PR #3056, opened 2017-05-25
Clean up Xapi_vm_lifecycle.check_operation_error

gaborigloi:
Get rid of the unnecessary get_info function.

Signed-off-by: Gabor Igloi <gabor.igloi@citrix.com>

PR #3276, opened 2017-10-10
VM.update_allowed_operations: cache some common values when checking operation errors in a loop

gaborigloi:
By specialising the function for a given VM.

### DIFF
--- a/ocaml/xapi/xapi_vm_lifecycle.ml
+++ b/ocaml/xapi/xapi_vm_lifecycle.ml
@@ -367,7 +367,7 @@ let maybe_get_guest_metrics ~__context ~ref =
     corresponding to the first error found. Checking stops at the first error.
     The "strict" param sets whether we require feature-flags for ops that need guest
     support: ops in the suspend-like and shutdown-like categories. *)
-let check_operation_error ~__context ~ref ~op ~strict =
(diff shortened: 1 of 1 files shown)


gaborigloi:
:warning: I have checked the code, and realized that there are even more values we could extract this way, possibly combined with `lazy` (which is not thread safe, but should be okay in this case), will try that when I get a chance.

gaborigloi:
This should undo the performance greression I introduced in https://github.com/xapi-project/xen-api/pull/3056

gaborigloi:
The changes are minimal, it's best to view them with `git` and the `--color-words` option.

lindig:
That's cool. I would have liked the explanation @gaborigloi gave me in the commit message. The idea is to to use partial evaluation of arguments and by this avoiding re-evaluation when only remaining arguments are applied. I'll have to take another look to understand what is going on.

jonludlam:
Ah, this is neat. Looks good to me!

gaborigloi:
I've update the commit message, but GitHub did not pick up the change. I'll close this PR and reopen it as a new one.

gaborigloi:
New PR is: https://github.com/xapi-project/xen-api/pull/3301

PR #3277, opened 2017-10-10
Revert "Factor out all_vm_operations"

gaborigloi:
Looping through more elemnts during VM.update_allowed_operations has a
performance impact. This commit restores the list of operations that we
check in VM.update_allowed_operations to the original set, which does
not include all VM operations.

This reverts commit 5497ab7b4569cd3422c0674295b92619294b8a09.

### DIFF
--- a/ocaml/xapi/test_vm_check_operation_error.ml
+++ b/ocaml/xapi/test_vm_check_operation_error.ml
@@ -1,4 +1,54 @@
 
+let all_vm_operations =
+  [ `assert_operation_valid
+  ; `awaiting_memory_live
+  ; `call_plugin
+  ; `changing_VCPUs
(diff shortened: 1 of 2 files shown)


gaborigloi:
We should decide how important this is, whether it's worth reverting or not.

gaborigloi:
I've run some perf tests with this change, and for each of those measurements, this change either improves performance a little bit, or it stays the same.
So we should decide whether we want to revert my change or not. I think we should revert it, unless we really want to include all the VM operations in VM.update_allowed_operations for some reason.

mseri:
Ping @robhoes @jonludlam  @edwintorok

robhoes:
Looks fine, but I've restarted Travis, because the build had failed.

PR #3279, opened 2017-10-11
CP-24907: Refine vusb in clear_scheduled_to_be_resident_on

taoyongd:
1. As vusb has no scheduled_to_be_resident_on field, so we
   don't need to do it for vusb.


taoyongd:
This commit will be included in PR https://github.com/xapi-project/xen-api/pull/3282

PR #3280, opened 2017-10-12
CA-267661: Fix Tools SR PBD.device_config on RPU

sharady:


sharady:
@jonludlam would you mind to review this please. Thanks

ocaml/xapi/xapi_db_upgrade.ml
```diff
@@ -453,6 +453,21 @@ let remove_restricted_pbd_keys = {
       ) (Db.PBD.get_all ~__context)
 }
 
+let update_tools_sr_pbd_device_config = {
+  description = "Updating Tools SR PBD.device_config";
+  version = (fun x -> x <= creedence);
+  fn = fun ~__context ->
+    let tools_srs = List.filter (fun self -> Db.SR.get_is_tools_sr ~__context ~self) (Db.SR.get_all ~__context) in
+    begin match tools_srs with
+    | sr :: others ->
```
mseri: Is it ok here to get any of them (the first apearing in this case) or we need to make sure that is a specific one?

sharady: It is supposed to be unique, but multiple records suggest duplicates of each other. So we can pick any of them.
Taken from https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/dbsync_master.ml#L140

jonludlam:
This looks good to me.

PR #3282, opened 2017-10-12
CP-24611: Pass host to PUSB scan function

taoyongd:
This PR contains follow modification.
1. PUSB.scan should take the host as a parameter, this PR is to add it.
2.  Add feature restriction for usb passthrough.
3. Refine error message when create VUSB which the USB_group already bind with vusbs.
4. Add specific error when set passthrough_enabled.


ocaml/xapi/message_forwarding.ml
```diff
@@ -852,13 +852,7 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
            Db.VGPU.set_scheduled_to_be_resident_on ~__context
              ~self:vgpu
              ~value:Ref.null)
-        (Db.VM.get_VGPUs ~__context ~self:vm);
-      List.iter
-        (fun vusb ->
-           Db.VUSB.set_attached ~__context
-             ~self:vusb
-             ~value:Ref.null)
-        (Db.VM.get_VUSBs ~__context ~self:vm)
```
jonludlam: I'm not convinced we don't need this - we're setting 'attached' in Vusbops at the same time that we're setting the VGPU scheduled_to_be_resident_on, so presumably we need to clear it on the error path in roughly the same place too?

ocaml/xapi/xapi_pusb.ml
```diff
@@ -29,8 +29,14 @@ let create ~__context ~uSB_group ~host ~other_config ~path
   debug "PUSB ref='%s' created" (Ref.string_of pusb);
   pusb
 
+let host_ref : [`host] Ref.t ref = ref Ref.null
+
 let scan_start ~__context usbs =
   let host = Helpers.get_localhost ~__context in
+  if !host_ref <> host then
```
jonludlam: we don't need this.

ocaml/xapi/xapi_pusb.ml
```diff
@@ -74,20 +80,21 @@ let start_thread f =
           ()
       done) ())
 
-let scan ~__context =
+let scan ~__context ~host =
   (* notify that scan is required. *)
   Mutex.execute mutex (fun () ->
+      host_ref := host;
```
jonludlam: again, no need for this. You're protecting against the call be routed to the wrong host, and if _that_ logic went wrong, xapi would fall apart completely :-)

ocaml/xapi/xapi.ml
```diff
@@ -919,7 +919,7 @@ let server_init() =
           "Synchronising tunnels on slave with master", [Startup.OnlySlave; Startup.NoExnRaising], Sync_networking.copy_tunnels_from_master ~__context;
 
           "SR scanning", [ Startup.OnlyMaster; Startup.OnThread ], Xapi_sr.scanning_thread;
-          "PUSB scanning", [], (fun () -> Xapi_pusb.scan_thread ~__context);
+          "PUSB scanning", [], (fun () -> Xapi_pusb.scan_thread ~__context ~host:(Helpers.get_localhost ~__context));
```
jonludlam: No need for host here.

ocaml/xapi/xapi_pusb.ml
```diff
@@ -29,8 +29,14 @@ let create ~__context ~uSB_group ~host ~other_config ~path
   debug "PUSB ref='%s' created" (Ref.string_of pusb);
   pusb
 
+let host_ref : [`host] Ref.t ref = ref Ref.null
```
lindig: This is a global variable. Can it contain anything other than the reference for localhost? And if not, how is this interacting with threads?

ocaml/idl/datamodel.ml
```diff
@@ -1499,6 +1499,8 @@ let _ =
     ~doc:"The USB group does not contain any PUSBs." ();
   error Api_errors.too_many_vusbs [ "number" ]
     ~doc:"The VM has too many VUSBs." ();
+  error Api_errors.usb_group_conflict [ "USB_group" ]
+    ~doc:"The USB_group must contain no vusb when creating VUSB object as now USB_group only have one PUSB." ();
```
jonludlam: "USB_groups are currently restricted to contain no more than one VUSB"

ocaml/xapi/message_forwarding.ml
taoyongd: In this function, let forward_to_suitable_host ~local_fn ~__context ~vm ~snapshot ?host_op op =
it will always call  
           if Db.is_valid_ref __context vm
                then clear_scheduled_to_be_resident_on ~__context ~vm))
Then when we start vm which already bind with a vusb, the attached will be cleared . I think this is not as expected.

ocaml/xapi/xapi_pusb.ml
taoyongd: how we pass the host here? just 
let scan_start ~__context usbs host =
?

ocaml/xapi/message_forwarding.ml
jonludlam: Ah yes, I see - in which case we're probably setting `attached` in the wrong place too. We should move all that logic into VM.start perhaps?

ocaml/xapi/xapi_pusb.ml
jonludlam: we don't need to pass it at all. It's always localhost.

ocaml/xapi/message_forwarding.ml
taoyongd: I think we can do it in  let allocate_vm_to_host ~__context ~vm ~host ~snapshot ?host_op () =
Originally I add a new function to do this which is ok I think.

ocaml/xapi/xapi_pusb.ml
taoyongd: updated.

ocaml/xapi/xapi_pusb.ml
taoyongd: updated.

ocaml/xapi/xapi.ml
taoyongd: updated.

ocaml/xapi/xapi_pusb.ml
taoyongd: no need host_ref here now.

ocaml/idl/datamodel.ml
taoyongd: updated.

jonludlam:
We'll merge this, but we'll need to address the wider issue of handling `attached` later.

taoyongd:
yes, https://issues.citrite.net/browse/CP-25165 will handle this.

PR #3284, opened 2017-10-12
CA-267946: Avoid Db.is_valid_ref RPC call when Ref is NULL

edwintorok:
e.g. in `update_allowed_operations` avoid repeated RPC calls that we know are
always going to return false when the field is not set.

Of course we'll still do the RPC call when it is not NULL to validate the ref.

### DIFF
--- a/ocaml/xapi/db.ml
+++ b/ocaml/xapi/db.ml
@@ -17,6 +17,8 @@
 
 include Db_actions.DB_Action
 let is_valid_ref __context r =
-  let t = Context.database_of __context in
-  let module DB = (val (Db_cache.get t) : Db_interface.DB_ACCESS) in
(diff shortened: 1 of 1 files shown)


jonludlam:
Incidentally, update_allowed_operations should be almost always called on the master, so this shouldn't affect that call too much.

edwintorok:
Could `force_state_reset` (which calls `update_allowed_operations`)  which is called from VM clone be executed somewhere other than the master? (e.g. if the VM is using local storage)

mseri:
Would this clean some of the "invalid reference ..." backtraces? Or the `DB.get` would not trigger any?

edwintorok:
It looks like exceptions are caught and just a bool is returned whether the ref is valid, and the `get` simply returns a first class module.

PR #3285, opened 2017-10-12
CA-267946: Check operation db memoize

edwintorok:
This improves upon https://github.com/xapi-project/xen-api/pull/3276 by storing the results of expensive Db RPC calls, and by only making them when necessary. 

(it implements the lazy operation in a safe way, `Lazy.t` is documented to be not thread-safe, and can raise `Lazy.Undefined` if used from 2 threads. although in update_allowed_operations we only ever use it from one thread, I didn't want to set a bad precedent of using thread-unsafe functions in threaded code).

On a 2 host pool this improves VM.clone from ~0.7s to ~0.5s, performance test pending for a 4 host pool.
Opened the PR to discuss the general approach.

### DIFF
--- a/ocaml/xapi/db.ml
+++ b/ocaml/xapi/db.ml
@@ -40,7 +40,7 @@ module LazyMemo : sig
       i.e. this is a shorter syntax for [memo f !!v] *)
   val ($!) : ('a -> 'b) -> 'a t -> 'b t
 end = struct
-  type 'a t = 'a Records.lzy ref
+  type 'a t = 'a Db_cache.lzy ref
 
 (* Lazy.t is not thread-safe (it can raise Lazy.undefined), so use a ref instead.
    We only want to initiate the (potentially RPC) call when we know we are
@@ -49,11 +49,11 @@ end = struct
    same value multiple times, but in assert_valid you only need some values for
    the current operation.
 *)
-  let memo f v = ref (Records.ToGet (fun () -> f v))
+  let memo f v = ref (Db_cache.ToGet (fun () -> f v))
 
-  let memoU f = ref (Records.ToGet f)
+  let memoU f = ref (Db_cache.ToGet f)
 
-  let (!!) = Records.lzy_get
+  let (!!) = Db_cache.lzy_get
 
   let ($!) f v = memo f !!v
 end
--- a/ocaml/xapi/db.ml
+++ b/ocaml/xapi/db.ml
@@ -20,3 +20,40 @@ let is_valid_ref __context r =
   let t = Context.database_of __context in
   let module DB = (val (Db_cache.get t) : Db_interface.DB_ACCESS) in
   DB.is_valid_ref t (Ref.string_of r)
+
+module LazyMemo : sig
+  type 'a t
+
+  (** [memo f v] constructs an ['a LazyMemo.t] lazy value for [f v] that is only evaluated the first time it is used.
+  *)
+  val memo : ('a -> 'b) -> 'a -> 'b t
+
+  (** [memoU f] constructs an ['a LazyMemo.t] equivalent to [memo f ()] *)
+  val memoU : (unit -> 'a) -> 'a t
+
+  (** [!! v] forces the evaluation of the lazy value [v].
+      If the evaluation succeeds the result is memoized and [f v] is not called again for the liftime of [v].
+  *)
+  val (!!) : 'a t -> 'a
+
+  (** [f $ v] constructs a new lazy value that will call [f !!v] when evaluated,
+      i.e. this is a shorter syntax for [memo f !!v] *)
+  val ($!) : ('a -> 'b) -> 'a t -> 'b t
+end = struct
+  type 'a t = 'a Records.lzy ref
+
+(* Lazy.t is not thread-safe (it can raise Lazy.undefined), so use a ref instead.
+   We only want to initiate the (potentially RPC) call when we know we are
+   going to need the result, and then memoize it for the lifetime of this value.
+   Useful for example in update_allowed_operations where sometimes you need the
+   same value multiple times, but in assert_valid you only need some values for
+   the current operation.
+*)
+  let memo f v = ref (Records.ToGet (fun () -> f v))
+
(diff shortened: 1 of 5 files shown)


ocaml/xapi/xapi_vm_lifecycle.ml
```diff
@@ -367,194 +365,205 @@ let maybe_get_guest_metrics ~__context ~ref =
...
+     let current_error = check current_error (fun () ->
+         let is_cbt_enabled vdi = try Db.VDI.get_cbt_enabled ~__context ~self:vdi with _ -> false in
+         if op = `migrate_send && List.exists is_cbt_enabled !!vdis then begin
+           let vdi = List.find is_cbt_enabled !!vdis in
+           Some (Api_errors.vdi_cbt_enabled, [ Ref.string_of vdi ])
+         end else None
+       ) in
+
+     let current_error = check current_error (fun () ->
+         if Helpers.rolling_upgrade_in_progress ~__context &&
```
gaborigloi: This also does a `Db.Pool.get_other_config`, so we could move this too out of the loop, but I guess that's not a big deal.

edwintorok: thanks for spotting that, I think we  could also swap the arguments of && to shortcircuit evaluation

ocaml/xapi/xapi_vm_lifecycle.ml
```diff
@@ -367,194 +365,207 @@ let maybe_get_guest_metrics ~__context ~ref =
...
+
+
+     (* Check if the VM is a control domain (eg domain 0).            *)
+     (* FIXME: Instead of special-casing for the control domain here, *)
+     (* make use of the Helpers.ballooning_enabled_for_vm function.   *)
+     let current_error = check current_error (fun () ->
+         if (op = `changing_VCPUs || op = `destroy) && Helpers.is_domain_zero ~__context !!vm_ref
+         then Some (Api_errors.operation_not_allowed, ["This operation is not allowed on dom0"])
+         else if vmr.Db_actions.vM_is_control_domain
+              && op <> `data_source_op
```
mseri: Using pattern matching here instead of spearate manual checks would be optimized by the compiler, but I am not sure it's worth it.

ocaml/xapi/xapi_vm_lifecycle.ml
```diff
@@ -367,194 +365,207 @@ let maybe_get_guest_metrics ~__context ~ref =
...
+            (Pervasiveext.maybe_with_default true
+               (fun gm -> let other = gm.Db_actions.vM_guest_metrics_other in
+                 not (List.mem_assoc "feature-quiesce" other || List.mem_assoc "feature-snapshot" other))
+               !!vmgmr)
+         then Some (Api_errors.vm_snapshot_with_quiesce_not_supported, [ ref_str ])
+         else None) in
+
+     (* Check for an error due to VDI caching/reset behaviour *)
+     let current_error = check current_error (fun () ->
+         if op = `checkpoint || op = `snapshot || op = `suspend || op = `snapshot_with_quiesce
```
mseri: here as well, changing it to a `match op` would be optimised by the compiler

ocaml/xapi/db.ml
```diff
@@ -20,3 +20,40 @@ let is_valid_ref __context r =
...
+  val memoU : (unit -> 'a) -> 'a t
+
+  (** [!! v] forces the evaluation of the lazy value [v].
+      If the evaluation succeeds the result is memoized and [f v] is not called again for the liftime of [v].
+  *)
+  val (!!) : 'a t -> 'a
+
+  (** [f $ v] constructs a new lazy value that will call [f !!v] when evaluated,
+      i.e. this is a shorter syntax for [memo f !!v] *)
+  val ($) : ('a -> 'b) -> 'a t -> 'b t
```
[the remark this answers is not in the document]
edwintorok: We could call it `($!)`, the problem with `(!$)` is that  AFAICT `!..` can only be a prefix operator (except for `!=`).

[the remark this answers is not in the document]
mseri: fine with me

edwintorok:
I've rebased on latest master due to merge conflicts, had to move one function to avoid a dependency loop too.

edwintorok:
The value is only cached for the duration of `update_allowed_operations` or `check_operation_error` itself, not globally. There could be race conditions of course if `update_allowed_operations` / `assert_operation_valid` is ever called without holding locks , but `assert_operation_valid` shouldn't be called without holding locks.
I'd be fine with delaying to merge this until after we branched for the current release if the performance regression is acceptable.

robhoes:
> I'd be fine with delaying to merge this until after we branched for the current release if the performance regression is acceptable.

I'm afraid that the performance regression isn't acceptable, so we should try our best to verify that this is correct.

gaborigloi:
Or should we just merge the first part of this PR, https://github.com/xapi-project/xen-api/pull/3276 first, as I think that's pretty low risk and @edwintorok said it made a difference?

edwintorok:
Yes I've kept the caching commits separate, if @gabori updates his PR to fix the merge conflict (or take https://github.com/xapi-project/xen-api/pull/3285/commits/72a792887bec907046a782bdf85848ded214202e which is already rebased) we could first merge that and decide on this one later.
@gabori's change improves performance of `update_allowed_operations` too, especially on single node configurations, and I don't see any particular downsides for it.

robhoes:
Oh right, I didn't read carefully. I agree with @edwintorok and @gaborigloi .

gaborigloi:
I took @edwintorok 's commit and updated my PR: https://github.com/xapi-project/xen-api/pull/3276

edwintorok:
In the long term it would be nice if we could find some solution that doesn't regress everytime we add more fields/features, maybe:

- have each host be responsible for the VMs it is running (which would also help with pool scalability), although that is quite an invasive change
- be more declarative in the "locks" an operation needs, and compute the operations possible from that (similar to `blocked_operations`), e.g. with a list of which operation is blocked by which operations and an operation is available if its block list is empty, but again that is an invasive change
...

PR #3288, opened 2017-10-13
CP-22019: Test VDI.data_destroy

minishrink:
Test that VDI.data_destroy:
- updates VDI.type to cbt_metadata
- updates content_id to: "/No content: this is a cbt_metadata VDI/"

### DIFF
--- a/ocaml/xapi/quicktest_cbt.ml
+++ b/ocaml/xapi/quicktest_cbt.ml
@@ -13,32 +13,36 @@
  *)
 
 open Client (* import module *)
+open Client (* import namespace for convenience, Client.VDI.func -> VDI.func *)
 open Quicktest_common
+
+(* Throw this exception within assertion helpers when assertion fails, instead of
+ * actually failing the test within a helper called from the main test body *)
 exception Test_failed of string
-open Client (* import namespace, don't have to type Client.VDI.function each time *)
 
-(* Helper for test failure due to unexpected error *)
+(* Helper for test failure due to unexpected error
+ * Can fail test here as this is called outside test body *)
 let report_failure error test =
-  failed test (Printf.sprintf "%s failed: %s" test.name
-                 (ExnHelper.string_of_exn error))
+  failed test (Printf.sprintf "%s failed: %s" test.name (ExnHelper.string_of_exn error))
 
-(* Define exception so that if test fails, exception is passed to try-with statement and fails there
- * so that the test only fails once and doesn't erroneously assume the test never started *)
+let get_cbt_status ~session_id ~vDI = VDI.get_cbt_enabled ~session_id ~rpc:!rpc ~self:vDI
+
 let test_assert ~test op ~msg =
   if not op then raise (Test_failed msg)
 
 let test_compare ~test left_op right_op ~msg =
   let op = (left_op = right_op) in test_assert ~test op ~msg
 
-let get_cbt_status ~session_id ~vDI = VDI.get_cbt_enabled ~session_id ~rpc:!rpc ~self:vDI
-
+(* This naming is used to identify VDIs to destroy later on *)
+let name_label = "qt-cbt"
 let name_description = "VDI for CBT quicktest"
-let make_vdi_from ~session_id ~sR = (* SR has VDI.create as allowed *)
+(* This is only called on VDI-create capable SRs *)
+let make_vdi_from ~session_id ~sR =
   VDI.create
     ~sR
     ~session_id
     ~rpc:!rpc
-    ~name_label:"qt-cbt"
+    ~name_label
     ~name_description
     ~_type:`user
     ~sharable:false
@@ -50,11 +54,13 @@ let make_vdi_from ~session_id ~sR = (* SR has VDI.create as allowed *)
     ~sm_config:[]
 
 
-(**** Test declarations ****)
+(* ******************
+ * Test declarations
+ * ******************)
 
-(* Test data_destroy and snapshot update the necessary fields *)
+(* Test enable/disable_cbt, data_destroy, and snapshot update the necessary fields *)
 let vdi_data_destroy_test ~session_id ~vDI =
-  let data_destroy_test = make_test "Testing VDI.data_destroy" 4 in
+  let data_destroy_test = make_test "Testing VDI.{enable/disable_cbt, data_destroy, snapshot}" 4 in
   try
     start data_destroy_test;
     debug data_destroy_test "Enabling CBT on original VDI";
@@ -63,7 +69,7 @@ let vdi_data_destroy_test ~session_id ~vDI =
       (get_cbt_status ~session_id ~vDI)
       ~msg:"VDI.enable_cbt failed";
 
-    debug data_destroy_test "Snapshotting original VDI.";
+    debug data_destroy_test "Snapshotting original VDI";
     let newvdi = VDI.snapshot ~session_id ~rpc:!rpc ~vdi:vDI ~driver_params:[] in
     test_assert ~test:data_destroy_test
       (get_cbt_status ~session_id ~vDI:newvdi)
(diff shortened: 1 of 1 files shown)


ocaml/xapi/quicktest_cbt.ml
```diff
@@ -70,6 +73,40 @@ let enable_disable_cbt_test ~session_id ~vDI =
   | Test_failed msg -> failed enable_cbt_test msg
   | e -> report_failure e enable_cbt_test
 
+(* Test data_destroy and snapshot update the necessary fields *)
```
gaborigloi: We discussed that we could remove the enable/disable test, because this test includes that one.

ocaml/xapi/quicktest_cbt.ml
```diff
@@ -49,27 +52,46 @@ let make_vdi_from ~session_id ~sR = (* SR has VDI.create as allowed *)
 
 (**** Test declarations ****)
 
-(* Test enable/disable CBT, test cbt_enabled:false for new VDI *)
-let enable_disable_cbt_test ~session_id ~vDI =
-  let enable_cbt_test = make_test "Testing VDI.enable/disable_CBT" 4 in
+(* Test data_destroy and snapshot update the necessary fields *)
```
gaborigloi: We also test here that `VDI.enable_cbt` and `VDI.disable_cbt` update the `cbt_enabled` field of the VDI accordingly when they return without errors.

ocaml/xapi/quicktest_cbt.ml
```diff
@@ -49,27 +52,46 @@ let make_vdi_from ~session_id ~sR = (* SR has VDI.create as allowed *)
...
-    test_assert ~test:enable_cbt_test
-      (not (get_cbt_status ~session_id ~vDI)) (* disable_cbt fails *)
-      ~msg:"VDI.disable_CBT failed";
-    success enable_cbt_test
-  with
-  | Test_failed msg -> failed enable_cbt_test msg
-  | e -> report_failure e enable_cbt_test
+    test_assert ~test:data_destroy_test
+      (not (get_cbt_status ~session_id ~vDI))
+      ~msg:"VDI.disable_cbt failed";
```
gaborigloi: After disabling CBT on the VDI, we could snapshot it again to see if a `false` `cbt_enabled` field also carries over properly. But it's fine to do that in a followup PR.

ocaml/xapi/quicktest_cbt.ml
```diff
@@ -81,13 +103,17 @@ let test ~session_id =
...
+        [ `vdi_enable_cbt ; `vdi_data_destroy ; `vdi_snapshot ]
       ]
       |> List.iter
-        (fun (test,list_vdi_ops) ->
+        (fun (test , list_vdi_ops) ->
            if List.for_all (fun vdi_op -> List.mem vdi_op sr_ops) list_vdi_ops
-           then test ()
+           then begin
+             debug cbt_test "Creating VDI...";
+             let vDI = make_vdi_from ~session_id ~sR in
```
gaborigloi: If I'm not mistaken, now we're creating a VDI for each test, so we can remove the old code at line 143, which used to create the global VDI share between all the tests?

ocaml/xapi/quicktest_cbt.ml
minishrink: In the same test, I presume?

ocaml/xapi/quicktest_cbt.ml
minishrink: You are correct.

ocaml/xapi/quicktest_cbt.ml
gaborigloi: Yes, I though we could do it here to save time.

ocaml/xapi/quicktest_cbt.ml
```diff
@@ -47,61 +54,93 @@ let make_vdi_from ~session_id ~sR = (* SR has VDI.create as allowed *)
...
   try
-    start enable_cbt_test;
-    test_assert ~test:enable_cbt_test
-      (not (get_cbt_status ~session_id ~vDI))
-      ~msg:"VDI.cbt_enabled field should be set to false for new VDIs";
+    start data_destroy_test;
+    debug data_destroy_test "Enabling CBT on original VDI";
     VDI.enable_cbt ~session_id ~rpc:!rpc ~self:vDI;
-    test_assert ~test:enable_cbt_test
+    test_assert ~test:data_destroy_test
```
gaborigloi: To simplify things, we could specialize this for this test to avoid having to repeatedly pass the `~test` argument:
```ocaml
let data_destroy_test =  ... in
let test_assert = test_assert ~test:data_destroy_test in
...
```

ocaml/xapi/quicktest_cbt.ml
```diff
@@ -47,61 +54,93 @@ let make_vdi_from ~session_id ~sR = (* SR has VDI.create as allowed *)
...
       (get_cbt_status ~session_id ~vDI)
       ~msg:"VDI.enable_cbt failed";
+
+    debug data_destroy_test "Snapshotting original VDI";
```
gaborigloi: We could also do the same as above, but for `debug`.

ocaml/xapi/quicktest_cbt.ml
```diff
@@ -47,61 +54,93 @@ let make_vdi_from ~session_id ~sR = (* SR has VDI.create as allowed *)
...
+    test_assert ~test:data_destroy_test
+      (not (get_cbt_status ~session_id ~vDI))
+      ~msg:"VDI.disable_cbt failed";
+
+    debug data_destroy_test "Destroying snapshot VDI data";
+    VDI.data_destroy ~session_id ~rpc:!rpc ~self:newvdi;
+    test_compare ~test:data_destroy_test
+      (VDI.get_type ~session_id ~rpc:!rpc ~self:newvdi)
+      `cbt_metadata
+      ~msg:"VDI.data_destroy failed to update VDI.type";
```
gaborigloi: We could also test here that `VDI.cbt_enabled` is still true for the snapshot VDI.

ocaml/xapi/quicktest_cbt.ml
```diff
@@ -47,61 +54,93 @@ let make_vdi_from ~session_id ~sR = (* SR has VDI.create as allowed *)
...
-    test_assert ~test:enable_cbt_test
+    test_assert ~test:data_destroy_test
       (get_cbt_status ~session_id ~vDI)
       ~msg:"VDI.enable_cbt failed";
+
+    debug data_destroy_test "Snapshotting original VDI";
+    let newvdi = VDI.snapshot ~session_id ~rpc:!rpc ~vdi:vDI ~driver_params:[] in
+    test_assert ~test:data_destroy_test
+      (get_cbt_status ~session_id ~vDI:newvdi)
+      ~msg:"VDI.snapshot failed, cbt_enabled field didn't carry over";
```
gaborigloi: I think we could also test that the snapshot also correctly inherits a `false` `cbt_enabled` field: if we snapshot a VDI on which CBT isn't enabled, the `cbt_enabled` field of the snapshot should also be `false`.

ocaml/xapi/quicktest_cbt.ml
minishrink: Maybe it'd be easier if I just renamed the test object to `test` in each individual test, so field punning takes care of the rest? Unless that loses clarity?

gaborigloi: I think it's the best if we minimize the amount of things that one has to read by completely omitting this parameter, in my opinion that makes the test clearer and easier to understand, but this is probably subjective.

ocaml/xapi/quicktest_cbt.ml
```diff
@@ -60,44 +60,45 @@ let make_vdi_from ~session_id ~sR =
 
 (* Test enable/disable_cbt, data_destroy, and snapshot update the necessary fields *)
 let vdi_data_destroy_test ~session_id ~vDI =
-  let data_destroy_test = make_test "Testing VDI.{enable/disable_cbt, data_destroy, snapshot}" 4 in
+  let test = make_test "Testing VDI.{enable/disable_cbt, data_destroy, snapshot}" 4 in
   try
-    start data_destroy_test;
-    debug data_destroy_test "Enabling CBT on original VDI";
+    start test;
+    let debug_test msg = debug test msg in
```
gaborigloi: Just wanted to mention that this can be simplified to
```ocaml
let debug_test = debug test in
```
because of higher-order functions. But there is no need to change this.

ocaml/xapi/quicktest_cbt.ml
```diff
@@ -47,61 +54,106 @@ let make_vdi_from ~session_id ~sR = (* SR has VDI.create as allowed *)
...
+      (get_cbt_status ~session_id ~vDI:snapshot)
+      ~msg:"VDI.snapshot failed, cbt_enabled field didn't carry over";
+
+    debug_test "Disabling CBT on original VDI";
     VDI.disable_cbt ~session_id ~rpc:!rpc ~self:vDI;
-    test_assert ~test:enable_cbt_test
-      (not (get_cbt_status ~session_id ~vDI)) (* disable_cbt fails *)
-      ~msg:"VDI.disable_CBT failed";
-    success enable_cbt_test
+    test_assert ~test
```
gaborigloi: We could do the same thing for `test_assert` as for `debug`, to simplify all these assertions:
```
let debug_test msg = debug test msg in
let test_assert = test_assert ~test in
...
```
and then we won't have to pass the `~test` param to `test_assert`.

gaborigloi:
@robhoes Can we merge this into master now, or should it wait / go to a different branch.

gaborigloi:
I'm gonna test this on a branch to see if the BVT still passes.

gaborigloi:
The quicktest part of the BVT succeeded, merging.

PR #3289, opened 2017-10-13
CP-22019: Test VDI.clone and VDI.copy

minishrink:
Test VDI.clone and VDI.copy set cbt_enabled to false when called on a CBT-enabled VDI, as specified in CBT design doc v13.

### DIFF
--- a/ocaml/xapi/quicktest_cbt.ml
+++ b/ocaml/xapi/quicktest_cbt.ml
@@ -169,12 +169,14 @@ let test ~session_id =
       ]
       |> List.iter
         (fun (test , list_vdi_ops) ->
-           if List.for_all (fun vdi_op -> List.mem vdi_op sr_ops) list_vdi_ops
-           then begin
-             debug cbt_test "Creating VDI. . .";
-             let vDI = make_vdi_from ~sR ~session_id in
-             test ~vDI end
-           else debug cbt_test "SR lacks capabilities for this test, skipping"
+           begin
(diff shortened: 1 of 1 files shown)


ocaml/xapi/quicktest_cbt.ml
```diff
@@ -112,6 +110,42 @@ let vdi_data_destroy_test ~session_id ~vDI =
   | Test_failed msg -> failed test msg
   | e -> report_failure e test
 
+(* Check VDI.(copy, clone, and snapshot) all properly update cbt_enabled
```
gaborigloi: The comment should be updated: we're testing snapshot in the above test, not this one.

ocaml/xapi/quicktest_cbt.ml
```diff
@@ -130,6 +164,8 @@ let test ~session_id =
       [
         (fun ~vDI -> vdi_data_destroy_test ~session_id ~vDI) ,
         [ `vdi_enable_cbt ; `vdi_disable_cbt ; `vdi_data_destroy ; `vdi_snapshot ]
+      ; (fun ~vDI -> vdi_clone_copy_test ~session_id ~sR ~vDI),
+        [ `vdi_enable_cbt ; `vdi_create ; `vdi_clone ; `vdi_snapshot ]
```
gaborigloi: Same as above:  we're not calling VDI.snapshot, so the SR doesn't need to support it.

gaborigloi:
I'm testing it on a new branch.

gaborigloi:
The quicktests passed, merging.

PR #3290, opened 2017-10-13
CP-24982: passthrough_enabled must be true before passthrough usb to vm

taoyongd:


jonludlam:
Looks good.

PR #3291, opened 2017-10-13
CA-268761: Trailing whitespaces are lost by the xapi database

mseri:
This is an issue in how xapi serialises the fields in the database xml dump.
Xapi creates row tags that have all the fields set as attributes and
by spec spaces in attrributes are stripped and normalised.
Currently xapi uses an internal encoding/decoding that escapse the
character that would otherwise be stripped, with the possible
exception of one space at the beginnin of the content and one
at the end.

This PR changes this escape mechanism by escaping all white spaces
singularly instead of in pairs. The unescape mechanism is
kept backward compatible.

### DIFF
--- a/ocaml/database/xml_spaces.ml
+++ b/ocaml/database/xml_spaces.ml
@@ -66,7 +66,7 @@ let map2_unlikely f s =
     s
 
 let protect_fn = function
-  |  ' ', Some ' ' -> Replace (protect_char, '_' )
+  |  ' ',       _  -> Expand  (protect_char, '.' )
   | '\t',       _  -> Expand  (protect_char, 't' )
   | '\n',       _  -> Expand  (protect_char, 'n' )
   | '\r',       _  -> Expand  (protect_char, 'r' )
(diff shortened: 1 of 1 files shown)


mseri:
This is passing tests and quicktest, but we may want to have a BVT and a BST before merging

lindig:
Just a typo:

Currently xapi uses an internal encoding/decoding that _escapes_ the
character that would otherwise be stripped [...]

jonludlam:
@mseri can we do an upgrade test? I think that's more likely to find issues than a BVT/BST (which I don't think does an upgrade, but I'm not 100% sure)

mseri:
The BVT and BST are ok. I thought BST was doing upgrades already, I am running a couple of upgrade specific tests for additional safety

mseri:
@lindig @jonludlam The update/upgrade tests have all passed. I am going to get a db and test loading it with XenCenter just to double check but there should be no problem there as well.

robhoes:
@mseri Have you finished your testing?

mseri:
Just for reference: tests had been completed, only an aesthetic issue with XenCenter has been found

PR #3292, opened 2017-10-16
CA-269367: Fix missing row PUSB while updating VUSB

taoyongd:
This PR is to fix the error log that when updating VUSB  in update_vusb funciton, which throw missing row PUSB.

### DIFF
--- a/ocaml/xapi/xapi_xenops.ml
+++ b/ocaml/xapi/xapi_xenops.ml
@@ -1957,6 +1957,7 @@ let update_vusb ~__context (id: (string * string)) =
           let pusb, pusb_r =
             Db.VM.get_VUSBs ~__context ~self:vm
             |> List.map (fun vusb -> Db.VUSB.get_attached ~__context ~self:vusb)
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_xenops.ml
```diff
@@ -1957,6 +1957,7 @@ let update_vusb ~__context (id: (string * string)) =
           let pusb, pusb_r =
             Db.VM.get_VUSBs ~__context ~self:vm
             |> List.map (fun vusb -> Db.VUSB.get_attached ~__context ~self:vusb)
+            |> List.filter (fun attached -> attached <> Ref.null)
```
lindig: This looks clean but I wonder whether we should call `is_valid_ref` or similar. This would also catch stale references.

taoyongd: Right, make sense.

jonludlam:
I think travis is using pre-CBT rebase RPMs

lindig:
Why is Travis failing? It used to work for this branch.

PR #3293, opened 2017-10-16
CP-25174: Handle VDI and PUSB conflict

taoyongd:
1. When we enable the PUSB for passthrough, we need to
   check if the correponding VDI has a vbd bound, as
   this will cause conflict when we attach vdi to vm and
   also passthrough the usb to vm.


taoyongd:
Hi @jonludlam   when we enable passthrough for pusb we will remove  corresponding vdis.
If user passthrough the usb to vm, then we can not re-introduce the vdi through disable passthrough . But if user didn't pasthrough, just enable  and disable, the vdi can be re-displayed. Do we need remove the re-display code thoroughly? That is  If user enable, then the vdi will not re-display?

ocaml/xapi/xapi_pusb.ml
```diff
@@ -109,6 +109,9 @@ let set_passthrough_enabled ~__context ~self ~value =
            List.iter (fun rf ->
              try
                if (get_sm_usb_path ~__context rf) = pusb_path then begin
+                 let vbds = Db.VDI.get_VBDs ~__context ~self:rf in
+                 if vbds <> [] then
+                   raise (Api_errors.Server_error(Api_errors.pusb_vdi_conflict, [ Ref.string_of self; Ref.string_of rf ]));
                  Xapi_vdi.forget ~__context ~vdi:rf
```
lindig: I think function `set_passthrough_enabled` needs to be refactored. The nesting of the control flow is too deep already and `List.iter` inside `List.iter` suggests that something is not well thought out here.

More specifically, if `vbs<>[]` the `raise` will make sure that `Xapi_vdi.forget ~__context ~vdi:rf` is never reached - it that correct?

taoyongd: `Xapi_vdi.forget ~__context ~vdi:rf` can be reached if a vdi is ok before raise error. Will this has any issue?

taoyongd: BTW, as PUSB may have more than one vdi, once one of the vdi have vbd bound, we need to make sure the set_passthrough_enabled failed.

taoyongd:
@jonludlam , how do you think about this PR?

ocaml/idl/datamodel.ml
```diff
@@ -1508,6 +1508,8 @@ let _ =
     ~doc:"The USB device is currently attached to a VM." ();
   error Api_errors.passthrough_not_enabled [ "PUSB"]
     ~doc:"The passthrough_enabled must be true before passthrough usb to vm." ();
+  error Api_errors.pusb_vdi_conflict [ "PUSB"; "VDI" ]
+    ~doc:"The PUSB and corresponding VDI conflicts as this vdi has vdb bound." ();
```
jonludlam: "The VDI corresponding to this PUSB has existing VBDs"

taoyongd: ok , error message refined.

jonludlam:
I think refactoring set_passthrough_enabled is a job for a different PR.

taoyongd:
Ok @jonludlam  , will raise another PR to refactor this.

PR #3294, opened 2017-10-16
CP-25172: Improve the error handling of scanner

TalonsLee:
CP-25172: Improve the error handling of scanner

If the policy file cannot be parsed, return 1 to indicate the error
instead of just ignoring the line.

### DIFF
--- a/scripts/usb-policy.conf
+++ b/scripts/usb-policy.conf
@@ -1,3 +1,6 @@
+# When you change this file, run 'xe pusb-scan' to confirm
+# the file can be parsed correctly.
+#
 # Syntax is an ordered list of case insensitive rules where # is line comment
 #  and each rule is (ALLOW | DENY) : ( match )*
 #  and each match is (class|subclass|prot|vid|pid|rel) = hex-number
--- a/scripts/usb_scan.py
+++ b/scripts/usb_scan.py
@@ -37,6 +37,11 @@ def log_list(l):
         log.debug(s)
 
 
+def log_exit(m):
+    log.error(m)
+    sys.exit(m)
+
+
 def hex_equal(h1, h2):
     """ check if the value of hex string are equal
 
@@ -375,15 +380,10 @@ class Policy(object):
         return False
 
     @staticmethod
-    def log_exit(m):
-        log.error(m)
-        sys.exit(m)
-
-    @staticmethod
     def parse_error(pos, end, target, line):
-        Policy.log_exit(
-            "Malformed policy rule, unable to parse '{}'. Ignored line: {}" \
-                .format(target[pos:end], line))
+        log_exit(
+            "Malformed policy rule, unable to parse '{}'. Ignored line: {}"
+            .format(target[pos:end], line))
 
     def parse_line(self, line):
         """ parse one line of policy file, generate rule, and append it to
@@ -415,8 +415,8 @@ class Policy(object):
             action, target = [part.strip() for part in line.split(":")]
         except ValueError as e:
             if line.rstrip():
-                self.log_exit("Caught error {}. Ignoring malformed rule line:"
-                              "{}".format(str(e), line))
+                log_exit("Caught error {}. Ignoring malformed rule line: {}"
+                         .format(str(e), line))
             # empty line, just return
             return
 
@@ -428,7 +428,7 @@ class Policy(object):
         elif action.lower() == "deny":
             rule[self._ALLOW] = False
         else:
-            self.log_exit("Malformed action'{}', ignoring rule line: {}".format(
+            log_exit("Malformed action'{}', ignoring rule line: {}".format(
                 action, line))
 
         # 4. parse key=value pairs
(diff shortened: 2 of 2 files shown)


scripts/usb_scan.py
```diff
@@ -374,10 +375,15 @@ def check_hex_length(self, name, value):
         return False
 
     @staticmethod
-    def log_parse_error(pos, end, target, line):
-        log.error(
-            "Malformed policy rule, unable to parse '{}'. Ignored line: {}"
-            .format(target[pos:end], line))
+    def log_exit(m):
+        log.error(m)
```
lindig: In the case of an error, is the error logged twice? Here, and when the exception is handled? Is using `sys.exit` with a message rather than an exit code idiomatic (I'm not a Python programmer).

TalonsLee: Comment message updated.

TalonsLee: And yes, 
the "log.error(m" will put log to /var/log/user.log
the sys.exit(m) will put log to stderr, so that xapi can get it.
Is it OK what we just call exit(1)?

krizex: I think it's better to keep it as is through the log will print twice. The first log is printed by `log.error` and output to `user.log`, and the second log is printed by `xapi` when catching `SystemExit exception` from the script, and the log will output to `xensource.log`. I think it will make it easier for us to track the log between `user.log` and `xensource.log` when debugging.

mseri: I agree with @krizex in this case. I'd leave it as is but I would probably move the function to the main scope, next to the log list. 

@lindig yes, using `sys.exit("message...")` is the idiomatic way to print the message on stderr and exit with 1.

scripts/usb_scan.py
```diff
@@ -374,10 +375,15 @@ def check_hex_length(self, name, value):
...
+        sys.exit(m)
+
+    @staticmethod
+    def parse_error(pos, end, target, line):
+        Policy.log_exit(
```
mseri: Why not `self.log_exit`?

scripts/usb_scan.py
```diff
@@ -616,7 +616,11 @@ def to_pusb(device):
         log.logToSyslog(level=logging.WARNING)
 
     # get usb info
-    devices, interfaces = get_usb_info()
+    try:
+        devices, interfaces = get_usb_info()
+    except Exception as e:
+        log.error("Failed to get usb info: {}".format(str(e)))
```
mseri: Make `log_exit` global and here use `log_exit` as well

scripts/usb_scan.py
krizex: Move the function to the top level will be good, so we can call it everywhere.

scripts/usb_scan.py
TalonsLee: OK. move log_exit() to main scope.

Because it's a staticmethod,
after moving it to main scope,  I can call it.

scripts/usb_scan.py
TalonsLee: Done.

scripts/usb_scan.py
```diff
@@ -375,15 +380,10 @@ def check_hex_length(self, name, value):
...
-        log.error(m)
-        sys.exit(m)
-
-    @staticmethod
     def parse_error(pos, end, target, line):
-        Policy.log_exit(
-            "Malformed policy rule, unable to parse '{}'. Ignored line: {}" \
-                .format(target[pos:end], line))
+        log_exit(
+            "Malformed policy rule, unable to parse '{}'. Ignored line: {}"
```
mseri: @lindig correctly pointed out that we are no longer skipping the line.
Change `Ignored line` to `Malformed line`.

scripts/usb_scan.py
```diff
@@ -415,8 +415,8 @@ def parse_line(self, line):
             action, target = [part.strip() for part in line.split(":")]
         except ValueError as e:
             if line.rstrip():
-                self.log_exit("Caught error {}. Ignoring malformed rule line:"
-                              "{}".format(str(e), line))
+                log_exit("Caught error {}. Ignoring malformed rule line: {}"
```
mseri: We are no longer skipping the line.
Change `Ignored line` to `Malformed line`.

scripts/usb_scan.py
```diff
@@ -428,7 +428,7 @@ def parse_line(self, line):
         elif action.lower() == "deny":
             rule[self._ALLOW] = False
         else:
-            self.log_exit("Malformed action'{}', ignoring rule line: {}".format(
+            log_exit("Malformed action'{}', ignoring rule line: {}".format(
```
mseri: We are no longer skipping the line.
Change `Ignored line` to `Malformed line`.

scripts/usb_scan.py
```diff
@@ -450,12 +450,12 @@ def parse_line(self, line):
                 self.parse_error(match.start(), match.end(), target, line)
 
             if not self.check_hex_length(name, value):
-                self.log_exit("hex'{}' length error, ignore line {}".format(
+                log_exit("hex'{}' length error, ignore line {}".format(
```
mseri: We are no longer skipping the line.
Change `Ignored line` to `Malformed line`.

scripts/usb_scan.py
```diff
@@ -450,12 +450,12 @@ def parse_line(self, line):
...
                     str(value), line))
 
             if name in rule:
-                self.log_exit("duplicated tag'{}' found, ignore line {}".
-                              format(name, line))
+                log_exit("duplicated tag'{}' found, ignore line {}".
```
mseri: We are no longer skipping the line.
Change `ignore line` to `. Malformed line`.

scripts/usb-policy.conf
```diff
@@ -5,9 +5,10 @@
 #  and each rule is (ALLOW | DENY) : ( match )*
 #  and each match is (class|subclass|prot|vid|pid|rel) = hex-number
 # Maximum hex value for class/subclass/prot is FF, and for vid/pid/rel is FFFF
```
mseri: I think this is the right thing to do, but this commit should be part of a different PR and not of this one

PR #3295, opened 2017-10-16
Should not check usb-passthrough licence when xapi starting

krizex:


lindig:
The code is fine. A slightly longer commit message would be useful:

```
Checking that the licensed USB passtrough feature is enabled must not happen during startup as it
could fail and this would break the startup sequence.
```

krizex:
Updated. thanks @lindig

PR #3297, opened 2017-10-17
CP-25186: Handle usb_policy error

taoyongd:
1. when usb policy file is modified, usb_scan script will exit(1).
  Xapi will log the error.

### DIFF
--- a/ocaml/xapi/xapi_pusb_helpers.ml
+++ b/ocaml/xapi/xapi_pusb_helpers.ml
@@ -90,8 +90,11 @@ let get_local_usb usbs =
 let get_script_stdout () =
   (** usb_scan is a script that will get all the usb details in current host, which will return json format data *)
   let usb_scan_script = "/opt/xensource/libexec/usb_scan.py" in
-  let stdout, stderr = Forkhelpers.execute_command_get_output usb_scan_script [] in
-  stdout
+  try
+    let stdout, stderr = Forkhelpers.execute_command_get_output usb_scan_script [] in
+    stdout
+  with Forkhelpers.Spawn_internal_error(stdout, stderr, Unix.WEXITED n) ->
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_pusb_helpers.ml
```diff
@@ -90,8 +90,12 @@ let get_local_usb usbs =
 let get_script_stdout () =
   (** usb_scan is a script that will get all the usb details in current host, which will return json format data *)
   let usb_scan_script = "/opt/xensource/libexec/usb_scan.py" in
-  let stdout, stderr = Forkhelpers.execute_command_get_output usb_scan_script [] in
-  stdout
+  try
+    let stdout, stderr = Forkhelpers.execute_command_get_output usb_scan_script [] in
+    stdout
+  with Forkhelpers.Spawn_internal_error(stdout, stderr, Unix.WEXITED n) as e ->
+    error "%s exitted with %d (%s; %s)" usb_scan_script n stdout stderr;
```
lindig: * I think including stdout and stderr in the log message is too much. The scanning script already logs a lot of errors by itself. I would suggest: "%s exited with %d" usb_scan_script n` is enough (`exitted` has a typo).
* This should not raise `Forkhelpers.Spawn_internal_error` but a more an internal xapi error.

taoyongd: Ok, make sense.  So just failwith is enough, right?

lindig: No. This is an API function and it should raise an `Api_errors.internal_error`

taoyongd: Ok, So we do not need error "%s exitted with %d (%s; %s)" usb_scan_script n. Can just raise an internal error with above messages. right?

taoyongd: @lindig  updated.

taoyongd:
hi @jonludlam . can this PR merge now?

PR #3298, opened 2017-10-17
CA-269588: BVT failed as the licence check issue.

taoyongd:
1. Should not check feature restriction as if the license expired
   the vm can not start.


lindig:
If this feature requires a license, isn't this the correct behaviour? Where is the license checked instead?

taoyongd:
Now in VUSB create will check the restriction.   If a USB already attached to vm, the user can continue to deatch/attach the USB, but can not attach other USBs

robhoes:
There are two options:

1. Have an assert like this, but _only call it when the VM actually has vUSBs_. Right now, it seems to be called for any VM.
2. Remove it, as in the PR.

Both are valid, depending on the requirements. Let's take this change now, and then sort out the requirements.

lindig:
Ok - we check the license when a USB device is introduced but not when it added to a VM. That's fine with me.

PR #3299, opened 2017-10-17
CP-25141 SCTX-2619 fix sync from xenopsd in on_xapi_restart

lindig:
When xapi starts, it queries xenopsd about the state of local VMs. In
SCTX-2619 we found that xenopsd passed data about a VM that was *not*
resident locally but was left over from a previous failed migration to
local host. This data was erroneously used to reset the power state of
that non-local VM. This commit fixes this.

If xenopsd reports the VM as running, it must indeed be resident on the
local host - so we reset the state of the VM to what xenopsd is
reporting.

If xenopsd reports the VM as suspended or halted, we only reset the
state of the VM if xapi believes the VM is resident on localhost.
Previously it would reset the state regardless - which caused the error.

### DIFF
--- a/ocaml/xapi/xapi_xenops.ml
+++ b/ocaml/xapi/xapi_xenops.ml
@@ -2023,17 +2023,19 @@ let on_xapi_restart ~__context =
       let vm = vm_of_id ~__context id in
       let xapi_power_state =
         xenapi_of_xenops_power_state (Some state.Vm.power_state) in
-      Xapi_vm_lifecycle.force_state_reset ~__context ~self:vm ~value:xapi_power_state;
       match xapi_power_state with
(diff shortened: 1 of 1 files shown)


jonludlam:
Is there adequate logging here? If force_state_reset doesn't log, we should make this function do more logging.

lindig:
I'll add `info` log lines.

lindig:
I made a second commit that adds logging on my branch but it does not show up here. https://github.com/lindig/xen-api/commit/1d0eb3feaa5532be3e966ed1e0a4133f183956d3

lindig:
There is some GitHub glitch here as is does not update the PR. Closing this and opening it again.

PR #3300, opened 2017-10-17
CP-25141 SCTX-2619 fix sync from xenopsd in on_xapi_restart

lindig:
When xapi starts, it queries xenopsd about the state of local VMs. In
SCTX-2619 we found that xenopsd passed data about a VM that was *not*
resident locally but was left over from a previous failed migration to
local host. This data was erroneously used to reset the power state of
that non-local VM. This commit fixes this.

If xenopsd reports the VM as running, it must indeed be resident on the
local host - so we reset the state of the VM to what xenopsd is
reporting.

If xenopsd reports the VM as suspended or halted, we only reset the
state of the VM if xapi believes the VM is resident on localhost.
Previously it would reset the state regardless - which caused the error.

CP-25141 log when VM's powerstate is reset


jonludlam:
Looks good now!

PR #3302, opened 2017-10-17
CA-269137: get_nbd_info: get subject from TLS cert

thomassa:
Use the x509 library to parse the certificate and read its
DNS subject names, if any.

(This replaces the stopgap of using Host.hostname for the
 value of the "subject" field of the vdi_nbd_server_info objects.)

### DIFF
--- a/ocaml/xapi/xapi_vdi.ml
+++ b/ocaml/xapi/xapi_vdi.ml
@@ -1072,12 +1072,16 @@ let _get_nbd_info ~__context ~self ~get_server_certificate =
     if ips = [] then [] else
     let cert = get_server_certificate ~host in
     let port = 10809L in
-    let subject = match Certificates.hostnames_of_pem_cert cert with
+    let subject = try match Certificates.hostnames_of_pem_cert cert with
       | [] -> (
           error "Found no subject DNS names in this hosts's certificate. Returning empty string as subject.";
           ""
         )
       | name :: _ -> name
+    with e -> (
+      error "get_nbd_info: failed to read subject from TLS certificate! Falling back to Host.hostname. Exn was %s" (ExnHelper.string_of_exn e);
+      Db.Host.get_hostname ~__context ~self:host
+    )
     in
     let template = API.{
       vdi_nbd_server_info_exportname = exportname;
--- a/ocaml/xapi/xapi_vdi.ml
+++ b/ocaml/xapi/xapi_vdi.ml
@@ -1072,8 +1072,13 @@ let _get_nbd_info ~__context ~self ~get_server_certificate =
     if ips = [] then [] else
     let cert = get_server_certificate ~host in
     let port = 10809L in
-    (* Stopgap measure: use hostname instead of reading a subject out of the cert. *)
-    let subject = Db.Host.get_hostname ~__context ~self:host in
+    let subject = match Certificates.hostnames_of_pem_cert cert with
+      | [] -> (
+          error "Found no subject DNS names in this hosts's certificate. Returning empty string as subject.";
+          ""
+        )
+      | name :: _ -> name
+    in
     let template = API.{
       vdi_nbd_server_info_exportname = exportname;
       vdi_nbd_server_info_address = "";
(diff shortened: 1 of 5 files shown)


thomassa:
This pull-request depends on adding some more libraries to xs-opam (x509 and its dependencies)
so don't merge yet.

xapi.opam
```diff
@@ -10,6 +10,7 @@ build-test: [[ "jbuilder" "runtest" "-p" name "-j" jobs ]]
 
 depends: [
   "jbuilder" {build & >= "1.0+beta11"}
+  "x509"
```
gaborigloi: We should move it to the `x` section, as this is more or less alphabetically ordered :smile:

ocaml/xapi/xapi_vdi.ml
```diff
@@ -1072,8 +1072,13 @@ let _get_nbd_info ~__context ~self ~get_server_certificate =
     if ips = [] then [] else
     let cert = get_server_certificate ~host in
     let port = 10809L in
-    (* Stopgap measure: use hostname instead of reading a subject out of the cert. *)
-    let subject = Db.Host.get_hostname ~__context ~self:host in
+    let subject = match Certificates.hostnames_of_pem_cert cert with
```
gaborigloi: I reckon we should fall back to the hostname in case the library fails here with an exception.

gaborigloi: As the [documentation of the library](http://docs.mirage.io/x509/X509/index.html) states:
>Missing is the handling of certificate revocation lists, online certificate status protocol, some X.509v3 extensions (such as policy and name constraints). The only supported key type is RSA.

I guess in these cases it will just throw an exception - so I guess if we have the above fallback, then we don't have to worry about this too much?

gaborigloi:
@lindig If I'm not mistaken, we just need to return a subject for which the certificate is valid, which the client can pass to the TLS library it uses to verify that the server's certificate matches that hostname/subject.
See http://xapi-project.github.io/xen-api/classes/vdi_nbd_server_info.html for the documentation of what we return here.

thomassa:
This pull-request depends on adding some more libraries to xs-opam (x509 and its dependencies)
so don't merge yet.

ocaml/xapi/xapi_vdi.ml
gaborigloi: Also, in case one host has an incorrect certificate, `get_nbd_info` will fail completely, and not return anything - not sure what the right behaviour is here - maybe that's fine.

thomassa:
@lindig, Gabor's explanation is right: the specification for this function is that the "subject" field of each vdi_nbd_server_info value must contain a single subject which is in the certificate. It doesn't matter which one.

As for catching and handling an exception... yes, I think that's worth doing; we should raise an alert when it happens, and fall back to hostname.

xapi.opam
thomassa: Agreed. It's here at the moment only because I started out by trying to use asn1-combinators directly, then switched to x509 without moving the entry.

thomassa:
I've just added a couple of extra commits, but I think it might be worth also defining a new alert in `Xapi_message`, and raising one when we catch an exception from the x509 library on attempting to parse the subject out of the certificate string.

Meanwhile, I've made xapi-project/xs-opam#158 to add the x509 library (and its dependencies) to xs-opam: once that one has been merged, this one should pass its test-build.

gaborigloi:
Sadly this PR will only build when we release xs-opam after that PR is merged, and update its spec file to point to the new version.
We do have a Travis build variant that uses OPAM to build xapi, the one which does the coverage analysis, but sadly that has been broken since the move to jbuilder.
So another way of building this PR would be to fix that.

gaborigloi:
Sadly this PR will only build when we release xs-opam after that PR is merged, and update its spec file to point to the new version.
We do have a Travis build variant that uses OPAM to build xapi, the one which does the coverage analysis, but sadly that has been broken since the move to jbuilder.
So another way of building this PR would be to fix that.

gaborigloi:
Yaaay, Travis finally passed :) - merging.

PR #3303, opened 2017-10-18
CA-269642: Revert snapshot should clean up VUSBs

taoyongd:
1. Need to clean up VUSBs when reverting snapshot.

### DIFF
--- a/ocaml/xapi/xapi_vm_snapshot.ml
+++ b/ocaml/xapi/xapi_vm_snapshot.ml
@@ -286,7 +286,7 @@ let safe_destroy_vusb ~__context ~rpc ~session_id vusb =
 
 (* Copy the VBDs and VIFs from a source VM to a dest VM and then delete the old disks. *)
 (* This operation destroys the data of the dest VM.                                    *)
-let update_vifs_vbds_and_vgpus ~__context ~snapshot ~vm =
+let update_vifs_vbds_vgpus_and_vusbs ~__context ~snapshot ~vm =
   let snap_VBDs = Db.VM.get_VBDs ~__context ~self:snapshot in
   let snap_VBDs_disk, snap_VBDs_CD =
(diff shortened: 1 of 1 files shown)


ocaml/xapi/message_forwarding.ml
```diff
@@ -1541,6 +1541,11 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
         then vm
         else Xapi_vm_snapshot.create_vm_from_snapshot ~__context ~snapshot in
 
+      (* As when vm snapshot is not allowed if vm has VUSBs, so revert should be disallowed *)
```
lindig: I wonder whether this could fail earlier to avoid creating the VM first only to fail now. But it is probably not worth the complexity. This looks straight forward to me.

taoyongd: yes, I think fail here is reasonable.

jonludlam: Not convinced (yet...!)

taoyongd: oh, how do you think about this? @jonludlam

jonludlam: OK, what is the difficulty with putting this test before the create_vm_from_snapshot?

taoyongd: If vm is not valid , how can we get the VUSBs?

taoyongd:
yes @lindig . I also notice that, but I think vusb is different from others, we will not set up new vusbs. So I think we can just keep as it. Does it make sense?

mseri:
I agree with @lindig. Even if we are not setting up new vusbs, the function is operating also on vusbs now

taoyongd:
Ok, updated.

PR #3304, opened 2017-10-18
CA-269596: Restrict vm checkpoint operation

taoyongd:


taoyongd:
As we restrict user to do snapshot operation when vm has VUSBs, but we can do snapshot (with disks and memory) in icenter, which actually is a vm checkpoint operation. So we need also restrict this.

jonludlam:
I think this was just an oversight/bug in a previous commit. I think there's enough context to know why this is like this.

PR #3305, opened 2017-10-18
CA-269586: Add device-model check when vm start

taoyongd:


taoyongd:
Close this PR. Branch name is not correct,should be CA-269586.

PR #3306, opened 2017-10-19
CA-269586: Add device-model check when vm start

taoyongd:

### DIFF
--- a/ocaml/xapi/vm_platform.ml
+++ b/ocaml/xapi/vm_platform.ml
@@ -100,6 +100,14 @@ let is_true ~key ~platformdata ~default =
   with Not_found ->
     default
 
+let is_valid_device_model ~key ~platformdata =
+  try
+    match List.assoc key platformdata with
+    | "qemu-upstream-compat" -> true
+    | _ -> false
+  with Not_found ->
+    false
+
 let sanity_check ~platformdata ~vcpu_max ~vcpu_at_startup ~hvm ~filter_out_unknowns =
   (* Filter out unknown flags, if applicable *)
   let platformdata =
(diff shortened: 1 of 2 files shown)


ocaml/xapi/vm_platform.ml
```diff
@@ -100,6 +100,14 @@ let is_true ~key ~platformdata ~default =
   with Not_found ->
     default
 
+let is_valid_device_model ~key ~platformdata =
+  try
+    match List.assoc key platformdata with
+    | "qemu-upstream-compat" -> true
+    | v -> false
```
lindig: I'd use the wildcard `_` as `v` is not used.

ocaml/xapi/vm_platform.ml
```diff
@@ -166,3 +174,8 @@ let check_restricted_flags ~__context platform =
 
   if is_true nested_virt platform false
   then Pool_features.assert_enabled ~__context ~f:Features.Nested_virt
+
+let check_restricted_device_model ~__context platform =
+  if not (is_valid_device_model device_model platform) then
+    raise (Api_errors.Server_error(Api_errors.invalid_value,
+            [Printf.sprintf "platform:%s when vm has VUSBs" device_model ; try List.assoc device_model platform with _ -> ""]))
```
lindig: Instead of returning `""` it would be better to return `(undefined)`.

The code can be slightly simplified to avoid long lines:

```
let check_restricted_device_model ~__context platform =
  if not (is_valid_device_model device_model platform) then
    raise Api_errors.(Server_error.invalid_value,
                      [ Printf.sprintf "platform:%s when vm has VUSBs" device_model 
                      ; try List.assoc device_model platform with Not_found -> "(undefined)"
                      ])
```

taoyongd: good suggestion of "undefined". Thanks.

ocaml/xapi/vm_platform.ml
taoyongd: Ok

ocaml/xapi/vm_platform.ml
taoyongd: updated

ocaml/xapi/vm_platform.ml
taoyongd: updated.

taoyongd:
Can now this PR ok for merge

jonludlam:
Yes please.

taoyongd:
Done @jonludlam

PR #3307, opened 2017-10-19
CA-269728: Unplug the USB from host, not detach the vusb from VM

taoyongd:
1.When unplug usb from host, we need to unplug vusb from vm if it is attached to vm.

### DIFF
--- a/ocaml/xapi/xapi_pusb.ml
+++ b/ocaml/xapi/xapi_pusb.ml
@@ -47,11 +47,10 @@ let scan_start ~__context usbs =
 
   List.filter (fun (rf, rc) -> USBSet.mem (extract_known_usb_info rc) (USBSet.diff known_usb local_usb)) (Db.PUSB.get_all_records ~__context)
   |> List.iter (fun (self, _) ->
-                  let usb_group = Db.PUSB.get_USB_group ~__context ~self in
-                  Db.PUSB.destroy ~__context ~self;
-                  let vusbs = Db.USB_group.get_VUSBs ~__context ~self:usb_group in
-                  List.iter (fun vusb -> Db.VUSB.destroy ~__context ~self:vusb) vusbs;
-                  Db.USB_group.destroy ~__context ~self:usb_group)
+                  try
+                    Xapi_pusb_helpers.destroy_pusb ~__context self;
+                  with _ -> error "failure while removing PUSB: %s" (Ref.string_of self);
+                )
 
 let cond = Condition.create ()
 let mutex = Mutex.create ()
--- a/ocaml/xapi/xapi_pusb.ml
+++ b/ocaml/xapi/xapi_pusb.ml
@@ -49,7 +49,7 @@ let scan_start ~__context usbs =
   |> List.iter (fun (self, _) ->
                   try
                     Xapi_pusb_helpers.destroy_pusb ~__context self;
-                  with _ -> error "failure while removing PUSB: %s" (Ref.string_of self);
+                  with e -> error "Caught exception while removing PUSB %s: %s" (Ref.string_of self) (Printexc.to_string e);
                 )
 
 let cond = Condition.create ()
--- a/ocaml/xapi/xapi_pusb_helpers.ml
+++ b/ocaml/xapi/xapi_pusb_helpers.ml
@@ -126,8 +126,8 @@ let destroy_pusb ~__context pusb =
     let currently_attached = Db.VUSB.get_currently_attached ~__context ~self:vusb in
     if currently_attached then
       Helpers.call_api_functions ~__context (fun rpc session_id ->
-        Client.Client.VUSB.unplug rpc session_id vusb)
+        Client.Client.VUSB.unplug rpc session_id vusb);
+    Db.VUSB.destroy ~__context ~self:vusb
   ) vusbs;
-  List.iter (fun vusb -> Db.VUSB.destroy ~__context ~self:vusb) vusbs;
   Db.PUSB.destroy ~__context ~self:pusb;
   Db.USB_group.destroy ~__context ~self:usb_group
--- a/ocaml/xapi/xapi_pusb_helpers.ml
+++ b/ocaml/xapi/xapi_pusb_helpers.ml
@@ -118,3 +118,16 @@ let find_or_create ~__context pusb =
   let name_label = "Group of " ^ pusb_rec.Db_actions.pUSB_vendor_id ^ " " ^ pusb_rec.Db_actions.pUSB_product_id ^ " USBs" in
   let group = Xapi_usb_group.create ~__context ~name_label ~name_description:"" ~other_config:[] in
   group
+
+let destroy_pusb ~__context pusb =
+  let usb_group = Db.PUSB.get_USB_group ~__context ~self:pusb in
+  let vusbs = Db.USB_group.get_VUSBs ~__context ~self:usb_group in
+  List.iter (fun vusb ->
+    let currently_attached = Db.VUSB.get_currently_attached ~__context ~self:vusb in
+    if currently_attached then
+      Helpers.call_api_functions ~__context (fun rpc session_id ->
+        Client.Client.VUSB.unplug rpc session_id vusb)
+  ) vusbs;
+  List.iter (fun vusb -> Db.VUSB.destroy ~__context ~self:vusb) vusbs;
+  Db.PUSB.destroy ~__context ~self:pusb;
+  Db.USB_group.destroy ~__context ~self:usb_group


taoyongd:
After looking at CP-25325. Which will make the design much simpler.

taoyongd:
This PR has been updated which will not affected by CP-25325

ocaml/xapi/xapi_pusb.ml
```diff
@@ -48,10 +48,17 @@ let scan_start ~__context usbs =
   List.filter (fun (rf, rc) -> USBSet.mem (extract_known_usb_info rc) (USBSet.diff known_usb local_usb)) (Db.PUSB.get_all_records ~__context)
   |> List.iter (fun (self, _) ->
                   let usb_group = Db.PUSB.get_USB_group ~__context ~self in
-                  Db.PUSB.destroy ~__context ~self;
                   let vusbs = Db.USB_group.get_VUSBs ~__context ~self:usb_group in
-                  List.iter (fun vusb -> Db.VUSB.destroy ~__context ~self:vusb) vusbs;
-                  Db.USB_group.destroy ~__context ~self:usb_group)
+                  List.iter (fun vusb ->
+                    let currently_attached = Db.VUSB.get_currently_attached ~__context ~self:vusb in
+                    if currently_attached then begin
```
jonludlam: I'd like to catch errors here and continue - so if we fail to unplug the VUSB, we should not destroy the PUSB and USB_group, but we should then continue to process the rest of the scan result.

taoyongd: Ok.  make sense

taoyongd: updated. please check if the error is ok. Thanks

taoyongd:
t

taoyongd:
changes updated.

ocaml/xapi/xapi_pusb.ml
```diff
@@ -48,10 +48,21 @@ let scan_start ~__context usbs =
...
                   let usb_group = Db.PUSB.get_USB_group ~__context ~self in
-                  Db.PUSB.destroy ~__context ~self;
                   let vusbs = Db.USB_group.get_VUSBs ~__context ~self:usb_group in
+                  List.iter (fun vusb ->
+                    let currently_attached = Db.VUSB.get_currently_attached ~__context ~self:vusb in
+                    if currently_attached then
+                      try
+                        Helpers.call_api_functions ~__context (fun rpc session_id ->
+                          Client.Client.VUSB.unplug rpc session_id vusb)
+                      with _ ->
```
mseri: Should we also add the specific error in the error message?

jonludlam: Hold on, this doesn't do what I suggested - I just wanted to log there had been a problem unplugging the VUSB and then move on to the next PUSB to remove.

taoyongd: totally right.

taoyongd: Will error be ok or just give debug log ?

taoyongd:
updated. @jonludlam  Thansk a lot.

lindig:
This is a difficult piece of logic. Here is just an idea that factors out the removal of a PUSB. 

```
(* new function *)
let destroy_pusb ~__context pusb =
  let group = Db.PUSB.get_USB_group ~__context ~self:pusb in
  Db.USB_group.get_VUSBs ~__context ~self:pusb
  |> List.iter (fun self -> 
      if Db.VUSB.get_currently_attached ~__context ~self then 
            Client.Client.VUSB.unplug rpc session_id vusb))
  |> List.iter (fun self -> Db.VUSB.destroy ~__context ~self)
  ;
  ; Db.PUSB.destroy ~__context ~self:pusb
  ; Db.USB_group.destroy ~__context ~self:group

 (* part of existing code *)
  Db.PUSB.get_all_records ~__context
  |> List.filter (fun (rf, rc) -> 
      USBSet.mem (extract_known_usb_info rc) 
        (USBSet.diff known_usb local_usb))
  |> fun pusb -> try destroy ~__context pusb with 
  | e -> error "removing stale PUSB %s failed (%s)" 
    (Ref.string_of pusb) (__LOC__)
```

If something goes wrong, we capture the exception at the outer level. This would not destroy the pusb and pusb group if a vusb cannot be unplugged or destroyed.

taoyongd:
good  suggestion @lindig

huizh:
Hi @lindig @robhoes, please could you merge this PR if no more comment?

Thanks.

mseri:
It would have been better to have the history of commits to be able to track changes across the various stages of review. @lindig and @jonludlam are not here and I have no way to see what changed between the version before their comments and the current version. This makes it much harder to figure out if the final version is what they wanted despite the code being short

ocaml/xapi/xapi_pusb.ml
```diff
@@ -47,11 +47,10 @@ let scan_start ~__context usbs =
...
   List.filter (fun (rf, rc) -> USBSet.mem (extract_known_usb_info rc) (USBSet.diff known_usb local_usb)) (Db.PUSB.get_all_records ~__context)
   |> List.iter (fun (self, _) ->
-                  let usb_group = Db.PUSB.get_USB_group ~__context ~self in
-                  Db.PUSB.destroy ~__context ~self;
-                  let vusbs = Db.USB_group.get_VUSBs ~__context ~self:usb_group in
-                  List.iter (fun vusb -> Db.VUSB.destroy ~__context ~self:vusb) vusbs;
-                  Db.USB_group.destroy ~__context ~self:usb_group)
+                  try
+                    Xapi_pusb_helpers.destroy_pusb ~__context self;
+                  with _ -> error "failure while removing PUSB: %s" (Ref.string_of self);
```
mseri: Please can you catch the error here and add it to the log line:
```ocaml
with err -> error "Caught exception whilt removing PUSB %s: (Ref.string_of self) (Printexc.to_String err);
```

taoyongd:
Thanks @mseri ,  you are right. Sorry for not leaving the commit history. For this PR, I confirmed with @jonludlam  about the change on Saturday and now this is ok for him(May he now cannot approve this as on vocation.)

ocaml/xapi/xapi_pusb_helpers.ml
```diff
@@ -118,3 +118,16 @@ let find_or_create ~__context pusb =
...
+  let vusbs = Db.USB_group.get_VUSBs ~__context ~self:usb_group in
+  List.iter (fun vusb ->
+    let currently_attached = Db.VUSB.get_currently_attached ~__context ~self:vusb in
+    if currently_attached then
+      Helpers.call_api_functions ~__context (fun rpc session_id ->
+        Client.Client.VUSB.unplug rpc session_id vusb)
+  ) vusbs;
+  List.iter (fun vusb -> Db.VUSB.destroy ~__context ~self:vusb) vusbs;
+  Db.PUSB.destroy ~__context ~self:pusb;
+  Db.USB_group.destroy ~__context ~self:usb_group
```
mseri: I assume the old version was catching the errors thus running the `PUSB.destroy` and `USB_group.destroy` functions in any case, while the current will do it only if the VUSB unplug&destroy are successful. Is that the change that happened? In this case, could you add a comment explaining that this function has to fail when an unplug fails, to prevent an incorrect PUSB and USB_group destruction?

Is there a reason we iterate twice on the vusbs instead of using a unique block?
```ocaml
List.iter (fun vusb ->
    let currently_attached = Db.VUSB.get_currently_attached ~__context ~self:vusb in
    if currently_attached then
        Client.Client.VUSB.unplug rpc session_id vusb);

    Db.VUSB.destroy ~__context ~self:vusb
) vusbs;

Comment (taoyongd) [reply ocaml/xapi/xapi_pusb_helpers.ml]: yes @mseri  Only when unplug vusb successfully , then we will destroy PUSB and USB_group. If we catch a exception when unplug vusb. We should skip destroy PUSB and USB_group and continue other PUSB scan

File: ocaml/xapi/xapi_pusb.ml (hunk 3 above)
Comment (taoyongd) [reply ocaml/xapi/xapi_pusb.ml]: Ok, Thanks @mseri

File: ocaml/xapi/xapi_pusb_helpers.ml (hunk 4 above)
Comment (robhoes) [reply ocaml/xapi/xapi_pusb_helpers.ml]: As @mseri says, let's merge the `List.iter`s.

Comment (taoyongd) [reply ocaml/xapi/xapi_pusb_helpers.ml]: @robhoes , just updated and add some commit comments in the second commit.

File: ocaml/xapi/xapi_pusb.ml [hunk 5]
```diff
@@ -49,7 +49,7 @@ let scan_start ~__context usbs =
   |> List.iter (fun (self, _) ->
                   try
                     Xapi_pusb_helpers.destroy_pusb ~__context self;
-                  with _ -> error "failure while removing PUSB: %s" (Ref.string_of self);
+                  with e -> error "Caught exception whilt removing PUSB %s:, %s:" (Ref.string_of self) (Printexc.to_string e);
```
mseri: Please correct the string:
```
Caught exception while removing PUSB %s: %s
```

taoyongd: Ok, I will just add it to the second commit.

taoyongd: done

taoyongd:
Ok Thanks @mseri

ocaml/xapi/xapi_pusb.ml
```diff
@@ -47,11 +47,10 @@ let scan_start ~__context usbs =
...
   List.filter (fun (rf, rc) -> USBSet.mem (extract_known_usb_info rc) (USBSet.diff known_usb local_usb)) (Db.PUSB.get_all_records ~__context)
   |> List.iter (fun (self, _) ->
-                  let usb_group = Db.PUSB.get_USB_group ~__context ~self in
-                  Db.PUSB.destroy ~__context ~self;
-                  let vusbs = Db.USB_group.get_VUSBs ~__context ~self:usb_group in
-                  List.iter (fun vusb -> Db.VUSB.destroy ~__context ~self:vusb) vusbs;
-                  Db.USB_group.destroy ~__context ~self:usb_group)
+                  try
+                    Xapi_pusb_helpers.destroy_pusb ~__context self;
+                  with e -> error "Caught exception whilt removing PUSB %s: %s" (Ref.string_of self) (Printexc.to_string e);
```
mseri: One last fix: `while` not `whilt`

taoyongd:
@mseri , build passed. Thanks

PR #3308, opened 2017-10-19
CA-269591: Clean vusb attached field when vm shutdown

taoyongd:
1.This solution is to slove the issue that vusb doesn't exist in xenopsd when reboot vm.
Another PR https://github.com/xapi-project/xenopsd/pull/406 in xenopsd will remove unplug vusb when vm shutdown. So xapi will check vusb attached when vm shutdown.

### DIFF
--- a/ocaml/xapi/xapi_xenops.ml
+++ b/ocaml/xapi/xapi_xenops.ml
@@ -2729,7 +2729,15 @@ let shutdown ~__context ~self timeout =
               raise Api_errors.(Server_error(internal_error, [
                   Printf.sprintf "shutdown: The VBD %s is still attached to VM %s"
                     (Ref.string_of vbd) (Ref.string_of self)]))
-         ) (Db.VM.get_VBDs ~__context ~self)
+         ) (Db.VM.get_VBDs ~__context ~self);
+       List.iter
+         (fun vusb ->
+            let attached = Db.VUSB.get_attached ~__context ~self:vusb in
+            if attached <> Ref.null then begin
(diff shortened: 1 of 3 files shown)


taoyongd:
To fix CA-269591, @jonludlam do you think if there's a better solution?

jonludlam:
Yes, better to do this in force_state_reset:
https://github.com/xapi-project/xen-api/blob/usb-passthrough/ocaml/xapi/xapi_vm_lifecycle.ml#L590

taoyongd:
Ok, better location to do this. Thanks @jonludlam

taoyongd:
@jonludlam  updated and test which is ok to clean.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -2597,7 +2597,11 @@ let maybe_cleanup_vm ~__context ~self =
   end
 
 let clear_vusb_attched ~__context ~self =
```
robhoes: Do we actually need this function still?

taoyongd: when  vm start failed. we need to clean both attached field of vusb and pusb. But here may be not the better location. We have a ticket to track where it should be. @robhoes  how do you think?

taoyongd: Ok, so you mean when vm start failed. the field will be reset , right? make sense , seems don't need this

taoyongd: no need this funciton.  Removed.

taoyongd:
close this PR as this will be included in https://github.com/xapi-project/xen-api/pull/3312

PR #3311, opened 2017-10-20
CA-269657: Detach VUSB before setting VM a template

makunterry:
Signed-off-by: Kun Ma <kun.ma@citrix.com>


jonludlam:
Looks good!

PR #3312, opened 2017-10-20
CP-25325: Rationalise handling of attached field in xapi

taoyongd:


ocaml/xapi/xapi_vm_lifecycle.ml
```diff
@@ -629,6 +629,11 @@ let force_state_reset_keep_current_operations ~__context ~self ~value:state =
            Db.PCI.set_scheduled_to_be_attached_to ~__context ~self:pci ~value:Ref.null
       )
       (Db.PCI.get_all ~__context);
+    List.iter
+      (fun vusb ->
+         Db.VUSB.set_currently_attached ~__context ~self:vusb~value:false;
```
taoyongd: add a space

ocaml/xapi/xapi_vusb.ml
```diff
@@ -25,14 +25,15 @@ let create ~__context ~vM ~uSB_group ~other_config =
...
   let attached_vusbs = Db.VM.get_VUSBs ~__context ~self:vM in
-  (* At most 6 VUSBS can be attached to one vm *)
-  if List.length attached_vusbs > 5 then
-    raise (Api_errors.Server_error (Api_errors.too_many_vusbs, ["6"]));
   let vusbs = Db.USB_group.get_VUSBs ~__context ~self:uSB_group in
   (* Currently USB_group only have one PUSB. So when vusb is created with a USB_group,
       another vusb can not create with the same USB_group. *)
   if vusbs <> [] then
     raise (Api_errors.Server_error(Api_errors.usb_group_conflict, [Ref.string_of uSB_group]));
+  (* At most 6 VUSBS can be attached to one vm *)
```
taoyongd: revert back

ocaml/xapi/xapi_pusb.ml
```diff
@@ -122,16 +122,17 @@ let set_passthrough_enabled ~__context ~self ~value =
...
+         (* If the USB is passthroughed to vm, need to unplug it firstly*)
+         if vusbs <> [] then begin
+           let currently_attached = Db.VUSB.get_currently_attached ~__context ~self:(List.hd vusbs) in
+           if currently_attached then
+             let vm = Db.VUSB.get_VM ~__context ~self:(List.hd vusbs) in
+             raise (Api_errors.Server_error(Api_errors.usb_already_attached, [Ref.string_of self; Ref.string_of vm]))
+           (* If vusb has been created, need to destroy it. *)
+         end;
+         Db.VUSB.destroy ~__context ~self:(List.hd vusbs);
          debug "set passthrough_enabled %b." value;
```
jonludlam: This `destroy` should be either in the block above, or otherwise it should be the `List.iter` it was before. I think I prefer the `List.iter`.

ocaml/xapi/xapi_vm_lifecycle.ml
jonludlam: Have you not even compiled this?

ocaml/xapi/xapi_pusb.ml
taoyongd: right.

ocaml/xapi/xapi_vm_lifecycle.ml
taoyongd: I compiled this. It works. I just go through again the code and find this.

taoyongd: every push I will compile the code.

ocaml/xapi/xapi_pusb.ml
taoyongd: updated.

ocaml/xapi/xapi_vm_lifecycle.ml
taoyongd: done

jonludlam:
There are several instances of `List.hd` in this PR without exception handlers - I think we need to catch and log these. I'll comment on each one to say what we should do.

ocaml/xapi/xapi_pusb.ml
```diff
@@ -122,16 +122,17 @@ let set_passthrough_enabled ~__context ~self ~value =
...
-           let vm = Db.VUSB.get_VM ~__context ~self:attached in
-           raise (Api_errors.Server_error(Api_errors.usb_already_attached, [Ref.string_of self; Ref.string_of vm]))
-         end;
          let usb_group = Db.PUSB.get_USB_group ~__context ~self in
          let vusbs = Db.USB_group.get_VUSBs ~__context ~self:usb_group in
-         (* If vusb has been created, need to destroy it. *)
-         List.iter (fun vusb -> Xapi_vusb.destroy ~__context ~self:vusb) vusbs;
+         (* If the USB is passthroughed to vm, need to unplug it firstly*)
+         if vusbs <> [] then begin
+           let currently_attached = Db.VUSB.get_currently_attached ~__context ~self:(List.hd vusbs) in
```
jonludlam: This List.hd is OK.

ocaml/xapi/xapi_pusb.ml
```diff
@@ -122,16 +122,17 @@ let set_passthrough_enabled ~__context ~self ~value =
...
+           if currently_attached then
+             let vm = Db.VUSB.get_VM ~__context ~self:(List.hd vusbs) in
```
jonludlam: And this one. Though we could avoid both by changing the
```ocaml
if vusbs <> [] then begin
```
to
```ocaml
match vusbs with
| [] -> ()
| [vusb] -> begin
```

ocaml/xapi/xapi_xenops.ml
```diff
@@ -695,9 +695,8 @@ module MD = struct
   let vusbs_of_vm ~__context (vmref, vm) =
     vm.API.vM_VUSBs
     |> List.map (fun self -> Db.VUSB.get_record ~__context ~self)
-    (* get attached PUSBs. *)
-    |> List.filter (fun vusb -> vusb.API.vUSB_attached <> Ref.null)
-    |> List.map (fun vusb -> Db.PUSB.get_record ~__context ~self:vusb.API.vUSB_attached)
+    |> List.map (fun vusb -> Db.USB_group.get_record ~__context ~self:vusb.API.vUSB_USB_group)
+    |> List.map (fun usb_group -> Db.PUSB.get_record ~__context ~self:(List.hd usb_group.API.uSB_group_PUSBs))
```
jonludlam: This one needs to be caught, and an Internal_error raised saying that there is no PUSB associated with the USB_group

ocaml/xapi/xapi_pusb.ml
```diff
@@ -122,16 +122,17 @@ let set_passthrough_enabled ~__context ~self ~value =
...
-         if attached <> Ref.null then begin
-           let vm = Db.VUSB.get_VM ~__context ~self:attached in
-           raise (Api_errors.Server_error(Api_errors.usb_already_attached, [Ref.string_of self; Ref.string_of vm]))
-         end;
          let usb_group = Db.PUSB.get_USB_group ~__context ~self in
          let vusbs = Db.USB_group.get_VUSBs ~__context ~self:usb_group in
-         (* If vusb has been created, need to destroy it. *)
-         List.iter (fun vusb -> Xapi_vusb.destroy ~__context ~self:vusb) vusbs;
+         (* If the USB is passthroughed to vm, need to unplug it firstly*)
+         if vusbs <> [] then begin
```
lindig: This code does not look safe to me because only the first element in `vusbs` is handled. 

* If we expect the list never to have more than one element, cases when there are more need tp be at least reported
* If the list can have multiple elements, why are they not treated the same? I would suggest a `List.iter` over the list.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -1956,18 +1955,19 @@ let update_vusb ~__context (id: (string * string)) =
...
-            |> List.map (fun vusb -> Db.VUSB.get_attached ~__context ~self:vusb)
-            |> List.filter (fun attached -> Db.is_valid_ref __context attached)
+            |> List.map (fun self -> Db.VUSB.get_USB_group ~__context ~self)
+            |> List.map (fun usb_group -> List.hd (Db.USB_group.get_PUSBs ~__context ~self:usb_group))
             |> List.map (fun self -> self, Db.PUSB.get_record ~__context ~self)
             |> List.find (fun (_, pusbr) -> "vusb" ^ pusbr.API.pUSB_path= (snd id))
           in
-          let vusb = Db.PUSB.get_attached ~__context ~self:pusb in
+          let usb_group = Db.PUSB.get_USB_group ~__context ~self:pusb in
+          let vusb = List.hd (Db.USB_group.get_VUSBs ~__context ~self:usb_group) in
```
jonludlam: This convoluted logic is here because we haven't got a better way of identifying which VUSB the update is associated with. Ideally we'd have something like the `device_number` that we have for VIFs - ie, a thing associated with the VUSB rather than the PUSB.

Since this is in the events_from_xenopsd code, let's catch and log exceptions here, but carry on.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -1956,18 +1955,19 @@ let update_vusb ~__context (id: (string * string)) =
         else begin
           let pusb, pusb_r =
             Db.VM.get_VUSBs ~__context ~self:vm
-            |> List.map (fun vusb -> Db.VUSB.get_attached ~__context ~self:vusb)
-            |> List.filter (fun attached -> Db.is_valid_ref __context attached)
+            |> List.map (fun self -> Db.VUSB.get_USB_group ~__context ~self)
+            |> List.map (fun usb_group -> List.hd (Db.USB_group.get_PUSBs ~__context ~self:usb_group))
```
mseri: We need to catch also this and raise an error saying that there is no PUSB, or we can avoid the `List.hd` and process empty strings in this case?

ocaml/xapi/xapi_xenops.ml
mseri: Can this be an empty list?

ocaml/xapi/xapi_xenops.ml
```diff
@@ -3130,9 +3124,10 @@ let task_cancel ~__context ~self =
 
 let md_of_vusb ~__context ~self =
   let vm = Db.VUSB.get_VM ~__context ~self in
-  let vusb = Db.VUSB.get_record ~__context ~self in
-  let pusb =Db.PUSB.get_record ~__context ~self:vusb.API.vUSB_attached in
-  MD.of_vusb ~__context ~vm:(Db.VM.get_record ~__context ~self:vm) ~pusb:pusb
+  let usb_group = Db.VUSB.get_USB_group ~__context ~self in
+  let pusb = List.hd (Db.USB_group.get_PUSBs ~__context ~self:usb_group) in
```
jonludlam: This one should be caught and an internal_error should be raised (as this is used in VUSB.unplug)

ocaml/xapi/xapi_pusb.ml
mseri: You can also protect the "exactly one" case adding `| _ :: _ :: _ -> error "too many vusb on the pusb"` with your second suggestion

ocaml/xapi/xapi_xenops.ml
mseri: I'd rater use 
```ocaml
match Db.USB_group.get_PUSBs ~__context ~self:usb_group with
[] -> error
_ :: _ :: _ -> error
[vusb] -> ...
```

ocaml/xapi/xapi_xenops.ml
taoyongd: Ok

ocaml/xapi/xapi_xenops.ml
taoyongd: yes

ocaml/xapi/xapi_pusb.ml
taoyongd: updated.

ocaml/xapi/xapi_xenops.ml
taoyongd: updated.

ocaml/xapi/xapi_xenops.ml
taoyongd: update, will raise internal error here ok?

taoyongd: updated.

ocaml/xapi/xapi_xenops.ml
taoyongd: updated.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -695,9 +695,18 @@ module MD = struct
...
-    |> List.map (fun vusb -> Db.PUSB.get_record ~__context ~self:vusb.API.vUSB_attached)
+    |> List.map (fun vusb -> Db.USB_group.get_record ~__context ~self:vusb.API.vUSB_USB_group)
+    |> List.map
+      (fun usb_group ->
+        let pusb =
+          try
+            List.hd usb_group.API.uSB_group_PUSBs
+          with _ ->
+            raise Api_errors.(Server_error(internal_error,
+              [Printf.sprintf "there is no PUSB associated with the USB_group"]))
```
lindig: This needs to include the reference of the USB group because otherwise the log is not helpful.

lindig: There is a pattern here to get the first element in a usb group - this should be factored into its own function such that we have all the error handling only once.

taoyongd: good suggestion, but here we need to wirte something to get the reference of USB_group

ocaml/xapi/xapi_pusb.ml
```diff
@@ -122,16 +122,20 @@ let set_passthrough_enabled ~__context ~self ~value =
...
-           raise (Api_errors.Server_error(Api_errors.usb_already_attached, [Ref.string_of self; Ref.string_of vm]))
-         end;
          let usb_group = Db.PUSB.get_USB_group ~__context ~self in
          let vusbs = Db.USB_group.get_VUSBs ~__context ~self:usb_group in
-         (* If vusb has been created, need to destroy it. *)
-         List.iter (fun vusb -> Xapi_vusb.destroy ~__context ~self:vusb) vusbs;
+         (* If the USB is passthroughed to vm, need to unplug it firstly*)
+         match vusbs with
+         | [] -> ()
+         | _ :: _ :: _ -> error "too many vusb on the usb_group"
```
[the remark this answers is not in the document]
mseri: It also should not be `error` but should rais a proper `Server_error` or `Internal_error`. My code snippet was just an example, not to take literally.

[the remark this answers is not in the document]
taoyongd: ok

ocaml/xapi/xapi_xenops.ml
taoyongd: updated.

jonludlam:
Jolly good. In it goes!

ocaml/xapi/xapi_pusb.ml
```diff
@@ -122,16 +122,21 @@ let set_passthrough_enabled ~__context ~self ~value =
...
          let vusbs = Db.USB_group.get_VUSBs ~__context ~self:usb_group in
-         (* If vusb has been created, need to destroy it. *)
-         List.iter (fun vusb -> Xapi_vusb.destroy ~__context ~self:vusb) vusbs;
+         (* If the USB is passthroughed to vm, need to unplug it firstly*)
+         match vusbs with
+         | [] -> ()
+         | _ :: _ :: _ -> raise Api_errors.(Server_error(internal_error,
+                            [Printf.sprintf "too many vusb on the USB_group: %s" (Ref.string_of usb_group)]))
+         | [vusb] -> begin
+           let currently_attached = Db.VUSB.get_currently_attached ~__context ~self:(List.hd vusbs) in
```
mseri: In this block you could have used directly `vusb` without the need of calling `List.hd` each time

ocaml/xapi/helpers.ml
```diff
@@ -1312,3 +1312,19 @@ let retry ~__context ~doc ?(policy = Policy.standard) f =
 
 let retry_with_global_lock ~__context ~doc ?policy f =
   retry ~__context ~doc ?policy (fun () -> with_global_lock f)
+
+let get_first_pusb ~__context usb_group =
+  try
+    List.hd (Db.USB_group.get_PUSBs ~__context ~self:usb_group)
+  with _ ->
```
mseri: I would have preferred the 
```ocaml
match ... with
| [] ->  error ...
| _ : _ : _ -> error ...
| [pusb] -> pusb 
```
approach for these two functions

PR #3313, opened 2017-10-20
CP-25328: update policy file, hubs are always denied

TalonsLee:
Remove the DENY hubs line from policy file, because hubs are always
ignored. Also add comment to make it clear.

### DIFF
--- a/scripts/usb-policy.conf
+++ b/scripts/usb-policy.conf
@@ -2,9 +2,10 @@
 #  and each rule is (ALLOW | DENY) : ( match )*
 #  and each match is (class|subclass|prot|vid|pid|rel) = hex-number
 # Maximum hex value for class/subclass/prot is FF, and for vid/pid/rel is FFFF
+#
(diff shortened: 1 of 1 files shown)


scripts/usb-policy.conf
```diff
@@ -2,9 +2,10 @@
 #  and each rule is (ALLOW | DENY) : ( match )*
 #  and each match is (class|subclass|prot|vid|pid|rel) = hex-number
 # Maximum hex value for class/subclass/prot is FF, and for vid/pid/rel is FFFF
+#
+# Hubs are always denied
```
mseri: Maybe let's make it more explicative:
```
# USB Hubs (calls 09) are always denied, independently of the rules in the policy file

Comment (TalonsLee) [reply scripts/usb-policy.conf]: class?

Comment (mseri) [reply scripts/usb-policy.conf]: Yes, thanks :)

**Comment by mseri** [comment]
Also rename the commit title and the PR title to
```
CP-25328: update policy file, hubs are always denied
```

PR #3314, opened 2017-10-20
Fix mutex in VUSB.create

robhoes:
The mutex is to protect the pre-checks, which limit the number of VUSBs per VM,
and per USB group, in the case that multiple VSUB.create calls come in. The
critical section must start before the `Db.VM.get_VUSBs` call for this to be
thread safe.

Also note that the final couple of lines of the function are part of the second
case in the `ha_restart_priority` match, even though the indentation suggests
otherwise. This is actually (luckily) alright, because the first case raises an
exception.

### DIFF
--- a/ocaml/xapi/xapi_vusb.ml
+++ b/ocaml/xapi/xapi_vusb.ml
@@ -24,27 +24,27 @@ let create ~__context ~vM ~uSB_group ~other_config =
   let vusb = Ref.make () in
   let uuid = Uuid.to_string (Uuid.make_uuid ()) in
   Pool_features.assert_enabled ~__context ~f:Features.USB_passthrough;
-  let attached_vusbs = Db.VM.get_VUSBs ~__context ~self:vM in
-  (* At most 6 VUSBS can be attached to one vm *)
-  if List.length attached_vusbs > 5 then
-    raise (Api_errors.Server_error (Api_errors.too_many_vusbs, ["6"]));
-  let vusbs = Db.USB_group.get_VUSBs ~__context ~self:uSB_group in
-  (* Currently USB_group only have one PUSB. So when vusb is created with a USB_group,
-      another vusb can not create with the same USB_group. *)
-  if vusbs <> [] then
-    raise (Api_errors.Server_error(Api_errors.usb_group_conflict, [Ref.string_of uSB_group]));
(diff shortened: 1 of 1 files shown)


robhoes:
Viewing the diff with `?w=1` is recommended.

ocaml/xapi/xapi_vusb.ml
```diff
@@ -24,27 +24,27 @@ let create ~__context ~vM ~uSB_group ~other_config =
...
-  (* We won't attach VUSB when VM ha_restart_priority is set to 'restart'  *)
-  let ha_restart_priority = Db.VM.get_ha_restart_priority ~__context ~self:vM in
-  match ha_restart_priority with
-    | hp when hp = Constants.ha_restart -> raise (Api_errors.Server_error(Api_errors.operation_not_allowed,
-        [Printf.sprintf "VM %s ha_restart_priority has been set to 'restart', can not create VUSB for it. " (Ref.string_of vM)]))
-    | _ -> ();
   Mutex.execute m (fun () ->
+    let attached_vusbs = Db.VM.get_VUSBs ~__context ~self:vM in
+    (* At most 6 VUSBS can be attached to one vm *)
+    if List.length attached_vusbs > 5 then
```
lindig: I think we should use `>=6` to align it closer with the comment and the documentation.

ocaml/xapi/xapi_vusb.ml
```diff
@@ -24,27 +24,27 @@ let create ~__context ~vM ~uSB_group ~other_config =
...
+      raise (Api_errors.Server_error (Api_errors.too_many_vusbs, ["6"]));
+    let vusbs = Db.USB_group.get_VUSBs ~__context ~self:uSB_group in
+    (* Currently USB_group only have one PUSB. So when vusb is created with a USB_group,
+        another vusb can not create with the same USB_group. *)
+    if vusbs <> [] then
+      raise (Api_errors.Server_error(Api_errors.usb_group_conflict, [Ref.string_of uSB_group]));
+    (* We won't attach VUSB when VM ha_restart_priority is set to 'restart'  *)
+    let ha_restart_priority = Db.VM.get_ha_restart_priority ~__context ~self:vM in
+    match ha_restart_priority with
```
lindig: I would have matched this directly:
```
match Db.VM.get_ha_restart_priority ~__context ~self:vM with
```

PR #3316, opened 2017-10-23
CA-270405: Can't set USB pass-through to disable

taoyongd:
1. No matter vusbs is [] nor not, we still need to
continue executing the set passthrough code.
2. List.hd vusbs is no need when match [vusb]

### DIFF
--- a/ocaml/xapi/xapi_pusb.ml
+++ b/ocaml/xapi/xapi_pusb.ml
@@ -125,17 +125,17 @@ let set_passthrough_enabled ~__context ~self ~value =
          let usb_group = Db.PUSB.get_USB_group ~__context ~self in
          let vusbs = Db.USB_group.get_VUSBs ~__context ~self:usb_group in
          (* If the USB is passthroughed to vm, need to unplug it firstly*)
-         match vusbs with
-         | [] -> ()
-         | _ :: _ :: _ -> raise Api_errors.(Server_error(internal_error,
-                            [Printf.sprintf "too many vusb on the USB_group: %s" (Ref.string_of usb_group)]))
-         | [vusb] -> begin
-           let currently_attached = Db.VUSB.get_currently_attached ~__context ~self:(List.hd vusbs) in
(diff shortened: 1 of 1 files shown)


mseri:
I don't understand how does this solve the issue of being unable to disable passthrough. It is still checking if there are vusbs attached and failing if there are, before trying to destroy them and then setting the enable/disable passthtrough

taoyongd:
@mseri . When vusb is [], we still need to do following code. 
`List.iter (fun vusb -> Db.VUSB.destroy ~__context ~self:vusb) vusbs;		           
 debug "set passthrough_enabled %b." value;
The original one will not go through the code vusb is [].

mseri:
Oh I see, it was the missing semicolon! Great, thanks

robhoes:
Not sure about the missing semicolon. The problem was that the last couple of lines were part of the last case in the match. A `begin`/`end` around the match would have fixed it as well (in fact, there was an `end` already, but the `begin` was in the wrong place).

mseri:
I disagree. If you test it with
```ocaml
let test lst = 
match lst with
| _ :: _ :: _ -> failwith "A"
| [] -> ()
| [a] -> if a then failwith "B";

Printf.printf "here!"
```

You have
```
- test []
;;
- : unit = ()
- test [true]
;;
Exception: Failure "B".
- test [false]
;;
here!- : unit = () 
```

mseri:
Maybe it would have been more coherent to move the `begin` as you suggest. Do you think we should change it or it is fine as it is?

robhoes:
Doesn't this prove that the print line is part of the `[a]` case?

robhoes:
I wouldn't change it anymore.

mseri:
It does indeed. It is clear if you try to move the `[] -> ()` as the final match expression. That would be a syntax error due to the missing semicolons 😄

robhoes:
Well, sure @mseri... :)

PR #3318, opened 2017-10-24
CA-270484: Restrict attach vusb to a template or pv guests

taoyongd:
1. Attaching vusb to a template or PV guests should be restricted
when creating VUSB even though it will not successfully attach to vm
when vm starts.

### DIFF
--- a/ocaml/xapi/xapi_vusb.ml
+++ b/ocaml/xapi/xapi_vusb.ml
@@ -24,6 +24,10 @@ let create ~__context ~vM ~uSB_group ~other_config =
   let vusb = Ref.make () in
   let uuid = Uuid.to_string (Uuid.make_uuid ()) in
   Pool_features.assert_enabled ~__context ~f:Features.USB_passthrough;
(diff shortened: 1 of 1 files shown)


robhoes:
I don't think we need this:

1. Templates are allowed to have VIFs, VBDs and VGPUs. It would be inconsistent if they can't have VUSBs.
2. The HVM check is not very useful, because whether a VM will boot HVM or PV can be changed after creating the VUSB. This should be checked when starting the VM instead (in xapi_xenops.ml), which may already be the case.

robhoes:
Even checking for HVM in xapi_xenopsd.ml may not be needed. I can't see this being done for vGPU, or PCI passthrough, either.

robhoes:
Ah, for vGPU, the check is in `Vgpuops.create_vgpus`.

taoyongd:
yes @robhoes . So we need to check  hvm in xapi_vusb.ml. How do you think so?

robhoes:
That's what I said, yes.

PR #3319, opened 2017-10-24
CP-22019: xapi quicktest checks VDI.update is CBT-compatible

minishrink:
Quicktest currently contains 2 tests for CBT:
- one checks VDI.snapshot and VDI.data_destroy retain CBT-relevant VDI fields
- as above but for VDI.copy and VDI.clone

This PR extends those both to call VDI.update after every CBT-relevant call, ensuring VDI.update preserves CBT-relevant fields
Note: this is a preliminary test, ideally would like to set incorrect database values and check for VDI.update to fix them, looking for a way to do this within a quicktest

### DIFF
--- a/ocaml/xapi/quicktest_cbt.ml
+++ b/ocaml/xapi/quicktest_cbt.ml
@@ -73,8 +73,8 @@ let test_vdi_update ~test ~session_id vDI =
   let vdi_before = list_fields_labels_of vDI in
   VDI.update ~session_id ~rpc:!rpc ~vdi:vDI;
   let vdi_after = list_fields_labels_of vDI in
-  let validate (fd, label) (_fd, _) =
-    test_compare ~test fd _fd ~msg:(Printf.sprintf "VDI update failed: incorrect %s value" label) in
+  let validate (fd_a, label) (fd_b, _) =
+    test_compare ~test fd_a fd_b ~msg:(Printf.sprintf "VDI update failed: incorrect %s value" label) in
   validate (fst vdi_before) (fst vdi_after);
   List.iter2 validate (snd vdi_before) (snd vdi_after)
 
--- a/ocaml/xapi/quicktest_cbt.ml
+++ b/ocaml/xapi/quicktest_cbt.ml
@@ -59,8 +59,8 @@ let make_vdi_from ~session_id ~sR =
 
 (* Helper for calling VDI.update and comparing changes in fields
  * after making an API call on a VDI *)
-let test_vdi_update ~test ~session_id ~vDI =
-  let list_bools_of ~vDI =
+let test_vdi_update ~test ~session_id vDI =
+  let list_bools_of vDI =
     let list_bools =
       [ VDI.get_cbt_enabled , "cbt-enabled"
       ; VDI.get_is_a_snapshot , "is-a-snapshot"
@@ -70,9 +70,9 @@ let test_vdi_update ~test ~session_id ~vDI =
     ( (VDI.get_type ~session_id ~rpc:!rpc ~self:vDI , "type" ), list_bools ) in
 
   (* Call a VDI function like enable_CBT, check fields after,  *)
-  let vdi_before = list_bools_of ~vDI in
+  let vdi_before = list_bools_of vDI in
   VDI.update ~session_id ~rpc:!rpc ~vdi:vDI;
-  let vdi_after = list_bools_of ~vDI in
+  let vdi_after = list_bools_of vDI in
   let validate (fd, label) (_fd, _) =
     test_compare ~test fd _fd ~msg:(Printf.sprintf "VDI update failed: incorrect %s value" label) in
   validate (fst vdi_before) (fst vdi_after);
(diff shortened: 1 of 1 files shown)


ocaml/xapi/quicktest_cbt.ml
```diff
@@ -74,21 +95,27 @@ let vdi_data_destroy_test ~session_id ~vDI =
     VDI.enable_cbt ~session_id ~rpc:!rpc ~self:vDI;
     assert_cbt_status true ~session_id ~test ~vDI
       ~msg:"VDI.enable_cbt failed";
+    test_vdi_update ~session_id ~test ~vDI;
 
     debug test "Snapshotting original VDI with CBT enabled";
     let snapshot = VDI.snapshot ~session_id ~rpc:!rpc ~vdi:vDI ~driver_params:[] in
     assert_cbt_status true ~session_id ~test ~vDI:snapshot
       ~msg:"VDI.snapshot failed, cbt_enabled field didn't carry over";
+    test_vdi_update ~session_id ~test ~vDI;
```
gaborigloi: See my comment below on line 118 - if we want to follow the same pattern as there, we should check both `snapshot` and `vDI` here.

ocaml/xapi/quicktest_cbt.ml
```diff
@@ -167,10 +169,11 @@ let vdi_clone_copy_test ~session_id ~sR ~vDI =
     [ true , vDI , "VDI.copy erroneously reset the original VDI's cbt_enabled to false"
     ; false , into_vdi , "VDI.copy failed to initialise cbt_enabled to false"
     ; false , vdi_copy_fresh , "VDI.copy erroneously reset the copied VDI's cbt_enabled field to true"
+    ; false , vdi_clone , "VDI.copy erroneously reset the cloned VDI's cbt_enabled field to true"
```
gaborigloi: I think this text should say `"VDI.clone failed to set cbt_enabled to false"` because we're doing the same check here as on line 156 - we don't call `VDI.copy` on `vdi_clone`, if I'm not mistaken, so it cannot reset the cloned VDI's fields?

gaborigloi: So I think this means that we can remove the `assert_cbt_status` check for `vdi_clone` at line 156, because we're doing the same check here?

minishrink:
Rebased on top of master branch

gaborigloi:
```
  Testing changed block tracking
                                               
   Testing SR: "Local storage"
                                                 
   Creating VDI. . .                                                            
    Testing VDI.{enable/disable_cbt, data_destroy, snapshot}                    
     Enabling CBT on original VDI                                               
     Snapshotting original VDI with CBT enabled                                 
     Disabling CBT on original VDI                                              
     Snapshotting original VDI with CBT disabled                                
     Destroying snapshot VDI data                                               
     Testing VDI.{enable/disable_cbt, data_destroy, snapshot} failed:[ Failed  ]
      INTERNAL_ERROR: [ Storage_interface.Vdi_does_not_exist("e111dcb[ Failed  ]
     d-a3ed-4a38-bb00-9036db04c457") ]                               [ Failed  ]
                                                                     [ Failed  ]
   Creating VDI. . .                                                            
    Testing VDI.clone and VDI.copy                                              
     Enabling CBT on original VDI                                               
     Cloning VDI                                                                
     Copying VDI into a freshly created VDI in same SR                          
     Copying differences between original VDI and fresh copy to a new           
      VDI                                                                       
                                                                     [ Success ]
   Destroying VDIs created in test. . .                                         
   Successfully destroyed all VDIs created for CBT test
                        
   Finished testing changed block tracking                                      
                                                                     [ Success ]
```

gaborigloi:
This error is from SM, and is expected, because `VDI.update` does not look at the metadata log files.

ocaml/xapi/quicktest_cbt.ml
```diff
@@ -98,6 +123,7 @@ let vdi_data_destroy_test ~session_id ~vDI =
       ~msg:"VDI.data_destroy failed to update VDI.type";
     assert_cbt_status true ~session_id ~test ~vDI:snapshot
       ~msg:"VDI snapshot cbt_enabled field erroneously set to false";
+    test_vdi_update ~session_id ~test snapshot;
```
gaborigloi: Quicktest fails at this step.

gaborigloi: Let's comment this out temporarily?

ocaml/xapi/quicktest_cbt.ml
```diff
@@ -123,7 +123,7 @@ let vdi_data_destroy_test ~session_id ~vDI =
       ~msg:"VDI.data_destroy failed to update VDI.type";
     assert_cbt_status true ~session_id ~test ~vDI:snapshot
       ~msg:"VDI snapshot cbt_enabled field erroneously set to false";
-    test_vdi_update ~session_id ~test snapshot;
+  (*  test_vdi_update ~session_id ~test snapshot; *)
```
gaborigloi: I think we could also mention here that it's blocked on CA-273981, that VDI.update currently does not work on a cbt_metadata VDI. This way we'll know that we can uncomment/remove this test when the issue is fixed/resolved.

ocaml/xapi/quicktest_cbt.ml
```diff
@@ -123,7 +123,9 @@ let vdi_data_destroy_test ~session_id ~vDI =
       ~msg:"VDI.data_destroy failed to update VDI.type";
     assert_cbt_status true ~session_id ~test ~vDI:snapshot
       ~msg:"VDI snapshot cbt_enabled field erroneously set to false";
-    test_vdi_update ~session_id ~test snapshot;
+  (*  test_vdi_update ~session_id ~test snapshot; 
+      temporarily comment this out as it is blocked on CA-273981
+      VDI.update doesn't currently work on cbt-metadata VDIs *)
```
gaborigloi: Actually we can use `SR.scan` until CA-273981 is resolved, so I think we should put that below this comment to still get some testing.

PR #3321, opened 2017-10-25
CP-25372: Prefer non-wildcard certificate subjects

thomassa:
In the VDI.get_nbd_info function, when reading a TLS certificate and
choosing a subject from it to include in the return value, avoid
wildcard subjects if possible: return a concrete hostname (DNS name)
if there is one, and a wildcard only as a last resort.


lindig:
{noformat}
commit 964ff8c1370ca0d6dd11f2937f84eeedc550f00c
Merge: f3d1e38fe e9b039097
Author: Marcello Seri <mseri@users.noreply.github.com>
Date:   Wed Oct 25 16:04:55 2017 +0100

    Merge pull request #3321 from thomassa/prefer-non-wildcard
    
    CP-25372: Prefer non-wildcard certificate subjects
{noformat}

This is on master.

PR #3324, opened 2017-10-26
CP-24132: unit test for usb_scan.py

TalonsLee:
add unit test for usb_scan.py

### DIFF
--- a/scripts/test_usb_scan.py
+++ b/scripts/test_usb_scan.py
@@ -1,0 +1,393 @@
+#!/usr/bin/env python
+#
+# unittest for usb_scan.py
+
+from collections import Mapping, Container, Iterable
+import mock
+from nose.tools import nottest
+import os
+import shutil
+import sys
+import tempfile
+import unittest
+
+sys.modules["xcp"] = mock.Mock()
+sys.modules["xcp.logger"] = mock.Mock()
+sys.modules["pyudev"] = mock.Mock()
+
+log_file = None
+
+
+class MocDeviceAttrs(Mapping):
+    def __init__(self, device):
+        self.d = device.get_attr()
+
+    def __iter__(self):
+        for name in self.d:
+            yield name
+
+    def __len__(self):
+        return len(self.d)
+
+    def __getitem__(self, name):
+        return self.d.get(name)
+
+
+class MocDevice(Mapping):
+
+    def __init__(self, d):
+        self.d = d
+
+    @property
+    def sys_name(self):
+        return self.d.get("name")
+
+    def get_prop(self):
+        return self.d.get("props")
+
+    def get_attr(self):
+        return self.d.get("attrs")
+
+    @property
+    def attributes(self):
+        return MocDeviceAttrs(self)
+
+    def __iter__(self):
+        for name in self.get_prop():
+            yield name
+
+    def __len__(self):
+        return len(self.get_prop())
+
+    def __getitem__(self, name):
+        return self.get_prop().get(name)
+
+
+class MocEnumerator(object):
+
+    def __init__(self, ds):
+        self.ds = ds
+
+    def __iter__(self):
+        for d in self.ds:
+            yield MocDevice(d)
+
+
+class MocContext(object):
+
+    def __init__(self, devices, interfaces):
+        self.devices = devices
+        self.interfaces = interfaces
+
+    def list_devices(self, **kwargs):
+        if "usb" == kwargs.pop("subsystem"):
+            dev_type = kwargs.pop("DEVTYPE")
+            if "usb_device" == dev_type:
+                return MocEnumerator(self.devices)
+            elif "usb_interface" == dev_type:
+                return MocEnumerator(self.interfaces)
+        return MocEnumerator([])
+
+
+def mock_setup(mod, devices, interfaces, path):
+    mod.log.error = test_log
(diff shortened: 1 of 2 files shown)


scripts/test_usb_scan.py
```diff
@@ -0,0 +1,393 @@
...
+def mock_setup(mod, devices, interfaces, path):
+    mod.log.error = test_log
+    mod.log.debug = test_log
+    mod.Policy._PATH = path
+    mod.pyudev.Context = mock.Mock(return_value=MocContext(
+        devices, interfaces))
+
+
+@nottest
+def test_log(m):
```
mseri: I would avoid using a file, instead I'd mock the stdout or the stderr and use print.
An example is
```python
>>> from StringIO import StringIO
>>> def foo():
...     print 'Something'
...
>>> @patch('sys.stdout', new_callable=StringIO)
... def test(mock_stdout):
...     foo()
...     assert mock_stdout.getvalue() == 'Something\n'
...
>>> test()
```
from the mock documentation at http://www.voidspace.org.uk/python/mock/patch.html

However it doesn't seem that you are using the logs content at all. In that case just replace the log funciton with a void function

scripts/test_usb_scan.py
```diff
@@ -0,0 +1,393 @@
...
+        pusbs = [usb_scan.to_pusb(d) for d in devices if
+                 d.is_ready() and policy.check(d)]
+
+        # pass pusbs in json to XAPI
+        self.assertItemsEqual(pusbs, moc_results)
+
+    @nottest
+    def test_usb_exit(self, moc_devices, moc_interfaces, moc_results,
+                        path="./scripts/usb-policy.conf"):
+        with self.assertRaises(SystemExit):
```
mseri: I think here you should also check that the failure did show the correct log error, at least the immutable part of it.
You could use `@raises(SystemExit)` as attribute and `assert` over the initial part of the error message for example

scripts/test_usb_scan.py
```diff
@@ -0,0 +1,393 @@
...
+
+    @nottest
+    def test_usb_common(self, moc_devices, moc_interfaces, moc_results,
+                        path="./scripts/usb-policy.conf"):
+        import usb_scan
+        mock_setup(usb_scan, moc_devices, moc_interfaces, path)
+
+        devices, interfaces = usb_scan.get_usb_info()
+
+        # debug info
```
mseri: We should not copy paste the code, instead we should refactor `usb_scan.py` to be better suited for testing.
I think we need to extract https://github.com/xapi-project/xen-api/blob/3bb747104ee0399c3d0e3c747b1d8ff95669b269/scripts/usb_scan.py#L627-L637 into a function that takes `devices` and `interfaces` as parameters and returns the object that we will serialise to json. Then this test should call that function with the mocked devices and interfaces

scripts/test_usb_scan.py
```diff
@@ -0,0 +1,393 @@
...
+        self.test_usb_exit([], [], [], "not_exit.conf")
+
+    @nottest
+    def test_usb_config_error_common(self, content):
+        path = os.path.join(self.work_dir, "usb-policy.conf")
+        with open(path, "w") as f:
+            f.write(content)
+        self.test_usb_exit([], [], [], path)
+
+    def test_usb_config_error_01(self):
```
mseri: This is just a stylistic thing. Here and below, give meaningful names to the error functions. For example, in this case, `test_usb_config_error_unexpected_characters` and in the next one `test_usb_config_error_duplicated_key`, and so on

scripts/test_usb_scan.py
TalonsLee: OK. remove log file, and make test_log empty.

scripts/test_usb_scan.py
TalonsLee: OK. This will change the usb_scan.py,
I have to run more test next week.

scripts/test_usb_scan.py
TalonsLee: done.

scripts/test_usb_scan.py
TalonsLee: OK.
    @nottest
    def test_usb_exit(self, devices, interfaces, results,
                      path="./scripts/usb-policy.conf", msg=""):
        with self.assertRaises(SystemExit) as cm:
            self.test_usb_common(devices, interfaces, results, path)
        exception = cm.exception
        print exception.message
        if msg:
            self.assertIn(msg, exception.message)

scripts/test_usb_scan.py
mseri: Yes, call it something like `make_pusbs_list(interfaces, devices)` and the unit test that function

scripts/usb_scan.py
```diff
@@ -607,19 +607,14 @@ def to_pusb(device):
...
+def make_pusbs_list(devices, interfaces):
+    """ check the USB devices and interfaces against policy file,
+    and return the pusb list that can be passed through
 
+    :param devices:([UsbDevice]) USB device list we found in host
+    :param interfaces:([UsbInterface]) USB interface list found in host
+    :return:([pusb]) the pusb list that can be passed through
+    """
     # debug info
     log_list(devices)
```
mseri: Please move these debug info logs to the main body before L648, or remove them completely

PR #3326, opened 2017-10-30
CA-178651: FCoE NIC is not prevented from unplugging

krizex:
Prevent unplugging PIF which is used by FCoE

When unplugging PIF, if the PIF is used by FCoE SR, we should prevent the action.

### DIFF
--- a/ocaml/xapi-consts/api_errors.ml
+++ b/ocaml/xapi-consts/api_errors.ml
@@ -118,6 +118,7 @@ let pif_incompatible_primary_address_type = "PIF_INCOMPATIBLE_PRIMARY_ADDRESS_TY
 let required_pif_is_unplugged = "REQUIRED_PIF_IS_UNPLUGGED"
 let pif_not_present = "PIF_NOT_PRESENT"
 let pif_does_not_allow_unplug = "PIF_DOES_NOT_ALLOW_UNPLUG"
+let pif_has_fcoe_sr_in_use = "PIF_HAS_FCOE_SR_IN_USE"
 let pif_unmanaged = "PIF_UNMANAGED"
 let cannot_plug_bond_slave = "CANNOT_PLUG_BOND_SLAVE"
 let cannot_add_vlan_to_bond_slave = "CANNOT_ADD_VLAN_TO_BOND_SLAVE"
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -605,6 +605,8 @@ let _ =
     ~doc:"This host has no PIF on the given network." ();
   error Api_errors.pif_does_not_allow_unplug [ "PIF" ]
     ~doc:"The operation you requested cannot be performed because the specified PIF does not allow unplug." ();
+  error Api_errors.pif_has_fcoe_sr_in_use ["PIF"; "SR"]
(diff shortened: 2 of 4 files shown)


ocaml/xapi-consts/api_errors.ml
```diff
@@ -118,6 +118,7 @@ let pif_incompatible_primary_address_type = "PIF_INCOMPATIBLE_PRIMARY_ADDRESS_TY
 let required_pif_is_unplugged = "REQUIRED_PIF_IS_UNPLUGGED"
 let pif_not_present = "PIF_NOT_PRESENT"
 let pif_does_not_allow_unplug = "PIF_DOES_NOT_ALLOW_UNPLUG"
+let pif_in_fcoe_not_allow_unplug = "PIF_IN_FCOE_NOT_ALLOW_UNPLUG"
```
YarsinCitrix: I think you missed change in datamodel.ml.

krizex: Yes, I will add this new-added error and the remaining errors to `datamodel.ml`

jonludlam:
Is the capability always there or only when the PIF is actually in use by a FCoE SR?

krizex:
@jonludlam If user enable FCoE on the ethernet interface that this PIF used, the fcoe capability always exists.

ocaml/xapi/xapi_network_attach_helpers.ml
```diff
@@ -56,6 +57,13 @@ let assert_pif_disallow_unplug_not_set ~__context pif =
   if (Db.PIF.get_disallow_unplug ~__context ~self:pif) then
     raise (Api_errors.Server_error(Api_errors.pif_does_not_allow_unplug, [ Ref.string_of pif ]))
 
+let assert_pif_not_in_fcoe ~__context pif =
+  let device_name = Db.PIF.get_device ~__context ~self:pif in
+  let dbg = Context.string_of_task __context in
+  let capabilities = Net.Interface.get_capabilities dbg ~name:device_name in
```
YarsinCitrix: For the purpose of this change, it is at unplug stage so the pif data should already be updated with the capacities from the Net.Interface. So the check should be done based on the pif data (in xapi database), in stead of another call to the network interface.

krizex: Seems make sense, I will check the code to find out how xapi synchronize the field value with the command line outputs.

krizex: Though Db.PIF.capabilities is setted when creating or during a new scan, that means its value won't change if we do not trigger `PIF.scan`. After user change the interface capabilities, it's necessary for user to run `xe pif-scan` to update.

ocaml/xapi/xapi_network_attach_helpers.ml
```diff
@@ -18,6 +18,7 @@ module D=Debug.Make(struct let name="xapi" end)
 open D
 
 open Db_filter_types
+open Network
```
YarsinCitrix: so you do not need this now?

robhoes:
I _do_ think that the capability reflects whether a NIC supports FCoE, and not whether is it currently using it. Information about the use of FCoE has to come from the storage layer, somehow, but I don't know if this is easily available.

jonludlam:
It's a bit of a nasty layering violation, but I don't see any better solution that we can do soon.

PR #3327, opened 2017-10-31
CA-262059: No SM backends support VDI.resize_online - should be deprecated & removed

krizex:
Deprecate VDI.resize_online because there are no sm backend support this feature.

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -3442,6 +3442,10 @@ let vdi_resize_online = call
     ~name:"resize_online"
     ~in_oss_since:None
     ~in_product_since:rel_rio
+    ~lifecycle: [
+      Published, rel_rio, "Online resize vdi";
+      Removed, rel_inverness, "Remove feature: online resize vdi"
+    ]
     ~params:[Ref _vdi, "vdi", "The VDI to resize"; Int, "size", "The new size of the VDI" ]
     ~doc:"Resize the VDI which may or may not be attached to running guests."
     ~allowed_roles:_R_VM_ADMIN
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -3443,8 +3443,8 @@ let vdi_resize_online = call
     ~in_oss_since:None
     ~in_product_since:rel_rio
     ~lifecycle: [
-      Published, rel_rio, "Online resize vdi";
-      Removed, rel_inverness, "Remove feature: online resize vdi"
+      Published, rel_rio, "";
+      Removed, rel_inverness, "Remove feature of online VDI resize because it is not supported by any of the storage backends."
     ]
     ~params:[Ref _vdi, "vdi", "The VDI to resize"; Int, "size", "The new size of the VDI" ]
     ~doc:"Resize the VDI which may or may not be attached to running guests."
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -3444,7 +3444,7 @@ let vdi_resize_online = call
     ~in_product_since:rel_rio
     ~lifecycle: [
       Published, rel_rio, "";
-      Removed, rel_inverness, "Remove feature of online VDI resize because it is not supported by any of the storage backends."
+      Removed, rel_inverness, "Online VDI resize is not supported by any of the storage backends."
     ]
     ~params:[Ref _vdi, "vdi", "The VDI to resize"; Int, "size", "The new size of the VDI" ]
     ~doc:"Resize the VDI which may or may not be attached to running guests."
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -3441,7 +3441,6 @@ let vdi_resize = call
 let vdi_resize_online = call
     ~name:"resize_online"
     ~in_oss_since:None
-    ~in_product_since:rel_rio
     ~lifecycle: [
       Published, rel_rio, "";
       Removed, rel_inverness, "Online VDI resize is not supported by any of the storage backends."
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -3614,15 +3614,6 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
            forward_vdi_op ~local_fn ~__context ~self:vdi
              (fun session_id rpc -> Client.VDI.resize rpc session_id vdi size))
 
-    let resize_online ~__context ~vdi ~size =
-      info "VDI.resize_online: VDI = '%s'; size = %Ld" (vdi_uuid ~__context vdi) size;
-      let local_fn = Local.VDI.resize_online ~vdi ~size in
-      let sR = Db.VDI.get_SR ~__context ~self:vdi in
-      with_sr_andor_vdi ~__context ~sr:(sR, `vdi_resize) ~vdi:(vdi, `resize_online) ~doc:"VDI.resize_online"
-        (fun () ->
-           forward_vdi_op ~local_fn ~__context ~self:vdi
-             (fun session_id rpc -> Client.VDI.resize_online rpc session_id vdi size))
-
     let generate_config ~__context ~host ~vdi =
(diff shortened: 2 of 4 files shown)


YarsinCitrix:
From the mere simple description from the CA ticket, it seems to me that the intention is we do not actually support online resize of the VDI, currently literally supported by vdi.resize_online, so we want to deprecate it. From code perspective, we know that vdi.resize is actually vdi.resize_online, and I did not find anything saying vdi.resize should be offline. So I have a concern on the overall solution here, as we simply deprecate the vdi.resize_online, we actually did not prevent any calling from API into vdi.resize (for online vdi) which actually do the work. Customer may also be confused to think if we merged resize_online to resize, as we simple remove one. So my suggestion is to enforce vdi.resize only do on offline VDIs, and also make this clear in customer documents related.

krizex:
@YarsinCitrix Yes, for now there are no document indicates that user only can resize vdi offline, but when user resize a vdi that in use, user will get an error "VDI_IN_USE". So it's better to add additional description about `xe vdi-resize` in XenServer Administrators guide.

YarsinCitrix:
You sure about the "VDI_IN_USE"? cos current call for vdi.resize_online/resize is also calling the same function body, so I do not think there will be an error if it is in use. I suspect we have to add something like that to indicate the case.

krizex:
@YarsinCitrix Yes, pls check following log.
```
Oct 31 07:50:39 dt77 xapi: [error|dt77|162297 UNIX /var/lib/xcp/xapi|dispatch:VDI.resize D:2967aaa883ce|backtrace] VDI.resize R:79678a01500e failed with exception Server_error(VDI_IN_USE, [ OpaqueRef:1e404c56-e0e9-c771-1fe8-bb6a78240a4a; resize ])
Oct 31 07:50:39 dt77 xapi: [error|dt77|162297 UNIX /var/lib/xcp/xapi|dispatch:VDI.resize D:2967aaa883ce|backtrace] Raised Server_error(VDI_IN_USE, [ OpaqueRef:1e404c56-e0e9-c771-1fe8-bb6a78240a4a; resize ])
Oct 31 07:50:39 dt77 xapi: [error|dt77|162297 UNIX /var/lib/xcp/xapi|dispatch:VDI.resize D:2967aaa883ce|backtrace] 1/6 xapi @ dt77 Raised at file ocaml/xapi/rbac.ml, line 236
Oct 31 07:50:39 dt77 xapi: [error|dt77|162297 UNIX /var/lib/xcp/xapi|dispatch:VDI.resize D:2967aaa883ce|backtrace] 2/6 xapi @ dt77 Called from file ocaml/xapi/server_helpers.ml, line 76
Oct 31 07:50:39 dt77 xapi: [error|dt77|162297 UNIX /var/lib/xcp/xapi|dispatch:VDI.resize D:2967aaa883ce|backtrace] 3/6 xapi @ dt77 Called from file ocaml/xapi/server_helpers.ml, line 91
Oct 31 07:50:39 dt77 xapi: [error|dt77|162297 UNIX /var/lib/xcp/xapi|dispatch:VDI.resize D:2967aaa883ce|backtrace] 4/6 xapi @ dt77 Called from file lib/pervasiveext.ml, line 22
Oct 31 07:50:39 dt77 xapi: [error|dt77|162297 UNIX /var/lib/xcp/xapi|dispatch:VDI.resize D:2967aaa883ce|backtrace] 5/6 xapi @ dt77 Called from file lib/pervasiveext.ml, line 26
Oct 31 07:50:39 dt77 xapi: [error|dt77|162297 UNIX /var/lib/xcp/xapi|dispatch:VDI.resize D:2967aaa883ce|backtrace] 6/6 xapi @ dt77 Called from file lib/backtrace.ml, line 176
```

gaborigloi:
Luckily we already have different checks for `VDI.resize` and `VDI.resize_online`, even though they share the same implementation. From `Xapi_vdi.check_operation_error`, which contains the VDI operation checks [done in the message forwarding layer](https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/message_forwarding.ml#L3328) (and is also [used](https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_vdi.ml#L310) to update the `allowed_operations` of the VDI), we can see that `resize_online` is allowed live, while `resize` isn't: https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_vdi.ml#L139 (In general, this function is the first place to go to if one wants to check when a given VDI operation is or isn't allowed.)

gaborigloi:
As @krizex suggested, we should make this "removed" straight away, because it probably never worked, so probably nobody is using it. This wouldn't be the only call that got removed without being deprecated: https://github.com/xapi-project/xen-api/blob/master/ocaml/idl/datamodel.ml#L4478

krizex:
PR updated.
Manual test passed:
```
[root@xrtuk-09-04 ~]# xe vdi-resize online=true uuid=8f3cb4ad-1736-4e0b-aa98-6cbc00846d97 disk-size=15GiB
This function is no longer available.
```

ocaml/idl/datamodel.ml
```diff
@@ -3442,6 +3442,10 @@ let vdi_resize_online = call
     ~name:"resize_online"
     ~in_oss_since:None
     ~in_product_since:rel_rio
+    ~lifecycle: [
+      Published, rel_rio, "Online resize vdi";
+      Removed, rel_inverness, "Remove feature: online resize vdi"
```
gaborigloi: Here we could explain that we removed this call because it was never supported. We could say something like:
```
Online VDI resize is not supported by any of the storage backends.
```

ocaml/idl/datamodel.ml
```diff
@@ -3442,6 +3442,10 @@ let vdi_resize_online = call
     ~name:"resize_online"
     ~in_oss_since:None
     ~in_product_since:rel_rio
+    ~lifecycle: [
+      Published, rel_rio, "Online resize vdi";
```
gaborigloi: I think it's fine to have the empty string here:
```
Published, rel_rio, "";
```
Otherwise, I don't know what we could say here, as this was never supported.

gaborigloi:
It looks like there are still many "removed" stubs lying around in our codebase that are unnecessary, as we discussed (for example, xapi_vmpp.ml, we can grep for `message_removed` for more). We should also remove all of these, but it's fine to do that in a separate ticket & PR.

krizex:
@gaborigloi Thanks for your comments, I have updated the PR. For the other `removed` stubs, I think we should create a separate ticket & PR to fix them as you said.

ocaml/idl/datamodel.ml
```diff
@@ -3442,6 +3442,10 @@ let vdi_resize_online = call
     ~name:"resize_online"
     ~in_oss_since:None
     ~in_product_since:rel_rio
+    ~lifecycle: [
+      Published, rel_rio, "";
+      Removed, rel_inverness, "Remove feature of online VDI resize because it is not supported by any of the storage backends."
```
gaborigloi: I think we should phrase this a bit differently: `"Remove online VDI resize feature because ..."`. @lindig any opinions about how to best phrase this?

krizex: That's an useful suggestion from native speaker, thank you. ;-)

lindig: Since it is obvious that we removed an API call, I would simply state the reason: `Online VDI resize is not supported by any of the storage backends.`

krizex: Updated per your comments.

ocaml/idl/datamodel.ml
```diff
@@ -3442,6 +3442,10 @@ let vdi_resize_online = call
     ~name:"resize_online"
     ~in_oss_since:None
     ~in_product_since:rel_rio
```
gaborigloi: I've just realized that we can remove `~in_product_since:rel_rio`, because that's already specified by the `~lifecycle`: https://github.com/xapi-project/xen-api/blob/master/ocaml/idl/datamodel.ml#L3687

krizex:
I was aware of that, but forgot to make that change... Will update tomorrow. Thanks

krizex:
@gaborigloi Updated.

PR #3329, opened 2017-11-03
[CA-271014] Only get local SRs on the physical utilisation thread

mcintyre94:
This greatly improves the performance of the pool master in a large pool. The current loop means that every 2 minutes each host in the pool queries, for each PBD (regardless of which host it's on), whether it's currently attached, and its SR. It then updates the xapi database with the RRDD physical utilisation value for those SRs it has the data. As an optimisation therefore, this commit adds a new method to only get PBDs on the calling host, and then the SRs for those will be those that it can update the RRDD data for.

The issue with this physical utilisation thread is the overhead on the pool master, which must handle all the database queries. This is now only one query to get the PBDs currently attached on the calling host, plus one more for each PBD currently plugged on localhost (to get its SR). It was previously one query to get all PBDs on the pool (regardless of host), plus an extra two queries for each PBD to get its currently_attached and then its SR.

If a pool has n hosts with m PBDs each (all currently attached), then we previously had n hosts making 2nm+1 database queries each every 2 minutes, and we now have n hosts making 1+m database queries each every 2 minutes.

Passing BVT: suite run 73964
Also see the ticket for examples of performance improvement from this fix on a large pool

Signed-off-by: Callum McIntyre <callum.mcintyre@citrix.com>

### DIFF
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -1016,7 +1016,7 @@ let get_all_plugged_srs ~__context =
     Eq (Field "currently_attached", Literal "true")) in
   List.setify (List.map (fun self -> Db.PBD.get_SR ~__context ~self) pbds_plugged_in)
 
-let get_all_plugged_srs_local ~__context =
+let get_local_plugged_srs ~__context =
   let localhost = get_localhost __context in
   let localhost = Ref.string_of localhost in
   let my_pbds_plugged_in = Db.PBD.get_refs_where  ~__context ~expr:(And (
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -1012,8 +1012,8 @@ let i_am_srmaster ~__context ~sr =
   get_srmaster ~__context ~sr = get_localhost ~__context
 
 let get_all_plugged_srs ~__context =
-  let pbds = Db.PBD.get_all ~__context in
-  let pbds_plugged_in = List.filter (fun self -> Db.PBD.get_currently_attached ~__context ~self) pbds in
+  let pbds_plugged_in = Db.PBD.get_refs_where ~__context ~expr:(
+    Eq (Field "currently_attached", Literal "true")) in
   List.setify (List.map (fun self -> Db.PBD.get_SR ~__context ~self) pbds_plugged_in)
 
 let get_all_plugged_srs_local ~__context =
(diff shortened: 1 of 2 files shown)


ocaml/xapi/helpers.ml
```diff
@@ -1016,6 +1016,16 @@ let get_all_plugged_srs ~__context =
   let pbds_plugged_in = List.filter (fun self -> Db.PBD.get_currently_attached ~__context ~self) pbds in
   List.setify (List.map (fun self -> Db.PBD.get_SR ~__context ~self) pbds_plugged_in)
 
+let get_all_plugged_srs_local ~__context =
+  let localhost = get_localhost __context in
+  let localhost = Ref.string_of localhost in
+  let my_pbds_plugged_in = Db.PBD.get_refs_where  ~__context ~expr:(And (
```
mseri: You could use `get_all_where` also in the function above, to reduce the Db calls also on the rest of `get_all_plugged_srs` invocations

mcintyre94: This sounds like a good idea :) Is there a reason you suggested `get_all_where` here while I used `get_refs_where` in the new function - and the `List.setify` line following is the same? Is `get_all_where` prefered when you're getting a lot of records?

mseri: You are right, you just need `get_refs_where`

mseri: It may be worth checking if getting all the relevant records is faster than making all the separate call to get the SRs or not.

mcintyre94:
My large pool environment doesn't really simulate realistic networking enough to make the judgement of `get_records_where` vs `get_refs_where`, and the performance difference will be very small compared to the current PR anyway. Unless anyone has a preference/policy either way I'd suggest we keep this as is for now - we can always make the smaller change if more realistic profiling suggests it'd help.

ocaml/xapi/helpers.ml
lindig: I'm happy to take this PR as it is already an improvement. I have only a minor comment: To my ear `get_all_local_plugged_srs` is easier to understand than `get_all_plugged_srs_local` or would that distort the meaning in some way? If not, I would like to suggest this change.

mcintyre94: How about `get_local_plugged_srs`? I think that's the most clear, agreed the local at the end is a bit awkard.

lindig: Sounds good to me.

robhoes:
The fixup commit should have been squashed before merging.

PR #3331, opened 2017-11-06
CA-269706: Add check to prevent snapshot-destroy command to destroy n…

YarsinCitrix:
…ormal none snapshot VM by mistake.

### DIFF
--- a/ocaml/xapi/cli_operations.ml
+++ b/ocaml/xapi/cli_operations.ml
@@ -2665,16 +2665,21 @@ let get_snapshot_uuid params =
   then List.assoc "uuid" params
   else raise (failwith "Required parameter not found: snapshot-uuid or uuid.")
 
+let get_snapshotVM_by_uuid rpc session_id snap_uuid =
+  let snap_ref = Client.VM.get_by_uuid rpc session_id snap_uuid in
+  if not (Client.VM.get_is_a_snapshot rpc session_id snap_ref) then failwith "This operation can only be done on VM snapshot.";
+  snap_ref
+
 let snapshot_revert printer rpc session_id params =
   let snap_uuid = get_snapshot_uuid params in
-  let snap_ref = Client.VM.get_by_uuid rpc session_id snap_uuid in
+  let snap_ref = get_snapshotVM_by_uuid rpc session_id snap_uuid in
   Client.VM.revert ~rpc ~session_id ~snapshot:snap_ref
 
 let snapshot_op op printer rpc session_id params =
   let new_name = List.assoc "new-name-label" params in
   let desc = if List.mem_assoc "new-name-description" params then Some (List.assoc "new-name-description" params) else None in
   let uuid = get_snapshot_uuid params in
-  let ref = Client.VM.get_by_uuid ~rpc ~session_id ~uuid in
+  let ref = get_snapshotVM_by_uuid rpc session_id uuid in
   let new_ref = op ~rpc ~session_id ~vm:ref ~new_name in
   ignore (may (fun desc -> Client.VM.set_name_description rpc session_id new_ref desc) desc);
   let new_uuid = Client.VM.get_uuid ~rpc ~session_id ~self:new_ref in
@@ -2689,14 +2694,13 @@ let snapshot_copy printer rpc session_id params =
 
 let snapshot_destroy printer rpc session_id params =
   let snap_uuid = get_snapshot_uuid params in
-  let snap_ref = Client.VM.get_by_uuid rpc session_id snap_uuid in
-  if not (Client.VM.get_is_a_snapshot rpc session_id snap_ref) then failwith "This operation can only destroy VM snapshot.";
+  let snap_ref = get_snapshotVM_by_uuid rpc session_id snap_uuid in
   if Client.VM.get_power_state rpc session_id snap_ref <> `Halted then Client.VM.hard_shutdown ~rpc ~session_id ~vm:snap_ref;
   Client.VM.destroy ~rpc ~session_id ~self:snap_ref
 
 let snapshot_uninstall fd printer rpc session_id params =
   let snap_uuid = get_snapshot_uuid params in
-  let snap_ref = Client.VM.get_by_uuid rpc session_id snap_uuid in
+  let snap_ref = get_snapshotVM_by_uuid rpc session_id snap_uuid in
   vm_uninstall_common fd printer rpc session_id params [snap_ref]
 
 let vm_copy printer rpc session_id params =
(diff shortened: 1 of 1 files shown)


ocaml/xapi/cli_operations.ml
```diff
@@ -2690,6 +2690,7 @@ let snapshot_copy printer rpc session_id params =
 let snapshot_destroy printer rpc session_id params =
   let snap_uuid = get_snapshot_uuid params in
   let snap_ref = Client.VM.get_by_uuid rpc session_id snap_uuid in
+  if (not Client.VM.get_is_a_snapshot rpc session_id snap_ref) then failwith "This operation can only destroy VM snapshot.";
```
minishrink: The whole expression `Client.VM.get_is_a_snapshot rpc session_id snap_ref` is of type bool, so you need to wrap it in parentheses before applying `not`
i.e. `(not (Client.VM.get_is_a_snapshot rpc session_id snap_ref))`

lindig: Not quite. It's enough to do 
```
if not (Client.VM.get_is_a_snapshot rpc session_id snap_ref) then ...
```

You can avoid binding names that are only used once (but this is a matter of taste). I'd also use `vm` or `snapshot` as a name as I don't think types should be encoded into names.

```
let vm = get_snapshot_uuid params |>  Client.VM.get_by_uuid rpc session_id in
```

YarsinCitrix: Thanks for catching this out! I made the change to parenthesize what's after not when I build and test, but seems missed the updates when I do PR.

ocaml/xapi/cli_operations.ml
```diff
@@ -2690,6 +2690,7 @@ let snapshot_copy printer rpc session_id params =
 let snapshot_destroy printer rpc session_id params =
   let snap_uuid = get_snapshot_uuid params in
   let snap_ref = Client.VM.get_by_uuid rpc session_id snap_uuid in
+  if not (Client.VM.get_is_a_snapshot rpc session_id snap_ref) then failwith "This operation can only destroy VM snapshot.";
```
lindig: There is grammar issue; it has to be either of these:

* This operation can only destroy *a* VM snapshot.
* This operation can only destroy VM *snapshots*.

Please:

* update the commit message such that it is clearly separated into a subject and a body. 
* do not force push over the current commit because this loses the discussion. We can always squash commits before merging.

gaborigloi:
I think we could generalize this solution to fix all `xe snapshot-...` operations, because all of them are affected. Maybe we could move this check into the `get_snapshot_uuid` function, because that seems to be called from all these snapshot CLI operations?

lindig:
I like the suggestion to check this. But in this case the function would not return a uuid but a vm reference. So maybe 

* create a function `get_snapshot` that takes the parameter list,
* checks that this is indeed a snapshot,
* returns the VM reference;
* update all functions to use `get_snapshot` instead of `get_snapshot_uuid`

YarsinCitrix:
All, it is a good suggestion to update all other similar command restrictions!
I am thinking if we do the check along with the current get_snapshot will complicate this function which is merely used to fetch parameter from a list; besides the concern is do check there will introduce another call to VM.get_by_uuid, which is actually unnecessary. So I added a function to do the VM.get_by_uuid along with the check, and updated other functions to call the new functions. Please comments, thank you in advance!

ocaml/xapi/cli_operations.ml
```diff
@@ -2665,16 +2665,21 @@ let get_snapshot_uuid params =
   then List.assoc "uuid" params
   else raise (failwith "Required parameter not found: snapshot-uuid or uuid.")
 
+let get_snapshotVM_by_uuid rpc session_id snap_uuid =
+  let snap_ref = Client.VM.get_by_uuid rpc session_id snap_uuid in
+  if not (Client.VM.get_is_a_snapshot rpc session_id snap_ref) then failwith "This operation can only be done on VM snapshot.";
```
lindig: Suggestion: `"This operation can only be performed on a VM snapshot. %s is not a VM snapshot." snap_uuid`.

gaborigloi:
I have checked, and wherever we use `get_snapshot_uuid`, it is immediately followed by a `VM.get_by_uuid`, and we do not actually use the snapshot's UUID anywhere else:
```
2669   let snap_uuid = get_snapshot_uuid params in
2670   let snap_ref = Client.VM.get_by_uuid rpc session_id snap_uuid in
```
So I think we could simplify the code a bit by replacing `get_snapshot_uuid` with `get_snapshot`, which would return the VM ref.

Also, I've checked that all of the CLI snapshot operations use `get_snapshot_uuid`, except one: the `let snapshot_export` function. So we'd have to update that code too to use this helper.

YarsinCitrix:
gaborigloi, it is good suggestion that we also fix similar issue in xe template- serial. I am in line with that. However, in case someday someone might have different opinion regarding changes for one of the two command(group), I would suggest to create another ticket and then initiate changes tracked by the new ticket specifically for the xe template- command (group). Would you agree?

I have updated the message as Chris suggested, which looks better. And I have also updated the part for snapshot-export part. A similar part will be done when I make changes for the xe template- command (group) there.

ocaml/xapi/cli_operations.ml
YarsinCitrix: Chris, I have updated the message as you suggested, would you fine with the current wording? Do you still have other comments, or may I proceed to rebase from master and squash the comments to prepare for merge?

ocaml/xapi/cli_operations.ml
```diff
@@ -3821,6 +3827,9 @@ let vm_export_aux obj_type fd printer rpc session_id params =
   let num = ref 1 in
   let uuid = List.assoc (obj_type ^ "-uuid") params in
   let ref = Client.VM.get_by_uuid rpc session_id uuid in
+  if obj_type = "snapshot" then
```
lindig: This could be simplified to 
```
if obj_type = "snapshot" && not (Client.VM.get_is_a_snapshot rpc session_id ref) then
  failwith (Printf.sprintf ..);
```

ocaml/xapi/cli_operations.ml
```diff
@@ -2665,16 +2665,22 @@ let get_snapshot_uuid params =
   then List.assoc "uuid" params
   else raise (failwith "Required parameter not found: snapshot-uuid or uuid.")
 
+let get_snapshotVM_by_uuid rpc session_id snap_uuid =
```
lindig: Minor point: `get_snapshot_vm_by_uuid` would be more common. I don't request a change.

gaborigloi:
@YarsinCitrix Agreed. I think we should do that next to make the behaviour consistent. Then the code for `snapshot_export` will reach its final form.

PR #3335, opened 2017-11-09
CA-271852 XenServer build date hardcoded to 1970-01-01

krizex:
Get build date from `xensource-inventory` instead of hard coding to `1970-01-01`.
This PR depends on the change of other repos, I have recorded them in the ticket.

### DIFF
--- a/ocaml/util/xapi_version.ml
+++ b/ocaml/util/xapi_version.ml
@@ -8,6 +8,6 @@
 
 let git_id = ""
 let hostname = "localhost"
-let date = "1970-01-01"
+let date = Build_info.date
 let xapi_version_major = 1
 let xapi_version_minor = 20
--- a/ocaml/util/jbuild
+++ b/ocaml/util/jbuild
@@ -1,7 +1,12 @@
+(rule
+ ((targets (build_info.ml))
+  (action  (with-stdout-to ${@} (run "date" "+let date=\"%Y-%m-%d\"")))))
+
 (library
  ((name xapi_version)
   (modules (
    xapi_version
+   build_info
   ))
   (libraries (
    xcp


ocaml/util/xapi_version.ml
```diff
@@ -1,13 +1,13 @@
...
- let product_brand () = Inventory.lookup ~default:"" "PRODUCT_BRAND" 
- let build_number () = Inventory.lookup ~default:"" "BUILD_NUMBER" 
+let product_version () = Inventory.lookup ~default:"" "PRODUCT_VERSION" 
+let product_version_text () = Inventory.lookup ~default:"" "PRODUCT_VERSION_TEXT" 
+let product_version_text_short () = Inventory.lookup ~default:"" "PRODUCT_VERSION_TEXT_SHORT" 
+let platform_name () = Inventory.lookup ~default:"" "PLATFORM_NAME" 
+let platform_version () = Inventory.lookup ~default:"0.0.0" "PLATFORM_VERSION" 
+let product_brand () = Inventory.lookup ~default:"" "PRODUCT_BRAND" 
+let build_number () = Inventory.lookup ~default:"" "BUILD_NUMBER" 
+let build_date () = Inventory.lookup ~default:"" "BUILD_DATE"
```
lindig: Maybe we should use some default like "2017-01-01"? On a current build, I see no `BUILD_DATE` entry in `/etc/xensource-inventory` such that this would fall back to the default. How are we going to get this into the inventory or is there some other PR that takes care of this?

lindig:
The commit message can be improved: it describes the problem but should describe the solution. What you said in the comment should go into the commit message.

ocaml/util/xapi_version.ml
krizex: Wrt to the default value, maybe we can keep the original default value `1970-01-01`.
There are some other PRs to other repos, I have posted them in the ticket.

PR #3337, opened 2017-11-14
CA-273239: Template-xxx CLI commands are allowed on non-template VMs

YarsinCitrix:


ocaml/xapi/cli_operations.ml
```diff
@@ -3827,6 +3833,9 @@ let vm_export_aux obj_type fd printer rpc session_id params =
   let num = ref 1 in
   let uuid = List.assoc (obj_type ^ "-uuid") params in
   let ref = Client.VM.get_by_uuid rpc session_id uuid in
+  if obj_type = "template" then
```
lindig: Simplify to:
```
if obj_type = "template" && not (Client.VM.get_is_a_template rpc session_id ref) then
...
```
The same simplification can be used for snapshots below.

PR #3338, opened 2017-11-15
CA-273649: add xapi restriction when attach several host's usbs

taoyongd:
This change is to add restriction when attaching usb to vm that
the host of the usb is not in possible_hosts that vm can boot on.


ocaml/idl/datamodel.ml
```diff
@@ -1515,7 +1515,9 @@ let _ =
   error Api_errors.pusb_vdi_conflict [ "PUSB"; "VDI" ]
     ~doc:"The VDI corresponding to this PUSB has existing VBDs." ();
   error Api_errors.vm_has_vusbs ["VM"]
-    ~doc:"The operation is not allowed when the VM has VUSBs." ()
+    ~doc:"The operation is not allowed when the VM has VUSBs." ();
+  error Api_errors.pusb_not_in_possible_hosts [ "pusb"; "possible_hosts" ]
+    ~doc:"The host of the pusb which will be attached to the vm must in possible hosts that the vm can boot on." ()
```
lindig: I think the error message is not clear and would suggest this one unless it does not capture the problem:
```
The VM can't boot on the host that belongs to the pusb; it can boot only on the following hosts.
```

ocaml/xapi/xapi_vusb.ml
```diff
@@ -34,6 +34,13 @@ let create ~__context ~vM ~uSB_group ~other_config =
         another vusb can not create with the same USB_group. *)
     if vusbs <> [] then
       raise (Api_errors.Server_error(Api_errors.usb_group_conflict, [Ref.string_of uSB_group]));
+
+    (* The host of the PUSB which will be attached to the vm must in possible_hosts that the vm can boot on.*)
```
lindig: The code is fine but I think the error condition is not well captured in the comment or the error message. Make this as clear as possible.

ocaml/idl/datamodel.ml
taoyongd: Better error message. Thanks

ocaml/xapi/xapi_vusb.ml
taoyongd: How about "All USBs that will be attached to the VM must be on the same host, as USB from host A can not attach to the vm resident on host  B. "

ocaml/xapi/xapi_vusb.ml
```diff
@@ -34,6 +34,14 @@ let create ~__context ~vM ~uSB_group ~other_config =
...
     if vusbs <> [] then
       raise (Api_errors.Server_error(Api_errors.usb_group_conflict, [Ref.string_of uSB_group]));
+
+    (* All USBs that will be attached to the VM must be on the same host, as USB from host A can
+      not attach to the vm resident on host B. *)
+    let possible_hosts = Xapi_vm.get_possible_hosts ~__context ~vm:vM in
+    let pusb = Helpers.get_first_pusb ~__context uSB_group in
+    let host = Db.PUSB.get_host ~__context ~self:pusb in
+    if not (List.mem host possible_hosts) then
+      raise (Api_errors.Server_error (Api_errors.pusb_not_in_possible_hosts, [Ref.string_of pusb; String.concat ";" (List.map Ref.string_of possible_hosts)]));
```
lindig: The error message is showing the references but I think it should show the UUIDs because this is how the user thinks about this.

taoyongd: you mean pusb uuid host uuid, right? In the error message it is showing that.
```The VM can't boot on the host that belongs to the pusb; it can boot only on the following hosts.
pusb: 8b0ae24d-58b3-263d-6aee-68737629e015
```

taoyongd:
``` raise (Api_errors.Server_error (Api_errors.pusb_not_in_possible_hosts, [Db.PUSB.get_uuid ~__context ~self:pusb; 
        String.concat ";" (List.map (fun host -> Db.Host.get_uuid ~__context ~self:host) possible_hosts)]));
``` 
This can get uuid. Seems a bit long.

lindig:
```
let pusb_uuid  = Db.PUSB.get_uuid ~__context ~self:pusb in
let host_uuids = List.map (fun self -> Db.Host.get_uuid ~__context ~self) possible_hosts in
  raise Api_errors.(Server_error (pusb_not_in_possible_hosts, [ pusb_uuid;  host_uuids]))
```

robhoes:
In other words, I think that we should not do this, and not attempt to "fix" this at all.

taoyongd:
Yes @robhoes, VM.start indeed do this. As now xenrt case catch error in VUSB.create in the pool scenario which I think it should catch in vm.start. I think we can ignore this fix and modify the xenrt case to do this. How do you think so?

robhoes:
How come VUSB.create already raises an error? If so, why is this PR adding another error?

taoyongd:
This error is throwed by xenrt case which it found call VUSB.create on this case should catch error but not. So itself throwed an error

robhoes:
OK, let's just close this then.

PR #3341, opened 2017-11-16
CA-274079: Removing one Pusb removes all Pusbs in the pool

taoyongd:
Two exactly same pusbs in two hosts of a pool will be deleted when
unplug one of the pusbs, this change is to fix it.

### DIFF
--- a/ocaml/xapi/xapi_pusb.ml
+++ b/ocaml/xapi/xapi_pusb.ml
@@ -31,10 +31,12 @@ let create ~__context ~uSB_group ~host ~other_config ~path
 
 let scan_start ~__context usbs =
   let host = Helpers.get_localhost ~__context in
-  let known_usb =
+  let local_pusb_records =
     Db.PUSB.get_all_records ~__context
     |> List.filter (fun (rf, rc) -> rc.API.pUSB_host = host)
-    |> get_known_usb
+  in
+  let known_usb =
+    local_pusb_records |> get_known_usb
   in
   let local_usb = get_local_usb usbs in
   (* Create the newly added pusbs *)
@@ -45,7 +47,7 @@ let scan_start ~__context usbs =
                Db.PUSB.set_USB_group ~__context ~self ~value:group
   ) (USBSet.diff local_usb known_usb);
 
-  List.filter (fun (rf, rc) -> USBSet.mem (extract_known_usb_info rc) (USBSet.diff known_usb local_usb)) (Db.PUSB.get_all_records ~__context)
+  List.filter (fun (rf, rc) -> USBSet.mem (extract_known_usb_info rc) (USBSet.diff known_usb local_usb)) local_pusb_records
   |> List.iter (fun (self, _) ->
                   try
                     Xapi_pusb_helpers.destroy_pusb ~__context self;
--- a/ocaml/xapi/xapi_pusb.ml
+++ b/ocaml/xapi/xapi_pusb.ml
@@ -30,6 +30,11 @@ let create ~__context ~uSB_group ~host ~other_config ~path
   pusb
 
 let scan_start ~__context usbs =
+  (* We compute two sets and operate on them:
+     (1) Local USB devices that are not yet in the database - we create entries for them.
+     (2) USB devices only present in the database for the local host - we delete
+     them from the database.
+  *)
   let host = Helpers.get_localhost ~__context in
   let known_pusbs_in_db =
     Db.PUSB.get_all_records ~__context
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_pusb.ml
```diff
@@ -31,10 +31,12 @@ let create ~__context ~uSB_group ~host ~other_config ~path
 
 let scan_start ~__context usbs =
   let host = Helpers.get_localhost ~__context in
-  let known_usb =
+  let local_pusb_records =
```
lindig: We now have 
* `local_pusb_records`
* `local_usb`
* `get_local_usb`
in this code and I think this is confusing. Can we find better names or avoid defining them all?

lindig: Thinking about it: maybe it's hard to avoid. I would like to suggest something else, though: `known_usb` should be become `known_local_usb`. Overall this commit is about making sure we only act on local usb devices and the error was that we operated on all usb devices in the pool.

robhoes: Maybe `known_local_usbs` then (in plural). What what does "known" actually mean? Is it that ones that are in the DB? Then how about `local_psubs_in_db`?

ocaml/xapi/xapi_pusb.ml
```diff
@@ -45,7 +47,7 @@ let scan_start ~__context usbs =
                Db.PUSB.set_USB_group ~__context ~self ~value:group
   ) (USBSet.diff local_usb known_usb);
 
-  List.filter (fun (rf, rc) -> USBSet.mem (extract_known_usb_info rc) (USBSet.diff known_usb local_usb)) (Db.PUSB.get_all_records ~__context)
+  List.filter (fun (rf, rc) -> USBSet.mem (extract_known_usb_info rc) (USBSet.diff known_usb local_usb)) local_pusb_records
```
robhoes: Why do we need to go through the whole list of DB records again? We need to destroy all PUSBs that are in the set difference `known_usb \ local_usb`, or "PUSBs that are in the DB, but not found in the host". The purpose of the `List.filter` seems to find the DB refs again that were thrown away above. Can we just not do that?

More importantly, I don't see how this changes anything, because PUSBs from other hosts would not have been in `known_usb \ local_usb` anyway?

ocaml/xapi/xapi_pusb.ml
taoyongd: Yes , "known" means in DB

robhoes: The names of the (existing) functions `get_known_usb` and `get_local_usb` are a bit misleading as well, since they take a list, then map it to some other data structure, and turn the outcome into a set.

ocaml/xapi/xapi_pusb.ml
taoyongd: yes, known_usb\ local_usb refer to local PUSB.  This change will indeed fix the issue. As if when use whole PUSB DB records, two exactly same PUSB records in two hosts will be filtered from this. As "rc" here will happen twice because of the same PUSBs

ocaml/xapi/xapi_pusb.ml
taoyongd: get_local_usb actually takes a json format data and get_known_usb takes a list

robhoes: Even more confusing then 😂

taoyongd: confusing more... :)

taoyongd: I think change konw_usb to known_usb_set and local_usb to local_usb_set will make this less confuse

taoyongd: local_pusb_records should be known_pusbs_in_db. How about these names?

robhoes: I think "local" is the one thing that _is_ clear, actually :) We always use that for "local to the host".

taoyongd: @robhoes  @lindig , please help to review the updates. Not sure now the name still confusing.

lindig:
I think this is fine for now. To make it much clearer would require to rewrite this have a more unified representation of USB devices coming from the local scan the xapi database. Maybe some comment would help:
```
We compute two sets and operate on them: 
(1) Local USB devices that are not yet in the database - we create entries for them. 
(2) USB devices only present in the database for the local host - we delete 
    them from the database.
```

taoyongd:
Ok, @lindig. Some comments will help. Thanks. I will add it.

PR #3342, opened 2017-11-17
CA-270463 block VM suspend if vGPU driver doesn't support it

lindig:
Suspending a VM must be blocked if the vGPU driver does not support it. This PR extends xapi_vm_lifecycle.check_vgpu() to check for it. The check implements a heuristic: if a pGPU does not provide metadata for compatibility checks (that are required when resuming the VM on a host), we assume suspending is not supported.

This PR refactors the existing code to use `List.for_all` for checking that all vGPUs have the necessary properties.

This has been manually tested: a VM using a vGPU driver that does not provide compatibility metadata can't be suspended because it is not an allowed operation.

### DIFF
--- a/ocaml/xapi/xapi_vm_lifecycle.ml
+++ b/ocaml/xapi/xapi_vm_lifecycle.ml
@@ -246,18 +246,27 @@ let check_vgpu ~__context ~op ~ref_str ~vgpus =
       List.assoc "restrict_vgpu_migration" restrictions = "false"
     with Not_found -> false
   in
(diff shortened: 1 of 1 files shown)


robhoes:
All good! But what's up with Travis?

lindig:
```
Error: External library "x509" not found.
-> required by "ocaml/xapi/jbuild (context default)"
Hint: try: jbuilder external-lib-deps --missing @install
```

This looks like a fallout from recent changes on master.

PR #3344, opened 2017-11-21
Bumped the API version to 2.8.

kc284:
Signed-off-by: Konstantina Chremmou <konstantina.chremmou@citrix.com>

### DIFF
--- a/ocaml/idl/datamodel_types.ml
+++ b/ocaml/idl/datamodel_types.ml
@@ -180,10 +180,9 @@ let release_order_full = [{
       branding   = "XenServer 7.1";
     }; {
       code_name     = Some rel_honolulu;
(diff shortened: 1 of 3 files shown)


ocaml/idl/datamodel_types.ml
```diff
@@ -180,7 +180,6 @@ let release_order_full = [{
       branding   = "XenServer 7.1";
     }; {
       code_name     = Some rel_honolulu;
-      (** TODO replace with the actual version numbers when Honolulu is released *)
       version_major = 2;
       version_minor = 6;
       branding   = "Unreleased";
```
gaborigloi: Should we update this too at some point?

gaborigloi: Maybe "XenServer 7.1 CU1"?

PR #3347, opened 2017-11-23
CA-271874: Local-to-local storage migration using CLI, without specif…

YarsinCitrix:
CA-271874: Local-to-local storage migration using CLI, without specifying VDI-map, fails badly

The issue is regarding the way the default SR is chosen, when no VDI mapping is provided. The current behavior is the default SR of the destine pool will be chosen. However in case of none shared SR in pool and the SR is not on the destine host, there will be problem.

The fix will change the strategy to choose an SR, in that SRs will be chosing based on the PBD the destine host is attached, shared preferred, excluding iso type.

### DIFF
--- a/ocaml/xapi/cli_operations.ml
+++ b/ocaml/xapi/cli_operations.ml
@@ -2843,15 +2843,34 @@ let vm_migrate printer rpc session_id params =
              let gpu_group = Client.GPU_group.get_by_uuid remote_rpc remote_session gpu_group_uuid in
              vgpu,gpu_group) (read_map_params "vgpu" params) in
 
-         let default_sr =
-           try let pools = Client.Pool.get_all remote_rpc remote_session in
-             printer (Cli_printer.PMsg "Selecting remote pool's default SR for migrating VDIs") ;
-             Some (Client.Pool.get_default_SR remote_rpc remote_session (List.hd pools))
+         let preferred_sr =
+           try
+             let query = Printf.sprintf "(field \"host\"=\"%s\") and (field \"currently_attached\"=\"true\")" (Ref.string_of host) in
+             let host_pbds = Client.PBD.get_all_records_where remote_rpc remote_session query in
+             let srs = List.map (fun (pbd_ref, pbd_rec) -> sr_record remote_rpc remote_session pbd_rec.API.pBD_SR) host_pbds in
+             let (sr, free_space) = List.fold_left (fun (sr, free_space) sr' ->
+               let sr_rec' = sr'.record () in
+               if sr_rec'.API.sR_content_type = "iso" then (sr, free_space)
+               else
+                 let free_space' = Int64.sub sr_rec'.API.sR_physical_size sr_rec'.API.sR_physical_utilisation in
+                 match sr with
+                 | None -> (Some sr', free_space')
+                 | Some sr ->
+                   let sr_rec = sr.record () in
+                   match sr_rec.API.sR_shared, sr_rec'.API.sR_shared with
+                   | true, false -> (Some sr, free_space)
+                   | false, true -> (Some sr', free_space')
+                   | _ -> if (free_space' > free_space) then (Some sr', free_space')
+                          else (Some sr, free_space)
+               ) (None, Int64.zero) srs in
+             match sr with
+             | Some sr -> Some (sr.getref ())
+             | _ -> None
            with _ -> None in
 
-         let vdi_map = match default_sr with
+         let vdi_map = match preferred_sr with
            | None -> vdi_map
-           | Some default_sr ->
+           | Some preferred_sr ->
              let vms = select_vms ~include_template_vms:true rpc session_id params
                  ( "host" :: "host-uuid" :: "host-name" :: "live" :: "force" :: "copy"
                    :: vm_migrate_sxm_params ) in
@@ -2864,7 +2883,7 @@ let vm_migrate printer rpc session_id params =
              let overrides = List.map (fun vdi ->
                  if List.mem_assoc vdi vdi_map
                  then (vdi, List.assoc vdi vdi_map)
-                 else (vdi, default_sr)
+                 else (vdi, preferred_sr)
                ) vdis in
              let filtered_orig_list = List.filter (fun (vdi,_) ->
                  not (List.mem_assoc vdi overrides)) vdi_map in
--- a/ocaml/xapi/cli_operations.ml
+++ b/ocaml/xapi/cli_operations.ml
@@ -2844,19 +2844,27 @@ let vm_migrate printer rpc session_id params =
              vgpu,gpu_group) (read_map_params "vgpu" params) in
 
          let preferred_sr =
+           (* The preferred SR is determined to be as the SR that the destine host has a PDB attached to it,
+              and among the choices of that the shared is preferred first(as it is recommended to have shared storage
+              in pool to host VMs), and then the one with the maximum available space *)
            try
-             let query = Printf.sprintf "(field \"host\"=\"%s\") and (field \"currently_attached\"=\"true\")" (Ref.string_of host) in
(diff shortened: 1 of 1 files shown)


ocaml/xapi/cli_operations.ml
```diff
@@ -2843,15 +2843,34 @@ let vm_migrate printer rpc session_id params =
...
+           try
+             let query = Printf.sprintf "(field \"host\"=\"%s\") and (field \"currently_attached\"=\"true\")" (Ref.string_of host) in
+             let host_pbds = Client.PBD.get_all_records_where remote_rpc remote_session query in
+             let srs = List.map (fun (pbd_ref, pbd_rec) -> sr_record remote_rpc remote_session pbd_rec.API.pBD_SR) host_pbds in
+             let (sr, free_space) = List.fold_left (fun (sr, free_space) sr' ->
+               let sr_rec' = sr'.record () in
+               if sr_rec'.API.sR_content_type = "iso" then (sr, free_space)
+               else
+                 let free_space' = Int64.sub sr_rec'.API.sR_physical_size sr_rec'.API.sR_physical_utilisation in
+                 match sr with
```
edwintorok: I think this would be a little clearer to read if you would split calculating the free space of an SR and finding the maximum into two separate passes through the srs list (although very slightly less efficient): first you'd `List.map` and calculate the free space like you do here (considering "iso" as `0L`) and return tuples of `(sr_shared, free_space)`.
Then as a second pass you would calculate the maximum with a fold.

ocaml/xapi/cli_operations.ml
```diff
@@ -2843,15 +2843,34 @@ let vm_migrate printer rpc session_id params =
              let gpu_group = Client.GPU_group.get_by_uuid remote_rpc remote_session gpu_group_uuid in
              vgpu,gpu_group) (read_map_params "vgpu" params) in
 
-         let default_sr =
-           try let pools = Client.Pool.get_all remote_rpc remote_session in
-             printer (Cli_printer.PMsg "Selecting remote pool's default SR for migrating VDIs") ;
-             Some (Client.Pool.get_default_SR remote_rpc remote_session (List.hd pools))
+         let preferred_sr =
+           try
+             let query = Printf.sprintf "(field \"host\"=\"%s\") and (field \"currently_attached\"=\"true\")" (Ref.string_of host) in
```
edwintorok: If you want to avoid escaping inside the string you can use `{|(field "host"...|}` (see https://caml.inria.fr/pub/docs/manual-ocaml/extn.html#sec263, this is a new but useful language feature in OCaml 4.02), but is fine as it is too.

ocaml/xapi/cli_operations.ml
```diff
@@ -2843,15 +2843,34 @@ let vm_migrate printer rpc session_id params =
...
+             let host_pbds = Client.PBD.get_all_records_where remote_rpc remote_session query in
+             let srs = List.map (fun (pbd_ref, pbd_rec) -> sr_record remote_rpc remote_session pbd_rec.API.pBD_SR) host_pbds in
```
robhoes: The use of the `sr_record` function (from records.ml) is a little confusing here, and also does a lot more work than you actually need. Just use `Client.SR.get_record`.

ocaml/xapi/cli_operations.ml
robhoes: I agree with @edwintorok. It would also be useful to describe the search policy in a comment, because it is not so obvious what is happening with the `shared` field.

YarsinCitrix: I have a slightly different opinion regarding this readability vs efficiency. In this case, the computing of the free_space is actually just one line of code. So pulling out of such logic out from the current fold_left with another fold_left will only slightly enhance the readability of the code in the current loop, but added complexity in the overall code. 

For the concern of the readability, I added some comments in the code, that first introduce the overall principle of the algorithm to determine the SR, and then explained the ISO, shared and available space consideration in details. Please check to see if you are fine with it in the later commit.

ocaml/xapi/cli_operations.ml
```diff
@@ -2843,15 +2843,42 @@ let vm_migrate printer rpc session_id params =
...
+             let query = Printf.sprintf {|(field "host"="%s") and (field "currently_attached"="true")|} (Ref.string_of host) in
+             let host_pbds = Client.PBD.get_all_records_where remote_rpc remote_session query in
+             let srs = List.map (fun (pbd_ref, pbd_rec) -> 
+               pbd_rec.API.pBD_SR, Client.SR.get_record remote_rpc remote_session pbd_rec.API.pBD_SR) host_pbds in
+             (* In the following loop, the current SR:sr' will be compared with previous checked ones, 
+                first if it is an ISO type, then pass this one for selection, then the only shared one from this and
+                prevoius one will be valued, and if not that case (both shared or none shared), choose the one with
+                more space available *)
+             let (sr, free_space) = List.fold_left (fun (sr, free_space) sr' ->
+               let sr_rec' = snd sr' in
```
robhoes: Instead of this, just do a more finegrained match in the function arguments, e.g.:

    List.fold_left (fun (sr, free_space) ((_, sr_rec') as sr') ->

ocaml/xapi/cli_operations.ml
```diff
@@ -2843,15 +2843,42 @@ let vm_migrate printer rpc session_id params =
...
+               if sr_rec'.API.sR_content_type = "iso" then (sr, free_space)
+               else
+                 let free_space' = Int64.sub sr_rec'.API.sR_physical_size sr_rec'.API.sR_physical_utilisation in
+                 match sr with
+                 | None -> (Some sr', free_space')
+                 | Some sr ->
+                   let sr_rec = snd sr in
```
robhoes: Here as well: `| Some ((_, sr_rec) as sr) ->`.

ocaml/xapi/cli_operations.ml
```diff
@@ -2843,15 +2843,42 @@ let vm_migrate printer rpc session_id params =
...
+                   match sr_rec.API.sR_shared, sr_rec'.API.sR_shared with
+                   | true, false -> (Some sr, free_space)
+                   | false, true -> (Some sr', free_space')
+                   | _ -> if (free_space' > free_space) then (Some sr', free_space')
+                          else (Some sr, free_space)
+               ) (None, Int64.zero) srs in
+             match sr with
+             | Some sr -> Some (fst sr)
```
robhoes: | Some (sr, _) -> sr

ocaml/xapi/cli_operations.ml
robhoes: Actually, the trick of considering the ISO as having zero free-space is not good enough: imagine that the ISO SR is the only shared SR reachable from the destination, then the policy will always select it because shared SRs are preferred. So we must filter out ISOs early. I think Yarsin's structure is fine now.

edwintorok: The comment looks good, thanks.

ocaml/xapi/cli_operations.ml
```diff
@@ -2843,15 +2843,42 @@ let vm_migrate printer rpc session_id params =
...
+              and among the choices of that the shared is preferred first(as it is recommended to have shared storage
+              in pool to host VMs), and then the one with the maximum available space *)
+           try
+             let query = Printf.sprintf {|(field "host"="%s") and (field "currently_attached"="true")|} (Ref.string_of host) in
+             let host_pbds = Client.PBD.get_all_records_where remote_rpc remote_session query in
+             let srs = List.map (fun (pbd_ref, pbd_rec) -> 
+               pbd_rec.API.pBD_SR, Client.SR.get_record remote_rpc remote_session pbd_rec.API.pBD_SR) host_pbds in
+             (* In the following loop, the current SR:sr' will be compared with previous checked ones, 
+                first if it is an ISO type, then pass this one for selection, then the only shared one from this and
+                prevoius one will be valued, and if not that case (both shared or none shared), choose the one with
```
edwintorok: small typo: `s/prevoius/previous/`

PR #3348, opened 2017-11-23
CA-273306: assert the correct VM power_state during VM.checkpoint

mg12:
The `xapi_vgpu.create` function expects that the passed vm record is always
    in the Halted power_state. However, during a VM.checkpoint, the power_state
    of the VM is Suspended and causes the internal Halted assertion to fail.
 
    This commit creates an internal `create'` function that parameterizes the
    expected power_state. This parameter is then used by the existing functions:
    * The original `xapi_vgpu.create` function behaves exactly as before, and it
    continues to "expect" the Halted power_state at all times.
    * The `xapi_vgpu.copy` function, called during VM snapshot, checkpoint and
    clone, expects the power_state of the VM associated with the VGPU, which
    conforms with the following successfully tested values:
 
    | Operation                 | expected VM power_state during VGPU.copy |
    +---------------------------+------------------------------------------+
    | VM.checkpoint             | Suspended                                |
    | VM.snapshot               | Halted                                   |
    | VM.revert_from_checkpoint | Halted                                   |
    | VM.clone                  | Halted                                   |

### DIFF
--- a/ocaml/xapi/xapi_vgpu.ml
+++ b/ocaml/xapi/xapi_vgpu.ml
@@ -21,12 +21,12 @@ let m = Mutex.create ()
 let valid_device device =
   device = "0"
 
-let create ~__context  ~vM ~gPU_group ~device ~other_config ~_type =
+let create' ~__context  ~vM ~gPU_group ~device ~other_config ~_type ~expected_vm_state =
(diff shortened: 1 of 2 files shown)


mg12:
pr to the wrong branch. closing.

PR #3349, opened 2017-11-23
CA-273306: assert the correct VM power_state during VM.checkpoint

mg12:
The `xapi_vgpu.create` function expects that the passed vm record is always                                                                                           
    in the Halted power_state. However, during a VM.checkpoint, the power_state                                                                                           
    of the VM is Suspended and causes the internal Halted assertion to fail.                                                                                           
                                                                                                                                
    This occurs because during a VM.{checkpoint,snapshot,clone} operation, a                                                                                           
    VGPU record is implicitly created in the xapi db when it is copied from the                                                                                           
    original VM. In VM.checkpoint specifically, this copy occurs while the VM is                                                                                           
    in the Suspended power_state. Measured power_states during tests:                                                                                           
                                                                                                   
    | Operation                 | measured VM power_state during VGPU.copy |                                                                                           
    +---------------------------+------------------------------------------+                                                                                           
    | VM.checkpoint             | Suspended                                |                                                                                           
    | VM.snapshot               | Halted                                   |                                                                                           
    | VM.revert_from_checkpoint | Halted                                   |                                                                                           
    | VM.clone/copy             | Halted                                   |                                                                                           
                                                                                                                                        
    There's no special reason to restrict the VGPU copy operation only to the                                                                                           
    Halted power_state. Other power_states, such as Suspended, are acceptable.                                                                                           
                                                                                                  
    The original `xapi_vgpu.create` function signature cannot be changed because                                                                                           
    it is autogenerated from the xapi datamodel, so this commit creates an                                                                                           
    internal `create'` function that accepts a boolean determining if the original                                                                                           
    power_state check for Halted should be executed or not. This parameter is                                                                                           
    then used by the existing datamodel-derived functions:                                                                                           
    * `xapi_vgpu.create` function behaves exactly as before, and it continues to                                                                                           
    assert the Halted power_state at all times.                                                                                           
    * The `xapi_vgpu.copy` function, called during VM snapshot, checkpoint and                                                                                           
    clone, does not execute the assertion anymore and therefore fixes the issue.                                                                                           
                                                                                             
    A priori, fixing the issue could be done by just removing the assertion in the                                                                                           
    `xapi_vgpu.create` function. The only side effect (apart from fixing the issue)                                                                                           
    is that it would be possible to use the XenAPI directly to call VGPU.create at                                                                                           
    any time -- `xe vgpu-create` would not enforce the VM to be in the Halted                                                                                           
    state anymore. This would only affect the Xapi db, creating a slightly                                                                                           
    inconsistent view of the properties of the VM until the VGPU is actually                                                                                           
    attached to the VM in the next VM.start.                                                                                           
                                                                                                                               
    However, in this case XenCenter will show this inconsistent view of the running                                                                                           
    VM, which will not have a VGPU attached but may show one in its 'properties',                                                                                           
    which could cause some confusion to the user, so it's probably better to keep                                                                                           
    the create function (as accessed from the XenAPI) protected by the assertion.                                                                                           
                                                                                                                      
    Therefore, after this change, `xe vgpu-create` will still return an error if                                                                                           
    the VM is not Halted:                                                                                           
                                                                                                                     
      # xe vgpu-create ...                                                                                           
      You attempted an operation on a VM that was not in an appropriate power state                                                                                           
      expected: halted                                                                                             
      actual: running


ocaml/xapi/xapi_vgpu.ml
```diff
@@ -21,12 +21,12 @@ let m = Mutex.create ()
 let valid_device device =
   device = "0"
 
-let create ~__context  ~vM ~gPU_group ~device ~other_config ~_type =
+let create' ~__context  ~vM ~gPU_group ~device ~other_config ~_type ~expected_vm_state =
```
mseri: Why not
 
```ocaml
let create ~__context  ~vM ~gPU_group ~device ~other_config ~_type ?(expected_vm_state = `Halted)
```
instead?

ocaml/xapi/xapi_vgpu.ml
```diff
@@ -66,12 +68,13 @@ let atomic_set_resident_on ~__context ~self ~value = assert false
 
 let copy ~__context ~vm vgpu =
   let all = Db.VGPU.get_record ~__context ~self:vgpu in
-  let vgpu = create ~__context
+  let vgpu = create' ~__context
```
mseri: I would add a comment explaining that here you expect only
```
| Operation                 | expected VM power_state during VGPU.copy |
+---------------------------+------------------------------------------+
| VM.checkpoint             | Suspended                                |
| VM.snapshot               | Halted                                   |
| VM.revert_from_checkpoint | Halted                                   |
| VM.clone                  | Halted                                   |
```

Why not making the expected power state explicit here in copy as well?

ocaml/xapi/xapi_vgpu.ml
mg12: This had caused an issue in the automatically-generated code in api_server.ml:
```
(** Actions module *)
module Actions = struct
  ...
  module VGPU = Xapi_vgpu
end
```
that fails to compile with
```
File "ocaml/xapi/api_server.ml", line 100, characters 47-54:
Error: Signature mismatch:
       ...
       In module VGPU:
       Values do not match:
         val create :
           __context:Context.t ->
           vM:[ `VM ] Ref.t ->
           gPU_group:[ `GPU_group ] Ref.t ->
           device:string ->
           other_config:(string * string) list ->
           _type:[ `VGPU_type ] Ref.t ->
           ?expected_vm_state:[< `Halted
                               | `Migrating
                               | `Paused
                               | `Running
                               | `ShuttingDown
                               | `Suspended
                               > `Halted `Paused `Running `Suspended ] ->
           unit -> [ `VGPU ] Ref.t
       is not included in
         val create :
           __context:Context.t ->
           vM:API.ref_VM ->
           gPU_group:API.ref_GPU_group ->
           device:string ->
           other_config:API.string_to_string_map ->
           _type:API.ref_VGPU_type -> API.ref_VGPU
```
the vgpu object in the datamodel is defined with
```
let vgpu =
  let create = call
      ~name:"create"
      ~versioned_params:[
        {param_type=(Ref _vm); param_name="VM"; param_doc=""; param_release=boston_release; param_default=None};
        {param_type=(Ref _gpu_group); param_name="GPU_group"; param_doc=""; param_release=boston_release; param_default=None};
        {param_type=String; param_name="device"; param_doc=""; param_release=boston_release; param_default=Some (VString "0")};
        {param_type=(Map (String, String)); param_name="other_config"; param_doc=""; param_release=boston_release; param_default=Some (VMap [])};
        {param_type=(Ref _vgpu_type); param_name="type"; param_doc=""; param_release=vgpu_tech_preview_release; param_default=(Some (VRef null_ref))};
      ]
```
I thought that would be better not to change the datamodel to match a new signature for `xapi_vgpu.create()`. -- instead, I kept the signature of `xapi_vgpu.create()` and used an internal function instead to handle the extra parameter.

ocaml/xapi/xapi_vgpu.ml
mg12: yes, that would be good, I'll update the comment

ocaml/xapi/xapi_vgpu.ml
mseri: Argh, yes if it has to play with the autogenerated code, let's go with the initial version. Please add also a comment on `create'` that explains why we need to duplicate it in such way.

mseri:
Could you figure out what was the reason to always assume the VM as Halted?

robhoes:
> Could you figure out what was the reason to always assume the VM as Halted?

I don't think that there is actually a need for this assertion at all. It should be fine to create a VGPU in the DB at any time.

robhoes:
This came from 91cb0ca7c195d2229a792ea7f9b9637e901fcaa3, without too much information about the reasons...

ocaml/xapi/xapi_vgpu.ml
```diff
@@ -66,12 +68,13 @@ let atomic_set_resident_on ~__context ~self ~value = assert false
...
       ~device:all.API.vGPU_device
       ~gPU_group:all.API.vGPU_GPU_group
       ~vM:vm
       ~other_config:all.API.vGPU_other_config
       ~_type:all.API.vGPU_type
+      ~expected_vm_state:(vm |> fun self -> Db.VM.get_power_state ~__context ~self)
```
robhoes: This is a little odd though: it essentially just bypasses the power-state check. How about having a parameter for just that instead? E.g. `~no_powerstate_check`.

ocaml/xapi/xapi_vgpu.ml
mg12: As robhoes suggested, it's probably simpler to just have a boolean flag to indicate that the powerstate should not be checked during copy, so I think there's no need to check states in copy -- it should just copy regardless of the vm state.

ocaml/xapi/xapi_vgpu.ml
mg12: done: df49140bfdc1f12e8b344e8f1be3fa00c58f8e33

mg12:
updated the commit with the suggestions, and added some explanation in the commit header about why it seems better to keep the Halted assertion for the create when not called from copy.

ocaml/xapi/xapi_vgpu.ml
mg12: done: df49140bfdc1f12e8b344e8f1be3fa00c58f8e33

ocaml/xapi/xapi_vgpu.ml
mseri: True. I agree

robhoes:
I like patches where the commit message is longer than the patch to the code itself 👍

PR #3350, opened 2017-11-23
WIP CP-24185 - support for VM.suspend/resume and vGPUs

lindig:
This is work in progress to add support for VM.suspend/VM.resume with vGPUs that support migration. Do not merge.

### DIFF
--- a/ocaml/xapi/xapi_gpumon.ml
+++ b/ocaml/xapi/xapi_gpumon.ml
@@ -49,17 +49,59 @@ module Nvidia = struct
       |> Stdext.Base64.encode
     in [key, metadata]
 
+  let vgpu_impl ~__context vgpu =
+    vgpu
+    |> (fun self -> Db.VGPU.get_type ~__context ~self)
+    |> (fun self -> Db.VGPU_type.get_implementation ~__context ~self)
+
+  (** [is_nvdia] is true, if [vgpu] is an NVDIA vGPU *)
+  let is_nvidia ~__context ~vgpu =
+    vgpu_impl ~__context vgpu = `nvidia
+
+  (** [reason_to_string] turns an incompatibility reason into a string
+   * for reporting it in an error message *)
+  let reason_to_string = function
+    | Gpumon_interface.Host_driver  -> "host-driver"
+    | Gpumon_interface.Guest_driver -> "guest-driver"
+    | Gpumon_interface.GPU          -> "gpu"
+    | Gpumon_interface.Other        -> "other"
+
+  (** [get_vgpu_metadata] relies on the assumption that there is at most
+   * one vGPU per VM. The underdyling problem is that there is no
+   * mapping between a vGPU and vgpu_instance in the NVIDIA library
+   * currently.
+   *)
+  let get_vgpu_compatibility_metadata ~__context ~vgpu =
+    let this = "get_vgpu_compatibility_metadata" in
+    try
+      let dbg   = Context.string_of_task __context in
+      let vm    = Db.VGPU.get_VM ~__context ~self:vgpu in
+      let domid = Db.VM.get_domid ~__context ~self:vm |> Int64.to_int in
+      Db.VGPU.get_resident_on ~__context ~self:vgpu
+      |> fun self -> Db.PGPU.get_PCI ~__context ~self
+      |> fun self -> Db.PCI.get_pci_id ~__context ~self
+      |> Gpumon_client.Client.Nvidia.get_vgpu_metadata dbg domid
+      |> function
+        | []      -> []
+        | [meta]  -> [key, Stdext.Base64.encode meta]
+        | _::_    -> failwith @@ Printf.sprintf
+          "%s: VM %s (dom %d) has more than one NVIDIA vGPU (%s)"
+          this (Ref.string_of vm) domid __LOC__
+    with
+      | Gpumon_interface.NvmlInterfaceNotAvailable ->
+        let host = Helpers.get_localhost ~__context |> Ref.string_of in
+        raise Api_errors.(Server_error (nvidia_tools_error, [host]))
+      | err ->
+        let msg = Printexc.to_string err in
+        raise Api_errors.(Server_error (internal_error, [msg]))
+
+
   (* N.B. the vgpu (and the vm) must be in the local host where this function runs *)
   let assert_pgpu_is_compatible_with_vm ~__context ~vm ~vgpu ~dest_host ~encoded_pgpu_metadata =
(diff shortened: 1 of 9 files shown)


ocaml/xapi/xapi_vm.ml
```diff
@@ -422,6 +422,15 @@ let power_state_reset ~__context ~vm =
 let suspend ~__context ~vm =
   Db.VM.set_ha_always_run ~__context ~self:vm ~value:false;
   debug "Setting ha_always_run on vm=%s as false during VM.suspend" (Ref.string_of vm);
+  (* set compatibility meta data for all NVIDIA vGPUs *)
```
mseri: I'd make this an helper in the `Xapi_gpumon` module as it is for `assert_pgpu_is_compatible_with_vm` and expose as little logic as we can outside that module

ocaml/xapi/xapi_gpumon.ml
```diff
@@ -130,6 +130,59 @@ module Nvidia = struct
...
+            vgpu_destination_incompatible,
+            [ String.concat ", " (List.map stringify reasons)
+            ; Ref.string_of vgpu
+            ; localhost'
+            ]))
+      | exception Not_found ->
+        raise Api_errors.(Server_error
+         (internal_error,
+          [ "can't obtain vGPU/pGPU compatibility meta data"
+          ; Ref.string_of vgpu
```
mseri: Whe should probably add also the `pgpu` here

ocaml/xapi/xapi_gpumon.ml
```diff
@@ -130,6 +130,59 @@ module Nvidia = struct
...
+          ]))
+      | exception Gpumon_interface.NvmlInterfaceNotAvailable ->
+        raise Api_errors.(Server_error (nvidia_tools_error, [localhost']))
+      | exception err ->
+        raise Api_errors.(Server_error
+          (internal_error, [Printexc.to_string err]))
+    in
+      if is_nvidia ~__context ~vgpu then check vgpu else ()
```
mseri: Maybe add a log in the `else` case, just to be sure...

ocaml/xapi/xapi_gpumon.ml
```diff
@@ -130,6 +130,59 @@ module Nvidia = struct
...
 
+  (** [assert_vgpu_pgpu_are_compatible] checks that vGPU and pGPU are
+   * comatible according to their abstract compatibility meta data. This
+   * needs to run on the host of the pGPU to be meaningful.
+   *)
+  let assert_vgpu_pgpu_are_compatible ~__context ~vgpu ~pgpu =
+    let dbg        = Context.string_of_task __context in
+    let localhost  = Helpers.get_localhost ~__context in
+    let localhost' = Ref.string_of localhost in
+    let stringify = function
```
mseri: let's factor this out. It's in both checks (here and above)

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -338,8 +338,14 @@ let assert_gpus_available ~__context ~self ~host =
   let reqd_groups =
     List.map (fun self -> Db.VGPU.get_GPU_group ~__context ~self) vgpus in
   let is_pgpu_available pgpu vgpu =
-    try Xapi_pgpu.assert_can_run_VGPU ~__context ~self:pgpu ~vgpu; true
-    with _ -> false
+    try
+      Xapi_pgpu.assert_can_run_VGPU ~__context ~self:pgpu ~vgpu;
+      Xapi_gpumon.Nvidia.assert_vgpu_pgpu_are_compatible
```
mseri: I think we need to make sure that we are calling this only for resume operations and that the metadata is in the db. I'd like to know from @robhoes if assert_gpus_available (called by assert_can_boot_here) is the right place, or where it should be called instead. I remember for the migration check, we had decided not to do it in such assert for a reason (but I don't remember the reason)

lindig: I have addressed the comments about factoring out stringify and reporting the pGPU in case no metadata can be found.

The problem, from where actually invoking the compatibility check is not yet solved.

robhoes: Most important is that this new check is only done for resuming indeed, because assert_can_boot_here is also called for starting VMs and other operations. If we make sure that the VGPU metadata field is always empty if the VM is not suspended, and that this check then passes, then it is probably fine.

ocaml/xapi/xapi_gpumon.ml
```diff
@@ -133,57 +133,62 @@ module Nvidia = struct
...
-          |> List.assoc key
-          |> Stdext.Base64.decode in
+        Db.PGPU.get_compatibility_metadata ~__context ~self:pgpu
+        |> List.assoc key
+        |> Stdext.Base64.decode in
       let vgpu_metadata () =
         Db.VGPU.get_compatibility_metadata ~__context ~self:vgpu
         |> fun md -> [List.assoc key md]
-        |> List.map Stdext.Base64.decode in
+                     |> List.map Stdext.Base64.decode in
```
mseri: ?

lindig:
I have fixed the indentation last pointed out by @mseri but also re-factored my commits such they make more sense in sequence.

ocaml/xapi/xapi_gpumon.ml
```diff
@@ -84,17 +126,94 @@ module Nvidia = struct
...
             vgpu_destination_incompatible,
-            [ String.concat ", " (List.map stringify reasons) (* There could be multiple reasons *)
+            [ String.concat ", " (List.map reason_to_string reasons)
+            (* There could be multiple reasons *)
             ; Ref.string_of vgpu
             ; Ref.string_of dest_host
             ]))
 
+  (** Predicate [vgpu_pgpu_are_compatible] checks that vGPU and pGPU are
+   * comatible according to their abstract compatibility metadata. This
```
mseri: compatible*

ocaml/xapi/xapi_xenops.ml
```diff
@@ -2819,6 +2819,7 @@ let resume ~__context ~self ~start_paused ~force =
        end;
        set_resident_on ~__context ~self;
        Db.VM.set_suspend_VDI ~__context ~self ~value:Ref.null;
+       Xapi_gpumon.clear_vgpu_metadata ~__context ~vm:self;
```
mseri: Add a comment here saying that this have to happen later in the process to avoid issues

lindig:
I'm testing this right now. Tests were failing because of a problem in the `vgpu` binary that is part of Xen.

PR #3351, opened 2017-11-27
Xapi coverage

edwintorok:
Depends on https://github.com/xapi-project/xcp-idl/pull/179

### DIFF
--- a/.travis-opam-coverage.sh
+++ b/.travis-opam-coverage.sh
@@ -1,8 +1,9 @@
+#!/bin/bash
 # SUMMARY:
 # Builds & tests xapi with coverage in a Ubuntu 16.04 Docker container with
 # OCaml 4.04.2, then uploads the coverage information to coveralls.
 
-set -uex
+set -ex
 
 # Currently there is no way of specifying OPAM depexts for multiple versions of
 # a given disto, and our current depexts only work with Ubuntu >= 16.04, due to
@@ -15,7 +16,7 @@ docker run --rm --volume=$PWD:/mnt --workdir=/mnt \
   --env "TRAVIS=$TRAVIS" \
   --env "TRAVIS_JOB_ID=$TRAVIS_JOB_ID" \
   ocaml/opam:ubuntu-16.04_ocaml-4.04.2 \
-  bash -uex -c '
+  bash -ex -c '
 sudo apt-get update
 
 sudo chown -R $(whoami) .
@@ -25,13 +26,39 @@ opam repository remove default
 opam repository add xs-opam https://github.com/xapi-project/xs-opam.git
 
 # install the dependencies of xapi
+sudo apt install -y libpci-dev
 opam pin add --no-action xapi .
 opam depext --yes xapi
 opam install --deps-only xapi
 
 # build and test xapi with coverage, then submit the coverage information to coveralls
-sudo apt-get install --yes wget
-wget https://raw.githubusercontent.com/simonjbeaumont/ocaml-travis-coveralls/master/travis-coveralls.sh
-COV_CONF="./configure" bash -ex travis-coveralls.sh
+
+opam install -y bisect_ppx ocveralls
+
+export BISECT_ENABLE=YES
+
+ulimit -s 16384
+
+# the test fails when run on debian-stable (it passed on Ubuntu though)
+# disable it when running coverage
+# -p does not work to filter out subdirectories though
+# because the test in pci does not say which package it is part of
+rm -rf ocaml/pci/lib_test
+jbuilder runtest
+
+declare -a outs
+outs=($(find . | grep bisect.*.out))
+bisect-ppx-report -I $(dirname ${outs[1]}) -text report ${outs[@]}
+bisect-ppx-report -I $(dirname ${outs[1]}) -summary-only -text summary ${outs[@]}
+if [ -n "$HTML" ]; then bisect-ppx-report -I $(dirname ${outs[1]}) -html ../html-report ${outs[@]}; fi
+
+if [ -n "$TRAVIS" ]; then
+  echo "\$TRAVIS set; running ocveralls and sending to coveralls.io..."
+  ocveralls --prefix _build/default ${outs[@]} --send
+else
+  echo "\$TRAVIS not set; displaying results of bisect-report..."
+  cat report
+  cat summary
+fi
 '
 
--- a/ocaml/util/jbuild
+++ b/ocaml/util/jbuild
@@ -1,6 +1,17 @@
+(* -*- tuareg -*- *)
+#require "unix"
+
+let coverage_rewriter =
+  let is_coverage = try Unix.getenv "BISECT_ENABLE" = "YES" with Not_found -> false in
+  if is_coverage then
+    "(preprocess (pps (bisect_ppx -conditional)))"
+  else
+    ""
+
+let () = Printf.ksprintf Jbuild_plugin.V1.send {|
 (rule
  ((targets (build_info.ml))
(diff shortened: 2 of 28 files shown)


ocaml/xapi/xapi.ml
```diff
@@ -189,6 +189,7 @@ let init_args() =
   (* Immediately register callback functions *)
   register_callback_fns();
   Xcp_service.configure ~options:Xapi_globs.all_options ~resources:Xapi_globs.Resources.xcp_resources ();
+  Xcp_coverage.dispatcher_init "xapi";
```
thomassa: How did you decide where in the start-up sequence to insert this?
There are many places that look as if they might be reasonable, but it is quite possible that some have non-obvious drawbacks.

thomassa:
The use-more-cpus commit needs a "Signed-off-by" in the commit message.

ocaml/xapi/xapi.ml
edwintorok: The other daemons (well xenopsd for now) would use `Xcp_service.configure2` to initialize, so I chose the most similar place to that, there are indeed many equally good places for it.
I think the initialization should meet these constraints:
* we initialize before shutting down in a successful run (so the coverage data gets written in the proper place)
* we do not initialize it for the CLI (we would run out of space in the host-installer ramdisk)
* we initialize before a testcase is finished (which triggers the coverage dump call)

edwintorok:
Yes this looked much better in OASIS where we only had to touch one file, maybe we can do something with `(:include ....)` and generate the to-be-included file from `configure` that enables coverage dumping.
Upstream's solution is to just use `bisect_ppx -conditional` always, but I am a bit worried of that, I want the build system to ensure we are not accidentally rewriting with coverage, not the tool.

mseri:
For the moment we will have to live with the autogenerated stuff but I hope there will be a better solution soon

gaborigloi:
We also need to fix the Coveralls integration. It went wrong when we switched to jbuilder, to script still uses oasis.

edwintorok:
For now I'll switch those jbuild files back to using the plugin and update this PR sometime next week.

edwintorok:
Updated jbuilder files to all use the coverage rewriter and rebased on master to avoid merge conflicts in jbuild files.

ocaml/util/jbuild
```diff
@@ -1,6 +1,17 @@
...
+  if is_coverage then
+    "(preprocess (pps (bisect_ppx -conditional)))"
+  else
+    ""
+
+let () = Printf.ksprintf Jbuild_plugin.V1.send {|
 (rule
  ((targets (build_info.ml))
-  (action  (with-stdout-to ${@} (run "date" "+let date=\"%Y-%m-%d\"")))))
+  (action  (with-stdout-to ${@} (run "date" "+let date=\"%%Y-%%m-%%d\"")))))
```
mseri: Why you need to double the `%`?

edwintorok: this is part of the `Printf` above, we need the `%` to be interpreted by the `date` command, not by Printf, hence we need to escape it.

mseri: Oh right, thanks

.travis-opam-coverage.sh
```diff
@@ -25,13 +25,37 @@ opam repository remove default
...
+
+# the test fails when run on debian-stable (it passed on Ubuntu though)
+# disable it when running coverage
+# -p doesn't work to filter out subdirectories though
+# because the test in pci doesn't say which package it is part of
+rm -rf ocaml/pci/lib_test
+jbuilder runtest
+
+
+outs=$(find . | grep bisect.*.out)
```
mseri: Is it worth to update this code to prevent issues as done in:

https://github.com/xapi-project/rrd-transport/commit/93d93c147d138c1edd9c1d367ae440e9db334523

?

mseri:
I am rerunning the travis job, the coverage was failing due to the datamodel stack overflow that should now be fixed in xs-opam

.travis-opam-coverage.sh
edwintorok: Had to use slightly different syntax to get all the values from the array, `$outs` only printed one:
```
+ outs=($(find . | grep bisect.*.out))
++ find .
++ grep 'bisect.*.out'
++ dirname ./_build/default/ocaml/xapi-types/bisect0001.out
+ bisect-ppx-report -I ./_build/default/ocaml/xapi-types -text report ./_build/default/ocaml/xapi-client/bisect0001.out
++ dirname ./_build/default/ocaml/xapi-types/bisect0001.out
+ bisect-ppx-report -I ./_build/default/ocaml/xapi-types -summary-only -text summary ./_build/default/ocaml/xapi-client/bisect0001.out
+ '[' -n '' ']'
```

Using the commit I pushed:
```
+ declare -a outs
+ outs=($(find . | grep bisect.*.out))
++ find .
++ grep 'bisect.*.out'
++ dirname ./_build/default/ocaml/xapi-types/bisect0001.out
+ bisect-ppx-report -I ./_build/default/ocaml/xapi-types -text report ./_build/default/ocaml/xapi-client/bisect0001.out ./_build/default/ocaml/xapi-types/bisect0001.out ./_build/default/ocaml/xapi/bisect0004.out ./_build/default/ocaml/xapi/bisect0001.out ./_build/default/ocaml/xapi/bisect0003.out ./_build/default/ocaml/xapi/bisect0002.out ./_build/default/ocaml/xapi/bisect0005.out
++ dirname ./_build/default/ocaml/xapi-types/bisect0001.out
+ bisect-ppx-report -I ./_build/default/ocaml/xapi-types -summary-only -text summary ./_build/default/ocaml/xapi-client/bisect0001.out ./_build/default/ocaml/xapi-types/bisect0001.out ./_build/default/ocaml/xapi/bisect0004.out ./_build/default/ocaml/xapi/bisect0001.out ./_build/default/ocaml/xapi/bisect0003.out ./_build/default/ocaml/xapi/bisect0002.out ./_build/default/ocaml/xapi/bisect0005.out
+ '[' -n '' ']'
```

edwintorok:
Looks like it couldn't find `jbuilder`, doing a test locally.

.travis-opam-coverage.sh
mseri: You are right, the script was broken. I've sent a PR to rrd-transport to fix it. We should consider rewriting it in a more robust way, and (as I think you suggested) store in a repo from where all the packages that need it can directly use it instead of replicating the code

.travis-opam-coverage.sh
```diff
@@ -25,13 +26,41 @@ opam repository remove default
...
+
+declare -a outs
+outs=($(find . | grep bisect.*.out))
+bisect-ppx-report -I $(dirname ${outs[1]}) -text report ${outs[@]}
+bisect-ppx-report -I $(dirname ${outs[1]}) -summary-only -text summary ${outs[@]}
+if [ -n "$HTML" ]; then bisect-ppx-report -I $(dirname ${outs[1]}) -html ../html-report ${outs[@]}; fi
+
+if [ -n "$TRAVIS" ]; then
+  echo "\$TRAVIS set; running ocveralls and sending to coveralls.io..."
+  ocveralls --prefix _build/default $outs --send
```
mseri: Don't you need `${outs[@]}` here as well?

mseri:
If you want to move the coverage script to a new repository, say `xapi-travis-scripts`, and get it from there, I think we could take this PR as an opportunity...

mseri:
If you want to completely fix the build you need to set the HTML variable to something:
```
bash: line 34: HTML: unbound variable
```

edwintorok:
Squashed the fixups, lets wait for one more build.

PR #3353, opened 2017-11-29
CP-23025 (JsonRpc protocol doc) and CA-274994 (JsonRpc error code improvement)

kc284:


ocaml/doc/wire-protocol.md
```diff
@@ -55,7 +57,7 @@ These types are mapped onto XML-RPC types in a straight-forward manner:
   XML-RPC `<struct>`, with the key as the name of the struct.  Note that the
   `(k -> v) map` type is only valid when `k` is a `string`, `ref`, or
   `int`, and in each case the keys of the maps are stringified as
-  above. For example, the `(string -> double) map` containing the mappings
+  above. For example, the `(string -> float) map` containing the mappings
```
kc284: I think this has always being wrong because the APi has floats not doubles, so here we need what the API has, not what it is mapped to.

edwintorok: OCaml has only one floating point type with double precision, so I agree it is confusing to write `double` in an example which otherwise uses OCaml syntax with arrows.

ocaml/xapi/api_server.ml
```diff
@@ -127,6 +127,10 @@ let is_himn_req req =
      | None -> false)
   | None -> false
 
+(* The API does not use the error.code and only retains it for compliance with
+  the JSON-RPC v2.0 specs. We set this always to a non-zero value because
+  some JsonRpc clients consider error.code 0 as no error*)
+let error_code_lit = 1L
```
euanh: :+1:

ocaml/doc/wire-protocol.md
mseri: Maybe we should mention this float/double quirck. The value pased via the rpc is going to be a double

kc284: It's mentioned a bit further up that for xml-rpc the API's float is mapped to double.
Also, further down I've described that for json-rpc the float is mapped to what consists an Intlit.

gaborigloi:
The new documentation wasn't uploaded to http://xapi-project.github.io/xen-api/wire-protocol.html - is that an issue?

PR #3355, opened 2017-12-04
CA-271867: Local-to-local storage migration fails with out-of-space e…

YarsinCitrix:
…rror

In this change, the original error message will be displayed to user in case of local SR space shortage. This will be confusing as in the scenario of VM migration the "specified SR" will be more likely to be the destine SR. The change will add another error to explicitly say the local SR in the VM migration scenario whereas the snapshot failed due to space shortage.

---
This change is [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3355)


ocaml/xapi/storage_migrate.ml
```diff
@@ -542,6 +542,10 @@ let start' ~task ~dbg ~sr ~vdi ~dp ~url ~dest =
     error " Caught exception %s:%s. Performing cleanup." Api_errors.sr_not_attached sr_uuid;
     perform_cleanup_actions !on_fail;
     raise (Api_errors.Server_error(Api_errors.sr_not_attached,[sr_uuid]))
+  | Storage_interface.Backend_error(code, _) when code = "SR_BACKEND_FAILURE_44" ->
+    error "Caught exception SR_BACKEND_FAILURE_44. Performing cleanup.";
```
lindig: It would be good to add some identifiers to the log message: `sr:%s vdi:%s url:%s dest:%s`.

ocaml/idl/datamodel.ml
```diff
@@ -1080,6 +1080,8 @@ let _ =
     ~doc:"The SR has no attached PBDs" ();
   error Api_errors.sr_full ["requested";"maximum"]
     ~doc:"The SR is full. Requested new size exceeds the maximum size" ();
+  error Api_errors.sr_source_space_insufficient [ ]
+    ~doc:"The source SR has insufficient available space for temporarily usage to proceed the operation." ();
```
edwintorok: I think that instead of `temporarily` this sounds better: `doesn't have sufficient temporary space available to proceed with the operation`

ocaml/xapi/storage_migrate.ml
YarsinCitrix: It is a good suggestion, as to provide the complete data for further debugging.  I checked the function, in the beginning, there is already a line covering all the information such as sr/vdi. So I think it is enough. Duplicating those information will make the log file large and not necessary.

Previous debug information is like:
debug "Mirror.start sr:%s vdi:%s url:%s dest:%s" sr vdi url dest;

ocaml/idl/datamodel.ml
YarsinCitrix: Good suggestion.

ocaml/idl/datamodel.ml
```diff
@@ -1080,6 +1080,8 @@ let _ =
     ~doc:"The SR has no attached PBDs" ();
   error Api_errors.sr_full ["requested";"maximum"]
     ~doc:"The SR is full. Requested new size exceeds the maximum size" ();
+  error Api_errors.sr_source_space_insufficient [ ]
```
krizex: I think it's better to add `SR` as the parameter of this exception

robhoes: I agree: add a parameter for the SR ref, so that it is easy to see which SR is full (a VM may have disks on multiple SRs).

lindig:
I like log messages that include what they are about rather than relying on context. However, if @robhoes thinks this is good as it is, I'm happy to accept it, too.

robhoes:
Travis says no:
```
File "ocaml/xapi/storage_migrate.ml", line 549, characters 92-94:
Error: This expression has type Storage_interface.sr = string
       but an expression was expected of type 'a Ref.t
```

robhoes:
Also, did you test that this definitely not catches the error if the _destination_ SR is full?

ocaml/xapi/storage_migrate.ml
```diff
@@ -545,7 +545,8 @@ let start' ~task ~dbg ~sr ~vdi ~dp ~url ~dest =
   | Storage_interface.Backend_error(code, _) when code = "SR_BACKEND_FAILURE_44" ->
     error "Caught exception SR_BACKEND_FAILURE_44. Performing cleanup.";
     perform_cleanup_actions !on_fail;
-    raise (Api_errors.Server_error(Api_errors.sr_source_space_insufficient, []))
+	
+    raise (Api_errors.Server_error(Api_errors.sr_source_space_insufficient, [ Ref.string_of sr ]))
```
robhoes: Compilation fails, because `sr` isn't of type `Ref.t`.

YarsinCitrix: Rob, thank you so much for helping to check it and give comments! I will check it and make updates accordingly later, possibly after I made some progress on the xs-opam update!

ocaml/xapi/storage_migrate.ml
```diff
@@ -542,6 +551,10 @@ let start' ~task ~dbg ~sr ~vdi ~dp ~url ~dest =
     error " Caught exception %s:%s. Performing cleanup." Api_errors.sr_not_attached sr_uuid;
     perform_cleanup_actions !on_fail;
     raise (Api_errors.Server_error(Api_errors.sr_not_attached,[sr_uuid]))
+  | Api_errors.Server_error(code, param) when code = Api_errors.sr_source_space_insufficient ->
```
mseri: Why do we have a leading space in the error strings of this pattern match and the above?

ocaml/xapi/storage_migrate.ml
```diff
@@ -542,6 +551,10 @@ let start' ~task ~dbg ~sr ~vdi ~dp ~url ~dest =
...
+    error " Caught exception %s:%s, performing cleanup." Api_errors.sr_source_space_insufficient (List.hd param);
+    perform_cleanup_actions !on_fail;
+    raise (Api_errors.Server_error(Api_errors.sr_source_space_insufficient, param))
```
jonludlam: I think this clause is redundant - the only thing it does is improve the logging in the case of the new error code. I think it would be better to change the `Printexc.to_string` in the generic exception handler clause below and instead call `Api_errors.to_string`

jonludlam:
Other than the redundant error handler, this looks good. I think the change to call `receive_cancel` is correct. I assume you added this because during testing you found it wasn't cleaned up on the destination?

ocaml/xapi/storage_migrate.ml
YarsinCitrix: Good point, updated.

YarsinCitrix:
Regarding "Other than the redundant error handler, this looks good. I think the change to call receive_cancel is correct. I assume you added this because during testing you found it wasn't cleaned up on the destination?",
Answer: yes, in the mirror failure case due to local space shortage, the exception is raised and the stop is not called, which include the remote receiv_cancel to remove the vdi that is leftover.

PR #3356, opened 2017-12-06
CA-266936: Introduce the PCI library as vendored dependency

mseri:
This will allow us to have more complete debug information in case of pci-related issues.

The ocaml-pci library is released under LGPL according to the source repository (no mention of any license on opam). If this is not compatible with xapi's license feel free to reject this PR.


gaborigloi:
Github nicely shows the license conditions here: https://github.com/simonjbeaumont/ocaml-pci/blob/master/LICENSE

mseri:
Thanks. I have added back the LICENSE file and documented the CHANGES

gaborigloi:
I think the `Same license (library) ` condition means that the vendored source code must have the same license, but not the code using it (xapi)?

mseri:
I think (and hope) so

gaborigloi:
If this is a long-term thing, than we could even remove `pci` from `xs-opam`.

mseri:
I have considered it. If this is merged we should temporarily remove it from `xs-opam` to avoid unnecessary duplicates. I believe we should get it out and upstream the changes at a certain point, then reintroduce it in xs-opam.

gaborigloi:
If both this library and xapi use the same license, then it's fine :smile:

mseri:
`diff LICENSE ocaml/pci/LICENSE` is empty. This has the same exact license as xapi

PR #3358, opened 2017-12-07
CA-274957: simplify and reorganise Nvidia VGPU compatibility check

mseri:
The function `assert_destination_pgpu_is_compatible_with_vm`
had a heuristic to select the destination pgpu that was
filtering over vgpu implementation, however this is `nvidia`
for any Nvidia card, resulting in potentially meaningless
choices.

The function was also assuming that the user-defined
vgpu_map was complete (all vgpus would be mapped) but
this is often not the case.

The function was called in two migration places, in one
we had preselected the pgpu, making any heuristic
a potential source of errors.

This commit splits the check into two assertions:
- `assert_destination_pgpu_is_compatible_with_vm`, to be called
  when the destination pgpu has already been selected
- `assert_destination_has_pgpu_compatible_with_vm`, using
  a heuristic (based on the one used by vm-import) to
  select a suitable pgpu group.

The new heuristic selects a gpu group by filtering
over the gpu types, making sure that the destination
group is compatible with the origin group.

### DIFF
--- a/ocaml/xapi/xapi_pgpu_helpers.ml
+++ b/ocaml/xapi/xapi_pgpu_helpers.ml
@@ -14,7 +14,8 @@
 module D=Debug.Make(struct let name="xapi" end)
 open D
 
-open Stdext
+open Xapi_stdext_monadic
+open Xapi_stdext_std
 
 let assert_VGPU_type_supported ~__context ~self ~vgpu_type =
   let supported_VGPU_types =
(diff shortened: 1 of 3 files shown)


ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -246,8 +246,17 @@ let pool_migrate ~__context ~vm ~host ~options =
   let vm_uuid = Db.VM.get_uuid ~__context ~self:vm in
   let xenops_vgpu_map = infer_vgpu_map ~__context vm in
 
-  (* Check pGPU compatibility for Nvidia vGPUs *)
-  Xapi_pgpu_helpers.assert_destination_pgpu_is_compatible_with_vm ~__context ~vm ~host ~vgpu_map:[] ();
+  (* Check pGPU compatibility for Nvidia vGPUs - at this stage we already know
+   * the vgpu <-> pgpu mapping. *)
+  let vgpus_map = List.map
+      (fun vgpu ->
+         let pgpu = Db.VGPU.get_scheduled_to_be_resident_on ~__context ~self:vgpu in
```
lindig: Maybe just: 
```
Db.VM.get_VGPUs ~__context ~self:vm
|> List.map (fun vgpu -> (vgpu,  Db.VGPU.get_scheduled_to_be_resident_on ~__context ~self:vgpu))
|> List.iter (fun (vgpu, pgpu) ->
  Xapi_pgpu_helpers.assert_destination_pgpu_is_compatible_with_vm ~__context ~vm ~host ~vgpu ~pgpu ())
```

ocaml/xapi/xapi_pgpu_helpers.ml
```diff
@@ -136,82 +137,108 @@ let assert_capacity_exists_for_VGPU_type ~__context ~self ~vgpu_type =
...
+            assert_destination_pgpu_is_compatible_with_vm ~__context ~vm ~vgpu ~pgpu ~host ?remote ()
+        end
+    ) vgpu_map;
+
+  (* Check that there is a potential pgpu candidate for each of the other vgpus *)
+  List.iter (fun vgpu -> 
+      let pgpus = get_pgpus_of_host host in
       test_compatibility vgpu pgpus
-      ) vgpus
+    ) unmapped
```
lindig: Missing `\n`

ocaml/xapi/xapi_pgpu_helpers.ml
```diff
@@ -136,82 +137,108 @@ let assert_capacity_exists_for_VGPU_type ~__context ~self ~vgpu_type =
...
+            raise Api_errors.(Server_error (vm_requires_gpu, [Ref.string_of vm; Ref.string_of gpu_group]))
+          | pgpu :: _ ->
+            (* We can run the compatibility test with any of the PGPUs *)
+            assert_destination_pgpu_is_compatible_with_vm ~__context ~vm ~vgpu ~pgpu ~host ?remote ()
+        end
+    ) vgpu_map;
+
+  (* Check that there is a potential pgpu candidate for each of the other vgpus *)
+  List.iter (fun vgpu -> 
+      let pgpus = get_pgpus_of_host host in
```
lindig: I think we don't need to obtain the `pGPUs` for every `vgpu` that we test - this binding can be moved out of `iter`.

ocaml/xapi/xapi_vm_migrate.ml
mseri: Indeed! Thanks

ocaml/xapi/xapi_pgpu_helpers.ml
```diff
@@ -136,82 +137,108 @@ let assert_capacity_exists_for_VGPU_type ~__context ~self ~vgpu_type =
...
   let test_compatibility vgpu pgpus =
-    match vgpu_impl vgpu with
-      | `passthrough | `gvt_g | `mxgpu -> ()
-      | `nvidia as impl -> 
-        let pgpu = get_first_pgpu_of_implementation vgpu impl pgpus
-        in test_nvidia_compatibility vgpu pgpu
+    match vgpu_impl ~__context vgpu with
+    | `passthrough | `gvt_g | `mxgpu -> ()
+    | `nvidia ->
+      let pgpu_types =
```
lindig: Not sure this is an improvement. I tend to write pipelines when there is a clear transformation and none of the intermediate value is required down below.
```
Db.VGPU.get_GPU_group ~__context ~self:vgpu
|> fun self -> Db.GPU_group.get_GPU_types ~__context ~self
|> fun pgpu_types -> get_first_suitable_pgpu pgpu_types vgpu pgpus
|> fun pgpu -> assert_destination_pgpu_is_compatible_with_vm ~__context ~vm ~vgpu ~pgpu ~host ?remote ()
```

ocaml/xapi/xapi_pgpu_helpers.ml
```diff
@@ -136,82 +137,108 @@ let assert_capacity_exists_for_VGPU_type ~__context ~self ~vgpu_type =
...
-    | Some (rpc, session_id) -> XenAPI.PGPU.get_supported_VGPU_types rpc session_id pgpu
+    | None -> Db.PGPU.get_compatibility_metadata ~__context ~self:pgpu
+    | Some (rpc, session_id) ->XenAPI.PGPU.get_compatibility_metadata rpc session_id pgpu
   in
+  let test_nvidia_compatibility vgpu pgpu = 
+    let pgpu_metadata =
+      let key = Xapi_gpumon.Nvidia.key in
+      try
+        get_compatibility_metadata pgpu
+        |> List.assoc key
```
lindig: I think using `Xapi_gpumon.Nvidia.key` directly is even clearer.

mseri:
This have been tested and did not show regressions

PR #3359, opened 2017-12-07
New api release; bumped api version; indentation.

kc284:


kc284:
Commit message amended.

PR #3363, opened 2017-12-12
CA-276638: Quicktest_cbt: don't fail in cleanup due to VDI refs becoming invalid

gaborigloi:
There is a race condition where we get the list of VDIs, but before we
process them some of them get removed from the DB, and the cleanup fails
with an invalid ref error.

This can happen for example when SM creates an internal VDI record in
the DB, and then later it removes it, in parallel with the cleanup at
the end of the CBT quicktest.


gaborigloi:
I have run the quicktest, it passed (the CBT one at least), and the VDIs still got cleaned up.

PR #3364, opened 2017-12-13
CA-275660 release vGPUs from VM on suspend

lindig:
When a VM is suspended, it must release all vGPUs it has allocated. So
far, this is only happens on shutdown but not suspend. This commit
rectifies this.

### DIFF
--- a/ocaml/xapi/xapi_vm_lifecycle.ml
+++ b/ocaml/xapi/xapi_vm_lifecycle.ml
@@ -606,13 +606,6 @@ let force_state_reset_keep_current_operations ~__context ~self ~value:state =
            (Pvs_proxy_control.find_proxy_for_vif ~__context ~vif))
       (Db.VM.get_VIFs ~__context ~self);
     List.iter
-      (fun vgpu ->
-         Db.VGPU.set_currently_attached ~__context ~self:vgpu ~value:false;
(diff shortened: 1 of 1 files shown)


lindig:
This has been tested manually by observing the entries for a vGPU of a VM that is suspended or halted.

ocaml/xapi/xapi_vm_lifecycle.ml
```diff
@@ -642,7 +635,15 @@ let force_state_reset_keep_current_operations ~__context ~self ~value:state =
     Db.VM.set_resident_on ~__context ~self ~value:Ref.null;
     (* make sure we aren't reserving any memory for this VM *)
     Db.VM.set_scheduled_to_be_resident_on ~__context ~self ~value:Ref.null;
-    Db.VM.set_domid ~__context ~self ~value:(-1L)
+    Db.VM.set_domid ~__context ~self ~value:(-1L);
+
+    (* release vGPUs associated with VM *)
+    Db.VM.get_VGPUs ~__context ~self
+    |> List.iter (fun vgpu ->
+        Db.VGPU.set_currently_attached ~__context ~self:vgpu ~value:false;
```
robhoes: I think that this must stay in the Halted-only case, just like the `currently_attached` fields on the other device types. This means that `currently_attached` remains `true` on suspended VMs. It is just the resource reservation (the lines below) that need to be cleared on suspend.

PR #3365, opened 2017-12-14
Add a fix and test for setting an empty string as a key in a map

jonludlam:
Also fix other tests that already rely on this working

### DIFF
--- a/ocaml/database/db_cache_types.ml
+++ b/ocaml/database/db_cache_types.ml
@@ -351,6 +351,7 @@ let remove_from_set key t =
 
 exception Duplicate
 let add_to_map ~idempotent key value t =
+  if String.length key = 0 then raise Db_exn.Empty_key_in_map;
   let t = Schema.Value.Unsafe_cast.pairs t in
   if List.mem_assoc key t && (not idempotent || List.assoc key t <> value) then raise Duplicate;
   Schema.Value.Pairs ((key, value) :: List.filter (fun (k, _) -> k <> key) t)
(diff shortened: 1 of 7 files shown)


ocaml/database/db_cache_types.ml
```diff
@@ -351,6 +351,7 @@ let remove_from_set key t =
 
 exception Duplicate
 let add_to_map ~idempotent key value t =
+  if String.length key = 0 then raise Db_exn.Empty_key_in_map;
```
lindig: This is more expensive than necessary because it always takes the string length. It's cheaper to test `if key = "" then .. `

ocaml/database/string_marshall_helper.ml
```diff
@@ -28,11 +28,14 @@ let ensure_utf8_xml string =
     warn "Whilst doing 'set' of structured field, string truncated to: '%s'." prefix;
   prefix
 
+let ensure_nonempty s =
+  if String.length s = 0 then raise Db_exn.Empty_key_in_map else s
```
lindig: Same as above

ocaml/database/db_cache_types.ml
jonludlam: https://github.com/ocaml/ocaml/blob/4.04/byterun/str.c#L39-L45

jonludlam: I'm not sure which is cheaper to be honest.

lindig: It's fine as it is because this operation is cheap, as your link to the C is showing.

PR #3373, opened 2017-12-22
xapi-types, xapi-database, idl: fix transitive dependencies issues

mseri:
And overall cleanup of dependencies. Includes the removal of `xapi-stdext-deprecated` from the database but does _not_ include the removal of `xstringext`.

This PR blocks the release of https://github.com/xapi-project/xen-api-libs-transitional/pull/35 as it removes the transitive dependencies re-added here.

### DIFF
--- a/ocaml/database/db_cache_test.ml
+++ b/ocaml/database/db_cache_test.ml
@@ -13,7 +13,7 @@
  *)
 
 open Db_cache_types
-open Stdext.Fun
+open Xapi_stdext_deprecated.Fun
 
 let create_test_db () =
   let schema = Test_schemas.many_to_many in
--- a/ocaml/database/db_cache_test.ml
+++ b/ocaml/database/db_cache_test.ml
@@ -13,28 +13,23 @@
  *)
 
 open Db_cache_types
-open Xapi_stdext_deprecated.Fun
 
 let create_test_db () =
   let schema = Test_schemas.many_to_many in
-  let db =
-    ((fun x -> x)
-     ++ (Db_backend.blow_away_non_persistent_fields schema)
-     ++ (Db_upgrade.generic_database_upgrade))
-      (Database.make schema) in
-
-  db
+  Database.make schema
+  |> Db_upgrade.generic_database_upgrade
+  |> Db_backend.blow_away_non_persistent_fields schema
 
 let check_many_to_many () =
   let db = create_test_db () in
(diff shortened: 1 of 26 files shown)


ocaml/database/db_cache_test.ml
```diff
@@ -30,11 +29,12 @@ let check_many_to_many () =
...
-  let db =
-    ((fun x -> x)
-     ++ (set_field "foo" "foo:1" "bars" (add_to_set "bar:1" (Schema.Value.Set [])))
-     ++ (add_row "foo" "foo:1" (Row.add 0L Db_names.ref (Schema.Value.String "foo:1") (Row.add 0L "bars" (Schema.Value.Set []) Row.empty)))
-     ++ (add_row "bar" "bar:1" (Row.add 0L Db_names.ref (Schema.Value.String "bar:1") (Row.add 0L "foos" (Schema.Value.Set []) Row.empty)))) db in
+  let db = db
+   |> add_row "bar" "bar:1" (Row.add 0L Db_names.ref (Schema.Value.String "bar:1") (Row.add 0L "foos" (Schema.Value.Set []) Row.empty))
+   |> add_row "foo" "foo:1" (Row.add 0L Db_names.ref (Schema.Value.String "foo:1") (Row.add 0L "bars" (Schema.Value.Set []) Row.empty))
+   |> set_field "foo" "foo:1" "bars" (add_to_set "bar:1" (Schema.Value.Set []))
+   |> (fun x -> x)
```
mseri: I wonder if this is really necessary now...

mseri:
As I mentioned in the commit message, the removal of `(++)` in favour of the `|>` should improve the database performances as the first requires a function call while the second is a an internal primitive and is also easier to optimize for the compiler.

The commit where this removal happens is the most delicate one, and it's the reason I've added @jonludlam as a reviewer.

ocaml/xapi-types/features.ml
```diff
@@ -165,4 +165,8 @@ let of_assoc_list l =
     with _ ->
       if List.mem f enabled_when_unknown then Some f else None
   in
-  Stdext.Listext.List.filter_map get_feature all_features
+  (* Filter_map to avoid having to carry the whole xapi-stdext-std *)
+  List.fold_right (fun f acc ->
```
gaborigloi: This does not change the order because of the fold_right

PR #3375, opened 2018-01-02
xapi subpackages: dependency cleanup

mseri:


ocaml/xapi-types/features.ml
```diff
@@ -165,4 +165,8 @@ let of_assoc_list l =
     with _ ->
       if List.mem f enabled_when_unknown then Some f else None
   in
-  Stdext.Listext.List.filter_map get_feature all_features
+  (* Filter_map to avoid having to carry the whole xapi-stdext-std *)
+  List.fold_right (fun f acc ->
```
edwintorok: Not that it matters in this case but this is not tail recursive.

ocaml/database/db_cache_impl.ml
```diff
@@ -296,14 +296,14 @@ let load connections default_schema =
...
-     ++ populate) empty in
-
-  db
+  let empty = Database.update_manifest (Manifest.update_schema (fun _ -> Some (default_schema.Schema.major_vsn, default_schema.Schema.minor_vsn))) (Database.make default_schema)
+  in
+
+  empty
+  |> populate
+  |> Db_upgrade.generic_database_upgrade
+  |> Db_backend.blow_away_non_persistent_fields default_schema
```
edwintorok: Nice, this is also much more natural to read than `++` which has a different order.

ocaml/database/db_upgrade.ml
```diff
@@ -40,6 +39,11 @@ let generic_database_upgrade db =
          Table.add g objref row tbl in
        let tbl = Table.fold add_fields_to_row tbl Table.empty in
        let g = Manifest.generation (Database.manifest db) in
-       ((Database.update ++ (TableSet.update g tblname Table.empty)) (fun _ -> tbl)) db
+       let perform_update =
+         (fun _ -> tbl)
```
edwintorok: if I read this right then this effectively ignores the `db` parameter below?
Could this be simplified to ```
tbl
|> TableSet.update g tblname Table.empty
|> Database.update
```?

File: ocaml/license/jbuild [hunk 4]
```diff
@@ -2,17 +2,13 @@
  ((name daily_license_check_main)
   (public_name daily-license-check)
   (package xapi)
+  (modules (:standard \ test_daily_license_check))
```
edwintorok: Would building that test require adding back these dependencies, or does the test simply not work?

ocaml/xapi-types/jbuild
```diff
@@ -37,6 +37,7 @@ let () = Printf.ksprintf Jbuild_plugin.V1.send {|
    uuidm
    xapi-consts
    xapi-stdext-date
+   xcp
```
edwintorok: You depend on `xcp` here, and on `xapi-idl` below. Which one is the preferred name?

ocaml/database/db_upgrade.ml
mseri: I have left it because I had fear of unexpected side effects, like some globals changing behind the back. I think it can be simplified but I'd like to double check the code and maybe ping @jonludlam

ocaml/license/jbuild
mseri: From what I could see the test has not been built or run for a while, I did not check what happens running it

ocaml/xapi-types/features.ml
mseri: True, I thought that `List.fold_left` and then `rev`, or a custom rewrite, was overkill for the feature list. I'll add a warning to the comment for future reference

ocaml/xapi-types/jbuild
mseri: I believe `xcp` is the current library name, `xapi-idl` the opam package. We will need the giant debt PR to fix all names after the jbuilder port is complete

ocaml/xapi-types/features.ml
edwintorok: Sounds good, just to prevent someone accidentally using it as an example for writing new code.

PR #3376, opened 2018-01-02
CA-276964 on import, set suspend_SR to null when unknown

lindig:
On importing a VM, we must check that its suspend_SR is meaningful and
set it to Null if it isn't. The existing code has the test flipped.

Helpers.check_sr_exists ~__context ~self:vm_record.API.vM_suspend_SR

The above call returns `Some vM_suspend_SR` if it exists and `None`
otherwise. The test needs to look for None but previously did not.


lindig:
This PR is currently being tested on a private build. Let's wait for the result before merging this.

lindig:
This commit is not the ultimate fix for CA-276964 (it is a testing issue) but I still believe it is a bug in the handling of `suspend_SR` which should be fixed.

lindig:
After the test case had been fixed, the private build passed the test: Job 1991767.

PR #3377, opened 2018-01-03
Warning fixes (broken off as requested from PR 3374)

kc284:
Was just made aware to the fact though that this may prevent a compiler downgrade.


robhoes:
I don't mind that too much. Besides, we should be able to take advantage of newer OCaml features, which would also break compat.

PR #3378, opened 2018-01-03
HA master shutdown

edwintorok:
When shutting the HA master we must ensure that we disable the redo log, otherwise we hang/fail when detaching the static VDIs or unplugging the PBD.

We should also ensure that slaves can no longer make requests to the master DB after we've started  to shutdown the master. This is done for now by moving where we stop the remote request thread, but it could also be done by killing the listening stunnel as discussed in the morning.

Tested on REQ-477 branch and was able to shutdown the HA master successfully several times.

While testing also found another bug in the new HA scripts, and I added a commit here that made debugging that easier.

### DIFF
--- a/ocaml/xapi/xapi_ha.ml
+++ b/ocaml/xapi/xapi_ha.ml
@@ -1566,5 +1566,19 @@ let before_clean_shutdown_or_reboot ~__context ~host =
           info "Still waiting to reboot after %.2f seconds" (Unix.gettimeofday () -. start)
         done
     end;
-    List.iter Static_vdis.detach_only (Static_vdis.list())
+
+    (* We must do this before attempting to detach the VDI holding the redo log,
+       otherwise we would either get an error later or hang.
+
+       Note that Xha_metadata_vdi is a VDI with reason = ha_metadata_vdi_reason and type=`redo_log:
+       type=`metadata is for DR *)
+    debug "About to close active redo logs";
+    Redo_log.with_active_redo_logs (Redo_log.shutdown);
+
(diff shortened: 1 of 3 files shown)


ocaml/xapi/xapi_host.ml
```diff
@@ -488,8 +488,17 @@ let shutdown_and_reboot_common ~__context ~host label description operation cmd
...
+          prevent slaves from sending (DB) requests.
+          If we are the slave we cannot shutdown the request thread yet
+          because we might need it when unplugging the PBDs
+        *)
+       Remote_requests.stop_request_thread();
+
   Xapi_ha.before_clean_shutdown_or_reboot ~__context ~host;
   Xapi_pbd.unplug_all_pbds ~__context;
+
   Remote_requests.stop_request_thread();
```
jonludlam: Is it OK to call this twice?

edwintorok: Should be: it sets a a variable and signals a condition, I pushed a fixup to do the `stop_request_thread` only once though, just in case we  change that in the future.

ocaml/xapi/xapi_ha.ml
```diff
@@ -558,6 +558,7 @@ module Monitor = struct
                 end
               end
             with e ->
+              log_backtrace ();
```
gaborigloi: Would this be better under the below debug line? Maybe that would look clearer in the logs.

edwintorok: I'm afraid we'll loose the backtrace then (`Format` sometimes raises/catches some internal exceptions)

PR #3380, opened 2018-01-05
CA-271406 don't call infer_vgpu_map when VM is not live

lindig:
During VM migration, a VM may shut down. When this happens, the VM's
vGPUs get released immediately. This needs to be detected before
infer_vgpu_map() is called as it relies on vGPUs being assigned.

This commit moves the call to infer_vgpu_map into the try-catch block
that deals with edge cases like this. Before infer_vgpu_map is called we
check that the VM is live and hence has its vGPUs still assigned. To
avoid a data race, the liveness check is done only after no more
events from xenopsd are accepted.

### DIFF
--- a/ocaml/xapi/xapi_vm_migrate.ml
+++ b/ocaml/xapi/xapi_vm_migrate.ml
@@ -33,6 +33,8 @@ module SMPERF=Debug.Make(struct let name="SMPERF" end)
 open Client
 open Xmlrpc_client
 
+exception VM_is_not_live of string (* should never escape this module *)
+
 let _sm = "SM"
 let _xenops = "xenops"
 let _host = "host"
@@ -203,7 +205,12 @@ let detach_local_network_for_vm ~__context ~vm ~destination =
 (** Return a map of vGPU device to PCI address of the destination pGPU.
   * This only works _after_ resources on the destination have been reserved
   * by a call to Message_forwarding.VM.allocate_vm_to_host (through
-  * Host.allocate_resources_for_vm for cross-pool migrations). *)
+  * Host.allocate_resources_for_vm for cross-pool migrations).
+  *
+  * We are extra careful to check that the VM has a valid pGPU assigned.
+  * During migration, a VM may suspend or shut down and if this happens
+  * the pGPU is released and getting the PCI will fail.
+  * *)
 let infer_vgpu_map ~__context ?remote vm =
   match remote with
   | None ->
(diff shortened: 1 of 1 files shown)


lindig:
When accepted, this needs also go into the master branch.

jonludlam:
This looks good to me. Local exception declaration, woo!

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -211,8 +211,11 @@ let infer_vgpu_map ~__context ?remote vm =
     List.map (fun self ->
       let vgpu = Db.VGPU.get_record ~__context ~self in
       let device = vgpu.API.vGPU_device in
+      let pgpu = vgpu.API.vGPU_scheduled_to_be_resident_on in
+      if pgpu = Ref.null then (* the VM was halted/suspended *)
```
jonludlam: Don't do this - do Db.is_valid_ref instead.

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -211,8 +211,11 @@ let infer_vgpu_map ~__context ?remote vm =
...
+        raise (Xenops_interface.Does_not_exist("PGPU", Ref.string_of vm));
```
jonludlam: and I don't like the idea of raising Xenops_interface errors - that's quite misleading.

jonludlam: Also, it'd be good to have a comment here saying precisely _why_ we're handling this carefully.

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -224,8 +235,12 @@ let infer_vgpu_map ~__context ?remote vm =
     List.map (fun self ->
       let vgpu = XenAPI.VGPU.get_record rpc session self in
       let device = vgpu.API.vGPU_device in
+      let pgpu = vgpu.API.vGPU_scheduled_to_be_resident_on in
+      if not (Db.is_valid_ref __context pgpu) then
```
jonludlam: This won't work. Hmm, I don't think we expose 'is_valid_ref' externally - perhaps we could emulate by just doing a get_uuid or something?

lindig:
@jonludlam I've cleaned up the commits and simplified the fix. Could you take another look?

jonludlam:
OK, looks good now.

PR #3382, opened 2018-01-05
Travis test: output only failure message

edwintorok:
This is not ideal but Travis is still failing inexplicably on master, lets try to reduce test output to a minimum: just the test summary and the error backtrace.


lindig:
I'm happy to try this until we have a better solution.

gaborigloi:
In 18c089e13565b260639c9e93f840001afa5e09fb we've redirected xapi's logging to the standard output on the REQ477 branch. Maybe that is the source of this issue?
Also, maybe this problem will be solved when we port our unit tests to https://github.com/mirage/alcotest . Our current plan is to write new unit tests using alcotest and compile them into a second unit test binary called by `make test`.

PR #3383, opened 2018-01-08
[CA-273985] Take clustering lock only if SM requires a cluster stack

mcintyre94:
Currently in xapi_pbd and xapi_sr we take a clustering lock (`with_clustering_lock`) regardless of the SR type. However we only need to do this if the SM type requires a cluster stack. This commit adds a new function `with_clustering_lock_if_needed`, which calls `with_clustering_lock` if the SM type requires a cluster stack, otherwise it executes the passed function without any wrapper. The calls in xapi_pdb and xapi_sr are changed to use this new function.

This is tested by the fact that nesting two calls to `with_clustering_lock` will deadlock. We test that the lock is correctly taken when needed by nesting two calls to `with_clustering_lock_if_needed`, which should both take the lock and therefore deadlock. If they don't then the test fails, otherwise it times out and passes. We test hat the lock is not taken when it's not needed by nesting two calls similarly, but this time one requires a cluster stack and the other doesn't. This shouldn't deadlock, which will cause the test the timeout and fail.

Signed-off-by: Callum McIntyre <callum.mcintyre@citrix.com>

### DIFF
--- a/ocaml/xapi/test_clustering.ml
+++ b/ocaml/xapi/test_clustering.ml
@@ -207,6 +207,56 @@ let test_assert_cluster_host_is_enabled_for_matching_sms =
   ; "test_assert_cluster_host_is_enabled_for_matching_sms_succeeds_if_no_cluster_host_exists_and_clustering_is_not_needed" >:: test_assert_cluster_host_is_enabled_for_matching_sms_succeeds_if_no_cluster_host_exists_and_clustering_is_not_needed
   ]
 
+
+(** Tests clustering lock is only taken if needed *)
+let test_clustering_lock_only_taken_if_needed = 
+  let nest_with_clustering_lock_if_needed ~__context ~timeout ~type1 ~type2 ~on_deadlock ~on_no_deadlock =
+    Helpers.timebox
+      ~timeout:timeout
+      ~otherwise: on_deadlock
+      (fun () ->
+        Xapi_clustering.with_clustering_lock_if_needed ~__context ~sr_sm_type:type1 (fun () -> 
+          Xapi_clustering.with_clustering_lock_if_needed ~__context ~sr_sm_type:type2 (fun () ->
+            on_no_deadlock ()        
+          )
+        )
+      )
+  in
+
+  let test_clustering_lock_only_taken_if_needed_nested_calls () =
+    let __context = T.make_test_database () in
+    let _ = T.make_sm ~__context ~_type:"type_corosync" ~required_cluster_stack:["corosync"] () in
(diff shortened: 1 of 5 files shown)


ocaml/xapi/xapi_clustering.ml
```diff
@@ -30,6 +30,22 @@ let with_clustering_lock f =
...
+  let sms_matching_sr_type =
+    Db.SM.get_records_where ~__context
+      ~expr:Db_filter_types.(Eq(Field "type", Literal sr_sm_type))
+  in
+  sms_matching_sr_type
+  |> List.map (fun (_sm_ref, sm_rec) -> sm_rec.API.sM_required_cluster_stack)
+  (* We assume that we only have one SM for each SR type, so this is only to satisfy type checking *)
+  |> List.flatten
+
+let with_clustering_lock_if_needed ~__context ~sr_sm_type f =
```
edwintorok: Could you move this and the above function just above `find_cluster_host`? It doesn't seem to be really necessary to move `get_required_cluster_stacks` uphere.

ocaml/xapi/test_clustering.ml
```diff
@@ -207,11 +207,60 @@ let test_assert_cluster_host_is_enabled_for_matching_sms =
   ; "test_assert_cluster_host_is_enabled_for_matching_sms_succeeds_if_no_cluster_host_exists_and_clustering_is_not_needed" >:: test_assert_cluster_host_is_enabled_for_matching_sms_succeeds_if_no_cluster_host_exists_and_clustering_is_not_needed
   ]
 
+
+(** Tests clustering lock is only taken if needed *)
+let test_clustering_lock_only_taken_if_needed = 
+  let nest_with_clustering_lock_if_needed ~__context ~type1 ~type2 ~on_deadlock ~on_no_deadlock =
+    Helpers.timebox
+      ~timeout: 1.0
```
edwintorok: Lets make this one smaller (e.g. `0.1`) if we do intend to detect the deadlocks during regular unit-testing

ocaml/xapi/xapi_clustering.ml
```diff
@@ -30,6 +30,22 @@ let with_clustering_lock f =
...
+  match get_required_cluster_stacks ~__context ~sr_sm_type with
+    | [] -> f ()
+    | required_cluster_stacks -> with_clustering_lock f
```
edwintorok: `required_cluster_stacks` is not actually used, you can use an underscore in front of the name to avoid the compiler warning about unused  variable like so : `_required_cluster_stacks` (so you still document what the variable is, and the fact that its not used).

PR #3384, opened 2018-01-08
CA-275786: Check PIF prerequisites when creating cluster

gaborigloi:
Previously these prerequisites were not checked, and the operation
succeeded when the PIF of the master wasn't permaplugged for example,
and we ended up with a cluster and a Cluster_host.

### DIFF
--- a/ocaml/xapi/test_clustering.ml
+++ b/ocaml/xapi/test_clustering.ml
@@ -257,6 +257,44 @@ let test_clustering_lock_only_taken_if_needed =
   ; "test_clustering_lock_taken_when_needed_nested_calls" >:: test_clustering_lock_taken_when_needed_nested_calls
   ]
 
+let test_assert_pif_prerequisites () =
+  let __context = Test_common.make_test_database () in
+  let network = Test_common.make_network ~__context () in
+  let localhost = Helpers.get_localhost ~__context in
+  let (_cluster, _cluster_host) = Test_common.make_cluster_and_cluster_host ~__context ~network ~host:localhost () in
+  let exn = "we_havent_decided_on_the_exception_yet" in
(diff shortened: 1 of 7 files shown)


gaborigloi:
I still need to test this.

ocaml/xapi/test_clustering.ml
```diff
@@ -207,11 +207,50 @@ let test_assert_cluster_host_is_enabled_for_matching_sms =
   ; "test_assert_cluster_host_is_enabled_for_matching_sms_succeeds_if_no_cluster_host_exists_and_clustering_is_not_needed" >:: test_assert_cluster_host_is_enabled_for_matching_sms_succeeds_if_no_cluster_host_exists_and_clustering_is_not_needed
   ]
 
+let test_assert_pif_prerequisites () =
+  let __context = Test_common.make_test_database () in
+  let network = Test_common.make_network ~__context () in
+  let localhost = Helpers.get_localhost ~__context in
+  let (_cluster, _cluster_host) = Test_common.make_cluster_and_cluster_host ~__context ~network ~host:localhost () in
```
edwintorok: The test doesn't seem to use these values, does it rely on some side effect of this, or can we safely delete this line?

gaborigloi: True, they are not used at the moment. However, this is just an old test that I've moved and then had to make some changes to make it compile. Maybe these are created to make the test more realistic 🤔

edwintorok: Ok lets keep them.

edwintorok:
You've got conflicts now, could be due to the other PRs that I merged

gaborigloi:
Tested this PR, cluster, GFS2 SR create, pool join, leave were all successful

PR #3386, opened 2018-01-08
CP-26197, CP-25397: Do not run xapi-clusterd if clustering is turned off

edwintorok:
Start the cluster daemon when:
* a Cluster_host / Cluster object is created
* on startup when we try to re-enable the Cluster_host

Stop the cluster daemon when:
* we destroy the Cluster_host or Cluster object

In general before any RPC call to the cluster daemon it has to be started.
So remember whether we started it or not, and if we didn't start it when it gets
the first RPC.

We could have made the daemon socket activated, but see CA-265469 systemd
retries don't work with socket activation reliably.

Also modified xenserver-release by dropping our branch to avoid auto-starting on boot.

This avoids having the xapi-clusterd ports open when nothing uses it.

Wrote unit tests for Cluster/Cluster_host creation/destruction with a mock RPC.

Jenkins test pending.

### DIFF
--- a/ocaml/xapi/create_storage.ml
+++ b/ocaml/xapi/create_storage.ml
@@ -42,14 +42,13 @@ let maybe_reenable_cluster_host __context =
   let host = Helpers.get_localhost __context in
   match Xapi_clustering.find_cluster_host ~__context ~host with
   | Some self ->
-      if not (Db.Cluster_host.get_enabled ~__context ~self) then
-        Xapi_cluster_host.enable ~__context ~self
+     Xapi_cluster_host.enable ~__context ~self
   | None -> ()
 
 let plug_unplugged_pbds __context =
   (* If the plug is to succeed for SM's requiring a cluster stack
    * we have to enable the cluster stack too if we have one *)
-  maybe_reenable_cluster_host __context;
+  log_and_ignore_exn(fun () -> maybe_reenable_cluster_host __context);
   let my_pbds = Helpers.get_my_pbds __context in
   List.iter
     (fun (self, pbd_record) ->
--- a/ocaml/xapi/xapi_clustering.ml
+++ b/ocaml/xapi/xapi_clustering.ml
@@ -163,12 +163,45 @@ let assert_cluster_has_one_node ~__context ~self =
   | 1 -> ()
   | n -> raise Api_errors.(Server_error(cluster_does_not_have_one_node, [string_of_int n]))
 
+module Daemon = struct
+  let started = ref false
+  let m = Mutex.create ()
+
+  let maybe_call_script ~__context script params =
+    match Context.get_test_clusterd_rpc __context with
+    | Some _ -> debug "in unit test, not calling %s %s" script (String.concat " " params)
+    | None -> ignore (Helpers.call_script script params)
+
+  let require ~__context =
+    Stdext.Threadext.Mutex.execute m (fun () ->
+        (* this function gets called on each RPC, it should only call out to `/sbin/service` when needed *)
+        if not !started then begin
+            debug "Cluster daemon: not started, starting it now";
+            maybe_call_script ~__context "/sbin/service" [ "xapi-clusterd"; "start" ];
+            started := true;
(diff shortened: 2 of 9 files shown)


ocaml/xapi/xapi_clustering.ml
```diff
@@ -163,12 +163,42 @@ let assert_cluster_has_one_node ~__context ~self =
...
+end
+
 (* xapi-clusterd only listens on message-switch,
  * the URL here would be for calling xapi-clusterd through an HTTP interface,
  * but that is not supported (yet).
  * Instead of returning an empty URL which wouldn't work just raise an
  * exception. *)
-let rpc = Cluster_client.rpc (fun () -> failwith "Can only communicate with xapi-clusterd through message-switch")
+let rpc ~__context =
+  Daemon.require ();
```
gaborigloi: Another option would be to move this into the `None` case below and that way we wouldn't need the checks in the `Daemon` module. But it's also fine as it is.

edwintorok: I do want to run it in both cases, but I could pass in a parameter instead, let me see if thats cleaner.

ocaml/xapi/xapi_clustering.ml
```diff
@@ -163,12 +163,42 @@ let assert_cluster_has_one_node ~__context ~self =
...
+  match Context.get_test_rpc __context with
```
gaborigloi: Hmm, maybe we could have a separate RPC for xapi-clusterd: `Context.test_clusterd_rpc`. This way we wouldn't have to worry about confusing xapi and xapi-clusterd RPCs.

edwintorok: Pushed a fixup to address this.

edwintorok:
I reverted the change that introduced Context.get_test_clusterd_rpc because the test was failing with a connection failure (there is no message-switch running).
Travis gave us a green tick because of a problem I introduced in log shrinking, see the fix here: https://github.com/xapi-project/xen-api/pull/3394

Also we need to always (attempt to) enable the Cluster_host on startup. Enable is idempotent (we have a unit test that checks that), and now that we start clusterd only on demand we need to send it at least one API call and wait for it to succeed.

gaborigloi:
@edwintorok Yep, we need to use both the `test_rpc` and `test_clusterd_rpc` for mocking xapi and xapi-clusterd functions, respectively. So for this reason I think the separation is nicer, the corresponding functions are kept together.

edwintorok:
Ah just for separation, OK, I'll put it back and just separate them

gaborigloi:
Yep, in https://github.com/xapi-project/xen-api/pull/3386/commits/4ca76eb2dd48520624b19a5a60f8177bb8c4b027 we forgot to also supply the mocked xapi RPCs in `test_rpc`, these got removed, only `test_clusterd_rpc` was specified.

edwintorok:
Updated the PR, the separation works now.

edwintorok:
Rebased the fixups.

ocaml/xapi/create_storage.ml
```diff
@@ -42,14 +42,13 @@ let maybe_reenable_cluster_host __context =
...
-      if not (Db.Cluster_host.get_enabled ~__context ~self) then
-        Xapi_cluster_host.enable ~__context ~self
+     Xapi_cluster_host.enable ~__context ~self
   | None -> ()
 
 let plug_unplugged_pbds __context =
   (* If the plug is to succeed for SM's requiring a cluster stack
    * we have to enable the cluster stack too if we have one *)
-  maybe_reenable_cluster_host __context;
+  log_and_ignore_exn(fun () -> maybe_reenable_cluster_host __context);
```
gaborigloi: So now we will just log an error below if we failed to enable clustering, instead of failing the whole function with an exception. Maybe that's better indeed, as it's closer to the original behaviour 🤔

gaborigloi: Yes, repair worked the last time I checked. It does call pool resync.

PR #3387, opened 2018-01-09
CP-26302: Add driver_name in PCI object

cheng-z:
Signed-off-by: Cheng Zhang <cheng.zhang@citrix.com>

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -9245,6 +9245,7 @@ let pci =
       field ~qualifier:StaticRO ~ty:String ~lifecycle:[Published, rel_clearwater_whetstone, ""] "subsystem_vendor_name" "Subsystem vendor name" ~default_value:(Some (VString ""));
       field ~qualifier:StaticRO ~ty:String ~lifecycle:[] "subsystem_device_id" "Subsystem device ID" ~default_value:(Some (VString "")) ~internal_only:true;
       field ~qualifier:StaticRO ~ty:String ~lifecycle:[Published, rel_clearwater_whetstone, ""] "subsystem_device_name" "Subsystem device name" ~default_value:(Some (VString ""));
+      field ~qualifier:StaticRO ~ty:String ~lifecycle:[Published, rel_kolkata, ""] "driver_name" "Driver name" ~default_value:(Some (VString ""));
       field ~qualifier:DynamicRO ~ty:(Ref _vm) ~lifecycle:[Published, rel_falcon, ""] ~internal_only:true "scheduled_to_be_attached_to" "The VM to which this PCI device is scheduled to be attached (passed through)" ~default_value:(Some (VRef null_ref));
     ]
     ()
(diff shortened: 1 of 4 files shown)


ocaml/xapi/xapi_pci_helpers.ml
```diff
@@ -28,8 +28,17 @@ type pci = {
   subsystem_vendor: pci_property option;
   subsystem_device: pci_property option;
   related: string list;
+  driver_name: string option;
 }
 
+let get_driver_name address =
+    try
+      let driver_path = Unix.readlink (Printf.sprintf "/sys/bus/pci/devices/%s/driver" address) in
```
mseri: We need to do this because `pciutils` does not know about it (see recent relevant commit: https://github.com/pciutils/pciutils/commit/11339c0d3d0c9064717983fa283bafcdc5298fe7)

ocaml/xapi/xapi_pci_helpers.ml
```diff
@@ -28,8 +28,17 @@ type pci = {
...
+      match Astring.String.cut ~sep:"/" ~rev:true driver_path with
+      | Some (prefix, suffix) -> Some suffix
+      | None -> None
+    with _ -> None
```
mseri: Please add a debug line printing the exception value in case of failure

ocaml/idl/datamodel.ml
```diff
@@ -9245,6 +9245,7 @@ let pci =
       field ~qualifier:StaticRO ~ty:String ~lifecycle:[Published, rel_clearwater_whetstone, ""] "subsystem_vendor_name" "Subsystem vendor name" ~default_value:(Some (VString ""));
       field ~qualifier:StaticRO ~ty:String ~lifecycle:[] "subsystem_device_id" "Subsystem device ID" ~default_value:(Some (VString "")) ~internal_only:true;
       field ~qualifier:StaticRO ~ty:String ~lifecycle:[Published, rel_clearwater_whetstone, ""] "subsystem_device_name" "Subsystem device name" ~default_value:(Some (VString ""));
+      field ~qualifier:StaticRO ~ty:String ~lifecycle:[Published, rel_kolkata, ""] "driver_name" "Driver name" ~default_value:(Some (VString ""));
```
robhoes: Please add new fields at the end of the list to avoid any possible compatibility issues.

ocaml/xapi/xapi_pci_helpers.ml
mseri: Let me clarify the comment. We cannot use `libpci` (part of `pciutils`) to read the driver name, because it does not have the functionality. In fact, `lspci` (part of `pciutils` as well) does it exactly as we are doing here (see the commit linked above). So I think this is the way to go to read the driver name

ocaml/xapi/xapi_pci_helpers.ml
cheng-z: Sure, I will

ocaml/idl/datamodel.ml
cheng-z: OK

cheng-z:
Hi @robhoes  and @mseri 
I've update the pull request and add a commit to fix your comments.
Please help to look and again and merge if no other concern.
Thanks.

robhoes:
It's good now, but why did travis not build it?

robhoes:
Hmm... this feature branch is not enabled for travis yet. I'll do that now.

PR #3389, opened 2018-01-09
CP-26469: Defined release jura and bumped the API version to 2.9.

kc284:


lindig:
I will merge this later today.

PR #3391, opened 2018-01-09
CP-26098: Upgrade xapi to use PPX-based v6d interface

minishrink:
This PR is the second of three which port the licensing daemon to PPX:
1. xcp-idl: [port V6d interface to use PPX](https://github.com/xapi-project/xcp-idl/pull/186)
2. xapi: update references to v6 interface
3. v6d: port daemons to ppx

This PR in particular updates xapi to use the following:
- ppx-based error handling (defined in v6 interface)
- new edition type (defined in v6 interface)
- some formatting changes for readability, as well as whitespace removal and ocp-indent

Note: CP-26098 is part of REQ-417

---
This change is [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3391)

### DIFF
--- a/ocaml/xapi/cli_operations.ml
+++ b/ocaml/xapi/cli_operations.ml
@@ -678,11 +678,11 @@ let make_param_funs getall getallrecs getbyuuid record class_name def_filters de
             let existing_params = try Hashtbl.find set_map_table set_map with Not_found -> [] in
             Hashtbl.replace set_map_table set_map ((key, v) :: existing_params)
           | None, None ->
-              let add_to_map = match field.add_to_map with Some f -> f | None -> failwith ("Map field '"^s^"' is read-only.") in
-              let remove_from_map = match field.remove_from_map with Some f -> f | None -> failwith (Printf.sprintf "Records broken (field %s)" s) in
-              let map = get_map () in
-              if List.mem_assoc key map then remove_from_map key;
-              add_to_map key v
+            let add_to_map = match field.add_to_map with Some f -> f | None -> failwith ("Map field '"^s^"' is read-only.") in
+            let remove_from_map = match field.remove_from_map with Some f -> f | None -> failwith (Printf.sprintf "Records broken (field %s)" s) in
+            let map = get_map () in
+            if List.mem_assoc key map then remove_from_map key;
+            add_to_map key v
           | Some _, Some _ -> failwith (Printf.sprintf "Broken Records (field %s)" s)
         end
       | Set s -> failwith "Cannot param-set on set fields"
@@ -828,7 +828,7 @@ let gen_cmds rpc session_id =
     ; Client.PVS_proxy.(mk get_all get_all_records_where get_by_uuid pvs_proxy_record "pvs-proxy" [] ["uuid"; "vif-uuid"; "pvs-site-uuid"; "currently-attached"; "cache-sr-uuid"] rpc session_id)
     ; Client.PVS_cache_storage.(mk get_all get_all_records_where get_by_uuid pvs_cache_storage_record "pvs-cache-storage" [] ["uuid"; "host-uuid"; "sr-uuid"; "pvs-site-uuid"; "size"] rpc session_id)
     ; Client.Feature.(mk get_all get_all_records_where get_by_uuid feature_record "feature" []
-        ["uuid"; "name-label"; "name-description"; "enabled"; "experimental"; "version"; "host-uuid"] rpc session_id)
+                        ["uuid"; "name-label"; "name-description"; "enabled"; "experimental"; "version"; "host-uuid"] rpc session_id)
     ; Client.SDN_controller.(mk get_all get_all_records_where get_by_uuid sdn_controller_record "sdn-controller" [] ["uuid"; "protocol"; "address"; "port"] rpc session_id)
     ; Client.PUSB.(mk get_all get_all_records_where get_by_uuid pusb_record "pusb" [] ["uuid"; "path"; "product-id"; "product-desc"; "vendor-id"; "vendor-desc"; "serial"; "version";"description"] rpc session_id)
     ; Client.USB_group.(mk get_all get_all_records_where get_by_uuid usb_group_record "usb-group" [] ["uuid";"name-label";"name-description"] rpc session_id)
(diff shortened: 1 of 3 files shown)


minishrink:
I found myself making a lot of formatting changes, so those might come in a separate PR.

ocaml/xapi/cli_operations.ml
```diff
@@ -208,14 +208,19 @@ let host_license_of_r host_r editions =
...
-  let edition_short = List.hd
-      (List.filter_map (fun (a, (_, b, _)) -> if a = edition then Some b else None) editions) in
+  let edition_str = host_r.API.host_edition in
+  let edition_short =
+    editions
+    |> List.filter_map
+      V6_interface.(fun ed -> if ed.ed_title = edition_str
+                     then Some ed.ed_code
+                     else None)
+    |> List.hd in
```
robhoes: We should really get rid of this List.hd...

ocaml/xapi/cli_operations.ml
```diff
@@ -208,14 +208,19 @@ let host_license_of_r host_r editions =
...
+     V6_interface.(fun ed -> if ed.ed_title = edition_str
+                                then Some ed.ed_code
+                                else None)
+    |> List.hd in
   {
     hostname = host_r.API.host_hostname;
     uuid = host_r.API.host_uuid;
     rstr = rstr;
-    edition = edition;
+    edition = edition_str;
```
robhoes: What's the benefit of the rename?

ocaml/xapi/cli_operations.ml
```diff
@@ -3249,16 +3254,15 @@ let with_license_server_changes printer rpc session_id params hosts f =
 
 let host_apply_edition printer rpc session_id params =
   let host =
-    if List.mem_assoc "host-uuid" params then
-      Client.Host.get_by_uuid rpc session_id (List.assoc "host-uuid" params)
-    else
-      get_host_from_session rpc session_id in
+    if List.mem_assoc "host-uuid" params
+    then Client.Host.get_by_uuid rpc session_id (List.assoc "host-uuid" params)
```
robhoes: This seems an unnecessary change?

ocaml/xapi/xapi_host.ml
```diff
@@ -1339,16 +1339,18 @@ let apply_edition_internal  ~__context ~host ~edition ~additional =
   let params =
     ("current_edition", current_edition) ::
     license_server @ current_license_params @ additional in
+  let edition_str = edition in
+  (* rename edition variable to prevent namespace clashes with V6_interface *)
```
robhoes: Where then? I don't see a `let edition` or something like that in there?

mseri: It could be renamed in the function definition itself: https://github.com/xapi-project/xen-api/pull/3391/files/634cfc05e57fb3fbaab6312554d0c34dab103857..733531584953d17ce23901200b4f06e6db122086#diff-126a0619d5403393ed49475b5dd96aa4R1326

robhoes: It can, but why?

ocaml/xapi/cli_operations.ml
minishrink: Replace it with a safer List.hd using options?

ocaml/xapi/cli_operations.ml
minishrink: I think I did that to keep some consistency with the renaming in v6d, but you're right in that it serves no purpose here so I'll remove it.

ocaml/xapi/cli_operations.ml
robhoes: It's usually better to use a match to get the first element of a list, and explicitly handle the `[]` case. The exception raised by `List.hd` in case the list is empty is often not what you want.

ocaml/xapi/xapi_host.ml
mseri: You are right, I thought we were in the same situation as in `v6d` but here it's not needed

ocaml/xapi/cli_operations.ml
minishrink: Satisfying the type system, which will otherwise recognise `edition` as a generated RPC type.

minishrink: Another way to deal with this would be to change the type name from edition to edition_repr or edition_rec, edition_type, or something

ocaml/xapi/xapi_host.ml
minishrink: As I understand it, prior to locally opening `V6_interface`, `edition` is interpreted as a string. Within the V6_interface namespace, it is interpreted as an RPC type instead of a string, which upsets the compiler.

Which do you think is the best solution?
1. Rename the `edition` type in the V6 interface to avoid clashes
2. Rename to `edition_str` within the function definition (this requires updating elsewhere as well)
3. Local rename for this small snippet of code (used in this PR)

ocaml/xapi/cli_operations.ml
minishrink: As the `edition_short` value is used in the host license record, should we handle a failure with `free`, `""`, or by raising an error?

ocaml/xapi/cli_operations.ml
minishrink: I may have gone on a slight spree in order to make some code more readable (have since moved that to a separate branch), though if/then/else just comes down to preference. Should I remove it, or just bundle it into a separate PR?

ocaml/xapi/xapi_host.ml
robhoes: Or 4: don't locally open `V6_interface` anymore, but just use `V6_interface.`.

minishrink: Also works!

ocaml/xapi/cli_operations.ml
robhoes: I happen to think that the original code is more readable ;)

Generally speaking, we should avoid making arbitrary style changes unless your changing those lines for another reason as well, or the new version is obviously Much Better™.

ocaml/xapi/xapi_host.ml
```diff
@@ -1400,7 +1400,7 @@ let apply_edition ~__context ~host ~edition ~force =
     raise (Api_errors.Server_error (Api_errors.ha_is_enabled, []))
   else
     let additional = if force then ["force", "true"] else [] in
-    apply_edition_internal ~__context ~host ~edition ~additional
+    apply_edition_internal ~__context ~host ~edition:edition ~additional
```
robhoes: That shouldn't be needed.

minishrink: Whoops, holdover from removing `edition_str`.

PR #3392, opened 2018-01-09
CA-279161 if unavailable, don't update pGPU compat data

lindig:
On startup, xapi updates its GPU records. This includes pGPU
compatibility meta data. Obtaining this data might fail if the pGPU is
used in pass-trough mode at the time. In the past this lead to empty
meta data overwriting the existing meta data. Since a restart of xapi
can happen for many reasons, this led to losing pGPU meta data which was
needed later.

This commit prevents existing pGPU meta data being overwritten if no
meta data can be obtained.

### DIFF
--- a/ocaml/xapi/xapi_pgpu.ml
+++ b/ocaml/xapi/xapi_pgpu.ml
@@ -48,9 +48,17 @@ let maybe_fetch_compatibility_metadata ~__context ~pgpu_pci =
     []
 
 let populate_compatibility_metadata ~__context ~pgpu ~pgpu_pci =
-  let () = Db.PGPU.set_compatibility_metadata ~__context ~self:pgpu
-    ~value:(maybe_fetch_compatibility_metadata ~__context ~pgpu_pci)
-  in ()
+  let this = "populate_compatibility_metadata" in
+  try
+    let value = fetch_compatibility_metadata ~__context ~pgpu_pci in
+    Db.PGPU.set_compatibility_metadata ~__context ~self:pgpu ~value
+  with
+  | Gpumon_interface.NvmlInterfaceNotAvailable ->
+      info "%s: can't get compat data for pgpu_pci:%s, keeping existing data"
+        this (Ref.string_of pgpu_pci)
+  | err ->
+      debug "%s: obtaining compat data for pgpu_pci:%s failed with %s"
+        this (Ref.string_of pgpu_pci) (Printexc.to_string err)
 
 let create ~__context ~pCI ~gPU_group ~host ~other_config
     ~supported_VGPU_types ~size ~dom0_access


ocaml/xapi/xapi_pgpu.ml
```diff
@@ -48,9 +48,17 @@ let maybe_fetch_compatibility_metadata ~__context ~pgpu_pci =
     []
 
 let populate_compatibility_metadata ~__context ~pgpu ~pgpu_pci =
-  let () = Db.PGPU.set_compatibility_metadata ~__context ~self:pgpu
-    ~value:(maybe_fetch_compatibility_metadata ~__context ~pgpu_pci)
```
robhoes: Can we now delete the `maybe_fetch_compatibility_metadata` function above?

ocaml/xapi/xapi_pgpu.ml
```diff
@@ -48,9 +48,17 @@ let maybe_fetch_compatibility_metadata ~__context ~pgpu_pci =
...
-  in ()
+  let this = "populate_compatibility_metadata" in
+  try
+    let value = fetch_compatibility_metadata ~__context ~pgpu_pci in
+    Db.PGPU.set_compatibility_metadata ~__context ~self:pgpu ~value
+  with
+  | Gpumon_interface.NvmlInterfaceNotAvailable ->
```
robhoes: What will happen if you replace an nvidia pGPU, which has `compatibility_metadata`, which a pGPU of another brand, in the same PCI slot?

ocaml/xapi/xapi_pgpu.ml
lindig: It is used in one more place. Will take a look.

ocaml/xapi/xapi_pgpu.ml
lindig: If you replace a card, you need a reboot - and this means, you should be able to get the correct data because no VM is using it. I believe the risk that we have stale data is very small and a much better trade-off than not having compatibility data when we need it.

ocaml/xapi/xapi_pgpu.ml
lindig: https://github.com/xapi-project/xen-api/blob/vgpu-migration/ocaml/xapi/xapi_pgpu.ml#L67

lindig: `maybe_fetch_compatibility_metadata` is used in the `create` function. I also don't like the code duplication between `populate_compatibility_metadata` and `fetch_compatibility_metadata` but reducing this would lead to more code in other places. So I think I would keep it.

ocaml/xapi/xapi_pgpu.ml
robhoes: If the new card is a non-nvidia one, then gpumon will raise `Gpumon_interface.NvmlInterfaceNotAvailable ` and still we want the metadata to be cleared.

robhoes: It may be that the entire PGPU object is recreated in this case, but it would be good to check.

robhoes: I checked in the code: it looks like you indeed get new PCI and PGPU objects if you replace the card with a different kind of card.

lindig: I have updated the commit message to explain why I think this is save.

PR #3393, opened 2018-01-09
CP-26457: Make multipath configuration a host parameter

gaborigloi:
And sync the ISCSI IQN and multipathing config files at startup.

### DIFF
--- a/ocaml/xapi/xapi_host_helpers.ml
+++ b/ocaml/xapi/xapi_host_helpers.ml
@@ -267,7 +267,7 @@ module Host_requires_reboot = struct
       )
 end
 
-module InitiatorName = struct
+module Configuration = struct
 
   let make_initiatorname_config iqn hostname =
     (* CA-18000: there is a 30 character limit to the initiator when talking to
@@ -286,6 +286,13 @@ module InitiatorName = struct
     let config_file = make_initiatorname_config iqn hostname in
     Unixext.write_string_to_file !Xapi_globs.iscsi_initiator_config_file config_file
 
+  let set_multipathing enabled =
+    let flag = !Xapi_globs.multipathing_config_file in
+    if enabled then Unixext.touch_file flag
+    else begin
+      if Sys.file_exists flag then Sys.remove flag
+    end
+
   let watch_other_configs ~__context delay =
     let loop token =
       Helpers.call_api_functions ~__context (fun rpc session_id ->
@@ -296,14 +303,24 @@ module InitiatorName = struct
           List.iter (fun ev ->
               match Event_helper.record_of_event ev with
                 | Event_helper.Host (host_ref, Some host_rec) -> begin
-                  let oc = host_rec.API.host_other_config in
-                  let iscsi_iqn = try Some (List.assoc "iscsi_iqn" oc) with _ -> None in
-                  match iscsi_iqn with
-                  | None -> ()
-                  | Some "" -> ()
-                  | Some iqn when iqn <> host_rec.API.host_iscsi_iqn ->
-                      Client.Client.Host.set_iscsi_iqn rpc session_id host_ref (List.assoc "iscsi_iqn" oc)
-                  | _ -> ()
+                    let oc = host_rec.API.host_other_config in
+                    let iscsi_iqn = try Some (List.assoc "iscsi_iqn" oc) with _ -> None in
+                    begin match iscsi_iqn with
+                      | None -> ()
+                      | Some "" -> ()
+                      | Some iqn when iqn <> host_rec.API.host_iscsi_iqn ->
+                        Client.Client.Host.set_iscsi_iqn rpc session_id host_ref iqn
+                      | _ -> ()
+                    end;
+                    (* Accepted values are "true" and "false" *)
+                    (* If someone deletes the multipathing other_config key, we don't do anything *)
+                    let multipathing = try Some (List.assoc "multipathing" oc |> Pervasives.bool_of_string) with _ -> None in
+                    begin match multipathing with
+                      | None -> ()
+                      | Some multipathing when multipathing <> host_rec.API.host_multipathing ->
+                        Client.Client.Host.set_multipathing rpc session_id host_ref multipathing
+                      | _ -> ()
+                    end
                   end
                 | _ -> ())
               events.Event_types.events;
@@ -319,7 +336,7 @@ module InitiatorName = struct
           try
             let rec inner token = inner (loop token) in inner ""
           with e ->
-            error "Caught exception in InitiatorName.start_watcher_thread: %s" (Printexc.to_string e);
+            error "Caught exception in Configuration.start_watcher_thread: %s" (Printexc.to_string e);
             Thread.delay 5.0;
         end;
       done) () |> ignore
--- a/ocaml/xapi/xapi_host_helpers.ml
+++ b/ocaml/xapi/xapi_host_helpers.ml
@@ -293,6 +293,14 @@ module Configuration = struct
       if Sys.file_exists flag then Sys.remove flag
     end
 
+  let sync_config_files ~__context =
+    (* If the host fields are not in sync with the values in other_config,
+       the other_config watcher thread will make sure that these functions will
+       be called again with the up to date values. *)
+    let self = Helpers.get_localhost ~__context in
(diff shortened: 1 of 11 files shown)


gaborigloi:
Tested:
Cluster, GFS2 SR create, pool join, leave still works.
```
[root@vhost-1 ~]# ls /var/run/nonpersistent/multipath_enabled
ls: cannot access /var/run/nonpersistent/multipath_enabled: No such file or directory
[root@vhost-1 ~]# cat /etc/iscsi/initiatorname.iscsi 
InitiatorName=iqn.2018-01.com.example:0e00b115
InitiatorAlias=vhost-1
[root@vhost-1 ~]# xe host-list hostname=vhost-1 params=iscsi_iqn,multipathing,uuid
uuid ( RO)            : dea25425-455c-4e56-a223-ad046a8e7e69
       iscsi_iqn ( RW): iqn.2018-01.com.example:0e00b115
    multipathing ( RW): false


[root@vhost-1 ~]# xe host-param-set uuid=dea25425-455c-4e56-a223-ad046a8e7e69 multipathing=true
[root@vhost-1 ~]# xe host-list hostname=vhost-1 params=iscsi_iqn,multipathing,uuid,other-config
uuid ( RO)            : dea25425-455c-4e56-a223-ad046a8e7e69
    other-config (MRW): multipathing: true; agent_start_time: 1515510461.; boot_time: 1515506363.; iscsi_iqn: iqn.2018-01.com.example:0e00b115
       iscsi_iqn ( RW): iqn.2018-01.com.example:0e00b115
    multipathing ( RW): true


[root@vhost-1 ~]# cat /var/run/nonpersistent/multipath_enabled
[root@vhost-1 ~]# ls -la /var/run/nonpersistent/multipath_enabled
-rw-r--r-- 1 root root 0 Jan  9 15:36 /var/run/nonpersistent/multipath_enabled
[root@vhost-1 ~]# xe host-param-set uuid=dea25425-455c-4e56-a223-ad046a8e7e69 multipathing=false
[root@vhost-1 ~]# xe host-list hostname=vhost-1 params=iscsi_iqn,multipathing,uuid,other-config
uuid ( RO)            : dea25425-455c-4e56-a223-ad046a8e7e69
    other-config (MRW): multipathing: false; agent_start_time: 1515510461.; boot_time: 1515506363.; iscsi_iqn: iqn.2018-01.com.example:0e00b115
       iscsi_iqn ( RW): iqn.2018-01.com.example:0e00b115
    multipathing ( RW): false


[root@vhost-1 ~]# ls -la /var/run/nonpersistent/multipath_enabled
ls: cannot access /var/run/nonpersistent/multipath_enabled: No such file or directory
[root@vhost-1 ~]# xe host-param-set uuid=dea25425-455c-4e56-a223-ad046a8e7e69 other-config:multipathing=true
[root@vhost-1 ~]# xe host-list hostname=vhost-1 params=iscsi_iqn,multipathing,uuid,other-config
uuid ( RO)            : dea25425-455c-4e56-a223-ad046a8e7e69
    other-config (MRW): multipathing: true; agent_start_time: 1515510461.; boot_time: 1515506363.; iscsi_iqn: iqn.2018-01.com.example:0e00b115
       iscsi_iqn ( RW): iqn.2018-01.com.example:0e00b115
    multipathing ( RW): true


[root@vhost-1 ~]# ls -la /var/run/nonpersistent/multipath_enabled
-rw-r--r-- 1 root root 0 Jan  9 15:37 /var/run/nonpersistent/multipath_enabled
[root@vhost-1 ~]# cat /var/run/nonpersistent/multipath_enabled
[root@vhost-1 ~]# xe host-param-set uuid=dea25425-455c-4e56-a223-ad046a8e7e69 other-config:multipathing=false
[root@vhost-1 ~]# xe host-list hostname=vhost-1 params=iscsi_iqn,multipathing,uuid,other-config
uuid ( RO)            : dea25425-455c-4e56-a223-ad046a8e7e69
    other-config (MRW): multipathing: false; agent_start_time: 1515510461.; boot_time: 1515506363.; iscsi_iqn: iqn.2018-01.com.example:0e00b115
       iscsi_iqn ( RW): iqn.2018-01.com.example:0e00b115
    multipathing ( RW): false


[root@vhost-1 ~]# ls -la /var/run/nonpersistent/multipath_enabled
ls: cannot access /var/run/nonpersistent/multipath_enabled: No such file or directory
[root@vhost-1 ~]#
```

ocaml/xapi/xapi_host_helpers.ml
```diff
@@ -296,14 +311,23 @@ module InitiatorName = struct
...
+                    let iscsi_iqn = try Some (List.assoc "iscsi_iqn" oc) with _ -> None in
+                    begin match iscsi_iqn with
+                      | None -> ()
+                      | Some "" -> ()
+                      | Some iqn when iqn <> host_rec.API.host_iscsi_iqn ->
+                        Client.Client.Host.set_iscsi_iqn rpc session_id host_ref iqn
+                      | _ -> ()
+                    end;
+                    (* Accepted values are "true" and "false" *)
+                    let multipathing = try Some (List.assoc "multipathing" oc |> Pervasives.bool_of_string) with _ -> None in
```
edwintorok: What if `multipathing` gets removed from `other-config`? We probably want to disable it, but might be tricky to implement because the setter itself temporarily removes it and adds it back.

ocaml/xapi/xapi_host_helpers.ml
```diff
@@ -286,6 +286,21 @@ module InitiatorName = struct
...
+    else begin
+      if Sys.file_exists flag then Sys.remove flag
+    end
+
+  let sync_config_files ~__context =
+    (* If the host fields are not in sync with the values in other_config,
+       the other_config watcher thread will make sure that these functions will
+       be called again with the up to date values. *)
+    let self = Helpers.get_localhost ~__context in
+    set_initiator_name (Db.Host.get_iscsi_iqn ~__context ~self);
```
edwintorok: Can this fail when there is no iscsi IQN configured?

gaborigloi: Right now will not fail, as it's a string, and it's the empty string by default. It will just write the empty string to the config file.

ocaml/xapi/xapi_host_helpers.ml
gaborigloi: I was thinking the same thing. It depends on how it was used. But it isn't easy indeed, so we shouldn't deal with that case if we don't have to.

ocaml/xapi/xapi_host_helpers.ml
gaborigloi: It's empty by default in the datamodel though 🤔

ocaml/xapi/xapi_host_helpers.ml
edwintorok: I checked XenCenter and it sets `multipathing: false` when turning off multipathing, so thats probably OK, and the admin guide says to set it to false too, so we are probably OK, but might be worth adding a comment here saying that if someone deletes the other-config multipathing key we don't do anything, and that could be improved later.

ocaml/xapi/xapi_host_helpers.ml
edwintorok: Yes empty = don't change it, I think.

edwintorok: Nevermind, looks like this is set by firstboot scripts, so lets not worry about empty IQN.

gaborigloi: Sounds good. As discussed, I'll disallow empty IQNs in host.set_iscsi_iqn

ocaml/xapi/xapi_host_helpers.ml
gaborigloi: Sounds good. As discussed, we shouldn't worry about this for the above reasons, plus the new setter will be the official way to change multipathing, so we don't have to worry too much about this cause it's just for backward compat.

gaborigloi: Added the comment. We can merge this now if it's ok, and then I can add the check for empty iscsi iqn in another pr.

gaborigloi:
created https://github.com/xapi-project/xen-api/pull/3398 for rejecting empty iscsi iqn.

PR #3395, opened 2018-01-10
CA-271406 infer_vgpu_map: VM might not be live

lindig:
During VM migration, a VM may shut down. When this happens, the VM's
vGPUs get released immediately.  This needs to be detected before or
while infer_vgpu_map() is called as it relies on vGPUs being assigned.
Otherwise we see an invalid handle to a PGPU.

This commit moves the call to infer_vgpu_map into the try-catch block
that deals with edge cases like this. If the VM is no longer live during
migration, we receive an new exception and deal with this.


lindig:
This was merged into the vGPU feature branch and needs to be merged into master, too.

PR #3396, opened 2018-01-10
CA-279161 if unavailable, don't update pGPU compat data

lindig:
On startup, xapi updates its GPU records. This includes pGPU
compatibility meta data. Obtaining this data might fail if the pGPU is
used in pass-trough mode at the time. In the past this lead to empty
meta data overwriting the existing meta data. Since a restart of xapi
can happen for many reasons, this led to losing pGPU meta data which was
needed later.

This commit prevents existing pGPU meta data being overwritten if no
meta data can be obtained. This could introduce stale data, so why is
this safe?

* Any xapi restart when pGPU are accessible will obtain the correct value.
  This provides a way to correct any potential problem.

* If the corresponding card is removed, a reboot is necessary and during
  xapi startup the pGPU object will be removed.

* If the corresponding card is replaced in the same PCI slot by a
  different card, a reboot is required during which pGPUs are
  accessible.  If the card does not provide pGPU metadata, we would not
  overwrite the existing data. However, provided the new card has a
  different PCI identifier, the old pGPU object will be removed entirely
  and a new object created.

### DIFF
--- a/ocaml/xapi/xapi_pgpu.ml
+++ b/ocaml/xapi/xapi_pgpu.ml
@@ -48,9 +48,17 @@ let maybe_fetch_compatibility_metadata ~__context ~pgpu_pci =
     []
 
 let populate_compatibility_metadata ~__context ~pgpu ~pgpu_pci =
-  let () = Db.PGPU.set_compatibility_metadata ~__context ~self:pgpu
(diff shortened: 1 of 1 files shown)


lindig:
This was cherry-picked from the vGPU feature branch where it was reviewed and merged. 
https://github.com/xapi-project/xen-api/pull/3392

PR #3397, opened 2018-01-10
static-vdis: need to save/pass uuid to SR.attach with SMAPIv3

edwintorok:
This is required for HA to work correctly after the SMAPIv3 API change that added `uuid` to `SR.attach`, otherwise we fail while enabling HA because we fail to attach the statefile VDI.

Tested on a pool of 2 hosts and I was able to enable HA.


edwintorok:
There is an SR.attach in the SMAPIv2 example script, but we didn't change the SMAPIv2 API (it already had the SR uuid), just the SMAPIv3 one.

PR #3398, opened 2018-01-10
Host.set_iscsi_iqn: reject empty string

gaborigloi:
Because it's invalid.

### DIFF
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -1762,6 +1762,7 @@ let allocate_resources_for_vm ~__context ~self ~vm ~live =
   ()
 
 let set_iscsi_iqn ~__context ~host ~value =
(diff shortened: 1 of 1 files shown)


gaborigloi:
i should add a nit t est too

ocaml/xapi/xapi_host.ml
```diff
@@ -1762,6 +1762,7 @@ let allocate_resources_for_vm ~__context ~self ~vm ~live =
   ()
 
 let set_iscsi_iqn ~__context ~host ~value =
+  if value = "" then raise Api_errors.(Server_error (invalid_value, ["value"; value]));
```
edwintorok: maybe make the error description a bit more explicit that the iqn is wrong, or will that be obvious from the API call itself?

gaborigloi: Yes, I chose this as this seems to be the standard error we throw in these cases. I specified here the iqn param name, `"value"`, as that seems to be the convention for this error param.

PR #3399, opened 2018-01-11
CP-26333 Add PIF.PCI field

krizex:
1. Set PIF.pci of physical PIF to corresponding pci ref when creating
2. Set PIF.pci of vlan master PIF to the PIF.pci of the underlying PIF
3. Set PIF.pci of Bond master or tunnel PIF to Ref.null
4. Refresh PIF.pci on xapi restarting for all kinds of PIFs

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -5659,6 +5659,7 @@ let pif =
       field ~lifecycle:[Published, rel_inverness, ""] ~qualifier:DynamicRO ~ty:pif_igmp_status ~default_value:(Some (VEnum "unknown")) "igmp_snooping_status" "The IGMP snooping status of the corresponding network bridge";
       field ~in_oss_since:None ~ty:(Set (Ref _network_sriov)) ~in_product_since:rel_kolkata ~qualifier:DynamicRO "sriov_physical_PIF_of" "Indicates which network_sriov this interface is physical of";
       field ~in_oss_since:None ~ty:(Set (Ref _network_sriov)) ~in_product_since:rel_kolkata ~qualifier:DynamicRO "sriov_logical_PIF_of" "Indicates which network_sriov this interface is logical of";
+      field ~qualifier:DynamicRO ~ty:(Ref _pci) ~lifecycle:[Published, rel_kolkata, ""] ~default_value:(Some (VRef null_ref)) "PCI" "Link to underlying PCI device";
     ]
     ()
 
--- a/ocaml/xapi/xapi_pif.ml
+++ b/ocaml/xapi/xapi_pif.ml
@@ -23,14 +23,23 @@ open Listext
 open Pervasiveext
 open Xstringext
 open Threadext
-
 open Network
+
+let get_device_pci ~__context ~host ~device =
+  let dbg = Context.string_of_task __context in
+  let pci_bus_path = Net.Interface.get_pci_bus_path dbg ~name:device in
+  let expr = Db_filter_types.(And (Eq (Field "pci_id", Literal (pci_bus_path)),
+                                   Eq (Field "host", Literal (Ref.string_of host)))) in
+  match Db.PCI.get_refs_where ~__context ~expr with
+  | pci :: _ -> pci
+  | _  -> Ref.null
 
 let refresh_internal ~__context ~self =
   let device = Db.PIF.get_device ~__context ~self in
   let network = Db.PIF.get_network ~__context ~self in
   let bridge = Db.Network.get_bridge ~__context ~self:network in
   let dbg = Context.string_of_task __context in
+  let host = Db.PIF.get_host ~__context ~self in
 
   (* Update the specified PIF field in the database, if
      	 * and only if a corresponding value can be read from
@@ -57,6 +66,12 @@ let refresh_internal ~__context ~self =
       (Db.PIF.set_MAC)
       (fun () -> Net.Interface.get_mac dbg ~name:device)
       (id);
+
+  maybe_update_database "PCI"
+    (Db.PIF.get_PCI)
+    (Db.PIF.set_PCI)
+    (fun () -> get_device_pci ~__context ~host ~device)
+    (Ref.string_of);
 
   maybe_update_database "MTU"
     (Db.PIF.get_MTU)
@@ -374,7 +389,7 @@ let pool_introduce
       ~ip_configuration_mode ~iP ~netmask ~gateway ~dNS
       ~bond_slave_of:Ref.null ~vLAN_master_of ~management
       ~other_config ~disallow_unplug ~ipv6_configuration_mode
-      ~iPv6 ~ipv6_gateway ~primary_address_type ~managed ~properties ~capabilities:[] in
+      ~iPv6 ~ipv6_gateway ~primary_address_type ~managed ~properties ~capabilities:[] ~pCI:Ref.null in
   pif_ref
 
 let db_introduce = pool_introduce
@@ -400,6 +415,7 @@ let introduce_internal
   in
   let dbg = Context.string_of_task __context in
   let capabilities = Net.Interface.get_capabilities dbg device in
+  let pci = get_device_pci ~__context ~host ~device in
 
   let pif = Ref.make () in
   debug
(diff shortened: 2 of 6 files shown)


krizex:
Travis check failed because it cannot get the corresponding correct `xcp-idl` package, I think the RC of the problem is we have not get a successful build yet.

robhoes:
I believe we had `PIF.PCI` (in capitals) in the design, consistent with e.g. `PGPU.PCI`.

ocaml/xapi/xapi_pif.ml
```diff
@@ -23,14 +23,27 @@ open Listext
 open Pervasiveext
 open Xstringext
 open Threadext
-
 open Network
 
+let get_device_pci ~__context ~host ~device =
+  let open Network_stats in
+  let stats = read_stats () in
+  if List.mem_assoc device stats then begin
```
mseri: We tend to do
```ocaml
try
  let stat = List.assoc device stats in
  ...
with Not_found -> Ref.null
```
in many other places to avoid the double lookup

krizex: ok

krizex:
Updated.

ocaml/xapi/xapi_vlan.ml
```diff
@@ -31,14 +31,15 @@ let create_internal ~__context ~host ~tagged_PIF ~tag ~network ~device =
...
+  let pci = Db.PIF.get_PCI ~__context ~self:tagged_PIF in
   Db.PIF.create ~__context ~ref:untagged_PIF ~uuid:(Uuid.to_string (Uuid.make_uuid ()))
     ~device ~device_name:device ~network ~host ~mAC:vlan_mac ~mTU ~vLAN:tag ~metrics
     ~physical:false ~currently_attached:false ~igmp_snooping_status:`unknown
     ~ip_configuration_mode:`None ~iP:"" ~netmask:"" ~gateway:"" ~dNS:"" ~bond_slave_of:Ref.null
     ~vLAN_master_of:vlan ~management:false
     ~other_config:[] ~disallow_unplug:false
     ~ipv6_configuration_mode:`None ~iPv6:[""] ~ipv6_gateway:"" ~primary_address_type:`IPv4 ~managed:true
-    ~properties:[] ~capabilities:[];
+    ~properties:[] ~capabilities:[] ~pCI:pci;
```
robhoes: I think that we should leave it NULL for any non-physical PIF (as for the bond master and tunnel access PIF above).

ocaml/idl/datamodel.ml
```diff
@@ -5659,6 +5659,7 @@ let pif =
       field ~lifecycle:[Published, rel_inverness, ""] ~qualifier:DynamicRO ~ty:pif_igmp_status ~default_value:(Some (VEnum "unknown")) "igmp_snooping_status" "The IGMP snooping status of the corresponding network bridge";
       field ~in_oss_since:None ~ty:(Set (Ref _network_sriov)) ~in_product_since:rel_kolkata ~qualifier:DynamicRO "sriov_physical_PIF_of" "Indicates which network_sriov this interface is physical of";
       field ~in_oss_since:None ~ty:(Set (Ref _network_sriov)) ~in_product_since:rel_kolkata ~qualifier:DynamicRO "sriov_logical_PIF_of" "Indicates which network_sriov this interface is logical of";
+      field ~qualifier:StaticRO ~ty:(Ref _pci) ~lifecycle:[Published, rel_kolkata, ""] ~default_value:(Some (VRef null_ref)) "PCI" "Link to underlying PCI device";
```
robhoes: This should be `DynamicRO`.

krizex: yes

ocaml/xapi/xapi_vlan.ml
krizex: yes

krizex:
1. Update code per comments
2. Accordingly update `PIF.PCI` when updating `PIF_metric.pci_bus_path`

robhoes:
@krizex, what was the reason for adding that last patch? Now there are two places where `PIF.PCI` is set: the PIF refresh code (triggered by dbsync) and the monitor code. Is the former not enough?

krizex:
@robhoes The scenario is when executing dbsync and at that moment networkd haven't wrote the device information to `/dev/shm/network_stats` yet, we will not get the correct PCI device info for PIF.
So I suggest to just keep the latter one(in the monitor code) to set the `PIF.PCI` because it triggers every 5 secs.

robhoes:
Doing this from the monitor code does not feel right, because this is not metadata that changes while xapi is running, so looking to update this everything 5s is unnecessary.

I think it is better to not rely on the network stats for this, but rather add an explicit call to network to get the PCI address, like we have explicit calls for the MTU and MAC (i.e. expose `Sysfs.get_pcibuspath`).

We do have to consider the order in which dbsync does things: it need to call `Xapi_pci.update_pcis` to create all PCI objects before `Xapi_pif.refresh_all`. This means that you need to move the call to `Xapi_pci.update_pcis` in `Dsync_slave.update_env` up to before `Xapi_pif.refresh_all`.

krizex:
@robhoes
In a word, I think we should implement a new IDL interface `Network.Interface.get_pci_bus_path` to get the pci bus path for the PIF device.

At the same time we should make the change as you said:
> This means that you need to move the call to Xapi_pci.update_pcis in Dsync_slave.update_env up to before Xapi_pif.refresh_all

The reasons is getting pci bus path from `/dev/shm/network_stats` is not reliable.
Because of the pci device info in the stats file `/dev/shm/network_stats` is written by networkd, so supposing when xapi query the info from networkd we should get the correct info. Unfortunately, though in systemd configuration file `xapi.service` should start after `xcp-networkd.service` and the start type of `xcp-networkd.service` is `Notify`, we still cannot ensure when xapi querying networkd, the stats file has been completed written because the stuff belongs to one thread of xcp-networkd and when xcp-networkd notifying systemd it is ready, we just can ensure the thread is started but cannot ensure the stats file has been written.

Actions:
1. create a new commit of move the call to `Xapi_pci.update_pcis` in `Dsync_slave.update_env` up to before `Xapi_pif.refresh_all`
2. Add new IDL interface `Network.Interface.get_pci_bus_path`

krizex:
Updated.
1. Move the call to Xapi_pci.update_pcis in Dsync_slave.update_env up to before Xapi_pif.refresh_all
2. Add new IDL interface Network.Interface.get_pci_bus_path and use it to get the pci bus path. 

This PR depends on https://github.com/xapi-project/xcp-idl/pull/188 and https://github.com/xapi-project/xcp-networkd/pull/125

ocaml/xapi/xapi_pif.ml
```diff
@@ -26,16 +26,12 @@ open Threadext
...
-  let stats = read_stats () in
-  try
-    let stat = List.assoc device stats in
-    let expr = Db_filter_types.(And (Eq (Field "pci_id", Literal (stat.pci_bus_path)),
-                                     Eq (Field "host", Literal (Ref.string_of host)))) in
-    match Db.PCI.get_refs_where ~__context ~expr with
-    | [pci] -> pci
-    | _  -> Ref.null
-  with Not_found -> Ref.null
+  let pci_bus_path = Net.Interface.get_pci_bus_path "get_pci_bus_path" ~name:device in
```
robhoes: Use `let dbg = Context.string_of_task __context in` for the debug-info in this call (see below in `refresh_internal`.

ocaml/xapi/xapi_pif.ml
```diff
@@ -26,16 +26,12 @@ open Threadext
...
+  let expr = Db_filter_types.(And (Eq (Field "pci_id", Literal (pci_bus_path)),
+                                   Eq (Field "host", Literal (Ref.string_of host)))) in
+  match Db.PCI.get_refs_where ~__context ~expr with
+  | [pci] -> pci
```
robhoes: Perhaps we can relax this a little by doing `| pci :: _` (i.e. just take the head of the list).

robhoes:
@krizex, please fix the two small things above, squash the patches, and update the commit message. Then it can go in after the xcp-idl/networkd PRs.

ocaml/xapi/xapi_pif.ml
krizex: sure

krizex:
Updated code per commit.
Squashed commit and update commit message accordingly.

ocaml/xapi/xapi_pif.ml
robhoes: It's just so that it then also covers the case where there happen to be more than one matching PCI object rather than exactly one. This should not happen in practice, but on the other hand, there should be no reason to return NULL in that case.

krizex: Ah yes

PR #3401, opened 2018-01-12
CA-279502: Set the correct release for the is_default_template field

mseri:


ocaml/idl/datamodel.ml
```diff
@@ -1892,7 +1892,7 @@ let vm_create_template = call
 let vm_set_is_default_template = call
     ~name:"set_is_default_template"
     ~hide_from_docs:true
-    ~lifecycle: [Published, rel_ely, "Allows to define XenServer default templates"]
+    ~lifecycle: [Published, rel_falcon, "Allows to define XenServer default templates"]
```
kc284: I believe we also need to change the version for the corresponding VM field as well.

mseri: Oops, right. Updated, I did not find other appearances of it in the datamodel other than the VM.

PR #3402, opened 2018-01-12
Add suite with alcotest

gaborigloi:
And test Valid_ref_list using alcotest.

### DIFF
--- a/ocaml/xapi/jbuild
+++ b/ocaml/xapi/jbuild
@@ -68,11 +68,12 @@ let () = Printf.ksprintf Jbuild_plugin.V1.send {|
 )
 
 (executables
- ((names (xapi_main suite quicktest))
-  (public_names (xapi - quicktestbin))
+ ((names (xapi_main suite suite_alcotest quicktest))
+  (public_names (xapi - - quicktestbin))
   (package xapi)
   (flags (:standard -bin-annot %s -warn-error +a-3-4-6-9-27-28-29-52))
   (libraries (
+   alcotest
    opasswd
    pam
    pciutil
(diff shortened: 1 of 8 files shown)


ocaml/xapi/jbuild
```diff
@@ -112,4 +113,11 @@ let () = Printf.ksprintf Jbuild_plugin.V1.send {|
   (action (run ${<} -runner sequential -verbose true))
  )
 )
+
+(alias
+ ((name runalcotest)
```
edwintorok: I think you can make both of these run as part of runtest if you put alcotest in another folder

gaborigloi: How can I do that?

edwintorok: See https://discuss.ocaml.org/t/jbuilder-support-for-multiple-test-executables/825 you can in fact have two `runtest` in same jbuild file.

gaborigloi: Thanks, that's great, I'll use two!

lindig:
xs-opam has been released. Can we merge this?

mseri:
Let's rerun travis first, once we have a build with the next `xs-opam` uploaded to the xs-buildenv

mseri:
Same issue as `message-switch`... There is finally a plan upstream to fix the `ppx_bisect`/coverage situation. Until that is complete you only need to install by hand the test dependencies in the `.coverage` script. It's a matter of adding `alcotest` to the `opam install -y oUnit ...` line

gaborigloi:
How about using `opam install --build-test`?

mseri:
Try. It did not work when I tried on message-switch.

gaborigloi:
Oh, that doesn't work because that tries to install the test deps of the dependencies transitively

PR #3406, opened 2018-01-18
CP-25699 Network sriov create/destroy

krizex:
1. Implement network sriov create/destroy
2. Call idl interface to enable/disable network sriov when plug/unplug
3. Update pci status after sriov plug/unplug
4. Check SRIOV pci compatibility before adding it to network
5. When disabling sriov, only call networkd when all the PIFs that use same
driver in the host has disabled sriov.

---
This change is [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3406)


robhoes:
I haven't reviewed this yet, but my initial comment would be that this looks like a lot of code for just a single commit... The commit message describes five items, which looks like a good way to split things up into more manageable patches.

krizex:
I can do that, but maybe cannot ensure that every patch could be able to compile...

krizex:
Split the code changes into 2 patches:
1. Patch about refactor out PIF type related stuffs.
2. Network sriov create/destroy implementation.

krizex:
Update:
Fix some corner case issues.
Add unit tests.

cheng-z:
Hi @robhoes 
Could you please help to review this PR as soon as possible?
It is now blocking our feature development for other PRs.
Thanks.

ocaml/idl/datamodel.ml
```diff
@@ -562,6 +562,19 @@ let _ =
   error Api_errors.user_is_not_local_superuser [ "msg" ]
     ~doc:"Only the local superuser can execute this operation" ();
 
+  (* SR-IOV errors *)
+  error Api_errors.network_sriov_already_enabled ["PIF"]
+    ~doc:"You tried to enable network SR-IOV on PIF which is already enabled" ();
```
mseri: I would just write `"The PIF selected for the SR-IOV network is already enabled."`

ocaml/idl/datamodel.ml
```diff
@@ -562,6 +562,19 @@ let _ =
...
+  error Api_errors.network_sriov_find_pf_from_vf_failed ["PCI"]
+    ~doc:"Failed to find physical function from virtual function" ();
+  error Api_errors.network_sriov_pif_has_vf_inuse ["PIF"; "VFs"]
+    ~doc:"Can not destroy network sriov object which has virtual function inuse" ();
```
mseri: `"The SR-IOV network object has virtual functions in use."`

ocaml/idl/datamodel.ml
```diff
@@ -562,6 +562,19 @@ let _ =
...
+  error Api_errors.network_sriov_enable_failed ["PIF"; "msg"]
+    ~doc:"Can not enable network SR-IOV on PIF" ();
```
mseri: `Can not -> Cannot`, but I would actually write `"Failed to enable SR-IOV network."`

ocaml/idl/datamodel.ml
```diff
@@ -562,6 +562,19 @@ let _ =
...
+  error Api_errors.network_sriov_disable_failed ["PIF"; "msg"]
+    ~doc:"Can not disable network SR-IOV on PIF" ();
```
mseri: `Can not -> Cannot`, but I would actually write `"Failed to disable SR-IOV network."`

ocaml/idl/datamodel.ml
```diff
@@ -562,6 +562,19 @@ let _ =
...
+  error Api_errors.network_sriov_pci_vendor_not_compatible ["PIF"; "network"]
+    ~doc:"Can not add incompatible network SR-IOV PIF into the network" ();
```
mseri: I would write `"The PIF is not compatible with the selected SR-IOV network."`

ocaml/idl/datamodel.ml
```diff
@@ -573,9 +586,12 @@ let _ =
     ~doc:"You tried to add a purpose to a network but the new purpose is not compatible with an existing purpose of the network or other networks." ();
   error Api_errors.pif_is_physical ["PIF"]
     ~doc:"You tried to destroy a PIF, but it represents an aspect of the physical host configuration, and so cannot be destroyed.  The parameter echoes the PIF handle you gave." ();
+  error Api_errors.pif_is_not_physical ["PIF"]
+    ~doc:"You tried to apply the operation which is only available on physical PIF" ();
```
mseri: `"You tried to perform an operation which is only available on physical PIF"`

ocaml/idl/datamodel.ml
```diff
@@ -608,6 +624,8 @@ let _ =
     ~doc:"The operation you requested cannot be performed because the specified PIF has FCoE SR in use." ();
   error Api_errors.pif_unmanaged [ "PIF" ]
     ~doc:"The operation you requested cannot be performed because the specified PIF is not managed by xapi." ();
+  error Api_errors.pif_is_not_sriov_capable [ "PIF" ]
+    ~doc:"You try to enable network SR-IOV on an incapable PIF" ();
```
mseri: `try -> tried` but I would write something like `"The selected PIF is not capable of network SR-IOV."`

ocaml/xapi/test_vlan.ml
```diff
@@ -170,10 +190,10 @@ let test =
     "test_create_internal" >:: test_create_internal;
     "test_create_unmanged_pif" >:: test_create_unmanaged_pif;
     "test_create_network_already_connected" >:: test_create_network_already_connected;
-    "test_create_pif_not_a_bond_slave" >:: test_create_pif_not_a_bond_slave;
-    "test_create_pif_not_vlan_slave" >:: test_create_pif_not_vlan_slave;
+    "test_create_pif_is_bond_slave" >:: test_create_pif_is_bond_slave;
```
mseri: Why did you rename these tests names?
Shouldn't we add the new tests in addition to the old ones instead?

ocaml/xapi/xapi_network_sriov.ml
```diff
@@ -0,0 +1,61 @@
...
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+open Stdext
```
mseri: Please do not do that. Instead do:
```
open Xapi_stdext_std
```

and then in the code prepend `Listext` only when needed.

To be honest I don't see any use of `Stdext` in this module, so you should actually remove these two lines.

ocaml/xapi/xapi_network_sriov_helpers.ml
```diff
@@ -0,0 +1,171 @@
...
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+open Stdext
+open Listext
```
mseri: Please do not do that. Instead do:
```
open Xapi_stdext_std
```

and then in the code prepend `Listext` only when needed.

To be honest I don't see any use of `Stdext` in this module, so you should actually remove these two lines.

ocaml/xapi/xapi_network_sriov_helpers.ml
```diff
@@ -0,0 +1,171 @@
...
+open Threadext
```
mseri: use `open Xapi_stdext_threads` instead and prepend the `Threadext` where needed, I think we use `execute` just once in this module

robhoes:
Sorry @krizex, but this PR is pretty much unreviewable right now. There are a couple of separate commits, but with all the fixups in top, all we can do now is review the entire diff, which is too large to make sense of. Could you please start by squashing the fixups?

ocaml/xapi/test_vlan.ml
krizex: I am renaming the test with correct name, for example, `test_create_pif_not_a_bond_slave` the original writer aims to test the scenario: cannot create vlan on a bond slave PIF. The test case actually test creating vlan on a PIF which is a bond slave, so I rename it to `test_create_pif_is_bond_slave`

krizex:
I have splitted the code changes into commits with size ~200 lines each one.
The biggest one is to re-struct the code so it is ~370 lines, but don't worry the effective code change I made is ~100 lines.

robhoes:
Reviewed 1 of 12 files at r5.
Review status: 0 of 16 files reviewed at latest revision, 27 unresolved discussions, some commit checks failed.

---

*[ocaml/xapi/nm.ml, line 239 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4Gz9fC_ctQb-MySvSt:-L4Gz9fC_ctQb-MySvSu:brtqnpg) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/nm.ml#L239)):*
> ```OCaml
>   let persistent = is_dom0_interface pif_rc in
>   let igmp_snooping = Some (Db.Pool.get_igmp_snooping_enabled ~__context ~self:(Helpers.get_pool ~__context)) in
>   let pif_ref = Db.PIF.get_by_uuid ~__context ~uuid:pif_rc.API.pIF_uuid in
> ```

The caller of `create_bridges`, which is just `bring_pif_up`, knows the PIF's ref, and should pass it as an argument to avoid a DB lookup.

---

*[ocaml/xapi/nm.ml, line 240 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H3KJBzUADOZd6i5n0:-L4H3KJBzUADOZd6i5n1:b-x4h2tm) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/nm.ml#L240)):*
> ```OCaml
>   let igmp_snooping = Some (Db.Pool.get_igmp_snooping_enabled ~__context ~self:(Helpers.get_pool ~__context)) in
>   let pif_ref = Db.PIF.get_by_uuid ~__context ~uuid:pif_rc.API.pIF_uuid in
>   match get_pif_type ~__context ~self:pif_ref with
> ```

Well, it may be better to pass the PIF record to `get_pif_type` instead (see below).

---

*[ocaml/xapi/nm.ml, line 246 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4GzgP5RfELbxqeM8QL:-L4GzgP5RfELbxqeM8QM:b4m5h8o) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/nm.ml#L246)):*
> ```OCaml
>                                                      igmp_snooping; other_config; persistent_b=persistent}],
>     []
>   | VLAN_master_on_physical _ ->
> ```

This case must cover the VLAN-on-bond case as well. Does it?

---

*[ocaml/xapi/nm.ml, line 247 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H3u6xuZBcz8kA39rD:-L4H3u6xuZBcz8kA39rE:bysr57g) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/nm.ml#L247)):*
> ```OCaml
>     []
>   | VLAN_master_on_physical _ ->
>     let vlan = pif_rc.API.pIF_VLAN_master_of in
> ```

Can't you get this from the match cases? Is seems that `VLAN_master_on_physical` has the VLAN as its argument.

---

*[ocaml/xapi/nm.ml, line 283 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4GzzeSxczfGlnHr25J:-L4GzzeSxczfGlnHr25K:b7azkx6) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/nm.ml#L283)):*
> ```OCaml
> 
> let rec destroy_bridges ~__context ~force pif_rc bridge =
>   let pif_ref = Db.PIF.get_by_uuid ~__context ~uuid:pif_rc.API.pIF_uuid in
> ```

Again, `bring_pif_down` should pass in the ref to avoid the DB call.

---

*[ocaml/xapi/nm.ml, line 287 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H-4b_D9jbOed80F-F:-L4H-4b_D9jbOed80F-G:bs4bkzc) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/nm.ml#L287)):*
> ```OCaml
>   | Tunnel_access _ ->
>     [bridge, false]
>   | VLAN_master_on_physical _ ->
> ```

Again, does this cover the VLAN-on-bond case?

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 25 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H0ua5cQDzIZyvA39h:-L4H0ua5cQDzIZyvA39i:b-klrb7l) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L25)):*
> ```OCaml
>   | Bond_master of ref_Bond
>   | Network_sriov_logical of ref_network_sriov
>   | VLAN_master_on_physical of ref_PIF
> ```

Ah, so "physical" includes bonds...?

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 26 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H8izOaHNsGFAgnnO6:-L4H8izOaHNsGFAgnnO7:b-pa07n) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L26)):*
> ```OCaml
>   | Network_sriov_logical of ref_network_sriov
>   | VLAN_master_on_physical of ref_PIF
>   | VLAN_master_on_network_sriov of ref_PIF
> ```

I find these two a little awkward, because the type encodes "something" at the same level as "something on something else". This could easily get out of hand. E.g. we might need a `Tunnel_on_VLAN_on_bond` at some point. I'm also worried about missing cases in this model, in the way that I was confused about the `VLAN_master_on_physical`.

Another option may be to have just `Tunnel | VLAN | Bond | SRIOV | Physical` in this list, and have a `get_pif_type` function return a list, e.g. `[VLAN; SRIOV; Physical]`, or `[Tunnel; VLAN; Bond; Physical]` (it's like the original `get_pif_type` called recursively). Or, to make the "slave of" relation explicit, a function that tells us "what's above us" in addition to this (seems to be useful only for bond-slave checks).

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 27 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H8YN31vUOLTltHRG0:-L4H8YN31vUOLTltHRG1:b-krzpkr) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L27)):*
> ```OCaml
>   | VLAN_master_on_physical of ref_PIF
>   | VLAN_master_on_network_sriov of ref_PIF
>   | Other_PIF
> ```

I'm missing the physical PIF here. Or is that what `Other_PIF` is? Or does `Other_PIF` include "I don't know; something must have gone wrong"?

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 29 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H1JuuLpdBBOUD7CTn:-L4H1JuuLpdBBOUD7CTo:b-pbwtn8) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L29)):*
> ```OCaml
>   | Other_PIF
> 
> let get_pif_type ~__context ~self =
> ```

This function does lots of DB-gets. On a pool slave, each on costs one round-trip to the master. This is why a PIF record is passed around in modules like `Nm`: you need to get the record from the master just once. It would be better to do the same here.

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 32 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H4AWKDXOJwtSwgaep:-L4H4AWKDXOJwtSwgaeq:b-awxglu) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L32)):*
> ```OCaml
>   (* Define a bind func to build the pipe for the monad.*)
>   (* We should ensure any 2 func that used in the pipeline should not return `Some` at the same time.*)
>   let (>>=) opt f =
> ```

Interesting monadic experiment :)

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 69 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H4f6cOPoCQrcsivBS:-L4H4f6dDVqcmYFNk6ID:b-8gymim) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L69)):*
> ```OCaml
>     >>= is_vlan_master_pif with
>   | Some x -> x
>   | None -> Other_PIF
> ```

This should be treated as an error, isn't it?

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 72 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H5cs0lcm7J4hpFXWp:-L4H5cs0lcm7J4hpFXWq:b4drnu0) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L72)):*
> ```OCaml
> 
> let vlan_is_allowed_on_pif ~__context ~tagged_PIF ~tag =
>   if Db.PIF.get_managed ~__context ~self:tagged_PIF <> true then
> ```

This is better put in an `assert_pif_is_managed` function. In fact, xapi_pif.ml has one of those, which we can move into this module, and call directly from xapi_vlan.ml and friends. It should not be part of the "allowed on" checks.

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 87 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H57sFT5ArrLe1qfOt:-L4H57sFT5ArrLe1qfOu:bjwy6zw) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L87)):*
> ```OCaml
>   end;
> 
>   (* Check the VLAN tag is sensible;  4095 is reserved for implementation use (802.1Q) *)
> ```

I would leave the checks below in `Xapi_vlan`, because they aren't part of the "allowed on" check. It's better to keep this module focussed on the latter.

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 100 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H6r1AIetk_F09ZMbx:-L4H6r1AIetk_F09ZMby:b-34vmyd) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L100)):*
> ```OCaml
> 
> let tunnel_is_allowed_on_pif ~__context ~transport_PIF =
>   if Db.PIF.get_managed ~__context ~self:transport_PIF <> true then
> ```

See above

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 110 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4HEHc_tXOkjWiCKeT2:-L4HEHcaAPHicSaiYVA9:b-x9pw6n) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L110)):*
> ```OCaml
>     raise (Api_errors.Server_error (Api_errors.cannot_add_tunnel_to_sriov_logical, [Ref.string_of transport_PIF]))
>   | VLAN_master_on_network_sriov _ ->
>     raise (Api_errors.Server_error (Api_errors.cannot_add_tunnel_to_vlan_on_sriov_logical, [Ref.string_of transport_PIF]))
> ```

Although it is a huge improvement compared to what we used to have, I think these functions are still a little error prone, because it is hard to spot mistakes and confirm that the definitions are complete (it is not "obviously correct"). I was hoping for a more direct way of expressing allowed relations, and use the same definition for all checks. E.g. things like "a bond can only sit on top of physical PIFs" and "a VLAN can only sit on top of physical, bond or SRIOV". Or more interesting, "a tunnel can sit on a VLAN unless the VLAN sits on top of an SRIOV PIF".

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3406)*

krizex:
Review status: 0 of 16 files reviewed at latest revision, 27 unresolved discussions, some commit checks failed.

---

*[ocaml/idl/datamodel.ml, line 567 at r4](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L3WWtp7Z3hZPD1dA_Pi-r4-567:-L4JP-EeNvdWpIm9ToKL:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/17e79d704509bae033e4ac869e44819056efe472/ocaml/idl/datamodel.ml#L567)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

I would just write `"The PIF selected for the SR-IOV network is already enabled."`
</blockquote></details>

Done.

---

*[ocaml/idl/datamodel.ml, line 571 at r4](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L3WWtp7Z3hZPD1dA_Pi-r4-571:-L4JOyvRUfGUYagZxvo5:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/17e79d704509bae033e4ac869e44819056efe472/ocaml/idl/datamodel.ml#L571)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

`"The SR-IOV network object has virtual functions in use."`
</blockquote></details>

Done.

---

*[ocaml/idl/datamodel.ml, line 573 at r4](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L3WWtp7Z3hZPD1dA_Pi-r4-573:-L4JOxwvvVXICnzzG4xJ:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/17e79d704509bae033e4ac869e44819056efe472/ocaml/idl/datamodel.ml#L573)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

`Can not -> Cannot`, but I would actually write `"Failed to enable SR-IOV network."`
</blockquote></details>

Done.

---

*[ocaml/idl/datamodel.ml, line 575 at r4](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L3WWtp7Z3hZPD1dA_Pi-r4-575:-L4JP0V4VSl-3D-_0DkZ:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/17e79d704509bae033e4ac869e44819056efe472/ocaml/idl/datamodel.ml#L575)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

`Can not -> Cannot`, but I would actually write `"Failed to disable SR-IOV network."`
</blockquote></details>

Done.

---

*[ocaml/idl/datamodel.ml, line 577 at r4](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L3WWtp7Z3hZPD1dA_Pi-r4-577:-L4JP1JLOH0KZC2skYqw:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/17e79d704509bae033e4ac869e44819056efe472/ocaml/idl/datamodel.ml#L577)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

I would write `"The PIF is not compatible with the selected SR-IOV network."`
</blockquote></details>

Done.

---

*[ocaml/idl/datamodel.ml, line 590 at r4](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L3WWtp7Z3hZPD1dA_Pi-r4-590:-L4JP3nXrAqOFUeac3Pn:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/17e79d704509bae033e4ac869e44819056efe472/ocaml/idl/datamodel.ml#L590)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

`"You tried to perform an operation which is only available on physical PIF"`
</blockquote></details>

Done.

---

*[ocaml/idl/datamodel.ml, line 628 at r4](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L3WWtp7Z3hZPD1dA_Pi-r4-628:-L4JP59rmJSV5AOaiqVt:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/17e79d704509bae033e4ac869e44819056efe472/ocaml/idl/datamodel.ml#L628)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

`try -> tried` but I would write something like `"The selected PIF is not capable of network SR-IOV."`
</blockquote></details>

Done.

---

*[ocaml/xapi/nm.ml, line 239 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4Gz9fC_ctQb-MySvSt:-L4JRns2bqPIh5YRmrA3:b2nhswy) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/nm.ml#L239)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

The caller of `create_bridges`, which is just `bring_pif_up`, knows the PIF's ref, and should pass it as an argument to avoid a DB lookup.
</blockquote></details>

Yes

---

*[ocaml/xapi/nm.ml, line 240 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H3KJBzUADOZd6i5n0:-L4JwvvDd-z0m9eFNTUz:b2nign6) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/nm.ml#L240)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

Well, it may be better to pass the PIF record to `get_pif_type` instead (see below).
</blockquote></details>

yes

---

*[ocaml/xapi/nm.ml, line 246 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4GzgP5RfELbxqeM8QL:-L4JQQjnDkgz6Rcd1Kyy:b-tkkop) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/nm.ml#L246)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

This case must cover the VLAN-on-bond case as well. Does it?
</blockquote></details>

Yes, but I think it could be better to introduce a new type `VLAN_master_on_bond` to make the type definition clearer.

---

*[ocaml/xapi/nm.ml, line 247 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H3u6xuZBcz8kA39rD:-L4JRAgzM26dLuWCZ8ze:b1jv8bd) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/nm.ml#L247)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

Can't you get this from the match cases? Is seems that `VLAN_master_on_physical` has the VLAN as its argument.
</blockquote></details>

No `VLAN_master_on_physical` does not has VLAN as its argument, but I think it's better to have. So I will refine the definition of `type pif_type_t` to make all types not only contain the pif_ref but also the bond/tunnel/network_sriov/vlan object.

---

*[ocaml/xapi/nm.ml, line 283 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4GzzeSxczfGlnHr25J:-L4JRqhW4fo27ZD_uwkR:b2nhswy) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/nm.ml#L283)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

Again, `bring_pif_down` should pass in the ref to avoid the DB call.
</blockquote></details>

Yes

---

*[ocaml/xapi/nm.ml, line 287 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H-4b_D9jbOed80F-F:-L4JSH2xD7jFbX3NGULB:b85dzpd) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/nm.ml#L287)):*
> Yes, but I think it could be better to introduce a new type `VLAN_master_on_bond` to make the type definition clearer.

---

*[ocaml/xapi/xapi_network_sriov.ml, line 15 at r4](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L3WWtp7Z3hZPD1dA_Pq-r4-15:-L4JSNL6eBUKwZCOPRL9:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/17e79d704509bae033e4ac869e44819056efe472/ocaml/xapi/xapi_network_sriov.ml#L15)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

Please do not do that. Instead do:
```
open Xapi_stdext_std
```

and then in the code prepend `Listext` only when needed.

To be honest I don't see any use of `Stdext` in this module, so you should actually remove these two lines.
</blockquote></details>

Done.

---

*[ocaml/xapi/xapi_network_sriov_helpers.ml, line 16 at r4](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L3WWtp7Z3hZPD1dA_Pr-r4-16:-L4JSP4USICKy1wVAaMP:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/17e79d704509bae033e4ac869e44819056efe472/ocaml/xapi/xapi_network_sriov_helpers.ml#L16)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

Please do not do that. Instead do:
```
open Xapi_stdext_std
```

and then in the code prepend `Listext` only when needed.

To be honest I don't see any use of `Stdext` in this module, so you should actually remove these two lines.
</blockquote></details>

Done.

---

*[ocaml/xapi/xapi_network_sriov_helpers.ml, line 17 at r4](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L3WWtp7Z3hZPD1dA_Pr-r4-17:-L4JSPWx0DFQFVqnUEVl:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/17e79d704509bae033e4ac869e44819056efe472/ocaml/xapi/xapi_network_sriov_helpers.ml#L17)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

use `open Xapi_stdext_threads` instead and prepend the `Threadext` where needed, I think we use `execute` just once in this module
</blockquote></details>

Done.

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 25 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H0ua5cQDzIZyvA39h:-L4JSRU3BwRIrEcyF5pQ:b-tkkop) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L25)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

Ah, so "physical" includes bonds...?
</blockquote></details>

Yes, but I think it could be better to introduce a new type `VLAN_master_on_bond` to make the type definition clearer.

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 26 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H8izOaHNsGFAgnnO6:-L4Jxejg3bvZ7swIYnWp:b-16fuj4) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L26)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

I find these two a little awkward, because the type encodes "something" at the same level as "something on something else". This could easily get out of hand. E.g. we might need a `Tunnel_on_VLAN_on_bond` at some point. I'm also worried about missing cases in this model, in the way that I was confused about the `VLAN_master_on_physical`.

Another option may be to have just `Tunnel | VLAN | Bond | SRIOV | Physical` in this list, and have a `get_pif_type` function return a list, e.g. `[VLAN; SRIOV; Physical]`, or `[Tunnel; VLAN; Bond; Physical]` (it's like the original `get_pif_type` called recursively). Or, to make the "slave of" relation explicit, a function that tells us "what's above us" in addition to this (seems to be useful only for bond-slave checks).
</blockquote></details>

Yes, the type definition is incomplete and I found it is not able to be complete because we should keep each 2 of them is exclusive.
So after write the relation tree about the network related object (VLAN/Tunnel/Bond/SRIOV), I think we can implement a FSM to get the attribute of the PIF by checking the network related object underneath.
The relation tree is: (lowercase name is non-terminal symbol and uppercase name is terminal symbol)
```
   tunnel : vlan
          | PHYSICAL
          | BOND_MASTER

   vlan : vlan
        | sriov
        | BOND_MASTER
        | PHYSICAL

   sriov : BOND_SLAVE
         | PHYSICAL
```
So for any given PIF, we can get a list contains the network related object. For example for a tunnel PIF on a VLAN on bond master we will get `[Tunnel(ref_tunnel, ref_PIF); VLAN(ref_VLAN, ref_PIF), BOND_MASTER(ref_Bond, ref_PIF)]`. Then we could execute the check with the list when creating VLAN/Tunnel/Bond/SRIOV etc.

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 27 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H8YN31vUOLTltHRG0:-L4JzrS3U2K03TEy3zwe:bnhrruu) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L27)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

I'm missing the physical PIF here. Or is that what `Other_PIF` is? Or does `Other_PIF` include "I don't know; something must have gone wrong"?
</blockquote></details>

Considering a new solution, see above.

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 29 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H1JuuLpdBBOUD7CTn:-L4JT1oYcO2dFWhTtnFW:b2nhswy) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L29)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

This function does lots of DB-gets. On a pool slave, each on costs one round-trip to the master. This is why a PIF record is passed around in modules like `Nm`: you need to get the record from the master just once. It would be better to do the same here.
</blockquote></details>

Yes

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 32 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H4AWKDXOJwtSwgaep:-L4JT5GWC_yn0APj1ZId:b2nh59u) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L32)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

Interesting monadic experiment :)
</blockquote></details>

;-)

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 69 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H4f6cOPoCQrcsivBS:-L4K-0s5AK29qv7muatr:bnhrruu) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L69)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

This should be treated as an error, isn't it?
</blockquote></details>

Considering a new solution, see above.

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 72 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H5cs0lcm7J4hpFXWp:-L4JTULUxy2nBOLRibFc:b2nhswy) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L72)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

This is better put in an `assert_pif_is_managed` function. In fact, xapi_pif.ml has one of those, which we can move into this module, and call directly from xapi_vlan.ml and friends. It should not be part of the "allowed on" checks.
</blockquote></details>

Yes

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 87 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H57sFT5ArrLe1qfOt:-L4JTgy_DgWIIUgOeb62:b-hd0ty) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L87)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

I would leave the checks below in `Xapi_vlan`, because they aren't part of the "allowed on" check. It's better to keep this module focussed on the latter.
</blockquote></details>

You are right

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 100 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H6r1AIetk_F09ZMbx:-L4JTkC9GlaZvuWRNyFa:b2nhswy) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L100)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

See above
</blockquote></details>

Yes

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 110 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4HEHc_tXOkjWiCKeT2:-L4K-2df9sEZ_CJ1GfbT:bnhrruu) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L110)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

Although it is a huge improvement compared to what we used to have, I think these functions are still a little error prone, because it is hard to spot mistakes and confirm that the definitions are complete (it is not "obviously correct"). I was hoping for a more direct way of expressing allowed relations, and use the same definition for all checks. E.g. things like "a bond can only sit on top of physical PIFs" and "a VLAN can only sit on top of physical, bond or SRIOV". Or more interesting, "a tunnel can sit on a VLAN unless the VLAN sits on top of an SRIOV PIF".
</blockquote></details>

Considering a new solution, see above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3406)*

krizex:
Review status: 0 of 16 files reviewed at latest revision, 27 unresolved discussions, some commit checks failed.

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 26 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H8izOaHNsGFAgnnO6:-L4K-Ipze-h5quF2mZ6f:blpvkka) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L26)):*
<details><summary><i>Previously, krizex (Yang Qian) wrote…</i></summary><blockquote>

Yes, the type definition is incomplete and I found it is not able to be complete because we should keep each 2 of them is exclusive.
So after write the relation tree about the network related object (VLAN/Tunnel/Bond/SRIOV), I think we can implement a FSM to get the attribute of the PIF by checking the network related object underneath.
The relation tree is: (lowercase name is non-terminal symbol and uppercase name is terminal symbol)
```
   tunnel : vlan
          | PHYSICAL
          | BOND_MASTER

   vlan : vlan
        | sriov
        | BOND_MASTER
        | PHYSICAL

   sriov : BOND_SLAVE
         | PHYSICAL
```
So for any given PIF, we can get a list contains the network related object. For example for a tunnel PIF on a VLAN on bond master we will get `[Tunnel(ref_tunnel, ref_PIF); VLAN(ref_VLAN, ref_PIF), BOND_MASTER(ref_Bond, ref_PIF)]`. Then we could execute the check with the list when creating VLAN/Tunnel/Bond/SRIOV etc.

</blockquote></details>

The FSM:


---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3406)*

krizex:
Review status: 0 of 16 files reviewed at latest revision, 27 unresolved discussions, some commit checks failed.

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 26 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H8izOaHNsGFAgnnO6:-L4KFrI2lew6DASu2hsf:b-hk8q4p) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L26)):*
<details><summary><i>Previously, krizex (Yang Qian) wrote…</i></summary><blockquote>

The FSM:


</blockquote></details>

The relation tree has to be updated to:
```
   tunnel : vlan
          | PHYSICAL
          | BOND_MASTER

   vlan : vlan
        | sriov
        | BOND_MASTER
        | PHYSICAL

   sriov : bond_slave
         | PHYSICAL

   bond_slave : PHYSICAL
```
And the FSM update accordingly


---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3406)*

krizex:
Review status: 0 of 16 files reviewed at latest revision, 27 unresolved discussions, some commit checks failed.

---

*[ocaml/xapi/nm.ml, line 246 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4GzgP5RfELbxqeM8QL:-L4KiOZ9bqS65X3wFpmj:b-qqy4wx) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/nm.ml#L246)):*
<details><summary><i>Previously, krizex (Yang Qian) wrote…</i></summary><blockquote>

Yes, but I think it could be better to introduce a new type `VLAN_master_on_bond` to make the type definition clearer.
</blockquote></details>

Ignore me ^ because we have a new solution now.

---

*[ocaml/xapi/nm.ml, line 247 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H3u6xuZBcz8kA39rD:-L4KiXSHLUaZgyj4aFAK:b-qqy4wx) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/nm.ml#L247)):*
<details><summary><i>Previously, krizex (Yang Qian) wrote…</i></summary><blockquote>

No `VLAN_master_on_physical` does not has VLAN as its argument, but I think it's better to have. So I will refine the definition of `type pif_type_t` to make all types not only contain the pif_ref but also the bond/tunnel/network_sriov/vlan object.
</blockquote></details>

Ignore me ^ because we have a new solution now.

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 25 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H0ua5cQDzIZyvA39h:-L4KigOs_2CV0CSmidEl:b-qqy4wx) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L25)):*
<details><summary><i>Previously, krizex (Yang Qian) wrote…</i></summary><blockquote>

Yes, but I think it could be better to introduce a new type `VLAN_master_on_bond` to make the type definition clearer.
</blockquote></details>

Ignore me ^ because we have a new solution now.

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 26 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H8izOaHNsGFAgnnO6:-L4Kni1fgWGwnDLKJ2Np:b-5nt6dy) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L26)):*
<details><summary><i>Previously, krizex (Yang Qian) wrote…</i></summary><blockquote>

The relation tree has to be updated to:
```
   tunnel : vlan
          | PHYSICAL
          | BOND_MASTER

   vlan : vlan
        | sriov
        | BOND_MASTER
        | PHYSICAL

   sriov : bond_slave
         | PHYSICAL

   bond_slave : PHYSICAL
```
And the FSM update accordingly


</blockquote></details>

Ignore me^.
After coding the solution, I find the relation tree can be updated to (treat bond slave as terminate symbol):
```
   tunnel : vlan
          | PHYSICAL
          | BOND_MASTER

   vlan : vlan
        | sriov
        | BOND_MASTER
        | PHYSICAL

   sriov : BOND_SLAVE
         | PHYSICAL
```
and the FSM should be (cut the line from `bond_slave` to `physical`)


---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3406)*

robhoes:
The way reviewable.io works is to do what we normally ask you not to do: force-push to the branch of your PR. Your latest commit should therefore be squashed into the first one. Reviewable will make sure that comments are not lost, of that the new diff is visible as a new revision.

---

Review status: 0 of 17 files reviewed at latest revision, 24 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-:-L4NIFGD9oIV0O0zoJtf:b-7wfj3l)*

krizex:
Got it, I will squash the fix into the existing commit.

---

Review status: 0 of 17 files reviewed at latest revision, 24 unresolved discussions, some commit checks failed.

---

*[ocaml/xapi/xapi_network_sriov.ml, line 38 at r6](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4HMeI0vEShbeDRT5gh:-L4ZZthiRqKPd1OjGUVB:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/96a92e56f59c2b5a928c7e50e4263dcc8b6800c3/ocaml/xapi/xapi_network_sriov.ml#L38)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

Just `~physical_PIF` is enough.
</blockquote></details>

Done.

---

*[ocaml/xapi/xapi_pif_helpers.ml, line 26 at r5](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-L4H8izOaHNsGFAgnnO6:-L4Yzsw5KAjfMRAUkWOl:bjjvdfd) ([raw file](https://github.com/xapi-project/xen-api/blob/96c64a9c410a545717be2961b33110900736b628/ocaml/xapi/xapi_pif_helpers.ml#L26)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

This is still a little odd, as the states are a mix of two different concepts: the kinds of PIF themselves, and their linker objects such as vlan, bond, tunnel and sriov. And a bond slave _is_ a physical PIF as well.

I also don't understand the self-edge at `vlan`: you can't put a VLAN on a VLAN. Also, this doesn't exclude tunnel-on-vlan-on-sriov. I'm not sure if you can encode that in a state diagram, because it needs to "remember" the previous state when you are in state vlan.
</blockquote></details>

For the self-edge at `vlan`, that's because of we have `vlan` on `vlan` scenario, see https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_vlan.ml#L58.

I introduced the FSM aims to get type of any given PIF, that means it does not contain the restriction when creating PIF. However you pointed out the issue of `tunnel-on-vlan-on-sriov`, indeed that is a problem and not restricted by the FSM.

On second thoughts, we can introduce a function to classify any given PIF into ONLY one of following types:
```
type pif_type_t =
  | Tunnel_access of ref_tunnel
  | VLAN_untagged of ref_VLAN
  | Network_sriov_logical of ref_network_sriov
  | Bond_master of ref_Bond
  | Physical of pIF_t
```
For `Tunnel_access`, `VLAN_untagged` and `Network_sriov_logical` PIF, we can recursive call the function to get the type of the underneath PIF.
So finally for any given PIF, we can recursive call the function to get a list of types of the PIFs that underneath the given PIF until reach the root PIF (the root should be `Bond_master` or `Physical`).
As we does not consider the restriction when getting PIF type, so we can deprecate the FSM now.

When creating network linker objects, we can apply the function to the given PIF and get a list of PIF types underneath, and execute the restriction checks with the list.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3406#-:-L4ZZPKa2osLeQPT_QPw:bb8neqd)*

krizex:
Close this one and split the changes into smaller PRs.

PR #3407, opened 2018-01-18
Added explicit dependency to message-switch-unix.

kc284:
This should fix the travis build.


kc284:
This should fix the travis build.

PR #3411, opened 2018-01-22
CA-270622: HA max tolerance is wrong for >16-host pools

fillzero:
Introduce a new algorithm to calculate max failure tolerance for large pools.

1. Assume all VMs are as big as the biggest
2. Calculate the number of biggest VM each host provides
3. Assume the biggest r (num_failures) hosts die, get left capacity of
left hosts
4. Compare the failed VM numbers with left capacity

More details and dev test result:
https://info.citrite.net/display/xenserver/HA+max+failure+tolerance+investigation

Signed-off-by: Liang Dai <liang.dai1@citrix.com>

### DIFF
--- a/ocaml/xapi/binpack.ml
+++ b/ocaml/xapi/binpack.ml
@@ -173,22 +173,24 @@ let simulate_failure config dead_host =
   { config with hosts = hosts; placement = placement; num_failures = config.num_failures - 1 }
 
 let simulate_failure_approximation config =
-  let rec sum_of_first_n n l = match l with
-    | [] -> failwith "sum_of_first_n"
-    | x::l -> if n=1 then x else x + (sum_of_first_n (n-1) l)
-  in
-  (* Assume all VMs are as big as the biggest *)
-  let vm_size = List.fold_left (fun acc (_, size) -> max acc size) 0L config.vms in
+  let sum = List.fold_left (+) 0 in
+  let rec drop n = function
+    | [] -> []
+    | x::xl when n>0 -> drop (n-1) xl
+    | l -> l in
 
-  (* How many VMs can each host support to power on, ascending order *)
-  let host_capacity = List.sort compare (List.map (fun (_, h_size) -> Int64.(to_int (div h_size vm_size))) config.hosts) in
+  (* Assume all VMs are as big as the biggest *)
+  let max_vm_size = List.fold_left (fun acc (_, size) -> max acc size) 0L config.vms in
+
+  (* How many VMs can each host support to power on, descending order *)
+  let hosts_by_capacity = List.sort less_than' (List.map (fun (_, h_size) -> Int64.(to_int (div h_size max_vm_size))) config.hosts) in
   (* Assume biggest hosts fail, get left capacity *)
-  let left_capacity = sum_of_first_n (List.length host_capacity - config.num_failures) host_capacity in
+  let left_capacity = drop config.num_failures hosts_by_capacity |> sum in
 
   (* VMs placed in each host, descending order *)
   let num_vms_per_host = List.sort less_than' (List.map (fun (host, _) -> List.length (List.filter (fun (vm, h) -> h = host) config.placement)) config.hosts) in
-  let failure_vm_length = sum_of_first_n config.num_failures num_vms_per_host in
-  if failure_vm_length > left_capacity then raise Stop
+  let vms_to_move = Xapi_stdext_std__Listext.List.take config.num_failures num_vms_per_host |> sum in
+  if vms_to_move > left_capacity then raise Stop
 
 (** For the nCr binpack strategy return true if a plan is always possible *)
 let plan_always_possible config =
--- a/ocaml/xapi/binpack.ml
+++ b/ocaml/xapi/binpack.ml
@@ -14,6 +14,7 @@
 
 (* Used to sort pairs into descending order of their second component *)
 let less_than (_, a) (_, b) = compare b a
+let less_than' a b = compare b a
 
 let rec insert compare elt sorted_list = match sorted_list with
   | [] -> [ elt ]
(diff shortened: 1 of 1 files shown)


lindig:
Please see my discussion of this approach at the internal wiki. I also have some comments on the code but would like to address the fundamental questions first.

lindig:
I believe the algorithm is sound. I'll leave some comment to improve the code.

ocaml/xapi/binpack.ml
```diff
@@ -171,6 +172,24 @@ let simulate_failure config dead_host =
   let placement = List.map (fun (vm, host) -> vm, (if List.mem_assoc vm plan then assoc "simulate_failure" vm plan else host)) config.placement in
   { config with hosts = hosts; placement = placement; num_failures = config.num_failures - 1 }
 
+let simulate_failure_approximation config =
+  let rec sum_of_first_n n l = match l with
+    | [] -> failwith "sum_of_first_n"
+    | x::l -> if n=1 then x else x + (sum_of_first_n (n-1) l)
+  in
+  (* Assume all VMs are as big as the biggest *)
+  let vm_size = List.fold_left (fun acc (_, size) -> max acc size) 0L config.vms in
```
lindig: A better name would be `max_vm_size` or `max_vm`.

ocaml/xapi/binpack.ml
```diff
@@ -171,6 +172,24 @@ let simulate_failure config dead_host =
...
+
+  (* How many VMs can each host support to power on, ascending order *)
+  let host_capacity = List.sort compare (List.map (fun (_, h_size) -> Int64.(to_int (div h_size vm_size))) config.hosts) in
```
lindig: A better name would be `hosts_by_capacity` and a comment that this is in ascending order.

ocaml/xapi/binpack.ml
```diff
@@ -171,6 +172,24 @@ let simulate_failure config dead_host =
   let placement = List.map (fun (vm, host) -> vm, (if List.mem_assoc vm plan then assoc "simulate_failure" vm plan else host)) config.placement in
   { config with hosts = hosts; placement = placement; num_failures = config.num_failures - 1 }
 
+let simulate_failure_approximation config =
+  let rec sum_of_first_n n l = match l with
```
lindig: I am not convinced by this function. I believe it is better to combine these functions:

* `sum: int list -> int`
* `take: int -> 'a list -> 'a list` (`Xapi_stdext_std__Listext.List.take`)
* `drop: int -> 'a list -> 'a list` (drops first n elements)

ocaml/xapi/binpack.ml
```diff
@@ -171,6 +172,24 @@ let simulate_failure config dead_host =
...
+  let vm_size = List.fold_left (fun acc (_, size) -> max acc size) 0L config.vms in
+
+  (* How many VMs can each host support to power on, ascending order *)
+  let host_capacity = List.sort compare (List.map (fun (_, h_size) -> Int64.(to_int (div h_size vm_size))) config.hosts) in
+  (* Assume biggest hosts fail, get left capacity *)
+  let left_capacity = sum_of_first_n (List.length host_capacity - config.num_failures) host_capacity in
+
+  (* VMs placed in each host, descending order *)
+  let num_vms_per_host = List.sort less_than' (List.map (fun (host, _) -> List.length (List.filter (fun (vm, h) -> h = host) config.placement)) config.hosts) in
+  let failure_vm_length = sum_of_first_n config.num_failures num_vms_per_host in
```
lindig: A better name would be `displaced_vms`, `vms_to_move` or similar. 

```
let vms_to_move = take config.num_failures num_vms_per_host |> sum
```

fillzero:
Thanks, updated.

ocaml/xapi/binpack.ml
```diff
@@ -171,6 +172,26 @@ let simulate_failure config dead_host =
...
+    | x::xl when n>0 -> drop (n-1) xl
+    | l -> l in
+
+  (* Assume all VMs are as big as the biggest *)
+  let max_vm_size = List.fold_left (fun acc (_, size) -> max acc size) 0L config.vms in
+
+  (* How many VMs can each host support to power on, ascending order *)
+  let hosts_by_capacity = List.sort compare (List.map (fun (_, h_size) -> Int64.(to_int (div h_size max_vm_size))) config.hosts) in
+  (* Assume biggest hosts fail, get left capacity *)
+  let left_capacity = drop config.num_failures hosts_by_capacity |> sum in
```
lindig: Is this correct? When `hosts_by_capacity` is in ascending order, this drops the `num_failure` hosts with the smallest capacity. I was expecting:

```
let left_capacity = List.rev hosts_by_capacity |> drop config.num_failures |> sum in
```

This could be also expressed using `take` but because one would need `List.length hosts_by_capacity` this wouldn't be much more efficient.

fillzero: Oh, Thanks.
Now change it to descending order:
```
  (* How many VMs can each host support to power on, descending order *)
  let hosts_by_capacity = List.sort less_than' (List.map (fun (_, h_size) -> Int64.(to_int (div h_size max_vm_size))) config.hosts) in
  (* Assume biggest hosts fail, get left capacity *)
  let left_capacity = drop config.num_failures hosts_by_capacity |> sum in
```

lindig:
I think the code is fine now. I am a bit reluctant to merge it because we probably would like to see more testing to convince ourselves that it is indeed correct. I'll discuss it in RIng3.

robhoes:
Yes, this should go onto a feature branch first, and not straight into master.

fillzero:
Thanks, raise https://github.com/xapi-project/xen-api/pull/3416 for feature branch.

PR #3414, opened 2018-01-24
Fixed an issue with device number in script.

davisaaronj:
Line 41 had a typo removed and extraneous p that broke this script from working.

Signed-off-by: Aaron Davis <aaron.davis@rackspace.com>

### DIFF
--- a/scripts/xe-edit-bootloader
+++ b/scripts/xe-edit-bootloader
@@ -37,7 +37,7 @@ while getopts "hu:n:p:f:" opt ; do
     h) usage ;;
     u) vm_uuid=${OPTARG} ;;
     n) vm_name="${OPTARG}" ;;
-    p) device_number=p${OPTARG} ;;
+    p) device_number=${OPTARG} ;;
     f) default_file_list="${OPTARG}" ;;
     *) echo "Invalid option"; usage ;;
     esac


mseri:
Thanks for the PR. Please can you add `Signed-off-by: Your Name <your@email>` at the bottom of the commit message?

davisaaronj:
Did I update my request as you wanted or did I need to add the signature somewhere else?

mseri:
Sorry, it needs to be in the text of the commit message itself. It is still missing there. You should `git commit --amend` and force push the updated commit message on the PR branch

robhoes:
Please look here for an explanation about what it means to sign off a patch: https://wiki.xenproject.org/wiki/Submitting_Xen_Project_Patches#Signing_off_a_patch.

davisaaronj:
Sorry for the delay I got the commit updated.

robhoes:
Did you push it...?

davisaaronj:
Sorry I thought it had pushed. I have pushed the update so the comment should now be correct.

mseri:
Thanks!

BobBall:
Unfortunately I suspect this pull request was incorrect.
I'm not convinced this is a typo in the script - this is actually a problem with the way kpartx works by default on different disk types.
On many systems kpartx will default to including a p before the partition number, but in some systems it will not (TBH I don't know what the difference is).

The correct fix is actually to force kpartx to use a defined partition format with "kpartx -pp" (and the equivalent for the delete)

BobBall:
@davisaaronj Please could you review + test https://github.com/xapi-project/xen-api/pull/3448

PR #3415, opened 2018-01-24
XOP-908/CA-279498: Zero-extend CPU features when comparing old and new sets

robhoes:
Backport of #3400.


robhoes:
Travis failed... we probably need to update the buildenv for this branch.

PR #3416, opened 2018-01-26
CA-270622: HA max tolerance is wrong for >16-host pools

fillzero:


fillzero:
The PR has been reviewed in https://github.com/xapi-project/xen-api/pull/3411.

PR #3418, opened 2018-01-28
CP-26717: Update Xapi to use PPX-based Gpumon IDL

minishrink:
This PR is the second of three to port GPU monitoring daemon gpumon to use PPX-based code generation for RPCs.

1. [Port Gpumon interface](https://github.com/xapi-project/xcp-idl/pull/196)
2. this
3. [Port Gpumon](https://github.com/xapi-project/xen-api/pull/3418)

### DIFF
--- a/ocaml/xapi/xapi_gpumon.ml
+++ b/ocaml/xapi/xapi_gpumon.ml
@@ -45,7 +45,7 @@ module Nvidia = struct
   let get_pgpu_compatibility_metadata ~dbg ~pgpu_pci_address =
     let metadata =
       pgpu_pci_address
-      |> Gpumon_client.Client.Nvidia.get_pgpu_metadata dbg 
+      |> Gpumon_client.Client.Nvidia.get_pgpu_metadata dbg
       |> Stdext.Base64.encode
     in [key, metadata]
 
@@ -70,7 +70,7 @@ module Nvidia = struct
    * one vGPU per VM. The underdyling problem is that there is no
    * mapping between a vGPU and vgpu_instance in the NVIDIA library
    * currently.
-   *)
+  *)
   let get_vgpu_compatibility_metadata ~__context ~vgpu =
     let this = "get_vgpu_compatibility_metadata" in
     try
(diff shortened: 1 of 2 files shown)


ocaml/xapi/xapi_gpumon.ml
```diff
@@ -79,21 +79,21 @@ module Nvidia = struct
...
       |> fun self -> Db.PGPU.get_PCI ~__context ~self
-      |> fun self -> Db.PCI.get_pci_id ~__context ~self
-      |> Gpumon_client.Client.Nvidia.get_vgpu_metadata dbg domid
-      |> function
-        | []      -> []
-        | [meta]  -> [key, Stdext.Base64.encode meta]
-        | _::_    -> failwith @@ Printf.sprintf
-          "%s: VM %s (dom %d) has more than one NVIDIA vGPU (%s)"
-          this (Ref.string_of vm) domid __LOC__
+                     |> fun self -> Db.PCI.get_pci_id ~__context ~self
```
mseri: Indentation here has gone wild as well... it would be nice to see if there is a flag to stop this behaviour in `ocp-indent`

ocaml/xapi/xapi_gpumon.ml
```diff
@@ -152,7 +151,7 @@ module Nvidia = struct
       let vgpu_metadata () =
         Db.VGPU.get_compatibility_metadata ~__context ~self:vgpu
         |> fun md -> [List.assoc key md]
-        |> List.map Stdext.Base64.decode in
+                     |> List.map Stdext.Base64.decode in
```
mseri: Weird indent

ocaml/xapi/xapi_gpumon.ml
minishrink: I think it assumes it's part of the same function, which would make sense. Not sure how you'd differentiate between that without parentheses/begin-end

mseri: By the use of the `|>`

minishrink: But what if the code above was like this:
`|> (function
  | x -> do_this
  | y -> some_function
        |> followed_by_this_one)`
I'm not sure how it would parse that without the parentheses

mseri: I would expect to use parentheses in that case in fact

minishrink: I think that's currently the situation; the original function wasn't wrapped in parentheses, so ocp-indent assumed the next `|> function` was part of the same lambda function, hence the weird indent.

mseri: Yes, I'd like to understand if it is right or not :)

PR #3421, opened 2018-01-30
Convert Test_vm_check_operation_error to alcotest

gaborigloi:


lindig:
Travis passes but does it execute the tests? I wasn't sure when glancing at the logs.

gaborigloi:
It's shown by the opam-coverage build at least: https://travis-ci.org/xapi-project/xen-api/jobs/335218466#L3782
In the main xenserver-build-env build job, we hide the logs: https://github.com/xapi-project/xen-api/blob/master/.travis-xenserver-build-env.sh#L12
When everything is converted into alcotest, we can undo that change, as it only shows the output of failed tests.

PR #3422, opened 2018-01-31
CA-271491: Creating NFS SR on 64-host pool ~7x slower than 16-host

taoyongd:
Adding a SR to a large pool(64 hosts) seems very slow which is caused by
serial PBD plug across the pool. This fix is to use async PBD plugwhen creating SR.

By doing this change, I have run Ring3 and Stroage BST which are both
passed(refer to 78821, 78820).

### DIFF
--- a/ocaml/xapi/xapi_sr.ml
+++ b/ocaml/xapi/xapi_sr.ml
@@ -240,12 +240,15 @@ let create  ~__context ~host ~device_config ~(physical_size:int64) ~name_label ~
   end;
   Helpers.call_api_functions ~__context
     (fun rpc session_id ->
-       List.iter
-         (fun self ->
-            try
-              Client.PBD.plug ~rpc ~session_id ~self
-            with e -> warn "Could not plug PBD '%s': %s" (Db.PBD.get_uuid ~__context ~self) (Printexc.to_string e))
-         pbds);
+       let tasks = List.fold_left (fun tasks self ->
+           try
+             let task = Client.Async.PBD.plug ~rpc ~session_id ~self in
+             task :: tasks
+           with e ->
+             warn "Could not plug PBD '%s': %s" (Db.PBD.get_uuid ~__context ~self) (Printexc.to_string e);
+             tasks)
+           [] pbds in
+       Tasks.wait_for_all ~rpc ~session_id ~tasks);
   sr_ref
 
 let check_no_pbds_attached ~__context ~sr =
--- a/ocaml/xapi/xapi_sr.ml
+++ b/ocaml/xapi/xapi_sr.ml
@@ -240,14 +240,7 @@ let create  ~__context ~host ~device_config ~(physical_size:int64) ~name_label ~
   end;
   Helpers.call_api_functions ~__context
     (fun rpc session_id ->
-       let tasks = List.fold_left (fun tasks self ->
-           try
-             let task = Client.Async.PBD.plug ~rpc ~session_id ~self in
-             task :: tasks
-           with e ->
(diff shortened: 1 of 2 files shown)


taoyongd:
Hi @jonludlam , could you please help to review this? Thanks a lot.

ocaml/xapi/xapi_sr.ml
```diff
@@ -240,12 +240,15 @@ let create  ~__context ~host ~device_config ~(physical_size:int64) ~name_label ~
...
-            try
-              Client.PBD.plug ~rpc ~session_id ~self
-            with e -> warn "Could not plug PBD '%s': %s" (Db.PBD.get_uuid ~__context ~self) (Printexc.to_string e))
-         pbds);
+       let tasks = List.fold_left (fun tasks self ->
+           try
+             let task = Client.Async.PBD.plug ~rpc ~session_id ~self in
+             task :: tasks
+           with e ->
+             warn "Could not plug PBD '%s': %s" (Db.PBD.get_uuid ~__context ~self) (Printexc.to_string e);
```
jonludlam: This almost certainly does very little - all you're checking for is that the async call was successfully started, not that it completed successfully. I think you're probably better off just replacing the whole fold_left thing with just

```ocaml
let tasks = List.map (fun self -> Client.Async.PBD.plug ~rpc ~session_id ~self) pbds in
Tasks.wait_for_all ~rpc ~session_id ~tasks
```

taoyongd: Hmm, will map check that the async call was completed successfully?

mseri: No, but `wait_for_all` will wait for the completion

taoyongd: yes.  So what the difference by using map instead of left_fold?

mseri: It's easier to read. The comment was more about the fact that the `try/with` in the fold is not needed because it does nothing, and in the end you are using a fold to just to a map, so you might as well use a map and make the code more readable

mseri: If anything it would become something like:
```ocaml
try
   let tasks = List.map (fun self -> Client.Async.PBD.plug ~rpc ~session_id ~self) pbds in
   Tasks.wait_for_all ~rpc ~session_id ~tasks
with err ->
   warn "Could not plug some PBDs: %s" (Printexc.to_string err);
```
and I am not even sure that the `err` that you get there is informative at all.

taoyongd: Ok. Understand your point. Thanks @mseri @jonludlam . 
If so, I would just use map as @jonludlam  suggests. Will update it later.   Thanks you.

ocaml/xapi/xapi_sr.ml
```diff
@@ -240,12 +240,8 @@ let create  ~__context ~host ~device_config ~(physical_size:int64) ~name_label ~
...
   Helpers.call_api_functions ~__context
     (fun rpc session_id ->
-       List.iter
-         (fun self ->
-            try
-              Client.PBD.plug ~rpc ~session_id ~self
-            with e -> warn "Could not plug PBD '%s': %s" (Db.PBD.get_uuid ~__context ~self) (Printexc.to_string e))
-         pbds);
+       let tasks = List.map (fun self -> Client.Async.PBD.plug ~rpc ~session_id ~self) pbds in
+       Tasks.wait_for_all ~rpc ~session_id ~tasks);
```
mseri: Note that `wait_for_all` does not fail if tasks have errored, so this should have retained the old partial success behaviour

PR #3423, opened 2018-01-31
[CP-25971] Introduce Cluster.pool_force_destroy

mcintyre94:
Cluster.pool_force_destroy is a best-effort function that performs as much Cluster cleanup as it can given the current state of the cluster. It is assumed that it typically will not complete successfully, because in the case that it can it would be better to call `Cluster.pool_destroy` instead. Instead it completes as much as it can, as cleanly as it can, and aims to give a useful error message if it cannot complete.

It first tries to cleanly disable (ie leave the cluster) each slave. We expect exceptions here, and ignore them. Next we try to force disable each host, including the master. We include the master here because if it's the only healthy node then we can at least remove it (it can't cleanly leave because it's the last node). This time we collect exceptions instead of ignoring them - but we attempt to force destroy all nodes before raising any error so we clean up everything that we possibly can. If there are no errors here then we destroy the cluster, but if a force destroy failed then we report that failure to the user and stop.

This commit also modifies `Cluster.destroy` to support clusters with 0 or 1 nodes, needed in the case where we managed to clean up all nodes, and makes `Cluster.pool_destroy` support the case where the pool master is Cluster_host is down in the same way that `Cluster.pool_force_destroy` does.

Tested scenarios:
- In a cluster of 3 nodes where 2 are down (so there is no quorum, and the up node can't leave), we clean up the active node successfully.
- In a cluster of 3 with all nodes available, we clean up all nodes and destroy the cluster successfully.

Signed-off-by: Callum McIntyre callum.mcintyre@citrix.com


ocaml/xapi/xapi_cluster.ml
```diff
@@ -69,13 +69,21 @@ let create ~__context ~network ~cluster_stack ~pool_auto_join ~token_timeout ~to
 
 let destroy ~__context ~self =
   let dbg = Context.string_of_task __context in
-  assert_cluster_has_one_node ~__context ~self;
-  let cluster_host = Db.Cluster.get_cluster_hosts ~__context ~self |> List.hd in
-  assert_cluster_host_has_no_attached_sr_which_requires_cluster_stack ~__context ~self:cluster_host;
+  let cluster_hosts = Db.Cluster.get_cluster_hosts ~__context ~self in
+  let cluster_host = match List.length cluster_hosts with
+    | 0 -> None
+    | 1 -> Some (cluster_hosts |> List.hd)
```
edwintorok: You can match the list directly and then you don't need to do List.hd here, i.e.
```
match cluster_hosts with
| [] -> None
| [ cluster_host ] -> Some cluster_host
| _ -> (* now you do List.length to report error *)
```
```

File: ocaml/xapi/xapi_cluster.ml [hunk 2]
```diff
@@ -100,17 +108,79 @@ let pool_create ~__context ~network ~cluster_stack ~token_timeout ~token_timeout
...
+    (fun exns_so_far cluster_host ->
+      Helpers.call_api_functions ~__context (fun rpc session_id ->
+        try
+          Client.Client.Cluster_host.force_destroy ~rpc ~session_id ~self:cluster_host;
+          exns_so_far
+        with e ->
+          Backtrace.is_important e;
+          let uuid = Client.Client.Cluster_host.get_uuid ~rpc ~session_id ~self:cluster_host in
+          debug "Ignoring exception while trying to force destroy cluster host %s: %s" uuid (ExnHelper.string_of_exn e);
+          exns_so_far @ [e]
```
edwintorok: It is usually better to construct lists by adding value to head (`O(1)`) vs concatenating (`O(n)`). You will get the list in reverse order, but thats fine in this case.

ocaml/xapi/xapi_cluster.ml
```diff
@@ -100,17 +108,79 @@ let pool_create ~__context ~network ~cluster_stack ~token_timeout ~token_timeout
...
+    begin
+    match exns with
+    | [] -> ()
+    | e :: _ -> raise Api_errors.(Server_error (cluster_force_destroy_failed, [ExnHelper.string_of_exn e]))
+    end;
+
+    try
+      Helpers.call_api_functions ~__context (fun rpc session_id ->
+          Client.Client.Cluster.destroy ~rpc ~session_id ~self)
+    with e ->
```
edwintorok: This try/with appears to be redundant, you can just drop it and let the exception escape on its own.

mcintyre94: Oops, thanks - I added that debugging to add the backtrace but I don't think we need to keep it. I'll remove it.

mcintyre94: Oops thanks, I added that debugging to get the backtrace but I don't think we need to keep it - I'll remove it.

ocaml/xapi/xapi_cluster.ml
```diff
@@ -70,10 +70,13 @@ let create ~__context ~network ~cluster_stack ~pool_auto_join ~token_timeout ~to
...
   let dbg = Context.string_of_task __context in
   let cluster_hosts = Db.Cluster.get_cluster_hosts ~__context ~self in
-  let cluster_host = match List.length cluster_hosts with
-    | 0 -> None
-    | 1 -> Some (cluster_hosts |> List.hd)
-    | n -> raise Api_errors.(Server_error(cluster_does_not_have_one_node, [string_of_int n]))
+  let cluster_host = match cluster_hosts with
+    | [] -> None
+    | [ cluster_host ] -> Some (cluster_host)
+    | _ -> (
```
edwintorok: I think the parenthesis shouldn't be needed here, due to the outer `let in`

ocaml/xapi/xapi_cluster.ml
```diff
@@ -152,7 +155,7 @@ let pool_force_destroy ~__context ~self =
           Backtrace.is_important e;
           let uuid = Client.Client.Cluster_host.get_uuid ~rpc ~session_id ~self:cluster_host in
           debug "Ignoring exception while trying to force destroy cluster host %s: %s" uuid (ExnHelper.string_of_exn e);
-          exns_so_far @ [e]
+          [e] @ exns_so_far
```
edwintorok: Sorry I meant to do this: `e :: exns_so_far`, that avoids concatenation

edwintorok:
I think this looks good now, can you squash the fixups?

ocaml/xapi/xapi_cluster.ml
```diff
@@ -100,17 +110,75 @@ let pool_create ~__context ~network ~cluster_stack ~token_timeout ~token_timeout
...
+        Client.Client.Cluster.destroy ~rpc ~session_id ~self)
+
 (* Helper function; concurrency checks are done in implementation of Cluster.destroy and Cluster_host.destroy *)
 let pool_destroy ~__context ~self =
   (* For now we assume we have only one pool, and that the cluster is the same as the pool.
      This means that the pool master must be a member of this cluster. *)
   let master = Helpers.get_master ~__context in
   let master_cluster_host =
     Xapi_clustering.find_cluster_host ~__context ~host:master
-    |> Xapi_stdext_monadic.Opt.unbox
```
gaborigloi: A while ago I tried another alternative change that makes `pool_destroy` more distributed: https://github.com/gaborigloi/xen-api/commit/b0bc29d6662f02e3d317810ab1bcaeab489f462d
This should work when the master has no `Cluster_host`. However, it didn't work because `Cluster.destroy` is always redirected to master. So we'd have to change message forwarding too.
What happens with this change if we destroy the cluster host of the master and the do a pool destroy on the cluster?

gaborigloi: We've discussed that this would fail if there is no master `Cluster_host` because in that case we'd try `Cluster_host.destroy` on the last one, and that would fail.

mcintyre94: As we discussed I'll remove my changes to `pool_destroy` and we'll use your patch later.

mcintyre94:
I've pushed a fixup which leaves `pool_destroy` unchanged

edwintorok:
If @gaborigloi agrees we can merge this now, using the squash+merge github button.

ocaml/xapi/xapi_cluster.ml
```diff
@@ -69,13 +69,23 @@ let create ~__context ~network ~cluster_stack ~pool_auto_join ~token_timeout ~to
 
 let destroy ~__context ~self =
   let dbg = Context.string_of_task __context in
-  assert_cluster_has_one_node ~__context ~self;
```
gaborigloi: I think this was the only use of this function, so I think we can remove it for now.

gaborigloi:
```
~/s/xen-api (feature/REQ477/master|✔) # rg assert_cluster_has_one_node
ocaml/xapi/xapi_clustering.ml
161:let assert_cluster_has_one_node ~__context ~self =

ocaml/xapi/xapi_cluster.ml
72:  assert_cluster_has_one_node ~__context ~self;
~/s/xen-api (feature/REQ477/master|✔) #
```

ocaml/xapi/xapi_cluster.ml
```diff
@@ -100,6 +110,65 @@ let pool_create ~__context ~network ~cluster_stack ~token_timeout ~token_timeout
...
+          e :: exns_so_far
+      )
+    )
+    [] all_remaining_cluster_hosts
+    in
+
+    begin
+    match exns with
+    | [] -> ()
+    | e :: _ -> raise Api_errors.(Server_error (cluster_force_destroy_failed, [ExnHelper.string_of_exn e]))
```
gaborigloi: According to the description of this error, we should pass in the cluster ref here instead of the exception.

ocaml/xapi/xapi_cluster.mli
```diff
@@ -38,6 +38,12 @@ val pool_create : __context:Context.t -> network:API.ref_network ->
     that creates the Cluster object and then creates Cluster_host objects for
     all hosts in the pool. *)
 
+val pool_force_destroy : __context:Context.t -> self:API.ref_Cluster -> unit
+(** [pool_force_destroy ~__context ~pool ~self] is the implementation of the XenAPI
```
gaborigloi: `pool_force_destroy ~__context ~self` without `~pool`

ocaml/xapi/xapi_cluster.mli
```diff
@@ -38,6 +38,12 @@ val pool_create : __context:Context.t -> network:API.ref_network ->
...
+    method 'Cluster.pool_force_create'. This is a convenience function that force
```
gaborigloi: `Cluster.pool_force_destroy`

ocaml/xapi/xapi_cluster.mli
```diff
@@ -38,6 +38,12 @@ val pool_create : __context:Context.t -> network:API.ref_network ->
...
     all hosts in the pool. *)
 
+val pool_force_destroy : __context:Context.t -> self:API.ref_Cluster -> unit
+(** [pool_force_destroy ~__context ~pool ~self] is the implementation of the XenAPI
+    method 'Cluster.pool_force_create'. This is a convenience function that force 
+    destroys the Cluster_host objects for all hosts in the pool and then destroys the
+    Cluster object. *)
+
 val pool_destroy : __context:Context.t -> self:API.ref_Cluster -> unit
 (** [pool_destroy ~__context ~pool ~self] is the implementation of the XenAPI
```
gaborigloi: `pool_destroy ~__context ~self` without ~pool

ocaml/xapi/xapi_cluster_host.mli
```diff
@@ -45,6 +45,11 @@ val create : __context:Context.t -> cluster:API.ref_Cluster -> host:API.ref_host
 (** [create ~__context ~cluster ~host] is implementation of the XenAPI call
     'Cluster_host.create'. It is the Cluster_host object constructor *)
 
+val force_destroy : __context:Context.t -> self:API.ref_Cluster_host -> unit
+(** [force_destroy ~__context ~self] is the implementation of the XenAPI call
+    'Cluster_host.force_destroy'. It forcefully removes the DB object and
```
gaborigloi: It seems that it does not always remove the DB object, only when it succeeded to destroy the cluster.

PR #3420, opened 2018-01-29
Port Test_vdi_allowed_operations to alcotest

gaborigloi:
Signed-off-by: Gabor Igloi <gabor.igloi@citrix.com>

PR #3425, opened 2018-02-01
CA-268763 - additional check for pool_size feature

jonludlam:
Testing during the pool-join prechecks is racy, so we now _also_ check in Xapi_host.create

### DIFF
--- a/ocaml/xapi/test_host.ml
+++ b/ocaml/xapi/test_host.ml
@@ -1,0 +1,83 @@
+(*
+ * Copyright (C) 2017 Citrix Systems Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+open Stdext
+open Test_common
+
+module D=Debug.Make(struct let name="test_xapi_xenops" end)
+open D
+
+let add_host __context name =
+  ignore(Xapi_host.create ~__context
+    ~uuid:(Uuid.to_string (Uuid.make_uuid()))
+    ~name_label:name ~name_description:"" ~hostname:"host2" ~address:"127.0.0.1"
+    ~external_auth_type:"" ~external_auth_service_name:"" ~external_auth_configuration:[]
+    ~license_params:[] ~edition:"" ~license_server:[] ~local_cache_sr:(Ref.null) ~chipset_info:[] ~ssl_legacy:false)
+
+(* Creates an unlicensed pool with the maximum number of hosts *)
+let setup_test () =
+  (* Create an unlicensed pool *)
+  let __context = make_test_database () in
+  let pool = Db.Pool.get_all ~__context |> List.hd in
+  Db.Pool.set_restrictions ~__context ~self:pool ~value:(Features.to_assoc_list []);
+
+  (* Add hosts until we're at the maximum unlicensed pool size *)
+  while (Db.Host.get_all ~__context |> List.length < Xapi_globs.restricted_pool_size) do
+    add_host __context "host";
+  done;
+  Alcotest.(check int) "initial number of hosts"
+    Xapi_globs.restricted_pool_size
+    (Db.Host.get_all ~__context |> List.length);
+  __context
+
+let test_host_join_restriction () =
+  let __context = setup_test () in
+
+  (* Check adding one more is a failure *)
+  Alcotest.check_raises "Should fail"
+    (Api_errors.Server_error(Api_errors.license_restriction, [Features.name_of_feature Features.Pool_size]))
+    (fun () -> ignore(add_host __context "badhost"));
+
+  (* License the pool *)
+  let pool = Db.Pool.get_all ~__context |> List.hd in
+  Db.Pool.set_restrictions ~__context ~self:pool ~value:(Features.to_assoc_list [Features.Pool_size]);
+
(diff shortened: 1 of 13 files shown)


ocaml/xapi/test_host.ml
```diff
@@ -0,0 +1,84 @@
...
+    (Api_errors.Server_error(Api_errors.license_restriction, [Features.name_of_feature Features.Pool_size]))
+    (fun () -> add_host __context "badhost");
+
+  (* License the pool *)
+  let pool = Db.Pool.get_all ~__context |> List.hd in
+  Db.Pool.set_restrictions ~__context ~self:pool ~value:(Features.to_assoc_list [Features.Pool_size]);
+
+  (* Adding hosts should now work *)
+  add_host __context "goodhost";
+  assert_equal ~msg:"one added OK" 4 (Db.Host.get_all ~__context |> List.length)
```
jonludlam: Not 4, but Xapi_globs.restricted_pool_size + 1

ocaml/xapi/xapi_host.ml
```diff
@@ -595,10 +595,10 @@ let is_host_alive ~__context ~host =
...
 
   (* fail-safe. We already test this on the joining host, but it's racy, so multiple concurrent
-  pool-join might succeed. *)
-  if pool_size_is_restricted ~__context &&
-    List.length (Db.Host.get_all ~__context) >= Xapi_globs.restricted_pool_size then
-    raise (Api_errors.Server_error(Api_errors.license_restriction, [Features.name_of_feature Features.Pool_size]));
+     pool-join might succeed. Note: we do it in this order to avoid a problem checking restrictions during
+     the initial setup of the database *)
+  if List.length (Db.Host.get_all ~__context) >= Xapi_globs.restricted_pool_size && pool_size_is_restricted ~__context
+  then raise (Api_errors.Server_error(Api_errors.license_restriction, [Features.name_of_feature Features.Pool_size]));
```
jonludlam: This change should be squashed into the previous commit

ocaml/xapi/suite.ml
```diff
@@ -69,6 +69,7 @@ let base_suite =
     Test_network_event_loop.test;
     Test_network.test;
     Test_pusb.test;
+    Test_host.test;
```
gaborigloi: Should we use alcotest for new tests (suite_alcotest.ml)?

jonludlam: Hmm, I think you might be right.

gaborigloi: API doc is here: https://mirage.github.io/alcotest/alcotest/Alcotest/index.html
And here are some examples:
* https://github.com/xapi-project/xen-api/pull/3421
* https://github.com/xapi-project/xen-api/pull/3420

ocaml/xapi/test_host.ml
```diff
@@ -0,0 +1,84 @@
...
+
+open Stdext
+open OUnit
+open Test_common
+
+module D=Debug.Make(struct let name="test_xapi_xenops" end)
+open D
+
+let add_host __context name =
+  ignore(Xapi_host.create ~__context
```
gaborigloi: It's not important, but I think we could use `Test_common.make_host ~__context ~name_label:name ()` here

ocaml/xapi/test_host.ml
```diff
@@ -0,0 +1,84 @@
...
+  assert_raises ~msg:"Should fail"
+    (Api_errors.Server_error(Api_errors.license_restriction, [Features.name_of_feature Features.Pool_size]))
+    (fun () -> add_host __context "badhost");
+
+  let new_vm = Test_common.make_vm ~__context ~name_label:"My test VM-CVM" () in
+  Db.VM.set_is_control_domain ~__context ~self:new_vm ~value:true;
+
+  (* Adding hosts should now work *)
+  add_host __context "goodhost";
+  assert_equal ~msg:"one added OK" 4 (Db.Host.get_all ~__context |> List.length)
```
gaborigloi: Same as above, Xapi_globs.restricted_pool_size + 1

ocaml/xapi/xapi_host.ml
```diff
@@ -66,6 +66,19 @@ let assert_safe_to_reenable ~__context ~self =
...
+  let cvm_exception =
+    let dom0 = Helpers.get_domain_zero ~__context in
+    Db.VM.get_records_where ~__context
+      ~expr:(Eq(Field "is_control_domain", Literal "true"))
+    |> List.exists (fun (vmref, vmrec) ->
+        (vmref <> dom0) &&
+        (Xapi_stdext_std.Xstringext.String.endswith "-CVM" vmrec.API.vM_name_label)
+      )
+  in
+  (not cvm_exception) && (not (Pool_features.is_enabled ~__context Features.Pool_size))
```
gaborigloi: So this means that we can have arbitrarily many VMs if we have a CVM but not the pool size feature?

jonludlam: arbitrarily many hosts, yes.

ocaml/xapi/test_host.ml
jonludlam: True, but given this is function I actually want to test, I thought I'd call it explicitly.

gaborigloi: Good idea, I agree.

jonludlam:
@gaborigloi using alcotest now!

ocaml/xapi/xapi_host.ml
gaborigloi: Oh ok - though this was not intentional, because I found this behaviour surprising.

ocaml/xapi/xapi_host.ml
```diff
@@ -66,6 +66,19 @@ let assert_safe_to_reenable ~__context ~self =
     List.iter (fun self -> Xapi_pif.abort_if_network_attached_to_protected_vms ~__context ~self) unplugged_pifs;
   end
 
+(* The maximum pool size allowed must be restricted to 3 hosts for the pool which does not have Pool_size feature *)
```
gaborigloi: Or `must be restricted to [Xapi_globs.restricted_pool_size] hosts`, in case we change that variable

PR #3428, opened 2018-02-02
 VDI.snapshot, clone: pass down complete vdi_info 

gaborigloi:
We use vdi_info_of_vdi_rec to ensure that the vdi_info we pass down to
SMAPIv2 contains the correct information about the VDI.
Previously when we were debugging an issue, we were looking at the
vdi_info printed out by SMAPIv2, and it was misleading as many of the
fields, such as is_a_snapshot and cbt_enabled, were incorrect because
they were just taken from the default_vdi_info.
I determine the snapshot time in SMAPIv2, to ensure that in the vdi_info
we only have to pass down the fields of the snapshotted VDI, not the new
fields of the snapshot. This means that SMAPIv3 plugins will also have
to set the snapshot time correctly, instead of relying on the old
behaviour.

### DIFF
--- a/ocaml/xapi/storage_access.ml
+++ b/ocaml/xapi/storage_access.ml
@@ -67,9 +67,9 @@ let find_content ~__context ?sr name =
 
 let vdi_info_of_vdi_rec __context vdi_rec =
   let content_id =
-    if List.mem_assoc "content_id" vdi_rec.API.vDI_other_config
-    then List.assoc "content_id" vdi_rec.API.vDI_other_config
-    else vdi_rec.API.vDI_location (* PR-1255 *)
+    try
+      List.assoc "content_id" vdi_rec.API.vDI_other_config
+    with Not_found -> vdi_rec.API.vDI_location (* PR-1255 *)
   in {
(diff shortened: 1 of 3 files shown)


ocaml/xapi/storage_access.ml
```diff
@@ -65,6 +65,33 @@ let find_content ~__context ?sr name =
        || (vdi_rec.API.vDI_location = name) (* PR-1255 *)
     ) all
 
+let vdi_info_of_vdi_rec __context vdi_rec =
+  let content_id =
+    if List.mem_assoc "content_id" vdi_rec.API.vDI_other_config
+    then List.assoc "content_id" vdi_rec.API.vDI_other_config
+    else vdi_rec.API.vDI_location (* PR-1255 *)
```
minishrink: Could this be more efficient as try-with statement? i.e. `try List.assoc [...] with _ -> vdi_rec.API.vDI_location`

gaborigloi: Yes, that's safer, I think Jon also prefers that style, but we should  catch the `Not_found` exception. Didn't spot it as I just moved around existing code.

PR #3429, opened 2018-02-02
CA-266936: Use a CArray instead of an ocaml Buffer

mseri:
The original code was passing an ocaml managed Buffer as
a buffer for `pciutils` to write the pci names. This can
have unexpected consequences, and in our case was causing
(rare and harmless) segfaults.

As recommended also in [ctypes FAQ](https://github.com/ocamllabs/ocaml-ctypes/wiki/FAQ#-whats-the-best-way-to-bind-c-functions-that-act-on-strings)
we can solve this issue by allocating a C array (that lives
outside the ocaml heap) and use it as the string buffer
expected by the C function. This is still managed by ocaml
and will be freed by the GC once it is no longer in use
but it will not change unexpectedly under our feet.

Signed-off-by: Marcello Seri <marcello.seri@gmail.com>

### DIFF
--- a/ocaml/pci/bindings/ffi_bindings.ml
+++ b/ocaml/pci/bindings/ffi_bindings.ml
@@ -75,8 +75,8 @@ module Bindings (F : Cstubs.FOREIGN) = struct
     let (-:) ty label = field pci_cap label ty
     let next = (ptr_opt pci_cap) -: "next"
     let id = uint16_t -: "id"
-    let type_ = uint16_t -: "type" 
-    let addr = int -: "addr" 
+    let type_ = uint16_t -: "type"
+    let addr = int -: "addr"
     let () = seal pci_cap
 
     type t = pci_cap structure ptr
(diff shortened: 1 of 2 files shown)


ocaml/pci/bindings/ffi_bindings.ml
```diff
@@ -264,15 +264,15 @@ module Bindings (F : Cstubs.FOREIGN) = struct
 
   let pci_lookup_name_1_ary =
     foreign "pci_lookup_name"
-      (Pci_access.t @-> string @-> int @-> int @-> int @-> returning string_opt)
```
edwintorok: There are a few more occurences of functions taking strings and returning strings. Are they safe, or we just don't use them?

ocaml/pci/lib/pci.ml
```diff
@@ -73,8 +73,8 @@ let maybe f = function
 let scan_bus = B.pci_scan_bus
 
 let with_string ?(size=1024) f =
```
edwintorok: Maybe add a comment here explaining why `string` is not suitable linking to the FAQ.
IIUC using `string` violates this rule from the FAQ: `string unsuitable for binding to C functions that write into the string,`. 

The caveats from `ocaml_string` do not apply (we do not release the runtime lock (the default) or call back into OCaml from these functions), because `string` actually does a copy, however IIUC the lifetime of the value returned by libpci is bound the the lifetime of the input parameter, which can be moved by the GC and cause problems.

ocaml/pci/bindings/ffi_bindings.ml
mseri: These are the only one that write in the buffer if I am correct, so the other should be fine as they are

ocaml/pci/lib/pci.ml
mseri: That is exactly right. I am going to steal your comment and add it as a comment

PR #3435, opened 2018-02-07
RFC - CA-281002 CA-271406 let XSM+vGPU fail if VM reboots

lindig:
When a VM shuts down or reboots during storage migration, the VGPU field
scheduled_to_be_resident_on is reset. This makes it impossible to
compute the VGPU map required for the migration. With this commit, we
don't try to recover from this but let the migration fail.

* CA-281002
* CA-271406


lindig:
This was accepted into Jura and should be accepted here to avoid that the branches diverge.

PR #3439, opened 2018-02-09
CP-26685 Adapt network interface when porting to ppx

krizex:
1. update idl call
2. update exception

This PR should goes in with https://github.com/xapi-project/xcp-idl/pull/189

### DIFF
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -65,8 +65,8 @@ let get_master ~__context =
 
 let get_management_iface_is_connected ~__context =
   let dbg = Context.string_of_task __context in
-  Net.Bridge.get_physical_interfaces dbg ~name:(Xapi_inventory.lookup Xapi_inventory._management_interface)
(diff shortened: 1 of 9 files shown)


minishrink:
Could you rename this PR since it doesn't contain changes to `xcp-idl`, but instead updates Xapi with your `xcp-idl` changes?

krizex:
Update the PR title and commit message accordingly.

krizex:
Ring3 BVT passed in job 79965
Ring3 BST passed job 79962
Network regression passed job 80071 except infrastructure issues of bond cases on BOURNE04, I have spoke it to Rama and got confirmed, the ticket has been assigned to him.

mseri:
I understand that the implementation does not support named arguments, but I thought that the interface does, am I wrong @jonludlam. Can we keep the named arguments here?

jonludlam:
No, I'm afraid the named parameters disappear from both client functions and server implementations. I think the only way to bring them back is to do code-gen ctypes-style, and I'd quite like to avoid that.

mseri:
I am going to do a manual merge to master

PR #3443, opened 2018-02-11
CA-271772: Improve time of computing HA max failure tolerance

fillzero:
XAPI uses bisect to pass fail-host-number argument, if we have N hosts
in pool, the total calculation executed times is log(N) + 1.

Each time, XAPI does the same operations:
1. Query DB
2. Choose heuristic algorithm
3. Execute algorithm to calculate max fail tolerance

Given step 1 doesn't depend on fail-host-number argument.
So cache the query db result to improve the performance.

Signed-off-by: Liang Dai <liang.dai1@citrix.com>

### DIFF
--- a/.travis-opam-coverage.sh
+++ b/.travis-opam-coverage.sh
@@ -30,6 +30,8 @@ sudo apt install -y libpci-dev
 opam pin add --no-action xapi .
 opam depext --yes xapi
 opam install --deps-only xapi
+# install the test dependencies
+opam install -y alcotest
 
 # build and test xapi with coverage, then submit the coverage information to coveralls
 
--- a/.travis-opam-coverage.sh
+++ b/.travis-opam-coverage.sh
@@ -30,8 +30,6 @@ sudo apt install -y libpci-dev
 opam pin add --no-action xapi .
 opam depext --yes xapi
 opam install --deps-only xapi
-# install the test dependencies
-opam install -y alcotest
 
 # build and test xapi with coverage, then submit the coverage information to coveralls
 
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ clean:
 	jbuilder clean
 
 test:
-	jbuilder runtest --no-buffer
+	jbuilder runtest --no-buffer -j $$(getconf _NPROCESSORS_ONLN)
 
 doc:
 	jbuilder build ocaml/doc/jsapi.exe ocaml/idl/datamodel_main.exe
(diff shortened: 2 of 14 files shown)


lindig:
Before we review this: 

* is this change effective - does it indeed improve performance noticeably? The design discussion so far only covered using a new approximation algorithm.
* Please don't user overlong subject lines in the commit message. The ... at the end of the subject line are in a unicode representation that can cause problems with some of our tools.

fillzero:
The change improves the performance noticeably.

My test environment: 64-hosts pool with 512 VMs running and protected.
Time of query db operation is 0.8 second, and total query times is log(64) + 1 = 7.
We can speed up the process if the result is cached.

By the way, PR https://github.com/xapi-project/xen-api/pull/3416 also improves the performance, 
reduce time complexity from O(n) to O(1).
```
-         ignore (List.fold_left (fun config _ -> simulate_worst_single_failure config) initial_config (mkints initial_config.num_failures));
+         if config.vms <> [] then simulate_failure_approximation config;
```

Please refer below link for detail result: 
https://issues.citrite.net/browse/CA-271772?focusedCommentId=2110810&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2110810

lindig:
How often is this computation required? I understand that the cache is making a difference but even without it, we have quite a performance improvement.

fillzero:
> What is the performance (in seconds) without this cache?

Based on my testing result: https://issues.citrite.net/secure/attachment/1337612/1337612_time_rerun.png
If we have 512 VMs running and protected on a 64-hosts pool, 
here is time to calculate max host failure tolerance for all these VMs:
```
Ring3 master branch: 20.8 seconds
With fix for CA-270622: 9.5 seconds
With fix for CA-270622 + CA-271772: 5.1 seconds
```

> How often is this computation required? 
Generally, this computation is only triggered when user set the VM `HA restart priority` on XenCenter.
But when the pool becomes overcommitted, this computation will be called to tell the user the actual value of `max failure tolerance`.
It will be triggered by any operation which could change the amount of host memory currently in use (VM.start, VM.resume, VM.migrate etc)

lindig:
I would like to propose that we proceed without merging this but to revisit this in case we really need the additional performance.

I haven't looked closely but I believe the actual fault tolerance can be computed efficiently when we know the state of the pool: 

* free memory on all hosts
* number of vms on all hosts
* maximum size of a vm in the pool

mseri:
Did you reach an agreement? Should we close this PR and reopen it if necessary?

lindig:
There is an email conversation about this. I believe we agree to close this. I've copied you in.

PR #3446, opened 2018-02-13
CP-25713: update usb_reset.py for usb pass-through with deprivileged QEMU

TalonsLee:


scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+# ./usb_reset.py attach device -d dom-id -p pid [-r]
+# ./usb_reset.py attach 2-2 -d 12 -p 4130
+# ./usb_reset.py attach 2-2 -d 12 -p 4130 -r
+# 1. reset device
+# if without -r, do step 2~4
+# 2. if it's the first USB device to pass-through
+#      a) bind mount /dev /sys in chroot directory (/var/xen/qemu/<domid))
+#      b) create new cgroup devices:/qemu-<domid>,
+#      c) blacklist all and add default device whitelist,
+#      d) join current qemu process to this cgourp
```
lindig: typo in `cgourp`

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+    cleanup.add_argument("-d", dest="dom_id", type=int, required=True,
+                         help="specify the dom_id of the VM")
+
+    return parser.parse_args()
 
-#USBDEVFS_RESET _IO('U', 20)
-USBDEVFS_RESET = (ord('U') << 8) | 20
 
+def get_root_dir(dom_id):
+    return "/var/xen/qemu/{}/".format(dom_id)
```
lindig: I'm not sure about Python conventions but it might be more natural to not have a trailing `/`. This way, appending a path would become `get_root_dir(id) + "/some/path"` rather than `get_root_dir(id) + "some/path"`.

rosslagerwall:
I think it would have made sense to squash these two changes together before submitting a PR.

scripts/usb_reset.py
TalonsLee: Thanks. It's easy for me to assume abc/ as a dir, and /edf probably a file.

scripts/usb_reset.py
TalonsLee: Done. Thanks.

TalonsLee:
OK. Will squash them. Run some test tomorrow, will push it then.

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
 def read_int(path):
     with open(path) as f:
         return int(f.readline())
 
 
-def reset_device(bus, dev):
-    with open("/dev/bus/usb/{0:03d}/{1:03d}".format(bus, dev), "w") as f:
-        fcntl.ioctl(f.fileno(), USBDEVFS_RESET, 0)
+def dev_path(device):
+    pat = re.compile(r"^\d+-\d+(\.\d+)*$")
```
rosslagerwall: Can you add a comment here with an example of what you're trying to match?

rosslagerwall: Should it be `r"^\d+-\d+(\.\d+)?$"` instead?

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+    if pat.match(device) is None:
+        log.debug("Unexpected device node: {}".format(device))
```
rosslagerwall: Rather use log.error()

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+        exit(1)
+    try:
+        bus = read_int("/sys/bus/usb/devices/{}/busnum".format(device))
+        dev = read_int("/sys/bus/usb/devices/{}/devnum".format(device))
+        return "/dev/bus/usb/{0:03d}/{1:03d}".format(bus, dev)
+    except (IOError, ValueError) as e:
+        log.debug("Failed to get device path {}: {}".format(device, str(e)))
```
rosslagerwall: Rather use log.error()

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+        exit(1)
 
 
-if __name__ == "__main__":
-    log.logToSyslog(level=logging.DEBUG)
+def get_ctl(path, mode):
```
rosslagerwall: What is `ctl`? You could add a docstring.

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+        st = os.stat(path)
+    except OSError:
+        return None
+    t = ""
+    if S_ISBLK(st.st_mode):
+        t = "b "
+    elif S_ISCHR(st.st_mode):
+        t = "c "
+    if t and (mode == "r" or mode == "rw"):
+        return t + str(os.major(st.st_rdev)) + ":" + str(os.minor(
```
rosslagerwall: I think this is cleaner:

    "%s%d:%d %s" % (t, os.major(st.st_rdev), os.minor(st.st_rdev), mode)

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+    try:
+        st = os.stat(path)
+    except OSError:
+        return None
+    t = ""
+    if S_ISBLK(st.st_mode):
+        t = "b "
+    elif S_ISCHR(st.st_mode):
+        t = "c "
+    if t and (mode == "r" or mode == "rw"):
```
rosslagerwall: ... and mode in ("r", "rw"):

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+        exit(1)
 
 
-if __name__ == "__main__":
-    log.logToSyslog(level=logging.DEBUG)
+def get_ctl(path, mode):
+    try:
+        st = os.stat(path)
+    except OSError:
+        return None
```
rosslagerwall: Why are we ignoring errors? I think any error setting up a device should be fatal.

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+    t = ""
+    if S_ISBLK(st.st_mode):
+        t = "b "
+    elif S_ISCHR(st.st_mode):
+        t = "c "
+    if t and (mode == "r" or mode == "rw"):
+        return t + str(os.major(st.st_rdev)) + ":" + str(os.minor(
+            st.st_rdev)) + " " + mode
+    return None
```
rosslagerwall: Why are we ignoring errors? I think any error setting up a device should be fatal.

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+    cg_dir = get_cg_dir(dom_id)
+    file_name = "devices.allow" if allow else "devices.deny"
+    try:
+        with open(cg_dir + file_name, "w", 0) as f:
+            ctl = get_ctl(path, "rw")
+            if ctl:
+                f.write(ctl)
+    except IOError as e:
+        # log and continue
+        log.debug("Failed to {} {}: {}".format(
```
rosslagerwall: Why are we ignoring errors? I think any error setting up a device should be fatal.

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+
+
+def setup_cgroup(dom_id, pid):
+    cg_dir = get_cg_dir(dom_id)
+
+    try:
+        os.mkdir(cg_dir, 0755)
+    except OSError as e:
+        if e.errno != errno.EEXIST:
+            log.debug("Failed to create cgroup: {}".format(cg_dir))
```
rosslagerwall: `log.error`

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+        with open(cg_dir + "tasks", "w", 0) as tasks, \
+                open(cg_dir + "devices.deny", "w", 0) as deny, \
+                open(cg_dir + "devices.allow", "w", 0) as allow:
+
+            # deny all by default
+            deny.write("a")
 
-    if len(sys.argv) != 2:
-        sys.exit("usage: {} device_node".format(sys.argv[0]))
+            # Only the "/dev/null" is required after CP-25630
```
rosslagerwall: I don't think this comment is helpful since whatever was required before CP-25630 is not relevant anymore.

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+            whitelist = [
```
rosslagerwall: Since there's only one element in this list you could probably collapse it all down to:

    allow.write(get_ctl("/dev/null", "rw"))

(Assuming get_ctl() is changed so that it can't fail.)

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+    except IOError as e:
+        log.debug("Failed to setup cgroup: {}".format(str(e)))
         exit(1)
 
+
+def mount(source, target, fs, flags=0):
+    if ctypes.CDLL(ctypes.util.find_library("c"), use_errno=True
+                   ).mount(source, target, fs, flags) < 0:
+        # log and continue
+        log.debug("Failed to mount {} ({}) to {} with flags {}: {}".
```
rosslagerwall: Why not fatal?

(Also, log.error)

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+                    allow.write(ctl)
 
-    device = sys.argv[1]
-    pattern = re.compile(r"^\d+-\d+(\.\d+)*$")
-    if pattern.match(device) is None:
-        log.debug("unexpected device node: {}".format(device))
+            tasks.write(str(pid))
+
+    except IOError as e:
+        log.debug("Failed to setup cgroup: {}".format(str(e)))
```
rosslagerwall: log.error

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+        log.debug("Failed to mount {} ({}) to {} with flags {}: {}".
+                  format(source, fs, target, flags,
+                         os.strerror(ctypes.get_errno())))
+
+
+def umount(target):
+    if ctypes.CDLL(ctypes.util.find_library("c"), use_errno=True
+                   ).umount(target) < 0:
+        # log and continue
+        log.debug("Failed to umount {}: {}".
```
rosslagerwall: log.error and for the rest of this file...

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+            fcntl.ioctl(f.fileno(), USBDEVFS_RESET, 0)
+    except IOError as e:
+        # log and continue
+        log.debug("Failed to reset {}: {}".format(path, str(e)))
+
+    if reset_only:
+        return
+
+    # set device file uid/gid
+    os.chown(path, pwd.getpwnam("qemu_base").pw_uid + dom_id,
```
rosslagerwall: I think it would be better if the uid & gid should come from the caller. This will allow us to change it more easily in the future (which I think we should do since it currently sometimes overlaps with other users).

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+             grp.getgrnam("qemu_base").gr_gid + dom_id)
+
+    root_dir = get_root_dir(dom_id)
+    dev_dir = root_dir + "dev/"
+    if not os.path.isdir(root_dir) or not os.path.isdir(dev_dir):
+        log.debug("Error: The chroot or dev directory doesn't exist")
```
rosslagerwall: log.error

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+
+    sys_dir = root_dir + "sys/"
+    # sys_dir could already be mounted because of PCI pass-through
+    if not os.path.isdir(sys_dir):
+        os.mkdir(sys_dir, 0755)
+    if not os.path.isdir(sys_dir + "devices/"):
+        mount("/sys", sys_dir, "sysfs")
+
+    # add device to cgroup allow list
+    device_ctl(path, dom_id, True)
```
rosslagerwall: `dom_id` is conventionally written as `domid`. (Throughout this file.)

scripts/usb_reset.py
```diff
@@ -14,46 +14,266 @@
...
+        t = "b "
+    elif S_ISCHR(st.st_mode):
+        t = "c "
+    if t and (mode == "r" or mode == "rw"):
+        return t + str(os.major(st.st_rdev)) + ":" + str(os.minor(
+            st.st_rdev)) + " " + mode
+    return None
+
+
+def device_ctl(path, dom_id, allow):
```
rosslagerwall: This API would be nicer is you renamed this to _device_ctl() and then added two wrappers, allow_device() and deny_device().

scripts/usb_reset.py
TalonsLee: Done.

scripts/usb_reset.py
TalonsLee: Done.

scripts/usb_reset.py
TalonsLee: Done.

scripts/usb_reset.py
TalonsLee: done.

scripts/usb_reset.py
TalonsLee: done.

scripts/usb_reset.py
TalonsLee: Done.

scripts/usb_reset.py
TalonsLee: done.

scripts/usb_reset.py
TalonsLee: "{} {}:{} {}".format(t, os.major(st.st_rdev), os.minor(
            st.st_rdev), mode)

scripts/usb_reset.py
TalonsLee: done.

scripts/usb_reset.py
TalonsLee: done.

scripts/usb_reset.py
TalonsLee: it can be "1-2.1.2".

ls /sys/bus/usb/devices/
1-0:1.0  1-2.1    1-2.1:1.0  1-2.1.2:1.0  2-2      3-0:1.0  3-1:1.0  4-1      usb1  usb3
1-2      1-2:1.0  1-2.1.2    2-0:1.0      2-2:1.0  3-1      4-0:1.0  4-1:1.0  usb2  usb4

TalonsLee: done.

scripts/usb_reset.py
TalonsLee: TODO, requires toolstack change.

scripts/usb_reset.py
TalonsLee: Done.

scripts/usb_reset.py
TalonsLee: done

scripts/usb_reset.py
TalonsLee: done.

scripts/usb_reset.py
TalonsLee: done.

scripts/usb_reset.py
TalonsLee: done.
But I wonder if I have to raise RuntimeError here,
and handle them in attach()

scripts/usb_reset.py
TalonsLee: done.

scripts/usb_reset.py
rosslagerwall: I was thinking you could just call exit(1) like in other places.

If the mount fails, usb_reset.py should fail and the VM should not start.

scripts/usb_reset.py
rosslagerwall: How many `(\.\d+)` segments can there be? Is it really unlimited (as you had in your original regexp)? You can do something like `(\.\d+){0,2}` to match a more specific number.

Also note that since you're using `match()`, you don't need to start the regexp with a `^`.

scripts/usb_reset.py
rosslagerwall: Yeah, that is fine.

scripts/usb_reset.py
lindig: I think this should be covered in a separate PR such that we can merge this.

rosslagerwall: Sure, that's fine.

scripts/usb_reset.py
mseri: Maybe the change has been lost when squashing? I still see `log.debug`

scripts/usb_reset.py
mseri: All these comments appear now as resolved but I think they are only hidden due to the new commit. This still needs and answer

scripts/usb_reset.py
mseri: Docstring is still missing

ocaml/xapi/xapi_db_upgrade.ml
```diff
@@ -521,6 +522,21 @@ let upgrade_vswitch_controller = {
         ignore (Xapi_sdn_controller.introduce ~__context ~protocol:`ssl ~address ~port:6632L)
 }
 
+let default_vm_platform_device_model = {
+  description = "Initialising unset VM.platform.device-model profiles";
+  version = (fun x -> x < jura);                                     (* initialise only for upgrades from previous versions before jura *)
```
Zhengchai: Would this be better if we use version number instead of project/release name?

scripts/usb_reset.py
```diff
@@ -14,46 +14,273 @@
...
 # 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 #
-# ./usb_reset.py 1-2
-# reset USB device 1-2
+# attach
+# ./usb_reset.py attach device -d dom-id -p pid [-r]
+# ./usb_reset.py attach 2-2 -d 12 -p 4130
+# ./usb_reset.py attach 2-2 -d 12 -p 4130 -r
+# 1. reset device
+# if without -r, do step 2~4
```
Zhengchai: you may need one comment line to explain the [-r] usage?

scripts/usb_reset.py
```diff
@@ -14,46 +14,273 @@
...
 
-def reset_device(bus, dev):
-    with open("/dev/bus/usb/{0:03d}/{1:03d}".format(bus, dev), "w") as f:
-        fcntl.ioctl(f.fileno(), USBDEVFS_RESET, 0)
+def dev_path(device):
+    # check device node pattern
+    # example: "4-1", "1-2.1.2"
+    pat = re.compile(r"^\d+-\d+(\.\d+)*$")
+    if pat.match(device) is None:
+        log.error("Unexpected device node: {}".format(device))
```
Zhengchai: where is the log prints into? i.e. which logfile in Dom0?

scripts/usb_reset.py
```diff
@@ -14,46 +14,273 @@
...
+        bus = read_int("/sys/bus/usb/devices/{}/busnum".format(device))
+        dev = read_int("/sys/bus/usb/devices/{}/devnum".format(device))
+        return "/dev/bus/usb/{0:03d}/{1:03d}".format(bus, dev)
+    except (IOError, ValueError) as e:
+        log.error("Failed to get device path {}: {}".format(device, str(e)))
+        exit(1)
 
 
-if __name__ == "__main__":
-    log.logToSyslog(level=logging.DEBUG)
```
Zhengchai: if this line is removed, where does the logs go into?

scripts/usb_reset.py
```diff
@@ -14,46 +14,273 @@
...
+# get the string to control device access for cgroup
+def get_ctl(path, mode):
+    try:
+        st = os.stat(path)
```
Zhengchai: looks this script needs to be executed in the privileged model in Dom0?

scripts/usb_reset.py
```diff
@@ -14,46 +14,273 @@
...
+    dev_dir = root_dir + "/dev"
+    sys_dir = root_dir + "/sys"
+    if os.path.isdir(dev_dir + "/bus"):
+        umount(dev_dir)
+    if os.path.isdir(sys_dir + "/devices"):
+        umount(sys_dir)
+
+
+if __name__ == "__main__":
+    log.logToSyslog(level=logging.DEBUG)
```
Zhengchai: OK, it is here.

scripts/usb_reset.py
TalonsLee: yes. by xapi.

scripts/usb_reset.py
TalonsLee: done.

scripts/usb_reset.py
TalonsLee: done.

scripts/usb_reset.py
TalonsLee: At most 5 hubs can be chained between controller and device. So I guess the longest one could be something like "1-2.1.2.1.2.1" ,So \d+-\d+(\.\d+){0,5}    .

scripts/usb_reset.py
TalonsLee: can I log and exit(1)?

scripts/usb_reset.py
rosslagerwall: Thanks. Since the number of chains can be quite large I think simply using `*` is fine.

scripts/usb_reset.py
rosslagerwall: You could, but it would be better practice to log and raise an exception.

rosslagerwall:
Why are several of Marcus's changes from October appearing in this PR?

scripts/usb_reset.py
TalonsLee: OK.    So pat = re.compile(r"\d+-\d+(\.\d+)*$")

taoyongd:
@TalonsLee , I think you should rebase your branch.

scripts/usb_reset.py
TalonsLee: I don't like to re-throw the exception, because we don't handle it somewhere(level) else. Any error is fatal.

rosslagerwall: It looks to me that both callers, `_device_ctl()` and `setup_cgroup()` catch the exception and exit appropriately.

scripts/usb_reset.py
TalonsLee: it's explained at line 22 and 69

scripts/usb_reset.py
TalonsLee: If so, do we have to raise another run-time exception at the end of get_ctrl, instead of exit(1)?

rosslagerwall: Yes, that's probably better.

lindig:
I will merge this when @rosslagerwall agrees,

scripts/usb_reset.py
```diff
@@ -14,46 +14,283 @@
...
+def get_ctl(path, mode):
+    """get the string to control device access for cgroup
+    :param path: the device file path
+    :param mode: either "r" or "rw"
+    :return: the string to control device access
+    """
+    try:
+        st = os.stat(path)
+    except OSError as e:
+        raise
```
rosslagerwall: It doesn't make any sense to catch the exception and then reraise it without doing anything. Why not just remove the try...except?

rosslagerwall: Alternatively, this would make sense if you also logged an error message in the exception handler.

scripts/usb_reset.py
```diff
@@ -14,46 +14,283 @@
...
 
-    device = sys.argv[1]
-    pattern = re.compile(r"^\d+-\d+(\.\d+)*$")
-    if pattern.match(device) is None:
-        log.debug("unexpected device node: {}".format(device))
+def _device_ctl(path, domid, allow):
+    cg_dir = get_cg_dir(domid)
+    file_name = "/devices.allow" if allow else "/devices.deny"
+    try:
+        with open(cg_dir + file_name, "w", 0) as f:
```
rosslagerwall: Why are you opening this unbuffered? Since the file is automatically written and closed when execution goes outside of the with statement context, setting it as unbuffered seems pointless.

scripts/usb_reset.py
TalonsLee: OK. Log error message in the exception handler here.

scripts/usb_reset.py
TalonsLee: It may seems pointless now, because only one write per file now.
I was writing multiple items(the white list),
it won't work if it's not buffered according to my test.
So I guess it's better to keep it unbuffered.

FYI: libvirt do virFileWriteStr(), it open, write only one item, and close.

rosslagerwall: This code is also set up to write a single item and then close the file (even if you had multiple calls to _device_ctl) so I still think it is unnecessary. If you want to keep it, please add a comment which says why it is needed.

TalonsLee: how about I use buffered in _device_ctl,
and keep the unbuffered in setup_cgroup?

rosslagerwall: > how about I use buffered in _device_ctl,
and keep it unbuffered in setup_cgroup?

Sure, with a comment in setup_cgroup().

mseri: @TalonsLee are you sure that it's not enough to add `f.flush()` and `os.fsync()` to enforce syncing the changes as well?

rosslagerwall: Just an `f.flush()` should be enough (this is a pseudo filesystem so there is nothing to sync to disk), although it is arguably simpler to just set it as unbuffered.

mseri: For small writes I agree with you

lindig:
@mseri please merge this when you think this is ready.

scripts/usb_reset.py
```diff
@@ -14,46 +14,285 @@
...
+def _device_ctl(path, domid, allow):
+    cg_dir = get_cg_dir(domid)
+    file_name = "/devices.allow" if allow else "/devices.deny"
+    try:
+        with open(cg_dir + file_name, "w") as f:
+            f.write(get_ctl(path, "rw"))
+    except (IOError, OSError, RuntimeError) as e:
+        log.error("Failed to {} {}: {}".format(
+            "allow" if allow else "deny", path, str(e)))
         exit(1)
```
TalonsLee: Consider xapi might call detach after cleanup,
in which case, the cgroup dir is already removed,
we might want to not call this exit(1) here.

taoyongd: There's a CA ticket to describe the case we met. https://issues.citrite.net/browse/CA-284857.

PR #3447, opened 2018-02-13
Reorg datamodel a little

jonludlam:
Move all objects into their own modules. Next step: move them into their own files. I'm submitting this incrementally as otherwise the rebasing will be very painful.


robhoes:
Slightly higher hanging fruit?

PR #3448, opened 2018-02-14
Fix kpartx arguments in xe-edit-bootloader

BobBall:
For reference, see https://github.com/openSUSE/multipath-tools/blob/master/kpartx/kpartx.c#L114
kpartx will attempt to determine the delimiter based on the device name - my reading is if we have a device name that ends in a digit, a 'p' delimiter will be automatically added by default. If the device name ends in any other character the delimiter is not added by default.

The only way to make this consistent is to request a delimiter on the command line (https://github.com/openSUSE/multipath-tools/blob/master/kpartx/kpartx.c#L382 will set delim != NULL)

### DIFF
--- a/scripts/xe-edit-bootloader
+++ b/scripts/xe-edit-bootloader
@@ -37,7 +37,7 @@ while getopts "hu:n:p:f:" opt ; do
     h) usage ;;
     u) vm_uuid=${OPTARG} ;;
     n) vm_name="${OPTARG}" ;;
-    p) device_number=${OPTARG} ;;
+    p) device_number=p${OPTARG} ;;
     f) default_file_list="${OPTARG}" ;;
     *) echo "Invalid option"; usage ;;
     esac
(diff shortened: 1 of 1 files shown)


BobBall:
Please do not merge until @davisaaronj has confirmed if this approach fixes the issue for Rackspace as well

BobBall:
I kept p because it's the default used internally in kpartx and the same as the original.
I'd worry that using special characters might cause problems in this - or other - scripts.  For example : may need to be escaped in the mount command and others?

cardoe:
I've also followed the code in kpartx and I agree with @BobBall's logic in the original commit message.

cardoe:
C's `isdigit()` only returns `true` if the value is between 0-9 (base 10) and since this is a UUID we are using base 16 so 0-f.

davisaaronj:
@BobBall Your change looks good to me.  I created #3450 to also account for the files for grub2 since some OS are now using that.

BobBall:
I will cancel this pull request based on the above comment - we should track the final fixes in 3450.

cardoe:
Bob, if its easier re-open this and merge this in and then #3450 can be rebased on top.

PR #3449, opened 2018-02-14
Update readme and maintainer info

minishrink:
This PR seeks to update the information a newcomer would see when looking to either build or contribute to xapi, such as:

- the maintainers list (switched out for mailing list address)
- the OCaml version Xapi builds on
- build instructions (may be worth removing these too?)

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -10,7 +10,7 @@ Xen-enabled hosts and resource pools, and co-ordinates resources
 within the pool. Xapi exposes the Xen API interface for many
 languages and is a component of the XenServer project.
 Xen API is written mostly in [OCaml](http://caml.inria.fr/ocaml/)
-4.01.0.
+4.04.2
 
 Xapi is the main component produced by the Linux Foundation's
 [Xapi Project](http://xenproject.org/developers/teams/xapi.html).
--- a/README.markdown
+++ b/README.markdown
@@ -37,20 +37,6 @@ more time for review.
 Maintainers
 -----------
 
-This is a short list of people who currently maintain this project.
-
-* Christian Lindig <christian.lindig@citrix.com>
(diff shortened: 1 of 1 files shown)


README.markdown
```diff
@@ -21,7 +21,7 @@ Build and Install
 The build install instructions are currently being written. The Xapi
 Project contains a large list of dependencies and sub-projects, which
 are actually quite difficult to build independently. To build xen-api
-from source, we recommend using [xenserver-core](https://github.com/xenserver/xenserver-core).
+from source, we recommend using [dune (previously named jbuilder)](https://github.com/ocaml/dune)
```
mseri: You have to use dune as the build tool, but that would not be enough. I think here we should say opam using he remote provided by xs-opam, and link to the xs-opam readme

README.markdown
```diff
@@ -22,6 +22,7 @@ The build install instructions are currently being written. The Xapi
 Project contains a large list of dependencies and sub-projects, which
 are actually quite difficult to build independently. To build xen-api
 from source, we recommend using [dune (previously named jbuilder)](https://github.com/ocaml/dune)
+and [opam](https://opam.ocaml.org/doc/Manual.html) using the [xs-opam](https://github.com/xapi-project/xs-opam) remote.
```
mseri: I'd completely get rid of the dune line, it is a necessary requirement not a suggestion. Just write

> To build xen-api from source, we recommend using [opam](https://opam.ocaml.org/doc/Manual.html) with the [xs-opam](https://github.com/xapi-project/xs-opam) remote, refer to its [README](https://github.com/xapi-project/xs-opam/blob/master/README.md#adding-the-opam-repository) for a more detailed explanation.

PR #3450, opened 2018-02-14
Includes changes from #3448 updated grub file list to account for grub2.

davisaaronj:
Changes in #3448 look good.  I also updated the file list to account for grub2.  

Signed-off-by: Aaron Davis <aaron.davis@rackspace.com>

### DIFF
--- a/scripts/xe-edit-bootloader
+++ b/scripts/xe-edit-bootloader
@@ -37,7 +37,7 @@ while getopts "hu:n:p:f:" opt ; do
     h) usage ;;
     u) vm_uuid=${OPTARG} ;;
     n) vm_name="${OPTARG}" ;;
-    p) device_number=${OPTARG} ;;
+    p) device_number=p${OPTARG} ;;
     f) default_file_list="${OPTARG}" ;;
     *) echo "Invalid option"; usage ;;
     esac
--- a/scripts/xe-edit-bootloader
+++ b/scripts/xe-edit-bootloader
@@ -84,7 +84,7 @@ function cleanup {
       rmdir ${mnt}
    fi
 
-   kpartx -d ${device}
+   kpartx -dvspp ${device}
 
    if [ ! -z "${vbd_uuid}" ]; then
       echo -n "Unplugging VBD: "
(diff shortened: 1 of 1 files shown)


scripts/xe-edit-bootloader
```diff
@@ -15,7 +15,7 @@ XE="@BINDIR@/xe"
 
 device_number=
 grub_conf=
-default_file_list="/boot/grub/menu.lst /grub/menu.lst /boot/grub/grub.cfg /grub/grub.cfg"
+default_file_list="/boot/grub/menu.lst /grub/menu.lst /boot/grub/grub.cfg /grub/grub.cfg /boot/grub2/grub.cfg /grub2/grub.cfg"
```
BobBall: If this list is going to be expanded, we should include the whole list from pygrub:
https://github.com/mirage/xen/blob/master/tools/pygrub/src/pygrub#L428
We should also make sure the list is in the same order (so the correct config file will be used if multiple config files are found)
We should therefore use:
default_file_list="/boot/grub/grub.cfg /grub/grub.cfg /boot/grub2/grub.cfg /grub2/grub.cfg"
default_file_list="$default_file_list /boot/isolinux/isolinux.cfg /boot/extlinux/extlinux.conf /boot/extlinux.conf /extlinux/extlinux.conf /extlinux.conf"
default_file_list="$default_file_list /boot/grub/menu.lst /boot/grub/grub.conf /grub/menu.lst"

BobBall:
I'm confused by this change.  In #3448 I had two commits - one to revert #3414 and one to correct the kpartx arguments.
In this pull request, we seem to have only included the second commit (fix to kpartx arguments) - but reading the code in commit, github is still displaying the 'p' in this pull request, but not in the target master branch.

I suspect this is github getting confused - but the pull request actually does still need to revert #3414 yet it is not doing so?

davisaaronj:
@BobBall I had some had some help from @cardoe but I think I now have the commits in a better order.  Please let me know if this change is acceptable.

BobBall:
LGTM, thanks for the updates :)

PR #3451, opened 2018-02-15
Use guarded matching instead of matching exact error message strings

minishrink:
Warning 52 is triggered by match statements like `with (Failure "hd") -> [...]`.
This commit changes these match statements to guarded matches with catch-all cases,
satisfying the compiler warning without changing functionality.

### DIFF
--- a/ocaml/database/redo_log.ml
+++ b/ocaml/database/redo_log.ml
@@ -375,14 +375,14 @@ let action_write_db marker generation_count write_fn sock datasockpath =
          write_fn datasock;
          R.debug "Finished writing database to data socket";
        with
-       | Sys_error("Connection reset by peer") ->
+       | Sys_error(x) when x=("Connection reset by peer") ->
          (* CA-41914: Note that if the block_device_io process internally
           * throws Timeout (or indeed any other exception), it will forcibly
           * close this connection, we'll see a Sys_error("Connection reset by
           * peer"). This can be safely suppressed because we'll hear all the
           * gory details in the response we read over the control socket. *)
          R.warn "I/O process forcibly closed the data socket while trying to write database to it. Await the response to see why it did that.";
-       | e ->
+       | Sys_error(_) as e | e ->
          (* We'll re-raise other exceptions, though. *)
          R.error "Got an unexpected exception while trying to write database to the data socket: %s. Re-raising." (Printexc.to_string e);
          raise e
--- a/ocaml/xapi/cli_operations.ml
+++ b/ocaml/xapi/cli_operations.ml
@@ -695,10 +695,11 @@ let make_param_funs getall getallrecs getbyuuid record class_name def_filters de
         try
           set v
         with
-        (* XXX: -- warning 52 -- this might break with new ocaml compilers *)
-          (Failure "int_of_string") -> failwith ("Parameter "^k^" must be an integer")
-        | (Failure "float_of_string") -> failwith ("Parameter "^k^" must be a floating-point number")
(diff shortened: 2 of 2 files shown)


robhoes:
>If your code raises this warning, you should not change the way you test for the specific string to avoid the warning (for example using a string equality inside the right-hand-side instead of a literal pattern), as your code would remain fragile. You should instead enlarge the scope of the pattern by matching on all possible values.

(https://caml.inria.fr/pub/docs/manual-ocaml/comp.html#ss%3Awarn52)

lindig:
This is a hack to silence the compiler but does not address the real problem: matching against strings in exceptions. However, there is little we can do about that in the short term and I would be happy to accept this to bring down the noise in compiler warnings.

robhoes:
I'm not in favour of changing the code to avoid warnings without actually fixing the issue that is being warned about. Then we might as well tell the compiler to disable the warning.

There _should_ be something that we can do about this. For the `Failure` exceptions, I think the way to do it is to catch them earlier, by push the try…with further inside, directly around the calls to `int_of_string` etc. But I'm not sure how easy that is.

mseri:
For the `Sys_error` change, unfortunately there is no good solution that I am aware of. I think this change is reasonable in that case as we are limited by the poor reporting of the ocaml `Sys` module.

For the `set` change, this is trickier. I had tried to fix it appropriately in the past (checking the various `int_of`, `bool_of`, ...) and as far as I remember the code became a giant unreadable mess due to various heuristics that couldn't be easily fixed without a greater rewrite. We could decide to leave the warning on in the hope that somebody decides to try and rewrite the functions completely, but I think this is risky and unlikely to happen, or we could accept this and at least get rid of the warnings (until one day a breaking change will break it and we will be forced to rewrite it properly).

mseri:
To be honest, trying to look for the `redo_log` and the `write_fn`, I could not find code that uses the `Sys` module and justifies those exception matches. I wonder if they are an historic leftover and we can replace them with proper matches on `Unix.error` instead.

The inability to track down what exception can be expected for a function is extremely limiting in these cases. I hope effects will help on this front

robhoes:
The redo-log exception handler comes from cf1ecbb7325fa699ae1e406ea778aa65956b4b08.

robhoes:
I think that the functions inside `Db_xml.To.fd` can result in the `Sys_error`.

mseri:
Yes, I was looking at those, but if you look at the whole module it seems to me that it uses `Unix` or `Xmlm` functions, and both of those have proper errors in the form of `Xmlm_error` and `Unix.error`. There are some small uses of `Unixext` around but not of `https://github.com/xapi-project/stdext/blob/master/lib/xapi-stdext-unix/unixext.ml#L252` that could be fixed to raise a unix error as well.

That is mostly why it is surprising to see the `Sys_error` and it is annoying to not be able to get the graph of possible exceptions from the compiler (that in some ways knows about it)

mseri:
Mostly I think the fault is due to the fact that functions using `in_channel` and `out_channel` are doomed to potentially raise `Sys_errors`, probably to avoid linking them to the `Unix` library. It could be that looking for those could help us clean this up, but it is probably a giant amount of work.

robhoes:
Yes, I think it is related to the use of channels, possibly inside Xmlm. Even though Xmlm defines its own exceptions, it may still raise others.

This is the problem with exceptions. Yes, I've also always wanted a way to know all possible exceptions that a function can raise. It shouldn't be too hard for a tool to annotate functions with exceptions if you have all the sources...

lindig:
Effect inference (what exceptions can be raised) is actually a difficult problem unless you annotate functions with exceptions like in Java.

lindig:
We can't agree on this - closing it.

PR #3452, opened 2018-02-15
Make sure we always go through the same code path during host shutdown

gaborigloi:
The `xapi-domains` service is responsible for performing the necessary actions before the host shuts down. Previously, we went through different code paths depending on whether we shut the host down using XenAPI (`host.shutdown`), or using the `shutdown` command. This PR ensures that the same code is executed in both cases. The only difference is that when using a XenAPI call, some prechecks will be done before shutting down the host to block the operation and notify the user in case of errors.
This PR also includes the OCaml code for VM evacuation that @edwintorok converted into OCaml from a Python script.

Now the new control flow looks like this:
```
                 host.shutdown                host.reboot
                           |                       |
                              shutdown prechecks
                           |                       |
           `/sbin/shutdown -h now`   `/sbin/shutdown -r now`
                           \                       /
         systemd performs ExecStop of `xapi-domains.service`
                                        |
                calls host.prepare_for_shutdown including common code

```

Tested on 1 host, pool slave, pool master, shutdown and evacuation worked fine.


ocaml/xapi/vm_evacuation.ml
```diff
@@ -0,0 +1,101 @@
...
+    vms
+    |> List.filter is_running
+    |> hard_shutdown
+  in
+
+  Client.Host.get_vms_which_prevent_evacuation ~rpc ~session_id ~self:host
+  |> Xapi_stdext_std.Listext.List.filter_map (fun (vm, _) ->
+         if self_managed_poweroff vm then None
+         else Some vm)
+  |> shutdown;
```
gaborigloi: Maybe we should ignore exceptions here?
Or something like:
```
ignore_exn( shutdown vms which prevent evacuation; evacuate);
ignore_exn( shutdown rest of the vms )
```

edwintorok: Yes, we could wrap lines 82-86 into a `log_and_ignore_exn`.

ocaml/idl/datamodel.ml
```diff
@@ -3158,8 +3158,8 @@ let host_reboot = call
     ~allowed_roles:_R_POOL_OP
     ()
 
-let host_prepare_for_shutdown = call
-    ~name:"prepare_for_shutdown"
+let host_prepare_for_poweroff = call
+    ~name:"prepare_for_poweroff"
```
edwintorok: We should probably squash all the rename and refactor commits into one before merging and merge the commits messages having both our signoffs on it.

gaborigloi: Done

ocaml/xapi/vm_evacuation.ml
gaborigloi: Done

PR #3453, opened 2018-02-16
Add SR Multipath capability

MarkSymsCtx:

### DIFF
--- a/ocaml/xapi/smint.ml
+++ b/ocaml/xapi/smint.ml
@@ -30,6 +30,7 @@ type capability =
   | Sr_stats
   | Sr_metadata
   | Sr_trim
+  | Sr_multipath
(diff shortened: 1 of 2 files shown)


jonludlam:
Fancy adding some tests similar to https://github.com/xapi-project/xen-api/commit/4cbad0d2dabe05d54267000f2bfd5098bc131958 ?

MarkSymsCtx:
Yep, will do

edwintorok:
How should XAPI behave when an SR does not support multipath? Should we add some checks when enabling/disabling multipath and plugging the PBDs?

MarkSymsCtx:
This is (at least at the moment) expected to be transparent to Xapi but used by XenCenter

PR #3455, opened 2018-02-16
CA-280981: Import dom0 record on pool join

robhoes:
This record may contain data, e.g. perfmon settings in other-config, that we would like to keep in the pool. Previously, the dom0 record was recreated in the pool after xapi first restart as a slave.


mseri:
Is there any record in `dom0` that can conflict with some new settings from the new pool? Or we will override them anyway?

robhoes:
@mseri This is about the VM record for dom0 ("Control domain on host: xxx"), of which there is one for each host, and contains data belonging to just that host. I can't see anything in there that depends on the pool or other hosts, so I think that it is safe for the joining host to "bring it along" to the pool, rather than having the host recreate it.

PR #3459, opened 2018-02-19
[Cp 26826] fsync XAPI master DB on major lifecycle operations

mcintyre94:
Best reviewed commit-by-commit. :)

- Adds a new mutable field `fsync_enabled` to `db_connection`
- Adds an `fsync` flag to `flush` DB functions and fsync the file descriptor before closing if it's set
- Refactor `preferred_write_db` because it used List.Hd which caused tests to fail after a later commit (and had a 9 year old 'FIXME' :) )
- Set `fsync enabled` in xapi startup and when clustering is enabled/disabled
- Add fsync on pool join when clustering is enabled

### DIFF
--- a/ocaml/database/db_connections.ml
+++ b/ocaml/database/db_connections.ml
@@ -15,8 +15,6 @@ module D = Debug.Make(struct let name = "xapi" end)
 module R = Debug.Make(struct let name = "redo_log" end)
 open D
 
-open Db_cache_types
-
 let get_dbs_and_gen_counts() =
   List.map (fun conn->(Parse_db_conf.generation_read conn, conn)) (Db_conn_store.read_db_connections())
 
@@ -105,10 +103,9 @@ let flush_dirty_and_maybe_exit dbconn ?(fsync=false) exit_spec =
 
 let flush dbconn db =
   debug "About to flush database: %s" dbconn.Parse_db_conf.path;
-  let fsync = dbconn.fsync_enabled in
   Db_conn_store.with_db_conn_lock dbconn
     (fun () ->
-       Backend_xml.flush dbconn db ~fsync
+       Backend_xml.flush dbconn db ~fsync:(dbconn.Parse_db_conf.fsync_enabled)
     )
 
 
--- a/ocaml/database/db_connections.ml
+++ b/ocaml/database/db_connections.ml
@@ -40,8 +40,10 @@ let choose connections = match List.filter exists connections with
     debug "Most recent db is %s (generation %Ld)" most_recent.Parse_db_conf.path gen;
     Some most_recent
 
-let preferred_write_db () =
-  List.hd (Db_conn_store.read_db_connections()) (* !!! FIX ME *)
+let preferred_write_db =
+  match Db_conn_store.read_db_connections() with
+  | [] -> None
+  | db :: _ -> Some db
 
 (* This is set by signal handlers. It instructs the db thread to call exit after the next flush *)
 let exit_on_next_flush = ref false
--- a/ocaml/database/db_connections.ml
+++ b/ocaml/database/db_connections.ml
@@ -15,6 +15,8 @@ module D = Debug.Make(struct let name = "xapi" end)
 module R = Debug.Make(struct let name = "redo_log" end)
 open D
 
+open Db_cache_types
+
 let get_dbs_and_gen_counts() =
   List.map (fun conn->(Parse_db_conf.generation_read conn, conn)) (Db_conn_store.read_db_connections())
 
@@ -44,7 +46,6 @@ let preferred_write_db () =
 (* This is set by signal handlers. It instructs the db thread to call exit after the next flush *)
 let exit_on_next_flush = ref false
 
-
 (* db flushing thread refcount: the last thread out of the door does the exit(0) when flush_on_exit is true *)
 open Xapi_stdext_threads
 let db_flush_thread_refcount_m = Mutex.create()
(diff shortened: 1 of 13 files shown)


ocaml/xapi/xapi_pool_transition.ml
```diff
@@ -131,15 +131,23 @@ let attempt_two_phase_commit_of_new_master ~__context (manual: bool) (peer_addre
...
   (* If this is an automatic transition then there is no other master to clash with and so
      we can restart immediately. NB if we are the master (and this code is being used to assert
      our authoritah on some slaves then we shouldn't restart otherwise we'll get into a restart loop. *)
   if not(manual) then
     if am_master_already
     then info "Not restarting since we are the master already"
-    else Db_cache_impl.flush_and_exit (Db_connections.preferred_write_db ()) Xapi_globs.restart_return_code;
+    else begin
+      let dbconn = Db_connections.preferred_write_db in
+      Xapi_stdext_monadic.Opt.iter (fun db ->
```
edwintorok: I think we should rather fail here if we have no DB connections, `flush_and_exit` is meant to exit XAPI, and if we have no DB connections we would silently not exit XAPI.

ocaml/database/db_connections.ml
```diff
@@ -15,6 +15,8 @@ module D = Debug.Make(struct let name = "xapi" end)
 module R = Debug.Make(struct let name = "redo_log" end)
 open D
 
+open Db_cache_types
```
edwintorok: Could you move this `open`  closer to the code that needs it?
I think only this `dbconn.fsync_enabled` would need it, which can be written as `dbconn.Db_cache_types.fsync_enabled` or `let open Db_cache_types in dbconn.fsync_enabled`

ocaml/database/db_connections.ml
```diff
@@ -102,9 +103,10 @@ let flush_dirty_and_maybe_exit dbconn exit_spec =
 
 let flush dbconn db =
   debug "About to flush database: %s" dbconn.Parse_db_conf.path;
+  let fsync = dbconn.fsync_enabled in
```
edwintorok: Since this variable is only used once you can pass it directly like this:
`Backend_xml.flush dbconn db ~fsync:(dbconn.Db_cache_types.fsync_enabled)`

ocaml/database/db_xml.ml
```diff
@@ -12,6 +12,7 @@
  * GNU Lesser General Public License for more details.
  *)
 open Db_cache_types
+open Xapi_stdext_unix
```
edwintorok: Please move this closer to where its used, i.e. `Xapi_stdext_unix.Unixext.fsync` if it is used in only one place.

ocaml/database/db_xml.ml
```diff
@@ -78,15 +79,23 @@ module To = struct
...
+    if fsync then begin
+      D.debug "fsyncing database before close";
+      Unixext.fsync fd
+    end
 
   let file (filename: string) db : unit =
     let fdescr = Unix.openfile filename [ Unix.O_WRONLY; Unix.O_CREAT; Unix.O_TRUNC ] 0o600 in
     Xapi_stdext_pervasives.Pervasiveext.finally
-      (fun () -> fd fdescr db)
+      (fun () ->
```
edwintorok: I think this change is not necessary, it was probably part of something that got fixed up and squashed? We can keep the `(fun () -> ...)` on a single line like it was before.

ocaml/xapi/xapi_fuse.ml
```diff
@@ -72,7 +72,8 @@ let light_fuse_and_dont_restart ?(fuse_length = !Xapi_globs.fuse_time) () =
                log_and_ignore_exn Xapi_stats.stop;
                log_and_ignore_exn Rrdd.backup_rrds;
                Thread.delay fuse_length;
-               Db_cache_impl.flush_and_exit (Db_connections.preferred_write_db ()) 0) ());
+               let dbconn = Xapi_stdext_monadic.Opt.unbox (Db_connections.preferred_write_db) in
+               Db_cache_impl.flush_and_exit dbconn 0) ());
```
gaborigloi: So we deliberately use `Opt.unbox` in this file to fail with an error, instead of `Opt.iter`, like in the other files?

mcintyre94: I figured since the whole thing is inside a try and `Db_connections.preferred_write_db ()` is already raising on `List.Hd` if there isn't one, this keeps the behaviour the same.

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -87,6 +87,9 @@ let create ~__context ~cluster ~host =
       | Result.Ok () ->
         Db.Cluster_host.create ~__context ~ref ~uuid ~cluster ~host ~enabled:true
           ~current_operations:[] ~allowed_operations:[] ~other_config:[];
+        (* Clustering is enabled, so fsync is enabled, sync the master DB *)
+        Helpers.call_api_functions ~__context
```
edwintorok: Could you move this before the `LocalClient.join`? I think this might be too late here as the master may have fenced when we do a `join`, and then it will have no record of this host ever being part of the cluster/pool, which is what we are trying to prevent.

ocaml/database/db_connections.ml
```diff
@@ -40,8 +40,10 @@ let choose connections = match List.filter exists connections with
     debug "Most recent db is %s (generation %Ld)" most_recent.Parse_db_conf.path gen;
     Some most_recent
 
-let preferred_write_db () =
-  List.hd (Db_conn_store.read_db_connections()) (* !!! FIX ME *)
+let preferred_write_db =
+  match Db_conn_store.read_db_connections() with
+  | [] -> None
+  | db :: _ -> Some db
```
edwintorok: You fixed the fixme, good spot.

ocaml/database/db_xml.ml
mcintyre94: It probably had a debug line :)

mcintyre94:
Sorry for the slightly nasty update there: I edited the commit that changed `preferred_write_db` (d08b492832884e6c959a70b45b731987e4080a60) because I wanted to change its commit message too, the fixups address the other changes. The other force pushed commits are just changed hashes because they follow d08b492832884e6c959a70b45b731987e4080a60.

ocaml/xapi/xapi_clustering.ml
```diff
@@ -13,6 +13,7 @@
  *)
 
 open Cluster_interface
+open Db_cache_types
```
edwintorok: Looks like there is one more `open` that should be moved.

jonludlam:
I'm not convinced this is worthwhile. Please don't merge this until we get a chance to discuss this.

ocaml/xapi/xapi_clustering.ml
mcintyre94: Looks like this one was left over + unused.

mcintyre94:
Closing this, we've discussed a different approach. I'll cherry-pick d08b492832884e6c959a70b45b731987e4080a60 from the branch independently.

PR #3466, opened 2018-02-23
[from-REQ-477] allow SMAPIv3 VDI snapshots to run on hosts other than the master

mcintyre94:
All applied cleanly on master.

### DIFF
--- a/ocaml/xapi/static_vdis.ml
+++ b/ocaml/xapi/static_vdis.ml
@@ -43,23 +43,19 @@ let permanent_vdi_attach ~__context ~vdi ~reason =
 let permanent_vdi_detach ~__context ~vdi =
   info "permanent_vdi_detach: vdi = %s; sr = %s"
     (Ref.string_of vdi) (Ref.string_of (Db.VDI.get_SR ~__context ~self:vdi));
-  Sm.call_sm_vdi_functions ~__context ~vdi
-    (fun srconf srtype sr -> Sm.vdi_detach srconf srtype sr vdi);
+  let vdi_uuid = Db.VDI.get_uuid ~__context ~self:vdi in
+  log_and_ignore_exn(fun () ->
+    ignore(Helpers.call_script !Xapi_globs.static_vdis
+           [ "detach"; vdi_uuid ]));
   ignore(Helpers.call_script !Xapi_globs.static_vdis
-           [ "del"; Db.VDI.get_uuid ~__context ~self:vdi ])
+           [ "del"; vdi_uuid ])
 
 (** Detach the VDI (by uuid) now and destroy the static configuration *)
 let permanent_vdi_detach_by_uuid ~__context ~uuid =
   info "permanent_vdi_detach: vdi-uuid = %s" uuid;
-  begin
-    try
-      (* This might fail because the VDI has been destroyed *)
-      let vdi = Db.VDI.get_by_uuid ~__context ~uuid in
-      Sm.call_sm_vdi_functions ~__context ~vdi
-        (fun srconf srtype sr -> Sm.vdi_detach srconf srtype sr vdi)
-    with e ->
-      warn "Ignoring exception calling SM vdi_detach for VDI uuid %s: %s (possibly VDI has been deleted while we were offline" uuid (ExnHelper.string_of_exn e)
-  end;
+  (* This might fail because the VDI has been destroyed *)
+  log_and_ignore_exn(fun () ->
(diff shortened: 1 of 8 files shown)


ocaml/xapi/storage_utils.ml
```diff
@@ -17,3 +17,36 @@ let string_of_vdi_type vdi_type =
...
+    that first attempts to call [local_fn call], and if that returns a [Redirect ip] exception then
+    it sends a remote RPC to the url built by [remote_url_of_ip].
+ *)
+let redirectable_rpc ~srcstr ~dststr ~remote_url_of_ip ~local_fn =
+  let rec rpc ~f call =
+    (* on first iteration this will be the [local_fn] supplied by the caller *)
+    let result = f call in
+    if result.Rpc.success then result
+    else begin
+        debug "Got failure: checking for redirect";
```
lindig: This spreads log messages that cannot be identified individually over several log lines which might no longer be next to each other in the log. Log should be self-contained.

edwintorok: Agreed, I'll fix this

edwintorok:
I've collapsed some of the mutliple debug calls to a single debug call, and repeated the rpc on some of the other debug messages. Not sure what the best approach would be here, lets keep this in mind as a use-case for the logging best practices seminar.

lindig:
Please squash the fixup commits and we will merge it.

ocaml/xapi/static_vdis.ml
```diff
@@ -43,23 +43,17 @@ let permanent_vdi_attach ~__context ~vdi ~reason =
...
   info "permanent_vdi_detach: vdi = %s; sr = %s"
     (Ref.string_of vdi) (Ref.string_of (Db.VDI.get_SR ~__context ~self:vdi));
-  Sm.call_sm_vdi_functions ~__context ~vdi
-    (fun srconf srtype sr -> Sm.vdi_detach srconf srtype sr vdi);
+  let vdi_uuid = Db.VDI.get_uuid ~__context ~self:vdi in
   ignore(Helpers.call_script !Xapi_globs.static_vdis
-           [ "del"; Db.VDI.get_uuid ~__context ~self:vdi ])
+           [ "detach"; vdi_uuid ]);
+  ignore(Helpers.call_script !Xapi_globs.static_vdis
+           [ "del"; vdi_uuid ])
```
mseri: What happens if any of these raises an exception? Is it fine to propagate it without performing `del` for example?

ocaml/xapi/static_vdis.ml
```diff
@@ -43,23 +43,17 @@ let permanent_vdi_attach ~__context ~vdi ~reason =
...
-    try
-      (* This might fail because the VDI has been destroyed *)
-      let vdi = Db.VDI.get_by_uuid ~__context ~uuid in
-      Sm.call_sm_vdi_functions ~__context ~vdi
-        (fun srconf srtype sr -> Sm.vdi_detach srconf srtype sr vdi)
-    with e ->
-      warn "Ignoring exception calling SM vdi_detach for VDI uuid %s: %s (possibly VDI has been deleted while we were offline" uuid (ExnHelper.string_of_exn e)
-  end;
+  (* This might fail because the VDI has been destroyed *)
+  ignore(Helpers.call_script !Xapi_globs.static_vdis [ "detach"; uuid ]);
```
mseri: As above, the previous version ignored the erros, this one would fail and reraise before reaching `"del"`

ocaml/xapi/storage_impl.ml
```diff
@@ -964,6 +964,7 @@ let initialise () =
 module Local_domain_socket = struct
   let path = Filename.concat "/var/lib/xcp" "storage"
 
+  (* receives external requests on Constants.sm_uri *)
```
mseri: Make it into a docstring, this can be useful to see during the autocompletion

ocaml/xapi/static_vdis.ml
edwintorok: Right, I think `log_and_ignore_exn` might be better here.

ocaml/xapi/static_vdis.ml
edwintorok: I pushed a change to use `log_and_ignore_exn`, that should preserve the original behaviour wrt to exceptions.
According to the comment the  VDI might have already been destroyed, and then detach can fail.

mseri:
I think the new version preserving the old behaviour is safer. Please, suqash the commits before merging

PR #3467, opened 2018-02-23
[From REQ-477] Quicktest_vdi_copy: Wait for VBD unplug and destroy VDIs created by this test

mcintyre94:
Quicktest_vdi_copy: clean up (destroy VDIs created by this test like the others do)
CA-277464: Quicktest_vdi_copy: wait for VBD unplug before destroying VDI


lindig:
This needs to be rebased

PR #3468, opened 2018-02-23
[From REQ-477] Cherry-pick testing changes around SMAPIv3

mcintyre94:
CP-24206: check snapshot works when VDI is activated on different hosts
CA-271525: Add a test_rpc field to the context
CA-271525: Move over all uses of 'test_mode' to use the mock rpc function instead.


mseri:
This is a much cleaner approach, even if it introduces some level of magical indirection

PR #3469, opened 2018-02-23
Tar_unix: replace Tar_unix.Archive.multicast_n_string with its implementation

mseri:
It will be removed in the next version of the library and it is a small and understandable amount of code.
Also... we were not using it for multicast so we could simplify it a bit.


mseri:
This was the upstream implementation: https://github.com/mirage/ocaml-tar/commit/c904b1753b8e863caadfb7f64a2b81a1b6f0ce62#diff-94b040c72b90cdd608bbc5698701df46L140

PR #3470, opened 2018-02-23
[From REQ-477] Cherry-pick addition of iSCSI IQN and multipathing changes

mcintyre94:
This adds the iSCSI IQN functionality and multipatching changes (they were closely intertwined, and only 2 commits were identified initially as multipathing anyway so I combined them here).

Note that the changes to `datamodel.ml` in a few commits were manually patched into `datamodel_host.ml` since the split was done since those commits were made.

### DIFF
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -1235,6 +1235,17 @@ let host_query_ha = call ~flags:[`Session]
       ~allowed_roles:_R_VM_OP
       ()
 
+  let set_iscsi_iqn = call
+    ~name:"set_iscsi_iqn"
+    ~lifecycle:[Published, rel_kolkata, ""]
+    ~doc:"Sets the initiator IQN for the host"
+    ~params:[
+      Ref _host, "host", "The host";
+      String, "value", "The value to which the IQN should be set"
+    ]
+    ~allowed_roles:_R_POOL_OP
+    ()
+
   (** Hosts *)
   let t =
     create_obj ~in_db:true ~in_product_since:rel_rio ~in_oss_since:oss_since_303 ~internal_deprecated_since:None ~persist:PersistEverything ~gen_constructor_destructor:false ~name:_host ~descr:"A physical host" ~gen_events:true
@@ -1346,6 +1357,7 @@ let host_query_ha = call ~flags:[`Session]
         apply_guest_agent_config;
         mxgpu_vf_setup;
         allocate_resources_for_vm;
+        set_iscsi_iqn;
       ]
       ~contents:
         ([ uid _host;
@@ -1400,8 +1412,7 @@ let host_query_ha = call ~flags:[`Session]
            field ~qualifier:DynamicRO ~in_product_since:rel_cream ~default_value:(Some (VSet [VInt 0L])) ~ty:(Set (Int)) "virtual_hardware_platform_versions" "The set of versions of the virtual hardware platform that the host can offer to its guests";
            field ~qualifier:DynamicRO ~default_value:(Some (VRef null_ref)) ~in_product_since:rel_ely ~ty:(Ref _vm) "control_domain" "The control domain (domain 0)";
            field ~qualifier:DynamicRO ~lifecycle:[Published, rel_ely, ""] ~ty:(Set (Ref _pool_update)) ~ignore_foreign_key:true "updates_requiring_reboot" "List of updates which require reboot";
-         field ~qualifier:DynamicRO ~lifecycle:[Published, rel_falcon, ""] ~ty:(Set (Ref _feature)) "features" "List of features available on this host";
-         field ~qualifier:StaticRO ~lifecycle:[Published, rel_kolkata, ""] ~default_value:(Some (VString "")) ~ty:String "iscsi_iqn" "The initiator IQN for the host";
-         field ~qualifier:StaticRO ~lifecycle:[Published, rel_kolkata, ""] ~default_value:(Some (VBool false)) ~ty:Bool "multipathing" "Specifies whether multipathing is enabled";
+           field ~qualifier:DynamicRO ~lifecycle:[Published, rel_falcon, ""] ~ty:(Set (Ref _feature)) "features" "List of features available on this host";
+           field ~qualifier:StaticRO ~lifecycle:[Published, rel_kolkata, ""] ~default_value:(Some (VString "")) ~ty:String "iscsi_iqn" "The initiator IQN for the host";
          ])
       ()
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -1246,6 +1246,17 @@ let host_query_ha = call ~flags:[`Session]
     ~allowed_roles:_R_POOL_OP
     ()
 
+  let set_multipathing = call
+    ~name:"set_multipathing"
+    ~lifecycle:[Published, rel_kolkata, ""]
+    ~doc:"Specifies whether multipathing is enabled"
+    ~params:[
+      Ref _host, "host", "The host";
+      Bool, "value", "Whether multipathing should be enabled"
+    ]
+    ~allowed_roles:_R_POOL_OP
+    ()
+
   (** Hosts *)
   let t =
     create_obj ~in_db:true ~in_product_since:rel_rio ~in_oss_since:oss_since_303 ~internal_deprecated_since:None ~persist:PersistEverything ~gen_constructor_destructor:false ~name:_host ~descr:"A physical host" ~gen_events:true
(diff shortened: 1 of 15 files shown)


jonludlam:
This looks fine to me.

mcintyre94:
Rebased on master now #3468 has merged :)

lindig:
This needs a rebase

mcintyre94:
Fixup: Moved test_host_helpers.ml to ocaml/tests/

edwintorok:
We should replace `rel_jura` with `rel_kolkata`, this change didn't make it into jura.

mcintyre94:
Replaced `rel_jura` with `rel_kolkata` per Edwin's comment ^

ocaml/idl/datamodel_host.ml
```diff
@@ -1400,6 +1424,8 @@ let host_query_ha = call ~flags:[`Session]
...
            field ~qualifier:DynamicRO ~default_value:(Some (VRef null_ref)) ~in_product_since:rel_ely ~ty:(Ref _vm) "control_domain" "The control domain (domain 0)";
            field ~qualifier:DynamicRO ~lifecycle:[Published, rel_ely, ""] ~ty:(Set (Ref _pool_update)) ~ignore_foreign_key:true "updates_requiring_reboot" "List of updates which require reboot";
-           field ~qualifier:DynamicRO ~lifecycle:[Published, rel_falcon, ""] ~ty:(Set (Ref _feature)) "features" "List of features available on this host"
+         field ~qualifier:DynamicRO ~lifecycle:[Published, rel_falcon, ""] ~ty:(Set (Ref _feature)) "features" "List of features available on this host";
+         field ~qualifier:StaticRO ~lifecycle:[Published, rel_kolkata, ""] ~default_value:(Some (VString "")) ~ty:String "iscsi_iqn" "The initiator IQN for the host";
+         field ~qualifier:StaticRO ~lifecycle:[Published, rel_kolkata, ""] ~default_value:(Some (VBool false)) ~ty:Bool "multipathing" "Specifies whether multipathing is enabled";
```
mseri: This will likely need a run of `ocp-indent`

ocaml/xapi/xapi_host_helpers.ml
```diff
@@ -266,3 +266,86 @@ module Host_requires_reboot = struct
...
+  let set_initiator_name iqn =
+    let hostname = Unix.gethostname () in
+    let config_file = make_initiatorname_config iqn hostname in
+    Unixext.write_string_to_file !Xapi_globs.iscsi_initiator_config_file config_file
+
+  let set_multipathing enabled =
+    let flag = !Xapi_globs.multipathing_config_file in
+    if enabled then Unixext.touch_file flag
+    else begin
+      if Sys.file_exists flag then Sys.remove flag
```
mseri: I'm way of `Sys`... we cannot catch the errors properly when using it. Can we do this using Unix and maybe handle potential exceptions with internal_error?

ocaml/xapi/xapi_host.ml
```diff
@@ -1766,3 +1768,21 @@ let mxgpu_vf_setup ~__context ~host =
...
   (* Implemented entirely in Message_forwarding *)
   ()
+
+let set_iscsi_iqn ~__context ~host ~value =
+  if value = "" then raise Api_errors.(Server_error (invalid_value, ["value"; value]));
+  (* Note, the following sequence is carefully written - see the
+     other-config watcher thread in xapi_host_helpers.ml *)
+  Db.Host.remove_from_other_config ~__context ~self:host ~key:"iscsi_iqn";
+  Db.Host.set_iscsi_iqn ~__context ~self:host ~value;
+  Db.Host.add_to_other_config ~__context ~self:host ~key:"iscsi_iqn" ~value;
```
mseri: This re-triggers an rpc call to set the iscsi_iqn, thing that we have "kind of" just done above.
Why such a complicated flow?

ocaml/xapi/xapi_host_helpers.mli
```diff
@@ -0,0 +1,134 @@
...
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+(** Common code between for dealing with Hosts.
```
mseri: Probably `between` is a leftover from an old version

ocaml/xapi/xapi_host_helpers.mli
```diff
@@ -0,0 +1,134 @@
...
+  val watch_other_configs : __context:Context.t -> float -> string -> string
+  (** [watch_other_configs ~__context timeout] returns a function that performs
+      one iteration of watching Host.other_config. If an update occurs this
+      will check whether the iscsi_iqn field in other-config is correctly
+      reflected in the field Host.iscsi_iqn, and if not it will call
+      Host.set_iscsi_iqn with the value specified in other-config. This is
+      intended to be run on the master. The returned function has type
+      [token -> token], where [token] is a string. The initial value should be
+      the empty string, and the returned value should be used for further
+      invocations. This function is exposed only for unit testing, and should
```
mseri: It is annoying that for unit testing we need to do this...

mseri:
As per current discussion, we need to update:
https://github.com/xapi-project/xen-api/blob/master/scripts/xe-set-iscsi-iqn
and/or the `firstboot` script calling it.

ocaml/idl/datamodel_host.ml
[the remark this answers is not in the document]
edwintorok: as discussed today new fields are safe.

ocaml/xapi/xapi_host_helpers.mli
edwintorok: There is a ppx to do inline testing, but that would require using yet another ppx...

mseri: Yes, I know, I wasn't suggesting to do anything. It's unfortunate, I only wish there was some compiler level support for tests behind mli

ocaml/xapi/xapi_host_helpers.ml
edwintorok: Replaced with Unixext.unlink_safe.

ocaml/xapi/xapi_host_helpers.ml
```diff
@@ -266,3 +266,86 @@ module Host_requires_reboot = struct
...
+      if String.length hostname > 30
+      then String.sub hostname 0 30
+      else hostname
+    in
+    Printf.sprintf
+      "InitiatorName=%s\nInitiatorAlias=%s\n"
+      iqn hostname_chopped
+
+  let set_initiator_name iqn =
+    let hostname = Unix.gethostname () in
```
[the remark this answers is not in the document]
edwintorok: We'd let this one to be treated as an internal error.

ocaml/xapi/xapi_host_helpers.ml
mseri: Thanks!

ocaml/xapi/xapi_host_helpers.ml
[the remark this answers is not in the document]
mseri: Sure :+1:

ocaml/xapi/xapi_host.ml
edwintorok: I've added a comment attempting to explain my understanding of it:
+  (* we need to first set the iscsi_iqn field and then the other-config field:
+   * setting the other-config field triggers and update on the iscsi_iqn
+   * field if they are different
+   * *)

edwintorok: I think the reason for the complicated behaviour is that we want to keep both fields in sync: when you update iscsi_iqn field we update other-config to match, and when you update other-config we update iscsi_iqn to match. We have to avoid an infinite loop though.

mseri: Yes, I think it's clearer. Thanks

ocaml/tests/test_host_helpers.ml
```diff
@@ -0,0 +1,126 @@
...
+  Db.Host.remove_from_other_config ~__context ~self:host1 ~key:"multipathing";
+  let token = watcher token in
+  assert_equal !calls [];
+  Db.Host.set_multipathing ~__context ~self:host1 ~value:false;
+  let token = watcher token in
+  assert_equal !calls [];
+  Db.Host.add_to_other_config ~__context ~self:host1 ~key:"multipathing" ~value:"false";
+  let _token = watcher token in
+  assert_equal !calls []
+
```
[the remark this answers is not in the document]
edwintorok: I don't think that setting multipathing should fail (unless the DB itself fails)

edwintorok:
Thanks for the review, lets wait until Travis build finishes.

PR #3472, opened 2018-02-23
Rearrange main executables and tests

mseri:
This moves the executables and tests in different scopes, reorganising and clarifying the dependencies.
This is a first step, further improvements will be to further refine the dependencies among those folders and replace `stdext` with the appropriate `xapi-stdext-` submodules.


ocaml/xapi/jbuild
```diff
@@ -67,11 +67,11 @@ let () = Printf.ksprintf Jbuild_plugin.V1.send {|
...
- ((names (xapi_main suite suite_alcotest quicktest))
-  (public_names (xapi - - quicktestbin))
-  (package xapi)
+(library
+ ((name xapi_internal)
+  (wrapped false)
   (flags (:standard -bin-annot %s -warn-error +a-3-4-6-9-27-28-29-52))
+  (modules (:standard \ xapi_main quicktest))
   (libraries (
    alcotest
```
gaborigloi: Can we also remove this?

gaborigloi:
Will it be easier to merge https://github.com/xapi-project/xen-api/pull/3468 first and then update this PR to move the new / modified files to their correct place?
Or I guess we should merge this first if reviewing those PRs would take long , and then  https://github.com/xapi-project/xen-api/pull/3468 will have to be updated.

mseri:
Yes, it will not be a problem to rebase after your changes in #3648. I think that should go first, once it passes travis tests

mseri:
Rebased on top of #3468.

`ls ocaml/xapi` now fits in a full-screen terminal :tada: :fearful:

ocaml/tests/jbuild
```diff
@@ -0,0 +1,58 @@
...
+)
+
+(alias
+((name runtest)
+ (package xapi)
+(deps (suite_alcotest.exe (files_recursively_in test_data)))
+(action (run ${<}))
+)
+)
+|} (flags rewriters) coverage_rewriter
```
robhoes: Do we still need to do this?

gaborigloi: Only for coverage, as far as I know.

gaborigloi: So I think we could remove `flags`.

mseri: It is still needed until we port `rpc` to `jbuilder`. A port is ready but there are some issues with `camlp4` and the way it interacts with ppx in `xapi-idl`, so I'd rather wait until the port is complete

mseri:
I've rebased to avoid the conflicts. I am going to merge this if Travis is happy

mseri:
The xenserver-build-env failure is due to idl changes not yet propagated. The Travis build with xs-opam has successfully passed. Merging

PR #3473, opened 2018-02-23
[From REQ-477] Cherry-pick of shutdown changes

mcintyre94:
** Do not merge before #3470, that adds an mli file that I need to make a change to here, will need to fix after rebasing after that merges**. This builds because that mli file doesn't exist on master, once it's there I'll need to add the change to it. It's added over a few commits in #3470 though, more straightforward to rebase later.

Again, adding this one despite the rebase and small fix needed because it's significant changes and would appreciate feedback! Needed a fair bit of manual tweaking around the pbd.ml (untangling clustering related changes from the bug fixes) and the datamodel due to the split.

- CA-274585: allow unplugging statefile VDI 
- CA-274585: unplug all local PBDs on shutdown/reboot 
- CP-24693: Tasks.with_tasks_destroy: add a function that waits for tasks with a timeout
- CA-277346: do not get stuck detaching metadata VDI
- CA-277346: stop (DB) requests when shutting down the master 
- Convert host evacuation script into ocaml
- Remove scripts/examples/python/shutdown.py 
- Don't do prechecks in Host.prepare_for_poweroff

### DIFF
--- a/ocaml/xapi-client/jbuild
+++ b/ocaml/xapi-client/jbuild
@@ -20,6 +20,8 @@ let () = Printf.ksprintf Jbuild_plugin.V1.send {|
  ((name xapi_client)
   (public_name xapi-client)
   (libraries (
+   mtime
+   mtime.clock.os
    xapi-types
    xapi-stdext-date
   ))
--- a/ocaml/xapi-client/tasks.ml
+++ b/ocaml/xapi-client/tasks.ml
@@ -14,35 +14,69 @@
 
 open Client
 
+module D = Debug.Make(struct let name = "tasks" end)
+
 module TaskSet = Set.Make(struct type t = API.ref_task let compare = compare end)
 
 (* Return once none of the tasks have a `pending status. *)
-let wait_for_all ~rpc ~session_id ~tasks =
+let wait_for_all_inner ~rpc ~session_id ?all_timeout ~tasks =
   let classes = List.map
       (fun task -> Printf.sprintf "task/%s" (Ref.string_of task))
       tasks
   in
+  let timeout_span = match all_timeout with
+    | Some t -> Some (t *. Mtime.s_to_ns |> Int64.of_float |> Mtime.Span.of_uint64_ns)
+    | None -> None in
+  let timer = Mtime_clock.counter () in
   let timeout = 5.0 in
   let rec wait ~token ~task_set =
-    if TaskSet.is_empty task_set then ()
-    else begin
-      let open Event_types in
-      let event_from_rpc = Client.Event.from ~rpc ~session_id ~classes ~token ~timeout in
-      let event_from = Event_types.event_from_of_rpc event_from_rpc in
-      let records = List.map Event_helper.record_of_event event_from.events in
-      (* If any records indicate that a task is no longer pending, remove that task from the set. *)
-      let pending_task_set = List.fold_left (fun task_set' record ->
-          match record with
-          | Event_helper.Task (t, Some t_rec) ->
-            if (TaskSet.mem t task_set') && (t_rec.API.task_status <> `pending) then
-              TaskSet.remove t task_set'
-            else
-              task_set'
-          | _ -> task_set') task_set records in
-      wait ~token:(event_from.Event_types.token) ~task_set:pending_task_set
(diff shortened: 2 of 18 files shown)


mcintyre94:
Removed 'Remove scripts/examples/python/shutdown.py' - that's self-contained and will require a synchronised xenserver-specs PR, we'll make it separately later.

lindig:
Is this ready to go? Do we need to clean up the history a bit?

edwintorok:
Tidied up the commits a bit, and we've reviewed and tested this in the feature branch. Might be useful if someone else takes a quick look as this affects the shutdown code of XAPI and is not feature flagged.

SM has implemented various workarounds (like introducing another service that gets run after xapi is shutdown and unmounts the SR) for unmounts getting stuck on shutdown.
With clustered filesystems XAPI really needs to ensure that it unplugs the PBDs before stopping clustering though, otherwise you either delay other hosts until the fencing timeout expires everytime you shutdown a host, and other hosts will treat the shut down host as crashed for the purposes of quorum calculation which is not what we wanted. 
If we cleanly unmount and leave the cluster on shutdown that is a better user experience, unifies the shutdown behaviour (regardless if you do `xe host-shutdown` or `/sbin/shutdown` you execute the same code), and it removes the need for some of the workaround in SM, hence this PR is not feature flagged.

edwintorok:
Just noticed that its missing this commit: 4f497c8880db62c18e6ae956c9f9a362538e6871

ocaml/xapi-client/jbuild
```diff
@@ -20,6 +20,8 @@ let () = Printf.ksprintf Jbuild_plugin.V1.send {|
  ((name xapi_client)
   (public_name xapi-client)
   (libraries (
+   mtime
+   mtime.clock.os
```
mseri: Do you need to update the opam file as well?

ocaml/xapi-client/tasks.ml
```diff
@@ -14,18 +14,30 @@
...
-    if TaskSet.is_empty task_set then ()
-    else begin
+    if TaskSet.is_empty task_set then true
+    else match timeout_span with
+      | Some span when Mtime.Span.compare (Mtime_clock.count timer) span > 0 ->
+            let tasks = TaskSet.elements task_set in
+            let tasks_str = tasks |> List.map Ref.really_pretty_and_small |> String.concat "," in
+            D.info "Waiting for tasks timed out on %s" tasks_str;
+            false
+      | _ ->
```
mseri: I think we should run `ocp-indent` on this file after these changes.

mseri:
`ensure_no_vms` contains a few `XXX` comments. Do we need to do anything about that?

edwintorok:
I think you can still unplug the PBDs when the machines are suspended, but I'll double check that.

mseri:
Please double check. Aside that, the current fixup version looks good to me.

edwintorok:
Might be useful to cancel the tasks as the python script did, e.g. if you were attempting to migrate a VM you probably can't hard shutdown it, will try to fix that.

mseri:
Thanks

edwintorok:
The `shutdown.py` script used `current_operations` field, did the same in my fixup. Unfortunately the type here is a string to enum map instead of task ref to enum map and we can't change that in the datamodel without breaking compatibility with existing API users.
Might be useful if there was a way to use the correct type internally at least, without breaking the external API.

edwintorok:
Thanks, squashed the fixups.

mseri:
Btw, I saw your discussion on slack and I was looking at the docs. What you do is not different from how we use the current operations in other parts of xapi. It is unfortunate but we cannot really do much about it with a reasonable effort.

mseri:
Feel free to merge when appropriate

lindig:
This needs to be merged after #3470.

mseri:
I am half way through that PR

edwintorok:
I've pushed a fixup for the indentation, I think that was the only outstanding action on this PR.

mseri:
Yes, I think this is fine. Once #3470 is ready and merged, we can merge this as well. Feel free to squash the fixup commits

PR #3474, opened 2018-02-24
CP-26022 SR-IOV pool join/leave

krizex:
1. Add network sriov precheck for pool join
2. Sync network sriov from master to slave after join
3. Gc sriovs when gc PIFs
4. Best effort bring up sriov logical PIFs when xapi start

---
This change is [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3474)

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -948,6 +948,8 @@ let _ =
     ~doc:"The host joining the pool must not have any bonds." ();
   error Api_errors.pool_joining_host_has_tunnels []
     ~doc:"The host joining the pool must not have any tunnels." ();
+  error Api_errors.pool_joining_host_has_network_sriovs []
+    ~doc:"The host joining the pool must not have any network SR-IOVs." ();
   error Api_errors.pool_hosts_not_compatible []
     ~doc:"The hosts in this pool are not compatible." ();
   error Api_errors.pool_hosts_not_homogeneous [ "reason" ]
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -5008,6 +5008,18 @@ let host_sync_tunnels = call ~flags:[`Session]
     ~allowed_roles:_R_POOL_OP
     ()
 
+let host_sync_network_sriovs = call ~flags:[`Session]
+    ~name:"sync_network_sriovs"
+    ~lifecycle:[Published, rel_kolkata, ""]
+    ~doc:"Synchronise network sriovs on given host with the master's network sriovs"
+    ~params:[
+      Ref _host, "host", "The host";
+    ]
+    ~hide_from_docs:true
+    ~pool_internal:true
+    ~allowed_roles:_R_POOL_OP
+    ()
+
 let host_sync_pif_currently_attached = call ~flags:[`Session]
     ~name:"sync_pif_currently_attached"
     ~lifecycle:[]
@@ -5162,6 +5174,7 @@ let host =
                 host_sm_dp_destroy;
                 host_sync_vlans;
                 host_sync_tunnels;
+                host_sync_network_sriovs;
                 host_sync_pif_currently_attached;
                 host_migrate_receive;
                 host_declare_dead;
--- a/ocaml/xapi-consts/api_errors.ml
+++ b/ocaml/xapi-consts/api_errors.ml
@@ -343,6 +343,7 @@ let pool_joining_host_management_vlan_does_not_match = "POOL_JOINING_HOST_MANAGE
 let pool_joining_host_has_non_management_vlans = "POOL_JOINING_HOST_HAS_NON_MANAGEMENT_VLANS"
 let pool_joining_host_has_bonds = "POOL_JOINING_HOST_HAS_BONDS"
 let pool_joining_host_has_tunnels = "POOL_JOINING_HOST_HAS_TUNNELS"
+let pool_joining_host_has_network_sriovs = "POOL_JOINING_HOST_HAS_NETWORK_SRIOVS"
 
 (*workload balancing*)
 let wlb_not_initialized = "WLB_NOT_INITIALIZED"
--- a/ocaml/xapi/db_gc_util.ml
+++ b/ocaml/xapi/db_gc_util.ml
@@ -62,11 +62,12 @@ let gc_PIFs ~__context =
   gc_connector ~__context Db.PIF.get_all Db.PIF.get_record (fun x->valid_ref __context x.pIF_host) (fun x->valid_ref __context x.pIF_network)
     (fun ~__context ~self ->
        (* We need to destroy the PIF, it's metrics and any VLAN/bond records that this PIF was a master of. *)
-       (* bonds/tunnels_to_gc is actually a list which is either empty (not part of a bond/tunnel)
+       (* bonds/tunnels/sriovs_to_gc is actually a list which is either empty (not part of a bond/tunnel/sriov)
         * or containing exactly one reference.. *)
        let bonds_to_gc = Db.PIF.get_bond_master_of ~__context ~self in
        let vlan_to_gc = Db.PIF.get_VLAN_master_of ~__context ~self in
        let tunnels_to_gc = Db.PIF.get_tunnel_access_PIF_of ~__context ~self in
+       let sriovs_to_gc = Db.PIF.get_sriov_logical_PIF_of ~__context ~self in
        (* Only destroy PIF_metrics of physical or bond PIFs *)
        if vlan_to_gc = Ref.null && tunnels_to_gc = [] then begin
          let metrics = Db.PIF.get_metrics ~__context ~self in
@@ -74,6 +75,7 @@ let gc_PIFs ~__context =
        end;
        (try Db.VLAN.destroy ~__context ~self:vlan_to_gc with _ -> ());
        List.iter (fun tunnel -> (try Db.Tunnel.destroy ~__context ~self:tunnel with _ -> ())) tunnels_to_gc;
+       List.iter (fun sriov -> (try Db.Network_sriov.destroy ~__context ~self:sriov with _ -> ())) sriovs_to_gc;
        List.iter (fun bond -> (try Db.Bond.destroy ~__context ~self:bond with _ -> ())) bonds_to_gc;
        Db.PIF.destroy ~__context ~self) 
       
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -2658,6 +2658,10 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
       info "Host.sync_tunnels: host = '%s'" (host_uuid ~__context host);
       Local.Host.sync_tunnels ~__context ~host
 
+    let sync_network_sriovs ~__context ~host =
+      info "Host.sync_network_sriovs: host = '%s'" (host_uuid ~__context host);
+      Local.Host.sync_network_sriovs ~__context ~host
+
     let sync_pif_currently_attached ~__context ~host ~bridges =
       info "Host.sync_pif_currently_attached: host = '%s'" (host_uuid ~__context host);
       Local.Host.sync_pif_currently_attached ~__context ~host ~bridges
--- a/ocaml/xapi/sync_networking.ml
+++ b/ocaml/xapi/sync_networking.ml
@@ -161,3 +161,8 @@ let copy_tunnels_from_master ~__context () =
   let host = !Xapi_globs.localhost_ref in
   Helpers.call_api_functions ~__context (fun rpc session_id -> Client.Host.sync_tunnels ~rpc ~session_id ~host)
 
+(** Copy network-sriovs from master *)
+let copy_network_sriovs_from_master ~__context () =
+  debug "Resynchronising network-sriovs";
+  let host = !Xapi_globs.localhost_ref in
+  Helpers.call_api_functions ~__context (fun rpc session_id -> Client.Host.sync_network_sriovs ~rpc ~session_id ~host)
--- a/ocaml/xapi/sync_networking.mli
+++ b/ocaml/xapi/sync_networking.mli
@@ -19,3 +19,4 @@ val fix_bonds : __context:Context.t -> unit -> unit
 val copy_bonds_from_master : __context:Context.t -> unit -> unit
 val copy_vlans_from_master : __context:Context.t -> unit -> unit
 val copy_tunnels_from_master : __context:Context.t -> unit -> unit
+val copy_network_sriovs_from_master : __context:Context.t -> unit -> unit
--- a/ocaml/xapi/xapi.ml
+++ b/ocaml/xapi/xapi.ml
@@ -906,7 +906,7 @@ let server_init() =
           "creating networks", [ Startup.OnlyMaster ], Create_networks.create_networks_localhost;
           (* CA-22417: bring up all non-bond slaves so that the SM backends can use storage NIC IP addresses (if the routing
              	 table happens to be right) *)
-          "Best-effort bring up of physical NICs", [ Startup.NoExnRaising ], Xapi_pif.start_of_day_best_effort_bring_up;
+          "Best-effort bring up of physical and sriov NICs", [ Startup.NoExnRaising ], Xapi_pif.start_of_day_best_effort_bring_up;
           "updating the vswitch controller", [], (fun () -> Helpers.update_vswitch_controller ~__context ~host:(Helpers.get_localhost ~__context));
           "initialising storage", [ Startup.NoExnRaising ],
           (fun () -> Helpers.call_api_functions ~__context Create_storage.create_storage_localhost);
--- a/ocaml/xapi/xapi.ml
+++ b/ocaml/xapi/xapi.ml
@@ -916,6 +916,7 @@ let server_init() =
           "watching networks for NBD-related changes", [Startup.OnThread], Network_event_loop.watch_networks_for_nbd_changes;
           (* CA-175353: moving VIFs between networks requires VMs to be resynced *)
           "Synchronising bonds on slave with master", [Startup.OnlySlave; Startup.NoExnRaising], Sync_networking.copy_bonds_from_master ~__context;
+          "Synchronising network sriovs on slave with master", [Startup.OnlySlave; Startup.NoExnRaising], Sync_networking.copy_network_sriovs_from_master ~__context;
           "Synchronising VLANs on slave with master", [Startup.OnlySlave; Startup.NoExnRaising], Sync_networking.copy_vlans_from_master ~__context;
           "Synchronising tunnels on slave with master", [Startup.OnlySlave; Startup.NoExnRaising], Sync_networking.copy_tunnels_from_master ~__context;
 
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -1620,6 +1620,49 @@ let sync_tunnels ~__context ~host =
   (* for each of the master's pifs, create a corresponding one on this host if necessary *)
   List.iter maybe_create_tunnel_for_me master_tunnel_pifs
 
+let sync_network_sriovs ~__context ~host =
+  let master = !Xapi_globs.localhost_ref in
+  let master_sriov_pifs = Db.PIF.get_records_where ~__context ~expr:(And (
+      Eq (Field "host", Literal (Ref.string_of master)),
+      Not (Eq (Field "sriov_logical_PIF_of", Literal "()"))
+    )) in
+  let slave_sriov_pifs = Db.PIF.get_records_where ~__context ~expr:(And (
+      Eq (Field "host", Literal (Ref.string_of host)),
+      Not (Eq (Field "sriov_logical_PIF_of", Literal "()"))
+    )) in
+  let slave_physical_pifs = Db.PIF.get_records_where ~__context ~expr:(And (
+      Eq (Field "host", Literal (Ref.string_of host)),
+      Eq (Field "physical", Literal "true")
+    )) in
+  let maybe_create_sriov (master_pif_ref, master_pif_rec) =
+    let sriov_network = master_pif_rec.API.pIF_network in
+    let existing_pif = List.filter (fun (_, slave_pif_rec) ->
+        slave_pif_rec.API.pIF_network = sriov_network
+      ) slave_sriov_pifs in
+    if existing_pif = [] then begin
+      let device = master_pif_rec.API.pIF_device in
+      let pifs = List.filter (fun (_, pif_rec) -> pif_rec.API.pIF_device = device) slave_physical_pifs in
+      match pifs with
+      | [] ->
+        info "Cannot sync network sriov because cannot find PIF whose device name is %s" device
+      | (pif_ref, pif_rec) :: _ ->
+        begin
+          try
+            Xapi_network_sriov.create ~__context ~pif:pif_ref ~network:sriov_network |> ignore
+          with
+          | Api_errors.Server_error (err, _) when err = Api_errors.network_has_incompatible_sriov_pifs ->
+            warn "Cannot sync network sriov on slave because PCI device of %s is different from the PIF of master in the same position" pif_rec.API.pIF_uuid
+          | Api_errors.Server_error (err, _) when err = Api_errors.network_sriov_already_enabled ->
+            warn "Cannot sync network sriov on slave because PIF %s on slave has enabled sriov in another network" pif_rec.API.pIF_uuid
+          | Api_errors.Server_error (err, _) when err = Api_errors.pif_is_not_sriov_capable ->
+            warn "Cannot sync network sriov on slave because PIF %s on slave is not sriov capable" pif_rec.API.pIF_uuid
+          | exn ->
+            error "Error occurs when syncing network sriov for PIF %s: %s" pif_rec.API.pIF_uuid (Printexc.to_string exn)
+        end
+    end
+  in
+  List.iter maybe_create_sriov master_sriov_pifs
+
 let sync_pif_currently_attached ~__context ~host ~bridges =
   (* Produce internal lookup tables *)
   let networks = Db.Network.get_all_records ~__context in
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -1519,30 +1519,30 @@ let sync_vlans ~__context ~host =
       Not (Eq (Field "VLAN_master_of", Literal (Ref.string_of Ref.null)))
     )) in
 
-  let get_network_of_pif_underneath_vlan vlan_pif =
-    let vlan = Db.PIF.get_VLAN_master_of ~__context ~self:vlan_pif in
-    let pif_underneath_vlan = Db.VLAN.get_tagged_PIF ~__context ~self:vlan in
-    Db.PIF.get_network ~__context ~self:pif_underneath_vlan
+  let get_network_of_pif_underneath_vlan vlan_pif_rec =
+    match Xapi_pif_helpers.get_pif_topo ~__context ~pif_rec:vlan_pif_rec with
+    | VLAN_untagged vlan :: _ ->
+      let pif_underneath_vlan = Db.VLAN.get_tagged_PIF ~__context ~self:vlan in
+      Db.PIF.get_network ~__context ~self:pif_underneath_vlan
+    | _ -> raise Api_errors.(Server_error(internal_error, [Printf.sprintf "Cannot find vlan from a vlan master PIF:%s" vlan_pif_rec.API.pIF_uuid]))
   in
-
   let maybe_create_vlan (master_pif_ref, master_pif_rec) =
     (* Check to see if the slave has any existing pif(s) that for the specified device, network, vlan... *)
+    (* On the master, we find the pif, p, that underlies the VLAN
+     * (e.g. "eth0" underlies "eth0.25") and then find the network that p's on: *)
+    let network_of_pif_underneath_vlan_on_master = get_network_of_pif_underneath_vlan master_pif_rec in
     let existing_pif = List.filter (fun (slave_pif_ref, slave_pif_record) ->
         (* Is slave VLAN PIF that we're considering (slave_pif_ref) the one that corresponds
            			 * to the master_pif we're considering (master_pif_ref)? *)
         true
         && slave_pif_record.API.pIF_network = master_pif_rec.API.pIF_network
         && slave_pif_record.API.pIF_VLAN = master_pif_rec.API.pIF_VLAN
-        && ((get_network_of_pif_underneath_vlan slave_pif_ref) =
-            (get_network_of_pif_underneath_vlan master_pif_ref))
+        && (get_network_of_pif_underneath_vlan slave_pif_record = network_of_pif_underneath_vlan_on_master)
       ) slave_vlan_pifs in
     (* if I don't have any such pif(s) then make one: *)
-    if List.length existing_pif = 0
+    if existing_pif = []
     then
       begin
-        (* On the master, we find the pif, p, that underlies the VLAN
-           				 * (e.g. "eth0" underlies "eth0.25") and then find the network that p's on: *)
-        let network_of_pif_underneath_vlan_on_master = get_network_of_pif_underneath_vlan master_pif_ref in
         let pifs = Db.PIF.get_records_where ~__context ~expr:(And (
             Eq (Field "host", Literal (Ref.string_of host)),
             Eq (Field "network", Literal (Ref.string_of network_of_pif_underneath_vlan_on_master))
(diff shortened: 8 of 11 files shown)


robhoes:
Four clear items in the commit message: there could be four nice and small commits as well.

---

Reviewed 12 of 12 files at r1.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[ocaml/xapi/xapi.ml, line 910 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6RF6yNZ4d2366pOyri:-L6RF6yNZ4d2366pOyrj:bi2onig) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi.ml#L910)):*
> ```OCaml
>              	 table happens to be right) *)
>           "Best-effort bring up of physical NICs", [ Startup.NoExnRaising ], Xapi_pif.start_of_day_best_effort_bring_up;
>           "Best-effort bring up of sriov PIFs", [ Startup.NoExnRaising ], Xapi_pif.best_effort_bring_up_sriov_pifs;
> ```

I think that this shouldn't be a separate item, but part of the one above.

---

*[ocaml/xapi/xapi_host.ml, line 1635 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6RniKaPe0KPpZ4e8hQ:-L6RniKaPe0KPpZ4e8hR:bdzn34c) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_host.ml#L1635)):*
> ```OCaml
>   let maybe_create_sriov (master_pif_ref, master_pif_rec) =
>     let sriov_network = master_pif_rec.API.pIF_network in
>     let exisiting_pif = List.filter (fun (_, slave_pif_rec) ->
> ```

`existing_pif`

---

*[ocaml/xapi/xapi_host.ml, line 1638 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6Ro7Q3WzkbNQvW5eLk:-L6Ro7Q3WzkbNQvW5eLl:be9hcm8) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_host.ml#L1638)):*
> ```OCaml
>         slave_pif_rec.API.pIF_network = sriov_network
>       ) slave_sriov_pifs in
>     if List.length exisiting_pif = 0 then begin
> ```

`if existing_pif = []` is slightly better.

---

*[ocaml/xapi/xapi_host.ml, line 1642 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6Rp_0YSXVNDIRas9nB:-L6Rp_0YSXVNDIRas9nC:b-ta3mev) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_host.ml#L1642)):*
> ```OCaml
>       let pifs = Db.PIF.get_records_where ~__context ~expr:(And (
>           Eq (Field "host", Literal (Ref.string_of host)),
>           Eq (Field "device", Literal (device))
> ```

This is not reliable, because there may be multiple PIFs with the same `device` on the same `host`. For example, VLAN and SR_IOV PIFs always share the device name of the physical PIF underneath. Hence the use of helpers such as `get_network_of_transport_pif` and `get_network_of_pif_underneath_vlan` in the sync functions above. You've got `get_pif_topo` now.

---

*[ocaml/xapi/xapi_host.ml, line 1651 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6RrESWkkZ8kFN2RmBv:-L6RrESWkkZ8kFN2RmBw:b-qr49qs) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_host.ml#L1651)):*
> ```OCaml
>           try
>             Xapi_network_sriov.create ~__context ~pif:pif_ref ~network:sriov_network |> ignore
>           with
> ```

We should just let the exceptions escape. They will then be logged by the startup sequence on the slave (the startup item has `Startup.NoExnRaising`).

---

*[ocaml/xapi/xapi_pif.ml, line 889 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6REgt5j4mDXBSwaKDK:-L6REgt5j4mDXBSwaKDL:b-5ztyfp) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_pif.ml#L889)):*
> ```OCaml
>   end
> 
> let best_effort_bring_up_sriov_pifs () =
> ```

I don't think that this needs to be a separate function. The `start_of_day_best_effort_bring_up` function above should also do SR-IOV PIFs. So `calculate_pifs_required_at_start_of_day` needs to include those PIFs.

---

*[ocaml/xapi/xapi_pif.mli, line 267 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6RFH8q7dmigk7MyGhQ:-L6RFH8q7dmigk7MyGhR:b-rxnr2h) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_pif.mli#L267)):*
> ```OCaml
> 
> (** Bring up (plug) sriov logical PIFs when the host starts up. *)
> val best_effort_bring_up_sriov_pifs : unit -> unit
> ```

As said above, I don't think that we need this.

---

*[ocaml/xapi/xapi_pool.ml, line 267 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6RGXWo61WSvRNLo60z:-L6RGXWo61WSvRNLo61-:b-liedti) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_pool.ml#L267)):*
> ```OCaml
>   let assert_no_network_sriovs_on_me () =
>     if Db.Network_sriov.get_all ~__context <> [] then begin
>       error "The current host has network-sriovs: it cannot join a new pool";
> ```

"has network SR-IOV enabled"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-:-L6R6eqVDfQjLrAcB5Hx:b-uwbz5v)*

mseri:
I agree with Rob. Is the CP correct? The ticket appears as done

---

Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[ocaml/idl/datamodel.ml, line 5001 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6S5fvB0zIjLb1cftxV:-L6S5fvB0zIjLb1cftxW:bu97e8y) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/idl/datamodel.ml#L5001)):*
> ```OCaml
> let host_sync_network_sriovs = call ~flags:[`Session]
>     ~name:"sync_network_sriovs"
>     ~lifecycle:[]
> ```

Shouldn't we populate the lifacycle even if this is hidden from the docs?

---

*[ocaml/xapi/db_gc_util.ml, line 78 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6S5vLghibqrbUPV6L6:-L6S5vLghibqrbUPV6L7:bbhrk5x) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/db_gc_util.ml#L78)):*
> ```OCaml
>        (try Db.VLAN.destroy ~__context ~self:vlan_to_gc with _ -> ());
>        List.iter (fun tunnel -> (try Db.Tunnel.destroy ~__context ~self:tunnel with _ -> ())) tunnels_to_gc;
>        List.iter (fun sriov -> (try Db.Network_sriov.destroy ~__context ~self:sriov with _ -> ())) sriovs_to_gc;
> ```

Is the destroy order here important?

---

*[ocaml/xapi/xapi_pif.ml, line 889 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6REgt5j4mDXBSwaKDK:-L6S75cYWFZw3vuhDtHf:b-rjmeg9) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_pif.ml#L889)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

I don't think that this needs to be a separate function. The `start_of_day_best_effort_bring_up` function above should also do SR-IOV PIFs. So `calculate_pifs_required_at_start_of_day` needs to include those PIFs.
</blockquote></details>

Do we need to be careful about the ordering, make sure that we bring up the sriovs after the physical pifs are up, or it does not matter? 

---

*[ocaml/xapi/xapi_pool.ml, line 267 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6RGXWo61WSvRNLo60z:-L6S6NIwRzHtruEUsoJc:birj98k) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_pool.ml#L267)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

"has network SR-IOV enabled"
</blockquote></details>

Should we add the host to the error metadata?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-:-L6S6BmDuGgrYUtbws81:b-1ckyfb)*

robhoes:
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[ocaml/idl/datamodel.ml, line 5001 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6S5fvB0zIjLb1cftxV:-L6SQ2m0Lfeb5sW_MBY0:bndi4vh) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/idl/datamodel.ml#L5001)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

Shouldn't we populate the lifacycle even if this is hidden from the docs?
</blockquote></details>

I'd be good practice as a reminder for developers, but otherwise the information won't be used anywhere.

---

*[ocaml/xapi/db_gc_util.ml, line 78 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6S5vLghibqrbUPV6L6:-L6SPRIhv_nZ_n3WrRtR:bze869i) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/db_gc_util.ml#L78)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

Is the destroy order here important?
</blockquote></details>

No, just one object will actually be deleted: only one of the `*_to_gc` lists will have at most one thing in it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3474)*

krizex:
Yes, I will split them into 4 commits. 
The CP ticket is used when developing and target to merge to staging branch, now I will update with a new CP ticket.

---

Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[ocaml/idl/datamodel.ml, line 5001 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6S5fvB0zIjLb1cftxV:-L6UXcDLNJ4aSeP_JC8w:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/idl/datamodel.ml#L5001)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

I'd be good practice as a reminder for developers, but otherwise the information won't be used anywhere.
</blockquote></details>

Done.

---

*[ocaml/xapi/xapi.ml, line 910 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6RF6yNZ4d2366pOyri:-L6U_EHvfbBgfT2aZSY5:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi.ml#L910)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

I think that this shouldn't be a separate item, but part of the one above.
</blockquote></details>

Done.

---

*[ocaml/xapi/xapi_host.ml, line 1635 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6RniKaPe0KPpZ4e8hQ:-L6U__vYhzH4Js2H_T-i:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_host.ml#L1635)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

`existing_pif`
</blockquote></details>

Done.

---

*[ocaml/xapi/xapi_host.ml, line 1638 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6Ro7Q3WzkbNQvW5eLk:-L6U_ePTKtOfeoBMt96v:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_host.ml#L1638)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

`if existing_pif = []` is slightly better.
</blockquote></details>

Done.

---

*[ocaml/xapi/xapi_host.ml, line 1642 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6Rp_0YSXVNDIRas9nB:-L6UcPwlZ26IsEncUVob:b-v3h4zu) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_host.ml#L1642)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

This is not reliable, because there may be multiple PIFs with the same `device` on the same `host`. For example, VLAN and SR_IOV PIFs always share the device name of the physical PIF underneath. Hence the use of helpers such as `get_network_of_transport_pif` and `get_network_of_pif_underneath_vlan` in the sync functions above. You've got `get_pif_topo` now.
</blockquote></details>

Yes, so I have to add a new restriction `PIF.physical = true`.
Wrt the two functions you mentioned, I will update them with a separate commit in this PR.

---

*[ocaml/xapi/xapi_host.ml, line 1651 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6RrESWkkZ8kFN2RmBv:-L6UvMt0xpid65sDilKG:b-duw95r) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_host.ml#L1651)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

We should just let the exceptions escape. They will then be logged by the startup sequence on the slave (the startup item has `Startup.NoExnRaising`).
</blockquote></details>

No. If we do this, it will stop the iteration once one sriov creation failed.

---

*[ocaml/xapi/xapi_pif.ml, line 889 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6REgt5j4mDXBSwaKDK:-L6UvaNCchSKODPIPwJ1:bb77ufg) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_pif.ml#L889)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

Do we need to be careful about the ordering, make sure that we bring up the sriovs after the physical pifs are up, or it does not matter? 
</blockquote></details>

Yes.

---

*[ocaml/xapi/xapi_pif.mli, line 267 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6RFH8q7dmigk7MyGhQ:-L6UvbvpWyYei_Vneyg0:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_pif.mli#L267)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

As said above, I don't think that we need this.
</blockquote></details>

Done.

---

*[ocaml/xapi/xapi_pool.ml, line 267 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6RGXWo61WSvRNLo60z:-L6Uw-Np81kNCL0SLAJN:bhzuiuj) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_pool.ml#L267)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

Should we add the host to the error metadata?
</blockquote></details>

Error message updated.
We do not need add host to the error message because this message only printed in the host which intend to join a pool.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-:-L6UX9D75n-xrSwz-oBX:b-19rd1a)*

mseri:
Review status: 0 of 11 files reviewed at latest revision, 13 unresolved discussions.

---

*[ocaml/idl/datamodel.ml, line 4999 at r6](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6VrIDhqBbCwfWYW1xk:-L6VrIDijfqDJuhX-Qwu:b4zegkm) ([raw file](https://github.com/xapi-project/xen-api/blob/d9daadc545a0f80e67bfc796ff300c680e965625/ocaml/idl/datamodel.ml#L4999)):*
> ```OCaml
> let host_sync_tunnels = call ~flags:[`Session]
>     ~name:"sync_tunnels"
>     ~lifecycle:[Published, rel_kolkata, ""]
> ```

You have updated the `host_sunc_tunnels` function instead of `sync_network_sriovs`

---

*[ocaml/xapi/xapi_host.ml, line 1527 at r6](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6Vrdis-WYbEaBlsfkA:-L6Vrdis-WYbEaBlsfkB:bqb7tmq) ([raw file](https://github.com/xapi-project/xen-api/blob/d9daadc545a0f80e67bfc796ff300c680e965625/ocaml/xapi/xapi_host.ml#L1527)):*
> ```OCaml
>       let pif_underneath_vlan = Db.VLAN.get_tagged_PIF ~__context ~self:vlan in
>       Db.PIF.get_network ~__context ~self:pif_underneath_vlan
>     | _ -> failwith (Printf.sprintf "Cannot find vlan from a vlan master PIF:%s" vlan_pif_rec.API.pIF_uuid)
> ```

I think we should avoid the use of `failwith`. I see it here and below, are we sure we want to raise untractable exceptions? Shouldn't this be an internal error?

---

*[ocaml/xapi/xapi_host.ml, line 1586 at r6](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6Vrwg5iPmZWahHme-b:-L6Vrwg5iPmZWahHme-c:bkcuhq9) ([raw file](https://github.com/xapi-project/xen-api/blob/d9daadc545a0f80e67bfc796ff300c680e965625/ocaml/xapi/xapi_host.ml#L1586)):*
> ```OCaml
>       let transport_pif = Db.Tunnel.get_transport_PIF ~__context ~self:tunnel in
>       Db.PIF.get_network ~__context ~self:transport_pif
>     | _ -> failwith (Printf.sprintf "PIF %s has no tunnel_access_PIF_of" access_pif_rec.API.pIF_uuid)
> ```

As I said above, I tend to be wary of `failwith`. Are we sure it is better than an internal error or some other specific error?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3474)*

krizex:
Review status: 0 of 11 files reviewed at latest revision, 13 unresolved discussions.

---

*[ocaml/idl/datamodel.ml, line 4999 at r6](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6VrIDhqBbCwfWYW1xk:-L6VwQjuAV5skr26n46o:b-k15u7p) ([raw file](https://github.com/xapi-project/xen-api/blob/d9daadc545a0f80e67bfc796ff300c680e965625/ocaml/idl/datamodel.ml#L4999)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

You have updated the `host_sunc_tunnels` function instead of `sync_network_sriovs`
</blockquote></details>

Oh, my bad.

---

*[ocaml/xapi/xapi_host.ml, line 1527 at r6](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6Vrdis-WYbEaBlsfkA:-L6VxomVTSHQypy-o6F-:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/d9daadc545a0f80e67bfc796ff300c680e965625/ocaml/xapi/xapi_host.ml#L1527)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

I think we should avoid the use of `failwith`. I see it here and below, are we sure we want to raise untractable exceptions? Shouldn't this be an internal error?
</blockquote></details>

Done.

---

*[ocaml/xapi/xapi_host.ml, line 1586 at r6](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6Vrwg5iPmZWahHme-b:-L6Vxv7r6SqoAPH1ffOZ:b-ci2f1g) ([raw file](https://github.com/xapi-project/xen-api/blob/d9daadc545a0f80e67bfc796ff300c680e965625/ocaml/xapi/xapi_host.ml#L1586)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

As I said above, I tend to be wary of `failwith`. Are we sure it is better than an internal error or some other specific error?
</blockquote></details>

No, I think internal error could be better.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3474)*

robhoes:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[ocaml/xapi/xapi_pif.ml, line 889 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6REgt5j4mDXBSwaKDK:-L6XUfaVyjQ-oz-_jC0r:b-5e7qtj) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_pif.ml#L889)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

The order probably doesn't matter...?
</blockquote></details>

If the physical PIF needs to be up for SR-IOV to work, then plugging an SR-IOV PIF should automatically plug the underlying physical one as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3474)*

krizex:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[ocaml/xapi/xapi_host.ml, line 1532 at r10](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6XS4lSli-5wyBcD67L:-L6ZLHHdbPID4U1_54OR:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/85fa01733575be000d0a62e1d34f654163e336af/ocaml/xapi/xapi_host.ml#L1532)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

Let's fix the indentation, while we're at it...
</blockquote></details>

Done.

---

*[ocaml/xapi/xapi_pif.ml, line 889 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6REgt5j4mDXBSwaKDK:-L6ZU4panvsR_rVsW58r:b-rnnkh5) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_pif.ml#L889)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

If the physical PIF needs to be up for SR-IOV to work, then plugging an SR-IOV PIF should automatically plug the underlying physical one as well.
</blockquote></details>

I think bring up physical PIF and SR-IOV PIF is two independent events, because bring up SR-IOV PIF does not depend on the bridge of the physical PIF. In `Xapi_pif.plug` when plugging SR-IOV logical PIF, we do not plug the underneath physical PIF automatically.

So the bring up order between SR-IOV PIFs and (bond master and physical PIF) does not matter, but I keep separate them to make the code clearer.

---

*[ocaml/xapi/xapi_pif.ml, line 866 at r8](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6XQ09OtN5hLv45DyYi:-L6ZM8sv2kPPU_szLwMq:b9gh154) ([raw file](https://github.com/xapi-project/xen-api/blob/2675a97b4bfa88bf1890e2a892fff9368faaf443/ocaml/xapi/xapi_pif.ml#L866)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

Why this filter? Shouldn't we just enabled SR-IOV for all?
</blockquote></details>

No, user could enable/disable SR-IOV by plug/unplug, we should respect user's choice when xapi restarting (include reboot).
So, when xapi restarting, we should plug the PIF which SR-IOV is enabled:
For sysfs PIF, we should plug it if `PIF.currently-attached = true`, because sysfs PIF need re-enable after reboot but we cannot distinguish xapi restart from host reboot, so the plan is always do this when xapi restart.
For modprobe PIF, we should plug it if `PIF.currently-attached = false` and `Network_sriov.requires_reboot = true`

I missed some of the logic above, will update accordingly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3474)*

mseri:
Reviewed 1 of 12 files at r2, 1 of 10 files at r15, 5 of 7 files at r16, 1 of 2 files at r17, 1 of 1 files at r18, 1 of 1 files at r19.
Review status: 10 of 11 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3474)*

robhoes:
Reviewed 10 of 10 files at r15, 7 of 7 files at r16, 2 of 2 files at r17, 1 of 1 files at r18.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[ocaml/xapi/xapi_pif.ml, line 889 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6REgt5j4mDXBSwaKDK:-L6aIlt4RMh0DpQe2TSd:boyf36v) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_pif.ml#L889)):*
<details><summary><i>Previously, krizex (Yang Qian) wrote…</i></summary><blockquote>

I think bring up physical PIF and SR-IOV PIF is two independent events, because bring up SR-IOV PIF does not depend on the bridge of the physical PIF. In `Xapi_pif.plug` when plugging SR-IOV logical PIF, we do not plug the underneath physical PIF automatically.

So the bring up order between SR-IOV PIFs and (bond master and physical PIF) does not matter, but I keep separate them to make the code clearer.
</blockquote></details>

Plugging the physical PIF does a little more than just creating the bridge: it also brings up the eth device in the kernel. In the SR-IOV case, this is the physical function (PF) device driver that gets engaged for the device. It may be the case that the PF device needs to be "up" in order for the VF devices to work.

If that is true, then we should auto-plug the physical PIF underneath the SR-IOV PIF if the latter is plugged. We should do some tests to verify this, and fix it in a separate PR, if needed. This won't affect the plug order in this function.

---

*[ocaml/xapi/xapi_pif.ml, line 866 at r8](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6XQ09OtN5hLv45DyYi:-L6aT9EvJ6QWXy46ueeN:brdq2jo) ([raw file](https://github.com/xapi-project/xen-api/blob/2675a97b4bfa88bf1890e2a892fff9368faaf443/ocaml/xapi/xapi_pif.ml#L866)):*
<details><summary><i>Previously, krizex (Yang Qian) wrote…</i></summary><blockquote>

No, user could enable/disable SR-IOV by plug/unplug, we should respect user's choice when xapi restarting (include reboot).
So, when xapi restarting, we should plug the PIF which SR-IOV is enabled:
For sysfs PIF, we should plug it if `PIF.currently-attached = true`, because sysfs PIF need re-enable after reboot but we cannot distinguish xapi restart from host reboot, so the plan is always do this when xapi restart.
For modprobe PIF, we should plug it if `PIF.currently-attached = false` and `Network_sriov.requires_reboot = true`

I missed some of the logic above, will update accordingly.
</blockquote></details>

There's a couple of points to note about this:

- Unplugging a PIF (or a PBD, a VIF, or anything else that can be unplugged), is considered to be relatively uncommon, and usually temporary. Disabling a device for a longer term can be done by removing the PIF completely (for SR-IOV, that means `network_sriov.destroy`).
- When a VM starts up, all PIFs that is needs are automatically plugged. If PIFs were already plugged, then the VM start will be a little quicker, because the network was already on "standby".
- It is standard practice for unplugged stuff to be automatically plugged again when the host (for PIFs, PBDs, ...), or a VM (for VIFs, VBDs, ...) restarts, even though they we unplugged before shutting down.
- For the physical devices, the main reason for doing to is that dom0 itself needs those devices to be connected (e.g. for the management network). But it also helps to have them on standby for VMs.
- However, we don't plug all types of PIFs on startup, for performance reasons. Especially for the case of VLANs and tunnels, there can in theory be >1000, which take long time to all plug. And in realilty, you may only need a fraction of those for VMs on the localhost. We do auto-plug VLAN PIFs with IP addresses, because dom0 will need those to be up.
- For SR-IOV, there is never more than 1 logical PIF per physical PIF, so plugging them all on startup should not cost too much. At least for the sysfs ones, so I would always plug those. The modprobe ones are a little awkward (as we know), because they may need a reboot, but I think that it is overall simpler to just plug them as well.
- Regarding `PIF.currently_attached`, there is one more startup item that we need to look  at: the `sync_pif_currently_attached` in xapi_host.ml. Currently, this updated the field based on whether a bridge exists or not, which isn't valid for SR-IOV. It runs in dbsync, before "Best-effort bring up of physical NICs", and will therefore break the code you have here in this patch. However, if we plug all SR-IOV PIFs on startup anyway, then this does not matter.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3474)*

krizex:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[ocaml/xapi/xapi_pif.ml, line 889 at r1](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6REgt5j4mDXBSwaKDK:-L6b3otT2uNBmqtqqwmP:b-kkofco) ([raw file](https://github.com/xapi-project/xen-api/blob/82ba75875052f5c720e6de9fd9d493254c4d3e23/ocaml/xapi/xapi_pif.ml#L889)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

Plugging the physical PIF does a little more than just creating the bridge: it also brings up the eth device in the kernel. In the SR-IOV case, this is the physical function (PF) device driver that gets engaged for the device. It may be the case that the PF device needs to be "up" in order for the VF devices to work.

If that is true, then we should auto-plug the physical PIF underneath the SR-IOV PIF if the latter is plugged. We should do some tests to verify this, and fix it in a separate PR, if needed. This won't affect the plug order in this function.
</blockquote></details>

OK, I will test the scenario. If needed I will create another PR to fix the issue above.

---

*[ocaml/xapi/xapi_pif.ml, line 866 at r8](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-L6XQ09OtN5hLv45DyYi:-L6b2ux0ZWLEtkIR05zr:b-pa9eaw) ([raw file](https://github.com/xapi-project/xen-api/blob/2675a97b4bfa88bf1890e2a892fff9368faaf443/ocaml/xapi/xapi_pif.ml#L866)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

There's a couple of points to note about this:

- Unplugging a PIF (or a PBD, a VIF, or anything else that can be unplugged), is considered to be relatively uncommon, and usually temporary. Disabling a device for a longer term can be done by removing the PIF completely (for SR-IOV, that means `network_sriov.destroy`).
- When a VM starts up, all PIFs that is needs are automatically plugged. If PIFs were already plugged, then the VM start will be a little quicker, because the network was already on "standby".
- It is standard practice for unplugged stuff to be automatically plugged again when the host (for PIFs, PBDs, ...), or a VM (for VIFs, VBDs, ...) restarts, even though they we unplugged before shutting down.
- For the physical devices, the main reason for doing to is that dom0 itself needs those devices to be connected (e.g. for the management network). But it also helps to have them on standby for VMs.
- However, we don't plug all types of PIFs on startup, for performance reasons. Especially for the case of VLANs and tunnels, there can in theory be >1000, which take long time to all plug. And in realilty, you may only need a fraction of those for VMs on the localhost. We do auto-plug VLAN PIFs with IP addresses, because dom0 will need those to be up.
- For SR-IOV, there is never more than 1 logical PIF per physical PIF, so plugging them all on startup should not cost too much. At least for the sysfs ones, so I would always plug those. The modprobe ones are a little awkward (as we know), because they may need a reboot, but I think that it is overall simpler to just plug them as well.
- Regarding `PIF.currently_attached`, there is one more startup item that we need to look  at: the `sync_pif_currently_attached` in xapi_host.ml. Currently, this updated the field based on whether a bridge exists or not, which isn't valid for SR-IOV. It runs in dbsync, before "Best-effort bring up of physical NICs", and will therefore break the code you have here in this patch. However, if we plug all SR-IOV PIFs on startup anyway, then this does not matter.
</blockquote></details>

Thanks for your comprehensive explanation, so for any existing sriov PIFs, we should plug them when xapi start. If user want to disable SR-IOV for some PIF forever, he should destroy rather than unplug it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3474)*

robhoes:
Reviewed 3 of 3 files at r20, 1 of 1 files at r21, 1 of 1 files at r22.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3474)*

robhoes:
:lgtm:

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xen-api/3474#-:-L6b7mZMX6-fBEVpZAHV:bnfp4nl)*

PR #3480, opened 2018-02-27
Convert Test_daemon_manager, Test_agility, Test_basic to Alcotest

gaborigloi:
We've recently started porting xapi's unit tests from OUnit, to Alcotest. 

In https://github.com/xapi-project/xen-api/pull/3402 we've added support for unit testing xapi using alcotest. Now there are two suites, one using ounit and another one using alcotest. New unit tests should use alcotest.

Moving to alcotest is beneficial because it has better output, thereby reducing log spam and making it easier to fix and write unit tests, and is probably lighter, more actively maintained, and more commonly used.

To get the full benefits of alcotest, we'd have to port all of our unit tests to it.

This PR ports some xapi unit tests from OUnit to Alcotest.

Useful links:

-  Alcotest website with link to docs: https://github.com/mirage/alcotest
- OUnit docs: http://ounit.forge.ocamlcore.org/api-ounit/OUnit.html


ocaml/tests/test_basic.ml
```diff
@@ -27,19 +23,7 @@ let test_mock_db () =
   ignore (Db.Blob.get_record ~__context ~self:blob_ref) ;
   ignore (Db.VM.get_all_records ~__context) ;
   let blob_name = Db.Blob.get_name_label ~__context ~self:blob_ref in
-  assert_equal blob_name "BLOB"
-
-let test_assert_licensed_storage_motion () = skip "TODO" ;
-  let __context = Mock.make_context_with_new_db "Mock context" in
-  let licensed = try Xapi_vm_migrate.assert_licensed_storage_motion ~__context; true
-    with _ -> false in
-  assert_bool "Not licensed for SXM" licensed
```
minishrink: Is this test no longer valid or is this function tested elsewhere?

ocaml/tests/test_daemon_manager.ml
```diff
@@ -138,19 +136,19 @@ let test_timeout_fail () =
...
       time_until_stopped = 5.0;
     };
-  assert_raises (Failure "stop failed")
+  Alcotest.check_raises "does not stop within timeout"
+    (Failure "stop failed")
     (fun () -> Mock_manager.with_daemon_stopped ~timeout:2.0 (fun () -> ()));
-  assert_equal !Mock_daemon.times_called_start 0;
-  assert_equal !Mock_daemon.times_called_stop 1
+  Alcotest.(check int) "times_called_start" !Mock_daemon.times_called_start 0;
+  Alcotest.(check int) "time_until_stopped" !Mock_daemon.times_called_stop 1
```
minishrink: I notice that lines 142-143 are repeated 7 times with different numbers, is it worth making an alias function?

ocaml/tests/test_basic.ml
gaborigloi: I decided to simply delete this because it was not run: at line 32, we have `skip "TODO"`

minishrink: Is it worth removing the `skip "TODO"` so it runs and including it?

gaborigloi: I don't know whether it works or not, but probably not because it was skipped. My thinking was that it's probably not worth trying to debug old legacy tests :thinking:

johnelse: FYI all the tests in Test_basic are really just demo tests that were written when unit tests first became a thing - they're not particularly useful as actual tests of the xapi codebase.

gaborigloi: Oh ok, thanks a lot for the info @johnelse  - then I'll just remove it completely :smile:

ocaml/tests/test_daemon_manager.ml
gaborigloi: Sounds good, done

PR #3483, opened 2018-03-02
Port test_clustering to Alcotest

minishrink:
Motivation: Alcotest allows for cleaner error reporting and debugging. Additionally, I will be submitting a PR that adds a unit test to test_clustering which is written in Alcotest, hence a port was necessary.

Changes in test_clustering:
- `assert_equal` now uses an Alcotest comparator
- every individual unit test is now a toplevel function

Changes in test_clustering_allowed_operations:
- OUnit.assert_bool -> Alcotest.(check bool)

### DIFF
--- a/ocaml/tests/test_clustering.ml
+++ b/ocaml/tests/test_clustering.ml
@@ -12,249 +12,259 @@
  * GNU Lesser General Public License for more details.
  *)
 
-(** Tests for the Xapi_clustering module *)
+(** Test for the Xapi_clustering module *)
 
 module T = Test_common
 
-(** Tests Xapi_clustering.get_required_cluster_stacks *)
+(** Test Xapi_clustering.get_required_cluster_stacks *)
+
+let assert_equal ~msg ~required_cluster_stacks ~to_set =
+  let module S = Set.Make(String) in
+  let setify sl = sl |> S.of_list |> S.elements in
+  Alcotest.(check (slist string String.compare)) msg
+    (setify required_cluster_stacks) (setify to_set)
+
+let test_zero_sms_in_database () =
+  let __context = T.make_test_database () in
+  let required_cluster_stacks = Xapi_clustering.get_required_cluster_stacks ~__context ~sr_sm_type:"" in
+  assert_equal
+    ~msg:"Asserted by test_zero_sms_in_database"
+    ~required_cluster_stacks ~to_set:[]
+
+let test_zero_sms_with_matching_type_which_do_require_cluster_stack () =
+  let __context = T.make_test_database () in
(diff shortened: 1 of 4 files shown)


ocaml/tests/test_clustering.ml
```diff
@@ -293,15 +306,10 @@ let test_assert_pif_prerequisites () =
...
-  ; test_assert_cluster_host_is_enabled_for_matching_sms
-  ; test_clustering_lock_only_taken_if_needed
-  ; "test_assert_pif_prerequisites" >:: test_assert_pif_prerequisites
-  ]
+  Alcotest.(check unit)
+    "assert_pif_prerequisites should pass after setting disallow_unplug:true"
+    (Xapi_clustering.assert_pif_prerequisites pif) ()
+
+let test_assert_pif_prerequisites =
+  [ "test_assert_pif_prerequisites", `Quick, test_assert_pif_prerequisites ]
```
gaborigloi: I think we should have all the tests in this file in a `tests` variable, like in `test_vdi_cbt.ml`, otherwise the main `alcotest_suite.ml` will get huge:
```
let tests =
  test_assert_pif_prerequisites @
  test_clustering_lock_only_taken_if_needed @ ...
```
And we should ensure that the names of the above tests are specific enough, because alcotest does not do the subgroup namespacing done by OUnit, because it does not have nested test groups.

ocaml/tests/test_clustering.ml
```diff
@@ -12,249 +12,259 @@
...
-    let required_cluster_stacks = Xapi_clustering.get_required_cluster_stacks ~__context ~sr_sm_type:"" in
-    assert_equal ~required_cluster_stacks ~to_set:[]
-  in
+(** Test Xapi_clustering.get_required_cluster_stacks *)
+
+let assert_equal ~msg ~required_cluster_stacks ~to_set =
+  let module S = Set.Make(String) in
+  let setify sl = sl |> S.of_list |> S.elements in
+  Alcotest.(check (slist string String.compare)) msg
+    (setify required_cluster_stacks) (setify to_set)
```
gaborigloi: @edwintorok Is it okay that `get_required_cluster_stacks` might return repeated elements, or should we put this into the code itself instead of the test?

edwintorok: I think that `get_required_cluster_stacks` should return either an empty list, or a list of one element. If it returns more than one element that is not going to work: different cluster stacks are mutually incompatible.
So perhaps we should fix the code to return an option, and raise an exception if the list ever contains more than 2 elements.

gaborigloi: I think the way this function was originally intended is that this is a disjunction not a conjunction: we require one of these cluster stacks for the SM to work, not all. So maybe this function assumes that there will be SMs in the future that work with multiple different cluster stacks.

PR #3485, opened 2018-03-05
CA-268511: Make PIF.disallow_unplug RO when clustering is enabled

minishrink:
This PR makes the following changes:
- PIF.disallow_unplug is now dynamic RO
- PIF.set_disallow_unplug throws `cannot_change_pif_properties` when a cluster object is attached to the network
- adds two tests to test_clustering to check the behaviour of `PIF.set_disallow_unplug` with and without clustering

Questions for reviewers about possible changes:
- should it raise a different exception if clustering is enabled?
- is the try-with unnecessary for getting `PIF.network`
- should the logging use `info` or` debug`?


ocaml/idl/datamodel.ml
```diff
@@ -954,6 +954,17 @@ module PIF = struct
              Api_errors.pif_has_fcoe_sr_in_use]
       ()
 
+  let pif_set_disallow_unplug = call
```
gaborigloi: We don't need to prefix this with `pif_` as it's in already the PIF module - this seems to be the convention in this module.

ocaml/idl/datamodel.ml
```diff
@@ -954,6 +954,17 @@ module PIF = struct
...
+    ~name:"set_disallow_unplug"
+    ~doc:"Set whether unplugging the PIF is allowed"
+    ~hide_from_docs:false
+    ~in_oss_since:None
+    ~in_product_since:rel_orlando
+    ~params: [ Ref _pif, "self", "Reference to the object"
+             ; Bool, "value", "New value to set" ]
+    ~allowed_roles:_R_POOL_OP
+    ()
+
```
minishrink: I assume this allowed role is fine?

gaborigloi: If that was the original role than yes.

ocaml/tests/test_clustering.ml
```diff
@@ -313,11 +313,65 @@ let test_assert_pif_prerequisites () =
...
+let check_disallow_unplug expected_value __context pif msg =
+  Alcotest.(check bool) msg
+    (Db.PIF.get_disallow_unplug ~__context ~self:pif)
+    expected_value
+
+(* Need host and network to make PIF *)
+let make_host_network_pif ~__context =
+  let host    = Test_common.make_host    ~__context () in
+  let network = Test_common.make_network ~__context () in
+  let pif     = Test_common.make_pif     ~__context ~network ~host () in
```
gaborigloi: We prefer not to align things because then we may have to change all the lines every time we add a new one.
This is one of the reasons we start list items with the semicolon: this way we only have to change one line when remove an item: http://elm-lang.org/docs/style-guide

ocaml/idl/datamodel.ml
gaborigloi: I've looked at the definitions of `let create_obj` and `let field`, and it seems that the field setter role will default to `~messages_default_allowed_roles`, if I'm not mistaken, which is this, so this is perfect.

ocaml/xapi/xapi_pif.ml
```diff
@@ -767,6 +767,31 @@ let set_property ~__context ~self ~name ~value =
...
     ) (self :: vlan_pifs)
 
+let pif_has_clusters ~__context (self : API.ref_PIF) =
+  debug "Looking up clusters for PIF: %s" (Ref.string_of self);
+  let network_of_pif =
+    try Db.PIF.get_network ~__context ~self with _ ->
+      failwith (Printf.sprintf "Could not find network of PIF: %s" (Ref.string_of self))
+  in
+  info "Got network: %s" (Ref.string_of network_of_pif);
+  (Db.Cluster.get_records_where ~__context
```
gaborigloi: We can use `get_refs_where` here - we don't need to obtain the whole record.

ocaml/xapi/message_forwarding.ml
```diff
@@ -3110,6 +3110,12 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
       let local_fn = Local.PIF.set_property ~self ~name ~value in
       do_op_on ~local_fn ~__context ~host (fun session_id rpc -> Client.PIF.set_property rpc session_id self name value)
 
+    let set_disallow_unplug ~__context ~self ~value =
+      info "PIF.set_disallow_unplug: PIF: %s; value = %s" (Ref.string_of self) (string_of_bool value);
+      let host = Db.PIF.get_host ~__context ~self in
+      let local_fn = Local.PIF.set_disallow_unplug ~self ~value in
+      do_op_on ~local_fn ~__context ~host (fun session_id rpc -> Client.PIF.set_disallow_unplug rpc session_id self value)
```
gaborigloi: I think we should not forward it but just do it locally on the master which has the DB, since we only do DB queries. (Like for `vdi_set_cbt_enabled`)

ocaml/xapi/xapi_pif.ml
```diff
@@ -767,6 +767,31 @@ let set_property ~__context ~self ~name ~value =
...
+    | []       -> false
+    | (a,_)::_ -> true
+
+let assert_no_clustering_enabled ~__context ~self =
+  if pif_has_clusters ~__context self
+  then raise Api_errors.(Server_error(Api_errors.cannot_change_pif_properties, [Ref.string_of self]))
+
+let set_disallow_unplug ~__context ~self ~value =
+  assert_no_clustering_enabled ~__context ~self;
+  if Db.PIF.get_disallow_unplug ~__context ~self <> value
```
gaborigloi: I think this idempotency is not necessary here, as there is nothing in the other `if` branch

ocaml/xapi/xapi_pif.ml
```diff
@@ -767,6 +767,31 @@ let set_property ~__context ~self ~name ~value =
...
+  info "Got network: %s" (Ref.string_of network_of_pif);
+  (Db.Cluster.get_records_where ~__context
+    ~expr:Db_filter_types.(Eq(Literal (Ref.string_of network_of_pif),Field "network")))
+    |> function
+    | []       -> false
+    | (a,_)::_ -> true
+
+let assert_no_clustering_enabled ~__context ~self =
+  if pif_has_clusters ~__context self
+  then raise Api_errors.(Server_error(Api_errors.cannot_change_pif_properties, [Ref.string_of self]))
```
gaborigloi: One issue with this exception is that its description currently talks about bonded PIFs (in `datamodel.ml`)

ocaml/xapi/message_forwarding.ml
robhoes: Yes, if the function act on the xapi DB only, than we should just call `Local...` directly.

ocaml/xapi/xapi_pif.ml
```diff
@@ -767,6 +767,31 @@ let set_property ~__context ~self ~name ~value =
...
+
+let set_disallow_unplug ~__context ~self ~value =
+  assert_no_clustering_enabled ~__context ~self;
+  if Db.PIF.get_disallow_unplug ~__context ~self <> value
+  then begin
+    info "PIF.set_disallow_unplug: PIF: %s, value: %s" (Ref.string_of self) (string_of_bool value);
```
robhoes: I would remove this debug line, since it's almost the same as the one in message_forwarding.

ocaml/xapi/xapi_pif.ml
robhoes: Indeed you don't save very much by this if-statement. In fact, the common case would be that the function takes 2 DB calls rather than just one. However, this is not such a big deal if this only runs on the master.

ocaml/xapi/xapi_pif.ml
robhoes: Let's introduce a new API error that is specific to clustering.

robhoes: Minor: you can remove the second mention of `Api_errors.`.

ocaml/xapi/xapi_pif.ml
```diff
@@ -767,6 +767,31 @@ let set_property ~__context ~self ~name ~value =
         Nm.bring_pif_up ~__context pif
     ) (self :: vlan_pifs)
 
+let pif_has_clusters ~__context (self : API.ref_PIF) =
+  debug "Looking up clusters for PIF: %s" (Ref.string_of self);
+  let network_of_pif =
+    try Db.PIF.get_network ~__context ~self with _ ->
```
robhoes: If PIF has a network by construction, and will be GC'ed if the network is deleted. So this is a little overly defensive.

ocaml/xapi/xapi_pif.ml
```diff
@@ -767,6 +767,31 @@ let set_property ~__context ~self ~name ~value =
...
+      failwith (Printf.sprintf "Could not find network of PIF: %s" (Ref.string_of self))
+  in
+  info "Got network: %s" (Ref.string_of network_of_pif);
```
robhoes: I would remove the logging in this function. We are logging far too much already...

ocaml/xapi/xapi_pif.ml
```diff
@@ -767,6 +767,31 @@ let set_property ~__context ~self ~name ~value =
...
+  (Db.Cluster.get_records_where ~__context
+    ~expr:Db_filter_types.(Eq(Literal (Ref.string_of network_of_pif),Field "network")))
```
robhoes: This actually works? I always see `Field` first and then `Literal`...

ocaml/tests/test_clustering.ml
edwintorok: I think this would be a good addition to https://github.com/lindig/ocaml-style

ocaml/xapi/xapi_pif.ml
edwintorok: Also when you add a debug message it would be nice if it printed UUIDs instead of OpaqueRefs, it makes debugging easier with `xe`. I've fixed that in a few places in the clustering code.
Don't fix it on this one since you're dropping it anyway, but in the future.

gaborigloi: I think that's not a huge problem if we have the database dump, because then we can match UUIDs to opaquerefs.

ocaml/tests/test_clustering.ml
gaborigloi: Indeed, what do you think @lindig ? I think it'd be good to mention that we shouldn't worry too much about aligning things but should keep it simple and practical, to minimise the effort when viewing diffs or changing code or rebasing.

ocaml/xapi/message_forwarding.ml
```diff
@@ -2666,14 +2666,14 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
-          Client.Host.set_iscsi_iqn rpc session_id host value)
+           Client.Host.set_iscsi_iqn rpc session_id host value)
 
     let set_multipathing ~__context ~host ~value =
       info "Host.set_multipathing: host='%s' value='%s'" (host_uuid ~__context host) (string_of_bool value);
       let local_fn = Local.Host.set_multipathing ~host ~value in
       do_op_on ~local_fn ~__context ~host
         (fun session_id rpc ->
-          Client.Host.set_multipathing rpc session_id host value)
+           Client.Host.set_multipathing rpc session_id host value)
```
gaborigloi: Does your editor ocp-indent automatically? :o

ocaml/xapi/message_forwarding.ml
```diff
@@ -3110,6 +3110,11 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
       let local_fn = Local.PIF.set_property ~self ~name ~value in
       do_op_on ~local_fn ~__context ~host (fun session_id rpc -> Client.PIF.set_property rpc session_id self name value)
 
+    let set_disallow_unplug ~__context ~self ~value =
+      let pif_uuid = Db.PIF.get_uuid ~__context ~self in
```
gaborigloi: I think it's best to stick the the convention here and only print the reference, this is what we seem to do everywhere else in message forwarding - for me, the inconsistency would be more confusing when looking at the logs, because I expect message-forwarding to log refs not UUIDs.

ocaml/xapi/message_forwarding.ml
```diff
@@ -3110,6 +3110,11 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
+      info "PIF.set_disallow_unplug: PIF uuid = %s; value = %s" (pif_uuid) (string_of_bool value);
```
gaborigloi: You can use the existing `pif_uuid` helper: https://github.com/xapi-project/xen-api/pull/3485/files#diff-9a639b670fb7e2432894cc91df8b27fbR3101

ocaml/xapi/xapi_pif.ml
```diff
@@ -767,6 +767,19 @@ let set_property ~__context ~self ~name ~value =
...
+  let network_ref = Db.PIF.get_network ~__context ~self |> Ref.string_of in
+  (Db.Cluster.get_refs_where ~__context
+     ~expr:Db_filter_types.(Eq(Literal network_ref,Field "network")))
+  |> function
+  | []   -> false
+  | a::_ -> true
+
+let set_disallow_unplug ~__context ~self ~value =
+  if pif_has_clustering_enabled ~__context self
+  then raise Api_errors.(Server_error(cannot_change_pif_properties, [Ref.string_of self]))
```
gaborigloi: As we Rob said, we should introduce a new API error that is specific to clustering.

ocaml/xapi/xapi_pif.ml
```diff
@@ -767,6 +767,19 @@ let set_property ~__context ~self ~name ~value =
         Nm.bring_pif_up ~__context pif
     ) (self :: vlan_pifs)
 
+let pif_has_clustering_enabled ~__context (self : API.ref_PIF) =
+  let network_ref = Db.PIF.get_network ~__context ~self |> Ref.string_of in
+  (Db.Cluster.get_refs_where ~__context
+     ~expr:Db_filter_types.(Eq(Literal network_ref,Field "network")))
```
gaborigloi: As discussed, maybe it's safer to put `Field` first and then `Literal` to follow the existing conventions.

ocaml/tests/test_clustering.ml
```diff
@@ -313,11 +313,65 @@ let test_assert_pif_prerequisites () =
...
+  let host,network,pif = make_host_network_pif ~__context in
+  check_disallow_unplug false __context pif
+    "check_disallow_unplug called by test_disallow_unplug_with_clustering to check default config";
+
+  (* PIF.disallow_unplug should become RO upon introduce cluster_host object *)
+  let _ = Test_common.make_cluster_and_cluster_host ~__context ~network ~host () in
+  Alcotest.check_raises
+    "check_disallow_unplug called by test_disallow_unplug_with_clustering after attaching cluster and cluster_host to network"
+    (Api_errors.(Server_error(clustering_enabled_on_network, [Ref.string_of network])))
+    (fun () -> Xapi_pif.set_disallow_unplug ~__context ~self:pif ~value:true)
```
edwintorok: You might want to add a second check that setting `disallow_unplug` to false raises as well. In the current implementation it is obvious it will, but the real goal is that we reject setting it to false.

ocaml/xapi/xapi_pif.ml
```diff
@@ -767,6 +767,21 @@ let set_property ~__context ~self ~name ~value =
...
+  (Db.Cluster.get_refs_where ~__context
+     ~expr:Db_filter_types.(Eq(Field "network", Literal network_ref)))
+  |> function
+  | []   -> false
+  | a::_ -> true
+
+let set_disallow_unplug ~__context ~self ~value =
+  if pif_has_clustering_enabled ~__context self
+  then
+    let network_ref = Db.PIF.get_network ~__context ~self |> Ref.string_of in
```
robhoes: We already got this from the DB in the function above, so this could be optimised a little. Not a big deal though.

robhoes:
Just added a minor comment, but generally LGTM.

lindig:
I would like to see commits to include the CA unless the the history is cleaned.

edwintorok:
I think all the commits could be squashed into one, it is logically one set of self-consistent changes.

PR #3486, opened 2018-03-05
CA-282112: Fix domain_type setting during SXM from older releases

robhoes:
The new `VM.domain_type` field need to be defined (derived from the
`HVM_boot_policy`) as soon as possible upon importing the VM record, since lots
of functionality now depends on it. Import pre-checks were failing because this
happened too late.


jonludlam:
Looks good! Have we checked for any other uses of `From.vM_t`?

robhoes:
@jonludlam, yes, the were only present in import.ml, and I've replaced them all by this new function.

PR #3488, opened 2018-03-05
Fix release info for domain_type fields

robhoes:

### DIFF
--- a/ocaml/idl/datamodel_vm.ml
+++ b/ocaml/idl/datamodel_vm.ml
@@ -1394,7 +1394,12 @@ let set_HVM_boot_policy = call ~flags:[`Session]
            field ~qualifier:DynamicRO ~ty:Bool ~lifecycle:[Published, rel_ely, ""] ~default_value:(Some (VBool false))
              "requires_reboot" "Indicates whether a VM requires a reboot in order to update its configuration, e.g. its memory allocation.";
            field ~qualifier:StaticRO ~ty:String ~in_product_since:rel_ely ~default_value:(Some (VString "")) "reference_label" "Textual reference to the template used to create a VM. This can be used by clients in need of an immutable reference to the template since the latter's uuid and name_label may change, for example, after a package installation or upgrade.";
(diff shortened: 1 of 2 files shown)


ocaml/idl/datamodel_vm.ml
```diff
@@ -1394,7 +1394,12 @@ let set_HVM_boot_policy = call ~flags:[`Session]
            field ~qualifier:DynamicRO ~ty:Bool ~lifecycle:[Published, rel_ely, ""] ~default_value:(Some (VBool false))
              "requires_reboot" "Indicates whether a VM requires a reboot in order to update its configuration, e.g. its memory allocation.";
            field ~qualifier:StaticRO ~ty:String ~in_product_since:rel_ely ~default_value:(Some (VString "")) "reference_label" "Textual reference to the template used to create a VM. This can be used by clients in need of an immutable reference to the template since the latter's uuid and name_label may change, for example, after a package installation or upgrade.";
-           field ~qualifier:StaticRO ~ty:domain_type ~lifecycle:[Published, rel_jura, ""] ~default_value:(Some (VEnum "unspecified")) "domain_type" "The type of domain that will be created when the VM is started";
+           field ~qualifier:StaticRO ~ty:domain_type
+             ~lifecycle:[
+               Prototyped, rel_jura, "Internal-only field; not yet in the public API";
```
edwintorok: in `datamodel.ml` this field is `Published`, this one is `Prototyped`. Is that intended?

robhoes:
Yes, this is just for the docs.

ocaml/idl/datamodel_vm.ml
robhoes: I did it this way, because the `VM.domain_type` field was part of the datamodel in jura, but it was marked as `internal_only`, and therefore not visible in the external API. Therefore, "published" in kolkata. The `VM_metrics.current_domain_type` was also includes in jura, but not `internal_only`, and I had therefore marked it as "published" in jura already (I wanted to also make this field internal, but somehow the datamodel didn't allow this).

However, even `VM_metrics.current_domain_type` wasn't actually useful in jura, because we never set it to anything other than `unspecified`. Perhaps it is better to call it "prototyped" as well for this reason.

robhoes: Submitted a fixup for this.

PR #3490, opened 2018-03-06
Remove Legacy API module

mseri:
I had an old experiment lying around, I've decided to resurrect it and polish it after a Slack discussion today. I've rewritten the commit history to reflect more it's evolution and avoid intermediate debugging commits.

I am going to test it on a private branch, this should not be merged until testing is done and we agree on what to do with the `Bigbuffer` story (look for `Bigbuffer.to_string`, it appears in a few places in the new code).


mseri:
I think originally the issue with `Bigbuffer` was that the maximum size of 32bits ocaml strings was 16Mb. The size of strings on 64 bit systems is on the order of 144 Petabytes (output of `Sys.max_string_length`, thanks @lindig)

gaborigloi:
Hope this will be merged soon as it will make my work easier ^^

mseri:
I'm running BVT and BST right now, if there are no further failures, and if somebody has the patience to review it in the meanwhile, we could proceed

gaborigloi:
Actually I only need the commits that remove the legacy module.

gaborigloi:
Figured out how to solve my problem, so it's not that urgent now :)

mseri:
The others are necessary to remove the Legacy module :P

mseri:
For reference:
```
~/src/xen-api master $ cat _build/default/ocaml/xapi-types/aPI.ml | wc -l
6862

~/src/xen-api no-legacy $ cat _build/default/ocaml/xapi-types/aPI.ml | wc -l
3144
```

ocaml/idl/ocaml_backend/gen_api.ml
```diff
@@ -82,10 +82,19 @@ let gen_record_type ~with_module highapi tys =
...
+        in
+        match default_value with
+          None -> "None"
+        | Some default -> sprintf "(Some (%s))" (Datamodel_values.to_ocaml_string ~v2:true default)
+      in
       let make_to_field fld =
-        sprintf "%s = %s_of_rpc (List.assoc \"%s\" x)" (field fld) (OU.alias_of_ty fld.DT.ty)
-          (String.concat "_" fld.DT.full_name)
+        sprintf {|%s = %s_of_rpc (assocer "%s" x %s)|} (field fld) (OU.alias_of_ty fld.DT.ty)
+          (String.concat "_" fld.DT.full_name) (get_default fld)
```
gaborigloi: It is not necessary to generate these `X_of_rpc`, and `rpc_of_X` things, rpclib can do it for us. I have a set of working patches, one of this removes these custom functions and moves to ones generated by rpclib:
https://github.com/gaborigloi/xen-api/commit/2c4fc73cf72c192feaaf091c80c78534a7729fd2

mseri: Not in this case, we need to customise them and replace the autogenerated ones.

mseri: Maybe we can use the new `[@default ...]` key but it needs more work to serialize it correctly and it needs further testing. It is an improvement that can be added later and we should keep in mind to try it

ocaml/idl/ocaml_backend/gen_api.ml
```diff
@@ -76,24 +76,23 @@ let gen_record_type ~with_module highapi tys =
...
+      let rpc_name fld = (String.concat "_" fld.DT.full_name) in
+      let get_default fld =
+        let default_value =
+          match fld.DT.ty with
+          | DT.Set (DT.Ref _) -> Some (DT.VSet [])
+          | _ -> fld.DT.default_value
+        in
+        match default_value with
+          None -> ""
+        | Some default -> sprintf "[@default (%s)]" (Datamodel_values.to_ocaml_string ~v2:true default)
```
gaborigloi: Jon said this is okay

mseri: :)

mseri: Unfortunately it doesn't but we have a plan to fix ocaml-rpc and make it possible

mseri:
As it is now, this code passed both the BVT and the BST :)

mseri:
If it is fine for you, I'd suggest that we review and proceed with the current changes, and we keep working to improve them to use only `ocaml-rpc` instead of generating the modules by hand. In the meanwhile ths would be further tested

ocaml/xapi/import.ml
```diff
@@ -1446,9 +1447,7 @@ let handle_all __context config rpc session_id (xs: obj list) =
 
 (** Read the next file in the archive as xml *)
 let read_xml hdr fd =
-  let xml_string = Bigbuffer.make () in
-  really_read_bigbuffer fd xml_string hdr.Tar_unix.Header.file_size;
-  Xml.parse_bigbuffer xml_string
+  Unixext.really_read_string fd (Int64.to_int hdr.Tar_unix.Header.file_size)
```
gaborigloi: This could be renamed to `read_next_file`, or we could add a `Xmlrpc.of_string` to the end.

ocaml/xapi/importexport.ml
```diff
@@ -16,52 +16,73 @@
...
+    { hostname        = Rpc.string_of_rpc (find _hostname)
+    ; date            = Rpc.string_of_rpc (find _date)
+    ; product_version = Rpc.string_of_rpc (find _product_version)
+    ; product_brand   = Rpc.string_of_rpc (find _product_brand)
+    ; build_number    = Rpc.string_of_rpc (find _build_number)
+    ; xapi_vsn_major  = Rpc.int_of_rpc (find _xapi_major)
+    ; xapi_vsn_minor  = Rpc.int_of_rpc (find _xapi_minor)
+    ; export_vsn      = try Rpc.int_of_rpc (find _export_vsn) with _ -> 0
+    }
+  | rpc -> raise (Failure(Printf.sprintf "version_of_rpc: malformed RPC %s" (Jsonrpc.to_string rpc)))
```
gaborigloi: This is a minor point, but I'd use `Rpc.to_string` here, because we're dealing with a generic Rpc structure here.

ocaml/xapi/importexport.ml
```diff
@@ -299,7 +280,12 @@ let remote_metadata_export_import ~__context ~rpc ~session_id ~remote_address ~r
            | `success -> begin
                debug "Remote metadata import succeeded";
                let result = Client.Task.get_result rpc session_id remote_task in
-               API.Legacy.From.ref_VM_set "" (Xml.parse_string result)
+               try
+                 result
+                 |> Xmlrpc.of_string
+                 |> API.ref_VM_set_of_rpc
+               with
+                 parse_error -> raise Api_errors.(Server_error (field_type_error, [Printexc.to_string parse_error]))
```
gaborigloi: This is a very minor point again, but the param of this error seems to the field:
```
  91   error Api_errors.field_type_error ["field"]                                                                                                                               
  92     ~doc:"The value specified is of the wrong type" ();
```

ocaml/xapi/xapi_vm_clone.ml
```diff
@@ -100,9 +100,12 @@ let wait_for_subtask ?progress_minmax ~__context task =
...
   let result = wait_for_subtask ?progress_minmax ~__context task in
-  let result = Xml.parse_string result in
-  let vdiref = API.Legacy.From.ref_VDI "" result in
-  vdiref
+  try
+    result 
+    |> Xmlrpc.of_string
+    |> API.ref_VDI_of_rpc
+  with
+    parse_error -> raise Api_errors.(Server_error (field_type_error, [Printexc.to_string parse_error]))
```
gaborigloi: Same as above.

ocaml/xapi/importexport.ml
mseri: Faor enough. I uesd json only because I find it more readable but it is more natural to use Rpc.to_string. I’ll change it

ocaml/database/db_interface.ml
```diff
@@ -12,9 +12,7 @@
  * GNU Lesser General Public License for more details.
  *)
 
-type response =
-  | String of string
-  | Bigbuf of Xapi_stdext_bigbuffer.Bigbuffer.t
+type response = String of string
```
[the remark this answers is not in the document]
mseri: Probably it is fine. I wanted to minimizethe changes

ocaml/xapi/import.ml
mseri: There was a reason I did not add the Xmlrpc.to_string but I think that woth the latest changes that reason is no longer there. I’ll fix it or let you know the reason :)

ocaml/xapi/importexport.ml
mseri: The error is bogus. Before we were not even printing the field. 

Using the printexc I am printing the real reason we cannot unmarshal the type. It is still a field type error, but with a useful string in the metadata. I think that even if is not what announced the errors with that content are way more readable and understandable. I agree that it is anstretch, but I will keep them as they are

mseri:
Thanks for the review. I think it’s worth making some of the changes before merging

ocaml/xapi/import.ml
mseri: Found it. The old Xva module uses a direct approach manipulating and marshalling/unmarshalling the xml. The `read_xml` function is used both by our new `Xmlrpc` and that old `Xml.parse_String |> Xva.from_xml`. Changing that is invasive and out of scope for this pr

ocaml/database/db_interface.ml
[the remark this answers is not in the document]
mseri: It's fine. Done

jonludlam:
This looks really good!

robhoes:
Nice cleanup: +202 −477 👍

PR #3491, opened 2018-03-07
Fix for update host selection and resource reservation for VM with SR-IOV VIFs

minli1:
This is a fix for host selection and reserve VF when vm without sr-iov VIF.
The function `get_local_underlying_pif ` was called in `assert_netsriov_available` and `reserve_sriov_vfs`.
These two functions will be called unconditionally when host selection and after selection to reserve VFs, so `get_local_underlying_pif ` should not raise an exception when no pif found due to exn will impacts VM start which hasn't sr-iov VIF.

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -3725,6 +3725,7 @@ module VMSS = struct
       ]
       ()
 
+<<<<<<< HEAD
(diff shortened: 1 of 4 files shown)


minli1:
Close ticket as feature branch changed, will PR base on newly feature branch soon.

PR #3492, opened 2018-03-07
CP-27272: Fix for VM without SR-IOV VIFs start failed

minli1:
This is a fix for host selection and reserve VF when vm without sr-iov VIF.
The function `get_local_underlying_pif ` was called in `assert_netsriov_available` and `reserve_sriov_vfs`.
These two functions will be called unconditionally when host selection and after selection to reserve VFs, so `get_local_underlying_pif ` should not raise an exception when no pif found due to exn will impacts VM start which hasn't sr-iov VIF.

### DIFF
--- a/ocaml/xapi/xapi_network_sriov_helpers.ml
+++ b/ocaml/xapi/xapi_network_sriov_helpers.ml
@@ -151,12 +151,8 @@ let get_sriov_networks_from_vm ~__context ~vm =
 (* Get localhost underlying pif with for the sr-iov network *)
 let get_local_underlying_pif ~__context ~network ~host =
   match Xapi_network_attach_helpers.get_local_pifs ~__context ~network ~host with
-    | [] -> raise Api_errors.(Server_error (internal_error, ["Cannot get local pif on network"]))
-    | pif :: _ ->
-      begin match get_underlying_pif ~__context ~pif with
-      | Some pif -> pif
-      | None -> raise Api_errors.(Server_error (internal_error, ["Cannot get underlying pif on sriov network"]))
-      end
+  | [] -> None
+  | pif :: _ -> get_underlying_pif ~__context ~pif
 
 (* Get remaining capacity for localhost on the given network, return None if no underlying_pif found or capacity = 0L *)
 let get_remaining_capacity_on_host ~__context ~host ~network =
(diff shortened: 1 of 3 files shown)


mseri:
Is there a ticket related to this change?

ocaml/xapi/xapi_network_sriov_helpers.ml
```diff
@@ -152,11 +152,7 @@ let get_sriov_networks_from_vm ~__context ~vm =
 let get_local_underlying_pif ~__context ~network ~host =
   match Xapi_network_attach_helpers.get_local_pifs ~__context ~network ~host with
     | [] -> raise Api_errors.(Server_error (internal_error, ["Cannot get local pif on network"]))
```
robhoes: I think you should make this return None as well rather than an exception. There are networks without PIFs as well (internal networks).

minli1:
Hi, @mseri  Yes, this is a defect found during testing after we merge feature branch and master, the CP ticket is CP-27272, I had changed the title to add the CP ticket. Thanks.
@robhoes  Thanks for your comment, updated.

cheng-z:
Hi @robhoes  and @mseri 
Please help to have a look again and merge if no other concern.
Thanks,

ocaml/xapi/xapi_network_sriov_helpers.ml
minli1: Hi, @robhoes  I had updated. I think here we indeed should return `None` for internal networks without PIFs as you mentioned. But also if the network is not _empty_, we should raise an exception because this means the host doesn't have the ability for running the VM with that `network`.
Assume VM with VGPU and SR-IOV VIF, the host was chosen by VGPU resources then we need to assert the VM can run on this host, if the selected host do not have any PIF on that `network`,  just return `None` for _get_local_pifs_ will cause the assert makes no sense.
Please take a look at the latest changes. Thanks.

robhoes: There is some overlap here with `assert_can_see_networks`, which is called in `assert_can_boot_here` before `assert_netsriov_available`. This function already checks whether the host has the PIFs that are needed by the VIFs of the VM. It seems to me that `assert_netsriov_available` should focus on just checking the available _capacity_ on the SR-IOV PIFs.

robhoes:
See inline comment. I think that we should not overload `Xapi_network_sriov_helpers.get_local_underlying_pif` with too many cases, and not mix options and exceptions. The function should just return `Some pif` if there is a local SR-IOV PIF on the given network, and `None` otherwise.

ocaml/xapi/xapi_network_sriov_helpers.ml
minli1: Yes, you are right. I didn't noticed `assert_can_see_networks` before, thanks. 
I will make `assert_netsriov_available` clean to just return an option without any internal exception.
Thanks, Rob!

minli1: Updated, please take a look again. Thanks a lot. @robhoes

robhoes:
Looks good now. Let's wait for Travis to confirm.

PR #3494, opened 2018-03-07
Add Test_client using Client module to test auto-generated layers

gaborigloi:

### DIFF
--- a/ocaml/tests/test_client.ml
+++ b/ocaml/tests/test_client.ml
@@ -1,0 +1,48 @@
+
+(** This module tests more layers of xapi than other tests by using the Client
+    module to make XenAPI calls. It tests many of the auto-generated files as
+    these XenAPI calls go through the client, server.ml, message forwarding,
+    and database layers. *)
+
+let setup_test () =
+  let __context = Test_common.make_test_database () in
+  Test_common.make_client_params ~__context
+
+(* Here we should have a unit test for each different type of method, such as
+   X.create, X.destroy, getters, and setters, to ensure that these are
+   generated correctly.
+   We use the Task class for these tests because it is simple enough that we do
+   not have to mock the storage or any other layers. *)
+
+let test_create () =
+  let rpc, session_id = setup_test () in
+  let self = Client.Client.Task.create ~rpc ~session_id ~label:"task_label" ~description:"task_description" in
+  Alcotest.(check string) "task label"
+    "task_label"
+    (Client.Client.Task.get_name_label ~rpc ~session_id ~self)
+
+let test_get_alls () =
+  let rpc, session_id = setup_test () in
+  let _t1 = Client.Client.Task.create ~rpc ~session_id ~label:"t1" ~description:"" in
+  let _t2 = Client.Client.Task.create ~rpc ~session_id ~label:"t2" ~description:"" in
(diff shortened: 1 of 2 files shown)


ocaml/tests/test_client.ml
```diff
@@ -0,0 +1,42 @@
...
+    and database layers. *)
+
+let setup_test () =
+  let __context = Test_common.make_test_database () in
+  Test_common.make_client_params ~__context
+
+let test_create () =
+  let rpc, session_id = setup_test () in
+  let self = Client.Client.Task.create ~rpc ~session_id ~label:"task_label" ~description:"task_description" in
+  Alcotest.(check string) "task label"
```
minishrink: This string could be a bit more descriptive about where a possible assertion failure would be raised.

ocaml/tests/test_client.ml
```diff
@@ -0,0 +1,42 @@
+
+(** This module tests more layers of xapi than other tests by using the Client
+    module to make XenAPI calls. It tests many of the auto-generated files as
+    these XenAPI calls go through the client, server.ml, message forwarding,
+    and database layers. *)
+
```
minishrink: I assume the difference between the target of these tests and of quicktests is that this actually tests the mechanism the quicktests use?

ocaml/tests/test_client.ml
```diff
@@ -0,0 +1,42 @@
...
+    |> List.map (fun (_, r) -> r.API.task_name_label)
+  in
+  Alcotest.(check (slist string String.compare)) "task labels from get_all_records"
+    ["t1";"t2";"t3";"t4"]
+    labels
+
+let test =
+  [ "test_create", `Quick, test_create
+  ; "test_get_alls", `Quick, test_get_alls
+  ]
```
minishrink: The current tests look at Task calls. I think it's worth renaming this to specify that, as there may be future tests added for different submodules.

gaborigloi: I'm planning test each type of method only once: so I'd have one test for `create`, `destroy`, getter, setter, etc.. And I would test each with only one class, does not matter which one, because I expect that the same code is responsible for generating each of the above function types for all the XenAPI classes. But I'll add a comment for this too.

ocaml/tests/test_client.ml
gaborigloi: For me, the main point of this is that this is a unit test, so I can catch issues in the autogenerated code by simply running `make test` :smile: 
But yes, I think it does test that mechanism.

ocaml/tests/test_client.ml
gaborigloi: Hmm OK, what should I write here?
However, for me the main point of these strings is to locate which assertion failed within a test case, but this test, `test_create`, has only one assertion, this is why I did not give much thought to to what string I should use here.

minishrink: That's true, it just looked a bit confusing at first because the string matches one of your assertion arguments. I think anything like "check task_label in test_create" works.

gaborigloi: Which unit test this assertion belongs to will be reported by Alcotest if it fails :)

gaborigloi: This is why Alcotest is awesome

PR #3496, opened 2018-03-07
database: resurrect additional test

mseri:
I need this fix (and another in xs-opam) to be able to run `jbuilder utop ocaml/xapi` and triage issues using the xapi internals.
 

mseri:
Yes, it was just about finding that the module was included in the database library instead of being used as a test. So the only change is to build and run it as such. It's nice when you can make this kind of changes with minimal invasivity :)

PR #3497, opened 2018-03-07
Simplify nbd_info Alcotest comparator

gaborigloi:
There is no need to write a printer by hand because we get a better one
for free generated rpclib.

### DIFF
--- a/ocaml/tests/alcotest_comparators.ml
+++ b/ocaml/tests/alcotest_comparators.ml
@@ -5,27 +5,12 @@ let error_code =
   let equal aa bb = fst aa = fst bb in
   Alcotest.testable fmt equal
 
-let vdi_nbd_server_info =
-  let to_string =
-    function
-      API.{
-        vdi_nbd_server_info_exportname = e;
-        vdi_nbd_server_info_address = a;
-        vdi_nbd_server_info_port = p;
-        vdi_nbd_server_info_cert = c;
-        vdi_nbd_server_info_subject = s
-      } -> Printf.sprintf "%s %s %Ld %s %s" e a p c s
-  in
-  let fmt = Fmt.of_to_string to_string in
-  let equal a b =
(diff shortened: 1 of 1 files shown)


ocaml/tests/alcotest_comparators.ml
```diff
@@ -5,27 +5,12 @@ let error_code =
...
-    let open API in
-    (a.vdi_nbd_server_info_exportname = b.vdi_nbd_server_info_exportname) &&
-    (a.vdi_nbd_server_info_address = b.vdi_nbd_server_info_address) &&
-    (a.vdi_nbd_server_info_port = b.vdi_nbd_server_info_port) &&
-    (a.vdi_nbd_server_info_cert = b.vdi_nbd_server_info_cert) &&
-    (a.vdi_nbd_server_info_subject = b.vdi_nbd_server_info_subject)
-  in
-  Alcotest.testable fmt equal
+(** Creates a [testable] using OCaml's polymorphic equality and [Rpc.t] -> [string] conversion for formatting *)
+let of_rpc_of rpc_of =
```
lindig: I find the names confusing. The `_of` suffix is could by used for any function with a single argument: `length_of xs` but we just use `length`. I would suggest to drop it.

Otherwise I think this is a solid improvement.

gaborigloi: `_of` here isn't a suffix but says: this is a comparator `of_` an `rpc_of` function

gaborigloi: Maybe `from_rpc_of` would be a better name?

edwintorok: `from_rpc_of_t`?

gaborigloi: Thanks, I've updated the name

PR #3499, opened 2018-03-08
CP-27381: Auto-plug SR-IOV physical PIF; CP-27329: Make Network SR-IOV Sync Do Plug on Slave

minglumlu:
Two commits:
CP-27381: Auto-plug SR-IOV physical PIF 

This change aims to bring up physical SR-IOV PIF when plugging logical
SR-IOV PIF. Because we observed that if a physical PIF is unplugged, and 
SR-IOV on this physical PIF can be enabled, but the generated VFs can 
not be used by VM due to the physical PIF's interface is down.

CP-27329: Make Network SR-IOV Sync Do Plug on Slave

During reboot of a slave XenServer in a pool, XAPI logic on slave will
try to synchronize some networks from master. These networks include
bond, sriov, vlan, and tunnel. There is a difference between bond/sriov
and vlan/tunnel: the sync of bond/sriov is done on slave host, while the 
sync of vlan/tunnel is done on master host. This is because sync of
vlan/tunnel is just to create DB objects without plug. Making it done
on master could improve the efficiency of creating DB objects.

But this is not true for bond and sriov, as both of them need to be
plugged in sync, which has to be done on slave.

This change will fix a bug which would cause the plug of sriov done on
master. Instead, it should be done on slave, just like sync of bond
network.

### DIFF
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -1175,18 +1175,6 @@ let host_query_ha = call ~flags:[`Session]
       ~allowed_roles:_R_POOL_OP
       ()
 
-  let sync_network_sriovs = call ~flags:[`Session]
-      ~name:"sync_network_sriovs"
-      ~lifecycle:[Published, rel_kolkata, ""]
-      ~doc:"Synchronise network sriovs on given host with the master's network sriovs"
-      ~params:[
-        Ref _host, "host", "The host";
-      ]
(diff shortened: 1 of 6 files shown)


minglumlu:
Shame on me to forget this case.
Thanks, @robhoes 
Add a new commit for this. And another minor commit to make "sriov" in debug message as "SRIOV", which make it consistent with other debug messages in xapi_pif.ml.

PR #3500, opened 2018-03-08
CA-282006: Prevent IP reconfiguration on a live cluster network

minishrink:
CA-282006 raises the issue of host fencing when trying to reconfigure IP on a live cluster network. This PR asserts no clustering is enabled on the chosen network for calls such as `PIF.reconfigure_IP(v6)`. and `PIF.forget`, as well as adding the relevant errors to the datamodel for these calls.

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -963,6 +963,7 @@ module PIF = struct
     ~params: [ Ref _pif, "self", "Reference to the object"
              ; Bool, "value", "New value to set" ]
     ~allowed_roles:_R_POOL_OP
+    ~errs:[Api_errors.clustering_enabled_on_network]
     ()
 
   let ip_configuration_mode = Enum ("ip_configuration_mode",
@@ -982,6 +983,7 @@ module PIF = struct
               ]
       ~in_product_since:rel_miami
       ~allowed_roles:_R_POOL_OP
+      ~errs:[Api_errors.clustering_enabled_on_network]
       ()
 
   let ipv6_configuration_mode = Enum ("ipv6_configuration_mode",
@@ -1001,6 +1003,7 @@ module PIF = struct
               ]
       ~lifecycle:[Prototyped, rel_tampa, ""]
       ~allowed_roles:_R_POOL_OP
+      ~errs:[Api_errors.clustering_enabled_on_network]
       ()
 
   let primary_address_type = Enum ("primary_address_type",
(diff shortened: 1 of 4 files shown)


ocaml/xapi/xapi_pif.ml
```diff
@@ -443,8 +443,17 @@ let introduce_internal
   (* return ref of newly created pif record *)
   pif
 
+let assert_no_clustering_enabled ~__context ~network =
```
edwintorok: You moved this function twice in this PR, could you put it in the correct place in the first commit please?

ocaml/idl/datamodel.ml
```diff
@@ -963,6 +963,7 @@ module PIF = struct
     ~params: [ Ref _pif, "self", "Reference to the object"
              ; Bool, "value", "New value to set" ]
     ~allowed_roles:_R_POOL_OP
+    ~errs:[Api_errors.clustering_enabled_on_network]
```
edwintorok: Looks like we are missing `errs` declarations on all the cluster APIs too. It would be nice if these errors were checked at compile time by the type system (e.g. by using a Result type) but that is probably a huge amount of work. For now can you open an internal minor ticket to fix the API error declarations on the cluster APIs?

ocaml/xapi/xapi_pif.ml
```diff
@@ -443,8 +443,17 @@ let introduce_internal
...
+  (Db.Cluster.get_refs_where ~__context
+    ~expr:Db_filter_types.(Eq(Field "network", Literal (Ref.string_of network))))
+    |> function
+    | []   -> ()
+    | _::_ -> raise Api_errors.(Server_error (clustering_enabled_on_network, [Ref.string_of network]))
```
edwintorok: We still want to allow IP address changes when in maintenance mode though, where clustering is temporarily disabled on a host. I think this function also needs to take a host parameter, and then use `Xapi_clustering.is_clustering_disabled_on_host`. If clustering is disabled allow the ip to be changed, otherwise forbid it. 
You can test this with XenCenter by putting one host into maintenance mode, and then trying to change the IP. XenCenter should internally disable the cluster host, change IP, enable it again.
Or if you want to do it from the cli you can `xe cluster-host-disable`.

minishrink: Should the host check occur before checking for clusters?

edwintorok: Yes

ocaml/quicktest/quicktest.ml
```diff
@@ -878,6 +879,7 @@ let _ =
   finally
     (fun () ->
        (try
+          maybe_run_test "reconfigure-ip-cluster" (fun () -> Quicktest_ca282006.test s);
```
edwintorok: Could this go in the same file as the other clustering tests?

minishrink: Technically yes, but I think we tend to keep quicktests in the quicktest folder by convention? If there are any other clustering quicktests, I could add them to the same file.

edwintorok: quicktest_cluster.ml would be fine, you are right they are different from unit tests, and this is the first clustering quicktest.

PR #3502, opened 2018-03-09
Implement SR.probe_ext

gaborigloi:


ocaml/idl/datamodel.ml
```diff
@@ -1691,6 +1691,61 @@ module Data_source = struct
...
+      ~messages_default_allowed_roles:(Some []) (* No messages, so no roles allowed to use them *)
+      ~contents:
+        [ (* The uuid is not needed here and only adds inconvenience. *)
+          field ~qualifier:DynamicRO ~lifecycle ~ty:String "name_label" "Short, human-readable label for the SR.";
+          field ~qualifier:DynamicRO ~lifecycle ~ty:String "name_description" "Longer, human-readable description of the SR. Descriptions are generally only displayed by clients whe
+ the user is examining SRs in detail.";
+          field ~qualifier:DynamicRO ~lifecycle ~ty:Int "free_space" "Number of bytes free on the backing storage (in bytes)";
+          field ~qualifier:DynamicRO ~lifecycle ~ty:Int "total_space" "Total physical size of the backing storage (in bytes)";
+          field ~qualifier:DynamicRO ~lifecycle ~ty:Bool "clustered" "Indicates whether the SR uses clustered local storage.";
+          field ~qualifier:DynamicRO ~lifecycle ~ty:health "health" "The health status of the SR.";
```
gaborigloi: Here & in xcp-idl we return less info than what is provided by SMAPIv3 - maybe we could extend them?

edwintorok: which info is missing?

ocaml/xapi/xapi_sr.ml
```diff
@@ -188,6 +188,53 @@ let probe ~__context ~host ~device_config ~_type ~sm_config =
        | Probe _ as x -> Xmlrpc.to_string (rpc_of_probe_result x)
     )
 
+let probe_ext ~__context ~host ~device_config ~_type ~sm_config =
```
gaborigloi: Maybe we should make probe call probe_ext and transform it into the old format, Jon recommended something like that.

ocaml/idl/datamodel.ml
gaborigloi: UUID, `sr` URI, and datasources: https://xapi-project.github.io/xapi-storage/volume.html#a-7

gaborigloi:
I still need to add a CLI.

gaborigloi:
I think we should merge this ASAP if the existing code is OK and add the improvements in a separate PR, because ppl are waiting for this. I've tested it from the Python clients and it worked as expected :) The optional field is represented as an array with 0 or 1 element in XML-RPC & Python.

gaborigloi:
I'll have to extend my relevant commit message to say that these optional types can only return types, but cannot be stored in the DB.

ocaml/idl/datamodel.ml
```diff
@@ -1795,6 +1850,16 @@ module SR = struct
       ~allowed_roles:_R_POOL_OP
       ()
 
+  let probe_ext = call
+      ~name:"probe_ext"
+      ~in_oss_since:None
+      ~in_product_since:rel_rio
```
edwintorok: I think this should be Kolkata.

mseri:
Has this also passed a BVT and a BST?

edwintorok:
Not yet, XenRT needs to be updated to the new format. This goes into `feature/REQ477/probe`, where we'll fix XenRT and run the BVT/BST.

ocaml/idl/ocaml_backend/genOCaml.ml
```diff
@@ -172,6 +179,10 @@ let gen_of_xmlrpc api tys = block
       ["let methodResponse = From.methodResponse"];
       ["let set f (xml: XMLRPC.xmlrpc) =";
        "  From.array f xml"];
+      ["let list_to_option = function";
+       "| [] -> None";
+       "| [x] -> Some x";
+       "| _ -> failwith \"0 or 1 list elements expected for option type\""];
```
mseri: Add a `log_backtrace` and raise a proper error? Not that is super important, this is going to disappear after https://github.com/xapi-project/xen-api/pull/3490 lands

gaborigloi:
I think once Marcello's changes in https://github.com/xapi-project/xen-api/pull/3490are merged, or maybe even before that , I'll take these changes and rebase, test, and submit them for master, to keep the db-related changes together.

ocaml/idl/ocaml_backend/genOCaml.ml
gaborigloi: Indeed, good point, but I think I'll just take your changes into our branch once it's ready.

mseri:
Ok, that's a pity, as changes in the db area can often lead to unexpected system test errors. I suggest you merge this ASAP and try to get a BVT/BST as soon as possible then

mseri:
Is 
```File "ocaml/xapi/xapi_sr.ml", line 212, characters 63-71:
Error: Unbound record field complete```
due to Travis misconfiguration?

**Comment by gaborigloi** [comment]
Yes, travis is not configured for this branch :/

**Comment by robhoes** [comment]
I just noticed the change in `datamodel_types.ml`. Has this been discussed with @jonludlam?

**Comment by edwintorok** [comment]
This will also require changes in xen-api-sdk to support the option type properly, which need to be discussed with @kc284, opened an internal ticket for that.

**Comment by gaborigloi** [comment]
@robhoes I've removed an unused module that doesn't seem to be used from any repo, so hopefully no discussion is required :)

**Comment by gaborigloi** [comment]
Yes, we've discussed it.

**Comment by gaborigloi** [comment]
Wrote a little script to recursively discover all configs from a given IP (run on a XS host):

```
[root@host ~]# cat test.py 
import sys
from pprint import pprint as pp
import XenAPI
s = XenAPI.xapi_local()
s.xenapi.login_with_password('root','password')
h = s.xenapi.host.get_all()[0]
ip = sys.argv[1]

def probe(configuration, calls):
        indent = ' ' * calls * 4
        print(indent + str(configuration))
        results = s.xenapi.SR.probe_ext(h, configuration,'gfs2', {})
        for result in results:
                if 'sr' in result:
                        pp(result['sr'])
                elif result['configuration'] != configuration:
                        probe(result['configuration'], calls + 1)

probe({'provider':'iscsi','ips':ip}, 0)
[root@host ~]# python test.py <IP>
{'ips': '...', 'provider': 'iscsi'}
    {'ips': '...', 'provider': 'iscsi', 'iqns': 'iqn.2009-01.xenrt.test:iscsi10ebc1c1'}
        {'ips': '...', 'iqns': 'iqn.2009-01.xenrt.test:iscsi10ebc1c1', 'ScsiId': '1494554000000000033626161633032390000000000000000', 'provider': 'iscsi'}
{'clustered': True,
 'free_space': '51513233408',
 'health': 'healthy',
 'name_description': '',
 'name_label': 'GFS2 Storage Repository 1',
 'total_space': '53682110464',
 'uuid': 'e3ff1e0e-a615-eda7-2958-c9dc59563ae5'}
[root@host ~]#
```

Good for verifying whether probe_ext works on the host works or not.

PR #3503, opened 2018-03-12
CA-285605: Make PIF.set_disallow_unplug idempotent

minishrink:
Previously, `PIF.set_disallow_unplug` raised `clustering_enabled_on_network` even when trying to set to the same value, if cluster objects were found, causing tests to fail unnecessarily. This PR makes `PIF.set_disallow_unplug` idempotent and updates the relevant tests to check this new behaviour.

### DIFF
--- a/ocaml/xapi/xapi_pif.ml
+++ b/ocaml/xapi/xapi_pif.ml
@@ -776,7 +776,7 @@ let pif_has_clustering_enabled ~__context (self : API.ref_PIF) network =
 
 let set_disallow_unplug ~__context ~self ~value =
   let network = Db.PIF.get_network ~__context ~self in
-  if pif_has_clustering_enabled ~__context self network
+  if ((Db.PIF.get_disallow_unplug ~__context ~self) <> value)&& pif_has_clustering_enabled ~__context self network
   then raise Api_errors.(Server_error(clustering_enabled_on_network, [Ref.string_of network]))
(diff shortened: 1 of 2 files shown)


ocaml/xapi/xapi_pif.ml
```diff
@@ -776,7 +776,7 @@ let pif_has_clustering_enabled ~__context (self : API.ref_PIF) network =
 
 let set_disallow_unplug ~__context ~self ~value =
   let network = Db.PIF.get_network ~__context ~self in
-  if pif_has_clustering_enabled ~__context self network
+  if ((Db.PIF.get_disallow_unplug ~__context ~self) <> value)&& pif_has_clustering_enabled ~__context self network
```
edwintorok: Could you add a space before the `&&`?

ocaml/xapi/xapi_pif.ml
```diff
@@ -776,7 +776,7 @@ let pif_has_clustering_enabled ~__context (self : API.ref_PIF) network =
...
   then raise Api_errors.(Server_error(clustering_enabled_on_network, [Ref.string_of network]))
   else Db.PIF.set_disallow_unplug ~__context ~self ~value
```
edwintorok: You could skip this if you found it to be equal, but its fine either way.

minishrink: Yeah, that's a better way to do it.

PR #3504, opened 2018-03-12
Port unit tests to Alcotest

minishrink:
Spent a lot of time on trains lately, so ported the following unit tests from OUnit to the Alcotest framework:

- test_vm_migrate
- test_ca91480
- test_no_migrate
- test_gpu_group
- test_pool_update
- test_pusb
- test_pool_apply_edition
- test_pvs_site
- test_pvs_proxy

### DIFF
--- a/ocaml/tests/test_gpu_group.ml
+++ b/ocaml/tests/test_gpu_group.ml
@@ -12,8 +12,9 @@
  * GNU Lesser General Public License for more details.
  *)
 
-open OUnit
 open Test_vgpu_common
+
+let assert_true msg b = Alcotest.(check bool) msg true b
 
 let test_supported_enabled_types () =
   let __context = Test_common.make_test_database () in
@@ -49,8 +50,8 @@ let test_supported_enabled_types () =
            "GPU group does not list %s as enabled"
            vgpu_type.Xapi_vgpu_type.model_name
        in
-       assert_bool msg_supported (List.mem vgpu_type_ref group_supported_types);
-       assert_bool msg_enabled (List.mem vgpu_type_ref group_enabled_types))
+       assert_true msg_supported (List.mem vgpu_type_ref group_supported_types);
+       assert_true msg_enabled (List.mem vgpu_type_ref group_enabled_types))
     vgpu_types_and_refs;
   (* Invalidate the PGPU's host ref, and run a GC pass; this should destroy the
      	 * pgpu, and clear the group's supported and enabled types. *)
@@ -62,18 +63,16 @@ let test_supported_enabled_types () =
   let group_enabled_types =
     Db.GPU_group.get_enabled_VGPU_types ~__context ~self:gPU_group
   in
-  assert_equal
-    ~msg:"PGPU has not been destroyed"
-    (Db.is_valid_ref __context pgpu) false;
-  assert_equal
-    ~msg:"GPU group still has supported types after GC"
-    group_supported_types [];
-  assert_equal
-    ~msg:"GPU group still has enabled types after GC"
-    group_enabled_types []
+  Alcotest.(check bool)
+    "PGPU has not been destroyed"
+    false (Db.is_valid_ref __context pgpu);
+  Alcotest.(check (list ((Alcotest_comparators.ref ()))))
+    "GPU group still has supported types after GC"
+    [] group_supported_types;
+  Alcotest.(check (list ((Alcotest_comparators.ref ()))))
+    "GPU group still has enabled types after GC"
+    [] group_enabled_types
 
 let test =
-  "test_gpu_group" >:::
-  [
-    "test_supported_enabled_types" >:: test_supported_enabled_types;
+  [ "test_supported_enabled_types", `Quick, test_supported_enabled_types
   ]
--- a/ocaml/tests/test_no_migrate.ml
+++ b/ocaml/tests/test_no_migrate.ml
@@ -12,11 +12,6 @@
  * GNU Lesser General Public License for more details.
  *)
 
-open OUnit
-open Test_common
-
-module LC = Xapi_vm_lifecycle
-
 let ops =
   [ `suspend
   ; `checkpoint
@@ -32,7 +27,7 @@ let op_string = function
   | _             -> "other"
 
 let testcases =
-  (*nest , nomig, force, permitted *)
+  (* nest , nomig, force, permitted *)
   [ false, false, false, true
   ; false, false, true , true
   ; false, true , false, false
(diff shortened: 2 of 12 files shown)


ocaml/tests/test_gpu_group.ml
```diff
@@ -62,18 +67,16 @@ let test_supported_enabled_types () =
   let group_enabled_types =
     Db.GPU_group.get_enabled_VGPU_types ~__context ~self:gPU_group
   in
-  assert_equal
-    ~msg:"PGPU has not been destroyed"
+  Alcotest.(check bool)
+    "PGPU has not been destroyed"
     (Db.is_valid_ref __context pgpu) false;
```
gaborigloi: We'll get better error messages if we put the expected value first:
```
─( 14:22:09 )─< command 0 >───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # #require "alcotest";;
─( 14:22:09 )─< command 1 >───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # Alcotest.(check int) 4 5;;
Error: This expression has type int but an expression was expected of type
         string
─( 14:22:12 )─< command 2 >───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # Alcotest.(check int) "asd" 4 5;;
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
ASSERT asd
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Exception: Failure "Error asd: expecting\n4, got\n5.".
─( 14:22:22 )─< command 3 >───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # 
```

mseri:
Looks good to me. I'd like somebody more acquainted with alcotest to also have a look

ocaml/tests/test_gpu_group.ml
```diff
@@ -12,9 +12,14 @@
  * GNU Lesser General Public License for more details.
  *)
 
-open OUnit
 open Test_vgpu_common
 
+let assert_true msg b = Alcotest.(check bool) msg true b
+let check_ref_map msg x y =
```
gaborigloi: I'd change this to `assert_map_empty`, because it's important that we're comparing to the empty list below, therefore the order doesn't matter and we can use `list` instead of `slist`

ocaml/tests/test_pool_update.ml
```diff
@@ -12,17 +12,18 @@
...
-open Stdext.Unixext
 
 let test_pool_update_destroy () =
   let __context = make_test_database () in
   let self = make_pool_update ~__context () in
   Xapi_pool_update.destroy ~__context ~self;
-  assert_equal (Db.is_valid_ref __context self) false
+  Alcotest.(check bool)
+    "test_pool_update_destroy: pool update ref should be invalid"
+    (Db.is_valid_ref __context self) false
```
gaborigloi: Same here: the expected value is the first

ocaml/tests/test_pvs_proxy.ml
```diff
@@ -12,88 +12,113 @@
...
-  let vIF = make_vif ~__context ~device:"0" () in
+  let __context = T.make_test_database () in
+  let site = T.make_pvs_site ~__context () in
+  let vIF = T.make_vif ~__context ~device:"0" () in
   let pvs_proxy = Xapi_pvs_proxy.create ~__context ~site ~vIF in
   Xapi_pvs_proxy.destroy ~__context ~self:pvs_proxy;
-  assert_equal (Db.is_valid_ref __context pvs_proxy) false
+  Alcotest.(check bool)
+    "test_destroy: PVS proxy ref should no longer be valid"
+    (Db.is_valid_ref __context pvs_proxy) false
```
gaborigloi: Same as above: `false` should be first - but this was an existing issue.

gaborigloi: Also true for all teh comparisons below.

ocaml/tests/test_no_migrate.ml
```diff
@@ -47,30 +42,30 @@ let testcases =
...
     |> function
-    | None        when permitted     -> assert_bool "success" true
-    | None                           -> assert_failure (Printf.sprintf "nv=%b nm=%b force=%b permitted=%b op=%s" nv nm force permitted (op_string op))
-    | Some (x,xs) when not permitted -> assert_bool "success" true
-    | Some (x,xs)                    -> assert_failure (Printf.sprintf "nv=%b nm=%b force=%b permitted=%b op=%s error was=%s" nv nm force permitted (op_string op) x)
+    | None        when permitted     -> Alcotest.(check pass) "success" () ()
+    | None                           -> Alcotest.fail (Printf.sprintf "nv=%b nm=%b force=%b permitted=%b op=%s" nv nm force permitted (op_string op))
+    | Some (x,xs) when not permitted -> Alcotest.(check pass) "success" () ()
+    | Some (x,xs)                    -> Alcotest.fail (Printf.sprintf "nv=%b nm=%b force=%b permitted=%b op=%s error was=%s" nv nm force permitted (op_string op) x)
   )
```
minishrink: `Alcotest.(check pass)` could just be a unit

ocaml/tests/test_vm_migrate.ml
```diff
@@ -12,57 +12,62 @@
...
-open Test_common
-
 let mac1 = "00:00:00:00:00:01"
 let mac2 = "00:00:00:00:00:02"
 
+let check_network_map =
+  Alcotest.(check (list
+                  (pair
+                    (Alcotest_comparators.ref ())
+                    (Alcotest_comparators.ref ()))))
```
gaborigloi: Since we're not using `slist`, this will break when the two lists have the same elements but in different order

ocaml/tests/test_vm_migrate.ml
```diff
@@ -12,57 +12,62 @@
...
+  let vm_vif1 = Test_common.make_vif ~__context ~mAC:mac1 () in
+  let vm_vif2 = Test_common.make_vif ~__context ~mAC:mac2 () in
+  let snap_vif1 = Test_common.make_vif ~__context ~mAC:mac1 () in
+  let snap_vif2 = Test_common.make_vif ~__context ~mAC:mac2 () in
   (* In reality this network won't be in the local database, but for our
      	 * purposes it is a meaningless UUID so it's OK for it to be in the local
      	 * database. *)
-  let network1 = make_network ~__context () in
+  let network1 = Test_common.make_network ~__context () in
   (* Check that a map with a single VIF -> network pair is unchanged. *)
```
gaborigloi: I think we can remove these comments now as they are part of the code :) (To avoid duplicating them)

PR #3505, opened 2018-03-12
CA-285349: replace -only-sr quicktest flag with -default-sr

minishrink:
As a result of the recent REQ-477 PR review, it was suggested to make the following switch:

- `-only-sr [name-label]`: former flag, only ran quicktests on SRs with a matching `name-label` field
- `-default-sr`: replacement flag, only runs quicktests on the pool's default SR

This flag is passed to all relevant quicktests which iterate through SRs. Currently this only includes the CBT and storage quicktests.

### DIFF
--- a/ocaml/quicktest/quicktest.ml
+++ b/ocaml/quicktest/quicktest.ml
@@ -828,7 +828,6 @@ let vm_powercycle_test s vm =
   success test
 
 
-
 let _ =
   let all_tests = [
     "storage";
@@ -857,7 +856,7 @@ let _ =
     "-iso-sr-path", Arg.String (fun x -> Quicktest_storage.iso_path := x), "Path to ISO SR";
     "-single", Arg.String (fun x -> tests_to_run := [ x ]), Printf.sprintf "Only run one test (possibilities are %s)" (String.concat ", " all_tests) ;
     "-all", Arg.Unit (fun () -> tests_to_run := all_tests), Printf.sprintf "Run all tests (%s)" (String.concat ", " all_tests);
-    "-only-sr", Arg.String (fun x -> Quicktest_storage.only_sr_name := Some x), "Run tests only on SR with specified name";
+    "-default-sr", Arg.Unit (fun () -> Quicktest_storage.use_default_sr := true), "Only run SR tests on the pool's default SR";
     "-nocolour", Arg.Clear Quicktest_common.use_colour, "Don't use colour in the output" ]
     (fun x -> match !host, !username, !password with
        | "", _, _ -> host := x; rpc := rpc_remote; using_unix_domain_socket := false;
--- a/ocaml/quicktest/quicktest_storage.ml
+++ b/ocaml/quicktest/quicktest_storage.ml
@@ -31,22 +31,23 @@ let sr_probe     = "SR_PROBE"
 let sr_update    = "SR_UPDATE"
 
 let iso_path = ref "/opt/xensource/packages/iso"
-
-let only_sr_name = ref None
+let use_default_sr = ref false
 
 (** Return a list of all SRs which have at least one plugged-in PBD ie those
     which we can use for stuff *)
 let list_srs session_id =
-  let all = Client.SR.get_all !rpc session_id in
+  let all =
+    if !use_default_sr
+    then begin
+      print_endline "  Running tests on default SR only";
+      [ (Quicktest_common.get_default_sr session_id) ]
+    end
+    else (Client.SR.get_all !rpc session_id)
+  in
   List.filter (fun sr ->
-      let pbds = Client.SR.get_PBDs !rpc session_id sr in
-      List.fold_left (||) false
-        (List.map (fun pbd -> Client.PBD.get_currently_attached !rpc session_id pbd) pbds)) all
-  (* Filter SR with specific type from CLI *)
-  |> List.filter (fun sr ->
-         match !only_sr_name with
-         | None -> true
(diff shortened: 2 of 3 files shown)


ocaml/quicktest/quicktest.ml
```diff
@@ -857,7 +859,7 @@ let _ =
     "-iso-sr-path", Arg.String (fun x -> Quicktest_storage.iso_path := x), "Path to ISO SR";
     "-single", Arg.String (fun x -> tests_to_run := [ x ]), Printf.sprintf "Only run one test (possibilities are %s)" (String.concat ", " all_tests) ;
     "-all", Arg.Unit (fun () -> tests_to_run := all_tests), Printf.sprintf "Run all tests (%s)" (String.concat ", " all_tests);
-    "-only-sr", Arg.String (fun x -> Quicktest_storage.only_sr_name := Some x), "Run tests only on SR with specified name";
+    "-default-sr", Arg.Unit (fun () -> use_default_sr := true), "Only run SR tests on the pool's default SR";
```
gaborigloi: @edwintorok do we want to actually *remove* the `-only-sr` flag, or could we have both?

gaborigloi: In my opinion it's better if we do it the same way as before, and move the `use_default_sr` ref into `Quicktest_storage`. This way it won't need this parameter, and we won't have to pass it through all the tests that do SR selection.

ocaml/quicktest/quicktest.ml
```diff
@@ -902,7 +904,7 @@ let _ =
           maybe_run_test "powercycle" (fun () -> with_vm s vm_powercycle_test);
           maybe_run_test "lifecycle" (fun () -> with_vm s Quicktest_lifecycle.test);
           maybe_run_test "copy" (fun () -> Quicktest_vdi_copy.start s sr);
-          maybe_run_test "import_raw_vdi" (fun () -> Quicktest_import_raw_vdi.start s);
```
edwintorok: Why this change?

ocaml/quicktest/quicktest_cbt.ml
```diff
@@ -123,7 +123,7 @@ let vdi_data_destroy_test ~session_id ~vDI =
       ~msg:"VDI.data_destroy failed to update VDI.type";
     assert_cbt_status true ~session_id ~test ~vDI:snapshot
       ~msg:"VDI snapshot cbt_enabled field erroneously set to false";
-  (*  test_vdi_update ~session_id ~test snapshot; 
+  (*  test_vdi_update ~session_id ~test snapshot;
```
edwintorok: Did you intentionally reindent code here? If not might be better to send a "reindent XAPI" pull request to master fist by running `make reindent`, rather than reindenting files in each PR. Then we can rebase from master with `-Xignore-space-change` and continue our work without having to reindent things.
The checklist from the ticket contains a reindent item as well.

ocaml/quicktest/quicktest_storage.ml
```diff
@@ -616,12 +616,11 @@ let foreach_sr session_id sr =
...
 
-let go s =
+let go s use_default_sr =
   let test = make_test "Listing available Storage Repositories" 0 in
   start test;
-  let srs = list_srs s in
+  let srs = list_srs s use_default_sr in
   debug test (Printf.sprintf "Found %d SRs" (List.length srs));
   success test;
-  if !only_sr_name = None then
```
edwintorok: I think the reason for skipping the ISO test was that the ISO test was failing on our branch. If you tried this on your box and all tests pass, then deleting this line is correct.

minishrink: Yes, it ran without problems on my dtbox

ocaml/quicktest/quicktest_cbt.ml
minishrink: Not intentionally, I forgot to turn off my vim plugin which automatically strips whitespace.

ocaml/quicktest/quicktest.ml
minishrink: I was under the impression we want to remove it.

edwintorok: -only-sr was added as a convenience flag while testing GFS2. Its just as convenient to change the default SR in Xencenter as passing the name, so I would not want to have two mechanisms if we do not need two.

edwintorok:
Gabor is right, lets keep the global ref.

ocaml/quicktest/quicktest.ml
minishrink: Ah, this happened because when I tested on my dtbox I commented out some of the tests so I could only run the relevant ones, I must've removed the semicolon while doing that, can undo it. Although I'm not sure why we include one for the final line within a try-with anyway, is it any different with or without?

edwintorok: If it already has a semicolon you get less changes when people add stuff to the end: you only need to add a new line, not modify the old to add the ; and add new line. As discussed in another PR the preferred style is to put the ; at the beginning of the line, but lets follow what the existing code already does here.

ocaml/quicktest/quicktest_cbt.ml
minishrink: Reindentation PR (here[https://github.com/xapi-project/xen-api/pull/3506])

minishrink:
I've removed any edits to quicktest_cbt.ml and kept all `default-sr` logic in quicktest_storage.ml, only changing the ref in quicktest.ml instead of passing it around to different quicktests.

PR #3506, opened 2018-03-12
CA-285349: Reindent Xapi

minishrink:
A reindentation of the entire xapi repo, as decided in the GFS2 PR review meeting.
Note: this PR only includes .ml and .mli files, not .mll or .mly


lindig:
I don't think this is a good idea as long as other feature branches are outstanding. It will make merging them potentially much harder.

robhoes:
I agree that it is not to right time to do this. We should at least wait for the `sr-iov` branch to be merged.

minishrink:
Closing until a more suitable time when outstanding feature branches have been merged.

PR #3507, opened 2018-03-13
Eliminate fd leak in timebox

krizex:
In function `timebox`, we create a new thread to handle the real stuff, but once
the stuff hangs forever, it will result in fd of one endian of the pipe leak.
So if the stuff times out, close the fd and ignore any exception if occurs. (The
main and new thread may close the fd at the same time once the stuff finished suddenly.)

### DIFF
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -1141,8 +1141,9 @@ let timebox ~timeout ~otherwise f =
           with e ->
             result := fun () -> raise e);
          Unix.close fd_out) () in
-  let _ = Thread.wait_timed_read fd_in timeout in
+  let finished = Thread.wait_timed_read fd_in timeout in
   Unix.close fd_in;
+  if not finished then ignore_exn (fun () -> Unix.close fd_out);
   !result ()
 
 (**************************************************************************************)


krizex:
@lindig Thanks, that's clearer than my commit message.
I'd like to wait for comments from other guys so that I could rework the change together.

ocaml/xapi/helpers.ml
```diff
@@ -1141,8 +1141,9 @@ let timebox ~timeout ~otherwise f =
           with e ->
             result := fun () -> raise e);
```
robhoes: Don't we leak both `fd_in` and `fd_out` if we hit this exception?

robhoes: No, we will still close `fd_in` below, because the exception only kills the thread. But I'm unsure about the value of `finished` in this case.

mseri: No that should be fine, it's packing the exception inside the result reference, it won't trigger until somebody reads it.

robhoes: Oh yes, I misread it.

lindig:
We agree here that it is fine. Please update the commit message and force push it - we will merge it.

robhoes:
@krizex can merge it himself after updating the commit message.

krizex:
Commit message updated.

ocaml/xapi/helpers.ml
YarsinCitrix: I think this is not good.

It seems even ocaml fd is actually numeric fd like linux/unix. So theoretically there could be race condition here, if both thread wake up and tries to close the fd, while in the meantime another thread (not related) may create a fd that is actually the same value, and the latter close could result in closing the other newly created fd. 

I would suggest mutex guard the fd closing with a flag. I have updated my original code with mutex implementation. Anyway, two places need to be implemented in the same way.

krizex: Yes, that should probably a problem because in the manpage (http://man7.org/linux/man-pages/man2/open.2.html) it says:
>        The return value of open() is a file descriptor, a small, nonnegative
>        integer that is used in subsequent system calls (read(2), write(2),
>        lseek(2), fcntl(2), etc.) to refer to the open file.  The file
>        descriptor returned by a successful call will be the lowest-numbered
>        file descriptor not currently open for the process.

lindig: Could you post your code here?

YarsinCitrix: @lindig Refer to:
https://github.com/xapi-project/xcp-networkd/pull/132/files

The code looks like below:
+		let timebox ~timeout f =
+			(* Mutex is used to guard the closing of fd_out in case of racing condition 
+			 * between the two threads against double close, whereas some other threads might
+			 * create fd that is just the same (as it is integar). *)
+			let fd_mutex = Xapi_stdext_threads.Threadext.Mutex.create () in
+			let fd_in, fd_out = Unix.pipe () in
+			let fd_closed = ref false in
+			let close_fd fd = Xapi_stdext_threads.Threadext.Mutex.execute fd_mutex
+				( fun () ->
+					if not !fd_closed then begin
+						fd_closed := true;
+						Unix.close fd
+					end
+				)
+			in
+			let _ =  Thread.create
+				(fun () ->
+					f ();
+					debug "Complete execution in thread.";
+					close_fd fd_out) () 
+			in
+			if Thread.wait_timed_read fd_in timeout then
+				debug "Timed execution completion."
+			else
+				debug "Timed execution timed out.";
+			Unix.close fd_in;
+			(* close the fd_out in case the other thread stuck long and causing fd leak *)
+			close_fd fd_out
+		in

PR #3508, opened 2018-03-13
CA-285349: split `unplug_all_pbds` into two internal functions

minishrink:
As per the GFS2 PR review, this change splits `Xapi_pbd.unplug_all_pbds` into two separate internal functions; one to unplug all PBDs and one to disable clustering, allowing for better decoupling of these functions. The external `Xapi_pbds.unplug_all_pbds` interface and logic is unchanged.

### DIFF
--- a/ocaml/xapi/xapi_pbd.ml
+++ b/ocaml/xapi/xapi_pbd.ml
@@ -236,8 +236,8 @@ let get_locally_attached ~__context =
             Eq (Field "host", Literal (Ref.string_of host)),
             Eq (Field "currently_attached", Literal "true"))))
 
-(* Called on shutdown: it unplugs all the PBDs and disables the cluster host.
-   If anything fails it throws an exception *)
+(* unplug_all_pbds and disable_clustering are called on shutdown
+ * Both throw exceptions if anything fails *)
 let unplug_all_pbds ~__context =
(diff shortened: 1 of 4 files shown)


ocaml/xapi/xapi_pbd.ml
```diff
@@ -247,10 +245,18 @@ let unplug_all_pbds ~__context =
...
   | None -> info "No cluster host found"
   | Some self ->
      info "Disabling cluster host";
      Xapi_cluster_host.disable ~__context ~self
+
+(* Called on shutdown: it unplugs all the PBDs and disables the cluster host.
+   If anything fails it throws an exception *)
+let unplug_all_pbds ~__context =
+  unplug_all_pbds_internal ~__context;
+  disable_clustering ~__context
```
edwintorok: Please change all the caller of unplug_all_pbds to call these two functions instead. The reason behind splitting the function was to make it clearer what it does: a function called unplug_all_pbds also disabling clustering is somewhat unexpected.

minishrink:
Moved `disable_clustering` to Xapi_cluster_host.ml as moving to Xapi_clustering.ml created a cyclical dependency, and this way it's directly beneath the `Xapi_cluster_host.disable` function it wraps

PR #3509, opened 2018-03-13
CA-280423: Open/close xapi-clusterd port when starting/stopping daemon

thomasmck:

### DIFF
--- a/ocaml/xapi/xapi_clustering.ml
+++ b/ocaml/xapi/xapi_clustering.ml
@@ -165,9 +165,10 @@ module Daemon = struct
     | None -> ignore (Helpers.call_script script params)
 
   let service = "xapi-clusterd"
+  let port = (string_of_int !Xapi_globs.xapi_clusterd_port)
   let enable ~__context =
     debug "Enabling and starting the clustering daemon";
-    maybe_call_script ~__context !Xapi_globs.firewall_port_config_script ["open"; (string_of_int !Xapi_globs.xapi_clusterd_port)];
+    maybe_call_script ~__context !Xapi_globs.firewall_port_config_script ["open"; port];
     maybe_call_script ~__context "/usr/bin/systemctl" [ "enable"; service ];
     maybe_call_script ~__context "/usr/bin/systemctl" [ "start"; service ];
     debug "Cluster daemon: enabled & started"
@@ -176,7 +177,7 @@ module Daemon = struct
     debug "Disabling and stopping the clustering daemon";
     maybe_call_script ~__context "/usr/bin/systemctl" [ "disable"; service ];
     maybe_call_script ~__context "/usr/bin/systemctl" [ "stop"; service ];
-    maybe_call_script ~__context !Xapi_globs.firewall_port_config_script ["close"; (string_of_int !Xapi_globs.xapi_clusterd_port)];
+    maybe_call_script ~__context !Xapi_globs.firewall_port_config_script ["close"; port];
     debug "Cluster daemon: disabled & stopped"
 end
 
--- a/ocaml/xapi/xapi_clustering.ml
+++ b/ocaml/xapi/xapi_clustering.ml
@@ -165,8 +165,10 @@ module Daemon = struct
     | None -> ignore (Helpers.call_script script params)
 
   let service = "xapi-clusterd"
+  let xapi_clusterd_port = "8896"
   let enable ~__context =
     debug "Enabling and starting the clustering daemon";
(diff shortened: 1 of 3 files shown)


ocaml/xapi/xapi_clustering.ml
```diff
@@ -165,8 +165,10 @@ module Daemon = struct
     | None -> ignore (Helpers.call_script script params)
 
   let service = "xapi-clusterd"
+  let xapi_clusterd_port = "8896"
   let enable ~__context =
     debug "Enabling and starting the clustering daemon";
+    maybe_call_script ~__context "/etc/xapi.d/plugins/firewall-port" ["open"; xapi_clusterd_port];
```
robhoes: This path, and the other strings in here, should be defined in xapi_globs.ml, such that they can be set from the config file.

thomasmck: @robhoes Just to make sure I understand properly we should move the path, the port number and the "open"/"close" strings to xapi_globs.ml?

And then just call them here with `Xapi_globs.xapi_clusterd_port`?

robhoes: I think that you can leave "open"/"close", and just do the path and port. Grep for e.g. `nbd_firewall_config_script` in xapi_globs.ml to see how to refine them and register them with the config file logic.

thomasmck: Should I actually update xapi_conf to include the port? looks like a look of ports are included but commented out. Thats what I've done atm.

robhoes: That's handy yes. How about adding the other one as well?

robhoes: Btw, they are commented out, with the default value filled in, as examples in the config file. That is common practice for such files. So you've done the right thing.

thomasmck: I didn't see any of the other paths included in there but I can include the one I've added anyways. Seeing as it is commented out there is no harm.

ocaml/xapi/xapi_clustering.ml
```diff
@@ -167,6 +167,7 @@ module Daemon = struct
   let service = "xapi-clusterd"
   let enable ~__context =
     debug "Enabling and starting the clustering daemon";
+    maybe_call_script ~__context !Xapi_globs.firewall_port_config_script ["open"; (string_of_int !Xapi_globs.xapi_clusterd_port)];
```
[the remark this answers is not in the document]
thomasmck: Sounds reasonable, I will amend it.

robhoes:
Looks like a rebase went wrong on this branch: the PR now contains a bunch of unrelated commits.

thomasmck:
Removed the other commits that got pulled in and made the other suggested changes.

ocaml/xapi/xapi_clustering.ml
```diff
@@ -165,8 +165,10 @@ module Daemon = struct
     | None -> ignore (Helpers.call_script script params)
 
   let service = "xapi-clusterd"
+  let port = (string_of_int !Xapi_globs.xapi_clusterd_port)
```
lindig: Careful: the port is defined by xapi startup and any changes to !Xapi_globs.xapi_clusterd_port over the lifetime of xapi will not be taken into account. You might want to put this inside `enable()` and `disable()` to read it again at the time of usage.

robhoes: Indeed, `port` may evaluate before the config file read takes place.

scripts/xapi.conf
```diff
@@ -378,6 +378,12 @@ sm-plugins=ext nfs iscsi lvmoiscsi dummy file hba rawhba udev iso lvm lvmohba lv
 # the vbd ring buffer
 # default-vbd3-polling-idle-threshold = 50
 
+# The communication port used by xapi-clusterd
+# xapi_clusterd_port = 8896
+
+# Path for the firewall-port script
+# firewall_port_config_script = /etc/xapi.d/plugins/firewall-port
```
robhoes: `firewall-port-config`

thomasmck:
Ok I think all comments have now been addressed. Let me know if there is anything I've missed.

thomasmck:
I agree. Github has the ability to do that as part of the merge right? I do I need to do it myself and re-push?

thomasmck:
@lindig @robhoes Do I need to squash manually myself or can it be done through github when merging?

PR #3510, opened 2018-03-13
Make SR.probe backward compatible for SMAPIv3

gaborigloi:


gaborigloi:
Depends on https://github.com/xapi-project/xcp-idl/pull/212 and https://github.com/xapi-project/xapi-storage-script/pull/59 - I've successfully built everything

ocaml/xapi/xapi_sr.ml
```diff
@@ -183,12 +183,44 @@ let probe ~__context ~host ~device_config ~_type ~sm_config =
...
     )
 
-let probe_ext ~__context ~host ~device_config ~_type ~sm_config =
+let probe = call_probe ~f:(
+    function
+    | Storage_interface.Raw x -> x
+    | Storage_interface.Probe probe_results ->
+      (* Here we try to mimic the XML document structure returned in the Raw
+         case by SMAPIv1, for backwards-compatibility *)
+      let module T = Xmlm_tree in
```
edwintorok: Maybe we should add ezxmlm to xs-opam (ezjsonm  is already there) and use that instead?

ocaml/xapi/xapi_sr.ml
```diff
@@ -183,12 +183,44 @@ let probe ~__context ~host ~device_config ~_type ~sm_config =
...
+          "SR"
+          [ T.el "size" [T.data @@ Int64.to_string total_space]
+          ; T.el "name_label" [T.data name_label]
+          ; T.el "name_description" [T.data name_description]
+          ]
+      in
+      let tree = T.el "SRlist" (List.map sr_info srs) in
+      let buf = Buffer.create 20 in
+      let output = Xmlm.make_output ~nl:true (`Buffer buf) in
+      T.output_doc ~tree ~output ~dtd:None;
```
edwintorok: Ezxmlm would probably have some helpers for doing these 4 lines

ocaml/idl/datamodel.ml
```diff
@@ -1712,7 +1712,8 @@ module Sr_stat = struct
       ~doccomments:[]
       ~messages_default_allowed_roles:(Some []) (* No messages, so no roles allowed to use them *)
       ~contents:
-        [ (* The uuid is not needed here and only adds inconvenience. *)
+        [
+          field ~qualifier:DynamicRO ~lifecycle ~ty:(Option String) "uuid" "Uuid that uniquely identifies this SR, if one is available.";
```
edwintorok: This is a new field introduced by this branch, so no need for additional version specifiers. We should probably squash it when merging to feature branch though.

ocaml/xapi/xapi_sr.ml
```diff
@@ -226,13 +231,15 @@ let probe_ext =
     function Healthy -> `healthy | Recovering -> `recovering
   in
   let to_xenapi_sr_stat Storage_interface.{
+      sr_uuid;
```
edwintorok: Would this require changes sm-cli?

ocaml/idl/ocaml_backend/ocaml_utils.ml
```diff
@@ -89,7 +89,7 @@ let rec ocaml_of_ty = function
   | Map(l, r) -> "("^alias_of_ty l^" * "^alias_of_ty r^") list"
   (*  | Ref "session" -> "Uuid.cookie" *)
   | Ref ty -> "[`"^ty^"] Ref.t"
-  | Option x -> failwith "ocaml_of_ty got an option"
+  | Option x -> ocaml_of_ty x ^ " option"
```
edwintorok: This is a fixup for the previous commit introducing the option type handling in the datamodel? Could you add a note to its commit message that this should be squashed?

ocaml/xapi/xmlm_tree.ml
```diff
@@ -0,0 +1,27 @@
+(** Conveniently output an XML tree using Xmlm *)
```
edwintorok: This is fine but if we can reuse upstream code like ezxmlm I'd prefer that.

edwintorok:
The changes look good, we might want to do the ezxmlm change, and check what happens with sm-cli

ocaml/xapi/xapi_sr.ml
gaborigloi: Sounds good. However, it's hard to branch xs-opam. So we should add it to master and remove this next time we rebase.

ocaml/xapi/xapi_sr.ml
gaborigloi: It does have some helpers for this

ocaml/xapi/xapi_sr.ml
gaborigloi: No, it built :)

ocaml/xapi/xmlm_tree.ml
gaborigloi: We can reuse it after getting it into xs-opam as mentioned above. I don't think we should delay this PR because of that, unless we can just get it into xs-opam, and then release it and put it into our branch asap.

mseri:
What happens if you use `~dtd:{|<?xml version="1.0" encoding="UTF-8"?>|}`?

gaborigloi:
I guess that will work but I expected it to work the same way as `Xmlm`.

mseri:
What if you omit `~dtd`? Is it worth having a look at the code and making a PR if needed?

gaborigloi:
I've realized what's going on, Ezxmlm sets `decl` to false when calling `Xmlm.make_output` from `to_string`, so the XML declaration won't be printed.

mseri:
Sounds like a good change to send a PR for

gaborigloi:
Opened https://github.com/avsm/ezxmlm/pull/8

PR #3511, opened 2018-03-14
CA-285897: Filter out un-attached VF-backed vifs in PCI metadata

minglumlu:
The background logic is that XAPI will try to construct all information
on a VM as metadata and push this metadata to Xenopsd to manage the VM.
Particularly the logic on vifs are: for the `start` of VM, XAPI will try
to push all vifs, no matter attached or not, to Xenopsd; while for the
`reboot` of VM, it will try to push the vifs which are currently
attached only.

SR-IOV VF backed vifs should follow same logic. This means XAPI should
not push the VFs' PCIs which are backend of un-attached vifs also.

This fix adds a filter on contructing PCI metata to filter out the vifs
which are backed by VF PCIs and currently un-attached.


cheng-z:
Looks good to me. @krizex  could you please help to have a look and merge if no other concern?
Thanks

mseri:
I think @robhoes was still reciewing it

robhoes:
OK, I think that this looks right.

PR #3512, opened 2018-03-14
CA-285349: Add clustering errors, minor refactoring [REQ-477]

minishrink:
- Replace some failures in Xapi_clustering with internal errors or API errors where possible
- Introduce new API error for clustering
- Refactor clustering assertion logic


ocaml/tests/test_cluster_host.ml
```diff
@@ -112,11 +112,10 @@ let test_destroy_forbidden_when_sr_attached () =
...
   in
   let _pbd : _ API.Ref.t = Test_common.make_pbd ~__context ~host ~sR ~currently_attached:true () in
-  let msg = ("Host has attached SR whose SM requires cluster stack " ^ cluster_stack) in
   Alcotest.check_raises
-    ("Should raise (Failure " ^ msg ^ ")")
+    ("Should raise cluster_stack_in_use: [ " ^ cluster_stack ^ " ] ")
     (* TODO: update this when the exceptions are finalized *)
-    (Failure msg)
+    Api_errors.(Server_error (cluster_stack_in_use, [ cluster_stack ]))
     (fun () -> Xapi_cluster_host.destroy ~__context ~self:cluster_host)
```
minishrink: Will soon update these tests to check for specific exceptions.

ocaml/tests/test_cluster_host.ml
```diff
@@ -112,11 +112,10 @@ let test_destroy_forbidden_when_sr_attached () =
     Test_common.make_sr ~__context ~_type:sr_type ()
   in
   let _pbd : _ API.Ref.t = Test_common.make_pbd ~__context ~host ~sR ~currently_attached:true () in
-  let msg = ("Host has attached SR whose SM requires cluster stack " ^ cluster_stack) in
   Alcotest.check_raises
-    ("Should raise (Failure " ^ msg ^ ")")
+    ("Should raise cluster_stack_in_use: [ " ^ cluster_stack ^ " ] ")
     (* TODO: update this when the exceptions are finalized *)
```
gaborigloi: This TODO can be removed now :)

ocaml/xapi/xapi_clustering.ml
```diff
@@ -61,17 +61,17 @@ let ip_of_pif (ref,record) =
     }*)
 let assert_pif_prerequisites pif =
   let assert_pif_permaplugged (ref,record) =
-    if not record.API.pIF_disallow_unplug then failwith (Printf.sprintf "PIF %s allows unplug" record.API.pIF_uuid);
-    if not record.pIF_currently_attached then failwith (Printf.sprintf "PIF %s not plugged" record.API.pIF_uuid)
+    if not record.API.pIF_disallow_unplug then
+      raise Api_errors.(Server_error (internal_error, [ "PIF allows unplug"; record.API.pIF_uuid ] ));
```
gaborigloi: I think we don't need to change this, as exceptions are automatically converted to INTERNAL_ERROR, but it's best to double-check

ocaml/xapi/xapi_clustering.ml
```diff
@@ -61,17 +61,17 @@ let ip_of_pif (ref,record) =
...
   ignore(ip_of_pif pif)
 
-let handle_error error =
-  (* TODO: replace with API errors? *)
-  match error with
-  | InternalError message -> failwith ("Internal Error: " ^ message)
-  | Unix_error message -> failwith ("Unix Error: " ^ message)
+let handle_error = function
+  | InternalError message -> raise Api_errors.(Server_error (internal_error, [ message ]))
+  | Unix_error message -> raise Api_errors.(Server_error (internal_error, [ "Unix Error: " ^ message ]))
```
gaborigloi: Same as above.

ocaml/xapi/xapi_clustering.ml
```diff
@@ -146,17 +146,12 @@ let assert_cluster_host_has_no_attached_sr_which_requires_cluster_stack ~__conte
       Db.PBD.get_currently_attached ~__context ~self:pbd)
       (Db.Host.get_PBDs ~__context ~self:host) in
   let srs = List.map (fun pbd -> Db.PBD.get_SR ~__context ~self:pbd) pbds in
-  List.iter
+  if List.exists
     (fun sr ->
        let sr_sm_type = Db.SR.get_type ~__context ~self:sr in
-       (* XXX This check is a bit too conservative, because the SR requires
-          only one of these cluster stacks to be configured and running. *)
```
gaborigloi: I think we should keep this comment.

ocaml/xapi-consts/api_errors.ml
```diff
@@ -591,3 +591,4 @@ let cluster_does_not_have_one_node = "CLUSTER_DOES_NOT_HAVE_ONE_NODE"
 let no_compatible_cluster_host = "NO_COMPATIBLE_CLUSTER_HOST"
 let cluster_force_destroy_failed = "CLUSTER_FORCE_DESTROY_FAILED"
 let clustering_enabled_on_network = "CLUSTERING_ENABLED_ON_NETWORK"
+let cluster_stack_in_use = "CLUSTER_STACK_IN_USE"
```
gaborigloi: We should define this in `ocaml/idl/datamodel_errors.ml` too

ocaml/xapi/xapi_clustering.ml
minishrink: The same comment (which no longer applies) or a similar comment pointing out the requirement of having at least one cluster stack configured and running?

ocaml/xapi-consts/api_errors.ml
minishrink: Whoops, forgot that

ocaml/xapi/xapi_clustering.ml
gaborigloi: I think this comment still applies, and we should keep it exactly as it is.

gaborigloi: (This is about the semantics of `get_required_cluster_stacks` - what it means exactly if we get multiple. From an old comment I figured out that this means that we need at least one of them.)

ocaml/idl/datamodel_errors.ml
```diff
@@ -1093,6 +1093,8 @@ let _ =
     ~doc:"Force destroy failed on a Cluster_host while force destroying the cluster." ();
   error Api_errors.clustering_enabled_on_network ["network"]
     ~doc:"The network has cluster objects attached." ()
+  error Api_errors.cluster_stack_in_use ["cluster stack"]
+    ~doc:"The cluster stack is already in use." ()
```
edwintorok: maybe say that the cluster stack is already in use by a plugged SR.

ocaml/xapi/xapi_clustering.ml
minishrink: @edwintorok Thoughts?

edwintorok: There is already an api error saying that pif does not allow unplug, for symmetry we could introduce an error saying that clustering requires the pif to disallow unplug. pif_allows_unplug.

minishrink:
I've rebased my branch and addressed the suggestions from @gaborigloi and @edwintorok, any further changes?

ocaml/xapi/xapi_clustering.ml
```diff
@@ -61,17 +61,17 @@ let ip_of_pif (ref,record) =
...
   ignore(ip_of_pif pif)
 
-let handle_error error =
-  (* TODO: replace with API errors? *)
-  match error with
-  | InternalError message -> failwith ("Internal Error: " ^ message)
-  | Unix_error message -> failwith ("Unix Error: " ^ message)
+let handle_error = function
+  | InternalError message -> raise Api_errors.(Server_error (internal_error, [ message ]))
+  | Unix_error message -> raise Api_errors.(Server_error (internal_error, [ "Unix Error: " ^ message ]))
```
gaborigloi: I think we do not need to explicitly specify `internal_error`, because the previous failwiths will be converted to this automatically. So probably we shouldn't change this block of code

ocaml/idl/datamodel_errors.ml
```diff
@@ -1092,7 +1094,9 @@ let _ =
   error Api_errors.cluster_force_destroy_failed ["cluster"]
     ~doc:"Force destroy failed on a Cluster_host while force destroying the cluster." ();
   error Api_errors.clustering_enabled_on_network ["network"]
-    ~doc:"The network has cluster objects attached." ()
+  ~doc:"The network has cluster objects attached." ();
+  error Api_errors.cluster_stack_in_use ["cluster stack"]
```
gaborigloi: We should say `cluster_stack` here, because this seems to be a kind of ID, we don't have space in any of these.

ocaml/idl/datamodel_errors.ml
```diff
@@ -1092,7 +1094,9 @@ let _ =
   error Api_errors.cluster_force_destroy_failed ["cluster"]
     ~doc:"Force destroy failed on a Cluster_host while force destroying the cluster." ();
   error Api_errors.clustering_enabled_on_network ["network"]
-    ~doc:"The network has cluster objects attached." ()
+  ~doc:"The network has cluster objects attached." ();
```
edwintorok: indentation?

PR #3513, opened 2018-03-14
Fix SR.probe quicktest

gaborigloi:
The quicktests passed after I've commented out the code checking for the probe capability - we still have to add that.


gaborigloi:
I get an empty array when the contents enclose by the tag are the empty string: `<tag></tag>`

PR #3514, opened 2018-03-15
CA-280342: Bad error message when migration fails

YarsinCitrix:
As the details is already logged in xenopsd, just hide the details to user.

Currently the mentioned error is only raised in one situation
where VM is being migrated, so make it more specific and hide
some of the details that is not user friendly.

Signed-off-by: YarsinCitrix <yarsin.he@citrix.com>


ocaml/idl/datamodel_errors.ml
```diff
@@ -587,6 +587,8 @@ let _ =
     ~doc:"VM cannot be resumed because it has no suspend VDI" ();
   error Api_errors.vm_migrate_failed [ "vm"; "source"; "destination"; "msg" ]
     ~doc:"An error occurred during the migration process." ();
+  error Api_errors.vm_migrate_contact_remote_service_failed []
+    ~doc:"Failed to contact service on the destination host." ();
   error Api_errors.vm_has_too_many_snapshots [ "vm" ]
```
minishrink: It looks like `vm_migrate_failed` is intentionally vague so you can provide the specific reason as the msg parameter, why not just raise `vm_migrate_failed [ vm ; source ; destination ; "Failed to contact service on the destination host" ]`? A new and more specific error may not be required.

mseri: I agree with @minishrink, I think we should use `vm_migrate_failed` (this will avoid also a necessary update in XenCenter) and we should keep the host in the error message: `vm_migrate_failed [ Ref.string_of vm ; ??? ; ??? ; Printf.sprintf "Failed to contact service on the destination host %s" x ]`. However we lack knowledge of the source and destination field, but in that case we would need to update the error in the idl and the usage in xenopsd to provide the relevant information. @robhoes what do you think?

jonludlam: We should try to avoid putting too much english text inside the errors - it makes things hard to translate. I'd quite like to see more structure in our errors in general so that we can report errors more precisely, but that's a bigger job than this. On balance I for this PR I think I'm going to side with @YarsinCitrix in using a new error message is appropriate.

PR #3515, opened 2018-03-15
CA-284492: Ignore vGPU live-migratability in migrate_send for halted VMs

robhoes:
If a VM is not running, then we don't have to worry about being able to
live-migrate its vGPUs, so don't bother checking if the vGPU supports it.

Note that the `migrate_send` operation in use for VM.migrate_send, but also for
VDI.pool_migrate.


jonludlam:
Looks good.

robhoes:
Fixed typos :)

PR #3517, opened 2018-03-16
CA-285511: TCUSBPassBetweenVMWithinPool: No route to host

taoyongd:
This error is caused by calling unplug vusb which will call usb_reset to
remove device from cgroup allow list. The root cause is that we didn't
filter vusb when creating metadata according vusb field `currently_attached`.
So if we create vusb by CLI when vm is running and then shutdown the vm, this
error will be raised by the usb_reset.py with "No such file or directory",
as the vusb is not plugged to vm and then the file is not created.

So we need to filter the vusb if currently_attached is false, it
shouldn't be pushed to xenopsd.

### DIFF
--- a/ocaml/xapi/xapi_xenops.ml
+++ b/ocaml/xapi/xapi_xenops.ml
@@ -718,7 +718,9 @@ module MD = struct
 
   let vusbs_of_vm ~__context (vmref, vm) =
     vm.API.vM_VUSBs
-    |> List.map (fun self -> Db.VUSB.get_USB_group ~__context ~self)
+    |> List.map (fun self -> Db.VUSB.get_record ~__context ~self)
+    |> List.filter (fun self -> self.API.vUSB_currently_attached)
+    |> List.map (fun self -> self.API.vUSB_USB_group)
     |> List.map (fun usb_group -> Helpers.get_first_pusb ~__context usb_group)
     |> List.map (fun self -> Db.PUSB.get_record ~__context ~self)
     |> List.map (fun pusb -> of_vusb ~__context ~vm ~pusb)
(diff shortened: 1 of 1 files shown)


taoyongd:
Hi @lindig , Please help to review this PR. Thanks.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -718,6 +718,7 @@ module MD = struct
 
   let vusbs_of_vm ~__context (vmref, vm) =
     vm.API.vM_VUSBs
+    |> List.filter (fun self -> Db.VUSB.get_currently_attached ~__context ~self)
```
krizex: As there are two types DB query of VUSB, I think we could use `Db.VUSB.get_record` first so that we could decrease the DB query amount.

lindig:
The code looks good but please edit the commit message to not overrun the subject line into the body. In particular the dots at the end may cause problems if they are in UTF8.

taoyongd:
Ok. Thanks @lindig .

taoyongd:
Hi @robhoes ,  could you also help to have a look about this? Thanks.

ocaml/xapi/xapi_xenops.ml
taoyongd: Thanks @krizex. Good point. Updated.

taoyongd:
Hi @lindig, @robhoes, if no comments, please consider merging this PR . Thanks.

taoyongd:
Hi @lindig , @robhoes, will this PR be able to merge or any comments? Thanks

lindig:
I will merge but as I see it the commit headline is still running into the body. Please keep it shorter:

```
CA-285511: TCUSBPassBetweenVMWithinPool:  No route to host
```

taoyongd:
Updated. Thanks.

PR #3519, opened 2018-03-16
[REQ477] CA-286294: Refactor pool_resync, add best-effort host_resync

minishrink:
The current `Xapi_cluster.pool_resync` function duplicates existing logic from `Xapi_clustering.create_as_necessary`, so the following changes were made:

- use `create_as_necessary` instead of duplicating code
- move this and host-enabling logic to `Xapi_cluster_host.host_resync` function (also add to interface)
- iterate over `host_resync` in `pool_resync`
- add logging

(This has passed all unit tests and the GFS2 BVT)

### DIFF
--- a/ocaml/xapi/xapi_cluster.ml
+++ b/ocaml/xapi/xapi_cluster.ml
@@ -194,17 +194,14 @@ let pool_destroy ~__context ~self =
   Helpers.call_api_functions ~__context (fun rpc session_id ->
       Client.Client.Cluster.destroy ~rpc ~session_id ~self)
 
-let pool_resync ~__context ~self =
-  (* First ensure that the Cluster_hosts that are enabled in xapi's DB are really enabled *)
-  Db.Cluster_host.get_all ~__context |> List.iter (fun cluster_host ->
-      if Db.Cluster_host.get_enabled ~__context ~self:cluster_host then
-        Helpers.call_api_functions ~__context (fun rpc session_id ->
-            (* We rely on the fact that Cluster_host.enable unconditionally
-               invokes the low-level enable operations and is idempotent. *)
-            Client.Client.Cluster_host.enable ~rpc ~session_id ~self:cluster_host)
-    );
-  (* Then create the missing Cluster_hosts *)
-  let pool_auto_join = Db.Cluster.get_pool_auto_join ~__context ~self in
-  if pool_auto_join then begin
-      Db.Host.get_all ~__context |> List.iter (fun host -> Xapi_cluster_host.create_as_necessary ~__context ~host)
-  end
+let pool_resync ~__context ~(self : API.ref_Cluster) =
+  List.iter
+    (fun host -> log_and_ignore_exn
+        (fun () ->
+           Xapi_cluster_host.resync_host ~__context ~host;
+           if is_clustering_disabled_on_host ~__context host
+           then raise Api_errors.(Server_error (no_compatible_cluster_host, [Ref.string_of host]))
+            (* If host.clustering_enabled then resync_host should successfully
+               find or create a matching cluster_host which is also enabled *)
+        )
+    ) (Xapi_pool_helpers.get_master_slaves_list ~__context)
--- a/ocaml/xapi/xapi_cluster_host.ml
+++ b/ocaml/xapi/xapi_cluster_host.ml
@@ -63,6 +63,16 @@ let create_as_necessary ~__context ~host =
         Client.Client.Cluster_host.create rpc session_id cluster_ref host) |> ignore
   | None -> ()
 
+let resync_host ~__context ~host =
+  create_as_necessary ~__context ~host;
+  match (find_cluster_host ~__context ~host) with
+    | None              -> () (* no clusters exist *)
+    | Some cluster_host ->    (* cluster_host and cluster exist *)
+      (* Cluster_host.enable unconditionally invokes the low-level enable operations and is idempotent. *)
+      if Db.Cluster_host.get_enabled ~__context ~self:cluster_host
+      then Helpers.call_api_functions ~__context
+        (fun rpc session_id -> Client.Client.Cluster_host.enable ~rpc ~session_id ~self:cluster_host)
+
 let create ~__context ~cluster ~host =
   (* TODO: take network lock *)
   with_clustering_lock (fun () ->
--- a/ocaml/xapi/xapi_cluster_host.mli
+++ b/ocaml/xapi/xapi_cluster_host.mli
@@ -69,3 +69,9 @@ val disable_clustering : __context:Context.t -> unit
 (** [disable_clustering ~__context] is a wrapper for Xapi_cluster_host.disable
     which finds the local cluster_host [self], calls [disable ~__context self]
     and logs its actions. *)
+
+val resync_host : __context:Context.t -> host:API.ref_host -> unit
+(** [resync_host ~__context ~host] checks for any clusters on the host. If one
+    exists but is not associated with a cluster_host, it creates one. If the
+    database indicates the cluster_host is enabled, host_resync enables it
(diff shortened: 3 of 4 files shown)


ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -63,6 +63,18 @@ let create_as_necessary ~__context ~host =
         Client.Client.Cluster_host.create rpc session_id cluster_ref host) |> ignore
   | None -> ()
 
+let resync_host ~__context ~host =
+  create_as_necessary ~__context ~host;
+  match (find_cluster_host ~__context ~host) with
+    | exception (Api_errors.(Server_error(_,_)) as err)
+                        -> raise err
```
edwintorok: This is redundant: if you do not catch the exception it will escape on  its own.

ocaml/xapi/xapi_cluster_host.mli
```diff
@@ -69,3 +69,8 @@ val disable_clustering : __context:Context.t -> unit
 (** [disable_clustering ~__context] is a wrapper for Xapi_cluster_host.disable
     which finds the local cluster_host [self], calls [disable ~__context self]
     and logs its actions. *)
+
+val resync_host : __context:Context.t -> host:API.ref_host -> unit
+(** [resync_host ~__context ~host] calls [create_if_necessary ~__context ~host],
```
edwintorok: Add a comment here saying that this is a no-op if no Cluster objects exists, i.e. it is a no-op if clustering is not turned off. Someone can infer this by recursively reading the comments on the called functions, but its better to be explicit

ocaml/xapi/xapi_cluster.ml
```diff
@@ -195,16 +195,6 @@ let pool_destroy ~__context ~self =
...
-            Client.Client.Cluster_host.enable ~rpc ~session_id ~self:cluster_host)
-    );
-  (* Then create the missing Cluster_hosts *)
-  let pool_auto_join = Db.Cluster.get_pool_auto_join ~__context ~self in
-  if pool_auto_join then begin
-      Db.Host.get_all ~__context |> List.iter (fun host -> Xapi_cluster_host.create_as_necessary ~__context ~host)
-  end
+  List.iter
+    (fun host -> Xapi_cluster_host.resync_host ~__context ~host)
+    (Db.Host.get_all ~__context)
```
edwintorok: This should act on  all hosts in the same pool, not on all hosts globally. We only support one pool at the moment, so the distinction is moot, but perhaps add a comment.

ocaml/xapi/xapi_cluster.ml
```diff
@@ -196,5 +196,5 @@ let pool_destroy ~__context ~self =
 
 let pool_resync ~__context ~self =
   List.iter
-    (fun host -> Xapi_cluster_host.resync_host ~__context ~host)
+    (fun host -> log_and_ignore_exn (fun () -> Xapi_cluster_host.resync_host ~__context ~host))
```
edwintorok: This loop is good, but might be useful to check at the end whether the resync succeeded and report back an error to the user. i.e. check that each host has a matching cluster host (if clustering is enabled), and that all cluster hosts are enabled.
Otherwise this API call always succeeds even if something failed.

ocaml/xapi/xapi_pbd.ml
```diff
@@ -124,9 +124,9 @@ let resync_cluster_stack_for_sr_type ~__context ~sr_sm_type =
      multiple clusters, we may want to change this behaviour. *)
   required_clusters
   |> List.iter
-    (fun (cluster_ref, cluster_rec) ->
+    (fun (_,_) ->
        Helpers.call_api_functions ~__context (fun rpc session_id ->
-           Client.Client.Cluster.pool_resync ~rpc ~session_id ~self:cluster_ref)
+           Client.Client.Cluster.pool_resync ~rpc ~session_id)
     )
```
minishrink: I could move the cluster_stack matching logic to `host_resync`?

minishrink:
Waiting on BVT results for latest changes.

ocaml/xapi/xapi_pbd.ml
```diff
@@ -116,18 +116,22 @@ let check_sharing_constraint ~__context ~sr =
...
-           Client.Client.Cluster.pool_resync ~rpc ~session_id ~self:cluster_ref)
-    )
+  Db.Cluster.get_all_records ~__context
+  |> List.filter (fun (cluster_ref, cluster_rec) -> List.mem cluster_rec.API.cluster_cluster_stack required_cluster_stacks)
+  (* Only one matching cluster should be found *)
+  |>  match_singleton_list
+  |> fun (cluster_ref, cluster_rec) -> begin
+      Db.Cluster_host.get_records_where ~__context
+        ~expr:Db_filter_types.(Eq (Field "cluster", Literal (Ref.string_of cluster_ref)))
+      |> match_singleton_list
```
edwintorok: I think this will fail here, because you can have multiple cluster hosts for a single cluster.
It would be a lot easier if you changed this function to also take a host parameter (the caller has the host)

ocaml/xapi/xapi_pbd.ml
```diff
@@ -113,21 +113,20 @@ let check_sharing_constraint ~__context ~sr =
...
-    (fun (cluster_ref, cluster_rec) ->
-       Helpers.call_api_functions ~__context (fun rpc session_id ->
-           Client.Client.Cluster.pool_resync ~rpc ~session_id ~self:cluster_ref)
+  if List.exists
+  (* check for a cluster_host associated with the host and with a cluster with a matching cluster_stack *)
+    (fun (_, cluster_host_rec) ->
+      let self = cluster_host_rec.API.cluster_host_cluster in
+      List.mem (Db.Cluster.get_cluster_stack ~__context ~self) required_cluster_stacks
     )
+    (Db.Cluster_host.get_records_where ~__context
```
edwintorok: Could this be simplified by using Xapi_clustering.find_cluster_host?

ocaml/xapi/xapi_pbd.ml
edwintorok: Its fine here, cluster stack is more SM related, see my other comment on simplifying this code by using find_cluster_host.

ocaml/xapi/xapi_pbd.ml
minishrink: Yes, that would be smarter than what I did :P I was concerned about unnecessarily repeating code as resync_host calls this, but that's after calling `create_as_necessary` so it makes sense.

ocaml/xapi/xapi_cluster.ml
```diff
@@ -194,17 +194,14 @@ let pool_destroy ~__context ~self =
...
+  List.iter
+    (fun host -> log_and_ignore_exn
+        (fun () ->
+           Xapi_cluster_host.resync_host ~__context ~host;
+           if is_clustering_disabled_on_host ~__context host
+           then raise Api_errors.(Server_error (no_compatible_cluster_host, [Ref.string_of host]))
+            (* If host.clustering_enabled then resync_host should successfully
+               find or create a matching cluster_host which is also enabled *)
+        )
+    ) (Xapi_pool_helpers.get_master_slaves_list ~__context)
```
edwintorok: I think error reporting could be improved here, but that can be done in a separate PR that goes through each clustering API and thinks about how error handling/reporting should look like for it.
We should probably have a Xapi_cluster_host.health_check and Xapi_cluster.health_check API that does some basic consistency checks (for cluster host it would check whether clustering is enabled and call a cluster_interface API diagnostics command and check if all good there), similarly the Xapi_cluster.health_check would check the health of all cluster hosts via their individual health check API, and also check whether we have cluster hosts for all hosts if auto join is enabled. Please add an entry about this on your checklist of GFS2 cleanups.

PR #3521, opened 2018-03-19
CA-286292: VLAN PIF on SR-IOV status is incorrect after restarting toolstack

krizex:
When synchronising PIF currently_attached, should ignore VLAN PIF on SR-IOV and
SR-IOV logical PIF because they don't have real vswitch bridge, so that the
bridge existance does not indicate the connection status of the PIF.

### DIFF
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -1679,8 +1679,12 @@ let sync_pif_currently_attached ~__context ~host ~bridges =
   let networks = Db.Network.get_all_records ~__context in
   let pifs = Db.PIF.get_records_where ~__context ~expr:(
       Eq (Field "host", Literal (Ref.string_of host))
-    ) in
-
+    ) |> List.filter (fun (_, pif_rec) ->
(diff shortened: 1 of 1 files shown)


mseri:
How do we decide if they are attached or not?

krizex:
The status will be updated every time plug/unplug the PIF. Xapi restarting should not change the status of SR-IOV logical PIF and VLAN PIF on SR-IOV.

ocaml/xapi/xapi_host.ml
```diff
@@ -1679,8 +1679,12 @@ let sync_pif_currently_attached ~__context ~host ~bridges =
   let networks = Db.Network.get_all_records ~__context in
   let pifs = Db.PIF.get_records_where ~__context ~expr:(
       Eq (Field "host", Literal (Ref.string_of host))
-    ) in
-
+    ) |> List.filter (fun (pif, pif_rec) ->
```
mseri: If you write `_pif` you'll avoid a warning for the unused `pif` variable

krizex: I'd like to replace it with `_`

cheng-z:
Looks good to me also.  Merge

PR #3522, opened 2018-03-19
Merge of qemu-datapath into REQ477/master branch

gaborigloi:

### DIFF
--- a/ocaml/xapi/xapi_vbd.ml
+++ b/ocaml/xapi/xapi_vbd.ml
@@ -31,6 +31,36 @@ let update_allowed_operations ~__context ~self : unit =
 let assert_attachable ~__context ~self : unit =
   assert_attachable ~__context ~self
 
+let print_fork_error f =
+  try
+    f ()
+  with Forkhelpers.Spawn_internal_error(stderr, stdout, status) as e ->
+    begin match status with
+      | Unix.WEXITED n ->
+        error "Forkhelpers.Spawn_internal_error(%s, %s, WEXITED %d)" stderr stdout n;
+        raise e
+      | Unix.WSIGNALED n ->
+        error "Forkhelpers.Spawn_internal_error(%s, %s, WSIGNALED %d)" stderr stdout n;
+        raise e
+      | Unix.WSTOPPED n ->
+        error "Forkhelpers.Spawn_internal_error(%s, %s, WSTOPPED %d)" stderr stdout n;
+        raise e
+    end
+
+let run_command cmd args =
+  debug "running %s %s" cmd (String.concat " " args);
+  let stdout, stderr = print_fork_error (fun () -> Forkhelpers.execute_command_get_output cmd args) in
+  stdout
+
+module NbdClient = struct
+  let start_nbd_client ~unix_socket_path ~export_name =
+    run_command !Xapi_globs.nbd_client_manager_script ["connect"; "--path"; unix_socket_path; "--exportname"; export_name]
+    |> String.trim
+
+  let stop_nbd_client ~nbd_device =
+    run_command !Xapi_globs.nbd_client_manager_script ["disconnect"; "--device"; nbd_device]
+    |> ignore
+end
 
 let set_mode ~__context ~self ~value =
   let vm = Db.VBD.get_VM ~__context ~self in
@@ -47,10 +77,24 @@ let plug ~__context ~self =
     Storage_access.attach_and_activate ~__context ~vbd:self ~domid
       (fun attach_info ->
          let params = attach_info.Storage_interface.params in
-         let prefix = "/dev/" in
-         let prefix_len = String.length prefix in
-         let path = String.sub params prefix_len (String.length params - prefix_len) in
-         Db.VBD.set_device ~__context ~self ~value:path;
+         let device_path =
+           let nbd_prefix = "hack|nbd:unix:" in
+           let is_nbd = Astring.String.is_prefix ~affix:nbd_prefix params in
+           if is_nbd then begin
+             debug "Using nbd-client for VBD.plug of VBD '%s', attach_info.params='%s'" (Ref.string_of self) params;
+             let export_name = "qemu_node" in
+             let unix_socket_path = params |> Astring.String.cuts ~empty:false ~sep:nbd_prefix |> List.hd |> String.split_on_char '|' |> List.hd in
+             NbdClient.start_nbd_client ~unix_socket_path ~export_name
+           end
+           else params
+         in
+         let device_path =
+           let prefix = "/dev/" in
+           let prefix_len = String.length prefix in
+           String.sub device_path prefix_len (String.length device_path - prefix_len)
+         in
+         debug "device path: %s" device_path;
+         Db.VBD.set_device ~__context ~self ~value:device_path;
          Db.VBD.set_currently_attached ~__context ~self ~value:true;
       )
   end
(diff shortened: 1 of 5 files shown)


ocaml/xapi/xapi_vbd.ml
```diff
@@ -47,10 +77,23 @@ let plug ~__context ~self =
     Storage_access.attach_and_activate ~__context ~vbd:self ~domid
       (fun attach_info ->
          let params = attach_info.Storage_interface.params in
-         let prefix = "/dev/" in
-         let prefix_len = String.length prefix in
-         let path = String.sub params prefix_len (String.length params - prefix_len) in
-         Db.VBD.set_device ~__context ~self ~value:path;
+         let device_path =
+           let nbd_prefix = "hack|nbd:unix:" in
```
mseri: :disappointed:

ocaml/xapi/xapi_vbd.ml
```diff
@@ -47,10 +77,23 @@ let plug ~__context ~self =
...
+           let is_nbd = String.startswith nbd_prefix params in
+           if is_nbd then begin
+             let export_name = "qemu_node" in
+             let unix_socket_path = params |> Astring.String.cuts ~empty:false ~sep:nbd_prefix |> List.hd |> String.split_on_char '|' |> List.hd in
```
mseri: This will raise an horrible looking and hard to debug error if there are empty lists on the way

gaborigloi: Yes, but it's isolated

scripts/nbd_client_manager.py
```diff
@@ -0,0 +1,184 @@
...
+        self._path = path
+        self._lock_file = None
+
+    def lock(self):
+        """Acquire the lock"""
+        flags = fcntl.LOCK_EX
+        self._lock_file = open(self._path, 'w+')
+        fcntl.flock(self._lock_file, flags)
+
+    def unlock(self):
```
mseri: if you add `__enter__` and `__exit__` as alias of these you can use `with FILE_LOCK: ` below

ocaml/xapi/xapi_vbd.ml
gaborigloi: We will come up with a proper design, but for now, the strategy is to make sure the new code is isolated, and this `"hack|..."` method is hopefully good enough for that.

ocaml/xapi/xapi_vbd.ml
gaborigloi: Maybe I could add a debug message above this with the content of `params`, so that we now what we got in case one of the `List.hd`s fail?

scripts/nbd_client_manager.py
gaborigloi: Thanks, that's a great idea, I'll add those :)

scripts/nbd_client_manager.py
```diff
@@ -0,0 +1,184 @@
...
+
+import argparse
+import logging
+import os
+import subprocess
+import time
+import fcntl
+
+
+LOGGER = logging.getLogger("nbd_client_manager")
```
mseri: I think this logs at `warning` level by default

gaborigloi: I think so, good spot, I'll increase the logging level to `debug`.

ocaml/xapi/xapi_vbd.ml
mseri: I think adding a preliminary debug with the params could be enough.
The only thing with List.hd is probably to define your own `hd_log` as something like
```ocaml
let hd_log logline = function
| h :: _ -> h
| [] -> debug logline; raise Not_found (* or whatever else List.hd raises *)
```

gaborigloi: Thanks, I've added a debug line.

scripts/nbd_client_manager.py
```diff
@@ -32,19 +32,26 @@ def __init__(self, path):
...
+    def _unlock(self):
         """Unlock and remove the lock file"""
         if self._lock_file:
             fcntl.flock(self._lock_file, fcntl.LOCK_UN)
             self._lock_file.close()
             self._lock_file = None
 
+    def __enter__(self):
+        self._lock()
+        return self
```
mseri: You can remove the `return` if you don't need to do anything to the lock, that is optional anyway

gaborigloi: Thanks, done

gaborigloi:
Should I also add this `nbd_client_manager.py` to `xapi_globs` as a nonessential executable?
`xcp-idl` does close to nothing for nonessential resources btw (only checks for duplicate keys), so it's essentially a no-op, the only advantage is that we declare the path to this script in the traditional place.

mseri:
I don't know. Ping @jonludlam

scripts/nbd_client_manager.py
mseri: :+1: There is a `LOGGER.setLogLevel()` call for that

gaborigloi: I think we need to do more configuration to redirect log messages into syslog using `SysLogHandler`.

gaborigloi:
Now logging should work fine:
```
[root@host ~]# python nbd_client_manager.py connect --path asdf --exportname fjdksl     
  File "nbd_client_manager.py", line 199, in <module>                                   
    args.func(args)   
  File "nbd_client_manager.py", line 154, in _connect_cli                               
    device = connect_nbd(path=args.path, exportname=args.exportname)                    
  File "nbd_client_manager.py", line 133, in connect_nbd                                
    nbd_device = _find_unused_nbd_device()  
  File "nbd_client_manager.py", line 116, in _find_unused_nbd_device                    
    if not _is_nbd_device_connected(nbd_device=nbd_device):                             
  File "nbd_client_manager.py", line 101, in _is_nbd_device_connected                   
    returncode = _call(cmd, error=False)    
  File "nbd_client_manager.py", line 71, in _call                                       
    close_fds=True    
  File "/usr/lib64/python2.7/subprocess.py", line 711, in __init__                      
    errread, errwrite)                      
  File "/usr/lib64/python2.7/subprocess.py", line 1327, in _execute_child               
    raise child_exception                   
OSError: [Errno 2] No such file or directory                                            
[root@host ~]# 
```

```
Mar 19 14:01:47 host Running cmd ['modprobe', 'nbd']                                    
Mar 19 14:01:47 host Running cmd ['nbd-client', '-check', '/dev/nbd0']                  
Mar 19 14:01:47 host [Errno 2] No such file or directory#012Traceback (most recent call last):#012  File "nbd_client_manager.py", line 199, in <module>#012    args.func(args)#012  File "nbd_client_manager.py", line 154, in _connect_cli#012    device = connect_nbd(path=args.path, exportname=args.exportname)#012  File "nbd_client_manager.py", line 133, in connect_nbd#012    nbd_device = _find_unused_nbd_device()#012  File "nbd_client_manager.py", line 116, in _find_unused_nbd_device#012    if not _is_nbd_device_connected(nbd_device=nbd_device):#012  File "nbd_client_manager.py", line 101, in _is_nbd_device_connected#012    returncode = _call(cmd, error=False)#012  File "nbd_client_manager.py", line 71, in _call#012    close_fds=True#012  File "/usr/lib64/python2.7/subprocess.py", line 711, in __init__#012    errread, errwrite)#012  File "/usr/lib64/python2.7/subprocess.py", line 1327, in _execute_child#012    raise child_exception#012OSError: [Errno 2] No such file or directory
```

gaborigloi:
I've decided to add it to `Xapi_globs`.

gaborigloi:
Thanks for the review, I think I've addressed all the comments. I'll squash the fixes and merge if this is ok now.

ocaml/xapi/xapi_vbd.ml
```diff
@@ -47,10 +77,24 @@ let plug ~__context ~self =
     Storage_access.attach_and_activate ~__context ~vbd:self ~domid
       (fun attach_info ->
          let params = attach_info.Storage_interface.params in
-         let prefix = "/dev/" in
-         let prefix_len = String.length prefix in
-         let path = String.sub params prefix_len (String.length params - prefix_len) in
-         Db.VBD.set_device ~__context ~self ~value:path;
+         let device_path =
+           let nbd_prefix = "hack|nbd:unix:" in
+           let is_nbd = String.startswith nbd_prefix params in
```
gaborigloi: I should use `Astring` here too :/

mseri: As you wish. Xapi is still not ported to Astring at all, feel free to use `Xstringext` here

gaborigloi: I've done it now

PR #3525, opened 2018-03-20
CP-27452: return uid/gid of device file

TalonsLee:
return the uid/gid of device file when attaching device, these ids
should be restored later.

### DIFF
--- a/scripts/usb_reset.py
+++ b/scripts/usb_reset.py
@@ -21,17 +21,18 @@
 # 1. reset device
 # if without -r, do step 2~4
 # 2. if it's the first USB device to pass-through
-#      a) bind mount /dev /sys in chroot directory (/var/xen/qemu/root-<domid))
+#      a) bind mount /dev /sys in chroot directory (/var/xen/qemu/root-<domid>)
 #      b) create new cgroup devices:/qemu-<domid>,
 #      c) blacklist all and add default device whitelist,
 #      d) join current qemu process to this cgroup
-# 3. set device file uid/gid to (qemu_base + dom-id)
-# 4. add current device to whitelist
+# 3. save device uid/gid to /var/run/nonpersistent/usb/<device>
+# 4. set device file uid/gid to (qemu_base + dom-id)
+# 5. add current device to whitelist
 #
 # detach
-# ./usb_reset.py detach device -d dom-id -i uid gid
-# ./usb_reset.py detach 2-2 -d 12 -i 0 0
-# 1. restore device file owner to uid/gid
+# ./usb_reset.py detach device -d dom-id
+# ./usb_reset.py detach 2-2 -d 12
+# 1. restore device file uid/gid from /var/run/nonpersistent/usb/<device>
 # 2. remove current device from whitelist
 #
 # cleanup
@@ -72,8 +73,6 @@ def parse_arg():
     detach.add_argument("device", help="the target usb device")
     detach.add_argument("-d", dest="domid", type=int, required=True,
                         help="specify the domid of the VM")
-    detach.add_argument("-i", dest="id", type=int, nargs=2, required=True,
-                        help="restore ownership to uid gid")
 
     cleanup = subparsers.add_parser("cleanup", help="clean up chroot directory")
     cleanup.add_argument("-d", dest="domid", type=int, required=True,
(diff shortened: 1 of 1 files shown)


scripts/usb_reset.py
```diff
@@ -90,6 +89,48 @@ def get_cg_dir(domid):
     return "/sys/fs/cgroup/devices/qemu-{}".format(domid)
 
 
+def get_ids_path(device):
+    usb_dir = "/var/run/nonpersistent/usb"
+    if not os.path.isdir(usb_dir):
+        os.makedirs(usb_dir)
+    return usb_dir + "/{}".format(device)
```
rosslagerwall: You can use os.path.join for this:

    return os.path.join(usb_dir, device)

scripts/usb_reset.py
```diff
@@ -90,6 +89,48 @@ def get_cg_dir(domid):
...
+        ids_info = "{} {}".format(stat.st_uid, stat.st_gid)
+    except OSError as e:
+        log.error("Failed to get stat of file {}: {}".format(path, str(e)))
+        exit(1)
+
+    try:
+        with open(get_ids_path(device), "w") as f:
+            f.write(ids_info)
+    except IOError as e:
+        log.error("Failed to save of file ids {}: {}".format(path, str(e)))
```
rosslagerwall: There's a stray `of`.

scripts/usb_reset.py
```diff
@@ -90,6 +89,48 @@ def get_cg_dir(domid):
...
+
+
+def save_device_ids(device):
+    path = dev_path(device)
+
+    try:
+        stat = os.stat(path)
+        ids_info = "{} {}".format(stat.st_uid, stat.st_gid)
+    except OSError as e:
+        log.error("Failed to get stat of file {}: {}".format(path, str(e)))
```
rosslagerwall: I'd suggest: `"Failed to stat {}: {}"`

scripts/usb_reset.py
```diff
@@ -90,6 +89,48 @@ def get_cg_dir(domid):
...
+    except IOError as e:
+        log.error("Failed to save of file ids {}: {}".format(path, str(e)))
+        exit(1)
+
+
+def restore_device_ids(device):
+    ids_path = get_ids_path(device)
+    try:
+        with open(ids_path) as f:
+            uid, gid = map(int, f.readline().split())
```
rosslagerwall: You can just use `read()` instead of `readline()`... since you don't actually write a newline previously.

scripts/usb_reset.py
```diff
@@ -90,6 +89,48 @@ def get_cg_dir(domid):
...
+
+    try:
+        with open(get_ids_path(device), "w") as f:
+            f.write(ids_info)
+    except IOError as e:
+        log.error("Failed to save of file ids {}: {}".format(path, str(e)))
+        exit(1)
+
+
+def restore_device_ids(device):
```
rosslagerwall: I would suggest `load_device_ids` since `restore_device_ids` sounds to me like it would actually be doing the `chown` as well.

scripts/usb_reset.py
```diff
@@ -90,6 +89,48 @@ def get_cg_dir(domid):
...
+    ids_path = get_ids_path(device)
+    try:
+        with open(ids_path) as f:
+            uid, gid = map(int, f.readline().split())
+    except (IOError, ValueError) as e:
+        log.error("Failed to restore device ids: {}".format(str(e)))
```
rosslagerwall: `restore` -> `load` as per my previous comment.

scripts/usb_reset.py
TalonsLee: done.

scripts/usb_reset.py
TalonsLee: done.

scripts/usb_reset.py
TalonsLee: done.

scripts/usb_reset.py
TalonsLee: this could be treated as the last line(EOF), according to the document it seems to be OK:
f.readline() reads a single line from the file; a newline character (\n) is left at the end of the string, and is only omitted on the last line of the file if the file doesn’t end in a newline.

scripts/usb_reset.py
TalonsLee: change "of" to "device"

scripts/usb_reset.py
rosslagerwall: Sure, it works but read() requires less processing and is shorter. It doesn't really matter either way.

scripts/usb_reset.py
TalonsLee: done.

scripts/usb_reset.py
```diff
@@ -90,6 +89,48 @@ def get_cg_dir(domid):
     return "/sys/fs/cgroup/devices/qemu-{}".format(domid)
 
 
+def get_ids_path(device):
+    usb_dir = "/var/run/nonpersistent/usb"
+    if not os.path.isdir(usb_dir):
+        os.makedirs(usb_dir)
```
mseri: This is racy. It is probably not a problem but should we do something like

```python
import errno

try:
    os.makedirs(usb_dir)
except os.error as err:
   if err.errno != errno.EEXIST:
     raise
```

or is not worth it?

rosslagerwall: Good spot, we should definitely do it this way.

TalonsLee: Done.

TalonsLee: actually I can also add this check at line 292, considering racing problem.

scripts/usb_reset.py
```diff
@@ -90,6 +89,53 @@ def get_cg_dir(domid):
     return "/sys/fs/cgroup/devices/qemu-{}".format(domid)
 
 
+def get_ids_path(device):
+    usb_dir = "/var/run/nonpersistent/usb"
+    if not os.path.isdir(usb_dir):
```
mseri: This check is no longer needed, if the folder is there it will fail with `errno.EEXIST` and will be ignored

TalonsLee: done.

PR #3526, opened 2018-03-20
CA-281178: Update allowed operations in consider_enabling_host

minglumlu:
In a pool, restarting the toolstack on a slave will go through two
stages: first shutting down, and next booting up again.  In the
shutting-down stage, Host.enabled will be set as false. Then in the
booting-up stage, XAPI will set it back to true in the 'initialising
storage' step.

The bug to be fixed in this ticket is that the allowed_operations of
the slave host is not updated correspondingly when setting 
Host.enabled back to true while booting up.

### DIFF
--- a/ocaml/xapi/xapi_host_helpers.ml
+++ b/ocaml/xapi/xapi_host_helpers.ml
@@ -224,6 +224,7 @@ let consider_enabling_host_nolock ~__context =
     if !Xapi_globs.on_system_boot then begin
       debug "Host.enabled: system has just restarted: setting localhost to enabled";
       Db.Host.set_enabled ~__context ~self:localhost ~value:true;
+      update_allowed_operations ~__context ~self:localhost;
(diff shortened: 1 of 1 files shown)


lindig:
The code looks good. I would suggest to re-word the commit message slightly:

```
In a pool, restarting the toolstack on a slave will go through two
stages: first shutting down, and next booting up again.  In the
shutting-down stage, Host.enabled will be set as false. Then in the
booting-up stage, XAPI will set it back to true in the 'initialising
storage' step.

The bug to be fixed in this ticket is that the allowed_operations of
the slave host is not updated correspondingly when setting 
Host.enabled back to true while booting up.
```

minglumlu:
Hi @lindig 
Thanks for your help. The commit message looks better now indeed.
Ming

minglumlu:
Hi @lindig 
Can you please let me know if you have any comment or concern on this PR?
Thanks!
Ming

PR #3527, opened 2018-03-22
CA-286338: Inter-host VM copy failed from source host to a slave in a pool

krizex:
In `Helpers.call_api_functions_internal`, if it is called in slave and the
session is not forwarded from master, then we may get a session without 'pool'
flag set, this behavior is inconsistent with the comments and the comments is
the correct.
So when we get the session from context, we should always make a new one if
it's not a inner pool session.


ocaml/xapi/helpers.ml
```diff
@@ -278,21 +278,14 @@ let call_api_functions_internal ~__context f =
...
-        ignore (Db.Session.get_pool ~__context ~self:session_id) ;
-        session_id
-    with _ ->
-      do_master_login ()
+        if Db.Session.get_pool ~__context ~self:session_id then Some session_id else None
+      with _ -> None
+    with
+    | Some session_id -> session_id
+    | None -> do_master_login ()
   in
```
mseri: This can be simplified with
```ocaml
match Context.get_session_id __context with
| session_id when Db.Session.get_pool ~__context ~self:session_id -> session_id
| _ -> do_master_login ()
| exception _ -> do_master_login ()
```
It may even be possible to do the following, but I have not tried it:
```ocaml
match Context.get_session_id __context with
| session_id when Db.Session.get_pool ~__context ~self:session_id -> session_id
| _ | exception _ -> do_master_login ()
```

For what concerns the logic I'd like somebody more familiar to have a look. Ping @jonludlam

krizex: I have read some code that mixed with `match` and `try` but cannot find it when I code.
Your code is what I aimed to find!

mseri: It's not the most readable thing, but usually I tend to go to the ocaml manual when I need to refresh weird new syntax :stuck_out_tongue_closed_eyes: This particular case is somewhat documented here https://caml.inria.fr/pub/docs/manual-ocaml/extn.html#sec264

krizex: Another post https://blog.janestreet.com/pattern-matching-and-exception-handling-unite/

mseri: Much nicer indeed! I should lurk more often in janestreet tech blog

krizex: Indeed `Db.Session.get_pool ~__context ~self:session_id` may raise exception and we have to catch it then `do_master_login ()`. So I will refine the code based on yours.

krizex:
Updated.

ocaml/xapi/helpers.ml
```diff
@@ -278,21 +278,15 @@ let call_api_functions_internal ~__context f =
...
-      if Pool_role.is_master() then
-        begin
-          let session_id = Context.get_session_id __context in
-          if Db.Session.get_pool ~__context ~self:session_id
-          then session_id
-          else do_master_login ()
-        end
-      else
-        let session_id = Context.get_session_id __context in
-        (* read any attr to test if session is still valid *)
```
mseri: Please keep this comment on top of `let in_pool ...`, it clarifies why the function call can fail

ocaml/xapi/helpers.ml
mseri: Well spotted

ocaml/xapi/helpers.ml
krizex: Yes, we should keep it.

jonludlam:
I think [this comment](https://github.com/xapi-project/xen-api/pull/3527/files#diff-0af27c45f2b1c2de120d2a9ae13ee5cbL269) is incorrect following your change, right?

krizex:
@jonludlam Yes, it's incorrect, we have to update the comments for the slave scenario.

krizex:
Updated the comments in the code. @jonludlam please have a look again, thanks.

ocaml/xapi/helpers.ml
```diff
@@ -267,32 +267,30 @@ let call_api_functions_internal ~__context f =
...
-        ignore (Db.Session.get_pool ~__context ~self:session_id) ;
-        session_id
-    with _ ->
-      do_master_login ()
+    let f () =
```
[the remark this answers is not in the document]
krizex: If the outer function is lengthy, I think we should avoid overwrite the function name when defining local function. However in this short function, I think it won't confuse other guys when reading the code.

PR #3530, opened 2018-03-23
CA-260638 Bonds are created on interfaces which participates in FCoE

krizex:
When creating bond, ensure slave PIF are not participating FCOE.


minli1:
I think the title is a little confused, the code intends to prevent bond creating on interfaces which are FCoE in use. The title seems need to say : `Bonds are not created on interfaces which participates in FCoE` or `Prevent bonds are created on interfaces which participates in FCoE`, also for the title of that ticket.

krizex:
The title illustrates the problem of this issue. @minli1

PR #3532, opened 2018-03-26
CP-26169: Check HA is not enabled before turning on clustering

thomasmck:


thomasmck:
As suggested by @edwintorok will move the check to xapi_pool_helpers where there are already some cluster related checks (also means we don't need to reindent everything).

ocaml/xapi/xapi_cluster.ml
```diff
@@ -26,7 +26,7 @@ let validate_params ~token_timeout ~token_timeout_coefficient =
 
 let create ~__context ~network ~cluster_stack ~pool_auto_join ~token_timeout ~token_timeout_coefficient =
   Pool_features.assert_enabled ~__context ~f:Features.Corosync;
-  (* TODO: take network lock *)
+
```
edwintorok: This is CP-24691, we haven't implemented it yet, please keep the todo.

ocaml/xapi/xapi_pool_helpers.ml
```diff
@@ -61,6 +61,7 @@ let valid_operations ~__context record _ref' =
     | [_] -> set_errors Api_errors.cluster_already_exists [] [ `cluster_create ]
     (* indicates a bug or a need to update this code (if we ever support multiple clusters in the pool *)
     | _::_ -> failwith "Multiple clusters exist in the pool"
+    | [] -> if ha_enabled then raise (Api_errors.Server_error(Api_errors.incompatible_cluster_stack_active, ["xhad"]))
```
edwintorok: You might want to get the actual cluster stack name from the pool instead of hardcoding it: `let current_stack = Db.Pool.get_ha_cluster_stack ~__context ~self:pool in`

ocaml/xapi/xapi_pool_helpers.ml
```diff
@@ -61,7 +62,8 @@ let valid_operations ~__context record _ref' =
     | [_] -> set_errors Api_errors.cluster_already_exists [] [ `cluster_create ]
     (* indicates a bug or a need to update this code (if we ever support multiple clusters in the pool *)
     | _::_ -> failwith "Multiple clusters exist in the pool"
-    | [] -> if ha_enabled then raise (Api_errors.Server_error(Api_errors.incompatible_cluster_stack_active, ["xhad"]))
+    (* cluster create cannot run if ha is already enabled *)
+    | [] -> if ha_enabled then raise (Api_errors.Server_error(Api_errors.incompatible_cluster_stack_active, [current_stack]))
```
edwintorok: Sorry, one more change, I think you need to use `set_errors .... [`cluster_create]` here, otherwise you'll always fail if HA is enabled and there is no clustering

ocaml/xapi/xapi_pool_helpers.ml
```diff
@@ -61,7 +61,7 @@ let valid_operations ~__context record _ref' =
     | [_] -> set_errors Api_errors.cluster_already_exists [] [ `cluster_create ]
     (* indicates a bug or a need to update this code (if we ever support multiple clusters in the pool *)
     | _::_ -> failwith "Multiple clusters exist in the pool"
-    | _ -> ()
```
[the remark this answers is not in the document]
edwintorok: `_` is equivalent to "anything matches". The reason the compiler complained is that you already matched every possibility and there was nothing left to match, you should always put the `_` match last:
`[_] ` matches lists with exactly one element.
`_ :: _` matches lists with at least one element (e.g. `one :: []` or `one :: [two; three]`, etc.) but since the previous one already matched  lists with one element, this left only lists of 2 or more to match
`_` matches any kind of list, but due to the previous matches the only remaining list that wasn't matched yet is the list of 0 elements, i.e. `[]`
you now try to match `[]`, but that already got matched by the previous `_`.

So in this case the `_` was equivalent to `[]`, its better to be explicit though and match on `[]` like you did.

thomasmck:
Squashed. Let me know if I should change anything else.

mseri:
Can you make the PR title more explicative?

PR #3533, opened 2018-03-26
CA-285596: Suppress host.set_iscsi_iqn during RPU

jonludlam:
When the pool is in RPU mode we no longer try to call `host.set_iscsi_iqn`
or `host.set_multipathing` if the database is inconsistent. When the pool
exits RPU mode we then run through the consistency check for all hosts and
do any calls that are necessary.

### DIFF
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -437,13 +437,16 @@ type boot_method =
 
 (** Returns the current value of the pool configuration flag *)
 (** that indicates whether a rolling upgrade is in progress. *)
+let rolling_upgrade_in_progress_of_oc oc =
+  List.mem_assoc Xapi_globs.rolling_upgrade_in_progress oc
+
 (* Note: the reason it's OK to trap exceptions and return false is that -- an exn will only happen if the pool record
    is not present in the database; that only happens on firstboot (when you're a master with no db and you're creating
    the db for the first time). In that context you cannot be in rolling upgrade mode *)
 let rolling_upgrade_in_progress ~__context =
   try
     let pool = get_pool ~__context in
-    List.mem_assoc Xapi_globs.rolling_upgrade_in_progress (Db.Pool.get_other_config ~__context ~self:pool)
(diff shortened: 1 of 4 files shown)


ocaml/xapi/helpers.ml
```diff
@@ -440,10 +440,13 @@ type boot_method =
 (* Note: the reason it's OK to trap exceptions and return false is that -- an exn will only happen if the pool record
```
mseri: You should probably move this comment below, where you define `rolling_upgrade_in_progress`

ocaml/xapi/xapi_host_helpers.ml
```diff
@@ -308,50 +308,67 @@ module Configuration = struct
...
+                Client.Client.Host.set_multipathing rpc session_id host_ref multipathing
+              | _ -> ()
+            end
+          in
+          let event_recs = List.map Event_helper.record_of_event events.Event_types.events in
+          let in_rpu = List.fold_left (fun in_rpu ev ->
+              match ev with
+              | Event_helper.Pool (_pool_ref, Some pool_rec) ->
+                let in_rpu = Helpers.rolling_upgrade_in_progress_of_oc pool_rec.API.pool_other_config in
+                if (not in_rpu) && was_in_rpu then begin
```
edwintorok: I am a bit confused on how `was_in_rpu` is set. RPU involves restarting XAPI, so if you got a 1 host pool will RPU be reported as in progress on startup, or will it immediately finish because it was the only host to be upgraded?

jonludlam: https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/db_gc.ml#L119-L168

jonludlam: This code doesn't care if it misses a transient RPU mode since the code it's invoking is entirely level-based rather than edge-based. All we're trying to do is make sure that if we do any suppression at all then we ensure at the end we don't miss anything by updating all hosts.

ocaml/xapi/helpers.ml
jonludlam: Ah, good catch thanks!

PR #3535, opened 2018-03-27
Review probe API changes for feature/REQ477/master

edwintorok:
This is a reduced set of changes, without the Option type that depends on xen-api-sdk.
The SR UUID change is required to not break the build, the rest transform the SMAPIv3 probe response into the legacy SMAPIv1 probe response, but only when the probe is complete.
When probe returns an incomplete result that requires further probing, we do not transform it into an exception, this was discussed with @jonludlam.
To get more information for SMAPIv3 SRs the upcoming `probe_ext` API should be used, that will be part of a different PR because it also depends on xen-api-sdk changes.

Drop the `Probe` variant from SR.probe, it will only be used by SR.probe_ext, due to the backwards compatible transform the CLI will report the SMAPIv1 compatible raw XML.


ocaml/xapi/xapi_sr.ml
```diff
@@ -183,11 +183,43 @@ let probe ~__context ~host ~device_config ~_type ~sm_config =
...
-       | Probe _ as x -> Xmlrpc.to_string (rpc_of_probe_result x)
+       Client.SR.probe ~dbg ~queue ~device_config ~sm_config |> f
     )
 
+let probe = call_probe ~f:(
+    function
+    | Storage_interface.Raw x -> x
+    | Storage_interface.Probe probe_results ->
+      (* Here we try to mimic the XML document structure returned in the Raw
+         case by SMAPIv1, for backwards-compatibility *)
```
mseri: Wouldn't it be better to add an optional version to the probe call, and when called with version say `>= 2` do not perform any conversion, so that new code can actually benefit from a structured safe response?

ocaml/xapi/xmlm_tree.ml
```diff
@@ -0,0 +1,27 @@
+(** Conveniently output an XML tree using Xmlm *)
```
mseri: Is this really needed or it is something that we can embed in xapi_sr, next to the thing that calls it?

edwintorok: We've discussed that we could use ezxmlm for this, but it is not in xs-opam yet. Once it is we can delete this file mostly.

gaborigloi: Maybe we'll use this from other files too and add more xmlm-related helpers, so this way it's more discoverable.

mseri: I think that manual parsing and unparsing of xml is a really bad idea, especially when we have the rpc that does it properly. It took a huge amount of time of cleaning up the use of xmlm from the legacy api, I am not kin on reintroducing it. I am really worried about introducing this new module to be honest.

ocaml/xapi/xapi_sr.ml
gaborigloi: Hmm, not sure, Jon said we should have a new `SR.probe_ext` call, that is the one that returns the structured data, in a way that is reflected by ocaml's type system too.

ocaml/xapi/xmlm_tree.ml
gaborigloi: The issue is that here we need to mimic exactly the legacy output format of SR.probe, which is like this:
```
<?xml version="1.0" ?>
<SRlist>
	<SR>
		<UUID>db208f01-736a-f62d-f341-49521629fb3d</UUID>
		<Devlist>/dev/sda3</Devlist>
		<size>954539704320</size>
	</SR>
</SRlist>
```
Therefore we need more control over the XML output, this is why we need xmlm, and cannot use ocaml-rpc.
I agree that for future uses, we should just use ocaml-rpc.

edwintorok: There are 2 things using the Xml output currently: Xapi_dr_task, and Quicktest_storage. They are not aware of the 'Probe' alternative and just try to parse the Xml, that parsing has to stay due to SMAPIv1 (the Xml comes straight from the python backend...). So we thought to also convert the SMAPIv3 response to that Xml for backwards compat, and introduce a new API `probe_ext` that returns a structured response (that will be in a future PR).

We could say that we do not support disaster recovery for SMAPIv3 yet (in fact the REQ explicitly excludes it), and instead have SMAPIv3 work *only* with probe_ext. This means we'd need to introduce a separate capability for it, but maybe thats best anyway.
When we do implement DR for SMAPIv3 we could do it by converting the SMAPIv1 raw Xml reply into the structured one: then we do not need to build an Xml by hand.

ocaml/xapi/xapi_sr.ml
edwintorok: We propose to introduce  a new API, probe_ext with the structed reply: https://github.com/xapi-project/xen-api/blob/feature/REQ477/probe/ocaml/idl/datamodel.ml#L1869

ocaml/xapi/xmlm_tree.ml
gaborigloi: Yes, not supporting the SMAPIv1 SR.probe for GFS2 / SMAPIv3 SRs is the other option. With  @jonludlam we found that SR.probe is being used by people, so I don't know whether it's okay to do that. I guess once we make SR.probe_ext work for SMAPIv1 SRs too, it will be reasonable to ask people to move to it, as it will work on all SRs. (This is why introducing a new capability may cause confusion in the future when SR.probe_ext will support everything.)

edwintorok: Should we make a decision on this after the TP, and just have `probe_ext` working for SMAPIv3 SRs, and probe throw a not implemented error? The main use case for probe in the TP is XenCenter which is updated to use `probe_ext`.

gaborigloi: Or we can leave it as it is, so that he quicktest keeps working, and make a decision after the TP.

mseri: I understand why you need more control now, I'd like not to make this become soon technical debt and instead try to improve things so that they are solid and as little hand-made or hacked as possible. I think the decision can be postponed until after the TP, but I'd like to see a plan being drafted and I'd be very happy if this plan includes getting rid of manually created and deserialised xml :stuck_out_tongue_winking_eye:

ocaml/xapi/xapi_sr.ml
mseri: I'm fine with that,I am happy to see there will be a new thing that uses a reasonable response format and that will be used by new code

ocaml/xapi/xmlm_tree.ml
mseri: Quecktest_storage does not need this module as I understand. This is just my opinion, but I still suggest to inliine the module next to where it is used and try to reduce its use as much as possible. I think if it is really indispensable for something else, it can be re-extracted later on.

gaborigloi: Sounds good, I'm happy to inline this module into the legacy probe function. Hopefully we won't need to do any manual XML processing and can just get rid of it when we get rid of `SR.probe`.

gaborigloi:
We can now include the `SR.probe_ext` and option type changes because the SDK is ready: https://github.com/xapi-project/xen-api-sdk/tree/feature/REQ-477

ocaml/xapi/xapi_sr.ml
jonludlam: It's always very hard to change the return type of an API call, sadly, and we know there are people out there who do actually use the probe call, so we don't want to break it unnecessarily.

PR #3536, opened 2018-03-28
CA-275591: remove dead code

krizex:
This change aims to remove dead code base on the analyzation result of
dead_code_analyzer (https://github.com/LexiFi/dead_code_analyzer)

It contains:
1. Remove unused variables
2. Remove unused methods

It does NOT remove the unused exported interfaces in header (*.mli) files


krizex:
Ring3 BVT passed in suite run 82033
Ring3 BST passed in suite run 82032

gaborigloi:
Looks good. Don't know whether we should keep the constants in `cli_key.ml` for reference, but I guess they will be in the git history, so it should be fine. Same for the unused fistpoints `xapi_fist.ml` - not sure whether we have any additional actions to take after removing them.

krizex:
Yes, I have the same concern with @gaborigloi . So please @jonludlam take a look of this change.

mseri:
Have you already run a BVT and a BST against this changes?

krizex:
@mseri Yes, Ring3 BST&BVT passed, I have posted the suite run ID above.

mseri:
Oh sorry, when reading I missed the comment

mseri:
Well, done! Let's double check with @jonludlam if we need to keep something for some reason

jonludlam:
Uh, how can coverage decrease when removing dead code!?

ocaml/xapi/helpers.ml
```diff
@@ -1048,19 +965,6 @@ let find_health_check_task ~__context ~sr =
       Eq (Field "name__description", Literal (Ref.string_of sr))
     ))
 
-(* Copy the snapshot metadata from [src_record] to the VM whose reference is [dst_ref]. *)
-(* If a lookup table is provided, then the field 'snapshot_of' is translated using this *)
-(* lookup table. *)
-let copy_snapshot_metadata rpc session_id ?lookup_table ~src_record ~dst_ref =
```
jonludlam: I'd like to know the history of this; when it came in and when it became dead code.

ocaml/xapi/xapi_sr_operations.ml
```diff
@@ -206,15 +206,6 @@ let update_allowed_operations ~__context ~self : unit =
   let keys = Hashtbl.fold (fun k v acc -> if v = None then k :: acc else acc) valid [] in
   Db.SR.set_allowed_operations ~__context ~self ~value:keys
 
-(** Someone is cancelling a task so remove it from the current_operations *)
-let cancel_task ~__context ~self ~task_id =
-  let all = List.map fst (Db.SR.get_current_operations ~__context ~self) in
-  if List.mem task_id all then
-    begin
-      Db.SR.remove_from_current_operations ~__context ~self ~key:task_id;
-      update_allowed_operations ~__context ~self
```
jonludlam: I wonder why we were calling `update_allowed_operations` here but not in `cancel_tasks` below this function...?

ocaml/xapi/xapi_vbd_helpers.ml
```diff
@@ -262,15 +262,6 @@ let update_allowed_operations ~__context ~self : unit =
   let keys = Hashtbl.fold (fun k v acc -> if v = None then k :: acc else acc) valid [] in
   Db.VBD.set_allowed_operations ~__context ~self ~value:keys
 
-(** Someone is cancelling a task so remove it from the current_operations *)
-let cancel_task ~__context ~self ~task_id =
-  let all = List.map fst (Db.VBD.get_current_operations ~__context ~self) in
-  if List.mem task_id all then
-    begin
-      Db.VBD.remove_from_current_operations ~__context ~self ~key:task_id;
-      update_allowed_operations ~__context ~self
```
jonludlam: Again, a suspicious lack of `update_allowed_operations` in the `cancel_tasks` function below.

mseri:
I think randomness in tests

krizex:
I think that's because `coveralls` mis calculated the sour code changes.
https://coveralls.io/jobs/35062005
How could `ocaml/xapi/cli_operations.ml` adds 4987 lines?

ocaml/xapi/helpers.ml
krizex: It comes in by the Initial import from hg on 2009, but it's hard to track when it became dead.

ocaml/xapi/xapi_vbd_helpers.ml
krizex: It also comes in by the Initial import from hg on 2009 and no changes after that.

ocaml/xapi/xapi_sr_operations.ml
krizex: Because this `cancel_tasks` functions only called in `Cancel_tasks.cancel_tasks_on_host` and in the end of the function there are a separate function to handle the `update_allowed_operations` stuff.

ocaml/xapi/xapi_vbd_helpers.ml
krizex: See above.

ocaml/xapi/helpers.ml
jonludlam: looks like it was here: 8e57c84c647076a2f6190ada0652f531796fdf90

jonludlam: OK, so it's not recent :-) I'm happy.

PR #3539, opened 2018-03-28
CA-281638: Set Pool.ha_cluster_stack when enabling/disabling clustering

minishrink:
If HA enabled on pool:
- set ha_cluster_stack to "corosync" when enabling clustering
- set ha_cluster_stack to "xhad" when disabling clustering

Linebreak separates assertions from API call logic

### DIFF
--- a/ocaml/xapi/xapi_cluster_host.ml
+++ b/ocaml/xapi/xapi_cluster_host.ml
@@ -136,6 +136,10 @@ let enable ~__context ~self =
       let network = Db.Cluster.get_network ~__context ~self:cluster in
       let pif = pif_of_host ~__context network host in
       assert_pif_prerequisites pif;
+
+      let pool = Helpers.get_pool ~__context in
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -136,6 +136,10 @@ let enable ~__context ~self =
       let network = Db.Cluster.get_network ~__context ~self:cluster in
       let pif = pif_of_host ~__context network host in
       assert_pif_prerequisites pif;
+
+      let pool = Helpers.get_pool ~__context in
+      if Db.Pool.get_ha_enabled ~__context ~self:pool
```
lindig: The code is fine and indentation is according to `ocp-indent`. However, I feel the code would be clearer with the conditional part indented, especially since there is no else branch.

```
if Db.Pool.get_ha_enabled ~__context ~self:pool then
  Db.Pool.set_ha_cluster_stack ~__context ~self:pool ~value:"corosync";
```
```

PR #3540, opened 2018-03-29
CA-282684: VDI migration with vGPU is failing when vGPU size is big

minli1:
When VM with GPU implementing VDI migration, resources (memory and Vgpu)
 will be reserved and then Vgpu and memory checks will be performed, when
 Vgpu has big size, checks will fail with no remaining capacity.
 This fix aims to check there has 'non_allocated' vgpus before check the
 GPU resources.

### DIFF
--- a/ocaml/xapi/xapi_vm_helpers.ml
+++ b/ocaml/xapi/xapi_vm_helpers.ml
@@ -349,12 +349,13 @@ let assert_host_has_iommu ~__context ~host =
   if List.assoc "iommu" chipset_info <> "true" then
     raise (Api_errors.Server_error (Api_errors.vm_requires_iommu, [Ref.string_of host]))
 
-(* Check if there are vgpus not allocated, since we may check the mem/vgpu after reservation *)
+(* Check if there are vgpus not allocated, since we may check the mem/vgpu after reservation 
+   Assumption: a VM can have only one vGPU *)
 let has_non_allocated_vgpus ~__context ~self =
-  let resident_pgpus = Db.VM.get_VGPUs ~__context ~self
-                       |> List.map (fun vgpu -> Db.VGPU.get_scheduled_to_be_resident_on ~__context ~self:vgpu) in
-  let non_allocated = List.filter (fun pgpu -> not (Db.is_valid_ref __context pgpu)) resident_pgpus in
-  non_allocated <> []
+  Db.VM.get_VGPUs ~__context ~self
+  |> List.map (fun vgpu -> Db.VGPU.get_scheduled_to_be_resident_on ~__context ~self:vgpu)
+  |> List.filter (fun pgpu -> not (Db.is_valid_ref __context pgpu))
+  |> (<>) []
 
 let assert_gpus_available ~__context ~self ~host =
   let this = "assert_gpus_available" in
--- a/ocaml/xapi/xapi_vm_helpers.ml
+++ b/ocaml/xapi/xapi_vm_helpers.ml
@@ -349,8 +349,7 @@ let assert_host_has_iommu ~__context ~host =
   if List.assoc "iommu" chipset_info <> "true" then
     raise (Api_errors.Server_error (Api_errors.vm_requires_iommu, [Ref.string_of host]))
 
-(* Check if there are vgpus not allocated, since we may check the mem/vgpu after reservation 
-   Assumption: a VM can have only one vGPU *)
+(* Check if there are vgpus not allocated, since we may check the mem/vgpu after reservation *)
 let has_non_allocated_vgpus ~__context ~self =
   Db.VM.get_VGPUs ~__context ~self
   |> List.map (fun vgpu -> Db.VGPU.get_scheduled_to_be_resident_on ~__context ~self:vgpu)
@@ -515,6 +514,7 @@ let assert_can_boot_here ~__context ~self ~host ~snapshot ?(do_sr_check=true) ?(
   assert_can_see_networks ~__context ~self ~host;
   if vm_needs_iommu ~__context ~self then
     assert_host_has_iommu ~__context ~host;
+  (* Assumption: a VM can have only one vGPU *)
   if has_non_allocated_vgpus ~__context ~self then
     assert_gpus_available ~__context ~self ~host;
   assert_usbs_available ~__context ~self ~host;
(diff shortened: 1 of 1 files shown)


minli1:
Job passed : 2085531

minli1:
May I ask help to review this change? @robhoes  and  @mseri 
Had some comments on this ticket with Marcello. At last, I think it's better to check the field `scheduled_to_be_resident_on` to indicate the VGPU had been reserved or not. Then the check will be avoid if so. For memory reservation, from the code I see there is no actually action to reserve the memory before the vm starts, but just set the `scheduled_to_be_resident_on`  to the host. So there is no need to avoid double checks for the memory I think.

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -507,7 +514,8 @@ let assert_can_boot_here ~__context ~self ~host ~snapshot ?(do_sr_check=true) ?(
   assert_can_see_networks ~__context ~self ~host;
   if vm_needs_iommu ~__context ~self then
     assert_host_has_iommu ~__context ~host;
-  assert_gpus_available ~__context ~self ~host;
+  if has_non_allocated_vgpus ~__context ~self then
+    assert_gpus_available ~__context ~self ~host;
   assert_usbs_available ~__context ~self ~host;
   assert_netsriov_available ~__context ~self ~host;
```
krizex: SR-IOV has the same issue.

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -349,6 +349,13 @@ let assert_host_has_iommu ~__context ~host =
   if List.assoc "iommu" chipset_info <> "true" then
     raise (Api_errors.Server_error (Api_errors.vm_requires_iommu, [Ref.string_of host]))
 
+(* Check if there are vgpus not allocated, since we may check the mem/vgpu after reservation *)
+let has_non_allocated_vgpus ~__context ~self =
```
krizex: I think the solution is unreliable.
Consider the scenario if the VM has 3 VGPUs but only 1 of them is non-allocated, it will still get the calculation result we need double amount (6) VGPUs.
So I think we'd better fix this issue in the `assert_*` functions and only alloc PGPU/SR-IOV for that non-allocated.

minli1: Thanks for comments. @krizex 
Consider the issue you mentioned before, but AFAIK that there is only one vgpu can be assigned to one VM, so the `list` at most get one element.
Yes, seems SR-IOV has the same issues, I'd like to have Rob's opinion on this, do we need to add checks on SR-IOV also to avoid double check? @robhoes

mseri: Yes, the code is written so that in the future we could be able to support multiple VGPUs with not too much effort, but as of now a lot of places are already relying on the fact that there is a VGPU only. I think this is fine here as well

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -349,6 +349,13 @@ let assert_host_has_iommu ~__context ~host =
...
+  let resident_pgpus = Db.VM.get_VGPUs ~__context ~self
+                       |> List.map (fun vgpu -> Db.VGPU.get_scheduled_to_be_resident_on ~__context ~self:vgpu) in
+  let non_allocated = List.filter (fun pgpu -> not (Db.is_valid_ref __context pgpu)) resident_pgpus in
+  non_allocated <> []
```
mseri: You can make this code a long chain:
```ocaml
Db.VM.get_VGPUs ~__context ~self
|> List.map (fun vgpu -> Db.VGPU.get_scheduled_to_be_resident_on ~__context ~self:vgpu)
|> List.filter (fun pgpu -> not (Db.is_valid_ref __context pgpu))
|> (<>) []
```

Or even
```ocaml
Db.VM.get_VGPUs ~__context ~self
|> List.exists (fun vgpu -> not (
     Db.VGPU.get_scheduled_to_be_resident_on ~__context ~self:vgpu
     |> Db.is_valid_ref __context
))
```

ocaml/xapi/xapi_vm_helpers.ml
mseri: Can you raise a ticket for this?

ocaml/xapi/xapi_vm_helpers.ml
minli1: Thanks, it really looks good. I'd like to change to the first one. Thanks. @mseri

ocaml/xapi/xapi_vm_helpers.ml
minli1: Will add a comment above the function to remind that `assumption of one VGPU per VM`
Thanks

ocaml/xapi/xapi_vm_helpers.ml
minli1: I will raise a ticket for SR-IOV for tracking.

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -349,12 +349,13 @@ let assert_host_has_iommu ~__context ~host =
   if List.assoc "iommu" chipset_info <> "true" then
     raise (Api_errors.Server_error (Api_errors.vm_requires_iommu, [Ref.string_of host]))
 
-(* Check if there are vgpus not allocated, since we may check the mem/vgpu after reservation *)
+(* Check if there are vgpus not allocated, since we may check the mem/vgpu after reservation 
+   Assumption: a VM can have only one vGPU *)
```
mseri: Let's move this Assumption comment below, at line 518, where you call `assert_gpus_available`. This function does not really care how many VGPUs there are

ocaml/xapi/xapi_vm_helpers.ml
mseri: Thanks

ocaml/xapi/xapi_vm_helpers.ml
minli1: Sure.

krizex:
LGTM. Please @minli1 submit another ticket to fix the SR-IOV issue.

PR #3542, opened 2018-04-04
CP-27343: SR probe: add clustering lock, assert cluster_host enabled

minishrink:
- Added an assertion for SR.probe as it is currently only safe with clustering enabled. 
- Wrapped `Xapi_sr.call_probe` (called by both `SR.probe` and `SR.probe_ext`) in a clustering lock in case the SR is in use elsewhere to prevent database inconsistency
- Removed local opens of `Storage_interface` and `Storage_access`

### DIFF
--- a/ocaml/xapi/xapi_sr.ml
+++ b/ocaml/xapi/xapi_sr.ml
@@ -170,20 +170,21 @@ let get_pbds ~__context ~self ~attached ~master_pos =
   | `Last -> slave_pbds @ master_pbds
 
 let call_probe ~__context ~host ~device_config ~_type ~sm_config ~f =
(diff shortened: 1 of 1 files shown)


mseri:
Why coveralls keeps messing the count?

minishrink:
I don't know how coveralls works, but is it possible that local opens contribute significantly?

mseri:
I don't think that's the issue. We have seen it happen in multiple PRs lately, even in ones where we ehere deleting lots of code

edwintorok:
This looks good, but please submit the PR to the `feature/REQ477/probe` branch where we collect the changes for `probe_ext` API, otherwise we will have conflicts.

edwintorok:
@minishrink please rebase there are too many commits in the PR, which means you probably have some duplicate commits that are already on the probe branch but with a different hash.

PR #3543, opened 2018-04-04
CA-286364: When creating guest_metrics, ensure all fields reflect cur…

jonludlam:
…rent state

In CA-286364, a reboot happened on a VM that had PV tools installed. This
triggered xapi's updates-from-xenopsd to look at the new state. This was slow
for other reasons and by the time we were looking at the events the VM had
reloaded its PV drivers.

When we notice a VM has a different start time we delete and recreate the guest
metrics. This object was, prior to this change, created with `PV_drivers_detected`
taking the default value of `false`.

We then took a look at the value of 'PV_drivers_detected' in the current state
of the VM and noticed that it hadn't changed since we last looked at it (before
the VM rebooted) - so we didn't set the field in the database.

The change in this CS ensures we always create the guest_metrics with the current
state of the VM, which we were already doing for all other fields in
`VM_guest_metrics`.


jonludlam:
Build failure looks related to the GFS merges.

PR #3544, opened 2018-04-04
[477 merge] CP-27346: SR.probe_ext API

edwintorok:
This is mostly @gaborigloi 's work, please review carefully the datamodel changes that introduce the option type, these having been reviewed yet as we only got the full build working recently.

At a high-level SR.probe_ext has a more structured reply type used by SMAPIv3, however this required some improvements to datamodel because it did not support option types.

Has to be merged together with:
https://github.com/xapi-project/xen-api-sdk/pull/146


ocaml/xapi/xmlm_tree.ml
```diff
@@ -0,0 +1,27 @@
...
+
+(** Creates a function deconstructing the above tree structure that can be
+    passed to Xmlm's tree output functions. *)
+let frag = function
+  | El (tag, children) -> `El (tag, children)
+  | Data s -> `Data s
+
+(** Outputs the given tree using [Xmlm.output_doc_tree] *)
+let output_doc ~tree ~output ~dtd =
+  Xmlm.output_doc_tree frag output (dtd, tree)
```
gaborigloi: I've inlined this in the SR.probe 1 PR in response to a review comment, so this file should be removed from the PR

edwintorok: Thanks, dropped the file now

edwintorok:
This was missing ticket numbers, I rebased and added them and also squashed the squashme commits.

ocaml/xapi/message_forwarding.ml
```diff
@@ -3263,6 +3267,12 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
       do_op_on ~local_fn ~__context ~host
         (fun session_id rpc -> Client.SR.probe ~rpc ~session_id ~host ~device_config ~_type ~sm_config)
 
+    let probe_ext ~__context ~host ~device_config ~_type ~sm_config =
+      info "SR.probe_ext: host = '%s'" (host_uuid ~__context host);
```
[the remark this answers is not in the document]
edwintorok: Using a uuid here makes it similar to SR.probe which also uses uuids.

ocaml/idl/ocaml_backend/gen_api.ml
```diff
@@ -93,18 +107,32 @@ let gen_record_type ~with_module highapi tys =
...
+        in
+        sprintf "%s = %s" (field fld) value
       in
 
       let type_t = sprintf "type %s_t = { %s }" obj_name (map_fields regular_def) in
       let others = if not with_module then
           []
         else [
-          sprintf "let rpc_of_%s_t x = Rpc.Dict [ %s ]" obj_name (map_fields make_of_field);
+          sprintf "let rpc_of_%s_t x = Rpc.Dict (Xapi_stdext_std.Listext.List.unbox_list [ %s ])" obj_name (map_fields make_of_field);
```
mseri: This is a rather heavy dependency. `Xapi_stdext_monadic.opt.map` and `Xapi_stdext_std.Listext.List.unbox_list` are so small and easy to implement that it could be worth adding them at line 175 with `assocer` and other local helpers, and avoid having to import two big modules for the api (also because they would then be needed in `xen-api-client` as well for no reason)

mseri: This is currently not failing only because we are getting the dependency transitively, but with further cleanup it could well break again.

ocaml/idl/datamodel.ml
```diff
@@ -1711,6 +1711,61 @@ module Data_source = struct
...
 end
 
+module Sr_stat = struct
+  let health = Enum ("sr_health", [
+      "healthy", "Storage is fully available";
+      "recovering", "Storage is busy recovering, e.g. rebuilding mirrors.";
+    ])
+
+  let t =
+    let lifecycle = [Published, rel_kolkata, ""] in
```
mseri: Is it published or preview here and below?

edwintorok: I think if we set it to `Prototyped` it won't show up in the docs. 
@jonludlam would you prefer `SR.probe_ext` to be in the docs for Kolkata or not?

kc284: I think stuff with Prototyped does appear in the docs - at least it is exposed in the SDK and we additionally mark it as experimental.

edwintorok: Hmm I do see some prototyped methods in the docs, like Host.get_management_interface, but all the clustering APIs are missing.

edwintorok: You are right, I've changed the lifecycle to say prototyped, we can set it to published in the next release.

ocaml/idl/ocaml_backend/gen_api.ml
```diff
@@ -93,18 +107,32 @@ let gen_record_type ~with_module highapi tys =
...
+        in
+        sprintf "%s = %s" (field fld) value
       in
 
       let type_t = sprintf "type %s_t = { %s }" obj_name (map_fields regular_def) in
       let others = if not with_module then
           []
         else [
-          sprintf "let rpc_of_%s_t x = Rpc.Dict [ %s ]" obj_name (map_fields make_of_field);
+          sprintf "let rpc_of_%s_t x = Rpc.Dict (unbox_list [ %s ])" obj_name (map_fields make_of_field);
```
jonludlam: what's this `unbox_list` bit here?

jonludlam: Ah, I see now.

ocaml/idl/ocaml_backend/gen_api.ml
```diff
@@ -135,8 +163,47 @@ let add_set_enums types =
...
+    in
+    match ty_ref with
+    | [] -> result @ ty_not_ref
+    | _ -> inner (result @ ty_not_ref) ty_ref
+  in
+  let result = inner [] types in
+  assert(List.length result = List.length types);
+  assert(List.sort compare result = List.sort compare types);
+  result
+
```
jonludlam: this looks familiar... and could do with quite careful review.

edwintorok: is it copied from elsewhere in xapi-project?

ocaml/idl/ocaml_backend/gen_api.ml
```diff
@@ -135,8 +163,47 @@ let add_set_enums types =
...
+      | DT.Record record when record = name -> true
+      | DT.Record record ->
+        let all_fields = DU.fields_of_obj (Dm_api.get_obj_by_name highapi ~objname:record) in
+        List.exists (fun fld -> references name fld.DT.ty) all_fields
+      | DT.Option ty -> references name ty
+    in
+    let (ty_ref,ty_not_ref) =
+      List.partition (fun ty -> match ty with
+      | DT.Record name ->
+        let referencing = List.filter (references name) remaining in
```
lindig: `List.length > 1` is an anti pattern. I'm not sure it is worth changing as there are probably better ways to optimize this.
```
match List.filter (references name) remaining with
| [] | [_] -> false
| _ -> true
```

edwintorok: When I read this code I initially thought its List.length > 0, and I could List.exists, but its not.

ocaml/idl/ocaml_backend/gen_api.ml
jonludlam: no: https://github.com/jonludlam/xen-api/commit/1cbe6ae058926bb740cc2fe7deddb9836af94011 - I didn't get around to making sure it was correct and submitting a PR - hence the comment about careful review :-)

edwintorok: My understanding is that the function works like this (not necesarily efficiently):
* it checks whether each type references any of the remaining types
* the types that do not reference anything are added straight to the sorted list: it doesn't matter in what order we emit these as they do not reference eachother, they could reference types emitted earlier though
* the remaining types are recursively checked with the same algorithm, since the leaf types got removed each type will reference one less, and thus the recursion eventually terminates

If you got self-referencing types or other loops the function would fail with a stackoverflow, but I think thats acceptable since this still happens at "build" time (when generating the ocaml API).

mseri: I got the same understanding. In the end we are relying on the fact that this is a build-time only thing and we can pay the little price it involves just that once. Also, it has not yet failed :D

Even without this, I think having circular references would have been a problem

jonludlam: Yes, that's how it's supposed to work. However, when I wrote this I recall it being a bit of a 'trial-and-error' style of development - it took a while to get the resulting code to compile. And that's all the testing it had at the time; that the resulting code compiled. I'd quite like to see some testing around this code.

edwintorok: You are right, if the code is wrong we would see a failure in compiling the generated code, so we know its not wrong at least for the types we currently have declared.

jonludlam: Absolutely true. Perhaps that's sufficient (for now).

edwintorok:
Squashed the fixups. @jonludlam are you with us merging this?

PR #3545, opened 2018-04-05
CP-25121, CP-26147: Prevent clustering operations during rolling pool upgrade or concurrently

minishrink:
Prevent all clustering operations during a rolling pool upgrade, wrap all clustering calls in a clustering lock, and add a test to check for this.

### DIFF
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -832,11 +832,11 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
     let clear_reserved_netsriov_vfs_on ~__context ~vm =
       Db.VM.get_VIFs ~__context ~self:vm
       |> List.iter (fun vif ->
-            let vf =  Db.VIF.get_reserved_pci ~__context ~self:vif in
-            Db.VIF.set_reserved_pci ~__context ~self:vif ~value:Ref.null;
-            if Db.is_valid_ref __context vf
-            then Db.PCI.set_scheduled_to_be_attached_to ~__context ~self:vf ~value:Ref.null
-          )
+          let vf =  Db.VIF.get_reserved_pci ~__context ~self:vif in
+          Db.VIF.set_reserved_pci ~__context ~self:vif ~value:Ref.null;
+          if Db.is_valid_ref __context vf
+          then Db.PCI.set_scheduled_to_be_attached_to ~__context ~self:vf ~value:Ref.null
+        )
 
     (* Notes on memory checking/reservation logic:
        		   When computing the hosts free memory we consider all VMs resident_on (ie running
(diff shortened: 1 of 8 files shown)


ocaml/xapi/xapi_cluster_host_helpers.ml
```diff
@@ -56,6 +56,7 @@ let get_operation_error ~__context ~self ~op =
   current_error
 
 let assert_operation_valid ~__context ~self ~op =
+  Helpers.assert_rolling_upgrade_not_in_progress ~__context;
```
edwintorok: We have to allow shutdown/reboot, i.e. disable/enable during a rolling pool upgrade, I think that is why the commented out code below had a list of what to allow.

ocaml/xapi/message_forwarding.ml
```diff
@@ -4285,6 +4285,7 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
       let pool = Db.Pool.get_all ~__context |> List.hd in (* assumes 1 pool in DB *)
       Xapi_pool_helpers.with_pool_operation ~__context ~self:pool ~doc:"Cluster.create" ~op:`cluster_create
         (fun () ->
+           Helpers.assert_rolling_upgrade_not_in_progress ~__context;
```
edwintorok: I think this should be part of xapi_pool_helpers.ml, in the valid_operations check for `cluster_create`. That way creating a cluster should get greyed out in XenCenter.

ocaml/xapi/message_forwarding.ml
```diff
@@ -4329,15 +4330,23 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
+        )
 
     let force_destroy ~__context ~self =
       info "Cluster_host.force_destroy";
       let local_fn = Local.Cluster_host.force_destroy ~self in
       let host = Db.Cluster_host.get_host ~__context ~self in
-      do_op_on ~__context ~local_fn ~host
-        (fun session_id rpc -> Client.Cluster_host.force_destroy rpc session_id self)
+      let cluster = Db.Cluster_host.get_cluster ~__context ~self in
+      Xapi_cluster_helpers.with_cluster_operation ~__context ~self:cluster ~doc:"Cluster_host.force_destroy" ~op:`destroy
```
edwintorok: This should be `remove`. `destroy` on the cluster object is for destroying the whole cluster.

ocaml/xapi/xapi_cluster_host_helpers.ml
minishrink: So we should allow all cluster_host operations during RPU and just prohibit cluster operations?

edwintorok: Yes, we want to prevent membership changes, but host lifecycle changes should be fine.

minishrink:
Updated PR to address Edvin's comments (commits are squashed) and refactor testing to reflect these changes.

minishrink:
I'd also like to investigate the possibility of adding a XenRT test for this.

lindig:
What is the status of this? If this requires a XenRT test, maybe indicate that this should not be merged.

edwintorok:
Note: this will also fix CP-26147, please add that into the title of the commit that changes message-forwarding

edwintorok:
Please include the ticket number in the indent commit too

edwintorok:
Travis build is working now, and @minishrink' manual test has shown that this code is working (you can still get back errors from the clustering daemon if hosts are offline, but that is expected, the RPU code is not blocking cluster_host.enable/disable), so we can merge this.

PR #3546, opened 2018-04-06
CA-287503: fix static-vdis for SMAPIv3

edwintorok:
There were no HA tests in the GFS2 functional or regression suites, just in a development branch which has now failed to enable HA on GFS2.
This is a regression due to the SR.attach API changes: static-vdis was not updated.

Instead of a seralized JSON in sr-uri SR.attach now takes a device-config string to string map, so we need to save and load this map to be able to attach the HA statefile.

I've done some testing on a 2 node GFS2 cluster and was able to enable/disable HA. There is a Jenkins build in progress, still remains to be tested that this patch has no effect on SMAPIv1 HA.

### DIFF
--- a/scripts/static-vdis
+++ b/scripts/static-vdis
@@ -2,12 +2,13 @@
 
 # Common functions for managing statically-attached (ie onboot, without xapi) VDIs
 
-import sys, os, subprocess, json, urlparse
+import sys, os, subprocess, json, urlparse, itertools
 import XenAPI, inventory, xmlrpclib
 
 main_dir = "/etc/xensource/static-vdis"
 
 xapi_storage_script = "/usr/libexec/xapi-storage-script"
+smapiv3_config = "device-config"
 
 def call_volume_plugin(name, command, args):
     args = [(xapi_storage_script + "/volume/org.xen.xapi.storage." + name + "/"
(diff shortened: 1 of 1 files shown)


scripts/static-vdis
```diff
@@ -237,19 +242,18 @@ def attach(vdi_uuid):
...
+                if not (os.path.exists(d + "/" + smapiv3_config)):
                     # SMAPIv1
                     config = read_whole_file(d + "/config")
                     driver = read_whole_file(d + "/driver")
                     path = call_backend_attach(driver, config)
                 else:
                     volume_plugin = read_whole_file(d + "/volume-plugin")
-                    sr_uuid = read_whole_file(d + "/sr-uuid")
-                    sr_uri = read_whole_file(d + "/sr-uri")
+                    configuration = json.loads(read_whole_file(d + "/" + smapiv3_config))
```
mseri: This is independent but, unless we need to support python 2.5, we could redefine `read/write_whole_file` Into more efficient and readable counterparts using the `io` module (https://docs.python.org/2/library/io.html), e.g.

```
Import io

def read_whole_file(filename):
    with io.open(filename,  mode="r", encoding="utf-8") as f:
        return f.read()
```

mseri:
This looks fine. Make sure to test it properly and check the backward snd upgrade/update compatibility

edwintorok:
Will do, note that HA must be turned off and back on during an update/upgrade, so changing what we store for static vdis should be fine (also we never shipped a SMAPIv3 plugin that uses HA before, pvsproxy uses a tmpfs which is not suitable for HA)

lindig:
Please remember that fixes for Kolkata need to go to branch `1.90-lcm`. We can cherry pick it from the master branch.

edwintorok:
Tested a master build: enable HA on NFS, disable HA, update to my build, enable HA, reboot, disable HA
Same thing when starting off with my build and GFS2 HA, enabling/disabling HA works now.

PR #3547, opened 2018-04-08
CA-286874: Redundant checks for SR-IOV when implementing VDI migration

minli1:
This fix along with CA-282684 aim to fix scenario when VM implementing VDI
migration VGPUs/Network SR-IOV (if have) will reserve resources first
then again check the remaining capacity. It means when there are not two
times resources than it request,the check will fail.


ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -410,18 +410,24 @@ let assert_usbs_available ~__context ~self ~host =
       ]))
     )
 
-(* Get SR-IOV Vifs by return a list of [(network1,(required_num1,PCI1));(network2,(required_num2,PCI2))....] 
+(* 1.To avoid redundant checks,if VF had been reserved, then it's no need to check the ability again.
```
minglumlu: Is this true: if any VF has been reserved, all other VFs are supposed to be reserved also?

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -410,18 +410,24 @@ let assert_usbs_available ~__context ~self ~host =
...
   let sriov_networks = List.fold_left (fun acc vif ->
-      let network = Db.VIF.get_network ~__context ~self:vif in
-      try
-        let required, pif = List.assoc network acc in
-        (network,(required + 1,pif)) :: (List.remove_assoc network acc)
-      with Not_found ->
-        match Xapi_network_sriov_helpers.get_local_underlying_pif ~__context ~network ~host with
-        | Some pif -> (network,(1,pif)) :: acc
-        | None -> acc
+      let reserved_pci = Db.VIF.get_reserved_pci ~__context ~self:vif in
```
minglumlu: How about to move the check into SR-IOV logic?
We may need to make sure it is a SR-IOV VF firstly, then check if it has been reserved.

ocaml/xapi/xapi_vm_helpers.ml
minli1: Yes, I think so. Because at here the VF(s) has been reserved already if have, during reservation of VF(s), if any one of the VF reservation failed, the whole reservation will fail and rollback the reserved VFs. Also, reservation failure will raise exn.
So we can make assumtion that if any VF been reserved. all other VFs are supposed to be reserved.

ocaml/xapi/xapi_vm_helpers.ml
minli1: Yes, but seems that we determine a VIF is SR-IOV or not is to whether we can get the underlying PIF.
While `get_local_underlying_pif` may consumes much effort I think, (it will call `get_local_underlying_pif` in the end which is recursive function).
Since plain VIF won't set the `reserved_pci` at all, if it's not `Ref.null` then must be SR-IOV VIF which has been reserved already, then we do not need to `get_local_underlying_pif` below. Right?
Correct me if it's wrong, thanks for comment.

minglumlu: At the moment, it is correct. I just have a little bit concern on treating `reserved_pci` standalone as an indicator of a SR-IOV VF-backed VIF.

minglumlu:
Please refine the commit message.

minli1:
Refined commit message, please help to look again @minglumlu 
Point out the incorrect places if have, thanks.

ocaml/xapi/xapi_vm_helpers.ml
robhoes: It's still checking per-VIF anyway, isn't it?

minli1: Yes, it's still checking per-VIF, do you mean it's better once we found one VIF has SR-IOV reserved, than we can cancel the assertion?

robhoes: No, I don't think that's better. I thought from Ming's comment that he read your code comment that way. But I looks like you didn't mean that.

lindig:
If this fix is for Kolkata, we need an additional PR for branch `1.90-lcm` or tell me to cherry-pick the commit.

lindig:
We are not merging this to Kolkata for now. It most likely will go to Kolkata during selected integration period.

ocaml/xapi/xapi_vm_helpers.ml
minli1: Thanks, I had refined the comment above the function.

minli1:
Hi, Lindig, thanks for your reminding. @lindig 
As this fix is for SR-IOV which is a new feature of K release, even though there wasn't much probability to come with the issue, but I think it's better to merge it to master ASAP.
Could you help to merge it, thanks!

lindig:
I would like to see @robhoes's review. Once it is on the xapi master branch, I can cherry-pick it to the 1.90-lcm branch.

robhoes:
Let's put it in 1.90-lcm as well.

minli1:
Please help to merge :)  @lindig

PR #3549, opened 2018-04-09
CA-281638: Set pool.ha_cluster_stack for Cluster.create/destroy

minishrink:
This is a follow-up to my previous PR [here](https://github.com/xapi-project/xen-api/pull/3539), which set the pool's HA cluster stack for `Cluster_host.enable/disable`. Unfortunately this wasn't enough to automatically update the value, so now this PR should cover the introduction and destruction of a cluster, along with enabling and disabling clustering.

Note: I'm adding unit tests _but_ they use a change made [in this PR](https://github.com/xapi-project/xen-api/pull/3554), which should be merged first.

### DIFF
--- a/ocaml/tests/test_clustering.ml
+++ b/ocaml/tests/test_clustering.ml
@@ -270,12 +270,12 @@ let test_clustering_lock_only_taken_if_needed =
   ]
 
 let test_assert_pif_prerequisites () =
-  let __context = Test_common.make_test_database () in
-  let network = Test_common.make_network ~__context () in
+  let __context = T.make_test_database () in
+  let network = T.make_network ~__context () in
   let localhost = Helpers.get_localhost ~__context in
-  let (_cluster, _cluster_host) = Test_common.make_cluster_and_cluster_host ~__context ~network ~host:localhost () in
+  let (_cluster, _cluster_host) = T.make_cluster_and_cluster_host ~__context ~network ~host:localhost () in
   let exn = "we_havent_decided_on_the_exception_yet" in
-  let pifref = Test_common.make_pif ~__context ~network ~host:localhost () in
+  let pifref = T.make_pif ~__context ~network ~host:localhost () in
   let pif = Xapi_clustering.pif_of_host ~__context network localhost in
   Alcotest.check_raises
     "test_assert_pif_prerequisites should fail at first"
@@ -324,14 +324,14 @@ let check_disallow_unplug expected_value __context pif msg =
 
 (* Need host and network to make PIF *)
 let make_host_network_pif ~__context =
-  let host = Test_common.make_host ~__context () in
-  let network = Test_common.make_network ~__context () in
-  let pif = Test_common.make_pif ~__context ~network ~host () in
+  let host = T.make_host ~__context () in
+  let network = T.make_network ~__context () in
+  let pif = T.make_pif ~__context ~network ~host () in
   (host, network, pif)
 
 (* Test PIF.set_disallow_unplug without cluster/cluster_host objects *)
 let test_disallow_unplug_no_clustering () =
-  let __context = Test_common.make_test_database () in
+  let __context = T.make_test_database () in
   let host,network,pif = make_host_network_pif ~__context in
 
   (* Test toggling disallow_unplug when disallow_unplug:false by default *)
@@ -342,7 +342,7 @@ let test_disallow_unplug_no_clustering () =
     "check_disallow_unplug called by test_disallow_unplug_no_clustering after setting disallow_unplug:true";
 
   (* Test toggling disallow_unplug when initialised to true *)
-  let pif_no_unplug = Test_common.make_pif ~__context ~network ~host ~disallow_unplug:true () in
+  let pif_no_unplug = T.make_pif ~__context ~network ~host ~disallow_unplug:true () in
   check_disallow_unplug true __context pif_no_unplug
     "check_disallow_unplug called by test_disallow_unplug_no_clustering when initialising disallow_unplug:true";
   Xapi_pif.set_disallow_unplug ~__context ~self:pif_no_unplug ~value:false;
@@ -350,7 +350,7 @@ let test_disallow_unplug_no_clustering () =
     "check_disallow_unplug called by test_disallow_unplug_no_clustering after setting disallow_unplug:false"
 
 let test_disallow_unplug_with_clustering () =
-  let __context = Test_common.make_test_database () in
+  let __context = T.make_test_database () in
   let host,network,pif = make_host_network_pif ~__context in
   check_disallow_unplug false __context pif
     "check_disallow_unplug called by test_disallow_unplug_with_clustering to check default config";
@@ -361,7 +361,7 @@ let test_disallow_unplug_with_clustering () =
     "check_disallow_unplug called by test_disallow_unplug_with_clustering after setting disallow_unplug:true";
 
   (* PIF.disallow_unplug should become RO upon introduce cluster_host object, should throw exception when changing value *)
-  let _ = Test_common.make_cluster_and_cluster_host ~__context ~network ~host () in
+  let _ = T.make_cluster_and_cluster_host ~__context ~network ~host () in
   Alcotest.check_raises
     "check_disallow_unplug called by test_disallow_unplug_with_clustering after attaching cluster and cluster_host to network"
     (Api_errors.(Server_error(clustering_enabled_on_network, [Ref.string_of network])))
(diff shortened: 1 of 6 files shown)


ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -162,9 +160,7 @@ let disable ~__context ~self =
       assert_operation_host_target_is_localhost ~__context ~host;
       assert_cluster_host_has_no_attached_sr_which_requires_cluster_stack ~__context ~self;
 
-      let pool = Helpers.get_pool ~__context in
-      if Db.Pool.get_ha_enabled ~__context ~self:pool then
-        Db.Pool.set_ha_cluster_stack ~__context ~self:pool ~value:"xhad";
+      set_ha_cluster_stack_if_ha_enabled ~__context ~value:"xhad";
```
gaborigloi: Hmm, maybe we don't need to set / unset this pool field if we only enable / disable one cluster_host?

minishrink: Perhaps it would make more sense to check across all cluster hosts?

ocaml/xapi/xapi_cluster.ml
```diff
@@ -88,6 +89,7 @@ let destroy ~__context ~self =
       Db.Cluster_host.destroy ~__context ~self:ch
     ) cluster_host;
     Db.Cluster.destroy ~__context ~self;
+    set_ha_cluster_stack_if_ha_enabled ~__context ~value:"xhad";
```
edwintorok: We should not hardcode "xhad" here. See Cluster_stack_constraints.choose_cluster_stack.
Ideally you would just call that to tell you the cluster stack, and then all you need is an update_ha_cluster_stack.

edwintorok: We should not hardcode "xhad" here. See `Cluster_stack_constraints.choose_cluster_stack`.
Ideally you would just call that to tell you the cluster stack, and then all you need is an update_cluster_stack.
I think `Cluster_stack_constraints.required_cluster_stack` should return 'corosync' if you got a Cluster object in your pool.
We should also set the cluster stack even if HA is not enabled, I think there was a ticket requiring to do that: even if you do not have HA enabled if you got corosync you *cannot* enable `xhad`.

ocaml/xapi/xapi_cluster_host.ml
gaborigloi: @edwintorok Why do we need to switch to xhad if we only disable one cluster host?

edwintorok: You are right, the cluster stack should not be changed at all in Xapi_cluster_host

edwintorok: Just simply creating a cluster should be enough to make it incompatible with xhad, but like I said in my other comment, the code should be changed in Cluster_constraints.ml too.

ocaml/xapi/cluster_stack_constraints.ml
```diff
@@ -26,19 +26,18 @@ let required_cluster_stack ~__context =
...
       let sr_type = Db.SR.get_type ~__context ~self:sr in
-      if List.mem_assoc sr_type constraints then
-        match List.assoc sr_type constraints with
+      try
+        (match List.assoc sr_type constraints with
         | [] -> None    (* No constraints *)
-        | l ->  Some l  (* Any one of these will do *)
-      else begin
+        | l ->  Some l)  (* Any one of these will do *)
+      with Not_found -> begin
```
edwintorok: you can also use `match | exception Not_found ->` instead of try match with.

ocaml/xapi/cluster_stack_constraints.ml
```diff
@@ -40,8 +40,17 @@ let required_cluster_stack ~__context =
...
+      failwith "Conflicting cluster stack demands."
+  in
   match required_stacks with
-  | [] -> active_cluster_stack  (* None of the attached SRs have constraints *)
+  | [] ->
+    (* None of the attached SRs have constraints *)
+    begin match active_cluster_stack with
+    | None -> None
+    | Some [ stack ] -> Some stack
+    | Some _ -> failwith_cluster_stack_conflict () end
```
edwintorok: the `end` should go onto the next line

ocaml/xapi/cluster_stack_constraints.ml
```diff
@@ -40,8 +40,17 @@ let required_cluster_stack ~__context =
...
+  let failwith_cluster_stack_conflict () =
+      error "Conflicting cluster stack demands.";
+      failwith "Conflicting cluster stack demands."
+  in
   match required_stacks with
-  | [] -> active_cluster_stack  (* None of the attached SRs have constraints *)
+  | [] ->
+    (* None of the attached SRs have constraints *)
+    begin match active_cluster_stack with
+    | None -> None
```
edwintorok: you could just match the empty list here, not sure if there is value in transforming the empty list to None, and other lists to Some ...

minishrink:
Updated, starting on tests now.

minishrink:
Added unit tests

ocaml/tests/test_clustering.ml
```diff
@@ -376,12 +376,205 @@ let test_disallow_unplug_ro_with_clustering_enabled =
...
+  choose_cluster_stack_should_select default ~__context;
+
+  (* Add two cluster, test choose_cluster_stack's filtering *)
+  for i = 0 to 1 do
+    let _ = T.make_cluster_and_cluster_host ~__context () in
+    choose_cluster_stack_should_select default_smapiv3 ~__context
+  done;
+
+  (* choose_cluster_stack should fail with multiple different cluster_stacks *)
+  let _ = T.make_cluster_and_cluster_host ~__context ~cluster_stack:"different_stack" () in
```
edwintorok: I think the creation would fail in a real situation here because you can't have more than one cluster stack, so this last part of test should probably be removed.

ocaml/tests/test_clustering.ml
```diff
@@ -376,12 +376,205 @@ let test_disallow_unplug_ro_with_clustering_enabled =
...
+  let _  : API.ref_PBD = T.make_pbd ~__context ~host ~currently_attached:true ~sR () in
+  sR
+
+let test_choose_cluster_stack_sms_no_clusters () =
+  let __context = T.make_test_database () in
+
+  (* create SMs with cluster_stack constraints *)
+  let _  : API.ref_SM = T.make_sm ~__context ~_type:"sm" ~required_cluster_stack:[ test_default ] () in
+  let _  : API.ref_SM = T.make_sm ~__context ~_type:"nfs" ~required_cluster_stack:[] () in
+  let _  : API.ref_SM = T.make_sm ~__context ~_type:"lvm" ~required_cluster_stack:[ default_smapiv3 ; default ] () in
```
edwintorok: please call this something else than lvm, lvmoiscsi requires no cluster stacks

ocaml/tests/test_clustering.ml
```diff
@@ -376,12 +376,205 @@ let test_disallow_unplug_ro_with_clustering_enabled =
...
+  | _ -> Alcotest.fail "only one PBD should be plugged into this SR"
+  end;
+  choose_cluster_stack_should_select default_smapiv3 ~__context;
+
+  (* default_smapiv3 is the only common cluster stack *)
+  let _  : API.ref_SM = T.make_sm ~__context ~_type:"hba" ~required_cluster_stack:[ test_default ; default_smapiv3 ] () in
+  let _ = create_and_plug_sr ~_type:"hba" ~__context () in
+  choose_cluster_stack_should_select default_smapiv3 ~__context;
+
+  let _  : API.ref_SM = T.make_sm ~__context ~_type:"gfs2" ~required_cluster_stack:[ default_smapiv3 ; default_smapiv3 ] () in
```
edwintorok: did you mean to repeat the stack twice here?

ocaml/tests/test_clustering.ml
```diff
@@ -376,12 +376,205 @@ let test_disallow_unplug_ro_with_clustering_enabled =
...
+let get_only_cluster_host ~__context : API.ref_Cluster_host =
+  match Db.Cluster_host.get_all ~__context with
+  | [ cluster_host ] -> cluster_host
+  | lst ->
+    let nodes = lst |> List.length |> string_of_int in
+    raise Api_errors.(Server_error (cluster_does_not_have_one_node, [ nodes ]))
+
+let test_pool_ha_cluster_stacks_no_ha_with_clustering () =
+  (* Test that cluster creation and destruction don't set
+   * HA cluster stacks, as HA is disabled *)
```
edwintorok: They should in fact set it, because if  you would enable HA you wouldn't be able to do so with xhad, thus xhad is not a valid choice anymore.

ocaml/xapi/xapi_clustering.ml
```diff
@@ -17,6 +17,13 @@ open Cluster_interface
 module D=Debug.Make(struct let name="xapi_clustering" end)
 open D
 
+(* Called by Cluster.create/destroy and Cluster_host.enable/disable *)
+let set_ha_cluster_stack_if_ha_enabled ~__context =
+  let pool = Helpers.get_pool ~__context in
+  let value = Cluster_stack_constraints.choose_cluster_stack ~__context in
+  if Db.Pool.get_ha_enabled ~__context ~self:pool then
```
edwintorok: I think you should drop the `if` we should always set it, because you can query this parameter with `xe`.
If you enable corosync and it shows ha cluster stack as `xhad`, then one might think that enabling HA could use `xhad` if a suitable heartbeat SR is chosen, but that is wrong, therefore once corosync is enabled we should update the acceptable choice for ha cluster stack.

ocaml/tests/test_clustering.ml
minishrink: Yes, I wanted to check that doesn't confuse `choose_cluster_stacks`.

ocaml/xapi/xapi_clustering.ml
```diff
@@ -19,10 +19,9 @@ open D
 
 (* Called by Cluster.create/destroy and Cluster_host.enable/disable *)
 let set_ha_cluster_stack_if_ha_enabled ~__context =
```
edwintorok: Please rename this function, you dropped the `if` below, drop it from the name too.

minishrink:
Note: this PR makes use of the changes from [this PR](https://github.com/xapi-project/xen-api/pull/3554) so I think both should be cherry-picked to 1.90-lcm.

PR #3550, opened 2018-04-10
CP-27560 move to Device Model "qemu-upstream-compat"

lindig:
The commits implement that imported and existing VMs use Device Model "qemu-upstream-compat". The change is implemented in ensure_device_model_profile_present() that either assigns this as the new default, or updates an existing device model. This function is also called during a xapi upgrade.

### DIFF
--- a/ocaml/idl/datamodel_common.ml
+++ b/ocaml/idl/datamodel_common.ml
@@ -7,7 +7,7 @@ open Datamodel_roles
               You do not have to bump vsn if you change/add/remove a message *)
 
 let schema_major_vsn = 5
-let schema_minor_vsn = 141
+let schema_minor_vsn = 142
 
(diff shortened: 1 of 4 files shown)


rosslagerwall:
The second commit message says `WIP`. Is this a mistake or does something still need to be done?

lindig:
This is a mistake. I will correct this. The one other thing I'm not sure: applying the upgrade rule could be limited to `<= Kolkata` but I'm not sure if this introduces the chance of missing something.

rosslagerwall:
> The one other thing I'm not sure: applying the upgrade rule could be limited to <= Kolkata but I'm not sure if this introduces the chance of missing something.

I think it should be fine to apply it regardless of the version.

lindig:
This passed BVT and BST. See the discussion on the ticket how xenopsd is affected.

PR #3552, opened 2018-04-11
CA-267687: Logs of VBD operation check that inspects operations of VBD's VDI do not match returned errors

minli1:
If operations of VBD's VDI is not copy(mirror), then the log will shows
"VBD operation  not allowed because VDI.current-operations =    "
unconditionally, this is wrong because only attach/plug/insert will be blocked
for VBD, other operations will succeed and also with the confusing log.


minli1:
As Gabor and Marcello had approved the PR, I think it also can merge and pick up in Kolkata even it's just improvement for the misleading logs. Thanks  @lindig

lindig:
This will only go into the K release if it is approved as an important fix.

PR #3554, opened 2018-04-11
CA-287854: Check for valid cluster_stack in Cluster.create

minishrink:
This PR addresses two issues:
- prevents users from creating a cluster with an invalid cluster_stack, or a 'valid' one (currently `xhad`) that might cause confusion in any HA or clustering checks
- cuts down on hardcoding cluster stacks in tests and other Xapi code to keep code maintainable if support for additional cluster stacks is added

by doing the following:
- defining constant values for cluster_stack comparison
- - `default_smapiv3_cluster_stack = "corosync"`
- - `supported_smapiv3_cluster_stacks = [ "corosync" ]`
- introducing the API error `invalid_cluster_stack` for any supplied cluster_stacks not found in the above list
- adding a test to catch this error
- replacing hardcoded `"corosync"` references with the constant


ocaml/tests/test_cluster.ml
```diff
@@ -37,7 +37,11 @@ let test_rpc ~__context call =
   | name, params ->
      failwith (Printf.sprintf "Unexpected RPC: %s(%s)" name (String.concat " " (List.map Rpc.to_string params)))
 
-let create_cluster ~__context =
+let create_cluster ~__context ?cluster_stack () =
+  let cluster_stack = match cluster_stack with
```
edwintorok: You can use ?(cluster_stack=Constants.default_smapiv3_cluster_stack) instead of this match

minishrink:
My [pool ha-cluster-stack PR](https://github.com/xapi-project/xen-api/pull/3549) uses this code to avoid hardcoding cluster stack values. If that PR is going into Kolkata, this one probably should too, otherwise it's fine.

PR #3555, opened 2018-04-12
XSI-6: Event class documentation enhancements.

kc284:
Extended documentation in the datamodel and special cased the class in the code generating the API reference markdown. Note that this PR does not conclude the ticket as the html reference generator has not been fixed yet.

### DIFF
--- a/ocaml/idl/markdown_backend.ml
+++ b/ocaml/idl/markdown_backend.ml
@@ -115,7 +115,8 @@ let is_deprecation_marker x =
   match x with | (Deprecated,_,_) -> true | _ -> false
 
 (* Make a markdown section for an API-specified message *)
-let markdown_section_of_message printer ~is_class_deprecated ~is_class_removed x =
+let markdown_section_of_message printer obj ~is_class_deprecated ~is_class_removed x =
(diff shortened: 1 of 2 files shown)


ocaml/idl/markdown_backend.ml
```diff
@@ -161,9 +166,14 @@ let markdown_section_of_message printer ~is_class_deprecated ~is_class_removed x
     List.iter (fun p -> printer (get_param_row p)) x.msg_params;
     printer "";
 
-    printer (sprintf "_Return Type:_ `%s`" return_type);
+    printer ("_Return Type:_" ^ (if is_event_from then " an event batch" else sprintf " `%s`" return_type));
     printer "";
-    let descr= desc_of_ty_opt x.msg_result in
+    let descr =
+      if is_event_from then
+        "A structure consisting of a token ('token'), a map of valid references per object type ('valid_ref_counts'), and a set of event records ('events')."
```
lindig: Long lines like this are difficult to review on GitHub.

mseri: I agree, but it is using the style of the rest of the file, so I'd accept them in this case

PR #3556, opened 2018-04-12
CA-287863: xe vm-shutdown complete the task too early

krizex:
When master forward a task to slave, slave will mark the task as completed once
finish processing it, however when master finishing the forwarding, it will also
mark the task as completed. Furthermore, when slave finished the task, that does
not mean the task is done, there may be some stuff that master has to handle, so
we have to ensure the task only complete once the master finish its work.

### DIFF
--- a/ocaml/xapi/server_helpers.ml
+++ b/ocaml/xapi/server_helpers.ml
@@ -55,57 +55,56 @@ let unknown_rpc_failure func =
 let parameter_count_mismatch_failure func expected received =
   API.response_of_failure Api_errors.message_parameter_count_mismatch [func; expected; received]
 
-(* Execute fn f in specified __context, marshalling result with "marshaller".
-   If has_task is set then __context has a real task in it that has to be completed. *)
-let exec ?marshaller ?f_forward ~__context f =
-  (* NB:
-     	  1. If we are a slave we process the call locally assuming the locks have
-     	  already been taken by the master
-     	  2. If we are the master, locks are only necessary for the potentially-forwarded
-     	  (ie side-effecting) operations and not things like the database layer *)
-  try
-    let result =
-      if not(Pool_role.is_master ())
-      then f ~__context (* slaves process everything locally *)
-      else match f_forward with
-        | None ->
-          (* this operation cannot be forwarded (eg database lookup); do it now *)
-          f ~__context
-        | Some forward ->
-          (* use the forwarding layer (NB this might make a local call ultimately) *)
-          forward ~local_fn:f ~__context
-    in
-    begin match marshaller with
-      | None    -> TaskHelper.complete ~__context None
-      | Some fn -> TaskHelper.complete ~__context (Some (fn result))
-    end;
-    result
-  with
-  | Api_errors.Server_error (a,b) as e when a = Api_errors.task_cancelled ->
-    Backtrace.is_important e;
-    TaskHelper.cancel ~__context;
-    raise e
-  | e ->
-    Backtrace.is_important e;
-    TaskHelper.failed ~__context e;
-    raise e
 
 (** WARNING: the context is destroyed when execution is finished if the task is not forwarded, in database and not called asynchronous. *)
 (*  FIXME: This function should not be used for external call : we should add a proper .mli file to hide it. *)
 let exec_with_context ~__context ?marshaller ?f_forward ?(called_async=false) f =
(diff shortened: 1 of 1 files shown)


ocaml/xapi/server_helpers.ml
```diff
@@ -75,6 +75,8 @@ let exec ?marshaller ?f_forward ~__context f =
           (* use the forwarding layer (NB this might make a local call ultimately) *)
           forward ~local_fn:f ~__context
     in
+    (* For forwarded task, we should not complete it here, the server which forward the task will complete it *)
+    if not (Context.forwarded_task __context) then
```
lindig: This code has become dependent on a condition but indentation has not changed to reflect this. Please check the indentation of the code.

robhoes:
Don't merge yet though, because some BST tests seem to be failing. Please check those out first.

ocaml/xapi/server_helpers.ml
```diff
@@ -75,19 +75,21 @@ let exec ?marshaller ?f_forward ~__context f =
           (* use the forwarding layer (NB this might make a local call ultimately) *)
           forward ~local_fn:f ~__context
     in
-    begin match marshaller with
+    (* For forwarded task, we should not complete it here, the server which forward the task will complete it *)
+    if not (Context.forwarded_task __context) then begin
+      match marshaller with
       | None    -> TaskHelper.complete ~__context None
       | Some fn -> TaskHelper.complete ~__context (Some (fn result))
```
krizex: @robhoes I have some concern about this. After this change, the result in the slave won't be updated to the task result, but instead will update with the execution result in the master, will it bring some other issues?

robhoes: I should hope that the result of the forwarded call is already always returned (unaltered) by the master to the original caller, especially for synchronous calls. So I think this is fine.

robhoes:
The failing test seems to be an import that times out. We should check how the task completion is handled for such calls, which don't come as RPCs through server.ml, but have their own HTTP handlers.

krizex:
Seems the installation of the build is failed, so all BST & BVT failed. I have updated with the latest specs of team/ring3/master and has started rerun the test again.
Privately I have tested this fix by uploading xapi to my hosts and the error messages gone.

robhoes:
Well, the installation was OK, but the 62-create-guest-templates firstboot script failed:
```
Load /usr/share/xapi/vm-templates/sles-11-sp4-64bit.json
Load /usr/share/xapi/vm-templates/ubuntu-12.04-64bit.json
Load /usr/share/xapi/vm-templates/base-windows-8.json
Insert 11fd3dc9-96cc-49af-b091-a2ca7e94c589
  File "/usr/bin/create-guest-templates", line 17, in <module>
    loader.insert_templates()
  File "/usr/lib/python2.7/site-packages/guesttemplates/loader.py", line 181, in insert_templates
    self._insert_template(i)
  File "/usr/lib/python2.7/site-packages/guesttemplates/loader.py", line 165, in _insert_template
    raise RuntimeError("Template import timeout")
RuntimeError: Template import timeout
```
This does a VM import.

krizex:
Hmm, I will take a look at that tomorrow.

robhoes:
Hmm... I wonder why `exec_with_forwarded_task` is used in `Xapi_http.with_context`. That looks odd to me. In this patch, I don't quite like the new `~with_task` param. I'll look a bit further...

krizex:
@robhoes The name of `has_task` comes from the original comment https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/server_helpers.ml#L59
I meant to introduce a new param `~need_complete` ;-)

robhoes:
Ah, that was another mysterious comment indeed :)

robhoes:
I think that it is declared a forwarded task, to make sure that the task is not destroyed when the operation completes; the task is created by the client, and should be destroyed by the client.

I also notice that the "forwarded task destroyed" log lines that we often see actually mean that the forwarded task was _not_ destroyed...

robhoes:
Well that is indeed the problem and one of the sources of confusion: the word forwarded is used to different things, or at least the function is "abused" for different things. I think that we should reserve the term "forwarding" for the action of passing a call from the master to a slave (as in "message_forwarding").

I have some ideas now to simplify some of this area, and will do that in follow-up PRs. I'll accept this PR now, because it already fixes a real issue.

PR #3557, opened 2018-04-12
CA-286165: Remove PCI dependencies of VFs from the same PF.

xiewei20082008:
The PCI bus path of SR-IOV VFs have the same function number, but actually different VFs do not have PCI dependencies with each other. So here we remove dependencies of VFs from the same PF

### DIFF
--- a/ocaml/xapi/xapi_pci.ml
+++ b/ocaml/xapi/xapi_pci.ml
@@ -89,12 +89,12 @@ let update_pf_vf_relations ~__context ~pcis =
       Some (Filename.basename (Unix.readlink path))
     with _ -> None
   in
-  let set_phyfn (vf_ref, vf_rec, phyfn_path) pfs =
+  let set_phyfn (vf_ref, (vf_rec, phyfn_path)) pfs =
     match phyfn_path with
     | Some phyfn_path ->
       begin
         try
-          let pf, _, _ = List.find (fun (_, pf_rec, _) -> phyfn_path = pf_rec.Db_actions.pCI_pci_id) pfs in
+          let pf, (_, _) = List.find (fun (_, (pf_rec, _)) -> phyfn_path = pf_rec.Db_actions.pCI_pci_id) pfs in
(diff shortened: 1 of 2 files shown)


xiewei20082008:
Dev test passed.
In Job 2098450, MxGPU can be detected after guest driver installed.

ocaml/xapi/xapi_pci.ml
```diff
@@ -89,21 +89,32 @@ let update_pf_vf_relations ~__context ~pcis =
...
   in
-  let set_phyfn (vf_ref, vf_rec, phyfn_path) pfs =
+  let set_phyfn (vf_ref, (vf_rec, phyfn_path)) pfs =
     match phyfn_path with
     | Some phyfn_path ->
       begin
         try
-          let pf, _, _ = List.find (fun (_, pf_rec, _) -> phyfn_path = pf_rec.Db_actions.pCI_pci_id) pfs in
+          let pf, (_, _) = List.find (fun (_, (pf_rec, _)) -> phyfn_path = pf_rec.Db_actions.pCI_pci_id) pfs in
           if vf_rec.Db_actions.pCI_physical_function <> pf then Db.PCI.set_physical_function ~__context ~self:vf_ref ~value:pf
```
robhoes: Isn't it enough to just clear the `PCI.dependencies` right here? Here we know that `vf_ref` is a VF, which should never have `dependencies`. This is what the old MxGPU function used to do.

xiewei20082008: I was thinking whether there is any case where a VF has PCI dependencies.
If we presume a VF should never have `dependencies`, then it looks more reasonable to fix this at `update_dependencies` (only update dependencies for PF), since it is wield to set the dependencies for VFs at a place and then remove them again at another place.
But yes, we can quickly fix this by clearing the PCI.dependencies right here.

robhoes: This is true, but right now, when `update_dependencies` runs, it doesn't yet know which PCIs are VFs. Then we'd need to reorder things and run `update_pf_vf_relations` before `update_dependencies`.

I think that we can assume that VFs never have dependencies.

robhoes:
I see that you've made #3559 for this now, so I'll close this.

PR #3559, opened 2018-04-12
CA-286165: Only non-VF PCI need to update dependencies

xiewei20082008:


xiewei20082008:
Just Having test on this method. Also works.

robhoes:
This may work, but it does a `get_phyfn_path` twice for every PCI, which is a filesystem operation:  once in `update_pf_vf_relations` and once in `update_dependencies`.

PR #3560, opened 2018-04-12
CA-287343: Update HA failure tolerance plan for corosync/GFS2 and add…

thomasmck:
… unit tests

### DIFF
--- a/ocaml/tests/test_ha_vm_failover.ml
+++ b/ocaml/tests/test_ha_vm_failover.ml
@@ -58,6 +58,7 @@ type pool = {
   master: host;
   slaves: host list;
   ha_host_failures_to_tolerate: int64;
+  cluster: int;
 }
 
 let string_of_vm {memory; name_label} =
@@ -68,12 +69,13 @@ let string_of_host {memory_total; name_label; vms} =
     memory_total name_label
     (Test_printers.list string_of_vm vms)
 
-let string_of_pool {master; slaves; ha_host_failures_to_tolerate} =
+let string_of_pool {master; slaves; ha_host_failures_to_tolerate; cluster} =
   Printf.sprintf
-    "{master = %s; slaves = %s; ha_host_failures_to_tolerate = %Ld}"
+    "{master = %s; slaves = %s; ha_host_failures_to_tolerate = %Ld; cluster = %d}"
     (string_of_host master)
     (Test_printers.list string_of_host slaves)
     ha_host_failures_to_tolerate
+    cluster
 
 let load_vm ~__context ~(vm:vm) ~local_sr ~shared_sr ~local_net ~shared_net =
   let vm_ref = make_vm ~__context
@@ -120,10 +122,9 @@ let load_host ~__context ~host ~local_sr ~shared_sr ~local_net ~shared_net =
   in
   host_ref
 
-let setup ~__context {master; slaves; ha_host_failures_to_tolerate} =
+let setup ~__context {master; slaves; ha_host_failures_to_tolerate; cluster} =
   let shared_sr = make_sr ~__context ~shared:true () in
   let shared_net = make_network ~__context ~bridge:"xenbr0" () in
-
   (* Remove all hosts added by make_test_database *)
   List.iter (fun host -> Db.Host.destroy ~__context ~self:host) (Db.Host.get_all ~__context);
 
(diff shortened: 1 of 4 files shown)


ocaml/xapi/xapi_ha_vm_failover.ml
```diff
@@ -318,7 +318,12 @@ let compute_max_host_failures_to_tolerate ~__context ?live_set ?protected_vms ()
   let protected_vms = match protected_vms with
     | None -> all_protected_vms ~__context
     | Some vms -> vms in
-  let nhosts = List.length (Db.Host.get_all ~__context) in
+  let total_hosts = List.length (Db.Host.get_all ~__context) in
+  (* For corosync HA less than half of the pool can fail whilst maintaining quorum *)
+  let corosync_ha_max_hosts = ((total_hosts - 1) / 2) in
```
edwintorok: I think that for corosync we need to look at loosing only the enabled hosts. If you got 10 hosts, with 5 of them cleanly shutdown we can tolerate the loss of 2 additional hosts, so max failures in that case is 7.
So the failures to tolerate would be: `count_of_disabled_hosts (* these already "failed" *) + ((count_of_enabled_hosts - 1)/2)`.

Otherwise if you want to shutdown hosts (which implies increasing the host failures you tolerate if you're already at the max) you'll be told that you can't shutdown more hosts because you reached the limits of HA failures to tolerate.

Calculating this would need to be done while holding the clustering lock with `Xapi_clustering.with_clustering_lock_if_needed `, and every time we enable or disable a cluster host we will need to update the HA plan. To avoid deadlock the function in xapi_ha_failover.ml won't take any locks, but all its callers will have to (Xapi_cluster_host.enable/disable already does, locks need to be added to other callers).

ocaml/xapi/xapi_ha_vm_failover.ml
```diff
@@ -318,7 +318,12 @@ let compute_max_host_failures_to_tolerate ~__context ?live_set ?protected_vms ()
...
+  let nhosts = match Db.Cluster.get_all ~__context with
+    | [] -> total_hosts
+    | _ -> corosync_ha_max_hosts in
```
edwintorok: Could you make this call a function in Xapi_clustering.ml to calculate the failures to tolerate? As mentioned above it needs to know about enabled/disabled hosts.

ocaml/tests/test_ha_vm_failover.ml
```diff
@@ -495,11 +509,62 @@ module AssertNewVMPreservesHAPlan = Generic.Make(Generic.EncapsulateState(struct
...
+                                            let tests = [
+                                              (* Three host pool with no VMs. *)
+                                              (
+                                                {
+                                                  master = {memory_total = gib 256L; name_label = "master"; vms = []};
+                                                  slaves = [
+                                                    {memory_total = gib 256L; name_label = "slave1"; vms = []};
+                                                    {memory_total = gib 256L; name_label = "slave2"; vms = []}
+                                                  ];
+                                                  ha_host_failures_to_tolerate = 3L;
```
edwintorok: Is this right? I think corosync could only tolerate the loss of one host in a 3 host pool.

ocaml/tests/test_ha_vm_failover.ml
```diff
@@ -495,11 +509,62 @@ module AssertNewVMPreservesHAPlan = Generic.Make(Generic.EncapsulateState(struct
...
+                                                1
+                                              );
+                                              (* Two hosts pool with no VMs  *)
+                                              (
+                                                {
+                                                  master = {memory_total = gib 256L; name_label = "master"; vms = []};
+                                                  slaves = [
+                                                    {memory_total = gib 256L; name_label = "slave1"; vms = []}
+                                                  ];
+                                                  ha_host_failures_to_tolerate = 2L;
```
edwintorok: If your pool only has 2 hosts how can you loose them all? I think you need some actual VMs for the tests to make sense, indeed if you have no VMs you  might as well turn the whole pool off, but thats not a realistic scenario.

ocaml/xapi/xapi_ha_vm_failover.ml
thomasmck: So the check we need is to see if the host is down and clustering is disabled? or can we just check is clustering is disabled?

edwintorok: You can check whether the associated Cluster_host is enabled or disabled. There is a helper in `Xapi_clustering.is_clustering_disabled_on_host` that you can use.

thomasmck: ok, will need to make some additional updates to the unit test framework to get this to work because currently we don't enable clustering on all hosts

ocaml/tests/test_ha_vm_failover.ml
```diff
@@ -552,10 +560,24 @@ module ComputeMaxFailures = Generic.Make(Generic.EncapsulateState(struct
...
                                               );
+                                              (* Two host pool with one down  *)
+                                              (
+                                                {
+                                                  master = {memory_total = gib 256L; name_label = "master"; vms = []};
+                                                  slaves = [
+                                                    {memory_total = gib 256L; name_label = "slave1"; vms = []}
+                                                  ];
+                                                  ha_host_failures_to_tolerate = 2L;
+                                                  cluster = 1;
```
thomasmck: Updated so cluster now dictates how many hosts to enable clustering on

edwintorok:
I think this looks correct now, please add some more tests for larger pools (4,5 hosts) and squash the commits.

PR #3565, opened 2018-04-17
CA-244573 XenMotion fails after previously attempted SXM is interrupted by XAPI restart and vm goes into suspended state 

krizex:
This PR fixes 2 issues in 2 commits:
1. when recover SXM state from persistent files, we just recover the data of `Send_state` in case of wrong format of code of `try...with...`

2. When executing SXM, the time window between dp creation and persistent is large,
it is possible lead to dp leak if xapi restarted in the time window.
Aiming to mitigate the impact of this issue, this change persistents the dp
before we sending it to remote to shrink the time window.

### DIFF
--- a/ocaml/xapi/storage_migrate.ml
+++ b/ocaml/xapi/storage_migrate.ml
@@ -45,14 +45,18 @@ module State = struct
   end
 
   module Send_state = struct
+    type remote_info = {
+      dp: dp;
+      vdi: vdi;
+      url: string
+    } [@@deriving rpc]
+
     type t = {
       url : string;
       dest_sr : sr;
-      remote_dp : dp;
+      remote_info: remote_info option;
       local_dp : dp;
-      mirror_vdi : vdi;
-      remote_url : string;
-      tapdev : Tapctl.tapdev;
+      tapdev : Tapctl.tapdev option;
       mutable failed : bool;
       mutable watchdog : Scheduler.handle option;
     } [@@deriving rpc]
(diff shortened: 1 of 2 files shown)


ocaml/xapi/storage_migrate.ml
```diff
@@ -423,6 +426,20 @@ let start' ~task ~dbg ~sr ~vdi ~dp ~url ~dest =
 
   let id = State.mirror_id_of (sr,local_vdi.vdi) in
 
+  debug "Adding to active local mirrors before sending: id=%s" id;
+  let alm = State.Send_state.({
+      url;
+      dest_sr=dest;
+      remote_dp="";
```
jonludlam: I don't like the idea of using empty strings as a signal that we've not finished setting up - I'd rather use option types here. Perhaps the values that are set in the 'Replacing active local mirrors' section below could all be a record that's an optional field in Send_state?

ocaml/xapi/storage_migrate.ml
```diff
@@ -130,10 +130,10 @@ module State = struct
         Hashtbl.iter (Hashtbl.replace table) (hashtbl_of_rpc ~of_rpc:Copy_state.t_of_rpc rpc))
 
   let load () =
-    try load_one (Send_table active_send) with _ -> ();
-      try load_one (Recv_table active_recv) with _ -> ();
-        try load_one (Copy_table active_copy) with _ -> ();
-          loaded := true
+    ignore_exn (fun () -> load_one (Send_table active_send));
+    ignore_exn (fun () -> load_one (Recv_table active_recv));
```
jonludlam: Argh! good catch here! This sort of problem is really annoying.

ocaml/xapi/storage_migrate.ml
krizex: Yes, I have putted them into an `option` object, and I think the tapdev should be an individual `option` object.

krizex:
Updated. Regression , BST & BVT has passed based on previous changes.
@jonludlam please help review it again, please **approve** if you don't have any further. However don't merge it, let's wait for the regression and BST&BVT results.

ocaml/xapi/storage_migrate.ml
```diff
@@ -586,7 +610,11 @@ let stat ~dbg ~id =
     | (Some receive_state, _, _) ->
       receive_state.Receive_state.remote_vdi, receive_state.Receive_state.leaf_vdi
     | (_, Some send_state, _) ->
-      snd (of_mirror_id id), send_state.Send_state.mirror_vdi
+      let dst_vdi = match send_state.Send_state.remote_info with
+        | Some remote_info -> remote_info.Send_state.vdi
+        | None -> ""
```
jonludlam: This doesn't look right, but I think that's more a problem with SMAPIv2 than your code.

jonludlam:
Looks good to me now from a functional point of view.

ocaml/xapi/storage_migrate.ml
```diff
@@ -734,22 +762,25 @@ let pre_deactivate_hook ~dbg ~dp ~sr ~vdi =
...
+            then (Thread.delay 1.0; wait ())
+            else st
+          in
+          let st = wait () in
+          debug "Got final stats after waiting %f seconds" (get_delta ());
+          if st.Stats.nbd_mirror_failed = 1
+          then begin
+            error "tapdisk reports mirroring failed";
+            s.failed <- true
+          end;
```
jonludlam: Approved.

PR #3568, opened 2018-04-18
Port some Xapi unit tests from OUnit to Alcotest framework

minishrink:
Ported a few simpler tests while waiting around for a build. We should make a list of which would be the most useful tests to port somewhere.

### DIFF
--- a/ocaml/tests/test_network.ml
+++ b/ocaml/tests/test_network.ml
@@ -1,138 +1,132 @@
 
-let test_purpose_setters =
-  let assert_equal ~msg p1 p2 = Ounit_comparators.NetworkPurposeSet.(assert_equal ~msg (of_list p1) (of_list p2)) in
+let assert_equal msg p1 p2 =
+  let network_purpose = Alcotest_comparators.from_rpc_of_t API.rpc_of_network_purpose in
+  Alcotest.(check (slist network_purpose compare))
+    msg
+    p1 p2
 
-  let with_test f =
-    let __context = Mock.make_context_with_new_db "Mock context" in
-    let network ~purpose = Test_common.make_network ~__context ~purpose () in
-    let add_purpose = Xapi_network.add_purpose ~__context in
-    let remove_purpose = Xapi_network.remove_purpose ~__context in
-    let get_purpose = Db.Network.get_purpose ~__context in
-    f network add_purpose remove_purpose get_purpose
-  in
+let with_test f =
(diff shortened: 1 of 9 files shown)


ocaml/tests/test_network.ml
```diff
@@ -1,138 +1,132 @@
...
-  in
-
-  let ((>:::), (>::)) = OUnit.((>:::), (>::)) in
-  "test_purpose_setters" >:::
-  [ "test_add_purpose" >:: test_add_purpose
-  ; "test_remove_purpose" >:: test_remove_purpose
-  ]
+let assert_equal msg p1 p2 =
+  let network_purpose = Alcotest_comparators.from_rpc_of_t API.rpc_of_network_purpose in
+  Alcotest.(check (list network_purpose))
```
gaborigloi: We need `slist` here because it's conceptually a set, the order doesn't matter.

ocaml/tests/test_xapi_db_upgrade.ml
```diff
@@ -42,88 +44,112 @@ let upgrade_vm_memory_for_dmc () =
...
 
 let upgrade_bios () =
 
   let check inventory bios_strings =
-    Unixext.mkdir_safe "/var/tmp" 0o755;
-    Unixext.write_string_to_file "/var/tmp/.previousInventory" inventory;
-    let __context = make_test_database () in
-    upgrade_bios_strings.fn ~__context;
+    Stdext.Unixext.mkdir_safe "/var/tmp" 0o755;
+    Stdext.Unixext.write_string_to_file "/var/tmp/.previousInventory" inventory;
```
gaborigloi: It's better to reference the `Xapi_stdext_unix` module directly, that way we don't have to depend on the whole of stdext.

ocaml/tests/test_workload_balancing.ml
```diff
@@ -12,18 +12,20 @@
...
-  let assert_succeed url host port =
-    assert_equal (split url) (host, port)
-  in
-  let assert_raise_url_invalid url =
-    assert_raises_api_error Api_errors.wlb_url_invalid ~args:[url] (fun () -> split url)
-  in
+let assert_succeed url host port =
+  Alcotest.(check (pair string int))
+    "test_split_host_port"
+    (split url) (host, port)
```
gaborigloi: The expected value should go first here

PR #3569, opened 2018-04-18
CA-284520: Adding pool pre-check for clustering enabled

thomasmck:

### DIFF
--- a/ocaml/xapi/xapi_pool.ml
+++ b/ocaml/xapi/xapi_pool.ml
@@ -70,6 +70,17 @@ let pre_join_checks ~__context ~rpc ~session_id ~force =
       raise (Api_errors.Server_error(Api_errors.ha_is_enabled, []));
     end in
 
+  (* I cannot join a Pool if I have Clustering enabled on me *)
+  let clustering_is_not_enabled_on_me () =
+    let host = Helpers.get_localhost ~__context in
+    match Xapi_clustering.find_cluster_host ~__context ~host with
+      | None -> ()
+      | Some cluster_host -> 
+        if (Db.Cluster_host.get_enabled ~__context ~self:cluster_host) then begin
+          error "Cannot join pool as Clustering is enabled";
+          raise (Api_errors.(Server_error( clustering_enabled, [Ref.string_of cluster_host])));
+    end in 
+
   (* CA-26975: Pool edition MUST match *)
   let assert_restrictions_match () =
     let my_edition = Db.Host.get_edition ~__context ~self:(Helpers.get_localhost ~__context) in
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_pool.ml
```diff
@@ -70,6 +70,14 @@ let pre_join_checks ~__context ~rpc ~session_id ~force =
       raise (Api_errors.Server_error(Api_errors.ha_is_enabled, []));
     end in
 
+  (* I cannot join a Pool if I have Clustering enabled on me *)
+  let clustering_is_not_enabled_on_me () =
+    let host = Helpers.get_localhost ~__context in
+    if not (Xapi_clustering.is_clustering_disabled_on_host ~__context host) then begin
+      error "Cannot join pool as Clustering is enabled";
+      raise (Api_errors.Server_error(Api_errors.clustering_enabled, []));
```
gaborigloi: Strictly speaking this error has a cluster_host parameter (in ocaml/idl/datamodel_errors.ml):
```ocaml
1127   error Api_errors.clustering_enabled ["cluster_host"]                                                                                                                      
1128     ~doc:"An operation was attempted while clustering was enabled on the cluster_host." ();
```
But it's not important as supplying that would require further changes.

minishrink: Minor suggestion: can just raise `Api_errors.(Server_error (clustering_enabled, [])) so you don't need the second `Api_errors`.

Also, I think it's worth providing the cluster_host parameter in the error message, though if you do that, you don't need to call `is_clustering_disabled_on_host` and can instead directly check the cluster_host's `enabled` field.

thomasmck: I don't think there is much value in passing the cluster_host as the main way people will be disabling this is through XC which doesn't require this knowledge (Only way to do this is to untick "clustering_enabled" for the pool which then disables it for all hosts).

robhoes: If there errors are defined with arguments in the datamodel, then we should return those arguments, because clients (through the API bindings) may expect them.

thomasmck: ok. Will make the update.

gaborigloi:
However, in this case we may or may not have a cluster_host, and that function works on a cluster host and not a host.

thomasmck:
Spoke with Aki about this yesterday who suggested that we could do 
```match Xapi_cluster_host.find_cluster_host with
| Some cluster_host -> Xapi_clustering.assert_cluster_host_enabled ~expected:false
| None -> ()```
However that we mean we lose out on the `error "Cannot join pool as Clustering is enabled";` line. Does that matter?

thomasmck:
I think its better to stick to the old format for raising the error log. Doesn't require many additional lines anyway.

edwintorok:
The last commit should be squashed, or a ticket number added.

PR #3570, opened 2018-04-18
Reduce amount of deprecation warnings when compiling xapi

mseri:
This will not solve all the warnings, but it moves slowly in the direction that we need to make xapi safe-string compliant and warning free

### DIFF
--- a/ocaml/idl/ocaml_backend/gen_rbac.ml
+++ b/ocaml/idl/ocaml_backend/gen_rbac.ml
@@ -65,12 +65,12 @@ let hash2uuid str =
   in
   Uuid.string_of_uuid (Uuid.uuid_of_int_array (int_array hex))
 
-let replace_char _str c1 c2 =
-  let str = String.copy _str in (*defensive copy*)
-  for i=0 to String.length str -1 do
-    if str.[i]=c1 then str.[i]<-c2 else ()
-  done;
-  str
+let replace_char str c1 c2 =
+  let buf = Bytes.of_string str in (*defensive copy*)
+  String.iteri (fun i _ ->
+    if str.[i]=c1 then Bytes.set buf i c2 else ()
+  ) str;
+  Bytes.unsafe_to_string buf
 
 let role_uuid name = hash2uuid name
 
--- a/ocaml/xapi-cli-protocol/cli_protocol.ml
+++ b/ocaml/xapi-cli-protocol/cli_protocol.ml
@@ -98,12 +98,12 @@ let marshal_int32 x =
   and b = (x >> 8) && 0xffl
   and c = (x >> 16) && 0xffl
   and d = (x >> 24) && 0xffl in
-  let result = String.make 4 '\000' in
-  result.[0] <- char_of_int (Int32.to_int a);
-  result.[1] <- char_of_int (Int32.to_int b);
-  result.[2] <- char_of_int (Int32.to_int c);
-  result.[3] <- char_of_int (Int32.to_int d);
-  result
+  let set buf pos v =
+    Bytes.set buf pos (char_of_int @@ Int32.to_int v)
+  in
+  let result = Bytes.make 4 '\000' in
+  List.iteri (set result) [a;b;c;d];
+  Bytes.unsafe_to_string result
 
 let marshal_int x = marshal_int32 (Int32.of_int x)
 
--- a/ocaml/xapi/cli_progress_bar.ml
+++ b/ocaml/xapi/cli_progress_bar.ml
@@ -22,7 +22,7 @@ module Make(T: Floatable) = struct
     max_value: T.t;
     mutable current_value: T.t;
     width: int;
-    line: string;
+    mutable line: string;
     mutable spin_index: int;
     start_time: float;
     mutable summarised: bool;
@@ -62,11 +62,13 @@ module Make(T: Floatable) = struct
 
   let string_of_bar t =
     let w = bar_width t t.current_value in
-    t.line.[1] <- spinner.(t.spin_index);
+    let line = Bytes.of_string t.line in
+    Bytes.set line 1 spinner.(t.spin_index);
     t.spin_index <- (t.spin_index + 1) mod (Array.length spinner);
     for i = 0 to w - 1 do
-      t.line.[prefix + i] <- (if i = w - 1 then '>' else '#')
+      Bytes.set line (prefix + i) (if i = w - 1 then '>' else '#')
     done;
+    t.line <- Bytes.unsafe_to_string line;
     let percent = Printf.sprintf "%3d" (percent t) in
     String.blit percent 0 t.line (t.width - 19) 3;
     let eta = eta t in
--- a/ocaml/xapi/sparse_encoding.ml
+++ b/ocaml/xapi/sparse_encoding.ml
@@ -51,16 +51,11 @@ module Marshal = struct
     and f = (x >>> 40) &&& 0xffL
     and g = (x >>> 48) &&& 0xffL
     and h = (x >>> 56) &&& 0xffL in
-    let result = String.make 8 '\000' in
-    result.[0] <- char_of_int (Int64.to_int a);
-    result.[1] <- char_of_int (Int64.to_int b);
-    result.[2] <- char_of_int (Int64.to_int c);
-    result.[3] <- char_of_int (Int64.to_int d);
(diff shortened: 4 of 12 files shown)


minli1:
Very nice to see much fewer warnings when compiling 👍

ocaml/idl/ocaml_backend/gen_rbac.ml
```diff
@@ -65,12 +65,12 @@ let hash2uuid str =
...
-  for i=0 to String.length str -1 do
-    if str.[i]=c1 then str.[i]<-c2 else ()
-  done;
-  str
+let replace_char str c1 c2 =
+  let buf = Bytes.of_string str in (*defensive copy*)
+  String.iteri (fun i _ ->
+    if str.[i]=c1 then Bytes.set buf i c2 else ()
+  ) str;
+  Bytes.unsafe_to_string buf
```
gaborigloi: I think we could replace this with `Astring.String.map` - performance probably doesn't matter in the autogenerator:
http://erratique.ch/software/astring/doc/Astring.String.html

ocaml/xapi-cli-protocol/cli_protocol.ml
```diff
@@ -213,7 +213,7 @@ let marshal_message = function
   | Blob x     -> marshal_int 11 ^ (marshal_blob_header x)
 
 let write_string (fd: Unix.file_descr) buf =
-  Unixext.really_write fd buf 0 (String.length buf)
+  Unix.write fd buf 0 (String.length buf) |> ignore
```
gaborigloi: `Return the number of bytes actually written. write repeats the writing operation until all bytes have been written or an error occurs.`

ocaml/xapi/cli_progress_bar.ml
```diff
@@ -22,7 +22,7 @@ module Make(T: Floatable) = struct
     max_value: T.t;
     mutable current_value: T.t;
     width: int;
-    line: string;
+    mutable line: string;
```
gaborigloi: Oh so for `Bytes.set`, we have to make this mutable?

gaborigloi:
Yes, IMO. But the more reviews, the better.

ocaml/xapi/sparse_encoding.ml
```diff
@@ -51,29 +51,23 @@ module Marshal = struct
...
+    ) [a; b; c; d; e; f; g; h];
+    Bytes.unsafe_to_string result
   let int32 x =
     let (>>>) a b = Int32.shift_right_logical a b
     and (&&&) a b = Int32.logand a b in
     let a = (x >>> 0) &&& 0xffl
     and b = (x >>> 8) &&& 0xffl
     and c = (x >>> 16) &&& 0xffl
     and d = (x >>> 24) &&& 0xffl in
-    let result = String.make 4 '\000' in
```
minishrink: Since you're ANDing them all with the same value, why not include that in the definition of `>>>`?

gaborigloi: I would find that confusing, because `>` is usually the left-shift operator.

minishrink: With a different operator of course

ocaml/xapi-cli-protocol/cli_protocol.ml
robhoes: I suspect that the `really_write` function comes from https://ocaml.github.io/ocamlunix/ocamlunix.html (Section 5.7). Has anything changed in `Unix.write` such that `really_write` isn't useful anymore?

gaborigloi: Not sure, we discussed with @mseri a while ago that `Unix.write` already does what we need, it writes everything according to the docs, unlike the `write` Linux command. So the it was assumed that `Unix.write` behaves like `write(2)`, and that's why `really_write` was created.

robhoes: See that link I posted. `Unix.write` does repeated `write()` system calls, but pipes and sockets are a bit special:

> The function write of the Unix module iterates the system call write until all the requested bytes are effectively written.
> `val write : file_descr -> string -> int -> int -> int`
> However, when the descriptor is a pipe (or a socket, see chapter 6), writes may block and the system call write may be interrupted by a signal. In this case the OCaml call to Unix.write is interrupted and the error EINTR is raised. The problem is that some of the data may already have been written by a previous system call to write but the actual size that was transferred is unknown and lost. This renders the function  write of the Unix module useless in the presence of signals.

gaborigloi: We have actually replaced `really_write` with `Unix.write`:
https://github.com/xapi-project/stdext/blob/001b4b7386e7a411de307825a19b3fe4c51a2770/lib/xapi-stdext-unix/unixext.ml#L448

robhoes: 😮

gaborigloi: But as you said, the implementation of `really_write` was equivalent to what `Unix.write` does, because it also just made repeated calls, and did not check the EINTR signal, so replacing it is probably harmless:
https://github.com/xapi-project/stdext/commit/1d501c6d85988019a96a5f219360a6f8acac2a60#diff-b9e2b89466fcaa1db5f80ce7b15fdf62L465
```difff
 (* Ideally, really_write would be implemented with optional arguments ?(off=0) ?(len=String.length string) *)
 let really_write_string fd string =
-  really_write fd string 0 (String.length string)
+  let payload = Bytes.unsafe_of_string string in
+  Unix.write fd payload 0 (Bytes.length payload) |> ignore
```

robhoes: The original implementation was:
```
let really_write fd string off n =
  let written = ref 0 in
  while !written < n
  do
    let wr = Unix.write fd string (off + !written) (n - !written) in
    written := wr + !written
  done
```
...which doesn't actually handle EINTR at all, so doesn't make much sense...

robhoes: A better version would use `Unix.single_write`.

robhoes: @jonludlam 🔼

gaborigloi: Yeah. I think it's fine to replace `really_write` with `Unix.write` though, because their behaviour is the same IMO :) - they both behave incorrectly in case of EINTR

robhoes: As on that website:
```
let rec restart_on_EINTR f x =
  try f x with Unix_error (EINTR, _, _) -> restart_on_EINTR f x
in
let rec really_write fd buffer offset len =
  let n = restart_on_EINTR (single_write fd buffer offset) len in
  if n < len then really_write fd buffer (offset + n) (len - n);;
```

robhoes: Well, if our `Unixext.really_write` was incorrect, than why don't we just fix it rather than replacing all its users with `Unix.write`?

gaborigloi: I think we've done a `really_write` -> `Unix.write` replacement in a past PR already, so we should undo that too, not sure which one, maybe @mseri remembers

mseri: @robhoes thanks for the rationale behind `really_write`! When I did deprecate it in stdext nobody could tell me why we had that broken implementation, I'm happy to fix it as you suggest and revert the deprecation and changes.

I think we can figure out where we did the change and revert it once we fix the `really_write`, it's likely one of the handful PR on moving to safe-strings, and in any case it was equivalent to what it had done for the last 10 years so it should not be a problem if we miss one.

Let's not merge this yet, I'll revert the Unixext changes and then send a PR to fix Stdext

mseri:
Yes, I agree. Thanks for digging out the rationale behind the old function

ocaml/xapi/sparse_encoding.ml
mseri: In this case I think it's better to keep the old code and stick to operators that look like the usual ones

mseri:
I can squash the reverts once this PR is approved

PR #3573, opened 2018-04-19
CA-288347: Do not take clustering lock in SR.probe

gaborigloi:
Remove the clustering lock in SR.probe - now the plugin no longer tries
to mount the fs, so it's not needed any more.


gaborigloi:
PR for the lcm branch: https://github.com/xapi-project/xen-api/pull/3572

PR #3575, opened 2018-04-20
CA-288312: Don't update HOSTNAME in /etc/sysconfig/network

robhoes:
This is no longer required. The following call to `hostnamectl` sets the
current hostname, and also updates `/etc/hostname` to persist it.


robhoes:
Same as #3574, but for 1.90-lcm.

PR #3577, opened 2018-04-23
Remove inactive with_vm quicktests

gaborigloi:
These quicktests are looking for a template that does not exist
currently, therefore they are always skipped.


minishrink:
Why not provide an existing template instead? Or would that not be possible with the current state of the test?

ocaml/quicktest/quicktest.ml
```diff
@@ -796,37 +755,6 @@ let async_test session_id =
 let make_vif ~session_id ~vM ~network ~device =
   Client.VIF.create ~rpc:!rpc ~session_id ~vM ~network ~mTU:0L ~mAC:"" ~device ~other_config:["promiscuous", "on"; "mtu", "1400"] ~qos_algorithm_type:"" ~qos_algorithm_params:[]
 
-let with_vm s f =
-  try
-    let (_: API.ref_VM) = find_template s vm_template in
```
mseri: If you stick with this, you should probably also remove `vm_template` from `quicktest_common`. What if instead we select the first debian or centos template from the list of available templates?

gaborigloi: I'm doing a quicktest rewrite at the moment, and for that it will be more convenient to just remove these temporarily. So I'll remove these in my quicktest rewrite PR, and I'll make a ticket to re-enable them.

PR #3578, opened 2018-04-23
CA-273986: Test Cluster.create cleans up refs when it fails

minishrink:
`Cluster.create` creates a cluster ref and a cluster_host ref. This PR ensures those refs are destroyed in the event of a failed `Cluster.create`. and adds a test to check this.

### DIFF
--- a/ocaml/tests/test_cluster.ml
+++ b/ocaml/tests/test_cluster.ml
@@ -39,7 +39,7 @@ let test_rpc ~__context call =
   | name, params ->
     failwith (Printf.sprintf "Unexpected RPC: %s(%s)" name (String.concat " " (List.map Rpc.to_string params)))
 
-let create_cluster ~__context ?(cluster_stack=Constants.default_smapiv3_cluster_stack) () =
+let create_cluster ~__context ?(cluster_stack=Constants.default_smapiv3_cluster_stack) ?(test_clusterd_rpc=test_clusterd_rpc) () =
   Context.set_test_rpc __context (test_rpc ~__context);
   Context.set_test_clusterd_rpc __context (test_clusterd_rpc ~__context);
   let network = Test_common.make_network ~__context () in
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_cluster.ml
```diff
@@ -71,7 +71,11 @@ let create ~__context ~network ~cluster_stack ~pool_auto_join ~token_timeout ~to
         set_ha_cluster_stack ~__context;
         cluster_ref
       | Result.Error error ->
-        D.warn "Error occurred during Cluster.create";
+        D.warn "Error occurred during Cluster.create. Cleaning up...";
+        Db.Cluster_host.destroy ~__context ~self:cluster_host_ref;
+        D.debug "Cluster.create cleanup: Destroyed cluster_host";
```
gaborigloi: I think we can omit these two debug lines, as these are just DB operations

ocaml/xapi/xapi_cluster.ml
```diff
@@ -71,7 +71,11 @@ let create ~__context ~network ~cluster_stack ~pool_auto_join ~token_timeout ~to
         set_ha_cluster_stack ~__context;
         cluster_ref
       | Result.Error error ->
-        D.warn "Error occurred during Cluster.create";
+        D.warn "Error occurred during Cluster.create. Cleaning up...";
+        Db.Cluster_host.destroy ~__context ~self:cluster_host_ref;
```
gaborigloi: Does this call fail if the ref isn't in the DB yet? In that case we should ignore exceptions that are raised.

ocaml/xapi/xapi_cluster.ml
```diff
@@ -71,7 +71,11 @@ let create ~__context ~network ~cluster_stack ~pool_auto_join ~token_timeout ~to
         set_ha_cluster_stack ~__context;
         cluster_ref
       | Result.Error error ->
```
gaborigloi: I think if we are in this match case, the refs aren't in the DB yet, because they are put into the DB in case of `Result.Ok`. So we don't have to clean them up.

edwintorok:
@gaborigloi can we merge this unit test? the PR is marked as changes requested, but I think the changes were done.

gaborigloi:
Yes we can, thanks for the heads-up, I missed the changes.

PR #3579, opened 2018-04-23
CA-288411: make the cache work after a toolstack restart

edwintorok:
This fixes a regression introduced in the fix for XOP-830,
where `add_caches` got dropped from the startup resynchronization logic.

This resulted in periodic writes to the database, which are especially
bad with redo log (or implicitly with HA) turned on.

### DIFF
--- a/ocaml/xapi/xapi_xenops.ml
+++ b/ocaml/xapi/xapi_xenops.ml
@@ -2261,6 +2261,12 @@ let resync_resident_on ~__context =
       Db.VM.set_resident_on ~__context ~self:vm ~value:localhost)
       xapi_thinks_are_nowhere;
 
+  List.iter (fun ((id, state), _queue_name) ->
+      match xenapi_of_xenops_power_state (Some state.Vm.power_state) with
+      | `Running | `Paused -> add_caches id;
+      | _ -> ()
+  ) xenopsd_vms_in_xapi;
+
   (* Sync VM state in Xapi for VMs not running on this host *)
   List.iter (fun (id, vm) ->
       info "VM %s was marked as resident here in the DB but isn't known to xenopsd. Resetting in DB" id;


ocaml/xapi/xapi_xenops.ml
```diff
@@ -2261,6 +2261,14 @@ let resync_resident_on ~__context =
       Db.VM.set_resident_on ~__context ~self:vm ~value:localhost)
       xapi_thinks_are_nowhere;
 
+  List.iter (fun ((id, state), queue_name) ->
+    let xapi_power_state =
+      xenapi_of_xenops_power_state (Some state.Vm.power_state) in
+    match xapi_power_state with
+      | `Running | `Paused -> add_caches id;
+      | `Suspended | `Halted -> ()
```
lindig: To avoid a runtime match failure, maybe use a wildcard instead? Or introduce another explicit match `| _ -> assert false`.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -2261,6 +2261,14 @@ let resync_resident_on ~__context =
       Db.VM.set_resident_on ~__context ~self:vm ~value:localhost)
       xapi_thinks_are_nowhere;
 
+  List.iter (fun ((id, state), queue_name) ->
```
lindig: `queue_name` is not used - so maybe use `_queue_name` to indicate that.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -2261,6 +2261,14 @@ let resync_resident_on ~__context =
...
+    let xapi_power_state =
```
lindig: This could be written as:

```
xenapi_of_xenops_power_state (Some state.Vm.power_state)
|> function 
| `Running | `Paused -> add_caches id
| _other -> ()
```

But I'm not sure it is better.

edwintorok: I agree that `xapi_power_state` should be eliminated, how about this:
```
match xenapi_of_xenops_power_state (Some state.Vm.power_state) with
| `Running | `Paused -> add_caches id;
| _ -> ()
```

ocaml/xapi/xapi_xenops.ml
edwintorok: I think in this case the match is exhaustive, if I remove the second line I get a build failure complaining I forgot those 2 cases. Unfortunately `_ -> .` does not work with polymorphic variants (to ensure that the _ can never be reached).
I'll change it to `_`, its safer that way.

edwintorok:
Thanks, force pushed some minor changes to address the comments (the patch is so small that I think using fixup commits wouldn't have helped readability).

robhoes:
Let's do this.

PR #3580, opened 2018-04-24
CP-27880, CA-288635: Reduce redo log traffic

edwintorok:
Analyzing the block_device_io `strace` logs noticed that:
 * the master DB is flushed in chunks of 16KiB on a file opened with O_DSYNC taking 1.68s for a 15MiB DB, the chunksize is too small. Note that ocaml runtime internally limits maximum size of a single write to 64KiB.
 * delta writes are usually twice as slow as need to be due to extra write of 16 zeros ("trample size") as a separate write call instead of concatenating with previous one (there is no buffering with O_DSYNC). In some cases writing those 16 bytes takes longer than writing the delta itself, depending how busy the disk is.
 * fields are updated with same value multiple times:

 Writing to the redo log has to be persistent, so the use of O_DSYNC is correct, however spinning disks can barely do more than 100 writes/s this way, so it is important that we buffer up as much data as possible and only wait for the write to finish when we really have to. (another possibility would be to open normally and call fdatasync, but that is not available in the Unix module).

Also there are some useful stats collected by xapi that are not part of a bugtool (`xe diagnostic-db-stats`, `xe diagnostic-timing-stats`, `xe diagnostic-net-stats`), if we collect them at runtime they should be in the bugtool. There were no stats collected for the redo log activity, so fix that.

I have dev-tested these changes on my dtbox, looking at the output with strace, and collecting a bugtool. I am preparing a jenkins build to do wider testing.


edwintorok:
SR83968 looks good, one test failed due to a bug in the testcase itself ("Trying to lock more luns than specified in SHAREDRESOURCES") and also affects master, I think this PR should be good to merge.

PR #3581, opened 2018-04-25
 CA-287921: nbd_client_manager: track nbd device -> nbd server mapping 

gaborigloi:
Merge with https://github.com/xapi-project/vhd-tool/pull/56

Keep track of the mapping of the connected /dev/nbdX NBD devices to the
Unix domain socket of the server and the exportname used for the
connection. This allows sparse_dd to look up the NBD connection details
based on the NBD device it receives, then it can extract the list of
allocated extents to copy from the NBD server using the BLOCK_STATUS NBD
extension.

Also, add randomized quicktest to check VDI.copy data integrity

### DIFF
--- a/ocaml/quicktest/quicktest_vdi_copy_data_integrity.ml
+++ b/ocaml/quicktest/quicktest_vdi_copy_data_integrity.ml
@@ -72,6 +72,16 @@ let with_open_vdi rpc session_id vdi mode f =
    Integrity tests
  * ---------------- *)
 
+let random_bytes length =
+  let buf = Bytes.create length in
+  let f = open_in "/dev/urandom" in
+  finally
+    (fun () ->
+       really_input f buf 0 length;
+       buf
+    )
+    (fun () -> close_in f)
+
 let write_random_data rpc session_id vdi =
   let size = Client.Client.VDI.get_virtual_size ~rpc ~session_id ~self:vdi in
   with_open_vdi rpc session_id vdi `RW
@@ -80,11 +90,8 @@ let write_random_data rpc session_id vdi =
        let max_writes = 100 in
        let max_write_length = 2048 in
 
-       let write_random_char_block offset length =
-         let buf =
-           let c = Random.int 256 |> Char.chr in
-           Bytes.make length c
-         in
+       let write_random_block offset length =
+         let buf = random_bytes length in
          assert (Unix.LargeFile.lseek fd offset Unix.SEEK_SET = offset);
          Unix.write fd buf 0 length |> ignore
        in
@@ -97,7 +104,7 @@ let write_random_data rpc session_id vdi =
          in
          let length = Random.int max_extent_length in
 
-         write_random_char_block offset length
+         write_random_block offset length
        in
 
        let num_extents = Random.int max_writes in
@@ -113,7 +120,7 @@ let fill rpc session_id vdi =
   in
   with_open_vdi rpc session_id vdi `RW
     (fun fd ->
-       let buf = Bytes.make size 'a' in
+       let buf = random_bytes size in
        Unix.write fd buf 0 size
     )
 
--- a/ocaml/quicktest/quicktest_vdi_copy_data_integrity.ml
+++ b/ocaml/quicktest/quicktest_vdi_copy_data_integrity.ml
@@ -1,0 +1,181 @@
+
+(* ---------------- *
+    Helper functions
+ * ---------------- *)
+
+let with_test test f =
+  Quicktest_common.start test;
+  try
+    f ();
+    Quicktest_common.success test
+  with e ->
+    Quicktest_common.failed test (Printexc.to_string e);
+    raise e
+
+let finally f cleanup =
+  let result =
+    try
+      f ()
+    with e -> begin
+        cleanup ();
+        raise e
+      end
+  in
+  cleanup ();
+  result
+
+let get_domain_zero rpc session_id =
+  Xapi_inventory.inventory_filename := "/etc/xensource-inventory";
+  let uuid = Xapi_inventory.lookup Xapi_inventory._control_domain_uuid in
+  Client.Client.VM.get_by_uuid ~rpc ~session_id ~uuid
+
+let with_attached_vdi rpc session_id vdi mode f =
+  let dom0 = get_domain_zero rpc session_id in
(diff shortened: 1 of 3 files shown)


gaborigloi:
Merge with https://github.com/xapi-project/vhd-tool/pull/56

ocaml/quicktest/quicktest_vdi_copy_data_integrity.ml
```diff
@@ -0,0 +1,181 @@
...
+let fill rpc session_id vdi =
+  let size =
+    Client.Client.VDI.get_virtual_size ~rpc ~session_id ~self:vdi
+    |> Int64.to_int
+  in
+  with_open_vdi rpc session_id vdi `RW
+    (fun fd ->
+       let buf = Bytes.make size 'a' in
+       Unix.write fd buf 0 size
+    )
```
gaborigloi: Probably I should fill the VDI with more random data, instead of using the same char everywhere. I just couldn't find a library that gives me a random `bytes` of a given length

ocaml/quicktest/quicktest_vdi_copy_data_integrity.ml
```diff
@@ -0,0 +1,181 @@
...
+  with_open_vdi rpc session_id vdi `RW
+    (fun fd ->
+
+       let max_writes = 100 in
+       let max_write_length = 2048 in
+
+       let write_random_char_block offset length =
+         let buf =
+           let c = Random.int 256 |> Char.chr in
+           Bytes.make length c
```
gaborigloi: Same comment as below - ideally this would be more random than just using the same char.

ocaml/quicktest/quicktest_vdi_copy_data_integrity.ml
mseri: I know only https://github.com/xavierleroy/cryptokit/blob/master/src/cryptokit.mli#L241

gaborigloi: Or we could copy from `/dev/urandom`?

mseri: If you look at the various implementations, that's what one of the options in cryptokit does :D
I think we can even use the Random module and fill the buffer.

edwintorok: Generating random numbers with a PRNG can be faster than `/dev/urandom` (especially if you just use something fast like a shift register, worse quality randomness but fast), however using `String.init` with `Random.int` is quite slow:
```
String.init Random: 0.225s
/dev/urandom: 0.045s
```

```
let time msg f =
  let t0 = Unix.gettimeofday () in
  ignore(f ());
  let t1 = Unix.gettimeofday () in
  Printf.printf "%s: %.3fs\n" msg (t1 -. t0)

let () =
  Random.self_init ();
  let n = 10_000_000 in

  time "String.init Random" (fun () ->
      String.init n (fun _ -> Char.chr (Random.int 255)));

  time "/dev/urandom" (fun () ->
      let f = open_in "/dev/urandom" in
      let r = really_input_string f n in
      close_in f;
      r)
```

A very fast and simple PRNG with reasonable quality is 2CMRES from here:
http://www.drdobbs.com/tools/fast-high-quality-parallel-random-number/231000484?pgno=2
This one should also be quite fast: http://www.pcg-random.org/
As long as you don't need it for cryptographic purposes either of these algorithms should work well.

gaborigloi: Thanks for the investigation, I think I'll go for `/dev/urandom` for now. There is also [mirage-random](https://github.com/mirage/mirage-random) in xs-opam, but it uses `Random.int`.

ocaml/quicktest/quicktest_vdi_copy_data_integrity.ml
```diff
@@ -0,0 +1,188 @@
...
+      check_vdi_unchanged test_nodata rpc session_id sr
+        ~prepare_vdi:(fun _vdi -> ());
+
+      let test_random = Quicktest_common.make_test "Copy of random VDI" 6 in
+      check_vdi_unchanged test_random rpc session_id sr
+        ~prepare_vdi:(write_random_data rpc session_id);
+
+      let test_full = Quicktest_common.make_test "Copy of full VDI" 6 in
+      check_vdi_unchanged test_full rpc session_id sr
+        ~prepare_vdi:(fill rpc session_id)
```
edwintorok: This is a good start. Might be useful to explicitly cover a few edge cases: hole at beginning of file (multiple of 64k), hole at end of file (multiple of 64k), write some data, then write random multiple of 64k zeroes (don't know if qemu would convert the zeroes to sparse, it might).

Also how do snapshots and leaf-coalesce affect this functionality? Might be useful to add a few tests that take write some random data with holes, take a snapshot, modify the original vdi again, check export of original vdi/snapshotted vdi, then delete snapshot, and do some copies again while SMGC is running. Although if there are any bugs there they would be in qemu/SMGC, so perhaps such a test is better written as part of the pool stress test. Could you verify if there are any such tests in the pool stress, and if not create a ticket to add some?

gaborigloi: Why should we write multiples of 64K blocks in the test? In my opinion completely random block lengths / offsets cover more edge cases. I could add another test case though for writing zeroes into the VDI.
I agree that those more complex scenarios should be run by XenRT. I'll check what VDI.copy tests we have at the moment.

edwintorok: Right, I think that you need a minimum of 64k, not necesarily a multiple of 64k. So 64k+random would cover more edge cases true.

gaborigloi: Not sure how we could detect this from a quicktest, maybe by checking the `physical_utilization` field if that's reliable. But there is a XenRT test that was failing because the copy did not took advantage of sparseness.

edwintorok: The VDI size is 4M, hence there is a reasonable chance that we will have >64k in our test. More tests could be added in a separate PR.

PR #3583, opened 2018-04-30
 CP-27911: Alcotest unit test ports

minishrink:
Porting the remaining OUnit tests that don't use test_highlevel, as this will require a separate port.

### DIFF
--- a/ocaml/tests/test_xapi_vbd_helpers.ml
+++ b/ocaml/tests/test_xapi_vbd_helpers.ml
@@ -12,48 +12,44 @@
  * GNU Lesser General Public License for more details.
  *)
 
-open OUnit
-open Test_common
+module T = Test_common
 
 (* Helpers for testing Xapi_vbd_helpers.valid_operations *)
 
 let setup_test ~__context ~vdi_fun =
-  let _sm_ref = make_sm ~__context () in
-  let sr_ref = make_sr ~__context () in
-  let (_: API.ref_PBD) = make_pbd ~__context ~sR:sr_ref () in
-  let vm_ref = make_vm ~__context () in
+  let _sm_ref = T.make_sm ~__context () in
+  let sr_ref = T.make_sr ~__context () in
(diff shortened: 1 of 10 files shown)


edwintorok:
That ticket number redirects to CP-27911, please use CP-27911.

ocaml/tests/test_xapi_vbd_helpers.ml
```diff
@@ -12,48 +12,44 @@
...
   let vbd_ref, vbd_record = setup_test ~__context ~vdi_fun in
-  assert_equal
-    ~cmp
-    ~printer:string_of_api_exn_opt
-    exc (Hashtbl.find (Xapi_vbd_helpers.valid_operations ~__context ~expensive_sharing_checks vbd_record vbd_ref) op)
+  let get_error_code_of op =
+    let valid_ops = (Xapi_vbd_helpers.valid_operations ~__context ~expensive_sharing_checks vbd_record vbd_ref) in
+    match Hashtbl.find valid_ops op with
+    | Some (code, _::_ ) -> Some code
+    | Some _ | None -> None
```
gaborigloi: I think this should be simplified to 
```ocaml
| Some (code, _ ) -> Some code | None -> None
```

minishrink: I don't think it would be specific to 2-element lists, rather it would skip out empty lists, but either way you're right that it needs changing.

gaborigloi: Yes sorry that's correct.

minishrink: That said, do you think I should remove the case that takes the first element of  a `pair option`? As it is, this is designed to accept `Some api_error_code,params`, not `Some individual_error`.

gaborigloi: Yeah, I think it's best to remove it.

minishrink: I forgot how typechecking works :P of course this only expects `(string * 'a) option`

PR #3585, opened 2018-05-01
[REQ-477] CP-26912: Enable live bond creation and destruction on clusters

minishrink:
This ticket is necessary for other post-MVP work concerning live bond creation, with the following stages:

- Addition of a `cluster_host.PIF` field
This allows us to represent the status of cluster-wide bond creation/moving the cluster onto a different network by seeing whether all cluster_host.PIFs are on the same network yet
**NOTE:** `Cluster.create` now takes a `PIF` parameter in order to create a cluster_host on the pool master

- Removal of Cluster.network field
The Cluster.network field is generally used to indirectly associate a PIF with a cluster_host and acquire IPs, therefore it became obsolete once an explicit PIF field was added
**NOTE:** `Cluster.pool_create` still takes a network parameter in order to assign PIFs to all hosts in the pool

- `Bond.create` updates the host's `cluster_host.PIF`
This is still a WIP as more changes are likely to follow

- Minor refactoring as follows:
  - addition of a `get_cluster_network` helper in Xapi_clustering.ml for use in `Xapi_cluster_host.create_as_necessary`
  - refactoring of quicktest_cluster.ml and cleaning up some of the clustering unit tests i.e. updating them with the exceptions the code raises
  - removal of obsolete code

This PR is posted as is for discussion, as there may still be changes to make. Additionally, I'm unsure what the best way to test this is: it's certainly worth writing unit tests, but that won't help test networkign changes (though I'll have a look at using a mock RPC for spoofing networkd responses)


ocaml/idl/datamodel_cluster_host.ml
```diff
@@ -20,6 +20,7 @@ let create = call
     ~params:
       [ Ref _cluster, "cluster", "Cluster to join"
       ; Ref _host,    "host",    "new cluster member"
+      ; Ref _pif,     "pif",     "interface to use"
```
lindig: Capitalization is inconsistent here: "Cluster" vs. "interface"

ocaml/xapi/xapi_bond.ml
```diff
@@ -375,6 +375,14 @@ let create ~__context ~network ~members ~mAC ~mode ~properties =
       Db.Bond.create ~__context ~ref:bond ~uuid:(Uuid.to_string (Uuid.make_uuid ())) ~master:master ~other_config:[]
         ~primary_slave ~mode ~properties ~links_up:0L;
 
+      (* If host is in a cluster, update the cluster_host PIF *)
```
robhoes: Only if the current `Cluster_host.PIF` is being bonded, right?

ocaml/quicktest/quicktest_cluster.ml
```diff
@@ -1,92 +1,88 @@
...
-        then begin
+    let enabled_cluster_hosts =
+      List.filter
+        (fun self -> C.Cluster_host.get_enabled ~session_id ~rpc ~self)
+        (C.Cluster_host.get_all ~session_id ~rpc)
+    in
+    if is_empty enabled_cluster_hosts
+    then Q.debug test_all_pifs "No PIFS with clustering enabled, skipping tests."
+    else begin
+      let pifs_with_clustering =
```
lindig: List.map and List.iter could be combined into a pipe:
```
xxx
|> List.map ..
|> List.iter ..
```

ocaml/xapi/xapi_bond.ml
robhoes: You also need to do the opposite in `destroy` below.

ocaml/tests/test_common.ml
```diff
@@ -211,7 +211,7 @@ let make_pool ~__context ~master ?(name_label="") ?(name_description="")
     ?(redo_log_vdi=Ref.null) ?(vswitch_controller="") ?(igmp_snooping_enabled=false) ?(restrictions=[])
     ?(current_operations=[]) ?(allowed_operations=[])
     ?(other_config=[Xapi_globs.memory_ratio_hvm; Xapi_globs.memory_ratio_pv])
-    ?(ha_cluster_stack="xhad") ?(guest_agent_config=[]) ?(cpu_info=[]) ?(policy_no_vendor_device=false) ?(live_patching_disabled=false)() =
+    ?(ha_cluster_stack=(!Xapi_globs.cluster_stack_default)) ?(guest_agent_config=[]) ?(cpu_info=[]) ?(policy_no_vendor_device=false) ?(live_patching_disabled=false) () =
```
lindig: Please avoid overlong lines and break this line up.

minishrink:
Updated in line with feedback.

ocaml/idl/datamodel_cluster.ml
```diff
@@ -28,8 +28,8 @@ let create = call
     ~doc:"Creates a Cluster object and one Cluster_host object as its first member"
     ~result:(Ref _cluster, "the new Cluster")
     ~versioned_params:
-            ([{param_type=Ref _network; param_name="network"; param_doc="the single network on which corosync carries out its inter-host communications"; param_release=kolkata_release; param_default=None};
-             {param_type=String; param_name="cluster_stack"; param_doc="simply the string 'corosync'. No other cluster stacks are currently supported"; param_release=kolkata_release; param_default=None};
+            ([{param_type=(Ref _pif); param_name="PIF"; param_doc="The PIF to connect the cluster's first cluster_host to"; param_release=kolkata_release; param_default=None};
```
edwintorok: I think this needs `lima_release`.

ocaml/idl/datamodel_cluster.ml
```diff
@@ -51,8 +51,8 @@ let pool_create = call
     ~doc:"Attempt to create a Cluster from the entire pool"
     ~result:(Ref _cluster, "the new Cluster")
     ~versioned_params:
-            ([{param_type=Ref _network; param_name="network"; param_doc="the single network on which corosync carries out its inter-host communications"; param_release=kolkata_release; param_default=None};
-             {param_type=String; param_name="cluster_stack"; param_doc="simply the string 'corosync'. No other cluster stacks are currently supported"; param_release=kolkata_release; param_default=None};
+          ([{param_type=String; param_name="cluster_stack"; param_doc="simply the string 'corosync'. No other cluster stacks are currently supported"; param_release=kolkata_release; param_default=None};
+            {param_type=Ref _network; param_name="network"; param_doc="the single network on which corosync carries out its inter-host communications"; param_release=kolkata_release; param_default=None};
```
edwintorok: minor: I see no functional changes in this hunk, maybe the original order should be kept and then it doesn't show up in the diff.

ocaml/idl/datamodel_cluster_host.ml
```diff
@@ -92,9 +93,13 @@ let t =
           ~ty:Bool "enabled" ~default_value:(Some (VBool false))
           "Whether the cluster host believes that clustering should be enabled on this host"
 
+      ; field  ~qualifier:StaticRO ~lifecycle
+          ~ty:(Ref _pif) "PIF" ~default_value:(Some (VRef null_ref))
+          "Reference to the Network object"
```
edwintorok: PIF object, not network

ocaml/xapi/cli_operations.ml
```diff
@@ -4939,7 +4926,7 @@ module Cluster = struct
     let network = Client.Network.get_by_uuid rpc session_id network_uuid in
     let token_timeout = get_float_param params "token-timeout" ~default:Constants.default_token_timeout_s in
     let token_timeout_coefficient = get_float_param params "token-timeout-coefficient" ~default:Constants.default_token_timeout_coefficient_s in
-    let cluster = Client.Cluster.pool_create ~rpc ~session_id ~network ~cluster_stack ~token_timeout ~token_timeout_coefficient in
+    let cluster = Client.Cluster.pool_create ~rpc ~session_id ~cluster_stack ~network ~token_timeout ~token_timeout_coefficient in
```
edwintorok: same as above: no functional change here, please keep original order.

ocaml/xapi/message_forwarding.ml
```diff
@@ -4309,9 +4309,9 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
         (fun () ->
            Local.Cluster.destroy ~__context ~self)
 
-    let pool_create ~__context ~network ~cluster_stack ~token_timeout ~token_timeout_coefficient =
+    let pool_create ~__context ~cluster_stack ~network ~token_timeout ~token_timeout_coefficient =
```
edwintorok: no change

ocaml/xapi/message_forwarding.ml
```diff
@@ -4309,9 +4309,9 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
       info "Cluster.pool_create";
-      Local.Cluster.pool_create ~__context ~network ~cluster_stack ~token_timeout ~token_timeout_coefficient
+      Local.Cluster.pool_create ~__context ~cluster_stack ~network ~token_timeout ~token_timeout_coefficient
```
edwintorok: no change

ocaml/xapi/xapi_cluster.ml
```diff
@@ -103,22 +102,22 @@ let destroy ~__context ~self =
     handle_error error
 
 (* helper function; concurrency checks are done in implementation of Cluster.create and Cluster_host.create *)
-let pool_create ~__context ~network ~cluster_stack ~token_timeout ~token_timeout_coefficient =
+let pool_create ~__context ~cluster_stack ~network ~token_timeout ~token_timeout_coefficient =
```
edwintorok: no change

ocaml/xapi/xapi_pif.ml
```diff
@@ -451,20 +451,17 @@ let introduce_internal
   pif
 
 (* Assertion passes if network has clusters attached but host has disabled clustering *)
-let assert_no_clustering_enabled ~__context ~network ~host =
+let assert_no_clustering_enabled ~__context ~host =
```
edwintorok: I think this should take a pif instead of a network then. You have to be able to reconfigure PIFs that are  not in use by clustering. The name of the function should probably be changed to `assert_no_clustering_enabled_on_pif`

ocaml/xapi/xapi_pif.ml
```diff
@@ -451,20 +451,17 @@ let introduce_internal
...
   if not (Xapi_clustering.is_clustering_disabled_on_host ~__context host)
-  then
-    (Db.Cluster.get_refs_where ~__context
-      ~expr:Db_filter_types.(Eq(Field "network", Literal (Ref.string_of network))))
-    |> function
-    | []   -> ()
-    | _::_ -> raise Api_errors.(Server_error (clustering_enabled_on_network, [Ref.string_of network]))
+  then begin match Xapi_clustering.find_cluster_host ~__context ~host with
+    | Some cluster_host ->
+      raise Api_errors.(Server_error (clustering_enabled, [Ref.string_of cluster_host]))
```
edwintorok: Looks like this always fails if clustering is enabled, which matches the function's name but in this case the function's name is wrong.

ocaml/xapi/xapi_cluster.ml
```diff
@@ -40,10 +40,10 @@ let create ~__context ~cluster_stack ~pool_auto_join ~token_timeout ~token_timeo
       let cluster_host_uuid = Uuidm.to_string (Uuidm.create `V4) in
       (* For now we assume we have only one pool
          TODO: get master ref explicitly passed in as parameter*)
-      let pool = Db.Pool.get_all ~__context |> List.hd in
-      let host = Db.Pool.get_master ~__context ~self:pool in
+      let host = Helpers.get_master ~__context in
+      let network = Db.Network.get_all ~__context |> List.hd in
```
edwintorok: This looks wrong and I think is reverted by a subsequent commit. It would be better to squash this commit with th e other one that removes the network parameter from everywhere, then the final diff of each commit would make sense.

ocaml/idl/datamodel_cluster.ml
minishrink: I think these changes are due to originally removing it as part of removing `Cluster.network` uses, then adding it back after deciding to pass a `network` param to `Cluster.pool_create`, so I'll try and change it back

minishrink:
Updated.

minishrink:
Updated, will squash the final approved version, although I think it's good to separate out some of these changes. @edvintorok we don't seem to have a `lima_release` available yet in the datamodel?

ocaml/idl/datamodel_cluster.ml
minishrink: Should that be available in the datamodel? It currently isn't.

lindig:
Before this is merged I'd like to point out that these commits contain some quite long lines: many over 150 characters and some over 200 characters. This creates problems for reviews and code merges. Please have a discussion about this.

ocaml/idl/datamodel_cluster.ml
edwintorok: You will need something like this: https://github.com/edwintorok/xen-api/commit/7ee65030c021da19eea1741174047df6bc424b93
You can cherry-pick that commit if you like

edwintorok:
It would be nice if the datamodel_cluster/cluster_host was reformatted to split the long lines.

edwintorok:
Please also bump `schema_minor_vsn` in datamodel_common.ml, since you added a field, the comment there says:
```
(* IMPORTANT: Please bump schema vsn if you change/add/remove a _field_.
              You do not have to bump vsn if you change/add/remove a message *)

```

edwintorok:
Also consider whether we can provide a compatibility `get_network` operation on the Cluster object.

ocaml/idl/datamodel_cluster.ml
```diff
@@ -28,10 +38,25 @@ let create = call
...
+
+        {param_type=String;
+         param_name="cluster_stack";
+         param_doc="simply the string 'corosync'. No other cluster stacks are currently supported";
+         param_release=kolkata_release;
+         param_default=None};
+
+        {param_type=Bool;
+         param_name="pool_auto_join";
+         param_doc="true if xapi is automatically joining new pool members to the cluster";
```
robhoes: I know this was already there, but still: don't use terms like "xapi" here. This goes into the API docs that should be independent of the implementation, daemons etc.

Better is something like "Indicates whether new pool members are automatically added to the cluster".

ocaml/xapi/xapi_bond.ml
```diff
@@ -333,19 +333,19 @@ let create ~__context ~network ~members ~mAC ~mode ~properties =
       (* 5. Members must not be the management interface if HA is enabled *)
       (* 6. Members must be PIFs that are managed by xapi *)
       (* 7. Members must have the same PIF properties *)
-      (* 8. Only the primary PIF should have a non-None IP configuration *)
+      (* 8. Only the primary PIF should have a valid IP configuration *)
```
robhoes: The word "should" isn't right: it's fine for a bond to have no IP at all. "Valid IP configuration" is vague as well, because "none" is also valid.

Better is: "At most one member may have an IP address".

ocaml/xapi/xapi_bond.ml
```diff
@@ -431,6 +431,21 @@ let create ~__context ~network ~members ~mAC ~mode ~properties =
             Db.PIF.set_disallow_unplug ~__context ~self:pif ~value:false)
           members
       end;
+
+      (* If the host has a cluster_host AND the cluster_host's PIF is a member, update the PIF *)
+      begin match Xapi_clustering.find_cluster_host ~__context ~host with
```
robhoes: To declutter this already very long function, and reduce duplication with the destroy code below, please put (some of) this block into a new function, e.g. `move_cluster_pif`.

ocaml/xapi/xapi_bond.ml
```diff
@@ -431,6 +431,21 @@ let create ~__context ~network ~members ~mAC ~mode ~properties =
...
+        then begin
+          let master_pif_rec = Db.PIF.get_record ~__context ~self:master in
+          Xapi_clustering.assert_pif_prerequisites (master,master_pif_rec);
+          debug "Setting cluster_host %s PIF to new bond PIF %s" (Ref.string_of cluster_host) (Ref.string_of master);
+          Db.Cluster_host.set_PIF ~__context ~self:cluster_host ~value:master
+        end
+      | None -> ()
+      end;
+
       TaskHelper.set_progress ~__context 1.0;
```
robhoes: You should refine the progress updates since there is a new step introduced now.

ocaml/xapi/xapi_bond.ml
robhoes: For the other things that potentially need moving, information is gathered at the top of the function. It would make sense to do the same for this new case.

ocaml/xapi/xapi_bond.ml
```diff
@@ -431,6 +431,21 @@ let create ~__context ~network ~members ~mAC ~mode ~properties =
...
       end;
+
+      (* If the host has a cluster_host AND the cluster_host's PIF is a member, update the PIF *)
+      begin match Xapi_clustering.find_cluster_host ~__context ~host with
+      | Some cluster_host ->
+        let cluster_pif_ref = Db.Cluster_host.get_PIF ~__context ~self:cluster_host in
+        if List.mem cluster_pif_ref members
+        then begin
+          let master_pif_rec = Db.PIF.get_record ~__context ~self:master in
+          Xapi_clustering.assert_pif_prerequisites (master,master_pif_rec);
```
robhoes: It's a bit late to do asserts now, because the bond has already been created. Aborting here would leave things in a bad state. This is why there is a "Validate constraints" section in the beginning on the function.

However, do we really need this check? I thought that we concluded that the bond master meets the prerequisites by construction?

ocaml/xapi/xapi_bond.ml
```diff
@@ -502,6 +517,19 @@ let destroy ~__context ~self =
         Db.PIF.set_disallow_unplug ~__context ~self:primary_slave ~value:true
       end;
 
+      (** Move Cluster_host to primary slave IF currently on bond *)
+      begin match Xapi_clustering.find_cluster_host ~__context ~host with
```
robhoes: Similar questions as above.

mseri:
Note that lima_release should point to a fake version to be updaated only at release

ocaml/xapi/xapi_bond.ml
minishrink: Yeah, that's true, will remove them.

ocaml/xapi/xapi_bond.ml
minishrink: Is this just for a progress bar? I.e. are the intervals meaningful or can I just put something in between 0.8 and 1.0, since moving the cluster PIF isn't too big a step?

minishrink:
@robhoes @edwintorok Updated, let me know if there's anything else or if we can merge.

ocaml/idl/datamodel_common.ml
```diff
@@ -7,7 +7,7 @@ open Datamodel_roles
               You do not have to bump vsn if you change/add/remove a message *)
 
 let schema_major_vsn = 5
-let schema_minor_vsn = 142
+let schema_minor_vsn = 152
```
edwintorok: Lets use 200 here please as discussed on the channel ("we could go to 200 as well and make this a policy: we start with the next hundred after a release"), and add a comment:
`(* When introducing a new release, bump the schema minor version to the next hundred, to leave a gap for potential hotfixes needing to increment the schema version.*)`

ocaml/xapi/xapi_bond.ml
robhoes: Yes, you can make up a number to more or less indicate the amount of process.

PR #3586, opened 2018-05-01
CP-27911: Port test_sdn_controller to Alcotest

minishrink:

### DIFF
--- a/ocaml/tests/test_sdn_controller.ml
+++ b/ocaml/tests/test_sdn_controller.ml
@@ -12,120 +12,132 @@
  * GNU Lesser General Public License for more details.
  *)
 
-open OUnit
-open Test_common
+module T = Test_common
 open Stdext.Unixext
 
 let test_sdn_controller_introduce_ok () =
   let protocol = `ssl in
   let address = "192.168.1.10" in
   let port = 6632L in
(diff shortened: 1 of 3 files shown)


ocaml/tests/test_sdn_controller.ml
```diff
@@ -12,120 +12,132 @@
...
+  Alcotest.check (Alcotest_comparators.from_rpc_of_t API.rpc_of_sdn_controller_protocol)
+    "test_sdn_controller_ssl_default_port: protocol"
+    protocol
+    (Db.SDN_controller.get_protocol ~__context ~self:sdn_controller);
+  Alcotest.(check string)
+    "test_sdn_controller_ssl_default_port: address"
+    address
+    (Db.SDN_controller.get_address ~__context ~self:sdn_controller);
+  Alcotest.(check int64)
+    "test_sdn_controller_ssl_default_port: port"
```
gaborigloi: There is no need to prefix these messages with the test name as Alcotest will say which test failed :)

minishrink: It does when outputting all of the test results, but I thought it might be handy to name the individual test within the test log itself, which doesn't include the title.

PR #3588, opened 2018-05-07
CA-289620: Remove redundant log of events_watch

krizex:
1. Remove redundent log
2. Add log for VUSB

### DIFF
--- a/ocaml/xapi/xapi_xenops.ml
+++ b/ocaml/xapi/xapi_xenops.ml
@@ -2087,43 +2087,22 @@ let rec events_watch ~__context cancel queue_name from =
            add_event ev;
            match ev with
            | Vm id ->
-             if Events_from_xenopsd.are_suppressed id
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_xenops.ml
```diff
@@ -2087,43 +2087,22 @@ let rec events_watch ~__context cancel queue_name from =
            add_event ev;
            match ev with
            | Vm id ->
-             if Events_from_xenopsd.are_suppressed id
-             then debug "ignoring xenops event on VM %s" id
-             else begin
-               debug "xenops event on VM %s" id;
-               update_vm ~__context id;
-             end
+             debug "xenops event on VM %s" id;
```
robhoes: I see that `update_*` already do similar logging, and event suppression is tested there. Is it still useful to log this line here, or shall we just drop all `debug` in this function?

PR #3591, opened 2018-05-10
CP-27911: Remove deprecated test_pr1510

minishrink:
This test was added years ago for a PR that has since been merged, but with little surviving code at this point. PR-1510 did the following:
- enabled pool-join without a running v6d (this change persists in
  xapi_pool.ml)
-  modified license_init.ml, which has since been reverted

This test never ran and tests code that is no longer in effect, so it is unnecessary.


robhoes:
The description does match... PR-1510 was for LACP bonding. These tests should be relevant.

minishrink:
Yeah, I wasn't really sure how they related. Maybe it's from a different repo originally? Anyway, why don't we currently run these tests?

PR #3592, opened 2018-05-10
CP-27899: Convert quicktests to Alcotest

gaborigloi:
And temporarily remove inactive with_vm tests - these quicktests were
looking for a template that does not exist currently, therefore they
were always skipped.

This commit makes the storage tests idempotent - they must not leave
temporary VDIs around. Also, now the storage tests are only run on the
SRs that support them, so the quicktests should run without errors on
any setup.


ocaml/quicktest/jbuild
```diff
@@ -31,8 +31,8 @@ let () = Printf.ksprintf Jbuild_plugin.V1.send {|
   (public_name quicktestbin)
   (package xapi)
   (flags (:standard -bin-annot %s -warn-error +a-3-4-6-9-27-28-29-52))
-  (libraries (
-   xapi_internal
-  ))
+  (libraries
+    (alcotest
```
mseri: You also use xapi-stdext-pervasives below and unix. I would also double check rpclib and threads

gaborigloi:
I think we shouldn't review the actual code, as it's huge, and I've tested it with local, NFS, and GFS2 SRs, and it passed, and I've also done a build and run the Quicktest TC on it, and that also passed. So I think we should merge this soon so we start benefiting from the change and fix any issues that come up. I should probably also give a seminar about how the new storage quictests can be written :)

mseri:
I agree, but having a brief read is already useful before merging

ocaml/quicktest/jbuild
```diff
@@ -33,6 +33,12 @@ let () = Printf.ksprintf Jbuild_plugin.V1.send {|
   (flags (:standard -bin-annot %s -warn-error +a-3-4-6-9-27-28-29-52))
   (libraries
     (alcotest
-     xapi_internal))
+     stdext
```
mseri: Remove stdext

ocaml/quicktest/jbuild
```diff
@@ -33,6 +33,12 @@ let () = Printf.ksprintf Jbuild_plugin.V1.send {|
...
+     xapi_internal
+     xapi-stdext-date
+     xapi-stdext-monadic
+     xapi-stdext-pervasives
+     xapi-stdext-std
+     xapi-stdext-threads))
```
mseri: Do you actually use all of the above?

mseri: `unix` is missing

ocaml/quicktest/jbuild
```diff
@@ -31,8 +31,14 @@ let () = Printf.ksprintf Jbuild_plugin.V1.send {|
   (public_name quicktestbin)
   (package xapi)
   (flags (:standard -bin-annot %s -warn-error +a-3-4-6-9-27-28-29-52))
-  (libraries (
-   xapi_internal
-  ))
+  (libraries
+    (alcotest
+     stdext
```
mseri: Please remove `stdext`

gaborigloi: But the `Stdext` module is actually used by the code, so will that work?

ocaml/quicktest/jbuild
gaborigloi: Yes, it's all used by quicktest code.

ocaml/quicktest/jbuild
mseri: Please get rid of it from the code. It should be a fairly mechanical replacemenf of `open Stdext open ...` with `open Xapi_stdext_...`

ocaml/quicktest/quicktest_event.ml
```diff
@@ -0,0 +1,274 @@
...
+  Thread.join i_should_succeed;
+  Alcotest.(check bool) "Event.from got cancelled by mistake" true !ok
+
+let object_level_event_test session_id =
+  let m = Mutex.create () in
+  let finished = ref false in
+  let reported_failure = ref false in
+  (* Let's play with templates *)
+  let vms = Client.Client.VM.get_all !QC.rpc session_id in
+  if List.length vms < 2 then failwith "Test needs 2 VMs";
```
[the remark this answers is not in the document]
gaborigloi: Good spot, I'll fix this later with the rest of the VM tests.

ocaml/quicktest/quicktest_import_export.ml
```diff
@@ -0,0 +1,108 @@
...
+             then Alcotest.fail
+                 (Printf.sprintf "Device %s varies in VDIness (original = %s; new = %s)"
+                    all.API.vBD_userdevice
+                    (Client.Client.VDI.get_uuid !rpc session_id orig_vbd.API.vBD_VDI)
+                    (Client.Client.VDI.get_uuid !rpc session_id all.API.vBD_VDI)
+                 )
+           | "3" ->
+             (* VDI should be different *)
+             if all.API.vBD_VDI = orig_vbd.API.vBD_VDI
+             then Alcotest.fail (Printf.sprintf "Device %s should not vary in VDIness" all.API.vBD_userdevice)
```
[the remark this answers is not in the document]
gaborigloi: Not sure, it wasn't written by me

ocaml/quicktest/storage_test.ml
```diff
@@ -0,0 +1,261 @@
...
+  let with_any session_id sr_info f =
+    let rpc = !Quicktest_common.rpc in
+    if List.mem `vdi_create sr_info.allowed_operations then begin
+      with_new session_id sr_info.sr f
+    end else begin
+      let self = Client.Client.SR.get_VDIs ~rpc ~session_id ~self:sr_info.sr |> List.hd in
+      f self
+    end
+end
+
```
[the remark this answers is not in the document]
gaborigloi: I've added these for clarity, I think they make the code more readable.

ocaml/quicktest/quicktest_cluster.ml
```diff
@@ -40,53 +34,45 @@ let reconfigure_ipv6 ~session_id ~self ~dNS =
...
-    failed test "PIF.reconfigure_ip should raise clustering_enabled_on_network."
+    Alcotest.fail "PIF.reconfigure_ip should raise clustering_enabled_on_network."
   with
   | Api_errors.(Server_error(code,_)) when code=Api_errors.clustering_enabled_on_network
-      -> debug test (Printf.sprintf "%s raised as expected." Api_errors.clustering_enabled_on_network);
-         success test
+      -> print_endline (Printf.sprintf "%s raised as expected." Api_errors.clustering_enabled_on_network)
   | Api_errors.(Server_error(_,_)) -> () (* Don't fail on other API errors, only test clustering *)
-  | Abort_test s -> failed test s
-  | e -> failed test (ExnHelper.string_of_exn e)
```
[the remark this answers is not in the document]
gaborigloi: The test will fail, as before, so the old behaviour is kept.

gaborigloi:
Thanks for the review, I've fixed the above issues.

ocaml/quicktest/quicktest_cluster.ml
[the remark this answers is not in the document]
minishrink: Should we at least wrap those exns with `Alcotest.fail`?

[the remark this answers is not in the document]
gaborigloi: That's not necessary, Alcotest does all the work for us, it automatically marks the test as failed if an exn happens, and prints the backtrace too :)
This is why alcotest is the best, it does everything out of the box

PR #3593, opened 2018-05-14
CA-275120: XenAPI methods only callable by SM are publicly visible

YarsinCitrix:
There are some calls, such as VDI.db_introduce and SR.set_physical_utilisation, that are publicly visible, but can only be called by SM, because we call Sm.assert_session_has_internal_sr_access in message forwarding for these functions. We should make these hidden by setting hide_from_docs to true in the datamodel. Going through all of the 'should be hidden' APIs is outside the scope of this ticket. This ticket only concerns hiding the public SR and VDI function that are only callable by SM.

Please refer to the ticket for more discussion.

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -2016,7 +2016,7 @@ module SR = struct
       ~flags:[`Session]
       ~doc:"Sets the SR's virtual_allocation field"
       ~hide_from_docs:true
-      ~allowed_roles:_R_POOL_OP
+      ~allowed_roles:_R_LOCAL_ROOT_ONLY
       ()
 
   let set_physical_size = call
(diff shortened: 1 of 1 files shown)


gaborigloi:
Should we also add
```ocaml
```
to each of these? This should be fine IMO as long as they are only called by the local SM, we have this permission for `set_cbt_enabed`. (Although this is not important)

ocaml/idl/datamodel.ml
```diff
@@ -2372,7 +2374,7 @@ module VDI = struct
       ~allowed_roles:_R_VM_ADMIN
       ()
 
-  let db_introduce = { pool_introduce with msg_name = "db_introduce"; msg_hide_from_docs = false }
+  let db_introduce = { pool_introduce with msg_name = "db_introduce"; msg_hide_from_docs = true }
 
   let db_forget = call
```
gaborigloi: Should we add `hide_from_docs` to this too, it was listed on the ticket?

YarsinCitrix: Well spotted!

gaborigloi:
@jonludlam said it's fine to change these to `LOCAL_ROOT_ONLY`, we should just check that they still work afterwards.

PR #3594, opened 2018-05-16
CP-27172: Implement Host.destroy integration with clustering daemon

edwintorok:
Introduced a new API `Cluster_host.forget`, because declare dead already has a different semantics in Xen API (see comments in commit message for full details).

There is this outstanding TODO on which I'd like some thoughts: "although things could go wrong if we try to join a host with the same IP as one that we are trying to declare as dead... (which can happen due to DHCP)."

If we have 2 hosts dead, declare one of them as dead (it gets added to pending forget), and then try to join a new node which reuses the same IP before marking the 2nd host as dead then we can't really tell the difference between the newly joined host and the one declared as dead.

### DIFF
--- a/ocaml/idl/datamodel_cluster.ml
+++ b/ocaml/idl/datamodel_cluster.ml
@@ -152,6 +152,11 @@ let t =
           ~ty:(Set (Ref _cluster_host)) "cluster_hosts"
           "A list of the cluster_host objects associated with the Cluster"
 
+      ; field ~qualifier:DynamicRO ~lifecycle:[ Prototyped, rel_lima, "" ]
+          ~ty:(Set String) "pending_forget" ~default_value:(Some (VSet []))
+          "Internal field used by Host.destroy to store the IP of cluster members \
+           marked as permanently dead but not yet removed"
+
       ; field   ~qualifier:StaticRO ~lifecycle
           ~ty:String "cluster_token" ~default_value:(Some (VString ""))
           "The secret key used by xapi-clusterd when it talks to itself on other hosts"
--- a/ocaml/idl/datamodel_cluster_host.ml
+++ b/ocaml/idl/datamodel_cluster_host.ml
@@ -66,6 +66,17 @@ let disable = call
     ~allowed_roles:_R_POOL_ADMIN
     ()
 
+let forget = call
+  ~name:"forget"
+  ~doc:"Permanently remove a dead host from the cluster. This host must never rejoin the cluster."
+  ~params:
+      [ Ref _cluster_host, "self", "the cluster_host to declare permanently dead and forget"
+      ]
+  ~lifecycle:[Prototyped, rel_lima, ""]
+  ~allowed_roles:_R_LOCAL_ROOT_ONLY
+  ~hide_from_docs:true
+  ()
+
 let t =
   create_obj
     ~name: _cluster_host
@@ -111,6 +122,7 @@ let t =
       ; destroy
       ; enable
       ; force_destroy
+      ; forget
       ; disable
       ]
     ()
--- a/ocaml/idl/datamodel_common.ml
+++ b/ocaml/idl/datamodel_common.ml
@@ -8,7 +8,7 @@ open Datamodel_roles
               When introducing a new release, bump the schema minor version to the next hundred
               to leave a gap for potential hotfixes needing to increment the schema version.*)
 let schema_major_vsn = 5
-let schema_minor_vsn = 200
+let schema_minor_vsn = 201
 
 (* Historical schema versions just in case this is useful later *)
 let rio_schema_major_vsn = 5
--- a/ocaml/tests/test_cluster_host.ml
+++ b/ocaml/tests/test_cluster_host.ml
@@ -19,7 +19,7 @@ let create_cluster ~__context pool_auto_join =
   let cluster_uuid = Uuidm.to_string (Uuidm.create `V4) in
   Db.Cluster.create ~__context ~ref:cluster_ref ~uuid:cluster_uuid ~cluster_token:"token"
     ~cluster_stack:Constants.default_smapiv3_cluster_stack ~token_timeout:5000L ~token_timeout_coefficient:1000L ~allowed_operations:[]
-    ~current_operations:[] ~pool_auto_join ~cluster_config:[] ~other_config:[];
+    ~current_operations:[] ~pool_auto_join ~cluster_config:[] ~other_config:[] ~pending_forget:[];
   cluster_ref
 
 let check_cluster_option =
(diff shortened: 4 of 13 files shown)


ocaml/idl/datamodel_common.ml
```diff
@@ -8,7 +8,7 @@ open Datamodel_roles
               When introducing a new release, bump the schema minor version to the next hundred
               to leave a gap for potential hotfixes needing to increment the schema version.*)
 let schema_major_vsn = 5
-let schema_minor_vsn = 200
```
minishrink: I assume this will be updated later on?

ocaml/tests/test_cluster_host.ml
```diff
@@ -122,11 +122,91 @@ let test_destroy_forbidden_when_sr_attached () =
...
+
+
+let test_forget () =
+  let __context = Test_common.make_test_database () in
+  let host2 = Test_common.make_host ~__context () in
+  let cluster, original_cluster_hosts = make ~__context [host2] in
+
+  Xapi_host.destroy ~__context ~self:host2;
+  let pending = Db.Cluster.get_pending_forget ~__context ~self:cluster in
+  Alcotest.(check (list string) "no pending forgets"
```
minishrink: You don't need to include the whole test in parentheses, just `Alcotest.(check (list string))`, although it's better to use `Alcotest.(check (slist string [String.compare or compare]))`.

ocaml/idl/datamodel_common.ml
```diff
@@ -8,7 +8,7 @@ open Datamodel_roles
...
+let schema_minor_vsn = 201
```
minishrink: I assume this will be updated later on?

edwintorok: I think we should update it everytime we make a change to the DB fields, otherwise its hard to test out the changes starting from a previous build and swapping out xapi binary.

edwintorok:
Rebased on latest feature branch

ocaml/idl/datamodel_cluster_host.ml
```diff
@@ -66,6 +66,17 @@ let disable = call
...
     ()
 
+let forget = call
+  ~name:"forget"
+  ~doc:"Permanently remove a dead host from the cluster. This host must never rejoin the cluster."
+  ~params:
+      [ Ref _cluster_host, "self", "the cluster_host to declare permanently dead and forget"
+      ]
+  ~lifecycle:[Prototyped, rel_lima, ""]
+  ~allowed_roles:_R_LOCAL_ROOT_ONLY
```
gaborigloi: Is this intentional, to make it only work from the CLI?

ocaml/idl/datamodel_cluster.ml
```diff
@@ -152,6 +152,10 @@ let t =
           ~ty:(Set (Ref _cluster_host)) "cluster_hosts"
           "A list of the cluster_host objects associated with the Cluster"
 
+      ; field ~qualifier:DynamicRO ~lifecycle:[ Prototyped, rel_lima, "" ]
+          ~ty:(Set String) "pending_forget" ~default_value:(Some (VSet []))
+          "A list of cluster members that were declared permanently dead and not yet removed"
+
```
gaborigloi: Should we set `~internal_only:true` to make this field invisible to the user, since the API call will be hidden too?

ocaml/xapi/db_gc_util.ml
```diff
@@ -110,6 +110,13 @@ let gc_VIFs ~__context =
 let gc_PBDs ~__context =
   gc_connector ~__context Db.PBD.get_all Db.PBD.get_record (fun x->valid_ref __context x.pBD_host) (fun x->valid_ref __context x.pBD_SR) Db.PBD.destroy
 
+let gc_Cluster_hosts ~__context =
+  gc_connector ~__context Db.Cluster_host.get_all Db.Cluster_host.get_record
+    (fun x -> valid_ref __context x.cluster_host_host)
+    (* TODO: a PIF after https://github.com/xapi-project/xen-api/pull/3585 *)
```
gaborigloi: That PR has been merged now (but it's fine by me to do this later)

ocaml/xapi/message_forwarding.ml
```diff
@@ -4379,5 +4379,24 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
+       * We might've run force destroy and this host would no longer have a cluster host
+       * *)
+      let other_hosts =
+        Db.Cluster.get_cluster_hosts ~__context ~self:cluster
+        |> List.filter ((<>) self) in
+      match other_hosts with
+      | [] -> info "No other cluster hosts, nothing to do" (* go ahead and finish Host.destroy *)
+      | other_cluster_host :: _ ->
+        let host = Db.Cluster_host.get_host ~__context ~self:other_cluster_host in
+        Xapi_cluster_helpers.with_cluster_operation ~__context ~self:cluster ~doc:"Cluster.remove" ~op:`remove
```
gaborigloi: Should we update the `doc` and the `op` here?

ocaml/xapi/message_forwarding.ml
```diff
@@ -4379,5 +4379,24 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
              (fun () ->
                 do_op_on ~__context ~local_fn ~host
                   (fun session_id rpc -> Client.Cluster_host.disable rpc session_id self)))
+
+    let forget ~__context ~self =
+      info "Cluster_host.forget cluster_host:%s" (Ref.string_of self);
+      let cluster = Db.Cluster_host.get_cluster ~__context ~self in
+      let local_fn = Local.Cluster_host.disable ~self in
+      (* We need to ask another host that has a cluster host to mark it as dead.
+       * We might've run force destroy and this host would no longer have a cluster host
```
gaborigloi: Hmm, this host must have a cluster_host though, because that's the `self` parameter of this function

ocaml/xapi/message_forwarding.ml
```diff
@@ -4379,5 +4379,24 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
+       * *)
+      let other_hosts =
+        Db.Cluster.get_cluster_hosts ~__context ~self:cluster
+        |> List.filter ((<>) self) in
+      match other_hosts with
```
gaborigloi: What happens if this gets forwarded to one of the dead hosts, or one in `pending_forget`, is that possible?
Should we keep trying until we find a live one?

ocaml/idl/datamodel_cluster.ml
```diff
@@ -152,6 +152,10 @@ let t =
           ~ty:(Set (Ref _cluster_host)) "cluster_hosts"
           "A list of the cluster_host objects associated with the Cluster"
 
+      ; field ~qualifier:DynamicRO ~lifecycle:[ Prototyped, rel_lima, "" ]
+          ~ty:(Set String) "pending_forget" ~default_value:(Some (VSet []))
+          "A list of cluster members that were declared permanently dead and not yet removed"
```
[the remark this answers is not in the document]
edwintorok: yes, list of cluster member IPs.

ocaml/idl/datamodel_cluster_host.ml
edwintorok: You are supposed to use Host.destroy, Cluster_host.forget is just an internal API used by XAPI. We need message-forwarding to send this request to a host that has clustering.

edwintorok: We could mark it as hidden in records.ml

ocaml/xapi/db_gc_util.ml
edwintorok: I rebased the PR but forgot to remove this line, will fix.

ocaml/xapi/message_forwarding.ml
edwintorok: It performs the same operation as remove, it kicks the member out of the cluster.

ocaml/xapi/message_forwarding.ml
edwintorok: Yes we need to find a live host somehow

ocaml/xapi/message_forwarding.ml
edwintorok: Yes we need to find a live host somehow, we need to retry if we get a HOST_OFFLINE error.

ocaml/idl/datamodel_cluster.ml
edwintorok: Actually that doesn't work, it would hide the field from records.ml too and the API, and we need it visible from the CLI for debugging.

gaborigloi: Hmm, I've asked the others. It would still be nice to make it `internal_only` somehow, so this does not show up to the user as a new field. But I guess it's not a disaster if it does, because it still conveys some useful information to them.

mseri: You can at least mention in the documentation that this is not meant to be used and it is for internal use only (like VM.create for example)

edwintorok: Added a fixup to mention this is an internal field, it is RO so the user won't be able to change it. And in fact its not bad that they can see it: the admin asked one host to be destroyed (permanently dead), but it is still part of the cluster as far as quorum is concerned until he declares all the unreachable hosts as permanently dead.
(to declare a host as permanently dead all the other members of the cluster must confirm, which they can't do if they're offline).

So if the admin wants to find out who are the members of the cluster at a particular moment in time they can look at the "cluster_hosts" field plus the "pending_forget" field. (the host/pif objects will be destroyed so if we didn't preserve the IP in pending_forget there would be no way to find out). (or they can run the low-level diagnostic commands from the clustering daemon)

mseri: Why do you need to add a manual newline there? I think we don't use it anywhere else in the messages

edwintorok: Indeed I just need a \ to join the strings but not an actual newline in the output, dropped.

mseri:
Remember to squash the fixups before merging!

ocaml/tests/test_cluster_host.ml
```diff
@@ -122,11 +122,91 @@ let test_destroy_forbidden_when_sr_attached () =
...
+  let __context = Test_common.make_test_database () in
+  let host2 = Test_common.make_host ~__context () in
+  let host3 = Test_common.make_host __context () in
+  let cluster, original_cluster_hosts = make ~__context [host2; host3] in
+
+  Xapi_host.destroy ~__context ~self:host3;
+
+  let pending = Db.Cluster.get_pending_forget ~__context ~self:cluster in
+  Alcotest.(check (list string) "1 pending forgets"
+              ["192.0.2.3"] pending);
```
lindig: This IP address is in a public address space. I think it's a good practice to use private spaces like `192.168.x.x`.

ocaml/tests/test_common.ml
```diff
@@ -500,8 +500,23 @@ let make_cluster_and_cluster_host ~__context ?(ref=Ref.make ()) ?(uuid=make_uuid
...
     ~token_timeout ~token_timeout_coefficient ~cluster_config ~other_config;
   let cluster_host_ref = make_cluster_host ~__context ~cluster:ref ~host ~pIF () in
   ref, cluster_host_ref
+
+let make_cluster_and_hosts ~__context extra_hosts =
+  let cluster_stack = "mock_cluster_stack" in
+  let network = make_network ~__context () in
+
+  let host = Helpers.get_localhost ~__context in
+  let pIF = make_pif ~__context ~network ~host ~iP:"192.0.2.1" () in
```
lindig: See above - I'd use a private address space to avoid any potential that this gets out.

ocaml/tests/test_cluster_host.ml
edwintorok: I used the IP address space recommended for documentation, they are reserved and should not be routable: https://tools.ietf.org/html/rfc5737. I agree that `192.168.x.x` is more well known, could change it to that as it is only used in a unit test.

lindig: That's cool. I did not know that these are also not routed.

PR #3596, opened 2018-05-17
CA-289319: Kolkata Update application failure: HANDLE_INVALID

krizex:
There is a race between post update-apply cleanup and db_gc.
To avoid the race condition, destroy the connector (Host_patch) before
destroying Pool_patch objects and do not call db_gc in post update-apply cleanup
procedure.


jonludlam:
Looks good to me. Thanks!

robhoes:
Travis says "yes". Merging.

PR #3598, opened 2018-05-17
CP-26199: Send alerts when Cluster_host.resync_host fails

minishrink:
`Cluster_host.resync_host` is iterated over all hosts in a pool by `Cluster.pool_resync`, and is also called at startup as part of the task `resync cluster host state`.

Question: should the new messages have a different priority? Also open to suggestions to make the formatting less awful.

### DIFF
--- a/ocaml/xapi-consts/api_messages.ml
+++ b/ocaml/xapi-consts/api_messages.ml
@@ -132,3 +132,7 @@ let host_cpu_features_down = addMessage "HOST_CPU_FEATURES_DOWN" 3L
 let host_cpu_features_up = addMessage "HOST_CPU_FEATURES_UP" 5L
 let pool_cpu_features_down = addMessage "POOL_CPU_FEATURES_DOWN" 5L
 let pool_cpu_features_up = addMessage "POOL_CPU_FEATURES_UP" 5L
+
+(* Cluster messages *)
+let cluster_host_creation_failed = addMessage "CLUSTER_HOST_CREATION_FAILED" 3L
+let cluster_host_enable_failed = addMessage "CLUSTER_HOST_ENABLE_FAILED" 3L
--- a/ocaml/xapi/xapi_cluster_host.ml
+++ b/ocaml/xapi/xapi_cluster_host.ml
@@ -53,30 +53,57 @@ let sync_required ~__context ~host =
     end
   | _ -> raise Api_errors.(Server_error (internal_error, ["Cannot have more than one Cluster object per pool currently"]))
 
+let call_api_function_with_alert ~__context ~msg ~cls ~obj_uuid ~body
+  ~(api_func : (Rpc.call -> Rpc.response) -> API.ref_session -> unit) =
+    Helpers.call_api_functions ~__context
+      (fun rpc session_id ->
+        try
+          api_func rpc session_id
+        with err ->
+          let body = Printf.sprintf "Error: %s\nMessage: %s" ExnHelper.(string_of_exn err) body in
+          Xapi_alert.add ~msg ~cls ~obj_uuid ~body;
+          raise err
+      )
+
 let create_as_necessary ~__context ~host =
   match sync_required ~__context ~host with
   | Some cluster_ref -> (* assume pool autojoin set *)
     let network = get_network_internal ~__context ~self:cluster_ref in
     let (pifref,pifrec) = Xapi_clustering.pif_of_host ~__context network host in
     fix_pif_prerequisites ~__context (pifref,pifrec);
-    Helpers.call_api_functions ~__context (fun rpc session_id ->
-        Client.Client.Cluster_host.create rpc session_id cluster_ref host pifref) |> ignore
+
+    (* Best effort Cluster_host.create *)
+    let body = Printf.sprintf "Unable to create cluster host on %s."
+        (Db.Host.get_name_label ~__context ~self:host) in
+    let obj_uuid = Db.Host.get_uuid ~__context ~self:host in
+    call_api_function_with_alert ~__context
+      ~msg:Api_messages.cluster_host_creation_failed
+      ~cls:`Host ~obj_uuid ~body
+      ~api_func:(fun rpc session_id -> ignore
+          (Client.Client.Cluster_host.create rpc session_id cluster_ref host pifref))
   | None -> ()
 
 let resync_host ~__context ~host =
   create_as_necessary ~__context ~host;
   match (find_cluster_host ~__context ~host) with
-    | None              -> () (* no clusters exist *)
-    | Some cluster_host ->    (* cluster_host and cluster exist *)
-      if Db.Cluster_host.get_enabled ~__context ~self:cluster_host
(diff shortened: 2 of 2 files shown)


ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -59,19 +59,46 @@ let create_as_necessary ~__context ~host =
     let network = get_network_internal ~__context ~self:cluster_ref in
     let (pifref,pifrec) = Xapi_clustering.pif_of_host ~__context network host in
     fix_pif_prerequisites ~__context (pifref,pifrec);
-    Helpers.call_api_functions ~__context (fun rpc session_id ->
-        Client.Client.Cluster_host.create rpc session_id cluster_ref host pifref) |> ignore
+    Helpers.call_api_functions ~__context
+      (fun rpc session_id ->
+        try
+          Client.Client.Cluster_host.create rpc session_id cluster_ref host pifref
+        with Api_errors.(Server_error (_, _)) as err ->
```
lindig: If a different error surfaces, should this be logged? Or can we ignore them here and let them escape? The same applies below.

ocaml/xapi-consts/api_messages.ml
```diff
@@ -132,3 +132,7 @@ let host_cpu_features_down = addMessage "HOST_CPU_FEATURES_DOWN" 3L
 let host_cpu_features_up = addMessage "HOST_CPU_FEATURES_UP" 5L
 let pool_cpu_features_down = addMessage "POOL_CPU_FEATURES_DOWN" 5L
 let pool_cpu_features_up = addMessage "POOL_CPU_FEATURES_UP" 5L
+
+(* Cluster messages *)
+let cluster_host_creation_failed = addMessage "CLUSTER_HOST_CREATION_FAILED" 3L
+let cluster_host_enable_failed = addMessage "CLUSTER_HOST_ENABLE_FAILED" 4L
```
edwintorok: If it failed the priority should not be 4, because 4 means it has recovered, see the table at the top of the file

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -59,19 +59,46 @@ let create_as_necessary ~__context ~host =
...
+          let body =
+            (Printf.sprintf
+              "Unable to create cluster host on %s"
```
edwintorok: This should also include a formatted error message, just knowing it failed won't help the admin fix the problem.

minishrink: If you mean it should include the error doc in datamodel_errors.ml, I know this is printed in the CLI, though I'm not sure how XenCenter presents it.

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -53,25 +53,58 @@ let sync_required ~__context ~host =
     end
   | _ -> raise Api_errors.(Server_error (internal_error, ["Cannot have more than one Cluster object per pool currently"]))
 
+let call_api_function_with_alert ~__context ~msg ~cls ~obj_uuid ~body ~(api_func : (Rpc.call -> Rpc.response) -> API.ref_session -> 'a) =
+  Helpers.call_api_functions ~__context
+    (fun rpc session_id ->
+       try
+         api_func rpc session_id |> ignore
```
edwintorok: I think it is better if `ignore` is not used, and the value is ignored in the caller where we know the actual type.
Otherwise this could mistakenly be a partial application that never runs code at runtime.

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -53,25 +53,58 @@ let sync_required ~__context ~host =
...
+       with err ->
+         error "%s" ExnHelper.(string_of_exn err);
+         Xapi_alert.add ~msg ~cls ~obj_uuid ~body;
```
edwintorok: Would be good if the body included the error message too, maybe concatenate it? `body ^ ": " ^ Exnhelper...)

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -53,25 +53,58 @@ let sync_required ~__context ~host =
...
-    Helpers.call_api_functions ~__context (fun rpc session_id ->
-        Client.Client.Cluster_host.create rpc session_id cluster_ref host pifref) |> ignore
+    let api_func =
+      (fun rpc session_id ->
+         Client.Client.Cluster_host.create rpc session_id cluster_ref host pifref
+      ) in
+    let body =
+      (Printf.sprintf
+         "Unable to create cluster host on %s."
+         (Db.Host.get_name_label ~__context ~self:host)
```
edwintorok: Minor comment: does this need to be split over so many lines? the parenthesis are not necessary, how long would it be if you put the `let body`, `Printf`, and format string on same line?

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -53,25 +53,58 @@ let sync_required ~__context ~host =
...
+        ) in
+      let body =
+        (Printf.sprintf
+           "Unable to create cluster host on %s"
+           (Db.Host.get_name_label ~__context ~self:host)
+        ) in
+      let obj_uuid = Db.Host.get_uuid ~__context ~self:host in
+      call_api_function_with_alert ~__context
+        ~msg:Api_messages.cluster_host_creation_failed
+        ~cls:`Host ~obj_uuid ~body ~api_func
```
edwintorok: `api_func` could probably just be declared directly here, like it was for `Helpers.call_api_functions`. i.e. you would replace the Helpers.call_api_functions with `call_api_function_with_alert`.

edwintorok:
You also got some new conflicts to solve now.

ocaml/xapi/xapi_cluster_host.ml
minishrink: Should we keep the error then?

edwintorok: No, I think it'll get logged by another exception handler up the call stack.

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -53,30 +53,56 @@ let sync_required ~__context ~host =
...
   | _ -> raise Api_errors.(Server_error (internal_error, ["Cannot have more than one Cluster object per pool currently"]))
 
+let call_api_function_with_alert ~__context ~msg ~cls ~obj_uuid ~body
+  ~(api_func : (Rpc.call -> Rpc.response) -> API.ref_session -> unit) =
+    Helpers.call_api_functions ~__context
+      (fun rpc session_id ->
+        try
+          api_func rpc session_id
+        with err ->
+          let body = Printf.sprintf "Error: %s\nMessage: %s" ExnHelper.(string_of_exn err) body in
```
edwintorok: Not sure how XenCenter handles messages with newlines, can you try to see if it looks alright?

minishrink: Apparently messages can only be one line, so I don't think I can include the error.

PR #3599, opened 2018-05-18
pci/lib_test: disable tests until we move back to upstream library

mseri:
This should fix the ocaml side of xapi Travis tests. I am still investigating how to fix python's tests


edwintorok:
The ocaml build succeeded, something wrong with python though
```
 E: Unable to locate package python-mock
E: Unable to locate package python-nose
```

PR #3603, opened 2018-05-18
quicktests: add support for comparing record fields

gaborigloi:


gaborigloi:
I've tested this and the quicktests passed on GFS2, NFS, and local ext

PR #3604, opened 2018-05-18
CP-24692: Prevent date race between allowing PIF.unplug and clustering

minishrink:
`PIF.disallow_unplug:true` is necessary for the entire time a cluster host is enabled, as well as being a prerequisite for the API calls `Cluster_host.{enable, create}`.

If `disallow_unplug` is set to false during either of these two clustering operations, it could create a race condition where both operations are allowed, but we then have a cluster_host that doesn't meet networking requirements and will therefore fail later on.

This PR introduces a condition wherein you cannot allow PIF.unplug if the pool cluster is currently creating or enable a cluster_host on *any* PIF, regardless of which PIF you are allowing unplug for. This should rarely cause an issue, as `PIF.set_disallow_unplug` is rarely called, although an operation like `Cluster.pool_{create, resync}` could block it for a while, depending on the pool size.


ocaml/idl/datamodel_errors.ml
```diff
@@ -1141,7 +1141,10 @@ let _ =
   error Api_errors.invalid_cluster_stack [ "cluster_stack" ]
     ~doc:"The cluster stack provided is not supported." ();
   error Api_errors.pif_not_attached_to_host [ "pif"; "host" ]
-    ~doc:"Cluster_host creation failed as the PIF provided is not attached to the host." ()
+    ~doc:"Cluster_host creation failed as the PIF provided is not attached to the host." ();
+  error Api_errors.clustering_operation_in_progress []
```
minishrink: Should we include a clustering operation string here?

edwintorok: Can you reuse other_operation_in_progress?

minishrink: Yes, but then I need to provide a ref to the cluster_host in use. Should I just include a specific error string in there instead, since blocking a PIF operation with "another clustering operation in progress"  might be confusing?

minishrink: e.g. `["Cluster_host"; "Cannot allow unplug on any PIF if enabling or creating a cluster host on any PIF"]`

minishrink:
@edwintorok updated

ocaml/idl/datamodel_errors.ml
edwintorok: ` error Api_errors.other_operation_in_progress ["class"; "object"]` you need to give a ref as the 2nd argument to `other_operation_in_progress`, not an arbitrary string.

minishrink: Should I just specify the PIF? I think that would be confusing for a user. We don't access the specific cluster_host in this, otherwise we could just simply block allowing unplug for that specific PIF.

minishrink: Unless we can get it from the task list?

edwintorok: Yeah just provide the cluster ref.

minishrink: Cool, done. Is this good to merge?

edwintorok: I think you can drop `clustering_operation_in_progress`, you are not using it anymore.

minishrink: Switched to `other_operation_in_progress`

PR #3605, opened 2018-05-21
XSI-6: Corrected and completed event class documentation enhancements. 

kc284:
Note I have not corrected jsapi.ml because this is only used by the SDK and I'm planning to deprecate it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ doc:
 
 doc-json:
 	jbuilder build ocaml/idl/json_backend/gen_json.exe
-	jbuilder exec -- ocaml/idl/json_backend/gen_json.exe
+	jbuilder exec -- ocaml/idl/json_backend/gen_json.exe -destdir _build/install/default/jekyll
 
 reindent:
 	git ls-files '*.ml*' '**/*.ml*' | xargs ocp-indent --syntax cstruct -i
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -24,7 +24,7 @@ doc:
 
 doc-json:
 	jbuilder build ocaml/idl/json_backend/gen_json.exe
-	jbuilder exec -- ocaml/idl/json_backend/gen_json.exe
+	jbuilder exec -- ocaml/idl/json_backend/gen_json.exe -destdir _build/install/default/jekyll
```
robhoes: Does something call `mkdir -p` for that directory?

ocaml/idl/json_backend/gen_json.ml
```diff
@@ -330,10 +339,14 @@ let releases objs =
...
   in
-  let release_info = JObject (List.map (fun rel -> code_name_of_release rel, changes_in_release rel) release_order) in
-  Xapi_stdext_unix.Unixext.write_string_to_file ("release_info.json") (string_of_json 0 release_info)
+  JObject (List.map (fun rel -> code_name_of_release rel, changes_in_release rel) release_order)
+
 
 let _ =
+  parse_args ();
+  let destdir = !destdir' in
+  Xapi_stdext_unix.Unixext.mkdir_rec destdir 0o755;
```
robhoes: Ah, there it is :)

PR #3606, opened 2018-05-21
Quicktest: check snapshot VDI fields

gaborigloi:
And reorganize the helpers a bit.

This test has already found an issue, CA-290241, and it's passing on local ext and NFS.

### DIFF
--- a/ocaml/quicktest/quicktest_common.ml
+++ b/ocaml/quicktest/quicktest_common.ml
@@ -34,22 +34,6 @@ let http request f =
     else SSL(SSL.make ~use_fork_exec_helper:false (), !Quicktest_args.host, 443) in
   with_transport transport (with_http request f)
 
-let vm_template = "Demo Linux VM"
-let other = "Other install media"
-
-exception Unable_to_find_suitable_vm_template
-
-let find_template session_id startswith =
-  let vms = Client.VM.get_all !rpc session_id in
-  match List.filter (fun self ->
-      (String.startswith startswith (Client.VM.get_name_label !rpc session_id self))
-      && (Client.VM.get_is_a_template !rpc session_id self)
-    ) vms with
-  | [] -> raise Unable_to_find_suitable_vm_template
-  | x :: _ ->
-    (* Printf.printf "Choosing template with name: %s\n" (Client.VM.get_name_label !rpc session_id x); *)
-    x
-
 let cli_cmd args =
   print_endline (String.concat " " ("$ xe" :: args));
   try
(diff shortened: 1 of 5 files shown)


ocaml/quicktest/quicktest_common.ml
```diff
@@ -117,7 +98,23 @@ let compare_fields cls fields original_rec new_rec =
...
-          "%s field %s should should be different, but is the same: '%s'"
+          "%s field %s should be different, but is the same: '%s'"
           cls field_name new_field
   in
   List.iter check fields
+
+module Time = struct
+  type t = float
+
+  let now () = Unix.gettimeofday ()
```
mseri: I think you also have Mtime at your disposal if you want something even more reliable

ocaml/quicktest/quicktest_common.ml
```diff
@@ -64,35 +48,32 @@ let cli_cmd args =
...
-  match List.filter (fun (_, r) -> List.mem_assoc Xapi_globs.is_guest_installer_network r.API.network_other_config) all with
-  | (rf, _) :: _ -> rf
-  | _ -> failwith "Could not find guest installer network"
-
-(** Return a host's domain zero *)
-let dom0_of_host session_id host =
-  Client.Host.get_control_domain !rpc session_id host
-
-let get_default_sr session_id =
-  Client.Pool.get_default_SR ~session_id ~rpc:!rpc ~self:(get_pool session_id)
```
mseri: So this was unused!

ocaml/quicktest/quicktest_http.ml
```diff
@@ -59,14 +59,6 @@ end
...
 module Secret_Auth_fails = struct (* {{{1 *)
 
-  let http request f =
-    let open Xmlrpc_client in
-    let transport =
-      if !Quicktest_args.using_unix_domain_socket
-      then Unix Xapi_globs.unix_domain_socket
-      else SSL(SSL.make ~use_fork_exec_helper:false (), !Quicktest_args.host, 443) in
-    with_transport transport (with_http request f)
-
```
minishrink: I take it this is obsolete?

gaborigloi: No, it was duplicated in quicktest_common

ocaml/quicktest/quicktest_common.ml
gaborigloi: I used stdext-date because that type is stored in the VDI record

PR #3607, opened 2018-05-21
CA-290237, CA-290473: fix pool-join on bonded networks and VLANs

minishrink:
This PR cleans up some of the breakages exposed by the recent bonding enabling (CP-26912).

Changes made to:

**startup sequence**
in order:
- create networking objects
- create cluster_host object (if necessary)
- wait for clustering IP
- resync cluster_host

**cluster_host operations**
- add `joined` field to represent cluster membership of host
- block all but Cluster_host.destroy (leave) if not joined
- separate Cluster_host.create into internal db and client calls

Consequences:
- `pool-join` should now work for bonded networks as well as VLANs


minishrink:
Submitted a Jenkins branch with the same name

jonludlam:
Yeah, the startup ordering stuff is really really fragile. Don't change it unless you must, and if you do, be prepared to spend a while debugging nasty corner cases :-(

edwintorok:
It might be better to fix this in another way that should also fix another bug with clustering not working on startup when the bond gets an IP too late:
* have a `wait_for_clustering_ip` function that is similar to `wait_for_management_ip`, except it waits for an IP address on the network used by the master's PIF, if there is no clustering it does nothing
* once it gets an IP it calls Cluster_host.resync_host (which does create_as_necessary), and this would allow corosync to succeed enabling itself (it'll fail to enable if we give it an IP address that is not yet assigned to an interface)
* after that call plug_unplugged_pbds

ocaml/xapi/xapi.ml
```diff
@@ -924,6 +922,8 @@ let server_init() =
           "Synchronising network sriovs on slave with master", [Startup.OnlySlave; Startup.NoExnRaising], Sync_networking.copy_network_sriovs_from_master ~__context;
           "Synchronising VLANs on slave with master", [Startup.OnlySlave; Startup.NoExnRaising], Sync_networking.copy_vlans_from_master ~__context;
           "Synchronising tunnels on slave with master", [Startup.OnlySlave; Startup.NoExnRaising], Sync_networking.copy_tunnels_from_master ~__context;
+          (* CA-290237: Create cluster objects after network objects so PIFs are available *)
+          "Create any necessary cluster_host objects", [ Startup.NoExnRaising ], (fun () -> Xapi_cluster_host.create_as_necessary __context (Helpers.get_localhost ~__context));
```
edwintorok: Please use `log_and_ignore_exn` here, so if something goes wrong XAPI doesn't go into a restart loop.

ocaml/xapi/xapi.ml
```diff
@@ -941,22 +941,34 @@ let server_init() =
...
-          )
+          end
         in
 
         Startup.run ~__context [
           "fetching database backup", [ Startup.OnlySlave; Startup.NoExnRaising ],
           (fun () -> Pool_db_backup.fetch_database_backup ~master_address:(Pool_role.get_master_address())
               ~pool_secret:!Xapi_globs.pool_secret ~force:None);
-          "wait management interface to come up", [ Startup.NoExnRaising ], wait_management_interface;
+          "wait management interface to come up, resync any cluster_hosts created", [ Startup.NoExnRaising ], wait_management_interface;
```
edwintorok: ... and plug any unplugged PBDs.

ocaml/xapi/xapi_pbd.ml
```diff
@@ -122,7 +122,10 @@ let resync_cluster_stack_for_sr_type ~__context ~sr_sm_type ~host =
   (* check cluster_host associated with both the host and a cluster with a matching cluster_stack *)
     let self  = Db.Cluster_host.get_cluster ~__context ~self:cluster_host in
     if List.mem (Db.Cluster.get_cluster_stack ~__context ~self) required_cluster_stacks
-    then Xapi_cluster_host.resync_host ~__context ~host
+    then begin
+      Xapi_cluster_host.create_as_necessary ~__context ~host;
```
edwintorok: I don't think that the object can be missing here, it got created during the startup sequence, or during join, this one probably doesn't need to be changed, and calling just resync_host is fine.

ocaml/xapi/xapi.ml
```diff
@@ -941,22 +941,34 @@ let server_init() =
                                    (fun () -> Debug_populate.do_populate ~vms:1000 ~vdis_per_vm:3 ~networks:10 ~srs:10 ~tasks:1000) ]
         );
 
+        let maybe_resync_cluster_host () =
+          let host = Helpers.get_localhost ~__context in
+          begin match Xapi_clustering.find_cluster_host ~__context ~host with
+          | Some self -> ignore
+            Xapi_mgmt_iface.(wait_for_clustering_ip ~__context ~self);
+            Xapi_cluster_host.resync_host ~__context ~host
```
edwintorok: Plugging the PBD already calls resync_host, so the only thing we need here is to wait for an IP.
Rename this to `maybe_wait_for_clustering_ip`.

ocaml/xapi/xapi.ml
edwintorok: Since this is a blocking action, if something goes wrong and the networking interface of clustering doesn't come up, it would be nice if we didn't block the whole of XAPI.
So perhaps introduce a second entry here below 'wait for management interface', that says "wait for clustering IP if any, and plug unplugged PBDs" that calls your `maybe_wait_for_clustering_ip` function above.

edwintorok:
I think the logic should work now, the commit message should probably say this about the order
- create network objects (unmodified)
- create cluster host if need be
...
- wait for management IP (unmodified)
- wait for clustering IP, plug any unplugged PBDs which would also resync the cluster host if needed

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -157,20 +157,33 @@ let enable_himn ~__context ~addr =
...
+(* CA-280237: Called in startup sequence after creating cluster_hosts *)
+let wait_for_clustering_ip ~__context ~(self : API.ref_Cluster_host) =
+  let pIF = Db.Cluster_host.get_PIF ~__context ~self in
+  let iP = ref (Db.PIF.get_IP ~__context ~self:pIF) in
+  let attached = ref (Db.PIF.get_currently_attached ~__context ~self:pIF) in
+  Mutex.execute ip_mutex (* Don't return until PIF is plugged AND has a valid IP *)
+    (fun () -> while !iP="" || not !attached do
+        Condition.wait ip_cond ip_mutex;
+        iP := Db.PIF.get_IP ~__context ~self:pIF;
+        attached := Db.PIF.get_currently_attached ~__context ~self:pIF
```
robhoes: This does do the same as `get_management_iface_is_connected`, which is checking the PIF carrier (link state). You can use `PIF_metrics.carrier` for that.

ocaml/xapi/xapi.ml
```diff
@@ -924,6 +922,9 @@ let server_init() =
           "Synchronising network sriovs on slave with master", [Startup.OnlySlave; Startup.NoExnRaising], Sync_networking.copy_network_sriovs_from_master ~__context;
           "Synchronising VLANs on slave with master", [Startup.OnlySlave; Startup.NoExnRaising], Sync_networking.copy_vlans_from_master ~__context;
           "Synchronising tunnels on slave with master", [Startup.OnlySlave; Startup.NoExnRaising], Sync_networking.copy_tunnels_from_master ~__context;
+          (* CA-290237: Create cluster objects after network objects so PIFs are available *)
+          "Create any necessary cluster_host objects", [ Startup.NoExnRaising ],
```
edwintorok: @minishrink we'll need to move this after we got an IP. Creating a cluster host object checks for the presence of an IP, and you get a NO_COMPATIBLE_CLUSTER_HOST error, caused by the create failing due to PIF_HAS_NO_NETWORK_CONFIGURATION

edwintorok: Creating the object not only creates the object in the Db but also tries to enable it, and creates the object in the db only if enabling succeeded.

edwintorok: Please add CA-290473 in the commit message when this is fixed.

edwintorok: Actually it might be better to have a function here just create the database object, and have the actual join/enable done elsewhere (in resync_host).
For that to work:
* add a new field to cluster_host called joined, default to true for backwards compat
* refactor `Cluster_host.create` into a create_db and create (where create calls create_db and then resync_host)
* `resync_host` will look at the joined field: if true then it calls daemon.enable+join, and sets joined to true if successful, otherwise it calls enable if enabled
* in the startup sequence here only call create_db

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -65,75 +46,74 @@ let call_api_function_with_alert ~__context ~msg ~cls ~obj_uuid ~body
...
+          let ip_list = List.map (fun self ->
+            let pref = Db.Cluster_host.get_PIF ~__context ~self in
+            let prec = Db.PIF.get_record ~__context ~self:pref in
+              ip_of_pif (pref,prec)
+            ) (Db.Cluster.get_cluster_hosts ~__context ~self:cluster) in
+          let result = Cluster_client.LocalClient.join (rpc ~__context) dbg cluster_token ip ip_list in
+          match result with
+          | Result.Ok () ->
+            Db.Cluster_host.set_joined ~__context ~self ~value:true;
+            debug "Cluster_host.create was successful; cluster_host: %s" (Ref.string_of self);
```
edwintorok: `join` was successful

ocaml/xapi/xapi.ml
```diff
@@ -956,7 +956,9 @@ let server_init() =
           | Some self ->
             debug "Waiting forever for cluster_host to gain an IP address";
             let ip = Xapi_mgmt_iface.(wait_for_clustering_ip ~__context ~self) in
-            debug "Got clustering IP %s, attempting to re-plug any remaining unplugged PBDs" ip;
+            debug "Got clustering IP %s, resyncing cluster_host %s" ip (Ref.string_of self);
+            Xapi_cluster_host.resync_host ~__context ~host;
```
edwintorok: Should not be needed, PBD plug already does this. I think you can drop this commit.

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -218,3 +198,31 @@ let disable_clustering ~__context =
...
+      | [] ->
+        if cluster_rec.API.cluster_pool_auto_join
+        then Some cluster_ref
+        else None
+      | _ -> raise Api_errors.(Server_error (internal_error, [ "Host cannot be associated with more than one cluster_host"; Ref.string_of host ]))
+    end
+  | _ -> raise Api_errors.(Server_error (internal_error, ["Cannot have more than one Cluster object per pool currently"]))
+
+(* If cluster found without local cluster_host, create one in db *)
+let create_as_necessary ~__context ~host =
```
edwintorok: Do you need to update the descriptions in the mli?

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -112,29 +93,68 @@ let create ~__context ~cluster ~host ~pif =
...
+            warn "Error occurred during Cluster_host.create";
+            handle_error error
+        end;
+        if Db.Cluster_host.get_enabled ~__context ~self then begin
+        (* [enable] unconditionally invokes low-level enable operations and is idempotent. *)
+          (* RPU reformats partition, losing service status, never re-enables clusterd *)
+          debug "Cluster_host %s is enabled, starting up xapi-clusterd" (Ref.string_of self);
+          Xapi_clustering.Daemon.enable ~__context;
+
+          (* Best-effort Cluster_host.enable *)
```
edwintorok: This doesn't seem to be best-effort anymore, if it fails the whole function fails. Thats fine, just remove the comment.

ocaml/xapi/create_storage.ml
```diff
@@ -41,17 +41,11 @@ let plug_all_pbds __context =
...
-  match Xapi_clustering.find_cluster_host ~__context ~host with
-  | Some self ->
-     Xapi_cluster_host.enable ~__context ~self
-  | None -> ()
 
 let plug_unplugged_pbds __context =
   (* If the plug is to succeed for SM's requiring a cluster stack
    * we have to enable the cluster stack too if we have one *)
-  log_and_ignore_exn(fun () -> maybe_reenable_cluster_host __context);
+  log_and_ignore_exn (fun () -> Xapi_cluster_host.resync_host ~__context ~host:Helpers.(get_localhost ~__context));
```
edwintorok: This line should be deleted completely. We resync when we get the IP, and in cluster-pool-resync. Thats it.

ocaml/xapi/xapi.ml
```diff
@@ -944,7 +944,9 @@ let server_init() =
           if management_if <> "" then begin
             debug "Waiting forever for the management interface to gain an IP address";
             let ip = wait_for_management_ip_address ~__context in
-            debug "Management interface got IP address: %s; attempting to re-plug any unplugged PBDs" ip;
+            debug "Management interface got IP address: %s, attempting to re-plug unplugged PBDs" ip;
+            (* This may fail without the clustering IP, which is why we attempt
+               another replugging in maybe_wait_for_clustering_ip *)
```
edwintorok: "attempt another plug"

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -65,76 +46,80 @@ let call_api_function_with_alert ~__context ~msg ~cls ~obj_uuid ~body
...
+  | Some self ->    (* cluster_host and cluster exist *)
+    if not (Db.Cluster_host.get_joined ~__context ~self) then begin
+       match join_internal ~__context ~self with
       | Result.Ok () ->
-        Db.Cluster_host.create ~__context ~ref ~uuid ~cluster ~host ~pIF:pifref ~enabled:true
-          ~current_operations:[] ~allowed_operations:[] ~other_config:[];
-        debug "Cluster_host.create was successful; cluster_host: %s" (Ref.string_of ref);
-        ref
+        Db.Cluster_host.set_joined ~__context ~self ~value:true;
+        Db.Cluster_host.set_enabled ~__context ~self ~value:true;
```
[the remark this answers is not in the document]
edwintorok: Please move these into join_internal

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -65,76 +46,80 @@ let call_api_function_with_alert ~__context ~msg ~cls ~obj_uuid ~body
...
+      Cluster_client.LocalClient.join (rpc ~__context) dbg cluster_token ip ip_list
+  )
+
+(* Enable cluster_host in client layer via clusterd *)
+let resync_host ~__context ~host =
+  match (find_cluster_host ~__context ~host) with
+  | None      -> () (* no clusters exist *)
+  | Some self ->    (* cluster_host and cluster exist *)
+    if not (Db.Cluster_host.get_joined ~__context ~self) then begin
+       match join_internal ~__context ~self with
```
edwintorok: We probably want the alert wrapper from enable here as well, so it would be better to wrap this whole part below `Some self` with `call_api_functions_with_alert`

ocaml/xapi/xapi_pbd.ml
```diff
@@ -213,7 +200,7 @@ let unplug ~__context ~self =
 
 let destroy ~__context ~self =
   if Db.PBD.get_currently_attached ~__context ~self
-  then raise (Api_errors.Server_error(Api_errors.operation_not_allowed, ["PBD is currently attached"]));
+  then raise Api_errors.(Server_error (operation_not_allowed, ["PBD is currently attached"]));
```
edwintorok: There are some unrelated cleanups in these commits. Thats fine, but could you move them to a separate commit?

ocaml/xapi/xapi_cluster_helpers.ml
```diff
@@ -59,8 +59,15 @@ let get_operation_error ~__context ~self ~op =
 
 let assert_operation_valid ~__context ~self ~op =
   begin match (Helpers.rolling_upgrade_in_progress ~__context), op with
-    | true, (`add | `remove | `destroy) -> raise Api_errors.(Server_error (not_supported_during_upgrade, []))
-    | _, _                              -> ()
+    | true, (`add | `remove | `destroy) ->
+      raise Api_errors.(Server_error (not_supported_during_upgrade, []))
+    | _, `remove -> ()
+    | _, _ -> begin
+      match Xapi_clustering.find_cluster_host ~__context ~host:Helpers.(get_localhost ~__context) with
```
edwintorok: Shouldn't this check be done on the cluster host's `assert_operation_valid`? Otherwise you only ever look at the master here.

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -79,6 +79,8 @@ let join_internal ~__context ~self =
 
       debug "Enabling clusterd and joining cluster_host %s" (Ref.string_of self);
       Xapi_clustering.Daemon.enable ~__context;
+      Db.Cluster_host.set_joined ~__context ~self ~value:true;
```
edwintorok: Please move these after the call to join: the call to join might fail and we want to set joined to true only if we actually succeeded.
So match on result Ok/Error should be inhere as well

minishrink: So `resync_host` just calls `join_internal`, then continues the `if enabled then` block? That works better actually.

ocaml/xapi/xapi_cluster_helpers.ml
minishrink: The problem is that the cluster_host helpers only recognise the `enable` and `disable` enums

edwintorok: Actually thats fine: we should be able to destroy a cluster host even if we never joined it.
Its just that the code in destroy will probably have to special case it, and if it was never joined, then do nothing on clusterd side, just delete the object from the DB.

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -65,76 +46,80 @@ let call_api_function_with_alert ~__context ~msg ~cls ~obj_uuid ~body
...
-        debug "Cluster_host.create was successful; cluster_host: %s" (Ref.string_of ref);
-        ref
+        debug "Cluster_host.create was successful; cluster_host: %s" (Ref.string_of self)
       | Result.Error error ->
         warn "Error occurred during Cluster_host.create";
         handle_error error
-    )
+    end;
+
+    if Db.Cluster_host.get_enabled ~__context ~self then begin
```
edwintorok: Sorry, one more change here: if we have just joined we should not try to go and enable it immediately: it will fail with other operation in progress (we are the other operation: Cluster.add).

So there should be an `else` here.

ocaml/xapi/xapi_cluster_host.ml
minishrink: For `LocalClient.join`, I assume?

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -246,3 +231,31 @@ let disable_clustering ~__context =
...
+    end
+  | _ -> raise Api_errors.(Server_error (internal_error, ["Cannot have more than one Cluster object per pool currently"]))
+
+(* If cluster found without local cluster_host, create one in db *)
+let create_as_necessary ~__context ~host =
+  match sync_required ~__context ~host with
+  | Some cluster -> (* assume pool autojoin set *)
+    let network = get_network_internal ~__context ~self:cluster in
+    let (pIF,pifrec) = Xapi_clustering.pif_of_host ~__context network host in
+    fix_pif_prerequisites ~__context (pIF,pifrec);
```
edwintorok: This is in the wrong place, it'll fail: we have no IP on startup where we create the object, this should be part of join.

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -65,76 +46,80 @@ let call_api_function_with_alert ~__context ~msg ~cls ~obj_uuid ~body
...
   with_clustering_lock (fun () ->
       assert_operation_host_target_is_localhost ~__context ~host;
       assert_cluster_host_can_be_created ~__context ~host;
-      assert_pif_attached_to ~host ~pif ~__context;
+      assert_pif_attached_to ~host ~pIF ~__context;
       let ref = Ref.make () in
-      let dbg = Context.string_of_task __context in
       let uuid = Uuidm.to_string (Uuidm.create `V4) in
+      let pifrec = Db.PIF.get_record ~__context ~self:pIF in
+      assert_pif_prerequisites (pIF,pifrec);
```
edwintorok: this will fail, should be part of join

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -65,76 +46,80 @@ let call_api_function_with_alert ~__context ~msg ~cls ~obj_uuid ~body
...
-
-let create ~__context ~cluster ~host ~pif =
-  (* TODO: take network lock *)
+(* Create xapi db object for cluster_host, resync_host calls clusterd *)
+let create_internal ~__context ~cluster ~host ~pIF : API.ref_Cluster_host =
   with_clustering_lock (fun () ->
       assert_operation_host_target_is_localhost ~__context ~host;
       assert_cluster_host_can_be_created ~__context ~host;
-      assert_pif_attached_to ~host ~pif ~__context;
+      assert_pif_attached_to ~host ~pIF ~__context;
```
edwintorok: this will fail, should be part of join.

edwintorok: Can you add a unit test that tries to create a cluster host object when the PIF exists, but the IP is empty?

ocaml/xapi/xapi_cluster_host.ml
edwintorok: Yes, but I think you can wrap things like this:
```
let resync_host ... =
  match ...
  | None -> ...
 | Some self ->
    call_api_functions_with_alert ~__context (fun ... ->
     if ...
         join ...
     else if ....
        enable
```
If an exception is raised by either of them you will raise an alert.
Note that call_api_functions does not mean immediately call one API function, it prepares the context for it, so you can call other functions too.

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -79,41 +79,42 @@ let join_internal ~__context ~self =
...
   | Some self ->    (* cluster_host and cluster exist *)
-    if not (Db.Cluster_host.get_joined ~__context ~self) then begin
-       match join_internal ~__context ~self with
-      | Result.Ok () ->
-        debug "Cluster_host.create was successful; cluster_host: %s" (Ref.string_of self)
-      | Result.Error error ->
-        warn "Error occurred during Cluster_host.create";
-        handle_error error
-    end;
+    if not (Db.Cluster_host.get_joined ~__context ~self)
```
edwintorok: This might look odd, but I think its better for ordering to keep it like this. The logical sequence is: join then later enable if not already enabled.

ocaml/xapi/xapi_cluster_host.ml
minishrink: After moving these assertions to `join_internal`, I don't think `create_internal` will need a lock, as it does nothing but create a ref, uuid, and db object

edwintorok:
The cleanup create_storage seems to be a no-op and not really needed for this PR, can we keep it for a future PR?

ocaml/xapi/xapi_cluster_host.ml
edwintorok: we don't want to accidentally create 2 DB objects though.

ocaml/xapi/xapi_cluster_host.ml
minishrink: We already test this in `test_clustering.test_assert_pif_prerequisites`, ensuring `pif_has_no_network_configuration` is raised if no IP exists.

edwintorok: We should have a unit test that tries to call Xapi_cluster_host.create and succeed with an empty IP.

ocaml/xapi/xapi_cluster_host.mli
```diff
@@ -19,31 +19,43 @@
...
 val sync_required : __context:Context.t -> host:API.ref_host ->
-  API.ref_Cluster option
+    API.ref_Cluster option
 (** [sync_required ~__context ~host] returns an option type indicating whether
     any action is required to sync the cluster. This will only be the case if
     the cluster object has [pool_auto_join] set and no corresponding
     Cluster_host object exists for the host. If one does not exist, the call
     will return the reference to the Cluster object wrapped in 'Some' *)
 
+val create_internal : __context:Context.t -> cluster:API.ref_Cluster ->
```
edwintorok: Why do you expose internal functions?

ocaml/xapi/xapi_cluster_host.mli
```diff
@@ -19,31 +19,43 @@
...
+    host:API.ref_host -> pIF:API.ref_PIF -> API.ref_Cluster_host
+(** [create_internal ~__context ~cluster ~host ~pif] creates cluster_host in DB *)
+
+val join_internal : __context:Context.t -> self:API.ref_Cluster_host -> unit
```
edwintorok: Same question.

ocaml/xapi/xapi_cluster_host.mli
```diff
@@ -72,11 +86,12 @@ val disable_clustering : __context:Context.t -> unit
 
 val resync_host : __context:Context.t -> host:API.ref_host -> unit
 (** [resync_host ~__context ~host] checks for the existence of a cluster_host.
-    If one is found and enabled in the db, it is enabled in the Client layer
-    too. Otherwise, nothing happens. *)
+    If one exists but hasn't joined the cluster, [join_internal] is called,
```
edwintorok: you can probably move the docs of join_internal inline here.

ocaml/xapi/xapi_cluster_host_helpers.ml
```diff
@@ -55,7 +55,15 @@ let get_operation_error ~__context ~self ~op =
...
 
+let assert_joined_cluster ~__context ~self = function
+  | `enable | `disable ->
+    (* Cannot enable nor disable without joining cluster *)
+    if not (Db.Cluster_host.get_joined ~__context ~self)
+    then raise Api_errors.(Server_error (cluster_host_not_joined, [ Ref.string_of self ]))
+  | _ -> ()
+
 let assert_operation_valid ~__context ~self ~op =
+  assert_joined_cluster ~__context ~self op;
```
edwintorok: This should be part of `get_operation_error`, because that is also used during update of allowed operations below.
I know currently the allowed operations are not updated properly, but lets avoid introducing more bugs :)

ocaml/xapi/xapi_cluster_helpers.ml
```diff
@@ -57,18 +57,13 @@ let get_operation_error ~__context ~self ~op =
...
-      raise Api_errors.(Server_error (not_supported_during_upgrade, []))
-    | _, `remove -> ()
-    | _, _ -> begin
-      match Xapi_clustering.find_cluster_host ~__context ~host:Helpers.(get_localhost ~__context) with
-      | Some self ->
-        if not (Db.Cluster_host.get_joined ~__context ~self)
-        then raise Api_errors.(Server_error (cluster_host_not_joined, [ Ref.string_of self ]))
-      | None -> () end
-  end;
+  assert_allowed_during_rpu __context op;
```
edwintorok: Same comment as below, now that you refactored it, its obvious that this is in the wrong place, please move it inside `get_operation_error`.

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -31,110 +31,112 @@ let fix_pif_prerequisites ~__context (pif_ref,pif_rec) =
...
 
-let create ~__context ~cluster ~host ~pif =
-  (* TODO: take network lock *)
-  with_clustering_lock (fun () ->
-      assert_operation_host_target_is_localhost ~__context ~host;
+      let host = Db.Cluster_host.get_host ~__context ~self in
       assert_cluster_host_can_be_created ~__context ~host;
-      assert_pif_attached_to ~host ~pif ~__context;
-      let ref = Ref.make () in
+      assert_pif_attached_to ~host ~pIF ~__context;
```
edwintorok: We need to query the pif again, otherwise the fix PIF prerequisites function before has done its job of setting the PIF to disallow unplug, but we still got the old PIF.
Would be better if we moved querying the PIF inside fix_pif_prerequisites, and just give fix_pif_prerequisites the PIF ref, not the PIF record (which can be out of date)

minishrink: In that case, should all the PIF helpers use the ref and not the record?

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -31,110 +31,112 @@ let fix_pif_prerequisites ~__context (pif_ref,pif_rec) =
...
+      let pIF = Db.Cluster_host.get_PIF ~__context ~self in
+      let pifrec = Db.PIF.get_record ~__context ~self:pIF in
+      fix_pif_prerequisites ~__context (pIF,pifrec);
 
-let create ~__context ~cluster ~host ~pif =
-  (* TODO: take network lock *)
-  with_clustering_lock (fun () ->
-      assert_operation_host_target_is_localhost ~__context ~host;
+      let host = Db.Cluster_host.get_host ~__context ~self in
       assert_cluster_host_can_be_created ~__context ~host;
```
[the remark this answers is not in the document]
minishrink: I'll move it to `create_internal`

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -31,110 +31,112 @@ let fix_pif_prerequisites ~__context (pif_ref,pif_rec) =
...
+        let p_ref = Db.Cluster_host.get_PIF ~__context ~self in
+        let p_rec = Db.PIF.get_record ~__context ~self:p_ref in
+          ip_of_pif (p_ref,p_rec)
         ) (Db.Cluster.get_cluster_hosts ~__context ~self:cluster) in
+
+      fix_pif_prerequisites ~__context (pIF,pifrec);
+
+      let host = Db.Cluster_host.get_host ~__context ~self in
+      assert_operation_host_target_is_localhost ~__context ~host;
+      assert_cluster_host_can_be_created ~__context ~host;
```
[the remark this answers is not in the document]
minishrink: No, but some are in the wrong place.

ocaml/xapi/xapi_cluster_host.ml
edwintorok: I think the code that modifies the PIF should be before the one that fetches the PIF record, so any code that modifies a PIF must not take the record as input, that way we can move the code that fetches the PIF record after all the code that modifies it. Currently the only code that modifies it is `fix_pif_prerequisites`

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -65,11 +67,9 @@ let join_internal ~__context ~self =
       let pIF = Db.Cluster_host.get_PIF ~__context ~self in
       let pifrec = Db.PIF.get_record ~__context ~self:pIF in
       fix_pif_prerequisites ~__context (pIF,pifrec);
-
-      let host = Db.Cluster_host.get_host ~__context ~self in
-      assert_cluster_host_can_be_created ~__context ~host;
-      assert_pif_attached_to ~host ~pIF ~__context;
-      assert_pif_prerequisites (pIF,pifrec);
+      (* Re-acquire PIF record after it has updated *)
```
minishrink: What kind of delay do we need here in order to refresh the record?

minishrink: Or should we just use `pif_of_host` here again?

ocaml/xapi/xapi_cluster_host.ml
minishrink: What should the mock RPC actually do? If it just returns a success, the code moves onto PIF prerequisites, at which point it raises `Api_errors.pif_has_no_network_configuration`, and does the same as the test mentioned above.

ocaml/xapi/xapi_cluster_host.ml
edwintorok: No delay is needed, just fetch it again.

edwintorok: Would be good if fix_pif_prerequisites fetched the record itself, to avoid accidentally reusing a stale record in the caller.

ocaml/xapi/xapi_cluster_host.ml
edwintorok: Lets create a test for `Xapi_cluster_host.create_as_necessary` then, which should succeed even if you have no network.

edwintorok: And then you can assign some IP to the PIF, and call resync_host, and check that it works in a second step in the unit test

ocaml/xapi/xapi_cluster_host.ml
minishrink: I've updated all the PIF helpers to fetch the record themselves now.

ocaml/xapi/create_storage.ml
edwintorok: Looks like we forgot to delete this line

ocaml/xapi/xapi.ml
edwintorok: typo in comment

ocaml/xapi/create_storage.ml
edwintorok: Looks like we forgot to delete this line

ocaml/xapi/xapi_pbd.ml
edwintorok: If you want to move these into a separate commit you can do so by doing a `git reset HEAD~1` and then `git add -p` and choose which changes to add to this commit, and then whats left can be in a next commit.

PR #3609, opened 2018-05-22
xapi-database, xapi-types: make safe-string compliant

mseri:
This is not enough to compile xapi with safe-strings, but it's a start


ocaml/xapi-types/ref.ml
```diff
@@ -79,6 +79,6 @@ let really_pretty_and_small x =
     Bytes.set r 0 c; Bytes.set r 1 ':';
     for i = 0 to 7 do Bytes.set r (i + 2) s.[prelen + i]; done;
     for i = 0 to 3 do Bytes.set r (i + 10)  s.[prelen + 8 + 1 + i]; done;
-    r
+    Bytes.unsafe_to_string r
```
gaborigloi: And these ones are safe because we no longer have access to the original `bytes`, so the string we return really is immutable.

PR #3610, opened 2018-05-24
make safe-string compliant

mseri:
Please review carefully. I have tried to minimise the interface changes and find a balance between the need of unsafe coercions and the new separation between strings and bytes.

Please refer to https://caml.inria.fr/pub/docs/manual-ocaml/libref/Bytes.html#3_Unsafeconversionsforadvancedusers for the limitations that we need to keep in mind if we use the unsafe conversions. As a rule of thumb, you should never mutate a string but you can convert it to bytes if you are passing it to a function that will not mutate it, and you should never use unsafe functions if the ownership is not clear of if the string is shared (e.g. in a fold). 

Make sure to read the Bytes section carefully before starting the review.


ocaml/database/block_device_io.ml
```diff
@@ -325,7 +325,7 @@ let transfer_data_from_sock_to_fd sock dest_fd available_space target_response_t
                (* Check that there's enough space *)
                if available_space - !total_length < len then raise NotEnoughSpace;
                (* Otherwise write it *)
-               Unixext.time_limited_write dest_fd len chunk target_response_time;
+               Unixext.time_limited_write dest_fd len (Bytes.unsafe_of_string chunk) target_response_time;
```
mseri: Before merging this PR, I plan to modify `Unixext.read_data_in_chunks` so that the `chunk` is bytes. This would make the signature closer to the `Unix` functions and reduce the amount of conversions that we need.

mseri:
This is now no-longer using any `Bytes.unsafe_of_string`, that are the most delicate as they convert immutable strings to mutable. The PR now depends on https://github.com/xapi-project/stdext/pull/31 and https://github.com/xapi-project/xs-opam/pull/276

A new version of xs-opam needs to be released before this can be merged!

.travis-xs-opam.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# replace the base remote with xs-opam
+opam repository remove default
+opam repository add xs-opam https://github.com/xapi-project/xs-opam.git
+
+# install the dependencies of xapi
+sudo apt install -y libpci-dev
+ls *.opam | xargs -n1 basename -s.opam | xargs -n1 -I '{}' opam pin add --no-action '{}' .
+opam depext --yes xapi
+opam install --deps-only xapi
```
gaborigloi: Should we replace the `.travis-opam-coverage.sh` script with this, and make it an error if it fails? That way we'd save some build time, because this is essentially doing the same things.

ocaml/database/block_device_io.ml
```diff
@@ -515,11 +516,11 @@ let action_writedelta block_dev_fd client datasock target_response_time =
   let generation_count = read_generation_count client in
   read_separator client;
   let length_str = read_length client in
-  let length = int_of_string length_str in
+  let length = int_of_string (Bytes.unsafe_to_string length_str) in
```
gaborigloi: This is just an idea: We could adopt a programming style where we override the previous let binding containing the `bytes` with a new one in these cases, so that we know for sure that the `bytes` is not accessible any more and therefore cannot be mutated:
```
   let length = read_length client in
   let length = int_of_string (Bytes.unsafe_to_string length) in
```

.travis-xs-opam.sh
mseri: I'll do that

ocaml/database/block_device_io.ml
gaborigloi: Since we only read 16 bytes here, would it be okay from a performance point of view to use `to_string`?

mseri: Yes, but it's overkill. This kind of use of `unsafe_to_string` is pervasive in some ocaml stdlib modules, it's safe to use here and it's not escaping anywhere (also, the danger is really with unsafe_of_string)

ocaml/graph/README.markdown
```diff
@@ -1,33 +0,0 @@
-# Describe object relationships via graphviz
```
gaborigloi: Nice!

ocaml/rfb/jbuild
```diff
@@ -1,21 +0,0 @@
-(* -*- tuareg -*- *)
```
gaborigloi: Nice! :D

ocaml/xapi/cli_progress_bar.ml
```diff
@@ -22,7 +22,7 @@ module Make(T: Floatable) = struct
     max_value: T.t;
     mutable current_value: T.t;
     width: int;
-    mutable line: string;
+    mutable line: bytes;
```
gaborigloi: I think we can remove `mutable` now?

mseri: no, we work on copies and then replace the line member

ocaml/xapi/cli_progress_bar.ml
```diff
@@ -62,18 +62,16 @@ module Make(T: Floatable) = struct
...
-      Bytes.set line (prefix + i) (if i = w - 1 then '>' else '#')
+      Bytes.set t.line (prefix + i) (if i = w - 1 then '>' else '#')
     done;
-    t.line <- Bytes.unsafe_to_string line;
     let percent = Printf.sprintf "%3d" (percent t) in
     String.blit percent 0 t.line (t.width - 19) 3;
     let eta = eta t in
     String.blit eta 0 t.line (t.width - 10) (String.length eta);
-    t.line
+    Bytes.to_string t.line
```
gaborigloi: All these overwrites of `line.t` are fine, because we always overwrite the same area, except for the progress bar which moves forward, so that's also fine

ocaml/xapi/cli_progress_bar.ml
gaborigloi: I thought we worked on `t.line` now because we've deleted the `<-` assignment.

mseri: I think you are right!

ocaml/database/block_device_io.ml
```diff
@@ -383,13 +383,13 @@ let send_failure client prefix error =
 
 (* Read a byte from the given file descriptor. If it is not '|' then raises ExpectedSeparator. *)
 let read_separator client =
-  let str = "\000" in
+  let str = Bytes.of_string "\000" in
   Unixext.really_read client str 0 1;
-  if str <> "|" then raise ExpectedSeparator
+  if str <> (Bytes.of_string "|") then raise ExpectedSeparator
```
edwintorok: Could this be written as `Bytes.get str 0 <> '|'?`

ocaml/database/block_device_io.ml
```diff
@@ -476,11 +476,12 @@ let action_writedb block_dev_fd client datasock target_response_time =
     (* If there's space, write some ASCII NULs over the next few bytes so that we trample on any data which may already exist on the block device *)
     let remaining_space = remaining_space - min_space_needed in
     let trample_size = if size_size > remaining_space then remaining_space else size_size in
-    Unixext.time_limited_write block_dev_fd trample_size (String.make trample_size '\000') target_response_time;
+    Unixext.time_limited_write block_dev_fd trample_size (Bytes.make trample_size '\000') target_response_time;
```
edwintorok: This could use `Unixext.time_limited_write_substring`

ocaml/database/block_device_io.ml
```diff
@@ -476,11 +476,12 @@ let action_writedb block_dev_fd client datasock target_response_time =
...
 
     (* Seek backwards in the block device to where the length is supposed to go and write it *)
     ignore_int (Unixext.seek_to block_dev_fd pos_to_write_length);
-    Unixext.time_limited_write block_dev_fd size_size (Printf.sprintf "%016d" total_length) target_response_time;
+    let total_length_str = Printf.sprintf "%016d" total_length in
+    Unixext.time_limited_write block_dev_fd size_size (Bytes.of_string total_length_str) target_response_time;
```
edwintorok: This could use Unixext.time_limited_write_substring

ocaml/tests/test_pool_license.ml
```diff
@@ -32,7 +32,7 @@ let string_of_date_opt = function
   | Some date -> Printf.sprintf "Some %s" (Date.to_string date)
 
 let f2d = Date.of_float
-let f2d2s = Date.to_string ++ Date.of_float
+let f2d2s f = f |> Date.of_float |> Date.to_string
```
edwintorok: Less deprecation warnings here I guess, nice.

ocaml/xapi/pool_db_backup.ml
```diff
@@ -34,7 +34,7 @@ let write_database (s: Unix.file_descr) ~__context =
     (* CA-18377: If we're in the middle of a rolling upgrade from Miami *)
     (* to Orlando, then only send a minimally-compliant Miami database. *)
     (* Orlando hosts will ignore this database and carry on.            *)
-    let len = String.length minimally_compliant_miami_database in
+    let len = Bytes.length minimally_compliant_miami_database in
     ignore (Unix.write s minimally_compliant_miami_database 0 len)
```
edwintorok: This could use `Unix.write_substring` and then you may not have to make `minimally_compliant_miami_database` mutable.

ocaml/xapi/rbac_audit.ml
```diff
@@ -222,7 +222,7 @@ let zip data = (* todo: remove i/o, make this more efficient *)
            (fun fd ->
               Gzip.compress fd (fun fd ->
                   let len = String.length data in
-                  let written = Unix.write fd data 0 len in
+                  let written = Unix.write fd (Bytes.of_string data) 0 len in
```
edwintorok: Another opportunity to use `Unix.write_substring`

ocaml/xapi/sparse_encoding.ml
```diff
@@ -67,15 +69,14 @@ module Marshal = struct
...
-
+    result
 end
 
 module Chunk = struct
   (** Represents an single block of data to write *)
   type t = {
     start: int64;
-    data: string;
+    data: bytes;
```
edwintorok: This was effectively mutable previously due to `unmarshal` writing to it, so should be fine.

ocaml/xapi/stream_vdi.ml
```diff
@@ -88,11 +88,11 @@ let made_progress __context progress n =
 let write_block ~__context filename buffer ofd len =
   let hdr = Tar_unix.Header.make filename (Int64.of_int len) in
   try
-    let csum = Sha1.to_hex (Sha1.string buffer) in
+    let csum = Sha1.to_hex (Sha1.string (Bytes.unsafe_to_string buffer)) in
     Tar_unix.write_block hdr (fun ofd -> Unix.write ofd buffer 0 len |> ignore) ofd;
     (* Write the checksum as a separate file *)
     let hdr' = Tar_unix.Header.make (filename ^ checksum_extension) (Int64.of_int (String.length csum)) in
-    Tar_unix.write_block hdr' (fun ofd -> ignore(Unix.write ofd csum 0 (String.length csum))) ofd
+    Tar_unix.write_block hdr' (fun ofd -> ignore(Unix.write ofd (Bytes.of_string csum) 0 (String.length csum))) ofd
```
edwintorok: `Unix.write_substring`

ocaml/xapi/stream_vdi.ml
```diff
@@ -177,8 +177,9 @@ let verify_inline_checksum ifd checksum_table =
   end;
   try
     let length' = Int64.to_int length in
-    let csum = String.make length' ' ' in
+    let csum = Bytes.make length' ' ' in
     Unixext.really_read ifd csum 0 length';
+    let csum = Bytes.unsafe_to_string csum in
```
edwintorok: This is safe because the old name is not accessible anymore, clever trick.

ocaml/xapi/xapi_session.ml
```diff
@@ -348,7 +349,9 @@ let slave_local_login ~__context ~psecret =
   Xapi_local_session.create ~__context ~pool:true
 
 (* Emergency mode login, uses local storage *)
-let slave_local_login_with_password ~__context ~uname ~pwd = wipe_params_after_fn [pwd] (fun () ->
+let slave_local_login_with_password ~__context ~uname ~pwd =
+  let pwd = Bytes.of_string pwd in
+  wipe_params_after_fn [pwd] (fun () ->
```
edwintorok: Due to the copy won't this leave the original `pwd` unwiped?

ocaml/xapi/xapi_session.ml
```diff
@@ -367,7 +370,9 @@ let slave_local_login_with_password ~__context ~uname ~pwd = wipe_params_after_f
       - try and authenticate remotely, passing the supplied username/password to the external auth/directory service. (Note: see below for definition of 'authenticate remotely')
    2. otherwise, Session.login_with_password will only attempt to authenticate against the local superuser credentials
 *)
-let login_with_password ~__context ~uname ~pwd ~version ~originator = wipe_params_after_fn [pwd] (fun () ->
+let login_with_password ~__context ~uname ~pwd ~version ~originator =
+  let pwd = Bytes.of_string pwd in
```
edwintorok: Same question as above

ocaml/xapi/xapi_session.ml
```diff
@@ -595,7 +600,10 @@ let login_with_password ~__context ~uname ~pwd ~version ~originator = wipe_param
       )
   )
 
-let change_password  ~__context ~old_pwd ~new_pwd = wipe_params_after_fn [old_pwd;new_pwd] (fun () ->
+let change_password  ~__context ~old_pwd ~new_pwd =
+  let old_pwd = Bytes.of_string old_pwd in
+  let new_pwd = Bytes.of_string new_pwd in
+  wipe_params_after_fn [old_pwd;new_pwd] (fun () ->
```
edwintorok: Same q as above

ocaml/database/block_device_io.ml
mseri: `write_substring` uses a "safe" unsafe conversion (ans in the ocaml Unix module), I prefer to avoid it unless necessary

ocaml/database/block_device_io.ml
mseri: Yes, I like this. Thanks!

ocaml/xe-cli/newcli.ml
```diff
@@ -349,11 +349,11 @@ let main_loop ifd ofd =
...
           let length = (Unix.stat x).Unix.st_size in
           marshal ofd (Blob (Chunk (Int32.of_int length)));
-          let buffer = String.make (1024 * 1024 * 10) '\000' in
+          let buffer = Bytes.make (1024 * 1024 * 10) '\000' in
           let left = ref length in
           while !left > 0 do
-            let n = Unix.read fd buffer 0 (min (String.length buffer) !left) in
-            Xapi_stdext_unix.Unixext.really_write ofd buffer 0 n;
+            let n = Unix.read fd buffer 0 (min (Bytes.length buffer) !left) in
+            Xapi_stdext_unix.Unixext.really_write ofd (Bytes.unsafe_to_string buffer) 0 n;
```
edwintorok: is there a `really_write_substring`

ocaml/database/block_device_io.ml
mseri: Here it makes sense to use the substring one to avoid a copy, I'll change it

ocaml/xapi/pool_db_backup.ml
mseri: Done!

ocaml/xapi/rbac_audit.ml
mseri: Done!

ocaml/xapi/stream_vdi.ml
mseri: Thanks

ocaml/xapi/xapi_session.ml
mseri: Until it is GCd away. We cannot really mutate strings or make them mutable with the unsafe conversion, this will have problems with compiler optimisations. The alternative is an invasive change to all the interfaces to accept bytes

ocaml/xe-cli/newcli.ml
mseri: `really_write_string`, I had forgotten about it

ocaml/xapi/xapi_session.ml
mseri: I have made some tests with the core dumps. It looks like even in normal use, it is GCd fast enough as I could not find it in the core dump at all. I will create a ticket to investigate what to do

ocaml/xe-cli/newcli.ml
edwintorok: lets keep it as it is for, we can update unixext if we see patterns like this a lot in the future.

ocaml/xapi/xapi_session.ml
edwintorok: Thanks, there is a ticket to investigate this, so we should probably not do more in this PR.

PR #3611, opened 2018-05-25
CA-290471: fix wrong token timeout in XenCenter

edwintorok:
Use same units for object default value and parameter default value, see commit message for explanation.
Tested by installing latest feature branch build, replacing xapi, and it has correctly updated the field in the database. The next time we build an SDK and rebuild XenCenter then XenCenter should get the correct value as well.

### DIFF
--- a/ocaml/idl/datamodel_cluster.ml
+++ b/ocaml/idl/datamodel_cluster.ml
@@ -17,6 +17,11 @@ let cluster_operation =
 
 let lifecycle = [Prototyped, rel_kolkata, ""]
 
+let lifecycle_timeout = [
+            Prototyped, rel_kolkata, "the unit is milliseconds";
+            Changed, rel_lima, "the unit is now seconds"
+          ]
+
 let timeout_params =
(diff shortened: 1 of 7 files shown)


ocaml/idl/datamodel_cluster.ml
```diff
@@ -171,13 +171,19 @@ let t =
...
+            Prototyped, rel_kolkata, "the unit is milliseconds";
+            Changed, rel_lima, "the unit is now seconds"
+          ]
+          ~ty:Float "token_timeout" ~default_value:(Some (VFloat Constants.default_token_timeout_s))
+          "The corosync token timeout in seconds"
+
+      ; field   ~qualifier:StaticRO ~lifecycle:[
+            Prototyped, rel_kolkata, "the unit is milliseconds";
+            Changed, rel_lima, "the unit is now seconds"
+          ]
```
minishrink: You could create a value for this specific lifecycle instead of passing it twice

PR #3613, opened 2018-05-30
CA-289623: Added missing branding; removed duplicate comment.

kc284:


kc284:
We may want to consider cherry picking this into `1.90-lcm`?

PR #3616, opened 2018-06-01
CA-289735: do not disable clustering daemon on shutdown from xapi-domains

edwintorok:
Just unplug the PBDs, and allow the clustering daemon to shut itself
down using the `xapi-clusterd-shutdown.service` unit.

This is to avoid a circular dependency between `xapi-domains`, `dlm`,
and `remote-fs` that prevents `dlm` from getting stopped from inside
`xapi-domains` on shutdown.

This needs to be merged together with matching PR on the clustering daemon.


edwintorok:
This can be merged together with the xenserver-specs and clustering daemon change. Wait until they're all approved and then merge.

PR #3618, opened 2018-06-05
 CA-290006 update snapshot device model profile

lindig:
A snapshot stores all relevant VM meta data inside VM.snapshot_metadata
which contains a platform value, which in turn contains a device model
profile. This needs to be updated when xapi's database is updated but so
far was not. Confusingly, the VM.platform field of a snapshot is not
used when restoring a snapshot. This commit updates the device model
profile inside the snapshot metadata.
    
Snapshot meta data is encoded as an s-expression. Updating it thus
requires decoding, updating, and encoding it again.

This fix also pulls in CA-290874 from the master branch to make sure the domain
type is handled correctly in snapshots.

### DIFF
--- a/ocaml/xapi/xapi_vm_snapshot.ml
+++ b/ocaml/xapi/xapi_vm_snapshot.ml
@@ -450,6 +450,14 @@ let extended_do_not_copy = [
   "children";
 ] @ do_not_copy
 
+(* Update the domain_type field if it does not match the HVM_boot_policy *)
+let ensure_domain_type_is_consistent ~__context ~vm =
+  Db.VM.get_HVM_boot_policy ~__context ~self:vm
+  |> fun p -> Xapi_vm_helpers.derive_domain_type ~hVM_boot_policy:p
+  |> fun value ->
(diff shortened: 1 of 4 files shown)


ocaml/xapi/xapi_vm_snapshot.ml
```diff
@@ -250,8 +250,8 @@ let copy_vm_fields ~__context ~metadata ~dst ~do_not_copy ~default_values =
   List.iter
     (fun (key,value) ->
        let value =
-         if List.mem_assoc key default_values
-         then List.assoc key default_values
+         if List.mem_assoc key overrides
+         then List.assoc key overrides
          else value in
```
minishrink: This could be simplified to `let value = try List.assoc key overrides with Not_found -> value`

lindig: This is code from the commits that I cherry-picked from the master branch. To keep rebasing simple,  I'd rather not change it. I agree that List.mem/List.assoc is an anti pattern.

jonludlam:
Looks good to me!

PR #3619, opened 2018-06-06
CA-287838: Wait for slave host to be enabled timed out: slave's dom0 …

krizex:
…metadata incomplete

After the fix of CA-280981, we import the dom0 metadata of slave to master
rather than re-create a new VM record for it when slave join a pool. However it
will make the VM record of dom0 of slave in an incorrect state.

This fix aims to fix the VM record state of dom0 of slave when slave restarting
xapi.


lindig:
The code looks fine but the commit message can be improved:

* Shorten the subject line to not flow into the body
* I am not familiar with "dom0 meta data". I assume it is the dom0 VM record.
* Grammar:  "After the fix of CA-280981, we import the dom0 VM metadata of the slave into the master
rather than re-creating a new VM record for it when the slave joins the pool. However, it
will leave the dom0 VM record of the of slave in an incorrect state."
* The commit message does not tell how the fix works - I don't see a connection between what the commit does and the description of the problem above.

PR #3621, opened 2018-06-07
CA-291136: wait for carrier on management, and do not fail on missing IP during parallel join

edwintorok:


edwintorok:
Added some more bugfixes for race conditions here, needs more testing.

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -72,11 +72,19 @@ let join_internal ~__context ~self =
...
+          let p_ref = Db.Cluster_host.get_PIF ~__context ~self in
+          let p_rec = Db.PIF.get_record ~__context ~self:p_ref in
+          (* parallel join: some hosts may not have an IP yet *)
+          try
+            let other_ip = ip_of_pif (p_ref,p_rec) in
+            if other_ip <> ip then Some other_ip
+            else None
+          with _ -> None
         ) (Db.Cluster.get_cluster_hosts ~__context ~self:cluster) in
+      if ip_list = [] then
```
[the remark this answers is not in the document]
edwintorok: I didn't want to reindent everything following it, this is `if` is not nested because it raises an exception immediately.

edwintorok:
@robhoes could you take a look at the changes in 'helpers.ml' and 'xapi_mgmt_iface.ml', they change some common code to make it reusable for clustering and avoid some race conditions.

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -160,31 +160,33 @@ let rebind ~__context =
...
+let wait_for_ip get_ip is_connected =
+  let rec loop () =
+    let ip = get_ip () in
+    let connected = is_connected () in
+    if ip = "" || not connected then begin
+      debug "wait_for_ip (IP=%s, connected:%b), waiting" ip connected;
+      Condition.wait ip_cond ip_mutex;
+      loop ()
+    end
+    else ip
```
minishrink: Pedantic nitpick: could have `end else ip` on one line, probably don't need `end` on its own

ocaml/xapi/helpers.ml
```diff
@@ -63,10 +63,16 @@ let get_pool ~__context = List.hd (Db.Pool.get_all ~__context)
...
   Db.Pool.get_master ~__context ~self:(get_pool ~__context)
 
-let get_management_iface_is_connected ~__context =
+let get_iface_is_connected ~__context iface =
   let dbg = Context.string_of_task __context in
-  Net.Bridge.get_physical_interfaces dbg (Xapi_inventory.lookup Xapi_inventory._management_interface)
-  |> List.exists (fun name -> Net.Interface.is_connected dbg name)
+  Net.Bridge.get_physical_interfaces dbg iface
+  |> List.exists (fun name ->
+    debug "Checking interface %s" name;
```
robhoes: Let's drop this debug line. It doesn't add much and we are already logging too much.

ocaml/xapi/helpers.ml
```diff
@@ -63,10 +63,16 @@ let get_pool ~__context = List.hd (Db.Pool.get_all ~__context)
 let get_master ~__context =
   Db.Pool.get_master ~__context ~self:(get_pool ~__context)
 
-let get_management_iface_is_connected ~__context =
+let get_iface_is_connected ~__context iface =
```
robhoes: Note that this is checking whether at least one of the physical interface on the given bridge is connected, so it's a little confusing that this function talks about interfaces. Please add a comment above the function, and replace `iface` by `bridge`.

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -178,20 +178,15 @@ let wait_for_management_ip ~__context =
...
   let pIF = Db.Cluster_host.get_PIF ~__context ~self in
-  let iP = ref (Db.PIF.get_IP ~__context ~self:pIF) in
-  Mutex.execute ip_mutex (* Don't return until PIF is plugged AND has a valid IP *)
-    (fun () -> while !iP="" || not (has_carrier ~__context ~self:pIF) do
-        Condition.wait ip_cond ip_mutex;
-        iP := Db.PIF.get_IP ~__context ~self:pIF
-      done);
-  !iP
+  let network = Db.PIF.get_network ~__context ~self:pIF in
+  let interface = Db.Network.get_bridge ~__context ~self:network in
```
robhoes: `bridge` again.

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -178,20 +178,15 @@ let wait_for_management_ip ~__context =
...
+  debug "Waiting for clustering IP on interface %s (%s)" interface (Ref.string_of pIF);
+  let get_ip () = Db.PIF.get_IP ~__context ~self:pIF in
```
robhoes: How about going a step further and doing a direct lookup of the IP as well? In the case of DHCP, the database will lag behind a little as well.

edwintorok: There is a Helpers.get_primary_ip_addr, I'll use that.

ocaml/xapi/helpers.ml
edwintorok: I think I was slightly confused here because I didn't expect a single physical interface to always have a bridge, though in that case it would be a bridge with 1 interface, so should be fine?

robhoes: The way networking is set up is that each physical interface is always on its own bridge, unless it is bonded and there are multiple physical interface on a single bridge. And the IP addresses of dom0 are always on bridges.

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -160,31 +160,33 @@ let rebind ~__context =
...
-          Condition.wait ip_cond ip_mutex;
-          ip := (match Helpers.get_management_ip_addr ~__context with Some x -> x | None -> "");
-          is_connected := (Helpers.get_management_iface_is_connected ~__context)
-        done);
-  !ip
+let wait_for_ip get_ip is_connected =
+  let rec loop () =
+    let ip = match get_ip () with Some x -> x | None -> "" in
+    let connected = is_connected () in
+    if ip = "" || not connected then begin
```
edwintorok: I could've just compared against None here, but we need the string for the debug line below, and in case something goes wrong and we get `Some ""` for the IP, we catch that here (although by looking at the code I don't think there is a way that could happen).

edwintorok:
Pushed some fixups, if they are good I'll squash the commits.

robhoes:
Looks good to me. Please go ahead and squash+merge.

PR #3626, opened 2018-06-11
CA-290450: quicktest_vdi_ops_data_integrity: test large VDIs too

gaborigloi:

### DIFF
--- a/ocaml/quicktest/quicktest_vdi_ops_data_integrity.ml
+++ b/ocaml/quicktest/quicktest_vdi_ops_data_integrity.ml
@@ -110,10 +110,10 @@ let checksum rpc session_id vdi =
       Digest.to_hex (Digest.file path)
     )
 
-let check_vdi_unchanged session_id ~prepare_vdi ~vdi_op sr_info =
+let check_vdi_unchanged session_id ~vdi_size ~prepare_vdi ~vdi_op sr_info =
   let rpc = !Quicktest_common.rpc in
   let sR = sr_info.Storage_test.sr in
-  Storage_test.VDI.with_new session_id sR (fun vdi ->
+  Storage_test.VDI.with_new ~virtual_size:vdi_size session_id sR (fun vdi ->
       prepare_vdi vdi;
       let checksum_original = checksum rpc session_id vdi in
       let new_vdi = vdi_op rpc session_id sR vdi in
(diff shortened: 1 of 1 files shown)


ocaml/quicktest/storage_test.ml
```diff
@@ -206,6 +206,10 @@ module Sr_filter = struct
   let not_type _type rpc session_id =
     List.filter (fun i -> Client.Client.SR.get_type ~rpc ~session_id ~self:i.sr <> _type)
 
+  let thin_pro rpc session_id =
+    List.filter (fun i -> match Client.Client.SR.get_type ~rpc ~session_id ~self:i.sr with
+        | "gfs2" | "lvm" -> true | _ -> false)
```
edwintorok: I'm confused, LVM is not thinly provisioned, if it was we wouldn't need GFS2.
What is the actual property that you need in your test? Is it enough free space to accomodate the 2GB VDI?

ocaml/quicktest/quicktest_vdi_ops_data_integrity.ml
```diff
@@ -144,16 +144,23 @@ let export_import_vdi rpc session_id ~exportformat sR vdi =
...
-  ; op_name^": random VDI", `Slow, check_vdi_unchanged ~prepare_vdi:(write_random_data session_id) ~vdi_op, filter
-  ; op_name^": full VDI", `Slow, check_vdi_unchanged ~prepare_vdi:(fill session_id) ~vdi_op, filter
+  [ op_name^": small empty VDI", `Slow, check_vdi_unchanged ~vdi_size:Sizes.(4L ** mib) ~prepare_vdi:(fun _vdi -> ()) ~vdi_op, f
+  ; op_name^": small random VDI", `Slow, check_vdi_unchanged ~vdi_size:Sizes.(4L ** mib) ~prepare_vdi:(write_random_data session_id) ~vdi_op, f
+  ; op_name^": small full VDI", `Slow, check_vdi_unchanged ~vdi_size:Sizes.(4L ** mib) ~prepare_vdi:(fill session_id) ~vdi_op, f
+  ]
+
+let large_data_integrity_tests session_id vdi_op op_name =
+  [ op_name^": ~2GiB empty VDI", `Slow, check_vdi_unchanged ~vdi_size:Sizes.(2L**gib) ~prepare_vdi:(fun _vdi -> ()) ~vdi_op, F.(f ||> thin_pro)
+  ; op_name^": ~2GiB random VDI", `Slow, check_vdi_unchanged ~vdi_size:Sizes.(2L**gib +* 16L**mib) ~prepare_vdi:(write_random_data session_id) ~vdi_op, F.(f ||> thin_pro)
```
edwintorok: Would it make sense to use a random size here >2GiB? i.e. 2gib + random between 0 and 16mib?
Might catch a few more interesting cases where the last chunk is a partial one.

ocaml/quicktest/storage_test.ml
gaborigloi: Good point, *I* was confused. What I want for this test is a sparse VDI, so that we can quickly copy only the allocated blocks. This is true for lvm, because, even though we allocate all the space, there is a VHD inside it. (And the filer can still make it thin-pro, if it's intelligent, it can lie and say the space is allocated even though it isn't) In XS we currently have 2 types of VDIs, raw and sparse, so what we want for this test is a sparse VDI, not a raw one. I think all the SRs support sparse VDIs (maybe except for the ISO SR).

ocaml/quicktest/quicktest_vdi_ops_data_integrity.ml
gaborigloi: Good idea, I've implemented that!

PR #3627, opened 2018-06-12
[WIP] CP-28477

minishrink:
This PR aims to make Cluster.pool_force_destroy succeed no matter what.

Still to do:
- add API timeout to `Cluster.pool_force_destroy`
- make more robust


edwintorok:
Note that the API timeout should already be available since https://github.com/xapi-project/xen-api-libs-transitional/pull/41 and https://github.com/xapi-project/xen-api/pull/3548, so it should "just work", we should just try and see what exception we get.

ocaml/xapi/xapi_cluster.ml
```diff
@@ -75,23 +75,24 @@ let create ~__context ~pIF ~cluster_stack ~pool_auto_join ~token_timeout ~token_
...
 
+
 let destroy ~__context ~self =
   let cluster_hosts = Db.Cluster.get_cluster_hosts ~__context ~self in
-  let cluster_host = match cluster_hosts with
-    | [] -> None
-    | [ cluster_host ] -> Some (cluster_host)
+  match cluster_hosts with
+    | [] ->
+      info "No cluster_hosts found, cannot run Cluster.destroy without force"
```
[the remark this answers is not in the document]
edwintorok: And in Cluster.pool_force_destroy we should try to call Xapi_cluster_host.forget on all the remaining cluster hosts, after we've tried destroy and force destroy on them and they failed.

edwintorok:
Another thing we should improve is Xapi_cluster_host.forget, which does not delete the cluster host. If you use Host.destroy thats fine, the GC will delete it, but with pool_force_destroy we won't have that, so its better if Xapi_cluster_host does the right thing.

ocaml/xapi/xapi_cluster.ml
```diff
@@ -116,84 +121,62 @@ let pool_create ~__context ~network ~cluster_stack ~token_timeout ~token_timeout
...
-let pool_force_destroy ~__context ~self =
-  (* For now we assume we have only one pool, and that the cluster is the same as the pool.
-     This means that the pool master must be a member of this cluster. *)
+let pool_destroy_common ~__context ~self ~force =
+  (* Prevent new hosts from joining if destroy fails *)
+  Db.Cluster.set_pool_auto_join ~__context ~self ~value:false;
   let master = Helpers.get_master ~__context in
   let master_cluster_host =
     Xapi_clustering.find_cluster_host ~__context ~host:master
+    |> Xapi_stdext_monadic.Opt.unbox
```
edwintorok: This would fail if there is no master

ocaml/xapi/xapi_cluster.ml
```diff
@@ -116,84 +121,62 @@ let pool_create ~__context ~network ~cluster_stack ~token_timeout ~token_timeout
...
-    debug "Cluster_host.force_destroy was successful"
-
-(* Helper function; concurrency checks are done in implementation of Cluster.destroy and Cluster_host.destroy *)
+  debug "Ignoring exceptions while trying to force destroy cluster hosts.";
+    foreach_cluster_host ~__context ~self ~log:false
+      ~fn:Client.Client.Cluster_host.force_destroy
+      all_remaining_cluster_hosts;
+
+  (* Finally, any cluster_hosts we couldn't force destroy will be forgotten *)
+  begin match Db.Cluster_host.get_all ~__context with
```
edwintorok: You probably don't need the match here, just call foreach with an empty list, it should be a no-op.

edwintorok:
Could you make this change too? 
> Another thing we should improve is Xapi_cluster_host.forget, which does not delete the cluster host. If you use Host.destroy thats fine, the GC will delete it, but with pool_force_destroy we won't have that, so its better if Xapi_cluster_host does the right thing.

ocaml/xapi/xapi_cluster.ml
```diff
@@ -108,103 +112,82 @@ let pool_create ~__context ~network ~cluster_stack ~token_timeout ~token_timeout
...
   cluster
 
-(* Helper function; if opn is None return all, else return those not equal to it *)
-let filter_on_option opn xs =
-  match opn with
-  | None -> xs
-  | Some x -> List.filter ((<>) x) xs
+let foreach_cluster_host ~__context ~self ~(fn : rpc:(Rpc.call -> Rpc.response) ->
+      session_id:API.ref_session -> self:API.ref_Cluster_host -> unit) ~log =
+  let wrapper = if log then log_and_ignore_exn else (fun _ -> ()) in
```
edwintorok: The bug is here: if log is false then this does not run `fn` at all, the correct way would be:
`let wrapper = if log then log_and_ignore_exn else (fun f -> f ()) in`

minishrink: time to hang my head in shame
or borrow the cone of shame from my dog

PR #3597, opened 2018-05-17
CP-28117: restart VMs in parallel when recovering from a host failure in HA

edwintorok:
I have dev-tested this using VMs created by this script, and then turning the power off on an HA slave, and watching VMs get restarted in the correct order (first all with order=0 got started in parallel, then order=1, then order=2).
```
#!/bin/bash
set -e
MEMORY='16MiB'
N=50
ORDER=2

if [ ! -f /boot/guest/test-vm.xen.gz ]; then
        wget https://github.com/xapi-project/xen-test-vm/releases/download/0.0.5/test-vm.xen.gz
        sha256sum test-vm.xen.gz
        mkdir -p /boot/guest
        mv test-vm.xen.gz /boot/guest
fi


. /etc/xensource-inventory
host_uuid="$INSTALLATION_UUID"

SR=$(xe pool-list params=default-SR --minimal)

for order in $(seq 0 "$ORDER"); do
    for i in $(seq 1 "$N"); do
        vm_uuid=$(xe vm-create name-label="mirage-$order-$i-$MEMORY")
        xe vm-param-set PV-kernel=/boot/guest/test-vm.xen.gz uuid="$vm_uuid"
        xe vm-memory-limits-set uuid="$vm_uuid" static-min="$MEMORY" dynamic-min="$MEMORY" dynamic-max="$MEMORY" static-max="$MEMORY"
        xe vm-param-set uuid="$vm_uuid" ha-restart-priority=restart
        xe vm-param-set uuid="$vm_uuid" order="$order"
        for i in $(seq 1 3); do
            vdi_uuid=$( xe vdi-create sr-uuid="$SR" name-label="mirage-$order-$i-$MEMORY-disk" virtual-size=1MiB)
            xe vbd-create vdi-uuid=$vdi_uuid vm-uuid=$vm_uuid device=autodetect >/dev/null
        done
    done
done
```

I'd love to have some unit tests for the VM restart code, but I couldn't find any. Should new unit tests be written for the HA VM restart code, or just the new helper module?

PR #3628, opened 2018-06-12
CP-28117: restart VMs in parallel when recovering from a host failure in HA

jonludlam:
This is PR #3597 rebased and with minor fixups. The following is from @edwintorok directly copied from the previous PR for reference:

I have dev-tested this using VMs created by this script, and then turning the power off on an HA slave, and watching VMs get restarted in the correct order (first all with order=0 got started in parallel, then order=1, then order=2).
```
#!/bin/bash
set -e
MEMORY='16MiB'
N=50
ORDER=2

if [ ! -f /boot/guest/test-vm.xen.gz ]; then
        wget https://github.com/xapi-project/xen-test-vm/releases/download/0.0.5/test-vm.xen.gz
        sha256sum test-vm.xen.gz
        mkdir -p /boot/guest
        mv test-vm.xen.gz /boot/guest
fi


. /etc/xensource-inventory
host_uuid="$INSTALLATION_UUID"

SR=$(xe pool-list params=default-SR --minimal)

for order in $(seq 0 "$ORDER"); do
    for i in $(seq 1 "$N"); do
        vm_uuid=$(xe vm-create name-label="mirage-$order-$i-$MEMORY")
        xe vm-param-set PV-kernel=/boot/guest/test-vm.xen.gz uuid="$vm_uuid"
        xe vm-memory-limits-set uuid="$vm_uuid" static-min="$MEMORY" dynamic-min="$MEMORY" dynamic-max="$MEMORY" static-max="$MEMORY"
        xe vm-param-set uuid="$vm_uuid" ha-restart-priority=restart
        xe vm-param-set uuid="$vm_uuid" order="$order"
        for i in $(seq 1 3); do
            vdi_uuid=$( xe vdi-create sr-uuid="$SR" name-label="mirage-$order-$i-$MEMORY-disk" virtual-size=1MiB)
            xe vbd-create vdi-uuid=$vdi_uuid vm-uuid=$vm_uuid device=autodetect >/dev/null
        done
    done
done
```

I'd love to have some unit tests for the VM restart code, but I couldn't find any. Should new unit tests be written for the HA VM restart code, or just the new helper module?

### DIFF
--- a/ocaml/xapi/xapi_ha_vm_failover.ml
+++ b/ocaml/xapi/xapi_ha_vm_failover.ml
@@ -705,6 +705,7 @@ let restart_auto_run_vms ~__context live_set n =
       Xapi_alert.add ~msg:Api_messages.ha_protected_vm_restart_failed ~cls:`VM ~obj_uuid ~body:""
     end in
 
+  let open TaskChains.Infix in
   (* execute the plan *)
   Helpers.call_api_functions ~__context
     (fun rpc session_id ->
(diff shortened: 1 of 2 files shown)


jonludlam:
I'd like to add a unit test or two to this PR before merging, but please continue to review the logic while I'm making the tests.

jonludlam:
Actually it's a giant PITA to test this stuff unit-fashion, so I'm not going to sink any more time into it.

ocaml/xapi/xapi_ha_vm_failover.ml
```diff
@@ -580,52 +731,80 @@ let restart_auto_run_vms ~__context live_set n =
...
+           of objects. 
+
+           The return is a list of 'a Result.results
+       *)
+       let map_parallel ~order_f f lst =
+         lst
+         |> Helpers.group_by ~ordering:`ascending order_f
+         |> ordered_map_concat (fun same_order ->
+             same_order
+             |> List.rev_map fst
```
lindig: To make it slightly more efficient, `map` and `fst` could be combined into `List.map (fun xs -> f (fst xs))`

ocaml/xapi/xapi_ha_vm_failover.ml
```diff
@@ -15,17 +15,167 @@
...
+    -> session_id:API.ref_session -> ('a * ('b, exn) Result.result t) list
+    -> ('a * ('b, exn) Result.result) list
+  (** [parallel ~__context lst] execute actions in [lst] in parallel.
+   * Returns the results from [lst], in an unspecified order, hence each item in [lst] is a tuple
+   * with arbitrary data (e.g. a VM ref) as first element, and the action as second:
+   * this way the results will be associated with some meaningful data.
+   *)
+end = struct
+  (** structured like a Freer Monad *)
+  type +'a t =
```
lindig: The variance annotation is only important if we expect `'a` to have subtypes - for example, being a list of polymorphic variants. Otherwise it would be enough to make it invariant  (no `+` or `-`).

robhoes:
@jonludlam https://s3-media4.fl.yelpcdn.com/bphoto/NKP9CaNZIXD7_qkRk9Onmg/o.jpg

PR #3629, opened 2018-06-12
nbd_client_manager: increase number of /dev/nbds and wait for a free one

gaborigloi:
We wait up to 5 minutes when all the 128 NBD devices are in use.

### DIFF
--- a/scripts/nbd_client_manager.py
+++ b/scripts/nbd_client_manager.py
@@ -68,28 +68,28 @@ def _call(cmd_args, error=True):
     If [error] and exit code != 0, log and throws a CalledProcessError.
     """
     LOGGER.debug("Running cmd %s", cmd_args)
-    p = subprocess.Popen(
+    proc = subprocess.Popen(
         cmd_args,
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
         close_fds=True
     )
 
-    stdout, stderr = p.communicate()
+    stdout, stderr = proc.communicate()
 
-    if error and p.returncode != 0:
+    if error and proc.returncode != 0:
         LOGGER.error(
             "%s exitted with code %d: %s",
             ' '.join(cmd_args),
-            p.returncode,
+            proc.returncode,
             stderr)
 
         raise subprocess.CalledProcessError(
-            returncode=p.returncode,
+            returncode=proc.returncode,
             cmd=cmd_args,
             output=stderr)
 
-    return p.returncode
+    return proc.returncode
 
 
 def _is_nbd_device_connected(nbd_device):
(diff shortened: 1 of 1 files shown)


gaborigloi:
I've tested this with a new quicktest that attaches 26 VBDs to dom0, and it passed.

scripts/nbd_client_manager.py
```diff
@@ -114,8 +114,9 @@ def _find_unused_nbd_device():
     """
     Returns the path of the first /dev/nbdX device that is not
     connected according to nbd-client.
+    Raises NbdDeviceNotFound if no devices are available.
     """
-    for device_no in range(0, 100):
+    for device_no in range(0, 1000):
```
edwintorok: you use 1000 here and 128 below, should this be 128 too?

gaborigloi: No, this is deliberate, to keep the old behaviour: when we go past 128, we'll get an exception due to nonexistent nbd device that we catch in the loop. This is to just keep the code simple.

gaborigloi:
I'll do more testing cause modprobe ran out of memory

gaborigloi:
I've modified the PR after discussion with @stefanopanella , and now the BVT passed, so this PR is ready for review :)

scripts/nbd_client_manager.py
```diff
@@ -155,14 +156,28 @@ def _remove_persistent_connect_info(device):
...
     """Connects to a free NBD device using nbd-client and returns its path"""
-    _call(['modprobe', 'nbd'])
-    with FILE_LOCK:
-        nbd_device = _find_unused_nbd_device()
-        cmd = ['nbd-client', '-unix', path, nbd_device, '-name', exportname]
-        _call(cmd)
-        _wait_for_nbd_device(nbd_device=nbd_device, connected=True)
-        _persist_connect_info(nbd_device, path, exportname)
-    return nbd_device
+    _call(['modprobe', 'nbd', 'nbds_max=24'])
```
edwintorok: Why 24 now, did you mean 1024?

gaborigloi: No, we've discussed with Stefano, we had to reduce the number due to the memory usage. We're not sure how much they use.

edwintorok: Could you add a comment about that? Also do you know how much more code has to be converted over to use NBD directly instead of the block device? 
VDI export should now use NBD directly, how about the emulation (qemu-trad/qemu-upstream), does it connect straight to NBD as well?

gaborigloi: Good idea, I've added a comment.
Currently nothing uses NBD directly, as far as I know. There are many places in the code that we need to convert, but we haven't started it yet.

edwintorok: Not even vhd-tool?

gaborigloi: No, we have a ticket for it though :disappointed:

PR #3631, opened 2018-06-13
CA-290526: Update location for cluster stack supported SRs file

thomasmck:

### DIFF
--- a/ocaml/xapi/xapi_globs.ml
+++ b/ocaml/xapi/xapi_globs.ml
@@ -715,7 +715,6 @@ let sm_dir = ref "/opt/xensource/sm"
 let web_dir = ref "/opt/xensource/www"
 
 let cluster_stack_root = ref "/usr/libexec/xapi/cluster-stack"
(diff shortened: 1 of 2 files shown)


thomasmck:
File with list of supported SRs introduced here: https://code.citrite.net/projects/XS/repos/xapi-storage-plugins/pull-requests/170/overview

ocaml/xapi/xapi_globs.ml
```diff
@@ -715,6 +715,7 @@ let sm_dir = ref "/opt/xensource/sm"
 let web_dir = ref "/opt/xensource/www"
 
 let cluster_stack_root = ref "/usr/libexec/xapi/cluster-stack"
+let cluster_stack_compatible_srs = ref "ha_statefile_srs"
```
edwintorok: Unless we intend to make this configurable you can just put this inside `xha_scripts.ml` like the other script names are as a regular string, not a ref.

thomasmck: Sure, will update now (will also need to update the other commit)

PR #3632, opened 2018-06-13
CA-290466: Add quicktest to test parallel VDI dom0 attach limit

gaborigloi:

### DIFF
--- a/ocaml/quicktest/quicktest_vdi.ml
+++ b/ocaml/quicktest/quicktest_vdi.ml
@@ -290,6 +290,13 @@ let vdi_general_test session_id sr_info =
         (fun () -> Client.Client.VBD.destroy !rpc session_id vbd)
     )
 
+let multiple_dom0_attach session_id sr_info =
+  let rec loop vdi = function
+    | 0 -> ()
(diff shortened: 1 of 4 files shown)


ocaml/quicktest/quicktest_vdi.ml
```diff
@@ -290,6 +290,13 @@ let vdi_general_test session_id sr_info =
         (fun () -> Client.Client.VBD.destroy !rpc session_id vbd)
     )
 
+let multiple_dom0_attach session_id sr_info =
```
lindig: I think I would factor out the iteration. If you don't want to use a `for` loop, I would do something like this:

```
  let rec iter n f = match n with
    | 0 -> ()
    | n -> f (); iter (n-1) f in

  iter 20 (fun () -> 
    Storage_test.VDI.with_any session_id sr_info
      (fun vdi -> 
        Storage_test.VDI.with_attached session_id vdi (fun _ -> ()))
```

gaborigloi: We need more than iteration - we need to nest the functions as well here, to have all the 20 VDIs attached in parallel.

lindig: Ok. Feel free to keep it as is. Maybe a comment what this code does is warranted because I seem to have misread it.

PR #3633, opened 2018-06-14
[smapiv3 merge] Implement & move to VDI.attach2 SMAPIv2 call, deprecate VDI.attach

gaborigloi:
We cannot change the original VDI.attach call, because during SXM, older
xapis call Remote.VDI.attach, therefore changing it's return type would
break backward compatibility and SXM. Therefore I kept the old
VDI.attach for backward compatibility, it just calls attach2 at the
Storage_impl layer and transforms its return type. Although that is not
strictly necessary, because in older xapis we ignore the return type of the
Remote.VDI.attach call.

Now only older xapis will be calling the deprecated VDI.attach during
SXM, newer versions starting from this one will use the new VDI.attach2
call.

---
This change is [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)

### DIFF
--- a/ocaml/xapi/attach_helpers.ml
+++ b/ocaml/xapi/attach_helpers.ml
@@ -100,3 +100,18 @@ let with_vbds rpc session_id __context vm vdis mode f =
                         (fun self -> safe_unplug rpc session_id self)) !vbds;
            List.iter (Helpers.log_exn_continue "destroying VBD on VM"
                         (fun self -> Client.VBD.destroy rpc session_id self)) !vbds))
+
+(** Separates the implementations of the given backend returned from
+    the VDI.attach2 SMAPIv2 call based on their type *)
+let implementations_of_backend backend =
+  let open Storage_interface in
+  List.fold_left
+    (fun (xendisks, blockdevices, files, nbds) implementation ->
+       match implementation with
+       | XenDisk xendisk -> (xendisk::xendisks, blockdevices, files, nbds)
+       | BlockDevice blockdevice -> (xendisks, blockdevice::blockdevices, files, nbds)
+       | File file -> (xendisks, blockdevices, file::files, nbds)
+       | Nbd nbd -> (xendisks, blockdevices, files, nbd::nbds)
+    )
+    ([], [], [], [])
+    backend.implementations
--- a/ocaml/xapi/storage_access.ml
+++ b/ocaml/xapi/storage_access.ml
@@ -447,7 +447,7 @@ module SMAPIv1 = struct
                    XenDisk {
                      params = attach_info_v1.Smint.params;
                      extra = attach_info_v1.Smint.xenstore_data;
-                     backend_type = "Tapdisk3"
+                     backend_type = "vbd3"
                    };
                    (* We always get a BlockDevice from SMAPIv3, never a File, not even for ISOs *)
                    BlockDevice {
--- a/ocaml/xapi/storage_access.ml
+++ b/ocaml/xapi/storage_access.ml
@@ -449,12 +449,11 @@ module SMAPIv1 = struct
                      extra = attach_info_v1.Smint.xenstore_data;
                      backend_type = "vbd3"
                    };
-                   (* We always get a BlockDevice from SMAPIv3, never a File, not even for ISOs *)
+                   (* Currently we always get a BlockDevice from SMAPIv1, never a File, not even for ISOs *)
                    BlockDevice {
                      path = attach_info_v1.Smint.params
                    }
-                 ];
-                 domain_uuid = "0"
+                 ]
                }
             ) in
         Mutex.execute vdi_read_write_m
--- a/ocaml/xapi/storage_access.ml
+++ b/ocaml/xapi/storage_access.ml
@@ -421,14 +421,14 @@ module SMAPIv1 = struct
       | Api_errors.Server_error(code, params) ->
         raise (Backend_error(code, params))
 
-    let attach context ~dbg ~dp ~sr ~vdi ~read_write =
+    let attach2 context ~dbg ~dp ~sr ~vdi ~read_write =
       try
-        let attach_info =
-          for_vdi ~dbg ~sr ~vdi "VDI.attach"
+        let backend =
+          for_vdi ~dbg ~sr ~vdi "VDI.attach2"
             (fun device_config _type sr self ->
                let attach_info_v1 = Sm.vdi_attach device_config _type sr self read_write in
                (* Record whether the VDI is benefiting from read caching *)
-               Server_helpers.exec_with_new_task "VDI.attach" ~subtask_of:(Ref.of_string dbg) (fun __context ->
+               Server_helpers.exec_with_new_task "VDI.attach2" ~subtask_of:(Ref.of_string dbg) (fun __context ->
                    let read_caching = not attach_info_v1.Smint.o_direct in
                    let on_key = read_caching_key ~__context in
                    let reason_key = read_caching_reason_key ~__context in
@@ -442,16 +442,29 @@ module SMAPIv1 = struct
                            ~value:(attach_info_v1.Smint.o_direct_reason)
                      )
                  );
-               { params = attach_info_v1.Smint.params;
-                 o_direct = attach_info_v1.Smint.o_direct;
-                 o_direct_reason = attach_info_v1.Smint.o_direct_reason;
-                 xenstore_data = attach_info_v1.Smint.xenstore_data; }
+               {
+                 implementations = [
+                   XenDisk {
+                     params = attach_info_v1.Smint.params;
+                     extra = attach_info_v1.Smint.xenstore_data;
+                     backend_type = "Tapdisk3"
+                   };
+                   (* We always get a BlockDevice from SMAPIv3, never a File, not even for ISOs *)
+                   BlockDevice {
+                     path = attach_info_v1.Smint.params
+                   }
+                 ];
+                 domain_uuid = "0"
+               }
             ) in
         Mutex.execute vdi_read_write_m
           (fun () -> Hashtbl.replace vdi_read_write (sr, vdi) read_write);
-        attach_info
+        backend
       with Api_errors.Server_error(code, params) ->
         raise (Backend_error(code, params))
+
+    let attach _ =
+      failwith "We'll never get here: attach is implemented in Storage_impl.Wrapper"
 
     let activate context ~dbg ~dp ~sr ~vdi =
       try
@@ -1344,7 +1357,7 @@ let attach_and_activate ~__context ~vbd ~domid f =
        on_vdi ~__context ~vbd ~domid
          (fun rpc dbg dp sr vdi ->
             let module C = Storage_interface.Client(struct let rpc = rpc end) in
-            let attach_info = C.VDI.attach dbg dp sr vdi read_write in
+            let attach_info = C.VDI.attach2 dbg dp sr vdi read_write in
             C.VDI.activate dbg dp sr vdi;
             f attach_info
          )
--- a/ocaml/xapi/storage_impl.ml
+++ b/ocaml/xapi/storage_impl.ml
@@ -467,12 +467,7 @@ module Wrapper = functor(Impl: Server_impl) -> struct
       | file::_, _, _ -> response file.Storage_interface.path
       | _, blockdev::_, _ -> response blockdev.Storage_interface.path
       | _, _, xendisk::_ ->
-        let backend_kind = match xendisk.Storage_interface.backend_type with
-          | "Tapdisk3" -> "vbd3"
-          | "Qdisk" -> "qdisk"
-          | "Blkback" -> "vbd"
-          | x -> x
-        in
+        let backend_kind = xendisk.Storage_interface.backend_type in
         response ~backend_kind ~xenstore_data:xendisk.Storage_interface.extra xendisk.Storage_interface.params
       | [], [], [] -> raise (Storage_interface.Backend_error (Api_errors.internal_error, ["No File, BlockDev, or XenDisk implementation in Datapath.attach response: " ^ (rpc_of_backend backend |> Jsonrpc.to_string)]))
 
--- a/ocaml/xapi/storage_impl.ml
+++ b/ocaml/xapi/storage_impl.ml
@@ -452,23 +452,25 @@ module Wrapper = functor(Impl: Server_impl) -> struct
     let attach context ~dbg ~dp ~sr ~vdi ~read_write =
       info "VDI.attach dbg:%s dp:%s sr:%s vdi:%s read_write:%b" dbg dp sr vdi read_write;
       let backend = attach2 context ~dbg ~dp ~sr ~vdi ~read_write in
-			(* VDI.attach2 should be used instead, VDI.attach is only kept for
-				 backwards-compatibility, because older xapis call Remote.VDI.attach during SXM.
-				 However, they ignore the return value, so in practice it does not matter what
-				 we return from here. *)
+      (* VDI.attach2 should be used instead, VDI.attach is only kept for
+         backwards-compatibility, because older xapis call Remote.VDI.attach during SXM.
+         However, they ignore the return value, so in practice it does not matter what
+         we return from here. *)
       let (xendisks, blockdevs, files, _nbds) = Attach_helpers.implementations_of_backend backend in
-      let response ?(backend_kind="vbd3") ?(xenstore_data=[]) params =
-        let xenstore_data = if List.mem_assoc "backend-kind" xenstore_data then xenstore_data else ["backend-kind", backend_kind] @ xenstore_data in
-        (* XXX We've thrown o_direct info away from the SMAPIv1 info during the conversion to SMAPIv3 attach info *)
-        { params; xenstore_data; o_direct = true; o_direct_reason = "" }
+      let response params =
+        (* We've thrown o_direct info away from the SMAPIv1 info during the conversion to SMAPIv3 attach info *)
+        (* The removal of these fields does not break read caching info propagation for SMAPIv1
+         * (REQ-49), because we put this information into the VDI's sm_config elsewhere,
+         * and XenCenter looks at the relevant sm_config keys. *)
+        { params; xenstore_data = []; o_direct = true; o_direct_reason = "" }
       in
       (* If Nbd is returned, then XenDisk must also be returned from attach2 *)
-      match files, blockdevs, xendisks with
-      | file::_, _, _ -> response file.Storage_interface.path
-      | _, blockdev::_, _ -> response blockdev.Storage_interface.path
-      | _, _, xendisk::_ ->
+      match xendisks, files, blockdevs with
+      | xendisk::_, _, _ ->
         let backend_kind = xendisk.Storage_interface.backend_type in
         response ~backend_kind ~xenstore_data:xendisk.Storage_interface.extra xendisk.Storage_interface.params
+      | _, file::_, _ -> response file.Storage_interface.path
+      | _, _, blockdev::_ -> response blockdev.Storage_interface.path
       | [], [], [] -> raise (Storage_interface.Backend_error (Api_errors.internal_error, ["No File, BlockDev, or XenDisk implementation in Datapath.attach response: " ^ (rpc_of_backend backend |> Jsonrpc.to_string)]))
 
     let activate context ~dbg ~dp ~sr ~vdi =
--- a/ocaml/xapi/storage_impl.ml
+++ b/ocaml/xapi/storage_impl.ml
@@ -110,7 +110,7 @@ let string_of_date x = Date.to_string (Date.of_float x)
 module Vdi = struct
   (** Represents the information known about a VDI *)
   type t = {
-    attach_info :  attach_info option;    (** Some path when attached; None otherwise *)
+    attach_info :  backend option;    (** Some path when attached; None otherwise *)
     dps: (Dp.t * Vdi_automaton.state) list; (** state of the VDI from each dp's PoV *)
     leaked: Dp.t list;                        (** "leaked" dps *)
   } [@@deriving rpc]
@@ -156,7 +156,7 @@ module Vdi = struct
     set_dp_state dp state' t
 
   let to_string_list x =
-    let title = Printf.sprintf "%s (device=%s)" (Vdi_automaton.string_of_state (superstate x)) (Opt.default "None" (Opt.map (fun x -> "Some " ^ Jsonrpc.to_string (rpc_of_attach_info x)) x.attach_info)) in
+    let title = Printf.sprintf "%s (device=%s)" (Vdi_automaton.string_of_state (superstate x)) (Opt.default "None" (Opt.map (fun x -> "Some " ^ Jsonrpc.to_string (rpc_of_backend x)) x.attach_info)) in
     let of_dp (dp, state) = Printf.sprintf "DP: %s: %s%s" dp (Vdi_automaton.string_of_state state) (if List.mem dp x.leaked then "  ** LEAKED" else "") in
     title :: (List.map indent (List.map of_dp x.dps))
 end
@@ -306,7 +306,7 @@ module Wrapper = functor(Impl: Server_impl) -> struct
             | Vdi_automaton.Nothing -> vdi_t
             | Vdi_automaton.Attach ro_rw ->
               let read_write = (ro_rw = Vdi_automaton.RW) in
-              let x = Impl.VDI.attach context ~dbg ~dp ~sr ~vdi ~read_write in
+              let x = Impl.VDI.attach2 context ~dbg ~dp ~sr ~vdi ~read_write in
               { vdi_t with Vdi.attach_info = Some x }
             | Vdi_automaton.Activate ->
               Impl.VDI.activate context ~dbg ~dp ~sr ~vdi; vdi_t
@@ -438,8 +438,8 @@ module Wrapper = functor(Impl: Server_impl) -> struct
                 Impl.VDI.epoch_begin context ~dbg ~sr ~vdi ~persistent
              ))
 
-    let attach context ~dbg ~dp ~sr ~vdi ~read_write =
-      info "VDI.attach dbg:%s dp:%s sr:%s vdi:%s read_write:%b" dbg dp sr vdi read_write;
+    let attach2 context ~dbg ~dp ~sr ~vdi ~read_write =
+      info "VDI.attach2 dbg:%s dp:%s sr:%s vdi:%s read_write:%b" dbg dp sr vdi read_write;
       with_vdi sr vdi
         (fun () ->
            remove_datapaths_andthen_nolock context ~dbg ~sr ~vdi Vdi.leaked
@@ -448,6 +448,33 @@ module Wrapper = functor(Impl: Server_impl) -> struct
                     (Vdi_automaton.Attach (if read_write then Vdi_automaton.RW else Vdi_automaton.RO)) in
                 Opt.unbox state.Vdi.attach_info
              ))
+
+    let attach context ~dbg ~dp ~sr ~vdi ~read_write =
+      info "VDI.attach dbg:%s dp:%s sr:%s vdi:%s read_write:%b" dbg dp sr vdi read_write;
+      let backend = attach2 context ~dbg ~dp ~sr ~vdi ~read_write in
+			(* VDI.attach2 should be used instead, VDI.attach is only kept for
+				 backwards-compatibility, because older xapis call Remote.VDI.attach during SXM.
+				 However, they ignore the return value, so in practice it does not matter what
+				 we return from here. *)
+      let (xendisks, blockdevs, files, _nbds) = Attach_helpers.implementations_of_backend backend in
+      let response ?(backend_kind="vbd3") ?(xenstore_data=[]) params =
+        let xenstore_data = if List.mem_assoc "backend-kind" xenstore_data then xenstore_data else ["backend-kind", backend_kind] @ xenstore_data in
+        (* XXX We've thrown o_direct info away from the SMAPIv1 info during the conversion to SMAPIv3 attach info *)
+        { params; xenstore_data; o_direct = true; o_direct_reason = "" }
+      in
+      (* If Nbd is returned, then XenDisk must also be returned from attach2 *)
+      match files, blockdevs, xendisks with
+      | file::_, _, _ -> response file.Storage_interface.path
+      | _, blockdev::_, _ -> response blockdev.Storage_interface.path
+      | _, _, xendisk::_ ->
+        let backend_kind = match xendisk.Storage_interface.backend_type with
+          | "Tapdisk3" -> "vbd3"
+          | "Qdisk" -> "qdisk"
+          | "Blkback" -> "vbd"
+          | x -> x
+        in
+        response ~backend_kind ~xenstore_data:xendisk.Storage_interface.extra xendisk.Storage_interface.params
+      | [], [], [] -> raise (Storage_interface.Backend_error (Api_errors.internal_error, ["No File, BlockDev, or XenDisk implementation in Datapath.attach response: " ^ (rpc_of_backend backend |> Jsonrpc.to_string)]))
 
     let activate context ~dbg ~dp ~sr ~vdi =
       info "VDI.activate dbg:%s dp:%s sr:%s vdi:%s" dbg dp sr vdi;
--- a/ocaml/xapi/storage_impl.ml
+++ b/ocaml/xapi/storage_impl.ml
@@ -467,8 +467,7 @@ module Wrapper = functor(Impl: Server_impl) -> struct
       (* If Nbd is returned, then XenDisk must also be returned from attach2 *)
       match xendisks, files, blockdevs with
       | xendisk::_, _, _ ->
-        let backend_kind = xendisk.Storage_interface.backend_type in
-        response ~backend_kind ~xenstore_data:xendisk.Storage_interface.extra xendisk.Storage_interface.params
+        response xendisk.Storage_interface.params
       | _, file::_, _ -> response file.Storage_interface.path
       | _, _, blockdev::_ -> response blockdev.Storage_interface.path
       | [], [], [] -> raise (Storage_interface.Backend_error (Api_errors.internal_error, ["No File, BlockDev, or XenDisk implementation in Datapath.attach response: " ^ (rpc_of_backend backend |> Jsonrpc.to_string)]))
--- a/ocaml/xapi/storage_access.mli
+++ b/ocaml/xapi/storage_access.mli
@@ -74,7 +74,7 @@ val transform_storage_exn: (unit -> 'a) -> 'a
     [attach_info] is the result of attaching a VDI which is also activated.
     This should be used everywhere except the migrate code, where we want fine-grained
     control of the ordering of attach/activate/deactivate/detach *)
-val attach_and_activate: __context:Context.t -> vbd:API.ref_VBD -> domid:int -> (Storage_interface.attach_info -> 'a) -> 'a
+val attach_and_activate: __context:Context.t -> vbd:API.ref_VBD -> domid:int -> (Storage_interface.backend -> 'a) -> 'a
 
 (** [deactivate_and_detach __context vbd domid] idempotent function which ensures
     that any attached or activated VDI gets properly deactivated and detached. *)
--- a/ocaml/xapi/storage_migrate.ml
+++ b/ocaml/xapi/storage_migrate.ml
@@ -253,29 +253,32 @@ let tapdisk_of_attach_info (backend:Storage_interface.backend) =
 
 
 let with_activated_disk ~dbg ~sr ~vdi ~dp f =
-  let path =
-    Opt.map
-      (fun vdi ->
-         let backend = Local.VDI.attach2 ~dbg ~dp ~sr ~vdi ~read_write:false in
-         let (_, blockdevs, files, _) = Attach_helpers.implementations_of_backend backend in
-         match blockdevs, files with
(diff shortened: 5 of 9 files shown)


gaborigloi:
These PRs have to be merged together:
* https://github.com/xapi-project/xcp-idl/pull/229
* xapi-storage-plugins
* https://github.com/xapi-project/xapi-storage-script/pull/70
* https://github.com/xapi-project/xapi-storage/pull/85
* https://github.com/xapi-project/xen-api/pull/3633
* https://github.com/xapi-project/xenopsd/pull/519
* qemu-trad.pg

ocaml/xapi/attach_helpers.ml
```diff
@@ -100,3 +100,36 @@ let with_vbds rpc session_id __context vm vdis mode f =
...
+    ([], [], [], [])
+    backend.implementations
+
+(** Extracts the UNIX domain socket path and the export name from the NBD URI in
+    the NBD information returned from the VDI.attach2 SMAPIv2 call.
+    This has the format nbd:unix:<domain-socket>:exportname=<name> *)
+let parse_nbd_uri nbd =
+  let Storage_interface.{ uri } = nbd in
+  let fail () =
+    raise Api_errors.(Server_error (Api_errors.internal_error, ["Unexpected NBD URI returned from the storage backend: " ^ uri]))
```
minishrink: The second `Api_errors` is redundant.

gaborigloi:
Thanks, fixed!

---







*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633#-:-LFHpzBN-197NgHCVKIB:b-x52gc1)*

edwintorok:
Reviewed 1 of 10 files at r1.
Review status: 1 of 10 files reviewed, 2 unresolved discussions (waiting on @gaborigloi)

---

*[ocaml/xapi/attach_helpers.ml, line 128 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFMIvOr0t6J1OV8lwXD:-LFMIvOr0t6J1OV8lwXE:b-kn9mv5) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/attach_helpers.ml#L128)):*
> ```OCaml
>   in
>   match String.split_on_char ':' uri with
>   | ["nbd"; "unix"; socket; exportname] -> begin
> ```

Isn't there a similar function in xenopsd too? Should this be put somewhere common?

---

*[ocaml/xapi/storage_mux.ml, line 126 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFMJ9V43CZODQsjpyAF:-LFMJ9V43CZODQsjpyAG:b8yxb0a) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_mux.ml#L126)):*
> ```OCaml
>   end
>   module DP = struct
>     (* We'll never get here, because DP is implemented in
> ```

What prevents us from deleting this module then?

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

gaborigloi:
*[ocaml/xapi/attach_helpers.ml, line 128 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFMIvOr0t6J1OV8lwXD:-LFMK8JQ0gohrDpMizbI:bs0wwvk) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/attach_helpers.ml#L128)):*
<details><summary><i>Previously, edwintorok (Török Edwin) wrote…</i></summary><blockquote>

Isn't there a similar function in xenopsd too? Should this be put somewhere common?
</blockquote></details>

Yes, there is, I'm not sure though where to put it . Last time I asked Jon about something like this, he said it's not worth the effort factoring out the duplication.

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

gaborigloi:
*[ocaml/xapi/storage_mux.ml, line 126 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFMJ9V43CZODQsjpyAF:-LFMKKdDCHDW83N-prER:bi4rmbk) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_mux.ml#L126)):*
<details><summary><i>Previously, edwintorok (Török Edwin) wrote…</i></summary><blockquote>

What prevents us from deleting this module then?
</blockquote></details>

It has to match the SMAPIv2 interface signature, therefore I had to include `Storage_skeleton.DP`.

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

edwintorok:
Review status: 1 of 10 files reviewed, 2 unresolved discussions (waiting on @edwintorok)

---

*[ocaml/xapi/attach_helpers.ml, line 128 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFMIvOr0t6J1OV8lwXD:-LFMKX208jgKsx-xjF8Z:brkllfh) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/attach_helpers.ml#L128)):*
> where to
Would xcp-idl work? The format of that URI is part of the API after all.

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

gaborigloi:
*[ocaml/xapi/attach_helpers.ml, line 128 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFMIvOr0t6J1OV8lwXD:-LFMbCNH8TfS8eG1GMlq:b-fznoh6) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/attach_helpers.ml#L128)):*
<details><summary><i>Previously, edwintorok (Török Edwin) wrote…</i></summary><blockquote>

> where to
Would xcp-idl work? The format of that URI is part of the API after all.
</blockquote></details>

Good idea, done!

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

edwintorok:
Review status: 1 of 8 files reviewed, 5 unresolved discussions (waiting on @edwintorok and @gaborigloi)

---

*[ocaml/xapi/storage_access.ml, line 466 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSxhgSB9558p39_5zJ:-LFSxhgT2Vzyl3lu7phC:bqvhzsm) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_access.ml#L466)):*
> ```OCaml
>         raise (Backend_error(code, params))
> 
>     let attach _ =
> ```

Would compilation fail if we deleted this?

---

*[ocaml/xapi/storage_impl.ml, line 462 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSy-pP0_e5s2U8g75O:-LFSy-pP0_e5s2U8g75P:btsfhk8) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_impl.ml#L462)):*
> ```OCaml
>       let response ?(backend_kind="vbd3") ?(xenstore_data=[]) params =
>         let xenstore_data = if List.mem_assoc "backend-kind" xenstore_data then xenstore_data else ["backend-kind", backend_kind] @ xenstore_data in
>         (* XXX We've thrown o_direct info away from the SMAPIv1 info during the conversion to SMAPIv3 attach info *)
> ```

I think there was something in SMAPIv1 using this, NFS maybe for read caching? Will that still work after this change?

---

*[ocaml/xapi/storage_impl.ml, line 471 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSyKhw6t9RCNK28O0C:-LFSyKhw6t9RCNK28O0D:baek9l5) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_impl.ml#L471)):*
> ```OCaml
>       | _, _, xendisk::_ ->
>         let backend_kind = match xendisk.Storage_interface.backend_type with
>           | "Tapdisk3" -> "vbd3"
> ```

Why not use an enum instead of a string? And if we used a string why do we need to convert it to a different string?

---

*[ocaml/xapi/storage_migrate.ml, line 266 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSyf7G60Fm9yDoFvbP:-LFSyf7G60Fm9yDoFvbQ:b-imzxjz) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_migrate.ml#L266)):*
> ```OCaml
>            path
>          | [], [] ->
>            Local.VDI.detach ~dbg ~dp ~sr ~vdi;
> ```

Shouldn't this be taken care of by an outer finally?
```
attach2
finally (fun() ->
  ....
 activate 
 finally (fun () ->
   .... 
   ) (.... deactivate ...))
(... detach... )
```

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

edwintorok:
Reviewed 8 of 10 files at r1, 4 of 4 files at r3.
Review status: all files reviewed, 9 unresolved discussions (waiting on @gaborigloi)

---

*[ocaml/xapi/storage_access.ml, line 466 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSxhgSB9558p39_5zJ:-LFT09NrEaAMf32yZXBs:b-fimt2s) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_access.ml#L466)):*
<details><summary><i>Previously, edwintorok (Török Edwin) wrote…</i></summary><blockquote>

Would compilation fail if we deleted this?
</blockquote></details>

See my comment on the other failwith below, wouldn't the implementation from the skeleton be enough to satisfy the signature?

---

*[ocaml/xapi/storage_migrate.ml, line 260 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFT0bfN5FZTEY3P07f-:-LFT0bfN5FZTEY3P07f0:b-5zp82g) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_migrate.ml#L260)):*
> ```OCaml
>       (fun vdi ->
>          let backend = Local.VDI.attach2 ~dbg ~dp ~sr ~vdi ~read_write:false in
>          let (_, blockdevs, files, _) = Attach_helpers.implementations_of_backend backend in
> ```

How about xendisk implementations?

---

*[ocaml/xapi/storage_mux.ml, line 209 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFT-Ljh9_IJltDq0O2i:-LFT-Ljh9_IJltDq0O2j:b-yv1ulq) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_mux.ml#L209)):*
> ```OCaml
>       C.VDI.epoch_begin ~dbg ~sr ~vdi ~persistent
>     let attach context ~dbg ~dp ~sr ~vdi ~read_write =
>       failwith "We'll never get here: attach is implemented in Storage_impl.Wrapper"
> ```

Wouldn't the dummy implementation from Storage_skeleton ensure we got an `attach` already? And if its never reached it doesn't matter if it raises Unimplemented or "we'll never get here"

---

*[ocaml/xapi/xapi_vbd.ml, line 79 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSzvOVDRBC8z17cPD1:-LFSzvOVDRBC8z17cPD2:b8zz25w) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/xapi_vbd.ml#L79)):*
> ```OCaml
>          let device_path =
>            match files, blockdevs, nbds with
>            | { path }::_, _, _ | _, { path }::_, _ -> path
> ```

What is the desired semantics if multiple paths of same type are returned. Is this for future extensibility?

---

*[ocaml/xapi/xapi_vbd.ml, line 80 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSzYv6ASxTTPKcO22m:-LFSzYv6ASxTTPKcO22n:b-cgbelt) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/xapi_vbd.ml#L80)):*
> ```OCaml
>            match files, blockdevs, nbds with
>            | { path }::_, _, _ | _, { path }::_, _ -> path
>            | _, _, nbd::_ ->
> ```

Previously we gave precedence to nbd if present, it looks like here we only use nbd if nothing else is available.
Shouldn't we still prefer NBD first if available?

---

*[ocaml/xapi/xapi_vbd.ml, line 84 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSzmTMA8ec6ktMIlyQ:-LFSzmTMA8ec6ktMIlyR:bsf0ixu) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/xapi_vbd.ml#L84)):*
> ```OCaml
>              let unix_socket_path, export_name = Attach_helpers.parse_nbd_uri nbd in
>              NbdClient.start_nbd_client ~unix_socket_path ~export_name
>            | _ -> raise (Storage_interface.Backend_error (Api_errors.internal_error, ["No File, BlockDevice or Nbd implementation in Datapath.attach response: " ^ (Storage_interface.rpc_of_backend attach_info |> Jsonrpc.to_string)]))
> ```

You could use `[], [], []` to make sure that we cover all non-empty possibilities above.

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

gaborigloi:
*[ocaml/xapi/storage_impl.ml, line 462 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSy-pP0_e5s2U8g75O:-LFTAOcp40MjJ_QdeB3d:bt0agi6) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_impl.ml#L462)):*
<details><summary><i>Previously, edwintorok (Török Edwin) wrote…</i></summary><blockquote>

I think there was something in SMAPIv1 using this, NFS maybe for read caching? Will that still work after this change?
</blockquote></details>

Hmm, not sure, I've completely removed this from the output of the new attach2 call. (The output of this old attach call does not matter at all because it's not used anywhere. We only call it from older xapis during migration but we ignore its output: https://github.com/xapi-project/xen-api/blob/eab3129c29c2a624c0a3db65410b936a974dc952/ocaml/xapi/storage_migrate.ml#L338 )
So we'd have to figure it out, but I think Stefano told me it's not used. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

gaborigloi:
*[ocaml/xapi/storage_impl.ml, line 471 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSyKhw6t9RCNK28O0C:-LFTAt4PD9A-abQ19FqW:b-6zk4o1) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_impl.ml#L471)):*
<details><summary><i>Previously, edwintorok (Török Edwin) wrote…</i></summary><blockquote>

Why not use an enum instead of a string? And if we used a string why do we need to convert it to a different string?
</blockquote></details>

Yes, this is in my TODO list, and I'm not sure either, I agree that the conversion might be unnecessary and an enum might be better - we'd have to discuss with Stefano and / or Jon

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

gaborigloi:
*[ocaml/xapi/storage_migrate.ml, line 260 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFT0bfN5FZTEY3P07f-:-LFTB9THCkyQ3_QMhGhD:b-tfuebs) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_migrate.ml#L260)):*
<details><summary><i>Previously, edwintorok (Török Edwin) wrote…</i></summary><blockquote>

How about xendisk implementations?
</blockquote></details>

XenDisk is for atttaching the VDI to a VM. BlockDev, File, and Nbd is for attaching to dom0.
These variants are documented in the api docs: https://xapi-project.github.io/xapi-storage/#datapath 

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

edwintorok:
Review status: all files reviewed, 6 unresolved discussions (waiting on @gaborigloi)

---

*[ocaml/xapi/storage_impl.ml, line 471 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSyKhw6t9RCNK28O0C:-LFTNWOv-uOAC-Kdnj2J:bidltk) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_impl.ml#L471)):*
<details><summary><i>Previously, gaborigloi wrote…</i></summary><blockquote>

Yes, this is in my TODO list, and I'm not sure either, I agree that the conversion might be unnecessary and an enum might be better - we'd have to discuss with Stefano and / or Jon
</blockquote></details>

Ok lets see whats possible

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

gaborigloi:
*[ocaml/xapi/storage_mux.ml, line 209 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFT-Ljh9_IJltDq0O2i:-LFWSDAbDH4sFzBs8QHg:bvtqc89) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_mux.ml#L209)):*
<details><summary><i>Previously, gaborigloi wrote…</i></summary><blockquote>

I just wanted to document the situation with the error message, so that people reading the code  will know why it isn't implemented here, and in case I've made a mistake, it will be helpful to see this in the logs.
</blockquote></details>

If I deleted this compilation would fail, we need to have all the functions in the signature. Including Storage_skeleton.VDI is not ideal, because then we don't get a compilation failure in case we add a new function to the interface that is missing from here.

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

gaborigloi:
*[ocaml/xapi/xapi_vbd.ml, line 79 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSzvOVDRBC8z17cPD1:-LFWSZbF1BKK_mbuL-Un:b-krcpue) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/xapi_vbd.ml#L79)):*
<details><summary><i>Previously, edwintorok (Török Edwin) wrote…</i></summary><blockquote>

What is the desired semantics if multiple paths of same type are returned. Is this for future extensibility?
</blockquote></details>

I agree this is not the best for legibility :D. We have two types that have a `path`: `BlockDevice` and `File` (see https://xapi-project.github.io/xapi-storage/#datapath ) , we discussed with Stefano that it's helpful to distinguish between the two because these are different and support different operations.

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

gaborigloi:
*[ocaml/xapi/storage_impl.ml, line 471 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSyKhw6t9RCNK28O0C:-LFWTxzE6otclJojUdM5:b-gx4e7e) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_impl.ml#L471)):*
<details><summary><i>Previously, edwintorok (Török Edwin) wrote…</i></summary><blockquote>

Ok lets see whats possible
</blockquote></details>

@jonludlam Would it be better to return a backend_type enum from `Datapath.attach` in `XenDisk`, which could be tapdisk, blktap, or qemu, instead of a `string`. Or would it cause problems for the future if we hardcoded the list of possibilities as an enum?

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

gaborigloi:
*[ocaml/xapi/xapi_vbd.ml, line 80 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSzYv6ASxTTPKcO22m:-LFWUXqp9IEZRSwe4Juw:b8445tf) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/xapi_vbd.ml#L80)):*
<details><summary><i>Previously, edwintorok (Török Edwin) wrote…</i></summary><blockquote>

Previously we gave precedence to nbd if present, it looks like here we only use nbd if nothing else is available.
Shouldn't we still prefer NBD first if available?
</blockquote></details>

I have this order, because currently, NBD is not too efficient, because we need to use `nbd-client`, so the reads/writes go through multiple layers. So basically it's ordered according to expected performance.

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

gaborigloi:
*[ocaml/xapi/xapi_vbd.ml, line 84 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSzmTMA8ec6ktMIlyQ:-LFWVIqz1xCKVWboqo5y:b-fznoh6) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/xapi_vbd.ml#L84)):*
<details><summary><i>Previously, edwintorok (Török Edwin) wrote…</i></summary><blockquote>

You could use `[], [], []` to make sure that we cover all non-empty possibilities above.
</blockquote></details>

Good idea, done!

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

gaborigloi:
*[a discussion](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFWVf459Skl4REW16dG:-LFWVf466RCfn19UZGe9:b-vb0oe6) (no related file):*
Btw, should we "connect" xen-api to reviewable?
> " This review will only because the repo is not connected."

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

gaborigloi:
*[ocaml/xapi/storage_mux.ml, line 209 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFT-Ljh9_IJltDq0O2i:-LFWqYIP74TM2tGJ3eav:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_mux.ml#L209)):*
<details><summary><i>Previously, edwintorok (Török Edwin) wrote…</i></summary><blockquote>

Could you add that as a comment?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

gaborigloi:
*[ocaml/xapi/xapi_vbd.ml, line 79 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSzvOVDRBC8z17cPD1:-LFWqbyc6OK5WIoPnf5r:b-ct23pl) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/xapi_vbd.ml#L79)):*
<details><summary><i>Previously, edwintorok (Török Edwin) wrote…</i></summary><blockquote>

That is fine, but what I meant to ask is what happens if the path list contains more than one element, i.e. what if you have multiple BlockDevice paths? We seem to only ever use the first, in which case shouldn't this be an option rather than a list? (Unless in the future we do want to do something more useful with a list of Blockdevice paths)
</blockquote></details>

We have a path list because of the types, because from `Datapath.attach` we return an enum list, and the `implementations_of_backend` functions sorts them out into different categories.
In practice we can consider it an option that is represented as a list. This way I did not have to do anything unsafe, and the code will still work if we do return multiple BlockDevices in the future, but I don't think it will happen. But since the interface allows it, I didn't want to make any more assumptions.

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

edwintorok:
Reviewed 1 of 1 files at r5.
Review status: all files reviewed, 2 unresolved discussions (waiting on @edwintorok and @gaborigloi)

---



*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

edwintorok:
Review status: all files reviewed, 2 unresolved discussions (waiting on @edwintorok and @gaborigloi)

---

*[ocaml/xapi/storage_migrate.ml, line 266 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSyf7G60Fm9yDoFvbP:-LFXYYBXDzR0iUvJYzTC:bcmh4zh) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_migrate.ml#L266)):*
<details><summary><i>Previously, edwintorok (Török Edwin) wrote…</i></summary><blockquote>

Shouldn't this be taken care of by an outer finally?
```
attach2
finally (fun() ->
  ....
 activate 
 finally (fun () ->
   .... 
   ) (.... deactivate ...))
(... detach... )
```
</blockquote></details>

I think this is the one we said it got forgotten

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

gaborigloi:
*[ocaml/xapi/storage_migrate.ml, line 266 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSyf7G60Fm9yDoFvbP:-LFXfX5G7fVEQUncDpdJ:b-896fix) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_migrate.ml#L266)):*
<details><summary><i>Previously, edwintorok (Török Edwin) wrote…</i></summary><blockquote>

I think this is the one we said it got forgotten
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

edwintorok:
Reviewed 1 of 1 files at r6.
Review status: all files reviewed, 1 unresolved discussion (waiting on @edwintorok)

---

*[ocaml/xapi/storage_migrate.ml, line 266 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSyf7G60Fm9yDoFvbP:-LFXgIPX0e99ybbOBUBo:bmo8snh) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_migrate.ml#L266)):*
<details><summary><i>Previously, gaborigloi wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Thanks you don't necesariy need to use different names for `attached_vdi` and `vdi`, it is the same `vdi` after all. You could Opt.map just on backend in let path.

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

gaborigloi:
*[ocaml/xapi/storage_migrate.ml, line 266 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSyf7G60Fm9yDoFvbP:-LFTFEvgCGqmp0rJ-lqJ:bqhvw9s) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_migrate.ml#L266)):*
<details><summary><i>Previously, edwintorok (Török Edwin) wrote…</i></summary><blockquote>

Thanks you don't necesariy need to use different names for `attached_vdi` and `vdi`, it is the same `vdi` after all. You could Opt.map just on backend in let path.
</blockquote></details>

`attached_vdi` conains both the `vdi` and the `backend`, I did this because in `let path` I need both the `vdi` and the `backend`, and I wanted to avoid a nested Opt.map

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

edwintorok:
Meanwhile you can add ticket numbers to the commits

---

Review status: all files reviewed, 1 unresolved discussion (waiting on @edwintorok)

---



*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633#-:-LFXjcG97PmgkRv38Emm:b-1a0h87)*

gaborigloi:
Done.

---







*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633#-:-LFXl6EeDuG6rqskCqQh:b-896fix)*

gaborigloi:
*[ocaml/xapi/storage_impl.ml, line 471 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSyKhw6t9RCNK28O0C:-LFaoraZ9pVoZaTQZuGx:bo7rjvx) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_impl.ml#L471)):*
<details><summary><i>Previously, gaborigloi wrote…</i></summary><blockquote>

@jonludlam Would it be better to return a backend_type enum from `Datapath.attach` in `XenDisk`, which could be tapdisk, blktap, or qemu, instead of a `string`. Or would it cause problems for the future if we hardcoded the list of possibilities as an enum?
</blockquote></details>

Jon told us that we should do no conversion at all, we should just directly return qdisk / vbd / vbd3 from SM.

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

gaborigloi:
*[ocaml/xapi/storage_impl.ml, line 471 at r2](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFSyKhw6t9RCNK28O0C:-LFbIbPr6tBR_7lpYMgX:b-aa7ku) ([raw file](https://github.com/xapi-project/xen-api/blob/5262441508c37e732ab4e4ad009330d03e4b2666/ocaml/xapi/storage_impl.ml#L471)):*
<details><summary><i>Previously, gaborigloi wrote…</i></summary><blockquote>

Jon told us that we should do no conversion at all, we should just directly return qdisk / vbd / vbd3 from SM.
</blockquote></details>

I've fixed this now, and also for xenopsd: https://github.com/xapi-project/xenopsd/commit/5fa569d779a4a406c4577ea68854d2a1884effbc
I'll also update the SMAPIv3 docs

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

edwintorok:
Reviewed 1 of 2 files at r9, 1 of 1 files at r10, 1 of 1 files at r11.
Review status: all files reviewed, 1 unresolved discussion (waiting on @gaborigloi)

---



*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

edwintorok:
@gaborigloi I think you need to update the attach-static-vdis script too, it looks for `implementation` in the output of datapath, and it should be looking at `implementations` now

---

Review status: all files reviewed, 1 unresolved discussion (waiting on @gaborigloi)

---



*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633#-:-LFcRTF0987r1OIAK_tS:b2awaqz)*

edwintorok:
Reviewed 1 of 1 files at r12, 1 of 1 files at r13.
Review status: all files reviewed, 1 unresolved discussion (waiting on @gaborigloi)

---



*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

ocaml/xapi/storage_access.ml
```diff
@@ -442,17 +442,30 @@ module SMAPIv1 = struct
...
+                     params = attach_info_v1.Smint.params;
+                     extra = attach_info_v1.Smint.xenstore_data;
+                     backend_type = "vbd3"
+                   };
+                   (* We always get a BlockDevice from SMAPIv3, never a File, not even for ISOs *)
+                   BlockDevice {
+                     path = attach_info_v1.Smint.params
+                   }
+                 ];
+                 domain_uuid = "0"
```
gaborigloi: Remove this

ocaml/xapi/storage_impl.ml
```diff
@@ -449,6 +449,28 @@ module Wrapper = functor(Impl: Server_impl) -> struct
...
+				 backwards-compatibility, because older xapis call Remote.VDI.attach during SXM.
+				 However, they ignore the return value, so in practice it does not matter what
+				 we return from here. *)
+      let (xendisks, blockdevs, files, _nbds) = Attach_helpers.implementations_of_backend backend in
+      let response ?(backend_kind="vbd3") ?(xenstore_data=[]) params =
+        let xenstore_data = if List.mem_assoc "backend-kind" xenstore_data then xenstore_data else ["backend-kind", backend_kind] @ xenstore_data in
+        (* XXX We've thrown o_direct info away from the SMAPIv1 info during the conversion to SMAPIv3 attach info *)
+        { params; xenstore_data; o_direct = true; o_direct_reason = "" }
+      in
+      (* If Nbd is returned, then XenDisk must also be returned from attach2 *)
```
gaborigloi: we'll always have xendisk

ocaml/xapi/storage_impl.ml
```diff
@@ -449,6 +449,28 @@ module Wrapper = functor(Impl: Server_impl) -> struct
...
+      match files, blockdevs, xendisks with
+      | file::_, _, _ -> response file.Storage_interface.path
+      | _, blockdev::_, _ -> response blockdev.Storage_interface.path
+      | _, _, xendisk::_ ->
```
gaborigloi: this should come first

ocaml/xapi/storage_impl.ml
```diff
@@ -449,6 +449,28 @@ module Wrapper = functor(Impl: Server_impl) -> struct
...
+      let backend = attach2 context ~dbg ~dp ~sr ~vdi ~read_write in
+			(* VDI.attach2 should be used instead, VDI.attach is only kept for
+				 backwards-compatibility, because older xapis call Remote.VDI.attach during SXM.
+				 However, they ignore the return value, so in practice it does not matter what
+				 we return from here. *)
+      let (xendisks, blockdevs, files, _nbds) = Attach_helpers.implementations_of_backend backend in
+      let response ?(backend_kind="vbd3") ?(xenstore_data=[]) params =
+        let xenstore_data = if List.mem_assoc "backend-kind" xenstore_data then xenstore_data else ["backend-kind", backend_kind] @ xenstore_data in
+        (* XXX We've thrown o_direct info away from the SMAPIv1 info during the conversion to SMAPIv3 attach info *)
+        { params; xenstore_data; o_direct = true; o_direct_reason = "" }
```
gaborigloi: leave xenstore_data empty

ocaml/xapi/storage_impl.ml
```diff
@@ -449,6 +449,28 @@ module Wrapper = functor(Impl: Server_impl) -> struct
                 Opt.unbox state.Vdi.attach_info
              ))
 
+    let attach context ~dbg ~dp ~sr ~vdi ~read_write =
+      info "VDI.attach dbg:%s dp:%s sr:%s vdi:%s read_write:%b" dbg dp sr vdi read_write;
+      let backend = attach2 context ~dbg ~dp ~sr ~vdi ~read_write in
+			(* VDI.attach2 should be used instead, VDI.attach is only kept for
```
gaborigloi: indentation wrong

edwintorok:
Reviewed 2 of 2 files at r14.
Review status: all files reviewed, 6 unresolved discussions (waiting on @gaborigloi)

---



*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

edwintorok:
Review status: all files reviewed, 7 unresolved discussions (waiting on @gaborigloi)

---

*[ocaml/xapi/storage_impl.ml, line 471 at r14](https://reviewable.io/reviews/xapi-project/xen-api/3633#-LFsEaqN--tQw0CaKt7c:-LFsEaqO6zF4C2LT7Hjv:baevqfo) ([raw file](https://github.com/xapi-project/xen-api/blob/d2656f0a87bc273a2a418aea93008d6283cdfb82/ocaml/xapi/storage_impl.ml#L471)):*
> ```
> response ~backend_kind ~xenstore_data:xendisk.Storage_interface.extra xendisk.Storage_interface.params
> ```
File "ocaml/xapi/storage_impl.ml", line 471, characters 8-16:
Error: This function has type string -> Storage_interface.attach_info
      ocamlc ocaml/xapi/.xapi_internal.objs/stream_vdi.{cmi,cmo,cmt}

---


*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

edwintorok:
Reviewed 1 of 1 files at r15.
Review status: all files reviewed, 7 unresolved discussions (waiting on @gaborigloi)

---



*Comments from [Reviewable](https://reviewable.io/reviews/xapi-project/xen-api/3633)*

PR #3636, opened 2018-06-19
Generalize quicktest filtering

gaborigloi:
Now all the parameters of the test cases are provided by the filter.
This makes it possible to later add test that use VM templates, if the
required template is available.
Or to test migration between two SRs, both of which are automatically
selected by the filtering framework. This new framework also makes it
possible to generate every possible combination of SRs, and take up to
10 at random, and then test VDI migration / copy between them.

Now even passing the connection parameters (rpc & session_id) is handled
by this framework, using the "conn" filter.


ocaml/quicktest/qt.ml
```diff
@@ -0,0 +1,217 @@
...
+type sr_info =
+  { sr: API.ref_SR
+  ; allowed_operations: API.storage_operations_set
+  ; capabilities: string list
+  }
+
+let init_session rpc username password =
+  Client.Client.Session.login_with_password ~rpc ~uname:username ~pwd:password ~version:Datamodel_common.api_version_string ~originator:"quick_test"
+
+let get_pool rpc session_id =
```
minishrink: Why not pattern match here?

ocaml/quicktest/qt.ml
```diff
@@ -0,0 +1,217 @@
...
+
+let cli_cmd args =
+  print_endline (String.concat " " ("$ xe" :: args));
+  try
+    let output = Xapi_stdext_std.Xstringext.String.rtrim (fst(Forkhelpers.execute_command_get_output !Quicktest_args.xe_path args)) in
+    print_endline output;
+    output
+  with
+  | Forkhelpers.Spawn_internal_error(log, output, Unix.WEXITED n) ->
+    Alcotest.fail (Printf.sprintf "CLI failed: exit code=%d output=[%s] log=[%s]" n output log)
```
minishrink: Use `Alcotest.failf` here

ocaml/quicktest/qt.ml
```diff
@@ -0,0 +1,217 @@
...
+    let self = make rpc session_id ~virtual_size sr in
+    with_destroyed rpc session_id self (fun () -> f self)
+
+  let with_any rpc session_id sr_info f =
+    if List.mem `vdi_create sr_info.allowed_operations then begin
+      with_new rpc session_id sr_info.sr f
+    end else begin
+      let self = Client.Client.SR.get_VDIs ~rpc ~session_id ~self:sr_info.sr |> List.hd in
+      f self
+    end
```
minishrink: Don't think we need `begin ... end` anywhere in here?

ocaml/quicktest/qt_filter.ml
```diff
@@ -0,0 +1,200 @@
...
+  let managed_vdis =
+    Client.Client.SR.get_VDIs rpc session_id sr
+    (* NB vhd backends may delete records beneath us *)
+    |> Valid_ref_list.filter
+      (fun vdi -> Client.Client.VDI.get_managed rpc session_id vdi)
+  in
+  List.length managed_vdis
+
+(** Called before the quicktests start to save the original state.
+    This data will be used by [finish] to check for any resouce leaks. *)
```
minishrink: type: resource

ocaml/quicktest/qt_filter.ml
```diff
@@ -0,0 +1,200 @@
...
+  Client.Client.SR.get_all_records ~rpc:!A.rpc ~session_id:!session_id
+  |> List.iter (fun (ref, sr) ->
+      Printf.printf "Hashtbl.find_opt vdi_count %s\n%!" sr.API.sR_uuid;
+      match Hashtbl.find_opt vdi_count sr.API.sR_uuid with
+      | Some before ->
+          Printf.printf "Hashtbl.find_opt vdi_count %s = %d\n%!" sr.API.sR_uuid before;
+          if List.mem `scan sr.API.sR_allowed_operations then begin
+            let after = count_vdis !A.rpc !session_id ref in
+            if after <> before then
+              failwith (Printf.sprintf "VDIs leaked on SR %s: before=%d, after=%d" sr.API.sR_uuid before after)
```
minishrink: `Alcotest.failf`?

ocaml/quicktest/qt_filter.ml
```diff
@@ -0,0 +1,200 @@
...
+         |> List.exists
+           (fun pbd -> Client.Client.PBD.get_currently_attached rpc session_id pbd)
+      )
+    |> List.map (get_sr_info rpc session_id)
+
+  let only sr () = [get_sr_info !A.rpc !session_id sr]
+
+  let all_srs = lazy (list_selected_srs !A.rpc !session_id)
+
+  let all =
```
minishrink: Do we need `begin ... end` here?

ocaml/quicktest/qt_filter.ml
```diff
@@ -0,0 +1,200 @@
...
+         let rpc = !A.rpc in
+         let session_id = !session_id in
+         let vdis = Client.Client.SR.get_VDIs ~rpc ~session_id ~self:sr_info.Qt.sr in
+         let vbds =
+           List.map (fun vdi -> Client.Client.VDI.get_VBDs ~rpc ~session_id ~self:vdi) vdis
+           |> List.concat
+         in
+         match vbds with
+         | [] -> true
+         | _::_ -> false
```
minishrink: Might be worth making an `is_empty` helper for use here and  in `with_any_vdi`?

ocaml/quicktest/quicktest_async_calls.ml
```diff
@@ -32,16 +30,16 @@ let async_test session_id sr_info =
                         | `cancelled -> "cancelled")
                        result);
       match status with
-      | `failure -> Alcotest.fail (Printf.sprintf "Failure of VDI copy! error_info: %s" (String.concat "," (Client.Client.Task.get_error_info !rpc session_id task)))
+      | `failure -> Alcotest.fail (Printf.sprintf "Failure of VDI copy! error_info: %s" (String.concat "," (Client.Client.Task.get_error_info rpc session_id task)))
```
minishrink: `Alcotest.failf`?

ocaml/quicktest/quicktest_async_calls.ml
```diff
@@ -14,15 +12,15 @@ let extract_ref xml =
...
+let async_test rpc session_id sr_info () =
+  let sr = sr_info.Qt.sr in
   print_endline "Async.VDI.copy test";
-  Storage_test.VDI.with_new session_id sr (fun newvdi ->
-      let task = Client.Client.Async.VDI.copy !rpc session_id newvdi sr Ref.null Ref.null in
-      wait_for_task_complete session_id task;
+  Qt.VDI.with_new rpc session_id sr (fun newvdi ->
+      let task = Client.Client.Async.VDI.copy rpc session_id newvdi sr Ref.null Ref.null in
+      wait_for_task_complete rpc session_id task;
       print_endline (Printf.sprintf "Task completed!%!");
```
minishrink: Why not just use `sprintf` with a newline char?

ocaml/quicktest/quicktest_iso_sr.ml
```diff
@@ -1,50 +1,51 @@
...
             if List.length files <> List.length vdis then begin
               Alcotest.fail (Printf.sprintf "%s has %d files; SR has %d VDIs" !iso_path (List.length files) (List.length vdis));
             end;
-            let locations = List.map (fun vdi -> Client.Client.VDI.get_location !rpc session_id vdi) vdis in
+            let locations = List.map (fun vdi -> Client.Client.VDI.get_location rpc session_id vdi) vdis in
             (* Check each file has a VDI.location *)
             List.iter (fun file ->
                 if not(List.mem file locations) then begin
                   Alcotest.fail (Printf.sprintf "ISO %s has no corresponding VDI" file);
                 end) files;
```
minishrink: `Alcotest.failf`, also we probably don't need `begin ... end` or the ending semicolon

ocaml/quicktest/quicktest_iso_sr.ml
```diff
@@ -1,50 +1,51 @@
...
             (* Check each VDI is read-only *)
             List.iter (fun vdi ->
-                let vdir = Client.Client.VDI.get_record !rpc session_id vdi in
+                let vdir = Client.Client.VDI.get_record rpc session_id vdi in
                 if not(vdir.API.vDI_read_only) then begin
                   Alcotest.fail (Printf.sprintf "ISO VDI has read_only set to false (%s)" vdir.API.vDI_name_label);
                 end;
                 print_endline (Printf.sprintf "ISO VDI %s looks ok" vdir.API.vDI_name_label);
```
minishrink: As above

ocaml/quicktest/quicktest_vm_import_export.ml
```diff
@@ -87,22 +85,22 @@ let import_export_test session_id sr_info =
...
                     all.API.vBD_userdevice
-                    (Client.Client.VDI.get_uuid !rpc session_id orig_vbd.API.vBD_VDI)
-                    (Client.Client.VDI.get_uuid !rpc session_id all.API.vBD_VDI)
+                    (Client.Client.VDI.get_uuid rpc session_id orig_vbd.API.vBD_VDI)
+                    (Client.Client.VDI.get_uuid rpc session_id all.API.vBD_VDI)
                  )
            | "3" ->
              (* VDI should be different *)
              if all.API.vBD_VDI = orig_vbd.API.vBD_VDI
              then Alcotest.fail (Printf.sprintf "Device %s should not vary in VDIness" all.API.vBD_userdevice)
```
minishrink: I still don't understand what VDIness is???

ocaml/quicktest/qt.ml
gaborigloi: True, I left it there to make the code clearer because we have more than one line here in the else case.

ocaml/quicktest/qt_filter.ml
gaborigloi: This is currently run outside alcotest - I might have a followup PR to convert this into a test case.

ocaml/quicktest/qt_filter.ml
gaborigloi: No - I left it here for clarity, so I don't have to remember the rules about how `if` expressions are parsed :)

ocaml/quicktest/qt_filter.ml
minishrink: Ah okay, I'll stop being pedantic then :P

ocaml/quicktest/quicktest_vm_import_export.ml
gaborigloi: I think VDIness means that the VDI of one VBD is the same as the VDI of another

minishrink:
I still suggest you make some of the codestyle changes I mentioned, but you've good reason not to make others, so I've approved this as the actual code looks good.

gaborigloi:
Thanks for the review, good suggestions, I've implemented most!

PR #3637, opened 2018-06-19
CP-28227: Add API errors to clustering datamodel

minishrink:


minishrink:
I think `Cluster.pool_destroy` errors are the same as `Cluster_host.destroy`, which would be a bit redundant. I wanted to include ones specific to the pool function, although `pool_resync` needs changing as I forgot it uses `log_and_ignore_exn`

PR #3638, opened 2018-06-19
XOP-948: New SR operations cause havoc during RPU

jonludlam:


ocaml/tests/test_sr_allowed_operations.ml
```diff
@@ -0,0 +1,79 @@
...
+  let sr_ref = make_sr ~__context () in
+  let (_: API.ref_PBD) = make_pbd ~__context ~sR:sr_ref () in
+  run sm_fun _sm_ref;
+  let sr_record = Db.SR.get_record ~__context ~self:sr_ref in
+  sr_ref, sr_record
+
+let check_same_error_code =
+  let open Alcotest in
+  let open Alcotest_comparators in
+  check (option error_code) "Same error code"
```
minishrink: You can just use `Alcotest.check_raises` for this, e.g.

```
Alcotest.check_raises
  "check blocked during RPU"
  Api_errors.(Server_error (not_supported_during_upgrade, []))
  (fun () -> Xapi_sr_operations.assert_operation_valid ~__context
                   ~self:sr_ref ~op:`vdi_enable_cbt)
```

ocaml/xapi/xapi_sr_operations.ml
```diff
@@ -86,6 +94,9 @@ let valid_operations ~__context ?op record _ref' : table =
         if Hashtbl.find table op = None
         then Hashtbl.replace table op (Some(code, params))) ops in
 
+  if Helpers.rolling_upgrade_in_progress ~__context then begin
+    set_errors Api_errors.not_supported_during_upgrade [] disallowed_during_rpu
+  end;
```
minishrink: I reckon you don't need `begin ... end` here, unless that's just to make it look clearer?

jonludlam: yep, it's to make it look clearer.

ocaml/tests/test_sr_allowed_operations.ml
jonludlam: true, but I'm following the same pattern as in `test_vdi_allowed_operations.ml`, and it'll make it easier for further tests of the allowed_operations errors to ignore the params of the exception.

PR #3639, opened 2018-06-20
CP-28561: Cluster.get_network returns common clustering network

minishrink:
Previously `Cluster.get_network` returned the network of the cluster_host's PIF on master, failing if no cluster_host could be found on master. Now it will check for a common network among the cluster_hosts before failing.

### DIFF
--- a/ocaml/xapi/xapi_clustering.ml
+++ b/ocaml/xapi/xapi_clustering.ml
@@ -127,24 +127,18 @@ let find_cluster_host ~__context ~host =
     raise Api_errors.(Server_error(internal_error, [msg; (Ref.string_of host)]))
   | _ -> None
 
-let get_master_pif ~__context =
-  match find_cluster_host ~__context ~host:Helpers.(get_master ~__context) with
-  | Some self -> Db.Cluster_host.get_PIF ~__context ~self
-  | None -> raise Api_errors.(Server_error (internal_error, [ "No cluster_host exists on master" ]))
-
+(** Best-effort attempt to find a network common to the entire cluster *)
 let get_network_internal ~__context ~self =
-  let cluster_network = Db.PIF.get_network ~__context ~self:(get_master_pif ~__context) in
-  if List.exists
-    (fun cluster_host ->
-      let pif = Db.Cluster_host.get_PIF ~__context ~self:cluster_host in
-      let cluster_host_network = Db.PIF.get_network ~__context ~self:pif in
-      cluster_host_network <> cluster_network
-    ) (Db.Cluster_host.get_all ~__context)
-  then
-    debug "Not all cluster hosts of cluster %s on same network %s"
-      (Ref.string_of self) (Ref.string_of cluster_network);
-
-  cluster_network
+  let network_of_cluster_host self =
+    Db.Cluster_host.get_PIF ~__context ~self
+    |> (fun self -> Db.PIF.get_network ~__context ~self)
+  in
+  let common network = List.for_all (fun self -> network = (network_of_cluster_host self)) in
+  match Db.Cluster.get_cluster_hosts ~__context ~self with
+  | [] -> failwith ("No cluster_hosts found for cluster " ^ (Ref.string_of self))
+  | ch :: other_chs when common (network_of_cluster_host ch) other_chs ->
+    network_of_cluster_host ch
+  | _ -> failwith ("No common network found for cluster " ^ (Ref.string_of self))
 
 let assert_cluster_host_enabled ~__context ~self ~expected =
   let actual = Db.Cluster_host.get_enabled ~__context ~self in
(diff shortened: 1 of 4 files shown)


ocaml/xapi/xapi_clustering.ml
```diff
@@ -132,19 +132,30 @@ let get_master_pif ~__context =
...
+  | None ->
+    info "no cluster_host on master, let's check the cluster_hosts";
+    (* Otherwise, we try to find a common network among the cluster_host PIFs *)
+    match Db.Cluster_host.get_all ~__context with
+    | ch :: other_chs when common (get_cluster_host_network ch) other_chs ->
+      info "looks like all cluster_host PIFs are on a common network";
+      get_cluster_host_network ch
+    | _ -> info "oops";
+      (* Finally, we fail *)
+      debug "Not all cluster hosts of cluster %s on same network" (Ref.string_of self);
```
gaborigloi: I think we should check this in case master has a cluster_host and log a debug message (to keep the old behaviour)

ocaml/xapi/xapi_clustering.ml
```diff
@@ -132,19 +132,30 @@ let get_master_pif ~__context =
...
+    get_cluster_host_network cluster_host
+    (* If cluster_host exists on master, we simply return the network of its PIF *)
+  | None ->
+    info "no cluster_host on master, let's check the cluster_hosts";
+    (* Otherwise, we try to find a common network among the cluster_host PIFs *)
+    match Db.Cluster_host.get_all ~__context with
+    | ch :: other_chs when common (get_cluster_host_network ch) other_chs ->
+      info "looks like all cluster_host PIFs are on a common network";
+      get_cluster_host_network ch
+    | _ -> info "oops";
```
gaborigloi: I think we also get this case when we have zero cluster_hosts, it's better to raise a different internal error in that case.

ocaml/xapi/xapi_clustering.ml
```diff
@@ -132,19 +132,30 @@ let get_master_pif ~__context =
...
+  let common network = List.for_all (fun self -> network = (get_cluster_host_network self)) in
+  begin match find_cluster_host ~__context ~host:(Helpers.get_master ~__context) with
+  | Some cluster_host ->
+    info "cluster_host on master!";
+    get_cluster_host_network cluster_host
+    (* If cluster_host exists on master, we simply return the network of its PIF *)
+  | None ->
+    info "no cluster_host on master, let's check the cluster_hosts";
+    (* Otherwise, we try to find a common network among the cluster_host PIFs *)
+    match Db.Cluster_host.get_all ~__context with
```
gaborigloi: If we want to be even more general & correct, than we should get the cluster hosts of the cluster `self` - this argument wasn't and isn't currently used.

ocaml/xapi/xapi_clustering.ml
```diff
@@ -132,19 +132,30 @@ let get_master_pif ~__context =
...
+    | ch :: other_chs when common (get_cluster_host_network ch) other_chs ->
+      info "looks like all cluster_host PIFs are on a common network";
+      get_cluster_host_network ch
+    | _ -> info "oops";
+      (* Finally, we fail *)
+      debug "Not all cluster hosts of cluster %s on same network" (Ref.string_of self);
+      raise Api_errors.(Server_error (internal_error, [ "No cluster network exists" ]))
```
gaborigloi: I think it might be sufficient to just do `failwith`, as all non-api errors might be converted into internalerror, but I'm not sure.

edwintorok: No *common* network exists, you have networks, they are just all different...

ocaml/xapi/xapi_clustering.ml
edwintorok: Yes, we could have multiple clusters, we should not expect them all to be on same network. Would be useful to write a helper function that gives you all the cluster hosts for a particular cluster.

gaborigloi: We can use the db getter: `Db.Cluster.get_cluster_hosts`

ocaml/xapi/xapi_clustering.ml
```diff
@@ -127,24 +127,26 @@ let find_cluster_host ~__context ~host =
...
-  cluster_network
+  let get_cluster_host_network self =
+    Db.Cluster_host.get_PIF ~__context ~self
+    |> (fun self -> Db.PIF.get_network ~__context ~self)
+  in
+  let common network = List.for_all (fun self -> network = (get_cluster_host_network self)) in
+  begin match find_cluster_host ~__context ~host:(Helpers.get_master ~__context) with
+  | Some cluster_host ->
+    get_cluster_host_network cluster_host
+    (* If cluster_host exists on master, we simply return the network of its PIF *)
```
gaborigloi: Maybe we could add a warning here like before, if not all cluster hosts are on the same network. Because I assume the cluster hosts should always be on the same network, whether or not the master is in the cluster.
And if this is true, I think we should fail in both cases - that's the best behaviour, because we call this when we add a cluster host. It's better not to add a cluster host to a cluster that inconsistent and falling apart because members are on different networks. This would also make the behaviour of this function more consistent - we return a common network if there is one , otherwise we fail.

gaborigloi:
I like how you moved some of the Cluster_host.get_alls to Cluster.get_cluster_hosts, because this is more correct, I think it's worth mentioning this too in the commit message.

PR #3640, opened 2018-06-20
[477 merge]

edwintorok:
Since we are going to enable the feature flag lets merge outstanding commits from the feature branch.


ocaml/xapi/message_forwarding.ml
```diff
@@ -4297,6 +4297,17 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
   end
 
   module Cluster = struct
+
+    let forward_cluster_op ~local_fn ~__context op =
+      let localhost = Helpers.get_localhost ~__context in
+      let is_local_cluster_host self = Db.Cluster_host.get_host ~__context ~self = localhost in
+      let host = match Db.Cluster_host.get_all ~__context with
+        | [] -> Helpers.get_master ~__context
+        | lst when List.exists is_local_cluster_host lst -> localhost
```
lindig: Just a nitpick: this matches a list of hosts but this is called `lst` here and `self` below which is a bit confusing. I would have called this match `hosts` and below would have used `host::_`.

ocaml/xapi/xapi_cluster.ml
```diff
@@ -108,103 +112,82 @@ let pool_create ~__context ~network ~cluster_stack ~token_timeout ~token_timeout
...
+  info "Forgetting any cluster_hosts that couldn't be destroyed.";
+  foreach_cluster_host ~__context ~self ~log:true
+    ~fn:Client.Client.Cluster_host.forget
+    (Db.Cluster_host.get_all ~__context);
+
+  let unforgotten_cluster_hosts = List.filter
+      (fun self -> not (Db.Cluster_host.get_joined ~__context ~self))
+      (Db.Cluster_host.get_all ~__context)
+  in
+  info "If forget failed on any remaining cluster_hosts, we now delete them competely";
```
lindig: This sentence is hard to parse. Should it be "If-forget failed .."?

edwintorok: I'll add a small commit that fixes these to this PR.

ocaml/xapi/message_forwarding.ml
edwintorok: I've changed it to use `cluster_host` and `cluster_hosts` (to avoid confusion with the `host` object which is different from a `Cluster_host` object)

PR #3641, opened 2018-06-21
CA-292372: Create SM objects for all running SMAPIv2 drivers

gaborigloi:
Previously we only created SM objects for those SMAPIv2 drivers that had
an SR of that type. This is inconsistent with the SMAPIv1 behaviour,
because we always create SMs for all the SMAPIv1 drivers.

This commit ensures that we create SMs for all the running SMAPIv2
drivers, even the ones we did not know about.

I've also made this function safer, because if it raises an exception,
it will cause the startup sequence to fail. Now the function will not
fail if we fail to query one of the SMAPIv2 plugins, it will just log an
error.

### DIFF
--- a/ocaml/xapi/storage_access.ml
+++ b/ocaml/xapi/storage_access.ml
@@ -962,6 +962,8 @@ exception Message_switch_failure
 
 (** Synchronise the SM table with the SMAPIv1 plugins on the disk and the SMAPIv2
     plugins mentioned in the configuration file whitelist. *)
+(* We have to be careful in this function, because an exception raised from
+   here will cause the startup sequence to fail *)
 let on_xapi_start ~__context =
   let existing = List.map (fun (rf, rc) -> rc.API.sM_type, (rf, rc)) (Db.SM.get_all_records ~__context) in
   let explicitly_configured_drivers = List.filter_map (function `Sm x -> Some x | `All -> None) !Xapi_globs.sm_plugins in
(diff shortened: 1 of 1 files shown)


gaborigloi:
I've tested it and before I had an SR, the gfs2 SM was there, even after the toolstack restart, and it wasn't duplicated after I've added a GFS2 SR.

ocaml/xapi/storage_access.ml
```diff
@@ -1022,22 +1007,60 @@ let on_xapi_start ~__context =
...
+  let with_query_result ty f =
+    let query ty =
+      let queue_name = !Storage_interface.queue_name ^ "." ^ ty in
+      let uri () = Storage_interface.uri () ^ ".d/" ^ ty in
+      let rpc = external_rpc queue_name uri in
+      let module C = Storage_interface.Client(struct let rpc = rpc end) in
+      let dbg = Context.string_of_task __context in
+      C.Query.query ~dbg
+    in
+    try
```
edwintorok: Couldn't we use log_and_ignore_exn here? We would see the stacktrace in the log in that case (and the stacktrace might be a combined ocaml+python one if everything works correctly)

ocaml/xapi/storage_access.ml
```diff
@@ -1007,12 +989,15 @@ let on_xapi_start ~__context =
...
-               running_smapiv2_driver_queues)
-          smapiv2_drivers
+        running_smapiv2_driver_queues
+        (* The results include the prefix itself, but that is the main storage
+           queue, we don't need it *)
+        |> List.filter ((<>) !Storage_interface.queue_name)
+        |> List.map (fun driver ->
+            (* Get the last component of the queue name: org.xen.xapi.storage.sr_type -> sr_type *)
+            (* split_on_char returns a non-empty list *)
+            String.split_on_char '.' driver |> List.rev |> List.hd
```
edwintorok: This is required because we want to register unknown drivers from SMAPIv3, right?

gaborigloi: Yep, exactly.

ocaml/xapi/storage_access.ml
gaborigloi: Good idea, done.

gaborigloi:
Diff when there is no GFS2 SR:
```diff
[root@talon ~]# diff nosr_1 nosr_2
0a1,9
> uuid ( RO)                : 80baa15c-5f71-d87c-f459-451dad126784
>           name-label ( RO): NFS Volume plugin
>     name-description ( RO): This plugin manages an NFS based SR
>                 type ( RO): nfs-ng
>               vendor ( RO): None
>            copyright ( RO): (C) 2017 Citrix Inc
>        configuration ( RO): uri: URI of the storage medium
> 
> 
54a64,72
> uuid ( RO)                : d2660329-4fd8-5881-54a4-377e19bde754
>           name-label ( RO): FILEBASED filesystem Volume plugin
>     name-description ( RO): This plugin creates a FILEBASED filesystem SR on an existing folder
>                 type ( RO): filebased
>               vendor ( RO): None
>            copyright ( RO): (C) 2017 Citrix Inc
>        configuration ( RO): uri: URI of the storage medium
> 
> 
115a134,142
> 
> 
> uuid ( RO)                : a709a7d5-52f4-2736-f306-83d5b06ee878
>           name-label ( RO): GFS2 filesystem Volume plugin
>     name-description ( RO): This plugin creates a GFS2 filesystem SR on a shared block device
>                 type ( RO): gfs2
>               vendor ( RO): None
>            copyright ( RO): (C) 2015 Citrix Inc
>        configuration ( RO): uri: URI of the storage medium
```
Diff when there is a GFS2 SR:
```diff
[root@talon ~]# diff sr_1 sr_2
0a1,9
> uuid ( RO)                : 80baa15c-5f71-d87c-f459-451dad126784
>           name-label ( RO): NFS Volume plugin
>     name-description ( RO): This plugin manages an NFS based SR
>                 type ( RO): nfs-ng
>               vendor ( RO): None
>            copyright ( RO): (C) 2017 Citrix Inc
>        configuration ( RO): uri: URI of the storage medium
>
>
54a64,72
> uuid ( RO)                : d2660329-4fd8-5881-54a4-377e19bde754
>           name-label ( RO): FILEBASED filesystem Volume plugin
>     name-description ( RO): This plugin creates a FILEBASED filesystem SR on an existing folder
>                 type ( RO): filebased
>               vendor ( RO): None
>            copyright ( RO): (C) 2017 Citrix Inc
>        configuration ( RO): uri: URI of the storage medium
>
>
118c136
< uuid ( RO)                : a709a7d5-52f4-2736-f306-83d5b06ee878
---
> uuid ( RO)                : 10eb59a5-98a4-e680-e223-a18100bcbcab
```

1 is without my changes and 2 is with my changes.

gaborigloi:
So the diff looks good to me, looks as expected.

PR #3642, opened 2018-06-25
CA-292621: ensure PIF has an IP before calling clustering methods

edwintorok:
CA-292621: ensure PIF has an IP before calling clustering methods

    The IP address would be updated asynchronously by the network monitor thread,
    so we can detect there is an IP on the link but fail later when trying to read
    the IP from the DB.
    We cannot simply wait on the condition variable until the DB is updated:
    the condition is not signalled. We could use the event mechanism to wait
    until the IP field gets updated, but that adds more connections to the
    master which hurts large pool scalability.
    Also when we wake up we couldn't been woken up due to a change on
    another PIF, not the one we were watching.

    To avoid race conditions always update the IP addresses of all PIFs on a
    networking change, and check for the presence of an IP address in the DB
    in the clustering network watcher thread.
    That way we know that when we get to the clustering code we will have an
    IP address.

### DIFF
--- a/ocaml/xapi/xapi_mgmt_iface.ml
+++ b/ocaml/xapi/xapi_mgmt_iface.ml
@@ -182,14 +182,14 @@ let wait_for_clustering_ip ~__context ~(self : API.ref_Cluster_host) =
   let pIF = Db.Cluster_host.get_PIF ~__context ~self in
   let network = Db.PIF.get_network ~__context ~self:pIF in
   let bridge = Db.Network.get_bridge ~__context ~self:network in
-  let primary_address_type = Db.PIF.get_primary_address_type ~__context ~self:pIF in
   debug "Waiting for clustering IP on bridge %s (%s)" bridge (Ref.string_of pIF);
-  let get_ip () = Helpers.get_primary_ip_addr ~__context bridge primary_address_type in
+  let get_ip () = match Db.PIF.get_IP ~__context ~self:pIF with "" -> None | s -> Some s in
   let is_connected () = Helpers.get_bridge_is_connected ~__context bridge in
   wait_for_ip get_ip is_connected
 
 let on_dom0_networking_change ~__context =
   debug "Checking to see if hostname or management IP has changed";
+  Helpers.update_pif_addresses ~__context;
   (* Need to update:
      	   1 Host.hostname
      	   2 Host.address
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -142,9 +142,7 @@ let assert_bacon_mode ~__context ~host =
       ));
   debug "Bacon test: VBDs OK"
 
-let signal_networking_change ~__context =
-  Helpers.update_pif_addresses ~__context;
-  Xapi_mgmt_iface.on_dom0_networking_change ~__context
+let signal_networking_change = Xapi_mgmt_iface.on_dom0_networking_change
 
 let signal_cdrom_event ~__context params =
   let find_vdi_name sr name =


ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -186,7 +186,12 @@ let wait_for_clustering_ip ~__context ~(self : API.ref_Cluster_host) =
   debug "Waiting for clustering IP on bridge %s (%s)" bridge (Ref.string_of pIF);
   let get_ip () = Helpers.get_primary_ip_addr ~__context bridge primary_address_type in
   let is_connected () = Helpers.get_bridge_is_connected ~__context bridge in
-  wait_for_ip get_ip is_connected
+  let value = wait_for_ip get_ip is_connected in
+  begin match primary_address_type with
+  | `IPv4 -> Db.PIF.set_IP ~__context ~self:pIF ~value
+  | `IPv6 -> Db.PIF.set_IPv6 ~__context ~self:pIF ~value:[value]
```
robhoes: This would remove any additional (IPv6) addresses that the interface may have (e.g. normally at least the link-local one).

edwintorok: Good point, should I get + add instead?

edwintorok:
Jenkins test passed, are there any other changes that should be done on this PR?

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -186,7 +186,15 @@ let wait_for_clustering_ip ~__context ~(self : API.ref_Cluster_host) =
   debug "Waiting for clustering IP on bridge %s (%s)" bridge (Ref.string_of pIF);
   let get_ip () = Helpers.get_primary_ip_addr ~__context bridge primary_address_type in
   let is_connected () = Helpers.get_bridge_is_connected ~__context bridge in
-  wait_for_ip get_ip is_connected
+  let value = wait_for_ip get_ip is_connected in
+  begin match primary_address_type with
+  | `IPv4 -> Db.PIF.set_IP ~__context ~self:pIF ~value
+  | `IPv6 ->
+    let old = Db.PIF.get_IPv6 ~__context ~self:pIF in
+    let value = if List.mem value old then old else value :: old in
```
robhoes: This would be no good if the IPv6 address has actually changed, because you'd get the old and the new one in the DB. I prefer to just get all addresses at once and put all in the DB field at once to avoid complicated reasoning about corner cases like this.

Furthermore, it would be better to have just one place where this field gets set, while we are now introducing new one. The existing function is `Helpers.update_pif_addresses`, which is called from `Xapi_host.signal_networking_change`, as well as from the startup sequence (dbsync). It updates the addresses, including the netmask, gateway etc., of all local DHCP-configured PIFs.

edwintorok: I see that Helpers.update_pif_addresses does already get called by `Xapi_host.signal_networking_change` as you say, but then why does xapi_cluster end up reading an empty IP, I think I'll need to investigate a bit deeper how the race condition trigerring this bug happens.

robhoes: Note that `on_dom0_networking_change`, which wakes up the condition variable, is called from two more places.

robhoes: I am not sure that the split between `Xapi_mgmt_iface.on_dom0_networking_change` and `Helpers.update_pif_addresses` is useful. In `Xapi_host.signal_networking_change` both are called consecutively. I can't think of a reason to not just always do that, e.g. move `Helpers.update_pif_addresses` from `Xapi_host.signal_networking_change` into `Xapi_mgmt_iface.on_dom0_networking_change`. That would probably solve your problem as well.

edwintorok: Running another test in Jenkins

edwintorok: There might still be a race, we update the IPs for management (no IP on clustering yet), we signal the condition variable, then the cluster IP watcher thread wakes up, reads the IP address on the link (lets say it just received one), then later we try to read the IP address from the DB and it would not be there because we haven't processed that event yet.

I think I could do both what you suggest, and just always rely on reading the IP address from the DB even in the watcher thread, that way given the scenario above we would get a 2nd event when the IP on the clustering interface is noticed by the network monitor thread, and we'll get the IP updated properly and available when our watcher wakes up the 2nd time (the 1st time it would just sleep again due to IP = "").

edwintorok:
Thanks, tested in SR88607, 88537, and 88535, all tests passed, I held off merging until the big SMAPIv3 API change is in, but thats in so merging.

PR #3643, opened 2018-06-26
CA-289996: Should declare VMs with SR-IOV VFs non-agile for HA purposes

krizex:
VM with SR-IOV VIF is treated as agileable VM when protecting, however base on
current implementation of HA algorithm we cannot ensure it could start on other
host.

This patch marks the VM with SR-IOV VIF as non-agileable.


krizex:
Travis job stage 2 `xs-opam` is failed because the site which store the package of `menhir` is down.
aka address of http://gallium.inria.fr/~fpottier/menhir/menhir-20180530.tar.gz is not accessible.

ocaml/xapi/agility.ml
```diff
@@ -72,6 +72,13 @@ let caching_vm_t_assert_agile ~__context (ok_srs, ok_networks) vm vm_t =
   if vm_t.API.vM_VUSBs <> [] then
     raise (Api_errors.Server_error
              (Api_errors.vm_has_vusbs, [Ref.string_of vm]));
+
+  List.iter (fun vif ->
+      if Db.VIF.get_reserved_pci ~__context ~self:vif <> Ref.null then
```
robhoes: This would only work if the VM is running, but we also want to check agility for halted VMs. You could use `Xapi_network_sriov_helpers.is_sriov_network` on the VIF's network to check if it is SR-IOV. May be best to do this in `check_vif` below, since we already get the network there, and take advantage of caching.

krizex: Ah, yes, I forget `VIF.reserved_pci` will be clear when shutting down.

krizex:
Updatde

ocaml/xapi/agility.ml
robhoes: It's still there...

robhoes:
The new code looks good, but a remnant of the old patch is still there.

krizex:
Sorry I did not consider that remnant logic. We are off site tomorrow, will get back to update this PR later.

krizex:
Updated, please have a look again @robhoes , thanks.

PR #3644, opened 2018-06-28
CA-292432: Add quicktest for static-vdis script

gaborigloi:

### DIFF
--- a/ocaml/quicktest/qt.ml
+++ b/ocaml/quicktest/qt.ml
@@ -5,6 +5,7 @@ type sr_info =
   { sr: API.ref_SR
   ; allowed_operations: API.storage_operations_set
   ; capabilities: string list
+  ; required_sm_api_version : string
   }
 
 let init_session rpc username password =
--- a/ocaml/quicktest/quicktest_static_vdis.ml
+++ b/ocaml/quicktest/quicktest_static_vdis.ml
@@ -1,0 +1,33 @@
+
+let run_cmd cmd =
+  print_endline cmd;
+  Alcotest.(check int) cmd 0 (Sys.command cmd)
+
+let static_vdis args =
+  let static_vdis = !Xapi_globs.static_vdis in
+  let cmd = static_vdis ^ " " ^ (String.concat " " args) in
+  run_cmd cmd
+
+let test_static_vdis rpc session_id sr_info () =
+  let sr = sr_info.Qt.sr in
+  Qt.VDI.with_new rpc session_id sr (fun self ->
+      let uuid = Client.Client.VDI.get_uuid ~rpc ~session_id ~self in
+      static_vdis ["add"; uuid; "quicktest_static_vdis"];
+      static_vdis ["list"];
+      static_vdis ["attach"; uuid];
+      static_vdis ["detach"; uuid];
+      let vbds = Client.Client.VDI.get_VBDs ~rpc ~session_id ~self in
+      Alcotest.(check int) "There must be no VBDs after static-vdis detach"
+        0 (List.length vbds);
+      static_vdis ["del"; uuid]
+    )
+
+let tests () =
+  let open Qt_filter in
+  [ [ "test_static_vdis for SMAPIv1", `Slow, test_static_vdis ]
+    (* We need to be able to run VDI.generate_config on SMAPIv1 SRs *)
+    |> conn |> sr SR.(all |> smapiv1 |> allowed_operations [`vdi_create; `vdi_destroy] |> has_capabilities Sr_capabilities.[vdi_generate_config])
+  ; [ "test_static_vdis for SMAPIv3", `Slow, test_static_vdis ]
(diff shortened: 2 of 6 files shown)


ocaml/quicktest/qt.ml
```diff
@@ -5,6 +5,7 @@ type sr_info =
   { sr: API.ref_SR
   ; allowed_operations: API.storage_operations_set
   ; capabilities: string list
+  ; required_sm_api_version : string
```
minishrink: Why not store these as floats or (int * int) for major * minor versions?

ocaml/quicktest/quicktest_static_vdis.ml
```diff
@@ -0,0 +1,33 @@
...
+  run_cmd cmd
+
+let test_static_vdis rpc session_id sr_info () =
+  let sr = sr_info.Qt.sr in
+  Qt.VDI.with_new rpc session_id sr (fun self ->
+      let uuid = Client.Client.VDI.get_uuid ~rpc ~session_id ~self in
+      static_vdis ["add"; uuid; "quicktest_static_vdis"];
+      static_vdis ["list"];
+      static_vdis ["attach"; uuid];
+      static_vdis ["detach"; uuid];
```
minishrink: `List.iter static_vdis [ ...]`?

ocaml/quicktest/quicktest_static_vdis.ml
```diff
@@ -0,0 +1,33 @@
...
+      Alcotest.(check int) "There must be no VBDs after static-vdis detach"
+        0 (List.length vbds);
+      run_cmd ("/opt/xensource/bin/static-vdis del " ^ uuid)
+    )
+
+let tests () =
+  let open Qt_filter in
+  [ [ "test_static_vdis for SMAPIv1", `Slow, test_static_vdis ]
+    (* We need to be able to run VDI.generate_config on SMAPIv1 SRs *)
+    |> conn |> sr SR.(all |> smapiv1 |> allowed_operations [`vdi_create; `vdi_destroy] |> has_capabilities Sr_capabilities.[vdi_generate_config])
```
minishrink: I assume `vdi_create` and `vdi_destroy` are common to VDI quicktests, maybe we should have a shorthand for this? Or would it just be pointless cruft?

ocaml/quicktest/qt.ml
gaborigloi: Because I did not want to parse it, but wanted to keep things simple for now

ocaml/quicktest/quicktest_static_vdis.ml
gaborigloi: In my opinion the code is more readable this way, because it looks like a command :)

ocaml/quicktest/quicktest_static_vdis.ml
gaborigloi: I think it's better this way, because we can just look at the test list and see exactly what the tests require.

ocaml/quicktest/quicktest_static_vdis.ml
```diff
@@ -0,0 +1,33 @@
+
+let run_cmd cmd =
+  print_endline cmd;
+  Alcotest.(check int) cmd 0 (Sys.command cmd)
+
+let static_vdis args =
+  let static_vdis = "/opt/xensource/bin/static-vdis" in
```
edwintorok: Xapi_globs.static_vdis

ocaml/quicktest/quicktest_static_vdis.ml
```diff
@@ -0,0 +1,33 @@
...
+  let cmd = static_vdis ^ " " ^ (String.concat " " args) in
+  run_cmd cmd
+
+let test_static_vdis rpc session_id sr_info () =
+  let sr = sr_info.Qt.sr in
+  Qt.VDI.with_new rpc session_id sr (fun self ->
+      let uuid = Client.Client.VDI.get_uuid ~rpc ~session_id ~self in
+      static_vdis ["add"; uuid; "quicktest_static_vdis"];
```
edwintorok: Would be nice if these could call `Static_vdis.permanent_vdi_attach`, etc. but due to Forkhelpers that is not trivial.
Maybe we could have a test mode for `Helpers.call_script` that instead of calling Forkhelpers just does it directly.

ocaml/quicktest/quicktest_static_vdis.ml
```diff
@@ -0,0 +1,33 @@
...
+      static_vdis ["list"];
+      static_vdis ["attach"; uuid];
+      static_vdis ["detach"; uuid];
+      let vbds = Client.Client.VDI.get_VBDs ~rpc ~session_id ~self in
+      Alcotest.(check int) "There must be no VBDs after static-vdis detach"
+        0 (List.length vbds);
+      run_cmd ("/opt/xensource/bin/static-vdis del " ^ uuid)
```
edwintorok: Why not `static_vdis ["del"; uuid]`?

gaborigloi:
Thanks, fixed

PR #3646, opened 2018-06-28
Update xapi.opam to match xs-opam meta data

lindig:
This adds a dependency as it is found (and tested) in the xs-opam
repository.

### DIFF
--- a/xapi.opam
+++ b/xapi.opam
@@ -43,6 +43,7 @@ depends: [
   "xenctrl"
   "xenstore"
   "yojson"
+  "alcotest"
 ]
 depexts: [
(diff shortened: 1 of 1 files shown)


xapi.opam
```diff
@@ -43,6 +43,7 @@ depends: [
   "xenctrl"
   "xenstore"
   "yojson"
+  "alcotest"
```
gaborigloi: Hmm I think it's incorrect in xs-opam then, because `alcotest` is already there at the top, so now it's duplicated

lindig: Yes - it's twice in xs-opam. So let's close this and correct it there.

lindig:
The fix needs to go into xs-opam.

PR #3647, opened 2018-07-03
Feature/cc/ccm to staging branch

makunterry:


makunterry:
@lindig 
Have combined all commits into one.

mg12:
- in the commit comment "Disable compression", I think we should add CP-28468 as a reference to justify the compression change.
- I've attempted to run the code `ldd ${STUNNEL} | grep -q "citrix\/libssl"` a few times, and it took 9ms or less -- I think that's ok.

scripts/init.d-xapissl
```diff
@@ -128,6 +127,12 @@ EOF
             # If we had v5.06 or newer we would need to allow SSLv3 with
             # writec 'options -NO_SSLv3'
             # but all earlier versions choke on that option.
+
+            # Only give option 'LEGACY_CLIENT_ACCEPT' when ssl-legacy == true
+            # and CCM installed
+            if ldd ${STUNNEL} | grep -q "citrix\/libssl"; then
```
lindig: I don't think we need the backslash here: `grep -q 'citrix/libssl'` would be better. It would also work with double quotes (at we have now) because '/` is not a character that needs escaping on the shell or for grep.

lindig:
I don't manage that branch so I don't think I should merge to it.

PR #3648, opened 2018-07-03
Replace OPasswd with Unixpwd

lindig:
This PR replaces the Opasswd library for updating passwords with a new library. The main motivation is
we suspect a problem with the ctypes-based binding for opasswd that we weren't able to track down (so far) that cause sometime corruption of passwords.

The code for the library is in unixpwd/ and it could be maintained as an independent library. Currently there is no point in doing that because xapi is the only client of this library and it is easier to maintain them together. The library contains code for unit-testing - both in C and OCaml. The uni-testing code builds binaries that exercise the library and that can be debugged using a debugger.

```
unixpwd/
├── c
│   ├── jbuild
│   ├── main.c   - unit testing
│   ├── Makefile - Makefile for unit testing
│   ├── unixpwd.c
│   ├── unixpwd.h
│   └── unixpwd_stubs.c
├── LICENSE
├── Makefile
├── README.md
├── src
│   ├── jbuild
│   ├── unixpwd.ml
│   └── unixpwd.mli
├── test
│   ├── jbuild
│   └── main.ml - unit testing
├── unixpwd.install
└── unixpwd.opam
```

The actual changes in xapi are quite small because and adds a bit of logging .


unixpwd/c/unixpwd.c
```diff
@@ -0,0 +1,282 @@
...
+char           *
+unixpwd_getpwd(const char *user)
+{
+    struct passwd   pwd,
+                   *pw;
+    char            buf[BUFLEN];
+
+    if (getpwnam_r(user, &pwd, buf, BUFLEN, &pw) == 0 && pw)
+        return strdup(pw->pw_passwd);
+    if (errno == 0)
```
edwintorok: The manpage says `If one wants to check errno after the call, it should be set to zero before the call.`

lindig: You got to love C libraries. Good catch. I noticed already that you need to check both return code and the `pw` pointer.

mseri:
We had noticed that comment from the man page, and indeed one of the suggestions was to rewrite the library using the reentrant versions, we decided against it because we have struggled enough and actually I think that the explicit bindings that @lindig implemented are way more readable than a reentrant version of opasswd with ctypes

unixpwd/c/unixpwd.c
lindig: I've added a commit to handle this

unixpwd/c/unixpwd.c
```diff
@@ -0,0 +1,284 @@
...
+        return rc;
+    }
+    if (fchown(tmp, statbuf.st_uid, statbuf.st_gid) != 0) {
+        rc = errno;
+        close(tmp);
+        return rc;
+    }
+    if (fchmod(tmp, statbuf.st_mode) != 0) {
+        rc = errno;
+        close(tmp);
```
edwintorok: minor: the cleanup code is the same for all three of these, could they be part of the same if `stat (...) != 0 || fchown(...) != 0 || fchmod(...) != 0)`?
Or another way I used to write C code is to have something that resembles a try/finally:
```
rc = 0;
tmp = ...
do {
 if (... something fails... ) {
    rc = <code>;
    break;
  }
 ...
} while(0);
close(tmp);
return rc;
```
(or errno instead of rc)

unixpwd/src/unixpwd.ml
```diff
@@ -0,0 +1,35 @@
...
+    external getspw :   string -> string          = "caml_unixpwd_getspw"
+    external get    :   string -> string          = "caml_unixpwd_get"
+    external setpwd :   string -> string -> unit  = "caml_unixpwd_setpwd"
+    external setspw :   string -> string -> unit  = "caml_unixpwd_setspw"
+    external unshadow : unit   -> string          = "caml_unixpwd_unshadow"
+end
+
+exception Error of string
+
+let wrap f = try f () with Failure msg -> raise (Error msg)
```
edwintorok: I think the stubs could raise this error directly, but it would be more complicated to do so, so this is probably fine.

unixpwd/c/unixpwd.c
lindig: I agree that there is some repetitiveness in the code that would be good to reduce. I like the first suggestion.

unixpwd/src/unixpwd.ml
lindig: Yes. I wanted to keep the C bindings small and therefore avoided dealing with custom types.

unixpwd/c/unixpwd_stubs.c
```diff
@@ -0,0 +1,137 @@
...
+    char           *user;
+    char           *passwd;
+    CAMLlocal1(pw);
+
+    user = String_val(caml_user);
+    passwd = unixpwd_getpwd(user);
+    if (passwd == NULL && errno != 0)
+        caml_failwith(strerror(errno));
+    if (passwd == NULL)
+        caml_failwith("unspecified error in caml_getpwd()");
```
robhoes: There is no function called `caml_getpwd`.

unixpwd/c/unixpwd.c
```diff
@@ -0,0 +1,258 @@
...
+    struct stat     statbuf;
+    int             rc;
+    int             updated = 0;
+
+    strncpy(tmp_name, TMP_PASSWD, sizeof tmp_name);
+    tmp = mkstemp(tmp_name);
+    if (tmp == -1)
+        return errno;
+    if (stat(ETC_PASSWD, &statbuf) != 0
+        || fchown(tmp, statbuf.st_uid, statbuf.st_gid) != 0
```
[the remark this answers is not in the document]
edwintorok: I think `||` and `&&` are guaranteed to short-circuit evaluation.

unixpwd/c/jbuild
```diff
@@ -0,0 +1,9 @@
+(jbuild_version 1)
+
+(library
+ ((name unixpwd.stubs)
+  (public_name unixpwd.stubs)
```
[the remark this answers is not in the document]
lindig: I am not sure but `unixpwd` is public and it can only refer to this when it is public, too. Trying to make this non-public leads to a complain:

ocaml/xapi/config_file_sync.ml
```diff
@@ -48,7 +53,9 @@ let transmit_config_files s =
    the entire /etc/password file. We need to make sure we send the
    "un-shadowed" passwd file, so that slaves don't overwrite root's
    password with an 'x'. *)
-let legacy_transmit_passwd s = unshadow () |> write_to_fd s
+let legacy_transmit_passwd s =
```
jonludlam: Is this comment still true? If so, it'd be helpful to know here when it will become false

unixpwd/c/unixpwd.c
```diff
@@ -0,0 +1,258 @@
...
+            pw->pw_passwd = password;
+            updated++;
+        }
+        putpwent(pw, tmp_file);
+    }
+    endpwent();
+
+    fclose(tmp_file);
+    if (rc != ENOENT)
+        return rc;
```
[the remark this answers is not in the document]
lindig: I am not so sure. The file could help diagnosing the problem.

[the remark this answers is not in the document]
robhoes: OK fair enough.

ocaml/xapi/config_file_sync.ml
robhoes: We can't rely on tag numbers before Ely to make a lot of sense unfortunately. A big of digging in the history suggests that, while this commit went into master quite long ago, just after the release of Clearwater, the first release it got into was Dundee. This is probably because Clearwater-Creedence-Cream were separate from master for quite long.

robhoes: In other words, we will still need this legacy function for a while longer to support upgrades.

lindig:
I have addressed the comments (in individual commits) and rebased on top of master.

ocaml/xapi/config_file_sync.ml
jonludlam: OK, while we're in the area can we adjust the comment to state precisely when we can remove the code?

unixpwd/c/unixpwd.c
```diff
@@ -0,0 +1,258 @@
...
+
+    strncpy(tmp_name, TMP_PASSWD, sizeof tmp_name);
+    tmp = mkstemp(tmp_name);
+    if (tmp == -1)
+        return errno;
+    if (stat(ETC_PASSWD, &statbuf) != 0
+        || fchown(tmp, statbuf.st_uid, statbuf.st_gid) != 0
+        || fchmod(tmp, statbuf.st_mode) != 0
+        || (tmp_file = fdopen(tmp, "w")) == NULL) {
+        rc = errno;
```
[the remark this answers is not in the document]
lindig: That is a good point - I think this should be checked

lindig:
Is this good to go?

PR #3649, opened 2018-07-05
CA-291017: Unable to connect server in pool of 64 physical hosts

krizex:
Stunnel restrict it could only use 125/256 of the fds that limited by system. We
set the fd limit to 1024 by default, so stunnel could only use 500 fds for
connection. It is not enough for the master host in pool of 64 hosts when the
pool experiencing high loads.

This fix target to reconfigure the fd limites for xapi and stunnel from 1024 to
4096 to resolve the issue.

### DIFF
--- a/scripts/init.d-xapissl
+++ b/scripts/init.d-xapissl
@@ -158,6 +158,8 @@ start() {
         "@LIBEXECDIR@/generate_ssl_cert" ${PEMFILE} ${CN}
     fi
     writeconffile
+    # Increase the fd limit to 2048, so stunnel could accept 1000 client connections
+    ulimit -n 2048
     start_daemon ${STUNNEL} ${SSLCONFFILE}
     RETVAL=$?
 


lindig:
The commit message could be improved:

```
Increase xapi's process limit for the number of open file descriptors to 4096. This is required for stunnel connections on the master of a pool with 64 hosts. The previous limit was 1024.
```

The change but itself is fine. At least aesthetically, connection multi-plexing is desirable. I don't know about resources (outside of fds) that could be saved by it.

lindig:
As pointed otu by @edwintorok, we can't have fd > 1024 when we are using `Unix.select`, as we do:

edwintorok:
There is a drop-in implementation of Unix.select here based on `poll` https://xenbits.xen.org/gitweb/?p=xen.git;a=blob;f=tools/ocaml/xenstored/select.ml;h=0455e163e306ceae4e0606e57845fde1660c15c7;hb=HEAD
https://xenbits.xen.org/gitweb/?p=xen.git;a=blob;f=tools/ocaml/xenstored/select.mli;h=39127791720a829deb4b43867661aba9372feec5;hb=HEAD

lindig:
My previous comment regarding the commit message is not correct: this does not change xapi's FD limit but that of stunnel. Maybe the commit message could be:

```
Increase stunnel's process limit for the number of open file descriptors to 4096. 
This is required for stunnel connections on the master of a pool with 64 hosts. 
The previous limit was 1024.
```

We still can't have more than 1024 FDs in xapi but this pull request does not change it.

PR #3650, opened 2018-07-05
CA-292063: detach static VDIs when HA is disarmed on boot

edwintorok:
HA is disarmed on boot if we determine that it got disabled on the pool
while we were off.
However the static VDIs stayed attached, which prevented the PBD from
getting unplugged, e.g. on shutdown. Normally the shutdown code would unplug
these but since HA was disabled that code does not run.

Clean up any static VDIs at the point where we disarm HA on boot by
releasing all HA resources, not just the monitor.
The monitor is stopped when releasing the HA resources, so replace the
call to stop a monitor with releasing the resources.

Dev-tested on a virtual pool and tap-ctl list wasn't showing any active disks after this patch and I was able to unplug the PBDs.

### DIFF
--- a/ocaml/xapi/xapi.ml
+++ b/ocaml/xapi/xapi.ml
@@ -468,7 +468,8 @@ let resynchronise_ha_state () =
          | true, false ->
            info "HA has been disabled on the Pool while we were offline; disarming HA locally";
            Localdb.put Constants.ha_armed "false";
-           Xapi_ha.Monitor.stop ()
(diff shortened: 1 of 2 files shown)


edwintorok:
Good point, the redo log might also be enabled for reasons other than HA, I'll add a second commit.

edwintorok:
I've opened ticket CA-293622 for that (the statefile/metadata VDIs are not deleted, not even with xHA).

gaborigloi:
Thanks!

PR #3651, opened 2018-07-05
Quicktest to run VDI ops on empty VDI with max supported size

gaborigloi:

### DIFF
--- a/ocaml/quicktest/qt_filter.ml
+++ b/ocaml/quicktest/qt_filter.ml
@@ -187,6 +187,11 @@ module SR = struct
   let smapiv3 =
     sr_filter (fun i -> not (is_smapiv1 i))
 
+  let has_type types sr_info =
+    List.mem (Client.Client.SR.get_type ~rpc:!A.rpc ~session_id:!session_id ~self:sr_info.Qt.sr) types
+
+  let thin_pro = sr_filter (has_type ["gfs2"; "nfs"; "smb"; "ext"; "file"])
+
   (** Creates a [Alcotest.test_case] from the given [storage_test_case] using the
       specified session ID and SR *)
   let specialise (name, speed, test) sr_info =
--- a/ocaml/quicktest/quicktest_max_vdi_size.ml
+++ b/ocaml/quicktest/quicktest_max_vdi_size.ml
@@ -1,0 +1,60 @@
+
+(** From XenServer 7.5 Configuration Limits:
+    https://docs.citrix.com/content/dam/docs/en-us/xenserver/current-release/downloads/xenserver-config-limits.pdf
+    Using (2TiB - 4GiB), instead of the (2TB - 4GB) limit defined in the above
+    document, does not work, we cannot create a VDI of that size. *)
+let max_vdi_size = Sizes.((2L ** tb) -* (4L ** gb))
+
+let with_max_vdi rpc session_id sr f =
+  Qt.VDI.with_new rpc session_id ~virtual_size:max_vdi_size sr
+    (fun vdi ->
+       (* We write some data to the very end of the VDI to ensure the IO code
+          gets tested with large offsets *)
+       Qt.VDI.with_open rpc session_id vdi `RW
+         (fun fd ->
+            let data = Bytes.make 1024 't' in
+            let offset = Int64.sub max_vdi_size 1024L in
+            assert (Unix.LargeFile.lseek fd offset Unix.SEEK_SET = offset);
+            assert (Unix.write fd data 0 1024 = 1024)
+         );
+       f rpc session_id sr vdi
+    )
+
+let test_copy rpc session_id sr_info () =
+  let sr = sr_info.Qt.sr in
+  with_max_vdi rpc session_id sr
+    (fun rpc session_id sr vdi ->
+       Qt.VDI.with_destroyed rpc session_id
+         (Client.Client.VDI.copy ~rpc ~session_id ~vdi ~base_vdi:API.Ref.null ~into_vdi:API.Ref.null ~sr)
+         (fun () -> ())
+    )
+
+(** This does not reproduce CA-292288 *)
+let test_export_import rpc session_id sr_info () =
+  (* We only want to export to a sparse VHD, not to a fully-inflated raw file *)
+  let format = "vhd" in
+  let sR = sr_info.Qt.sr in
+  with_max_vdi rpc session_id sR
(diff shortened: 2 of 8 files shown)


gaborigloi:
Tested on GFS2 & NFS SRs, on GFS2 it did not run the export/import test, as intended.

ocaml/quicktest/qt_filter.ml
```diff
@@ -187,6 +187,11 @@ module SR = struct
   let smapiv3 =
     sr_filter (fun i -> not (is_smapiv1 i))
 
+  let has_type types sr_info =
+    List.mem (Client.Client.SR.get_type ~rpc:!A.rpc ~session_id:!session_id ~self:sr_info.Qt.sr) types
+
+  let thin_pro = sr_filter (has_type ["gfs2"; "nfs"; "smb"; "ext"; "file"])
```
edwintorok: Would be nice if the SR themselves exposed some capability that helped us tell the difference.

ocaml/quicktest/quicktest_max_vdi_size.ml
```diff
@@ -0,0 +1,46 @@
+
+(** From XenServer 7.5 Configuration Limits
+    https://docs.citrix.com/content/dam/docs/en-us/xenserver/current-release/downloads/xenserver-config-limits.pdf *)
+let max_vdi_size = Sizes.((2L ** tb) -* (4L ** gb))
```
edwintorok: Is it `TB` or `TiB`? Not sure which one the config limits meant, since this is a limit due to sizes of integers I would assume that it would somehow be related to powers of 2, and thus `tib` and `gib` instead of `tb` and `gb`.
We should check which one works, and update the config limits docs to be correct.

ocaml/quicktest/quicktest_max_vdi_size.ml
```diff
@@ -0,0 +1,46 @@
...
+         )
+         (fun () -> Sys.remove file)
+    )
+
+let tests () =
+  let open Qt_filter in
+  [ [ "VDI.copy", `Slow, test_copy ] |> conn |> sr SR.(all |> allowed_operations [`vdi_create; `vdi_destroy] |> thin_pro)
+  ; ["VDI.export/import to/from VHD file", `Slow, test_export_import]
+    (* GFS2 does not export to a sparse VHD file *)
+    |> conn |> sr SR.(all |> allowed_operations [`vdi_create; `vdi_destroy] |> thin_pro |> not_type "gfs2")
```
edwintorok: Can we query what export formats an SR supports?

gaborigloi: We cannot query this - what I meant is that GFS2 exports fully-inflated VHD files, not sparse ones, because that hasn't been implemented. I'll clarify the comment.

ocaml/quicktest/quicktest_max_vdi_size.ml
gaborigloi: I just assumed that the document is correct, and it's `TB`.
But I'll check, so that we know what is the truth. I wouldn't change the document based on one experiment, I think more testing goes into verifying these limits. I just viewed this document as a contract of what XenServer should be able to do. So I'd rather not change it now, because probably other people have also been relying on this contract.

But if it supports `TiB`, I can maybe add another "stress test" with this limit.

ocaml/quicktest/qt_filter.ml
gaborigloi: I don't think they do, sadly, I've checked the SR and SM fields.  Is there a way to figure this out @stefanopanella ?

ocaml/quicktest/quicktest_max_vdi_size.ml
gaborigloi: It failed:
```
[ERROR]             Quicktest_max_vdi_size                    0   VDI.copy on SR [NFS Storage Repository 1].
[ERROR]             Quicktest_max_vdi_size                    1   VDI.export/import to/from VHD file on SR [NFS Storage Repository 1].
[OK]                Quicktest_static_vdis                     0   test_static_vdis for SMAPIv1 on SR [NFS Storage Repository 1].
-- Quicktest_max_vdi_size.000 [VDI.copy on SR [NFS Storage Repository 1].] Failed --
in _build/_tests/Quicktest_max_vdi_size.000.output:
[exception] Server_error(SR_BACKEND_FAILURE_79, [ ; VDI Invalid size [opterr=VDI size must be between 1 MB and 2093050 MB];  ])
Raised at file "ocaml/xapi-client/client.ml", line 6, characters 31-75
Called from file "ocaml/xapi-client/client.ml", line 18, characters 55-110
Called from file "ocaml/xapi-client/client.ml", line 11506, characters 6-55
Called from file "ocaml/quicktest/qt.ml", line 125, characters 4-321
Called from file "ocaml/quicktest/qt.ml", line 146, characters 15-51
Called from file "src/alcotest.ml", line 269, characters 8-14


The full test results are available in `_build/_tests`.
2 errors! in 59.932s. 92 tests run.
Fatal error: exception Alcotest.Test_error
Raised at file "lib/xapi-stdext-pervasives/pervasiveext.ml", line 26, characters 6-15
Called from file "ocaml/quicktest/quicktest.ml", line 20, characters 2-1023
```

gaborigloi: So I'd leave the test as it is, because `TiB` doesn't work

edwintorok: Ok good to know, can you add a comment saying that TiB doesn't work and TB is the correct limit?

gaborigloi: Sure, done.

ocaml/quicktest/qt_filter.ml
edwintorok: Since this is just test code it is probably fine to hardcode, just wanted to check there isn't another way.

gaborigloi: Stefano said this is the right approach. It's best to just filter based on SR type.

gaborigloi:
These tests are quite quick, they finished on an NFS SR in like 20 seconds.

edwintorok:
I think this looks good. We should add a comment saying whether this can reproduce the >1TiB VDI import problem or not (it is still a valuable test to have even if it cannot)

gaborigloi:
I've added a comment.

PR #3653, opened 2018-07-09
CP-28753: add CLUSTER_HOST_FENCING message

edwintorok:
This message is created by clustering daemon fencing hooks,
and needed by XenCenter to localise the message.


ocaml/xapi-consts/api_messages.ml
```diff
@@ -46,6 +46,8 @@ let ha_host_failed = addMessage "HA_HOST_FAILED" 3L
 
 let ha_host_was_fenced = addMessage "HA_HOST_WAS_FENCED" 4L
```
gaborigloi: Just one suggestion: this could be moved to line 138 where we have the otehr clusterting messages

edwintorok: Thanks, moved.

PR #3654, opened 2018-07-09
CP-27915: test that VBD.create isn't allowed for cbt_metadata VDI

gaborigloi:


edwintorok:
This looks good, looking at the ticket more unit tests are expected, will those be part of this PR or separate PRs?

gaborigloi:
No, I've decided not to do the rest, because I think they are hard to write, so it's not really worth it.
This unit test was the most important, after this I'll close the ticket.

PR #3655, opened 2018-07-10
CP-28753: drop unused CLUSTER_HOST_CREATION_FAILED message

edwintorok:
We only use CLUSTER_HOST_ENABLE_FAILED, after some refactoring create is
calling the same resync_host API that calls enable, so there is only
message that can be raised.

Also add a comment that the cluster host fencing script is raised by
external script, so that it doesn't get removed next time someone does a
dead code cleanup.


edwintorok:
Thanks, fixed the typo

PR #3656, opened 2018-07-10
CA-287525 make unxipwd an internal library

lindig:
This simplifies the jbuild files and makes the library internal.


edwintorok:
Might be worth checking whether the xenopsd RPM can still be built, I think it depends on XAPI, not sure whether it'd also want to link with unixpwd transitively.

lindig:
I suggest to merge this and to keep an eye on builds in xs-opam.

PR #3657, opened 2018-07-10
CA-292676: Filter out missing VDIs when looking for some to use

jonludlam:


jonludlam:
Good plan, done.

PR #3659, opened 2018-07-11
CA-289650: Wait for the pidfile from udhcpd before releasing lock

jonludlam:
The pidfile is written _after_ udhcpd daemonizes, so we need to wait
for a short amount of time until it turns up.

### DIFF
--- a/ocaml/xapi/xapi_udhcpd.ml
+++ b/ocaml/xapi/xapi_udhcpd.ml
@@ -162,7 +162,18 @@ let restart_nolock () =
   let pid = try Unixext.pidfile_read !Xapi_globs.udhcpd_pidfile with _ -> None in
   Opt.iter Unixext.kill_and_wait pid;
   let (_: string * string) = execute_command_get_output !Xapi_globs.busybox [ "udhcpd"; !Xapi_globs.udhcpd_conf ] in
-  ()
+  let start = Mtime_clock.counter () in
+  let rec wait_for_pid n =
+    let now = Mtime_clock.count start in
+    if Mtime.Span.to_s now > 30.0 then failwith "Failed to start udhcpd";
+    let pid = try Unixext.pidfile_read !Xapi_globs.udhcpd_pidfile with _ -> None in
+    match pid with
+    | Some _ -> ()
+    | None ->
+      if n>0 && n mod 10 = 0 then debug "Continuing to wait for the pidfile from udhcpd";
+      Thread.delay 0.1;
+      wait_for_pid (n+1)
+  in wait_for_pid 0
 
 let find_lease_nolock vif =
   try Some (List.find (fun l -> l.vif = vif) !assigned)


ocaml/xapi/xapi_udhcpd.ml
```diff
@@ -162,7 +162,18 @@ let restart_nolock () =
   let pid = try Unixext.pidfile_read !Xapi_globs.udhcpd_pidfile with _ -> None in
   Opt.iter Unixext.kill_and_wait pid;
   let (_: string * string) = execute_command_get_output !Xapi_globs.busybox [ "udhcpd"; !Xapi_globs.udhcpd_conf ] in
-  ()
+  let start = Mtime_clock.elapsed_ns () in
+  let rec wait_for_pid n =
+    let now = Mtime_clock.elapsed_ns () in
+    if Int64.sub now start > Int64.mul 30L 1000000000L then failwith "Failed to start udhcpd";
```
edwintorok: You can use http://erratique.ch/software/mtime/doc/Mtime_clock.html#counters and compare spans

ocaml/xapi/xapi_udhcpd.ml
```diff
@@ -162,7 +162,18 @@ let restart_nolock () =
   let pid = try Unixext.pidfile_read !Xapi_globs.udhcpd_pidfile with _ -> None in
   Opt.iter Unixext.kill_and_wait pid;
   let (_: string * string) = execute_command_get_output !Xapi_globs.busybox [ "udhcpd"; !Xapi_globs.udhcpd_conf ] in
```
edwintorok: I wonder whether this should be a systemd service, that way we'd know we do not have multiple copies running, but this PR fixes the immediate problem so we can do that later.

ocaml/xapi/xapi_udhcpd.ml
lindig: I believe the Mtime module provides constants for this kind of conversion and also for measuring elapsed time. So this code could be made more idiomatic by using `Mtime.span` values. https://github.com/dbuenzli/mtime/blob/master/src/mtime_clock.mli

edwintorok:
Thanks for fixing this, if I understand correctly the reason it was broken before was that xapi was looking at the pidfile to find old udhcpds and kill them, however udhcpd takes a while to start up and write out a pidfile, so if we try to run this function multiple times in quick succession we end up spawning more udhcpd without realising. By waiting for the pidfile to be available we guarantee that if we get called again a pidfile will be found and we can kill the process properly.

ocaml/xapi/xapi_udhcpd.ml
```diff
@@ -162,7 +162,18 @@ let restart_nolock () =
...
-  ()
+  let start = Mtime_clock.elapsed_ns () in
+  let rec wait_for_pid n =
+    let now = Mtime_clock.elapsed_ns () in
+    if Int64.sub now start > Int64.mul 30L 1000000000L then failwith "Failed to start udhcpd";
+    let pid = try Unixext.pidfile_read !Xapi_globs.udhcpd_pidfile with _ -> None in
+    match pid with
+    | Some _ -> ()
+    | None ->
+      if n>10 then debug "Continuing to wait for the pidfile from udhcpd";
```
robhoes: Perhaps `if n > 0 && n mod 10 = 0` to avoid too much spam after 1s?

robhoes:
Please copy Edvin's comment into the commit message ;)

ocaml/xapi/xapi_udhcpd.ml
jonludlam: :+1:

ocaml/xapi/xapi_udhcpd.ml
jonludlam: I'd worry then about whether the bridge exists or not - perhaps @robhoes could comment?

robhoes: So you mean a systemd service that is not included in the host boot sequence, but still (re)started from this function? I think that would work.

jonludlam: Ah I see, yes that would make sense

ocaml/xapi/xapi_udhcpd.ml
```diff
@@ -162,7 +162,18 @@ let restart_nolock () =
   let pid = try Unixext.pidfile_read !Xapi_globs.udhcpd_pidfile with _ -> None in
   Opt.iter Unixext.kill_and_wait pid;
   let (_: string * string) = execute_command_get_output !Xapi_globs.busybox [ "udhcpd"; !Xapi_globs.udhcpd_conf ] in
-  ()
+  let start = Mtime_clock.counter () in
+  let rec wait_for_pid n =
+    let now = Mtime_clock.count start in
```
lindig: Is this correct? `now` is taken at every iteration of the loop. How is this guarantee to fail at some point when there is no upper limit for `n`?

PR #3662, opened 2018-07-12
CP-28844: Allow pool operator to perform clustering operations

edwintorok:
Other dangerous operations like enabling HA and Host destroy can be
performed by pool operator, so this one should be too.
Discussed with @robertbreker , the REQ was wrong here.

We should probably wait until the internal PR on the design is merged before merging this though.


edwintorok:
The design change got merged, so this should be good to go.

PR #3664, opened 2018-07-13
CA-289997: Fix for VDI.allowed_operations

jonludlam:
Fix a confusion between 'None' and 'Some []' leading to the field VDI.allowed_operations being computed without ever referencing any VBDs.

### DIFF
--- a/ocaml/xapi/xapi_vdi.ml
+++ b/ocaml/xapi/xapi_vdi.ml
@@ -283,7 +283,7 @@ let assert_operation_valid ~__context ~self ~(op:API.vdi_operations) =
     None -> ()
   | Some (a,b) -> raise (Api_errors.Server_error (a,b))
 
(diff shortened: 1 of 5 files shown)


ocaml/xapi/xapi_vdi.ml
```diff
@@ -307,7 +307,7 @@ let update_allowed_operations_internal ~__context ~self ~sr_records ~pbd_records
   in
   let all = Db.VDI.get_record_internal ~__context ~self in
   let allowed =
-    let check x = match check_operation_error ~__context ~sr_records ~pbd_records ~vbd_records ha_enabled all self x with None ->  [ x ] | _ -> [] in
+    let check x = match check_operation_error ~__context ~sr_records ~pbd_records ?vbd_records ha_enabled all self x with None ->  [ x ] | _ -> [] in
```
edwintorok: I see you pass the optional type through here, so we do not loose/overwrite the vbd records, but instead look it up from the DB.

PR #3666, opened 2018-07-19
CA-293399: Ensure we don't allow invalid utf8 strings into the db

jonludlam:


jonludlam:
I'd be mildly surprised if this is bad for performance. I vote we put it in and watch the rage stats.

PR #3667, opened 2018-07-20
CP-28116: reintroduce VM quicktests

gaborigloi:
I haven't reintroduced the other two VM quicktests that I've deleted, test_vhd_locking_hook, and vm_powercycle_test, because I think the former does not fail in case of product fault, and the latter is complex.


lindig:
Overall this looks good but I can't review this in detail. I would be happy to merge this and see how the tests perform.

PR #3668, opened 2018-07-20
CA-286723: Replace timebox function with something less dangerous.

jonludlam:


ocaml/xapi/helpers.ml
```diff
@@ -250,6 +250,26 @@ let make_rpc ~__context rpc : Rpc.response =
...
+      let resources = Locking_helpers.Thread_state.get_acquired_resources_by_task task_id in
+      List.iter Locking_helpers.kill_resource resources
+    in
+    Xapi_periodic_scheduler.add_to_queue (Ref.string_of task_id) Xapi_periodic_scheduler.OneShot timeout cancel;
+    let transport =
+      if Pool_role.is_master ()
+      then Unix(Xapi_globs.unix_domain_socket)
+      else SSL(SSL.make ~use_stunnel_cache:true ~task_id:(Ref.string_of task_id) (), Pool_role.get_master_address (), !Xapi_globs.https_port)
+    in
+    let result = XMLRPC_protocol.rpc ~srcstr:"xapi" ~dststr:"xapi" ~transport ~http rpc in
```
krizex: If the `timeout` is quite small, is it possible that the stunnel process has not been created when the `cancel` function been called, so the `Locking_helpers.kill_resource` won't take effect?

ocaml/xapi/xapi_xenops.ml
```diff
@@ -2349,8 +2349,19 @@ let events_from_xapi () =
                 while true do
                   let api_timeout = 60. in
                   let timeout = 30. +. api_timeout +. !Db_globs.master_connection_reset_timeout in
-                  let otherwise () = info "Event.from timed out in %f seconds, abort the events listening loop and retry" timeout; raise Exit in
-                  let from = Helpers.timebox ~timeout ~otherwise (fun () -> XenAPI.Event.from ~rpc ~session_id ~classes ~token:!token ~timeout:api_timeout |> event_from_of_rpc) in
+                  let timebox_rpc = Helpers.make_timeboxed_rpc ~__context timeout in
```
robhoes: This sits inside a `Helpers.call_api_functions` where we also make a connection, get an `rpc` function and do a login. We need the session, but it seems that we now have two connections, while we really need to cut down on slave->master connections.

ocaml/xapi/helpers.ml
```diff
@@ -250,6 +250,26 @@ let make_rpc ~__context rpc : Rpc.response =
     else SSL(SSL.make ~use_stunnel_cache:true (), Pool_role.get_master_address(), !Xapi_globs.https_port) in
   XMLRPC_protocol.rpc ~srcstr:"xapi" ~dststr:"xapi" ~transport ~http rpc
 
+let make_timeboxed_rpc ~__context timeout rpc : Rpc.response =
+  let subtask_of = Ref.string_of (Context.get_task_id __context) in
+  Server_helpers.exec_with_new_task "timeboxed_rpc" ~subtask_of:(Context.get_task_id __context) (fun __context ->
```
robhoes: We call this from `events_from_xapi`, which calls `Server_helpers.exec_with_new_task "xapi events"` followed by `Helpers.call_api_functions`, and now again `Server_helpers.exec_with_new_task`. This is at least confusing, perhaps worse...?

`make_rpc` ⬆️ does not seem to need it?

ocaml/xapi/helpers.ml
```diff
@@ -250,6 +250,26 @@ let make_rpc ~__context rpc : Rpc.response =
...
+    let open Xmlrpc_client in
+    let http = xmlrpc ~subtask_of ~version:"1.1" "/" in
+    let task_id = Context.get_task_id __context in
+    let cancel () =
```
robhoes: There is significant overlap with `make_rpc`. How about modifying that function instead and adding an optional `timeout` parameter?

robhoes: Perhaps `call_api_functions` ⬇️ can take the `timeout` param as well, and pass it to `make_rpc`, which it calls internally. But then we'd have to make it so that the returned `rpc` function sets the timer itself each time it is called, rather than the make-function doing this.

ocaml/tests/test_vdi_cbt.ml
```diff
@@ -379,11 +379,16 @@ let test_data_destroy =
...
         (* We need a 1-second tolerance here, otherwise the test will fail *)
         let timebox_timeout = timeout +. 1.0 in
-        Helpers.timebox
-          ~timeout:timebox_timeout
-          ~otherwise:(fun () -> Alcotest.fail (Printf.sprintf "data_destroy did not return in %f seconds" timebox_timeout))
-          (fun () -> Xapi_vdi._data_destroy ~__context ~self:vDI ~timeout)
-       in
+        let success = ref false in
+        let raisedexn = ref None in
+        ignore (bg (fun () -> (try Xapi_vdi._data_destroy ~__context ~self:vDI ~timeout with e -> raisedexn := Some e); success := true));
```
robhoes: `success` is a bit ambiguous here... it seems that it means "not timed out", but an exception could still have been raised, so the function-under-test did not succeed.

ocaml/xapi/xapi_xenops.ml
jonludlam: Not quite: we only actually make a connection _during_ an RPC call: until the `rpc` parameter is supplied to `make_rpc` it's only partially applied. So I don't believe we're creating more connections with this change.

ocaml/xapi/helpers.ml
jonludlam: Right, that's because I need to associate the 'resources' with the temporary task so I can kill the stunnel independently of any other resources associated with the task.

ocaml/xapi/helpers.ml
jonludlam: We could probably coalesce them a little. Also, as above, the make_rpc function is only partially applied until it gets the 'rpc' argument, so the timer is indeed not set until you're actually doing an RPC.

ocaml/tests/test_vdi_cbt.ml
jonludlam: Good point, will fix this.

ocaml/xapi/helpers.ml
jonludlam: You're quite right, it'd be a very similar situation to CA-285840, and it's about the only thing that would functionally be better with the original suggestion of using cohttp. However, I believe it's less bad in this case because we're making a connection _to_ the master rather than from as in CA-285840 (so 1/n as likely), and will only go wrong if we've lost our master (or route to the master), in which case either it comes back (and the stunnel connects with/without error), or we get an emergency_reset_master call in which case xapi restarts.

ocaml/xapi/xapi_xenops.ml
robhoes: Oh, I see! I didn't realise that this was partially applied.

ocaml/xapi/helpers.ml
robhoes: Aha, interesting. It would be nice to have some comments in the code about this (also about the partial application).

jonludlam: Comments added.

krizex:
I think this change puts us into a more risky situation.
Assuming the connection timeout happens, we will kill stunnel process in the `cancel` function, at the same time in https://github.com/xapi-project/xen-api-libs-transitional/blob/master/http-svr/xmlrpc_client.ml#L300 we will 1) kill this process agagin if do not need to reuse it, 2) put it into cache and kill it once get it from cache in the next time. 
Before this change, we face the fd race condition which may close a fd unexpectedly and it will only impact the xapi process, however after this change we face the killing process race condition which may kill a process unexpectedly, that will impact the system and is more dangerous.

PR #3669, opened 2018-07-23
CA-293399: Corrupted PCI information in xapi database

krizex:
The language binding to the C function is:
```
foreign "pci_lookup_name"
    (Pci_access.t @-> ptr char @-> int @-> int @-> int @-> returning string_opt)
```
The return value `string_opt` is constructed by the memory content of the second
parameter `ptr char`, however the corresponding OCaml value of `ptr char` is
GC-able (because we no longer reference it) before coping it to the return
value. So the return value may be incorrect.

Indeed this is bug of CTypes that it does not keep the OCaml variable alive
through the C binding invocation.

We make a local function call to foll the compiler to make the corresponding
OCaml variable alive through the C binding invocation.

### DIFF
--- a/ocaml/pci/lib/pci.ml
+++ b/ocaml/pci/lib/pci.ml
@@ -83,7 +83,11 @@ let with_string ?(size=1024) f =
    * to the lifetime of the input parameter, which can be moved
    * by the GC and cause problems. *)
   let buf = CArray.make char ~initial:'\x00' size in
-  f (CArray.start buf) size
+  let s = CArray.start buf in
+  let r = f s size in
+  (* Keep `s` alive through the C binding invocation in `f` *)
+  ignore (Sys.opaque_identity (List.hd [s]));
+  r
 
 let lookup_class_name pci_access class_id =
   with_string (fun buf size ->


edwintorok:
Making the code "complicated enough" so the compiler doesn't optimize it a certain way is fragile and can break on any compiler, surrounding code or compiler flag changes.

If you want to prevent optimizations you should use `Sys.opaque_identity`, which was designed for that purpose. I would expect something like the code below to work the same as this patch, but more future proof with compiler changes. If it doesn't then there is probably a deeper problem/bug in ctypes that should be investigated:
```ocaml
let s = CArray.start buf in
let r = f s size in
ignore (Sys.opaque_identity s); (* `r` can reference `s`, keep it alive through the foreign call. *)
r
```

mseri:
The code follows the recommendations from the CTypes FAQs: https://github.com/ocamllabs/ocaml-ctypes/wiki/FAQ#strings

I think an issue should also be opened on the ctypes issue tracker for this

krizex:
@mseri Exactly. Looks like the fix of https://github.com/ocamllabs/ocaml-ctypes/pull/473 is not completely.

lindig:
I agree with @edwintorok. The problem is to keep the value alive but the solution is too fragile. I did some experiments and found that keeping it alive is not always easy. I was not aware of `Sys.opaque_identity` and like the proposal. Another idea I had was:

```
let tag = ref 0 (* global *)

let s = CArray.start buf in
let r = f s size in
tag := Obj.repr s |> Obj.tag;
r
```

This also forces `s` to be kept alive by accessing it after the call to f. But it also relies on the fact that the compiler can't prove that the tag is not consumed and the update could be optimised away.

robhoes:
The C function that is being bound here is a little different from the one in the ctypes FAQ. I think that the main problem is that the C function takes a `char *` as one of its arguments, and may return the _same_ pointer (or NULL, or a string literal). The CArray object on the OCaml heap has a block with a pointer to the C array, and a finaliser to free the array when this block object is collected. A reference to this block is kept alive all the way into the C call as far as I can see. However, it is lost when the C call returns, and becomes at risk of being collected by the GC. The return value of the C call is a pointer to the same thing, but put in _another_ block on the heap, rather than reusing the original block to prevent the collection.

robhoes:
I tried `Sys.opaque_identity` on my test program. While it may prevent certain optimisation from being done, it looks like it does not help to keep the value reachable.

krizex:
@robhoes Yes, I met the same issue.

krizex:
@lindig I have tried
```
let s = CArray.start buf in
let r = f s size in
ignore (Sys.opaque_identity (Obj.repr s));
r
```
to keep `s` alive, however it does not work. I think the reason is `Obj.repr` is a stub function so may be optimized by the compiler. So I change the code to
```
  let s = CArray.start buf in
  let r = f s size in
  ignore (Sys.opaque_identity (List.hd [s]));
  r
```
Test ongoing.

lindig:
`Sys.opaque_identity` is unlikely to work. I believe to reliably keep the value alive we need to read from it and use the value, hence my suggestion to read the tag (because it is a cheap operation) and to store in a global reference. The compiler can't prove that this reference is not used.

krizex:
IMO `Sys.opaque_identity` works and it only works at compile time, so the code wrapped by it will be there in the binary. However there still have some runtime optimization, so if we write `Sys.opaque_identity (x)` while x is just a variable, it will be optimized out at runtime as it does not have side-effect. If we replace the `x`  with some function, it will be kept there and `Sys.opaque_identity` works.

I have tested the latest patch and it passed ~5200 loops while the failure test fails in no more than 100 loops.

PR #3670, opened 2018-07-23
CA-294281: Network reset: warn if networkd.db could not be deleted

robhoes:


robhoes:
Updated title.

PR #3675, opened 2018-07-27
UEFI: implement VM.start/stop

edwintorok:
Needs to be merged together with:
https://github.com/xapi-project/xcp-idl/pull/234
https://github.com/xapi-project/xen-api/pull/3675
https://github.com/xapi-project/xenopsd/pull/536
https://github.com/xapi-project/xenops-cli/pull/25

I've tested this on a box and was able to boot/shutdown Debian 9 and Windows 10 with UEFI.

I've opened this to get some feedback on the overall approach

### DIFF
--- a/ocaml/idl/datamodel_vm.ml
+++ b/ocaml/idl/datamodel_vm.ml
@@ -1331,7 +1331,7 @@ let set_HVM_boot_policy = call ~flags:[`Session]
            namespace ~name:"PV" ~contents:pv ();
            namespace ~name:"HVM" ~contents:hvm ();
            field ~ty:(Map(String, String)) "platform" "platform-specific configuration";
-           field ~lifecycle:[Prototyped, rel_next, ""] ~ty:(Map(String, String)) "NVRAM"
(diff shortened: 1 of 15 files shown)


ocaml/idl/datamodel_vm.ml
```diff
@@ -1331,6 +1331,9 @@ let set_HVM_boot_policy = call ~flags:[`Session]
            namespace ~name:"PV" ~contents:pv ();
            namespace ~name:"HVM" ~contents:hvm ();
            field ~ty:(Map(String, String)) "platform" "platform-specific configuration";
+           field ~lifecycle:[Prototyped, rel_naples, ""] ~ty:(Map(String, String)) "NVRAM"
```
robhoes: We normally put new fields at the end to avoid potential problems with compatibility if the constructor takes positional arguments.

Also, I believe we were going to make this read-only and add an explicit setter that is blocked if the power state is not halted?

edwintorok: I'll move it. The read-only+setter is going to be part of the next PR.

edwintorok:
Thanks, moved the datamodel field and updated based on xcp-idl changes.

PR #3676, opened 2018-07-30
Extend quicktest docstrings

gaborigloi:


edwintorok:
This looks good, but I think we should fix the dependency failure in xs-opam

gaborigloi:
https://github.com/xapi-project/xs-opam/pull/303 should fix this

gaborigloi:
Build passes :tada:

PR #3679, opened 2018-08-06
CP-28923: Add sm capability for large and thinly provisioned VDIs

thomasmck:

### DIFF
--- a/ocaml/xapi/smint.ml
+++ b/ocaml/xapi/smint.ml
@@ -39,6 +39,8 @@ type capability =
   | Vdi_attach_offline
   | Vdi_reset_on_boot
   | Vdi_configure_cbt
+  | Large_vdi
+  | Thin_provisioning
 
 type feature = capability * int64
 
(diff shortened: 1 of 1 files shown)


gaborigloi:
We can optionally add these as `storage_operations` enums, if these things may be useful for api users:
https://xapi-project.github.io/xen-api/classes/sr.html

thomasmck:
@gaborigloi If we add them as storage_operation enums does that mean when creating the SR from the cli they can specify that is has the "LARGE_VDIS" or "THIN_PROVISIONING" features?

gaborigloi:
It would be in the read-only `allowed-operations` SR field.
But actually, now that I think about it, these aren't really SR operations.

edwintorok:
If we are going to use the Thin_provisioning capability, then we should probably add it to the existing SRs as well (Local storage Ext, SMB, NFS, and filebased, GFS2, nfs-ng).

thomasmck:
@edvint Change for GFS2 is merged. Made the changes for existing SRs but waiting for Stefano to create a feature branch for me to merge to.

edwintorok:
thanks, my username here is edwintorok though, edvint is someone else unrelated to me.

PR #3681, opened 2018-08-15
CA-294874 xe-toolstack-restart: add message-switch

lindig:
Add message-switch to the list of services being restarted by 
xe-toolstack-restart.
    

lindig:
I agree that this script could be improved.

robhoes:
I believe that, when initially switching to systemd, we tried `systemctl restart` with the whole list of services, but that didn't seem to work well. I'm not sure if we tried first `stop` followed by `start`.

PR #3685, opened 2018-08-23
CP-29250 Removing oldval, and adding a unit test

rwdrich:
In the 'good old days', VM and VBD entries were linked to each other and when a VBD was created, the vm reference was entered. To fill in the entry in the VM record, a fake event was triggered to cause an update. If this vbd is then moved from one vm to another, the old value of the vm is needed to update the vbd record in the vm afterwards. Because the method for linking of the two records has changed, this is just a bit of tidying, with a unit test to ensure the behaviour is maintained.

### DIFF
--- a/ocaml/tests/test_event.ml
+++ b/ocaml/tests/test_event.ml
@@ -57,6 +57,32 @@ let test_event_from_ev_rel () =
   Alcotest.(check int) "list length" 1 (List.length vm_ev);
   Alcotest.(check int) "list length" 1 (List.length vbd_ev)
 
+let test_events_from_obj () =
+  (* Test that events on one object create the correct number of events on other objects *)
+  let __context, session_id = event_setup_common () in
+  let vma = make_vm __context () in
+  let evs = Xapi_event.from __context ["vm"] "" 30.0 |> parse_event_from in
+  let tok = evs.token in
+  let vm_ev = List.filter (fun ev -> ev.ty="vm") evs.events in
+  Alcotest.(check int) "Creation of dom0 and test vma" 2 (List.length vm_ev);
+
+  let vmb = make_vm __context () in
+  let evs2 = Xapi_event.from __context ["vm"] tok 30.0 |> parse_event_from in
+  let tok2 = evs2.token in
+  let vm_ev2 = List.filter (fun ev -> ev.ty="vm") evs2.events in
+  Alcotest.(check int) "Creation of test vmb" 1 (List.length vm_ev2);
+
+  let vbd = make_vbd ~__context ~vM:vma () in
+  let evs3 = Xapi_event.from __context ["vm"] tok2 30.0 |> parse_event_from in
+  let tok3 = evs3.token in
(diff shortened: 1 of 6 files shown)


ocaml/tests/test_event.ml
```diff
@@ -57,6 +57,36 @@ let test_event_from_ev_rel () =
   Alcotest.(check int) "list length" 1 (List.length vm_ev);
   Alcotest.(check int) "list length" 1 (List.length vbd_ev)
 
+let test_events_from_obj () =
+  (* Test that events on one object create the correct number of events on other objects *)
+  let __context, session_id = event_setup_common () in
+  let evs = Xapi_event.from __context ["vm"] "" 30.0 |> parse_event_from in
+  let vma = make_vm __context () in
```
jonludlam: Aren't these in the wrong order? Don't you want to do a thing, _then_ get the events?

ocaml/xapi/eventgen.ml
```diff
@@ -101,46 +97,17 @@ let database_callback event db =
...
-  | WriteField (tblname, objref, fldname, oldval, newval) ->
-    let events_old_val =
-      if is_valid_ref oldval then
-        let oldval = Schema.Value.Unsafe_cast.string oldval in
-        events_of_other_tbl_refs
-          (List.map (fun (tbl,fld) ->
-               (tbl, oldval, find_get_record tbl ~__context:context ~self:oldval)) (other_tbl_refs_for_this_field tblname fldname))
-      else [] in
-    let events_new_val =
-      if is_valid_ref newval then
```
jonludlam: You can remove the similar code in the Create and Delete clauses.

ocaml/tests/test_event.ml
```diff
@@ -57,6 +57,36 @@ let test_event_from_ev_rel () =
...
+  let evs3 = Xapi_event.from __context ["vm"] tok2 30.0 |> parse_event_from in
+  let vbd = make_vbd ~__context ~vM:vma () in
+  let tok3 = evs3.token in
+  let vm_ev3 = List.filter (fun ev -> ev.ty="vm") evs3.events in
+  Alcotest.(check int) "list length" 1 (List.length vm_ev3);
+  let vbd_ev = List.filter (fun ev -> ev.ty="vbd") evs3.events in
+  Alcotest.(check int) "list length" 0 (List.length vbd_ev);
+
+  let evs4 = Xapi_event.from __context ["vm"] tok3 30.0 |> parse_event_from in
+  Db.VBD.set_VM __context vbd vmb;
```
jonludlam: The only thing you've changed in the events code in this PR is what happens after a WriteField event, and you're not actually invoking the events logic after this, which is the only thing that can cause a WriteField event.

lindig:
Can we either progress this or close it?

lindig:
@jonludlam is this good for a merge?

jonludlam:
I need to go over this some more.

lindig:
I'm closing this for now.

PR #3690, opened 2018-09-03
CP-29238: Merging from *Feature/cc/ccm* for CCM

makunterry:


lindig:
The code change is quite small. Is there a good reason to not put all of this into a single commit? This looks clean but I am not familiar with the technical details of the ciphers so can't really comment on them.

cheng-z:
As the time for master merge close and all test been passed. We will merge the code to master if no other concern. Please shout if any comments.

krizex:
I don't have much knowledge about this story, it's better to have @jonludlam review and give some feedback.

lindig:
Please combine this into a single commit with one a comprehensive commit message. This is all related and I see not good reason to split this up.

jonludlam:
I think it'll be useful to keep a pointer to the CA/CP tickets involved though - I'd like this as code comments too (e.g. when specifying the key length of 2048 mention CA-296204)

ocaml/xapi/xapi_globs.ml
```diff
@@ -832,7 +832,7 @@ let xenopsd_queues = ref ([
 
 let default_xenopsd = ref "org.xen.xapi.xenops.xenlight"
 
-let ciphersuites_good_outbound = ref "!EXPORT:RSA+AES128-SHA256"
+let ciphersuites_good_outbound = ref "!EXPORT:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-SHA384:RSA+AES128-SHA256:AES128-SHA:AES256-SHA256"
```
edwintorok: Is the order here correct? I think AES256 should be preferred ahead of AES128 everywhere.
The requirements list this order, where AES_256 precedes all of AES_128:
TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384, TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384, TLS_RSA_WITH_AES_256_CBC_SHA256, TLS_RSA_WITH_AES_128_CBC_SHA, TLS_RSA_WITH_AES_128_CBC_SHA256
(although I think preferring SHA256 over SHA is the right thing as done in this commit, so the only thing that might have to be changed here is to move the last AES256-SHA256 earlier.

scripts/generate_ssl_cert
```diff
@@ -52,7 +52,7 @@ do
 	echo "DNS.${i}=${x}" >> config
 done
 
-openssl genrsa 1024 > privkey.rsa
+openssl genrsa 2048 > privkey.rsa
```
edwintorok: 2048 is good, but the requirements mention 3072. Which one is correct?

scripts/init.d-xapissl
```diff
@@ -75,7 +75,7 @@ writeconffile () {
 
     # (This "good" list must match, or at least contain one of,
     #  the ciphersuites-good-outbound list in /etc/xapi.conf.)
-    GOOD_CIPHERS='RSA+AES128-SHA256'
+    GOOD_CIPHERS='ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-SHA384:RSA+AES128-SHA256:AES128-SHA:AES256-SHA256'
```
edwintorok: Same comment on cipher order as above.

scripts/generate_ssl_cert
jonludlam: This is why I mentioned the comments - CA-296204 details why 3072 failed - it involves a static length buffer.

cheng-z: Yes. We are now confirmed to use 2048. I will change the requirement.

cheng-z:
Thank you all for your comment.
As @makunterry is now on paternity leave. Please @xiewei20082008 rework the comment and create another PR and link to this one.
Also please close this PR and use the new PR to do code merge.

xiewei20082008:
Close this PR please, I raised a new PR here [https://github.com/xapi-project/xen-api/pull/3692/].

cheng-z:
Close it as https://github.com/xapi-project/xen-api/pull/3692 will track on the code change.

PR #3691, opened 2018-09-05
CA-293678 SXM in partially upgraded pool with pre-7.3 hosts fails mig…

YarsinCitrix:
…ration checks

assert_can_migrate_sender is added in 7.3 along with some code reoganization.
So if in any chance this is called from master to slave in RPC but slave is
before 7.3, the slave will raise exception and hance fail the entire code due
to this unknown method.

In assert_can_migrate_sender, it is mostly regarding pgpu check, which may not
be necessary in older release, so just ignore this unknown method.

Signed-off-by: YarsinCitrix <yarsin.he@citrix.com>

### DIFF
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -1684,7 +1684,7 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
           (fun session_id rpc -> Client.VM.assert_can_migrate_sender rpc session_id vm dest live vdi_map vif_map vgpu_map options)
       with
       | Api_errors.Server_error(code, params) when code=Api_errors.message_method_unknown ->
-        debug "VM.assert_can_migrate_sender not known by destine, assuming older version host no need to do this."
+        warn "VM.assert_can_migrate_sender is not known by destination, assuming it can ignore this check."
 
     let assert_can_migrate ~__context ~vm ~dest ~live ~vdi_map ~vif_map ~options ~vgpu_map =
       info "VM.assert_can_migrate: VM = '%s'" (vm_uuid ~__context vm);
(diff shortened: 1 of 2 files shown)


ocaml/xapi/message_forwarding.ml
```diff
@@ -1679,8 +1679,12 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
       info "VM.assert_can_migrate_sender: VM = '%s'" (vm_uuid ~__context vm);
       let local_fn = Local.VM.assert_can_migrate_sender ~vm ~dest ~live ~vdi_map ~vif_map ~vgpu_map ~options in
-      forward_vm_op ~local_fn ~__context ~vm
-        (fun session_id rpc -> Client.VM.assert_can_migrate_sender rpc session_id vm dest live vdi_map vif_map vgpu_map options)
+      try
+        forward_vm_op ~local_fn ~__context ~vm
+          (fun session_id rpc -> Client.VM.assert_can_migrate_sender rpc session_id vm dest live vdi_map vif_map vgpu_map options)
+      with
+      | Api_errors.Server_error(code, params) when code=Api_errors.message_method_unknown ->
+        debug "VM.assert_can_migrate_sender not known by destine, assuming older version host no need to do this."
```
krizex: I think the log level should at least be `warning` and it's better to replace the `params` with `_` in the pattern match statement.

YarsinCitrix: No, the current "cannot marshall arguments for..." error is only debug, so warning is not necessary. Just keep the two similar handling the same level.

YarsinCitrix: Just followed the same style for the (code, params) from somewhere else, not used params in this style is also seen from various codes, so not infringing either.

krizex: 1. The log level of this kind of log is something we don't expect, so it's better to log all of them with `warning` level.
2. I prefer using `_` instead named variable for useless variable because it makes the code clean and easy to read (I need to remember less variables when reading the code...)

lindig: Grammar: "VM.assert_can_migrate_sender not known by **destination, assuming it can ignore this check**". I agree that this should be at priority warning.

lindig:
Ping @YarsinCitrix.

robhoes:
Looks like this got merged, while the review was not yet complete.......?

YarsinCitrix:
All review comments have been address. As a simple change, I just merged it. Just want to resolve this also avoiding efforts of others. So the review can only be 'complete'ed by reviewers?

robhoes:
As a general rule, we always want every change to be reviewed by someone else, even the smallest ones. It's just a safety net, because it is always possible to make a mistake. It might sometimes seem unnecessary, but once you start making exceptions, the rules get weaker and people tend to get sloppy. And for small changes the additional review effort is minimal as well. So we prefer to keep it simple and always have everything reviewed.

YarsinCitrix:
OK, will wait more time for final confirmation next time. Thanks for the reminder!

PR #3696, opened 2018-09-10
CA-290696: Update task to un-cancellable after xenopsd notify xapi

krizex:
Set the task to un-cancellable after xenopsd updated it to un-cancellable

### DIFF
--- a/ocaml/xapi/xapi_xenops.ml
+++ b/ocaml/xapi/xapi_xenops.ml
@@ -2062,8 +2062,8 @@ let update_task ~__context queue_name id =
       if not task_t.Task.cancellable then begin
         let allowed_operations = Db.Task.get_allowed_operations ~__context ~self in
         if List.mem `cancel allowed_operations then begin
-          let allowed_operations' = List.filter (fun x -> x != `cancel) allowed_operations in
-          debug "Set task %s to not cancellable." (Ref.string_of self);
+          let allowed_operations' = List.filter (fun x -> x <> `cancel) allowed_operations in
+          debug "Set task %s to not cancellable." (Ref.really_pretty_and_small self);
           Db.Task.set_allowed_operations ~__context ~self ~value:allowed_operations'
         end
       end
--- a/ocaml/xapi/xapi_xenops.ml
+++ b/ocaml/xapi/xapi_xenops.ml
@@ -2058,7 +2058,15 @@ let update_task ~__context queue_name id =
     let task_t = Client.TASK.stat dbg id in
     match task_t.Task.state with
     | Task.Pending x ->
-      Db.Task.set_progress ~__context ~self ~value:x
+      Db.Task.set_progress ~__context ~self ~value:x;
+      if not task_t.Task.cancellable then begin
+        let allowed_operations = Db.Task.get_allowed_operations ~__context ~self in
+        if List.mem `cancel allowed_operations then begin
+          let allowed_operations' = List.filter (fun x -> x != `cancel) allowed_operations in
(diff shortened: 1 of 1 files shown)


krizex:
This PR depends on https://github.com/xapi-project/xcp-idl/pull/240 and https://github.com/xapi-project/xenopsd/pull/550

lindig:
Please update the commit message such that the topic does not overflow:
```
CA-290696 - block VM migration cancelation after VM suspended
```

The logic looks sounds to me.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -2058,7 +2058,15 @@ let update_task ~__context queue_name id =
...
     match task_t.Task.state with
     | Task.Pending x ->
-      Db.Task.set_progress ~__context ~self ~value:x
+      Db.Task.set_progress ~__context ~self ~value:x;
+      if not task_t.Task.cancellable then begin
+        let allowed_operations = Db.Task.get_allowed_operations ~__context ~self in
+        if List.mem `cancel allowed_operations then begin
+          let allowed_operations' = List.filter (fun x -> x != `cancel) allowed_operations in
+          debug "Set task %s to not cancellable." (Ref.string_of self);
+          Db.Task.set_allowed_operations ~__context ~self ~value:allowed_operations'
```
robhoes: What about `TaskHelper.set_not_cancellable ~__context`?

krizex: `TaskHelper.set_not_cancellable ~__context` will set the task of the `__context` to not cancellable, but in this case, the task we have to update status not in `__context`

ocaml/xapi/xapi_xenops.ml
```diff
@@ -2058,7 +2058,15 @@ let update_task ~__context queue_name id =
     let task_t = Client.TASK.stat dbg id in
     match task_t.Task.state with
     | Task.Pending x ->
-      Db.Task.set_progress ~__context ~self ~value:x
+      Db.Task.set_progress ~__context ~self ~value:x;
+      if not task_t.Task.cancellable then begin
+        let allowed_operations = Db.Task.get_allowed_operations ~__context ~self in
+        if List.mem `cancel allowed_operations then begin
+          let allowed_operations' = List.filter (fun x -> x != `cancel) allowed_operations in
```
robhoes: You should use `<>` rather than `!=`.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -2058,7 +2058,15 @@ let update_task ~__context queue_name id =
...
+          debug "Set task %s to not cancellable." (Ref.string_of self);
```
robhoes: Here you can use `Ref.really_pretty_and_small` to get a task ref as in the logs (e.g. `R:xxxxxxxx`), which makes it easier to follow.

ocaml/xapi/xapi_xenops.ml
krizex: Oh, my bad.

ocaml/xapi/xapi_xenops.ml
krizex: Does this function only suitable for task ref?

robhoes: It works on any ref, but we seem to use it just for tasks. Also, the prefixes `R` (a "real" ref for something in the DB) and `D` (a dummy ref; nothing in the DB) are only relevant for tasks.

robhoes:
This is good, but can't be merged until the xcp-idl patch has been rebased and merged.

PR #3697, opened 2018-09-10
CP-29389: Delete duplicate function 'parse_device_config'

minli1:
The code shows the functionality of `parse_device_config` is the same as `read_map_params` + key_name , and also the latter one supports `device-config-key` syntax.
So there is no need to keep the duplicate code.

### DIFF
--- a/ocaml/xapi/cli_operations.ml
+++ b/ocaml/xapi/cli_operations.ml
@@ -87,7 +87,9 @@ let waiter printer rpc session_id params task =
        Client.Task.destroy rpc session_id task
     )
 
-(* Return the list of k=v pairs for maps *)
+(* Return the list of k=v pairs for maps.
+   Works for key which is not follow by a ':',
+   also match old syntax 'device-config-key' for backwards compatability *)
 let read_map_params name params =
   let len = String.length name + 1 in (* include ':' *)
   let filter_params = List.filter (fun (p,_) -> (String.startswith name p) && (String.length p > len)) params in
@@ -1419,13 +1421,6 @@ let parse_host_uuid ?(default_master=true) rpc session_id params =
     else failwith "Required parameter not found: host-uuid"
   end
 
-let parse_device_config params =
-  (* Ack! We're supposed to use the format device-config:key=value but we need to match device-config-key=value for *)
-  (* backwards compatability *)
-  let len = String.length "device-config:" in
-  let filter_params = List.filter (fun (p,_) -> (String.startswith "device-config" p) && (String.length p > len)) params in
-  List.map (fun (k,v) -> String.sub k len (String.length k - len),v) filter_params
-
 (* SR create destroy list param-list param-get param-set param-add param-remove *)
 
 let sr_create fd printer rpc session_id params =
@@ -1438,7 +1433,7 @@ let sr_create fd printer rpc session_id params =
     with _ -> 0L in
   let _type=List.assoc "type" params in
   let content_type = List.assoc_default "content-type" params "" in
-  let device_config = parse_device_config params in
+  let device_config = read_map_params "device-config" params in
   (* If the device-config parameter is of the form k-filename=v, then we assume the
      	   key is 'k' and the value is stored in a file named 'v' *)
   let suffix = "-filename" in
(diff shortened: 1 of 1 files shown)


minli1:
Job 2251370 and 2251369 passed.

minli1:
Hi, @lindig 
1. We don't have releases that are not EOL to support `device-config-key` key for sr-create/sr-probe
2. From the code, you can find [here](https://github.com/xapi-project/xen-api/pull/3697/files#diff-d08f7885870f1c8bd4c6dbb283e4322aL1427) that it will parse the tuple by sub the string with length of `device-config`, just think that if the user have `device-config-key:target=xxx` in the params, it will have have `key:target=xxx` at last. And SM layer will fails with `not correct params`.
3. The comment is wrong, it's actually not deal with `device-config-key` and it's not needed for now.
3. Above all, I had small test by sr-create with `device-config` and `device-config-key`.

**device-config**
`xe sr-create type=lvmoiscsi name-label=new_iscsi device-config:target=<ip_address> device-config:SCSIid=<scsiid> device-config:targetIQN=<targetIQN> shared=true`
9eb04e46-3e8f-08a1-e5cb-b53b5430fa30
**device-config-key**
`xe sr-create type=lvmoiscsi name-label=new_iscsi device-config-key:target=<ip_address> device-config-key:SCSIid=<scsiid> device-config-key:targetIQN=<targetIQN> shared=true`
Error code: SR_BACKEND_FAILURE_95
Error parameters: , The request is missing the target parameter,

minli1:
Commit message updated. @lindig  Thanks.

gaborigloi:
I think the comment meant that we should substitute `key` in `device-config-key=value` for the actual name of the key. I think this should work:
```
xe sr-create type=lvmoiscsi name-label=new_iscsi device-config-target=<ip_address> device-config-SCSIid=<scsiid> device-config-targetIQN=<targetIQN> shared=true
```
But if all the releases that use this format are EOL, we should definitely remove it.
So I think the commit message should be updated to only mention this reason (1.)

minli1:
Thanks! Gabor. @gaborigloi 
You are right, `device-config-target` works fine after quick test just now.
But seems the function '[parse_device_config](https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/cli_operations.ml#L91)' and '[read_map_params](https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/cli_operations.ml#L1422)' with a name(device-config) are same.
What do you think? Thanks.

minli1:
Commit message updated.

gaborigloi:
It looks good to me. Good to merge if we definitely don't need to support the old syntax.
The question is whether some people still have old scripts that use the old syntax and that need to keep working. I'm not sure about that. Is is okay to drop support for this syntax @robhoes @jonludlam ?

edwintorok:
I'm not familiar with the history of device-config, but looking through XenRT it looks like it is using the following deprecated formats:
```
                       "device-config-location=\"\\$(touch /tmp/CA-10038)\" "
            args.append("device-config-chappassword=%s" % (p))
            args.append("device-config-chapuser=%s" % (u))
            args.append("device-config-incoming_chappassword=%s" % (p))
            args.append("device-config-incoming_chapuser=%s" % (u))
            args.append("device-config-location=%s/%s" %
            args.append("device-config-location=%s/SRs/%s/sr" %
        args.append("device-config-LUNid=%u" % (lun.getLunID()))
        args.append("device-config-location=\"//%s/%s\"" %
        args.append("device-config-target=%s" % (lun.getServer()))
        args.append("device-config-targetIQN=%s" % (lun.getTargetName()))
```
If you want to drop support for this format please update XenRT to not use it anymore.

minli1:
Hi, @gaborigloi @edwintorok 
Thanks for the review! Actually I am not dropping support for `device-config-key`, it's just because of the `parse_device_config` and `read_map_params` + <device-config> are the same.
So even change to the second function, the old syntax is also working, I've tested on them. Please see the above comment. https://github.com/xapi-project/xen-api/pull/3697#issuecomment-420662326

minli1:
Updated commit message and comment. @gaborigloi 
Ask Rob to have a look at it, thanks. @robhoes

PR #3698, opened 2018-09-11
Update opam files

lindig:
The opam files have been updated using "dune external-lib-deps" to find the real dependencies.


xapi-database.opam
```diff
@@ -9,14 +9,24 @@ build: [[ "jbuilder" "build" "-p" name ]]
...
+  "ppx_deriving_rpc"
   "ppx_sexp_conv"
-  "xapi-libs-transitional"
-  "xapi-idl"
+  "rpclib"
+  "sexpr"
+  "base-threads"
+  "uuid"
+  "xapi-consts"
+  "xapi-datamodel"
```
edwintorok: This causes a loop and Travis complains:
```
The actions to process have cyclic dependencies:
  - install xapi-database.master -> install xapi-datamodel.master -> install xapi-database.master
  - install xapi-datamodel.master -> install xapi-consts.master -> install xapi-datamodel.master
  - install xapi-database.master -> install xapi-datamodel.master -> install xapi-consts.master -> install xapi-database.master
```

xapi.opam
```diff
@@ -10,38 +10,52 @@ build-test: [[ "jbuilder" "runtest" "-p" name "-j" jobs ]]
...
+  "xapi-stdext-unix"
   "xapi-tapctl"
   "xapi-test-utils"
   "xapi-types"
   "xapi-xenopsd"
+  "xcp"
+  "xcp-inventory"
   "xenctrl"
-  "xenstore"
+  "xml-light2"
```
edwintorok: This seems to come from perftest, but this is what the dune file says, so updating it in opam is correct, we can look at pruning the dependencies from dune later if needed.

xapi-database.opam
```diff
@@ -9,14 +9,22 @@ build: [[ "jbuilder" "build" "-p" name ]]
 
 depends: [
   "jbuilder" {build & >= "1.0+beta11"}
-  "rpc"
+
+  "gzip"
+  "http-svr"
+  "ocaml-migrate-parsetree"
```
mseri: Omp appears in multiple packages but it should not, maybe something is wrong in some dune/jbuilder file

edwintorok: We should keep an eye out when updating the other packages, eventually we should find the one responsible for it.

PR #3699, opened 2018-09-12
Refine the format of generated db_actions code

krizex:


krizex:
Diffs of the generated file: [diff.txt](https://github.com/xapi-project/xen-api/files/2373968/diff.txt)

robhoes:
Hmm... is this useful?

krizex:
Yes, it helps me when reading these generated code

lindig:
I have no strong opinion. Code that emits code is difficult to write and I don't mind one way or the other.

PR #3701, opened 2018-09-13
CA-264499: xapi fails when receiving unknown enum from newer release

krizex:
Xapi will fail if receive new introduced enumeration value which cannot
recognize by code.

If the unknown enumeration is just a plain value that not wrapped by other
container, we won't handle and leave it fails as is.
If the unknown enumeration warpped by `set` or `map`, we will filter out the
enumeration which we cannot recognize in old version code.


lindig:
Closing this. Please re-open it for more discussion.

PR #3707, opened 2018-09-18
CA-298373: Set ssl_legacy=true by default in Havana

xiewei20082008:
Job passed - id 2258127.

### DIFF
--- a/ocaml/xapi/create_misc.ml
+++ b/ocaml/xapi/create_misc.ml
@@ -156,7 +156,7 @@ let read_localhost_info () =
     dom0_static_max = dom0_static_max;
     ssl_legacy = try (
       bool_of_string (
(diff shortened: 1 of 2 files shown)


cheng-z:
Small change. LGTM.

cheng-z:
@lindig  can I ask your help to have a look also.
In havana, when backport master change, we need to do so to keep backward keep the same.

cheng-z:
Hi @lindig 
It is not a backport, it is for havana only. Changed the commit message now.
Merge the code now as the code seems all good.

xiewei20082008:
> This needs a better commit message. If this is a backport, what is the commit that is backported? If this required for some other backport, please explain it in the commit message.

This is not a backport - it is just a CA for Havana.

PR #3709, opened 2018-09-18
CP-29420: finer grained locking for xenopsd metadata

edwintorok:
Xenopsd has a global metadata lock for creating/deleting and updating the VM cache table.
This means that if you have a VM start that triggers a xenopsd metadata push then all events from other VMs get blocked until this metadata push finishes. On a slave this can be especially slow because it needs to wait for the RPC from master.

This change improves bootstorm on slaves.

The PR is not specific to UEFI, so after some more testing on the UEFI branch I intend to also submit this to master.

### DIFF
--- a/ocaml/xapi/xapi_xenops.ml
+++ b/ocaml/xapi/xapi_xenops.ml
@@ -19,7 +19,6 @@ open Network
 
 open Stdext
 open Xstringext
-open Listext
 open Threadext
 open Pervasiveext
 open Fun
@@ -202,7 +201,7 @@ let rtc_timeoffset_of_vm ~__context (vm, vm_t) vbds =
     vdis
     |> List.map (fun self -> (self, Db.VDI.get_record ~__context ~self))
     |> List.filter (fun (_, record) -> record.API.vDI_on_boot = `reset)
-    |> List.filter_map (fun (reference, record) ->
+    |> Listext.List.filter_map (fun (reference, record) ->
         Opt.of_exception (fun () ->
             reference,
             List.assoc Vm_platform.timeoffset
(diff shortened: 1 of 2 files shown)


ocaml/xapi/xapi_xenops.ml
```diff
@@ -1039,29 +1039,41 @@ module Xapi_cache = struct
...
 
   let cache = Hashtbl.create 10 (* indexed by Vm.id *)
+  let mutex = Mutex.create ()
+  let with_lock f =
+    Mutex.execute mutex f
+
+  let register id initial_value =
+    with_lock (fun () ->
+        debug "xapi_cache: creating cache for %s" id;
+        if not(Hashtbl.mem cache id) || (Hashtbl.find cache id = None)
```
lindig:
```
if (try Hashtbl.find cache id = None with Not_found -> true) then
```

Probing the hash table twice is an anti pattern. Because of the lock we don't get a race here but I think it should be still avoided.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -1039,29 +1039,41 @@ module Xapi_cache = struct
...
+        then Hashtbl.replace cache id initial_value)
+
+  let unregister id =
+    with_lock (fun () ->
+      debug "xapi_cache: deleting cache for %s" id;
+      Hashtbl.remove cache id)
+
+  let update id f =
+    let find_nolock id =
+      if Hashtbl.mem cache id
```
lindig: Same as above. If you don't like exceptions, you can use `Hashtbl.find_opt`

edwintorok: Good point, when the original code was written find_opt did not exist, but we can use it now. I'll update.

edwintorok:
I've made some changes to use find_opt, in certain places it looks a bit odd, because we store an `'a option` inside the Hashtbl, so we get `Some None` at some point. I tried to refactor the code to use the `'a` directly, but I think the distinction is important for `add_caches`, where we just register an `id` with an "empty" cached value vs not having the id in the hashtbl at all (observable by calling `list`).
If you have suggestions on how to improve this I'd gladly do it.

Restructured the `update` function, I don't think having that `f` separate from the lookup code helped, I've instead folded it into the update function itself (it takes a `newvalue` and tells the caller whether it performed the update or not), and added a comment.

PR #3710, opened 2018-09-21
CP-29169: call varstore-rm on UEFI VM clone

edwintorok:
On clone certain UEFI variables may need to be cleared.
Call `varstore-rm -c <uuid>` to perform this.

The actual variables to clear are defined in the `/etc/xapi.d/efi-clone` directory,
however the format of NVRAM is managed by `varstored`, hence XAPI can't
clear them itself and needs to invoke the helper executable.

There is a related specfile change to make this directory owned by XAPI, however this builds even with the old specfile.


lindig:
Can we merge this? I looked over the code and it looks fine.

PR #3711, opened 2018-09-24
Updates to update precheck extensions

robhoes:
This makes the extensions the same as on master.


robhoes:
CC: @srowe

robhoes:
Travis says:
```
ocamlfind: Package `ppx_sexp_conv' not found
```
That's not because of this PR.

PR #3713, opened 2018-09-25
XSO-886: Do not add micro version to the pv_drivers_version if the latter is empty.

kc284:
I'm suggesting this change to fix the above issue whereby querying a VM's PV-drivers-version returns on occasion (for example, when it's a new VM without PV drivers) the following:
```
[root@xxxx ~]# xe vm-param-get uuid=2ff4edfd-3477-6507-63fd-e0055a578912 param-name=PV-drivers-version
micro: -1
```
This looks weird because major/minor are not there. The original code adds a micro version to pv_drivers_version if it's missing, but I think this should only happen if pv_drivers_version is not empty.

### DIFF
--- a/ocaml/xapi/xapi_guest_agent.ml
+++ b/ocaml/xapi/xapi_guest_agent.ml
@@ -233,7 +233,7 @@ let get_initial_guest_metrics (lookup: string -> string option) (list: string ->
   (* to avoid breakage whilst 'micro' is added to linux and windows agents, default this field
      to -1 if it's not present in xenstore *)
   let pv_drivers_version =
-    if List.mem_assoc "micro" pv_drivers_version then pv_drivers_version (* already there; do nothing *)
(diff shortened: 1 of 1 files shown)


lindig:
Travis is somewhat broken: the build based on xenserver-buildenv failed. We might want to move to an `xs-opam` based build only.

ocaml/xapi/xapi_guest_agent.ml
```diff
@@ -233,7 +233,7 @@ let get_initial_guest_metrics (lookup: string -> string option) (list: string ->
   (* to avoid breakage whilst 'micro' is added to linux and windows agents, default this field
      to -1 if it's not present in xenstore *)
   let pv_drivers_version =
-    if List.mem_assoc "micro" pv_drivers_version then pv_drivers_version (* already there; do nothing *)
+    if List.length pv_drivers_version = 0 || List.mem_assoc "micro" pv_drivers_version then pv_drivers_version (* already there; do nothing *)
```
lindig: Minor point: `if pv_drivers_version = []` would be better to check for an empty list.

PR #3715, opened 2018-09-25
CA-298318: Set the SM name_label correctly

jonludlam:


edwintorok:
nice spot @jonludlam . IIUC here `query_result.name` comes from the ppx generated `query_result`, which contains a `name` field. Would it be possible to change the PPX to generate field names that are namespaces and wouldn't clash with something a user would normally write, or would that mean we have to also update manually written code? (e.g. `mirage_ocaml_rpc_rpcty_name`)

PR #3716, opened 2018-09-26
CP-29521: VDI import/export in TAR format

gaborigloi:

### DIFF
--- a/ocaml/xapi/export_raw_vdi.ml
+++ b/ocaml/xapi/export_raw_vdi.ml
@@ -36,12 +36,6 @@ let localhost_handler rpc session_id vdi (req: Http.Request.t) (s: Unix.file_des
          debug "export_raw_vdi task_id = %s; vdi = %s; format = %s; content-type = %s; filename = %s"
            (Ref.string_of task_id) (Ref.string_of vdi) (Importexport.Format.to_string format) content_type filename;
          let copy base_path path =
-           let headers = Http.http_200_ok ~keep_alive:false () @ [
-               Http.Hdr.task_id ^ ":" ^ (Ref.string_of task_id);
-               Http.Hdr.content_type ^ ":" ^ content_type;
-               Http.Hdr.content_disposition ^ ": attachment; filename=\"" ^ filename ^ "\""
-             ] in
-           Http_svr.headers s headers;
            try
              debug "Copying VDI contents...";
              Vhd_tool_wrapper.send ?relative_to:base_path (Vhd_tool_wrapper.update_task_progress __context)
@@ -51,14 +45,29 @@ let localhost_handler rpc session_id vdi (req: Http.Request.t) (s: Unix.file_des
              raise (Api_errors.Server_error (Api_errors.vdi_io_error, ["Device I/O errors"])) in
          begin
            try
-             Sm_fs_ops.with_block_attached_device __context rpc session_id vdi `RO
-               (fun path ->
-                  match Importexport.base_vdi_of_req ~__context req with
-                  | Some base_vdi ->
-                    Sm_fs_ops.with_block_attached_device __context rpc session_id base_vdi `RO
-                      (fun base_path -> copy (Some base_path) path)
-                  | None -> copy None path
-               )
+             let headers = Http.http_200_ok ~keep_alive:false () @ [
+                 Http.Hdr.task_id ^ ":" ^ (Ref.string_of task_id);
+                 Http.Hdr.content_type ^ ":" ^ content_type;
+                 Http.Hdr.content_disposition ^ ": attachment; filename=\"" ^ filename ^ "\""
+               ] in
+             Http_svr.headers s headers;
+             match format with
+             | Raw | Vhd ->
+               Sm_fs_ops.with_block_attached_device __context rpc session_id vdi `RO
+                 (fun path ->
+                    match Importexport.base_vdi_of_req ~__context req with
+                    | Some base_vdi ->
+                      Sm_fs_ops.with_block_attached_device __context rpc session_id base_vdi `RO
+                        (fun base_path -> copy (Some base_path) path)
+                    | None -> copy None path
+                 )
+             | Tar ->
+               (* We need to keep refreshing the session to avoid session timeout *)
+               let refresh_session = Xapi_session.consider_touching_session rpc session_id in
+               (* The third element of the tuple, the size, is currently ignored by the function *)
+               let size = Client.Client.VDI.get_virtual_size ~rpc ~session_id ~self:vdi in
+               Stream_vdi.send_all refresh_session s __context rpc session_id [(Xapi_globs.vdi_tar_export_dir, vdi, size)];
+               Tar_unix.write_end s;
            with e ->
              Backtrace.is_important e;
              TaskHelper.failed ~__context e;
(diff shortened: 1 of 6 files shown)


ocaml/xapi/import_raw_vdi.ml
```diff
@@ -71,17 +71,26 @@ let localhost_handler rpc session_id vdi_opt (req: Request.t) (s: Unix.file_desc
...
-                                  content_type ] in
-                  Http_svr.headers s headers;
-                  if chunked
-                  then Vhd_tool_wrapper.receive (Vhd_tool_wrapper.update_task_progress __context) "raw" "chunked" s None path "" prezeroed
-                  else Vhd_tool_wrapper.receive (Vhd_tool_wrapper.update_task_progress __context) (Importexport.Format.to_string format) "none" s req.Request.content_length path "" prezeroed
-               );
+             let headers = Http.http_200_ok ~keep_alive:false () @
+                           [ Http.Hdr.task_id ^ ":" ^ (Ref.string_of task_id);
+                             content_type ] in
+             Http_svr.headers s headers;
```
gaborigloi: There is a small change in behaviour here: we'll send these headers, even if we fail to attach the VDI afterwards.
But this is probably not a big deal, because we can also fail during import.

ocaml/xapi/importexport.ml
```diff
@@ -340,6 +343,7 @@ module Format = struct
   let content_type = function
     | Raw -> "application/octet-stream"
     | Vhd -> "application/vhd"
+    | Tar -> "application/tar"
```
edwintorok: There is no `application/tar` in `/etc/mime.types`, there is an `application/x-tar` that you could use though. If you run `file --mime foo.tar` it will say `application/x-tar`.

ocaml/xapi/export_raw_vdi.ml
```diff
@@ -51,14 +45,29 @@ let localhost_handler rpc session_id vdi (req: Http.Request.t) (s: Unix.file_des
...
+                        (fun base_path -> copy (Some base_path) path)
+                    | None -> copy None path
+                 )
+             | Tar ->
+               (* We need to keep refreshing the session to avoid session timeout *)
+               let refresh_session = Xapi_session.consider_touching_session rpc session_id in
+               (* The third element of the tuple, the size, is currently ignored by the function *)
+               let size = Client.Client.VDI.get_virtual_size ~rpc ~session_id ~self:vdi in
+               Stream_vdi.send_all refresh_session s __context rpc session_id [(Xapi_globs.vdi_tar_export_dir, vdi, size)];
+               Tar_unix.write_end s;
```
edwintorok: Where did we write the beginning?

ocaml/xapi/import_raw_vdi.ml
```diff
@@ -71,17 +71,26 @@ let localhost_handler rpc session_id vdi_opt (req: Request.t) (s: Unix.file_desc
...
+                      if chunked
+                      then Vhd_tool_wrapper.receive (Vhd_tool_wrapper.update_task_progress __context) "raw" "chunked" s None path "" prezeroed
+                      else Vhd_tool_wrapper.receive (Vhd_tool_wrapper.update_task_progress __context) (Importexport.Format.to_string format) "none" s req.Request.content_length path "" prezeroed
+                   );
+               | Tar ->
+                 (* We need to keep refreshing the session to avoid session timeout *)
+                 let refresh_session = Xapi_session.consider_touching_session rpc session_id in
+                 let size = Client.VDI.get_virtual_size ~rpc ~session_id ~self:vdi in
+                 (* VDIs exported as TAR archives will always have inline checksums *)
+                 Stream_vdi.recv_all_vdi refresh_session s __context rpc session_id ~has_inline_checksums:true ~force:false [(Xapi_globs.vdi_tar_export_dir, vdi, size)] |> ignore;
```
edwintorok: Why do we ignore the return value?

ocaml/xapi/import_raw_vdi.ml
edwintorok: I think this comment was meant for the previous patch hunk. You could make a `send_headers` function and call it from both `copy` where it used to be called previously, and in the `Tar` branch as well.

ocaml/xapi/export_raw_vdi.ml
```diff
@@ -51,14 +45,29 @@ let localhost_handler rpc session_id vdi (req: Http.Request.t) (s: Unix.file_des
...
-                    Sm_fs_ops.with_block_attached_device __context rpc session_id base_vdi `RO
-                      (fun base_path -> copy (Some base_path) path)
-                  | None -> copy None path
-               )
+             let headers = Http.http_200_ok ~keep_alive:false () @ [
+                 Http.Hdr.task_id ^ ":" ^ (Ref.string_of task_id);
+                 Http.Hdr.content_type ^ ":" ^ content_type;
+                 Http.Hdr.content_disposition ^ ": attachment; filename=\"" ^ filename ^ "\""
+               ] in
+             Http_svr.headers s headers;
```
edwintorok: You could make a `send_headers` function and call it from both `copy` where it used to be called previously, and in the `Tar` branch as well to keep the old behaviour.

ocaml/xapi/import_raw_vdi.ml
gaborigloi: My question is, is that really necessary, or is this slight behaviour change okay? (As we can fail later during import anyway) - I don't want to complicate the code if it's not necessary.

ocaml/xapi/importexport.ml
gaborigloi: Thanks, I'll use x-tar. Looks like people use all kinda mimes: https://stackoverflow.com/questions/39058006/proper-mime-media-type-for-tar-files

ocaml/xapi/import_raw_vdi.ml
gaborigloi: Because we return the checksum table (file -> checksum) mapping.
In VM import, for old formats we store this table in the archive, and we compare it with the computed checksum table.
But here we don't use the checksums.

ocaml/xapi/export_raw_vdi.ml
gaborigloi: I think we don't need it, I followed the example of `export.ml` where we just call this at the end.
In `send_all`, we write the header for each file.
I'v enabled import-export quicktests in tar format with randomized data, and they seem to work fine.
I also managed to untar the export, it only complained about invalid dates.

gaborigloi:
Is this good to merge now?
(Or is the small behavioural change problematic where we still send the headers even if we fail to attach the VDI?)

edwintorok:
I'd be fine with it, but I'd like a second opinion from @jonludlam: this patch changes the place where we send the 200 OK headers, previously we would've caught a few more rare errors, before we sent 200, now instead we'll see those errors only after we sent the 200 OK headers.
Either way there is still a possibility for failure after 200 OK is sent (e.g. we could fail to read the VDI, or loose connectivity) so the client should handle failures after receiving a 200 HTTP code.

gaborigloi:
Does this look okay @jonludlam ?

jonludlam:
There has always been the possibility of a failure after the `200 OK` was sent so I don't see that we're making things much worse with this patch.

PR #3718, opened 2018-09-27
CP-28951: Add message of xen low memory alarm

krizex:

### DIFF
--- a/ocaml/xapi-consts/api_messages.ml
+++ b/ocaml/xapi-consts/api_messages.ml
@@ -132,6 +132,7 @@ let host_cpu_features_down = addMessage "HOST_CPU_FEATURES_DOWN" 3L
 let host_cpu_features_up = addMessage "HOST_CPU_FEATURES_UP" 5L
 let pool_cpu_features_down = addMessage "POOL_CPU_FEATURES_DOWN" 5L
 let pool_cpu_features_up = addMessage "POOL_CPU_FEATURES_UP" 5L
+let host_low_memory = addMessage "HOST_LOW_MEMORY" 2L
 
 (* Cluster messages *)
 let cluster_host_enable_failed = addMessage "CLUSTER_HOST_ENABLE_FAILED" 3L


ocaml/xapi-consts/api_messages.ml
```diff
@@ -132,6 +132,7 @@ let host_cpu_features_down = addMessage "HOST_CPU_FEATURES_DOWN" 3L
 let host_cpu_features_up = addMessage "HOST_CPU_FEATURES_UP" 5L
 let pool_cpu_features_down = addMessage "POOL_CPU_FEATURES_DOWN" 5L
 let pool_cpu_features_up = addMessage "POOL_CPU_FEATURES_UP" 5L
+let host_low_memory = addMessage "HOST_LOW_MEMORY" 1L
```
edwintorok: I think severity of 1 is too high, I think it should be at least a 2, see definitions at top of this file a host crashing due to out of memory would be 2, and if its not an imminent crash then 3:
```
1        Data-loss imminent    Take action now or your data may be permanently lost (e.g. corrupted)
 2        Service-loss imminent Take action now or some service(s) may fail (e.g. host / VM crash)
 3        Service degraded      Take action now or some service may suffer (e.g. NIC bond degraded without HA)
 4        Service recovered     Notice that something just improved (e.g. NIC bond repaired)
 5        Informational         More day-to-day stuff (e.g. VM started, suspended, shutdown, rebooted etc)
```

krizex: Thank you. I was not aware of that comments.

ocaml/xapi-consts/api_messages.ml
```diff
@@ -132,6 +132,7 @@ let host_cpu_features_down = addMessage "HOST_CPU_FEATURES_DOWN" 3L
 let host_cpu_features_up = addMessage "HOST_CPU_FEATURES_UP" 5L
 let pool_cpu_features_down = addMessage "POOL_CPU_FEATURES_DOWN" 5L
 let pool_cpu_features_up = addMessage "POOL_CPU_FEATURES_UP" 5L
+let host_low_memory = addMessage "HOST_LOW_MEMORY" 2L
```
robhoes: This isn't quite precise enough. The condition is that Xen itself is low on memory. There may still be free memory for VMs. So perhaps `XEN_LOW_MEMORY`.

ocaml/xapi-consts/api_messages.ml
edwintorok: It would be better if we used a variant type to specify severity, and then map that to an int somewhere else, instead of using magic numbers, but refactoring this whole file is outside the scope of this PR.

ocaml/xapi-consts/api_messages.ml
krizex: Isn't it true that free memory of xen equals to free memory of the host?
When I test the implementation, I continuously create VMs on the host to consume the memory and finally trigger the VIRQ_ENOMEM event of xen.

robhoes: I'm sure entirely sure how Xen's own memory management works. Perhaps @andyhhp could make a suggestion?

robhoes: Thanks, so @krizex is right.

PR #3722, opened 2018-09-30
CP-29196: Let stunnel in FIPS mode for both server/client

minli1:


minli1:
This PR is with xen-api-libs-transitional changes https://github.com/xapi-project/xen-api-libs-transitional/pull/52
As Wei Xie(@xiewei20082008 ) is on marriage leave, help to create PR for him.
BVT pass.

scripts/init.d-xapissl
```diff
@@ -90,7 +90,7 @@ writeconffile () {
     echo "; Autogenerated by ${0}" > $SSLCONFFILE
     writec '; during xapi start-up.'
     writec '; '
-    if [ ${ANCIENT_STUNNEL} = 0 ]; then
+    if [ ${ANCIENT_STUNNEL} = 0 ] && [ x$CC_PREPARATIONS != 'xtrue' ]; then
```
lindig: I think it would be better to use `x"$CC_PREPARATIONS" != x"true"` to highlight that the `x` is a dummy.

lindig:
Could the cert bundle be created during installation in the spec file?

ocaml/xapi/certificates.ml
```diff
@@ -146,7 +146,7 @@ let host_uninstall is_cert ~name =
   debug "Uninstalling %s %s" (get_type is_cert) name;
   try
     Sys.remove filename;
-    rehash()
+    ignore(execute_command_get_output "/opt/xensource/bin/update-ca-bundle.sh" [])
```
edwintorok: If rehash is not used anymore it should be deleted.

ocaml/xapi/certificates.ml
```diff
@@ -130,7 +130,7 @@ let host_install is_cert ~name ~cert =
     mkdir_cert_path is_cert;
     write_string_to_file filename cert;
     Unix.chmod filename (cert_perms is_cert);
-    rehash()
+    ignore(execute_command_get_output "/opt/xensource/bin/update-ca-bundle.sh" [])
```
edwintorok: Would be good to put this into a function, `update_ca_bundle`.

scripts/update-ca-bundle.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+#
+# Copyright (c) Citrix Systems 2008. All rights reserved.
+#
+
+set -e
+
+mkdir -p /etc/stunnel
+find /etc/stunnel/certs -name '*.pem' | xargs cat > /etc/stunnel/xapi-stunnel-ca-bundle.pem
```
edwintorok: Would be better if this created a temporary file during concatenation and atomically moved it when complete to avoid the situation where `stunnel` could see an incomplete file.

Also how is stunnel going to notice that there is a new certificate, do you need to signal it, restart it somehow?

ocaml/xapi/certificates.ml
minli1: Good point, will update.

ocaml/xapi/certificates.ml
minli1: rehash is also used here: https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/certificates.ml#L110

scripts/init.d-xapissl
minli1: 👍   Will update, thanks!

scripts/update-ca-bundle.sh
minli1: Thanks for your comments! @edwintorok 
Just to confirm, do you mean it's better to create a temp file to put all the content inside of `*.pem`? After that `cat` the content into `xapi-stunnel-ca-bundle.pem` to avoid incomplete file.
Like:
```find /etc/stunnel/certs -name '*.pem' | xargs cat > /etc/stunnel/temp 
cp /etc/stunnel/temp  /etc/stunnel/xapi-stunnel-ca-bundle.pem
rm -f /etc/stunnel/temp```

For the second, if there is a new certificate, the user should install the certificate with `xe pool-certificate-install`

Comment (edwintorok) [reply scripts/update-ca-bundle.sh]: `cp` is not atomic and would have the same problem as original `cat`, please use `mv /etc/stunnel/xapi-stunnel-ca-bundle.pem.tmp /etc/stunnel/xapi-stunnel-ca-bundle.pem`

Comment (minli1) [reply scripts/update-ca-bundle.sh]: Updated, thanks!

File: ocaml/xapi/certificates.ml [hunk 5]
```diff
@@ -28,8 +28,10 @@ let c_rehash = "/usr/bin/c_rehash"
 let pem_certificate_header = "-----BEGIN CERTIFICATE-----"
 let pem_certificate_footer = "-----END CERTIFICATE-----"
 
+let certificate_path = "/etc/stunnel/certs"
```
lindig: This could also go into `ocaml/xapi-consts/constants.ml` but I'm fine with keeping it here.

ocaml/xapi/certificates.ml
edwintorok: But that code would have no effect anymore since we use the certificate bundle instead of directory, should the code be removed?

minli1:
Thanks for the review! @lindig @edwintorok 
Since we are pending for the PR to run the integration test, please help to merge this one and xapi-project/xen-api-libs-transitional#52.

PR #3726, opened 2018-10-05
CP-28857 - New database copying

rwdrich:
Write out to disk
Turned off db copying
Add delay vars and reset counter
Poll master for db updates
Pass context, calculate in file
Force update when changing db in context
Fixed unit tests
using mtime


ocaml/database/schema.ml
```diff
@@ -35,7 +35,7 @@ module Value = struct
     | String of string
     | Set of string list
     | Pairs of (string * string) list
-  [@@deriving sexp]
+  [@@deriving sexp, rpc]
```
edwintorok: Are we sure we want to use `rpc` and not `rpcty` here?

rwdrich:
I need to add a don't merge label to this, I'm running the Ring3 BST on this on the next build, just thought I'd open the PR ready anyway

ocaml/xapi/context.ml
```diff
@@ -84,6 +84,9 @@ let string_of x =
 
 let database_of x = x.database
 
+let set_database x db =
```
edwintorok: This does not mutate `x`, perhaps a different name than `set` would be better (`update`?)

ocaml/xapi/pool_db_backup.ml
```diff
@@ -250,7 +250,7 @@ let fetch_database_backup ~master_address ~pool_secret ~force =
 (* Master sync thread *)
 let pool_db_backup_thread () = Debug.with_thread_named "pool_db_backup_thread" (fun () ->
     Server_helpers.exec_with_new_task "Pool DB sync" (fun __context ->
-        while (true) do
+        while (false) do
```
edwintorok: Should this be based on a configuration value in `/etc/xapi.conf`? If enabled we use the new mechanism (and have `false` here), if not we use the old mechanism. That way we can switch between the two without rebuilding.

ocaml/xapi/xapi_vm_clone.ml
```diff
@@ -399,8 +400,8 @@ let clone ?(snapshot_info_record) disk_op ~__context ~vm ~new_name =
               List.iter (fun (vbd, newvdi, _) ->
                   let vbd = Xapi_vbd_helpers.copy ~__context ~vm:ref ~vdi:newvdi vbd in
                   (* CA-58405: when we make a clone/snapshot/checkpoint we consider the clone/snapshot/checkpoint VM
-                     					   to "own" all the clone/snapshot/checkpoint *disks* irrespective of the ownership of the original
-                     					   disks. We wish the clone/snapshot/checkpoint disks to be cleaned up with the VM. *)
+                     to "own" all the clone/snapshot/checkpoint *disks* irrespective of the ownership of the original
+                     disks. We wish the clone/snapshot/checkpoint disks to be cleaned up with the VM. *)
```
edwintorok: The indentation changes should probably go into a separate commit, there's less noise that way when reviewing the main commit.

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,184 @@
...
+open Http
+open Xmlrpc_client
+open Db_cache_types
+open Stdext
+open Threadext
+
+type stat  = {created : int64; modified : int64; deleted : int64;    } [@@deriving rpc]
+type field = {fldname : string; stat : stat; value : Schema.Value.t; } [@@deriving rpc]
+type row   = {objref  : string; stat : stat; value : field list;     } [@@deriving rpc]
+type table = {tblname : string; stat : stat; value : row list;       } [@@deriving rpc]
```
edwintorok: minor stylistic: I don't think we usually align `}` like this, if you add a new field you'll have to realign everything again.
We usually put each field on its own line instead.

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,184 @@
...
+
+let get_del_from_table_name token ts table =
+  Table.fold_over_deleted token (fun key {created; modified; deleted} acc ->
+      ({key; table; stat={created;modified;deleted}})::acc) (TableSet.find table ts) []
+
+let get_deleted ~__context ~token db =
+  let tables =
+    let all =
+      let objs = List.filter (fun x -> x.Datamodel_types.gen_events) (Dm_api.objects_of_api Datamodel.all_api) in
+      List.map (fun x->x.Datamodel_types.name) objs in
```
edwintorok: `List.map` is not tail recursive, it can overflow the stack if the list is large. There's an alternative one in `xapi-stdext` .
If we move to 4.07 this would be a good candidate for use of `Seq` though (or we can start using `seq` with 4.06 if we add the library I think).

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,184 @@
...
+  let handle_table table = Table.fold_over_recent token (fun objref {created; modified; deleted} value acc
+                                                          -> {objref; stat={created;modified;deleted}; value=(handle_obj value)}::acc) table [] in
+  TableSet.fold_over_recent token (fun tblname {created; modified; deleted} value acc
+                                    -> {tblname; stat={created;modified;deleted}; value=(handle_table value)}::acc) ts []
+
+let get_delta ~__context ~token =
+  let db = Db_ref.get_database (Context.database_of __context) in
+  let db_access =
+    fun () -> Mutex.execute Db_lock.dbcache_mutex (
+        fun () -> {fresh_token=Manifest.generation (Database.manifest db);
```
edwintorok: minor style: prefer to put `fun () ->` on the previous line

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,184 @@
...
+  Xapi_slave_db.slave_db :=
+    Database.update (
+      (fun _ -> newval)
+      |> Row.update mtime fldname newval
+      |> Table.update mtime objref Row.empty
+      |> TableSet.update mtime tblname Table.empty) !Xapi_slave_db.slave_db
+
+let unpack_row __context (table:table) tblname row =
+  List.iter (fun (f:field)
+              -> Mutex.execute Xapi_slave_db.slave_db_mutex
```
edwintorok: minor style: arrow should go onto previous line.

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,184 @@
...
+let apply_changes (delta:delta) __context =
+  if (Db_cache_types.Manifest.generation (Db_cache_types.Database.manifest !Xapi_slave_db.slave_db)) < delta.fresh_token then
+    begin
+      Xapi_slave_db.slave_db := Database.set_generation delta.fresh_token !Xapi_slave_db.slave_db;
+      apply_updates __context delta.tables;
+      apply_deletes __context delta.deletes;
+    end;
+  (* Need to check that we haven't missed anything *)
+  if not (List.for_all (fun x -> x)
+      (List.map2 count_eq_check (object_count !Xapi_slave_db.slave_db) (object_count (Db_ref.get_database (Context.database_of __context))))) then
```
edwintorok: `List.for_all identity (List.map2 f ..)` is better written as `List.for_all2 f`

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,184 @@
...
+  let tbls, objs = List.split (info_extract d.tables) in
+  List.iter2 (send_notification __context "mod") tbls objs
+
+let write_db_to_disk =
+  debug "Writing slave db to slave disk";
+  Db_cache_impl.sync (Db_conn_store.read_db_connections ()) !Xapi_slave_db.slave_db
+
+let endless_loop ~__context () =
+  let uri = Constants.database_backup_uri in
+  let delay = Mtime.Span.of_uint64_ns 300_000_000_000L in
```
edwintorok: Isn't there a constant you could use to specify this in seconds?

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,184 @@
...
+      (with_http req
+         (fun (response, fd) ->
+            let res = match response.Http.Response.content_length with
+              | None -> raise Content_length_required
+              | Some l -> Xapi_stdext_unix.Unixext.really_read_string fd (Int64.to_int l)
+            in
+            let delta_change = delta_of_rpc (Jsonrpc.of_string res) in
+            apply_changes delta_change __context;
+            handle_notifications __context delta_change;
+            if (Mtime.span (Mtime_clock.now ()) !now) > delay then
```
edwintorok: I think you can use the `counter` function in Mtime to do this for you, you initialize it, and call `count` when you want to know how much has elapsed.

robhoes:
Initial comment... This is one large commit that deserves to be split up in a couple of separate ones to make it easier to review, and easier to understand later when looking back at the history. The various line items in the commit message already suggest ways of splitting it. Also it is better to keep whitespace fixes of existing code in separate commit.

Each commit should have a clear message explaining why and whay, and a "Signed-off-by" line.

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,184 @@
...
+
+let info_extract (tlist:table list) =
+  let tbls = List.map (fun (i:table) -> i.tblname) tlist in
+  let rows = List.concat (List.map (fun tbl -> List.map (fun row -> row.objref) tbl.value) tlist) in
+  List.concat (List.map ((fun objlist tblname -> List.combine [tblname] objlist) rows) tbls)
+
+let handle_notifications __context (d:delta) =
+  List.iter ((fun __context (d:del_tables) -> send_notification __context "del" d.table d.key) __context) d.deletes;
+  let tbls, objs = List.split (info_extract d.tables) in
+  List.iter2 (send_notification __context "mod") tbls objs
```
edwintorok: Could this and info_extract be written with a `fold` instead? It looks like we are constructing a lot of intermediate values, only to iterate over them.

ocaml/xapi/xapi_database_backup.ml
lindig: The style below leaves room to grow and for comments. Keeping at semicolon at the front is a Haskell-ism that is spreading in the OCaml community.
```
type field = 
  { fldname :   string    (** room for comment *)
  ; stat :      stat
  ; value :     Schema.Value.t
  } [@@deriving rpc]
```

ocaml/xapi/xapi_database_backup.ml
rwdrich: I'm not sure that seq is correct for this? Wouldn't the benefit of Seq be from a list that isn't fully created? I already have the full list in the reply from `(Dm_api.objects_of_api Datamodel.all_api)` , or am I missing something

ocaml/database/schema.ml
rwdrich: I'm not sure I understand the difference between the two sufficiently. But rpcty causes errors in the type of Value for me

rwdrich: Ok, I've had a chat with Jon, and he's pointed out what the problem is, this is my (paraphrased) understanding of it.
In trying to use rpcty with Schema.Value, the auto generated code has multiple instances of `String`. `let open! Rpc.Types in` is called above each of the type options, and as such inside the definition of `String of string`, the compiler ~doesn't know~ gets the wrong `String`. This means that it's trying to apply the rpc String, where it should be applying the Schema.Value String.
I think it's something that Jon has added to the infinite 'to do' list, but in the mean time the solution is, 'Use rpc, rather than rpcty'

ocaml/xapi/xapi_host_helpers.ml
```diff
@@ -309,6 +309,7 @@ module Configuration = struct
 
   let watch_other_configs ~__context delay =
     let loop (token, was_in_rpu) =
+      let __context = Xapi_slave_db.update_context_db __context in
```
robhoes: As discussed in person, this does not look to be enough to get the event loop use the local DB, because the `Event.from` call still goes through the API, and the master.

ocaml/xapi/xapi_vm_clone.ml
```diff
@@ -32,6 +32,7 @@ let delete_disks rpc session_id disks =
     ) disks
 
 let wait_for_subtask ?progress_minmax ~__context task =
+  let __context = Xapi_slave_db.update_context_db __context in
```
robhoes: In this case, the context is switched over to the local DB too early. This affects DB operations (such as `Db_actions.DB_Action.Task.set_current_operations`) that are not for the event loop, which still need to use the master DB.

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,175 @@
+module D=Debug.Make(struct let name="xapi_database_backup" end)
+open D
+
+open Http
+open Xmlrpc_client
+open Db_cache_types
+open Stdext
+open Threadext
+
+type stat = {created : int64; modified : int64; deleted : int64;} [@@deriving rpc]
```
robhoes: This is identical to `Db_cache_types.Stat.t`, correct? But with rpc-deriving added to it.

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,175 @@
...
+type table = {tblname : string; stat : stat; value : row list;} [@@deriving rpc]
+
+type del_tables = {key: string; table: string; stat : stat;} [@@deriving rpc]
+type delta = {fresh_token: int64; tables : table list; deletes : del_tables list;} [@@deriving rpc]
+type edits = {ts : TableSet.t; tables : string list}
+
+type count = {tblname:string; count: int; del:int;}
+type pair = {tbl:string; row:string;}
+
+let empty_schema = Schema.Value.unmarshal String ""
```
robhoes: Is this an empty-string value rather than an empty schema? You can probably just write `Schema.Value.String ""`?

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,175 @@
...
+let empty_field = {fldname=""; stat={created=0L;modified=0L;deleted=0L;}; value=empty_schema;}
+let empty_row   = {objref =""; stat={created=0L;modified=0L;deleted=0L;}; value=[empty_field];}
```
robhoes: This row isn't really "empty"... Would `value=[]` not do?

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,175 @@
...
+
+open Http
+open Xmlrpc_client
+open Db_cache_types
+open Stdext
+open Threadext
+
+type stat = {created : int64; modified : int64; deleted : int64;} [@@deriving rpc]
+type field = {fldname : string; stat : stat; value : Schema.Value.t;} [@@deriving rpc]
+type row = {objref  : string; stat : stat; value : field list;} [@@deriving rpc]
```
robhoes: It may be clearer to use `fields` rather than `value`, and `rows` below.

ocaml/xapi/xapi.ml
```diff
@@ -933,6 +934,7 @@ let server_init() =
           "Starting SR physical utilisation scanning", [Startup.OnThread], (Xapi_sr.physical_utilisation_thread ~__context);
           "Caching metadata VDIs created by foreign pools.", [ Startup.OnlyMaster; ], cache_metadata_vdis;
           "Stats reporting thread", [], Xapi_stats.start;
+          "Slave database backup", [ Startup.OnlySlave; Startup.OnThread; ], (Xapi_database_backup.endless_loop ~__context);
```
robhoes: This may be too late in the startup sequence: if any of the startup items above need the slave DB. We can do this immediately after establishing the master connection, so move it to the beginning of this list.

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,175 @@
...
+  let _dbreply = db_access () in
+  _dbreply
+
+let handler (req: Request.t) s _ =
+  req.Http.Request.close <- true;
+  let token = Generation.of_string (snd (List.find (fun (x,y) -> x = "token") req.query)) in
+  Xapi_http.with_context "Database backup" req s (fun __context ->
+      let d = get_delta __context token in
+      let str_update = (Jsonrpc.to_string (rpc_of_delta d)) in
+      Http_svr.response_str req s str_update
```
robhoes: You could use a pipeline here:
```
token
|> get_delta __context
|> rpc_of_delta
|> Jsonrpc.to_string
|> Http_svr.response_str req s
```

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,175 @@
...
+        {fresh_token=Manifest.generation (Database.manifest db);
+         tables=(check_for_updates token db);
+         deletes=(get_deleted __context token db)})
+  in
+  let _dbreply = db_access () in
+  _dbreply
+
+let handler (req: Request.t) s _ =
+  req.Http.Request.close <- true;
+  let token = Generation.of_string (snd (List.find (fun (x,y) -> x = "token") req.query)) in
```
robhoes: Let's handle the `Not_found` exception that `List.find` may raise.

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,175 @@
...
+let info_extract (tlist:table list) =
+  List.fold_left (fun pl (t:table) -> (fl t.tblname t.value) @ pl) [] tlist
+
+let handle_notifications __context (d:delta) =
+  List.iter ((fun __context (d:del_tables) -> send_notification __context "del" {tbl=d.table; row=d.key}) __context) d.deletes;
+  List.iter (send_notification __context "mod") (info_extract d.tables)
+
+let endless_loop ~__context () =
+  let uri = Constants.database_backup_uri in
+  while (true) do
```
robhoes: This loop isn't going to be endless if an exception happens inside of it, for example if the master connection breaks. You need to then restart the loop after a short delay.

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,175 @@
...
+    let addr = Pool_role.get_master_address() in
+    let psec = !Xapi_globs.pool_secret in
+    let token_str = Int64.to_string (Manifest.generation (Database.manifest !Xapi_slave_db.slave_db)) in
+    let req = Xapi_http.http_request ~cookie:["pool_secret", psec] ~query:[("token", token_str)] Http.Get uri in
+    let transport = SSL(SSL.make (), addr, !Xapi_globs.https_port) in
+    with_transport transport
+      (with_http req
+         (fun (response, fd) ->
+            let res = match response.Http.Response.content_length with
+              | None -> raise Content_length_required
```
robhoes: This exception, for example, is unhandled and prevents any further local DB updates until you restart xapi.

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,175 @@
...
+      Http_svr.response_str req s str_update
+    )
+
+(* Slave Functions *)
+
+let send_notification (c:Context.t) reason p =
+  let record = (Eventgen.find_get_record p.tbl ~__context:c ~self:p.row) () in
+  begin match record with
+    | None -> debug "Could not send %s event for %s in %s table" reason p.row p.tbl;
+    | Some record -> debug "Sending %s event for %s object in %s table" reason p.row p.tbl;
```
robhoes: This is probably going to result in lots and lots of log lines? This may be handy for dev testing, but not in the final version of the patch.

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,175 @@
...
+  let tables = get_table_list db in
+  List.map (counter db) tables
+
+let count_eq_check c1 c2 =
+  c1.tblname = c2.tblname && c1.count = c2.count && c1.del = c2.del
+
+let apply_changes (delta:delta) __context =
+  if (Db_cache_types.Manifest.generation (Db_cache_types.Database.manifest !Xapi_slave_db.slave_db)) < delta.fresh_token then
+    begin
+      Xapi_slave_db.slave_db := Database.set_generation delta.fresh_token !Xapi_slave_db.slave_db;
```
robhoes: No `Mutex.execute`?

robhoes: Hmm... I'm a bit uncomfortable about the use of the global `Xapi_slave_db.slave_db` in many places and the use of the mutex at different levels (e.g. below here and inside `unpack_row`). Some of the functions about take a db argument and do functional updates, but others with directly on the global db, so it is a little hard to see that this is correct, and no deadlocks can occur. How about adding a db argument to all functions?

E.g.
```
let new_db =
  !Xapi_slave_db.slave_db
  |> apply_updates ...
  |> apply_deletes ...
in
Xapi_slave_db.slave_db := new_db
```

And just use one `Mutex.execute` around `apply_changes`?

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,175 @@
...
+      apply_updates __context delta.tables;
+      apply_deletes __context delta.deletes;
+    end;
+  (* Need to check that we haven't missed anything *)
```
robhoes: Why?

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,175 @@
...
+      |> TableSet.update mtime tblname Table.empty) !Xapi_slave_db.slave_db
+
+let unpack_row __context (table:table) tblname row =
+  List.iter (fun (f:field)->
+      Mutex.execute Xapi_slave_db.slave_db_mutex
+        (fun () -> make_change __context f.stat.modified tblname row.objref f.fldname f.value)
+    ) row.value
+
+let unpack_table __context (t:table) =
+  List.iter (unpack_row __context t t.tblname) (if t.value = [] then [empty_row] else t.value)
```
robhoes: I still don't really follow the `empty_row` thing here. Can't we just have a table without rows?

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -191,9 +191,16 @@ let handle_notifications __context (d:delta) =
   List.iter ((fun __context (d:del_tables) -> send_notification __context "del" {tbl=d.table; row=d.key}) __context) d.deletes;
   List.iter (send_notification __context "mod") (info_extract d.tables)
 
+let write_db_to_disk =
+  debug "Writing slave db to slave disk";
+  Db_cache_impl.sync (Db_conn_store.read_db_connections ()) !Xapi_slave_db.slave_db
+
 let endless_loop ~__context () =
   let uri = Constants.database_backup_uri in
+  let delay_f = 300.0 in
```
robhoes: We should move this to Xapi_globs and make it configurable from the config file (like the other one you did).

ocaml/xapi/pool_db_backup.ml
```diff
@@ -250,7 +250,7 @@ let fetch_database_backup ~master_address ~pool_secret ~force =
 (* Master sync thread *)
 let pool_db_backup_thread () = Debug.with_thread_named "pool_db_backup_thread" (fun () ->
     Server_helpers.exec_with_new_task "Pool DB sync" (fun __context ->
-        while (true) do
+        while (not !Xapi_globs.slave_dbs) do
```
robhoes: We should make `write_db_to_disk` from the local DB also conditional on this flag. I'd also prefer to not do `Debug.with_thread_named` and `Server_helpers.exec_with_new_task`, so leave this `while true`, and put an if-statement on top.

ocaml/xapi/xapi_database_backup.ml
rwdrich: Yes. I can't seem to get it to work with `Db_cache_types.Stat.t`, I can't make that type work with deriving rpc

ocaml/xapi/context.ml
rwdrich: I've changed this to update

ocaml/xapi/pool_db_backup.ml
rwdrich: I've changed this to use the config file. But this will probably change again very soon re: Rob's comments

ocaml/xapi/xapi_database_backup.ml
rwdrich: It counts the objects in case we are in the scenario where more than the maximum length of the deleted items list have occurred. In which case the objects that we have may not be correct. Though I'm not sure what we do in this scenario is necessarily 'the right thing'.

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,218 @@
...
+  end
+
+let apply_deletes __context (deletes:del_tables list) =
+  List.iter
+    (fun dtbl ->
+       (fun tblname objref ->
+          if Db_cache_impl.is_valid_ref (Context.database_of __context) objref then
+            Xapi_slave_db.slave_db := Db_cache_types.remove_row tblname objref !Xapi_slave_db.slave_db
+          else
+            Printf.printf "Not valid ref"
```
robhoes: Where do you want this to get printed...?

rwdrich:
Currently building after the latest changes - I'll run a BST on the build if it passes and then update here

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,249 @@
...
+      db
+    end
+
+let apply_changes (delta:delta) __context =
+  if (Db_cache_types.Manifest.generation (Db_cache_types.Database.manifest !Xapi_slave_db.slave_db)) < delta.fresh_token then
+    begin
+      let new_db =
+        !Xapi_slave_db.slave_db
+        |> Database.set_generation delta.fresh_token
+        |> List.fold_right make_change_with_db_reply (flatten_table_to_update delta.tables)
```
robhoes: `List.fold_right` unfortunately isn't tail recursive (`List.fold_left` is). This would be a problem is the list can be long.

rwdrich: Very true. I've changed this to use fold_left instead

ocaml/xapi/xapi_event.ml
```diff
@@ -604,6 +604,16 @@ let inject ~__context ~_class ~_ref =
 
 (* Internal interface ****************************************************)
 
+let create_call_task ~__context loc =
+  debug "Creating wakeup for %s" loc;
+  let task_id = (TaskHelper.XapiEvent ("TaskForEvents: " ^ loc)) in
```
robhoes: This doesn't give you a task in the DB. You could use `Xapi_task.create`.

robhoes: By the way, `TaskHelper` is weirdly entangled with the `Context` module... I've got some patches to clean it up a little, but that's for later.

rwdrich: Using Xapi_task from here creates a cyclic dependency. It should still be possible, but might need a bit of a shuffle

robhoes: Or just do the same as that function does, but inline, e.g.:
```
let task_id =
  Context.make ~task_description:"something useful" ~task_in_database:true "suitable name"
  |> Context.get_task_id
in
```

ocaml/xapi/message_forwarding.ml
```diff
@@ -698,10 +698,10 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
         let unfinished = List.exists (fun state -> state = `pending) statuses in
         if unfinished
         then begin
-          let from = Helpers.call_api_functions ~__context
-              (fun rpc session_id -> Client.Event.from ~rpc ~session_id ~classes ~token ~timeout:30.0) in
+          let timeout = 30.0 in
+          let classes = classes in
```
robhoes: Really?

rwdrich: To be extra sure...
oops, there was a point where some of that line was different and thus useful. Removed now

rwdrich:
I believe so, but a unit test for that would be a good idea.

A PN counter on the master would not always be the same as a PN counter on the slave. If we insert something, and then immediately delete it, the delta taken from the changes will not insert it, because it will only have the information about the delete, so the counters wouldn't be the same, even though the state of the db would be. Whereas instead if we count the objects in the db at that time, and the length of the deleted list, then they will be the same

ocaml/database/db_cache_types.ml
```diff
@@ -144,7 +145,7 @@ module Table = struct
     let rec loop xs acc = match xs with
       | (created,deleted,r)::xs ->
         let new_acc =
-          if (deleted > since) && (created <= since)
+          if (deleted >= since) && (created <= since)
```
edwintorok: Are you sure you still want to change this if you made Event.from behave the same?

rwdrich: No. I don't want to make this change. Thought I'd removed that. That also affects my test results.

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,357 @@
...
+
+let compare_by_tblname (c1: (string * int)) (c2: (string * int)) =
+  String.compare (fst c1) (fst c2)
+
+let count_check (c1_list: (string * int) list) (c2_list: (string * int) list) =
+  let c1_list = List.sort compare_by_tblname c1_list in
+  let c2_list = List.sort compare_by_tblname c2_list in
+  try
+    List.for_all2 (fun (a,b) (c,d) -> a=c && b=d) c1_list c2_list
+  with invalid_arg -> false
```
lindig: `invalid_arg` is an identifier here that binds an exception and not an exception itself - because it is lowercase. Is this intended?

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,357 @@
...
+  | "" -> make_empty_table update.stat update.tblname db
+  | _ -> (match update.fldname with
+      | "" -> make_from_barrier update.stat update.tblname update.objref db
+      | _ -> make_change update.stat update.tblname update.objref update.fldname update.value db)
+
+let make_delete_with_db_reply db delete =
+  remove_row delete.table delete.key db
+
+let modified_check db tblname =
+  let _t = TableSet.find tblname (Database.tableset db) in
```
lindig: An underscore typically indicates an unused value. To indicate a derived value; `t'` (prime) could be used.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -2370,87 +2370,63 @@ let events_from_xapi () =
...
-           debug "Woken event thread: updating list of event subscriptions"
-         | e ->
-           debug "Caught %s listening to events from xapi" (string_of_exn e);
-           (* Start from scratch *)
-           token := "";
-           Thread.delay 15.
+         (* We register for events on resident_VMs only *)
+         let resident_VMs = Db.Host.get_resident_VMs ~__context ~self:localhost in
+
+         let uuids = List.map (fun self -> Db.VM.get_uuid ~__context ~self) resident_VMs in
```
lindig: Are lists here the right data structure? A string set is easy to have: `Set.Make(String)`.

rwdrich: This change is potentially taking place here: https://github.com/xapi-project/xen-api/pull/3790 as it is not directly linked to this PR

ocaml/database/slave_backup_master_connection.ml
```diff
@@ -0,0 +1,262 @@
...
+            open_secure_connection()
+          with _ -> () (* oh well, maybe nextime... *)
+        end
+    end
+  done;
+  !result
+
+let execute_remote_fn string =
+  let host = (!get_master_address) () in
+  Db_lock.with_lock
```
robhoes: If we use the same lock as used by the regular DB calls on the slave, aren't we then starving those?
And if we really need the same lock, does it then help to have a separate connection here?

ocaml/database/slave_backup_master_connection.ml
```diff
@@ -0,0 +1,262 @@
...
+
+(* kill the stunnel underlying the connection.. When the master dies
+   then read/writes to the connection block for ages waiting for the
+   TCP timeout. By killing the stunnel, we can get these calls to
+   unblock. (We do not clean up after killing stunnel, it is still the
+   duty of the person who initiated the stunnel connection to call
+   Stunnel.disconnect which closes the relevant fds and does wait_pid on
+   the (now dead!) stunnel process.
+*)
+let force_connection_reset () =
```
robhoes: This file is now mostly a copy of master_connection.ml, which is not what we want. We need to make master_connection more generic where needed, while being careful about all the parts of xapi that interact with it. E.g., this `force_connection_reset` function must probably always be called on both connections.

Since `Master_connection` has so many `ref` values on the top level, we probably need to turn most of it into a functor, while providing a common high-level interface where needed.

robhoes: For reference, the "interesting" interactions with the master connection are things like management IP addresses changes and pool master changes/failover.

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,357 @@
...
+  Mutex.execute Db_lock.dbcache_mutex (fun () ->
+      let db = Db_ref.get_database (Context.database_of __context) in
+      let tables = check_for_updates token db in
+      let deleted = get_deleted token db in
+      let t = Manifest.generation (Database.manifest db) in
+      {fresh_token=t;
+       last_event_token=Int64.pred t;
+       tables=tables;
+       deletes=deleted;
+       counts=object_count db})
```
robhoes: Both `object_count` and `get_deleted` internally call `get_table_list db`. We can pull that out and do it only once.

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,357 @@
...
+
+let get_gen db =
+  db
+  |> Database.manifest
+  |> Manifest.generation
+
+(* Master Functions *)
+
+let get_table_list db =
+  TableSet.fold_over_recent (-2L) (fun table stat value acc -> table::acc) (Database.tableset db) []
```
robhoes: Just `TableSet.fold` would probably do here, since you aren't actually asking for the "recent" tables?

ocaml/xapi/xapi_xenops.ml
```diff
@@ -2370,87 +2370,63 @@ let events_from_xapi () =
...
-                      | _ -> warn "Received event for something we didn't register for!"
-                    ) from.events;
-                  token := from.token;
-                  Events_from_xapi.broadcast !token;
-                done
-             )
-         with
-         | Api_errors.Server_error(code, _) when code = Api_errors.session_invalid ->
-           debug "Woken event thread: updating list of event subscriptions"
-         | e ->
```
robhoes: We don't need the exception case for `session_invalid` anymore, but we do need the catch-all one, otherwise we risk exiting from the event loop.

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,357 @@
...
+    if Mtime.Span.to_s now > delay_seconds then write_db_to_disk;
+    if changes.last_event_token > (get_gen !Xapi_slave_db.slave_db) then begin
+      token_str := Int64.to_string (get_gen !Xapi_slave_db.slave_db);
+    end else
+      token_str := Int64.to_string changes.last_event_token;
+  done
+
+let slave_db_backup_loop ~__context () =
+  while (true) do
+    Xapi_event.register_hooks ();
```
robhoes: This should go outside the loop.

ocaml/xapi/xapi_database_backup.ml
```diff
@@ -0,0 +1,351 @@
...
+    let new_master_address = Pool_role.get_master_address () in
+    let new_pool_secret = !Xapi_globs.pool_secret in
+    if new_master_address <> !addr || new_pool_secret <> !psec then begin
+      debug "The address or pool secret has changed - Make a new connection";
+      addr := new_master_address;
+      psec := new_pool_secret;
+      start_stunnel_connection ();
+    end;
+    let now = Mtime_clock.count start in
+    let changes =
```
rwdrich: I don't think that this is sleeping on changes anymore. This will be a relatively tight loop querying the master

rwdrich: This is fixed now

lindig:
This has been open for over a year. Closing this for now.

PR #3729, opened 2018-10-09
Private/thomasmc/cp 29604

thomasmck:
Done some basic manual testing using utop with the functions copied in and it looked like it worked:

utop # process [] 0 [{length = 7L; status_flags = 3l}; {length = 5L; status_flags = 1l}; {length = 6L; status_flags = 3l}; {length = 6L; status_flags = 1l};{length = 1L; status_flags = 3l};{length = 1L; status_flags = 1l};{length = 2L; status_flags = 3l}];;
- : int list * int = ([0; 1; 2; 3; 4; 5], 28)

### DIFF
--- a/ocaml/xapi/stream_vdi.ml
+++ b/ocaml/xapi/stream_vdi.ml
@@ -111,6 +111,7 @@ let cycle_descriptors descriptor_list offset =
     if a > b then []
     else a :: range (a + 1) b
   in
+  (* Insert proper not_empty function here *)
   let not_empty status_flag =
     match (Int32.to_int status_flag) with
       | 3 -> true
@@ -120,7 +121,8 @@ let cycle_descriptors descriptor_list offset =
     match (not_empty descriptor.status_flags) with
         | true -> begin
             let start_chunk = offset / increment in
-            let end_chunk = (Int64.to_int descriptor.length + offset) / increment in
+            (* If the chunk ends at the boundary don't include the next chunk *)
+            let end_chunk = (Int64.to_int descriptor.length + offset - 1) / increment in
             let chunks = range start_chunk end_chunk in
             chunks, descriptor.length
             end
--- a/ocaml/xapi/stream_vdi.ml
+++ b/ocaml/xapi/stream_vdi.ml
@@ -141,13 +141,11 @@ let cycle_descriptors descriptor_list offset =
   let rec process acc offset = function
     | [] -> acc, offset
     | x::xs -> begin
-        (* Insert proper increment here *)
-        let increment = 5 in
-        let chunks, add_offset = find_blocks x offset increment in
+        let chunks, add_offset = find_blocks x offset (Int64.to_int chunk_size) in
         process (add_unique_chunks acc chunks) (offset + (Int64.to_int add_offset)) xs
     end
   in
-  let chunks, offset = process [] 0 descriptor_list  in
+  let chunks, offset = process [] offset descriptor_list  in
   chunks, offset
 
 (** Stream a set of VDIs split into chunks in a tar format in a defined order. Return an
--- a/ocaml/xapi/stream_vdi.ml
+++ b/ocaml/xapi/stream_vdi.ml
@@ -109,6 +109,7 @@ type descriptor = {
 let flag_hole = 1l
 let flag_zero = 2l
 
+(** [descriptor_list] should be an list of non-overlapping extents, ordered from lowest offset to highest. *)
 let cycle_descriptors descriptor_list offset =
   (* Output range includes start and end points *)
   let rec range acc start_chunk end_chunk =
--- a/ocaml/xapi/stream_vdi.ml
+++ b/ocaml/xapi/stream_vdi.ml
@@ -134,7 +134,7 @@ let cycle_descriptors descriptor_list offset =
   in
 
   let rec add_unique_chunks acc = function
-    | [] -> List.rev acc
+    | [] -> acc
     | x::xs -> add_unique_chunks (check_exists x acc) xs
   in
 
--- a/ocaml/xapi/stream_vdi.ml
+++ b/ocaml/xapi/stream_vdi.ml
@@ -100,6 +100,52 @@ let write_block ~__context filename buffer ofd len =
     then raise (Api_errors.Server_error (Api_errors.client_error, [ExnHelper.string_of_exn e]))
     else raise e
 
+type descriptor = {
+  length : int64;
+  status_flags : int32;
+}
+
+
+let cycle_descriptors descriptor_list offset =
+  let round_up a b =
+    if a mod b == 0
+        then (a / b)
+        else (a / b) + 1
+  in
+  let rec range a b =
+    if a > b then []
+    else a :: range (a + 1) b
+  in
+  let not_empty status_flag =
+    match (Int32.to_int status_flag) with
+      | 3 -> true
+      | _ -> false
+  in
+  let find_blocks descriptor offset increment =
+    (* List of chunks to be copied *)
+    (* Assume existence of "not_empty" function for status flags *)
+    match (not_empty descriptor.status_flags) with
+        | true -> begin
+            (* Rounds down *)
+            let start_chunk = offset / increment in
+            let num_chunks = round_up (Int64.to_int descriptor.length) increment in
+            let chunks = range start_chunk (start_chunk + num_chunks) in
+            chunks, descriptor.length
+            end
+        | _ -> [], descriptor.length
+  in
+  let rec loop descriptor_list offset chunks =
+      if (List.length descriptor_list) == 1 then begin
+          let new_chunks, add_offset = find_blocks (List.hd descriptor_list) offset 1024 in
+          (chunks@new_chunks), (Int64.add add_offset (Int64.of_int offset))
+      end
+      else begin
+          let new_chunks, add_offset = loop [List.hd descriptor_list] offset chunks in
+          let rec_chunks, rec_offset = loop (List.tl descriptor_list) (Int64.to_int add_offset) new_chunks in
+          rec_chunks, rec_offset
+      end in
+  let chunks, offset = loop descriptor_list offset [] in
+  chunks, offset
 
 (** Stream a set of VDIs split into chunks in a tar format in a defined order. Return an
     association list mapping tar filename -> string (containing the SHA1 checksums) *)
--- a/ocaml/xapi/stream_vdi.ml
+++ b/ocaml/xapi/stream_vdi.ml
@@ -100,18 +100,13 @@ let write_block ~__context filename buffer ofd len =
     then raise (Api_errors.Server_error (Api_errors.client_error, [ExnHelper.string_of_exn e]))
     else raise e
 
+(* Example = [{"length": 2097152, "flags": 0}, {"length": 1902848, "flags": 3}]*)
 type descriptor = {
   length : int64;
   status_flags : int32;
 }
 
-
 let cycle_descriptors descriptor_list offset =
-  let round_up a b =
-    if a mod b == 0
-        then (a / b)
-        else (a / b) + 1
-  in
   let rec range a b =
     if a > b then []
     else a :: range (a + 1) b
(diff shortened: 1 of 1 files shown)


ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +100,53 @@ let write_block ~__context filename buffer ofd len =
...
+    match (Int32.to_int status_flag) with
+      | 3 -> true
+      | _ -> false
+  in
+  let find_blocks descriptor offset increment =
+    match (not_empty descriptor.status_flags) with
+        | true -> begin
+            let start_chunk = offset / increment in
+            (* If the chunk ends at the boundary don't include the next chunk *)
+            let end_chunk = (Int64.to_int descriptor.length + offset - 1) / increment in
```
gaborigloi: Ah I see, so this works because chunks start at `0`

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +100,53 @@ let write_block ~__context filename buffer ofd len =
...
+(* Example = [{"length": 2097152, "flags": 0}, {"length": 1902848, "flags": 3}]*)
+type descriptor = {
+  length : int64;
+  status_flags : int32;
+}
+
+let cycle_descriptors descriptor_list offset =
+  let rec range a b =
+    if a > b then []
+    else a :: range (a + 1) b
```
mseri: This is not tail recursive

gaborigloi: Good catch! We should use an accumulator instead.

thomasmck: What is the issue with it not being tail recursive?

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +100,53 @@ let write_block ~__context filename buffer ofd len =
...
+
+  let rec process acc offset = function
+    | [] -> acc, offset
+    | x::xs -> begin
+        let chunks, add_offset = find_blocks x offset (Int64.to_int chunk_size) in
+        process (add_unique_chunks acc chunks) (offset + (Int64.to_int add_offset)) xs
+    end
+  in
+  let chunks, offset = process [] offset descriptor_list  in
+  List.map (fun a -> a * Int64.to_int chunk_size) chunks, offset
```
thomasmck: @stefanopanella Updated to return chunks in chunk_size format

ocaml/xapi/stream_vdi.ml
gaborigloi: If the list is not processed tail recursively, every element will result in a new function call & stack frame, and we might get a stack overflow for big lists: https://dev.realworldocaml.org/lists-and-patterns.html#scrollNav-5

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +100,53 @@ let write_block ~__context filename buffer ofd len =
...
+  let rec add_unique_chunks acc = function
+    | [] -> acc
+    | x::xs -> add_unique_chunks (check_exists x acc) xs
+  in
+
+  let rec process acc offset = function
+    | [] -> acc, offset
+    | x::xs -> begin
+        let chunks, add_offset = find_blocks x offset (Int64.to_int chunk_size) in
+        process (add_unique_chunks acc chunks) (offset + (Int64.to_int add_offset)) xs
```
gaborigloi: Maybe it's safer to use int64 everywhere so that we don't have to rely on the assumption that ints on XS are 64-bit (which should be true always)

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +100,53 @@ let write_block ~__context filename buffer ofd len =
...
+  status_flags : int32;
+}
+
+let cycle_descriptors descriptor_list offset =
+  let rec range a b =
+    if a > b then []
+    else a :: range (a + 1) b
+  in
+  (* Insert proper not_empty function here *)
+  let not_empty status_flag =
```
lindig: Unless you expect more cases to come up, this could be simplified:
```
let not_empty status_flag =
  Int32.to_int status_flag = 3
```

ocaml/xapi/stream_vdi.ml
lindig: Indentation looks too deep for the cases. It's a bit unusual to use match rather than if for boolean expressions but it results it a nice table-like layout.

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +100,53 @@ let write_block ~__context filename buffer ofd len =
...
+            let start_chunk = offset / increment in
+            (* If the chunk ends at the boundary don't include the next chunk *)
+            let end_chunk = (Int64.to_int descriptor.length + offset - 1) / increment in
+            let chunks = range start_chunk end_chunk in
+            chunks, descriptor.length
+            end
+        | false -> [], descriptor.length
+  in
+  let check_exists a b =
+    if List.exists ((=)a) b = false then b@[a]
```
lindig: Could this be written with `List.mem`? (`List.mem a b`). The name could be improved: what is being checked here?

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +100,53 @@ let write_block ~__context filename buffer ofd len =
...
+       else b
+  in
+
+  let rec add_unique_chunks acc = function
```
lindig: Could this be expressed with `List.filter`? Given a list of elements `xs`, retain all those that satisfy a predicate.

ocaml/xapi/stream_vdi.ml
thomasmck: Ah looks like a have used a mix of 2 and 4 space identations. What should it be? Would you prefer I swap to if/else?

ocaml/xapi/stream_vdi.ml
thomasmck: Discussed that this is not actually needed here. Can do in Stefano's function instead

ocaml/xapi/stream_vdi.ml
thomasmck: Yep will update

ocaml/xapi/stream_vdi.ml
thomasmck: @gaborigloi mentioned that there is some `Nbd.Protocol.AllocationMetadataFlag.of_int32` function we can use in the final implementation so might leave as is for now.

ocaml/xapi/stream_vdi.ml
thomasmck: Discussed with Gabor/Stefano that only need to check it doesn't match the last value added to the accumulator rather than comparing against the whole list so will update.

ocaml/xapi/stream_vdi.ml
thomasmck: Will be reworking `check_exists` function so will see if using List.filter is valid with the update.

ocaml/xapi/stream_vdi.ml
gaborigloi: 2 spaces. We use the ocp-indent program to reindent files (but that might change other parts of the file too)

ocaml/xapi/stream_vdi.ml
gaborigloi: It's empty if either bit 0 or 1 is set: https://github.com/xapi-project/vhd-tool/blob/master/src/nbd_input.ml#L32
So we need to slightly change this

gaborigloi: And we can't use that `Nbd` function instead, because that will get merged later, it's in the nice to have epic.

ocaml/xapi/stream_vdi.ml
edwintorok: minor nitpick: ints are just 63 bit, the last bit is used to mark whether the value is an integer or a pointer to another block.

ocaml/xapi/stream_vdi.ml
thomasmck: @gaborigloi so Christian's suggestion is valid as there is only one options for non-empty?

gaborigloi: If either of these bits is set, the block is empty.
See NBD_STATE_HOLE and NBD_STATE_ZERO in the NBD protocol docs: https://github.com/NetworkBlockDevice/nbd/blob/master/doc/proto.md
So we need to check for the presence of these bits in the status_flag(s) integer

gaborigloi: I think we need to `or` it with `0b0011`, because if *either* of these bits is set, the extent is empty.

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,37 +100,48 @@ let write_block ~__context filename buffer ofd len =
...
 
+let flag_hole = 1l
+let flag_zero = 2l
+
 let cycle_descriptors descriptor_list offset =
-  let rec range a b =
-    if a > b then []
-    else a :: range (a + 1) b
+  let rec range acc a b =
+    if a > b then acc
```
gaborigloi: Do we need to reverse it with `List.rev`?

gaborigloi: Or we could loop from b down to a

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +100,65 @@ let write_block ~__context filename buffer ofd len =
...
+    | [] -> a::b
+    | _ -> begin
+      let b_rev = List.rev b in
+      if Int64.equal a (List.hd b_rev) then b
+      else List.rev (a::b_rev)
+      end
+  in
+
+  let rec add_unique_chunks acc = function
+    | [] -> acc
```
gaborigloi: I think it's more efficient to not do any `List.rev`s in `check_exists` but only do one here where we return `acc`

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +100,65 @@ let write_block ~__context filename buffer ofd len =
...
+        let start_chunk = Int64.div offset increment in
+        (* If the chunk ends at the boundary don't include the next chunk *)
+        let end_chunk = Int64.div (Int64.add (Int64.add descriptor.length  offset) Int64.minus_one ) increment in
+        let chunks = range [] start_chunk end_chunk in
+        chunks, descriptor.length
+        end
+      | true -> [], descriptor.length
+  in
+  (* Only compare to most recent chunk added*)
+  let check_exists a b =
```
gaborigloi: This could be renamed to something else to reflect the fact that it's also adding it to the list, not just checking if it's there.
Maybe like `add_new`

ocaml/xapi/stream_vdi.ml
thomasmck: ok will see how I can rework it. Had been trying to always keep stuff in order (i.e. [1;2;3...]) for simplicity but I think it results in using more revs than is strictly necessary.

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +100,64 @@ let write_block ~__context filename buffer ofd len =
...
+  status_flags : int32;
+}
+
+let flag_hole = 1l
+let flag_zero = 2l
+
+let cycle_descriptors descriptor_list offset =
+  let rec range acc a b =
+    if a < b then acc
+    else range (a::acc) (Int64.add a Int64.minus_one) b
```
mseri: You can simplify these: `Int64.(add a minus_one)`

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +100,64 @@ let write_block ~__context filename buffer ofd len =
...
+    (* We assume the destination is prezeroed, so we do not have to copy zeroed extents *)
+    (has_flag flag_hole) || (has_flag flag_zero)
+  in
+
+  let find_blocks descriptor offset increment =
+    match (is_empty descriptor) with
+      | false -> begin
+        let start_chunk = Int64.div offset increment in
+        (* If the chunk ends at the boundary don't include the next chunk *)
+        let end_chunk = Int64.div (Int64.add (Int64.add descriptor.length  offset) Int64.minus_one ) increment in
```
mseri: Here as well:
```ocaml
let open Int64 in div (add (add descriptor.length offset) minus_one ) increment
```

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +100,64 @@ let write_block ~__context filename buffer ofd len =
...
+        let chunks = range [] end_chunk start_chunk in
+        chunks, descriptor.length
+        end
+      | true -> [], descriptor.length
+  in
+  (* Only compare to most recent chunk added*)
+  let add_new a b =
+    match b with
+    | [] -> a::b
+    | _ -> begin
```
mseri: Why not 
```ocaml
| hd :: _ when Int64.equal a hd -> b
| _ -> a::b
```

ocaml/xapi/stream_vdi.ml
gaborigloi: Good idea, using local module opens: http://dev.realworldocaml.org/files-modules-and-programs.html#scrollNav-6

ocaml/xapi/stream_vdi.ml
gaborigloi: See http://dev.realworldocaml.org/lists-and-patterns.html#scrollNav-6 for an explanation of the when close: it can add extra conditions to a pattern

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +100,61 @@ let write_block ~__context filename buffer ofd len =
     then raise (Api_errors.Server_error (Api_errors.client_error, [ExnHelper.string_of_exn e]))
     else raise e
 
+type descriptor = {
+  length : int64;
+  status_flags : int32;
+}
+
+let flag_hole = 1l
+let flag_zero = 2l
```
gaborigloi: We could add something like this comment above these:
```
(* Flags for extents returned for the base:allocation metadata context, documented at https://github.com/NetworkBlockDevice/nbd/blob/master/doc/proto.md *)
```

lindig:
* It seems to me that `add_new` relies on undocumented assumptions. Why is it enough to compare only against the head of the list to avoid duplicates? If this cannot be guaranteed, I would recommend using a set data type.

* `process` could be expressed as a fold over descriptors but we can leave it as it is.

gaborigloi:
Add new relies on the assumption that we get the extents in the correct order, from left to right.
https://github.com/xapi-project/vhd-tool/blob/master/src/nbd_input.ml#L23-L25
We could document it as a precondition of the main function.

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +100,63 @@ let write_block ~__context filename buffer ofd len =
...
+type descriptor = {
+  length : int64;
+  status_flags : int32;
+}
+
+(* Flags for extents returned for the base:allocation NBD metadata context, documented at https://github.com/NetworkBlockDevice/nbd/blob/master/doc/proto.md *)
+let flag_hole = 1l
+let flag_zero = 2l
+
+let cycle_descriptors descriptor_list offset =
```
gaborigloi: We could put a comment above this function, something like this:
```
(** [descriptor_list] should be an list of non-overlapping extents, ordered from lowest offset to highest. *)
```

gaborigloi:
Thanks for the code, We can close this now, it's included in https://github.com/xapi-project/xen-api/pull/3734

PR #3733, opened 2018-10-11
CP-28675: Add a VM_SECURE_BOOT_FAILED message type

rosslagerwall:
This message is emitted by varstored on behalf of the guest when a
security violation occurs and Secure Boot is turned on.

Signed-off-by: Ross Lagerwall <ross.lagerwall@citrix.com>


edwintorok:
Perhaps worth adding a comment that this is raised by an external script so that people don't delete this as dead code, as done here: https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi-consts/api_messages.ml#L140

rosslagerwall:
The comment already says that. I can make it more explicit if you want.

PR #3734, opened 2018-10-15
CP-29605: Write function to query sparseness of VDIs on GFS2 SRs and do sparse data export

stefanopanella:
Try to use sparseness information in case the vdi to be exported in tar format has an associated NBD.
This will speed-up dramatically XVA VM export (tar) and vdi-export (tar) in case the VDI is scarcely written.

### DIFF
--- a/ocaml/xapi/stream_vdi.ml
+++ b/ocaml/xapi/stream_vdi.ml
@@ -30,6 +30,9 @@ open D
 (** Inside the tar we divide each VDI into small 'chunk_size' blocks: *)
 let chunk_size = Int64.mul 1024L 1024L (* 1 MiB *)
 
+(** Maximum range for sparseness query *)
+let max_sparseness_size = Int64.mul 10240L chunk_size
+
 let checksum_extension = ".checksum"
 
 type vdi = string (* directory prefix in tar file *) * API.ref_VDI * Int64.t (* size to send/recieve *)
@@ -41,7 +44,7 @@ let with_open_vdi __context rpc session_id vdi_ref mode flags perms f =
     (fun dom0_path ->
        debug "with_open_vdi opening: %s" dom0_path;
        let ofd = Unix.openfile dom0_path flags perms in
-       Pervasiveext.finally (fun () -> f ofd) (fun () -> Unix.close ofd))
+       Pervasiveext.finally (fun () -> f ofd dom0_path) (fun () -> Unix.close ofd))
 
 (** Used to sort VDI prefixes into a canonical order for streaming. Currently lexicographic
     sort on the externalised reference (used as a 'directory name') *)
@@ -100,6 +103,98 @@ let write_block ~__context filename buffer ofd len =
     then raise (Api_errors.Server_error (Api_errors.client_error, [ExnHelper.string_of_exn e]))
     else raise e
 
+let get_device_numbers path =
+  let rdev = (Unix.LargeFile.stat path).Unix.LargeFile.st_rdev in
+  let major = rdev / 256 and minor = rdev mod 256 in
+  (major, minor)
+
+let is_nbd_device path =
+  let nbd_device_num = 43 in
+  let (major, _) = get_device_numbers path in
+  major = nbd_device_num
+
+type nbd_connect_info =
+  { path : string
+  ; exportname : string
+  } [@@deriving rpc]
+
+let get_nbd_device path =
+  let nbd_device_prefix = "/dev/nbd" in
+  if Astring.String.is_prefix ~affix:nbd_device_prefix path && is_nbd_device path then begin
+    let nbd_number =
+      String.sub path (String.length nbd_device_prefix) (String.length path - String.length nbd_device_prefix)
+    in
+    let { path; exportname } =
+      let persistent_nbd_info_dir = "/var/run/nonpersistent/nbd" in
+      let filename = persistent_nbd_info_dir ^ "/" ^ nbd_number in
+      Xapi_stdext_unix.Unixext.string_of_file filename
+      |> Jsonrpc.of_string
+      |> nbd_connect_info_of_rpc
+    in
+    Some (path, exportname)
+  end else None
+
+type extent = {
+  flags : int32;
+  length : int64;
+} [@@deriving rpc]
+
+type extent_list = extent list [@@deriving rpc]
+
+(* Flags for extents returned for the base:allocation NBD metadata context, documented at https://github.com/NetworkBlockDevice/nbd/blob/master/doc/proto.md *)
+let flag_hole = 1l
+let flag_zero = 2l
+
+(** [descriptor_list] should be an list of non-overlapping extents, ordered from lowest offset to highest. *)
+let cycle_descriptors descriptor_list offset =
+  (* Output range includes start and end points *)
+  let rec range acc start_chunk end_chunk =
+    if end_chunk < start_chunk then acc
+    else range (end_chunk::acc) start_chunk Int64.(add end_chunk minus_one)
+  in
+
+  let is_empty e =
+    let has_flag flag =
+      Int32.logand e.flags flag = flag
+    in
+    (* We assume the destination is prezeroed, so we do not have to copy zeroed extents *)
+    (has_flag flag_hole) || (has_flag flag_zero)
+  in
+
+  let find_blocks descriptor offset increment =
+    match (is_empty descriptor) with
+      | false -> begin
+        let start_chunk = Int64.div offset increment in
+        (* If the chunk ends at the boundary don't include the next chunk *)
+        let end_chunk = let open Int64 in div (add (add descriptor.length offset) minus_one) increment in
+        let chunks = range [] start_chunk end_chunk in
+        chunks, descriptor.length
+        end
+      | true -> [], descriptor.length
+  in
+  (* Only compare to most recent chunk added*)
+  let add_new a b =
+    match b with
+    | hd :: _ when Int64.equal a hd -> b
+    | _ -> a::b
+  in
+
+  let rec add_unique_chunks acc = function
+    | [] -> List.rev acc
+    | x::xs -> add_unique_chunks (add_new x acc) xs
+  in
+
+  let rec process acc offset = function
+    | [] -> acc, offset
+    | x::xs -> begin
+      let chunks, add_offset = find_blocks x offset chunk_size in
+      process (add_unique_chunks (List.rev acc) chunks) (Int64.add offset add_offset) xs
+      end
+  in
+
+  let chunks, _ = process [] offset descriptor_list  in
+  chunks
+
 
 (** Stream a set of VDIs split into chunks in a tar format in a defined order. Return an
     association list mapping tar filename -> string (containing the SHA1 checksums) *)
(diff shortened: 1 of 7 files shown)


stefanopanella:
build error is not my fault :)

robhoes:
Hmmm:
```
[ERROR] The current version of OPAM cannot read the repository "xs-opam". 
        You should upgrade to at least version 2.0.
```

gaborigloi:
We need to use the `opam2` docker images in `.travis-*.sh`

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +103,98 @@ let write_block ~__context filename buffer ofd len =
...
+  let rec add_unique_chunks acc = function
+    | [] -> List.rev acc
+    | x::xs -> add_unique_chunks (add_new x acc) xs
+  in
+
+  let rec process acc offset = function
+    | [] -> acc, offset
+    | x::xs -> begin
+      let chunks, add_offset = find_blocks x offset chunk_size in
+      process (add_unique_chunks (List.rev acc) chunks) (Int64.add offset add_offset) xs
```
gaborigloi: Okay, actually I think the complexity of this is still quadratic in the number of extents, because we reverse the entire list `acc` every time we add a new element. I think there is a neat way in which we can get rid of all the List.revs - we just need to keep `acc` in decreasing order by not doing any List.revs on it
Then in the end we'll only have to do one List.rev on the final list.

ocaml/xapi/stream_vdi.ml
```diff
@@ -113,53 +208,95 @@ let send_all refresh_session ofd ~__context rpc session_id (prefix_vdis: vdi lis
...
+            write_block ~__context filename buffer ofd this_chunk_size;
+            made_progress __context progress (Int64.of_int this_chunk_size);
+          in
+          let rec stream_from_offset (offset: int64) =
+            let remaining = Int64.sub size offset in
+            if remaining > 0L then begin
+              let this_chunk_size = min (Int64.to_int chunk_size) (Int64.to_int remaining) in
+              let this_chunk_no = Int64.div offset chunk_size in
+              let now = Unix.gettimeofday () in
+              let time_since_transmission = now -. !last_transmission_time in
```
edwintorok: I think we prefer to use `Mtime_clock.count` to use a monotonic timer, gettimeofday can go backwards or have large jumps it is a wall clock, we want just relative durations.

ocaml/xapi/stream_vdi.ml
```diff
@@ -113,53 +208,95 @@ let send_all refresh_session ofd ~__context rpc session_id (prefix_vdis: vdi lis
...
+              if offset = 0L || remaining <= chunk_size || time_since_transmission > 5. then begin
+                actually_write_chunk (Int64.to_int this_chunk_no) this_chunk_size;
+                stream_from_offset (Int64.add offset (Int64.of_int this_chunk_size))
+              end else begin
+                let remaining = Int64.mul (Int64.div (Int64.sub remaining 1L) chunk_size) chunk_size in
+                (* Get sparseness for next 10GB or until the end rounded by last chunk, whichever is smaller *)
+                let sparseness_size = min max_sparseness_size remaining in
+                let output, _ = Forkhelpers.execute_command_get_output "/opt/xensource/libexec/get_nbd_extents.py" ["--path"; path; "--exportname"; exportname; "--offset"; Int64.to_string offset; "--length"; Int64.to_string sparseness_size] in
```
edwintorok: Might want to put the path into Xapi_globs. There is also a way for xapi to check for presence of the script on startup if you list it in required resources.

ocaml/xapi/stream_vdi.ml
```diff
@@ -113,53 +208,95 @@ let send_all refresh_session ofd ~__context rpc session_id (prefix_vdis: vdi lis
...
+                let extents = extent_list_of_rpc (Jsonrpc.of_string output) in
+                let chunks = cycle_descriptors extents offset in
+                List.iter (
+                  fun chunk ->
+                    begin
+                    actually_write_chunk (Int64.to_int chunk) (Int64.to_int chunk_size);
+                    end
```
edwintorok: begin/end is redundant here.

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +103,98 @@ let write_block ~__context filename buffer ofd len =
...
+
+  let rec process acc offset = function
+    | [] -> acc, offset
+    | x::xs -> begin
+      let chunks, add_offset = find_blocks x offset chunk_size in
+      process (add_unique_chunks (List.rev acc) chunks) (Int64.add offset add_offset) xs
+      end
+  in
+
+  let chunks, _ = process [] offset descriptor_list  in
```
edwintorok: Is the intent here to filter out zeroes and holes? Could we simply use a List.filter is_empty_descriptor instead?

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +103,98 @@ let write_block ~__context filename buffer ofd len =
...
+  } [@@deriving rpc]
+
+let get_nbd_device path =
+  let nbd_device_prefix = "/dev/nbd" in
+  if Astring.String.is_prefix ~affix:nbd_device_prefix path && is_nbd_device path then begin
+    let nbd_number =
+      String.sub path (String.length nbd_device_prefix) (String.length path - String.length nbd_device_prefix)
+    in
+    let { path; exportname } =
+      let persistent_nbd_info_dir = "/var/run/nonpersistent/nbd" in
```
edwintorok: Who writes this information?

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +103,98 @@ let write_block ~__context filename buffer ofd len =
     then raise (Api_errors.Server_error (Api_errors.client_error, [ExnHelper.string_of_exn e]))
     else raise e
 
+let get_device_numbers path =
+  let rdev = (Unix.LargeFile.stat path).Unix.LargeFile.st_rdev in
+  let major = rdev / 256 and minor = rdev mod 256 in
+  (major, minor)
+
+let is_nbd_device path =
+  let nbd_device_num = 43 in
```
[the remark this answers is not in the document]
gaborigloi: It's the major device number for NBD devices
https://www.kernel.org/doc/Documentation/admin-guide/devices.txt
Try `ls -l /dev/` and that will show the major & minor dev numbers

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +103,98 @@ let write_block ~__context filename buffer ofd len =
...
+type nbd_connect_info =
+  { path : string
+  ; exportname : string
+  } [@@deriving rpc]
+
+let get_nbd_device path =
+  let nbd_device_prefix = "/dev/nbd" in
+  if Astring.String.is_prefix ~affix:nbd_device_prefix path && is_nbd_device path then begin
+    let nbd_number =
+      String.sub path (String.length nbd_device_prefix) (String.length path - String.length nbd_device_prefix)
```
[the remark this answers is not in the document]
gaborigloi: Thanks, good idea! We could just drop in the code you suggested.

ocaml/xapi/stream_vdi.ml
edwintorok: It would be better to not copy/paste common functionality, but attempt to restructure the code to reuse the existing code instead, I haven't looked to deeply in what the existing code does and how much effort that would be.

stefanopanella: existing code is looping in a very different way because it can not rely on sparseness. It looks quite hard to refactor in a way that would not be even uglier. ALso as @gaborigloi said, this way we know that we are not introducing bugs to existing code because of this.

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +103,98 @@ let write_block ~__context filename buffer ofd len =
...
+    (* We assume the destination is prezeroed, so we do not have to copy zeroed extents *)
+    (has_flag flag_hole) || (has_flag flag_zero)
+  in
+
+  let find_blocks descriptor offset increment =
+    match (is_empty descriptor) with
+      | false -> begin
+        let start_chunk = Int64.div offset increment in
+        (* If the chunk ends at the boundary don't include the next chunk *)
+        let end_chunk = let open Int64 in div (add (add descriptor.length offset) minus_one) increment in
```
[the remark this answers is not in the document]
stefanopanella: OK

ocaml/xapi/stream_vdi.ml
stefanopanella: OK, there was a print that I have removed, I will change

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +103,98 @@ let write_block ~__context filename buffer ofd len =
...
+        let chunks = range [] start_chunk end_chunk in
+        chunks, descriptor.length
+        end
+      | true -> [], descriptor.length
+  in
+  (* Only compare to most recent chunk added*)
+  let add_new a b =
```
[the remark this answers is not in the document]
stefanopanella: like
let add_new acc b =
?

ocaml/xapi/stream_vdi.ml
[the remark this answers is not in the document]
stefanopanella: should I just change
`String.sub path (String.length nbd_device_prefix) (String.length path - String.length nbd_device_prefix)`
with
`Astring.String.with_range ~first:(String.length nbd_device_prefix) path`
?

ocaml/xapi/stream_vdi.ml
gaborigloi: Good point. The original code had the sime issue

ocaml/xapi/stream_vdi.ml
gaborigloi: https://github.com/xapi-project/xen-api/blob/master/scripts/nbd_client_manager.py
We should add it as a comment

ocaml/xapi/stream_vdi.ml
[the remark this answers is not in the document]
gaborigloi: Yes, I think that should work

ocaml/xapi/stream_vdi.ml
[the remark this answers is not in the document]
gaborigloi: I think so

ocaml/xapi/stream_vdi.ml
stefanopanella: mmm, not sure I follow this, what is Xapi_globs?

ocaml/xapi/stream_vdi.ml
gaborigloi: I think we could try to fix this mtime issue in our new code because it's a serious issue in my opinion, it could result in intermittent issues.

I agree with Stefano that for the reasons in my above comment, it's best to leave the old code unchanged for now.

Later R3 should fix the old one outside the scope of this req.

ocaml/xapi/stream_vdi.ml
gaborigloi: ocaml/xapi/xapi_globs.ml

ocaml/xapi/stream_vdi.ml
gaborigloi: There is, but I think it's not worth making so big a change now, because this code will be replaced eventually.

ocaml/xapi/stream_vdi.ml
stefanopanella: @edwintorok do you agree with @gaborigloi ? If yes I can do it that way.

ocaml/xapi/stream_vdi.ml
stefanopanella: OK i will do that

ocaml/xapi/stream_vdi.ml
stefanopanella: Should I leave this as is than?

ocaml/xapi/stream_vdi.ml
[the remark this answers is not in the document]
edwintorok: I think b was your accumulator in this function, but having the accumulator first makes more sense I agree.

ocaml/xapi/stream_vdi.ml
edwintorok: I think @gaborigloi mentioned a problem that this is currently quadratic, which could make this slow. If filtering is what you actually want to do, and your list of offsets is already sorted it might be easier to do that instead of building a lot of intermediate lists, it would simplify the code and make it smaller.

edwintorok: > It is, and then we split them into chunks, but I think it's not worth making so big a change now, because this code will be replaced eventually.

What will it get replaced with? Could we implement the replacement directly?

stefanopanella: we have already removed the quadratic part in last commit. Regarding changing this code, when every SR type will support CBT API we can use the same to query sparseness but we can not do that now.
@edwintorok Are you OK wit it? If not, can you tell me how should I modify the existing code?

stefanopanella:
Added a new commit to address all the issues.

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +103,105 @@ let write_block ~__context filename buffer ofd len =
...
+  (* This works in reverse order *)
+  let rec process acc offset = function
+    | [] -> acc
+    | x::xs ->
+      let chunks, add_offset = get_non_empty_chunks x offset chunk_size in
+      process (add_unique_chunks acc chunks) (Int64.add offset add_offset) xs
+  in
+
+  let chunks = process [] offset descriptor_list  in
+  List.rev chunks
```
edwintorok: The List.rev could be inside `process`, that way it would be self-contained.

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +103,105 @@ let write_block ~__context filename buffer ofd len =
...
+        let start_chunk = Int64.div offset increment in
+        (* If the chunk ends at the boundary don't include the next chunk *)
+        let end_chunk = Int64.(div (add (add descriptor.length offset) minus_one) increment) in
+        let chunks = range [] start_chunk end_chunk in
+        chunks, descriptor.length
+      | true -> [], descriptor.length
+  in
+
+  (* Only compare to most recent chunk added *)
+  let add_new acc b =
```
edwintorok: This could be named `add_new_unique`, see also below.

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +103,105 @@ let write_block ~__context filename buffer ofd len =
...
+    match acc with
+    | hd :: _ when Int64.equal b hd -> acc
+    | _ -> b::acc
+  in
+
+  (* Output decreasing chunk numbers, x should be increasing and acc should be decreasing *)
+  let rec add_unique_chunks acc = function
```
edwintorok: This is a hand-coded List.fold_left, see below

ocaml/xapi/stream_vdi.ml
```diff
@@ -100,6 +103,105 @@ let write_block ~__context filename buffer ofd len =
...
+    | [] -> acc
+    | x::xs -> add_unique_chunks (add_new acc x) xs
+  in
+
+  (* This works in reverse order *)
+  let rec process acc offset = function
+    | [] -> acc
+    | x::xs ->
+      let chunks, add_offset = get_non_empty_chunks x offset chunk_size in
+      process (add_unique_chunks acc chunks) (Int64.add offset add_offset) xs
```
edwintorok: I think this is equivalent to:
`process (List.fold_left add_new_unique acc chunks) (Int64.add offset add_offset)`.

I'm still thinking how else this could be simplified, it should be possible to implement this with fewer lines of code.

gaborigloi: Agreed - however I think it's not worth spending too much time making this PR perfect, because hopefully both branches of this code will be rewritten soon if/when we implement generic CBT for all SRs - in that case we'll use CBT instead of NBD and will have only one branch.

So if the above issues do not cause performance problems (e.g. quadratic complexity) or bugs in case of some corner cases, I think we should merge it?

edwintorok: I couldn't find a situation where this code would give incorrect results so far, although would've been nice to have unit tests.
Does the quicktest construct a lot of scenarios with holes/zeroes and other corner cases?

gaborigloi: Yes. It test empty VDI, full random VDI, random VDI with random number of random extents (random offset, length, data).

gaborigloi: Stefano has run it 100 times and it did not fail :)

ocaml/xapi/stream_vdi.ml
edwintorok: Should be equivalent to:
```ocaml
 let add_unique_chunks acc lst = List.fold_left add_new acc lst
```

PR #3737, opened 2018-10-15
CA-297297 Make create_row idempotent

rwdrich:
If a row already exists, and all of the fields passed in are the same,
we should do nothing and return successfully.

If a row already exists, and any of the fields are different to what was
passed in, we should fail.
 - For this I've created a new exception - Integrity_violation.

I've also updated the database tests to cover these cases

### DIFF
--- a/ocaml/database/db_cache_impl.ml
+++ b/ocaml/database/db_cache_impl.ml
@@ -170,14 +170,16 @@ let fld_check t tblname objref (fldname, value) =
   (v = value, fldname, v)
 
 let create_row t tblname kvs' new_objref =
-  if is_valid_ref t new_objref then
-      let uniq_check_list = List.map (fld_check t tblname new_objref) kvs' in
-      let failure_opt = List.find_opt (fun (tf, _, _) -> not tf) uniq_check_list in
-      match failure_opt with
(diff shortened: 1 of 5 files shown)


ocaml/database/db_cache_impl.ml
```diff
@@ -165,8 +165,19 @@ let create_row_locked t tblname kvs' new_objref =
...
 let create_row t tblname kvs' new_objref =
-  with_lock (fun () -> create_row_locked t tblname kvs' new_objref)
+  if is_valid_ref t new_objref then
+      let uniq_check_list = List.map (fld_check t tblname new_objref) kvs' in
+      let failure_opt = List.find_opt (fun (tf, _, _) -> not tf) uniq_check_list in
+      match failure_opt with
+      | Some (_, f, v) -> raise (Integrity_violation (tblname, f, v))
+      | _ -> ()
+  else
+    with_lock (fun () -> create_row_locked t tblname kvs' new_objref)
```
robhoes: We should probably move the lock around the whole thing, otherwise two threads could create rows for the same ref after the `is_valid_ref` check.

rwdrich:
I've cherry picked Stefano's commit across from his PR, to fix the travis build for this PR

PR #3738, opened 2018-10-16
CA-297343: Use transform_xenops_exn in pool_migrate

robhoes:
This translates more of the Xenops_interface errors into proper API errors,
like the other lifecycle functions inside Xapi_xenops do.

### DIFF
--- a/ocaml/xapi/xapi_vm_migrate.ml
+++ b/ocaml/xapi/xapi_vm_migrate.ml
@@ -273,9 +273,11 @@ let pool_migrate ~__context ~vm ~host ~options =
         Xapi_network.with_networks_attached_for_vm ~__context ~vm ~host (fun () ->
             (* XXX: PR-1255: the live flag *)
             info "xenops: VM.migrate %s to %s" vm_uuid xenops_url;
-            migrate_with_retry ~__context queue_name dbg vm_uuid [] [] xenops_vgpu_map xenops_url;
-            (* Delete all record of this VM locally (including caches) *)
-            Xapi_xenops.Xenopsd_metadata.delete ~__context vm_uuid;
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -294,16 +296,7 @@ let pool_migrate ~__context ~vm ~host ~options =
             end;
           with _ -> ()
         end;
-        match exn with
-        | Xenops_interface.Xenopsd_error Failed_to_acknowledge_shutdown_request ->
-          raise Api_errors.(Server_error (vm_failed_shutdown_ack, [Ref.string_of vm]))
-        | Xenops_interface.Xenopsd_error Cancelled _ ->
-          TaskHelper.raise_cancelled ~__context
-        | Xenops_interface.Xenopsd_error Storage_backend_error (code, _) ->
-          raise Api_errors.(Server_error (sr_backend_failure, [code]))
```
edwintorok: I checked the other exceptions and they are handled similarly by `transform_xenops_exn` (IIUC then Canceled is handled better because it looks up the id), but the storage exception is handled differently.
`transform_xenops_exn` would reraise using the `code`, and this code would reraise as `sr_backend_failure`. Should we update `transform_xenopsd_exn`?

robhoes: Hmm, indeed. @jonludlam, is SR_BACKEND_FAILURE the API error that we need for all cases `Xenopsd_error Storage_backend_error`?

jonludlam: I think actually `Storage_backend_error` is a wrapper for real `Api_error` errors, so the new code after this PR is doing it correctly.

PR #3739, opened 2018-10-16
CA-299944: Cross pool update apply failed because HTTP connection not enabled in CC mode

minli1:
Job 2288870 passed.

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -433,7 +433,10 @@ module Pool_update = struct
       ~doc:"Attach the pool update VDI"
       ~in_oss_since:None
       ~in_product_since:rel_ely
-      ~params:[ Ref _pool_update, "self", "The update to be attached"]
+      ~versioned_params:[
+        {param_type=Ref _pool_update; param_name="self"; param_doc="The update to be attached"; param_release=ely_release; param_default=None};
+        {param_type=Bool; param_name="use_localhost_proxy"; param_doc="Use the localhost proxy"; param_release=ely_release; param_default=Some (VBool false)};
+      ]
       ~result:(String, "The file URL of pool update")
       ~allowed_roles:_R_POOL_OP
       ()
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -435,7 +435,7 @@ module Pool_update = struct
       ~in_product_since:rel_ely
       ~versioned_params:[
         {param_type=Ref _pool_update; param_name="self"; param_doc="The update to be attached"; param_release=ely_release; param_default=None};
-        {param_type=Bool; param_name="use_localhost_proxy"; param_doc="Use the localhost proxy"; param_release=ely_release; param_default=Some (VBool false)};
+        {param_type=Bool; param_name="use_localhost_proxy"; param_doc="Use the localhost proxy"; param_release=naples_release; param_default=Some (VBool false)};
       ]
       ~result:(String, "The file URL of pool update")
       ~allowed_roles:_R_POOL_OP
--- a/ocaml/tests/test_pool_update.ml
+++ b/ocaml/tests/test_pool_update.ml
@@ -41,19 +41,20 @@ let test_assert_space_available () =
 
 let test_download_restriction () =
   Xapi_globs.host_update_dir := ".";
-  let assert_no_dots s =
+  let assert_no_dots (_, s) =
     Alcotest.(check bool)
       "test_download_restriction: path must not contain ."
       true
       (String.index_opt s '.' = None)
   in
+  let uuid = Helpers.get_localhost_uuid () in
   let test path =
     path
     |> Filename.concat Constants.get_pool_update_download_uri
-    |> Xapi_pool_update.path_from_uri
+    |> Xapi_pool_update.path_and_host_from_uri
     |> assert_no_dots
   in
-  List.iter test ["myfile"; ".."; "%2e%2e"]
+  List.map (fun s -> String.concat uuid [s] ) ["/myfile"; "/.."; "/%2e%2e"] |> List.iter test
 
 let test =
   [ "test_pool_update_destroy", `Quick, test_pool_update_destroy
--- a/ocaml/tests/test_pool_update.ml
+++ b/ocaml/tests/test_pool_update.ml
@@ -25,12 +25,11 @@ let test_pool_update_destroy () =
 let test_pool_update_refcount () =
   let assert_equal = Alcotest.(check int) "assertion called by test_pool_update_refcount" in
   let __context = Mock.make_context_with_new_db "Mock context" in
-  let uuid = Helpers.get_localhost_uuid () in
   let vdi = make_vdi ~__context ~virtual_size:4096L () in
-  Xapi_pool_update.with_inc_refcount ~__context ~uuid ~vdi (fun ~__context ~uuid ~vdi -> ());
-  Xapi_pool_update.with_inc_refcount ~__context ~uuid ~vdi (fun ~__context ~uuid ~vdi -> assert_equal 0 1);
-  Xapi_pool_update.with_dec_refcount ~__context ~uuid ~vdi (fun ~__context ~uuid ~vdi -> assert_equal 0 1);
-  Xapi_pool_update.with_dec_refcount ~__context ~uuid ~vdi (fun ~__context ~uuid ~vdi -> ())
+  Xapi_pool_update.with_inc_refcount ~__context ~uuid:"a" ~vdi (fun ~__context ~uuid ~vdi -> ());
+  Xapi_pool_update.with_inc_refcount ~__context ~uuid:"a" ~vdi (fun ~__context ~uuid ~vdi -> assert_equal 0 1);
+  Xapi_pool_update.with_dec_refcount ~__context ~uuid:"a" ~vdi (fun ~__context ~uuid ~vdi -> assert_equal 0 1);
+  Xapi_pool_update.with_dec_refcount ~__context ~uuid:"a" ~vdi (fun ~__context ~uuid ~vdi -> ())
 
 let test_assert_space_available () =
   let free_bytes = 1_000_000L in
@@ -47,14 +46,13 @@ let test_download_restriction () =
       true
       (String.index_opt s '.' = None)
   in
-  let uuid = Helpers.get_localhost_uuid () in
   let test path =
     path
     |> Filename.concat Constants.get_pool_update_download_uri
     |> Xapi_pool_update.path_and_host_from_uri
     |> assert_no_dots
   in
-  List.map (fun s -> String.concat uuid [s] ) ["/myfile"; "/.."; "/%2e%2e"] |> List.iter test
+  List.iter test ["/myfile"; "/.."; "/%2e%2e"]
 
 let test =
   [ "test_pool_update_destroy", `Quick, test_pool_update_destroy
--- a/ocaml/xapi/xapi_pool_update.ml
+++ b/ocaml/xapi/xapi_pool_update.ml
@@ -543,16 +543,24 @@ let path_and_host_from_uri uri =
       |> Stdext.Unixext.resolve_dot_and_dotdot
     in
     (host,resolved_path)
-  | _ -> failwith "Expecting "
+  | _ -> failwith "Expecting both host and path in the uri"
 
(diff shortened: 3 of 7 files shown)


ocaml/xapi/xapi_pool_update.ml
```diff
@@ -532,21 +532,42 @@ let resync_host ~__context ~host =
...
   debug "pool_update.pool_update_download_handler %s" filepath;
-  if not(String.startswith !Xapi_globs.host_update_dir filepath) || not (Sys.file_exists filepath) then begin
-    debug "Rejecting request for file: %s (outside of or not existed in directory %s)" filepath !Xapi_globs.host_update_dir;
-    Http_svr.response_forbidden ~req s
-  end else begin
-    Fileserver.response_file s filepath;
-    req.Request.close <- true
-  end
+  if host_uuid <> localhost_uuid then
+    proxy_request req s host_uuid
```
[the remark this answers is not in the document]
minli1: I think the `host_uuid` comes from `Db.Host.get_uuid ~__context ~self:host`.
And for safety, we can check the uuid to be a valid host uuid here. @jonludlam

ocaml/idl/datamodel.ml
```diff
@@ -433,7 +433,10 @@ module Pool_update = struct
       ~doc:"Attach the pool update VDI"
       ~in_oss_since:None
       ~in_product_since:rel_ely
-      ~params:[ Ref _pool_update, "self", "The update to be attached"]
+      ~versioned_params:[
+        {param_type=Ref _pool_update; param_name="self"; param_doc="The update to be attached"; param_release=ely_release; param_default=None};
+        {param_type=Bool; param_name="use_localhost_proxy"; param_doc="Use the localhost proxy"; param_release=ely_release; param_default=Some (VBool false)};
```
edwintorok: Is `ely_release` correct here, or should we specify naples_release here and havana_release in the backport?

ocaml/xapi/xapi_pool_update.ml
```diff
@@ -532,21 +532,42 @@ let resync_host ~__context ~host =
...
+  let host_and_path = String.sub_to_end uri (String.length Constants.get_pool_update_download_uri) in
+  match String.split ~limit:2 '/' host_and_path with
+  | [host; untrusted_path] ->
+    let resolved_path = untrusted_path
+      |> Filename.concat !Xapi_globs.host_update_dir
+      |> Uri.pct_decode
+      |> Stdext.Unixext.resolve_dot_and_dotdot
+    in
+    (host,resolved_path)
+  | _ -> failwith "Expecting "
```
edwintorok: Did you mean to say what you're expecting here?

ocaml/xapi/xapi_pool_update.ml
```diff
@@ -532,21 +532,42 @@ let resync_host ~__context ~host =
...
+
+let proxy_request req s host_uuid =
+  Server_helpers.exec_with_new_task "pool_update_proxy_request" (fun __context ->
+    let host = Db.Host.get_by_uuid ~__context ~uuid:host_uuid in
+    let ip = Db.Host.get_address ~__context ~self:host in
+    let transport = Xmlrpc_client.(SSL(SSL.make (), ip, 443)) in
```
edwintorok: I assume we do not intend to use SNI, and we do not need to send the actual hostname along.

ocaml/xapi/xapi_pool_update.ml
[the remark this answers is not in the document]
edwintorok: When a XenServer slave sends it then it is valid, but this is an unauthenticated API, anyone can send anything here, and we should be robust against that. If we raise an exception when we detect the invalid host uuid that should be enough though.

ocaml/xapi/xapi_pool_update.ml
alexbrett: It's not SNI we'd need to worry about, but certificate validation - we currently require the certificate to have a SAN of the host IP, as lots of things connect purely by IP (in fact I'm not sure we even have the FQDN available to us), so this should be fine here.

ocaml/xapi/xapi_pool_update.ml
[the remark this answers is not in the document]
krizex: We already have it in `proxy_request`: `let host = Db.Host.get_by_uuid ~__context ~uuid:host_uuid in`

ocaml/xapi/xapi_pool_update.ml
```diff
@@ -532,21 +532,42 @@ let resync_host ~__context ~host =
...
+    Xmlrpc_client.with_transport transport (fun fd ->
+      Unixext.really_write_string fd (Http.Request.to_wire_string req);
+      Unixext.proxy s fd)
```
krizex: We have to mark the request closed here by `req.Request.close <- true` so Http server will close the socket, otherwise the sock will be left open until client close the other peer.

ocaml/xapi/xapi_pool_update.ml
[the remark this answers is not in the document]
krizex: Step further, is it possible to restrict the proxy only used for the request comes from localhost? Aka the client outside the localhost cannot use this proxy. It will be safer.

ocaml/xapi/xapi_pool_update.ml
minli1: Yes, will write more of `excepting` here.

ocaml/xapi/xapi_pool_update.ml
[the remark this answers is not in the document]
jonludlam: Unfortunately not, since the external stunnel reconnects over localhost, hence xapi can't tell the difference between a call that really honestly originated from yum running on localhost and stunnel running on localhost.

In any case, what will happen is that you'll get an exception raised by the `Db.Host.get_by_uuid`. The only thing we need to do is verify that at least the fd is closed, and preferably we send a reasonable HTTP error before we close it.

ocaml/idl/datamodel.ml
minli1: You are right, but I don't find anywhere define `havana_release` in 1.14-lcm which is havana branch. Do we change the mechanism for the release definition?

ocaml/xapi/xapi_pool_update.ml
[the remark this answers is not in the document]
minli1: Hi, all. I have updated the code.
Trying to catch the exception for `Db.Host.get_by_uuid`, if invalid then returns a bad http request, finally close the fd.
@jonludlam

minli1:
I have refined the comments, mainly trying to handle the invalid host uuid scenario.
So ask help again to review it. Thanks. @jonludlam @edwintorok @krizex @alexbrett

ocaml/xapi/xapi_pool_update.ml
```diff
@@ -532,21 +532,50 @@ let resync_host ~__context ~host =
...
+      |> Uri.pct_decode
+      |> Stdext.Unixext.resolve_dot_and_dotdot
+    in
+    (host,resolved_path)
+  | _ -> failwith "Expecting both host and path in the uri"
+
+let proxy_request req s host_uuid =
+  Server_helpers.exec_with_new_task "pool_update_proxy_request" (fun __context ->
+    let host =
+      try Some (Db.Host.get_by_uuid ~__context ~uuid:host_uuid) with _ -> None
```
alexbrett: I don't know enough about xapi here, so this is probably not an issue at all, but is `get_by_uuid` safe to untrusted data, i.e. is there any possibility an attacker could encode something in it that could lead to a SQL injection (or equivalent, as I know we aren't actually using a SQL DB)?

edwintorok: An `xe host*` API would also at some point do a get_by_uuid, all it does is a lookup in a map/list matching the uuid exactly as it is, so this API is safe against injection.
There is another API get_all_records_where that allows specifying a filter string, but even if you manage to inject it that expression is interpreted by XAPI and all you can do is match on fields and perform boolean operations, it is never given to a C API call, or some DB.

ocaml/tests/test_pool_update.ml
```diff
@@ -41,19 +41,20 @@ let test_assert_space_available () =
...
 let test_download_restriction () =
   Xapi_globs.host_update_dir := ".";
-  let assert_no_dots s =
+  let assert_no_dots (_, s) =
     Alcotest.(check bool)
       "test_download_restriction: path must not contain ."
       true
       (String.index_opt s '.' = None)
   in
+  let uuid = Helpers.get_localhost_uuid () in
```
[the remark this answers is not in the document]
minli1: You are right, Rob. Probably `get_localhost_uuid ` returns blank and it still works below. But I found another place in the file that use it too, so I refine it with a `fake` uuid there to avoid confusion. Thanks.
https://github.com/xapi-project/xen-api/blame/master/ocaml/tests/test_pool_update.ml#L28

ocaml/tests/test_pool_update.ml
```diff
@@ -41,19 +41,20 @@ let test_assert_space_available () =
...
   let test path =
     path
     |> Filename.concat Constants.get_pool_update_download_uri
-    |> Xapi_pool_update.path_from_uri
+    |> Xapi_pool_update.path_and_host_from_uri
     |> assert_no_dots
   in
-  List.iter test ["myfile"; ".."; "%2e%2e"]
+  List.map (fun s -> String.concat uuid [s] ) ["/myfile"; "/.."; "/%2e%2e"] |> List.iter test
```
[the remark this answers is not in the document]
minli1: Drop the UUID here, thanks. 👍

ocaml/idl/datamodel.ml
krizex: Please resolve this when you changing it to `naples_release`

ocaml/xapi/xapi_pool_update.ml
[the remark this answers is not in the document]
krizex: There is already an exception handler at https://github.com/xapi-project/xen-api-libs-transitional/blob/master/http-svr/http_svr.ml#L457 and for this case, the code will send a HTTP 500 error and close the socket automatically.
IMO the `proxy_request` is fine, we don't need to handle the exception, just leave the work to upper layer.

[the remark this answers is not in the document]
minli1: I agree with @krizex. And I think there has a very low probability to have a bad uuid here, the exception will handle by upper layer too. @robhoes  Needs your review and confirm. Thank you.

ocaml/xapi/xapi_pool_update.ml
```diff
@@ -532,21 +532,50 @@ let resync_host ~__context ~host =
...
+    |Some host ->
+      let ip = Db.Host.get_address ~__context ~self:host in
+      let transport = Xmlrpc_client.(SSL(SSL.make (), ip, 443)) in
+      Xmlrpc_client.with_transport transport (fun fd ->
+        Unixext.really_write_string fd (Http.Request.to_wire_string req);
+        Unixext.proxy s fd)
+    |None ->
+      debug "Caught exception while get Host by uuid %s" host_uuid;
+      Http_svr.response_badrequest ~req s;
+    req.Request.close <- true
```
robhoes: In spite of the indentation, this line is part of the `| None ->` case (only). I think that rather than having it here, we should put this line near the top of `pool_update_download_handler`, because we always want to handle just one request, not matter what the response will be, so we never want to keep the connection alive after the response.

Even before this patch, we don't actually close the connection on the path that calls `Http_svr.response_forbidden`, which looks wrong to me. Moving the line to the top would fix that as well.

Searching for `close <- true` reveals that many similar handlers do it this way.

jonludlam: For both 'forbidden' and 'badrequest' we instruct the client to close the connection, which should cause us to close too (see https://github.com/xapi-project/xen-api-libs-transitional/blob/master/http-svr/http.ml#L30-L33 and https://github.com/xapi-project/xen-api-libs-transitional/blob/master/http-svr/http.ml#L57-L60). However, it's good form for us to close proactively so we should set `req.Request.close` to true in all cases, as @robhoes suggests.

jonludlam: Note that the we may end up handling multiple requests on the one connection - since after the initial headers we proxy all the traffic, any subsequent requests will also be proxied, until either the client or remote server closes the connection.

PR #3743, opened 2018-10-19
CA-300210: Add 'CIPHER_SERVER_PREFERENCE' option in xapi ssl config file.

minli1:
To meet the requirement, add `CIPHER_SERVER_PREFERENCE` to let the server use the preference of cipher. 
Test passed. 


robhoes:
@edwintorok says that this is good :)

PR #3745, opened 2018-10-22
CA-297137: Don't update current_domain_type if xenopsd returns undefined

robhoes:
Under some circumstances, for example if a domain is being destroyed, a VM.stat
call to xenopsd may return an undefined domain type if the domain is not
halted. We assume that this is always a transient state, and ignore it.


robhoes:
Yep, I agree...

PR #3747, opened 2018-10-23
Increase VDI size for metadata backups

dermotctx:
Increase VDI size from 250 to 500MiB due to it running out space during metadata backups.


robhoes:
Thanks Dermot! Could you please add a Signed-off-by line to the commit message? See https://wiki.xenproject.org/wiki/Submitting_Xen_Project_Patches#Signing_off_a_patch.

PR #3749, opened 2018-10-25
CA-300115 lower VM.assert_operation_valid permissions

lindig:
Set the permissions for VN.assert_operation_valid to READ_ONLY as this
call doesn't mutate anything and does not reveal critical information.


lindig:
I've just fixed a typo in the commit message.

PR #3750, opened 2018-10-26
CP-29698 expunge Xenctrl.with_intf

lindig:
Xenctrl.with_intf is deprecated. The local replacement opens a
connection only once and remembers it.

### DIFF
--- a/ocaml/xapi/xenctrlx.ml
+++ b/ocaml/xapi/xenctrlx.ml
@@ -1,0 +1,27 @@
+
+module D = Debug.Make(struct let name = "xenctrlx" end)
+
+let handle = ref None
+
+let with_intf f =
+  match !handle with
+    | Some xc -> f xc
+    | None ->
+      let xc =
+        try
+          Xenctrl.interface_open ()
+        with
+        | e ->
+            let msg = Printexc.to_string e in
+            failwith @@ "failed to open xenctrl interface: "^msg
+      in
+        handle := Some (xc);
+        f xc
+ 
+let close () =
+  match !handle with
+  | Some xc -> Xenctrl.interface_close xc
+  | None    -> ()
+
+let () = at_exit close
+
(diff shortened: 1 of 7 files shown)


lindig:
This passed a BVT. There are more daemons that have to expunge `Xenctrl.with_intf` - so I'd like to discuss whether this is the right approach.

edwintorok:
Can the interface be open on startup? Might simplify things a bit if it can just refer to the global Xenctrlx.interface that is always open. Although that would make error reporting more difficult (if the open fails), and using globals has to be done carefully if we're multithreaded, like XAPI is.

ocaml/xapi/xenctrlx.ml
```diff
@@ -0,0 +1,27 @@
...
+    | None ->
+      let xc =
+        try
+          Xenctrl.interface_open ()
+        with
+        | e ->
+            let msg = Printexc.to_string e in
+            failwith @@ "failed to open xenctrl interface: "^msg
+      in
+        handle := Some (xc);
```
mseri: Why not making the open lazy, and Lazy.force it in with_intf? That should also be thread safe right?

mseri: But then it would not be nice for close

mseri: Oh, no, it would be ok, you could use Lazy.is_val

lindig: I believe `Xenctrl.with_intf` should implement this to avoid that every client has to do it - I will propose it. The problem there is that we have low-level `open/close` but also `with_intf` which maintains a global handle. A client should not use both methods (although it would not be dangerous).

lindig:
I believe this is the wrong approach and `with_intf` should be provided again by `Xenctrl` (which is part of Xen).

lindig:
Closing as I like to propose the approach mentioned above.

PR #3751, opened 2018-10-29
CP-29857: defer setting device-model until VM.start

edwintorok:
We want to be able to choose BIOS/UEFI boot mode from XenCenter, so the device-model should not be part of the template.
Did a basic smoke test on this, and will run more tests.


edwintorok:
Thanks, we've run a nightly (still being triaged), which should hopefully highlight any problems caused by this.

PR #3754, opened 2018-10-30
CA-290024: Reject booting pv-iommu VMs on a host where the premap is yet to complete

zli:

### DIFF
--- a/ocaml/xapi/xapi_host_helpers.ml
+++ b/ocaml/xapi/xapi_host_helpers.ml
@@ -204,6 +204,21 @@ let assert_startup_complete () =
     (fun () -> if not (!startup_complete) then
         raise (Api_errors.Server_error (Api_errors.host_still_booting, [])))
 
+let assert_host_pviommu_ready =
+  let is_ready = ref false in
+  fun () ->
+    (if not !is_ready then
+       try
+         let s = Unixext.string_of_file "/sys/kernel/pv_iommu_ready" in
+         is_ready := Scanf.sscanf s " %d " (=) 1;
+       with _ ->
+         (* If we had difficulty reading/converting the key (e.g. the key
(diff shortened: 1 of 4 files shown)


robhoes:
So this appears to just make the VM.start fail if pv-iommu is not yet ready, rather than causing the call to wait/block until it is ready. This may be the right thing to do, but the commit message is a little misleading.

zli:
> So this appears to just make the VM.start fail if pv-iommu is not yet ready, rather than causing the call to wait/block until it is ready. 

Yes, this is to follow Bob's suggest on the ticket. 

> This may be the right thing to do, but the commit message is a little misleading.

Commit message is updated.

ocaml/xapi/xapi_host_helpers.ml
```diff
@@ -204,6 +204,22 @@ let assert_startup_complete () =
...
         raise (Api_errors.Server_error (Api_errors.host_still_booting, [])))
 
+let assert_host_pviommu_ready =
+  let is_ready = ref false in
+  fun () ->
+    (if not !is_ready then
+       try
+         let s = Unixext.string_of_file "/sys/kernel/pv_iommu_ready" in
+         let s = Xstringext.String.strip Xstringext.String.isspace s in
+         is_ready := int_of_string s = 1;
```
lindig: I think this could be done simpler:
```
let is_ready =  
  try 
    let s = Unixext.string_of_file "/sys/kernel/pv_iommu_ready" in
    Scanf.sscanf " %d " (fun d -> d=1)  
  with _ -> true
```

Note that this will scan strings like "1" or " 1" successfully.

zli: Updated with scanf function suggested. 
Though the mutable variable `is_ready` is on purpose, to avoid re-reading/re-parsing once it already becomes `true`.

PR #3755, opened 2018-11-05
CA-294900 remove network_sriov on network reset

lindig:
During a host network reset, remove any network_sriov object that refers
to a local PIF.

### DIFF
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -1527,7 +1527,17 @@ let reset_networking ~__context ~host =
                 Db.PIF_metrics.destroy ~__context ~self:metrics
               end;
               Db.PIF.destroy ~__context ~self;
-            ) local_pifs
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_host.ml
```diff
@@ -1527,7 +1527,15 @@ let reset_networking ~__context ~host =
                 Db.PIF_metrics.destroy ~__context ~self:metrics
               end;
               Db.PIF.destroy ~__context ~self;
-            ) local_pifs
+  ) local_pifs;
+  List.iter (fun self ->
```
krizex: Why not follow the styles above that cleanup the `bonds`, `vlans` and `tunnels`? It could make the code easier to understand and maintain.

lindig: I don't think that the existing code is particularly clear but I'm happy to change it.

PR #3758, opened 2018-11-06
CA-298465: Only try to detach locally attached updates when booting

zli:

### DIFF
--- a/ocaml/xapi/xapi_pool_update.ml
+++ b/ocaml/xapi/xapi_pool_update.ml
@@ -110,6 +110,13 @@ let with_inc_refcount ~__context ~uuid ~vdi f =
         Hashtbl.replace updates_to_attach_count_tbl uuid (count + 1)
     )
 
+let get_locally_attached ~__context ~uuid ~vdi =
+    let mount_dir = Filename.concat !Xapi_globs.host_update_dir uuid in
+    let mount_dir_opt = if Sys.file_exists mount_dir then Some mount_dir else None in
+    let dom0 = Helpers.get_domain_zero ~__context in
+    let vbds = List.filter (fun self -> Db.VBD.get_VM ~__context ~self = dom0) (Db.VDI.get_VBDs ~__context ~self:vdi) in 
+    (mount_dir_opt, vbds)
+
 let detach_helper ~__context ~uuid ~vdi =
   with_dec_refcount ~__context ~uuid ~vdi
     (fun ~__context ~uuid ~vdi ->
@@ -453,9 +460,11 @@ let destroy ~__context ~self =
 let detach_attached_updates __context =
   Db.Pool_update.get_all ~__context |>
   List.iter ( fun self ->
-      Helpers.log_exn_continue ("detach_attached_updates: update_uuid " ^ (Db.Pool_update.get_uuid ~__context ~self))
-        (fun () -> ignore(Helpers.call_api_functions ~__context
-                            (fun rpc session_id -> Client.Pool_update.detach ~rpc ~session_id ~self))) ()
+      let uuid = Db.Pool_update.get_uuid ~__context ~self in
+      let vdi = Db.Pool_update.get_vdi ~__context ~self in
+      match get_locally_attached ~__context ~uuid ~vdi with
+      | None, [] -> ()
+      | _ -> Helpers.log_exn_continue ("detach_attached_updates: update_uuid " ^ uuid) (fun () -> detach_helper ~__context ~uuid ~vdi) ()
     )
 
 let resync_host ~__context ~host =


ocaml/xapi/xapi_pool_update.ml
```diff
@@ -453,9 +458,11 @@ let destroy ~__context ~self =
 let detach_attached_updates __context =
   Db.Pool_update.get_all ~__context |>
   List.iter ( fun self ->
-      Helpers.log_exn_continue ("detach_attached_updates: update_uuid " ^ (Db.Pool_update.get_uuid ~__context ~self))
-        (fun () -> ignore(Helpers.call_api_functions ~__context
-                            (fun rpc session_id -> Client.Pool_update.detach ~rpc ~session_id ~self))) ()
+      let uuid = Db.Pool_update.get_uuid ~__context ~self in
```
minli1: We request `uuid` and `vdi` here, but in `detach` function, we request these two again. I am wondering if we can improve them by not requesting Db action too much.

ocaml/xapi/xapi_pool_update.ml
```diff
@@ -110,35 +110,40 @@ let with_inc_refcount ~__context ~uuid ~vdi f =
...
+             Helpers.log_exn_continue ("detach_helper: unmounting " ^ mount_point)
+               (fun () -> umount mount_point) ()
+           end;
+           Helpers.log_exn_continue ("pool_update.detach_helper: rm " ^ mount_point)
+             (fun () ->
+                let output, _ = Forkhelpers.execute_command_get_output "/bin/rm" ["-r"; mount_point] in
+                debug "pool_update.detach_helper Mountpoint removed (output=%s)" output) ();
+           Helpers.log_exn_continue ("pool_update.detach_helper: rmdir " ^ mount_point_parent_dir)
+             (fun () ->
+                let output, _ = Forkhelpers.execute_command_get_output "/bin/rmdir" ["--ignore-fail-on-non-empty"; mount_point_parent_dir] in
```
edwintorok: Why not use `Unix.rmdir` here (wrapped in a try with if you want to ignore non-empty errors)

ocaml/xapi/xapi_pool_update.ml
zli: Updated.

ocaml/xapi/xapi_pool_update.ml
zli: This part of code was simply shifted around due to code refactoring, and I didn't try to change the way it was implemented.

Regarding the reason not using `Unix.rmdir` instead, I guess it was just for dealing with non-empty directories (which `Unix.rmdir` doesn't do).

ocaml/xapi/xapi_pool_update.ml
minli1: 👍

ocaml/xapi/xapi_pool_update.ml
```diff
@@ -110,35 +110,40 @@ let with_inc_refcount ~__context ~uuid ~vdi f =
         Hashtbl.replace updates_to_attach_count_tbl uuid (count + 1)
     )
 
+let get_locally_attached ~__context ~uuid ~vdi =
+    let mount_dir = String.concat "/" [!Xapi_globs.host_update_dir; uuid] in
```
lindig: Could use `File.concat` instead.

ocaml/xapi/xapi_pool_update.ml
```diff
@@ -110,35 +110,40 @@ let with_inc_refcount ~__context ~uuid ~vdi f =
...
+    let mount_dir_opt = if Sys.file_exists mount_dir then Some mount_dir else None in
```
lindig: This could test for `Sys.is_directory` as well and avoid the test later.

ocaml/xapi/xapi_pool_update.ml
lindig: Code to remove a directory recursively:
```
  let rmtree path =
    let (//) = Filename.concat in
    let rec rm path =
      let st = Unix.lstat path in
      match st.Unix.st_kind with
      | Unix.S_DIR ->
        Sys.readdir path |> Array.iter (fun file -> rm (path // file));
        Unix.rmdir path
      | _ -> Unix.unlink path
    in try rm path with
    | exn ->
      let exn' = Printexc.to_string exn in
      let msg  = Printf.sprintf "failed to remove %s: %s" path exn' in
      failwith msg
```

ocaml/xapi/xapi_pool_update.ml
zli: Update.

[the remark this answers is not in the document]
zli: Updated.

ocaml/xapi/xapi_pool_update.ml
zli: It would be nicer if we can find similar function from a more established/tested library instead of having to cook our own, otherwise sticking with shell calling out is not necessarily worse, as any potential corner cases have been dealt with `rm`.

Anyway, I've now left `detach_helper` completely untouched, as the refactoring seemed to have caused more confusion than benefits. Now, the semantic of `detach_helper` is just going ahead straightly and do the necessity same as before, leaving the judgement on any preconditions to its caller, which includes `detach`, `extract_update_info`, `detach_attached_updates` etc.

ocaml/xapi/xapi_pool_update.ml
zli: The test inside `detach_helper` was on the `vdi` sub-directory, this was on the parent directory. My previous patch did avoid re-testing the same directory.

But anyway, with the new patch, I've now left `detach_helper` completely untouched, as the refactoring seemed to have caused more confusion than benefits. Now, the semantic of `detach_helper` is just going ahead straightly and do the necessity same as before, leaving the judgement on any preconditions to its caller, which includes `detach`, `extract_update_info`, `detach_attached_updates` etc.

ocaml/xapi/xapi_pool_update.ml
```diff
@@ -110,35 +110,40 @@ let with_inc_refcount ~__context ~uuid ~vdi f =
...
   with_dec_refcount ~__context ~uuid ~vdi
     (fun ~__context ~uuid ~vdi ->
-       let mount_point_parent_dir = String.concat "/" [!Xapi_globs.host_update_dir; uuid] in
-       let mount_point = Filename.concat mount_point_parent_dir "vdi" in
-       debug "pool_update.detach_helper %s from %s" uuid mount_point;
-       if try Sys.is_directory mount_point with _ -> false then begin
-         Helpers.log_exn_continue ("detach_helper: unmounting " ^ mount_point)
-           (fun () -> umount mount_point) ()
-       end;
+       let mount_dir_opt, vbds = get_locally_attached ~__context ~uuid ~vdi in
```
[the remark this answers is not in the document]
zli: `detach_helper` was called by several  callers, which demands more refactoring in the end.

Anyway, I've now left `detach_helper` completely untouched, as the refactoring seemed to have caused more confusion than benefits. Now, the semantic of `detach_helper` is just going ahead straightly and do the necessity same as before, leaving the judgement on any preconditions to its caller, which includes `detach`, `extract_update_info`, `detach_attached_updates` etc.

PR #3759, opened 2018-11-13
CP-30002: xe vm-add-tags/vm-remove-tags

edwintorok:
Xapi has an API to add/remove VM tags that XenCenter uses, but there was no CLI equivalent.
I used this a while ago when doing some large scale boot tests and wanted to group my VMs from the commandline.


robhoes:
You can already do this with `xe vm-param-add uuid=<uuid> param-name=tags param-key=...`, and `xe vm-param-remove`.

PR #3760, opened 2018-11-13
CA-300103 Avoid duplicated tools SRs on pool join

lindig:
This is work in progress. On joining a pool, the SR for the Guest Tools ISO is a special case. This series of commits tries to avoid that multiple such SRs are present in the final pool.

### DIFF
--- a/ocaml/xapi/dbsync_master.ml
+++ b/ocaml/xapi/dbsync_master.ml
@@ -131,26 +131,35 @@ let create_tools_sr __context name_label name_description sr_introduce maybe_cre
     (Xapi_globs.i18n_original_value_prefix ^ "name_label"), name_label;
     (Xapi_globs.i18n_original_value_prefix ^ "name_description"), name_description
   ] in
+  let destroy self =
+    try Db.SR.destroy ~__context ~self with e ->
+      warn "failed to destroy redundant tools SR %s: %s"
+        (Ref.string_of self) (Printexc.to_string e) in
   let sr =
-    let tools_srs = List.filter (fun self -> Db.SR.get_is_tools_sr ~__context ~self) (Db.SR.get_all ~__context) in
+    let srs = Db.SR.get_all ~__context in
+    let tools_srs =
+      List.filter (fun self -> Db.SR.get_is_tools_sr ~__context ~self) srs in
+    let old_srs =
+      List.filter (fun self ->
+          let other_config = Db.SR.get_other_config ~__context ~self in
+          (Db.SR.get_is_tools_sr ~__context ~self = false) &&
+          ((List.mem_assoc Xapi_globs.tools_sr_tag other_config) ||
+           (List.mem_assoc Xapi_globs.xensource_internal other_config))
+        ) srs in
     match tools_srs with
     | sr :: others ->
       (* Let there be only one Tools SR *)
-      List.iter (fun self -> Db.SR.destroy ~__context ~self) others;
+      List.iter destroy others;
+      List.iter destroy old_srs;
       sr
     | [] ->
       (* First check if there is an SR with the old tags on it, which needs upgrading (set is_tools_sr). *)
       (* We cannot do this in xapi_db_upgrade, because that runs later. *)
-      let old_srs =
-        List.filter (fun self ->
-            let other_config = Db.SR.get_other_config ~__context ~self in
-            (List.mem_assoc Xapi_globs.tools_sr_tag other_config) ||
-            (List.mem_assoc Xapi_globs.xensource_internal other_config)
-          ) (Db.SR.get_all ~__context)
-      in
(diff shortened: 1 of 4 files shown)


ocaml/xapi/xapi_pool.ml
```diff
@@ -554,57 +554,39 @@ let rec create_or_get_host_on_master __context rpc session_id (host_ref, host) :
...
-          pbds_on_master
+  try
+    Client.SR.get_by_uuid ~rpc ~session_id ~uuid:my_uuid
+  with _ ->
+    if sr.API.sR_is_tools_sr then
+      (* find the tools ISO and return it *)
+      try
+        Client.SR.get_all_records ~rpc ~session_id
+        |> List.find (fun (_, sr) -> sr.API.sR_is_tools_sr)
+        |> (fun (ref,_) -> ref)
```
edwintorok: Should we check that the device config is the same and what should happen if it isn't?
What would happen if you start a VM with the tools ISO plugged in, and migrate it to a host which has a newer tools SR? (e.g. during an RPU)
If the contents of the tools ISO is different then I think the SRs should have different UUIDs to avoid accidentally changing the contents of a plugged ISO in a guest.

lindig: This is a good point. Should we scan for VMs that that a tools ISO attached?

robhoes: The host that is joining a pool cannot have any running VMs (another pre-join check further up in this file), so this should not be a problem. This function is not related to RPU either.

ocaml/xapi/xapi_pool.ml
```diff
@@ -554,57 +554,39 @@ let rec create_or_get_host_on_master __context rpc session_id (host_ref, host) :
...
-        List.exists
-          (fun (_, x) ->
-            Listext.List.set_equiv x.API.pBD_device_config
-              my_pbd.API.pBD_device_config )
-          pbds_on_master
+  try
+    Client.SR.get_by_uuid ~rpc ~session_id ~uuid:my_uuid
+  with _ ->
+    if sr.API.sR_is_tools_sr then
+      (* find the tools ISO and return it *)
```
robhoes: "tools SR"

ocaml/xapi/xapi_pool.ml
```diff
@@ -554,57 +554,39 @@ let rec create_or_get_host_on_master __context rpc session_id (host_ref, host) :
...
+      try
+        Client.SR.get_all_records ~rpc ~session_id
+        |> List.find (fun (_, sr) -> sr.API.sR_is_tools_sr)
+        |> (fun (ref,_) -> ref)
+      with Not_found ->
+        let msg = Printf.sprintf "can't find SR of tools iso %s" my_uuid in
```
robhoes: "tools SR %s"

ocaml/xapi/dbsync_master.ml
```diff
@@ -131,8 +131,14 @@ let create_tools_sr __context name_label name_description sr_introduce maybe_cre
     (Xapi_globs.i18n_original_value_prefix ^ "name_label"), name_label;
     (Xapi_globs.i18n_original_value_prefix ^ "name_description"), name_description
   ] in
+  let destroy self =
+    try Db.SR.destroy ~__context ~self with e ->
+    info "failed to destroy bogus tools SR %s: %s"
```
robhoes: `warn` is probably appropriate.

ocaml/xapi/dbsync_master.ml
```diff
@@ -149,8 +155,10 @@ let create_tools_sr __context name_label name_description sr_introduce maybe_cre
           ) (Db.SR.get_all ~__context)
       in
       match old_srs with
-      | sr :: _ ->
-        Db.SR.set_is_tools_sr ~__context ~self:sr ~value:true; sr
+      | sr :: others ->
+        Db.SR.set_is_tools_sr ~__context ~self:sr ~value:true;
+        List.iter destroy others; (* destroy bogus Tool SRs CA-300103 *)
```
robhoes: We want to do this cleanup of the "old" and "bogus" Tools SRs also if we did find a Tools SR (with `is_tools_sr = true`), so in the `| sr :: []` case.

By the way, the commit message is not quite right: this is not executed on pool join, but when the master xapi starts. Pool join is fixed by the previous commit already. This code is meant to fix things up for pools that already have the bogus tools SRs due the bug that the previous commit fixes.

robhoes:
> Avoid duplicated tool ISOs on pool join

"tools SRs"

robhoes:
It may be worth adding some unit tests to test_dbsync_master.ml. That would give us more confidence that the corner cases are working.

lindig:
I have added two unit tests.

PR #3761, opened 2018-11-13
CP-29936: override control/feature-suspend if data/cant_suspend_reason is set

edwintorok:
The guest might be able to suspend, but QEMU may not support it,
for example if the guest has not unplugged its NVME disks yet.
Remove the "feature-suspend" from guest metrics and replace it with
the reason error from QEMU when this happens, this will cause the
allowed operations on the VM to be updated accordingly.

Related to https://github.com/xapi-project/xenopsd/pull/571

### DIFF
--- a/ocaml/xapi/xapi_guest_agent.ml
+++ b/ocaml/xapi/xapi_guest_agent.ml
@@ -204,9 +204,13 @@ let mutex = Mutex.create ()
    a directory from xenstore. Both are relative to the guest's domainpath. *)
 let get_initial_guest_metrics (lookup: string -> string option) (list: string -> string list) =
   let all_control = list "control" in
-  let to_map kvpairs = List.concat (List.map (fun (xskey, mapkey) -> match lookup xskey with
-      | Some xsval -> [ mapkey, xsval ]
-      | None -> []) kvpairs) in
+  let cant_suspend_reason = lookup "data/cant_suspend_reason" in
+  let to_map kvpairs = List.concat (List.map (fun (xskey, mapkey) ->
+      match lookup xskey, xskey, cant_suspend_reason with
+      | Some xsval, "control/feature-suspend", Some reason ->
+        ["data-cant-suspend-reason", reason]
+      | Some xsval, _, _ -> [ mapkey, xsval ]
+      | None, _, _ -> []) kvpairs) in
 
   let get_tristate xskey =
     match lookup xskey with


ocaml/xapi/xapi_guest_agent.ml
```diff
@@ -205,7 +205,12 @@ let mutex = Mutex.create ()
 let get_initial_guest_metrics (lookup: string -> string option) (list: string -> string list) =
   let all_control = list "control" in
   let to_map kvpairs = List.concat (List.map (fun (xskey, mapkey) -> match lookup xskey with
-      | Some xsval -> [ mapkey, xsval ]
+      | Some xsval ->
+        if xskey = "control/feature-suspend" then
```
lindig: This could be also guarded using `when` is a new match case above the existing one.

ocaml/xapi/xapi_guest_agent.ml
```diff
@@ -218,6 +223,10 @@ let get_initial_guest_metrics (lookup: string -> string option) (list: string ->
     | Some value -> ["data-ts",value]
     | None -> [] in
 
+  let cant_suspend = match lookup "data/cant_suspend_reason" with
```
robhoes: Where is `cant_suspend` used?

edwintorok: I'll need to refactor this a bit, I ended up with some code duplication.

ocaml/xapi/xapi_guest_agent.ml
edwintorok: I refactored this to use pattern matching.

ocaml/xapi/xapi_guest_agent.ml
```diff
@@ -204,9 +204,13 @@ let mutex = Mutex.create ()
    a directory from xenstore. Both are relative to the guest's domainpath. *)
 let get_initial_guest_metrics (lookup: string -> string option) (list: string -> string list) =
   let all_control = list "control" in
-  let to_map kvpairs = List.concat (List.map (fun (xskey, mapkey) -> match lookup xskey with
-      | Some xsval -> [ mapkey, xsval ]
-      | None -> []) kvpairs) in
+  let cant_suspend_reason = lookup "data/cant_suspend_reason" in
```
lindig: I find this code a bit difficult to parse (because of the `kvpairs` argument to `List.map`) but also have no good suggestion that is much better.

edwintorok: Would `kvpairs |> List.map ...` be clearer?

PR #3765, opened 2018-11-16
CA-302194 XSI-87 move code to end of on_xapi_restart

lindig:
This fixes the previous commit - the code for building the
tree in xen should be run last to be equivalent with the code
on the master branch.


robhoes:
Is that the right commit message?

PR #3767, opened 2018-11-19
REQ-638: Support checkpoint for  Nvidia vGPU enabled VMs

minli1:
To support checkpoint for Nvidia vGPU enabled VMs, this PR contains:

- Remove restriction for VMs with vGPUs
- Do not clear `resident_on` for VM and VGPU in checkpoint.

### DIFF
--- a/ocaml/xapi/xapi_vm_lifecycle.ml
+++ b/ocaml/xapi/xapi_vm_lifecycle.ml
@@ -279,7 +279,7 @@ let check_vgpu ~__context ~op ~ref_str ~vgpus ~power_state =
   | `pool_migrate | `migrate_send
     when List.for_all is_migratable  vgpus
       && List.for_all is_suspendable vgpus -> None
-  | `suspend
+  | `suspend | `checkpoint
     when List.for_all is_suspendable vgpus -> None
   | `pool_migrate | `migrate_send | `suspend | `checkpoint ->
     Some (Api_errors.vm_has_vgpu, [ref_str])
--- a/ocaml/xapi/xapi_vm_lifecycle.ml
+++ b/ocaml/xapi/xapi_vm_lifecycle.ml
@@ -584,6 +584,10 @@ let update_allowed_operations ~__context ~self =
   if Db.is_valid_ref __context appliance then
     Xapi_vm_appliance_lifecycle.update_allowed_operations ~__context ~self:appliance
 
+let checkpoint_in_progress ~__context ~vm =
+  Listext.List.setify (List.map snd (Db.VM.get_current_operations ~__context ~self:vm))
+  |> List.mem `checkpoint
+
 (** 1. Called on new VMs (clones, imports) and on server start to manually refresh
     the power state, allowed_operations field etc.  Current-operations won't be
     cleaned 
@@ -637,7 +641,8 @@ let force_state_reset_keep_current_operations ~__context ~self ~value:state =
     Db.VM.set_requires_reboot ~__context ~self ~value:false
   end;
 
-  if state = `Halted || state = `Suspended then begin
+  (* Do not clear resident_on for VM and VGPU in a checkpoint operation *)
+  if state = `Halted || (state = `Suspended && not (checkpoint_in_progress ~__context ~vm:self)) then begin
     Db.VM.set_resident_on ~__context ~self ~value:Ref.null;
     (* make sure we aren't reserving any memory for this VM *)
     Db.VM.set_scheduled_to_be_resident_on ~__context ~self ~value:Ref.null;
--- a/ocaml/xapi/xapi_vm_snapshot.ml
+++ b/ocaml/xapi/xapi_vm_snapshot.ml
@@ -221,8 +221,6 @@ let checkpoint ~__context ~vm ~new_name =
   (* restore the power state of the VM *)
   if power_state = `Running
   then begin
-    let localhost = Helpers.get_localhost ~__context in
-    Db.VM.set_resident_on ~__context ~self:vm ~value:localhost;
     debug "Performing a slow resume";
     Xapi_xenops.resume ~__context ~self:vm ~start_paused:false ~force:false;
   end;
--- a/ocaml/xapi/xapi_xenops.ml
+++ b/ocaml/xapi/xapi_xenops.ml
@@ -2897,7 +2897,8 @@ let suspend ~__context ~self =
               Client.VM.suspend dbg id d |> sync_with_task __context ~cancellable:false queue_name;
               Events_from_xenopsd.wait queue_name dbg id ();
               Xapi_vm_lifecycle.assert_final_power_state_is ~__context ~self ~expected:`Suspended;
-              if not(Db.VM.get_resident_on ~__context ~self = Ref.null) then
+              if not(Xapi_vm_lifecycle.checkpoint_in_progress ~__context ~vm:self)
+              && not(Db.VM.get_resident_on ~__context ~self = Ref.null) then
                 raise Api_errors.(Server_error(internal_error, [
                     Printf.sprintf "suspend: The VM %s is still resident on the host" (Ref.string_of self)]));
             with e ->


minli1:
vGPU suite: 96833 verified.
vGPU migration Job: 2327692 success.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -613,6 +613,19 @@ module MD = struct
...
   let get_target_pci_address ~__context vgpu =
+    (* Reallocate a vgpu to during vm checkpoint *)
+    let vm = vgpu.Db_actions.vGPU_VM in
+    let implementation = Db.VGPU_type.get_implementation ~__context~self:vgpu.Db_actions.vGPU_type in
+    let vgpu =
+      match Db.VM.get_power_state ~__context ~self:vm, implementation with
+      | `Suspended, `nvidia ->
+        let host = Helpers.get_localhost ~__context in
+        let record = Db.VM.get_record ~__context ~self:vm in
+        Helpers.with_global_lock(fun () -> Vgpuops.create_vgpus ~__context host (vm, record) true);
```
robhoes: This does not look like the right place to call this. It is normally `Message_forwarding.allocate_vm_to_host` that does this. Why is that not happening here? Could you explain the sequence of operations?

minli1: Thanks. @robhoes 
1. Yes, normally the function `forward_to_suitable_host`  which include `allocate_vm_to_host` does this VM start-`Message_forwarding.start` or  Resume-`Message_forwarding.resume`; However, in vm's checkpoint, it will call `Xapi_xenops.resume` directly at [here](https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_vm_snapshot.ml#L227), so it won't get a valid vGPU through Message_forwarding.
Thus, before this fix, it will raise an exn in [here](https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_xenops.ml#L622) because of there is no vGPU in VM.resume.

2. For the main sequence of operations:
Xapi_vm_snapshot.checkpoint 
-> Xapi_xenops.suspend (it will remove the vGPU allocated in function`Xapi_vm_lifecycle.force_state_reset`) 
-> Xapi_vm_clone.clone (checkpoint operation) 
-> Xapi_xenops.resume

minli1: Plus, I run more vGPU migration jobs with this fix. 2327822 and 2327823.
As it was merged to feature branch at first, we are developing more test cases base on feature branch. Thanks. @robhoes

ocaml/xapi/xapi_vm_lifecycle.ml
```diff
@@ -279,7 +279,7 @@ let check_vgpu ~__context ~op ~ref_str ~vgpus ~power_state =
   | `pool_migrate | `migrate_send
     when List.for_all is_migratable  vgpus
       && List.for_all is_suspendable vgpus -> None
-  | `suspend
+  | `suspend | `checkpoint
```
krizex: The original condition is `| \`suspend`

ocaml/xapi/xapi_xenops.ml
```diff
@@ -613,6 +613,19 @@ module MD = struct
       (List.combine (Range.to_list (Range.make 0 (List.length devs))) devs)
 
   let get_target_pci_address ~__context vgpu =
+    (* Reallocate a vgpu to during vm checkpoint *)
+    let vm = vgpu.Db_actions.vGPU_VM in
+    let implementation = Db.VGPU_type.get_implementation ~__context~self:vgpu.Db_actions.vGPU_type in
```
krizex: add space after `~__context`

minli1: 👍

ocaml/xapi/xapi_vm_lifecycle.ml
krizex: Sorry, I misunderstand OCaml match syntax.
> Pattern matching with guards corresponds to the evaluation of a conditional expression immediately after the pattern is matched. If this expression comes back true, then the expression associated with that pattern is evaluated, otherwise pattern matching continues with the following pattern.

So `| a | b when c` means the following evaluation steps:
1. match `a` or `b`
2. if matched, check condition `c`

minli1:
@TalonsLee

ocaml/xapi/xapi_xenops.ml
```diff
@@ -613,6 +613,19 @@ module MD = struct
       (List.combine (Range.to_list (Range.make 0 (List.length devs))) devs)
 
   let get_target_pci_address ~__context vgpu =
+    (* Reallocate a vgpu to during vm checkpoint *)
+    let vm = vgpu.Db_actions.vGPU_VM in
+    let implementation = Db.VGPU_type.get_implementation ~__context ~self:vgpu.Db_actions.vGPU_type in
+    let vgpu =
+      match Db.VM.get_power_state ~__context ~self:vm, implementation with
+      | `Suspended, `nvidia ->
```
edwintorok: What happens if it is not Nvidia, but AMD or Intel? Can we support suspend on them with the same code?

minli1: 1. For this function `get_target_pci_address`, if AMD or Intel, then the record of vgpu will be the original one, nothing changes with the old logic.
2. For AMD or Intel vGPU vms suspendable, please see https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_vm_lifecycle.ml#L265, only nvidia vGPU support suspend and checkpoint(the newly added).

Thanks, Edvin.

ocaml/xapi/xapi_xenops.ml
robhoes: OK, I see what you mean. However, while it may work in some cases, fixing it this way has a couple of problems:
1. Resource allocation for VMs is important and complicated, and best done from a single place in the code. It is already quite hard to see how this works right now, and adding such calls to Xapi_xenops would make it worse.
2. The job of Xapi_xenops functions like `of_nvidia_vgpu` is to just get information from the DB and turn this into xenopsd config. This patch mixes in resourcing here, which does not fit.
3. The `with_global_lock` function must only be called on the master, usually from Message_forwarding (it's just a mutex), and the same is true for resource allocation functions like `create_vgpus`. Now this code runs on the host where the VM is running, which means that the locking does not work, which can lead to concurrency problems.

Taking a step back, it seems to me that for a VM checkpoint, while internally the VM goes through a suspend/resume cycle, the VM should ideally keep its resources (such as vGPUs), rather than release and reacquire them.

minli1: Thanks, Rob. You are right, it's not good to mix the resource reservation in xenopsd anymore.
I had refined the code to get all the `record` of the vgpus and then reset the field of `resident_on` if have.
Please help to review again on the latest change. Thanks!

ocaml/xapi/xapi_vm_snapshot.ml
```diff
@@ -223,6 +226,11 @@ let checkpoint ~__context ~vm ~new_name =
   then begin
     let localhost = Helpers.get_localhost ~__context in
     Db.VM.set_resident_on ~__context ~self:vm ~value:localhost;
+    match vgpu_records with
+    | [] -> ()
+    | vgpu_rec :: _ ->
+      let vgpu = Db.VGPU.get_by_uuid ~__context ~uuid:vgpu_rec.Db_actions.vGPU_uuid in
+      Db.VGPU.set_resident_on ~__context ~self:vgpu ~value:vgpu_rec.Db_actions.vGPU_resident_on;
```
robhoes: There is a race window now, where the pGPU can be stolen from the VM that is being checkpointed by another VM starting up. This is because after the `suspend` call returns (L209), and during the `clone`, the `VPGU.resident_on` field is clear.

In fact, the same thing is already true for memory, due to `VM.resident_on` being cleared at the same time.

How about addressing both of these in `force_state_reset_keep_current_operations`, by having that function recognise that a checkpoint is in progress, and if so, not clear those fields? I think we can determine this from the `current_operations` and then update the if-statement in https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_vm_lifecycle.ml#L640.

minli1: Thanks, Rob. @robhoes  That makes sense.
I had updated the PR base on your comment, please help to review again. Thanks so much.

robhoes:
OK, I think that this looks good now. Have you tested this on a couple of checkpoints with and without vGPUs?

minli1:
Thanks, @robhoes . I have manually tested on checkpoint with/without vGPUs, both worked.
vGPU migration job 2341532 and checkpoint(without vGPU) job 2341534 running.

robhoes:
OK, all 💚 .

PR #3771, opened 2018-11-23
Reference xapi-inventory instead of xcp-inventory; the latter is being deprecated.

kc284:
Goes with https://github.com/xapi-project/xs-opam/pull/350


lindig:
What about the Travis failure?

PR #3772, opened 2018-11-26
CA-300644: return immediately when attach-static-vdis failed

taoyongd:
When slave host cannot connect to NFS server(HA metadata VDI and HA state file VDI reside on), from the XenRT case it drops the port to NFS, the slave will in emergency mode, but when NFS server comes back by allowing the port in iptables, what's  expected is the slave host will quit emergency mode. But we met a scenario that when attaching first VDI(the NFS didn't come back), the attach-static-vdis will fail, but succeed on attaching the second one(as now NFS server can be accessed). Xapi will marked it as failed and retry to attach the VDIs after 10s, so in the second round, the first VDI will be attached successfully(Because it failed in the first round) and then the second VDI will fail to attach(as it succeeds in the first round).

Even though we discussed in the ticket CA-300644 that the test may need to rethink the procedure. But I raised this PR to know why we didn't return immediately when attaching VDI failed? Is it by design? 
And I didn't come up with a reason not to do this change.

### DIFF
--- a/scripts/init.d-attach-static-vdis
+++ b/scripts/init.d-attach-static-vdis
@@ -39,6 +39,7 @@ attach_all(){
 	    if [ $? -ne 0 ]; then
 	       RC=1
 	       logger "Attempt to attach VDI: ${UUID} failed -- skipping (Error was: ${OUTPUT})"
+	       return $RC
 	    fi
 	done
 	return $RC


scripts/init.d-attach-static-vdis
```diff
@@ -39,6 +39,7 @@ attach_all(){
 	    if [ $? -ne 0 ]; then
 	       RC=1
 	       logger "Attempt to attach VDI: ${UUID} failed -- skipping (Error was: ${OUTPUT})"
+	       return $RC
```
lindig: This could be also implemented with a `break`.

taoyongd: Yes. :+1:  Here we neeed to understand whether it is reasonable. I didn't come up with a reason not do this.

taoyongd:
Hi @jonludlam @edwintorok , any thoughts about this PR? Thanks.

lindig:
Ping @jonludlam

scripts/init.d-attach-static-vdis
taoyongd: I think we can remember which static VDI is attached successfully, but to make it simpler, it may not need to do that.

robhoes:
We have discussed this, and agree that it is the right approach.

PR #3773, opened 2018-11-26
CA-299554 XSI-132 use correct vCPU count for dom0

lindig:
The number of vCPUs in dom0 was set to the number of CPUs in the host -
which is wrong. Read the correct number from /proc/cpuinfo from inside
dom0.

### DIFF
--- a/ocaml/xapi/create_misc.ml
+++ b/ocaml/xapi/create_misc.ml
@@ -89,6 +89,14 @@ let make_kpatch_list () =
 
 open Xstringext
 
+(** [count_cpus] returns the number of CPUs found in /proc/cpuinfo *)
+let count_cpus () =
+  let cpuinfo = "/proc/cpuinfo" in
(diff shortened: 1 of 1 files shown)


ocaml/xapi/create_misc.ml
```diff
@@ -89,6 +89,26 @@ let make_kpatch_list () =
...
+      loop [] in
+  let with_file filename f =
+    let ic = open_in filename in
+    Pervasiveext.finally
+      (fun () -> f ic)
+      (fun () -> close_in ic) in
+  let re = Re.compile @@ Re.Pcre.re "^processor\\s*:\\s+\\d+" in
+  let matches line = Re.matches re line <> [] in
+  let count n line = if matches line then n+1 else n in
+  with_file cpuinfo read |> List.fold_left count 0
```
robhoes: What about `Unixext.file_lines_fold`?

lindig: Will take a look.

PR #3774, opened 2018-11-26
BP CA-299554 XSI-132 use correct vCPU count for dom0

lindig:
The number of vCPUs in dom0 was set to the number of CPUs in the host -
which is wrong. Read the correct number from /proc/cpuinfo from inside
dom0.

Backport of 4a2cc12d1

### DIFF
--- a/ocaml/xapi/create_misc.ml
+++ b/ocaml/xapi/create_misc.ml
@@ -89,6 +89,14 @@ let make_kpatch_list () =
 
 open Xstringext
 
+(** [count_cpus] returns the number of CPUs found in /proc/cpuinfo *)
+let count_cpus () =
+  let cpuinfo = "/proc/cpuinfo" in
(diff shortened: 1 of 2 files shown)


ocaml/xapi/create_misc.ml
```diff
@@ -89,6 +89,14 @@ let make_kpatch_list () =
 
 open Xstringext
 
+(** [count_cpus] returns the number of CPUs found in /proc/cpuinfo *)
+let count_cpus () =
+  let cpuinfo = "/proc/cpuinfo" in
+  let re = Re_perl.compile @@ Re_perl.re {|^processor\s*:\s+\d+|} in
```
edwintorok: I wasn't sure whether this'd work on older versions of OCaml, but according to https://caml.inria.fr/pub/docs/manual-ocaml/extn.html `{|quoted strings|}` were introduced in OCaml 4.02 so should be fine.

robhoes: We should really get travis on this branch...

PR #3775, opened 2018-11-27
CP-29757: block SXM of encrypted VDIs

gaborigloi:
This will not block the "Move VM" operation on a halted VM in XenCenter,
because XenCenter moves the VM manually by copying the VDIs of the VM
and creating & plugging new VBDs.

But this is fine, because we wanted to block SXM of encrypted VDIs for
XenAPI users. And both the VM.migrate_send call and "xe vm-migrate" will
be blocked for encrypted VDIs, even if the VM is halted.

### DIFF
--- a/ocaml/xapi/xapi_vm_migrate.ml
+++ b/ocaml/xapi/xapi_vm_migrate.ml
@@ -138,15 +138,24 @@ let assert_sr_support_operations ~__context ~vdi_map ~remote ~ops =
     )
 
 (** Check that none of the VDIs that are mapped to a different SR have CBT
-    enabled. This function must be called with the complete [vdi_map],
-    which contains all the VDIs of the VM.
+    or encryption enabled. This function must be called with the complete
+    [vdi_map], which contains all the VDIs of the VM.
     [check_vdi_map] should be called before this function to verify that this
     is the case. *)
-let assert_no_cbt_enabled_vdi_migrated ~__context ~vdi_map =
+let assert_can_migrate_vdis ~__context ~vdi_map =
+  let assert_cbt_not_enabled vdi =
+    if Db.VDI.get_cbt_enabled ~__context ~self:vdi then
+      raise Api_errors.(Server_error(vdi_cbt_enabled, [Ref.string_of vdi]))
+  in
+  let assert_not_encrypted vdi =
+    let sm_config = Db.VDI.get_sm_config ~__context ~self:vdi in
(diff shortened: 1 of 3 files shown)


ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -138,15 +138,24 @@ let assert_sr_support_operations ~__context ~vdi_map ~remote ~ops =
...
     is the case. *)
-let assert_no_cbt_enabled_vdi_migrated ~__context ~vdi_map =
+let assert_can_migrate_vdis ~__context ~vdi_map =
+  let assert_cbt_not_enabled vdi =
+    if Db.VDI.get_cbt_enabled ~__context ~self:vdi then
+      raise Api_errors.(Server_error(vdi_cbt_enabled, [Ref.string_of vdi]))
+  in
+  let assert_not_encrypted vdi =
+    let other_config = Db.VDI.get_sm_config ~__context ~self:vdi in
+    if List.exists (fun (key, _value) -> key = "key_hash") other_config  then
```
jonludlam: can you please change the name of this variable to `sm_config`? `other_config` has other connotations

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -138,15 +138,24 @@ let assert_sr_support_operations ~__context ~vdi_map ~remote ~ops =
...
-let assert_no_cbt_enabled_vdi_migrated ~__context ~vdi_map =
+let assert_can_migrate_vdis ~__context ~vdi_map =
+  let assert_cbt_not_enabled vdi =
+    if Db.VDI.get_cbt_enabled ~__context ~self:vdi then
+      raise Api_errors.(Server_error(vdi_cbt_enabled, [Ref.string_of vdi]))
+  in
+  let assert_not_encrypted vdi =
+    let sm_config = Db.VDI.get_sm_config ~__context ~self:vdi in
+    if List.exists (fun (key, _value) -> key = "key_hash") sm_config then
+      failwith ("Migration of encrypted VDI " ^ (Ref.string_of vdi) ^ " is not allowed")
```
lindig: Why is this not an API error like above?

edwintorok: is there another Api error that we could raise here? I think this one would get translated to an internal error, which may be confusing as there is nothing wrong with XAPI, but a user error.

PR #3778, opened 2018-11-28
CP-29002: add unit test for VM.NVRAM field

edwintorok:
This should not be settable while a VM is running, one has to use the special
set_NVRAM_EFI_variables call that is restricted by RBAC.

### DIFF
--- a/ocaml/tests/test_vm_check_operation_error.ml
+++ b/ocaml/tests/test_vm_check_operation_error.ml
@@ -68,6 +68,27 @@ let test_null_vdi () =
         all_vm_operations
     )
 
+let test_vm_set_nvram_running () =
+  with_test_vm (fun __context vm_ref ->
+      Db.VM.set_power_state ~__context ~self:vm_ref ~value:`Halted;
+      let old_nvram = ["EFI-variables", "AAAA"] in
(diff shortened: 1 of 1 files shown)


ocaml/tests/test_vm_check_operation_error.ml
```diff
@@ -68,6 +68,27 @@ let test_null_vdi () =
...
+        (fun () ->
+           Api_server.Forwarder.VM.set_NVRAM ~__context ~self:vm_ref ~value:["EFI-variables", "BBBB"]);
+      let read_nvram = Db.VM.get_NVRAM ~__context ~self:vm_ref in
+      Alcotest.(check (list (pair string string))) "NVRAM not updated" old_nvram read_nvram;
+
+      let new_vars = "CCCC" in
+      let new_nvram = ["EFI-variables", new_vars] in
+      Api_server.Forwarder.VM.set_NVRAM_EFI_variables ~__context ~self:vm_ref ~value:new_vars;
+      let read_nvram = Db.VM.get_NVRAM ~__context ~self:vm_ref in
+      Alcotest.(check (list (pair string string))) "NVRAM not updated" new_nvram read_nvram;
```
robhoes: This looks like the right test but the wrong string.

PR #3781, opened 2018-12-03
Use dune and define profile "gprof" for profiling

lindig:
Switch to to Dune and define a profile for profiling with gprof while
keeping jbuild files for now. Since Dune has stricter warning flags in
its dev profile, we have to use the release profile to avoid braking the
build until this is cleaned up. Switching to a different profile is easy
from the command line:

  make PROFILE=dev


dune
```diff
@@ -0,0 +1,20 @@
+(env
+ (gprof
+  (ocamlopt_flags
+   (:standard -g -p -w -39)
```
edwintorok: I think `-g` is there by default, otherwise backtraces wouldn't work, but it doesn't hurt to be explicit.

dune
```diff
@@ -0,0 +1,20 @@
...
+  )
+ )
+ (dev
+  (flags
+   (:standard -g -w -39)
+  )
+ )
+ (release
+  (flags
+   (:standard -w -39)
```
mseri: Are you sure you want to get rid of `g` here? The code may become undebuggable

edwintorok: -g should be part of :standard

dune
lindig: According to the logs, we are still compiling with `-g`.

PR #3782, opened 2018-12-04
CP-30032: spawn varstore-rm in a chroot

edwintorok:
Previously we've run it as root, sandbox it instead.
There is no domain id associated with a VM clone, so we just use
the same user `qemu_base + 0` for sandboxing. This is less than ideal, but better than running as root.

### DIFF
--- a/ocaml/xapi/xapi_xenops.ml
+++ b/ocaml/xapi/xapi_xenops.ml
@@ -194,8 +194,7 @@ let varstore_rm_with_sandbox ~__context ~vm_uuid f =
   let domid = 0 in
   let chroot, socket_path = Xenops_sandbox.Varstore_guard.start dbg ~domid ~vm_uuid ~paths:[] in
   Xapi_stdext_pervasives.Pervasiveext.finally (fun () -> f chroot socket_path)
-    (fun () ->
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_xenops.ml
```diff
@@ -189,15 +189,28 @@ let firmware_of_vm vm =
         bad]))
   | exception Not_found -> default_firmware
 
+let varstore_rm_with_sandbox ~__context ~vm_uuid f =
+  let dbg = Context.string_of_task __context in
+  let domid = 0 in
+  let chroot, socket_path = Xenops_sandbox.Varstore_guard.start dbg ~domid ~vm_uuid ~paths:[] in
+  Xapi_stdext_pervasives.Pervasiveext.finally (fun () -> f chroot socket_path)
+    (fun () ->
+       let (_: string list) = Xenops_sandbox.Varstore_guard.stop dbg ~domid ~paths:[] in ())
```
lindig: How about `ignore @@ Xenops_sandbox.Varstore_guard.stop dbg ~domid ~paths:[]`

edwintorok: I agree that this looked ugly, I've actually fixed this in the xenopsd PR by making `stop` return unit, updated this PR too now.

PR #3783, opened 2018-12-05
CA-304576: Allow checkpoint on suspended VM with Nvidia vGPU

minli1:
XS support checkpoint when VM in suspend state:
https://docs.citrix.com/en-us/xenserver/current-release/dr/snapshots.html

When VM with Nvidia vGPU becoming a suspended state, it will update
allowed_operations to exclude checkpoint which is not correct.

This commit is refining to check the power_state and is_nvidia to
decide checkpoint permission.

So now there are two conditions to allow checkpoint for vGPUs VMs:
1. Checking all vgpus on VM are suspendable by calling `is_suspendable`
2. power state is `suspended` and the vGPU type is `NVIDIA`

### DIFF
--- a/ocaml/xapi/xapi_vm_lifecycle.ml
+++ b/ocaml/xapi/xapi_vm_lifecycle.ml
@@ -279,6 +279,8 @@ let check_vgpu ~__context ~op ~ref_str ~vgpus ~power_state =
   | `pool_migrate | `migrate_send
     when List.for_all is_migratable  vgpus
       && List.for_all is_suspendable vgpus -> None
+  | `checkpoint
+    when power_state = `Suspended -> None
   | `suspend | `checkpoint
     when List.for_all is_suspendable vgpus -> None
   | `pool_migrate | `migrate_send | `suspend | `checkpoint ->


ocaml/xapi/xapi_vm_lifecycle.ml
```diff
@@ -352,7 +356,7 @@ let maybe_get_guest_metrics ~__context ~ref =
   then Some (Db.VM_guest_metrics.get_record_internal ~__context ~self:ref)
   else None
 
-(** Take an internal VM record and a proposed operation. Return None iff the operation
```
lindig: "iff" means "if and only if" and comes from mathematical writing. I suspect it is not a typo.

ocaml/xapi/xapi_vm_lifecycle.ml
```diff
@@ -279,7 +279,11 @@ let check_vgpu ~__context ~op ~ref_str ~vgpus ~power_state =
   | `pool_migrate | `migrate_send
     when List.for_all is_migratable  vgpus
       && List.for_all is_suspendable vgpus -> None
-  | `suspend | `checkpoint
+  | `checkpoint
+    when (List.for_all is_suspendable vgpus)
+      || (power_state = `Suspended
+      && List.for_all (fun vgpu -> Xapi_gpumon.Nvidia.is_nvidia ~__context ~vgpu) vgpus) -> None
```
lindig: Is indentation correct here? The structure is `(a || (b && c))`.

lindig: The logic looks correct to me.

lindig:
Why is Travis failing?

minli1:
Thanks for your review! @lindig 
The travis failed due to:
```
# File "_build/.dune/default/ocaml/database/jbuild", line 32, characters 3-6:
# 32 |    xcp
#         ^^^
# Error: Library "xcp" not found.
# Hint: try: dune external-lib-deps --missing -p xapi-database @install
```
Do you have any idea of this? The feature branch req-638/master is not rebased for days.

ocaml/xapi/xapi_vm_lifecycle.ml
minli1: Yes, that's the structure.  `(a || (b && c))`

minli1:
Updated. @lindig  Thanks.

edwintorok:
I think the error about xcp is due to the changes we did recently in xs-opam

lindig:
Package `xcp` has been renamed to `xapi-idl` on the master branch. So it might be time to update the feature branch.

ocaml/xapi/xapi_vm_lifecycle.ml
robhoes: Why do we need to check for nvidia again? I thought that the point of the helper functions was to hide that. Besides, if the power state is currently "suspended", then doesn't that already imply the right kind of vGPU?

minli1: Emm, you are right. @robhoes  
If a VM with vGPU could went into a `suspended` state, `Nvidia` checking must already happened in `checkpoint`. 
I will refine it. Thanks.

robhoes: So the `is_suspendable` check is only valid for running VMs, not for suspended ones. So I think that we first need to check the power state. Therefore, I think that we should leave the existing case as is, and add the following above it:
```
| `checkpoint
  when power_state = `Suspended -> None
```

ocaml/xapi/xapi_vm_lifecycle.ml
robhoes: They probably meant "iff" indeed, but I'm not sure whether it is actually used correctly :)

minli1:
Thanks, I have updated according to your comment. @robhoes

PR #3786, opened 2018-12-07
CA-281176: hide deprecated VDI operations from the allowed_operations…

zli:
… field

### DIFF
--- a/ocaml/xapi/xapi_vdi.ml
+++ b/ocaml/xapi/xapi_vdi.ml
@@ -299,11 +299,12 @@ let update_allowed_operations_internal ~__context ~self ~sr_records ~pbd_records
    * re-enable (and maybe alter) the relevant part of
    * test_update_allowed_operations.
   *)
-  let all_ops = Listext.List.set_difference
-      (* Older XenServers choke on ops they don't recognise during SXM, so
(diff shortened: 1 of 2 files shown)


ocaml/xapi/xapi_vdi.ml
```diff
@@ -305,6 +305,7 @@ let update_allowed_operations_internal ~__context ~self ~sr_records ~pbd_records
       Xapi_globs.pre_ely_vdi_operations
       [`blocked]
   in
+  let all_ops = Listext.List.set_difference all_ops Xapi_globs.deprecated_vdi_operations in
```
lindig: I find the handling of `blocked` and `force_unlock` a bit inconsistent. One is exposed globally and the other is not. I would have done:

```
let all_ops =
  Xapi_globs.pre_ely_vdi_operations
  |> List.filter (function
       | `blocked -> false (* CA-260245 *)
       | `force_unlock -> false (* CA-281176 *)
       | _ -> true )
```

This would keep it all in one place.

zli: Updated.

zli:
@robhoes Unit test updated.

PR #3787, opened 2018-12-10
Remove some unused binaries, tests, and values

gaborigloi:
And add a new "make check" target.


Makefile
```diff
@@ -11,6 +11,10 @@ PROFILE=release
 build:
 	dune build @install -j $(JOBS) --profile=$(PROFILE)
 
+# Quickly verify that the code compiles, without actually building it
+check:
+	dune build @check -j $(JOBS) --profile=$(PROFILE)
```
edwintorok: Does this require a newer version of dune?

ocaml/tests/cancel_tests.ml
```diff
@@ -1,431 +0,0 @@
...
-
-let make_rpc ?dbg () xml =
-  let open Xmlrpc_client in
-  let http = {
-    (xmlrpc ~version:"1.0" "/") with
-    Http.Request.additional_headers = Opt.default [] (Opt.map (fun dbg -> [ "X-Http-other-config-dbg", dbg ]) dbg)
-  } in
-  XMLRPC_protocol.rpc ~srcstr:"graph" ~dststr:"xapi" ~transport:(TCP(!host, !port)) ~http xml
-
-let wait_for_guest_agent ~rpc ~session_id ~vm =
```
edwintorok: This looks like a useful test, should we add it to quicktest instead of deleting it (assuming it stil works)?

ocaml/tests/storage_impl_test.ml
```diff
@@ -1,535 +0,0 @@
...
-  create_vdi_test "sr";
-
-  if !total_errors = 0 then begin
-    info "OK";
-    exit 0;
-  end else begin
-    info "%d errors detected" !total_errors;
-    exit 1
-  end
-
```
edwintorok: This test looks useful, do we have equivalents for it in quicktest, or can we add this to quicktest?

ocaml/tests/test_pr1510.ml
```diff
@@ -1,105 +0,0 @@
...
-
-let suite =
-  "pr1510_suite" >:::
-  [
-    "test_lacp_timeout_prop(slow)" >:: test_lacp_timeout_prop "slow";
-    "test_lacp_timeout_prop(fast)" >:: test_lacp_timeout_prop "fast";
-    "test_lacp_aggregation_key(42)" >:: test_lacp_aggregation_key "42";
-    "test_lacp_aggregation_key_vsctl" >:: test_lacp_aggregation_key_vsctl "42";
-    "test_lacp_defaults_bond_create" >:: test_lacp_defaults_bond_create;
-  ]
```
edwintorok: This could be part of alcotests, it appears to mock out calls to the actual ovs cli, so it should? work even in travis?

Makefile
gaborigloi: Yep, they recently introduced this

ocaml/tests/cancel_tests.ml
gaborigloi: We can create a ticket for that.

ocaml/tests/storage_impl_test.ml
gaborigloi: No, but it's a layering violation. I think it's not worth the effort. But we can still add a ticket to re-add it.

ocaml/tests/test_pr1510.ml
gaborigloi: It does not currently compile. We can add a ticket to re-add it.

ocaml/tests/storage_impl_test.ml
edwintorok: Ok lets drop this test.

gaborigloi:
Thanks for the review. I think we should merge it and add tickets to re-introduce the removed tests because that's much more work.

gaborigloi:
Thanks!

PR #3788, opened 2018-12-11
CA-300719: Block export of >2TB VDI to VHD format.

GeraldElder-Vass:
Blocks the export of >2TB VDI to VHD format operations instantly.


ocaml/xapi/export_raw_vdi.ml
```diff
@@ -53,6 +53,12 @@ let localhost_handler rpc session_id vdi (req: Http.Request.t) (s: Unix.file_des
              Http_svr.headers s headers;
              match format with
              | Raw | Vhd ->
+             if format = Vhd 
+             then
```
gaborigloi: This should be indented to be in line with the subsequent lines

ocaml/xapi/export_raw_vdi.ml
```diff
@@ -53,6 +53,12 @@ let localhost_handler rpc session_id vdi (req: Http.Request.t) (s: Unix.file_des
              Http_svr.headers s headers;
              match format with
              | Raw | Vhd ->
+               if format = Vhd then begin
+                 let size = Db.VDI.get_virtual_size ~__context ~self:vdi in
+                 if size > 2199023255552L
```
edwintorok: Our supported limit is 2TiB-4GiB, see https://docs.citrix.com/en-us/xenserver/current-release/system-requirements/configuration-limits.html, I think we should use that.
I think this constant is already defined here: https://github.com/xapi-project/xen-api/blob/680e7b1feed1186e998a845fe4979bc9dfc9cd4a/ocaml/quicktest/quicktest_max_vdi_size.ml#L6

The constant itself should probably be moved to `constants.ml`, and both this code and the unit test could refer to it.

I think @gaborigloi tested various values in a PR a while ago and we determined that TB means TiB in those limits. (from our official docs it is not clear whether TB=10^12 bytes, or TB=TiB=2^40 bytes

ocaml/quicktest/quicktest_max_vdi_size.ml
```diff
@@ -3,17 +3,16 @@
     https://docs.citrix.com/content/dam/docs/en-us/xenserver/current-release/downloads/xenserver-config-limits.pdf
     Using (2TiB - 4GiB), instead of the (2TB - 4GB) limit defined in the above
     document, does not work, we cannot create a VDI of that size. *)
```
gaborigloi: We should move this comment as well to the new place of max_vdi_size

ocaml/xapi/export_raw_vdi.ml
```diff
@@ -53,6 +53,12 @@ let localhost_handler rpc session_id vdi (req: Http.Request.t) (s: Unix.file_des
              Http_svr.headers s headers;
              match format with
              | Raw | Vhd ->
+               if format = Vhd then begin
+                 let size = Db.VDI.get_virtual_size ~__context ~self:vdi in
+                 if size > Constants.max_vhd_size
+                 then raise (Api_errors.Server_error(Api_errors.vdi_too_large, [ Ref.string_of vdi; Int64.to_string size]))
```
gaborigloi: We should pass the max size into the error instead of the VDi's size

ocaml/idl/datamodel_errors.ml
```diff
@@ -706,6 +706,8 @@ let _ =
     ~doc:"The operation cannot be performed because this VDI has rrd stats" ();
   error Api_errors.vdi_too_small [ "vdi"; "minimum size" ]
     ~doc:"The VDI is too small. Please resize it to at least the minimum size." ();
+  error Api_errors.vdi_too_large [ "vdi"; "maximum size" ]
+    ~doc:"The VDI is too large. Cannot export to VHD >2TB" ();
```
lindig: This is confusing. II talks about the VDI being too large and refers to "VHD > 2TB". It's not obvious what is too large. How about "Cannot export VDI larger than 2TB as a VHD" or similar.

ocaml/quicktest/quicktest_max_vdi_size.ml
```diff
@@ -1,19 +1,12 @@
-
-(** From XenServer 7.5 Configuration Limits:
-    https://docs.citrix.com/content/dam/docs/en-us/xenserver/current-release/downloads/xenserver-config-limits.pdf
-    Using (2TiB - 4GiB), instead of the (2TB - 4GB) limit defined in the above
-    document, does not work, we cannot create a VDI of that size. *)
-let max_vdi_size = Sizes.((2L ** tb) -* (4L ** gb))
-
 let with_max_vdi rpc session_id sr f =
-  Qt.VDI.with_new rpc session_id ~virtual_size:max_vdi_size sr
+  Qt.VDI.with_new rpc session_id ~virtual_size:Constants.max_vhd_size sr
```
lindig: Again - what is too large: the VHD or the VDI?

ocaml/xapi/export_raw_vdi.ml
```diff
@@ -53,6 +53,12 @@ let localhost_handler rpc session_id vdi (req: Http.Request.t) (s: Unix.file_des
              Http_svr.headers s headers;
              match format with
              | Raw | Vhd ->
+               if format = Vhd then begin
+                 let size = Db.VDI.get_virtual_size ~__context ~self:vdi in
+                 if size > Constants.max_vhd_size
```
lindig: Again - the code obtains a VDI size and compares it a gainst a VHD size.

ocaml/idl/datamodel_errors.ml
gaborigloi: True. The last sentence should be deleted, this is a more generic error message.

ocaml/quicktest/quicktest_max_vdi_size.ml
gaborigloi: True again. We should rename the function to `with_max_vhd`.

ocaml/xapi/export_raw_vdi.ml
gaborigloi: This is fine here, because we are exporting a VHD.

PR #3790, opened 2018-12-14
Use sets not lists

rwdrich:
Lists:

* "adding" an element: O(1)
* length: O(n)
* accessing cell i: O(i)
* finding an element: O(n)
* diff: O(n^2) - from xapi-stdext-std/listext.ml#L176

Sets:

* "adding" an element: O(log n)
* returning the number of elements: O(n)
* finding an element: O(log n)
* diff: O(n)

* of_list: O(n log (n))
There is the overhead of creating the sets, but with larger lists I
think this will be negligible


ocaml/xapi/xapi_xenops.ml
```diff
@@ -14,6 +14,8 @@
 
 module D=Debug.Make(struct let name="xenops" end)
 open D
+module StringSet = Set.Make(struct type t = string let compare = compare end)
+module VMSet = Set.Make(struct type t = [ `VM ] API.Ref.t let compare = compare end)
```
robhoes: It looks like you are not using `VMSet`.

rwdrich: True! I had previously used it, but realised that the case where I was using it was just converting a list to a set and back again, so removed where it was used, forgot to remove the module declaration

ocaml/xapi/xapi_xenops.ml
```diff
@@ -2388,14 +2389,14 @@ let events_from_xapi () =
...
-                let extra_in_cache = Listext.List.set_difference cached uuids in
-                if missing_in_cache <> []
-                then error "events_from_xapi: missing from the cache: [ %s ]" (String.concat "; " missing_in_cache);
-                if extra_in_cache <> []
-                then error "events_from_xapi: extra items in the cache: [ %s ]" (String.concat "; " extra_in_cache);
+                let uuids = StringSet.of_list (List.rev_map (fun self -> Db.VM.get_uuid ~__context ~self) resident_VMs) in
+                let cached = StringSet.of_list (Mutex.execute metadata_m (fun () -> Xenops_cache.list_nolock ())) in
+                let missing_in_cache = StringSet.diff uuids cached in
+                let extra_in_cache = StringSet.diff cached uuids in
+                if missing_in_cache <> StringSet.empty
```
edwintorok: There is a StringSet.is_empty you could use here instead, usually it is best to avoid polymorphic comparisons on sets (the empty set would probably work, but comparing otherwise identical sets with polymorphic compare could result in them being considered different).

rwdrich: Ok, I've changed this to use the `is_empty` function instead

edwintorok: Sorry I didn't mean to literally use `String.is_empty`, obviously preserve the original meaning of the test and check that it is not empty by negating it.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -2388,14 +2389,14 @@ let events_from_xapi () =
...
-                then error "events_from_xapi: missing from the cache: [ %s ]" (String.concat "; " missing_in_cache);
-                if extra_in_cache <> []
-                then error "events_from_xapi: extra items in the cache: [ %s ]" (String.concat "; " extra_in_cache);
+                let uuids = StringSet.of_list (List.rev_map (fun self -> Db.VM.get_uuid ~__context ~self) resident_VMs) in
+                let cached = StringSet.of_list (Mutex.execute metadata_m (fun () -> Xenops_cache.list_nolock ())) in
+                let missing_in_cache = StringSet.diff uuids cached in
+                let extra_in_cache = StringSet.diff cached uuids in
+                if StringSet.is_empty missing_in_cache
+                then error "events_from_xapi: missing from the cache: [ %s ]" (String.concat "; " (StringSet.elements missing_in_cache));
+                if StringSet.is_empty extra_in_cache
```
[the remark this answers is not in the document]
rwdrich: This is wrong. Corrected the condition

ocaml/xapi/xapi_xenops.ml
```diff
@@ -14,6 +14,7 @@
 
 module D=Debug.Make(struct let name="xenops" end)
 open D
+module StringSet = Set.Make(struct type t = string let compare = compare end)
```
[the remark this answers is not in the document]
rwdrich: Thank you

ocaml/xapi/xapi_xenops.ml
rwdrich: Yes, it was a mistake, Christian also pointed it out

robhoes:
There is now a conflict since the guefi merge, which needs to be resolved.

edwintorok:
The conflict was fixed, the build failure is because the xenserver branch is not out of date (due to BVT taking too long to finish, there were no new packages uploaded last night).

PR #3792, opened 2019-01-02
CP-30244: Avoid checking power_state for checkpoint while creating VGPU

minli1:
For VM with vGPU and checkpoint, migration will fail because of target
    host creating vGPU needs VM in 'halted' power_state but the checkpoint
    is actually in 'suspended' power_state.
    
    This commit is to avoid checking power_state for VMs which
    'is_a_snapshot' filed is true.
    

ocaml/xapi/xapi_vgpu.ml
```diff
@@ -26,7 +26,7 @@ let create' ~__context  ~vM ~gPU_group ~device ~other_config ~_type ~powerstate_
   let uuid = Uuid.to_string (Uuid.make_uuid ()) in
   if not (Pool_features.is_enabled ~__context Features.GPU) then
     raise (Api_errors.Server_error (Api_errors.feature_restricted, []));
-  if powerstate_check then
+  if powerstate_check && not (Db.VM.get_is_a_snapshot ~__context ~self:vM) then
```
lindig: For a snapshot the power check is skipped entirely. Should we not check for a different power state in case it is a snapshot, e.g.:

```
let expected = match Db.VM.get_is_a_snapshot ~__context ~self:vM with
  | true  -> `Suspended 
  | false -> `Halted in
if power_check then
  api_vm_lifecycle.assert_initial_power_state_is ~__context ~self:vM ~expected;
```

minli1: Thanks for your comment. @lindig 
However not all snapshots are `suspended` state. When it's a snapshot(disk only), it's `Halted` and when it's a checkpoint(disk+memory), it's `Suspended` because it was taken while the VM is suspended.
So we can't get the `expected` simply by checking it's a snapshot or not.
Per my understanding, we are calling `VGPU.create` for snapshot(checkpoint) only in migration on destination host, so we may not care about the snapshot's power_state not only for snapshot but also for checkpoint.

minli1: This change will not impact the normal VM at all.

lindig: I see and would be fine with your patch. This exposes a weakness in `assert_initial_power_state_is` that we can't easily check for a set of potential power states (or pass a predicate for the check).

minli1: Yes, I agree with you. Probably creating vGPU on a snapshot(checkpoint), we don't need to care about the power_state. :) Thanks.

minli1:
Job 2374418,2374421 passed.

ocaml/xapi/xapi_vgpu.ml
minli1: Hi, Rob. Thanks for your useful comments. 
1. 
The original [VGPU.create](https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_vgpu.ml#L65) doesn't have params to let us get `Db.VM.get_is_a_snapshot` and pass inside the [VGPU.create'](https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_vgpu.ml#L24).
2. If I get `powerstate_check` inside `create'`, then it may impact the function in [VGPU.copy](https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_vgpu.ml#L77), because it has powerstate_check setting to false to pass inside `create'`, I suspect VGPU.copy will fail in scenario if I implement in this way.

The `powerstate_check` was introduced in below commit to void failure of checkpoint with vGPU.
https://github.com/minli1/xen-api/commit/e28080849bfc3875606958b33bb04aacd169034e
This commit create new function `create'` to have the param, so now I think it's probably better to 
call `create'` with `powerstate_check = false` while VM import, it will not impact other functions at all.  
Please help to see the updated code again. Thanks.

krizex: Export `create'` is not a good idea, we need a better name of this function.

minli1: I rename `create'` to `create_with_powercheck ` :)

ocaml/xapi/import.ml
```diff
@@ -1273,9 +1273,10 @@ module VGPU : HandlerTools = struct
     | Found_VGPU _ ->
       handle_dry_run __context config rpc session_id state x precheck_result
     | Create vgpu_record -> begin
+        (* No need to check powerstate during VM import *)
         let vgpu = log_reraise "failed to create VGPU" (fun value ->
-            let vgpu = Client.VGPU.create ~rpc ~session_id ~vM:value.API.vGPU_VM ~gPU_group:value.API.vGPU_GPU_group
-                ~device:value.API.vGPU_device ~other_config:value.API.vGPU_other_config ~_type:value.API.vGPU_type in
+            let vgpu = Xapi_vgpu.create_with_powercheck ~__context ~device:value.API.vGPU_device ~gPU_group:value.API.vGPU_GPU_group
```
robhoes: I'm a afraid that this is not right. VGPU.create is done through the API, because it needs to be executed on the pool master. There is no way to add extra parameters.

ocaml/xapi/xapi_vgpu.ml
robhoes: `create_with_powercheck` is not great, because it doesn't always do a powerstate check... ;)

robhoes: 1) You can just add the params:
```
let create ~__context  ~vM ~gPU_group ~device ~other_config ~_type =
  let powerstate_check = not (Db.VM.get_is_a_snapshot ~__context ~self:vM) in
  create' ~__context  ~vM ~gPU_group ~device ~other_config ~_type ~powerstate_check
```
2) The `copy` function always calls `create'` with `~powerstate_check:false`, so it should never have problems?

minli1: Done. Thanks, Rob. @robhoes

PR #3795, opened 2019-01-09
Merge REQ-638 to master

minli1:
This PR is to merge REQ-638 feature branch to master.
Include below commits(already get reviewed when PR to feature branch):
- CP-29673: allow checkpoint op for VM with VGPU.
- CA-302456: Do not clear 'resident_on' during a checkpoint operation.
- CA-304576: Allow checkpoint on suspended VM with Nvidia VGPU.
- CA-307012: Avoid checking power_state for checkpoint while creating VGPU.

The QA report based on the feature branch:
https://info.citrite.net/display/xenserver/QA+Status+Report+for+REQ-638%3A+Support+Snapshots+for+NVIDIA+vGPU-Enabled+VMs


ocaml/xapi/xapi_vm_lifecycle.ml
```diff
@@ -584,6 +586,10 @@ let update_allowed_operations ~__context ~self =
   if Db.is_valid_ref __context appliance then
     Xapi_vm_appliance_lifecycle.update_allowed_operations ~__context ~self:appliance
 
+let checkpoint_in_progress ~__context ~vm =
+  Listext.List.setify (List.map snd (Db.VM.get_current_operations ~__context ~self:vm))
```
lindig: We can keep it but I don't see why `setify` is required. If the same item is multiple times in the list, `List.mem` will return the same result and overall it would be more efficient.

PR #3796, opened 2019-01-14
Checksum 

RoryBG:
VDI checksumming is now done using fast non-cryptographic hash functions. This change replaces calls to the Sha1 and Digest class with calls to a SeaHash binding.

### DIFF
--- a/ocaml/seahash/hash.ml
+++ b/ocaml/seahash/hash.ml
@@ -1,0 +1,16 @@
+open Ctypes
+open Foreign
+
+let bytes = foreign "bytes" (ocaml_bytes @-> uint64_t @-> returning uint64_t)
+let file = foreign "file" (string @-> returning uint64_t)
+
+let to_hex hash =
+	Printf.sprintf "%016LX" hash;;
+
+let bytes buff = 
+	let size = Unsigned.UInt64.of_int (Bytes.length buff) in
+	Unsigned.UInt64.to_int64(bytes (ocaml_bytes_start buff) size);;
+
+let file name = 
+	close_in (open_in_bin name); (* To emulate the behaviour of the Digest class, also throws an error if the file does not exist *)
+	Unsigned.UInt64.to_int64(file name);; (* We know the file exists at this point *)
--- a/ocaml/seahash/hash.ml
+++ b/ocaml/seahash/hash.ml
@@ -1,8 +1,11 @@
 open Ctypes
 open Foreign
 
-let bytes = foreign "bytes" (ocaml_bytes @-> uint64_t @-> returning uint64_t)
-let file = foreign "file" (string @-> returning uint64_t)
+let bytes = foreign "hash_bytes" (ocaml_bytes @-> uint64_t @-> returning uint64_t)
+let file = foreign "hash_filedesc" (int @-> returning uint64_t)
+
+let fd_of_int (x: int) : Unix.file_descr = Obj.magic x;;
+let int_of_fd (x: Unix.file_descr) : int = Obj.magic x;;
 
 let to_hex hash =
 	Printf.sprintf "%016LX" hash;;
@@ -11,6 +14,17 @@ let bytes buff =
 	let size = Unsigned.UInt64.of_int (Bytes.length buff) in
 	Unsigned.UInt64.to_int64(bytes (ocaml_bytes_start buff) size);;
 
-let file name = 
-	close_in (open_in_bin name); (* To emulate the behaviour of the Digest class, also throws an error if the file does not exist *)
-	Unsigned.UInt64.to_int64(file name);; (* We know the file exists at this point *)
+let file file_name = 
+	if Sys.is_directory file_name || not(Sys.file_exists file_name)
+	then raise Not_found
+	else begin
+		let fd = Unix.openfile file_name [O_RDONLY] 0o666 in
+		let csum = Unsigned.UInt64.to_int64(file (int_of_fd fd)) in
+		
+		if csum = 0L
+		then raise End_of_file (* Unexpected EOF *)
+		else begin
+			Unix.close fd;
+			csum
+		end
+	end
--- a/ocaml/seahash/libseahash/Makefile
+++ b/ocaml/seahash/libseahash/Makefile
@@ -1,6 +1,6 @@
 
 all:
-	gcc -O3 -std=c11 -c seahash.c -o libseahash_stubs.a
+	cc -O3 -std=gnu11 -c seahash.c -o libseahash_stubs.a
 
 clean:
 	rm -f ./libseahash_stubs.a
--- a/ocaml/seahash/libseahash/Makefile
+++ b/ocaml/seahash/libseahash/Makefile
@@ -1,0 +1,6 @@
+
+all:
+	gcc -O3 -std=c11 -c seahash.c -o libseahash_stubs.a
+
+clean:
+	rm -f ./libseahash_stubs.a
--- a/ocaml/seahash/libseahash/seahash.c
+++ b/ocaml/seahash/libseahash/seahash.c
@@ -1,0 +1,167 @@
+#include <stdint.h>
+#include <assert.h>
+#include <string.h>
+#include <stdio.h>
+
+/*
+	This code is a C implementation of the SeaHash algorithm (see: http://ticki.github.io/blog/seahash-explained/)
+	The author also provides a reference implemenation written in Rust, the license for this reference implementation is below: 
+*/
+
+/*
+	The MIT License (MIT)
+
+	Copyright (c) 2016 Ticki
+
+	Permission is hereby granted, free of charge, to any person obtaining a copy
+	of this software and associated documentation files (the "Software"), to deal
+	in the Software without restriction, including without limitation the rights
+	to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+	copies of the Software, and to permit persons to whom the Software is
+	furnished to do so, subject to the following conditions:
+
+	The above copyright notice and this permission notice shall be included in all
+	copies or substantial portions of the Software.
+-fsyntax-only
+	THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+	IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+	FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+	AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+	LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+	OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+	SOFTWARE.
+*/
+
+#define BUFF_SIZE (61440)
+
+uint64_t A, B, C, D;
+
+void init() {
+	// These default values come from the SeaHash specification
+	A = 0x16F11FE89B0D677C,
+	B = 0xB480A793D8E6C86C,
+	C = 0x6FE2E5AAF078EBC9,
+	D = 0x14F994A4C5259381;
+}
+
+uint64_t diffuse(uint64_t x) {
+	const uint64_t p = 0x6EED0E9DA4D94A4F;
+	x *= p;
+	x ^= (x >> 32) >> (x >> 60);
+	x *= p;
+	return x;
+}
+
+uint64_t finalise(const uint64_t size) {
+	A ^= B;
+	C ^= D;
+	A ^= C;
+	A ^= size;
+	return diffuse(A);
+}
+
+void rotate(uint64_t x) {
+	x = diffuse(A ^ x);
+	A = B; // Rotate the lanes
+	B = C;
+	C = D;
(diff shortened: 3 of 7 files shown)


RoryBG:
The C code contains some tests in the `main` function to check the algorithm's correctness. So it can be compiled into an object file or it can be tested by compiling and executing it as a program. I have run these tests with `-fsanitize=address` and `-fsanitize=undefined` (along with Clang's static analysis).

jonludlam:
That's an impressive increase in coverage :-P

ocaml/seahash/libseahash/seahash.c
```diff
@@ -0,0 +1,144 @@
...
+		if(size == 0) break; // EOF
+		update(data, size);
+		sum += size;
+	}
+
+	fclose(in);
+	return finalise(sum);
+}
+
+int main(int argc, char **argv) {
```
edwintorok: This one should go into a separate file, I'm surprised the linker picked the correct `main()` when building xapi.

ocaml/seahash/libseahash/seahash.c
```diff
@@ -0,0 +1,144 @@
...
+#include <stdio.h>
+
+/*
+	This code is a C implementation of the SeaHash algorithm (see: http://ticki.github.io/blog/seahash-explained/)
+	The author also provides a reference implemenation written in Rust (which follows the MIT license). 
+*/
+
+#define BUFF_SIZE (61440)
+
+uint64_t A, B, C, D;
```
edwintorok: It is usually best to avoid global variables, especially when dealing with threaded code. Even though in its current state OCaml is single threaded, that might change in the future, and we can also explicitly drop the runtime lock if we'd be using bigstrings that do not move.
It would be best to put these into a struct.

ocaml/seahash/libseahash/seahash.c
```diff
@@ -0,0 +1,144 @@
...
+
+void init() {
+	// These default values come from the SeaHash specification
+	A = 0x16F11FE89B0D677C,
+	B = 0xB480A793D8E6C86C,
+	C = 0x6FE2E5AAF078EBC9,
+	D = 0x14F994A4C5259381;
+}
+
+uint64_t diffuse(uint64_t x) {
```
edwintorok: If you don't intend on exporting these functions then we should use `static inline` to allow the compiler to do more optimizations.

ocaml/seahash/libseahash/seahash.c
gaborigloi: We could move these tests into xapi's unit tests to avoid linking them into xapi.
That way we'll also test the ocaml bindings.

ocaml/seahash/libseahash/seahash.c
```diff
@@ -0,0 +1,144 @@
...
+	x = diffuse(A ^ x);
+	A = B; // Rotate the lanes
+	B = C;
+	C = D;
+	D = x;
+}
+
+void update(const void *data, const uint64_t size) {
+	const uint64_t freq = ((size >> 3) >> 2);
+	const uint64_t *head = (uint64_t *) data;
```
edwintorok: Might want to check the alignment requirements here before casting, and make sure the address is aligned to multiples of 8.
Even though x86-64 does not normally have alignment requirements (as opposed to Sparc) if the compiler succeeds in auto-vectorizing the code then it would crash (likely with SIGBUS) even on x86-64 if it is not aligned properly, because vectorized code *does* have alignment requirements.

ocaml/seahash/hash.ml
```diff
@@ -0,0 +1,16 @@
...
+
+let to_hex hash =
+	Printf.sprintf "%016LX" hash;;
+
+let bytes buff = 
+	let size = Unsigned.UInt64.of_int (Bytes.length buff) in
+	Unsigned.UInt64.to_int64(bytes (ocaml_bytes_start buff) size);;
+
+let file name = 
+	close_in (open_in_bin name); (* To emulate the behaviour of the Digest class, also throws an error if the file does not exist *)
```
gaborigloi: You can use `Sys.file_exists` and `Sys.is_directory` to check the file: https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&ved=2ahUKEwi55qyD6O3fAhULzqQKHdl-Dh8QFjAAegQICRAB&url=https%3A%2F%2Fcaml.inria.fr%2Fpub%2Fdocs%2Fmanual-ocaml%2Flibref%2FSys.html&usg=AOvVaw0ZFy5l_8_hAeiq59UjyXpa

ocaml/seahash/libseahash/seahash.c
```diff
@@ -0,0 +1,144 @@
...
+uint64_t bytes(const void *data, const uint64_t size) {
+	init();
+	update(data, size);
+	return finalise(size);
+}
+
+uint64_t file(const char *name) {
+	FILE *in = fopen(name, "rb");
+	if(in == NULL) {
+		fprintf(stderr, "Cannot open file: %s\n", name);
```
edwintorok: A library printing to standard error is not very nice, it should raise an exception instead.

ocaml/seahash/libseahash/seahash.c
```diff
@@ -0,0 +1,144 @@
...
+	// NB: we read backwards because we need little-endian
+	for(uint8_t i = 1; i <= rem; i++) {
+		x <<= 8;
+		x += body[rem - i];
+	}
+
+	rotate(x);
+}
+
+uint64_t bytes(const void *data, const uint64_t size) {
```
edwintorok: Might be better to call this process_bytes, it is a bit confusing to have it called the same name as the type.

ocaml/seahash/libseahash/Makefile
```diff
@@ -0,0 +1,6 @@
+
+all:
+	gcc -O3 -std=c11 -c seahash.c -o libseahash_stubs.a
```
edwintorok: This should use CC, that way you also get the hardening flags that an RPM build would usually set. You probably also get an easier way to run things like clang static analyzer and coverity if you allow CC to be overridable.

ocaml/seahash/libseahash/seahash.c
```diff
@@ -0,0 +1,144 @@
+#include <stdint.h>
```
edwintorok: What is the copyright of this file?

ocaml/seahash/hash.ml
```diff
@@ -0,0 +1,16 @@
+open Ctypes
```
edwintorok: Copyright?

ocaml/seahash/libseahash/seahash.c
```diff
@@ -0,0 +1,144 @@
...
+	fclose(in);
+	return finalise(sum);
+}
+
+int main(int argc, char **argv) {
+	// These hash outputs were precomputed using the SeaHash reference implementation (which is written in Rust).
+	char temp[4096];
+
+	char *a = ""; // 0 bytes
+	strcpy(temp, a);
```
lindig: Never use `strcpy`! Use `strlcpy`

ocaml/seahash/libseahash/seahash.c
lindig: I think this function should take a file descriptor or a `FILE` object, not a file name and leave opening and closing to the caller.

ocaml/xapi/stream_vdi.ml
```diff
@@ -407,10 +409,14 @@ let recv_all_vdi refresh_session ifd (__context:Context.t) rpc session_id ~has_i
              in
              Unixext.really_read ifd buffer 0 (Int64.to_int length);
              Unix.write ofd buffer 0 (Int64.to_int length) |> ignore;
-             let csum = Sha1.to_hex (Sha1.string (Bytes.unsafe_to_string buffer)) in
-
-             checksum_table := (file_name, csum) :: !checksum_table;
-
+             
+             let csum = if String.endswith file_name checksum_extension
```
lindig: Use `Filename.check_suffix`

ocaml/xapi/stream_vdi.ml
```diff
@@ -324,7 +326,7 @@ let verify_inline_checksum ifd checksum_table =
     let csum = Bytes.unsafe_to_string csum in
     Tar_unix.Archive.skip ifd (Tar_unix.Header.compute_zero_padding_length hdr);
     (* Look up the relevant file_name in the checksum_table *)
-    let original_file_name = String.sub file_name 0 (String.length file_name - (String.length checksum_extension)) in
+    let original_file_name = String.sub file_name 0 (String.index file_name '.') in
```
lindig: Use `Filename.remove_extension`

ocaml/seahash/hash.ml
```diff
@@ -0,0 +1,16 @@
...
+open Foreign
+
+let bytes = foreign "bytes" (ocaml_bytes @-> uint64_t @-> returning uint64_t)
+let file = foreign "file" (string @-> returning uint64_t)
```
lindig: Like I said below, I think this should take a file descriptor or `FILE` object, not a file name. This makes the function more flexible (can read from a pipe, for example) and dealing with opening and closing is left to the caller.

ocaml/seahash/libseahash/seahash.c
```diff
@@ -0,0 +1,144 @@
...
+}
+
+uint64_t file(const char *name) {
+	FILE *in = fopen(name, "rb");
+	if(in == NULL) {
+		fprintf(stderr, "Cannot open file: %s\n", name);
+		return 0;		
+	}
+	init();
+	uint8_t *data[BUFF_SIZE];
```
lindig: `BUFF_SIZE` is quite large for a stack-allocated object.

ocaml/seahash/libseahash/seahash.c
```diff
@@ -0,0 +1,144 @@
...
+		C ^= *head++;
+		D ^= *head++;
+
+		A = diffuse(A);
+		B = diffuse(B);
+		C = diffuse(C);
+		D = diffuse(D);
+	}
+
+	const uint8_t remain = size - (size & 0xFFFFFFFFFFFFFFE0); // How many bytes weren't hashed above (n < 32)
```
lindig: Could this be expressed as `size & 0x1F` or similar? I'm not  comfortable with this subtraction.

ocaml/seahash/libseahash/seahash.c
gaborigloi: It will be removed & moved to ocaml unit tests

ocaml/seahash/libseahash/seahash.c
gaborigloi: For the user, it will look quite nice, `Seahash.Hash.bytes`.
To me this reads like a command "hash bytes!".

ocaml/seahash/libseahash/seahash.c
RoryBG: I wanted to keep IO operations down so I chose a pretty big buffer size. Though I do agree something like 16K would probably be more appropriate here.

ocaml/seahash/libseahash/seahash.c
```diff
@@ -0,0 +1,144 @@
...
+	if(in == NULL) {
+		fprintf(stderr, "Cannot open file: %s\n", name);
+		return 0;		
+	}
+	init();
+	uint8_t *data[BUFF_SIZE];
+
+	uint64_t sum = 0;
+	for(;;) {
+		const uint64_t size = (uint64_t) fread(data, 1, BUFF_SIZE, in);
```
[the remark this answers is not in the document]
RoryBG: From the docs:
> fread does not distinguish between end-of-file and error, and callers must use `feof` and `ferror` to determine which occurred. 

So I'll add code to detect if a read error has occured, however I'm not sure what our response would be, I don't think we could recover. Once an error has occured that whole FILE stream can be considered broken. As per the docs:

> If an error occurs, the resulting value of the file position indicator for the stream is indeterminate. If a partial element is read, its value is indeterminate. 

I think the only thing we can do in this case is throw an exception in this case.

ocaml/seahash/libseahash/seahash.c
```diff
@@ -0,0 +1,145 @@
...
+	for(uint8_t i = 1; i <= rem; i++) {
+		x <<= 8;
+		x += body[rem - i];
+	}
+
+	rotate(state, x);
+}
+
+uint64_t hash_bytes(const void *data, const uint64_t size) {
+	SeaState state = init();
```
lindig: This assigns a complex value. See comment above.

ocaml/seahash/libseahash/seahash.c
```diff
@@ -0,0 +1,145 @@
...
+	update(&state, data, size);
+	return finalise(&state, size);
+}
+
+uint64_t hash_file(FILE *in) {
+	SeaState state = init();
+	uint64_t data[BUFF_SIZE >> 3];
```
lindig: This declares a large buffer on the stack and multiplies the size by 8. Why do we need a large buffer? The OS is reading much smaller blocks anyway.

ocaml/seahash/hash.ml
```diff
@@ -0,0 +1,30 @@
+open Ctypes
+open Foreign
+
+let bytes = foreign "hash_bytes" (ocaml_bytes @-> uint64_t @-> returning uint64_t)
+let file = foreign "hash_filedesc" (int @-> returning uint64_t)
+
+let fd_of_int (x: int) : Unix.file_descr = Obj.magic x;;
```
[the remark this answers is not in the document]
gaborigloi: I don't think that can be done with ctypes :(. We took this code from fd-send-recv, which does the same.

lindig:
Closing this after internal discussion.

PR #3797, opened 2019-01-16
Zstd VM export

johnelse:
This pull request implements VM export and import using the Zstd algorithm. Compression algorithm is specified on export, and automatically detected on import via the gzip/zstd magic strings.

Comments/open questions:

* I've added a feature flag to `features.ml` mainly so clients can tell whether a server supports this feature. xapi does not actually check this feature flag before doing a zstd import/export, but I can add this if necessary.
* There is no check that the zstd binary is installed on the host. This could be done by calling `Zstd.available ()` and throwing an exception if it is not installed, but since this is not checked for `gzip` I haven't implemented this. For XCP-ng we plan to add `zstd` as a dependency of the xapi RPM.

### DIFF
--- a/ocaml/xapi-types/features.ml
+++ b/ocaml/xapi-types/features.ml
@@ -62,6 +62,7 @@ type feature =
   | USB_passthrough
   | Network_sriov
   | Corosync
+  | Zstd_export
 [@@deriving rpc]
 
 type orientation = Positive | Negative
@@ -113,6 +114,7 @@ let keys_of_features =
     USB_passthrough, ("restrict_usb_passthrough", Negative, "USB_passthrough");
     Network_sriov, ("restrict_network_sriov", Negative, "Network_sriov");
     Corosync, ("restrict_corosync", Negative, "Corosync");
+    Zstd_export, ("restrict_zstd_export", Negative, "Zstd_export");
   ]
 
 (* A list of features that must be considered "enabled" by `of_assoc_list`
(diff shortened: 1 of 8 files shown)


lindig:
Thank you for your contribution. Having so far only glanced at it, the code looks good. Since this introduces new capabilities, I expect some discussion about this. It might be helpful to mark all commits with a "zstd import" prefix or similar to keep this together.

johnelse:
Thanks for the comment, I've updated all commits with the tag "Zstd export" - the content is unchanged.

lindig:
We generally like this feature but might not merge it immediately because we are currently in a feature freeze.

lindig:
We would now be in a position to merge this. Could you rebase and check that the Travis build succeeds?

johnelse:
Thanks! I'll rebase this tonight.

lindig:
This PR introduced a regression; we backed it out again to fix it: it's no longer possible to import VMs that are in ZIP format (not GZIP).

johnelse:
Ah apologies - where do you want to go from here? This should just need an extra check for the .zip magic string (`50 4B 03 04`) - I can test that and remake this pull request with the fix if you'd be happy with that.

lindig:
You will see the revert commits on the `master` branch. You could create a new PR that re-introduces the commits and the fix in addition. @psafont is also looking at it here. We backed it out just to reduce the pressure to come up with a fix immediately.

psafont:
More than one extra check might be needed.

From gzip's manual:
> gunzip can currently decompress files created by gzip, zip, compress or pack.

This includes .z files using Lempel-Ziv-Welch (`1F 9D`), I'm not sure if there are more formats supported or if they are even used in the wild for compressing VMs.

PR #3798, opened 2019-01-16
CA-307829: XSI-216 Add active state in update_vgpu

taoyongd:
See https://github.com/xapi-project/xenopsd/pull/592


taoyongd:
Any comments on this?
Thanks.

taoyongd:
Yes. We should keep it attached during the reboot.
Updated.

PR #3806, opened 2019-01-29
CP-30527: Reduce direct usage of Xenctrl

psafont:
This commit tries reduces direct calls to Xenctrl by replacing it with xenopsd communication, it also tries to reduce communication between xapi and xenopsd to gather host information, as well as making it more apparent when it's happening.

I'm unsure about not being able to retrieve the actual memory values when xenopsd is down, feedback on how to tackle this is appreciated.

I'm also unsure on whether it would be beneficial to replace the `make` in the name of methods in `create_misc.ml` to better reflect that they write to the database.

Xenctrl is still used for testing in `ocaml/quicktest/quicktest_vm_lifecycle.ml` and is used to periodically update information about the host in `ocaml/xapi/monitor_dbcalls.ml` and `ocaml/xapi/monitor_types.ml`. This last usage needs a more thought'out approach to not have a negative performance impact.

### DIFF
--- a/ocaml/xapi/create_misc.ml
+++ b/ocaml/xapi/create_misc.ml
@@ -111,25 +111,13 @@ let read_dom0_memory_usage () =
 
 let read_localhost_info () =
   let xen_verstring, total_memory_mib =
-    try
-      let xc = Xenctrl.interface_open () in
-      Xenctrl.interface_close xc;
-      let open Xapi_xenops_queue in
-      let module Client = (val make_client (default_xenopsd ()) : XENOPS) in
-      let stat = Client.HOST.stat "read_localhost_info" in
-      let xen_verstring = stat.hypervisor.version in
-      let total_memory_mib =
-        Client.HOST.get_total_memory_mib "read_localhost_info" in
-      xen_verstring, total_memory_mib
-    with e ->
-      if Pool_role.is_unit_test ()
-      then "0.0.0", 0L
-      else begin
-        warn "Failed to read xen version";
-        match Balloon.get_memtotal () with
-        | None -> "unknown", 0L
-        | Some x -> "unknown", Int64.(div x (mul 1024L 1024L))
-      end
+    let open Xapi_xenops_queue in
+    let module Client = (val make_client (default_xenopsd ()) : XENOPS) in
+    let stat = Client.HOST.stat "read_localhost_info" in
+    let xen_verstring = stat.hypervisor.version in
+    let total_memory_mib =
+      Client.HOST.get_total_memory_mib "read_localhost_info" in
+    xen_verstring, total_memory_mib
   and linux_verstring =
     let verstring = ref "" in
     let f line =
--- a/ocaml/xapi/create_misc.ml
+++ b/ocaml/xapi/create_misc.ml
@@ -43,6 +43,9 @@ type host_info = {
   total_memory_mib: int64;
   dom0_static_max: int64;
   ssl_legacy: bool;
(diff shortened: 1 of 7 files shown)


edwintorok:
You should add Signed-off-by to each commit (see the DCO link, it'll tell you how, there is a `-s` flag for `git commit --amend`, the problem likely comes from the fixups since by default they'll miss the signoff)

psafont:
Thanks for the heads-up, I'll fix it at the end of the review.

ocaml/xapi/create_misc.ml
```diff
@@ -109,28 +112,12 @@ let read_dom0_memory_usage () =
...
-      Xenctrl.interface_close xc;
-      let open Xenctrl in
-      let xen_verstring = Printf.sprintf "%d.%d%s" v.major v.minor v.extra in
-      let total_memory_mib =
-        let open Xapi_xenops_queue in
-        let module Client = (val make_client (default_xenopsd ()) : XENOPS) in
-        Client.HOST.get_total_memory_mib "read_localhost_info" in
-      xen_verstring, total_memory_mib
-    with e ->
-      if Pool_role.is_unit_test ()
```
edwintorok: Do the unit tests pass now, or do you need to add some special handling in the unit test to mock out the API call?

psafont: I didn't mock any API call to make the unit tests pass neither on Travis nor on my localhost.

ocaml/xapi/create_misc.ml
```diff
@@ -109,28 +112,12 @@ let read_dom0_memory_usage () =
...
-      then "0.0.0", 0L
-      else begin
-        warn "Failed to read xen version";
-        match Balloon.get_memtotal () with
-        | None -> "unknown", 0L
-        | Some x -> "unknown", Int64.(div x (mul 1024L 1024L))
-      end
+let read_localhost_info ~__context =
+  let open Xapi_xenops_queue in
```
lindig: Do we need this `open`?

psafont: It it necessary to be able to use `XENOPS` as Xenops_client.Client, on the following line.

There might be a better way to use the Xenopsd API client, but this is how I saw it being used everywhere else.

lindig: Otherwise is would be `Xapi_xenops_queue.XENOPS`. Let's keep iusing `open`.

ocaml/xapi/create_misc.ml
robhoes: `read_localhost_info` does not seem to be used from a unit test right now, so the exception handler is probably not useful for that. The other part  of the exception handler looks like it is meant to cover the case where xapi is not running on a Xen system, which isn't something that is relevant anymore.

ocaml/xapi/dbsync_slave.ml
```diff
@@ -122,21 +122,11 @@ let refresh_localhost_info ~__context info =
...
-    try
-      let xc = Xenctrl.interface_open () in
-      Xenctrl.interface_close xc; (* we're on xen *)
-      let dbg = Context.string_of_task __context in
-      let open Xapi_xenops_queue in
-      let module Client = (val make_client (default_xenopsd ()): XENOPS) in
-      let mib = Client.HOST.get_total_memory_mib dbg in
-      Int64.mul 1024L (Int64.mul 1024L mib)
-    with _ ->
-      warn "Failed to detect xen, querying /proc/meminfo";
```
robhoes: Same comment as above. I don't think that we need this exception handler anymore.

PR #3807, opened 2019-01-29
Faster Checksum Algorithm

RoryBG:
With these changes Xapi should see significant performance improvements when doing VM and VDI imports and exports. Formal benchmarks will be finished in the next day or so. These changes are also maintain backwards compatability. 

Moved from SHA1 checksum to xxHash checksum as per discussion. The xxHash checksum bindings come from opam: https://opam.ocaml.org/packages/xxhash/ and information about the xxHash algorithm can be found here: https://github.com/Cyan4973/xxHash

### DIFF
--- a/ocaml/xapi/stream_vdi.ml
+++ b/ocaml/xapi/stream_vdi.ml
@@ -21,6 +21,7 @@ open Debug
 open Http
 open Forkhelpers
 open Pervasiveext
+open XXHash
 
 exception Failure of string
 
@@ -33,7 +34,8 @@ let chunk_size = Int64.mul 1024L 1024L (* 1 MiB *)
 (** Maximum range for sparseness query *)
 let max_sparseness_size = Int64.mul 10240L chunk_size
 
-let checksum_extension = ".checksum"
+let checksum_extension = ".xxhash"
+let checksum_supported_extension = [checksum_extension; ".checksum"] (* Supported checksum formats, .checksum == SHA1 *)
 
 type vdi = string (* directory prefix in tar file *) * API.ref_VDI * Int64.t (* size to send/recieve *)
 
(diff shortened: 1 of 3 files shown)


xapi.opam
```diff
@@ -28,6 +28,7 @@ depends: [
   "rrdd-plugin"
   "sexpr"
   "sha"
+  "xxhash"
```
edwintorok: I think `xxhash` needs to be added to `xs-opam`, or included as a subdir in xapi (probably add to xs-opam + add patch file for the issue we know about would be the way forward).

ocaml/xapi/stream_vdi.ml
```diff
@@ -407,18 +408,26 @@ let recv_all_vdi refresh_session ifd (__context:Context.t) rpc session_id ~has_i
...
-             let csum = Sha1.to_hex (Sha1.string (Bytes.unsafe_to_string buffer)) in
+             
+             let buffer_string = Bytes.unsafe_to_string buffer in
+             
+             let csum_hdr = Tar_unix.Header.get_next_header ifd in (* Header of the checksum file *)
+	     let csum_file_name = csum_hdr.Tar_unix.Header.file_name in
+             
+             let csum = if Filename.check_suffix csum_file_name checksum_extension (* Infer the hash algorithm *)
+             	then Printf.sprintf "%016LX" (XXH64.hash buffer_string)
+             	else Sha1.to_hex (Sha1.string buffer_string)
```
edwintorok: This works, although another way to implement this would be to have a function that receives the `Filename.extension csum_file_name`, and the string buffer, does a `match` on the extension and returns the checksum.

lindig:
Please make Travis work - vendoring xxHash should do it.

ocaml/xapi/stream_vdi.ml
```diff
@@ -21,6 +21,7 @@ open Debug
 open Http
 open Forkhelpers
 open Pervasiveext
+open XXHash
```
lindig: Avoid this. The interface should be small enough that using a qualified name is bearable.

xapi.opam
edwintorok: Agreed ("vendoring" in dune can be done by just dropping xxhash into a subdirectory in `ocaml/`)

gaborigloi: Sounds good, but on the long term I think we should merge back to upstream.

lindig:
I'm not sure what to do with this but I am leaning towards closing this. I'd like to see a PR build on Travis before we merge it unless there is a good reason why it can't. With a vendored xxHash package, this should be possible.

gaborigloi:
The problem is that xxhash is not available on ubuntu 18.04, which we use for our travis builds.

lindig:
I don't think we want to merge something into Xapi where we can't get Travis to work. We should switch away from Ubuntu if we can't get xxHash there.

gaborigloi:
I agree. This PR is not yet finished.

lindig:
Closing this. Please update and re-open.

PR #3811, opened 2019-02-04
CA-310173: use a persistent path for multipath_enabled

edwintorok:
/var/run/nonpersistent would get wiped on each boot, this normally wouldn't be a problem due to `sync_config_files` which runs on startup, except for HA.
HA requires all this information to be available before XAPI is up, in the attach-static-vdis script.

Changing this requires a coordinated change in sm/xapi-storage-plugins.
/cc @MarkSymsCtx


robhoes:
Has the coordinated change been coordinated?

MarkSymsCtx:
No and we're not sure if this is the correct way of doing it anyway.,

robhoes:
OK, then I'll close this PR for now. Please reopen if/when ready.

PR #3813, opened 2019-02-05
CP-30434: Save XAPI uefi_certificates to varstored directory if the certificate files don't already exist there and parse secureboot=auto

rwdrich:


ocaml/xapi/xapi_vm.ml
```diff
@@ -261,9 +261,19 @@ let start ~__context ~vm ~start_paused ~force =
...
 
+  (* Check to see if secure boot should be parsed *)
+  let pool = Helpers.get_pool ~__context in
+  if List.assoc "secureboot" platform = "auto" then
+    Db.VM.set_platform ~__context ~self:vm
+      ~value:(("secureboot", string_of_bool (Db.Pool.get_uefi_certificates ~__context
+                                               ~self:pool <> "")) :: List.remove_assoc "secureboot" platform);
 
-  (* Check to see if we're using correct device-model when vm has VUSBs*)
+  if not (List.exists (fun e -> e = "PK.auth") (Array.to_list (Sys.readdir "/usr/share/varstored"))) then
```
lindig: You can use `Array.exists` directly.

ocaml/xapi/xapi_vm.ml
```diff
@@ -261,9 +261,19 @@ let start ~__context ~vm ~start_paused ~force =
...
+    let open_file = open_out "/usr/share/varstored/xapi_uefi_certificates.tar" in
```
lindig: I would use `finally` to avoid any leaks of a file descriptor.

ocaml/xapi/xapi_vm.ml
```diff
@@ -261,9 +261,19 @@ let start ~__context ~vm ~start_paused ~force =
     end else vmr
   in
 
+  (* Check to see if secure boot should be parsed *)
+  let pool = Helpers.get_pool ~__context in
+  if List.assoc "secureboot" platform = "auto" then
+    Db.VM.set_platform ~__context ~self:vm
```
lindig: I'm not sure but maybe "add_to_platform" could be also used - I don't know how it handles existing entries.

robhoes: The `platform` field is set just above here as well. It would also be better to avoid doing it twice for the same field, since we are potentially running on a slave.

rwdrich: I think
`if List.mem_assoc key t && (not idempotent || List.assoc key t <> value) then raise Duplicate;`
If the key already exists, and either idempotent is true, or the current value is different to the new value - then raise duplicate. So it reads to me that add_to_platform wouldn't work well for existing keys?

lindig: I agree. Keep the current code.

ocaml/xapi/xapi_vm.ml
```diff
@@ -249,21 +249,30 @@ let start ~__context ~vm ~start_paused ~force =
...
+  (* Check to see if secure boot should be parsed *)
+  let pool = Helpers.get_pool ~__context in
+  if List.assoc "secureboot" !platform = "auto" then
+    platform := (("secureboot", string_of_bool (Db.Pool.get_uefi_certificates ~__context ~self:pool <> "")) :: List.remove_assoc "secureboot" !platform);
+  Db.VM.set_platform ~__context ~self:vm ~value:!platform;
 
-  (* Check to see if we're using correct device-model when vm has VUSBs*)
+  if not (Array.exists (fun e -> e = "PK.auth") (Sys.readdir "/usr/share/varstored")) then
+    let open_file = open_out "/usr/share/varstored/xapi_uefi_certificates.tar" in
+    finally (fun () -> fprintf open_file "%s" (Xapi_stdext_base64.Base64.decode (Db.Pool.get_uefi_certificates ~__context ~self:pool)))
```
lindig: How long is the string? Why not write the string to `open_file` without using `fprintf` (which adds nothing)? if writint fails, should it be logged?

ocaml/xapi/xapi_vm.ml
lindig: `Sys.readdir` can fail with an exception. This should be at least logged.

ocaml/xapi/xapi_vm.ml
```diff
@@ -249,21 +249,30 @@ let start ~__context ~vm ~start_paused ~force =
   let default_value = match Xapi_xenops.firmware_of_vm vmr with
     | Bios -> Vm_platform.fallback_device_model_default_value
     | Uefi _ -> Vm_platform.fallback_device_model_default_value_uefi in
-  let platform = vmr.API.vM_platform
+  let platform = ref( vmr.API.vM_platform
```
lindig: Why do we need a reference here? Isn't it possible to determine the value of platform before using it later?

ocaml/xapi/xapi_vm.ml
```diff
@@ -256,14 +256,36 @@ let start ~__context ~vm ~start_paused ~force =
...
   let vmr =
     if platform <> vmr.API.vM_platform then begin
-      Db.VM.set_platform ~__context ~self:vm ~value:platform;
       { vmr with API.vM_platform = platform }
     end else vmr
   in
 
+  (* Check to see if secure boot should be parsed *)
+  let pool = Helpers.get_pool ~__context in
+  if List.assoc "secureboot" platform = "auto" then
```
edwintorok: The platform field is already updated above (see ensure_device_model_profile_present), so setting secureboot status should also be done there, perhaps by introducing a new function `|> update_platform_secureboot`, then you wouldn't need to modify the code that call Db.VM.set_platform.

ocaml/xapi/xapi_vm.ml
```diff
@@ -256,14 +256,36 @@ let start ~__context ~vm ~start_paused ~force =
...
+    Db.VM.set_platform ~__context ~self:vm
+      ~value:(("secureboot", string_of_bool (Db.Pool.get_uefi_certificates ~__context ~self:pool <> "")) :: List.remove_assoc "secureboot" platform)
+  else
+    Db.VM.set_platform ~__context ~self:vm ~value:platform;
 
-  (* Check to see if we're using correct device-model when vm has VUSBs*)
+  let dir = ref [||] in
```
edwintorok: This should go into a function of its own.

ocaml/xapi/xapi_vm.ml
```diff
@@ -256,14 +256,36 @@ let start ~__context ~vm ~start_paused ~force =
...
+    dir := Sys.readdir "/usr/share/varstored";
+  with Sys_error _ ->
+    begin
+      debug "varstored directory does not exist";
+      Xapi_stdext_unix.Unixext.mkdir_rec "/usr/share/varstored" 0o755
+    end;
+
+  if not (Array.exists (fun e -> e = "PK.auth") !dir) then
+    let open_file = open_out "/usr/share/varstored/xapi_uefi_certificates.tar" in
+    finally (fun () ->
```
edwintorok: We'll also need to extract the tar file, there is a Tar_unix module that is already used in xapi that could help.

ocaml/xapi/xapi_vm.ml
```diff
@@ -256,14 +256,36 @@ let start ~__context ~vm ~start_paused ~force =
...
+    Db.VM.set_platform ~__context ~self:vm ~value:platform;
 
-  (* Check to see if we're using correct device-model when vm has VUSBs*)
+  let dir = ref [||] in
+  try
+    dir := Sys.readdir "/usr/share/varstored";
+  with Sys_error _ ->
+    begin
+      debug "varstored directory does not exist";
+      Xapi_stdext_unix.Unixext.mkdir_rec "/usr/share/varstored" 0o755
```
edwintorok: The path should probably go into Xapi_globs, to allow overriding it. An entry should be added to nonessential_dirs.

rwdrich:
Don't merge. Need to move away from deprecated package

ocaml/xapi/xapi_vm.ml
```diff
@@ -238,6 +238,29 @@ let assert_memory_constraints ~__context ~vm platformdata =
...
+    end;
+
+  if not (Array.exists (fun e -> e = "PK.auth") !dir) then
+    let file = Unix.openfile "xapi_uefi_certificates.tar" [Unix.O_RDWR; Unix.O_CREAT] 0o755 in
+    let contents = (Xapi_stdext_base64.Base64.decode (Db.Pool.get_uefi_certificates ~__context ~self)) in
+    let hdr = Tar_unix.Header.make "xapi_uefi_certificates.tar" (Int64.of_int (String.length contents)) in
+    Tar_unix.write_block hdr (fun file -> Unixext.really_write_string file contents) file
```
edwintorok: You need to extract the tarfile, but this one creates another tar file.
Instead you should use Tar_unix.extract or extract_gen, see https://github.com/mirage/ocaml-tar/blob/4a6cad9a5e947e5b7c5953c3e4f6a9535a0f593a/unix/tar_unix.ml#L99-L105

ocaml/xapi/xapi_vm.ml
```diff
@@ -256,11 +256,13 @@ let save_uefi_certificates_to_dir ~__context ~self =
...
-    let file = Unix.openfile "xapi_uefi_certificates.tar" [Unix.O_RDWR; Unix.O_CREAT] 0o755 in
-    let contents = (Xapi_stdext_base64.Base64.decode (Db.Pool.get_uefi_certificates ~__context ~self)) in
-    let hdr = Tar_unix.Header.make "xapi_uefi_certificates.tar" (Int64.of_int (String.length contents)) in
-    Tar_unix.write_block hdr (fun file -> Unixext.really_write_string file contents) file;
-    Tar_unix.Archive.extract (fun _ -> !Xapi_globs.varstore_dir) file
+    begin
+      let file = Unix.openfile "xapi_uefi_certificates.tar" [Unix.O_RDWR; Unix.O_CREAT] 0o755 in
+      let contents = (Xapi_stdext_base64.Base64.decode (Db.Pool.get_uefi_certificates ~__context ~self)) in
+      let hdr = Tar_unix.Header.make "xapi_uefi_certificates.tar" (Int64.of_int (String.length contents)) in
+      Tar_unix.write_block hdr (fun file -> Unixext.really_write_string file contents) file;
```
edwintorok: I think these 2 lines may not be required, otherwise you wrap the tar into another tar, and when you extract it you are still left with a tar?

rwdrich: but the contents of the uefi_certificates field is a string, and when I'm extracting from a tar file I need a file descriptor?

edwintorok: You can save the field to the .tar file directly (without using the tar module, it already is a tar).
There are some helpers in unixext for saving files

ocaml/xapi/xapi_vm.ml
```diff
@@ -238,6 +238,35 @@ let assert_memory_constraints ~__context ~vm platformdata =
...
 
+let update_platform_secureboot ~__context ~self platform =
+  if List.assoc "secureboot" platform = "auto" then
+    let t = ("secureboot", string_of_bool (Db.Pool.get_uefi_certificates ~__context ~self <> "")) in
+    List.append [t] (List.remove_assoc "secureboot" platform)
+  else
+    platform
+
+let save_uefi_certificates_to_dir ~__context ~self =
+  let dir = ref [||] in
```
lindig: Why is this a reference? I expected something like
```
let dir = try Sys.readdir !Xapi_globs.varstore_dir with Sys_error -> ..; [| (*empty *) |] in
```
as there doesn't seem to be an update to `dir`.

ocaml/xapi/xapi_vm.ml
```diff
@@ -238,6 +238,35 @@ let assert_memory_constraints ~__context ~vm platformdata =
...
+  if not (Array.exists (fun e -> e = "PK.auth") !dir) then
+    begin
+      let filename = "xapi_uefi_certificates.tar" in
+      try
+        let file = Unix.openfile filename [Unix.O_RDWR; Unix.O_CREAT] 0o755 in
+        let contents = (Xapi_stdext_base64.Base64.decode (Db.Pool.get_uefi_certificates ~__context ~self)) in
+        Unixext.write_string_to_file filename contents;
+        Tar_unix.Archive.extract (fun _ -> !Xapi_globs.varstore_dir) file
+      with _ ->
+        debug "Error writing uefi certificates to file";
```
lindig: This leaks `file` as it is not closed.

ocaml/xapi/xapi_vm.ml
rwdrich: Thank you. I struggled to get the scoping of the exception handling to work for me with the dir being accessed afterwards

ocaml/xapi/xapi_vm.ml
```diff
@@ -238,6 +238,32 @@ let assert_memory_constraints ~__context ~vm platformdata =
...
+  if List.assoc "secureboot" platform = "auto" then
+    let t = ("secureboot", string_of_bool (Db.Pool.get_uefi_certificates ~__context ~self <> "")) in
+    List.append [t] (List.remove_assoc "secureboot" platform)
+  else
+    platform
+
+let save_uefi_certificates_to_dir ~__context ~self =
+  let dir = try Sys.readdir !Xapi_globs.varstore_dir
+    with Sys_error _ ->
+      Xapi_stdext_unix.Unixext.mkdir_rec "/usr/share/varstored" 0o755;
```
lindig: This could still fail if `/usr/share/varstored` is not a directory but an existing file. It might be worth checking the exception more carefully and to fail in such a case.

ocaml/xapi/xapi_vm.ml
```diff
@@ -246,22 +246,26 @@ let update_platform_secureboot ~__context ~self platform =
...
+          let filename = "xapi_uefi_certificates.tar" in
+          try
+            begin
+              let file = Unix.openfile filename [Unix.O_RDWR; Unix.O_CREAT] 0o755 in
+              let contents = (Xapi_stdext_base64.Base64.decode (Db.Pool.get_uefi_certificates ~__context ~self)) in
+              Unixext.write_string_to_file filename contents;
+              Tar_unix.Archive.extract (fun _ -> !Xapi_globs.varstore_dir) file;
+              finally (fun () -> Unix.close file) (fun () -> ())
+            end
+          with _ ->
```
edwintorok: Would be good to know what the error was, there is a `D.log_and_ignore_exn` you could use for that.

ocaml/xapi/xapi_vm.ml
```diff
@@ -238,6 +238,36 @@ let assert_memory_constraints ~__context ~vm platformdata =
...
+            debug "Error writing uefi certificates to file";
+        end
+      with Sys_error _ ->
+        debug "varstored directory does not exist";
+    end
+
```
edwintorok: Would be good to keep track of whether this function succeeded or not, either by making it return `true/false`, or (re)raising an exception when something fails.
For this code it is probably better to reraise the exception, e.g. `failwith "Failed to save UEFI certificates: varstored directory does not exist"`. I think this will get converted to an internal error at the API level, probably worth double checking by triggerring the failure on purpose.

ocaml/xapi/xapi_vm.ml
rwdrich: This with doesn't exist anymore by changing to if's above

ocaml/xapi/xapi_vm.ml
rwdrich: The difficulty comes when the file already exists, and so the function is a noop. Has it succeeded? It technically hasn't written out the file, but it's a positive path through the function

rwdrich:
Don't merge

ocaml/xapi/xapi_vm.ml
```diff
@@ -238,6 +238,32 @@ let assert_memory_constraints ~__context ~vm platformdata =
    same VM twice during memory calculations to determine whether a given VM can start on a particular host..
 *)
 
+let update_platform_secureboot ~__context ~self platform =
+  if List.assoc "secureboot" platform = "auto" then
+    let t = ("secureboot", string_of_bool (Db.Pool.get_uefi_certificates ~__context ~self <> "")) in
+    List.append [t] (List.remove_assoc "secureboot" platform)
```
lindig: `List.append [x] xs` is the same as `x :: xs`

ocaml/xapi/xapi_vm.ml
```diff
@@ -238,6 +238,32 @@ let assert_memory_constraints ~__context ~vm platformdata =
...
+  else
+    platform
+
+let save_uefi_certificates_to_dir ~__context ~self =
+  let err () =
+    debug "Error writing uefi certificates to file";
```
lindig: This would benefit from including the specifics: what file is missing in what directory.

ocaml/xapi/xapi_vm.ml
```diff
@@ -238,6 +238,32 @@ let assert_memory_constraints ~__context ~vm platformdata =
...
+    if not (Sys.file_exists uefi_platform_key) then
+      if Sys.file_exists !Xapi_globs.varstore_dir && Sys.is_directory !Xapi_globs.varstore_dir then
+        let filename = "xapi_uefi_certificates.tar" in
+        Unixext.with_file filename [Unix.O_RDWR; Unix.O_CREAT] 0o755 (fun fd ->
+            let contents = (Xapi_stdext_base64.Base64.decode (Db.Pool.get_uefi_certificates ~__context ~self)) in
+            Unixext.write_string_to_file filename contents;
+            Tar_unix.Archive.extract (fun _ -> !Xapi_globs.varstore_dir) fd)
+      else
+        err ()
+  with Sys_error _ ->
```
edwintorok: Is this try/with required here? I think you replaced all the Sys. function calls with Sys.file_exists which do not raise.

ocaml/xapi/xapi_vm.ml
```diff
@@ -238,6 +238,32 @@ let assert_memory_constraints ~__context ~vm platformdata =
...
+  else
+    platform
+
+let save_uefi_certificates_to_dir ~__context ~self =
+  let err () =
+    debug "Error writing uefi certificates to file";
+    failwith "Failed to save UEFI certificates: varstored directory does not exist or is not a directory" in
+
+  try
+    let uefi_platform_key = !Xapi_globs.varstore_dir ^ "PK.auth" in
```
edwintorok: PK.auth will be there all the time now, replace this with KEK.auth

ocaml/xapi/xapi_vm.ml
rwdrich: Do I want to reraise the exception. What's the expected path for a non-UEFI host? Should there be a condition before the function call, or should the function be a no-op on a BIOS machine (because varstored won't be running, so won't have any dir)?

ocaml/xapi/xapi_vm.ml
```diff
@@ -238,6 +238,35 @@ let assert_memory_constraints ~__context ~vm platformdata =
    same VM twice during memory calculations to determine whether a given VM can start on a particular host..
 *)
 
+let update_platform_secureboot ~__context ~self platform =
+  try
+    match List.assoc "secureboot" platform with
+    | "auto" -> ("secureboot", string_of_bool (Db.Pool.get_uefi_certificates ~__context ~self <> "")) :: (List.remove_assoc "secureboot" platform)
+    | _ -> platform
+  with Not_found -> platform
```
edwintorok: This can also be done by using `| exception Not_found -> platform` instead of wrapping with try/with but that is a recent language feature, we can keep this as is.

ocaml/xapi/xapi_vm.ml
```diff
@@ -238,6 +238,35 @@ let assert_memory_constraints ~__context ~vm platformdata =
...
+  let err () =
+    debug "Error writing uefi certificates to file: varstored directory does not exist or is not a directory";
+    failwith "Failed to save UEFI certificates: varstored directory does not exist or is not a directory"
+  in
+  let uefi_key = !Xapi_globs.varstore_dir ^ "KEK.auth" in
+  if not (Sys.file_exists uefi_key) then
+    if Sys.file_exists !Xapi_globs.varstore_dir && Sys.is_directory !Xapi_globs.varstore_dir then begin
+      let contents = (Xapi_stdext_base64.Base64.decode (Db.Pool.get_uefi_certificates ~__context ~self:pool)) in
+      if contents <> "" then
+        let filename = "xapi_uefi_certificates.tar" in
```
edwintorok: Please delete the file after you're finished extracting it.

edwintorok: Also would be useful to print a debug message here, so we know we've extracted the certificates.

ocaml/xapi/xapi_vm.ml
edwintorok: The directory is created by the host installer when the varstored package is installed, should always be there.
If it is missing nothing will notice that these files are missing, and starting a UEFI VM would fail regardless of secureboot already, so if we want to check anything related to that then secureboot is not the place for it.

Lets drop this whole `else try` branch.

ocaml/xapi/xapi_vm.ml
rwdrich: It's tidier, and I think looks better. I'd like to change it

PR #3816, opened 2019-02-15
CA-309048 handle domain sockets for wsproxy

lindig:
Add support for using Unix domain sockets for wsproxy requests.

### DIFF
--- a/ocaml/xapi/console.ml
+++ b/ocaml/xapi/console.ml
@@ -91,11 +91,9 @@ let ensure_proxy_running () =
 
 let ws_proxy __context req protocol address s =
   let port = match address with
-    | Port p -> p
-    | Path _ ->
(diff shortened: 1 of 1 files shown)


ocaml/xapi/console.ml
```diff
@@ -129,7 +127,7 @@ let ws_proxy __context req protocol address s =
               let wsprotocol = match ty with
                 | Ws_helpers.Hixie76 -> "hixie76"
                 | Ws_helpers.Hybi10 -> "hybi10" in
-              let message = Printf.sprintf "%s:%s:%d" wsprotocol protocol port in
+              let message = Printf.sprintf "%s:%s:%s" wsprotocol protocol port  in
```
robhoes: Now that `port` can be a port or a path, it may be better to give it another name.

PR #3819, opened 2019-02-22
CA-310971: consider enabling the host again after plugging in clustered PBDs in HA mode

edwintorok:
When HA is enabled the host only gets enabled if all its PBDs got
plugged.
However clustered PBDs only get plugged after quorum is established,
which might be after the startup sequence has finished.

There is a watcher thread for that, and when the clustered PBDs do get
plugged in we should attempt to enable the host again.

Rerunning the xenrt test that was consistently failing before on this GFS2 HA test is now passing.


edwintorok:
Need to wait for SIFIX approval before merging to master (also wait for more tests to run)

robhoes:
:shipit:

edwintorok:
release team approved the bugfix, I'll retest this in xenrt once more.

robhoes:
@edwintorok, feel free to merge this when the test is good.

edwintorok:
Nice timing, the testcase just passed 2m ago

PR #3824, opened 2019-03-01
CA-311823: Do not raise alerts for "new CPU features"

robhoes:
These alerts may cause confusion, because sometimes the "new features" are not
actually new hardware features, but ones that are newly exposed by Xen to
guests.

We are keeping the "lost features" alerts, which are far more useful, and also
extend the logging to include the old as well as the new features.

### DIFF
--- a/ocaml/xapi/create_misc.ml
+++ b/ocaml/xapi/create_misc.ml
@@ -519,15 +519,15 @@ let create_host_cpu ~__context host_info =
   if not (is_equal before after) && before <> [||] then begin
     let lost = is_strict_subset (intersect before after) before in
     let gained = is_strict_subset (intersect before after) after in
-    let body = Printf.sprintf "The CPU features of this host have changed.%s%s"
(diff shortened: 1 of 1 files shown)


ocaml/xapi/create_misc.ml
```diff
@@ -519,15 +519,15 @@ let create_host_cpu ~__context host_info =
...
-    in
-    info "%s" body;
+        (string_of_features before);
 
-    if not (Helpers.rolling_upgrade_in_progress ~__context) then
-      let (name, priority) = if lost then Api_messages.host_cpu_features_down else Api_messages.host_cpu_features_up in
+    if not (Helpers.rolling_upgrade_in_progress ~__context) && lost then
+      let (name, priority) = Api_messages.host_cpu_features_down in
       let obj_uuid = Db.Host.get_uuid ~__context ~self:host in
+      let body = Printf.sprintf "The CPU features of this host have changed. Some features have gone away." in
```
edwintorok: This could be a constant defined outside.

PR #3825, opened 2019-03-07
CP-29134 Update versions for Naples release

lindig:
This commit updates version numbers for the Naples release and adds a
record for the Oslo release.

* Database Version: 5.400
* API Version: 2.12
* Branding


ocaml/xapi/xapi_globs.ml
```diff
@@ -52,8 +52,8 @@ let xapi_user_agent = "xapi/"^(string_of_int version_major)^"."^(string_of_int v
  *
  * Please consult the XenCenter maintainers before changing these numbers, because a corresponding change
  * will need to be made in XenCenter *)
-let xencenter_min_verstring = "2.11"
-let xencenter_max_verstring = "2.11"
+let xencenter_min_verstring = "2.12"
+let xencenter_max_verstring = "2.12"
```
edwintorok: Looks like we always bump these 2 together, and consider this is a platform release I think it is right.

PR #3826, opened 2019-03-08
CA-312227: fix extraction of uefi certificates

edwintorok:
Extracting failed if the files were not present.
Have to first create the file and then extract.

Also add a sanity check for the filename: it must not contain slashes
to avoid path traversal attacks.


ocaml/xapi/xapi_vm.ml
```diff
@@ -244,6 +244,14 @@ let update_platform_secureboot ~__context ~self platform =
   | "auto" -> ("secureboot", string_of_bool (Db.Pool.get_uefi_certificates ~__context ~self <> "")) :: (List.remove_assoc "secureboot" platform)
   | _ -> platform
 
+let extract_certificate_file name =
+  if String.contains name '/' then
+	(* Internal error: tarfile not created correctly *)
```
lindig: Indentation seems wider than usual.

edwintorok: Thanks, fixed.

PR #3827, opened 2019-03-13
CP-30614: Use RRD to update memory stats

psafont:
Memory statistics are now collected from rrd files in 2 new modules. These modules follow the same structure as `ocaml/xapi/monitor_pvs_proxy.ml` I'd like to reduce the amount of repeated/shared code in them, I'm open to suggestions.

Left to do: change how domain information is gathered from a string in `ocaml/xapi/monitor_types.ml`

Depends on https://github.com/xapi-project/xcp-rrdd/pull/89

### DIFF
--- a/ocaml/xapi/monitor_mem_host.ml
+++ b/ocaml/xapi/monitor_mem_host.ml
@@ -1,0 +1,83 @@
+(*
+ * Copyright (C) Citrix Systems Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+module Mtxext = Stdext.Threadext.Mutex
+module Lstext = Stdext.Listext.List
+module Mcache = Monitor_dbcalls_cache
+
+module D = Debug.Make(struct let name = "monitor_mem_host" end)
+open D
+
+let get_changes () =
+  let named_dss = List.flatten (List.map (fun filename ->
+      try
+        let datasources = Monitor_types.datasources_from_filename filename in
+        Mcache.log_errors_from filename;
+
+        datasources
+        |> Lstext.filter_map (function
+          | Rrd.Host, ds when List.mem ds.Ds.ds_name ["memory_total_kib"; "memory_free_kib"]
+              -> Some ds
+          | _ -> None (* we are only interested in Host memory stats *)
+          )
+        |> List.map (function ds ->
+          let value =
+            match ds.Ds.ds_value with
+            | Rrd.VT_Int64 v -> Memory.bytes_of_kib v
+            | Rrd.VT_Float v -> Memory.bytes_of_kib (Int64.of_float v)
+            | Rrd.VT_Unknown -> -1L
+          in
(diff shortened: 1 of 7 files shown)


robhoes:
I haven't looked at this in details yet, but my initial thought is that the duplication among the three monitor files could be reduced by using a functor. This is a good exercise in OCaml as well :)

ocaml/xapi/monitor_pvs_proxy.ml
```diff
@@ -60,7 +57,7 @@ let get_changes () =
                 | Rrd.VT_Float v -> int_of_float v
                 | Rrd.VT_Unknown -> -1
               in
-              if ds.Ds.ds_name = Xapi_globs.pvs_proxy_status_ds_name then
+              if ds.Ds.ds_name = "pvscache_status" then
```
robhoes: Why would you "unglobal" this? I think that it is good practice to have constant strings factored out.

psafont: To open a conversation ;)

It's not a "xapi global" constant string at all, it gives the wrong impression that is shared or reused, at this point having it as a global provides indirection without benefit.

It might make sense to put it into xcp-idl and share the value with the plugin that generates the data. Same with the other metrics-related string constants.

robhoes:
And don't forget the final piece of the puzzle: removing the actual `xenctrl` dependency from the `dune` file :)

ocaml/xapi/monitor_mem_host.ml
```diff
@@ -0,0 +1,89 @@
...
+
+module Mtxext = Stdext.Threadext.Mutex
+module Lstext = Stdext.Listext.List
+module Strext = Stdext.Xstringext.String
+module Mcache = Monitor_dbcalls_cache
+
+module D = Debug.Make(struct let name = "monitor_mem_vms" end)
+open D
+
+module StringSet = Set.Make(struct type t = string let compare = compare end)
```
lindig:
```
module StringSet = Set.Make(String)
```

ocaml/xapi/monitor_mem_vms.ml
```diff
@@ -0,0 +1,81 @@
...
+
+module Mtxext = Stdext.Threadext.Mutex
+module Lstext = Stdext.Listext.List
+module Strext = Stdext.Xstringext.String
+module Mcache = Monitor_dbcalls_cache
+
+module D = Debug.Make(struct let name = "monitor_mem_vms" end)
+open D
+
+module StringSet = Set.Make(struct type t = string let compare = compare end)
```
lindig: see above

ocaml/xapi/monitor_types.ml
```diff
@@ -15,8 +15,7 @@
  * @group Performance Monitoring
 *)
 
-open Stdext
-open Xstringext
+module Strext = Stdext.Xstringext.String
```
mseri: Here you use it, but you may as well ditch it and use `Astring`, as we did in few other modules already (with the intention of killing `Stdext.Xstringext` once we were able to)

lindig:
Is this ready to be merged?

psafont:
Please wait, I found more updateable packages that do not seem to break compatibility.

ocaml/xapi/monitor_mem_host.ml
```diff
@@ -0,0 +1,83 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+module Mtxext = Stdext.Threadext.Mutex
+module Lstext = Stdext.Listext.List
+module Mcache = Monitor_dbcalls_cache
+
+module D = Debug.Make(struct let name = "monitor_mem_vms" end)
```
robhoes: Wrong name string here.

PR #3828, opened 2019-03-13
XSI-132 CA-312644 CA-299554 update dom0 vcpu count

lindig:
This complements XSI-132 CA-312644 and commit

  4a2cc12d1a72854174b25773f510afcfa8c46aad

The number of vCPUs of dom0 need to be updated on xapi restart in the
dom0 metrics which usually are only written on first boot.

### DIFF
--- a/ocaml/xapi/create_misc.ml
+++ b/ocaml/xapi/create_misc.ml
@@ -166,6 +166,11 @@ let minimum x y = if x < y then x else y
 
 let (+++) = Int64.add
 
+(** [mkints n] creates a list [1; 2; .. ; n] *)
+let rec mkints = function
+  | 0 -> []
+  | n -> (mkints (n - 1) @ [n])
+
 (** Ensures that the database has all the necessary records for domain *)
 (** zero, and that the records are up-to-date. Includes the following: *)
 (**     1. The domain zero record.                                     *)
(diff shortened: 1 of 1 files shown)


ocaml/xapi/create_misc.ml
```diff
@@ -348,9 +348,13 @@ and update_domain_zero_record ~__context ~domain_zero_ref (host_info: host_info)
     Db.VM.set_requires_reboot ~__context ~self:domain_zero_ref ~value:false
   end;
   let localhost = Helpers.get_localhost ~__context in
+  let metrics   = Db.VM.get_metrics ~__context ~self:domain_zero_ref in
   Db.VM.set_power_state ~__context ~self:domain_zero_ref ~value:`Running;
   Db.VM.set_domid ~__context ~self:domain_zero_ref ~value:0L;
-  Helpers.update_domain_zero_name ~__context localhost host_info.hostname
+  Helpers.update_domain_zero_name ~__context localhost host_info.hostname;
+  Db.VM_metrics.set_VCPUs_number ~__context ~self:metrics
```
robhoes: You also need to update `VM.VCPUs_max`, `VM.VCPUs_at_startup` and `VM.VCPUs_params`. See where all `vcpus` is used in `create_domain_zero_record`.

lindig: `VM.VCPUs_params` is a list. But what about `vCPUs_utilisation`?

robhoes: Oh yes, not VCPU_params. But yes to `VCPU_utilisation` in the metrics.

robhoes: Cleaner would be to update the metrics from `ensure_domain_zero_metrics_record` (perhaps add `update_domain_zero_metrics_record`, since all those function are split between the main record and the metrics.

lindig: I've created `update_domain_zero_metrics_record` but I don't see how `vCPUs_utilisation` could be updated.

lindig: I believe this has been addressed.

ocaml/xapi/create_misc.ml
```diff
@@ -196,9 +201,12 @@ and ensure_domain_zero_record ~__context (host_info: host_info): [`VM] Ref.t =
   let (domain_zero_ref, found) =
     try       ref_lookup (), true
     with _ -> ref_create (), false in
-  if found
-  then update_domain_zero_record ~__context ~domain_zero_ref host_info
-  else create_domain_zero_record ~__context ~domain_zero_ref host_info;
+  if found then begin
+    update_domain_zero_record ~__context ~domain_zero_ref host_info;
+    update_domain_zero_metrics_record ~__context ~domain_zero_ref
```
robhoes: Let's add this to `ensure_domain_zero_metrics_record` instead to avoid more 🍝

lindig: Done.

PR #3829, opened 2019-03-14
CP-29682: changed the checksum algorithm from SHA1 to xxHash

RoryBG:
Backwards compatability is maintained and xxHash has been vendored for this PR.

### DIFF
--- a/ocaml/xxhash/lib/XXHash.ml
+++ b/ocaml/xxhash/lib/XXHash.ml
@@ -1,0 +1,91 @@
+(*
+Copyright (c) 2016 Pieter Goetschalckx
+
+Permission is hereby granted, free of charge, to any person obtaining
+a copy of this software and associated documentation files (the
+"Software"), to deal in the Software without restriction, including
+without limitation the rights to use, copy, modify, merge, publish,
+distribute, sublicense, and/or sell copies of the Software, and to
+permit persons to whom the Software is furnished to do so, subject to
+the following conditions:
+
+The above copyright notice and this permission notice shall be
+included in all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*)
+
+exception XXH_error
+
+module type XXHASH = sig
+  type hash
+  type state
+
+  val hash : ?seed:hash -> string -> hash
+  val reset : ?seed:hash -> state -> unit
+  val update : state -> string -> unit
+  val digest : state -> hash
+
+  val with_state : ?seed:hash -> (state -> unit) -> hash
+end
+
+let check errorcode =
+  if errorcode != 0 then raise XXH_error
+
+module Make (Bindings : Xxhash_bindings.BINDINGS) = struct
+
+  type hash = Bindings.hash
+  type state = Bindings.state
+
+  let hash ?(seed=Bindings.default_seed) input =
+    let length = String.length input |> Unsigned.Size_t.of_int in
+    Bindings.internal_of_hash seed
+    |> Bindings.hash input length
+    |> Bindings.hash_of_internal
+
+  let create () =
+    Bindings.create ()
+
+  let reset ?(seed=Bindings.default_seed) state =
+    Bindings.internal_of_hash seed
+    |> Bindings.reset state
+    |> check
+
+  let update state input =
+    let length = String.length input |> Unsigned.Size_t.of_int in
+    Bindings.update state input length
+    |> check
+
+  let digest state =
+    Bindings.digest state |> Bindings.hash_of_internal
+
+  let free state =
+    Bindings.free state |> check
+
+  let with_state ?(seed=Bindings.default_seed) f =
+    let with_state_unsafe state =
+      reset ~seed state;
+      f state;
+      digest state
+    in
+
+    let state = create () in
+    
+    let output =
+      Xapi_stdext_pervasives.Pervasiveext.finally
+      (fun () -> with_state_unsafe state)
+      (fun () -> free state)
+    in
+    output
+    
+end
+
+module C = Xxhash_bindings.C (Xxhash_generated)
+module XXH32 : (XXHASH with type hash = nativeint) = Make (C.XXH32)
+module XXH64 : (XXHASH with type hash = int64) = Make (C.XXH64)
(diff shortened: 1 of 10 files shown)


RoryBG:
Both libxxhash0 and libxxhash-dev are not in Debian 9 so Travis fails to build, the other build fails because xxHash is not in the spec file yet (awaiting a bitbucket merge).

edwintorok:
I think you will need to update the build dependencies on one of the .opam files in the repo to install xxhash C library, otherwise the xs-opam build fails (the xenserver one is expected to fail until your other PR is merged):
```
E: Unable to locate package libxxhash-dev
E: Unable to locate package libxxhash0
```

lindig:
* `xxhash-devel` has been added to the spec file as a dependency. 
* `libxxhash-dev ` need to be added to `.opam` as a depext depency to make it available for Travis.

ocaml/xxhash/lib/XXHash.ml
```diff
@@ -0,0 +1,61 @@
+exception XXH_error
```
lindig: Please add a copyright notice.

ocaml/xxhash/lib/XXHash.ml
```diff
@@ -0,0 +1,61 @@
...
+  val digest : state -> hash
+  val free : state -> unit
+
+  val with_state : ?seed:hash -> (state -> unit) -> hash
+end
+
+let check errorcode =
+  if errorcode != 0 then raise XXH_error
+
+module XXHash (Bindings : Xxhash_bindings.BINDINGS) = struct
```
lindig: This is module `XXHash.XXHash`. It would be more idiomatic to call this `Make`, such that we get `XXHash.Make(Bindings)`.

ocaml/xxhash/lib/XXHash.ml
```diff
@@ -0,0 +1,61 @@
...
+    |> check
+
+  let digest state =
+    Bindings.digest state |> Bindings.hash_of_internal
+
+  let free state =
+    Bindings.free state |> check
+
+  let with_state ?(seed=Bindings.default_seed) f =
+    let state = create () in
```
lindig: If any of the functions can raise an exception, the free call could be missed. It it probably better to use `finally` to make sure `free` is always called`.

ocaml/xxhash/lib/XXHash.mli
```diff
@@ -0,0 +1,46 @@
...
+
+module type XXHASH = sig
+  type hash
+  type state
+
+  (** {2 Short input} *)
+
+  (** [hash ~seed input] returns the hash of [input]. This is equivalent to
+      {[let state = create () in
+reset ~seed state;
```
lindig: Fix indentation.

RoryBG:
> Since `xapi.opam` was updated with the library dependencies, it's not immediately obvious to me why Travis would not succeed. Is Travis not respecting depext dependencies?

It looks like Travis is running on Ubuntu 14.04.3 LTS, I think that means its running a pretty old version of Debian. From what I can see [here](https://pkgs.org/download/xxhash), it seems that xxHash was recently added to Debian 10. So maybe this is why Travis is unable to find them.

lindig:
Going to Travis, it can't find the build for me. Could you post a link to the Travis build or trigger a new one?

RoryBG:
> Going to Travis, it can't find the build for me. Could you post a link to the Travis build or trigger a new one?

[Here is the most recent build. ](https://travis-ci.org/xapi-project/xen-api/builds/506381790)

lindig:
The distro being used for the `xs-opam` build is defined in `.travis-xs-opam.sh`, which is currently `debian-9-ocaml-4.07` and does not have`xxhash`. The distros available are: https://hub.docker.com/r/ocaml/opam2/tags. Is xxhash in one of the Ubuntus? It is a unfortunate that we we might have to change the docker image for `xxhash`.

RoryBG:
> Is xxhash in one of the Ubuntus? It is a unfortunate that we we might have to change the docker image for `xxhash`

xxHash is in Ubuntu 18.10 and onwards. Which means it is not yet in any LTS version of Ubuntu (and I think the next Ubuntu LTS release will be 20.04 in a year's time). 

This means we might have to update the docker image to a non-LTS Ubuntu version to get Travis to work, though I think [pinning](https://help.ubuntu.com/community/PinningHowto) might be able to fix the package issue without needing to upgrade the image.

lindig:
I believe my suggestion above works - I downloaded the image and it contains OCaml 4.07 and it can install `libxxhash-dev`.

ocaml/xxhash/lib/XXHash.ml
```diff
@@ -0,0 +1,94 @@
...
+  type state
+
+  val hash : ?seed:hash -> string -> hash
+  val create : unit -> state
+  val reset : ?seed:hash -> state -> unit
+  val update : state -> string -> unit
+  val digest : state -> hash
+  val free : state -> unit
+
+  val with_state : ?seed:hash -> (state -> unit) -> hash
```
lindig: I believe there should be no public way to create state such that `with_state` is the only way to obtain and release state; hence, `create` and `free` should be private and not show up in `XXHASH`.

ocaml/xxhash/lib/XXHash.mli
```diff
@@ -0,0 +1,48 @@
...
+  val reset : ?seed:hash -> state -> unit
+
+  (** [update state input] adds [input] to [state]. *)
+  val update : state -> string -> unit
+
+  (** [digest state] returns the hash all input added to [state]. *)
+  val digest : state -> hash
+
+  (** [free state] releases the memory used by [state]. *)
+  val free : state -> unit
```
lindig: See above.

ocaml/xxhash/stubs/xxhash_bindings.ml
```diff
@@ -0,0 +1,71 @@
...
+    let update = F.foreign "XXH32_update" (ptr state_t @-> string @-> size_t @-> returning int)
+    let digest = F.foreign "XXH32_digest" (ptr state_t @-> returning uint)
+  end
+
+  module XXH64 = struct
+    type hash = int64
+    type internal = Unsigned.ullong
+
+    let default_seed = 0L
+    let hash_of_internal = Unsigned.ULLong.to_int64
```
lindig: What is the purpose of the internal type and is the conversion between `hash` and `internal` guaranteed to be lossless and if not, why would that not be a problem? I'm worried that this could corrupt a hash.

lindig:
I think this looks good. I have one question about the `internal` type. Would like @robhoes and @psafont to take a look, too, before we merge this.

ocaml/xxhash/stubs/xxhash_bindings.ml
RoryBG: The conversion between `hash` and `internal` is guaranteed to be lossless as both types are guaranteed to be 64-bits in size.

I'm not 100% sure as to the purpose of the type though, could be there to improve readability.

RoryBG:
Once @robhoes is happy I'll rebase the branch ready for the merge.

lindig:
Please rebase - we will merge this if Travis is green.

RoryBG:
Travis currently fails on the `python-nosetests` build due to a public key error.

lindig:
The python tests don't worry me; this looks like a problem on Travis side.

lindig:
Are you sure the rebase is correct? There are too many commits in this.

RoryBG:
> 
> 
> Are you sure the rebase is correct? There are too many commits in this.

Looks like I accidently rebased onto `master` rather than onto the `checksum` branch, which has pulled in a bunch of commits from master. I'll fix it.

RoryBG:
I have now squashed the commits so this can be merged.

PR #3831, opened 2019-03-19
CA-311705: Add VDI usage checking for metadata backup scripts.

lindig:
VM metadata backup will fail with no errors because of the VDI
is running out of space.

- Add backup VDI usage check before perform the backup, return error
if usage greater than the usage alert.(Default 90%)
- Add '-f' option to ingore the checking.

Backport of 4ac33cd52a1cf67c445c52db25c340a12e13bc0a


lindig:
Ping @minli1 - could you review this, please?

PR #3832, opened 2019-03-19
CP-30756: Replace Base64 library

psafont:
Now we use Base64 3.1.0 instead of using the extended custom library.

The encoding functions should be safe to replace.

For the decoding there are 2 possible issues:
* The new calls could fail more frequently because not they are not sanitized anymore. (especially for gpumon)
* The exception from the call in ocaml/xapi/xapi_gpumon.ml, Line 103 are not caught: how to report the failing call?

### DIFF
--- a/ocaml/xapi/cli_operations.ml
+++ b/ocaml/xapi/cli_operations.ml
@@ -3097,7 +3097,7 @@ let host_license_add fd printer rpc session_id params =
   | Some license ->
     debug "Checking license [%s]" license;
     (try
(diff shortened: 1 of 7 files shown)


mseri:
Do extremely careful testing of this change, not just BVT&BST. When we cleaned up stedext last year we did not remove Base64 because something was deeply relying on the sanitization and started failing reliably. It may be that is no longer an issue though.

psafont:
New changes pushed to accommodate guefi code that usess base64.

PR #3838, opened 2019-04-04
CA-314290: Allow to specify SMBIOS type2 info from the toolstack

TalonsLee:
The SMBIOS type 2 info is required for some guest app to run properly.
This commit let the toolstack accept those strings and write to xenstore.


krizex:
The code changes looks good to me.
Please describe the problem and your solution in the commit message.

lindig:
Please provide a log message that summarizes what this does beyond the subject line. The fix is not not so small that it would be obvious. One grammar aspect: "Allow *to* specify SMBIOS type2 info from *the* toolstack".

PR #3839, opened 2019-04-05
CA-312226 XSI-251 clear unexpected vGPU metadata on shutdown

lindig:
A vGPU is supposed to hold vGPU compatibility metadata only while it is
suspended. Beyond that, it can deadlock the start of other machines as
it causes gpumon being contacted in vgpu_pgpu_are_compatible() while
gpumon might be is stopped as part of VM.start of another VM. This
commit adds code to detect, log, and remove this metadata when a VM is
halted to avoid the problem.  We still need to find the root cause why
the situation can arise in the first place.
    
### DIFF
--- a/ocaml/xapi/xapi_vm_lifecycle.ml
+++ b/ocaml/xapi/xapi_vm_lifecycle.ml
@@ -619,7 +619,15 @@ let force_state_reset_keep_current_operations ~__context ~self ~value:state =
       (Db.VM.get_attached_PCIs ~__context ~self);
     List.iter
       (fun vgpu ->
-         Db.VGPU.set_currently_attached ~__context ~self:vgpu ~value:false)
+         Db.VGPU.set_currently_attached ~__context ~self:vgpu ~value:false;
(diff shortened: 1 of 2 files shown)


ocaml/xapi/xapi_vm_lifecycle.ml
```diff
@@ -655,7 +655,15 @@ let force_state_reset_keep_current_operations ~__context ~self ~value:state =
     |> List.iter (fun vgpu ->
         Db.VGPU.set_resident_on ~__context ~self:vgpu ~value:Ref.null;
         Db.VGPU.set_scheduled_to_be_resident_on
-          ~__context ~self:vgpu ~value:Ref.null)
+          ~__context ~self:vgpu ~value:Ref.null;
+        let keys = Db.VGPU.get_compatibility_metadata ~__context ~self:vgpu in
+        if state = `Halted && List.mem_assoc Xapi_gpumon.Nvidia.key keys then begin
```
robhoes: I think that we should move this into the `state = `Halted` conditional above.

PR #3840, opened 2019-04-08
CA-314317: Protect PVS-cache get_or_recreate_vdi by mutex

robhoes:
The function `Pvs_cache_vdi.get_or_recreate_vdi` may be called for multiple
concurrently starting VMs. Without the mutex, this can lead to multiple VDIs
being created, while the should only be one for a given `PVS_cache_storage`
object.

### DIFF
--- a/ocaml/xapi/pvs_cache_vdi.ml
+++ b/ocaml/xapi/pvs_cache_vdi.ml
@@ -64,15 +64,19 @@ let get_vdi ~__context ~self =
     end
   end
 
+let m = Mutex.create ()
+
 let get_or_recreate_vdi ~__context ~self =
-  match get_vdi ~__context ~self with
-  | None ->
-    let sR = Db.PVS_cache_storage.get_SR ~__context ~self in
-    let size = Db.PVS_cache_storage.get_size ~__context ~self in
-    let vdi = create_vdi ~__context ~sR ~size in
-    Db.PVS_cache_storage.set_VDI ~__context ~self ~value:vdi;
-    vdi
-  | Some vdi -> vdi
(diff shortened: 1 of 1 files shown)


robhoes:
Draft, because I am still running some tests.

ocaml/xapi/pvs_cache_vdi.ml
```diff
@@ -64,15 +64,19 @@ let get_vdi ~__context ~self =
...
-  | None ->
-    let sR = Db.PVS_cache_storage.get_SR ~__context ~self in
-    let size = Db.PVS_cache_storage.get_size ~__context ~self in
-    let vdi = create_vdi ~__context ~sR ~size in
-    Db.PVS_cache_storage.set_VDI ~__context ~self ~value:vdi;
-    vdi
-  | Some vdi -> vdi
+  Stdext.Threadext.Mutex.execute m (fun () ->
+    match get_vdi ~__context ~self with
+    | None ->
```
lindig: Could we restrict this to the `None` case?

robhoes: Indeed, `get_vdi` needs to be under the mutex.

edwintorok:
There is also a destroy_vdi with similar logic, perhaps worth using the same mutex there as well? Otherwise you might race between creating and destroying the VDI

robhoes:
@edwintorok If the destroy and create functions are called at the same time, then we would likely have problems higher up, which won't be solved by this mutex.

robhoes:
Ehm... this was in "draft" state because I was still testing.

PR #3841, opened 2019-04-11
REQ-720: CP-31058: Datamodel changes for multiple Nvidia VGPU support

minli1:
- Add compatible_types_in_vm and compatible_types_on_pgpu into
   vgpu_type, show the first one in parameter list.
 - Change to use new way to read from NVIDIA config file and
   initilize compatible* list.
 - Add device as an option paramater in vgpu.create and show
   in parameter list.
 - Update test cases accordingly.

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -4823,7 +4823,9 @@ module VGPU_type = struct
         field ~qualifier:DynamicRO ~ty:(Set (Ref _gpu_group)) ~lifecycle:[Published, rel_vgpu_productisation, ""] "enabled_on_GPU_groups" "List of GPU groups in which at least one have this VGPU type enabled";
         field ~qualifier:StaticRO ~ty:implementation ~lifecycle:[Published, rel_dundee, ""] ~default_value:(Some (VEnum "passthrough")) "implementation" "The internal implementation of this VGPU type";
         field ~qualifier:StaticRO ~ty:String ~lifecycle:[Published, rel_dundee, ""] ~default_value:(Some (VString "")) "identifier" "Key used to identify VGPU types and avoid creating duplicates - this field is used internally and not intended for interpretation by API clients";
-        field ~qualifier: StaticRO ~ty:Bool ~lifecycle:[Published, rel_dundee, ""] ~default_value:(Some (VBool false)) "experimental" "Indicates whether VGPUs of this type should be considered experimental";
+        field ~qualifier:StaticRO ~ty:Bool ~lifecycle:[Published, rel_dundee, ""] ~default_value:(Some (VBool false)) "experimental" "Indicates whether VGPUs of this type should be considered experimental";
+        field ~qualifier:DynamicRO ~ty:(Set String) ~lifecycle:[Published, rel_plymouth, ""] ~default_value:(Some (VSet [])) "compatible_types_in_vm" "List of VGPU types which are compatible in one VM";
+        field ~qualifier:DynamicRO ~ty:(Set String) ~lifecycle:[Published, rel_plymouth, ""] ~internal_only:true ~default_value:(Some (VSet [])) "compatible_types_on_pgpu" "List of VGPU types which are compatible on one PGPU";
       ]
       ()
 end
--- a/ocaml/xapi/xapi_vgpu_type.ml
+++ b/ocaml/xapi/xapi_vgpu_type.ml
@@ -102,6 +102,8 @@ type vgpu_type = {
   internal_config : (string * string) list;
   identifier : Identifier.t;
   experimental : bool;
+  compatible_types_in_vm : string list;
+  compatible_types_on_pgpu : string list;
 }
 
 let passthrough_gpu = {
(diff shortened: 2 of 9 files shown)


ocaml/idl/datamodel.ml
```diff
@@ -4823,7 +4823,9 @@ module VGPU_type = struct
...
         field ~qualifier:StaticRO ~ty:String ~lifecycle:[Published, rel_dundee, ""] ~default_value:(Some (VString "")) "identifier" "Key used to identify VGPU types and avoid creating duplicates - this field is used internally and not intended for interpretation by API clients";
-        field ~qualifier: StaticRO ~ty:Bool ~lifecycle:[Published, rel_dundee, ""] ~default_value:(Some (VBool false)) "experimental" "Indicates whether VGPUs of this type should be considered experimental";
+        field ~qualifier:StaticRO ~ty:Bool ~lifecycle:[Published, rel_dundee, ""] ~default_value:(Some (VBool false)) "experimental" "Indicates whether VGPUs of this type should be considered experimental";
+        field ~qualifier:DynamicRO ~ty:(Set String) ~lifecycle:[Published, rel_plymouth, ""] ~default_value:(Some (VSet [])) "compatible_types_in_vm" "List of vgpu types that can start on one VM";
+        field ~qualifier:DynamicRO ~ty:(Set String) ~lifecycle:[Published, rel_plymouth, ""] ~internal_only:true ~default_value:(Some (VSet [])) "compatible_types_on_pgpu" "List of vgpu types that compatible on one PGPU";
```
lindig: The documentation "List of vgpu types that compatible on one PGPU" is lacking a verb and makes no sense as it is.

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -280,6 +297,7 @@ module Nvidia_old = struct
             (fun pdev_id -> pdev_id, None)
       in
       (* NVIDIA key is "device_id:subdevice_id", N.B. not subvendor id *)
+      (* Since old config file don't include multiple or comptible list, define empty list here *)
```
edwintorok: s/comptible/compatible

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -467,24 +490,25 @@ module Vendor = functor (V : VENDOR) -> struct
       ~is_pci_hidden =
     let vgpu_types = make_vgpu_types ~__context ~pci in
     (* Temporarily fall back to old nvidia module *)
-    if vendor_id = Nvidia_old.vendor_id then begin
+    info "Using new Nvidia config file";
+    (* if vendor_id = Nvidia_old.vendor_id then begin
```
edwintorok: Does commenting this out have any implication for updates from an older version?
Or did we drop the Nvidia_old.vendor_id cards from the HCL?

ocaml/xapi/xapi_vgpu_type.ml
minli1: Thanks, @edwintorok 
Will update.

ocaml/idl/datamodel.ml
minli1: Thanks. @lindig Will update.

ocaml/xapi/xapi_vgpu_type.ml
minli1: Actually it's not related to dropping any nvidia_old cards from HCL. Here the old means just the way to read from Nvidia config file as Nvidia changed their config file from separated to one.
So I just enable the new way to read config(b66e46cc38e40ac6f0a1d3f3576575b09f49825c).

ocaml/idl/datamodel.ml
```diff
@@ -4823,7 +4823,9 @@ module VGPU_type = struct
...
         field ~qualifier:StaticRO ~ty:implementation ~lifecycle:[Published, rel_dundee, ""] ~default_value:(Some (VEnum "passthrough")) "implementation" "The internal implementation of this VGPU type";
         field ~qualifier:StaticRO ~ty:String ~lifecycle:[Published, rel_dundee, ""] ~default_value:(Some (VString "")) "identifier" "Key used to identify VGPU types and avoid creating duplicates - this field is used internally and not intended for interpretation by API clients";
-        field ~qualifier: StaticRO ~ty:Bool ~lifecycle:[Published, rel_dundee, ""] ~default_value:(Some (VBool false)) "experimental" "Indicates whether VGPUs of this type should be considered experimental";
+        field ~qualifier:StaticRO ~ty:Bool ~lifecycle:[Published, rel_dundee, ""] ~default_value:(Some (VBool false)) "experimental" "Indicates whether VGPUs of this type should be considered experimental";
+        field ~qualifier:DynamicRO ~ty:(Set String) ~lifecycle:[Published, rel_plymouth, ""] ~default_value:(Some (VSet [])) "compatible_types_in_vm" "List of VGPU types in which are compatible in one VM";
```
lindig: _List of VGPU types **in** which are compatible in one VM_ - drop the "in", it does not make sense (to me). Also on the next line. If you think the "in" is important, please explain what you would like to express and we will try to find a concise documentation.

minli1: Yes, you are right. Need not "in" here. Thanks.
 Updated.

ocaml/idl/datamodel.ml
```diff
@@ -4823,7 +4823,9 @@ module VGPU_type = struct
...
         field ~qualifier:StaticRO ~ty:implementation ~lifecycle:[Published, rel_dundee, ""] ~default_value:(Some (VEnum "passthrough")) "implementation" "The internal implementation of this VGPU type";
         field ~qualifier:StaticRO ~ty:String ~lifecycle:[Published, rel_dundee, ""] ~default_value:(Some (VString "")) "identifier" "Key used to identify VGPU types and avoid creating duplicates - this field is used internally and not intended for interpretation by API clients";
-        field ~qualifier: StaticRO ~ty:Bool ~lifecycle:[Published, rel_dundee, ""] ~default_value:(Some (VBool false)) "experimental" "Indicates whether VGPUs of this type should be considered experimental";
+        field ~qualifier:StaticRO ~ty:Bool ~lifecycle:[Published, rel_dundee, ""] ~default_value:(Some (VBool false)) "experimental" "Indicates whether VGPUs of this type should be considered experimental";
+        field ~qualifier:DynamicRO ~ty:(Set String) ~lifecycle:[Published, rel_plymouth, ""] ~default_value:(Some (VSet [])) "compatible_types_in_vm" "List of VGPU types which are compatible in one VM";
```
robhoes: I just noticed that the type here and below is `Set String`. I think we want `Set (Ref _vgpu_type)`.

minli1: Yes, my first thought is also Ref vgpu_type, but actually we can't have the Ref before the vgpu_type created.
The compatible* were initialized during vgpu_type creation.
Like we would like to have A,B,C into type A's compatible list but actually B and C may not created now.

robhoes: You have to then create all of them first, and then fill in all of the compatible fields, which isn't too hard. It's not a problem yet anyway, because `A` will now only have `[A]` or `[]`.

minli1: OK, I will make some changes in the coming commits. Thanks.

PR #3842, opened 2019-04-12
Backport of HFP-510

minli1:
This backport includes HFX-2242 and HFX-2263,
As there are large conflicts between master and dundee-lcm, some functions are missing in the old branch.
I have to backport some dependencies. 
See the commit message or the CP-31010 ticket for details.

### DIFF
--- a/ocaml/xapi/xapi_guest_agent.ml
+++ b/ocaml/xapi/xapi_guest_agent.ml
@@ -164,7 +164,7 @@ let get_initial_guest_metrics (lookup: string -> string option) (list: string ->
   {pv_drivers_version; os_version; networks; other; memory; device_id; last_updated; can_use_hotplug_vbd; can_use_hotplug_vif;}
 
 
-let create_and_set_guest_metrics (lookup: string -> string option) (list: string -> string list) ~__context ~domid ~uuid =
+let create_and_set_guest_metrics (lookup: string -> string option) (list: string -> string list) ~__context ~domid ~uuid ~pV_drivers_detected =
   let initial_gm = get_initial_guest_metrics lookup list
   in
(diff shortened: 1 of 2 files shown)


minli1:
Travis failed due to:
```
+ eval make
++ make
omake -j 8 phase1
make: omake: Command not found
make: *** [all] Error 127
The command "bash travis-build-repo.sh" exited with 2.
```
I think it's not related to the backport, build successfully with hg way.

lindig:
Don't worry about Travis - we can't expect it to work on this branch unless we put a lot of effort into creating a docker container that has all dependencies.

lindig:
This is not HFX-510 but HFP-510.

lindig:
I think we should re-consider backporting HFX-2242 because of its dependencies. It seems to me that the amount of code required is so large that it introduces too much risk.

lindig:
Why is `HFX-2263- XOP-830: Fix the resynchronisation logic on xapi restart` required for `HFX-2263,CA-302194 XSI-87 apply guest agent config on start`?

minli1:
Because of function `resync_all_vms ` is missing before that.
```
This is backport of 6e705d3, to resolve dependency of
2ccd151.
```
@lindig

lindig:
I think this is wrong. The point of 2ccd151ed5aedc1061efeb8f5806a47828a1cadf is to call `apply_guest_agent_config()`. This code is already in place - so the bug does not exist in Dundee. Indeed, it appears that this code was lost when `resync_all_vms()` was introduced and it had to be re-introduced. So I would suggest to remove both commits for HFX-2263.

lindig:
There is a discussion that HFX-2242 could be fixed in a different and much simpler way. See HFP-510.

minli1:
Close this PR as I am making this with a new one.

PR #3844, opened 2019-04-15
HFP-510 for Dundee-lcm.

minli1:
Backport for
- HFX-2242
- HFX-2263


ocaml/xapi/xapi_xenops.ml
```diff
@@ -2023,6 +2024,11 @@ let on_xapi_restart ~__context =
 		Db.VM.set_resident_on ~__context ~self:vm ~value:Ref.null;
 	) xapi_vms_not_in_xenopsd
 
+  info "applying guest agent configuration during restart";
```
lindig: I doubt this compiles because you need a semicolon before this sequence.

minli1: Yes, updated.

PR #3846, opened 2019-04-16
REQ-720: CP-31058: Update Nvidia data structures according to IDL change

minli1:
- Add 'virtual_pci_address' to implementation.
 - Add 'type_id' to internal_config.
 - Delete Manual setup checking of vgpus_of_vm.
 - Delete old way of getting Nvidia config.


ocaml/xapi/xapi_xenops.ml
```diff
@@ -647,29 +647,25 @@ module MD = struct
     let pci_address = Db.PCI.get_pci_id ~__context ~self:pci in
     Xenops_interface.Pci.address_of_string pci_address
 
+  let get_virtual_pci_address ~__context vgpu =
+   let device = vgpu.Db_actions.vGPU_device in
+   let virtual_pci_address  =  Xapi_globs.vgpu_pci_prefix ^ device ^".0" in
```
rwdrich: Does this rely on the device always being plugged into slot 0? Because it's in the prefix of the address. What will we do in the cases of multiple vgpus, or we want to access a vgpu in another slot?

minli1: Yes, since it's virtual PCI address which xapi defines here and pass them to DEMU.
Previously, when only one vgpu allowed, the virtual PCI address is defined in DEMU with slot always set 0. So after discussion, we decided to define virtual PCI address by xapi now and DEMU will handle the slot number.

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -64,13 +65,14 @@ module Identifier = struct
       match id with
       | Passthrough -> "passthrough"
       | Nvidia nvidia_id ->
-        Printf.sprintf "nvidia,%04x,%s,%04x,%04x"
+        Printf.sprintf "nvidia,%04x,%s,%04x,%04x,%s"
```
lindig: The existing parameters are treated as hexadecimal numbers. Why are we using a string here for the new parameter?

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -288,11 +290,14 @@ module Nvidia_old = struct
                   (List.assoc "plugin0.num_heads" args) in
               let max_instance = Int64.of_string
                   (List.assoc "plugin0.max_instance" args) in
+              (* Define "0" here because now we are using new way to read from nvidia conf file  *)
+              let type_id = "0" in
```
lindig: Should this not be an integer that is converted to a string representation when needed?

ocaml/xapi/xapi_vgpu_type.ml
minli1: Hi, @lindig 
Actually type_id are not hexadecimal numbers.
See the `vgpuConfig.xml` file:
```
  <vgpuType id="11" name="GRID M60-0B" class="NVS">
    <devId vendorId="0x10de" deviceId="0x13F2" subsystemVendorId="0x10de" subsystemId="0x1176"></devId>
    <framebuffer>0x1A000000</framebuffer>
    <numHeads>2</numHeads>
    <display width="2560" height="1600"></display>
    <mappableVideoSize>0x400000</mappableVideoSize>
    <fbReservation>0x6000000</fbReservation>
    <frlConfig>0x2d</frlConfig>
    <cudaEnabled>0x0</cudaEnabled>
    <multiVgpuSupported>0x0</multiVgpuSupported>
    <encoderCapacity>0x64</encoderCapacity>
    <bar1Length>0x100</bar1Length>
    <frame_rate_limiter>1</frame_rate_limiter>
    <license>GRID-Virtual-PC,2.0;GRID-Virtual-WS,2.0;Quadro-Virtual-DWS,5.0;GRID-Virtual-WS-Ext,2.0</license>
```
Other number fields like deviceID are defined as hexadecimal numbers and start with 0x, but `vgpuType id ` is not treated as number here, I think string can be flexbility in the future?
Thanks.

ocaml/xapi/xapi_vgpu_type.ml
minli1: Here is the old way of getting the config from Nvidia. I think now they are deprecated. I just make the compile pass and leave a comment above. We could delete the old way of this actually :)

ocaml/xapi/xapi_vgpu_type.ml
lindig: ok. It's not obvious what value ID can take, so let's keep the string.

johnelse: Hello, I am a bit rusty on how vGPU works now, but the format of the VGPU_type IDs is described [here](http://xapi-project.github.io/xapi/futures/vgpu-type-identifiers.html).

If you need to change the format of the identifier for existing VGPU_types you'll probably want to bump `VGPU_type.Identifier.version` and implement some upgrade logic. The identifier string is used on import as a primary key for identifying VGPU_types, as identical VGPU_types will have different UUIDs across different pools.

minli1: :) Many thanks for your useful comment. 
- Yes, I will bump up the version.
- This PR is just some fundamental changes, and the more changes like upgrade logic will be added soon. Thanks for your remind, that's helpful.

minli1:
Hi, @lindig 
Asking your help to approve and merge this PR if no other concern.
Many thanks.

ocaml/xapi/xapi_xenops.ml
robhoes: Instead of constructing a string and then converting to the `address` type, why not constructing the `address` record directly?

ocaml/xapi/xapi_vgpu_type.ml
robhoes: Is it really needed to extend the identifier string? Is the existing one not good enough anymore as a primary key? It seems to me that the type ID is redundant information, and we might as well leave it out so that we don't need to worry about the upgrade/import cases?

ocaml/xapi/xapi_xenops.ml
```diff
@@ -647,29 +647,30 @@ module MD = struct
...
     let vgpu_type = vgpu.Db_actions.vGPU_type in
-    let internal_config =
-      Db.VGPU_type.get_internal_config ~__context ~self:vgpu_type in
-    let config_file =
-      try List.assoc Xapi_globs.vgpu_config_key internal_config
-      with Not_found -> failwith "NVIDIA vGPU config file not specified"
-    in
-    let config_file =
+    let identifier = Db.VGPU_type.get_identifier ~__context ~self:vgpu_type in
+    let type_id =
```
robhoes: The identifier is not meant to be parsed. Add the type_id to the `internal_config` field of `VGPU_type`, as was done for the conf file.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -647,29 +647,30 @@ module MD = struct
...
-        Printf.sprintf "%s,%s" config_file extra_args
-      with Not_found -> config_file
+        Scanf.sscanf identifier "%04d,nvidia,%04x,%04x,%04x,%04x,%s" (fun _ _ _ _ _ type_id -> type_id)
+      with _ ->
+        Scanf.sscanf identifier "%04d,nvidia,%04x,,%04x,%04x,%s" (fun _ _ _ _ type_id -> type_id)
     in
     let implementation =
       Nvidia {
         physical_pci_address = None; (* unused *)
-        config_file;
```
robhoes: Removing this field is likely to break the upgrade case.

ocaml/xapi/xapi_vgpu_type.ml
robhoes: Let's delete the old code (in a separate commit), as it really won't work anymore now that we need this `type_id`.

ocaml/xapi/xapi_xenops.ml
minli1: let get_virtual_pci_address ~__context vgpu =
    let open Pci in
    let device = vgpu.Db_actions.vGPU_device in
    {
      domain = 0000;
      bus = int_of_string device;
      dev = 0;
      fn = 0;
    }
Updated like above. Thanks, Rob,

ocaml/xapi/xapi_xenops.ml
minli1: Yes, I will remove from `identifier` to `internal_config.` Thanks.

ocaml/xapi/xapi_vgpu_type.ml
minli1: Yes, now I will make changes to internal_config and delete the type_id from the identifier.

ocaml/xapi/xapi_xenops.ml
minli1: Sure, I will add this back here and IDL.
And I am a little concern about during upgrade/migrate case, as we keep the old field and adding new fields, can you point where to handle the new fields? I am wondering this is a common way for xenopsd handle new things.

ocaml/xapi/xapi_vgpu_type.ml
minli1: OK, I will delete in a seperate commit.

minli1:
I had refined all the comments above.
And A new IDL PR to add `config_file` back is here: https://github.com/xapi-project/xcp-idl/pull/271

ocaml/xapi/xapi_xenops.ml
robhoes: Probably `bus = 0; dev = int_of_string device`?

minli1: :) Yes, the code pushed is right. I made a mistake here :)

ocaml/xapi/xapi_xenops.ml
robhoes: It may be possible to handle this in xenopsd, in the `import_metadata` function in xenops_server.ml. You need to do something special if `config_file = None`. The `virtual_pci_address` can just be set to `0000:00:0a.0` in that case, but I don't know where you'll get the `type_id` from....

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -256,187 +256,6 @@ module Passthrough = struct
       [find_or_create ~__context passthrough_gpu]
 end
 
-module Nvidia_old = struct
```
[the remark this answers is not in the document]
minli1: Yes, It's compile.
In `find_or_create_supported_types`, the old way was commented so it may not obvious. I will delete it now.

ocaml/xapi/xapi_xenops.ml
michael2012z: @robhoes , in upgrade scenario, can we keep the handling of type_id and config_file fields in XAPI in this way?
When the metadata is imported, we check "config_file", if it isn't none, in xapi we find the corresponding type_id, and allocate the virtual_pci_address (or write a hardcode "0000:00:0a.0"). By parsing the conf_file file name, we can find the matching type_id.
Then we can guarantee XAPI always send valid type_id and virtual_pci_address to Xenopsd.

minli1:
Hi, @robhoes 
Would you please approve and merge this if no other comments? For the import cases, we will have a test and trying to handle it in the coming PRs.
Thanks.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -647,29 +647,30 @@ module MD = struct
     let pci_address = Db.PCI.get_pci_id ~__context ~self:pci in
     Xenops_interface.Pci.address_of_string pci_address
 
+  let get_virtual_pci_address ~__context vgpu =
+    let open Pci in
+    let device = vgpu.Db_actions.vGPU_device |> int_of_string in
```
lindig: Why do we need a conversion string -> int -> string -> int? At a minimum, `device` should be a string that is converted once to an integer. A string that is a hexadecimal number can be converted by `int_of_string( "0x"^str)` to an int.

minli1: Since the device is string but decimal.
Then I have to at first convert it to a hex by Printf.sprintf. and then convert the hex string to int.
You are right that if the device is hex then the `int_of_string( "0x"^str)` is good.

minli1: Thanks. 
But how about the device is string but decimal, it's not hex.
Then I had to convert it to hex by Printf.sprintf and then convert to int.

lindig: What is the format of vgpu.Db_actions.vGPU_device? In either case, one conversion from string to int should be enough.

minli1: vgpu.Db_actions.vGPU_device is string of decimal.
like device = 10, then the pci_address, should be 0000:0:0a.0.
So device should be string"10" ->10(int) ->"0a" (string) -> 0a(int). because of the type address is defined:
{
  domain: int
  bus:int
  dev:int
  fn:int
}

lindig: The address record defines `dev` to be int, so a single `int_of_string` is enough. There is no such thing as `0a(int)`. Decimal and hexadecimal are string representations of an `int` value and you can convert the integer `dev` to whatever you need when you need a string.

minli1: Thanks, after checking it's just working with int_of_string without the prefix.

minli1:
Hi, @lindig Would you like to merge this PR? Thanks.

lindig:
@robhoes had a lot of comments so I would like to him to sign this off.

minli1:
Thanks. @lindig 
Hi, @robhoes  May I ask you to review this again and merge it as we are holding some PRs which depend on this one.

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -306,6 +306,7 @@ module Nvidia_old = struct
                   (List.assoc "plugin0.num_heads" args) in
               let max_instance = Int64.of_string
                   (List.assoc "plugin0.max_instance" args) in
+              (* Define "0" here because now we are using new way to read from nvidia conf file  *)
```
robhoes: Leftover comment?

minli1: Ah, will delete.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -741,44 +742,24 @@ module MD = struct
 
   let vgpus_of_vm ~__context (vmref, vm) =
     let open Vgpu in
-    if Vgpuops.vgpu_manual_setup_of_vm vm
```
robhoes: The function `Vgpuops.vgpu_manual_setup_of_vm` is now unused, so please delete the definition.

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -688,7 +691,7 @@ module Vendor_nvidia = struct
       max_resolution_x = conf.max_x;
       max_resolution_y = conf.max_y;
       size = Int64.div Constants.pgpu_default_size conf.max_instance;
-      internal_config = [Xapi_globs.vgpu_config_key, conf.file_path];
+      internal_config = [Xapi_globs.vgpu_config_key, conf.file_path; "type_id", conf.type_id];
```
robhoes: We don't need the config file path anymore, so remove it from here, and remove the key from xapi_globs. But add a value to xapi_globs for the type_id.

minli1: OK, got it.

ocaml/xapi/xapi_xenops.ml
minli1: OK, will remove.

minli1:
Thanks, @robhoes  All comments ⬆️ refined.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -647,29 +647,30 @@ module MD = struct
...
-    in
-    let config_file =
-      try
-        let extra_args =
-          List.assoc Xapi_globs.vgpu_extra_args_key vm.API.vM_platform in
-        Printf.sprintf "%s,%s" config_file extra_args
-      with Not_found -> config_file
-    in
+    let type_id = Db.VGPU_type.get_internal_config ~__context ~self:vgpu_type 
+      |> List.assoc "type_id" in
```
robhoes: ;)

minli1: Reslolved.

robhoes:
I think that Travis isn't working on this branch. We need to make sure that it does.

minli1:
> I think that Travis isn't working on this branch. We need to make sure that it does.

Yes, because of IDL is ahead of this with UUID thing.  It will be fixed after this change get merged.
https://github.com/xapi-project/xen-api/pull/3849#issuecomment-487527388

PR #3847, opened 2019-04-18
CA-314381: fix race condition in secure boot startup

edwintorok:
```
VM1                                 VM2
Sys.file_exists ... -> false
create xapi_uefi_certificates.tar  Sys.file_exists ... -> false
                                   create xapi_uefi_certificates.tar
extract xapi_uefi_certificates.tar
remove xapi_uefi_certificates.tar
                                   extract xapi_uefi_certificates.tar:ENOENT
```

The race condition could be solved either with a lock, or with using
unique names for temporary files.
This patch uses unique names for the temporary file to avoid the race
condition.

### DIFF
--- a/.travis-xs-opam.sh
+++ b/.travis-xs-opam.sh
@@ -3,7 +3,7 @@
 set -ex
 
 PACKAGE="xapi"
-PINS="xapi:."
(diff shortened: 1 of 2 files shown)


lindig:
Is the branch out of date? Travis can't build this.

edwintorok:
I think we're not pinning enough things in Travis, it takes xapi from the right branch, but it takes the datamodel from master, we need to pin all sub-packages, I'll try to fix it.

PR #3848, opened 2019-04-23
CP-30294: Bumped the API minor version and the client min/max version to 2.13

kc284:


kc284:
By corresponding change we mean that XenCenter should have an SDK aware of the bumped version - which XenCenter has as of https://github.com/xenserver/xenadmin/commit/fd083584a9076b721feadd5e7486a361d72801dc

ocaml/xapi/xapi_globs.ml
```diff
@@ -52,8 +52,8 @@ let xapi_user_agent = "xapi/"^(string_of_int version_major)^"."^(string_of_int v
  *
  * Please consult the XenCenter maintainers before changing these numbers, because a corresponding change
  * will need to be made in XenCenter *)
-let xencenter_min_verstring = "2.12"
-let xencenter_max_verstring = "2.12"
+let xencenter_min_verstring = "2.13"
+let xencenter_max_verstring = "2.13"
```
lindig:
```

File: ocaml/xapi/xapi_globs.ml [hunk 2]
```diff
@@ -52,8 +52,8 @@ let xapi_user_agent = "xapi/"^(string_of_int version_major)^"."^(string_of_int v
  *
  * Please consult the XenCenter maintainers before changing these numbers, because a corresponding change
  * will need to be made in XenCenter *)
-let xencenter_min_verstring = "2.12"
-let xencenter_max_verstring = "2.12"
+let xencenter_min_verstring = "2.13"
```
lindig:
```

Comment (lindig) [inline ocaml/xapi/xapi_globs.ml]: I think we should use something line the code below:
```
let xencenter_min_verstring = Printf.sprintf "%Ld.%Ld"
  Datamodel.api_version_major Datamodel.api_version_minor
```

kc284: The reason why this is hardcoded is that in theory one might want to allow different versions of XenCenter to connect to the server; for example, if we wanted naples XenCenter to connect to a plymouth server. This did happen some time in the remote past. Nowadays I'm not aware of a case when these numbers differ from the API version - in which case we could get rid of xencenter_min_verstring, xencenter_max_verstring completely - however, I can't promise it won't happen again. I suppose we can change it to use the variable and if we have another unusual use case hardcode it only on that branch.

lindig: I understand that we want no to use `Datamodel.api_version_major` directly for the exact reason you are saying. But in the common case they are identical and we could derive one from the other - which is what I am suggesting.

PR #3850, opened 2019-04-24
CP-31124: Add vGPU uuid as parameter to get metadata.

michael2012z:
vGPU uuid is needed in gpumon in the case of multiple vGPUs.

### DIFF
--- a/ocaml/xapi/xapi_gpumon.ml
+++ b/ocaml/xapi/xapi_gpumon.ml
@@ -77,16 +77,17 @@ module Nvidia = struct
       let dbg   = Context.string_of_task __context in
       let vm    = Db.VGPU.get_VM ~__context ~self:vgpu in
       let domid = Db.VM.get_domid ~__context ~self:vm |> Int64.to_int in
+      let vgpu_uuid = Db.VGPU.get_uuid ~__context ~self:vgpu in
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_gpumon.ml
```diff
@@ -77,10 +77,11 @@ module Nvidia = struct
...
       let domid = Db.VM.get_domid ~__context ~self:vm |> Int64.to_int in
+      let vgpu_uuid = Db.VGPU.get_uuid ~__context ~self:vgpu in
       Db.VGPU.get_resident_on ~__context ~self:vgpu
       |> (fun self -> Db.PGPU.get_PCI ~__context ~self)
       |> (fun self -> Db.PCI.get_pci_id ~__context ~self)
-      |> Gpumon_client.Client.Nvidia.get_vgpu_metadata dbg domid
+      |> (fun pgpu_address -> Gpumon_client.Client.Nvidia.get_vgpu_metadata dbg domid pgpu_address vgpu_uuid)
       |> (function
           | []      -> []
           | [meta]  -> [key, Stdext.Base64.encode meta]
```
lindig: I think the error message below should report the UUID. What happens is that we found more than one vGPU for a given domid and uuid.

michael2012z: Updated.

PR #3851, opened 2019-04-24
CA-315107 Create xapi-init-complete systemd target

TimSmithCtx:
Create a xapi-init-complete.target and a xapi-wait-init-complete.service
which is wanted by it. The service invokes the existing
xapi-wait-init-complete so that it is only done in one place now.

The SPEC file will need altering to enable the xapi-wait-init-complete
service on installation.

xapi-domains.service (and any other service which needs to start after
Xapi has completed initialisation) can now declare

Wants=xapi-init-complete.target
After=xapi-init-complete.target

to ensure that Xapi will be done initialising before they are started.

### DIFF
--- a/scripts/xapi-domains.service
+++ b/scripts/xapi-domains.service
@@ -1,7 +1,8 @@
 [Unit]
 Description=Start/stop domains on dom0 start/shutdown
 Requires=proc-xen.mount
-After=remote-fs.target xapi.service
+Wants=xapi-init-complete.target
+After=remote-fs.target xapi-init-complete.target
 
 [Service]
 Type=oneshot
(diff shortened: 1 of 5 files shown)


scripts/xapi-domains.service
```diff
@@ -1,7 +1,8 @@
 [Unit]
 Description=Start/stop domains on dom0 start/shutdown
 Requires=proc-xen.mount
-After=remote-fs.target xapi.service
+Wants=xapi-init-complete.target
+After=remote-fs.target xapi-init-complete.target
```
robhoes: Will this still do the right thing on shutdown, i.e. stop xapi-domains before xapi?

TimSmithCtx: It doesn't change the shutdown behaviour. Since xapi-domains is started after xapi, it will be shut down before.

PR #3854, opened 2019-04-29
Changed the checksum algorithm from SHA1 to xxHash

RoryBG:
Backwards compatability is maintained and significant performance improvements seen. 

Signed-off-by: RoryBG <rory.bertuzzi-glover@citrix.com>

### DIFF
--- a/.travis-xs-opam.sh
+++ b/.travis-xs-opam.sh
@@ -5,7 +5,8 @@ set -ex
 PACKAGE="xapi"
 PINS="xapi:."
 BASE_REMOTE="https://github.com/xapi-project/xs-opam.git"
-DISTRO="debian-9-ocaml-4.07"
+DISTRO="debian-unstable"
+OCAML_VERSION="4.07"
 
 wget https://raw.githubusercontent.com/ocaml/ocaml-ci-scripts/master/.travis-docker.sh
 . .travis-docker.sh
--- a/ocaml/xapi/dune
+++ b/ocaml/xapi/dune
@@ -57,6 +57,7 @@ let () = Printf.ksprintf Jbuild_plugin.V1.send {|
    mtime
    mtime.clock.os
    sha
+   xxhash
    tar
    tar-unix
    xapi-tapctl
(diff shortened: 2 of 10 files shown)


edwintorok:
What is the status of the automated performance tests in RAGE on this branch? Are they showing an improvement?

RoryBG:
> 
> 
> What is the status of the automated performance tests in RAGE on this branch? Are they showing an improvement?

The RAGE graphs show an improvement, however because the automated tests are with 24GB Windows VMs, the performance improvement is quite small. I ran some manual tests with much larger VMs and they saw significant performance improvements, for example:
```
Machine: perfuk-01-03 (440 GB VM) VM-Export
Old Xapi: 36 mins @ 205 MB/s
New Xapi: 27mins @ 270 MB/s
```
As you'd expect, the larger the VM, the larger the performance improvement.

edwintorok:
Could we add a RAGE test with the larger VM size?

lindig:
Before we merge this, we should update xs-opam to use debian-unstable and to add the opam file from this PR. Otherwise xs-opam will fail to build after the merge.

RoryBG:
> 
> 
> Could we add a RAGE test with the larger VM size?

We could do, the problem is that currently the test just installs a Windows VM and then tries to export it. 

So we'd have to do change it a Linux VM and XenRT probably has some functionality to call `dd` within a Linux guest to generate a random file of a given size.

edwintorok:
You could attach a 2nd disk to the VM, and fill it from within Dom0 if it is easier,  for example:
```
echo 'dd if=/dev/urandom of=/dev/$DEVICE bs=1M || true' >fill.sh

DISK_SIZE=100GiB
VM_UUID=...
POOL_UUID=$(xe pool-list --minimal)
SR_UUID=$(xe pool-param-get param-name=default-SR uuid=${POOL_UUID} --minimal)
VDI_UUID=$(xe vdi-create sr-uuid=${SR_UUID} name-label='export-test' virtual-size=${DISK_SIZE})
VBD_UUID=$(xe vbd-create vm-uuid=${VM_UUID} vdi-uuid=${VDI_UUID} device=2)
/opt/xensource/debug/with-vdi ${VDI_UUID} sh ./fill.sh
```

lindig:
PR to for xs-opam: https://github.com/xapi-project/xs-opam/pull/383

lindig:
We merged the xs-opam commit and I'd be ready to merge this. Any objections?

edwintorok:
Not an objection per se, but would be good to wait until we have the updated automated perf tests, considering that the reason to have this feature was performance. That way we'll know that we don't regress this in the future.

PR #3855, opened 2019-04-30
CP-30660: Dry run the allocate vGPU to pGPU

liulinC:
This commit two functionalities:
1. Dry run allocation vGPU to pGPUs
2. Support assign multiple vGPUs to VM

Test done: manual dev test

### DIFF
--- a/ocaml/xapi/vgpuops.ml
+++ b/ocaml/xapi/vgpuops.ml
@@ -18,7 +18,7 @@ open Stdext
 open Listext
 open Xstringext
 
-type vgpu = {
+type vgpu_t = {
   vgpu_ref: API.ref_VGPU;
   gpu_group_ref: API.ref_GPU_group;
   devid: int;
@@ -27,7 +27,7 @@ type vgpu = {
   requires_passthrough: [ `PF | `VF ] option;
 }
 
-let vgpu_of_vgpu ~__context vm_r vgpu =
+let vgpu_of_ref ~__context vgpu =
   let vgpu_r = Db.VGPU.get_record ~__context ~self:vgpu in
   {
     vgpu_ref = vgpu;
@@ -39,7 +39,7 @@ let vgpu_of_vgpu ~__context vm_r vgpu =
   }
 
 let vgpus_of_vm ~__context vm_r =
-  List.map (vgpu_of_vgpu ~__context vm_r) vm_r.API.vM_VGPUs
+  List.map (vgpu_of_ref ~__context ) vm_r.API.vM_VGPUs
 
 let fail_creation vm vgpu =
   match vgpu.requires_passthrough with
@@ -55,10 +55,48 @@ let fail_creation vm vgpu =
         Ref.string_of vgpu.gpu_group_ref
       ]))
 
-let allocate_vgpu_to_gpu ~__context vm host vgpu =
+let allocate_vgpu_to_gpu ?(dry_run=false) ?(pre_allocate_list=[]) ~__context vm host vgpu =
+  (* Get all pGPU from the host *)
   let available_pgpus = Db.Host.get_PGPUs ~__context ~self:host in
+  (* Get all pGPU from the required groups *)
   let compatible_pgpus = Db.GPU_group.get_PGPUs ~__context ~self:vgpu.gpu_group_ref in
   let pgpus = List.intersect compatible_pgpus available_pgpus in
+
+  let pgpu_can_hold_vgpu pgpu vgpu =
+    try Xapi_pgpu.assert_can_run_VGPU ~__context ~self:pgpu ~vgpu;true
+    with e -> false in
+  (* Get all pGPU that can hold the vGPU *)
+  let active_pgpus = List.filter ( fun pgpu -> pgpu_can_hold_vgpu pgpu vgpu.vgpu_ref ) pgpus in
+
+  let remaining_capacity_for_vgpu_from_pgpu vgpu pgpu =
+    let db_remaining =  Helpers.call_api_functions ~__context
+        (fun rpc session_id ->
+           Client.Client.PGPU.get_remaining_capacity ~rpc ~session_id
+             ~self:pgpu ~vgpu_type:vgpu.type_ref) in
+
+    let size_of_vgpu vgpu_ref =
+        let vgpu_type = Db.VGPU.get_type ~__context ~self:vgpu_ref in
+        Int64.to_float (Db.VGPU_type.get_size ~__context ~self:vgpu_type) in
+
+    let current_vgpu_size = size_of_vgpu vgpu.vgpu_ref in
+
+    let size_over_current_vGPU other =
+      let other_size = size_of_vgpu other in
+      other_size /. current_vgpu_size in
+
+    (* Check if any pre_allocation existed, the pre_allocation is a set of vGPU allocation that
+     * not reflected in the database, usually in the dry run mode, with following format
+     * [(v1,p1);(v2,p2);(v3,p1)...]*)
+    let virtual_allocation = List.fold_left (fun num ele ->
+        match ele with
+        |(cvgpu,cpgpu) when cpgpu = pgpu -> num +. (size_over_current_vGPU cvgpu)
+        |(_,_) -> num )
+        0. pre_allocate_list in
+    let virtual_available_allocation = Int64.of_float (ceil virtual_allocation) in
+    let result = Int64.sub db_remaining virtual_available_allocation in
+    if result < 0L then raise Api_errors.(Server_error (internal_error, ["The remaining cappacity cannot be negative"]))
+    else result in
+
   (* Sort the pgpus in lists of equal optimality for vGPU placement based on
    * the GPU groups allocation algorithm *)
   let sort_desc =
@@ -67,11 +105,8 @@ let allocate_vgpu_to_gpu ~__context vm host vgpu =
     | `breadth_first -> true
   in
   let sorted_pgpus = Helpers.sort_by_schwarzian ~descending:sort_desc
-      (fun pgpu ->
-         Helpers.call_api_functions ~__context (fun rpc session_id ->
-             Client.Client.PGPU.get_remaining_capacity ~rpc ~session_id
-               ~self:pgpu ~vgpu_type:vgpu.type_ref))
-      pgpus
+      (fun pgpu -> remaining_capacity_for_vgpu_from_pgpu vgpu pgpu )
+      active_pgpus
   in
   let rec choose_pgpu = function
     | [] -> None
@@ -85,9 +120,9 @@ let allocate_vgpu_to_gpu ~__context vm host vgpu =
   match choose_pgpu sorted_pgpus with
   | None -> fail_creation vm vgpu
   | Some pgpu ->
-    Db.VGPU.set_scheduled_to_be_resident_on ~__context
-      ~self:vgpu.vgpu_ref ~value:pgpu;
-    pgpu
+    if not dry_run then
+      Db.VGPU.set_scheduled_to_be_resident_on ~__context ~self:vgpu.vgpu_ref ~value:pgpu;
+    (vgpu.vgpu_ref,pgpu)::pre_allocate_list
 
 (* Take a PCI device and assign it, and any dependent devices, to the VM *)
 let add_pcis_to_vm ~__context host vm pci =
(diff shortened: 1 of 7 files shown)


ocaml/xapi/xapi_pgpu_helpers.ml
```diff
@@ -50,20 +50,30 @@ let get_allocated_VGPUs ~__context ~self =
...
-     in
-     if running_type <> vgpu_type
-     then raise (Api_errors.Server_error (
-         Api_errors.vgpu_type_not_compatible_with_running_type,
-         [
-           Ref.string_of self;
-           Ref.string_of vgpu_type;
-           Ref.string_of running_type;
-         ])))
+  let allocated_vgpu_list = get_allocated_VGPUs ~__context ~self in
```
lindig: I think this code is complicated and hard to understand. Would it be simpler to use sets rather than lists?

ocaml/xapi/vgpuops.ml
```diff
@@ -55,10 +55,46 @@ let fail_creation vm vgpu =
...
+      let self_size = size_of_vgpu self in
+      let other_size = size_of_vgpu other in
+      (Int64.to_float self_size) /. (Int64.to_float other_size) in
+
+    (* Check if any pre_allocation existed, the pre_allocation is a set of vGPU allocation that
+     * not reflected in the database, usually in the dry run mode, with following format
+     * [(v1,p1);(v2,p2);(v3,p1)...]*)
+    let virtual_allocation = List.fold_left (fun num ele ->
+        match ele with
+        |(cvgpu,cpgpu) when cpgpu = pgpu -> num +. (vGPU_size_of cvgpu vgpu.vgpu_ref)
```
lindig: `vgpu.vgpu_ref` is constant over the list but its size is computed over and over again. I suspect `other` in `vGPU_size_of` is constant and could be computed once.

liulinC: good point, refined for `vgpu.vgpu_ref`,
I think `other` need to be queried from the database everytime, unless we `cache` the result in memory,
Does that complicated?

ocaml/xapi/xapi_pgpu_helpers.ml
liulinC: code refined, hopfully to make it a little simper, 
The purpose of this code is to make sure the new vGPU type is compatible with all the residenting/scheduled vGPU types, 
Change `List` to `Set` maybe not helpful to simplify the code.

liulinC:
@lindig @robhoes May I know is there any further concern about this PR?

ocaml/xapi/xapi_pgpu_helpers.ml
```diff
@@ -50,20 +50,33 @@ let get_allocated_VGPUs ~__context ~self =
...
+  (** Now check whether the requested type is permitted *)
+  match allocated_vgpu_list with
+  | [] -> () (* Not allocated on this pgpu, does not need to check compatibility*)
+  | hd::tail ->
+    let get_compatible_types_on_pgpu vgpu=
+      let vgpu_type = Db.VGPU.get_type ~__context ~self:vgpu in
+      Db.VGPU_type.get_compatible_types_on_pgpu ~__context ~self:vgpu_type in
+    (* The new vGPU type must in the compatible_types_on_pgpu of all the allocated vGPUs*)
+    let grant_vgpu_type_list = List.fold_left
```
lindig: This code is paying a price for using lists instead of sets. Why do we need to `sort_uniq` before we can pass the result to `List.intersect`?

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -358,63 +358,30 @@ let has_non_allocated_vgpus ~__context ~self =
...
-  let avail_groups =
-    List.map
-      (fun self -> Db.VGPU.get_GPU_group ~__context ~self)
-      runnable_vgpus
-  in
-  let not_available = List.set_difference reqd_groups avail_groups in
+  let open Vgpuops in
+  let vgpurs = List.map (Vgpuops.vgpu_of_ref ~__context)  vgpus in
+  try
+    ignore (List.fold_left (fun pre_allocate_list vgpu -> Vgpuops.allocate_vgpu_to_gpu ~dry_run:true ~pre_allocate_list ~__context self host vgpu) [] vgpurs)
```
lindig: Is `List.fold_left` the right choice when we ignore the result? Could `List.iter` be used?

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -358,63 +358,30 @@ let has_non_allocated_vgpus ~__context ~self =
...
-      let usb_host = Db.PUSB.get_host ~__context ~self:pusb in
-      assert (usb_host = host)
-    with _ -> raise (Api_errors.Server_error (Api_errors.operation_not_allowed,
-      [Printf.sprintf "VUSB %s is not available on Host %s"
-        (Ref.string_of vusb)
-        (Ref.string_of host)
-      ]))
+      try
+        let usb_group = Db.VUSB.get_USB_group ~__context ~self:vusb in
+        let pusb = List.hd (Db.USB_group.get_PUSBs ~__context ~self:usb_group) in
```
lindig: `List.hd` can fail. It's better to use pattern matching unless you convince me that this can't fail.

ocaml/xapi/xapi_pgpu_helpers.ml
```diff
@@ -50,20 +50,33 @@ let get_allocated_VGPUs ~__context ~self =
...
-     then raise (Api_errors.Server_error (
-         Api_errors.vgpu_type_not_compatible_with_running_type,
-         [
-           Ref.string_of self;
-           Ref.string_of vgpu_type;
-           Ref.string_of running_type;
-         ])))
+  let allocated_vgpu_list = get_allocated_VGPUs ~__context ~self in
+  (** Now check whether the requested type is permitted *)
+  match allocated_vgpu_list with
```
lindig: This is more complicated than needed. Just call `List.for_all compatible_with_new_vGPU_type allocated_vgpu_list` and it would return `true` for the empty list (and never call the predicate).

ocaml/xapi/xapi_vm_helpers.ml
lindig: Any comment?

liulinC:
thanks @lindig , code refined basing on your comments, 
May I ask your approve if no other further concern? 
thanks in advance.

robhoes:
General comments:
* The commit messages are not very clear even though there is a lot happening in the commits. The way that capacity checking is done is changing considerably, and it would be good to explain how in the commit message.
* There seems to be a lot of (unrelated) whitespace being changed. Next time please do that in a separate commit (or not at all).

liulinC:
@robhoes , Thanks for your comments. 
- the commit message has been changed accordingly
- well noted about the indents. Will only indent the changed code

ocaml/xapi/vgpuops.ml
```diff
@@ -55,10 +55,48 @@ let fail_creation vm vgpu =
...
+        let vgpu_type = Db.VGPU.get_type ~__context ~self:vgpu_ref in
+        Int64.to_float (Db.VGPU_type.get_size ~__context ~self:vgpu_type) in
+
+    let current_vgpu_size = size_of_vgpu vgpu.vgpu_ref in
+
+    let size_over_current_vGPU other =
+      let other_size = size_of_vgpu other in
+      other_size /. current_vgpu_size in
+
+    (* Check if any pre_allocation existed, the pre_allocation is a set of vGPU allocation that
```
robhoes: Rather than doing all this pre-allocation checking here, isn't it better to make `Xapi_pgpu.assert_can_run_VGPU` (and functions called from it) aware of the pre-allocations? E.g. it's already taking into account resident and scheduled vGPUs through `Xapi_pgpu_helpers.get_allocated_VGPUs`. It seems unnecessary and confusing to again do computations about sizes and capacity right here.

ocaml/idl/datamodel_common.ml
```diff
@@ -325,6 +325,11 @@ let rio_release =
   ; internal_deprecated_since=None
   }
 
+let oslo_release =
```
robhoes: There won't be an Oslo release.

ocaml/idl/datamodel.ml
```diff
@@ -4545,6 +4545,13 @@ module PGPU = struct
           param_release = vgpu_tech_preview_release;
           param_default = None;
         };
+        {
```
robhoes: Let's not add this to the datamodel. As we only need this additional list for internal computations, it should not be part of the external API.

ocaml/xapi/xapi_pgpu.ml
```diff
@@ -305,14 +305,14 @@ let set_GPU_group ~__context ~self ~value =
                   [gpu_type; "[" ^ String.concat ", " group_types ^ "]"]))
     )
 
-let get_remaining_capacity ~__context ~self ~vgpu_type =
-  match Xapi_pgpu_helpers.get_remaining_capacity_internal ~__context ~self ~vgpu_type with
+let get_remaining_capacity ~__context ~self ~vgpu_type ~pre_allocate_list =
```
robhoes: This is the entry point for the API, so adding the parameter here made it necessary to change the datamodel. Don't do that, and leave this function unchanged. Anything internal that calls `Xapi_pgpu.get_remaining_capacity` can just call `Xapi_pgpu_helpers.get_remaining_capacity`, which does the exact same thing and can receive the new argument. That's what "helpers" and "internal" functions are for.

ocaml/xapi/xapi_pgpu_helpers.mli
```diff
@@ -43,20 +43,26 @@ val assert_no_resident_VGPUs_of_type :
...
+  (exn, int64) Stdext.Either.t
 
 (* Return the number of VGPUs of the specified type for which capacity
  * remains on the PGPU. *)
 val get_remaining_capacity :
   __context:Context.t ->
   self:API.ref_PGPU ->
-  vgpu_type:API.ref_VGPU_type -> int64
+  vgpu_type:API.ref_VGPU_type ->
+  pre_allocate_list:(API.ref_VGPU * API.ref_PGPU) list ->
```
robhoes: Make this an optional argument, so you don't need all the `~pre_allocate_list:[]`.

ocaml/xapi/xapi_pgpu_helpers.ml
```diff
@@ -109,14 +109,17 @@ let get_remaining_capacity_internal ~__context ~self ~vgpu_type =
        * VM, the remaining capacity is binary. We simply return 0 if the PCI
        * device is currently passed-through to a VM, or is scheduled to be,
        * and 1 otherwise. *)
+      let inuse = List.exists (fun (_,pgpu) -> pgpu = self ) pre_allocate_list in
```
robhoes: `pre_allocated` rather than `inuse`?

ocaml/xapi/xapi_pgpu_helpers.ml
```diff
@@ -109,14 +109,17 @@ let get_remaining_capacity_internal ~__context ~self ~vgpu_type =
...
       let pci = Db.PGPU.get_PCI ~__context ~self in
       let scheduled = List.length (get_scheduled_VGPUs ~__context ~self) > 0 in
       let attached = Db.PCI.get_attached_VMs ~__context ~self:pci <> [] in
-      convert_capacity (if scheduled || attached then 0L else 1L)
+      convert_capacity (if scheduled || attached || inuse then 0L else 1L)
     else begin
       (* For virtual VGPUs, we calculate the number of times the VGPU_type's
        * size fits into the PGPU's (size - utilisation). *)
       let pgpu_size = Db.PGPU.get_size ~__context ~self in
+      let pgpu_pre_allocatd_vGPUs = List.filter (fun (vgpu,pgpu)->pgpu=self) pre_allocate_list
```
robhoes: Typo `pgpu_pre_allocatd_vGPUs` -> `pgpu_pre_allocated_vGPUs`. Also please add some spacing around the operators `->` and `=` (and `@` below), and use `_` for unused parameters.

ocaml/xapi/xapi_pgpu_helpers.mli
```diff
@@ -43,20 +43,26 @@ val assert_no_resident_VGPUs_of_type :
...
+  int64
 
 (** Check that the PGPU has capacity to run the specified VGPU. *)
 val assert_capacity_exists_for_VGPU_type :
   __context:Context.t ->
   self:API.ref_PGPU ->
-  vgpu_type:API.ref_VGPU_type -> unit
+  vgpu_type:API.ref_VGPU_type ->
+  pre_allocate_list:(API.ref_VGPU * API.ref_PGPU) list ->
```
robhoes: Do you really need the new argument on this function? It always seems to be `[]`.

PR #3858, opened 2019-05-02
CA-316165: workaround - disable nondeterministic unit test

edwintorok:
It uses Thread.delay and assumes that events get processed within 0.4s.
That may not be enough if the CI is under high load.

Should be replaced with looking for events and a (larger) timeout
instead. Disable for now to restore build stability.


MarkSymsCtx:
Unit tests with threads are almost universally a fail unless the test can lock step the threads in a deterministic manner. The test certainly needs to be in control of what the testee views as time so as to make the test repeatable.

PR #3857, opened 2019-04-30
REQ-720: CP-31160: Support for multiple Nvidia VGPU creation

liulinC:
Min is on vacation, raise PR on behavior of him

PR #3861, opened 2019-05-05
Continue CP-31160: Support creation of multiple vGPU

liulinC:


liulinC:
This PR is the follow up of
https://github.com/xapi-project/xen-api/pull/3857

As I cannot push to min's github repo directly.

@minli1 @lindig @robhoes  Could you please help to review at your early convenience.
Thanks in advance.

liulinC:
@lindig @robhoes May I know is there any further concern about this PR?

ocaml/xapi/xapi_vgpu.ml
```diff
@@ -42,19 +62,27 @@ let create' ~__context  ~vM ~gPU_group ~device ~other_config ~_type ~powerstate_
...
   in
+  (* during multiple vgpus creation:
+     1. Underlying vgpu_type should support multiple
+     2. _type must be listed on the all vgpu_type's compatible lists*)
+  let existing = Db.VM.get_VGPUs ~__context ~self:vM in
+  if existing <> [] then begin
+    let types = List.map (fun vgpu -> Db.VGPU.get_type ~__context ~self:vgpu) existing in
+    let is_in_compatible_lists _type vgpu_type =
+      let compatible_lists = Db.VGPU_type.get_compatible_types_in_vm ~__context ~self:vgpu_type in
+      List.mem _type compatible_lists
```
lindig: It is a bit unusual to use `_type` (with a an underscore) for a parameter that is actually uses. In particular, when this is a named parameter. We can leave this for now.

minli1:
Thanks @lindig @liulinC 
Please @robhoes help to take a look. :)

ocaml/xapi/vgpuops.ml
```diff
@@ -129,9 +129,8 @@ let reserve_free_virtual_function ~__context vm pf =
 let add_vgpus_to_vm ~__context host vm vgpus =
   (* Only support a maximum of one virtual GPU per VM for now. *)
   (try Db.VM.remove_from_other_config ~__context ~self:vm ~key:Xapi_globs.vgpu_pci with _ -> ());
-  match vgpus with
-  | [] -> ()
-  | vgpu :: _ ->
+  List.iter (fun vgpu -> debug "Vgpus located on VM(%s) are: %s" (Ref.string_of vm) (Ref.string_of vgpu.vgpu_ref)) vgpus;
+  List.iter (fun vgpu ->
```
robhoes: Two `iter`s?

robhoes: What is this change doing here anyway? This was already in another PR?

ocaml/xapi/xapi_vgpu.ml
```diff
@@ -17,9 +17,31 @@ open D
...
+let get_valid_device ~__context ~device ~vM =
+  let d = int_of_string device in
+  if d >= min_valid_vgpu_device && d <= max_valid_vgpu_device then device
+  else if d = 0 then begin
+    let all = Db.VM.get_VGPUs ~__context ~self:vM in
+    let all_existing_devices = List.map (fun self -> Db.VGPU.get_device ~__context ~self |> int_of_string) all in
+    try
+      let valid = List.find (fun d -> not (List.mem d all_existing_devices)) all_valid_devices in
+      string_of_int valid
+    with Not_found -> raise (Api_errors.Server_error (Api_errors.invalid_device, [device]))
```
robhoes: This error condition means that we have run out of guest PCI slots, not that device 0 is invalid. So we should introduce a new API error for it.

ocaml/xapi/xapi_vgpu.ml
```diff
@@ -42,19 +62,27 @@ let create' ~__context  ~vM ~gPU_group ~device ~other_config ~_type ~powerstate_
...
+    in
+    if not (List.for_all (fun vgpu_type -> is_in_compatible_lists _type vgpu_type) types) then
+      raise (Api_errors.Server_error (Api_errors.vgpu_type_not_compatible, [Ref.string_of _type]))
+  end;
 
-  Stdext.Threadext.Mutex.execute m (fun () ->
-      (* Check to make sure the device is unique *)
-      let all = Db.VM.get_VGPUs ~__context ~self:vM in
-      let all_devices = List.map (fun self -> Db.VGPU.get_device ~__context ~self) all in
-      if List.mem device all_devices then
```
robhoes: We have lost this uniqueness check. The function `get_valid_device` does not do it if `device <> 0`.

ocaml/xapi/xapi_vgpu.ml
```diff
@@ -42,19 +62,27 @@ let create' ~__context  ~vM ~gPU_group ~device ~other_config ~_type ~powerstate_
...
-        raise (Api_errors.Server_error (Api_errors.device_already_exists, [device]));
 
-      Db.VGPU.create ~__context ~ref:vgpu ~uuid ~vM ~gPU_group ~device
+  Stdext.Threadext.Mutex.execute m (fun () ->
```
robhoes: The device checks need to happen under this mutex (that's what the mutex was for).

ocaml/xapi/xapi_vgpu.ml
```diff
@@ -42,19 +62,27 @@ let create' ~__context  ~vM ~gPU_group ~device ~other_config ~_type ~powerstate_
                     (Api_errors.invalid_value, ["type"; Ref.string_of _type]))
     end
   in
+  (* during multiple vgpus creation:
+     1. Underlying vgpu_type should support multiple
+     2. _type must be listed on the all vgpu_type's compatible lists*)
+  let existing = Db.VM.get_VGPUs ~__context ~self:vM in
```
robhoes: We are doing the same DB call in `get_valid_device` just above here. Let's do it just once.

ocaml/xapi/xapi_vgpu.ml
```diff
@@ -42,19 +62,27 @@ let create' ~__context  ~vM ~gPU_group ~device ~other_config ~_type ~powerstate_
...
+  if existing <> [] then begin
+    let types = List.map (fun vgpu -> Db.VGPU.get_type ~__context ~self:vgpu) existing in
+    let is_in_compatible_lists _type vgpu_type =
+      let compatible_lists = Db.VGPU_type.get_compatible_types_in_vm ~__context ~self:vgpu_type in
+      List.mem _type compatible_lists
+    in
+    if not (List.for_all (fun vgpu_type -> is_in_compatible_lists _type vgpu_type) types) then
```
robhoes: You could just write `if not (List.for_all (is_in_compatible_lists _type) types) then`.

ocaml/xapi/xapi_vgpu.ml
robhoes: I think that this was done because `type` is a keyword. In fact, the datamodel code generator may have added this prefix.

ocaml/xapi/xapi_vgpu.ml
```diff
@@ -42,19 +62,27 @@ let create' ~__context  ~vM ~gPU_group ~device ~other_config ~_type ~powerstate_
                     (Api_errors.invalid_value, ["type"; Ref.string_of _type]))
     end
   in
+  (* during multiple vgpus creation:
+     1. Underlying vgpu_type should support multiple
+     2. _type must be listed on the all vgpu_type's compatible lists*)
+  let existing = Db.VM.get_VGPUs ~__context ~self:vM in
+  if existing <> [] then begin
```
robhoes: You don't really need this if-statement.

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -182,6 +188,26 @@ let find_and_update ~__context vgpu_type =
...
+(* since vgpu_type refs had been created in update, just need to get ref from vgpu_type.compatible* lists *)
+let update_compatible_lists ~__context vgpu_type vgpu_type_ref rc =
+  let get_type_from_model_name model =
+    let open Db_filter_types in
+    let expr = Eq(Field "model_name", Literal model) in
+    match Db.VGPU_type.get_internal_records_where ~__context ~expr with
+    | (_ref, _) :: _ -> _ref
+    | _ -> raise (Api_errors.Server_error(Api_errors.internal_error, ["Could not find vgpu_type according to model name" ]))
+  in
+  let compatible_types_in_vm =  List.map (fun model -> get_type_from_model_name model) vgpu_type.compatible_types_in_vm in
```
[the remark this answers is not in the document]
liulinC: The code 
```
let compatible_types_in_vm = if vgpu_type.compatible_types_in_vm = [] then [] else [vgpu_type] in
```
should work well in current situation, namely a VM only have one type of vGPU allowed, 
but in the future, the VM may have multiple types, then the list should not only contain itselfl, 
The value should be parsed from `vgpu_type.compatible_types_in_vm`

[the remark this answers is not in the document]
liulinC: Actually, Even for `create` function, the list should also be parsed from the sring list `vgpu_type.compatible_types_in_vm`,

But I think the problem here is, during the execution of `create` for one vGPU type, it may refer to other type of vGPU in its `compatible_types_in_vm`, which may have not been created at all, then We do not know its `Ref` value.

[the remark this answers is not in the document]
robhoes: Exactly, that is the problem in the create function. To solve that for the future, more general case, we need to have two rounds: first create all, then fill in the compatible lists. But we don't necessarily have to do that yet.

My main point is that we should do it the same for both the create and the update case (consistency).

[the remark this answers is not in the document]
liulinC: sure, if we do not have to do the two round ATM, it looks like the `create`  methon wins as it simpler, 
I will update the `udpate` function according.

ocaml/xapi/vgpuops.ml
liulinC: yes,   you are right, 
This is involved as Min and I developed parallelly, 
I will delete it here and merge in my PR.

ocaml/xapi/xapi_vgpu.ml
liulinC: will fix it.

ocaml/xapi/xapi_vgpu.ml
liulinC: fixed in new commit

ocaml/xapi/xapi_vgpu.ml
liulinC: fixed in new commit

ocaml/xapi/xapi_vgpu.ml
liulinC: fixed in new commit

ocaml/xapi/xapi_vgpu.ml
liulinC: fixed in new commit

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -129,9 +129,15 @@ let create ~__context ~vendor_name ~model_name ~framebuffer_size ~max_heads
   Db.VGPU_type.create ~__context ~ref ~uuid ~vendor_name ~model_name
     ~framebuffer_size ~max_heads ~max_resolution_x ~max_resolution_y
     ~size ~internal_config ~implementation ~identifier ~experimental
-    ~compatible_types_in_vm ~compatible_types_on_pgpu;
+    ~compatible_types_in_vm:[] ~compatible_types_on_pgpu:[];
   debug "VGPU_type ref='%s' created (vendor_name = '%s'; model_name = '%s')"
     (Ref.string_of ref) vendor_name model_name;
+  if compatible_types_in_vm <> [] then
```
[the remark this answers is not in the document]
liulinC: fixed in new commit

ocaml/xapi/xapi_vgpu.ml
liulinC: fixed in new commit

ocaml/xapi/xapi_vgpu.ml
```diff
@@ -17,9 +17,31 @@ open D
...
+    if low > high then [] else low :: aux (low+1) high in
+  aux low high
+
+let all_valid_devices = range min_valid_vgpu_device max_valid_vgpu_device
+
+let get_valid_device ~__context ~device ~vM ~vGPUs =
+  let d = int_of_string device in
+  let all_existing_devices = List.map (fun self -> Db.VGPU.get_device ~__context ~self |> int_of_string) vGPUs in
+  let device_in_use device = List.mem device all_existing_devices in
+  if d >= min_valid_vgpu_device && d <= max_valid_vgpu_device && not (device_in_use d)  then device
```
robhoes: We should raise `device_already_exists`, as before, if `device_in_use d`. I'll push a fix.

ocaml/xapi-consts/api_errors.ml
```diff
@@ -316,8 +316,10 @@ let pgpu_insufficient_capacity_for_vgpu = "PGPU_INSUFFICIENT_CAPACITY_FOR_VGPU"
 let vgpu_type_not_enabled = "VGPU_TYPE_NOT_ENABLED"
 let vgpu_type_not_supported = "VGPU_TYPE_NOT_SUPPORTED"
 let vgpu_type_not_compatible_with_running_type = "VGPU_TYPE_NOT_COMPATIBLE_WITH_RUNNING_TYPE"
+let vgpu_type_not_compatible = "VGPU_TYPE_NOT_COMPATIBLE"
 let vgpu_destination_incompatible = "VGPU_DESTINATION_INCOMPATIBLE"
 let nvidia_tools_error = "NVIDIA_TOOLS_ERROR"
+let vgpu_vm_ran_out_of_pci_slot = "VGPU_VM_RUN_OUT_OF_PCI_SLOT"
```
robhoes: I'm changing this to `vm_pci_bus_full` so that we can use the same error for other PCI devices as well.

robhoes:
I've updated and pushed this as https://github.com/xapi-project/xen-api/commit/93a3762a6ec0f3baba3fe94c89ccc3093f356289.

PR #3862, opened 2019-05-06
Fix test failures and make new test cases

michael2012z:
What's in the PR:
1.  Fixed test failures caused by data model change;
2.  Added test cases on metadata with multiple vGPUs.
3.  Removed platform/vgpu_extra_args test cases for vGPU. 
4.  Removed test cases around Nvidia_old.

### DIFF
--- a/ocaml/tests/test_xenopsd_metadata.ml
+++ b/ocaml/tests/test_xenopsd_metadata.ml
@@ -200,7 +200,7 @@ module VideoRam = Generic.Make(Generic.EncapsulateState(struct
                                  ]
                                end))
 
-let test_uuid = "uuiduuid-uuid-uuid-uuid-uuiduuiduuid"
+let test_uuid_base = "uuiduuid-uuid-uuid-uuid-uuiduuid"
 
 module GenerateVGPUMetadata = Generic.Make(Generic.EncapsulateState(struct
                                              open Test_vgpu_common
@@ -230,7 +230,7 @@ module GenerateVGPUMetadata = Generic.Make(Generic.EncapsulateState(struct
                                                     in
                                                     let (_ : API.ref_VGPU) =
                                                       make_vgpu ~__context
-                                                        ~vm_ref ~scheduled_to_be_resident_on:pgpu_ref ~uuid:test_uuid vgpu_type in
+                                                        ~vm_ref ~scheduled_to_be_resident_on:pgpu_ref ~uuid:(test_uuid_base^"0000") vgpu_type in
                                                     ())
                                                  pgpus_and_vgpu_types
 
@@ -264,7 +264,7 @@ module GenerateVGPUMetadata = Generic.Make(Generic.EncapsulateState(struct
                                                      virtual_pci_address = Some {domain=0; bus = 0; dev = 0; fn = 0;};
                                                      config_file = None;
                                                      type_id = "type_id_1";
-                                                     uuid = test_uuid;
+                                                     uuid = test_uuid_base^"0000";
                                                    })
                                                ];
                                                (* One Intel vGPU. *)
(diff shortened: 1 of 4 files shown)


lindig:
The Travis failures are caused by problems downloading dependencies from upstream.

ocaml/tests/test_xenopsd_metadata.ml
```diff
@@ -226,18 +194,14 @@ module VideoRam = Generic.Make(Generic.EncapsulateState(struct
...
-                                   {oc = []; platform=vgpu_platform_data @ ["videoram", "8"]}, 16;
-                                   (* videoram values larger than the default should be allowed for vGPU. *)
-                                   {oc = []; platform=vgpu_platform_data @ ["videoram", "32"]}, 32;
                                    (* Other VGA options shouldn't affect the videoram setting. *)
                                    {oc = []; platform=["vga", "cirrus"]}, 4;
                                    {oc = []; platform=["vga", "cirrus"; "videoram", "8"]}, 8;
                                  ]
                                end))
 
+let test_uuid_base = "uuiduuid-uuid-uuid-uuid-uuiduuid"
```
lindig: This is obviously not a uuid. It might not matter but I would prefer something that has at least the correct syntactical structure while still not being random.

```
let uuid_with_suffix suffix =
 Printf.sprintf "11111111-1111-1111-1111-11111111%s" suffix
```

ocaml/tests/test_xenopsd_metadata.ml
```diff
@@ -226,18 +194,14 @@ module VideoRam = Generic.Make(Generic.EncapsulateState(struct
...
+
```
lindig: The code below is indented too far which makes reviewing it harder than necessary.

ocaml/tests/test_xenopsd_metadata.ml
```diff
@@ -266,7 +230,7 @@ module GenerateVGPUMetadata = Generic.Make(Generic.EncapsulateState(struct
                                                     in
                                                     let (_ : API.ref_VGPU) =
                                                       make_vgpu ~__context
-                                                        ~vm_ref ~scheduled_to_be_resident_on:pgpu_ref vgpu_type in
+                                                        ~vm_ref ~scheduled_to_be_resident_on:pgpu_ref ~uuid:(test_uuid_base^"0000") vgpu_type in
```
lindig: Use a function to create these uuids.

ocaml/tests/test_xenopsd_metadata.ml
```diff
@@ -297,7 +261,10 @@ module GenerateVGPUMetadata = Generic.Make(Generic.EncapsulateState(struct
                                                [
                                                  Xenops_interface.Vgpu.(Nvidia {
                                                      physical_pci_address = None;
-                                                     config_file = "/usr/share/nvidia/vgx/grid_k100.conf";
+                                                     virtual_pci_address = Some {domain=0; bus = 0; dev = 0; fn = 0;};
+                                                     config_file = None;
+                                                     type_id = "type_id_1";
+                                                     uuid = test_uuid_base^"0000";
```
lindig: Same as above: use a function.

ocaml/tests/test_xenopsd_metadata.ml
michael2012z: Agree. But I prefix zero's.

ocaml/tests/test_xenopsd_metadata.ml
michael2012z: Indeed, it was. This line "module GenerateVGPUMetadata = Generic.Make(Generic.EncapsulateState(struct" make the auto-indenting go far in the editor. I am going to change the indent of the whole file for easy read.

michael2012z:
@lindig, I improved the code per your comments. 
And I shortened the code indentation of some functions. Now it will be a little bit difficult to see the difference.

PR #3863, opened 2019-05-08
CA-316241 redirect stderr output from probe-device-for-file

lindig:
probe-device-for-file is called from xe-restore-metadata which is called
from xsconsole. Output to stderr, stdin is shown to the user and it
contains irrelevant and confusing output from probe-device-for-file
which is coming from the fsimage library. Since we can't change the
fsimage library, stderr for the call to fsimage (only) is redirected to
/dev/null. This does not make us miss errors as these are raised as
exceptions.

### DIFF
--- a/scripts/probe-device-for-file
+++ b/scripts/probe-device-for-file
@@ -4,6 +4,37 @@
 
 import os, sys
 import fsimage
+from contextlib import contextmanager
+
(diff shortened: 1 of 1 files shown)


edwintorok:
Would be good to get this fixed in Xen's libfsimage eventually, for now we use a workaround.

PR #3864, opened 2019-05-08
CA-315688: Bumped API version to 2.14 for the plymouth release

kc284:
Also, do not autogenerate placeholders for unreleased API versions in the docs.


kc284:
I'm going to merge this so as to update the SDK for XenCenter.

PR #3867, opened 2019-05-14
Remove the last direct Xenctrl dependency

robhoes:


robhoes:
Travis failed because of this:
```
File "ocaml/xapi/xapi_xenops.ml", line 198, characters 15-49:
Error: This function has type string -> domid:int -> unit
    ocamlopt ocaml/xapi/.xapi_internal.objs/native/stream_vdi.{cmx,o}
```
This is because it has not yet got the xenopsd update from https://github.com/xapi-project/xenopsd/commit/f75a6477adf710c5b82a1004b7d7d0d4c33b3a0a.

robhoes:
@psafont I've added some more text in the commit message.

PR #3871, opened 2019-05-21
CP-31268: copy SMBIOS type2 info from host to VM

TalonsLee:
Backport CP-31084 change, only keep the bios string copy function.

### DIFF
--- a/ocaml/xapi/bios_strings.ml
+++ b/ocaml/xapi/bios_strings.ml
@@ -90,7 +90,10 @@ let get_host_bios_strings ~__context =
   info "Getting host BIOS strings.";
   (* named BIOS strings *)
   let dmidecode_strings = ["bios-vendor"; "bios-version"; "system-manufacturer";
-                           "system-product-name"; "system-version"; "system-serial-number"] in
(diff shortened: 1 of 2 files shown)


robhoes:
What's the target branch? It should be one of the *-lcm branches if it is for a hotfix.

lindig:
The original commit is 94d2ff73711becc9cc92d01606ee38929808eb14 and contains more changes. Why can we take only part of them?

TalonsLee:
> What's the target branch? It should be one of the *-lcm branches if it is for a hotfix.

release/havana/lcm

Should we merge all commits in this hotfix to this this upd-branch first?

TalonsLee:
> The original commit is [94d2ff7](https://github.com/xapi-project/xen-api/commit/94d2ff73711becc9cc92d01606ee38929808eb14) and contains more changes. Why can we take only part of them?

As discussed in xsi-278, do this with back-port dependent commit,
so there is no cmdline setting function, just bios-copy.

lindig:
Please add to the commit message the hash of the original commit and the reason why we can only take some parts.

robhoes:
For Havana hotfixes, just use 1.14-lcm.

PR #3879, opened 2019-05-29
Prepare for feature/guefi/master merge

edwintorok:
Depends on https://github.com/xapi-project/xenopsd/pull/627


ocaml/xapi/xapi_vm.ml
```diff
@@ -238,6 +238,39 @@ let assert_memory_constraints ~__context ~vm platformdata =
...
+    Xapi_stdext_pervasives.Pervasiveext.finally
+      (fun () -> output_string out contents)
+      (fun () -> close_out out);
+    Unixext.with_file filename [Unix.O_RDONLY] 0 f
+  ) (fun () -> Sys.remove filename)
+
+let save_uefi_certificates_to_dir ~__context ~pool ~vm =
+  let uefi_key = !Xapi_globs.varstore_dir ^ "KEK.auth" in
+  if not (Sys.file_exists uefi_key) then
+    if Sys.file_exists !Xapi_globs.varstore_dir && Sys.is_directory !Xapi_globs.varstore_dir then
```
lindig: We can keep this as is but I think

* it would have been enough to test for `is_directory`
* if the content is empty, a log line would probably be helpful - I assume this should not happen.

PR #3880, opened 2019-05-29
Support extra_args for vGPU configration

liulinC:

As titled, add support for extra_args for vgpu.

Customer can set extra args for the vgpu by command 
```xe vgpu-param-set uuid=xxxx extra_args="key1=v1,key2=v2"```
This is very similer to the previous verion of 
``` xe vm-param-set uuid=xxx platform:vgpu_extra_args="key1=v1,key2=v2"```


liulinC:
@lindig  I will rebase them to a single commits when we get the PR approved before merge to master.

liulinC:
@robhoes @lindig  May I ask Is there any further concern about this PR?

PR #3882, opened 2019-06-04
CP-31400: Update API error text.

michael2012z:
Replace "VGPU" with "virtual GPU", as VGPU should be avoided in user interface.

### DIFF
--- a/ocaml/idl/datamodel_errors.ml
+++ b/ocaml/idl/datamodel_errors.ml
@@ -277,7 +277,7 @@ let _ =
   error Api_errors.vgpu_type_not_compatible_with_running_type ["pgpu"; "type"; "running_type"]
     ~doc:"The VGPU type is incompatible with one or more of the VGPU types currently running on this PGPU" ();
   error Api_errors.vgpu_type_not_compatible ["type"]
-    ~doc:"You tried to create a VGPU that is not compatible with existing types on the VM." ();
+    ~doc:"Cannot create a virtual GPU that is incompatible with the existing types on the VM." ();
   error Api_errors.vgpu_destination_incompatible ["reason"; "vgpu"; "host"]
     ~doc:"The VGPU is not compatible with any PGPU in the destination." ();
   error Api_errors.nvidia_tools_error ["host"]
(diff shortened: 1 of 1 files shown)


lindig:
`VGPU` is used in many more places - i.e., right above. Should they not all be replaced?

MihaelaStoica:
> 
> 
> `VGPU` is used in many more places - i.e., right above. Should they not all be replaced?

In the C# SDK used by XenCenter, we have a mechanism of overwriting the API error text, so for example we replace the `vgpu_type_not_compatible_with_running_type` error message "VGPU type is not compatible with one or more of the VGPU types currently running on this PGPU"  with "The VM cannot start because all GPUs are fully used or are running other types of virtual GPU".
But we would prefer to not have to overwrite these messages.

michael2012z:
We use a lot of "VGPU" in XAPI error text,  they were replaced in XenCenter when presenting in GUI.
But this string here is very close to what XC need, changing it can save duplicate work in XC.

ocaml/idl/datamodel_errors.ml
```diff
@@ -277,7 +277,7 @@ let _ =
   error Api_errors.vgpu_type_not_compatible_with_running_type ["pgpu"; "type"; "running_type"]
     ~doc:"VGPU type is not compatible with one or more of the VGPU types currently running on this PGPU" ();
   error Api_errors.vgpu_type_not_compatible ["type"]
-    ~doc:"You tried to create a VGPU that is not compatible with existing types on the VM." ();
+    ~doc:"You tried to create a virtual GPU that is not compatible with the existing types on the VM." ();
```
kc284: This "You tried" bit should be removed completely (from all the messags in this file really) because it is not user friendly. It should read instead "Cannot create a virtual GPU that is incompatible with the existing types on the VM."

lindig:
@MihaelaStoica Are we talking about the same thing? I was talking about the `doc` string. Is XenCenter matching these or the more symbolic `vgpu_type_not_compatible_with_running_type`? I was only suggesting to refer to VGPUs consistently in these doc strings but not to change the symbolic names.

MihaelaStoica:
> 
> 
> @MihaelaStoica Are we talking about the same thing? I was talking about the `doc` string. Is XenCenter matching these or the more symbolic `vgpu_type_not_compatible_with_running_type`? I was only suggesting to refer to VGPUs consistently in these doc strings but not to change the symbolic names.

Yes, I was referring to the `doc` string.

edwintorok:
@michael2012z is this PR targeting the right branch? it looks like there are 15 commits including a lot of other vgpu code, not just the doc update.

michael2012z:
@edwintorok Thanks for reminding. I have rebased and adjusted the target branch.

michael2012z:
Thanks. Will update later.

michael2012z:
The string was updated.
I modified 2 "You tried" usages. There are more sentences with mentioning "You tried" in this file that need to be reorganized, but it's not good idea for me the modify too much English text. :)

PR #3884, opened 2019-06-05
REQ-720: Support multiple vGPUs in one VM

michael2012z:
Feature integration to master.


ocaml/xapi/xapi_vgpu.ml
```diff
@@ -17,9 +17,35 @@ open D
...
 let m = Mutex.create ()
 
-(* Only allow device = "0" for now, as we support just a single vGPU per VM *)
-let valid_device device =
-  device = "0"
+(* 1. Since multiple vgpus support, only device number = 0 or specify [11,31] since demu restriction.
+   2. device = 0 will requires xapi to give a valid number
+    - First sort the device number and the trying to find the next valid from range *)
+
+let min_valid_vgpu_device = 11
```
edwintorok: How does this work for UEFI?
According to the guest UEFI design document vGPU addresses should be assigned from 8 onwards.

(There is also the question of what happens if you mix SRIOV and vGPU, or vGPU and pass-through GPU, how do we avoid device number clashes? Although I think that case was problematic already even before this feature)

PR #3886, opened 2019-06-06
CA-320458: Upgrade vGPU default device id from 0 to 11

liulinC:

For XS with version before Naples, it has only one vGPU device with 0 as its default id.
For XS from Polymoth, it may have multiple vGPU device, 11 is the default device id.
This commit upgrade the default device ids from 0 to 11 during upgradation.

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -4771,7 +4771,7 @@ module VGPU = struct
         uid _vgpu ~lifecycle:[Published, rel_boston, ""];
         field ~qualifier:DynamicRO ~ty:(Ref _vm) ~lifecycle:[Published, rel_boston, ""] "VM" "VM that owns the vGPU";
         field ~qualifier:DynamicRO ~ty:(Ref _gpu_group) ~lifecycle:[Published, rel_boston, ""] "GPU_group" "GPU group used by the vGPU";
-        field ~qualifier:DynamicRO ~ty:String ~lifecycle:[Published, rel_boston, ""] ~default_value:(Some (VString "0")) "device" "Order in which the devices are plugged into the VM";
+        field ~qualifier:DynamicRO ~ty:String ~lifecycle:[Published, rel_boston, "Only 1 vGPU with device id 0 is supported"; Changed, rel_plymouth, "Multiple vGPUs are supported, device refers to guest PCI slot"] ~default_value:(Some (VString "0")) "device" "Guest PCI slot (a value of 0 means auto-assign to first empty slot, valid slot is in range of [11,31] for multi-VGPU purpose)";
         field ~qualifier:DynamicRO ~ty:Bool ~lifecycle:[Published, rel_boston, ""] ~default_value:(Some (VBool false)) "currently_attached" "Reflects whether the virtual device is currently connected to a physical device";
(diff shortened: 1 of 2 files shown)


ocaml/xapi/xapi_db_upgrade.ml
```diff
@@ -592,6 +592,19 @@ let upgrade_cluster_timeouts = {
         update_milliseconds Db.Cluster.get_token_timeout_coefficient Db.Cluster.set_token_timeout_coefficient;
     )
 }
+(* For XS Before Naples with single vGPU, it has 0 as its default device number,
+ * For XS After Naples with multiple vGPU, the valid device number is [11,31],
+ * 11 is the default device number.
+ * This function upgrade the default device number *)
+let upgrade_vgpu_default_device_id = {
+  description = "Upgrade default vGPU device ID from 0 to 11";
+  version = (fun _ -> true); (* Alwasys check and update when detected *)
```
edwintorok: Minor tyepo Alwasys/Always

edwintorok:
Should the documentation in datamodel.ml for 'device' be updated to say that since Plymouth a value of 0 means auto-assign to first empty slot, and values 11+ can be used to pick a specific slot for multi-vGPU purposes?

liulinC:
@edwintorok  update accordingly, please help to review.

ocaml/idl/datamodel.ml
```diff
@@ -4771,7 +4771,7 @@ module VGPU = struct
         uid _vgpu ~lifecycle:[Published, rel_boston, ""];
         field ~qualifier:DynamicRO ~ty:(Ref _vm) ~lifecycle:[Published, rel_boston, ""] "VM" "VM that owns the vGPU";
         field ~qualifier:DynamicRO ~ty:(Ref _gpu_group) ~lifecycle:[Published, rel_boston, ""] "GPU_group" "GPU group used by the vGPU";
-        field ~qualifier:DynamicRO ~ty:String ~lifecycle:[Published, rel_boston, ""] ~default_value:(Some (VString "0")) "device" "Order in which the devices are plugged into the VM";
+        field ~qualifier:DynamicRO ~ty:String ~lifecycle:[Published, rel_boston, ""] ~default_value:(Some (VString "0")) "device" "Since Plymouth a value of 0 means auto-assign to first empty slot, valid slot is in range of [11,31] for multi-VGPU purpose";
```
edwintorok: The part about 'since Plymouth' can be put into a `lifecycle` parameter like this (see example in datamodel_cluster.ml):
```
~lifecycle:[  Published, rel_boston, "Only 1 vGPU with device id 0 is supported"
               ;  Changed, rel_plymouth, "Multiple vGPUs are supported, device refers to guest PCI slot" ]
"Guest PCI slot (a value of 0 means auto-assign to first empty slot, valid slot is in range of [11,31] for multi-VGPU purpose)"
```

liulinC: thanks @edwintorok ,updated

robhoes:
We probably need to do something similar in the import code, to cover the cases of imports of old VMs and cross-pool migration from older server versions.

liulinC:
Thanks @robhoes , I will check it later.

PR #3887, opened 2019-06-11
Modifications to error messages; includes CA-258385.

kc284:
Apart from fixing CA-258385, this PR moves here most of the error messages with which the C# SDK overrides the original API errors. Most of the changes are conversion of 'host' to 'server', which I think is ok given there was already some mixed used of host/server in the messages. The advantage of moving those messages here is that we won't need anymore to override them in the C# SDK, and the corrected messages will be available in the other SDK flavours as well as the API docs, which means uniform terminology throughout components. Let me know if there are any concerns.

### DIFF
--- a/ocaml/idl/datamodel_errors.ml
+++ b/ocaml/idl/datamodel_errors.ml
@@ -620,7 +620,7 @@ let _ =
   error Api_errors.vm_requires_iommu ["host"]
     ~doc:"You attempted to run a VM on a host which doesn't have I/O virtualization (IOMMU/VT-d) enabled, which is needed by the VM." ();
   error Api_errors.vm_host_incompatible_version_migrate ["host"; "vm"]
-    ~doc:"You attempted to migrate a VM to a destination host which is older than the source host." ();
+    ~doc:"Cannot migrate a VM to a destination host which is older than the source host." ();
   error Api_errors.vm_host_incompatible_version ["host"; "vm"]
     ~doc:"This VM operation cannot be performed on an older-versioned host during an upgrade." ();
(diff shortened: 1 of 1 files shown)


ocaml/idl/datamodel_errors.ml
```diff
@@ -620,7 +620,7 @@ let _ =
   error Api_errors.vm_requires_iommu ["host"]
     ~doc:"You attempted to run a VM on a host which doesn't have I/O virtualization (IOMMU/VT-d) enabled, which is needed by the VM." ();
   error Api_errors.vm_host_incompatible_version_migrate ["host"; "vm"]
-    ~doc:"You attempted to migrate a VM to a destination host which is older than the source host." ();
+    ~doc:"Cannot migrate a VM to a destination host which is older than the source host." ();
```
lindig: This uses "host" where the preferred term seems to be "server". Or is "host" specifically used in the context of migration?

kc284: Hmm, no it's simpler than this, this belongs to the ones that have escaped our attention. We have converted host to server only in the overrides and this was not overriden (which is why I want to reduce the overrides to the bare minimum: if something needs rephrasing, there should be one place to do it, the API source).

lindig: Shall we merge and fix this some other time or would you like to change it?

kc284: Let's merge it now; there are more tweaks we can make in a future PR, but right now the objective of this PR is to strip down the SDK overrides.

PR #3888, opened 2019-06-12
List RRD directory once per monitor poll

psafont:
Before it was being listed 3 times per poll, once per each RRD monitored.

### DIFF
--- a/ocaml/xapi/monitor_mem_vms.ml
+++ b/ocaml/xapi/monitor_mem_vms.ml
@@ -19,7 +19,7 @@ module Mcache = Monitor_dbcalls_cache
 module D = Debug.Make(struct let name = "monitor_mem_vms" end)
 open D
 
-let get_changes () =
+let get_changes rrd_files =
   List.iter (fun filename ->
       try
         let datasources = Monitor_types.datasources_from_filename filename in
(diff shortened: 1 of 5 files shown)


ocaml/xapi/monitor_types.ml
```diff
@@ -37,7 +37,6 @@ let vif_device_of_string x =
 let find_rrd_files prefix =
```
robhoes: Should this now be `let find_rrd_files ()`?

psafont: indeed!

ocaml/xapi/monitor_mem_vms.ml
```diff
@@ -55,7 +55,9 @@ let set_changes ?except () =
       Mcache.transfer_map ?except ~source:Mcache.vm_memory_tmp ~target:Mcache.vm_memory_cached
     )
 
-let update () =
+let update rrd_files =
+  let is_vm_rrd = (Astring.String.is_prefix ~affix:Xapi_globs.metrics_prefix_mem_vms) in
```
lindig: No '(' .. ')' needed.

ocaml/xapi/monitor_pvs_proxy.ml
```diff
@@ -75,7 +75,9 @@ let pvs_proxy_status_of_int = function
   | 4 -> `incompatible_protocol_version
   | _ -> failwith "Unknown status"
 
-let update () =
+let update rrd_files =
+  let is_proxy_rrd = (Astring.String.is_prefix ~affix:Xapi_globs.metrics_prefix_pvs_proxy) in
```
lindig: No '(' .. ')' needed.

PR #3889, opened 2019-06-12
CA-312654: domains must be stopped before xapi is

psafont:
Otherwise VM migrations may be interrupted

### DIFF
--- a/scripts/xapi-domains.service
+++ b/scripts/xapi-domains.service
@@ -2,7 +2,8 @@
 Description=Start/stop domains on dom0 start/shutdown
 Requires=proc-xen.mount
 Wants=xapi-init-complete.target
-After=remote-fs.target xapi-init-complete.target
+# xapi.service is needed for shutdown ordering
+After=remote-fs.target xapi-init-complete.target xapi.service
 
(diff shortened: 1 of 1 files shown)


scripts/xapi-domains.service
```diff
@@ -2,7 +2,8 @@
 Description=Start/stop domains on dom0 start/shutdown
 Requires=proc-xen.mount
 Wants=xapi-init-complete.target
-After=remote-fs.target xapi-init-complete.target
```
edwintorok: The ordering should already be there from xapi-init-complete.target -> xapi-wait-init-complete.service -> xapi.service.
Although maybe systemd gets confused about target and service here?

edwintorok: Either way it is better to have xapi here explicitly, which we used to have prior to 49d0fbcef5b5712a3352762dae2ae97b4e28bf70#diff-4bbf7bc94efde8928c515b5fece9c92f

psafont: I'm testing the change, it's not clear to me why is xapi being stopped in parallel with xapi-domains.

PR #3890, opened 2019-06-14
CA-321983: Fix cross-pool migration for VGPU VMs

robhoes:


ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -708,6 +707,79 @@ module Nvidia = Vendor(Vendor_nvidia)
...
+        with Parse_error e ->
+          error "Ignoring error parsing %s: %s\n%s\n" conf_file
+            (Printexc.to_string e) (Printexc.get_backtrace ());
+          read_configs ac tl
+    in
+    let conf_files = Array.to_list (Sys.readdir conf_dir) in
+    debug "Reading NVIDIA vGPU config files %s/{%s}"
+      conf_dir (String.concat ", " conf_files);
+    read_configs []
+      (List.map (fun conf -> String.concat "/" [conf_dir; conf]) conf_files)
```
lindig: Use `Filename.concat`

ocaml/xapi/xapi_globs.ml
```diff
@@ -446,6 +446,8 @@ let vgt_fence_sz = "vgt_fence_sz"
 
 let mxgpu_vgpus_per_pgpu = "vgpus_per_pgpu"
 
+let nvidia_compat_lookup_file = "/tmp/nvidia_compat_lookup"
```
lindig: Should this filename have an extension? Like `.vgx` or similar?

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -708,6 +707,79 @@ module Nvidia = Vendor(Vendor_nvidia)
...
+             match (String.split ' ' s ~limit:2) with
+             | k :: [v] -> (k, v)
+             | _ -> ("", "")
+          ) args in
+      (* plugin0.pdev_id will either be just the physical device id, or of the
+       * form "device_id:subdevice_id" *)
+      let pdev_id, psubdev_id =
+        let pdev_id_data = (List.assoc "plugin0.pdev_id" args) in
+        try
+          Scanf.sscanf pdev_id_data "\"0x%x:0x%x\""
```
lindig: You could use `{|...|}` and avoid quoting `\"` - also below.

ocaml/xapi/xapi_globs.ml
edwintorok: What code uses this file?

robhoes: It's not a vgx file or anything like it, but our own data. I don't think that an extension helps?

`/var/run/nonpersistent/xapi/` may be better, yes.

robhoes:
I've added a "fixup", plus a new patch that depends on an idl change that I'll submit soon.

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -707,6 +707,79 @@ module Nvidia = Vendor(Vendor_nvidia)
...
+          pdev_id;
+          psubdev_id;
+          vdev_id;
+          vsubdev_id;
+        })
+      )
+    with e ->
+      raise (Parse_error e)
+
+  let read_config_dir conf_dir =
```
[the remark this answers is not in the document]
robhoes: This is parsing the old config files, which xapi isn't doing anymore, not the XML.

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -707,6 +707,79 @@ module Nvidia = Vendor(Vendor_nvidia)
...
+  let create_compat_lookup_file __context =
+    let open Db_filter_types in
+    read_config_dir conf_dir
+    |> List.map (fun (conf_file, identifier) ->
+      let identifier_string = Identifier.to_string identifier in
+      let expr = Eq (Field "identifier", Literal identifier_string) in
+      match Db.VGPU_type.get_internal_records_where ~__context ~expr with
+      | [_, rc] ->
+        let type_id = List.assoc Xapi_globs.vgpu_type_id rc.Db_actions.vGPU_type_internal_config in
+        Printf.sprintf "%s %s %s\n" conf_file type_id rc.Db_actions.vGPU_type_uuid
```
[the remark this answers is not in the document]
robhoes: OK, that's a mistake. I'll need to find another solution for that... :(

ocaml/xapi/xapi_vgpu_type.ml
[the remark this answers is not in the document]
liulinC: No, it is not from the old config file, it comes from the new xml file, and stored in the `internal_cofnig` field of `Vgpu_type` table, see
https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_vgpu_type.ml#L517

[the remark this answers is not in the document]
robhoes: I mean that this function is read the old conf files. We a mapping from conf file path to type_id. The conf file path is not present in the XML, and the type_id is not present in the conf files. So we have to read both and match them up based on the PCI device IDs (the "identifier").

robhoes:
I've updated the patches and added some more. The changes are:
- The old commit wrote the VGPU_type UUID to the compat file, but this isn't what is needed. Xenopsd needs the VGPU UUID, but we can't easily pass this one. Luckily, it isn't really required and the PR contains an alternative.
- The range of `VPGU.device` field is scaled back to start from 0, as it was before, to avoid complications.

ocaml/xapi/xapi_vgpu.ml
```diff
@@ -17,12 +17,15 @@ open D
...
+      limitations on the guest PCI bus.
+   2. The guest PCI bus slot will be equal to the device number plus 11.
+   3. device = 0 is recognised as a special value to automatically pick the next
+      available device number from the allowed range.
+*)
 
-let min_valid_vgpu_device = 11
-let max_valid_vgpu_device = 31
+let min_valid_vgpu_device = 0
+let max_valid_vgpu_device = 20
```
edwintorok: In the future we could have more devices here for UEFI, if we start assignments from 8, but that requires the PCI slot assignment rework in XAPI.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -653,7 +653,7 @@ module MD = struct
     {
       domain = 0000;
       bus = 0;
-      dev = int_of_string device;
```
liulinC: should we move `11` to the xapi_globals, and give it a meaningful name, as it looks important.

robhoes: Yes, good idea.

ocaml/xapi/xapi_vgpu.ml
liulinC: @edwintorok  the real pci address used in VM plugs 11, namely [11,31].

ocaml/xapi/xapi_xenops.ml
robhoes: I've updated it.

liulinC: sorry, I cannot find it, did you forgot to push your commit?

robhoes: I had pushed the wrong one :) It's there now...

PR #3891, opened 2019-06-14
CA-322044: throttle number of active SR scans

edwintorok:
This will need to be merged and tested on master first.
I have tested this only on Havana for now.


lindig:
Can we review and merge this?

PR #3892, opened 2019-06-17
CA-321787: Block migrate/suspend/resume when there is no pGPU compatible with vGPU

liulinC:
This commit check the NVML compatibility of the vGPU and the pGPU.
The vGPU can only be assigned bo a pGPU with NVML compatibility

1. The `Xapi_gpumon.Nvidia.vgpu_pgpu_are_compatible` is moved to Xapi_pgpu.assert_can_run_VGPU in https://github.com/xapi-project/xen-api/pull/3855/files#diff-3fb19a7212efdf3b669fd4374ac02838L315 as I throught every pGPU assert need such checks. But it caused other inconsistency behaviors when doing the RPU. So I revert it back.
2. Add the  `Xapi_gpumon.Nvidia.vgpu_pgpu_are_compatible` when allocate a vGPU to a pGPU to guarantee only the compatible pGPUs are allocated.

### DIFF
--- a/ocaml/xapi/vgpuops.ml
+++ b/ocaml/xapi/vgpuops.ml
@@ -65,6 +65,7 @@ let allocate_vgpu_to_gpu ?(dry_run=false) ?(pre_allocate_list=[]) ~__context vm 
   (* Make a asso list as follows
    * [(p1,capacity1);(p2,capacity2)...] *)
   let pgpu_capacity_assoc = List.intersect compatible_pgpus available_pgpus
+                            |> List.filter (fun pgpu -> Xapi_gpumon.Nvidia.vgpu_pgpu_are_compatible ~__context ~pgpu ~vgpu:vgpu.vgpu_ref) (* Filter all compatible pGPUs *)
                             |> List.map (fun self -> (self, Xapi_pgpu_helpers.get_remaining_capacity  ~__context ~pre_allocate_list ~self ~vgpu_type))
                             |> List.filter (fun (_,capacity) -> capacity > 0L ) in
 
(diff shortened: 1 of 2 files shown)


ocaml/xapi/xapi_pgpu.ml
```diff
@@ -312,12 +312,7 @@ let get_remaining_capacity ~__context ~self ~vgpu_type =
 
 let assert_can_run_VGPU ~__context ~self ~vgpu =
```
robhoes: I can't seem to find any callers to this function apart from in the unit tests. Am I missing something?

robhoes: I think that it was removed in 30d9517cd27450ca06beff0b12f44e966c030703 and replaced by the dry-run stuff in `allocate_vgpu_to_gpu`?

ocaml/xapi/vgpuops.ml
```diff
@@ -65,6 +65,7 @@ let allocate_vgpu_to_gpu ?(dry_run=false) ?(pre_allocate_list=[]) ~__context vm
   (* Make a asso list as follows
    * [(p1,capacity1);(p2,capacity2)...] *)
   let pgpu_capacity_assoc = List.intersect compatible_pgpus available_pgpus
+                            |> List.filter (fun pgpu -> Xapi_gpumon.Nvidia.vgpu_pgpu_are_compatible ~__context ~pgpu ~vgpu:vgpu.vgpu_ref) (* Filter all compatible pGPUs *)
```
robhoes: I believe we now rely on this function to be called by `Xapi_vm_helpers.assert_gpus_available` with `dry_run=true`, and when this gpumon check returns says that the GPU is incompatible, then an error is raised in L84.

robhoes: So did we somehow lose the compat check completely?

PR #3894, opened 2019-06-21
CA-319960: Remove user KRBTGT from cache before checking

minglumlu:
PBIS will return success when it queries user KRBTGT even if the domain is
offline but the user KRBTGT is in the cache. This makes the XAPI fail to
check the availability of domain by this appoach.

This commit adds logic to remove the user KRBTGT from cache before XAPI
checks the availabililty of domain.

### DIFF
--- a/ocaml/xapi/extauth_plugin_ADpbis.ml
+++ b/ocaml/xapi/extauth_plugin_ADpbis.ml
@@ -71,6 +71,8 @@ struct
 
   let user_friendly_error_msg = "The Active Directory Plug-in could not complete the command. Additional information in the logs."
 
+  let mutex_check_availability = Locking_helpers.Named_mutex.create "IS_SERVER_AVAILABLE"
+
   open Stdext.Xstringext
 
   let splitlines s = String.split_f (fun c -> c = '\n') (String.replace "#012" "\n" s)
@@ -521,7 +523,8 @@ struct
     In addition, there are some event hooks that auth modules implement as follows:
 *)
 
-  let is_pbis_server_available max =
+  let _is_pbis_server_available max =
+    let username = "KRBTGT" in (* domain name prefix automatically added by our internal AD plugin functions *)
     let rec test i = (* let's test this many times *)
       if i > max then false (* we give up *)
       else begin (* let's test *)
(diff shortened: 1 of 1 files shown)


ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -563,9 +556,18 @@ struct
     in
     begin
       debug "Testing if external authentication server is accepting requests...";
+      let full_username = get_full_subject_name username in
+      begin try
+          ignore(pbis_common "/opt/pbis/bin/ad-cache" ["--delete-user"; "--name"; full_username])
+        with _ -> ()
```
lindig: I believe in case of an exception, this should be logged. Should this not also mean that the function should return false?

edwintorok: I agree about logging the exception.
What happens if the user we ask to delete is not in the cache, would pbis fail? I couldn't find a manpage for `ad-cache` that would document its behaviour.

minglumlu: Sure. I will log the exception to make it more clear.
If the user does not exist in cache, pbis will fail. This deletion is just to make sure it is not in cache. This forces `test 0` to retrieve the user from remote server, rather than cache.

minglumlu:
Hi @mg12 
May I please have your review on this change? Thanks.

ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -521,27 +523,18 @@ struct
...
-           let sid = get_subject_identifier username in (* use our well-known KRBTGT builtin username in AD *)
+           ignore(get_subject_identifier username); (* use our well-known KRBTGT builtin username in AD *)
            (* OK, we found this username! *)
            debug "Request %i/%i to external authentication server successful: user %s was found" i max username;
-
-           (* (2) CA-25427: test (1) above may succeed (because of pbis caching stale AD information) *)
-           (* even though the AD domain is offline (error 32888), usually because /etc/resolv.conf is not *)
-           (* pointing to the AD server. This test should catch if the domain is offline by calling find-by-sid *)
-           (* using a domain SID. We must use a _domain_ SID. A universal SID like S-1-1-0 doesn't work for this test. *)
-           let (_: (string*string) list) = query_subject_information sid in (* use KRBTGT's domain SID *)
```
mg12: any special reason for removing this check? otherwise, we should probably keep it to have a 2-layer defense for the domain offline problem.

minglumlu: My understanding is based on the comment:
"(* (2) CA-25427: test (1) above may succeed (because of pbis caching stale AD information)"

The fix makes sure that pbis has no stale information from cache, the first check should be enough for the availability check. 
Please correct me if my understanding is wrong.
Or in reverse, is it possible that the first check passed but the second one failed?

minglumlu:
Hi @mg12  
Could you please look at the change again? I just kept the test () to minimize the change. Thanks.

PR #3895, opened 2019-06-24
CA-322450 xe-restore-metadata: ImportError: fsimage

lindig:
The Python module fsimage has been renamed to xenfsimage as part of Xen 4.12.
This commit changes the client accordingly.

### DIFF
--- a/scripts/probe-device-for-file
+++ b/scripts/probe-device-for-file
@@ -3,7 +3,10 @@
 # Checks for the existence of a file on a device
 
 import os, sys
-import fsimage
+try:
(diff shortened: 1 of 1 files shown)


scripts/probe-device-for-file
```diff
@@ -3,7 +3,10 @@
 # Checks for the existence of a file on a device
 
 import os, sys
-import fsimage
+try:
+   import xenfsimage
+except: ImportError
```
edwintorok: I think the `:` should go at the end of the line

PR #3898, opened 2019-06-25
Open VxLAN port for SDN controller

benjamreis:
The commit opens the VxLAN IANA port (4789 udp) when a SDN controller is introduced and close it when its removed. 

This allow the SDN controller to creates VxLAN without having to manually change iptables rules.

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -1388,12 +1388,19 @@ module VLAN = struct
 end
 
 module Tunnel = struct
+  let tunnel_protocol = Enum ("tunnel_protocol", [
+      "gre", "GRE protocol";
+      "vxlan", "VxLAN Protocol";
+    ])
 
   let create = call
       ~name:"create"
       ~doc:"Create a tunnel"
-      ~params:[ Ref _pif, "transport_PIF", "PIF which receives the tagged traffic";
-                Ref _network, "network", "Network to receive the tunnelled traffic" ]
+      ~versioned_params:[
+        {param_type=Ref _pif; param_name="transport_PIF"; param_doc="PIF which receives the tagged traffic"; param_release=dundee_release; param_default=None};
+        {param_type=Ref _network; param_name="network"; param_doc="Network to receive the tunnelled traffic"; param_release=dundee_release; param_default=None};
+        {param_type=tunnel_protocol; param_name="protocol"; param_doc="Protocol used for the tunnel (GRE or VxLAN)"; param_release=next_release; param_default=Some(VEnum "gre")}
+      ]
       ~result:(Ref _tunnel, "The reference of the created tunnel object")
       ~lifecycle:[Published, rel_cowley, "Create a tunnel"]
       ~allowed_roles:_R_POOL_OP
@@ -1409,7 +1416,8 @@ module Tunnel = struct
       ()
 
   let t =
-    create_obj ~in_db:true ~lifecycle:[Published, rel_cowley, "A tunnel for network traffic"] ~in_oss_since:None ~persist:PersistEverything ~gen_constructor_destructor:false ~name:_tunnel ~descr:"A tunnel for network traffic" ~gen_events:true
+    create_obj ~in_db:true ~lifecycle:[Published, rel_cowley, "A tunnel for network traffic"]
+      ~in_oss_since:None ~persist:PersistEverything ~gen_constructor_destructor:false ~name:_tunnel ~descr:"A tunnel for network traffic" ~gen_events:true
       ~doccomments:[]
       ~messages_default_allowed_roles:_R_POOL_OP
       ~doc_tags:[Networking]
@@ -1420,6 +1428,7 @@ module Tunnel = struct
           field ~qualifier:StaticRO ~ty:(Ref _pif) ~lifecycle:[Published, rel_cowley, "The interface used by the tunnel"] "transport_PIF" "The interface used by the tunnel" ~default_value:(Some (VRef ""));
           field ~ty:(Map(String, String)) ~lifecycle:[Published, rel_cowley, "Status information about the tunnel"] "status" "Status information about the tunnel" ~default_value:(Some (VMap [VString "active", VString "false"]));
           field ~lifecycle:[Published, rel_cowley, "Additional configuration"] ~default_value:(Some (VMap [])) ~ty:(Map(String, String)) "other_config" "Additional configuration";
+          field ~ty:tunnel_protocol ~default_value:(Some (VEnum "gre")) ~lifecycle:[Published, rel_next, "Add protocol field to tunnel"] "protocol" "The protocol used for tunneling (either GRE or VxLAN)";
         ])
       ()
 end
--- a/ocaml/idl/datamodel_common.ml
+++ b/ocaml/idl/datamodel_common.ml
@@ -106,7 +106,7 @@ let tech_preview_releases = [
 (* Normally xencenter_min_verstring and xencenter_max_verstring in the xapi_globs should be set to the same value,
  * but there are exceptions: please consult the XenCenter maintainers if in doubt. *)
 let api_version_major = 2L
-let api_version_minor = 15L
+let api_version_minor = 16L
 let api_version_string =
   Printf.sprintf "%Ld.%Ld" api_version_major api_version_minor
 let api_version_vendor = "XenSource"
@@ -200,6 +200,12 @@ let get_product_releases in_product_since =
     | x::xs when code_name_of_release x = in_product_since -> "closed"::in_product_since::(List.map code_name_of_release xs)
     | x::xs -> go_through_release_order xs
   in go_through_release_order release_order
+
+let next_release =
+  { internal = get_product_releases rel_next
+  ; opensource=get_oss_releases None
+  ; internal_deprecated_since=None
+  }
 
 let stockholm_release =
   { internal = get_product_releases rel_stockholm
--- a/ocaml/xapi-cli-server/cli_operations.ml
+++ b/ocaml/xapi-cli-server/cli_operations.ml
@@ -3829,7 +3829,10 @@ let vlan_destroy printer rpc session_id params =
 let tunnel_create printer rpc session_id params =
   let network = Client.Network.get_by_uuid rpc session_id (List.assoc "network-uuid" params) in
   let pif = Client.PIF.get_by_uuid rpc session_id (List.assoc "pif-uuid" params) in
-  let tunnel = Client.Tunnel.create rpc session_id pif network in
+  let protocol = Record_util.tunnel_protocol_of_string (
+    List.assoc_opt "protocol" params |> Option.value ~default:"gre"
+  ) in
+  let tunnel =  Client.Tunnel.create rpc session_id pif network protocol in
   let pif' = Client.Tunnel.get_access_PIF rpc session_id tunnel in
   let uuid = Client.PIF.get_uuid rpc session_id pif' in
   printer (Cli_printer.PList [uuid])
--- a/ocaml/xapi/nm.ml
+++ b/ocaml/xapi/nm.ml
@@ -485,19 +485,26 @@ let bring_pif_up ~__context ?(management_interface=false) (pif: API.ref_PIF) =
           let igmp_snooping' = if igmp_snooping then `enabled else `disabled in
           if igmp_snooping' <> rc.API.pIF_igmp_snooping_status then
             Db.PIF.set_igmp_snooping_status ~__context ~self:pif ~value:igmp_snooping'
-        end
+        end;
+
+        (* Open VxLAN UDP port (4789) if a VxLAN tunnel access PIF is plugged *)
+        match Xapi_pif_helpers.get_pif_type rc with
+        | Tunnel_access tunnel_ref ->
+          let protocol = Db.Tunnel.get_protocol ~__context ~self:tunnel_ref in
+          begin match protocol with
+          | `vxlan ->
+            debug "Opening VxLAN UDP port for tunnel with protocol 'vxlan'";
+            ignore @@ Helpers.call_script !Xapi_globs.firewall_port_config_script ["open"; "4789"; "udp"]
+          | `gre -> ()
+          end;
+        | _ -> ()
     )
 
 let bring_pif_down ~__context ?(force=false) (pif: API.ref_PIF) =
   with_local_lock (fun () ->
       let rc = Db.PIF.get_record ~__context ~self:pif in
       let open Xapi_pif_helpers in
-      match get_pif_topo ~__context ~pif_rec:rc with
-      | Network_sriov_logical _ :: _ ->
-        Xapi_network_sriov_helpers.sriov_bring_down ~__context ~self:pif
-      | VLAN_untagged _ :: Network_sriov_logical _ :: _ ->
-        Db.PIF.set_currently_attached ~__context ~self:pif ~value:false
-      | _ ->
+      let close_network_interface cleanup_f () =
         Network.transform_networkd_exn pif (fun () ->
             let dbg = Context.string_of_task __context in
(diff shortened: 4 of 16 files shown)


benjamreis:
Hi!

Any update on this?
Is there an issue with the code?
Perhaps I can answer some questions?

Thanks!

psafont:
Sorry for the late reply.

There's a general sense of uncertainty about the security implications of opening that port. We're trying to get ahold of somebody that can generate a threat model around it.

olivierlambert:
Hi there! Any news on the consequence to do so?

olivierlambert:
Okay so a bit of context explaining it:

1. Our [SDN controller](https://xen-orchestra.com/blog/xo-sdn-controller/) is using the exact same calls than the Citrix DVSC controller.
2. OVS in all XenServer 7.x supports both GRE and VxLAN tunnels
3. DVSC only uses GRE tunnels because it was engineered before VxLAN support in OVS

So this PR only allow usage of VxLAN tunnels through the firewall, for hosts being able to create and communicate through those tunnels. It doesn't change the way a SDN controller (DVSC or XO SDN) are talking to OVS, nor add a "new feature" in OVS.

benjamreis:
I've added a commit to limitate the opening of the VxLAN port when the sdn-controller is not DVSC because it is not needed.

benjamreis:
So here is what this PR does with my latest commits:
- When there's no sdn controller set, the VxLAN port (udp 4789) is closed
- If a controller is set using active ssl (It's the case for Citrix' DVSC) the VxLAN port is closed.
- if a controller is set using passive ssl (it's the case for XO's SDN Controller) the VxLAN port is open.

So only when the SDN controller is `pssl:6640` the port is open, is it enough? I can't restrain it more because there's no way to differentiate XO's Controller other than `pssl:6640`. On top of that I believe that another SDN Controller could want to use VxLAN (even in the active ssl case possibly).

What do you think?

robhoes:
Hi there, I'm sorry for joining the conversation so late.

My first response was going to be that we should be very careful with opening new ports in the firewall, because it widens the attack surface of the hosts. Therefore, as a rule, ports should only be opened when needed, and when the user can reasonably expect this to happen.

You have already partially addressed this by further restricting when the port is opened. However, as you hinted at yourself, certain assumptions are used in the patch to derive which kind of controller is connected and what its needs are, that will not be true in general.

Coming back to the general rule above, it follows that we should only open the VxLAN port when the user enables a feature that requires VxLAN. The function that you are modifying is one that enables an SDN controller and setting up the OVS connection to it, and has no explicit awareness of specific controller features, such as VxLAN tunnels.

I would like to propose an alternative, by extending the `tunnel` API. Currently, tunnels can be created and destroyed in the API, and the controller is responsible for setting up actual tunnel end-points for them on the virtual switches. Citrix' DVSC uses GRE tunnels, and your controller uses VxLAN, but the choice of protocol is not explicit in the `tunnel` API (see http://xapi-project.github.io/xen-api/classes/tunnel.html). How about adding a field to the `tunnel` class for this, e.g. `tunnel.protocol` with enum values `gre` and `vxlan`. Then when the PIF belonging to a tunnel is plugged, the VxLAN port will be opened only if `tunnel.protocol = vxlan`, and closed on unplug.

What do you think?

olivierlambert:
I think this is doable in the future, BUT will require more changes all around layers (and backport changes on previous versions).

Also @robhoes  can you estimate the extra work needed to achieve that, roughly?

@benjamreis and @stormi what do you think?

benjamreis:
Hi,

I think it's a good idea, but harder to implement. i guess it can be tricky to know when to close the VxLAN port if there's many tunnel using VxLAN and one is unplugged for instance.

But it is a cleaner solution for sure.

robhoes:
In the first instance, we could omit closing the port on unplug to simplify. Otherwise, the xapi changes are very simple. Looking at your other PR, you already know how to change/add API fields. For the PIF plug code, look in nm.ml (bring_pif_up).

benjamreis:
I'm focusing on my other PR right now, once i'm done with it i'll do the refont of this one. :)

Thanks for all your inputs!

benjamreis:
Small question: the auto-completion of [xe-cli proposes `ssl` as a default protocol](https://github.com/xapi-project/xen-api/blob/master/ocaml/xe-cli/bash-completion#L373-L377) (for the `sdn-controller` protocol) is it possible to change it in the case of tunnels?
Should I just change the of the param? USe `tunnel-protocol` seems to specific since it's an arg or `tunnel-create`...

What do you think?

ocaml/xapi/xapi_bond.ml
```diff
@@ -167,7 +167,7 @@ let move_tunnel ~__context host new_transport_PIF old_tunnel =
 
   (* Create new tunnel object and access PIF *)
   let new_tunnel, new_access_PIF =
-    Xapi_tunnel.create_internal ~__context ~transport_PIF:new_transport_PIF ~network ~host in
+    Xapi_tunnel.create_internal ~__context ~transport_PIF:new_transport_PIF ~network ~host ~protocol:`gre in
```
robhoes, line 170: The protocol should be copied from the tunnel that is being moved.

ocaml/xapi/xapi_host.ml
```diff
@@ -1698,7 +1698,7 @@ let sync_tunnels ~__context ~host =
         | [(pif_ref,_)] ->
           (* this is the PIF on which we want as transport PIF; let's make it *)
           ignore (Xapi_tunnel.create_internal ~__context ~transport_PIF:pif_ref
-                    ~network:master_pif_rec.API.pIF_network ~host)
+                    ~network:master_pif_rec.API.pIF_network ~host ~protocol:`gre)
```
robhoes, line 1701: Here also we must copy the protocol from the tunnel that is being copied.

benjamreis:
Done. I'm working on the code to open/close the vxlan port in `nm.ml` now.

benjamreis:
My latest commit open the UDP VxLAN port (4789) when an `access_PIF` of a `tunnel` with `protocol=vxlan` is plugged.
When the last `tunnel` with `protocol=vxlan` `access_PIF` is unplugged, the port is closed.

benjamreis:
Any feedback is welcome! :)

benjamreis:
Hi @robhoes ! On top of reviewing my PR when you can, can you answer me as well:
I'm implementing a new feature for our SDN Controller: managing openflow rules. Therefore I need to open the openflow port (`TCP 6653`) on a XenServer. Where should I put the code to open it?

benjamreis:
Up to date with master :)

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -3823,7 +3823,8 @@ let vlan_destroy printer rpc session_id params =
 let tunnel_create printer rpc session_id params =
   let network = Client.Network.get_by_uuid rpc session_id (List.assoc "network-uuid" params) in
   let pif = Client.PIF.get_by_uuid rpc session_id (List.assoc "pif-uuid" params) in
-  let tunnel = Client.Tunnel.create rpc session_id pif network in
+  let protocol = Record_util.tunnel_protocol_of_string (Listext.assoc_default "protocol" params "gre") in
```
psafont, line 3826: Could you use `(List.assoc_opt params "protocol" |> Option.value ~default:"gre")`? We're trying to stop using the Listext library.

ocaml/xapi/xapi_host.ml
```diff
@@ -1737,8 +1737,12 @@ let sync_tunnels ~__context ~host =
           ()
         | [(pif_ref,_)] ->
           (* this is the PIF on which we want as transport PIF; let's make it *)
+          let (_, tunnel_rec) = List.hd (Db.Tunnel.get_records_where ~__context ~expr: (
```
psafont, line 1740: `List.hd` will raise an exception on an empty list, please treat that case with a match structure instead to raise an appropriate error:
```
let tunnel_rec = match Db.tunnel.get_ ... with
| (_, tunnel_rec) :: _ -> tunnel_rec
| [] -> raise Server_error ...
```

benjamreis: Done! :)

ocaml/xapi-cli-server/cli_operations.ml
benjamreis: Done! :)

ocaml/xe-cli/bash-completion
```diff
@@ -360,8 +360,16 @@ _xe()
...
-                IFS=$'\n,'
-                set_completions 'ssl' "$value"
+                case "${OLDSTYLE_WORDS[1]}" in
+                    sdn-controller-introduce)
+                        IFS=$'\n,'
+                        set_completions 'ssl' "$value"
+                        ;;
+                    tunnel-create)
+                        IFS=$'\n,'
+                        set_completions 'gre' "$value"
```
edwintorok, line 370: Isn't `vxlan` a valid completion too now?

benjamreis: Done! :)

ocaml/xapi/xapi_host.ml
```diff
@@ -1737,8 +1737,16 @@ let sync_tunnels ~__context ~host =
           ()
         | [(pif_ref,_)] ->
           (* this is the PIF on which we want as transport PIF; let's make it *)
+          let tunnel_rec =
+            match Db.Tunnel.get_records_where ~__context ~expr: (
+              Eq (Field "transport_PIF", Literal (Ref.string_of pif_ref))
```
robhoes, line 1742: I think that you will always get an empty list here, because you are looking for the tunnel that you are about the create.

You need to get the protocol from the tunnel of the master through `master_pif_rec`:
```
let protocol =
  match Xapi_pif_helpers.get_pif_type master_pif_rec with
  | Tunnel_access tunnel -> Db.Tunnel.get_protocol ~__context ~self:tunnel
  | _ -> (* raise internal error *)
in
```
Bonus points for merge this with `get_network_of_transport_pif` above and saving some DB calls.

ocaml/idl/datamodel.ml
```diff
@@ -1388,14 +1388,22 @@ module VLAN = struct
...
+      "vxlan", "VxLAN Protocol";
+    ])
 
   let create = call
       ~name:"create"
       ~doc:"Create a tunnel"
       ~params:[ Ref _pif, "transport_PIF", "PIF which receives the tagged traffic";
-                Ref _network, "network", "Network to receive the tunnelled traffic" ]
+                Ref _network, "network", "Network to receive the tunnelled traffic";
+                tunnel_protocol, "protocol", "Protocul used for the tunnel (GRE or VxLAN)" ]
```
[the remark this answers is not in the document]
psafont: I've used stockholm for the time being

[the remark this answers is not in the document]
robhoes: Are you doing to remember to change it?

ocaml/xapi/xapi_host.ml
```diff
@@ -1725,26 +1726,18 @@ let sync_tunnels ~__context ~host =
     if existing_pif = []
     then
       begin
-        (* On the master, we find the network the tunnel transport PIF is on *)
-        let network_of_transport_pif_on_master = get_network_of_transport_pif master_pif_rec in
+        (* On the master, we find the network the tunnel transport PIF is on and its protocol *)
+        let master_transport_pif, protocol = get_network_of_transport_pif master_pif_rec in
```
robhoes, line 1730: The old name was better than `master_transport_pif`: its a network, not a PIF.

ocaml/xapi/nm.ml
```diff
@@ -520,32 +515,44 @@ let bring_pif_down ~__context ?(force=false) (pif: API.ref_PIF) =
...
+      in
+      match get_pif_topo ~__context ~pif_rec:rc with
+      | Network_sriov_logical _ :: _ ->
+        Xapi_network_sriov_helpers.sriov_bring_down ~__context ~self:pif
+      | VLAN_untagged _ :: Network_sriov_logical _ :: _ ->
+        Db.PIF.set_currently_attached ~__context ~self:pif ~value:false
+      | Tunnel_access tunnel_ref :: _ ->
+          (* If last tunnel with VxLAN protocol is unplugged, close VxLAN UDP port *)
+          let maybe_close_port () =
+            let host = Db.PIF.get_host ~__context ~self:pif in
```
robhoes, line 529: You have the record, so avoid the DB call:

     let host = rc.API.pIF_host in

ocaml/xapi/nm.ml
```diff
@@ -520,32 +515,44 @@ let bring_pif_down ~__context ?(force=false) (pif: API.ref_PIF) =
...
+            begin match Db.Tunnel.get_protocol ~__context ~self:tunnel_ref with
+            | `vxlan ->
+              let expr =
+                And (Eq (Field "protocol", Literal (Record_util.tunnel_protocol_to_string `vxlan))
+                 , Eq (Field "host", Literal (Ref.string_of host)))
```
robhoes, line 534: Unfortunately, the `tunnel` class does not have a `host` field. You need to get it off one of the attached PIFs. So first get all vxlan tunnels, then map to their access PIFs, and filter out the ones that don't match `host` and are not currently_attached. Probably change the code below into a fold and do the last two steps all at once.

ocaml/idl/datamodel.ml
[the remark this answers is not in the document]
psafont: I'll add the new release as part of this PR

ocaml/idl/datamodel_common.ml
```diff
@@ -106,7 +106,7 @@ let tech_preview_releases = [
 (* Normally xencenter_min_verstring and xencenter_max_verstring in the xapi_globs should be set to the same value,
  * but there are exceptions: please consult the XenCenter maintainers if in doubt. *)
 let api_version_major = 2L
-let api_version_minor = 15L
+let api_version_minor = 16L
```
psafont, line 109: Should we bump the minor version or is it better to wait?

benjamreis:
Commits squashed wth more explicit commit message.

ocaml/idl/datamodel.ml
```diff
@@ -1420,6 +1428,7 @@ module Tunnel = struct
           field ~qualifier:StaticRO ~ty:(Ref _pif) ~lifecycle:[Published, rel_cowley, "The interface used by the tunnel"] "transport_PIF" "The interface used by the tunnel" ~default_value:(Some (VRef ""));
           field ~ty:(Map(String, String)) ~lifecycle:[Published, rel_cowley, "Status information about the tunnel"] "status" "Status information about the tunnel" ~default_value:(Some (VMap [VString "active", VString "false"]));
           field ~lifecycle:[Published, rel_cowley, "Additional configuration"] ~default_value:(Some (VMap [])) ~ty:(Map(String, String)) "other_config" "Additional configuration";
+          field ~ty:tunnel_protocol ~default_value:(Some (VEnum "gre")) ~lifecycle:[Published, rel_stockholm, "Add protocol field to tunnel"] "protocol" "The protocol used for tunneling (either GRE or VxLAN)";
```
robhoes, line 1431: `rel_next`

stormi:
So, it turns out that tunnel protocol field will cause a transition issue for us because we have users that already have vxlan tunnels open and after an upgrade their tunnels will be seen as GRE (default value). This would imply that users would suddently not be able to use their tunnel anymore, due to the port being closed. And since the protocol field is StaticRO apparently we can't update it after that.

We do have the information elsewhere in the DB, so we could update that field when we detect that it needs updating, but we are not familiar with the way one can handle that kind of situation during DB upgrades, and whether there is a way to set an initial value for a StaticRO field that is different from the default.

What would you advise to do?

lindig:
In general, database upgrade are handled in `ocaml/xapi/xapi_db_upgrade.ml` which has a set of rules (at the bottom) which are invoked on update and can be used to modify the database.

PR #3899, opened 2019-06-26
CA-321930 XSI-374 GPU compatibility check not done on snapshot snapshot

lindig:
When saving a snapshot, make sure to save VGPU metadata such that it is
available on resume.

### DIFF
--- a/ocaml/xapi/xapi_vgpu.ml
+++ b/ocaml/xapi/xapi_vgpu.ml
@@ -50,7 +50,8 @@ let get_valid_device ~__context ~device ~vM ~vGPUs =
   else
     raise Api_errors.(Server_error (invalid_device, [device]))
 
-let create' ~__context  ~vM ~gPU_group ~device ~other_config ~_type ~powerstate_check =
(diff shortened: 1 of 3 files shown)


ocaml/xapi/xapi_vgpu.ml
```diff
@@ -119,6 +123,7 @@ let copy ~__context ~vm vgpu =
       ~other_config:all.API.vGPU_other_config
       ~_type:all.API.vGPU_type
       ~powerstate_check:false
+      ~compatibility_metadata:all.API.vGPU_compatibility_metadata
```
edwintorok: So the bug here was that during a VM copy/snapshot we did not preserve the compatibility metadata?

lindig: Yes. We missed that while developing the feature.

robhoes: ..._and_ we didn't create it, which is why the patch to xapi_vm_snapshot.ml is needed.

PR #3901, opened 2019-07-01
CA-322749: Make toolstack work with old Nvidia drivers

liulinC:
This PR make xapi work with old NVidia drivers that has no multiple vGPU support as followings

   - Check whether the Nvidia host installed suport multiple vGPU
   -  Make up old config_file argument for demu in case host driver does not support multiple vGPU
   - Make all VGPU_type does not compatible with any one (including themselves) if host driver does not support multiple vGPU

How xapi judge whether a Nvidia driver support multple vGPU:
Add configration variable nvidia_multi_vgpu_enabled_driver_versions
   
   nvidia_multi_vgpu_enabled_driver_versions is a list and configurable
    in xapi confie file /etc/xapi.conf, it stating the version of Nvidia
    host drivers that have multiple vGPU support. The elements in the list
    have following two formats:
    1. "major_version.minor_versoin": the specific version that support
    multiple vGPU.
    2. "major_version.minor_version+": any version biger than this version
    is taken to support multiple vGPU.
Currently installed Nvidia vGPU version is read from file:  `/usr/lib64/libnvidia-vgpu.so`, per suggested by  Rob.

I will delete the `nvidia_compat_lookup_file` related code after this PR merged.

### DIFF
--- a/ocaml/tests/test_vgpu_type.ml
+++ b/ocaml/tests/test_vgpu_type.ml
@@ -21,72 +21,7 @@ open Xapi_vgpu_type
 let mib x = List.fold_left Int64.mul x [1024L; 1024L]
 
 module NvidiaTest = struct
-  (*
-  let string_of_vgpu_conf conf =
-    let open Identifier in
-    let open Nvidia_old in
-    Printf.sprintf "%04x %s %04x %04x %Ld"
-      conf.identifier.pdev_id
-      (match conf.identifier.psubdev_id with
-       | Some id -> Printf.sprintf "Some %04x" id
-       | None -> "None")
-      conf.identifier.vdev_id
-      conf.identifier.vsubdev_id
-      conf.framebufferlength
-
-  let print_vgpu_conf conf =
-    Printf.printf "%s\n" (string_of_vgpu_conf conf)
-
-  module OfConfFile = Generic.Make(struct
-      module Io = struct
-        type input_t = string
-        type output_t = Nvidia_old.vgpu_conf
-
-        let string_of_input_t x = x
-        let string_of_output_t = string_of_vgpu_conf
-      end
-
-      let transform = Nvidia_old.of_conf_file
-
-      let tests = [
-        "test_data/test_vgpu_subdevid.conf",
-        Nvidia_old.({
-            identifier = Identifier.({
-                pdev_id = 0x3333;
-                psubdev_id = Some 0x4444;
-                vdev_id = 0x1111;
-                vsubdev_id = 0x2222;
-              });
-            framebufferlength = 0x10000000L;
-            num_heads = 2L;
-            max_instance = 8L;
-            max_x = 1920L;
-            max_y = 1200L;
-            file_path = "test_data/test_vgpu_subdevid.conf";
-            compatible_model_names_in_vm = [];
-            compatible_model_names_on_pgpu = [];
-          });
-        "test_data/test_vgpu_nosubdevid.conf",
-        Nvidia_old.({
-            identifier = Identifier.({
-                pdev_id = 0x3333;
-                psubdev_id = None;
-                vdev_id = 0x1111;
-                vsubdev_id = 0x2222;
-              });
-            framebufferlength = 0x10000000L;
-            num_heads = 2L;
-            max_instance = 8L;
-            max_x = 1920L;
-            max_y = 1200L;
-            file_path = "test_data/test_vgpu_nosubdevid.conf";
-            compatible_model_names_in_vm = [];
-            compatible_model_names_on_pgpu = [];
-          });
-      ]
-    end)
-*)
-  let string_of_vgpu_conf conf =
+    let string_of_vgpu_conf conf =
     let open Identifier in
     let open Vendor_nvidia in
     Printf.sprintf "%04x %s %04x %04x %Ld %Ld %Ld %Ldx%Ld"
@@ -199,6 +134,33 @@ module NvidiaTest = struct
             })
         ];
         
+      ]
+    end)
+
+  module HostDriverMultipleVgpuSupport = Generic.Make(struct
+      module Io = struct
+        type input_t = (string * string list)
+        type output_t = bool
+
+        let string_of_input_t (host_driver_version, supported_driver_versions) =
+          Printf.sprintf "(%s, %s)" host_driver_version (String.concat "," supported_driver_versions)
+
+        let string_of_output_t x = string_of_bool x
+      end
+
+      let transform (host_driver_version, supported_versions) = Xapi_vgpu_type.Vendor_nvidia.host_driver_supports_multi_vgpu ~host_driver_version ~supported_versions
+
+      let tests = [
+        ("430.19",["430.19"]), true; (* Test specific version should match *)
+        ("430.19",["430.89";"430.19"]), true; (* Test specific version should match *)
+        ("430.19",["430.19";"typo_error"]), true; (* Test specific version should match *)
+        ("430.19",["430.19typo"]), false; (* Test specific version with typo should not match *)
+        ("430.19",["430.18"]), false; (* Test specific version should not match *)
+        ("430.18",["430.19"]), false; (* Test specific version should not match *)
+        ("430.400",["430.200+"]), true; (* Higher version should match *)
+        ("430.100",["430.200+"]), false; (* Higher version should match *)
+        ("100.800",["430.200+"]), false; (* Lower version should not match *)
+        ("530.400",["430.200+";"typo_error"]), true; (* One wrong format config should not affact others *)
       ]
     end)
 
@@ -546,6 +508,7 @@ let test =
   "test_vgpu_type" >:::
   [
     "nvidia_read_whitelist" >::: NvidiaTest.ReadWhitelist.tests;
+    "nvidia_host_driver_support_multiple_vgpu" >::: NvidiaTest.HostDriverMultipleVgpuSupport.tests;
     "nvidia_print_nv_types" >:: NvidiaTest.print_nv_types;
     "intel_read_whitelist_line" >::: IntelTest.ReadWhitelistLine.tests;
     "intel_read_whitelist" >::: IntelTest.ReadWhitelist.tests;
--- a/ocaml/xapi/xapi_globs.ml
+++ b/ocaml/xapi/xapi_globs.ml
@@ -447,7 +447,9 @@ let vgt_fence_sz = "vgt_fence_sz"
 let mxgpu_vgpus_per_pgpu = "vgpus_per_pgpu"
 
 let nvidia_vgpu_first_slot_in_guest = 11
-let nvidia_compat_lookup_file = ref "/var/run/nonpersistent/xapi/nvidia_compat_lookup"
+let nvidia_host_driver_file = ref "/usr/lib64/libnvidia-vgpu.so"
+let nvidia_compat_conf_dir = "/usr/share/nvidia/vgx"
+let nvidia_compat_config_file_key = "config_file"
 
 let wlb_timeout = "wlb_timeout"
 let wlb_reports_timeout = "wlb_reports_timeout"
@@ -890,6 +892,9 @@ let sm_plugins = ref [ ]
 let accept_sm_plugin name =
   List.(fold_left (||) false (map (function `All -> true | `Sm x -> String.lowercase_ascii x = String.lowercase_ascii name) !sm_plugins))
 
+let nvidia_multi_vgpu_enabled_driver_versions = ref ["430.19"]
+let nvidia_default_host_driver_version = "0.0"
+
 let other_options = [
   gen_list_option "sm-plugins"
     "space-separated list of storage plugins to allow."
@@ -980,8 +985,9 @@ let other_options = [
   "db_idempotent_map", Arg.Set Db_globs.idempotent_map,
   (fun () -> string_of_bool !Db_globs.idempotent_map), "True if the add_to_<map> API calls should be idempotent";
 
-  "nvidia_compat_lookup_file", Arg.Set_string nvidia_compat_lookup_file,
-  (fun () -> !nvidia_compat_lookup_file), "Path to the file with NVidia vGPU compat data for use by xenopsd";
+  "nvidia_multi_vgpu_enabled_driver_versions", Arg.String (fun x -> nvidia_multi_vgpu_enabled_driver_versions := String.split ',' x),
+  (fun () -> String.concat "," !nvidia_multi_vgpu_enabled_driver_versions), "list of nvidia host driver versions with multiple vGPU supported.
+  if a version end with +, it means any driver version greater or equal than that version";
 ]
 
 let all_options = options_of_xapi_globs_spec @ other_options
--- a/ocaml/xapi/xapi_vgpu_type.ml
+++ b/ocaml/xapi/xapi_vgpu_type.ml
@@ -363,7 +363,18 @@ module Vendor_nvidia = struct
   let pt_when_vgpu = true
   let whitelist_file () = !Xapi_globs.nvidia_whitelist
   let device_id_of_conf conf = conf.identifier.Identifier.pdev_id
+  
+  let get_host_driver_version () = try Scanf.sscanf (Unix.readlink !Xapi_globs.nvidia_host_driver_file) "libnvidia-vgpu.so.%s" (fun x -> x)
+    with _ -> info "Can not get the file version of %s, use the default value" !Xapi_globs.nvidia_host_driver_file; Xapi_globs.nvidia_default_host_driver_version
 
+  let host_driver_supports_multi_vgpu ~host_driver_version ~supported_versions =
+    let version = float_of_string host_driver_version in
+    List.exists (fun pattern ->
+      match String.index_opt pattern '+' with
+      | Some i -> (try float_of_string (String.sub pattern 0 i) <= version with _ -> false)
+      | None -> try (float_of_string pattern) = version with _ -> false
+    ) supported_versions
+    
   (* A type to capture the XML tree with functions for use in Xmlm.input_doc_tree *)
   type tree = E of string * (string * string) list * tree list | D of string
   let el ((_, name), attr) children = E (name, List.map (fun ((_, k), v) -> k, v) attr, children)
@@ -432,6 +443,9 @@ module Vendor_nvidia = struct
       - Find the vgpuType elements that match the vgpu type ids found above,
         and construct vgpu_conf records.
     *)
+    let host_driver_version = get_host_driver_version () in
+    let driver_supports_multiple_vgpus = host_driver_supports_multi_vgpu ~host_driver_version
+      ~supported_versions:!Xapi_globs.nvidia_multi_vgpu_enabled_driver_versions in
     List.filter_map (fun vgpu_type ->
       let id = get_attr "id" vgpu_type in
       if List.mem_assoc id vgpu_ids then
@@ -456,12 +470,12 @@ module Vendor_nvidia = struct
            - Read  'multiVgpuSupported' from config, 1L indicates this type support multiple 
            - Currently always initialize 'compatible_types_on_pgpu' to itself only since we 
              don't support yet *)
-        let multi_vgpu_supported = Int64.of_string (get_data (find_one_by_name "multiVgpuSupported" vgpu_type)) in
+        let multi_vgpu_support = Int64.of_string (get_data (find_one_by_name "multiVgpuSupported" vgpu_type)) in
         let name = get_attr "name" vgpu_type in
-        info "Getting multiple vGPU supported from config file: %Ld, model: %s" multi_vgpu_supported name;
+        info "Getting multiple vGPU supported from config file: %Ld, model: %s" multi_vgpu_support name;
         let compatible_model_names_in_vm, compatible_model_names_on_pgpu =
-          match multi_vgpu_supported with
-          | 1L -> [name], [name]
+          match multi_vgpu_support, driver_supports_multiple_vgpus with
+          | 1L, true -> [name], [name]
           | _ -> [], [name]
         in
         Some {identifier; framebufferlength;
@@ -708,7 +722,7 @@ module Intel = Vendor(Vendor_intel)
 module AMD = Vendor(Vendor_amd)
 
 module Nvidia_compat = struct
-  let conf_dir = "/usr/share/nvidia/vgx"
+  let conf_dir = Xapi_globs.nvidia_compat_conf_dir
 
   let of_conf_file file_path =
     try
@@ -764,23 +778,26 @@ module Nvidia_compat = struct
     read_configs []
       (List.map (fun conf -> Filename.concat conf_dir conf) conf_files)
 
-  let create_compat_lookup_file __context =
+  let create_compat_config_file __context =
     try
       let open Db_filter_types in
-      read_config_dir conf_dir
-      |> List.map (fun (conf_file, identifier) ->
-        let identifier_string = Identifier.to_string identifier in
-        let expr = Eq (Field "identifier", Literal identifier_string) in
-        match Db.VGPU_type.get_internal_records_where ~__context ~expr with
-        | [_, rc] ->
-          let type_id = List.assoc Xapi_globs.vgpu_type_id rc.Db_actions.vGPU_type_internal_config in
-          Printf.sprintf "%s %s\n" conf_file type_id
-        | _ -> "" (* Type is not relevant: ignore *)
-      )
-      |> String.concat ""
-      |> Stdext.Unixext.write_string_to_file !Xapi_globs.nvidia_compat_lookup_file
+      let host_driver_version = Vendor_nvidia.get_host_driver_version () in
+      let host_driver_supports_multiple = Vendor_nvidia.host_driver_supports_multi_vgpu ~host_driver_version ~supported_versions:!Xapi_globs.nvidia_multi_vgpu_enabled_driver_versions in
+      if not host_driver_supports_multiple then
+        read_config_dir conf_dir (* Nvidia host driver does not support multiple vGPU, create compat config file *)
+        |> List.iter (fun (conf_file, identifier) ->
+            let identifier_string = Identifier.to_string identifier in
+            let expr = Eq (Field "identifier", Literal identifier_string) in
+            match Db.VGPU_type.get_internal_records_where ~__context ~expr with
+            | [vgpu_type_ref, rc] ->
+              let updated_config = rc.Db_actions.vGPU_type_internal_config
+                |> List.remove_assoc Xapi_globs.nvidia_compat_config_file_key
+                |> (fun x -> (Xapi_globs.nvidia_compat_config_file_key, conf_file)::x) in
+              Db.VGPU_type.set_internal_config ~__context ~self:vgpu_type_ref ~value:updated_config
+            | _ -> () (* Type is not relevant: ignore *)
+          )
     with e ->
-      error "Failed to create NVidia compat lookup file: %s\n%s\n"
+      error "Failed to create NVidia compat config_file: %s\n%s\n"
         (Printexc.to_string e) (Printexc.get_backtrace ())
 end
 
--- a/ocaml/xapi/xapi_xenops.ml
+++ b/ocaml/xapi/xapi_xenops.ml
@@ -663,16 +663,17 @@ module MD = struct
     let physical_pci_address = get_target_pci_address ~__context vgpu in
     let virtual_pci_address = get_virtual_pci_address ~__context vgpu in
     let vgpu_type = vgpu.Db_actions.vGPU_type in
-    let type_id = Db.VGPU_type.get_internal_config ~__context ~self:vgpu_type 
-                  |> List.assoc Xapi_globs.vgpu_type_id in
+    let type_id, config_file = Db.VGPU_type.get_internal_config ~__context ~self:vgpu_type
+                               |> (fun x -> (List.assoc_opt Xapi_globs.vgpu_type_id x),
+                                            (List.assoc_opt Xapi_globs.nvidia_compat_config_file_key x)) in
     let uuid = vgpu.Db_actions.vGPU_uuid in
     let extra_args = vgpu.Db_actions.vGPU_extra_args in
     let implementation =
       Nvidia {
         physical_pci_address = None; (* unused *)
-        config_file = None; (* unused *)
+        config_file;
         virtual_pci_address;
-        type_id  = Some type_id;
+        type_id;
         uuid = Some uuid;
         extra_args;
       }
--- a/ocaml/tests/test_xenopsd_metadata.ml
+++ b/ocaml/tests/test_xenopsd_metadata.ml
@@ -231,6 +231,7 @@ module GenerateVGPUMetadata =
           module State = Test_state.XapiDb
 
           let load_input __context (vm_config, pgpus_and_vgpu_types) =
+            Xapi_globs.nvidia_multi_vgpu_enabled_driver_versions := [Xapi_globs.nvidia_default_host_driver_version];
             let vm_ref = load_vm_config __context vm_config in
             List.iteri
               (fun index (pgpu, vgpu_type) ->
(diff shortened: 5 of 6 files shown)


ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -363,6 +363,30 @@ module Vendor_nvidia = struct
...
+  
+  let host_driver_version = try Scanf.sscanf (Unix.readlink !Xapi_globs.nvidia_host_driver_file) "libnvidia-vgpu.so.%s" (fun x -> x)
+    with _ -> info "can not get the file version of the %s, use the default value" !Xapi_globs.nvidia_host_driver_file; Xapi_globs.nvidia_default_host_driver_version
+
+  let is_host_driver_support_multi_vgpu () =
+    List.mem host_driver_version !Xapi_globs.nvidia_multi_vgpu_enabled_driver_versions ||
+    (
+      let from_version_on_mark = '+' in
+      let host_driver_major_version, host_driver_minor_version = try Scanf.sscanf host_driver_version "%d.%d" (fun x y -> x , y)
+        with _ -> info "get error when parse the host driver version: %s" host_driver_version;
```
lindig: Grammar: _got error error when parsing_; I'd say: `failed to parse nVIDIA host driver version %s"`

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -363,6 +363,30 @@ module Vendor_nvidia = struct
...
+          Scanf.sscanf default_host_driver_version "%d.%d" (fun x y -> x , y) in (* If cannot parse host driver version, we use the default one,
+                                                                                    which would be taken as not support multiple vGPU *)
+      !Xapi_globs.nvidia_multi_vgpu_enabled_driver_versions
```
lindig: Would it be easier to solve this using regular expressions? You can easily check the general format and extract the version numbers.

```
let rx' = Re.(compile @@ Pcre.re {|([0-9]+)\.([0-9]+)(\+?)|})
Re.Pcre.extract rx' "5.6+"
- : string array = [|"5.6+"; "5"; "6"; "+"|]
```

ocaml/xapi/xapi_xenops.ml
```diff
@@ -667,10 +667,19 @@ module MD = struct
                   |> List.assoc Xapi_globs.vgpu_type_id in
     let uuid = vgpu.Db_actions.vGPU_uuid in
     let extra_args = vgpu.Db_actions.vGPU_extra_args in
+    let config_file = match Xapi_vgpu_type.Vendor_nvidia.is_host_driver_support_multi_vgpu () with
+      | true -> None
+      | false -> (* Host driver does not support multiple vGPU, prepare config_file for the legency mode *)
+        let config_dir =  "/usr/share/nvidia/vgx" in
```
lindig: Should this be in `Xapi_globs`?

ocaml/xapi/xapi_vgpu_type.ml
liulinC: thanks, my utop does not have `Re` module, I will figure that how that works and change accordly.

ocaml/xapi/xapi_globs.ml
```diff
@@ -890,7 +891,7 @@ let sm_plugins = ref [ ]
 let accept_sm_plugin name =
   List.(fold_left (||) false (map (function `All -> true | `Sm x -> String.lowercase_ascii x = String.lowercase_ascii name) !sm_plugins))
 
-let nvidia_multi_vgpu_enabled_driver_versions = ref ["430.19"]
+let nvidia_multi_vgpu_enabled_driver_versions = ref (["430.19"])
```
robhoes: The `()` are not needed.

ocaml/xapi/xapi_vgpu_type.ml
liulinC: updated

ocaml/xapi/xapi_xenops.ml
liulinC: good point

ocaml/xapi/xapi_xenops.ml
```diff
@@ -667,10 +667,19 @@ module MD = struct
...
+        let model_name_in_config_file = Db.VGPU_type.get_model_name ~__context ~self:vgpu_type (* like "GRID M10-8Q" *)
+                                        |> String.split_on_char ' ' |> String.concat "_" |> String.lowercase_ascii in
+        let path = Printf.sprintf "%s/%s.conf" config_dir model_name_in_config_file in
```
robhoes: Is this guaranteed to work for all models? A safer alternative is to repurpose `Nvidia_compat.create_compat_lookup_file` in xapi_vgpu_type.ml to add the `config_file` path into the `internal_config` of the types, so that we can just get it here like the `type_id` above. We could add the path to `internal_config` only when then driver check fails, so that we only have to do that check once.

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -365,27 +365,27 @@ module Vendor_nvidia = struct
...
-                           with _ -> None)
-      |> List.exists ( fun (major_version, minor_version) ->
-          match host_driver_major_version - major_version, host_driver_minor_version - minor_version with
-          | _ as x, _ when x > 0 -> true
-          | 0, (_ as y) when y >= 0 -> true
-          | _ -> false
-        )
+      |> List.map (fun x -> Re.Pcre.extract valid_version_pattern x)
+      |> List.filter (fun x -> (Array.length x) = expected_length)
+      |> List.exists ( fun x -> let major_version = Array.get x 1 |> int_of_string in
```
lindig: This does not make use of the optional `+` in the version string.

ocaml/xapi/xapi_xenops.ml
liulinC: how to  identify which `config_file` belong to which type, it may still need to use the method above.
It is a good point to store it in xapi db and only compute once is a good point

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -365,27 +365,27 @@ module Vendor_nvidia = struct
...
-      |> List.filter (fun x -> String.get x (String.length x -1) = from_version_on_mark)(* Ending with "+" *)
-      |> List.filter_map ( fun x -> try Scanf.sscanf x "%d.%d" (fun x y -> Some (x , y))
-                           with _ -> None)
-      |> List.exists ( fun (major_version, minor_version) ->
-          match host_driver_major_version - major_version, host_driver_minor_version - minor_version with
-          | _ as x, _ when x > 0 -> true
-          | 0, (_ as y) when y >= 0 -> true
-          | _ -> false
-        )
+      |> List.map (fun x -> Re.Pcre.extract valid_version_pattern x)
```
lindig: The `extract` can fail with `Not_found` when the string does not match the regular expression - this exception needs to be handled.

ocaml/xapi/xapi_xenops.ml
robhoes: That is already handled in `Nvidia_compat`. You just need to change the bit in `create_compat_lookup_file` that writes to the file, to instead updated the DB.

liulinC: okay, I will check it

ocaml/xapi/xapi_vgpu_type.ml
lindig: Consider:
```
let nvidia_multi_vgpu_enabled_driver_versions = ref ["430.19"]
	
```
This version will not pass through this filter list and no deriver will be found compatible.

liulinC: Yes, I see and I have  already did a change for that, please see my latest commit.

TalonsLee:
Do we need this?

1)
major_version.minor_version

both versions match.

2)
major_version.minor_version+

same major version, minor version larger than this

3)
major_version+. minor_version+

major version larger than this,
or (same major version and minor larger than this)

lindig:
I think it would be good to have a function that just takes the current driver version and the list of supported driver version and returns true/false if the current driver version is sufficient. This can be easily tested and does not mix discovering driver versions. I'm not confident that the current solution is correct.

robhoes:
...and a unit test for the version checking function that @lindig suggested.

liulinC:
and  @lindig and @robhoes , will split that into a single function and add a UT for it.

liulinC:
thanks @lindig and @robhoes  for the comment, will split that into a single function and add UT for it.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -667,10 +667,16 @@ module MD = struct
                   |> List.assoc Xapi_globs.vgpu_type_id in
     let uuid = vgpu.Db_actions.vGPU_uuid in
     let extra_args = vgpu.Db_actions.vGPU_extra_args in
+    let config_file = match Xapi_vgpu_type.Vendor_nvidia.is_host_driver_support_multi_vgpu () with
```
robhoes: We can avoid the call to this function here (where it is called very often), and move it into `create_compat_config_file` (which is called just once, when xapi starts). I.e. only add the `config_file` key to the `internal_config` map if needed.

Then over here you just do something like
```
let config_file =
  if List.mem_assoc Xapi_globs.nvidia_compat_config_file_key internal_config then
    Some (List.assoc Xapi_globs.nvidia_compat_config_file_key internal_config)
  else
    None
in
```

liulinC: oh, I see.

robhoes: Oh, there is now a `List.assoc_opt` function in the standard lib. Then it becomes just:
```
let config_file = List.assoc_opt Xapi_globs.nvidia_compat_config_file_key internal_config in
```

lindig:
Can you confirm that the unit test for the version check is working?

liulinC:
@lindig I think so, as  I have following logs in the UT output
```
3604 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: Start test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:0:(430.19, 430.19) -> true.
3605 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: Test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:0:(430.19, 430.19) -> true is successful.
3606 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: End test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:0:(430.19, 430.19) -> true.
3607 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: Start test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:1:(430.19, 430.89,430.19) -> true.
3608 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: Test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:1:(430.19, 430.89,430.19) -> true is successful.
3609 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: End test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:1:(430.19, 430.89,430.19) -> true.
3610 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: Start test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:2:(430.19, 430.19,typo_error) -> true.
3611 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: Test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:2:(430.19, 430.19,typo_error) -> true is successful.
3612 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: End test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:2:(430.19, 430.19,typo_error) -> true.
3613 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: Start test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:3:(430.19, 430.18) -> false.
3614 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: Test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:3:(430.19, 430.18) -> false is successful. 
3615 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: End test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:3:(430.19, 430.18) -> false.
3616 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: Start test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:4:(430.18, 430.19) -> false.
3617 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: Test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:4:(430.18, 430.19) -> false is successful. 
3618 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: End test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:4:(430.18, 430.19) -> false.
3619 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: Start test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:5:(430.400, 430.200+) -> true.
3620 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: Test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:5:(430.400, 430.200+) -> true is successful.
3621 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: End test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:5:(430.400, 430.200+) -> true.
3622 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: Start test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:6:(430.100, 430.200+) -> false.
3623 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: Test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:6:(430.100, 430.200+) -> false is successful.
3624 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: End test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:6:(430.100, 430.200+) -> false.
3625 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: Start test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:7:(530.400, 430+.200) -> true.
3626 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: Test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:7:(530.400, 430+.200) -> true is successful.
3627 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: End test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:7:(530.400, 430+.200) -> true.
3628 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: Start test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:8:(100.400, 430+.200) -> false.
3629 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: Test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:8:(100.400, 430+.200) -> false is successful.
3630 2019-07-03T08:47:47+00:00 007dfd27ae70#00 I: End test base_suite:9:test_vgpu_type:1:nvidia_host_driver_support_multiple_vgpu:8:(100.400, 430+.200) -> false.
```

ocaml/xapi/xapi_xenops.ml
```diff
@@ -667,10 +667,12 @@ module MD = struct
                   |> List.assoc Xapi_globs.vgpu_type_id in
     let uuid = vgpu.Db_actions.vGPU_uuid in
     let extra_args = vgpu.Db_actions.vGPU_extra_args in
+    let config_file = Db.VGPU_type.get_internal_config ~__context ~self:vgpu_type
```
robhoes: Let's call `Db.VGPU_type.get_internal_config` just once (it's a slave->master call each time). It's already done in L666.

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -764,23 +793,28 @@ module Nvidia_compat = struct
...
-        | [_, rc] ->
-          let type_id = List.assoc Xapi_globs.vgpu_type_id rc.Db_actions.vGPU_type_internal_config in
-          Printf.sprintf "%s %s\n" conf_file type_id
-        | _ -> "" (* Type is not relevant: ignore *)
-      )
-      |> String.concat ""
-      |> Stdext.Unixext.write_string_to_file !Xapi_globs.nvidia_compat_lookup_file
+      let host_driver_version = Vendor_nvidia.get_host_driver_version () in
+      match Vendor_nvidia.is_host_driver_support_multi_vgpu ~host_driver_version ~supported_versions:!Xapi_globs.nvidia_multi_vgpu_enabled_driver_versions with
+    | false -> read_config_dir conf_dir (* Nvidia host driver does not support multiple vGPU, create compat config file *)
```
robhoes: The indentation looks weird here.

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -363,6 +363,33 @@ module Vendor_nvidia = struct
...
+      |> List.filter (fun x -> Array.get x 2 = "+" || Array.get x 4 = "+") (* Filter all version that from them on *)
+      |> List.exists ( fun x -> let major_version = Array.get x 1 |> int_of_string in
+                       let from_major_on = Array.get x 2 |> (fun e -> e = "+" ) in
+                       let minor_version = Array.get x 3 |> int_of_string in
+                       let from_minor_on = Array.get x 4 |> (fun e -> e = "+" ) in
+                       debug "host_driver_major_version: %d, host_driver_minor_version: %d, major_version: %d, from_major_on: %s , minor_version: %d, from_minor_on: %s"
+                       host_driver_major_version host_driver_minor_version major_version (string_of_bool from_major_on) minor_version (string_of_bool from_minor_on);
+                       match host_driver_major_version - major_version, host_driver_minor_version - minor_version with
+                       (* For example, the host version is 430.19 *)
+                       | _ as x, _ when (x > 0 && from_major_on) -> true (* 420+,34, in case of master release *)
```
lindig: This can be simplified:
```
| x, y when .. ->
| 0, y when .. ->
```

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -363,6 +363,33 @@ module Vendor_nvidia = struct
...
+    (
+      let valid_version_pattern = Re.(compile @@ Pcre.re {|^([0-9]+)(\+?)\.([0-9]+)(\+?)$|}) in
+      (* 0. total string, 1. major version, 2. + or "" 3. minor version. 4. + or "" *)
+      let host_driver_major_version, host_driver_minor_version =
+        try Re.Pcre.extract valid_version_pattern host_driver_version |> (fun x -> Array.get x 1 |> int_of_string, Array.get x 3 |> int_of_string)
+        with Not_found -> raise (Api_errors.Server_error (Api_errors.internal_error,[Printf.sprintf "Failed to parse host driver version: %s" host_driver_version])) in
+      List.filter_map (fun x -> try Some (Re.Pcre.extract valid_version_pattern x) with _ -> None) supported_versions
+      |> List.filter (fun x -> Array.get x 2 = "+" || Array.get x 4 = "+") (* Filter all version that from them on *)
+      |> List.exists ( fun x -> let major_version = Array.get x 1 |> int_of_string in
+                       let from_major_on = Array.get x 2 |> (fun e -> e = "+" ) in
```
lindig: This could be simplified - but would not mind if it stays.
```
Array.get x 2 = '+'
```

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -764,23 +793,28 @@ module Nvidia_compat = struct
...
+      let host_driver_version = Vendor_nvidia.get_host_driver_version () in
+      match Vendor_nvidia.is_host_driver_support_multi_vgpu ~host_driver_version ~supported_versions:!Xapi_globs.nvidia_multi_vgpu_enabled_driver_versions with
+    | false -> read_config_dir conf_dir (* Nvidia host driver does not support multiple vGPU, create compat config file *)
+      |> List.iter (fun (conf_file, identifier) ->
+          let identifier_string = Identifier.to_string identifier in
+          let expr = Eq (Field "identifier", Literal identifier_string) in
+          match Db.VGPU_type.get_internal_records_where ~__context ~expr with
+          | [vgpu_type_ref, rc] ->
+            let internal_config = rc.Db_actions.vGPU_type_internal_config in
+            let updated_config = try match (List.assoc Xapi_globs.nvidia_compat_config_file_key internal_config) with
```
robhoes: Rather than having the exception handling, you could use `List.assoc_opt`:
```
let updated_config =
  match List.assoc_opt Xapi_globs.nvidia_compat_config_file_key internal_config with
  | Some c when c = config_file ->
  | Some c                      ->
  | None                        ->
in
```
(you don't actually need the first case).

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -363,6 +363,33 @@ module Vendor_nvidia = struct
...
+  let is_host_driver_support_multi_vgpu ~host_driver_version ~supported_versions =
+    List.mem host_driver_version supported_versions ||
+    (
+      let valid_version_pattern = Re.(compile @@ Pcre.re {|^([0-9]+)(\+?)\.([0-9]+)(\+?)$|}) in
+      (* 0. total string, 1. major version, 2. + or "" 3. minor version. 4. + or "" *)
+      let host_driver_major_version, host_driver_minor_version =
+        try Re.Pcre.extract valid_version_pattern host_driver_version |> (fun x -> Array.get x 1 |> int_of_string, Array.get x 3 |> int_of_string)
+        with Not_found -> raise (Api_errors.Server_error (Api_errors.internal_error,[Printf.sprintf "Failed to parse host driver version: %s" host_driver_version])) in
+      List.filter_map (fun x -> try Some (Re.Pcre.extract valid_version_pattern x) with _ -> None) supported_versions
+      |> List.filter (fun x -> Array.get x 2 = "+" || Array.get x 4 = "+") (* Filter all version that from them on *)
```
lindig: This `filter` can only be passed by a version that has a `+` in it. Why does this pass the unit tests that have no `+`?

```
        ("430.19",["430.89";"430.19"]), true; (* Test specific version should match *)
```

ocaml/tests/test_vgpu_type.ml
```diff
@@ -202,6 +137,34 @@ module NvidiaTest = struct
...
+        ("430.19",["430.89";"430.19"]), true; (* Test specific version should match *)
+        ("430.19",["430.19";"typo_error"]), true; (* Test specific version should match *)
+        ("430.19",["430.18"]), false; (* Test specific version should not match *)
+        ("430.18",["430.19"]), false; (* Test specific version should not match *)
+        ("430.400",["430.200+"]), true; (* LCM release, higher version should match *)
+        ("430.100",["430.200+"]), false; (* LCM release, lower version should not match *)
+        ("530.400",["430+.200"]), true; (* master release, higher version should match *)
+        ("100.400",["430+.200"]), false; (* master release, lower version should not match *)
+        ("530.400",["430+.200";"typo_error"]), true; (* LCM release*)
+        ("530.400",["430+.200+"]), true; (* LCM release*)
```
robhoes: How about `("100.100",["430.200+"]), false`?

liulinC: I will add this one

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -363,6 +363,33 @@ module Vendor_nvidia = struct
   let pt_when_vgpu = true
   let whitelist_file () = !Xapi_globs.nvidia_whitelist
   let device_id_of_conf conf = conf.identifier.Identifier.pdev_id
+  
+  let get_host_driver_version () = try Scanf.sscanf (Unix.readlink !Xapi_globs.nvidia_host_driver_file) "libnvidia-vgpu.so.%s" (fun x -> x)
+    with _ -> info "Can not get the file version of %s, use the default value" !Xapi_globs.nvidia_host_driver_file; Xapi_globs.nvidia_default_host_driver_version
+
+  let is_host_driver_support_multi_vgpu ~host_driver_version ~supported_versions =
```
robhoes: This function is so complicated that it is impossible to review... :( Does it have to have all these cases with `+` for major and minor separately? Without that, and just a + for the whole thing, you could just do something like this?
```
let is_host_driver_support_multi_vgpu ~host_driver_version ~supported_versions =
  let version = float_of_string host_driver_version in
    List.fold_left (fun result pattern ->
      match List.index_opt pattern '+' with
      | Some i -> result || (float_of_string (String.sub pattern i) <= version)
      | None   -> result || (float_of_string pattern = version)
    ) false supported_versions
```

lindig: I tend to agree. Do we really need the flexibility the mechanism implements? Is there a reason to believe that any higher version than a minimum is not usable? I can imagine a number of simpler approaches:

* Have a minimum version and accept all versions greater than that. We can always add a blacklist of versions that we would want to exclude.
* Use a list of regular expressions. Any version that matches at least one of the regular expressions is accepted.

lindig:
```

Comment (liulinC) [reply ocaml/xapi/xapi_vgpu_type.ml]: As I know, `430.27` does not support multiple vGPU.

File: ocaml/xapi/xapi_vgpu_type.ml (hunk 14 above)
Comment (liulinC) [reply ocaml/xapi/xapi_vgpu_type.ml]: see ` List.mem host_driver_version supported_versions`, but never mind, I will change the function as Rob suggested.

**Comment by liulinC** [comment]
It seems there will be more discussion with Nvidia about how to determine whether a host driver support multiple vGPU, should we close/pause this PR ATM?

File: ocaml/xapi/xapi_vgpu_type.ml (hunk 16 above)
Comment (lindig) [reply ocaml/xapi/xapi_vgpu_type.ml]: It would be easy to exclude `430.27` by creating a match to `false` as the 1st or 2nd entry. That is the point of this: it's easy because it uses to full power of OCaml. The disadvantage is that it cannot be controlled from the command line - but even that would not be difficult.

Comment (liulinC) [reply ocaml/xapi/xapi_vgpu_type.ml]: Thanks Lindig for your advice.  This is still under discussing, but Rob's suggestion looks good so far. 
let's wait for the final decision, and perhaps we can just take Rob's advice if possible.

Comment (liulinC) [reply ocaml/xapi/xapi_vgpu_type.ml]: Thanks @robhoes  and @lindig  for the discussion about this PR.
As @robhoes suggestion looks good so far so good, and the logic of how to determine whether a host  driver is encapsulated into a single function (which is easy to change if necessary), can I suggest we just take Rob's suggestion in this PR and continute the review. 

I changed the `List.fold_left` to `List.exist` as we does not need to iterate all the elements if there is one elements return true.

File: ocaml/xapi/xapi_vgpu_type.ml [hunk 17]
```diff
@@ -363,7 +363,18 @@ module Vendor_nvidia = struct
   let pt_when_vgpu = true
   let whitelist_file () = !Xapi_globs.nvidia_whitelist
   let device_id_of_conf conf = conf.identifier.Identifier.pdev_id
-
+  
+  let get_host_driver_version () = try Scanf.sscanf (Unix.readlink !Xapi_globs.nvidia_host_driver_file) "libnvidia-vgpu.so.%s" (fun x -> x)
+    with _ -> info "Can not get the file version of %s, use the default value" !Xapi_globs.nvidia_host_driver_file; Xapi_globs.nvidia_default_host_driver_version
+
+  let is_host_driver_support_multi_vgpu ~host_driver_version ~supported_versions =
```
lindig: `is_host_driver_support_multi_vgpu` is grammatically awkward. How about `host_driver_supports_multi_vgpu`?
Avoid overlong lines in new code.

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -432,6 +443,8 @@ module Vendor_nvidia = struct
       - Find the vgpuType elements that match the vgpu type ids found above,
         and construct vgpu_conf records.
     *)
+    let host_driver_version = get_host_driver_version () in
+    let driver_support_multiple_vgpu = is_host_driver_support_multi_vgpu ~host_driver_version ~supported_versions:!Xapi_globs.nvidia_multi_vgpu_enabled_driver_versions in
```
lindig: Again, grammatically awkward. 
* `driver_supports_multiple_vgpus` 
* `multi_vgpu_support`
Avoid overlong lines in new code.

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -764,23 +777,26 @@ module Nvidia_compat = struct
...
-        | _ -> "" (* Type is not relevant: ignore *)
-      )
-      |> String.concat ""
-      |> Stdext.Unixext.write_string_to_file !Xapi_globs.nvidia_compat_lookup_file
+      let host_driver_version = Vendor_nvidia.get_host_driver_version () in
+      let host_driver_support_multiple = Vendor_nvidia.is_host_driver_support_multi_vgpu ~host_driver_version ~supported_versions:!Xapi_globs.nvidia_multi_vgpu_enabled_driver_versions in
+      if not host_driver_support_multiple then
+        read_config_dir conf_dir (* Nvidia host driver does not support multiple vGPU, create compat config file *)
+        |> List.iter (fun (conf_file, identifier) ->
+            let identifier_string = Identifier.to_string identifier in
```
lindig: Maybe just `id`

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -764,23 +777,26 @@ module Nvidia_compat = struct
...
+            let expr = Eq (Field "identifier", Literal identifier_string) in
+            match Db.VGPU_type.get_internal_records_where ~__context ~expr with
+            | [vgpu_type_ref, rc] ->
+              let updated_config = rc.Db_actions.vGPU_type_internal_config
+                |> List.remove_assoc Xapi_globs.nvidia_compat_config_file_key (* If the key does not exist, just return the original list *)
```
lindig: Avoid overlong lines in new code.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -663,16 +663,16 @@ module MD = struct
     let physical_pci_address = get_target_pci_address ~__context vgpu in
     let virtual_pci_address = get_virtual_pci_address ~__context vgpu in
     let vgpu_type = vgpu.Db_actions.vGPU_type in
-    let type_id = Db.VGPU_type.get_internal_config ~__context ~self:vgpu_type 
-                  |> List.assoc Xapi_globs.vgpu_type_id in
+    let type_id, config_file = Db.VGPU_type.get_internal_config ~__context ~self:vgpu_type
+                               |> (fun x -> (List.assoc_opt Xapi_globs.vgpu_type_id x), (List.assoc_opt Xapi_globs.nvidia_compat_config_file_key x)) in
```
lindig: Avoid overlong lines in new code.

ocaml/xapi/xapi_vgpu_type.ml
liulinC: `multi_vgpu_supported` is actually from Nvidia config file `multiVgpuSupported`, I will change the others.

ocaml/xapi/xapi_vgpu_type.ml
liulinC: all other parts use `identifier_string` instead of `id`,  like 
```145 let find_and_update ~__context vgpu_type =
  146   let identifier_string = Identifier.to_string vgpu_type.identifier in
```
maybe we shall keep consistency.

liulinC:
Merge this commit as Nvidia confirmed they would not backport their multiple vGPU feature to the old Versions, which would make a new lower version support multiple vGPU while the bigger version not.
So rob's suggestion looks good.

PR #3902, opened 2019-07-01
Remove PVS Proxy from a VM when it is templated

rwdrich:
From CA-322682, In a PVS Environment, if a VM has a proxy enabled for it, and the vm is then converted into a template, the proxy isn't removed. This then means that the template has an existing vif, and as such no further streamed pvs vms can be created from the template as the device already exists.

I've tested this on an environment that I had left over from the MRGs, can will probably add a few screenshots of this onto CA-322682


robhoes:
Looks good! Could you please add the CA number, and the first paragraph of the PR text to the commit message?

PR #3905, opened 2019-07-04
CP-31729

lippirk:
@robhoes i don't really want to merge yet, i just have a few questions

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -1310,6 +1310,7 @@ module Bond = struct
           field ~lifecycle:[Published, rel_boston, ""] ~qualifier:DynamicRO ~default_value:(Some (VEnum "balance-slb")) ~ty:mode "mode" "The algorithm used to distribute traffic among the bonded NICs";
           field ~in_oss_since:None ~in_product_since:rel_tampa ~qualifier:DynamicRO ~ty:(Map(String, String)) ~default_value:(Some (VMap [])) "properties" "Additional configuration properties specific to the bond mode.";
           field ~in_oss_since:None ~in_product_since:rel_tampa ~qualifier:DynamicRO ~ty:Int ~default_value:(Some (VInt 0L)) "links_up" "Number of links up in this bond";
+          field ~lifecycle:[Published, rel_quebec, ""] ~qualifier:DynamicRO ~ty:Bool ~default_value:(Some (VBool true)) "auto_update_mac" "true if the MAC was taken from the primary slave when the bond was created, and false if the client specified the MAC"
         ]
       ()
 end
--- a/ocaml/idl/datamodel_common.ml
+++ b/ocaml/idl/datamodel_common.ml
@@ -8,7 +8,7 @@ open Datamodel_roles
               When introducing a new release, bump the schema minor version to the next hundred
               to leave a gap for potential hotfixes needing to increment the schema version.*)
 let schema_major_vsn = 5
-let schema_minor_vsn = 501
+let schema_minor_vsn = 502
 
 (* Historical schema versions just in case this is useful later *)
 let rio_schema_major_vsn = 5
--- a/ocaml/tests/test_common.ml
+++ b/ocaml/tests/test_common.ml
@@ -190,8 +190,8 @@ let make_vlan ~__context ~tagged_PIF ~untagged_PIF ~tag ?(other_config=[]) () =
 
 let make_network_sriov = Xapi_network_sriov.create_internal
 
-let make_bond ~__context ?(ref=Ref.make ()) ?(uuid=make_uuid ()) ~master ?(other_config=[]) ?(primary_slave=Ref.null) ?(mode=`activebackup) ?(properties=[]) () =
-  Db.Bond.create ~__context ~ref ~uuid ~master ~other_config ~primary_slave ~mode ~properties ~links_up:0L ~auto_update_mac:false;
+let make_bond ~__context ?(ref=Ref.make ()) ?(uuid=make_uuid ()) ~master ?(other_config=[]) ?(primary_slave=Ref.null) ?(mode=`activebackup) ?(properties=[]) ?(auto_update_mac=false) () =
+  Db.Bond.create ~__context ~ref ~uuid ~master ~other_config ~primary_slave ~mode ~properties ~links_up:0L ~auto_update_mac;
   ref
 
 let make_tunnel = Xapi_tunnel.create_internal
--- a/ocaml/tests/test_common.ml
+++ b/ocaml/tests/test_common.ml
@@ -191,7 +191,7 @@ let make_vlan ~__context ~tagged_PIF ~untagged_PIF ~tag ?(other_config=[]) () =
 let make_network_sriov = Xapi_network_sriov.create_internal
 
 let make_bond ~__context ?(ref=Ref.make ()) ?(uuid=make_uuid ()) ~master ?(other_config=[]) ?(primary_slave=Ref.null) ?(mode=`activebackup) ?(properties=[]) () =
-  Db.Bond.create ~__context ~ref ~uuid ~master ~other_config ~primary_slave ~mode ~properties ~links_up:0L;
+  Db.Bond.create ~__context ~ref ~uuid ~master ~other_config ~primary_slave ~mode ~properties ~links_up:0L ~auto_update_mac:false;
   ref
 
 let make_tunnel = Xapi_tunnel.create_internal
--- a/ocaml/tests/test_nm.ml
+++ b/ocaml/tests/test_nm.ml
@@ -1,0 +1,66 @@
+open Test_highlevel
+open Test_common
+open OUnit
+
+module MaybeUpdateMasterPifMac = Generic.Make(Generic.EncapsulateState(struct
+  module Io = struct
+    type input_t = (bool * string option) (* auto_update_mac, new_primary_slave_mac *)
+
+    type output_t = (string * string) (* record bond master mac, db bond master mac *)
+
+    let show_primary_slave_mac_change = function
+      | None -> "primary_slave mac unchanged"
+      | Some new_primary_slave_mac -> Printf.sprintf "primary_slave mac -> %s" new_primary_slave_mac
+
+    let string_of_input_t = Test_printers.pair Test_printers.bool show_primary_slave_mac_change
+    let string_of_output_t (rc,db) = Printf.sprintf "Bond master macs. record: %s db : %s" rc db
+  end
+
+  module State = Test_state.XapiDb
+
+  let original_primary_slave_mac = "aa:11:11:11:11:11"
+  let original_master_mac = "cc:11:11:11:11:11"
+  let a_new_mac = "12:34:56:78:90:ab"
+
+  let load_input (__context : State.state_t) (auto_update_mac, o_new_primary_slave_mac) =
+    let network = Test_common.make_network ~__context () in
+    let host = Test_common.make_host ~__context () in
+    let bond_master = Test_common.make_pif ~__context ~network ~host () ~mAC:original_master_mac in
+    let primary_slave = Test_common.make_pif ~__context ~network ~host ~bond_slave_of:bond_master () ~mAC:"aa:11:11:11:11:11" in
+    let _ = Test_common.make_pif ~__context ~network ~host ~bond_slave_of:bond_master () ~mAC:"bb:11:11:11:11:11" (* lowly slave *) in
+    let bond = Test_common.make_bond ~__context ~master:bond_master ~primary_slave () in
+    (Db.Bond.set_auto_update_mac ~__context ~self:bond ~value:auto_update_mac;
+     match o_new_primary_slave_mac with
+      | None -> ()
+      | Some new_primary_slave_mac -> Db.PIF.set_MAC ~__context ~self:primary_slave ~value:new_primary_slave_mac)
+
+  let extract_output (__context : State.state_t) _ =
+    let bond = List.hd (Db.Bond.get_all ~__context) in
+    let bond_record = Db.Bond.get_record ~__context ~self:bond in
+    let bond_master = bond_record.API.bond_master in
+    let bond_master_record = Db.PIF.get_record ~__context ~self:bond_master in
+    let bond_master_record = (Nm.maybe_update_master_pif_mac ~__context bond_record bond_master_record bond_master) in
+
+    (* these mac addresses should be equal; we test them both *)
+    (bond_master_record.API.pIF_MAC, Db.PIF.get_MAC ~__context ~self:bond_master)
+
+  let tests = [
+        (* primary slave mac does not change & auto_update_mac disabled *)
+        ((false, None), (original_master_mac, original_master_mac));
+
+        (* primary slave mac does not change & auto_update_mac enabled *)
+        ((true, None), (original_primary_slave_mac, original_primary_slave_mac));
+
+        (* primary slave mac does change & auto_update_mac disabled *)
+        ((false, Some a_new_mac), (original_master_mac, original_master_mac));
+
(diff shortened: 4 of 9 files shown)


ocaml/idl/datamodel_common.ml
```diff
@@ -8,7 +8,7 @@ open Datamodel_roles
               When introducing a new release, bump the schema minor version to the next hundred
               to leave a gap for potential hotfixes needing to increment the schema version.*)
 let schema_major_vsn = 5
-let schema_minor_vsn = 501
+let schema_minor_vsn = 502
```
lippirk: @psafont said not to bother with bumping these versions? ( however the 'add a new record' guide says to )

ocaml/xapi/nm.ml
```diff
@@ -322,54 +322,68 @@ let bring_pif_up ~__context ?(management_interface=false) (pif: API.ref_PIF) =
...
-            let slaves = Db.Bond.get_slaves ~__context ~self:bond in
+            let dbBond = Db.Bond.get_record ~__context ~self:bond in
+
+            if dbBond.API.bond_auto_update_mac then begin
+              (* Make sure the MAC address of the PIF Bond master matches the MAC address of the Bond's primary slave *)
+              let primarySlaveRef = dbBond.API.bond_primary_slave in
+              let primarySlaveMAC = Db.PIF.get_MAC ~__context ~self:primarySlaveRef in
+              if !rc.pIF_MAC <> primarySlaveMAC then begin
+                Db.PIF.set_MAC ~__context ~self:pif ~value:primarySlaveMAC;
+                rc := Db.PIF.get_record ~__context ~self:pif;
```
lippirk: the reason i made `rc` a ref was so that we can conditionally make this `get_record` call. would you say the cost of introducing ref is worth this small benefit?

ocaml/idl/datamodel.ml
```diff
@@ -1310,6 +1310,7 @@ module Bond = struct
           field ~lifecycle:[Published, rel_boston, ""] ~qualifier:DynamicRO ~default_value:(Some (VEnum "balance-slb")) ~ty:mode "mode" "The algorithm used to distribute traffic among the bonded NICs";
           field ~in_oss_since:None ~in_product_since:rel_tampa ~qualifier:DynamicRO ~ty:(Map(String, String)) ~default_value:(Some (VMap [])) "properties" "Additional configuration properties specific to the bond mode.";
           field ~in_oss_since:None ~in_product_since:rel_tampa ~qualifier:DynamicRO ~ty:Int ~default_value:(Some (VInt 0L)) "links_up" "Number of links up in this bond";
+          field ~lifecycle:[Published, rel_naples, ""] ~in_oss_since:None ~qualifier:DynamicRO ~ty:Bool ~default_value:(Some (VBool true)) "auto_update_mac" "true if the MAC was taken from the primary slave when the bond was created, and false if the client specified the MAC"
```
lippirk: oss = Open source?

i wasn't sure about the `lifecycle` field either?

ocaml/xapi/nm.ml
psafont: I think that adding a new control variable would be better than changing rc to be a reference, as the latter introduces quite a few avoidable code changes.

For example:
```
let mac_changed = ref false in

[...]

if !rc.pIF_MAC <> primarySlaveMAC then
  mac_changed := true;

[...]

let rc = if !mac_changed then [...] else rc in
```

This keeps the changes localised and it's very easy to understand what happens (change the name of the variable or add a comment if it's not clear)

ocaml/idl/datamodel_common.ml
psafont: I mispoke, the comment above the field is pretty clear :)

ocaml/xapi/xapi_bond.ml
```diff
@@ -385,7 +386,7 @@ let create ~__context ~network ~members ~mAC ~mode ~properties =
         ~ipv6_configuration_mode:`None ~iPv6:[""] ~ipv6_gateway:"" ~primary_address_type:`IPv4 ~managed:true
         ~properties:pif_properties ~capabilities:[] ~pCI:Ref.null;
       Db.Bond.create ~__context ~ref:bond ~uuid:(Uuid.to_string (Uuid.make_uuid ())) ~master:master ~other_config:[]
-        ~primary_slave ~mode ~properties ~links_up:0L;
+        ~primary_slave ~mode ~properties ~links_up:0L ~auto_update_mac:auto_update_mac;
```
psafont: This can be shortened to simply `~auto_update_mac` because both the parameter and the variable have the same name

ocaml/idl/datamodel.ml
robhoes: `in_oss_since`: you can just leave that out (it's not actually used anymore).

`lifecycle` defines when the new field was added (in which release). That would be Quebec (as things stand now).

ocaml/idl/datamodel_common.ml
```diff
@@ -90,7 +90,7 @@ let lima_release_schema_major_vsn = 5
 let lima_release_schema_minor_vsn = 203
 
 let naples_release_schema_major_vsn = 5
-let naples_release_schema_minor_vsn = 400
+let naples_release_schema_minor_vsn = 401
```
robhoes: Don't change this: it is just documenting the schema version of the (already released) Naples release.

ocaml/xapi/xapi_bond.ml
```diff
@@ -277,7 +277,8 @@ let create ~__context ~network ~members ~mAC ~mode ~properties =
 
   (* Validate MAC parameter; note an empty string is OK here, since that means 'inherit MAC from
      	 * primary slave PIF' (see below) *)
-  if mAC <> "" && (not (Helpers.is_valid_MAC mAC)) then
+  let didClientSpecifyMAC = mAC <> "" in
```
robhoes: Stylistic point: we don't use CamelCase, but `x_y_z`.

ocaml/xapi/nm.ml
```diff
@@ -333,10 +334,25 @@ let bring_pif_up ~__context ?(management_interface=false) (pif: API.ref_PIF) =
...
+                mac_changed := true;
+              end;
+            end;
+
+            let slaves = dbBond.API.bond_slaves in
+
             List.iter (fun self -> Db.PIF.set_currently_attached ~__context ~self ~value:false) slaves
         end;
 
+        let rc = if !mac_changed then Db.PIF.get_record ~__context ~self:pif else rc in
```
robhoes: An alternative to eliminate the `get_record` call and the `ref` is to "return" an updated PIF record from the above `match` statement. A structure like:
```
let rc =
  match rc.API.pIF_bond_master_of with
  | [] -> rc
  | bond :: _ ->
    ...
    List.iter ...slaves;
    if ... then
      {rc with mAC = primarySlaveMAC}
    else
      rc
in
```

ocaml/xapi/xapi_bond.ml
```diff
@@ -277,8 +277,8 @@ let create ~__context ~network ~members ~mAC ~mode ~properties =
 
   (* Validate MAC parameter; note an empty string is OK here, since that means 'inherit MAC from
      	 * primary slave PIF' (see below) *)
-  let didClientSpecifyMAC = mAC <> "" in
-  if didClientSpecifyMAC && (not (Helpers.is_valid_MAC mAC)) then
+  let did_client_specify_MAC = mAC <> "" in
```
psafont: It's probably a good idea to use "user" rather than "client". The second one may have more than one meaning depending on the context :)

ocaml/xapi/nm.ml
```diff
@@ -331,27 +330,19 @@ let bring_pif_up ~__context ?(management_interface=false) (pif: API.ref_PIF) =
...
-
-            let slaves = dbBond.API.bond_slaves in
-
-            List.iter (fun self -> Db.PIF.set_currently_attached ~__context ~self ~value:false) slaves
-        end;
-
-        let rc = if !mac_changed then Db.PIF.get_record ~__context ~self:pif else rc in
+              let primary_slave_ref = bond_record.API.bond_primary_slave in
+              let primary_slave_MAC = Db.PIF.get_MAC ~__context ~self:primary_slave_ref in
+              if rc.pIF_MAC <> primary_slave_MAC then {rc with API.pIF_MAC = primary_slave_MAC} else rc
```
robhoes: We still need to update the MAC in the DB, as well as updating the local record that we pass around in this function.

lippirk:
@robhoes To test, I have:

- verified that creating a bond with the mac param sets `auto_update_mac` to `false` (and conversely that excluding the mac param sets `auto_update_mac` to `true`.

- added the following at the end of `refresh_internal`
```
  (* If PIF is a primary slave, fake its mac address *)
  if pif.API.pIF_bond_slave_of <> Ref.null then
    let bond_ref = pif.API.pIF_bond_slave_of in
    let prim_slave_ref = Db.Bond.get_primary_slave ~__context ~self:bond_ref in
    if Db.PIF.get_uuid ~__context ~self:prim_slave_ref = pif.API.pIF_uuid then
      Db.PIF.set_MAC ~__context ~self ~value:"00:1b:21:0f:3a:b7
```
and verified that when xapi restarts, `auto_update_mac = true` means the bond's master pif has its mac address copied from the primary slave, but not when `auto_update_mac = false`. 


I feel like it would be nice to work this into a test case! would that me too much effort?

ocaml/tests/test_common.ml
```diff
@@ -191,7 +191,7 @@ let make_vlan ~__context ~tagged_PIF ~untagged_PIF ~tag ?(other_config=[]) () =
 let make_network_sriov = Xapi_network_sriov.create_internal
 
 let make_bond ~__context ?(ref=Ref.make ()) ?(uuid=make_uuid ()) ~master ?(other_config=[]) ?(primary_slave=Ref.null) ?(mode=`activebackup) ?(properties=[]) () =
-  Db.Bond.create ~__context ~ref ~uuid ~master ~other_config ~primary_slave ~mode ~properties ~links_up:0L;
+  Db.Bond.create ~__context ~ref ~uuid ~master ~other_config ~primary_slave ~mode ~properties ~links_up:0L ~auto_update_mac:false;
```
robhoes: Make this an optional argument `?(auto_update_mac=false)`.

robhoes:
@lippirk Yes, let's add a unit test. You can do this by factoring out the code that you've added to `bring_pif_up` into a separate function, which can be tested on its own.

robhoes:
By the way, I don't see the changes for adding the new field to the CLI...?

ocaml/tests/test_common.ml
lippirk: i can't figure out how to make this argument optional, and could not find an example elsewhere?

(i tried to add an element to the versioned_params list in datamodel.ml, Bond.create, but this didn't work)

lippirk: aah, you mean an optional argument of `make_bond`... :+1:

ocaml/xapi/nm.mli
```diff
@@ -31,3 +31,6 @@ val with_local_lock : (unit -> 'a) -> 'a
 (** [is_dom0_interface pif_r] returns true if pif_r is a network interface
     	which has a dom0 endpoint *)
 val is_dom0_interface : API.pIF_t -> bool
+
+(* Internals *)
+val update_master_pif_mac : __context:Context.t -> API.bond_t -> API.pIF_t -> API.ref_PIF -> API.pIF_t
```
lippirk: this feels wrong, and the tests also felt awkward to write... what are your thoughts?

ocaml/xapi/records.ml
```diff
@@ -144,6 +144,7 @@ let bond_record rpc session_id bond =
           ~set_in_map:(fun k v -> Client.Bond.set_property rpc session_id bond k v) ();
         make_field ~name:"primary-slave" ~get:(fun () -> get_uuid_from_ref (x ()).API.bond_primary_slave) ();
         make_field ~name:"links-up" ~get:(fun () -> Int64.to_string (x ()).API.bond_links_up) ();
+        make_field ~name:"auto_update_mac" ~get:(fun () -> string_of_bool (x ()).bond_auto_update_mac) ();
```
robhoes: We always use `-` in the CLI rather than `_`.

robhoes:
I think that the test will be easier to write if you use the helper module and functor as in https://github.com/xapi-project/xen-api/blob/master/ocaml/tests/test_helpers.ml#L22.

ocaml/xapi/nm.ml
```diff
@@ -309,6 +309,20 @@ let determine_static_routes net_rc =
   else
     []
 
+let update_master_pif_mac ~__context bond_record (bond_master_record : API.pIF_t) bond_master =
```
robhoes: We often use the prefix `maybe_` for functions that may or may not do something depending on the input.

lippirk:
> I think that the test will be easier to write if you use the helper module and functor as in https://github.com/xapi-project/xen-api/blob/master/ocaml/tests/test_helpers.ml#L22.

I definitely prefer the output from alcotest! (from the code i got the impression that the encapsulate state functor currently works only with ounit??)

robhoes:
Looks like it does, yes. Perhaps we should take the opportunity to switch it (https://github.com/xapi-project/xapi-test-utils) over to alcotest...?

lippirk:
> Looks like it does, yes. Perhaps we should take the opportunity to switch it (https://github.com/xapi-project/xapi-test-utils) over to alcotest...?

It looks like this has already been raised: https://issues.citrite.net/browse/CP-28368

lippirk:
> Looks like it does, yes. Perhaps we should take the opportunity to switch it (https://github.com/xapi-project/xapi-test-utils) over to alcotest...?

will do that once this is merged :+1:

ocaml/tests/test_nm.ml
```diff
@@ -0,0 +1,72 @@
...
+    let show_auto_update_mac = fun auto_update_mac -> "auto_update_mac="^string_of_bool auto_update_mac
+    let show_primary_slave_mac_change = fun o_new_primary_slave_mac -> match o_new_primary_slave_mac with
+      | None -> "primary_slave mac unchanged"
+      | Some new_primary_slave_mac -> "primary_slave mac -> "^new_primary_slave_mac
+
+    let string_of_input_t = Test_printers.pair show_auto_update_mac show_primary_slave_mac_change
+    let string_of_output_t (rc,db) = "Bond master macs. record: "^rc^" "^"db : "^db
+  end
+
+  module State  = struct
```
robhoes: You could also do:
```
module State = Test_state.XapiDb
```
See test_state.ml. This is what we tend to use for unit tests that need a xapi DB.

ocaml/tests/test_nm.ml
```diff
@@ -0,0 +1,72 @@
...
+open Test_common
+open OUnit
+
+module MaybeUpdateMasterPifMac = Generic.Make(Generic.EncapsulateState(struct
+  module Io = struct
+    type input_t = (bool * string option) (* auto_update_mac, new_primary_slave_mac *)
+
+    type output_t = (string * string)  (* record bond master mac, db bond master mac *)
+
+    let show_auto_update_mac = fun auto_update_mac -> "auto_update_mac="^string_of_bool auto_update_mac
```
robhoes: The `Test_printers` module has handy combinators for this kind of thing; `bool` for this case.

ocaml/tests/test_nm.ml
```diff
@@ -0,0 +1,72 @@
...
+    let show_primary_slave_mac_change = fun o_new_primary_slave_mac -> match o_new_primary_slave_mac with
```
robhoes: You can replace all the characters after the `=` on this line with `function`.

ocaml/tests/test_nm.ml
lippirk: I did try that, but i want to carry around refs to the bond and the pif master as well

PR #3906, opened 2019-07-08
CA-319960: Remove user KRBTGT from cache before checking

minglumlu:
PBIS will return success when it queries user KRBTGT even if the domain is
offline but the user KRBTGT is in the cache. This makes the XAPI fail to
check the availability of domain by this appoach.

This commit adds logic to remove the user KRBTGT from cache before XAPI
checks the availabililty of domain.

### DIFF
--- a/ocaml/auth/extauth_plugin_ADpbis.ml
+++ b/ocaml/auth/extauth_plugin_ADpbis.ml
@@ -42,6 +42,8 @@ struct
 
   let user_friendly_error_msg = "The Active Directory Plug-in could not complete the command. Additional information in the logs."
 
+  let mutex_check_availability = Locking_helpers.Named_mutex.create "IS_SERVER_AVAILABLE"
+
   let splitlines s = String.split_f (fun c -> c = '\n') (String.replace "#012" "\n" s)
 
   let pbis_common_with_password (password:string) (pbis_cmd:string) (pbis_args:string list) =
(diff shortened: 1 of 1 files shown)


ocaml/auth/extauth_plugin_ADpbis.ml
```diff
@@ -531,9 +534,22 @@ struct
     in
     begin
       debug "Testing if external authentication server is accepting requests...";
+      let full_username = get_full_subject_name username in
+      begin try
+          ignore(pbis_common "/opt/pbis/bin/ad-cache" ["--delete-user"; "--name"; full_username])
+        with
+        | Not_found -> ()
```
lindig: This is not expected - should it be logged? I assume this would be raised by `get_full_subject_name`. What does this mean for the result of `is_pbis_server_available`?

minglumlu: Hi @lindig 
Generally, this is a backport PR from master to havana LCM branch.
The original PR against master is https://github.com/xapi-project/xen-api/pull/3894

Specifically, for your question, the exception `Not_found` might be raised from the `/opt/pbis/bin/ad-cache --delete-user`, which aims to make sure there is no the user in cache.   By deleting the user from cache, it will force PBIS to connect to remote server through network. This helps XAPI to determine if the server is available. The user `KRBTGT` is a well-known user.  So I think logging the `Not_found` is not necessary here.

`get_full_subject_name` is only to concatenate its domain name with the user. I.E. from `KRBTGT` to `citrite.net\KRBTGT`

PR #3908, opened 2019-07-16
CA-322146: set NBD device scheduler and max_sectors to more efficient…

MarkSymsCtx:
… values

### DIFF
--- a/scripts/nbd_client_manager.py
+++ b/scripts/nbd_client_manager.py
@@ -167,6 +167,16 @@ def connect_nbd(path, exportname):
                 _call(cmd)
                 _wait_for_nbd_device(nbd_device=nbd_device, connected=True)
                 _persist_connect_info(nbd_device, path, exportname)
+                nbd = (nbd_device[len('/dev/'):]
+                       if nbd_device.startswith('/dev/') else nbd_device)
+                with open("/sys/block/" + nbd + "/queue/scheduler", "w") as fd:
+                    fd.write("none")
+                # Set the NBD queue size to the same as the qcow2 cluster size
+                with open("/sys/block/" + nbd + "/queue/max_sectors_kb", "w") as fd:
+                    fd.write("512")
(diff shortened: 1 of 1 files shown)


scripts/nbd_client_manager.py
```diff
@@ -167,6 +167,14 @@ def connect_nbd(path, exportname):
                 _call(cmd)
                 _wait_for_nbd_device(nbd_device=nbd_device, connected=True)
                 _persist_connect_info(nbd_device, path, exportname)
+                nbd = (nbd_device[len('/dev/'):]
+                       if nbd_device.startswith('/dev/') else nbd_device)
+                with open("/sys/block/" + nbd + "/queue/scheduler", "w") as fd:
+                    fd.write("none")
+                # Set the NBD queue size to the same as the qcow2 cluster size
+                with open("/sys/block/" + nbd + "/queue/max_sectors_kb", "w") as fd:
```
edwintorok: Note that this might increase memory usage, see https://sourceforge.net/p/nbd/mailman/message/30592319/ to 128MiB instead of 32MiB, do you need to adjust nr_requests?

MarkSymsCtx: We could do that, I think it only actually gets used with QD=1 and it's not clear if the request buffers get allocated upfront or only when needed.

TimSmithCtx: I would argue we could even drop in lower, to 16 or even 8
Since our NBD server is on the same host we should rather queue closer to the hardware if we're going to queue, and since we are using scheduler=none there is no benefit to the scheduler from using a deeper queue (as there would otherwise be)

edwintorok: There might be opportunities for coalescing, although coalescing is more beneficial on the iSCSI blockdevice side than pure localhost.

MarkSymsCtx: Set to 8

MarkSymsCtx:
Stress test completed.

PR #3909, opened 2019-07-16
CA-316165: Convert some Thread.delay to Delay.wait

zli:
Following suggestions on the ticket, here we convert some unconditional wait
(Thread.delay) to conditional wait (Delay.wait).

The pattern is, in these tests we expect to observe something to happen, but it
has to be within a reasonable time frame. From testing point of view, we can
not assume a function will always return (especially if it was implemented
wrong) and we certainly don't want to wait indefinitely.

Previously with Thread.delay, we unconditionally wait for a timeout, then check
the observed value. Not pretty, but it doesn't require any kind of
collaboration from the observed party. The problem is, if we set the timeout
tight, it might time out if the system is extremely stressed. But if we set the
timeout too relax, we end up wasting time every single run.

Now With Delay.wait, we wait conditionally on both a timeout and a
signal/condition. In these case, we can now set the timeout generously (say,
times higher than the usual value, as a real hard timeout), but knowing that in
most cases the wait will end much earlier as usual (upon some
signal/condition) instead of breaking out of the timeout.

### DIFF
--- a/ocaml/tests/test_vdi_cbt.ml
+++ b/ocaml/tests/test_vdi_cbt.ml
@@ -12,6 +12,7 @@
 -  GNU Lesser General Public License for more details.
   *)
 
+open Stdext
 
 let register_smapiv2_server (module S: Storage_interface.Server_impl) sr_ref =
   let module S = Storage_interface.Server(S)() in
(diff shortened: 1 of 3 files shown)


ocaml/tests/test_vdi_cbt.ml
```diff
@@ -386,14 +387,13 @@ let test_data_destroy =
...
+        let timebox_timeout = timeout +. 10. in
+        let wait_hdl = Threadext.Delay.make() in
         let raisedexn = ref None in
-        ignore (bg (fun () -> (try Xapi_vdi._data_destroy ~__context ~self:vDI ~timeout with e -> raisedexn := Some e); completed := true));
-        Thread.delay timebox_timeout;
-        if not !completed then
-          Alcotest.fail (Printf.sprintf "data_destroy did not return in %f seconds" timebox_timeout);
+        ignore (bg (fun () -> (try Xapi_vdi._data_destroy ~__context ~self:vDI ~timeout with e -> raisedexn := Some e); Threadext.Delay.signal wait_hdl));
+        if Threadext.Delay.wait wait_hdl timebox_timeout then
+          Alcotest.fail (Printf.sprintf "data_destroy did not return in %f seconds" timeout);
```
edwintorok: I think this should be `timebox_timeout` here.

zli: You're right, updated.

PR #3913, opened 2019-07-24
CA-322045: make light_fuse idempotent

edwintorok:
light_fuse* can trigger reboots and saving of RRDs and the database.
It is enough to do this once, even if we are asked to do it multiple
times (in fact it is counter-productive to do it multiple times because
        we would queueu up N RRD backups when we only wanted one).

Rebooting the system would eventually tell xapi to shutdown and trigger
the other fuse, hence the need for different fuse types: Exit when it is
for xapi itself, and Reboot when it is about the system.

I tried flooding it with `xe host-shutdown-agent` in a loop, and it logged the no-op message, and then properly exited when the main fuse was done.

### DIFF
--- a/ocaml/xapi/xapi_fuse.ml
+++ b/ocaml/xapi/xapi_fuse.ml
@@ -24,8 +24,20 @@ let time f =
   (try f () with e -> warn "Caught exception while performing timed function: %s" (Printexc.to_string e));
   Unix.gettimeofday () -. start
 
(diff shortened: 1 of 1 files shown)


edwintorok:
Although a toolstack restart works, the host now behaves oddly after a reboot, it refuses to exit maintenance mode, lets wait a bit with merging this.

PR #3914, opened 2019-07-24
sr_health_check: Actually make the thread start instead of dead code

edwintorok:
OCaml 4.08.1+rc1 found this:
```
File "ocaml/xapi/xapi_sr_operations.ml", line 236, characters 18-705:
236 | ..................Thread.create (fun () ->
237 |               let rec loop () =
238 |                 Thread.delay 30.;
239 |                 let info = C.SR.stat dbg (Storage_interface.Sr.of_string (Db.SR.get_uuid ~__context ~self)) in
240 |                 if not (Db.Task.get_status ~__context ~self:task = `cancelling) &&
...
247 |                 end
248 |               in
249 |               loop ()
250 |             )
Error (warning 5): this function application is partial,
```

And it is indeed correct, this is dead code, we never actually created
the thread, because the `_` was of type `unit -> Thread.t`.
Be explicit about the type of the ignored argument and fix it.

None of the SMAPIv3 SRs that we have actually implement a recovering
health check status, but when they do it would be confusing to have this
code that looks sort of OK but be completely broken.

### DIFF
--- a/ocaml/tests/test_xapi_xenops.ml
+++ b/ocaml/tests/test_xapi_xenops.ml
@@ -60,7 +60,7 @@ let test_xapi_restart_inner () =
   let open Xenops_interface in
   let open Xapi_xenops_queue in
   let module Client = (val make_client "simulator" : XENOPS) in
-  let _ = Client.VM.list "dbg" in
+  let _ : 'a list = Client.VM.list "dbg" () in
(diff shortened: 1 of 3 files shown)


edwintorok:
Apparently we have more of these, :hammer: will update the PR shortly

PR #3916, opened 2019-07-25
Add the ability to run xapi while out-of-sync with Xen and/or libxenctrl

robhoes:
Depends on:
- https://github.com/xapi-project/ezxenstore/pull/13
- https://github.com/xapi-project/ocaml-rrdd-plugin/pull/31
- https://github.com/xapi-project/rrd-transport/pull/57
- https://github.com/xapi-project/xenopsd/pull/646


lindig:
The commit message for ac6e420 has a typo in the subject line (no/not) and in the body (does/do).

robhoes:
I have run the usual regression-test suites. Most of the changes only kick in in the out-of-sync situation, where previously xapi wouldn't even start, so they probably won't regress that bit :)

edwintorok:
Lets put a ticket number on one of the commits so it is easier to track in the changelog.

robhoes:
Thanks, I've updated the commit messages to fix the type and include the ticket numbers.

PR #3921, opened 2019-07-31
CA-299343: Explain reason when operation fails due to DMC being unava…

robhoes:
…ilable

When changing the memory settings on a VM, setting a dynamic range is not
allowed for three possible reasons. This patch extends the error message, which
is returned by xe.


lindig:
Maybe edit the subject line to fit into the space limit.
```
CA-299343: give reason when DMC fails
```

PR #3922, opened 2019-07-31
REQ-627 Feature Branch

lindig:
This series of commits sets up the REQ-627 feature branch that prepares xapi to support SR-IOV for NVIDIA GPUs.


ocaml/xapi/vgpuops.ml
```diff
@@ -102,7 +102,13 @@ let add_pcis_to_vm ~__context host vm pci =
...
-  let value = String.concat "," (List.map Pciops.to_string devs) in
+  let value =
+    Db.VM.get_other_config ~__context ~self:vm
+    |> (fun assoc -> try List.assoc Xapi_globs.vgpu_pci assoc with Not_found -> "")
+    |> ( function (* add to existing entry *)
+       | ""  -> String.concat "," (List.map Pciops.to_string devs)
+       | str -> String.concat "," (str :: List.map Pciops.to_string devs)
+       ) in
   debug "Adding PCIs to a VM with 'other config': %s" value;
   Db.VM.add_to_other_config ~__context ~self:vm ~key:Xapi_globs.vgpu_pci ~value
```
robhoes: Does this work now? Do you not need to remove first and then re-add?

ocaml/xapi/xapi_xenops.ml
```diff
@@ -657,6 +657,18 @@ module MD = struct
       fn = 0;
     }
 
+  (** Return the virtual function (VF) for a VGPU operated in SR-IOV
+   * mode, or None otherwise
+   *)
+  let sriov_vf ~__context vgpu =
+    match vgpu.Db_actions.vGPU_pci with
```
robhoes: Where are you setting this field? So far, the PCI address only goes into the `other-config` field.

ocaml/xapi/vgpuops.ml
lindig: Is `Db.VM.add_to_other_config` overwriting the existing entry (key) with the new value? The new value is constructed by adding the new value to the existing value.

ocaml/xapi/xapi_xenops.ml
lindig: I agree, this does not work as the value is never set.

lindig:
My current plan:

* Raise error if Nvidia script can't be found
* Make path to Nvidia script configurable
* Use vGPU PCI address in datamodel and avoid using other-config key on VM

lindig:
I have addressed the comments, please take a fresh look.

ocaml/idl/datamodel.ml
```diff
@@ -4780,6 +4780,11 @@ module VGPU = struct
         field ~qualifier:DynamicRO ~ty:(Ref _pgpu) ~lifecycle:[Published, rel_dundee, ""] "scheduled_to_be_resident_on" "The PGPU on which this VGPU is scheduled to run" ~default_value:(Some (VRef null_ref));
         field ~qualifier:DynamicRO ~ty:(Map (String,String)) ~lifecycle:[Published, rel_inverness, ""] ~default_value:(Some (VMap [])) "compatibility_metadata" "VGPU metadata to determine whether a VGPU can migrate between two PGPUs";
         field ~qualifier:RW ~ty:String ~lifecycle:[Published, rel_quebec, ""] ~default_value:(Some (VString "")) "extra_args" "Extra arguments for vGPU and passed to demu";
+        field ~qualifier:DynamicRO
+          ~ty:(Ref _pci)
+          ~lifecycle:[Published, rel_quebec, ""]
+          ~default_value:(Some (VRef null_ref))
+          "pci" "PCI device";
```
robhoes: For consistency with other API fields, we should capitalise this name: `PCI`.
Also, please explain in the doc string that this is the PCI device that is passed through to the VM for this VGPU, either a full device pass-through, or an SR-IOV virtual function.

robhoes:
The commit message of d46cbde seems to be truncated. Other than that, it all looks good to me.

lindig:
I have addressed the comments and rebased. I will address the build issues and would merge this once they are resolved.

PR #3923, opened 2019-08-01
Remove unused xapi-netdev dependency

robhoes:


robhoes:
@edwintorok https://github.com/xapi-project/xs-opam/pull/401

PR #3924, opened 2019-08-01
Clean up old network scripts

robhoes:


scripts/init.d-attach-static-vdis
```diff
@@ -10,9 +10,6 @@ STATE_DIR=/etc/xensource/static-vdis
 # Source function library.
 . /etc/init.d/functions
 
-# Needed for interface-reconfigure below
-. /etc/xensource-inventory
```
edwintorok: Are we sure that is the only reason it was needed for?

robhoes: I can't see any variables from the inventory in this file. I ran the `attach-static-vdis` command on a host to check.

PR #3926, opened 2019-08-05
CP-31859 Remove support for VSS

lippirk:

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -3751,7 +3751,7 @@ module VMSS = struct
                     [
                       "snapshot", "The snapshot is a disk snapshot";
                       "checkpoint", "The snapshot is a checkpoint";
-                      "snapshot_with_quiesce", "The snapshot is a VSS";
+                      "snapshot_with_quiesce", "Support for VSS has been removed.";
                     ])
 
   let frequency = Enum ("vmss_frequency",
--- a/ocaml/idl/datamodel_vm.ml
+++ b/ocaml/idl/datamodel_vm.ml
@@ -213,7 +213,7 @@ let power_behaviour =
   (* VM.snapshot *)
   let snapshot_with_quiesce = call
       ~name:"snapshot_with_quiesce"
-      ~in_product_since: rel_orlando
+      ~lifecycle:[ Published, rel_orlando, ""; Removed, rel_quebec, "VSS support has been removed" ]
       ~doc:"Snapshots the specified VM with quiesce, making a new VM. Snapshot automatically exploits the capabilities of the underlying storage repository in which the VM's disk images are stored (e.g. Copy on Write)."
       ~result: (Ref _vm, "The reference of the newly created VM.")
       ~params:[
--- a/ocaml/xapi/xapi_vm_snapshot.ml
+++ b/ocaml/xapi/xapi_vm_snapshot.ml
@@ -57,115 +57,12 @@ let is_flag_set ~xs ~flag ~domid ~vm =
       flag (Ref.string_of vm) domid (Printexc.to_string e);
     false
 
-let quiesce_enabled ~xs ~domid ~vm =
-  let aux x = is_flag_set ~xs ~domid ~vm ~flag:x in
-  aux "feature-snapshot" || aux "feature-quiesce"
-
 (* we want to compare the integer at the end of a common string, ie. strings as x="/local/..../3" *)
 (* and y="/local/.../12". The result should be x < y.                                             *)
 let compare_snapid_chunks s1 s2 =
   if String.length s1 <> String.length s2
   then String.length s1 - String.length s2
   else compare s1 s2
-
-(* wait for the VSS provider (or similar piece of software running inside the guest) to quiesce *)
-(* the applications of the VM and to call VM.snapshot. After that, the VSS provider is supposed *)
-(* to tell us if everything happened nicely.                                                    *)
-let wait_for_snapshot ~__context ~vm ~xs ~domid ~new_name =
-  let value = Watch.value_to_appear (snapshot_path ~xs ~domid "status") in
-  match Watch.wait_for ~xs ~timeout:!Xapi_globs.snapshot_with_quiesce_timeout value with
-  | "snapshot-created" ->
-    (* Get the transportable snap ID *)
-    debug "wait_for_snapshot: getting the transportable ID";
-    let snapid = xs.Xenstore.Xs.directory (snapshot_path ~xs ~domid "snapid") in
-    let snapid = List.sort compare_snapid_chunks snapid in
-    let read_chunk x = xs.Xenstore.Xs.read (snapshot_path ~xs ~domid ("snapid/" ^ x)) in
-    let snapid = String.concat "" (List.map read_chunk snapid) in
-
-    (* Get the uuid of the snapshot VM *)
-    debug "wait_for_snapshot: getting uuid of the snapshot VM";
-    let snapshot_uuid =
-      try xs.Xenstore.Xs.read (snapshot_path ~xs ~domid "snapuuid")
-      with _ ->
-        error "The snapshot has not been correctly created; did snapwatchd create a full VM snapshot?";
-        raise (Api_errors.Server_error (Api_errors.vm_snapshot_with_quiesce_failed, [ Ref.string_of vm ])) in
-    let snapshot_ref =
-      try Db.VM.get_by_uuid ~__context ~uuid:snapshot_uuid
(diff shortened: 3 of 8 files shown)


ocaml/idl/datamodel_errors.ml
```diff
@@ -373,14 +373,8 @@ let _ =
     ~doc:"Cannot restore this VM because it would create a duplicate" ();
   error Api_errors.duplicate_mac_seed [ "seed" ]
     ~doc:"This MAC seed is already in use by a VM in the pool" ();
-  error Api_errors.vm_snapshot_with_quiesce_failed [ "vm" ]
```
robhoes: Please leave these error definitions. They are part of the API definition, and API bindings, or software using the bindings, may break if you remove them.

ocaml/idl/datamodel_vm.ml
```diff
@@ -214,17 +214,13 @@ let power_behaviour =
   let snapshot_with_quiesce = call
       ~name:"snapshot_with_quiesce"
       ~in_product_since: rel_orlando
```
robhoes: Please replace this `~in_product_since` with a `~lifecycle` arg that has `Published` and `Removed` entries.

ocaml/xapi-consts/api_errors.ml
```diff
@@ -173,9 +173,6 @@ let vbd_tray_locked = "VBD_TRAY_LOCKED"
 let vbd_missing = "VBD_MISSING"
 let vm_no_empty_cd_vbd = "VM_NO_EMPTY_CD_VBD"
 let vm_snapshot_failed = "VM_SNAPSHOT_FAILED"
-let vm_snapshot_with_quiesce_failed = "VM_SNAPSHOT_WITH_QUIESCE_FAILED"
-let vm_snapshot_with_quiesce_timeout = "VM_SNAPSHOT_WITH_QUIESCE_TIMEOUT"
-let vm_snapshot_with_quiesce_plugin_does_not_respond = "VM_SNAPSHOT_WITH_QUIESCE_PLUGIN_DEOS_NOT_RESPOND"
```
robhoes: Just noticed the typo here!!

ocaml/idl/datamodel_vm.ml
robhoes: Adding `Removed` will trigger a standard `MESSAGE_REMOVED` error to be raised, so you don't need to define a new one.

ocaml/idl/datamodel.ml
```diff
@@ -3751,7 +3751,7 @@ module VMSS = struct
                     [
                       "snapshot", "The snapshot is a disk snapshot";
                       "checkpoint", "The snapshot is a checkpoint";
-                      "snapshot_with_quiesce", "The snapshot is a VSS";
+                      "snapshot_with_quiesce", "Support for VSS has been removed. The snapshot is a VSS";
```
robhoes: Remove the second sentence?

ocaml/idl/datamodel_vm.ml
```diff
@@ -214,17 +214,14 @@ let power_behaviour =
...
       ~params:[
         Ref _vm, "vm", "The VM to be snapshotted";
         String, "new_name", "The name of the snapshotted VM"
       ]
-      ~errs:[Api_errors.vm_bad_power_state; Api_errors.sr_full; Api_errors.operation_not_allowed;
-             Api_errors.vm_snapshot_with_quiesce_failed;
-             Api_errors.vm_snapshot_with_quiesce_timeout;
-             Api_errors.vm_snapshot_with_quiesce_plugin_does_not_respond;
-             Api_errors.vm_snapshot_with_quiesce_not_supported ]
+      ~errs:[ Api_errors.vm_snapshot_with_quiesce_not_supported ]
```
lippirk: @robhoes  should i create a new api error to use here? (since otherwise the semantics of the error will change)

olivierlambert:
Hi there! Can you provide more info on the context? Eg:

1. Why removing this feature?
2. Is there a replacement planned? Because it's widely used for Windows users having Active Directory or MS SQL DB, when doing a snapshot in live.

ocaml/idl/datamodel_vm.ml
robhoes: I'd just leave the old list of errors here, to document what the function used to do.

ocaml/idl/datamodel_vm.ml
```diff
@@ -214,17 +214,14 @@ let power_behaviour =
   let snapshot_with_quiesce = call
       ~name:"snapshot_with_quiesce"
       ~in_product_since: rel_orlando
-      ~doc:"Snapshots the specified VM with quiesce, making a new VM. Snapshot automatically exploits the capabilities of the underlying storage repository in which the VM's disk images are stored (e.g. Copy on Write)."
+      ~lifecycle:[ Published, rel_orlando, ""; Removed, rel_quebec, "" ]
+      ~doc:"Support for this feature has been removed. Snapshots the specified VM with quiesce, making a new VM. Snapshot automatically exploits the capabilities of the underlying storage repository in which the VM's disk images are stored (e.g. Copy on Write)."
```
robhoes: Put the support-removed note in the string at the end of the lifecycle tuple, and leave the doc string as it was. This drives the documentation generator to show the right message. Also, you can now remove `~in_product_since`, because `~lifecycle` makes it redundant.

edwintorok:
The feature relies on undocumented behaviour of the VSS system in Windows, and modifies already created snapshots in Dom0 in order to make them bootable (IIUC VSS in Windows is meant for read-only snapshots, not r/w bootable ones).
It currently only supports MBR partitions, if the disk uses GPT the snapshot gets corrupted.
There are other problems with VSS like a hard timeout built into Windows during which you must complete the snapshot, which causes the operation to fail often.
Rather than adding more technical debt we prefer to remove the feature since it cannot be made to work reliably.

Alternatives for users:
* full disk+memory snapshot (this takes longer because it needs to write the entire memory to the snapshot)
* regular disk snapshots: the underlying filesystem in Windows should be a journaled filesystem that should be able to deal with crash recovery and crash consistency

olivierlambert:
Thanks for the answer @edwintorok 

We are using VSS at scale here (hundreds and hundreds of users since 2015) and by default (we fallback to regular snap if VSS snap fails).

Are we that extremely lucky to have no report of VSS enabled backup restore failure? We are about of some shortcomings:

* VSS timeout failure on Windows (hardcoded)
* VSS broken on non-US Windows (true story… there's even an XSO opened on that topic, confirmed by a lot of people)

But in the end, it works better than non-VSS snapshot (and therefore backup after that).

Full disk + memory snapshot are great, but IIRC, there's **no way to restore a memory snapshot** in XAPI. I agree that might be the neatest alternative, we coded that in XO but we hit a wall when we saw we could restore this memory snapshot… I think we should work together to make it happen :+1:

edwintorok:
Please open a separate issue or XSO about the ability to restore memory snapshots, `xe snapshot-revert` followed by `xe vm-resume` worked for me with a Windows 10 VM on Naples.

olivierlambert:
@edwintorok as I said in my edit, it's not restoring the snapshot per se, it's when you export the memory disk, to re-import it later (the point of a backup is not just a snapshot, but to store the content elsewhere). I'm opening an issue here if it's fine for you :+1:

olivierlambert:
Issue created! https://github.com/xapi-project/xen-api/issues/3927 Sorry for PR pollution :wink:

ocaml/xapi/xapi_vm_snapshot.ml
```diff
@@ -57,115 +57,16 @@ let is_flag_set ~xs ~flag ~domid ~vm =
...
-        Api_errors.xen_vss_req_error_creating_snapshot_xml_string ]
-    then raise (Api_errors.Server_error (error_str, [ Ref.string_of vm; error_code () ]))
-    else raise (Api_errors.Server_error (Api_errors.vm_snapshot_with_quiesce_failed, [ Ref.string_of vm; error_str ]))
-
-  | e ->
-    failwith (Printf.sprintf "wait_for_snapshot: unexpected result (%s)" e)
-
-(* We fail if the guest does not support quiesce mode. Normally, that should be detected *)
-(* dynamically by the xapi_vm_lifecycle.update_allowed_operations call.                  *)
 let snapshot_with_quiesce ~__context ~vm ~new_name =
```
robhoes: I think you can delete the entire function definition, as it is no longer called from anywhere?

lindig:
Is this ready to be merged?

lippirk:
> Is this ready to be merged?

@robhoes are you happy?

PR #3929, opened 2019-08-07
CP-12980: Prepare XenAPI.py package

psafont:
Todo:
- [x] Readme: examples now won't work unless XenAPI package is installed. (+ instructions on how to build and how to install)
- [x] Version: need to decide the version number scheme and how to set it
- [x] metadata: decide new email and url
- [x] build: enable building the package in the makefiles


scripts/Makefile
```diff
@@ -143,7 +143,7 @@ install:
 	$(IPROG) host-backup-restore/host-restore $(DESTDIR)$(LIBEXECDIR)
 # example/python
 	$(IDATA) examples/python/XenAPIPlugin.py $(DESTDIR)$(SITE_DIR)/
-	$(IDATA) examples/python/XenAPI.py $(DESTDIR)$(SITE_DIR)/
+	$(IDATA) examples/python/XenAPI/XenAPI.py $(DESTDIR)$(SITE_DIR)/
```
edwintorok: might be useful to add a Makefile target that creates the pypi package/wheel etc. that we can then call from the specfile.

psafont: The target should also generate the version number for the package before building it.

scripts/examples/python/setup.py
```diff
@@ -0,0 +1,19 @@
+from setuptools import setup, find_packages
+
+setup(name='XenAPI',
+      version='1.2',
```
kc284: It would make sense to keep the module version in sync with the API version defined in the datamodel_types.ml (current value is 2.14), even if the module does not change that often. I think it would be confusing for people if the module uses a version scheme different from the API (which is reflected in the API reference and the other SDK flavours). For now we can add it manually and later consider using a mechanism to autogenerated it (for example `ocaml/doc/jsapi.ml` and the SDK generator use mustache for wirting out the file).

scripts/examples/python/setup.py
```diff
@@ -0,0 +1,19 @@
...
+      description='XenAPI SDK, for communication with Citrix Hypervisor.',
+      author='Citrix Systems, Inc.',
+      author_email='discard@citrix.com',
```
kc284: Should we use xen-api@lists.xenproject.org instead?

psafont: It looks a bit strange since in pypi: the author is Citrix, but the mailto link points to the mail list (on the left-hand side) https://test.pypi.org/project/XenAPI/2.14/

scripts/examples/python/setup.py
psafont: I may be able to extract the API version number from `setup.py` on build-time from `datamodel_types.ml`. I've hardcoded the version for the time being.

scripts/examples/python/setup.cfg
```diff
@@ -0,0 +1,27 @@
+[metadata]
+name = XenAPI
+version = attr: src.VERSION
+description = XenAPI SDK, for communication with Citrix Hypervisor.
+author = Citrix Systems, Inc.
+author_email = xen-api@lists.xenproject.org
+url = http://community.citrix.com/display/xs/Download+SDKs
+project_urls =
+    Bug Tracker = https://github.com/xapi-project/xen-api/issues,
```
edwintorok: This comma ends up in the URL, which then causes it to 404, can you drop the comma?

scripts/examples/python/setup.cfg
```diff
@@ -0,0 +1,27 @@
+[metadata]
+name = XenAPI
+version = attr: src.VERSION
+description = XenAPI SDK, for communication with Citrix Hypervisor.
+author = Citrix Systems, Inc.
+author_email = xen-api@lists.xenproject.org
+url = http://community.citrix.com/display/xs/Download+SDKs
```
edwintorok: That has a lot of 404s, would this be a better starting point: https://developer-docs.citrix.com/citrix-hypervisor/

kc284: I think this might be better https://www.citrix.com/community/citrix-developer/citrix-hypervisor-developer/

edwintorok: It is a bit difficult to find it from there:
https://www.citrix.com/community/citrix-developer/citrix-hypervisor-developer/citrix-hypervisor-developing.html 404
https://developer-docs.citrix.com/ works but have to choose CH again to get to:
https://developer-docs.citrix.com/citrix-hypervisor/

How about we use https://developer-docs.citrix.com/citrix-hypervisor/ as the link, or make https://www.citrix.com/community/citrix-developer/citrix-hypervisor-developer/citrix-hypervisor-developing.html  redirect to it so its easier to find?

scripts/examples/python/setup.cfg
psafont: let me see, the manual says the entries in a dict are comma-separated

scripts/examples/python/setup.cfg
kc284: I don't think it is correct to link to the docs page, it is too restricted. We should be linking to the main developer page for citrix hypervisor. The 404s have to be raised as bugs with the relevant maintaining team and fixed, but they are not the python package's problem.

psafont: Docs in the SDK about python don't match the usecase of using pypi to install the package: https://developer-docs.citrix.com/projects/citrix-hypervisor-sdk/en/latest/getting-started/#python

I'd be more inclined to use developer-docs.citrix.com/citrix-hypervisor as a homepage, even if it's just because it's easier to go to the docs about python.

In the end it might be better to just direct users to this repository as it's more practical for them.

psafont:
A preview on how the package appears in the index can be found in https://test.pypi.org/project/XenAPI/

scripts/examples/python/setuppy_gen.ml
```diff
@@ -0,0 +1,11 @@
...
+
+let setup_templ = Mustache.of_string
+    "from setuptools import setup\n\
+     \n\
+     setup(version='{{version}}')\n"
+
+let json = `O [ "version", `String Datamodel_common.api_version_string ]
+
+let () =
+  Xapi_stdext_unix.Unixext.write_string_to_file filename @@ Mustache.render setup_templ json
```
kc284: I think in this particular case mustache might be a bit of an overkill since the rendered file is so small that a simple print would suffice - unless you're planning with extensibility in mind.

psafont: I can change it. I used it because it was already being used elsewhere and what takes the most time to compile is `Datamodel_common` in any case.

psafont:
squashing fixups, nosetests passed: https://travis-ci.org/xapi-project/xen-api/jobs/571236378

Failures in the other jobs are unrelated:
```
# Error: Library "bisect_ppx" not found.
# Hint: try: dune external-lib-deps --missing -j 71 -p xapi-idl @@default
# File "varstore/privileged/dune", line 8, characters 36-46:
# 8 |   (preprocess (pps ppx_deriving_rpc bisect_ppx -conditional)))
#                                         ^^^^^^^^^^
# Error: Library "bisect_ppx" not found.
```
```
dune build @install -j 2 --profile=release
File "_build/.dune/default/ocaml/xapi/dune", line 48, characters 3-18:
48 |    ezxenstore.core
        ^^^^^^^^^^^^^^^
Error: Library "ezxenstore.core" not found
```

PR #3930, opened 2019-08-12
CA-322204: Flush logs to disk before fencing

psafont:
In case we self-fence, let's wait for logs to be flushed properly so there's visibility of what happened on the logs.

### DIFF
--- a/ocaml/xapi/xapi_ha.ml
+++ b/ocaml/xapi/xapi_ha.ml
@@ -13,9 +13,9 @@
  *)
 
 (* Functions for implementing 'High Availability' (HA). File is divided into 3 sections:
-      + scripts and functions which form part of the HA subsystem interface
-      + internal API calls used for arming and disarming individual hosts
-      + external API calls (Pool.enable_ha, Pool.disable_ha) used for turning on/off HA pool-wide
+   + scripts and functions which form part of the HA subsystem interface
+   + internal API calls used for arming and disarming individual hosts
+   + external API calls (Pool.enable_ha, Pool.disable_ha) used for turning on/off HA pool-wide
 *)
 
 module D = Debug.Make(struct let name="xapi_ha" end)
@@ -81,7 +81,7 @@ let get_uuid_to_ip_mapping () =
   String_unmarshall_helper.map (fun x -> x) (fun x -> x) v
 
 (** Without using the Pool's database, returns the IP address of a particular host
-    	named by UUID. *)
+    named by UUID. *)
 let address_of_host_uuid uuid =
   let table = get_uuid_to_ip_mapping () in
   if not(List.mem_assoc uuid table) then begin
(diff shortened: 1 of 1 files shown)


psafont:
My guess is that it happens asynchronously from the call and the fence happens too fast.
For example `Xensource.log` gets written with 0s instead of the correct loglines:

```
0013e010  5f 70 69 64 3d 31 33 37  31 32 0a 00 00 00 00 00  |_pid=13712......|
0013e020  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
0013e8b0  00 00 4a 75 6e 20 32 30  20 30 30 3a 32 37 3a 35  |..Jun 20 00:27:5|
0013e8c0  31 20 6c 6f 63 61 6c 68  6f 73 74 20 78 63 70 2d  |1 localhost xcp-|
0013e8d0  72 72 64 64 3a 20 5b 64  65 62 75 67 7c 6c 6f 63  |rrdd: [debug|loc|
0013e8e0  61 6c 68 6f 73 74 2e 6c  6f 63 61 6c 64 6f 6d 61  |alhost.localdoma|
0013e8f0  69 6e 7c 30 20 7c 7c 72  72 64 64 5f 6d 61 69 6e  |in|0 ||rrdd_main|
0013e900  5d 20 52 65 61 64 69 6e  67 20 63 6f 6e 66 69 67  |] Reading config|
0013e910  75 72 61 74 69 6f 6e 20  66 69 6c 65 20 2e 2e 0a  |uration file ...|
```

edwintorok:
It depends on the loglevel, see the rsyslog configuration in `/etc/rsyslog.d/xenserver.conf`, `/var/log/crit.log` is written synchronously, everything else is asynchronous (the `-` in front of the filenames).
I don't know about `journalctl` though, it should not be involved in logging at all, we log to syslog, nor journald, and I'm not sure if rsyslog has a flush command.

lindig:
That was my point: `systemd` does not shut down, so its logs should not be affected. @edwintorok is correct that regular logs go through rsyslogd and this is unrelated with journalctl.

psafont:
I was under the impression that the syslog function interfaced with the "syslogd" daemon, in the case of CH that is journald, which is controlled by journalctl.

```
       --flush
           Asks the journal daemon to flush any log data stored in /run/log/journal into /var/log/journal, if persistent storage is
           enabled. This call does not return until the operation is complete. Note that this call is idempotent: the data is only
           flushed from /run/log/journal into /var/log/journal once during system runtime, and this command exits cleanly without
           executing any operation if this has already happened. This command effectively guarantees that all data is flushed to
           /var/log/journal at the time it returns.
```

edwintorok:
You can try this out on a host: edit those scripts to return a failure, and see if you got any log messages you expected before/after the change.
To use journald I think one has to provide special non-portable code that calls `sd_journal_send` and similar functions, `syslog` goes through `/dev/log` which in our case is `rsyslog`.

PR #3931, opened 2019-08-13
CA-324959: write out the DNS entries that 05-prepare-networking would…

edwintorok:
… expect

It expects entries of the form DNS%d=%s (starting from 1) and concatenates them with ","
instead of a single DNS= entry.

On master this has been fixed by changing what host-installer/firstboot
scripts do (they would write the DNS= entry), however for the backport
for backwards compatibility we change what the toolstack writes.
Apparently firstboot and xapi always disagreed on the format ever since the DNS= entry was introduced: xapi would write out something that firstboot was not able to restore.

The value written out by xe-reset-networking would contain commas,
but it doesn't really matter because it would just get concatenated
again.

Draft PR, in the process of testing it.

### DIFF
--- a/ocaml/xapi/xapi_pool.ml
+++ b/ocaml/xapi/xapi_pool.ml
@@ -978,7 +978,9 @@ let eject ~__context ~host =
           "\nIP=" ^ pif.API.pIF_IP ^
           "\nNETMASK=" ^ pif.API.pIF_netmask ^
           "\nGATEWAY=" ^ pif.API.pIF_gateway ^
-          "\nDNS=" ^ pif.API.pIF_DNS ^ "\n"
+          (pif.API.pIF_DNS |> String.split ','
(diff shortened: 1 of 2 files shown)


robhoes:
Interesting, then we should also fix:

* https://github.com/xenserver/xsconsole/blob/master/plugins-base/XSFeatureNetworkReset.py#L349
* https://github.com/xapi-project/xcp-networkd/blob/master/lib/network_config.ml#L81

edwintorok:
Good point, will raise some Havana PRs for those too

edwintorok:
Test worked, confirmed that without this `/etc/resolv.conf` is lost, and with this its not.

edwintorok:
Something not right the network reset script still.

edwintorok:
Needed a 2nd fix in xcp-networkd: https://github.com/xapi-project/xcp-networkd/pull/168 working well now.

edwintorok:
I'll send a PR to fix that for master.

PR #3932, opened 2019-08-13
REQ-627 CA-325126 don't let SRIOV PCI GPU block VM ops

lindig:
Previously any PCI attached to a VM would block some VM operations.
Exempt NVIDIA SRIOV GPUs from this. We run checks if we find any other
PCI device attached to a VM.

### DIFF
--- a/ocaml/xapi/xapi_vm_lifecycle.ml
+++ b/ocaml/xapi/xapi_vm_lifecycle.ml
@@ -239,12 +239,6 @@ let report_concurrent_operations_error ~current_ops ~ref_str =
   in
   Some (Api_errors.other_operation_in_progress,["VM." ^ current_ops_str; ref_str])
 
-(* Suspending, checkpointing and live-migrating are not (yet) allowed if a PCI device is passed through *)
-let check_pci ~op ~ref_str =
-  match op with
-  | `suspend | `checkpoint | `pool_migrate | `migrate_send -> Some (Api_errors.vm_has_pci_attached, [ref_str])
-  | _ -> None
-
 let check_vgpu ~__context ~op ~ref_str ~vgpus ~power_state =
   let is_migratable vgpu =
     try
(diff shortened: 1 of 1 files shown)


robhoes:
Would it be cleaner to move this into `check_pci`?

lindig:
I've tried to split up the code but it became quite difficult to comprehend. So this is my current proposal.

ocaml/xapi/xapi_vm_lifecycle.ml
```diff
@@ -496,10 +499,18 @@ let check_operation_error ~__context ~ref =
...
-      else None) in
+  let current_error = check current_error @@ fun () ->
+    let sriov_pcis = nvidia_sriov_pcis ~__context vmr.Db_actions.vM_VGPUs in
+    let is_not_sriov pci = not @@ List.mem pci sriov_pcis in
+    let pcis = vmr.Db_actions.vM_attached_PCIs in
+    match op with
+    | `suspend
+    | `checkpoint
+    | `pool_migrate
+    | `migrate_send when List.exists is_not_sriov pcis ->
```
edwintorok: Is the intention here for the `when` to apply to all ops?

lindig: Yes. Is this not what is happening?

ocaml/xapi/xapi_vm_lifecycle.ml
```diff
@@ -354,6 +348,15 @@ let maybe_get_guest_metrics ~__context ~ref =
...
   else None
 
+(* PCI devices that belong to NVIDIA SRIOV cards *)
+let nvidia_sriov_pcis ~__context vgpus =
+  vgpus
+  |> List.filter_map (fun vgpu -> Db.VGPU.get_type ~__context ~self:vgpu
+    |> fun typ -> Db.VGPU_type.get_implementation ~__context ~self:typ
+    |> function
+      | `nvidia_sriov -> Some (Db.VGPU.get_PCI ~__context ~self:vgpu)
+      | _ -> None)
```
edwintorok: If SRIOV vGPUs are now migratable should this require changes in agility.ml? Looks like we still consider all VMs with vGPUs non-agile for HA purposes, although if your pool has all compatible vGPUs they might be.

edwintorok: Nevermind agility is unrelated to migration.

ocaml/xapi/xapi_vm_lifecycle.ml
edwintorok: It is, having the variants listed on separate rows made me think the intent was for it to apply just to the last variant, perhaps some parenthesis would help to make the intent explicit?

lindig: I've made the formatting more suggestive.

PR #3933, opened 2019-08-14
CA-322204: Synchronously log when fencing is about to occur.

psafont:


psafont:
This needs to be verified

psafont:
I build a sample executable using `critical` and it logged the line to `/var/log/crit.log`:
```
# cat /var/log/crit.log 
Aug 14 14:29:57 lcy2-dt119 happy_fencer.exe: [critical|lcy2-dt119|0 ||happy_fencer] Host will self-fence via its own watchdog for safety
```

edwintorok:
Try also calling the fencing binary from your executable

psafont:
Just tested and I confirm that it logs both to `crit.log` and `daemon.log` just before fencing
```
# grep -ri fencer /var/log
/var/log/daemon.log:Aug 14 14:56:25 localhost happy_fencer.exe: [critical|lcy2-dt119|0 ||happy_fencer] Host will self-fence via its own watchdog for safety
/var/log/crit.log:Aug 14 14:56:25 localhost happy_fencer.exe: [critical|lcy2-dt119|0 ||happy_fencer] Host will self-fence via its own watchdog for safety
```

PR #3934, opened 2019-08-14
REQ-627 CP-32057 narrow definition of "related" PCIs

lindig:
PCI devices are considered to be related if they only differ in their
function. Related PCI devices are passed to a guest together.
This definition is insufficient when dealing with virtual functions,
which must be excluded.

This commit narrows the definition of "related" accordingly: virtual PCI
devices are excluded. In addition, the commit tries to make the
definition more explicit by defining predicates.

### DIFF
--- a/ocaml/xapi/xapi_pci_helpers.ml
+++ b/ocaml/xapi/xapi_pci_helpers.ml
@@ -41,6 +41,32 @@ let get_driver_name address =
       debug "get_driver_name: for %s failed with %s" address (Printexc.to_string e); 
       None
 
+let address_of_dev x =
+  let open Pci.Pci_dev in
+  Printf.sprintf "%04x:%02x:%02x.%d" x.domain x.bus x.dev x.func
+
+(** [is_virtual_pci "0000:37:00.4"] is true, if this designates a
+ * virtual PCI function (VF), false otherwise. Only a VF has a "physfn"
+ * symbolic link.
+ *)
+let is_virtual_pci addr =
+  let path = Printf.sprintf "/sys/bus/pci/devices/%s/physfn" addr in
+  try
+    ignore @@ Unix.readlink path; true
+  with _ -> false
+
+(** [is_related_to x y] is true, if two non-virtual PCI devices
+ *  only differ in their function.
+ *)
+let is_related_to (x:Pci.Pci_dev.t) (y:Pci.Pci_dev.t) =
+  let open Pci.Pci_dev in
+  x.domain = y.domain
+  && x.bus = y.bus
+  && x.dev = y.dev
+  && x.func <> y.func
+  && not @@ is_virtual_pci @@ address_of_dev x
+  && not @@ is_virtual_pci @@ address_of_dev y
+
 let get_host_pcis () =
   let default ~msg v =
     match v with
@@ -54,7 +80,6 @@ let get_host_pcis () =
           let open Pci_dev in
           debug "get_host_pcis: vendor=%04x device=%04x class=%04x"
             d.vendor_id d.device_id d.device_class;
-          let address_of_dev x = Printf.sprintf "%04x:%02x:%02x.%d" x.domain x.bus x.dev x.func in
           let vendor = { id = d.vendor_id
                        ; name = lookup_vendor_name access d.vendor_id
                                 |> default ~msg:"vendor name" }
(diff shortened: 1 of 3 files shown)


edwintorok:
The change itself LGTM, do you also need to revert the hack that was put in place previously?
Might be worth thinking about how this would affect nested virtualization, would the nested hypervisor need access to the virtual functions? (nested virt is experimental atm though)

rosslagerwall:
LGTM. Do we know how this worked previously with network SR-IOV since I would expect the same issue?

ocaml/xapi/xapi_pci_helpers.ml
```diff
@@ -80,11 +105,7 @@ let get_host_pcis () =
                           ; name = lookup_class_name access d.device_class
                                    |> default ~msg:"class name" }
           in
-          let related_devs =
-            List.filter (fun d' ->
-                let slot x = (x.domain, x.bus, x.dev) in
-                slot d' = slot d && d' <> d
-              ) devs in
+          let related_devs = List.filter (is_related_to d) devs in
```
robhoes: I think this will now trigger _lots_ of (redundant) `readlink` calls to the file system? And similar calls are also done higher up in xapi_pci.ml. It may be better to add the `physfn` to `type pci` in a first pass over the devices.

ocaml/xapi/xapi_pci_helpers.ml
```diff
@@ -41,6 +41,32 @@ let get_driver_name address =
       debug "get_driver_name: for %s failed with %s" address (Printexc.to_string e); 
       None
 
+let address_of_dev x =
```
robhoes: We have very similar functions in pciops.ml. I can't recall who we ended up with two helper modules for PCI operations (probably refactoring due to mutual module dependencies). Perhaps it is better to merge them?

ocaml/xapi/xapi_pci_helpers.ml
lindig: The `readlink` is only called when devices are likely to be related; the `pci` type being used here is not the one defined above but in `ocaml/pci` - I'm a bit reluctant to extend it because it looks like it was vendored but I'll take a look.

ocaml/xapi/xapi_pci_helpers.ml
edwintorok: They're also slighlty different, treating the function either as hex or decimal (which as I've been just explaind doesn't matter because the max is 7). However the usual way to format it is without a leading 0, so we should unify these PCI formatting functions (there is one in xcp-idl)

lindig: These printing functions depend on different types and are not compatible. In this particular instance I think having a local definition is actually clearer. But I share the frustration about the many types defined around PCI devices.

lindig: I've updated this PR:

* The previous hack around the problem is removed and will lead to an internal error if it re-appears.
* To address performance concerns, queries into sys-fs are now cached. 
* I don't understand why the problem did not surface with network SR-IOV.

ocaml/xapi/xapi_pci_helpers.ml
```diff
@@ -41,6 +41,53 @@ let get_driver_name address =
...
+  type addr = string
+
+  (** [is_virtual_pci "0000:37:00.4"] is true, if this designates a
+   * virtual PCI function (VF), false otherwise. Only a VF has a "physfn"
+   * symbolic link.
+   *)
+  let is_virtual addr =
+    let path = Printf.sprintf "/sys/bus/pci/devices/%s/physfn" addr in
+    try
+      ignore @@ Unix.readlink path; true
```
edwintorok: Would Sys.file_exists work instead?

ocaml/xapi/xapi_pci_helpers.ml
```diff
@@ -41,6 +41,53 @@ let get_driver_name address =
...
+    with _ -> false
+
+  let make () = Hashtbl.create 100
+  let is_virtual t addr =
+    try
+      Hashtbl.find t addr
```
edwintorok: There is a `Hashtbl.find_opt` since 4.05 that we could use.

ocaml/xapi/xapi_pci_helpers.ml
lindig: This is almost equivalent. `readlink` would fail if this is not a symbolic link, whereas `Sys.file_exsts` would succeed. Neither would fail if the target of the symbolic link does not exist.

PR #3936, opened 2019-08-15
CA-325319 Fix host-display script console handling

alexbrett:
There were two distinct issues with the host-display script. Firstly, it always
assumed com1 was the serial port in use, when in some cases it may be e.g. com2.
Secondly if you go through a disable/enable cycle on a normal (VGA) setup, the
result would be the loss of `console=hvc0` from the dom0 command line, because
the `set-dom0` command will *replace* all existing entries, not just append one
to the line.

These changes resolve these issues.

Signed-off-by: Alex Brett <alex.brett@citrix.com>

### DIFF
--- a/scripts/host-display
+++ b/scripts/host-display
@@ -11,6 +11,7 @@ necessarily correct at the time this script is called.
 
 import os
 import os.path
+import re
 import subprocess
 import sys
 import syslog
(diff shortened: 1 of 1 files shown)


scripts/host-display
```diff
@@ -80,6 +81,17 @@ COMMANDS = {
...
+def _get_com_port():
+    """
+        Determine from the current bootloader config which COM port is in use
+    """
+    (_, stdout, _) = doexec([XEN_CMDLINE, "--get-xen", "console"])
+    m = re.match("console=.*(com\d).*", stdout.readline().strip())
+    if m:
+        return m.group(1)
+    else:
+        return "com1"
```
edwintorok: If everything goes well this shouldn't be reached, unless there was a previous run with the old host-display script that wiped the com1 entry.

alexbrett: Indeed, I felt it a good precaution to have something present just in case however

edwintorok:
There seems to be a general problem with this script (unrelated to your change) in that it affects all grub menu entries, even non-serial ones, so after a disable/enable you end up with com1 everywhere.

PR #3939, opened 2019-08-19
Gather the list of host datasources

edwintorok:
To help diagnose bugs like XSI-443.


psafont:
Would it be worth to also print the vm sources?

PR #3940, opened 2019-08-20
CP-12980: maintain import behaviour on package

psafont:
This keeps backwards compatibility with unpackaged XenAPI.py,
Previously users importing had to change imports to `from XenAPI import XenAPI`


psafont:
Yes, the previous import was not exposing the contents of the `XenAPI.py` module  at a package level while the new one does.

PR #3942, opened 2019-08-27
CA-325988: do not lose newlines from dmidecode

edwintorok:
In 4156d7fdcb300a5d72eff0c0fabb8ecd4c98c266 we started to filter out
lines with '#' from dmidecode, however by doing so we also lost newlines
from strings.

Since 94d2ff73711becc9cc92d01606ee38929808eb14 we start getting newlines
in baseboard-strings, and we are reporting the wrong values to the
guest.

So just concatenate with newlines, removing the newlines never seem to
have been the intention of the filtering code here.


edwintorok:
I'd merge this and then run the test on the affected machine, setting up a new CI branch is not worth the overhead for this one.

alexbrett:
Trivial one, but it should be `lose` and not `loose` in the commit message

PR #3943, opened 2019-08-30
CA-325988: pass both boards to xenstore

edwintorok:
Entries for multiple boards separate with newlines (as in dmidecode),
but perhaps a better design here would be to have separate xenstore
subkeys for baseboard-manufacturer/0 and baseboard-manufacturer/1.

The board type is lost and not passed through at the moment either.


psafont:
Superseded by 3957 and 3959

PR #3944, opened 2019-08-30
CA-326174: fix race condition between SR.scan and VDI.forget

edwintorok:
VDI.forget was not marking the SR with an in-progress operation so SR.scan could've run while VDI.forget was in progress.
After a VDI.forget the VDI could disappear, causing SR.scan to report a failure.

By default only the plug operation is marked as able to execute concurrently, add VDI.forget to that list:
as long as all we do is run VDI.forget in parallel its fine, but as soon as we try to combine it with
something like SR.scan the SR.scan should get OTHER_OPERATION_IN_PROGRESS error.
VDI.forget should be safe to parallelise because it is just a DB operation, and it was parallelized before.

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -1795,6 +1795,7 @@ module SR = struct
             "vdi_data_destroy", "Deleting the data of the VDI";
             "vdi_list_changed_blocks", "Exporting a bitmap that shows the changed blocks between two VDIs";
             "vdi_set_on_boot", "Setting the on_boot field of the VDI";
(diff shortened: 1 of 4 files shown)


edwintorok:
VDI.forget should be fairly quick (just a db update), and will be serialized at the DB level, but there is no reason to reject it at the XAPI level: that lock should clear up soon enough, so we can benefit from concurrent VDI.forget. It probably doesn't matter much in practice, but just to keep the API behaving as it did previously (i.e. you don't get suddenly get other operation in progress when doing concurrent VDI.forgets).

PR #3947, opened 2019-09-06
REQ-627 CA-325131 add vGPU.PCI to vgpu map for xenopsd

lindig:
When Xapi calls VM.migrate in xenopsd by RPC the call includes a map
that handles assigning a new PCI address for the PGPU. We need a similar
mechanism for the PCI VF that is part of a vGPU. Otherwise changes in
this field would not reach Xenopsd.

This commits extends the mapping by including the vGPU.PCI value if it
is not null. It does so by using a special key in the map.

### DIFF
--- a/ocaml/xapi/xapi_vm_migrate.ml
+++ b/ocaml/xapi/xapi_vm_migrate.ml
@@ -222,7 +222,7 @@ let detach_local_network_for_vm ~__context ~vm ~destination =
   * the pGPU is released and getting the PCI will fail.
   * *)
 let infer_vgpu_map ~__context ?remote vm =
-  let vf_device x = "vf:"^x in
+  let vf_device_of x = "vf:"^x in
   match remote with
   | None ->
     let f vgpu =
@@ -237,7 +237,7 @@ let infer_vgpu_map ~__context ?remote vm =
         |> fun self -> Db.PCI.get_pci_id ~__context ~self
         |> Xenops_interface.Pci.address_of_string in
       let pf_device = vgpu.API.vGPU_device in
-      let vf_device = vf_device pf_device in
+      let vf_device = vf_device_of pf_device in
       if vgpu.API.vGPU_PCI <> API.Ref.null then
         [ pf_device, pf ()
         ; vf_device, vf ()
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -222,37 +222,58 @@ let detach_local_network_for_vm ~__context ~vm ~destination =
   * the pGPU is released and getting the PCI will fail.
   * *)
 let infer_vgpu_map ~__context ?remote vm =
+  let vf_device x = "vf:"^x in
```
edwintorok: Perhaps this should be `vf_device_of` because `vf_device` is also used as a simple value later on.

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -222,37 +222,58 @@ let detach_local_network_for_vm ~__context ~vm ~destination =
...
+        |> fun self -> XenAPI.PGPU.get_PCI rpc session self
+        |> fun self -> XenAPI.PCI.get_pci_id rpc session self
+        |> Xenops_interface.Pci.address_of_string in
+      let vf () =
+        vgpu.API.vGPU_PCI
+        |> fun self -> XenAPI.PCI.get_pci_id rpc session self
+        |> Xenops_interface.Pci.address_of_string in
+      let pf_device = vgpu.API.vGPU_device in
+      let vf_device = vf_device pf_device in
+      if vgpu.API.vGPU_PCI <> API.Ref.null then
```
edwintorok: There is quite a bit of preexisting code duplication between this and the `remote None` case, the only difference is the form of API calls (through an rpc or straight to the DB).
Looks like we'd need something like the localfn/remote rpc pairings we got in message-switch if we want to improve this. The cleanup could be done in a separate PR, for example:
```
|> fun self -> XenAPI.PGPU.get_PCI rpc session self
```
becomes
```
let dbop ~localfn ~remotefn = match remote with None -> localfn () | Some {rpc; session} -> remotefn rpc session in
...
|> fun self -> dbop ~localfn:(fun () -> Db.PCI.get_pci_id ~__context ~self) ~remotefn:(fun rpc session -> XenAPI.PGPU.get_PCI rpc session self)
```

lindig: I wrestled with the code duplication and I'm unhappy about it. The local and remote functions differ in a way that I found it difficult to unify this.

psafont: I don't see a good, simple way to deduplicate the code, unfortunately.

PR #3952, opened 2019-09-12
fixup! REQ-627 pass PCI VF for GPU SR-IOV to xenopsd

lindig:
This fixes an oversight in an earlier commit: xapi creates a vgpu object which it passes to xenopsd. This contains a VF PCI field. We have to make sure to use it only when we are indeed using SRIOV mode. This code checks the implementation of the VGPU to decide this.

### DIFF
--- a/ocaml/xapi/xapi_xenops.ml
+++ b/ocaml/xapi/xapi_xenops.ml
@@ -660,12 +660,19 @@ module MD = struct
     }
 
   (** Return the virtual function (VF) for a VGPU operated in SR-IOV
-   * mode, or None otherwise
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_xenops.ml
```diff
@@ -663,9 +663,15 @@ module MD = struct
    * mode, or None otherwise
    *)
   let sriov_vf ~__context vgpu =
+    let is_sriov () =
```
edwintorok: Perhaps add a comment saying that this is needed for pass-through of whole-device (which might still be an SRIOV device, but we decide to pass it through in  its entirety instead of individual VF).

PR #3953, opened 2019-09-13
CA-325330 add error for VGPU driver incompatibility

lindig:
Add a more specific API error message for when the Nvidia host driver
does not support migration as is the case for Linux.

### DIFF
--- a/ocaml/xapi/xapi_gpumon.ml
+++ b/ocaml/xapi/xapi_gpumon.ml
@@ -125,6 +125,15 @@ module Nvidia = struct
       | Gpumon_interface.Compatible ->
         info "VM %s Nvidia vGPU is compatible with the destination pGPU on host %s"
           (Ref.string_of vm) (Ref.string_of dest_host)
+      | Gpumon_interface.(Incompatible reasons)
+        when List.mem Gpumon_interface.Guest_driver reasons ->
+        raise Api_errors.(Server_error (
+            vgpu_guest_driver_limit,
+            [ String.concat ", " (List.map reason_to_string reasons)
+            (* There could be multiple reasons *)
+            ; Ref.string_of vm
+            ; Ref.string_of dest_host
+            ]))
       | Gpumon_interface.(Incompatible reasons) ->
         raise Api_errors.(Server_error (
             vgpu_destination_incompatible,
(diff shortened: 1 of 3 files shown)


lindig:
I updated the commit message: this is about the guest driver not supporting migration.

ocaml/xapi/xapi_gpumon.ml
```diff
@@ -125,6 +125,15 @@ module Nvidia = struct
       | Gpumon_interface.Compatible ->
         info "VM %s Nvidia vGPU is compatible with the destination pGPU on host %s"
           (Ref.string_of vm) (Ref.string_of dest_host)
+      | Gpumon_interface.(Incompatible reasons)
+        when List.mem Gpumon_interface.Guest_driver reasons ->
```
psafont: I would rather have a function selecting the preferred error, inside the `Gpumon_interface.(Incompatible reasons)` clause rather than having a new clause with most of it duplicated.

lindig: I agree that this is not nice. The problem is that the low-level driver returns a list of reasons for incompatibility. I think it is correct to give preference to one reason at this level rather than further down the stack.

psafont: I don't think selecting an error based on the list of reasons needs to be done at a lower level. Wouldn't a small function before this block suffice?

lindig: I might fail to understand - what would this function do? `List.mem` seems pretty simple to me.

psafont: I think something like this would make sense:
```
      let error_from reasons = match List.mem Gpumon_interface.Guest_driver reasons with
      | true -> vgpu_guest_driver_limit
      | false -> vgpu_destination_incompatible
      in
      match compatibility with
      | Gpumon_interface.Compatible ->
        info "VM %s Nvidia vGPU is compatible with the destination pGPU on host %s"
          (Ref.string_of vm) (Ref.string_of dest_host)
      | Gpumon_interface.(Incompatible reasons) ->
        raise Api_errors.(Server_error (
            error_from reasons,
            [ String.concat ", " (List.map reason_to_string reasons)
            (* There could be multiple reasons *)
            ; Ref.string_of vgpu
            ; Ref.string_of dest_host
            ]))
```
It decouples the selection of the error from the rest of the error handling, removing the duplication, and allowing for further changes on the selection of the API error.

lindig: The error messages in the two cases take different parameters. Once you account for that,  I don't think there is much to gain here.

PR #3955, opened 2019-09-17
CA-326621 Remove VM.migrate op when finished

lippirk:
In 80a59e5f54fee845df, we lost calls which
cleared the VM.migrate current operation from
the database after the operation had finished.
Here we reintroduce these calls.

### DIFF
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -1687,9 +1687,16 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
             (* Intra-pool: reserve resources on the destination host, then
              * forward the call to the source. *)
             let snapshot = Db.VM.get_record ~__context ~self:vm in
-            (fun ~local_fn ~__context ~vm op ->
-               allocate_vm_to_host ~__context ~vm ~host ~snapshot ~host_op:`vm_migrate ();
-               forward_vm_op ~local_fn ~__context ~vm op)
+            let clear_migrate_op () =
+              Helpers.with_global_lock
+                (fun () ->
+                  finally_clear_host_operation ~__context ~host ~host_op:`vm_migrate ();
+                  clear_scheduled_to_be_resident_on ~__context ~vm) in
+            fun ~local_fn ~__context ~vm op ->
+              finally
+                (fun () -> allocate_vm_to_host ~__context ~vm ~host ~snapshot ~host_op:`vm_migrate ();
+                           forward_vm_op ~local_fn ~__context ~vm op)
+                clear_migrate_op;
           else
             (* Cross pool: just forward to the source host. Resources on the
              * destination will be reserved separately. *)


ocaml/xapi/message_forwarding.ml
```diff
@@ -1687,9 +1687,16 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
              * forward the call to the source. *)
             let snapshot = Db.VM.get_record ~__context ~self:vm in
-            (fun ~local_fn ~__context ~vm op ->
-               allocate_vm_to_host ~__context ~vm ~host ~snapshot ~host_op:`vm_migrate ();
-               forward_vm_op ~local_fn ~__context ~vm op)
+            let clear_migrate_op () =
+              Helpers.with_global_lock
+                (fun () ->
+                  finally_clear_host_operation ~__context ~host ~host_op:`vm_migrate ();
+                  clear_scheduled_to_be_resident_on ~__context ~vm) in
```
lippirk: `finally_clear_host_operation` fixes the issue, but `clear_scheduled_to_be_resident_on` also looks important in the case that there is a vgpu

psafont: Is there any reason to call the db-changing functions here isntead of inside `allocate_vm_to_host`? It seems strange that it's done here and not where in the function where the DB is changed.

lippirk: you don't want to do it in there, because it does not give the right behaviour - you want to wait till the op has finished, i.e. after `foward_vm_op`

psafont:
`allocate_vm_to_host` seems to have a problematic design: it changes the database, but expects the callers to revert some of the database changes in some cases. It is not clear how this should be communicated.
It's probably better to change the design in this case.

ocaml/xapi/message_forwarding.ml
```diff
@@ -1687,9 +1687,16 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
+            fun ~local_fn ~__context ~vm op ->
+              finally
+                (fun () -> allocate_vm_to_host ~__context ~vm ~host ~snapshot ~host_op:`vm_migrate ();
+                           forward_vm_op ~local_fn ~__context ~vm op)
+                clear_migrate_op;
```
lippirk: todo: possibly reraise an error if there is one when clearing the migrate operation

ocaml/xapi/message_forwarding.ml
```diff
@@ -1687,9 +1687,11 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
              * forward the call to the source. *)
             let snapshot = Db.VM.get_record ~__context ~self:vm in
-            (fun ~local_fn ~__context ~vm op ->
-               allocate_vm_to_host ~__context ~vm ~host ~snapshot ~host_op:`vm_migrate ();
-               forward_vm_op ~local_fn ~__context ~vm op)
+            fun ~local_fn ~__context ~vm op ->
+              finally
+                (fun () -> allocate_vm_to_host ~__context ~vm ~host ~snapshot ~host_op:`vm_migrate ();
+                           forward_vm_op ~local_fn ~__context ~vm op)
+                (finally_clear_host_operation ~__context ~host ~host_op:`vm_migrate);
```
lippirk: i'm not sure which of the following options is better:

```
# initial attempt
finally
  (fun () -> allocate_vm_to_host ...
                 forward_vm_op ...)
  (finally_clear_host_operation ...)
```

```
# most similar to existing code
# but if allocate_vm_to_host fails after setting things
# in the db, we will have inconsistent state
allocate_vm_to_host ...
finally
  (fun () -> foward_vm_op ...)
  (finally_clear_host_operation ...)
```

```
# what i think would be most robust
finally_with_exception
  (fun () -> allocate_vm_to_host ...
                 foward_vm_op ...)
  (fun op_exception -> match op_exception with
    | Some ex -> finally_clear_host_operation ..; raise ex
    | None -> finally_clear_host_operation )
```

lippirk: i've opened https://github.com/xapi-project/stdext/pull/40 to allow me to achieve the 3rd option above

edwintorok: You could also put the finally_clear_host_operation into a try/with that just logs the exception and doesn't raise, so that your finally function never raises. That way the original exception is preserved, and we follow the semantics of the upcoming `Fun.protect` in ocaml 4.08: https://caml.inria.fr/pub/docs/manual-ocaml/libref/Fun.html

lippirk: i think we _should_ raise the exception, so as to preserve the old behaviour (unless this is definitely a bad idea?)

you're right, i shouldn't have been using finally here

ocaml/xapi/message_forwarding.ml
```diff
@@ -1688,10 +1688,15 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
-                           forward_vm_op ~local_fn ~__context ~vm op)
-                (finally_clear_host_operation ~__context ~host ~host_op:`vm_migrate);
+              let res = try
+                allocate_vm_to_host ~__context ~vm ~host ~snapshot ~host_op:`vm_migrate ();
+                forward_vm_op ~local_fn ~__context ~vm op
+              with ex ->
+                debug "migrate_send: allocate_vm_to_host failed. Clearing host operations. Error: %s" (Printexc.to_string ex);
+                finally_clear_host_operation ~__context ~host ~host_op:`vm_migrate ();
+                raise ex in
+              finally_clear_host_operation ~__context ~host ~host_op:`vm_migrate ();
```
psafont: `finally_clear_host_operation` is called now before the migration takes places, and only called afterwards if there's an error.

I think the previous commit with `finally` would have worked just fine:
If there's an exception in `allocate_vm_to_host` we reset the database state up twice, which is strange, but should get us to an inconsistent state.
If there's an exception within `forward_vm_op` the cleanup happens once.

If there's an exception in `finally_clear_host_operation` the error is logged by `finally` as a warning.

If there are no exceptions it should work just fine.

Unless I'm missing something, using the previous commit would me my preferred solution.

lippirk: i have gone for the 1st commit since the only difference was the `clear_scheduled_to_be_resident_on` call

PR #3956, opened 2019-09-19
CA-326244: do not include host name in log format

gaborigloi:
Syslog outputs it anyway, so it is pointless to include it twice in the
log line, and it complicates the Debug module of xcp-idl.


edwintorok:
Unfortunately sometimes we only see `localhost` in logs, have you checked that after removing this we get a sensible name?

gaborigloi:
Yep that's a bug in the hostname cache in the Debug module of xcp-idl.
But I've removed that completely, see https://github.com/xapi-project/xcp-idl/pull/293
Now only syslog will be responsible for logging the hostname, which should fix the issue.

edwintorok:
https://github.com/xapi-project/xen-api/pull/3960 to fix the syslog hostname caching bug.

PR #3957, opened 2019-09-19
CA-325988: Use one baseboard value when more than one is on the system

psafont:
The current code uses `dmidecode -s` to gather baseboard values.
This is problematic because it's impossible to distinguish between multiple values and a single value with multiple lines.

This adds parser using the angstrom library which can replace all ad-hoc code that was used before. As an example, there's a commit that replaces how OEM info is gathered.

Parser has been dev-tested with outputs from workstations and servers

- [x] Include tests for parser
- [x] Integration testing
- [x] Replace oem code
- [x] Replace baseboard code
- [x] Replace bios code
- [x] Replace system code

### DIFF
--- a/ocaml/tests/test_bios_strings.ml
+++ b/ocaml/tests/test_bios_strings.ml
@@ -1,0 +1,118 @@
+(** This module tests the dmidecode processing
+ *)
+open Bios_strings
+
+let load_test_data file = Stdext.Unixext.string_of_file @@ "test_data/" ^ file
+
+let baseboard_two_string = load_test_data "bios_baseboard_two.dmidecode"
+
+let baseboard_two_record = [
+  { name="Base Board Information"
+  ; values=[ "Manufacturer", "TestA Inc."
+           ; "Product Name", "Not Specified"
+           ; "Version", "A0"
+           ; "Serial Number", " .DEADBEEF"
+           ; "Asset Tag", "          "
+           ; "Features", "Board is a hosting board\nBoard is removable\nBoard is replaceable\nBoard is hot swappable"
+           ; "Location In Chassis", "Slot 00"
+           ; "Type", "Server Blade"
+           ]
+  };
+  { name="Base Board Information"
+  ; values=[ "Manufacturer", "TestB Inc."
+           ; "Product Name", "      "
+           ; "Version", "   "
+           ; "Serial Number", "FOOBAR"
+           ; "Asset Tag", "          "
+           ; "Features", "Board is removable\nBoard is replaceable\nBoard is hot swappable"
+           ; "Location In Chassis", "Slot 00"
+           ; "Type", "Interconnect Board"
+           ]
+  }]
+
+let baseboard_two =
+  [ "baseboard-manufacturer", "TestA Inc."
+  ; "baseboard-product-name", ""
+  ; "baseboard-version", "A0"
+  ; "baseboard-serial-number", ".DEADBEEF"
+  ]
+
+let baseboard_empty =
+  [ "baseboard-manufacturer", ""
+  ; "baseboard-product-name", ""
+  ; "baseboard-version", ""
+  ; "baseboard-serial-number", ""
+  ]
+
+let oem_string = load_test_data "bios_oem.dmidecode"
+
+let oem_empty =
+  [ "oem-1", "Xen"
+  ; "oem-2", "MS_VM_CERT/SHA1/bdbeb6e0a816d43fa6d3fe8aaef04c2bad9d3e3d"
+  ]
+
+let oem = oem_empty @
+  [ "oem-3", "Test System"
+  ; "oem-4", "1[087C]"
+  ; "oem-5", "3[1.0]"
+  ; "oem-6", "12[www.test.com]"
+  ; "oem-7", "14[1]"
+  ; "oem-8", "15[0]"
+  ; "oem-9", "27[10567471934]"
+  ]
+
+let invalid_string = load_test_data "bios_invalid.dmidecode"
+
+let with_array_string = load_test_data "bios_with_array.dmidecode"
+
+let with_array = [
+  { name="BIOS Language Information"
+  ; values=[ "Installable Languages", "en|US|iso8859-1\n<BAD INDEX>"
+           ; "Currently Installed Language", "en|US|iso8859-1"
+           ]
+  }]
+
+let pp_record fmt {name; values} =
+  Format.fprintf fmt "{name=%s; values=%a}" name Fmt.(Dump.list @@ pair ~sep:comma string string) values
+
+let alco_record = Alcotest.testable pp_record (=)
+
+let check_values = Alcotest.(check @@ list @@ pair string string)
+let check_records = Alcotest.(check @@ result (list alco_record) string)
+
+let values_from_string str =
+  match Angstrom.parse_string P.records str with
+  | Ok (r :: _) -> r.values
+  | Ok [] -> []
+  | Error msg -> []
+
+let test_parser () =
+  check_records "Empty string produces an empty list" (Ok [])
+    (Angstrom.parse_string P.records "");
+
+  check_records "Invalid records must be discarded and stop the parser" (Ok [])
+    (Angstrom.parse_string P.records invalid_string);
+
+  check_records "Two records with same name is valid input" (Ok baseboard_two_record)
+    (Angstrom.parse_string P.records baseboard_two_string);
+
+  check_records "Arrays must be parsed as multi-line values" (Ok with_array)
+    (Angstrom.parse_string P.records with_array_string)
+
+let test_baseboard () =
+  check_values "Baseboard values must have empty values when input is empty" baseboard_empty
+    (get_baseboard_strings @@ fun _ _ -> []);
+  check_values "Second baseboard values must be discarded" baseboard_two
+    (get_baseboard_strings @@ fun _ _ -> values_from_string baseboard_two_string)
+
+let test_oem () =
+  check_values "OEM default values must be used when input is empty" oem_empty
+    (get_oem_strings @@ fun _ _ -> []);
+  check_values "OEM default values must be listed first" oem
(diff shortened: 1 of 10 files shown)


ocaml/xapi/bios_strings.ml
```diff
@@ -32,26 +78,49 @@ let get_bios_string name =
...
+  | Ok records -> records
+  | Error msg -> []
+  in
+  let values = match List.length records = 0 with
+  | true -> (List.hd records).values
+  | false -> []
+  in
+
+  let convert i (_, value) = Printf.sprintf "oem-%d" i, remove_invisible value in
+  standard @ List.mapi convert values
```
gaborigloi: I think the numbering should start from (List.length standard) + 1, like in the old code, otherwise the oem-X numbers will be duplicated

ocaml/xapi/bios_strings.ml
```diff
@@ -32,26 +78,49 @@ let get_bios_string name =
...
+let get_baseboard_strings () =
+  let keys = ["baseboard-manufacturer"; "baseboard-product-name";
+              "baseboard-version"; "baseboard-serial-number"] in
+
+  let output = get_output_of_type "2" in
+  let records = match Angstrom.parse_string P.record_list output with
+  | Ok records -> records
+  | Error msg -> []
+  in
+  let values = match List.length records = 0 with
```
gaborigloi: Should this be the other way around because currently we try to run List.hd if the list is empty which will fail
We could
```
match records with
| [record] -> record.values
| record::_ -> warning more records (if this is unexpected); record.values
| [] -> warning no records (if this is unexpected); []
```

ocaml/xapi/bios_strings.ml
```diff
@@ -32,26 +78,49 @@ let get_bios_string name =
...
-    in
-    standard @ (loop (n+1) 0)
-  with _ -> standard
+  let output = get_output_of_type "11" in
+
+  let records = match Angstrom.parse_string P.record_list output with
+  | Ok records -> records
+  | Error msg -> []
+  in
+  let values = match List.length records = 0 with
```
gaborigloi: Same as above - it seems list.hd will always fail cause we run it when the list is empty

ocaml/xapi/bios_strings.ml
```diff
@@ -23,6 +23,50 @@ let remove_invisible str =
...
+  let key = skip is_tab *> take_till is_sep
+  let v_line = skip is_tab *> skip is_tab *> token
+
+  let single_line =
+    lift2 (fun k v -> k, v)
+    (key <* single_sep) (* k *)
+    (token) (* v *)
+
+  let multi_line =
+    lift2 (fun k v_lines -> k, String.concat "\n" v_lines)
```
gaborigloi: Looks good but IDK how general we want to be cause there can be some output that does not match either of these e.g.
```
BIOS Language Information
        Language Description Format: Long
        Installable Languages: 2
                en|US|iso8859-1
                <BAD INDEX>
        Currently Installed Language: en|US|iso8859-1
```
But if we will never encounter these then it's fine

ocaml/xapi/bios_strings.ml
```diff
@@ -23,6 +23,50 @@ let remove_invisible str =
...
+ * without its type, handle nor size *)
+type record = {name: string; values: (string * string) list}
+
+(* Parser module for the output of command dmidecode -q *)
+module P = struct
+  open Angstrom
+  let is_eol =   function | '\n' -> true | _ -> false
+  let is_tab =   function | '\t' -> true | _ -> false
+  let is_space = function | ' '  -> true | _ -> false
+  let is_sep =   function | ':'  -> true | _ -> false
```
lindig: I think it would be good to include as a comment a typical entry that the code is supposed to parse.

ocaml/xapi/bios_strings.ml
psafont: Did the current code give problems regarding malformed output? I think that logging a warning in the case that the call to dmidecode fails or parsing it should be enough.

gaborigloi: This is not malformed, but when I get the entire output with my dmidecode, it displays stuff like this towards the end.

psafont: I misunderstood, I can add that case (I hadn't seen that array output in any of the machines)

psafont: Now the parser supports this format, the end result is a single string with endlines separating the different values. I think we can get away with this since we were not processing these kinds of values before.

ocaml/xapi/bios_strings.ml
```diff
@@ -23,6 +23,68 @@ let remove_invisible str =
...
+    | Some n -> return (Some n)
+    | None -> fail "Couldn't parse number"
+  let v_line = tab *> tab *> token
+
+  (* Inline value example:
+   *	String 1: Dell System
+   *)
+  let inline_value =
+    lift2 (fun k v -> k, v)
+    (key) (* k *)
```
lindig: No need for `(...)`

ocaml/xapi/bios_strings.ml
```diff
@@ -23,6 +23,68 @@ let remove_invisible str =
...
+   *	Features:
+   *		Board is a hosting board
+   *		Board is replaceable
+   *
+   *	Items: 1
+   *		0x0000 (OEM-specific)
+   *)
+  let multi_value =
+    lift3 (fun k _ v_lines -> k, String.concat "\n" v_lines)
+    (key) (* k *)
```
lindig: Same as above.

ocaml/xapi/bios_strings.ml
```diff
@@ -32,26 +94,51 @@ let get_bios_string name =
...
+  in
+  let values = match records with
+  | record :: _ -> record.values
+  | [] -> warn "No BIOS records found"; []
+  in
+
+  let convert (key, value) =
+    let key = key
+      |> String.lowercase_ascii
+      |> Str.global_replace (Str.regexp " ") "-"
```
lindig: I think we should avoid using `Str`. It is considered a bad regexp library and use `Re` instead. Here we might even use a simpler string function.

ocaml/xapi/bios_strings.ml
```diff
@@ -32,26 +94,51 @@ let get_bios_string name =
...
-        else
-          loop index c
-      with _ -> []
-    in
-    standard @ (loop (n+1) 0)
-  with _ -> standard
+  let output = get_output_of_type "11" in
+  let start_index = List.length standard + 1 in
+
+  let records = match Angstrom.parse_string P.record_list output with
```
lindig:
```
match Angstrom.parse_string P.record_list output with
| Ok (r::_) ->
| Ok [] ->
| Error msg ->
```

ocaml/xapi/bios_strings.ml
psafont: Re's equivalent would be `|> Re.(replace_string (compile (char ' ')) ~by:"-")`

Question is, is it that much better that it needs to add a new dependency to xapi?
The benefits of using Re in this case are not apparent to me, I guess because this is not doing complex regex queries, just replacing a character in a string.

lindig: Use `String.map`
```
val map : (char -> char) -> string -> string
String.map f s applies function f in turn to all the characters of s (in increasing index order) and stores the results in a new string that is returned.
```

ocaml/xapi/bios_strings.ml
```diff
@@ -23,35 +23,117 @@ let remove_invisible str =
   let str = String.concat "\n" l in
   String.fold_left (fun s c -> if c >= ' ' && c <= '~' then s ^ (String.of_char c) else s) "" str
```
lindig: This is pretty inefficient. If we have `Astring` available, `Astring.String.filter` would be a good choice.

psafont: I'll do that.

I think most uses of this function can be dropped and just be used for the HP special case. I'll double check if dmidecode can return invisible characters.

psafont: I'd also like to see the output in the special case, I suspect the filtering of "#-commented" lines comes from trying to process dmidecode output.

psafont: In the HP special case only 6 bytes get read, all the processing and trimming done is unnecessary.

ocaml/tests/test_bios_strings.ml
```diff
@@ -0,0 +1,97 @@
+(** This module tests the dmidecode processing
+ *)
+open Bios_strings
+
+let baseboard_two_string = "Base Board Information
```
lindig: For long literal strings, I think `{| .. |}` is a good convention. But this will work, of course.

ocaml/xapi/bios_strings.ml
```diff
@@ -13,45 +13,130 @@
...
+    key (* k *)
+    ((end_of_line >>| fun _ -> None) <|> array_count) (* number of elements, ignored *)
+    (many1 v_line) (* v_lines *)
+
+  let record =
+    lift2 (fun name values -> {name; values})
+    (token) (* name *)
+    (many (multi_value <|> inline_value)) (* values *)
+
+  let record_list = sep_by1 end_of_line record
```
lindig: Some suggestions for names:
* `record_list` -> `records`
* `v_lines` -> `values` or `items`
* `token` -> `value` or `item` (token is a very generic term inside a parser)

psafont: I'd rather use `value_lines` rather than `values`, as I treat it as a single value.

I'm writing tests for the parser, in particular I've found out I need to fix the case where a record cannot be parsed entirely, but a partial record is generated, it would be pretty nasty to debug if it ever happened in a live system.

ocaml/tests/test_bios_strings.ml
psafont: Decided to stash the strings into files: they have spaces before eols, some editors might be too eager to trim the characters away.

PR #3959, opened 2019-09-20
CA-325988: Limit BIOS strings to a single line

psafont:
In the case where more than one line is found in `remove_invisible` we assume that there are multiple values and pick the first one.

This is safe to do on all 3 usages of the function:
1. When used for oem_strings the string is guaranteed to not contain any eols (the string always ends at the first eol after it starts: `String.index_from result b '\n'`)
2. When used for hp_rombios anything that is not "COMPAQ" gets converted to "" (the 6 bytes that get read); eols are irrelevant in this case.
3. When used in `get_bios_string` none of the values gathered are expected to contain eols (manufacturer names, versions and serial numbers)


psafont:
On master dropping uncommented lines has been removed; It's also not found in any of the three cases here. We still need to split the string by newlines to drop the lines other than the first.

PR #3960, opened 2019-09-20
Fix incorrect hostname in syslog: send HUP when hostname is changed

edwintorok:
Prior to this change:

```
/opt/xensource/libexec/set-hostname foo
logger testing
systemctl kill -s HUP rsyslog
logger testing
```
```
Sep 20 14:25:35 piquet root: testing
Sep 20 14:26:30 foo root: testing
```

Thanks to @srowe for the suggestion.

### DIFF
--- a/scripts/set-hostname
+++ b/scripts/set-hostname
@@ -12,4 +12,6 @@ HOSTNAME=$1
 # Set current hostname
 hostnamectl set-hostname "$HOSTNAME"
 
+systemctl kill -s HUP rsyslog
+
 exit 0


gaborigloi:
I see so I guess this is better than `systemctl restart rsyslog.service` cause it does not have to restart

edwintorok:
Would be nice if `rsyslog` had a `reload` action, but it doesn't...

gaborigloi:
But sighup doesn't stop it , or does it, so it's as good as a reload?

PR #3962, opened 2019-09-30
CA-314317: Protect PVS-cache get_or_recreate_vdi by mutex

lippirk:
The function `Pvs_cache_vdi.get_or_recreate_vdi` may be called for multiple
concurrently starting VMs. Without the mutex, this can lead to multiple VDIs
being created, while the should only be one for a given `PVS_cache_storage`
object.

### DIFF
--- a/ocaml/xapi/pvs_cache_vdi.ml
+++ b/ocaml/xapi/pvs_cache_vdi.ml
@@ -64,15 +64,19 @@ let get_vdi ~__context ~self =
     end
   end
 
+let m = Mutex.create ()
+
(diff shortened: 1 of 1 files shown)


edwintorok:
Hint: you can use `git cherry-pick -x` when cherry-picking commits to automatically add the commit hash you cherry picked from into the commit message, so it is easier to compare.

lippirk:
> Hint: you can use `git cherry-pick -x` when cherry-picking commits to automatically add the commit hash you cherry picked from into the commit message, so it is easier to compare.

that's handy! have updated the message

PR #3963, opened 2019-10-01
Remove spammy log lines

robhoes:
It appears to be very common for devices in dom0 to not have a driver link
(/sys/bus/pci/devices/XX/driver), i.e. to not be bound to a driver. This
currently leads to lots of alarming lines in the logs. However, it is nothing
to worry about, and not worth logging.


robhoes:
@lindig IMO there are just too many of those lines, and they are not useful enough.

PR #3964, opened 2019-10-02
CA-32641 Orphaned dbsync tasks cleaned up

lippirk:
When xapi repeatedly fails to start it was not
clearing failed dbsync tasks from the database.
Assigning the `resident_on` field fixes this
for a pool scenario.

### DIFF
--- a/ocaml/xapi/dbsync.ml
+++ b/ocaml/xapi/dbsync.ml
@@ -47,6 +47,8 @@ let create_host_metrics ~__context =
 
 
 let update_env () =
+  Server_helpers.exec_with_new_task "set localhost_ref"
+    (fun __context -> Xapi_globs.localhost_ref := Helpers.get_localhost ~__context);
   Server_helpers.exec_with_new_task "dbsync (update_env)" ~task_in_database:true
     (fun __context ->
        let other_config =
--- a/ocaml/xapi/dbsync_slave.ml
+++ b/ocaml/xapi/dbsync_slave.ml
@@ -241,9 +241,6 @@ let update_env __context sync_keys =
       create_localhost ~__context info;
     );
 
-  (* record who we are in xapi_globs *)
-  Xapi_globs.localhost_ref := Helpers.get_localhost ~__context;
-
   switched_sync Xapi_globs.sync_set_cache_sr (fun () ->
       try
         let cache_sr = Db.Host.get_local_cache_sr ~__context ~self:(Helpers.get_localhost ~__context) in


ocaml/xapi/dbsync.ml
```diff
@@ -47,6 +47,14 @@ let create_host_metrics ~__context =
...
 let update_env () =
+
+  (** Allows code in cancel_tasks.ml to clean up any orphaned tasks *)
+  let set_me_as_resident_on ~__context =
+    let task = Context.get_task_id __context in
+    let me_ref = Helpers.get_localhost ~__context in
+    Db.Task.set_resident_on ~__context ~self:task ~value:me_ref;
+  in
+
   Server_helpers.exec_with_new_task "dbsync (update_env)" ~task_in_database:true
```
robhoes: This function sets the `resident_on` field of the new task to `!Xapi_globs.localhost_ref`: https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/taskHelper.ml#L43. However, `Xapi_globs.localhost_ref` is set to the local host ref only later in dbsync: https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/dbsync_slave.ml#L245. Perhaps we can set this ref earlier. For example, set it right immediately where it is defined (https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_globs.ml#L36), rather than making it null.

lippirk: i didn't want to do that because I was worried about breaking other clients, but if you think it wouldn't then that would certainly make sense

robhoes: Oh no, we can't because `Helpers.get_localhost` needs the DB, which isn't immediately available. I think that we could set the ref right above here though.

lippirk: i misunderstood this yesterday, but it makes sense now.

what might be even better is to have `Helpers.get_localhost` cache the host ref as a side effect

ocaml/xapi/dbsync.ml
```diff
@@ -59,6 +69,8 @@ let update_env () =
          | _ ->
            error "Multiple pool objects detected -- this should never happen";
            [] in
+       cache_localhost_ref ~__context;
+       set_me_as_resident_on ~__context;
```
edwintorok: This wouldn't be needed anymore if you move the place where the localhost ref is set, like @robhoes said above.

lippirk: my bad, yes

lippirk:
after a few tries i think it makes most sense to do this. reasoning:

ideally we would assign the glob before `"starting up database engine", [ Startup.OnlyMaster ], start_database_engine;`, in `xapi.ml`, but that doesn't work

if we do it _after_ `start_database_engine` then we have left it too late since it calls `Dbsync.setup()`

ocaml/xapi/dbsync.ml
```diff
@@ -71,7 +71,16 @@ let update_env () =
...
+
+  let cache_xapi_globs () =
+    Server_helpers.exec_with_new_task "cache_xapi_globs" ~task_in_database:false
+    (fun __context ->
+      if !Xapi_globs.localhost_ref = Ref.null then
+        Xapi_globs.localhost_ref := Helpers.get_localhost ~__context)
+  in
+
   try
+    cache_xapi_globs ();
```
lippirk: realised this morning that i don't know whether `task1` blocks `task2` or not:

```
Server_helpers.exec_with_new_task "task1" (fun _context -> ...);
Server_helpers.exec_with_new_task "task2" (fun _context -> ...);
```

if not, then I need to change this code

robhoes: I think that all we need is `Server_helpers.exec_with_new_task "set localhost_ref" (fun __context -> Xapi_globs.localhost_ref := Helpers.get_localhost ~__context)` immediately under `let update_env () =`. And delete the corresponding line in dbsync_slave.ml.

lippirk: so `exec_with_new_task` does block?

lippirk:
please can you approve this

PR #3965, opened 2019-10-04
REQ-627 more options to infer SRIOV mode

lindig:
This supersedes commit "REQ-627 add hack to identify SR-IOV GPUs".

Previously we addressed NVidia T4 cards using SR-IOV provided it is
activated in the global configuration switch nvidia_t4_sriov. The T4
cards were detected based on their IDs.

This commit introduces more options based on the value of
nvidia_t4_sriov in xapi.conf:

* "default" - the addressing mode is taken from vgpuConfig.xml
* "true"    - T4 cards are addressed using SR-IOV, legacy otherwise
* "false"   - all cards are addressed using legacy mode

### DIFF
--- a/ocaml/xapi/xapi_globs.ml
+++ b/ocaml/xapi/xapi_globs.ml
@@ -894,7 +894,13 @@ let accept_sm_plugin name =
 
 let nvidia_multi_vgpu_enabled_driver_versions = ref ["430.19";"430.42";"440.00+"]
 let nvidia_default_host_driver_version = "0.0"
-let nvidia_t4_sriov = ref true
+
+type nvidia_t4_sriov =
+  | Nvidia_T4_SRIOV
+  | Nvidia_LEGACY
+  | Nvidia_DEFAULT
+
+let nvidia_t4_sriov = ref Nvidia_T4_SRIOV
 
 let other_options = [
   gen_list_option "sm-plugins"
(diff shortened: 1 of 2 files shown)


ocaml/xapi/xapi_globs.ml
```diff
@@ -894,7 +894,13 @@ let accept_sm_plugin name =
...
 let nvidia_multi_vgpu_enabled_driver_versions = ref ["430.19";"430.42";"440.00+"]
 let nvidia_default_host_driver_version = "0.0"
-let nvidia_t4_sriov = ref true
+
+type nvidia_t4_sriov =
+  | Nvidia_T4_SRIOV
+  | Nvidia_LEGACY
+  | Nvidia_DEFAULT
+
+let nvidia_t4_sriov = ref Nvidia_T4_SRIOV
```
edwintorok: I assume this will get set to Nvidia_DEFAULT before merging, and just set to T4 on feature branch?
That is fine, another approach would be to change the default in `SOURCES/xenopsd/xenopsd-conf` on the feature branch.

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -451,17 +467,16 @@ module Vendor_nvidia = struct
     List.filter_map (fun vgpu_type ->
       let is_sriov id =
         try (* IDs of T4 cards *)
-          !Xapi_globs.nvidia_t4_sriov && List.mem (int_of_string id)
```
edwintorok: The call to this function is guarded by the global flag, ok.

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -451,17 +467,16 @@ module Vendor_nvidia = struct
...
+          List.mem (int_of_string id)
             [230; 231; 232; 233; 234; 225; 226;
              227; 228; 229; 222; 252; 223; 224]
         with Not_found | Failure _ -> false
```
edwintorok: What can raise Not_found here? Wouldn't List.mem return false?

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -473,7 +488,10 @@ module Vendor_nvidia = struct
             psubdev_id;
             vdev_id = int_of_string (get_attr "deviceId" devid);
             vsubdev_id = int_of_string (get_attr "subsystemId" devid);
-            sriov = sriov
+            sriov = match !Xapi_globs.nvidia_t4_sriov with
+              | Xapi_globs.Nvidia_DEFAULT  -> sriov (* from XML *)
+              | Xapi_globs.Nvidia_T4_SRIOV -> is_sriov id (* true if T4 card *)
+              | Xapi_globs.Nvidia_LEGACY   -> false (* don't use SRIOV *)
```
edwintorok: Should the sriov status be logged at debug level?
Perhaps all 3 (`sriov` from xml, `is_sriov id` and the final `sriov`) to help debug issues with the xml/drivers.

ocaml/xapi/xapi_globs.ml
lindig: Yes, we will use `Nvidia_DEFAULT` (using the XML) before merging. This is just here to simplify development.

ocaml/xapi/xapi_vgpu_type.ml
lindig: You are right, `Not_found` can't happen. I will remove it.

ocaml/xapi/xapi_vgpu_type.ml
lindig: The configuration flag is always logged; currently not logged is what each card is using. Could log is as well because it is indeed difficult so see.

PR #3967, opened 2019-10-10
REQ-627 CA-328215 pass pci_passtrough for PV

lindig:
Pass the pci_passtrough flag to xenopsd. We assume the domain will use
PCI passtrough when the key "pci" is present in the VM's other_config.

### DIFF
--- a/ocaml/xapi/xapi_xenops.ml
+++ b/ocaml/xapi/xapi_xenops.ml
@@ -353,6 +353,7 @@ let builder_of_vm ~__context (vmref, vm) timeoffset pci_passthrough vgpu =
         |false -> true
       end;
       vncterm_ip = None (*None PR-1255*);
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_xenops.ml
```diff
@@ -353,6 +353,7 @@ let builder_of_vm ~__context (vmref, vm) timeoffset pci_passthrough vgpu =
         |false -> true
       end;
       vncterm_ip = None (*None PR-1255*);
+      pci_passthrough = List.mem_assoc "pci" vm.API.vM_other_config;
```
robhoes: I think that the (right) value of `pci_passthrough` is available as an argument to the outer function `builder_of_vm`. So this can just be `pci_passthrough;`.

PR #3968, opened 2019-10-11
CA-326349

lippirk:


ocaml/xapi/xapi_host.ml
```diff
@@ -596,10 +596,25 @@ let list_methods ~__context =
 let is_slave ~__context ~host = not (Pool_role.is_master ())
 
 let ask_host_if_it_is_a_slave ~__context ~host =
-  let local_fn = is_slave ~host in
-  Server_helpers.exec_with_subtask ~__context "host.ask_host_if_it_is_a_slave" (fun ~__context ->
-    (Message_forwarding.do_op_on_localsession_nolivecheck ~local_fn ~__context
-      ~host (fun session_id rpc -> Client.Client.Pool.is_slave rpc session_id host)))
+  let ask_if_slave_with_warn_log_when_slow ~__context =
```
edwintorok: Perhaps just `ask_and_warn_when_slow` is enough

ocaml/xapi/xapi_host.ml
```diff
@@ -596,10 +596,25 @@ let list_methods ~__context =
...
+    let local_fn = is_slave ~host in
+    let timeout = 10. in
+    let task_name = Context.get_task_id __context |> Ref.string_of in
+    let log_host_slow_to_respond () =
+      D.warn "ask_host_if_it_is_a_slave: host taking a long time to respond - IP: %s; uuid: %s"
+        (Db.Host.get_address ~__context ~self:host) (Db.Host.get_uuid ~__context ~self:host)
+    in
+
```
edwintorok: Perhaps rearm the scheduler with an exponential backoff so it is obvious if we are stuck for too long, and we don't flood the logs either:
```
let rec log_host_slow_to_respond timeout () =
....
Xapi_periodic_scheduler.add_to_queue task_name Xapi_periodic_scheduler.OneShot (log_host_slow_to_respond (timeout *. 2.))
in
```

edwintorok:
Good idea to use the scheduler for showing the warning, added some suggestions on improving the warnings.

scripts/xe-toolstack-restart
```diff
@@ -32,6 +32,13 @@ SERVICES="perfmon xapi v6d xenopsd xenopsd-xc xenopsd-xenlight
   xcp-rrdd-xenpm xcp-rrdd-gpumon xcp-rrdd xcp-networkd squeezed forkexecd
   $MPATHALERT xapi-storage-script xapi-clusterd varstored-guard message-switch"
 
+# the following command gives ugly output
+/opt/xensource/libexec/xapissl stop > /dev/null 2>&1
+kill_stunnel_exit_code=$?
+if [[ $kill_stunnel_exit_code != 0 ]]; then
+        echo "Failed to kill stunnel processes!";
+fi
```
lindig: If the command failed, we have no idea why. Maybe:
```
ERROR=$(/opt/xensource/libexec/xapissl stop > /dev/null 2>&1)
if [[ $? != 0 ]]; then
   echo "Failed to kill stunnel processes!"
   echo "$ERROR"
fi
```

I don't have enough experience with shell scripts to know whether that is a viable idea.

ocaml/xapi/xapi_host.ml
lippirk: i tried to test with the following

```
let rec log_host_slow_to_respond timeout' () =
      D.warn "ask_host_if_it_is_a_slave: host taking a long time to respond - IP: %s; uuid: %s"
        (Db.Host.get_address ~__context ~self:host) (Db.Host.get_uuid ~__context ~self:host);
      Unix.sleep timeout';
      log_host_slow_to_respond (timeout') ();
    in

    Xapi_periodic_scheduler.add_to_queue task_name Xapi_periodic_scheduler.OneShot (float_of_int timeout) (log_host_slow_to_respond timeout);
    Unix.sleep 10;
    let res = Message_forwarding.do_op_on_localsession_nolivecheck ~local_fn ~__context ~host
      (fun session_id rpc -> Client.Client.Pool.is_slave rpc session_id host)
    in
    Xapi_periodic_scheduler.remove_from_queue task_name;
    res
  in
```
but it the logs keep on printing regardless; so it seems that once the scheduler has started executing the callback you have given it, removing the task from the queue will not kill the callback

scripts/xe-toolstack-restart
lippirk: i'll test this

ocaml/xapi/xapi_host.ml
edwintorok: Don't call the function recursively, schedule a new Oneshot event instead.
Of course that once the callback starts it is not interruptable, this is not Lwt.

scripts/xe-toolstack-restart
edwintorok: I think the redirections need to be in the opposite order, otherwise it all goes to `/dev/null`: https://stackoverflow.com/a/549776

ocaml/xapi/xapi_host.ml
robhoes: Perhaps add a maximum interval when resetting the timer, e.g.: `min (timeout *. 2.) 300.`.

scripts/xe-toolstack-restart
lippirk: that actually was my intention because the output is messy

instead it would be better to store stdout and stderr somewhere, and display them if there was a problem

scripts/xe-toolstack-restart
```diff
@@ -32,6 +32,16 @@ SERVICES="perfmon xapi v6d xenopsd xenopsd-xc xenopsd-xenlight
   xcp-rrdd-xenpm xcp-rrdd-gpumon xcp-rrdd xcp-networkd squeezed forkexecd
   $MPATHALERT xapi-storage-script xapi-clusterd varstored-guard message-switch"
 
+tmp_file=$(mktemp)
+/opt/xensource/libexec/xapissl stop > $tmp_file 2>&1
+kill_stunnel_exit_code=$?
+if [[ $kill_stunnel_exit_code != 0 ]]; then
+  echo $'\nFailed to kill stunnel processes! Output:'
```
lindig: What is `$` doing here? I have not seen used it like this.

ocaml/xapi/xapi_host.ml
```diff
@@ -596,10 +596,24 @@ let list_methods ~__context =
...
-  let local_fn = is_slave ~host in
-  Server_helpers.exec_with_subtask ~__context "host.ask_host_if_it_is_a_slave" (fun ~__context ->
-    (Message_forwarding.do_op_on_localsession_nolivecheck ~local_fn ~__context
-      ~host (fun session_id rpc -> Client.Client.Pool.is_slave rpc session_id host)))
+  let ask_and_warn_when_slow ~__context =
+    let local_fn = is_slave ~host in
+    let timeout = 10 in
+    let task_name = Context.get_task_id __context |> Ref.string_of in
+    let ip, uuid = Db.Host.get_address ~__context ~self:host, Db.Host.get_uuid ~__context ~self:host in
+    let rec log_host_slow_to_respond timeout' () =
```
lindig: The `timeout'` is here to distinguish it from from `timeout` but I think it would be fine to use `timeout` and shadow the outer timeout as we don't need access to it.

ocaml/xapi/xapi_host.ml
```diff
@@ -596,10 +596,24 @@ let list_methods ~__context =
...
+      D.warn "ask_host_if_it_is_a_slave: host taking a long time to respond - IP: %s; uuid: %s" ip uuid;
```
lindig: It's fine to include the name of the function; another option is to use `__LOC__` which contains the current source code location.

lippirk:
> Some commits don't have meaningful subject lines - I think they should.

Are you talking about the PR ones? I plan on squashing those once the changes are approved

scripts/xe-toolstack-restart
```diff
@@ -32,6 +32,16 @@ SERVICES="perfmon xapi v6d xenopsd xenopsd-xc xenopsd-xenlight
...
+  cat $tmp_file
```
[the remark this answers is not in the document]
lippirk: i'm calling `rm $tmp_file` just below?

i agree personalising the tmp file name is a good idea

ocaml/xapi/xapi_host.ml
lippirk: that's cool, i didn't know that's what `__LOC__` was for

scripts/xe-toolstack-restart
lippirk: without the `$`, echo prints the `\n` literally as `\n` rather than a new line

ocaml/xapi/xapi_host.ml
lippirk: hmm, i suppose it is also preferable here because it ensures i won't accidentally use `timeout` where i might have actually wanted to use `timeout'`. 

I have always thought of shadowing as being bad. It would be nice to get other toolstackers' opinions on shadowing

ocaml/xapi/xapi_host.ml
lippirk: I will stick with using the function name unless you have an objection

scripts/xe-toolstack-restart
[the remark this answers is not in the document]
lindig: Sorry - I had overlooked that. Might want to use `rm -f` which would not fail if the file does not exist. But this is debatable.

scripts/xe-toolstack-restart
lindig: I would suggest to avoid this; instead:
```
echo
echo 'Failed to kill stunnel processes! Output:'
```

psafont: COnsider `printf '\nFailed to kill stunnel processes! Output:'`

lippirk: i went with `printf`

ocaml/xapi/xapi_host.ml
edwintorok: Shadowing is bad when you have 2 modules defining the same value, and you open both of them: when you open the 2nd module code using the 1st module would break (or worse call the function from the wrong module if their types actually match). In this case it is better to refer to the function with `Module.function` instead of open (the compiler does warn when open shadows things). If that is too long you can define a local module or function alias and use that.

For values inside a function I do not find shadowing particularly bad, as long as both values refer to semantically the same thing. In some cases it is even helpful, like in the example you point out: it prevents you from accessing the wrong `timeout` value by accident.
I would object to calling everything `let x = ` though, in that case shadowing would be a bad thing, because it'd be hard to see what the `x` actually means in each case.

PR #3970, opened 2019-10-15
Revert "CA-32641 Orphaned dbsync tasks cleaned up"

robhoes:
Reverts xapi-project/xen-api#3964. This broke some tests, which we will need to investigate.


lippirk:
to make the original commit work, we could assign the `resident_on` for this task immediately after `Xapi_globs.localhost_ref` has been assigned

PR #3973, opened 2019-10-18
REQ-627 PCI handling

lindig:
This is work towards making PCI handling work: during an SR-IOV migration we need to update a VMs attached PCI devices. This will need support from xenopsd and this is just the xapi part. Note that we not yet support migration of SRIOV vGPUs which is the main motivation for this.


lindig:
Will explore this first on a feature branch.

PR #3974, opened 2019-10-21
CA-329107: Improve the order of cipher suites for stunnel

minglumlu:
This commit reverses the order of cipher suites from
"ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES256-GCM-SHA384"
to
"ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-SHA384".

The reason is that GCM is more secure than CBC (which is used in
CDHE-RSA-AES256-SHA384).


lindig:
Please re-open once this has been reviewed for security implications.

PR #3977, opened 2019-10-24
REQ-627 CA-328075 after migration, remove stale PCI

lindig:
This commit adds code to remove stale entries in a VM's list of attached
PCI devices. See CA-328075 for a detailed description and why this
problem occurs and why it is difficult to fix properly. The code assumes
that after a migration the only PCI devices attached to a VM can be
vGPUs because PCI devices can't be migrated so far.

The second commit fixes an earlier commit for releasing PCI objects from
a VM when it is halted or suspended.


lindig:
Updated the commit message.

PR #3978, opened 2019-10-24
Use systemd to daemonize wsproxy

psafont:
Uses socket activation to run wsproxy on-demand. Needs https://github.com/xapi-project/wsproxy/pull/21


psafont:
Tests are green

PR #3980, opened 2019-10-25
Corrected spelling to match the docs. Use en-us spelling. Removed unused error.

kc284:


ocaml/idl/datamodel_errors.ml
```diff
@@ -1133,8 +1133,6 @@ let _ =
     ~doc:"USB_groups are currently restricted to contain no more than one VUSB." ();
   error Api_errors.usb_already_attached [ "PUSB"; "VM" ]
     ~doc:"The USB device is currently attached to a VM." ();
-  error Api_errors.passthrough_not_enabled [ "PUSB"]
```
kc284: I don't see this one being raised anywhere. If it's needed we need to correct the text at least.

lindig:
Given that the CI has passed I assume, too, that it is safe to remove this.

PR #3981, opened 2019-10-30
Cluster host enable starts daemon

lippirk:


ocaml/xapi/helpers.ml
```diff
@@ -1180,24 +1180,18 @@ let retry ~__context ~doc ?(policy = Policy.standard) f =
...
-      | Api_errors.Server_error(code, objref :: _ ) as e when code = Api_errors.other_operation_in_progress ->
+    try
+      if TaskHelper.is_cancelling ~__context then begin
+        error "%s locking failed: task has been cancelled" doc;
+        TaskHelper.cancel ~__context;
+        raise (Api_errors.Server_error(Api_errors.task_cancelled, [ Ref.string_of (Context.get_task_id __context) ]))
+      end;
+      f ()
+    with
+    | Api_errors.Server_error(code, objref :: _ ) as e when code = Api_errors.other_operation_in_progress ->
```
psafont: I wonder if this could be written as
`| Api_errors.(Server_error(other_operation_in_progress, objref :: _ )) as e ->`

lippirk: unfortunately not, will accept string literals though

ocaml/xapi/helpers.ml
```diff
@@ -1180,24 +1180,18 @@ let retry ~__context ~doc ?(policy = Policy.standard) f =
...
-      | Api_errors.Server_error(code, objref :: _ ) as e when code = Api_errors.other_operation_in_progress ->
+    try
+      if TaskHelper.is_cancelling ~__context then begin
+        error "%s locking failed: task has been cancelled" doc;
+        TaskHelper.cancel ~__context;
+        raise (Api_errors.Server_error(Api_errors.task_cancelled, [ Ref.string_of (Context.get_task_id __context) ]))
+      end;
+      f ()
+    with
+    | Api_errors.(Server_error(other_operation_in_progress, objref :: _ )) as e ->
```
edwintorok: I don't think that this is correct: this will match on any `Server_error` and bind the value called `other_operation_in_progress` to the error code.
Previously it was also checking that the code was matching `other_operation_in_progress`. Unfortunately without a `when` I don't know of a way in OCaml to treat this as a literal instead of a value to bind.

psafont: Just tested, the explanation is correct.

ocaml/xapi/helpers.ml
```diff
@@ -1180,24 +1180,18 @@ let retry ~__context ~doc ?(policy = Policy.standard) f =
...
+      f ()
+    with
+    | Api_errors.Server_error(code, objref :: _ ) as e when code = Api_errors.other_operation_in_progress ->
         debug "%s locking failed: caught transient failure %s" doc (ExnHelper.string_of_exn e);
-        state := queue_thread (fun () -> Policy.wait ~__context !state e)
-    done;
-    match !result with
-    | Some x -> x
-    | None -> failwith "this should never happen" in
+        loop @@ queue_thread (fun () -> Policy.wait ~__context state e)
```
lindig: I believe this is not tail recursive. Add `loop [@tailcall]` for a check by the compiler.

PR #3984, opened 2019-11-11
Ca 328130

lippirk:


ocaml/xapi/xapi_pusb.ml
```diff
@@ -47,11 +47,24 @@ let scan_start ~__context usbs =
...
+      match USBSet.find_first_opt (fun elt -> USB.compare elt (extract_known_usb_info rec') = 0) local_usb_set with
+      | None -> ()
+      | Some x -> Db.PUSB.set_speed ~__context ~self ~value:x.USB.speed
+    end
+  in
+
+  (* Speed was added to the usb datamodel, so we need
+     to update that field for any usbs that were plugged
+     in before that update. *)
+  List.iter refresh_speed known_pusbs_in_db;
```
lippirk: ideally I would simply replace the existing records completely so we are robust to datamodel changes, but I think that might cause problems for any USB that is plugged (see try/catch below)

ocaml/idl/datamodel.ml
```diff
@@ -5256,6 +5256,7 @@ module PUSB = struct
           field ~qualifier:StaticRO ~ty:String ~lifecycle "description" "USB device description" ~default_value:(Some (VString ""));
           field ~qualifier:DynamicRO ~ty:Bool ~lifecycle "passthrough_enabled" "enabled for passthrough" ~default_value:(Some (VBool false));
           field ~qualifier:RW ~ty:(Map (String,String)) ~lifecycle:[Published, rel_inverness, ""] "other_config" "additional configuration" ~default_value:(Some (VMap []));
+          field ~qualifier:StaticRO ~ty:String ~lifecycle:[Published, rel_quebec, ""] "speed" "USB device speed"  ~default_value:(Some (VString ""));
```
edwintorok: Why is this a `String`, and not an `Int`?

psafont: Cannot be an Int, see https://github.com/xapi-project/xenopsd/pull/683#discussion_r344745406

lindig:
```
File "ocaml/xapi/xapi_xenops.ml", line 811, characters 8-13:
- 811 |         speed = speed;
-               ^^^^^
- Error: Unbound record field speed
```

* Travis fails with the above error. It this a cross-repo dependency? 
* Before we merge this, please squash the !fixup commits

lippirk:
> ```
> File "ocaml/xapi/xapi_xenops.ml", line 811, characters 8-13:
> - 811 |         speed = speed;
> -               ^^^^^
> - Error: Unbound record field speed
> ```
> 
>     * Travis fails with the above error. It this a cross-repo dependency?
> 
>     * Before we merge this, please squash the !fixup commits

- yes (xenopsd + xcp-idl + xapi)
- will do

PR #3985, opened 2019-11-11
CP-32437: `assert_can_boot_here`  should perform CPUID checks

edwintorok:
Details in the commit messages: this is needed so that XenCenter can use an API call instead of hardcoding/duplicating logic from XAPI.


edwintorok:
(Draft, haven't tested it yet)

PR #3989, opened 2019-11-18
CA-325582: Force unloading SR metrics from memory on destroy and forget

psafont:
While this might interfere with the archival of SR metrics due to lifecycle mismatch of SR and Host; this is preferable to have an unusable system due to memory bloat in xcp-rrdd.

The proper solution is to report these metrics as SR ones and not as Host ones, but the migration needs a bit of thought.

Depends on https://github.com/xapi-project/xcp-rrdd/pull/106

### DIFF
--- a/ocaml/xapi/xapi_sr.ml
+++ b/ocaml/xapi/xapi_sr.ml
@@ -424,7 +424,10 @@ let maybe_copy_sr_rrds ~__context ~sr =
     with Rrd_interface.Rrdd_error(Archive_failed(msg)) ->
       warn "Archiving of SR RRDs to stats VDI failed: %s" msg
 
-let forget_common ~__context ~sr =
+(* Removes the SR and dependent objects from the database (VDIs and PBDs).
+   This function is meant to be used only inside this module.
+ *)
+let _destroy ~__context ~sr =
   (* CA-325582: Since the pbds are not mounted anymore there are no more
      metrics being reported, the current ones can be deleted. *)
   let short_uuid = String.sub (Db.SR.get_uuid ~__context ~self:sr) 0 8 in
@@ -454,7 +457,7 @@ let forget_common ~__context ~sr =
 let forget  ~__context ~sr =
   (* NB we fail if ANY host is connected to this SR *)
   check_no_pbds_attached ~__context ~sr;
-  forget_common ~__context ~sr
+  _destroy ~__context ~sr
 
 (** Remove SR from disk and remove SR record from database. (This operation uses the SR's associated
     PBD record on current host to determine device_config reqd by sr backend) *)
@@ -484,7 +487,7 @@ let destroy  ~__context ~sr =
   let sm_cfg = Db.SR.get_sm_config ~__context ~self:sr in
 
   Xapi_secret.clean_out_passwds ~__context sm_cfg;
-  forget_common ~__context ~sr
+  _destroy ~__context ~sr
 
 let update ~__context ~sr =
   let open Storage_access in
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_sr.ml
```diff
@@ -424,22 +424,42 @@ let maybe_copy_sr_rrds ~__context ~sr =
     with Rrd_interface.Rrdd_error(Archive_failed(msg)) ->
       warn "Archiving of SR RRDs to stats VDI failed: %s" msg
 
+let forget_common ~__context ~sr =
```
lindig: It looks like "delete", "forget", and "destroy" are quite synonymous here. The metrics are destroyed/removed but the underlying devices of course not. The "common" part in the name is not obvious - metrics with matching names are removed but this reflects internal workings - I'd say "remove_inactive_dss"

psafont: The device has been removed at this point. Here we are updating the xapi database and metrics to reflect this. I'll try to think of a better name.

ocaml/xapi/xapi_sr.ml
```diff
@@ -424,22 +424,47 @@ let maybe_copy_sr_rrds ~__context ~sr =
     with Rrd_interface.Rrdd_error(Archive_failed(msg)) ->
       warn "Archiving of SR RRDs to stats VDI failed: %s" msg
 
+(* Removes the SR and dependent objects from the database (VDIs and PBDs).
+   This function is meant to be used only inside this module.
```
lippirk: add an `.mli`?

ocaml/xapi/xapi_sr.ml
```diff
@@ -424,22 +424,47 @@ let maybe_copy_sr_rrds ~__context ~sr =
...
+   Also removes the associated metrics from xcp-rrdd until SR metrics are
+   correctly reported as SR ones (CA-325582)
+ *)
+let _destroy ~__context ~sr =
+  (* CA-325582: Since the pbds are not mounted anymore there are no more
+     metrics being reported, the current ones can be deleted. *)
+  let short_uuid = String.sub (Db.SR.get_uuid ~__context ~self:sr) 0 8 in
+  let is_sr_metric = Astring.String.is_suffix ~affix:short_uuid in
```
lippirk: i would hide `short_uuid` since `is_sr_metric` is the thing you care about

```
let is_sr_metric =
  let short_uuid = String.sub (Db.SR.get_uuid ~__context ~self:sr) 0 8 in
  Astring.String.is_suffix ~affix:short_uuid in
in
```

psafont: The binding is only available inside the function, I don't see the benefit of hiding it to the rest of the function as it doesn't seem to cause any conflicts or confusing situation and I 'd prefer to have a flat structure if there aren't obvious downsides.

ocaml/xapi/xapi_sr.ml
```diff
@@ -424,22 +424,47 @@ let maybe_copy_sr_rrds ~__context ~sr =
...
+
+  (* SR data sources are currently stored in memory as host ones.
+     Pick the ones that match the short uuid and remove them,
+     this prevents these metrics from being archived *)
+  Rrdd.query_possible_host_dss ()
```
robhoes: These are local calls to xcp-rrdd? Then I assume that they must be run on the "SR master". Keep in mind that `SR.forget` is not forwarded and always runs on the pool master. If the SR master is not the pool master, it will not work.

robhoes: So where exactly are these metrics? On every host that has a PBD for the SR? Or just on the SR master?

psafont: The metrics are stored in every host that has a PBD for the SR

psafont:
Closing in favour of an approach that unloads the metrics from all the hosts with PDBs from to the SR.

PR #3990, opened 2019-11-19
fixup! CA-307578 Cluster_host.enable host starts clusterd

lippirk:


psafont:
Could you change the commit text and the name of the PR?

We really do not want to have master history to be rewritten and I advise against autosquashing a commit into master.

PR #3995, opened 2019-11-21
Fix compiler warnings

lippirk:


ocaml/database/redo_log.ml
```diff
@@ -378,7 +378,7 @@ let action_write_db marker generation_count write_fn sock datasockpath =
          write_fn datasock;
          R.debug "Finished writing database to data socket";
        with
-       | Sys_error("Connection reset by peer") ->
+       | Sys_error e when e = "Connection reset by peer" ->
```
psafont: This gets rid of the issue, but the fragility from depending on the particular string values is still present.

Not sure how to solve this

ocaml/xapi/cli_operations.ml
```diff
@@ -32,6 +32,8 @@ open Records
 let failwith str = raise (Cli_util.Cli_failure str)
 exception ExitWithError of int
 
+let (+++) f g a b = f (g a b)
```
psafont: Could we make the code that uses `+++` more readable?

Alternatively I would rather get `+++` defined in the `vm_migrate` scope only, where it's used

ocaml/database/redo_log.ml
mseri: Unfortunately no unless you drop the use of any function raising Sys_error. This is why we left them as they were. You could silence the warnings with `[@warning "-X"]` where X is the warning number

mseri: Similarly in cli_operations

PR #3996, opened 2019-11-22
CA-330961 Clean the yum cache before prechecking/applying an update

alexbrett:
It is possible to have two update ISOs which have the same update name, but
contain different versions of the update. The most obvious example is a driver
disk, where we might have different versions depending on the platform version
in use.

There is a potential issue in that yum will cache the repository metadata based
on the name of the repository (which in our case is the name of the update).
If the repository 'on disk' has an older timestamp than the cached data, then
yum will ignore the 'on disk' data, and use the cached information.

In the scenario where a user has erroneously attempted to apply the update from
the wrong version, we can end up in a situation where despite attempting to
apply the correct update, the metadata has been cached from the incorrect one,
and thus yum will refuse to install it due to a dependency failure (this is what
happened in XSI-539).

To avoid this, clear the yum cache before we attempt a precheck or update apply.

### DIFF
--- a/scripts/extensions/pool_update.apply
+++ b/scripts/extensions/pool_update.apply
@@ -25,6 +25,8 @@ ERROR_MESSAGE_DOWNLOAD_PACKAGE = 'Error downloading packages:\n'
 ERROR_MESSAGE_START = 'Error: '
 ERROR_MESSAGE_END = 'You could try '
 
+class EnvironmentFailure(Exception):
+    pass
 
 class ApplyFailure(Exception):
     pass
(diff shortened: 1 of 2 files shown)


scripts/extensions/pool_update.apply
```diff
@@ -44,9 +46,18 @@ def failure_message(code, params):
 
 
 def execute_apply(session, update_package, yum_conf_file):
-    cmd = ['yum', 'upgrade', '-y', '--noplugins', '-c', yum_conf_file, update_package]
     yum_env = os.environ.copy()
     yum_env['LANG'] = 'C'
+
+    cmd = ['yum', 'clean', 'all', '--noplugins', '-c', yum_conf_file]
+    p = subprocess.Popen(cmd, shell=False, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, close_fds=True, env=yum_env)
+    output, _ = p.communicate()
```
edwintorok: This seems to be duplicated in quite a few places, would `subprocess.check_output` be a suitable replacement?
Or if we need the logging could we have a helper function `call_and_log_cmd`? `remove_group` has almost the same code copy/pasted

alexbrett:
@edwintorok  no - not for Quebec (it'd be nice to have it there, but by no means essential, this has been an issue since Ely so not a new problem...)

scripts/extensions/pool_update.apply
alexbrett: So I just copied what the existing code was doing for consistency - I agree `check_output` would be cleaner, but I wonder if it's more confusing to have a mix of styles in the same script?

lindig: I think this is fine for this as it is clear and concise enough.

PR #3997, opened 2019-11-25
CA-324461: Resyncing host locally

lippirk:


ocaml/xapi/helpers.ml
```diff
@@ -127,9 +127,12 @@ let get_management_ip_addr ~__context =
 let get_localhost_uuid () =
   Xapi_inventory.lookup Xapi_inventory._installation_uuid
 
-let get_localhost ~__context : API.ref_host  =
-  let uuid = get_localhost_uuid () in
-  Db.Host.get_by_uuid ~__context ~uuid
+let get_localhost ~__context : API.ref_host =
+  let localhost_ref = !Xapi_globs.localhost_ref in
+  match !Xapi_globs.localhost_ref = Ref.null with
```
edwintorok: This could be an `if` statement

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -230,6 +197,49 @@ let enable ~__context ~self =
...
+    with err ->
+      Backtrace.is_important err;
+      let body = Printf.sprintf "Error: %s\nMessage: %s" ExnHelper.(string_of_exn err) body in
+      Xapi_alert.add ~msg:Api_messages.cluster_host_enable_failed ~cls:`Host ~obj_uuid ~body;
+      raise err
+
+let resync_host ~__context ~host =
+  let localhost = Helpers.get_localhost ~__context in
+  if not (host = localhost) then begin
+    D.error "resync_host: expected to be called with localhost %s, but got called with %s. This is a bug." (Ref.string_of localhost) (Ref.string_of host);
```
edwintorok: If it is a bug, then raise an exception

ocaml/idl/datamodel_cluster_host.ml
```diff
@@ -91,6 +91,17 @@ let forget = call
   ~hide_from_docs:true
   ()
 
+let resync_host = call
```
edwintorok: I don't think we need to expose this API, we got pool_resync already, see below.

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -230,6 +197,49 @@ let enable ~__context ~self =
...
+          end
+    with err ->
+      Backtrace.is_important err;
+      let body = Printf.sprintf "Error: %s\nMessage: %s" ExnHelper.(string_of_exn err) body in
+      Xapi_alert.add ~msg:Api_messages.cluster_host_enable_failed ~cls:`Host ~obj_uuid ~body;
+      raise err
+
+let resync_host ~__context ~host =
+  let localhost = Helpers.get_localhost ~__context in
+  if not (host = localhost) then begin
```
lippirk: i don't really want to include a host parameter here, but if i understand correctly i have to because it needs to be included in order for somebody to resync another host

that is, the forwarding should have been done already in messaging_forwarding.ml

ocaml/xapi/message_forwarding.ml
```diff
@@ -4473,5 +4473,9 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
           | _ -> find_first_live rest
       in
       find_first_live other_hosts
+
+    let resync_host ~__context ~host =
+      info "Cluster_host.resync_host host: %s" (Ref.string_of (Helpers.get_localhost ~__context));
+      Xapi_cluster_host_helpers.resync_host ~__context ~host
```
edwintorok: I think that pool_resync should do the forwarding using `do_op_on`, and likely move the implementation of pool_resync from xapi_cluster.ml here, since it is all about forwarding API calls.

ocaml/xapi/helpers.ml
robhoes: The above ref is set to a value that is returned by this `get_localhost` function, while the function now inspects the ref. See https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/dbsync_slave.ml#L245. Checking whether this is safe makes my brain explode.

In general, before making a change like this, we need to be able to guarantee that all callers of this function are happy to receive a cached version of the ref. Otherwise things may break in unexpected places.

edwintorok: Is it expected that localhost's Ref ever changes after it is set?
I agree that the potential infinite recursion is worrying, perhaps we should have a `get_localhost_uncached` and have the code that initially sets it call that instead to be simpler?

lippirk: will drop this commit and test on another branch

ocaml/idl/datamodel_cluster_host.ml
lippirk: we do because `pool_resync` needs to be able to call `resync_host` on the various hosts. it does not need a host parameter though, which is good

edwintorok: Interesting, how did this ever work before then? Was it just silently resyncing the wrong host?

edwintorok: I think we just never tested `xe cluster-pool-resync` thoroughly enough before, lets add the new API call then.

lippirk: i think it must have been, but we probably didn't notice because it was still making the enable rpc call

ocaml/xapi/xapi_cluster_host.ml
edwintorok: It is better to double check that we made the changes correctly, perhaps it should be an `assert` instead of an `if` statement though.
You do need a host object to query the cluster host object on, but as above, that can always be just the local host. The functions in `xapi_host` take a `host` or `self` parameter too, so lets keep it.

lippirk: this should no longer be necessary, provided `pool_resync` is calling `resync_host` correctly!

lippirk:
Closing for now, this seems to reliably cause CA-321975

PR #4000, opened 2019-11-26
CA-322704 error with clustering across subnets

lippirk:
We should not be able to enable clustering in
pools where hosts are in more than one subnet,
because xapi-clusterd won't work. Here we provide
a helpful error message.

### DIFF
--- a/ocaml/idl/datamodel_errors.ml
+++ b/ocaml/idl/datamodel_errors.ml
@@ -1165,6 +1165,8 @@ let _ =
     ~doc:"The last cluster host cannot be destroyed. Destroy the cluster instead" ();
   error Api_errors.no_cluster_hosts_reachable ["cluster"]
     ~doc:"No other cluster host was reachable when joining" ();
+  error Api_errors.cluster_inconsistent_subnet []
+    ~doc:"The host joining the cluster is not on the same network" ();
 
   error Api_errors.xen_incompatible []
     ~doc:"The current version of Xen or its control libraries is incompatible with the Toolstack." ()
--- a/ocaml/tests/test_helpers.ml
+++ b/ocaml/tests/test_helpers.ml
@@ -284,10 +284,40 @@ module CIDRCheckers = Generic.MakeStateless (struct
     ]
   end)
 
+module AreOnSameSubnetTests = Generic.MakeStateless (struct
+  module Io = struct
+    type input_t = string * string list
+    type output_t = bool
+
+    let string_of_input_t (netmask, ips) = Printf.sprintf "netmask: %s; ips: %s" netmask (String.concat ";" ips)
+    let string_of_output_t = function
+    | false -> "Not_same_subnet"
+    | true -> "Same_subnet"
+  end
+
+  let transform (netmask, ips) =
+    let netmask = Ipaddr.of_string_exn netmask in
+    let ips = List.map Ipaddr.of_string_exn ips in
+    match Helpers.are_on_same_subnet_exn netmask ips with
+    | `Not_same_subnet -> false
+    | `Same_subnet     -> true
+    | `Empty_ips       -> failwith "empty_ips"
+
+  let tests = `QuickAndAutoDocumented [
+    ("255.255.255.0", ["192.0.2.0"]),                                true;
+    ("255.255.255.0", ["192.0.2.0"; "192.0.2.1"]),                   true;
+    ("255.255.255.0", ["192.0.2.0"; "192.0.2.1"; "192.0.2.214"]),    true;
+    ("255.255.255.0", ["192.0.2.0"; "192.2.0.1"]),                   false;
+    ("255.0.0.0",     ["192.0.2.0"; "192.2.0.1"; "192.123.243.97"]), true;
+    ("255.0.0.0",     ["192.0.2.0"; "192.2.0.1"; "203.0.2.0"]),      false;
+  ]
+end)
+
 let tests = make_suite "helpers_" [
     "determine_gateway", DetermineGateway.tests;
     "assert_is_valid_tcp_udp_port", PortCheckers.tests;
     "assert_is_valid_tcp_udp_port_range", PortRangeCheckers.tests;
     "assert_is_valid_ip", IPCheckers.tests;
     "assert_is_valid_cidr", CIDRCheckers.tests;
+    "test_are_on_same_subnet", AreOnSameSubnetTests.tests;
   ]
--- a/ocaml/tests/test_helpers.ml
+++ b/ocaml/tests/test_helpers.ml
@@ -298,10 +298,7 @@ module AreOnSameSubnetTests = Generic.MakeStateless (struct
   let transform (netmask, ips) =
     let netmask = Ipaddr.of_string_exn netmask in
     let ips = List.map Ipaddr.of_string_exn ips in
-    match Helpers.are_on_same_subnet_exn netmask ips with
-    | `Not_same_subnet -> false
-    | `Same_subnet     -> true
-    | `Empty_ips       -> failwith "empty_ips"
+    Helpers.are_on_same_subnet_exn netmask ips
 
   let tests = `QuickAndAutoDocumented [
     ("255.255.255.0", ["192.0.2.0"]),                                true;
@@ -313,6 +310,25 @@ module AreOnSameSubnetTests = Generic.MakeStateless (struct
   ]
 end)
 
+module IntersectIpsTests = Generic.MakeStateless (struct
+  module Io = struct
+    type input_t = string list
+    type output_t = string
+
+    let string_of_input_t ips = Printf.sprintf "ips: [%s]" (String.concat "; " ips)
+    let string_of_output_t x = x
+  end
+
+  let transform ips =
+    let ips = List.map Ipaddr.of_string_exn ips in
+    Helpers.intersect_ips_exn ips |> Ipaddr.to_string
+
+  let tests = `QuickAndAutoDocumented [
+    ["255.255.255.0"; "255.255.0.0"], "255.255.0.0";
+    ["255.64.8.1"; "192.63.4.3"; "166.64.8.1"], "128.0.0.1";
+  ]
+end)
+
 let tests = make_suite "helpers_" [
     "determine_gateway", DetermineGateway.tests;
     "assert_is_valid_tcp_udp_port", PortCheckers.tests;
@@ -320,4 +336,5 @@ let tests = make_suite "helpers_" [
     "assert_is_valid_ip", IPCheckers.tests;
     "assert_is_valid_cidr", CIDRCheckers.tests;
     "test_are_on_same_subnet", AreOnSameSubnetTests.tests;
+    "test_intersect_ips", IntersectIpsTests.tests;
   ]
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -1230,3 +1230,32 @@ let host_supports_hvm ~__context host =
 let env_with_path env_vars =
   Array.of_list @@
   List.map (fun (k,v) -> Printf.sprintf "%s=%s" k v) (("PATH", String.concat ":" Forkhelpers.default_path)::env_vars)
+
+let are_on_same_subnet_exn netmask ips =
+  let ensure_same_protocol ips =
+    let coerce_v4 xs = xs |> List.map (function
+      | Ipaddr.V4 x -> x
+      | V6 _        -> raise Api_errors.(Server_error (internal_error, ["expected IPv4, but got IPv6"])))
+    in
+    let coerce_v6 xs = xs |> List.map (function
+      | Ipaddr.V6 x -> x
+      | V4 _        -> raise Api_errors.(Server_error (internal_error, ["expected IPv6, but got IPv4"])))
+    in
+    match netmask with
+    | Ipaddr.V4 nm -> `Ip_v4s (nm, coerce_v4 ips)
+    | V6        nm -> `Ip_v6s (nm, coerce_v6 ips)
+  in
+  match ensure_same_protocol ips with
+  | `Ip_v4s (netmask, ip::ips) ->
+    let network_addr = Ipaddr.V4.Prefix.of_netmask netmask ip in
+    if List.for_all (fun x -> Ipaddr.V4.Prefix.mem x network_addr) ips then
+      `Same_subnet
+    else
+      `Not_same_subnet
+  | `Ip_v6s (netmask, ip::ips) ->
+    let network_addr = Ipaddr.V6.Prefix.of_netmask netmask ip in
+    if List.for_all (fun x -> Ipaddr.V6.Prefix.mem x network_addr) ips then
+      `Same_subnet
+    else
+      `Not_same_subnet
+  | `Ip_v4s (_, []) | `Ip_v6s (_, []) -> `Same_subnet
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -1231,31 +1231,39 @@ let env_with_path env_vars =
   Array.of_list @@
   List.map (fun (k,v) -> Printf.sprintf "%s=%s" k v) (("PATH", String.concat ":" Forkhelpers.default_path)::env_vars)
 
-let are_on_same_subnet_exn netmask ips =
-  let ensure_same_protocol ips =
-    let coerce_v4 xs = xs |> List.map (function
-      | Ipaddr.V4 x -> x
-      | V6 _        -> raise Api_errors.(Server_error (internal_error, ["expected IPv4, but got IPv6"])))
-    in
-    let coerce_v6 xs = xs |> List.map (function
-      | Ipaddr.V6 x -> x
-      | V4 _        -> raise Api_errors.(Server_error (internal_error, ["expected IPv6, but got IPv4"])))
-    in
-    match netmask with
-    | Ipaddr.V4 nm -> `Ip_v4s (nm, coerce_v4 ips)
-    | V6        nm -> `Ip_v6s (nm, coerce_v6 ips)
+let ensure_same_protocol (ips: Ipaddr.t list) =
+  let coerce_v4 xs = xs |> List.map (function
+    | Ipaddr.V4 x -> x
+    | V6        x -> raise Api_errors.(Server_error (internal_error, [Printf.sprintf "expected IPv4, but got an IPv6: %s" (Ipaddr.V6.to_string x)])))
   in
+  let coerce_v6 xs = xs |> List.map (function
+    | Ipaddr.V6 x -> x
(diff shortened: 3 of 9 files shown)


ocaml/xapi/xapi_cluster.ml
```diff
@@ -47,6 +47,29 @@ let create ~__context ~pIF ~cluster_stack ~pool_auto_join ~token_timeout ~token_
       assert_pif_prerequisites (pIF,pifrec);
       let ip = ip_of_pif (pIF,pifrec) in
 
+      let assert_pool_on_same_subnet () =
+        let are_on_same_subnet =
+          try
+            let parse_ip ip_str = Scanf.sscanf ip_str "%i.%i.%i.%i" (fun a b c d -> (a, b, c, d)) in
```
lindig: This is clear and simple. But we also have the `Ipaddr` which provides functions for this. I'm not sure what to recommend.

ocaml/xapi/xapi_cluster.ml
```diff
@@ -47,6 +47,29 @@ let create ~__context ~pIF ~cluster_stack ~pool_auto_join ~token_timeout ~token_
...
+            let to_network_addr (a1, b1, c1, d1) =
+              let (a2, b2, c2, d2) = netmask in
+              (a1 land a2, b1 land b2, c1 land c2, d1 land d2)
+            in
+            let pool_host_ips =
+               Xapi_pool_helpers.get_master_slaves_list ~__context |>
+               List.map (fun x -> Db.Host.get_address ~__context ~self:x) |>
+               List.map parse_ip
+            in
+            pool_host_ips |> List.map to_network_addr |> List.for_all (fun x -> x = List.hd pool_host_ips)
```
lindig: Why is this the right test? This compares the masked IP of each slave against the IP of the master? Maybe add a comment.

ocaml/xapi/xapi_cluster.ml
lippirk: i will use that. i assume I will also be able to find a `to_network_address` function

ocaml/xapi/xapi_cluster.ml
lippirk: After googling it looked like to check some IP addresses are on the same subnet, we do a bitwise and of each ip with the netmask and verify that they all give the same answer

I'll add a comment to explain this

ocaml/xapi/dbsync_slave.ml
```diff
@@ -44,6 +44,7 @@ let get_my_ip_addr ~__context =
 
 let create_localhost ~__context info =
   let ip = get_my_ip_addr ~__context in
+  let ip = if ip = "" then "127.0.0.1" else ip in
```
lippirk: just realised this is not in test code, but in prod code, should probably fix this another way 

(the issue is the unit tests do not set up the host with an ip address or a netmask)

ocaml/xapi/xapi_cluster.ml
```diff
@@ -47,6 +47,50 @@ let create ~__context ~pIF ~cluster_stack ~pool_auto_join ~token_timeout ~token_
       assert_pif_prerequisites (pIF,pifrec);
       let ip = ip_of_pif (pIF,pifrec) in
 
+      let assert_pool_on_same_subnet () =
+        (** we gather ip addresses from all the hosts and also the netmask from the PIF.
+          * if all the ip addresses do not have the same network address, then we raise an
+            error *)
+        let are_on_same_subnet =
```
lippirk: unfortunately i think we need to handle ip4/ip6 separately because the `ipaddr` library doesn't give us an interface which is quite as rich when we have an `Ipaddr.t` (as opposed to an `Ipaddr.V4.t` or `Ipaddr.V6.t`)

ocaml/xapi/xapi_cluster.ml
```diff
@@ -47,6 +47,50 @@ let create ~__context ~pIF ~cluster_stack ~pool_auto_join ~token_timeout ~token_
...
+            | `Ip_v4s (netmask, ip::ips) ->
+              let network_addr = Ipaddr.V4.Prefix.of_netmask netmask ip in
+              List.for_all (fun x -> Ipaddr.V4.Prefix.mem x network_addr) ips
+            | `Ip_v6s (netmask, ip::ips) ->
+              let network_addr = Ipaddr.V6.Prefix.of_netmask netmask ip in
+              List.for_all (fun x -> Ipaddr.V6.Prefix.mem x network_addr) ips
+            | `Ip_v4s (_, []) | `Ip_v6s (_, []) -> raise Api_errors.(Server_error (internal_error, ["impossible: pool should contain at least one member"]))
+        in
+        if not are_on_same_subnet then
+          raise Api_errors.(Server_error (internal_error, ["not every host is on the same subnet"]))
```
robhoes: If the error is due to something that the user has done wrong, then it is not an internal error. Instead, define a new API error in the datamodel and api_errors.ml.

ocaml/xapi/xapi_cluster.ml
```diff
@@ -47,6 +47,50 @@ let create ~__context ~pIF ~cluster_stack ~pool_auto_join ~token_timeout ~token_
...
+              (* this shouldn't happen because we shouldn't be storing invalid ips *)
+              | Error (`Msg msg) -> raise Api_errors.(Server_error (internal_error, [Printf.sprintf "could not parse ip address (%s), msg: %s" ip_str msg]))
+            in
+            let pool_host_ips =
+               Xapi_pool_helpers.get_master_slaves_list ~__context |>
+               List.map (fun x -> Db.Host.get_address ~__context ~self:x) |>
+               List.map parse_ip
+            in
+            let ensure_same_protocol ips =
+              let netmask = pifrec.API.pIF_netmask |> parse_ip in
```
robhoes: What PIF is this `pifrec` from? It is one of the host management PIFs, i.e. one of PIFs that have their IP addresses represented in the `host.address` fields that are used in `pool_host_ips`?

ocaml/xapi/dbsync_slave.ml
lindig:
```

File: ocaml/xapi/xapi_cluster.ml (hunk 6 above)
Comment (lippirk) [reply ocaml/xapi/xapi_cluster.ml]: are you saying I should use `master --> netmask instead of pifrec --> netmask`?

the pif comes from the command line

File: ocaml/xapi/dbsync_slave.ml (hunk 3 above)
Comment (psafont) [reply ocaml/xapi/dbsync_slave.ml]: What's the source that tells us if ipv4 or ipv6 should be used? Is assuming  ipv4 always here safe?

File: ocaml/xapi/xapi_cluster.ml (hunk 6 above)
Comment (robhoes) [reply ocaml/xapi/xapi_cluster.ml]: What is the network that the clustering system is using? If it is always the management network of the pool (as this PR is assuming), then why do we get another PIF from the API call? What it is for?

File: ocaml/xapi/xapi_cluster.ml [hunk 7]
```diff
@@ -47,6 +47,50 @@ let create ~__context ~pIF ~cluster_stack ~pool_auto_join ~token_timeout ~token_
...
+              let coerce_v6 xs = xs |> List.map (function
+                | Ipaddr.V6 x -> x
+                | V4 _        -> raise Api_errors.(Server_error (internal_error, ["expected IPv6, but got IPv4"])))
+              in
+              match netmask with
+              | Ipaddr.V4 nm -> `Ip_v4s (nm, coerce_v4 ips)
+              | V6        nm -> `Ip_v6s (nm, coerce_v6 ips)
+            in
+            match ensure_same_protocol pool_host_ips with
+            | `Ip_v4s (netmask, ip::ips) ->
```
robhoes: All this special casing and coercing of v4/v6 feel like something the ipaddr library should be able to abstract and hide from us...?

ocaml/xapi/dbsync_slave.ml
lippirk: I was more getting at the fact that I shouldn't be changing this line for the purpose of a test, as what I'm effectively writing is `let ip = if RUNNING_UNIT_TESTS then "127.0.0.1" else ip in...`. 

The issue that `Net.Interface` is not mocked, so we have bits of code like:

```
let get_primary_ip_addr ~__context iface primary_address_type =
  if iface = "" then
    None
  else
    (* actually get ip address, using Net.Interface *)
```
But now that I have a test which needs to use an IP we are stuck

@psafont  Ip4 is the only option available as a user at the moment, but we have IPv6 support in theory, so we probably shouldn't assume v4

ocaml/xapi/xapi_cluster.ml
lippirk: i mentioned above that the joint interface isn't as powerful (as far as i can tell).

with the joint v4/v6 interface you can't provide a complete ip address as a netmask, only using as a single integer; and I assume we have to support arbitrary netmasks (at least in the ipv4 case)?

ocaml/xapi/xapi_cluster.ml
lippirk: it appears that the only purpose of the PIF is to provide the `local_ip` address for the `init_config` which is then passed to clusterd - so this is clearly the right network. I'm not sure if getting the addresses from the master/slaves list is correct though

lippirk: @edwintorok do you know if xapi knows about the IP addresses of the cluster network? perhaps the check instead needs to be made in xapi-clusterd

lippirk:
I think I am get IPs from the correct network now

I still need to run some tests

ocaml/tests/test_helpers.ml
```diff
@@ -284,10 +284,40 @@ module CIDRCheckers = Generic.MakeStateless (struct
     ]
   end)
 
+module AreOnSameSubnetTests = Generic.MakeStateless (struct
+  module Io = struct
+    type input_t = string * string list
```
lindig: Minor style issue: I don't think types need to be marked with a `_t` suffix. They live in their own name space and can't collide with value names. It is a customary to use `t` for the central type of a module, though (like in `Stack.t`).

ocaml/xapi/helpers.ml
```diff
@@ -1230,3 +1230,32 @@ let host_supports_hvm ~__context host =
 let env_with_path env_vars =
   Array.of_list @@
   List.map (fun (k,v) -> Printf.sprintf "%s=%s" k v) (("PATH", String.concat ":" Forkhelpers.default_path)::env_vars)
+
+let are_on_same_subnet_exn netmask ips =
```
lindig: Would a `bool` as a result not be as good? The name indicates that this a predicate. On the client side we would have 
```
match Helpers.are_on_same_subnet_exn netmask ips with
| true ->
| false ->
```

(or using an `if`) - I don't think there could be confusion of the meaning with the name chosen as is.

ocaml/xapi/helpers.ml
```diff
@@ -1230,3 +1230,32 @@ let host_supports_hvm ~__context host =
...
+  let ensure_same_protocol ips =
+    let coerce_v4 xs = xs |> List.map (function
+      | Ipaddr.V4 x -> x
+      | V6 _        -> raise Api_errors.(Server_error (internal_error, ["expected IPv4, but got IPv6"])))
```
lindig: Maybe include the culprit in the error message?

ocaml/xapi/helpers.ml
```diff
@@ -1230,3 +1230,32 @@ let host_supports_hvm ~__context host =
...
+      | V4 _        -> raise Api_errors.(Server_error (internal_error, ["expected IPv6, but got IPv4"])))
+    in
+    match netmask with
+    | Ipaddr.V4 nm -> `Ip_v4s (nm, coerce_v4 ips)
+    | V6        nm -> `Ip_v6s (nm, coerce_v6 ips)
+  in
+  match ensure_same_protocol ips with
+  | `Ip_v4s (netmask, ip::ips) ->
+    let network_addr = Ipaddr.V4.Prefix.of_netmask netmask ip in
+    if List.for_all (fun x -> Ipaddr.V4.Prefix.mem x network_addr) ips then
```
lindig: When returning a bool, this would be come simpler as we would have boolean expressions.

edwintorok:
Why are all the python changes showing up in this PR? Can you try rebasing on top of latest master?

ocaml/xapi/helpers.ml
```diff
@@ -1230,3 +1230,32 @@ let host_supports_hvm ~__context host =
...
+      | Ipaddr.V6 x -> x
+      | V4 _        -> raise Api_errors.(Server_error (internal_error, ["expected IPv6, but got IPv4"])))
+    in
+    match netmask with
+    | Ipaddr.V4 nm -> `Ip_v4s (nm, coerce_v4 ips)
+    | V6        nm -> `Ip_v6s (nm, coerce_v6 ips)
+  in
+  match ensure_same_protocol ips with
+  | `Ip_v4s (netmask, ip::ips) ->
+    let network_addr = Ipaddr.V4.Prefix.of_netmask netmask ip in
```
edwintorok: The netmask is not guaranteed to be the same, especially if they come from different subnets.
Perhaps all these checks could be done by intersecting the netmasks instead?

ocaml/tests/test_helpers.ml
lippirk: i think adding the `_t` helps with readability, since we use snake case for both types and variable names

psafont: `input_t` and `output_t` seem to be a convention in the tests of xapi, I found 128 uses:
```
$ rg 'type [^ ]+put_t' -c | cut -d ":" -f 2 | jq -s 'add'
128
```

lippirk: to would be easy to change (you just need to change the signature of the `MakeStateless` functor, and the compiler will tell you what needs fixing), but i can't see it being worthwhile

ocaml/xapi/helpers.ml
```diff
@@ -1230,3 +1230,40 @@ let host_supports_hvm ~__context host =
...
+  let and_ipv6 x y =
+    let (x_ho, x_lo) = Ipaddr.V6.to_int64 x in
+    let (y_ho, y_lo) = Ipaddr.V6.to_int64 y in
+    Ipaddr.V6.of_int64 (Int64.logand x_ho y_ho, Int64.logand x_lo y_lo)
+  in
+  let and_ipv4 x y = Int32.logand (Ipaddr.V4.to_int32 x) (Ipaddr.V4.to_int32 y) |> Ipaddr.V4.of_int32 in
+  match ensure_same_protocol ips with
+  | `Ip_v4s (ip::ips) -> Ipaddr.V4 (List.fold_left and_ipv4 ip ips)
+  | `Ip_v6s (ip::ips) -> Ipaddr.V6 (List.fold_left and_ipv6 ip ips)
+  | `Ip_v4s [] | `Ip_v6s [] -> failwith "impossible"
```
lindig: Minor comment: this case is not impossible in the sense that the program can never reach this state. Instead, you assume the list of IPs is never empty. I think a better error would be something like `list of IPs must not be empty` or similar.

ocaml/xapi/xapi_clustering.ml
```diff
@@ -154,6 +154,24 @@ let assert_operation_host_target_is_localhost ~__context ~host =
   if host <> Helpers.get_localhost ~__context then
     raise Api_errors.(Server_error (internal_error, [ "A clustering operation was attempted from the wrong host" ]))
 
+(* a new member of a cluster should reside on the same network as the cluster *)
+let assert_same_subnet ~__context ~cluster ~pIF =
+  let parse_ip ip_str =
+    match Ipaddr.of_string ip_str with
+    | Ok ip_addr       -> ip_addr
+    (* this shouldn't happen because we shouldn't be storing invalid ips *)
+    | Error (`Msg msg) -> raise Api_errors.(Server_error (internal_error, [Printf.sprintf "could not parse ip address (%s), msg: %s" ip_str msg]))
```
lindig: Something like this could make this a bit easier:

```
let internal_error fmt = 
  Printf.kprintf (fun msg -> raise Api_errors.(Server_error (internal_error, msg)))
...
internal_error "could not parse IP address %s: %s" ip_str msg

File: ocaml/xapi/xapi_clustering.ml [hunk 15]
```diff
@@ -154,6 +154,24 @@ let assert_operation_host_target_is_localhost ~__context ~host =
...
+  in
+  let ip_of_pif pIF = ip_of_pif (pIF, Db.PIF.get_record ~__context ~self:pIF) |> Cluster_interface.str_of_address |> parse_ip in
+  let netmask_of_pif pIF = Db.PIF.get_netmask ~__context ~self:pIF |> parse_ip in
+
+  let network = get_network_internal ~__context ~self:cluster in
+  let pifs = Db.Network.get_PIFs ~__context ~self:network |> List.cons pIF in
+  let ips = List.map ip_of_pif pifs in
+  let netmask = List.map netmask_of_pif pifs |> Helpers.intersect_ips_exn in
+  if not (Helpers.are_on_same_subnet_exn netmask ips) then
```
edwintorok: What if you have a PIF with netmask 255.255.255.255 (i.e. a /32 address) and IP 192.168.0.2,
and another PIF with netmask 255.255.255.0 (i.e. a /24 address) and IP 192.168.0.3.
When you intersect the netmasks I think you get 255.255.255.0 and conclude they are in the same subnet.
I think you should have a function for intersecting netmasks (using OR), or if you want to use the one for IPs then you should first invert the netmask, do the and, and then invert again.

Perhaps it'd be easier to avoid intersecting in the first place: you have one `pIF` to check, which has an IP address: do a `List.for_all` and check that this IP is a member of all the subnets that already exist (using Ipaddr.Prefix.mem).
That would also avoid having to handle IPv4 vs IPv6 separately.

robhoes: Yes, I don't think that intersecting netmasks is the right thing to do. If the constraint is that hosts need to be on the same subnet, then that means that the network prefixes (derived from IP + netmask) for all hosts are identical.

ocaml/idl/datamodel_errors.ml
```diff
@@ -1165,6 +1165,8 @@ let _ =
     ~doc:"The last cluster host cannot be destroyed. Destroy the cluster instead" ();
   error Api_errors.no_cluster_hosts_reachable ["cluster"]
     ~doc:"No other cluster host was reachable when joining" ();
+  error Api_errors.cluster_inconsistent_subnet []
```
robhoes: You could consider adding some helpful arguments to the error.

ocaml/xapi/xapi_clustering.ml
robhoes: For the case of adding a new host to an existing cluster, you could assume that the current cluster host are already on the same subnet. So just pick one and compare its subnet to the new host's?

lindig:
This has been lingering here for a while. Do we have a plan?

edwintorok:
There are some possibilities described at https://github.com/xapi-project/xen-api/pull/4000#discussion_r358163216, although we may have more important bugs to work on, this is a minor.

PR #4001, opened 2019-11-28
CA-325940 SRIOV switching between vGPU/pass-through

lindig:
When trying to allocate a PCI VF for an NVidia SRIOV GPU, xapi looks
into its database. The database may contain an entry for a VF that
actually does not exist. This happens, when previously the corresponding
PCI PF was used in pass-through mode. Therefore, xapi needs to call
nvidia_vf_setup in every case. The setup function (which runs on the
relevant host) checks for the prensence of VFs before activating them.

Xapi calls Nvidia's sriov-manage on a PCI to activate PCI VFs. If
previously the PCI device was used as pass-through, it is still bound to
the pciback driver. It needs to be unbound and bound to the "nvidia"
driver first. This commit adds code to do so.

These commits come with corresponding changes in xenopsd but they can be merged
independently.

### DIFF
--- a/ocaml/xapi/vgpuops.ml
+++ b/ocaml/xapi/vgpuops.ml
@@ -100,11 +100,11 @@ let add_pcis_to_vm ~__context host vm vgpu pci =
     Db.VM.add_to_platform ~__context ~self:vm ~key:Xapi_globs.igd_passthru_key ~value:"true"
 
 let reserve_free_virtual_function ~__context vm impl pf =
+  let host = Db.PCI.get_host ~__context ~self:pf in
   let rec get retry =
     match Pciops.reserve_free_virtual_function ~__context vm pf with
     | Some vf -> vf
     | None when retry ->
-        let host = Db.PCI.get_host ~__context ~self:pf in
         begin match impl with
         | `mxgpu ->
           Helpers.call_api_functions ~__context @@ fun rpc session_id ->
@@ -125,7 +125,17 @@ let reserve_free_virtual_function ~__context vm impl pf =
         (* This probably means that our capacity checking went wrong! *)
         raise Api_errors.(Server_error (internal_error, ["No free virtual function found"]))
   in
-  get true
+  match impl with
+  | `nvidia_sriov ->
+    (* Always call the setup function. If we are coming out of
+     * pass-through, xapi assumes the VFs are active but they are not - so
+     * activate them. The setup function itself avoids unnecessary
+     * work *)
+    Helpers.call_api_functions ~__context @@ fun rpc session_id ->
+        Client.Client.Host.nvidia_vf_setup ~rpc ~session_id ~host ~pf
+        ~enable:true;
+    get false
+  | _ -> get true
 
 let add_vgpus_to_vm ~__context host vm vgpus =
   debug "vGPUs allocated to VM (%s) are: %s" (Ref.string_of vm)
--- a/ocaml/xapi/xapi_pgpu.ml
+++ b/ocaml/xapi/xapi_pgpu.ml
@@ -366,18 +366,76 @@ let mxgpu_vf_setup ~__context =
 
 (* This must be run LOCALLY on the host that is about to start a VM that is
  * going to use a vgpu backed by an nvidia pgpu. *)
+let nvidia_vf_setup_mutex = Mutex.create ()
 let nvidia_vf_setup ~__context ~pf ~enable =
-  let script  = !Xapi_globs.nvidia_sriov_manage_script in
-  let enable' = if enable then "-e" else "-d" in
-  if Sys.file_exists script then begin
-    let id =
-      if pf = Ref.null then "ALL" else Db.PCI.get_pci_id ~__context ~self:pf in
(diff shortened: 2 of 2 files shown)


ocaml/xapi/vgpuops.ml
```diff
@@ -125,7 +125,17 @@ let reserve_free_virtual_function ~__context vm impl pf =
...
   in
-  get true
+  match impl with
+  | `nvidia_sriov ->
+    (* Always call the setup function. If we are coming out of
+     * pass-through, xapi assumes the VFs are active but they are not - so
+     * activate them. The setup function itself avoids unnecessary
+     * work *)
+    Helpers.call_api_functions ~__context @@ fun rpc session_id ->
+        Client.Client.Host.nvidia_vf_setup ~rpc ~session_id ~host ~pf
```
edwintorok: I thought that it would be better to call the Xapi_host method directly here, but you might need some additional locking (see below), so this is probably better as it is written currently.

ocaml/xapi/xapi_pgpu.ml
```diff
@@ -367,17 +367,76 @@ let mxgpu_vf_setup ~__context =
 (* This must be run LOCALLY on the host that is about to start a VM that is
  * going to use a vgpu backed by an nvidia pgpu. *)
 let nvidia_vf_setup ~__context ~pf ~enable =
```
edwintorok: Does this function need some locking? e.g. in message-forwarding
Although the actions performed by it are idempotent you could have 2 threads racing to enable the extra VFs when number of VFs is zero, which may not work well.

robhoes: I agree, especially now that multiple VM-starts can get here at the same time. I think that a local mutex in this function would suffice.

ocaml/xapi/xapi_pgpu.ml
```diff
@@ -367,17 +367,76 @@ let mxgpu_vf_setup ~__context =
...
+  let sprintf   = Printf.sprintf in
+  let fail msg  = Api_errors.(Server_error (internal_error, [msg])) in
+  let script    = !Xapi_globs.nvidia_sriov_manage_script in
+  let enable'   = if enable then "-e" else "-d" in
+  let bind_path = "/sys/bus/pci/drivers/nvidia/bind" in
+  let unbind_path pci = sprintf "/sys/bus/pci/devices/%s/driver/unbind" pci in
+  let driver_path pci = sprintf "/sys/bus/pci/devices/%s/driver" pci in
+  let pci       = Db.PCI.get_pci_id ~__context ~self:pf in
+
+  let driver_of pci =
```
robhoes: There is also `Xapi_pci_helpers.get_driver_name`.

ocaml/xapi/xapi_pgpu.ml
```diff
@@ -367,17 +367,76 @@ let mxgpu_vf_setup ~__context =
...
+    try
+      let fn fd = Unixext.really_write fd pci 0 (String.length pci) in
+      Unixext.with_file path [ Unix.O_WRONLY ] 0o640 fn
+    with
+      e ->
+        error "failed to write to %s to re-bind PCI %s to Nvidia driver: %s"
+          path pci (Printexc.to_string e);
+        raise (fail (sprintf "Can't rebind PCI %s driver" pci)) in
+
+  begin match driver_of pci with
```
robhoes: It may be clearer to put this match statement in a `bind` function, and the one below in an `activate` function, and then have a three-line "program" at the bottom.

robhoes:
Regarding the code duplication with xenopsd, from an architectural point of view, I think that we should move all PCI device management from xenopsd to xapi. Xenopsd is meant to set up the virtual side of things, and when it is asked by xapi to (e.g.) start a VM, then the required physical environment should have already been prepared. In the same way, xapi ensures that networks are configured (PIF plugged) before calling xenopsd.

ocaml/xapi/vgpuops.ml
robhoes: @edwintorok, the functions in this module always run on the master, while the nvidia setup needs to be done on the host that will run the VM. So we have to go through the API forwarding in any case.

lindig:
I've addressed the comments and the locking - please take another look.

ocaml/xapi/xapi_pgpu.ml
```diff
@@ -366,18 +366,73 @@ let mxgpu_vf_setup ~__context =
...
+      write_to bind_path pci in
+
+  let activate_vfs pci =
+    match num_vfs pci with
+    | None ->
+      let msg = sprintf "Can't determine number of VFs for PCI %s" pci in
+      raise (fail msg)
+    | Some 0 when Sys.file_exists script ->
+      debug "PCI %s has 0 VFs - calling %s" pci script;
+      Mutex.execute nvidia_vf_setup_mutex @@ fun () ->
```
robhoes: I think that the mutex needs to be wider: get the lock _before_ assessing whether you need to call the enable script, i.e. before counting the VFs.

PR #4002, opened 2019-11-29
Allow guests to see floppies

lippirk:
This is a backport of CA-330162 to 8.1.

It includes what I hope is a fix for failing BVTs.

Needs to go in alongside https://github.com/xapi-project/xenopsd/pull/691

### DIFF
--- a/ocaml/xapi/cli_operations.ml
+++ b/ocaml/xapi/cli_operations.ml
@@ -1338,8 +1338,8 @@ let vbd_create printer rpc session_id params =
   let _type =
     if List.mem_assoc "type" params
     then match String.lowercase_ascii (List.assoc "type" params) with
-      | "cd" -> `CD | "disk" -> `Disk
-      | x -> failwith (Printf.sprintf "Unknown type: %s (should be \"cd\" or \"disk\"" x)
+      | "cd" -> `CD | "disk" -> `Disk | "floppy" -> `Floppy
+      | x -> failwith (Printf.sprintf {|"Unknown type: %s (should be "cd", "disk" or "floppy"|} x)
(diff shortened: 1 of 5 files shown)


lindig:
This fixup is in response to a test failure. It is not obvious to me why floppies are not an allowed VBD devices.

edwintorok:
Before merging to LCM we should fix master

edwintorok:
I think the API for `get_allowed_VBD_devices` is a bit incomplete since it does not specify disk type.
Lets say you want to create a new Disk, so you query the allowed devices and use one of the responses. If that includes floppy devices then you try to use a floppy device to create a disk device, which won't work.
Similarly if you want to use get_allowed_VBD_devices to create a floppy you're out of luck, because there is no such API. You can only have 2 floppy slots though, and very likely you'd be using just the 1st one so I wouldn't modify the API just to support a 2nd floppy disk more easily.

lindig:
We will cherry-pick the fixup into the master branch after we have merged this. So no extra PR for master is required.

lippirk:
> I think the API for `get_allowed_VBD_devices` is a bit incomplete since it does not specify disk type.
> Lets say you want to create a new Disk, so you query the allowed devices and use one of the responses. If that includes floppy devices then you try to use a floppy device to create a disk device, which won't work.
> Similarly if you want to use get_allowed_VBD_devices to create a floppy you're out of luck, because there is no such API. You can only have 2 floppy slots though, and very likely you'd be using just the 1st one so I wouldn't modify the API just to support a 2nd floppy disk more easily.

I think this is worth doing if we add a regression test for floppies, but we can do this at the time

PR #4003, opened 2019-11-29
CA-325582: Unload SR metrics from memory on SR.forget and SR.destroy

psafont:
The approach is to reuse the SR.forget and SR.destroy API calls so that they call xcp-rrdd in _all_ the hosts with PBDs from the SR to remove the metrics from memory; while making sure their actual functionality keeps being executed by a single host, like they were before.

I don't think that the instructions are executed always in the same hosts as they were before, the intention is to do so, those check will probably change.

### DIFF
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -3295,16 +3295,18 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
       let local_forget = Local.SR.forget ~sr in
       with_sr_marked ~__context ~sr ~doc:"SR.destroy" ~op:`destroy
         (fun () ->
-          Xapi_sr.assert_all_pbds_unplugged ~__context ~sr;
+          let pbd_state = Xapi_sr.assert_all_pbds_unplugged ~__context ~sr in
           Xapi_sr.assert_sr_not_indestructible ~__context ~sr;
           Xapi_sr.assert_sr_not_local_cache ~__context ~sr;
 
           forward_sr_op ~consider_unplugged_pbds:true ~local_fn:local_destroy ~__context ~self:sr
             (fun session_id rpc -> Client.SR.destroy rpc session_id sr);
 
-          (* forward call to unload SR metrics from memory *)
-          forward_sr_all_op ~local_fn:local_forget ~__context ~self:sr
-            (fun session_id rpc -> Client.SR.forget rpc session_id sr);
+          match pbd_state with
+          | `No_pbds            -> (* no SR metrics to remove *) ()
+          | `All_pbds_unplugged ->
+            forward_sr_all_op ~local_fn:local_forget ~__context ~self:sr
+              (fun session_id rpc -> Client.SR.forget rpc session_id sr);
 
           (* don't forward this is just a db call *)
           Xapi_sr.really_forget ~__context ~sr
@@ -3315,13 +3317,13 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
       let local_fn = Local.SR.forget ~sr in
       with_sr_marked ~__context ~sr ~doc:"SR.forget" ~op:`forget
         (fun () ->
-          Xapi_sr.assert_all_pbds_unplugged ~__context ~sr;
+          match Xapi_sr.assert_all_pbds_unplugged ~__context ~sr with
+          | `No_pbds            -> (* there are no SR metrics to remove *) ()
+          | `All_pbds_unplugged -> (* unload SR metrics from memory *)
+            forward_sr_all_op ~local_fn ~__context ~self:sr
+              (fun session_id rpc -> Client.SR.forget rpc session_id sr);
 
-          (* forward call to unload SR metrics from memory *)
-          forward_sr_all_op ~local_fn ~__context ~self:sr
-            (fun session_id rpc -> Client.SR.forget rpc session_id sr);
-
-          (* don't forward this is just a db call *)
+          (* don't forward - this is just a db call *)
           Xapi_sr.really_forget ~__context ~sr
           )
 
@@ -3944,8 +3946,13 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
     let destroy ~__context ~self =
       info "PBD.destroy: PBD '%s'" (pbd_uuid ~__context self);
       let sr = Db.PBD.get_SR ~__context ~self in
+      let host = Db.PBD.get_host ~__context ~self in
       SR.with_sr_marked ~__context ~sr ~doc:"PBD.destroy" ~op:`pbd_destroy
-        (fun () -> Local.PBD.destroy ~__context ~self)
+        (fun () -> (* Forget metrics on the host which owns that SR. This is safe
+                      because on a host there is at most one pbd per sr *)
+                   do_op_on ~local_fn:(Local.SR.forget ~sr) ~__context ~host
+                     (fun session_id rpc -> Client.SR.forget rpc session_id sr);
+                   Local.PBD.destroy ~__context ~self)
 
     (* -------- Forwarding helper functions: ------------------------------------ *)
 
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -161,25 +161,29 @@ let log_exn_ignore ?(doc = "performing unknown operation") f x =
 
 
 (* Given an SR, return a PBD to use for some storage operation. *)
-(* In the case of SR.destroy we need to be able to forward the SR operation when all
-   PBDs are unplugged - this is the reason for the consider_unplugged_pbds optional
-   argument below. All other SR ops only consider plugged PBDs... *)
-let choose_pbd_for_sr ?(consider_unplugged_pbds=false) ~__context ~self () =
+(* In the case of SR.destroy and forget we need to be able to forward the SR
+   operation when all PBDs are unplugged. This is the reason for the
+   consider_unplugged_pbds optional argument below. All other SR ops only
+   consider plugged PBDs. *)
+let choose_pbds_for_sr ?(consider_unplugged_pbds=false) ~__context ~self () =
   let all_pbds = Db.SR.get_PBDs ~__context ~self in
-  let plugged_pbds = List.filter (fun pbd->Db.PBD.get_currently_attached ~__context ~self:pbd) all_pbds in
+  let plugged_pbds = List.filter (fun self -> Db.PBD.get_currently_attached ~__context ~self) all_pbds in
   let pbds_to_consider = if consider_unplugged_pbds then all_pbds else plugged_pbds in
-  if Helpers.is_sr_shared ~__context ~self then
-    let master = Helpers.get_master ~__context in
-    let master_pbds = Db.Host.get_PBDs ~__context ~self:master in
-    (* shared SR operations must happen on the master *)
-    match Listext.intersect pbds_to_consider master_pbds with
-    | pbd :: _ -> pbd (* ok, master plugged *)
-    | [] -> raise (Api_errors.Server_error(Api_errors.sr_no_pbds, [ Ref.string_of self ])) (* can't do op, master pbd not plugged *)
-  else
-    match pbds_to_consider with
-    | [] -> raise (Api_errors.Server_error(Api_errors.sr_no_pbds, [ Ref.string_of self ]))
-    | pdb :: _ -> pdb
+  let pbd_candidates =
+    if Helpers.is_sr_shared ~__context ~self then
+      let master = Helpers.get_master ~__context in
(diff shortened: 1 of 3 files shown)


psafont:
Github is ordering the commits chronological creation, not by precedence order :(

ocaml/xapi/message_forwarding.ml
```diff
@@ -161,25 +161,29 @@ let log_exn_ignore ?(doc = "performing unknown operation") f x =
...
-   argument below. All other SR ops only consider plugged PBDs... *)
-let choose_pbd_for_sr ?(consider_unplugged_pbds=false) ~__context ~self () =
+(* In the case of SR.destroy and forget we need to be able to forward the SR
+   operation when all PBDs are unplugged. This is the reason for the
+   consider_unplugged_pbds optional argument below. All other SR ops only
+   consider plugged PBDs. *)
+let choose_pbds_for_sr ?(consider_unplugged_pbds=false) ~__context ~self () =
   let all_pbds = Db.SR.get_PBDs ~__context ~self in
-  let plugged_pbds = List.filter (fun pbd->Db.PBD.get_currently_attached ~__context ~self:pbd) all_pbds in
+  let plugged_pbds = List.filter (fun self -> Db.PBD.get_currently_attached ~__context ~self) all_pbds in
```
lippirk: perhaps write a helper for this? looks like it is used in multiple places, e.g. 

https://github.com/xapi-project/xen-api/blob/74412d584c9e49669256c3243ccdbdb1a84eff0c/ocaml/xapi/agility.ml#L28

ocaml/xapi/message_forwarding.ml
```diff
@@ -3226,6 +3230,16 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
         with _ -> raise Not_found in
       do_op_on ~local_fn ~__context ~host op
 
+    (* Forward SR operation to all hosts that have suitable, plugged or
+       unplugged PBDs and the pool master *)
+    let forward_sr_all_op ~local_fn ~__context ~self op =
+      let module HostSet = Set.Make(struct type t = API.ref_host let compare = Pervasives.compare end) in
+      choose_pbds_for_sr ~consider_unplugged_pbds:true ~__context ~self ()
+      |> List.map (fun pbd -> Db.PBD.get_host ~__context ~self:pbd)
+      |> List.cons (Helpers.get_master ~__context)
```
lippirk: it looks like you are using a set just so you can cons the master? i think you should just do this instead:

```
let pbds_for_sr = if List.mem master pbds_for_sr then pbds_for_sr else master :: pbds_for_sr in
```

if `choose_pbds_for_sr` does indeed return duplicates, should it?

ocaml/xapi/message_forwarding.ml
```diff
@@ -3238,6 +3252,34 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
+          pbd.API.pBD_SR = sr && pbd.API.pBD_currently_attached)
+      in
+
+      if any_pbds_attached_to_sr then
+        raise (Api_errors.Server_error(Api_errors.sr_has_pbd, [ Ref.string_of sr ]))
+
+    let assert_sr_not_indestructible ~__context ~sr =
+      let oc = Db.SR.get_other_config ~__context ~self:sr in
+
+      match List.assoc_opt "indestructible" oc with
```
lippirk: do we have a file containing consts for this sort of thing?

ocaml/xapi/message_forwarding.ml
psafont: I guess it's simpler that way, with equivalent cost.

ocaml/xapi/message_forwarding.ml
psafont: Not sure it's worth doing it in this PR, it's a pattern that's repeated a lot in xapi. Could be tackled all in a another PR where it can be treated more systematically, it might need more considerations.

ocaml/xapi/message_forwarding.ml
```diff
@@ -161,25 +161,29 @@ let log_exn_ignore ?(doc = "performing unknown operation") f x =
...
+      let master = Helpers.get_master ~__context in
+      let master_pbds = Db.Host.get_PBDs ~__context ~self:master in
+      (* shared SR operations must happen on the master *)
+      Listext.intersect pbds_to_consider master_pbds
+    else
+      pbds_to_consider
+  in
+  match pbd_candidates with
+  | [] -> raise (Api_errors.Server_error(Api_errors.sr_no_pbds, [ Ref.string_of self ]))
+  | pdbs -> pdbs
```
edwintorok: Did you mean `pbds`?

ocaml/xapi/message_forwarding.ml
```diff
@@ -161,25 +161,29 @@ let log_exn_ignore ?(doc = "performing unknown operation") f x =
...
 
+let choose_pbd_for_sr ?consider_unplugged_pbds ~__context ~self () =
+  List.hd @@ choose_pbds_for_sr ?consider_unplugged_pbds ~__context ~self ()
```
edwintorok: Shouldn't this need the old wrapper for empty lists to raise `Api_errors.sr_no_pbds`? This would raise an internal error from List instead now.

ocaml/xapi/message_forwarding.ml
robhoes: Yes: xapi_globs.ml.

ocaml/xapi/message_forwarding.ml
```diff
@@ -3270,15 +3312,22 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
-    (* don't forward this is just a db call *)
     let forget ~__context ~sr =
       info "SR.forget: SR = '%s'" (sr_uuid ~__context sr);
+      let local_fn = Local.SR.forget ~sr in
       with_sr_marked ~__context ~sr ~doc:"SR.forget" ~op:`forget
         (fun () ->
-           Local.SR.forget ~__context ~sr)
+          assert_all_pbds_unplugged ~__context ~sr;
+
+          forward_sr_all_op ~local_fn ~__context ~self:sr
```
edwintorok: What happens if a host is offline/unreachable in the pool? Should we make this function best-effort, or should we have it fail if it cannot reach all hosts?

ocaml/xapi/message_forwarding.ml
```diff
@@ -3238,6 +3238,34 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
       Sm.assert_session_has_internal_sr_access ~__context ~sr:self;
       Local.SR.set_physical_utilisation ~__context ~self ~value
 
+    let assert_all_pbds_unplugged ~__context ~sr =
```
robhoes: I would prefer to keep the assert function definitions themselves in xapi_sr.ml, and just call them from message_forwarding.ml.

robhoes: Or xapi_sr_operations.ml perhaps.

ocaml/xapi/message_forwarding.ml
psafont: Protein Data Banks where a thing in my previous job =_=

ocaml/xapi/message_forwarding.ml
robhoes: `List.hd` is generally to be avoided. A match to handle the `[]` case is better. What's the use of redefining the function anyway?

ocaml/xapi/message_forwarding.ml
robhoes: We are already on the master, so why would we forward the call to the master?

ocaml/xapi/message_forwarding.ml
psafont: `sr_no_pbds` is being raised in the case where the list is empty (the match is right above this function), making it impossible to see an internal error from list. Would you rather have the match to be repeated?

ocaml/xapi/message_forwarding.ml
psafont: I can try removing the master from the list, if it's present (that's the case of shared SRs) and directly perform the operation on master after all the calls are forwarderded. It will make the function have caveats and will be more difficult to use, however.

psafont: I can try removing the master from the list, if it's present (that's the case of shared SRs) and directly perform the operation on master after all the calls are forwarderded. It will make the function have caveats and will be more difficult to use, however.

robhoes: Probably not worth it.

ocaml/xapi/message_forwarding.ml
psafont: I opted for the second option, it looks like the functions in in `xapi_sr.ml` are only accessible if they are exposed through the datamodel api.

ocaml/xapi/message_forwarding.ml
psafont: I think it's needed to ensure the SR is not deleted from the database by the master; before the Database call in `destroy` happens in the SR master.

Where the operations are run in branch `master`:

| Op      | Shared | Not Shared (Master) | Not Shared (Slave)
| :------ | -----: | ------------------: | -----------------:
| Destroy | master | master              | slave
| Forget  | master | master              | master

Right now this branch doesn't behave like that. I'm changing xapi_sr destroy so it doesn't depend on forget and can accomplish the behaviour in the table. Then I'll adapt the selection mechanism based on that.

For reference, this is the desired behaviour:

| Op      | Shared | Not Shared (Master) | Not Shared (Slave)
| :------ | -----: | ------------------: | -----------------:
| Destroy | master | master              | slave
| Forget  | master | master              | master
| RRD rm  | all    | all                 | all

ocaml/xapi/message_forwarding.ml
psafont: I think that would be best, we don't want the operation to fail while it was successful before.
This will mean that if the forwarding fails for one host, the behaviour of those SR metrics will be the same as before, so it won't introduce a regression, even if it's not ideal.

Customers still have the option to manually delete the metrics if all else fails.

ocaml/xapi/message_forwarding.ml
psafont: Latest change removes the need to add the pool master to the list: if the SR is shared or if the SR is managed by the master it will already contain the reference to master. in the other case the RRD deletion does not need to be forwarded to it.

In any case if master is present the call will still be local because of the logic in `do_op_on`.

ocaml/xapi/message_forwarding.ml
```diff
@@ -161,25 +161,29 @@ let log_exn_ignore ?(doc = "performing unknown operation") f x =
...
-  else
-    match pbds_to_consider with
-    | [] -> raise (Api_errors.Server_error(Api_errors.sr_no_pbds, [ Ref.string_of self ]))
-    | pdb :: _ -> pdb
+  let pbd_candidates =
+    if Helpers.is_sr_shared ~__context ~self then
+      let master = Helpers.get_master ~__context in
+      let master_pbds = Db.Host.get_PBDs ~__context ~self:master in
+      (* shared SR operations must happen on the master *)
+      Listext.intersect pbds_to_consider master_pbds
```
robhoes: So this function will only return the master PBD for shared SRs (as before). But that isn't good enough for the new use case, is it?

psafont:
With the latest fixup a list of all the pbds in all slaves from an SR.
I've used set instead of list for the worst-case scenario for the partition with 64 (pbds of the SR, 1 per host) * the number of PBDs on master operations.

Time to test it.

psafont:
Last commit looks good it's on the good path. PBD.destroy should to call `SR.forget` in the host where the pbd is destroyed so the metrics are still cleaned up.

ocaml/xapi/message_forwarding.ml
```diff
@@ -3899,8 +3946,13 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
       info "PBD.destroy: PBD '%s'" (pbd_uuid ~__context self);
       let sr = Db.PBD.get_SR ~__context ~self in
+      let host = Db.PBD.get_host ~__context ~self in
       SR.with_sr_marked ~__context ~sr ~doc:"PBD.destroy" ~op:`pbd_destroy
-        (fun () -> Local.PBD.destroy ~__context ~self)
+        (fun () -> (* Forget metrics on the host which owns that SR. This is safe
+                      because on a host there is at most one pbd per sr *)
+                   do_op_on ~local_fn:(Local.SR.forget ~sr) ~__context ~host
+                     (fun session_id rpc -> Client.SR.forget rpc session_id sr);
+                   Local.PBD.destroy ~__context ~self)
```
lippirk: > Last commit looks good it's on the good path. PBD.destroy should to call `SR.forget` in the host where the pbd is destroyed so the metrics are still cleaned up.

@psafont pretty sure it does here?

psafont: Pretty sure I need new glasses :)

PR #4004, opened 2019-12-04
CP-28369: remove unused daemonize code

psafont:
Systemd manages daemonization of mpathalert


psafont:
Note that currently the service file was launching the command with only the `-delay` parameter: https://github.com/xapi-project/xen-api/blob/master/scripts/mpathalert.service

PR #4007, opened 2019-12-20
Ca 325582 clean up dead SR metrics

lippirk:
Make sure data sources for an SR are not kept in memory in xcp-rrdd when the SR is destroyed or forgotten.

### DIFF
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -167,9 +167,9 @@ let log_exn_ignore ?(doc = "performing unknown operation") f x =
 (* Given an SR, return a PBD to use for some storage operation. *)
 (* In the case of SR.destroy and forget we need to be able to forward the SR
    operation when all PBDs are unplugged. This is the reason for the
-   consider_unplugged_pbds optional argument below. All other SR ops only
+   consider_unplugged_pbds argument below. All other SR ops only
    consider plugged PBDs. *)
-let choose_pbds_for_sr ?(consider_unplugged_pbds=false) ~__context ~self () =
+let choose_pbds_for_sr ~consider_unplugged_pbds ~__context ~self () =
   let module R = Stdlib.Result in
   let all_pbds = Db.SR.get_PBDs ~__context ~self in
   let plugged_pbds = List.filter (fun self -> Db.PBD.get_currently_attached ~__context ~self) all_pbds in
@@ -193,8 +193,8 @@ let choose_pbds_for_sr ?(consider_unplugged_pbds=false) ~__context ~self () =
   | []   -> Error `Sr_no_pbds
   | pbds -> Ok pbds)
 
-let choose_pbd_for_sr ?consider_unplugged_pbds ~__context ~self () =
-  match choose_pbds_for_sr ?consider_unplugged_pbds ~__context ~self () with
+let choose_pbd_for_sr ~consider_unplugged_pbds ~__context ~self () =
+  match choose_pbds_for_sr ~consider_unplugged_pbds ~__context ~self () with
   | Ok (x::_)         -> x
   | Ok []             -> failwith "expected 'choose_pbds_for_sr' to return a non-empty list"
   | Error `Sr_no_pbds -> raise (Api_errors.Server_error(Api_errors.sr_no_pbds, [ Ref.string_of self ]))
@@ -1528,7 +1528,7 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
                if Db.VM.get_power_state ~__context ~self:snapshot = `Suspended then begin
                  let suspend_VDI = Db.VM.get_suspend_VDI ~__context ~self:snapshot in
                  let sr = Db.VDI.get_SR ~__context ~self:suspend_VDI in
-                 let pbd = choose_pbd_for_sr ~__context ~self:sr () in
+                 let pbd = choose_pbd_for_sr ~consider_unplugged_pbds:false ~__context ~self:sr () in
                  let host = Db.PBD.get_host ~__context ~self:pbd in
                  let metrics = Db.Host.get_metrics ~__context ~self:host in
                  let live = Db.is_valid_ref __context metrics && (Db.Host_metrics.get_live ~__context ~self:metrics) in
@@ -2081,7 +2081,7 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
 
     let import ~__context ~url ~sr ~full_restore ~force =
       info "VM.import: url = '%s' sr='%s' force='%b'" "(url filtered)" (Ref.string_of sr) force;
-      let pbd = choose_pbd_for_sr ~__context ~self:sr () in
+      let pbd = choose_pbd_for_sr ~consider_unplugged_pbds:false ~__context ~self:sr () in
       let host = Db.PBD.get_host ~__context ~self:pbd in
       do_op_on ~local_fn:(Local.VM.import ~url ~sr ~full_restore ~force) ~__context ~host (fun session_id rpc -> Client.VM.import rpc session_id url sr full_restore force)
 
@@ -3228,8 +3228,8 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
     (* -------- Forwarding helper functions: ------------------------------------ *)
 
     (* Forward SR operation to host that has a suitable plugged (or unplugged) PBD  *)
-    let forward_sr_op ?consider_unplugged_pbds ~local_fn ~__context ~self op =
-      let pbd = choose_pbd_for_sr ?consider_unplugged_pbds ~__context ~self () in
+    let forward_sr_op ?(consider_unplugged_pbds=false) ~local_fn ~__context ~self op =
+      let pbd = choose_pbd_for_sr ~consider_unplugged_pbds ~__context ~self () in
       let host = Db.PBD.get_host ~__context ~self:pbd in
       do_op_on ~local_fn ~__context ~host op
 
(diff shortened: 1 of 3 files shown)


ocaml/xapi/message_forwarding.ml
```diff
@@ -3899,8 +3952,13 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
       info "PBD.destroy: PBD '%s'" (pbd_uuid ~__context self);
       let sr = Db.PBD.get_SR ~__context ~self in
+      let host = Db.PBD.get_host ~__context ~self in
       SR.with_sr_marked ~__context ~sr ~doc:"PBD.destroy" ~op:`pbd_destroy
-        (fun () -> Local.PBD.destroy ~__context ~self)
+        (fun () -> (* Forget metrics on the host which owns that SR. This is safe
+                      because on a host there is at most one pbd per sr *)
+                   do_op_on ~local_fn:(Local.SR.forget ~sr) ~__context ~host
+                     (fun session_id rpc -> Client.SR.forget rpc session_id sr);
+                   Local.PBD.destroy ~__context ~self)
```
lippirk: as far as i'm concerned this PR is working as expected and is good to go

however, i would like to get your opinion on changing the `PBD.destroy` api @robhoes ?

edwintorok: What if this is a shared SR that I have detached from one host and did a PBD destroy on? Wouldn't forgetting the SR be problematic for the other hosts that are still using it?

lippirk: yes - i will test something like the following then

```
fun () -> if (shared_sr && i_am_last_host connected_to_sr) || not shared_sr then
                forget_sr ()
              pbd_destroy ()

```

lippirk: i have realised that `Client.SR.forget` and `Local.SR.forget` are actually a bit different, since `Client.SR.forget` performs in addition a `Xapi_sr.really_forget`, so I need to think quite carefully about what is going on

lippirk: i think the solution is to allow `SR.forget` when an SR has no PBDs, and for `PBD.destroy` to do what it did before

ocaml/xapi/message_forwarding.ml
```diff
@@ -30,6 +30,10 @@ open D
 module Audit = Debug.Make(struct let name="audit" end)
 let info = Audit.debug
 
+module PBDSet = Set.Make (struct
+  type t = API.ref_PBD
+  let compare = Pervasives.compare
```
lindig: This should be `Stdlib.compare`

ocaml/xapi/message_forwarding.ml
```diff
@@ -161,25 +165,35 @@ let log_exn_ignore ?(doc = "performing unknown operation") f x =
...
+      pbds_to_consider
+  in
+  match pbd_candidates with
+  | [] ->
+    let err = if consider_unplugged_pbds then Api_errors.sr_no_pbds else Api_errors.sr_no_plugged_pbds in
+    raise (Api_errors.Server_error(err, [ Ref.string_of self ]))
+  | pbds -> pbds
 
+let choose_pbd_for_sr ~consider_unplugged_pbds ~__context ~self () =
+  List.hd @@ choose_pbds_for_sr ~consider_unplugged_pbds ~__context ~self ()
```
lindig: Can this fail? I fhink a `match` with an explicit error (or at least a `failwith`) for the empty list would be good defense.

ocaml/xapi/message_forwarding.ml
```diff
@@ -161,25 +165,35 @@ let log_exn_ignore ?(doc = "performing unknown operation") f x =
...
 (* Given an SR, return a PBD to use for some storage operation. *)
-(* In the case of SR.destroy we need to be able to forward the SR operation when all
-   PBDs are unplugged - this is the reason for the consider_unplugged_pbds optional
-   argument below. All other SR ops only consider plugged PBDs... *)
-let choose_pbd_for_sr ?(consider_unplugged_pbds=false) ~__context ~self () =
+(* In the case of SR.destroy and forget we need to be able to forward the SR
+   operation when all PBDs are unplugged. This is the reason for the
+   consider_unplugged_pbds optional argument below. All other SR ops only
+   consider plugged PBDs. *)
+let choose_pbds_for_sr ~consider_unplugged_pbds ~__context ~self () =
```
lindig: Previously the argument was optional and with a default. Now it is no longer optional - unlike what the comment says.

ocaml/xapi/message_forwarding.ml
robhoes: @lippirk That sounds right. I recall talking about this a couple of weeks ago.

By the way, when you say `Client.SR.forget` and `Local.SR.forget` are different, do you mean that the former has things going on in `Message_forwarding` while the latter doesn't? That is not exactly true. In general, the `Client` module is for sending an call through the API layer to a xapi daemon on the same or another host. In `Message_forwarding`, this is normally used to forward a call from the master to a slave host. When a slave receives a forwarded API call, it does not go through message forwarding again, but the local function is called directly.

lippirk: yes that's what i meant

ah i see, i thought that we go through message forwarding again. so when an rpc is sent from one xapi to another it contains information about whether it has already been forwarded and the receiving xapi unpacks this accordingly into a `Local.X.Y` call?

robhoes: See https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/server_helpers.ml#L76.

lippirk:
thanks for reviewing. i am happy with the functionality, and am confident the failing storage quicktests were attributable to issues fixed in https://github.com/xapi-project/xen-api/pull/4012.

i should mention that a couple of storage quicktests still fail in a pool, but these failures can also be seen on master

i will wait for a final review from @psafont

ocaml/xapi/message_forwarding.ml
```diff
@@ -3295,24 +3295,45 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
           Xapi_sr.assert_sr_not_indestructible ~__context ~sr;
           Xapi_sr.assert_sr_not_local_cache ~__context ~sr;
 
-          forward_sr_op ~consider_unplugged_pbds:true ~local_fn ~__context ~self:sr
-            (fun session_id rpc -> Client.SR.destroy rpc session_id sr))
+          forward_sr_op ~consider_unplugged_pbds:true ~local_fn:local_destroy ~__context ~self:sr
+            (fun session_id rpc -> Client.SR.destroy rpc session_id sr);
+
+          begin
+            match pbd_state with
```
psafont: These 2 lines can be put into a single line to use less indentation. Alternatively `let () = [match ...] in` can be used.

ocaml/xapi/message_forwarding.ml
```diff
@@ -161,25 +165,37 @@ let log_exn_ignore ?(doc = "performing unknown operation") f x =
...
-    match pbds_to_consider with
-    | [] -> raise (Api_errors.Server_error(Api_errors.sr_no_pbds, [ Ref.string_of self ]))
-    | pdb :: _ -> pdb
-
+  let pbds = PBDSet.of_list pbds_to_consider in
+  let pbd_candidates =
+    if Helpers.is_sr_shared ~__context ~self then
+      let master = Helpers.get_master ~__context in
+      let master_pbds = PBDSet.of_list (Db.Host.get_PBDs ~__context ~self:master) in
+      (* Operation run on shared SRs depend on the first pbd of the list to be
```
robhoes: The old function is stricter than this, as it raises an API error if `pbds_to_consider` does not contain a master PBD. Now, `sr_master_pbds` can be empty, while `rest_pbds` is not, without an error being raised, and a caller of `choose_pbd_for_sr` gets a non-master PBD.

ocaml/xapi/message_forwarding.ml
```diff
@@ -161,25 +165,37 @@ let log_exn_ignore ?(doc = "performing unknown operation") f x =
...
+         on the master. *)
+      let sr_master_pbds, rest_pbds = PBDSet.partition (fun pbd ->
+          PBDSet.mem pbd master_pbds) pbds in
+      List.rev_append (PBDSet.elements sr_master_pbds) (PBDSet.elements rest_pbds)
+    else
+      pbds_to_consider
+  in
+  match pbd_candidates with
+  | [] ->
+    let err = if consider_unplugged_pbds then Api_errors.sr_no_pbds else Api_errors.sr_no_plugged_pbds in
```
robhoes: Do we need to change the error reporting here? This is essentially an API change, which has not been justified.

Furthermore, the `SR_NO_PBDs` error has doc string `"The SR has no attached PBDs"` (where "attached" is the same as "plugged"), so the old error actually meant the same as your new error code.

Finally, if we really need a new error, the besides defining it in `api_errors.ml`, you must also add it to `datamodel_errors.ml`.

ocaml/xapi/xapi_sr.ml
```diff
@@ -369,13 +369,15 @@ let create  ~__context ~host ~device_config ~(physical_size:int64) ~name_label ~
...
-  let any_pbds_attached_to_sr = Db.PBD.get_all_records ~__context
-  |> List.exists (fun (_ref, pbd) ->
-      pbd.API.pBD_SR = sr && pbd.API.pBD_currently_attached)
-  in
-
-  if any_pbds_attached_to_sr then
-    raise (Api_errors.Server_error(Api_errors.sr_has_pbd, [ Ref.string_of sr ]))
+  let pbds = Db.SR.get_PBDs ~__context ~self:sr in
+  match pbds with
+  | []   -> `No_pbds
```
robhoes: An assert function that returns something (other than `()`) is a bit odd...

ocaml/xapi/message_forwarding.ml
```diff
@@ -3295,24 +3295,45 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
 
-          forward_sr_op ~consider_unplugged_pbds:true ~local_fn ~__context ~self:sr
-            (fun session_id rpc -> Client.SR.destroy rpc session_id sr))
+          forward_sr_op ~consider_unplugged_pbds:true ~local_fn:local_destroy ~__context ~self:sr
+            (fun session_id rpc -> Client.SR.destroy rpc session_id sr);
+
+          begin
+            match pbd_state with
+            | `No_pbds            -> (* no SR metrics to remove *) ()
+            | `All_pbds_unplugged ->
```
robhoes: Do we really need to distinguish the no-PBD case here? Can we not just let the `forward_sr_all_op` be a no-op in that case?

In this light, perhaps it is better if `choose_pbds_for_sr` does not raise errors, but delegates that to `choose_pbd_for_sr`.

ocaml/xapi/message_forwarding.ml
lippirk: okay, i will remove this extra error since it is not adding much information

PR #4011, opened 2020-01-07
CP-32461: Delete CPUID leveling checks from WLB

TalonsLee:
Enable do_cpuid_check when recommendation from WLB is filtered. Set zero
score if the vm_incompatible_with_this_host Api_errors is detected.

### DIFF
--- a/ocaml/xapi/xapi_vm_helpers.ml
+++ b/ocaml/xapi/xapi_vm_helpers.ml
@@ -537,9 +537,25 @@ let retrieve_wlb_recommendations ~__context ~vm ~snapshot =
   List.iter
     (fun (h, r) ->
        try
-         assert_can_boot_here ~__context ~self:vm ~host:h ~snapshot ~do_cpuid_check:false ();
+         assert_can_boot_here ~__context ~self:vm ~host:h ~snapshot ~do_cpuid_check:true ();
          Hashtbl.replace recs h r;
        with
+       (* CPUID checks are being removed from WLB, this will become the responsibility of XAPI
+        * (and ultimately Xen/libxc). Previously WLB would've assigned a score of 0.0 and a reason
+        * of "HostCPUFeaturesIncompatible" to hosts that can't boot a VM due to CPU features missing.
+        * WLB could call assert_can_boot_here instead of doing its own CPUID checks, however that
+        * would add a lot of extra API calls (at least Number_of_hosts * Number_of_VMs).
+        * Instead do this check in XAPI itself, and override WLB's response for backwards compatibility
+        * with 0.0, and the same reason code that WLB would've used. This should keep observable behaviour
+        * identical and work both with old and new WLBs.
+        *)
+       | Api_errors.Server_error(x, _) when x=Api_errors.vm_incompatible_with_this_host ->
+         begin match r with
+         | id :: _ :: rec_id :: _ ->
+           Hashtbl.replace recs h [id; "0.0"; rec_id; "HostCPUFeaturesIncompatible"]
+         | _ ->
+           debug "Missing element of recommendation of host %s!" (Ref.string_of h);
+         end
        | Api_errors.Server_error(x, y) -> Hashtbl.replace recs h (x :: y))
     (retrieve_vm_recommendations ~__context ~vm);
   if ((Hashtbl.length recs) <> (List.length (Helpers.get_live_hosts ~__context)))


TalonsLee:
2024 is the ZeroSocreReason(HostCPUFeaturesIncompatible = 2024) used in WLB.

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -537,9 +537,14 @@ let retrieve_wlb_recommendations ~__context ~vm ~snapshot =
...
        try
-         assert_can_boot_here ~__context ~self:vm ~host:h ~snapshot ~do_cpuid_check:false ();
+         assert_can_boot_here ~__context ~self:vm ~host:h ~snapshot ~do_cpuid_check:true ();
          Hashtbl.replace recs h r;
        with
+       | Api_errors.Server_error(x, _) when x=Api_errors.vm_incompatible_with_this_host ->
+         try
+           let zeroScoreReasonVmIncompatible = "2024" in
+           Hashtbl.replace recs h [List.nth r 0; "0"; List.nth r 2; zeroScoreReasonVmIncompatible]
+         with Invalid_argument _ ->  debug "Missing element of recommendation of host %s!" (Ref.string_of h)
```
edwintorok: Instead of try/with you could also use a match:
```ocaml
| Api_errors.Server_error(x, y) when x=Api_errors.vm_incompatible_with_this_host ->
   begin match r with
   |  id :: _ :: rec_id :: _ -> 
      let zeroScoreReasonVmIncompatible = "2024" in
      Hashtbl.replace recs h [id; "0.0"; rec_id; zeroScoreReasonVmIncompatible]
   | _ ->
    debug "Missing element of recommendation of host %s!" (Ref.string_of h);
    Hashtbl.replace recs h (x :: y)
   end
```

Also replace `0` with `0.0`, that is what some of the rest of the code seems to expect, see in function `choose_host_for_vm` this line: `"WLB"; "0.0"; rec_id; zero_reason]`

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -537,9 +537,14 @@ let retrieve_wlb_recommendations ~__context ~vm ~snapshot =
   List.iter
     (fun (h, r) ->
        try
-         assert_can_boot_here ~__context ~self:vm ~host:h ~snapshot ~do_cpuid_check:false ();
+         assert_can_boot_here ~__context ~self:vm ~host:h ~snapshot ~do_cpuid_check:true ();
          Hashtbl.replace recs h r;
        with
+       | Api_errors.Server_error(x, _) when x=Api_errors.vm_incompatible_with_this_host ->
+         try
```
edwintorok: Perhaps add a comment here explaining why this needs to be done, something like this:
> CPUID checks are being removed from WLB, this will become the responsibility of XAPI (and ultimately Xen/libxc).
Previously WLB would've assigned a score of 0.0 and a reason of 2024 (vm incompatible) to hosts that cannot boot a VM due to CPU features missing. WLB could call `assert_can_boot_here` instead of doing its own CPUID checks, however that would add a lot of extra API calls (at least Number_of_hosts * Number_of_VMs).
Instead do this check in XAPI itself, and override WLB's response for backwards compatibility with `0.0`, and the same reason code that WLB would've used. This should keep observable behaviour identical and work both with old and new WLBs.

ocaml/xapi/xapi_vm_helpers.ml
TalonsLee: "r" can be a 3 or 4 elements list, see function retrieve_vm_recommendations.
So I use list element access.
OK, I will try match.

I set it to "0.0" first, but WLB set zero score to "0".
OK. I will set it to 0.0 to match the logic in xapi.

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -537,9 +537,25 @@ let retrieve_wlb_recommendations ~__context ~vm ~snapshot =
...
+        * WLB could call assert_can_boot_here instead of doing its own CPUID checks, however that
+        * would add a lot of extra API calls (at least Number_of_hosts * Number_of_VMs).
+        * Instead do this check in XAPI itself, and override WLB's response for backwards compatibility
+        * with 0.0, and the same reason code that WLB would've used. This should keep observable behaviour
+        * identical and work both with old and new WLBs.
+        *)
+       | Api_errors.Server_error(x, _) when x=Api_errors.vm_incompatible_with_this_host ->
+         begin match r with
+         | id :: _ :: rec_id :: _ ->
+           Hashtbl.replace recs h [id; "0.0"; rec_id; "HostCPUFeaturesIncompatible"]
```
edwintorok: The string definetely looks nicer, thought this has to be 2024 though, does this still pass the XenRT test?

TalonsLee: yes, jobid: 2770097, 2770098 pass

PR #4012, opened 2020-01-08
CA-325582 fix some storage quicktests for pools

lippirk:
This does not fix all failures, but some.

Some storage quicktests were failing on pools
because by default it runs each test for each SR
a host can see - including the local storage of
another host in the pool. To fix we add a filter
to remove such SRs.

`vdi_snapshot_in_pool` was also iterating over
hosts - it should not have been, because we
don't necessarily have access to that host's SRs.
Instead, only test snapshot on localhost.

### DIFF
--- a/ocaml/quicktest/qt.ml
+++ b/ocaml/quicktest/qt.ml
@@ -38,6 +38,13 @@ let cli_cmd args =
   | e ->
     Alcotest.fail ("CLI failed" ^ (Printexc.to_string e))
 
+(** Look up key in /etc/xensource-inventory *)
+let inventory_lookup k =
+  Xapi_inventory.inventory_filename := "/etc/xensource-inventory";
+  Xapi_inventory.lookup k
+
+let localhost_uuid = inventory_lookup Xapi_inventory._installation_uuid
+
 module Test = struct
   let assert_raises_match exception_match fn =
     try
@@ -119,8 +126,7 @@ module VM = struct
     Client.Client.Host.get_control_domain rpc session_id host
 
   let get_dom0 rpc session_id =
-    Xapi_inventory.inventory_filename := "/etc/xensource-inventory";
-    let uuid = Xapi_inventory.lookup Xapi_inventory._control_domain_uuid in
+    let uuid = inventory_lookup Xapi_inventory._control_domain_uuid in
     Client.Client.VM.get_by_uuid ~rpc ~session_id ~uuid
 end
 
--- a/ocaml/quicktest/qt_filter.ml
+++ b/ocaml/quicktest/qt_filter.ml
@@ -116,19 +116,32 @@ module SR = struct
     ; required_sm_api_version
     }
 
-  let list_selected_srs rpc session_id =
-    Client.Client.SR.get_all rpc session_id
-    |> List.filter
-      (fun sr ->
-         Client.Client.SR.get_PBDs rpc session_id sr
-         |> List.exists
-           (fun pbd -> Client.Client.PBD.get_currently_attached rpc session_id pbd)
-      )
-    |> List.map (get_sr_info rpc session_id)
+  let list_srs_connected_to_localhost rpc session_id =
+    let is_attached = Client.Client.PBD.get_currently_attached ~rpc ~session_id in
+    let all_attached = List.for_all (fun pbd -> is_attached pbd) in
+    let does_pbd_belong_to_me pbd =
+      let pbd_host = Client.Client.PBD.get_host ~rpc ~session_id ~self:pbd in
+      let pbd_host_uuid = Client.Client.Host.get_uuid ~rpc ~session_id ~self:pbd_host in
+      pbd_host_uuid = Qt.localhost_uuid
+    in
+    let is_shared_or_local_and_my_pbds_attached sr =
+      let pbds = Client.Client.SR.get_PBDs rpc session_id sr in
+      let my_pbds = List.find_all does_pbd_belong_to_me pbds in
+      match pbds, my_pbds with
+      | [], _         -> Printf.eprintf "SR (%s) has no pbds %s\n" (Ref.string_of sr) __LOC__;
+                         false
+      | [_], [my_pbd] -> (* Local SR *) is_attached my_pbd
+      | _::_, []      -> Printf.eprintf "This host doesn't own any PBDs from shared SR (%s) %s\n" (Ref.string_of sr) __LOC__;
+                         false
+      | _::_, _::_    -> (* Shared SR *) my_pbds |> all_attached
+    in
+    Client.Client.SR.get_all rpc session_id |>
+    List.filter is_shared_or_local_and_my_pbds_attached |>
+    List.map (get_sr_info rpc session_id)
 
   let only sr () = [get_sr_info !A.rpc !session_id sr]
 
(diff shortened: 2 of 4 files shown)


ocaml/quicktest/qt_filter.ml
```diff
@@ -128,7 +128,24 @@ module SR = struct
...
+  let all_srs =
+    let is_shared_or_local sr_info =
+      (** allows us to ignore srs which don't belong to us.
+        * this remedies some failing tests if we run the quicktest binary on a host in a pool *)
+      let sr = sr_info.Qt.sr in
+      let session_id = !session_id in
+      let rpc = !A.rpc in
+      let pbds = Client.Client.SR.get_PBDs ~rpc ~session_id ~self:sr in
+      match pbds with
+      | [pbd] -> begin
```
lindig: The `begin`/`end` is not required for the right-hand side of a `match` case.

ocaml/quicktest/quicktest_args.ml
```diff
@@ -10,6 +10,19 @@ let username = ref ""
 let password = ref ""
 
 let host = ref ""
+let localhost_uuid =
```
lindig:
```
let localhost_uuid = Inventory.lookup "INSTALLATION_UUID"
```

ocaml/quicktest/qt_filter.ml
```diff
@@ -128,7 +128,24 @@ module SR = struct
...
+          let pbd_host = Client.Client.PBD.get_host ~rpc ~session_id ~self:pbd in
+          let pbd_host_uuid = Client.Client.Host.get_uuid ~rpc ~session_id ~self:pbd_host in
+          pbd_host_uuid = A.localhost_uuid (* sr is not shared, test it only if it is ours *)
+        end
+      | []    -> Printf.eprintf "SR (%s) has no pbds %s" (Ref.string_of sr) __LOC__; false
+      | _::_  -> true (* sr is shared, and we can see it *)
```
lindig: Why does a single pbd signal a PBD that needs to be checked but multiple (pattren `_::_`) do not?

ocaml/quicktest/qt_filter.ml
gaborigloi: I think it's because of how it interacts with `list_selected_srs`, but I think @lippirk it would be cleaner to merge this functionality into `list_selected_srs`, because it seems that function is only used from here, and maybe rename it to something more descriptive, like `list_available_srs`.

ocaml/quicktest/qt_filter.ml
psafont: AFAIK shared SRs are always visible to the master.

ocaml/quicktest/quicktest_args.ml
lippirk: i tried that, but it doesn't work

at the time I put it down to the fact that the `quicktest` binary is actually a client to xapi, rather than part of the xapi daemon

lippirk: if there is a cleaner way to get the localhost uuid however, i would much rather do that

ocaml/quicktest/qt_filter.ml
lippirk: yes you are right, i should check that at least one of the pbds on the shared sr belongs to the host where we are running the binary

lippirk: on reflection i am not sure if this is right.

my assumption was that we have access to a shared SR iff one of its PBDs belongs to us. this is probably not true

lippirk: after asking on slack, a shared sr should be accessible to each host in the pool, so the original code is correct. however, i will add a sanity check and print a warning if this is not the case

ocaml/quicktest/qt_filter.ml
```diff
@@ -116,36 +116,41 @@ module SR = struct
...
-      | [pbd] -> begin
-          let pbd_host = Client.Client.PBD.get_host ~rpc ~session_id ~self:pbd in
-          let pbd_host_uuid = Client.Client.Host.get_uuid ~rpc ~session_id ~self:pbd_host in
-          pbd_host_uuid = A.localhost_uuid (* sr is not shared, test it only if it is ours *)
-        end
+      | [pbd] -> (* sr is not shared, test it only if it is ours *)
+        does_pbd_belong_to_me pbd
       | []    -> Printf.eprintf "SR (%s) has no pbds %s" (Ref.string_of sr) __LOC__; false
-      | _::_  -> true (* sr is shared, and we can see it *)
+      | _::_ as pbds  -> (** sr is shared, so we have access to is
```
lindig:
```
| [] ->
| [pbd] ->
| pbds -> when List.exists does_pbd_belong_to_me pbds ->
| _  ->
```
I would have expected `List.for_all` rather than `List.exists` but maybe knowing one is good enough.

lippirk: `for_all` is not right - each host in the pool should have a PBD for any shared SR

ocaml/quicktest/quicktest_args.ml
lindig: The Inventory module is part of xcp-inventory. Would it be possible to use that? If not, I think the solution is fine.

https://github.com/xapi-project/xcp-inventory

The implementation is a single file. So it could be also vendored.

psafont: I think adding it to the libraries in dune's executable definition would work, see `ocaml/util/dune`

lippirk: `Xapi_inventory` is already a dependency of `quicktestbin` (presumably via `xapi_internal`)

the reason it wasn't working was because you have to set `inventory_filename`, but that wasn't happening early enough.

Interestingly setting it at the top of main (`quicktest.ml`) doesn't seem to work, so i have set it when we assign `localhost_uuid`

ocaml/quicktest/qt.ml
```diff
@@ -119,7 +119,6 @@ module VM = struct
     Client.Client.Host.get_control_domain rpc session_id host
 
   let get_dom0 rpc session_id =
-    Xapi_inventory.inventory_filename := "/etc/xensource-inventory";
```
lippirk: helper in qt

ocaml/quicktest/qt_filter.ml
```diff
@@ -116,19 +116,41 @@ module SR = struct
...
+                           * do a sanity check that we own one of the pbds *)
+        if List.exists does_pbd_belong_to_me pbds then
+          true
+        else begin
+          Printf.eprintf "SR (%s) is shared, but we do not own any of its pbds... %s" (Ref.string_of sr) __LOC__;
+          false
+        end
+    in
+    let has_at_least_one_attached_pbd sr =
```
lippirk: this is wrong - need to make sure that sr's pbd that is ours is attached, not simply that any pbd is attached

ocaml/quicktest/quicktest_args.ml
```diff
@@ -10,6 +10,9 @@ let username = ref ""
...
+  Xapi_inventory.inventory_filename := "/etc/xensource-inventory";
+  Xapi_inventory.lookup Xapi_inventory._installation_uuid
```
lippirk: move to qt

PR #4014, opened 2020-01-14
CA-325582 backport to 7.1 CU2

lippirk:
To help avoid memory usage growing in xcp-rrdd, we delete SR metrics whenever an SR is forgotten or destroyed

### DIFF
--- a/ocaml/xapi/xapi_sr.ml
+++ b/ocaml/xapi/xapi_sr.ml
@@ -271,12 +271,8 @@ let create  ~__context ~host ~device_config ~(physical_size:int64) ~name_label ~
   sr_ref
 
 let assert_all_pbds_unplugged ~__context ~sr =
-  let any_pbds_attached_to_sr = Db.PBD.get_all_records ~__context
-  |> List.exists (fun (_ref, pbd) ->
-      pbd.API.pBD_SR = sr && pbd.API.pBD_currently_attached)
-  in
-
-  if any_pbds_attached_to_sr then
+  let pbds = Db.SR.get_PBDs ~__context ~self:sr in
+  if List.exists (fun self -> Db.PBD.get_currently_attached ~__context ~self) pbds then
     raise (Api_errors.Server_error(Api_errors.sr_has_pbd, [ Ref.string_of sr ]))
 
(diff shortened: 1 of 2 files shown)


lippirk:
should go it at the same time as https://github.com/xapi-project/xcp-rrdd/pull/109

the fixups contain work that was done to make the cherry picks compile (aside from merge conflicts)

lindig:
This fixups are adding functionality that are part of the standard library today. These are looking good and I would not squash them.

psafont:
I'm not sure about the value of adding code to add missing functions in Ocaml 4.02, it's something additional to keep track of as it will get sprinkled around the code in each backport, leading to code duplication. The alternative is like we're doing now, translate the idioms to the old ways. We can discuss about this.

edwintorok:
There are various projects/"shims" that allow you to use new OCaml standard library features when having an old compiler. We could import one of those into xs-opam for Havana and save us the extra effort for backporting.

edwintorok:
http://opam.ocaml.org/packages/stdcompat/ tries to backport everything and is fairly large (I'd configure it with --disable-magic, I don't trust Obj.magic much...)

There are more targeted backports:
http://opam.ocaml.org/packages/stdlib-shims/ is for Pervasives -> Stdlib.
http://opam.ocaml.org/packages/seq/seq.0.2.2/ is for Seq (but you'd be missing the List.of_seq, etc. so not very useful on its own)

lippirk:
as discussed yesterday, i have moved the extra utils to stdext (apart from `is_suffix`, since this is pretty much an alias for `endswith`)

ocaml/xapi/xapi_sr.ml
```diff
@@ -354,13 +361,13 @@ let maybe_copy_sr_rrds ~__context ~sr =
    each PBD got unplugged. *)
 let unload_metrics_from_memory ~__context ~sr =
   let short_uuid = String.sub (Db.SR.get_uuid ~__context ~self:sr) 0 8 in
-  let is_sr_metric = Astring.String.is_suffix ~affix:short_uuid in
```
psafont: I'd prefer using directly `endswith` here for the single use, there's less scaffolding code and the translation is very easy to understand.

PR #4015, opened 2020-01-14
Server certificate validation

psafont:
Add private key and certificate sanity checking checking, this is not meant to replace online certificate chain checking.

I also took the opportunity to make the existing interface clearer.


ocaml/xapi/certificates.ml
```diff
@@ -334,3 +335,57 @@ let validate_private_key pkcs8_private_key =
   |> function
     | Ok priv -> priv
     | Error `Msg (err, msg) -> raise_server_error msg err
+
+let validate_certificate kind str now private_key =
+  let open Rresult in
+  let ensure_keys_match private_key certificate =
+    let public_key = X509.Certificate.public_key certificate in
+    match public_key, private_key with
+    | `RSA pub, `RSA priv when pub = Nocrypto.Rsa.pub_of_priv priv ->
```
edwintorok: Is polymorphic equality the only way to compare these, or would the library provide a comparison function?

ocaml/xapi/certificates.ml
```diff
@@ -334,3 +335,57 @@ let validate_private_key pkcs8_private_key =
...
+        Ok certificate
+    | _ ->
+        Error (`Msg (server_certificate_key_mismatch, []))
+  in
+
+  let ensure_validity ~time certificate =
+    let to_string = Ptime.to_rfc3339 in
```
edwintorok: Should we also include some grace period here?
i.e. if someone tries to install a certificate that expires the next day we should probably warn them in that case too?
i.e. have a minimum validity period of lets say 30 days?

ocaml/xapi/certificates.ml
```diff
@@ -305,3 +305,35 @@ let hostnames_of_pem_cert pem =
...
+    else
+      Ok priv
+  in
+
+  let raw_pem = Cstruct.of_string pkcs8_private_key in
+  let open Rresult in
+  X509.Private_key.decode_pem raw_pem
+  |> R.reword_error
+      (fun (`Msg err_msg) ->
+        if Astring.String.is_infix ~affix:"multi-prime RSA" err_msg then
```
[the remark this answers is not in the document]
psafont: they are all defined as parse_error "message", unfortunately

ocaml/xapi/certificates.ml
psafont: There are alerts planned to warn users about expiring server certificates, I'm not sure if it's worth doing the check here as well. I'll take note and see if it's easy to reuse those checks here as well.

ocaml/xapi-consts/api_errors.ml
```diff
@@ -550,6 +550,16 @@ let crl_does_not_exist = "CRL_DOES_NOT_EXIST"
 let crl_already_exists = "CRL_ALREADY_EXISTS"
 let crl_name_invalid = "CRL_NAME_INVALID"
 let crl_corrupt = "CRL_CORRUPT"
+let server_certificate_key_invalid = "SERVER_CERTIFICATE_KEY_INVALID"
```
robhoes: Don't forget to define these errors with their doc strings and parameters in `datamodel_errors.ml`.

ocaml/tests/test_certificates.ml
```diff
@@ -0,0 +1,206 @@
...
+     error_message list ) *)
+let corrupt_certificates =
+  [ "cert_bogus",
+    "pkey_rsa_2048",
+    "2020-02-01T00:00:00+00:00",
+    server_certificate_invalid,
+    []
+  ]
+
+let key_chain = List.map (fun _ -> `RSA (Nocrypto.Rsa.generate 2048)) (List.init 3 Fun.id)
```
lippirk: is this not equivalent to `List.init 3 (`RSA (Nocrypto.Rsa.generate 2048))`?

psafont: I agree this can be made shorter.

ocaml/xapi/certificates.ml
psafont: The certificate module provides
```
val supports_keytype : t -> key_type -> bool

    supports_keytype certificate key_type is result, whether public key of the certificate matches the given key_type.
```
But I don't see a function to extract keytype from a private or public key. I don't think I can avoid polymorphic equality.

psafont:
I want the DCO check to fail with the travis commit on the feature branch as it needs to be removed when merging to master.

ocaml/xapi/certificates.ml
```diff
@@ -24,33 +20,37 @@ open Client
 module D=Debug.Make(struct let name="certificates" end)
 open D
 
+type t_trusted = Certificate | CRL
```
lippirk: i don't know what a CRL is, but wikipedia says:

> '[a CRL] is "a list of digital certificates that have been revoked by the issuing certificate authority (CA) before their >scheduled expiration date and should no longer be trusted"'

so i'm confused as to why it is part of `t_trusted`

lippirk: is it the list that we trust, rather than the items on the list?

ocaml/xapi/certificates.ml
```diff
@@ -305,3 +305,32 @@ let hostnames_of_pem_cert pem =
...
+  |> R.reword_error
+      (fun (`Msg err_msg) ->
+        let unknown_algorithm = "Unknown algorithm " in
+        if Astring.String.is_prefix ~affix:"multi-prime RSA" err_msg then
+          `Msg (server_certificate_key_rsa_multi_not_supported, [])
+        else if Astring.String.is_prefix ~affix:unknown_algorithm err_msg then
+          `Msg (server_certificate_key_algorithm_not_supported,
+            [Astring.String.with_range ~first:(String.length unknown_algorithm) err_msg])
+        else
+          `Msg (server_certificate_key_invalid, []))
```
lippirk: it might be helpful to log the original error here, in case we run into this (assuming you can confidently filter out the secrets)

ocaml/xapi/certificates.ml
```diff
@@ -305,3 +305,32 @@ let hostnames_of_pem_cert pem =
...
+  let open Rresult in
+  X509.Private_key.decode_pem raw_pem
+  |> R.reword_error
+      (fun (`Msg err_msg) ->
+        let unknown_algorithm = "Unknown algorithm " in
+        if Astring.String.is_prefix ~affix:"multi-prime RSA" err_msg then
+          `Msg (server_certificate_key_rsa_multi_not_supported, [])
+        else if Astring.String.is_prefix ~affix:unknown_algorithm err_msg then
+          `Msg (server_certificate_key_algorithm_not_supported,
+            [Astring.String.with_range ~first:(String.length unknown_algorithm) err_msg])
```
lippirk: could `err_msg` contain a secret?

ocaml/xapi/certificates.ml
psafont: The type represents certificates that are trusted by the pool to validate the other party on TLS connections, this includes the list of invalid certs, because the software "trusts" the list to be correct and true.

This is in juxtaposition of the server certificate which is not used to validate other certificates on connection, but is shown to the other party.

ocaml/xapi/certificates.ml
psafont: I have not seen the library code exposing secrets willfully on the debug messages.

ocaml/xapi/certificates.ml
psafont: The messages have been exposed in a hard-coded way due to internationalization concerns. If there are more failure modes more API errors will have to be introduced.

ocaml/xapi/certificates.ml
```diff
@@ -74,100 +75,110 @@ let not_safe_chars name =
...
-
-let host_install is_cert ~name ~cert =
-  if not_safe is_cert name then
-    raise_name_invalid is_cert name;
-  let filename = library_filename is_cert name in
+let cert_perms kind =
+  let stat = Unix.stat (library_path kind) in
+  let mask = 0o666 in
+  let perm = stat.Unix.st_perm land mask in
+  debug "%d %d" perm stat.Unix.st_perm;
```
lindig: Has this log line enough context to be useful? These are just two numbers.

psafont: I don't think so, I'm partial to removing the logline.

I changed this function because it's most probably going to be reused for the new certificates, which need different permissions.

psafont:
Merging this now that the code compatible with X509 0.9.0.

I'll keep the comments in mind for doing further cleanups in certificates.ml

PR #3800, opened 2019-01-18
Allow migrate_send during RPU

stormi:
It is needed to avoid downtime during RPU when VMs are running on local
storage and is already allowed for cross-pool storage migration between
mismatching XAPI versions, provided the destination has a higher version.

Related to #3764

I must admit that there may be a good reason why it was disabled in the first place, but I think that the reasons might not be valid anymore. If you look at the commit history, it was removed with commit 6760dad which says "Migrate_send wasn't a Miami operation". My guess is that the list of allowed actions was downgraded to those that "Miami" was able to handle. If I'm right, it has become less and less likely that someone would attempt an upgrade from that version (is it supported at all?).

Local migration tests went fine.

PR #4016, opened 2020-01-20
CA-333610 avoid long running idle connection during migration

lippirk:


ocaml/xapi/message_forwarding.ml
```diff
@@ -1727,7 +1743,23 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
+                  API.ref_VM_of_rpc x
+                 with e ->
+                   failwith @@ Printf.sprintf "impossible: VM ref was not placed in task's (ref = %s) result field. error: %s" (Ref.string_of t) (Printexc.to_string e)
+                 end
+               | `Cancelling | `Cancelled -> raise (Api_errors.Server_error (Api_errors.task_cancelled, [ Ref.string_of t]))
+               | `Server_error e -> raise e
+               | `Could_not_parse_xml_rpc e -> raise e
+               | `Unknown_error -> failwith "impossible: couldn't extract error info from task"
+               | `Pending ->  failwith "impossible: just waited on task"
+               | `Null -> failwith "impossible: vm ref wasn't written to task result")
```
lippirk: using `Client.Async.VM` instead of `Client.VM` introduces a lot of new error conditions (i hope most of these are actually impossible)

I'm not sure whether to introduce a new API error or whether there are suitable existing ones (there seem to be no generic `MIGRATION_FAIL` errors)

stormi:
Hi. This is a comment I've been refraining to do for months, but it is becoming frustrating :)

To us outside Citrix doors, `CA-333610` means nothing, so in the absence of a detailed commit message, it's very hard to feel included in the development.

Is that something that may evolve in the future?

Best regards, Samuel

lippirk:
> Someone made me realize that by going to the individual commits in the PR I would get a detailed commit message. My bad!

Glad you managed to resolve this yourself :+1: 

We use `CA-333610` to track the issue internally - I would be surprised if that were to change. The majority of our commits should have good messages

lippirk:
@robhoes 

further to our conversation yesterday about whether this would break pool upgrades - it looks like the PR to allow `migrate_send` during RPUs has not been merged https://github.com/xapi-project/xen-api/pull/3800

ocaml/xapi/message_forwarding.ml
```diff
@@ -1727,7 +1743,23 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
+               D.info "VM.migrate_send waiting for migration to complete... on task %s" (Ref.string_of t);
+               wait_for_tasks ~__debug:true ~__context ~tasks:[t];
+               match rpc_of_task ~__context ~self:t with
+               | `Ok x -> begin
+                 try
+                  API.ref_VM_of_rpc x
+                 with e ->
+                   failwith @@ Printf.sprintf "impossible: VM ref was not placed in task's (ref = %s) result field. error: %s" (Ref.string_of t) (Printexc.to_string e)
+                 end
+               | `Cancelling | `Cancelled -> raise (Api_errors.Server_error (Api_errors.task_cancelled, [ Ref.string_of t]))
```
lindig:
```
Api_errors.(Server_error (task_cancelled, [..]))
```

ocaml/xapi/message_forwarding.ml
lindig: Should these not be mapped to internal error if we expect them to never occur? You could use:

```
let fail msg  = Api_errors.(Server_error (internal_error, [msg])) 
```

robhoes:
I think that the approach of the async call is a good solution here. For the way that it is implemented, I would like it to be made more general, so that the pattern becomes easy to reuse for other API calls.

Firstly, I would like to avoid changes down in the implementation functions of the API calls, such as `Xapi_vm_migrate.migrate_send`. Those functions should not be modifying the task completeness and related state, which should be done high up in the server (e.g. `Server_helpers`).

Secondly, we need to think about compatibility more generally, where an upgraded master may forward a call to a slave that is still running an older version. This is another reason to not want changes down in functions like `Xapi_vm_migrate.migrate_send`. As you wrote above, this happens to be not a problem for this particular case (yet...), but it would be good to not rely on that.

In this upgrade case, I think that it is acceptable to simply not have the new behaviour, and have the old synchronous forwarding instead.

What we could do, is do the calls to slave in the same way as today, without `Async.`, so that everything works on older versions as before. But we add some additional information to the call, e.g. in the form of HTTP cookies or additional headers, which a slave running the old version just ignores. When that information is present, a slave running the new version picks this up in `Server_helpers.do_dispatch` (or around there), and executes the function asynchronously, and completes the task at the end.

Thoughts?

lippirk:
@robhoes this sounds good! i like the idea of putting this information in the http header and think we could get this working pretty easily (however, it surprises me that http requests do not contain api versions already?)

edwintorok:
additional headers might be better, we filter the cookies in logs, so might be more difficult to debug what happens. Having a generic way of specifying multiple API versions for a single API function might be useful.
Elsewhere (in SMAPIv2) we've introduced new API calls with a `v2`, `v3`, etc in their name as an approach to versioning. Since migration is mostly an internal API call we could do it that way as well, although its a bit more ugly.

lippirk:
i've modified the code to use the `User-Agent` header. unfortunately xapi uses the default value set by `xen-api-libs`... as a hack, i've shadowed the xmlrpc_client module

Also, the `Xapi_version` module doesn't seem to be reliable, so we should use something else.

edwintorok:
Xapi_version is hardcoded, nothing updates it. We should update it based on information from `git` and `.gitattributes` files. Was hoping we can defer this until we have Koji in place, because then we only need to care about one mode (git would always be present) instead of 2 (git for local development, and static tarball exported from git for rpms).

lippirk:
Perhaps we could use a git hook on github to do perform this update after each merge. For now it might be sensible to at least update it to this value:

```
[root@xrtmia-07-05 log]# xe host-list params=all | grep version
                     API-version-major ( RO): 2
                     API-version-minor ( RO): 14
 ```

**Comment by robhoes** [comment]
I don't think a version number is going to help us here? I meant adding an explicit option to the headers that means "execute asynchronously and complete".

**Comment by lippirk** [comment]
@robhoes @edwintorok i've fixed this so that it doesn't change the datamodel anymore

it's nearly working, the only issue that i've noticed is that some `Async.SR.update` tasks stick around after migrations involving versions of the api with/without the change (in particular inter pool, when everybody has been updated apart from the destination). i'm going to run some tests to see if anything pops up to make it more obvious what's going on

File: ocaml/idl/ocaml_backend/ocaml_syntax.ml [hunk 3]
```diff
@@ -77,33 +77,42 @@ end
 
 module Module = struct
   type e = Let of Let.t | Module of t | Type of Type.t
-  and t = { name: string;     (** OCaml module name *)
-            preamble: string list; (** Convenient place for helper functions, opens etc *)
-            letrec: bool;     (** True for all the let bindings to be mutually recursive*)
-            args: string list; (** for functor *)
+  and t = { name: string;           (** OCaml module name *)
+            preamble: string list;  (** Convenient place for helper functions, opens etc *)
+            postamble: string list; (** True for all the let bindings to be mutually recursive*)
```
robhoes: Looks like the comments haven't moved down?

ocaml/idl/ocaml_backend/ocaml_syntax.ml
```diff
@@ -77,33 +77,42 @@ end
...
+            letrec: bool;           (** for functor *)
+            args: string list;
             elements: e list }
 
-  let make ?(preamble=[]) ?(letrec=false) ?(args=[]) ~name ~elements () =
-    { name = name; preamble = preamble; letrec = letrec;
-      args = args; elements = elements }
+  let make ?(preamble=[]) ?(letrec=false) ?(args=[]) ?(postamble=[]) ~name ~elements () =
+    { name = name;
```
robhoes: Since your touch this: it can be shortened to `{name; preamble; postamble; letrec; args; elements}`.

ocaml/xapi/server_helpers.ml
```diff
@@ -118,26 +128,44 @@ let dispatch_exn_wrapper f =
   with exn -> let code, params = ExnHelper.error_of_exn exn in API.response_of_failure code params
 
 let do_dispatch ?session_id ?forward_op ?self called_async supports_async called_fn_name op_fn
-    marshaller_fn fd http_req label generate_task_for =
+    marshaller_fn fd http_req label sync_ty generate_task_for =
```
robhoes: The `called_async` arg is now obsolete, because it's just the same as ``sync_ty <> `Sync``. Let's remove it, and then also the `__async` value can be removed from server.ml and the helper above.

ocaml/xapi/server_helpers.ml
```diff
@@ -118,26 +128,44 @@ let dispatch_exn_wrapper f =
...
-      Rpc.success (marshaller_fn result)
+    in
+    let internal_async () =
+      (** similar to `default_async`, except we are zealous when completing the task
+        * this should be used when converting sync xapi <-> xapi calls to async *)
+      ignore (Thread.create
+                (fun () ->
+                  try
+                    let res = exec_with_context ~__context ~called_async ?f_forward:forward_op ~marshaller:marshaller_fn op_fn |> marshaller_fn in
+                    TaskHelper.complete ~__context (Some res)
```
robhoes: I think that we should leave `exec_with_context` be the only place were tasks are completed. This function already has a notion of `need_complete` inside of it, which we could pull out.

ocaml/xapi/message_forwarding.ml
```diff
@@ -1727,7 +1750,22 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
                assert_can_migrate ~__context ~vm ~dest ~live ~vdi_map ~vif_map ~vgpu_map ~options
              );
            forwarder ~local_fn ~__context ~vm
-             (fun session_id rpc -> Client.VM.migrate_send rpc session_id vm dest live vdi_map vif_map options vgpu_map)
+             (fun session_id rpc ->
+               let res =
```
robhoes: This retry logic could be pulled out into a function for general use.

lippirk:
the BVT is passing, running Ring3 now

i managed to break `VM.revert` (it would hang forever) by changing the code which waits for tasks from...

```
let classes = List.map (fun x -> Printf.sprintf "task/%s" (Ref.string_of x)) (our_task::tasks) in
```
to
```
let tasks = our_task::(List.filter (fun t -> t != our_task) tasks) in
let classes = List.map (fun x -> Printf.sprintf "task/%s" (Ref.string_of x)) tasks in
```
i'm still not sure why this didn't work, but i've undone it now. in retrospect this was not worth changing at all

ocaml/idl/ocaml_backend/ocaml_syntax.ml
```diff
@@ -77,33 +77,42 @@ end
 
 module Module = struct
   type e = Let of Let.t | Module of t | Type of Type.t
-  and t = { name: string;     (** OCaml module name *)
-            preamble: string list; (** Convenient place for helper functions, opens etc *)
-            letrec: bool;     (** True for all the let bindings to be mutually recursive*)
-            args: string list; (** for functor *)
+  and t = { name: string;           (** OCaml module name *)
+            preamble: string list;  (** Convenient place for helper functions, opens etc *)
+            postamble: string list;
```
psafont: Could add a docstring for the psotamble so it's in line with the rest?

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,68 @@ let host_supports_hvm ~__context host =
...
+        let from = Event_types.event_from_of_rpc from in
+        process from.Event_types.token
+      end else
+        ()
+    in
+    process ""
+
+  let to_result ~__context ~of_rpc ~name ~t =
+    wait_for ~__context ~tasks:[t];
+    let fail msg = raise (Api_errors.Server_error (Api_errors.internal_error, [Ref.string_of t; Printf.sprintf "%s: %s"name msg])) in
```
psafont: you could use `Api_errors.(Server_error (internal_error, ...))` to avoid repetition, likewise further down

ocaml/xapi/server_helpers.ml
```diff
@@ -117,44 +124,52 @@ let dispatch_exn_wrapper f =
     f()
   with exn -> let code, params = ExnHelper.error_of_exn exn in API.response_of_failure code params
 
-let do_dispatch ?session_id ?forward_op ?self called_async supports_async called_fn_name op_fn
-    marshaller_fn fd http_req label generate_task_for =
+let do_dispatch ?session_id ?forward_op ?self supports_async called_fn_name op_fn
+    marshaller fd http_req label sync_ty generate_task_for =
```
psafont: is there a reason why the parameter position of "async-ness" was moved?

lippirk: yes - `sync_ty` takes values of `Sync, Async, InternalAsync`, so we would be duplicating data by keeping the other async variable around

psafont: I understand that, the question is why is its position further back.

lippirk:
bst+bvt finally passing :tada: i believe this is ready to go now

ocaml/idl/ocaml_backend/gen_client.ml
```diff
@@ -188,21 +188,28 @@ let gen_module api : O.Module.t =
     "    | rpc -> failwith (\"Client.rpc: \" ^ Rpc.to_string rpc)";
   ]
   in
+  let postamble = [
+    "module Async = AsyncF(struct let async_qualifier = \"\" end)";
```
lindig: Using `{| .. |}` could avoid quoting.

ocaml/idl/ocaml_backend/ocaml_syntax.ml
```diff
@@ -77,33 +77,42 @@ end
...
+        | Let y -> Let.items_of ~prefix:(if x.letrec then "and" else "let") y
+        | Module x -> items_of x
+        | Type x -> [ Type.item_of x ] in
+      let opening = "module " ^ x.name ^ " = " ^
+                    (if x.args = []
+                     then ""
+                     else String.concat " " (List.map (fun x -> "functor(" ^ x ^ ") ->") x.args)) ^
+                    "struct" in
+      [ Line opening;
+        Indent  (
```
lindig: I'd use `List.concat`rather than repeatedly `@`:
```
let indent = List.concat
  [ ...
  ; ...
  ]
in Indent indent
```
Indent

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,66 @@ let host_supports_hvm ~__context host =
...
+
+let try_internal_async ~__context
+                       ~(debug_tag : string) (* e.g. debug_tag="VM.migrate_send" *)
+                       (internal_async_fn : unit -> API.ref_task)
+                       (sync_fn : unit -> 'a) (* only used when we are talking to a version of the api without InternalAsync *)
+                       (marshaller : Rpc.t -> 'a) =
+  let res = try
+      let t = internal_async_fn () in `new_api t
+    with
+      Api_errors.Server_error(code, params) when code=Api_errors.message_method_unknown -> `old_api
```
lindig: The exception will escape if then `when` condition is not met. Is this as expected? You would also do:
```
match internal_async_fn () with
| t ->
| exception Api_errors.Server_error(code, params) when code=Api_errors.message_method_unknown ->
| exception other ->
```

ocaml/xapi/server_helpers.ml
```diff
@@ -37,17 +37,27 @@ let async_wire_name = "Async."
...
   tbl
 
-let is_async x =
-  String.length x > async_length && (String.sub x 0 async_length = async_wire_name)
-
-let remove_async_prefix x =
-  String.sub x async_length (String.length x - async_length)
+(** removes Async. or InternalAsync. prefixes if they exist
+  * NB. X_y api call format does not work with Async (and didn't previously) *)
+let sync_ty_and_maybe_remove_prefix x =
```
lindig: Are these optimizations worth it? I would have used `Astring.String.is_prefix` because it avoids creating string copies like `String.sub`

ocaml/xapi/helpers.ml
lippirk: yes - this is as expected. the reason we catch the `message_method_unknown` is that we want to know when we are contact a version of xapi which doesn't know about internal async calls, in which case we revert to a synchronous call.

i'll add a comment explaining this, as it's not clear

ocaml/xapi/server_helpers.ml
lippirk: i should definitely be using `is_prefix`!

it is necessary to remove the `Async.` or `InternalAsync.` suffices because the pattern match on the method name in `server.ml` is not expecting them.

ocaml/xapi/server_helpers.ml
```diff
@@ -37,17 +37,27 @@ let async_wire_name = "Async."
...
+  let len_x = String.length x in
+  (* expect more async calls than internal_async, so check for async first *)
+  if len_x > async_length && Astring.String.is_prefix ~affix:async_wire_name x then
```
psafont: Is it the first check needed for performance? (It's not needed functionally)

robhoes:
> i'm still not sure why this didn't work, but i've undone it now. in retrospect this was not worth changing at all

Perhaps the use of `!=` rather than `<>`?

robhoes:
I think that you need a rebase.

lippirk:
> > i'm still not sure why this didn't work, but i've undone it now. in retrospect this was not worth changing at all
> 
> Perhaps the use of `!=` rather than `<>`?

yup... sigh
```
utop # "asdf" != "asdf";;
- : bool = true
utop # "asdf" <> "asdf";;
- : bool = false
```

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,69 @@ let host_supports_hvm ~__context host =
 let env_with_path env_vars =
   Array.of_list @@
   List.map (fun (k,v) -> Printf.sprintf "%s=%s" k v) (("PATH", String.concat ":" Forkhelpers.default_path)::env_vars)
+
+module Task  = struct (* can't place these functions in task helpers due to circular dependencies *)
+  let wait_for ~__context ~tasks =
+    let our_task = Context.get_task_id __context in
+    (** note that in this definition of `classes` we might duplicate our_task
+      * it turns out that filtering out our_task breaks _at least_ `VM.revert`, so best to leave it like this *)
```
robhoes: I'm sceptical :)

ocaml/xapi/server_helpers.ml
lippirk: no, i can get rid of it (it's a hangover from how it was done before - using `String.sub`)

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,69 @@ let host_supports_hvm ~__context host =
...
+      let t = internal_async_fn () in `new_api t
+    with
+      Api_errors.Server_error(code, params) when code=Api_errors.message_method_unknown -> `old_api
+  in
+  match res with
+  | `old_api ->
+    D.info "%s: InternalAsync call was not known by destination - trying sync call instead" debug_tag;
+    sync_fn ()
+  | `new_api t ->
+    if t != Context.get_task_id __context then D.error "try_internal_async: expected InternalAsync task to match ours";
```
robhoes: Here's another `!=`...

ocaml/xapi/helpers.ml
lippirk: yup, removed this comment

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,67 @@ let host_supports_hvm ~__context host =
...
+                       (sync_fn : unit -> 'a) (* only used when we are talking to a version of the api without InternalAsync *)
+                       (marshaller : Rpc.t -> 'a) =
+  let res = try
+      let t = internal_async_fn () in `new_api t
+    with
+      Api_errors.Server_error(code, params) when code=Api_errors.message_method_unknown -> `old_api
+  in
+  match res with
+  | `old_api ->
+    D.info "%s: InternalAsync call was not known by destination - trying sync call instead" debug_tag;
```
robhoes: Do you really need to manually add the `debug_tag`? Since the calls are all wrapped in `Debug.with_thread_associated` (in server_helpers.ml), this should get included automatically. Also, you don't need the `D.` prefix, as the module is opened.

ocaml/xapi/server_helpers.mli
```diff
@@ -0,0 +1,42 @@
+val exec_with_new_task : ?http_other_config:(string * string) list ->
```
robhoes: You need to add the license and copyright stuff to the top of this new file.

ocaml/xapi/server_helpers.ml
```diff
@@ -58,7 +67,7 @@ let parameter_count_mismatch_failure func expected received =
 
 (** WARNING: the context is destroyed when execution is finished if the task is not forwarded, in database and not called asynchronous. *)
 (*  FIXME: This function should not be used for external call : we should add a proper .mli file to hide it. *)
-let exec_with_context ~__context ?marshaller ?f_forward ?(called_async=false) ?(has_task=false) f =
+let exec_with_context ~__context ~need_complete ?marshaller ?f_forward ?(called_async=false) f =
   (* Execute fn f in specified __context, marshalling result with "marshaller".
      If has_task is set then __context has a real task in it that has to be completed. *)
```
robhoes: `has_task` is gone.

robhoes:
You are no longer using a subtask, probably because you had issues with the task progress reporting. The problem is now that the main task gets completed early, while the master is still doing some work (e.g. resetting operations when coming out of `with_vm_operation`). I am worried that this will lead to subtle problems.

ocaml/xapi/server_helpers.ml
lippirk: yes,

we demand the caller compute `need_complete` instead, as that is all that `has_task` was used for

lippirk:
I've made changes to put the internal async request inside a subtask. testing now

ocaml/xapi/server_helpers.ml
```diff
@@ -164,9 +182,7 @@ let exec_with_forwarded_task ?http_other_config ?session_id ?origin task_id f =
     (fun ~__context -> f __context)
 
 let exec_with_subtask ~__context ?task_in_database ?task_description task_name f =
```
robhoes: It may be easier and cleaner to wrap the client call (on the master) in `exec_with_subtask`, and leave the server as it was.

lippirk: yes, that would be sensible

lippirk: though i need to make sure that if we fallback to the synchronous call, then we are not using a subtask

ocaml/xapi/server_helpers.ml
robhoes: Yep, so that code comment needs to go away.

lippirk: ah - done

lippirk:
@robhoes when you cancel a parent task will all subtasks get cancelled as well?

if not then i may need to make a further amendment

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,102 @@ let host_supports_hvm ~__context host =
...
+      TaskHelper.exn_if_cancelling ~__context; (* First check if _we_ have been cancelled *)
+      let statuses = List.filter_map (fun task -> try Some (Db.Task.get_status ~__context ~self:task) with _ -> None) tasks in
+      let unfinished = List.exists (fun state -> state = `pending) statuses in
+      if unfinished
+      then begin
+        let from = call_api_functions ~__context
+            (fun rpc session_id -> Client.Client.Event.from ~rpc ~session_id ~classes ~token ~timeout:30.0) in
+        debug "Using events to wait for tasks: %s" (String.concat "," classes);
+        let from = Event_types.event_from_of_rpc from in
+        cb (); (** be careful not to include logic in `cb` that could modify tasks we are waiting on indefinitely *)
```
lippirk: realised that we have to be careful here to not unnecessarily change the tasks we are listening for events on

as a precaution, i made `mirror_progress` more future proof - we don't want to mirror our progress on our own task

edwintorok:
How did the testing go of this PR?

lippirk:
> How did the testing go of this PR?

it was green! still waiting on another review from @robhoes

PR #4017, opened 2020-01-21
CP-32863: storage driver domain support

edwintorok:
Draft PR, needs more testing in particular with XenCenter.
See commit messages for explanations.

The goal here is to allow storage driver domain to be cleanly shut down at the appropriate time in the shutdown sequence.
Storage driver domain provide storage (backend) to other VMs, so they can only be shut down once no VM is using them, and their associated PBD is unplugged.

The usual shutdown process first shuts the VMs down and then unplugs PBDs, so this needs an adjustment to run  in 2 phases (VMs except driver domains, PBD unplug + VMs that are driver domains).

Also add protection against accidentally shutting the VM down (e.g. from XenCenter, or as a side-effect of other operations done by XenCenter such as host shutdown/reboot).

TBD: XenCenter still tries to shut this VM down (and now correctly fails instead of crashing the other VMs), we might have to mark the VM as a control domain too, not just system domain.


ocaml/xapi/vm_evacuation.ml
```diff
@@ -34,6 +34,10 @@ let ensure_no_vms ~__context ~rpc ~session_id ~evacuate_timeout =
     get_resident_vms ~__context ~self:host |> snd
     |> List.filter (fun vm -> is_running vm && not (self_managed_poweroff vm))
   in
+  let get_driver_domains () =
+    get_resident_vms ~__context ~self:host |> fst
+    |> List.filter (fun vm -> is_running vm && not (Db.VM.get_is_control_domain ~__context ~self:vm))
```
lippirk: to me this filter reads as 'all running domUs', but i imagined your driver domains would be a strict subset of your domUs?

lippirk: ah i understand from `get_resident_vms` why this works... that name is not the best

ocaml/xapi/xapi_vm_lifecycle.ml
```diff
@@ -223,6 +223,13 @@ let check_snapshot ~vmr ~op ~ref_str =
   then None
   else Some (Api_errors.vm_is_snapshot, [ref_str; Record_util.vm_operation_to_string op])
 
+let check_system_domain ~vmr ~op ~ref_str ~pbd =
+  let allowed = [`start; `start_on; `unpause; `data_source_op; `update_allowed_operations;
+                 `assert_operation_valid; `power_state_reset; `query_services; `snapshot ] in
```
lippirk: can we hide snapshot behind some development flag?

ocaml/xapi/vm_evacuation.ml
edwintorok: Yes, the use of `fst` and `snd` here is confusing, I'll use a let binding instead to make it more explicit what the 2 sides mean.

ocaml/xapi/xapi_vm_lifecycle.ml
edwintorok: I dropped entire special casing of driver domains, just treating them as control domains now. Which is what the documentation for control domains in the IDL says: domain 0 or driver domains.

ocaml/xapi/export.ml
```diff
@@ -556,17 +556,26 @@ let metadata_handler (req: Request.t) s _ =
...
-           with e ->
-             export_error := Some e)
+       let writer_thread = Thread.create (
+           Debug.with_thread_named "metadata export writer thread" (fun () ->
+             (* lock all the VMs before exporting their metadata *)
+             let needs_close = ref true in
+             try
+               List.iter (fun vm -> lock_vm ~__context ~vm ~task_id `metadata_export) vm_refs;
+               finally
+                 (fun () -> export_metadata ~with_snapshot_metadata:export_snapshots ~preserve_power_state:true ~include_vhd_parents ~__context ~vms:vm_refs write_fd)
```
lindig: Is usually `export_metadata` closing the fd but is not in the case of an exception? That looks like a problem in `export_metadata`. If not, could the closing of the fd not be handled by an outer `finally` without a `needs_close` logic?

ocaml/xapi/vm_evacuation.ml
```diff
@@ -105,7 +105,19 @@ let ensure_no_vms ~__context ~rpc ~session_id ~evacuate_timeout =
 
       evacuate ());
 
-  log_and_ignore_exn (fun () -> get_running_domains () |> shutdown)
+  log_and_ignore_exn (fun () -> get_running_domains () |> shutdown);
+
+  (* The driver domains shouldn't be in use at this point, so we should be able to
+   * shut them down *)
+  log_and_ignore_exn (fun () ->
```
lindig: This could be made syntactically nicer:
```
 log_and_ignore_exn @@ fun () -> ..
```

ocaml/xapi/export.ml
edwintorok: We are closing it in both situations, but we want to close it sooner in case lock_vm succeeded (i.e. before unlock_vm). This is all very convoluted, and there is still a bug here when lock_vm fails halfway through: we do not unlock any of the VMs that were locked.

So instead I got rid of the 2 close situations and just close it in the finally.
`lock_vm` now stores a list of VMs it has successfully locked. When done we go through the VMs that we locked and unlock them: this works both when everything succeeds (full list as before), everything fails (empty list), or partial success (we only unlock the VMs that we have previously successfully locked).

psafont:
The base branch should be rebased to keep up with master as well.

ocaml/xapi/vm_evacuation.ml
```diff
@@ -105,7 +105,19 @@ let ensure_no_vms ~__context ~rpc ~session_id ~evacuate_timeout =
...
       evacuate ());
 
-  log_and_ignore_exn (fun () -> get_running_domains () |> shutdown)
+  log_and_ignore_exn (fun () -> get_running_domains () |> shutdown);
+
+  (* The driver domains shouldn't be in use at this point, so we should be able to
+   * shut them down *)
+  log_and_ignore_exn @@ fun () ->
+      get_resident_vms ~__context ~self:host |> fst
+      |> List.filter_map (fun vm ->
```
lindig: This is mixing a functional map with a side effect inside the predicate. I think I would prefer to filter first and to iterate separately.

edwintorok: I've split this so it first does the filtering for pbds, vms and then the unplugs and shutdowns.

edwintorok:
Thanks, I'll merge this and then update the feature branch.

PR #4018, opened 2020-01-22
REQ-627 CA-333495 add Xapi_pci.dequarantine

lindig:
Before we can bind PCI drivers we need to dequarantine the device. This
is done by calling into Xenopsd to let it do the work. (Hence, this must
happen on the host where the device is.) We employ this in the context
of setting up Nvidia SRIOV devices.

This PR depends on https://github.com/xapi-project/xcp-idl/pull/302 which has been reverted but needs to be active. Because of cross-repo dependencies, Travis is broken.

### DIFF
--- a/ocaml/xapi/xapi_pgpu.ml
+++ b/ocaml/xapi/xapi_pgpu.ml
@@ -375,6 +375,8 @@ let nvidia_vf_setup ~__context ~pf ~enable =
   let bind_path = "/sys/bus/pci/drivers/nvidia/bind" in
   let unbind_path pci = sprintf "/sys/bus/pci/devices/%s/driver/unbind" pci in
   let pci       = Db.PCI.get_pci_id ~__context ~self:pf in
+  let addr_of pci = Xenops_interface.Pci.address_of_string pci in
+  let dequarantine pci = Xapi_pci.dequarantine ~__context (addr_of pci) in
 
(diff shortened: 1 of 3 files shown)


ocaml/xapi/xapi_pgpu.ml
```diff
@@ -422,6 +426,7 @@ let nvidia_vf_setup ~__context ~pf ~enable =
       raise (fail msg)
     | Some 0 when Sys.file_exists script ->
       debug "PCI %s has 0 VFs - calling %s" pci script;
+      dequarantine pci;
```
edwintorok: I think this is redundant, because this is always called after `bind_to_nvidia`, but since this is an idempotent operation it is good practice to do it whenever we "assert" that the device needs to be in Dom0.

robhoes: It still isn't redundant if the device was already bound to the nvidia driver, because `bind_to_nvidia` does not call `dequarantine` then. However, it may be simpler to just always dequarantine in `bind_to_nvidia` (as the first thing that is done in that function), and not here anymore, so that there is just a single point where the `dequarantine` function is called.

lindig:
The fixup commit move the de-quarantining earlier into the control flow - thus reducing the number of calls sites. This commit will be squashed before I merge it.

PR #4022, opened 2020-01-31
CP-32686: Native certificate installation

psafont:
This adds ocaml code equivalent to the `scripts/generate_ssl_cert` script.

Needs https://github.com/xapi-project/stdext/pull/43 to work correctly

### DIFF
--- a/scripts/generate_ssl_cert
+++ b/scripts/generate_ssl_cert
@@ -64,14 +64,11 @@ openssl req -batch -new -key privkey.rsa -days 3650 -config config -out cert.csr
 # the extensions too, but "req" does not allow the -extfile option.)
 openssl x509 -extfile config -req -signkey privkey.rsa -in cert.csr -outform PEM -out signedpubcert.pem -days 3650
 
-openssl dhparam 512 > dh.pem
-
 popd
 
 tmpcert="${DIR}/tmpcertcombined.pem"
 
-(cat ${DIR}/privkey.rsa; echo ""; cat ${DIR}/signedpubcert.pem; \
- echo ""; cat ${DIR}/dh.pem) > "${tmpcert}"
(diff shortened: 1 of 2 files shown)


scripts/generate_ssl_cert
```diff
@@ -45,11 +45,11 @@ subjectAltName = @alt_names
 [alt_names]
 @eof
 
-i=0
+i=1
 for x in $dnsnames
 do
-	let i=i+1
 	echo "DNS.${i}=${x}" >> config
+	(( i=i+1 ))
```
lindig: `((i++))` does work as well

scripts/generate_ssl_cert
```diff
@@ -64,14 +64,11 @@ openssl req -batch -new -key privkey.rsa -days 3650 -config config -out cert.csr
 # the extensions too, but "req" does not allow the -extfile option.)
 openssl x509 -extfile config -req -signkey privkey.rsa -in cert.csr -outform PEM -out signedpubcert.pem -days 3650
 
-openssl dhparam 512 > dh.pem
-
 popd
 
 tmpcert="${DIR}/tmpcertcombined.pem"
```
lindig: Should this use `mktemp`?

scripts/generate_ssl_cert
psafont: it does

scripts/generate_ssl_cert
psafont: Better to use it, even if It's bound to be replaced by the ocaml code

psafont: privkey.rsa and signedpubcert.pem should also use mktemp as well.
The fact that they do not get deleted after they get consolidated into the single cert is also iffy.

I wonder if something else depends on those files being present.

lindig:
The webhook is complaining about sign-offs - I suspect that is the fixup. Can we squash it?

PR #4023, opened 2020-02-04
Fix parsing of platform:cores-per-socket to avoid divide-by-0 error

andyhhp:
[root@lcy2-dt106 ~]# xe vm-start uuid=39b8d59d-7cc7-7e71-8107-28ba5e2ad74e
  The server failed to handle your request, due to an internal error. The given message may give details useful for debugging the problem.
  message: Division_by_zero

Also reject a negative cores-per-socket, which is equally nonsensical, and
only gets caught by the sanity checks in the Xen:

  [root@lcy2-dt106 ~]# xe vm-start uuid=39b8d59d-7cc7-7e71-8107-28ba5e2ad74e
  The server failed to handle your request, due to an internal error. The given message may give details useful for debugging the problem.
  message: xenopsd internal error: (Failure
    "Error from xenguesthelper: xenguest: xc_domain_set_cores_per_socket: [22]
    Invalid argument")

Also take the opportunity to make the error messages helpful, by returning the
current values which are being objected to.

### DIFF
--- a/ocaml/xapi/vm_platform.ml
+++ b/ocaml/xapi/vm_platform.ml
@@ -145,16 +145,22 @@ let sanity_check ~platformdata ?firmware ~vcpu_max ~vcpu_at_startup ~domain_type
   end;
   if check_cores_per_socket && (List.mem_assoc "cores-per-socket" platformdata) then
     begin
+      let cps_str = List.assoc "cores-per-socket" platformdata in
+      let vcpus = Int64.to_int(vcpu_max) in
       try
-        let cores_per_socket = int_of_string(List.assoc "cores-per-socket" platformdata) in
+        let cps = int_of_string(cps_str) in
+
         (* VCPUs_max has to be a multiple of cores per socket *)
-        if ((Int64.to_int(vcpu_max) mod cores_per_socket) <> 0) then
-          raise (Api_errors.Server_error(Api_errors.invalid_value,
-                                         ["platform:cores-per-socket";
-                                          "VCPUs_max must be a multiple of this field"]))
+        if cps < 1 || ((vcpus mod cps) <> 0) then
+          raise (Api_errors.Server_error(
+                     Api_errors.invalid_value,
+                     [Printf.sprintf "platform:cores-per-socket (value %d)" cps;
+                      Printf.sprintf "VCPUs_max (value %d) must be a multiple of cores-per-socket" vcpus]))
       with Failure msg ->
-        raise (Api_errors.Server_error(Api_errors.invalid_value, ["platform:cores-per-socket";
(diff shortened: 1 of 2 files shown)


ocaml/xapi/vm_platform.ml
```diff
@@ -145,16 +145,22 @@ let sanity_check ~platformdata ?firmware ~vcpu_max ~vcpu_at_startup ~domain_type
   end;
   if check_cores_per_socket && (List.mem_assoc "cores-per-socket" platformdata) then
     begin
+      let cps_str = List.assoc "cores-per-socket" platformdata in
+      let vcpus = Int64.to_int(vcpu_max) in
       try
-        let cores_per_socket = int_of_string(List.assoc "cores-per-socket" platformdata) in
+        let cps = int_of_string(cps_str) in
```
psafont: Negative integers can be rejected at this stage with Scanf:
```
# Scanf.sscanf "-4" "%u" Fun.id;;
Exception:
Stdlib.Scanf.Scan_failure
 "scanf: bad input at char number 0: character '-' is not a decimal digit".
```

edwintorok: Not necessarily a user friendly error though

ocaml/xapi/vm_platform.ml
```diff
@@ -145,16 +145,22 @@ let sanity_check ~platformdata ?firmware ~vcpu_max ~vcpu_at_startup ~domain_type
...
-        if ((Int64.to_int(vcpu_max) mod cores_per_socket) <> 0) then
-          raise (Api_errors.Server_error(Api_errors.invalid_value,
-                                         ["platform:cores-per-socket";
-                                          "VCPUs_max must be a multiple of this field"]))
+        if cps < 1 || ((vcpus mod cps) <> 0) then
+          raise (Api_errors.Server_error(
+                     Api_errors.invalid_value,
+                     [Printf.sprintf "platform:cores-per-socket (value %d)" cps;
+                      Printf.sprintf "VCPUs_max (value %d) must be a multiple of cores-per-socket" vcpus]))
       with Failure msg ->
```
psafont: Seeing that code is guarding against negative values for `cps` I think it should do the same for 0. The modulo operation can fail with the exception `Division_by_zero` otherwise.

The exception can be caught using `with Failure _ | Division_by_zero ->` here

edwintorok: `cps < 1` is true when `cps = 0`, and short-circuiting the `||` is part of the language specification: https://caml.inria.fr/pub/docs/manual-ocaml/expr.html#sec147

psafont: fair enough

PR #4024, opened 2020-02-05
CP-32681: Add certificates to DB schema

psafont:
This allows the xapi to easily expose the server certificates metadata
to clients. In particular their validity period and fingerprint.

There a quite a lot of pulls and levers to configure when adding a class to the schema db, althought they can be changed later on
So there are some things I'm don't know about:
- `~gen_constructor_destructor` what would it be needed for? I would only like to create an object when the certificate is installed.
- I'm unsure whether we want to force a 1 to 1 relationship between host and server certificate. I've added it as a one-to-many in case this could change in the future.

In addition I'm not fond of using the hand-crafted `Date` module from stdext, seems pretty limited.

### DIFF
--- a/ocaml/idl/datamodel_certificate.ml
+++ b/ocaml/idl/datamodel_certificate.ml
@@ -1,0 +1,51 @@
+(*
+ * Copyright (C) 2020 Citrix Systems Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+open Datamodel_types
+open Datamodel_common
+open Datamodel_roles
+
+let lifecycle = [Published, rel_stockholm, ""]
+
+let t =
+  create_obj
+    ~name: _certificate
+    ~descr:"Description"
+    ~doccomments:[]
+    ~gen_constructor_destructor:false
+    ~gen_events:true
+    ~in_db:true
+    ~lifecycle
+    ~persist:PersistEverything
+    ~in_oss_since:None
+    ~messages_default_allowed_roles:_R_READ_ONLY
+    ~contents:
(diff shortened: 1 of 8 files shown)


robhoes:
The option `~gen_constructor_destructor:true` adds predefined `create` and `destroy` functions to your class' datamodel. In this case, you don't want that, because we manage these objects in a different way, e.g. `certificates` objects get created as a result of calling another API call.

robhoes:
> I'm unsure whether we want to force a 1 to 1 relationship between host and server certificate. I've added it as a one-to-many in case this could change in the future.

I agree that we should allow a host to have multiple certificates in the datamodel (even if we use it as 1:1 in the first instance).

ocaml/xapi/records.ml
```diff
@@ -2181,3 +2181,31 @@ let cluster_host_record rpc session_id cluster_host =
...
+  ; record    = x
+  ; getref    = (fun () -> !_ref)
+  ; fields =
+    [ make_field ~name:"uuid"
+      ~get:(fun () -> (x ()).API.certificate_uuid)
+      ()
+    ; make_field ~name:"host"
+      ~get:(fun () -> (x ()).API.certificate_host |> get_uuid_from_ref)
+      ()
+    ; make_field ~name:"not_before"
```
robhoes: We use `-` rather than `_` in the CLI.

robhoes:
Perhaps we should create the file `datamodel_certificates.ml` for this. `datamodel.ml` had become too big and messy, so we started splitting it up.

psafont:
> I agree that we should allow a host to have multiple certificates in the datamodel (even if we use it as 1:1 in the first instance).
If we want to model the currently-existing trusted certificates (including CRLs) we can add a relationship between a certificate and a pool as well.

I don't think it's needed right now, but it's something to keep in mind.

ocaml/idl/datamodel_certificate.ml
```diff
@@ -0,0 +1,51 @@
...
+    ~in_db:true
+    ~lifecycle
+    ~persist:PersistEverything
+    ~in_oss_since:None
+    ~messages_default_allowed_roles:_R_READ_ONLY
+    ~contents:
+      [ uid   _certificate ~lifecycle
+      ; field   ~qualifier:StaticRO ~lifecycle
+          ~ty:(Ref _host) "host" ~default_value:(Some (VRef null_ref))
+          "The host where the certificate installed"
```
lindig: Is this missing a "gets" or "is" before "installed"?

psafont: yep!

PR #4025, opened 2020-02-06
CA-334797: Disable TLSv1.3 when legacy is true

minglumlu:
This fix is to disable TLSv1.3 when legacy is true.


lippirk:
@minglumlu i about to change xapi so that legacy ssl is not supported, you may want to cherry pick that instead when i am done

PR #4029, opened 2020-02-07
Remove legacy ssl

lippirk:


ocaml/xapi/xapi_sync.ml
```diff
@@ -35,12 +35,8 @@ let sync_host ~__context host =
...
           and session = Xapi_session.slave_login ~__context ~host:(Helpers.get_localhost ~__context) ~psecret:!Xapi_globs.pool_secret in
           Unix.putenv "XSH_SESSION" (Ref.string_of session);
-          Unix.putenv "XSH_SSL_LEGACY" (string_of_bool (Db.Host.get_ssl_legacy ~__context ~self:host));
-          (match !Xapi_globs.ciphersuites_good_outbound with
-           | "" -> raise (Api_errors.Server_error (Api_errors.internal_error,["Xapi_sync found no good ciphersuites in Xapi_globs."]))
-           | c -> Unix.putenv "XSH_GOOD_CIPHERSUITES" c
-          );
-          Unix.putenv "XSH_LEGACY_CIPHERSUITES" !Xapi_globs.ciphersuites_legacy_outbound;
+
+          Unix.putenv "XSH_SSL_LEGACY" "false";
```
lippirk: i should possibly remove this altogether

ocaml/xapi/message_forwarding.ml
```diff
@@ -2162,18 +2158,7 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
-        else None
-      in
-      let local_fn = Local.Host.set_ssl_legacy ~self ~value in
-      let fn () =
-        do_op_on ~local_fn ~__context ~host:self
-          (fun session_id rpc ->
-             Client.Host.set_ssl_legacy rpc session_id self value)
-      in
-      tolerate_connection_loss fn success 30.
+      Local.Host.set_ssl_legacy ~__context ~self ~value
```
psafont: Can't this just be a no-op?

lippirk: `Local.Host.set_ssl_legacy` is equivalent to `Xapi_host.set_ssl_legacy` which I have changed to be a no-op

ocaml/xapi-types/features.mli
```diff
@@ -50,7 +50,7 @@ type feature =
   | Xen_motion                   (** Enable XenMotion feature *)
   | Guest_ip_setting             (** Enable use of Guest ip seting *)
   | AD                           (** Enable use of Active Directory *)
-  | Ssl_legacy_switch            (** Enable the control switch for SSL/TLS legacy-mode. *)
+  | Ssl_legacy_switch            (** Enable the control switch for SSL/TLS legacy-mode - no longer supported *)
```
lippirk: delete this variant

lippirk:
TODO:
1. amend `init.d-xapissl` to remove SSLv2 & EXTRA_CIPHERS

```
scripts/init.d-xapissl
98:ciphers = !SSLv2:${GOOD_CIPHERS}${EXTRA_CIPHERS}
```

2. rename `init.d-xapissl` to match its actual name when the hypervisor is installed on a host

3. the following script will be broken:

```
scripts/poweron/iLO.py
30:     # It would be nice to specify all protocols ("sslv23") and then use
31:     # options to disable sslv2 and sslv3, but such a Context fails to
37:             M2Crypto.m2.SSL_OP_NO_SSLv2 |
40:     ctx.set_cipher_list("!SSLv2:RSA+AES128-SHA256:RSA+AES256-SHA:RSA+AES128-SHA:RSA+RC4-SHA")

```

ocaml/xapi/message_forwarding.ml
psafont: It can raise an exception, which is why I'm asking if it can be a no-op

lippirk: ah i see. well it could be, but a user would be confused if they see a `set_ssl_legacy ~value:true` command "succeed", but not actually do anything

ocaml/idl/datamodel_host.ml
```diff
@@ -1458,7 +1458,7 @@ let host_query_ha = call ~flags:[`Session]
...
            field ~qualifier:DynamicRO ~lifecycle:[Published, rel_inverness, ""] ~ty:(Set (Ref _pusb)) "PUSBs" "List of physical USBs in the host";
-           field ~qualifier:StaticRO ~lifecycle:[Published, rel_dundee, ""] ~ty:Bool ~default_value:(Some (VBool true)) "ssl_legacy" "Allow SSLv3 protocol and ciphersuites as used by older server versions. This controls both incoming and outgoing connections. When this is set to a different value, the host immediately restarts its SSL/TLS listening service; typically this takes less than a second but existing connections to it will be broken. API login sessions will remain valid.";
+           field ~qualifier:StaticRO ~lifecycle:[Published, rel_dundee, ""; Deprecated, rel_stockholm, "Legacy SSL no longer supported"] ~ty:Bool ~default_value:(Some (VBool true)) "ssl_legacy" "Allow SSLv3 protocol and ciphersuites as used by older server versions. This controls both incoming and outgoing connections. When this is set to a different value, the host immediately restarts its SSL/TLS listening service; typically this takes less than a second but existing connections to it will be broken. API login sessions will remain valid.";
```
edwintorok: Should default value be false now?

ocaml/idl/datamodel_pool.ml
```diff
@@ -576,6 +577,7 @@ open Datamodel_types
       ~in_oss_since:None
       ~lifecycle:[
         Published, rel_dundee, "";
+        Deprecated, rel_stockholm, "Legacy SSL no longer supported"
       ]
       ~params:[Ref _pool, "self", "(ignored)";]
       ~doc:"Sets ssl_legacy true on each host, pool-master last. See Host.ssl_legacy and Host.set_ssl_legacy."
```
edwintorok: The docs are wrong here, it should set it to `false`

ocaml/tests/binpack_test.ml
```diff
@@ -107,7 +107,7 @@ let try_impossible_cases () =
     try
       check_plan config dead_hosts bad_plan;
       failwith "bad plan was not detected"
-    with (Failure "bad plan") -> ()
+    with (Failure e) when e = "bad plan" -> ()
```
edwintorok: Shouldn't we have a dedicated exception BadPlan instead?

ocaml/xapi/xapi_host.ml
```diff
@@ -640,6 +640,7 @@ let is_host_alive ~__context ~host =
 
 let create ~__context ~uuid ~name_label ~name_description ~hostname ~address ~external_auth_type ~external_auth_service_name ~external_auth_configuration ~license_params ~edition ~license_server ~local_cache_sr ~chipset_info ~ssl_legacy =
 
+  let ssl_legacy = false in
```
edwintorok: You could move this below to where it is used

ocaml/xapi/xapi_mgmt_iface.mli
```diff
@@ -41,5 +41,5 @@ val rebind : __context:Context.t -> unit
 (** Start a server thread on the given HIMN address if the server is not yet running *)
 val enable_himn : __context:Context.t -> addr:string -> unit
 
-(** Restart stunnel to make it pick up a change to host.ssl_legacy *)
+(** Restart stunnel *)
```
edwintorok: Is this function obsolete now? If it has no other callers we can delete it too

ocaml/xe-cli/newcli.ml
```diff
@@ -233,16 +227,7 @@ let parse_args =
...
 let open_tcp_ssl server =
   let port = get_xapiport true in
-  debug "Connecting via%s stunnel to [%s] port [%d]%s\n%!"
-    (if !allow_ssl_legacy then " legacy-mode" else "") server port
-    (match !ciphersuites with None -> "" | Some c -> " with ciphersuites "^c);
-  Stunnel.set_legacy_protocol_and_ciphersuites_allowed !allow_ssl_legacy;
-  (match !ciphersuites with
-   | None -> ()
-   | Some c -> (* Use only the specified ones, none of Stunnel's built-in defaults. *)
-     Stunnel.set_good_ciphersuites c;
```
edwintorok: I assume that you'll modify stunnel in xapi-stdext to have a good set of default good ciphersuites

scripts/init.d-xapissl
```diff
@@ -24,8 +24,6 @@ PEMFILE="@ETCDIR@/xapi-ssl.pem"
 SSLPIDFILE="/var/run/xapissl.pid"
```
edwintorok: You can also drop the whole comment section at the top, the include of lsb init functions, or just in general we do not need this to be a shell script at all, we could move it all inside ocaml, all it needs to do is generate a config file, and call a script to potentially generate a certificate.

With the new stunnel soon to be merged there will also be a systemd service `stunnel@` that we could use, perhaps open a ticket to do this work independent of what you are currently doing.

ocaml/idl/datamodel_host.ml
lippirk: i'll change this to `false` for clarity

i believe this shouldn't make a difference (my understanding is `default_value` will only get used when upgrading to a version where this field was introduced to the schema)

ocaml/xapi/xapi_mgmt_iface.mli
psafont: This function may be useful to reload certificates, don't delete it just yet.

ocaml/xe-cli/newcli.ml
lippirk: i will check that the 'good ciphersuites' in stdext are the same as in xapi

ocaml/xapi/xapi_mgmt_iface.mli
lippirk: yes

ocaml/xe-cli/newcli.ml
edwintorok: Might be better to find a 'home' for the ciphersuite definition, and have every other daemon/etc. use that. For example this could be put somewhere in `xapi-idl`.

lippirk: this provides another reason for moving the bash scripts to ocaml

scripts/init.d-xapissl
lippirk: i will make a ticket for this

psafont:
What's up with the rebase?

lippirk:
> What's up with the rebase?

i rebased against master (this branch is feature/req-811/master)

ocaml/xapi/sparse_dd_wrapper.ml
```diff
@@ -71,15 +71,8 @@ let dd_internal progress_cb base prezeroed infile outfile size =
...
-                            | "" -> raise (Api_errors.Server_error
-                                             (Api_errors.internal_error,["Vdi_copy found no good ciphersuites in Xapi_globs."]))
-                            | s -> s
-                          );
-                        "-legacy-ciphersuites"; !Xapi_globs.ciphersuites_legacy_outbound
-                      ] @ (if Stunnel.is_legacy_protocol_and_ciphersuites_allowed () then [ "-ssl-legacy" ] else []
-                          ) @ (if prezeroed then [ "-prezeroed" ] else []
-                              ) @ (Opt.default [] (Opt.map (fun x -> [ "-base"; x ]) base)) in
+                      ] @ if prezeroed then [ "-prezeroed" ] else []
+                        @ Opt.default [] (Opt.map (fun x -> [ "-base"; x ]) base) in
```
psafont: Could you change `Opt` for `Option` here?

scripts/init.d-xapissl
```diff
@@ -40,10 +39,6 @@ fi
 [ $(which stunnel4) ] && STUNNEL=$(which stunnel4)
 [ -z ${STUNNEL} ] && STUNNEL=$(which stunnel)
```
psafont: I don't think we've had an `stunnel4` binary for a long time; `STUNNEL=$(command -v stunnel)` could be used instead

lippirk:
i have removed the commit which changed the iLO script - hopefully we can remove that altogether

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -40,16 +40,12 @@ let stunnel_m = Mutex.create ()
 
 let restart_stunnel_nomutex ~__context ~accept =
   info "Restarting stunnel (accepting connections on %s)" accept;
-  let back_compat ~__context =
-    if Stunnel.is_legacy_protocol_and_ciphersuites_allowed ()
-    then [ "back_compat_6_5" ]
-    else []
+  let xapissl_args = [ "restart"
+                     ; accept
+                     ; "permfile=" ^ !Xapi_globs.server_cert_path
```
psafont: is `permfile` the correct name for the parameter? It looks like it should be `pemfile` since it's the path of the pem (certificate file)

scripts/init.d-xapissl
```diff
@@ -111,30 +89,13 @@ debug = authpriv.5
...
-            # With multiple sslVersion entries, only the last is used.
-            writec 'sslVersionMax = TLSv1.2'
-            writec 'sslVersionMin = TLSv1'
-
-            # If we had v5.06 or newer we would need to allow SSLv3 with
-            # writec 'options -NO_SSLv3'
-            # but all earlier versions choke on that option.
-        fi
-    fi
+    writec 'sslVersionMin = TLSv1.2'
```
edwintorok: Add `sslVersionMax = TLSv1.2` too, otherwise you also enable TLSv1.3 and we're not ready for that one yet.

ocaml/xapi/xapi_mgmt_iface.ml
lippirk: i'll fix this when i convert this script to ocaml

lippirk:
this should be good to merge now hopefully - i plan on putting the bash rewrites in another PR since this one is already quite large

ocaml/xapi/xapi_pool.ml
```diff
@@ -2139,9 +2133,7 @@ let set_ssl_legacy_on_each_host ~__context ~self ~value =
   in
   Xapi_pool_helpers.call_fn_on_slaves_then_master ~__context f
 
-let disable_ssl_legacy = set_ssl_legacy_on_each_host ~value:false
-
-let enable_ssl_legacy = set_ssl_legacy_on_each_host ~value:true
+let disable_ssl_legacy ~__context ~self = warn "disable_ssl_legacy: doing nothing"
```
edwintorok: You can also delete `set_ssl_legacy_on_each_host`

ocaml/tests/binpack_test.ml
```diff
@@ -42,16 +44,16 @@ let check_plan config dead_hosts plan =
   let memory_remaining = account config.hosts config.vms plan in
   (*    List.iter (fun mem -> Printf.printf "%Ld\n" mem) free; *)
   (* No host should be overcommitted: *)
-  if List.fold_left (||) false (List.map (fun x -> x < 0L) (List.map snd memory_remaining)) then failwith "bad plan";
+  if List.fold_left (||) false (List.map (fun x -> x < 0L) (List.map snd memory_remaining)) then raise BadPlan;
```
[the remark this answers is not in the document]
lippirk: i think you suggested it :P 

i've put it in a 'maintenance' commit

PR #4030, opened 2020-02-10
CA-334951: Ignore 0-star recommendations from WLB

psafont:
Before, if all the recommendations were 0 for some reason xapi would needlessly migrate all the VMs.

This also introduces a small change in the API in the case an internal error happened when processing a retrieve_wlb_recommendations API call: now it's more consistent and more data is returned.

After talking with @liulinC it's clear that vm recommendation always contains both the `Stars` field and the `ZeroScore`. The code now expects those values to be congruent or it considers the whole response nonsense.

### DIFF
--- a/ocaml/xapi/workload_balancing.ml
+++ b/ocaml/xapi/workload_balancing.ml
@@ -388,7 +388,7 @@ let perform_wlb_request ?auth ?url ?enable_log ~meth ~params
   | Some s -> s
   | None -> raise_internal_error []
 
-let val_num i =
+let float_or_raise i =
   try
     ignore (float_of_string i);
     i
@@ -414,9 +414,9 @@ let retrieve_vm_recommendations ~__context ~vm =
         let source = "WLB" in
         let host_uuid = get_child_data "HostUuid" in
         let host = Db.Host.get_by_uuid ~__context ~uuid:host_uuid in
-        let id = get_child_data "RecommendationId" |> val_num in
+        let id = get_child_data "RecommendationId" |> float_or_raise in
         let zero_reason = get_child_data "ZeroScoreReason" in
-        let stars = get_child_data "Stars" |> val_num |> float_of_string in
+        let stars = get_child_data "Stars" |> float_or_raise |> float_of_string in
 
         let recommendation = match stars, zero_reason with
         | 0., reason ->
@@ -433,7 +433,8 @@ let retrieve_vm_recommendations ~__context ~vm =
         (host, recommendation)
       with
       | Xml_parse_failure error ->
-        (* let this parse error carry on upwards,  perform_wlb_request will catch it and check the rest of the xml for an error code *)
+        (* let this parse error carry on upwards, perform_wlb_request will
+           catch it and check the rest of the xml for an error code *)
         raise (Xml_parse_failure error)
       | Db_exn.Read_missing_uuid (_,_,_)
       | Db_exn.Too_many_values (_,_,_) ->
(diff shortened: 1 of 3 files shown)


ocaml/xapi/workload_balancing.ml
```diff
@@ -24,6 +24,16 @@ open D
 
 exception Xml_parse_failure of string
 
+(* This is a stub used in the pattern_matching below to silence a
+ * warning in the newer ocaml compilers *)
+exception Float_of_string_failure
```
lindig: Maybe add the string that we can't concert to this error such that we can report it.

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -772,49 +773,32 @@ let choose_host_uses_wlb ~__context =
...
+      let possible_rec (host, recommendation) =
+        match recommendation with
+        | Recommendation {source; id; score} ->
+            Some (host, score, id)
+        | _ ->
+            None
       in
+      let all_hosts = retrieve_wlb_recommendations ~__context ~vm ~snapshot
+        |> List.filter_map possible_rec
+        |> List.sort (fun (h, s, r) (h', s', r') ->
```
lindig:
```
(fun (_, s, _) (_, s', _) -> Float.compare s' s)
```

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -772,49 +773,32 @@ let choose_host_uses_wlb ~__context =
...
+              if s < s' then 1 else if s > s' then -1 else 0)
+      in
+      debug "Hosts sorted in priority: %s"
+        (List.fold_left
```
lindig:
```
all_hosts
|> List.map (fun (h,s,r) -> (Printf.sprintf "%s %f," (Db.Host.get_name_label ~__context ~self:h) s) 
|> String.concat ","
```

ocaml/xapi/workload_balancing.ml
psafont: The string of the Failure is simply `"float_of_string"`, unfortunately. We're not removing any information doing this.

ocaml/xapi/workload_balancing.ml
```diff
@@ -424,6 +443,9 @@ let retrieve_vm_recommendations ~__context ~vm =
       | Db_exn.Too_many_values (_,_,_) ->
         raise_malformed_response' "VMGetRecommendations"
           "Invalid VM or host UUID" "unknown"
+      | Float_of_string_failure ->
```
edwintorok: Seems like you've deleted the code that raises this  exception, can the exception and this match be deleted too?

psafont: Oops. The original exception can still be thrown.

Although it's unfortunate Stdlib raises a parametrized `Failure`, I think it's not worth having a special case to suppress the warning for matching against a string.

psafont: Checking again, it's not even needed as the `val_num` function already does this. I'm changing its name so it's obvious what is doing and removing the match case.

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -772,49 +773,30 @@ let choose_host_uses_wlb ~__context =
...
+      let possible_rec (host, recommendation) =
+        match recommendation with
+        | Recommendation {source; id; score} ->
+            Some (host, score, id)
+        | _ ->
+            None
       in
+      let all_hosts = retrieve_wlb_recommendations ~__context ~vm ~snapshot
+        |> List.filter_map possible_rec
+        |> List.sort (fun (_, s, _) (_, s', _) -> Float.compare s' s)
```
edwintorok: Don't you need to reverse the list here? The previous comparison function sorted high values first, this one sorts low values first.

psafont: `s` and `s'` are reversed to have high values first:

```
# List.sort (fun s s' -> Float.compare s' s) [1.; 2.; 3.];;
- : float list = [3.; 2.; 1.]
```

edwintorok: can you add a comment about that? (or name the function something like `cmp_descending`)

psafont:
waiting on https://github.com/xapi-project/xen-api/pull/4035 so Seq can be used for further cleanups

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -550,14 +551,14 @@ let retrieve_wlb_recommendations ~__context ~vm ~snapshot =
...
-         end
-       | Api_errors.Server_error(x, y) -> Hashtbl.replace recs h (x :: y))
-    (retrieve_vm_recommendations ~__context ~vm);
+           Hashtbl.replace recs host (Impossible {source; id; reason="HostCPUFeaturesIncompatible"})
+       | Api_errors.Server_error(x, y) ->
+           let reason = String.concat ";" (x :: y) in
+           Hashtbl.replace recs host (Impossible {source; id; reason})
+      )
+      | host, impossible_load ->
+        Hashtbl.replace recs host impossible_load
```
lippirk: probably worth adding a log line here

psafont: Not sure why, these get ignored.

ocaml/xapi/workload_balancing.ml
```diff
@@ -393,56 +399,61 @@ let val_num i =
...
-        if ((List.length children) != List.length((Helpers.get_live_hosts ~__context)))
-        then
-          raise_malformed_response "VMGetRecommendations"
-            "List of returned reccomendations is not equal to the number of hosts in pool"
-            inner_xml
-        else
-          List.map (fun x -> extract_data x) children
-      | _ -> assert false (*the is_childless should catch this *)
+    match recs with
+    | Xml.Element ( _, _, children) when List.length children <> 0 ->
```
lindig:
```
| Xml.Element ( _, _, _::_ as children)
```

This would avoid the `when`

PR #4031, opened 2020-02-10
CA-333712: use wrapped string type for pool internal API calls

edwintorok:


ocaml/xapi-types/secretString.ml
```diff
@@ -0,0 +1,32 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+type t = string [@@deriving rpc]
+
+let of_string s = s
+
+let _pool_secret = "pool_secret"
```
lindig: Why using a leading underscore?

PR #4032, opened 2020-02-12
CP-32678: update to X509 0.9.0

psafont:
Previously hostname could be empty if there were no hostnames available
in the certificate.
Now the hostname from xapi's database is picked instead.

This PR needs https://github.com/xapi-project/xs-opam/pull/433 as it's a breaking change.

### DIFF
--- a/ocaml/xapi/xapi_vdi.ml
+++ b/ocaml/xapi/xapi_vdi.ml
@@ -1065,7 +1065,7 @@ let _get_nbd_info ~__context ~self ~get_server_certificate =
         let port = 10809L in
         let module Host_set = X509.Certificate.Host_set in
         let select_a_hostname = function
-          | set when set = Host_set.empty ->
+          | set when Host_set.is_empty set ->
               Error (`Msg "Found no subject DNS names in this hosts's certificate.")
           | set ->
               let strict_or_wildcard = function
--- a/ocaml/xapi/xapi_vdi.ml
+++ b/ocaml/xapi/xapi_vdi.ml
@@ -299,7 +299,7 @@ let update_allowed_operations_internal ~__context ~self ~sr_records ~pbd_records
    * re-enable (and maybe alter) the relevant part of
    * test_update_allowed_operations.
   *)
-  let all_ops = 
+  let all_ops =
     Xapi_globs.pre_ely_vdi_operations
     |> List.filter (function
        | `blocked -> false (* CA-260245 *)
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_vdi.ml
```diff
@@ -1059,21 +1059,26 @@ let _get_nbd_info ~__context ~self ~get_server_certificate =
...
+              (* select the first strict hostname, or a wildcard one otherwise *)
+              let strict, wildcard = Host_set.partition strict_or_wildcard set in
+              let hosts = if Host_set.is_empty strict then wildcard else strict in
+              Ok (Host_set.min_elt hosts |> snd |> Domain_name.to_string)
         in
-        let subject = match seek (Certificates.hostnames_of_pem_cert cert) with
+        let hosts = Certificates.hostnames_of_pem_cert cert in
+        let subject = match Rresult.R.bind hosts select_a_hostname with
           | Ok hostname -> hostname
           | Error (`Msg e) ->
```
lippirk: it looks like you can remove ```Msg``

psafont: I cannot

psafont: I cannot, I have to conform to X509's API.

ocaml/xapi/xapi_vdi.ml
```diff
@@ -1059,21 +1059,26 @@ let _get_nbd_info ~__context ~self ~get_server_certificate =
...
-              error "Found no subject DNS names in this hosts's certificate. Returning empty string as subject.";
-              ""
-            )
-            | last :: [] -> last (* Better to return a possible wildcard than nothing *)
-            | name :: xs -> if (String.contains name '*') then _seek xs else name
-          in
-            Rresult.R.bind hostnames (fun v -> Ok (_seek Domain_name.(Set.elements v |> List.map to_string)))
+        let module Host_set = X509.Certificate.Host_set in
+        let select_a_hostname = function
+          | set when set = Host_set.empty ->
```
edwintorok: I think is_empty should be used here, we shouldn't rely on polymorphic equality working on sets.
If you take empty, then add an entry and remove it, you are not guaranteed to get back something exactly equal to empty (set may have some internal fields that are different, we don't know)

psafont: useful read: https://blog.janestreet.com/the-perils-of-polymorphic-compare/

PR #4033, opened 2020-02-14
CP-31856: Propagates nbd params to xenopsd

BenSimsCitrix:
I have a candidate build (private/bensi/CP-31856) and all seems well with my local testing.
I need to run this through a Storage BST and a boot storm after the XenRT issues are resolved.

NOTE! The changes in this PR need to be merged in conjunction with the SM changes or everything will be broken!!

### DIFF
--- a/ocaml/xapi/sm_exec.ml
+++ b/ocaml/xapi/sm_exec.ml
@@ -265,6 +265,7 @@ let parse_attach_result (xml : Xml.xml) =
   rethrow_parse_failures (Xml.to_string_fmt xml) (fun () ->
       let info = XMLRPC.From.structure xml in
       let params = XMLRPC.From.string (safe_assoc "params" info) in
+      let params_nbd = XMLRPC.From.string (safe_assoc "params_nbd" info) in
       let o_direct =
         try XMLRPC.From.boolean (safe_assoc "o_direct" info)
         with _ -> true
@@ -282,6 +283,7 @@ let parse_attach_result (xml : Xml.xml) =
       in
       {
         params;
+        params_nbd;
         o_direct;
         o_direct_reason;
         xenstore_data;
--- a/ocaml/xapi/smint.ml
+++ b/ocaml/xapi/smint.ml
@@ -142,6 +142,7 @@ let query_result_of_sr_driver_info x = {
 
 type attach_info = {
   params : string;
+  params_nbd: string;
   o_direct : bool;
   o_direct_reason : string;
   xenstore_data : (string * string) list;
--- a/ocaml/xapi/storage_access.ml
+++ b/ocaml/xapi/storage_access.ml
@@ -454,9 +454,11 @@ module SMAPIv1 = struct
                      extra = attach_info_v1.Smint.xenstore_data;
                      backend_type = "vbd3"
                    };
-                   (* Currently we always get a BlockDevice from SMAPIv1, never a File, not even for ISOs *)
                    BlockDevice {
                      path = attach_info_v1.Smint.params
+                   };
+                   Nbd {
+                     uri = attach_info_v1.Smint.params_nbd
                    }
                  ]
                }
--- a/ocaml/xapi/sm.ml
+++ b/ocaml/xapi/sm.ml
@@ -144,7 +144,7 @@ let vdi_attach dconf driver sr vdi writable =
   let result = (Sm_exec.exec_xmlrpc (driver_filename driver) call) in
   try Sm_exec.parse_attach_result result
   with _ ->
-    { params = Sm_exec.parse_attach_result_legacy result; o_direct = true; o_direct_reason = ""; xenstore_data = []; }
+    { params = Sm_exec.parse_attach_result_legacy result; params_nbd = ""; o_direct = true; o_direct_reason = ""; xenstore_data = []; }
 
 let vdi_detach dconf driver sr vdi =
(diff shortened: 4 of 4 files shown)


BenSimsCitrix:
Storage BST running 124063

ocaml/xapi/storage_access.ml
```diff
@@ -454,9 +454,11 @@ module SMAPIv1 = struct
                      extra = attach_info_v1.Smint.xenstore_data;
                      backend_type = "vbd3"
                    };
-                   (* Currently we always get a BlockDevice from SMAPIv1, never a File, not even for ISOs *)
```
lippirk: perhaps you could amend this comment rather than deleting it to explain what has changed

ocaml/xapi/sm_exec.ml
```diff
@@ -265,6 +265,7 @@ let parse_attach_result (xml : Xml.xml) =
   rethrow_parse_failures (Xml.to_string_fmt xml) (fun () ->
       let info = XMLRPC.From.structure xml in
       let params = XMLRPC.From.string (safe_assoc "params" info) in
+      let params_nbd = XMLRPC.From.string (safe_assoc "params_nbd" info) in
```
robhoes: It this new param now mandatory? This code raises an error if it is absent.

ocaml/xapi/smint.ml
```diff
@@ -142,6 +142,7 @@ let query_result_of_sr_driver_info x = {
 
 type attach_info = {
   params : string;
+  params_nbd: string;
```
robhoes: It may be better to define this new fields as a `string option` to better handle the cases where this (new) field has not been specified. That would avoid the `params_nbd = ""` above, where we apparently encode the don't-have-this option as an empty string, which is less explicit.

ocaml/xapi/sm_exec.ml
BenSimsCitrix: The param_nbd is  always be passed back from driver/blktap2.py see 
https://github.com/xapi-project/sm/pull/482 

This commit is dangerous if not taken in conjunction with the sm changes.

I don't think there is any way this code would be called in a none SMAP V1 context, but perhaps @edwintorok  can confirm

ocaml/xapi/smint.ml
BenSimsCitrix: Good idea, I will code it up.

ocaml/xapi/sm_exec.ml
robhoes: To make it less dangerous, and the type is changed to an option type, then we can just set the value here to `None` is the key is missing, unless it is important to report this as an error right here.

BenSimsCitrix: Great, That will make integration simpler with SM

ocaml/xapi/storage_access.ml
```diff
@@ -454,9 +454,11 @@ module SMAPIv1 = struct
...
                    BlockDevice {
```
edwintorok: You can just drop BlockDevice from here, and use Nbd.
Then you don't need to change the order of priority in xenopsd.

edwintorok: Although @MarkSymsCtx  just mentioned CDROMs where we do need a blockdevice.

edwintorok: Ideally the SM backend would also specify priorities instead of having XAPI guess.

If everything worked correctly then the most efficient ways to access the device would be:
* Path (O_DIRECT)
* Blockdevice (O_DIRECT)
* NBD

However we currently prefer NBD over Blockdevice because tapdisk's blockdevice support is something we want to remove, and something that uses pagecache when it shouldn't.
Blockdevice is also currently not more efficient than NBD because it apparently wouldn't use O_DIRECT. However that sounds like something to fix in the SM layer.

Thus currently NBD is preferrable to BlockDevices when used for Qemu. However when used for import/export in Dom0 (or any other reason why it might be plugged into Dom0) then NBD would spawn an nbd-client, and we'd get a blockdevice  (which would run an extra process, and allocate extra memory for the buffer, although hopefully a fixed size and not unbounded page-cache).

Seems like XAPI/xenopsd would gain a lot of implementation-specific workarounds (for CD-ROMs, blockdevice memory usage, etc.), where really the implementation would know what is most efficient and tell XAPI about that.

For SMAPIv3 we could change the API to say that the order you return the implementations in is the preferred priority order (except XenDisk would only be used for VMs).
For SMAPIv1 unfortunately the API doesn't specify a list of implementations, there are just various parameters.

Since we're adding a new optional parameter to SMAPIv1, could we instead make it return the SMAPIv3 implementation list then?
i.e. instead of 
```ocaml
params: string
param_nbd: string option
```

Have this:
```
implementations: Storage_interface.implementation list
```

And in parse_attach_result make `params` optional and look for `implementations` in the XML, and if found transform that to an `implementation list`. This would be more work, but would make SMAPIv1 closer to SMAPIv3 and add more flexibility for the backend to make appropriate choices. 
E.g. an ISOSR could always return a blockdevice (or even a file) to work around issues with Qemu and fd passing, and once that is fixed it could revert to returning NBD again.

ocaml/xapi/storage_access.ml
BenSimsCitrix: done

lindig:
This was last touched months ago. Is there a plan or can we close this?

PR #4034, opened 2020-02-17
CA-334909: Only update if the update is needed

rwdrich:
There seem to be three cases for the value of domain type, in the snapshot metadata.
If it's not there, we need to fill it in
If it's there, but is unspecified, it's from a Jura era snapshot, so should be updated from `unspecified
If it's there and filled in, then we shouldn't change the value

### DIFF
--- a/ocaml/xapi/xapi_vm_snapshot.ml
+++ b/ocaml/xapi/xapi_vm_snapshot.ml
@@ -348,13 +348,15 @@ let extended_do_not_copy = [
   "children";
 ] @ do_not_copy
 
+let domain_type_to_string = function
+  | `pv -> "pv"
+  | `hvm -> "hvm"
+
 (* Update the domain_type field if it does not match the HVM_boot_policy *)
-let ensure_domain_type_is_consistent ~__context ~vm =
-  Db.VM.get_HVM_boot_policy ~__context ~self:vm
-  |> fun p -> Xapi_vm_helpers.derive_domain_type ~hVM_boot_policy:p
-  |> fun value ->
-    if Db.VM.get_domain_type ~__context ~self:vm <> value then
-      Db.VM.set_domain_type ~__context ~self:vm ~value
+let ensure_domain_type_is_consistent ~__context ~snap_metadata =
+  match (Stdlib.List.assoc_opt "domain_type" snap_metadata) with
+  | Some "unspecified"  | None -> ("domain_type", domain_type_to_string (Xapi_vm_helpers.derive_domain_type (List.assoc "HVM_boot_policy" snap_metadata))) :: List.remove_assoc "domain_type" snap_metadata
+  | _ -> snap_metadata
 
 (* This function has to be done on the master *)
 let revert_vm_fields ~__context ~snapshot ~vm =
@@ -369,8 +371,8 @@ let revert_vm_fields ~__context ~snapshot ~vm =
     if post_MNR
     then do_not_copy
     else extended_do_not_copy in
+  let snap_metadata = ensure_domain_type_is_consistent ~__context ~snap_metadata in
   copy_vm_fields ~__context ~metadata:snap_metadata ~dst:vm ~do_not_copy ~overrides;
-  ensure_domain_type_is_consistent ~__context ~vm;
   TaskHelper.set_progress ~__context 0.1
 
 let revert ~__context ~snapshot ~vm =
@@ -423,8 +425,8 @@ let	create_vm_from_snapshot ~__context ~snapshot =
              ~__context rpc session_id snap_record in
          begin try
              Db.VM.set_uuid ~__context ~self:new_vm ~value:vm_uuid;
+             let snap_metadata = ensure_domain_type_is_consistent ~__context ~snap_metadata in
              copy_vm_fields ~__context ~metadata:snap_metadata ~dst:new_vm ~do_not_copy:do_not_copy ~overrides;
-             ensure_domain_type_is_consistent ~__context ~vm:new_vm;
              List.iter (fun (snap,_) -> Db.VM.set_snapshot_of ~__context ~self:snap ~value:new_vm) snapshots;
              new_vm
            with e ->


rwdrich:
Additionally, I've moved the domain type check to _before_ the vm fields are copied over. The intention being that we have a snapshot from _some_ era; this function now ensures that that snapshot is consistent in the current era, and it is then used to copy the fields across afterwards

ocaml/xapi/xapi_vm_snapshot.ml
```diff
@@ -348,13 +348,16 @@ let extended_do_not_copy = [
...
 (* Update the domain_type field if it does not match the HVM_boot_policy *)
-let ensure_domain_type_is_consistent ~__context ~vm =
-  Db.VM.get_HVM_boot_policy ~__context ~self:vm
-  |> fun p -> Xapi_vm_helpers.derive_domain_type ~hVM_boot_policy:p
-  |> fun value ->
-    if Db.VM.get_domain_type ~__context ~self:vm <> value then
-      Db.VM.set_domain_type ~__context ~self:vm ~value
+let ensure_domain_type_is_consistent ~__context ~snap_metadata =
+  match (List.assoc_opt "domain_type" snap_metadata) with
+  | Some `unspecified -> ("domain_type", domain_type_to_string (Xapi_vm_helpers.derive_domain_type (List.assoc "HVM_boot_policy" snap_metadata))) :: List.remove_assoc "domain_type" snap_metadata
```
lippirk: it'd be nice to optimize that - i assume `domain_type` often exists in `snap_metadata` and doesn't change too much

lippirk: it's a shame `snap_metadata` isn't a hashtable

ocaml/xapi/xapi_vm_snapshot.ml
```diff
@@ -423,8 +426,8 @@ let	create_vm_from_snapshot ~__context ~snapshot =
              ~__context rpc session_id snap_record in
          begin try
              Db.VM.set_uuid ~__context ~self:new_vm ~value:vm_uuid;
+             let snap_metadata = ensure_domain_type_is_consistent ~__context ~snap_metadata:snap_metadata in
```
lippirk: `~snap_metadata:snap_metadata` can just be `~snap_metadata`

ocaml/xapi/xapi_vm_snapshot.ml
```diff
@@ -369,8 +372,8 @@ let revert_vm_fields ~__context ~snapshot ~vm =
     if post_MNR
     then do_not_copy
     else extended_do_not_copy in
+  let snap_metadata = ensure_domain_type_is_consistent ~__context ~snap_metadata in
```
edwintorok: I'm not sure that moving it improves things. Now you need to refer to fields by string name, that is not typechecked.
Previously you could refer to HVM_boot_policy by using get/set_HVM_boot_policy.

ocaml/xapi/xapi_vm_snapshot.ml
edwintorok: I think the common case will be that domain type is present and specified, the unspecified and None cases are the fallbacks.

edwintorok:
Talked with @rwdrich  that perhaps the simplest fix here is to delete the special handling of domain_type from the snapshot revert code completely, and perhaps replace it by an assert that domain type is not unspecified (if we do find such a case we should find out how we got there and fix that).

When upgrading from <= Jura we already have code in xapi_db_upgrade that sets the domain type for all VMs/snapshots/etc. Upgrading to Stockholm requires a host reboot, thus an implied toolstack restart, so this upgrade should always happen.
When we import a VM I don't think it can have snapshots, and the VM object itself has a correct domain type set if unspecified by Xapi_vm.create.

@robhoes would there be a case here that we missed here?

ocaml/xapi/xapi_vm_snapshot.ml
```diff
@@ -348,13 +348,16 @@ let extended_do_not_copy = [
...
-  Db.VM.get_HVM_boot_policy ~__context ~self:vm
-  |> fun p -> Xapi_vm_helpers.derive_domain_type ~hVM_boot_policy:p
-  |> fun value ->
-    if Db.VM.get_domain_type ~__context ~self:vm <> value then
-      Db.VM.set_domain_type ~__context ~self:vm ~value
+let ensure_domain_type_is_consistent ~__context ~snap_metadata =
+  match (List.assoc_opt "domain_type" snap_metadata) with
+  | Some `unspecified -> ("domain_type", domain_type_to_string (Xapi_vm_helpers.derive_domain_type (List.assoc "HVM_boot_policy" snap_metadata))) :: List.remove_assoc "domain_type" snap_metadata
+  | Some _ -> snap_metadata
+  | None -> ("domain_type", domain_type_to_string (Xapi_vm_helpers.derive_domain_type (List.assoc "HVM_boot_policy" snap_metadata))) :: snap_metadata
```
edwintorok: This line and the unspecified one can be unified, List.remove_assoc is idempotent.

rwdrich:
Additionally, this doesn't fix the original bug. If the domain type is not specified, it is incorrectly calculated

psafont:
Could you add the explanation in the commit message?

ocaml/xapi/xapi_vm_snapshot.ml
robhoes: There are three cases, which Richard's patch implements:
1. The snapshot metadata includes `domain_type` which is not `unspecified`. All is good.
2. The snapshot metadata includes `domain_type`, but it has the value `unspecified`. We need to use `Xapi_vm_helpers.derive_domain_type` to set the `domain_type`.
3. The snapshot metadata does not include `domain_type`. We need to use `Xapi_vm_helpers.derive_domain_type` to set the `domain_type`.

This move was to detect case 3. The old `ensure_domain_type_is_consistent` function didn't have this information, because `copy_vm_fields` only overrides fields that are present in the snapshot metadata, and there is scope for stale metadata to be left behind. However, having to use the untyped data is not great.

Another option is to leave the `ensure_domain_type_is_consistent` call where it was, but add `snap_metadata` as an argument to it. Then it can still do the checks to detect case 2 and 3, while using the normal DB API to do the update. I don't think that we can't get away from checking for the untyped string `"unspecified"` in the snapshot metadata though.

PR #4035, opened 2020-02-17
CP-33121: remove some usages from stdext libraries

psafont:
Remove unneeded open statements

Unfortunately there aren't many gains when replacing the listext functions, the types for the API DB are not optimal.

Processing the lists from the DB API as sets is a common pattern, changing the API to provide Sets of references would clean up the awkward conversions using `Seq`.


ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -1033,10 +1037,13 @@ let list_required_vdis ~__context ~self =
...
 (* Find the SRs of all VDIs which have VBDs attached to the VM. *)
 let list_required_SRs ~__context ~self =
-  let vdis = list_required_vdis ~__context ~self in
-  let srs = List.map (fun vdi -> Db.VDI.get_SR ~__context ~self:vdi) vdis in
-  let srs = List.filter (fun sr -> Db.SR.get_content_type ~__context ~self:sr <> "iso") srs in
-  List.setify srs
+  list_required_vdis ~__context ~self
+  |> List.to_seq
+  |> Seq.map (fun self -> Db.VDI.get_SR ~__context ~self)
+  |> Seq.filter (fun self -> Db.SR.get_content_type ~__context ~self <> "iso")
```
lippirk: changing to seq here (and above) might affect the order in which DB calls are done, but since we aren't doing any writing we should be fine :+1:

ocaml/xapi/xapi_vm.ml
```diff
@@ -693,9 +688,10 @@ let copy ~__context ~vm ~new_name ~sr =
   (* See if the supplied SR is suitable: it must exist and be a non-ISO SR *)
   (* First the existence check. It's not an error to not exist at all. *)
   let sr = try ignore(Db.SR.get_uuid ~__context ~self:sr); Some sr with _ -> None in
-  maybe (fun sr -> debug "Copying disks to SR: %s" (Db.SR.get_uuid ~__context ~self:sr)) sr;
+  Option.fold ~none:() ~some:(fun sr ->
```
mseri: Why not `Option.iter` here and below?

ocaml/xapi/xapi_vm_appliance.ml
```diff
@@ -127,11 +123,15 @@ let assert_can_be_recovered ~__context ~self ~session_to =
...
-  List.setify(List.flatten sr_list)
+  Db.VM_appliance.get_VMs ~__context ~self
+  |> List.to_seq
+  |> Seq.map (fun vm ->
+      Xapi_vm_helpers.get_SRs_required_for_recovery ~__context ~self:vm ~session_to
+      |> List.to_seq)
+  |> Seq.flat_map Fun.id
+  |> Xapi_vm_helpers.SRSet.of_seq
+  |> Xapi_vm_helpers.SRSet.to_seq
+  |> List.of_seq
```
lippirk: it's unfortunate that we use so many intermediate data structures... not sure what the best solution to the list problem is

ocaml/xapi/xapi_vm.ml
psafont: missed that function when reading the module :)

ocaml/xapi/xapi_vm_appliance.ml
psafont: In this case a `Seq.deduplicate` would have worked wonders, I haven't found an ergonomical way to express deduplication.

I've tried to return a seq here instead of a list, I ended up changing quite a lot of code elsewhere.

ocaml/xapi/xapi_vm_helpers.ml
psafont: If there's a race condition with another operation, interleaving the `get_SR` and `get_content_type` calls might change the end result, but the issue would be more pronounced before, as not interleaving the calls separates them further in time.

lippirk: i.e. this is actually better :+1:

ocaml/xapi/xapi_vm_appliance.ml
```diff
@@ -127,11 +123,15 @@ let assert_can_be_recovered ~__context ~self ~session_to =
...
 
 let get_SRs_required_for_recovery ~__context ~self ~session_to =
-  let vms = Db.VM_appliance.get_VMs ~__context ~self in
-  let sr_list = List.map
-      (fun vm -> Xapi_vm_helpers.get_SRs_required_for_recovery ~__context ~self:vm ~session_to)
-      vms in
-  List.setify(List.flatten sr_list)
+  Db.VM_appliance.get_VMs ~__context ~self
+  |> List.to_seq
+  |> Seq.map (fun vm ->
```
edwintorok: Can't you do the flat_map here instead of a line below?

ocaml/xapi/xapi_vm_appliance.ml
```diff
@@ -127,11 +123,15 @@ let assert_can_be_recovered ~__context ~self ~session_to =
...
+      Xapi_vm_helpers.get_SRs_required_for_recovery ~__context ~self:vm ~session_to
+      |> List.to_seq)
+  |> Seq.flat_map Fun.id
+  |> Xapi_vm_helpers.SRSet.of_seq
+  |> Xapi_vm_helpers.SRSet.to_seq
```
edwintorok: You can use `Xapi_vm_helpers.SRSet.elements` to get a list directly, instead of converting to a seq and then to a list

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -94,8 +95,8 @@ let set_is_a_template ~__context ~self ~value =
     List.iter (fun vusb -> try Db.VUSB.destroy ~__context ~self:vusb with _ -> ()) vusbs;
     (* Destroy any attached pvs proxies *)
     Db.VM.get_VIFs ~__context ~self
-    |> List.rev_map (fun v -> Pvs_proxy_control.find_proxy_for_vif ~__context ~vif:v)
-    |> List.unbox_list
+    |> List.filter_map (fun vif -> Pvs_proxy_control.find_proxy_for_vif ~__context ~vif)
+    |> List.rev
```
edwintorok: I think rev_map was used just to avoid stack overflow, not sure whether you need this List.rev here, but to preserve the behaviour of the original code lets keep it if we don't understand whether there are some ordering constraints

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -1033,10 +1037,13 @@ let list_required_vdis ~__context ~self =
...
-  let srs = List.map (fun vdi -> Db.VDI.get_SR ~__context ~self:vdi) vdis in
-  let srs = List.filter (fun sr -> Db.SR.get_content_type ~__context ~self:sr <> "iso") srs in
-  List.setify srs
+  list_required_vdis ~__context ~self
+  |> List.to_seq
+  |> Seq.map (fun self -> Db.VDI.get_SR ~__context ~self)
+  |> Seq.filter (fun self -> Db.SR.get_content_type ~__context ~self <> "iso")
+  |> SRSet.of_seq
+  |> SRSet.to_seq
+  |> List.of_seq
```
edwintorok: elements, see above

ocaml/xapi/xapi_vm_helpers.ml
psafont: yep, couldn't find any evidence whether the order of the destroy operations matter.

PR #4036, opened 2020-02-17
CA-334951: Ignore 0-star recommendations from WLB (backport)

psafont:
This is a backport of https://github.com/xapi-project/xen-api/pull/4030, minus the cleanups done after the actual fixes.

### DIFF
--- a/ocaml/xapi/xapi_vm.ml
+++ b/ocaml/xapi/xapi_vm.ml
@@ -61,8 +61,16 @@ let assert_can_boot_here ~__context ~self ~host =
   assert_can_boot_here ~__context ~self ~host ~snapshot ()
 
 let retrieve_wlb_recommendations ~__context ~vm =
+  let open Workload_balancing in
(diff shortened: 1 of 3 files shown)


ocaml/xapi/xapi_vm.ml
```diff
@@ -61,8 +61,16 @@ let assert_can_boot_here ~__context ~self ~host =
   assert_can_boot_here ~__context ~self ~host ~snapshot ()
 
 let retrieve_wlb_recommendations ~__context ~vm =
+  let open Workload_balancing in
```
lippirk: why did you add this `open`?

psafont: The syntax `Module.(...)` is not supported in OCaml 4.02.3, so I used a local open instead to avoid repetition.

lindig:
@lippirk have your comments been addressed such that we could merge this?

PR #4038, opened 2020-02-21
CP-32678: Add API to to install server certificates

psafont:
Adds an API call to install user-created server certificates.

This does _not_ add parameters to xe.

Draft PR as I have some doubts around severing the connection between hosts when the certificates need to be reloaded. Especially regarding the DB update that needs to be done, and the function that checks whether the recovery of the connection happened successfully.

If something is not clear from the code or the commit messages please let me know.

### DIFF
--- a/ocaml/idl/datamodel_common.ml
+++ b/ocaml/idl/datamodel_common.ml
@@ -8,7 +8,7 @@ open Datamodel_roles
               When introducing a new release, bump the schema minor version to the next hundred
               to leave a gap for potential hotfixes needing to increment the schema version.*)
 let schema_major_vsn = 5
-let schema_minor_vsn = 504
+let schema_minor_vsn = 600
 
 (* Historical schema versions just in case this is useful later *)
 let rio_schema_major_vsn = 5
--- a/ocaml/idl/datamodel_common.ml
+++ b/ocaml/idl/datamodel_common.ml
@@ -200,6 +200,12 @@ let get_product_releases in_product_since =
     | x::xs -> go_through_release_order xs
   in go_through_release_order release_order
 
+let stockholm_release =
+  { internal = get_product_releases rel_stockholm
+  ; opensource = get_oss_releases None
+  ; internal_deprecated_since = None
+  }
+
 let quebec_release =
   { internal = get_product_releases rel_quebec
   ; opensource = get_oss_releases None
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -878,7 +878,7 @@ let host_query_ha = call ~flags:[`Session]
       ~pool_internal:true
       ~hide_from_docs:true
       ~name:"certificate_install"
-      ~doc:"Install an SSL certificate to this host."
+      ~doc:"Install a TLS CA certificate to this host."
       ~params:[Ref _host, "host", "The host";
                String, "name", "A name to give the certificate";
                String, "cert", "The certificate"]
@@ -891,7 +891,7 @@ let host_query_ha = call ~flags:[`Session]
       ~pool_internal:true
       ~hide_from_docs:true
       ~name:"certificate_uninstall"
-      ~doc:"Remove an SSL certificate from this host."
+      ~doc:"Remove a TLS CA certificate from this host."
       ~params:[Ref _host, "host", "The host";
                String, "name", "The certificate name"]
       ~allowed_roles:_R_LOCAL_ROOT_ONLY
@@ -903,7 +903,7 @@ let host_query_ha = call ~flags:[`Session]
       ~pool_internal:true
       ~hide_from_docs:true
       ~name:"certificate_list"
-      ~doc:"List all installed SSL certificates."
+      ~doc:"List the filenames of all installed TLS CA certificates."
       ~params:[Ref _host, "host", "The host"]
       ~result:(Set(String),"All installed certificates")
       ~allowed_roles:_R_LOCAL_ROOT_ONLY
@@ -915,7 +915,7 @@ let host_query_ha = call ~flags:[`Session]
       ~pool_internal:true
       ~hide_from_docs:true
       ~name:"crl_install"
-      ~doc:"Install an SSL certificate revocation list to this host."
+      ~doc:"Install a TLS Certificate Revocation List to this host."
       ~params:[Ref _host, "host", "The host";
                String, "name", "A name to give the CRL";
                String, "crl", "The CRL"]
@@ -928,7 +928,7 @@ let host_query_ha = call ~flags:[`Session]
       ~pool_internal:true
       ~hide_from_docs:true
       ~name:"crl_uninstall"
-      ~doc:"Remove an SSL certificate revocation list from this host."
+      ~doc:"Uninstall a TLS certificate revocation list from this host."
       ~params:[Ref _host, "host", "The host";
                String, "name", "The CRL name"]
       ~allowed_roles:_R_LOCAL_ROOT_ONLY
@@ -940,7 +940,7 @@ let host_query_ha = call ~flags:[`Session]
       ~pool_internal:true
       ~hide_from_docs:true
       ~name:"crl_list"
-      ~doc:"List all installed SSL certificate revocation lists."
+      ~doc:"List the filenames of all installed TLS Certificate Revocation Lists."
       ~params:[Ref _host, "host", "The host"]
       ~result:(Set(String),"All installed CRLs")
       ~allowed_roles:_R_LOCAL_ROOT_ONLY
@@ -952,7 +952,7 @@ let host_query_ha = call ~flags:[`Session]
       ~pool_internal:true
       ~hide_from_docs:true
       ~name:"certificate_sync"
-      ~doc:"Resync installed SSL certificates and CRLs."
+      ~doc:"Makes installed TLS CA certificates and CRLs available to all programs using OpenSSL."
       ~params:[Ref _host, "host", "The host"]
       ~allowed_roles:_R_LOCAL_ROOT_ONLY
       ()
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -878,7 +878,7 @@ let host_query_ha = call ~flags:[`Session]
       ~pool_internal:true
       ~hide_from_docs:true
       ~name:"certificate_install"
-      ~doc:"Install a TLS CA certificate to this host."
+      ~doc:"Install a TLS CA certificate on this host."
       ~params:[Ref _host, "host", "The host";
                String, "name", "A name to give the certificate";
                String, "cert", "The certificate"]
@@ -952,9 +952,31 @@ let host_query_ha = call ~flags:[`Session]
       ~pool_internal:true
       ~hide_from_docs:true
       ~name:"certificate_sync"
-      ~doc:"Makes installed TLS CA certificates and CRLs available to all programs using OpenSSL."
+      ~doc:"Make installed TLS CA certificates and CRLs available to all programs using OpenSSL."
       ~params:[Ref _host, "host", "The host"]
       ~allowed_roles:_R_LOCAL_ROOT_ONLY
+      ()
+
+  let install_server_certificate = call
+      ~in_oss_since:None
+      ~lifecycle:[Published, rel_stockholm, ""]
+      ~name:"install_server_certificate"
+      ~doc:"Install the TLS server certificate."
+      ~versioned_params:
+        [{ param_type=Ref _host; param_name="host"; param_doc="The host"
+         ; param_release=stockholm_release; param_default=None}
+        ;{ param_type=String; param_name="certificate"
+         ; param_doc="The server certificate, in PEM form"
+         ; param_release=stockholm_release; param_default=None}
+        ;{ param_type=String; param_name="private_key"
+         ; param_doc="The unencrypted private key used to sign the certificate, \
+                      in PKCS#8 form"
+         ; param_release=stockholm_release; param_default=None}
+        ;{ param_type=String; param_name="certificate_chain"
+         ; param_doc="The certificate chain, in PEM form"
+         ; param_release=stockholm_release; param_default=Some (VString "")}
+        ]
+      ~allowed_roles:_R_POOL_ADMIN
       ()
 
   let get_server_certificate = call
@@ -1379,6 +1401,7 @@ let host_query_ha = call ~flags:[`Session]
         crl_list;
         certificate_sync;
         get_server_certificate;
+        install_server_certificate;
         update_pool_secret;
         update_master;
         attach_static_vdis;
--- a/ocaml/idl/datamodel_pool.ml
+++ b/ocaml/idl/datamodel_pool.ml
@@ -407,7 +407,7 @@ open Datamodel_types
       ~in_oss_since:None
       ~in_product_since:rel_george
       ~name:"certificate_install"
-      ~doc:"Install an SSL certificate pool-wide."
+      ~doc:"Install a TLS CA certificate, pool-wide."
       ~params:[String, "name", "A name to give the certificate";
                String, "cert", "The certificate"]
       ~allowed_roles:_R_POOL_OP
@@ -417,7 +417,7 @@ open Datamodel_types
       ~in_oss_since:None
       ~in_product_since:rel_george
       ~name:"certificate_uninstall"
-      ~doc:"Remove an SSL certificate."
+      ~doc:"Remove a pool-wide TLS CA certificate."
       ~params:[String, "name", "The certificate name"]
       ~allowed_roles:_R_POOL_OP
       ()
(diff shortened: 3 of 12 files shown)


ocaml/xapi/message_forwarding.ml
```diff
@@ -2607,6 +2607,32 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
+    let install_server_certificate ~__context ~host ~certificate ~private_key ~certificate_chain =
+      let local_fn =
+        Local.Host.install_server_certificate
+          ~host
+          ~certificate
+          ~private_key
+          ~certificate_chain
+      in
+      let success () =
+        (* check that fingerprint of the installed cert matches? *)
```
lippirk: it is also worth checking we can contact the host we just installed a cert on?

lippirk: maybe that belongs to the code immediately after we `reconfigure_stunnel`

ocaml/xapi/xapi_host.ml
```diff
@@ -1126,10 +1126,33 @@ let crl_list ~__context ~host =
...
+let install_server_certificate ~__context ~host ~certificate ~private_key ~certificate_chain =
+  let pem_chain = match certificate_chain with
+  | ""        -> None
+  | pem_chain -> Some pem_chain
+  in
+  let certificate =
+    Certificates.install_server_certificate ~pem_chain certificate private_key
+  in
+  let date_of_ptime time = Date.of_float (Ptime.to_float_s time) in
+  let map f (a, b) = (f a, f b) in
```
lippirk: personally i think `let date_of_ptime (a, b) = (date_of_ptime a, date_of_ptime b)` is better

ocaml/xapi/xapi_host.ml
```diff
@@ -1126,10 +1126,33 @@ let crl_list ~__context ~host =
...
+    map date_of_ptime (X509.Certificate.validity certificate)
+  in
+  let fingerprint =
+    X509.Certificate.fingerprint Nocrypto.(`SHA256) certificate
+    |> Cstruct.to_string
+  in
+  let uuid = Uuid.(to_string (make_uuid ())) in
+  let ref = Ref.make () in
+  Db.Certificate.create ~__context ~ref ~uuid ~host ~not_before ~not_after ~fingerprint;
+  (* wait for db sync? next call severs all connections done with stunnel *)
```
lippirk: probably want some logging here :)

ocaml/xapi/message_forwarding.ml
```diff
@@ -2607,6 +2607,32 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
+        Some ()
+      in
+
+      let fn () = do_op_on ~local_fn ~__context ~host
+        (fun session_id rpc ->
+          Client.Host.install_server_certificate
```
lippirk: you might want to make this an async call instead

this way the host which is restarting stunnel can let you know whether or not it successfully installed the certificate _before_ it kills all of its stunnel connections

ocaml/xapi/message_forwarding.ml
psafont: I'll make an async call, so this won't be needed

ocaml/xapi/message_forwarding.ml
psafont: I think I'll wait until your PR is in to add async

ocaml/idl/datamodel_host.ml
```diff
@@ -878,7 +878,7 @@ let host_query_ha = call ~flags:[`Session]
       ~pool_internal:true
       ~hide_from_docs:true
       ~name:"certificate_install"
-      ~doc:"Install an SSL certificate to this host."
+      ~doc:"Install a TLS CA certificate to this host."
```
lindig: While we are here; "install .. to" seems to me unusual. A native speaker should comment on this but "install on", "install for" seem to me conventional.

ocaml/idl/datamodel_host.ml
```diff
@@ -915,7 +915,7 @@ let host_query_ha = call ~flags:[`Session]
       ~pool_internal:true
       ~hide_from_docs:true
       ~name:"crl_install"
-      ~doc:"Install an SSL certificate revocation list to this host."
+      ~doc:"Install an TLS Certificate Revocation List to this host."
```
lindig: Same as above.

ocaml/idl/datamodel_host.ml
```diff
@@ -952,11 +952,33 @@ let host_query_ha = call ~flags:[`Session]
       ~pool_internal:true
       ~hide_from_docs:true
       ~name:"certificate_sync"
-      ~doc:"Resync installed SSL certificates and CRLs."
+      ~doc:"Makes installed TLS CA certificates and CRLs available to all programs using OpenSSL."
```
lindig: I think this should be "Make .." to stay with the imperative form used above.

ocaml/xapi/certificates.ml
```diff
@@ -387,7 +390,8 @@ let validate_certificate kind pem now private_key =
           `Msg (server_certificate_chain_invalid, []))
     )
   |> function
-    | Ok _ -> ()
+    | Ok (cert :: _)  -> cert
```
lindig: Do you want to log that there were several entries but only the first is used?
```
| Ok [cert] ->
| Ok (cert::_) -> log ..
```

psafont: This returns the leaf certificate if the chain. In the case of the leaf certificate we want to return it to add its metadata to the database, in the case of the chain, we don't care right now. Returning a certificate in the case of a chain is only done to keep the type checker happy.

In the future the function will change as we will most probably have to verify the whole chain every time, with the information of the leaf certificate added to the database, with a possibility to show the certs in the chain as well.

psafont:
Regarding the lifecycle of the installed certificates and the database objects: right now the ones installed with the bash script are not being tracked, the intention is to change them.

The operation with a single-host pool work, tested with using the console in the lab, manually. Installing a certificate on a slave is not tested.

ocaml/idl/datamodel_host.ml
```diff
@@ -952,11 +952,33 @@ let host_query_ha = call ~flags:[`Session]
...
+      ~name:"install_server_certificate"
+      ~doc:"Install the TLS server certificate."
+      ~versioned_params:
+        [{ param_type=Ref _host; param_name="host"; param_doc="The host"
+         ; param_release=stockholm_release; param_default=None}
+        ;{ param_type=String; param_name="certificate"
+         ; param_doc="The server certificate, in PEM form"
+         ; param_release=stockholm_release; param_default=None}
+        ;{ param_type=String; param_name="private_key"
+         ; param_doc="The unecrypted private key used to sign the certificate, \
```
lippirk: is the multiline string really necessary here?

psafont: It already has a width of 80 chars :)

ocaml/tests/test_certificates.ml
```diff
@@ -216,6 +223,23 @@ let invalid_chain_cert_tests =
...
+
+let pp_hash_test =
+  List.map (fun (hashable, expected) ->
+    let test_hash hashable repr () =
+      let digest = Cstruct.of_string hashable
+      |> Nocrypto.Hash.digest `SHA256
+      in
+      Alcotest.(check string) "fingerprints must match" expected (Certificates.pp_hash digest)
+    in
+    Printf.sprintf {|Validation of hash printing of "%s"|} hashable,
```
lippirk: you haven't used `repr`

this suggests you have been silencing some warnings locally, which you ought to fix

ocaml/idl/datamodel_host.ml
robhoes: `unecrypted` -> `unencrypted` ;)

ocaml/tests/test_certificates.ml
psafont: warnings are always silenced when using `make test`, across the whole toolstack.

I'll fix this now.

ocaml/xapi/message_forwarding.ml
```diff
@@ -2605,6 +2605,33 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
+      let task = Context.get_task_id __context in
+      let local_fn =
+        Local.Host.install_server_certificate
+          ~host
+          ~certificate
+          ~private_key
+          ~certificate_chain
+      in
+      let success () =
+        let progress = Db.Task.get_progress ~__context ~self:task in
```
lippirk: i believe ``let success () = Task.get_status ~__context ~self:task = `success`` would be preferred

(for example migrate_send's progress gets set to 1 and then back to zero again a couple of times... :/)

ocaml/xapi/xapi_host.ml
```diff
@@ -1124,10 +1124,52 @@ let crl_list ~__context ~host =
...
+  let current_server_certs = Db.Certificate.get_all_records ~__context
+  |> List.filter_map (fun (reference, record) ->
+    if record.API.certificate_host = host then
+        Some reference
+    else
+        None
+    )
+  in
+
+  let pem_chain = match certificate_chain with
```
lippirk: worth trimming here? :man_shrugging:

ocaml/xapi/xapi_host.ml
```diff
@@ -1124,10 +1124,52 @@ let crl_list ~__context ~host =
...
 
 let get_server_certificate ~__context ~host =
-  Certificates.get_server_certificate()
+  Certificates.get_server_certificate ()
+
+let install_server_certificate ~__context ~host ~certificate ~private_key ~certificate_chain =
+  if Db.Pool.get_ha_enabled ~__context ~self:(Helpers.get_pool ~__context) then
+    raise Api_errors.(Server_error (ha_is_enabled, []));
+
+  let current_server_certs = Db.Certificate.get_all_records ~__context
```
lippirk: what if there are no certs?

psafont: Right now we don't care as this is not used for installing the self-signed cert from the first boot. When adding the emergency call we'll see what needs to happen here.

ocaml/xapi/xapi_host.ml
psafont: pem_chain is an "optional" parameter, `no chain` is represented by the empty string

ocaml/xapi/message_forwarding.ml
lippirk: Actually, I'm wrong here - setting the task status to ```success`` would make the client think that the task had completed, when it in fact hasn't

psafont: I used this scheme because I saw it was being used elsewhere. The way it works is: I set the progress 1 one just before restarting stunnel. As soon as it reconnect checks that the task completion is set to 1 and end successfully. If for any reason the task fails beforehand and the connection is reset this check will fail.

I used this to follow convention, I'm more than happy to change it to something else if it's better.

lippirk: if it works, and since the same technique is used elsewhere, i would leave it as it is


the only status states are:

```
  | `pending 
  | `success 
  | `failure 
  | `cancelling
  | `cancelled
```

as stated above we can't use `` `success ``,  so we can't use the task status, and there are no other valid candidates

ocaml/idl/datamodel_common.ml
```diff
@@ -8,7 +8,7 @@ open Datamodel_roles
               When introducing a new release, bump the schema minor version to the next hundred
               to leave a gap for potential hotfixes needing to increment the schema version.*)
 let schema_major_vsn = 5
-let schema_minor_vsn = 504
+let schema_minor_vsn = 505
```
edwintorok: Can you bump this to next hundred like the comment says?

ocaml/idl/datamodel_host.ml
```diff
@@ -952,11 +952,33 @@ let host_query_ha = call ~flags:[`Session]
...
+        [{ param_type=Ref _host; param_name="host"; param_doc="The host"
+         ; param_release=stockholm_release; param_default=None}
+        ;{ param_type=String; param_name="certificate"
+         ; param_doc="The server certificate, in PEM form"
+         ; param_release=stockholm_release; param_default=None}
+        ;{ param_type=String; param_name="private_key"
+         ; param_doc="The unencrypted private key used to sign the certificate, \
+                      in PKCS#8 form"
+         ; param_release=stockholm_release; param_default=None}
+        ;{ param_type=String; param_name="certificate_chain"
```
edwintorok: Chain to...? Any root CA installed system-wide, or the above installed TLS CA certificates with the other API, or both?

ocaml/tests/test_certificates.ml
```diff
@@ -181,12 +185,15 @@ let invalid_leaf_cert_tests =
   List.map test_invalid_leaf_cert invalid_leaf_certificates
 
 let test_valid_cert_chain chain time pkey () =
-  validate_certificate Chain chain time pkey
+  let _leaf = validate_certificate Chain chain time pkey in ()
```
edwintorok: sometimes you use an explicit type below, I think that is better.
Otherwise you could've just used ignore, but using explicit type is preferable to make sure you actually call the function, and not just curry it.

ocaml/idl/datamodel_common.ml
psafont: I didn't bump to the next hundred because I saw that the latest release further down was 4xx, now I see that quebec is not on the list of releases here, it should be added as well.

ocaml/idl/datamodel_host.ml
psafont: None of those, it's to the first self-signed cert / root CA cert that the other party  in the TLS connection encounters in the verification chain.

ocaml/idl/datamodel_host.ml
```diff
@@ -915,7 +915,7 @@ let host_query_ha = call ~flags:[`Session]
       ~pool_internal:true
       ~hide_from_docs:true
       ~name:"crl_install"
-      ~doc:"Install an SSL certificate revocation list to this host."
+      ~doc:"Install an TLS Certificate Revocation List on this host."
```
lindig: Typo: "a" TLS

ocaml/idl/datamodel_pool.ml
```diff
@@ -435,7 +435,7 @@ open Datamodel_types
       ~in_oss_since:None
       ~in_product_since:rel_george
       ~name:"crl_install"
-      ~doc:"Install an SSL certificate revocation list, pool-wide."
+      ~doc:"Install an TLS Certificate Revocation List, pool-wide."
```
lindig: Typo: "a" TLS

ocaml/idl/datamodel_pool.ml
```diff
@@ -454,16 +454,16 @@ open Datamodel_types
...
+      ~result:(Set(String), "The names of all installed CRLs")
       ~allowed_roles:_R_POOL_OP
       ()
 
   let certificate_sync = call
       ~in_oss_since:None
       ~in_product_since:rel_george
       ~name:"certificate_sync"
-      ~doc:"Sync SSL certificates from master to slaves."
+      ~doc:"Copy the TLS CA certificates and CRLs of the master to the slaves."
```
lindig: Maybe use: "to all slaves"

ocaml/tests/test_certificates.ml
```diff
@@ -216,6 +223,23 @@ let invalid_chain_cert_tests =
     "Validation of an unsupported certificate chain", `Quick, test_cert)
   corrupt_chain_certificates
 
+let hashables =
+  [ "foobar",
```
lindig: Can we find a better name?

psafont: any string will do, I could change it to "hashable", but that won't add any information.

lindig: `__LOC__`, maybe?

psafont: I think that would make changing this file unnecessarily difficult

ocaml/xapi/xapi_host.ml
```diff
@@ -1129,6 +1129,48 @@ let certificate_sync ~__context ~host =
...
+  |> List.filter_map (fun (reference, record) ->
+    if record.API.certificate_host = host then
+        Some reference
+    else
+        None
+    )
+  in
+
+  let pem_chain = match certificate_chain with
+  | ""        -> None
```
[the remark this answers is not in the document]
psafont: Setting the default value to the empty string expression doesn't seem to allow to make the argument of the function optional (using `?certificate_chain=""`)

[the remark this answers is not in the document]
robhoes: No, this function always gets all parameters. If a client omits an optional parameter of an API function, then the default is substituted before you get here.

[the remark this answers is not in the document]
psafont: Hm I tried to omit the parameter in `cli_operations.ml` when there is no chain, but the type system would not allow me as the parameter is named, not optional.

This is when using `Client.Host.install_server_certificate`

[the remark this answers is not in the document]
robhoes: Yeah, unfortunately our OCaml client module doesn't support optional parameters.

robhoes:
👍 now squash and merge :)

PR #4039, opened 2020-02-21
Move init.d-xapissl script to ocaml land

lippirk:
`init.d-xapissl` write the stunnel config and manages the stunnel service. This can be simplified now that an stunnel service file has been written - at the same time we rewrite the script in ocaml for coherence.

Preliminary testing is positive (though I have only tested it against master so far, not also with the other changes in REQ-811). I am going to run some BST+BVTs.


ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,104 @@ let host_supports_hvm ~__context host =
...
+end = struct
+
+  let cert = !Xapi_globs.server_cert_path
+
+  (** protect ourselves from concurrent writes to files *)
+  module File : sig
+    (** create or update stunnel config file *)
+    val update_config : accept:string -> unit
+    (** calls `generate_ssl_cert` bash script
+      * TODO: don't generate certificates in bash? *)
```
psafont: This is planned using the work done in branch `feature/REQ-453/master`

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,104 @@ let host_supports_hvm ~__context host =
...
+        ; Printf.sprintf "cert = %s" cert
+        ; Printf.sprintf "ciphers = %s" Xcp_const.good_ciphersuites
+        ; "curve = secp384r1"
+        ; "TIMEOUTclose = 1"
+        ; "options = CIPHER_SERVER_PREFERENCE"
+        ; "sslVersion = TLSv1.2"
+        ]
+      in
+      if Sys.file_exists xapi_stunnel_conf then Sys.remove xapi_stunnel_conf;
+      Stdext.Unixext.write_string_to_file xapi_stunnel_conf conf_contents
```
edwintorok: There is a function that does an atomic replace, we could use that, then you don't need to remove the old one, the atomic replace will take care of it.

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,104 @@ let host_supports_hvm ~__context host =
...
+        | None | Some _ -> begin
+          current_accept := Some accept;
+          rewrite_xapi_ssl_config_file ~accept
+        end
+      )
+
+    let gen_pem_if_not_exist () =
+      Mutex.execute m (fun () ->
+        if not (Sys.file_exists cert) then begin
+          let hostname = call_script "/usr/bin/hostname" ["-f"] |> Astring.String.trim in
```
edwintorok: You can use `Unix.gethostname ()`

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,104 @@ let host_supports_hvm ~__context host =
...
+          let cn =
+            if Astring.String.is_prefix ~affix:"localhost" hostname then
+              "mgmt_ip"
```
edwintorok: I think you should call `Helpers.get_management_ip_addr` here.

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,104 @@ let host_supports_hvm ~__context host =
...
+            else if Astring.String.is_infix ~affix:"." hostname then
+              hostname
+            else
+              "mgmt_ip"
```
edwintorok: as above, this has to be a function call

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,104 @@ let host_supports_hvm ~__context host =
...
+    function
+    | "enabled"  -> true
+    | "disabled" -> false
+    | _          -> D.error "Stunnel.is_enabled: unable to discern if stunnel@xapi is enabled. assuming that it is disabled"; false
+
+  let restart ~accept =
+    File.update_config ~accept;
+    File.gen_pem_if_not_exist ();
+    if not @@ is_enabled () then systemctl_ "enable";
+    systemctl_ "restart"
```
edwintorok: I think we might have to modify the default stunnel service file via systemd drop-ins, the old startup script also had this:
```
    # Increase the fd limit to 2048, so stunnel could accept 1000 client connections
    ulimit -n 2048
```

That one now has to be set using systemd directives.

scripts/Makefile
```diff
@@ -46,7 +46,6 @@ install:
 	$(IPROG) db.conf.skel $(DESTDIR)$(ETCDIR)/db.conf
 	$(IPROG) rio.db.conf.skel $(DESTDIR)$(ETCDIR)/db.conf.rio
 	$(IPROG) init.d-xapi $(DESTDIR)$(LIBEXECDIR)/xapi-init
-	$(IPROG) init.d-xapissl $(DESTDIR)$(LIBEXECDIR)/xapissl
```
edwintorok: You'll need to update the specfile too

ocaml/xapi/xapi_mgmt_iface.mli
```diff
@@ -42,4 +42,4 @@ val rebind : __context:Context.t -> unit
 val enable_himn : __context:Context.t -> addr:string -> unit
 
 (** Restart stunnel *)
-val reconfigure_stunnel : __context:Context.t -> unit
+val reconfigure_stunnel : unit -> unit
```
edwintorok: I think a context would be useful, you'll need to get the management IP, etc.

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -38,33 +38,14 @@ let update_mh_info interface =
...
-                     ; Xcp_const.good_ciphersuites
-                     ]
-  in
-  let (_ : Thread.t) = Thread.create (fun () ->
-      Mutex.execute management_m (fun () ->
-          Forkhelpers.execute_command_get_output !Xapi_globs.xapissl_path xapissl_args
-        )
-    ) () in
-  ()
+  Helpers.Stunnel.restart ~accept
```
edwintorok: This previously used to be a thread, and now is inline, it might be better to keep it in a thread

edwintorok:
This is a good direction, and should make stunnel more robust (systemd will now also restart it if it fails)

ocaml/xapi/helpers.ml
lippirk: you are right, i misread this part of the bash script (i didn't realise ``f`` is a function call in bash)

```
        case "${CN}" in
            localhost*)
                CN=`mgmt_ip`;;
            *.*)
                :;;
            *)
                CN=`mgmt_ip`;;
        esac
        "@LIBEXECDIR@/generate_ssl_cert" ${PEMFILE} ${CN}
```

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,108 @@ let host_supports_hvm ~__context host =
...
+      Mutex.execute m (fun () ->
+        match !current_accept with
+        | Some current_accept when current_accept = accept -> ()
+        | None | Some _ -> begin
+          current_accept := Some accept;
+          rewrite_xapi_ssl_config_file ~accept
+        end
+      )
+
+    let gen_pem_if_not_exist ~__context =
```
psafont: I'd be interested in triggering this function, when there's a certificate present but it's expired.

One option to allow for this would be to provide the file check as a parameter to move it outside if the function (restart would provide the check)

I'm open to other ideas as this isn't super clean, but the handling of the mutex makes it tricky.

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,108 @@ let host_supports_hvm ~__context host =
...
+          let hostname = Unix.gethostname () in
+          let cn =
+            if Astring.String.is_prefix ~affix:"localhost" hostname then
+              get_management_ip_addr ~__context
+            else if Astring.String.is_infix ~affix:"." hostname then
+              None
+            else
+              get_management_ip_addr ~__context
+          in
+          call_script "/opt/xensource/libexec/generate_ssl_cert" [cert; Opt.default hostname cn] |> ignore
```
psafont: Prefer using Stdlib's `Option.value ~default:hostname cn`.

ocaml/xapi/helpers.ml
lippirk: it would probably make sense if you became the owner of this function, and this stunnel helper becomes a client (or we figure out a way to decouple restarting stunnel and writing the stunnel config completely)

that way we you could handle concurrent writes yourself without the confusion of using this mutex

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,108 @@ let host_supports_hvm ~__context host =
...
+    Astring.String.trim |>
+    function
+    | "enabled"  -> true
+    | "disabled" -> false
+    | _          -> D.error "Stunnel.is_enabled: unable to discern if stunnel@xapi is enabled. assuming that it is disabled"; false
+
+  let restart ~__context ~accept =
+    try
+      File.update_config ~accept;
+      File.gen_pem_if_not_exist ~__context;
```
psafont: Previously this was done on every start of the daemon (on `start ()`)
While I don't like the idea of having logic for creating certs run every time stunnel is started, I wonder if the replacement is enough: stunnel could be started up without xapi being involved, missing the creation of the certificate.

I don't think it's happening right now, but with all the changes on the service management and firstboot some breakage could slip through.

Could you check this suspicion so we're confident this change doesn't add brittleness?

edwintorok: You could add another systemd service file, and make stunnel@xapi depend on it (it should be possible to do this from your gencert.service by adding a WantedBy or something similar).
Then systemd would guarantee us that it is available.

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,108 @@ let host_supports_hvm ~__context host =
...
+    the stunnel daemon but they do this directly (not via ocaml). *)
+module Stunnel : sig
+  (** restart stunnel, possibly changing the config file *)
+  val restart : __context:Context.t -> accept:string -> unit
+end = struct
+
+  let cert = !Xapi_globs.server_cert_path
+
+  (** protect ourselves from concurrent writes to files *)
+  module File : sig
```
lindig: * A better name for the module might be `Config`. 
* Use `update` for `update_config`
* Use `generate_pem`. It is already idempotent - I would not encode this in the name. But you could add a `?(force=true)` to enforce an update.

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,108 @@ let host_supports_hvm ~__context host =
...
+      let conf_contents = String.concat "\n" [
+          "; autogenerated by helpers.ml"
+        ; "fips = no"
+        ; "pid = /var/run/xapissl.pid"
+        ; "socket = r:TCP_NODELAY=1"
+        ; "socket = a:TCP_NODELAY=1"
+        ; "socket = l:TCP_NODELAY=1"
+        ; "socket = r:SO_KEEPALIVE=1"
+        ; "socket = a:SO_KEEPALIVE=1"
+        ; idle_timeout
```
lindig: The `idle_timeout` could be inlined here - might be a little more obvious.

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,108 @@ let host_supports_hvm ~__context host =
...
+      )
+
+    let gen_pem_if_not_exist ~__context =
+      Mutex.execute m (fun () ->
+        if not (Sys.file_exists cert) then begin
+          let hostname = Unix.gethostname () in
+          let cn =
+            if Astring.String.is_prefix ~affix:"localhost" hostname then
+              get_management_ip_addr ~__context
+            else if Astring.String.is_infix ~affix:"." hostname then
```
lindig: This logic does not match a non-qualified (local) hostname like `example`. I am a bit surprised about the logic here where we return None for some names but not others.

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,108 @@ let host_supports_hvm ~__context host =
...
+              get_management_ip_addr ~__context
+          in
+          call_script "/opt/xensource/libexec/generate_ssl_cert" [cert; Opt.default hostname cn] |> ignore
+        end
+      )
+
+  end
+
+  let systemctl cmd = call_script "/usr/bin/systemctl" [cmd; "stunnel@xapi"]
+  let systemctl_ cmd = systemctl cmd |> ignore
```
lindig: Do we need both? Another option:
```
ignore @@ call_script "/usr/bin/systemctl" [cmd; "stunnel@xapi"]
```

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,108 @@ let host_supports_hvm ~__context host =
...
+
+  let is_enabled () =
+    systemctl "is-enabled" |>
+    Astring.String.trim |>
+    function
+    | "enabled"  -> true
+    | "disabled" -> false
+    | _          -> D.error "Stunnel.is_enabled: unable to discern if stunnel@xapi is enabled. assuming that it is disabled"; false
```
lindig: Show the string that was matched in the error message.

ocaml/xapi/helpers.ml
edwintorok: We also have some systemctl in xapi_clustering.ml. If this is more generally useful now feel free to move it (as long as it doesn't introduce a dependency cycle).

lippirk: i'm using `systemctl_ "restart"` and `systemctl "is-enabled"`

i'll look into merging these modules

ocaml/xapi/helpers.ml
lippirk: i've simply lifted this logic directly from the bash script:

```
        case "${CN}" in
            localhost*)
                CN=`mgmt_ip`;;
            *.*)
                :;;
            *)
                CN=`mgmt_ip`;;
        esac
```

lippirk:
bst+bvt all green

if you both want the systemd gencert service, then i will do that (but as i said i think it is overkill)

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,118 @@ let host_supports_hvm ~__context host =
...
+        (fun fd -> let (_: int) = Unix.single_write_substring fd conf_contents 0 len in ())
+
+    let update ~accept =
+      Mutex.execute m (fun () ->
+        match !current_accept with
+        | Some current_accept when current_accept = accept -> ()
+        | None | Some _ -> begin
+          current_accept := Some accept;
+          rewrite_xapi_ssl_config_file ~accept
+        end
```
psafont: The begin and end are not needed here.

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,118 @@ let host_supports_hvm ~__context host =
...
+  module Config : sig
+    (** create or update stunnel config file *)
+    val update : accept:string -> unit
+    (** calls `generate_ssl_cert` bash script
+      * TODO: don't generate certificates in bash *)
+    val generate_pem : __context:Context.t -> force:bool -> unit
+  end = struct
+    let xapi_stunnel_conf = "/etc/stunnel/xapi.conf"
+    let m = Mutex.create ()
+    let current_accept = ref None
```
psafont: Isn't this variable replicating `stunnel_accept` in ` ocaml/xapi/xapi_mgmt_iface.ml`? I don't see `stunnel_accept` being modified anymore whenever it's being restarted.

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,118 @@ let host_supports_hvm ~__context host =
...
+        | Some current_accept when current_accept = accept -> ()
+        | None | Some _ -> begin
+          current_accept := Some accept;
+          rewrite_xapi_ssl_config_file ~accept
+        end
+      )
+
+    let generate_pem ~__context ~force =
+      Mutex.execute m (fun () ->
+        if not (Sys.file_exists cert) then begin
```
lindig: The `force` would take effect here: `if force || not @@ Sys.file_exists cert then`

ocaml/xapi/helpers.ml
```diff
@@ -1260,7 +1260,9 @@ end = struct
     let rewrite_xapi_ssl_config_file ~accept =
       let conf_contents = String.concat "\n" [
           "; autogenerated by helpers.ml"
-        ; "fips = no"
+        ; (match Sys.getenv_opt "CC_PREPARATIONS" with
+          | Some "true"   -> "fips = yes"
```
edwintorok: You could say `"fips = " ^ (if fips then "yes" else "no")`, and set `fips` based on the env var as you did, just in case we need to do anything else to turn fips on /off

lippirk:
As discussed, I will create a gencert systemd service (but in a new PR)

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,122 @@ let host_supports_hvm ~__context host =
...
+              get_management_ip_addr ~__context
+          in
+          D.debug "Helpers.Stunnel.gen_pem_if_not_exist: writing pem file";
+          let (_: string) = call_script "/opt/xensource/libexec/generate_ssl_cert" [cert; Option.value ~default:hostname cn]
+          in ()
+        end
+      )
+  end
+
+  let systemctl cmd = call_script "/usr/bin/systemctl" [cmd; "stunnel@xapi"]
```
robhoes: We should define this path in xapi_globs.ml and add the `systemctl` executable to the list of `essential_executables` in there. I see that it is used in this way in a couple of other places, so they could do with a cleanup as well.

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,122 @@ let host_supports_hvm ~__context host =
 let env_with_path env_vars =
   Array.of_list @@
   List.map (fun (k,v) -> Printf.sprintf "%s=%s" k v) (("PATH", String.concat ":" Forkhelpers.default_path)::env_vars)
+
+(** wrapper around the stunnel@xapi systemd service.
+  * there exist scripts (e.g. xe-toolstack-restart) which also manipulate
+    the stunnel daemon but they do this directly (not via ocaml). *)
+module Stunnel : sig
```
robhoes: It's not a problem, but I would have just put this into a new file.

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,122 @@ let host_supports_hvm ~__context host =
...
+
+  (** protect ourselves from concurrent writes to files *)
+  module Config : sig
+    (** create or update stunnel config file *)
+    val update : accept:string -> unit
+    (** calls `generate_ssl_cert` bash script
+      * TODO: don't generate certificates in bash *)
+    val generate_pem : __context:Context.t -> force:bool -> unit
+  end = struct
+    let xapi_stunnel_conf = "/etc/stunnel/xapi.conf"
```
robhoes: This should also go into xapi_globs.ml.

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,122 @@ let host_supports_hvm ~__context host =
...
+          let cn =
+            if Astring.String.is_prefix ~affix:"localhost" hostname then
+              get_management_ip_addr ~__context
+            else if Astring.String.is_infix ~affix:"." hostname then
+              None
+            else
+              get_management_ip_addr ~__context
+          in
+          D.debug "Helpers.Stunnel.gen_pem_if_not_exist: writing pem file";
+          let (_: string) = call_script "/opt/xensource/libexec/generate_ssl_cert" [cert; Option.value ~default:hostname cn]
```
robhoes: This should also go into globs. See how `xapissl` is done as an example. Also, we should never hardcode paths like`/opt/xensource/libexec`

ocaml/xapi/helpers.ml
```diff
@@ -1234,3 +1234,122 @@ let host_supports_hvm ~__context host =
...
+    let m = Mutex.create ()
+    let current_accept = ref None
+
+    let rewrite_xapi_ssl_config_file ~accept =
+      let fips = match Sys.getenv_opt "CC_PREPARATIONS" with
+                 | Some "true"   -> true
+                 | Some _ | None -> false
+      in
+      let conf_contents = String.concat "\n" [
+          "; autogenerated by helpers.ml"
```
robhoes: "by xapi"

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -29,43 +29,17 @@ let management_interface_server = ref []
...
+let restart_stunnel ~__context ~accept =
   info "Restarting stunnel (accepting connections on %s)" accept;
-  let xapissl_args = [ "restart"
-                     ; accept
-                     ; "permfile=" ^ !Xapi_globs.server_cert_path
-                     ; Xcp_const.good_ciphersuites
-                     ]
-  in
-  let (_ : Thread.t) = Thread.create (fun () ->
-      Mutex.execute management_m (fun () ->
```
robhoes: This `management_m` mutex is interesting...

I see that `run` takes the mutex. This calls `start`, which calls `restart_stunnel`, which calls `restart_stunnel_nomutex`, which gets us here and blocks. Luckily it's in a thread, so `start` and `run` can complete to unlock the mutex, so that stunnel can be restarted. I'm not sure this worked as intended... Unless there is a very good reason to postpone restarting stunnel until after `run` completes, I think that we're better off without the mutex (as in your patch)... :)

ocaml/xapi/helpers.ml
lippirk: if the module were to get much bigger i would agree

ocaml/xapi/xapi_mgmt_iface.ml
lippirk: i postulate that putting the call to the xapissl script was a workaround for this mutex being locked overzealously

i believe we can remove the thread (in my initial patch, i did not wrap the call to `Helpers.Stunnel.restart` in a thread and tests were looking good)

edwintorok:
@lippirk can you squash your fixup? should be good to go then

lippirk:
> @lippirk can you squash your fixup? should be good to go then

yup, i will don't worry

i'm gonna check that the CC tests pass, and then merge

PR #4040, opened 2020-02-24
CP-33121: Stop using Xstringext in cli_operations

psafont:

### DIFF
--- a/ocaml/xapi/cli_operations.ml
+++ b/ocaml/xapi/cli_operations.ml
@@ -19,7 +19,6 @@ open Cli_protocol
 open Cli_util
 open Cli_cmdtable
 open Stdext
-open Xstringext
 open Pervasiveext
 open Listext
 
@@ -92,7 +91,10 @@ let waiter printer rpc session_id params task =
    also match old syntax 'device-config-key' for backwards compatability *)
 let read_map_params name params =
   let len = String.length name + 1 in (* include ':' *)
-  let filter_params = List.filter (fun (p,_) -> (String.startswith name p) && (String.length p > len)) params in
+  let filter_params =
+    List.filter (fun (p,_) ->
+      (Astring.String.is_prefix ~affix:name p) && (String.length p > len)
+    ) params in
   List.map (fun (k,v) -> String.sub k len (String.length k - len),v) filter_params
 let read_set_params name params = List.map fst (read_map_params name params)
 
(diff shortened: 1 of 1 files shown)


ocaml/xapi/cli_operations.ml
```diff
@@ -167,7 +169,9 @@ let diagnostic_net_stats printer rpc session_id params =
   let uri (_, u, _) =
     not (List.mem_assoc "uri" params)
     || (String.lowercase_ascii u = String.lowercase_ascii (List.assoc "uri" params)) in
-  let has_param x = not(List.mem_assoc "params" params) || (List.mem x (String.split ',' (List.assoc "params" params))) in
+  let has_param x =
+    not (List.mem_assoc "params" params)
+    || (List.mem x (String.split_on_char ',' (List.assoc "params" params))) in
```
edwintorok: can this use List.assoc_opt here instead of mem+assoc?

ocaml/xapi/cli_operations.ml
```diff
@@ -832,7 +839,10 @@ let pool_disable_binary_storage printer rpc session_id params =
 
 let pool_ha_enable printer rpc session_id params =
   let config = read_map_params "ha-config" params in
-  let uuids = if List.mem_assoc "heartbeat-sr-uuids" params then String.split ',' (List.assoc "heartbeat-sr-uuids" params) else [] in
+  let uuids = if List.mem_assoc "heartbeat-sr-uuids" params then
```
edwintorok: assoc_opt

ocaml/xapi/cli_operations.ml
psafont: Can't, this `List` is from stdext, I'll add it to https://github.com/xapi-project/xen-api/pull/4038 I already have removed Listext for this reason.

psafont: See https://github.com/xapi-project/xen-api/commit/acd69b306

lindig:
The code for handling parameter lists is a good candidate for simplification. I understand that this is not the point of this PR but I am horrified by the endless string matching and splitting. A first step would be to document what key and values we actually support and to come up with an interface for a sub module.

ocaml/xapi/cli_operations.ml
```diff
@@ -1871,7 +1885,9 @@ let select_vms ?(include_control_vms = false) ?(include_template_vms = false) rp
   let params = if not include_template_vms then ("is-a-template"    , "false") :: params else params in
   let vm_name_or_ref = try Some (
       (* Escape every quote character *)
-      List.assoc "vm" params |> String.replace "\"" "\\\""
+      let quote = Re.(compile (str {|"|})) in
```
edwintorok: you could compile the regex just once if you put this somewhere global

psafont: I've moved it outside the function and reused it for sr and host parameters.

PR #4041, opened 2020-02-26
Havana Backports

lindig:
This PR is for a series of backports:

* be8ff590f CA-333712: use wrapped string type for pool internal API calls   
* 010b3759f CA-329466: Simplify logging for plugins' parameters              
* 49bf6d71e CA-329835 Improve logging                                        
* 1523addcd CA-329843: broaden usage of secrets API                          
* 6fe63ab87 CA-330902 host-bugreport-upload args via env vars                
* b69de0058 CA-330902 Improve logging - backport it                          
* cef84c2a7 CA-330902 Improve logging                                        
* 6609d0346 CA-330693: Limit access to state.db to just root                 
* 49351479d CA-330961 Clean the yum cache before prechecking/applying

Except for CA-333712 the backports have not required considerable effort.

### DIFF
--- a/ocaml/xapi/storage_migrate.ml
+++ b/ocaml/xapi/storage_migrate.ml
@@ -25,8 +25,19 @@ open Pervasiveext
 open Xmlrpc_client
 open Threadext
 
-let local_url () = Http.Url.(Http { host="127.0.0.1"; auth=None; port=None; ssl=false }, { uri = "/services/SM"; query_params=["pool_secret",!Xapi_globs.pool_secret] } )
-let remote_url ip = Http.Url.(Http { host=ip; auth=None; port=None; ssl=true }, { uri = "/services/SM"; query_params=["pool_secret",!Xapi_globs.pool_secret] } )
+let with_cookie t request =
+  let cookie = [("pool_secret", t)] in
+  {request with Http.Request.cookie}
+
+let storage_url ?pool_secret url = url, pool_secret
+
+let local_url () = Http.Url.(Http { host="127.0.0.1"; auth=None; port=None;
+                                    ssl=false }, { uri = Constants.sm_uri; query_params=[] } )
+                   |> storage_url ~pool_secret:!Xapi_globs.pool_secret
+
+let remote_url ip = Http.Url.(Http { host=ip; auth=None; port=None; ssl=true },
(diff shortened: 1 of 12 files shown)


lippirk:
your backports of my commits look correct

ocaml/idl/datamodel.ml
```diff
@@ -18,7 +18,7 @@ open Datamodel_types
 (* IMPORTANT: Please bump schema vsn if you change/add/remove a _field_.
               You do not have to bump vsn if you change/add/remove a message *)
 let schema_major_vsn = 5
-let schema_minor_vsn = 109
+let schema_minor_vsn = 110
```
lippirk: which field was removed/added?

edwintorok: I don't think we should change the version number in a backport unless we really have to. It might make upgrades to next release more difficult (because they would be partially upgrading, partially downgrading the database when installing a new RTM build).
This might've been bumped to trigger the update code, but I see I used `fun _ -> true` in the upgrade code, so the version number is ignored. I think it'd be safe to not bump the version number here, as long as the upgrade code is still triggered.

ocaml/xapi/storage_migrate.ml
```diff
@@ -25,8 +25,19 @@ open Pervasiveext
...
+  {request with Http.Request.cookie}
+
+let storage_url ?pool_secret url = url, pool_secret
+
+let local_url () = Http.Url.(Http { host="127.0.0.1"; auth=None; port=None;
+                                    ssl=false }, { uri = Constants.sm_uri; query_params=[] } )
+                   |> storage_url ~pool_secret:!Xapi_globs.pool_secret
+
+let remote_url ip = Http.Url.(Http { host=ip; auth=None; port=None; ssl=true },
+                              { uri = "/services/SM"; query_params=["pool_secret",!Xapi_globs.pool_secret] } )
```
edwintorok: We must not put this into the query params, it must go into the cookie.

ocaml/idl/datamodel.ml
lindig: This might be a mistake. In on of the original commits the schema was updated but this might not be required for the backport.

lindig:
I've squashed the fixups and will merge this. Thanks for the reviews.

PR #4027, opened 2020-02-06
Firstboot toolstack changes

rosslagerwall:
This must be merged along with the spec file change.

PR #4042, opened 2020-03-02
gencert systemd service to generate ssl certificate for the stunnel daemon

lippirk:
final PR for REQ-811

Instead of xapi figuring out whether it needs to generate a certificate, leave it up to systemd

### DIFF
--- a/ocaml/gencert/gencert.ml
+++ b/ocaml/gencert/gencert.ml
@@ -1,0 +1,67 @@
+(*
+ * Copyright (C) Citrix Systems Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+(* Usage: gencert <path> *)
+
+let inventory = "/etc/xensource-inventory"
+let generate_ssl_cert = "/opt/xensource/libexec/generate_ssl_cert"
+
+module Lib = Gencertlib.Lib
+
+let generate_cert_or_fail ~path ~cn =
+  let args = [path; cn] in
+  let (stdout, stderr) = Forkhelpers.execute_command_get_output generate_ssl_cert args in
+
+  Printf.printf {|generate_ssl_cert stdout: "%s"|} stdout;
+  Printf.printf {|generate_ssl_cert stderr: "%s"|} stderr;
+  if Sys.file_exists path then
+    Printf.printf "file exists (%s), assuming cert was created" path
+  else begin
+    Printf.eprintf "file doesn't exist (%s): cert was not created" path;
+    exit 1
+  end
+
+let main ~dbg ~path =
+  let init_inventory () =
+    Inventory.inventory_filename := inventory
+  in
+  init_inventory ();
+  let hostname = Unix.gethostname () in
+  let cn =
+    if Astring.String.is_prefix ~affix:"localhost" hostname then
+      Lib.get_management_ip_addr ~dbg
+    else if Astring.String.is_infix ~affix:"." hostname then
+      None
+    else
+      Lib.get_management_ip_addr ~dbg
+  in
+  let cn = Option.value ~default:hostname cn in
+  generate_cert_or_fail ~path ~cn
+
+let () =
+  let dbg = Printf.sprintf "gencert - %f" (Unix.gettimeofday ()) in
+
+  if Array.length Sys.argv <> 2 then begin
+    Printf.eprintf "Usage: gencert PATH";
+    exit 1
+  end;
+
+  let path = Sys.argv.(1) in
+
+  if Sys.file_exists path then begin
+      Printf.printf "file already exists at path (%s) - doing nothing" path;
+      exit 0
+    end
+  else
+    main ~dbg ~path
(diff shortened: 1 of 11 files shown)


scripts/10-stunnel-increase-number-of-file-descriptors.conf
```diff
@@ -1,2 +1,6 @@
+[Unit]
```
lippirk: should maybe put dependency on gencert in another conf file?

ocaml/gencert/lib.ml
```diff
@@ -0,0 +1,29 @@
+
```
lippirk: i needed this code from xapi, so am exposing it as a library from gencert

i don't think this is right, perhaps it belongs in the network client library instead?

ocaml/gencert/gencert.ml
```diff
@@ -0,0 +1,67 @@
+(*
+ * Copyright (C) 2006-2009 Citrix Systems Inc.
```
edwintorok: Best to remove the year, its wrong.

scripts/10-stunnel-increase-number-of-file-descriptors.conf
```diff
@@ -1,2 +1,6 @@
...
+Requires=gencert.service
```
edwintorok: This can be put into gencert.service itself if you invert the direction, i.e. instead of Requires say RequiredBy.
Ask @TimSmithCtx or @rosslagerwall  for the preferred way to do this with systemd.

scripts/Makefile
```diff
@@ -54,6 +54,7 @@ install:
 	$(IPROG) control-domain-params-init $(DESTDIR)$(LIBEXECDIR)/control-domain-params-init
 	$(IPROG) reset-and-reboot $(DESTDIR)$(LIBEXECDIR)/reset-and-reboot
 	$(IDATA) cdrommon@.service $(DESTDIR)/usr/lib/systemd/system/cdrommon@.service
+	$(IDATA) gencert.service $(DESTDIR)/usr/lib/systemd/system/gencert.service
```
edwintorok: Where is gencert.service?

lippirk: mb, it's in the fixup

scripts/10-stunnel-increase-number-of-file-descriptors.conf
lippirk: yes, it looks like i want `RequiredBy`

lippirk: inverting the dependencies means that we have to enable gencert for the following symlink to be created:

```
/etc/systemd/system/stunnel@xapi.service.requires/gencert.service to /usr/lib/systemd/system/gencert.service.
```

if we don't enable it, we get an error in daemon.log because the cert never gets created.

whereas if we just leave the gencert dependency in the stunnel@xapi drop-in, we don't have to worry about enabling gencert (so i would suggest that this is preferred)

ocaml/gencert/lib.ml
psafont: Does `ocaml/xapi/helpers.ml > get_management_ip_addr` not work in this case? (It retrieves the ip from the inventory)

Even `ocaml/xapi/xapi_mgmt_iface.ml` uses this function.

psafont: I see that you've changed _after_ this commit; it's clear now that this deals with ipv6 as well and the function I mention gets changed to use this one as well.

ocaml/xapi/helpers.ml
```diff
@@ -1322,9 +1301,9 @@ end = struct
     | unknown    -> D.error "Stunnel.is_enabled: expected 'enabled' or 'disabled', but got: %s" unknown; false
 
   let restart ~__context ~accept =
+    (* generating certificates is handled by gencert systemd service *)
```
psafont: I think this comment will be more confusing than helpful for future readers

scripts/gencert.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Generate xapi ssl certificate
+Requires=xapi.service forkexecd.service xapi-init-complete.target
+After=xapi.service forkexecd.service xapi-init-complete.target
+
+[Service]
+Type=oneshot
+ExecStart=/opt/xensource/libexec/gencert /etc/xensource/xapi-ssl.pem
+RemainAfterExit=yes
```
psafont: Does the default of `no` introduce problems? I thought this was needed in case the service forked and systemd lost visibility on it but still needs to stop it.

Reading the docs it seems the default should be OK in this case for dependency purposes:
```
Behavior of oneshot is similar to simple; however, the service manager will consider the unit up after the main
               process exits. It will then start follow-up units.
```

ocaml/gencert/lib.ml
lippirk: i think i'll just get rid of the commit 'share get_management_ip_logic'

i.e. I'll merge it into the previous

ocaml/xapi/helpers.ml
lippirk: i've made it more clear now

scripts/gencert.service
lippirk: i doubt it would cause problems, but if we set `RemainAfterExit=no`, then systemd will run the gencert binary each time we start/restart stunnel which isn't necessary

lippirk: the firstboot changes use `RemainAfterExit=yes` also https://github.com/xapi-project/xen-api/pull/4027/files

ocaml/gencert/lib.ml
```diff
@@ -0,0 +1,42 @@
...
+let get_management_ip_addr ~dbg =
+  let iface = Inventory.lookup Inventory._management_interface in
+  let primary_address_type =
+    match String.lowercase_ascii (Inventory.lookup Inventory._management_address_type ~default:"ipv4") with
+    | "ipv4" -> `IPv4
+    | "ipv6" -> `IPv6
+    | s      -> raise (Unexpected_address_type (Printf.sprintf "Expected 'ipv4' or 'ipv6', got %s" s))
+  in
+  if iface = "" then
```
lindig: I think the nesting of the logic here is quite deep. 

```
try 
  if iface = "" || not @@ Net.Interface.exists dbg iface then
    None
  else
   ...
with
  _ -> None

lippirk:
thanks for reviewing - bvt/bst looking good

PR #4043, opened 2020-03-05
CA-336011: WLB recommend VMs to the same host

liulinC:
WLB provide recommendation basing on history data, the data takes
time to be awared by WLB. If multiple VM starts at the same time,
all the VMs are recommended to the same host, as WLB does not has
the just started VM running data.

This commit heat the just chosen host and cool down the host as time
goes by to choose the best host for any VM.

### DIFF
--- a/ocaml/xapi/xapi_vm_helpers.ml
+++ b/ocaml/xapi/xapi_vm_helpers.ml
@@ -31,6 +31,13 @@ module SRSet = Set.Make(struct
   let compare = Stdlib.compare
 end)
 
+module Cool_down_map = Map.Make(struct
+  type t = API.ref_host
+  let compare = Stdlib.compare
+end)
+
+let cool_down_data = ref Cool_down_map.empty
+
 let compute_memory_overhead ~__context ~vm =
   let vm_record = Db.VM.get_record ~__context ~self:vm in
   Memory_check.vm_compute_memory_overhead vm_record
@@ -790,12 +797,22 @@ let choose_host_uses_wlb ~__context =
       (Db.Pool.get_other_config ~__context
          ~self:(Helpers.get_pool ~__context)))
 
+
 (** Given a virtual machine, returns a host it can boot on, giving   *)
 (** priority to an affinity host if one is present. WARNING: called  *)
 (** while holding the global lock from the message forwarding layer. *)
 let choose_host_for_vm ~__context ~vm ~snapshot =
   if choose_host_uses_wlb ~__context then
     try
+      let current_time = Unix.time() in
+      let filter_calm_down value =
+        List.filter (fun e -> current_time -. e <= !Xapi_globs.wlb_host_cool_down_time) value
+      in
+      let rec pow base = function
+        | 0 -> 1.0
+        | 1 -> base
+        | n -> base *. (pow base (n-1))
+      in
       let possible_rec (host, recommendation) =
         match recommendation with
         | Recommendation {source; id; score} ->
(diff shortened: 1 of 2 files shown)


ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -790,25 +797,62 @@ let choose_host_uses_wlb ~__context =
...
+        let entry_str = hostname ^ " " ^ time_str ^ "\n" in
+        debug "cool_down_data entry %s" entry_str
+      in
       let cmp_descending (_, s, _) (_, s', _) =
         Float.compare s' s
       in
+      (* WLB provide recommendation basing on history data which take time to be awared by WLB
+       * In case of start multiple VMs at the same time, WLB choose the best host for the first
+       * VM, then the second VM starts, WLB does not has the first VM running data and the
+       * influence. Xapi heat and cool down the host to choose the best one.
```
robhoes: What if the best choice of host for the next VM would actually be the same host as the last VM started on? If I understand it correctly, this logic seems to arbitrarily penalise this host?

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -790,25 +797,62 @@ let choose_host_uses_wlb ~__context =
...
+      let cool_down_host (host, score, id) =
+        match Cool_down_map.find_opt host !cool_down_data with
+        | None -> (host, score, id)
+        | Some times ->
+          let penalty = List.length times |> pow !Xapi_globs.wlb_host_cool_down_penalty in
+          let penaltied_value = score *. penalty in
+          (host, (penaltied_value), id)
+      in
+      let heat_host = function
+        | Some times -> Some (current_time::times)
```
edwintorok: Can this list become very long? Should it have an upper-bound?

ocaml/xapi/xapi_vm_helpers.ml
psafont: I would like to see a sequence diagram to see the problematic interaction to see where and how it can be better tackled.

I have doubts this is the right approach, I would rather have WLB not recommend moving a VM until it the load in the pool has stabilized and can do reliable recommendations.

ocaml/xapi/xapi_vm_helpers.ml
liulinC: The list contains the VM starts on the host during last N second, (N configurable with default value 10, it will not be huge)

ocaml/xapi/xapi_vm_helpers.ml
liulinC: @robhoes yes, this logic arbitrarily penalise this host,  That make sense as the just started VM will surely add new load to this host. But WLB has no data about how much load it will add to the this host.

@pauldu The problem is dscribed [here](https://sequencediagram.org/index.html#initialData=C4S2BsFMAIFkFdygA5WgNVtAzsAhgE7DbR7DTAAWM2eAtjKA9MgQPYBGUdAUD5gEYAtEIB8ACTa4BALmgCAdBlgCc+IgBpoAdQAyAIWgFIAYzZ0GAOwAm0SdK0APPMhDQTlNlJifpfe8DConr6cgBMSiHubOBQJuSCRvCWliCWAObQaQBmbKQ20PDI1mSQtgGqeNnAkASkHGzw5AIADDimbDbYfJhhIhJSwGFyAMxKvdAAyurEWlHGZlblg2FOLm64hAmwYdCddisK-LAjYgHD0AAs4ydTM9hzBtDWbJAklmzkANYfAO4U1FI2WyeHiyhGe0sB1wqzUIFihWwMAA8uBbCV8FoFuYltChkpnK53L4kXiwvw6Jd+uc5ABWJSTEB0eGECh5TAjfwrMQhOSWSD-XxDaAMYAEEAmEhFDFlNk6AxHIA)  Please let me know if that is not clear.
In short, in case of multiple VM starts at the same time, WLB does not know which VM will be chose by xapi to start the VM, and how much the just started vm will affact the host (WLB needs time to know that with new later metrics.) 
Even more, we would not know when WLB is ready for the change.

Actually, This change does not make big differece as it looks like, as most of time, the cool_down_data is emty. (it only has the VM starts on the host in recent 10 seconds)

Shall we have a meeting to discuss that?

robhoes: I am also concerned that we are confusing the policies here.

We have two clear ways for `choose_host_for_vm` to do its job, which is to find a host for a given VM to start on:
1. WLB is asked for recommendations, and xapi follows this.
2. WLB is not enabled, and xapi does it's own assessment.

What happens in this patch looks like a combination of both, which makes it hard to reasons about what is going on and who is in charge of the placement policy. I would like to avoid this.

Is it not possible to implement this cool-down policy in WLB itself?

liulinC: Yes, it is possible, But the effect may not as good as here.

WLB only provide recommendations, it provide a score to every host,
It actually does not know which host is chosen by xapi.
It can only know the result until got a VM start Event. so it will be a short delay here.
But technically, we can cool down host in WLB side,
If lots of requests come in very short time (say 1~2 seconds),  WLB will have no change to heat the host.

If head/cool down is concern in xapi, let me try in WLB and see how it works.

robhoes: The other problem is that `choose_host_for_vm` is not always used for VMs to be started. For example, `VM.start_on` does not use it (as you explicitly specify a host), and similarly VM migration is not taken into account. So you're not catching all "start" events with this patch, only the instances where WLB has done a recommendation.

The other way around, `choose_host_for_vm` may be used when _not_ starting a VM, so you are getting false positives.

robhoes:
I think that we have agreed to fix this inside WLB instead, so let's close this.

PR #4044, opened 2020-03-06
Remove iLO script

lippirk:
Our iLO scripts uses TLS<1.2. Rather than replace it, remove it


ocaml/idl/datamodel_host.ml
```diff
@@ -1044,7 +1044,7 @@ let host_query_ha = call ~flags:[`Session]
       ~doc:"Set the power-on-mode, host, user and password "
       ~params:[
         Ref _host, "self", "The host";
-        String, "power_on_mode", "power-on-mode can be empty,iLO,wake-on-lan, DRAC or other";
+        String, "power_on_mode", "power-on-mode can be empty, wake-on-lan, DRAC or other";
```
edwintorok: I think we'll need a Changed rel_stockholm to say that iLO mode is removed.

scripts/poweron/power-on.py
```diff
@@ -41,18 +41,14 @@ def main(session, args):
     
     power_on_config = session.xenapi.host.get_power_on_config(remote_host)
     
-    if mode == "iLO" or mode=="DRAC" :
+    if mode=="DRAC" :
```
edwintorok: What if someone had this configured to iLO previously? Should we log something here?

lindig:
The Travis failure looks suspicious:
```
- 1259 |         ; Printf.sprintf "ciphers = %s" Xcp_const.good_ciphersuites
-                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
- Error: Unbound module Xcp_const
[ERROR] The compilation of xapi failed at "/home/opam/.opam/4.08/bin/dune build -p xapi".
```

Is this a missing dependency?

edwintorok:
I think that is due to other changes on the feature branch.
Travis isn't set up for the feature branch, so it builds those from master, and master doesn't have that.
Perhaps we could already merge the IDL changes from the feature branch, they are unlikely to change?

lindig:
Any reason to not merge this?

lippirk:
> Any reason to not merge this?

yes - we're waiting on approval to remove iLO

scripts/poweron/power-on.py
```diff
@@ -40,30 +40,34 @@ def main(session, args):
     mode = session.xenapi.host.get_power_on_mode(remote_host)
```
psafont: I'd prefer if we used an Enum to represent the values for mode, but that its only supported starting in python 3.4 or through a library, without talking about increased burden to maintain the values of the enum. :(

PR #4047, opened 2020-03-09
Enable testing for daily license checks

psafont:


ocaml/license/dune
```diff
@@ -1,23 +1,21 @@
...
+  )
+)
 
-let () = Printf.ksprintf Jbuild_plugin.V1.send {|
 (executable
   (name daily_license_check_main)
   (public_name daily-license-check)
   (package xapi)
-  (modules (:standard \ test_daily_license_check))
+  (modules daily_license_check_main)
```
edwintorok: If you moved the file that was excluded then you can just remove this line completely.

ocaml/tests/test_daily_license_check.ml
```diff
@@ -0,0 +1,83 @@
...
+    Expired ["host1"; "host0"]
+  ; (["expiry", "20160101T00:00:00Z"],
+    ["host0", ["expiry", "20170601T00:00:00Z"];
+     "host1", ["expiry", "20150601T00:00:00Z"]]),
+    Expired ["host1"]
+]
+
+let test =
+  let tests () =
+    List.iter test_expiry expiry_samples;
```
edwintorok: You could map each test from expiry_samples to an alcotest test, and then we don't need to use List.iter (and if you have more than 1 failure Alcotest can tell you about each one individually).
e.g. if you add a `()` parameter to `test_expiry`, then something like:
`List.mapi (fun i spec -> Printf.sprintf "Test license checks as run daily #%d" i, `Quick, test_expiry spec) expiry_samples`

ocaml/license/dune
psafont: Since I'm building a library with the other ml file in the folder I can't remove this line.

edwintorok:
Lets wait for Travis and then we can squash and merge it

ocaml/tests/test_daily_license_check.ml
```diff
@@ -0,0 +1,82 @@
...
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+open Daily_license_check
```
lindig: Do we need this? Below this is opened again locally

psafont: we do not

PR #4049, opened 2020-03-12
CP-32398 must not use POD when using SRIOV vGPU

lindig:
Add a check into VM start to check its memory configuration if the VM
has an SRIOV vGPU. We disable ballooning (DMC) and POD (populate on
demand) by pinning memory to static max. The main point is disabling POD
but we found it difficult to implement while still permitting
ballooning. So we are disabling both by demanding that the memory
configuration is fixed at the upper limit at the start of the VM.

### DIFF
--- a/ocaml/xapi/xapi_vm.ml
+++ b/ocaml/xapi/xapi_vm.ml
@@ -225,15 +225,26 @@ let unpause ~__context ~vm =
 let set_xenstore_data ~__context ~self ~value =
   Xapi_xenops.set_xenstore_data ~__context ~self value
 
-(* CP-18860: check memory limits if using nested_virt on start *)
+let is_sriov_vgpu ~__context vgpu =
+  let vgpu = Db.VGPU.get_record_internal ~__context ~self:vgpu in
+  let ty = vgpu.Db_actions.vGPU_type in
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_vm.ml
```diff
@@ -225,15 +225,26 @@ let unpause ~__context ~vm =
...
+  if Vm_platform.is_true ~key:"nested-virt" ~platformdata ~default:false then begin
+    (* CP-18860: check memory limits if using nested_virt on start *)
+    let c = C.get ~__context ~vm_ref:vm in
+    C.assert_valid_and_pinned_at_static_max ~constraints:c ~reason:"nested virt"
+  end;
+  let vgpus = Db.VM.get_VGPUs ~__context ~self:vm in
+  if List.exists (is_sriov_vgpu ~__context) vgpus then begin
+    (* CP-32398: must not use POD when using SRIOV vGPU *)
+    let c = C.get ~__context ~vm_ref:vm in
+    C.assert_valid_and_pinned_at_static_max ~constraints:c ~reason:"SRIOV vGPU"
```
edwintorok: Nice, we already had functions that check this.

PR #4050, opened 2020-03-12
CP-32696: Add alerts for certificate expiration

psafont:
Adds a daily cron job that checks the validity of the certificates for all hosts.
It also checks alerts for hosts with installed certificates and clears them before adding the new alerts.

Depends on https://github.com/xapi-project/xen-api/pull/4038

### DIFF
--- a/ocaml/alerts/certificate/certificate_check.ml
+++ b/ocaml/alerts/certificate/certificate_check.ml
@@ -1,0 +1,70 @@
+module XenAPI = Client.Client
+module Date = Xapi_stdext_date.Date
+
+let seconds_per_day = 3600. *. 24.
+let seconds_per_30_days = 30. *. seconds_per_day
+
+let days_until_expiry epoch expiry =
+  int_of_float (expiry /. seconds_per_day -. epoch /. seconds_per_day)
+
+let get_certificate_attributes rpc session =
+  XenAPI.Certificate.get_all_records rpc session
+  |> List.map (fun (_, certificate) ->
+    certificate.API.certificate_host,
+    certificate.API.certificate_not_after
+  )
+
+let generate_alert epoch (host, expiry) =
+  let days = days_until_expiry epoch (Date.to_float expiry) in
+  if days > 30 then
+    host, None
+  else
+    let expiring_message = "The TLS server certificate is expiring soon." in
+    let expired_message = "The TLS server certificate has expired." in
+
+    let body msg =
+      Printf.sprintf "<message>%s</message><date>%s</date>" msg (Date.to_string expiry)
+    in
+    let message, alert =
+      if days < 0 then
+        body expired_message, Api_messages.host_server_certificate_expired
+      else if days < 8 then
+        body expiring_message, Api_messages.host_server_certificate_expiring_07
+      else if days < 15 then
+        body expiring_message, Api_messages.host_server_certificate_expiring_14
+      else
+        body expiring_message, Api_messages.host_server_certificate_expiring_30
+    in
+    host, Some (message, alert)
+
+let execute rpc session (host, alert) =
+  let host_uuid = XenAPI.Host.get_uuid rpc session host in
+  (* we need to remove any host_server_certificate_expiring messages affecting
+     this host as it needs to be refreshed *)
+  let obsolete_messages = XenAPI.Message.get_all_records rpc session
+  |> List.filter_map (fun (ref, record) ->
+      let prefix = Api_messages.host_server_certificate_expiring in
+      if Astring.String.is_prefix ~affix:prefix record.API.message_name
+           && record.API.message_obj_uuid = host_uuid then
+        Some ref
+      else
+        None
+      )
+  in
+  List.iter (fun self -> XenAPI.Message.destroy rpc session self) obsolete_messages;
+
+  match alert with
+  | None ->
+      ()
+  | Some (message, (alert, priority)) ->
+    ignore (XenAPI.Message.create rpc session alert priority `Host host_uuid message)
+
+let alert rpc session =
+  let now = Unix.time () in
+  let send_alert_maybe attributes =
+    attributes
+    |> generate_alert now
+    |> execute rpc session
+  in
+  get_certificate_attributes rpc session
+  |> List.iter send_alert_maybe
--- a/ocaml/xapi-consts/api_messages.ml
+++ b/ocaml/xapi-consts/api_messages.ml
@@ -12,13 +12,15 @@
  * GNU Lesser General Public License for more details.
  *)
 
-(*
+(** Each message string in this file has to be unique and it can only have one
+   priority.
+
 Priority Name                  Description
 -------- --------------------- -----------------------------------------------------------------------------
-1        Data-loss imminent    Take action now or your data may be permanently lost (e.g. corrupted)
-2        Service-loss imminent Take action now or some service(s) may fail (e.g. host / VM crash)
-3        Service degraded      Take action now or some service may suffer (e.g. NIC bond degraded without HA)
-4        Service recovered     Notice that something just improved (e.g. NIC bond repaired)
+1        Critical              Take action now or your data may be permanently lost (e.g. corrupted)
+2        Major                 Take action now or some service(s) may fail (e.g. host / VM crash)
+3        Warning               Take action now or some service may suffer (e.g. NIC bond degraded without HA)
+4        Minor                 Notice that something just improved (e.g. NIC bond repaired)
 5        Informational         More day-to-day stuff (e.g. VM started, suspended, shutdown, rebooted etc)
 *)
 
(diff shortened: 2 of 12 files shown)


psafont:
Not sure why the DCO bot is not happy

psafont:
Only the last commit in the series should be reviewed, the rest is part of https://github.com/xapi-project/xen-api/pull/4038

edwintorok:
Commit sha: b4f24b0, Author: Pau Ruiz Safont, Committer: Pau Ruiz Safont; The sign-off is missing.

psafont:
ah, good, we don't want it merged to master.

psafont:
This PR needs a change in the packaging as we're installing 2 more files.

robhoes:
You probably want to target the feature branch?

ocaml/alerts/certificate/certificate_check.ml
```diff
@@ -0,0 +1,63 @@
...
+module Date = Xapi_stdext_date.Date
+
+let seconds_per_day = 3600. *. 24.
+let seconds_per_30_days = 30. *. seconds_per_day
+
+let days_until_expiry epoch expiry =
+  int_of_float (expiry /. seconds_per_day -. epoch /. seconds_per_day)
+
+let get_certificate_attributes rpc session =
+  XenAPI.Certificate.get_all_records rpc session
```
robhoes: Since this get the certs of all hosts in the pool, we need to make sure that the script runs on just one host (the master). I can't remember if/how the license-check does that.

ocaml/alerts/certificate/certificate_check.ml
```diff
@@ -0,0 +1,63 @@
...
+    in
+    host, Some (message, alert)
+
+let execute rpc session (host, alert) =
+  let host_uuid = XenAPI.Host.get_uuid rpc session host in
+  (* we need to remove any host_server_certificate_expiring messages affecting
+     this host as it needs to be refreshed *)
+  let obsolete_messages = XenAPI.Message.get_all_records rpc session
+  |> List.filter_map (fun (ref, record) ->
+      if record.API.message_name = (fst Api_messages.host_server_certificate_expiring_07)
```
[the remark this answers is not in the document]
psafont: the first element of the 3 tuples is the same string, this was the most convenient way to extract the information, would you prefer to put a name to the string in `api_messages` and use it here?

ocaml/alerts/certificate/certificate_check.ml
psafont: I haven't seen anything ensuring this on license-check, which is bizarre, as I would expect it to just execute on a single host.

ocaml/xapi-consts/api_messages.ml
```diff
@@ -139,3 +141,9 @@ let cluster_host_enable_failed = addMessage "CLUSTER_HOST_ENABLE_FAILED" 3L
 
 (* raised by external script in clustering daemon, do not delete this: it is not dead code *)
 let cluster_host_fencing = addMessage "CLUSTER_HOST_FENCING" 2L
+
+(* Certificate expiration messages *)
+let host_server_certificate_expiring_30 = addMessage "HOST_SERVER_CERTIFICATE_EXPIRING_30" 3L
+let host_server_certificate_expiring_14 = addMessage "HOST_SERVER_CERTIFICATE_EXPIRING_14" 2L
+let host_server_certificate_expiring_07 = addMessage "HOST_SERVER_CERTIFICATE_EXPIRING_07" 1L
```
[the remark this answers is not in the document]
psafont: If hosts in a pool cannot establish connections among themselves, or appliances cannot connect to them, isn't data-loss a possibility, this is present in the design.

ocaml/alerts/certificate/certificate_check.ml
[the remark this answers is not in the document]
robhoes: They are three different strings, starting with "HOST_SERVER_CERTIFICATE_EXPIRING_" + a number...?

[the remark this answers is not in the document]
psafont: ahhh, right. forgot to change this part when I made each to use a different one.

kc284:
I'd like to raise a couple of concerns:
- The alert is issued daily, however I believe the check should also run once immediately after certificate installation. While dev testing I noticed that, if installed a certificate expiring imminently, there would be no immediate alert. Presumably I would have to wait a day to see it. Consider an extreme case where a user installed a certificate with 1 day left: by the next day it would be too late.
- There is an issue with alert internationalisation, namely the certificate expiry date is hidden inside the message body which cannot be used on the UI as is. I have extracted the date ok with a regex, however this solution is a bit inelegant. It would be nicer if the message contained parameters (or an xml part like the performance alarms) from where the necessary values could be extracted.

ocaml/xapi-consts/api_messages.ml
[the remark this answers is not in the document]
robhoes: A cert expiring is similar to a network link failing – the service will break, but you shouldn't lose any data. However, level 2 seems a bit too weak for this case again ("some service may fail").

This is how the Citrix doc describes them: https://docs.citrix.com/en-us/citrix-hypervisor/monitor-performance.html#alerts. Level 1 is "Critical". Let's leave it this way.

robhoes:
👍 for running the check and alerting immediately after installing a new certificate.

psafont:
Last commit adds ad-hoc xml formatting for internazionalization. It also removes the hostname from the message as this is can be inferred from the host uuid which is sent, likewise with the days remaining.

ocaml/xapi-consts/api_messages.ml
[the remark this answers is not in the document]
psafont: I'll read those docs and add them to the comments.

ocaml/alerts/certificate/certificate_check.ml
robhoes: Let's update the cron script to only continue if the file `/etc/xensource/pool.conf` reads `master`.

psafont: I've made it fail if it's not run on a master, it seems more informative than exiting with 1 or not doing anything silently. Being successful with an info message could be another option, but I'm not sure where would that be logged, if at all.

psafont:
With the latest commit I think the issues have been addressed.

ocaml/alerts/certificate/certificate_check.ml
```diff
@@ -38,6 +38,10 @@ let generate_alert epoch (host, expiry) =
     host, Some (message, alert)
 
 let execute rpc session (host, alert) =
+  if Xapi_stdext_unix.Unixext.string_of_file "/etc/xensource/pool.conf" <> "master" then (
+    raise (Failure "Alerts cannot only be executed from a master host")
```
robhoes: Well, they can, but we want to do it in the cron job of just one host. I would therefore put this check into the cron script, which can just exit if not the master.

ocaml/alerts/certificate/certificate_check.ml
robhoes: IMO the program should always work, on any host. What we are constraining, is the scheduling by cron.

ocaml/xapi/message_forwarding.ml
```diff
@@ -2630,7 +2630,13 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
             certificate_chain
         );
       in
-      tolerate_connection_loss fn success 30.
+      tolerate_connection_loss fn success 30.;
+      try
+        let _, _ = Forkhelpers.execute_command_get_output "/opt/xensource/libexec/alert-certificate-check" [] in
```
robhoes: The path should be defined in xapi_globs.ml.

psafont: It's not, I'll define it, maybe we can slowly start using it as the path appears peppered around the code.

robhoes: I didn't mean `/opt/xensource/libexec`, but define the name of the command `alert-certificate-check` itself. In xapi_globs.ml, define it as `let alert_certificate_check = ref "alert-certificate-check"` and add it to `Resources.essential_executables` at the bottom of the file. This is used by logic in `Xcp_services.configure` (from xcp-idl), which is called at start up. This will search for the executables in `$PATH` and paths defined in the config file: we have `search-path = /opt/xensource/libexec:/opt/xensource/bin` in xapi.conf. See this being logged when xapi starts. So there is no need to have `/opt/xensource/libexec` in the code at all.

scripts/certificate-check
```diff
@@ -1,5 +1,12 @@
 #!/bin/sh
 
+HOST_ROLE=$(cat /etc/xensource/pool.conf)
+
+if [ "$HOST_ROLE" = "master\n" ]; then
+  echo "Not checking certificate expiration for alerts, this is not a master host"
```
robhoes: No, it _is_ the master.

psafont: Fixed

scripts/certificate-check
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+HOST_ROLE=$(cat /etc/xensource/pool.conf)
```
lindig: I think it would be safer to do:
```
if grep -q master /etc/xensource/pool.conf; then
..
fi
```
This protects against slight changes in that file, like a space, newline or something else that a `==` would miss

psafont: I think wwe want to use `-qv` to invert the match

scripts/certificate-check
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+if ! grep -q master /etc/xensource/pool.conf; then
+  echo "Not checking certificate expiration for alerts, this is not a master host"
+  exit 0
+ fi
```
lindig: indentation 😇

psafont: ✨

psafont:
Travis is failing due to connectivity problems.

ocaml/xapi/xapi_globs.ml
```diff
@@ -791,6 +791,8 @@ let list_domains = ref "/usr/bin/list_domains"
 
 let xen_cmdline_script = ref "/opt/xensource/libexec/xen-cmdline"
 
+let alert_certificate_check = ref "/opt/xensource/libexec/alert-certificate-check"
```
robhoes: Drop the `/opt/xensource/libexec/` part, and make it just `ref "alert-certificate-check"`. See my related comment. The "examples" above are not right.

PR #4052, opened 2020-03-13
fix travis for quebec

lippirk:
Testing travis with quebec

### DIFF
--- a/.travis-xs-opam.sh
+++ b/.travis-xs-opam.sh
@@ -4,10 +4,9 @@ set -ex
 
 PACKAGE="xapi"
 PINS="xapi:. xapi-cli-protocol:. xapi-client:. xapi-consts:. xapi-database:. xapi-datamodel:. xapi-types:. xe:."
-BASE_REMOTE="https://github.com/xapi-project/xs-opam.git"
 
 wget https://raw.githubusercontent.com/ocaml/ocaml-ci-scripts/master/.travis-docker.sh
-wget https://raw.githubusercontent.com/xapi-project/xs-opam/master/tools/xs-opam-ci.env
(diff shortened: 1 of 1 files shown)


.travis-xs-opam.sh
```diff
@@ -4,10 +4,10 @@ set -ex
 
 PACKAGE="xapi"
 PINS="xapi:. xapi-cli-protocol:. xapi-client:. xapi-consts:. xapi-database:. xapi-datamodel:. xapi-types:. xe:."
-BASE_REMOTE="https://github.com/xapi-project/xs-opam.git"
+BASE_REMOTE="https://github.com/xapi-project/xs-opam.git#quebec"
```
psafont: This is already defined in the xs-opam-ci.env. the variables for setting the remote and its branch are different ones, looking at it.

edwintorok:
If there are LCM branches for components, do we want to update xs-opam to point to the LCM branches of the component instead of a version number? Probably not worth it for Quebec, but for Stockholm which is more long lived.

psafont:
The problem is that strategy makes us create an LCM branch for all repositories right away. We can pin it to a version initially, and when the version needs to be changed pin it to a branch instead.

lippirk:
> If there are LCM branches for components, do we want to update xs-opam to point to the LCM branches of the component instead of a version number? Probably not worth it for Quebec, but for Stockholm which is more long lived.

i agree with @psafont that we should do it for branches only when necessary, since the vast majority of repos we will not change (unless specsavers happens soon)

psafont:
I'd really like to have a PR for the xs-opam branch before merging this one

lippirk:
> I'd really like to have a PR for the xs-opam branch before merging this one

done - i'll update this according to those changes

PR #4053, opened 2020-03-13
CA-336685 improve error for mem constraints violation

lindig:
To enable better error messages in XenCenter, introduce two
new errors for memory constraint violations and use them
in xapi_vm_memory_constraints.ml

* MEMORY_CONSTRAINT_VIOLATION_ORDER
* MEMORY_CONSTRAINT_VIOLATION_MAXPIN


ocaml/idl/datamodel_errors.ml
```diff
@@ -107,6 +107,12 @@ let _ =
     ~doc:"A VDI with the specified location already exists within the SR" ();
   error Api_errors.memory_constraint_violation ["constraint"]
     ~doc:"The dynamic memory range does not satisfy the following constraint." ();
+  error Api_errors.memory_constraint_violation_order ["constraint"]
+    ~doc:"The dynamic memory range violates static_min <= dynamic_min <=
```
edwintorok: Are you sure you want to have a newline in the string here? I think that'll translate into a newline in the error message which may not be what we want (I think XenCenter only displays a single line)

ocaml/idl/datamodel_errors.ml
```diff
@@ -107,6 +107,10 @@ let _ =
     ~doc:"A VDI with the specified location already exists within the SR" ();
   error Api_errors.memory_constraint_violation ["constraint"]
     ~doc:"The dynamic memory range does not satisfy the following constraint." ();
+  error Api_errors.memory_constraint_violation_order ["constraint"]
+    ~doc:"The dynamic memory range violates static_min <= dynamic_min <= dynamic_max <= static_max." ();
```
kc284: I would leave the term constraint in: `The dynamic memory range violates the constraint static_min <= dynamic_min <= dynamic_max <= static_max.` Same below.

ocaml/idl/datamodel_errors.ml
```diff
@@ -107,6 +107,10 @@ let _ =
     ~doc:"A VDI with the specified location already exists within the SR" ();
   error Api_errors.memory_constraint_violation ["constraint"]
     ~doc:"The dynamic memory range does not satisfy the following constraint." ();
+  error Api_errors.memory_constraint_violation_order ["constraint"]
```
robhoes: I think that the parameters need to be better defined. Right now, it is still a long string that includes the doc string below and (for maxpin) a reason appended to that string. It would be better if the error code + arguments are "machine readable", so that they can be easily internationalised etc. by a client such as XenCenter.

So, I think that the `_order` error does not need an argument, while the `maxpin` error needs a reason error that tells the client why this stricter constraint applies.

lindig:
I have refined the PR: wording (@kc284), and dropped the reason parameter for order constraints (@robhoes).

ocaml/xapi/xapi_vm_memory_constraints.ml
```diff
@@ -66,12 +66,12 @@ module Vm_memory_constraints : T = struct
   let assert_valid ~constraints =
     if not (are_valid ~constraints)
     then raise (Api_errors.Server_error (
-        Api_errors.memory_constraint_violation, [order_constraint]))
+        Api_errors.memory_constraint_violation_order, []))
 
   let assert_valid_and_pinned_at_static_max ~constraints ~reason =
     if not (are_valid_and_pinned_at_static_max ~constraints)
     then raise (Api_errors.Server_error (
-        Api_errors.memory_constraint_violation, [equality_constraint reason]))
```
robhoes: You can delete L58-65 now.

PR #4054, opened 2020-03-13
CP-32649: Use Stdlib's Result

psafont:
The library result is deprecated and it doesn't have the same interface as the standard library.

### DIFF
--- a/ocaml/xapi/xapi_ha_vm_failover.ml
+++ b/ocaml/xapi/xapi_ha_vm_failover.ml
@@ -15,27 +15,27 @@
 module D = Debug.Make(struct let name="xapi_ha_vm_failover" end)
 open D
 
-type task_result = (Rpc.t, exn) Result.result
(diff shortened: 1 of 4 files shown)


lippirk:
i expected to see the result library disappear from a dune file

psafont:
We use `Rresult`, since it imports `Result` there's not much we can do about that.

ocaml/xapi/xapi_ha_vm_failover.ml
```diff
@@ -15,27 +15,27 @@
 module D = Debug.Make(struct let name="xapi_ha_vm_failover" end)
 open D
 
-type task_result = (Rpc.t, exn) Result.result
+type task_result = (Rpc.t, exn) Stdlib.Result.t
```
edwintorok: `Stdlib.Result.t` is just `result`: if you have a compiler new enough to have stdlib it will be new enough to have result.

PR #4055, opened 2020-03-13
CA-336258: fix API forwarder for JSONRPC calls

robhoes:


robhoes:
Depends on https://github.com/xapi-project/xen-api-libs-transitional/pull/75

robhoes:
I'm not sure why there are two travis jobs (push and pr), but the pr one passed.

PR #4057, opened 2020-03-17
Enable to create a VM in Suspended mode with suspend_VDI set

benjamreis:
Signed-off-by: benjamreis <benjamin.reis@vates.fr>

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -1477,7 +1477,7 @@ end
 (* These are included in vbds and vifs -- abstracted here to keep both these uses consistent *)
 let device_status_fields =
   [
-    field ~ty:Bool ~qualifier:DynamicRO "currently_attached" "is the device currently attached (erased on reboot)";
+    field ~ty:Bool ~qualifier:StaticRO ~default_value:(Some (VBool false)) ~lifecycle:[Changed, rel_stockholm, "Become static to allow plugged VIF and VBD creation for Suspended VM"] "currently_attached" "is the device currently attached (erased on reboot)";
     field ~ty:Int ~qualifier:DynamicRO "status_code" "error/success code associated with last attach-operation (erased on reboot)";
     field ~ty:String ~qualifier:DynamicRO "status_detail" "error/success information associated with last attach-operation status (erased on reboot)";
     field ~ty:(Map(String, String)) ~qualifier:DynamicRO "runtime_properties" "Device runtime properties"
@@ -3004,7 +3004,7 @@ module VBD = struct
            field ~qualifier:StaticRO ~ty:(Ref _vm) "VM" "the virtual machine";
            field ~qualifier:StaticRO ~ty:(Ref _vdi) "VDI" "the virtual disk";
 
-           field ~qualifier:DynamicRO "device" "device seen by the guest e.g. hda1";
+           field ~qualifier:StaticRO ~ty:String ~default_value:(Some (VString "")) ~lifecycle:[Changed, rel_stockholm, "Become static to allow plugged VBD creation for Suspended VM"] "device" "device seen by the guest e.g. hda1";
            field "userdevice" "user-friendly device name e.g. 0,1,2,etc.";
            field ~ty:Bool "bootable" "true if this VBD is bootable";
            field ~qualifier:StaticRO ~ty:mode "mode" "the mode the VBD should be mounted with";
--- a/ocaml/idl/datamodel_vm.ml
+++ b/ocaml/idl/datamodel_vm.ml
@@ -1284,6 +1284,7 @@ let set_NVRAM_EFI_variables = call ~flags:[`Session]
                    ]
       ~lifecycle:[
         Published, rel_rio, "";
+        Changed, rel_stockholm, "possibility to create a VM in suspended mode with a suspend_VDI set";
       ]
       ~messages_default_allowed_roles:_R_VM_ADMIN
       ~messages:[ snapshot; snapshot_with_quiesce; clone; copy; revert; checkpoint;
@@ -1361,13 +1362,13 @@ let set_NVRAM_EFI_variables = call ~flags:[`Session]
       ~contents:
         ([ uid _vm;
          ] @ (allowed_and_current_operations operations) @ [
-           field ~writer_roles:_R_VM_OP ~qualifier:DynamicRO ~ty:power_state "power_state" "Current power state of the machine";
            namespace ~name:"name" ~contents:(names oss_since_303 RW) ();
+           field ~writer_roles:_R_VM_OP ~qualifier:StaticRO ~default_value:(Some (VEnum "Halted")) ~lifecycle:[Changed, rel_stockholm, "Become static to allow Suspended VM creation"] ~ty:power_state "power_state" "Current power state of the machine";
 
            field ~ty:Int "user_version" "Creators of VMs and templates may store version information here.";
            field ~effect:true ~ty:Bool "is_a_template" "true if this is a template. Template VMs can never be started, they are used only for cloning other VMs";
            field ~ty:Bool ~default_value:(Some (VBool false)) ~qualifier:DynamicRO ~writer_roles:_R_POOL_ADMIN ~lifecycle:[Published, rel_falcon, "Identifies default templates"] "is_default_template" "true if this is a default template. Default template VMs can never be started or migrated, they are used only for cloning other VMs";
-           field ~qualifier:DynamicRO ~ty:(Ref _vdi) "suspend_VDI" "The VDI that a suspend image is stored on. (Only has meaning if VM is currently suspended)";
+           field ~qualifier:StaticRO ~default_value:(Some (VRef null_ref)) ~lifecycle:[Changed, rel_stockholm, "Become static to allow Suspended VM creation"] ~ty:(Ref _vdi) "suspend_VDI" "The VDI that a suspend image is stored on. (Only has meaning if VM is currently suspended)";
 
            field ~writer_roles:_R_VM_POWER_ADMIN ~qualifier:DynamicRO ~ty:(Ref _host) "resident_on" "the host the VM is currently resident on";
            field ~writer_roles:_R_VM_POWER_ADMIN ~in_oss_since:None ~qualifier:DynamicRO ~default_value:(Some (VRef null_ref)) ~ty:(Ref _host) "scheduled_to_be_resident_on" "the host on which the VM is due to be started/resumed/migrated. This acts as a memory reservation indicator";
@@ -1395,14 +1396,14 @@ let set_NVRAM_EFI_variables = call ~flags:[`Session]
            field  ~ty:(Map(String, String)) "other_config" "additional configuration" ~map_keys_roles:["pci", _R_POOL_ADMIN; ("folder",(_R_VM_OP));("XenCenter.CustomFields.*",(_R_VM_OP))];
            field ~qualifier:DynamicRO ~ty:Int "domid" "domain ID (if available, -1 otherwise)";
            field ~qualifier:DynamicRO ~in_oss_since:None ~ty:String "domarch" "Domain architecture (if available, null string otherwise)";
-           field ~in_oss_since:None ~qualifier:DynamicRO ~ty:(Map(String, String)) "last_boot_CPU_flags" "describes the CPU flags on which the VM was last booted";
+           field ~in_oss_since:None ~qualifier:StaticRO ~ty:(Map(String, String)) ~default_value:(Some (VMap [])) "last_boot_CPU_flags" "describes the CPU flags on which the VM was last booted";
            field ~qualifier:DynamicRO ~ty:Bool "is_control_domain" "true if this is a control domain (domain 0 or a driver domain)";
            field ~qualifier:DynamicRO ~ty:(Ref _vm_metrics) "metrics" "metrics associated with this VM";
            field ~qualifier:DynamicRO ~ty:(Ref _vm_guest_metrics) "guest_metrics" "metrics associated with the running guest";
            (* This was an internal field in Rio, Miami beta1, Miami beta2 but is now exposed so that
               	   it will be included automatically in Miami GA exports and can be restored, important if
               	   the VM is in a suspended state *)
-           field ~in_oss_since:None ~internal_only:false ~in_product_since:rel_miami ~qualifier:DynamicRO ~ty:String "last_booted_record" "marshalled value containing VM record at time of last boot, updated dynamically to reflect the runtime state of the domain" ~default_value:(Some (VString ""));
+           field ~in_oss_since:None ~internal_only:false ~in_product_since:rel_miami ~qualifier:StaticRO ~ty:String "last_booted_record" "marshalled value containing VM record at time of last boot, updated dynamically to reflect the runtime state of the domain" ~default_value:(Some (VString ""));
            field ~in_oss_since:None ~ty:String "recommendations" "An XML specification of recommended values and ranges for properties of this VM";
            field ~effect:true ~in_oss_since:None ~ty:(Map(String, String)) ~in_product_since:rel_miami ~qualifier:RW "xenstore_data" "data to be inserted into the xenstore tree (/local/domain/<domid>/vm-data) after the VM is created." ~default_value:(Some (VMap []));
            field ~writer_roles:_R_POOL_OP ~in_oss_since:None ~ty:Bool ~in_product_since:rel_orlando ~internal_deprecated_since:rel_boston ~qualifier:StaticRO "ha_always_run" "if true then the system will attempt to keep the VM running as much as possible." ~default_value:(Some (VBool false));
--- a/ocaml/xapi/xapi_vbd.ml
+++ b/ocaml/xapi/xapi_vbd.ml
@@ -132,8 +132,17 @@ let unplug_force_no_safety_check = unplug_force
 let autodetect_mutex = Mutex.create ()
 
 (** VBD.create doesn't require any interaction with xen *)
-let create  ~__context ~vM ~vDI ~userdevice ~bootable ~mode ~_type ~unpluggable ~empty
-    ~other_config ~qos_algorithm_type ~qos_algorithm_params =
+let create  ~__context ~vM ~vDI ~device ~userdevice ~bootable ~mode ~_type ~unpluggable ~empty
+    ~other_config ~currently_attached ~qos_algorithm_type ~qos_algorithm_params =
+
+  (* TODO: Raise bad power state error (once all API clients make sure to onlu call the needed params in the create method) when:
+    - power_state == `Halted and (device <> "" || currently_attached)
+  *)
+
+  let power_state = Db.VM.get_power_state ~__context ~self:vM in
+  let suspended = (power_state = `Suspended) in
+  let _device = if suspended then device else "" in
+  let _currently_attached = if suspended then currently_attached else false in
 
   if not empty then begin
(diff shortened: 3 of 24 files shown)


benjamreis:
This patch enables to create a VM in `Suspended` state with a `suspend_VDI` set
This add to the `vm.create` method:
- `power_state` with default value: `Halted`
- `suspend_VDI` with default value: `Ref.null`
- `last_booted_record` with default value: `""`
- `last_boot_CPU_flags` with default value: `[]`
The default behavior is not changed.

ocaml/xapi/xapi_vm.ml
```diff
@@ -505,6 +507,17 @@ let create ~__context ~name_label ~name_description
...
+  if power_state = `Halted && suspend_VDI != Ref.null then begin
+    error "No suspend_VDI should be provided if VM created in `Halted state";
+  end
+  else if power_state = `Suspended && suspend_VDI = Ref.null then begin
+    error "VM created in `Suspended state needs a suspend_VDI";
+  end
+  else begin
+    error "Bad power state for VM creation";
+  end;
+
```
benjamreis: I'd like some feedback here, I'm not sure what to do in this error cases, for now there's only log but I'm willing to change it accordingly to your comments.

Thanks

ocaml/xapi/xapi_vm.ml
```diff
@@ -505,6 +507,17 @@ let create ~__context ~name_label ~name_description
     ~nVRAM
   : API.ref_VM =
 
+  (* TODO: Error handling, but how? raise? or return null ref? or put default values? *)
+  if power_state = `Halted && suspend_VDI != Ref.null then begin
```
lippirk: you probably want structural inequality (`<>`) rather than physical inequality (`!=`)

edwintorok:
You must've had this VM running somewhere else and suspended it to end up with a suspend VDI.
If you now create a new VM object with same suspend VDI then the VM will resume with a different UUID, which may have unforeseen consequences. (Unlike a migration or suspend/resume where the VM will end up with same UUID as it had when it started).

What is the use case here? Can you use built-in hibernation support in various OSes to achieve the same outcome, but without having to use the suspend VDI?

ocaml/xapi/xapi_vm.ml
benjamreis: Indeed!

ocaml/xapi/xapi_vm.ml
lippirk: you can throw a VM bad power state error - https://github.com/xapi-project/xen-api/blob/cdcc916a3ed5fd3040d1bdcc33bf4faa46292f48/ocaml/idl/datamodel_vm.ml#L353

you will need to make sure you add a 'changed' to the datamodel for this function

robhoes:
Besides what @edwintorok wrote, since you have an empty `last_booted_record`, the VM generally won't have right runtime state in xenopsd.

Note that you can already export and import suspended VMs. Does that help you?

ocaml/xapi/xapi_vm.ml
benjamreis: Thanks for the feedback, will do!

olivierlambert:
@edwintorok Let me explain the context. The goal is to able to backup a VM with its RAM, to be certain it will stay coherent despite VSS agent disappearance. This will also allow VM replication/copy with its RAM to another host.

@robhoes This is coupled with our existing Delta backup/Continuous Replication feature in XO (used in production since 2015 IIRC, for snapshots). We first tried with VM metadata but it was almost every time catastrophic (import broken, missing fields, impossible to restore on older hosts). Since we switched

Process is like this:

1. Creating a VM check point
2. Export the VM object, VHD and Suspend VDI somewhere
3. Restore: recreate the VM object, import VHD and suspend VDI. Then start it.

Note that we already test on our own PoC and it worked well. If you have actual example of problems with UUID change on restore, I'd like to know more.

benjamreis:
To further what @olivierlambert  said:

The VM created suspended would by a "clone" of the snapshoted VM, so it will have the same `last_booted_record` and `last_boot_CPU_flags` the default values are only used to maintain the default behavious when a VM is created `Halted`.

edwintorok:
So perhaps add a check that when creating a suspended VM the last_* records must not be empty

benjamreis:
Hi all,

I've got a little issue with my latest (uncommited) changes: I've added a block a the begining of create that looks like this:
```ocaml
  (* `ignore` is necessary here or the compilation will break because this method doesn't return in this error cases *)
  ignore(
    if power_state = `Halted then
      if suspend_VDI <> Ref.null then begin
        raise (Api_errors.Server_error (Api_errors.vm_bad_power_state, ["No suspend_VDI should be provided if VM created in `Halted state"]))
      end
    else if power_state = `Suspended then
      if suspend_VDI = Ref.null then begin
        raise (Api_errors.Server_error (Api_errors.vm_bad_power_state, ["VM created in `Suspended state needs a suspend_VDI"]))
      end;
      if last_booted_record = "" then begin
        raise (Api_errors.Server_error (Api_errors.vm_bad_power_state, ["VM created in `Suspended state needs non empty last_booted_record"]))
      end;
      if last_boot_CPU_flags = [] then begin
        raise (Api_errors.Server_error (Api_errors.vm_bad_power_state, ["VM created in `Suspended state needs non empty last_boot_CPU_flags"]))
      end
    else begin
      raise (Api_errors.Server_error (Api_errors.vm_bad_power_state, ["Bad power state for VM creation "; Record_util.power_to_string power_state; " should be `Halted or `Suspended"]))
    end
  );
```

It throws an exception when:
- `power_state = `Halted` and a `suspend_VDI` is set
- `power_state = `Suspended` and no `suspend_VDI` is set
- `power_state = `Suspended` and no `last_booted_record` is set
- `power_state = `Suspended` and no `last_boot_CPU_flags` is set
- `power_state` is not `Halted` nor `Suspended`

However with this, the xapi test fails (it didn't before this specific changes), after investigating I found out that the test created Suspended VM with no `last_booted_record` nor `last_boot_CPU_flags` an I can't find where.

I thought I modified every call to VM.create to put `Halted` as `power_state` but it seems I missed one and I can't find it.

Could help me?

lippirk:
> Hi all,
> 
> I've got a little issue with my latest (uncommited) changes: I've added a block a the begining of create that looks like this:
> 
> ```ocaml
>   (* `ignore` is necessary here or the compilation will break because this method doesn't return in this error cases *)
>   ignore(
>     if power_state = `Halted then
>       if suspend_VDI <> Ref.null then begin
>         raise (Api_errors.Server_error (Api_errors.vm_bad_power_state, ["No suspend_VDI should be provided if VM created in `Halted state"]))
>       end
>     else if power_state = `Suspended then
>       if suspend_VDI = Ref.null then begin
>         raise (Api_errors.Server_error (Api_errors.vm_bad_power_state, ["VM created in `Suspended state needs a suspend_VDI"]))
>       end;
>       if last_booted_record = "" then begin
>         raise (Api_errors.Server_error (Api_errors.vm_bad_power_state, ["VM created in `Suspended state needs non empty last_booted_record"]))
>       end;
>       if last_boot_CPU_flags = [] then begin
>         raise (Api_errors.Server_error (Api_errors.vm_bad_power_state, ["VM created in `Suspended state needs non empty last_boot_CPU_flags"]))
>       end
>     else begin
>       raise (Api_errors.Server_error (Api_errors.vm_bad_power_state, ["Bad power state for VM creation "; Record_util.power_to_string power_state; " should be `Halted or `Suspended"]))
>     end
>   );
> ```
> 
> It throws an exception when:
> 
>     * `power_state = `Halted`and a`suspend_VDI` is set
> 
>     * `power_state = `Suspended`and no`suspend_VDI` is set
> 
>     * `power_state = `Suspended`and no`last_booted_record` is set
> 
>     * `power_state = `Suspended`and no`last_boot_CPU_flags` is set
> 
>     * `power_state` is not `Halted` nor `Suspended`
> 
> 
> However with this, the xapi test fails (it didn't before this specific changes), after investigating I found out that the test created Suspended VM with no `last_booted_record` nor `last_boot_CPU_flags` an I can't find where.
> 
> I thought I modified every call to VM.create to put `Halted` as `power_state` but it seems I missed one and I can't find it.
> 
> Could you help me?

firstly i would use a match statement to make this a bit more readable

```
match power_state with
| `Halted when suspend_VDI <> Ref.null -> throw
| `Suspended when suspend_VDI <> Ref.null -> throw 
... etc.
```

it might be helpful if you copy the error that you get? (i assume you're running `make test`)

benjamreis:
Thanks for the quick reply!

Indeed a match statement seems more readable, will do.

As for the test yes I run `make test` and I got 148 failed tests (0 test failed before I added the exceptions).

I've got these at the end of every failed tests:
```
[exception] Server_error(VM_BAD_POWER_STATE, [ VM created in `Suspended state needs non empty last_booted_record ])
Raised at file "ocaml/xapi/xapi_vm.ml", line 566, characters 38-196
Called from file "ocaml/tests/test_common.ml", line 132, characters 2-918
```

But I give Halted to the create method by default: https://github.com/xapi-project/xen-api/pull/4057/files#diff-3b56a3e32a3ef8e749aa75009fc29b19R131 and `Suspended is given nowhere.`

lippirk:
the problem is that `make_vm` takes `power_state` as an optional parameter, so it can be overridden by a caller - i.e. you will need to check all the places where `make_vm` gets called

benjamreis:
Good news: the test are now succesfull!
Going from `if then else` block within `if then else` block to a `match` solved the problem.

I think there was a mistake in the imbrication of my `if`...

Anyway now it works and I raise when the create params are incoherent.

ocaml/xapi/xapi_vm.ml
```diff
@@ -551,6 +553,15 @@ let create ~__context ~name_label ~name_description
     ~nVRAM
   : API.ref_VM =
 
+  (* `ignore` is necessary here or the compilation will break because this method doesn't return in this error cases *)
```
lindig: Small simplifications:
* ` ignore @@ match ...`
*  `raise Api_errors.(Server_error (vm_bad_power_state, …`
* please avoid overlong lines - they are hard to review

ocaml/idl/datamodel_vm.ml
```diff
@@ -1361,13 +1362,13 @@ let set_NVRAM_EFI_variables = call ~flags:[`Session]
...
+           field ~writer_roles:_R_VM_OP ~qualifier:StaticRO ~default_value:(Some (VEnum "Halted")) ~ty:power_state "power_state" "Current power state of the machine";
 
            field ~ty:Int "user_version" "Creators of VMs and templates may store version information here.";
            field ~effect:true ~ty:Bool "is_a_template" "true if this is a template. Template VMs can never be started, they are used only for cloning other VMs";
            field ~ty:Bool ~default_value:(Some (VBool false)) ~qualifier:DynamicRO ~writer_roles:_R_POOL_ADMIN ~lifecycle:[Published, rel_falcon, "Identifies default templates"] "is_default_template" "true if this is a default template. Default template VMs can never be started or migrated, they are used only for cloning other VMs";
-           field ~qualifier:DynamicRO ~ty:(Ref _vdi) "suspend_VDI" "The VDI that a suspend image is stored on. (Only has meaning if VM is currently suspended)";
+           field ~qualifier:StaticRO ~default_value:(Some (VRef (null_ref))) ~ty:(Ref _vdi) "suspend_VDI" "The VDI that a suspend image is stored on. (Only has meaning if VM is currently suspended)";
```
lindig:
```
(Some (VRef null_ref))
```

ocaml/xapi/xapi_vm.ml
benjamreis: Done! I split the `raise` lines to be more readable. :)

ocaml/idl/datamodel_vm.ml
benjamreis: Done!

benjamreis:
Latest commit add the following modifications:
- add device and currently_attached to the VBD.create method to allow the creation of an already plugged VBD. the new fields can only be used if the VM is suspened (else it throws). Default behaviour is kept.
- add currently_attached to the VIF.create method to allow the creation of an already plugged VIF. the new fields can only be used if the VM is suspened (else it throws). Default behaviour is kept.

The VM created Suspended needs to have already plugged VIFs and VBDs

ocaml/idl/datamodel.ml
```diff
@@ -1466,7 +1466,7 @@ end
 (* These are included in vbds and vifs -- abstracted here to keep both these uses consistent *)
 let device_status_fields =
   [
-    field ~ty:Bool ~qualifier:DynamicRO "currently_attached" "is the device currently attached (erased on reboot)";
+    field ~ty:Bool ~qualifier:StaticRO ~default_value:(Some (VBool false)) "currently_attached" "is the device currently attached (erased on reboot)";
```
psafont: please a `~lifecycle:[Changed, rel_stockholm, "explanation"]` parameter in all the changed `field`s.

benjamreis: Done! :)

benjamreis:
I'e added the default values for 
- `VM`: `power_state` and `suspend_VDI`
- `VBD`: `device` and `currently_attached`
- `VIF`: `currently_attached`

in the `create_from_record` to keep the same behaviour.

benjamreis:
I'm done with dev and testing so any feedback is welcomed :)

ocaml/xapi/xapi_vbd.ml
```diff
@@ -132,8 +132,16 @@ let unplug_force_no_safety_check = unplug_force
...
-let create  ~__context ~vM ~vDI ~userdevice ~bootable ~mode ~_type ~unpluggable ~empty
-    ~other_config ~qos_algorithm_type ~qos_algorithm_params =
+let create  ~__context ~vM ~vDI ~device ~userdevice ~bootable ~mode ~_type ~unpluggable ~empty
+    ~other_config ~currently_attached ~qos_algorithm_type ~qos_algorithm_params =
+
+  if device <> "" || currently_attached then begin
+    match (Db.VM.get_power_state ~__context ~self:vM) with
+      | `Suspended -> ()
+      | _ -> raise (Api_errors.(Server_error (
+          vm_bad_power_state, ["Plugged VBD creation only allowed for suspended VM"])
```
edwintorok, line 142: You should also say which VM the error is for. This is not yet enforced by the type checker, but see datamodel_errors.ml you need 3 arguments here, not 1: `error Api_errors.vm_bad_power_state ["vm"; "expected"; "actual"]`
See examples elsewhere in the code:
`Api_errors.vm_bad_power_state [ Ref.string_of vm; expected; actual `

ocaml/xapi/xapi_vif.ml
```diff
@@ -34,8 +34,17 @@ let unplug_force ~__context ~self =
...
 let create  ~__context ~device ~network ~vM
-    ~mAC ~mTU ~other_config ~qos_algorithm_type ~qos_algorithm_params ~locking_mode ~ipv4_allowed ~ipv6_allowed : API.ref_VIF =
-  create ~__context ~device ~network ~vM ~currently_attached:false
+    ~mAC ~mTU ~other_config  ~currently_attached ~qos_algorithm_type ~qos_algorithm_params ~locking_mode ~ipv4_allowed ~ipv6_allowed : API.ref_VIF =
+
+  if currently_attached then begin
+    match (Db.VM.get_power_state ~__context ~self:vM) with
+      | `Suspended -> ()
+      | _ -> raise (Api_errors.(Server_error (
+          vm_bad_power_state, ["Plugged VIF creation only allowed for suspended VM"])
```
edwintorok, line 43: as above, which VM?

ocaml/xapi/xapi_vm.ml
```diff
@@ -551,6 +553,22 @@ let create ~__context ~name_label ~name_description
     ~nVRAM
   : API.ref_VM =
 
+  begin match power_state with
+    | `Halted when suspend_VDI <> Ref.null ->
+      raise (Api_errors.(Server_error (vm_bad_power_state, ["No suspend_VDI should be provided if VM created in `Halted state"])))
+    | `Suspended when suspend_VDI = Ref.null || last_booted_record = "" || last_boot_CPU_flags = [] ->
+      raise (Api_errors.(Server_error (
+        vm_bad_power_state,
+        ["VM created in `Suspended state needs a suspend_VDI, non empty last_booted_record and non empty last_boot_CPU_flags"])
```
edwintorok, line 562: For which VM?

ocaml/xapi/xapi_vm.ml
```diff
@@ -551,6 +553,22 @@ let create ~__context ~name_label ~name_description
...
+      ))
+    | `Running | `Paused ->
+      raise (Api_errors.(Server_error (
+        vm_bad_power_state,
+        ["Bad power state for VM creation "; Record_util.power_to_string power_state; " should be `Halted or `Suspended"])
```
edwintorok, line 567: which VM?

ocaml/xapi/xapi_vbd.ml
benjamreis: Done! :)

ocaml/xapi/xapi_vif.ml
benjamreis: Done

ocaml/xapi/xapi_vm.ml
benjamreis: Done

ocaml/xapi/xapi_vm.ml
benjamreis: Done

olivierlambert:
Hi guys, any news for this merge?

benjamreis:
PR is up to date with master. :)

BTW this patch has been included in the 8.1 version of XCP-ng, some of our user reports that using XenDesktop of XCP-ng host triggers the new power state error I added.

It seems that the optional arguments I added for VM/VBD/VIF creation are filled by XenDesktop (I guess from an object record used for the creation) so I thinkk when this is merged XenDesktop shouldn't use this argument and let the default value be used to maintain the same behavior. :)

Regards!

benjamreis:
Last commit remove the throws in case the new params of VM/VBD/VIF create methods are misused.
Instead of throwing, I use the default values when `power_state <> `Halted`

This will fix the break of XenDesktop calls to the API.

benjamreis:
i put back an error raising in case the VM is created `Suspended` but have no `suspend_VDI` or `last_boot_CPU_flags` or `last_booted_record`.

This is good to be reviewed again. :)

benjamreis:
Checked the logs of CI and the failure seems unrelated to my changes.
Build and tests are ok on my side.

lindig:
We have some interdependent merges. Coul you try running Travis again?

lindig:
Is this meant to go into the Stockholm release? Please prefix commits with a ticket numner.

benjamreis:
We'd like to merge is the soonest we can yes. IDK about ticket number though can you elaborate or redirect me to a doc?

Thanks

lindig:
Sorry - I mistook this for a different series of commits. There is no ticket number but it would be good to group the commits by giving them a common prefix but it's not absolutely required. This is unlikely to make it into the next release (because we've done all the testing) but I expect it to be merged once we branched off.

benjamreis:
What I can do is squash all commit into one once, the PR is approved (it makes sense since all of them are semantically linked together). What do you think?
And if its not possible for this release then the the next one.

benjamreis:
Great! I'll let you squash before the merge then. :)

lippirk:
@benjamreis personally I would like to see a more detailed commit message (and of course that it is squashed into a single / sensible number of commits). I think as the patch author you are best placed to do this

PR #4058, opened 2020-03-19
improve invalid VIF map error message for intra-pool migration

lippirk:
During an intra-pool migration, the VM's VIFs must be on the same network, so make that explicit in the error message


ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -1255,7 +1255,7 @@ let assert_can_migrate  ~__context ~vm ~dest ~live ~vdi_map ~vif_map ~options ~v
       Xapi_vm_helpers.assert_can_boot_here ~__context ~self:vm ~host:remote.dest_host ~snapshot ~do_sr_check:false ~do_cpuid_check ();
       if vif_map <> [] then
         raise (Api_errors.Server_error(Api_errors.operation_not_allowed, [
-            "VIF mapping is not allowed for intra-pool migration"]));
+          "VIF mapping is not allowed for intra-pool migration"; "all VIFs must be on the same network"]));
```
robhoes: The API error is defined as follows in datamodel_errors.ml:

```
  error Api_errors.operation_not_allowed ["reason"]
    ~doc:"You attempted an operation that was not allowed." ();
```
So it must have a single argument for the reason, not two.

lippirk: ok, i'll fix that

PR #4061, opened 2020-03-20
Add emergency call to reset TLS server certificates

psafont:
If successful it deletes the current certificate and places a new,
self-signed on instead. It also deletes the database record of the
server certificate currently installed in the host.
This call is only available locally only.

Its implementation is barebones as it's replicating what happens when gencert is called on first startup.

This is meant to change with https://github.com/xapi-project/xen-api/tree/feature/REQ-811/master and after both features get merged the bash script will be replaced by an ocaml binary, as well as xapi making an effort to add the existing self-signed cert to the database.

### DIFF
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -958,7 +958,6 @@ let host_query_ha = call ~flags:[`Session]
       ()
 
   let install_server_certificate = call
-      ~in_oss_since:None
       ~lifecycle:[Published, rel_stockholm, ""]
       ~name:"install_server_certificate"
       ~doc:"Install the TLS server certificate."
@@ -977,6 +976,15 @@ let host_query_ha = call ~flags:[`Session]
          ; param_release=stockholm_release; param_default=Some (VString "")}
         ]
       ~allowed_roles:_R_POOL_ADMIN
+      ()
+
+  let emergency_reset_server_certificate = call
+      ~flags:[`Session]
+      ~lifecycle:[Published, rel_stockholm, ""]
+      ~name:"emergency_reset_server_certificate"
+      ~doc:"Delete the current TLS server certificate and replace by a new, self-signed one. This should only be used with extreme care."
+      ~versioned_params: []
+      ~allowed_roles:_R_LOCAL_ROOT_ONLY
       ()
 
   let get_server_certificate = call
@@ -1402,6 +1410,7 @@ let host_query_ha = call ~flags:[`Session]
         certificate_sync;
         get_server_certificate;
         install_server_certificate;
+        emergency_reset_server_certificate;
         update_pool_secret;
         update_master;
         attach_static_vdis;
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -1133,14 +1133,8 @@ let install_server_certificate ~__context ~host ~certificate ~private_key ~certi
   if Db.Pool.get_ha_enabled ~__context ~self:(Helpers.get_pool ~__context) then
     raise Api_errors.(Server_error (ha_is_enabled, []));
 
-  let current_server_certs = Db.Certificate.get_all_records ~__context
-  |> List.filter_map (fun (reference, record) ->
-    if record.API.certificate_host = host then
-        Some reference
-    else
-        None
-    )
-  in
+  let expr = Db_filter_types.(Eq (Field "host", Literal (Ref.string_of host))) in
+  let current_server_certs = Db.Certificate.get_refs_where ~__context ~expr in
 
   let pem_chain = match certificate_chain with
   | ""        -> None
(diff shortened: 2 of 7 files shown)


ocaml/idl/datamodel_host.ml
```diff
@@ -979,6 +979,16 @@ let host_query_ha = call ~flags:[`Session]
       ~allowed_roles:_R_POOL_ADMIN
       ()
 
+  let emergency_server_certificate_reset = call
+      ~flags:[`Session]
+      ~in_oss_since:None
+      ~lifecycle:[Published, rel_stockholm, ""]
+      ~name:"emergency_server_certificate_reset"
```
robhoes: I'd say `emergency_reset_server_certificate`, for consistency with `install_server_certificate` and `get_server_certificate`. Although I know that some of the other emergency commands have the verb at the end... :(

ocaml/idl/datamodel_host.ml
```diff
@@ -979,6 +979,16 @@ let host_query_ha = call ~flags:[`Session]
       ~allowed_roles:_R_POOL_ADMIN
       ()
 
+  let emergency_server_certificate_reset = call
+      ~flags:[`Session]
+      ~in_oss_since:None
```
robhoes: You don't actually need this if you have `~lifecycle`.

ocaml/xapi/xapi_host.ml
```diff
@@ -1171,6 +1171,24 @@ let install_server_certificate ~__context ~host ~certificate ~private_key ~certi
   (* sever all connections done with stunnel *)
   Xapi_mgmt_iface.reconfigure_stunnel ~__context
 
+let emergency_server_certificate_reset ~__context =
+  let generate_ssl_cert = "/opt/xensource/libexec/generate_ssl_cert" in
```
robhoes: This of course need to get defined in xapi_globs.ml, but I think that @lippirk has done that already on his feature branch.

ocaml/xapi/xapi_host.ml
```diff
@@ -1171,6 +1171,24 @@ let install_server_certificate ~__context ~host ~certificate ~private_key ~certi
...
+
+  let args = [!Xapi_globs.xapissl_path; Helper_hostname.get_hostname ()] in
+  ignore @@ Forkhelpers.execute_command_get_output generate_ssl_cert args;
+
+  let self = Helpers.get_localhost ~__context in
+
+  (* Delete records of the server certificate in this host *)
+  Db.Certificate.get_all_records ~__context
```
robhoes: We should assume the worst-case situation and the main reason for having this command, which is to fix things up when your on a pool slave, the current certificate is bad, and because of that the master connection is broken. That means no DB calls are possible until this is fixed. So perhaps we need to reset stunnel first?

ocaml/xapi/xapi_host.ml
```diff
@@ -1171,6 +1171,24 @@ let install_server_certificate ~__context ~host ~certificate ~private_key ~certi
...
+  |> List.filter_map (fun (reference, record) ->
```
robhoes: For this kind of operations, take a look at `get_refs_where`. It's a little awkward, but more efficient, since the filtering happens on the master. E.g.

    let expr = Db_filter_types.(Eq (Field "host", Literal (Ref.string_of self))) in
    Db.Certificate.get_refs_where ~__context ~expr

ocaml/xapi/xapi_host.ml
psafont: I would like to avoid the case where the certificate metadata does not match the certificate that's actually installed. I would rather try to do the database calls and simply ignore the errors than not doing the calls at all.

ocaml/xapi/xapi_host.ml
psafont: The intention is to remove the bash script altogether. Not sure if the merge between the branches will come before or after I get around to replace it.

ocaml/idl/datamodel_host.ml
psafont: 😵

ocaml/xapi/xapi_host.ml
psafont: I forgot to reset stunnel :scream:

ocaml/idl/datamodel_host.ml
psafont: should I change the CLI name as well?

robhoes: It's better to make it consistent...

psafont: I think having inconsistent names would be consistent as well ;p

psafont:
travis failure is due to the removal of Xcp_coverage, let me rebase the branches.

psafont:
travis is running at https://travis-ci.org/github/xapi-project/xen-api/jobs/665869144

PR #4062, opened 2020-03-22
Fix build: drop Xcp_coverage

edwintorok:
This has been removed from xcp-idl and causing the build to fail now.
Also simplify the dune files and drop unused coverage from there too.


psafont:
It's so strange that we were enabling coverage collection from tests

PR #4063, opened 2020-03-24
CA-336735: preserve formatting of stars for wlb recommendations

psafont:
Interfaces such as xe were exposing this string, with the conversion to
and from float the format was changed. With this make an extra effort to
preserve the formatting for the usual range of values.

Values lower than 0.0001 will still show up with the exponential format:
9.99e-5

### DIFF
--- a/ocaml/xapi/xapi_vm.ml
+++ b/ocaml/xapi/xapi_vm.ml
@@ -60,9 +60,17 @@ let assert_can_boot_here ~__context ~self ~host =
   assert_can_boot_here ~__context ~self ~host ~snapshot ~do_cpuid_check:true ()
 
 let retrieve_wlb_recommendations ~__context ~vm =
+  let compat_string_of_float f =
+    let repr = string_of_float f in
+    (* Original format has always at least one fractional decimal digit *)
+    if Astring.String.is_suffix ~affix:"." repr then
+      repr ^ "0"
+    else
+      repr
+  in
   let transform_for_api = function
     | host, Workload_balancing.(Recommendation {source; id; score}) ->
-      host, [source; (string_of_float score); id]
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_vm.ml
```diff
@@ -60,6 +60,14 @@ let assert_can_boot_here ~__context ~self ~host =
   assert_can_boot_here ~__context ~self ~host ~snapshot ~do_cpuid_check:true ()
 
 let retrieve_wlb_recommendations ~__context ~vm =
+  let string_of_float f =
```
edwintorok: Can you call this `compat_string_of_float`, that makes it clearer when reading the place it gets called that this is not the standard one.

ocaml/xapi/xapi_vm.ml
```diff
@@ -60,9 +60,17 @@ let assert_can_boot_here ~__context ~self ~host =
   assert_can_boot_here ~__context ~self ~host ~snapshot ~do_cpuid_check:true ()
 
 let retrieve_wlb_recommendations ~__context ~vm =
+  let compat_string_of_float f =
+    let repr = string_of_float f in
+    (* Original format has always at least one fractional decimal digit *)
+    if Astring.String.is_suffix ~affix:"." repr then
```
lindig: I assume performance is not an issue. Otherwise: `str.[String.length str - 1] = '.'` could be a cheaper predicate. In general, Atring is good about performance so I would not worry about this.

psafont: I read astring's source code around comparisons and tends to do the minimal amount of work needed, I don't think we need to worry that much about a function that's rarely called.

PR #4064, opened 2020-03-27
xapissl: re-read MANAGEMENT_INTERFACE in each retry

cheese:
If the old MANAGEMENT_INTERFACE is not configured, e.g. cable unplugged,
    xapissl will loop forever, even MANAGEMENT_INTERFACE is re-set to another
    interface during the loop.


lippirk:
hi @cheese, we have a feature branch on which we have removed this bash script

this infinite loop should no longer occur on this branch (if you're interested the code that will be used to get the management IP is here ocaml/gencert/lib.ml on branch feature/REQ-811/master). we are going to merge this shortly

PR #4066, opened 2020-03-27
REQ-453 re-expose reconfigure_stunnel

lippirk:
REQ-811 removed the ability for arbitrary code inside xapi to restart
stunnel, but this is required for REQ-453, so we reintroduce it.

### DIFF
--- a/ocaml/xapi/xapi_mgmt_iface.ml
+++ b/ocaml/xapi/xapi_mgmt_iface.ml
@@ -30,15 +30,31 @@ let listening_all = ref false
 let listening_localhost = ref false
 let listening_himn = ref false
 let management_m = Mutex.create ()
+let stunnel_accept_m = Mutex.create ()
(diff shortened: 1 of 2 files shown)


lippirk:
@psafont i'm going to test this - to double check, is this definitely what you need for REQ-453?

psafont:
Yes, the function is used by the other feature branch.

Does this implementation work now that the stunnel service is being driven by systemd unit files?

lippirk:
> Yes, the function is used by the other feature branch.
> 
> Does this implementation work now that the stunnel service is being driven by systemd unit files?

i added an fake xe endpoint to call `reconfigure_stunnel` and it works as expected :)

lippirk:
> It's ok if reconfigure doesn't restart stunnel as it means that stunnel has not been configured yet.

i'm not super happy with this, but that is the behaviour on `master`

PR #4067, opened 2020-03-31
CA-337087 migration tasks not cancelling correctly

lippirk:
Migrating a VM from a master would not cancel as expected

Also fix a potential race condition


lippirk:
I'd like to test this a bit more before merging.

I'm not too happy about how the forwarding logic is based on master/slave, and this also requires an is_master check

ocaml/xapi/helpers.ml
```diff
@@ -1242,10 +1242,23 @@ end = struct (* can't place these functions in task helpers due to circular depe
   let wait_for_ ~__context ~tasks ~propagate_cancel cb =
     let our_task = Context.get_task_id __context in
     let classes = List.map (fun x -> Printf.sprintf "task/%s" (Ref.string_of x)) (our_task::tasks) in
+    let maybe_cancel_children () =
+      let tasks_str = List.map Ref.short_string_of tasks |> String.concat "; " in
```
psafont: This can be also be formatted with the Fmt module: `let tasks_str = Fmt.(str "%a" Dump.(list string)) (Ref.short_string_of tasks) in`; it will surround each element with double quotes.

robhoes:
Can't we use `do_op_on`? That's what message_forwarding normally uses to decide between `Local` and `Client`.

lippirk:
> Can't we use `do_op_on`? That's what message_forwarding normally uses to decide between `Local` and `Client`.

i think the secret is to keep all of the forwarding inside `message_forwarding.ml` ;)

ocaml/xapi/message_forwarding.ml
```diff
@@ -1694,7 +1694,20 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
             fun ~local_fn ~__context ~vm op ->
               finally
                 (fun () -> allocate_vm_to_host ~__context ~vm ~host ~snapshot ~host_op:`vm_migrate ();
-                           forward_vm_op ~local_fn ~__context ~vm op)
+                           let internal_async_op ~__context ~wait =
```
robhoes, line 1697: I've been looking at this patch for a while now, and have to admit that I don't really follow what's going on anymore.

Part of the problem is that the actual API calls that are being executed (`{Local;Client;Client.InternalAsync}.VM.migrate_send`) are at three different places inside the function `migrate_send`. Furthermore, we have `forward_vm_op`, which inside it calls `do_op_on`, but there is also a call to `do_op_on_no_localhost` used for just the internal-async call.

The structure of the function was already quite complicated, because it includes three cases: non-live migration, live intra-pool migration, and live cross-pool migration. I believe that you have now "treated" just the live intra-pool case with internal-async logic, while at least the cross-pool case would also benefit from it.

Unfortunately, we have ended up with a mix of forwarding logic and different API calls that is hard to reason about.

lippirk: have you tried diffing against the commit before the original patch? that might make it slightly easier to see what's going on

i will try to make it more explicit which migration case we're in, perhaps that would make it easier as well

i will look into adding functionality for the cross-pool case. admittedly i haven't been testing with the stunnel idle timeout set to 60s because i've just been trying to get it to work - this should flag any problems with the cross-pool case

robhoes: Yep, I am familiar with the original code ;)

lippirk: i also got rid of some unnecessary function args (e.g. __context, local_fn etc.), since these are all in scope anyway

lippirk:
cross pool migration working as expected, going to run BST

ocaml/xapi/message_forwarding.ml
```diff
@@ -1679,44 +1679,54 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
-            (* Intra-pool: reserve resources on the destination host, then
-             * forward the call to the source. *)
+          if Db.is_valid_ref __context host then `Live_intrapool host else `Live_interpool
+        end else `Non_live
+      in
+      let forward_migrate_send () =
+        let forward_internal_async () =
+          let internal_async_op ~__context ~wait = (* this will be executed using the InternalAsync subtask's context *)
+            let host = Db.VM.get_resident_on ~__context ~self:vm in
+            do_op_on_no_localhost ~__context ~host (fun session_id rpc ->
```
robhoes, line 1693: I still don't think this `do_op_on_no_localhost` call is right. Here, `host` is the destination of the migration, but `VM.migrate_send` needs to run on the source host, where the VM is currently running (resident). The function `forward_vm_op` already gives you an `rpc` function and `session_id` for this source host, as is being passed to `op` (a.k.a. `Client.VM.migrate_send`).

lippirk: i don't think using the pre-existing `rpc` and `session_id` will work because it is associated with the parent task context (`Async.VM.migrate_send`). we need to create a new session so that the source host will write progress to the subtask context (`InternalAsync.VM.migrate_send`)

robhoes: Ignore my comment about the `host` being wrong; I seem to have misread it.

The `__context` on this line is the same as the one passed to `forward_vm_op` below. The main difference between `do_op_on` and `do_op_on_no_localhost` is that you always call the remote op (with `call_slave_with_session`), even if `host` is the master (where this code is running). So even if the `host` is the master, you'll go through the API again rather than calling `Local.VM.migrate_send`. If `host` is a slave, then you'll get a new rpc function (a slave TCP connection) and session (a slave login token) for it, and will throw away the ones that `forward_vm_op` has just created from the same context?

lippirk: it's not the same context - see that it is passed as a param to `internal_async_op`, which is then called in during the subtask:

```
let res = Server_helpers.exec_with_subtask ~__context ~task_in_database:true subtask_label (fun ~__context ->
      try
        `internal_async_successful (
          let wait t =
            info "try_internal_async: waiting for task to complete: t = ( %s )" (Ref.string_of t);
            Task.to_result ~__context:__parent_context ~of_rpc:marshaller ~t
          in
          (** callers are expected to perform the forwarding, but using the subtask context we provide.
            * we let callers choose how to wait on the outcome of the API call. this allows them to
              destroy sessions at the appropriate time *)
          internal_async_fn ~__context ~wait
        )
      with Api_errors.Server_error(code, params) when code=Api_errors.message_method_unknown -> `old_api
    )
```

as for the `do_op_on_no_localhost`, this won't be called if we are running on the master, because `forward_vm_op` is a wrapper around `do_op_on`. perhaps it is easier to see that it won't get called if we do some inlining:

```
forward_vm_op ~local_fn ~__context ~vm (fun session_id rpc ->
            Helpers.try_internal_async ~__context
                                       "InternalAsync.VM.migrate_send"
                                       API.ref_VM_of_rpc
                                       (fun ~__context ~wait -> 
                                         let host = Db.VM.get_resident_on ~__context ~self:vm in
                                         do_op_on_no_localhost ~__context ~host (fun session_id rpc ->
                                           wait (Client.InternalAsync.VM.migrate_send rpc session_id vm                                     dest live vdi_map vif_map options vgpu_map)
                                       )
                                         (fun () -> op session_id rpc)
                                       )
                                       (fun () -> op session_id rpc)
)
```

robhoes: OK, so I misread it again 🙄 

So the point is to not create a subtask when running on the master (`local_fn`). Previously it did, but that bypassed your progress/completion/cancel-check mirror loop. So you're actually creating a variant of `forward_vm_op` where the local and remote functions get different contexts/tasks.

It's a shame about the wasted connection and slave login though, which would happen in the common case. I wonder if that can be avoided? I still don't really like `do_op_on_no_localhost`... I we have to use it, then I'd prefer to just call `do_op_on ~local_fn`, even if we "know" that the `local_fn` won't be used (it' not like `do_op_on_no_localhost` has any "protection" for the case that `host` is the master).

Also, why did you factor out the `wait` function?

lippirk: >So the point is to not create a subtask when running on the master (local_fn). 

>Previously it did, but that bypassed your progress/completion/cancel-check mirror loop.

>So you're actually creating a variant of forward_vm_op where the local and remote functions get different contexts/tasks.
this is all correct ^

>It's a shame about the wasted connection and slave login though, which would happen in the common case. I wonder if that can be avoided?
it is a shame. it is definitely avoidable - we could have the slave create the subtask instead of the master - what do you think?

>I still don't really like do_op_on_no_localhost... I we have to use it, then I'd prefer to just call do_op_on ~local_fn, 
i'll do that (i weighed up that way vs this, guess...i picked the wrong one :P)

>Also, why did you factor out the wait function?
the internal async call is inside `do_op_on_no_localhost`, which destroys the session it creates when the op exits (=> session is destroyed too early)

lippirk:
bst on this new patch looking good

PR #4068, opened 2020-03-31
CA-337113: Turn create_tools_sr into remove_tools_sr

robhoes:
Tools SR.

The removal of the Tools SR is best-effort: a warning is logged if it fails (as
done before for redundant SRs). Failure occurs if there is still some VM that
has a Tools VDI attached to it. The cleanup code runs each time the master xapi
start, so once this situation has been resolved, the SR will eventually get
cleaned up.


ocaml/tests/suite_alcotest.ml
```diff
@@ -71,7 +71,6 @@ let () =
       @ Test_vm.tests
       @ Test_cpuid_helpers.tests
       @ Test_pool_cpuinfo.tests
-      @ Test_dbsync_master.tests
```
lippirk, line 74: this made me suspicious, but it looks like the dbsync master tests were actually only testing the tools sr :heavy_check_mark:

ocaml/xapi/dbsync_master.ml
```diff
@@ -107,90 +107,23 @@ let release_locks ~__context =
...
-    )
+  let srs = Db.SR.get_all ~__context in
+  let tools_srs =
+    List.filter (fun self -> Db.SR.get_is_tools_sr ~__context ~self) srs in
+  let old_srs =
+    List.filter (fun self ->
+        let other_config = Db.SR.get_other_config ~__context ~self in
+        (Db.SR.get_is_tools_sr ~__context ~self = false) &&
+        ((List.mem_assoc Xapi_globs.tools_sr_tag other_config) ||
+         (List.mem_assoc Xapi_globs.xensource_internal other_config))
```
lippirk, line 123: why does having a `xensource_internal` property mean it's an old sr?

robhoes: It's from the days before we had a proper `is_tools_sr` field, and even from before the days that we had a `tools_sr_tag` other_config key.

psafont: this check has just been moved, I'd rather keep the logic the same in here unless there's a reason to change it.

robhoes: Indeed.

PR #4070, opened 2020-04-01
CP-32663: Round of fixes for certificate installation

psafont:
The fixes are:
- correctly install certificates when using emergency reset
- fix edge-cases for alerting on the last and first day the period of alerting
- Change format of alerts to have a single root
- Output dates in the API  certificate errors in UTC
- Produce a readable error in the case an empty intermediate chain is provided
- Move the certificate installation code to gencert so it can be used by the gencert binary as well as xapi.


scripts/generate_ssl_cert
```diff
@@ -7,17 +7,24 @@
 set -e
 FILE=$1
 CN=$2
+FORCE=$3
 set -eu
-if [ -z "${CN}" ] || [ -z "${FILE}" ]; then
-	echo "usage: $0 <certname> <cn>"
+
+if [ -z "${CN}" ] || [ "${CN}" = "--force" ] || [ -z "${FILE}" ] || [ "${FILE}" = "--force" ]; then
```
lindig, line 13: It might be better to use `getopt(1)` for command line handling. https://www.shellscript.sh/tips/getopts/

psafont: I'm working on removing the shell script with ocaml code, now that the branch is rebased on top of Ben's work with an ocaml binary for generating certs (now it simply calls the script)

psafont: getopt doesn't seem to support positional arguments at all.

lindig: Maybe it's not worth doing.
https://stackoverflow.com/questions/11742996/is-mixing-getopts-with-positional-parameters-possible

ocaml/alerts/certificate/certificate_check.ml
```diff
@@ -2,10 +2,12 @@ module XenAPI = Client.Client
 module Date = Xapi_stdext_date.Date
 
 let seconds_per_day = 3600. *. 24.
-let seconds_per_30_days = 30. *. seconds_per_day
 
 let days_until_expiry epoch expiry =
-  int_of_float (expiry /. seconds_per_day -. epoch /. seconds_per_day)
+  let days = expiry /. seconds_per_day -. epoch /. seconds_per_day in
+  match Float.sign_bit days with
```
lippirk, line 8: why not `days <= 0.`?

ocaml/alerts/certificate/certificate_check.ml
```diff
@@ -23,7 +25,7 @@ let generate_alert epoch (host, expiry) =
     let expired_message = "The TLS server certificate has expired." in
 
     let body msg =
-      Printf.sprintf "<message>%s</message><date>%s</date>" msg (Date.to_string expiry)
+      Printf.sprintf "<body><message>%s</message><date>%s</date></body>" msg (Date.to_string expiry)
```
lippirk, line 28: could `msg` ever come from some malicious user?

ocaml/xapi/certificates.ml
```diff
@@ -364,7 +364,7 @@ let validate_certificate kind pem now private_key =
   in
 
   let ensure_validity ~time certificate =
-    let to_string = Ptime.to_rfc3339 in
+    let to_string = Ptime.to_rfc3339 ~tz_offset_s:0 in
```
lippirk, line 367: :+1:

ocaml/alerts/certificate/certificate_check.ml
psafont: Because of how floats are [stored](https://en.wikipedia.org/wiki/Single-precision_floating-point_format#IEEE_754_single-precision_binary_floating-point_format:_binary32): every float has a sign bit, this is why `-0.` is an actual value.

Checking the bit should also be the most efficient way to check for this.

ocaml/alerts/certificate/certificate_check.ml
psafont: No, the binding is local (in `general_alert`) and it's only fed local values (2 possibilities, `expiring_message` and `expired_mesage`)

ocaml/alerts/certificate/certificate_check.ml
lippirk: okay, in this case, i assume you are happy that the certificate has exactly expired, we will report 0 days until expiry, rather than -1?
(since `Float.sign_bit 0. = true`)

psafont: Yes, I don't think it makes a difference which alert the user is seeing, both have the same priority, and I don't think it's worth tweaking for a case that's _really_ unlikely to happen in a real environment.

lippirk: agreed

ocaml/alerts/certificate/certificate_check.ml
```diff
@@ -61,10 +59,23 @@ let execute rpc session (host, alert) =
 
 let alert rpc session =
   let now = Unix.time () in
+  (* Message starting with [host_server_certificate_\{expiring,expired\}] may
+     need to be refreshed, gather them just once *)
+  let previous_messages = XenAPI.Message.get_all_records rpc session
+  |> List.filter (fun (ref, record) ->
+      let expiring_or_expired name =
```
lindig, line 66: I think `message` would be a better name but I understand it comes from the name of the field.

ocaml/gencert/lib.ml
```diff
@@ -32,3 +32,115 @@ let get_management_ip_addr ~dbg =
...
+  let ensure_sha256_signature_algorithm certificate =
+    match X509.Certificate.signature_algorithm certificate with
+    | Some (_, `SHA256) ->
+        Ok certificate
+    | _ ->
+        Error (`Msg (server_certificate_signature_not_supported, []))
+  in
+
+  let raw_pem = Cstruct.of_string pem in
+  (match kind with
```
lindig, line 99: Why do we need `(` here?

psafont: It's not needed, probably a leftover when it wasn't the last statement in the function.

PR #4071, opened 2020-04-02
Enable dormant tests

psafont:
The PCI tests are not being run, the database tests are not run under opam and the db test  unit_test_sql didn't seem to be run in any case.
Ported the database tests to alcotest so now the ounit dependency can be dropped from xapi.

It's still used to thest ocaml-pci, but that is probably better done upstream, along with the fixes applied in this repo.


psafont:
It's the same, I guess it was used as a way to test the performance.

PR #4077, opened 2020-04-15
CP-31118: Avoid xapi as module name in logs

psafont:
Instead use the filename as is more informative to locate the code that
produces the loglines.

This changes the modules that were missed by
44a0cfcc09198b8273f908c93d0ff18c6b2d6b5c

The change was generated by running

```
$ for file in (rg --files-with-matches -F "Debug.Make(struct let name=\"xapi\" end)")
      set filename (echo (basename (string split -r -m1 . $file)[1]))
      sed -i "s/name=\"xapi\"/name=\"$filename\"/g" "$file"
  end
```

### DIFF
--- a/ocaml/xapi/attach_helpers.ml
+++ b/ocaml/xapi/attach_helpers.ml
@@ -14,7 +14,7 @@
 open Stdext.Pervasiveext
 open Client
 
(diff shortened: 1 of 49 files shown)


lippirk:
nice !

PR #4078, opened 2020-04-15
CA-337929 remove gencert xapi-wait-init-complete dependency

lippirk:
stunnel@xapi.service depends on gencert, so if the
xapi-wait-init-complete service fails, it takes 4 minutes for stunnel to
start. This was seen when rebooting slaves, but is easily fixed by
removing this dependency

### DIFF
--- a/scripts/gencert.service
+++ b/scripts/gencert.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=Generate xapi ssl certificate
-Requires=xapi.service forkexecd.service xapi-init-complete.target
-After=xapi.service forkexecd.service xapi-init-complete.target
+Requires=forkexecd.service
+After=forkexecd.service
 
 [Service]
 Type=oneshot


lippirk:
we should also fix the problem with the xapi-wait-init-complete.service failing

scripts/gencert.service
```diff
@@ -1,7 +1,7 @@
 [Unit]
 Description=Generate xapi ssl certificate
-Requires=xapi.service forkexecd.service xapi-init-complete.target
-After=xapi.service forkexecd.service xapi-init-complete.target
+Requires=xapi.service forkexecd.service
+After=xapi.service forkexecd.service
```
edwintorok, line 4: Can you remove xapi.service too? All you need there is the inventory, do you actually need to communicate with XAPI?

lippirk: you're right

edwintorok:
Is xapi-wait-init-complete timing out due to the master not being able to reach the slave (that would need stunnel running, but stunnel needs gencert, and gencert needs xapi-wait-init-complete)? i.e. does it reproducibly always fail like this on a slave?

PR #4079, opened 2020-04-16
CA-337903 xapi-domains requires stunnel

lippirk:
Stunnel is required to start/stop domains on slaves.

Fixes CA-337894 (and other related bugs).

### DIFF
--- a/scripts/xapi.service
+++ b/scripts/xapi.service
@@ -15,6 +15,7 @@ After=xcp-networkd.service
 After=xcp-rrdd.service
 After=xenopsd-xc.service
 After=xenstored.service
+After=stunnel@xapi.service
 
 Conflicts=shutdown.target
 


scripts/xapi-domains.service
```diff
@@ -1,9 +1,9 @@
 [Unit]
 Description=Start/stop domains on dom0 start/shutdown
 Requires=proc-xen.mount
-Wants=xapi-init-complete.target
+Wants=xapi-init-complete.target stunnel@xapi.service
```
lippirk, line 4: i've been testing with only 'After += stunnel@xapi.service', but i think we probably also want this ^

(`Wants` rather than `Requires`, since we if we are starting VMs on the master we shouldn't need stunnel)

scripts/xapi-domains.service
```diff
@@ -1,9 +1,9 @@
...
 # xapi.service is needed for shutdown ordering
-After=remote-fs.target xapi-init-complete.target xapi.service
+After=remote-fs.target xapi-init-complete.target xapi.service stunnel@xapi.service
```
edwintorok, line 6: I thought that After is redundant here, but according to the docs apparently not. Wants does not introduce an ordering dependency ... so this is correct.

scripts/xapi-domains.service
edwintorok: "Wants" is probably fine. We don't want a "Requires" because that'd also stop xapi-domains, if stunnel is stopped which would be bad because we do have code that restarts stunnel and we wouldn't expect VMs to be shutdown as a side-effect.

lindig: I think it would be good to have this on its own line - it makes patches easier

lippirk: > "Wants" is probably fine. We don't want a "Requires" because that'd also stop xapi-domains, if stunnel is stopped which would be bad because we do have code that restarts stunnel and we wouldn't expect VMs to be shutdown as a side-effect.

i'm struggling to find documentation that says systemd will automatically stops its services. can you convince me that adding a `Requires=stunnel@xapi.service` would cause that behaviour?

my understanding of the difference between `Requires` and `Wants` is that `Requires` adds a hard dependency, whereas `Wants` adds a soft one

psafont: What soft and hard dependencies mean:
```
Wants=
Units listed in this option will be started if the configuring unit is. However, if the listed units fail to start or cannot be added to the transaction, this has no impact on the validity of the transaction as a whole, and this unit will still be started.
```
```
Requires=
If this unit gets activated, the units listed will be activated as well. If one of the other units fails to activate, and an ordering dependency After= on the failing unit is set, this unit will not be started. Besides, with or without specifying After=, this unit will be stopped if one of the other units is explicitly stopped.
```

lippirk: i think only `After/Before` will affect the shutdown order

psafont: That's what I understood from the documentation as well.

lippirk:
@robhoes suggested we add `After=stunnel@xapi.service` to the xapi.service config. this is strange because it's xapi that actually starts stunnel. however it should fix the shutdown order once and for all

i'll run some tests with that change

lindig:
Can we un-draft this?

lippirk:
unfortunately this PR doesn't actually fix CA-337903, but it does fix a lot of other bugs, so i think it is worth merging asap

robhoes:
I think we should drop the xapi-domains.service change now?

lindig:
I was about to merge this and can't comment on Rob's suggestion.

lippirk:
> I think we should drop the xapi-domains.service change now? Those changes are not required and will only add to the confusion. Also, I think the `Wants=` clause in there is wrong, because it suggests that xapi-domains can start stunnel@xapi, but it can't actually...

ok, i will remove that. i think we should merge it immediately though - i don't think it's worth spending time testing it on my branch

PR #4080, opened 2020-04-17
CA-337875 base not always passed to sparse_dd

lippirk:
Due to operator precedences, the "-base" argument would not be passed to
sparse_dd whenever `prezeroed` was true. We fix this by using
List.concat instead of @, and parens by recommendation of ocamlformat

### DIFF
--- a/ocaml/xapi/sparse_dd_wrapper.ml
+++ b/ocaml/xapi/sparse_dd_wrapper.ml
@@ -66,14 +66,17 @@ let dd_internal progress_cb base prezeroed infile outfile size =
        try match Forkhelpers.with_logfile_fd "sparse_dd"
                    (fun log_fd ->
                       let sparse_dd_path = !Xapi_globs.sparse_dd in
-                      let args = [
-                        "-machine";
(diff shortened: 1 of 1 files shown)


ocaml/xapi/sparse_dd_wrapper.ml
```diff
@@ -66,14 +66,17 @@ let dd_internal progress_cb base prezeroed infile outfile size =
...
-                      ] @ if prezeroed then [ "-prezeroed" ] else []
-                        @ Option.value ~default:[] (Option.map (fun x -> [ "-base"; x ]) base) in
+                      let args =
+                        List.concat
+                          [ [ "-machine";
+                              "-src";  infile;
+                              "-dest"; outfile;
+                              "-size"; Int64.to_string size;
+                              "-good-ciphersuites"; Xcp_const.good_ciphersuites; ]
+                          ; (if prezeroed then [ "-prezeroed" ] else [])
```
edwintorok, line 76: Well now the parenthesis really are redundant, but nevermind, lets keep them.

PR #4081, opened 2020-04-20
Add support for replacing the TLS server certificates

psafont:
This includes API, CLI and alerts to support the feature.

Installed certificates using the API or CLI show up in the database, with its public key fingerprint and calidity period. The default certificates are _not_ stored in the database. Neither are the certificates that are generated with the emergency-server-certificate command.


psafont:
Certificate tests are broken due to a change in signature on certificate generation originating on the X509 0.11.0 update. This does not seem to affect the functionality itself. I'm rebuilding the dev environment to fix this.

lindig:
Travis fails because it relies on `xs-opam#master` while we will build this on `xs-opam#6.31.0` internally. We can accept this failure here for now.

psafont:
The old travis run passes: https://travis-ci.org/github/xapi-project/xen-api/builds/675325741

The new one is not because of the update in xs-opam of X509. This is a one-line fix: https://github.com/xapi-project/xen-api/pull/4082

ocaml/xapi/message_forwarding.ml
```diff
@@ -2578,6 +2578,42 @@ module Forward = functor(Local: Custom_actions.CUSTOM_ACTIONS) -> struct
...
+      let local_fn =
+        Local.Host.install_server_certificate
+          ~host
+          ~certificate
+          ~private_key
+          ~certificate_chain
+      in
+      let success () =
+        let progress = Db.Task.get_progress ~__context ~self:task in
+        if progress = 1. then Some () else None
```
lindig, line 2592: You might want to make this `if progress >= 1.0` because you are dealing with `float`

ocaml/gencert/test_data/certificates/pkey_rsa_8192.pem
```diff
@@ -0,0 +1,100 @@
+-----BEGIN PRIVATE KEY-----
```
edwintorok, line 1: Various security tools may winge at having private keys inside repositories. It is only test data in this case, could these be generated at runtime instead in the future? E.g. with openssl.

psafont:
> This depends on #4082 to fix the build if I got it right.

Yes, that will be merged a later date, it needs to go in at the same time as a new release for xs-opam. I would prefer to have that merged after the feature work, as this is more pressing than the upstream packages.

We have a happy travis build showing in the PR to show it was working before the xs-opam change.

ocaml/gencert/test_data/certificates/pkey_rsa_8192.pem
psafont: We could generate them, I elected not to do that as it introduces more complexity in the test code.

It will be hard to get rid of this error if the tools are grepping for this string, it's needed to generate one of the invalid private keys.

ocaml/xapi/message_forwarding.ml
psafont: Is that the case even if the value is set explicitly to 1.0, that is, no arithmetics are used to set `progress`?

PR #4084, opened 2020-04-20
CA-338141 mirror other_config when performing InternalAsync operation

lippirk:
We need to mirror other_config in addition to progress, because we have
tests which rely on this property.


psafont:
Travis fails on certificate tests due to the X509 version mismatch.
The rest of the tests are fine

PR #4085, opened 2020-04-22
CP-33511 improve mail alarm script

GeorgeBurac:


scripts/mail-alarm
```diff
@@ -34,66 +34,43 @@ FromLineOverride=YES
 ma_username="__dom0__mail_alarm"
 
 mail_language_pack_path = '/etc/xapi.d/mail-languages/'
```
edwintorok, line 36: Please include the ticket number in the commit title, it makes finding it later much easier.

scripts/mail-alarm
```diff
@@ -34,66 +34,43 @@ FromLineOverride=YES
...
+    return session.xenapi.pool.get_other_config(opaque_ref)
 
 def get_VM_params(uuid):
-    session = XenAPI.xapi_local()
-    session.xenapi.login_with_password(ma_username, "", "1.0", "xen-api-scripts-mail-alarm")
     try:
-        try:
-            opaque_ref = session.xenapi.VM.get_by_uuid(uuid)
-            return session.xenapi.VM.get_record(opaque_ref)
-        finally:
```
edwintorok, line 83: Indentation was wrong here, I think this code used to be broken, now its fixed.

scripts/mail-alarm
```diff
@@ -34,66 +34,43 @@ FromLineOverride=YES
 ma_username="__dom0__mail_alarm"
 
 mail_language_pack_path = '/etc/xapi.d/mail-languages/'
+session = XenAPI.xapi_local()
```
edwintorok, line 37: You now never login, I think you need to login at least once. At least for the purpose of making the session identifiable in xensource.log/audit.log as coming from the mail alarm script.

scripts/test_mail-alarm.py
```diff
@@ -53,7 +53,7 @@ def setUp(self):
         try:
             self.work_dir = tempfile.mkdtemp(prefix='test-mail-alarm-')
             log_file_global = os.path.join(self.work_dir, 'user.log')
-            src_file = './scripts/mail-alarm'
+            src_file = './mail-alarm'
```
edwintorok, line 56: Can you give an example why this was needed? Are there other tests that rely on this script and invoke it in a different way, or was this a purely manual test before?

scripts/mail-alarm
```diff
@@ -730,6 +707,7 @@ def main():
 
     finally:
         os.unlink(fname)
+        session._logout()
```
edwintorok, line 710: This leaks a session if `main()` returns before the `try` block is reached. I think the try/finally should be at a higher scope, for example below around the invocation of mail itself, or perhaps by using an at exit handler.

edwintorok:
Please add a `signed-off-by` line to your commit, see https://elinux.org/Developer_Certificate_Of_Origin for details, this will make the DCO bot pass.
If you have configured your `git` correctly, then `git commit --amend -s` will add the signed-off-by line. (If not then see `git config --global user.name "Your Name"` and `git config --global user.email "youremail"`).

Please don't get the discouraged by the many comments here, they are meant to point out some issues that you may not have been aware of, with the goal of improving the PR and expanding your knowledge. Feel free to ask for more details by replying to my individual comments, you don't necessarily have to agree with everything I said.

scripts/test_mail-alarm.py
psafont: Tests are failing because of this change, please check the file `.travis-python-nosetests-3.sh` to see how the tests are run. For local testing I recommend a python virtual environment with `mock` and `six` installed.

scripts/mail-alarm
psafont: Indentation in the file is also wrong elsewhere. Running [black](https://black.readthedocs.io/en/stable/) on the file might a good thing to do. The changes it produces should added last as a separate commit.

scripts/mail-alarm
psafont: Right now `session` is a global variable. I would remove it from the global scope, make it local in `main()` and add the session as a parameter to all the functions that use it. This will allow for `session` and the temporary file to be cleaned up with a context manager using `with` construction.

Documentation on how to define a context manager: https://docs.python.org/2/library/contextlib.html

If you need an example, here's one that changes the working directory and goes back the execution leave the `with block`: https://github.com/common-workflow-language/cwltool/blob/master/tests/util.py#L103 and how it's used:  https://github.com/common-workflow-language/cwltool/blob/beab66d649dd3ee82a013322a5e830875e8556ba/tests/test_fetch.py#L86

lindig:
Before we can review this further, please make sure your commit is on top the master branch such that we don't see unrelated commits here. This means you need to rebase your own branch on top of the master branch. Given that this is all work on single small file, your PR should comprise a single commit.

lindig:
We cleaned up the history such that this PR now contains only the actual commits.

scripts/mail-alarm
```diff
@@ -675,61 +662,67 @@ class XapiMessage:
...
-    if len(sys.argv) < 2:
-        log_err("Expected at least 1 argument but got none: [\"%s\"]." % (' '.join(sys.argv)))
-        raise Exception("Insufficient arguments")
+    try:
+        session = XenAPI.xapi_local()
+        session.xenapi.login_with_password(ma_username, "", "1.0", "xen-api-scripts-mail-alarm")
+    
+        if len(sys.argv) < 2:
+            log_err("Expected at least 1 argument but got none: [\"%s\"]." % (' '.join(sys.argv)))
+            raise Exception("Insufficient arguments")
```
psafont, line 671: I'd prefer if this were done before starting the xapi login.

psafont:
This looks much better now, thanks for the effort! :)

I think it's safe to squash the two commits in here with a single one. And after that. if you're up for it you could try using context manager for the temporary file and create a new one for the xen api lib. Maybe integrate it into XenAPI.py so other clients can also use it.

scripts/mail-alarm
```diff
@@ -676,60 +661,68 @@ class XapiMessage:
 
 def main():
     if len(sys.argv) < 2:
-        log_err("Expected at least 1 argument but got none: [\"%s\"]." % (' '.join(sys.argv)))
-        raise Exception("Insufficient arguments")
+            log_err("Expected at least 1 argument but got none: [\"%s\"]." % (' '.join(sys.argv)))
```
lindig, line 664: Indentation looks wrong. Please run all of this trough some automation indention tool

scripts/mail-alarm
```diff
@@ -676,60 +661,68 @@ class XapiMessage:
...
+            raise Exception("Insufficient arguments")
+    
+    
+    try:
+        session = XenAPI.xapi_local()
+        ma_username="__dom0__mail_alarm"
+        session.xenapi.login_with_password(ma_username, "", "1.0", "xen-api-scripts-mail-alarm")
```
psafont: I think this makes sense to be out of the `try:` block: it doesn't make sense to logout when we're not logged in yet.

lindig:
I have lost track where this stands. Are there unresolved comments? I would like to see a better commit message to explain in what way this is an improvement. The code reformatting doesn't need additional documentation - so keep this as two commits.

psafont:
The code changes look good to me, only one session is opened and used throughout the execution.

GeorgeBurac:
> I have lost track where this stands. Are there unresolved comments? I would like to see a better commit message to explain in what way this is an improvement. The code reformatting doesn't need additional documentation - so keep this as two commits.

I changed the commit message to "CP-33511 reduced XenAPI sessions to one per mail-alarm script invocation". Once it builds, I think all should be good.

PR #4086, opened 2020-04-22
CP-32980: security policy support for storage driver domains

edwintorok:


ocaml/xapi/system_domains.ml
```diff
@@ -78,6 +78,8 @@ let pbd_of_vm ~__context ~vm =
   then Some(Ref.of_string (List.assoc storage_driver_domain_key other_config))
   else None
 
+let driver_domain_ssid = ref 0x44525644l
```
lippirk, line 81: is there any justification as to why this wouldn't clash with the SSID of some ordinary domain? or is this a magic number from xen?

ocaml/idl/datamodel_vm.ml
```diff
@@ -1392,7 +1392,11 @@ let set_NVRAM_EFI_variables = call ~flags:[`Session]
              Published, rel_rio, "PCI bus path for pass-through devices";
              Deprecated, rel_boston, "Field was never used"]
              "PCI_bus" "PCI bus path for pass-through devices";
-           field  ~ty:(Map(String, String)) "other_config" "additional configuration" ~map_keys_roles:["pci", _R_POOL_ADMIN; ("folder",(_R_VM_OP));("XenCenter.CustomFields.*",(_R_VM_OP))];
+           field  ~ty:(Map(String, String)) "other_config" "additional configuration"
+             ~map_keys_roles:[ "pci", _R_POOL_ADMIN
+                             ; "storage_driver_domain", _R_POOL_ADMIN
+                             ; "is_system_domain", _R_POOL_ADMIN
+                             ; ("folder",(_R_VM_OP));("XenCenter.CustomFields.*",(_R_VM_OP))];
```
lippirk, line 1399: it'd be nice to make the parentheses consistent for this list

ocaml/xapi/system_domains.ml
edwintorok: ordinary domains get a ssid of 0 always. It is a magic number until we have proper Flask security policy support in Xen. (at which point we can refer to the policy instead of hardcoding it in the source code).

PR #4087, opened 2020-04-23
CP-32669: Remove vendored PCI library

psafont:
We need a few changes because one change in the vendored library was not needed to fix an issue and it was not upstreamed.

I also took the liberty to replace a function using [pciutil](https://github.com/xapi-project/xen-api-libs-transitional/blob/master/pciutil/pciutil.ml) with pci. This was the only usage I found for pciutl, so I think that package can be deleted quite soon :)


ocaml/xapi/monitor_master.ml
```diff
@@ -37,10 +36,19 @@ let update_configuration_from_master () =
...
   (* FIXME : put a lazy cache *)
-  let v, d = Pciutil.parse vendor device in
-  (match v with None -> "" | Some x -> x),
-  (match d with None -> "" | Some x -> x)
+  let pci_ids =
+    id_of_string vendor >>= fun vendor_id ->
+    id_of_string device >>= fun device_id ->
+    Pci.(with_access lookup_vendor_name) vendor_id >>= fun vendor_str ->
+    Pci.(with_access lookup_device_name) vendor_id device_id >>= fun device_str ->
+    Some (vendor_str, device_str)
```
psafont, line 49: For names and vendors without a name pcutil returned `Unknown vendor <vendor_id>`, now it returns `Vendor <vendor_id>`.

Unsure if this is bound to break on the upgrade case or even if it's important.

edwintorok: Pciutil doesn't depend on libpci (which we know had buggy interfaces with OCaml) so should be more reliable. However we're already using pci elsewhere in xapi, so might as well use it here.

ocaml/xapi/xapi_pci_helpers.ml
```diff
@@ -103,33 +98,31 @@ let get_host_pcis () =
             d.vendor_id d.device_id d.device_class;
           let vendor = { id = d.vendor_id
                        ; name = lookup_vendor_name access d.vendor_id
```
lippirk, line 100: how did `lookup_vendor_name` change from returning an `t option` to returning a `t`? was it previously from `pciutil` but now from `pci`?

could this change cause unexpected exceptions?

psafont: > how did lookup_vendor_name change from returning an t option to returning a t? was it previously from pciutil but now from pci?

This is an interface change between the versioned and the upstream pci versions, as explained in the commit:
```
The interface changes because the vendored library started using Option
with the intention to fix crashes. After the fact it was seen what the
actual solution to the crashes was. The proper solution was upstreamed
and the interface change was not rolled back in xapi, until now.
```

> i assume it returns the empty string instead of None now?

Now it return `Vendor <provided id>` for example `Vendor 00ffef`.

PR #4088, opened 2020-04-27
REQ-821: Separate out xapi's CLI server into its own library

minli1:
This make the code structure stronger and enforces that property the the CLI server inside xapi is just another API client.


robhoes:
This looks good. All these patches have been reviewed by various people before the PR was raised.

lindig:
I agree with @robhoes but would like to wait for Travis before we merge this.

robhoes:
Absolutely: no merge without a happy Travis.

robhoes:
Travis indicates that xapi-consts.opam needs a xapi-inventory dependency now.

ocaml/xe-cli/newcli.ml
```diff
@@ -293,7 +300,16 @@ let handle_unmarshal_failure ex ifd = match ex with
     else raise e
   | e -> raise e
 
-let main_loop ifd ofd =
+let assert_upload_filename_in_args filename permitted_filenames =
+  (* Prefix match instead of exact match is used here to workaround the old xva format that request files
+   * are not in the command line *)
+  let is_filename_permited = List.exists (fun file_in_arg -> Astring.String.is_prefix ~affix:file_in_arg filename)
```
edwintorok, line 303: This is not necessarily enough, you also want to block `/../` after the prefix, or check the canonicalized path.

ocaml/xapi/xapi_diagnostics.ml
```diff
@@ -0,0 +1,18 @@
+module D = Debug.Make(struct let name="xapi_diagnostics" end)
+open D
+
+let gc_compact ~__context ~session ~host =
+  raise Api_errors.(Server_error(not_implemented, [ "gc_compact" ]))
```
edwintorok: Why not implemented?

ocaml/xe-cli/newcli.ml
```diff
@@ -127,11 +127,8 @@ let parse_port (x: string) =
 
 let parse_eql arg =
   try
```
edwintorok, line 129: I don't think try/with is needed here anymore.

ocaml/xapi/xapi_diagnostics.ml
minli1: hmm, this is an outdated line. @liulinC was adding a Class at first and then add the messages. It's here to make it build

ocaml/xe-cli/newcli.ml
minli1: Yes, I logged this and we can refine it during the more restrict testing. Thanks.

edwintorok: Also I think similar filtering should be applied for download, not just upload.

minli1: IIRC there is a discussion about the whole name-checking instead of prefix, since now we have the old style of XVA. after remove the old style of XVA, we will change the checking here I think.

edwintorok: See fileserver.ml, there is already a helper: Stdext.Unixext.resolve_dot_and_dotdot.

minli1: Will log these places you point out. Thank you.

PR #4090, opened 2020-04-28
CA-337113: Turn create_tools_sr into remove_tools_sr

robhoes:
Tools SR.

The removal of the Tools SR is best-effort: a warning is logged if it fails (as
done before for redundant SRs). Failure occurs if there is still some VM that
has a Tools VDI attached to it. The cleanup code runs each time the master xapi
start, so once this situation has been resolved, the SR will eventually get
cleaned up.

This was previously reviewed on the tools-sr branch – see #4068.


lindig:
I assume this needs updated tests before we should merge it?

robhoes:
We may need test to be updated – let's not merge yet.

robhoes:
Turns out that we need a fix in quicktest as well.

robhoes:
Quicktest fix added. However, there are reports of RPU issues as well :(

ocaml/quicktest/quicktest_vm_import_export.ml
```diff
@@ -81,7 +70,7 @@ let import_export_test rpc session_id sr_info vm_template () =
                 if all.API.vBD_empty <> orig_vbd.API.vBD_empty
                 then Alcotest.fail (Printf.sprintf "Device %s varies in emptiness" all.API.vBD_userdevice);
                 match all.API.vBD_userdevice with
```
lindig, line 72: Should this not use a wildcard to catch unexpected value? Otherwise we would get a match runtime exception.

robhoes: There is one below.

robhoes:
I'm closing this until I've sorted out the RPU issue.

PR #4091, opened 2020-04-29
CA-338596: Upload files limit should deal with the dot style

liulinC:
Xe client permit upload filenames with "..", This commit set limitation for this.

### DIFF
--- a/ocaml/xapi/export.ml
+++ b/ocaml/xapi/export.ml
@@ -556,17 +556,27 @@ let metadata_handler (req: Request.t) s _ =
        let task_id = Ref.string_of (Context.get_task_id __context) in
        let read_fd, write_fd = Unix.pipe () in
        let export_error = ref None in
-       let writer_thread = Thread.create (fun () ->
-           (* lock all the VMs before exporting their metadata *)
-           List.iter (fun vm -> lock_vm ~__context ~vm ~task_id `metadata_export) vm_refs;
-           try
-             finally
-               (fun () -> export_metadata ~with_snapshot_metadata:export_snapshots ~preserve_power_state:true ~include_vhd_parents ~__context ~vms:vm_refs write_fd)
-               (fun () ->
-                  Unix.close write_fd;
-                  List.iter (fun vm -> unlock_vm ~__context ~vm ~task_id) vm_refs)
-           with e ->
-             export_error := Some e)
+       let writer_thread = Thread.create (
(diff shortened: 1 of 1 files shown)


ocaml/xe-cli/newcli.ml
```diff
@@ -300,10 +298,14 @@ let handle_unmarshal_failure ex ifd = match ex with
 let assert_upload_filename_in_args filename permitted_filenames =
   (* Prefix match instead of exact match is used here to workaround the old xva format that request files
    * are not in the command line *)
-  let is_filename_permited = List.exists (fun file_in_arg -> Astring.String.is_prefix ~affix:file_in_arg filename)
-      permitted_filenames in
+  let forbidden_path = ".." in
+  let is_filename_permited =  permitted_filenames
+                              |> List.exists (fun file_in_arg -> Astring.String.is_prefix ~affix:file_in_arg filename)
+                              && not (Stdext.Xstringext.String.has_substr filename forbidden_path) in
```
minli1, line 304: only check the filename include the `..` may block the normal cases that working well before.
See the user input the path like `../a.file`, the server will ask the client to upload this file but it will fail with this change.
Maybe we can use canonicalized path or give some more specific check on the `..`.
However, we need to think about the effort to remove the old style of XVA, after this we can have the whole string name checking which is the best choice.

liulinC: updated

ocaml/xe-cli/newcli.ml
```diff
@@ -126,9 +126,7 @@ let parse_port (x: string) =
     raise Usage
 
 let parse_eql arg =
```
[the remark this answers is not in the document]
liulinC: updated

ocaml/xe-cli/newcli.ml
```diff
@@ -300,8 +296,12 @@ let handle_unmarshal_failure ex ifd = match ex with
 let assert_upload_filename_in_args filename permitted_filenames =
   (* Prefix match instead of exact match is used here to workaround the old xva format that request files
    * are not in the command line *)
-  let is_filename_permited = List.exists (fun file_in_arg -> Astring.String.is_prefix ~affix:file_in_arg filename)
-      permitted_filenames in
+  let open Xapi_stdext_std in
+  let open Xapi_stdext_unix in
+  let is_filename_permited =  permitted_filenames
```
edwintorok, line 301: This should be moved to a function of its own so you can reuse it for download.

PR #4092, opened 2020-04-29
No more Tools ISO

robhoes:
Tools SR.


robhoes:
This replaces #4090.

robhoes:
I'm still running some tests on this.

ocaml/xapi/create_storage.ml
```diff
@@ -75,13 +75,42 @@ let maybe_create_pbd rpc session_id sr device_config me =
...
+        ((List.mem_assoc Xapi_globs.tools_sr_tag other_config) ||
+         (List.mem_assoc Xapi_globs.xensource_internal other_config))
+      ) srs in
+  let unplug_and_maybe_destroy sr =
+    (* Unplug me *)
+    let pbd, _ = Sm.get_my_pbd_for_sr __context sr in
+    Db.PBD.set_currently_attached ~__context ~self:pbd ~value:false;
+    (* Destroy any Tools SRs *)
+    try
+      Client.SR.forget rpc session_id sr;
```
lippirk, line 95: >We therefore let every host unplug its own PBD for the
Tools SR, and the last one to do so deletes the SR (`SR.forget` will only
proceed if all PBDs are unplugged).

why `SR.forget` rather than `SR.destroy`?

ocaml/xapi/create_storage.ml
```diff
@@ -75,13 +75,42 @@ let maybe_create_pbd rpc session_id sr device_config me =
...
+      info "Tools SR %s has been removed" (Ref.string_of sr)
+    with
+    | Api_errors.Server_error (e, _) when e = Api_errors.sr_has_pbd ->
+      info "Tools SR %s could not be removed: it is still in use by another host" (Ref.string_of sr)
+    | e ->
+      warn "Failed to destroy redundant Tools SR %s: %s"
```
lippirk, line 101: do you mean `forget` instead of `destroy` here?

ocaml/xapi/create_storage.ml
robhoes: Destroy means deleting VDIs, but that's not what we want. We just want to removed the DB objects.

ocaml/xapi/create_storage.ml
robhoes: Yeah, "destroy the DB object" and "forget the SR"... it's confusing... :)

lippirk:
do we want to deprecate the `is_tools_sr` field?
```
field ~qualifier:DynamicRO ~lifecycle:[Published, rel_dundee, ""] ~ty:Bool ~default_value:(Some (VBool false)) "is_tools_sr" "True if this is the SR that contains the Tools ISO VDIs";
```

ocaml/quicktest/quicktest_vm_import_export.ml
```diff
@@ -81,7 +70,7 @@ let import_export_test rpc session_id sr_info vm_template () =
                 if all.API.vBD_empty <> orig_vbd.API.vBD_empty
                 then Alcotest.fail (Printf.sprintf "Device %s varies in emptiness" all.API.vBD_userdevice);
                 match all.API.vBD_userdevice with
```
lindig, line 72: I think I've said this before: a wildcard match would avoid runtime match exceptions.

ocaml/xapi/create_storage.ml
```diff
@@ -75,13 +75,42 @@ let maybe_create_pbd rpc session_id sr device_config me =
...
+  let unplug_and_maybe_destroy sr =
+    (* Unplug me *)
+    let pbd, _ = Sm.get_my_pbd_for_sr __context sr in
+    Db.PBD.set_currently_attached ~__context ~self:pbd ~value:false;
+    (* Destroy any Tools SRs *)
+    try
+      Client.SR.forget rpc session_id sr;
+      info "Tools SR %s has been removed" (Ref.string_of sr)
+    with
+    | Api_errors.Server_error (e, _) when e = Api_errors.sr_has_pbd ->
```
lindig, line 98: Could use:
```
Api_errors.(Server_error (e, _) when e = sr_has_pb) ->
```

ocaml/quicktest/quicktest_vm_import_export.ml
robhoes: It's there on L86, isn't it?

robhoes:
> do we want to deprecate the is_tools_sr field?

Yes, good idea.

stormi:
Hi! I understand that there must be good reasons to drop the tools ISO in next CH release, but as a XAPI downstream user, we'd like to keep this feature. We rely on it to distribute the linux tools and may do too for windows tools in the future.

Would it be realistic to make configurable rather than completely dropped?

ocaml/quicktest/quicktest_vm_import_export.ml
lindig: You are correct - so forget my comment

robhoes:
@stormi Let me add a config file option for it, which defaults to false (no Tools SR), but can be set to true in the xapi.conf file. Would that do?

stormi:
@robhoes Yes, I think so

ocaml/xapi/create_storage.ml
```diff
@@ -75,13 +75,44 @@ let maybe_create_pbd rpc session_id sr device_config me =
...
+    List.filter (fun self ->
+        let other_config = Db.SR.get_other_config ~__context ~self in
+        (Db.SR.get_is_tools_sr ~__context ~self = false) &&
+        ((List.mem_assoc Xapi_globs.tools_sr_tag other_config) ||
+         (List.mem_assoc Xapi_globs.xensource_internal other_config))
+      ) srs in
+  let unplug_and_maybe_destroy sr =
+    (* Unplug me *)
+    let pbd, _ = Sm.get_my_pbd_for_sr __context sr in
+    Db.PBD.set_currently_attached ~__context ~self:pbd ~value:false;
```
edwintorok, line 92: I assume this is to prevent more uses of the tools SR, should the below forget fail because the SR is in-use (e.g. by a plugged in cd-rom device)

robhoes: The unplug means that this host is ready for the Tools SR to be removed. SR.forget will only work if all PBDs are unplugged.

nagilum99:
> Hi! I understand that there must be good reasons to drop the tools ISO in next CH release, but as a XAPI downstream user, we'd like to keep this feature. We rely on it to distribute the linux tools and may do too for windows tools in the future.

As I use that function quite often: I'm curious about the replacement for that. For Linux & Windows guests. ISO-Version was pretty universal and easy to use.

PR #4095, opened 2020-05-04
CA-336258: fix API forwarder for JSONRPC calls

robhoes:
Backport for Havana.

### DIFF
--- a/ocaml/xapi/api_server.ml
+++ b/ocaml/xapi/api_server.ml
@@ -101,7 +101,7 @@ module D=Debug.Make(struct let name="xapi" end)
 open D
 
 (** Forward a call to the master *)
-let forward req body rpc =
+let forward req rpc =
   let open Xmlrpc_client in
   let transport = SSL(SSL.make ~use_stunnel_cache:true (), Pool_role.get_master_address(), !Xapi_globs.https_port) in
   XMLRPC_protocol.rpc ~srcstr:"xapi" ~dststr:"xapi" ~transport ~http:{ req with Http.Request.frame = true } rpc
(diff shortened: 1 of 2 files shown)


ocaml/xapi/api_server.ml
```diff
@@ -197,7 +197,7 @@ let jsoncallback req bio _ =
     let response = Jsonrpc.a_of_response
         ~empty:Bigbuffer.make
         ~append:(fun buf s -> Bigbuffer.append_substring buf s 0 (String.length s))
-        (callback1 false req fd (Some body) rpc) in
+        (callback1 false req fd rpc) in
```
psafont, line 201: Seems strange to use `false` for `is_json` when the rpc is using Jsonrpc. On master thr first parameter is indeed true. 

Any idea on why has this changed?

robhoes: This is mysterious indeed. From what I can see, this has been `false` since 2012 when the `jsoncallback` was introduced. Then it was changed in 2017 when @kc284 properly implemented support for JSON-RPC.

Note that we actually have three HTTP endpoints (see xapi.ml):
```
  ("post_root", (Http_svr.BufIO (Api_server.callback false)));
  ("post_json", (Http_svr.BufIO (Api_server.callback true)));
  ("post_jsonrpc", (Http_svr.BufIO (Api_server.jsoncallback)));
```

These map to paths:
* `post_root` = `/`
* `post_json` = `/json`
* `post_jsonrpc` = `/jsonrpc`

The first two have existed "for ever". It seems to me that `/json` accepts XML-RPC with returns JSON in the response, while /jsonrpc is proper JSON-RPC. I bet that @jonludlam knows more about this.

So the `is_json` parameter may have had slightly different semantics. The function `jsoncallback` does its own JSONification, so it probably doesn't rely on `callback1` with `true`. However, now it does, due to my fix (and on master it already did due to other changes)!

Long story short: I'm going to fix this.

robhoes: This got merged early by accident. However, I did put in the fix: https://github.com/xapi-project/xen-api/commit/2792a44f42137ba86943b2b6eec915b3cbf5aa5e.

Now I'm going to set up a branch protection policy on this branch...

PR #4097, opened 2020-05-04
CA-337546: resource leak handling

edwintorok:
Needs to be merged together with https://github.com/xapi-project/xen-api-libs-transitional/pull/78

### DIFF
--- a/ocaml/database/master_connection.ml
+++ b/ocaml/database/master_connection.ml
@@ -19,6 +19,7 @@
 *)
 
 open Xapi_stdext_threads.Threadext
+open Safe_resources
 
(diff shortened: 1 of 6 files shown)


edwintorok:
I used a result type in the xapi-libs-transitional PR, I can try to see whether something like that would work here.

edwintorok:
Dropped the fallback ref code, use result type instead.
Note that there are more Unix.pipe in xapi that could be converted, but I'll leave that for later, we may not have to convert them.

lindig:
The DCO test is complaining - are commits signed off?

edwintorok:
There was a fixup without a signoff, I've squashed it now

PR #4098, opened 2020-05-07
CP-33121: Remove Stdext.monadic usages

psafont:
Most of the changes involve changing Either to result and Opt to Option.

The changes for Opt are mechanical and have been automated, for Either some fine-tuning for pretty printers, most of it was mechanical as well, please check the changes around vgpu code that were done as part of removing Either from the testcode. 

Some of the mechanical substitutions made:
```
sed -i "s/Opt.iter/Option.iter/g"
sed -i "s/Opt.map/Option.map/g"
sed -i "s/Opt.default /Option.value ~default:/g"
sed -i "s/Opt.unbox/Option.get/g"
sed -i "s/Opt.is_some/Option.is_some/g"
sed -i "s/Stdext.Option/Option/g"
```

Xapi does not depend anymore on xapi-test-utils for building, quicktest didn't really need it.

There's is still quite a bit of work to get rid of the parent Stdext package from xapi, 174 files still use it. I'll see if I can automate most of the changes, even if it is to use the subpackage directly.

### DIFF
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -1775,7 +1775,7 @@ let sync_pif_currently_attached ~__context ~host ~bridges =
   List.iter
     (fun (pif, pif_r) ->
        let bridge = List.assoc pif pif_to_bridge in
-       let currently_attached = Opt.default false (Opt.map (fun x -> List.mem x bridges) bridge) in
+       let currently_attached = Option.value ~default:false (Option.map (fun x -> List.mem x bridges) bridge) in
        if pif_r.API.pIF_currently_attached <> currently_attached then begin
          Db.PIF.set_currently_attached ~__context ~self:pif ~value:currently_attached;
          debug "PIF %s currently_attached <- %b" (Ref.string_of pif) currently_attached;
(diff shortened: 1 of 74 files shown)


ocaml/xapi/xha_interface.ml
```diff
@@ -180,7 +180,7 @@ module DaemonConfiguration = struct
     }
 
   let int_parameter (name, param) =
-    Opt.default [] (Opt.map (fun x -> [ xml_leaf_element name (string_of_int x) ]) param)
+    Option.value ~default:[] (Option.map (fun x -> [ xml_leaf_element name (string_of_int x) ]) param)
```
lippirk, line 183: can we replace this with something simpler?

```
match param with None -> [] | Some x -> [ xml_leaf_element name ... ]
```

if there are lots of statements like this, then maybe don't bother, but it's something to keep in mind

ocaml/xapi/xapi_host.ml
```diff
@@ -1775,7 +1775,7 @@ let sync_pif_currently_attached ~__context ~host ~bridges =
   List.iter
     (fun (pif, pif_r) ->
        let bridge = List.assoc pif pif_to_bridge in
-       let currently_attached = Opt.default false (Opt.map (fun x -> List.mem x bridges) bridge) in
+       let currently_attached = Option.value ~default:false (Option.map (fun x -> List.mem x bridges) bridge) in
```
lippirk, line 1778: similar thing here

ocaml/xapi/xha_interface.ml
edwintorok: Would `Option.fold` make this shorter? It would do the default value+map as one operation.

lippirk:
This is quite difficult to wade through, but it looks correct to me

PR #4099, opened 2020-05-07
CA-338137: Fix upgrade case in generate-iscsi-iqn

robhoes:
The generate-iscsi-iqn script sets the iscsi_iqn of a host on first boot, and
also runs on upgrade (this used to be a firstboot script).

Before XS 7.5, there was no dedicated `host.iscsi_iqn` field, and instead the
host IQN was stored in an other-config key. Upon upgrade from pre-Kolkata, the
new `iscsi_iqn` field is initialised to the empty string. In this case, the
generate-iscsi-iqn script is meant to copy the other-config key and reapply the
iscsi_iqn in order to set the new field. If the new field was already set, it
should leave it alone.

However, the if-statement for the latter is wrong and needs quotes. Without
that, it always evaluates to "true", even when `iscsi_iqn` is the empty string,
which causes the IQN to be regenerated rather than preserved.

### DIFF
--- a/scripts/generate-iscsi-iqn
+++ b/scripts/generate-iscsi-iqn
@@ -38,8 +38,9 @@ geniqn() {
 }
 
 # On upgrade, the host param will already be there - reuse it:
-if [ -n $(xe host-param-get uuid=${INSTALLATION_UUID} param-name=iscsi_iqn 2>/dev/null) ]; then
-    IQN=$(xe host-param-get uuid=${INSTALLATION_UUID} param-name=iscsi_iqn 2>/dev/null)
+IQN_PARAM="$(xe host-param-get uuid=${INSTALLATION_UUID} param-name=iscsi_iqn 2>/dev/null)"
+if [ -n "${IQN_PARAM}" ]; then
+    IQN="${IQN_PARAM}"
 elif xe host-param-get uuid=${INSTALLATION_UUID} param-name=other-config --minimal 2>/dev/null | grep -q "iscsi_iqn:" ; then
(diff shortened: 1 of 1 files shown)


lippirk (dismissed):

scripts/generate-iscsi-iqn
```diff
@@ -38,8 +38,8 @@ geniqn() {
 }
 
 # On upgrade, the host param will already be there - reuse it:
-if [ -n $(xe host-param-get uuid=${INSTALLATION_UUID} param-name=iscsi_iqn 2>/dev/null) ]; then
-    IQN=$(xe host-param-get uuid=${INSTALLATION_UUID} param-name=iscsi_iqn 2>/dev/null)
+if [ -n "$(xe host-param-get uuid=${INSTALLATION_UUID} param-name=iscsi_iqn 2>/dev/null)" ]; then
```
psafont, line 41: Using `[[ -n $(...) ]]` may be better if variables used inside need quoting to prevent splitting.
Another option would be using a temporary variable like `_IQN` to separate collecting the output from the testing, it also avoids executing the command twice.

robhoes: I'll add the variable, because executing the xe command twice is indeed a bit silly.

psafont (approved):

edwintorok (approved):
LGTM.
Perhaps these scripts should be in python/ocaml instead of bash which don't have quoting problems such as these.

Or we should run shellcheck on all the shellscripts and fail if shellcheck reports warnings/errors.

PR #4100, opened 2020-05-14
CA-334811 assign xapi version automatically

lippirk:
When building with opam, we use `dune subst` to interpolate the version.
We add `dune subst` to the build steps in all opam files to avoid silent
failures if we move xapi_version.ml to a different package. dune
recommends this when integrating with opam anyway.

We cannot use `dune subst` when building the xapi rpms (no git context).
To get around this we expose a make target. For completeness this works
in a developer context, where git is available.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,24 +3,25 @@ include config.mk
 XAPIDOC=_build/install/default/xapi/doc
 JOBS = $(shell getconf _NPROCESSORS_ONLN)
 PROFILE=release
+XAPI_VERSION ?= $(shell git describe --always --dirty || echo "NO_GIT")
 
 .PHONY: build clean test doc python reindent install uninstall
 
 build:
-	dune build @install -j $(JOBS) --profile=$(PROFILE)
+	XAPI_VERSION=$(XAPI_VERSION) dune build @install -j $(JOBS) --profile=$(PROFILE)
 
 # Quickly verify that the code compiles, without actually building it
 check:
-	dune build @check -j $(JOBS) --profile=$(PROFILE)
+	XAPI_VERSION=$(XAPI_VERSION) dune build @check -j $(JOBS) --profile=$(PROFILE)
 
 clean:
 	dune clean
 
 test:
-	dune runtest --profile=$(PROFILE) --no-buffer -j $(JOBS)
+	XAPI_VERSION=$(XAPI_VERSION) dune runtest --profile=$(PROFILE) --no-buffer -j $(JOBS)
 
 doc:
-	dune build --profile=$(PROFILE) ocaml/idl/datamodel_main.exe
+	XAPI_VERSION=$(XAPI_VERSION) dune build --profile=$(PROFILE) ocaml/idl/datamodel_main.exe
 	dune build --profile=$(PROFILE) -f @ocaml/doc/jsapigen
 	mkdir -p $(XAPIDOC)/html
 	cp -r _build/default/ocaml/doc/api $(XAPIDOC)/html
--- a/ocaml/util/gen_build_info_ml.sh
+++ b/ocaml/util/gen_build_info_ml.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+set -e
+
+DATE=$(date "+%Y-%m-%d")
+
+if [[ -z "$XAPI_VERSION" ]]; then
+  printf "XAPI_VERSION not set" 1>&2
+  exit 1
+fi
+XAPI_VERSION="${XAPI_VERSION#v}"
+
+printf "let date = \"%s\"\n\n" "$DATE"
+printf "let version = \"%s\"\n" "$XAPI_VERSION"
--- a/ocaml/util/xapi_version.ml
+++ b/ocaml/util/xapi_version.ml
@@ -9,5 +9,8 @@
 let git_id = ""
 let hostname = "localhost"
 let date = Build_info.date
-let xapi_version_major = 1
-let xapi_version_minor = 20
+let (xapi_version_major, xapi_version_minor) =
+  try
+    Scanf.sscanf Build_info.version "%d.%d.%s" (fun maj min _rest -> (maj, min))
+  with _ ->
+    failwith (Printf.sprintf "Couldn't determine xapi version - got unexpected XAPI_VERSION='%s'" Build_info.version)
--- a/xapi-client.opam
+++ b/xapi-client.opam
@@ -4,6 +4,7 @@ authors: [ "xen-api@lists.xen.org" ]
 homepage: "https://github.com/xapi-project/xen-api"
 bug-reports: "https://github.com/xapi-project/xen-api/issues"
 dev-repo: "git+https://github.com/xapi-project/xen-api.git"
+build-env: [[ XAPI_VERSION = "v0.0.0" ]]
 build: [[ "dune" "build" "-p" name ]]
 
 depends: [
--- a/ocaml/util/dune
+++ b/ocaml/util/dune
@@ -1,6 +1,7 @@
 (rule
   (targets build_info.ml)
-  (action (with-stdout-to %{targets} (run "date" "+let date=\"%Y-%m-%d\"")))
+  (deps %{workspace_root}/ocaml/util/gen_build_info_ml.sh)
+  (action (with-stdout-to %{targets} (bash "./gen_build_info_ml.sh")))
 )
 
 (library
--- a/xapi-cli-protocol.opam
+++ b/xapi-cli-protocol.opam
@@ -4,6 +4,7 @@ authors: [ "xen-api@lists.xen.org" ]
 homepage: "https://github.com/xapi-project/xen-api"
 bug-reports: "https://github.com/xapi-project/xen-api/issues"
 dev-repo: "git+https://github.com/xapi-project/xen-api.git"
+build-env: [[ XAPI_VERSION = "v0.0.0" ]]
 build: [[ "dune" "build" "-p" name ]]
 
 depends: [
--- a/xapi-consts.opam
+++ b/xapi-consts.opam
@@ -4,6 +4,7 @@ authors: [ "xen-api@lists.xen.org" ]
 homepage: "https://github.com/xapi-project/xen-api"
 bug-reports: "https://github.com/xapi-project/xen-api/issues"
 dev-repo: "git+https://github.com/xapi-project/xen-api.git"
+build-env: [[ XAPI_VERSION = "v0.0.0" ]]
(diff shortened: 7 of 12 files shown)


ocaml/util/xapi_version.ml
```diff
@@ -9,5 +9,14 @@
...
+let (xapi_version_major, xapi_version_minor) =
+  (* v will be interpolated when:
+     - creating an RPM (`make version`)
+     - compiling with opam (`dune subst`)
+  *)
+  let v = "%%VERSION_NUM%%" in
+  try
+    Scanf.sscanf v "%d.%d.%s" (fun maj min _rest -> (maj, min))
+  with _ ->
+    (0, 0)
```
psafont, line 22: I would rather makes this utterly fail. We don't want to release a version that advertises itself as 0.0.

I would also like to expose the hash commit (now part of `_rest`), if it's available. I guess a `xapi_version_revision = Some rev` would do, along with modifying the code that exposes the version.

Makefile
```diff
@@ -40,6 +41,10 @@ doc-json:
 reindent:
 	git ls-files '*.ml*' '**/*.ml*' | xargs ocp-indent --syntax cstruct -i
 
+version:
```
robhoes, line 44: Should we let the `build` target depend on this, so that it remains a "do everything" target?

psafont: I'm not sure about the suggestion, when building from the repository we generally want to use `dune subst` so the version can be grabbed from `git describe`.

I think this target is mainly useful only when building from tarball, when `git describe` is not available.

lippirk: i didn't let the build target depend on version, because you then have to carry around a change in your working directory:

```
[ben@piglet:xen-api {master} (CA-334811)]$ git diff
diff --git a/ocaml/util/xapi_version.ml b/ocaml/util/xapi_version.ml
index 0d903aef8..a67b54072 100644
--- a/ocaml/util/xapi_version.ml
+++ b/ocaml/util/xapi_version.ml
@@ -15,7 +15,7 @@ let (xapi_version_major, xapi_version_minor) =
      - creating an RPM (`make version`)
      - compiling with opam (`dune subst`)
   *)
-  let v = "%%VERSION_NUM%%" in
+  let v = "1.244.0-1-g0303cdaf1" in
   try
     Scanf.sscanf v "%d.%d.%s" (fun maj min _rest -> (maj, min))
   with _ ->
```

Using `dune subst` is actually a lot worse, because it adds a version field to all the opam files...

```
[ben@piglet:xen-api {master} (CA-334811)]$ git status
On branch CA-334811
Your branch is up-to-date with 'origin/CA-334811'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   Makefile
        modified:   dune-project
        modified:   ocaml/util/xapi_version.ml
        modified:   xapi-cli-protocol.opam
        modified:   xapi-client.opam
        modified:   xapi-consts.opam
        modified:   xapi-database.opam
        modified:   xapi-datamodel.opam
        modified:   xapi-types.opam
        modified:   xapi.opam
        modified:   xe.opam

no changes added to commit (use "git add" and/or "git commit -a")
```

robhoes: Yeah, it would be much nicer if the build does not change any tracked files. If it can't be done by environment variables, then perhaps by creating a temporary file that is included (that's how we used to do it).

ocaml/util/xapi_version.ml
lippirk: if we allow this to fail, then you can't compile xapi without interpolating the version (i think this is because of the autogenerated code)

but interpolating the version means you have to carry around version changes in the working directory which is a pain

the  rev is available (if you use dune subst) it looks like this = 1.244.0-1-g0303cdaf1. i think we should leave implementing this until we know exactly what we need for koji

lippirk:
i noticed we can use the dune rule stanza here. there already exists a rule to create a file containing the date:

```
 (rule
   (targets build_info.ml)
   (action (with-stdout-to %{targets} (run "date" "+let date=\"%Y-%m-%d\"")))
 )

```

so i'll tweak this to add the version. this makes @psafont happy because we can reject invalid versions, and it makes @robhoes happy because we don't need a new make target - just a new env var. additionally we won't have to deal with any uncommited changes floating around during development

i don't think `dune subst` is useful at all here

Makefile
```diff
@@ -3,11 +3,12 @@ include config.mk
 XAPIDOC=_build/install/default/xapi/doc
 JOBS = $(shell getconf _NPROCESSORS_ONLN)
 PROFILE=release
+XAPI_VERSION=$(shell git describe --always --dirty)
```
robhoes, line 6: I think that this would just override the variable, if you have specified it on the command line. Perhaps use `?=` instead of `=`.

ocaml/util/gen_build_info_ml.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e
+
+DATE=$(date "+%Y-%m-%d")
+# shellcheck disable=SC2001
+XAPI_VERSION=$(echo "$XAPI_VERSION" | sed "s/^v//g")
+
+printf "let date = \"%s\"\n\n" "$DATE"
```
psafont, line 8: Please do no add the date to the executable as part of the build process. It breaks reproducible builds.

robhoes: We've already got that. It was just done in the dune file itself.

Makefile
lippirk: `make  -e ENV=var` works, but `ENV=var make` doesn't - i suspect `?=` will fix that

ocaml/util/gen_build_info_ml.sh
lippirk: this was already there:

```
  (action (with-stdout-to %{targets} (run "date" "+let date=\"%Y-%m-%d\"")))
```
it gets used here:

```
let software_version () =
  (* In the case of XCP, all product_* fields will be blank. *)
  List.filter (fun (_, value) -> value <> "")
    [_product_version, Xapi_version.product_version ();
...
     _date,            Xapi_version.date]
```

lippirk:
i forgot about opam builds, so i've reintroduced `dune subst`... i believe the version now works locally, for opam and for rpms

ocaml/util/gen_build_info_ml.sh
psafont: Let's leave this for another PR.

psafont:
I wouldn't worry about opam builds: it's never used to build binaries for the actual product, as long as it builds it's fine. If we worry about it we could make a test for it so it never fails.

lippirk:
> I wouldn't worry about opam builds: it's never used to build binaries for the actual product, as long as it builds it's fine. If we worry about it we could make a test for it so it never fails.

the version gets used during compilation (by the autogen code), so opam _does_ need to pass the correct version - otherwise travis won't pass. i'm trying to fix that now

we could instead use a default version like before (`0.0`), but as we discussed earlier this is not ideal

ocaml/util/xapi_version.ml
```diff
@@ -9,5 +9,9 @@
 let git_id = ""
 let hostname = "localhost"
 let date = Build_info.date
-let xapi_version_major = 1
-let xapi_version_minor = 20
+let (xapi_version_major, xapi_version_minor) =
+  try
+    Scanf.sscanf Build_info.version "%d.%d.%s" (fun maj min _rest -> (maj, min))
```
edwintorok, line 14: I see that you rely on the tag having at least 3 components, for xapi that is probably a safe assumption.

ocaml/util/xapi_version.ml
```diff
@@ -9,5 +9,9 @@
...
+  with _ ->
+    Printf.eprintf "Could not determine xapi_version (are you building with opam?) defaulting to (0,0)";
+    (0, 0)
```
edwintorok, line 17: As discussed it would be good to make this fail.
There is already a .gitarchive-info in this folder, but that only tells you the commit hash, and since opam just creates something like 'master.tar.gz' (for xs-opam), the actual version is lost.

Would a suitable workaround be to set in the opam file: `XAPI_VERSION=v0.0.0` in the build command? And then we can make this fail.

xapi-client.opam
```diff
@@ -4,6 +4,7 @@ authors: [ "xen-api@lists.xen.org" ]
 homepage: "https://github.com/xapi-project/xen-api"
 bug-reports: "https://github.com/xapi-project/xen-api/issues"
 dev-repo: "git+https://github.com/xapi-project/xen-api.git"
+build-env: [[ XAPI_VERSION = "v0.0.0" ]]
```
psafont, line 7: This could be set to opam's `version`, once we drop `master` as the version and start using the actual versions of the packages.

lippirk: yes, we can do this when we pin for 8.2 :+1:

Makefile
```diff
@@ -3,11 +3,12 @@ include config.mk
 XAPIDOC=_build/install/default/xapi/doc
 JOBS = $(shell getconf _NPROCESSORS_ONLN)
 PROFILE=release
+XAPI_VERSION ?= $(shell git describe --always --dirty)
```
lindig, line 6: This can fail when there is no git context. Would it be better to account for this?
```
API_VERSION ?= $(shell git describe --always --dirty || echo "000000")
```

ocaml/util/gen_build_info_ml.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e
+
+DATE=$(date "+%Y-%m-%d")
+# shellcheck disable=SC2001
+XAPI_VERSION=$(echo "$XAPI_VERSION" | sed "s/^v//g")
```
lindig, line 6: * Should this fail when there is no XAPI_VERSION defined? 
* The substitution can be done like this `XAPI_VERSION=${XAPI_VERSION#v}`

ocaml/util/xapi_version.ml
```diff
@@ -9,5 +9,8 @@
 let git_id = ""
 let hostname = "localhost"
 let date = Build_info.date
-let xapi_version_major = 1
-let xapi_version_minor = 20
+let (xapi_version_major, xapi_version_minor) =
+  try
+    Scanf.sscanf Build_info.version "%d.%d.%s" (fun maj min _rest -> (maj, min))
+  with _ ->
+    failwith "Couldn't determine xapi version (set XAPI_VERSION env var)"
```
lindig, line 16: Include the contents of `Build_info.version`

PR #4102, opened 2020-05-18
maintenance: prepare for ocamlformat

psafont:
With this change ocamlformat can be applied to the code without errors.

Ignors all the datamodel files as the format is quite special there and applying ocamlformat makes it hard to read.

### DIFF
--- a/ocaml/xapi/memory_check.ml
+++ b/ocaml/xapi/memory_check.ml
@@ -19,8 +19,8 @@ let ( --- ) = Int64.sub
 let ( *** ) = Int64.mul
 let ( /// ) = Int64.div
 
-(** Calculates the amounts of 'normal' and 'shadow' host memory needed *)
-(** to run the given guest with the given amount of guest memory.      *)
+(** Calculates the amounts of 'normal' and 'shadow' host memory needed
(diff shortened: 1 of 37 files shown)


ocaml/xapi/memory_check.ml
```diff
@@ -108,12 +108,15 @@ let vm_compute_used_memory ~__context policy vm_ref =
...
 type host_memory_summary = {
+  host_maximum_guest_memory_bytes: int64
   (** The maximum amount of memory that guests can use on this host. *)
-  host_maximum_guest_memory_bytes: int64;
+  ;
+  resident: API.ref_VM list
   (** list of VMs which have a domain running here *)
-  resident: API.ref_VM list;
+  ;
+  scheduled: API.ref_VM list
```
edwintorok, line 117: What happened here? I would've expected the `;` to go on same line as one of `resident`/`scheduled`

psafont: The field separator is `;`, so the field and the comment must not be separated by `;`

I can format it as ocamlformat would:
```
let ... =
{
; field : type (** doc *)
; field2 : type (** doc *)
)

PR #4103, opened 2020-05-19
CA-338602: lwsmd daemon should not be running when AD is not configured

liulinC:
CA-338602: lwsmd daemon should not be running when AD is not configured
    
    This commit suppose lwsmd service is not bootup by default, and
    manage lwsmd service as follows
    1. during xapi start, it will start lwsmd if the host is authed
    with AD
    2. start lwsmd during enabling AD auth
    3. stop lwsmd during disabling AD auth
    
### DIFF
--- a/ocaml/xapi/extauth_plugin_ADpbis.ml
+++ b/ocaml/xapi/extauth_plugin_ADpbis.ml
@@ -20,6 +20,56 @@ open D
 
 open Stdext.Xstringext
 
+let lwsmd_service = "lwsmd"
+
+module Lwsmd = struct
+
+  let is_ad_enabled ~__context =
+    Helpers.get_localhost ~__context
+    |> (fun self -> Db.Host.get_external_auth_type ~__context ~self)
+    |> (fun x -> x = Xapi_globs.auth_type_AD_Likewise)
+
+  let is_active() =
+    Fe_systemctl.is_active lwsmd_service
+
+  type t =
+    | Start
+    | Stop
+
+  let to_string = function
+    | Start -> "start"
+    | Stop -> "stop"
+
+  let operate ?(wait_until_success=false) op =
+    let op_str = op |> to_string in
+    try
+      ignore (Forkhelpers.execute_command_get_output !Xapi_globs.systemctl [op_str; lwsmd_service]);
+      let timeout = 5. in
+      if wait_until_success then
+        let success_cond = match op with
+          | Start -> is_active
+          | Stop -> Fun.negate is_active in
+        try
+          Helpers.retry_until_timeout ~timeout (Printf.sprintf "trying to %s %s" op_str lwsmd_service) success_cond
+        with _ -> debug "Fail to %s %s timeout %f" op_str lwsmd_service timeout
+    with e -> error "Fail to %s %s with error %s" op_str lwsmd_service (ExnHelper.string_of_exn e)
+
+  let stop ~wait_until_success =
+    operate ~wait_until_success Stop
+
+  let start ~wait_until_success =
+    operate ~wait_until_success Start
+
+  let init_service  ~__context =
+    (* This function is called during xapi start *)
+    (* it will start lwsmd service if the host is authed with AD *)
+    (* Xapi does wait lwsmd service boot up success as following reasons
+     * 1. The waiting will slow down xapi bootup
+     * 2. Xapi still needs to boot up even lwsmd bootup fail
+     * 3. Xapi does not need to use lwsmd functionality during its bootup *)
+    if is_ad_enabled ~__context then start ~wait_until_success:false
+end
+
 let match_error_tag (lines:string list) =
   let err_catch_list =
     [ "DNS_ERROR_BAD_PACKET",       Auth_signature.E_LOOKUP;
--- a/ocaml/xapi/extauth_plugin_ADpbis.ml
+++ b/ocaml/xapi/extauth_plugin_ADpbis.ml
@@ -453,10 +453,10 @@ struct
       (* the space-replacement char used by pbis is defined at /etc/pbis/lsassd.conf *)
       let current_lw_space_replacement = '+' in
       String.iteri (fun i c ->
-        if c = current_lw_space_replacement then
-          Bytes.set defensive_copy i ' '
-        else ()
-      ) lwname;
+          if c = current_lw_space_replacement then
+            Bytes.set defensive_copy i ' '
+          else ()
+        ) lwname;
       Bytes.unsafe_to_string defensive_copy
     in
     let get_value name ls = if List.mem_assoc name ls then List.assoc name ls else "" in
--- a/ocaml/xapi/extauth_plugin_ADpbis.ml
+++ b/ocaml/xapi/extauth_plugin_ADpbis.ml
@@ -40,7 +40,12 @@ module Lwsmd = struct
     | Start -> "start"
     | Stop -> "stop"
 
-  let operate ?(wait_until_success=false) op =
+  let enable_nsswitch() =
+    try
+      ignore (Forkhelpers.execute_command_get_output !Xapi_globs.domain_join_cli_cmd ["configure"; "--enable"; "nsswitch"]);
+    with e -> error "Fail to run %s with error %s" !Xapi_globs.domain_join_cli_cmd (ExnHelper.string_of_exn e)
+
+  let operate ~wait_until_success op =
     let op_str = op |> to_string in
     try
       ignore (Forkhelpers.execute_command_get_output !Xapi_globs.systemctl [op_str; lwsmd_service]);
@@ -63,11 +68,18 @@ module Lwsmd = struct
   let init_service  ~__context =
     (* This function is called during xapi start *)
     (* it will start lwsmd service if the host is authed with AD *)
-    (* Xapi does wait lwsmd service boot up success as following reasons
+    (* Xapi does not wait lwsmd service to boot up success as following reasons
      * 1. The waiting will slow down xapi bootup
      * 2. Xapi still needs to boot up even lwsmd bootup fail
      * 3. Xapi does not need to use lwsmd functionality during its bootup *)
-    if is_ad_enabled ~__context then start ~wait_until_success:false
+    if is_ad_enabled ~__context then (
+      start ~wait_until_success:false;
+      (* Xapi help to enable nsswitch during bootup if it find the host is authed with AD
+       * nsswitch will be automatically enabled with command domainjoin-cli
+       * but this enabling is necessary when the host authed with AD upgrade
+       * As it will not run the domainjoin-cli command again *)
+      enable_nsswitch()
+    )
 end
 
 let match_error_tag (lines:string list) =
@@ -739,12 +751,9 @@ struct
         (* execute the pbis domain join cmd *)
         try
           let (_: (string*string) list) =
-            pbis_common_with_password
-              pass
-              "/opt/pbis/bin/domainjoin-cli"
-              (["join"]
-               @ ou_params @ disabled_module_params @
-               ["--ignore-pam";"--notimesync";domain;user])
+            [["join"]; ou_params; disabled_module_params; ["--ignore-pam"; "--notimesync"; domain; user]]
+            |> List.concat
+            |> pbis_common_with_password pass !Xapi_globs.domain_join_cli_cmd
           in
 
           let max_tries = 60 in (* tests 60 x 5.0 seconds = 300 seconds = 5minutes trying *)
@@ -795,7 +804,7 @@ struct
            begin (* no windows admin+pass have been provided: leave the pbis host in the AD database *)
              (* execute the pbis domain-leave cmd *)
              (* this function will raise an exception if something goes wrong *)
-             let (_: (string*string) list) = pbis_common "/opt/pbis/bin/domainjoin-cli" ["leave"] in
+             let (_: (string*string) list) = pbis_common !Xapi_globs.domain_join_cli_cmd ["leave"] in
              ()
            end
          else
@@ -806,7 +815,7 @@ struct
              let user = convert_nt_to_upn_username (get_full_subject_name ~use_nt_format:false _user) in
              (* execute the pbis domain-leave cmd *)
              (* this function will raise an exception if something goes wrong *)
-             let (_: (string*string) list) = pbis_common_with_password pass "/opt/pbis/bin/domainjoin-cli" ["leave";user] in
+             let (_: (string*string) list) = pbis_common_with_password pass !Xapi_globs.domain_join_cli_cmd ["leave";user] in
              ()
            end;
          None (* no failure observed in pbis *)
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -1184,6 +1184,17 @@ let retry ~__context ~doc ?(policy = Policy.standard) f =
 let retry_with_global_lock ~__context ~doc ?policy f =
   retry ~__context ~doc ?policy (fun () -> with_global_lock f)
 
+(* Retry function f until success or timeout, f should return true on success *)
+let rec retry_until_timeout ?(interval=0.1) ?(timeout=5.) doc f =
+  match f() with
+  | true -> ()
+  | false -> let next_interval = interval *. 1.5 in
+    let next_timeout = timeout -. interval in
+    if next_timeout < 0. then
+      raise Api_errors.(Server_error(internal_error, [Printf.sprintf "retry %s failed" doc]));
+    Thread.delay interval;
+    retry_until_timeout ~interval:next_interval ~timeout:next_timeout doc f
+
 let get_first_pusb ~__context usb_group =
   try
     List.hd (Db.USB_group.get_PUSBs ~__context ~self:usb_group)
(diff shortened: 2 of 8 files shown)


lindig:
Is this the right approach? Would it be possible to handle this by systemd?

psafont:
I also think this should be handled by systemd if possible, is the AD service using systemd as well?

liulinC:
AD service (lwsmd) is handled by Systemd.
but it is expected to not started if the host is not authed with AD. (currently systemd will start it on boot).  Such info is only available from xapi, not systemd.
This approach suppose the `lwsmd` service is enabled by systemd, but not started by default, and only start it (through systemd) when necessary.

liulinC:
> Is this the right approach? Would it be possible to handle this by systemd?

systemd can only start/stop the service, service dependencies, and which level the service should start, My understanding about this, it cannot be handled by systemd as it does not know whether the host is authed with AD, which is only availabe from xapi

psafont:
One alternative would be to use a [file](https://www.freedesktop.org/software/systemd/man/systemd.path.html) that can be monitored by systemd, this is more complicated than it should be, however: https://unix.stackexchange.com/questions/286769/can-i-use-systemd-to-start-and-stop-a-service-based-on-the-presence-of-a-file

So I'm dubious it's better than the approach taken by this PR.

Forkexec has native support for systemd services, could you use that instead of executing the command?
https://github.com/xapi-project/forkexecd/blob/master/lib/fe_systemctl.mli

liulinC:
> One alternative would be to use a [file](https://www.freedesktop.org/software/systemd/man/systemd.path.html) that can be monitored by systemd, this is more complicated than it should be, however: https://unix.stackexchange.com/questions/286769/can-i-use-systemd-to-start-and-stop-a-service-based-on-the-presence-of-a-file
> 
> So I'm dubious it's better than the approach taken by this PR.
> 
> Forkexec has native support for systemd services, could you use that instead of executing the command?
> https://github.com/xapi-project/forkexecd/blob/master/lib/fe_systemctl.mli

https://github.com/xapi-project/forkexecd/blob/master/lib/fe_systemctl.mli  says
`[start_transient ?env ?properties ~service cmd args] generates and starts a transient systemd
 [service] that will execute [cmd args].` 
but here I need to run `lwsmd` as a long-term running service if necessary.  `fe_systemctl` may not appropriate for this case.
I will check https://www.freedesktop.org/software/systemd/man/systemd.path.html in detail later, but my first thought is writing extra file should be more complicated and implicit than run command.

psafont:
I think two approaches I mentioned add more complexity than the current approach. I was simply mentioning alternatives since both and christian have reservations with the current way.

liulinC:
> I think two approaches I mentioned add more complexity than the current approach. I was simply mentioning alternatives since both and christian have reservations with the current way.

aha, I see, sorry for my wrong reading 😄

lindig:
Good discussion. So consensus is the start of the service needs to be triggered by xapi as done by this PR?

ocaml/xapi/xapi.ml
```diff
@@ -669,6 +669,9 @@ let server_init() =
     (* only tries to call a hook script if there's some external auth_type defined, otherwise it is a useless operation *)
     if auth_type <> "" then
       try
+        if auth_type = Extauth.auth_type_AD_Likewise then
+          ignore (Forkhelpers.execute_command_get_output !Xapi_globs.systemctl ["start"; Extauth_plugin_ADpbis.lwsmd_service]);
```
psafont, line 673: Please use `Extauth_plugin_ADpbis.start_lwsmd` instead

psafont:
> Good discussion. So consensus is the start of the service needs to be triggered by xapi as done by this PR?

I think there is no other way, the AD configuration bit is stored in xapi's database, so it xapi's responsibility to start the ad service at the right time, the question would be what is the appropriate mechanism to do so.

ocaml/xapi/xapi_globs.ml
```diff
@@ -466,6 +466,11 @@ let redo_log_alert_key = "metadata_lun_alerts"
 let serialize_pool_enable_disable_extauth = Mutex.create()
 (* CP-695: controls our asynchronous persistent initialization of the external authentication service during Xapi.server_init *)
 
+(* Auth types *)
+let auth_type_NONE = ""
+let auth_type_AD_Likewise = "AD"
+let auth_type_PAM = "PAM"
```
psafont, line 472: I would prefer a variant type with a `auth_type_of_string` function to just 3 string values. It's a shame it's not encoded like that in the database.

Could the database (de)serialization be changed to use this?

liulinC: yes, this is what I am thinking at first,
then we need to (de)serialization database values. 
It is a kind of code refine, and I just focus on issues in the ticket itself.

ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -20,6 +20,35 @@ open D
...
+    ignore (Forkhelpers.execute_command_get_output !Xapi_globs.systemctl ["start"; lwsmd_service]);
+    Thread.delay !Xapi_globs.lwsmd_service_bootup_time;
+  with e -> error "Fail to start %s with error %s" lwsmd_service (ExnHelper.string_of_exn e)
+
+let init_lwsmd_service  ~__context =
+  (* This function is called during xapi start *)
+  (* it will start lwsmd service if the host is authed with AD, stop lwsmd service otherwise *)
+  let op = match (is_ad_enabled ~__context) with
+    | true -> "start"
+    | false -> "stop" in
```
psafont, line 42: I'd like the service to be disabled by default instead of letting it start automatically and stopping it afterwards.

liulinC: yes, disable the auto-start of `lwsmd` is what I am trying to do . that is the first version.
But then I found the service is `Requred` and then bootup by `upgrade-likewise-to-pbis`. 
So the simplest way I found here is start/stop as it looks now.

psafont: > So the simplest way I found here is start/stop as it looks now.

The problem with this is that now we have a race condition. If somehow systemd decides to start the AD service after xapi run stop the service will keep running.

liulinC: yes, that is the reason I change `xapi.service` to sync the bootup, you can see it in the top commit

ocaml/xapi/xapi.ml
liulinC: updated

ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -20,6 +20,33 @@ open D
...
+let is_ad_enabled ~__context =
+  Helpers.get_localhost ~__context
+  |> (fun self -> Db.Host.get_external_auth_type ~__context ~self)
+  |> (fun x -> x = Xapi_globs.auth_type_AD_Likewise)
+
+let start_lwsmd ~__context =
+  (* if current host is auth with AD, then start lwsmd *)
+  try
+    ignore (Forkhelpers.execute_command_get_output !Xapi_globs.systemctl ["start"; lwsmd_service]);
+    Thread.delay !Xapi_globs.lwsmd_service_bootup_time;
```
psafont, line 34: We don't know after the delay whether the service has started or not. forkexechelpers has a function to gather the status of a systemd service to check this and fail accordingly.

Consider as well adding a maximum wait time after which we consider the service to have failed if it hasn't booted up yet.

liulinC: :+1: 
do you mean https://github.com/xapi-project/forkexecd/blob/master/lib/fe_systemctl.mli#L32

psafont: Correct

ocaml/xapi/extauth_plugin_ADpbis.ml
liulinC: @psafont  per discussion with PM, this PR will not be backport to older XS which can be upgraded from XS 6.x, 
and XS 7.x all use `likewise` instead of `pbis`, 
so we can safely disable `upgrade-likewise-to-pbis` , I have updated the PR, please have a look.

ocaml/xapi/helpers.ml
```diff
@@ -1184,6 +1184,15 @@ let retry ~__context ~doc ?(policy = Policy.standard) f =
...
   retry ~__context ~doc ?policy (fun () -> with_global_lock f)
 
+(* Retry function f until success or timeout, f should return true on success *)
+let rec retry_until_timeout ?(interval=0.1) ?(timeout=2.) doc f =
+  if timeout <= 0.0 then raise Api_errors.(Server_error(internal_error,
+                                                        [Printf.sprintf "retry %s failed with interval %f and timeout %f" doc interval timeout]));
+  match f() with
+  | true -> ()
+  | false -> Thread.delay interval;
+    retry_until_timeout ~interval ~timeout:(timeout -. interval) doc f
```
lindig, line 1194: You could use `interval *. 1.5` to increasingly slow down the loop but use a longer timeout of 5s.

ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -20,6 +20,55 @@ open D
...
+let lwsmd_service = "lwsmd"
+
+module Lwsmd = struct
+
+  let is_ad_enabled ~__context =
+    Helpers.get_localhost ~__context
+    |> (fun self -> Db.Host.get_external_auth_type ~__context ~self)
+    |> (fun x -> x = Xapi_globs.auth_type_AD_Likewise)
+
+  let is_bootup() =
```
psafont, line 32: I recommend keeping the name same as the library used: `is_active`

ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -20,6 +20,55 @@ open D
...
+    Fe_systemctl.is_active lwsmd_service
+
+  let is_stoped() =
+    not (is_bootup())
```
psafont, line 36: I don't see the value of this function, just use `Fun.negate is_active` where it's needed.

ocaml/xapi/helpers.ml
liulinC: sounds interesting, it will have more check density at beging while fewer check density later.
this possible the expected behavior

ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -20,6 +20,55 @@ open D
...
+  let stop ~wait_until_success =
+    operate ~wait_until_success Stop
+
+  let start ~wait_until_success =
+    operate ~wait_until_success Start
+
+  let init_service  ~__context =
+    (* This function is called during xapi start *)
+    (* it will start lwsmd service if the host is authed with AD *)
+    if is_ad_enabled ~__context then start ~wait_until_success:false
```
psafont, line 69: I guess the code does not wait for the service to be active becase
1. We do not want to add more time to xapi boot process
2. We do not want xapi to fail if he external AD fails

Regardless of the reason it would be good to explain them, could you add a comment with the reasons for this?
I'm happy with the comment being in the commit message or here in the code.

liulinC: updated

ocaml/xapi/extauth_plugin_ADpbis.ml
liulinC: updated

ocaml/xapi/helpers.ml
liulinC: updated, please have a look

liulinC:
@psafont @lindig  thanks for the comments, 
the concerns are addressed, could you please to review, and let me know if any further concern left, thanks

ocaml/xapi/helpers.ml
```diff
@@ -1184,6 +1184,17 @@ let retry ~__context ~doc ?(policy = Policy.standard) f =
...
 
+(* Retry function f until success or timeout, f should return true on success *)
+let rec retry_until_timeout ?(interval=0.1) ?(timeout=5.) doc f =
+  match f() with
+  | true -> ()
+  | false -> let next_interval = interval *. 1.5 in
+    let next_timeout = timeout -. interval in
+    if next_timeout < 0. then
+      raise Api_errors.(Server_error(internal_error, [Printf.sprintf "retry %s failed" doc]));
+    Thread.delay interval;
```
lindig, line 1195: This needs to use `next_interval`. It would be easier to just say `let interval = interval *. 1.5` above.

liulinC: No, this should be `interval` 
as `let next_timeout = timeout -. interval`,
I changed the code and it becomes a little more complicated to understand.
The change is basing on following thinking
If we sleep, we should try once more after wake up. (previous version may sleep, and raise error after wait up other than retry)

liulinC: > This needs to use `next_interval`. It would be easier to just say `let interval = interval *. 1.5` above.

No, this should be interval
as let next_timeout = timeout -. interval,
I changed the code and it becomes a little more complicated to understand.
The change is basing on following thinking
If we sleep, we should try once more after wake up. (previous version may sleep, and raise error after wait up other than retry)

liulinC:
> Seem my comment about `next_interval`

please check my replay and see whether that make sense.

ocaml/xapi/helpers.ml
lindig: Ok - so `Thread.delay` uses the old interval and the next iteration will use the next interval.

ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -20,6 +20,69 @@ open D
...
+  let stop ~wait_until_success =
+    operate ~wait_until_success Stop
+
+  let start ~wait_until_success =
+    operate ~wait_until_success Start
+
+  let init_service  ~__context =
+    (* This function is called during xapi start *)
+    (* it will start lwsmd service if the host is authed with AD *)
+    (* Xapi does wait lwsmd service boot up success as following reasons
```
lindig, line 73: The comment is misleading: I understand from the code that Xapi does not wait for the start-up.

ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -690,11 +753,8 @@ struct
         try
           let (_: (string*string) list) =
             pbis_common_with_password
-              pass
-              "/opt/pbis/bin/domainjoin-cli"
-              (["join"]
-               @ ou_params @ disabled_module_params @
-               ["--ignore-pam";"--notimesync";domain;user])
+              pass domain_join_cli_path
+              (["join"] @ ou_params @ disabled_module_params @ ["--ignore-pam";"--notimesync";domain;user])
```
lindig, line 757: Use `List.concat` and avoid `@` - it can lead to subtle bugs in connection with `if..then..else`

liulinC: updated

ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -20,6 +20,69 @@ open D
...
+    | Start -> "start"
+    | Stop -> "stop"
+
+  let enable_nsswitch() =
+    let cmd = domain_join_cli_path in
+    try
+      ignore (Forkhelpers.execute_command_get_output cmd ["configure"; "--enable"; "nsswitch"]);
+    with e -> error "Fail to run %s with error %s" cmd (ExnHelper.string_of_exn e)
+
+  let operate ?(wait_until_success=false) op =
```
lippirk, line 50: this optional parameter isn't necessary

ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -20,6 +20,69 @@ open D
 
 open Stdext.Xstringext
 
+let domain_join_cli_path = "/opt/pbis/bin/domainjoin-cli"
```
lippirk, line 23: you might want to add this to `Xapi_globs` and include it one of the lists of executables (see `Xapi_globs.ml:Resources`)

liulinC: > you might want to add this to `Xapi_globs` and include it one of the lists of executables (see `Xapi_globs.ml:Resources`)

updated, please have a look. thanks

ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -20,6 +20,67 @@ open D
...
+     * 1. The waiting will slow down xapi bootup
+     * 2. Xapi still needs to boot up even lwsmd bootup fail
+     * 3. Xapi does not need to use lwsmd functionality during its bootup *)
+    if is_ad_enabled ~__context then
+      start ~wait_until_success:false;
+    (* Xapi help to enable nsswitch during bootup if it find the host is authed with AD
+     * nsswitch will be automaticlly enabled with command domainjoin-cli
+     * but this enabling is necessary when the host authed with AD upgrade
+     * As it will not run the domainjoin-cli command again *)
+    enable_nsswitch()
```
psafont, line 81: nsswitch is enabled unconditionally. Looking at the comment this doesn't seem the intention.

To avoid this use this construction:
```
if cond then (
  statement ;
  another
)
```

liulinC: right, my fault, updated.

liulinC:
xenrt job: 2911118~2911127 looks good

PR #4107, opened 2020-05-26
CA-338565: Improve error message.

minli1:

### DIFF
--- a/ocaml/xe-cli/newcli.ml
+++ b/ocaml/xe-cli/newcli.ml
@@ -358,6 +358,8 @@ let main_loop ifd ofd permitted_filenames =
           marshal ofd (Blob End);
           Unix.close fd
         with
+        | Unix.Unix_error(Unix.EPIPE,_,_) ->
+            raise (ClientSideError (Printf.sprintf "Failed to upload file %s" x))
         | e -> marshal ofd (Response Failed)
       end
     | Command (HttpConnect(url)) ->
@@ -662,7 +664,9 @@ let main () =
          | Unix.WSIGNALED c -> "killed by signal " ^ (Xapi_stdext_unix.Unixext.string_of_signal c)
          | Unix.WSTOPPED c -> "stopped by signal " ^ string_of_int c)
     | Upload_filename_error e ->
-      error "Upload file error: %s" e
+      error "Upload file error: %s.\n" e
+    | ClientSideError e ->
+      error "Client Side error: %s.\n" e
     | e ->
       error "Unhandled exception\n%s\n" (Printexc.to_string e) in
   List.iter (fun p ->


ocaml/xe-cli/newcli.ml
```diff
@@ -342,11 +342,11 @@ let main_loop ifd ofd permitted_filenames =
     | Command (Debug x) -> debug "debug from server: %s\n%!" x
     | Command (Load x) ->
       begin
+        let length = (Unix.stat x).Unix.st_size in
```
liulinC, line 345: should this inside `try`

ocaml/xe-cli/newcli.ml
```diff
@@ -358,6 +358,11 @@ let main_loop ifd ofd permitted_filenames =
           marshal ofd (Blob End);
           Unix.close fd
         with
+        | Unix.Unix_error(Unix.EPIPE,fn,arg) as e ->
+          if length > Constants.max_cli_upload_bytes then
```
lippirk, line 362: if we exceed the upload limit, don't we want to fail before trying to read? (i.e. should we move this check before the try block?)

ocaml/xe-cli/newcli.ml
```diff
@@ -358,6 +358,11 @@ let main_loop ifd ofd permitted_filenames =
...
+            raise (ClientSideError (Printf.sprintf "The upload size of file '%s' probably exceeds the limit" x))
```
lindig, line 363:
```
"The size %d of file %s exceeds the upload limit of %d"
```

ocaml/xe-cli/newcli.ml
robhoes: The server-side (xapi) uses `Constants.max_cli_upload_bytes` to implement the size constraint. This code is for the client (xe), which may be older or newer than the server. Using the same `Constants.max_cli_upload_bytes` is a guess at best, because future versions of xapi may be stricter or less strict (older ones didn't have the constraint), while the same xe binary can be used to communicate with them.

It would therefore be better to not rely on `Constants.max_cli_upload_bytes` here. It would be better if xapi can tell xe why it is going to hang up, but to keep it simple, I would drop the if statement here and just report that the file upload failed. E.g. just `"Failed to upload file %s"`.

minli1:
Comments addressed.

ocaml/xe-cli/newcli.ml
```diff
@@ -662,7 +664,9 @@ let main () =
          | Unix.WSIGNALED c -> "killed by signal " ^ (Xapi_stdext_unix.Unixext.string_of_signal c)
          | Unix.WSTOPPED c -> "stopped by signal " ^ string_of_int c)
     | Upload_filename_error e ->
-      error "Upload file error: %s" e
+      error "Upload file error: %s.\n" e
+    | ClientSideError e ->
+      error "Client Side error: %s.\n" e
```
lindig, line 669: Why is this having a `\n`? This is going into the syslog.

ocaml/xe-cli/newcli.ml
```diff
@@ -358,6 +358,8 @@ let main_loop ifd ofd permitted_filenames =
           marshal ofd (Blob End);
           Unix.close fd
         with
+        | Unix.Unix_error(Unix.EPIPE,_,_) ->
```
lindig, line 361: Does it make sense to capture more errors than just `EPIPE` for this error message>

ocaml/xe-cli/newcli.ml
minli1: The error message in cli will goes to the stderr, it will be connected with shell if without a `\n`.
https://github.com/xapi-project/xen-api/blob/master/ocaml/xe-cli/newcli.ml#L47

ocaml/xe-cli/newcli.ml
minli1: Other errors apart from the `EPIPE` will be captured by the next match `| e ->`
I think other errors will have more info and captured by other match, the `EPIPE` happens when the server disconnected during the file transfer.

minli1:
Comments replied, please take a look, thank you. @lindig

PR #4112, opened 2020-05-27
CA-339656 use HOME when generating SSL certificate

lippirk:
openssl was complaining about being "unable to write 'random state'".
A fix for this is to set HOME or RANDFILE. We opt for the former to
avoid depleting the system's entropy.
     
openssl will now write to $HOME/.rnd

### DIFF
--- a/ocaml/gencert/gencert.ml
+++ b/ocaml/gencert/gencert.ml
@@ -15,16 +15,14 @@
 (* Usage: gencert <path> *)
 
 let inventory = "/etc/xensource-inventory"
-let generate_ssl_cert = "/opt/xensource/libexec/generate_ssl_cert"
 
 module Lib = Gencertlib.Lib
 module D = Debug.Make(struct let name = "gencert" end)
 
 let generate_cert_or_fail ~path ~cn =
-  let args = [path; cn] in
-  let (stdout, stderr) = Forkhelpers.execute_command_get_output generate_ssl_cert args in
-  let (stdout, stderr) = (String.escaped stdout, String.escaped stderr) in
+  let (stdout, stderr) = Lib.call_generate_ssl_cert ~args:[path; cn] in
 
+  let (stdout, stderr) = (String.escaped stdout, String.escaped stderr) in
   D.debug {|generate_ssl_cert stdout: "%s"|} stdout;
   D.debug {|generate_ssl_cert stderr: "%s"|} stderr;
   if Sys.file_exists path then
--- a/ocaml/gencert/gencert.ml
+++ b/ocaml/gencert/gencert.ml
@@ -22,9 +22,13 @@ module D = Debug.Make(struct let name = "gencert" end)
 let generate_cert_or_fail ~path ~cn =
   let (stdout, stderr) = Lib.call_generate_ssl_cert ~args:[path; cn] in
 
-  let (stdout, stderr) = (String.escaped stdout, String.escaped stderr) in
-  D.debug {|generate_ssl_cert stdout: "%s"|} stdout;
-  D.debug {|generate_ssl_cert stderr: "%s"|} stderr;
+  let debug_lines prefix std_x =
+    let lines = String.split_on_char '\n' std_x in
+    List.iter (fun l -> D.debug {|%s: "%s"|} prefix l) lines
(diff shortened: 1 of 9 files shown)


psafont:
Does this work with the `emergency-host-server-certificate-reset` call? I'm wary of the added error condition when `HOME` is not set.

lippirk:
> Does this work with the `xe emergency-host-server-certificate-reset` call? I'm wary of the added error condition when `HOME` is not set.

we get `HOME` from systemd:

```
[Service]
User=root
```

so i believe it would work. if you are not convinced we could make `gencert` more lenient by not failing if `HOME` isn't set

psafont:
looks like it should work, as forkexecd is used:
```
let emergency_reset_server_certificate ~__context =
  let generate_ssl_cert = "/opt/xensource/libexec/generate_ssl_cert" in

  let args = [!Xapi_globs.server_cert_path; Helper_hostname.get_hostname (); "--force"] in
  ignore @@ Forkhelpers.execute_command_get_output generate_ssl_cert args;
```

I'll doublecheck with a build, just in case

lippirk:
> looks like it should work, as forkexecd is used:
> 
> ```
> let emergency_reset_server_certificate ~__context =
>   let generate_ssl_cert = "/opt/xensource/libexec/generate_ssl_cert" in
> 
>   let args = [!Xapi_globs.server_cert_path; Helper_hostname.get_hostname (); "--force"] in
>   ignore @@ Forkhelpers.execute_command_get_output generate_ssl_cert args;
> ```
> 
> I'll doublecheck with a build, just in case

thanks for pointing this out - i didn't realise we call this script in two different locations. i'll have to update this invocation

lindig:
HOME is defined in Posix and must be defined (as I understand it). Is the change here where SSL writes it or that HOME was not defined?

ocaml/gencert/lib.ml
```diff
@@ -33,6 +36,14 @@ let get_management_ip_addr ~dbg =
       Some (List.hd addrs)
   with _ -> None
 
+let call_generate_ssl_cert ~args =
+  let home_env = match Sys.getenv_opt "HOME" with
+  | None -> D.warn "environment variable 'HOME' is unavailable"; ""
+  | Some h -> "HOME" ^ h
+  in
+  let env = [| "PATH=" ^ (String.concat ":" Forkhelpers.default_path); home_env |] in
+  Forkhelpers.execute_command_get_output generate_ssl_cert ~env args
```
psafont, line 45: Does it make sense to use Xapi_globs or Xapi_consts here to remove the duplicated definition?

lippirk:
> HOME is defined in Posix and must be defined (as I understand it). Is the change here where SSL writes it or that HOME was not defined?

the latter, previously openssl wasn't writing the random state to a file at all. to tell `generate_ssl_cert` what `HOME` is, we have to propagate it through systemd & through forkexecd

ocaml/gencert/lib.ml
```diff
@@ -33,6 +34,14 @@ let get_management_ip_addr ~dbg =
       Some (List.hd addrs)
   with _ -> None
 
+let call_generate_ssl_cert ~args =
+  let home_env = match Sys.getenv_opt "HOME" with
+  | None -> D.warn "environment variable 'HOME' is unavailable"; ""
```
lindig, line 39: We could use a fallback like `HOME=/root`.

ocaml/gencert/gencert.ml
```diff
@@ -15,18 +15,20 @@
...
 module Lib = Gencertlib.Lib
 module D = Debug.Make(struct let name = "gencert" end)
 
 let generate_cert_or_fail ~path ~cn =
-  let args = [path; cn] in
-  let (stdout, stderr) = Forkhelpers.execute_command_get_output generate_ssl_cert args in
-  let (stdout, stderr) = (String.escaped stdout, String.escaped stderr) in
+  let (stdout, stderr) = Lib.call_generate_ssl_cert ~args:[path; cn] in
+
+  let print_neatly prefix std_x =
```
lindig, line 25: This is actually writing to the debug log, so something like `debug_each_line` would be more descriptive.

PR #4114, opened 2020-05-28
CA-339329 firstboot scripts shouldn't sync DB when ugprading

lippirk:
There are 4 problematic scripts - we fix 2 of them here.

To preserve old behaviour of the firstboot scripts, we don't sync the
DB during an upgrade.

If we do sync the DB during an upgrade, then we get an error complaining
that incoming DBs have an incompatible schema version. This happens
because hosts are told to send the the minimally compliant Miami DB
- see pool_db_backup.ml.

### DIFF
--- a/scripts/generate-iscsi-iqn
+++ b/scripts/generate-iscsi-iqn
@@ -7,6 +7,9 @@ set -e
 export FIRSTBOOT_DATA_DIR=/etc/firstboot.d/data
 export XENSOURCE_INVENTORY=/etc/xensource-inventory
 
+UPGRADE="false"
+[ -r ${FIRSTBOOT_DATA_DIR}/host.conf ] && . ${FIRSTBOOT_DATA_DIR}/host.conf
+
 . ${XENSOURCE_INVENTORY}
 IQN_CONF=${FIRSTBOOT_DATA_DIR}/iqn.conf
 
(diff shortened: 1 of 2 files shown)


scripts/generate-iscsi-iqn
```diff
@@ -55,7 +58,10 @@ xe host-param-set uuid=${INSTALLATION_UUID} iscsi_iqn="${IQN}"
 echo Set iSCSI IQN: "${IQN}"
 echo IQN=\'"${IQN}"\' >${IQN_CONF}
 
-# Ensure changes are synced to disk
-xe pool-sync-database
+
+if [ ! "$UPGRADE" = "true" ]; then
```
lindig, line 62: Would it be clearer to use `[  "$UPGRADE" != "true" ]`?

lindig: Given that `UPGRADE` is set to `false` unconditionally above, does this need a test at all?

lippirk: yes, we need the check - `UPGRADE` potentially gets sourced by this:

```
[ -r ${FIRSTBOOT_DATA_DIR}/host.conf ] && . ${FIRSTBOOT_DATA_DIR}/host.conf
```

scripts/generate-iscsi-iqn
```diff
@@ -7,6 +7,9 @@ set -e
 export FIRSTBOOT_DATA_DIR=/etc/firstboot.d/data
 export XENSOURCE_INVENTORY=/etc/xensource-inventory
 
+UPGRADE="false"
+[ -r ${FIRSTBOOT_DATA_DIR}/host.conf ] && . ${FIRSTBOOT_DATA_DIR}/host.conf
```
psafont, line 11: Using `-r` instead of the usual `-e` seems to be useful as the executable flag in the file doesn't matter anyway. From the manpage:


> `.  filename [arguments]`
> `source filename [arguments]`
> Read and execute commands from filename in the current shell environment and return the exit status of the last command executed from filename.  If  filename  does     
> not contain a slash, filenames in PATH are used to find the directory containing filename.  **The file searched for in PATH need not be executable.**

PR #4119, opened 2020-06-01
Remove warnings

GeorgeBurac:


lindig:
This needs a proper rebase as there are unrelated changes in the commit

psafont:
Second commit has the actual changes (minus formatting), but it also has some unrelated changes.

ocaml/xapi-cli-protocol/tar_helpers.ml
```diff
@@ -0,0 +1,79 @@
+module Identity = struct
```
lippirk, line 1: it might be more appropriate to put `tar_helpers.ml` in the util library (i.e. move it to the util folder)

ocaml/xapi/helpers.ml
```diff
@@ -1424,4 +1424,4 @@ end = struct
       Backtrace.is_important e;
       D.error "Helpers.Stunnel.restart: failed to restart stunnel";
       raise e
-end
+end
```
lippirk, line 1427: it's preferable to not stage whitespace changes like this

ocaml/xapi/stream_vdi.ml
```diff
@@ -323,7 +323,7 @@ let verify_inline_checksum ifd checksum_table hdr =
     let csum = Bytes.make length' ' ' in
     Unixext.really_read ifd csum 0 length';
     let csum = Bytes.unsafe_to_string csum in
-    Tar_unix.Archive.skip ifd (Tar_unix.Header.compute_zero_padding_length hdr);
+    Tar_helpers.skip  ifd (Tar_unix.Header.compute_zero_padding_length hdr);
```
lippirk, line 326: you have some whitespace to fix around your usages of `Tar_helpers.skip `

lippirk:
you might be able to  remove `xapi_internal`'s dependency on `tar` and `tar-unix` now (see `ocaml/xapi/dune`)

personally i would have two commits - one fixing the datetime warnings, and the other fixing the tar warnings

thank you for fixing these warnings - they have been annoying us all for a long time!

ocaml/xapi-cli-protocol/tar_helpers.ml
GeorgeBurac: Okay

ocaml/xapi-cli-protocol/tar_helpers.ml
```diff
@@ -0,0 +1,77 @@
...
+module Writer = struct
+  type out_channel = Unix.file_descr
+  type 'a t = 'a Direct.t
+  (* XXX: there's no function to write directly from a bigarray *)
+  let really_write (fd : out_channel) buffer =
+    let s = Cstruct.to_string buffer |> Bytes.of_string in
+    if Bytes.length s > 0
+    then ignore (Unix.write fd s 0 (Bytes.length s))
+end
+module TarH = Tar.HeaderWriter(Direct)(Writer)
```
lindig, line 16: A better name would be `TarHeader`.

ocaml/xapi-cli-protocol/tar_helpers.ml
```diff
@@ -0,0 +1,77 @@
...
+
+module TarHelper : sig
+val skip : Unix.file_descr -> int -> unit
```
lindig, line 19: This should be indented. Maybe run this code through `ocamlformat`

ocaml/xapi-cli-protocol/tar_helpers.ml
```diff
@@ -0,0 +1,77 @@
...
+  type 'a t = 'a Direct.t
+  (* XXX: there's no function to write directly from a bigarray *)
+  let really_write (fd : out_channel) buffer =
+    let s = Cstruct.to_string buffer |> Bytes.of_string in
+    if Bytes.length s > 0
+    then ignore (Unix.write fd s 0 (Bytes.length s))
+end
+module TarH = Tar.HeaderWriter(Direct)(Writer)
+
+module TarHelper : sig
```
lindig, line 18: We now have `Tar_helpers.TarHelper`. Maybe introduce `tar_helpers.mli` and avoid a sub module here.

lindig:
A few general comments:

* The PR subject could be improved: Remove compiler warnings about Tar
* The commit subject of each commit should include a ticket number when there is one

ocaml/util/tar_helpers.ml
```diff
@@ -0,0 +1,80 @@
...
+  in
+  loop n
+
+let copy_n ifd ofd n =
+  let buffer = Bytes.create 16384 in
+  let rec loop remaining =
+  if remaining = 0L then ()
+  else (
+    let this = Int64.(to_int (min (of_int (Bytes.length buffer)) remaining)) in
+    let n = input (Unix.in_channel_of_descr ifd) buffer 0 this in
```
edwintorok, line 59: This creates a new input channel every time we go around the loop

ocaml/util/tar_helpers.ml
```diff
@@ -0,0 +1,80 @@
...
+    if n = 0 then 
+      raise End_of_file;
+    (
+      try
+        output (Unix.out_channel_of_descr ofd) buffer 0 n
```
edwintorok, line 64: Similarly this creates a new output channel every time we go around the loop. Who flushes it? Who closes it?
I think that the channels should be created once outside of the loop, and then at the end the out channel should be flushed, but even then since you're doing your own buffering here anyway could just use the Unix.read/Unix.write functions directly.

ocaml/util/tar_helpers.ml
GeorgeBurac: True. I don't know how I didn't spot that. Thank you so much, Edvin. :)

ocaml/util/tar_helpers.ml
```diff
@@ -0,0 +1,75 @@
...
+  really_input ifd s 0 (Cstruct.len buffer);
+  Cstruct.blit_from_bytes s 0 buffer 0 (Cstruct.len buffer)
+
+let skip ifd n =
+  let buffer = Cstruct.create 4096 in
+  let rec loop n =
+    if n <= 0 then ()
+    else (
+      let amount = min n (Cstruct.len buffer) in
+      really_read ifd (Cstruct.sub buffer 0 amount);
```
edwintorok, line 47: And lseek would've been more efficient here, but that only works on files and not pipes/sockets. It looks like we need to deal with streams, so we can't use lseek.
This could be optimized so we avoid using a Cstruct buffer, and allocate just a temporary bytes only once outside the loop, but we can leave optimizations for some other time.

ocaml/util/tar_helpers.ml
```diff
@@ -0,0 +1,75 @@
...
+  let return x = x
+  let ( >>= ) m f = f m
+end
+
+module Writer = struct
+  type out_channel = Unix.file_descr
+  type 'a t = 'a Identity.t
+
+  let really_write fd buffer =
+    let s = Cstruct.to_string buffer |> Bytes.of_string in
```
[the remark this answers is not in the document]
GeorgeBurac: Thank you! :)

PR #4120, opened 2020-06-01
CA-340484 bump xapi version in DB after upgrade/update

lippirk:
We got round to bumping the hardcoded API version in b81af6e, but
the API version is also stored in the DB, which was not changed.
This was causing problems with the firstboot create-guest-template
service since ultimately it gets a host's API version from the DB,
and then uses that version to perform a template import, leading to
an IMPORT_INCOMPATIBLE_VERSION error.

### DIFF
--- a/ocaml/xapi/create_misc.mli
+++ b/ocaml/xapi/create_misc.mli
@@ -44,4 +44,4 @@ val create_host_cpu : __context:Context.t -> host_info -> unit
 val create_pool_cpuinfo : __context:Context.t -> unit
 val create_chipset_info : __context:Context.t -> host_info -> unit
 val create_updates_requiring_reboot_info : __context:Context.t -> host:[`host] Ref.t -> unit
-val get_xapi_verstring : unit -> string
+val get_xapi_version : unit -> string
--- a/ocaml/xapi/create_misc.mli
+++ b/ocaml/xapi/create_misc.mli
@@ -44,3 +44,4 @@ val create_host_cpu : __context:Context.t -> host_info -> unit
 val create_pool_cpuinfo : __context:Context.t -> unit
 val create_chipset_info : __context:Context.t -> host_info -> unit
 val create_updates_requiring_reboot_info : __context:Context.t -> host:[`host] Ref.t -> unit
+val get_xapi_verstring : unit -> string
--- a/ocaml/xapi/dbsync_slave.ml
+++ b/ocaml/xapi/dbsync_slave.ml
@@ -89,10 +89,10 @@ let refresh_localhost_info ~__context info =
   let old_software_version = Db.Host.get_software_version ~__context ~self:host in
   match List.assoc_opt "xapi" old_software_version with
   | None              -> D.error "host is missing xapi version. ref = %s" (Ref.string_of host)
-  | Some xapi_version -> (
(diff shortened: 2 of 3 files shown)


lippirk:
testing in progress

ocaml/xapi/xapi_db_upgrade.ml
```diff
@@ -612,6 +612,17 @@ let remove_legacy_ssl_support = {
   fn=fun ~__context -> Db.Host.get_all ~__context |> List.iter (fun self -> Db.Host.set_ssl_legacy ~__context ~self ~value:false)
 }
 
+let update_xapi_version = {
+  description = "Bump API version";
+  version=(fun _ -> true);
+  fn=fun ~__context ->
+    Db.Host.get_all ~__context |>
+    List.iter (fun self ->
+      Db.Host.set_API_version_major ~__context ~self ~value:(Int64.of_int Constants.version_major);
```
edwintorok, line 621: What happens during an RPU when the master got upgraded, but the slave hasn't yet? This looks more like a pool version than a host version now.
If this causes other problems could we set the API version on toolstack startup just for the current host? That'd give a better picture...

edwintorok: Dbsync_slave refresh_localhost_info already has some code to do this refresh (and apparently runs on master too).
Is the problem that the import code runs too soon, before that version is updated?

lippirk:
as @psafont pointed out, i was updating the wrong version, so I've fixed that now

i moved the version update to `refresh_localhost_info`. `create-guest-templates.service` ought run after this, because it waits for `xapi-init-complete.service`

ocaml/xapi/create_misc.mli
```diff
@@ -44,3 +44,4 @@ val create_host_cpu : __context:Context.t -> host_info -> unit
 val create_pool_cpuinfo : __context:Context.t -> unit
 val create_chipset_info : __context:Context.t -> host_info -> unit
 val create_updates_requiring_reboot_info : __context:Context.t -> host:[`host] Ref.t -> unit
+val get_xapi_verstring : unit -> string
```
lindig, line 47: `get_xapi_verstring` - the fact that this is a string is covered by the type, so I would prefer `get_xapi_version`. This will lead to clearer code below because "version" will be used more consistently.

ocaml/xapi/dbsync_slave.ml
```diff
@@ -84,6 +84,18 @@ let refresh_localhost_info ~__context info =
...
+
+  (* update xapi version *)
+  let old_software_version = Db.Host.get_software_version ~__context ~self:host in
+  match List.assoc_opt "xapi" old_software_version with
+  | None              -> D.error "host is missing xapi version. ref = %s" (Ref.string_of host)
+  | Some old_xapi_version -> (
+    let current_xapi_version = Create_misc.get_xapi_version () in
+    if old_xapi_version <> current_xapi_version then
+      let value = ("xapi", current_xapi_version) :: List.remove_assoc "xapi" old_software_version in
+      Db.Host.set_software_version ~__context ~self:host ~value
```
edwintorok, line 96: Hmm, didn't Create_misc.create_software_version above already call this?

lippirk:
this is the wrong fix, closing

PR #4123, opened 2020-06-02
CP-33774 PSR orchestration

lippirk:
We implement a function which orchestrates a pool secret rotation.
We handle failure by splitting the PSR into idempotent chunks, and
saving checkpoints at the end of each chunk. This will allow us to
resume from failed PSRs, and importantly reduce the chance of hosts
becoming locked out with the wrong pool secret.

### DIFF
--- a/ocaml/tests/suite_alcotest.ml
+++ b/ocaml/tests/suite_alcotest.ml
@@ -79,4 +79,5 @@ let () =
        ; Test_extauth_plugin_ADpbis.tests
        ; Test_helpers.tests
        ; Test_datamodel_utils.tests
+       ; Test_psr.tests
        ])
--- a/ocaml/tests/suite_alcotest.ml
+++ b/ocaml/tests/suite_alcotest.ml
@@ -4,77 +4,79 @@ let () =
      so we will probably not exceed the 4MB limit in Travis *)
   Debug.log_to_stdout () ;
   Alcotest.run "Base suite"
-    ([
-       ("Test_valid_ref_list", Test_valid_ref_list.test)
-     ; ("Test_sdn_controller", Test_sdn_controller.test)
-     ; ("Test_pci_helpers", Test_pci_helpers.test)
-     ; ("Test_vdi_allowed_operations", Test_vdi_allowed_operations.test)
-     ; ("Test_sr_allowed_operations", Test_sr_allowed_operations.test)
-     ; ("Test_vm_migrate", Test_vm_migrate.test)
-     ; ("Test_no_migrate", Test_no_migrate.test)
-     ; ("Test_vm_check_operation_error", Test_vm_check_operation_error.test)
-     ; ("Test_vm_helpers", Test_vm_helpers.test)
-     ; ("Test_xapi_vbd_helpers", Test_xapi_vbd_helpers.test)
-     ; ("Test_host", Test_host.test)
-     ; ("Test_host_helpers", Test_host_helpers.test)
-     ; ("Test_vdi_cbt", Test_vdi_cbt.test)
-     ; ("Test_sr_update_vdis", Test_sr_update_vdis.test)
-     ; ("Test_xapi_db_upgrade", Test_xapi_db_upgrade.test)
-     ; ("Test_db_lowlevel", Test_db_lowlevel.test)
-     ; ("Test_vlan", Test_vlan.test)
-     ; ("Test_network", Test_network.test)
-     ; ("Test_network_sriov", Test_network_sriov.test)
-     ; ("Test_bond", Test_bond.test)
-     ; ("Test_tunnel", Test_tunnel.test)
-     ; ("Test_agility", Test_agility.test)
-     ; ("Test_daemon_manager", Test_daemon_manager.test)
-     ; ("Test_cluster", Test_cluster.test)
-     ; ("Test_cluster_host", Test_cluster_host.test)
-     ; ("Test_clustering", Test_clustering.test)
-     ; ( "Test_clustering_allowed_operations"
-       , Test_clustering_allowed_operations.test )
-     ; ("Test_client", Test_client.test)
-     ; ("Test_ca91480", Test_ca91480.test)
-     ; ("Test_pgpu", Test_pgpu.test)
-     ; ("Test_gpu_group", Test_gpu_group.test)
-     ; ("Test_pool_apply_edition", Test_pool_apply_edition.test)
-     ; ("Test_pool_update", Test_pool_update.test)
-     ; ("Test_pool_db_backup", Test_pool_db_backup.test)
-     ; ("Test_pool_restore_database", Test_pool_restore_database.test)
-     ; ("Test_workload_balancing", Test_workload_balancing.test)
-     ; ("Test_pusb", Test_pusb.test)
-     ; ("Test_pvs_site", Test_pvs_site.test)
-     ; ("Test_pvs_proxy", Test_pvs_proxy.test)
-     ; ("Test_pvs_server", Test_pvs_server.test)
-     ; ("Test_event", Test_event.test)
-     ; ("Test_vm_placement", Test_vm_placement.test)
-     ; ("Test_vm_memory_constraints", Test_vm_memory_constraints.test)
-     ; ("Test_xapi_xenops", Test_xapi_xenops.test)
-     ; ("Test_network_event_loop", Test_network_event_loop.test)
-     ; ("Test_vgpu_type", Test_vgpu_type.test)
-     ; ("Test_storage_migrate_state", Test_storage_migrate_state.test)
-     ; ("Test_bios_strings", Test_bios_strings.test)
-     ; ("Test_certificates", Test_certificates.test)
-     ]
-    @ Test_guest_agent.tests
-    @ Test_nm.tests
-    @ Test_xenopsd_metadata.tests
-    @ Test_http.tests
-    @ Test_ha_vm_failover.tests
-    @ Test_map_check.tests
-    @ Test_pool_license.tests
-    @ Test_features.tests
-    @ Test_platformdata.tests
-    @ Test_sm_features.tests
-    @ Test_vgpu_type.tests
-    @ Test_pgpu_helpers.tests
-    @ Test_storage_migrate_state.tests
-    @ Test_vm.tests
-    @ Test_cpuid_helpers.tests
-    @ Test_pool_cpuinfo.tests
-    @ Test_dbsync_master.tests
-    @ Test_pvs_cache_storage.tests
-    @ Test_extauth_plugin_ADpbis.tests
-    @ Test_helpers.tests
-    @ Test_datamodel_utils.tests
-    )
+    (List.concat
+       [
+         [
+           ("Test_valid_ref_list", Test_valid_ref_list.test)
+         ; ("Test_sdn_controller", Test_sdn_controller.test)
+         ; ("Test_pci_helpers", Test_pci_helpers.test)
+         ; ("Test_vdi_allowed_operations", Test_vdi_allowed_operations.test)
+         ; ("Test_sr_allowed_operations", Test_sr_allowed_operations.test)
+         ; ("Test_vm_migrate", Test_vm_migrate.test)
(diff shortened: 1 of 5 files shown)


ocaml/tests/suite_alcotest.ml
```diff
@@ -75,4 +75,5 @@ let () =
       @ Test_pvs_cache_storage.tests
       @ Test_extauth_plugin_ADpbis.tests
       @ Test_helpers.tests
-      @ Test_datamodel_utils.tests);
+      @ Test_datamodel_utils.tests
+      @ Test_psr.tests );
```
lindig: It doesn't matter much here, but `List.concat` would be more efficient. We also had some nasty surprised when using `@` together with if-then-else.

ocaml/tests/test_psr.ml
```diff
@@ -0,0 +1,122 @@
...
+(* we don't make the distinction between a slave & master here.
+   it is intended that whilst the master orchestrates the procedure
+   in the real implementation, the way the master should tell itself
+   to update the PS is no different to the way the master should tell
+   a slave *)
+module Impl = struct
+  type ps = string
+  type pss = (ps * ps)
+  type fistpoint = No_fp
+                 | Accept_new_ps
```
lindig: The module uses an acronym (psr) and `ps` is used here, too. Looking from the outside it might be good to spell out `pool_secret` somewhere.

ocaml/tests/test_psr.ml
```diff
@@ -0,0 +1,142 @@
...
+  let save_checkpoint x = checkpoint := Some x
+
+  let retrieve_checkpoint () = !checkpoint
+
+  let backup pss = backups := Some pss
+
+  let retrieve () = Option.get !backups
+
+  let tell_accept_new_ps (_, staged_ps) h =
+    if !h.fistpoint = Accept_new_ps then
```
lindig, line 45: How about
```
match !h.fistpoint with
| Accept_new_ps ->
| .. ->
```

I think this is easier to read and works better with algebraic types.

ocaml/tests/test_psr.ml
```diff
@@ -0,0 +1,142 @@
...
+  faulty_host := {!faulty_host with fistpoint} ;
+  let r = PSR.start (default_ps, new_ps) hosts in
+  Alcotest.check psr_r "PSR should fail" r err ;
+  faulty_host := {!faulty_host with fistpoint= No_fp} ;
+  let r = PSR.start (default_ps, "not_used") hosts in
+  check_psr_success r new_ps hosts
+
+let test_fails_twice () =
+  let hosts = Impl.mk_hosts 3 in
+  let faulty_host_1 = List.nth hosts 2 in
```
lindig, line 111: Slightly confusing to select host 2 but call it host 1.

ocaml/tests/test_psr.ml
```diff
@@ -0,0 +1,142 @@
+let checkpoint = ref None
+
+let backups = ref None
+
+let default_ps = "default_ps"
```
lindig, line 5: The code uses acronyms in identifiers for modules, types, and values: ps and cp. Looking from the outside, this can be hard to read. I think some more explicit names for modules would help the reader.

ocaml/tests/test_psr.ml
lippirk: i mentioned the abbreviations in the mli file, but you are right - this is an indication i should change the name of the type

ocaml/tests/test_psr.ml
```diff
@@ -0,0 +1,159 @@
...
+          ; fistpoint= No_fistpoint
+          })
+
+  let save_checkpoint x = checkpoint := Some x
+
+  let retrieve_checkpoint () = !checkpoint
+
+  let backup pool_secrets = backups := Some pool_secrets
+
+  let retrieve () = Option.get !backups
```
psafont, line 52: This will raise an exception if it gets called before `let backup` is called, is it not better to return an option so the caller can deal with it?

lippirk: i wrote `val retrieve : unit -> pool_secrets option` originally, but it actually doesn't make much sense. if we can't fetch the pool secret from the backup that's a fatal error (i.e. the implementor should handle that). the reason it's a fatal error is that we fetch from the backup only when there is a checkpoint to resume from (you'll see that the function to get the checkpoint returns an option)

there are racy, and i ran into this when writing the tests :-1: i think this is somewhat useful here because i do want to make sure that the pool secret backup is cleaned up (for example). i'm not sure what i'd use as an alternative to model writing/reading to a file?

ocaml/xapi/xapi_psr.ml
```diff
@@ -0,0 +1,143 @@
...
+    type checkpoint =
+      | No_checkpoint
+      | Backup
+      | Accept_new_pool_secret
+      | Send_new_pool_secret
+      | Finish
+
+    exception Cannot_parse_checkpoint of string
+
+    let string_of_checkpoint = function
```
edwintorok, line 45: You can use the RPC derivers to generate serialization and deserialization functions for you, avoiding all the boilerplate.

ocaml/xapi/xapi_psr.ml
```diff
@@ -0,0 +1,143 @@
...
+            try
+              List.iter (tell_accept_new_pool_secret pool_secrets) hosts ;
+              None
+            with e -> Some Failed_during_accept_new_pool_secret
+          in
+          match res with
+          | None ->
+              (go [@tailcall]) pool_secrets hosts Send_new_pool_secret
+          | Some e ->
+              save_checkpoint (string_of_checkpoint Accept_new_pool_secret) ;
```
edwintorok, line 100: I think the checkpoint should be saved when you enter the go function, prior to doing the matching, that way you don't need all these match None/Some on each state, and also you would save a checkpoint even if xapi or the host crashes in the middle of running the operation.

As for implementing the save checkpoint: perhaps the redolog code can be used here? That already has necessary logic for recovering from failures, etc. We could require the redolog to be enabled (which doens't necessarily mean that HA has to be enabled, in fact HA should be disabled...) before doing a PSR

lippirk:
unfortunately the tests have become more complicated because the slaves and the master are now distinguished (the reason for doing this was that i found in the real implementation it matters what order you cleanup the hosts after the rotation has finished). i also increased the coverage by testing more fistpoint combinations

ocaml/tests/test_psr.ml
```diff
@@ -0,0 +1,350 @@
...
+      Printf.sprintf "Failed_during_accept_new_pool_secret: %s"
+        (string_of_int x)
+  | Failed_during_send_new_pool_secret x ->
+      Printf.sprintf "Failed_during_send_new_pool_secret: %s" (string_of_int x)
+  | Failed_during_cleanup_slave x ->
+      Printf.sprintf "Failed_during_cleanup_slave: %s" (string_of_int x)
+  | Failed_during_cleanup_master x ->
+      Printf.sprintf "Failed_during_cleanup_master: %s" (string_of_int x)
+
+let map_r f = function
```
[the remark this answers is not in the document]
lippirk: i added an extra variant in my other PR (#4127) - `UnknownError of exn`, which makes it harder to model with result (but i agree in its form here to use `result` would definitely have been better)

ocaml/tests/test_psr.ml
```diff
@@ -0,0 +1,350 @@
...
+
+let new_pool_secret = "new_pool_secret"
+
+(* we could also test with a fistpoint 'During' the computation,
+   but this adds even more complexity to the test *)
+type fistpoint_time = Before | After
+
+let string_of_fistpoint_time = function Before -> "Before" | After -> "After"
+
+type firstpoint_place =
```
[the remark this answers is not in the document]
lippirk: i renamed this and added a comment to make it clearer

ocaml/xapi/xapi_psr.mli
```diff
@@ -0,0 +1,60 @@
+(** psr = pool secret rotation
+  * ps = pool secret = ptoken *)
+
+type 'a r =
+  | Success
+  | Failed_during_accept_new_pool_secret of 'a
+  | Failed_during_send_new_pool_secret of 'a
+  | Failed_during_cleanup_slave of 'a
+  (* very unlikely - user needs to cleanup manually *)
```
robhoes, line 9: We can't tell users to delete files in dom0. We should just keep trying, for example on startup.

ocaml/xapi/xapi_psr.ml
```diff
@@ -0,0 +1,143 @@
+(** important invariants:
+  * given a list of hosts containing both the master and
+    the slaves, the master will always be at the head *)
+module D = Debug.Make (struct
+  let name = "xapi_psr"
+end)
+
+type 'a r =
```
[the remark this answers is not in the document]
robhoes: I think that it is a good idea to make use of the `result` type for this kind of thing. This is now the standard. You could then also replace the use of exceptions below with a monadic style around `result` (see Rresult).

ocaml/xapi/xapi_psr.ml
```diff
@@ -0,0 +1,143 @@
...
+
+  val cleanup_master : unit -> unit
+end
+
+module Make =
+functor
+  (Impl : Impl)
+  ->
+  struct
+    open Impl
```
robhoes, line 44: I think that not opening will make the code clearer.

ocaml/xapi/xapi_psr.ml
```diff
@@ -0,0 +1,143 @@
...
+                     D.error
+                       "failed while telling hosts to send new pool secret. \
+                        error= %s"
+                       (Printexc.to_string e) ;
+                     raise (R (Failed_during_send_new_pool_secret host)))
+          in
+          (go [@tailcall]) pool_secrets hosts Cleanup_slaves
+      | Cleanup_slaves ->
+          save_checkpoint (string_of_checkpoint Cleanup_slaves) ;
+          let[@warning "-8"] (master :: slaves) = hosts in
```
robhoes, line 107: Perhaps have `master` and `slaves` as arguments to `go`, so you don't need this?

ocaml/xapi/xapi_psr.mli
lippirk: :heart: that will make the code simpler

lippirk:
have addressed the comments, notable changes:

 - use result type
 - slave -> member
 - put master cleanup in a retry loop (this also changes the unit test a little)

ocaml/xapi/xapi_psr.ml
```diff
@@ -0,0 +1,157 @@
...
+                   Ok ()
+                 with e ->
+                   D.error "failed while telling hosts to cleanup. error= %s"
+                     (Printexc.to_string e) ;
+                   Error (Failed_during_cleanup_member, member))
+          >>= fun () ->
+          (go [@tailcall]) pool_secrets master members Cleanup_master
+      | Cleanup_master ->
+          Impl.save_checkpoint (string_of_checkpoint Cleanup_master) ;
+          Helpers.retry_forever ~interval:1. ~doc:"cleanup master after PSR"
```
robhoes, line 140: This means that the API call will potentially hang forever, and the caller will not be able to cancel it. What is the consequence of this failure state? Everything works with the new token, but the master still accepts the old one? It may be better to treat it like the other failure cases: fail with and error describing the situation and let the user retry the call.

ocaml/tests/test_psr.ml
```diff
@@ -0,0 +1,340 @@
...
+      (option (pair string string))
+      "pool secret backups were cleaned up" None psr_state.pool_secret_backups)
+
+(* we test almost all combinations of hosts and fistpoints. the only
+   case we don't test is one case of master cleanup failure,
+   since it is slightly different - it has its own unit test (see below).
+ *)
+let almost_all_possible_fists ~num_hosts =
+  let multiply fp_times fp_actions host_ids =
+    List.map
```
psafont, line 241: Maybe next time we'll be able to use `List.concat_map`...

lippirk: :crossed_fingers:

PR #4124, opened 2020-06-04
Technical debt round

psafont:
These changes are
- removing stdext usages (now only xapi_internal uses it directly)
- making apparent all dependencies on the dune metadata (it's quite staggering the amount of libraries used)
- adding depexts for fedora
- remove most warnings when not compiling with `--profile=release`, these found when running `dune utop`.


mseri:
What does prevent to get rid of sexplib and just use sexplib0?

psafont:
> What does prevent to get rid of sexplib and just use sexplib0?

This doesn't seem available in sexplib0:
```
File "ocaml/database/schema.ml", line 14, characters 5-16:
14 | open Sexplib.Std
          ^^^^^^^^^^^
```

Maybe it's easily solvable, but I haven't investigated it.

mseri:
In 90% of the cases, which may mean that it is not enough in this case, you can replace

```ocaml
open Sexplib.Std
open Sexplib.Conv
```
with
```ocaml
open Sexplib0.Sexp_conv
```

psafont:
Couldn't remove it from the xe server as it's deserializing s-exprs if backtraces from strings, something which doesn't seem available in sexplib0.

psafont:
> Have you tested that the string splitting split_f produces the same results? Those usually have corner cases around handling of empty, or repeated delimiters.

Yes, the cpuid tests exercise the code and found the difference in behaviour. I couldn't find any other difference when manually testing it.

PR #4125, opened 2020-06-09
CA-340776: disconnect from stunnel cleanly

edwintorok:
Needs testing


edwintorok:
Manually building an 'xe' and copying it to Dom0 of 2 hosts in a pool worked.

psafont:
This should also be integrated into the `1.249.0-lcm` branch

edwintorok:
There are still some bugs in xe, when running with strace I've seen some EBADF (both from old and new xe), let me see if those are easy to fix as well.

PR #4127, opened 2020-06-10
CP-33777- Real PSR implementation

lippirk:
Using the interface defined in https://github.com/xapi-project/xen-api/pull/4123, we implement pool secret rotation for real hosts.

### DIFF
--- a/ocaml/idl/datamodel_errors.ml
+++ b/ocaml/idl/datamodel_errors.ml
@@ -452,7 +452,7 @@ let _ =
   error Api_errors.host_master_cannot_talk_back [ "ip" ]
     ~doc:"The master reports that it cannot talk back to the slave on the supplied management IP address." ();
   error Api_errors.host_unknown_to_master [ "host" ]
-    ~doc:"The master says the host is not known to it. Perhaps the Host was deleted from the master's database? Perhaps the slave is pointing to the wrong master?" ();
+    ~doc:"The master says the host is not known to it. Perhaps the Host was deleted from the master's database, the slave is pointing to the wrong master, or a pool secret rotation failed?" ();
   error Api_errors.host_broken []
     ~doc:"This server failed in the middle of an automatic failover operation and needs to retry the failover action." ();
   error Api_errors.host_has_resident_vms [ "host" ]
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -1325,6 +1325,41 @@ let host_query_ha = call ~flags:[`Session]
     ~allowed_roles:_R_LOCAL_ROOT_ONLY
     ()
 
+  let notify_accept_new_pool_secret = call
+    ~name:"notify_accept_new_pool_secret"
+    ~lifecycle:[Published, rel_next, ""]
+    ~doc:"Notifies host that they must begin accepting requests containing the new pool secret"
+    ~params:[
+      Ref _host, "host", "The host";
+      SecretString, "new_ps", "New pool secret"
+    ]
+    ~allowed_roles:_R_LOCAL_ROOT_ONLY
+    ~hide_from_docs:true
+    ()
+
+  let notify_send_new_pool_secret = call
+    ~name:"notify_send_new_pool_secret"
+    ~lifecycle:[Published, rel_next, ""]
+    ~doc:"Notifies host that they must begin sending requests with the new pool secret"
+    ~params:[
+      Ref _host, "host", "The host";
+      SecretString, "new_ps", "New pool secret"
+    ]
+    ~allowed_roles:_R_LOCAL_ROOT_ONLY
+    ~hide_from_docs:true
+    ()
+
+  let cleanup_pool_secret = call
+    ~name:"cleanup_pool_secret"
+    ~lifecycle:[Published, rel_next, ""]
+    ~doc:"Cleanup old pool secret on recipient host"
+    ~params:[
+      Ref _host, "host", "The host"
+    ]
+    ~allowed_roles:_R_LOCAL_ROOT_ONLY
+    ~hide_from_docs:true
+    ()
+
   (** Hosts *)
   let t =
     create_obj ~in_db:true ~in_product_since:rel_rio ~in_oss_since:oss_since_303 ~internal_deprecated_since:None ~persist:PersistEverything ~gen_constructor_destructor:false ~name:_host ~descr:"A physical host" ~gen_events:true
@@ -1443,6 +1478,9 @@ let host_query_ha = call ~flags:[`Session]
         set_iscsi_iqn;
         set_multipathing;
         set_uefi_certificates;
+        notify_accept_new_pool_secret;
+        notify_send_new_pool_secret;
+        cleanup_pool_secret;
       ]
       ~contents:
         ([ uid _host;
--- a/ocaml/ptoken/lib.ml
+++ b/ocaml/ptoken/lib.ml
@@ -1,0 +1,4 @@
+let gen_token () =
+  let uuid () = Uuid.to_string (Uuid.make_uuid ()) in
+  let uuids = String.concat "/" [uuid (); uuid (); uuid ()] in
+  SecretString.of_string uuids
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -409,7 +409,7 @@ let call_api_functions_internal ~__context f =
   let do_master_login () =
     let session =
       Client.Client.Session.slave_login rpc (get_localhost ~__context)
-        !Xapi_globs.pool_secret
+        (Xapi_globs.pool_secret ())
     in
     require_explicit_logout := true ;
     session
@@ -455,7 +455,7 @@ let call_emergency_mode_functions hostname f =
     XMLRPC_protocol.rpc ~srcstr:"xapi" ~dststr:"xapi" ~transport ~http
   in
   let session_id =
-    Client.Client.Session.slave_local_login rpc !Xapi_globs.pool_secret
+    Client.Client.Session.slave_local_login rpc (Xapi_globs.pool_secret ())
   in
   finally
     (fun () -> f rpc session_id)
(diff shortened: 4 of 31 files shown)


lippirk:
This PR currently includes all the commits in #4123, but it can still be reviewed separately since #4123 is mainly unit tests

ocaml/idl/datamodel_host.ml
```diff
@@ -1325,6 +1325,37 @@ let host_query_ha = call ~flags:[`Session]
     ~allowed_roles:_R_LOCAL_ROOT_ONLY
     ()
 
+  let notify_accept_new_pool_secret = call
+    ~name:"notify_accept_new_pool_secret"
+    ~lifecycle:[Published, rel_stockholm, ""]
```
robhoes, line 1330: `rel_next`

ocaml/idl/datamodel_host.ml
```diff
@@ -1325,6 +1325,37 @@ let host_query_ha = call ~flags:[`Session]
     ~allowed_roles:_R_LOCAL_ROOT_ONLY
     ()
 
+  let notify_accept_new_pool_secret = call
```
robhoes, line 1328: As discussed, let's add `hide_from_docs:true` to all of these, as they are internal calls.

ocaml/xapi/message_forwarding.ml
```diff
@@ -3361,6 +3365,28 @@ functor
...
+        (* we do not need to verify the old pool secret here, since it has
+           already been verified (see xapi_http.assert_credentials_ok) *)
+        info "Host.notify_accept_new_pool_secret: host='%s'"
+          (host_uuid ~__context host) ;
+        let local_fn = Local.Host.notify_accept_new_pool_secret ~host ~new_ps in
+        do_op_on ~__context ~host ~local_fn (fun session_id rpc ->
+            Client.Host.notify_accept_new_pool_secret rpc session_id host new_ps)
+
+      let notify_send_new_pool_secret ~__context ~host =
+        info "Host.notify_accept_new_pool_secret: host='%s'"
```
robhoes, line 3379: accept -> send

ocaml/xapi/xapi_http.ml
```diff
@@ -130,8 +130,16 @@ let assert_credentials_ok realm ?(http_action = realm) ?(fn = Rbac.nofn)
     | None, Some pool_secret, _
       when SecretString.equal pool_secret !Xapi_globs.pool_secret ->
         fn ()
-    | None, Some _, _ ->
-        raise (Http.Unauthorised realm)
+    | None, Some pool_secret, _ -> (
+      (* if a pool secret rotation is in progress, we accept multiple pool secrets *)
+      match !Xapi_globs.psr_pool_secrets with
```
robhoes, line 135: Rather than having both `Xapi_globs.pool_secret` and `Xapi_globs.psr_pool_secrets` for almost the same purpose, could we change the former to `Xapi_globs.pool_secrets` and make it a list of all currently accepted secrets?

ocaml/xapi/xapi_globs.ml
```diff
@@ -34,7 +34,13 @@ let _ =
...
       exit restart_return_code
 
-let pool_secret = ref (SecretString.of_string "")
+(* - this will usually be a singleton list.
+   - during pool secret rotation it will contain multiple elements,
+     but the tail will be dropped when it has completed.
+   - the head is always the pool secret that should be sent in requests *)
+let pool_secrets : SecretString.t list ref = ref []
+
+let pool_secret () = List.hd !pool_secrets
```
robhoes, line 43: Please avoid `List.hd` and instead raise a descriptive error if the list is empty.

ocaml/xapi/helpers.ml
```diff
@@ -860,26 +860,24 @@ let pool_has_different_host_platform_versions ~__context =
...
+      let ps =
+        Unixext.string_of_file !Xapi_globs.pool_secret_path
+        |> SecretString.of_string
+      in
+      pool_secrets := [ps] ;
+      Db_globs.pool_secret :=
+        ps |> SecretString.rpc_of_t |> Db_secret_string.t_of_rpc
+    with _ ->
+      (* No pool secret exists. *)
+      let module Ptoken = Genptokenlib.Lib in
```
psafont, line 874: `pool_secret` used to be updated in this branch, `pool_secrets` is not. It's not clear to me why this is the case

ocaml/xapi/xapi_psr.ml
```diff
@@ -15,9 +15,12 @@
 (** important invariants:
   * given a list of hosts containing both the master and
     the members, the master will always be at the head *)
-module D = Debug.Make (struct
-  let name = "xapi_psr"
-end)
+
+module D = Debug.Make (struct let name = "xapi_psr" end)
+
+module Unixext = Stdext.Unixext
```
psafont, line 21: I'm trying to drop `Stdext`, Please use `Xapi_stdext_unix.Unixext` instead

ocaml/xapi/xapi_psr.ml
```diff
@@ -26,6 +29,23 @@ type failure =
...
+let user_facing_error_message ~__context error =
+  let failure, self = error in
+  let host_name = Db.Host.get_name_label ~__context ~self in
+  let try_again_msg = "Please check the management network and try again." in
+  match failure with
+  | Failed_during_accept_new_pool_secret ->
+      Printf.sprintf "%s refused to start accepting new pool secret. %s"
+        host_name try_again_msg
+  | Failed_during_send_new_pool_secret ->
+      Printf.sprintf "%s refused to start sending new pool secret. %s" host_name
```
psafont, line 41: The messages seem too convoluted. I think they can be made simpler while not losing meaning, like `%s refused the new pool secret. %s` and `%s refused to send a new pool secret. %s`.

ocaml/xapi/xapi_psr_util.ml
```diff
@@ -0,0 +1,40 @@
...
+        [old_pool_secret_backup_path; new_pool_secret_backup_path]
+        |> List.map (fun path ->
+               Unixext.string_of_file path |> SecretString.of_string)
+      in
+      (* be careful to load them in the correct order *)
+      Xapi_globs.pool_secrets :=
+        if SecretString.equal (Xapi_globs.pool_secret ()) (List.hd secrets) then
+          secrets
+        else
+          List.rev secrets
```
psafont, line 40: I would rather add a bit of repetition when processing pool secret paths and keep both secrets in separate variables rather than having to do this strange handling.

ocaml/xapi/xapi_psr.ml
lippirk: 'refused the new pool secret' implies that the host refused some request because of a pool secret, but that isn't what is going on here. What I mean is that the operation to tell the host to start accepting the new pool secret failed (and similarly for the other message).

I'll change it to `operation to tell $host to begin accepting new pool secret failed`

ocaml/idl/datamodel_errors.ml
```diff
@@ -452,7 +452,7 @@ let _ =
   error Api_errors.host_master_cannot_talk_back [ "ip" ]
     ~doc:"The master reports that it cannot talk back to the slave on the supplied management IP address." ();
   error Api_errors.host_unknown_to_master [ "host" ]
-    ~doc:"The master says the host is not known to it. Perhaps the Host was deleted from the master's database? Perhaps the slave is pointing to the wrong master?" ();
+    ~doc:"The master says the host is not known to it. Perhaps the Host was deleted from the master's database? Perhaps the slave is pointing to the wrong master? Perhaps a pool secret rotation failed?" ();
```
lindig, line 455: The repeated use of `Perhaps` could be replaced by `Or`

ocaml/ptoken/lib.ml
```diff
@@ -0,0 +1,4 @@
+let gen_token () =
+  let uuid _ = Uuid.to_string (Uuid.make_uuid ()) in
```
lindig, line 2: Minor: but why not use `()` as argument for `uuid`?

ocaml/xapi/xapi_psr.ml
```diff
@@ -145,11 +167,214 @@ functor
...
+  if List.length !Xapi_globs.pool_secrets <= 1 then
+    D.info "xapi_psr.ml:cleanup_internal: already cleaned up"
+  else
+    (* if we don't remove pool secret backups, they will be loaded into memory when xapi starts *)
+    let files_to_remove =
+      List.append additional_files_to_remove
+        [old_pool_secret_backup_path; new_pool_secret_backup_path]
+    in
+    List.iter
+      (fun path -> if Sys.file_exists path then Sys.remove path)
```
lindig, line 199: In general it is better to avoid a test and just do the action and deal with the failure - this avoids race conditions.

ocaml/xapi/xapi_psr.ml
```diff
@@ -145,11 +167,214 @@ functor
...
+            Unix.write_substring fd checkpoint 0 (String.length checkpoint)
+          in
+          ())
+
+    let retrieve_checkpoint () =
+      if not (Sys.file_exists checkpoint_path) then
+        None
+      else
+        let checkpoints = Unixext.read_lines ~path:checkpoint_path in
+        match checkpoints with
```
lindig, line 231: I think it's better to do the `Unixext.read_lines ~path:checkpoint_path ` inside the match and also match exceptions. This could include handling that the file does not exist. There is an unlikely race here between the test earlier and the read.

ocaml/xapi/xapi_psr.ml
```diff
@@ -145,11 +167,214 @@ functor
...
+    let cleanup_master =
+      cleanup_internal
+        ~additional_files_to_remove:[checkpoint_path; checkpoint_tmp_path]
   end
 
-let start ~__context = failwith "not impl"
+let notify_new ~new_ps =
+  match !Xapi_globs.pool_secrets with
+  | _ :: _ :: _ ->
+      D.info "xapi_psr.ml:notify_new: already accepting new_ps"
```
lindig, line 278: I'd spell out `new_ps` in a log gmessage

ocaml/xapi/xapi_psr.ml
```diff
@@ -145,11 +167,214 @@ functor
...
+          Unixext.with_file new_pool_secret_backup_path [Unix.O_RDONLY] perm
+            (fun ifd ->
+              Unixext.with_file !Xapi_globs.pool_secret_path
+                [Unix.O_WRONLY] perm (fun ofd ->
+                  try
+                    let (_ : int64) = Unixext.copy_file ifd ofd in
+                    ()
+                  with e ->
+                    D.error
+                      "xapi_psr.ml:notify_send: copy pool_secret_path failed" ;
```
lindig, line 306: Should the path be included in the message?

ocaml/xapi/xapi_psr.ml
```diff
@@ -145,11 +167,214 @@ functor
...
+      raise
+        Api_errors.(
+          Server_error (host_is_slave, [Pool_role.get_master_address ()])) ;
+    let live_hosts = Helpers.get_live_hosts ~__context |> HostSet.of_list in
+    let all_hosts_list = Xapi_pool_helpers.get_master_slaves_list ~__context in
+    let all_hosts = all_hosts_list |> HostSet.of_list in
+    ( if not (HostSet.equal all_hosts live_hosts) then
+        let offline_host =
+          HostSet.diff all_hosts live_hosts
+          |> HostSet.find_first (fun _ -> true)
```
lindig, line 348: Is this the same as `HostSet.min_elt`? You could also start with the set difference and save the `equal`.

ocaml/xapi/xapi_psr.ml
psafont: The exception thrown when the file is not found looks like this:
```
open_in "doesn't exist";;
Exception: Sys_error "doesn't exist: No such file or directory".
```

lippirk: thanks! i also didn't know you could match on exceptions in match blocks

psafont: It got added in ocaml 4.02, so it's available for all backports as well :)

ocaml/xapi/xapi_psr.ml
```diff
@@ -145,11 +167,223 @@ functor
...
+            D.error "unexpected checkpoint file format: %s" checkpoint_path ;
+            raise
+              Api_errors.(
+                Server_error
+                  (internal_error, ["unexpected checkpoint file format"]))
+        | exception e ->
+            let fail_msg = "failed to read checkpoint file" in
+            D.error "%s %s. error: %s" fail_msg checkpoint_path
+              (Printexc.to_string e) ;
+            raise Api_errors.(Server_error (internal_error, [fail_msg]))
```
psafont, line 247: the `if .. else ` can be replaced with a `| exception SysError msg when Astring.String.is_infix ~affix:"No such file or directory" msg -> None`

This makes the race condition to disappear as well, even if the matching is fragile, but that can't be avoided, unfortunately

PR #4128, opened 2020-06-11
CA-338596 vendor fpath module

lindig:
This vendors the fpath module such that we can use it to check paths for the XE client. I've decided that this is easier than adding the module to the xs-opam library collection because it is only used here and only in the context of a backport. You might disagree.


lippirk:
I agree that is easier and sensible, so long as fpath is only used in xapi

I confirmed that the source has not been altered during the vendoring by comparing to fpath v0.7.2

PR #4129, opened 2020-06-11
CA-340776: move stunnel disconnection to the end where it was

edwintorok:
Disconnecting from stunnel as soon as we finished uploading the
requested file causes some unexplained EOFs.
XE used to disconnect from stunnel only at the end, so restore this
behaviour. We can address the stunnel bug revealed by this another time.

Also need to keep the channels alive longer, so manually create them
instead of using the with_ variant that closes them when we're done with them.


edwintorok:
Rebased on LCM branch, we won't merge this to master just yet.

PR #4133, opened 2020-06-15
Maximise string sharing in database

lindig:
The commits below add a weak string pool that is used to hold strings used in the Xapi database. Sharing among strings is maximised by looking up the string before it is entered into the database as a key or value. The goal is to reduce memory consumption of the database. The general access pattern is that we have many more reads than insertions such that the added lookup should not matter. Some early benchmarks show that memory consumption of xapi is reduced by 10% or more.

### DIFF
--- a/ocaml/database/db_cache_types.ml
+++ b/ocaml/database/db_cache_types.ml
@@ -16,6 +16,20 @@ open Db_exn
 
 module Time = struct type t = Generation.t end
 
+module HashedString = struct
+  type t = string
+
+  let equal = String.equal
+
+  let hash = Hashtbl.hash
+end
+
+module StringPool = Weak.Make (HashedString)
+
+let share =
+  let pool = StringPool.create 2048 in
+  StringPool.merge pool
+
 module Stat = struct
   type t = {created: Time.t; modified: Time.t; deleted: Time.t}
 
@@ -30,6 +44,8 @@ module StringMap = struct
 
     let compare = Stdlib.compare
   end)
+
+  let add key v t = add (share key) v t
 
   let update key default f t =
     let v = try find key t with Not_found -> default in
@@ -125,6 +141,17 @@ functor
 
 module Row = struct
   include Make (Schema.Value)
+
+  let add gen key v =
+    add gen key
+    @@
+    match v with
+    | Schema.Value.String x ->
+        Schema.Value.String (share x)
+    | Schema.Value.Pairs ps ->
+        Schema.Value.Pairs (List.map (fun (x, y) -> (share x, share y)) ps)
+    | Schema.Value.Set xs ->
+        Schema.Value.Set (List.map share xs)
 
   type t = map_t
 


psafont:
This looks good at first sight, are there any unit tests we can add to check that behaviour didn't change?

lindig:
I have not added unit tests. The observable behaviour from inside the program should not change. And even if I miss some strings and insert them directly as before, this would not cause a problem - just reduce the effectiveness. The main challenge are the many layers of the data structures for the database are built and to make sure we actually maximise the sharing. There is pseudo inheritance coming in from `include` statements.

ocaml/database/db_cache_types.ml
```diff
@@ -16,6 +16,25 @@ open Db_exn
...
+
+  let hash = Hashtbl.hash
+end
+
+module StringPool = Weak.Make (HashedString)
+
+let pool = StringPool.create 2048
+
+let unique str =
+  match StringPool.find_opt pool str with
```
edwintorok, line 32: I think `StringPool.merge pool str` would already do this. The docs say "merge t x returns an instance of x found in t if any, or else adds x to t and return x."

ocaml/database/db_cache_types.ml
```diff
@@ -126,6 +147,17 @@ functor
...
   include Make (Schema.Value)
 
+  let add gen key v =
+    add gen key
+    @@
+    match v with
+    | Schema.Value.String x ->
+        Schema.Value.String (unique x)
+    | Schema.Value.Pairs ps ->
+        Schema.Value.Pairs (List.map (fun (x, y) -> (unique x, unique y)) ps)
```
edwintorok, line 157: You could change `add_to_map` to do the unique call, so we don't have to walk the map twice.

ocaml/database/db_cache_types.ml
```diff
@@ -126,6 +147,17 @@ functor
...
+    | Schema.Value.Set xs ->
+        Schema.Value.Set (List.map unique xs)
```
edwintorok, line 159: Same here with `add_to_set`.
In fact the type of Pairs and Set is rather inefficient, the former should be a map, and the latter a true set. That would speed up some of the add/remove/lookup functions since they wouldn't have to walk a list anymore.
And potentially generate less garbage as well since removing a value wouldn't involve rewriting the whole list, but just updating up to a `log` depth on a tree.
Perhaps that could come as an optimization in a separate PR.

lindig: I would like to keep this PR focused on the sharing aspect but I agree that the abundance of lists in a database at least looks inefficient.

lindig: I have simplified the implementation. I am reluctant to modify `add_to_map` because it spreads out where we would add the sharing.

lindig:
Travis fails but it is not because of the code in the PR but something about Travis

ocaml/database/db_cache_types.ml
```diff
@@ -26,14 +26,9 @@ end
...
-let pool = StringPool.create 2048
-
-let unique str =
-  match StringPool.find_opt pool str with
-  | None ->
-      StringPool.add pool str ; str
-  | Some s ->
-      s
+let share =
+  let pool = StringPool.create 2048 in
```
psafont, line 30: Why has `pool` been moved from the module scope to function scope?

lindig: I moved it because nobody else needs the pool. Is this wrong in the way that it would re-create the pool every time? My intention was to create a closure.

psafont: Just tested and you're right, it creates the closure and pool gets reused correctly

ocaml/database/db_cache_types.ml
```diff
@@ -16,6 +16,20 @@ open Db_exn
 
 module Time = struct type t = Generation.t end
 
+module HashedString = struct
+  type t = string
+
+  let equal = String.equal
+
+  let hash = Hashtbl.hash
```
mseri, line 24: Not sure it's worth, but you can have an extra perf improvement taking inspiration from https://github.com/ocaml/ocaml/pull/8878

lindig: We are on OCaml 4.08 – I'd rather wait for `String.hash` or similar in the standard library than trying to do something clever here.

lippirk:
You mentioned in commit c0ef3e1b1 that you are passing the string through a function `unique`, but I can't see any such function in your change - perhaps you need to update the commit message

lindig:
That's fair - I've renamed it to `share`

lindig:
I probably also condense this down to a single commit now that everything is quite simple

lindig:
The Travis run failed a test that most likely is caused by a using a different container than before. This is unrelated to the code change.

psafont:
Fix for the failed test is in https://github.com/xapi-project/xen-api/pull/4150

lindig:
I can't re-run the test and suggest we merge this when there are no other concerns.

PR #4134, opened 2020-06-15
maintenance: move tar_helpers to xapi_aux

psafont:
This avoids adding dependencies to xapi-consts which was causing
failures when building it on travis.


ocaml/xapi-cli-server/dune
```diff
@@ -11,7 +11,7 @@
    result
    rresult
    sexplib
-   tar-unix
+   tar
```
lippirk, line 14: are we missing a dependency on `xapi-aux` here?

```
[ben@piglet:xapi-cli-server {stockholm-test} ((HEAD detached at psafont/tar))]$ rg Tar
<..>
cli_operations.ml:                    Tar_helpers.write_end sock;
```

psafont: xapi_aux is on line 24 ;)

PR #4135, opened 2020-06-15
fix pool config parsing

lippirk:
This was broken in ee226e0, since Astring.String.cut
doesn't behave as expected.

### DIFF
--- a/ocaml/xapi-aux/pool_role.ml
+++ b/ocaml/xapi-aux/pool_role.ml
@@ -50,11 +50,11 @@ let read_pool_role () =
   try
     let s = Astring.String.trim
         (Unixext.string_of_file !Constants.pool_config_file) in
-    match Astring.String.cut ~sep:":" s with
-    | Some ( "master", "")  -> Master
-    | Some ( "slave", m_ip) -> Slave m_ip
-    | Some ( "broken", "")  -> Broken
-    | _ -> failwith "cannot parse pool_role from pool config file"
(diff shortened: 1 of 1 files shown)


lippirk:
```
# Astring.String.cut ~sep:":" "master";;
- : (string * string) option = None

# Astring.String.cuts ~sep:":" "master";;
- : string list = ["master"]
```

ocaml/xapi-aux/pool_role.ml
```diff
@@ -50,11 +50,11 @@ let read_pool_role () =
...
     let s = Astring.String.trim
         (Unixext.string_of_file !Constants.pool_config_file) in
-    match Astring.String.cut ~sep:":" s with
-    | Some ( "master", "")  -> Master
-    | Some ( "slave", m_ip) -> Slave m_ip
-    | Some ( "broken", "")  -> Broken
-    | _ -> failwith "cannot parse pool_role from pool config file"
+    match Astring.String.cuts ~sep:":" s with
+    | ["master"]      -> Master
+    | ["slave"; m_ip] -> Slave m_ip
```
psafont, line 55: Could you use Slave (String.concat ":" m_ip), just in case? It's unfortunate astring doesn't have a direct replacement to the split with limit.

lippirk: yes, for ipv6 (NB. this was not the original behaviour)

PR #4136, opened 2020-06-15
CA-335033 avoid idle connections during VDI copy

lippirk:
Prefer to use Client.InternalAsync.VDI.copy over
Client.VDI.copy to avoid long running idle
connections when instructing a slave to perform
a VDI.copy.

For more information see CA-333610, where the
same problem was fixed for VM.migrate_send


lippirk:
I have rebased - please can I get a couple of re-approves?

PR #4138, opened 2020-06-17
Merge stunnel fixes from LCM branch

edwintorok:
Now that the formatting changes are in this can be merged. There were still some conflicts in the XE cli, I fixed by picking the LCM version.
I used git merge and solve the conflicts rather than cherry-pick to make merges easier in the future.

### DIFF
--- a/ocaml/xe-cli/newcli.ml
+++ b/ocaml/xe-cli/newcli.ml
@@ -40,6 +40,8 @@ let xedebug = ref false
 
 let xedebugonfail = ref false
 
+let stunnel_processes = ref []
+
 let debug_channel = ref None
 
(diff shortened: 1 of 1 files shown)


lindig:
The change is fairly small - why does this show up as 7 commits?

edwintorok:
There is a merge of the formatting commits on LCM with the one on master, to tell git they are equivalent: 2 merge commits + the formatting commit (this is a noop with 0 diff), merge of LCM branch: 1 commit. And the 3 commits from the LCM branch to fix the bug (I think it took multiple attempts until we found a fix that works).
Its a lot, but now `git merge origin/1.249-lcm` says there is nothing more there.

PR #4139, opened 2020-06-17
CI: use github actions

psafont:
This approach substitutes the current shellscript for travis, as just simply running like in https://github.com/xapi-project/xs-opam/pull/468 produced failures.

The part that loads the .env file is a bit clunky since it needs to remove ` export ` from every line.

This approach has the benefit that it better shows how a developer would build and test the repository using an opam switch instead of running an opaque scripts that only is run for CI.

When running the jobs now it's plain simple how much do the steps take, most of the time being spent compiling the dependencies. Cache is used to speed up build by about 8 minutes, the cache is invalidated daily to minimize problems with packages in xs-opam that are not versioned and use 'master' as their version.

A daily scheduled run serves as a way to retest the build without a cache and rebuild the cache to spped up any jobs thay may be generated by PRs that day.

Sample run: https://github.com/psafont/xen-api/actions/runs/172460064

A similar setup is needed to enable testing in `1.249-lcm`, a bit more thinking is needed around the cache, as scheduled runs are only triggered for `master` branch.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,75 @@
+name: Build and test
+
+on:
+  push:
+  pull_request:
+  schedule:
+    # run daily, this refreshes the cache
+    - cron: '13 2 * * *'
+
+jobs:
+  python-test:
+    name: Python tests
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        test: ["", "-3"]
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Run python tests
+        run: bash ./.travis-python-nosetests${{ matrix.test }}.sh
+
+  ocaml-test:
+    name: Ocaml tests
+    runs-on: ubuntu-20.04
+    env:
+      package: "xapi-cli-protocol xapi-client xapi-consts xapi-database xapi-datamodel xapi-types xapi xe"
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Pull configuration from xs-opam
+        run: |
+          curl --fail --silent https://raw.githubusercontent.com/xapi-project/xs-opam/master/tools/xs-opam-ci.env | cut -f2 -d " " > .env
+
+      - name: Load environment file
+        id: dotenv
+        uses: falti/dotenv-action@v0.2.4
+
+      - name: Retrieve date for cache key
+        id: cache-key
+        run: echo "::set-output name=date::$(/bin/date -u "+%Y%m%d")"
+        shell: bash
+
+      - name: Restore opam cache
+        id: opam-cache
+        uses: actions/cache@v2
+        with:
+          path: "~/.opam"
+          # invalidate cache daily, gets built daily using a scheduled job
+          key: ${{ steps.cache-key.outputs.date }}
+
+      - name: Use ocaml
+        uses: avsm/setup-ocaml@v1.1.0
+        with:
+          ocaml-version: ${{ steps.dotenv.outputs.ocaml_version_full }}
+          opam-repository: ${{ steps.dotenv.outputs.repository }}
+
+      - name: Install dependencies
+        run: |
+          opam pin add . --no-action
+          opam depext -u ${{ env.package }}
+          opam install ${{ env.package }} --deps-only --with-test -v
+
+      - name: Build
+        run: |
+          opam exec -- ./configure
+          opam exec -- make
+        env:
+          XAPI_VERSION: "v0.0.0-${{ github.sha }}"
+
+      - name: Run tests
+        run: opam exec -- make test


.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
...
+    env:
+      package: "xapi"
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Pull configuration from xs-opam
+        run: |
+          curl https://raw.githubusercontent.com/psafont/xs-opam/gh-act/tools/xs-opam-ci.env 2> /dev/null | cut -f2 -d " " > .env
```
psafont, line 31: This url needs to be changed to point to xapi-project/xs-opam/master when ready

.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
...
+          opam pin add . --no-action
+          opam depext -u ${{ env.package }}
+          opam install ${{ env.package }} --deps-only --with-test -v
+
+      - name: Build
+        run: |
+          opam exec -- ./configure
+          opam exec -- make
+        env:
+          XAPI_VERSION: "v0.0.0-${{ github.sha }}"
```
psafont, line 58: Not sure how else to manage this.

mseri:
Have you also looked at https://github.com/ocurrent/ocaml-ci (which enables caching and reactive retrigger on dependency changes)?

psafont:
I had a look, I don't think their instance is public. Maybe setting one up is something @olivierlambert would be willing to investigate.

olivierlambert:
Sure, we can investigate. What do you need exactly?

.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Pull configuration from xs-opam
+        run: |
+          curl https://raw.githubusercontent.com/psafont/xs-opam/gh-act/tools/xs-opam-ci.env 2> /dev/null | cut -f2 -d " " > .env
+
+      - name: Load environment file
+        id: dotenv
+        uses: falti/dotenv-action@v0.2.4
+
+      - name: Use ocaml
+        uses: avsm/setup-ocaml@v1
```
lindig, line 38: I think this includes quite a large opam repo that we later remove - so the download it wasteful in that regard.

.github/workflows/main.yml
lindig: I would suggest to to use curl flags 
* `--fail` to detect failure
* `--silent` to avoid other output

.github/workflows/main.yml
psafont: This would need a modification of the action to support replacing the repository before creating a switch.

lindig: I realise that this is an external action that relies on outside containers - just wanted to point out the inefficiency here that is undesirable.

psafont:
> Sure, we can investigate. What do you need exactly?

Set up and configure the ocaml-ci as a service in the infrastructure, link it to a xen-api repository. This will allow to test against travis and github, especially see how reliable and fast it is.

.github/workflows/main.yml
psafont: I understand, I also though the same when I was writing the file. I was pointing to the solution to solve the issue you brought up.

mseri: I think Anil would be happy if he receives a PR for the customisation (or an issue to start the discussion)

avsm: @mseri speaks accurately, as always :-)

olivierlambert:
@psafont so I enabled the app and it's activated on https://github.com/xcp-ng/xen-api repo

.github/workflows/main.yml
psafont: See https://github.com/avsm/setup-ocaml/pull/27

psafont:
I'm waiting on https://github.com/avsm/setup-ocaml/pull/27 to be merged before making this PR ready for review

.github/workflows/main.yml
lippirk: I think this is fine

psafont:
We can see both workflows being triggered [here](https://github.com/psafont/xen-api/pull/3) main is triggered because of pushing to `master` and stockholm is triggered by having `1.249-lcm` as the base of the PR.

psafont:
Testing cache right now. It brought down the build time to less than 6 minutes, but I'm afraid it makes opam not rebuild some dependencies when it's needed: https://github.com/psafont/xen-api/runs/878570649?check_suite_focus=true

psafont:
I've added a nigtlhy job to be scheduled along with an opam cache with the date as a key. This makes sure the cache generated are invalidated daily and rebuilt by the scheduled job. This minimizes problems with dependencies getting out of date while still accelerating all the build jobs.

This is not enabled for stockholm-based PRs, and it's not clear if it would work the same way as scheduled runs only get triggered for `master`.

psafont:
Adding fpath to the dependencies of xe showed that all the packages had to be mentioned when installing dependencies.

PR #4143, opened 2020-06-22
CA-341442: is_json does not mean is_jsonrpc after all

robhoes:
A previous patch changed `jsoncallback` to call `callback1` with
`is_json:true`. Unfortunately, this broke the responses. The fix is to add a
new parameter `is_jsonrpc` for this case.

### DIFF
--- a/ocaml/xapi/api_server.ml
+++ b/ocaml/xapi/api_server.ml
@@ -101,10 +101,10 @@ module D=Debug.Make(struct let name="xapi" end)
 open D
 
 (** Forward a call to the master *)
-let forward req call is_json =
+let forward req call is_jsonrpc =
   let open Xmlrpc_client in
   let transport = SSL(SSL.make ~use_stunnel_cache:true (), Pool_role.get_master_address(), !Xapi_globs.https_port) in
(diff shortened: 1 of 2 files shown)


psafont:
could the `is_json` option in `forward` also be changed at the function definition as well?

robhoes:
@psafont Good idea. I've updated it.

ocaml/xapi/api_server.ml
```diff
@@ -122,7 +122,7 @@ let is_himn_req req =
   | None -> false
 
 (* This bit is called directly by the fake_rpc callback *)
-let callback1 is_json req fd call =
+let callback1 is_json is_jsonrpc req fd call =
```
lippirk, line 125: can we add a comment explaining why we need  both `is_json` and `is_jsonrpc`? on the surface it looks like accidental complexity

robhoes: It is accidental complexity :) But sure, I'll add a comment.

robhoes:
I've updated the comment. Could I get a re-review?

robhoes:
Arrg... I forgot the signed-off-by line.

PR #4144, opened 2020-06-24
CA-341518 halt VM if it fails a migration

lippirk:
See https://github.com/xapi-project/xenopsd/pull/713

If a VM fails a migration, xenopsd should clean up the VM's runtime state, and xapi should halt the VM.

### DIFF
--- a/ocaml/xapi/xapi_vm_migrate.ml
+++ b/ocaml/xapi/xapi_vm_migrate.ml
@@ -189,6 +189,7 @@ let rec migrate_with_retries ~__context queue_name max try_no dbg vm_uuid
     xenops_vdi_map xenops_vif_map xenops_vgpu_map xenops =
   let open Xapi_xenops_queue in
   let module Client = (val make_client queue_name : XENOPS) in
+  let vm = Db.VM.get_by_uuid ~__context ~uuid:vm_uuid in
   let progress = ref "(none yet)" in
   let f () =
     progress := "Client.VM.migrate" ;
@@ -227,6 +228,18 @@ let rec migrate_with_retries ~__context queue_name max try_no dbg vm_uuid
           "xenops: not retrying migration: caught %s from %s in attempt %d of \
            %d."
           (Printexc.to_string e) !progress try_no max ;
+        ( try
+            Xapi_vm_lifecycle.force_state_reset ~__context ~self:vm
+              ~value:`Halted ;
+            info
+              "migrate_with_retries: successfully set state of VM='%s' to \
+               `Halted"
+              (Ref.string_of vm)
+          with e ->
+            error
+              "migrate_with_retries: failed to set state of VM='%s' to `Halted"
+              (Ref.string_of vm)
+        ) ;
         raise e
 
 let migrate_with_retry ~__context queue_name =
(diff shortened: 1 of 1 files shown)


stormi:
For users, a failed migration that ends in a shutdown VM means downtime. Is there no way to ensure the VM keeps running on the sender host if the migration failed?

(Note: I haven't read the patches)

lindig:
This would benefit from a review by @robhoes. I simply don't know whether shutting down the VM is the right policy here. @stormi note that this is resolving an issue where the VM can't be shut down after migration, which is also not a good situation to be in.

edwintorok:
In some cases it might be able to resume the VM (fast-resume) into the same domain if we haven't actually unpaused the VM on the remote side. There is some interested in implementing this (CP-29222), but it is not quite a simple as it sounds.
Meanwhile we can improve the situation with the current migrations: if the migration fails the VM *is* actually not working already, it is just that xenopsd is not available, and the next time you try to migrate it you get some weird internal errors (which prevents host evacuation, which prevents hotfix application). So all this patch does is to sync xenopsd's state with reality.
If we can improve that in the future and make it resume instead that'd be great, but having this code would be useful in that case too (we can only resume the guest in some conditions, and maybe only certain kinds of guests, and that fast resume can itself fail, in which case we need to clean up anyway).

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -227,6 +228,18 @@ let rec migrate_with_retries ~__context queue_name max try_no dbg vm_uuid
           "xenops: not retrying migration: caught %s from %s in attempt %d of \
            %d."
           (Printexc.to_string e) !progress try_no max ;
+        ( try
+            Xapi_vm_lifecycle.force_state_reset ~__context ~self:vm
```
edwintorok, line 232: Should we add a safety check here that the VM on the source is not running (i.e. paused, etc.)? If it is still running then there is no need to kill it, we can just let it keep running, the migration failed but we haven't made any permanent changes to the source VM yet.

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -1035,7 +1035,7 @@ let migrate_send' ~__context ~vm ~dest ~live ~vdi_map ~vif_map ~vgpu_map
       true
     with _ -> false
   in
-  let is_same_host = is_intra_pool && remote.dest_host == localhost in
```
robhoes, line 1038: :o !!!

ocaml/xapi/xapi_vm_migrate.ml
robhoes: This is a dangerous operation, and I don't think it does what you want.

We should let xenopsd handle the cleanup, in the xenopsd PR. If that starts working, then xapi will update the VM state through the usual xenopsd event mechanism.

robhoes:
I really want the first patch, but I think we should drop the second.

ocaml/xapi/xapi_vm_migrate.ml
robhoes: There is also existing fallback cleanup code in the exception handlers in `pool_migrate` and `migrate_send`, using `Xapi_xenops.shutdown` to give xenopsd another kick to clean up if required.

lippirk:
Okay, when I get time I'll see if I can get xenopsd to do the clean up. I'll close this and open a new PR with the equality changes

PR #4146, opened 2020-06-24
CA-341149: Ensure a wait happen when the heartbeat connection fails

psafont:
This was happening in some cases, but not all

Cherry-pick of https://github.com/xapi-project/xen-api/commit/991de5bfaaee877b9a044b178d23ee56442588e9


psafont:
it's not, weird. it's supposed to be the commit in https://github.com/xapi-project/xen-api/pull/4140

lindig:
But that commit has not been merged - that's why I could not find it.

psafont:
The hash doesn't match in any case, I need to change it

PR #4147, opened 2020-06-24
CA-341149: Ensure a wait happen when the heartbeat connection fails

psafont:
This was happening in some cases, but not all

Backport of https://github.com/xapi-project/xen-api/commit/991de5bfaaee877b9a044b178d23ee56442588e9


lindig:
The hash that refers to the original implementation does not exist for me

psafont:
the commit is supposed to be the one in https://github.com/xapi-project/xen-api/pull/4140

PR #4149, opened 2020-06-29
XSO-974: correctly reflect datamodel changes in stockholm

psafont:
The datamodel changes in https://github.com/xapi-project/xen-api/pull/4057 incorrectly state that were introduced in the stockholm release, this feature was only added after stockhlm branched off.

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -1486,7 +1486,7 @@ end
 (* These are included in vbds and vifs -- abstracted here to keep both these uses consistent *)
 let device_status_fields =
   [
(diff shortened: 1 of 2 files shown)


psafont:
I don't like the travis failure, it happens when installing the dependencies for travis-opam, before the depencies for xen-api are compiled

avsm:
Can you reproduce that Travis failure from a Dockerfile? I can't locally, but I don't have time to dig any further right now.

psafont:
> Can you reproduce that Travis failure from a Dockerfile? I can't locally, but I don't have time to dig any further right now.

I've got a broken container setup and it doesn't seem easily fixable so I cannot test it now either. hopefully I'll make time to reinstall the OS soon.

PR #4150, opened 2020-07-01
Fix travis tests

psafont:
The testing container changed the debian distribution changed, this changed the PCI tests.

Depending on the distributions' pci package the string for red hat might change, adapt the test so it accepts the two values the string might take.

Error looks like this:

```
Value "Red Hat, Inc." is not any of the expected ones: ["Red Hat, Inc"]
```

### DIFF
--- a/ocaml/pci/lib_test/test_pci.ml
+++ b/ocaml/pci/lib_test/test_pci.ml
@@ -59,23 +59,28 @@ let test_lookup_functions () =
      Device: 82371AB/EB/MB PIIX4 ACPI [7113]
      SVendor:        Red Hat, Inc [1af4]
      SDevice:        Qemu virtual machine [1100] *)
-  let test_lookup = assert_equal ~printer:(fun x -> x) in
+  let test_lookup valid_values value =
+    assert_bool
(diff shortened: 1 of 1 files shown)


ocaml/pci/lib_test/test_pci.ml
```diff
@@ -59,23 +59,28 @@ let test_lookup_functions () =
      Device: 82371AB/EB/MB PIIX4 ACPI [7113]
      SVendor:        Red Hat, Inc [1af4]
      SDevice:        Qemu virtual machine [1100] *)
-  let test_lookup = assert_equal ~printer:(fun x -> x) in
+  let test_lookup valid_values value =
```
lippirk, line 62: We could have turned this into `test_contains`, but this is more strict :+1:

psafont: We want to get rid of the pci module very soon anyway, so I just changed the code to pass the tests.

lindig: A name like `any` or `any_of` would suggest at the call site that we are expecting at least one match.

PR #4151, opened 2020-07-01
pci: fix tests for all distributions (backport)

lindig:
Depending on the distributions' pci package the string for red hat might
change, adapt the test so it accepts the two values the string might
take.

Backport of 7cdc390db


psafont:
Before this fix can be verified https://github.com/xapi-project/xs-opam/pull/479 needs to be merged

PR #4152, opened 2020-07-02
CA-319021 fixed resident_on field in Task

GeorgeBurac:


edwintorok:
If you have a ticket number (I see one in your branch name) please add in the commit title too, e.g. `CA-319021: ..`. It is easier to find what got fixed where.
The unit tests are failing here though, can you reproduce that failure locally?

PR #4155, opened 2020-07-06
CA-341988 avoid update keys="." in xapi database

lippirk:
I have tested this on Havana (we can't really test on master since there are no hotfixes for master).

Once this is merged I will create PRs for Stockholm and Havana

### DIFF
--- a/ocaml/xapi/xapi_pool_update.ml
+++ b/ocaml/xapi/xapi_pool_update.ml
@@ -43,7 +43,7 @@ type update_info = {
   ; name_label: string
   ; name_description: string
   ; version: string
-  ; key: string
+  ; key: string option
   ; installation_size: int64
   ; after_apply_guidance: API.after_apply_guidance list
   ; enforce_homogeneity: bool
@@ -309,7 +309,13 @@ let guidance_from_string = function
 let parse_update_info xml =
   match xml with
   | Xml.Element ("update", attr, children) ->
-      let key = try List.assoc "key" attr with _ -> "" in
+      let key =
+        Option.bind (List.assoc_opt "key" attr) (function
+          | "" ->
+              None
+          | s ->
+              Some (Filename.basename s))
+      in
       let uuid =
         try List.assoc "uuid" attr
         with _ ->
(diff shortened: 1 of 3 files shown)


ocaml/xapi/xapi_pool_update.ml
```diff
@@ -309,7 +309,13 @@ let guidance_from_string = function
 let parse_update_info xml =
   match xml with
   | Xml.Element ("update", attr, children) ->
-      let key = try List.assoc "key" attr with _ -> "" in
+      let key =
+        List.assoc_opt "key" attr |> Option.value ~default:"" |> function
+        | "" ->
+            ""
```
robhoes, line 315: Can we make `key` an option type instead? Then we are forced to handle it properly without relying on the special value `""`. Or would that cause too much churn?

lippirk: It doesn't cause too much churn so I'll change it

Unfortunately we don't get too much benefit because the DB still stores the key as a plain string (and when constructing the yum config we get the key from the DB rather than from values of type `update_info`)

ocaml/xapi/xapi_pool_update.ml
```diff
@@ -310,7 +309,10 @@ let guidance_from_string = function
 let parse_update_info xml =
   match xml with
   | Xml.Element ("update", attr, children) ->
-      let key = try List.assoc "key" attr with _ -> "" in
+      let key =
+        List.assoc_opt "key" attr |> fun k ->
+        Option.bind k (function "" -> None | s -> Some (Filename.basename s))
```
psafont, line 314: Using `(List.assoc_opt "key" attr)` directly as the first parameter for the bind seems like a simpler way to code this.

ocaml/xapi/xapi_pool_update.ml
```diff
@@ -19,7 +19,6 @@ open Http
 open Forkhelpers
 open Xml
 open Helpers
-open Listext
```
psafont, line 22: thanks!

PR #4156, opened 2020-07-07
CA-341988 stockholm backport

lippirk:


lippirk:
This travis build should succeed once https://github.com/xapi-project/xs-opam/pull/481 is merged

lippirk:
@psafont now that https://github.com/xapi-project/xs-opam/pull/480 is merged, should I expect to see github actions being run here?

psafont:
Github actions is not enabled for this repository yet, see #4139

ocaml/xapi/xapi_pool_update.ml
```diff
@@ -524,38 +531,42 @@ let introduce ~__context ~vdi =
...
-                 (Helpers.call_api_functions ~__context (fun rpc session_id ->
-                      Client.Pool_update.apply rpc session_id self host)) ;
-               acc
-             with e ->
-               debug "Caught exception while pool_apply %s: %s"
-                 (Ref.string_of host)
-                 (ExnHelper.string_of_exn e) ;
-               host :: acc)
-           []
+      HostSet.fold
```
lindig, line 552: I'm a bit surprised about the `ignore` inside a `fold`. Could this be done with `HashSet.iter` instead?

edwintorok: It appears to accumulate errors, so a straight iter wouldn't work, but I agree that `ignore` inside a fold is awkward.
Perhaps a clearer way to write this would be to List.map the outcome to a result, then filter out only the errors from that list. We can refactor this on master, would keep the backport identical to master for now.

edwintorok:
The travis build fails here, is the CI pointing to the LCM branch correctly here?

psafont:
> The travis build fails here, is the CI pointing to the LCM branch correctly here?

A rebase against current stockholm should trigger travis and github jobs @lippirk

PR #4157, opened 2020-07-07
Error message improvement.

kc284:
This message has an override in xen-api-sdk which corrects the language. This practice should generally be avoided, unless absolutely necessary, because the content of the override may get out of sync as it has happened in this case. This PR corrects the original message and this one https://github.com/xapi-project/xen-api-sdk/pull/177 removes the outdated SDK override.

### DIFF
--- a/ocaml/idl/datamodel_errors.ml
+++ b/ocaml/idl/datamodel_errors.ml
@@ -452,7 +452,7 @@ let _ =
   error Api_errors.host_master_cannot_talk_back [ "ip" ]
     ~doc:"The master reports that it cannot talk back to the slave on the supplied management IP address." ();
   error Api_errors.host_unknown_to_master [ "host" ]
-    ~doc:"The master says the host is not known to it. Perhaps the Host was deleted from the master's database, the slave is pointing to the wrong master, or a pool secret rotation failed?" ();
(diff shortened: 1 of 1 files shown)


ocaml/idl/datamodel_errors.ml
```diff
@@ -452,7 +452,7 @@ let _ =
   error Api_errors.host_master_cannot_talk_back [ "ip" ]
     ~doc:"The master reports that it cannot talk back to the slave on the supplied management IP address." ();
   error Api_errors.host_unknown_to_master [ "host" ]
-    ~doc:"The master says the host is not known to it. Perhaps the Host was deleted from the master's database, the slave is pointing to the wrong master, or a pool secret rotation failed?" ();
+    ~doc:"The master says the server is not known to it. Perhaps the server was deleted from the master's database, or is pointing to the wrong master, or a pool secret rotation failed?" ();
```
lippirk, line 455: I think the following would read a bit better (i.e. wording it in terms of some sanity checks the user can do)

```
The master says the server is not known to it. Is the server in the master's database and pointing to the correct master? Did a pool secret rotation fail?
```

Apologies for not doing this properly when I touched it last

kc284: As discussed, I corrected the message to state the sanity check in terms of the result rather than the action causing it.

PR #4160, opened 2020-07-14
CA-34357 add PSR feature flag

lippirk:
We don't actually want to ever disable this this
feature, but the flag is used by XC to determine
whether to show PSR button or not.

### DIFF
--- a/ocaml/xapi-types/features.ml
+++ b/ocaml/xapi-types/features.ml
@@ -63,6 +63,7 @@ type feature =
   | Network_sriov
   | Corosync
   | Zstd_export
+  | Pool_secret_rotation
 [@@deriving rpc]
 
 type orientation = Positive | Negative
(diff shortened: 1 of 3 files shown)


ocaml/xapi-types/features.ml
```diff
@@ -63,6 +63,7 @@ type feature =
   | Network_sriov
   | Corosync
   | Zstd_export
+  | PoolSecretRotation
```
robhoes, line 66: For consistency, let's call it `Pool_secret_rotation`.

ocaml/xapi-types/features.ml
```diff
@@ -117,6 +118,8 @@ let keys_of_features =
   ; (Network_sriov, ("restrict_network_sriov", Negative, "Network_sriov"))
   ; (Corosync, ("restrict_corosync", Negative, "Corosync"))
   ; (Zstd_export, ("restrict_zstd_export", Negative, "Zstd_export"))
+  ; ( PoolSecretRotation
+    , ("restrict_PoolSecretRotation", Negative, "PoolSecretRotation") )
```
robhoes, line 122: Again for consistency: `restrict_pool_secret_rotation`.

robhoes:
> We don't actually want to ever disable this this feature

There may still be value in adding a check for this flag in the main PSR function, so that you can't accidentally start a PSR on a pool that still has some pre-PSR hosts in it, during a rolling upgrade/update. This works, because at the pool level, the feature only becomes enabled once all hosts expose it.

robhoes:
By the way, unfortunately, the branch rule for feature branches with pattern `feature/*` does not appear to be matching on this branch, so there is no 2-approver check on this PR. Perhaps we need to add a `feature/*/*` pattern.

robhoes:
Yes, I've added `feature/*/*`, and now we need 2 approvers.

PR #4162, opened 2020-07-15
CP-34379 don't proceed with rotation if PSR state is inconsistent

lippirk:
We add some sanity checks before performing PSR
operations.

The main motivation for this is that if the master
changes after a failed PSR then we want to block
the next PSR. The idea is to avoid lockout at all
costs, even though manual intervention may be
required to fix the issue.

### DIFF
--- a/ocaml/idl/datamodel_types.ml
+++ b/ocaml/idl/datamodel_types.ml
@@ -267,8 +267,8 @@ let release_order_full = [{
      code_name     = Some rel_stockholm;
      version_major = 2;
      version_minor = 15;
-     branding      = "Unreleased";
-     release_date  = None;
+     branding      = "Citrix Hypervisor 8.2";
+     release_date  = Some "June 2020";
    }; {
      code_name     = Some rel_next;
      version_major = 2;
--- a/ocaml/idl/datamodel_types.ml
+++ b/ocaml/idl/datamodel_types.ml
@@ -64,7 +64,7 @@ let rel_naples = "naples"
 let rel_oslo = "oslo"
 let rel_quebec = "quebec"
 let rel_stockholm = "stockholm"
-let rel_next = "next"
+let rel_stockholm_psr = "stockholm_psr"
 
 type api_release = {
   code_name: string option;
@@ -270,10 +270,10 @@ let release_order_full = [{
      branding      = "Citrix Hypervisor 8.2";
      release_date  = Some "June 2020";
    }; {
-     code_name     = Some rel_next;
+     code_name     = Some rel_stockholm_psr;
      version_major = 2;
-     version_minor = 16;
-     branding      = "Unreleased";
+     version_minor = 15;
+     branding      = "Citrix Hypervisor 8.2 Hotfix n";
      release_date  = None;
    }
   ]
--- a/ocaml/xapi/xapi_psr.ml
+++ b/ocaml/xapi/xapi_psr.ml
@@ -274,8 +274,7 @@ functor
           Client.Host.cleanup_pool_secret rpc session_id host)
 
     let cleanup_master =
-      cleanup_internal
-        ~additional_files_to_remove:[checkpoint_path; checkpoint_tmp_path]
+      cleanup_internal ~additional_files_to_remove:[checkpoint_path]
   end
 
 let notify_new ~new_ps =
--- a/ocaml/xapi/xapi_psr.ml
+++ b/ocaml/xapi/xapi_psr.ml
@@ -465,62 +465,90 @@ end)
 
 module Ptoken = Genptokenlib.Lib
 
-(* at most one PSR should be
-   running at any given time *)
-let m = Mutex.create ()
-
-let start ~__context =
-  (* checked preconditions:
-     * the feature is enabled (to prevent a PSR when not all hosts have upgraded to this version)
-     * we are the master
-     * HA disabled
-     * RPU not running
-     * all hosts in pool 'alive' *)
-  let f () =
-    Pool_features.assert_enabled ~__context ~f:Features.Pool_secret_rotation ;
-    if
-      not
-        (Helpers.is_pool_master ~__context
-           ~host:(Helpers.get_localhost ~__context))
-    then
+let start =
+  let m = Mutex.create () in
+  let with_lock f =
+    (* prevents multiple concurrent PSRs *)
+    if Mutex.try_lock m then (
+      try f () ; Mutex.unlock m with e -> Mutex.unlock m ; raise e
+    ) else
       raise
         Api_errors.(
-          Server_error (host_is_slave, [Pool_role.get_master_address ()])) ;
-    Assert.master_state_valid () ;
-    let live_hosts = Helpers.get_live_hosts ~__context |> HostSet.of_list in
-    let all_hosts_list = Xapi_pool_helpers.get_master_slaves_list ~__context in
-    let all_hosts = all_hosts_list |> HostSet.of_list in
-    let offline_hosts = HostSet.diff all_hosts live_hosts in
-    ( if not (HostSet.is_empty offline_hosts) then
-        let offline_host = HostSet.min_elt offline_hosts in
+          Server_error (internal_error, ["pool secret rotation already running"]))
+  in
+  fun ~__context ->
+    let self = Helpers.get_pool ~__context in
+    let set_up_psr_pending_flag asserts =
+      let was_already_pending = Db.Pool.get_is_psr_pending ~__context ~self in
+      let maybe_set_pending value =
+        if not was_already_pending then
+          Db.Pool.set_is_psr_pending ~__context ~self ~value
+      in
+      maybe_set_pending true ;
+      try asserts ()
+      with e ->
+        (* only set pending to false if we just set it to true *)
+        maybe_set_pending false ; raise e
+    in
+    let assert_no_ha () =
+      let is_ha_enabled = Db.Pool.get_ha_enabled ~__context ~self in
+      if is_ha_enabled then
+        raise Api_errors.(Server_error (ha_is_enabled, []))
+    in
+    let assert_we_are_master () =
+      if
+        not
+          (Helpers.is_pool_master ~__context
+             ~host:(Helpers.get_localhost ~__context))
+      then
         raise
           Api_errors.(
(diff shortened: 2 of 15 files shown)


ocaml/xapi/xapi_psr.ml
```diff
@@ -186,24 +187,134 @@ functor
...
+        false
+    | false, true | true, false ->
+        raise
+          Api_errors.(
+            Server_error
+              (internal_error, ["do_backups_exist: invalid backup state"]))
+
+  let does_checkpoint_exist () = Sys.file_exists checkpoint_path
+
+  let backups_match ~old_ps ~new_ps =
```
lindig, line 226: A comment on style: this mixes logic and logging. I think it would be clearer to have a predicate for the matching and do the logging where it is used:

```
let backups_match = 
  let old_backup, new_backup = read_backups () in
  SecretString.equal old_backup old_ps
           && SecretString.equal new_backup new_ps
...
if not  backups_match then ..
```

ocaml/xapi/xapi_psr.ml
```diff
@@ -248,78 +359,116 @@ functor
...
-  | _ :: _ :: _ ->
-      D.info "xapi_psr.ml:notify_new: already accepting new pool secret"
-  | _ ->
-      let old_ps = Xapi_globs.pool_secret () in
+  | [priority_1_ps; priority_2_ps] ->
+      (* check disk state *)
+      Assert.backups_match ~old_ps ~new_ps ;
+      (* check runtime state *)
+      if
+        SecretString.equal priority_2_ps new_ps
```
lindig, line 392:
```
SecretString.(equal priority_2_ps new_ps && equal priority_1_ps old_ps)
```
I believe this could be used in other places as well.

ocaml/xapi/xapi_psr_util.ml
```diff
@@ -5,12 +5,18 @@ module Unixext = Stdext.Unixext
 
 let checkpoint_path = "/var/lib/xcp/psr_cp"
 
-let checkpoint_tmp_path = "/var/lib/xcp/psr_cp.tmp"
-
 let old_pool_secret_backup_path = "/var/lib/xcp/ptoken.old"
 
 let new_pool_secret_backup_path = "/var/lib/xcp/ptoken.new"
 
+let read_backups () =
```
lindig, line 12: Can this fail and would it be good to handle exceptions here?

ocaml/xapi/xapi_psr.ml
psafont: To me it seems like it's also mixing the result of reading the files and the result of the match.
This could be encoded as `(boolean, string) result` to completely separate business logic from error-treating logic.

```
match do_backups_match with
| Ok true -> ()
| _ -> raise Api_errors.(...)
```
The false and the failure to read ca be separated here as well to log them if need be.

lippirk: This is combination with your comment below regarding `read_backups` has tidied this up

lippirk: I don't gain much by returning a `result`, since we do always expect the backups to exist whenever we ask for them (see https://github.com/xapi-project/xen-api/pull/4123#discussion_r434637042)

ocaml/xapi/xapi_psr.ml
```diff
@@ -248,78 +348,114 @@ functor
...
   match !Xapi_globs.pool_secrets with
-  | _ :: _ :: _ ->
-      D.info "xapi_psr.ml:notify_new: already accepting new pool secret"
-  | _ ->
-      let old_ps = Xapi_globs.pool_secret () in
+  | [priority_1_ps; priority_2_ps] ->
+      (* check disk state *)
+      Assert.backups_match ~old_ps ~new_ps ;
+      (* check runtime state *)
+      if SecretString.(equal priority_2_ps new_ps && equal priority_1_ps old_ps)
```
lindig, line 380: Has this been run through `make format`?

ocaml/xapi/xapi_psr_util.ml
```diff
@@ -5,12 +5,22 @@ module Unixext = Stdext.Unixext
...
+  try
+    ( old_pool_secret_backup_path
+      |> Unixext.string_of_file
+      |> SecretString.of_string
+    , new_pool_secret_backup_path
+      |> Unixext.string_of_file
+      |> SecretString.of_string )
+  with e ->
+    D.error "xapi_psr_util.ml:read_backups failed" ;
+    raise Api_errors.(Server_error (internal_error, ["failed to read backups"]))
```
lindig, line 22: This does not capture the string for `e` - so it will be difficult to diagnose. You might also want to include the paths for the files.

ocaml/xapi/xapi_psr.ml
lippirk: It has

ocaml/xapi/xapi_psr.ml
```diff
@@ -186,24 +187,123 @@ functor
...
+(* we include some sanity checks for the following invariants:
+     - only the master should have a checkpoint
+     - either (a) all the master psr state exists (=> resuming a failed psr)
+       or     (b) none of the state exists (=> starting a new psr)
+     - existing pool secret backups are consistent with requested pool secret
+       changes
+     - the runtime state, i.e. Xapi_globs.pool_secrets (although this check
+       is not done here, but at each stage separately)
+   these invariants could be broken if for example a psr fails, the master
+   changes, and then a new psr starts on the new master
```
edwintorok, line 199: You should make pool secret rotation and HA be mutually exclusive.
Master promotion can occur at any time when HA is enabled (it cannot be blocked, it is essential to the functioning of HA), so there should be an assert that HA is not enabled, and raise `Api_errors.ha_is_enabled,` otherwise if one attempts to initiate a pool secret rotation. I see this is already present in `xapi_psr.ml`.

Similarly in the HA enable function there should be a check that no pool secret rotation is in progress.
I think this'll need a new operation in allowed operations for the pool in xapi_pool_helpers, and to remove `ha_enable` when psr becomes active.

lippirk: I opted for adding a flag to the database to solve this - that way we can block HA in between a failed PSR and the next attempt. I didn't see much benefit adding PSR to the operations list as well as this flag

lippirk:
I had to rebase against the changes which include a feature flag check

ocaml/xapi/xapi_psr.ml
```diff
@@ -329,61 +465,72 @@ end)
...
-          Server_error (host_is_slave, [Pool_role.get_master_address ()])) ;
-    let live_hosts = Helpers.get_live_hosts ~__context |> HostSet.of_list in
-    let all_hosts_list = Xapi_pool_helpers.get_master_slaves_list ~__context in
-    let all_hosts = all_hosts_list |> HostSet.of_list in
-    let offline_hosts = HostSet.diff all_hosts live_hosts in
-    ( if not (HostSet.is_empty offline_hosts) then
-        let offline_host = HostSet.min_elt offline_hosts in
+          Server_error (internal_error, ["pool secret rotation already running"]))
+  in
+  fun ~__context ->
```
psafont, line 479: It's a bit difficult to see in diff here, this construct is used to create a closure with the mutex m and and the function with_lock outside of `start ~__context`, isn't it?

I think that defining them at the top level would be easier to understand, is there a reason why it wasn't done that way?

lippirk: Yes, that's right. Personally I prefer putting it in the scope of `start`, that way it's obvious that only `start` can be using it (i.e. one less thing to consider if we were to run into a locking issue in the future)

ocaml/xapi/xapi_psr.ml
```diff
@@ -329,61 +465,72 @@ end)
...
-        Server_error (internal_error, ["pool secret rotation already running"]))
+    with_lock (fun () ->
+        Pool_features.assert_enabled ~__context ~f:Features.Pool_secret_rotation ;
+        assert_we_are_master () ;
+        Assert.master_state_valid () ;
+        let[@warning "-8"] (master :: members) = assert_all_hosts_alive () in
+        assert_no_ha () ;
+        assert_no_rpu () ;
+        let module PSR = Make (Impl (struct let __context = __context end)) in
+        Db.Pool.set_is_psr_pending ~__context ~self ~value:true ;
```
psafont, line 523: I think this should be set before asserting ha is not enabled to avoid race conditions, however small they might be. (and the flag reset if the assertion fails)

ocaml/xapi/xapi_psr.ml
```diff
@@ -329,61 +465,72 @@ end)
...
+          PSR.start
+            (Xapi_globs.pool_secret (), Ptoken.gen_token ())
+            ~master ~members
+        in
+        match r with
+        | Ok () ->
+            Db.Pool.set_is_psr_pending ~__context ~self ~value:false ;
+            D.info "Xapi_psr.start: pool secret rotation successful"
+        | Error e ->
+            let err_msg = user_facing_error_message ~__context e in
```
psafont, line 534: It's not clear why is_psr_pending is not reset in case of failure, is the pool secret rotation going to be retried? How?

lippirk: Yes, this is intended - if a PSR fails, the user is expected to fix whatever caused the issue and then run the command again

ocaml/idl/datamodel_types.ml
```diff
@@ -267,13 +267,13 @@ let release_order_full = [{
...
+     branding      = "Citrix Hypervisor 8.2";
+     release_date  = Some "June 2020";
    }; {
-     code_name     = Some rel_next;
+     code_name     = Some rel_stockholm_psr;
      version_major = 2;
-     version_minor = 16;
-     branding      = "Unreleased";
+     version_minor = 15;
+     branding      = "Citrix Hypervisor 8.2 Hotfix n";
```
lindig, line 276: I assume this is a placeholder. Just in the case it does leak out, I'd remove the trailing ` n` or keep the "Unreleased". Technically the string does not matter, only the version number.

lippirk: This actually introduces a conflict with 1.249-lcm because of https://github.com/xapi-project/xen-api/pull/4168, so we'll have an extra to nudge here to make sure we remove this placeholder

PR #4163, opened 2020-07-15
Moved tests into separate suites to reduce test exec time

GeorgeBurac:
moved test_daemon_manager, test_vdi_cbt, test_event, test_clustering, test_cluster_host, test_cluster, test_pusb, test_network_sriov, test_vm_placement and test_vm_helpers into separate suites to reduce execution time


ocaml/tests/dune
```diff
@@ -2,6 +2,16 @@
   (name suite_alcotest)
   (package xapi)
   (flags (:standard -warn-error +a-3-4-6-9-27-28-29-52))
+  (modules (:standard \
+   test_daemon_manager
```
lippirk, line 6: I think this is getting to the point where we should consider moving test executables to their own subfolder (so that we can avoid having to adding exceptions to this list)

GeorgeBurac: Okay, thank you. I'll try to move them.

psafont: I think having subdirectoriesper area being tested would make sense, like all the clustering tests being in `ocaml/test/cluster` and `tests/network` containing `test_vlan` `test_network_sriov` and storage

ocaml/tests/test_vm_placement.ml
```diff
@@ -596,6 +596,15 @@ module Categorisation = struct
...
 
+
+let () =
+  Suite_init.harness_init () ;
+  (* Alcotest hides the standard output of successful tests,
+     so we will probably not exceed the 4MB limit in Travis *)
+  Debug.log_to_stdout () ;
+  Alcotest.run "Test VM Placement suite" ["test", test]
+
+
```
psafont, line 607: Why isn't this at the end of the file? there's a `test` being defined at line 642

ocaml/tests/test_vm_helpers.ml
```diff
@@ -569,3 +569,10 @@ let test =
     , `Quick
     , test_get_group_key_vgpu_and_netsriov )
   ]
+
+let () =
+  Suite_init.harness_init () ;
+  (* Alcotest hides the standard output of successful tests,
+     so we will probably not exceed the 4MB limit in Travis *)
+  Debug.log_to_stdout () ;
+  Alcotest.run "Test VM Helpers suite" ["test", test]
```
lippirk, line 578: You could add a helper to avoid repeating this pattern, something like:

```
let run_with_init name test =
   Suite_init.harness_init () ;
  (* Alcotest hides the standard output of successful tests,
     so we will probably not exceed the 4MB limit in Travis *)
  Debug.log_to_stdout () ;
  Alcotest.run name ["test", test]
```

ocaml/tests/test_daemon_manager.ml
```diff
@@ -145,3 +145,6 @@ let test =
   ; ("test_timeout_succeed", `Slow, test_timeout_succeed)
   ; ("test_timeout_fail", `Slow, test_timeout_fail)
   ]
+
+let () =
+  Alcotest.run "Daemon Manager suite" ["test", test]
```
lippirk, line 150: To avoid this 'No newline at end of file', I'd do change this setting in vscode: https://stackoverflow.com/questions/44704968/visual-studio-code-insert-new-line-at-the-end-of-files

ocaml/tests/test_bond.ml
```diff
@@ -218,3 +218,6 @@ let test =
     , `Quick
     , test_create_bond_into_sriov_vlan_network )
   ]
+
+(* let () =
+  Alcotest.run "Test Bond suite" ["test", test] *)
```
lippirk, line 223: This should be removed

lippirk:
I think we'd benefit from a more detailed commit message - it should explain, at least, what you've done and the motivation for it.

You could also consider squashing both commits into a single commit, if you think that is appropriate

ocaml/tests/test_daemon_manager.ml
psafont: I recommend installing ocamlformat using opam and running `make format` ;)

lippirk:
A good sanity check for this change would be to count the number of tests that are executed on master, vs the number of tests that are executed on this branch and make sure that they give the same answer - we don't want to have dropped any tests

psafont:
There are two issues: the biggest test suite is not being run at all and isntead only some of the vgpu tests are being run. Another, bigger problem is that there are race condition among the executables created: sometimes they try to grab the same socket file. The tests need to be serialized, made to use different sockets, or if possible avoid using the socket at all, if possible.

psafont:
Not sure what the failures in travis mean, they are all about missing files:

```
- [exception] (Sys_error "/tmp/xapi-test/81aa08.tmp: No such file or directory")
- Raised at file "string.ml", line 115, characters 19-34
- Called from file "sexp.ml", line 112, characters 13-47
```
I've not been able to reproduce the errors locally

lippirk:
> There are two issues: the biggest test suite is not being run at all and isntead only some of the vgpu tests are being run. Another, bigger problem is that there are race condition among the executables created: sometimes they try to grab the same socket file. The tests need to be serialized, made to use different sockets, or if possible avoid using the socket at all, if possible.

@GeorgeBurac  has this been resolved with your changes to `harness_init`?

psafont:
> > There are two issues: the biggest test suite is not being run at all and isntead only some of the vgpu tests are being run. Another, bigger problem is that there are race condition among the executables created: sometimes they try to grab the same socket file. The tests need to be serialized, made to use different sockets, or if possible avoid using the socket at all, if possible.
> 
> @GeorgeBurac has this been resolved with your changes to `harness_init`?

I think so, only the issue with temporary directories remains.

GeorgeBurac:
> > There are two issues: the biggest test suite is not being run at all and isntead only some of the vgpu tests are being run. Another, bigger problem is that there are race condition among the executables created: sometimes they try to grab the same socket file. The tests need to be serialized, made to use different sockets, or if possible avoid using the socket at all, if possible.
> 
> @GeorgeBurac has this been resolved with your changes to `harness_init`?

Yes, it appears so, thanks to Pau. However, I am struggling to reproduce the missing file errors.

psafont:
Issue also happens when using github: https://github.com/psafont/xen-api/runs/889346544?check_suite_focus=true, it's not just a problem with travis

lindig:
This needs a better subject line like "Move tests into separate suits to reduce test exec time".

edwintorok:
I think the problem here is that these tests all use the same inventory filename in `/tmp/xapi-test/`, but nothing actually creates `/tmp/xapi-test`.
Previously creating the xapi listening socket with `Xapi.listen_unix_socket` has created that directory as a side-effect, but that is not called for all tests now anymore.
It'd probably be more reliable if you call `Unixext.mkdir_safe Test_common.working_area` from `harness_init` so that the directory exists when you start the test. It would also be useful to move that temp dir somewhere underneath `_build` so that a `make clean` removes it, perhaps just make the path relative in `working_area` instead of absolute.

GeorgeBurac:
> I think the problem here is that these tests all use the same inventory filename in `/tmp/xapi-test/`, but nothing actually creates `/tmp/xapi-test`.
> Previously creating the xapi listening socket with `Xapi.listen_unix_socket` has created that directory as a side-effect, but that is not called for all tests now anymore.
> It'd probably be more reliable if you call `Unixext.mkdir_safe Test_common.working_area` from `harness_init` so that the directory exists when you start the test. It would also be useful to move that temp dir somewhere underneath `_build` so that a `make clean` removes it, perhaps just make the path relative in `working_area` instead of absolute.

Thank you very much for that, Edvin. :)

GeorgeBurac:
> I think the problem here is that these tests all use the same inventory filename in `/tmp/xapi-test/`, but nothing actually creates `/tmp/xapi-test`.
> Previously creating the xapi listening socket with `Xapi.listen_unix_socket` has created that directory as a side-effect, but that is not called for all tests now anymore.
> It'd probably be more reliable if you call `Unixext.mkdir_safe Test_common.working_area` from `harness_init` so that the directory exists when you start the test. It would also be useful to move that temp dir somewhere underneath `_build` so that a `make clean` removes it, perhaps just make the path relative in `working_area` instead of absolute.

Thank you very much for that, Edvin. :)

Issue #4154, opened 2020-07-06
RFC proposal; Expose xen's `sched-gran` in XAPI


benjamreis:
The goal is to expose  xen's `sched-gran` setting in to a xapi's `Host` object in order to enable a user to set it.

How to do it:
- Add a new `sched-gran` RW field in `Host` with the corresponding getter and setter
- the setter and getter can call:
    - setter: call `/opt/xensource/libexec/xen-cmdline --set-xen sched-gran=<cpu | core | socket>`
    - getter: call `/opt/xensource/libexec/xen-cmdline --get-xen sched-gran` (if empty line returned then the retrun value should be `cpu` else the returned value by the script)

What do you guys think?

benjamreis:
More info:

We'd like to offer our users in Xen Orchestra the possibility to select the schedule granularity of Xen. It'd would be in an Advanced view only for users that know what they're doing.

Changing granularity from `cpu` (default value) to `core`allows to prevents cross domain side-channel attack while still being faster than disabling the hyperthreading.

PR #4164, opened 2020-07-17
Expose xen's scheduler granularity into the `Host`

benjamreis:
Solves https://github.com/xapi-project/xen-api/issues/4154 & https://github.com/xcp-ng/xcp/issues/407

This PR adds 2 new methods to the `Host` object:
- `set_sched_gran`: sets Xen's scheduler granularity (either `core`, `cpu` or `socket`)
- `get_sched_gran`: gets Xen's scheduler granularity

It uses `/opt/xensource/libexec/xen-cmdline`

This allows a user to select the xen's scheduler granularity and choose between efficiency and safety. For instance core scheduling prevents cross domain side-channel attack while being more efficient than disabling hyperthreading.

### DIFF
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -1364,6 +1364,34 @@ let host_query_ha = call ~flags:[`Session]
     ~hide_from_docs:true
     ()
 
+  let host_sched_gran = Enum("host_sched_gran", [
+    "core", "core scheduling";
+    "cpu", "CPU scheduling";
+    "socket", "socket scheduling";
+  ])
+
+  let set_sched_gran = call
+    ~name:"set_sched_gran"
+    ~lifecycle:[Prototyped, rel_next, ""]
+    ~doc:"Sets xen's sched-gran on a host. See: https://xenbits.xen.org/docs/unstable/misc/xen-command-line.html#sched-gran-x86"
+    ~params:[
+      Ref _host, "self", "The host";
+      host_sched_gran, "value", "The sched-gran to apply to a host"
+    ]
+    ~allowed_roles:_R_LOCAL_ROOT_ONLY
+    ()
+
+  let get_sched_gran = call
+    ~name:"get_sched_gran"
+    ~lifecycle:[Prototyped, rel_next, ""]
+    ~doc:"Gets xen's sched-gran on a host"
+    ~params:[
+      Ref _host, "self", "The host";
+    ]
+    ~allowed_roles:_R_LOCAL_ROOT_ONLY
+    ~result:(host_sched_gran, "The host's sched-gran")
+    ()
+
   (** Hosts *)
   let t =
     create_obj ~in_db:true ~in_product_since:rel_rio ~in_oss_since:oss_since_303 ~internal_deprecated_since:None ~persist:PersistEverything ~gen_constructor_destructor:false ~name:_host ~descr:"A physical host" ~gen_events:true
@@ -1485,6 +1513,8 @@ let host_query_ha = call ~flags:[`Session]
         notify_accept_new_pool_secret;
         notify_send_new_pool_secret;
         cleanup_pool_secret;
+        set_sched_gran;
+        get_sched_gran;
       ]
       ~contents:
         ([ uid _host;
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -3404,6 +3404,20 @@ functor
         let local_fn = Local.Host.cleanup_pool_secret ~host ~old_ps ~new_ps in
         do_op_on ~__context ~host ~local_fn (fun session_id rpc ->
             Client.Host.cleanup_pool_secret rpc session_id host old_ps new_ps)
+
+      let set_sched_gran ~__context ~self ~value =
+        info "Host.set_sched_gran: host='%s' sched='%s'"
+          (host_uuid ~__context self)
+          (Record_util.host_sched_gran_to_string value) ;
+        let local_fn = Local.Host.set_sched_gran ~self ~value in
+        do_op_on ~local_fn ~__context ~host:self (fun session_id rpc ->
+            Client.Host.set_sched_gran rpc session_id self value)
+
+      let get_sched_gran ~__context ~self =
+        info "Host.get_sched_gran: host='%s'" (host_uuid ~__context self) ;
+        let local_fn = Local.Host.get_sched_gran ~self in
+        do_op_on ~local_fn ~__context ~host:self (fun session_id rpc ->
+            Client.Host.get_sched_gran rpc session_id self)
     end
 
     module Host_crashdump = struct
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -2389,3 +2389,42 @@ let notify_send_new_pool_secret ~__context ~host ~old_ps ~new_ps =
 
 let cleanup_pool_secret ~__context ~host ~old_ps ~new_ps =
   Xapi_psr.cleanup ~__context ~old_ps ~new_ps
+
+let set_sched_gran ~__context ~self ~value =
+  if Helpers.get_localhost ~__context <> self then
+    failwith "Forwarded to the wrong host" ;
+  if not !Xapi_globs.allow_host_sched_gran_modification then
+    raise
+      Api_errors.(
+        Server_error (operation_not_allowed, ["Disabled by xapi.conf"])) ;
+  let arg =
+    Printf.sprintf "sched-gran=%s" (Record_util.host_sched_gran_to_string value)
+  in
+  let args = ["--set-xen"; arg] in
+  try
+    let _ = Helpers.call_script !Xapi_globs.xen_cmdline_script args in
+    ()
+  with e ->
+    error "Failed to update sched-gran: %s" (Printexc.to_string e) ;
+    raise
+      Api_errors.(
+        Server_error (internal_error, ["Failed to update sched-gran"]))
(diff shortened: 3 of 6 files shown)


benjamreis:
I'd like to mutualize https://github.com/xapi-project/xen-api/pull/4164/files#diff-9a639b670fb7e2432894cc91df8b27fbR3365-R3368 and https://github.com/xapi-project/xen-api/pull/4164/files#diff-126a0619d5403393ed49475b5dd96aa4R2396-R2399
Without making it a XAPI method, is it possible?

benjamreis:
It seems this call: https://github.com/xapi-project/xen-api/pull/4164/files#diff-126a0619d5403393ed49475b5dd96aa4R2405

Only call the script on master... Is it possible to select the host on which the script is called?

edwintorok:
Changing security related parameters for Xen require a lot of technical knowledge from the user: it is expected they read the Xen documentation on the cmdline flag, and associated Xen security bulletins and made a performance/security tradeoff.
If they know how to SSH into Dom0 and run this command it is assumed they have that knowledge (or take responsibility if it doesn't work).

It would probably be best to hide this behind a flag, and exposed only to advanced users. E.g. you could add a flag in xapi.conf (see how other flags like `pass-through-pif-carrier` are added) that defaults on in your specfile, and defaults off when absent in ours. If the flag is not present the API call should fail, e.g. the flag could be called 'xen-security-controllable'?

stormi:
Isn't anyone calling a XAPI method which is not exposed in a CLI or GUI tool an advanced user already?

robhoes:
I wonder if there is much benefit to adding this into the API. We don't tend to directly expose low-level settings. These are persistent settings that have existing command line tools (`xen-cmdline` in this case) that allow tweaking them. Or do you intend to expose this is a UI?

benjamreis:
@robhoes yes we'd like to put it in an Advanced tab in Xen Orchestra. :)

ocaml/xapi/message_forwarding.ml
```diff
@@ -3361,6 +3361,23 @@ functor
           (host_uuid ~__context host)
           value ;
         Local.Host.set_uefi_certificates ~__context ~host ~value
+
+      let host_sched_gran_to_string = function
```
lindig, line 3365: Would it be possible to use regular variants instead of polymorphic variants? (This would require a type definition)

ocaml/xapi/xapi_host.ml
```diff
@@ -2382,3 +2382,52 @@ let set_multipathing ~__context ~host ~value =
   Db.Host.add_to_other_config ~__context ~self:host ~key:"multipathing"
     ~value:(string_of_bool value) ;
   Xapi_host_helpers.Configuration.set_multipathing value
+
```
lindig, line 2385: Would it make sense to put the types and conversions into a local module?
```
module SchedGran = struct
  type granularity of Core | CPU | Socket
  let of_string = ..
  let to_string = ..
end

Comment (benjamreis) [reply ocaml/xapi/xapi_host.ml:2385]: Sure, no pb. :)

**Comment by benjamreis** [comment]
A little help would be appreciated here:
As said above, I noticed that with my code the `xen-cmdline` script is only called on the pool's master.
It means one can only alter the scheduling granularity of the master...
Can someone point me an example where the script is called on a specific host? I believe it's possible since it's used to set the dom0 memory entries in `grub.cfg` but I didn't find where and how. :/

I looked at `set_memory_static_range` and `set_memory_limits` in `xapi_vm.ml` and their corresponding methods in `message_forwarding.ml` but I couldn't find any clue unfortunately.

Thanks in advance!

File: ocaml/xapi/message_forwarding.ml [hunk 3]
```diff
@@ -3361,6 +3361,23 @@ functor
...
+        | `core -> "core"
+        | `cpu -> "cpu"
+        | `socket -> "socket"
+
+      let set_sched_gran ~__context ~self ~sched =
+        info "Host.set_sched_gran: host='%s' sched='%s'"
+          (host_uuid ~__context self)
+          (host_sched_gran_to_string sched) ;
+        Local.Host.set_sched_gran ~__context ~self
```
edwintorok, line 3374: This needs to be forwarded to the proper host, as it is written it'll always run on the pool master as you've noticed.
Look at other functions in this module, e.g. `request_backup`.

benjamreis: Thanks a lot! That was really helpful :)

benjamreis:
> Changing security related parameters for Xen require a lot of technical knowledge from the user: it is expected they read the Xen documentation on the cmdline flag, and associated Xen security bulletins and made a performance/security tradeoff.
> If they know how to SSH into Dom0 and run this command it is assumed they have that knowledge (or take responsibility if it doesn't work).
> 
> It would probably be best to hide this behind a flag, and exposed only to advanced users. E.g. you could add a flag in xapi.conf (see how other flags like `pass-through-pif-carrier` are added) that defaults on in your specfile, and defaults off when absent in ours. If the flag is not present the API call should fail, e.g. the flag could be called 'xen-security-controllable'?

I'm trying to do it. I've added en entry in `xapi_globs` as for the `xapi.conf` entry, it's in `monitor_master.ml` right? the `update_configuration_from_master` method?

Thanks

ocaml/xapi/xapi_host.ml
```diff
@@ -2407,6 +2407,8 @@ let host_sched_gran_to_string = function
 let set_sched_gran ~__context ~host ~sched =
   if Helpers.get_localhost ~__context <> host then
     failwith "Forwarded to the wrong host" ;
+  if not !Xapi_globs.allow_host_sched_gran_modification then
+    failwith "Forbidden operation: scheduler granularity can't be changed" ;
```
edwintorok, line 2411: Perhaps raising `operation_not_allowed` would be more appropriate here:
```
raise (Api_errors.Server_error(Api_errors.operation_not_allowed, [ "Disabled by xapi.conf" ]))
```

edwintorok:
Thanks for the changes, this is looking a lot better now.

ocaml/xapi/xapi_host.ml
benjamreis: Where do you think I should declare this module ?
In its own file seems a bit too much for me.

ocaml/xapi/message_forwarding.ml
robhoes: The polymorphic variant comes from the autogenerated code for enum types, so we can't change that here. However, to- and of-string functions are normally in record_util.ml, so we should move this function there.

ocaml/xapi/xapi_host.ml
```diff
@@ -2382,3 +2382,72 @@ let set_multipathing ~__context ~host ~value =
...
+      Api_errors.(
+        Server_error (internal_error, ["Failed to update sched-gran"]))
+
+let get_sched_gran ~__context ~host =
+  if Helpers.get_localhost ~__context <> host then
+    failwith "Forwarded to the wrong host" ;
+  let args = ["--get-xen"; "sched-gran"] in
+  try
+    let ret = Helpers.call_script !Xapi_globs.xen_cmdline_script args in
+    match ret with
```
robhoes, line 2431: It's more robust to call `String.trim` on `ret` before comparing, to remove any leading and trailing whitespace.

ocaml/xapi/message_forwarding.ml
benjamreis: Even though `sched-gran` won't be a field of an `host` record?
I'd be happy to do that because it solves my problem of mutualizing the to_string function.

ocaml/xapi/xapi_host.ml
robhoes: I just commented on message_forwarding: this type is autogenerated from the datamodel enum, so we can't change it. However, we should put the to/of string functions together (and not define them twice), and the usual place for them is record_util.ml.

ocaml/idl/datamodel_host.ml
```diff
@@ -1325,6 +1325,34 @@ let host_query_ha = call ~flags:[`Session]
...
+    "socket", "socket scheduling";
+  ])
+
+  let set_sched_gran = call
+    ~name:"set_sched_gran"
+    ~lifecycle:[Prototyped, rel_next, ""]
+    ~doc:"Sets xen's sched-gran on a host. See: https://xenbits.xen.org/docs/unstable/misc/xen-command-line.html#sched-gran-x86"
+    ~params:[
+      Ref _host, "host", "The host";
+      host_sched_gran, "sched", "The sched-gran to apply to a host"
```
robhoes, line 1340: The usual name for the value parameter of setters in the API is `value`.

robhoes: The usual name for the first parameter is actually `self` – for the autogenerated getters and setters at least. In this case, I don't mind `host`, because we have a lot of precedents.

ocaml/xapi/xapi_host.ml
```diff
@@ -2382,3 +2382,72 @@ let set_multipathing ~__context ~host ~value =
...
+    | "\n" ->
+        `cpu
+    | "sched-gran=cpu\n" ->
```
robhoes, line 2434: Bonus points for using the of_string function, as this is yet another enum/string conversion.

edwintorok (approved):
There are merge conflicts, but otherwise looks good.

benjamreis:
Thanks for the reviews, I'll rebase on master ASAP so it can be merged quickly :+1:

benjamreis:
Rebase done, I've reviewed the patch after the rebase in case I missed something and everything looks good to me! :)
Ready to be merged when you want :+1:

PR #4171, opened 2020-07-27
Feature/REQ-802: merge from master

edwintorok:
Merge from master, and fix conflicts.
Best way to review this is to diff against master, that is a fairly small diff.
`git diff origin/master..edwintorok/feature/REQ-802`

I tested that this builds, I still need to test whether it works too.


edwintorok:
This is outdated, will reopen with a new refresh.

PR #4172, opened 2020-07-30
CA-342171 fix get_server_localtime

lippirk:
Fix get_server_localtime by using a string to represent the localtime, rather than Date.iso8601, which is supposed to be UTC

### DIFF
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -855,9 +855,10 @@ let host_query_ha = call ~flags:[`Session]
       ~name:"get_server_localtime"
       ~in_oss_since:None
       ~in_product_since:rel_cowley
+      ~lifecycle:[Published, rel_cowley, ""; Changed, rel_next, "Return string instead of datetime"]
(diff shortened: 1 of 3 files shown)


ocaml/idl/datamodel_host.ml
```diff
@@ -855,9 +855,10 @@ let host_query_ha = call ~flags:[`Session]
       ~name:"get_server_localtime"
       ~in_oss_since:None
       ~in_product_since:rel_cowley
+      ~lifecycle:[Published, rel_cowley, ""; Changed, rel_next, "Return string instead of datetime"]
       ~params:[Ref _host, "host", "The host whose clock should be queried"]
       ~doc:"This call queries the host's clock for the current time in the host's local timezone"
-      ~result:(DateTime, "The current local time")
+      ~result:(String, "The current local time")
```
robhoes, line 861: This will change the type of the XML-RPC response from `<dateTime.iso8601>` to `<string>`, which is a backwards-incompatible change that existing clients may (still) not tolerate.

PR #4173, opened 2020-07-30
CA-342171 fix get_server_localtime

lippirk:
We previously changed the Stdext.Date library to
only accept UTC. This of course doesn't work
correctly when we are trying to work with a
server's localtime. The library has been modified
to handle non-UTC times again, so we use those
changes here.


lippirk:
Sibling of https://github.com/xapi-project/stdext/pull/50/files

lindig:
I think we can merge this once we have updated xs-opam on ring3/master

PR #4175, opened 2020-08-12
CP-33121: Remove dependency on stdext metapackage

psafont:
Most of the commits are mechanically replacing how the sublibraries are imported to not go through Stdext.

The second-to-last commit is the one with manual changes, removing the `open Stdext ` statement broke usages in the files, forcing replacements for the imports or the usages, as I saw fit.

From here on each the stdext metalibrary is not used anymore and all its sublibraries can be managed independently, be gradually trimmed to remove the unused functionality and make them support the actually needed functionality.

- listext: used mainly for set operations, could it be made to work around actual sets?
- xstringext: it's main use is the special split function with a limited number of splits, the rest can be marked as deprecated and replace the implementation with equivalents so it's evident how to replace their usages.
- threadext: used for Mutex.execute. I haven't seen it used for anything else. is there a better way to package this function?
- unixext: actually useful, an effort to test and document the functionality would be very positive.
- pervasiveext: main use is the finally function, the rest of the function should be deprecated and replaced (mostly by the `Option` functions)
- encodings and date: actually useful and with tests. quite a bit of work has been done here :)


edwintorok:
xstringext: we already have astring, we can use that as replacement if we ensure the semantics matches for the converted functions (likely requires quite a bit of testing).
listext: be careful not to introduce stack overflows if you replace listext's list map_tr. Replacements could be in form of Seq for all the map operations we do. Yes, I'd like if we used actual sets, needs some changes in XAPI's IDL to accomodate

psafont:
the xstringext function I mentioned doesn't have an equivalent in astring so an alternative way is still needed, see https://github.com/dbuenzli/astring/pull/17

lindig:
I think I like the general direction to make more explicit what we are using such that it becomes easier to phase out libraries.

psafont:
BVT and BST are both green.

lippirk:
I didn't spot any mistakes, thanks for doing a suite run

lindig (approved):

PR #4176, opened 2020-08-12
CA-341155: Fix console refresh when starting management server + other improvements

robhoes:


robhoes:
Reviewers: I suggest reviewing patch-by-patch, and with whitespace changes suppressed for the "restructure" patch.

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -20,25 +20,11 @@ module D = Debug.Make (struct let name = "xapi_mgmt_iface" end)
 
 open D
 
-(** Keep track of the management interface server thread *)
+module StringSet = Set.Make (struct
```
lindig, line 23:
```
module StringSet = Set.Make(String)
```

robhoes: Oh, good one!

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -48,114 +30,151 @@ let update_mh_info interface =
   in
   ()
 
-let _restart_stunnel_no_cache ~__context ~accept =
-  let (_ : Thread.t) =
-    Thread.create (fun () -> Helpers.Stunnel.restart ~__context ~accept) ()
-  in
-  ()
+module Stunnel = struct
+  let stunnel_accept_m = Mutex.create ()
```
lindig, line 34: The names in this module don't need to include `stunnel`  as this is already provided by the module.

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -48,114 +30,151 @@ let update_mh_info interface =
...
+     to communicate with each other. *)
+  let start ~__context ?addr () =
+    let socket, accept =
+      match addr with
+      | None -> (
+          info "Starting new server (listening on all IP addresses)" ;
+          try
+            (* Is it IPv6 ? *)
+            let addr = Unix.inet6_addr_any in
+            ( Xapi_http.bind (Unix.ADDR_INET (addr, Constants.http_port))
```
lindig, line 87: It seems strange to me to take the port number from `Constants` but to used a fixed string here at the same time - also below.

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -48,114 +30,151 @@ let update_mh_info interface =
...
+    management_servers := socket :: !management_servers ;
+    Stunnel.restart_stunnel ~__context ~accept ;
+    if Pool_role.is_master () && addr = None then
+      (* NB if we synchronously bring up the management interface on a master with a blank
+         database this can fail... this is ok because the database will be synchronised later *)
+      Server_helpers.exec_with_new_task "refreshing consoles" (fun __context ->
+          Dbsync_master.set_master_ip ~__context ;
+          Dbsync_master.refresh_console_urls ~__context)
+
+  type listening_mode = Off | Any | Local of StringSet.t
```
lindig, line 116: StringSet might not be the best name - it's a set of addresses (which are strings).

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -48,114 +30,151 @@ let update_mh_info interface =
...
   Xapi_inventory.update Xapi_inventory._management_address_type
     (Record_util.primary_address_type_to_string primary_address_type) ;
   update_mh_info interface
 
+let himn = ref None
+
+let management_m = Mutex.create ()
+
+let next_server_mode ~mgmt_enabled =
+  let localhost = "127.0.0.1" in
```
lindig, line 158: Is this correct when we are using IPv6?

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -48,114 +30,151 @@ let update_mh_info interface =
   in
   ()
 
-let _restart_stunnel_no_cache ~__context ~accept =
-  let (_ : Thread.t) =
-    Thread.create (fun () -> Helpers.Stunnel.restart ~__context ~accept) ()
-  in
-  ()
+module Stunnel = struct
```
lippirk, line 33: it's nice to add a sig so that you can be sure nobody is using `stunnel_accept_m` outside this module

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -48,114 +30,151 @@ let update_mh_info interface =
...
+                D.warn
+                  "reconfigure_stunnel: accept is not set, so not restarting \
+                   stunnel"
+          | Some accept ->
+              fun () -> _restart_stunnel_no_cache ~__context ~accept)
+    in
+    f ()
+end
+
+module Server = struct
```
lippirk, line 65: similarly it'd be good to have a sig here

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -48,114 +30,151 @@ let update_mh_info interface =
...
+     to the management IP. *)
+  let management_servers = ref []
+
+  let stop () =
+    debug "Shutting down the old management interface (if any)" ;
+    List.iter (fun i -> Http_svr.stop i) !management_servers ;
+    management_servers := []
+
+  (* Even if xapi listens on all IP addresses, there is still an interface appointed as
+     _the_ management interface. Hosts in a pool use the IP addresses of this interfaces
```
lippirk, line 77: presumably you want `this interface` rather than `this interfaces`

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -48,114 +30,151 @@ let update_mh_info interface =
...
+      Server_helpers.exec_with_new_task "refreshing consoles" (fun __context ->
+          Dbsync_master.set_master_ip ~__context ;
+          Dbsync_master.refresh_console_urls ~__context)
+
+  type listening_mode = Off | Any | Local of StringSet.t
+
+  (* This is an idempotent function that leaves servers running if they do not
+     require any changes. *)
+  let update' ~__context = function
+    | c, n when c = n ->
```
lippirk, line 121: i think `current, next` would be a lot clearer

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -48,112 +30,149 @@ let update_mh_info interface =
...
-  Http_svr.start Xapi_http.server socket ;
-  management_interface_server := socket :: !management_interface_server ;
-  restart_stunnel ~__context ~accept ;
-  if Pool_role.is_master () && addr = None then
-    (* NB if we synchronously bring up the management interface on a master with a blank
-       		   database this can fail... this is ok because the database will be synchronised later *)
-    Server_helpers.exec_with_new_task "refreshing consoles" (fun __context ->
-        Dbsync_master.set_master_ip ~__context ;
-        Dbsync_master.refresh_console_urls ~__context)
+  let current_mode = ref Off
```
lippirk, line 136: why not expose this externally using a function rather than a ref? like you did for `himn`

ocaml/xapi/xapi_mgmt_iface.ml
robhoes: Indeed, we should use `Constant.https_port` for 443.

ocaml/xapi/xapi_mgmt_iface.ml
robhoes: This is an interesting question. We have experimental and incomplete management support for IPv6, and I'm hesitant to change things like this now. I think it should be reviewed if/when the time comes to put effort into making IPv6 support first class.

robhoes:
@lippirk @lindig Thanks for your comments. I've addressed them in the new commit.

PR #4177, opened 2020-08-12
CA-343230 improve update precheck failure messages

lippirk:


ocaml/xapi/xapi_pool_update.ml
```diff
@@ -98,8 +98,27 @@ let updates_to_attach_count_tbl : (string, int) Hashtbl.t = Hashtbl.create 10
...
+  let vbds = get_update_vbds ~__context ~vdi in
+  List.iter
+    (fun self ->
+      if not (Db.VBD.get_currently_attached ~__context ~self) then
+        let msg =
+          Printf.sprintf
+            "pool_update: expected VBD='%s' to be attached but it isn't!"
+            (Ref.string_of self)
+        in
+        raise Api_errors.(Server_error (internal_error, [msg])))
```
psafont, line 116: A more informative message could be produced if all the unattached vbds are shown instead of just the first one.

This can be done with:
```ocaml
match List.filter (Fun.not vbd_attached) vbds with
| [] -> ()
| unattached_vbds -> raise ...
```

ocaml/xapi/xapi_pool_update.ml
```diff
@@ -146,15 +166,12 @@ let detach_helper ~__context ~uuid ~vdi =
...
-          let dom0 = Helpers.get_domain_zero ~__context in
-          let vbds = Client.VDI.get_VBDs ~rpc ~session_id ~self:vdi in
           List.iter
             (fun self ->
-              if Client.VBD.get_VM ~rpc ~session_id ~self = dom0 then (
-                Client.VBD.unplug ~rpc ~session_id ~self ;
-                Client.VBD.destroy ~rpc ~session_id ~self
-              ))
+              Client.VBD.unplug ~rpc ~session_id ~self ;
+              Client.VBD.destroy ~rpc ~session_id ~self)
```
psafont, line 174: I don't think this change has been explained in the commit message, what was the issue before?

ocaml/xapi/xapi_pool_update.ml
lindig: I second this. I was about to suggest (but like Pau's code better):
I wonder whether it is cleaner to first find all non-attached VBDs (`List.filter`) and then raise an error for all of them. The current code bails out when the first one is found (which would be the typical case, I assume). It's still fine.

```
let dom0 = Helpers.get_domain_zero ~__context in
let missing = 
  Db.VDI.get_VBDs ~__context ~self:vdi
  |> List.filter (fun self -> Db.VBD.get_VM ~__context ~self = dom0)
  |> List.filter (fun self -> (Db.VBD.get_currently_attached ~__context ~self)
```

ocaml/xapi/xapi_pool_update.ml
lippirk: this is functionally equivalent to before (i believe) - i just factored out `get_update_vbds`

PR #4179, opened 2020-08-13
CA-342551: Avoid replacing certificate alerts

psafont:
Previously the certificate alerts were replaced every time the script
generating them was run. Now the script only replaces them when the new
alert is different.

### DIFF
--- a/ocaml/alerts/certificate/certificate_check.ml
+++ b/ocaml/alerts/certificate/certificate_check.ml
@@ -40,26 +40,41 @@ let generate_alert epoch (host, expiry) =
     in
     (host, Some (message, alert))
 
-let execute rpc session previous_messages (host, alert) =
+let execute rpc session existing_messages (host, alert) =
   let host_uuid = XenAPI.Host.get_uuid rpc session host in
-  let obsolete_messages =
-    previous_messages
-    |> List.filter_map (fun (ref, record) ->
-           if record.API.message_obj_uuid = host_uuid then
-             Some ref
-           else
-             None)
+  let messages_in_host =
+    List.filter
+      (fun (_, record) -> record.API.message_obj_uuid = host_uuid)
+      existing_messages
   in
-  List.iter
-    (fun self -> XenAPI.Message.destroy rpc session self)
-    obsolete_messages ;
-  match alert with
-  | None ->
-      ()
-  | Some (message, (alert, priority)) ->
+  let is_outdated (ref, record) =
+    match alert with
+    | None ->
+        true
+    | Some (body, (name, priority)) ->
+        record.API.message_body <> body
+        || record.API.message_name <> name
+        || record.API.message_priority <> priority
+  in
+  (* CA-342551: messages need to be deleted if the pending alert regard the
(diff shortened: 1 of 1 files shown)


ocaml/alerts/certificate/certificate_check.ml
```diff
@@ -42,24 +42,36 @@ let generate_alert epoch (host, expiry) =
...
+  let is_current (ref, record) =
+    match alert with
+    | None ->
+        false
+    | Some (body, (name, priority)) ->
+        record.API.message_body = body
+        && record.API.message_name = name
+        && record.API.message_priority = priority
+  in
+  let _, outdated_messages = List.partition is_current host_messages in
```
psafont, line 59: This could be changed by a `List.filter` if it hurts readability. These alerts will most probably be affected by adding the metadata of the whole chain of certificates in the database.

lippirk: I am not seeing the advantage of using partition

robhoes: I also don't see the advantage, since you drop the first return value.

ocaml/alerts/certificate/certificate_check.ml
```diff
@@ -42,24 +42,36 @@ let generate_alert epoch (host, expiry) =
...
+        record.API.message_body <> body
+        || record.API.message_name <> name
+        || record.API.message_priority <> priority
+  in
+  let outdated_messages = List.filter is_outdated host_messages in
+  (* CA-342551: messages should only be replaced if a different message needs
+     to be emmited. This means that messages are not automatically dismissed
+     when a new certificate is introduced; and messages are not replaced by
+     equivalent ones. *)
+  let new_alert_needed = outdated_messages <> [] || host_messages = [] in
```
lindig, line 64: The code above reads a bit confusing: why do we need an alert for something that is outdated? Could use of the word "pending" help? We have pending, yet undelivered, alerts?

psafont: I can understand the confusion, a new alert needs to be emitted iff
1. An alert for this host has been generated (`alert` matches `Some ...`)
2. The alert is not identical to any of the existing alerts, that is `new_alert_needed` matches `true`.

would a name like `replacement_is_pending` for the condition make it less confusing?

lindig: The confusion is coming from `outdated` - it implies we don't need to care. But if we do have to care, `outdated` is not good - but something like `pending` would be better - it implies we still need to do something. Or, `outdated` could mean it represents the past - in which case `history` or `delivered` would make sense.

psafont: The `outdated` messages are not pending, they are indeed outdated/obsoleted/do not represent current state of the system and need to be `XenAPI.Message.destroy`if `alert` contains a message. That is, a replacement is needed.

I don't think `history` or `delivered` makes sense, this is already `previous_messages`.

I'll try to find better names for all of these to make it less confusing all around.

PR #4180, opened 2020-08-13
CP-32138: rely on systemd to have wsproxy available

psafont:
Changes the name of the socket used from wsproxy to ws_proxy as systemd
does not have a capability to take over existing sockets. This would
have been a problem when upgrading as xapi does not clean up the socket
and cannot be controlled in all cases.

Must be merged along https://github.com/xapi-project/wsproxy/pull/23

### DIFF
--- a/ocaml/xapi/console.ml
+++ b/ocaml/xapi/console.ml
@@ -103,29 +103,12 @@ let real_proxy __context _ _ vnc_port s =
     debug "Proxy exited"
   with exn -> debug "error: %s" (ExnHelper.string_of_exn exn)
 
-let check_wsproxy () =
-  try
-    let pid = int_of_string (Unixext.string_of_file "/var/run/wsproxy.pid") in
-    Unix.kill pid 0 ; true
-  with _ -> false
-
-let ensure_proxy_running () =
-  if check_wsproxy () then
-    ()
-  else (
-    ignore
-      (Forkhelpers.execute_command_get_output "/opt/xensource/libexec/wsproxy"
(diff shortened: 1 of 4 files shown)


lindig:
Maybe @lippirk can assist but I believe it should be "Rely on systemd …"

psafont:
I want to test the `ExecStartPre` in the socket file. It should delete the existing socket before systemd starts listening to it.
> ExecStartPre=, ExecStartPost=
> Takes one or more command lines, which are executed before or after the listening sockets/FIFOs are created and bound, respectively. The first token of the command line must be an absolute filename, then followed by arguments for the process. Multiple command lines may be specified following the same scheme as used for ExecStartPre= of service unit files.

lindig:
Can we un-draft this?

psafont:
I have time and a windows VM to test this manually hoping to undraft this very soon

psafont:
When installing the wsproxy and xapi-core rpms into the system the socket needs to be started manually before the service can be started successfully or an error happens:

```
Oct  1 12:45:26 localhost systemd[1]: Got more than one socket.
Oct  1 12:45:26 localhost systemd[1]: wsproxy.service failed to run 'start' task: Invalid argument
```

The good thing is that the execstart for the socket works, not sure how to start the socket as soon as it's installed to avoid this new issue.

psafont:
Almost there... the last issue to overcome is to start `wsproxy.socket` when xapi_core is installed. This can be done manually in the xapi.spec file, but I need to test it.

PR #4182, opened 2020-08-14
HIMN detach fix

robhoes:


lindig (approved):

lindig:
The Travis problem could be ignored:
```
[ERROR] The sources of the following couldn't be obtained, aborting:
7874          - ounit.2.2.2: curl error code 429
7875          - ounit2.2.2.2: curl error co
```

ocaml/xapi/xapi_network.ml
```diff
@@ -146,9 +138,13 @@ let attach_internal ?(management_interface = false) ?(force_bringup = false)
...
 
-let detach ~__context ~bridge_name ~managed =
+let detach ~__context net =
   let dbg = Context.string_of_task __context in
-  if managed && Net.Interface.exists dbg bridge_name then (
+  let bridge_name = net.API.network_bridge in
+  if is_himn net then
+    Xapi_mgmt_iface.reconfigure_himn ~__context ~addr:None ;
+  if net.API.network_managed && Net.Interface.exists dbg bridge_name
+  then (
```
lippirk, line 147: So this is the important bit to get rid of the HIMN bridge. I assume that `deregister_vif` is called during VM shutdown

lippirk (approved):

psafont (approved):

ocaml/xapi/xapi_network.ml
robhoes: Yes, the main point is the `Xapi_mgmt_iface.reconfigure_himn` call, and this comes from `deregister_vif`, which is called for every VIF that gets shutdown, including on VM shutdown (but also hot unplug).

PR #4184, opened 2020-08-17
maintenance: don't use physical equality in migrate_send

lippirk:


stormi:
What are the practical consequences of using those physical equalities?

lindig:
There is similar code in ocaml/xapi/helpers.ml which could be fixed as well

lindig:
Physical equality is almost always the wrong choice. For any value that is not atomic (like integers and tags in algebraic data types) it's likely that `==` will be false but `=` will be true because not the same physical instance is compared.

lindig:
More places to review which probably should use `<>` and not `!=`

```
: xen-api $ git grep --name-only '!=' '**/*.ml' 
ocaml/xapi/helpers.ml
ocaml/xapi/message_forwarding.ml
ocaml/xapi/vm_placement.ml
ocaml/xapi/xapi_host.ml
ocaml/xapi/xapi_services.ml
ocaml/xapi/xapi_vm.ml
ocaml/xapi/xha_interface.ml
ocaml/xxhash/lib/XXHash.ml
```

lippirk:
> More places to review which probably should use `<>` and not `!=`
> 
> ```
> : xen-api $ git grep --name-only '!=' '**/*.ml' 
> ocaml/xapi/helpers.ml
> ocaml/xapi/message_forwarding.ml
> ocaml/xapi/vm_placement.ml
> ocaml/xapi/xapi_host.ml
> ocaml/xapi/xapi_services.ml
> ocaml/xapi/xapi_vm.ml
> ocaml/xapi/xha_interface.ml
> ocaml/xxhash/lib/XXHash.ml
> ```

I'll go and change those other usages :+1:

ocaml/xapi/xapi_host.ml
```diff
@@ -442,7 +442,7 @@ let compute_evacuation_plan_wlb ~__context ~self =
       let target_uuid = List.hd (List.tl detail) in
       let target_host = Db.Host.get_by_uuid ~__context ~uuid:target_uuid in
       if
-        Db.Host.get_control_domain ~__context ~self:target_host != v
+        Db.Host.get_control_domain ~__context ~self:target_host = v
```
lippirk, line 445: oops

ocaml/xapi/helpers.ml
```diff
@@ -793,14 +793,8 @@ let max_version_in_pool : __context:Context.t -> int list =
...
-let rec string_of_int_list : int list -> string = function
-  | [] ->
-      ""
-  | x :: xs ->
-      if xs == [] then
-        string_of_int x
-      else
-        string_of_int x ^ "." ^ string_of_int_list xs
+let string_of_int_list =
+  List.fold_left (fun acc x -> Printf.sprintf "%s.%s" (string_of_int x) acc) ""
```
undu, line 797: I wonder if it's worth using
```
List.map string_of_int intlist |> String.concat "."
```

psafont, line 797: I wonder if it's worth using
```
List.map string_of_int intlist |> String.concat "."
```

ocaml/xapi/vm_placement.ml
```diff
@@ -309,15 +309,17 @@ let select_host_from_categories categories hosts validate_host
...
         in
-        if host != None then
-          host
-        else
-          select hosts_remaining categories_remaining
+        match host with
+        | Some _ ->
+            host
+        | None ->
+            select hosts_remaining categories_remaining
```
psafont, line 321:
```suggestion
        Option.value ~default:(select hosts_remaining categories_remaining) host
```

ocaml/xapi/xapi_host.ml
psafont: Please double-check the change is correct.

> Note: if the vm being moved is dom0 then this is a power management rec and this check does not apply

I understood the comment as dom0 being excluded from the same-host check. Is this reasoning wrong?

ocaml/xapi/helpers.ml
robhoes: I think that it's also reversing the list and appending a `.` at the end. I like Pau's suggestion.

ocaml/xapi/vm_placement.ml
robhoes: Here Pau's suggestion is not correct, because the option is not to be unboxed (`host` is returned, not the arg of `Some`).

ocaml/xapi/vm_placement.ml
```diff
@@ -309,15 +309,17 @@ let select_host_from_categories categories hosts validate_host
...
+    | hosts, category :: categories_remaining -> (
         let host, hosts_remaining =
           select_host_from_category category hosts validate_host
             generate_random_value
         in
-        if host != None then
-          host
-        else
-          select hosts_remaining categories_remaining
+        match host with
```
lindig, line 317:
```
Option.value host ~default:(select hosts_remaining categories_remaining)
```

ocaml/xapi/vm_placement.ml
lindig: I was falling into the same trap. `host` is still an option value, indeed

ocaml/xapi/helpers.ml
lindig:
```

File: ocaml/xapi/helpers.ml [hunk 5]
```diff
@@ -793,8 +793,7 @@ let max_version_in_pool : __context:Context.t -> int list =
   in
   List.fold_left max_version [] versions
 
-let string_of_int_list =
-  List.fold_left (fun acc x -> Printf.sprintf "%s.%s" (string_of_int x) acc) ""
+let string_of_int_list xs = List.map string_of_int xs |> String.concat ""
```
robhoes, line 796: The final string should be `"."`.

psafont: I checked again, this function is unused:

```
$ rg string_of_int_list
ocaml/xapi/helpers.ml
796:let rec string_of_int_list : int list -> string = function
803:        string_of_int x ^ "." ^ string_of_int_list xs
```

lippirk: third time lucky..

ocaml/xapi/helpers.ml
```diff
@@ -793,8 +793,6 @@ let max_version_in_pool : __context:Context.t -> int list =
   in
   List.fold_left max_version [] versions
 
-let string_of_int_list xs = List.map string_of_int xs |> String.concat ""
```
robhoes: 🤣

lippirk: finally nailed it :P

PR #4186, opened 2020-08-18
maintenance: ensure all fistpoints work as expected

lippirk:
The existence of two fistpoints was only ever checked
when the Xapi_fist module is loaded at runtime, and
that result cached. I.e. from xapi's point of view,
these two fistpoints would be either always present,
or always not present. Instead we want to query the
filesystem each time we check for a fistpoint.


lindig:
I agree that it is a bit dubious that fistpoints are only checked at startup rather than dynamically provided the code for the check is executed afterwards

PR #4187, opened 2020-08-18
CP-33780: add pool secret rotation fistpoints

lippirk:


ocaml/xapi/xapi_fist.ml
```diff
@@ -29,6 +29,11 @@ let fistpoint name =
     true
   with _ -> false
 
+let rec hang_while fp =
```
lindig, line 32: Maybe maintain a counter and emit a log line if `counter mod 10 = 0`? This way you can observe from the outside what is happening

lippirk:
I didn't set up CI on this feature branch, so we expect the CI to fail. I have tested the changes and they work as expected

ocaml/xapi/xapi_fist.ml
```diff
@@ -23,16 +23,30 @@ open D
 
 (** {2 (Fill in title!)} *)
 
+let prefix = "/tmp/fist_"
+
 let fistpoint name =
   try
-    Unix.access ("/tmp/fist_" ^ name) [Unix.F_OK] ;
+    Unix.access (Printf.sprintf "%s%s" prefix name) [Unix.F_OK] ;
```
lindig, line 30: The same code is used below. This suggest a function `path_for: string -> string` would be useful that constructs the filename and that has `prefix` as a local binding. Now you would have `Unix.access (path_for name) [Unix.F_OK]`

robhoes (approved):

Issue #4153, opened 2020-07-03
Open port for SDN controller


benjamreis:
Hi!

Would it be possible to add a feature to allow to open specific port for the SDN controller?
I'm willing to do it in a PR, but I'd like to be sure XAPI team is on board with it before putting the work!

So: I want to open the openflow port (TCP 6653) for my SDN controller.
For now it's always opened when a SDN controller is set. That's BAD.

I think a ggod way to answer my need is to allow XAPI's `sdn-controller` to open/close ports specific to SDN protocols (ie: OpenFlow)

What do you think?
I think @robhoes is the one I should ask the question.

Thanks

robhoes:
Hi there, I think extending the `SDN_controller` class (https://xapi-project.github.io/xen-api/classes/sdn_controller.html) to include a list of local ports that need to be opened could be a good solution.

benjamreis:
Yeah that's what I thought.

I'll work on it then!

Thanks for your prompt reply.

PR #4188, opened 2020-08-18
Open/Close ports with SDN Controller

benjamreis:
Adds to `SDN_controller`:
- New field: `opened_ports` the ports by the SDN Controller
- New method:  `open_port` opens a port on all the pool's hosts
- New method: `close_port` closes a port on all the pool's hosts

Solves https://github.com/xapi-project/xen-api/issues/4153

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -5185,6 +5185,11 @@ module SDN_controller = struct
       "pssl", "Passive ssl connection";
     ])
 
+  let sdn_port_protocol = Enum ("sdn_port_protocol", [
+      "tcp", "TCP";
+      "udp", "UDP";
+  ])
+
   let introduce = call
       ~name:"introduce"
       ~doc:"Introduce an SDN controller to the pool."
@@ -5203,6 +5208,30 @@ module SDN_controller = struct
       ~doc:"Remove the OVS manager of the pool and destroy the db record."
       ~params: [ Ref _sdn_controller, "self", "this SDN controller"]
       ~lifecycle
+      ~allowed_roles:_R_POOL_OP
+      ()
+
+  let open_port = call
+      ~name:"open_port"
+      ~doc:"Open a port and add it to the list of opened ports"
+      ~params: [
+        Ref _sdn_controller, "self", "this SDN controller";
+        sdn_port_protocol, "protocol", "The protocol (tcp | udp)";
+        Int, "port", "The port to open"
+      ]
+      ~lifecycle:[Prototyped, rel_next, ""]
+      ~allowed_roles:_R_POOL_OP
+      ()
+
+  let close_port = call
+      ~name:"close_port"
+      ~doc:"Close a port and remove it to the list of opened ports"
+      ~params: [
+        Ref _sdn_controller, "self", "this SDN controller";
+        sdn_port_protocol, "protocol", "the protocol (tcp | udp)";
+        Int, "port", "the port to close"
+      ]
+      ~lifecycle:[Prototyped, rel_next, ""]
       ~allowed_roles:_R_POOL_OP
       ()
 
@@ -5232,10 +5261,15 @@ module SDN_controller = struct
         ; field   ~qualifier:StaticRO ~lifecycle
             ~ty:Int "port" ~default_value:(Some (VInt 0L))
             "TCP port of the controller"
+
+        ; field   ~qualifier:RW ~lifecycle:[Prototyped, rel_next, ""]
+            ~ty:(Set(String)) "opened_ports" ~default_value:(Some (VSet [])) "List of opened ports"
         ]
       ~messages:
         [ introduce
         ; forget
+        ; open_port
+        ; close_port
         ]
       ()
 end
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -5211,21 +5211,21 @@ module SDN_controller = struct
       ~allowed_roles:_R_POOL_OP
       ()
 
-  let open_port = call
-      ~name:"open_port"
-      ~doc:"Open a port and add it to the list of opened ports"
+  let add_local_port = call
+      ~name:"add_local_port"
+      ~doc:"Add a port to the list of required open ports"
       ~params: [
         Ref _sdn_controller, "self", "this SDN controller";
         sdn_port_protocol, "protocol", "The protocol (tcp | udp)";
-        Int, "port", "The port to open"
+        Int, "port", "The required port"
       ]
       ~lifecycle:[Prototyped, rel_next, ""]
       ~allowed_roles:_R_POOL_OP
       ()
 
-  let close_port = call
-      ~name:"close_port"
-      ~doc:"Close a port and remove it to the list of opened ports"
+  let remove_local_port = call
+      ~name:"remove_local_port"
+      ~doc:"Remove a port from the list of required open ports"
       ~params: [
         Ref _sdn_controller, "self", "this SDN controller";
         sdn_port_protocol, "protocol", "the protocol (tcp | udp)";
@@ -5263,13 +5263,13 @@ module SDN_controller = struct
             "TCP port of the controller"
 
         ; field   ~qualifier:RW ~lifecycle:[Prototyped, rel_next, ""]
-            ~ty:(Set(String)) "opened_ports" ~default_value:(Some (VSet [])) "List of opened ports"
+            ~ty:(Set(String)) "local_ports" ~default_value:(Some (VSet [])) "List of required opened ports"
         ]
       ~messages:
         [ introduce
         ; forget
-        ; open_port
-        ; close_port
+        ; add_local_port
+        ; remove_local_port
         ]
       ()
 end
--- a/ocaml/xapi-cli-server/cli_operations.ml
+++ b/ocaml/xapi-cli-server/cli_operations.ml
@@ -7181,24 +7181,6 @@ module SDN_controller = struct
     let uuid = List.assoc "uuid" params in
     let ref = Client.SDN_controller.get_by_uuid rpc session_id uuid in
     Client.SDN_controller.forget rpc session_id ref
-
-  let open_port printer rpc session_id params =
-    let uuid = List.assoc "uuid" params in
-    let ref = Client.SDN_controller.get_by_uuid rpc session_id uuid in
-    let protocol =
-      Record_util.sdn_port_protocol_of_string (List.assoc "protocol" params)
-    in
-    let port = Int64.of_string (List.assoc "port" params) in
-    Client.SDN_controller.open_port rpc session_id ref protocol port
-
-  let close_port printer rpc session_id params =
-    let uuid = List.assoc "uuid" params in
-    let ref = Client.SDN_controller.get_by_uuid rpc session_id uuid in
-    let protocol =
-      Record_util.sdn_port_protocol_of_string (List.assoc "protocol" params)
-    in
-    let port = Int64.of_string (List.assoc "port" params) in
-    Client.SDN_controller.close_port rpc session_id ref protocol port
 end
 
 module PUSB = struct
--- a/ocaml/xapi-cli-server/cli_operations.ml
+++ b/ocaml/xapi-cli-server/cli_operations.ml
@@ -7181,6 +7181,24 @@ module SDN_controller = struct
     let uuid = List.assoc "uuid" params in
     let ref = Client.SDN_controller.get_by_uuid rpc session_id uuid in
     Client.SDN_controller.forget rpc session_id ref
+
+  let open_port printer rpc session_id params =
+    let uuid = List.assoc "uuid" params in
+    let ref = Client.SDN_controller.get_by_uuid rpc session_id uuid in
+    let protocol =
+      Record_util.sdn_port_protocol_of_string (List.assoc "protocol" params)
+    in
+    let port = Int64.of_string (List.assoc "port" params) in
+    Client.SDN_controller.open_port rpc session_id ref protocol port
+
+  let close_port printer rpc session_id params =
+    let uuid = List.assoc "uuid" params in
+    let ref = Client.SDN_controller.get_by_uuid rpc session_id uuid in
+    let protocol =
+      Record_util.sdn_port_protocol_of_string (List.assoc "protocol" params)
+    in
+    let port = Int64.of_string (List.assoc "port" params) in
+    Client.SDN_controller.close_port rpc session_id ref protocol port
 end
 
 module PUSB = struct
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -5295,13 +5295,7 @@ functor
           (sdn_controller_uuid ~__context self)
           (Record_util.sdn_port_protocol_to_string protocol)
           (Int64.to_string port) ;
-        let local_fn = Local.SDN_controller.add_local_port ~self ~protocol ~port in
-        List.iter
-          (fun host ->
-            do_op_on ~local_fn ~__context ~host (fun session_id rpc ->
-                Client.SDN_controller.add_local_port rpc session_id self protocol
-                  port))
-          (Db.Host.get_all ~__context)
+        Local.SDN_controller.add_local_port ~__context ~self ~protocol ~port
 
       let remove_local_port ~__context ~self ~protocol ~port =
         info
@@ -5310,13 +5304,7 @@ functor
           (sdn_controller_uuid ~__context self)
           (Record_util.sdn_port_protocol_to_string protocol)
           (Int64.to_string port) ;
-        let local_fn = Local.SDN_controller.remove_local_port ~self ~protocol ~port in
-        List.iter
-          (fun host ->
-            do_op_on ~local_fn ~__context ~host (fun session_id rpc ->
-                Client.SDN_controller.remove_local_port rpc session_id self protocol
-                  port))
-          (Db.Host.get_all ~__context)
+        Local.SDN_controller.remove_local_port ~__context ~self ~protocol ~port
     end
 
     module PUSB = struct
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -5288,33 +5288,33 @@ functor
           (sdn_controller_uuid ~__context self) ;
         Local.SDN_controller.forget ~__context ~self
 
-      let open_port ~__context ~self ~protocol ~port =
+      let add_local_port ~__context ~self ~protocol ~port =
         info
-          "SDN_controller.open_port: sdn_controller = '%s', protocol = '%s', \
+          "SDN_controller.add_local_port: sdn_controller = '%s', protocol = '%s', \
            port = '%s'"
           (sdn_controller_uuid ~__context self)
           (Record_util.sdn_port_protocol_to_string protocol)
           (Int64.to_string port) ;
-        let local_fn = Local.SDN_controller.open_port ~self ~protocol ~port in
+        let local_fn = Local.SDN_controller.add_local_port ~self ~protocol ~port in
         List.iter
           (fun host ->
             do_op_on ~local_fn ~__context ~host (fun session_id rpc ->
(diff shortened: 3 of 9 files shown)


ocaml/idl/datamodel.ml
```diff
@@ -5206,6 +5211,30 @@ module SDN_controller = struct
       ~allowed_roles:_R_POOL_OP
       ()
 
+  let open_port = call
+      ~name:"open_port"
+      ~doc:"Open a port and add it to the list of opened ports"
+      ~params: [
+        Ref _sdn_controller, "self", "this SDN controller";
+        sdn_port_protocol, "protocol", "The protocol (tcp | udp)";
+        Int, "port", "The port to open"
```
edwintorok, line 5220: Can the port number be hardcoded instead? Adding the ability to open arbitrary ports won't sit well with the security team.

benjamreis: I thought about, like having a map and specifying a key; for instance `openflow` -> `tcp:6653`.
But I'm not sure how to have an exhaustiv list of all the ports that could be used by a sdn controller.

ocaml/xapi/message_forwarding.ml
```diff
@@ -5287,6 +5287,20 @@ functor
         info "SDN_controller.forget: sdn_controller = '%s'"
           (sdn_controller_uuid ~__context self) ;
         Local.SDN_controller.forget ~__context ~self
+
+      let open_port ~__context ~self ~protocol ~port =
+        info "SDN_controller.open_port: sdn_controller = '%s', protocol = '%s', port = '%s'"
+          (sdn_controller_uuid ~__context self)
+          (Record_util.sdn_port_protocol_to_string protocol)
+          (Int64.to_string port);
+        Local.SDN_controller.open_port ~__context ~self ~protocol ~port
```
edwintorok, line 5296: This would open the port on only one host (the one receiving the API call, which may be a slave).
Do you need this port opened everywhere?

ocaml/idl/datamodel.ml
edwintorok: Perhaps it would also be useful to enable this only if a feature flag is enabled (see features.ml/features.mli). That way if the feature is not enabled you are not exposed to the security risks of having additional ports open.

ocaml/xapi/message_forwarding.ml
benjamreis: Yes I noticed it while I was testing. I need to do it on all the pools host indeed but I think I know how since I dit it [here](https://github.com/xapi-project/xen-api/pull/4164/files#diff-9a639b670fb7e2432894cc91df8b27fbR3370-R3371)

ocaml/idl/datamodel.ml
benjamreis: Indeed.

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -7181,6 +7181,20 @@ module SDN_controller = struct
...
+    let protocol = Record_util.sdn_port_protocol_of_string (List.assoc "protocol" params) in
+    let port = Int64.of_string (List.assoc "port" params) in
+    Client.SDN_controller.open_port rpc session_id ref protocol port
+
+  let close_port printer rpc session_id params =
+    let uuid = List.assoc "uuid" params in
+    let ref = Client.SDN_controller.get_by_uuid rpc session_id uuid in
+    let protocol = Record_util.sdn_port_protocol_of_string (List.assoc "protocol" params) in
+    let port = Int64.of_string (List.assoc "port" params) in
+    Client.SDN_controller.close_port rpc session_id ref protocol port
```
lippirk, line 7197: I suspect you need to run `make format` for this section

benjamreis: Probably, I always run it after I'm done with my dev. :)

lippirk:
Please can you add some details to the commit message, e.g. the motivation for the patch

benjamreis:
Xen Orchestra's SDN controller has a new feature using OpenFlow.
The OpenFlow port used to talk to an host's OpenVSwitch is `tcp:6653`.
I'd like to add the possibility to open/close port with XAPI's `SDN_controller` so that any SDN Controller can opens/closes the ports it needs.

I discussed it with @robhoes in https://github.com/xapi-project/xen-api/issues/4153

lippirk:
> Xen Orchestra's SDN controller has a new feature using OpenFlow.
> The OpenFlow port used to talk to an host's OpenVSwitch is `tcp:6653`.
> I'd like to add the possibility to open/close port with XAPI's `SDN_controller` so that any SDN Controller can opens/closes the ports it needs.
> 
> I discussed it with @robhoes in #4153

I know, it'd be nice to see this information in the commit message however :)

benjamreis:
Yes, I'll put it when I rebase and squash my commits. :)

ocaml/xapi-cli-server/cli_operations.ml
benjamreis: I'm wondering, are these modifications necessary?
I'm not sure I understand the uses of this file since the method is not available in `xe` cli.

and if necessary, should I iter through all the pools host like I do in `message_forwarding.ml`?

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -7181,6 +7181,24 @@ module SDN_controller = struct
...
+    Client.SDN_controller.open_port rpc session_id ref protocol port
+
+  let close_port printer rpc session_id params =
+    let uuid = List.assoc "uuid" params in
+    let ref = Client.SDN_controller.get_by_uuid rpc session_id uuid in
+    let protocol =
+      Record_util.sdn_port_protocol_of_string (List.assoc "protocol" params)
+    in
+    let port = Int64.of_string (List.assoc "port" params) in
+    Client.SDN_controller.close_port rpc session_id ref protocol port
```
benjamreis, line 7201: I'm wondering, are these modifications necessary?
I'm not sure I understand the uses of this file since the method is not available in xe cli.

and if necessary, should I iter through all the pools host like I do in message_forwarding.ml?

benjamreis: Is it possible to iterate on all pool's host to call `Client.SDN_controller.close_port rpc session_id ref protocol port`?
In `message_forwarding` there's `do_op_on` but I didn't find something similar in this file.

robhoes:
I think that it would be good to have a bit more discussion on the design for this before diving into implementation details. What is the intended use case?

Here are some questions:
* The current design has a list of opened ports and functions to open and close ports. Does this put the onus on a client to open and close ports at the right time?
* What happens if a host reboots? Is the port going to be automatically reopened or does the client need to reopen it?
* From a security point of view, are we comfortable with a client being able to open/close any port on all hosts in a pool at the same time?
* We have ports being opened in other places in the code, including in the code that manager SDN controllers: https://github.com/xapi-project/xen-api/blob/master/scripts/plugins/openvswitch-config-update. How does this new feature relate to that?

I would be in favour of a design where the client expresses which ports are needed for the controller by setting a field in the `SDN_controller` object, and letting xapi handle the opening and closing of ports, at the right times and while taking a holistic view of the state of the host firewalls to manage possible conflicts. Also, we should be able to restrict the range of ports that can be opened?

ocaml/xapi/message_forwarding.ml
```diff
@@ -5287,6 +5287,36 @@ functor
...
+          (sdn_controller_uuid ~__context self)
+          (Record_util.sdn_port_protocol_to_string protocol)
+          (Int64.to_string port) ;
+        let local_fn = Local.SDN_controller.close_port ~self ~protocol ~port in
+        List.iter
+          (fun host ->
+            do_op_on ~local_fn ~__context ~host (fun session_id rpc ->
+                Client.SDN_controller.close_port rpc session_id self protocol
+                  port))
+          (Db.Host.get_all ~__context)
```
lippirk, line 5319: Have you tested that both of these functions are idempotent?

For example, suppose you run `SDN_controller.close_port`, and it succeed on host A,B, but fails on C. Then the user will probably want to run `SDN_controller.close_port` again once they have fixed the problem on C - will that work?

(In addition we probably want to have the loops closing/opening ports to close/open as many ports as possible, and then raise any exceptions later)

benjamreis:
> I think that it would be good to have a bit more discussion on the design for this before diving into implementation details. What is the intended use case?
Sure, at first my need was top open the port `tcp:6653` for XO's SDN Controller to use `openflow`
 with the hosts.
First I patched [this code](https://github.com/xapi-project/xen-api/blob/master/scripts/plugins/openvswitch-config-update#L223-L248) to open the port whenever a SDN Controller is set.
But if the SDN Controller does not use Openflow then the port is open for no reason, which is bad.
Therefore I think it's best to allow a XAPI client to open/close is only when necessary.

> 
> Here are some questions:
> 
>     * The current design has a list of opened ports and functions to open and close ports. Does this put the onus on a client to open and close ports at the right time?
Yes

> 
>     * What happens if a host reboots? Is the port going to be automatically reopened or does the client need to reopen it?
If the host reboot, the opened port would stay open with the current code.

> 
>     * From a security point of view, are we comfortable with a client being able to open/close any port on all hosts in a pool at the same time?
I understand how that could be a problem. My need for openflow requires all host's port to be opened but perhaps some scenarii doesn't require that. Maybe a I can a host to be specified in the method so that it can be selected on which host this needs to be done.

> 
>     * We have ports being opened in other places in the code, including in the code that manager SDN controllers: https://github.com/xapi-project/xen-api/blob/master/scripts/plugins/openvswitch-config-update. How does this new feature relate to that?
My code use the same mechanism: the `/etc/xapi.d/plugins/firewall-port` script.

> 
> 
> I would be in favour of a design where the client expresses which ports are needed for the controller by setting a field in the `SDN_controller` object, and letting xapi handle the opening and closing of ports, at the right times and while taking a holistic view of the state of the host firewalls to manage possible conflicts. Also, we should be able to restrict the range of ports that can be opened?
The problem here is that XAPI don't know when the OpenFlow port would be needed to be open for instance.
Restricting the allowed port range to be opened is a good idea indeed. I also thought of providing a determined list of port that could be open, but I'm not sure it's possible to know in advance all the port that could be need by a SDN controller (the lis could be expended later though if someone needs it)

ocaml/xapi/xapi_sdn_controller.ml
```diff
@@ -69,3 +69,27 @@ let forget ~__context ~self =
     List.iter
       (fun host -> Helpers.update_vswitch_controller ~__context ~host)
       (Db.Host.get_all ~__context)
+
+let open_port ~__context ~self ~protocol ~port =
+  Pool_features.assert_enabled ~__context ~f:Features.Sdn_port_management ;
```
lippirk, line 74: Putting this `Pool_features.assert_enabled` here means that we are checking the feature flag `n` times, where `n` is the number of hosts in the pool. For Citrix Hypervisor, I believe this is unnecessary and that you can move this assert to `message_forwarding.ml` so that we only check it once (however I'm not sure how you check the xapi feature flags at xcp-ng)

benjamreis: Indeed! Silly me.

ocaml/xapi/message_forwarding.ml
benjamreis: Yes it'd work, you can call several times `open_port`/`close_port` without any problem, even if the port is already opened/closed.

So you think i should catch the error in the loop and throw it after the loop completion? What if several iterations fail, should i throw as many error?

robhoes:
I am uncomfortable with this `open/close/opened_ports` design also because there is a risk of getting out of sync with reality. The `opened_ports` fields reflects the ports that the controller wants to be open, but it can't be relied upon to find the ports that are actually open on each host of the pool. This is because TCP/UDP ports are not exclusively owned by the SDN-controller sub-system.

Fields in the API usually fall into two categories:
* Input fields, which state a desired configuration that the Toolstack is supposed to implement. An example is the `VIF.MAC` field, which has the MAC that a VIF is supposed to get whenever is it "attached" to a VM (usually when the VM starts).
* Output fields, which reflect the current state of the system. For example the `VM.resident_on` field, which is updated by xapi to indicate on which host a VM is running. Many such fields are kept in sync when xapi restarts by an important part of the startup sequence called `dbsync`.

The `opened_ports` field looks like the latter, except that it isn't kept up-to-date by xapi. It could also be considered an input field, but the Toolstack does not act on it, so it really isn't. 

> But if the SDN Controller does not use Openflow then the port is open for no reason, which is bad.
> Therefore I think it's best to allow a XAPI client to open/close is only when necessary.
> (...)
> The problem here is that XAPI don't know when the OpenFlow port would be needed to be open for instance.

What "open when needed" means is a little open for interpretation. A while back we had the case of the VXLAN port that needed to be opened, and we added that to the VXLAN-specific datamodel objects, and open/close code was added to the relevant PIF plug/unplug code. This worked nicely for this use case. I assumed that the ports that are part of the `SDN_controller` configuration itself should be opened whenever the controller config is instantiated. There is one port opened already in `openvswitch-config-update` (in the code that you mentioned).

Then, going back to the beginning, it is indeed bad to open e.g. the OpenFlow port for all controllers unconditionally. But if we have a per-controller field that states precisely which ports this controller needs, then it is still a conscious choice by the administrator to open those ports. There won't be any surprises, even if the ports are open all the time. In that case, we just need a new field, without the open and close functions in the API, as the ports will be opened when `openvswitch-config-update` runs.

> Restricting the allowed port range to be opened is a good idea indeed. I also thought of providing a determined list of port that could be open, but I'm not sure it's possible to know in advance all the port that could be need by a SDN controller (the lis could be expended later though if someone needs it)

Perhaps we could add a config file option for the allowed port ranges that may be used.

benjamreis:
I understand your point. So you're suggesting, if I understood correctly, 
- to have a list of ports to open for the sdn controller stored in a config file
- whenever a sdn controller is set, open this ports
- close them when it's unset. 

Is that right?

robhoes:
I am proposing:
* A new field in the `SDN_controller` class for additional ports that the controller requires. There are already `address/port/protocol` fields, which are the connection details for the remote (controller) side, and the OVS uses these to connect to the controller. The new field could be called `local_ports` of type `string set` for the additional local ports that need to be opened. Entries could be of the form `tcp:xxx`, as you did in the current patch, or just numbers if we don't need UDP.
* A config file option to restrict which ports can be entered for `local_ports`. It could be a list of allowed port ranges. We need to think about a useful format.
* Opening/closing the ports in `openvswitch-config-update`.

benjamreis:
I think I can start with point 1 & 3 and meanwhile we can find the formats of the allowed ports config and I'll add it then. :)

benjamreis:
I assume my `<open/close>_port` method should become `<add/remove>_local_port`

benjamreis:
Should I keep the feature to enable it or the config file is enough?

robhoes:
If we keep the default for the allowed ports list (settable in the config file) to be "empty", then IMO this is sufficient, as it would not let anyone open ports through the API unless the config file on the host is edited.

ocaml/xapi/message_forwarding.ml
```diff
@@ -5287,6 +5287,36 @@ functor
...
+          (sdn_controller_uuid ~__context self)
+          (Record_util.sdn_port_protocol_to_string protocol)
+          (Int64.to_string port) ;
+        let local_fn = Local.SDN_controller.add_local_port ~self ~protocol ~port in
+        List.iter
+          (fun host ->
+            do_op_on ~local_fn ~__context ~host (fun session_id rpc ->
+                Client.SDN_controller.add_local_port rpc session_id self protocol
+                  port))
+          (Db.Host.get_all ~__context)
```
benjamreis, line 5304: I think this part is no longer required since the port opening/closing is hanfled by `openvswitch-config-update`

robhoes: If the operation is just a DB update, then you don't need any forwarding, and just execute the `Local` functions. However, think about whether there is a need to do "live" updates or not.

scripts/plugins/openvswitch-config-update
```diff
@@ -91,6 +92,16 @@ def update(session, args):
...
 
+    # Open port required by the SDN Controller
+    if controller_local_ports:
+        for local_port in controller_local_ports:
+            splitted = local_port.split(':')
+            protocol = splitted[0]
+            port = splitted[1]
+            cmd = [iptablectl, 'open', port, protocol]
+            output = subprocess.Popen(cmd, stdout=subprocess.PIPE).communicate()
+
```
benjamreis, line 103: I'm not sure I put it in the right place. Just tried it and the port didn't open.
I thought the `update` method of this script would be called when the SDN Controller fields was updated.

benjamreis:
Hmm, I'm not sure how to close a port after it's removed from `SDN_controller.local_ports`.
Should I close all allowed ports and then re-open the ones in the `local_ports` field? Not very efficient IMHO.

ocaml/idl/datamodel.ml
```diff
@@ -5232,10 +5261,15 @@ module SDN_controller = struct
         ; field   ~qualifier:StaticRO ~lifecycle
             ~ty:Int "port" ~default_value:(Some (VInt 0L))
             "TCP port of the controller"
+
+        ; field   ~qualifier:RW ~lifecycle:[Prototyped, rel_next, ""]
```
robhoes, line 5265: I think this needs to have `DynamicRO` and and explicit `set_local_ports` function. With `RW`, you'll get that function generated for you, but it would only set the DB field. You need to override it to set the DB _and_ update the OVS. You have add/remove functions above, but they don't match the field name (which has "ports" not "port"). It may be easier to not have the add/remove functions, but just `set_local_ports`.

robhoes: However, I see now that you've added better typed parameters to the add/remove functions, with a protocol enum. Let's leave it like that, but we still need the field to have `DynamicRO` to avoid the auto-generated setters.

benjamreis: Yes, I like this design.

ocaml/xapi/message_forwarding.ml
```diff
@@ -5287,6 +5287,24 @@ functor
...
           (sdn_controller_uuid ~__context self) ;
         Local.SDN_controller.forget ~__context ~self
+
+      let add_local_port ~__context ~self ~protocol ~port =
+        info
+          "SDN_controller.add_local_port: sdn_controller = '%s', protocol = '%s', \
+           port = '%s'"
+          (sdn_controller_uuid ~__context self)
+          (Record_util.sdn_port_protocol_to_string protocol)
+          (Int64.to_string port) ;
```
robhoes, line 5297: You could use `%Ld` to avoid `Int64.to_string` (see `introduce` above).

robhoes:
> Hmm, I'm not sure how/when to close a port after it's removed from SDN_controller.local_ports.
> Should I close all allowed ports and then re-open the ones in the local_ports field? Not very efficient IMHO.

Tricky indeed, because we don't want to close ports from other services by mistake. We may need to store somewhere which ports are currently opened for OVS, so we can check which ones to close.

Also, are you still adding the config file option to restrict the ports that can be opened?

benjamreis:
> > Hmm, I'm not sure how/when to close a port after it's removed from SDN_controller.local_ports.
> > Should I close all allowed ports and then re-open the ones in the local_ports field? Not very efficient IMHO.
> 
> Tricky indeed, because we don't want to close ports from other services by mistake. We may need to store somewhere which ports are currently opened for OVS, so we can check which ones to close.
> 
> Also, are you still adding the config file option to restrict the ports that can be opened?
Yes I still have to add it, didn't think about how yet, haven't got the time. But yeah it's still in the pipes.

lindig:
This has received a lot of comments and updates but none recently. I suspect most people lost track where this is standing. Can we close this and open a new PR or get a summary what is left to do?

benjamreis:
Sure, I can't work on it for now, too much time taken by IPv6.
I'll reopen a PR once I'm back at it, with all we've said in the comments.

psafont:
Marking the PR as draft, feel free to address comments and undraft it

robhoes:
This is really old... does anyone still care about it? I'll close it, because it would be quite out of date now.

PR #4190, opened 2020-08-20
Block PSR if pool operations are running, and vice versa

lippirk:
We want to avoid the master changing whilst a pool secret
rotation is pending, so in particular we should block
pool-designate-new-master.

We do _not_ block pool-emergency-transition-to-master, but note
that it should not be used lightly.

### DIFF
--- a/ocaml/xapi/xapi_ha.ml
+++ b/ocaml/xapi/xapi_ha.ml
@@ -1677,15 +1677,6 @@ let enable __context heartbeat_srs configuration =
   let pool = Helpers.get_pool ~__context in
   if Db.Pool.get_ha_enabled ~__context ~self:pool then
     raise (Api_errors.Server_error (Api_errors.ha_is_enabled, [])) ;
-  if Xapi_pool_helpers.pool_secret_rotation_pending ~__context then
-    raise
-      Api_errors.(
-        Server_error
-          ( internal_error
-          , [
-              "a pool secret rotation is pending, please complete it before \
-               attempting to enable HA"
-            ] )) ;
   Pool_features.assert_enabled ~__context ~f:Features.HA ;
   (* Check that all of our 'disallow_unplug' PIFs are currently attached *)
   let unplugged_ununpluggable_pifs =
--- a/ocaml/xapi/xapi_pool.ml
+++ b/ocaml/xapi/xapi_pool.ml
@@ -1651,7 +1651,7 @@ let designate_new_master ~__context ~host =
     if Db.Pool.get_ha_enabled ~__context ~self:pool then
       raise (Api_errors.Server_error (Api_errors.ha_is_enabled, [])) ;
     (* Only the master can sync the *current* database; only the master
-       		   knows the current generation count etc. *)
+       knows the current generation count etc. *)
     Helpers.call_api_functions ~__context (fun rpc session_id ->
         Client.Pool.sync_database rpc session_id) ;
     let all_hosts = Db.Host.get_all ~__context in
--- a/ocaml/xapi/xapi_pool_helpers.ml
+++ b/ocaml/xapi/xapi_pool_helpers.ml
@@ -20,7 +20,10 @@ open Db_filter
 open Record_util
 open Api_errors
 
-let all_operations = [`ha_enable; `ha_disable; `cluster_create]
+(* psr is not included in this list because it can be considered in progress
+   in between api calls (i.e. wrapping it inside with_pool_operation won't work) *)
+let all_operations =
+  [`ha_enable; `ha_disable; `cluster_create; `designate_new_master]
 
 (** Returns a table of operations -> API error options (None if the operation would be ok) *)
 let valid_operations ~__context record _ref' =
@@ -42,6 +45,7 @@ let valid_operations ~__context record _ref' =
       (`ha_enable, Api_errors.ha_enable_in_progress, [])
     ; (`ha_disable, Api_errors.ha_disable_in_progress, [])
     ; (`cluster_create, Api_errors.cluster_create_in_progress, [])
+    ; (`designate_new_master, Api_errors.designate_new_master_in_progress, [])
     ]
   in
   List.iter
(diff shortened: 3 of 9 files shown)


ocaml/xapi/xapi_pool.ml
```diff
@@ -1646,6 +1646,15 @@ let sync_database ~__context =
 
 (* This also means me, since call will have been forwarded from the current master *)
 let designate_new_master ~__context ~host =
+  if Xapi_pool_helpers.pool_secret_rotation_pending ~__context then
```
edwintorok, line 1649: Should you also block pool secret rotation when designate new master is pending? Usually this is handled by using allowed operations.
Perhaps all these designate new master, and PSR should gain an operation type in xapi_pool_helpers, and then you can easily declare in one place that they are mutually exclusive.

edwintorok: Also I assume you already block pool secret rotation when HA is enabled? (master transitions can happen unpredictably and they cannot be blocked by design: if old master dies then another pool member must be promoted to master)

lippirk: We discussed adding a PSR operation to xapi pool helpers in this PR (https://github.com/xapi-project/xen-api/pull/4162). We elected to use a flag in the database instead, because a PSR can be pending in between calls to Pool.secret_rotate (which didn't fit nicely with the semantics of current operations - in particular you can't wrap `Xapi_psr.start` inside `with_pool_operation` and expect it to work).

I'll add designate_new_master to xapi_pool_helpers, that's a good idea.

(and yes, HA and PSR are already mutually exclusive)

ocaml/xapi/xapi_ha.ml
```diff
@@ -1668,6 +1668,7 @@ let disable __context =
   let pool = Helpers.get_pool ~__context in
   if not (Db.Pool.get_ha_enabled ~__context ~self:pool) then
     raise (Api_errors.Server_error (Api_errors.ha_not_enabled, [])) ;
+  Xapi_pool_helpers.assert_designate_new_master_not_in_progress ~__context ;
```
edwintorok, line 1671: I think this is redundant, you should've been blocked from starting the designate operation if HA is enabled. Also HA does designate new masters on its own.

ocaml/xapi/xapi_pool.ml
```diff
@@ -1646,12 +1646,14 @@ let sync_database ~__context =
 
 (* This also means me, since call will have been forwarded from the current master *)
 let designate_new_master ~__context ~host =
+  Xapi_pool_helpers.assert_pool_secret_rotation_not_pending ~__context ;
```
edwintorok, line 1649: I think that the allowed operation checking should take care of this instead of having to add N checks for each of the N operation implementations. So these are redundant.
HA is enabled is the exception because it is a permanent state rather than a single operation.

ocaml/xapi/xapi_pool_helpers.mli
```diff
@@ -32,7 +32,11 @@ val ha_disable_in_progress : __context:Context.t -> bool
 
 val ha_enable_in_progress : __context:Context.t -> bool
 
-val pool_secret_rotation_pending : __context:Context.t -> bool
+val assert_pool_secret_rotation_not_pending : __context:Context.t -> unit
+
+val assert_designate_new_master_not_in_progress : __context:Context.t -> unit
+
+val assert_cluster_create_not_in_progress : __context:Context.t -> unit
```
edwintorok, line 39: These are probably redundant, the allowed operations checking in with_pool_operation should call the allowed operation checker and reject things there.

psafont (approved):

ocaml/xapi/xapi_pool.ml
edwintorok: See 
```
let with_pool_operation ~__context ~self ~doc ~op f =
  let task_id = Ref.string_of (Context.get_task_id __context) in
  Helpers.retry_with_global_lock ~__context ~doc (fun () ->
      assert_operation_valid ~__context ~self ~op ;
      Db.Pool.add_to_current_operations ~__context ~self ~key:task_id ~value:op) 
```

As long as you taught the assert_operation_valid to reject the incompatible concurrent ops (which you did by declaring the operation and adding it to the list of incompatible ops) that is all you really need to do.

lippirk:
Thanks for the comments @edwintorok, it's a lot simpler now (I tested and those asserts between HA and designate new master were indeed redundant). We can also move the assert that a PSR is not pending inside `with_pool_operations` to simply even further

ocaml/xapi/xapi_pool.ml
lippirk: I think you're confusing HA enable (the operation) and HA enabled (the state); nonetheless there is some existing code which wants to know if HA enable is in progress, so I'll leave these functions here

edwintorok (approved):

ocaml/xapi/xapi_pool_helpers.ml
```diff
@@ -112,6 +116,9 @@ let update_allowed_operations ~__context ~self : unit =
 (** Add to the Pool's current operations, call a function and then remove from the
     current operations. Ensure the allowed_operations are kept up to date. *)
 let with_pool_operation ~__context ~self ~doc ~op f =
+  if Db.Pool.get_is_psr_pending ~__context ~self:(Helpers.get_pool ~__context)
```
edwintorok, line 119: Wouldn't the assert_operation_valid below make this one redundant too?

edwintorok: Or is it becasue PSR is more like HA: instead of one long running API call you have something that starts it

lippirk: PSR isn't a 'proper' `pool_operation` so isn't in the `valid_operations` table, therefore we have to add the check specifically. However I'll move the check to `assert_operation_valid` because that makes more sense (this won't affect  behaviour anywhere else because this is the only usage)

PR #4191, opened 2020-09-01
CA-322708 - VM should not be allowed to start during dead storage migration

GeorgeBurac:
During VM storage migration it was possible to start the VM.

This leads to errors (VDI in use and illegal storage transition).
To resolve this issue the start and start_on operation were removed from the allowed_operations list in the is_allowed_concurrently function.
This removes the Start and Start On operations from the XenCenter vm menu, thus preventing the user from encountering the errors.

### DIFF
--- a/ocaml/xapi/xapi_vm_lifecycle.ml
+++ b/ocaml/xapi/xapi_vm_lifecycle.ml
@@ -115,8 +115,6 @@ let is_allowed_concurrently ~(op : API.vm_operations) ~current_ops =
     ; ([`migrate_send], `metadata_export)
     ; ([`migrate_send], `clean_shutdown)
     ; ([`migrate_send], `clean_reboot)
-    ; ([`migrate_send], `start)
-    ; ([`migrate_send], `start_on)
     ]
   in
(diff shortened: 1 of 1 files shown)


lindig:
If `start` is removed, `start_on` should be removed as well. But whether this is a good solution overall, I have to look deeper.

lippirk:
I think the commit message should explain the issue that xapi has, and why this changes fixes it

robhoes:
I agree with both comments above. Apart from that, this should be all that is required.

lindig:
Unrelated to this change: I don't like the implementation of `is_allowed_concurrently` or the problem is more complicated than I thought. We have a set (list) of current operations `ops` and the question is whether we `op` can be executed in addition. Is this different from asking whether `op` can be executed in parallel to each `o` in `ops`? `List.for_all (fun o -> allowed o op)` ops` and `allowed` should be implemented with pattern matching.

lippirk (approved):
Perhaps have a maximum of 72 characters per line in your commit message

robhoes (approved):

PR #4192, opened 2020-09-01
CP-33823 use OCaml for self signing SSL certs

lindig:
This replaces the bash script that invokes openssl for generating SSL certificates (/etc/xensource/xapi-ssl.pem). The certificate is now generated directly using an OCaml X509 implementation. Before this can be merged, corresponding changes in xapi.spec need to be merged.


ocaml/gencert/selfcert.ml
```diff
@@ -0,0 +1,125 @@
...
+    [X509.Distinguished_name.(Relative_distinguished_name.singleton (CN name))]
+  in
+  let req = X509.Signing_request.create issuer privkey in
+  sign days privkey pubkey issuer req alt_names >>= fun cert ->
+  let cert_pem = X509.Certificate.encode_pem cert in
+  let key_pem = X509.Private_key.encode_pem privkey in
+  write_certs certfile key_pem cert_pem
+
+let host name alt_names pemfile =
+  let expire_days = 3650 in
```
edwintorok, line 86: I assume the 10 years here is meant to match our LTSR lifetime, so if your certs expire you should at least upgrade by that time, right?

What happens if someone's clocks go wrong though on boot (e.g. failing NVRAM battery, and no internet connection to synchronize the clock). Would that prevent the pool members from communicating with each-other (we do recommend using NTP, which should mostly solve this issue). I think we have to be sure to run the certificate generation *after* we got a stable time from NTP though.

lindig: This is what xapi does so far:
```
ocaml/gencert/selfcert.ml:  let expire_days = 3650 in
```

robhoes:
I would prefer to do this on a feature branch first, together with related follow-up work.

lindig:
Let's do this on a feature branch.

PR #4194, opened 2020-09-02
Get CC_PREPARATIONS from xs-inventory

lippirk:
The xapi daemon is not passed the CC_PREPARATIONS env var,
so stunnel fips mode could not be enabled as expected.
Instead get CC_PREPARATIONS from /etc/xensource-inventory.


ocaml/xapi/helpers.ml
```diff
@@ -1700,17 +1700,23 @@ end = struct
...
-        | Some _ | None ->
-            false
+        Xapi_inventory.lookup ~default:"false" "CC_PREPARATIONS" |> function
+        | "true" ->
+            "yes"
+        | "false" ->
+            "no"
+        | unknown ->
+            info
+              "CC_PREPARATIONS has unexpected value '%s'. assuming fips mode \
```
lindig, line 1710: The error message could include that either `true` or `false` are expected.

lindig (approved):

robhoes (approved):

PR #4196, opened 2020-09-03
CP-33823 use OCaml to self-sign certificates

lindig:
* Remove genereate_ssl_cert
* implement signing in OCaml
* still use openssl to generate RSA keys


ocaml/xapi/xapi_globs.ml
```diff
@@ -1183,6 +1183,9 @@ module Resources = struct
     ; ( "generate_ssl_cert"
       , Constants.generate_ssl_cert
       , "script to generate SSL certificates to be used by XAPI" )
```
lippirk, line 1185: This needs updating?

ocaml/gencert/selfcert.mli
```diff
@@ -0,0 +1,9 @@
+val host : string -> string list -> string -> unit
+(** [certify hostname alt_names path] creates (atomically) a PEM file at
+[path] for hostname with alternative names [alt_names]. *)
+
+module CLI : sig
+  val main : unit -> unit
```
lippirk, line 6: I expected to see this binary installed in the Makefile

lindig: This is not installed. I could remove it as I'm not building the stand-alone binary.

ocaml/xapi/xapi_globs.ml
lippirk: I think we should update the doc string (and perhaps rename `generate_ssl_cert` -> `gencert`)

ocaml/gencert/selfcert.ml
```diff
@@ -0,0 +1,165 @@
...
+let selfsign name alt_names length days certfile =
+  let rsa =
+    try
+      generate_private_key length
+      |> Cstruct.of_string
+      |> X509.Private_key.decode_pem
+      |> R.get_ok
+      |> function
+      | `RSA x ->
+          x
```
psafont, line 108: This will fail without giving useful information about the cause, how about

```suggestion
      |> Result.fold
           ~ok:(function `RSA x -> x)
           ~error:(fun (`Msg e) -> failwith e)

```

ocaml/gencert/selfcert.ml
```diff
@@ -0,0 +1,165 @@
...
+    with e ->
+      D.error "generating RSA key for %s failed: %s" certfile
+        (Printexc.to_string e) ;
+      failwith (Printf.sprintf "generating RSA key for %s failed" certfile)
```
psafont, line 112: Would the gencert binary log correctly the error message? This could be used instead to make sure it's shown everywhere
```suggestion
      let err_msg =
        Printf.sprintf "generating RSA key for %s failed: %s" certfile
          (Printexc.to_string e)
      in
      D.error "%s" err_msg ; failwith err_msg
```

ocaml/gencert/selfcert.ml
```diff
@@ -0,0 +1,165 @@
...
+  | None ->
+      R.error_msgf "can't represent %d as time span" days
+
+let add_dns_names extension = function
+  | [] ->
+      extension
+  | names ->
+      X509.Extension.(
+        add Subject_alt_name
+          (false, X509.General_name.(singleton DNS names))
```
psafont, line 60: Do you have any information on what the `false` means here, X509 documentation was not helpful when I checked.

ocaml/gencert/selfcert.ml
```diff
@@ -14,23 +14,28 @@
...
+
+(** initialize the random number generator at program startup when this
+module is loaded. *)
+let () = Mirage_crypto_rng_unix.initialize ()
 
 let defer f = Fun.protect ~finally:f
 
 (** [write_cert] writes a PEM file to [path]. It attempts to do that
   atomically by writing to a temporary file in the same directory first and
   renaming the file at the end *)
```
psafont: Could you mention that the file is a PKCS12? (private key + certificate)

ocaml/gencert/selfcert.ml
```diff
@@ -14,23 +14,28 @@
...
 let write_certs path key cert =
-  let delimit = "\n" in
+  let delim = "\n" in
   let temp_dir = Filename.dirname path in
   let tmp = Filename.temp_file ~temp_dir "certify-" ".tmp" in
   try
     let fd = Unix.openfile tmp [Unix.O_WRONLY] 0o600 in
```
psafont, line 36: This should be 0o400, like in https://github.com/xapi-project/xen-api/blob/master/ocaml/gencert/lib.ml#L165

I'm not happy that there are two functions for writing a PKCS12 to disk, could you change this function so the installation one in lib.ml uses this one? i would also prefer if `UX.atomic_write_to_file` was used as well to avoid  duplicated functionality.

lindig:
I've addressed the comments. While it would be safe to push this to master we can also put this on a feature branch.

ocaml/xapi/xapi_host.ml
```diff
@@ -1400,10 +1399,10 @@ let install_server_certificate ~__context ~host ~certificate ~private_key
   Xapi_mgmt_iface.reconfigure_stunnel ~__context
 
 let emergency_reset_server_certificate ~__context =
-  let args =
-    [!Xapi_globs.server_cert_path; Helper_hostname.get_hostname (); "--force"]
-  in
-  ignore @@ Gencert.call_generate_ssl_cert ~args ;
+  let xapi_ssl_pem = !Xapi_globs.server_cert_path in
+  let common_name = Helper_hostname.get_hostname () in
+  let alt_names = [] in
```
psafont, line 1404: I think  Gencert.get_hostnames would better as it would give us alt names as well.

I would also be tempted to replace the implementation of get_hostname with the get_hostnames, but I think that's out of scope of this PR

lindig: That function is not in the libray but in the binary - so would have to move it.

ocaml/gencert/selfcert.ml
lindig: I don't. I looked at the code here but I agree that it is not obvious. The generated certs look correct to me:

https://github.com/yomimono/ocaml-certify/blob/primary/src/common.ml#L22-L27

psafont (approved):

ocaml/gencert/selfcert.ml
lindig: The parameter identifies "critical" extensions and I believe this parameter is not used here.

PR #4197, opened 2020-09-03
Give xapi the option of generating pool secrets using a script

lippirk:
- remove genptoken & genptoken.service
- generate pool secrets optionally via script


ocaml/xapi/helpers.ml
```diff
@@ -1804,5 +1782,79 @@ end = struct
...
+              ] )) ;
+    SecretString.of_string x
+
+  let _make () =
+    match !Xapi_globs.pool_secret_gen_method with
+    | `script ->
+        call_script !Xapi_globs.gen_pool_secret_script []
+    | `urandom ->
+        (* uuids require 16 bytes, and the pool secret is
+           made up of 3 of them, so we need 16 * 3 bytes *)
```
psafont, line 1823: Can the uuid module from transitional be used here? It already can generate uuids from urandom, we can always use its internals to use uuidm if need be.

ocaml/xapi/helpers.ml
```diff
@@ -1819,20 +1819,7 @@ end = struct
...
-        Unixext.with_file "/dev/urandom" [Unix.O_RDONLY] 0 (fun fd ->
-            let (_ : int) = Unix.read fd bs 0 (bytes_per_uuid * num_uuids) in
-            ()) ;
-        Stdlib.List.init num_uuids Stdlib.Fun.id
-        |> List.map (fun i ->
-               Bytes.sub bs (i * bytes_per_uuid) bytes_per_uuid
-               |> Uuidm.v4
-               |> Uuidm.to_string)
-        |> String.concat "/"
+        Uuid.(make_uuid_urnd () |> to_string)
```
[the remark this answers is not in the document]
lippirk: Thanks, you are right! The following is correct:

```
# Stdlib.List.init 3 (fun _ -> Uuid.(make_uuid_urnd () |> to_string)) |> String.concat "/";;
- : string =
"7f96f5dc-c52e-79de-093c-90f701c8fd47/c01e524a-68f6-aea0-3c1d-8c705ba1082a/419c4138-edb4-af51-2114-f168fd073603"
```

ocaml/xapi/xapi_globs.ml
```diff
@@ -1194,6 +1216,9 @@ module Resources = struct
     ; ( "generate_ssl_cert"
       , Constants.generate_ssl_cert
       , "script to generate SSL certificates to be used by XAPI" )
+    ; ( "gen_pool_secret_script"
+      , gen_pool_secret_script
+      , "Generates new pool secrets" )
```
psafont, line 1221: Even if urandom is used, the script will need to be in the filesystem. Not sure if it should be required. Is there another solution for this?
In any case @stormi could work around this by placing an empty file.

lippirk: As discussed we'll default to using /dev/urandom, and only use the script if the file /usr/bin/pool_secret_wrapper exists

psafont (approved):

ocaml/xapi/helpers.ml
```diff
@@ -1804,5 +1782,70 @@ end = struct
...
+
+  let _make () =
+    (* by default we generate the pool secret using /dev/urandom,
+       but if a script to generate the pool secret exists, use
+       that instead *)
+    try
+      Unix.access !Xapi_globs.gen_pool_secret_script [Unix.X_OK] ;
+      call_script !Xapi_globs.gen_pool_secret_script []
+    with _ ->
+      Stdlib.List.init 3 (fun _ -> Uuid.(make_uuid_urnd () |> to_string))
```
lindig, line 1825: I think this should be logged. Someone could sabotage the script - it looks it is used but in fact is not. If the script exists but raises an error, that would be concerning - so I think this should be handled with more nuance.

lindig (changes requested):

ocaml/xapi/helpers.ml
lippirk: That makes sense. I don't think we should log the exception that occurs however, because we want to avoid any chance of logging the secret

ocaml/xapi/helpers.ml
```diff
@@ -1804,5 +1782,83 @@ end = struct
...
+    (* by default we generate the pool secret using /dev/urandom,
+       but if a script to generate the pool secret exists, use that instead *)
+    let make_urandom () =
+      Stdlib.List.init 3 (fun _ -> Uuid.(make_uuid_urnd () |> to_string))
+      |> String.concat "/"
+    in
+    let make_script () =
+      try call_script !Xapi_globs.gen_pool_secret_script []
+      with _ ->
+        info "helpers.ml:PoolSecret._make: script failed, using urandom instead" ;
```
lippirk, line 1827: I've updated this so that if the script fails, we log and revert to /dev/urandom

lindig (approved):

PR #4198, opened 2020-09-07
CA-332605 - Fixed Bad error message for vcpu/cores-per-socket

GeorgeBurac:
Creating a Windows 8.1 VM in XenCenter by allocating 1 vCPU and
topology 1 core-per-socket would lead to an incorrectly formatted
error (i.e. "Failed","Starting VM 'Windows 8.1 (64-bit) (1)'
The value 'VCPUs_max must be a multiple of this field' is invalid
for field 'platform:cores-per-socket'.","xrtuk-01-01","Dec 6, 2019 12:55 PM")

The incorrectly formatted error made it more difficult to understand the issue
causing it.

### DIFF
--- a/ocaml/idl/datamodel_errors.ml
+++ b/ocaml/idl/datamodel_errors.ml
@@ -411,6 +411,8 @@ let _ =
     ~doc:"An error occured while restoring the memory image of the specified virtual machine" ();
   error Api_errors.vm_pv_drivers_in_use [ "vm" ]
     ~doc:"VM PV drivers still in use" ();
+  error Api_errors.vcpu_max_not_cores_per_socket_multiple ["vcpu_max"; "cores_per_socket"]
+    ~doc:"VCPUs_max must be a multiple of cores-per-socket" ();
   (* VM appliance errors *)
   error Api_errors.operation_partially_failed [ "operation" ]
     ~doc:"Some VMs belonging to the appliance threw an exception while carrying out the specified operation" ();
--- a/ocaml/xapi/vm_platform.ml
+++ b/ocaml/xapi/vm_platform.ml
@@ -196,21 +196,17 @@ let sanity_check ~platformdata ?firmware ~vcpu_max ~vcpu_at_startup ~domain_type
         if cps < 1 || vcpus mod cps <> 0 then
           raise
             (Api_errors.Server_error
-               ( Api_errors.invalid_value
+               ( Api_errors.vcpu_max_not_cores_per_socket_multiple
                , [
-                   Printf.sprintf "platform:cores-per-socket (value %d)" cps
-                 ; Printf.sprintf
-                     "VCPUs_max (value %d) must be a multiple of \
-                      cores-per-socket"
-                     vcpus
-                 ] ))
+                   string_of_int vcpus; cps_str
+                 ]
+ ))
       with Failure msg ->
         raise
           (Api_errors.Server_error
              ( Api_errors.invalid_value
              , [
-                 Printf.sprintf "platform:cores-per-socket (value %s)" cps_str
-               ; "Value is not a valid int"
+               "platform:cores-per-socket"; cps_str
                ] ))
   ) ;
   (* Add usb emulation flags.
--- a/ocaml/tests/test_platformdata.ml
+++ b/ocaml/tests/test_platformdata.ml
@@ -152,28 +152,28 @@ module SanityCheck = Generic.MakeStateless (struct
         ( ([("cores-per-socket", "4")], None, false, 6L, 6L, `hvm)
         , Error
             (Api_errors.Server_error
-               ( Api_errors.invalid_value
+               ( Api_errors.vcpu_max_not_cores_per_socket_multiple
                , [
-                   "platform:cores-per-socket (value 4)"
-                 ; "VCPUs_max (value 6) must be a multiple of cores-per-socket"
+                     "6"
+                   ; "4"
                  ] )) )
       ; (* Check VCPUs configuration - hvm failure scenario*)
         ( ([("cores-per-socket", "0")], None, false, 6L, 6L, `hvm)
         , Error
             (Api_errors.Server_error
-               ( Api_errors.invalid_value
+               ( Api_errors.vcpu_max_not_cores_per_socket_multiple
                , [
-                   "platform:cores-per-socket (value 0)"
-                 ; "VCPUs_max (value 6) must be a multiple of cores-per-socket"
+                   "6"
+                   ; "0"
                  ] )) )
       ; (* Check VCPUs configuration - hvm failure scenario*)
         ( ([("cores-per-socket", "-1")], None, false, 6L, 6L, `hvm)
         , Error
             (Api_errors.Server_error
-               ( Api_errors.invalid_value
+               ( Api_errors.vcpu_max_not_cores_per_socket_multiple
                , [
-                   "platform:cores-per-socket (value -1)"
-                 ; "VCPUs_max (value 6) must be a multiple of cores-per-socket"
+                   "6"
+                   ; "-1"
                  ] )) )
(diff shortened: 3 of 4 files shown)


ocaml/xapi/vm_platform.ml
```diff
@@ -198,19 +198,16 @@ let sanity_check ~platformdata ?firmware ~vcpu_max ~vcpu_at_startup ~domain_type
             (Api_errors.Server_error
                ( Api_errors.invalid_value
                , [
-                   Printf.sprintf "platform:cores-per-socket (value %d)" cps
-                 ; Printf.sprintf
-                     "VCPUs_max (value %d) must be a multiple of \
-                      cores-per-socket"
-                     vcpus
+                   Printf.sprintf "platform:cores-per-socket"
```
edwintorok, line 201: Might be useful to still log the more detailed reason why the failure occured.
(Or introduce a new kind of API error that can give more details about why a value is wrong)

psafont: There's a similar case with certificates and validity dates:
There are 2 API errors, `server_certificate_not_valid_yet` and `server_certificate_expired` each one has two parameters. The erroneous the refused value, and the the closest valid value. These can be used to construct message that can be easily understood.
You can see the errors defined in
https://github.com/xapi-project/xen-api/blob/99d98bdd3c00e7f5053940d3d3430289c01ecee3/ocaml/idl/datamodel_errors.ml#L1051-L1054

GeorgeBurac: Thank you so much for that. :slightly_smiling_face:

ocaml/idl/datamodel_errors.ml
```diff
@@ -411,6 +411,8 @@ let _ =
     ~doc:"An error occured while restoring the memory image of the specified virtual machine" ();
   error Api_errors.vm_pv_drivers_in_use [ "vm" ]
     ~doc:"VM PV drivers still in use" ();
+  error Api_errors.invalid_VCPU_max ["field"; "value"]
```
lindig, line 414: The field seem redundant because always the same field is used but it does no harm.

lindig (approved):

ocaml/idl/datamodel_errors.ml
psafont: I think replacing the field with `cores-per-socket` would allow the UI to inform the user what's wrong and how to fix it:
```
error Api_errors.vcpu_max_not_cores_per_socket_multiple ["vcpu_max"; "cores_per_socket"]
```
which the can be used as
```
Api_errors.(Server_error vcpu_max_not_cores_per_socket_multiple [vcpus; cpds])
```

robhoes: @psafont's suggestion is better.

GeorgeBurac: Thank you so much. I will update it now. :)

ocaml/xapi/vm_platform.ml
```diff
@@ -196,21 +196,18 @@ let sanity_check ~platformdata ?firmware ~vcpu_max ~vcpu_at_startup ~domain_type
...
             (Api_errors.Server_error
-               ( Api_errors.invalid_value
+               ( Api_errors.vcpu_max_not_cores_per_socket_multiple
                , [
-                   Printf.sprintf "platform:cores-per-socket (value %d)" cps
-                 ; Printf.sprintf
-                     "VCPUs_max (value %d) must be a multiple of \
-                      cores-per-socket"
-                     vcpus
+                   Printf.sprintf "platform:cores-per-socket"
```
robhoes, line 201: You now need to fill in `vcpus` here.

robhoes:
The parameters of the new raised errors don't match the definition now. See Pau's comment for how it should look.

ocaml/xapi/vm_platform.ml
```diff
@@ -196,21 +196,18 @@ let sanity_check ~platformdata ?firmware ~vcpu_max ~vcpu_at_startup ~domain_type
...
-                     vcpus
+                 ; Printf.sprintf "%d" vcpus
                  ] ))
       with Failure msg ->
         raise
           (Api_errors.Server_error
-             ( Api_errors.invalid_value
+             ( Api_errors.vcpu_max_not_cores_per_socket_multiple
              , [
                  Printf.sprintf "platform:cores-per-socket (value %s)" cps_str
```
psafont, line 209: This should be `Printf.sprintf "%s" cps_str`

ocaml/xapi/vm_platform.ml
```diff
@@ -196,21 +196,18 @@ let sanity_check ~platformdata ?firmware ~vcpu_max ~vcpu_at_startup ~domain_type
         if cps < 1 || vcpus mod cps <> 0 then
           raise
             (Api_errors.Server_error
-               ( Api_errors.invalid_value
+               ( Api_errors.vcpu_max_not_cores_per_socket_multiple
```
robhoes, line 199: It needs to be:
```
Api_errors.vcpu_max_not_cores_per_socket_multiple, [string_of_int vcpus; cps_str]
```

ocaml/xapi/vm_platform.ml
```diff
@@ -196,21 +196,18 @@ let sanity_check ~platformdata ?firmware ~vcpu_max ~vcpu_at_startup ~domain_type
...
-                 ; Printf.sprintf
-                     "VCPUs_max (value %d) must be a multiple of \
-                      cores-per-socket"
-                     vcpus
+                 ; Printf.sprintf "%d" vcpus
                  ] ))
       with Failure msg ->
         raise
           (Api_errors.Server_error
-             ( Api_errors.invalid_value
```
psafont, line 210: This should still be `Api_errors.invalid_value`: it's a string that cannot be transformed into a string. i don't think `vcpus` should be displayed at all

robhoes (changes requested):
See my and Pau's comments above.

ocaml/xapi/vm_platform.ml
GeorgeBurac: Thank you! :)

ocaml/xapi/vm_platform.ml
```diff
@@ -196,21 +196,17 @@ let sanity_check ~platformdata ?firmware ~vcpu_max ~vcpu_at_startup ~domain_type
...
+                 ]
+ ))
       with Failure msg ->
         raise
           (Api_errors.Server_error
              ( Api_errors.invalid_value
              , [
-                 Printf.sprintf "platform:cores-per-socket (value %s)" cps_str
-               ; "Value is not a valid int"
+                 Printf.sprintf "platform:cores-per-socket (value %s)" cps_str;
```
psafont, line 209: I would leave it as it was, it was explaining the root of the problem :)

robhoes: The error is defined as:
```
  error Api_errors.invalid_value ["field"; "value"]
    ~doc:"The value given is invalid" ();
```
Therefore, it needs two arguments: the field name and the input value. No other text, just plain data.

So:
```
[ "platform:cores-per-socket"; cps_str]
```

ocaml/xapi/vm_platform.ml
```diff
@@ -196,21 +196,17 @@ let sanity_check ~platformdata ?firmware ~vcpu_max ~vcpu_at_startup ~domain_type
...
+                 ]
+ ))
       with Failure msg ->
         raise
           (Api_errors.Server_error
              ( Api_errors.invalid_value
              , [
-                 Printf.sprintf "platform:cores-per-socket (value %s)" cps_str
-               ; "Value is not a valid int"
+               "platform:cores-per-scoket"; cps_str
```
robhoes, line 209: Spelling mistake...

psafont (approved):

robhoes (approved):

PR #4199, opened 2020-09-07
XSI-795 CA-343951 fix Nvidia version parsing

lindig:
The version of the host driver is read from the library file name and
used to decide if multi-vgpu is supported. The version used to 123.56
but now contains another dot and thus caused the parsing to fail. We now
split the string at the dots and compare them piece wise as numbers (not
as strings). This means 1 = 1.0 = 1.0.0 and so on and "" = 0.

The format of the string is not completely validated. We split the
string and pass the elements to int_of_string to extract the version
number.

Xapi's configuration marks a version as "123.4+" to indicate the all
versions 123.4 and later are supported. We find such a version but again
do not validate the format completely.

I have changed the tests slightly: if there are any format problems detected 
in any of the version strings, the feature is not supported. It was more dynamic 
earlier but that is more difficult to keep and can lead to nasty surprises when
you update a driver.


ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -421,19 +421,52 @@ module Vendor_nvidia = struct
...
-          try float_of_string (String.sub pattern 0 i) <= version
-          with _ -> false
-        )
-        | None -> (
-          try float_of_string pattern = version with _ -> false
-        ))
-      supported_versions
+    try
+      let latest =
+        List.find_opt (fun v -> String.contains v '*') supported_versions
```
robhoes, line 461: It used to be `+` rather than `*`?

lindig: Yeah - and the unit tests found it, too. Fixed now. I have tested this otherwise manually.

lindig: There is another test failure - have to review it

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -421,19 +421,58 @@ module Vendor_nvidia = struct
...
-          with _ -> false
-        )
-        | None -> (
-          try float_of_string pattern = version with _ -> false
-        ))
-      supported_versions
+    try
+      let driver = nvidia_version host_driver_version in
+      let supported = List.map nvidia_version supported_versions in
+      let latest =
```
lippirk, line 463: I'm a bit confused - do you mean `earliest` here? My understanding is that the earliest version that supports `multi_vgpu` will be marked with a `+`, e.g. 1.2 onwards, or 1.2+

ocaml/tests/test_vgpu_type.ml
```diff
@@ -174,10 +174,10 @@ module NvidiaTest = struct
         ; (* Test specific version should match *)
           (("430.19", ["430.89"; "430.19"]), true)
         ; (* Test specific version should match *)
-          (("430.19", ["430.19"; "typo_error"]), true)
-        ; (* Test specific version should match *)
```
psafont, line 178: Comments are confusing here, they're attached to the wrong element.

psafont (approved):

lippirk (approved):

ocaml/xapi/xapi_vgpu_type.ml
lindig: Yeah, maybe not the best name. This whole idea is a bit odd and I only kept it for compatibility with what we have. For example: `[1,3+,2+]` is semantically `[1,2+]` - it does not make sense to have more than one entry with `+` but I'm only looking for the first one.

ocaml/tests/test_vgpu_type.ml
lippirk: There doesn't seem to be much benefit having these comments in the first place

PR #4200, opened 2020-09-07
CA-344268: Fix timing issue in PBIS available check

lippirk:
If the lwsmd service has only just started when xapi checks to see
whether PBIS / AD is available, then removing the KRBTGT user from
the PBIS cache fails. This is not a problem, but we should try again
to remove KRBTGT from the cache rather than erroring out completely.

To remedy this, we move the call to `ad-cache --delete-user` inside
the retry loop. Additionally, the retry loop has been decoupled from the
checking

This was tested by running a thread which restarts lwsmd periodically,
and then restarting xapi:

```
$ restart_lwsmd () { while true; do systemctl restart lwsmd; sleep 3; done; }
$ restart_lwsmd &
$ systemctl restart xapi
```

Unfortunately this only catches the issue some of the time; after
restarting xapi potentially several times, you should see the following
in xensource.log:

Auth_signature.Auth_failure("The authentication request could not be handled")

### DIFF
--- a/ocaml/xapi/extauth_plugin_ADpbis.ml
+++ b/ocaml/xapi/extauth_plugin_ADpbis.ml
@@ -734,83 +734,94 @@ module AuthADlw : Auth_signature.AUTH_MODULE = struct
     In addition, there are some event hooks that auth modules implement as follows:
 *)
 
-  let _is_pbis_server_available max =
-    let username = "KRBTGT" in
-    (* domain name prefix automatically added by our internal AD plugin functions *)
-    let rec test i =
-      (* let's test this many times *)
-      if i > max then
-        false (* we give up *)
-      else
-        (* let's test *)
-
-        (* (1) we _need_ to use a username contained in our domain, otherwise test (2) doesn't work *)
-        (* Microsoft KB/Q243330 article provides the KRBTGT account as a well-known built-in SID in AD *)
-        (* Microsoft KB/Q229909 article says that KRBTGT account cannot be renamed or enabled, making *)
-        (* it the perfect target for such a test using a username (Administrator account can be renamed). *)
-        try
-          let username = "KRBTGT" in
-          (* domain name prefix automatically added by our internal AD plugin functions *)
-          let sid = get_subject_identifier username in
-          (* use our well-known KRBTGT builtin username in AD *)
-          (* OK, we found this username! *)
-          debug
-            "Request %i/%i to external authentication server successful: user \
-             %s was found"
-            i max username ;
-          (* (2) CA-25427: test (1) above may succeed (because of pbis caching stale AD information) *)
-          (* even though the AD domain is offline (error 32888), usually because /etc/resolv.conf is not *)
-          (* pointing to the AD server. This test should catch if the domain is offline by calling find-by-sid *)
-          (* using a domain SID. We must use a _domain_ SID. A universal SID like S-1-1-0 doesn't work for this test. *)
-          let (_ : (string * string) list) = query_subject_information sid in
-          (* use KRBTGT's domain SID *)
-          debug
-            "Request %i/%i to external authentication server successful: sid \
-             %s was found"
-            i max sid ;
-          true
+  let _is_pbis_server_available max_tries =
+    (* we _need_ to use a username contained in our domain, otherwise the following tests won't work.
+       Microsoft KB/Q243330 article provides the KRBTGT account as a well-known built-in SID in AD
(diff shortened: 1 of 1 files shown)


ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -734,83 +734,94 @@ module AuthADlw : Auth_signature.AUTH_MODULE = struct
...
+  let _is_pbis_server_available max_tries =
+    (* we _need_ to use a username contained in our domain, otherwise the following tests won't work.
+       Microsoft KB/Q243330 article provides the KRBTGT account as a well-known built-in SID in AD
+       Microsoft KB/Q229909 article says that KRBTGT account cannot be renamed or enabled, making
+       it the perfect target for such a test using a username (Administrator account can be renamed) *)
+    let krbtgt = "KRBTGT" in
+    let try_clear_cache () =
+      match get_full_subject_name krbtgt with
+      | exception e ->
+          info "_is_pbis_server_available: failed to get full subject name" ;
```
lindig, line 746: I'd include the account you want to look up.

ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -734,83 +734,94 @@ module AuthADlw : Auth_signature.AUTH_MODULE = struct
...
-            test (i + 1)
-        | e ->
-            (* ERROR: anything else means that the server is NOT responding adequately *)
-            debug
-              "Request %i/%i to external authentication server failed, waiting \
-               5 secs to try again: %s"
-              i max
+        | () ->
+            true
+        | exception Not_found ->
```
lindig, line 756: Could you add a comment why this means the service is available?

ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -734,83 +734,94 @@ module AuthADlw : Auth_signature.AUTH_MODULE = struct
...
+            (ExnHelper.string_of_exn e) ;
+          false
+    in
+    (* in order to properly test that PBIS and the AD domain
+       are running correctly, we first remove KRBTGT from the
+       cache. if we didn't, then in the case that PBIS is running
+       and the AD domain is offline, [ can_fetch_sid ] will
+       erroneously return success *)
+    (* to avoid clearing the cache twice, we carry around a checkpoint *)
+    let rec pbis_functioning = function
```
lindig, line 800: Is this recursive function not just a sequence of two tests?

lindig (commented):
The existing code looks a bit convoluted and I wonder whether a monadic error handling would have been simpler.

ocaml/xapi/extauth_plugin_ADpbis.ml
lippirk: I've added a comment

The purpose of `try_clear_cache` is simply to remove the KRBTGT user from the cache to set up for the 'real test', `try_fetch_sid`. If the cache is populated, then `try_fetch_sid` will not perform an end to end query to the AD server

ocaml/xapi/extauth_plugin_ADpbis.ml
lippirk: Yes there are only two operations, the idea is to avoid calling `try_clear_cache` twice (though the code would still work if we did)

lippirk:
> The existing code looks a bit convoluted and I wonder whether a monadic error handling would have been simpler.

I'll see if using result or option is nicer

ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -734,83 +734,94 @@ module AuthADlw : Auth_signature.AUTH_MODULE = struct
...
+          (* that means that pbis is responding to at least cached subject queries.
+             in this case, KRBTGT wasn't found in the AD domain. this usually indicates that the
+             AD domain is offline/inaccessible to pbis, which will cause problems, specially
+             to the ssh python hook-script, so we need to try again until KRBTGT is found, indicating
+             that the domain is online and accessible to pbis queries *)
+          debug
+            "Request to external authentication server returned KRBTGT \
+             Not_found" ;
+          Error ()
+      | e ->
```
lindig, line 792: I think the error string could be returned by `R.error_msgf` and logged once at the end before the outermost function returns. But I'm fine with keeping this as it is. The disadvantage of that is that only one error could be logged.

lindig (approved):

robhoes (approved):
I'm not a domain expert, but this looks good as far as I can see.

PR #4203, opened 2020-09-10
CP-34936 don't log result from pool_secret_wrapper

lippirk:
We don't want to log the pool secret. So we add a new option to
`call_script` which ensures nothing is logged in either the success case
or the failure case.

### DIFF
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -1833,7 +1833,7 @@ end = struct
       |> String.concat "/"
     in
     let make_script () =
-      try call_script !Xapi_globs.gen_pool_secret_script []
+      try call_script ~log_output:Never !Xapi_globs.gen_pool_secret_script []
       with _ ->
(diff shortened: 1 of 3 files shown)


ocaml/xapi/helpers.ml
```diff
@@ -39,15 +39,21 @@ let log_exn_continue msg f x =
   with e ->
     debug "Ignoring exception: %s while %s" (ExnHelper.string_of_exn e) msg
 
-let call_script ?(log_successful_output = true) ?env script args =
+let call_script ?(log_successful_output = true)
```
lindig, line 42: Is this calling for a type to capture this and the default `SuccessOnly`? I'd call the parameter `logging` that would control this.

```
type log = Always | Never | SuccessOnly | ErrorOnly
```

lippirk: I thought there would be a lot more usages of `call_script`, but it wasn't that bad to change the interface

lindig (approved):

psafont (approved):

PR #4204, opened 2020-09-11
CP-33823 - replace script with OCaml for self signing SSL certs

lindig:
I prematurely pushed this to feature/REQ-403. To facilitate review, here is a PR for master. Overall, this replaces a bash script with OCaml code to create SSL certs but still uses openssl to create private keys.


ocaml/gencert/selfcert.ml
```diff
@@ -0,0 +1,125 @@
...
+module is loaded. *)
+let () = Mirage_crypto_rng_unix.initialize ()
+
+(** [write_cert] writes a PKCS12 file to [path]. The typical file
+ extension would be ".pem". It attempts to do that atomically by
+ writing to a temporary file in the same directory first and renaming
+ the file at the end *)
+let write_certs path pkcs12 =
+  let f () =
+    UX.atomic_write_to_file path 0o400 @@ fun fd ->
```
lippirk, line 32: I expected `0o600` to be necessary here? (`600=rw- --- ---`, and `400=r-- --- ---`)

psafont: This allows to move the certificate file, but not overwrite it with an empty file, add certificates the end, etc.

It's a small barrier to protect against accidental writes more than anything.

ocaml/gencert/selfcert.ml
```diff
@@ -0,0 +1,125 @@
...
+  | None ->
+      R.error_msgf "can't represent %d as time span" days
+
+let add_dns_names extension = function
+  | [] ->
+      extension
+  | names ->
+      X509.Extension.(
+        add Subject_alt_name
+          (false, X509.General_name.(singleton DNS names))
```
edwintorok, line 52: The `false` here indicates that the extension is not critical (i.e. can be ignored if the other side doesn't support it).
I think that is fine, they can fall back to the CN.

ocaml/xapi/xapi_host.ml
```diff
@@ -1400,10 +1399,16 @@ let install_server_certificate ~__context ~host ~certificate ~private_key
...
+  let common_name, alt_names =
+    match Gencertlib.Lib.hostnames () with
+    | cn :: alt ->
+        (cn, alt)
+    | [] ->
+        (Helper_hostname.get_hostname (), [])
+    (* should never happen *)
   in
-  ignore @@ Gencert.call_generate_ssl_cert ~args ;
+  Gencertlib.Selfcert.host common_name alt_names xapi_ssl_pem ;
```
edwintorok, line 1411: I think we always want to include the main name in the subject alt names.
The CN field is deprecated in X509 certs, the subject-alt-names is supposed to be used instead, so this should always contain the full list of hostnames

edwintorok: also https://tools.ietf.org/html/rfc6125#section-6.4.4

ocaml/gencert/selfcert.ml
lindig: Yes, this is not well documented - I had some comments on the earlier PR and looked at the X509 source code.

ocaml/xapi/xapi_host.ml
lindig: Ok - so don't split the names - use the full list in alt_names?

psafont: I agree with adding all hostnames into the SANs.

I'm not opposed to removing the CN from the certificate: it would not affect RFC-compliant clients, but I think it's worth communicating to teams building xapi clients they will not be able to rely on its presence anymore. (not that there were any)

lindig: I've added a commit that makes sure this is happening.

ocaml/gencert/selfcert.ml
```diff
@@ -122,4 +122,7 @@ let selfsign name alt_names length days certfile =
 let host name alt_names pemfile =
   let expire_days = 3650 in
   let length = 2048 in
-  selfsign name alt_names length expire_days pemfile |> R.failwith_error_msg
+  (* make sure name is part of alt_names because CN is deprecated and
+     that there are no duplicates *)
+  let alt = Astring.String.uniquify (name :: alt_names) in
```
psafont, line 127: Didn't know the Astring offered [this](https://github.com/dbuenzli/astring/blob/master/src/astring_string.ml#L748), the implementation is sure better than [stdext's](https://github.com/xapi-project/stdext/blob/master/lib/xapi-stdext-std/listext.ml#L18)

psafont (approved):

edwintorok (approved):

PR #4205, opened 2020-09-14
Update feature/REQ-403 branch

lindig:
This aligns REQ-403 (certificate checking) with master. There should be no difference between the two because we merged the self-signing changes into master already.


edwintorok (approved):

psafont (approved):

psafont:
The only change is the addition of all the hostnames as SANs into the self-signed certificates, which was the last change introduced.

PR #4209, opened 2020-09-14
feature/REQ-819/master -> master

lippirk:


lindig:
I assume this code has been reviewed before as part of the feature work - so will just give it a glance here.

lindig (approved):

lippirk:
> I assume this code has been reviewed before as part of the feature work - so will just give it a glance here.

Thanks! Yes, it's been reviewed already, so in theory all we need to check is 89fa576 and 9015232 (where conflicts were resolved + formatting)

psafont (approved):
The changes I've overviewed were all reasonable, without unexpected changes, same with the commits, they seem to be the same as the ones that were originally reviewed for the feature branch.

PR #4211, opened 2020-09-16
CA-265116 rename and deprecate Pool cert functions

lindig:
The names of some pool certificate functions are misleading. We
deprecate them and introduce the same functionality under better
names.

* certificate_install -> install_ca_certificate
* certificate_uninstall -> uninstall_ca_certificate

Update the CLI documentation for functions that handle Certificate
Revocation Lists.

### DIFF
--- a/ocaml/idl/datamodel_pool.ml
+++ b/ocaml/idl/datamodel_pool.ml
@@ -411,7 +411,21 @@ open Datamodel_types
       ~params:[String, "name", "A name to give the certificate";
                String, "cert", "The certificate"]
       ~allowed_roles:_R_POOL_OP
+      ~internal_deprecated_since:rel_next
       ()
+
+
+  let install_ca_certificate = call
+      ~in_oss_since:None
+      ~in_product_since:rel_next
+      ~name:"install_ca_certificate"
+      ~doc:"Install a TLS CA certificate, pool-wide."
+      ~params:[String, "name", "A name to give the certificate";
+               String, "cert", "The certificate"]
+      ~allowed_roles:_R_POOL_OP
+      ~internal_deprecated_since:rel_next
+      ()
+
 
   let certificate_uninstall = call
       ~in_oss_since:None
@@ -420,6 +434,17 @@ open Datamodel_types
       ~doc:"Remove a pool-wide TLS CA certificate."
       ~params:[String, "name", "The certificate name"]
       ~allowed_roles:_R_POOL_OP
+      ~internal_deprecated_since:rel_next
+      ()
+
+  let uninstall_ca_certificate = call
+      ~in_oss_since:None
+      ~in_product_since:rel_george
+      ~name:"uninstall_ca_certificate"
+      ~doc:"Remove a pool-wide TLS CA certificate."
+      ~params:[String, "name", "The certificate name"]
+      ~allowed_roles:_R_POOL_OP
+      ~internal_deprecated_since:rel_next
       ()
 
   let certificate_list = call
@@ -429,6 +454,7 @@ open Datamodel_types
       ~doc:"List the names of all installed TLS CA certificates."
       ~result:(Set(String),"All installed certificates")
       ~allowed_roles:_R_POOL_OP
+      ~internal_deprecated_since:rel_next
       ()
 
   let crl_install = call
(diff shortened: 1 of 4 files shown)


ocaml/idl/datamodel_pool.ml
```diff
@@ -411,15 +411,40 @@ open Datamodel_types
...
       ()
 
+
+  let install_ca_certificate = call
+      ~in_oss_since:None
+      ~in_product_since:rel_next
+      ~name:"install_ca_certificate"
+      ~doc:"Install a TLS CA certificate, pool-wide."
+      ~params:[String, "name", "A name to give the certificate";
+               String, "cert", "The certificate"]
```
psafont, line 424: I'd like this to mention that it needs to be in PEM format

ocaml/idl/datamodel_pool.ml
```diff
@@ -411,15 +411,40 @@ open Datamodel_types
...
+      ~allowed_roles:_R_POOL_OP
+      ~internal_deprecated_since:rel_next
```
psafont, line 426: I don't think this is already deprecated :)

psafont:
I've just seen that the corresponding calls for host should also be replaced: https://github.com/xapi-project/xen-api/blob/master/ocaml/idl/datamodel_host.ml#L875-L910

I'm also not sure we should keep the same signature. The name of the file the certificate is saved to should not matter to users. Could we merge this into the feature branch instead of master? It's bound to change as part of the work to store CA certificates into the database.

ocaml/idl/datamodel_pool.ml
lindig: Oops - must have attached it to the wrong call.

lindig:
This will go into a feature branch

ocaml/idl/datamodel_pool.ml
```diff
@@ -409,17 +409,40 @@ open Datamodel_types
       ~name:"certificate_install"
       ~doc:"Install a TLS CA certificate, pool-wide."
       ~params:[String, "name", "A name to give the certificate";
-               String, "cert", "The certificate"]
+               String, "cert", "The certificate in PEM format"]
       ~allowed_roles:_R_POOL_OP
+      ~internal_deprecated_since:rel_next
```
robhoes, line 414: Please use `~lifecycle` to replace `~in_oss_since`, `~in_product_since` and `~internal_deprecated_since`. This allows you to provide information about the deprecation, e.g. that the function has been replaced by another one.

ocaml/idl/datamodel_pool.ml
```diff
@@ -409,17 +409,40 @@ open Datamodel_types
...
       ()
 
+
+  let install_ca_certificate = call
+      ~in_oss_since:None
+      ~in_product_since:rel_next
```
robhoes, line 420: These two can also be replaced by a single `~lifecycle`.

ocaml/idl/datamodel_pool.ml
```diff
@@ -409,17 +409,40 @@ open Datamodel_types
...
       ~name:"certificate_uninstall"
       ~doc:"Remove a pool-wide TLS CA certificate."
       ~params:[String, "name", "The certificate name"]
       ~allowed_roles:_R_POOL_OP
+      ~internal_deprecated_since:rel_next
+      ()
+
+  let uninstall_ca_certificate = call
+      ~in_oss_since:None
+      ~in_product_since:rel_george
```
robhoes, line 441: Same comment as above. In additional, this should say `rel_next`.

ocaml/xapi-cli-server/cli_frontend.ml
```diff
@@ -346,9 +346,25 @@ let rec cmdtable_data : (string * cmd_spec) list =
...
+      ; help= "Uninstall a pool-wide TLS CA certificate."
+      ; implementation= No_fd Cli_operations.pool_certificate_uninstall
+      ; flags= [Deprecated ["Use pool-uninstall-ca-certificate"]]
+      } )
+  ; ( "pool-install-ca-certificate"
+    , {
+        reqd= ["filename"]
+      ; optn= []
+      ; help= "Install a TLS CA certificate, pool-wide."
+      ; implementation= With_fd Cli_operations.pool_certificate_install
```
robhoes, line 364: I think you should rename the functions in `Cli_operations` as well. You could link both old and new to functions with the new name.

PR #4212, opened 2020-09-16
CA-333441 - restarting ISCSI daemon after setting initiator IQN

GeorgeBurac:
Setting the initiator IQN after attempting to connect
to and iSCSI target using an older IQN causes the daemon
to not be updated with the new IQN.

The daemon needs to be restarted in order to update
the IQN value used.

### DIFF
--- a/ocaml/xapi/xapi_host_helpers.ml
+++ b/ocaml/xapi/xapi_host_helpers.ml
@@ -392,7 +392,7 @@ module Host_requires_reboot = struct
 end
 
 module Configuration = struct
-  let make_initiatorname_config iqn hostname =
+  let make_set_initiator_args iqn hostname =
     (* CA-18000: there is a 30 character limit to the initiator when talking to
        Dell MD3000i filers, so we limit the size of the initiator name in all cases *)
     let hostname_chopped =
@@ -401,14 +401,12 @@ module Configuration = struct
       else
         hostname
     in
-    Printf.sprintf "InitiatorName=%s\nInitiatorAlias=%s\n" iqn hostname_chopped
+    [iqn; hostname_chopped]
 
   let set_initiator_name iqn =
     let hostname = Unix.gethostname () in
-    let config_file = make_initiatorname_config iqn hostname in
-    Unixext.write_string_to_file
-      !Xapi_globs.iscsi_initiator_config_file
-      config_file
+    let args = make_set_initiator_args iqn hostname in
+    ignore(Helpers.call_script !Xapi_globs.set_iSCSI_initiator_script args)
 
   let set_multipathing enabled =
     let flag = !Xapi_globs.multipathing_config_file in
--- a/ocaml/xapi/xapi_globs.ml
+++ b/ocaml/xapi/xapi_globs.ml
@@ -382,6 +382,9 @@ let startup_script_hook = ref "xapi-startup-script"
 (* Executed when a rolling upgrade is detected starting or stopping *)
 let rolling_upgrade_script_hook = ref "xapi-rolling-upgrade"
 
+(* Sets IQN and restarts iSCSI daemon if required *)
+let set_iSCSI_initiator_script =  ref "/opt/xensource/libexec/set-iscsi-initiator"
+
 (* Executed during startup when the host is authed with AD
  * or the host is joining or leaving AD *)
 let domain_join_cli_cmd = ref "/opt/pbis/bin/domainjoin-cli"
@@ -1200,6 +1203,8 @@ module Resources = struct
     ; ( "openssl_path"
       , Constants.openssl_path
       , "Path for openssl command to generate RSA keys" )
+    ; ( "set-iscsi-initiator", set_iSCSI_initiator_script
+      , "Path to set-iscsi-initiator script")
     ]
 
   let nonessential_executables =


ocaml/xapi/xapi_host_helpers.ml
```diff
@@ -401,14 +401,15 @@ module Configuration = struct
...
   let set_initiator_name iqn =
     let hostname = Unix.gethostname () in
     let config_file = make_initiatorname_config iqn hostname in
-    Unixext.write_string_to_file
-      !Xapi_globs.iscsi_initiator_config_file
-      config_file
+    let shell_params = String.concat " " config_file in 
+    let shell_script = "sh /opt/xensource/libexec/set-iscsi-initiator" in 
+    let shell_command = String.concat " " [shell_script; shell_params] in
+    Sys.command shell_command
```
edwintorok, line 412: We usually use forkhelpers to run commands.

ocaml/xapi/xapi_host.ml
```diff
@@ -2370,7 +2370,7 @@ let set_iscsi_iqn ~__context ~host ~value =
    * *)
   Db.Host.set_iscsi_iqn ~__context ~self:host ~value ;
   Db.Host.add_to_other_config ~__context ~self:host ~key:"iscsi_iqn" ~value ;
-  Xapi_host_helpers.Configuration.set_initiator_name value
+  ignore (Xapi_host_helpers.Configuration.set_initiator_name value)
```
edwintorok, line 2373: you return an int, but ignore it. I think instead any errors should be handled inside the function (e.g. non-zero exitcode would mean failure). See the point about using forkhelpers though.

ocaml/xapi/xapi_host_helpers.ml
edwintorok: The name of the script should also be part of Xapi_globs, and part of the required scripts list.

ocaml/xapi/xapi_host.ml
GeorgeBurac: Thank you for that, Edvin. :)

ocaml/xapi/xapi_host_helpers.ml
GeorgeBurac: Thank you! I will update it.

robhoes: We have `Helpers.call_script`.

ocaml/xapi/xapi_host.ml
GeorgeBurac: Now this returns a string and error handling appears to be done by the call_script helper function. Are there other changes required here?

robhoes: Now you don't need to ignore here anymore, since the function returns `()` before.

ocaml/xapi/xapi_host_helpers.ml
```diff
@@ -401,14 +401,13 @@ module Configuration = struct
...
+    [iqn; hostname_chopped]
 
   let set_initiator_name iqn =
     let hostname = Unix.gethostname () in
     let config_file = make_initiatorname_config iqn hostname in
-    Unixext.write_string_to_file
-      !Xapi_globs.iscsi_initiator_config_file
-      config_file
+    ignore(Helpers.call_script ~log_output:On_failure 
+    !Xapi_globs.set_iSCSI_initiator_script config_file)
```
robhoes, line 410: Indentation looks wrong.

ocaml/xapi/xapi_host_helpers.ml
```diff
@@ -401,14 +401,13 @@ module Configuration = struct
       else
         hostname
     in
-    Printf.sprintf "InitiatorName=%s\nInitiatorAlias=%s\n" iqn hostname_chopped
+    [iqn; hostname_chopped]
 
   let set_initiator_name iqn =
     let hostname = Unix.gethostname () in
     let config_file = make_initiatorname_config iqn hostname in
```
robhoes, line 408: The names `config_file` and `make_initiatorname_config` don't make sense anymore, which is confusing. I suggest `let args = make_set_initiator_args ...`.

ocaml/xapi/xapi_host_helpers.ml
```diff
@@ -401,14 +401,13 @@ module Configuration = struct
...
-    Unixext.write_string_to_file
-      !Xapi_globs.iscsi_initiator_config_file
-      config_file
+    ignore(Helpers.call_script ~log_output:On_failure
```
robhoes, line 409: I would leave the default for `log_output`, so just omit the param.

ocaml/xapi/xapi_host_helpers.ml
GeorgeBurac: Thank you! :)

ocaml/xapi/xapi_host.ml
robhoes: You've resolved the conversation, but the `ignore` is still there.

ocaml/xapi/xapi_host_helpers.ml
```diff
@@ -401,14 +401,12 @@ module Configuration = struct
...
+    [iqn; hostname_chopped]
 
   let set_initiator_name iqn =
     let hostname = Unix.gethostname () in
-    let config_file = make_initiatorname_config iqn hostname in
-    Unixext.write_string_to_file
-      !Xapi_globs.iscsi_initiator_config_file
-      config_file
+    let args = make_set_initiator_args iqn hostname in
+      ignore(Helpers.call_script !Xapi_globs.set_iSCSI_initiator_script args)
```
robhoes, line 409: Indentation is still off, now that it's gone to a single line.

ocaml/xapi/xapi_host.ml
GeorgeBurac: Apologies, I thought that I fixed it.

robhoes (approved):

edwintorok (approved):
This works correctly I think

PR #4214, opened 2020-09-17
CA-256116 rename pool and host functions that handle CA certs

lindig:
Make it more obvious that pool and host functions are handling CA-issued certificates by renaming them while keeping the old name around but deprecated.

### DIFF
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -873,8 +873,6 @@ let host_query_ha = call ~flags:[`Session]
       ()
 
   let certificate_install = call
-      ~in_oss_since:None
-      ~in_product_since:rel_george
       ~pool_internal:true
       ~hide_from_docs:true
       ~name:"certificate_install"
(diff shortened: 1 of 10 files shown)


ocaml/xapi/xapi_pool.ml
```diff
@@ -2130,8 +2130,12 @@ let send_test_post = Remote_requests.send_test_post
 
 let certificate_install = Certificates.(pool_install CA_Certificate)
```
psafont, line 2131: Since they are not API-facing function, I'd like the previous functions to be removed and their usages to be changed to the new ones, same with `xapi_host`

psafont: These cannot be removed  because of how the API definitions and these module interact.

psafont (approved):

lindig:
The CA ticket number in the commit messages had digits swapped. I fixed that with a force push.

ocaml/idl/datamodel_host.ml
```diff
@@ -915,7 +948,7 @@ let host_query_ha = call ~flags:[`Session]
       ~pool_internal:true
       ~hide_from_docs:true
       ~name:"crl_install"
-      ~doc:"Install a TLS Certificate Revocation List to this host."
+      ~doc:"Install a TLS CA-issued Certificate Revocation List to this host."
```
lippirk, line 951: might be overkill, but do we need to add a `Changed` to the lifecycle when we are modifying docs?

lindig: I think this would add noise - because as a user you would expect a different behaviour when there is none.

lippirk (approved):

ocaml/idl/datamodel_host.ml
psafont: I don't think so, we've modified docs before and the lifecycle was not changed.

robhoes: `Changed` is for interface changes that are in some way not backwards compatible. See https://xapi-project.github.io/xen-api/evolution.html.

So: no.

PR #4216, opened 2020-09-22
CP-34465: Add signing fields to certificate records

psafont:
These allow to model certificate chains and let xapi manage which
certificates are to be installed as pool-wide CA certificates, which are
to be installed in a single host, or both.

This change does not ensure that the fields are correctly filled and are up-to-date with the certificates in the filesystem.
To do that several codepaths will need to be modified: Pool CA certificate installation and uninstallation; host certificate reset, host certificate installation; and startup to syncronize CA certificates and server certificate.

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -5598,6 +5598,8 @@ let all_relations =
     (_network_sriov, "logical_PIF"), (_pif, "sriov_logical_PIF_of");
 
     (_certificate, "host"), (_host, "certificates");
+    (_certificate, "pool"), (_pool, "certificates");
+    (_certificate, "validated_by"), (_certificate, "validates");
   ]
 
 (** the full api specified here *)
--- a/ocaml/idl/datamodel_certificate.ml
+++ b/ocaml/idl/datamodel_certificate.ml
@@ -18,6 +18,13 @@ open Datamodel_roles
 
 let lifecycle = [Published, rel_stockholm, ""]
 
+let certificate_type =
+  Enum ( "certificate_type"
+       , [ "host", "Certificate that identifies a single host"
+         ; "pool", "Certificate that is trusted by the whole pool"
+         ; "host_and_pool", "Certificate that both identifies a host and is trusted by the whole pool."
+         ])
+
 let t =
   create_obj
     ~name: _certificate
@@ -32,9 +39,15 @@ let t =
     ~messages_default_allowed_roles:_R_READ_ONLY
     ~contents:
       [ uid   _certificate ~lifecycle
-      ; field   ~qualifier:StaticRO ~lifecycle
+      ; field ~qualifier:DynamicRO ~lifecycle:[Published, rel_next, ""]
+          ~ty:certificate_type "type"
+          "The type of certificate, can be a host certificate, a CA certificate, or both"
+      ; field ~qualifier:StaticRO ~lifecycle
           ~ty:(Ref _host) "host" ~default_value:(Some (VRef null_ref))
-          "The host where the certificate is installed"
+          "The host that is identified by the certificate"
+      ; field ~qualifier:StaticRO ~lifecycle:[Published, rel_next, ""]
+          ~ty:(Ref _pool) "pool" ~default_value:(Some (VRef null_ref))
+          "The pool that trusts this certificate"
       ; field ~qualifier:StaticRO ~lifecycle
           ~ty:(DateTime) "not_before" ~default_value:(Some (VDateTime Date.never))
           "Date after which the certificate is valid"
@@ -44,6 +57,12 @@ let t =
       ; field ~qualifier:StaticRO ~lifecycle
           ~ty:(String) "fingerprint" ~default_value:(Some (VString ""))
           "The certificate's fingerprint / hash"
+      ; field ~qualifier:DynamicRO ~lifecycle:[Published, rel_next, ""]
+      ~ty:(Set (Ref _certificate)) "validated_by"
+          "The trusted certificates that validate this host certificate"
+      ; field ~qualifier:DynamicRO ~lifecycle:[Published, rel_next, ""]
+          ~ty:(Set (Ref _certificate)) "validates"
+          "The certificates that are validates by this pool certificate"
       ]
     ~messages:
       [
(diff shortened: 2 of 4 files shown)


lindig (commented):
This looks good but I was confused by "signs" - because it could be a verb or a noun and I was reading it as a noun.

ocaml/idl/datamodel_certificate.ml
```diff
@@ -44,6 +57,12 @@ let t =
       ; field ~qualifier:StaticRO ~lifecycle
           ~ty:(String) "fingerprint" ~default_value:(Some (VString ""))
           "The certificate's fingerprint / hash"
+      ; field ~qualifier:StaticRO ~lifecycle:[Published, rel_next, ""]
+          ~ty:(Ref _certificate) "signed_by" ~default_value:(Some (VRef null_ref))
```
edwintorok, line 61: I think this is the part that you said that duplicates the information in the X509 chain.
Does signed_by also imply that the signature is a valid? What happens when the certificate expires, or is revoked by a CRL or OCSP responder? Perhaps we should treat it as signed_by=has a valid digital signature from the other certificate, and to leave expiration and revocation validation up to OpenSSL (which probably won't do revocation validation on its own though).

edwintorok: Also note that certificates are not uniquely identified in a chain, it is sometimes possible to construct different trust paths starting from same end user certificate, see https://medium.com/@sleevi_/path-building-vs-path-verifying-the-chain-of-pain-9fbab861d7d6, we should also consider the situation where a single certificate has multiple possible issuers, and multiple paths (i.e. instead of treating it as a chain, treat it as a graph).
Or perhaps we put the burden on the user to choose one valid trust path, and then we store just that in the DB?

ocaml/idl/datamodel.ml
```diff
@@ -5598,6 +5598,8 @@ let all_relations =
     (_network_sriov, "logical_PIF"), (_pif, "sriov_logical_PIF_of");
 
     (_certificate, "host"), (_host, "certificates");
+    (_certificate, "pool"), (_pool, "certificates");
+    (_certificate, "signed_by"), (_certificate, "signs");
```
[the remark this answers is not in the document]
psafont: Not sure how else to word the relation, certificates are always signed (or certified) by a certificate, which can be itself or a different one. The relation of certificates A and B when the certificate A contains the signature generated by certificate B is usually expressed as B signs A.

Examples from the RFC:
> Users of a public key require confidence that the associated private key is owned by the correct remote subject (person or system) with which an encryption or digital signature mechanism will be used. This confidence is obtained through the use of public key certificates, which are data structures that bind public key values to subjects. The binding is asserted by having a trusted CA digitally sign each certificate.

> The signatureAlgorithm field contains the identifier for the cryptographic algorithm used by the CA to sign this certificate.

ocaml/idl/datamodel_certificate.ml
psafont: > Perhaps we should treat it as signed_by=has a valid digital signature from the other certificate, and to leave expiration and revocation validation up to OpenSSL (which probably won't do revocation validation on its own though).

This is how it's intended to be used: there are already fields for validity dates and CRL are only used when users do an explicit API call, xapi does not use them on the background to achieve revocation. (instead certificates are remove from the trust root)

> we should also consider the situation where a single certificate has multiple possible issuers, and multiple paths

This is indeed an issue. Currently the pool level certificates are all trusted. Hosts may have intermediate certificates along the server certificate to have be able to validated using trusted certificates.

One of the ideas to model all the intermediate certificates was to alert users of about-to-expire certificates. I wanted to also model the difference between self-signed certificates and CA-issued ones, but this is already done by the field `_type`. I'm rethinking the need to model the paths within the database and how it relates to the needs for the feature.

ocaml/idl/datamodel.ml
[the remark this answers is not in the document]
psafont: These have been changed to a `validates` `validated_by` with only two types of certificates, trusted ones and server ones. Intermediate certificates are omitted in the database.

ocaml/idl/datamodel_certificate.ml
psafont: > we store a pool level certificate (which itself can chain up to a trusted root somewhere in our store), and then check that host level certificates are signed by that certificate, i.e. we only have 2 levels: the CA level and the issued host level, and no more levels?

This is the situation with the new proposal: `pool` certificates are trust anchors for all the hosts in the pool and can validate host certificates. `host` certificates are server certificate and identify a single host. They are validated by at least one pool certificate. Auto-generated certificates are `pool_and_host` because they are both trust anchors and identify a server.

lippirk (approved):

ocaml/idl/datamodel.ml
```diff
@@ -5598,6 +5598,8 @@ let all_relations =
     (_network_sriov, "logical_PIF"), (_pif, "sriov_logical_PIF_of");
 
     (_certificate, "host"), (_host, "certificates");
+    (_certificate, "pool"), (_pool, "certificates");
+    (_certificate, "validated_by"), (_certificate, "validates");
```
edwintorok, line 5602: I was worried that this would create a loop, but we already have similar situations with VM snapshots/snapshot_of, so this should be fine.

ocaml/idl/datamodel_certificate.ml
```diff
@@ -32,9 +39,15 @@ let t =
     ~messages_default_allowed_roles:_R_READ_ONLY
     ~contents:
       [ uid   _certificate ~lifecycle
-      ; field   ~qualifier:StaticRO ~lifecycle
+      ; field ~qualifier:DynamicRO ~lifecycle:[Published, rel_next, ""]
+          ~ty:certificate_type "type"
+          "The type of certificate, can be a host certificate, a CA certificate, or both"
```
edwintorok, line 44: Do we need to assign a default value or otherwise handle upgrades?

edwintorok (commented):
I think we need some code in the DB upgrade, so we c

ocaml/idl/datamodel_certificate.ml
psafont: Upgrades need to be handled, all existing certificates are have the type `` `host``.
Similarly, the validated_by need to be populated.

robhoes: If you synchronise all `Certificate` objects for a host in the dbsync phase of xapi's startup sequence, then you don't need special "upgrade" code.

ocaml/idl/datamodel.ml
psafont: In the documentation it's specified that setting the `validated_by` field will automatically update the corresponding `validates` field in the referenced object with the correct reference.

ocaml/idl/datamodel_certificate.ml
psafont: I've dropped the `db_upgrade` code, the new fields will be updated in `update_env` from `dbsync_master.ml` instead. I'd like to do that in a separate PR and merge this to build an SDk with these changes.

edwintorok: I assume the FQP will contain some test that checks that upgrades work when you got these certificates installed (wouldn't rely on RPU tests catching this, they don't use the full features of the product).

edwintorok (approved):

PR #4217, opened 2020-09-22
Update definitions of HTTP actions

robhoes:


robhoes:
@kc284 This has some impact on the SDK:
* Some handler definitions are deleted.
* Additional parameters are added for SDK-exposed HTTP actions. Note that the added arguments are optional, but there is currently no way to express this. I also believe that the C# code generator doesn't use optional arguments, so these changes may impact the users of these bindings. Is that what we want?

lippirk:
Can we also add the necessary files to be able to generate your HTTP endpoints document? (with `make doc` perhaps)

robhoes:
> Can we also add the necessary files to be able to generate your HTTP endpoints document?

Yep, but I'd do that separately.

lippirk (approved):

kc284:
The SDK code could be modified to generate optional arguments for the http actions, but if there is no way to tell them apart from the mandatory parameters, we'll resort to adding manually the old version of these methods as overloads.

robhoes:
We can of course change the type in the datamodel to express more details about the args, such as optionality and possible values. Does C# have optional function args with defaults? Or do you need overloads?

We also have some that accept `uuid` as well as `ref`, but only one is required. I didn't even add those yet to avoid confusion, but I could add the additional one as optional.

kc284:
Recent versions of C# understand optional arguments, but the SDK generator was written before that time, hence it generates method overloads instead, it will need some work to change it. However, this does not prevent you from defining the Ocaml parameters in any way you want. The option to use the opaque_ref instead of the uuid would be useful for the consumers of the SDK. Note though that there may be some calls where the parameter is called "uuid", when in reality this is an opaque_ref; it would be nice if the definition could hold additionally parameter decsription so it can be output in the generated code.

kc284 (approved):

lindig (approved):
Has this been ocamlformat'ed?

robhoes:
OK thanks, let's put this in as it is now. Then I'll do a proposal for an extended type later.

robhoes:
> Has this been ocamlformat'ed?

`ocaml/idl/datamodel*` has been excluded from the formatter.

PR #4218, opened 2020-09-22
CP-34472 login fail alerts

lippirk:
We store information about failed login attempts in memory. This information is made persistent every 60 minutes, by generating a report and storing it in an alert. Each time the report is made, we forget completely about previous failed attempts - in this way, reports only contain information about the last 60 minutes.

### DIFF
--- a/ocaml/idl/datamodel_errors.ml
+++ b/ocaml/idl/datamodel_errors.ml
@@ -115,6 +115,8 @@ let _ =
   (* Session errors *)
   error Api_errors.session_authentication_failed []
     ~doc:"The credentials given by the user are incorrect, so access has been denied, and you have not been issued a session handle." ();
+  error Api_errors.session_authorization_failed ["username"; "msg"]
+    ~doc:"The credentials given by the user are correct, but the user could not be authorized, so access has been denied, and you have not been issued a session handle." ();
   error Api_errors.session_invalid ["handle"]
     ~doc:"You gave an invalid session reference. It may have been invalidated by a server restart, or timed out. You should get a new session handle, using one of the session.login_ calls. This error does not invalidate the current connection. The handle parameter echoes the bad value given." ();
   error Api_errors.change_password_rejected [ "msg" ]
--- a/ocaml/tests/test_session.ml
+++ b/ocaml/tests/test_session.ml
@@ -1,0 +1,168 @@
+open Test_common
+module Date = Xapi_stdext_date.Date
+
+let now = Date.of_string "2020-09-22T14:57:11Z"
+
+let future = Date.of_string "2020-09-22T15:03:13Z"
+
+let fail_login ~__context ~uname ~originator ~now () =
+  try
+    Xapi_session._record_login_failure ~__context ~now ~uname ~originator
+      ~record:`log_and_alert (fun () ->
+        if Random.bool () then
+          raise Api_errors.(Server_error (session_authentication_failed, []))
+        else
+          raise (Auth_signature.Auth_failure "Auth failure"))
+  with e -> ()
+
+let success_login ~__context ~uname ~originator ~now () =
+  Xapi_session._record_login_failure ~__context ~now ~uname ~originator
+    ~record:`log_and_alert Fun.id
+
+let make_ctx ~user_agent =
+  let open Context in
+  match user_agent with
+  | None ->
+      make "test_ctx"
+  | Some _ ->
+      let rq = {Http.Request.empty with user_agent} in
+      (* it doesn't matter which fd is used to here, we are just satisying the
+         type system. we use stderr because then we don't need to worry about
+         closing it *)
+      make ~origin:(Http (rq, Unix.stderr)) "text_ctx"
+
+let repeat n f =
+  for _ = 1 to n do
+    f ()
+  done
+
+let run_unknown_client_logins () =
+  let __context = make_ctx ~user_agent:None in
+  repeat 50
+    (success_login ~__context ~uname:(Some "good_user")
+       ~originator:(Some "nice_origin") ~now) ;
+  repeat 10 (fail_login ~__context ~uname:None ~originator:None ~now) ;
+  repeat 50
+    (success_login ~__context ~uname:(Some "good_user")
+       ~originator:(Some "nice_origin") ~now)
+
+let run_known_client_logins () =
+  let __context = make_ctx ~user_agent:(Some "UA") in
+  let __context_no_UA = make_ctx ~user_agent:None in
+  repeat 50
+    (success_login ~__context ~uname:(Some "good_user")
+       ~originator:(Some "nice_origin") ~now) ;
+  repeat 2
+    (fail_login ~__context:__context_no_UA ~uname:(Some "usr1") ~now
+       ~originator:(Some "origin1")) ;
+  repeat 3
+    (fail_login ~__context:__context_no_UA ~uname:None
+       ~originator:(Some "origin2") ~now) ;
+  repeat 4 (fail_login ~__context ~uname:None ~originator:None ~now) ;
+  repeat 6
+    (fail_login ~__context ~uname:(Some "usr4") ~originator:None ~now:future) ;
+  let () =
+    (* this client fails now and then in the future (to test timestamp) *)
+    repeat 9
+      (fail_login ~__context:__context_no_UA ~uname:(Some "usr5")
+         ~originator:(Some "origin5") ~now) ;
+    repeat 1
+      (fail_login ~__context:__context_no_UA ~uname:(Some "usr5")
+         ~originator:(Some "origin5") ~now:future)
+  in
+  repeat 50
+    (success_login ~__context ~uname:(Some "good_user")
+       ~originator:(Some "nice_origin") ~now:future)
+
+let test_fetching_failed_login_stats_twice_yields_none () =
+  let _ = Xapi_session.get_failed_login_stats () in
+  let stats = Xapi_session.get_failed_login_stats () in
+  Alcotest.(check @@ option string)
+    "no extra stats have been accumulated" stats None
+
+let test_only_failed_logins_from_unknown_clients () =
+  let _ = Xapi_session.get_failed_login_stats () in
+  run_unknown_client_logins () ;
+  let stats = Xapi_session.get_failed_login_stats () |> Option.get in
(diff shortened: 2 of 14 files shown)


ocaml/tests/test_session.ml
```diff
@@ -0,0 +1,136 @@
...
+  | None ->
+      make "test_ctx"
+  | Some _ ->
+      let rq = {Http.Request.empty with user_agent} in
+      (* it doesn't matter which fd is used to here, we are just satisying the
+         type system. we use stderr because then we don't need to worry about
+         closing it *)
+      make ~origin:(Http (rq, Unix.stderr)) "text_ctx"
+
+let repeat n f =
```
lindig, line 30:
```
let repeat n f = for _ = 1 to n do f (); done
```

ocaml/xapi/xapi_session.ml
```diff
@@ -27,6 +27,228 @@ open Client
...
+
+  module TopThree = struct
+    let maybe_drop len = function
+      | [] ->
+          []
+      | _ :: rest as all ->
+          if len >= 3 then rest else all
+
+    (* util to maintain a list of up to 3 offending addresses - least bad to most bad *)
+    let maybe_add xs x =
```
lindig, line 132: Not that it matters too much but using `List.length` and `List.sort` for this reeks of inefficiency. I believe this should be a linear insertion into an ordered list of pairs of length 3.

ocaml/xapi/xapi_session.ml
```diff
@@ -27,6 +27,228 @@ open Client
...
+    type key = client
+
+    type value = {num_failed_attempts: int; last_failed_attempt: Date.iso8601}
+
+    type table_t = (key, value) Hashtbl.t
+
+    let table = Hashtbl.create 10
+
+    let record_client k ~now =
+      Mutex.execute m (fun () ->
```
lindig, line 178: It looks like we are keeping a counter for all clients. Given how rarely the top3 are reported, is it not enough to query the hash table and sort the results to obtain the 3 top values?

psafont: I would try to obtain the last 3 values when it needs to be calculated instead of all the time.

ocaml/xapi/xapi_session.ml
```diff
@@ -27,6 +27,228 @@ open Client
...
+    with e ->
+      (* we don't expect this function to fail, but if it does we don't want to block callers *)
+      error "AuthFail.on_fail_with_uname: unexpected error: '%s'"
+        (Printexc.to_string e)
+end
+
+let _record_login_failure ~__context ~now ~uname ~originator ~record f =
+  try f ()
+  with e ->
+    (* naively assume any exception corresponds to an authentication failure *)
```
psafont, line 242: How reliable is this? Don't we have specific exception to signal authentication failure?

ocaml/xapi/xapi_session.ml
```diff
@@ -27,6 +27,228 @@ open Client
...
+                  num_known_client_failed_attempts
+                  + unknown_client_failed_attempts
+              ; top_3_worst_clients= List.rev top_3_worst_clients
+              ; unknown_client_failed_attempts
+              })
+  end
+
+  let get_stats_string () = Stats.get () |> Option.map string_of_stats
+
+  let on_fail ~__context ~now ~uname ~originator ~record =
```
psafont, line 222: `~record` doesn't seem to be used here

ocaml/xapi/xapi_session.ml
```diff
@@ -27,6 +27,228 @@ open Client
...
+             checking, including %i of unknown origin! The worst known \
+             offenders were:\n\
+             %s"
+            total_num_failed_attempts unknown_client_failed_attempts worst
+        else
+          Printf.sprintf
+            "We found a total of %i failed attempts to login since last \
+             checking! The worst offenders were:\n\
+             %s"
+            total_num_failed_attempts worst
```
psafont, line 122: For certificate alerts we used xml to structure messages, if I remember correctly to parse dates independently:

```
 Printf.sprintf "<body><message>%s</message><date>%s</date></body>" msg (Date.to_string expiry)
```

@kc284 What format to display these alerts would be better suited for xencenter? Would xml help you to display the dates in here, while the rest of the message is more freeform?

psafont:
The changes are looking good. Just the small niggle that Christian mentioned and the comment about the output format for alerts.

Some observations: I would have liked not to add the code to the xen-api binary but we don't have much of a choice here since it's xapi that keeps track of these. The failed logins could have been reported as operational metrics, we don't have a metrics infrastructure able to cope with these, however.

ocaml/xapi/xapi_session.ml
kc284: Yes, presumably these numbers need to be shown by the UI, hence isolating them in a separate field would be much easier to parse, otherwise XenCenter has to use a text based regex which will break if in future the wording on the message changes. Json would also work, if you want smaller data. E.g.:
```
<body>
    <offence>
        <message>Total failed attempts to login since last checking</message>
        <number>%i</number>
    </offence>
    <offence>
        <message>Attempts of unknown origin:</message>
        <number>%i</number>
    </offence>
    <offence>
        <message>Worst known offenders</message>
        <number>%s</number>
    </offence>
</body>
```

kc284: Or maybe use different tag names instead of "offence" so you're not not order bound, e.g. "total", "unknown", "offenders".

ocaml/xapi/xapi_session.ml
lippirk: I found a better way of doing this

ocaml/xapi/xapi_session.ml
lippirk: > How reliable is this?
It depends how you define reliable

false positive = FP = we logged a failed login request as an auth failure when it wasn't really an auth failure
false negative = FN = an login failed because of invalid credentials and we didn't record it

The current implementation has the potential to cause FPs, but will never result in a FN. If we catch only on `Auth_signature.Auth_failure` and the code behaves exactly as we expect, then we'll never get FNs or FPs :+1: However my confidence of that isn't particularly high after looking at `xapi_session.ml:login_with_password`

In practice it probably doesn't matter which method we choose

psafont: Could the code in the logins be changed to use `Auth_failure` when appropiate? I would prefer to minimize false positives to have minimum noise if there are no auth failures. Customers might get worried at first and call support,or if it becomes too noisy they will tend to ignore this kind of alerts.

lippirk: To be clear, you are talking about looking at the implementations of the login functions (e.g. `login_with_password`), and making sure they from `Auth_failure` when we expect? I will look into this

psafont: That's correct. Makes those function throw `Auth_failure` so we can rely on it for logging / alerting.

lippirk:
I've modified the stats report to be xml based now. When writing the XenCenter parser we need to remember that there might be an client IP field (one doesn't exist at the moment because we don't have that information yet inside xapi).

I've combed through `login_with_password`. I believe it will now throw errors with code `SESSION_AUTHENTICATION_FAILED` exactly when there is an authentication failure (previously it was defaulting to throwing this error, even if the user had successfully authenticated)

ocaml/xapi/xapi_session.ml
```diff
@@ -27,6 +27,232 @@ open Client
...
+          let unknown_client_failed_attempts = !unknown_ctr in
+          if Hashtbl.length table = 0 && unknown_client_failed_attempts = 0 then
+            None
+          else
+            let num_known_client_failed_attempts, top_3_worst_clients =
+              Hashtbl.fold
+                (fun client {num_failed_attempts; last_failed_attempt}
+                     (ctr, worst_so_far) ->
+                  ( ctr + num_failed_attempts
+                  , up_to_3 worst_so_far
```
psafont, line 200: This looks lineal to me, and only called when the values need to be collected for the alerts, which seems like a good compromise to me.

ocaml/xapi/xapi_session.ml
```diff
@@ -608,7 +609,8 @@ let login_with_password ~__context ~uname ~pwd ~version ~originator =
                       uname subject_identifier
                       (Context.get_origin __context)
                       msg ;
-                    thread_delay_and_raise_error uname msg
+                    thread_delay_and_raise_error
+                      ~error:Api_errors.session_invalid uname msg
```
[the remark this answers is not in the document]
lippirk: I've replaced the session_invalid instances I added with session_authorization_failed (there was one instance where it wasn't appropriate - when the auth service is unavailable - I changed that to internal_error)

ocaml/idl/datamodel_errors.ml
```diff
@@ -115,6 +115,8 @@ let _ =
   (* Session errors *)
   error Api_errors.session_authentication_failed []
     ~doc:"The credentials given by the user are incorrect, so access has been denied, and you have not been issued a session handle." ();
+  error Api_errors.session_authorization_failed ["username"; "msg"]
+    ~doc:"User was successfully authenticated, but could not be authorized" ();
```
psafont, line 119: I would try to follow the message structure and vocabulary of the previous error, similar to
"The credentials given by the user do not give them enough permissions, so access has been denied."

ocaml/xapi/authx.ml
```diff
@@ -134,7 +134,11 @@ module AuthX : Auth_signature.AUTH_MODULE = struct
         (* no exception raised, then authentication succeeded *)
       with Failure msg -> raise (Auth_signature.Auth_failure msg)
     in
-    get_subject_identifier username
+    try get_subject_identifier username
+    with Not_found ->
+      raise
+        (Auth_signature.Auth_failure
+           (Printf.sprintf "Could not find user's SID (uname='%s')" username))
```
psafont, line 141: The actual reason is that neither the user's id nor the group id could be found.

psafont (approved):

lindig:
I feel the record keeping and reporting is a bit more complicated than required. It mixes an imperative hash table with a functional style. Using either a `Map` or a record with mutable counter and date could have been simpler. Anyway, this is well encapsulated.

lindig (approved):

PR #4223, opened 2020-10-06
CA-333441 - restarting ISCSI daemon after setting initiator IQN

robhoes:
(Re-adding patch b4b46341 which was reverted in 569c7e1c)

Setting the initiator IQN after attempting to connect
to and iSCSI target using an older IQN causes the daemon
to not be updated with the new IQN.

The daemon needs to be restarted in order to update
the IQN value used.


robhoes:
The original patch was reverted only due to missing dependences, which we have now resolved.

lindig (approved):

psafont (approved):

PR #4224, opened 2020-10-09
maintenance: allow tests to run in a sandbox

psafont:
Since `/var/tmp` is not writable when inside opam's sandbox, the code for upgrading the database is changed to read the inventory filepath from `/tmp` when running the tests. A environment variable is used, in a similar way to https://github.com/xapi-project/forkexecd/pull/44

### DIFF
--- a/ocaml/tests/test_xapi_db_upgrade.ml
+++ b/ocaml/tests/test_xapi_db_upgrade.ml
@@ -58,10 +58,9 @@ let upgrade_vm_memory_for_dmc () =
     (r.API.vM_memory_static_min <= r.API.vM_memory_static_max)
 
 let upgrade_bios () =
+  let tmp_filename = "/tmp/previousInventory" in
   let check inventory bios_strings =
-    Xapi_stdext_unix.Unixext.mkdir_safe "/var/tmp" 0o755 ;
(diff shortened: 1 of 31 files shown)


ocaml/tests/test_xapi_db_upgrade.ml
```diff
@@ -58,10 +58,9 @@ let upgrade_vm_memory_for_dmc () =
     (r.API.vM_memory_static_min <= r.API.vM_memory_static_max)
 
 let upgrade_bios () =
+  let tmp_filename = "/tmp/.previousInventory" in
```
lindig, line 61: Using fixed names in a global directory risks conflicts. Should we not generate a file name?

psafont: The issue is that both the writer (test) and the reader (db_upgrade) have to use the same filename. I've tried to inject the value of the generated filename to the db upgrade function using the current interface to no avail. I wanted to avoid being too disruptive around that code.

lindig (approved):

ocaml/tests/test_xapi_db_upgrade.ml
lindig: Maybe not use a hidden file? We would never look for it.

lippirk (approved):

PR #4225, opened 2020-10-09
CA-333441: Do not fail the startup sequence if the iSCSI initiator cannot be set

edwintorok:
With HA on or during a toolstack restart we do expect other iSCSI
sessions to be active, so this cannot be changed.
Log and ignore instead of blocking XAPI startup.


robhoes (approved):

lindig:
This needs a rebase because we reverted the patch that introduced the problem.

robhoes (approved):

psafont (approved):
The cherry-picked commit fails the DCO check, this is fine.

PR #4226, opened 2020-10-09
Revert "CA-333441 - restarting ISCSI daemon after setting initiator IQN"

lindig:
Reverting this because we suspect interaction with the High Availability
code and see a failing test case.

This reverts commit b4b46341cd20d456529d62b483ef5e9359407330.


psafont (approved):

edwintorok (approved):
We can test the fix here afterwards: https://github.com/xapi-project/xen-api/pull/4225

PR #4229, opened 2020-10-12
REQ-403 CP-34468 add Host.reset_server_certificate

lindig:
Deprecate Host.emergency_reset_server_certificate and add
Host.reset_server_certificate. The new method is synonym for the old one
but can be called by the pool admin.

### DIFF
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -1011,9 +1011,18 @@ let host_query_ha = call ~flags:[`Session]
       ~allowed_roles:_R_POOL_ADMIN
       ()
 
+  let reset_server_certificate = call
+      ~flags:[`Session]
+      ~lifecycle:[Published, rel_next, ""]
+      ~name:"reset_server_certificate"
+      ~doc:"Delete the current TLS server certificate and replace by a new, self-signed one. This should only be used with extreme care."
+      ~params:[Ref _host, "host", "The host"]
+      ~allowed_roles:_R_POOL_ADMIN
+      ()
+
   let emergency_reset_server_certificate = call
       ~flags:[`Session]
-      ~lifecycle:[Published, rel_stockholm, ""]
+      ~lifecycle:[ Published, rel_stockholm, ""]
       ~name:"emergency_reset_server_certificate"
       ~doc:"Delete the current TLS server certificate and replace by a new, self-signed one. This should only be used with extreme care."
       ~versioned_params: []
@@ -1514,6 +1523,7 @@ let host_query_ha = call ~flags:[`Session]
         get_server_certificate;
         install_server_certificate;
         emergency_reset_server_certificate;
+        reset_server_certificate;
         update_pool_secret;
         update_master;
         attach_static_vdis;
--- a/ocaml/xapi-cli-server/cli_frontend.ml
+++ b/ocaml/xapi-cli-server/cli_frontend.ml
@@ -695,6 +695,17 @@ let rec cmdtable_data : (string * cmd_spec) list =
             Cli_operations.host_emergency_reset_server_certificate
       ; flags= [Neverforward]
       } )
+  ; ( "host-reset-server-certificate"
+    , {
+        reqd= []
+      ; optn= []
+      ; flags= [Host_selectors]
+      ; help=
+          "Deletes the current TLS server certificate in the host and installs \
+           a new, self-signed one."
+      ; implementation=
+          No_fd_local_session Cli_operations.host_reset_server_certificate
+      } )
   ; ( "host-management-reconfigure"
     , {
         reqd= ["pif-uuid"]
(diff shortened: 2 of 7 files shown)


ocaml/xapi/message_forwarding.ml
```diff
@@ -3230,9 +3230,11 @@ functor
                     failed."
                  ] ))
 
-      let emergency_reset_server_certificate ~__context =
+      let reset_server_certificate ~__context =
         Local.Host.emergency_reset_server_certificate ~__context
```
psafont, line 3234: This call needs to be routed to the appropriate member.

ocaml/idl/datamodel_host.ml
```diff
@@ -1011,9 +1011,19 @@ let host_query_ha = call ~flags:[`Session]
       ~allowed_roles:_R_POOL_ADMIN
       ()
 
+  let reset_server_certificate = call
+      ~flags:[`Session]
+      ~lifecycle:[Published, rel_next, ""]
+      ~name:"reset_server_certificate"
+      ~doc:"Delete the current TLS server certificate and replace by a new, self-signed one. This should only be used with extreme care."
+      ~versioned_params: []
```
psafont, line 1019: The call is missing a `host` parameter with type `Ref _host`. Before since the call was only local this parameter was implicit.

psafont (changes requested):

lindig:
Sorry - this is quite incomplete. It also needs CLI support.

psafont:
I didn't think much of it, it's a draft after all ;)

lindig:
I updated the PR. I am not sure about the forwarding mechanism - this needs always be executed on the host that that is given on the CLI/call and is this what message forwarding does?

psafont:
All xmlrpc calls are taken by the master, which forwards the call to the correct host, or executes the local function. This is done with the `do_op_on` function, just like you've done.

ocaml/xapi/xapi_host.ml
```diff
@@ -1403,7 +1403,12 @@ let install_server_certificate ~__context ~host ~certificate ~private_key
   (* sever all connections done with stunnel *)
   Xapi_mgmt_iface.reconfigure_stunnel ~__context
 
-let emergency_reset_server_certificate ~__context =
+let reset_server_certificate ~__context ~host =
+  let self = Helpers.get_localhost ~__context in
+  ( if host <> self then
+      let msg = "reset_server_certificate host != localhost" in
+      raise Api_errors.(Server_error (internal_error, [msg]))
+  ) ;
```
psafont, line 1411: I don't think this check belongs to this layer, it should be contained to the `message_forwarding` module.

ocaml/xapi-cli-server/cli_frontend.ml
```diff
@@ -695,6 +695,17 @@ let rec cmdtable_data : (string * cmd_spec) list =
             Cli_operations.host_emergency_reset_server_certificate
       ; flags= [Neverforward]
       } )
+  ; ( "host-reset-server-certificate"
+    , {
+        reqd= ["host-uuid"]
```
robhoes, line 700: Better use `flags= [Host_selectors]` rather than adding an explicit `host-uuid` param. See other commands for examples.

ocaml/idl/datamodel_host.ml
```diff
@@ -1011,9 +1011,19 @@ let host_query_ha = call ~flags:[`Session]
...
+      ~doc:"Delete the current TLS server certificate and replace by a new, self-signed one. This should only be used with extreme care."
+      ~params:[Ref _host, "host", "The host"]
+      ~allowed_roles:_R_POOL_ADMIN
+      ()
+
   let emergency_reset_server_certificate = call
       ~flags:[`Session]
-      ~lifecycle:[Published, rel_stockholm, ""]
+      ~lifecycle:[ Published, rel_stockholm, ""
+                 ; Deprecated, rel_next, "Use Host.reset_server_certificate"]
```
robhoes, line 1026: It may be worth keeping this as an official API call, as it can be used on a pool member with a broken master connection (I think).

ocaml/xapi/xapi_host.ml
robhoes: This assertion does no harm, but we can rely on the forwarding layer to have done the right thing.

robhoes (approved):
Looks good to me now.

ocaml/xapi/xapi_host.ml
```diff
@@ -1403,7 +1403,8 @@ let install_server_certificate ~__context ~host ~certificate ~private_key
   (* sever all connections done with stunnel *)
   Xapi_mgmt_iface.reconfigure_stunnel ~__context
 
-let emergency_reset_server_certificate ~__context =
+let reset_server_certificate ~__context ~host =
```
psafont, line 1406: `~host` is unused, but needs to be defined because of how the definition machinery

psafont (approved):

PR #4234, opened 2020-10-21
CP-35026 log API client's IP addresses

lippirk:
This includes plumbing the client IP address from `STUNNEL_PROXY` (https://github.com/xapi-project/xen-api-libs-transitional/pull/84) into values of type `Context.t`.

The client IP is given to the Debug module (https://github.com/xapi-project/xcp-idl/pull/317/files) when we begin processing an API request (`exec_with_context`). The Debug module handles actually logging the info.

Here is an example of what it looks like:

```
[ info||293 10.71.8.14->INET :::80|session.login_with_password D:0488288ebf28|xapi_session]
```


robhoes:
It would be nice to also indicate if the incoming connection came through stunnel or not, since we have all the information now. Perhaps add `HTTP` or `HTTPS`? E.g.

    [ info||293 HTTPS 10.71.8.14->INET :::80|session.login_with_password D:0488288ebf28|xapi_session]

The `INET`/`UNIX` string actually feels a bit redundant, because it is clear from the bit after it which one it is: an IP/port such as `:::80` or a path such as `/var/lib/xcp/xapi`. Perhaps we could drop that (in `Xapi_http.bind`), so the log becomes simply:

    [ info||293 HTTPS 10.71.8.14->:::80|session.login_with_password D:0488288ebf28|xapi_session]

lippirk:
> It would be nice to also indicate if the incoming connection came through stunnel or not, since we have all the information now. Perhaps add `HTTP` or `HTTPS`? E.g.
> 
> ```
> [ info||293 HTTPS 10.71.8.14->INET :::80|session.login_with_password D:0488288ebf28|xapi_session]
> ```
> 
> The `INET`/`UNIX` string actually feels a bit redundant, because it is clear from the bit after it which one it is: an IP/port such as `:::80` or a path such as `/var/lib/xcp/xapi`. Perhaps we could drop that (in `Xapi_http.bind`), so the log becomes simply:
> 
> ```
> [ info||293 HTTPS 10.71.8.14->:::80|session.login_with_password D:0488288ebf28|xapi_session]
> ```

Thanks, these are good suggestions! It's now looking like this:

```
[debug||262 HTTPS 10.62.49.51->:::80|session.slave_local_login D:73559a489c73|xapi_session]
[ info||182 HTTP 127.0.0.1->:::80|session.logout D:7972001e5cfa|xapi_session]
```

and in other cases where we don't know the client, or the client is internal:

```
[debug||274 /var/lib/xcp/xapi||dummytaskhelper]
[error||290 :::80||backtrace]
[ info||182 :::80||taskhelper]
```

We could actually add more information in the backtrace case - that's a change for xcp-idl anyhow

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -84,32 +84,43 @@ end = struct
...
+                   port localhost)
+        in
+        (Xapi_http.bind sockaddr, desc addr sockaddr)
+      in
       match addr with
       | None -> (
           info "Starting new server (listening on all IP addresses)" ;
           try
-            (* Is it IPv6 ? *)
+            (* this may succeed even if IPv6 is not used, but we must try [inet6_addr_any]
```
robhoes, line 115: This was meant to catch the case where ipv6 is disabled in the kernel. The call to bind would then fail.

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -84,32 +84,43 @@ end = struct
      _the_ management interface. Hosts in a pool use the IP addresses of this interface
      to communicate with each other. *)
   let start ~__context ?addr () =
+    let port = !Constants.https_port in
     let socket, stunnel_accept =
```
robhoes, line 88: What is confusing in the original code, and still in the new code, is how the xapi listening socket and stunnel's accept parameter are derived together. In the new function, it still look a little odd that the bind function produces the stunnel param as well.

The localhost aspect there comes from the fact that the case where a server is started on a specific IP coincides with the case where external management is to be disabled.

ocaml/xapi/context.ml
```diff
@@ -152,6 +152,46 @@ let trackid ?(with_brackets = false) ?(prefix = "") __context =
...
+let _clean_addr_of_string ip =
+  (* in the IPv4 case, users should see 127.0.0.1 rather than ::ffff:127.0.0.1 *)
+  let ipv4_affix = "::ffff:" in
+  ( if Astring.String.is_prefix ~affix:ipv4_affix ip then
+      Astring.String.drop ~max:(String.length ipv4_affix) ip
+  else
+    ip
+  )
+  |> Ipaddr.of_string
+  |> Rresult.R.to_option
```
psafont, line 170:
```suggestion
  |> Result.to_option
```

robhoes:
The tests show a problem:
```
File "ocaml/xapi/server_helpers.ml", line 157, characters 9-19:
157 |       |> marshaller
               ^^^^^^^^^^
Error: This expression has type (?client:string -> 'a) -> Rpc.t
       but an expression was expected of type 'a -> 'b
       The type variable 'a occurs inside ?client:string -> 'a
make: *** [Makefile:12: build] Error 1
```

ocaml/xapi/xapi_mgmt_iface.ml
lippirk: I see. My hope was that `domain_of_socket` would tell us that `Unix.inet6_addr_any` actually returned a ipv4 address, but it doesn't. I'll revert back to the old behaviour (though the observable behaviour is coincidentally equivalent)

lippirk:
> The tests show a problem:
> 
> ```
> File "ocaml/xapi/server_helpers.ml", line 157, characters 9-19:
> 157 |       |> marshaller
>                ^^^^^^^^^^
> Error: This expression has type (?client:string -> 'a) -> Rpc.t
>        but an expression was expected of type 'a -> 'b
>        The type variable 'a occurs inside ?client:string -> 'a
> make: *** [Makefile:12: build] Error 1
> ```

I think github-ci is not using the newest version of xcp-idl. I guess this is because of the caching, so this test should pass tomorrow

psafont:
We can try the same strategy around unversioned packages as in xs-opam: remove before caching. This means it will take more time per run, though, it will be interesting to see how much.

ocaml/xapi/xapi_mgmt_iface.ml
robhoes: `Unix.inet6_addr_any` is just a constant.

lippirk: Ah this makes sense now... :facepalm: 

I tried decoupling `stunnel_accept` from the creating of the `socket`, but I don't think it makes the code any better, so I've just reverted this file

psafont (approved):
I  believe this will be really helpful going forward 🎉

robhoes (approved):
This is great and will be a big help when triaging issues.

olivierlambert:
Great feature! Just wanted to say thanks for this :+1:

Is there any discussion to think about expose that somehow in the API or would it be preferable to create a separate plugin to parse the log? Leveraging this would allow to increase overall security (eg we have an UI in XO to expose our own audit log. Combining it with host audit log could be great addition to compare both and find discrepancies or attacks)

psafont:
The logging system needs a revamp. We currently use [custom libraries](https://github.com/xapi-project/xcp-idl/blob/master/lib/debug.ml) instead of using the community standard [logs](https://opam.ocaml.org/packages/logs/) library.

Remote syslog servers have been talked about, it would need changing the libraries with a [syslog reporter for `logs`](https://opam.ocaml.org/packages/logs-syslog/), and configure it to send the interesting log messages to these servers. (with an API to configure it through xapi)

Would that be useful for XO?

lippirk:
> Great feature! Just wanted to say thanks for this +1
> 
> Is there any discussion to think about expose that somehow in the API or would it be preferable to create a separate plugin to parse the log? Leveraging this would allow to increase overall security (eg we have an UI in XO to expose our own audit log. Combining it with host audit log could be great addition to compare both and find discrepancies or attacks)

What kind information would you like to see exposed by the API exactly? We're actually planning to use this client IP feature to add to limit the number of connections somebody can open from the same IP address. Hopefully we can find some overlap here

olivierlambert:
Remote syslog is also an approach yes, that could be done in parallel I suppose.

However for XO, having a turnkey tool/view that could just ask XAPI about access (time/IP/command) would be easier for us (I imagine on the `host` class). I don't know what's @julien-f opinion on that too

julien-f:
The ideal implementation for XO is an integration with the records and events system but that seems like a big departure from current implementation.

A simpler solution would be to allow downloading the log as a file (raw is acceptable but JSON would be better), either all entries or the a range of entries.

lippirk:
> The ideal implementation for XO is an integration with the records and events system but that seems like a big departure from current implementation.

We have actually implemented a feature which raises xapi alerts when a user fails to login: https://github.com/xapi-project/xen-api/pull/4218 (we will extend this slightly to include IP address information soon). I think this could be extended even further to expose time/IP/command information from xapi requests like Olivier you suggested, though it would be a bit of work.

> A simpler solution would be to allow downloading the log as a file (raw is acceptable but JSON would be better), either all entries or the a range of entries.

Would the aforementioned 'export logs to remote syslog server' suffice for the time being? https://support.citrix.com/article/CTX222747

olivierlambert:
https://github.com/xapi-project/xen-api/pull/4218 is a nice feature, thanks! Indeed, that would be really interesting to add few more information so those alerts will be "complete" :+1: 

About remote syslog: we are aware of the feature (which is nice and relevant if you have the knowledge in your company to use it and analyse them, eg with Kibana etc.). However, they aren't a lot of users/companies able to do that for their own infrastructure, and for them, having a single dashboard/UI to interact is what they need.

So it means we should configure Xen Orchestra to get the logs, but in the same time, it means another "channel" to talk with the host. But that's exactly why we **love** XAPI: being able to manage everything from the API itself, without extra means of communication (no remote SSH commands, no scripts etc.)

However, I'm aware those features requires work, so just trying to find a balance :)

PR #4235, opened 2020-10-21
Update feature/req-403 from master

lindig:
This merges master into feature/req-403 to keep it up-to-date.


psafont (approved):

lindig:
There are some sign-offs missing. This is annoying and I think we should ignore the problem for this merge.

lippirk (approved):

PR #4240, opened 2020-10-23
CP-35021 assert compatibility before suspend

lindig:
This is just for initial discussion. I will refine the commit message.


lippirk:
This change looks fine, but why would a VM be running with an incompatible vgpu/pgpu in the first place?

lindig:
I agree. This is something NVidia suggested as a test to see if suspend is supported. This can only work when we can start a VM without such test. We can't migrate to into this state because it would have to pass this test.

lippirk (approved):
In any case it can't hurt to have this code. Maybe add a comment to say that this check is here to see whether suspend is supported?

psafont (approved):

robhoes (approved):

PR #4241, opened 2020-10-23
maintenance: Schedule weekly run for 1.249-lcm

psafont:
Since the dependencies barely change, I felt like a weekly run / cache
is more than enough for testing. It depends on the rest of the packages
being attached to a commit / tag in opam instead of following a branch.

This commit is indeed on master branch, this is due to a limitation in github actions, it's the same as the scheduled runs in xs-opam: https://github.com/xapi-project/xs-opam/blob/master/.github/workflows/stockholm-lcm.yml

### DIFF
--- a/.github/workflows/1.249-lcm.yml
+++ b/.github/workflows/1.249-lcm.yml
@@ -1,0 +1,84 @@
+name: Build and test (1.249-lcm, scheduled)
+
+on:
+  schedule:
+    # run every Monday, this refreshes the cache
+    - cron: '13 2 * * 1'
+
+jobs:
+  python-test:
+    name: Python tests
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        test: ["", "-3"]
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          ref: '1.249-lcm'
+
+      - name: Run python tests
+        run: bash .github/python-nosetests${{ matrix.test }}.sh
+
+  ocaml-test:
+    name: Ocaml tests
+    runs-on: ubuntu-20.04
+    env:
+      package: "xapi-cli-protocol xapi-client xapi-consts xapi-database xapi-datamodel xapi-types xapi xe"
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          ref: '1.249-lcm'
+
+      - name: Pull configuration from xs-opam
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/1.249-lcm.yml
```diff
@@ -0,0 +1,84 @@
...
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        test: ["", "-3"]
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          ref: '1.249-lcm'
```
psafont, line 20: All checkout actions must use the ref `1.249-lcm` to checkout the correct branch.

.github/workflows/1.249-lcm.yml
```diff
@@ -0,0 +1,84 @@
...
+        run: echo "::set-output name=date::$(/bin/date -u "+%Y%W")"
+        shell: bash
+
+      - name: Restore opam cache
+        id: opam-cache
+        uses: actions/cache@v2
+        with:
+          path: "~/.opam"
+          # invalidate cache every week, gets built using a scheduled job
+          key: ${{ steps.cache-key.outputs.date }}-1.249
```
psafont, line 56: This cache key should be accessible from the rest of the branches, including `1.249-lcm`, according to documentation.

lippirk:
> Since the dependencies barely change, I felt like a weekly run / cache
> is more than enough for testing. It depends on the rest of the packages
> being attached to a commit / tag in opam instead of following a branch.

Presumably xen-api builds will be affected by other dependencies other than whatever we use in xs-opam (e.g. xenopsd)?

Is there a way to rebuild the cache manually?

lippirk (approved):

psafont:
> Presumably xen-api builds will be affected by dependencies other than whatever we use in xs-opam (e.g. xenopsd)?

I don't understand the question. Currently all the  packages in the stockholm branch in xs-opam (except xen-api's) have a source url bound to a tag and not a branch, so if they change, the `opam upgrade` command in the workflow will update ithe package even if the cache is not up-to-date.

> Is there a way to rebuild the cache manually?

I think the cache is overwritten on every successful build using it.

lindig (approved):
Could we do this also in xs-opam for the entire toolstack?

psafont:
We haven't done much work on the new github action workflow in xs-opam. I wanted to exercise it a bit before moving the lcm CI to it. I could upgrade dune, rresult and ppx_bisect to test it.

PR #4245, opened 2020-10-30
REQ-403 add module to split PEM files

lindig:
Add a module to split a PEM file into its components and return them as
strings.


psafont (approved):
Tests cases make it very easy to  check the correctness of the code, they look comprehensive to me.

psafont (approved):

lippirk (approved):

ocaml/xapi/xapi_globs.ml
```diff
@@ -730,6 +731,8 @@ let pool_secret_path = ref (Filename.concat "/etc/xensource" "ptoken")
 (* Path to server ssl certificate *)
 let server_cert_path = ref (Filename.concat "/etc/xensource" "xapi-ssl.pem")
 
+let stunnel_cert_path = ref "/etc/stunnel/xapi-stunnel-ca-bundle.pem"
```
psafont, line 734:
```suggestion
let stunnel_ca_cert_bundle_path = ref "/etc/stunnel/xapi-stunnel-ca-bundle.pem"
let stunnel_ca_certs_path = ref "/etc/stunnel/certs"
```
There are actually two paths for ca certificates, one pem file containing all ca certs, and one directory with a file per certificate.

psafont (changes requested):

ocaml/xapi/xapi_globs.ml
lindig: I think it would be better to deal only with one source of truth - either the PEM or the individual files

psafont: I will check if stunnel supports reading from a directory instead of a single pem file.
Handling certificates as files is convenient for installing and removing individual certificates because they have individual filenames. This is exploited by the current interface and I was thinking of exploiting it to keep them in sync with the database.

lippirk: I would suggest that we avoid duplicating this path (it's defined here, though not exposed via the mli yet https://github.com/xapi-project/xen-api-libs-transitional/blob/dc2e8d18f650f801744e099e94847adf4410cab7/stunnel/stunnel.ml#L27)

I would agree with Christian, and suggest that we use only one. If we were to use two, then we have to pass two parameters to stunnel (`CApath` and `CAfile`), but it's not obvious which one would take precedence

lindig: I updated it to use the directory as global configuration paramete.

psafont (approved):

PR #4247, opened 2020-11-05
Add Pool.enable_tls_verification to the API and include IP address in SANs of self-signed certs

lippirk:
Here we implement a naive version of `Pool.enable_tls_verification` by setting the global flag in the stunnel library.

We also change the way that xapi generates self-signed slightly, by making the IP address the CN, and adding the IP as a SAN.

Sibling: https://github.com/xapi-project/xen-api-libs-transitional/pull/85

### DIFF
--- a/ocaml/gencert/lib.ml
+++ b/ocaml/gencert/lib.ml
@@ -23,14 +23,20 @@ let hostnames () =
   let hostname = Unix.gethostname () in
   let fqdns =
     Unix.getaddrinfo hostname "" [Unix.AI_CANONNAME]
-    |> List.filter_map (fun addrinfo ->
-           match addrinfo.Unix.ai_canonname with
-           | "" ->
-               None
-           | name ->
-               Some name)
+    |> List.map (fun x -> x.Unix.ai_canonname)
   in
-  match fqdns with [] -> [hostname] | fqdns -> fqdns
+  hostname :: fqdns
+  |> List.filter_map (fun x ->
+         let x = Astring.String.trim x in
+         if
+           String.equal "" x
+           || String.equal "localhost" x
+           || Ipaddr.of_string x |> Stdlib.Result.is_ok
+         then
+           None
+         else
+           Some x)
+  |> Astring.String.uniquify
 
 let get_management_ip_addr ~dbg =
   let iface = Inventory.lookup Inventory._management_interface in
--- a/ocaml/gencert/selfcert.ml
+++ b/ocaml/gencert/selfcert.ml
@@ -43,22 +43,35 @@ let expire_in days =
   | None ->
       R.error_msgf "can't represent %d as time span" days
 
-let add_dns_names extension = function
-  | [] ->
-      extension
-  | names ->
-      X509.Extension.(
-        add Subject_alt_name
-          (false, X509.General_name.(singleton DNS names))
-          extension)
+let sans alt_names ip =
+  let ip_cstruct =
+    Ipaddr.of_string ip
+    |> Stdlib.Result.to_option
+    |> Option.map (function
+         | Ipaddr.V4 addr ->
+             Cstruct.of_string (Ipaddr.V4.to_octets addr)
+         | Ipaddr.V6 addr ->
+             Cstruct.of_string (Ipaddr.V6.to_octets addr))
+  in
+  let sans =
+    X509.General_name.(
+      singleton DNS alt_names
+      |>
+      match ip_cstruct with
+      | None ->
+          D.error "selfcert.ml: failed to convert ip='%s' to cstruct" ip ;
+          Fun.id
+      | Some ip ->
+          add IP [ip])
+  in
+  X509.Extension.(add Subject_alt_name (false, sans) X509.Extension.empty)
 
-let sign days key pubkey issuer req alt_names =
+let sign days privkey pubkey issuer req alt_names ip =
   expire_in days >>= fun (valid_from, valid_until) ->
-  match (key, pubkey) with
+  match (privkey, pubkey) with
   | `RSA priv, `RSA pub when Rsa.pub_of_priv priv = pub ->
-      let extensions = add_dns_names X509.Extension.empty alt_names in
-      X509.Signing_request.sign ~valid_from ~valid_until ~extensions req key
-        issuer
+      X509.Signing_request.sign ~valid_from ~valid_until
+        ~extensions:(sans alt_names ip) req privkey issuer
       |> R.reword_error (fun _ -> Printf.sprintf "signing failed" |> R.msg)
   | _ ->
       R.error_msgf "public/private keys don't match (%s)" __LOC__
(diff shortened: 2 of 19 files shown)


ocaml/gencert/lib.ml
```diff
@@ -23,14 +23,19 @@ let hostnames () =
...
-    |> List.filter_map (fun addrinfo ->
-           match addrinfo.Unix.ai_canonname with
-           | "" ->
-               None
-           | name ->
-               Some name)
+    |> List.map (fun x -> x.Unix.ai_canonname)
   in
-  match fqdns with [] -> [hostname] | fqdns -> fqdns
+  hostname :: fqdns
```
lippirk, line 28: I didn't include any lookups in /etc/hosts - I don't think this is necessary (`hostname :: fqnds` includes everything that I expected during testing). My justification is that self-signed certs of course work by default, when verification is disabled, and we expect users to create their own certificates anyway when enabling verification

psafont: `Unix.getaddrinfo` resolves the hostnames to IP using `nsswitch.conf`, which uses currently `hosts:      files dns myhostname`, which means mappings in `/etc/hosts` takes precedence over dns resolution.

ocaml/gencert/selfcert.ml
```diff
@@ -43,22 +43,33 @@ let expire_in days =
...
+    |> Option.map (function
+         | Ipaddr.V4 addr ->
+             Cstruct.of_string (Ipaddr.V4.to_octets addr)
+         | Ipaddr.V6 addr ->
+             Cstruct.of_string (Ipaddr.V6.to_octets addr))
+  in
+  let sans =
+    X509.General_name.(
+      singleton DNS alt_names
+      |> match ip_cstruct with None -> Fun.id | Some ip -> add IP [ip])
```
psafont, line 59: I think having an ip that cannot be parsed should be logged.

ocaml/gencert/selfcert.ml
```diff
@@ -107,22 +118,17 @@ let selfsign name alt_names length days certfile =
   in
   let privkey = `RSA rsa in
   let pubkey = `RSA (Rsa.pub_of_priv rsa) in
-  let issuer =
-    [X509.Distinguished_name.(Relative_distinguished_name.singleton (CN name))]
-  in
-  let req = X509.Signing_request.create issuer privkey in
```
psafont, line 113: I'm not sure why this code was moved, it may make backporting more difficult

ocaml/gencert/lib.ml
psafont: The uniquify to this list was dropped on the `Selfcert.host` function, I think it should be reintroduced here to avoid duplicates..

ocaml/idl/datamodel_pool.ml
```diff
@@ -804,6 +813,7 @@ open Datamodel_types
          ; field ~in_product_since:rel_inverness ~qualifier:DynamicRO ~ty:Bool ~default_value:(Some (VBool false)) "igmp_snooping_enabled" "true if IGMP snooping is enabled in the pool, false otherwise."
          ; field ~in_product_since:rel_quebec ~qualifier:RW ~ty:String ~default_value:(Some (VString "")) "uefi_certificates" "The UEFI certificates allowing Secure Boot"
          ; field ~in_product_since:rel_stockholm_psr ~qualifier:RW ~ty:Bool ~default_value:(Some (VBool false)) "is_psr_pending" "True if either a PSR is running or we are waiting for a PSR to be re-run"
-         ; field ~qualifier:DynamicRO ~lifecycle:[Published, rel_next, ""] ~ty:(Set (Ref _certificate)) "certificates" "CA certificates installed in the pool";
+         ; field ~qualifier:DynamicRO ~lifecycle:[Published, rel_next, ""] ~ty:(Set (Ref _certificate)) "certificates" "CA certificates installed in the pool"
+         ; field ~qualifier:RW ~in_product_since:rel_next ~lifecycle:[Published, rel_next, ""] ~ty:Bool ~default_value:(Some (VBool false)) "tls_verification_enabled" "True iff TLS certificate verification is enabled"
```
psafont, line 817: Long term this should be enabled by default and it must not be allowed to turn it off during normal operation.

ocaml/xapi/message_forwarding.ml
```diff
@@ -922,6 +922,17 @@ functor
         info "Pool.assert_can_enable_tls_verification: pool = '%s'"
           (current_pool_uuid ~__context) ;
         Local.Pool.assert_can_enable_tls_verification ~__context
+
+      let enable_tls_verification ~__context =
+        info "Pool.enable_tls_verification: pool = '%s'"
+          (current_pool_uuid ~__context) ;
+        let self = Helpers.get_pool ~__context in
+        let local_fn = Local.Pool.enable_tls_verification in
```
psafont, line 930: This should call `Local.Pool.assert_can_enable_tls_verification` before trying to enable verification.

ocaml/gencert/lib.ml
lippirk: Indeed, so `getaddrinfo` is using /etc/hosts implicitly. I've reintroduced the uniquify here

ocaml/gencert/selfcert.ml
lippirk: I did it to reduce the number of params to the sign function, but you are right. I've moved the code back

ocaml/idl/datamodel_pool.ml
lippirk: Indeed. I believe it must always be RW unfortunately, in order to implement `Pool.emergency_disable_tls_verification`

robhoes: I think this should be `DynamicRO`. We don't want clients to set this directly, and rather use the API call above, don't we?

lippirk: > xapi itself can always write to any field

I misunderstood the meaning of DynamicRO. In this case I definitely agree with you and Pau!

ocaml/gencert/selfcert.ml
```diff
@@ -43,22 +43,35 @@ let expire_in days =
...
+      | None ->
+          D.error "selfcert.ml: failed to convert ip='%s' to cstruct" ip ;
+          Fun.id
+      | Some ip ->
+          add IP [ip])
+  in
+  X509.Extension.(add Subject_alt_name (false, sans) X509.Extension.empty)
 
-let sign days key pubkey issuer req alt_names =
+let sign days privkey pubkey issuer req cn alt_names ip =
```
psafont, line 69: the cn parameter is not used in the function.

lippirk:
Please can I get another review on this? It'd be nice to merge both this PR and https://github.com/xapi-project/xen-api-libs-transitional/pull/85

(The CI failure is expected - the signature of `with_remove` changed in https://github.com/xapi-project/xen-api-libs-transitional/pull/85)

psafont (approved):

lindig:
As pointed out by GitHub this needs a rebase.

lindig (approved):

PR #4254, opened 2020-11-12
CP-35280: Verify certificate chains when installing certificates

psafont:
For now it contains the changes in gencert validation needed to enable it.
    
Now when installing a certificate the trust anchors from stunnel are used to validate the certificate. This means that additional checks are done to the chains and intermediate certificates they contain.
    
Now hashes are enforced for the whole chain (sha2), the chains must be in order from leaf to root certificate. Additionally, properties of x509v3 like path len are enforced when checking the chains.
    
When validating a chain a list of root cert fingerprints that validate the certificate are returned. This is needed to fill the database with correct information.

TODO:
- [ ] add unit tests for validation of full chains
- [ ] Add the information to the database, incorporating all changes from the WIP branch: https://github.com/psafont/xen-api/tree/balances


lindig:
Small thing: but I think `Gencertlib.Lib` is not looking good and I would suggest to rename this library to `gencert`as we are accumulating more modules. For documentation I would find a  `validation.mli` useful.

ocaml/gencert/validation.ml
```diff
@@ -0,0 +1,247 @@
...
+           D.info {|Failed to validate private key because "%s"|} err_msg ;
+           `Msg (server_certificate_key_invalid, [])
+         ))
+  >>= ensure_key_length
+
+let has_valid_time time certificate =
+  let not_before, not_after = X509.Certificate.validity certificate in
+  Ptime.(is_later ~than:not_before time && is_earlier ~than:not_after time)
+
+let load_server_certificate pem =
```
lindig, line 62: This is not a good name as it implies loading from the file system. This is just decoding a PEM.

ocaml/gencert/validation.ml
```diff
@@ -0,0 +1,247 @@
...
+
+let verify_chain_of_trust ~server ~private_key ~intermediates ~time ~anchors =
+  validate_server_certificate server time private_key >>= fun server ->
+  (* build all paths *)
+  let paths = X509.Validation.build_paths server intermediates in
+  let anchors = List.filter (has_valid_time time) anchors in
+  (* return valid chains *)
+  valid_chains time anchors paths
+
+let load_ca_cert_bundle ca_cert_bundle_path =
```
lindig, line 195: Here load means loading from the file system - which I think is the right name,

ocaml/gencert/lib.ml
```diff
@@ -65,107 +65,36 @@ let get_management_ip_addr ~dbg =
...
-
-let validate_certificate kind pem now private_key =
-  let ensure_keys_match private_key certificate =
-    let public_key = X509.Certificate.public_key certificate in
-    match (public_key, private_key) with
-    | `RSA pub, `RSA priv when pub = Mirage_crypto_pk.Rsa.pub_of_priv priv ->
-        Ok certificate
-    | _ ->
-        Error (`Msg (server_certificate_key_mismatch, []))
+let install_server_certificate ?(pem_chain = None) ~pem_leaf ~pkcs8_private_key
```
lindig, line 70: The validations imply that an invalid cert cannot be installed. There is some friction with the startup behaviour where we also verify the certificate but in addition check if TLS verification is on and currently would install the certificate if TLS is off. I believe this needs to be simplified and would suggest to never install a certificate that is known to be invalid even with TLS validation off as it would create problems down the road.

lippirk:
> Small thing: but I think `Gencertlib.Lib` is not looking good and I would suggest to rename this library to `gencert`as we are accumulating more modules. For documentation I would find a `validation.mli` useful.

I agree that gencert is a bit confusing now. However, my vote would be to leave the refactoring until the churn in there has stopped (I'm currently modifying it in another open PR, and will be changing it more in a future PR)

ocaml/gencert/validation.ml
```diff
@@ -0,0 +1,247 @@
...
+  Xapi_stdext_unix.Unixext.string_of_file ca_cert_bundle_path
+  |> Cstruct.of_string
+  |> X509.Certificate.decode_pem_multiple
+  |> function
+  | Ok (_ :: _ as trust_anchors) ->
+      Ok trust_anchors
+  | _ ->
+      Error (`Msg (server_certificate_chain_invalid, []))
+
+let validate_certificate_chain pem_chain =
```
lindig, line 205: Is this a good name? The function decodes the PEM and does some syntactic checking but is this a validation? The name implies more to me.

ocaml/gencert/validation.ml
```diff
@@ -51,58 +51,197 @@ let validate_private_key pkcs8_private_key =
...
+
+let verify_single_chain time anchors chain =
+  let open X509.Validation in
+  let rec climb pathlen = function
+    | cert :: issuer :: certs ->
+        validate_intermediate_certificate issuer time >>= fun _ ->
+        signs pathlen issuer cert >>= fun () ->
+        climb (succ pathlen) (issuer :: certs)
+    | [c] ->
+        let anchors = issuers_of anchors c in
```
lippirk, line 163: It's not obvious whether these anchors have are 'valid'. I believe they are because of `List.filter (has_valid_time time) anchors` below, but it would be simpler to do this check / filtering inside `load_ca_cert_bundle`

ocaml/gencert/validation.ml
```diff
@@ -51,58 +51,197 @@ let validate_private_key pkcs8_private_key =
...
-    | `RSA pub, `RSA priv when pub = Mirage_crypto_pk.Rsa.pub_of_priv priv ->
-        Ok certificate
-    | _ ->
-        Error (`Msg (server_certificate_key_mismatch, []))
-  in
+let has_valid_time time certificate =
+  let not_before, not_after = X509.Certificate.validity certificate in
+  Ptime.(is_later ~than:not_before time && is_earlier ~than:not_after time)
+
+let load_server_certificate pem =
```
lippirk, line 62: Similarly to my comment about anchors, I think it would make sense to perform the cert validation in this constructor (unless there's a reason for keeping expired certs around?)

ocaml/gencert/lib.ml
psafont: I guess on startup if tls verification is off nothing is done; or else the pem is read and installed using `install_server_certificate`. It adds some unnecessary writing, but it adds more regularity. The exceptions have to be caught correctly so they don't interrupt the startup process, however.

ocaml/gencert/validation.ml
psafont: This the "validation" that is done to chains in 8.2. I agree it doesn't provide much value right now, but I'm unsure on how to factor it out. It's a consequence on the function signature chosen for `install_server_certificate`.

ocaml/gencert/lib.ml
lindig: The current plan is the check the validity of the PEM at startup. If the verification code in this function could be pulled out, it could be re-used. Otherwise it would have to be replicated as this function verifies the certificate and installs it.

psafont: I think exposing the chain validation functionality by itself is perfectly acceptable, that should be enough to validate the PKCS12 file present in the filesystem at startup.

ocaml/gencert/validation.ml
psafont: I've changed the function that simply decode pem formats and make some sanity checks to `decode_...` to avoid confusion.

ocaml/gencert/lib.ml
psafont: I've offloaded all the decoding and validation to the function `validating_trust_anchors`. This returns the decoded server certificate and the list of validating anchors, which is allowed to be empty.

I'm pondering whether this should be allowed, as it means the certificate could not be validated and therefore it should return an error instead.

ocaml/gencert/validation.ml
psafont: This has been renamed to `decode_certificate_chain`.

ocaml/gencert/validation.ml
psafont: This function does not exist and its functionality is contained in `validate_server_certificate`

lindig (approved):
I like the changes and improved documentation.

robhoes (approved):
I haven't reviewed the details, but from a high-level point of view this looks good to me. I'm happy to merge this into the feature branch.

Issue #4256, opened 2020-11-13
Enable IPV6 through xe-enable-ipv6


benjamreis:
Hi

Enabling IPv6 has been discussed in [this PR](https://github.com/xapi-project/xcp-networkd/pull/156/files) ; from what I understood there were a disagrement with splitting the default/global enablement.

This fix looks good to me regardless of how to enable ipv6: https://github.com/xapi-project/xcp-networkd/pull/156/files#diff-b6603d795adbfdc73d5d1482e296a5d8ff25d5cf24184df2c974ff9f71395c81L527-R543

How about changing `xe-enable-ipv6` to modify the entries of `/etc/sysctl.d/90-net.conf`? Would that be good?
```
net.ipv6.conf.all.disable_ipv6 = 1
net.ipv6.conf.default.disable_ipv6 = 1
```

Thanks!

benjamreis:
Also: should the ipv6 address be stored in the `address` field of the host when enabled?

benjamreis:
Why is it not possible to change the primary address type of the management interface?
https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_pif.ml#L782-L785

What happens in the case of an IPv6 only network?

benjamreis:
To answer my own question:
> Also: should the ipv6 address be stored in the address field of the host when enabled?
It is, if the management interface PIf's `primary-address-type` is set to ipv6.

PR #4259, opened 2020-11-18
xe-enable-ipv6 edits net.ipv6.conf.{ all | default }.disable_ipv6

benjamreis:
Would solve: https://github.com/xapi-project/xen-api/issues/4256

This question remains: https://github.com/xapi-project/xen-api/issues/4256#issuecomment-727892082

### DIFF
--- a/scripts/xe-enable-ipv6
+++ b/scripts/xe-enable-ipv6
@@ -23,7 +23,7 @@ fi
 . /etc/xensource-inventory
 
 if [ $1 = "enable" ]; then
-	rm /etc/modprobe.d/disable-ipv6
+	rm /etc/modprobe.d/disable-ipv6.conf
 
 	# Enable IPv6 networking
 	sed -i '/^NETWORKING_IPV6=/d' /etc/sysconfig/network
@@ -46,8 +46,8 @@ elif [ $1 = "disable" ]; then
 	fi
 
 	chkconfig ip6tables off
- 	echo "alias ipv6 no" > /etc/modprobe.d/disable-ipv6
-	echo "alias net-pf-10 off" >> /etc/modprobe.d/disable-ipv6
+ 	echo "alias ipv6 no" > /etc/modprobe.d/disable-ipv6.conf
+	echo "alias net-pf-10 off" >> /etc/modprobe.d/disable-ipv6.conf
 	sed -i '/^NETWORKING_IPV6=/d' /etc/sysconfig/network
 	sed -i '/^IPV6_AUTOCONF=/d' /etc/sysconfig/network
 	echo "NETWORKING_IPV6=NO" >> /etc/sysconfig/network
--- a/scripts/xe-enable-ipv6
+++ b/scripts/xe-enable-ipv6
@@ -32,10 +32,8 @@ if [ $1 = "enable" ]; then
 	echo "IPV6_AUTOCONF=NO" >> /etc/sysconfig/network
 	chkconfig ip6tables on
 
-	sed -i '/^net.ipv6.conf.all.disable_ipv6=/d' /etc/sysctl.d/91-net-ipv6.conf
-	sed -i '/^net.ipv6.conf.default.disable_ipv6=/d' /etc/sysctl.d/91-net-ipv6.conf
-	echo "net.ipv6.conf.all.disable_ipv6=0" >> /etc/sysctl.d/91-net-ipv6.conf
-	echo "net.ipv6.conf.default.disable_ipv6=0" >> /etc/sysctl.d/91-net-ipv6.conf
+	echo "net.ipv6.conf.all.disable_ipv6=0" > /etc/sysctl.d/91-net-ipv6.conf
+	echo "net.ipv6.conf.default.disable_ipv6=0" > /etc/sysctl.d/91-net-ipv6.conf
 
 	echo "IPv6 enabled.  You may now need to reboot the host"
 elif [ $1 = "disable" ]; then
@@ -55,10 +53,8 @@ elif [ $1 = "disable" ]; then
 	echo "NETWORKING_IPV6=NO" >> /etc/sysconfig/network
 	echo "IPV6_AUTOCONF=NO" >> /etc/sysconfig/network
 
-	sed -i '/^net.ipv6.conf.all.disable_ipv6=/d' /etc/sysctl.d/91-net-ipv6.conf
-	sed -i '/^net.ipv6.conf.default.disable_ipv6=/d' /etc/sysctl.d/91-net-ipv6.conf
-	echo "net.ipv6.conf.all.disable_ipv6=1" >> /etc/sysctl.d/91-net-ipv6.conf
-	echo "net.ipv6.conf.default.disable_ipv6=1" >> /etc/sysctl.d/91-net-ipv6.conf
+	echo "net.ipv6.conf.all.disable_ipv6=1" > /etc/sysctl.d/91-net-ipv6.conf
+	echo "net.ipv6.conf.default.disable_ipv6=1" > /etc/sysctl.d/91-net-ipv6.conf
 
 	echo "IPv6 disabled.  You may now need to reboot the host"
 else
--- a/scripts/xe-enable-ipv6
+++ b/scripts/xe-enable-ipv6
@@ -33,7 +33,7 @@ if [ $1 = "enable" ]; then
 	chkconfig ip6tables on
 
 	echo "net.ipv6.conf.all.disable_ipv6=0" > /etc/sysctl.d/91-net-ipv6.conf
-	echo "net.ipv6.conf.default.disable_ipv6=0" > /etc/sysctl.d/91-net-ipv6.conf
+	echo "net.ipv6.conf.default.disable_ipv6=0" >> /etc/sysctl.d/91-net-ipv6.conf
 
 	echo "IPv6 enabled.  You may now need to reboot the host"
 elif [ $1 = "disable" ]; then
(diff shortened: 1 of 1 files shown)


benjamreis:
with this:
- the files `/proc/sys/net/ipv6/conf/{ all | default }/disable_ipv6` are being modified.
- but not `net.ipv6.conf.{ all | default }.disable_ipv6` entries in `/etc/sysctl.d/90-net.conf`, is this enough?

scripts/xe-enable-ipv6
```diff
@@ -23,20 +23,23 @@ fi
 . /etc/xensource-inventory
 
 if [ $1 = "enable" ]; then
-	rm /etc/modprobe.d/disable-ipv6 
-	
+	rm /etc/modprobe.d/disable-ipv6
```
benjamreis, line 26: shouldn't the file be named `/etc/modprobe.d/disable-ipv6.conf`?
like here: https://github.com/xcp-ng/host-installer/blob/512877bfab35a729e4986658b8d71365235bbfb9/netutil.py#L438

psafont: I would think so as well. the installer is definitely writing to `etc/modprobe.d/disable-ipv6.conf`.

benjamreis: I can add the edit to my PR if you want. :)

lippirk: If you modify this, make sure to modify other references in the 'disable' flow:

```
echo "alias ipv6 no" > /etc/modprobe.d/disable-ipv6
echo "alias net-pf-10 off" >> /etc/modprobe.d/disable-ipv6
```

benjamreis: On it!

lippirk (approved):

scripts/xe-enable-ipv6
```diff
@@ -23,33 +23,39 @@ fi
...
 	# Enable IPv6 networking
 	sed -i '/^NETWORKING_IPV6=/d' /etc/sysconfig/network
 	sed -i '/^IPV6_AUTOCONF=/d' /etc/sysconfig/network
 	echo "NETWORKING_IPV6=YES" >> /etc/sysconfig/network
 	echo "IPV6_AUTOCONF=NO" >> /etc/sysconfig/network
 	chkconfig ip6tables on
-	
+
+	sysctl -q -w "net.ipv6.conf.all.disable_ipv6=0"
+	sysctl -q -w "net.ipv6.conf.default.disable_ipv6=0"
```
benjamreis, line 36: it seems this command overrides the entries but it's lost at reboot.Is there a way to write in `/etc/sysctl.d/90-net.conf`?

lippirk: I don't think there is a particularly nice way, but you could use a combination of `grep`, `sed` and `>>`

benjamreis: Yes I was afraid of that. But i didn't find anything offering another solution. ^^

psafont: Would it make sense to move the ipv6 settings to `/etc/sysctl.d/91-net-ipv6.conf`. So the whole file can be written instead of being a partial update?

benjamreis: No idea, I'm okay to do that if you and everyone in the xapi team are in favor. :)

psafont: Could you please go ahead with the new config file?

benjamreis: On it! :)

benjamreis: Done, should i check that the file alreydy exists? and also should i remove the settings from `90-net.conf`? Or should it be done elsewhere.

scripts/xe-enable-ipv6
```diff
@@ -32,8 +32,8 @@ if [ $1 = "enable" ]; then
 	echo "IPV6_AUTOCONF=NO" >> /etc/sysconfig/network
 	chkconfig ip6tables on
 
-	sysctl -q -w "net.ipv6.conf.all.disable_ipv6=0"
-	sysctl -q -w "net.ipv6.conf.default.disable_ipv6=0"
+	echo "net.ipv6.conf.all.disable_ipv6=0" >> /etc/sysctl.d/91-net-ipv6.conf
```
psafont, line 37: Shouldn't this be just `>` to write on top of the existing file instead of appending the contents? (same in line 56)

benjamreis: I used sed to remove the entries.
So that a `net.ipv6.conf.<iface>.disable_ipv6` entry can be put and not erased.

benjamreis: I think I should do the same sed command for `90-net.conf`  to be sure that the `disable_ipv6` entries are removed from this file?

benjamreis:
PR has been merged, I think this issue can be closed. :)

Thanks for the help! :+1:

scripts/xe-enable-ipv6
psafont: It's easier to reason about this if every time the contents of the file are written instead of searching code to delete then append contents.

Unless I've missed something this code is the only code that writes to this file, there's no point in being "nice" and expect that the file might have more contents.

benjamreis: I guess you're right yeah.
I'll change that then.
I think I still need to seek and erase the entry in `90-net.conf`

psafont: I don't see any mentions in 90-net.conf here. Is it coming from the default centos configuration?

benjamreis:
CI failure is unrelated to my changes.

scripts/xe-enable-ipv6
```diff
@@ -32,10 +32,8 @@ if [ $1 = "enable" ]; then
 	echo "IPV6_AUTOCONF=NO" >> /etc/sysconfig/network
 	chkconfig ip6tables on
 
-	sed -i '/^net.ipv6.conf.all.disable_ipv6=/d' /etc/sysctl.d/91-net-ipv6.conf
-	sed -i '/^net.ipv6.conf.default.disable_ipv6=/d' /etc/sysctl.d/91-net-ipv6.conf
-	echo "net.ipv6.conf.all.disable_ipv6=0" >> /etc/sysctl.d/91-net-ipv6.conf
-	echo "net.ipv6.conf.default.disable_ipv6=0" >> /etc/sysctl.d/91-net-ipv6.conf
+	echo "net.ipv6.conf.all.disable_ipv6=0" > /etc/sysctl.d/91-net-ipv6.conf
+	echo "net.ipv6.conf.default.disable_ipv6=0" > /etc/sysctl.d/91-net-ipv6.conf
```
psafont, line 36:
```suggestion
	echo "net.ipv6.conf.default.disable_ipv6=0" >> /etc/sysctl.d/91-net-ipv6.conf
```

scripts/xe-enable-ipv6
```diff
@@ -55,10 +53,8 @@ elif [ $1 = "disable" ]; then
 	echo "NETWORKING_IPV6=NO" >> /etc/sysconfig/network
 	echo "IPV6_AUTOCONF=NO" >> /etc/sysconfig/network
 
-	sed -i '/^net.ipv6.conf.all.disable_ipv6=/d' /etc/sysctl.d/91-net-ipv6.conf
-	sed -i '/^net.ipv6.conf.default.disable_ipv6=/d' /etc/sysctl.d/91-net-ipv6.conf
-	echo "net.ipv6.conf.all.disable_ipv6=1" >> /etc/sysctl.d/91-net-ipv6.conf
-	echo "net.ipv6.conf.default.disable_ipv6=1" >> /etc/sysctl.d/91-net-ipv6.conf
+	echo "net.ipv6.conf.all.disable_ipv6=1" > /etc/sysctl.d/91-net-ipv6.conf
+	echo "net.ipv6.conf.default.disable_ipv6=1" > /etc/sysctl.d/91-net-ipv6.conf
```
psafont, line 57:
```suggestion
	echo "net.ipv6.conf.default.disable_ipv6=1" >> /etc/sysctl.d/91-net-ipv6.conf
```

scripts/xe-enable-ipv6
psafont: Both lines have to be added ;)

benjamreis: Yeah I saw that, not fully awake this morning i guess ^^'

psafont (approved):

scripts/xe-enable-ipv6
robhoes: On CH:
```
# rpm -qf /etc/sysctl.d/90-net.conf 
xenserver-release-config-8.2.50-8.xs8.x86_64
# tail -n3 /etc/sysctl.d/90-net.conf 
# Disable IPv6 on interfaces
net.ipv6.conf.all.disable_ipv6 = 1
net.ipv6.conf.default.disable_ipv6 = 1
```

robhoes: By default, on my host, IPv6 is disabled in /etc/sysctl.d/90-net.conf, while the kernel ipv6 module is still loaded. So I don't think that we need that  /etc/modprobe.d/disable-ipv6 stuff anymore?

And this PR now seems to create a 91-net.conf file that conflicts with 90-net.conf?

benjamreis: The modprobe thing is probably useless yes.
This PR creates a new config file IPv6 specific yes.
Indeed for now it conflicts with 90-net.conf, I'm currently searching where the 90-net.conf ipv6 entries are written in the file.

robhoes: For Citrix Hypervisor, the 90-net.conf file is just a static file in the xenserver-release-config RPM.

stormi: And it's the same in XCP-ng, only the name of the RPM changes.

It's defined as `%config(noreplace)` so once put there it can be altered by users or anything without being replaced in future updates, except ISO upgrades.

Should the installer and/or XAPI modify it when needed to enable ipv6?

Or should the defaults be different at installation, and/or the ipv6-related settings extracted to a separate configuration file so that we can let the installer/XAPI rewrite it entirely when needed?

robhoes: Perhaps splitting up those conf files is the best here. So in xenserver-release-config and the xcp-ng equivalent, move the IPv6 lines from 90-net.conf to 91-net-ipv6.conf, and then this current PR seems to do the right thing (but just delete the modprobe stuff while we are here).

stormi: This looks like a good solution to me.

benjamreis: LGTM as well, i'll do the change for modprobe and we'll change the config file in our rpm :)

Thanks for the help!

robhoes (approved):

PR #4260, opened 2020-11-18
CP-34463: pool.conf should use FQDN

lippirk:
Slaves store an address to the master in pool.conf. Historically this
has been an IP address, however for REQ-403 we need to use an fqdn
instead.

### DIFF
--- a/ocaml/xapi-aux/pool_role.ml
+++ b/ocaml/xapi-aux/pool_role.ml
@@ -23,11 +23,7 @@ module D = Debug.Make (struct let name = "pool_role" end)
 open D
 
 (** The role of this node *)
-type t =
-  | Master
-  | Slave of string
-  (* IP address *)
-  | Broken
+type t = Master | Slave of string | Broken
 
 let role = ref None
 
@@ -101,8 +97,8 @@ exception This_host_is_broken
 
 let get_master_address () =
   match get_role () with
-  | Slave ip ->
-      ip
+  | Slave addr ->
+      addr
   | Master ->
       raise This_host_is_a_master
   | Broken ->
--- a/ocaml/xapi-aux/pool_role.mli
+++ b/ocaml/xapi-aux/pool_role.mli
@@ -12,7 +12,7 @@
  * GNU Lesser General Public License for more details.
  *)
 
-type t = Master | Slave of string  (** IP address *) | Broken
+type t = Master | Slave of string  (** IP address or FQDN *) | Broken
 
 val with_pool_role_lock : (unit -> unit) -> unit
 
@@ -41,4 +41,5 @@ exception This_host_is_a_master
 exception This_host_is_broken
 
 val get_master_address : unit -> string
-(** If this node is a slave, returns the IP address of its master *)
+
+(** If this node is a slave, returns the address of its master. *)
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -1021,15 +1021,16 @@ let is_valid_MAC mac =
          acc && String.length s = 2 && validchar s.[0] && validchar s.[1])
        true l
 
-(** Returns true if the supplied IP address looks like one of mine *)
+(** Returns true if the supplied address looks like one of mine *)
 let this_is_my_address ~__context address =
   let dbg = Context.string_of_task __context in
-  let inet_addrs =
+  let ipv4s =
     Net.Interface.get_ipv4_addr dbg
       (Xapi_inventory.lookup Xapi_inventory._management_interface)
   in
-  let addresses = List.map Unix.string_of_inet_addr (List.map fst inet_addrs) in
-  List.mem address addresses
+  let addresses = List.map Unix.string_of_inet_addr (List.map fst ipv4s) in
+  let hostnames = Gencertlib.Lib.dns_sans () in
+  List.mem address (List.append addresses hostnames)
 
 (** Returns the list of hosts thought to be live *)
 let get_live_hosts ~__context =
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -1041,44 +1041,6 @@ let get_live_hosts ~__context =
       try Db.Host_metrics.get_live ~__context ~self:metrics with _ -> false)
     hosts
 
-let gethostbyname_family host family =
-  let throw_resolve_error () =
-    failwith (Printf.sprintf "Couldn't resolve hostname: %s" host)
-  in
-  let getaddr x =
-    match x with
-    | Unix.ADDR_INET (addr, port) ->
-        addr
-    | _ ->
-        failwith "Expected ADDR_INET"
-  in
-  let he =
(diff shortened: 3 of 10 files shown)


ocaml/xapi-aux/pool_role.ml
```diff
@@ -23,11 +23,7 @@ module D = Debug.Make (struct let name = "pool_role" end)
 open D
 
 (** The role of this node *)
-type t =
-  | Master
-  | Slave of string
-  (* IP address *)
```
psafont, line 29: We should keep the comment to note the string is the adress of the master host

ocaml/xapi-aux/pool_role.mli
```diff
@@ -12,7 +12,11 @@
  * GNU Lesser General Public License for more details.
  *)
 
-type t = Master | Slave of string  (** IP address *) | Broken
+type t =
+  | Master
+  | Slave of string
+  (* ip or fqdn *)
+  | Broken
```
psafont, line 19:
```suggestion
type t = Master | Slave of string  (** IP address or FQDN *) | Broken
```
We can convert this to a doccomment

ocaml/xapi-aux/pool_role.mli
```diff
@@ -41,4 +45,5 @@ exception This_host_is_a_master
 exception This_host_is_broken
 
 val get_master_address : unit -> string
-(** If this node is a slave, returns the IP address of its master *)
+
+(* If this node is a slave, returns the address of its master. *)
```
psafont, line 49: Was was it changed from a doccoment?

ocaml/xapi/helpers.ml
```diff
@@ -1021,15 +1021,20 @@ let is_valid_MAC mac =
...
       (Xapi_inventory.lookup Xapi_inventory._management_interface)
   in
   let addresses = List.map Unix.string_of_inet_addr (List.map fst inet_addrs) in
-  List.mem address addresses
+  let other_names =
+    let hostname = Unix.gethostname () in
+    let fqdns = Gencertlib.Lib.fqdns_of_hostname hostname in
+    hostname :: fqdns
+  in
+  List.mem address (List.append addresses other_names)
```
psafont, line 1037: I think changing the names of the list to `ipv4s` and `hostnames` would help with readability (ipv6 addresses are missing here, pinging @benjamreis)

benjamreis: Should I add it in my PR? after a rebase.

ocaml/xapi/xapi.ml
```diff
@@ -559,6 +564,33 @@ let resynchronise_ha_state () =
...
+              if current_master_address <> master_fqdn then (
+                D.info
+                  "set_fqdn_in_pool_conf: updating pool.conf with master_fqdn: \
+                   %s"
+                  master_fqdn ;
+                Slave master_fqdn |> Xapi_pool_transition.set_role
+              ) else
+                D.info "set_fqdn_in_pool_conf: not modifying pool.conf"
+          | None ->
+              D.error
```
lindig, line 588: This logs an error but does not communicate the error to the caller. How does the caller become aware of the problem?

ocaml/xapi/xapi_ha.ml
```diff
@@ -78,36 +78,51 @@ let local_failover_decisions_are_ok () =
...
-      (Db.Host.get_all_records ~__context)
+(** Since the liveset info doesn't include the host address, we persist these ourselves *)
+let write_uuid_to_address_mapping ~__context =
+  let uuid_fqdn_of_host (_, host) =
+    let uuid = host.API.host_uuid in
+    let hostname = host.API.host_hostname in
+    match Gencertlib.Lib.fqdn_of_hostname hostname with
+    | None ->
+        raise
+          Api_errors.(
```
psafont, line 89: wouldn't `hostname` be enough in this case? I'm wary about introducing exceptions in the startup process.

ocaml/xapi-aux/pool_role.mli
lippirk: I've changed it back

ocaml/xapi/helpers.ml
lippirk: @benjamreis You won't want to rebase on top of this branch because it includes feature work, but indeed it would be good to make this function work correctly in the ipv6 case

benjamreis: Sure, I can make a new PR pointing to this one. I think it's the simpler solution. What do you think?

ocaml/xapi/xapi.ml
lippirk: This is being called as part of the xapi startup sequence, so if it raised then I believe it would take down xapi's main thread. We could run this in a separate thread, but that seems like overkill

ocaml/xapi/xapi_ha.ml
lippirk: While testing https://github.com/xapi-project/xen-api/pull/4247/files, I found that stunnel will refuse to connect unless the `checkHost` parameter matches the FQDN (hostname doesn't work), so not throwing here _could_ lead to subtle bugs

I'm not sure how you worked out that this is called in the startup process - I see that it gets used by `preconfigure_ha`, which is presumably only called when enabling ha?

ocaml/xapi/helpers.ml
benjamreis: Hmm, for some reason I can't make a PR towards `lippirk/pool-conf`. How can I supply my patch?

psafont: It will take months until this work is merged to master. We merge master to the feature branch from time you time. I'd recommend to have the patch done against master and we'll resolve the conflict when merging master to the feature branch, it should be an easy conflict to resolve.

ocaml/xapi/xapi_ha.ml
psafont: i don't know how I reached that conclusion, my bad!

ocaml/xapi/helpers.ml
benjamreis: https://github.com/xapi-project/xen-api/pull/4261 here you go.

ocaml/xapi/helpers.ml
```diff
@@ -1021,15 +1021,20 @@ let is_valid_MAC mac =
...
       (Xapi_inventory.lookup Xapi_inventory._management_interface)
   in
-  let addresses = List.map Unix.string_of_inet_addr (List.map fst inet_addrs) in
-  List.mem address addresses
+  let addresses = List.map Unix.string_of_inet_addr (List.map fst ipv4s) in
+  let hostnames =
+    let hostname = Unix.gethostname () in
+    let dns_sans = Gencertlib.Lib.dns_sans_of_hostname hostname in
+    hostname :: dns_sans
+  in
```
psafont, line 1036:
```suggestion
  let hostnames = Gencertlib.Lib.dns_sans () in
```

psafont (approved):

lindig (approved):

Issue #4264, opened 2020-11-25
console location are wrong when the management interface is on IPv6


benjamreis:
Hi!

During my IPv6 investigations I found that the `location` field of a console is wrong when in IPv6:
```	
# xe console-param-list uuid=10dd2ea7-a936-2fed-49f7-de644ffc5c6b 
uuid ( RO)             : 10dd2ea7-a936-2fed-49f7-de644ffc5c6b
          vm-uuid ( RO): f45a6608-1ec2-4da7-808c-5104bac657cd
    vm-name-label ( RO): Control domain on host: localhost.localdomain
         protocol ( RO): VT100
         location ( RO): https://::1/console?ref=OpaqueRef:233bcd2e-3f23-4880-8cc4-6413c21b7218
     other-config (MRW): 
16:34



# xe console-param-list uuid=b5182380-6605-2162-9688-db1f1bb60678 
uuid ( RO)             : b5182380-6605-2162-9688-db1f1bb60678
          vm-uuid ( RO): f45a6608-1ec2-4da7-808c-5104bac657cd
    vm-name-label ( RO): Control domain on host: localhost.localdomain
         protocol ( RO): RFB
         location ( RO): https://::1/console?ref=OpaqueRef:19f301f1-b4c8-42d4-b9f2-3a7dc3a20242
     other-config (MRW):
```

When in an HTTP(S) URL an IPv6 needs to be surrended by `[]`
So the location should be like this: `https://[<ip6>]/console?ref=...`

I don't know where this field is filled but I am willing to try to provide patch if given direction. :)

benjamreis:
Is there other places where http url are created from an ip address?

robhoes:
Look here: https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/dbsync_master.ml#L74-L92

benjamreis:
PR merged

PR #4265, opened 2020-11-25
Fix IPv6 console location

benjamreis:
In a HTTP URL, IPv6 addresses needs to be between `[]`.

Solves https://github.com/xapi-project/xen-api/issues/4264


lippirk (approved):

robhoes (approved):
Thanks!

PR #4266, opened 2020-11-26
ci: ensure opam points to the correct opam repo

lippirk:


.github/workflows/main.yml
```diff
@@ -60,9 +60,11 @@ jobs:
 
       - name: Install dependencies
         run: |
-          # HACK: this should be set already by the 'Use ocaml' step, but it is not
+          # HACK: the repo should be set already by the 'Use ocaml' step, but it is not
           # (this might be because of the cache)
+          opam repo remove default
```
psafont: If the repo is added and removed it's already up-to-date. I recommend not adding this line, updating the url and updating the repo should be cheaper than cloning it whole.

psafont (approved):
Looks a good workaround to the strange interaction between the cache and the ocaml-setup action. I don't think this would hurt if it ends up to master, but it might be preferrable to remove it before merging to master.

lippirk:
> Looks a good workaround to the strange interaction between the cache and the ocaml-setup action. I don't think this would hurt if it ends up to master, but it might be preferrable to remove it before merging to master.

Indeed - maybe we can keep this out of master for now. Let's first see if we encounter the problem on other branch (strangely we haven't had an issue with 1.249-lcm)

psafont:
that's because 1.249-lcm uses a separate cache from the master one, check the name.

edwintorok:
Is the cache per branch or do the various branches keep overwriting each-other's caches?
Perhaps it'd be best to include the branch name automatically in the cache key?

psafont:
We could make a per-branch cache but that would mean that PRs would lose the cache from the master branch, which is the main purpose of accelerating the CI.

lindig (approved):

PR #4267, opened 2020-11-26
CP-34942: update dmidecode parser for angstrom 0.14

psafont:
The code relied on previous behavior to only parse a prefix and not the
whole text. This means in practice that if a failure is found the parser
emits the list of successfully parsed records instead of failing.

Should be merged along https://github.com/xapi-project/xs-opam/pull/524

### DIFF
--- a/ocaml/tests/test_bios_strings.ml
+++ b/ocaml/tests/test_bios_strings.ml
@@ -108,8 +108,10 @@ let check_values = Alcotest.(check @@ list @@ pair string string)
 
 let check_records = Alcotest.(check @@ result (list alco_record) string)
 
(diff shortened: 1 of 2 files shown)


edwintorok:
I assume you've tested all this together on Jenkins?

edwintorok (approved):

lindig (approved):

psafont:
I've tested it locally. This behaviour is captured by unit tests so I didn't there was a need to test it using Jenkins.

lindig:
This will also affect REQ-403 because `pem.ml` also uses angstrom. But it is a one-line change.

PR #4268, opened 2020-11-27
Allow migration on IPv6 only host

benjamreis:

Still testing.

### DIFF
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -2246,20 +2246,45 @@ let migrate_receive ~__context ~host ~network ~options =
            ( Api_errors.host_cannot_attach_network
            , [Ref.string_of host; Ref.string_of network] ))
   in
-  let ip = Db.PIF.get_IP ~__context ~self:pif in
-  ( if String.length ip = 0 then
-      match Db.PIF.get_ip_configuration_mode ~__context ~self:pif with
-      | `None ->
-          raise
-            (Api_errors.Server_error
-               (Api_errors.pif_has_no_network_configuration, [Ref.string_of pif]))
-      | `DHCP ->
-          raise
-            (Api_errors.Server_error
-               (Api_errors.interface_has_no_ip, [Ref.string_of pif]))
-      | _ ->
-          failwith "No IP address on PIF"
+  let primary_address_type =
+    Db.PIF.get_primary_address_type ~__context ~self:pif
+  in
+  let ip, configuration_mode =
+    match primary_address_type with
+    | `IPv4 ->
+      ( Db.PIF.get_IP ~__context ~self:pif,
+      Db.PIF.get_ip_configuration_mode ~__context ~self:pif )
+    | `IPv6 ->
+      let configuration_mode =
+        Db.PIF.get_ipv6_configuration_mode ~__context ~self:pif
+      in
+      match Db.PIF.get_IPv6 ~__context ~self:pif with
+      | [] -> ("", configuration_mode)
+      | ip::_ ->
+        (* The CIDR is also stored in the IPv6 field of a PIF. *))
+        let ipv6 =
+          match String.split_on_char '/' ip with
+          | hd::_ -> hd
+          | _ -> ""
+        in
+        (ipv6, configuration_mode)
+  in
+  ( if ip = "" then
+    match configuration_mode with
+    | `None ->
+      raise
+        (Api_errors.Server_error
+          (Api_errors.pif_has_no_network_configuration, [Ref.string_of pif]))
+    | _ ->
+      raise
+        (Api_errors.Server_error
+          (Api_errors.interface_has_no_ip, [Ref.string_of pif]))
   ) ;
+  let ip =
+    match primary_address_type with
+    | `IPv4 -> ip
+    | `IPv6 -> Http.Url.maybe_wrap_IPv6_literal ip
+  in
   let sm_url =
     Printf.sprintf "http://%s/services/SM?session_id=%s" ip new_session_id
   in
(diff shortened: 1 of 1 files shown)


benjamreis:
For now, when attempting to migrate a VM towards an IPv6 only host I get `PIF_HAS_NO_NETWORK_CONFIGURATION` because only IPv4 configuration mode is checked.

This Pr does the same for IPv6.

ocaml/xapi/xapi_host.ml
```diff
@@ -2246,20 +2246,50 @@ let migrate_receive ~__context ~host ~network ~options =
...
+    Db.PIF.get_primary_address_type ~__context ~self:pif
+  in
+  let ip =
+    match primary_address_type with
+    | `IPv4 ->
+      Db.PIF.get_IP ~__context ~self:pif
+    | `IPv6 ->
+      List.hd (Db.PIF.get_IPv6 ~__context ~self:pif)
+  in
   ( if String.length ip = 0 then
```
lindig, line 2259:
```
if ip - "" then
```

ocaml/xapi/xapi_host.ml
```diff
@@ -2246,20 +2246,50 @@ let migrate_receive ~__context ~host ~network ~options =
...
-          raise
-            (Api_errors.Server_error
-               (Api_errors.pif_has_no_network_configuration, [Ref.string_of pif]))
-      | `DHCP ->
-          raise
-            (Api_errors.Server_error
-               (Api_errors.interface_has_no_ip, [Ref.string_of pif]))
-      | _ ->
-          failwith "No IP address on PIF"
+      match primary_address_type with
```
lindig, line 2260: I think a flatter structure would be more readable:

```
match primary_address_type, Db.PIF.get_ip_configuration_mode ~__context ~self:pif with
| `IPv4, `None ->
| `IPv4, `DHCP ->
..
```

benjamreis: Yes, much better indeed. Thanks !

benjamreis: No need in fact, since both configuration_mode can be compared to `None` and `DHCP`

ocaml/xapi/xapi_host.ml
```diff
@@ -2246,20 +2246,36 @@ let migrate_receive ~__context ~host ~network ~options =
...
+  let primary_address_type =
+    Db.PIF.get_primary_address_type ~__context ~self:pif
+  in
+  let ip, configuration_mode =
+    match primary_address_type with
+    | `IPv4 ->
+      ( Db.PIF.get_IP ~__context ~self:pif,
+      Db.PIF.get_ip_configuration_mode ~__context ~self:pif )
+    | `IPv6 ->
+      ( List.hd (Db.PIF.get_IPv6 ~__context ~self:pif),
```
robhoes, line 2258: Avoid `List.hd`, which gives nasty exceptions. You could make `ip` a list, and just remove the `List.hd` here and add `[ ... ]` above.

benjamreis: How can i get the first elem of a list "cleanly" then?

benjamreis: Because later I need the value `ip` as a string to add the extra `[]` around IPv6 for URL.

ocaml/xapi/xapi_host.ml
```diff
@@ -2246,20 +2246,36 @@ let migrate_receive ~__context ~host ~network ~options =
...
+      Db.PIF.get_ipv6_configuration_mode ~__context ~self:pif )
+  in
+  ( if ip = "" then
+    match configuration_mode with
+    | `None ->
+      raise
+        (Api_errors.Server_error
+          (Api_errors.pif_has_no_network_configuration, [Ref.string_of pif]))
+    | `DHCP ->
```
robhoes, line 2267:
```
| `DHCP | `Autoconf` ->
```
I think?

ocaml/xapi/xapi_host.ml
```diff
@@ -2246,20 +2246,36 @@ let migrate_receive ~__context ~host ~network ~options =
...
+      raise
+        (Api_errors.Server_error
+          (Api_errors.interface_has_no_ip, [Ref.string_of pif]))
+    | _ ->
+      failwith "No IP address on PIF"
```
robhoes, line 2272: This is a weird case actually. It's only for mode=static without IP, which is "impossible" by construction, and therefore and internal error. However, we might as well raise the error above here as well... Then you can collapse this and the previous case under a single `| _ ->`.

ocaml/xapi/xapi_host.ml
robhoes:
```
match primary_address_type, ip with
    | `IPv4, ip :: _ -> ip
    | `IPv6, ip :: _ -> Http.Url.maybe_wrap_IPv6_literal ip
```

benjamreis: what if I do this:

```ocaml
  let ip, configuration_mode =
    match primary_address_type with
    | `IPv4 ->
      ( Db.PIF.get_IP ~__context ~self:pif,
      Db.PIF.get_ip_configuration_mode ~__context ~self:pif )
    | `IPv6 ->
      ( match Db.PIF.get_IPv6 ~__context ~self:pif with
         | [] -> ""
         | hd::tl -> hd,
      Db.PIF.get_ipv6_configuration_mode ~__context ~self:pif )
  in
```

robhoes: That's fine as well.

benjamreis: Yours is more concise, i'll do that.

lindig (approved):

benjamreis:
Removed the `WIP`, the issue I encountered were in other repo. :+1:

robhoes (approved):
This looks good.

PR #4269, opened 2020-11-27
Continue fixing console location in IPv6

benjamreis:

I missed some places where consoles' location were computed.

### DIFF
--- a/ocaml/xapi/dbsync_master.ml
+++ b/ocaml/xapi/dbsync_master.ml
@@ -84,15 +84,7 @@ let refresh_console_urls ~__context =
             | "" ->
                 ""
             | address ->
-              let address =
-                match Xapi_stdext_unix.Unixext.domain_of_addr address with
-                | Some x when x = Unix.PF_INET ->
-                  address
-                | Some x when x = Unix.PF_INET6 ->
-                  "[" ^ address ^ "]"
-                | _ ->
-                  ""
-              in
+              let address = Http.Url.maybe_wrap_IPv6_literal address in
               Printf.sprintf "https://%s%s?ref=%s" address
                 Constants.console_uri (Ref.string_of console)
           in


benjamreis:
I think there's nothing to do here:
https://github.com/xapi-project/xen-api/blob/master/ocaml/vncproxy/vncproxy.ml#L86-L100

is that correct?

ocaml/xapi/create_misc.ml
```diff
@@ -337,6 +337,15 @@ and create_domain_zero_console_record_with_protocol ~__context ~domain_zero_ref
...
     Db.Host.get_address ~__context ~self:(Helpers.get_localhost ~__context)
   in
+  let address =
+    match Xapi_stdext_unix.Unixext.domain_of_addr address with
+    | Some x when x = Unix.PF_INET ->
+      address
+    | Some x when x = Unix.PF_INET6 ->
+      Http.Url.maybe_wrap_IPv6_literal address
+    | _ ->
+      ""
```
lippirk, line 347: it may be better to keep the old behaviour, and return `address` when `x` is a unix socket (though it shouldn't make a difference as far as I can tell). It would also simplify the match slightly:

```
let address =
  match domain_of_addr address with
  | Some x when x = Unix.PF_INET6 ->
    Http.Url.maybe_wrap_IPv6_literal address
  | _ -> address
```

ocaml/xapi/dbsync_master.ml
```diff
@@ -89,7 +89,7 @@ let refresh_console_urls ~__context =
                 | Some x when x = Unix.PF_INET ->
                   address
                 | Some x when x = Unix.PF_INET6 ->
-                  "[" ^ address ^ "]"
+                  Http.Url.maybe_wrap_IPv6_literal address
```
lippirk, line 92: Does this fix a bug where we might have accidentally wrapped addresses in two sets of braces `[]`?

ocaml/xapi/xapi_xenops.ml
```diff
@@ -2049,6 +2049,15 @@ let update_vm ~__context id =
                       (fun (protocol, _) ->
                         let ref = Ref.make () in
                         let uuid = Uuid.to_string (Uuid.make_uuid ()) in
+                        let uri =
```
lippirk, line 2052: You can give uri a similar treatment as above

ocaml/xapi/dbsync_master.ml
benjamreis: If a port si specified then the `[]` are already present.
I don't know if it can happen in this specific case but I think using a dedicated method is better than handling manually.

lippirk: Indeed - a good spot :+1:

lippirk (approved):
I just noticed that `| Some Unix.PF_INET6` is more idiomatic than `| Some x when x =...` - feel to make that change too :stuck_out_tongue:

benjamreis:
> I just noticed that `| Some Unix.PF_INET6` is more idiomatic than `| Some x when x =...` - feel to make that change too

The more concise the better ;)

benjamreis:
Just Saw: https://github.com/xapi-project/xen-api-libs/blob/master/http-svr/http.ml#L619-L621

It means I don't need to check the address family before calling `maybe_unwrap_IPv6_literal`, i'll make one more commit to remove al the checks, this will simplify my changes.

benjamreis:
There are many other instances of `http://%s, address` and `https://%s, address` i the code, but they're not related to a VM console so I'll make another PR to modify the other instances. :)

lindig (approved):

ocaml/xapi/xapi_xenops.ml
```diff
@@ -2049,6 +2049,7 @@ let update_vm ~__context id =
                       (fun (protocol, _) ->
                         let ref = Ref.make () in
                         let uuid = Uuid.to_string (Uuid.make_uuid ()) in
+                        let uri = Http.Url.maybe_wrap_IPv6_literal uri in
```
robhoes, line 2052: Here, `uri` isn't just an address (see L2014), so you'll need to do this higher up.

ocaml/xapi/create_misc.ml
```diff
@@ -337,6 +337,7 @@ and create_domain_zero_console_record_with_protocol ~__context ~domain_zero_ref
   let address =
     Db.Host.get_address ~__context ~self:(Helpers.get_localhost ~__context)
   in
+  let address = Http.Url.maybe_wrap_IPv6_literal address in
```
robhoes, line 340:
```suggestion
  let address =
    Db.Host.get_address ~__context ~self:(Helpers.get_localhost ~__context)
    |> Http.Url.maybe_wrap_IPv6_literal
  in
```

robhoes (changes requested):
See above.

ocaml/xapi/xapi_xenops.ml
benjamreis: Indeed!

benjamreis:
Most changes were already done after rebasing on master. ^^

robhoes (approved):

PR #4270, opened 2020-11-30
REQ-403 emergency disable tls verification

lippirk:
An escape hatch is needed in case communication breaks down in a pool.
Importantly verification will only be disabled for the host on which you
make this emergency call. This introduces a discrepancy between the pool
database and the local database, which can be fixed by executing
Pool.enable_tls_verification once again.

### DIFF
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -2461,3 +2461,38 @@ let emergency_disable_tls_verification ~__context =
   Localdb.put Constants.tls_verification_enabled "false" ;
   (* disable verification now *)
   Stunnel.set_verify_tls_certs false
+
+let health_check ~__context =
+  let tls_verification_enabled_locally =
+    Localdb.get_with_default Constants.tls_verification_enabled "false"
+    |> bool_of_string
+  in
+  let tls_verification_enabled_pool_wide =
+    Db.Pool.get_tls_verification_enabled ~__context
+      ~self:(Helpers.get_pool ~__context)
+  in
+  (* Only add an alert if (a) we found a problem and (b) an alert doesn't already exist *)
+  if
+    tls_verification_enabled_pool_wide
+    && tls_verification_enabled_pool_wide <> tls_verification_enabled_locally
+  then
+    let alert_exists =
+      Helpers.call_api_functions ~__context (fun rpc session ->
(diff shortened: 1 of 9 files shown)


robhoes:
Do we have a way to make an API client aware of the fact that verification is disabled on a particular host, while it's meant to be enabled on a pool level, so that they know that they should call the enable function again?

lippirk:
> Do we have a way to make an API client aware of the fact that verification is disabled on a particular host, while it's meant to be enabled on a pool level, so that they know that they should call the enable function again?

I don't believe this is planned, but that sounds like a good idea. Some ideas:

1. add an assertion that local tls verification settings match the pool setting to a health check
2. add a parameter to host record (so that you can run `xe host-list params=tls-verification-enabled`)

I quite like option 1, as it shouldn't require any changes for XenCenter

lindig (approved):
This looks good to me on the code level (no opinion on the feature but it makes sense to me).

ocaml/xapi/xapi_host.ml
```diff
@@ -2461,3 +2461,36 @@ let emergency_disable_tls_verification ~__context =
...
+          Client.Client.Message.get_all_records rpc session
+          |> List.exists (fun (_, record) ->
+                 record.API.message_name = fst Api_messages.local_health_check))
+    in
+
+    if not alert_exists then
+      let self = Helpers.get_localhost ~__context in
+      let msg =
+        Printf.sprintf
+          "TLS verification is not enabled on %s - please re-enable!"
```
psafont, line 2490: Could we separate the message into two xml tags? I think it would help with translation.

It might even be desirable to create a message type exclusive to this check and then only the host might need to be transmitted.

@kc284 do you have recommendations about this?


Here's what I mean by 'two xml tags'
```
<message>%s</message><host>%s</host>
```

kc284: Yes. The message text itself is not very important for the implementing client because it will be localised, but the client will need to extract the host's ref from the alert; retrieving it from an xml tag is safer than extracting it via a regex which will break should the message text change in future. May I also suggest to wrap the above in one more tag, e.g. `<body><message>%s</message><host>%s</host></body>`? So the xml parser does not complain about 2 root nodes when the xml is fed to it as is (same as we did for the certificate alerts).

psafont (approved):

PR #4271, opened 2020-11-30
CA-348700: Block VDI.copy if on-boot=reset

robhoes:
Performing a VDI copy on a VDI that has reset-on-boot enabled may lead
to the VDI to be wiped completely, rather than reset to a predefined
state, when a VM is started. Storage migrate is blocked for VMs that
have such VDIs, and VDI.copy should have the same constraint.


robhoes:
Still to be tested.

psafont (approved):

lindig (approved):

edwintorok (approved):

PR #4272, opened 2020-12-01
Merge master to feature/REQ-403 

psafont:
This makes the branch compatible with angstrom 0.14, includes a commit to adapt the feature code to the latest library as well.

Depends on https://github.com/xapi-project/xs-opam/pull/527


lippirk (approved):

lindig (approved):

lindig:
The CI fails because it does not (yet) use the updated opam, I suspect. I would suggest to merge and figure this out locally rather than using the CI.

PR #4273, opened 2020-12-01
CP-35280: Install self-signed certificates as trust anchors, add more information to database about certificates

psafont:
Now the trust anchors that verify a server certificate in a pool are store in the server certificate's database record. This means that CA certificates must be also recorded in the database when installed.

CA certificates are not removed from the database when uninstalled yet. This is for another issue that tackles CA certificates exclusively.

For self-signed certificates, these are generated, installed normally as pool ca certificates with their UUID as names and then the database record is marked as host_and_pool and the host is added to the record, as well as the field verified_by.

### DIFF
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -1377,7 +1377,7 @@ let install_server_certificate ~__context ~host ~certificate ~private_key
   let pem_chain =
     match certificate_chain with "" -> None | pem_chain -> Some pem_chain
   in
-  let certificate =
+  let certificate, root_fingerprints =
     Certificates.install_server_certificate ~pem_leaf:certificate
       ~pkcs8_private_key:private_key ~pem_chain
   in
@@ -1425,7 +1425,9 @@ let reset_server_certificate ~__context ~host =
         xs
   in
   let cn = ip in
-  Gencertlib.Selfcert.host cn alt_names xapi_ssl_pem ip ;
+  let _certificate, _root_fingerprint =
+    Gencertlib.Selfcert.generate cn alt_names xapi_ssl_pem ip
+  in
   (* Reset stunnel to try to restablish TLS connections *)
   Xapi_mgmt_iface.reconfigure_stunnel ~__context ;
   (* Delete records of the server certificate in this host *)
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -1366,6 +1366,55 @@ let certificate_sync ~__context ~host = Certificates.local_sync ()
 let get_server_certificate ~__context ~host =
   Certificates.get_server_certificate ()
 
+let add_certificate_to_db ~__context ~host certificate root_fingerprints =
+  let date_of_ptime time = Date.of_float (Ptime.to_float_s time) in
+  let dates_of_ptimes (a, b) = (date_of_ptime a, date_of_ptime b) in
+  let not_before, not_after =
+    dates_of_ptimes (X509.Certificate.validity certificate)
+  in
+  let raw_fingerprint =
+    X509.Certificate.fingerprint Mirage_crypto.Hash.(`SHA256) certificate
+  in
+  let fingerprint = Certificates.pp_hash raw_fingerprint in
+  let uuid = Uuid.(to_string (make_uuid ())) in
+  let ref = Ref.make () in
+  let _type =
+    if List.mem raw_fingerprint root_fingerprints then `host_and_pool else `host
+  in
+  Db.Certificate.create ~__context ~ref ~uuid ~_type ~host ~pool:Ref.null
+    ~not_before ~not_after ~fingerprint ;
+  (* Link host to trust anchors *)
+  let root_certs =
+    root_fingerprints
+    |> List.map (fun root_fingerprint ->
+           let expr =
+             Db_filter_types.(
+               Eq
+                 ( Field "fingerprint"
+                 , Literal (Certificates.pp_hash root_fingerprint) ))
+           in
+           Db.Certificate.get_refs_where ~__context ~expr)
+    |> List.flatten
+    |> List.sort_uniq Stdlib.compare
+  in
+  Db.Certificate.set_validated_by ~__context ~self:ref ~value:root_certs ;
+  ref
+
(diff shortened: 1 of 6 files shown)


ocaml/xapi/xapi_host.ml
```diff
@@ -1366,45 +1366,92 @@ let certificate_sync ~__context ~host = Certificates.local_sync ()
...
+    dates_of_ptimes (X509.Certificate.validity certificate)
+  in
+  let raw_fingerprint =
+    X509.Certificate.fingerprint Mirage_crypto.Hash.(`SHA256) certificate
+  in
+  let fingerprint = Certificates.pp_hash raw_fingerprint in
+  let uuid = Uuid.(to_string (make_uuid ())) in
+  let ref = Ref.make () in
+  let _type =
+    if List.mem raw_fingerprint root_fingerprints then `host_and_pool else `host
```
lippirk, line 1383: Would `self_signed` be a better name than `host_and_pool`?

ocaml/xapi/xapi_host.ml
```diff
@@ -1366,45 +1366,92 @@ let certificate_sync ~__context ~host = Certificates.local_sync ()
...
+                 , Literal (Certificates.pp_hash root_fingerprint) ))
+           in
+           Db.Certificate.get_refs_where ~__context ~expr)
+    |> List.flatten
+    |> List.sort_uniq Stdlib.compare
+  in
+  Db.Certificate.set_validated_by ~__context ~self:ref ~value:root_certs ;
+  ref
+
+let uninstall_outdated_certificates ~__context outdated_server_certs =
```
lippirk, line 1404: This is a misleading name since we are not checking that certs are outdated here ( `let uninstall_certificates ~__context certs` is more accurate ? )

ocaml/xapi/xapi_host.ml
```diff
@@ -1425,15 +1472,30 @@ let reset_server_certificate ~__context ~host =
...
+
   (* Reset stunnel to try to restablish TLS connections *)
   Xapi_mgmt_iface.reconfigure_stunnel ~__context ;
-  (* Delete records of the server certificate in this host *)
-  let expr =
-    Db_filter_types.(Eq (Field "host", Literal (Ref.string_of self)))
-  in
-  Db.Certificate.get_refs_where ~__context ~expr
-  |> List.iter (fun self -> Db.Certificate.destroy ~__context ~self)
+  uninstall_outdated_certificates ~__context current_server_certs
```
lippirk, line 1498: If we restart stunnel before uninstalling the old certificates, then won't the pool potentially have some connections which are verified using the old certificates? (this might be bad for long running connections established at start up, but equally might not be a concern for us, since those 'old' certificates are still valid)

(In reality this is probably not an issue, because I believe the call to `reconfigure_stunnel` is threaded, and so by the time stunnel is back up the old certificates will have been removed)

ocaml/xapi/xapi_pool.ml
```diff
@@ -2130,7 +2130,59 @@ let retrieve_wlb_recommendations ~__context = get_opt_recommendations ~__context
...
+     To install a self signed:
+       1. write to filesystem
+       2. distribute cert & add to db (here)
+       3. modify the db record so it's a self-signed certificate
+  *)
+  if refs <> [] then
+    raise Api_errors.(Server_error (internal_error, [name])) ;
+
+  (* only names with .pem extension are allowed. *)
+  let uuid, name =
```
lippirk, line 2149: What if `name` already has a .pem extension? If that's not allowed then we should at least check

Also you didn't add .pem in the None case which is confusing

psafont: You're completely right, this is just minimal work to make the normal certificate installation to work. There is an ticket that specifically targets proper CA certificate installation (and removal!). In general CA certificates must _not_ be replaced or removed if they are the only trust anchor for a server certificate (host certificate). Self-signed need special management.

Currently the implementation details in the certificates module are being leaked. I'll try to see if I can plug some of the problems, but I don't want to spend too much time since it's going to be replaced soon.

ocaml/xapi/xapi_host.ml
psafont: I'm torn about this. It has a very specific meaning in the database:

```
         ; "host_and_pool", "Certificate that both identifies a host and is trusted by the whole pool."
```

lippirk: Let's leave it as it is. It's simpler if the name matches the name in the database

ocaml/xapi/xapi_pool.ml
psafont: For the time being, I've simplified the setup by assuming the name passed to the api is just a name and not the full valid filename. This makes it impossible for a user to rewrite a xapi-installed certificate.

ocaml/xapi/xapi_pool.ml
```diff
@@ -2130,7 +2130,62 @@ let retrieve_wlb_recommendations ~__context = get_opt_recommendations ~__context
...
+  let expr = Db_filter.(Eq (Field "uuid", Literal name)) in
+  let refs = Db.Certificate.get_refs_where ~__context ~expr in
+  (* Do not allow users to overwrite certificates installed by xapi
+     To install a self signed:
+       1. write to filesystem
+       2. distribute cert & add to db (here)
+       3. modify the db record so it's a self-signed certificate
+  *)
+  if refs <> [] then
+    raise Api_errors.(Server_error (internal_error, [name])) ;
```
lippirk, line 2147:
```suggestion
    raise Api_errors.(Server_error (internal_error, [Printf.sprintf "certificate already exists with name '%s'" name])) ;
```

lindig (approved):

lippirk (approved):

ocaml/xapi/xapi_host.ml
psafont: > won't the pool potentially have some connections which are verified using the old certificates?

I don't think so: any new connection established after the stunnel reset will offer the new certificate to clients. This is why it's safe to delete the server certificates from the pool's trust anchors at this point and not safe to do before.

> I believe the call to reconfigure_stunnel is threaded, and so by the time stunnel is back up the old certificates will have been removed

That might be a problem. in this case the call must be made blocking.

PR #4274, opened 2020-12-03
REQ-819 stockholm schema version bump

lippirk:
We added `is_psr_pending` field in a Stockholm
hotfix, so we should have bumped the schema version.

### DIFF
--- a/ocaml/idl/datamodel_common.ml
+++ b/ocaml/idl/datamodel_common.ml
@@ -8,7 +8,7 @@ open Datamodel_roles
               When introducing a new release, bump the schema minor version to the next hundred
               to leave a gap for potential hotfixes needing to increment the schema version.*)
 let schema_major_vsn = 5
-let schema_minor_vsn = 601
+let schema_minor_vsn = 602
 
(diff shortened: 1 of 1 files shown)


edwintorok (approved):
It would be good if we implemented a "schema checksum" and stored the expected schema checksum for each version number in a table in this file.
Then we would detect situations that need a version bump that we missed (e.g. in a backport).

The schema checksum could be something as simple as a sorted list of classes and fields, and perhaps some stable serializaiton of types (we rarely change types, so as a first approximation just the fields would be good enough).

lippirk:
> It would be good if we implemented a "schema checksum" and stored the expected schema checksum for each version number in a table in this file.
> Then we would detect situations that need a version bump that we missed (e.g. in a backport).
> 
> The schema checksum could be something as simple as a sorted list of classes and fields, and perhaps some stable serializaiton of types (we rarely change types, so as a first approximation just the fields would be good enough).

That a good idea - we may even be able to calculate a checksum for each datamodel class directly, for example for the pool class using `Datamodel_pool.t`

robhoes (approved):

PR #4275, opened 2020-12-03
maintenance: detect schema updates which are missing version bumps

lippirk:
Implementation of @edwintorok 's suggestion to use hashing to detect datamodel changes: https://github.com/xapi-project/xen-api/pull/4274#pullrequestreview-543739076

### DIFF
--- a/ocaml/idl/schematest.ml
+++ b/ocaml/idl/schematest.ml
@@ -1,0 +1,27 @@
+let hash x = Digest.string x |> Digest.to_hex
+
+(* BEWARE: if this changes, check that schema has been bumped accordingly *)
+let last_known_schema_hash = "5a6605a08c21f658bebcbfee9e93dc62"
+
+let current_schema_hash : string =
+  let open Datamodel_types in
+  let hash_of_obj x =
+    List.map rpc_of_content x.contents
+    |> List.map Jsonrpc.to_string
+    |> String.concat ""
+    |> hash
+  in
+  Datamodel.all_system |> List.map hash_of_obj |> String.concat ":" |> hash
+
+let () =
+  if last_known_schema_hash <> current_schema_hash then (
+    Printf.eprintf
+      {|
+
+New schema hash ('%s') doesn't match the last known one. Please bump the
+datamodel schema versions if necessary, and update 'last_known_schema_hash'.
+
+|}
+      current_schema_hash ;
+    exit 1
+  )
--- a/ocaml/idl/dune
+++ b/ocaml/idl/dune
@@ -50,3 +50,9 @@
    xapi-datamodel
   )
 )
+
+(test
+  (name schematest)
+  (modules schematest)
+  (libraries xapi_datamodel)
+)


ocaml/idl/datamodel_types.ml
```diff
@@ -510,6 +510,9 @@ type obj = {
   obj_doc_tags: doc_tag list;
 } [@@deriving rpc]
 
+let hash_of_obj x =
+  rpc_of_obj x |> Jsonrpc.to_string |> Hashtbl.hash
```
edwintorok, line 514: Hashtbl.hash can change between OCaml versions, perhaps use `Digest.string` instead

edwintorok: Perhaps sorting might be useful.
See  this code from the testsuite:
```
 let rec canonicalise r =
    match r with
>>  | Rpc.Dict x ->
      Rpc.Dict
        (List.sort (fun (s1, _) (s2, _) -> compare s1 s2) x
        |> List.map (fun (x, y) -> x, canonicalise y))
    | Rpc.Enum ys -> Rpc.Enum (List.map canonicalise ys)
    | x -> x
```

edwintorok (commented):
Thanks, this is a good start!

ocaml/idl/schematest.ml
```diff
@@ -0,0 +1,39 @@
...
+
+let schema_hash : int =
+  [
+    Datamodel_certificate.t
+  ; Datamodel_cluster_host.t
+  ; Datamodel_cluster.t
+  ; Datamodel_diagnostics.t
+  ; Datamodel_host.t
+  ; Datamodel_pool.t
+  ; Datamodel_vm.t
```
edwintorok, line 14: Is there a way to get this list from the datamodel itself?
E.g. adding a completely new class should require a bump as well. There is a Datamodel.all_system list, would that work?

ocaml/idl/datamodel_types.ml
lippirk: I'm not convinced this is so useful - why would you want to reorder the fields (other than if you're adding a new field anyway)? 

If you do want to do some stylistic change to the datamodel, then you can do it in one go and then you only have to add a single new hash to the `hash_to_version` map

ocaml/idl/schematest.ml
```diff
@@ -0,0 +1,58 @@
...
+      failwith "New schema hash"
+  | Some (major, minor) ->
+      if
+        major <> Datamodel_common.schema_major_vsn
+        || minor <> Datamodel_common.schema_minor_vsn
+      then (
+        Printf.eprintf
+          {|
+
+Got known schema hash, but this corresponds to an unexpected schema version. did you bump the schema version unnecessarily?
```
psafont, line 51: This is what usually happens when bumping the major version, it's done when there is a new release, not when there are changes in the schema. (and then the minor is also usually bumped along with it)

lippirk: That's a good point. I think this means we'll want hashes to correspond to multiple versions, instead of just one version

(We could alternatively include the version in the hash, but this creates a chicken vs egg problem like @lindig mentioned!)

psafont: Ah, I think there is a change needed in this case so it makes sense to exit here with an error since the version in this file doesn't match the one in the datamodel.

I suggest to only change the message to say that these two values are out of sync and must match.

lindig (approved):
I think this is a good idea. The hash depends on the serialisation of the IDL and is sensitive to syntactic changes. However, this is still more stable than the internal OCaml representation, which also could have been used. 

Once the IDL is updated, it hashes to a new value. I don't see how we can ever expect to find a hash of a previous version and therefore the purpose of keeping them around. It would mean that we reverted to a previous version or that we have a hash collision.

psafont (approved):

ocaml/idl/schematest.ml
```diff
@@ -0,0 +1,72 @@
...
+    (Datamodel_common.schema_minor_vsn + 1)
+
+let major_bump hash =
+  Printf.sprintf {|("%s", [(%i, %i)])|} hash
+    (Datamodel_common.schema_major_vsn + 1)
+    0
+
+let current_schema_hash : string =
+  let hash_of_obj x =
+    Datamodel_types.rpc_of_obj x |> Jsonrpc.to_string |> hash
```
robhoes, line 21: We only need to look at the `contents` of an object (the fields). You don't need a new schema version if you're just adding a function.

ocaml/idl/datamodel_types.ml
robhoes: The order of fields is actually important, and shouldn't change.

robhoes (approved):

psafont:
Something's not right:
```
Download action repository 'actions/checkout@v2'
Warning: Failed to download action 'api.github.com/repos/actions/checkout/tarball/5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f'. Error: Response status code does not indicate success: 500 (Internal Server Error).
```

The DCO bot is also acting wonky, probably because of the change of the base branch

lippirk:
> Something's not right:
> 
> ```
> Download action repository 'actions/checkout@v2'
> Warning: Failed to download action 'api.github.com/repos/actions/checkout/tarball/5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f'. Error: Response status code does not indicate success: 500 (Internal Server Error).
> ```
> 
> The DCO bot is also acting wonky, probably because of the change of the base branch

I was getting 500 earlier as well... silly github

PR #4277, opened 2020-12-10
CP-35284 validate and install updated host certs at start

lindig:
If a host certificate is present in the file system at xapi start:

* verify and install it if none was present
* check if the information about it is still valid
* verify, update, and install it if it has changed

Installing a new self-signed certificate at startup in a pool is
dangerous unless the host is the only member of the pool. Check the pool
size to verify this.

If verification fails when a valid host certificate is required, enter
emergency mode because communication will be impossible with an invalid
host certificate.

### DIFF
--- a/ocaml/xapi/certificates_sync.ml
+++ b/ocaml/xapi/certificates_sync.ml
@@ -1,0 +1,150 @@
+module D = Debug.Make (struct let name = "certificates_sync" end)
+
+open D
+module Unixext = Xapi_stdext_unix.Unixext
+module Date = Xapi_stdext_date.Date
+module G = Gencertlib
+module GP = Gencertlib.Pem
+module GV = Gencertlib.Validation
+open Api_errors
+open Rresult
+
+let ( let* ) = ( >>= )
+
+let is_tls_verification_enabled () =
+  Localdb.get_with_default Constants.tls_verification_enabled "false"
+  |> bool_of_string
+
+let is_self_signed cert =
+  R.return
+    X509.(
+      Distinguished_name.equal (Certificate.subject cert)
+        (Certificate.issuer cert))
+
+let pool_size ~__context = Db.Pool.get_all ~__context |> List.length
+
+let remove_cert ~__context cert =
+  info "Removing certificate %s" (Ref.string_of cert) ;
+  Db.Certificate.destroy ~__context ~self:cert
+
+(* the host cert is under a fixed name - so the new cert in the file
+   system already replaced the old one. We must not remove it from the
+   file system *)
+
+(** verify and install a certificate or return with an error *)
+let install ~__context ~host cert =
+  try
+    let host_pem = cert.GP.host_cert in
+    let* host_cert =
+      Cstruct.of_string host_pem
+      |> X509.Certificate.decode_pem
+      |> R.reword_error (fun (`Msg msg) ->
+             D.info {|Failed to decode certificate because "%s"|} msg ;
+             `Msg (server_certificate_invalid, []))
+    in
+    let* self_signed = is_self_signed host_cert in
+    match self_signed with
+    | true ->
+        let raw = X509.Certificate.encode_pem host_cert in
+        let pem = Cstruct.to_string raw in
+        (* pass own cert as root *)
+        let ref =
+          Xapi_host.add_certificate_to_db ~__context ~host host_cert [raw]
+        in
+        ( match pool_size ~__context with
+        | 1 ->
+            let uuid = Db.Certificate.get_uuid ~__context ~self:ref in
+            Certificates.(
+              pool_install CA_Certificate ~__context ~name:uuid ~cert:pem)
+        | _ ->
+            debug "Not installing updated self-signed host cert in pool"
+        ) ;
+        R.ok ()
+    | false -> (
+        let ca_cert_bundle_path = !Xapi_globs.stunnel_cert_path in
+        let pem_leaf = cert.GP.host_cert in
+        let pkcs8_private_key = cert.GP.private_key in
+        let pem_intermediates = String.concat "\n\n" cert.GP.other_certs in
+        let time = Ptime_clock.now () in
+        let tls_on = is_tls_verification_enabled () in
+        match
+          GV.validating_trust_anchors ~pem_leaf ~pkcs8_private_key
+            ~pem_intermediates ~time ~ca_cert_bundle_path
+        with
+        | Error _ as error_msg when tls_on ->
+            error "Host certificate is invalid (TLS verification is enabled)" ;
+            error_msg
+        | Ok (host_cert, roots) ->
+            let hashes = List.map X509.Certificate.encode_pem roots in
+            let _ref =
+              Xapi_host.add_certificate_to_db ~__context ~host host_cert hashes
+            in
+            R.ok ()
+        | Error (`Msg (msg, _)) ->
+            warn
+              "The host certificate is invalid (TLS verification is not \
+               enabled)" ;
+            warn "The host certificate is invalid: %s" msg ;
+            R.ok ()
+        | Error _ ->
+            warn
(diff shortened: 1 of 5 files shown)


ocaml/xapi/certificates_sync.ml
```diff
@@ -0,0 +1,144 @@
...
+let ( let* ) = ( >>= )
+
+let is_tls_verification_enabled () =
+  Localdb.get_with_default Constants.tls_verification_enabled "false"
+  |> bool_of_string
+
+let is_self_signed cert =
+  R.return
+    X509.(
+      Distinguished_name.equal (Certificate.subject cert)
```
edwintorok, line 21: This doesn't necessarily mean that it was actually signed with its own key, or that the signature is valid.
See https://github.com/mirleft/ocaml-x509/blob/master/lib/validation.ml#L322-L330 for a more comprehensive check.
Perhaps you could use  `is_self_signed + X509.Validation.valid_ca`

psafont: I'm pretty sure that `X509.Validation.valid_ca` will fail for the certificates that are generated with the xapi-generated certificates in all existing versions. Changing this would mean updating how the certs are generated and gating the old certificates.

ocaml/xapi/certificates_sync.ml
```diff
@@ -0,0 +1,144 @@
...
+        let pkcs8_private_key = cert.GP.private_key in
+        let pem_intermediates = String.concat "\n\n" cert.GP.other_certs in
+        let time = Ptime_clock.now () in
+        let tls_on = is_tls_verification_enabled () in
+        match
+          GV.validating_trust_anchors ~pem_leaf ~pkcs8_private_key
+            ~pem_intermediates ~time ~ca_cert_bundle_path
+        with
+        | Error _ as error_msg when tls_on ->
+            error "Host certificate is invalid but required by TLS" ;
```
lippirk, line 75:
```suggestion
            error "Host certificate is invalid (TLS verification is enabled)" ;
```

ocaml/xapi/certificates_sync.ml
```diff
@@ -0,0 +1,144 @@
...
+            error_msg
+        | Ok (host_cert, roots) ->
+            let hashes = List.map X509.Certificate.encode_pem roots in
+            let _ref =
+              Xapi_host.add_certificate_to_db ~__context ~host host_cert hashes
+            in
+            R.ok ()
+        | Error _ as error_msg ->
+            warn "The host certificate is invalid but TLS is off" ;
```
lippirk, line 84:
```suggestion
            warn "Host certificate is invalid (TLS verification is not enabled)" ;
```

lippirk: I hope TLS is not off!

ocaml/xapi/certificates_sync.ml
```diff
@@ -0,0 +1,144 @@
...
+module G = Gencertlib
+module GP = Gencertlib.Pem
+module GV = Gencertlib.Validation
+open Api_errors
+open Rresult
+
+let ( let* ) = ( >>= )
+
+let is_tls_verification_enabled () =
+  Localdb.get_with_default Constants.tls_verification_enabled "false"
```
lippirk, line 15: Maybe it is worth checking here that the pool state (Db.Pool.tls_verification_enabled) matches the local state (what you are using here) - perhaps we should block installation in this case? Though this may be overzealous

ocaml/xapi/certificates_sync.ml
```diff
@@ -0,0 +1,144 @@
...
+      let* unchanged = is_unchanged ~__context cert_ref cert in
+      if unchanged then (
+        info "Active server certificate for host %s is unchanged" host_uuid ;
+        Ok ()
+      ) else (
+        info "Server certificate for host %s changed - updating" host_uuid ;
+        let* () = install ~__context ~host cert in
+        remove_cert ~__context cert_ref
+      )
+  | cert_refs ->
```
lippirk, line 135: We could merge this match case with the previous one, which would save a little code:
```
| certs_ref::rest as cert_refs ->
  let* unchanged = rest = [] && is_unchanged ~__context cert_ref cert
```

ocaml/xapi/certificates_sync.ml
```diff
@@ -0,0 +1,144 @@
...
+      info "Server certificate for host %s changed - updating" host_uuid ;
+      let* () = install ~__context ~host cert in
+      let rec loop = function
+        | [] ->
+            Ok ()
+        | c :: cs ->
+            remove_cert ~__context c >>= fun () -> loop cs
+      in
+      loop cert_refs
```
psafont, line 144: This can be done with:
```ocaml
      List.iter (remove_cert  ~__context) cert_refs
```
By changing remove_cert to return `()`.

This makes it more obvious that the code works with the empty case as well, if the error message is massaged a bit. I'm thinking along the lines of `info "Server certificate in the database is not the same as in the filesystem, removing the existing %i certificates for host %s" (List.length cert_refs) host_uuid;`

ocaml/xapi/certificates_sync.ml
```diff
@@ -0,0 +1,144 @@
...
+open Rresult
+
+let ( let* ) = ( >>= )
+
+let is_tls_verification_enabled () =
+  Localdb.get_with_default Constants.tls_verification_enabled "false"
+  |> bool_of_string
+
+let is_self_signed cert =
+  R.return
```
psafont, line 19: I don't think it's necessary to wrap the boolean in a result here

ocaml/xapi/certificates_sync.ml
```diff
@@ -0,0 +1,144 @@
...
+            error_msg
+        | Ok (host_cert, roots) ->
+            let hashes = List.map X509.Certificate.encode_pem roots in
+            let _ref =
+              Xapi_host.add_certificate_to_db ~__context ~host host_cert hashes
+            in
+            R.ok ()
+        | Error _ as error_msg ->
+            warn "The host certificate is invalid but TLS is off" ;
+            error_msg
```
psafont, line 85: Won't returning the error as-is make the host enter maintenance mode, even if the certificates is not verified?

If a host was just updated and it fails here it will look like the host failed to upgrade. I think logging this and continuing with an `Ok ()`

Installing an invalid certificate with this new version will just not be possible since we enforce the validation of the certificate.

ocaml/xapi/certificates_sync.ml
lindig: I'd like to keep the cases separate: one is the normal case, the other is where we have more than one cert for the host. This should never happen. We install the new one and remove all old ones.

ocaml/xapi/certificates_sync.ml
lindig: There is a difference: My loop fails with an error when `remove_cert` returns an error whereas using `List.iter` would continue and gobble up the error. Not sure what the right approach is here.

ocaml/xapi/certificates_sync.ml
lindig: During boot we cant rely on the master connection - so I'd like to keep this local.

lindig:
I've addressed the comments and made the change @psafont suggested.

ocaml/xapi/certificates_sync.ml
lippirk: We should probably log that something unexpected happened then

lippirk (approved):

ocaml/xapi/certificates_sync.ml
psafont: Doesn't remove_cert return always `Ok ()`?

lindig: You are right - so this is purely hypothetical and could be simplified. Will do.

psafont (approved):

PR #4278, opened 2020-12-10
REQ-403 deprecate Pool.wlb_verify_cert flag, but ensure it still works

lippirk:
We need to thread the `wlb_verify_cert` flag to the stunnel client library, so that even if Pool.tls_verification_enabled is not set, wlb verification still works as expected

Needs to go in simultaneously with https://github.com/xapi-project/xen-api-libs-transitional/pull/89 - expect CI here to fail

### DIFF
--- a/ocaml/xapi/workload_balancing.ml
+++ b/ocaml/xapi/workload_balancing.ml
@@ -295,6 +295,7 @@ let wlb_request ~__context ~host ~port ~auth ~meth ~params ~handler ~enable_log
   let body = wlb_body meth params in
   let request = wlb_request host meth body auth in
   let pool = Helpers.get_pool ~__context in
+  let verify_cert = Db.Pool.get_wlb_verify_cert ~__context ~self:pool in
   let pool_other_config = Db.Pool.get_other_config ~__context ~self:pool in
   let timeout =
     try
@@ -310,8 +311,8 @@ let wlb_request ~__context ~host ~port ~auth ~meth ~params ~handler ~enable_log
          (filtered_headers (Http.Request.to_header_list request)))
       body ;
   try
-    Remote_requests.perform_request ~__context ~timeout ~host ~port ~request
-      ~handler ~enable_log
+    Remote_requests.perform_request ~__context ~timeout ~verify_cert ~host ~port
+      ~request ~handler ~enable_log
   with
   | Remote_requests.Timed_out ->
       raise_timeout timeout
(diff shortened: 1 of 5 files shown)


edwintorok (commented):
Will enabling pool level TLS verification automatically enable the WLB one?
Why should we be able to disable WLB TLS verification, can't we just always enable it (regardless of Pool level TLS verification)

lippirk:
> Will enabling pool level TLS verification automatically enable the WLB one?

I'm not sure if flipping the WLB verification flag when we enable pool wide verification is a good idea, because then `xe host-emergency-disable-tls-verification` wouldn't disable WLB verification. I think setting wlb_verify_cert to false actually makes more sense, since the idea is that you shouldn't rely on that flag anymore

> Why should we be able to disable WLB TLS verification, can't we just always enable it (regardless of Pool level TLS verification)

I don't think we should enable it by default because we'd need to set up `/etc/stunnel/xapi-stunnel-ca-bundle.pem` to contain the WLB cert

ocaml/xapi/workload_balancing.ml
```diff
@@ -295,6 +295,7 @@ let wlb_request ~__context ~host ~port ~auth ~meth ~params ~handler ~enable_log
   let body = wlb_body meth params in
   let request = wlb_request host meth body auth in
   let pool = Helpers.get_pool ~__context in
+  let verify_cert = Db.Pool.get_wlb_verify_cert ~__context ~self:pool in
```
psafont, line 298:
```suggestion
  let verify_cert =
    Db.Pool.(
      tls_verification_enabled ~__context ~self:pool
      || get_wlb_verify_cert ~__context ~self:pool)
  in
```
I think reverting the commit is not enough: the server certificates should be verified when Pool.tls_verification_enabled is enabled, regardless of the value of the get_wlb_verify_cert.

lippirk: This isn't necessary because we do this check in xen-api-libs-transitional already (https://github.com/xapi-project/xen-api-libs-transitional/pull/89/files#r540361707)

I think it's cleaner to keep checks against the global verification flag only inside the stunnel library

psafont:
> Will enabling pool level TLS verification automatically enable the WLB one?

Yes, validating the peer when establishing TLS connections is a requirement. This is why we wanted to remove this setting altogether.

> Why should we be able to disable WLB TLS verification, can't we just always enable it (regardless of Pool level TLS verification)

Once pool-level TLS verification is enabled it cannot be disabled, this includes connections made with WLB appliances.

I don't see a big reason to disallow users from enabling and disabling WLB certificate verification while pool-wide TLS verification, it doesn't add complexity given that we already have to support being on and off for the upgrade case.

psafont (approved):

lindig (approved):

PR #4280, opened 2020-12-16
REQ-403 pool conf fixes

lippirk:
This fixes an issue where the pool role's master address was out of sync between the pool.conf file and the pool role ref

### DIFF
--- a/ocaml/xapi-aux/pool_role.ml
+++ b/ocaml/xapi-aux/pool_role.ml
@@ -75,15 +75,22 @@ let read_pool_role () =
       Broken
     )
 
-let get_role () =
-  with_pool_role_lock (fun _ ->
-      match !role with
-      | Some x ->
-          x
-      | None ->
-          let r = read_pool_role () in
-          role := Some r ;
-          r)
+let get_role_without_lock () =
+  match !role with
+  | Some x ->
+      x
+  | None ->
+      let r = read_pool_role () in
+      role := Some r ;
+      r
+
+let get_role () = with_pool_role_lock get_role_without_lock
+
+let set_role_for_next_boot r =
+  with_pool_role_lock @@ fun () ->
+  let current = get_role_without_lock () in
+  Unixext.write_string_to_file !Constants.pool_config_file (string_of r) ;
+  current
 
 let is_master () = get_role () = Master
 
--- a/ocaml/xapi-aux/pool_role.ml
+++ b/ocaml/xapi-aux/pool_role.ml
@@ -92,6 +92,8 @@ let set_role_for_next_boot r =
   Unixext.write_string_to_file !Constants.pool_config_file (string_of r) ;
   current
 
+let unsafe_set_role_ref r = with_pool_role_lock (fun () -> role := Some r)
+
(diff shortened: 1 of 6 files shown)


ocaml/xapi/xapi_pool_transition.ml
```diff
@@ -24,12 +24,10 @@ open D
...
 (** Reset the role on disk, takes effect on next server restart only! *)
 let set_role r =
-  let open Pool_role in
-  let old_role = get_role () in
-  with_pool_role_lock (fun () ->
-      Unixext.write_string_to_file !Constants.pool_config_file (string_of r)) ;
-  Localdb.put Constants.this_node_just_became_master
-    (string_of_bool (old_role <> Master && r = Master))
+  let current_role = Pool_role.set_role_for_next_boot r in
+  (current_role <> Master && r = Master)
```
psafont, line 28: Shouldn't the conditions be switched?

```suggestion
  (current_role = Master && r <> Master)
```
using `new_role` instead of `r` would make this clearer.

psafont, line 28: can the names be tweaked to show the relation between them?
Something like `runtime_role` and `next_boot_role`.

ocaml/xapi/xapi.ml
```diff
@@ -577,6 +598,7 @@ let maybe_set_fqdn_in_pool_conf () =
           match Gencertlib.Lib.fqdn_of_hostname master_hostname with
           | Some master_fqdn ->
               if persisted_master_address <> master_fqdn then (
+                assert_can_ping master_fqdn ;
```
psafont, line 601: I think it would be preferable to investigate what race condition is being solved by waiting 5 seconds, and try to see what we can do about that other than waiting some seconds as it seems fragile.

lippirk: I don't think we have a race condition necessarily, we genuinely may not be able to reach whatever FQDN we have calculated the master to be at (in which case it would be a bad idea to go ahead and change the pool config).

Part of me would like to do the pool config changes only when TLS verification is enabled, since at that point we can expect all the hosts to be online, and we can give the user an error message if the FQDNs are wrong. Admittedly we could propagate an error message anyway using the host health check we added: https://github.com/xapi-project/xen-api/commit/fb61b55e51f9d7a0199b588d1df432321ca3a8a5

ocaml/xapi-aux/pool_role.ml
```diff
@@ -75,15 +75,24 @@ let read_pool_role () =
...
+      x
+  | None ->
+      let r = read_pool_role () in
+      role := Some r ;
+      r
+
+let get_role () = with_pool_role_lock get_role_no_lock
+
+let set_role_for_next_boot r =
+  with_pool_role_lock (fun () ->
```
lindig, line 90: It's a matter of preference, but you can do this and avoid the long `(..)`
```
with_pool_role_lock @@ fun () ->
```

lindig (approved):

psafont (approved):

PR #4284, opened 2020-12-18
REQ-403 alert fixups

lippirk:


ocaml/xapi/xapi_session.ml
```diff
@@ -251,7 +251,7 @@ let _record_login_failure ~__context ~now ~uname ~originator ~record f =
       on_fail e
 
 let record_login_failure ~__context ~uname ~originator ~record f =
-  let now = Date.localtime () in
+  let now = Unix.time () |> Date.of_float in
```
lippirk, line 254: This improves the XenCenter usability in the case where you have pools in several timezones

psafont (approved):

ocaml/xapi/xapi_session.ml
kc284: Beyond what XenCenter expects, there are already other alerts issued by xapi that use UTC, so it's easier overall to keep the same behaviour for newly added alerts.

kc284 (approved):

PR #4285, opened 2021-01-04
CA-343646: generate certificate alerts

psafont:
The lastest change prevented creating new alerts created when there were
no other certificate alerts present.

Having no alerts now doesn't dismiss any existing alerts, as stated by the
comment.

### DIFF
--- a/ocaml/alerts/certificate/certificate_check.ml
+++ b/ocaml/alerts/certificate/certificate_check.ml
@@ -41,12 +41,6 @@ let generate_alert epoch (host, expiry) =
     (host, Some (message, alert))
 
 let execute rpc session existing_messages (host, alert) =
-  let host_uuid = XenAPI.Host.get_uuid rpc session host in
-  let messages_in_host =
-    List.filter
-      (fun (_, record) -> record.API.message_obj_uuid = host_uuid)
-      existing_messages
-  in
   (* CA-342551: messages need to be deleted if the pending alert regard the
      same host and has newer, updated information.
      If the pending alert has the same metadata as the existing host message
@@ -56,6 +50,12 @@ let execute rpc session existing_messages (host, alert) =
      they are not destroyed since no alert is automatically dismissed. *)
   match alert with
   | Some (message, (alert, priority)) ->
+      let host_uuid = XenAPI.Host.get_uuid rpc session host in
+      let messages_in_host =
+        List.filter
+          (fun (_, record) -> record.API.message_obj_uuid = host_uuid)
+          existing_messages
+      in
       let is_outdated (ref, record) =
         record.API.message_body <> message
         || record.API.message_name <> alert
--- a/ocaml/alerts/certificate/certificate_check.ml
+++ b/ocaml/alerts/certificate/certificate_check.ml
@@ -47,15 +47,6 @@ let execute rpc session existing_messages (host, alert) =
       (fun (_, record) -> record.API.message_obj_uuid = host_uuid)
       existing_messages
   in
-  let is_outdated (ref, record) =
-    match alert with
-    | None ->
-        true
-    | Some (body, (name, priority)) ->
-        record.API.message_body <> body
-        || record.API.message_name <> name
-        || record.API.message_priority <> priority
-  in
   (* CA-342551: messages need to be deleted if the pending alert regard the
      same host and has newer, updated information.
      If the pending alert has the same metadata as the existing host message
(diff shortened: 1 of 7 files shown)


ocaml/alerts/certificate/certificate_check.ml
```diff
@@ -41,39 +41,29 @@ let generate_alert epoch (host, expiry) =
...
-        outdated_messages ;
-      ignore
-        (XenAPI.Message.create rpc session alert priority `Host host_uuid
-           message)
-  | _ ->
+  match alert with
+  | Some (message, (alert, priority)) ->
+      let host_uuid = XenAPI.Host.get_uuid rpc session host in
+      let (_ : [> `message] Client.Id.t API.Ref.t) =
+        XenAPI.Message.create rpc session alert priority `Host host_uuid message
```
edwintorok, line 55: Won't this now always create a new message and thus change the message UUID contrary to what the comment states?

edwintorok: One way to fix this would be perhaps to modify the previous commit to:
`match (alert, replacement_is_pending)` to `match (alert, replacement_is_pending || previous_messages = [])`. Also `replacement_is_pending` can be written as `outdated_messages <> []`.

edwintorok: I'd put all the existing_messages into a Set, make any changes needed to the set (e.g. add new message), then diff the old and the new sets and figure out what messages you have to create or destroy (there are set subtraction functions). And for building the Set I'd ignore the message UUID (thus ensuring that the message's UUID won't change: it'll retain the UUID of the  first message of its kind, at least on OCaml >=4.03 according to the Set's docs).

psafont: You're right this doesn't work to avoid replacing messages that are still valid. Using sets it's a bit difficult since there's a mismatch between existing messages and the message that may be emitted: I need the reference to be able to delete messages, but I can't have the reference of the newly created message until it's created. I'll think more about this.

psafont: I've replaced the filter with a partition: If there are any elements that are still current no new alerts get emmitted.

ocaml/alerts/certificate/certificate_check.ml
```diff
@@ -41,39 +41,38 @@ let generate_alert epoch (host, expiry) =
...
+
       List.iter
         (fun (self, _) -> XenAPI.Message.destroy rpc session self)
-        outdated_messages ;
-      ignore
-        (XenAPI.Message.create rpc session alert priority `Host host_uuid
-           message)
-  | _ ->
+        outdated ;
+      if current = [] then
```
lindig, line 69: Why do we need to do something when `current = []`? Should this be `<>`?

psafont: `current` being empty means that the new alert is different from any existing alerts so it should be emitted.

This may happen in two cases:
1. there are no certificate alerts in the system about the host: the new alert is generated
2. There is at least one alert in the system but these have information that is different from the new alert: they are deleted (they are present in `outdated`) and the new alert is generated.

edwintorok (approved):
LGTM, this could have some unit tests as discussed, those can come in a separate PR.

ocaml/alerts/certificate/certificate_check.ml
lindig: So the message we need either exists already (in current), or we need to create it. It seems to me that `current` should have at most one member because to be in current all messages have to agree on message, alert, and priority. Not sure it's worth acting on this if more than one message is here.

ocaml/alerts/certificate/certificate_check.ml
```diff
@@ -47,33 +47,32 @@ let execute rpc session existing_messages (host, alert) =
...
-  in
   (* CA-342551: messages need to be deleted if the pending alert regard the
      same host and has newer, updated information.
      If the pending alert has the same metadata as the existing host message
      it must not be emmited and the existing message must be retained, this
      prevents changing the message UUID.
      In the case there are alerts regarding the host but no alert is pending
      they are not destroyed since no alert is automatically dismissed. *)
-  let outdated_messages = List.filter is_outdated messages_in_host in
-  let replacement_is_pending = List.exists is_outdated messages_in_host in
```
lippirk, line 67: Out of interest would simply

```
let replacement_is_pending = List.exists is_outdated messages_in_host || messages_in_host = []
```

have done the trick?

lippirk (approved):

lindig (approved):

ocaml/alerts/certificate/certificate_check.ml
psafont: It's not exactly the same.

`current` being empty can only happen when `messages_in_host` is empty or `outdated_messages` contains all of its elements.

The equivalent condition would be:
```
let replacement_is_pending = List.for_all is_outdated messages_in_host in
```
But this would mean the removal of elements is behind the flag, which is not the case with the current implementation.

It may not make a difference since only one condition per host should be present. (every time an alert is created the rest are deleted)

psafont:
Tests are failing because they expect a new alert being generated whenever the script is run. This conflicts with the idea of not replacing alerts whenever possible. I'll change the tests so the alerts are only changed when the message changes.

I don't think it makes sense to merge this until the tests are changed as they'll keep failing.

PR #4286, opened 2021-01-05
REQ-403 fix first boot failure

lippirk:
Before running '"Update host certificate if changed"' startup task, we need gencert to have created xapi-ssl.pem, otherwise the host will be placed in emergency mode and the first boot scripts will fail.

I've included some refactorings alongside this fix

### DIFF
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -1634,6 +1634,8 @@ let try_internal_async ~__context (marshaller : Rpc.t -> 'b)
 module SystemdService (Name : sig
   val name : string
 end) : sig
+  val is_active : unit -> bool
+
   val ensure_enabled_and_restart : unit -> unit
 end = struct
   let name = Name.name
@@ -1650,6 +1652,8 @@ end = struct
       )
     in
     exit_code = 0
+
+  let is_active () = is_cmd "is-active"
 
   let is_enabled () = is_cmd "is-enabled"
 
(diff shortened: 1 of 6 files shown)


ocaml/xapi/helpers.ml
```diff
@@ -1631,13 +1631,46 @@ let try_internal_async ~__context (marshaller : Rpc.t -> 'b)
...
+  val ensure_enabled_and_restart : unit -> unit
+end = struct
+  let name = Name.name
+
+  let systemctl cmd = call_script !Xapi_globs.systemctl [cmd; name]
+
+  let systemctl_ cmd = systemctl cmd |> ignore
+
+  let is_cmd cmd =
+    let exit_code =
```
lindig, line 1648: Is it possible to implement it like this? I believe it would be clearer but I'm not sure about pattern matching in the `exception` case.
```
let exit_code = match systemctl_ cmd with
| () -> 0
| exception Forkhelpers.Spawn_internal_error (_, _, Unix.WEXITED n) -> n
| exception e -> raise e
```

ocaml/xapi/xapi.ml
```diff
@@ -516,6 +516,21 @@ let server_run_in_emergency_mode () =
   wait_to_die () ; exit 0
 
 let update_certificates ~__context () =
+  let module GencertService = Helpers.SystemdService (struct
```
lindig, line 519: Is this not a problem that should be solved by service dependencies?

psafont: I've asked Ben the same question elsewhere, it can't be done unless a third service sets MANAGEMENT_INTERFACE in xensource-inventory. 

> I'm wondering if the dependency could be done with the systemd unit files instead of a waiting loop in xapi, was there something that made you avoid that solution?

> that won't work (at the moment, maybe there is a better way) because xapi sets MANAGEMENT_INTERFACE in xensource-inventory, which gencert needs

PR #4287, opened 2021-01-05
CP-35599 add new interface for certs

lindig:
This introduces module Certs which provides an interface between the
on-disk representation of certificates and their internal, more detailed
structure. This shall help to maintain the consistency between the two
as we want to make it the only interface for all of Xapi to use.

### DIFF
--- a/ocaml/xapi/certs.ml
+++ b/ocaml/xapi/certs.ml
@@ -1,0 +1,272 @@
+(*
+ * Copyright (C) Citrix Systems Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+open Rresult
+module Pem = Gencertlib.Pem
+
+type uuid = Uuidm.t
+
+type paths = {host_cert_path: string; ca_certs_dir: string; ca_crls_dir: string}
+
+type host = {
+    private_key: X509.Private_key.t
+  ; cert: X509.Certificate.t
+  ; intermediates: X509.Certificate.t list
+}
+
+type ca = {cert: X509.Certificate.t; uuid: uuid; mutable deleted: bool}
+
+type crl = {name: string; cert: X509.Certificate.t; mutable deleted: bool}
+
+type t = {
+    host: host option
+  ; mutable ca: ca list
+  ; mutable crl: crl list
+  ; paths: paths
+}
+
+let ( let* ) = R.bind
+
+let defer f = Fun.protect ~finally:f
+
+let default_paths () =
+  {
+    host_cert_path= !Xapi_globs.server_cert_path
+  ; ca_certs_dir= "/etc/stunnel/certs"
+  ; ca_crls_dir= Stunnel.crl_path
+  }
+
+(* local, private, mutable state *)
+let t : t option ref = ref None
+
+(* calls in this API are protected by this mutex *)
+let mx = Mutex.create ()
+
+let with_mutex f =
+  Mutex.lock mx ;
+  defer (fun () -> Mutex.unlock mx) f
+
+let get_unsafe () =
+  match !t with
+  | Some t ->
+      R.ok t
+  | None ->
+      R.error_msgf "Certificate store has not been initialised %s" __LOC__
+
+let has_extension ext path = Filename.extension path = ext
+
+let filter_dir accept path =
+  Sys.readdir path
+  |> Array.to_list
+  |> List.filter accept
+  |> List.map (Filename.concat path)
+
+let decode_key pem = Cstruct.of_string pem |> X509.Private_key.decode_pem
+
+let decode_cert pem = Cstruct.of_string pem |> X509.Certificate.decode_pem
+
+let decode_multi pem =
+  Cstruct.of_string pem |> X509.Certificate.decode_pem_multiple
+
+let read_host path =
+  match Sys.file_exists path with
+  | true ->
+      let* pems =
+        Pem.parse_file_using Pem.host_pem path |> function
+        | Error msg ->
+            R.error_msg msg
+        | Ok _ as ok ->
+            ok
+      in
+      let* private_key = decode_key pems.Pem.private_key in
+      let* cert = decode_cert pems.Pem.host_cert in
+      let* intermediates =
+        decode_multi (String.concat "\n" pems.Pem.other_certs)
+      in
+      Ok (Some {private_key; cert; intermediates})
+  | false ->
+      Ok None
+
+let read_crl path =
+  let name = Filename.basename path in
+  let* pem =
+    Pem.parse_file_using Pem.cert path |> function
+    | Error msg ->
+        R.error_msg msg
+    | Ok _ as ok ->
+        ok
+  in
+  let* cert = decode_cert pem in
+  Ok {cert; name; deleted= false}
+
+let read_ca path =
+  let name = Filename.basename path |> Filename.chop_extension in
+  let* uuid =
+    match Uuidm.of_string name with
+    | Some uuid ->
+        Ok uuid
+    | None ->
+        R.error_msgf "%s is not a uuid filename" path
+  in
+  let* pem =
+    Pem.parse_file_using Pem.cert path |> function
+    | Error msg ->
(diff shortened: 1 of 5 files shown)


ocaml/xapi/certs.ml
```diff
@@ -0,0 +1,414 @@
...
+
+  type ca = {cert: X509.Certificate.t; uuid: uuid; mutable deleted: bool}
+
+  type cr = {name: string; cert: X509.Certificate.t; mutable deleted: bool}
+
+  (* [path] value to be used normally but other paths could be used for
+     testing *)
+  val default_paths : unit -> paths
+
+  (* initialise this module and read satte from disk *)
```
psafont, line 137: I suspect `satte` is a German word :)

ocaml/xapi/certs.ml
```diff
@@ -0,0 +1,414 @@
...
+  val init : paths -> (unit, [> R.msg]) result
+
+  (* write back stack in memory to disk *)
+  val sync : unit -> (unit, [> R.msg]) result
+
+  (* remove CA cert *)
+  val rm_ca : uuid -> unit -> (unit, [> R.msg]) result
+
+  (* remove CR cert *)
```
psafont, line 146: CRL is much more widespread than CR, could we use that instead?

ocaml/xapi/certs.ml
```diff
@@ -0,0 +1,414 @@
...
+  val rm_cr : string -> unit -> (unit, [> R.msg]) result
+
+  (* access CR cert *)
+  val get_cr : unit -> ((string * X509.Certificate.t) list, [> R.msg]) result
+
+  (* access CA cert *)
+  val get_ca : unit -> ((uuid * X509.Certificate.t) list, [> R.msg]) result
```
psafont, line 153: I'm wondering whether we would want a `CertificateMap` instead of a list.

ocaml/xapi/certs.ml
```diff
@@ -0,0 +1,414 @@
...
+    ; intermediares: X509.Certificate.t list
+  }
+
+  type ca = {cert: X509.Certificate.t; uuid: uuid; mutable deleted: bool}
+
+  type cr = {name: string; cert: X509.Certificate.t; mutable deleted: bool}
+
+  type t = {
+      host: host option
+    ; mutable ca: ca list
```
psafont, line 183: I'm wondering whether we would want a `CertificateMap` instead of a list.

psafont (approved):

ocaml/xapi/certs.ml
```diff
@@ -0,0 +1,414 @@
...
+
+open Rresult
+
+let ( let* ) = R.bind
+
+let defer f = Fun.protect ~finally:f
+
+(* Pem provides parsing of *.pem files. The returned strings still need
+to be decoded *)
+module Pem : sig
```
lippirk, line 23: This module looks identical to ocaml/gencert/pem.ml, is there a reason you're not using that?

ocaml/xapi/certs.ml
```diff
@@ -0,0 +1,414 @@
...
+  type paths = {
+      host_cert_path: string
+    ; ca_certs_dir: string
+    ; ca_crls_dir: string
+  }
+
+  type host = {
+      private_key: X509.Private_key.t
+    ; cert: X509.Certificate.t
+    ; intermediares: X509.Certificate.t list
```
lippirk, line 126: Do you mean intermediates here? or possibly intermediaries

ocaml/xapi/certs.ml
```diff
@@ -0,0 +1,414 @@
...
+      host: host option
+    ; mutable ca: ca list
+    ; mutable cr: cr list
+    ; paths: paths
+  }
+
+  let default_paths () =
+    {
+      host_cert_path= !Xapi_globs.server_cert_path
+    ; ca_certs_dir= "/etc/stunnel/certs"
```
lippirk, line 191: I think we're using a CA bundle rather than a CA directory (the path to which is `Stunnel.certificates_bundle_path`)

ocaml/xapi/certs.ml
lindig: Indeed, there is a lot of overlap. The gencert/pem.ml only reads the host *.pem but does not read CA certs from stunnel. I could change gencert/pem.ml to a wider interface and use that. The other reason is that I tried to develop this code independent of existing code.

ocaml/xapi/certs.ml
lindig: I believe we need both and need a discussion how to handle that. But it should not be difficult to read and generate bundles as well.

ocaml/xapi/certs.ml
```diff
@@ -0,0 +1,414 @@
...
+  let t : t option ref = ref None
+
+  (* calls in this API are protected by this mutex *)
+  let mx = Mutex.create ()
+
+  let with_mutex f =
+    Mutex.lock mx ;
+    defer (fun () -> Mutex.unlock mx) f
+
+  let get () =
```
lippirk, line 205: Can we call this `get_unsafe`/`get_no_lock`? So that it's obvious it ought to be used inside `with_mutex`, and that we're not going to get a deadlock

ocaml/xapi/certs.ml
lindig: I expect the list to be short. A hash would also provide the mutable data structure that we need here.

ocaml/xapi/certs.ml
psafont: In practice both are needed. We require the names to be able to identify individual CA certificates. The bundle is needed for stunnel.

ocaml/xapi/certs.ml
```diff
@@ -0,0 +1,414 @@
...
+  let sync_ca dir (ca : ca) =
+    let name = Printf.sprintf "%s.pem" (Uuidm.to_string ca.uuid) in
+    let path = Filename.concat dir name in
+    let rm path = Sys.remove path in
+    match ca.deleted with
+    | true ->
+        R.trap_exn rm path |> R.error_exn_trap_to_msg
+    | false ->
+        let pem = X509.Certificate.encode_pem ca.cert |> Cstruct.to_string in
+        write_to path pem
```
lippirk, line 321: I think we need to sync with the bundle here, which is more complicated :(

ocaml/xapi/certs.ml
lindig: Only as in "die satte Kuh" (the well-fed cow). I meant to use "state".

ocaml/xapi/certs.ml
lippirk: Oh I see now. Is the idea that we write the CA certs to individual files, and then concatenate them all to the bundle file? (Though I don't see the code for that)

ocaml/xapi/certs.ml
lippirk: Yes I see what's going on now. I guess we need to decide when to generate new bundles

psafont: Right now the bundle is refreshed with a script that runs everytime CAs are synchronized in the host.

ocaml/xapi/certs.ml
```diff
@@ -0,0 +1,414 @@
...
+  val init : paths -> (unit, [> R.msg]) result
+
+  (* write back stack in memory to disk *)
+  val sync : unit -> (unit, [> R.msg]) result
+
+  (* remove CA cert *)
+  val rm_ca : uuid -> unit -> (unit, [> R.msg]) result
+
+  (* remove CR cert *)
+  val rm_cr : string -> unit -> (unit, [> R.msg]) result
```
psafont, line 147: I think we talked about exposing only atomic operations, meaning sync would be used as part of the exposed operations and not be exposed itself.

I think it would be a safer interface to use, and it could be done in subsequent PRs.

ocaml/xapi/certs.ml
```diff
@@ -0,0 +1,334 @@
...
+open Rresult
+
+(* module [Sync] provides an interface to certificates stored on
+disk. Certificates are read upon [init], potentially manipulated, and
+written back on [sync]. No other code should write these files. Writing
+back on [sync] provides a primitive form of transactions: in case of a
+fatal error, we could avoid writing a bad state to disk. This is
+currently not actively used, though *)
+
+module Sync : sig
```
lippirk, line 24: I think this sig should be in an mli file really, because whilst the Sync module is restricted, functions added to this file in the future will not be

lippirk (approved):

ocaml/xapi/certs.ml
lindig: I can do that. I was keeping it here while things are in flux.

ocaml/xapi/certs.ml
lindig: Right now, `sync` is never called implicitly. So unless we change the semantics of update functions or don't want to sync back to disk, it would be required.

lindig:
I will squash commits and un-draft this.

PR #4297, opened 2021-01-12
CP-35372: Repository: datamodel

minglumlu:
This commit introduces datamodel related changes for repository function.

 The repository function makes XAPI be able to update dom0 of Xen hosts from
 remote YUM repository.

### DIFF
--- a/ocaml/idl/datamodel_errors.ml
+++ b/ocaml/idl/datamodel_errors.ml
@@ -1198,8 +1198,25 @@ let _ =
     ~doc:"No other cluster host was reachable when joining" ();
 
   error Api_errors.xen_incompatible []
-    ~doc:"The current version of Xen or its control libraries is incompatible with the Toolstack." ()
+    ~doc:"The current version of Xen or its control libraries is incompatible with the Toolstack." ();
 
+  (* repository and updates errors *)
+  error Api_errors.set_repository_in_progress []
+    ~doc:"The operation could not be performed because a repository setup is in progress." ();
+  error Api_errors.invalid_repository_name ["name"]
+    ~doc:"The name of the repository is invalid." ();
+  error Api_errors.invalid_base_url ["url"]
+    ~doc:"The base url in the repository is invalid." ();
+  error Api_errors.repository_already_exists []
+    ~doc:"The repository already exists." ();
+  error Api_errors.repository_is_in_use []
+    ~doc:"The repository is in use." ();
+  error Api_errors.reposync_in_progress []
+    ~doc:"The pool is syncing with the enabled remote YUM repository." ();
+  error Api_errors.local_pool_repo_cleanup_failed []
+    ~doc:"Failed to clean up local pool repository on master." ();
+  error Api_errors.no_repository_enabled []
+    ~doc:"There is no repository being enabled." ()
 
 
 let _ =
--- a/ocaml/idl/datamodel_errors.ml
+++ b/ocaml/idl/datamodel_errors.ml
@@ -1203,18 +1203,16 @@ let _ =
   (* repository and updates errors *)
   error Api_errors.set_repository_in_progress []
     ~doc:"The operation could not be performed because a repository setup is in progress." ();
-  error Api_errors.invalid_repository_name ["name"]
-    ~doc:"The name of the repository is invalid." ();
   error Api_errors.invalid_base_url ["url"]
     ~doc:"The base url in the repository is invalid." ();
-  error Api_errors.repository_already_exists []
+  error Api_errors.repository_already_exists ["ref"]
     ~doc:"The repository already exists." ();
   error Api_errors.repository_is_in_use []
     ~doc:"The repository is in use." ();
   error Api_errors.reposync_in_progress []
     ~doc:"The pool is syncing with the enabled remote YUM repository." ();
-  error Api_errors.local_pool_repo_cleanup_failed []
-    ~doc:"Failed to clean up local pool repository on master." ();
+  error Api_errors.repository_cleanup_failed []
+    ~doc:"Failed to clean up local repository on master." ();
   error Api_errors.no_repository_enabled []
     ~doc:"There is no repository being enabled." ()
 
--- a/ocaml/idl/datamodel_pool.ml
+++ b/ocaml/idl/datamodel_pool.ml
@@ -8,6 +8,7 @@ open Datamodel_types
             "ha_disable", "Indicates this pool is in the process of disabling HA";
             "cluster_create", "Indicates this pool is in the process of creating a cluster";
             "designate_new_master", "Indicates this pool is in the process of changing master";
+            "set_repository", "Indicates this pool is in the process of configuring repository";
           ])
 
   let enable_ha = call
@@ -647,6 +648,17 @@ open Datamodel_types
     ~allowed_roles:_R_POOL_ADMIN
     ()
 
+  let set_repository = call
+      ~name:"set_repository"
+      ~in_product_since:rel_next
+      ~doc:"Set the enabled repository for updates"
+      ~params:[
+        Ref _pool, "self", "The pool";
+        Ref _repository, "value", "The repository to be enabled"
+      ]
+      ~allowed_roles:_R_POOL_ADMIN
+      ()
+
   (** A pool class *)
   let t =
     create_obj
@@ -721,6 +733,7 @@ open Datamodel_types
         ; add_to_guest_agent_config
         ; remove_from_guest_agent_config
         ; rotate_secret
+        ; set_repository
         ]
       ~contents:
         ([uid ~in_oss_since:None _pool] @
@@ -764,5 +777,6 @@ open Datamodel_types
          ; field ~in_product_since:rel_inverness ~qualifier:DynamicRO ~ty:Bool ~default_value:(Some (VBool false)) "igmp_snooping_enabled" "true if IGMP snooping is enabled in the pool, false otherwise."
          ; field ~in_product_since:rel_quebec ~qualifier:RW ~ty:String ~default_value:(Some (VString "")) "uefi_certificates" "The UEFI certificates allowing Secure Boot"
          ; field ~in_product_since:rel_stockholm_psr ~qualifier:RW ~ty:Bool ~default_value:(Some (VBool false)) "is_psr_pending" "True if either a PSR is running or we are waiting for a PSR to be re-run"
+         ; field ~in_product_since:rel_next ~qualifier:DynamicRO ~ty:(Ref _repository) ~default_value:(Some (VRef null_ref)) "repository" "The enabled repository for updates"
          ])
       ()
--- a/ocaml/idl/datamodel_repository.ml
+++ b/ocaml/idl/datamodel_repository.ml
@@ -1,0 +1,83 @@
+(*
+ * Copyright (C) Citrix Systems Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+open Datamodel_types
+open Datamodel_common
+open Datamodel_roles
+
+let lifecycle = [Published, rel_next, ""]
+
+let introduce = call
+    ~name:"introduce"
+    ~in_oss_since:None
+    ~lifecycle
+    ~doc:"Add the configuration for a new repository"
+    ~params:[
+      String, "name_label", "The name of the repository";
+      String, "name_description", "The description of the repository";
+      String, "binary_url", "Base URL of binary packages in this repository";
+      String, "source_url", "Base URL of source packages in this repository";
+    ]
+    ~result:(Ref _repository, "The ref of the created repository record.")
+    ~allowed_roles:_R_POOL_OP
+    ()
+
+let forget = call
+    ~name:"forget"
+    ~in_oss_since:None
+    ~lifecycle
+    ~doc:"Remove the repository record from the database"
+    ~params:[Ref _repository, "self", "The repository to be removed from the database"]
+    ~allowed_roles:_R_POOL_OP
+    ()
+
+let t =
+  create_obj
+    ~name: _repository
+    ~descr:"Repository for updates"
+    ~doccomments:[]
+    ~gen_constructor_destructor:false
+    ~gen_events:true
+    ~in_db:true
+    ~lifecycle
+    ~persist:PersistEverything
+    ~in_oss_since:None
+    ~messages_default_allowed_roles:_R_POOL_OP
+    ~messages: [introduce; forget]
+
+    ~contents:
+      [ uid       _repository ~lifecycle;
+        namespace ~name:"name" ~contents:(names ~writer_roles:_R_POOL_OP None RW) ();
+        field     ~qualifier:StaticRO ~lifecycle
+          ~ty:String
+          ~default_value:(Some (VString ""))
+          "binary_url"
+          "Base URL of binary packages in this repository";
+        field     ~qualifier:StaticRO ~lifecycle
+          ~ty:String
+          ~default_value:(Some (VString ""))
+          "source_url"
+          "Base URL of source packages in this repository";
+        field     ~qualifier:DynamicRO ~lifecycle
+          ~ty:String
+          ~default_value:(Some (VString ""))
+          "hash"
+          "SHA256 checksum of latest updateinfo.xml.gz in this repository";
+        field     ~qualifier:DynamicRO ~lifecycle
+          ~ty:Bool
+          ~default_value:(Some (VBool false))
+          "up_to_date"
+          "True if all hosts in pool is up to date with this repository";
+      ]
+    ()
--- a/ocaml/xapi-cli-server/cli_frontend.ml
+++ b/ocaml/xapi-cli-server/cli_frontend.ml
@@ -3093,6 +3093,22 @@ let rec cmdtable_data : (string * cmd_spec) list =
       ; implementation= No_fd Cli_operations.Cluster_host.force_destroy
       ; flags= []
       } )
+  ; ( "repository-introduce"
+    , {
+        reqd= ["name-label"; "binary-url"; "source-url"]
+      ; optn= ["name-description"]
+      ; help= "Add the configuration for a new repository."
+      ; implementation= No_fd Cli_operations.Repository.introduce
+      ; flags= []
+      } )
+  ; ( "repository-forget"
+    , {
+        reqd= ["uuid"]
+      ; optn= []
+      ; help= "Remove the repository record from the database."
+      ; implementation= No_fd Cli_operations.Repository.forget
+      ; flags= []
+      } )
   ]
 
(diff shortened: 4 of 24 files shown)


ocaml/xapi/repository.ml
```diff
@@ -0,0 +1,975 @@
...
+      in
+      let resolved_path =
+        let len = String.length Constants.get_repository_uri in
+        begin match String.sub_to_end req.Request.uri len with
+          | untrusted_path ->
+            untrusted_path
+            |> Filename.concat repo_name
+            |> Filename.concat !Xapi_globs.local_pool_repo_dir
+            |> Uri.pct_decode
+            |> Xapi_stdext_unix.Unixext.resolve_dot_and_dotdot
```
edwintorok, line 951: It would be good to abstract this "resolve a path safely relative to a root" into a function of its own.
There is already nearly identical code to this elsewhere in xapi, see de2c5887ae4532a9790cda904f78bb0672a16e41 for a security issue we had to fix in that. The code here looks ok, but if we duplicate this kind of security-checking code we risk them getting out-of-date with eachother (e.g. a security fix for one might miss the other) or introducing subtle bugs.

edwintorok: Probably best to wrap Fileserver.response_file and have a whitelist of dirs that it is allowed to use.

ocaml/xapi/repository.ml
```diff
@@ -0,0 +1,975 @@
...
+  Db.Repository.set_hash ~__context ~self ~value:hash;
+  hash
+
+let update_of_json j =
+  let open Yojson.Basic.Util in
+  let u = {
+    name = (member "name" j |> to_string);
+    arch = (member "arch" j |> to_string);
+    new_ver_rel = (member "newVerRel" j |> to_string);
+    old_ver_rel = (try Some (member "oldVerRel" j |> to_string) with _ -> None);
```
psafont, line 868: I don't think the versions should just be string.

I think this is where they should be parsed into `{version, release}`

For example:
```
let ver_rel_of_string str = match Astring.cut ~sep:"-" with
| Some (v, r) when v <> "" && r <> "" ->
  {version=v; release=r}
| None ->
  None


let new_ver_rel = (member "newVerRel" j |> to_string |> ver_rel_of_string)

File: ocaml/xapi/repository.ml [hunk 3]
```diff
@@ -112,8 +112,23 @@ type updateinfo = {
   ; guidance_applicabilities: applicability list
   ; spec_info: string
   ; url: string
+  ; ty: string
```
Zhengchai, line 115: could the ty be more meaningful name?

ocaml/idl/datamodel_pool.ml
```diff
@@ -8,6 +8,9 @@ open Datamodel_types
             "ha_disable", "Indicates this pool is in the process of disabling HA";
             "cluster_create", "Indicates this pool is in the process of creating a cluster";
             "designate_new_master", "Indicates this pool is in the process of changing master";
+            "set_repository", "Indicates this pool is in the process of setting repository";
+            "updates_sync", "Indicates this pool is in the process of syncing updates";
+            "updates_list", "Indicates this pool is in the process of listing updates";
```
lindig, line 13: * `of setting up the repository` or `configuring the repository`
* `listening for updates`

ocaml/idl/datamodel_errors.ml
```diff
@@ -1198,7 +1198,41 @@ let _ =
     ~doc:"No other cluster host was reachable when joining" ();
 
   error Api_errors.xen_incompatible []
-    ~doc:"The current version of Xen or its control libraries is incompatible with the Toolstack." ()
+    ~doc:"The current version of Xen or its control libraries is incompatible with the Toolstack." ();
+
+  (* repository and updates errors *)
+  error Api_errors.set_repository_in_progress []
+    ~doc:"The operation could not be performed because setting repository is in progress." ();
```
lindig, line 1205: `because setting up of the repository` or `because a repository setup is in progress`

ocaml/idl/datamodel_errors.ml
```diff
@@ -1198,7 +1198,41 @@ let _ =
...
+  error Api_errors.updates_sync_in_progress []
+    ~doc:"The operation could not be performed because updates syncing is in progress." ();
+  error Api_errors.updates_get_in_progress []
+    ~doc:"The operation could not be performed because updates getting is in progress." ();
+  error Api_errors.invalid_updateinfo_xml []
```
lindig, line 1210: `because getting updates is in progress`

ocaml/idl/datamodel_errors.ml
```diff
@@ -1198,7 +1198,41 @@ let _ =
...
+    ~doc:"The content of updateinfo.xml is invalid." ();
+  error Api_errors.invalid_repository_name ["name"]
+    ~doc:"The name of repository is invalid." ();
```
lindig, line 1213: `The name of the repository is invalid.`

ocaml/idl/datamodel_repository.ml
```diff
@@ -0,0 +1,82 @@
...
+    ]
+    ~result:(Ref _repository, "The ref of the created repository record.")
+    ~allowed_roles:_R_POOL_OP
+    ()
+
+let forget = call
+    ~name:"forget"
+    ~in_oss_since:None
+    ~in_product_since:rel_next
+    ~doc:"Remove a repository record from the database"
```
lindig, line 39: `Remove the repository from the database` - the repo in the argument will be removed so it is better to use the definite article (the)

ocaml/idl/datamodel_repository.ml
```diff
@@ -0,0 +1,82 @@
...
+    ~params:[Ref _repository, "self", "The repository to forget about"]
```
lindig, line 40: "forget about" is a bit too casual. "The repository to be removed from the database."

ocaml/xapi-cli-server/cli_frontend.ml
```diff
@@ -3093,6 +3101,22 @@ let rec cmdtable_data : (string * cmd_spec) list =
...
+      ; optn= []
+      ; help= "Create a repository object representing a YUM repository."
+      ; implementation= No_fd Cli_operations.Repository.introduce
+      ; flags= []
+      } )
+  ; ( "repository-forget"
+    , {
+        reqd= ["uuid"]
+      ; optn= []
+      ; help= "Forget about a repository."
```
lindig, line 3116: Remove a repository from the database.

ocaml/xapi-consts/constants.ml
```diff
@@ -152,6 +152,13 @@ let audit_log_uri = "/audit_log"
 let get_pool_update_download_uri = "/update/"
 
 (* ocaml/xapi/xapi_pool_update.ml *)
+
+let get_repository_uri = "/repository/" (* ocaml/xapi/repository.ml *)
```
lindig, line 156: Does this really depend on the trailing `/`? Is there a difference in an URI between `a/b` and `a/b/`? If not, I would prefer to drop the trailing slash.

ocaml/xapi/message_forwarding.ml
```diff
@@ -610,6 +610,14 @@ functor
           Ref.string_of vusb
       with _ -> "invalid"
 
+     let repository_uuid ~__context repository =
+       try
+         if Pool_role.is_master () then
+           Db.Repository.get_uuid __context repository
+         else
+           Ref.string_of repository
+       with _ -> "invalid"
```
lindig, line 619: This does not seem like a good solution. It would be better to use an option value if this can be invalid.

ocaml/xapi/xapi_pool.ml
```diff
@@ -2814,3 +2814,97 @@ let remove_from_guest_agent_config ~__context ~self ~key =
...
+
+let get_updates_handler (req : Http.Request.t) s _ =
+  debug "Pool.get_updates_handler: received reqeust";
+  req.Http.Request.close <- true;
+  let query = req.Http.Request.query in
+  let prefix = !Xapi_globs.repository_prefix in
+  Xapi_http.with_context "Getting available updates" req s (fun __context ->
+      let hosts = match List.assoc "host_refs" query with
+          | v ->
+              List.map (fun ref_str -> Ref.of_string ref_str) (Str.split (Str.regexp " +") v)
```
lindig, line 2892: Avoid the `Str` module - use `Re`

psafont: `Astring.String.cuts ~sep:" +" v` is another option

ocaml/xapi/xapi_pool.ml
```diff
@@ -2814,3 +2814,97 @@ let remove_from_guest_agent_config ~__context ~self ~key =
...
+                  create_pool_repository ~__context ~self:repository ~prefix;
+                  let h = with_pool_repository ~__context ~self:repository ~prefix (fun () ->
+                       try_set_available_updates ~__context ~self:repository ~prefix)
+                  in
+                  Some h)
+                (fun () -> reposync_unlock ())
+            | false -> None (* still hold the reposync lock *))
+        else raise Api_errors.(Server_error (reposync_in_progress, [])))
+  in
+  match hash with
```
robhoes, line 2866: This function has some issues with operations and locking.

This match statement has a separate `with_pool_operation`, which means that in theory someone can get in between and "steal" the operation, and make this function fail halfway. It's better to have a single `with_pool_operation` for the whole function.

I also don't like how the function can return in various places, where only sometimes `reposync_unlock` is called at the end, and sometimes it is unlocked earlier. It is probably fine now, but unnecessarily hard to verify, and someone may mess it up later by accident.

Why not put the whole `None` case below inside the `false` case in line 2863, and move the `finally` outside the `match force with` statement? You can then probably just use `with_reposync_lock`?

robhoes: Is the only difference between the `force` true/false cases the `cleanup` call? Then we can get rid of several duplicated lines.

robhoes (changes requested):
I've reviewed everything except repository.ml, which is a large new 1000-line file. I don't think it is practical do review all of that in a single PR. Let's use this PR for all the datamodel stuff and related infrastructure (which mostly looks fine), so we can get that in first. Please have a separate PR (or multiple) for the functionality in repository.ml, ideally split into logical commits.

ocaml/xapi/repository.ml
```diff
@@ -0,0 +1,127 @@
...
+ * GNU Lesser General Public License for more details.
+ *)
+
+module Unixext = Xapi_stdext_unix.Unixext
+
+module D = Debug.Make (struct let name = "repository" end)
+
+open D
+
+let cmd_yum = "/usr/bin/yum"
```
robhoes, line 21: This should be defined in `xapi_globs.ml` under `essential_executables`.

ocaml/xapi/repository.ml
```diff
@@ -0,0 +1,127 @@
...
+let cmd_rm = "/usr/bin/rm"
```
robhoes, line 22: Don't use `rm` through the shell, but `Unix.unlink`.

ocaml/xapi/repository.ml
```diff
@@ -0,0 +1,127 @@
...
+  with _ -> warn "Unable to clean YUM cache of %s-%s" prefix name
+
+let cleanup ~__context ~self ~prefix =
+  if self <> Ref.null then
+    let name = Db.Repository.get_name_label ~__context ~self in
+    try
+      clean_yum_cache ~prefix ~name;
+      ignore (Helpers.call_script
+                cmd_rm
+                ["-f"; (Printf.sprintf "/etc/yum.repos.d/%s-%s.repo" prefix name)]);
```
robhoes, line 123: Is it useful to have the name_label as part of this file? We only allow enabling one at a time, so why don't we just use a fixed name, settable in the config file? Then also we don't need to restrict the format of the name_label and worry about changes to it.

ocaml/xapi/repository.ml
```diff
@@ -0,0 +1,127 @@
...
+let assert_name_is_valid ~name_label =
+  match Re.Str.string_match (Re.Str.regexp "^[A-Za-z]+[A-Za-z0-9\\-]+$") name_label 0 with
+  | true -> ()
+  | false -> raise Api_errors.(Server_error (invalid_repository_name, [name_label]))
+
+let assert_url_is_valid ~url =
+  match !Xapi_globs.repository_domain_name_allowlist with
+  | [] -> ()
+  | l ->
+    let result = l |> List.exists (fun domain_name ->
```
robhoes, line 50: Can we use `Uri.of_string` (see e.g. `guest_agent_config_requirements` in xapi_pool.ml)?

ocaml/xapi/repository.ml
```diff
@@ -0,0 +1,127 @@
...
+    begin match result with
+      | true -> ()
+      | false -> raise Api_errors.(Server_error (invalid_base_url, [url]))
+    end
+
+let introduce ~__context ~name_label ~name_description ~binary_url ~source_url =
+  assert_name_is_valid ~name_label;
+  assert_url_is_valid ~url:binary_url;
+  assert_url_is_valid ~url:source_url;
+  match Db.Repository.get_by_name_label ~__context ~label:name_label with
```
robhoes, line 67: Isn't it more useful to test for uniqueness based on the URLs? The name is just an arbitrary string.

ocaml/xapi/repository.ml
```diff
@@ -0,0 +1,127 @@
...
+  | [] ->
+    create_repository_record  ~__context ~name_label ~name_description ~binary_url ~source_url
+  | _ ->
+    error "A repository with same name_label '%s' already exists" name_label;
+    raise Api_errors.(Server_error (repository_already_exists, []))
```
robhoes, line 72: You could add the ref as an argument (also add it to the definition in the datamodel).

ocaml/xapi/repository.ml
```diff
@@ -0,0 +1,127 @@
...
+
+let forget ~__context ~self =
+  let pool = Helpers.get_pool ~__context in
+  let enabled = Db.Pool.get_repository ~__context ~self:pool in
+  if enabled = self then
+    raise Api_errors.(Server_error (repository_is_in_use, []))
+  else
+    Db.Repository.destroy ~__context ~self
```
robhoes, line 80: Should we call `cleanup`, or is it always "clean" when not enabled?

ocaml/xapi/repository.ml
```diff
@@ -0,0 +1,127 @@
...
+
+let reposync_try_lock () =
+  Mutex.try_lock reposync_mutex
+
+let reposync_unlock () =
+  Mutex.unlock reposync_mutex
```
robhoes, line 86: I'd avoid exposing these if possible. The with_ function below is safer.

ocaml/xapi/repository.ml
```diff
@@ -0,0 +1,127 @@
...
+  try
+    let params =
+        [
+          "--disablerepo=*";
+          Printf.sprintf "--enablerepo=%s-%s" prefix name;
+          "clean";
+          "expire-cache";
+        ]
+    in
+    ignore (Helpers.call_script cmd_yum params)
```
robhoes, line 113: Does it log the command output on failure?

robhoes (changes requested):
Most of yesterday's comments were addressed, which is good. Now I've got some questions on repository.ml.

ocaml/xapi/repository.ml
minglumlu: I think it is always "clean" when not enabled, as `pool.set_repository` does the cleanup for a repository to be disabled.

ocaml/xapi/repository.ml
minglumlu: It is used to remove a non-empty directory.
It looks like "Unix.unlink" works for a file only.  And Unix. rmdir works for an empty directory.
May I please know any existing utility before writing one?

ocaml/xapi/repository.ml
minglumlu: I can't see any benefit from `Uri.of_string` and `Uri.scheme` as I can' find how these functions check the URL.
Please correct me if I missed something by reading https://github.com/mirage/ocaml-uri/blob/master/lib/uri.ml.

ocaml/xapi/repository.ml
robhoes: If it needs a recursive delete, then you could copy the function from https://github.com/xapi-project/xenopsd/blob/master/lib/xenops_utils.ml#L186-L203.

ocaml/xapi/repository.ml
robhoes: I assume that `Uri.of_string` returns an error if it fails to parse the given string as a well-formed URL. We can use it to sanity check the whole URL. You still need to check whether the domain is in the allowlist. I think we can use `Uri.host` and match the suffix.

I find it hard to verify that the regexp below that all it needs to do (but perhaps it's just me :)).

ocaml/xapi/repository.ml
```diff
@@ -0,0 +1,118 @@
...
+ *)
+
+module Unixext = Xapi_stdext_unix.Unixext
+
+module D = Debug.Make (struct let name = "repository" end)
+
+open D
+
+let yum_repos_config_dir  = "/etc/yum.repos.d"
+let pool_repo_name = "pool-repo"
```
robhoes, line 22: Just put these in xapi_globs please.

robhoes (approved):

ocaml/xapi/repository.ml
```diff
@@ -0,0 +1,116 @@
...
+  let pool = Helpers.get_pool ~__context in
+  let enabled = Db.Pool.get_repository ~__context ~self:pool in
+  if enabled = self then
+    raise Api_errors.(Server_error (repository_is_in_use, []))
+  else
+    Db.Repository.destroy ~__context ~self
+
+let with_reposync_lock f =
+  if Mutex.try_lock reposync_mutex then
+    Xapi_stdext_pervasives.Pervasiveext.finally
```
lippirk, line 82: It's probably better to use this finally equivalent from the stdlib: `Fun.protect ~finally:(fun () -> Mutex.unlock reposync_mutex) f `

ocaml/xapi/repository.ml
```diff
@@ -0,0 +1,116 @@
...
+              begin match List.exists (fun d -> Astring.String.is_suffix ("." ^ d) h) l with
+                | true -> ()
+                | false ->
+                  let msg = "host is not in allowlist" in
+                  raise Api_errors.(Server_error (internal_error, [msg]))
+              end
+          end
+        | None -> raise Api_errors.(Server_error (internal_error, ["invalid host in url"]))
+      end
+    | _ -> raise Api_errors.(Server_error (internal_error, ["invalid scheme in url"]))
```
lippirk, line 57: Can we put the url in these error messages to make debugging easier? Alternatively we could just log it

minglumlu: These internal errors will be caught below and raise a invalid_base_url which is exposed to users. In this API error, the url will be there as a parameter.

lippirk (approved):

ocaml/xapi/repository.ml
psafont: I don't think the stdext's `finally` is equivalent to `Fun.protect`, they handle some exceptions differently. Let me find the github thread where these were talked about.

minglumlu: I looked at this for a while. It looks like that the `Xapi_stdext_pervasives.Pervasiveext.finally` has an extra feature "Backtrace.is_important exn".
To be honest, I don't clearly understand the function of this: https://github.com/xapi-project/backtrace/blob/master/lib/backtrace.ml#L195

But as this `finally` is widely used in Toolstack, I think it would be fine to use it here.

psafont: The difference is `Fun.protect` does not try to do any work when the finalizing function throws an exception and covers the original exception. The ocaml designers consider this a programming error.

`Pervasives.finally` protects against these bugs and logs the exception raised by the finaliser.

PR #4300, opened 2021-01-21
CP-35372: Repository: pool.sync_updates

minglumlu:
This commit implements 'pool.sync_updates' API for repository function.

The handler of this API will:
1. sync/download latest updates and metadata from remote repository;
2. create local repository on the master host in the pool;
3. check if there are available updates on each host in the pool.

### DIFF
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -1326,13 +1326,46 @@ let resolve_uri_path ~root ~uri_path =
   |> Filename.concat root
   |> Uri.pct_decode
   |> Xapi_stdext_unix.Unixext.resolve_dot_and_dotdot
-  |> (fun x -> match (String.startswith (root^"/") x), (Sys.file_exists x) with
+  |> (fun x -> match (Astring.String.is_prefix ~affix:(root^"/") x), (Sys.file_exists x) with
       | true, true -> x
       | _ ->
         let msg =
           Printf.sprintf "Failed to resolve uri path '%s' under '%s': %s" uri_path root x
         in
         raise Api_errors.(Server_error (internal_error, [msg])))
+
+let run_in_parallel ~funs ~capacity =
+  let rec run_in_parallel' acc funs capacity =
+    let rec split_for_first_n acc n l =
+      match n, l with
+      | n, h::t when n > 0 -> split_for_first_n (h :: acc) (n - 1) t
+      | _ -> acc, l
+    in
+    let run f =
+      let result = ref `Not_started in
+      let wrapper r = try r := `Succ (f ()) with e -> r := `Fail e in
+      let th = Thread.create wrapper result in
+      th, result
+    in
+    let get_result (th, result) =
+      Thread.join th;
+      match !result with
+      | `Not_started -> `Error (Failure "The thread in run_in_parallel is not started")
+      | `Succ s -> `Ok s
+      | `Fail e -> `Error e
+    in
+    let to_be_run, remaining = split_for_first_n [] capacity funs in
+    match to_be_run with
+    | [] -> acc
+    | _ ->
+      let finished =
+        List.map run to_be_run
+        |> List.map get_result
+        |> List.map (function `Ok s -> s | `Error e -> raise e)
+      in
+      run_in_parallel' (List.rev_append finished acc) remaining capacity
+  in
+  run_in_parallel' [] funs capacity
 
 (**************************************************************************************)
 (* The master uses a global mutex to mark database records before forwarding messages *)
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -1320,6 +1320,20 @@ let rmtree path =
     let msg = Printf.sprintf "failed to remove %s: %s" path exn' in
     failwith msg
 
+let resolve_uri_path ~root ~uri_path =
+  let open Xapi_stdext_std.Xstringext in
+  uri_path
+  |> Filename.concat root
+  |> Uri.pct_decode
+  |> Xapi_stdext_unix.Unixext.resolve_dot_and_dotdot
+  |> (fun x -> match (String.startswith (root^"/") x), (Sys.file_exists x) with
+      | true, true -> x
+      | _ ->
+        let msg =
+          Printf.sprintf "Failed to resolve uri path '%s' under '%s': %s" uri_path root x
+        in
+        raise Api_errors.(Server_error (internal_error, [msg])))
+
 (**************************************************************************************)
 (* The master uses a global mutex to mark database records before forwarding messages *)
 
--- a/ocaml/xapi/repository.ml
+++ b/ocaml/xapi/repository.ml
@@ -136,6 +136,15 @@ let remove_repo_conf_file repo_name =
 
 let write_yum_config ?(source_url=None) binary_url repo_name =
   let file_path = Filename.concat !Xapi_globs.yum_repos_config_dir (repo_name ^ ".repo") in
+  let gpgkey_path =
+    Filename.concat !Xapi_globs.rpm_gpgkey_dir !Xapi_globs.repository_gpgkey_name
+  in
+  if !Xapi_globs.repository_gpgcheck then
+    (if not (Sys.file_exists gpgkey_path) then
+       raise Api_errors.(Server_error (internal_error, ["gpg key file does not exist"]));
+     if not ( (Unix.lstat gpgkey_path).Unix.st_kind = Unix.S_REG ) then
+       raise Api_errors.(Server_error (internal_error,
+                                       ["gpg key file is not a regular file"])));
   let content_of_binary =
     [
       Printf.sprintf "[%s]" repo_name;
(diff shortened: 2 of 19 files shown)


ocaml/xapi/repository.ml
```diff
@@ -114,3 +129,476 @@ let cleanup_pool_repo () =
...
+       end
+     | _ ->
+       error "Missing 'repomd' node";
+       raise Api_errors.(Server_error (invalid_repomd_xml, []))
+     | exception e ->
+       error "Failed to parse repomd.xml: %s" (ExnHelper.string_of_exn e);
+       raise Api_errors.(Server_error (invalid_repomd_xml, []))
+    end
+
+let run_in_parallel ~funs ~capacity =
```
lippirk, line 254: I think this belongs in helpers.ml

ocaml/xapi/repository.ml
```diff
@@ -114,3 +129,476 @@ let cleanup_pool_repo () =
...
+    match Yojson.Basic.Util.member "updates" json with
+    | `List [] -> false
+    | _ -> true
+    | exception e ->
+      let uuid = Db.Host.get_uuid ~__context ~self:host in
+      error "Invalid updates from host uuid='%s': %s" uuid (ExnHelper.string_of_exn e);
+      raise Api_errors.(Server_error (get_host_updates_failed, [uuid]))
+  in
+  let funs = List.map (fun h -> are_updates_available h) hosts in
+  let rets_of_hosts = run_in_parallel funs capacity_in_parallel in
```
lippirk, line 339: If `are_updates_available` fails on any of the hosts we'll throw `get_host_updates_failed` which I believe is correct

ocaml/xapi/repository.ml
```diff
@@ -114,3 +129,476 @@ let cleanup_pool_repo () =
...
+            !Xapi_globs.local_repo_name
+          ]
+      in
+      ignore (Helpers.call_script !Xapi_globs.yum_config_manager_cmd config_params);
+      f ())
+    (fun () -> remove_repo_conf_file !Xapi_globs.local_repo_name)
+
+let assert_yum_error output =
+  let errors = ["Updateinfo file is not valid XML"] in
+  let open Xapi_stdext_std.Xstringext in
```
lippirk, line 442: You have used Xstringext in some places and Astring in others (astring would be preferred)

ocaml/xapi/repository.ml
```diff
@@ -114,3 +129,476 @@ let cleanup_pool_repo () =
...
+  List.iter (fun err ->
+    if String.has_substr output err then (
+      error "Error from 'yum updateinfo list': %s" err;
+      raise Api_errors.(Server_error (invalid_updateinfo_xml, [])))
+    else ()) errors;
+  output
+
+let pkg_of_fullname s =
+  (* s I.E. "libpath-utils-0.2.1-29.el7.x86_64" *)
+  let r = Re.Str.regexp "^\\(.+\\)\\.\\(noarch\\|x86_64\\)$" in
```
lippirk, line 452: You can use `{| |}` and avoid the double escape :)

```
{|^\(.+\)\.\(noarch\|x86_64\)$|}
```

ocaml/xapi/repository.ml
robhoes: Yes, well done on this quite complicated procedure. It looks right to me.

robhoes (changes requested):
There is a lot of parsing and many yum/rpm command calls happening, which I can't really verify. However, I believe the code looks mostly quite good. I just had a couple of minor comments.

ocaml/xapi/repository.ml
minglumlu: Indeed it would be much more clear to get rid of these backslashes. But not got this passed build after trying with long time :(

ocaml/xapi/xapi_globs.ml
```diff
@@ -862,6 +862,22 @@ let yum_repos_config_dir  = ref "/etc/yum.repos.d"
...
+
+let reposync_cmd = ref "/usr/bin/reposync"
+
+let createrepo_cmd = ref "/usr/bin/createrepo_c"
+
+let modifyrepo_cmd = ref "/usr/bin/modifyrepo_c"
+
+let rpm_cmd = ref "/usr/bin/rpm"
+
+let gpgkey_path = ref "/etc/pki/rpm-gpg/RPM-GPG-KEY-Citrix"
```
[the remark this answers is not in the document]
minglumlu: The procedure to deploy and rotate this key has not been determined.
But I think the directory `/etc/pki/rpm-gpg` is a common one for RPM gpg keys of many vendors.

ocaml/xapi/xapi_pool.ml
```diff
@@ -2831,3 +2831,29 @@ let set_repository ~__context ~self ~value =
             Db.Pool.set_repository ~__context ~self ~value;
             Db.Repository.set_hash ~__context ~self:value ~value:"";
             Db.Repository.set_up_to_date ~__context ~self:value ~value:false))
+
+let sync_updates ~__context ~self ~force =
+  let open Repository in
+  match force with
+  | true ->
```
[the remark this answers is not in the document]
minglumlu: `with_pool_operation` will be used by following APIs:
1. set_repository
2. get_updates: in next PR
3. sync_updates
4. apply_updates: in next of next PR

While in `sync_updates`, the `sync` may take long time, which may fail other APIs like `get_updates`.
The code here is to use another lock `reposync_mutex` to protect the sync between `set_repository` and `sync_updates`.  In other words, `get_updates` can be served even a `sync_updates` is in syncing.

ocaml/xapi/repository.ml
minglumlu: Once joined all hosts in one batch, and if there is an error, it will raise exception, and will not continue on next batch.
But in one batch, I think it's better to wait for joining all (e.g. 16) hosts. Otherwise the main thread may lose control on un-finished threads.

ocaml/xapi/helpers.ml
```diff
@@ -1320,6 +1320,20 @@ let rmtree path =
     let msg = Printf.sprintf "failed to remove %s: %s" path exn' in
     failwith msg
 
+let resolve_uri_path ~root ~uri_path =
+  let open Xapi_stdext_std.Xstringext in
+  uri_path
+  |> Filename.concat root
+  |> Uri.pct_decode
+  |> Xapi_stdext_unix.Unixext.resolve_dot_and_dotdot
+  |> (fun x -> match (String.startswith (root^"/") x), (Sys.file_exists x) with
```
psafont: Please use `Astring.String.is_prefix ~affix:(root ^
"/") x` instead:
```suggestion
  uri_path
  |> Filename.concat root
  |> Uri.pct_decode
  |> Xapi_stdext_unix.Unixext.resolve_dot_and_dotdot
  |> (fun x -> match (Astring.String.is_prefix ~affix:(root ^
"/") x, Sys.file_exists x) with
```

robhoes: I see that this change has now been made.

ocaml/xapi/xapi_pool.ml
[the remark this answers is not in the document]
robhoes: So operations like get_updates and apply_updates can run at the same time as `sync`, but not with `cleanup_repo` and `create_pool_repository`? And `with_reposync_lock` just exists so that you can take `sync` outside of `Xapi_pool_helpers.with_pool_operation`? But not in the `true` case, because you need to do a cleanup under with_pool_operation first anyway? Let's at least add a comment about this ;)

robhoes (approved):

ocaml/xapi/repository.ml
```diff
@@ -114,3 +129,476 @@ let cleanup_pool_repo () =
...
+      ~is_local_superuser:true ~subject:Ref.null ~auth_user_sid:""
+      ~auth_user_name:"" ~rbac_permissions:[]
+  in
+  let request = Xapi_http.http_request
+      ~cookie:[("session_id", Ref.string_of host_session_id)]
+      ~query:[("installed", (string_of_bool installed))]
+      Http.Get Constants.get_host_updates_uri
+  in
+  let host_name = (Db.Host.get_hostname ~__context ~self:host) in
+  let host_addr = Db.Host.get_address ~__context ~self:host in
```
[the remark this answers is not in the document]
lippirk:
```
$ git diff upstream/master..upstream/feature/REQ-403 | grep set_address ; echo $?
1
```
Good spot, thankfully it'll be a tiny change :)

ocaml/xapi/repository.ml
lippirk: Okay, nevermind! It is clear what's going on right now anyway

lippirk (approved):

ocaml/xapi/xapi_pool.ml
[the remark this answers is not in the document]
minglumlu: Exactly. I will add more comments in next PR for this. Thanks.

PR #4302, opened 2021-01-22
XSI-804 ensure HVM boot params consistent

lippirk:
HVM VMs need a boot order in order for qemu to successfully start them.
This check ensures early and obvious failure when the boot order is not
present, whereas previously VM.start would silently fail.

### DIFF
--- a/ocaml/xapi-consts/constants.ml
+++ b/ocaml/xapi-consts/constants.ml
@@ -163,6 +163,8 @@ let hvm_boot_policy_bios_order = "BIOS order"
    Value is the boot string we send to qemu-dm (eg cd, dc, dcn, etc) *)
 let hvm_boot_params_order = "order"
 
+let hvm_default_boot_order = "cd"
+
 (* Key we put in VM.other_config when we upgrade a VM from Zurich/Geneva to Rio *)
 let vm_upgrade_time = "upgraded at"
 
--- a/ocaml/xapi/import_xva.ml
+++ b/ocaml/xapi/import_xva.ml
@@ -71,7 +71,13 @@ let make __context rpc session_id srid (vms, vdis) =
             other_platform
       in
       let hVM_boot_policy = if vm.is_hvm then "BIOS order" else "" in
-      let hVM_boot_params = if vm.is_hvm then [("order", "cd")] else [] in
+      let hVM_boot_params =
+        if vm.is_hvm then
+          let open Constants in
+          [(hvm_boot_params_order, hvm_default_boot_order)]
+        else
+          []
+      in
       let domain_type = Xapi_vm_helpers.derive_domain_type ~hVM_boot_policy in
       let vm_ref =
         Client.VM.create ~rpc ~session_id ~name_label:(vm.vm_name ^ " import")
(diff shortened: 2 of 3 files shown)


ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -58,6 +58,30 @@ let ensure_domain_type_is_specified ~__context ~self =
...
+          None
+    in
+    match msg with
+    | None ->
+        ()
+    | Some msg ->
+        raise
+          Api_errors.(
+            Server_error
+              ( internal_error
```
psafont, line 82: If this is an internal error, is there something the user can do to fix the root cause?

Can xapi try to recover from this and give a default for boot params' order?

lippirk: Xapi could guess the boot order, but I'm not sure that's a good idea

I was expecting that the user would call something like `xe vm-param-set HVM-boot-params:order=cdn`

psafont: It might be worth creating an API error for this situation so xencenter can have more control over this.

@kc284 what do you think?

robhoes: We have https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_xenops.ml#L404, which fills in `"cd"` by default, but only if `order` is missing from the map rather than empty. We could change that to put in the default for `""` as well.

robhoes: If we raise an error, then it should not be an internal error, because that implies a bug rather than a problem with user input. You could use `invalid_value`.

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -58,6 +58,30 @@ let ensure_domain_type_is_specified ~__context ~self =
...
+    in
+    match msg with
+    | None ->
+        ()
+    | Some msg ->
+        raise
+          Api_errors.(
+            Server_error
+              ( invalid_value
+              , [Printf.sprintf "VM=%s %s" (Ref.short_string_of self) msg] ))
```
robhoes, line 83: This error is defined as `error Api_errors.invalid_value ["field"; "value"]` in datamodel_errors.ml, so you need something like `["HVM_boot_params:order", value_from_db]`.

robhoes: By the way, we are not at all consistent with these errors in our code base :(

ocaml/xapi/xapi_vm_helpers.ml
lippirk: Ah I didn't see your comment earlier about xapi_xenops filling in the order parameter with a default. That sounds like a better solution to me - as long as assuming a boot order isn't going to make the VM explode

robhoes (approved):

ocaml/xapi/xapi_vm_helpers.ml
psafont: Can we add this default value to `ocaml/xapi-consts/constants.ml` and use it in all sites?

```
ocaml/xapi-consts/constants.ml
164:let hvm_boot_params_order = "order"

xen-api/ocaml/xapi/import_xva.ml
74:      let hVM_boot_params = if vm.is_hvm then [("order", "cd")] else [] in
xen-api/ocaml/xapi/xapi_xenops.ml
404:    ; boot_order= string vm.API.vM_HVM_boot_params "cd" "order"
```

lippirk: I will, though this import xva code is going to be removed soon iirc

psafont (approved):

Issue #4301, opened 2021-01-21
DNS not stored when PIF configured in IPv6


benjamreis:
Hi all,

I'm still working on IPv6 support and I've noticed that when i call `PIF.reconfigure_ipv6`, the DNS i provide to the method is not set in the PIF params.

I can't find the root cause of the issue so I'm seeking some guidance. :)

Thanks

lippirk:
This is surprising - in the code I see the following:

```
(* xapi_pif.ml *)
let reconfigure_ipv6 =
....
  if dNS <> "" then Db.PIF.set_DNS ~__context ~self ~value:dNS ;
....
```

Maybe you could add some debugging around this area to see exactly what's going on

benjamreis:
Yes I suspect the issue comes after in `NM.bring_pif_ip` but didn't found for now. ^^
I'll add logs indeed.

robhoes:
Yes, `bring_pif_up` constructs `dns` only based on if there is a static IPv4 config (see `let ipv4_conf, ipv4_gateway, dns =`).

benjamreis:
Yeah I saw that, but didn't know why. It seems there's nothing IPv4 specific for DNS.

robhoes:
You're right, it's not specific to IPv4. And this code is definitely not correct for IPv6-only use cases :)

The main thing to note is how the DNS field – as well as IP/netmask/gateway – is an input field (i.e. set by the client) for static configurations, but an output field (i.e. set by xapi) for DHCP (or autoconf for IPv6). And only input parameters should be sent to xcp-network in `bring_pif_up`. So you need to work out potential conflicts if `IP_configuration_mode` is `DHCP` but `IPv6_configuration_mode` is `static`, etc.

benjamreis:
I see, what about the `primary_address_type` choses if we look for ip or ipv6 configuration mode?

benjamreis:
Actually, if a pif as many ip configuration, it might have several DNSs configured no?

robhoes:
It is probably a very unusual situation – we just need a deterministic way of handling it. Using `primary_address_type` to choose in case of conflict could be a good option.

I'd also point you at https://github.com/xapi-project/xcp-networkd/blob/master/networkd/network_server.ml#L305-L311, which controls whether DNS servers requested from the DHCP server in DHCP mode (only if the interface is "the DNS interface"). There is no such code for the IPv6 case now, so it probably never requests DNS servers there.

robhoes:
[...]

PR #4303, opened 2021-01-22
fixes regarding DNS management

benjamreis:

https://github.com/xapi-project/xen-api/issues/4301

### DIFF
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -173,7 +173,16 @@ let determine_gateway_and_dns_ifs ~__context
       ~expr:
         (And
            ( Eq (Field "host", Literal (Ref.string_of localhost))
-           , Not (Eq (Field "ip_configuration_mode", Literal "None"))
+           , Or
+               ( And
+                   ( Not (Eq (Field "ip_configuration_mode", Literal "None"))
+                   , Eq (Field "primary_address_type", Literal "IPv4")
+                   )
+               , And
+                   ( Not (Eq (Field "ipv6_configuration_mode", Literal "None"))
+                   , Eq (Field "primary_address_type", Literal "IPv6")
+                   )
+               )
            )
         )
   in
--- a/ocaml/xapi/nm.ml
+++ b/ocaml/xapi/nm.ml
@@ -552,6 +552,7 @@ let bring_pif_up ~__context ?(management_interface = false) (pif : API.ref_PIF)
                   bond_record.API.bond_slaves ;
                 maybe_update_master_pif_mac ~__context bond_record rc pif
           in
+          let address_type = rc.API.pIF_primary_address_type in
           Network.transform_networkd_exn pif (fun () ->
               let persistent = is_dom0_interface rc in
               let gateway_if, dns_if =
@@ -576,12 +577,12 @@ let bring_pif_up ~__context ?(management_interface = false) (pif : API.ref_PIF)
               Net.Bridge.make_config dbg false bridge_config ;
               Net.Interface.make_config dbg false interface_config ;
               (* Configure IPv4 parameters and DNS *)
-              let ipv4_conf, ipv4_gateway, dns =
+              let ipv4_conf, ipv4_gateway =
                 match rc.API.pIF_ip_configuration_mode with
                 | `None ->
-                    (None4, None, ([], []))
+                    (None4, None)
                 | `DHCP ->
-                    (DHCP4, None, ([], []))
+                    (DHCP4, None)
                 | `Static ->
                     let conf =
                       Static4
@@ -597,30 +598,7 @@ let bring_pif_up ~__context ?(management_interface = false) (pif : API.ref_PIF)
                       else
                         None
                     in
-                    let dns =
-                      if rc.API.pIF_DNS <> "" then
-                        let nameservers =
-                          List.map Unix.inet_addr_of_string
-                            (String.split ',' rc.API.pIF_DNS)
-                        in
-                        let domains =
-                          if List.mem_assoc "domain" rc.API.pIF_other_config
-                          then (
-                            let domains =
-                              List.assoc "domain" rc.API.pIF_other_config
-                            in
-                            try String.split ',' domains
-                            with _ ->
-                              warn "Invalid DNS search domains: %s" domains ;
-                              []
-                          ) else
-                            []
-                        in
-                        (nameservers, domains)
-                      else
-                        ([], [])
-                    in
-                    (conf, gateway, dns)
+                    (conf, gateway)
               in
               let ipv4_routes = determine_static_routes net_rc in
               (* Configure IPv6 parameters *)
@@ -660,6 +638,35 @@ let bring_pif_up ~__context ?(management_interface = false) (pif : API.ref_PIF)
                     in
                     (conf, gateway)
               in
+              let static =
+                match address_type with
+                | `IPv4 ->
+                    rc.API.pIF_ip_configuration_mode = `Static
+                | `IPv6 ->
+                    rc.API.pIF_ipv6_configuration_mode = `Static
+              in
+              let dns =
+                match (static, rc.API.pIF_DNS) with
+                | false, _ | true, "" ->
+                    ([], [])
+                | true, pif_dns ->
+                    let nameservers =
+                      List.map Unix.inet_addr_of_string
+                        (String.split ',' pif_dns)
+                    in
+                    let domains =
+                      match List.assoc_opt "domain" rc.API.pIF_other_config with
+                      | None ->
+                          []
+                      | Some domains -> (
+                        try String.split ',' domains
+                        with _ ->
+                          warn "Invalid DNS search domains: %s" domains ;
+                          []
+                      )
+                    in
+                    (nameservers, domains)
+              in
               let mtu = determine_mtu rc net_rc in
               let ethtool_settings, ethtool_offload =
                 determine_ethtool_settings rc.API.pIF_properties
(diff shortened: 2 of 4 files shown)


ocaml/xapi/nm.ml
```diff
@@ -618,6 +596,36 @@ let bring_pif_up ~__context ?(management_interface = false) (pif : API.ref_PIF)
...
+                      if List.mem_assoc "domain" rc.API.pIF_other_config
+                      then (
+                        let domains =
+                          List.assoc "domain" rc.API.pIF_other_config
+                        in
+                        try String.split ',' domains
+                        with _ ->
+                          warn "Invalid DNS search domains: %s" domains ;
+                          []
+                      ) else []
```
psafont, line 622: Consider using `match List.assoc_opt with None -> [] | Some domains -> ...`

benjamreis: instead of using both `mem_assoc` and `assoc` right?

benjamreis:
This does not solve the overrides of DNS when one PIF is configured `Static` on one family address and `DHCP` on the other.

I propose to catch the overides here and to have check on family and configuration mode to set accordingly the namesrvers (concataining the Static ones to the ones given by DHCP).
What do you think?

ocaml/xapi/nm.ml
psafont: That's correct, that would mean traversing the list once instead of twice.

lindig (approved):
The code looks ok but there is a lot of string manipulation going on and using empty strings for results which are not good for long-term code health. These problems existed mostly before, though.

benjamreis:
Would it make sense to add an ipv6_DNS field? like its done for the gateway?

robhoes:
Please describe the fixes in the commit message: the problem and the choices made to fix it.

benjamreis:
The commit message has been changed. Is it okay?

ocaml/xapi/nm.ml
```diff
@@ -618,6 +597,37 @@ let bring_pif_up ~__context ?(management_interface = false) (pif : API.ref_PIF)
...
+                        with _ ->
+                          warn "Invalid DNS search domains: %s" domains ;
+                          []
+                      )
+                    in
+                    (nameservers, domains)
+                  else
+                    ([], [])
+                else
+                  ([], [])
```
lippirk, line 629: A match would make this slightly neater

```suggestion
                match static, rc.API.pIF_DNS with
                | false, _ | true, "" -> ([], [])
                | true, pif_dns ->
                    let nameservers =
                      List.map Unix.inet_addr_of_string
                        (String.split ',' pif_dns)
                    in
                    let domains =
                      match List.assoc_opt "domain" rc.API.pIF_other_config with
                      | None ->
                          []
                      | Some domains -> (
                        try String.split ',' domains
                        with _ ->
                          warn "Invalid DNS search domains: %s" domains ;
                          []
                      )
                    in
                    (nameservers, domains)
```

ocaml/xapi/nm.ml
```diff
@@ -642,13 +652,33 @@ let bring_pif_up ~__context ?(management_interface = false) (pif : API.ref_PIF)
...
+            match address_type with
+            | `IPv4 ->
+                rc.API.pIF_IP
+            | `IPv6 -> (
+              match rc.API.pIF_IPv6 with
+              | [] ->
+                  ""
+              | hd :: _ -> (
+                match String.split_on_char '/' hd with [ip; _] -> ip | _ -> ""
+              )
```
lippirk, line 678: Can you add a comment to show an example of what `pIF_IPv6` looks like? It's not obvious to me why it would contain slashes

And if it doesn't contain a slash, would it not be reasonable to assume the whole string is a valid IP? However, this should be obvious from an example

ocaml/xapi/nm.ml
psafont: I don't think `XStringext.String.split` can raise exceptions, the try with seems useless here. It would be eveen better to replace it with a `String.split_on_char`

What would be invalid DNS domains? if the list is empty or there are empty strings in it?

ocaml/xapi/xapi_pif.ml
```diff
@@ -709,7 +709,14 @@ let reconfigure_ipv6 ~__context ~self ~mode ~iPv6 ~gateway ~dNS =
   Db.PIF.set_ipv6_configuration_mode ~__context ~self ~value:mode ;
   Db.PIF.set_ipv6_gateway ~__context ~self ~value:gateway ;
   Db.PIF.set_IPv6 ~__context ~self ~value:[iPv6] ;
-  if dNS <> "" then Db.PIF.set_DNS ~__context ~self ~value:dNS ;
+  (* Keep IPv4 DNS entries *)
+  let pif_dns =
```
lippirk, line 713: What if `dNS` is already in `pif_dns`? We probably don't want to add it to appear twice in the database

scripts/network-init
```diff
@@ -36,24 +36,29 @@ prepare_networking() {
...
                 dhcp|none|autoconf)
                     $XE pif-reconfigure-ipv6 uuid=${PIF} mode=${MODEV6} || true
                     ;;
                 static)
-                    $XE pif-reconfigure-ipv6 uuid=${PIF} mode=static IPv6=${IPv6} gateway=${IPv6_GATEWAY} || true
+                    isStatic=true
+                    $XE pif-reconfigure-ipv6 uuid=${PIF} mode=static IPv6=${IPv6} gateway=${IPv6_GATEWAY} ${DNS+DNS=$DNS} || true
                     ;;
             esac
+            if [ $isStatic = true ] ; then
```
lippirk, line 58:
```suggestion
            if $isStatic; then
```

psafont: I don't agree with this, it makes the check more brittle:
```
$ (export isStatic="" && if $isStatic; then echo yes; else echo no; fi)
yes
```

Keeping the brackets to interpret the variable as a boolean is a good idea in this case.

```
$ (export isStatic="" && if [ $isStatic ]; then echo yes; else echo no; fi)
no
```

lippirk: Maybe just keep the existing @benjamreis if it works, as bash is rather confusing:

```
$ export isStatic=false && if [ $isStatic ]; then echo "true"; fi
true
```

benjamreis: I'm not an expert in bash, it's working and from my research it seems to be the standard way to do it. So i'll jsut remove `= true` since `isStatic` is a boolean.
I need to keep the `[]` because the brackets made the check on the value of `isStatic` instead of it definitions.

ocaml/xapi/xapi_pif.ml
benjamreis: This can't happen since when setting dns in IPv6 we only keep IPv4 dns entries and same thing the other way.

And the method check above in the code that the new dns entries are in the correct ip family.

But anyway this is still not good enough to work when using both v4 and v6 I've opened [an issue](https://github.com/xapi-project/xen-api/issues/4301) to find a better way of doing it, I proposed to add a new field for IPv6 dns.

ocaml/xapi/nm.ml
benjamreis: Yes, the IPv6 are stored in the xapi db like this: `ipv6/netmask` instead of having a netmask field.
I'll add a comment.

ocaml/xapi/nm.ml
benjamreis: Since we take it from the PIF record, I don't see how there can be an invalid entries since it's tested when set on the PIF.

ocaml/xapi/helpers.ml
```diff
@@ -141,7 +141,13 @@ let determine_gateway_and_dns_ifs ~__context
       ~expr:
         (And
            ( Eq (Field "host", Literal (Ref.string_of localhost))
-           , Not (Eq (Field "ip_configuration_mode", Literal "None")) ))
+           , Or
```
robhoes, line 144: This actually (with the current code in xcp-networkd) assumes the "gateway interface" for IPv4 and IPv6 is the same. This is the interface that is used for the default route, and the gateway IP is taken off the relevant PIF.

The PIF class in the datamodel has separate gateway fields for IPv4 (`gateway`) and IPv6 (`ipv6_gateway`). But xcp-networkd only has one "gateway interface". So this will go wrong if you have (only) an IPv4 address on one PIF and (only) and IPv6 address on another PIF.

For DNS, the "DNS interface" is just the PIF that the DNS address is taking from, which is shared among IPv4 and IPv6, which is OK.

benjamreis: I'm not sure to understand. Should I change my modification or should I create a PR in xcp-networkd for this case?

robhoes: Sorry, the current PR is about DNS, which is fine. But since you are looking at improving IPv6 generally, you may want to think about the gateway thing.

benjamreis: I'll do a different Pr then. But I'm not sure what i should do.

edwintorok (approved):
Ideally the scripts here would be moved inside XAPI, but otherwise LGTM.

edwintorok:
Sorry that it has taken so long to merge this, but the PR has a conflict and cannot be merged. @benjamreis can you please merge from master and fix the conflict? AFAICT it is just about the placement of the `))` due to ocamlformat.
I've pushed an example merge that solves the conflict here: https://github.com/xapi-project/xen-api/tree/private/edvint/xcp-ng-add-ipv6-dns, all that would need to be done here is to force push that to the xcp-ng origin.

robhoes (approved):

benjamreis:
Hi! Thanks for the reviews and approvals.

I've resolved the merge conflicts and review the changes. This should be good to merge. :)

Regards

ocaml/xapi/nm.ml
```diff
@@ -660,6 +638,35 @@ let bring_pif_up ~__context ?(management_interface = false) (pif : API.ref_PIF)
...
+                    rc.API.pIF_ipv6_configuration_mode = `Static
+              in
+              let dns =
+                match (static, rc.API.pIF_DNS) with
+                | false, _ | true, "" ->
+                    ([], [])
+                | true, pif_dns ->
+                    let nameservers =
+                      List.map Unix.inet_addr_of_string
+                        (String.split ',' pif_dns)
```
psafont, line 655: This reminds me there's cleanup to do around stdext's string

PR #4306, opened 2021-01-22
Use reconfigure_ipv6 when necessary

benjamreis:

### DIFF
--- a/ocaml/xapi/xapi.ml
+++ b/ocaml/xapi/xapi.ml
@@ -581,37 +581,41 @@ let check_network_reset () =
             args
         in
         let device = List.assoc "DEVICE" args in
-        let vlan =
-          if List.mem_assoc "VLAN" args then
-            Some (List.assoc "VLAN" args)
-          else
-            None
-        in
+        let vlan = List.assoc_opt "VLAN" args in
         let mode =
-          match List.assoc "MODE" args with
-          | "static" ->
-              `Static
-          | "dhcp" | _ ->
-              `DHCP
+          match List.assoc_opt "MODE" args with
+          | Some "static" ->
+              Some `Static
+          | Some "dhcp" | Some _ ->
+              Some `DHCP
+          | None ->
+              None
         in
-        let iP =
-          if List.mem_assoc "IP" args then List.assoc "IP" args else ""
-        in
+        let iP = Option.value ~default:"" (List.assoc_opt "IP" args) in
         let netmask =
-          if List.mem_assoc "NETMASK" args then
-            List.assoc "NETMASK" args
-          else
-            ""
+          Option.value ~default:"" (List.assoc_opt "NETMASK" args)
         in
         let gateway =
-          if List.mem_assoc "GATEWAY" args then
-            List.assoc "GATEWAY" args
-          else
-            ""
+          Option.value ~default:"" (List.assoc_opt "GATEWAY" args)
         in
-        let dNS =
-          if List.mem_assoc "DNS" args then List.assoc "DNS" args else ""
+        let mode_v6 =
+          match List.assoc_opt "MODE_V6" args with
+          | Some "static" ->
+              Some `Static
+          | Some "autoconf" ->
+              Some `Autoconf
+          | Some "dhcp" | Some _ ->
+              Some `DHCP
+          | None ->
+              None
         in
+        (* This field is in the format: <ipv6>/<cidr> *)
+        let iPv6 = Option.value ~default:"" (List.assoc_opt "IPV6" args) in
+        let gateway_v6 =
+          Option.value ~default:"" (List.assoc_opt "GATEWAY_V6" args)
+        in
+        let dNS = Option.value ~default:"" (List.assoc_opt "DNS" args) in
+
         (* Get the existing network for management vlan *)
         let existing_network =
           match vlan with
@@ -675,8 +679,16 @@ let check_network_reset () =
               | None ->
                   phy_pif
             in
-            Xapi_pif.reconfigure_ip ~__context ~self:pif ~mode ~iP ~netmask
-              ~gateway ~dNS ;
+            Option.iter
+              (fun mode ->
+                Xapi_pif.reconfigure_ip ~__context ~self:pif ~mode ~iP ~netmask
+                  ~gateway ~dNS)
+              mode ;
+            Option.iter
+              (fun mode ->
+                Xapi_pif.reconfigure_ipv6 ~__context ~self:pif ~mode ~iPv6
+                  ~gateway:gateway_v6 ~dNS)
+              mode_v6 ;
             Xapi_host.management_reconfigure ~__context ~pif) ;
     (* Remove trigger file *)
     Unix.unlink Xapi_globs.network_reset_trigger


benjamreis:
I'm not sure what to do [here](https://github.com/xapi-project/xen-api/blob/master/scripts/examples/python/renameif.py#L101) (it's in an example folder, so i guess nothing.) and [here](https://github.com/xapi-project/xen-api/blob/master/ocaml/perftest/createpool.ml#L557-L559) (perftest folder so maybe nothing as well). I don't understand what's done.

ocaml/xapi/xapi.ml
```diff
@@ -675,8 +675,16 @@ let check_network_reset () =
...
-              ~gateway ~dNS ;
+            (* Reconfigure accordingly to primary_address_type *)
+            ( match Db.PIF.get_primary_address_type ~__context ~self:pif with
+            | `IPv4 ->
+                Xapi_pif.reconfigure_ip ~__context ~self:pif ~mode ~iP ~netmask
+                  ~gateway ~dNS
+            | `IPv6 ->
+                Xapi_pif.reconfigure_ipv6 ~__context ~self:pif ~mode
+                  ~iPv6:(iP ^ "/" ^ netmask)
+                  ~gateway ~dNS
```
psafont, line 686: I think it would be better to add the netmask parameter to the call, just like the ipv4 version

benjamreis: The thing is, the `IPv6` field of a PIF is a list of `ip6/cidr` string.
So it'd mean maintaining 2 lists and knowing which cidr goes with which ip.

robhoes (changes requested):
Note that this is code that does a network reset on xapi startup, after using the `xe-reset-networking` script. The script takes the new network config from the user and writes it to a file, which this function picks up. I think we should start by adding IPv6 options there, and explicitly writing them to the file, rather than overloading `iP` and `netmask` here, where `netmask` indeed isn't used for IPv6 in the same way. We need a separate IPv6 option, and then we don't need to look at the `primary_address_type`.

benjamreis:
@robhoes What do you think of it now?
This way a network reset could reconfigure both IPv4 and 6 of a pif if all the fields are present.

ocaml/xapi/xapi.ml
```diff
@@ -609,6 +611,32 @@ let check_network_reset () =
...
+          | Some "autoconf" ->
+              Some `Autoconf
+          | Some "dhcp" | Some _ ->
+              Some `DHCP
+          | None ->
+              None
+        in
+        let iPv6 =
+          if List.mem_assoc "IPV6" args then List.assoc "IPV6" args else ""
+        in
```
psafont, line 627: This idiom avoid going through the list twice: 
```suggestion
        let iPv6 = Option.value ~default:"" (List.assoc_opt "IPV6" args) in
```

It could be used for the netmask and gateway as well.

ocaml/xapi/xapi.ml
```diff
@@ -675,8 +703,29 @@ let check_network_reset () =
...
             in
-            Xapi_pif.reconfigure_ip ~__context ~self:pif ~mode ~iP ~netmask
-              ~gateway ~dNS ;
+            ( match mode with
+            | Some mode ->
+                Xapi_pif.reconfigure_ip ~__context ~self:pif ~mode ~iP ~netmask
+                  ~gateway ~dNS
+            | None ->
+                ()
+            ) ;
```
psafont, line 712: `Option.iter` is useful here: 
```suggestion
            Option.iter (fun mode ->
                Xapi_pif.reconfigure_ip ~__context ~self:pif ~mode ~iP ~netmask
                  ~gateway ~dNS)
              mode ;
```

benjamreis: Didn't know about that, thanks it's usefull indeed!

ocaml/xapi/xapi.ml
benjamreis: Yes, I can do it for the IPv4 fields as well, I was looking for such a pattern.

ocaml/xapi/xapi.ml
```diff
@@ -609,6 +611,32 @@ let check_network_reset () =
           else
             ""
         in
+        let mode_v6 =
```
lindig, line 614: This could be made simpler using `let*` but it's a matter of taste and being used to it. Below I'm demonstrating the use of the `Option.bind` monad.

```
let args = [ "MODE_V6", "autoconf" ]

let f args =
  let ( let* ) = Option.bind in
  let* mode = List.assoc_opt "MODE_V6" args in
  match mode with 
  | "static" -> Some `Static
  | "autoconf" -> Some `Autoconf
  | "dhcp" | _ -> Some `DHCP
```

ocaml/xapi/xapi.ml
```diff
@@ -675,8 +703,29 @@ let check_network_reset () =
...
+            ( match mode_v6 with
+            | Some mode_v6 ->
+                let iPv6 =
+                  match (iPv6, netmask_v6) with
+                  | "", _ ->
+                      ""
+                  | iPv6, "" ->
+                      iPv6
+                  | iPv6, netmask_v6 ->
+                      iPv6 ^ "/" ^ netmask_v6
```
robhoes, line 722: Let's keep it simple and define `IPV6` in the file to be in CIDR format.

ocaml/xapi/xapi.ml
```diff
@@ -581,37 +581,40 @@ let check_network_reset () =
...
+          | Some "static" ->
+              Some `Static
+          | Some "autoconf" ->
+              Some `Autoconf
+          | Some "dhcp" | Some _ ->
+              Some `DHCP
+          | None ->
+              None
         in
+        let iPv6 = Option.value ~default:"" (List.assoc_opt "IPV6" args) in
```
psafont, line 612: Can we add a comment somewhere that this field is in CIDR notation?

psafont (approved):
Minor nit aside, this looks good to get squashed into a single commit and be merged

ocaml/xapi/xapi.ml
benjamreis: Indeed, I'll add it and rebase :)

benjamreis: Done!

lippirk (approved):

ocaml/xapi/xapi.ml
```diff
@@ -581,37 +581,41 @@ let check_network_reset () =
...
+              Some `Autoconf
+          | Some "dhcp" | Some _ ->
+              Some `DHCP
+          | None ->
+              None
         in
+        (* This field is in the format: <ipv6>/<cidr> *)
+        let iPv6 = Option.value ~default:"" (List.assoc_opt "IPV6" args) in
+        let gateway_v6 =
+          Option.value ~default:"" (List.assoc_opt "GATEWAY_v6" args)
```
robhoes, line 615: `GATEWAY_V6` (capital V) would be more consistent.

robhoes (approved):
This is fine, but does require matching changes to xe-reset-networking to be useful at all.

ocaml/xapi/xapi.ml
benjamreis: indeed, i did the changes in our installer. I'll do them for xe-reset-networking in a separate PR.

benjamreis:
indeed, i did the changes in our xsconsole. I'll do them for xe-reset-networking in a separate PR.

ocaml/xapi/xapi.ml
benjamreis: Done :)

PR #4307, opened 2021-01-23
CP-35372: Repository: HTTPs endpoint "/updates"

minglumlu:
This commit implements HTTPs endpoint "/updates" for repository function.

With this HTTPs endpoint, user can get the available updates from hosts in the
pool.

### DIFF
--- a/ocaml/xapi/repository.ml
+++ b/ocaml/xapi/repository.ml
@@ -18,6 +18,10 @@ module D = Debug.Make (struct let name = "repository" end)
 
 open D
 
+module UpdateIdSet = Set.Make (String)
+
+module GuidanceStrSet = Set.Make (String)
+
 let capacity_in_parallel = 16
 
 let reposync_mutex = Mutex.create ()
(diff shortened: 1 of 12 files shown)


ocaml/xapi/xapi_pool.ml
```diff
@@ -2857,3 +2870,61 @@ let sync_updates ~__context ~self ~force =
...
+        Xapi_pool_helpers.with_pool_operation
+          ~__context
+          ~self:(Helpers.get_pool ~__context)
+          ~doc:"pool.get_updates"
+          ~op:`get_updates
+          (fun () ->
+             try
+               Repository.with_pool_repository
+                 (fun () ->
+                    let json_str = Yojson.Basic.pretty_to_string
```
lippirk, line 2912: Why are we pretty printing json just before sending it over the wire?

I would have expected this function to be more suitable:

```
val to_string : ?⁠buf:Bi_outbuf.t -> ?⁠len:int -> ?⁠std:bool -> t -> string
(** Write a compact JSON value to a string*)
```

ocaml/xapi/repository.ml
```diff
@@ -583,3 +702,291 @@ let get_repository_handler (req : Http.Request.t) s _ =
...
+  List.map (fun g ->
+      match g with
+      | Xml.Element ("guidance", _, [Xml.PCData v]) ->
+        guidance_of_string v
+      | _ ->
+        error "Unknown node in <absolute|recommended_guidances>";
+        raise Api_errors.(Server_error (invalid_updateinfo_xml, []))
+  ) gs
+
+let parse_applicabilities apps =
```
lippirk, line 723: It's a shame these bits of information aren't transported using json, because then we could use a library to parse it (https://github.com/ahrefs/atd) - I'm not aware of an equivalent library for xml parsing

(I assume there is a good reason for using xml?)

ocaml/xapi/repository.ml
```diff
@@ -583,3 +702,291 @@ let get_repository_handler (req : Http.Request.t) s _ =
...
+     raise Api_errors.(Server_error (createrepo_failed, [])));
+  let updateinfo_xml_gz_path = Filename.concat repo_dir md.location in
+  match Sys.file_exists updateinfo_xml_gz_path with
+  | false ->
+    error "File %s doesn't exist" updateinfo_xml_gz_path;
+    raise Api_errors.(Server_error (invalid_updateinfo_xml, []))
+  | true ->
+    with_updateinfo_xml updateinfo_xml_gz_path parse_updateinfo_xml_file
+
+let compare_segment s1 s2 =
```
lippirk, line 841: is `compare_version_strings` a better name?

ocaml/xapi/repository.ml
minglumlu: This xml is defined and used by YUM utilities (and DNF in future).

ocaml/xapi/repository.ml
lindig: I think a comment how these version strings look like would help to document the code. It's difficult to say whether the code is correct or not without having some examples. Version strings are notorious for having surprising ordering properties to the point that I believe this could be its own module.

robhoes (commented):
The file repository.ml is getting up to 1000 lines long already (and counting), and contains a mix of:
* High-level API implementation functions (`introduce`, `forget`).
* High-level HTTP handlers.
* Master-only functions that call HTTP handlers on members.
* Calls to YUM/RPM tools.
* Parsers for XML and JSON from the above.
* JSON construction code.
* Types.
* Locking.
* Various helper functions.

Without a clearer structure, no one is going to understand this code anymore after a while. Sub-modules may help, or splitting up the file. I know that you have two follow-up PRs already prepared, and restructuring the code may break those in complicated ways. So we can leave it like this now, but we should come back to this immediately after those are in.

ocaml/xapi/repository.ml
minglumlu: More examples are added. And fixed a few bugs with respect to these examples. Thanks.

minglumlu:
> The file repository.ml is getting up to 1000 lines long already (and counting), and contains a mix of:
> 
> * High-level API implementation functions (`introduce`, `forget`).
> * High-level HTTP handlers.
> * Master-only functions that call HTTP handlers on members.
> * Calls to YUM/RPM tools.
> * Parsers for XML and JSON from the above.
> * JSON construction code.
> * Types.
> * Locking.
> * Various helper functions.
> 
> Without a clearer structure, no one is going to understand this code anymore after a while. Sub-modules may help, or splitting up the file. I know that you have two follow-up PRs already prepared, and restructuring the code may break those in complicated ways. So we can leave it like this now, but we should come back to this immediately after those are in.
> 
> Otherwise, this mostly looks fine.

Thanks for this. Will get this done (resort them into sub-modules) before unit tests PR.

robhoes (approved):

ocaml/xapi/repository.ml
```diff
@@ -583,3 +702,328 @@ let get_repository_handler (req : Http.Request.t) s _ =
...
+  let normalize v =
+    let split_letters_and_numbers s =
+      let r = Re.Str.regexp "^\\([^0-9]+\\)\\([0-9]+\\)$" in
+      if Re.Str.string_match r s 0 then
+        [(Re.Str.replace_first r "\\1" s); (Re.Str.replace_first r "\\2" s)]
+      else [s]
+    in
+    v
+    |> Astring.String.cuts ~sep:"."
+    |> List.map (fun s -> Astring.String.cuts ~sep:"_" s)
```
lippirk, line 872: You want `concat_map` here (concat is just an alias for flatten)

```
List.concat (List.map f l) = List.concat_map f l
```

ocaml/xapi/repository.ml
```diff
@@ -583,3 +702,328 @@ let get_repository_handler (req : Http.Request.t) s _ =
...
+    else f ()
+  in
+  let rec compare_segments l1 l2 =
+    match l1, l2 with
+    | c1 :: t1, c2 :: t2 ->
+      let f () = compare_segments t1 t2 in
+      (match c1, c2 with
+       | Int s1, Int s2 ->
+         if s1 > s2 then Gt
+         else if s1 = s2 then f ()
```
lippirk, line 891: If you inline f, then you can get the compiler to check that your recursive calls are tail recursive with `(compare_segments [@tailcall]) t1 t2`

ocaml/xapi/xapi_pool.ml
```diff
@@ -2857,3 +2870,58 @@ let sync_updates ~__context ~self ~force =
...
+
+let get_updates_handler (req : Http.Request.t) s _ =
+  debug "Pool.get_updates_handler: received request";
+  req.Http.Request.close <- true;
+  let query = req.Http.Request.query in
+  Xapi_http.with_context "Getting available updates" req s (fun __context ->
+      let all_hosts = Db.Host.get_all ~__context in
+      let hs =
+        match List.assoc "host_refs" query with
+          | v ->
```
lindig, line 2882: I suspect that indentation is not correct.

ocaml/xapi/xapi_pool.ml
```diff
@@ -2857,3 +2870,58 @@ let sync_updates ~__context ~self ~force =
...
+                  not (Db.is_valid_ref __context ref && List.mem ref all_hosts)
+                in
+                match (List.exists is_invalid l), l with
+                | true, _ | false, [] -> None
+                | false, l -> Some l)
+          | exception Not_found -> Some all_hosts
+      in
+      match hs with
+      | Some hosts ->
+        let codes_of_404 =
```
lindig, line 2895: A better name would be simply `failures`. The name suggests currently that these errors are resulting from 404 errors but it is the reverse: these errors lead to a 404 response.

lippirk (approved):

ocaml/xapi/xapi_pool.ml
lindig: What happens when `hosts = []`? Is this still working as intended?

lippirk: the case `hosts = []` is converted into a none just above here. it would probably be simpler to avoid the `list option` and just match using `[] -> ... | _::_ ->` instead

ocaml/xapi/repository.ml
minglumlu: Thanks. It's great to learn about this annotation. :)
I resolved the `f` completely as I found no more usages on it.

ocaml/xapi/repository.ml
minglumlu: This requires "Since 4.10.0".  I got build failure `Unbond value List.concat_map`.  :(

PR #4308, opened 2021-01-25
XSI-828 fix snapshot metadata lookup

lippirk:


ocaml/xapi/xapi_vm_snapshot.ml
```diff
@@ -409,14 +409,28 @@ let extended_do_not_copy =
 
 let domain_type_to_string = function `pv -> "pv" | `hvm -> "hvm"
 
-(* Update the domain_type field if it does not match the HVM_boot_policy *)
+(* Update the domain_type field if it does not match the HVM boot policy *)
 let ensure_domain_type_is_consistent ~__context ~snap_metadata =
+  let hvm_boot_policy () =
+    (* XSI-828 the HVM boot policy is actually stored in the database as HVM__boot_policy
+     * we keep the check for HVM_boot_policy 'just in case' *)
```
psafont, line 416: How about a giving priority to the name with `__`?

```ocaml
List.assoc_opt "HVM__boot_policy"
|> Option.fold ~none:(List.assoc_opt "HVM_boot_policy") ~some:Fun.id
|> Option.value ~default:Constants.hvm_default_boot_policy
```

ocaml/xapi/xapi_vm_snapshot.ml
```diff
@@ -409,14 +409,28 @@ let extended_do_not_copy =
...
+    let rec go = function
+      | [] ->
+          D.info "couldn't find HVM boot policy in snap metadata, assuming '%s'"
+            Constants.hvm_default_boot_policy ;
+          Constants.hvm_default_boot_policy
```
robhoes, line 421: You can't assume this. It is the same as saying that "I don't know the domain type, so I'll assume it's HVM".  If we can't find the domain_type _and_ not even the HVM_boot_policy, which is a highly unlikely (now that you've fixed it :) ), then I think it is better to make it a hard failure.

ocaml/xapi/xapi_vm_snapshot.ml
```diff
@@ -409,14 +409,28 @@ let extended_do_not_copy =
 
 let domain_type_to_string = function `pv -> "pv" | `hvm -> "hvm"
 
-(* Update the domain_type field if it does not match the HVM_boot_policy *)
+(* Update the domain_type field if it does not match the HVM boot policy *)
 let ensure_domain_type_is_consistent ~__context ~snap_metadata =
+  let hvm_boot_policy () =
+    (* XSI-828 the HVM boot policy is actually stored in the database as HVM__boot_policy
+     * we keep the check for HVM_boot_policy 'just in case' *)
+    let if_none f x = if Option.is_none x then f () else x in
```
lindig, line 417: This is quite clever. Maybe adding a comment like "probing to find a policy, using the first one we find or fail"? I wonder whether this is an of a more general principle.

robhoes (approved):

ocaml/xapi/xapi_vm_snapshot.ml
psafont: I general I've found that composing functions around `None` or `Error ...` values is usually much more clunky than composing around `Some ...` or `Ok ...`. I think the ecosystem could provide better functions for composition around these.

psafont (approved):

lindig (approved):

PR #4309, opened 2021-01-26
CA-342527: simplify code around rbac checks

psafont:
The rbac checks around call arguments spends a lot of effort checking that two lists are always the same length. Since we control the source, we can generate a list of pairs instead. Other list traversals were avoided as well.

Special case for parsing VMPP arguments when log-auditing was removed as the code was not trivial, introduced IO and the VMPP calls were deprecated in 6.0 so the code has been unused since then.

BST + BVT 144004 is mostly green, except for a problem with RPU, xenserver-inventory is not updated after upgrading.

I couldn't find any tests for RBAC, but I would think this code would be excercised a lot in e2e tests anyway.

### DIFF
--- a/ocaml/idl/ocaml_backend/gen_server.ml
+++ b/ocaml/idl/ocaml_backend/gen_server.ml
@@ -211,7 +211,7 @@ let operation (obj : obj) (x : message) =
   let rbac_check_begin =
     if has_session_arg then
       [
-        "let arg_names = "
+        "let arg_name_params = List.combine ("
         ^ List.fold_right
             (fun arg args -> "\"" ^ arg ^ "\"::" ^ args)
             orig_string_args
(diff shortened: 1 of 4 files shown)


ocaml/idl/ocaml_backend/gen_server.ml
```diff
@@ -211,7 +211,7 @@ let operation (obj : obj) (x : message) =
   let rbac_check_begin =
     if has_session_arg then
       [
-        "let arg_names = "
```
psafont, line 214: Previously an error in this code resulted in the RBAC checks failing due to 'wrong length'. Now it will fail here, I-m not sure how the exception will manifest. It's worth mentioning that all `List.combine` usages are done inside a pattern match that enforces the number of elements in the list `__params`.

lindig (approved):
It's difficult to see what the existing code does in the first place. The changes look alright to me but again, it's difficult to see the bigger picture. Simplifying this code is the right thing to do.

lippirk (approved):

psafont:
> BST + BVT 144004 is mostly green, except for a problem with RPU, xenserver-inventory is not updated after upgrading.

undrafted since I found other builds with the same failure, it's an unrelated failure

PR #4311, opened 2021-01-27
CP-35627: Repository: host.apply_updates

minglumlu:
This commit implements API 'host.apply_updates' for repository function.

The handler of this API will apply available updates on a specified host. And
after that, it will perform necessary actions like rebooting host, restarting
toolstack etc. depending on the requests of updates applied.

### DIFF
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -3445,15 +3445,17 @@ functor
           ~self:(Helpers.get_pool ~__context)
           ~doc:"Host.apply_updates"
           ~op:`apply_updates
-          (fun () -> Local.Host.apply_updates ~__context ~self ~hash)
-
-      let restart_device_models ~__context ~self =
-        let uuid = host_uuid ~__context self in
-        info "Host.restart_device_models : host = '%s'" uuid;
-        let local_fn = Local.Host.restart_device_models ~self in
-        do_op_on ~local_fn ~__context ~host:self (fun session_id rpc ->
-            Client.Host.restart_device_models rpc session_id self)
-
+          (fun () ->
+             let reboot_host =
+               with_host_operation ~__context ~self
+                 ~doc:"Host.apply_updates" ~op:`apply_updates (fun () ->
+                     Local.Host.apply_updates ~__context ~self ~hash)
+             in
+             if reboot_host then (
+               let local_fn = Local.Host.reboot ~host:self in
+               do_op_on ~local_fn ~__context ~host:self (fun session_id rpc ->
+                   Client.Host.reboot rpc session_id self));
+             reboot_host)
     end
 
     module Host_crashdump = struct
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -2617,43 +2617,6 @@ functor
     end
 
     module Host = struct
-      (** Add to the Host's current operations, call a function and then remove from the
-        current operations. Ensure the allowed_operations are kept up to date. *)
-      let with_host_operation ~__context ~self ~doc ~op f =
-        let task_id = Ref.string_of (Context.get_task_id __context) in
-        (* CA-18377: If there's a rolling upgrade in progress, only send Miami keys across the wire. *)
-        let operation_allowed ~op =
-          false
-          || (not (Helpers.rolling_upgrade_in_progress ~__context))
-          || List.mem op Xapi_globs.host_operations_miami
-        in
-        Helpers.retry_with_global_lock ~__context ~doc (fun () ->
-            Xapi_host_helpers.assert_operation_valid ~__context ~self ~op ;
-            if operation_allowed ~op then
-              Db.Host.add_to_current_operations ~__context ~self ~key:task_id
-                ~value:op ;
-            Xapi_host_helpers.update_allowed_operations ~__context ~self) ;
-        (* Then do the action with the lock released *)
-        finally f (* Make sure to clean up at the end *) (fun () ->
-            try
-              if operation_allowed ~op then (
-                Db.Host.remove_from_current_operations ~__context ~self
-                  ~key:task_id ;
-                Helpers.Early_wakeup.broadcast
-                  (Datamodel_common._host, Ref.string_of self)
-              ) ;
-              let clustered_srs =
-                Db.SR.get_refs_where ~__context
-                  ~expr:(Eq (Field "clustered", Literal "true"))
-              in
-              if clustered_srs <> [] then
-                (* Host powerstate operations on one host may affect all other hosts if
-                 * a clustered SR is in use, so update all hosts' allowed operations. *)
-                Xapi_host_helpers.update_allowed_operations_all_hosts ~__context
-              else
-                Xapi_host_helpers.update_allowed_operations ~__context ~self
-            with _ -> ())
-
       let create ~__context ~uuid ~name_label ~name_description ~hostname
           ~address ~external_auth_type ~external_auth_service_name
           ~external_auth_configuration =
@@ -2857,7 +2820,7 @@ functor
       let shutdown ~__context ~host =
         info "Host.shutdown: host = '%s'" (host_uuid ~__context host) ;
         let local_fn = Local.Host.shutdown ~host in
-        with_host_operation ~__context ~self:host ~doc:"Host.shutdown"
+        Xapi_host_helpers.with_host_operation ~__context ~self:host ~doc:"Host.shutdown"
           ~op:`shutdown (fun () ->
             do_op_on ~local_fn ~__context ~host (fun session_id rpc ->
                 Client.Host.shutdown rpc session_id host))
@@ -2865,7 +2828,7 @@ functor
       let reboot ~__context ~host =
         info "Host.reboot: host = '%s'" (host_uuid ~__context host) ;
         let local_fn = Local.Host.reboot ~host in
-        with_host_operation ~__context ~self:host ~doc:"Host.reboot" ~op:`reboot
+        Xapi_host_helpers.with_host_operation ~__context ~self:host ~doc:"Host.reboot" ~op:`reboot
           (fun () ->
             do_op_on ~local_fn ~__context ~host (fun session_id rpc ->
                 Client.Host.reboot rpc session_id host))
@@ -2880,7 +2843,7 @@ functor
 
       let power_on ~__context ~host =
         info "Host.power_on: host = '%s'" (host_uuid ~__context host) ;
-        with_host_operation ~__context ~self:host ~doc:"Host.power_on"
+        Xapi_host_helpers.with_host_operation ~__context ~self:host ~doc:"Host.power_on"
           ~op:`power_on (fun () ->
             (* Always executed on the master *)
             Local.Host.power_on ~__context ~host)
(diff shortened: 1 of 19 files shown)


ocaml/xapi/repository.ml
```diff
@@ -1024,3 +1024,128 @@ let get_pool_updates_in_json ~__context ~hosts =
   | e ->
     error "getting updates for pool failed: %s" (ExnHelper.string_of_exn e);
     raise Api_errors.(Server_error (get_updates_failed, []))
+
+let apply ~__context ~host =
+  (* This function runs on slave host *)
+  with_local_repository ~__context (fun () ->
+      let params =
+          [
+            "-y"; "--disablerepo=*";
```
lindig, line 1033: Has this code been run through `make format`? I suspect not.

ocaml/xapi/repository.ml
```diff
@@ -1024,3 +1024,128 @@ let get_pool_updates_in_json ~__context ~hosts =
...
+
+let set4 = set_of_list [EvacuateHost; RestartToolstack; EvacuateHost]
+
+let assert_valid_guidances gs =
+  match List.map (fun x -> guidance_of_string x) gs with
+  | [RebootHost]
+  | [EvacuateHost]
+  | [RestartToolstack]
+  | [RestartDeviceModel] -> ()
+  | l when GuidanceStrSet.equal (set_of_list l) set1 -> ()
```
lindig, line 1063: There are several calls `GuidanceStrSet.equal (set_of_list l) set1 ` - so this could be abstracted and used subsequently:

```
let eq list set = GuidanceStrSet.equal (set_of_list list) set

File: ocaml/xapi/repository.ml [hunk 3]
```diff
@@ -1024,3 +1024,128 @@ let get_pool_updates_in_json ~__context ~hosts =
...
+              let immediate_guidances =
+                eval_guidances ~updates_info ~updates:l ~kind:Recommended
+              in
+              assert_valid_guidances immediate_guidances;
+              Helpers.call_api_functions ~__context (fun rpc session_id ->
+                  Client.Client.Repository.apply ~rpc ~session_id ~host);
+              (* TODO: absolute guidances *)
+              apply_immediate_guidances ~__context ~host immediate_guidances))
+  with
+  | Api_errors.(Server_error (code, _)) as e when code <> Api_errors.internal_error ->
```
lindig, line 1146: I'm a bit confused about the matching here. The second case is reached only for errors that are not `Api_errors.Server_error` because all `Api_errors.Server_error` except internal ones are re-raised in the first match.  Why are exceptions split this way?

ocaml/xapi/repository.ml
psafont: when merging to master it will have to be formatted in any case :)

minglumlu: Will format all fo these in next PR for refining the code structure.

ocaml/xapi/repository.ml
minglumlu: This is to hide `internal_error` and other non `Api_errors` from user of API.

ocaml/xapi/repository.ml
```diff
@@ -1024,3 +1024,129 @@ let get_pool_updates_in_json ~__context ~hosts =
...
+    let num_of_host = List.length (Db.Host.get_all ~__context) in
+    let open Client in
+    Helpers.call_api_functions ~__context (fun rpc session_id ->
+        match guidances with
+        | ["RebootHost"] ->
+          Client.Host.reboot ~rpc ~session_id ~host
+        | ["EvacuateHost"] -> ()
+        | ["RestartDeviceModel"] ->
+          Client.Host.restart_device_models ~rpc ~session_id ~self:host
+        | l when GuidanceStrSet.equal (GuidanceStrSet.of_list l) set1 ->
```
lindig, line 1086: `eq` from above could be used here, too

ocaml/xapi/repository.ml
```diff
@@ -1024,3 +1024,129 @@ let get_pool_updates_in_json ~__context ~hosts =
...
+       f ())
+    (fun () ->
+       Db.Host.remove_from_other_config ~__context ~self:host
+         ~key:Xapi_globs.host_in_updating)
+
+let apply_updates ~__context ~self ~host ~hash =
+  (* This function runs on master host *)
+  try
+    with_host_in_updating ~__context ~host (fun () ->
+        with_pool_repository (fun () ->
```
lindig, line 1125: Deep nesting could be avoided by:

```
with_host_in_updating ~__context ~host @@ fun () ->
with_pool_repository @@ fun () ->
```

The disadvantage is that the end of the functions would be less obvious.

lindig (approved):

ocaml/xapi/repository.ml
```diff
@@ -1024,3 +1025,132 @@ let get_pool_updates_in_json ~__context ~hosts =
...
+    Helpers.call_api_functions ~__context (fun rpc session_id ->
+        match guidances with
+        | [RebootHost] ->
+          Client.Host.reboot ~rpc ~session_id ~host
+        | [EvacuateHost] -> ()
+        | [RestartDeviceModel] ->
+          Client.Host.restart_device_models ~rpc ~session_id ~self:host
+        | [RestartToolstack] ->
+          Client.Host.restart_agent ~rpc ~session_id ~host
+        | l when eq l set1 ->
```
robhoes, line 1089: Perhaps add a comment about what the set is, to make it easier to read (below as well).
```suggestion
        | l when eq l set1 ->  (* EvacuateHost; RestartToolstack *)
```

ocaml/xapi/repository.ml
```diff
@@ -1024,3 +1025,132 @@ let get_pool_updates_in_json ~__context ~hosts =
...
+
+let apply_immediate_guidances ~__context ~host guidances =
+  try
+    let num_of_host = List.length (Db.Host.get_all ~__context) in
+    let open Client in
+    Helpers.call_api_functions ~__context (fun rpc session_id ->
+        match guidances with
+        | [RebootHost] ->
+          Client.Host.reboot ~rpc ~session_id ~host
+        | [EvacuateHost] -> ()
```
robhoes, line 1084: Please explain why we aren't doing any evacuating here.

ocaml/xapi/repository.ml
```diff
@@ -1024,3 +1025,132 @@ let get_pool_updates_in_json ~__context ~hosts =
...
+    raise Api_errors.(Server_error (internal_error, [msg]))
+
+let apply_immediate_guidances ~__context ~host guidances =
+  try
+    let num_of_host = List.length (Db.Host.get_all ~__context) in
+    let open Client in
+    Helpers.call_api_functions ~__context (fun rpc session_id ->
+        match guidances with
+        | [RebootHost] ->
+          Client.Host.reboot ~rpc ~session_id ~host
```
robhoes, line 1083: What if host = localhost (master) for reboot and restart_agent? Are we certain that there is nothing left to do after this function call? For example, we may miss the "cleanup" part of any with_ functions that we are nested inside, such as with_host_in_updating.

ocaml/xapi/xapi_pool.ml
```diff
@@ -665,9 +665,17 @@ let pre_join_checks ~__context ~rpc ~session_id ~force =
            ( Api_errors.license_restriction
            , [Features.name_of_feature Features.Pool_size] ))
   in
+  let assert_pool_master_not_in_updating () =
+    let master = get_master ~rpc ~session_id in
```
robhoes, line 669: Perhaps just block if `apply_updates` is in `pool.current_operations`. It's best not to add new hosts to the pool if anything is updating.

robhoes (changes requested):

ocaml/xapi/repository.ml
minglumlu: I think they have 10-sec delay for XAPI call to complete remaining works.
But I defer the `Host.reboot` as reboot conflicts with apply_updates.

ocaml/xapi/xapi_pool_transition.ml
```diff
@@ -201,7 +201,8 @@ let attempt_two_phase_commit_of_new_master ~__context (manual : bool)
 let become_another_masters_slave master_address =
   let new_role = Pool_role.Slave master_address in
   if Pool_role.get_role () = new_role then
-    debug "We are already a slave of %s; nothing to do" master_address
+    (Repository.cleanup_pool_repo ();
```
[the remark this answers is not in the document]
minglumlu: Will raise another PR for this.

ocaml/xapi/repository.ml
```diff
@@ -1062,42 +1062,84 @@ let assert_valid_guidances = function
...
+        None
+      | `Paused, true ->
+        error "VM 'ref=%s' is paused, can't restart its device models" (Ref.string_of ref);
+        Some ref
+      | _ ->
+        (* No device models are running for this VM *)
+        None)
+  |> function
+  | [] -> ()
+  | _ :: _ ->
```
robhoes, line 1105: You could use `Option.iter` instead.

ocaml/xapi/repository.ml
```diff
@@ -1062,42 +1062,84 @@ let assert_valid_guidances = function
...
     raise Api_errors.(Server_error (internal_error, [msg]))
 
+let restart_device_models ~__context host =
+  (* Restart device models of all running HVM VMs on the host by doing
+   * local migrations. *)
+  Db.Host.get_resident_VMs ~__context ~self:host
+  |> List.map (fun self -> (self, Db.VM.get_record ~__context ~self))
+  |> List.filter (fun (_, record) -> not record.API.vM_is_control_domain)
+  |> List.filter_map (fun (ref, _) ->
+      match Db.VM.get_power_state ~__context ~self:ref,
```
robhoes, line 1091: You've got the `power_state` in the record (like `control_domain`), so you don't need another DB call.

ocaml/xapi/repository.ml
robhoes: I think `restart_agent` at least just executes `xe-toolstack-restart` immediately.

ocaml/xapi/message_forwarding.ml
```diff
@@ -3436,6 +3436,26 @@ functor
         let local_fn = Local.Host.get_sched_gran ~self in
         do_op_on ~local_fn ~__context ~host:self (fun session_id rpc ->
             Client.Host.get_sched_gran rpc session_id self)
+
+      let apply_updates ~__context ~self ~hash =
+        let uuid = host_uuid ~__context self in
+        info "Host.apply_updates: host = '%s'; hash = '%s'" uuid hash;
+        Xapi_pool_helpers.with_pool_operation
```
robhoes, line 3443: You still don't exist `with_pool_operation` before the reboot or xapi restart. To make it easier, perhaps apply all guidance cases outside of this operation. And move this entire block of code into the Xapi_host function, so you don't need to worry about the return value of the API, and return unit as before (as it all runs on the master).

robhoes (changes requested):
Nearly there, just the timing of the restarts/reboots needs to be addressed.

ocaml/xapi/repository.ml
minglumlu: Not quite understand this :(

`val iter : ('a -> unit) -> 'a option -> unit
iter f o is f v if o is Some v and () otherwise.`

Here we have a list of `ref option`, and we aim to raise exception if the list is not empty.

ocaml/xapi/repository.ml
minglumlu: Is this still true?
https://github.com/xapi-project/xen-api/blob/master/ocaml/idl/datamodel_host.ml#L438

robhoes: I don't think so, looking at `Xapi_host.restart_agent`.

ocaml/xapi/repository.ml
robhoes: Yes, I misread this.

robhoes (approved):

PR #4312, opened 2021-01-27
CA-351391: Make certificate alerts ignore CA certs

psafont:
Currently it's assumed by the alert generator that all certificates have
a valid reference to a host. This is not true anymore as we model CA
certificates.

Ignore the error when the assumption does not hold for the time being


lippirk (approved):

lindig (approved):
It's probably safer the way it is. An alternative would be 
```
| .. when Db.is_valid_ref __context host ->
| _ -> ()
```

PR #4314, opened 2021-02-05
Two commits for fixing-up of repository funciton

minglumlu:
Two commits for fixing-up:

Fixup: Repository: Check existence before removing repo dir
    This commit adds an existence checking before removing the local pool repository
    directory. This fixes bug in cases of inital pool.set_repository and changing
    pool master.

Fixup: Repository: Cleanup local pool repo when becoming a new slave
    This commit adds cleanup when a host becoming a new slave.

### DIFF
--- a/ocaml/xapi/repository.ml
+++ b/ocaml/xapi/repository.ml
@@ -244,8 +244,7 @@ let cleanup_pool_repo () =
     clean_yum_cache !Xapi_globs.pool_repo_name;
     Unixext.unlink_safe (Filename.concat !Xapi_globs.yum_repos_config_dir
                            !Xapi_globs.pool_repo_name);
-    if Sys.file_exists !Xapi_globs.local_pool_repo_dir then
-      Helpers.rmtree !Xapi_globs.local_pool_repo_dir
+    Helpers.rmtree !Xapi_globs.local_pool_repo_dir
   with e ->
     error "Failed to cleanup pool repository: %s" (ExnHelper.string_of_exn e);
     raise Api_errors.(Server_error (repository_cleanup_failed, []))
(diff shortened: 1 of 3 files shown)


ocaml/xapi/repository.ml
```diff
@@ -244,7 +244,8 @@ let cleanup_pool_repo () =
     clean_yum_cache !Xapi_globs.pool_repo_name;
     Unixext.unlink_safe (Filename.concat !Xapi_globs.yum_repos_config_dir
                            !Xapi_globs.pool_repo_name);
-    Helpers.rmtree !Xapi_globs.local_pool_repo_dir
+    if Sys.file_exists !Xapi_globs.local_pool_repo_dir then
```
lippirk, line 247: To me it would make sense to put this check inside `Helpers.rmtree` itself

lippirk (approved):

ocaml/xapi/xapi_pool_transition.ml
```diff
@@ -206,7 +206,8 @@ let become_another_masters_slave master_address =
     debug "Setting pool.conf to point to %s" master_address ;
     set_role new_role ;
     run_external_scripts false ;
-    Xapi_fuse.light_fuse_and_run ()
+    Xapi_fuse.light_fuse_and_run () ;
```
psafont, line 209: This gives the next functions 5 seconds to complete until it is cancelled, are you sure you don't want to clean up the the repository _before_ this call?

minglumlu: Aah, I see. This `fuse` will restart the toolstack. Thanks.

robhoes (approved):

robhoes:
@minglumlu Please "autosquash" so we'll end up with two commits (github can't do this automatically, only squashing into a single commit).

lindig (approved):

PR #4316, opened 2021-02-08
CP-35373: Restructure code of repository with modules and helpers

minglumlu:
This commit re-structures the code of repository with sub modules and helpers.

### DIFF
--- a/ocaml/xapi/repository.mli
+++ b/ocaml/xapi/repository.mli
@@ -12,7 +12,7 @@
  * GNU Lesser General Public License for more details.
  *)
 
-type guidance = RebootHost | RestartToolstack | EvacuateHost | RestartDeviceModel
+open Repository_helpers
 
 val introduce :
      __context:Context.t
@@ -70,10 +70,10 @@ val apply_updates :
      __context:Context.t
   -> host:[`host] API.Ref.t
   -> hash:string
-  -> guidance list
+  -> Guidance.t list
 
 val apply_immediate_guidances :
      __context:Context.t
   -> host:[`host] API.Ref.t
-  -> guidances:guidance list
+  -> guidances:Guidance.t list
   -> unit
--- a/ocaml/xapi/repository.mli
+++ b/ocaml/xapi/repository.mli
@@ -12,8 +12,6 @@
  * GNU Lesser General Public License for more details.
  *)
 
-open Repository_helpers
-
 val introduce :
      __context:Context.t
   -> name_label:string
@@ -70,10 +68,10 @@ val apply_updates :
      __context:Context.t
   -> host:[`host] API.Ref.t
   -> hash:string
-  -> Guidance.t list
+  -> Repository_helpers.Guidance.t list
 
 val apply_immediate_guidances :
      __context:Context.t
   -> host:[`host] API.Ref.t
-  -> guidances:Guidance.t list
+  -> guidances:Repository_helpers.Guidance.t list
   -> unit
--- a/ocaml/xapi/repository_helpers.ml
+++ b/ocaml/xapi/repository_helpers.ml
@@ -1,0 +1,732 @@
+(*
+ * Copyright (C) Citrix Systems Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+module D = Debug.Make (struct let name = "repository_helpers" end)
+
+open D
+
+module Unixext = Xapi_stdext_unix.Unixext
+
+module GuidanceStrSet = Set.Make (String)
+
+let exposing_pool_repo_mutex = Mutex.create ()
+
+module Pkg = struct
+  type t = {
+      name: string
+    ; version: string
+    ; release: string
+    ; arch: string
+  }
+
+  let of_fullname s =
+    (* s I.E. "libpath-utils-0.2.1-29.el7.x86_64" *)
+    let r = Re.Str.regexp "^\\(.+\\)\\.\\(noarch\\|x86_64\\)$" in
+    match Re.Str.string_match r s 0 with
+    | true ->
+      (try
+         let pkg = Re.Str.replace_first r "\\1" s in (* pkg is, e.g. "libpath-utils-0.2.1-29.el7" *)
+         let arch = Re.Str.replace_first r "\\2" s in (* arch is "<noarch|x86_64>" *)
+         let pos1 = String.rindex pkg '-' in
+         let pos2 = String.rindex_from pkg (pos1 - 1) '-' in
+         let release = String.sub pkg (pos1 + 1) ((String.length pkg) - pos1 - 1) in
+         let version = String.sub pkg (pos2 + 1) (pos1 - pos2 - 1) in
+         let name = String.sub pkg 0 pos2 in
+         Some { name; version; release; arch }
+       with e ->
+         let msg =
+           Printf.sprintf "Failed to split rpm name '%s': %s" s (ExnHelper.string_of_exn e) in
+         raise Api_errors.(Server_error (internal_error, [msg])))
+    | false -> None
+
+  let to_ver_rel_json pkg =
+    `Assoc [
+      ("version", `String pkg.version);
+      ("release", `String pkg.release);
+    ]
+
+  let to_name_arch_string pkg =
+    pkg.name ^ "." ^ pkg.arch
+end
+
+module Update = struct
+  type t = {
+      name: string
+    ; arch: string
+    ; old_version: string option
+    ; old_release: string option
+    ; new_version: string
+    ; new_release: string
+    ; update_id: string option
+  }
+
+  let of_json j =
+    let open Yojson.Basic.Util in
+    {
+      name = (member "name" j |> to_string);
+      arch = (member "arch" j |> to_string);
+      new_version = (member "newVerRel" j |> member "version" |> to_string);
+      new_release = (member "newVerRel" j |> member "release" |> to_string);
+      old_version = (try Some (member "oldVerRel" j |> member "version" |> to_string)
+                     with _ -> None);
+      old_release = (try Some (member "oldVerRel" j |> member "release" |> to_string)
+                     with _ -> None);
+      update_id = (member "updateId" j |> to_string_option)
+    }
+end
+
+module Guidance = struct
+  type t =
+    | RebootHost
+    | RestartToolstack
+    | EvacuateHost
+    | RestartDeviceModel
+
+  type guidance_kind = Absolute | Recommended
+
+  let to_string = function
+    | RebootHost -> "RebootHost"
+    | RestartToolstack -> "RestartToolstack"
+    | EvacuateHost -> "EvacuateHost"
+    | RestartDeviceModel -> "RestartDeviceModel"
+
+  exception Unknown_guidance
+
+  let of_string = function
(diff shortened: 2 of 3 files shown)


ocaml/xapi/repository_helpers.ml
```diff
@@ -0,0 +1,732 @@
...
+      raise Api_errors.(Server_error (internal_error, [msg]))
+
+  let resort_guidances gs=
+    match GuidanceStrSet.find_opt "RebootHost" gs with
+    | Some _ -> GuidanceStrSet.singleton "RebootHost"
+    | None -> gs
+end
+
+module Applicability = struct
+  type inequality = Lt | Eq | Gt
```
lindig, line 166: A better name would be `order`

ocaml/xapi/repository_helpers.ml
```diff
@@ -0,0 +1,732 @@
...
+        if Re.Str.string_match r s 0 then
+          [(Re.Str.replace_first r "\\1" s); (Re.Str.replace_first r "\\2" s)]
+        else [s]
+      in
+      let concat_map f l = List.concat (List.map f l) in
+      v
+      |> Astring.String.cuts ~sep:"."
+      |> concat_map (fun s -> Astring.String.cuts ~sep:"_" s)
+      |> concat_map (fun s -> split_letters_and_numbers s)
+      |> List.map (fun s -> try Int (int_of_string s) with _ -> Str s)
```
lindig, line 276: Is this new code? The detection of integers is not safe:

```
utop # int_of_string "0x10";;
- : int = 16
```

ocaml/xapi/repository_helpers.ml
```diff
@@ -0,0 +1,732 @@
...
+    in
+    let rec compare_segments l1 l2 =
+      match l1, l2 with
+      | c1 :: t1, c2 :: t2 ->
+        begin match c1, c2 with
+          | Int s1, Int s2 ->
+            if s1 > s2 then Gt
+            else if s1 = s2 then compare_segments t1 t2
+            else Lt
+          | Int s1, Str s2 -> Gt
```
lindig, line 286: According to this, any number is greater than any string. Is this correct? "1" > "a". This is not explicit in the examples above.

lindig (commented):
I have no comments about the restructuring - which looks good to me. But I have doubts that the code for comparing version strings is correct.

ocaml/xapi/repository_helpers.ml
psafont: IIRC this model the version constraints, which is subtly different from ordering. This means that `Lte` and `Gte` need to be modelled. This is done currently with a list like `[Lt; Eq]`.
I would prefer if Lte and Gte were added to the variant type instead of handling lists.

ocaml/xapi/repository_helpers.ml
minglumlu: Merged into feature branch already. 
This commit re-structure the code with modules and helpers.

ocaml/xapi/repository_helpers.ml
minglumlu: Following examples are for this:
"1.0" ">" "1.a"
"2a" "<" "2.0"

psafont: Please add unit tests that prove these, it makes reviewing and maintaining it so much easier.

minglumlu: That's in progress and will be next PR.

ocaml/xapi/repository.mli
```diff
@@ -12,7 +12,7 @@
  * GNU Lesser General Public License for more details.
  *)
 
-type guidance = RebootHost | RestartToolstack | EvacuateHost | RestartDeviceModel
+open Repository_helpers
```
robhoes, line 15: I think this adds everything from the helpers into the interface, which defeats part of the purpose of having an mli file. Instead, you could change below `Guidance.t` to `Repository_helpers.Guidance.t`.

ocaml/xapi/repository_helpers.ml
minglumlu: `Lte` and `Gte` are actually useless in actual comparison. For example, `compare s1 s2 ` will return `Lt`, `Eq`, or `Gt`. It will not return `Lte` or `Gte`.

lindig: The use of lists (which represent conjunction) also seems ad-hoc to me. We are parsing a string "<=" as a version constrained and need to check whether two versions respect the constrained. It is more important to model the problem than what `compare` returns. Implementing the `<=` constraint it terms of compare is easy: `leq x y = lt x Y || eq x y`. I don't see why we need to represent what `compare` returns as a sum type.

minglumlu: Thanks. In this case, I think following should present the modeling:
`
type inequality = Lt | Eq | Gt | Lte | Gte

type order = Lt_ | Eq_ | Gt_

let lt (v1, r1) (v2, r2) ->
  match (compare v1 v2), (compare r1 r2) with
  | Lt_, _ | Eq_, Lt_ -> true
  | _ -> false

let gt (v1, r1) (v2, r2) ->
  match (compare v1 v2), (compare r1 r2) with
  | Gt_, _ | Eq_, Gt_ -> true
  | _ -> false

let eq (v1, r1) (v2, r2) ->
  match (compare v1 v2), (compare r1 r2) with
  | Eq_, Eq_ -> true
  | _ -> false

let lte vr1 vr2 = lt vr1 vr2 || eq vr1 vr2
let gte vr1 vr2 = lt vr1 vr2 || eq vr1 vr2

let eval_applicability vr1 vr2
match inequality
| Lt -> lt vr1 vr2
| Lte -> lte vr1 vr2
| Gt  -> gt vr1 vr2
| Gte -> gte vr1 vr2
| Eq -> eq vr1 vr2
`

lippirk: I think it would be better to avoid the `Invalid` variant - so that if you have a value of type `inequality`, you don't have to do an awkward match on `Invalid` (unless of course `Invalid` is genuinely a state we expect to see)

robhoes: I don't know if changing these types makes thing better. It certainly doesn't look _simpler_ to me.

lindig: I think this is a log clearer. The underscores could be avoided by using `LT`, `GT`, `EQ` for `order`.

robhoes (approved):

ocaml/xapi/repository_helpers.ml
minglumlu: > I think it would be better to avoid the Invalid variant - so that if you have a value of type inequality, you don't have to do an awkward match on Invalid (unless of course Invalid is genuinely a state we expect to see)

A default value is needed for this. Not found a better way :(

minglumlu: > An alternative name for for inequality could be constraint.

For the name, this is inherited from the tag name of updateinfo.xml.  I think it would be better to keep the consistency.

ocaml/xapi/repository_helpers.ml
```diff
@@ -0,0 +1,766 @@
...
+          [(Re.Str.replace_first r "\\1" s); (Re.Str.replace_first r "\\2" s)]
+        else [s]
+      in
+      let concat_map f l = List.concat (List.map f l) in
+      v
+      |> Astring.String.cuts ~sep:"."
+      |> concat_map (fun s -> Astring.String.cuts ~sep:"_" s)
+      |> concat_map (fun s -> split_letters_and_numbers s)
+      |> List.map (fun s ->
+          match int_of_string s with
```
lindig, line 279: I think the solution is to match against a regexp `[0-9]+` to decide if a string is a number and then pass it to `int_of_string`. If the match fails, it is a string.

ocaml/xapi/repository_helpers.ml
psafont: > A default value is needed for this. Not found a better way :(
You could use `inequality Option` instead. I think the same should be done with the rest of the fields in the records.
I see that's a pain to contruct because of how xml is parsed. Maybe you can use a "partial record" with optional fields, at the end of the xml parsing you can discard records with None fields and construct a "complete" record if you're successful.

lippirk: > A default value is needed for this. Not found a better way :(

you can use `None` for default values (from the `option` type). you should be able to write a parse function with the following signature:

```
parse_xml : string -> t option
```
where `t` represents a completely valid result (i.e. the parse function should handle invalid xml states for us, so we don't have worry about it elsewhere in the code)

minglumlu:
Hi @psafont  @lindig 
I appended a new commit to resolve the issues raised in your comments. I'm sorry for long delay for this.

ocaml/xapi/repository_helpers.ml
```diff
@@ -188,30 +188,41 @@ module Applicability = struct
...
-    | "gt"  -> Gt
-    | "lt"  -> Lt
-    | "eq"  -> Eq
-    | _ -> raise Invalid_inequality
+    | "gte" -> Some Gte
+    | "lte" -> Some Lte
+    | "gt"  -> Some Gt
+    | "lt"  -> Some Lt
+    | "eq"  -> Some Eq
+    | _ -> None
```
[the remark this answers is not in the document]
minglumlu: I don't quite understand this. You mean to remove line# 204 above?
The pattern matching in function `inequality_of_string` needs to be exhaustive.

ocaml/xapi/repository_helpers.ml
```diff
@@ -0,0 +1,778 @@
...
+
+  let string_of_inequality = function
+    | Some Lt -> "<"
+    | Some Eq -> "="
+    | Some Gt -> ">"
+    | Some Gte -> ">="
+    | Some Lte -> "<="
+    | _ -> "InvalidInequality"
+
+  let inequality_of_string = function
```
lindig, line 198: Why do we need to ever convert from strings like "gte" to an internal representation? We are parsing strings like "<=" and I understand that this need error handling. But I don't understand how strings like "gte" can arise.

ocaml/xapi/repository_helpers.ml
```diff
@@ -0,0 +1,778 @@
...
+
+  exception Invalid_inequality
+
+  let string_of_inequality = function
+    | Some Lt -> "<"
+    | Some Eq -> "="
+    | Some Gt -> ">"
+    | Some Gte -> ">="
+    | Some Lte -> "<="
+    | _ -> "InvalidInequality"
```
lippirk, line 196: This is simpler:
```suggestion
  let string_of_inequality = function
    | Lt -> "<"
    | Eq -> "="
    | Gt -> ">"
    | Gte -> ">="
    | Lte -> "<="
```

ocaml/xapi/repository_helpers.ml
lippirk: `inequality_of_string` is used in `of_xml`, which implies these strings appear in xml files we are trying to parse ... is that right @minglumlu ?

I think a unit test would really help here. We can show an example of the sort of xml file we want to be able to parse, and prove that we are parsing it correctly

lindig: If that's the case, then some error handling is justified. However, I was expecting that XML decoding would not expose us to strings like "gte" if they are coming from ">=" originally. Overall, this still looks convoluted to me but i'm happy to move on and close this PR.

ocaml/xapi/repository_helpers.ml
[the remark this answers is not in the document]
psafont: I mean none of these two functions should be using Option. If one needs to raise an exception when converting to `inequality` that is fine, it is managed by the caller.

ocaml/xapi/repository_helpers.ml
minglumlu: The value in XML is designed as "gte", "lte", "lt" or "gt".  The string like ">=" is used only for logging.
And a following PR will have all unit tests against all of these.

minglumlu: I changed the string ">=" to "gte" for consistency.

psafont (approved):

lippirk (approved):

PR #4318, opened 2021-02-12
ci: count usages of List.hd

psafont:
We've decided that usage of this function is widespread and can lead to
uncaught exceptions.

The recommendation is to match the list or use List.nth_opt list 0

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -82,3 +82,17 @@ jobs:
         run: |
           opam list | awk -F " " '$2 == "master" { print $1 }' |  xargs opam uninstall
           opam pin list | cut -f1 -d "." | xargs opam unpin
+
+  deprecation-test:
+    name: Deprecation tests
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Generate empty configuration for make to be happy
+        run: touch config.mk
+
+      - name: Number of List.hd usages has not changed
+        run: make list-hd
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -82,3 +82,22 @@ jobs:
...
+      - name: Retrieve number of usages of List.hd
+        id: list-hd
+        shell: bash
+        run: echo "::set-output name=count::$(grep -r --include=*.ml --count 'List.hd' | cut -d ':' -f 2 | paste -sd+ - | bc)"
+
+      - name: Number of List.hd usages has not changed
+        if: ${{ steps.list-hd.outputs.count != 301 }}
+        run: exit $COUNT
+        env:
+          COUNT: ${{ steps.list-hd.outputs.count }}
```
lippirk, line 103: Is there a reason we're implementing this assertion in this yaml file? Would it be simpler to have a make target instead, e.g. `make ci-assertions`?

psafont: It needs to be run in the CI if we want to use the assertion as a gate, so I thought it would be simpler if only the CI was involved.

I tried implementing it as a dune test, using diff or cram, but the dune sandboxing prevents the commands having access to all the files.

lippirk: > It needs to be run in the CI if we want to use the assertion as a gate, so I thought it would be simpler if only the CI was involved.

I don't agree here, I think I'd rather see a make target (or even just a bash/python script). Also if we would like to add a new check to the ci in the future, I'd argue that a bash script is easier to work with

lindig: Could we use `git grep ... -- **/*.ml` rather than plain grep?

lippirk (approved):

lindig (approved):

PR #4321, opened 2021-02-19
CP-36094 add SNI to stunnel server config

lindig:
Add SNI "pool" to the stunnel server configuration. This will be used by
clients to request a specific certificate. For now, this is simply
the self-signed certificate used by xapi.

This is safe to merge into xapi. Currently no client it using the SNI. A client needs
access to the public cert of the server before it makes sense to use this feature
and currently the mechanism to distribute these certificates is not yet in place.

### DIFF
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -1736,6 +1736,12 @@ end = struct
           ; "TIMEOUTclose = 1"
           ; "options = CIPHER_SERVER_PREFERENCE"
           ; "sslVersion = TLSv1.2"
+          ; ""
+          ; "# xapi connections use SNI 'pool' to request a cert"
+          ; "[pool]"
+          ; "connect = 80"
+          ; "sni = xapi:pool"
+          ; Printf.sprintf "cert = %s" cert
           ]
       in
       let len = String.length conf_contents in


ocaml/xapi/helpers.ml
```diff
@@ -1736,6 +1736,12 @@ end = struct
           ; "TIMEOUTclose = 1"
           ; "options = CIPHER_SERVER_PREFERENCE"
           ; "sslVersion = TLSv1.2"
+          ; ""
+          ; "# xapi connections use SNI 'pool' to request a cert"
+          ; "[pool]"
+          ; "connect = 80"
+          ; "sni = xapi:pool"
+          ; Printf.sprintf "cert = %s" cert
```
lippirk, line 1744: If I understand correctly, this shouldn't break master, because if a client does not specify an SNI parameter (which should always be the case right now) settings inside the [xapi] section will be used

However don't we need the [pool] section to have the same properties as above, for example the ciphers or the tls version?

ocaml/xapi/helpers.ml
```diff
@@ -1736,6 +1736,12 @@ end = struct
           ; "TIMEOUTclose = 1"
           ; "options = CIPHER_SERVER_PREFERENCE"
           ; "sslVersion = TLSv1.2"
+          ; ""
+          ; "# xapi connections use SNI 'pool' to request a cert"
+          ; "[pool]"
+          ; "connect = 80"
```
psafont, line 1742: Isn't this only used when using stunnel as a client?

lindig: All I can say is that this works and I also believe the parameter is required but I might be wrong.

ocaml/xapi/helpers.ml
lindig: No. The `xapi` in `xapi:pool` refers to the `[xapi]` section for parameters.

ocaml/xapi/helpers.ml
psafont: The manual is not very clear, but it looks like it's only used when initiating connections:
```
      connect = [HOST:]PORT                                                                                                                                   
           connect to a remote address

           If no host is specified, the host defaults to localhost.
       
           Multiple connect options are allowed in a single service section.

           If host resolves to multiple addresses and/or if multiple connect options are specified, then the remote address is chosen using a round-robin
           algorithm.
```

ocaml/xapi/helpers.ml
lippirk: Ah I see. We presumably specify `connect` a second time, because the alternative 'sni cert' may be the certificate for an HTTP server running on a different port (however in our case it's the same server)

lippirk (approved):

ocaml/xapi/helpers.ml
lippirk: Isn't `connect = x` used to say to stunnel 'please turn incoming HTTPS requests into HTTP requests, and send them to x'?

lindig: This is what we want, right? Forwarding an HTTPS request to port 80 but unencrypted.

robhoes (approved):

psafont (approved):

PR #4330, opened 2021-02-24
xapi-cli-server: Use semicolons for lists

psafont:
Now most lists are separated by "; ", using the same code. (this needs a note in the changelog as it's breaking behaviour)

Also generate an empty list when setting ipvx_allowed if the list is empty, there's code that actually depends on it being an empty list in some cases.

### DIFF
--- a/ocaml/xapi-cli-server/records.ml
+++ b/ocaml/xapi-cli-server/records.ml
@@ -14,10 +14,7 @@
 (* String-based interface to the API *)
 
 open Client
-open Db_cache (* eek! *)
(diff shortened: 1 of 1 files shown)


lippirk (approved):
I thought I had already reviewed this fairly recently?

psafont:
for the feature branch :)

lindig:
I wonder whether this has the potential to break script and XenRT in particular - which might be expecting a comma in lists.

lindig (approved):

PR #4332, opened 2021-02-24
REQ-403 CP-33822 add IP address as SAN in self-signed certs

psafont:
Make certificates generated by gencert and xapi be consistent.

Previously the management IP could be used as the CN only when a certificate was generated using gencert, but not through xapi. The IP was used only in the case where `localhost` was the hostname.

Now the CN is always the management network's IP. The hostnames are used as DNS SANs and the management IP is added to the IP SANs.


ocaml/xapi-aux/networking_info.ml
```diff
@@ -0,0 +1,117 @@
...
+let _cached_hostname = ref ""
+
+let _cached_hostname_m = Mutex.create ()
+
+let get_hostname () =
+  Mutex.execute _cached_hostname_m (fun () ->
+      ( if !_cached_hostname = "" then
+          _cached_hostname :=
+            try filter_newline (get_process_output "/bin/hostname")
+            with _ -> "unknown"
```
lippirk, line 42: I think it's possible for this `"unknown"` string to get put in the certificate (presumably that is bad)

ocaml/xapi-aux/networking_info.mli
```diff
@@ -0,0 +1,33 @@
...
+    [hostname]. The result may be cached. *)
+
+val reget_hostname : unit -> string
+(** [reget_hostname ()] returns the hostname as returned by the command
+    [hostname]. The result is never cached and the command is called every
+    time. *)
+
+exception Unexpected_address_type of string
+
+val hostnames : unit -> string list
```
lippirk, line 24: I think `dns_names` would be a better name for this function (my understanding is that we can have multiple FQDNs, but only one hostname)

psafont: sure thing.
For reference, this is how the names look like in an 8.2 certificate:
`DNS:lcy2-dt119, DNS:lcy2-dt119.xenrtcloud`

ocaml/xapi-aux/networking_info.ml
psafont: Makes sense, I'll try to use an option here.

ocaml/xapi-aux/networking_info.ml
```diff
@@ -0,0 +1,79 @@
...
+let ip_addr_of_string ip =
+  Ipaddr.of_string ip
+  |> Stdlib.Result.to_option
+  |> Option.map (function
+       | Ipaddr.V4 addr ->
+           Cstruct.of_string (Ipaddr.V4.to_octets addr)
+       | Ipaddr.V6 addr ->
+           Cstruct.of_string (Ipaddr.V6.to_octets addr))
+
+let get_management_ip_addr ~dbg =
```
lindig, line 49: Small note on the name: `_addr` is redundant, isn't it? What else would we expect behind the name?

ocaml/xapi/xapi_host.ml
```diff
@@ -1400,15 +1400,21 @@ let install_server_certificate ~__context ~host ~certificate ~private_key
...
-        (Helper_hostname.get_hostname (), [])
-    (* should never happen *)
-  in
-  Gencertlib.Selfcert.host common_name alt_names xapi_ssl_pem ;
+  let cn, ip =
+    let dbg = Context.string_of_task __context in
+    match Networking_info.get_management_ip_addr ~dbg with
+    | None ->
+        let msg =
+          "xapi_host.ml:reset_server_certificate: failed to get management IP"
```
lindig, line 1408: You could use `__LOC__`

lindig (approved):
Looking good.

ocaml/xapi-aux/networking_info.ml
psafont: This is the historical name of the function. This also reflects the name commonly used in Unix: `ip_addr`

psafont:
This includes some changes around the retrieval of the hostname to make it more robust, which I didn't anticipate but seem like a good PR to do the cleanup.

I could have gone further by cleaning up host_info but it's better to make this from growing even bigger. The software version should be independent of the host information, I'm not sure why they're tied together.

lippirk (approved):

PR #4334, opened 2021-03-01
Repository: Unit tests and bug fixes during test

minglumlu:


ocaml/tests/test_helpers.ml
```diff
@@ -347,6 +347,50 @@ module CIDRCheckers = Generic.MakeStateless (struct
...
+    `QuickAndAutoDocumented
+      [
+          ( ([], 0), Ok [] )
+        ; ( ([f], 0), Ok [] )
+        ; ( ([f], 1), Ok ["hello"] )
+        ; ( ([f], 2), Ok ["hello"] )
+        ; ( ([f; f], 1), Ok ["hello"; "hello"] )
+        ; ( ([f; f], 2), Ok ["hello"; "hello"] )
+        ; ( ([f; f], 3), Ok ["hello"; "hello"] )
+        ; ( ((List.init 64 (fun _ -> f)), 64), Ok (List.init 64 (fun _ -> "hello")) )
```
lippirk, line 384: A stronger test would be something like this (to ensure that `run_in_parallel` preserves the order of its arguments)

```suggestion
        ; ( let range_64 = List.init 64 Fun.id |> List.map string_of_int in
            ( (range_64, 64), Ok range_64 ) )
```

psafont: `let range_64 = List.init 64 string_of_int in` seems clearer to me ;)

ocaml/tests/test_repository_helpers.ml
```diff
@@ -0,0 +1,1218 @@
...
+  end
+
+  let transform input =
+    try Ok (Update.of_json input)
+    with e -> Error e
+
+  let tests =
+    `QuickAndAutoDocumented
+      [
+        (`Assoc [
```
lippirk, line 85: I think we should use a raw json string, rather than this intermediate Yojson type. `{| ... |}` comes in handy here (you don't need to escape `"`)

```
let ok_json_test_cases = [
  {| { "name": "libpath-utils", ... } |}, Update.{ name="libpath-utils", ... },
  ...
]
```

ocaml/tests/test_repository_helpers.ml
```diff
@@ -0,0 +1,1218 @@
...
+
+  let updateinfo = with_updateinfo normal
+
+  let transform input =
+    try Ok (UpdateInfoMetaData.of_xml (Xml.parse_string input))
+    with e -> Error e
+
+  let tests =
+    `QuickAndAutoDocumented
+      [ (with_repomd "",
```
lippirk, line 312: Similarly to the json tests above, I think it'd be easier to read this test if we were testing a single xml string for each test case, e.g.

```
let bad_test_cases = [
  ({|<?xml version="1.0" encoding="UTF-8"?><repomd></repomd>|}, invalid_repomd_xml ;
 ...
]
```

ocaml/xapi/repository_helpers.ml
```diff
@@ -467,8 +477,14 @@ module UpdateInfo = struct
       raise Api_errors.(Server_error (invalid_updateinfo_xml, []))
     | ui -> ui
 
-  let of_xml_file xml_file_path =
-    match Xml.parse_file xml_file_path with
+  let assert_no_dup_update_id l =
+    let comp x y = compare x.id y.id in
+    if List.length (List.sort_uniq comp l) = List.length (List.sort comp l) then l
```
lindig, line 482: Something like the code below would be more efficient. However, just keep what you have.

```
let rec has_duplicates = function
| [] -> false
| x::y::_ when x = y -> true
| x::y::z -> has_duplicates (y::z)

let assert_no_dup_update_id l = List.sort l |> has_duplicates |> not
```

ocaml/xapi/repository_helpers.ml
```diff
@@ -364,51 +370,55 @@ module UpdateInfoMetaData = struct
...
   }
 
+  let assert_valid_repomd = function
+    | { checksum = ""; _ } | { location = ""; _ } ->
+      error "Missing 'checksum' or 'location' in updateinfo in repomod.xml";
+      raise Api_errors.(Server_error (invalid_repomd_xml, []))
+    | umd -> umd
+
+  let of_xml = function
+    | Xml.Element ("repomd", _, children) ->
```
lindig, line 380: I believe this code could be made simpler using monadic error handling. For the success case, all expected information needs to be found and if any of that is missing, the fails with an error message. However, this needs some thoughts and I can't just write it down here myself.

lindig (approved):

ocaml/xapi/repository_helpers.ml
minglumlu: How about following?
`  let of_xml xml =
    let open Rresult.R.Infix in
    Ok xml
    >>= (function
        | Xml.Element ("repomd", _, children) ->
          let get_updateinfo_node = function
            | Xml.Element ("data", attrs, nodes) ->
              (match List.assoc_opt "type" attrs with
               | Some "updateinfo" -> Some nodes
               | _ -> None)
            | _ -> None
          in
          Ok (List.filter_map get_updateinfo_node children)
        | _ ->
          Error "Missing 'repomd' node")
    >>= (function
        | [l] ->
          Ok (List.fold_left (fun md n ->
              match n with
              | Xml.Element ("checksum", _, [Xml.PCData v]) ->
                {md with checksum = v}
              | Xml.Element ("location", attrs, _) ->
                (try {md with location = (List.assoc "href" attrs)}
                 with _ -> md)
              | _ -> md
            ) { checksum = ""; location = "" } l)
        | _ -> Error "Missing or multiple 'updateinfo' node(s)")
    >>= (function
        | { checksum = ""; _ } | { location = ""; _ } ->
          Error "Missing 'checksum' or 'location' in updateinfo in repomod.xml"
        | umd -> Ok umd)`

minglumlu: How about following?
``` 
 let of_xml xml =
    let open Rresult.R.Infix in
    Ok xml
    >>= (function
        | Xml.Element ("repomd", _, children) ->
          let get_updateinfo_node = function
            | Xml.Element ("data", attrs, nodes) ->
              (match List.assoc_opt "type" attrs with
               | Some "updateinfo" -> Some nodes
               | _ -> None)
            | _ -> None
          in
          Ok (List.filter_map get_updateinfo_node children)
        | _ ->
          Error "Missing 'repomd' node")
    >>= (function
        | [l] ->
          Ok (List.fold_left (fun md n ->
              match n with
              | Xml.Element ("checksum", _, [Xml.PCData v]) ->
                {md with checksum = v}
              | Xml.Element ("location", attrs, _) ->
                (try {md with location = (List.assoc "href" attrs)}
                 with _ -> md)
              | _ -> md
            ) { checksum = ""; location = "" } l)
        | _ -> Error "Missing or multiple 'updateinfo' node(s)")
    >>= (function
        | { checksum = ""; _ } | { location = ""; _ } ->
          Error "Missing 'checksum' or 'location' in updateinfo in repomod.xml"
        | umd -> Ok umd)```

File: ocaml/tests/test_repository_helpers.ml (hunk 3 above)
Comment (minglumlu) [reply ocaml/tests/test_repository_helpers.ml:312]: Thanks. Just want to make my understanding clear: do you mean following example would be better?

```
[ ({|<?xml version="1.0" encoding="UTF-8"?>
     <repomd>
     </repomd>
    |},
   Error Api_errors.(Server_error (invalid_repomd_xml, [])))
; ({|<?xml version="1.0" encoding="UTF-8"?>
     <repomd>
       <data type="primary"></data>
     </repomd>
    |},
   Error Api_errors.(Server_error (invalid_repomd_xml, [])))`
]
```
And the point is: it would be easier to read the test cases with complete XML/JSON string embedded in each case. And this overwhelms benefit of reducing the duplication of those XML/JSON strings.

Comment (lippirk) [reply ocaml/tests/test_repository_helpers.ml:312]: Yes, that's what I meant! I also mentioned this alternative:

> If there a lots of xml test string, you could put the strings in their own files

By that, I meant have a folder for each type of xml you want to test, e.g.

```
├── centaurus
│   ├── invalid_repomd_xml
│   │   ├── 1.xml
│   │   ├── 2.xml
│   │   └── 3.xml
│   ├── valid_repomd_xml
│   │   ├── 1.xml
│   │   ├── 2.xml
```

and then your ocaml tests become something like this

```
let test () =
  Sys.readdir "ocaml/tests/centaurus/invalid_repomd_xml" |> Array.map (fun fname -> Unixext.string_of_file fname |> try_parse |> assert_error) ;
  Sys.readdir "ocaml/tests/centaurus/valid_repomd_xml" |> Array.map (fun fname -> Unixext.string_of_file fname |> try_parse |> assert_ok)
```

I think I have a preference for this latter way, but it's up to you

minglumlu:
(Sorry for a mistake in rebase.)
I appended two commits for fixing up:
1. Per @lippirk 's comments, rewrote cases for bettering reading these cases;
2. fixed another bug found in unit testing

robhoes (approved):
Well done!

PR #4335, opened 2021-03-02
Maintenance: remove compilation warnings and use pci library in xs-opam

psafont:
This limits the usage of Listext and externalizes the pci library, which stabilized years ago and hasn't practically changed in years.

The listext warnings will manifest if https://github.com/xapi-project/xs-opam/pull/552 is merged before this PR.


lippirk:
I thought we already removed the PCI library in #4087, when did it get re-vendored?

lippirk (approved):

psafont:
It got revendored because the vendored library had some changes which were not available in the other repository. This made xapi to crash. The fixes were backported and published into opam as version 1.0.3.

lindig (approved):

PR #4336, opened 2021-03-02
CP-36096: Generate two certificates at startup

psafont:
Now two certificates are generated at startup: the additional
certificate will be used for host-to-host communications and contains
the host UUID as the issuer and subject's Common Name. Its validity
period and key type are the same as the other certificate: 3650 days and
RSA 2048 bits, respectively.

Drops gencert's dependency on mirage_crypto_rng as it doesn't generate
the private keys itself anymore.

### DIFF
--- a/ocaml/gencert/gencert.ml
+++ b/ocaml/gencert/gencert.ml
@@ -12,13 +12,25 @@
  * GNU Lesser General Public License for more details.
  *)
 
-(* Usage: gencert <path> *)
-
 let inventory = "/etc/xensource-inventory"
 
 module Lib = Gencertlib.Lib
 
 module D = Debug.Make (struct let name = "gencert" end)
+
+(* There are two types of certificates that can be generated, each one is
+   served depending on the SNI of the petition being received *)
+module SNI = struct
+  type t = Default | Xapi_pool
+
+  let of_string = function
+    | "default" ->
+        Some Default
+    | "xapi:pool" ->
+        Some Xapi_pool
+    | _ ->
+        None
+end
 
 let generate_cert_or_fail ~generator ~path =
   generator path ;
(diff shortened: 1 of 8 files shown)


ocaml/xapi/xapi_globs.ml
```diff
@@ -731,6 +731,10 @@ let pool_secret_path = ref (Filename.concat "/etc/xensource" "ptoken")
 (* Path to server ssl certificate *)
 let server_cert_path = ref (Filename.concat "/etc/xensource" "xapi-ssl.pem")
 
+(* Path to server certificate used for host-to-host TLS connections *)
+let server_cert_internal_path =
```
lippirk, line 735: This should be added to `essential_files`

ocaml/gencert/gencert.ml
```diff
@@ -51,11 +68,17 @@ let () =
...
+    match sni_of_string sni with
+    | Some sni ->
+        main ~dbg ~path ~sni
+    | None ->
+        D.error "SNI must be default or xapi:pool" ;
+        exit 1
+  )
   | _ ->
-      D.error "Usage: %s PATH" program_name ;
+      D.error "Usage: %s PATH [default|xapi:pool]" program_name ;
```
lippirk, line 83: I think `sni` is required, so according to 'docopt' the usage string should use round brackets `(default|xapi:pool)`

ocaml/gencert/gencert.ml
```diff
@@ -51,11 +68,17 @@ let () =
...
       D.info "file already exists at path (%s) - doing nothing" path ;
       exit 0
-  | [|_; path|] ->
-      main ~dbg ~path
+  | [|_; path; sni|] -> (
+    match sni_of_string sni with
+    | Some sni ->
+        main ~dbg ~path ~sni
+    | None ->
+        D.error "SNI must be default or xapi:pool" ;
```
lippirk, line 79:
```suggestion
        D.error "SNI must be default or xapi:pool, but got %s" sni;
```

scripts/gencert.service
```diff
@@ -1,12 +1,13 @@
...
+Description=Generate TLS certificates for xapi
 Requires=forkexecd.service
 After=forkexecd.service
 
 [Service]
 User=root
 Type=oneshot
-ExecStart=/opt/xensource/libexec/gencert /etc/xensource/xapi-ssl.pem
+ExecStart=/bin/sh -c '/opt/xensource/libexec/gencert /etc/xensource/xapi-ssl.pem default \
+                   && /opt/xensource/libexec/gencert /etc/xensource/xapi-pool-tls.pem xapi:pool'
```
lippirk, line 10: I think gencert.service runs after a toolstack restart, so this should generate 'xapi-pool-tls.pem' after an upgrade, as expected :+1:

psafont: Note that the command exits with 0 if the cert already exists, which means that xapi-pool-tls.pem will be generated even if xapi-ssl.pem exists.

lippirk: Yes, I made sure of that :)

lippirk (approved):

ocaml/gencert/gencert.ml
```diff
@@ -12,14 +12,24 @@
...
-
 let inventory = "/etc/xensource-inventory"
 
 module Lib = Gencertlib.Lib
 
 module D = Debug.Make (struct let name = "gencert" end)
 
+(* There are two types of certificates that can be generated, each one is
+   served depending on the SNI of the petition being received *)
+type sni = Default | Xapi_pool
```
lindig, line 23: This looks good. I'm not sure that an extra type is worth it or whether binding names to strings would have been enough. The next step up would be a local module:

```
module SNI = struct
  type t = ..
  let of_string = function ..
end
```

This leads to nice names where this is used. Just an idea, though.

lindig (approved):

PR #4340, opened 2021-03-05
CA-352329: Revert original formatting of lists in xapi-cli-server

psafont:
Also define functions so conventions can be easily followed

### DIFF
--- a/ocaml/xapi-cli-server/records.ml
+++ b/ocaml/xapi-cli-server/records.ml
@@ -160,9 +160,23 @@ let get_name_from_ref r =
     )
   with _ -> nid
 
-let get_list_from_refs f rs = String.concat "; " (List.map f rs)
+let get_uuid_from_ref_or_null r =
+  if r = Ref.null then
+    nullref
+  else
+    get_uuid_from_ref r
 
-let get_uuids_from_refs rs = get_list_from_refs get_uuid_from_ref rs
+let concat_with_semi = String.concat "; "
+
+(* Some lists are separated with commas and cannot be changed to preserve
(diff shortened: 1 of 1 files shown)


edwintorok (approved):
I compared this to reverting the 3 xapi-cli-server commits on master and wherever `",` was used I use the `comma` variant used, so LGTM

edwintorok:
Except maybe for roles where we might now show nothing and previously we only skipped the item that was failing:
```
       ; make_field ~name:"roles"
           ~get:(fun () ->
-            String.concat "; "
-              ( try
-                  List.map
+            try
+              map_and_concat
                 (fun self ->
                   try Client.Role.get_name_label rpc session_id self
                   with _ -> nid)
                 (Client.Subject.get_roles rpc session_id subject)
-                with _ -> []
-              ))
+            with _ -> "")

```

ocaml/xapi-cli-server/records.ml
```diff
@@ -168,10 +168,16 @@ let get_uuid_from_ref_or_null r =
 
 let concat_with_semi = String.concat "; "
 
+let concat_with_comma = String.concat ", "
```
robhoes, line 171: Let's add a comment to discourage new uses of this function, `map_and_concat_comma` and `get_uuids_from_refs_outdated` (explaining why).

robhoes: That was quick...

psafont: ;)

robhoes (approved):

PR #4341, opened 2021-03-05
CP-36099 REQ-403 modelling CA certs in the database

lippirk:
This PR adds some new fields to certs in the database. We also make sure that certs are added to the database when they are changed via the API (this means `Host.reset_server_certificate`, `Host.install_server_certificate` and `Pool.install_ca_certificate`)

### DIFF
--- a/ocaml/idl/datamodel_certificate.ml
+++ b/ocaml/idl/datamodel_certificate.ml
@@ -18,6 +18,13 @@ open Datamodel_roles
 
 let lifecycle = [Published, rel_stockholm, ""]
 
+let certificate_type =
+  Enum ( "certificate_type",
+         [ "ca", "Certificate that is trusted by the whole pool"
+         ; "host", "Certificate that identifies a single host"
+         ]
+       )
+
 let t =
   create_obj
     ~name: _certificate
@@ -32,6 +39,16 @@ let t =
     ~messages_default_allowed_roles:_R_READ_ONLY
     ~contents:
       [ uid   _certificate ~lifecycle
+      ; field ~qualifier:DynamicRO ~lifecycle:[Published, rel_next, ""]
+          (* Any cert records existing before the introduction of the 'name' field
+           * have no name, because they are of type 'host' *)
+          ~ty:String "name" ~default_value:(Some (VString ""))
+          "The name of the certificate, only present on certificates of type 'ca'"
+      ; field ~qualifier:DynamicRO ~lifecycle:[Published, rel_next, ""]
+          (* Any cert records existing before the introduction of the 'host' field
+           * are of type host *)
+          ~ty:certificate_type "type"  ~default_value:(Some (VEnum "host"))
+          "The type of the certificate, either 'ca' or 'host'"
       ; field   ~qualifier:StaticRO ~lifecycle
           ~ty:(Ref _host) "host" ~default_value:(Some (VRef null_ref))
           "The host where the certificate is installed"
--- a/ocaml/xapi/certificates.ml
+++ b/ocaml/xapi/certificates.ml
@@ -19,6 +19,8 @@ module D = Debug.Make (struct let name = "certificates" end)
 
 open D
 
+let () = Mirage_crypto_rng_unix.initialize ()
+
 (* Certificate locations:
   * a) stunnel external             = /etc/xensource/xapi-ssl.pem
   * b) stunnel SNI (internal)       = /etc/xensource/xapi-pool-tls.pem
--- a/ocaml/xapi/certificates.ml
+++ b/ocaml/xapi/certificates.ml
@@ -19,6 +19,16 @@ module D = Debug.Make (struct let name = "certificates" end)
 
 open D
 
+(* Certificate locations:
+  * a) stunnel external             = /etc/xensource/xapi-ssl.pem
+  * b) stunnel SNI (internal)       = /etc/xensource/xapi-pool-tls.pem
+  * c) user trusted cert folder     = /etc/stunnel/certs/
+  * d) internal trusted cert folder = /etc/stunnel/certs-pool/
+  * e) appliance trusted bundle     = /etc/stunnel/xapi-stunnel-ca-bundle.pem
+  * f) host-in-pool trusted bundle  = /etc/stunnel/xapi-pool-ca-bundle.pem
+  *
+  * Note that the bundles (e) and (f) are generated automatically using the contents of (c) and (d) respectively *)
+
 type t_trusted = CA_Certificate | CRL
 
 let c_rehash = "/usr/bin/c_rehash"
--- a/ocaml/xapi/certificates.ml
+++ b/ocaml/xapi/certificates.ml
@@ -39,6 +39,15 @@ let pem_certificate_footer = "-----END CERTIFICATE-----"
 
 let ca_certificates_path = "/etc/stunnel/certs"
 
+let pem_of_string x =
+  match Cstruct.of_string x |> X509.Certificate.decode_pem with
+  | Error _ ->
+      D.error "pem_of_string: failed to parse certificate string" ;
+      raise
+        Api_errors.(Server_error (invalid_value, ["certificate"; "<omitted>"]))
+  | Ok x ->
+      x
+
 let library_path = function
   | CA_Certificate ->
       ca_certificates_path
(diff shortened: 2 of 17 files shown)


ocaml/xapi/xapi_host.ml
```diff
@@ -1381,31 +1468,21 @@ let install_server_certificate ~__context ~host ~certificate ~private_key
...
-  Db.Certificate.create ~__context ~ref ~uuid ~host ~not_before ~not_after
-    ~fingerprint ;
+  let (_ : API.ref_Certificate) =
+    add_cert_to_db ~__context ~host ~_type:`host certificate
+  in
   let task = Context.get_task_id __context in
   (* mark task as done *)
   Db.Task.set_progress ~__context ~self:task ~value:1.0 ;
   (* sever all connections done with stunnel *)
   Xapi_mgmt_iface.reconfigure_stunnel ~__context
```
lippirk, line 1481: Do we want to uninstall old ```host`` certificates at the end of `install_server_certificate`?

ocaml/xapi/xapi_host.ml
```diff
@@ -1381,31 +1468,21 @@ let install_server_certificate ~__context ~host ~certificate ~private_key
...
 
 let reset_server_certificate ~__context ~host =
```
lippirk, line 1483: Here we've changed `reset_server_certificate` to add the new cert to the database, and remove any old ones. However this means that `emergency_reset_server_certificates` requires a database connection, and therefore won't necessarily work in 'emergency situations' (i.e. slave can't connect to the master, and therefore doesn't have a database connection!)

Perhaps `emergency_reset_server_certificate` should create a new cert, without adding it to the database; then on xapi start up we would need to do a sync of the certs on the file system with the database

ocaml/xapi/xapi_host.ml
```diff
@@ -1340,11 +1340,98 @@ let get_uncooperative_resident_VMs ~__context ~self = []
...
+     | `ca ->
+         let name = record.API.certificate_name in
+         Certificates.(host_uninstall CA_Certificate name)
+     ) ;
+     Db.Certificate.destroy ~__context ~self
+
 let install_ca_certificate ~__context ~host ~name ~cert =
-  Certificates.(host_install CA_Certificate name cert)
+  let certificate = pem_of_string cert in
+  assert_valid_ca certificate ;
```
lippirk, line 1426: Whilst not strictly necessary, I thought it would be nice to stop users accidentally installing a non-CA cert here

We could add a similar check that CA=false in the 'Basic Constraints' x509 extension, when we install a ```host`` cert

ocaml/idl/datamodel_certificate.ml
```diff
@@ -32,6 +39,12 @@ let t =
     ~messages_default_allowed_roles:_R_READ_ONLY
     ~contents:
       [ uid   _certificate ~lifecycle
+      ; field ~qualifier:DynamicRO ~lifecycle:[Published, rel_next, ""]
+          ~ty:String "name"
+          "The name of the certificate, only present on certificates of type 'ca'"
+      ; field ~qualifier:DynamicRO ~lifecycle:[Published, rel_next, ""]
```
lippirk, line 45: I would guess that existing certs should have type ```ca`` by default? Though I'm not sure what name we would give them (perhaps just their uuid)

ocaml/xapi/xapi_host.ml
```diff
@@ -1340,11 +1340,98 @@ let get_uncooperative_resident_VMs ~__context ~self = []
...
+         Certificates.(pool_uninstall CA_Certificate) ~__context ~name
+     | `ca ->
+         let name = record.API.certificate_name in
+         Certificates.(host_uninstall CA_Certificate name)
+     ) ;
+     Db.Certificate.destroy ~__context ~self
+
 let install_ca_certificate ~__context ~host ~name ~cert =
-  Certificates.(host_install CA_Certificate name cert)
+  let certificate = pem_of_string cert in
```
lippirk, line 1425: I believe this is a valid assumption: user certs ought to be in the pem format

ocaml/xapi/xapi_host.ml
psafont: yes, the database should reflect the filesystem. since there can only be one server certificate per host, the old should be removed.

ocaml/idl/datamodel_certificate.ml
psafont: Existint certs in the In 8.2 Only host certificates are present in the patabase they should be of type `host

lindig:
Why are we not storing the path for a certificate in the database and instead rely on conventions? Certificates are not all in the same place and this suggests to me the path would be a useful information.

ocaml/xapi/xapi_host.ml
psafont: I believe restricting the trusted certificates that users can install is out of scope for the feature, this should be part of a future enhancement.

In any case, right now self-signed certs without basic constraints are allowed to be installed so I'm wary to break currently-working configurations.

ocaml/xapi/xapi_host.ml
psafont: Absolutely.

ocaml/xapi/xapi_host.ml
lippirk: Fair enough, I will remove this assertion

ocaml/xapi/xapi_host.ml
```diff
@@ -1340,10 +1340,98 @@ let get_uncooperative_resident_VMs ~__context ~self = []
...
+  List.iter
+    (fun (self, record) ->
+      ( match record.API.certificate_type with
+      | `host ->
+          let name = record.API.certificate_uuid in
+          Certificates.(pool_uninstall CA_Certificate) ~__context ~name
+      | `ca ->
+          let name = record.API.certificate_name in
+          Certificates.(host_uninstall CA_Certificate name)
+      ) ;
```
psafont, line 1420: Replacing a user-installed host certificate does not involve CA certificates anymore. The code should be deleted.

ocaml/xapi/xapi_host.ml
```diff
@@ -1340,10 +1340,98 @@ let get_uncooperative_resident_VMs ~__context ~self = []
...
+      Db.Certificate.destroy ~__context ~self)
+    old_certs ;
+  Xapi_mgmt_iface.reconfigure_stunnel ~__context
+
 let install_ca_certificate ~__context ~host ~name ~cert =
-  Certificates.(host_install CA_Certificate name cert)
+  let certificate = pem_of_string cert in
+  Certificates.(host_install CA_Certificate name cert) ;
+  let (_ : API.ref_Certificate) =
```
psafont, line 1428: This call must not modify the database, only pool.install_ca_certificate does when the cert is present in all the hosts' filesystems

psafont (changes requested):
Aside from the two issues raised, xapi_pool has to be modified to correctly fill in the information in the database for ca certificates.

ocaml/xapi/certificates.ml
```diff
@@ -143,6 +153,86 @@ let raise_corrupt kind n =
...
+    let self =
+      match Db.Certificate.get_refs_where ~__context ~expr with
+      | [x] ->
+          x
+      | xs ->
+          let ref_str =
+            xs |> List.map Ref.short_string_of |> String.concat ", "
+          in
+          D.error
+            "expected 1 certificate with name='%s', but found several: [ %s ]"
```
psafont, line 229: This case also happens when the list is empty, not just when there's more than one.

The invalid value is a good error the return in that case.

If there is more than one returned I would say that's an internal error and the user should contact support.

ocaml/xapi/xapi_host.ml
```diff
@@ -1340,10 +1340,34 @@ let get_uncooperative_resident_VMs ~__context ~self = []
 (* Dummy implementation for a deprecated API method. *)
 let get_uncooperative_domains ~__context ~self = []
 
+let replace_host_certificate ~__context ~host ~complete_task
+    (write_cert_fs : unit -> X509.Certificate.t) : unit =
+  (* NOT atomic
```
psafont, line 1345: Do we want a lock of some type to make it atomic?

ocaml/xapi/xapi_host.ml
```diff
@@ -1413,19 +1411,18 @@ let reset_server_certificate ~__context ~host =
...
-
-let emergency_reset_server_certificate ~__context =
-  let host = Helpers.get_localhost ~__context in
-  reset_server_certificate ~__context ~host
+  Gencertlib.Selfcert.host ~name ~dns_names ~ips xapi_ssl_pem
+
+let reset_server_certificate ~__context ~host =
+  let dbg = Context.string_of_task __context in
+  let write_cert_fs () = _new_host_cert ~dbg in
+  replace_host_certificate ~__context ~host ~complete_task:false write_cert_fs
```
psafont, line 1419: Thinking of it, now that resetting the server certificate is not local, this should be set to 'true' and add a tolerate_connection_loss  in the corresponding function in message_forwarding.

ocaml/xapi/xapi_host.ml
lippirk: With this comment I meant that if we write a cert to the file system but the database writes fail then we'll get into a dodgy state. However now that I removed the code which does `pool_uninstall` this is probably less of a concern (but could still happen)

Nonetheless a lock would be good in case two api users try to modify the certs at the same time

ocaml/xapi/xapi_host.ml
lippirk: This requires removing the ```Session`` flag from the call in datamodel_host.ml - I'm not sure why that flag was there in the first place, possibly because it was copied from `emergency_reset_server_certificate`

ocaml/xapi/certificates.ml
```diff
@@ -19,6 +19,16 @@ module D = Debug.Make (struct let name = "certificates" end)
...
 
+let () = Mirage_crypto_rng_unix.initialize ()
+
+(* Certificate locations:
+  * a) stunnel external             = /etc/xensource/xapi-ssl.pem
+  * b) stunnel SNI (internal)       = /etc/xensource/xapi-pool-tls.pem
+  * c) user trusted cert folder     = /etc/stunnel/certs/
+  * d) internal trusted cert folder = /etc/stunnel/certs-pool/
+  * e) appliance trusted bundle     = /etc/stunnel/xapi-stunnel-ca-bundle.pem
+  * f) host-in-pool trusted bundle  = /etc/stunnel/xapi-pool-ca-bundle.pem *)
```
lindig, line 30: e) and f) are created by a script from c) and d). Are we calling this script (update-ca-bundle.sh) whenever we add or remove certs?

ocaml/xapi/message_forwarding.ml
```diff
@@ -3259,25 +3259,25 @@ functor
...
-          let progress = Db.Task.get_progress ~__context ~self:task in
-          if progress = 1. then Some () else None
-        in
         let fn () =
           do_op_on ~local_fn ~__context ~host (fun session_id rpc ->
               Client.Host.install_server_certificate rpc session_id host
                 certificate private_key certificate_chain)
         in
-        tolerate_connection_loss fn success 30. ;
+        tolerate_connection_loss fn (_success ~__context) 30. ;
```
psafont, line 3280: Note to myself: I believe the code following this expression is obsolete. (It used to clear the alert regarding certificate expiry, CA-342551)

ocaml/xapi/certificates.ml
psafont: e) is called by existing API calls installing CA certificates in `ocaml/xapi/certificates`
f) is not done yet and it must happen on pool join and on first-boot (startup)

psafont:
@robhoes could you double-check the changes in the datamodel?

ocaml/xapi/xapi_host.ml
```diff
@@ -1418,19 +1413,18 @@ let reset_server_certificate ~__context ~host =
...
-  let host = Helpers.get_localhost ~__context in
-  reset_server_certificate ~__context ~host
+  Gencertlib.Selfcert.host ~name ~dns_names ~ips xapi_ssl_pem
+
+let reset_server_certificate ~__context ~host =
+  let dbg = Context.string_of_task __context in
+  let write_cert_fs () = _new_host_cert ~dbg in
+  replace_host_certificate ~__context ~host write_cert_fs
+
+let emergency_reset_server_certificate ~(__context : 'a) =
```
robhoes, line 1423: Does this emergency call still work on a pool member if the master connection is broken due to problems with the certificate?

ocaml/xapi/certificates.ml
lippirk: I'll mention in the comment that these are auto-generated (however I think actually adding that functionality is out of scope for this PR)

robhoes:
The datamodel changes look fine.

ocaml/xapi/xapi_host.ml
lippirk: I believe so, I just tried it out (Host.reset_server_certificate is of course expected to fail, but the emergency call still works)

psafont: It does not modify the database, this means that when using the call the filesystem and the database will be out of sync. There is work incoming to synchronize them on xapi startup.

The call affect the certificate offered to external API clients, and therefore it cannot break pool operations.

ocaml/xapi/xapi_host.ml
```diff
@@ -1361,48 +1363,46 @@ let certificate_sync ~__context ~host = Certificates.local_sync ()
 let get_server_certificate ~__context ~host =
   Certificates.get_server_certificate ()
 
+let with_cert_lock : (unit -> 'a) -> 'a =
+  let cert_m = Mutex.create () in
```
psafont, line 1367: Shouldn't `cert_m` be bound outside of `with_cert_lock` to ensure it's shared in all calls of `with_cert_lock`?

lippirk: This makes a closure, so it should be shared. This is similar, but you can easily see that it works in utop

```
let with_ctr : unit -> unit =
 let r = ref 0 in
 fun () -> Printf.printf "%i" !r; r := !r + 1
```

psafont (approved):

lindig (approved):

PR #4342, opened 2021-03-08
Initialise Mirage RNG on startup

lindig:
The Mirage crypto modules use a random number generator and it needs to
be initialised. We do this with a side effect during startup: the code
is initialised as the part of the Selfcert module when it initialises
its global values.


robhoes (approved):

edwintorok (approved):
This is apparently needed by the RSA blinding operation used during signing (it needs to generate some ephemeral keys)

PR #4346, opened 2021-03-11
CP-36113: Imported the SDK repo

kc284:
The last two commits (8cc7111 and 1e8b0ab) contain new, unreviewed code; everything else was reviewed in the SDK repo.


robhoes:
Welcome back SDK!

psafont:
Nice! could you add the `xen-api-sdk` package to the tests? https://github.com/xapi-project/xen-api/blob/master/.github/workflows/main.yml#L28

Makefile
```diff
@@ -32,6 +33,32 @@ doc:
...
+	mkdir -p $(XAPISDK)/powershell
+	mkdir -p $(XAPISDK)/python/samples
+	cp -r _build/default/ocaml/sdk-gen/c/autogen/* $(XAPISDK)/c
+	cp -r _build/default/ocaml/sdk-gen/csharp/autogen/* $(XAPISDK)/csharp
+	cp -r _build/default/ocaml/sdk-gen/java/autogen/* $(XAPISDK)/java
+	cp -r _build/default/ocaml/sdk-gen/powershell/autogen/* $(XAPISDK)/powershell
+	cp -r ocaml/sdk-gen/python/samples/*.py $(XAPISDK)/python/samples
+	cp scripts/examples/python/XenAPI/XenAPI.py $(XAPISDK)/python
+	sh ocaml/sdk-gen/windows-line-endings.sh $(XAPISDK)/csharp
+	sh ocaml/sdk-gen/windows-line-endings.sh $(XAPISDK)/powershell
```
robhoes, line 60: This looks like something we eventually could get dune to do? I should be possible to get dune to automatically put the installation targets into the right location under `_build/install`.

kc284: The install stanza does not seem to support multiple files/directories yet, this issue is still open: https://github.com/ocaml/dune/issues/1868. I also tried the sites stanza, but it needs a higher dune version. Any other suggestions?

robhoes: Hmm... I don't think its the install stanza you need here, which is for installing files into the wider filesystem. The build command puts the build targets under `_build/install`. So I think these files need to be marked as targets somehow (I'm probably getting the terminology wrong). I'll take a look as well.

robhoes: Oh, I see now that it is indeed the install "stanza" that you need, which is implicit when your use `public_name` in `executable` and things like that.

robhoes: Anyway, let's leave it like this for now (we'll have something similar for `make doc`).

robhoes (approved):

psafont:
The first check complains about formatting. This can be fixed by installing ocamlformat using `opam install ocmalformat` in an opam switch with xs-opam as default repo and running make format.

I could make the change and push it, if you like.

psafont (approved):
The DCO failing it's unfortunate but I would rather merge the history without changing it.

robhoes:
Yes, we'll ignore the DCO check for this kind of thing.

PR #4349, opened 2021-03-11
Support to enable multiple repositories

minglumlu:
This series update repository functions to support multiple repositories being enabled at same time. But in these enabled repositories, there should be one and only one update repository.
If a repository is an update repository, XAPI will expect that it has updateinfo.xml and will parse it to get update metadata. A new bool field "update" is added into repository datamodel. If it is true, this repository is an update repository.

### DIFF
--- a/ocaml/idl/datamodel_errors.ml
+++ b/ocaml/idl/datamodel_errors.ml
@@ -1203,8 +1203,8 @@ let _ =
     ~doc:"The current version of Xen or its control libraries is incompatible with the Toolstack." ();
 
   (* repository and updates errors *)
-  error Api_errors.set_repository_in_progress []
-    ~doc:"The operation could not be performed because a repository setup is in progress." ();
+  error Api_errors.configure_repositories_in_progress []
+    ~doc:"The operation could not be performed because other repository(ies) is(are) already being configured." ();
   error Api_errors.invalid_base_url ["url"]
     ~doc:"The base url in the repository is invalid." ();
   error Api_errors.repository_already_exists ["ref"]
@@ -1217,6 +1217,8 @@ let _ =
     ~doc:"Failed to clean up local repository on master." ();
   error Api_errors.no_repository_enabled []
     ~doc:"There is no repository being enabled." ();
+  error Api_errors.multiple_update_repositories_enabled []
+    ~doc:"There is more than one update repository being enabled." ();
   error Api_errors.sync_updates_in_progress []
     ~doc:"The operation could not be performed because syncing updates is in progress." ();
   error Api_errors.reposync_failed []
--- a/ocaml/idl/datamodel_pool.ml
+++ b/ocaml/idl/datamodel_pool.ml
@@ -861,6 +861,6 @@ open Datamodel_types
          ; field ~in_product_since:rel_quebec ~qualifier:RW ~ty:String ~default_value:(Some (VString "")) "uefi_certificates" "The UEFI certificates allowing Secure Boot"
          ; field ~in_product_since:rel_stockholm_psr ~qualifier:RW ~ty:Bool ~default_value:(Some (VBool false)) "is_psr_pending" "True if either a PSR is running or we are waiting for a PSR to be re-run"
          ; field ~qualifier:DynamicRO ~in_product_since:rel_next ~lifecycle:[Published, rel_next, ""] ~ty:Bool ~default_value:(Some (VBool false)) "tls_verification_enabled" "True iff TLS certificate verification is enabled"
-         ; field ~in_product_since:rel_next ~qualifier:DynamicRO ~ty:(Set (Ref _repository)) "repositories" "The set of currently enabled repositories"
+         ; field ~in_product_since:rel_next ~qualifier:DynamicRO ~ty:(Set (Ref _repository)) ~ignore_foreign_key:true "repositories" ~default_value:(Some (VSet [])) "The set of currently enabled repositories"
          ])
       ()
--- a/ocaml/idl/datamodel_pool.ml
+++ b/ocaml/idl/datamodel_pool.ml
@@ -8,7 +8,7 @@ open Datamodel_types
             "ha_disable", "Indicates this pool is in the process of disabling HA";
             "cluster_create", "Indicates this pool is in the process of creating a cluster";
             "designate_new_master", "Indicates this pool is in the process of changing master";
-            "set_repository", "Indicates this pool is in the process of configuring repository";
+            "configure_repositories", "Indicates this pool is in the process of configuring repositories";
             "sync_updates", "Indicates this pool is in the process of syncing updates";
             "get_updates", "Indicates this pool is in the process of getting updates";
             "apply_updates", "Indicates this pool is in the process of applying updates";
@@ -688,13 +688,35 @@ open Datamodel_types
     ~allowed_roles:_R_POOL_ADMIN
     ()
 
-  let set_repository = call
-      ~name:"set_repository"
+  let set_repositories = call
+      ~name:"set_repositories"
       ~in_product_since:rel_next
-      ~doc:"Set the enabled repository for updates"
+      ~doc:"Set enabled set of repositories"
       ~params:[
         Ref _pool, "self", "The pool";
-        Ref _repository, "value", "The repository to be enabled"
+        Set (Ref _repository), "value", "The set of repositories to be enabled"
+      ]
+      ~allowed_roles:_R_POOL_ADMIN
+      ()
+
+  let add_repository = call
+      ~name:"add_repository"
+      ~in_product_since:rel_next
+      ~doc:"Add a repository to the enabled set"
+      ~params:[
+        Ref _pool, "self", "The pool";
+        Ref _repository, "value", "The repository to be added to the enabled set"
+      ]
+      ~allowed_roles:_R_POOL_ADMIN
+      ()
+
+  let remove_repository = call
+      ~name:"remove_repository"
+      ~in_product_since:rel_next
+      ~doc:"Remove a repository from the enabled set"
+      ~params:[
+        Ref _pool, "self", "The pool";
+        Ref _repository, "value", "The repository to be removed"
       ]
       ~allowed_roles:_R_POOL_ADMIN
       ()
@@ -788,7 +810,9 @@ open Datamodel_types
         ; add_to_guest_agent_config
         ; remove_from_guest_agent_config
         ; rotate_secret
-        ; set_repository
+        ; set_repositories
+        ; add_repository
+        ; remove_repository
         ; sync_updates
         ]
       ~contents:
@@ -837,6 +861,6 @@ open Datamodel_types
          ; field ~in_product_since:rel_quebec ~qualifier:RW ~ty:String ~default_value:(Some (VString "")) "uefi_certificates" "The UEFI certificates allowing Secure Boot"
          ; field ~in_product_since:rel_stockholm_psr ~qualifier:RW ~ty:Bool ~default_value:(Some (VBool false)) "is_psr_pending" "True if either a PSR is running or we are waiting for a PSR to be re-run"
          ; field ~qualifier:DynamicRO ~in_product_since:rel_next ~lifecycle:[Published, rel_next, ""] ~ty:Bool ~default_value:(Some (VBool false)) "tls_verification_enabled" "True iff TLS certificate verification is enabled"
-         ; field ~in_product_since:rel_next ~qualifier:DynamicRO ~ty:(Ref _repository) ~default_value:(Some (VRef null_ref)) "repository" "The enabled repository for updates"
+         ; field ~in_product_since:rel_next ~qualifier:DynamicRO ~ty:(Set (Ref _repository)) "repositories" "The set of currently enabled repositories"
          ])
       ()
--- a/ocaml/idl/datamodel_repository.ml
+++ b/ocaml/idl/datamodel_repository.ml
@@ -70,6 +70,9 @@ let t =
     ~contents:
       [ uid       _repository ~lifecycle;
         namespace ~name:"name" ~contents:(names ~writer_roles:_R_POOL_OP None RW) ();
+        field   ~qualifier:DynamicRO ~lifecycle
+          ~ty:(Set (Ref _pool)) "enabled_on"
+          "The pools where the repository is enabled on";
         field     ~qualifier:StaticRO ~lifecycle
           ~ty:String
           ~default_value:(Some (VString ""))
--- a/ocaml/idl/datamodel_repository.ml
+++ b/ocaml/idl/datamodel_repository.ml
@@ -28,6 +28,7 @@ let introduce = call
       String, "name_description", "The description of the repository";
       String, "binary_url", "Base URL of binary packages in this repository";
       String, "source_url", "Base URL of source packages in this repository";
+      Bool, "update", "True if the repository is an update repository. This means that updateinfo.xml will be parsed";
     ]
     ~result:(Ref _repository, "The ref of the created repository record.")
     ~allowed_roles:_R_POOL_OP
@@ -70,9 +71,6 @@ let t =
     ~contents:
       [ uid       _repository ~lifecycle;
         namespace ~name:"name" ~contents:(names ~writer_roles:_R_POOL_OP None RW) ();
-        field   ~qualifier:DynamicRO ~lifecycle
-          ~ty:(Set (Ref _pool)) "enabled_on"
-          "The pools where the repository is enabled on";
         field     ~qualifier:StaticRO ~lifecycle
           ~ty:String
           ~default_value:(Some (VString ""))
@@ -83,11 +81,16 @@ let t =
           ~default_value:(Some (VString ""))
           "source_url"
           "Base URL of source packages in this repository";
+        field     ~qualifier:StaticRO ~lifecycle
+           ~ty:Bool
+           ~default_value:(Some (VBool false))
+           "update"
+           "True if updateinfo.xml in this repository needs to be parsed";
         field     ~qualifier:DynamicRO ~lifecycle
           ~ty:String
           ~default_value:(Some (VString ""))
           "hash"
-          "SHA256 checksum of latest updateinfo.xml.gz in this repository";
+          "SHA256 checksum of latest updateinfo.xml.gz in this repository if its 'update' is true";
         field     ~qualifier:DynamicRO ~lifecycle
           ~ty:Bool
           ~default_value:(Some (VBool false))
--- a/ocaml/idl/schematest.ml
+++ b/ocaml/idl/schematest.ml
@@ -1,7 +1,7 @@
 let hash x = Digest.string x |> Digest.to_hex
 
 (* BEWARE: if this changes, check that schema has been bumped accordingly *)
-let last_known_schema_hash = "5115945f27703c84a08dcc94a10469e5"
+let last_known_schema_hash = "ddaf771dae1e25c706d6bd86d6e264a5"
 
 let current_schema_hash : string =
   let open Datamodel_types in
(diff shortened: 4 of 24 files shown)


ocaml/xapi/repository_helpers.ml
```diff
@@ -684,27 +698,56 @@ let get_cachedir repo_name =
...
+              write_yum_config url repo_name ;
+              clean_yum_cache repo_name ;
+              let config_params =
+                [
+                  "--save"
+                ; Printf.sprintf "--setopt=%s.sslverify=false" repo_name
+                ; repo_name
+                ]
+              in
+              ignore (Helpers.call_script !Xapi_globs.yum_config_manager_cmd config_params) ;
```
lindig, line 739: If this fails with an exception, what happens?

lindig (commented):
Mostly small stuff; looking good.

ocaml/idl/datamodel_pool.ml
```diff
@@ -788,7 +810,9 @@ open Datamodel_types
         ; add_to_guest_agent_config
         ; remove_from_guest_agent_config
         ; rotate_secret
-        ; set_repository
+        ; set_repositories
+        ; add_repositories
+        ; remove_repositories
```
lippirk, line 815: I think these functions should be called `add_repository` and `remove_repository` respectively, since they only deal with a single repository

ocaml/idl/datamodel_pool.ml
```diff
@@ -688,13 +688,35 @@ open Datamodel_types
...
-        Ref _repository, "value", "The repository to be enabled"
+        Set (Ref _repository), "value", "The set of repositories to be enabled"
+      ]
+      ~allowed_roles:_R_POOL_ADMIN
+      ()
+
+  let add_repositories = call
+      ~name:"add_repositories"
+      ~in_product_since:rel_next
+      ~doc:"Add a repository into the enabled set"
```
lippirk, line 705:
```suggestion
      ~doc:"Add a repository to the enabled set"
```

ocaml/idl/datamodel_pool.ml
```diff
@@ -688,13 +688,35 @@ open Datamodel_types
...
+      ~params:[
+        Ref _pool, "self", "The pool";
+        Ref _repository, "value", "The repository to be added into the enabled set"
```
lippirk, line 708:
```suggestion
        Ref _repository, "value", "The repository to be added to the enabled set"
```

ocaml/idl/schematest.ml
```diff
@@ -1,7 +1,7 @@
 let hash x = Digest.string x |> Digest.to_hex
 
 (* BEWARE: if this changes, check that schema has been bumped accordingly *)
-let last_known_schema_hash = "5115945f27703c84a08dcc94a10469e5"
+let last_known_schema_hash = "862ad9d640be8f53ec8c1b3085b83332"
```
lippirk, line 4: Since the schema has changed, the schema version should be bumped (`grep -re 'let schema_.*_vsn'
`)

ocaml/tests/test_repository_helpers.ml
```diff
@@ -1304,7 +1322,8 @@ module GetRpmUpdateInJson = Generic.MakeStateless (struct
             ),
             "libpath-utils.noarch    0.2.2-1.el7      epel" (* repository name is "epel" *)
           ),
-          None
+          Error Api_errors.(Server_error(
+              internal_error, ["Found update from un-managed repository"]))
```
lippirk, line 1326:
```suggestion
              internal_error, ["Found update from unmanaged repository"]))
```

if the repository is invalid, maybe you want to use an `invalid_value` error, rather than internal error?

ocaml/idl/datamodel_repository.ml
```diff
@@ -28,6 +28,7 @@ let introduce = call
       String, "name_description", "The description of the repository";
       String, "binary_url", "Base URL of binary packages in this repository";
       String, "source_url", "Base URL of source packages in this repository";
+      Bool, "update", "True if updateinfo.xml in this repository should be parsed";
```
lippirk, line 31:
```suggestion
      Bool, "parse_updateinfo", "True if updateinfo.xml in this repository should be parsed";
```

ocaml/xapi/repository.ml
```diff
@@ -51,51 +52,63 @@ let with_reposync_lock f =
...
-let cleanup_pool_repo () =
+let cleanup_all_pool_repositories () =
   try
-    clean_yum_cache !Xapi_globs.pool_repo_name;
-    Unixext.unlink_safe (Filename.concat !Xapi_globs.yum_repos_config_dir
-                           !Xapi_globs.pool_repo_name);
+    clean_yum_cache "*" ;
+    let prefix = !Xapi_globs.remote_repository_prefix ^ "-" in
+    Array.to_list (Sys.readdir !Xapi_globs.yum_repos_config_dir)
+    |> List.iter (fun n ->
```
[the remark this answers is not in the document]
lippirk: You can also use `Array.iter` instead of `Array.to_list .. |> List.iter`

ocaml/xapi/repository.ml
```diff
@@ -257,14 +332,20 @@ let get_repository_handler (req : Http.Request.t) s _ =
...
   let repomd_xml_path = Filename.concat repodata_dir "repomd.xml" in
   let md = UpdateInfoMetaData.of_xml_file repomd_xml_path in
-  if hash <> md.UpdateInfoMetaData.checksum then
-    (error "Unexpected mismatch between XAPI DB and YUM DB. Need to do pool.sync-updates again.";
-     raise Api_errors.(Server_error (createrepo_failed, [])));
+  if hash <> md.UpdateInfoMetaData.checksum then (
+    let msg =
+      "Unexpected mismatch between XAPI DB and YUM DB. Need to do pool.sync-updates again."
+    in
+    error "%s: %s" repo_name msg;
```
lippirk, line 346: It might be useful to log the checksums

ocaml/idl/datamodel_repository.ml
```diff
@@ -70,6 +71,9 @@ let t =
     ~contents:
       [ uid       _repository ~lifecycle;
         namespace ~name:"name" ~contents:(names ~writer_roles:_R_POOL_OP None RW) ();
+        field   ~qualifier:DynamicRO ~lifecycle
+          ~ty:(Set (Ref _pool)) "enabled_on"
+          "The pools where the repository is enabled on";
```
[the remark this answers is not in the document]
robhoes: There is always just one pool in the DB, so this could just be a single `ref`. And then is this field actually useful?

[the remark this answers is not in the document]
robhoes: But if you make it a single ref, you get a one-to-many relation, where you can only set the "one" side, which is this field. And that will make the pool setters go wrong. Personally, I'd just remove this field.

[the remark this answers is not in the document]
minglumlu: This `enabled_on` was added after I got error like:
```
gen_api_main ocaml/xapi-client/client.ml (exit 2)
(cd _build/default/ocaml/xapi-client && ../idl/ocaml_backend/gen_api_main.exe -filterinternal true -filter closed -mode client -output client.ml)
Fatal error: exception (Failure "Set(Ref _) field is not in relations table: pool.repositories")
```
And I had to add relation between them. Thus a "enabled_on" field in Repository. (But just as you pointed out, single ref of "enabled_on" makes the pool setters go wrong. And again I had to change type of "enabled_on" to set as well).

Back to the origin, it looks like the set "repositories" of pool does require a relation? May I get a pointer on how to avoid the error above without relation?

ocaml/xapi/repository_helpers.ml
minglumlu: The exception will be captured by difference API functions and raise proper error based on the specific API function.
In API function handler get_host_updates_in_json, it will raise get_host_updates_failed.
In API function handler Repository.apply_updates, it will raise apply_updates_failed
And the error message will be logged.

ocaml/idl/schematest.ml
minglumlu: There was a schema change in feature branch.
This is another change on top of that one. I would expect that in merging into master, there would be only one bumping-up on the schema version. So I don't bump it here.

minglumlu:
Hi @lindig @lippirk @robhoes 
Thanks for your comments. I resolved most of them. But following one may need more discussion:
https://github.com/xapi-project/xen-api/pull/4349#discussion_r594810491

ocaml/idl/schematest.ml
lippirk: Okay, my bad. In that case we'll leave it

ocaml/idl/datamodel_repository.ml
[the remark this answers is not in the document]
lippirk: Is it happening because you added the relation: `(_pool, "repositories"), (_repository, "enabled_on");`?

I'm also wondering why the API calls `add_repository`, `remove_repository`, `set_repositories` were added to datamodel_pool.ml and not datamodel_repository.ml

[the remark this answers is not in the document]
robhoes: I think that adding `~ignore_foreign_key:true` to your `set ref` field definition will avoid this. It tells the code generator not to expect a relation.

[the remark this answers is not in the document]
minglumlu: Hi Rob, with your help, the relation was dropped by adding `~ignore_foreign_key:true` in `pool.repositories` now. Thanks.

ocaml/xapi/repository_helpers.ml
```diff
@@ -684,27 +698,56 @@ let get_cachedir repo_name =
         let msg = Printf.sprintf "Not found cachedir for repository %s" repo_name in
         raise Api_errors.(Server_error (internal_error, [msg])))
 
-let with_local_repository ~__context f =
+let get_enabled_repositories ~__context =
+  let pool = Helpers.get_pool ~__context in
+  Db.Pool.get_repositories ~__context ~self:pool
+
+let get_remote_repository_name ~__context ~self =
+    !Xapi_globs.remote_repository_prefix ^ "-" ^ (Db.Repository.get_name_label ~__context ~self)
```
robhoes, line 706: The `name_label` field is now used as an important key that is used internally in names and URLs, where previously it wasn't. The datamodel defines it as a RW field along with `name_description`, and can be used by a client to put in strings without validation and at any time. If we are going to use the `name_label` as part of paths in the file system or URLs, then we need to carefully constrain them (e.g. check for bad characters) and encode where needed. We'd also need to handle changes to the field.

An alternative could be to use the UUID instead, which is fixed and has a known format.

robhoes (changes requested):
This mostly looks fine, although it is hard to check functional correctness in such a large PR. However, we can have some confidence thanks to the unit tests.

I only have one major issue about the use of `name_label`, as in the code comment.

ocaml/idl/datamodel_repository.ml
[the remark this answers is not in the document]
minglumlu: Hi Ben, I think the `add_repository`, `remove_repository` and `set_repositories` are pool level methods. So I added them into `datamodel_pool.ml`.

ocaml/xapi/repository_helpers.ml
minglumlu: It's good idea to use UUID. I just appended a new commit for that.

robhoes (approved):

minglumlu:
Hi @robhoes ,
Please note that I appended a fixup for wrong expectation on return of `Db.Repository.get_by_uuid` after your approve.

robhoes:
Thanks Ming, that looks fine.

ocaml/xapi/message_forwarding.ml
```diff
@@ -940,10 +940,21 @@ functor
                do_op_on ~local_fn ~__context ~host (fun session_id rpc ->
                    Client.Pool.enable_tls_verification rpc session_id))
 
-      let set_repository ~__context ~self ~value =
-        info "Pool.set_repository : pool = '%s'; value = %s"
-          (pool_uuid ~__context self) (repository_uuid ~__context value);
-        Local.Pool.set_repository ~__context ~self ~value
+      let set_repositories ~__context ~self ~value =
+        info "Pool.set_repositories : pool = '%s'; value = %s"
```
lippirk, line 944:
```suggestion
        info "Pool.set_repositories : pool = '%s'; value = [ %s ]"
```

ocaml/xapi/message_forwarding.ml
```diff
@@ -940,10 +940,21 @@ functor
...
+          (pool_uuid ~__context self)
+          (String.concat "; " (List.map (repository_uuid ~__context) value)) ;
+        Local.Pool.set_repositories ~__context ~self ~value
+
+      let add_repository ~__context ~self ~value =
+        info "Pool.add_repository : pool = '%s'; value = %s"
```
lippirk, line 950:
```suggestion
        info "Pool.add_repository : pool = '%s'; repo_uuid = %s"
```

similarly in `remove_repository`

ocaml/idl/datamodel_errors.ml
```diff
@@ -1217,6 +1217,8 @@ let _ =
     ~doc:"Failed to clean up local repository on master." ();
   error Api_errors.no_repository_enabled []
     ~doc:"There is no repository being enabled." ();
+  error Api_errors.multiple_update_repositories_enabled []
+    ~doc:"There are more than one update repository being enabled." ();
```
lindig, line 1221: Grammar: there *is* more than one

ocaml/xapi/repository.ml
```diff
@@ -51,51 +52,63 @@ let with_reposync_lock f =
...
+    Sys.readdir !Xapi_globs.yum_repos_config_dir
+    |> Array.iter (fun file ->
+        let open Astring.String in
+        if ( (is_prefix ~affix:prefix file) && (is_suffix ~affix:".repo" file) ) then (
+          let path = Filename.concat !Xapi_globs.yum_repos_config_dir file in
+          Unixext.unlink_safe path)) ;
     Helpers.rmtree !Xapi_globs.local_pool_repo_dir
   with e ->
-    error "Failed to cleanup pool repository: %s" (ExnHelper.string_of_exn e);
+    error "Failed to cleanup all pool repositories: %s" (ExnHelper.string_of_exn e);
```
lindig, line 67: Grammar: "failed to clean up" (cleanup is a noun and you need a verb here)

ocaml/xapi/repository.ml
```diff
@@ -51,51 +52,63 @@ let with_reposync_lock f =
...
+let cleanup_pool_repo ~__context ~self =
+  let repo_name = get_remote_repository_name ~__context ~self in
+  try
+    clean_yum_cache repo_name;
+    Unixext.unlink_safe (Filename.concat
+                           !Xapi_globs.yum_repos_config_dir
+                           (repo_name ^ ".repo"));
+    Helpers.rmtree (Filename.concat !Xapi_globs.local_pool_repo_dir repo_name)
+  with e ->
+    error "Failed to cleanup pool repository %s: %s" repo_name (ExnHelper.string_of_exn e);
```
lindig, line 79: Same as above: "clean up"

ocaml/xapi/repository.ml
```diff
@@ -51,51 +52,63 @@ let with_reposync_lock f =
...
-          "--save";
-          if !Xapi_globs.repository_gpgcheck then "--setopt=repo_gpgcheck=1"
-          else "--setopt=repo_gpgcheck=0";
-          Printf.sprintf "%s" !Xapi_globs.pool_repo_name;
-        ]
+      [
+        "--save"
+      ; if !Xapi_globs.repository_gpgcheck then "--setopt=repo_gpgcheck=1"
+        else "--setopt=repo_gpgcheck=0"
+      ; Printf.sprintf "%s" repo_name
```
lindig, line 94: This could be simplified to just `repo_name` since it's already a string.

ocaml/xapi/repository_helpers.ml
```diff
@@ -890,3 +962,32 @@ let consolidate_updates_of_host ~updates_info host updates_of_host =
       ("updates", `List (List.map (fun uid -> `String uid) (UpdateIdSet.elements uids)))]
   in
   (json_of_host, uids)
+
+let append_by_key l k v =
```
lindig, line 966: Is it legal to have a `(key, [v;v;])`: the same value is added multiple times to `key`?

lindig (approved):

ocaml/xapi/xapi_pool.ml
```diff
@@ -666,9 +673,8 @@ let pre_join_checks ~__context ~rpc ~session_id ~force =
            , [Features.name_of_feature Features.Pool_size] ))
   in
   let assert_no_hosts_in_updating () =
-    let pool = Helpers.get_pool ~__context in
     if List.exists (fun (_, op) -> op = `apply_updates)
-        (Client.Pool.get_current_operations ~rpc ~session_id ~self:pool) then
+        (Client.Pool.get_current_operations ~rpc ~session_id ~self:remote_pool) then
```
lippirk, line 677: This does indeed look like a bug to me. I thought that maybe we already perform this check somewhere else in the code for the remote pool instead, but I can't find it! The only problem with this change is that we are now not checking if the joining host is updating

My personal preference would be to perform this check on the remote pool itself (for example you could do it in `xapi_pool.ml:initial_auth` perhaps), and then you can leave the code in `pre_join_checks` alone

ocaml/idl/datamodel_errors.ml
```diff
@@ -1203,8 +1203,8 @@ let _ =
     ~doc:"The current version of Xen or its control libraries is incompatible with the Toolstack." ();
 
   (* repository and updates errors *)
-  error Api_errors.set_repository_in_progress []
-    ~doc:"The operation could not be performed because a repository setup is in progress." ();
+  error Api_errors.configure_repositories_in_progress []
+    ~doc:"The operation could not be performed because repositories setup is in progress." ();
```
lippirk, line 1207: The following sounds a bit more natural:

```suggestion
  error Api_errors.repository_configuration_in_progress []
    ~doc:"The operation could not be performed because another repository is already being configured" ();
```

ocaml/xapi/repository.ml
```diff
@@ -25,20 +25,21 @@ module UpdateIdSet = Set.Make (String)
...
-  create_repository_record ~__context ~name_label ~name_description ~binary_url ~source_url
+  create_repository_record
+    ~__context ~name_label ~name_description ~binary_url ~source_url ~update
 
 let forget ~__context ~self =
   let pool = Helpers.get_pool ~__context in
-  let enabled = Db.Pool.get_repository ~__context ~self:pool in
-  if enabled = self then
+  let enabled = Db.Pool.get_repositories ~__context ~self:pool in
+  if List.mem self enabled then
```
lippirk, line 42: Is this right? If a repository is in the database, then it is enabled?

ocaml/xapi/xapi_pool.ml
robhoes: Many of the pre-join checks look at properties of the remote pool using the API. I may not matter much, unless we put the remote pool into a kind of join-receive state such that it would block conflicting operations during the join. But we currently don't have such a mechanism.

I do think that Ben is right about still needing the local check. I also can't find a local operations check elsewhere.

ocaml/xapi/repository_helpers.ml
minglumlu: The final usage of this list `[v;v]` is to check if it is empty or not. So same value being added multiple times will not get wrong result.

ocaml/xapi/xapi_pool.ml
minglumlu: I just added a check on local joining host.

lippirk (approved):

robhoes:
@minglumlu It's all approved now. Please squash the fixups and merge. Thanks!

minglumlu:
Squashed and merged: `git diff d185c409a fd7a21522`

Issue #4337, opened 2021-03-03
Provide a network to host.evacuate


olivierlambert:
Hi there,

When you do a `host.evacuate`, it will use the management network by default. However, storage network could be faster to do it (often 1G vs 10G). If `host.evacuate` could have an extra optional arg like a network UUID, it would be really efficient to evacuate a host on a selected network.

Another way to do it would be to have a kind of "pool wide default migration network" that will be used by all method that doesn't specific anything for VM migration inside a pool with shared storage. But the second option seems more prone to side effects than the first one.

Any opinion about this?

robhoes:
I'd be happy for such a thing to be added. The `host.evacuate` call just does a bunch of `VM.pool_migrate` calls internally, so that is the first thing to update. It has an `options` argument, where we could accept a `network` key. It's just a matter of getting the right IP address here: https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_vm_migrate.ml#L315-L320. Then we could also add a `network` arg to `host.evacuate`.

olivierlambert:
Okay so to recap to be sure I understand correctly:

1. We add the optional `network` parameter in `host.evacuate`
2. If the parameter is detected, it will be simply passed to every `VM.pool_migrate`
3. However, `VM.pool_migrate` doesn't accept yet such parameter. So we have to also update it.

Is that correct?

robhoes:
Yep, so start with 3 :)

olivierlambert:
Thanks a lot @robhoes for your quick feedback :+1:

benjamreis:
I'm assuming the `options` argument of `pool_migrate` is an assoc list. Is that correct?

benjamreis:
PR has been merged, I think this issue can be closed. :)

Thanks for the help! +1

PR #4350, opened 2021-03-15
Enable to choose a migration network in `VM.pool_migrate` and `Host.evacuate`:

benjamreis:
- Add `network` key in `options` expecting a network ref to try to migrate on
- If not possible, fallback on management address

### DIFF
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -336,7 +336,14 @@ let host_query_ha = call ~flags:[`Session]
       ~in_product_since:rel_miami
       ~name:"evacuate"
       ~doc:"Migrate all VMs off of this host, where possible."
-      ~params:[Ref _host, "host", "The host to evacuate"]
+      ~lifecycle:[
+        Published, rel_miami, "";
+        Extended, rel_next, "Enable migration network selection."
+      ]
+      ~versioned_params:[
+        {param_type=Ref _host; param_name="host"; param_doc="The host to evacuate"; param_release=miami_release; param_default=None};
+        {param_type=Ref _network; param_name="network"; param_doc="Optional preferred network for migration"; param_release=next_release; param_default=(Some (VRef null_ref))}
+      ]
       ~allowed_roles:_R_POOL_OP
       ()
 
--- a/ocaml/xapi-cli-server/cli_operations.ml
+++ b/ocaml/xapi-cli-server/cli_operations.ml
@@ -4767,10 +4767,16 @@ let host_all_editions printer rpc session_id params =
   printer (Cli_printer.PList editions)
 
 let host_evacuate printer rpc session_id params =
+  let network =
+    List.assoc_opt "network-uuid" params
+    |> Option.fold ~none:Ref.null ~some:(fun uuid ->
+           Client.Network.get_by_uuid rpc session_id uuid)
+  in
   ignore
     (do_host_op rpc session_id ~multiple:false
-       (fun _ host -> Client.Host.evacuate rpc session_id (host.getref ()))
-       params [])
+       (fun _ host ->
+         Client.Host.evacuate rpc session_id (host.getref ()) network)
+       params ["network-uuid"])
 
 let host_get_vms_which_prevent_evacuation printer rpc session_id params =
   let uuid = List.assoc "uuid" params in
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -548,7 +548,7 @@ let compute_evacuation_plan ~__context ~host =
            Using original algorithm" ;
         compute_evacuation_plan_no_wlb ~__context ~host
 
-let evacuate ~__context ~host =
+let evacuate ~__context ~host ~network =
   let task = Context.get_task_id __context in
   let plans = compute_evacuation_plan ~__context ~host in
   (* Check there are no errors in this list *)
@@ -566,9 +566,26 @@ let evacuate ~__context ~host =
     match plan with
     | Migrate host ->
         ( try
+            ( if network <> Ref.null then
+                let hosts = Db.Host.get_all ~__context in
+                List.iter
+                  (fun host ->
+                    ignore
+                    @@ Xapi_network_attach_helpers
+                       .assert_valid_ip_configuration_on_network_for_host
+                         ~__context ~self:network ~host)
+                  hosts
+            ) ;
+            let with_network_option =
+              if network <> Ref.null then
+                [("network", Ref.string_of network)]
+              else
+                []
+            in
+            let options = ("live", "true") :: with_network_option in
             Helpers.call_api_functions ~__context (fun rpc session_id ->
                 Client.Client.VM.pool_migrate ~rpc ~session_id ~vm ~host
-                  ~options:[("live", "true")])
+                  ~options)
           with
         | Api_errors.Server_error (code, params)
           when code = Api_errors.vm_bad_power_state ->
--- a/ocaml/xapi/xapi_pif_helpers.ml
+++ b/ocaml/xapi/xapi_pif_helpers.ml
@@ -242,3 +242,11 @@ let is_device_underneath_same_type ~__context pif1 pif2 =
     (pci_rec.Db_actions.pCI_vendor_id, pci_rec.Db_actions.pCI_device_id)
   in
   get_device_info pif1 = get_device_info pif2
+
+let get_primary_address ~__context ~pif =
+  match Db.PIF.get_primary_address_type ~__context ~self:pif with
+  | `IPv4 -> (
+    match Db.PIF.get_IP ~__context ~self:pif with "" -> None | ip -> Some ip
+  )
+  | `IPv6 ->
+      List.nth_opt (Db.PIF.get_IPv6 ~__context ~self:pif) 0
--- a/ocaml/xapi/xapi_vm_migrate.ml
+++ b/ocaml/xapi/xapi_vm_migrate.ml
@@ -312,9 +312,18 @@ let pool_migrate ~__context ~vm ~host ~options =
   let queue_name = queue_of_vm ~__context ~self:vm in
   let module XenopsAPI = (val make_client queue_name : XENOPS) in
   let session_id = Ref.string_of (Context.get_session_id __context) in
-  let ip =
-    Http.Url.maybe_wrap_IPv6_literal (Db.Host.get_address ~__context ~self:host)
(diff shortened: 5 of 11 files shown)


benjamreis:
Solves: https://github.com/xapi-project/xen-api/issues/4337

benjamreis:
I'll rebase and keep only 2 commits (one for `VM.pool_migrate` and one for `Host.evacuate`) at the end before merge.

benjamreis:
Tested and working, good to be reviewed!

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -312,9 +312,33 @@ let pool_migrate ~__context ~vm ~host ~options =
...
+    | None ->
+        Db.Host.get_address ~__context ~self:host
+    | Some network_ref -> (
+        let pifs =
+          Db.Network.get_PIFs ~__context ~self:(Ref.of_string network_ref)
+        in
+        try
+          let pif =
+            List.find (fun x -> host = Db.PIF.get_host ~__context ~self:x) pifs
+          in
```
robhoes, line 327: `Xapi_network_attach_helpers.get_local_pifs`

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -312,9 +312,33 @@ let pool_migrate ~__context ~vm ~host ~options =
...
+          match Db.PIF.get_primary_address_type ~__context ~self:pif with
+          | `IPv4 ->
+              Db.PIF.get_IP ~__context ~self:pif
+          | `IPv6 -> (
+            match Db.PIF.get_IPv6 ~__context ~self:pif with
+            | [] ->
+                Db.Host.get_address ~__context ~self:host
```
robhoes, line 334: We should raise an error, because the client has explicitly asked to use a certain network.

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -312,9 +312,33 @@ let pool_migrate ~__context ~vm ~host ~options =
...
+            | ip :: _ ->
+                ip
+          )
+        with Not_found -> Db.Host.get_address ~__context ~self:host
```
robhoes, line 338: We should raise an error, because the client has explicitly asked to use a certain network.

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -312,9 +312,33 @@ let pool_migrate ~__context ~vm ~host ~options =
...
+      )
   in
+  let ip = Http.Url.maybe_wrap_IPv6_literal address in
```
robhoes, line 341: Don't we have a helper function for this? If not, we probably should ;)

ocaml/idl/datamodel_host.ml
```diff
@@ -336,7 +336,14 @@ let host_query_ha = call ~flags:[`Session]
       ~in_product_since:rel_miami
       ~name:"evacuate"
       ~doc:"Migrate all VMs off of this host, where possible."
-      ~params:[Ref _host, "host", "The host to evacuate"]
+      ~lifecycle:[
+        Published, rel_miami, "";
+        Changed, rel_next, "Enable migration network selection."
```
robhoes, line 341: You should use `Extended`, because this is a compatible change.

ocaml/xapi/xapi_host.ml
```diff
@@ -566,9 +566,15 @@ let evacuate ~__context ~host =
     match plan with
     | Migrate host ->
         ( try
+            let options =
+              if network = Ref.null then
+                [("live", "true")]
+              else
+                [("live", "true"); ("network", Ref.string_of network)]
```
robhoes, line 573:
```suggestion
              ("live", "true")
              :: (if network <> Ref.null then ["network", Ref.string_of network] else [])
```

ocaml/xapi/xapi_vm_migrate.ml
benjamreis: I thought it more about a preferred network migration, but I'm okay with raising an error. :)

ocaml/xapi/xapi_vm_migrate.ml
benjamreis: Indeed!

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -312,9 +312,39 @@ let pool_migrate ~__context ~vm ~host ~options =
...
+                  Xapi_pif_helpers.get_primary_address ~__context ~pif <> None)
+                pifs
+            in
+            Option.get (Xapi_pif_helpers.get_primary_address ~__context ~pif)
+          with Not_found ->
+            failwith
+              (Printf.sprintf
+                 "No PIF with IP configuration for host: [%s] on network: [%s]"
+                 (Ref.string_of host) (Ref.string_of network))
+        )
```
robhoes, line 344: By definition, there will only be zero or one local PIF for a host on any given network.

```suggestion
        | pif :: _ -> (
          match Xapi_pif_helpers.get_primary_address ~__context ~pif with
          | Some ip -> ip
          | None ->
            raise
              Api_errors.(Server_error (interface_has_no_ip, [Ref.string_of pif]))
        )
```

benjamreis:
Unrelated, is it possible to add a signoff when i accept a suggestion ? I have to amend the commit everytime to  not break the DCO ^^'

ocaml/xapi/xapi_host.ml
```diff
@@ -566,9 +566,18 @@ let evacuate ~__context ~host =
     match plan with
     | Migrate host ->
         ( try
+            let options =
+              ("live", "true")
+              ::
+              ( if network <> Ref.null then
+                  [("network", Ref.string_of network)]
```
robhoes, line 573: We should assert here that the network is suitable to avoid migration failures half-way through an evacuation. So we should do similar checks as in `pool_migrate` to assert that the network has a PIF with IP configuration on each host in the pool.

benjamreis: The calls `pool_migrate`, I don't see the point of replicating the assertion. Am i missing something?

benjamreis: Should iterate `assert_can_attach_network_on_host` on all pool's hosts?

vmpr:
great job guys! +1

ocaml/xapi/xapi_host.ml
benjamreis: Okay, I tried something, what do you think?

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -4765,9 +4765,17 @@ let host_all_editions printer rpc session_id params =
   printer (Cli_printer.PList editions)
 
 let host_evacuate printer rpc session_id params =
+  let network =
```
lindig, line 4768: Could be written as:
```
let network = List.assoc_opt "network" params |> Option.fold ~none:Ref.null ~some:Ref.of_string
```

ocaml/xapi/xapi_host.ml
lindig: I don't like too much mixing building a list and checking assertion at the same time.

lindig (approved):

ocaml/xapi-cli-server/cli_operations.ml
benjamreis: Yes, it's simpler!

benjamreis:
I've taken the assert out of the option list creation.

benjamreis:
Hi! 

Any more modification required? We have tested locally and through our own auto tests and the feature is validated.

Thanks

ocaml/xapi-cli-server/cli_operations.ml
psafont: Was there a reason not to change the code?

ocaml/xapi/xapi_host.ml
```diff
@@ -566,9 +566,28 @@ let evacuate ~__context ~host =
...
+            ) ;
+            let options =
+              ("live", "true")
+              ::
+              ( if network <> Ref.null then
+                  [("network", Ref.string_of network)]
+              else
+                []
+              )
+            in
```
psafont, line 587:
```suggestion
            let with_network_option =
              if if network <> Ref.null then
                [("network", Ref.string_of network)]
              else
                []
            in
            let options = ("live", "true") :: with_network_option in
```

ocaml/xapi/xapi_pif_helpers.ml
```diff
@@ -242,3 +242,16 @@ let is_device_underneath_same_type ~__context pif1 pif2 =
...
+  | `IPv4 -> (
+    match Db.PIF.get_IP ~__context ~self:pif with "" -> None | ip -> Some ip
+  )
+  | `IPv6 -> (
+    match Db.PIF.get_IPv6 ~__context ~self:pif with
+    | [] ->
+        None
+    | ip :: _ ->
+        Some ip
+  )
```
psafont, line 257:
```suggestion
    List.nth_opt (Db.PIF.get_IPv6 ~__context ~self:pif) 0
```

ocaml/xapi/xapi_host.ml
benjamreis: Why is there 2 `if`s in your suggestion?

psafont: it's a typo

psafont (approved):

ocaml/xapi-cli-server/cli_operations.ml
benjamreis: I forgot ^^'
Done now :)

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -4766,11 +4766,8 @@ let host_all_editions printer rpc session_id params =
 
 let host_evacuate printer rpc session_id params =
   let network =
-    match List.assoc_opt "network-uuid" params with
-    | None ->
-        Ref.null
-    | Some uuid ->
-        Client.Network.get_by_uuid rpc session_id uuid
+    List.assoc_opt "network-uuid" params
+    |> Option.fold ~none:Ref.null ~some:Ref.of_string
```
psafont, line 4770: This looks wrong, shouldn't this be 
```suggestion
    |> Option.fold ~none:Ref.null ~some:(Client.Network.get_by_uuid rpc session_id)
```
instead?

benjamreis: Yes! Good catch ^^'

PR #4351, opened 2021-03-15
CP-36100-2 Use explicit config for TLS Stunnel verification

lindig:
When Stunnel creates a conncetion as a client it can verify this
connection using certificates. We have intra pool connections and
connections opened to appliances, which use different verification
methods. This needs to be reflected beyond yes/no. This patch makes
(more) explicit the configuration to use for verifcation.

Some call sites still use no certificate checking (passing None to
SSL.make) and this will need to be adressed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,12 @@ list-hd:
 	echo counted $$LIST_HD usages ;\
 	test $$LIST_HD -eq 300
 
-quality-gate: list-hd ;
+verify-cert:
+	@NONE=$$( git grep -r --count 'verify_cert:None' -- **/*.ml | cut -d ':' -f 2 | paste -sd+ - | bc) ;\
+	echo "counted $$NONE usages of verify_cert:None" ;\
+	test $$NONE -eq 6
+
+quality-gate: list-hd verify-cert ;
 
 install: build doc sdk
 	mkdir -p $(DESTDIR)$(SBINDIR)
--- a/ocaml/database/master_connection.ml
+++ b/ocaml/database/master_connection.ml
@@ -76,8 +76,9 @@ let force_connection_reset () =
       | Some () ->
           purge_stunnels verify_cert
     in
-    purge_stunnels true ;
-    purge_stunnels false ;
+    purge_stunnels None ;
+    purge_stunnels (Some Stunnel.pool) ;
+    purge_stunnels (Some Stunnel.appliance) ;
     info
       "force_connection_reset: all cached connections to the master have been \
        purged"
@@ -157,9 +158,10 @@ let on_database_connection_established = ref (fun () -> ())
 let open_secure_connection () =
   let host = !get_master_address () in
   let port = !Db_globs.https_port in
+  let verify_cert = Stunnel_client.pool () in
   Stunnel.with_connect ~use_fork_exec_helper:true ~extended_diagnosis:true
     ~write_to_log:(fun x -> debug "stunnel: %s\n" x)
-    host port
+    ~verify_cert host port
   @@ fun st_proc ->
   let fd_closed = Thread.wait_timed_read Unixfd.(!(st_proc.Stunnel.fd)) 5. in
   let proc_quit =
--- a/ocaml/perftest/perfutil.ml
+++ b/ocaml/perftest/perfutil.ml
@@ -25,7 +25,7 @@ let rpc xml =
 let remoterpc host xml =
   let open Xmlrpc_client in
   XMLRPC_protocol.rpc ~srcstr:"perftest" ~dststr:"remotexapi"
-    ~transport:(SSL (SSL.make (), host, 443))
+    ~transport:(SSL (SSL.make ~verify_cert:None (), host, 443))
     ~http:(xmlrpc ~version:"1.1" "/")
     xml
 
--- a/ocaml/vncproxy/vncproxy.ml
+++ b/ocaml/vncproxy/vncproxy.ml
@@ -73,7 +73,11 @@ let _ =
           ~http xml
     | host ->
         XMLRPC_protocol.rpc ~srcstr:"vncproxy" ~dststr:"xapi"
-          ~transport:(SSL (SSL.make ~use_fork_exec_helper:false (), host, 443))
+          ~transport:
+            (SSL
+               ( SSL.make ~verify_cert:None ~use_fork_exec_helper:false ()
+               , host
+               , 443 ))
           ~http xml
   in
   let find_vm rpc session_id vm =
@@ -94,7 +98,11 @@ let _ =
           (Printf.sprintf "%s?ref=%s" Constants.console_uri (Ref.string_of vm))
       in
       let transport =
-        SSL (SSL.make ~use_fork_exec_helper:false (), address, 443)
+        SSL
+          ( SSL.make ~verify_cert:(Stunnel_client.pool ())
+              ~use_fork_exec_helper:false ()
+          , address
+          , 443 )
       in
       with_transport transport
         (with_http http (fun (response, fd) ->
--- a/ocaml/vncproxy/vncproxy.ml
+++ b/ocaml/vncproxy/vncproxy.ml
@@ -75,7 +75,8 @@ let _ =
         XMLRPC_protocol.rpc ~srcstr:"vncproxy" ~dststr:"xapi"
           ~transport:
             (SSL
-               ( SSL.make ~verify_cert:None ~use_fork_exec_helper:false ()
+               ( SSL.make ~verify_cert:(Stunnel_client.pool ())
+                   ~use_fork_exec_helper:false ()
                , host
                , 443 ))
           ~http xml
--- a/ocaml/xapi-cli-server/cli_operations.ml
+++ b/ocaml/xapi-cli-server/cli_operations.ml
@@ -4082,7 +4082,9 @@ let vm_migrate printer rpc session_id params =
       let open Xmlrpc_client in
       let http = xmlrpc ~version:"1.0" "/" in
       XMLRPC_protocol.rpc ~srcstr:"cli" ~dststr:"dst_xapi"
-        ~transport:(SSL (SSL.make ~use_fork_exec_helper:false (), ip, 443))
+        ~transport:
+          (SSL
+             (SSL.make ~verify_cert:None ~use_fork_exec_helper:false (), ip, 443))
         ~http xml
     in
     let username = List.assoc "remote-username" params in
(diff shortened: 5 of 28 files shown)


ocaml/database/master_connection.ml
```diff
@@ -157,9 +158,10 @@ let on_database_connection_established = ref (fun () -> ())
 let open_secure_connection () =
   let host = !get_master_address () in
   let port = !Db_globs.https_port in
+  let verify_cert = None (* CL *) in
```
psafont, line 161: I would think this is a pool connection as it connects to the pool master.

ocaml/perftest/perfutil.ml
```diff
@@ -25,7 +25,7 @@ let rpc xml =
 let remoterpc host xml =
   let open Xmlrpc_client in
   XMLRPC_protocol.rpc ~srcstr:"perftest" ~dststr:"remotexapi"
-    ~transport:(SSL (SSL.make (), host, 443))
+    ~transport:(SSL (SSL.make ~verify_cert:None (), host, 443))
```
psafont, line 28: @edwintorok Do we know under which circumstances this connection is opened? Is it always to another host in the pool?

ocaml/vncproxy/vncproxy.ml
```diff
@@ -73,7 +73,11 @@ let _ =
           ~http xml
     | host ->
         XMLRPC_protocol.rpc ~srcstr:"vncproxy" ~dststr:"xapi"
-          ~transport:(SSL (SSL.make ~use_fork_exec_helper:false (), host, 443))
+          ~transport:
+            (SSL
+               ( SSL.make ~verify_cert:None ~use_fork_exec_helper:false ()
```
psafont, line 78: I would think this is a pool connection: we're connecting to the member of the pool that can offer the session to the VM.

ocaml/perftest/perfutil.ml
lindig: It would be acceptable to not use verification for debugging or other non-production code. It would be also safe to set verification if we know for certain the destination.

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -4082,7 +4082,9 @@ let vm_migrate printer rpc session_id params =
       let open Xmlrpc_client in
       let http = xmlrpc ~version:"1.0" "/" in
       XMLRPC_protocol.rpc ~srcstr:"cli" ~dststr:"dst_xapi"
-        ~transport:(SSL (SSL.make ~use_fork_exec_helper:false (), ip, 443))
+        ~transport:
+          (SSL
+             (SSL.make ~verify_cert:None ~use_fork_exec_helper:false (), ip, 443))
```
psafont, line 4087: This one is a tough one. It's' cross-pool migration started from the cli. If trust is to be established beforehand the cli needs to be changed to have the certificate of the remote host as trusted while using the SNI 'pool'.

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -5181,7 +5183,7 @@ let vm_import fd printer rpc session_id params =
               let open Xmlrpc_client in
               let transport =
                 SSL
-                  ( SSL.make ~use_stunnel_cache:true
+                  ( SSL.make ~use_stunnel_cache:true ~verify_cert:None
```
psafont, line 5186: This code path imports a zurich / geneva XVA. I don't think we can do much here certificate-wise, and I don't know how used this use-case is. I remember @robhoes saying this code should just be dropped.

ocaml/xe-cli/newcli.ml
```diff
@@ -305,12 +305,14 @@ let exit_status = ref 1
 
 let with_open_tcp_ssl server f =
   let port = get_xapiport true in
+  let verify_cert = None in
```
psafont, line 308: This usage feels like it should be solved by having trust-on-first-use like in xencenter. Using a store for host certificates in the filesystem would work. This is outside the scope of the current work, though.

ocaml/xsh/xsh.ml
```diff
@@ -73,11 +73,12 @@ let proxy (ain : Unix.file_descr) (aout : Unix.file_descr)
 
 let with_open_tcp_ssl server f =
   let port = 443 in
+  let verify_cert = None (* CL *) in
```
psafont, line 76: What is this binary used for?

ocaml/xapi/xapi.ml
```diff
@@ -783,6 +783,13 @@ let set_stunnel_timeout () =
     Stunnel.timeoutidle := Some timeout
   with _ -> debug "Using default stunnel timeout (usually 43200)"
 
+let init_tls_verification ~__context () =
+  (* set global flag based on database entry *)
+  let pool = Helpers.get_pool ~__context in
+  let default = Db.Pool.get_tls_verification_enabled ~__context ~self:pool in
```
lippirk, line 789: This is not right in the case where a user emergency disabled verification on a host with `Host.emergency_disable_tls_verification`

I think `Stunnel_client.set_verify_by_default` should persist the setting as a file flag, and then we don't even this function

lindig: We need to be aware that the library is used by different binaries: at least xapi and xe, So who needs to maintain the state?

lindig: Alternatively, is there a way to check whether we are in emergency mode and initialize the default correctly here?

ocaml/vncproxy/vncproxy.ml
lindig: I've updated this.

ocaml/xapi/xapi.ml
lippirk: > Alternatively, is there a way to check whether we are in emergency mode and initialize the default correctly here?

on master right now, the check is just whether /var/xapi/verify-certificates exists or not (but with your changes that won't work anymore)

> We need to be aware that the library is used by different binaries: at least xapi and xe, So who needs to maintain the state?

It's clear that xapi should maintain at least its own state. Xapi could also maintain the state of xe clients which are running on hosts in a pool, but this might not be a good idea - e.g. what if you want to make an xe call to another pool? So you make a good point :+1: 

If we say that each binary should decide for itself how it wants to maintain its verification state, then having this `init_tls_verification` function does make sense (xe presumably needs one too?). Ultimately, for the xapi binary, the database can't be used for maintaining this state, so I would propose a file flag

lindig: How about: pool-enable-tls-verifcation removes a file flag and host-emergency-disable-tls-verification creates one? On startup we check for the file.

lippirk: > How about: pool-enable-tls-verifcation removes a file flag and host-emergency-disable-tls-verification creates one? On startup we check for the file.

This would mean 'verification enabled if and only if Db.Pool.get_tls_verification_enabled && file doesn't exist'. I think it will have to be: 'file exists if and only if verification enabled' (because of the chicken and the egg problem where we would need to contact the database to decide whether to enable verification, but in order to read the database we have to open a connection, but we don't know whether it should be verified or not ...)

lindig: For now I would suggest we merge this and know that emergency mode for TLS does not survive a toolstack restart. But we have to think about of the semantics of these emergency modes. I feel there is an asymmetry how it is enabled and disabled

ocaml/xapi/xapi.ml
```diff
@@ -783,6 +783,21 @@ let set_stunnel_timeout () =
...
+  (* set global flag based on database entry *)
+  let file = Xapi_globs.disable_tls_verification_path in
+  match Sys.file_exists file with
+  | true ->
+      warn "TLS verification is disabled on this host: %s" file ;
+      Stunnel_client.set_verify_by_default false
+  | false ->
+      let pool = Helpers.get_pool ~__context in
+      let default =
+        Db.Pool.get_tls_verification_enabled ~__context ~self:pool
```
lippirk, line 796: I mentioned this yesterday - I don't think this will work https://github.com/xapi-project/xen-api/pull/4351#discussion_r599630590

In order to decide whether to verify tls connections, slaves would have to connect to the master, but how do they know whether to verify that connection?

If you change `Xapi_globs.disable_tls_verification_path` to instead be `Xapi_globs.enable_tls_verification_path`, then the problem goes away

ocaml/xe-cli/newcli.ml
```diff
@@ -305,12 +305,14 @@ let exit_status = ref 1
...
+  (* CL *)
   debug "Connecting via stunnel to [%s] port [%d]\n%!" server port ;
   (* We don't bother closing fds since this requires our close_and_exec wrapper *)
   let open Safe_resources in
   Stunnel.with_connect ~use_fork_exec_helper:false
     ~write_to_log:(fun x -> debug "stunnel: %s\n%!" x)
-    ~extended_diagnosis:(!debug_file <> None) server port
+    ~verify_cert ~extended_diagnosis:(!debug_file <> None) server port
```
psafont, line 315: I'd like to have a test in the Makefile / CI to track the number on usages where verify_cert is None. We really want to have at most a couple of this for the justified use-cases.

Can this be changed to `~verify_cert:None`? It would be greppable and easy to count the usages.

See https://github.com/xapi-project/xen-api/commit/b5f213f022f63d629c036411a51f68a25a7577b1#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52 for an example on how it could be integrated.

lindig: I've made this more explicit

psafont: Nice, thanks! could you add a step to test it to .github/workflows/main.yml?

Makefile
```diff
@@ -74,7 +74,12 @@ list-hd:
 	echo counted $$LIST_HD usages ;\
 	test $$LIST_HD -eq 300
 
-quality-gate: list-hd ;
+verify-cert:
+	@NONE=$$( git grep -r --count 'verify_cert:None' | awk 'BEGIN { FS=":" } { c += $$2 } END { print c }') ;\
```
psafont, line 78:
```suggestion
	@NONE=$$( git grep -r --count 'verify_cert:None' -- **/*.ml | awk 'BEGIN { FS=":" } { c += $$2 } END { print c }') ;\
```

Makefile
```diff
@@ -74,7 +74,12 @@ list-hd:
...
+	echo "counted $$NONE usages of verify_cert:None" ;\
+	test $$NONE -eq 8
```
psafont, line 80:
```suggestion
	@NONE=$$(git grep -r --count '~verify_cert:None' -- **/*.ml | cut -d ':' -f 2 | paste -sd+ - | bc) ;\
	echo "counted $$NONE usages of verify_cert:None" ;\
	test $$NONE -eq 6
```

psafont: using awk along the .ml filter returns '2' for some reason.

lindig: You are right - the code needs to restrict the grep to `*.ml` files as otherwise the Makefile is counted as well - just found out.

lippirk (approved):

psafont (approved):

PR #4352, opened 2021-03-22
maintenance: allow tests to run in an opam sandbox

psafont:
Seems like the opam sandbox also made the tests break while using the container-based workflow for xs-opam. This change fixes them, as it happened with master.

The datamodel tests were also using the private name instead of the public one which broke them.

This change can be tested by creating an opam switch with the 8.2 packages and installing xapi while forcing opam to run its tests:

```
$ opam switch create stockholm 4.08.1
$ eval $(opam env --switch="stockholm" --set-switch)
$ opam repo add xs-opam-stockholm git+https://github.com/xapi-project/xs-opam.git#release/stockholm/lcm
$ opam repo remove default
$ opam pin add xapi . -t
```

### DIFF
--- a/ocaml/tests/dune
+++ b/ocaml/tests/dune
@@ -27,3 +27,7 @@
   )
   (action (run ./check-no-xenctrl %{x}))
 )
+
+(env
+  (_ (env-vars (XAPI_TEST 1)))
+)
--- a/ocaml/idl/dune
+++ b/ocaml/idl/dune
@@ -50,5 +50,5 @@
 (test
   (name schematest)
   (modules schematest)
-  (libraries xapi_datamodel)
+  (libraries xapi-datamodel)
 )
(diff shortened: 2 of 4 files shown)


ocaml/tests/dune
```diff
@@ -27,3 +27,7 @@
   )
   (action (run ./check-no-xenctrl %{x}))
 )
+
+(env
+  (_ (env-vars (XAPI_TEST 1)))
```
lippirk, line 32: `(_` looks like a typo! I assume it's correct though

lippirk (approved):

ocaml/tests/dune
psafont: `_` is a wildcard for the environment name, IIRC

lindig (approved):

PR #4353, opened 2021-03-23
CP-36510 REQ-403 distribute certs during Pool.enable_tls_verification

lippirk:
You can now execute `xe pool-enable-tls-verification` and it 'just works' (without having to manually set up the bundles)

It works as expected in the usual case, where no hosts go offline, nothing bad happens, etc. It has not been tested in failure cases.

From a high level, `cert_distrib.go` is supposed to distribute certs
between hosts. More specifically, it does the following:

 - aggregates a collection of host certs on the master
 - instructs hosts to save this collection
 - instructs hosts to regenerate their bundles, using said collection

### DIFF
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -999,6 +999,19 @@ let host_query_ha = call ~flags:[`Session]
       ~allowed_roles:_R_LOCAL_ROOT_ONLY
       ()
 
+  (* this internal call is used to process values of type [Cert_distrib.Wire.command] *)
+  let cert_distrib_atom = call
+      ~pool_internal:true
+      ~hide_from_docs:true
+      ~lifecycle:[Published, rel_next, ""]
+      ~name:"cert_distrib_atom"
+      ~doc:""
+      ~params:[Ref _host, "host", "The host";
+               String, "command", "The string encoded command"]
+      ~result:(String, "The string encoded result")
+      ~allowed_roles:_R_LOCAL_ROOT_ONLY
+      ()
+
   let get_server_certificate = call
       ~in_oss_since:None
       ~lifecycle:[Published, rel_george, ""; Changed, rel_inverness, "Now available to all RBAC roles."]
@@ -1540,6 +1553,7 @@ let host_query_ha = call ~flags:[`Session]
         set_sched_gran;
         get_sched_gran;
         emergency_disable_tls_verification;
+        cert_distrib_atom;
       ]
       ~contents:
         ([ uid _host;
--- a/ocaml/xapi/cert_distrib.ml
+++ b/ocaml/xapi/cert_distrib.ml
@@ -1,0 +1,247 @@
+module Unixext = Xapi_stdext_unix.Unixext
+
+module D = Debug.Make (struct let name = "cert_distrib" end)
+
+module XenAPI = Client.Client
+
+type existing_cert_strategy = Erase_old | Merge [@@deriving sexp]
+
+type 'a remote_call =
+  API.ref_host -> (Rpc.call -> Rpc.response) -> API.ref_session -> 'a
+
+module WireProtocol = struct
+  module Sexp = Sexplib.Sexp
+  open Sexplib0.Sexp_conv
+
+  type cert_blob = string [@@deriving sexp]
+
+  type cert = {uuid: string; blob: cert_blob} [@@deriving sexp]
+
+  type command =
+    | Collect
+    | Write of (existing_cert_strategy * cert list)
+    | GenBundle
+  [@@deriving sexp]
+
+  type result = CollectResult of cert_blob | WriteResult | GenBundleResult
+  [@@deriving sexp]
+
+  let string_of_command x = x |> sexp_of_command |> Sexp.to_string
+
+  let command_of_string x =
+    try Some (x |> Sexp.of_string |> command_of_sexp) with _ -> None
+
+  let dbg_of_command = function
+    | Collect ->
+        "Collect"
+    | Write (Erase_old, _) ->
+        "Write (Erase_old)"
+    | Write (Merge, _) ->
+        "Write (Merge)"
+    | GenBundle ->
+        "GenBundle"
+
+  let string_of_result x = x |> sexp_of_result |> Sexp.to_string
+
+  let result_of_string x =
+    try Some (x |> Sexp.of_string |> result_of_sexp) with _ -> None
+
+  let dbg_of_result = function
+    | CollectResult _ ->
+        "CollectResult"
+    | WriteResult ->
+        "WriteResult"
+    | GenBundleResult ->
+        "GenBundleResult"
+end
+
+let raise_internal ?e ?details msg : 'a =
+  let e =
+    Option.fold ~none:""
+      ~some:(fun e -> e |> Printexc.to_string |> Printf.sprintf "exception: %s")
+      e
+  in
+  let details = Option.value ~default:"" details in
+  [msg; details; e] |> String.concat ". " |> D.error "%s" ;
+  raise Api_errors.(Server_error (internal_error, [msg]))
+
+(* eventually the remote calls should probably become API calls in the datamodel
+ * but they remain here for quick development *)
+module Worker : sig
+  val local_exec : __context:Context.t -> command:string -> string
+
+  val remote_collect_cert : WireProtocol.cert_blob remote_call
+
+  val remote_write_certs_fs :
+    existing_cert_strategy -> WireProtocol.cert list -> unit remote_call
+
+  val remote_regen_bundle : unit remote_call
+end = struct
+  open WireProtocol
+
+  let my_cert = !Xapi_globs.server_cert_internal_path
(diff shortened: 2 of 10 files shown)


scripts/update-ca-bundle.sh
```diff
@@ -5,6 +5,16 @@
 
 set -e
 
+CERTS_DIR="${1:-/etc/stunnel/certs}"
```
psafont, line 8: This is also being modified by https://github.com/xapi-project/xen-api/pull/4351

It probably would be good to anticipate this and try to avoid any merge conflicts.

lindig: I've included the updated script also in my xapi PR for cert verification - so we might want to avoid conflicts.

ocaml/xapi/cert_distrib.ml
```diff
@@ -0,0 +1,262 @@
...
+    | Error reason ->
+        raise_internal
+          ~details:
+            (Printf.sprintf "get_my_cert: couldn't parse %s because %s" my_cert
+               reason)
+          ~msg:(Printf.sprintf "failed to parse %s" my_cert)
+    | Ok {host_cert} ->
+        host_cert
+
+  let mv a b =
```
lindig, line 88: I think it would be better to implement this directly here. Likewise below for rm
```
(** [rmtree path] removes a file or directory recursively without following
      symbolic links. It may raise [Failure] *)
  let rmtree path =
    let ( // ) = Filename.concat in
    let rec rm path =
      let st = Unix.lstat path in
      match st.Unix.st_kind with
      | Unix.S_DIR ->
          Sys.readdir path |> Array.iter (fun file -> rm (path // file)) ;
          Unix.rmdir path
      | _ ->
          Unix.unlink path
```

ocaml/xapi/dune
```diff
@@ -74,6 +74,7 @@
    rrdd-plugin.base
    rrdd-plugin.local
    sexplib
+   sexplib0
```
psafont, line 77: Unrelated note to the PR: I wish we could just use 1 library for S-expressions instead of 3 (on top of that there's also another in xs-opam that dune uses: csexp)

ocaml/xapi/cert_distrib.ml
lindig: See https://github.com/xapi-project/xenopsd/blob/master/lib/xenops_utils.ml#L161-L203 - I would expect similar code in xapi but don't know from the top of my head.

lindig (commented):
I like this module. I think the protocol that is being implemented here could be put a bit more explicitly into comments. I have some initial comments but need to think about it some more.

ocaml/xapi/cert_distrib.ml
lippirk: I've added rmrf and mv to helpers.ml (implemented in ocaml) - I couldn't see these functions in xapi already, which is a bit surprising

ocaml/idl/datamodel_host.ml
```diff
@@ -1029,6 +1029,18 @@ let host_query_ha = call ~flags:[`Session]
       ~allowed_roles:_R_LOCAL_ROOT_ONLY
       ()
 
+  let cert_distrib_atom = call
```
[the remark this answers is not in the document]
lippirk: The idea is that the variants in `Cert_distrib.Wire.payload` are the 'atoms', maybe I should write a comment about that somewhere

ocaml/xapi/dune
lippirk: why have n sexp libraries when you can have n+1?

psafont:
I think it was missing on the ticket, but the design requires the master host to verify the connections to all the members of the pool (including itself) to avoid enabling TLS certificate verification if anything failed during the setup.

ocaml/xapi/cert_distrib.mli
```diff
@@ -0,0 +1,15 @@
+type existing_cert_strategy = Erase | Merge
```
lippirk, line 1: @psafont  I added this - it could be useful for pool join:

```
Cert_distrib.go ~from_hosts:existing_pool_hosts ~to_hosts:[joining_host] ~existing_cert_strategy:Erase ;
Cert_distrib.go ~from_hosts:[joining_host] ~to_hosts:existing_pool_hosts ~existing_cert_strategy:Merge ;
```

but maybe we only want to distribute a joiner's cert to other slaves once it has 'officially' joined?

lindig (approved):

ocaml/xapi/cert_distrib.ml
```diff
@@ -0,0 +1,247 @@
...
+  let pool_certs_bk = Printf.sprintf "%s.bk" pool_certs
+
+  let get_my_cert () : cert_blob =
+    let open Gencertlib.Pem in
+    match parse_file my_cert with
+    | Error reason ->
+        raise_internal
+          ~details:(Printf.sprintf "reason: %s" reason)
+          (Printf.sprintf "failed to parse %s" my_cert)
+    | Ok {host_cert} ->
```
psafont, line 95: I'd prefer if it was made clear fields are being dropped here.
```suggestion
    | Ok {host_cert; _} ->
```

ocaml/xapi/cert_distrib.mli
psafont: We want to distribute certificates before the joiner restarts itself, there's no way to know if the host has just joined a pool after that point, so the filesystems must be correct before it restarts. The database will be refreshed when it sees its internal certificate is missing from the db.

ocaml/xapi/cert_distrib.ml
```diff
@@ -0,0 +1,247 @@
+module Unixext = Xapi_stdext_unix.Unixext
+
+module D = Debug.Make (struct let name = "cert_distrib" end)
+
+module XenAPI = Client.Client
+
+type existing_cert_strategy = Erase | Merge [@@deriving sexp]
```
psafont, line 7: It's not immediately clear what this implies on the operation. I was thinking if instead of Erase. something like `Erase_old` or `Only_new` would make more sense. An explanation of the implication on the mli would be welcome as well.

ocaml/xapi/cert_distrib.ml
```diff
@@ -0,0 +1,247 @@
...
+   * NB: if any individual call fails, then we don't continue with the distribution.
+   *     for example: suppose that the master is unable to obtain the cert from a host,
+   *     then we are guaranteed not to modify the trusted certs on any hosts. however
+   *     we do not guarantee 'atomicity', so if regenerating the bundle on one host
+   *     fails, then state across the pool will most likely become inconsistent, and
+   *     manual intervention may be required *)
+  let uuid host = Db.Host.get_uuid ~__context ~self:host in
+
+  Helpers.call_api_functions ~__context @@ fun rpc session_id ->
+  let blobs =
```
psafont, line 231: I'm thinking about how to change this so we can get rollback when cycling the internal certificates.

We could change this function here to request new certificates to hosts (and save them in a temp location).

After that, change the distribution so it's a two step process: 1. backup the certs and copy the new certificates (arm change) 2. remove the backup (finish_distribution) or remove the newly-copied certs and restore from backup (rollback)

ocaml/xapi/cert_distrib.ml
```diff
@@ -0,0 +1,247 @@
...
+    | GenBundle
+  [@@deriving sexp]
+
+  type result = CollectResult of cert_blob | WriteResult | GenBundleResult
+  [@@deriving sexp]
+
+  let string_of_payload x = x |> sexp_of_payload |> Sexp.to_string
+
+  let payload_of_string x =
+    try Some (x |> Sexp.of_string |> payload_of_sexp) with _ -> None
```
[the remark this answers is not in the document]
lippirk: Well the alternative is to just have separate messages in the datamodel for each step in the distribution procedure. This is what we do for Pool secret rotation (for example: https://github.com/xapi-project/xen-api/blob/6eedefc121d3015d73f2978d9c8871fee551fcbe/ocaml/xapi/xapi_host.ml#L2421)

I mentioned my motivation for defining my own protocol below:
```
(* eventually the remote calls should probably become API calls in the datamodel
 * but they remain here for quick development *)
```

Personally I prefer the current implementation because _pretty much_ everything to do with cert distribution can be kept in just this file, and we don't have to expose internal details in the datamodel. However I also appreciate the argument for putting all of these calls in datamodel, as that gives us explicit versioning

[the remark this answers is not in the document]
psafont: Don't worry about this, I'm just rambling because I haven't found a nice way to get rid of the `base` dependency ;)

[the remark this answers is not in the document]
lindig: There is a `sexplib0` package from JaneStreet with zero dependencies. Would that work?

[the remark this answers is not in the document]
psafont: Unfortunately, no. sexplib0 is unable to parse a string to generate a structured s-expression.

psafont (approved):

PR #4355, opened 2021-03-26
CP-36099 add host_internal cert type to db schema

lippirk:


psafont:
Adding some code to output red text for the hash check would help see it:

```
New schema hash ('0ab12bc6f78eeae82a987239b4b100a7') doesn't match the last known one. Please bump the
```

`\033[31;1mRED TEXT\033[0m NORMAL TEXT`

psafont (changes requested):
I think `ocaml/alerts/certificate/certificate_check.ml` should be changed as well to not delete alerts that affect `host_internal certificates (easiest way is probably to change `messages_in_host` (line 57) to filter by message name, just like records are filtered in 91-96)

psafont (approved):

lindig (approved):

PR #4356, opened 2021-03-30
CA-352988: Allow local live migration even when host is disabled

minglumlu:

### DIFF
--- a/ocaml/xapi/xapi_vm_helpers.ml
+++ b/ocaml/xapi/xapi_vm_helpers.ml
@@ -661,8 +661,12 @@ let assert_can_boot_here ~__context ~self ~host ~snapshot ~do_cpuid_check
   validate_basic_parameters ~__context ~self ~snapshot ;
   assert_host_is_live ~__context ~host ;
   assert_matches_control_domain_affinity ~__context ~self ~host ;
+  let is_local_live_migration =
+    (Db.VM.get_power_state ~__context ~self) = `Running
+       && (Db.VM.get_resident_on ~__context ~self) = host
+  in
   (* CA-233580: allow control domains to start on the host even if the latter is disabled *)
-  if not (Db.VM.get_is_control_domain ~__context ~self) then
+  if not (Db.VM.get_is_control_domain ~__context ~self) && not is_local_live_migration then
     assert_host_is_enabled ~__context ~host ;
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -661,8 +661,15 @@ let assert_can_boot_here ~__context ~self ~host ~snapshot ~do_cpuid_check
   validate_basic_parameters ~__context ~self ~snapshot ;
   assert_host_is_live ~__context ~host ;
   assert_matches_control_domain_affinity ~__context ~self ~host ;
+  let is_local_live_migration =
+    if (Db.VM.get_power_state ~__context ~self) = `Running
+       && (Db.VM.get_resident_on ~__context ~self) = host then
+      true
+    else
+      false
```
psafont, line 669:
```suggestion
    (Db.VM.get_power_state ~__context ~self) = `Running
       && (Db.VM.get_resident_on ~__context ~self) = host
```

lippirk:
If a host is disabled, is that not most likely because it is broken and the master cannot contact it?

Therefore doesn't removing this check just move the problem somewhere else?

robhoes:
Allowing a VM power-state operation on a disabled host contradicts the meaning of enabled/disabled. You haven't explained on the PR why this is useful. In your scenario, can we not just re-enable the host before doing this localhost migration?

minglumlu:
> Allowing a VM power-state operation on a disabled host contradicts the meaning of enabled/disabled. You haven't explained on the PR why this is useful. In your scenario, can we not just re-enable the host before doing this localhost migration?

Basically this host is still in updating. Keeping it disabled would protect this host from new VM start from outside.

minglumlu:
> If a host is disabled, is that not most likely because it is broken and the master cannot contact it?
> 
> Therefore doesn't removing this check just move the problem somewhere else?
This change only allows a running VM on a disabled host to migrate on the same host. Indeed this allows a little bit more than before. But I don't think it breaks the meaning of disabled too much. The disabled host is still in protect basically.

psafont:
I would be really helpful this information would be available in the commit message as well :)

robhoes:
Yes, @minglumlu please update the commit message.

lindig (approved):

PR #4359, opened 2021-03-30
CP-36509 update db for xapi_ssl.pem on startup if changed

lindig:
The host certificate /etc/xensource/xapi-ssl.pem has a corresponding
entry in the xapi database. At startup we check for changes and update
the database entry if the certificate has changed.

The update of the database entry is not yet implemented because the
corresponding code does not yet exist.

### DIFF
--- a/ocaml/xapi/certificates_sync.ml
+++ b/ocaml/xapi/certificates_sync.ml
@@ -10,14 +10,6 @@ open Rresult
 
 let ( let* ) = ( >>= )
 
-let is_self_signed cert =
-  R.return
-    X509.(
-      Distinguished_name.equal (Certificate.subject cert)
-        (Certificate.issuer cert))
-
-let pool_size ~__context = Db.Pool.get_all ~__context |> List.length
-
 let remove_cert ~__context cert =
   info "Removing certificate %s" (Ref.string_of cert) ;
   Db.Certificate.destroy ~__context ~self:cert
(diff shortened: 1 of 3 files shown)


ocaml/xapi/certificates_sync.ml
```diff
@@ -0,0 +1,87 @@
+module D = Debug.Make (struct let name = "certificates_sync" end)
```
lippirk, line 1: Can we have an mli for this file?

lippirk:
We'll want to sync xapi-pool-tls with the database too, is that going to be done in another PR?

psafont:
I think so, it's not trivial because the startup code is part of the pool join mechanism.

lippirk (approved):

ocaml/xapi/certificates_sync.ml
```diff
@@ -0,0 +1,87 @@
...
+
+(** determine if the database is up to date by comparing the fingerprint
+  of xapi-ssl.pem with the entry in the database *)
+let is_unchanged ~__context cert_ref cert =
+  let ref_hash = Db.Certificate.get_fingerprint ~__context ~self:cert_ref in
+  let cert_hash =
+    X509.Certificate.fingerprint Mirage_crypto.Hash.(`SHA256) cert
+    |> Certificates.pp_hash
+  in
+  Ok (cert_hash = ref_hash)
```
psafont, line 42: The function can't return an error, it's worth dropping the result here.
```suggestion
  cert_hash = ref_hash
```

psafont (approved):

ocaml/xapi/certificates_sync.ml
lindig: This needs a change at the caller, too. I've added it.

PR #4362, opened 2021-04-01
CA-353011: Clean up certificates from unknown hosts

psafont:

### DIFF
--- a/ocaml/xapi/db_gc_util.ml
+++ b/ocaml/xapi/db_gc_util.ml
@@ -178,7 +178,7 @@ let gc_PGPUs ~__context =
 
 let gc_VGPU_types ~__context =
   (* We delete a VGPU_type iff it does not appear in the supported_VGPU_types
-     	 * of any PGPU _and_ there doesn't exist a VGPU with this VGPU_type *)
+     of any PGPU _and_ there doesn't exist a VGPU with this VGPU_type *)
   let open Db_filter_types in
   let garbage =
     Db.VGPU_type.get_records_where ~__context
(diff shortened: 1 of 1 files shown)


ocaml/xapi/db_gc_util.ml
```diff
@@ -222,6 +222,23 @@ let gc_host_metrics ~__context =
...
+  let all_certificates = Db.Certificate.get_all_records ~__context in
+  let get_certificates =
+    List.map (fun self -> Db.Host.get_certificates ~__context ~self)
+  in
+  let host_certificates =
+    get_certificates (Db.Host.get_all ~__context) |> List.flatten
+  in
+  List.iter
+    (fun (cert, record) ->
+      if
```
lindig, line 235: I would use `List.filter` followed by `List.iter` - it's not more efficient but I think it fits the paradigm and is readable.

lindig: I think a comment about the intention would be also good: which certs you want to remote

lindig: not (A || B) == (not A) && (not B) - would the  latter logic be more readable?

robhoes:
Why not use `gc_connector`?

ocaml/xapi/db_gc_util.ml
```diff
@@ -222,6 +222,23 @@ let gc_host_metrics ~__context =
...
     all_host_metrics
 
+let gc_certificates ~__context =
+  let all_certificates = Db.Certificate.get_all_records ~__context in
+  let get_certificates =
+    List.map (fun self -> Db.Host.get_certificates ~__context ~self)
+  in
+  let host_certificates =
+    get_certificates (Db.Host.get_all ~__context) |> List.flatten
+  in
```
lippirk, line 232: you can use List.concat_map, I'd also just do it in one line

lippirk (approved):

psafont:
As I understood it, `gc_connector` is used for classes that represent a many to many relationship. Certificates is does not fit that description and it must be handled separately.

ocaml/xapi/db_gc_util.ml
psafont: I did the && logic before, not sure which one is less confusing. I'll add a comment to state we want to remove the certificates that have a host associates (`host and `internal_host) with an invalid host reference.

ocaml/xapi/db_gc_util.ml
psafont: understand the eagerness, this needs to be backported to stockholm :p

lindig (approved):

ocaml/xapi/db_gc_util.ml
psafont: I'll change it in the backport ;p

lindig:
Undraft and we merge it?

PR #4364, opened 2021-04-08
XSI-995 remove dangling PCI references from VM records

lippirk:
A customer has seen error messages such as:

```
xe vm-reset-powerstate --multiple --force
operation failed on <vm_uuid>. You gave an invalid object reference. The object may have recently been deleted. The class parameter gives the type of reference given, and the handle parameter echoes the bad value given.
class: PCI
handle: <PCI_ref>
```

It is not clear how they arrived in such a state, but we should be able to get them out by considering the case the the PCI is not valid in `force_state_reset_keep_current_operations`. This should fix both VM hard shutdown and VM reset powerstate

### DIFF
--- a/ocaml/xapi/xapi_vm_lifecycle.ml
+++ b/ocaml/xapi/xapi_vm_lifecycle.ml
@@ -819,10 +819,13 @@ let force_state_reset_keep_current_operations ~__context ~self ~value:state =
            Db.VGPU.set_scheduled_to_be_resident_on ~__context ~self:vgpu
              ~value:Ref.null ;
            Db.VGPU.set_PCI ~__context ~self:vgpu ~value:Ref.null) ;
-    (* release PCIs *)
-    List.iter
-      (fun pci -> Db.PCI.remove_attached_VMs ~__context ~self:pci ~value:self)
-      (Db.VM.get_attached_PCIs ~__context ~self) ;
+    Db.VM.get_attached_PCIs ~__context ~self
+    |> List.iter (fun pci ->
+           if Db.is_valid_ref __context pci then
+             Db.PCI.remove_attached_VMs ~__context ~self:pci ~value:self
+           else
+             (* XSI-995 pci does not exist, so remove it from the vm record *)
+             Db.VM.remove_attached_PCIs ~__context ~self ~value:pci) ;
     List.iter
       (fun pci ->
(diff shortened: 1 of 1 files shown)


psafont:
I thought that removing dangling references was usually done by https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/db_gc_util.ml is there a reason to avoid it and have a new mechanism?

lippirk:
> I thought that removing dangling references was usually done by https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/db_gc_util.ml is there a reason to avoid it and have a new mechanism?

I've moved the code to db_gc, there wasn't really a reason to a avoid it (though NB that the rest of the code in db_gc_util.ml tries to remove whole rows from the database, as far as I can tell, whereas the code I've added is only modifying existing VM rows). 

I originally had a config flag which a user would have to set for this code to run. However if we think it's safe, then maybe we shouldn't bother - so that it will actually get tested by our suites

lindig:
Is the existing GC mechanism not removing non-existent PCI references (like @psafont suggeted)?

lippirk:
> Is the existing GC mechanism not removing non-existent PCI references (like @psafont suggested)? The bug suggests it does not, though. There could be a race because it might take a while before the GC kicks in. But this should have cleared the reference eventually and obviously it did not.

I don't think so, since I don't see any PCI related code in db_gc_util - only GPU related stuff. So I suspect this is simply a bug we haven't seen ourselves before

lippirk:
@edwintorok  was worried about potential performance issues if we added this fix to the db_gc, so I've added a fixup to make it only happen once on xapi startup (similar to the original commit)

lippirk:
Instead of writing a new function to check VM records at start up, fix `force_state_reset_keep_current_operations` instead. This makes the change simpler

edwintorok (approved):
I would've hoped that the `all_relations` in datamodel ensured consistency, but it seems to work only in one direction (i.e. if you update attached_PCIs on a PCI object it updates attached_VMs).

Perhaps in the long run we would have something that runs on startup and ensures that all the `all_relations` are still valid (snapshot restores can invalidate some of the assumptions there I think).

robhoes (approved):
I believe that this is the right fix, thanks!

PR #4365, opened 2021-04-09
Add datamodel option to log Db.X.destroy calls

lippirk:
The motivation for change is to make it easier to figure out, from a bugtool, what happened in the xapi database

To see what this PR does, it's easiest to see the diff of the auto-generated file, db_actions.ml:

```
     and destroy ~__context ~self =
       let self = DM_to_String.ref_PCI self in
       let __t = Context.database_of __context in
       let module DB = (val (Db_cache.get __t) : Db_interface.DB_ACCESS) in
-      DB.delete_row __t "PCI" self
+      D.debug "deleting row from PCI table: ref=%s" self ; DB.delete_row __t "PCI" self
     (**  *)
```

Some ways we might want to extend the change:
 - add logging to other tables, apart from PCI
 - write to a file other than xensource.log
 - log other db calls, such as 'create'

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -4155,6 +4155,7 @@ module Event = struct
       obj_allowed_roles=_R_POOL_ADMIN;
       obj_implicit_msg_allowed_roles=_R_ALL;
       obj_doc_tags=[];
+      db_logging=None;
     }
 end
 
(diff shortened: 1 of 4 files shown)


lindig:
I think it looks good. I would have preferred to put the log statement on a line by its own but it's fine.

lindig:
We recently encountered the problem with certificates - so it might be valuable to add it for them, too. But I'd be happy to merge this now

lindig (approved):

psafont (approved):

PR #4368, opened 2021-04-13
CP-36658 remove certs of ejected host

lindig:
When a host leaves a pool, the remaining hosts in the pool must no
longer trust it. This is implemented by removing the leaving host's
certificates from the database and file system.

### DIFF
--- a/ocaml/xapi/certificates_sync.ml
+++ b/ocaml/xapi/certificates_sync.ml
@@ -97,3 +97,44 @@ let update ~__context =
   let* () = sync ~__context ~type':`host in
   let* () = sync ~__context ~type':`host_internal in
   Ok ()
+
+let internal_error fmt =
+  fmt
+  |> Printf.kprintf @@ fun msg ->
+     error "%s" msg ;
+     raise Api_errors.(Server_error (internal_error, [msg]))
+
+let remove_from_db ~__context cert =
+  try
+    Db.Certificate.destroy ~__context ~self:cert ;
+    info "removed host certificate %s from db" (Ref.string_of cert)
+  with e ->
+    internal_error "failed to remove cert %s: %s" (Ref.string_of cert)
+      (Printexc.to_string e)
+
+let path host_uuid =
+  let prefix = !Xapi_globs.trusted_pool_certs_dir in
+  Filename.concat prefix (Printf.sprintf "%s.pem" host_uuid)
+
+let host_certs_of ~__context host =
+  List.concat
+    [
+      Certificates.Db_util.get_host_certs ~__context ~host ~type':`host
+    ; Certificates.Db_util.get_host_certs ~__context ~host ~type':`host_internal
+    ]
+
+let eject_certs_from_db ~__context certs =
+  certs |> List.iter (remove_from_db ~__context)
+
+let eject_certs_from_fs_for ~__context host =
+  (* the cert is not identified by its UUID in the file system but
+     by the UUID of the host it belongs to *)
+  let host_uuid = Db.Host.get_uuid ~__context ~self:host in
+  let file = path host_uuid in
+  try
+    Sys.remove file ;
+    Certificates.update_ca_bundle () ;
+    info "removed host certificate %s" file
+  with e ->
+    internal_error "failed to remove cert %s on pool eject" file
+      (Printexc.to_string e)
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -745,8 +745,22 @@ functor
         info "Pool.eject: pool = '%s'; host = '%s'"
           (current_pool_uuid ~__context)
           (host_uuid ~__context host) ;
+        let master = Helpers.get_master ~__context in
         let local_fn = Local.Pool.eject ~host in
+        let other =
+          Db.Host.get_all ~__context
+          |> List.filter (fun h -> h <> host && h <> master)
+        in
+        (* eject host but don't remove it from DB yet *)
         do_op_on ~local_fn ~__context ~host (fun session_id rpc ->
+            Client.Pool.eject rpc session_id host) ;
+        (* call eject on all other slaves first *)
+        other
+        |> List.iter (fun h ->
+               do_op_on ~local_fn ~__context ~host:h (fun session_id rpc ->
+                   Client.Pool.eject rpc session_id host)) ;
+        (* finally clean up on master *)
+        do_op_on ~local_fn ~__context ~host:master (fun session_id rpc ->
             Client.Pool.eject rpc session_id host)
 
       let designate_new_master ~__context ~host =
--- a/ocaml/xapi/xapi_pool.ml
+++ b/ocaml/xapi/xapi_pool.ml
@@ -1410,7 +1410,7 @@ let unplug_pbds ~__context host =
       List.iter (fun sr -> Client.SR.forget ~rpc ~session_id ~sr) srs_to_delete)
 
 (* This means eject me, since will have been forwarded from master  *)
-let eject ~__context ~host =
+let eject_self ~__context ~host =
   (* If HA is enabled then refuse *)
   let pool = Helpers.get_pool ~__context in
   if Db.Pool.get_ha_enabled ~__context ~self:pool then
@@ -1574,8 +1574,6 @@ let eject ~__context ~host =
     Net.reset_state () ;
     Xapi_inventory.update Xapi_inventory._current_interfaces "" ;
     debug "Pool.eject: deleting Host record (the point of no return)" ;
-    (* delete me from the database - this will in turn cause PBDs and PIFs to be GCed *)
-    Db.Host.destroy ~__context ~self:host ;
     Create_misc.create_pool_cpuinfo ~__context ;
     (* Update pool features, in case this host had a different license to the
        		 * rest of the pool. *)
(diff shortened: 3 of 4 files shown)


lindig:
This code is not yet called from the code for pool eject. I would like to understand: each host in the pool must remove the certs from the file system but the removal from the database needs to happen only once.

psafont:
>  I would like to understand: each host in the pool must remove the certs from the file system but the removal from the database needs to happen only once.

This is correct. The pool eject calls gets forwarded to the member that is about to be ejected and driven from there.

I would try to add the call to delete the certificates from the filesystems after the pool.eject call in the member has succeeded and be it driven from the master, this is in message_forwarding.

It could be called by the member being ejected right before it deleted the host from the database, but at that point the networking info has been deleted, so I'm not sure if it'll work.

@robhoes do you have any strong feelings about this approach? I don't see there's anything similar case that exists.

ocaml/xapi/xapi_pool.ml
```diff
@@ -1620,6 +1620,22 @@ let eject ~__context ~host =
...
+  let self_eject = host = Helpers.get_localhost ~__context in
+  let self_master = Pool_role.is_master () in
+  match (self_eject, self_master) with
+  | true, false ->
+      eject ~__context ~host
+  | false, false ->
+      Certificates_sync.eject_certs_from_fs_for ~__context host
+  | false, true ->
+      Certificates_sync.eject_certs_from_fs_for ~__context host ;
+      Certificates_sync.eject_certs_from_db_for ~__context host
```
lippirk, line 1635: I think it's safer if this line is in message_forwarding.ml - removing the cert from the DB must be the absolute last step, because otherwise calls to `Certificates_sync.eject_certs_from_fs_for ~__context host` will fail

lindig: That is a good point - will move that back to message forwarding

ocaml/xapi/xapi_pool.ml
```diff
@@ -1620,6 +1620,21 @@ let eject ~__context ~host =
...
+let eject ~__context ~host =
+  let self_eject = host = Helpers.get_localhost ~__context in
+  let self_master = Pool_role.is_master () in
+  match (self_eject, self_master) with
+  | true, false ->
+      eject ~__context ~host
+  | false, false ->
+      Certificates_sync.eject_certs_from_fs_for ~__context host
+  | false, true ->
+      Certificates_sync.eject_certs_from_fs_for ~__context host
```
lippirk, line 1634: This can now be simplified

ocaml/xapi/certificates_sync.ml
```diff
@@ -97,3 +97,53 @@ let update ~__context =
...
+     error "%s" msg ;
+     raise Api_errors.(Server_error (internal_error, [msg]))
+
+let path ~__context cert =
+  let uuid = Db.Certificate.get_uuid ~__context ~self:cert in
+  let type' = Db.Certificate.get_type ~__context ~self:cert in
+  let prefix =
+    match type' with
+    | `host ->
+        !Xapi_globs.trusted_certs_dir
```
psafont, line 113: The trusted certs dir is used by `ca certs, not `host ones. The correct filename for the certificate can be build by concatenating it with the field `name` in their record.

lindig: We use the following paths to build bundles which we pass to stunnel for client connections:

```
regen_bundle "/etc/stunnel/certs"      "/etc/stunnel/xapi-stunnel-ca-bundle.pem"
regen_bundle "/etc/stunnel/certs-pool" "/etc/stunnel/xapi-pool-ca-bundle.pem"
```
These correspond to these names:

```
let trusted_certs_dir = ref "/etc/stunnel/certs"
let trusted_pool_certs_dir = ref "/etc/stunnel/certs-pool"
```

So where are the certs that belong to a host in the pool that need to be removed and which are tracked in the database as host and host-internal certificates? According to `xe` they don't have a meaningful name, which could be a different problem:

```

Comment (lindig) [reply ocaml/xapi/certificates_sync.ml:113]: The correct paths are:
```
Xapi_globs.server_cert_path
Xapi_globs.server_cert_internal_path
```

```

ocaml/xapi/message_forwarding.ml
```diff
@@ -746,8 +746,14 @@ functor
...
         let local_fn = Local.Pool.eject ~host in
-        do_op_on ~local_fn ~__context ~host (fun session_id rpc ->
-            Client.Pool.eject rpc session_id host)
+        let hosts = Xapi_pool_helpers.get_master_slaves_list ~__context in
+        (* call eject on all slaves but only slave "host" will eject
+           itself *)
+        hosts
+        |> List.iter (fun h ->
+               do_op_on ~local_fn ~__context ~host:h (fun session_id rpc ->
+                   Client.Pool.eject rpc session_id host)) ;
```
lippirk, line 755: Is the order in which you call Pool.eject on all the hosts important?

Suppose that the master-slave list looks like `[ master ; slave_1 ; slave_2 ]`, and that `slave_2` is the ejectee. The master will first remove the ejectee's certificates, then the master will ask `slave_1` to remove the ejectee's certificates, and then finally the master will ask `slave_2` to eject. However that will fail because the master no longer trusts slave_2

It might be safer to do this:

```
Client.Pool.eject rpc session_id ejectee ;
hosts |> List.filter (fun h -> h <> ejectee) |> List.iter (fun h -> Client.Pool.eject rpc session_id) ;
```

lindig: I agree - the core eject operation needs to happen first before we can remove certs.

lindig:
Un-drafted this now. Code is quite a bit simpler now and works in my manual tests. `Xapi_pool.eject()` now either ejects a host as before, or removes the certs of that host from remaining hosts in the pool - except the master, which is handled in `message_forwarding`.

ocaml/xapi/xapi_pool.ml
```diff
@@ -1620,6 +1620,20 @@ let eject ~__context ~host =
       (fun () -> Xapi_fuse.light_fuse_and_reboot_after_eject ()) ;
     Xapi_hooks.pool_eject_hook ~__context
 
+(** eject [host] from the pool. This code is run on all hosts in the
+pool but only [host] will eject itself. *)
+let eject ~__context ~host =
```
psafont, line 1625: This is surprising, can the existing eject in line 1387 have its name changed? Maybe `self_eject`?

ocaml/xapi/xapi_pool.ml
```diff
@@ -1620,6 +1620,20 @@ let eject ~__context ~host =
...
+  let local = Helpers.get_localhost ~__context in
+  let master = Helpers.get_master ~__context in
+  let host_uuid = Db.Host.get_uuid __context host in
+  match (host = local, host = master) with
+  | true, false ->
+      eject ~__context ~host
+  | false, _ ->
+      Certificates_sync.eject_certs_from_fs_for ~host_uuid
```
psafont, line 1633: Although it makes sense to call directly this function for all the `other` hosts in message_forwarding I'd rather keep it this way because it's not exposed in the API.

ocaml/xapi/message_forwarding.ml
```diff
@@ -745,9 +745,27 @@ functor
...
+        let other =
+          Xapi_pool_helpers.get_master_slaves_list ~__context
+          |> List.filter (fun h -> h <> host && h <> master)
+        in
+        (* call eject on all other slaves first; they can't communicate
+           with host after that *)
+        other
+        |> List.iter (fun h ->
+               do_op_on ~local_fn ~__context ~host:h (fun session_id rpc ->
+                   Client.Pool.eject rpc session_id host)) ;
```
psafont, line 761: This should be done _after_ the host has self-ejected. The ejection may fail when HA is enabled or there's a VM resident in this host. Is there anything stopping this from being done after the ejection?

psafont (changes requested):

ocaml/xapi/message_forwarding.ml
robhoes: > they can't communicate with host after that

They don't need to communicate with the ejected host. Only the master does, and only to get the outcome of the rejection.

psafont: If the ejection fails the hosts will not have the certificate to contact the host. If, for some reason the master rotates, it won't be able to forward calls to the host that was tried to be ejected.

lindig: If the host is ejected first, it is impossible to later refer to it by its ref afterwards, as is required when calling `Xapi_pool.eject` on all other hosts.

lindig: If the cleanup of the remaining host needs to happen after the host was ejected it needs a new internal API call that refers to the host by its UUID. This is ugly but avoids the worse problem what happens if eject fails.

psafont: I see, it's not possible to change the ref to a uuid because it's part of the API.
In that case would it be possible to  separate the pre_checks that can make the eject to fail into another function so they can be run before deleting certificates?

Otherwise somehow exposing the function `eject_certs_from_fs_for` to be able to retrieve the uuid beforehand and do the file-system after the ejectee has rebooted.

lindig: We could delay the removal from the database and only do it once:
* the host ejected itself
* all other hosts removed the certs
This would allows us to use the API call using the ref

lindig:
Next attempt; now we eject the host first, then remove its cert from all other pool members, and finally on the master and from the database. If the eject fails, we can still recover from this.

ocaml/xapi/message_forwarding.ml
```diff
@@ -745,8 +745,22 @@ functor
         info "Pool.eject: pool = '%s'; host = '%s'"
           (current_pool_uuid ~__context)
           (host_uuid ~__context host) ;
+        let master = Helpers.get_master ~__context in
         let local_fn = Local.Pool.eject ~host in
+        let other =
+          Xapi_pool_helpers.get_master_slaves_list ~__context
```
psafont, line 751: If the master host is removed from the list, why use this call instead of `Db.Host.get_all ~__context`?

psafont (approved):
Great work!

ocaml/xapi/message_forwarding.ml
lindig: Good point - will use it instead

PR #4370, opened 2021-04-15
CP-36059: Manage winbind daemon lifecycle

liulinC:

### DIFF
--- a/ocaml/xapi/extauth_ad.ml
+++ b/ocaml/xapi/extauth_ad.ml
@@ -1,0 +1,79 @@
+(*
+ * Copyright (C) Citrix Systems Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+(**
+ * @AD group Access Control
+*)
+
+open Db_actions
+
+module D = Debug.Make (struct let name = "extauth_ad" end)
+
+open D
+
+module AD_type = struct
+  exception Unknown_AD_type of string
+
+  type t = Pbis | Winbind
+
+  let of_string = function
+    | "pbis" ->
+        Pbis
+    | "winbind" ->
+        Winbind
+    | _ as at ->
+        error "Unknown AD type '%s'" at ;
+        raise (Unknown_AD_type at)
+end
+
+module AD = struct
+  let start_backend_daemon ~wait_until_success = function
+    | AD_type.Pbis ->
+        Extauth_plugin_ADpbis.Lwsmd.start ~wait_until_success ~timeout:5.
+    | AD_type.Winbind ->
+        Extauth_plugin_ADwinbind.Winbind.start ~wait_until_success ~timeout:5.
+
+  let stop_backend_daemon ~wait_until_success = function
+    | AD_type.Pbis ->
+        Extauth_plugin_ADpbis.Lwsmd.stop ~wait_until_success ~timeout:3.
+    | AD_type.Winbind ->
+        Extauth_plugin_ADwinbind.Winbind.stop ~wait_until_success ~timeout:3.
+
+  let init_service ~__context = function
+    | AD_type.Pbis ->
+        Extauth_plugin_ADpbis.Lwsmd.init_service ~__context
+    | AD_type.Winbind ->
+        Extauth_plugin_ADwinbind.Winbind.init_service ~__context
+end
+
+let start_backend_daemon ~wait_until_success =
+  !Xapi_globs.extauth_ad_backend
+  |> AD_type.of_string
+  |> AD.start_backend_daemon ~wait_until_success
+
+let stop_backend_daemon ~wait_until_success =
+  !Xapi_globs.extauth_ad_backend
+  |> AD_type.of_string
+  |> AD.stop_backend_daemon ~wait_until_success
+
+let init_service ~__context =
+  !Xapi_globs.extauth_ad_backend
+  |> AD_type.of_string
+  |> AD.init_service ~__context
+
+let methods =
+  match !Xapi_globs.extauth_ad_backend |> AD_type.of_string with
+  | Pbis ->
+      Extauth_plugin_ADpbis.AuthADlw.methods
+  | Winbind ->
+      Extauth_plugin_ADwinbind.AuthADWinbind.methods
--- a/ocaml/xapi/extauth_ad.mli
+++ b/ocaml/xapi/extauth_ad.mli
@@ -1,0 +1,11 @@
+(* start AD external auth backend daemon *)
+val start_backend_daemon : wait_until_success:bool -> unit
+
+(* stop AD external auth backend daemon *)
+val stop_backend_daemon : wait_until_success:bool -> unit
+
+(* init AD external auth backend service *)
+val init_service : __context:Context.t -> unit
+
+(* methods to implement auth signature *)
+val methods : Auth_signature.t
--- a/ocaml/xapi/extauth_plugin_ADwinbind.ml
+++ b/ocaml/xapi/extauth_plugin_ADwinbind.ml
@@ -1,0 +1,147 @@
+(*
+ * Copyright (C) Citrix Systems Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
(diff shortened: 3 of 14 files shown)


ocaml/xapi/xapi_systemctl.ml
```diff
@@ -0,0 +1,51 @@
...
+type t = Start | Stop
+
+let to_string = function Start -> "start" | Stop -> "stop"
+
+let operate ~wait_until_success ~service ~timeout op =
+  let op_str = op |> to_string in
+  try
+    debug "%s %s" op_str service ;
+    ignore
+      (Forkhelpers.execute_command_get_output !Xapi_globs.systemctl
```
psafont, line 31: Fe_systemctl is not used here because it creates and deletes the service files on start and stop, something that's is not convenient to use here.

There are a few non-transient services managed from xapi, so there's an oportunity to consolidate these systemctl calls.

ocaml/xapi/extauth_ad.ml
```diff
@@ -0,0 +1,74 @@
+(*
+ * Copyright (C) 2006-2009 Citrix Systems Inc.
```
lippirk, line 2:
```suggestion
 * Copyright (C) Citrix Systems Inc.
```

ocaml/xapi/extauth_ad.ml
```diff
@@ -0,0 +1,74 @@
...
+module Ad_type = struct
+  type t
+
+  let of_string = function
+    | "pbis" ->
+        Pbis
+    | "winbind" ->
+        Winbind
+    | _ as at ->
+        debug "Unknow ad type of %s" at ;
```
lippirk, line 37:
```suggestion
        error "Unknown AD type '%s'" at ;
```

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -0,0 +1,147 @@
+(*
+ * Copyright (C) 2006-2009 Citrix Systems Inc.
```
lippirk, line 2:
```suggestion
 * Copyright (C) Citrix Systems Inc.
```

ocaml/xapi/xapi_systemctl.ml
```diff
@@ -0,0 +1,51 @@
+(*
+ * Copyright (C) 2006-2009 Citrix Systems Inc.
```
lippirk, line 2:
```suggestion
 * Copyright (C) Citrix Systems Inc.
```

lippirk:
@liulinC please can you add interface files (.mli files) for any new files you add (e.g. for xapi_systemctl.ml and  extauth_plugin_ADwinbind.ml). Otherwise this is looking good

ocaml/xapi/extauth_ad.ml
```diff
@@ -0,0 +1,74 @@
...
+        Extauth_plugin_ADwinbind.Winbind.stop ~wait_until_success ~timeout:3.
+
+  let init_service ~__context = function
+    | Pbis ->
+        Extauth_plugin_ADpbis.Lwsmd.init_service ~__context
+    | Winbind ->
+        Extauth_plugin_ADwinbind.Winbind.init_service ~__context
+end
+
+let start_backend_daemon ~wait_until_success =
```
lippirk, line 61: Suppose we were using PBIs, and then xapi crashed, but before rebooting, we changed the AD backend to winbind. Would we then have both daemons running? 

I doubt we want this - maybe in this function we could make we sure we shutdown any other daemons before starting the current backend

liulinC: change the AD backend is only a temporary emergency  solution, for your case, we would expect customer (or script) to
- stop xapi (skip if xaip crash)
- stop winbind daemon
- change ad_backend in xapi.conf
- start xapi 

If we coding that in xapi,  and always stop pbis before start winbind,  this will always introduce extra failures in xapi logs when it try to stop PBIS.  As the PBIS baries and services is not installed at all.

lippirk: I think we can tell systemd itself to not start both daemons at once, with a `Conflicts` directive, e.g. `Conflicts=winbind.service`. This is probably a simpler solution than having xapi manage this

ocaml/xapi/extauth_ad.ml
```diff
@@ -0,0 +1,74 @@
...
+ * @AD group Access Control
+*)
+
+open Db_actions
+
+module D = Debug.Make (struct let name = "extauth_ad" end)
+
+open D
+
+exception Unknown_Ad_type of string
```
lindig, line 24: I would suggest to spell this consistently as `AD` where possible; maybe `Unknown_AD_type`. I believe this should be declared in the module below.

ocaml/xapi/extauth_ad.ml
```diff
@@ -0,0 +1,74 @@
...
+
+type ad_type = Pbis | Winbind
+
+module Ad_type = struct
```
lindig, line 28: `AD_type` as suggested above

ocaml/xapi/extauth_ad.ml
```diff
@@ -0,0 +1,74 @@
...
+  type t
```
lindig, line 29: What is this type doing? It has no implementation and this is not a signature.

ocaml/xapi/extauth_ad.ml
```diff
@@ -0,0 +1,74 @@
...
+
+open Db_actions
+
+module D = Debug.Make (struct let name = "extauth_ad" end)
+
+open D
+
+exception Unknown_Ad_type of string
+
+type ad_type = Pbis | Winbind
```
lindig, line 26: Should this not be declared in `AD_type` as `AD_type.t`?

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -0,0 +1,147 @@
...
+      Takes a username and password, and tries to authenticate against an already configured
+      auth service (see XenAPI requirements Wiki page for details of how auth service configuration
+      takes place and the appropriate values are stored within the XenServer Metadata).
+      If authentication is successful then a subject_id is returned representing the account
+      corresponding to the supplied credentials (where the subject_id is in a namespace managed by
+      the auth module/service itself -- e.g. maybe a SID or something in the AD case).
+      Raises auth_failure if authentication is not successful
+  *)
+
+  let authenticate_username_password username password =
```
lindig, line 49: I suggest to use labels because you have two strings here: `~username` `~password`

ocaml/xapi/xapi_systemctl.ml
```diff
@@ -0,0 +1,51 @@
...
+    ignore
+      (Forkhelpers.execute_command_get_output !Xapi_globs.systemctl
+         [op_str; service]) ;
+    if wait_until_success then
+      let is_active = Fe_systemctl.is_active service in
+      let success_cond () =
+        match op with Start -> is_active | Stop -> is_active |> not
+      in
+      try
+        Helpers.retry_until_timeout ~timeout
```
lindig, line 39: How is this implemented? We want to avoid a tight loop here.

ocaml/xapi/xapi_systemctl.ml
```diff
@@ -0,0 +1,51 @@
...
+
+module D = Debug.Make (struct let name = "xapi_systemctl" end)
+
+open D
+
+type t = Start | Stop
+
+let to_string = function Start -> "start" | Stop -> "stop"
+
+let operate ~wait_until_success ~service ~timeout op =
```
lindig, line 26: A better name might be `perform`

ocaml/xapi/extauth_plugin_ADwinbind.ml
lippirk: I believe Lin is implementing a pre-existing interface here, so it's probably not worth changing this

ocaml/xapi/extauth_ad.ml
liulinC: Use `Conflicts` directive sounds a clean solution to me,  but 
- if we put this `Conflicts` directive in PBIS (lwsmd), then we need to update PBIS before we remove PBIS,  (we will need to include update PBIS binaries to remove PBIS0, sounds strange!
- if we put this `Conflicts` directive in winbind, we have to setup new patchqueu in winbind, and patch winbind for this.

I do not think it worth a patchqueu just for this temporary solution (PBIS does not live long as we will finally remove it),   so I create ticket in CoP for this and future improvements.
but for now, let just make sure we will not start both by scripts to change backends.

ocaml/xapi/xapi_systemctl.ml
liulinC: I do not think there is a tight loop here
https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/helpers.ml#L1490

lippirk (approved):

ocaml/xapi/extauth_ad.mli
```diff
@@ -0,0 +1,11 @@
+(* start AD external auth backend daemon *)
```
robhoes, line 1: We also need a copyright header on this new file.

psafont (approved):

ocaml/xapi/extauth_ad.mli
liulinC: oh, sorry, I have missed this comment,
I will add a license in the next commit, sorry about this.

PR #4372, opened 2021-04-15
CA-341715: Sync certificates after bringing up mgmnt IF

psafont:
Otherwise the synchronization fails on first boot and puts the host on
maintenance mode.

### DIFF
--- a/ocaml/xapi/xapi.ml
+++ b/ocaml/xapi/xapi.ml
@@ -336,15 +336,12 @@ let update_certificates ~__context () =
   | Ok () ->
       info "successfully synced certificates"
   | Error (`Msg (msg, _)) ->
-      error "Failed to update host certificates: %s" msg ;
-      server_run_in_emergency_mode ()
+      error "Failed to update host certificates: %s" msg
   | exception e ->
-      error "Failed to update host certificates: %s" (Printexc.to_string e) ;
-      server_run_in_emergency_mode ()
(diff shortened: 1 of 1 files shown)


lippirk (approved):

ocaml/xapi/xapi.ml
```diff
@@ -376,6 +373,7 @@ let bring_up_management_if ~__context () =
     ) ;
     (* Start the Host Internal Management Network, if needed. *)
     Xapi_network.check_himn ~__context ;
+    update_certificates ~__context () ;
```
robhoes, line 376: Maybe put this after line 362. That's when we are in the case that stunnel will definitely have started, and we have more reasons to log with `error` if the certs can not be synced.

lippirk: Don't we want to regenerate the bundle _before_ starting the stunnel server? I assume this is one of the things `update_certificates` does

robhoes: I assumed that it just updates DB records.

robhoes: The function name could be made clearer.

psafont: The function updates the database record to match the contents of /etc/xensource/xapi_ssl.pem

robhoes (approved):

PR #4373, opened 2021-04-16
CP-36690 at startup, sync host certs with DB

lindig:
CP-36690 at startup, sync host certs with DB
    
    Each host has two server certificates that identifies it, one for
    internal and external communication each and both are tracked in the
    Xapi database. At Xapi startup, the database entries for them need to be
    kept in sync with the certificates in the file system. This was done so
    far for only one of them (xapi-ssl.pem) but needs to be done for both.
    This commit generalizes the code to do that.


robhoes (approved):
I think this is correct.

psafont (approved):

PR #4374, opened 2021-04-20
CP-36579: Add cache in memory for available updates

minglumlu:
This commit aims to improve the HTTP endpoint `/updates`.

Prior to this commit, HTTP endpoint `/updates` handler will always get
the available updates from all hosts in the pool. This will take quite
long duration.

With this commit, the result of available updates will be stored in a
cache in memory which is established in `pool.sync_updates`. With this
cache, HTTP endpoint `/updates` will return the result from cache
immediately.

Meanwhile, there are a few changes for management of the cache as well.

### DIFF
--- a/ocaml/xapi/repository.ml
+++ b/ocaml/xapi/repository.ml
@@ -24,6 +24,7 @@ module UpdateIdSet = Set.Make (String)
 
 let capacity_in_parallel = 16
 let reposync_mutex = Mutex.create ()
+let updates_in_cache = ref None
 
 let introduce ~__context ~name_label ~name_description ~binary_url ~source_url ~update =
   assert_url_is_valid ~url:binary_url;
@@ -146,16 +147,15 @@ let http_get_host_updates_in_json ~__context ~host ~installed =
         raise Api_errors.(Server_error (get_host_updates_failed, [ref])))
     (fun () -> Xapi_session.destroy_db_session ~__context ~self:host_session_id)
 
-let get_host_updates_by_repository ~__context enabled host () =
+let group_host_updates_by_repository ~__context enabled host updates_of_host =
   (* Return updates grouped by repository. Example:
    * [ ("repo-00", [u0, u1]);
    *   ("repo-01", [u3, u4]);
    *   ... ...
    * ]
    *)
-  let json = http_get_host_updates_in_json ~__context ~host ~installed:false in
   (* TODO: live patches *)
-  match Yojson.Basic.Util.member "updates" json with
+  match Yojson.Basic.Util.member "updates" updates_of_host with
   | `List updates ->
     List.fold_left
       (fun acc update ->
@@ -191,21 +191,29 @@ let get_host_updates_by_repository ~__context enabled host () =
     raise Api_errors.(Server_error (get_host_updates_failed, [ref]))
 
 let set_available_updates ~__context =
-  with_pool_repositories @@ fun () ->
   ignore (get_single_enabled_update_repository ~__context) ;
   let hosts = Db.Host.get_all ~__context in
-  let enabled = get_enabled_repositories ~__context in
   let funs =
-    List.map
-      (fun h -> get_host_updates_by_repository ~__context enabled h)
-      hosts
+    List.map (fun host ->
+        fun () ->
+          ( host, (http_get_host_updates_in_json ~__context ~host ~installed:true) )
+      ) hosts
   in
-  let rets_of_hosts = Helpers.run_in_parallel funs capacity_in_parallel in
+  let rets = with_pool_repositories (fun () ->
+      Helpers.run_in_parallel funs capacity_in_parallel)
+  in
+  updates_in_cache := Some rets ;
+  let enabled = get_enabled_repositories ~__context in
+  (* Group updates by repository for each host *)
+  let updates_of_hosts =
+    List.map (fun (h, updates_of_host) ->
+        group_host_updates_by_repository ~__context enabled h updates_of_host) rets
+  in
   (* Group updates by repository for all hosts *)
   let updates_by_repository =
     List.fold_left
       (fun acc l -> List.fold_left (fun acc' (x, y) -> append_by_key acc' x y) acc l)
-      [] rets_of_hosts
+      [] updates_of_hosts
   in
   let checksums =
     List.filter_map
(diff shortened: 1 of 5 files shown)


ocaml/xapi/repository.ml
```diff
@@ -24,6 +24,7 @@ module UpdateIdSet = Set.Make (String)
 
 let capacity_in_parallel = 16
 let reposync_mutex = Mutex.create ()
+let updates_in_cache = ref None
```
lippirk, line 27: I would use a lock when accessing / setting this ref - can you use `reposync_mutex` for this purpose perhaps?

ocaml/xapi/repository.ml
```diff
@@ -360,34 +368,29 @@ let parse_updateinfo ~__context ~self =
...
-    in
-    `Assoc [
-      ("hosts", `List updates_of_hosts);
-      ("updates", `List (UpdateIdSet.elements ids_of_updates
-                         |> List.map (fun uid  ->
-                             UpdateInfo.to_json (List.assoc uid updates_info))));
-      ("hash", `String (Db.Repository.get_hash ~__context ~self:repository))]
+    (match !updates_in_cache with
+     | None ->
+         raise Api_errors.(Server_error (updates_require_sync, []))
```
lippirk, line 373: Is there a reason not to do the sync here? Then we don't need a new API error

ocaml/xapi/xapi_pool.ml
```diff
@@ -2926,7 +2930,10 @@ let remove_repository ~__context ~self ~value =
     Repository.with_reposync_lock @@ fun () ->
     List.iter
       (fun x ->
-         if x = value then Repository.cleanup_pool_repo ~__context ~self:x)
+         if x = value then (
+           Repository.cleanup_pool_repo ~__context ~self:x ;
+           Repository.reset_updates_in_cache ()
```
lippirk, line 2935: I would have expected `cleanup_pool_repo` to do this reset anyway. Should `set_up_to_date` be doing this reset also? If so then we don't need to expose `reset_updates_in_cache` anymore in the respository.mli file

ocaml/xapi/repository.ml
minglumlu: The `pool.allowed_operation` will protect this cache.

ocaml/xapi/repository.ml
minglumlu: `sync` is a API call. I don't think this should be done in HTTP endpoint `/updates` which is kind of another API. In other words, this would mean a API call contains another API call.

ocaml/xapi/xapi_pool.ml
minglumlu: The case of `set_up_to_date` is actually to add new repositories, which indeed requires `reset_updates_in_cache`.

ocaml/xapi/repository.ml
lippirk: Please can you add a comment to mention that?

minglumlu: Sure. Thanks.

lindig:
I would like to discuss the design of this. If we need to remember the updates for each host, should that not be in the database? I don't like the idea of caches too much unless performance is critical - how often do we need this operation to justify the complexity?

minglumlu:
> I would like to discuss the design of this. If we need to remember the updates for each host, should that not be in the database? I don't like the idea of caches too much unless performance is critical - how often do we need this operation to justify the complexity?

To get the HTTP endpoint `/updates` respond quickly, the updates data have to be be collected in advance as collecting the data may take more than 30 seconds in large pool. The data could be stored either in XAPI DB, or in memory.
And what's more, the data volume of updates is large as well. There may be lots of RPM packages to be updated and they have to be  organised per host.  It was considered in design that storing such large volume of data in XAPI DB is not proper.

robhoes:
Yes, it would certainly be _easier_ to store the updates data in the DB. However, I don't think the current DB is suitable to store large blobs of data.

robhoes (approved):

ocaml/xapi/repository.ml
```diff
@@ -191,21 +192,29 @@ let get_host_updates_by_repository ~__context enabled host () =
...
   ignore (get_single_enabled_update_repository ~__context) ;
   let hosts = Db.Host.get_all ~__context in
-  let enabled = get_enabled_repositories ~__context in
   let funs =
-    List.map
-      (fun h -> get_host_updates_by_repository ~__context enabled h)
-      hosts
+    List.map (fun host ->
+        fun () ->
+          ( host, (http_get_host_updates_in_json ~__context ~host ~installed:true) )
```
edwintorok, line 200: `get_host_updates_by_repository` has previously called `http_get_host_updates_in_json` with `installed:false`, but it is called with `installed:true` here now, why?
I'd assume that installed:true is used here so that the cache will be useful for the updates query below, but I'm not sure about what effects it has on this function...

ocaml/xapi/repository.ml
```diff
@@ -24,6 +24,8 @@ module UpdateIdSet = Set.Make (String)
 
 let capacity_in_parallel = 16
 let reposync_mutex = Mutex.create ()
+(* The cache below is protected by pool's current_operations locking mechanism *)
+let updates_in_cache = ref None
```
edwintorok, line 28: If this is a mutable associative list then why not use a Hashtbl instead? It should be more efficient and have a more natural API to make updates to it than an associative list. If you need a functional datastructure then a reference to a Map might be better suited than an associative list (XAPI tends to use lists a lot, even when something else would be both more efficient and have a better API)

ocaml/xapi/xapi_pool.ml
lippirk: Then can we keep `reset_updates_in_cache` as an implementation detail of `repository.ml`? Based on your comment I don't see a reason to expose it

ocaml/xapi/repository.ml
minglumlu: When it was called `http_get_host_updates_in_json`, it only served for `sync_updates` purpose. `sync_updates` doesn't need the extra data returned with call parameter `installed:true`.
Now `get_host_updates_by_repository` serves for both `sync_udpates` and the cache which will be used by HTTP endpoint `/updates` handler later.

ocaml/xapi/xapi_pool.ml
minglumlu: I mean i.e. in the function `add_repository`, there is no other functions exposed from `module Repository` to cover the reset bits. Basically only those `Db.Repository.*` functions are used in this function.

ocaml/xapi/repository.ml
minglumlu: Thanks. I changed it to `Map`.

minglumlu:
Hi @lippirk @lindig @edwintorok 
Could you please look at my responses and new commit? Thanks.

ocaml/xapi/xapi_pool.ml
edwintorok: If all the repository functions need to call reset they can do that themselves like @lippirk said, so here you could call `Repository.cleanup_pool_repo ~__context ~self:x` and that call would internally call the reset function, and similarly for all the other Repository functions.

minglumlu: Please bear with me as I still can't get the point:
For example, in this function, I can't see which one could take the `reset` bits on?
```
let add_repository ~__context ~self ~value =
  Xapi_pool_helpers.with_pool_operation
    ~__context
    ~self
    ~doc:"pool.add_repository"
    ~op:`configure_repositories
    @@ fun () ->
    Repository.with_reposync_lock @@ fun () ->
    let existings = Db.Pool.get_repositories ~__context ~self in
    if not (List.mem value existings) then (
      Db.Pool.add_repositories ~__context ~self ~value ;
      Db.Repository.set_hash ~__context ~self:value ~value:"" ;
      Db.Repository.set_up_to_date ~__context ~self:value ~value:false ;
      Repository.reset_updates_in_cache ()
    )
```
@lippirk  suggested `set_up_to_date`, but this is a function in `Db`. I don't think I could overwrite it.

robhoes: > If all the repository functions need to call reset they can do that themselves

Some of the functions that need to call reset are implemented here, like `add_repository`. So is does need the reset to be exposed for those. Then for consistency, I'd make all reset calls explicit in the functions in this module (xapi_pool), as Ming has done.

ocaml/xapi/xapi_pool.ml
```diff
@@ -3021,6 +3028,10 @@ let get_updates_handler (req : Http.Request.t) s _ =
                when (List.mem failure failures_of_404) ->
                  error "404: can't get updates for pool: %s" (ExnHelper.string_of_exn e);
                  Http_svr.headers s (Http.http_404_missing ())
+             | Api_errors.(Server_error (failure, _)) as e
+               when not (List.mem failure failures_of_404) ->
+                 error "getting updates for pool failed: %s" (ExnHelper.string_of_exn e);
+                 raise e
```
lindig, line 3034: Is this the right exception to raise? Below a different one is raised for a similar case.

robhoes: I think that this one here is already a proper API error, which can be re-raised. Below is a different exception, which needs to be turned into an API error.

edwintorok (approved):

lindig (approved):

PR #4375, opened 2021-04-20
CP-34467: Pre-join checks for TLS verification

psafont:
Aside from checking that the TLS verification matches in both the joiner and the joined pool, I've removed from the xapi_pool all the list.hd when retrieving the reference of the remote pool and return a better error other than "Not_found".

### DIFF
--- a/ocaml/xapi/xapi_pool.ml
+++ b/ocaml/xapi/xapi_pool.ml
@@ -665,6 +665,26 @@ let pre_join_checks ~__context ~rpc ~session_id ~force =
            ( Api_errors.license_restriction
            , [Features.name_of_feature Features.Pool_size] ))
   in
+  let assert_tls_verification_matches () =
+    let remote_pool = List.hd (Client.Pool.get_all rpc session_id) in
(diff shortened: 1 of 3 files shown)


lippirk (approved):

ocaml/xapi/xapi_pool.ml
```diff
@@ -43,8 +43,16 @@ let rpc host_address xml =
       (Api_errors.Server_error
          (Api_errors.pool_joining_host_connection_failed, []))
 
+let get_pool ~rpc ~session_id =
+  match Client.Pool.get_all ~rpc ~session_id with
+  | [] ->
+      let err_msg = "Remote host does not belong to any pool." in
```
lindig, line 49: "any pool" sounds like there could be many pools but can we have more than one pool?

lindig (approved):

ocaml/xapi/xapi_pool.ml
psafont: I can change it to "a pool". I wish this call would just return a single reference, always. Not sure why it's not this way.

PR #4378, opened 2021-04-22
CP-36085: Join domain with winbind

liulinC:


ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -34,6 +36,91 @@ module Winbind = struct
...
+      raise (Auth_service_error (E_GENERIC, err_msg))
+
+let assert_hostname_valid () =
+  let hostname =
+    (fun __context ->
+      Helpers.get_localhost ~__context |> fun host ->
+      Db.Host.get_hostname ~__context ~self:host)
+    |> Server_helpers.exec_with_new_task "retrieving hostname"
+  in
+  let all_numbers hostname =
```
lindig, line 55: This is fine if it is all we need but these kinds of tests are usually better implemented with regular expression matching as they are more flexible. I'd use the `Re` module for this.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -34,6 +36,91 @@ module Winbind = struct
...
+           ( E_GENERIC
+           , Printf.sprintf "hostname '%s' cannot contain only digits." hostname
+           ))
+  | false ->
+      ()
+
+let assert_domain_equal_service_name ~service_name ~config_params =
+  (* For legeacy support, if domain exist in config_params, it must be equal to service_name *)
+  let domain_key = "domain" in
+  if List.mem_assoc domain_key config_params then
```
lindig, line 71: Use `List.assoc_opt` and do this without `List.mem`

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -34,6 +36,91 @@ module Winbind = struct
...
+  let key = "Pre-Win2k Domain" in
+  let lines =
+    Helpers.call_script ~log_output:On_failure net_cmd
+      ["ads"; "lookup"; "-S"; domain]
+  in
+  match Xapi_cmd_result.of_output_opt ~sep:':' ~key ~lines with
+  | Some v ->
+      v
+  | None ->
+      raise (Auth_service_error (E_LOOKUP, "Failed to lookup domain workgroup"))
```
lindig, line 94: Include the domain in the error message

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -34,6 +36,91 @@ module Winbind = struct
...
+      ; "winbind offline logon = Yes"
+      ; "winbind refresh tickets = Yes"
+      ; "winbind enum groups = no"
+      ; "winbind enum users = no"
+      ; Printf.sprintf "workgroup = %s" workgroup
+      ; "idmap config * : range = 3000000-3999999"
+      ; "idmap config connapp : backend = rid"
+      ; "idmap config connapp : range = 2000000-2999999"
+      ; "idmap config * : backend = tdb"
+      ]
```
lindig, line 117: You might want to add a `""` (empty line) as otherwise the last line won't have a newline at the end. This is not. a problem for most config files, though.

ocaml/xapi/xapi_cmd_result.mli
```diff
@@ -0,0 +1,30 @@
...
+(* Module to parse result from command line output
+ * The result should be key value pairs seperated by sep like ':' *)
+
+type t
+
+(* Empty result set *)
+val empty : t
+
+(* Add one line to the result set *)
+val touch : ?sep:char -> string -> t -> t
```
lindig, line 24: The name is not a good match for the description. Would `add` not be better?

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -34,6 +36,91 @@ module Winbind = struct
     debug "init_service To be implemented in CP-36085"
 end
 
+let net_cmd = "/usr/bin/net"
```
edwintorok, line 39: We usually add commands like these to `xapi_globs.ml`. See here for an example how to add it: https://github.com/xapi-project/xen-api/commit/273339d5f5ff97edf8dc907f291d7d4a6376094a.
Since xapi can work without AD/winbind this should probably go into `nonessential_executables`, so that XAPI doesn't fail to start if it is missing.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -34,6 +36,91 @@ module Winbind = struct
...
+let extract_ou_config ~config_params =
+  try
+    let ou = from_config ~name:"ou" ~err_msg:"" ~config_params in
+    ([("ou", ou)], [Printf.sprintf "createcomputer=%s" ou])
+  with Auth_service_error _ -> ([], [])
+
+let query_domain_workgroup domain =
+  let key = "Pre-Win2k Domain" in
+  let lines =
+    Helpers.call_script ~log_output:On_failure net_cmd
```
edwintorok, line 87: This will raise an exception on failure, I see you intend to handle failure below, in which case you need to catch `Forkhelpers.Spawn_internal_error (stderr, stdout, status) as e`, see `helpers.ml`

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -34,6 +36,91 @@ module Winbind = struct
...
+        "# autogenerated by xapi"
+      ; "[global]"
+      ; "kerberos method = secrets and keytab"
+      ; Printf.sprintf "realm = %s" domain
+      ; "security = ADS"
+      ; "template shell = /bin/bash"
+      ; "winbind offline logon = Yes"
+      ; "winbind refresh tickets = Yes"
+      ; "winbind enum groups = no"
+      ; "winbind enum users = no"
```
edwintorok, line 111: To handle complex inter-domain/forest trusts (like our own CITRITE) I think you need this setting too:
```
winbind use krb5 enterprise principals = yes
```

This should allow you to join the machine to FOO.EXAMPLE.COM, but login with EXAMPLE.COM credentials if trust is properly set up between FOO.EXAMPLE.COM and EXAMPLE.COM (e.g. because you only have logins on EXAMPLE.COM, and nobody has logins on FOO.EXAMPLE.COM, but EXAMPLE.COM doesn't allow joining machines, but FOO.EXAMPLE.COM does)

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -34,6 +36,91 @@ module Winbind = struct
...
+      ; Printf.sprintf "workgroup = %s" workgroup
+      ; "idmap config * : range = 3000000-3999999"
+      ; "idmap config connapp : backend = rid"
```
edwintorok, line 114: What does `connapp` mean here?

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: Yes, this is what we need and what PBIS currently does,  I would like to keep this behavior for backward compatibility,
but regular expression sounds a better solution and I would like to use it.
btw, For expression  in ocaml, I only found https://ocaml.org/api/Str.html#VALstring_match
Can you give me a link to refer the `Re` module?

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -105,7 +192,45 @@ module AuthADWinbind : Auth_signature.AUTH_MODULE = struct
...
+
+    let workgroup = query_domain_workgroup service_name in
+    config_winbind_damon service_name workgroup ;
+
+    let args =
+      ["ads"; "join"; service_name; "-U"; user; "--no-dns-updates"] @ ou_param
+    in
+    debug "Joining domain %s with user %s and workgroup: %s" service_name user
+      workgroup ;
+    let env = [|Printf.sprintf "PASSWD=%s" pass|] in
```
edwintorok, line 226: Please make sure this doesn't get logged, even on failure (on failure forkexecd likes to print its debug output, which might include env vars).

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: nice catch, this is my testing domain name!
I will update it to the domain name dynamically

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: The code does not print the env, https://github.com/xapi-project/forkexecd/blob/master/lib/forkhelpers.ml#L291
and I mocked a failure in my local dev environment, and It do not print the password on failure.

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: The document https://www.samba.org/samba/docs/current/man-html/smb.conf.5.html says:
```
It is not really required to know about the trust topology. We can just rely on the [K]DCs of our primary domain (e.g. PRIMARY.A.EXAMPLE.COM) and use enterprise principals e.g. upnfromB@B.EXAMPLE.COM@PRIMARY.A.EXAMPLE.COM and follow the WRONG_REALM referrals in order to find the correct DC. The final principal might be userfromB@INTERNALB.EXAMPLE.PRIVATE.
```
But set this to `yes` looks like make things better ?

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: The config file works perfectly without empty line, and I would like to just keep it.

ocaml/xapi/extauth_plugin_ADwinbind.ml
edwintorok: Yes I think trust topology is complicated and `winbindd` may not be able to get the full picture, so when you set that config option it'll rely on the (K)DC servers telling it where it should go to authenticate the user.
I was not able to join a machine and authenticate to our internal AD without this setting. Have you tried with and without this setting?

ocaml/xapi/extauth_plugin_ADwinbind.ml
lippirk: you can write `Helpers.call_script ~log_output:Never ...` if you want to be doubly sure

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -105,7 +192,45 @@ module AuthADWinbind : Auth_signature.AUTH_MODULE = struct
...
+        ~config_params
+    in
+
+    assert_hostname_valid () ;
+
+    let service_name =
+      (fun __context ->
+        Helpers.get_localhost ~__context |> fun host ->
+        Db.Host.get_external_auth_service_name ~__context ~self:host)
+      |> Server_helpers.exec_with_new_task
```
lippirk, line 211: what's the rationale for creating a new task here? (particularly for such a small piece of work)

ocaml/xapi/xapi_cmd_result.mli
```diff
@@ -0,0 +1,30 @@
...
+
+(* Find result from the result set *)
+val find : string -> t -> string option
+
+(* find result from the whole result lines *)
+val of_output_opt : sep:char -> key:string -> lines:string -> string option
```
lippirk, line 30: I would like to see some unit tests for these functions, as it's not clear to me exactly what the format of `lines` is supposed to be

lippirk, line 30: I would like to see a unit test for this function, as it's not clear to me exactly what the format of `lines` is supposed to be

ocaml/xapi/xapi_cmd_result.mli
```diff
@@ -0,0 +1,30 @@
...
+type t
+
+(* Empty result set *)
+val empty : t
+
+(* Add one line to the result set *)
+val touch : ?sep:char -> string -> t -> t
+
+(* Find result from the result set *)
+val find : string -> t -> string option
```
lippirk, line 27: Are you using any of these outside of the context of the the `xapi_cmd_result.ml` file? If not, you can remove them from the .mli file. The idea of the .mli file is that you only expose what is necessary, in order to provide a clean interface

ocaml/xapi/extauth_plugin_ADwinbind.ml
lippirk: You can see an example here https://github.com/xapi-project/xen-api/blob/ea27a74a532d2e881f8a88a55319287dc5302e7b/ocaml/xapi/create_misc.ml#L113

ocaml/xapi/xapi_cmd_result.mli
liulinC: some function is not used by me, like `touch` and `find`,  
but I take it as general module and provide functions that is necessary for future usage.

ocaml/xapi/xapi_cmd_result.mli
liulinC: I had a unittest file `test_xapi_cmd_result.ml` in this commit

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: I port such code pieces from PBIS, 
for me, just want to refer the argument `__context` from new task, 
is there any other way I can get `__context` argument in this function?

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: thanks @lippirk  for the help, but I maybe better to log the output on failure to help identify problems.

ocaml/xapi/extauth_plugin_ADwinbind.ml
lippirk: ah i see, it is strange the existing interface does not provide you with a context. There isn't another way really

Personally I would wrap the whole `on_enable` function in a task, and call it `on_enable`, but this is fine

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: I do not find other files need to refer this command, so I put in the this file other than the global file, 
also, the `nonessential_executables` table contains command configurable, while this command not.
but I will follow the suggestion if that is what tool stack team does.

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: I will follow the suggestion and give it a try

ocaml/xapi/xapi_cmd_result.mli
liulinC: The `find` is what currently I used, and it extract a value from the all output for `key`
However, there are a chance, that the output contains lots of key value pairs,  and we may want to query several values basing on `key`s,  and we do not want to execute the command several times.

In this case, `add` and `empty` will be useful to let the caller parses line by line.

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: I had a tried and updated the details in the channel.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -34,6 +38,118 @@ module Winbind = struct
...
+  Unixext.atomic_write_to_file smb_config 0o0644 (fun fd ->
+      let (_ : int) = Unix.single_write_substring fd conf_contents 0 len in
+      ())
+
+let from_config ~name ~err_msg ~config_params =
+  match List.assoc_opt name config_params with
+  | Some v ->
+      v
+  | _ ->
+      raise (Auth_service_error (E_GENERIC, err_msg))
```
lindig, line 96: You could include the key that was not found

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -34,6 +38,118 @@ module Winbind = struct
...
+
+let assert_hostname_valid () =
+  let hostname =
+    (fun __context ->
+      Helpers.get_localhost ~__context |> fun host ->
+      Db.Host.get_hostname ~__context ~self:host)
+    |> Server_helpers.exec_with_new_task "retrieving hostname"
+  in
+  let all_numbers hostname =
+    Re.Perl.re {|^\d+$|} |> Re.Perl.compile |> fun re ->
```
lindig, line 106: You could do the compilation just once by binding the result to an identifier outside the function and only do the match here.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -34,6 +38,118 @@ module Winbind = struct
...
+         ( E_GENERIC
+         , Printf.sprintf "hostname '%s' cannot contain only digits." hostname
+         ))
+
+let assert_domain_equal_service_name ~service_name ~config_params =
+  (* For legeacy support, if domain exist in config_params, it must be equal to service_name *)
+  let domain_key = "domain" in
+  match List.assoc_opt domain_key config_params with
+  | Some domain ->
+      if domain <> service_name then
```
lindig, line 121: This could be expressed with `when` and an extra case that makes more explicit what happens if `domain = service_name`. Here we have a pattern and an if without an else.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -34,6 +38,118 @@ module Winbind = struct
...
+  | Some u, Some p -> (
+      (* Clean machine account in DC *)
+      let env = [|Printf.sprintf "PASSWD=%s" p|] in
+      let args = ["ads"; "leave"; "-U"; u] in
+      try
+        Helpers.call_script ~env net_cmd args |> ignore ;
+        debug "Clean machine account in domain %s successfully" service_name
+      with _ ->
+        let msg =
+          Printf.sprintf "Failed to clean machine account in domain  %s"
```
lindig, line 144: Maybe "Failed the clean machine count" as "clean machine count" is not a verb" or "failed to perform a …"

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: the `err_msg` should contain what is necessary, like some key is required

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: `machine account` is none,  does `clean the machine account` make sense to you?

lindig: Sorry, I had misread that but I'm still not sure what the command does. Is it cleaning the machine account? If so, it "Cleaning the machine account succeeded", "Cleaning the machine account failed" would make sense to me.

lindig (approved):

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: The command will leave domain, and it will clean the machine account in Domain Controller for us.  (and destroy the machine password)

liulinC: updated

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -34,6 +38,128 @@ module Winbind = struct
...
+
+let assert_hostname_valid () =
+  let hostname =
+    (fun __context ->
+      Helpers.get_localhost ~__context |> fun host ->
+      Db.Host.get_hostname ~__context ~self:host)
+    |> Server_helpers.exec_with_new_task "retrieving hostname"
+  in
+  let all_numbers hostname = Re.matches all_number_re hostname <> [] in
+  if all_numbers hostname then
```
lippirk, line 109:
```suggestion
  let all_numbers = Re.matches all_number_re hostname <> [] in
  if all_numbers then
```

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -34,6 +38,128 @@ module Winbind = struct
...
+let from_config ~name ~err_msg ~config_params =
+  match List.assoc_opt name config_params with
+  | Some v ->
+      v
+  | _ ->
+      raise (Auth_service_error (E_GENERIC, err_msg))
+
+let all_number_re = Re.Perl.re {|^\d+$|} |> Re.Perl.compile
+
+let assert_hostname_valid () =
```
lippirk, line 101: I am fine with keeping this assertion, I'm just wondering why we need it (also it'd be better to make the assertion _before_ putting the hostname in the database)

lippirk (approved):

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: well, It is from PBIS, see https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/extauth_plugin_ADpbis.ml#L884

I had some test with winbind, and found join domain can still success even the hostname are all numbers, 
let's just keep it for now for backward compatibilities, and remove it on demand.

PR #4379, opened 2021-04-22
CP-36750: Block enabling TLS verification on pool ops

psafont:
It must not be run concurrently with other pool operations as it might
disrupt the certificate distribution. It must not be run when HA is
enabled as it might fence hosts and change the master while the
operation is in progress.

Confirm it works when HA is enabled:
```
# xe pool-param-get uuid=6991f2ea-9674-5997-fce7-a26122ad838a param-name=ha-enabled 
true
# xe pool-enable-tls-verification 
The operation could not be performed because HA is enabled on the Pool
```

### DIFF
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -929,14 +929,18 @@ functor
         let self = Helpers.get_pool ~__context in
         let local_fn = Local.Pool.enable_tls_verification in
         let all_hosts = Xapi_pool_helpers.get_master_slaves_list ~__context in
-        Db.Pool.set_tls_verification_enabled ~__context ~self ~value:true ;
-        Cert_distrib.(
-          go ~__context ~existing_cert_strategy:Erase_old ~from_hosts:all_hosts
-            ~to_hosts:all_hosts) ;
-        all_hosts
-        |> List.iter (fun host ->
-               do_op_on ~local_fn ~__context ~host (fun session_id rpc ->
-                   Client.Pool.enable_tls_verification rpc session_id))
+
+        Xapi_pool_helpers.with_pool_operation ~__context
+          ~doc:"Pool.enable_tls_verification" ~self ~op:`tls_verification_enable
+          (fun () ->
+            Cert_distrib.(
+              go ~__context ~existing_cert_strategy:Erase_old
+                ~from_hosts:all_hosts ~to_hosts:all_hosts) ;
+            all_hosts
+            |> List.iter (fun host ->
+                   do_op_on ~local_fn ~__context ~host (fun session_id rpc ->
+                       Client.Pool.enable_tls_verification rpc session_id)) ;
+            Db.Pool.set_tls_verification_enabled ~__context ~self ~value:true)
     end
 
     module VM = struct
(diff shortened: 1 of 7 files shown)


ocaml/xapi/message_forwarding.ml
```diff
@@ -929,14 +929,18 @@ functor
...
+          ~doc:"Pool.enable_tls_verification" ~self ~op:`tls_verification_enable
+          (fun () ->
+            Cert_distrib.(
+              go ~__context ~existing_cert_strategy:Erase_old
+                ~from_hosts:all_hosts ~to_hosts:all_hosts) ;
+            all_hosts
+            |> List.iter (fun host ->
+                   do_op_on ~local_fn ~__context ~host (fun session_id rpc ->
+                       Client.Pool.enable_tls_verification rpc session_id)) ;
+            Db.Pool.set_tls_verification_enabled ~__context ~self ~value:true)
```
lippirk, line 943: :+1: setting this flag last is definitely more sensible - now if you run `Pool.enable_tls_verification` and it fails, the db state will be correct

ocaml/xapi/message_forwarding.ml
```diff
@@ -929,14 +929,18 @@ functor
...
         let all_hosts = Xapi_pool_helpers.get_master_slaves_list ~__context in
-        Db.Pool.set_tls_verification_enabled ~__context ~self ~value:true ;
-        Cert_distrib.(
-          go ~__context ~existing_cert_strategy:Erase_old ~from_hosts:all_hosts
-            ~to_hosts:all_hosts) ;
-        all_hosts
-        |> List.iter (fun host ->
-               do_op_on ~local_fn ~__context ~host (fun session_id rpc ->
-                   Client.Pool.enable_tls_verification rpc session_id))
+
```
lippirk, line 932: My understanding is that users must have turned HA off before using `Pool.enable_tls_verification`. Is that correct?

If so, I would steal `assert_no_ha` and `assert_no_rpu` from xapi_psr.ml and use them here - I think `with_pool_operation` will only block if ha is _being_ enabled, not if it _is_ enabled (though it would be best to double-check). See https://github.com/xapi-project/xen-api/blob/a214de185c97178a6fa2c8a9d08a45624b3b6886/ocaml/xapi/xapi_psr.ml#L540

ocaml/xapi/xapi_pool_helpers.ml
```diff
@@ -63,9 +72,11 @@ let valid_operations ~__context record _ref' =
...
-    set_errors Api_errors.ha_is_enabled [] [`ha_enable]
-  else
+  if ha_enabled then (
+    set_errors Api_errors.ha_is_enabled [] [`ha_enable] ;
+    (* TLS verification is not allowed to run if HA is enabled *)
+    set_errors Api_errors.ha_is_enabled [] [`tls_verification_enable]
+  ) else
     set_errors Api_errors.ha_not_enabled [] [`ha_disable] ;
   (* cluster create cannot run during a rolling pool upgrade *)
   if Helpers.rolling_upgrade_in_progress ~__context then
```
lippirk, line 82: It may be prudent to block `Pool.enable_tls_verification` when an RPU is in progress

ocaml/xapi/message_forwarding.ml
psafont: Yup. For increased resiliency public-key exchange happens first, then it's enabled in the hosts and finally the flag is set on the database. Each stage until the last is reversible.

lindig (approved):

lippirk (approved):

PR #4380, opened 2021-04-23
CA-329462 Cluster.create should clean up if it fails

lippirk:
This was a known bug (raised as part of XSI-501), but I have been
running into it a lot recently while testing certificate checking in
xapi-clusterd.

There are two obvious places where we might fail:
a) starting the cluster daemon on the master
b) telling all the slaves to start their cluster daemons and to join
   cluster

Previously, if either of these fails, we could be left with cluster
daemons running unnecessarily, and / or cluster records lying around in
the database. You are prevented from enabling clustering again until
these records are deleted manually, which is annoying.


lindig (approved):
Looks good to me.

psafont (approved):

PR #4384, opened 2021-04-27
CP-34467: Exchange certificate when a hosts joins a pool

psafont:
After passing the pre-join checks the joiner and the pool exchange
certificates and place them in the filesystem, ready to be used. The
database is not modified for any of the two sides. When rebooting xapi's
startup code in the joining host will detect its internal certificate is
not in the database and will add it.

Current code does not try to merge the trusted certificates that exist
in the joining host and the pool.

I've tested it manually successfully, I'll try to throw automated testing to make sure it's robust

### DIFF
--- a/ocaml/idl/datamodel_pool.ml
+++ b/ocaml/idl/datamodel_pool.ml
@@ -75,6 +75,21 @@ open Datamodel_types
       ~allowed_roles:_R_POOL_OP
       ()
 
+  (* This is a map of uuid -> cert_blob *)
+  let pool_certs = Map (String, String)
+
+  let exchange_certificates_on_join = call
+      ~name:"exchange_certificates_on_join"
+      ~in_oss_since:None
+      ~in_product_since:rel_next
+      ~params:[String, "uuid", "The uuid of the joining host";
+               String, "certificate", "The contents of the joiner's certificate";
+              ]
+      ~result:(pool_certs, "The contents of the pool's certificates")
+      ~doc:"Install the pool certificate of a joiner and return the pool's certificates"
+      ~hide_from_docs:true
+      ~allowed_roles:_R_POOL_OP
+      ()
 
   let slave_reset_master = call ~flags:[`Session]
       ~name:"emergency_reset_master"
@@ -704,6 +719,7 @@ open Datamodel_types
         ; join_force
         ; eject
         ; initial_auth
+        ; exchange_certificates_on_join
         ; transition_to_master
         ; slave_reset_master
         ; recover_slaves
--- a/ocaml/xapi/cert_distrib.ml
+++ b/ocaml/xapi/cert_distrib.ml
@@ -76,6 +76,14 @@ module Worker : sig
     existing_cert_strategy -> WireProtocol.cert list -> unit remote_call
 
   val remote_regen_bundle : unit remote_call
+
+  val local_write_cert_fs :
+       __context:Context.t
+    -> existing_cert_strategy
+    -> WireProtocol.cert list
+    -> unit
+
+  val local_regen_bundle : __context:Context.t -> unit
 end = struct
   open WireProtocol
 
@@ -203,6 +211,22 @@ end = struct
         ()
     | r ->
         unexpected_result "remote_regen_bundle" r
+
+  let local_write_cert_fs ~__context strategy certs =
+    let command = WireProtocol.string_of_command (Write (strategy, certs)) in
+    match result_or_fail (local_exec ~__context ~command) with
+    | WriteResult ->
+        ()
+    | r ->
+        unexpected_result "local_write_certs_fs" r
+
+  let local_regen_bundle ~__context =
+    let command = WireProtocol.string_of_command GenBundle in
+    match result_or_fail (local_exec ~__context ~command) with
+    | GenBundleResult ->
+        ()
+    | r ->
+        unexpected_result "local_regen_bundle" r
 end
 
 let local_exec = Worker.local_exec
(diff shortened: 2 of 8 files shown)


ocaml/xapi/xapi_pool.ml
```diff
@@ -1357,6 +1388,16 @@ let join_force ~__context ~master_address ~master_username ~master_password =
...
 
+let pre_join_exchange_certificates ~__context ~uuid ~certificate :
+    API.string_to_string_map =
+  let to_hosts = Xapi_pool_helpers.get_master_slaves_list ~__context in
+  let joiner_certificate =
+    Cert_distrib.make_joiner_cert ~__context ~uuid ~certificate
+  in
+  Cert_distrib.import_joiner ~__context ~joiner_certificate ~to_hosts ;
+  Cert_distrib.make_pool_certs ~__context
+  |> List.map Cert_distrib.host_cert_to_string_to_string
```
lippirk, line 1399: I think all of this should be internal to `cert_distrib.ml`, then you can get away with exposing a single function in `cert_distrib.mli`. something like:

```
val send_joiner_certs_to_pool : __context -> certificate -> to_hosts -> map
```

ocaml/idl/datamodel_pool.ml
```diff
@@ -75,6 +75,21 @@ open Datamodel_types
       ~allowed_roles:_R_POOL_OP
       ()
 
+  (* This is a map of uuid -> cert_blob *)
+  let pool_certs = Map (String, String)
+
+  let pre_join_exchange_certificates = call
```
lippirk, line 81: I find it a bit strange that some of the cert distrib is done here, and the rest is done in cert_distrib.ml. I would prefer if it were all done in cert_distrib.ml (i think the way to do that would be to add command to `WireProtocol.command`), or alternatively if all the commands in cert_distrib.ml were made part of the datamodel

lindig: Linguistically this is a bit awkward. Should this be a "pre-join certificates exchange" or a "certificates pre-join check"?

ocaml/idl/datamodel_pool.ml
```diff
@@ -75,6 +75,21 @@ open Datamodel_types
...
+      ~name:"pre_join_exchange_certificates"
+      ~in_oss_since:None
+      ~in_product_since:rel_next
+      ~params:[String, "uuid", "The uuid of the joiner";
```
lindig, line 85: uuid of the joining host (joiner) - to make explicit the kind of the uuid

ocaml/idl/datamodel_pool.ml
```diff
@@ -75,6 +75,21 @@ open Datamodel_types
...
+               String, "certificate", "The contents of the joiner's certificate";
```
lindig, line 86: I assume this is only the public part of the certificate. Mention this?

ocaml/xapi/certificates.ml
```diff
@@ -396,18 +396,23 @@ let pool_uninstall kind ~__context ~name =
 
 (* Extracts the server certificate from the server certificate pem file.
    It strips the private key as well as the rest of the certificate chain. *)
-let get_server_certificate () =
-  match Gencertlib.Pem.parse_file !Xapi_globs.server_cert_path with
+let get_public_certficate_from_pkcs12 path =
```
lindig, line 399: For me `get` implies more effort than a `read`. In some way this is true because we are parsing the file but I would still this consider a read, like a database or filesystem read. I would use `get` for something that involves a network request or user interaction.

lindig (approved):

ocaml/idl/datamodel_pool.ml
psafont: I think I may be able to avoid changing the datamodel by reusing the untrusted connection that the joiner creates to direct the function call of `cert_distrib_atom` to the pool.

psafont: Done!

ocaml/idl/datamodel_pool.ml
psafont: I've removed the new API call

ocaml/xapi/cert_distrib.mli
```diff
@@ -17,3 +17,14 @@ val go :
...
+type host_cert
+
+val exchange_certificates_with_pool :
+     rpc:(Rpc.call -> Rpc.response Client.Id.t)
+  -> session_id:API.ref_session
+  -> uuid:string
+  -> certificate:string
+  -> host_cert list
+
+val trust_joined_pool : __context:Context.t -> pool_certs:host_cert list -> unit
```
lippirk, line 30: Why can't we merge these two functions? Is it because `exchange_certificates_with_pool` requires an 'unverified' rpc, yet `trust_joined_pool` doesn't?

Would it be possible to handle all of this inside cert_distrib.ml as well? (from a high level i expect one function `val distribute_certificates_between_new_joiner_and_pool : unverified_rpc -> __context -> joiner_uuid -> ... etc ... -> unit`)

psafont: > Why can't we merge these two functions? 

I just didn't think so far ahead while simplifying the code :)

lippirk: It's looking good now! Thanks

lippirk (approved):

lindig:
I'd suggest we squash the fixups before merging.

psafont:
I have to roll back the removal of the additional API: the current API call for cert distribution is only allowed to be used within a pool. I would like to keep it that way as makes it easer to think about the context where it's used: it's only used among hosts in a pool and cannot be triggered directly through any API client.

ocaml/idl/datamodel_pool.ml
psafont: Technically the private key is not part of the certificate, the format used to stored the certificate along the private key is PKCS #12. Note that the certificate records never contain any information about private keys.

lippirk (approved):

psafont:
I've validated the changes are working and it's good to reviewed and merged

ocaml/xapi/cert_distrib.ml
```diff
@@ -203,40 +211,53 @@ end = struct
...
     | r ->
         unexpected_result "remote_regen_bundle" r
+
+  let local_write_cert_fs ~__context strategy certs =
+    let command = WireProtocol.string_of_command (Write (strategy, certs)) in
+    match result_or_fail (local_exec ~__context ~command) with
+    | WriteResult ->
+        ()
+    | r ->
+        unexpected_result "local_write_certs_fs" r
```
lindig, line 221: There is a new `__FUNCTION__` macro in OCaml 4.12 but we can't use it yet.

ocaml/xapi/cert_distrib.ml
```diff
@@ -245,3 +266,37 @@ let go ~__context ~from_hosts ~to_hosts ~existing_cert_strategy =
   List.iter
     (fun host -> Worker.remote_regen_bundle host rpc session_id)
     to_hosts
+
```
lindig, line 269: The names `from_host` and `to_host` are a bit odd, `source` and `destination` (`src`, `dst`) feel more natural to me.

ocaml/xapi/cert_distrib.ml
```diff
@@ -245,3 +266,37 @@ let go ~__context ~from_hosts ~to_hosts ~existing_cert_strategy =
...
+  List.iter
+    (fun host ->
+      Worker.remote_write_certs_fs Merge [joiner_certificate] host rpc
+        session_id)
+    to_hosts ;
+  List.iter
+    (fun host -> Worker.remote_regen_bundle host rpc session_id)
+    to_hosts
+
+(* This function is called in the host that is joining a pool *)
```
lindig, line 290: "is called _on_ the host"?

ocaml/xapi/cert_distrib.ml
```diff
@@ -245,3 +266,37 @@ let go ~__context ~from_hosts ~to_hosts ~existing_cert_strategy =
...
+let import_joining_pool_certs ~__context ~pool_certs =
+  let pool_certs =
+    List.map (fun (uuid, blob) -> WireProtocol.{uuid; blob}) pool_certs
+  in
+  Worker.local_write_cert_fs ~__context Merge pool_certs ;
+  Worker.local_regen_bundle ~__context
+
+(* This function is called in the pool that is incorporating a new host *)
```
lindig, line 298: "accepting"?

lindig (approved):

robhoes (approved):
I have reviewed the high-level changes in the pool join code, which look good to me. I'll leave the details about the exchange mechanism in cert_distrib.ml to the other reviewers.

ocaml/xapi/cert_distrib.ml
psafont: I think it makes the code to be read in a quite natural way so I tried to keep that style.

ocaml/xapi/cert_distrib.ml
psafont: I'd say the pool doesn't have much choice in the matter. ;)

PR #4387, opened 2021-04-29
CA-353553 add API error for when NVidia GPU is misconfigured

lindig:
Xapi sets up GPU types based on an XML configuration that tells it
whether a card is SR-IOV capable. When we actually use the card and
enable virtual functions and find the card is not correctly configured
for SR-IOV, we raised so far an internal error. This patch adds a new
API error to report the case more clearly.

### DIFF
--- a/ocaml/xapi/xapi_pgpu.ml
+++ b/ocaml/xapi/xapi_pgpu.ml
@@ -450,8 +450,13 @@ let nvidia_vf_setup ~__context ~pf ~enable =
   let activate_vfs pci =
     match num_vfs pci with
     | None ->
-        let msg = sprintf "Can't determine number of VFs for PCI %s" pci in
-        raise (fail msg)
+        let host = Db.PCI.get_host ~__context ~self:pf in
+        let device = Db.PCI.get_device_name ~__context ~self:pf in
+        error "Can't determine number of VFs for PCI %s" pci ;
+        raise
+          Api_errors.(
+            Server_error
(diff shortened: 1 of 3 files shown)


psafont (approved):

ocaml/xapi/xapi_pgpu.ml
```diff
@@ -450,8 +450,13 @@ let nvidia_vf_setup ~__context ~pf ~enable =
...
     | None ->
-        let msg = sprintf "Can't determine number of VFs for PCI %s" pci in
-        raise (fail msg)
+        let host = Db.PCI.get_host ~__context ~self:pf in
+        let host_uuid = Db.Host.get_uuid ~__context ~self:host in
+        let device = Db.PCI.get_device_name ~__context ~self:pf in
+        error "Can't determine number of VFs for PCI %s" pci ;
+        raise
+          Api_errors.(
+            Server_error (nvidia_sriov_misconfigured, [host_uuid; device]))
```
robhoes, line 459: We should return refs in API errors rather than UUIDs.

lindig: Will fix it. The seconds parameter is a string but I could also return something else like the PCI ref but thought this was less helpful

robhoes: Think about whether the string is useful to an API client, which might want to parse it. In this case, a name may be fine if it is just for presenting info to the user, but it won't be as unambigious as e.g. the PCI bus ID.

robhoes (approved):

PR #4395, opened 2021-05-05
CP-34198: Start membership naming for pool hosts

psafont:
Pool hosts are now called 'members', with one of them being the master
host.

There's no consensus on what to do with the term master, but this naming
scheme allows us to stop using the term slave.

Citrix's writing guides is not very self-explanatory,it would be interesting if people would propose alternatives to "master" in any case.
 https://brand.citrix.com/application-guides/writing/unbiased-language/#section_10

### DIFF
--- a/ocaml/xapi/xapi_cluster.ml
+++ b/ocaml/xapi/xapi_cluster.ml
@@ -13,6 +13,7 @@
  *)
 
 open Xapi_clustering
+module Client = Client.Client
 
 module D = Debug.Make (struct let name = "xapi_cluster" end)
 
@@ -46,7 +47,7 @@ let create ~__context ~pIF ~cluster_stack ~pool_auto_join ~token_timeout
       let cluster_host_uuid = Uuid.(to_string (make ())) in
       (* For now we assume we have only one pool
          TODO: get master ref explicitly passed in as parameter*)
-      let host = Helpers.get_master ~__context in
+      let host = Helpers.get_coordinator ~__context in
       let pifrec = Db.PIF.get_record ~__context ~self:pIF in
       assert_pif_prerequisites (pIF, pifrec) ;
       let ip = ip_of_pif (pIF, pifrec) in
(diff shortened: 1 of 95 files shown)


lindig:
I think it's important to state whether semantics change or not. In my understanding a slave is a member but not a master. Whereas any host (master or not) is a member of a pool.

lippirk:
I would also be happy with master -> primary, slave -> secondary like the branding says, but slave -> member is probably fine

lippirk (approved):

lindig (approved):

ocaml/xapi/xapi_cluster.ml
```diff
@@ -151,7 +153,7 @@ let pool_destroy_common ~__context ~self ~force =
         List.filter (( <> ) master_ch) all_hosts
   in
   foreach_cluster_host ~__context ~self ~log:force
-    ~fn:Client.Client.Cluster_host.destroy slave_cluster_hosts
```
edwintorok, line 154: There is another instance of `slave` here

psafont: Yes, there are a lot of mentions to slaves around the whole codebase. I concentrated the efforts of this PR around xapi_pool_helpers and its users. I can extend it a bit more if we find the terminology good enough to use.

psafont:
Converted the changes to use primary / secondary as it's being picked up by more people. I've also replaced a lot of usaged. There are about 55 more files with mentions to slave, there are quite a few that cannot be removed because they affect the API or because they don't refer to host but to other concepts that don't involve host (like bridges), which we don't have control of.

This is a starting pass and not all usages have been changed. The networking module specially is difficult to change because it mixes different types of "slaves" and it's not clear if the name or comment is referring to hosts or not.

psafont:
Drafting this as a decisions needs to be reached among the whole hypervisor team about which terms to use

psafont:
Closing this, it's obvious we don't have capacity for this currently.

PR #4396, opened 2021-05-06
CP-35348 cover alerts for internal and CA certificates

lindig:
Extend the alert mechansim for pool-internal and CA certificates.

### DIFF
--- a/ocaml/alerts/certificate/certificate_check.ml
+++ b/ocaml/alerts/certificate/certificate_check.ml
@@ -1,7 +1,17 @@
 module XenAPI = Client.Client
 module Date = Xapi_stdext_date.Date
 
+type cert =
+  | CA of API.ref_pool * API.datetime
+  | Host of API.ref_host * API.datetime
+  | Internal of API.ref_host * API.datetime
+
 let seconds_per_day = 3600. *. 24.
+
+let internal_error fmt =
+  fmt
+  |> Printf.kprintf @@ fun msg ->
+     raise Api_errors.(Server_error (internal_error, [msg]))
 
 let days_until_expiry epoch expiry =
   let days = (expiry /. seconds_per_day) -. (epoch /. seconds_per_day) in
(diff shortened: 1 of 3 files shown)


ocaml/alerts/certificate/certificate_check.ml
```diff
@@ -1,8 +1,17 @@
 module XenAPI = Client.Client
 module Date = Xapi_stdext_date.Date
 
+type entity = Host of API.ref_host | Pool of API.ref_pool
+
+type cert = {entity: entity; expiry: API.datetime}
```
psafont, line 6: There are three different messages, depending on the type of certificate. The current code only differentiates between two types. The certificates for internal communication use different messages because the way to resolve its expiry is different from the normal one.

> `HOST_INTERNAL_CERTIFICATE_EXPIRING_30`, `HOST_INTERNAL_CERTIFICATE_EXPIRING_14`, `HOST_INTERNAL_CERTIFICATE_EXPIRING_07`, `HOST_INTERNAL_CERTIFICATE_EXPIRED`: Similar to host certificates, the newly-introduced hosts' internal server certificates are monitored for expiry dates and alerts are generated about them. The body for this type of message is:
> `<body><message>The TLS server certificate for internal communications {is expiring soon|has expired}.</message><date>20210302T02:00:01Z</date></body>`

lindig: The existing code is tedious by not being parameterized over type of certificate and expiration and we are increasing this problem. The internal certificate is always crated by us and I think a simpler handling can be justified.

psafont: I've tried to make the handling simpler, the big issue here is the lack of changing the certificate provided by stunnel while it's running. This means it needs to be restarted, this means breaking active connections, with the risks of interrupting operations.

I would rather make users be aware of this fact rather than trying to automate it and risk corrupting data.

psafont (approved):
This looks good, one tricky bit is finding outdated alerts using the uuid of the object the certificate refers to, which can now be a host or a pool. The code assumes uuids are unique even among classes of objects which is fine.

lindig:
It's in the name of `uuid`; if we reuse UUIDs, this is another problem.

lippirk (approved):

PR #4397, opened 2021-05-06
CP-35523: Block access to the website on port 80

robhoes:
By default, only HTTPS is allowed, as a small step to improve security.
This only affects the web content on the GET / handler (for the time
being).  There is a config file option to re-enable HTTP access if
required.

### DIFF
--- a/ocaml/xapi/fileserver.ml
+++ b/ocaml/xapi/fileserver.ml
@@ -85,27 +85,35 @@ let send_file (uri_base : string) (dir : string) (req : Request.t)
   let uri_base_len = String.length uri_base in
   let s = Buf_io.fd_of bio in
   Buf_io.assert_buffer_empty bio ;
-  let uri = req.Request.uri in
-  try
-    let relative_url =
-      String.sub uri uri_base_len (String.length uri - uri_base_len)
-    in
-    (* file_path is the thing which should be served *)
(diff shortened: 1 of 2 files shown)


edwintorok:
We could also add a config flag to reject API requests on http, leaving only the bare minimum handlers that currently don't support https yet on http.

robhoes:
We could, yes, but switching off HTTP for more things would require more testing and communication. And ideally we'd just invest the time in making sure we can block HTTP for everything. This one is just a quick win, since the website doesn't really contain anything interesting anymore.

edwintorok (approved):

lindig (approved):

psafont (approved):
xcp-ng uses this page as an entry-point to inform the user on how they can manage the host, it includes an installer for the xen orchestra appliance.

Browsers will show a warning because of the self-signed certificate installed by default. Not much can be done in this case other than documenting that it's normal to see the wanring.

robhoes:
> xcp-ng uses this page as an entry-point to inform the user on how they can manage the host, it includes an installer for the xen orchestra appliance.

For xcp-ng, they might want to use the new config file option (`website-https-only = false`) to re-enable HTTP, if the certificate warning for HTTPS is undesirable.

PR #4400, opened 2021-05-07
CA-334758 - Allow vbds from sds to be plugged into dom0

BenSimsCitrix:
I need to do run some BSTs on this code hence the no merge.

This change in conjunction with xenopsd changes has allowed the with-vdi testing to progress to the next stage.

### DIFF
--- a/ocaml/xapi/xapi_vbd.ml
+++ b/ocaml/xapi/xapi_vbd.ml
@@ -77,9 +77,14 @@ let plug ~__context ~self =
   let domid = Int64.to_int (Db.VM.get_domid ~__context ~self:vm) in
   let force_loopback_vbd = Helpers.force_loopback_vbd ~__context in
   let hvm = Helpers.has_qemu_currently ~__context ~self:vm in
-  if
-    System_domains.storage_driver_domain_of_vbd ~__context ~vbd:self = vm
-    && not force_loopback_vbd
+  if domid = 0
+  then (
+    (* SDS vm support *)
+    Xapi_xenops.vbd_plug ~__context ~self
+  )
+  else if
+    (System_domains.storage_driver_domain_of_vbd ~__context ~vbd:self = vm
+    && not force_loopback_vbd)
   then (
     debug "VBD.plug of loopback VBD '%s'" (Ref.string_of self) ;
     Storage_access.attach_and_activate ~__context ~vbd:self ~domid


ocaml/xapi/xapi_vbd.ml
```diff
@@ -77,9 +77,14 @@ let plug ~__context ~self =
   let domid = Int64.to_int (Db.VM.get_domid ~__context ~self:vm) in
   let force_loopback_vbd = Helpers.force_loopback_vbd ~__context in
   let hvm = Helpers.has_qemu_currently ~__context ~self:vm in
-  if
-    System_domains.storage_driver_domain_of_vbd ~__context ~vbd:self = vm
-    && not force_loopback_vbd
+  if domid = 0
+  then (
+    (* SDS vm support *)
```
edwintorok, line 82: You could say 'storage driver domain' support to make it clearer what this is

edwintorok: Are you sure this is still needed though, the line below `storage_driver_domain_of_vbd` should give you the uuid of the driver domain, and then `vm` should be the uuid of dom0 when we plug the vbd into dom0, they should be different and it should take the `else` branch. If that is not happening do we have a bug in `storage_driver_domain_of_vbd`?

edwintorok: Can you add some debug statements before the if to print the ref of the driver domain, and that of the vm?

ocaml/xapi/xapi_vbd.ml
```diff
@@ -77,9 +77,14 @@ let plug ~__context ~self =
...
+    Xapi_xenops.vbd_plug ~__context ~self
```
robhoes, line 83: Can we merge this with the `else` branch below, which also does `Xapi_xenops.vbd_plug` plus some other checks that may be relevant here as well?

ocaml/xapi/xapi_vbd.ml
```diff
@@ -77,9 +77,14 @@ let plug ~__context ~self =
   let domid = Int64.to_int (Db.VM.get_domid ~__context ~self:vm) in
   let force_loopback_vbd = Helpers.force_loopback_vbd ~__context in
   let hvm = Helpers.has_qemu_currently ~__context ~self:vm in
-  if
-    System_domains.storage_driver_domain_of_vbd ~__context ~vbd:self = vm
-    && not force_loopback_vbd
+  if domid = 0
```
robhoes, line 80: Use `Helpers.is_domain_zero`.

robhoes (changes requested):

ocaml/xapi/xapi_vbd.ml
robhoes: I agree with @edwintorok that, if the relevant PBD has the `storage_driver_domain_key` set in `other_config`, then this may not be necessary.

psafont:
Marking as draft since it's a work in progress

edwintorok:
Closing this as obsolete since we don't support storage driver domain as a feature anymore.

PR #4405, opened 2021-05-14
maintenance: use conf-pam instead of depexts

psafont:
This allows to especialize the detection in a single package

This change has been present in xs-opam for some months

### DIFF
--- a/xapi.opam
+++ b/xapi.opam
@@ -16,6 +16,7 @@ depends: [
   "angstrom"
   "base64"
   "cdrom"
+  "conf-pam"
   "ctypes"
   "ctypes-foreign"
   "domain-name"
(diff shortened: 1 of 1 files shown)


lippirk (approved):

lindig (approved):

lindig:
What does `conf-opam` do? It pushes the declaration of dependencies into another opam package such that we don't have to maintain them?

psafont:
> What does `conf-opam` do? It pushes the declaration of dependencies into another opam package such that we don't have to maintain them?

Yes, this is exactly what all conf packages do. They declare the depexts and on top of that they also detect the presence of the c library by trying to compile some code.

For example: https://github.com/ocaml/opam-repository/blob/master/packages/conf-xen/conf-xen.1/files/test.c

PR #4408, opened 2021-05-14
CP-36097 xapi should talk to cluster daemons using hostnames as well as IPs

lippirk:
- replaces usages of `IPv4` with `IP_with_hostname`
- when enabling tls verification, run clusterd prechecks

### DIFF
--- a/ocaml/idl/datamodel_cluster.ml
+++ b/ocaml/idl/datamodel_cluster.ml
@@ -172,9 +172,13 @@ let t =
           ~ty:(Set (Ref _cluster_host)) "cluster_hosts"
           "A list of the cluster_host objects associated with the Cluster"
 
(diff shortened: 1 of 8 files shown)


lindig:
I'm a bit uncomfortable with pairing names with IPs. An address should be either of the two but a pair creates the possibility of a contradiction. An IP can have many names and a name should always resolve to a name. Why do we need such a pair?

psafont (approved):
The "hostnames" are needed for verifying the certificate of the other peer in the TLS connections, they are not used for resolution. These are host UUIDs so they don't change in the lifetime of a cluster, I think it's safe to do.

At the xcp-idl level I would prefer if the addresses contained 'UUIDs' instead of 'hostnames'. The fact they are used as hostnames in TLS seems like and implementation details that is being leaked here. 

That said I would prefer if we started using specific types for addresses instead of plain strings.

lippirk:
Closing because the design has changed

PR #4409, opened 2021-05-18
CP-36092: Enable winbind debug log and wait for winbind daemon during domain join

liulinC:
The debug log is print to stderr
This commit enable the debug with highest level and logged on error

### DIFF
--- a/ocaml/xapi/extauth_plugin_ADwinbind.ml
+++ b/ocaml/xapi/extauth_plugin_ADwinbind.ml
@@ -27,6 +27,8 @@ let net_cmd = !Xapi_globs.net_cmd
 
 let tdb_tool = !Xapi_globs.tdb_tool
 
+let debug_level = !Xapi_globs.winbind_debug_level |> string_of_int
+
 let winbind_service = "winbind"
 
 module Winbind = struct
@@ -52,7 +54,7 @@ let query_domain_workgroup domain =
   try
     let lines =
       Helpers.call_script ~log_output:On_failure net_cmd
-        ["ads"; "lookup"; "-S"; domain]
+        ["ads"; "lookup"; "-S"; domain; "-d"; debug_level]
     in
     match Xapi_cmd_result.of_output_opt ~sep:':' ~key ~lines with
     | Some v ->
@@ -82,6 +84,7 @@ let config_winbind_damon ~domain ~workgroup =
       ; "idmap config * : range = 3000000-3999999"
       ; Printf.sprintf "idmap config %s: backend = rid" domain
       ; Printf.sprintf "idmap config %s: range = 2000000-2999999" domain
+      ; Printf.sprintf "log level = %s" debug_level
       ; "idmap config * : backend = tdb"
       ; "" (* Empty line at the end *)
       ]
@@ -151,7 +154,7 @@ let clean_machine_account ~service_name = function
   | Some u, Some p -> (
       (* Clean machine account in DC *)
       let env = [|Printf.sprintf "PASSWD=%s" p|] in
-      let args = ["ads"; "leave"; "-U"; u] in
+      let args = ["ads"; "leave"; "-U"; u; "-d"; debug_level] in
       try
         Helpers.call_script ~env net_cmd args |> ignore ;
         debug "Cleaning the machine account for domain %s succeeded"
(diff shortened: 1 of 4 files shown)


psafont:
Could we have the level to be configurable instead? I'd like to get a low level by default and be able to turn it on selectively. For example within Citrix's internal testing

stormi:
For the clarity of git history, may I ask that the commit message also states what logs it changes the debug level for? I suppose it's related to winbind, based on the name of the branch, but it's not obvious from the commit message alone.

liulinC:
> Could we have the level to be configurable instead? I'd like to get a low level by default and be able to turn it on selectively. For example within Citrix's internal testing

Yes, we can do that, 
but just clarify,  the debug message is only logged on command error.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -27,6 +27,8 @@ let net_cmd = !Xapi_globs.net_cmd
 
 let tdb_tool = !Xapi_globs.tdb_tool
 
+let debug_level = "10"
```
lindig, line 30: If we make this a binding, it should have type `int`.

robhoes: Constants like this should go into xapi_globs.ml, and made configurable from xapi.conf by adding them to `xapi_globs_spec` near the bottom of that file.

ocaml/xapi/extauth_plugin_ADwinbind.mli
```diff
@@ -0,0 +1,36 @@
...
+ * GNU Lesser General Public License for more details.
+ *)
+
+(**
+ * @winbind group Access Control
+*)
+
+module Winbind : sig
+  (* start winbind service *)
+  val start : timeout:float -> wait_until_success:bool -> unit
```
lindig, line 21: The role of `init` is under-documented. Is it legal to call `start` before calling `init` or not?

lindig (approved):

ocaml/xapi/extauth_plugin_ADwinbind.mli
liulinC: really good question,  I have just implemented the init, please have a check about it.
For your question, yes, the init actually do nothing more than start for winbind.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -25,19 +25,54 @@ open Auth_signature
...
+    Xapi_systemctl.start ~timeout ~wait_until_success name
 
   let stop ~timeout ~wait_until_success =
-    Xapi_systemctl.stop ~timeout ~wait_until_success winbind_service
+    Xapi_systemctl.stop ~timeout ~wait_until_success name
 
   let init_service ~__context =
-    debug "init_service To be implemented in CP-36085"
+    if is_ad_enabled ~__context then
+      start ~wait_until_success:false ~timeout:5.
```
lindig, line 50: You might want to add an `info` log line in the `else` case because you would not call `init_service` is you knew that it would do nothing.

ocaml/xapi/extauth_plugin_ADwinbind.mli
lindig: With the new change, what is the difference between `init` and `start`? For me, `init` looks like a guarded `start`. Could we drop `init` and put all this logic into `start`?

liulinC: During xapi init, it will call the init service of different daemon servcies, 

init_service can perform other stuff like some confiuration update, and start the daemon, like PBIS will check configure.

winbind just happen to only start the deamon ATM, so I would like to just keep it.

lippirk (approved):

quality-gate.sh
```diff
@@ -26,7 +26,7 @@ mli-files () {
   MLIS=$(git ls-files -- '**/*.mli' | xargs -I {} sh -c "echo {} | cut -f 1 -d '.'" \;)
   MLS=$(git  ls-files -- '**/*.ml'  | xargs -I {} sh -c "echo {} | cut -f 1 -d '.'" \;)
   num_mls_without_mlis=$(comm -23 <(sort <<<"$MLS") <(sort <<<"$MLIS") | wc -l)
-  if [ "$num_mls_without_mlis" -eq 389 ]; then
+  if [ "$num_mls_without_mlis" -eq 391 ]; then
```
lindig, line 29: I don't think this number should ever go up - that is the point of this gate.

lindig (changes requested):
Don't introduce ML without MLI files

quality-gate.sh
psafont: I think this is because modules from tests are being counted as well, this PR should ignore them: https://github.com/xapi-project/xen-api/pull/4413

lindig (approved):

psafont:
Please rebase on top of master for the changes in the quality gate :)

PR #4414, opened 2021-05-20
CP-35955: Add absolute guidances in pending_guidances

minglumlu:
There are two kinds of update guidances designed:
1. Recommended guidances: basically these guidances will be applied by
XAPI automatically after host being updated;
2. Absolute guidances: these guidances requires user to apply manually.

Both kinds of guidances look like: RebootHost, RestartToolstack, and
RestartDeviceModel.

This commit introduces function to add absolute guidances into
host.pending_guidances or VM.pending_guidances. In this way, users can
get to know the pending guidances need to be done by them manually.

### DIFF
--- a/ocaml/xapi-cli-server/record_util.ml
+++ b/ocaml/xapi-cli-server/record_util.ml
@@ -185,6 +185,12 @@ let host_operation_to_string = function
   | `apply_updates ->
       "apply_updates"
 
+let update_guidance_to_string = function
+  | `reboot_host ->
+      "reboot_host"
+  | `restart_toolstack ->
+      "restart_toolstack"
+
 let vdi_operation_to_string : API.vdi_operations -> string = function
   | `clone ->
       "clone"
--- a/ocaml/xapi-cli-server/record_util.ml
+++ b/ocaml/xapi-cli-server/record_util.ml
@@ -190,6 +190,8 @@ let update_guidance_to_string = function
       "reboot_host"
   | `restart_toolstack ->
       "restart_toolstack"
+  | `restart_device_model ->
+      "restart_device_model"
 
 let vdi_operation_to_string : API.vdi_operations -> string = function
   | `clone ->
--- a/ocaml/xapi/repository.ml
+++ b/ocaml/xapi/repository.ml
@@ -166,10 +166,10 @@ let http_get_host_updates_in_json ~__context ~host ~installed =
         debug "host %s returned updates: %s" host_name json_str ;
         Yojson.Basic.from_string json_str
       with e ->
-        let ref = Ref.string_of host in
-        error "Failed to get updates from host ref='%s': %s" ref
+        let host' = Ref.string_of host in
+        error "Failed to get updates from host ref='%s': %s" host'
           (ExnHelper.string_of_exn e) ;
-        raise Api_errors.(Server_error (get_host_updates_failed, [ref]))
+        raise Api_errors.(Server_error (get_host_updates_failed, [host']))
       )
     (fun () -> Xapi_session.destroy_db_session ~__context ~self:host_session_id)
 
@@ -180,7 +180,6 @@ let group_host_updates_by_repository ~__context enabled host updates_of_host =
    *   ... ...
    * ]
    *)
-  (* TODO: live patches *)
   match Yojson.Basic.Util.member "updates" updates_of_host with
   | `List updates ->
       List.fold_left
@@ -206,14 +205,14 @@ let group_host_updates_by_repository ~__context enabled host updates_of_host =
           )
         [] updates
   | _ ->
-      let ref = Ref.string_of host in
-      error "Invalid updates from host ref='%s': No 'updates'" ref ;
-      raise Api_errors.(Server_error (get_host_updates_failed, [ref]))
+      let host' = Ref.string_of host in
+      error "Invalid updates from host ref='%s': No 'updates'" host' ;
+      raise Api_errors.(Server_error (get_host_updates_failed, [host']))
   | exception e ->
-      let ref = Ref.string_of host in
-      error "Invalid updates from host ref='%s': %s" ref
+      let host' = Ref.string_of host in
+      error "Invalid updates from host ref='%s': %s" host'
         (ExnHelper.string_of_exn e) ;
-      raise Api_errors.(Server_error (get_host_updates_failed, [ref]))
+      raise Api_errors.(Server_error (get_host_updates_failed, [host']))
 
 let set_available_updates ~__context =
   ignore (get_single_enabled_update_repository ~__context) ;
@@ -349,7 +348,6 @@ let get_host_updates_in_json ~__context ~installed =
           |> List.filter_map
                (get_rpm_update_in_json ~rpm2updates ~installed_pkgs)
         in
-        (* TODO: live patches *)
         `Assoc [("updates", `List updates)]
     )
   with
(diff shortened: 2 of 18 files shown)


ocaml/xapi/repository.ml
```diff
@@ -544,6 +563,28 @@ let apply_immediate_guidances ~__context ~host ~guidances =
...
+         | RestartToolstack ->
+             Db.Host.add_pending_guidances ~__context ~self:host
+               ~value:`RestartToolstack
+         | RestartDeviceModel ->
+             set_pending_restart_device_models ~__context ~host
+         | g ->
+             warn "Unsupported pending guidance %s, ignore it."
+               (Guidance.to_string g))
+  with e ->
+    let ref = Ref.string_of host in
```
lindig, line 583: `ref` is a bad name - it refers to a host; Something like `host'` would be better.

ocaml/xapi/repository_helpers.ml
```diff
@@ -173,12 +173,17 @@ module Guidance = struct
         let msg = error_msg l in
         raise Api_errors.(Server_error (internal_error, [msg]))
 
-  let resort_guidances gs =
+  let resort_guidances ~kind gs =
     match GuidanceStrSet.find_opt "RebootHost" gs with
```
lippirk, line 177: I think there are some unnecessary conversions between strings and `Guidance.t` (see `Guidance.eq` for example). We could use a set of `Guidance.t` instead (and we get some more type safety)

```
module GuidanceSet = Set.Make(struct type t = Guidance.t ;; let compare = Stdlib.compare ) ;;

..

match GuidanceSet.find_opt RebootHost gs with
| Some _ -> GuidanceSet.singleton RebootHost
...
```

ocaml/xapi/repository_helpers.ml
```diff
@@ -173,12 +173,17 @@ module Guidance = struct
...
     | Some _ ->
         GuidanceStrSet.singleton "RebootHost"
-    | None ->
-        gs
+    | None -> (
+      match kind with
```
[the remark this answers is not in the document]
lippirk: or match on both kind and `gs` at once:

```
match (find_opt RebootHost gs, kind) with
```

ocaml/xapi/repository_helpers.ml
```diff
@@ -970,7 +975,7 @@ let eval_guidances ~updates_info ~updates ~kind =
       GuidanceStrSet.union acc
         (eval_guidance_for_one_update ~updates_info ~update:u ~kind))
     GuidanceStrSet.empty updates
-  |> Guidance.resort_guidances
+  |> Guidance.resort_guidances ~kind
   |> GuidanceStrSet.elements
   |> List.map Guidance.of_string
```
lippirk, line 980: At the moment you are converting from a string set, to a string list, and then doing a list diff later on

We could try to use a set the whole way through, and convert to a json string list only right at the end

(again I would recommend this `module GuidanceSet = Set.Make(struct type t = Guidance.t ;; let compare = Stdlib.compare ) ;;`)

ocaml/xapi/repository_helpers.ml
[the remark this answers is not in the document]
minglumlu: Thanks both. Reworked.

ocaml/xapi/repository_helpers.ml
minglumlu: Thanks. A new module `GuidanceSet` is added for this.

ocaml/xapi/repository_helpers.ml
```diff
@@ -173,12 +179,14 @@ module Guidance = struct
...
-        GuidanceStrSet.singleton "RebootHost"
-    | None ->
+  let resort_guidances ~kind gs =
+    match (find_opt RebootHost gs, kind) with
+    | Some _, _ ->
+        singleton RebootHost
+    | None, Recommended ->
         gs
+    | None, Absolute ->
+        filter (fun g -> g != EvacuateHost) gs
```
robhoes, line 189: You should use `<>` (`!=` would have probably worked this time, but only by chance as the type is a very simple variant type).

ocaml/xapi/repository.ml
```diff
@@ -457,41 +455,62 @@ let apply ~__context ~host =
...
+  |> List.filter_map f
   |> function
   | [] ->
       ()
   | _ :: _ ->
-      let msg = "Can't restart device models for some VMs" in
+      let msg =
+        Printf.sprintf "Can't %s device models for some VMs" action_label
+      in
       raise Api_errors.(Server_error (internal_error, [msg]))
```
robhoes, line 476: It looks like this would occur (only) for paused VMs. It's not an internal error then, as anyone can pause VMs. It should be an explicit API error that the client can handle.

ocaml/xapi/repository.ml
```diff
@@ -510,39 +529,61 @@ let apply_immediate_guidances ~__context ~host ~guidances =
...
   with e ->
-    let ref = Ref.string_of host in
-    error "applying immediate guidances on host ref='%s' failed: %s" ref
+    let host' = Ref.string_of host in
+    error "applying immediate guidances on host ref='%s' failed: %s" host'
       (ExnHelper.string_of_exn e) ;
-    raise Api_errors.(Server_error (apply_guidance_failed, [ref]))
+    raise Api_errors.(Server_error (apply_guidance_failed, [host']))
+
+let apply_pending_guidances ~__context ~host ~guidances =
```
robhoes, line 566: The word "apply" is a bit vague, but the function `apply_immediate_guidances` above actually executes the actions suggested by the guidances, whereas this function just records them in the DB. Perhaps call it `set_pending_guidances` to avoid confusion.

ocaml/xapi/repository.ml
```diff
@@ -510,39 +529,61 @@ let apply_immediate_guidances ~__context ~host ~guidances =
...
+         | RestartDeviceModel ->
+             set_pending_restart_device_models ~__context ~host
+         | g ->
+             warn "Unsupported pending guidance %s, ignore it."
+               (Guidance.to_string g))
+  with e ->
+    let host' = Ref.string_of host in
+    error "applying pending guidances on host ref='%s' failed: %s" host'
+      (ExnHelper.string_of_exn e) ;
+    raise Api_errors.(Server_error (apply_guidance_failed, [host']))
```
robhoes, line 586: I don't think this is the right error either. In this case, the functions called inside of this function shouldn't raise any exceptions because it is just DB sets and nothing is actually "applied".

robhoes (changes requested):

ocaml/xapi/xapi.ml
```diff
@@ -1290,6 +1290,13 @@ let server_init () =
                 Helpers.call_api_functions ~__context (fun rpc session_id ->
                     Xapi_pool_update.resync_host __context
                       (Helpers.get_localhost ~__context)) )
+          ; ( "Cleanup pending RestartToolstack guidance"
+            , [Startup.NoExnRaising]
+            , fun () ->
+                Helpers.call_api_functions ~__context (fun rpc session_id ->
+                    Xapi_host.remove_pending_guidance __context
```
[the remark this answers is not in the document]
minglumlu: I thought a slave should update the DB by an API call.  Now it looks like the pool DB can be updated by DB operation issued from slave directly.

ocaml/xapi-cli-server/record_util.ml
```diff
@@ -184,6 +184,14 @@ let host_operation_to_string = function
   | `apply_updates ->
       "apply_updates"
 
+let update_guidance_to_string = function
```
lindig, line 187: It seems that we have both regular and polymorphic variants to describe guidance. Is this required?

ocaml/xapi/xapi.ml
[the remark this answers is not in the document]
robhoes: Any host can read and write the DB directly through the DB module. On the master, this acts on the local in-memory DB. On other hosts, the DB calls are automatically forwarded to the pool master on the (persistent) DB connection (see db.ml and db_cache.ml).

ocaml/xapi-cli-server/record_util.ml
robhoes: The polymorphic one is generated from the datamodel enum. There is no way around that one.

robhoes (approved):
I'm generally happy with the code now, but there are some conflicts that need to be resolved (perhaps formatting changes).

minglumlu:
Rebased from master to resolve conflicts caused by application of new formatting rules.

minglumlu:
Hi @lippirk @lindig 
May I have your review again on this? Thanks :-)

lippirk (approved):

PR #4415, opened 2021-05-20
CA-354689 don't fail if host cert to be removed doesn't exist

lindig:
Without certificate checking, there is no certificate to be removed in a
pool when a member is ejected. Previously we failed to recognise the
case where certificate checking is not enabled.

The fix tries to keep it simple rather than introducing complex code
paths based on the certificate checking beeing enabled. During pool
eject we remove certficates of the ejected host from the databse and the
file system. If certificate checking is not enabled, there is nothing to
be removed from the database and the file system. We fix here the case
for the file system by no longer failing if the file does not exist in
the first place.

The alternative design would branch on the status of certficate checking
and I believe this introduces more complexity without adding any
robustness.


psafont (approved):
I agree with the solution. The are many ways the filesystem is populated with the other hosts' certificates, depending on when they joined the pool and when was the pool upgraded. It's better to deal with the discrepancies at the level when they are being deleted.

lindig:
It looks like my `make format` does not agree with the CI anymore after that was upgraded - this is causing the current failure. But it means that we have to reformat all code and this is not related just to my code change.

psafont (approved):

robhoes (approved):

PR #4417, opened 2021-05-20
maintenance: run make format

psafont:
This is needed because formatting change for ocamlformat 0.18
Previously the setting we used sometimes added a newline before closing parentheses, now it's always added, creating a whole lot of lines.


psafont:
Closing in favour of #4415

PR #4418, opened 2021-05-21
Winbind plugin implementations

lippirk:
This PR implements ocaml wrappers around some wbinfo, ntlm_auth and net. I've tested the changes using this: https://github.com/xapi-project/xen-api/commit/ff36708793e27844d5495031f49b03c53b1afadd (and also `xe subject-add/remove`)

### DIFF
--- a/ocaml/tests/test_extauth_plugin_ADwinbind.ml
+++ b/ocaml/tests/test_extauth_plugin_ADwinbind.ml
@@ -41,5 +41,317 @@ module ExtractOuConfig = Generic.MakeStateless (struct
       ]
 end)
 
+let test_domainify_uname =
+  let open Extauth_plugin_ADwinbind in
+  let check uname exp () =
+    let msg = Printf.sprintf "%s -> %s" uname exp in
+    let ac = domainify_uname ~domain:"domain.net" uname in
+    Alcotest.(check string) msg exp ac
+  in
+  let matrix =
+    [
+      ("KRBTGT", "KRBTGT")
+    ; ({|user|}, {|user@domain.net|})
+    ; ({|user@domain.net|}, {|user@domain.net|})
+    ; ({|DOMAIN\user|}, {|DOMAIN\user|})
+      (* if username already contains a domain, DO NOT try and correct it *)
+    ; ({|user@unknowndomain.net|}, {|user@unknowndomain.net|})
+    ; ({|UNKNOWNDOMAIN\user|}, {|UNKNOWNDOMAIN\user|})
+    ]
+  in
+  matrix
+  |> List.map @@ fun (inp, exp) ->
+     (Printf.sprintf "%s -> %s" inp exp, `Quick, check inp exp)
+
+let test_parse_wbinfo_uid_info =
+  let open Extauth_plugin_ADwinbind.Wbinfo in
+  let string_of_result x =
+    match x with
+    | Ok x ->
+        Printf.sprintf "Ok %s" (string_of_uid_info x)
+    | Error () ->
+        "Error ()"
+  in
+  let uid_info = Test_util.alcotestable_of_pp string_of_result in
+  let check stdout exp () =
+    let msg = Printf.sprintf "<omit> -> %s" (string_of_result exp) in
+    let ac = parse_uid_info stdout in
+    Alcotest.(check uid_info) msg exp ac
+  in
+
+  let matrix =
+    [
+      ( {|CONNAPP\ladmin:*:3000003:3000174:ladmin:/home/CONNAPP/ladmin:/bin/bash|}
+      , Ok
+          {
+            user_name= {|CONNAPP\ladmin|}
+          ; uid= 3000003
+          ; gid= 3000174
+          ; gecos= {|ladmin|}
+          } )
+    ; ( {|CONNAPP\locked:*:3000004:3000174::/home/CONNAPP/locked:/bin/bash|}
+      , Ok
+          {user_name= {|CONNAPP\locked|}; uid= 3000004; gid= 3000174; gecos= ""}
+      )
+    ; ({|blah|}, Error ())
+    ]
+  in
+  matrix |> List.map @@ fun (inp, exp) -> (inp, `Quick, check inp exp)
+
+let test_parse_ldap_stdout =
+  let open Extauth_plugin_ADwinbind.Ldap in
+  let string_of_result = function
+    | Ok x ->
+        Printf.sprintf "Ok %s" (string_of_user x)
+    | Error x ->
+        Printf.sprintf "Error %s" x
+  in
+  let ldap_user = Test_util.alcotestable_of_pp string_of_result in
+  let check stdout exp () =
+    let msg = Printf.sprintf "<omit> -> %s" (string_of_result exp) in
+    let ac = parse_user stdout in
+    Alcotest.(check ldap_user) msg exp ac
+  in
+
+  let stdout_ladmin =
+    {|Got 1 replies
+
+objectClass: top
+objectClass: person
+objectClass: organizationalPerson
+objectClass: user
+cn: ladmin
+givenName: ladmin
+distinguishedName: CN=ladmin,CN=Users,DC=conappada,DC=local
+instanceType: 4
+whenCreated: 20210203072119.0Z
+whenChanged: 20210518070906.0Z
+displayName: ladmin
+uSNCreated: 70509
+memberOf: CN=Domain Admins,CN=Users,DC=conappada,DC=local
+memberOf: CN=Enterprise Admins,CN=Users,DC=conappada,DC=local
+memberOf: CN=Administrators,CN=Builtin,DC=conappada,DC=local
+uSNChanged: 362555
+name: ladmin
+objectGUID: 1a29218b-f145-45ce-a181-e471864c934b
+userAccountControl: 66048
+badPwdCount: 0
+codePage: 0
+countryCode: 0
+badPasswordTime: 132657962285300918
+lastLogoff: 0
+lastLogon: 132663205168179135
+pwdLastSet: 132573331165736472
+primaryGroupID: 513
+objectSid: S-1-5-21-3143668282-2591278241-912959342-1178
+adminCount: 1
+accountExpires: 9223372036854775807
+logonCount: 1166
+sAMAccountName: ladmin
+sAMAccountType: 805306368
+userPrincipalName: ladmin@conappada.local
+objectCategory: CN=Person,CN=Schema,CN=Configuration,DC=conappada,DC=local
+dSCorePropagationData: 20210203085144.0Z
+dSCorePropagationData: 16010101000000.0Z
+lastLogonTimestamp: 132657953463490742
+msDS-SupportedEncryptionTypes: 24
+  |}
+  in
+  let stdout_krbtgt =
+    {|
+
+
+    Got 1 replies
+
+objectClass: top
+objectClass: person
+objectClass: organizationalPerson
+objectClass: user
+cn: krbtgt
+description: Key Distribution Center Service Account
+distinguishedName: CN=krbtgt,CN=Users,DC=conappada,DC=local
+instanceType: 4
+whenCreated: 20201029064320.0Z
+whenChanged: 20201029065834.0Z
+uSNCreated: 12324
+memberOf: CN=Denied RODC Password Replication Group,CN=Users,DC=conappada,DC=local
+uSNChanged: 12767
+showInAdvancedViewOnly: TRUE
+name: krbtgt
+objectGUID: d898ace2-862f-456d-b5da-e012f412a8a2
+userAccountControl: 514
+badPwdCount: 0
+codePage: 0
+countryCode: 0
+badPasswordTime: 0
+lastLogoff: 0
+lastLogon: 0
+pwdLastSet: 132484274001314113
+primaryGroupID: 513
+objectSid: S-1-5-21-3143668282-2591278241-912959342-502
+adminCount: 1
+accountExpires: 9223372036854775807
+logonCount: 0
+sAMAccountName: krbtgt
+sAMAccountType: 805306368
+servicePrincipalName: kadmin/changepw
+objectCategory: CN=Person,CN=Schema,CN=Configuration,DC=conappada,DC=local
+isCriticalSystemObject: TRUE
+dSCorePropagationData: 20201029065834.0Z
+dSCorePropagationData: 20201029064320.0Z
+dSCorePropagationData: 16010101000416.0Z
+msDS-SupportedEncryptionTypes: 0
+    |}
+  in
+  let stdout_locked =
+    {| Got 1 replies
+
+objectClass: top
+objectClass: person
+objectClass: organizationalPerson
+objectClass: user
+cn: locked
+givenName: locked
+distinguishedName: CN=locked,CN=Users,DC=conappada,DC=local
+instanceType: 4
+whenCreated: 20210303032518.0Z
+whenChanged: 20210303041203.0Z
+displayName: locked
+uSNCreated: 101671
+uSNChanged: 101737
+name: locked
+objectGUID: 26f8c3e3-e4cd-4bd1-b0fb-4e4f84c65a49
+userAccountControl: 532
+badPwdCount: 0
+codePage: 0
+countryCode: 0
+badPasswordTime: 132592179051171201
+lastLogoff: 0
+lastLogon: 132592972097050310
+pwdLastSet: 132592177189171588
+primaryGroupID: 513
+objectSid: S-1-5-21-3143668282-2591278241-912959342-1202
+accountExpires: 9223372036854775807
+logonCount: 3
+sAMAccountName: locked
+sAMAccountType: 805306368
+userPrincipalName: locked@conappada.local
+lockoutTime: 0
+objectCategory: CN=Person,CN=Schema,CN=Configuration,DC=conappada,DC=local
+dSCorePropagationData: 16010101000000.0Z
+lastLogonTimestamp: 132592177310358648
+
+
+
+    |}
+  in
+  let stdout_expired =
+    {|Got 1 replies
+
+objectClass: top
+objectClass: person
+objectClass: organizationalPerson
+objectClass: user
+cn: experied
+givenName: experied
+distinguishedName: CN=experied,CN=Users,DC=conappada,DC=local
+instanceType: 4
+whenCreated: 20210302074856.0Z
+whenChanged: 20210302081509.0Z
+displayName: experied
+uSNCreated: 101115
+uSNChanged: 101243
+name: experied
+objectGUID: caaa3e6a-fe9e-4c0c-9704-931f7445fb3f
+userAccountControl: 512
+badPwdCount: 16
+codePage: 0
+countryCode: 0
+badPasswordTime: 132591456191947359
+lastLogoff: 0
+lastLogon: 0
+pwdLastSet: 132591449361503688
+primaryGroupID: 513
+objectSid: S-1-5-21-3143668282-2591278241-912959342-1200
+accountExpires: 132665213390000000
+logonCount: 0
+sAMAccountName: experied
+sAMAccountType: 805306368
+userPrincipalName: experied@conappada.local
+lockoutTime: 0
+objectCategory: CN=Person,CN=Schema,CN=Configuration,DC=conappada,DC=local
+dSCorePropagationData: 16010101000000.0Z
+msDS-SupportedEncryptionTypes: 0|}
+  in
+  let matrix =
+    [
+      ( stdout_ladmin
+      , Ok
+          {
+            upn= "ladmin@conappada.local"
+          ; account_disabled= false
+          ; account_expired= false
+          ; account_locked= false
+          ; password_expired= false
+          } )
+    ; ( stdout_locked
+      , Ok
+          {
+            upn= "locked@conappada.local"
+          ; account_disabled= false
+          ; account_expired= false
+          ; account_locked= true
+          ; password_expired= false
+          } )
+    ; ( stdout_expired
+      , Ok
+          {
+            upn= "experied@conappada.local"
+          ; account_disabled= false
+          ; account_expired= true
+          ; account_locked= false
+          ; password_expired= false
+          } )
+    ; (stdout_krbtgt, Error "ldap missing key 'userPrincipalName'")
+    ; ("Got 0 replies", Error "ldap parsing failed ': got 0 replies'")
+    ; ( "complete garbage"
+      , Error "ldap parsing failed 'unexpected header: string'" )
+    ]
+  in
+  matrix |> List.map @@ fun (inp, exp) -> ("<omit inp>", `Quick, check inp exp)
+
+let test_wbinfo_exception_of_stderr =
+  let open Extauth_plugin_ADwinbind.Wbinfo in
+  let open Auth_signature in
+  let string_of_result x =
+    match x with
+    | Some ex ->
+        Printf.sprintf "Some %s" (Printexc.to_string ex)
+    | None ->
+        "None"
+  in
+  let ex = Test_util.alcotestable_of_pp string_of_result in
+  let check stderr exp () =
+    let msg = Printf.sprintf "<omit> -> %s" (string_of_result exp) in
+    let ac = exception_of_stderr stderr in
+    Alcotest.(check ex) msg exp ac
+  in
+  let matrix =
+    [
+      ( "failed to call wbcLookupName: WBC_ERR_DOMAIN_NOT_FOUND\x0ACould not \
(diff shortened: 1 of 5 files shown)


ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -31,6 +33,12 @@ let tdb_tool = !Xapi_globs.tdb_tool
 
 let debug_level = !Xapi_globs.winbind_debug_level |> string_of_int
 
+let call_wbinfo (args : string list) : (string, exn) result =
+  try
+    let stdout = Helpers.call_script ~log_output:Never wb_cmd args in
```
liulinC, line 38: if it is a generic function (other than only check krbtgt in retry_until_timeout), it should not set `log_outgput:Never` as we want the debug messages on exception.

liulinC:
The Sanity check looks good, 
but I think dev test to call such APIs (like from python SDK) is necessary as part of the dev test.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -221,7 +468,32 @@ let clean_local_resources () : unit =
...
+  (* subject_id get_subject_identifier(string subject_name)
+
+      Takes a subject_name (as may be entered into the XenCenter UI when defining subjects --
+      see Access Control wiki page); and resolves it to a subject_id against the external
+      auth/directory service.
+      Raises Not_found (*Subject_cannot_be_resolved*) if authentication is not succesful.
+  *)
+  let get_subject_identifier subject_name =
+    let subject_name =
+      domainify_uname ~domain:(get_service_name ()) subject_name
```
lippirk, line 493: I believe it's a good idea to domainify the username here. I found instances where

```
wbinfo -n user@domain.net
```
works,

but
 
```
wbinfo -n user
```

does not work

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -233,8 +505,9 @@ module AuthADWinbind : Auth_signature.AUTH_MODULE = struct
       Raises auth_failure if authentication is not successful
   *)
 
-  let authenticate_username_password username password =
-    "authenticate_ticket To be implemented in CP-35399"
+  let authenticate_username_password uname password =
+    let () = maybe_raise (ntlm_auth uname password) in
```
lippirk, line 509: @liulinC the behaviour of ntlm_auth is really strange. I found that sometimes it will not accept the username in anything other than SAM format (that is, `echo "$PASSWD" | netlm_auth --username user` doesn't work, but `echo "$PASSWD" | ntlm_auth --username "DOMAIN\user"` does work)

I think we might need to convert `uname` to SAM format just in case

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -233,8 +297,24 @@ module AuthADWinbind : Auth_signature.AUTH_MODULE = struct
       Raises auth_failure if authentication is not successful
   *)
 
-  let authenticate_username_password username password =
-    "authenticate_ticket To be implemented in CP-35399"
+  let authenticate_username_password uname password =
+    (* example:
+     *
+     * $ wbinfo -a user@domain.net%PASSWD
```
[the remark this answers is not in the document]
lippirk: I'm now passing the password in on stdin, which won't appear process trees, etc.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -31,6 +45,245 @@ let tdb_tool = !Xapi_globs.tdb_tool
...
+    Rpcmarshal.marshal user.Rpc.Types.ty x |> Jsonrpc.to_string
+
+  let parse_user stdout : (user, string) result =
+    let module Map = Map.Make (String) in
+    let module P = struct
+      open Angstrom
+
+      let is_space = function ' ' -> true | _ -> false
+
+      let s = take_while is_space
```
lindig, line 94: While I think `ws` is acceptable as a name in a scanner, this is a step too far and would prefer a longer name. I understand the problem of having similar names for predicates and parsers which you solved by using `is_` as prefix. I think this should be `spaces`. I am a bit surprised we use both spaces and white space - this could be confusing for the user.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -31,6 +45,245 @@ let tdb_tool = !Xapi_globs.tdb_tool
...
+          choice ~failure_msg:"unexpected header"
+            [string "Got" *> char ' ' *> ns <* char ' ' <* string "replies"]
+        in
+        match num_replies with
+        | "1" ->
+            return ()
+        | _ ->
+            Printf.sprintf "got %s replies" num_replies |> fail
+
+      let p =
```
lindig, line 126: This needs a better name or a comment.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -31,6 +45,245 @@ let tdb_tool = !Xapi_globs.tdb_tool
...
+    let* upn = get_string "userPrincipalName" in
+    let* user_account_control = get_int Int32.of_string "userAccountControl" in
+    let* account_expires = get_int Int64.of_string "accountExpires" in
+    let account_expired =
+      (* see https://docs.microsoft.com/en-us/windows/win32/adschema/a-accountexpires *)
+      let windows_nt_time_to_unix_time x =
+        Int64.sub (Int64.div x 10000000L) 11644473600L
+      in
+      match account_expires with
+      | 0L | 9223372036854775807L ->
```
lindig, line 157: What is this magic number?

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -31,6 +45,245 @@ let tdb_tool = !Xapi_globs.tdb_tool
...
+    (* example:
+     * $ wbinfo -s S-1-5-21-3143668282-2591278241-912959342-502
+       CONNAPP\krbtgt 1 *)
+    let args = ["-s"; sid] in
+    let* stdout =
+      call_wbinfo args <!> fun _ ->
+      generic_ex "'wbinfo %s' failed" (String.concat " " args)
+    in
+    (* we need to drop everything after the last space *)
+    match List.rev (String.split_on_char ' ' stdout) with
```
lindig, line 218: It might be easier to use regular expressions for problems like this.

lindig (commented):
I think Angstrom is the right approach for parsing structured output. But I see now that we are using a mixed approach of using string functions and Angstrom. Regular expressions would be also quite powerful for anything with not too much structure.

ocaml/xapi/extauth_plugin_ADwinbind.ml
lippirk: this number in particular is actually `Int64.max_int`, so I'll replace that

(it's mentioned in that link above)

ocaml/xapi/extauth_plugin_ADwinbind.ml
lippirk: You're right, this is much cleaner

ocaml/xapi/extauth_plugin_ADwinbind.ml
lippirk: I've doing some renaming :+1: 

`s` was used for separating key values pairs like `key: value`, but I was probably being too generous accepting multiple spaces, so I've changed it to just `char ' '` instead

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -31,6 +45,255 @@ let tdb_tool = !Xapi_globs.tdb_tool
...
+        Error
+          (generic_ex "unable to find SID in output of 'wbinfo %s'"
+             (String.concat " " args))
+
+  let name_of_sid =
+    (* example:
+     * $ wbinfo -s S-1-5-21-3143668282-2591278241-912959342-502
+       CONNAPP\krbtgt 1 *)
+    (* we need to drop everything after the last space *)
+    let regex = Re.Perl.(compile (re {|^([^\s].*)\ \d+$|})) in
```
liulinC, line 216: per my test, you need to match the ending new line at the end,  something like `let regex = Re.Perl.(compile (re {|^([^\s].*)\ \d+\s$|}))`   `\s` is used to match the ending new line, , tried `\n` for newline, but not work.

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: Yes, you are right, this is an existing issue with PBIS, and we need to fix it. 
please update it in this PR if you like.

ocaml/xapi/extauth_plugin_ADwinbind.ml
lippirk: I couldn't get your regex to work - but both the following work just fine for me:

```bash
# don't allow trailing whitespace
echo "$(wbinfo -s S-1-5-21-1076320343-3137401123-2908695819-456812)" | sed -E -e 's/([^s].*)\ [0-9]+/\1/g'

# do allow trailing whitespace
echo "$(wbinfo -s S-1-5-21-1076320343-3137401123-2908695819-456812)" | sed -E -e 's/([^s].*)\ [0-9]+\s*/\1/g'
```

lippirk: (i've updated the code to allow trailing whitespace)

ocaml/tests/test_extauth_plugin_ADwinbind.ml
```diff
@@ -41,5 +41,290 @@ module ExtractOuConfig = Generic.MakeStateless (struct
...
+        Printf.sprintf "Error %s" x
+  in
+  let ldap_user = Test_util.alcotestable_of_pp string_of_result in
+  let check stdout exp () =
+    let msg = Printf.sprintf "<omit> -> %s" (string_of_result exp) in
+    let ac = parse_user stdout in
+    Alcotest.(check ldap_user) msg exp ac
+  in
+
+  let stdout_ladmin =
```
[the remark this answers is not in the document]
lippirk: The type I'm parsing to is this:

```
  type user = {
            upn: string
          ; account_disabled: bool
          ; account_expired: bool
          ; account_locked: bool
          ; password_expired: bool }
```

Internally I indeed parse to a (string, string) map, and then do some calculations to get those booleans

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -265,8 +588,26 @@ module AuthADWinbind : Auth_signature.AUTH_MODULE = struct
...
-  let query_subject_information subject_identifier =
-    failwith "extauth_plugin authenticate_ticket not implemented"
+  let query_subject_information (sid : string) =
+    (* we have to aggregate information from wbinfo and / or an ldap query (using net binary)
+     * to begin with, we don't even know if the sid belongs to a group or a user
+     * so we first do a test with [gid_of_sid] + [uid_of_sid] *)
+    let res =
+      let* name = Wbinfo.name_of_sid sid in
+      match Wbinfo.uid_of_sid sid with
+      | Ok uid ->
```
liulinC, line 598: why need to use `uid` as identifier for the user here, other than `SID` directly?
`uid` may not right (consistent) and may not trusted as 
- hcp_nss will override the uid to `0`, to force all users login as root, we only support one single root in dom0
- pbis and users have different uid for the same domain user, for upgrade case, we cannot keep it consistent.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -245,15 +541,42 @@ module AuthADWinbind : Auth_signature.AUTH_MODULE = struct
...
-  *)
-  let get_subject_identifier _subject_name =
-    "get_subject_identifier To be implemented in CP-36087"
+  let query_subject_information_group (name : string) (gid : int) (sid : string)
+      =
+    [
+      ("subject-name", name)
+    ; ("subject-gid", string_of_int gid)
+    ; ("subject-sid", sid)
+    ; ("subject-is-group", string_of_bool true)
```
liulinC, line 550: why not just `; ("subject-is-group", "true")`

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -31,6 +45,255 @@ let tdb_tool = !Xapi_globs.tdb_tool
...
+    let args = ["-S"; sid] in
+    let* stdout =
+      call_wbinfo args <!> fun _ ->
+      generic_ex "'wbinfo %s' failed" (String.concat " " args)
+    in
+    try Ok (String.trim stdout |> int_of_string)
+    with _ ->
+      Error (generic_ex "parsing 'wbinfo %s' failed" (String.concat " " args))
+
+  type uid_info = {user_name: string; uid: int; gid: int; gecos: string}
```
liulinC, line 268: It is really good to extract information into struct (types) like `user` and `uid_info` 👍 , 
Just a question, should the type be defined inside the modules, like `uid_info`, 
as `uid_info` does not looks only specific to the `wbinfo` command.

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: I have re-consider this, currently implementation probably good enough as
- hcp_nss only override uid to 0 during resolve uid like `getpwnam_r`, here we ask wbinfo `winbind` for the UID directly, so it is not overrided.
- The uid here is only used temporary, for the upgraded case, the UID is also regenerated by winbind and used to query winbind info, so it should be fine.

So, currently implementation looks good to me,  Just for your info, if you do want to distinguish a user/group from SID, you can use  `wbinfo -s, --sid-to-name=SID,`, and it will tell you it is a group or user, 1 for user, 2 for group, like

```[root@xrtuk-14-05 test_nss]# wbinfo -s S-1-5-21-3143668282-2591278241-912959342-1174
CONNAPP\test_group 2
[root@xrtuk-14-05 test_nss]# wbinfo -s S-1-5-21-3143668282-2591278241-912959342-1178
CONNAPP\ladmin 1```

File: ocaml/xapi/extauth_plugin_ADwinbind.ml [hunk 14]
```diff
@@ -31,6 +45,255 @@ let tdb_tool = !Xapi_globs.tdb_tool
...
+    let _stdout =
+      Helpers.call_script ~log_output:Never ~stdin:passwd
+        !Xapi_globs.ntlm_auth_cmd args
+    in
+    Ok ()
+  with e -> Error e
+
+let call_ldap (query : string) : (string, exn) result =
+  try
+    let args = ["ads"; "search"; "--machine-pass"; query] in
```
liulinC, line 60: add `"-d";  debug_level` to the command line

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -31,6 +45,255 @@ let tdb_tool = !Xapi_globs.tdb_tool
...
+
+      let is_whitespace = function
+        | ' ' | '\n' | '\t' | '\r' ->
+            true
+        | _ ->
+            false
+
+      let ws = skip_while is_whitespace
+
+      let header =
```
liulinC, line 107: Maybe that is because I am not familiar with `Angstrom`,  The command result parse looks quite complicated to me.

why not just reuse the `xapi_cmd_result` and have `:` as the `sep`, add lines to the resultset, 
if the resultset is empy, then no replies, otherwise, just look up the key in the resultset?

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -233,8 +515,22 @@ module AuthADWinbind : Auth_signature.AUTH_MODULE = struct
...
-    "authenticate_ticket To be implemented in CP-35399"
+  let authenticate_username_password uname password =
+    (* the ntlm_auth binary expects the username to be in either SAM or UPN format.
+     * we use wbinfo to try to convert the provided [uname] into said format.
+     * as a last ditch attempt, we try to auth with the provided [uname]
+     *
+     * see CA-346287 for more information *)
+    let orig_uname = uname in
+    let sid = get_subject_identifier uname in
+    (let* wbinfo_ified_uname = Wbinfo.name_of_sid sid in
```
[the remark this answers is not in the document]
lippirk: by 'wbinfoified uname', i mean that the uname been transformed by wbinfo

you can add -ify to some english words to make them verbs (https://www.collinsdictionary.com/dictionary/english/ify). adding -ified makes them past tense, or turns them into adjectives (wbinifoified is an adjective here)

a more standard example: simple -> simplify, simplified

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -245,15 +541,42 @@ module AuthADWinbind : Auth_signature.AUTH_MODULE = struct
...
-  (* subject_id get_subject_identifier(string subject_name)
-
-      Takes a subject_name (as may be entered into the XenCenter UI when defining subjects --
-      see Access Control wiki page); and resolves it to a subject_id against the external
-      auth/directory service.
-      Raises Not_found (*Subject_cannot_be_resolved*) if authentication is not succesful.
-  *)
-  let get_subject_identifier _subject_name =
-    "get_subject_identifier To be implemented in CP-36087"
+  let query_subject_information_group (name : string) (gid : int) (sid : string)
```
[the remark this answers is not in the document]
lippirk: i define them outside just because i find it slightly more readable, but it doesn't really matter

ocaml/xapi/extauth_plugin_ADwinbind.ml
lippirk: they are functionally equivalent

ocaml/xapi/extauth_plugin_ADwinbind.ml
lippirk: We need to get the uid, since it's one of the fields returned when you call `query_subject_information`

> pbis and users have different uid for the same domain user, for upgrade case, we cannot keep it consistent.

true, after the upgrade we'll need to make sure that any subject rows in the database are updated accordingly :+1: 

> and it will tell you it is a group or user, 1 for user, 2 for group

thanks, I will add that simplification!

ocaml/xapi/extauth_plugin_ADwinbind.ml
lippirk: It is specific to wbinfo in the sense that we obtain the information from wbinfo (so I believe the `Wbinfo` module is in the right place for it)

ocaml/xapi/extauth_plugin_ADwinbind.ml
lippirk: My issue with xapi_cmd_result is that is lenient if it can't parse a line - it simply ignores the error and moves on. Additionally I specifically want to make sure that all of the key value pairs are in one 'block'. Angstrom is also convenient for parsing the header

I agree it looks quite complicated, which is a shame

liulinC: we can ignore the headers,  and just parse the values, 
If you do care about the record numbers, perhaps just check whether the resultset is empty. 
If you does not found the target key/value pair, then you can report error.

Just a little bit worried needs to get familiar with `Angstrom` to maintain the codes.  you can make the decision.

liulinC:
also, The stderr is printed to console on error, 
```
[root@xrtuk-14-05 ssh]# xe vm-list -s localhost -u ladmin@conappada.local -pw 1234567
The server failed to handle your request, due to an internal error. The given message may give details useful for debugging the problem.
message: session.login_with_password
<extra>: INTERNAL_ERROR: [ Subprocess exited with unexpected code 1; stdout = [ Password: 
NT_STATUS_NO_SUCH_USER: The specified account does not exist. (0xc0000064)
 ]; stderr = [  ] ]
```

liulinC:
Looks like there is nothing necessary on xapi exit, so please clear the debug message :
https://github.com/lippirk/xen-api/blob/private/bena/winbind/ocaml/xapi/extauth_plugin_ADwinbind.ml#L754

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -31,6 +54,268 @@ let tdb_tool = !Xapi_globs.tdb_tool
...
+  try
+    let args = ["--username"; uname] in
+    let _stdout =
+      Helpers.call_script ~log_output:Never ~stdin:passwd
+        !Xapi_globs.ntlm_auth_cmd args
+    in
+    Ok ()
+  with e ->
+    D.info "ntlm auth failed with uname='%s'" uname ;
+    Error e
```
liulinC, line 67: when the auth failed, the higher layer expected a `Auth_signature.Auth_failure`
 https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_session.ml#L824
cli will match the error code, and provide  messages back to user ( `Authentication failed` in this case), otherwise, just return Server Internal error,  like `The server failed to handle your request, due to an internal error. The given message may give details useful for debugging the problem.                      
message: session.login_with_password
<extra>: INTERNAL_ERROR: [ Subprocess exite`

Hope the exceptions to be raised can be checked for all the reminding APIs,  in higher layer codes or refer to PBIS,

liulinC (commented):
I have just finish the second round of review, 
the code generally looks good,

Please check the exceptions to be raised in the APIs.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -31,6 +54,268 @@ let tdb_tool = !Xapi_globs.tdb_tool
...
+      {
+        upn
+      ; account_expired
+      ; account_disabled= logand user_account_control disabled_bit <> 0l
+      ; account_locked= logand user_account_control lockout_bit <> 0l
+      ; password_expired= logand user_account_control passw_expire_bit <> 0l
+      }
+
+  let query_user sid =
+    let query = Printf.sprintf "(&(objectClass=user)(objectSid=%s))" sid in
```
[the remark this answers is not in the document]
lippirk: thanks, I've simplified this

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -31,6 +54,268 @@ let tdb_tool = !Xapi_globs.tdb_tool
...
+
+  let uid_of_sid sid =
+    let args = ["--sid-to-uid"; sid] in
+    let* stdout =
+      call_wbinfo args <!> fun _ ->
+      generic_ex "'wbinfo %s' failed" (String.concat " " args)
+    in
+    try Ok (String.trim stdout |> int_of_string)
+    with _ ->
+      Error (generic_ex "parsing 'wbinfo %s' failed" (String.concat " " args))
```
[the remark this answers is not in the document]
lippirk: I avoided logging stdout here because `Helpers.call_script ~log_output:On_failure` will have logged it already

ocaml/xapi/extauth_plugin_ADwinbind.ml
lippirk: I added a function to extract error messages from the wbinfo stderr, and turn them into the expected exception (I believe these match up roughly with what we have for pbis)

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -196,6 +518,18 @@ let clean_local_resources () : unit =
     error "%s : %s" msg (ExnHelper.string_of_exn e) ;
     raise (Auth_service_error (E_GENERIC, msg))
 
+let domainify_uname ~domain uname =
+  let open Astring.String in
```
lindig, line 522: This looks to me like something that could be more reliably checked with a regular expression. However, I'm fine with it as is.

lindig (approved):

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -31,6 +60,299 @@ let tdb_tool = !Xapi_globs.tdb_tool
...
+    let regex = Re.Perl.(compile (re {|.*(WBC_ERR_[A-Z_]*).*|})) in
+    let get_regex_match x =
+      Option.bind (Re.exec_opt regex x) (fun g ->
+          match Re.Group.all g with [|_; code|] -> Some code | _ -> None)
+    in
+    fun stderr ->
+      get_regex_match stderr
+      |> Option.map (fun code ->
+             match code with
+             | "WBC_ERR_AUTH_ERROR" ->
```
liulinC, line 219: I suppose this error list comes from https://github.com/samba-team/samba/blob/master/nsswitch/libwbclient/wbclient.h#L36

a comment to link it will be good.

liulinC (commented):
Looks good to me, 
please squash the comments before the merge,  thanks.

liulinC (approved):

PR #4420, opened 2021-05-24
REQ-403: check certs haven't expired before installing them

lippirk:


ocaml/xapi/certificates.ml
```diff
@@ -282,12 +285,33 @@ let cert_perms kind =
   debug "%d %d" perm stat.Unix.st_perm ;
   perm
 
+let assert_valid kind ~name ~cert =
```
psafont, line 288: Validation of certificates is usually done in `ocaml/gencert/lib.ml` could you add the function there instead? 

The function `ensure_validity` can be extracted from `validate_certificate` and be reused to generate new error messages, (`ca_certificate_not_valid_yet` and `ca_certificate_expired`)

ocaml/xapi/certificates.ml
```diff
@@ -282,12 +285,33 @@ let cert_perms kind =
...
+  ) else
+    ()
+
 let host_install kind ~name ~cert =
   if is_unsafe kind name then
     raise_name_invalid kind name ;
   let filename = library_filename kind name in
   if Sys.file_exists filename then
     raise_already_exists kind name ;
+  assert_valid kind ~name ~cert ;
```
psafont, line 314: Can this check be done on on xapi_pool.ml in the certificate_install function? This forces the check to be done earlier, on the main host.

This also means that the check is _not_ performed for CRLs, which would fail because they cannot be parsed as normal certificates.

psafont (changes requested):
The check should not be performed for CRLs, and I'd like to reuse the existing code in gencert. In the long term run I'd like to drop the code in certificates and use cert_distrib instead that's I'm being so pushy here to avoid adding code in certificates.

ocaml/xapi/certificates.ml
```diff
@@ -282,12 +285,33 @@ let cert_perms kind =
...
+let assert_valid kind ~name ~cert =
+  let cert =
+    match X509.Certificate.decode_pem (Cstruct.of_string cert) with
+    | Error (`Msg msg) ->
+        D.error "assert_valid: invalid cert. name='%s', msg='%s'" name msg ;
+        raise_corrupt kind name
+    | Ok cert ->
+        cert
+  in
+  let from', to' = X509.Certificate.validity cert in
```
lindig, line 297: You could use `starting`,`ending` to avoid using reserved words. I think the meaning would be a little more obvious.

lindig (approved):

lindig (approved):

ocaml/gencert/lib.ml
```diff
@@ -57,6 +57,58 @@ let validate_private_key pkcs8_private_key =
      )
   >>= ensure_key_length
 
+(* TODO gencert shouldn't have anything to do with CA certificates.
+ * This file should probably become a library with generally useful functions
+ * for operating on certificates *)
+type cert_kind = CA | Server
+
+let pem_of_string ~kind x =
```
psafont, line 65: Instead of using variant, it might be nicer to add the error parameter as
```ocaml
let pem_of_string ~kind x invalid_error =
```
Same with assert_not_expired:
```ocaml
let assert_not_expired ~kind ~now error_not_yet error_expired certificate =
```

lippirk: I agree this is probably the right thing to do, at the cost of an awkward interface

psafont (approved):

PR #4423, opened 2021-05-26
CP-37014 verify TLS-based RPC before enabling it

lindig:
Before enabling certificate checking, check that all hosts can
communicate when using it. This is implemented in a ping call that uses
certificate checking before it is enabled by default. A host pings all
pool members and all hosts do this based on code in message_forwarding.

### DIFF
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -988,14 +988,22 @@ functor
         Xapi_pool_helpers.with_pool_operation ~__context
           ~doc:"Pool.enable_tls_verification" ~self ~op:`tls_verification_enable
           (fun () ->
+            debug "Pool.enable_tls_verification start ... (1/2)" ;
             Cert_distrib.exchange_certificates_among_all_members ~__context ;
+            while Xapi_fist.pause_after_cert_exchange () do
+              debug "Pool.enable_tls_verification sleeping on fistpoint" ;
+              Thread.delay 5.0
+            done ;
             all_hosts
             |> List.iter (fun host ->
                    do_op_on ~local_fn ~__context ~host (fun session_id rpc ->
                        Client.Pool.enable_tls_verification rpc session_id
-                   )
+                   ) ;
+                   debug "Pool.enable_tls_verification enabling on host %s"
+                     (Ref.string_of host)
                ) ;
-            Db.Pool.set_tls_verification_enabled ~__context ~self ~value:true
+            Db.Pool.set_tls_verification_enabled ~__context ~self ~value:true ;
+            debug "Pool.enable_tls_verification completed (2/2)"
         )
 
       let set_repositories ~__context ~self ~value =
--- a/ocaml/xapi/xapi_pool.ml
+++ b/ocaml/xapi/xapi_pool.ml
@@ -3240,7 +3240,41 @@ let alert_failed_login_attempts () =
             ~body:stats
       )
 
+let perform ~local_fn ~__context ~host op =
+  let rpc' context hostname (task_opt : API.ref_task option) xml =
+    let open Xmlrpc_client in
+    let verify_cert = Some Stunnel.pool (* verify! *) in
+    let task_id = Option.map Ref.string_of task_opt in
+    let http = xmlrpc ?task_id ~version:"1.0" "/" in
+    let port = !Constants.https_port in
+    let transport = SSL (SSL.make ~verify_cert ?task_id (), hostname, port) in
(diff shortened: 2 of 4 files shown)


ocaml/xapi/xapi_pool.ml
```diff
@@ -3228,7 +3228,41 @@ let alert_failed_login_attempts () =
...
+        , !Constants.https_port
+        )
+    in
+    let http =
+      xmlrpc ?task_id:(Option.map Ref.string_of task_opt) ~version:"1.0" "/"
+    in
+    XMLRPC_protocol.rpc ~srcstr:"xapi" ~dststr:"dst_xapi" ~transport ~http xml
+  in
+  let open Message_forwarding in
+  do_op_on_common ~local_fn ~__context ~host op (call_slave_with_session rpc')
```
psafont, line 3249: I'm surprised this doesn't create a module circular dependency and dune is happy about this

psafont (approved):
The approach looks sensible. It's a bit tough to understand what's going on, some comment explaining the intent alongside the code would help there.

psafont (approved):

lindig:
I have added some comments. I think we need to see how long this takes in larger pools and whether we need to make it more asynchronous. However, this would be a one-time cost.

ocaml/xapi/message_forwarding.ml
```diff
@@ -988,14 +988,17 @@ functor
...
           (fun () ->
+            debug "Pool.enable_tls_verification start ... (1/3)" ;
             Cert_distrib.exchange_certificates_among_all_members ~__context ;
             all_hosts
             |> List.iter (fun host ->
                    do_op_on ~local_fn ~__context ~host (fun session_id rpc ->
                        Client.Pool.enable_tls_verification rpc session_id
-                   )
+                   ) ;
+                   debug "Pool.enable_tls_verification continuing ...  (2/3)"
```
lippirk, line 998: This will print (2/3) n times. Maybe you want to say instead `"enabling on host=%s" host`

ocaml/xapi/message_forwarding.ml
```diff
@@ -988,14 +988,17 @@ functor
         Xapi_pool_helpers.with_pool_operation ~__context
           ~doc:"Pool.enable_tls_verification" ~self ~op:`tls_verification_enable
           (fun () ->
+            debug "Pool.enable_tls_verification start ... (1/3)" ;
             Cert_distrib.exchange_certificates_among_all_members ~__context ;
```
lippirk, line 992: If you want to add a fistpoint for testing as @psafont  suggested, I'd have thought here is a good place (just after the exchange)

ocaml/xapi/xapi_pool.ml
lippirk: xapi_pool isn't actually used in message_forwarding.ml (we get the desired runtime behaviour because of the functor)

ocaml/xapi/xapi_pool.ml
```diff
@@ -3228,7 +3228,50 @@ let alert_failed_login_attempts () =
...
+            ()
+        , hostname
+        , !Constants.https_port
+        )
+    in
+    let http =
+      xmlrpc ?task_id:(Option.map Ref.string_of task_opt) ~version:"1.0" "/"
+    in
+    XMLRPC_protocol.rpc ~srcstr:"xapi" ~dststr:"dst_xapi" ~transport ~http xml
+  in
```
lippirk, line 3248: Instead of redefining the rpc function, we could add a optional `force_cert_check` parameter to `message_forwarding.ml:remote_rpc_retry` and use that here

ocaml/xapi/xapi_pool.ml
```diff
@@ -3228,7 +3228,50 @@ let alert_failed_login_attempts () =
...
+
+(** [ping host] calls [Pool.is_slave] using a TLS connection that uses
+certficate checking. We ignore the result but are interested in any
+failures that would indicate connection problems, which would raise an
+exception. We can't use the standard
+[Message_forwarding.do_op_on_common] because certificate checking is not
+yet enables. Hence we the local [perform], which uses certificate
+checking *)
+
+let ping ~__context host =
```
lippirk, line 3260:
```suggestion
let ping_with_tls_verif ~__context host =
```

ocaml/xapi/xapi_pool.ml
lindig: I am not sure I agree. This requires quite a bit of parameter threading in a code base that is complex already. Here we do one thing locally and that is it. I've cleaned up the code a bit.

lindig:
I have added a fistpoint for testing and I believe this works

lippirk (approved):

PR #4426, opened 2021-05-27
Manage extauth-hook-AD plugin

liulinC:


scripts/plugins/extauth-hook-AD.py
```diff
@@ -9,138 +9,268 @@
...
 
-etc_pamd_sshd_start_boilerplate = """#%PAM-1.0
+    def apply(self):
+        self._apply_to_cache()
+        self._install()
+
+    def _install(self):
+        with tempfile.NamedTemporaryFile(prefix="extauth-", delete=False) as f:
+            lines = [line.encode() for line in self._lines]
+            f.writelines(lines)
```
lippirk, line 75: I'd use `join` here (and remove this line below `self._lines = ["{}\n".format(l) for l in lines[:-1]] + lines[-1:]`)

```suggestion
            f.write('\n'.join([l.encode() for l in self._lines])
```

lippirk: In you can remove occurrences of `\n` from a lot of places with this

scripts/plugins/extauth-hook-AD.py
```diff
@@ -9,138 +9,268 @@
...
+        self._sep = sep
+        self._values = {}
+
+    def _update_value(self, key, value):
+        self._values[key] = value
+
+    def _apply_value(self, key, value):
+        try:
+            target_line = "{}{}{}\n".format(key, self._sep, value)
+            for idx, line in enumerate(self._lines):
```
lippirk, line 171: instead of using a list of encoded strings to store the key value pairs, we should really use a dictionary

scripts/plugins/extauth-hook-AD.py
liulinC: the point here is to  add `\n` to each line except the last line, can the suggestion match this?

scripts/plugins/extauth-hook-AD.py
liulinC: The key value pairs do stored in dict in `self._values`,   

the `self._lines` are raw line read from original existing configuration file.

scripts/plugins/extauth-hook-AD.py
```diff
@@ -9,138 +9,268 @@
...
-    # Active Directory
+def refresh_dynamic_pam(session, args):
     try:
-        fd, fname = tempfile.mkstemp(prefix="sshd-", dir="/etc/pam.d")
-        os.write(fd, etc_pamd_sshd_ad_disabled)
-        os.rename(fname, "/etc/pam.d/sshd")
+        ConfigManager(session, args).refresh_dynamic_pam()
         return str(True)
-    except:
-        return "ERROR_1: revert_etc_pamd_ssh failed"
```
liulinC, line 124: should return string value on exception

scripts/plugins/extauth-hook-AD.py
lippirk: yes, see this example:

```
>>> '\n'.join(['a', 'b', 'c'])
'a\nb\nc'
```

scripts/plugins/extauth-hook-AD.py
lippirk: > The key value pairs do stored in dict in self._values,

Indeed, you have `def _update_value`, so why do we need `def _apply_value` as well? It's like you are storing key values pairs in both `self._values` and in `self._lines` which I don't understand

scripts/plugins/extauth-hook-AD.py
```diff
@@ -9,138 +9,268 @@
...
+        self._update_value("shadow", modules)
+
+
+class SshdConfig(KeyValueConfig):
+    def __init__(self, session, args, ad_enabled=True):
+        super(SshdConfig, self).__init__("/etc/ssh/sshd_config", session, args, ad_enabled, sep=" ")
+        value = "yes" if ad_enabled else "no"
+        self._values["ChallengeResponseAuthentication"] = value
+        self._values["GSSAPIAuthentication"] = value
+        self._values["GSSAPICleanupCredentials"] = value
```
lippirk, line 210: You abstracted away the `self._values` dictionary in the `KeyValueConfig` class with `def _update_value`, so you should be consistent here

scripts/plugins/extauth-hook-AD.py
```diff
@@ -9,138 +9,268 @@
...
+            f.writelines(lines)
+            os.rename(f.name, self._file_path)
+
+
+class StaticPam(ADConfig):
+    ad_pam_format = """#%PAM-1.0
 auth        required      pam_env.so
 auth        sufficient    pam_unix.so try_first_pass nullok
-auth        sufficient    /lib/security/pam_lsass.so try_first_pass
+auth        sufficient    {ad_module} try_first_pass try_authtok
```
lippirk, line 83: Will changing `try_first_pass` to `try_authtok` break pbis?

scripts/plugins/extauth-hook-AD.py
```diff
@@ -9,138 +9,268 @@
...
-session     sufficient    /lib/security/pam_lsass.so
+session     sufficient    {ad_module}
 
-account    required       pam_nologin.so
-account     required      /lib/security/pam_lsass.so unknown_ok
-# Start of list of allowed AD groups and users
-"""
+account     required      pam_nologin.so
+account     required      {ad_module} unknown_ok
+account     include       hcp_users
```
lippirk, line 95: I don't think we want to add this hcp_users here?

scripts/plugins/extauth-hook-AD.py
```diff
@@ -9,138 +9,268 @@
...
+account     required      pam_unix.so"""
```
lippirk, line 96: If winbind requires this, is it safer to add it only when the backend is winbind?

lippirk: (similarly for `session sufficient pam_lsass.so`)

scripts/plugins/extauth-hook-AD.py
liulinC: `self._lines` is general for all kinds of configurations,  (not only for key value pair configuration),    It is used as cache (which will be write to file as configuration).  It will be initialized with file existing content on demand. 

`self._values` is only specific for key value pair configurations,  as you said, it is a dict to store key, values. 
Since KeyValueCOnfig needs to reuse the parent function like `_install`, so it needs to update its value to `self._lines` on its `apply()`

liulinC:
Thank @lippirk for the help, could you please help to review at your convenient , 😄

scripts/plugins/extauth-hook-AD.py
lippirk: > every subject (user or group) with Pool Admin role is added to /etc/pam.d/hcp_users as follows

so hcp_users is actually a list of pool admins, rather than a list of hcp users? calling the file `/etc/pam.d/pool_admins` would make more sense to me

lippirk: additionally do we want to add `account include hcp_users` in the pbis case?

scripts/plugins/extauth-hook-AD.py
lippirk: in that case can we store the 'key value pair configuration' separately from the 'general configuration'? I think that would simplify this code here

> Since KeyValueCOnfig needs to reuse the parent function like _install, so it needs to update its value to self._lines on its apply()

I think it would be simpler to have the install function generate the string from an intermediate representation, rather than having to maintain/update the line strings

scripts/plugins/extauth-hook-AD.py
liulinC: yes, we now we only add pool admin to this file, as we only permit pool admin to login ATM, 

yes, we do need to add `account include hcp_users` for pbis case, this is not specific to pbis or winbind, 
we should align both.

scripts/plugins/extauth-hook-AD.py
liulinC: > in that case can we store the 'key value pair configuration' separately from the 'general configuration'? I think that would simplify this code here
> 
> > Since KeyValueCOnfig needs to reuse the parent function like _install, so it needs to update its value to self._lines on its apply()
> 
They are stored separately from the `general configuration`, let's have a talk offline. 
> I think it would be simpler to have the install function generate the string from an intermediate representation, rather than having to maintain/update the line strings

scripts/plugins/extauth-hook-AD.py
```diff
@@ -9,138 +9,284 @@
...
+        # and the list of subjects from xapi
+        subjects = self._session.xenapi.subject.get_all()
+        admin_role = self._session.xenapi.role.get_by_name_label('pool-admin')[0]
         # Add each subject which contains the admin role
         for opaque_ref in subjects:
-            subject_rec = session.xenapi.subject.get_record(opaque_ref)
+            subject_rec = self._session.xenapi.subject.get_record(opaque_ref)
             sid = subject_rec['subject_identifier']
+            name = subject_rec["other_config"]["subject-name"]
+            is_group = True if subject_rec["other_config"]["subject-is-group"] == "true" else False
```
lippirk, line 153:
```suggestion
            is_group = subject_rec["other_config"]["subject-is-group"] == "true"
```

lippirk (approved):
I think the code inside `KeyValueConfig` could be made much simpler by maintaining a dictionary instead of modifying the config strings themselves (see my comments above)

I would have perhaps been more confident about these changes not affecting the PBIS implementation if we had separated the plugins into two files: `extauth-hook-pbis-AD.py`, and `extauth-hook-pbis-AD.py`

Other than these two things, the changes look mostly correct

scripts/plugins/extauth_hook_ad.py
```diff
@@ -0,0 +1 @@
+extauth-hook-AD.py
```
lippirk, line 1: why is this new file necessary?

liulinC: This is for unittest,  and it will not be installed into dom0

The file name contains `-`, which is not supported by module names.

liulinC:
> I think the code inside `KeyValueConfig` could be made much simpler by maintaining a dictionary instead of modifying the config strings themselves (see my comments above)
> 
> I would have perhaps been more confident about these changes not affecting the PBIS implementation if we had separated the plugins into two files: `extauth-hook-pbis-AD.py`, and `extauth-hook-pbis-AD.py`
> 
> Other than these two things, the changes look mostly correct

The `KeyValueConfig` is not exactly match the key value pairs, as some special lines, like comments,  empty lines. 
Also, the line order needs to be kept.  (Perhaps `OrderedDict` can help with that)
I am considering whether workaround this issue simper than current solution.

I understand improving confidence by keeping `extauth-hook-pbis-AD.py` unchanged, but it will leave some duplicated code (splitting the common code into a module requires modify PBIS anyway).    But the pbis plugin will be deleted finally. 

So it maybe a good idea to keep pbis plugin unchanged.

scripts/plugins/extauth-hook-AD.py
```diff
@@ -9,138 +9,304 @@
...
+        self._args = args
+        self._lines = []
+        self._backend = self._get_ad_backend(args)
+        self._ad_enabled = ad_enabled
+        self._file_mode = file_mode
+        if load_existing:
+            if os.path.exists(self._file_path):
+                with open(self._file_path, 'r') as f:
+                    lines = f.readlines()
+                    self._lines = [l.strip() for l in lines]
```
psafont, line 66:
```suggestion
        if load_existing and os.path.exists(self._file_path):
            with open(self._file_path, 'r') as f:
                lines = f.readlines()
                self._lines = [l.strip() for l in lines]
```

scripts/plugins/extauth-hook-AD.py
```diff
@@ -9,138 +9,304 @@
...
+        subjects = self._session.xenapi.subject.get_all()
+        admin_role = self._session.xenapi.role.get_by_name_label('pool-admin')[0]
         # Add each subject which contains the admin role
         for opaque_ref in subjects:
-            subject_rec = session.xenapi.subject.get_record(opaque_ref)
+            subject_rec = self._session.xenapi.subject.get_record(opaque_ref)
             sid = subject_rec['subject_identifier']
+            name = subject_rec["other_config"]["subject-name"]
+            is_group = subject_rec["other_config"]["subject-is-group"] == "true"
+            logger.debug("Permit %s with sid %s is_group as %s", name, sid, is_group)
```
psafont, line 155: Shouldn't this logline be part of the if statement?

scripts/plugins/extauth-hook-AD.py
```diff
@@ -9,138 +9,304 @@
...
             if admin_role in subject_rec['roles']:
-                config.add_subject(sid)
-        config.install()
+                self._add_subject(name, is_group)
+
+    def _add_subject(self, name, is_group):
+        condition = "ingroup {}".format(name) if is_group else "= {}".format(name)
+        self._lines.append("account sufficient pam_succeed_if.so user {} ".format(condition))
```
psafont, line 161:
```suggestion
        condition = "ingroup" if is_group else "="
        self._lines.append("account sufficient pam_succeed_if.so user {} {} ".format(condition, name))

File: scripts/plugins/extauth-hook-AD.py [hunk 13]
```diff
@@ -9,138 +9,304 @@
...
+    def _install(self):
+        if self._ad_enabled:
+            super(DynamicPam, self)._install()
+        else:
+            if os.path.exists(self._file_path):
+                os.remove(self._file_path)
+
+
+class KeyValueConfig(ADConfig):
+    _empty_line_prefix = "__key_vluae_config_empty_line_prefix_" # Presume normal config does not have such keys
```
psafont, line 172: This seems to have a typo

```suggestion
    _empty_line_prefix = "__key_value_config_empty_line_prefix_" # Presume normal config does not have such keys
```

scripts/plugins/extauth-hook-AD.py
```diff
@@ -9,138 +9,304 @@
...
+    def _is_empty_line(self, line):
+        return line.startswith(self._empty_line_prefix)
+
+    def _load_values(self):
+        for idx, line in enumerate(self._lines):
+            if line == "": # Empty line
+                key = "{}{}".format(self._empty_line_prefix, str(idx)) # Generate a unique key to store multiple empty lines
+                self._values[key] = self._empty_value
+            elif self._is_comment_line(line):
+                self._values[line] = self._empty_value # Store the comment as key, Presume comments are unique
```
psafont, line 194: can't index be used here, like on the empty lines?

psafont (changes requested):
While I don't know much about PAM, the code in general looks like it's in good shape.
I'll ask the tests to be run by the CI before I can approve the PR, currently only the tests in examples are being run:

```
+ nosetests3 scripts scripts/examples scripts/examples/python
..................
----------------------------------------------------------------------
Ran 18 tests in 0.057s

OK
```

scripts/plugins/extauth-hook-AD.py
liulinC: oh, it should be, nice catch

scripts/plugins/extauth-hook-AD.py
liulinC: yes, it can be used,  but I think that is over designed, 

The index needs to be append to the comments and split out during save,
The files we update actually does not have duplicated comments, so let's just keep it simple.

liulinC:
> While I don't know much about PAM, the code in general looks like it's in good shape.
> I'll ask the tests to be run by the CI before I can approve the PR, currently only the tests in examples are being run:
> 
> ```
> + nosetests3 scripts scripts/examples scripts/examples/python
> ..................
> ----------------------------------------------------------------------
> Ran 18 tests in 0.057s
> 
> OK
> ```

The CI run the unittest, please help to check, thanks.

.github/python-nosetests.sh
```diff
@@ -6,5 +6,8 @@ set -uex
 sudo add-apt-repository -y "deb http://archive.ubuntu.com/ubuntu/ trusty main universe"
 sudo apt-get update
 sudo apt-get install -y python-mock python-nose
+curl https://bootstrap.pypa.io/pip/2.7/get-pip.py |sudo python2
+pip install enum
```
psafont, line 10: I'd prefer something like this instead if curl-ing a script:
```suggestion
sudo apt-get install -y pip
pip install enum mock nose
```

psafont (changes requested):
Please add the tests to `.github/python-nosetests-3.sh` as well to run the tests under python 3 as well.

.github/python-nosetests.sh
liulinC: Cann install pip simply by run `apt-get install -y pip` (install pip3), or `apt-get install -y python-pip`, (get following errors about packages dependencies)

Btw, pip officially recommend instal with curl, 
https://pip.pypa.io/en/stable/installing/

```The following packages have unmet dependencies:
62
 python-pip : Depends: python (>= 2.7)
63
              Depends: python (< 2.8)
64
              Depends: python:any (>= 2.7.1-0ubuntu2)
65
              Depends: python-colorama but it is not going to be installed
66
              Depends: python-distlib but it is not going to be installed
67
              Depends: python-requests but it is not going to be installed
68
              Recommends: python-dev-all (>= 2.6) but it is not installable
69
E: Unable to correct problems, you have held broken packages.
70
Error: Process completed with exit code 100.

psafont: The package `python2-pip` or similar doesn't seem to exist in ubuntu 20.04 LTS

psafont (approved):
The code is tested on CI, both on python2 and python3

PR #4427, opened 2021-05-27
CA-354834 log ref, uuid when adding CA cert

lindig:
To track adding and removal of CA certificates, log the uuid and ref
under which it is added to the database.


psafont (approved):

lippirk:
we may as well add it to the remove functions too

edwintorok (approved):

PR #4428, opened 2021-05-28
CP-36178: Add basic precheck function for updates

robhoes:
Also includes some minor improvements.

### DIFF
--- a/ocaml/idl/datamodel_pool.ml
+++ b/ocaml/idl/datamodel_pool.ml
@@ -748,6 +748,19 @@ open Datamodel_types
       ~allowed_roles:_R_POOL_OP
       ()
 
+  let check_update_readiness = call
+      ~name:"check_update_readiness"
+      ~lifecycle:[Published, rel_next, ""]
+      ~doc:"Check if the pool is ready to be updated. If not, report the reasons."
+      ~params:[
+        Ref _pool, "self", "The pool";
+        Bool, "requires_reboot", "Assume that the update will require host reboots"
+      ]
+      ~result:(Set(Set (String)), "A set of error codes with arguments, if the pool is
+        not ready to update. An empty list means the pool can be updated.")
+      ~allowed_roles:_R_READ_ONLY
+      ()
+
   (** A pool class *)
   let t =
     create_obj
@@ -830,6 +843,7 @@ open Datamodel_types
         ; add_repository
         ; remove_repository
         ; sync_updates
+        ; check_update_readiness
         ]
       ~contents:
         ([uid ~in_oss_since:None _pool] @
--- a/ocaml/idl/datamodel_pool.ml
+++ b/ocaml/idl/datamodel_pool.ml
@@ -224,21 +224,6 @@ open Datamodel_types
       ~allowed_roles:_R_POOL_OP
       ()
 
-  let slave_network_report = call
-      ~name:"slave_network_report"
-      ~in_oss_since:None
-      ~in_product_since:rel_rio
-      ~doc:"Internal use only"
-      ~params:[Map (String, String), "phydevs", "(device,bridge) pairs of physical NICs on slave";
-               Map (String, String), "dev_to_mac", "(device,mac) pairs of physical NICs on slave";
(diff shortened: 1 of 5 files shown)


ocaml/xapi/rbac_audit.ml
```diff
@@ -443,19 +443,24 @@ let has_to_audit action =
           [
             (* list of _actions_ filtered out from the audit log *)
             "session.local_logout"
+          ; "session_local_logout"
```
lippirk, line 446: why are we duplicating these?

ocaml/xapi/xapi_pool.ml
```diff
@@ -3325,6 +3321,28 @@ let sync_updates ~__context ~self ~force =
       List.iter (fun x -> create_pool_repository ~__context ~self:x) enabled ;
       set_available_updates ~__context
 
+let check_update_readiness ~__context ~self ~requires_reboot =
```
lippirk, line 3324: by this logic, we consider a pool 'ready to upgrade' if there are no miscreant VMs, and if all the hosts are up

what about if there's a cluster running, or HA is enabled, etc. ?

ocaml/xapi/xapi_pool.ml
```diff
@@ -3325,6 +3321,28 @@ let sync_updates ~__context ~self ~force =
...
+  List.map
```
lippirk, line 3325:
```suggestion
  List.concat_map
```

sadly there is no `filter_concat_map`

ocaml/xapi/message_forwarding.ml
```diff
@@ -1021,6 +1021,12 @@ functor
           (pool_uuid ~__context self)
           (string_of_bool force) ;
         Local.Pool.sync_updates ~__context ~self ~force
+
+      let check_update_readiness ~__context ~self ~requires_reboot =
+        info "Pool.check_update_readiness: pool = '%s'; requires_reboot = %s"
```
lippirk, line 1026: you can use %b here

ocaml/idl/datamodel_pool.ml
```diff
@@ -763,6 +748,19 @@ open Datamodel_types
...
 
+  let check_update_readiness = call
+      ~name:"check_update_readiness"
+      ~lifecycle:[Published, rel_next, ""]
+      ~doc:"Check if the pool is ready to be updated. If not, report the reasons."
+      ~params:[
+        Ref _pool, "self", "The pool";
+        Bool, "requires_reboot", "Assume that the update will require host reboots"
+      ]
+      ~result:(Set(Set (String)), "A set of error codes with arguments, if the pool is
```
lippirk, line 759: as far as I can tell, it should be `Set (Pair String))`? not sure if the datamodel supports that though

robhoes: No, it is not a pair, but list: the first element is always the error code and then there are zero or more arguments (as in datamodel_errors.ml). A better type would be `string * string list`, but the datamodel doesn't support tuples.

ocaml/xapi/xapi_pool.ml
robhoes: Yeah, it just so happens that our update service already checks for HA and WLB, and disables them proactively. I would still add checks for them for extra safety though.

minglumlu (approved):

ocaml/idl/datamodel_pool.ml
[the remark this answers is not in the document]
robhoes: Unfortunately, the datamodel types are not flexible enough to make this much better; see my reply to Ben's comment above.

ocaml/xapi/xapi_pool.ml
robhoes: I've added the HA check now. We are not currently convinced that anything else is needed.

ocaml/xapi/xapi_pool.ml
```diff
@@ -3325,6 +3321,28 @@ let sync_updates ~__context ~self ~force =
...
+    (fun host ->
```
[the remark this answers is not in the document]
robhoes: I've made this change now.

PR #4430, opened 2021-06-02
CA-355179: Support epoch in RPM;CA-355180: Improve parsing output of 'yum list updates'

minglumlu:
CA-355179: Support epoch in RPM

    This commit adds logic to support epoch in RPM packages.

    From https://rpm-packaging-guide.github.io/,  "the Epoch version would
    win out over the traditional Name-Version-Release marker that signifies
    versioning for RPM Packages."

    CA-355180: Improve parsing output of 'yum list updates'

    This commit improves the logic of parsing output of "yum list updates".
    Usually each line of the output would look like:
    "qemu-dp.x86_64                             2:2.12.0-2.0.11.xs8 local"
    But for some reason, it could be:
    "qemu-dp.x86_64
                                 2:2.12.0-2.0.11.xs8 local".
    This means one line can be separated by newline as well. This commit
    aims to handle these cases.

### DIFF
--- a/ocaml/tests/test_repository_helpers.ml
+++ b/ocaml/tests/test_repository_helpers.ml
@@ -1833,12 +1833,14 @@ module EvalGuidanceForOneUpdate = Generic.MakeStateless (struct
       ]
 end)
 
-module GetRpmUpdateInJson = Generic.MakeStateless (struct
+module GetUpdateInJson = Generic.MakeStateless (struct
   module Io = struct
     (* (name.arch, updateId) list: from "yum updateinfo list updates"
      * (name.arch, Pkg.t) list): from "rpm -qa"
-     * string: output line of "yum list updates" *)
-    type input_t = ((string * string) list * (string * Pkg.t) list) * string
+     * (Pkg.t * string): from "yum list updates"
+     *)
+    type input_t =
+      ((string * string) list * (string * Pkg.t) list) * (Pkg.t * string)
 
     type output_t = (Yojson.Basic.t option, exn) result
 
@@ -1851,7 +1853,7 @@ module GetRpmUpdateInJson = Generic.MakeStateless (struct
                  (list (pair string string))
                  (list (pair string (record @@ fields_of_pkg)))
               )
-              string
+              (pair (record @@ fields_of_pkg) string)
           )
       )
 
@@ -1863,8 +1865,8 @@ module GetRpmUpdateInJson = Generic.MakeStateless (struct
           Fmt.(str "%a" exn) e
   end
 
-  let transform ((rpm2updates, installed_pkgs), line) =
-    try Ok (get_rpm_update_in_json ~rpm2updates ~installed_pkgs line)
+  let transform ((update_ids, installed_pkgs), (new_pkg, repo)) =
+    try Ok (get_update_in_json ~update_ids ~installed_pkgs (new_pkg, repo))
     with e -> Error e
 
   let tests =
@@ -1900,7 +1902,17 @@ module GetRpmUpdateInJson = Generic.MakeStateless (struct
                 )
               ]
             )
-          , "libpath-utils.noarch    0.2.2-1.el7      epel"
+          , ( Pkg.
+                {
+                  name= "libpath-utils"
+                ; epoch= None
+                ; version= "0.2.2"
+                ; release= "1.el7"
+                ; arch= "noarch"
+                }
+              
+            , "epel"
+            )
             (* repository name is "epel" *)
           )
         , Error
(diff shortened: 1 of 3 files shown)


robhoes (approved):
This PR extends the current yum repository management functionality to recognise the "epoch" field, which was previously not taken into account. The change is a lot larger than I had hoped due to all the additional parsing and (un)marshalling.

It was pointed out to me that using a parser combinator library like Angstrom would be better than all this ad hoc parsing using the string operators and regexps. I tend to agree with that, as this may make the code as a whole easier to read and understand. However, the problem here is not simply parsing the output of a single command. Similar data is coming in from updates XML files, and there is JSON being returned to clients. At this point, I think we should stick with what we have now, and we can reassess the situation once the dust has settled. Thankfully we have unit test, otherwise we'd be lost.

minglumlu:
> using a parser combinator library like Angstrom would be better
Sure. I will look at the parser combinator to see how to raise more PRs for better handling both outputs of command lines and JSON. Thanks for guidance!

ocaml/xapi/repository_helpers.ml
```diff
@@ -20,13 +20,83 @@ module UpdateIdSet = Set.Make (String)
...
+     *   "(none):1.1.11-34".
+     *
+     * These may come from:
+     *   "yum list updates",
+     *   "yum updateinfo list updates", or
+     *   "rpm -qa" *)
+    let open Rresult.R.Infix in
+    ( ( match Astring.String.cuts ~sep:":" epoch_ver_rel with
+      | [e; vr] -> (
+        try Result.Ok (Epoch.of_string e, vr)
```
psafont, line 71: There is not need to write `Result` when using `Ok` or `Error`

ocaml/xapi/repository_helpers.ml
```diff
@@ -51,47 +123,104 @@ module Pkg = struct
...
+      let msg = "Can't construct an update from json" in
+      error "%s: %s" msg (ExnHelper.string_of_exn e) ;
+      raise Api_errors.(Server_error (internal_error, [msg]))
 
   let to_string u =
-    Printf.sprintf "%s.%s %s-%s -> %s-%s from %s:%s" u.name u.arch
+    Printf.sprintf "%s.%s %s:%s-%s -> %s:%s-%s from %s:%s" u.name u.arch
+      ( match u.old_epoch with
+      | Some e -> (
+        match e with Some i -> string_of_int i | None -> Epoch.epoch_none
```
lippirk, line 211: Is there a reason not to use `Epoch.to_string` here?

ocaml/xapi/repository_helpers.ml
```diff
@@ -51,47 +123,104 @@ module Pkg = struct
...
+          Epoch.epoch_none
+      )
       (Option.value u.old_version ~default:"unknown")
       (Option.value u.old_release ~default:"unknown")
+      ( match u.new_epoch with
+      | Some i ->
+          string_of_int i
+      | None ->
+          Epoch.epoch_none
+      )
```
lippirk, line 223: can we use `Epoch.to_string` here too?

ocaml/xapi/repository_helpers.ml
```diff
@@ -51,47 +123,95 @@ module Pkg = struct
...
+          release arch
+    | None ->
+        Printf.sprintf "%s-%s-%s.%s.rpm" name version release arch
 end
 
 module Update = struct
   type t = {
       name: string
     ; arch: string
+    ; old_epoch: Epoch.t option
```
psafont, line 158: Could you explain why is `Epoch.t option` used here instead of `Epoch.t`? I couldn't spot a difference in behaviour between `None` and `Some None`. This change would also simplify the code around comparisons.

ocaml/tests/test_repository_helpers.ml
```diff
@@ -1990,6 +2620,156 @@ module GuidanceSetResortGuidancesTest = Generic.MakeStateless (struct
...
+      List.fold_left parse_line_of_list_updates ((None, None, None), []) lines
+    in
+    updates
+
+  let tests =
+    `QuickAndAutoDocumented
+      [
+        ( [
+            "xenserver-status-report.noarch"
+          ; "         1.3.1-1.xs8         \
```
lippirk, line 2646: It'd be good to understand why `yum` is outputting such inconsistent formats... but at least our parsing is looking bulletproof :+1:

lindig (approved):

ocaml/xapi/repository_helpers.ml
minglumlu: Yes. Oh, silly me.

ocaml/xapi/repository_helpers.ml
minglumlu: An update has:
1. new package, this comes from `yum list updates`
2. old package, this comes from `rpm -qa`

Since they come from different sources, one maybe case is that a package in `yum list updates` but not in `rpm -qa`. 
And meanwhile, an epoch itself might be None as well. So epoch is represented as `Epoch.t`(`int option`), while the old epoch is  represented as `Epoch.t option`

ocaml/tests/test_repository_helpers.ml
minglumlu: Indeed. I can't get the output manually. And I can't find a configuration of YUM to control that. But we did observe this from XAPI log.

minglumlu:
Hi @psafont @lippirk 
I've addressed your comments. May I have your reviews again? Thanks.

lippirk (approved):

PR #4431, opened 2021-06-03
Build the doc-json target as part of the install target. Also:

kc284:
Restructured output so it's easier to drop it into xapi-project.github.io.
Added missing release date (or the release is not generated by gen-json.exe).

### DIFF
--- a/ocaml/idl/datamodel_types.ml
+++ b/ocaml/idl/datamodel_types.ml
@@ -55,7 +55,6 @@ let rel_indigo = "indigo"
 let rel_dundee = "dundee"
 let rel_ely = "ely"
 let rel_falcon = "falcon"
-let rel_honolulu = "honolulu"
 let rel_inverness = "inverness"
 let rel_jura = "jura"
 let rel_kolkata = "kolkata"
@@ -210,12 +209,6 @@ let release_order_full = [{
      version_minor = 6;
      branding      = "XenServer 7.1";
      release_date  = Some "February 2017";
-   }; {
-     code_name     = Some rel_honolulu;
-     version_major = 2;
-     version_minor = 6;
-     branding      = "XenServer 7.1 CU1";
-     release_date  = Some "September 2017";
    }; {
      code_name     = Some rel_falcon;
      version_major = 2;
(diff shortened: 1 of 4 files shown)


psafont (approved):

ocaml/idl/datamodel_types.ml
```diff
@@ -275,7 +275,7 @@ let release_order_full = [{
      version_major = 2;
      version_minor = 15;
      branding      = "Citrix Hypervisor 8.2 Hotfix 2";
```
edwintorok, line 277: Should we add a 7.1 CU2 to this list? CU1 is there, but CU2 seems to have ben forgotten

edwintorok (approved):

ocaml/idl/datamodel_types.ml
kc284: It should probably be the other way round, i.e. CU1 should be removed. There were no public API changes either in 7.1 CU1 or 7.1 CU2 (apart from a few new API errors, which don't have version info). Indeed this page http://xapi-project.github.io/xen-api/releases/honolulu.html is empty, so I don't know why CU1 was defined.

kc284:
Don't merge this yet, I may add a modification pending on the outcome of the conversation on https://github.com/xapi-project/xapi-project.github.io/pull/250

kc284:
Any idea why removal of `rel_honolulu` from the datamodel_types.ml is causing the Ocaml tests failure? I don't see it being used anywhere else, am I missing something?

robhoes:
@kc284 
```
New schema hash ('633a99b46dda090677598aad3a830f76') doesn't match the last known one. Please bump the

  schematest alias ocaml/idl/runtest (exit 1)
(cd _build/default/ocaml/idl && ./schematest.exe)
```

kc284:
I haven't changed any fields of API objects in this PR, only removed an unused release from Datamodel_types, I think this does not warrant a schema bump, correct?

robhoes:
Yeah, but still the hash may be different for some reason. So don't bump the schema version, but update the hash for the test to pass.

kc284:
Can this be merged now?

robhoes (approved):

PR #4434, opened 2021-06-11
CP-36098 introduce host-refresh-server-certificates

lindig:
Some initial scaffolding and code to refresh a host certificate. I believe the current approach can't work because we can't atomically update a server certificate and make all pool members trust it.

### DIFF
--- a/ocaml/xapi-cli-server/cli_frontend.ml
+++ b/ocaml/xapi-cli-server/cli_frontend.ml
@@ -3071,6 +3071,15 @@ let rec cmdtable_data : (string * cmd_spec) list =
       ; flags= [Host_selectors]
       }
     )
+  ; ( "host-refresh-server-certificates"
+    , {
+        reqd= []
+      ; optn= ["host-uuid"]
+      ; help= "Refresh server certificate on host with local host as default"
+      ; implementation= No_fd Cli_operations.host_refresh_server_certificates
+      ; flags= [Host_selectors]
+      }
+    )
   ; ( "host-server-certificate-install"
     , {
         reqd= ["certificate"; "private-key"]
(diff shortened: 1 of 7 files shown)


ocaml/xapi-cli-server/cli_frontend.ml
```diff
@@ -3071,6 +3071,15 @@ let rec cmdtable_data : (string * cmd_spec) list =
       ; flags= [Host_selectors]
       }
     )
+  ; ( "host-refresh-server-certificates"
+    , {
+        reqd= []
+      ; optn= []
+      ; help= "Refresh server certificate on host with local host as default"
```
lippirk, line 3078: 'local host as default' is a bit misleading, in the case where xe is not running in dom0

lippirk:
If we wanted to use machinery in cert_distrib.ml to do the cert/key cycling, I would expect it to look similar to this https://github.com/xapi-project/xen-api/blob/6b5352cbb35bce67f332981b33d1380c572b3926/ocaml/xapi/cert_distrib.ml#L348

Instead of collecting certs from every host, we only need to collect a cert from one host (the one who is generating a new certificate). We'd tell everyone about our new cert, regen bundles, make the switch to serve the new cert, then finally tell everyone to delete the old cert, and once more regen bundles. Even though we are regenerating bundles a lot of times, it is not too complicated. The thing we have to be careful of (something that pool secret rotation worries about a lot) is if a cert/key cycle fails, then performing another one should not clobber any state left over from the failed attempt

We talked about changing the file names of certs (currently we are using the format `$hostuuid.pem`) - is there a reason we can't use a temporary name like `$hostuuid.new.pem` for the 'new cert' and swap the files over when we are done? We would be out of sync with the database until the cycling is complete

ocaml/xapi-cli-server/cli_frontend.ml
lindig: I agree; the actual implementation does not have a default. I am also confused by the use of `host` and `host-uuid` parameter. Function `do_op_on_host` evaluates a `host` parameter but other API function use `host-uuid`.

```

lindig:
If we know that there are at most two certs per host (current and new), a scheme using `.new.pem` would work and had the benefit of still using known names.

lindig:
This needs a different approach.

PR #4435, opened 2021-06-11
ci: build the sdk when running the CI

psafont:
The xml file is fetched straight from the SM master branch and it's placed where the makefile expects it

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -76,6 +76,12 @@ jobs:
       - name: Run tests
         run: opam exec -- make test
 
+      - name: Build SDK
+        run: |
+          mkdir -p /opt/xensource/sm
+          wget -O /opt/xensource/sm/XE_SR_ERRORCODES.xml https://raw.githubusercontent.com/xapi-project/sm/master/drivers/XE_SR_ERRORCODES.xml
+          opam exec -- make sdk
+
       - name: Uninstall unversioned packages and remove pins
         # This should purge them from the cache, unversioned package have
         # 'master' as its version
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ scripts/examples/python/setup.py
 scripts/examples/python/XenAPI.egg-info/
 scripts/examples/python/build/
 scripts/examples/python/dist/
+
+# ignore file needed for building the SDK
+ocaml/sdk-gen/csharp/XE_SR_ERRORCODES.xml


Makefile
```diff
@@ -45,19 +44,6 @@ sdk:
 		@ocaml/sdk-gen/csharp/generate \
 		@ocaml/sdk-gen/java/generate \
 		@ocaml/sdk-gen/powershell/generate
-	mkdir -p $(XAPISDK)/c
```
kc284, line 48: Why have these lines been removed?

.github/XE_SR_ERRORCODES.xml
```diff
@@ -0,0 +1,880 @@
+<?xml version="1.0" ?>
```
kc284, line 1: Wouldn't it be better to have a way to copy this over from the SM repo instead of keeping here a file that will most likely get out of sync with the original one?

psafont: sure, it can be fetched using wget. Last update was on 2020. Is it important to keep it in sync?

robhoes: This copy is just for CI testing. Consider it a mock file just for this purpose.

robhoes:
This correctly detects the SDK build problem:
```
gen_c_binding alias ocaml/sdk-gen/c/generate (exit 2)
(cd _build/default/ocaml/sdk-gen/c && ./gen_c_binding.exe)
[S(Set);[S(Set);S(String)]]Fatal error: exception "Assert_failure ocaml/sdk-gen/c/gen_c_binding.ml:1382:6"
gen_powershell_binding alias ocaml/sdk-gen/powershell/generate (exit 2)
(cd _build/default/ocaml/sdk-gen/powershell && ./gen_powershell_binding.exe)
Fatal error: exception "Assert_failure ocaml/sdk-gen/powershell/common_functions.ml:179:6"
gen_csharp_binding alias ocaml/sdk-gen/csharp/generate (exit 2)
(cd _build/default/ocaml/sdk-gen/csharp && ./gen_csharp_binding.exe)
[S(Set);[S(Set);S(String)]]Fatal error: exception "Assert_failure ocaml/sdk-gen/csharp/gen_csharp_binding.ml:1279:6"
make: *** [Makefile:38: sdk] Error 1
Error: Process completed with exit code 2.
```

Makefile
robhoes: I think we should keep these lines in the Makefile (at least for now). Is there another way to factor out the `dune build` lines?

psafont: I've reintroduced the lines

.github/XE_SR_ERRORCODES.xml
psafont: hmm, let me try to download the file to the /opt/... location within the CI, that way we don't have to modify the Makefile

robhoes (approved):
Brilliant :) Let's see if the checks on this PR can be made to pass after merging #4436.

robhoes:
Looks like it works.

.github/XE_SR_ERRORCODES.xml
kc284: The xml file is not updated too often and at the moment the output generated via this test is not used for anything, but theoretically this might change in future. Unless it causes issues, I'd rather have it done the right way, if only to avoid giving the wrong impression.

Makefile
kc284: What exactly is the issue with these lines? Is it just repetitiveness (the logic is the same as the doc target)?

kc284 (approved):

PR #4437, opened 2021-06-15
CP-37241: Disable machine account on domain leave

liulinC:
PBIS just disable machine account during domain leave
This commit let winbind keep this behavior for backward compatibility
instead of delete the machine account

### DIFF
--- a/ocaml/tests/test_data/repository_pkg_of_fullname_all
+++ b/ocaml/tests/test_data/repository_pkg_of_fullname_all
@@ -253,7 +253,6 @@ intel-igb-5.3.5.20-2.xs8.x86_64
 nss-pem-1.0.3-4.el7.x86_64
 gssproxy-0.7.0-17.el7.x86_64
 nss-sysinit-3.36.0-7.el7_5.x86_64
-pbis-open-8.2.3-1.7.8.xs8.x86_64
 mailx-12.5-19.el7.x86_64
 system-config-firewall-base-1.2.29-10.el7.noarch
 libgomp-4.8.5-28.el7_5.1.x86_64
@@ -464,7 +463,6 @@ broadcom-bnxt-en-1.10.0_216.0.119.1-2.xs8.x86_64
 tar-1.26-34.el7.x86_64
 avago-megaraid-sas-07.713.01.00+rc1-2.xs8.x86_64
 nss-softokn-3.36.0-5.el7_5.x86_64
-pbis-open-upgrade-8.2.2-3.xs8.x86_64
 nss-tools-3.36.0-7.el7_5.x86_64
 forkexecd-1.22.0-7.xs8.x86_64
 libunistring-0.9.3-9.el7.x86_64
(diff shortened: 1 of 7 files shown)


ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -484,17 +484,17 @@ let persist_extauth_config ~domain ~user ~ou_conf ~workgroup =
...
   |> Server_helpers.exec_with_new_task "update external_auth_configuration"
 
-let clean_machine_account ~service_name = function
+let disable_machine_account ~service_name = function
   | Some u, Some p -> (
-      (* Clean machine account in DC *)
+      (* Disable machine account in DC *)
       let env = [|Printf.sprintf "PASSWD=%s" p|] in
-      let args = ["ads"; "leave"; "-U"; u; "-d"; debug_level] in
+      let args = ["ads"; "leave"; "-U"; u; "--keep-account"; "-d"; debug_level] in
```
lippirk, line 491: When you leave an AD domain using XenCenter, you can choose whether to leave the machine account enabled or not, so I would have expected to see some boolean switch alongside `--keep-account`?

liulinC: if user name and password is not provides (in XenCenter), then the machine account is not touched, just local leave is performed. 
if user name and password is provided (in XenCenter), we originally  delete the machine account, but we now just disable the machine account like PBIS

lippirk (approved):

liulinC:
@lippirk @lindig @psafont  Thanks for the help, 
I added to new commit, could you please help to review again?

ocaml/tests/test_data/repository_pkg_of_fullname_all
```diff
@@ -253,7 +253,6 @@ intel-igb-5.3.5.20-2.xs8.x86_64
 nss-pem-1.0.3-4.el7.x86_64
 gssproxy-0.7.0-17.el7.x86_64
 nss-sysinit-3.36.0-7.el7_5.x86_64
-pbis-open-8.2.3-1.7.8.xs8.x86_64
```
lippirk, line 256: why have these been removed?

lippirk:
Looks fine to me, once the `ocaml/tests/test_data/repository_pkg_of_fullname_all` changes have been reverted

Another note is that the 'Avoid using List.hd' commit contains some unrelated changes

lippirk (approved):

ocaml/tests/test_data/repository_pkg_of_fullname_all
liulinC: We are going to remove PBIS related packages, 
I know this is just for unittest, but remove it makes no bad.

liulinC (commented):
> Looks fine to me, once the `ocaml/tests/test_data/repository_pkg_of_fullname_all` changes have been reverted
> 
> Another note is that the 'Avoid using List.hd' commit contains some unrelated changes

Yes, Just merged them into this commit, to avoid too many commits.
Others just some minor refine.

robhoes (approved):

PR #4438, opened 2021-06-17
CA-355571: Include accumulative updates for updates description and guidances

minglumlu:
CA-355571: Include accumulative updates for updates description and guidances

When applying updates on a host, there are 2 parts would be considered:
1. Latest updates, which include binary RPM packages to be installed.
The 'yum list updates' will get these RPM packages;
2. Accumulative updates. Some of these updates may be overrided by the
latest updates. These can be get from 'yum updateinfo list updates'.

E.g. pkg-a-1.0.0-r1.x86_64 has been installed on a host. There is an
update pkg-a-1.0.1-r1.x86_64 which is included in update-1. But this
update-1 has not been applied by user. Now there is another update-2
with pkg-a-1.0.2-r1.x86_64. In this case, the latest update is update-2
(pkg-a-1.0.2-r1.x86_64), the accumulative updates are update-1 and
update-2.

With this commit, the metadata (summary, description) of all
accumulative updates are included in overall update information; and
guidances of all accumulative updates are evaluated as well.
But the RPM packages to be installed are still from latest updates.

CA-355571: Refine precedence between guidances

This commit aims to refine the precedence between guidances as:
RestartToolstack < RebootHost
EvacuateHost < RebootHost
RestartDeviceModel < RebootHost
RestartDeviceModel < EvacuateHost

This implies following as well:
EvacauteHost and RestartToolstack can be presented together;
RestartDeviceModel and RestartToolstack can be presented together.

### DIFF
--- a/ocaml/xapi/repository.ml
+++ b/ocaml/xapi/repository.ml
@@ -537,7 +537,6 @@ let restart_device_models ~__context ~host =
 let apply_immediate_guidances ~__context ~host ~guidances =
   (* This function runs on master host *)
   try
-    let num_of_hosts = List.length (Db.Host.get_all ~__context) in
     let open Client in
     Helpers.call_api_functions ~__context (fun rpc session_id ->
         let open Guidance in
@@ -561,16 +560,6 @@ let apply_immediate_guidances ~__context ~host ~guidances =
         | l when GuidanceSet.eq_set2 l ->
             (* RestartDeviceModel and RestartToolstack *)
             restart_device_models ~__context ~host ;
-            Client.Host.restart_agent ~rpc ~session_id ~host
-        | l when GuidanceSet.eq_set3 l ->
-            (* RestartDeviceModel and EvacuateHost *)
-            (* Evacuating host restarted device models already *)
-            if num_of_hosts = 1 then restart_device_models ~__context ~host ;
-            ()
-        | l when GuidanceSet.eq_set4 l ->
-            (* EvacuateHost, RestartToolstack and RestartDeviceModel *)
-            (* Evacuating host restarted device models already *)
-            if num_of_hosts = 1 then restart_device_models ~__context ~host ;
             Client.Host.restart_agent ~rpc ~session_id ~host
         | l ->
             let host' = Ref.string_of host in
--- a/ocaml/xapi/repository.ml
+++ b/ocaml/xapi/repository.ml
@@ -328,21 +328,31 @@ let create_pool_repository ~__context ~self =
 let get_host_updates_in_json ~__context ~installed =
   try
     with_local_repositories ~__context (fun repositories ->
-        (* (name_arch, updateID) list *)
-        let update_ids = get_updates_from_updateinfo ~__context repositories in
+        (* (pkg, update_id) list *)
+        let accumulative_updates =
+          get_updates_from_updateinfo ~__context repositories
+        in
         (* (name_arch, pkg) list *)
         let installed_pkgs =
           match installed with true -> get_installed_pkgs () | false -> []
         in
-        (* (name_arch, (pkg, repo)) list *)
-        let update_pkgs = get_updates_from_list_updates repositories in
+        (* (pkg, repo) list *)
+        let latest_updates = get_updates_from_list_updates repositories in
         List.iter (fun r -> clean_yum_cache r) repositories ;
-        let updates =
-          List.filter_map
-            (get_update_in_json ~update_ids ~installed_pkgs)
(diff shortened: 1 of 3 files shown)


ocaml/xapi/repository_helpers.ml
```diff
@@ -1235,7 +1208,90 @@ let get_updates_from_list_updates repositories =
...
+  let is_same_name_arch pkg1 pkg2 =
+    pkg1.name = pkg2.name && pkg1.arch = pkg2.arch
+  in
+  match
+    List.find_opt (fun (pkg', _) -> is_same_name_arch pkg pkg') latest_updates
+  with
+  | Some (pkg', repo) ->
+      if
+        Applicability.gt pkg.epoch pkg.version pkg.release pkg'.epoch
+          pkg'.version pkg'.release
```
lippirk, line 1240: Maybe `repository_helpers.ml` should have a function for comparing packages

ocaml/xapi/repository_helpers.ml
```diff
@@ -283,24 +280,34 @@ module GuidanceSet = struct
...
-    | None, Absolute ->
-        filter (fun g -> g <> EvacuateHost) gs
+
+  let precedences =
+    [
+      (RebootHost, [RestartToolstack; EvacuateHost; RestartDeviceModel])
+    ; (EvacuateHost, [RestartDeviceModel])
+    ]
+
+  let resort_guidances ~kind gs =
```
lippirk, line 300: I'm having a hard time understand this function - is it supposed to be equivalent to something like this? (I think it gets simpler if you use sets, so take `gs` to be a set below)

```
open GuidanceSet

let clean_guidances ~kind gs =
  let gs = 
    if mem RebootHost gs then
      gs |> remove RestartToolstack |> remove EvacuateHost |> remove RestartDeviceModel 
    else 
      gs
  in
  let gs = if mem EvauateHost gs then gs |> remove RestartDeviceModel else gs
  in match kind with
  | Recommended -> gs
  | Absolute -> remove EvacuateHost gs
```

lippirk: or use set difference instead of `remove .. |> remove .. |> remove ..`

ocaml/xapi/repository_helpers.ml
minglumlu: Indeed. I can move the comparison from Applicability to Pkg.

ocaml/xapi/repository_helpers.ml
minglumlu: Using `Set` is a good idea. How about this:
```
let precedences =
  [
    (RebootHost, GuidanceSet.of_list [RestartToolstack; EvacuateHost; RestartDeviceModel])
  ; (EvacuateHost, GuidanceSet.of_list [RestartDeviceModel])
  ]

let resort_guidances ~kind gs =
  let gs' =
    List.fold_left
      (fun acc (higher, lowers) ->
        if GuidanceSet.mem higher gs then GuidanceSet.diff gs lowers
        else gs)
      gs precedences
  in
  match kind with
  | Recommended -> gs'
  | Absolute -> GuidanceSet.remove EvacuateHost gs'
```

I want to keep the `precedences` as I think this makes the precedence relation easy to be understood and updated in future.

lippirk: this looks good to me!

lippirk (approved):

ocaml/xapi/repository.ml
```diff
@@ -602,23 +612,34 @@ let apply_updates ~__context ~host ~hash =
...
           http_get_host_updates_in_json ~__context ~host ~installed:true
+        in
+        let latest_updates =
+          host_updates
           |> Yojson.Basic.Util.member "updates"
           |> Yojson.Basic.Util.to_list
           |> List.map Update.of_json
         in
-        match all_updates with
+        match latest_updates with
```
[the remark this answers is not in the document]
minglumlu: Yes. But there is another case. `latest_updates` is used as the set of updates to be installed in `Client.Client.Repository.apply` actually. It is possible that `accumulative_updates` is empty, but `latest_updates` is not. (This means some packages are missing in updateinfo. But this will not break the actual updates since updateinfo is only information of the updates. In other words, we only try best to retrieve information from `accumulative_updates` )
So here `latest_updates` is the most accurate indicator to determine if a host could be updated.

robhoes (approved):

robhoes:
All looks fine. @minglumlu please squash the fix-ups before merging.

PR #4440, opened 2021-06-17
Cert Refresh

lindig:
Just for discussion. This is my approach to refresh server certificates. This rotation of the certs works both on the client and the server, including updating bundles and the database. However, I don't know yet how best to activate them without losing connections. Hence, the stunnel re-configuration is commented out.

### DIFF
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -1029,10 +1029,10 @@ let host_query_ha = call ~flags:[`Session]
       ~allowed_roles:_R_READ_ONLY
       ()
 
-  let refresh_server_certificates = call
+  let refresh_server_certificate = call
       ~lifecycle:[Published, rel_next, ""]
-      ~name:"refresh_server_certificates"
-      ~doc:"Replace the self-signed certficates for the host with new ones."
+      ~name:"refresh_server_certificate"
+      ~doc:"Replace the internal self-signed host certficate with a new one."
       ~params:[Ref _host, "host", "The host"]
       ~allowed_roles:_R_POOL_ADMIN
       ()
@@ -1552,7 +1552,7 @@ let host_query_ha = call ~flags:[`Session]
         crl_list;
         certificate_sync;
         get_server_certificate;
-        refresh_server_certificates;
+        refresh_server_certificate;
         install_server_certificate;
         emergency_reset_server_certificate;
         reset_server_certificate;
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -1029,6 +1029,14 @@ let host_query_ha = call ~flags:[`Session]
       ~allowed_roles:_R_READ_ONLY
       ()
 
+  let refresh_server_certificates = call
+      ~lifecycle:[Published, rel_next, ""]
+      ~name:"refresh_server_certificates"
+      ~doc:"Replace the self-signed certficates for the host with new ones."
+      ~params:[Ref _host, "host", "The host"]
+      ~allowed_roles:_R_POOL_ADMIN
+      ()
+
   let display =
     Enum ("host_display", [
         "enabled", "This host is outputting its console to a physical display device";
@@ -1544,6 +1552,7 @@ let host_query_ha = call ~flags:[`Session]
         crl_list;
         certificate_sync;
         get_server_certificate;
+        refresh_server_certificates;
         install_server_certificate;
         emergency_reset_server_certificate;
         reset_server_certificate;
--- a/ocaml/xapi/cert_distrib.ml
+++ b/ocaml/xapi/cert_distrib.ml
@@ -439,3 +439,17 @@ let import_joining_pool_ca_certificates ~__context ~ca_certs =
   Worker.local_write_cert_fs ~__context ApplianceCertificate Merge
     appliance_certs ;
   Worker.local_regen_bundle ~__context
+
+let distribute_new_host_cert ~__context ~host ~content =
+  let hosts = Db.Host.get_all ~__context in
+  let uuid = Db.Host.get_uuid ~__context ~self:host in
+  let file =
+    WireProtocol.{filename= Printf.sprintf "%s.new.pem" uuid; content}
+  in
+  let job rpc session_id host =
+    Worker.remote_write_certs_fs HostPoolCertificate Merge [file] host rpc
+      session_id
+  in
+  Helpers.call_api_functions ~__context @@ fun rpc session_id ->
+  List.iter (fun host -> job rpc session_id host) hosts ;
+  List.iter (fun host -> Worker.remote_regen_bundle host rpc session_id) hosts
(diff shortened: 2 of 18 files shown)


ocaml/xapi/xapi_host.ml
```diff
@@ -1465,7 +1465,11 @@ let replace_host_certificate ~__context ~type' ~host
   let old_certs = Db_util.get_host_certs ~__context ~type' ~host in
   let new_cert = write_cert_fs () in
```
psafont, line 1466: This overwrites the existing certificate. What happens when there's a failure and the previous certificates needs to be reinstated?

ocaml/idl/datamodel_host.ml
```diff
@@ -1029,6 +1029,14 @@ let host_query_ha = call ~flags:[`Session]
       ~allowed_roles:_R_READ_ONLY
       ()
 
+  let refresh_server_certificates = call
+      ~lifecycle:[Published, rel_next, ""]
+      ~name:"refresh_server_certificates"
+      ~doc:"Replace the self-signed certficates for the host with new ones."
+      ~params:[Ref _host, "host", "The host"]
+      ~allowed_roles:_R_POOL_ADMIN
```
psafont, line 1037: This should not be documented, IMO. Users will find confusing having "refresh_server_certificates" and "reset_server_certificate".
I think recycling certificates in this way only makes sense for internal certificates, dropping the 's' among other changes would make it less confusing even for us.

ocaml/xapi/xapi_host.ml
lindig: Maybe I should rename them to `*.bak` such that they would not be picked up by bundling but they would be around for recovery.

ocaml/idl/datamodel_host.ml
lindig: Ok - will limit this to the internal certificates. But to be clear: a server has two self-signed certificates, one for internal and one for external use. So don't rotate the external one? I don't think this makes sense from a security perspective.

psafont: The external one is already handled by the call `reset_server_certificate`. It's a different call because users might have installed their own certificates and they just want to replace the self-signed ones without affecting the external certificates.

ocaml/xapi/message_forwarding.ml
```diff
@@ -3546,6 +3546,25 @@ functor
             Client.Host.get_server_certificate rpc session_id host
         )
 
+      let refresh_server_certificate ~__context ~host =
```
lippirk, line 3549: We probably want a lock to prevent concurrent refreshes

lippirk: Maybe we want to error if a refresh is already in progress

ocaml/xapi/cert_distrib.ml
```diff
@@ -439,3 +439,17 @@ let import_joining_pool_ca_certificates ~__context ~ca_certs =
...
   Worker.local_regen_bundle ~__context
+
+let distribute_new_host_cert ~__context ~host ~content =
+  let hosts = Db.Host.get_all ~__context in
+  let uuid = Db.Host.get_uuid ~__context ~self:host in
+  let file =
+    WireProtocol.{filename= Printf.sprintf "%s.new.pem" uuid; content}
+  in
+  let job rpc session_id host =
+    Worker.remote_write_certs_fs HostPoolCertificate Merge [file] host rpc
```
lippirk, line 450: what happens if `file` already exists on the remote host? We probably don't want to overwrite it, because the remote host could be relying on it

ocaml/xapi/cert_refresh.ml
```diff
@@ -0,0 +1,102 @@
...
+  let ips = [ip] in
+  Gencertlib.Selfcert.host ~name ~dns_names ~ips path
+
+let host ~__context ~type' =
+  let host = Helpers.get_localhost ~__context in
+  let dbg = Context.string_of_task __context in
+  let pem = cert_path type' in
+  let path = new_cert_path type' in
+  let cert = new_host_cert ~dbg ~path in
+  let bak = backup_cert_path type' in
```
lippirk, line 59: We might want to think about the case where `bak` exists on the file system at this point - it probably means that a previous cert refresh failed. Do we just error out in this case or try to resolve the problem?

lippirk: Ah I see we don't actually remove `bak` at the end. The idea is that it is useful to keep around in case the user needs to manually intervene because of a failure?

I'm thinking about the case where a cert refresh has failed - the user's first instinct after running `xe cert-refresh` and seeing a failure is going to be to run it again, so it would be nice not to overwrite `bak` in this case

lindig: Good point! Maybe we should fail if the backup exists

ocaml/xapi/cert_refresh.ml
```diff
@@ -50,6 +50,12 @@ let new_host_cert ~dbg ~path : X509.Certificate.t =
   let ips = [ip] in
   Gencertlib.Selfcert.host ~name ~dns_names ~ips path
 
+let reload_stunnel () =
+  let systemctl cmd =
+    Helpers.call_script !Xapi_globs.systemctl [cmd; "stunnel@xapi"] |> ignore
+  in
+  systemctl "reload-or-restart"
+
```
lippirk, line 58: For consistency you could put this function in `Helpers.Stunnel`

We'll probably want to use it here too https://github.com/xapi-project/xen-api/blob/99d1f2d9c34265ebf0f78650960f3a4b5fb8fc87/ocaml/xapi/cert_distrib.ml#L229 (though that is out of scope for this PR)

ocaml/xapi/cert_distrib.mli
```diff
@@ -50,3 +50,7 @@ val import_joining_pool_ca_certificates :
     This parameter must be the result of
     [exchange_ca_certificates_with_joiner]. This function was designed
     as part of pool join and is unlikely to be useful elsewhere. *)
+
+val distribute_new_host_cert :
+  __context:Context.t -> host:[`host] API.Ref.t -> content:string -> unit
+(** distribute a new (additioanl) certificate for [host] in the pool *)
```
lippirk, line 56:
```suggestion
(** distribute a new (additional) certificate for [host] in the pool *)
```

lindig: Fixed it.

ocaml/xapi/cert_refresh.ml
lindig: Done

lindig:
This should not be merged before we have support for systemd-based reload of Stunnel merged.

ocaml/xapi/message_forwarding.ml
lippirk: your call to `with_pool_operation` should error if a refresh is already in progress

lippirk (approved):
We should use the cert distrib lock introduced here: https://github.com/xapi-project/xen-api/pull/4441/files . Whilst it seems like the lock induced by the cert refresh pool operation should be sufficient, it does not mutually exclude pool.join cert distribs and Host.refresh_server_certificate

However I'm happy to get this in and fix that later

psafont (approved):

PR #4443, opened 2021-06-24
REQ-403 concurrency fixes part 2

lippirk:
update_ca_bundle.sh is not thread safe because it manipulates files in
/etc/stunnel. We only allow one instance to be running at once.


psafont:
The lock for the bundle is going to be held sometimes while holding the lock in cert_distrib, since there is no wait on other threads after releasing the bundle lock there shouldn't be a lockout.

I don't think there's a possibility to get a deadlock either since the order of lock acquisition is always the cert_distrib one first then the bundle one, or just the bundle one

psafont (approved):

lindig (approved):

PR #4444, opened 2021-06-24
REQ-403 concurrency fixes part 3

lippirk:
This fixes parallel pool joins

The idea is that the joiners only exchanges certs with the primary host during Pool.join. The secondary hosts will get joiners' certificates because of `am_i_missing_certs_loop`

### DIFF
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -1467,6 +1467,19 @@ let host_query_ha = call ~flags:[`Session]
     ~allowed_roles:_R_POOL_ADMIN
     ()
 
+  let copy_primary_host_certs = call
+    ~name:"copy_primary_host_certs"
+    ~in_oss_since:None
+    ~in_product_since:rel_next
+    ~doc:"useful for secondary hosts that are missing some certs"
+    ~params:[
+      Ref _host, "host", "this host receives a copy of the primary host's trusted certificates";
+    ]
+    ~allowed_roles:_R_POOL_ADMIN
+    ~hide_from_docs:true
+    ~pool_internal:true
+    ()
+
   (** Hosts *)
   let t =
     create_obj ~in_db:true ~in_product_since:rel_rio ~in_oss_since:oss_since_303 ~internal_deprecated_since:None ~persist:PersistEverything ~gen_constructor_destructor:false ~name:_host ~descr:"A physical host" ~gen_events:true
@@ -1596,6 +1609,7 @@ let host_query_ha = call ~flags:[`Session]
         emergency_reenable_tls_verification;
         cert_distrib_atom;
         apply_updates;
+        copy_primary_host_certs;
       ]
       ~contents:
         ([ uid _host;
--- a/ocaml/xapi/cert_distrib.ml
+++ b/ocaml/xapi/cert_distrib.ml
@@ -481,6 +481,34 @@ let exchange_certificates_with_joiner ~__context ~uuid ~certificate =
   Worker.local_write_cert_fs ~__context HostPoolCertificate Merge
     [joiner_certificate] ;
   Worker.local_regen_bundle ~__context ;
+  let () =
+    (* now that the primary host trusts the joiner, perform best effort
+     * distribution to remaining hosts (this is not strictly necessary,
+     * as the 'am i missing certs thread' should pick up missing certs) *)
+    let secondary_hosts = Xapi_pool_helpers.get_slaves_list ~__context in
+    Helpers.call_api_functions ~__context @@ fun rpc session_id ->
+    secondary_hosts
+    |> List.iter (fun host ->
+           try
+             Worker.remote_write_certs_fs HostPoolCertificate Merge
+               [joiner_certificate] host rpc session_id
+           with e ->
+             D.warn
+               "exchange_certificates_with_joiner: sending joiner cert to %s \
+                failed. ex: %s"
+               (Ref.short_string_of host) (Printexc.to_string e)
+       ) ;
+
+    secondary_hosts
+    |> List.iter (fun host ->
+           try Worker.remote_regen_bundle host rpc session_id
+           with e ->
+             D.warn
+               "exchange_certificates_with_joiner: failed to regen bundle on \
+                %s. ex: %s"
+               (Ref.short_string_of host) (Printexc.to_string e)
+       )
+  in
   get_local_pool_certs () |> List.map WireProtocol.pair_of_certificate_file
 
 (* This function is called on the host that is joining a pool *)
--- a/ocaml/xapi/cert_distrib.ml
+++ b/ocaml/xapi/cert_distrib.ml
@@ -472,29 +472,16 @@ let copy_certs_to_host ~__context ~host =
     (get_local_ca_certs ()) host rpc session_id ;
   Worker.remote_regen_bundle host rpc session_id
 
-let import_joiner ~__context ~uuid ~certificate ~to_hosts =
-  let joiner_certificate =
-    HostPoolProvider.certificate_of_id_content uuid certificate
-  in
-  Helpers.call_api_functions ~__context @@ fun rpc session_id ->
-  List.iter
-    (fun host ->
-      Worker.remote_write_certs_fs HostPoolCertificate Merge
-        [joiner_certificate] host rpc session_id
-      )
-    to_hosts ;
-  List.iter
-    (fun host -> Worker.remote_regen_bundle host rpc session_id)
-    to_hosts
-
 (* This function is called on the pool that is incorporating a new host *)
 let exchange_certificates_with_joiner ~__context ~uuid ~certificate =
   lock @@ fun () ->
-  let all_hosts = Db.Host.get_all ~__context in
-  import_joiner ~__context ~uuid ~certificate ~to_hosts:all_hosts ;
-  Helpers.call_api_functions ~__context @@ fun rpc session_id ->
-  collect_pool_certs ~__context ~rpc ~session_id ~from_hosts:all_hosts
-    ~map:WireProtocol.pair_of_certificate_file
+  let joiner_certificate =
+    HostPoolProvider.certificate_of_id_content uuid certificate
+  in
+  Worker.local_write_cert_fs ~__context HostPoolCertificate Merge
+    [joiner_certificate] ;
+  Worker.local_regen_bundle ~__context ;
+  get_local_pool_certs () |> List.map WireProtocol.pair_of_certificate_file
 
 (* This function is called on the host that is joining a pool *)
 let import_joining_pool_certs ~__context ~pool_certs =
@@ -509,21 +496,10 @@ let collect_ca_certs ~__context ~names =
 (* This function is called on the pool that is incorporating a new host *)
 let exchange_ca_certificates_with_joiner ~__context ~import ~export =
   lock @@ fun () ->
-  let all_hosts = Db.Host.get_all ~__context in
   let appliance_certs = List.map WireProtocol.certificate_file_of_pair import in
-
-  Helpers.call_api_functions ~__context @@ fun rpc session_id ->
-  List.iter
-    (fun host ->
-      Worker.remote_write_certs_fs ApplianceCertificate Merge appliance_certs
-        host rpc session_id
-      )
-    all_hosts ;
-
-  List.iter
-    (fun host -> Worker.remote_regen_bundle host rpc session_id)
-    all_hosts ;
-
+  Worker.local_write_cert_fs ~__context ApplianceCertificate Merge
+    appliance_certs ;
+  Worker.local_regen_bundle ~__context ;
   collect_ca_certs ~__context ~names:export
 
 (* This function is called on the host that is joining a pool *)
--- a/ocaml/xapi/cert_distrib.ml
+++ b/ocaml/xapi/cert_distrib.ml
@@ -1,4 +1,5 @@
 module Unixext = Xapi_stdext_unix.Unixext
+module StringSet = Set.Make (String)
 
 module D = Debug.Make (struct let name = "cert_distrib" end)
 
@@ -103,8 +104,10 @@ module HostPoolProvider = struct
 
   let store_path = !Xapi_globs.trusted_pool_certs_dir
 
+  let cert_fname_of_host_uuid uuid = uuid ^ ".pem"
+
   let certificate_of_id_content uuid content =
-    WireProtocol.{filename= uuid ^ ".pem"; content}
+    WireProtocol.{filename= cert_fname_of_host_uuid uuid; content}
 
   let read_certificate uuid =
     let open Gencertlib.Pem in
@@ -117,6 +120,8 @@ module HostPoolProvider = struct
         certificate_of_id_content uuid host_cert
 end
 
+let string_of_file path = Unixext.read_lines path |> String.concat "\n"
+
 module ApplianceProvider = struct
   let store_path = !Xapi_globs.trusted_certs_dir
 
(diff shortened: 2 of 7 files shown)


psafont:
Having a delay of almost 10 seconds when distributing certificates might introduce failed connections in the pool when the join finished faster than that. I would rather _not_ remove certificate distribution on join and allow failures while having the delayed certificate distribution when that fails to reduce issues.

ocaml/idl/datamodel_host.ml
```diff
@@ -1459,6 +1459,18 @@ let host_query_ha = call ~flags:[`Session]
     ~allowed_roles:_R_POOL_ADMIN
     ()
 
+  let copy_primary_host_certs = call
+    ~name:"copy_primary_host_certs"
+    ~in_oss_since:None
+    ~in_product_since:rel_next
+    ~doc:"useful for secondary slaves that are missing some certs"
```
lindig, line 1466: I think this could be confusing as there are no primary slaves. It's secondary hosts or slaves.

ocaml/idl/datamodel_host.ml
```diff
@@ -1459,6 +1459,18 @@ let host_query_ha = call ~flags:[`Session]
...
+    ~params:[
+      Ref _host, "host", "the primary host will copy its set of trusted certs onto this host";
```
lindig, line 1468: The host updates its trusted certificates by receiving them from the primary host in the pool.

ocaml/xapi/cert_distrib.ml
```diff
@@ -399,29 +399,83 @@ let exchange_certificates_among_all_members ~__context =
...
 
-let import_joiner ~__context ~uuid ~certificate ~to_hosts =
-  let joiner_certificate =
-    HostPoolProvider.certificate_of_id_content uuid certificate
+let ( (get_local_ca_certs : unit -> WireProtocol.certificate_file list)
+    , (get_local_pool_certs : unit -> WireProtocol.certificate_file list) ) =
+  let g path () =
+    Sys.readdir path
+    |> Array.to_list
+    |> List.map (fun filename ->
```
lindig, line 407: Should we filter for `Filename.check_suffix` to only include `*.pem` but no `*.bak` or similar?

ocaml/xapi/cert_distrib.ml
```diff
@@ -399,29 +399,83 @@ let exchange_certificates_among_all_members ~__context =
...
+  in
+  missing_pool_certs () || missing_ca_certs ()
+
+let copy_certs_to_host ~__context ~host =
+  lock @@ fun () ->
+  D.debug "copy_certs_to_host: sending my certs to host %s"
+    (Ref.short_string_of host) ;
+  if am_i_missing_certs ~__context then
+    D.error
+      "i have been asked to copy my certs to %s but i myself am missing \
```
lindig, line 459: "I have been tasked to ..."?

ocaml/xapi/cert_distrib.mli
```diff
@@ -6,6 +6,12 @@ val exchange_certificates_among_all_members : __context:Context.t -> unit
     certificates from all members in a pool and installed on all of them. On
     success, new bundles will have been generated on the members. *)
 
+val am_i_missing_certs : __context:Context.t -> bool
```
lindig, line 9: Would it be more useful to return the list or set of missing certs?

lippirk: We could do that. At the moment the logic is roughly: if i am missing any certs then ask the primary host to send me all its certs

If we were to request a subset of certs then i guess this would be more efficient on the network, but with a possible increased risk of certs becoming stale (if we consider the primary host as the source of truth). The 'am i missing certs thread' should only very rarely actually copy certs from one host to another, so I doubt it would be a problem

ocaml/xapi/cert_distrib.ml
lippirk: I think you are right and we should take only `*.pem` (since `update-ca-bundle.sh` generates the bundle only using `*.pem` files, so anything else does not end up in the bundle)

lippirk:
> Having a delay of almost 10 seconds when distributing certificates might introduce failed connections in the pool when the join finished faster than that. I would rather _not_ remove certificate distribution on join and allow failures while having the delayed certificate distribution when that fails to reduce issues.

I've addressed this in 0887865

ocaml/xapi/cert_distrib.ml
```diff
@@ -117,17 +118,16 @@ module HostPoolProvider = struct
         certificate_of_id_content uuid host_cert
 end
 
+let string_of_file path = Unixext.read_lines path |> String.concat "\n"
```
lindig, line 121: Could simply be called "read"; the types are doing the test.

ocaml/xapi/cert_distrib.ml
```diff
@@ -399,29 +399,115 @@ let exchange_certificates_among_all_members ~__context =
...
+  let missing_pool_certs () =
+    let expected_files =
+      Db.Host.get_all ~__context
+      |> List.map (fun self -> Db.Host.get_uuid ~__context ~self)
+      |> List.map (fun x -> x ^ ".pem")
+      |> StringSet.of_list
+    in
+    let actual_files =
+      Sys.readdir HostPoolProvider.store_path
+      |> Array.to_seq
```
lindig, line 430: Don't we need to filter based on extension here, too?

ocaml/xapi/cert_distrib.ml
```diff
@@ -399,29 +399,115 @@ let exchange_certificates_among_all_members ~__context =
...
+             | _ ->
+                 None
+         )
+      |> List.map (fun x -> x ^ ".pem")
+      |> StringSet.of_list
+    in
+    let actual_files =
+      Sys.readdir ApplianceProvider.store_path
+      |> Array.to_seq
+      |> StringSet.of_seq
```
lindig, line 451: Same here?

lindig (approved):

ocaml/xapi/cert_distrib.ml
lippirk: I don't think it matters whether or not we remove the 'non .pems' from this set: the strings in `expected_files` all end in ".pem", so `expected_files` is a subset of `actual_files` if and only if it is a subset of `Set.union actual_files S`, where `S` is any set containing strings not ending in `.pem`

(same applies below)

ocaml/xapi/cert_distrib.ml
```diff
@@ -399,6 +399,73 @@ let exchange_certificates_among_all_members ~__context =
...
+  in
+  (g ApplianceProvider.store_path, g HostPoolProvider.store_path)
+
+let am_i_missing_certs ~__context : bool =
+  (* compare what's in the database with what's on my filesystem *)
+  let missing_pool_certs () =
+    let expected_files =
+      Db.Host.get_all ~__context
+      |> List.map (fun self -> Db.Host.get_uuid ~__context ~self)
+      |> List.map (fun x -> x ^ ".pem")
```
psafont, line 425: This is duplicated in the HostPoolProvider. It's probably a good idea to have this function enclosed there to remove duplication, even if it's longer

ocaml/xapi/cert_distrib.ml
```diff
@@ -399,6 +399,73 @@ let exchange_certificates_among_all_members ~__context =
...
+    let expected_files =
+      Db.Certificate.get_all ~__context
+      |> List.filter_map (fun self ->
+             match Db.Certificate.get_type ~__context ~self with
+             | `ca ->
+                 Some (Db.Certificate.get_name ~__context ~self)
+             | _ ->
+                 None
+         )
+      |> List.map (fun x -> x ^ ".pem")
```
psafont, line 445: I think this is wrong. CA cert records have .pem in their names. This would have been avoided by adding to the providers the transformation from id to filename. this is now done as part of the certificate_of_id_content.

ocaml/idl/datamodel_host.ml
edwintorok: If we want to avoid the master/slave terminology then we should use just "secondary host" here, and considering you use 'primary' elsewhere it'd make more sense to drop the 'slave' terminology.

ocaml/xapi/cert_distrib.ml
```diff
@@ -399,6 +399,73 @@ let exchange_certificates_among_all_members ~__context =
...
+      |> List.map (fun self -> Db.Host.get_uuid ~__context ~self)
+      |> List.map (fun x -> x ^ ".pem")
+      |> StringSet.of_list
+    in
+    let actual_files =
+      Sys.readdir HostPoolProvider.store_path
+      |> Array.to_seq
+      |> StringSet.of_seq
+    in
+    StringSet.subset expected_files actual_files |> not
```
edwintorok, line 433: Might be useful to print the missing certs in a debug message, that can be addressed in a future PR

ocaml/xapi/cert_distrib.ml
```diff
@@ -399,6 +399,73 @@ let exchange_certificates_among_all_members ~__context =
...
+             match Db.Certificate.get_type ~__context ~self with
+             | `ca ->
+                 Some (Db.Certificate.get_name ~__context ~self)
+             | _ ->
+                 None
+         )
+      |> List.map (fun x -> x ^ ".pem")
+      |> StringSet.of_list
+    in
+    let actual_files =
```
edwintorok, line 448: Can this and the above functions be factored out into something common?

ocaml/xapi/xapi.ml
```diff
@@ -844,6 +844,29 @@ let init_tls_verification () =
...
+        D.debug "am_i_missing_certs_loop: i am missing certs!" ;
+        let host = get_localhost ~__context in
+        call_api_functions ~__context @@ fun rpc session_id ->
+        Client.Host.copy_primary_host_certs ~rpc ~session_id ~host
+      )
+    with e ->
+      D.error "am_i_missing_certs_loop: exception (ignoring and continuing): %s"
+        (Printexc.to_string e)
+  in
+  let rec loop () = inner () ; Thread.delay delay ; (loop [@tailcall]) () in
```
edwintorok, line 867: Why does there have to be a delay here, and if there has to be can you make it an exponential backoff that starts with a small delay (0.5s and ends up 10s at max?)
There is a Repeat_with_uniform_backoff in helpers.ml, could use that as a start and maybe add an exponential backoff policy later (I think there are exponential backoffs elsewhere in xapi already)

ocaml/xapi/xapi_host.ml
```diff
@@ -2652,6 +2652,8 @@ let alert_if_tls_verification_was_emergency_disabled ~__context =
 let cert_distrib_atom ~__context ~host ~command =
   Cert_distrib.local_exec ~__context ~command
 
+let copy_primary_host_certs ~__context ~host = ()
```
psafont, line 2655: Can we add a comment this is added to appease the type system?

ocaml/xapi/cert_distrib.ml
```diff
@@ -479,6 +479,34 @@ let exchange_certificates_with_joiner ~__context ~uuid ~certificate =
   Worker.local_write_cert_fs ~__context HostPoolCertificate Merge
     [joiner_certificate] ;
   Worker.local_regen_bundle ~__context ;
+  let () =
+    (* now that the primary host trusts the joiner, perform best effort
+     * distribution to remaining hosts (this is not strictly necessary,
+     * as the 'am i missing certs thread' should pick up missing certs) *)
+    let secondary_hosts = Xapi_pool_helpers.get_slaves_list ~__context in
```
edwintorok, line 486: There is an API to sync config files (currently `/etc/shadow`), could that be reused for syncing the certificates?

ocaml/idl/datamodel_host.ml
```diff
@@ -1459,6 +1459,18 @@ let host_query_ha = call ~flags:[`Session]
...
+  let copy_primary_host_certs = call
+    ~name:"copy_primary_host_certs"
+    ~in_oss_since:None
+    ~in_product_since:rel_next
+    ~doc:"useful for secondary hosts that are missing some certs"
+    ~params:[
+      Ref _host, "host", "this host receives a copy of the primary host's trusted certificates";
+    ]
+    ~allowed_roles:_R_POOL_ADMIN
+    ~hide_from_docs:true
```
robhoes, line 1471: If this call is only to be used within the pool, then you can add `~pool_internal:true`. This show up in the `~intra_pool_only` argument to `Session_check.check`, which is used by the API server to authenticate incoming calls.

ocaml/xapi/xapi_host.ml
robhoes: Or move the actual business logic (`Cert_distrib.copy_certs_to_host ~__context ~host`) here, which would be the usual place (leaving message_forwarding for matter related to handling pool operations and forwarding).

lippirk: it's important that the master is the host coordinating the cert copy, because otherwise the cert distrib lock would not be taken. so the options are:

 - slave forwards copy request to master via message forwarding (as currently implemented)
 - we perform the forwarding inside cert_distrib.ml (is this what you were suggesting implicitly @robhoes  ...?)

ocaml/xapi/cert_distrib.ml
lippirk: good idea (i'll address it here, it's not too difficult)

ocaml/xapi/xapi_host.ml
robhoes: No, I mean that still `copy_primary_host_certs` is called through the API, with the implementation right here in xapi_host.ml. To make sure that the function runs on the master only, message_forwarding would only have a call to `Local.Host. copy_primary_host_certs`, no forwarding to a slave with `Client.Host.copy_primary_host_certs` through `do_op_on`. Slaves call the function only through `Client.Host.copy_primary_host_certs` and `Helpers.call_api_functions`, which go to the master.

lippirk: yes this is the same as what I have done, modulo some copy pasting (which I will do now)

ocaml/xapi/cert_distrib.ml
lippirk: in theory? (though my opinion is that at this point reimplementing the code we already have for distributing certificates to use `config_file_sync.ml` is creating unnecessary work for ourselves)

ocaml/xapi/xapi.ml
lippirk: I'm not sure I understand what advantage an expontential backoff would give us here? The idea of this loop is that if a slave's certificates become out of sync with what's the db, then it will ask the master for an update. In theory this we should only have to do this very rarely (it is an unfortunate solution for the cases of parallel pool join, or if certificate sync went wrong for some reason)

> Why does there have to be a delay here

i put the delay in so we are not querying the database and checking the file system in a tight loop. 10s was chosen arbitrarily

lippirk:
I ran tests again with the latest changes and they are still looking green

ocaml/xapi/xapi.ml
```diff
@@ -295,7 +295,30 @@ let on_master_restart ~__context =
     debug
       "failed to set Host_metrics.live to true immediately; will have to wait \
        for regular heartbeat to arrive: %s"
-      (ExnHelper.string_of_exn e)
+      (ExnHelper.string_of_exn e) ;
+    let am_i_missing_certs () =
```
robhoes, line 299: Did you mean for this to be inside the exception handler?

lippirk: nope :)

lippirk: thanks, now I realise why ocamlformat didn't like my indentation

psafont (approved):

lippirk:
Reran the tests - still looking green

PR #4446, opened 2021-06-25
CA-355964: Hostname change will affect the authentication

liulinC:
Winbind just use the hostname as the netbios_name to talk to DC
if hostname is changed after join domain, the auth will failed
This commit fix the issue by

* Store the netbios_name in xapi db during join domain
* Configure netbios_name to winbind daemon during xapi bootup
* Migrate netbios_name from PBIS db if domain joined but xapi
database does not have the value, to support update/upgrade

### DIFF
--- a/ocaml/tests/test_extauth_plugin_ADwinbind.ml
+++ b/ocaml/tests/test_extauth_plugin_ADwinbind.ml
@@ -84,6 +84,7 @@ module ParseValueFromPbis = Generic.MakeStateless (struct
       [
         ( "X'58005200540055004B002D00300032002D003000330024000000'"
         , "XRTUK-02-03" )
+      ; ("X'4C004F00430041004C0048004F0053005400320024000000'", "LOCALHOST2")
       ]
 end)
 
--- a/ocaml/tests/test_extauth_plugin_ADwinbind.ml
+++ b/ocaml/tests/test_extauth_plugin_ADwinbind.ml
@@ -41,6 +41,50 @@ module ExtractOuConfig = Generic.MakeStateless (struct
       ]
 end)
 
+module Range = Generic.MakeStateless (struct
+  module Io = struct
+    type input_t = int * int * int
+
+    type output_t = int list
+
+    let string_of_input_t (x, y, z) = Printf.sprintf "%d , %d, %d" x y z
+
+    let string_of_output_t = Test_printers.(list int)
+  end
+
+  let transform (s, e, step) = Extauth_plugin_ADwinbind.range s e step
+
+  let tests =
+    `QuickAndAutoDocumented
+      [
+        ((0, 10, 1), [0; 1; 2; 3; 4; 5; 6; 7; 8; 9])
+      ; ((0, 1, 1), [0])
+      ; ((0, 0, 1), [])
+      ; ((1, 0, 1), [])
+      ]
+end)
+
+module ParseValueFromPbis = Generic.MakeStateless (struct
+  module Io = struct
+    type input_t = string
+
+    type output_t = string
+
+    let string_of_input_t = Test_printers.(string)
+
+    let string_of_output_t = Test_printers.(string)
+  end
+
+  let transform s = Extauth_plugin_ADwinbind.parse_value_from_pbis s
+
+  let tests =
+    `QuickAndAutoDocumented
+      [
+        ( "X'58005200540055004B002D00300032002D003000330024000000'"
+        , "XRTUK-02-03" )
+      ]
+end)
+
 let test_domainify_uname =
   let open Extauth_plugin_ADwinbind in
   let check uname exp () =
@@ -349,6 +393,8 @@ let test_wbinfo_exception_of_stderr =
 let tests =
   [
     ("ADwinbind:extract_ou_config", ExtractOuConfig.tests)
+  ; ("ADwinbind:test_range", Range.tests)
+  ; ("ADwinbind:test_parse_value_from_pbis", ParseValueFromPbis.tests)
   ; ("ADwinbind:test_domainify_uname", test_domainify_uname)
   ; ("ADwinbind:test_parse_wbinfo_uid_info", test_parse_wbinfo_uid_info)
   ; ("ADwinbind:test_parse_ldap_stdout", test_parse_ldap_stdout)
--- a/ocaml/tests/test_extauth_plugin_ADwinbind.ml
+++ b/ocaml/tests/test_extauth_plugin_ADwinbind.ml
@@ -52,7 +52,8 @@ module Range = Generic.MakeStateless (struct
     let string_of_output_t = Test_printers.(list int)
   end
 
-  let transform (s, e, step) = Extauth_plugin_ADwinbind.range s e step
+  let transform (s, e, step) =
+    Extauth_plugin_ADwinbind.Migrate_from_pbis.range s e step
 
   let tests =
     `QuickAndAutoDocumented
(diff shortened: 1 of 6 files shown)


ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -58,8 +58,12 @@ let wb_cmd = !Xapi_globs.wb_cmd
 
 let tdb_tool = !Xapi_globs.tdb_tool
 
+let sqlite3 = !Xapi_globs.sqlite3
```
robhoes, line 61: This is unused.

ocaml/tests/test_extauth_plugin_ADwinbind.ml
```diff
@@ -41,6 +41,50 @@ module ExtractOuConfig = Generic.MakeStateless (struct
       ]
 end)
 
+module Range = Generic.MakeStateless (struct
+  module Io = struct
+    type input_t = int * int * int
```
lindig, line 46: I assume the names are dictated by the functor and can't be changed. I find them a bit verbose.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -58,8 +58,12 @@ let wb_cmd = !Xapi_globs.wb_cmd
...
+
 let debug_level = !Xapi_globs.winbind_debug_level |> string_of_int
 
+let pbis_db_path = "/var/lib/pbis/db/registry.db"
```
robhoes, line 65: This should go into xapi_globs

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -547,6 +560,80 @@ let domainify_uname ~domain uname =
   else
     Printf.sprintf "%s@%s" uname domain
 
+let range s e step =
```
lindig, line 563: All of this looks brittle to me. What are we trying to do here? We have an SQL database but seem to use low-level string scanning. What is encoded in this string that we are reading here?

liulinC: The `range` just like python `range`,  like `range 0 10 1 -> [0;1;2;3;4;5;6;7;8;9]`, see the unittest,

I suppose the question is about `parse_value_from_pbis`

This function parsing the values extracted from PBIS database (sqlite3), and decode it, see the unittest for the input and output.
The PBIS databases values are formatted as `X"xxxxxxxx"`,  where "xxxxxxx" are values in hex. 
All the data are formatted this way, like 
```
sqlite3 registry.db "select QUOTE(Value) from regvalues1 where ValueName='SamAccountName'"
X'4C004F00430041004C0048004F0053005400320024000000' 
```
For the parsing details, refer to the code, 

This migrate should only happens after update/upgrade from CH which joined domain with PBIS.

liulinC:
One possible refine:

Store netbios_name migrated from PBIS in xapi db, to skip re-migrate on next xapi restart, (Just for case, update/upgrade from CH joined domain with PBIS)

ocaml/xapi/extauth_plugin_ADwinbind.ml
lindig: This should be based around a function that takes a hex-string and returns the decoded string (a byte or char sequence). An easy way to do this is using `Hex.of_string` from the `hex` package.

liulinC: `Hex.of_string` is used finally , refer to the code, 
but before that, there are some preparing work like strip the tailing and heading, remvoe the splitting and ending `00` , etc,

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -547,6 +560,80 @@ let domainify_uname ~domain uname =
...
+      [pbis_db_path; sql]
+  in
+  if String.length value <= min_valid_pbis_value_length then
+    raise (Auth_service_error (E_GENERIC, Printf.sprintf "No value for %s" key))
+  else
+    value
+
+let remove_tailing s len = String.sub s 0 (String.length s - len)
+
+let parse_value_from_pbis raw_value =
```
lippirk, line 585: I would expect regexes to make this a little simpler? Assuming you have a function `extract_group` for obtaining the 1st regex group match, I would expect something like this to work

```
raw_value
|> extract_group {|X'([^']*)'|}
|> ... 
|> s -> Hex.to_string (`Hex s) 
|> extract_group {|(.*)\$|}
```

lippirk:
I think we would really benefit by partitioning the code used for the PBIS migration from the rest of the winbind code (perhaps using a `PBIS_migrate` module within this file)

ocaml/xapi/extauth_plugin_ADwinbind.ml
lindig: Or `Scanf.sscanf "X'abcd'" "X'%s@'" (fun s -> s)` -> `abcd`

ocaml/tests/test_extauth_plugin_ADwinbind.ml
liulinC: I think so, lots of other test test use it, so let's just keep to for consistency.

liulinC:
@robhoes @lippirk @lindig Thanks for the comments, 
could you please help to review again?

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -693,6 +695,31 @@ module Winbind = struct
...
+    let upper_char_start = 65 in
+    let upper_char_len = 26 in
+    let rec aux n acc =
+      if n > 0 then
+        aux (n - 1)
+          ((upper_char_start + Random.int upper_char_len |> char_of_int) :: acc)
+      else
+        acc
+    in
+    aux len [] |> List.to_seq |> String.of_seq
```
lippirk, line 711: `String.init len (fun _ -> upper_char_start + Random.int upper_char_len |> char_of_int)`

should do the same thing and is a bit simpler

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -564,18 +642,38 @@ module Winbind = struct
   let stop ~timeout ~wait_until_success =
     Xapi_systemctl.stop ~timeout ~wait_until_success name
 
-  let configure () =
+  let migrate_netbios_name ~__context =
```
lippirk, line 645: this function should be in the `Migrate_from_pbis` module?

lippirk (approved):

lippirk (approved):

liulinC:
@lippirk @robhoes @lindig @psafont 
Thanks for the help to review this PR, 
As this PR pending for a while, new commits for different issues comes in, 
you can review them commit by commits if you like, 
Again, thanks for the help.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -547,6 +560,80 @@ let domainify_uname ~domain uname =
...
+  let value =
+    Helpers.call_script ~log_output:On_failure !Xapi_globs.sqlite3
+      [pbis_db_path; sql]
+  in
+  if String.length value <= min_valid_pbis_value_length then
+    raise (Auth_service_error (E_GENERIC, Printf.sprintf "No value for %s" key))
+  else
+    value
+
+let remove_tailing s len = String.sub s 0 (String.length s - len)
```
psafont: This may encounter problems if len is bigger than the length of s. I propose using 
```ocaml
Astring.String.drop ~rev:true ~max:len s
```

psafont (approved):
I'm not an expert in this domain, the code has some parsing based on regex which may not be ideal but it all seems reasonable to me. The code is partially tested so it should be reasonably safe to merge

PR #4447, opened 2021-06-25
Maintenance: fix unixpwd warnings about loosing const qualifier

edwintorok:
Most of the warnings like below can be fixed by changing type from
`char*` to `const char*`.
```
unixpwd_stubs.c:77:10: warning: assignment discards 'const' qualifier from pointer target type [enabled by default]
     user = String_val(caml_user);
          ^
```

Except for the password, where the `putpwent` API takes a `char*`
instead of `const char*`.
We can't trust that it doesn't internally modify it, so take a copy and
pass that instead, use `caml_stat_strdup`+`caml_stat_free` for this
which automatically raise OCaml exceptions on NULL.

### DIFF
--- a/unixpwd/c/unixpwd_stubs.c
+++ b/unixpwd/c/unixpwd_stubs.c
@@ -28,7 +28,7 @@ CAMLprim        value
 caml_unixpwd_getpwd(value caml_user)
 {
     CAMLparam1(caml_user);
-    char           *user;
+    const char           *user;
     char           *passwd;
     CAMLlocal1(pw);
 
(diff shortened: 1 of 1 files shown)


lippirk (approved):
This looks correct when comparing to the `stub_gethostbyname` example here https://ocaml.org/manual/intfc.html (minus the runtime releasing/acquiring)

lindig:
Run `make indent` in `unixpwd/c` to fix the indentation.

lindig (approved):

robhoes (approved):

edwintorok:
https://github.com/xapi-project/xen-api/pull/4450 for the indent fix

PR #4448, opened 2021-06-25
CA-356017: fix migration failures

edwintorok:
CrossPool migrations were failing:

```
Jun 25 10:15:54 xrtuk-13-01 xapi: [error||3390 ||backtrace] Raised Db_exn.DBCache_NotFound("missing row", "VM", "OpaqueRef:24fa7464-93f7-475f-9389-b66955c430ec")
Jun 25 10:15:54 xrtuk-13-01 xapi: [error||3390 ||backtrace] 1/11 xapi Raised at file ocaml/database/db_cache_impl.ml, line 52
Jun 25 10:15:54 xrtuk-13-01 xapi: [error||3390 ||backtrace] 2/11 xapi Called from file ocaml/database/db_cache_impl.ml, line 57
Jun 25 10:15:54 xrtuk-13-01 xapi: [error||3390 ||backtrace] 3/11 xapi Called from file ocaml/xapi/db_actions.ml, line 4939
Jun 25 10:15:54 xrtuk-13-01 xapi: [error||3390 ||backtrace] 4/11 xapi Called from file ocaml/xapi/message_forwarding.ml, line 2453
Jun 25 10:15:54 xrtuk-13-01 xapi: [error||3390 ||backtrace] 5/11 xapi Called from file ocaml/xapi/rbac.ml, line 225
Jun 25 10:15:54 xrtuk-13-01 xapi: [error||3390 ||backtrace] 6/11 xapi Called from file ocaml/xapi/rbac.ml, line 234
Jun 25 10:15:54 xrtuk-13-01 xapi: [error||3390 ||backtrace] 7/11 xapi Called from file ocaml/xapi/server_helpers.ml, line 100
Jun 25 10:15:54 xrtuk-13-01 xapi: [error||3390 ||backtrace] 8/11 xapi Called from file ocaml/xapi/server_helpers.ml, line 118
Jun 25 10:15:54 xrtuk-13-01 xapi: [error||3390 ||backtrace] 9/11 xapi Called from file lib/xapi-stdext-pervasives/pervasiveext.ml, line 24
Jun 25 10:15:54 xrtuk-13-01 xapi: [error||3390 ||backtrace] 10/11 xapi Called from file lib/xapi-stdext-pervasives/pervasiveext.ml, line 38
Jun 25 10:15:54 xrtuk-13-01 xapi: [error||3390 ||backtrace] 11/11 xapi Called from file lib/debug.ml, line 230
```

After the migration finishes the VM will be truly gone, so we can't
[...]

PR #4449, opened 2021-06-25
Only add XAPI message for VM when migration is live and intrapool

benjamreis:
If it's not the xapi creation will fail as the VM might be in another pool


benjamreis:
See https://github.com/xapi-project/xen-api/pull/4448

edwintorok (approved):

psafont (approved):

robhoes (approved):

PR #4451, opened 2021-06-25
CA-355625 reload Stunnel instead of restart after cert change

lindig:
When changing a host's certificate, it is enough to reload Stunnel
rather than restarting it. This keeps existing connections alive. This
requires support in the systemd configuration of stunnel.

### DIFF
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -1474,7 +1474,7 @@ let replace_host_certificate ~__context ~type' ~host
   List.iter (Db_util.remove_cert_by_ref ~__context) old_certs ;
   let task = Context.get_task_id __context in
   Db.Task.set_progress ~__context ~self:task ~value:1.0 ;
-  Xapi_mgmt_iface.reconfigure_stunnel ~__context
+  Helpers.Stunnel.reload ()
 
 let install_server_certificate ~__context ~host ~certificate ~private_key
     ~certificate_chain =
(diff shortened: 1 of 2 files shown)


lippirk:
We may want to apply similar treatment here `ocaml/xapi/cert_distrib.ml:restart_stunnel`

ocaml/xapi/xapi_host.ml
```diff
@@ -1474,7 +1474,7 @@ let replace_host_certificate ~__context ~type' ~host
   List.iter (Db_util.remove_cert_by_ref ~__context) old_certs ;
   let task = Context.get_task_id __context in
   Db.Task.set_progress ~__context ~self:task ~value:1.0 ;
-  Xapi_mgmt_iface.reconfigure_stunnel ~__context
```
psafont, line 1477: This allows to delete the function reconfigure_stunnel, the other user of the function is never called (restart_stunnel in xapi/cert_distrib.ml)

lindig: The existing code for Stunnel reconfiguration is quite complex. Is all we ever need this simple reload?

lippirk (approved):

ocaml/xapi/xapi_host.ml
psafont: If it's for the stunnel server only: yes

psafont (approved):

PR #4452, opened 2021-06-28
CA-341715: control-domain-params-init: skip on upgrade

robhoes:
This makes it behave the same as network-init and other *-init services,
as well as the old firstboot scripts.


lippirk (approved):

lindig:
This looks good to me. But would it be more readable by avoiding the nesting and use this?

```
if [ condition -a condition ]; then

fi
```

lindig (approved):

robhoes:
@lindig Then you'd waste the `xe vm-param-get` call in the upgrade case?

PR #4455, opened 2021-07-06
CA-356325: Could not ssh to a host with a pooladmin user with upper case

liulinC:
User is permitted to ssh to dom0 by adding their name pam configure files
However, ssh expected the name in lower case other than the real name

This commit fix the issue by lookup name in subject details, which contain
the expected name

### DIFF
--- a/scripts/plugins/extauth-hook-AD.py
+++ b/scripts/plugins/extauth-hook-AD.py
@@ -176,18 +176,18 @@ class DynamicPam(ADConfig):
         if len(comps) < 2:
             logger.error("subject %s from sid %s is not valid", subject, sid)
             return None, None
-        # CONNAPP\test_user 1  # 1: user, 2: group
-        is_group = True if comps[1].strip() =="2" else False
-        _name = comps[0].strip()
-        # The _name returned name contains uppercase, need to format it from subject details
-        return self._query_subject_name(_name, is_group), is_group
+        # CONNAPP\IUgroup 1  # 1: user, 2: group
+        is_group = comps[1].strip() == "2"
+        real_name = comps[0].strip()
+        # The real_name returned name contains uppercase, need to format it from subject details
+        return self._query_subject_name(real_name, is_group), is_group
 
-    def _query_subject_name(self, _name, is_group):
+    def _query_subject_name(self, real_name, is_group):
         sub_cmd = "--group-info" if is_group else "--user-info"
-        cmd = ["/usr/bin/wbinfo", sub_cmd, _name]
+        cmd = ["/usr/bin/wbinfo", sub_cmd, real_name]
         subject_detail = run_cmd(cmd)
         if not subject_detail:
-           logger.error("Failed to lookup subject details %s", _name)
+           logger.error("Failed to lookup subject details %s", real_name)
            return None
         # CONNAPP\ugroup:x:3000009: --> group details
         # CONNAPP\iugroup:*:3000016:3000000:IUgroup:/home/CONNAPP/iugroup:/bin/bash --> user details
--- a/scripts/plugins/extauth-hook-AD.py
+++ b/scripts/plugins/extauth-hook-AD.py
@@ -177,8 +177,21 @@ class DynamicPam(ADConfig):
             logger.error("subject %s from sid %s is not valid", subject, sid)
             return None, None
         # CONNAPP\test_user 1  # 1: user, 2: group
-        name, category = comps[0].strip(), comps[1].strip()
-        return name, True if category == "2" else False
+        is_group = True if comps[1].strip() =="2" else False
+        _name = comps[0].strip()
+        # The _name returned name contains uppercase, need to format it from subject details
+        return self._query_subject_name(_name, is_group), is_group
+
+    def _query_subject_name(self, _name, is_group):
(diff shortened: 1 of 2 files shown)


scripts/plugins/extauth-hook-AD.py
```diff
@@ -177,8 +177,21 @@ def _query_subject(self, sid):
             logger.error("subject %s from sid %s is not valid", subject, sid)
             return None, None
         # CONNAPP\test_user 1  # 1: user, 2: group
-        name, category = comps[0].strip(), comps[1].strip()
-        return name, True if category == "2" else False
+        is_group = True if comps[1].strip() =="2" else False
```
lindig, line 180: I assume this can be simplified to `is_group = comps[1].strip() == "2"` (also added a space after `==`)

scripts/plugins/extauth-hook-AD.py
```diff
@@ -177,8 +177,21 @@ def _query_subject(self, sid):
...
+        _name = comps[0].strip()
```
lindig, line 181: Why is this prefixed with an underscore?

scripts/plugins/test_extauth_hook_AD.py
```diff
@@ -144,9 +144,18 @@ def test_not_permit_non_admin_group(self, mock_rename, mock_chmod):
 
     @patch("extauth_hook_ad.run_cmd")
     def test_query_subject_success(self, mock_cmd, mock_rename, mock_chmod):
+        def cmd_result(cmd):
```
lindig, line 147: I find the name confusing. Is this not simulating the execution of a command? Something like `mock` or `mock_exec` would tell me more.

lindig (approved):

lippirk:
It seems that here we are re-implementing what we already have in the ocaml

Does the subject record not already contain `subject-name` and `subject-is-group`?

scripts/plugins/extauth-hook-AD.py
liulinC: ah, yes, my bad

scripts/plugins/extauth-hook-AD.py
liulinC: it is name, but not the name wanted, 
let me try to find a `name` for it.

liulinC:
> It seems that here we are re-implementing what we already have in the ocaml
> 
> Does the subject record not already contain `subject-name` and `subject-is-group`?

Actually, the initial version just query such information from xapi sdk, but then we got a bug.
The root cause is, user can call SDK to add a subject only with SID, then leave `subject-name` and `subject-is-group` empty, 
details, see https://github.com/xapi-project/xen-api/pull/4446/commits/d252253257e3e12f9cb350147913494fff5a44c9

Also, I find PBIS query such I info plugin by running pbis command.

lippirk:
> The root cause is, user can call SDK to add a subject only with SID, then leave `subject-name` and `subject-is-group` empty,
> details, see [d252253](https://github.com/xapi-project/xen-api/commit/d252253257e3e12f9cb350147913494fff5a44c9)

should we fix that by having xapi fill in the extra information? Then we can rely on the database containing necessary information

liulinC:
current design is , during add subject, xapi just add the record into database, and there is an interval task to update the subject information,  I think this can benifit the performance and do not want to break this design.

scripts/plugins/extauth-hook-AD.py
```diff
@@ -176,9 +176,22 @@ def _query_subject(self, sid):
...
+    def _query_subject_name(self, real_name, is_group):
+        sub_cmd = "--group-info" if is_group else "--user-info"
+        cmd = ["/usr/bin/wbinfo", sub_cmd, real_name]
+        subject_detail = run_cmd(cmd)
+        if not subject_detail:
+           logger.error("Failed to lookup subject details %s", real_name)
+           return None
+        # CONNAPP\ugroup:x:3000009: --> group details
+        # CONNAPP\iugroup:*:3000016:3000000:IUgroup:/home/CONNAPP/iugroup:/bin/bash --> user details
+        return subject_detail.split(":")[0]
```
edwintorok, line 194: Will this also fix the case when we have a `#` in the group name?
That is a bug in the current PBIS code that groups starting with `#` (like many of our internal groups) do not work, because `#` is a comment character in PAM and cannot be escaped.
Are the names returned here guaranteed to be compatible with PAM and not contain characters like `#`?
Might be useful to add a testcase specifically for that.

liulinC: @edwintorok  I googled a while about how to escape a `#` for pam configuration (and quick go through the pam code, ),  but not found, 
is it a limitation of pam configuration?  any suggestion about this?

We probably can enum all the users in the group, but That sounds not a good idea.

edwintorok: Looks like we would need to patch PAM to fix this. Lets consider it out of scope for this PR for now

edwintorok (approved):

liulinC:
Thanks all or the help, 
I would like to follow the suggestion to let xapi manage the xapi database, and make it trust worthy, 

Plugin just query such info from xapi and all its responsibility is to manage the configuration file. 

So I decline this PR and followup in PR
https://github.com/xapi-project/xen-api/pull/4456

PR #4456, opened 2021-07-06
Private/linl/ca 356007

liulinC:
CA-356007: Failed to add subject if subject Sid changed recently in DC
    
This commit fix the issue by querying sid from ldap instead of winbind
as winbind use cache which may be outdated


ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -83,6 +83,28 @@ module Ldap = struct
   let string_of_user x =
     Rpcmarshal.marshal user.Rpc.Types.ty x |> Jsonrpc.to_string
 
+  let sid_of_name name =
```
lindig, line 86: A better name could by `sid_from_ldap`

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -629,17 +621,13 @@ let clean_local_resources () : unit =
...
-  if
-    is_infix ~affix:domain uname
-    || is_infix ~affix:"@" uname
-    || is_infix ~affix:{|\|} uname
-    || uname = krbtgt
-  then
-    uname
-  else
-    Printf.sprintf "%s@%s" uname domain
+let undomainify_uname uname =
```
lindig, line 624: Would it be easier to extract the desired substring using a regular expression and do this with a single operation? I don't like the piecemeal destruction of strings because it tends to be unreliable.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -699,7 +692,8 @@ module Winbind = struct
   let random_string len =
     let upper_char_start = 65 in
     let upper_char_len = 26 in
-    String.init len (fun _ -> upper_char_start + Random.int upper_char_len |> char_of_int)
+    String.init len (fun _ ->
```
lindig, line 695: I would like it better if you implemented a `rand_char: unit -> char` and used that in `String.init`

ocaml/xapi/xapi_cmd_result.ml
```diff
@@ -36,4 +36,9 @@ let find key results = RMap.find_opt key results
 let of_output_opt ~sep ~key ~lines =
   String.split_on_char '\n' lines
   |> List.fold_left (fun results line -> add ~sep line results) empty
-  |> find key
+  |> RMap.find_opt key
+
+let of_output ~sep ~key lines =
```
lindig, line 41: Why do we need this when we have ` of_output_opt` already? Can we not just call that function and use the result?

liulinC: Lots of other structures like `Map` provides both, 
I think if we provides both, the caller can decides which one to use, handle the return value or exception.

edwintorok: Sure, but in that case you can implement `of_output` by using `of_output_opt`, you can avoid copy/pasting the code.

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: there are some function like `name_of_sid`, I like to keep consistency.

This function is in module `Ldap`, I suppose this indicate query by ldap already.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -692,8 +692,10 @@ module Winbind = struct
   let random_string len =
     let upper_char_start = 65 in
```
lindig, line 693:
```
let upper_char_start = Char.code 'A' in
```

lindig (approved):

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -717,11 +713,7 @@ end
 
 module AuthADWinbind : Auth_signature.AUTH_MODULE = struct
   let get_subject_identifier' subject_name =
-    let* domain =
-      try Ok (get_domain_info_from_db ()).service_name with e -> Error e
-    in
-    let subject_name = domainify_uname ~domain subject_name in
-    Wbinfo.sid_of_name subject_name
+    undomainify_uname subject_name |> Ldap.sid_of_name
```
edwintorok, line 716: What if you have the same username in 2 different domains?
Will this allow `DOMAIN_B\user` to login with the privileges of `DOMAIN_A\user` but with the password of `DOMAIN_B\user`?

liulinC: Here is not writing the `/etc/pam.d/hcp_user`, 
`/etc/pam.d/hcp_users` has domain name prefixed.

edwintorok (changes requested):
Discussed with @liulinC that there are more changes pending here that will address the domain confusion and other issues with user name formats, marking as needs changes until updated PR is available.

lippirk (approved):

liulinC:
@edwintorok @edwintorok @lindig 
Thanks for the help to review this PR, 

As I made quite more change, could you please help to re-revivew this PR, thanks.

lindig (approved):

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -796,27 +843,35 @@ module AuthADWinbind : Auth_signature.AUTH_MODULE = struct
...
+        ("subject-name", user_name)
       ; ("subject-gecos", gecos)
       ; ( "subject-displayname"
-        , if gecos = "" || gecos = "<null>" then name else gecos )
+        , if display_name != "" then
+            display_name
+          else if gecos != "" && gecos != "<null>" then
+            gecos
+          else
+            user_name )
```
lippirk, line 870: you have to be careful in ocaml because `!=` does not do what you might expect (see `!=` vs `<>` https://stackoverflow.com/a/1412688)

please do the following replacement `s/!=/<>/g`

liulinC: nice catch!

ocaml/xapi/xapi_cmd_result.ml
```diff
@@ -33,7 +33,11 @@ let add ?(sep = ':') line results =
...
 let find key results = RMap.find_opt key results
 
-let of_output_opt ~sep ~key ~lines =
+let of_output ~sep ~key lines =
   String.split_on_char '\n' lines
   |> List.fold_left (fun results line -> add ~sep line results) empty
-  |> find key
+  |> RMap.find key
+
+let of_output_opt ~sep ~key ~lines =
```
edwintorok, line 41: Usually this is written the other way around: write the `_opt` function using other `_opt` functions and then write the exception raising function using the `_opt` functions. This is because raising an exception is more time consuming than returning `None`. If you wrap the exception returning function with Some/None then you might be catching more exceptions than intended, but also spending more time because an exception is raised (including capturing of a stacktrace) only to be ignored.

edwintorok (commented):
Otherwise I think looks good, it doesn't have the problem with dropping domains from the user. Login performance/caching is to be determined separately.

scripts/plugins/extauth-hook-AD.py
```diff
@@ -159,30 +147,18 @@ def _apply_to_cache(self):
...
-    def _add_subject(self, name, is_group):
-        condition = "ingroup" if is_group else "="
-        self._lines.append("account sufficient pam_succeed_if.so user {} {} ".format(condition, name))
+                self._add_subject(subject_rec)
+
+    def _add_subject(self, subject_rec):
+        try:
+            sid = subject_rec['subject_identifier']
+            name = subject_rec["other_config"]["subject-name"]
+            is_group = subject_rec["other_config"]["subject-is-group"] == "true"
```
lippirk, line 156: nice! I think this is much better than using wbinfo

edwintorok (approved):

scripts/plugins/extauth-hook-AD.py
```diff
@@ -159,30 +147,18 @@ def _apply_to_cache(self):
...
+            logger.debug("Permit %s with sid %s is_group as %s", name, sid, is_group)
+            condition = "ingroup" if is_group else "="
```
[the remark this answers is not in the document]
liulinC: condition should be `ingroup` if it is group, else, should be `=`. to match exactly user.

ocaml/xapi/xapi_cmd_result.ml
lindig: I am not sure the rationale is sound. An exception in OCaml is very fast because it skips over all frames on the stack to the frame where is is caught (unlike in C++, for example). A regular return (like `None`) typically involves pattern matching in every function/frame until where it is properly handled. But exceptions are not tracked by the type system, making them more prone to programmer error.

PR #4457, opened 2021-07-06
Merge xen-api-client

robhoes:
This is a subtree merge of the http://github.com/xapi-project/xen-api-client repo into `ocaml/`, plus cleanups in the final commit.


psafont:
I can't find any specific tests regarding xapi-client in the test output, are there any tests in its code?

robhoes:
@psafont Note that the imported thing is `xen-api-client`, while there was already a `xapi-client`. The former does run it's old ounit test case. The output is a bit hidden, but looks like this:
```
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
failed to read response
..
Ran: 2 tests in: 0.00 seconds.
OK
```
It looks odd, but the test actually passes.

psafont (approved):

edwintorok (approved):
Do we need to explicitly list all the packages in `dune install`, or would it work if we just told it to install everything? i.e. is there a dune package that we do *not* want to install?

robhoes:
> Do we need to explicitly list all the packages in dune install, or would it work if we just told it to install everything? i.e. is there a dune package that we do not want to install?

The Makefile is currently using dune just to install the libraries, probably because the binaries goes into unusual locations.

PR #4458, opened 2021-07-07
CP-36097 REQ-403 ferry cluster pems

lippirk:
Xapi will generate a pem when it creates a cluster. This cert goes in the cluster database. New cluster hosts will get a copy of the cluster pem when they join (such that they can trust the cluster they are joining)

We also provide a mechanism such that we can introduce pems to an existing cluster. This may be necessary before enabling TLS verification

### DIFF
--- a/ocaml/gencert/selfcert.ml
+++ b/ocaml/gencert/selfcert.ml
@@ -18,6 +18,8 @@ open Rresult (* introduces >>= >>| and R *)
 
 module D = Debug.Make (struct let name = "gencert_selfcert" end)
 
+let ( let* ) = Result.bind
+
 (** initialize the random number generator at program startup when this
 module is loaded. *)
 let () = Mirage_crypto_rng_unix.initialize ()
@@ -81,7 +83,7 @@ let generate_private_key length =
   let stdout, _stderr = call_openssl args in
   stdout
 
-let selfsign issuer extensions key_length days certfile =
+let selfsign' issuer extensions key_length days =
   let rsa =
     try
       generate_private_key key_length
@@ -93,21 +95,25 @@ let selfsign issuer extensions key_length days certfile =
           x
     with e ->
       let msg =
-        Printf.sprintf "generating RSA key for %s failed: %s" certfile
-          (Printexc.to_string e)
+        Printf.sprintf "generating RSA key failed: %s" (Printexc.to_string e)
       in
       D.error "%s" msg ; failwith msg
   in
   let privkey = `RSA rsa in
   let pubkey = `RSA (Rsa.pub_of_priv rsa) in
   let req = X509.Signing_request.create issuer privkey in
-  sign days privkey pubkey issuer req extensions >>= fun cert ->
+  let* cert = sign days privkey pubkey issuer req extensions in
   let key_pem = X509.Private_key.encode_pem privkey in
   let cert_pem = X509.Certificate.encode_pem cert in
   let pkcs12 =
     String.concat "\n\n" [Cstruct.to_string key_pem; Cstruct.to_string cert_pem]
   in
-  write_certs certfile pkcs12 >>| fun () -> cert
+  Ok (cert, pkcs12)
+
+let selfsign issuer extensions key_length days certfile =
+  let* cert, pkcs12 = selfsign' issuer extensions key_length days in
+  let* () = write_certs certfile pkcs12 in
+  Ok cert
 
 let host ~name ~dns_names ~ips pemfile =
   let expire_days = 3650 in
@@ -131,3 +137,16 @@ let xapi_pool ~uuid pemfile =
   selfsign issuer extensions key_length expire_days pemfile
   >>| (fun _ -> ())
   |> R.failwith_error_msg
+
+let xapi_cluster ~cn =
+  (* in practical terms, never expire *)
+  let expire_days = 365 * 2000 in
+  let key_length = 2048 in
+  let issuer =
+    [X509.Distinguished_name.(Relative_distinguished_name.singleton (CN cn))]
+  in
+  let extensions = X509.Extension.empty in
+  let _, pkcs12 =
+    selfsign' issuer extensions key_length expire_days |> R.failwith_error_msg
+  in
+  pkcs12
--- a/ocaml/gencert/selfcert.ml
+++ b/ocaml/gencert/selfcert.ml
@@ -35,7 +35,7 @@ let write_certs path pkcs12 =
   in
   R.trap_exn f () |> R.error_exn_trap_to_msg
 
-let expire_in days =
+let expire_in_days days =
   let seconds = days * 24 * 60 * 60 in
   let start = Ptime_clock.now () in
   match Ptime.(add_span start @@ Span.of_int_s seconds) with
@@ -44,12 +44,14 @@ let expire_in days =
   | None ->
       R.error_msgf "can't represent %d as time span" days
 
+let expire_never () = (Ptime_clock.now (), Ptime.max)
+
 let sans dns_names ips =
   let sans = X509.General_name.(singleton DNS dns_names |> add IP ips) in
   X509.Extension.(singleton Subject_alt_name (false, sans))
 
-let sign days privkey pubkey issuer req extensions =
-  expire_in days >>= fun (valid_from, valid_until) ->
+let sign expiration privkey pubkey issuer req extensions =
+  let valid_from, valid_until = expiration in
   match (privkey, pubkey) with
   | `RSA priv, `RSA pub when Rsa.pub_of_priv priv = pub ->
       X509.Signing_request.sign ~valid_from ~valid_until ~extensions req privkey
@@ -83,7 +85,7 @@ let generate_private_key length =
   let stdout, _stderr = call_openssl args in
   stdout
 
-let selfsign' issuer extensions key_length days =
+let selfsign' issuer extensions key_length expire =
   let rsa =
     try
       generate_private_key key_length
@@ -102,7 +104,7 @@ let selfsign' issuer extensions key_length days =
   let privkey = `RSA rsa in
   let pubkey = `RSA (Rsa.pub_of_priv rsa) in
   let req = X509.Signing_request.create issuer privkey in
-  let* cert = sign days privkey pubkey issuer req extensions in
+  let* cert = sign expiration privkey pubkey issuer req extensions in
   let key_pem = X509.Private_key.encode_pem privkey in
   let cert_pem = X509.Certificate.encode_pem cert in
   let pkcs12 =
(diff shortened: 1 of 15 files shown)


ocaml/gencert/selfcert.ml
```diff
@@ -93,21 +95,25 @@ let selfsign issuer extensions key_length days certfile =
           x
     with e ->
       let msg =
-        Printf.sprintf "generating RSA key for %s failed: %s" certfile
-          (Printexc.to_string e)
+        Printf.sprintf "generating RSA key failed: %s" (Printexc.to_string e)
```
edwintorok, line 98: This will loose the original stacktrace.
I would suggest to add a `Backtrace.is_important e` prior to `let msg` to ensure that the stacktrace is saved and we see the full stacktrace in the logs, even if we change the error message here

ocaml/gencert/selfcert.ml
```diff
@@ -93,21 +95,25 @@ let selfsign issuer extensions key_length days certfile =
...
       in
       D.error "%s" msg ; failwith msg
```
edwintorok, line 100: This is a bit weird mix of result monad chaining, and exceptions.
*If* we wanted we could use the `Rresult` module, which has a `trap_exn` function that captures an exception together with its backtrace in the `Error` variant.
https://erratique.ch/software/rresult/doc/Rresult.R.html

ocaml/gencert/selfcert.ml
```diff
@@ -131,3 +137,16 @@ let xapi_pool ~uuid pemfile =
   selfsign issuer extensions key_length expire_days pemfile
   >>| (fun _ -> ())
   |> R.failwith_error_msg
+
+let xapi_cluster ~cn =
+  (* in practical terms, never expire *)
+  let expire_days = 365 * 2000 in
+  let key_length = 2048 in
```
edwintorok, line 144: Can we share these settings with the one used for xapi itself?

ocaml/xapi/message_forwarding.ml
```diff
@@ -1002,6 +1002,8 @@ functor
               debug "Pool.enable_tls_verification sleeping on fistpoint" ;
               Thread.delay 5.0
             done ;
+            Db.Cluster.get_all ~__context
+            |> List.iter (Xapi_cluster_helpers.Pem.maybe_write_new ~__context) ;
```
edwintorok, line 1006: Do we need to tell each one to write it? Or can we tell it to just the local clusterd, and let the local clusterd run distribute state across the cluster?

ocaml/gencert/selfcert.ml
psafont: We need to implement certificate rotation for the cluster to do that safely.

ocaml/gencert/selfcert.ml
```diff
@@ -131,3 +137,16 @@ let xapi_pool ~uuid pemfile =
   selfsign issuer extensions key_length expire_days pemfile
   >>| (fun _ -> ())
   |> R.failwith_error_msg
+
+let xapi_cluster ~cn =
+  (* in practical terms, never expire *)
+  let expire_days = 365 * 2000 in
```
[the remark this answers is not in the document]
psafont: Isn't that `Ptime.max`?

ocaml/gencert/selfcert.ml
```diff
@@ -42,12 +44,14 @@ let expire_in days =
...
 
+let expire_never () = (Ptime_clock.now (), Ptime.max)
+
 let sans dns_names ips =
   let sans = X509.General_name.(singleton DNS dns_names |> add IP ips) in
   X509.Extension.(singleton Subject_alt_name (false, sans))
 
-let sign days privkey pubkey issuer req extensions =
-  expire_in days >>= fun (valid_from, valid_until) ->
+let sign expire privkey pubkey issuer req extensions =
```
lindig, line 53: Hm. Days is a noun but expire is a verb. Should this not be `expiration`?

ocaml/gencert/selfcert.ml
lippirk: I switched to just using the result monad. I think it's easier to not use `trap_exn` because the debug library in xcp-idl expects `Backtrace.t` for logging, whereas `trap_exn` gives us `Printexn.raw_backtrace` (it's not clear to me that we can convert between the two)

ocaml/xapi/message_forwarding.ml
lippirk: The way it works is that xapi finds an enabled cluster host, gives it a pem blob, and it's that cluster host's job to tell the rest of the cluster hosts (with `distribute_state`). Later on xapi tells each cluster daemon to restart, and at that point they will be using the new pem.

If this distribute state call fails to update one of the cluster hosts, I am envisioning that an admin would (a) emergency disable verification on that host (b) fix the network problem (c) disable+(re)enable the cluster host

ocaml/gencert/selfcert.ml
```diff
@@ -102,51 +104,58 @@ let selfsign' issuer extensions key_length days =
...
+    let extensions = sans dns_names ips in
+    (* make sure name is part of alt_names because CN is deprecated and
+       that there are no duplicates *)
+    selfsign issuer extensions key_length expire pemfile
   in
-  let extensions = sans dns_names ips in
-  (* make sure name is part of alt_names because CN is deprecated and
-     that there are no duplicates *)
-  selfsign issuer extensions key_length expire_days pemfile
-  |> R.failwith_error_msg
```
psafont, line 122: Can the |> be changed back? It's getting unnecessary churn here.

ocaml/gencert/selfcert.ml
```diff
@@ -102,51 +104,58 @@ let selfsign' issuer extensions key_length days =
...
+  let res =
+    let* expire = expire_in 3650 in
+    let key_length = 2048 in
+    let issuer =
+      [
+        X509.Distinguished_name.(Relative_distinguished_name.singleton (CN uuid))
+      ]
+    in
+    let extensions = X509.Extension.empty in
+    let* _ = selfsign issuer extensions key_length expire pemfile in
```
psafont: Can the type it returns be defined instead of using a wildcard?

ocaml/gencert/selfcert.ml
```diff
@@ -82,25 +82,19 @@ let call_openssl args =
...
-          x
-    with e ->
-      let msg =
-        Printf.sprintf "generating RSA key failed: %s" (Printexc.to_string e)
-      in
-      D.error "%s" msg ; failwith msg
-  in
+  let* s = generate_private_key key_length in
+  let* pem = s |> Cstruct.of_string |> X509.Private_key.decode_pem in
+  let rsa = match pem with `RSA x -> x in
```
psafont, line 97: This results in a runtime exception if it's not an RSA private key, the opposite of what you're trying to do ;)

This is not relevant in the current version of mirage-crypto, but it becomes problematic as soon as we update: https://github.com/mirleft/ocaml-x509/blob/main/lib/private_key.ml#L14

lippirk: perhaps it is easier if I remove this commit

psafont: I think it's fine to get it all in the same function

lippirk: I don't quite follow how
```
  let* pem = s |> Cstruct.of_string |> X509.Private_key.decode_pem in
  let rsa = match pem with `RSA x -> x in
```
can give an exception ?

I think I will reword the error from `decode_pem` however

lippirk:
bb7dc24ae9 ensures that when we perform a certificate refresh, we will restart the cluster daemon if necessary

ocaml/gencert/selfcert.ml
psafont: The match can return an exception:
```
# let privkey = `ESDCE 1;;
val privkey : [> `ESDCE of int ] = `ESDCE 1

# let rsa = match privkey with `RSA x -> x;;
Line 1, characters 10-40:
Warning 8: this pattern-matching is not exhaustive.
Here is an example of a case that is not matched:
`ESDCE _
Exception: Match_failure ("//toplevel//", 1, 10).
```

lippirk: ah, for some reason I was not seeing this warning. maybe I have an old version of x509. I'll fix this

psafont: Yeah, this error cannot happening _yet_, but it may be emmited after upgrading to the next version of X509, which has many types of keys.

ocaml/xapi/cert_refresh.ml
```diff
@@ -56,6 +56,17 @@ let new_host_cert ~dbg ~path : X509.Certificate.t =
   let ips = [ip] in
   Gencertlib.Selfcert.host ~name ~dns_names ~ips path
 
+let restart_cluster_daemon ~__context =
```
psafont, line 59:
```suggestion
let maybe_restart_cluster_daemon ~__context =
```

psafont (approved):

psafont:
This needs a rebase

ocaml/gencert/selfcert.ml
```diff
@@ -74,60 +78,90 @@ let call_openssl args =
...
+      let stdout, _stderr = call_openssl args in
+      Ok stdout
     with e ->
-      let msg =
-        Printf.sprintf "generating RSA key for %s failed: %s" certfile
-          (Printexc.to_string e)
-      in
-      D.error "%s" msg ; failwith msg
+      let msg = "generating RSA key failed" in
+      D.error "selfcert.ml: %s" msg ;
```
lindig, line 91: You could use `__LOC__` or `__FUNCTION__`

ocaml/gencert/selfcert.ml
```diff
@@ -74,60 +78,90 @@ let call_openssl args =
...
+  let* () = write_certs certfile pkcs12 in
+  Ok cert
 
 let host ~name ~dns_names ~ips pemfile =
-  let expire_days = 3650 in
-  let key_length = 2048 in
-  let issuer =
-    [X509.Distinguished_name.(Relative_distinguished_name.singleton (CN name))]
+  let res =
+    let* expiration = expire_in 3650 in
```
lindig, line 126: Using `10 * 365` could make the intent more obvious; likewise a name like `lifetime_days`.

ocaml/gencert/selfcert.ml
```diff
@@ -74,60 +78,90 @@ let call_openssl args =
...
-  (* make sure name is part of alt_names because CN is deprecated and
-     that there are no duplicates *)
-  selfsign issuer extensions key_length expire_days pemfile
-  |> R.failwith_error_msg
+  R.failwith_error_msg res
 
 let xapi_pool ~uuid pemfile =
-  let expire_days = 3650 in
+  let res =
+    let* expiration = expire_in 3650 in
```
lindig, line 142: Same as above

lindig (approved):

ocaml/gencert/selfcert.ml
lippirk: I didn't know about `__FUNCTION__`, but unfortunately that is 4.12 :(

PR #4459, opened 2021-07-08
CP-37370 refresh pool-internal certs using cron

lindig:
Call `xe host-refresh-server-certificate` from cron. Do it hourly before release for increased testing but revert to doing it daily before release. This requires a corresponding change in `xapi.spec` for packaging.

### DIFF
--- a/scripts/certificate-refresh
+++ b/scripts/certificate-refresh
@@ -1,0 +1,11 @@
+#! /bin/bash
+# wait randomly up to 10m (600s) and refresh the pool-internal
+# host certificate of this host
+
+set -e
+
+source /etc/xensource-inventory
+host="$INSTALLATION_UUID"
+/usr/bin/sleep $((RANDOM % 600))
+/opt/xensource/bin/xe host-refresh-server-certificate host="$host"
+exit 0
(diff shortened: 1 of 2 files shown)


lippirk (approved):
If for some reason 'xe host-refresh-server-certificate' fails during a cron, I hope that the `am_i_missing_certs` code will fix any problems

we might still get into a pickle if the master changes at the wrong time, but I think the best way to see is to get this tested

scripts/certificate-refresh
```diff
@@ -0,0 +1,11 @@
+#! /bin/bash
+# wait randomly up to 10m (600s) and refresh the pool-internal
```
psafont, line 2: This seems quite a small amount of time, it will probably trigger some parallel operations. This is good for testing, for a release this will probably be changed to be daily. In that case it might do good to have a period of 30 minutes (or more) instead of 10.

psafont (approved):

lippirk:
for what it's worth, I believe this cronjob could cause issues in the following scenario:

 - pool.join starts, joiner obtains primary host's cert
 - whilst the joiner is rebooting, cronjob triggers a cert refresh on the master (crucially the joiner is not yet in the pool's db, so doesn't receive the new cert)
 - joiner comes back online and has the wrong cert for the primary host

lindig:
CA-356724 fixed the reliability of certificate refresh - so it seems to me that this is good to merge. It needs a corresponding merge from `xapi.spec` to install the cron job.

robhoes (approved):

PR #4460, opened 2021-07-12
REQ-403 concurrency fixes pt 4

lippirk:
Put cert refresh under the cert distrib lock.

The main aim of this is to mutually exclude both
'copy_primary_host_certs' and cert refresh (tls verification enable is
already excluded via the refresh cert pool op).

It is handy to have both locks, since the pool op excludes pool refresh
with things like HA enable/disable.

### DIFF
--- a/ocaml/xapi/cert_distrib.mli
+++ b/ocaml/xapi/cert_distrib.mli
@@ -1,3 +1,12 @@
+val lock : (unit -> 'a) -> 'a
+(** where possible, the master host should control the certificate
+  * distributions. this allows us to coordinate multiple parties that are
+  * trying to modify /etc/stunnel at the same time with [lock]!
+  *
+  * we apply this lock to all top level distribution calls, with the exception of
+  * the pool join functions that execute on the joiner.
+  *)
+
 val local_exec : __context:Context.t -> command:string -> string
 (** execute a string encoded job, returning a string encoded result *)
 
--- a/ocaml/xapi/cert_distrib.mli
+++ b/ocaml/xapi/cert_distrib.mli
@@ -1,3 +1,17 @@
+(*
+ * Copyright (C) Citrix Systems Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+
 val local_exec : __context:Context.t -> command:string -> string
 (** execute a string encoded job, returning a string encoded result *)
 
--- a/ocaml/xapi/cert_distrib.mli
+++ b/ocaml/xapi/cert_distrib.mli
@@ -1,12 +1,3 @@
-val lock : (unit -> 'a) -> 'a
-(** where possible, the master host should control the certificate
-  * distributions. this allows us to coordinate multiple parties that are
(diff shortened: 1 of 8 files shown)


psafont (approved):

ocaml/xapi/cert_distrib.mli
```diff
@@ -1,3 +1,12 @@
+val lock : (unit -> 'a) -> 'a
```
robhoes, line 1: Separate issue: this file seems to lack its copyright headers.

robhoes:
>It is handy to have both locks, since the pool op excludes pool refresh
with things like HA enable/disable.

Interactions and exclusions between pool operations can be encoded in `Xapi_pool_helpers.with_pool_operation` (as done for operations on other classes, like host and VM). Using a local lock on the master, like `Cert_distrib` may work as well, but it may be clearer (and less error prone for people modifying this code later) to put all high-level locking considerations together?

lippirk:
> > It is handy to have both locks, since the pool op excludes pool refresh
> > with things like HA enable/disable.
> 
> Interactions and exclusions between pool operations can be encoded in `Xapi_pool_helpers.with_pool_operation` (as done for operations on other classes, like host and VM). Using a local lock on the master, like `Cert_distrib` may work as well, but it may be clearer (and less error prone for people modifying this code later) to put all high-level locking considerations together?

I agree it would be nice to use `with_pool_operation` here (maybe replacing the existing `cert_refresh op with a more general `cert_distrib op)

The one thing we need to make sure is that if you try to do two `cert_distrib ops at the same time, then `with_pool_operation` will wait rather than throw an exception (otherwise we'll break parallel pool join again :) ); looking at the code this seems like what it does already though

robhoes:
I believe that the use of `Helpers.retry_with_global_lock` make it so that the call blocks and will attempts to get the lock later, if the exception from `assert_operation_valid` is `Api_errors.other_operation_in_progress`.

psafont:
Tests are failing, but I cannot see which test is failing

lippirk:
This PR has changed a bit from its inception. The current state is:

Cert distribution ops get their own pool op; `with_pool_operation` is a replacement for `Cert_distrib.lock`. There is some fiddling to make sure existing pool jobs interact with these new pool ops correctly

There are still some tricky cases that are not covered here, for example Pool.join - if the primary host cert refresh cron job decides to refresh in the middle of a pool join I suspect bad things would happen

ocaml/xapi/xapi_host.ml
```diff
@@ -2674,7 +2674,11 @@ let alert_if_tls_verification_was_emergency_disabled ~__context =
 let cert_distrib_atom ~__context ~host ~command =
   Cert_distrib.local_exec ~__context ~command
 
-let copy_primary_host_certs = Cert_distrib.copy_certs_to_host
+let copy_primary_host_certs ~__context ~host =
+  Xapi_pool_helpers.with_pool_operation ~__context ~op:`copy_primary_host_certs
```
robhoes, line 2678: Is this called _on_ the primary host?

robhoes (approved):
This all looks good to me, assuming you have verified that all calls to `Xapi_pool_helpers.with_pool_operation` are made only on the primary host.

ocaml/xapi/xapi_host.ml
lippirk: yes

lippirk: it would be clearer if this `with_pool_operation` call was made in message_forward, so i'll move it there

lippirk:
> This all looks good to me, assuming you have verified that all calls to `Xapi_pool_helpers.with_pool_operation` are made only on the primary host.

The `cert_distrib.ml:lock` that this PR replaces had the same requirement, I tried not to break it here :)

ocaml/xapi/xapi_host.ml
robhoes: Yes indeed, if we put this sort of thing in message_forwarding, then it guaranteed to be primary-only.

PR #4461, opened 2021-07-12
quality-gate: error if somebody used physical equality

lippirk:


quality-gate.sh
```diff
@@ -39,6 +39,24 @@ mli-files () {
...
+  # prefer =,<> to  ==,!=
+  if git grep -r --count ' == ' -- '**/*.ml' ':!ocaml/sdk-gen/**/*.ml'; then
+    echo "ERROR expected no usages of ' == '" 1>&2
+    exit 1
+  else
+    echo "OK found no usages of ' == '"
+  fi
+
+  if git grep -r --count ' != ' -- '**/*.ml' ':!ocaml/sdk-gen/**/*.ml'; then
+    echo "ERROR expected no usages of ' != '" 1>&2
```
psafont, line 52: Could we suggest using <> in the message?, it's not clear what the action should be taken unless the user knows the difference in equalities.

robhoes (approved):

psafont (approved):

PR #4462, opened 2021-07-12
CP-36098 don't refresh certs if any host is offline

lindig:
When a host refreshes its internal certificate, it distributes it to all
pool members which will need it for future communication. If any host is
offline, it would miss out. Hence, we fail the operatio if that is the
case and don't refresh a host's certificate.


psafont (approved):
This will create a visible error in the cronjob and warn the user when a host is offline when a certificate change is triggered

lippirk (approved):

PR #4463, opened 2021-07-12
Add client_certificate_auth fields and functions

robhoes:
Two new fields:
- pool.client_certificate_auth_enabled
- pool.client_certificate_auth_disabled

Two functions to modify these fields:
- pool.enable_client_certificate_auth
- pool.disable_client_certificate_auth

The functions just set the DB fields for the moment. Code to
actually enable/disable client-certificate auth will follow.


lippirk (approved):
Looks good, though you have some unit tests to fix

lindig:
Why do we need two functions rather than one that takes a boolean to enable/disable authentication?

robhoes:
> Why do we need two functions rather than one that takes a boolean to enable/disable authentication?

That's what I had first, actually. However, I wanted to set the name and enable in a single step, so the enable function needed a name param. For disabling, you don't need to give a name, so that param becomes awkward.

Furthermore, we have enable+disable functions for other pool operations, so it fits in: enable_ha/disable_ha, enable_binary_storage/disable_binary_storage, enable_external_auth/disable_external_auth, enable_redo_log/disable_redo_log, enable_ssl_legacy/disable_ssl_legacy, enable_local_storage_caching/disable_local_storage_caching.

lindig (approved):

PR #4464, opened 2021-07-13
XenAPI.Session: raise exception on attempted forwarding of python magic methods

edwintorok:
Python calls magic methods for certain operations, e.g.
`not session` calls `session.__nonzero__`
and `session == foo` calls `session.__eq__`.

In the case of `__nonzero__` the failure is obvious: XAPI will return an RPC error
because it doesn't implement __nonzero__, which will be a dict and not an integer and Python raises a type error.
For a recent example of such a bug, see: https://github.com/xapi-project/sm/pull/562/commits/c52bcdf4e219126cb2b3b7a163902b85912ce4a3

`__eq__` is more problematic because the result is an error, a dict, which can be converted to an always true boolean.

Explicitly check for magic methods (starting with `__`) and reject it as not implemented.
This will give a clear stacktrace and error message, allowing to locate these bugs more easily.
Also prints the method name and arguments, which could help identify the callsite.

### DIFF
--- a/scripts/examples/python/XenAPI/XenAPI.py
+++ b/scripts/examples/python/XenAPI/XenAPI.py
@@ -113,6 +113,9 @@ class UDSTransport(xmlrpclib.Transport):
         for key, value in self._extra_headers:
             connection.putheader(key, value)
 
+def notimplemented(name, *args, **kwargs):
+    raise NotImplementedError("XMLRPC proxies do not support python magic methods", name, *args, **kwargs)
+
 class Session(xmlrpclib.ServerProxy):
     """A server proxy and session manager for communicating with xapi using
     the Xen-API.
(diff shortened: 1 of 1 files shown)


edwintorok:
You should now get stacktraces like this:
```
  File "x.py", line 3, in <module>
    session == "foo"
  File "/usr/lib/python2.7/site-packages/XenAPI.py", line 222, in <lambda>
    return lambda *args, **kwargs: notimplemented(name, args, kwargs)
  File "/usr/lib/python2.7/site-packages/XenAPI.py", line 117, in notimplemented
    raise NotImplementedError("XMLRPC proxies do not support python magic methods", name, *args, **kwargs)
NotImplementedError: ('XMLRPC proxies do not support python magic methods', '__eq__', ('foo',), {})
```

psafont:
Could `__eq__` be treated in a special case so the error says "XMLRPC proxies do not support equality nor other python magic methods"? This way users don't have to know that `==` is implemented as the magic method `__eq__`.

MarkSymsCtx:
> Could `__eq__` be treated in a special case so the error says "XMLRPC proxies do not support equality nor other python magic methods"? This way users don't have to know that `==` is implemented as the magic method `__eq__`.

It's not just `__eq__`, that's just an example. The start of this journey was `__nonzero__` being called as a result of `if not session:`. There are probably other builtins that Python will magically call to handle operator overloads.

lippirk (approved):

lindig (approved):

psafont (approved):

PR #4465, opened 2021-07-14
CP-37500: Support multiple Domain controller in domain

liulinC:
User information is queried through ldap
When joining domain, a machine account and pasword is created in
the joined domain controller, and such info is used to by ldap
query for authentication.

However, if the domain have multiple domain controller,
the joined domain needs time to sync the data to other controllers.
Before the syncing is finished, if one lday query goes to other
domain controller, then it will failed as auth error.

This commit fix the issue by providing winbind krb5.conf to ldap query
This will guides ldap query goes to the right domain controller

### DIFF
--- a/ocaml/xapi/extauth_plugin_ADwinbind.ml
+++ b/ocaml/xapi/extauth_plugin_ADwinbind.ml
@@ -58,6 +58,8 @@ let wb_cmd = !Xapi_globs.wb_cmd
 
 let tdb_tool = !Xapi_globs.tdb_tool
 
+let domain_krb5_dir = Filename.concat Xapi_globs.samba_dir "lock/smb_krb5"
+
 let debug_level () = !Xapi_globs.winbind_debug_level |> string_of_int
 
 let ntlm_auth uname passwd : (unit, exn) result =
@@ -189,16 +191,12 @@ module Ldap = struct
       ; password_expired= logand user_account_control passw_expire_bit <> 0l
       }
 
-  let net_ads (sid : string) : (string, exn) result =
-    try
-      let args = ["ads"; "sid"; "-d"; debug_level (); "--machine-pass"; sid] in
-      let stdout =
-        Helpers.call_script ~log_output:On_failure !Xapi_globs.net_cmd args
-      in
-      Ok stdout
-    with _ -> Error (generic_ex "net ads query failed")
-
-  let query_user sid domain =
+  let query_user sid domain_netbios domain =
+    let domain_krb5_cfg =
+      Filename.concat domain_krb5_dir
+        (Printf.sprintf "krb5.conf.%s" domain_netbios)
+    in
+    let env = [|Printf.sprintf "KRB5_CONFIG=%s" domain_krb5_cfg|] in
     let* stdout =
       try
         let args =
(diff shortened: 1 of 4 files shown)


ocaml/xapi/xapi_globs.ml
```diff
@@ -911,12 +911,14 @@ let ntlm_auth_cmd = ref "/usr/bin/ntlm_auth"
 
 let winbind_debug_level = ref 2
 
-let winbind_cache_time = ref 300
+let winbind_cache_time = ref 10
```
lippirk, line 914: what's the benefit of having a low cache time here? I would expect the data on the windows server to be changing pretty infrequently

docs for this config setting: https://www.samba.org/samba/docs/current/man-html/smb.conf.5.html#WINBINDCACHETIME

liulinC: See the updated design page.
xapi query such info from cache, like query sid from cache, and query user information from DC by sid, 
if the user infomate is updated ( delete and re-add back), then xapi cannot resolve the SID.

This is the time how often winbind refresh its cache.  and the load looks okay,

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -745,7 +746,9 @@ module Winbind = struct
   let random_string len =
     let upper_char_start = 65 in
```
lindig, line 747: `Char.code('A')` would be more readable

lindig: Do we have to make sure this is random enough? The range of random characters is quite limited and if the string is also short, it could be enumerated. How important is this from a security perspective?

lindig (approved):

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: Every join domain will create a machine account and coresponding passwrod in domain controller, 
The mache account name lengh is limitted to 15 chars.

If the  hostname longger than 15 bits, we will construct one
`7_chars_of_prefix_of_hostname`-`7_chars_of_random`

Yes, the random characters is limitted,  (26*26*26*26*26*26*26),  
but we  only need the it when the host name longger than 15, and the how many depends on how many users join domain. I do not expect customer can join domain so many times.
BTW, PBIS do the same way.
This does not related to security.

lippirk (approved):

liulinC:
Check failed with unrelated errors to download packages, 

Ocaml unittest passed in my local dev envioronment , so just merge this PR, 
details, see

```Error:  The sources of the following couldn't be obtained, aborting:
          - menhir.20210310:
            https://gitlab.inria.fr/fpottier/menhir/-/archive/20210310/archive.tar.gz (Bad checksum, expected md5=1cbc71c0bc1f3ddc3e71d5c1f919fd1a)
          - menhirLib.20210310:
            https://gitlab.inria.fr/fpottier/menhir/-/archive/20210310/archive.tar.gz (Bad checksum, expected md5=1cbc71c0bc1f3ddc3e71d5c1f919fd1a)
          - menhirSdk.20210310:
            https://gitlab.inria.fr/fpottier/menhir/-/archive/20210310/archive.tar.gz (Bad checksum, expected md5=1cbc71c0bc1f3ddc3e71d5c1f919fd1a)
'opam install xapi-cli-protocol xapi-client xapi-consts xapi-database xapi-datamodel xapi-types xapi xe xen-api-sdk --deps-only --with-test -v' failed.
Error: Process completed with exit code 40```

PR #4466, opened 2021-07-14
CP-34726: Add client certificate auth to stunnel config

robhoes:
If `pool.client_certificate_auth_enabled` is `true`, then
the incoming stunnel config is modified to verify the certificate
offered by the client. If valid, then the connection is forwarded
to a new unix socket where no further authentication is required.
The socket set up is still to do.


ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -48,10 +48,17 @@ end = struct
   let accept_cached = ref None
 
   let _restart_no_cache ~__context ~accept =
+    let client_auth_name =
```
lippirk, line 51: Do we have a chicken/egg problem on secondary hosts here? I believe this code will query the db just before starting stunnel for the first time on a _secondary_ host, but if we have db access that implies stunnel is already running ..?

I wonder if we even need to add this extra bit of stunnel config on secondary hosts in the first place (since all connections from the outside are coming in via the primary host)

robhoes: Keep in mind that there are separate stunnel processes for outgoing and incoming connections. This config is for the incoming one only, and I don't think that this problem exists.

Having said that, I only wanted to expose this functionality on the the primary host, and I'll need to add some code to do that.

robhoes: Updated.

lippirk: yup, you are right :)

lippirk (approved):

robhoes:
I think that the CI failed due to some unrelated idl changes that we merged today.

ocaml/xapi/helpers.ml
```diff
@@ -1849,55 +1855,80 @@ end = struct
...
+           ; ""
+           ; "[xapi]"
+           ; Printf.sprintf "accept = %s" accept
+           ; Printf.sprintf "cert = %s" cert
+           ; Printf.sprintf "ciphers = %s" Xcp_const.good_ciphersuites
+           ; "curve = secp384r1"
+           ; "options = CIPHER_SERVER_PREFERENCE"
+           ; "sslVersion = TLSv1.2"
+           ]
+          @ ( match client_auth_name with
```
lindig, line 1885: I think we should use `List.concat` on the other level and avoid `@` here.

ocaml/xapi/helpers.ml
```diff
@@ -1826,14 +1830,16 @@ end = struct
...
     (** create or update stunnel config file *)
   end = struct
     let m = Mutex.create ()
 
     let current_accept = ref None
 
-    let rewrite_xapi_ssl_config_file ~accept =
+    let current_client_auth_name = ref None
+
+    let rewrite_xapi_ssl_config_file ~accept ~client_auth_name =
```
lindig, line 1842: I think `rewrite` does not capture what this code is doing - it does not read the file and modifies it. It generates a configuration. So I would call it `emit_xapi_ssl_config` or `generate_...`.

ocaml/xapi/helpers.ml
```diff
@@ -1849,55 +1855,80 @@ end = struct
...
+          | Some current_accept, Some current_client_auth_name
+            when current_accept = accept
+                 && current_client_auth_name = client_auth_name ->
               ()
-          | None | Some _ ->
+          | _ ->
               current_accept := Some accept ;
-              rewrite_xapi_ssl_config_file ~accept
+              current_client_auth_name := Some client_auth_name ;
+              rewrite_xapi_ssl_config_file ~accept ~client_auth_name
```
lindig, line 1931: maybe add a comment here how the new configuration will be picked up

ocaml/xapi/helpers.ml
robhoes: Then `@` becomes `;`, but is that clearer/better?

robhoes:
@lippirk @lindig I've updated this PR with a couple of commits to remove some complexity in the area.

ocaml/xapi/xapi_stunnel.ml
```diff
@@ -0,0 +1,198 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+open Xapi_stdext_threads.Threadext
+module Unixext = Xapi_stdext_unix.Unixext
+
+module D = Debug.Make (struct let name = "stunnel" end)
```
lippirk, line 18: I'd make this `xapi_stunnel_server`, to avoid confusion with https://github.com/xapi-project/xen-api-libs-transitional/blob/656e5e750d77a98e6656ac1c37d9309be795b0b6/stunnel/stunnel.ml#L16

ocaml/xapi/xapi_stunnel.mli
```diff
@@ -0,0 +1,23 @@
...
+
+(** Wrapper around the stunnel@xapi systemd service.
+    There exist scripts (e.g. xe-toolstack-restart) which also manipulate
+    the stunnel daemon but they do this directly (not via ocaml). *)
+
+val restart : __context:Context.t -> accept:string -> unit
+(** restart stunnel, possibly changing the config file *)
+
+val reload : unit -> unit
+(** reload (potentially updated) configuration *)
```
lippirk, line 23: As far as I can tell from the implementation, the only way to change the config is to call `restart`, which means calling `reload` wouldn't actually load a new config (this is probably correct - if you change the config completely sending a SIGHUP is probably not the right thing to do)

tldr; the purpose of reload is for when you've changed your trusted certs? the comment should probably talk about that rather than the config

ocaml/xapi/xapi_pool.ml
```diff
@@ -3404,12 +3404,12 @@ let check_update_readiness ~__context ~self ~requires_reboot =
 
 let enable_client_certificate_auth ~__context ~self ~name =
   (* TODO: actually enable client-certificate auth *)
```
lippirk, line 3406: in a future PR you will want to do an stunnel restart, rather than a reload? (because reloading does not reload the config)

restarting would of course drop the connection, so it's not ideal

lindig: I believe an Stunnel reload (via systemd which is sending a signal) does reload the configuration.

ocaml/xapi/helpers.ml
lindig: With `@` you have to be careful about precedence if you are combine it with `if-then-else`.

ocaml/xapi/xapi_stunnel.mli
robhoes: I think you're right yes. It's pre-existing code/comment, but I can add a commit to make it clearer now that I'm here.

ocaml/xapi/xapi_pool.ml
robhoes: Yes, I think that's the only option.

robhoes: OK, I'll try it out.

lippirk (approved):

ocaml/xapi/helpers.ml
robhoes: IMO "rewrite" is correct in the sense that it writes the file again to update its contents.

robhoes: Done.

lindig (approved):

PR #4472, opened 2021-07-21
CP-37549: Address subject-add performance issue

liulinC:
subject-add needs to perform ldap query for subject information like subject-name
lockout, disabled, etc. This takes long time for subject add

This commit address the issue by
1. Only query subject information from sid once
2. Query and cache closest KDC and update regularly

### DIFF
--- a/ocaml/xapi/extauth_plugin_ADwinbind.ml
+++ b/ocaml/xapi/extauth_plugin_ADwinbind.ml
@@ -228,7 +228,7 @@ module Ldap = struct
     parse_user stdout <!> generic_ex "%s"
 end
 
-type domain_name_type = | Name | NetbiosName
+type domain_name_type = Name | NetbiosName
 
 module Wbinfo = struct
   let exception_of_stderr =
@@ -236,7 +236,8 @@ module Wbinfo = struct
     let regex = Re.Perl.(compile (re {|.*(WBC_ERR_[A-Z_]*).*|})) in
     let get_regex_match x =
       Option.bind (Re.exec_opt regex x) (fun g ->
-          match Re.Group.all g with [|_; code|] -> Some code | _ -> None)
+          match Re.Group.all g with [|_; code|] -> Some code | _ -> None
+      )
     in
     fun stderr ->
       get_regex_match stderr
@@ -261,7 +262,8 @@ module Wbinfo = struct
                  Not_found
              | _ ->
                  Auth_service_error
-                   (E_GENERIC, Printf.sprintf "unknown error code: %s" code))
+                   (E_GENERIC, Printf.sprintf "unknown error code: %s" code)
+         )
 
   let call_wbinfo (args : string list) : (string, exn) result =
     let generic_err () =
@@ -315,29 +317,31 @@ module Wbinfo = struct
      * *)
     let args = ["--domain-info"; from_name] in
     let* stdout = call_wbinfo args in
-    let key = match target_name_type with
-    | Name -> "Alt_Name"
-    | NetbiosName -> "Name" in
-    try
-      Ok(Xapi_cmd_result.of_output ~sep:':' ~key stdout)
+    let key =
+      match target_name_type with Name -> "Alt_Name" | NetbiosName -> "Name"
+    in
+    try Ok (Xapi_cmd_result.of_output ~sep:':' ~key stdout)
     with _ -> Error (parsing_ex args)
 
   let is_domain_netbios_valid domain_netbios =
     let min_valid_domain_length = 1 in
     match domain_name_of ~target_name_type:Name ~from_name:domain_netbios with
-    | Ok name -> String.length name >= min_valid_domain_length
-    | Error _ -> false
+    | Ok name ->
+        String.length name >= min_valid_domain_length
+    | Error _ ->
+        false
 
   let domain_of_uname uname =
     match String.split_on_char '\\' uname with
     | [domain_netbios; _] ->
-      let* domain_name = domain_name_of ~target_name_type:Name ~from_name:domain_netbios in
-      Ok (domain_netbios, domain_name)
+        let* domain_name =
+          domain_name_of ~target_name_type:Name ~from_name:domain_netbios
+        in
+        Ok (domain_netbios, domain_name)
     | _ ->
         Error (generic_ex "Invalid domain user name %s" uname)
 
-
-  let all_domain_netbios() =
+  let all_domain_netbios () =
     (*
      * List all domains (trusted and own domain)
      * wbinfo --all-domains
@@ -355,8 +359,10 @@ module Wbinfo = struct
      * *)
     let args = ["--all-domains"] in
     let* stdout = call_wbinfo args in
-    Ok(String.split_on_char '\n' stdout
-      |> List.filter (fun x -> is_domain_netbios_valid x))
+    Ok
+      (String.split_on_char '\n' stdout
+      |> List.filter (fun x -> is_domain_netbios_valid x)
+      )
 
   type name = User of string | Other of string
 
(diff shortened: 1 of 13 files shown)


liulinC:
Just leave the format issue here, I will format it before merge to avoid merge conflict.

ocaml/xapi/xapi_globs.ml
```diff
@@ -915,6 +915,10 @@ let winbind_cache_time = ref 60
 
 let winbind_machine_pwd_timeout = ref (7 * 24 * 3600)
 
+let winbind_update_closest_kdc_interval = ref (3600. *. 24.) (* every day *)
+
+let winbind_update_closest_kdc_delay = ref 5. (* Delay 5 seconds to make way for xapi startup *)
```
lippirk, line 920: this is not used. It's also probably not a good idea to try and guess an upper bound for how long it takes for xapi to start

liulinC: my bad, I comment it out for debug.

My understand about it is the delay to execute the function after push to queue. 
We need a value for this during xapi startup, to avoid take too many resources from xapi.
Or we can just use 0. if no concern.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -765,6 +805,68 @@ module Winbind = struct
...
+      kdcs_of_domain domain
+      |> List.map (fun kdc -> debug "Got domain '%s' kdc '%s'" domain kdc; kdc)
+      |> List.map (fun kdc -> mtime_this ~name:kdc ~f:(fun() -> Ldap.query_user krbtgt_sid domain_netbios_name kdc |> maybe_raise))
+      |> List.filter (fun (_, span) -> Mtime.Span.equal Mtime.Span.max_span span|> not)
+      |> List.sort (fun (_, s1) (_, s2) -> Mtime.Span.compare s1 s2)
+      |> hd (Printf.sprintf "domain %s does not has valid kdc" domain)
+      |> fst
+    with _ ->
+      debug "Failed to lookup domain %s closest kdc" domain;
+      invalid_kdc
```
lippirk, line 848: maybe you would do this in javascript, but not in ocaml :)

please use `Ok/Error` to encode whether the computation succeeded or not, rather than trying to encode all the information with a string

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -765,6 +805,68 @@ module Winbind = struct
...
+
+  let update() =
+    try
+      Wbinfo.all_domain_netbis() |> maybe_raise
+      |> List.map (fun domain_netbios -> Wbinfo.domain_name_of ~target_name_type:Name ~from_name:domain_netbios |> maybe_raise)
+      |> List.map (fun domain -> domain, lookup domain)
+      |> List.filter (fun (_, kdc) -> kdc <> invalid_kdc)
+      |> List.iter (fun (domain, kdc) -> update_db ~domain ~kdc)
+    with e -> error "Failed to update domain kdc %s" (Printexc.to_string e)
```
lippirk, line 857: I would use the result monad here, rather than trying to converting any errors to exceptions

e.g.
```
let* domains = Wbinfo.all_domain_netbis () in ...
```

also if we fail to get the domain name of any 'domain_netbios', then we don't update the db at all; you may want to check that this is the right behaviour

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -294,28 +303,61 @@ module Wbinfo = struct
...
-        with _ -> Error (parsing_ex args)
-      )
+    | [domain_netbios; _] ->
+      let* domain_name = domain_name_of ~target_name_type:Name ~from_name:domain_netbios in
+      Ok (domain_netbios, domain_name)
     | _ ->
         Error (generic_ex "Invalid domain user name %s" uname)
 
+
+  let all_domain_netbis() =
```
lippirk, line 340: i'm not sure what the difference between netbis and netbios is, is this just a typo?

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -765,6 +805,68 @@ module Winbind = struct
...
+  let from_db domain =
+    (fun __context ->
+     let self = Helpers.get_localhost ~__context in
+     Db.Host.get_external_auth_configuration ~__context ~self|> List.assoc_opt domain)
+  |> Server_helpers.exec_with_new_task "query domain closest kdc"
+
+  let lookup domain =
+    try
+      let krbtgt_sid = Wbinfo.sid_of_name (Printf.sprintf "%s@%s" krbtgt domain) |> maybe_raise in
+      let domain_netbios_name = Wbinfo.domain_name_of ~target_name_type:NetbiosName ~from_name: domain |> maybe_raise in
```
lippirk, line 837: similarly here, remove the try catch and use `let* x = y`

ocaml/xapi/xapi_globs.ml
lippirk: fair enough. As I understand it, the delay time before trying to find the closest kdc is really 5s + the amount of time it takes to check whether winbind is ready, but I assume the latter does not really take that long in the average case

fwiw, it may also be simpler to just put these constants in the winbind ml file, but I am not too strongly opinionated on that

liulinC: I put it into xapi_glob.ml to make it configurable.

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: Yes, this is expected, Find the closest KDC is only best effort.
If closest KDC is not stored in xaipdb, xapi will just choose the first one as before.

liulinC: I have updated the code to use monad as possible,
But I would ike to leave this one, as I expect this function to return void other than Result, 
which is expected by the `Xapi_periodic_scheduler.add_to_queue`

lippirk: you can still use the `let*` bindings, even if you function returns unit, you just have to put it into a block, e.g:

```
let res : ('a, 'b) result =
  let* x = y in
  ...
  Ok x
in
maybe_raise res
```

lippirk (approved):

robhoes (approved):
I don't understand all the details, but think it's is broadly fine. However, please run `make format` and squash the patches.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -765,6 +805,67 @@ module Winbind = struct
...
+  |> Server_helpers.exec_with_new_task "query domain closest kdc"
+
+  let lookup domain =
+    try
+      let* krbtgt_sid = Wbinfo.sid_of_name (Printf.sprintf "%s@%s" krbtgt domain) in
+      let* domain_netbios_name = Wbinfo.domain_name_of ~target_name_type:NetbiosName ~from_name: domain in
+
+      kdcs_of_domain domain
+      |> List.map (fun kdc -> debug "Got domain '%s' kdc '%s'" domain kdc; kdc)
+      |> List.map (fun kdc -> mtime_this ~name:kdc ~f:(fun() -> Ldap.query_user krbtgt_sid domain_netbios_name kdc))
```
edwintorok, line 839: If one of the KDCs is down this will fail, however the admin can just do a toolstack restart once they fixed the KDC, so this should be fine. If needed we could ignore failures here and use filter_map

edwintorok (approved):

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: If we find the closest DC, we will trust it and use it for the next whole day, 
If the chosen DC is down, yes, this issue can be simply resolved by xapi restart.
Also, this check interval can be configured in xapi configure file. (we do not expected the DC fail often)

PR #4476, opened 2021-07-23
REQ-403 add fist to Cert_distrib.exchange_certificate_among_all_members

lippirk:
The idea here is that a test harness can place `/tmp/fist_exchange_certificates_among_all_members` on the file system of the primary host, and cause subsequent calls to`Pool.enable_tls_verification` to fail during its critical section. This will allow the test harness to mimic network failures, and reliably test their effect on this api call

### DIFF
--- a/ocaml/xapi/cert_distrib.ml
+++ b/ocaml/xapi/cert_distrib.ml
@@ -364,7 +364,7 @@ let take_and_append n x xs =
   in
   loop 0 [] xs
 
-let exchange_certificates_among_all_members ~__context =
+let exchange_certificates_in_pool ~__context =
   (* here we coordinate the certificate distribution. from a high level
      we do the following:
      a) collect certs from all the members, aggregating them on the master.
@@ -387,7 +387,7 @@ let exchange_certificates_among_all_members ~__context =
      *   - throw an error at a random point
      *   - print out what is going to execute for debugging purposes
      *)
-    match Xapi_fist.exchange_certificates_among_all_members () with
+    match Xapi_fist.exchange_certificates_in_pool () with
     | None ->
         Fun.id
     | Some seed ->
@@ -401,16 +401,13 @@ let exchange_certificates_among_all_members ~__context =
                   Api_errors.(
                     Server_error
                       ( internal_error
-                      , [
-                          "/tmp/fist_exchange_certificates_among_all_members \
-                           FIST!"
-                        ]
+                      , ["/tmp/fist_exchange_certificates_in_pool FIST!"]
                       )
                   )
             )
           in
           let ops' = take_and_append rand_i throw_op ops in
-          D.debug "exchange_certificates_among_all_members: we are about to..." ;
+          D.debug "exchange_certificates_in_pool: we are about to..." ;
           List.iteri (fun i (desc, _) -> D.debug "%d. %s" i desc) ops' ;
           ops'
   in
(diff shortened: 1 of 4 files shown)


ocaml/xapi/xapi_fist.ml
```diff
@@ -122,3 +122,6 @@ let hang_psr psr_checkpoint =
       "psr_cleanup"
   )
   |> hang_while
+
+let exchange_certificates_among_all_members () =
```
lindig, line 126: Would `exchange_certificates_in_pool` be clearer`? `all_members` makes me wonder if they are different from the pool.

lippirk:
I made a couple of changes:

1. as requested, if you insert a fist point, a summary of what will happen is printed:

```
[debug||783 /var/lib/xcp/xapi|pool.enable_tls_verification D:c6023c9e0353|cert_distrib] exchange_certificates_among_all_members: we are about to...
[debug||783 /var/lib/xcp/xapi|pool.enable_tls_verification D:c6023c9e0353|cert_distrib] 0. send certs to b4cb481b
[debug||783 /var/lib/xcp/xapi|pool.enable_tls_verification D:c6023c9e0353|cert_distrib] 1. send certs to dbb4cb7a
[debug||783 /var/lib/xcp/xapi|pool.enable_tls_verification D:c6023c9e0353|cert_distrib] 2. FIST
```

2. when creating the fist file, you must add a seed. if you want to make a test reproducible, just use the same seed (and the same pool size)

ocaml/xapi/cert_distrib.ml
```diff
@@ -372,21 +372,77 @@ let exchange_certificates_among_all_members ~__context =
...
+                  Api_errors.(
+                    Server_error
+                      ( internal_error
+                      , ["/tmp/fist_exchange_certificates_in_pool FIST!"]
+                      )
+                  )
+            )
+          in
+          let ops' =
+            List.fold_left
```
psafont, line 400: Fold_left tends to make transformation on lists harder to inderstand. This looks like it has the same behaviour as `Xapi_stdext_std.Listext.List.take (rand_i - 1) @ [throw_op]`, this is not very efficient, adapting the code for `take` might be best here:
https://github.com/xapi-project/stdext/blob/master/lib/xapi-stdext-std/listext.ml#L61
```
  let take_and_append n op list =
    let rec loop i acc = function
      | x :: xs when (i - 1) < n ->
          loop (i + 1) (x :: acc) xs
      | _ ->
          List.rev acc
    in
    loop 0 [op] list
```

ocaml/xapi/cert_distrib.ml
```diff
@@ -372,21 +372,77 @@ let exchange_certificates_among_all_members ~__context =
...
+                if i < rand_i then
+                  (i + 1, op :: acc)
+                else if i = rand_i then
+                  (i + 1, throw_op :: acc)
+                else
+                  (i + 1, acc)
+                )
+              (0, []) ops
+            |> snd
+            |> List.rev
```
psafont, line 411: Fold_left tends to make transformation on lists harder to inderstand. This looks like it has the same behaviour as `Xapi_stdext_std.Listext.List.take (rand_i - 1) @ [throw_op]`, this is not very efficient, adapting the code for `take` might be best here:
https://github.com/xapi-project/stdext/blob/master/lib/xapi-stdext-std/listext.ml#L61
```suggestion
            let take_and_append n op list =
              let rec loop i acc = function
                | x :: xs when i < n ->
                    loop (i + 1) (x :: acc) xs
                | _ ->
                    List.rev acc
              in
              loop 1 [op] list
            in
            take_and_append rand_i throw_op ops
```

psafont (approved):

robhoes (approved):

robhoes:
This looks good, but please squash the fixup before merging.

PR #4478, opened 2021-07-26
CA-356901: Perform ldap query if winbind failed to resovle subject

liulinC:
name to subject identifier

winbind has cache timeout set to 60s, so winbind sync its cached data
with domain controller every 60 seconds.
If a user is newly created in DC within 60s, winbind failed to resovle it.
This commit fix this issue by perform ldap query on winbin fail

### DIFF
--- a/ocaml/xapi/extauth_plugin_ADwinbind.ml
+++ b/ocaml/xapi/extauth_plugin_ADwinbind.ml
@@ -90,21 +90,61 @@ let ntlm_auth uname passwd : (unit, exn) result =
   with _ -> Error (auth_ex uname)
 
 let get_domain_info_from_db () =
-  (fun __context ->
-    let host = Helpers.get_localhost ~__context in
-    let service_name =
-      Db.Host.get_external_auth_service_name ~__context ~self:host
-    in
-    let workgroup, netbios_name =
-      Db.Host.get_external_auth_configuration ~__context ~self:host |> fun l ->
-      (List.assoc_opt "workgroup" l, List.assoc_opt "netbios_name" l)
-    in
-    {service_name; workgroup; netbios_name}
-    )
-  |> Server_helpers.exec_with_new_task
-       "retrieving external auth domain workgroup"
+  Server_helpers.exec_with_new_task "retrieving external auth domain workgroup"
+  @@ fun __context ->
+  let host = Helpers.get_localhost ~__context in
+  let service_name =
+    Db.Host.get_external_auth_service_name ~__context ~self:host
+  in
+  let workgroup, netbios_name =
+    Db.Host.get_external_auth_configuration ~__context ~self:host
+    |> fun config ->
+    (List.assoc_opt "workgroup" config, List.assoc_opt "netbios_name" config)
+  in
+  {service_name; workgroup; netbios_name}
 
 module Ldap = struct
+  module Escape = struct
+    (*
+     * Escape characters according to
+     * https://docs.microsoft.com/en-gb/windows/win32/adsi/search-filter-syntax?redirectedfrom=MSDN#special-characters
+     * *)
+
+    let reg_star = {|*|} |> Re.str |> Re.compile
+
+    let reg_left_bracket = {|(|} |> Re.str |> Re.compile
+
+    let reg_right_bracket = {|)|} |> Re.str |> Re.compile
+
+    let reg_backward_slash = {|\|} |> Re.str |> Re.compile
+
+    let reg_null = "\000" |> Re.str |> Re.compile
+
+    let reg_slash = {|/|} |> Re.str |> Re.compile
+
+    let escape_map =
+      [
+        (* backward slash goes first as others will include backward slash*)
+        (reg_backward_slash, {|\5d|})
+      ; (reg_star, {|\2a|})
+      ; (reg_left_bracket, {|\28|})
+      ; (reg_right_bracket, {|\29|})
+      ; (reg_null, {|\00|})
+      ; (reg_slash, {|\2f|})
+      ]
+
+    let escape str =
+      List.fold_left
+        (fun acc element ->
+          let reg = fst element in
+          let value = snd element in
+          Re.replace_string reg ~by:value acc
+          )
+        str escape_map
+  end
+
+  let escape str = Escape.escape str
+
   type user = {
       name: string
     ; display_name: string
@@ -223,7 +263,7 @@ module Ldap = struct
       ; password_expired= logand user_account_control passw_expire_bit <> 0l
       }
 
-  let env_of_lookup domain_netbios =
+  let env_of_krb5 domain_netbios =
     let domain_krb5_cfg =
       Filename.concat domain_krb5_dir
         (Printf.sprintf "krb5.conf.%s" domain_netbios)
@@ -231,7 +271,7 @@ module Ldap = struct
     [|Printf.sprintf "KRB5_CONFIG=%s" domain_krb5_cfg|]
 
   let query_user sid domain_netbios kdc =
-    let env = env_of_lookup domain_netbios in
+    let env = env_of_krb5 domain_netbios in
     let* stdout =
       try
         (* Query KDC instead of use domain here
@@ -259,7 +299,9 @@ module Ldap = struct
 
   let query_sid ~name ~kdc ~domain_netbios =
     let key = "objectSid" in
-    let env = env_of_lookup domain_netbios in
+    let env = env_of_krb5 domain_netbios in
+    let name = escape name in
+    (* Escape name to avoid injection detection *)
     let query = Printf.sprintf "(|(sAMAccountName=%s)(name=%s))" name name in
     let args =
       [
@@ -676,11 +718,9 @@ let from_config ~name ~err_msg ~config_params =
 let all_number_re = Re.Perl.re {|^\d+$|} |> Re.Perl.compile
 
 let get_localhost_name () =
-  (fun __context ->
-    Helpers.get_localhost ~__context |> fun host ->
-    Db.Host.get_hostname ~__context ~self:host
-    )
-  |> Server_helpers.exec_with_new_task "retrieving hostname"
+  Server_helpers.exec_with_new_task "retrieving hostname" @@ fun __context ->
+  Helpers.get_localhost ~__context |> fun host ->
+  Db.Host.get_hostname ~__context ~self:host
 
 let assert_hostname_valid ~hostname =
   let all_numbers = Re.matches all_number_re hostname <> [] in
@@ -716,13 +756,12 @@ let persist_extauth_config ~domain ~user ~ou_conf ~workgroup ~netbios_name =
         ]
         @ ou_conf
   in
-  (fun __context ->
-    Helpers.get_localhost ~__context |> fun self ->
-    Db.Host.set_external_auth_configuration ~__context ~self ~value ;
-    Db.Host.get_name_label ~__context ~self
-    |> debug "update external_auth_configuration for host %s"
-    )
-  |> Server_helpers.exec_with_new_task "update external_auth_configuration"
+  Server_helpers.exec_with_new_task "update external_auth_configuration"
+  @@ fun __context ->
+  Helpers.get_localhost ~__context |> fun self ->
+  Db.Host.set_external_auth_configuration ~__context ~self ~value ;
+  Db.Host.get_name_label ~__context ~self
+  |> debug "update external_auth_configuration for host %s"
 
 let disable_machine_account ~service_name = function
   | Some u, Some p -> (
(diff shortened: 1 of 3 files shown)


lippirk (approved):
LGTM - even though this is making an extra request sometimes, it'll only happen when users enter an incorrect user name (which should be rare)

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -81,6 +89,21 @@ let ntlm_auth uname passwd : (unit, exn) result =
     Ok ()
   with _ -> Error (auth_ex uname)
 
+let get_domain_info_from_db () =
+  (fun __context ->
+    let host = Helpers.get_localhost ~__context in
```
lindig, line 94: Why is this explicitly curried and `__context` not an ordinary parameter?

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -81,6 +89,21 @@ let ntlm_auth uname passwd : (unit, exn) result =
...
+    let service_name =
+      Db.Host.get_external_auth_service_name ~__context ~self:host
+    in
+    let workgroup, netbios_name =
+      Db.Host.get_external_auth_configuration ~__context ~self:host |> fun l ->
```
lindig, line 99: `fun l ` is a poor name - this should be `auth` or `config`

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -200,12 +223,15 @@ module Ldap = struct
       ; password_expired= logand user_account_control passw_expire_bit <> 0l
       }
 
-  let query_user sid domain_netbios kdc =
+  let env_of_lookup domain_netbios =
```
lindig, line 226: The name could be misleading. This is the environment for the lookup, not of the lookup. I'd call it `krb5_config` because it defines the environment variable `KRB5_CONFIG`.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -345,6 +400,25 @@ module Wbinfo = struct
     | _ ->
         Error (generic_ex "Invalid domain user name %s" uname)
 
+  let domain_and_user_of_uname uname =
```
lindig, line 403: The `of_uname` is redundant. A function always depends in its parameters. There are similar names below. I'm not sure this convention is too useful - but it also does not hurt. If you prefer long names, `split_into_domain_and_user` would work, or a shorter `split` as well.

edwintorok:
There is a typo in the title here and several places in the commit message: `resovle` -> `resolve`.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -81,6 +89,21 @@ let ntlm_auth uname passwd : (unit, exn) result =
...
+    let service_name =
+      Db.Host.get_external_auth_service_name ~__context ~self:host
+    in
+    let workgroup, netbios_name =
+      Db.Host.get_external_auth_configuration ~__context ~self:host |> fun l ->
+      (List.assoc_opt "workgroup" l, List.assoc_opt "netbios_name" l)
+    in
+    {service_name; workgroup; netbios_name}
+    )
+  |> Server_helpers.exec_with_new_task
```
edwintorok, line 104: This could be at the beginning of the function like this:
```ocaml
Server_helpers.exec_with_new_task "retrieving external auth domain workgroup" @@ fun __context ->
...
```

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -227,9 +253,38 @@ module Ldap = struct
...
         Ok stdout
-      with _ -> Error (generic_ex "ldap query failed")
+      with _ -> Error (generic_ex "ldap query user info from sid failed")
     in
     parse_user stdout <!> generic_ex "%s"
+
+  let query_sid ~name ~kdc ~domain_netbios =
+    let key = "objectSid" in
+    let env = env_of_lookup domain_netbios in
+    let query = Printf.sprintf "(|(sAMAccountName=%s)(name=%s))" name name in
```
edwintorok, line 263: doesn't this need escaping? It looks like someone can inject arbitrary values into the LDAP query this way if they add the right number of `)` into the login name.

edwintorok (changes requested):
We probably need a generic function to escape arguments that we pass to ldapquery that can be user controlled.

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: because there are many places the callers do not have a valid `__context`

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: yes, I see, 
There are places use it and I like to just keep it for consistency

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: I will escape the name, but I do not lose my sleep with it as
# This only happens when the winbind cache is missed, winbind cache is refreshed every 60 seconds.
# This just query the subject SID, and does not update system,  Add subject and change subject role will need pool-admin permission.

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: yes,  it sounds redundant, but I prefer this name from caller side, as we know what is outputted from the function.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -259,7 +258,8 @@ module Ldap = struct
 
   let query_sid ~name ~kdc ~domain_netbios =
     let key = "objectSid" in
-    let env = env_of_lookup domain_netbios in
+    let env = env_of_krb5 domain_netbios in
+    let name = String.escaped name in
```
edwintorok, line 262: Escaping must be done by a function that understands LDAP query syntax and escapes characters with special meaning in an LDAP query. `String.escaped` in OCaml just escapes OCaml specific characters.
See here for a list of special characters and how to escape them for LDAP queries: https://docs.microsoft.com/en-gb/windows/win32/adsi/search-filter-syntax?redirectedfrom=MSDN#special-characters

Also depending on what the string argument and where it is used you may have to escape it differently, for example DNs used alone need additional escaping: https://ldapwiki.com/wiki/DN%20Escape%20Values

lindig: This might be well placed into an in-line sub module where you construct the query:
```
module LDAPquery : sig
  type t
  val sid: string -> string -> t
  val to_string: t -> string
end
```

Internally, `t` would be a sum type that represents the different queries that you want to support. Escaping would be either implemented in `sid` or `to_string` - both would be aware of the role of each parameter.

liulinC: we already have a ldap module and I have put it there, please have a check, thanks @lindig @edwintorok

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -81,7 +89,47 @@ let ntlm_auth uname passwd : (unit, exn) result =
...
+
+  let escape str =
+    String.to_seq str
+    |> List.of_seq
+    |> List.map (fun k ->
+           match List.assoc_opt k escape_map with
+           | Some v ->
+               v
+           | None ->
+               String.make 1 k
```
edwintorok, line 129: Creating lots of small 1-byte strings and then concatenating them together is not very efficient.
We can actually use sequences more efficiently here:
There is a `Seq.map`, and it is more efficient to apply this prior to conversion to list (sequences are traversed once and Seq.map operates on one element at a time, whereas List.map creates intermediate lists). 
And because we sometimes want to keep characters as they are (replace 1 char with 1 char), and sometimes expand them (escaping grows the string) then we can use Seq.flat_map for this purpose. So something like:
```ocaml
let escape_map = function
 | '*' -> String.to_seq "\\2a"
...
  | c -> Seq.return c

let escape str =
   str |> String.to_seq
  |> Seq.flat_map escape_map
  |> String.of_seq
```

edwintorok: (Although in this case this is not performance-critical code, the faster version is actually shorter than the original, if one wants to optimize this further could use a Buffer and build the string that way, but I think the sequence is a nicer, more functional abstraction, so we can keep that).

lindig: Good call. The fastest way to do this would be to scan for characters to be escaped - if none are in the string, which I would assume is the common case, just do nothing, otherwise do what Edvin suggests. The predicate `needs_escaping` would stop at the first characters that needs an escape or go to the end of the string otherwise and return false.

liulinC: Yes, this code is not performance critical,  ( I would say more than 99% cases will just get the subject ID from winbind cache, and thus, this code is not ran at all).
please se the updated code, which I think is more clear.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -81,7 +89,61 @@ let ntlm_auth uname passwd : (unit, exn) result =
...
+
 module Ldap = struct
+  module Escape = struct
+    (*
+     * Escape characters according to
+     * https://docs.microsoft.com/en-gb/windows/win32/adsi/search-filter-syntax?redirectedfrom=MSDN#special-characters
+     * *)
+    let reg_star = Str.regexp "*"
+
+    let reg_left_bracket = Str.regexp "("
```
edwintorok, line 114: Str is not thread safe, and XAPI is a multithreaded application. We should use `re` here, use `Re.Posix` instead. 
In fact I see we've already merged some code that uses Re.Str, which has the same issues as Str, it is not thread safe, because it relies on global internal state (you can see this from the function prototypes too: matched_group doesn't take the regex or the match object as parameter, it just relies on global state).

edwintorok: I've opened an issue here to document this upstream https://github.com/ocaml/ocaml-re/issues/194

liulinC: `matched_group` does not takes regex or match object, indicate it use global state some where,
but I am not sure about the using `global_replace` here, it takes regular expression, replacement, and the whole string, 
`val global_replace : regexp -> string -> string -> string`, which looks like does not needs extra state.

liulinC:
updated, @lindig @edwintorok  please help to review at your convenient, thanks

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -81,7 +89,62 @@ let ntlm_auth uname passwd : (unit, exn) result =
...
+      ; (reg_left_bracket, {|\28|})
+      ; (reg_right_bracket, {|\29|})
+      ; (reg_null, {|\00|})
+      ; (reg_slash, {|\2f|})
+      ]
+
+    let escape str =
+      List.fold_left
+        (fun acc element ->
+          let reg = fst element in
```
edwintorok, line 139: This can also be written as `let reg, value = element`, or `fun acc (reg, value)`, but is fine like this too.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -200,12 +263,15 @@ module Ldap = struct
       ; password_expired= logand user_account_control passw_expire_bit <> 0l
       }
 
-  let query_user sid domain_netbios kdc =
+  let env_of_krb5 domain_netbios =
     let domain_krb5_cfg =
       Filename.concat domain_krb5_dir
         (Printf.sprintf "krb5.conf.%s" domain_netbios)
```
edwintorok, line 269: Is this created by winbind itself? I can't find any code in xapi that would create this file

edwintorok (approved):
I'm not an expert on winbind, so this will have to rely on testing to prove that this works correctly and securely. I can't spot any immediate issues in terms of OCaml language.

edwintorok:
Might be useful to add a comment to `get_subject_identifier'` that it is called in the login path with a yet unauthenticated user, and to treat parameters as untrusted (so things like the LDAP escaping are especially important here, otherwise an unauthenticated user could trick XAPI into authenticated them as another user via LDAP query injection).

edwintorok:
There are some concerns on whether the login process here is secure (but that is unrelated to this PR). Could a user set up their own domain server then attempt to login with attacker@attackercontrolled.example.com, which would successfully authenticate them and when asked to map the username to a SID or the SID to a name it would map it to a SID that exists in our AD hierarchy, and map the name to niceuser@trusteddomain.example.com (assuming that the pool is joined to trusteddomain.example.com).

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: yes, winbind create it, we just re-use it (for the KDCs configured)

liulinC:
> There are some concerns on whether the login process here is secure (but that is unrelated to this PR). Could a user set up their own domain server then attempt to login with [attacker@attackercontrolled.example.com](mailto:attacker@attackercontrolled.example.com), which would successfully authenticate them and when asked to map the domain to a KDC it would map it to an attacker controlled one, and map the username to a SID or the SID to a name it would map it to a SID that exists in our AD hierarchy, and map the name to [niceuser@trusteddomain.example.com](mailto:niceuser@trusteddomain.example.com) (assuming that the pool is joined to trusteddomain.example.com).

The faked DC will not serve as normal DC, as during ldap query,  net command will authenticate with domain controller by kerberos protocol, I do not think the fake server can hold the machine password for the machine account created during domain join.    There is an xenrt automation test for this by @xihuany

liulinC:
> Might be useful to add a comment to `get_subject_identifier'` that it is called in the login path with a yet unauthenticated user, and to treat parameters as untrusted (so things like the LDAP escaping are especially important here, otherwise an unauthenticated user could trick XAPI into authenticating them as another user via LDAP query injection).

If you use ldap query injection and successfully become another user, you still needs to provide the password for that user.

liulinC:
The security concen is addressed in PSEC-8206, if still any further concern, let's talk there as it should not block this PR.

PR #4479, opened 2021-07-26
Stunnel.reload: wait 5s by default

lindig:
The stunnel reload is initiated via systemd, which in turn sends a
signal to stunnel. This is an asynchronous call and we can't know for
sure that stunnel has reloaded its configuration when the call returns.
This can create a race condition in that we try using the new
configuration when indeed it is not yet ready. To avoid this, wait 5
seconds. There is no good mechanism to make this synchronous.

### DIFF
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -1826,8 +1826,8 @@ module Stunnel : sig
   val restart : __context:Context.t -> accept:string -> unit
   (** restart stunnel, possibly changing the config file *)
 
-  val reload : unit -> unit
-  (** reload (potentially updated) configuration *)
+  val reload : ?wait:float -> unit -> unit
+  (** reload (potentially updated) configuration and wait 5s *)
 end = struct
   let cert = !Xapi_globs.server_cert_path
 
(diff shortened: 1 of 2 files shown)


ocaml/xapi/helpers.ml
```diff
@@ -1907,7 +1907,10 @@ end = struct
 
   let systemctl_ cmd = systemctl cmd |> ignore
 
-  let reload () = systemctl_ "reload-or-restart"
+  let reload ?(wait = 5.0) () =
+    systemctl_ "reload-or-restart" ;
+    (* We can't be sure that the reload is finished, so wait a moment *)
+    Unix.sleepf wait
```
edwintorok, line 1913: This releases the runtime lock and should allow other threads to run, so should be fine.

edwintorok (approved):
If this still proves to be unreliable could look into other ways of checking when the reload has finished, but I don't see anything better than polling with a sleep between. There is a `ui_config_reloaded` in the stunnel source code, but it is an empty function.

lippirk:
This is going to be slow for large pools (reloading stunnel on each host will take at least `#hosts * 5 seconds`)

I think we should do a single sleep on the host that is refreshing its certificate. In pseudocode I would expect something like:

```
# Host.refresh_server_certificate
for host in pool:
    send_new_cert(host)
    sighup_stunnel(host)

sleep(5)
# hopefully by this point we have give all other stunnels enough time to trust our new cert
switch_to_new_cert()
```

ocaml/xapi/helpers.ml
robhoes: We normally use `Thread.delay`. I don't know if `sleep` is nowadays suitable for threaded programs, but it wasn't in the past.

robhoes (approved):

psafont (approved):

PR #4480, opened 2021-07-27
More on client authentication

robhoes:


ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -117,6 +118,49 @@ end = struct
...
+      let sock_path = Xapi_globs.unix_domain_socket_clientcert in
+      Unixext.mkdir_safe (Filename.dirname sock_path) 0o700 ;
+      Unixext.unlink_safe sock_path ;
+      let domain_sock = Xapi_http.bind (Unix.ADDR_UNIX sock_path) in
+      Http_svr.start Xapi_http.server domain_sock ;
+      configure_port true ;
+      management_server := Some domain_sock
+    )
+
+  let stop () =
```
lippirk, line 154: From what I can tell, if a master changes to a slave this will be called when the toolstack reboots. This means the port gets closed as expected

Similarly `stop` is called when disabling client auth, so the port is closed in this case too :+1:

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -117,6 +118,49 @@ end = struct
...
+  let configure_port cmd =
+    let cmd' = if cmd then "open" else "close" in
+    let _ =
+      Helpers.call_script
+        !Xapi_globs.firewall_port_config_script
+        [cmd'; string_of_int Constants.https_port_clientcert]
+    in
+    ()
+
+  let start () =
```
lippirk, line 143: If IPv6 was enabled/disabled whilst xapi was running, I suspect we'd need something more in this function. However, I don't even know if that is possible, and it is presumably not supported anyway

ocaml/xapi/xapi_stunnel_server.mli
```diff
@@ -16,7 +16,7 @@
     There exist scripts (e.g. xe-toolstack-restart) which also manipulate
     the stunnel daemon but they do this directly (not via ocaml). *)
 
-val restart : __context:Context.t -> accept:string -> unit
+val restart : __context:Context.t -> bool -> unit
```
lippirk, line 19: a named parameter would be readable here

ocaml/xapi/xapi_mgmt_iface.ml
robhoes: This part is independent from IP addresses. It combines management of the unix socket (no TCP/IP) and the TCP port entry in the firewall (independent of IP addressing).

ocaml/xapi/xapi_mgmt_iface.ml
robhoes: I think that is true, yes.

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -117,6 +118,49 @@ end = struct
...
+  let management_server = ref None
+
+  let must_be_running ~__context ~mgmt_enabled =
+    let pool = Helpers.get_pool ~__context in
+    mgmt_enabled
+    && Pool_role.is_master ()
+    && Db.Pool.get_client_certificate_auth_enabled ~__context ~self:pool
+
+  let configure_port cmd =
+    let cmd' = if cmd then "open" else "close" in
```
lindig, line 135: I would prefer a `match` to translate a `cmd` into a string but obviously this works as along as `cmd` is a boolean.

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -117,6 +118,49 @@ end = struct
...
+    let _ =
+      Helpers.call_script
+        !Xapi_globs.firewall_port_config_script
+        [cmd'; string_of_int Constants.https_port_clientcert]
+    in
+    ()
+
+  let start () =
+    if !management_server = None then (
```
lindig, line 144: A `match` would avoid the `( .. )` required by `if` and would enforce making the `else` case explicit.

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -143,9 +187,12 @@ let next_server_mode ~mgmt_enabled =
 
 let mgmt_is_enabled () = Server.current_mode () = Any
 
-let run ~__context ~mgmt_enabled =
+let run ~__context ?mgmt_enabled () =
+  let mgmt_enabled = Option.value ~default:(mgmt_is_enabled ()) mgmt_enabled in
   Mutex.execute management_m (fun () ->
```
lindig, line 192: If the rest of the function has to be executed under the lock `@@ fun () ->` would work, too.

ocaml/xapi/xapi_stunnel_server.ml
```diff
@@ -49,67 +49,71 @@ end = struct
...
+        ; "socket = a:TCP_NODELAY=1"
+        ; "socket = l:TCP_NODELAY=1"
+        ; "socket = r:SO_KEEPALIVE=1"
+        ; "socket = a:SO_KEEPALIVE=1"
+        ; "TIMEOUTclose = 1"
+        ; ( match Sys.getenv_opt "STUNNEL_IDLE_TIMEOUT" with
+          | None ->
+              "; no idle timeout"
+          | Some x ->
+              sprintf "TIMEOUTidle = %s" x
```
lindig, line 76: Should we check that this has an expected format? Do we expect this to be a number?

lindig (approved):

ocaml/xapi/xapi_stunnel_server.ml
```diff
@@ -181,8 +181,22 @@ let update_certificates ~__context () =
...
+      in
+      match (management_enabled, ipv6_enabled) with
+      | true, true ->
+          ":::"
+      | true, false ->
+          ""
+      | false, true ->
+          "::1:"
+      | false, false ->
+          "127.0.0.1:"
```
edwintorok, line 198: Can't we simplify this and always use 127.0.0.1?
Looking at xapi_mgmt_iface.ml in next_server_mode it always listens at least on 127.0.0.1 (and that'll work regardless of whether we also listen to IPv6 or not).

ocaml/xapi/xapi_mgmt_iface.ml
```diff
@@ -187,7 +187,8 @@ let next_server_mode ~mgmt_enabled =
 
 let mgmt_is_enabled () = Server.current_mode () = Any
 
-let run ~__context ~mgmt_enabled =
+let run ~__context ?mgmt_enabled () =
```
edwintorok, line 190: Why is this turned into an optional argument? I see it does get explicitly called with true or false every time.

ocaml/xapi/xapi_stunnel_server.ml
robhoes: It should be a number, yes. It's not new code though, and I don't really want to change its behaviour in this PR.

ocaml/xapi/xapi_mgmt_iface.ml
robhoes: I do prefer an old-fashioned if statement sometimes :) Especially for simple Boolean checks. A matter of taste I guess.

ocaml/xapi/xapi_stunnel_server.ml
robhoes: This controls what stunnel listens on. To accept external connections, it needs to listen on the actual IP addresses of the host.

PR #4482, opened 2021-07-28
configure: work around read-only /tmp found in opam's 2.1.0

psafont:
Use TMPDIR instead if it's provided

### DIFF
--- a/configure
+++ b/configure
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-D=$(mktemp -d /tmp/configure.XXXXX)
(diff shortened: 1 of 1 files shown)


lippirk:
it's not obvious to me where this environment variable gets set? however the change seems sensible

lippirk (approved):

edwintorok:
OPAM sets the environment variable. If it is not set then it'll default to /tmp

edwintorok (approved):

lindig (approved):

PR #4484, opened 2021-07-29
CA-356441: Restart sshd to apply sshd configuration

liulinC:

### DIFF
--- a/scripts/plugins/extauth-hook-AD.py
+++ b/scripts/plugins/extauth-hook-AD.py
@@ -13,6 +13,7 @@ import abc
 import XenAPIPlugin
 import XenAPI
 import sys
+import subprocess
 import os
 import tempfile
 import logging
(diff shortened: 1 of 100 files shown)


liulinC:
Per test result, if we does not restart sshd after configuration update, ssh still works well, 
but if the password is wrong, the error message is promoted only after user logged in.

scripts/plugins/extauth-hook-AD.py
```diff
@@ -248,6 +260,10 @@ def __init__(self, session, args, ad_enabled=True):
         self._update_key_value("GSSAPIAuthentication", value)
         self._update_key_value("GSSAPICleanupCredentials", value)
 
+    def apply(self):
+        super(SshdConfig, self).apply()
+        run_cmd(["/usr/bin/systemctl", "restart", "sshd"])
```
edwintorok, line 265: SSHd is smart enough to keep existing connections open over a restart (it uses a child process), but if all you want is to load a new config then `reload` might be better here

lindig: Indeed, `sshd.service` supports `reload-or-restart` which would use `restart` as a fallback if it were not supported.

lindig (changes requested):
Use `systemctl reload-or-restart sshd`

liulinC:
Thanks Edwin and Lindig for the comments, 
Since the feature master is merged to master, I will just close this PR and raise a new PR target to master to followup.

PR #4485, opened 2021-07-29
REQ-403 change type of cert generated during cert refresh

lippirk:
We should be creating a cert with CN=$uuid, rather than with CN=$ip


psafont (approved):
how did we miss this, ouch :(

lindig (approved):

PR #4486, opened 2021-07-29
REQ-403 CA-356724 add random OU to xapi-pool-tls.pem

lippirk:
This seems to fix the intermediate bundle problem

### DIFF
--- a/ocaml/gencert/selfcert.ml
+++ b/ocaml/gencert/selfcert.ml
@@ -137,13 +137,18 @@ let host ~name ~dns_names ~ips pemfile =
   in
   R.failwith_error_msg res
 
+let serial_stamp () = Unix.gettimeofday () |> string_of_float
+
 let xapi_pool ~uuid pemfile =
   let res =
     let* expiration = expire_in_days (365 * 10) in
     let key_length = 2048 in
     let issuer =
       [
-        X509.Distinguished_name.(Relative_distinguished_name.singleton (CN uuid))
+        X509.Distinguished_name.(
+          Relative_distinguished_name.of_list
+            [CN uuid; Serialnumber (serial_stamp ())]
+        )
       ]
     in
     let extensions = X509.Extension.empty in


ocaml/gencert/selfcert.ml
```diff
@@ -137,13 +137,17 @@ let host ~name ~dns_names ~ips pemfile =
   in
   R.failwith_error_msg res
 
+let random_string () = Random.int64 Int64.max_int |> Int64.to_string
```
lindig, line 140: To help anybody who ever looks at these certs, maybe we should use a string that has a bit more information, like `"xapi timestamp %d"` and `%d` is a timestamp or `"xapi random %d"`.

lindig (approved):

lippirk:
I opted for putting the current unix time in the serial number field instead, rather than editing the OU field (organizational unit)

I tested again and it still works as desired

lindig:
What are we doing with https://github.com/xapi-project/xen-api/pull/4479? I think it's worth keeping because the argument is still sound: we can't know how long the reload takes

lippirk:
> What are we doing with #4479? I think it's worth keeping because the argument is still sound: we can't know how long the reload takes

Indeed, there are still opportunities for race conditions!

To be prudent we should add a sleep either side of the reload:
 - the sleep before increase the size of the window for any incoming connections expecting the old certificate
 - the sleep after gives the stunnel server some time to start serving the new certificate, before we begin deleting the old one

ocaml/gencert/selfcert.ml
```diff
@@ -137,13 +137,18 @@ let host ~name ~dns_names ~ips pemfile =
   in
   R.failwith_error_msg res
 
+let serial_stamp () = Unix.gettimeofday () |> string_of_float
```
psafont, line 140: Using floats here seems fine, in the RFCs I couldn't find any requirement for the field relating to the issuer, and for the subject only uniqueness is required per CA:

> The serialNumber attribute type SHALL, when present, be used to
      differentiate between names where the subject field would
      otherwise be identical.  This attribute has no defined semantics
      beyond ensuring uniqueness of subject names.  It MAY contain a
      number or code assigned by the CA or an identifier assigned by a
      government or civil authority.  It is the CA's responsibility to
      ensure that the serialNumber is sufficient to resolve any subject
      name collisions.

psafont (approved):

PR #4487, opened 2021-07-29
CP-36409: fix storage driver domain race conditions on startup

edwintorok:
On startup XAPI still thought that the storage driver domain was running at the time it tried to plug a PBD using a storage driver domain (during XAPI startup we first create storage and then reset power state).
Also plugging a PBD triggers a boot of the storage driver domain, but then XAPI immediately attempts to plug in the PBD, and the (SMAPIv3) plugin might fail to communicate with the driver domain since it hasn't finished booting.

There was already code to wait for a driver domain to boot up, fix it up to use a simpler /HEALTHCHECK GET request instead of xmlrpc to poll for startup completion and call it.


edwintorok:
still has bugs will reopen

PR #4489, opened 2021-08-02
CA-349946: Fix startup race conditions with storage driver domains

edwintorok:
When plugging PBDs on startup have to ensure that:
- we wait for the driver domain to boot up and respond before calling the SMAPIv3 plugin to attach it
- ensure we don't deadlock by having cancelable timeouts on driver domain bootup wait, and doing the PBD plug after the xenops event queue is up

Also clean up after failed or missing VBD unplugs: when the system reboots explicitly transition Dom0 VM through Halted and Running power states to trigger the usual cleanups (detach virtual devices). Although take care not to reset other fields like resident on, etc.

### DIFF
--- a/ocaml/xapi/system_domains.ml
+++ b/ocaml/xapi/system_domains.ml
@@ -187,14 +187,19 @@ let pingable ip () =
 
 let queryable ~__context transport () =
   let open Xmlrpc_client in
-  let rpc =
-    XMLRPC_protocol.rpc ~srcstr:"xapi" ~dststr:"remote_smapiv2" ~transport
-      ~http:(xmlrpc ~version:"1.0" "/")
+  let host = match transport with
+  | Unix _ -> None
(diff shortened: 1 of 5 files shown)


rwdrich (approved):

ocaml/xapi/system_domains.ml
```diff
@@ -187,14 +187,19 @@ let pingable ip () =
 
 let queryable ~__context transport () =
   let open Xmlrpc_client in
-  let rpc =
-    XMLRPC_protocol.rpc ~srcstr:"xapi" ~dststr:"remote_smapiv2" ~transport
-      ~http:(xmlrpc ~version:"1.0" "/")
+  let host = match transport with
+  | Unix _ -> None
```
robhoes, line 191: I don't think that ocamlformat will approve of this.

robhoes (approved):
This looks good to me.

PR #4490, opened 2021-08-05
CP-37692: Introduce RBAC role for client-auth sessions

robhoes:


ocaml/idl/datamodel_roles.ml
```diff
@@ -6,9 +6,12 @@ let role_vm_power_admin = "vm-power-admin"
 let role_vm_admin = "vm-admin"
 let role_vm_operator = "vm-operator"
 let role_read_only = "read-only"
+let role_client_cert = "client-cert"
```
edwintorok, line 9: I think these are roughly sorted by level of privilege, so this should go near pool operator here.

ocaml/idl/datamodel_roles.ml
```diff
@@ -22,6 +25,7 @@ let role_description = [
   role_vm_admin,"The VM Administrator role can manage VMs and templates";
   role_vm_operator,"The VM Operator role can use VMs and interact with VM consoles";
   role_read_only,"The Read-Only role can log in with basic read-only access";
+  role_client_cert,"The Client Certificate role has access to funtionality granted to those who authenticate using a client certificate"
```
edwintorok, line 28: as above, I think the list should be kept sorted.

ocaml/xapi/xapi_session.ml
```diff
@@ -711,15 +711,35 @@ let login_with_password ~__context ~uname ~pwd ~version ~originator =
   let pwd = Bytes.of_string pwd in
   wipe_params_after_fn [pwd] (fun () ->
       (* !!! Do something with the version number *)
-      if Context.preauth ~__context then
+      match Context.preauth ~__context with
+      | Some socket_kind when socket_kind = `root ->
```
edwintorok, line 715: the `when` is redundant here you should be able to write:`Some 'root`, and that will also properly check exhaustiveness.

ocaml/xapi/xapi_session.ml
```diff
@@ -711,15 +711,35 @@ let login_with_password ~__context ~uname ~pwd ~version ~originator =
...
         (* in this case, the context origin of this login request is a unix socket bound locally to a filename *)
         (* we trust requests from local unix filename sockets, so no need to authenticate them before login *)
         login_no_password_common ~__context ~uname:(Some uname) ~originator
           ~host:(Helpers.get_localhost ~__context)
           ~pool:false ~is_local_superuser:true ~subject:Ref.null
           ~auth_user_sid:"" ~auth_user_name:uname ~rbac_permissions:[]
           ~db_ref:None
-      else
+      | Some socket_kind when socket_kind = `client_cert ->
```
edwintorok, line 723: as above

ocaml/xapi/xapi_session.ml
```diff
@@ -711,15 +711,35 @@ let login_with_password ~__context ~uname ~pwd ~version ~originator =
...
+                 (Api_errors.internal_error, [Datamodel_roles.role_client_cert ^ " role not found"])
+              )
+        in
+        let rbac_permissions = Xapi_role.get_permissions_name_label ~__context ~self:role in
+        login_no_password_common ~__context ~uname:(Some uname) ~originator
+          ~host:(Helpers.get_localhost ~__context)
+          ~pool:false ~is_local_superuser:false ~subject:Ref.null
+          ~auth_user_sid:"" ~auth_user_name:uname ~rbac_permissions
+          ~db_ref:None
+      | Some _ | None ->
```
edwintorok: You should remove the `Some` and let OCaml's exhaustiveness analysis prove that you handled both kinds.

ocaml/idl/datamodel_pool.ml
```diff
@@ -712,7 +712,7 @@ open Datamodel_types
         Ref _pool, "self", "The pool";
         Set (Ref _repository), "value", "The set of repositories to be enabled"
       ]
-      ~allowed_roles:_R_POOL_ADMIN
+      ~allowed_roles:_R_POOL_OP
```
edwintorok, line 715: Do we have a baked-in gpg key and we only trust RPMs signed by that gpg key when using the repository?
(I think this is what happens with updates, but I'm not sure what happens with a Yum repository, do we just trust all RPMs inside it by default, or only those signed by us?) 
If not then pool operator could escalate itself to pool admin by creating a Yum repository itself, replacing XAPI rpm with one that disables all authentication.

robhoes: Yes indeed, that is the reason: you can only install signed RPMs, and a pool-operator cannot change which key is accepted.

ocaml/idl/datamodel_roles.ml
robhoes: I wanted to keep the original set of roles together.

ocaml/idl/datamodel_pool.ml
edwintorok: I assume we (will) have a test for that.

edwintorok (approved):

rwdrich (approved):

Issue #4491, opened 2021-08-06
USB+PCI Passthrough combined fails to call usb_reset.py


jeremfg:
This was originally reported [here](https://github.com/xcp-ng/xcp/issues/511), and thus observed on XCP-ng.

----

Environment:
Hardware platform: SuperMicro X10SDV-TLN4F
XCP-ng version: 8.2

I am trying to pass a USB device AND a PCI peripheral to the same VM. Once assignations are properly made and I try to start the VM, the boot fails and XCP-ng Center shows the following error:

> "Failed","Starting VM 'Cosmos'
> Internal error: xenopsd internal error: Call to usb reset failed: Forkhelpers.Spawn_internal_error("usage: usb_reset.py attach [-h] -d DOMID -p PID [-r RESET_ONLY] device\nusb_reset.py attach: error: argument -r: expected one argument\n", "", _)
> Time: 00:00:03","sol","Aug 1, 2021 9:30 PM"

It doesn't seem to matter which USB peripheral(s) I try to pass, but as long as there is one while my GPU is also attached, it fails.
If I remove the GPU and only pass USB devices, the VM boots successfully and the guest is able to use the USB peripheral.

Passthrough for PCI works, Passthrough for USB works. But doing both at the same time fails, preventing the VM from booting.

Based on the error message, it appears that an argument is missing when calling usb_reset.py under the specific condition that a PCI device is also attached. But I have no idea where this gets called.

P.S. The issue is also mentioned [here](https://forum.level1techs.com/t/xcp-pci-passthrough-usb-passthrough-does-not-work-together/162110), albeit for XCP-ng 8.1 in that case.

edwintorok:
Take a look at this commit: https://github.com/xapi-project/xenopsd/commit/4fb4f3bb37e34364c61ef1ec26ac5798ae3e2ae9
It will call `usb_reset.py` with `-r` (when other PCI devices are passed-through), but `usb_reset.py` expects an integer argument for `-r` so that won't work.
This can be fixed either by changing xenopsd to submit an integer along with `-r`, or changing usb_reset.py to accept `-r` as a flag and not require an integer, I'd prefer the latter.

See the documentation for argparse: https://docs.python.org/2.7/library/argparse.html, and modify usb_reset.py on your system to accept a flag instead of integer, I think a `store_true` action instead of `type=int` should work.

If that works then can you submit a pull request to fix scripts/usb_reset.py in this repo?

PR #4492, opened 2021-08-06
Fix issue #4491: USB device reset for Privileged VMs

jeremfg:
When a VM has PCI devices attached, xenopsd will call a usb reset with parameter "-r" for USB passthrough devices to indicate a privileged VM.
xen-api doesn't parse option "-r" properly and is wrongly expecting an integer. This PR fixes this.


psafont:
Do the VMs that gave you trouble boot successfully with this change?

lippirk (approved):

jeremfg:
> Do the VMs that gave you trouble boot successfully with this change?

Yes. I've kept the change locally and use this.

psafont (approved):

PR #4493, opened 2021-08-09
CA-357062: xapi did not successfully start/restart

liulinC:
During xapi startup, main thread start new thread to startup management
interfacea and wait for ip by condition variable between sub-thread
The condition variable should be synced and protected by mutex both
in main thread and sub-thread

### DIFF
--- a/ocaml/xapi/xapi_mgmt_iface.ml
+++ b/ocaml/xapi/xapi_mgmt_iface.ml
@@ -245,6 +245,7 @@ let wait_for_clustering_ip ~__context ~(self : API.ref_Cluster_host) =
   wait_for_ip get_ip is_connected
 
 let on_dom0_networking_change ~__context =
+  Mutex.execute ip_mutex @@ fun () ->
   debug "Checking to see if hostname or management IP has changed" ;
   Helpers.update_pif_addresses ~__context ;
   (* Need to update:
@@ -285,4 +286,4 @@ let on_dom0_networking_change ~__context =
   ) ;
   Helpers.update_domain_zero_name ~__context localhost new_hostname ;
   debug "Signalling anyone waiting for the management IP address to change" ;
-  Mutex.execute ip_mutex (fun () -> Condition.broadcast ip_cond)
+  Condition.broadcast ip_cond


liulinC:
The confidence for this PR is low, see CA-357062 for further details.

lippirk:
> The condition variable should be synced and protected by mutex both
in main thread and sub-thread

I agree with this (this is what the docs say https://ocaml.org/api/Condition.html), however the current implementation already satisfies this as far as I can tell?

liulinC:
see  following code, 
what will happens if main thread is switch out between line 220 and 221,  and switch to sub-thread, where sub-thread will update the condition variable? can current implemenatation handle it?
```217 let wait_for_ip get_ip is_connected =
218   let rec loop () =
219     let ip = match get_ip () with Some x -> x | None -> "" in                                                     
220     let connected = is_connected () in
221     if ip = "" || not connected then (
222       debug "wait_for_ip (IP=%s, connected:%b), waiting" ip connected ;                                           
223       Condition.wait ip_cond ip_mutex ;
224       loop ()
225     ) else
226       ip
227   in
228   Mutex.execute ip_mutex loop

lippirk:
Ah I think I understand what you mean now: the docs talk about `D is a shared data structure, m its mutex, and c is a condition variable`. Previously we were placing the mutex around our condition variable, but not `D`, which in this case is 'networking state in the database'

psafont:
> which in this case is 'networking state in the database'
I don't think this is accurate, the waiting thread protects the inventory file for both the IP and the interface being connected.

liulinC:
> the waiting thread protects the inventory file for both the IP and the interface being connected

I do not understand `the waiting thread protects the inventory file for both the IP and the interface being connected`  well, 
For my finding , see CA-357062 for the details.

edwintorok:
Anything that can change the output of `get_ip` or `is_connected` should be protected by the mutex, such that while the mutex is held inside `loop` it cannot change, and will only change once it sleeps on the Condition variable and releases the mutex.

edwintorok:
It might be better to wrap the mutex + data that it protects + condition var into another abstract type to ensure that modifications always hold the mutex, and that waiting happens correctly. This might be tricky if the change happens outside of XAPI as a side-effect of running another command to bring up networking though.

liulinC:
> Anything that can change the output of `get_ip` or `is_connected` should be protected by the mutex, such that while the mutex is held inside `loop` it cannot change, and will only change once it sleeps on the Condition variable and releases the mutex.

👍 yes, My thinking about the fix is to put `Anything that can change the output of get_ip or is_connected ` into the mutex. 
However, I am not very clear about the codes which will affect the output, so I said the confidence for this PR is low. 
BTW, I will focus on other stuff recent days and will comeback later.

lindig:
I believe this PR should be rejected because it seems to require a different approach

liulinC:
Thanks all for the review, As this ticket is minor and  I got no time on this for this moment, 
Let 's close it for now, and discuss in the ticket.
Again, thanks all.

PR #4494, opened 2021-08-09
REQ-403 verification after pool eject

lippirk:
This PR achieves two objectives:
 - fixes a bug where /var/xapi/verify-certificates was lingering around after pool eject (causing mismatch between db tls verification state + actual state)
 - when we want to 'switch on' the feature by default, reverting f47c98d will ensure that ejected hosts 'come back' with verification enabled

In the future, for new installs, we can also add `FIRSTBOOT_ENABLE_TLS_VERIFICATION='true'` to /etc/xensource-inventory to get verification by default

### DIFF
--- a/scripts/network-init
+++ b/scripts/network-init
@@ -108,6 +108,12 @@ rename_network_label() {
     done
 }
 
+maybe-enable-tls-verification() {
+  verification_enabled="$(${XE} pool-list params=tls-verification-enabled --minimal)"
+  if [ "$FIRSTBOOT_ENABLE_TLS_VERIFICATION" == "true" ] && [ "$verification_enabled" != "true" ] ; then
+      ${XE} pool-enable-tls-verification
+  fi
+}
 
 if [ "$UPGRADE" = "true" ]; then
     rename_network_label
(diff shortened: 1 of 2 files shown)


scripts/network-init
```diff
@@ -108,6 +108,12 @@ rename_network_label() {
     done
 }
 
+maybe-enable-tls-verification() {
```
lindig, line 111: Given that `$XENSOURCE_INVENTORY` is in shell syntax, would not not be easier to `source $XENSOURCE_INVENTORY` and simply look for the value of `FIRSTBOOT_ENABLE_TLS_VERIFICATION`?

psafont: note: `$XENSOURCE_INVENTORY` is sourced in line 11

psafont (approved):

scripts/network-init
```diff
@@ -108,6 +108,12 @@ rename_network_label() {
...
+  verif_enabled="$(${XE} pool-list params=tls-verification-enabled --minimal)"
+  if [ "$FIRSTBOOT_ENABLE_TLS_VERIFICATION" == "true" ] && [ "$verif_enabled" != "true" ] ; then
```
lindig, line 113: I find `verif_enabled` a bit odd as a name. "I shot the verif, I didn't shoot the deputy?"

lindig (approved):

PR #4495, opened 2021-08-10
CP-36863: Expose local YUM repository only on TLS interface

minglumlu:
In one update of continuous update model, a local YUM repository is
enalbed on the pool master host. The pool slaves can access this YUM
repository for updating.

This commit restricts the service port of the local YUM repository on
the pool master to TLS interface only.

### DIFF
--- a/ocaml/xapi/repository.ml
+++ b/ocaml/xapi/repository.ml
@@ -373,7 +373,9 @@ let get_repository_handler (req : Http.Request.t) s _ =
   let open Xapi_stdext_std.Xstringext in
   debug "Repository.get_repository_handler URL %s" req.Request.uri ;
   req.Request.close <- true ;
-  if is_local_pool_repo_enabled () then (
+  if Fileserver.access_forbidden req s then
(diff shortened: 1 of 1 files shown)


lippirk (approved):

ocaml/xapi/repository.ml
```diff
@@ -373,7 +373,9 @@ let get_repository_handler (req : Http.Request.t) s _ =
   let open Xapi_stdext_std.Xstringext in
   debug "Repository.get_repository_handler URL %s" req.Request.uri ;
   req.Request.close <- true ;
-  if is_local_pool_repo_enabled () then (
+  if Fileserver.access_forbidden req s then
```
lippirk, line 376: > This commit restricts the service port of the local YUM repository on
the pool master to TLS interface only.

FYI this change will not block plain http requests on the local unix socket, but I assume that is allowed

psafont (approved):

lindig (approved):

PR #4496, opened 2021-08-12
CP-37866 add Host.tls_verification_enabled field

lindig:
TLS verification is enabled on the pool level but can be disabled
locally (in an emergency) for a host. To provide visibility if this has
happened, this adds field Host.tls_verification_enabled for XenCenter to
use it.

### DIFF
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -916,6 +916,16 @@ let create ~__context ~uuid ~name_label ~name_description ~hostname ~address
   let metrics = Ref.make () in
   make_new_metrics_object metrics ;
   let host_is_us = uuid = Helpers.get_localhost_uuid () in
+  let tls_verification_enabled =
+    match (host_is_us, Db.Pool.get_all ~__context) with
+    | true, _ ->
+        Stunnel_client.get_verify_by_default ()
+    | false, [pool] ->
+        Db.Pool.get_tls_verification_enabled ~__context ~self:pool
+    | _ ->
+        false
+    (* no or multiple pools *)
+  in
   Db.Host.create ~__context ~ref:host ~current_operations:[]
     ~allowed_operations:[]
     ~software_version:(Xapi_globs.software_version ())
@@ -941,7 +951,8 @@ let create ~__context ~uuid ~name_label ~name_description ~hostname ~address
         [0L]
       )
     ~control_domain:Ref.null ~updates_requiring_reboot:[] ~iscsi_iqn:""
-    ~multipathing:false ~uefi_certificates:"" ~editions:[] ~pending_guidances:[] ;
+    ~multipathing:false ~uefi_certificates:"" ~editions:[] ~pending_guidances:[]
+    ~tls_verification_enabled ;
   (* If the host we're creating is us, make sure its set to live *)
   Db.Host_metrics.set_last_updated ~__context ~self:metrics
     ~value:(Date.of_float (Unix.gettimeofday ())) ;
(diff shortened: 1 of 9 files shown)


ocaml/xapi/xapi_host.ml
```diff
@@ -2628,15 +2630,19 @@ let get_sched_gran ~__context ~self =
...
 
 let emergency_reenable_tls_verification ~__context =
   (* NB: Should only be used after running emergency_disable_tls_verification.
      Xapi_pool.enable_tls_verification is not used because it introduces a
      dependency cycle. *)
+  let self = Helpers.get_localhost ~__context in
   Stunnel_client.set_verify_by_default true ;
-  Helpers.touch_file Xapi_globs.verify_certificates_path
+  Helpers.touch_file Xapi_globs.verify_certificates_path ;
+  Db.Host.set_tls_verification_enabled ~__context ~self ~value:false
```
lippirk, line 2645: i think you mean `true` here?

lindig: Good catch

ocaml/xapi/xapi_host.mli
```diff
@@ -514,12 +514,12 @@ val set_sched_gran :
 val get_sched_gran :
   __context:Context.t -> self:API.ref_host -> API.host_sched_gran
 
-val emergency_disable_tls_verification : __context:'a -> unit
+val emergency_disable_tls_verification : __context:Context.t -> unit
```
lippirk, line 517: I'm not so sure about this

``__context:`a`` was communicating that you can't use the `__context` parameter in this function - `emergency_disable_tls_verification` and `emergency_reenable_tls_verification` are emergency comands, so we can expect that the host they are being called on cannot reach the master. In particular we can expect 'Db.X.y ~__context` to hang or fail

ocaml/xapi/xapi_host.ml
lippirk: see my comment below, I think trying to modify the database in an emergency call will often not end well

lippirk: maybe, for the case when the slave will reconnect to the master after running the emergency call, a good place for this code is here: https://github.com/xapi-project/xen-api/blob/62e144780c6f1cccb17fc4decaa010365167b398/ocaml/xapi/xapi.ml#L1158

lindig: So you would suggest a different strategy?  We could not update the database in the emergency command that disables TLS verification but we have to update the database at some point as otherwise there is no point to have this field.

lindig:
I've updated this. When a host disables TLS verification we try to update the DB (but it might fail) and we also report the state of the TLS verification on Xapi startup. That should be good enough to detect that verification is off for a host even when it is on for a pool in general.

ocaml/xapi/xapi_host.ml
```diff
@@ -916,6 +916,12 @@ let create ~__context ~uuid ~name_label ~name_description ~hostname ~address
   let metrics = Ref.make () in
   make_new_metrics_object metrics ;
   let host_is_us = uuid = Helpers.get_localhost_uuid () in
+  let tls_verification_enabled =
+    if host_is_us then
+      Stunnel_client.get_verify_by_default ()
+    else
+      false
```
psafont, line 923: If I'm not mistaken this is use when a host joins a pool. In that case it should be set to the default of the pool instead of just `false` unconditionally.

ocaml/xapi/xapi_host.ml
```diff
@@ -2629,14 +2636,31 @@ let get_sched_gran ~__context ~self =
...
   Stunnel_client.set_verify_by_default false ;
-  Unixext.unlink_safe Xapi_globs.verify_certificates_path
+  Unixext.unlink_safe Xapi_globs.verify_certificates_path ;
+  try
+    (* we update the database on a best-effort basis because we
+       might not have a connection *)
+    let self = Helpers.get_localhost ~__context in
+    Db.Host.set_tls_verification_enabled ~__context ~self ~value:false
+  with e ->
+    error "Failed to update database after TLS verication was disabled: %s"
```
psafont, line 2646: I think downgrading this to warning or even info is warranted since the exception is expected.

ocaml/xapi/xapi_host.ml
```diff
@@ -2629,14 +2636,31 @@ let get_sched_gran ~__context ~self =
...
+      (Printexc.to_string e) ;
+    raise
+      Api_errors.(
+        Server_error
+          ( internal_error
+          , ["Failed to update database after TLS verification disabled"]
```
psafont, line 2652: I think the message must be reworded to reduce user confusion, the exit status won't be 0 (success) even though this is the expected path:
"TLS verification disabled successfully. Failed to contact the coordinator to update the database."

ocaml/xapi/xapi_host.ml
```diff
@@ -2629,14 +2639,34 @@ let get_sched_gran ~__context ~self =
 let emergency_disable_tls_verification ~__context =
   (* NB: the tls-verification state on this host will no longer agree with state.db *)
   Stunnel_client.set_verify_by_default false ;
-  Unixext.unlink_safe Xapi_globs.verify_certificates_path
+  Unixext.unlink_safe Xapi_globs.verify_certificates_path ;
+  try
+    (* we update the database on a best-effort basis because we
+       might not have a connection *)
+    let self = Helpers.get_localhost ~__context in
+    Db.Host.set_tls_verification_enabled ~__context ~self ~value:false
```
lippirk, line 2647: So what actually happens here in practice when you have lost connection to the coordinator and you try to disable verification? I ask because I recall trying to make a similar sort of change in the past, and getting an indefinite hang rather than an exception

lippirk: (I'm thinking about the situation where the user has a genuine networking problem, but attempts to disable tls verification to see if it helps)

lindig: I don't know. If networking breaks down and this is not related because of TLS problems, I think you have all kinds of problems, probably your VMs can't talk to storage anymore. Not sure we can guarantee anything at that point

lindig: I now see failed unit tests - need to investigate

lindig: The failure is coming from this change:

psafont (approved):

lippirk (approved):
I am still a bit wary of this change (it may emergency commands unreliable in when you don't have a session available with the master):

```
-val emergency_reenable_tls_verification : __context:'a -> unit
+val emergency_reenable_tls_verification : __context:Context.t -> unit
...
-val emergency_reenable_tls_verification : __context:Context.t -> unit
+val emergency_reenable_tls_verification : __context:'a -> unit
```

but if you have tested it and it works as expected then that is fine

PR #4497, opened 2021-08-13
CA-357151 REQ-403 ca cert fixes

lippirk:
- ca certs from a joiner must be added to the database
- 'xe pool-certificate-list' should read from db rather than the file system (though these ought to be functionally equivalent)

### DIFF
--- a/ocaml/xapi/cert_distrib.ml
+++ b/ocaml/xapi/cert_distrib.ml
@@ -565,10 +565,32 @@ let collect_ca_certs ~__context ~names =
 
 (* This function is called on the pool that is incorporating a new host *)
 let exchange_ca_certificates_with_joiner ~__context ~import ~export =
+  let module C = Certificates in
   let appliance_certs = List.map WireProtocol.certificate_file_of_pair import in
+  (* perform parsing straight away, so we avoid write invalid certs to disk *)
+  let parsed =
+    List.map
+      (fun WireProtocol.{filename; content} ->
+        let () =
+          if C.(is_unsafe CA_Certificate filename) then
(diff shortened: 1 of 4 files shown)


ocaml/xapi/cert_distrib.ml
```diff
@@ -565,10 +565,32 @@ let collect_ca_certs ~__context ~names =
 
 (* This function is called on the pool that is incorporating a new host *)
 let exchange_ca_certificates_with_joiner ~__context ~import ~export =
+  let open Certificates in
```
lindig, line 568: Another option to avoid long module names would be to `let module C = Certificates in ... C.pem_of_string ..` such that names are still qualified.

ocaml/xapi/cert_distrib.ml
```diff
@@ -565,10 +565,32 @@ let collect_ca_certs ~__context ~names =
...
+        (filename, cert)
+        )
+      appliance_certs
+  in
   Worker.local_write_cert_fs ~__context ApplianceCertificate Merge
     appliance_certs ;
   Worker.local_regen_bundle ~__context ;
+  List.iter
+    (fun (name, cert) ->
+      let (_ : API.ref_Certificate) =
```
lindig, line 588: Why do we need the `let` here?

lindig (approved):

ocaml/xapi/cert_distrib.ml
psafont: My guess would be to ensure the expected type

psafont (approved):

PR #4498, opened 2021-08-13
CP-37898: change default back to PBIS

edwintorok:
There are some outstanding changes needed before we can switch to
winbind by default.
Since the code still supports both PBIS and winbind lets change the
default to PBIS to unblock testing.

### DIFF
--- a/ocaml/xapi/xapi_globs.ml
+++ b/ocaml/xapi/xapi_globs.ml
@@ -901,7 +901,7 @@ let repository_gpgcheck = ref true
 
 type xapi_globs_spec_ty = Float of float ref | Int of int ref
 
-let extauth_ad_backend = ref "winbind"
+let extauth_ad_backend = ref "pbis"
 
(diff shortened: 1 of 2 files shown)


lippirk (approved):

psafont (approved):

lindig (approved):

ocaml/xapi/xapi_globs.ml
```diff
@@ -901,7 +901,7 @@ let repository_gpgcheck = ref true
 
 type xapi_globs_spec_ty = Float of float ref | Int of int ref
 
-let extauth_ad_backend = ref "winbind"
+let extauth_ad_backend = ref "pbis"
```
lindig, line 904: I would be nicer to use a sum type here instead of a string.

liulinC:
If we have to do this, we need to install PBIS binaries also.

PR #4499, opened 2021-08-17
CP-37898: Make winbind encryption types configurable

liulinC:
Trunk limit the encryption methods to `strong` to improve securirty.
However, There may be some existing environment with legacy encryption
like RC4-HMAC. Limit the `strong` encryption may break LCM customers.

This commit set encryption type to `strong` as default
and will configure it to `all` during back port

Details: https://www.samba.org/samba/docs/current/man-html/smb.conf.5.html

### DIFF
--- a/ocaml/xapi/extauth_plugin_ADwinbind.ml
+++ b/ocaml/xapi/extauth_plugin_ADwinbind.ml
@@ -691,7 +691,10 @@ let config_winbind_damon ~domain ~workgroup ~netbios_name =
       ; Printf.sprintf "winbind cache time = %d" !Xapi_globs.winbind_cache_time
       ; Printf.sprintf "machine password timeout = %d"
           !Xapi_globs.winbind_machine_pwd_timeout
-      ; "kerberos encryption types = strong"
+      ; Printf.sprintf "kerberos encryption types = %s"
+          (Kerberos_encryption_types.Winbind.to_string
+             !Xapi_globs.winbind_kerberos_encryption_type
+          )
       ; Printf.sprintf "workgroup = %s" workgroup
       ; Printf.sprintf "netbios name = %s" netbios_name
       ; "idmap config * : range = 3000000-3999999"
--- a/ocaml/xapi/xapi_globs.ml
+++ b/ocaml/xapi/xapi_globs.ml
@@ -917,6 +917,9 @@ let winbind_machine_pwd_timeout = ref (7 * 24 * 3600)
 
 let winbind_update_closest_kdc_interval = ref (3600. *. 24.) (* every day *)
 
+let winbind_kerberos_encryption_type =
+  ref Kerberos_encryption_types.Winbind.Strong
+
 let tdb_tool = ref "/usr/bin/tdbtool"
 
 let sqlite3 = ref "/usr/bin/sqlite3"
@@ -1254,6 +1257,20 @@ let other_options =
     , Arg.Set_string extauth_ad_backend
     , (fun () -> !extauth_ad_backend)
     , "Which AD backend used to talk to DC"
+    )
+  ; ( "winbind_kerberos_encryption_type"
+    , Arg.String
+        (fun s ->
+          Option.iter
+            (fun k -> winbind_kerberos_encryption_type := k)
+            (Kerberos_encryption_types.Winbind.of_string s)
+          )
+    , (fun () ->
+        Kerberos_encryption_types.Winbind.to_string
+          !winbind_kerberos_encryption_type
+        )
+    , "Encryption types to use when operating as Kerberos client \
+       [strong|legacy|all]"
     )
   ; ( "website-https-only"
     , Arg.Set website_https_only
--- a/ocaml/xapi-aux/kerberos_encryption_types.ml
+++ b/ocaml/xapi-aux/kerberos_encryption_types.ml
@@ -1,0 +1,40 @@
+(*
+ * Copyright (C) 2006-2009 Citrix Systems Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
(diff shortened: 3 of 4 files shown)


ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -70,6 +70,34 @@ type domain_info = {
...
+    | "legacy" ->
+        Legacy
+    | "strong" ->
+        Strong
+    | _ as v ->
+        debug "Invalid kerberos type %s use default %s" v (to_string default) ;
+        default
+
+  let get () =
+    !Xapi_globs.winbind_kerberos_encryption_type |> of_string |> to_string
```
lindig, line 98: This should return `t` or otherwise it does not make much sense.

ocaml/xapi/xapi_globs.ml
```diff
@@ -917,6 +917,8 @@ let winbind_machine_pwd_timeout = ref (7 * 24 * 3600)
 
 let winbind_update_closest_kdc_interval = ref (3600. *. 24.) (* every day *)
 
+let winbind_kerberos_encryption_type = ref "strong"
```
lindig, line 920: Why are we strong a string and not the proper type?

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: why it should return `t`, 
I want to use `of_string` and `to_string` to check the string value validation and provide default value if not, does not make sense?

ocaml/xapi/xapi_globs.ml
liulinC: I think the configuration part only take care of the configurations, and let the logic part (winbind) take care of the types.

lindig: By using the proper type you need to validate the incoming string only once. Now anything could be in that string and any client needs to validate it

ocaml/xapi/extauth_plugin_ADwinbind.ml
psafont: > why it should return t,

The client that needs to check properties of the value will be forced to parse again the string. I prefer the interface to return `t` and the user convert to string if needed.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -70,6 +70,34 @@ type domain_info = {
         (* Persist netbios_name to support hostname change *)
 }
 
+module KerberosEncryptionTypes = struct
```
lippirk, line 73: i believe the purpose of this module is to validate `Xapi_globs.winbind_kerberos_encryption_type`, which you do in just one function:

```
let get_encryption_type () =
  let s = !Xapi_globs.winbind_kerberos_encryption_type in
  if List.mem s ["strong"; "all"; "legacy" ] then s else ( debug "some warning" ;  "strong" )
```

i'm not sure i see the benefit of the variant type

liulinC:
Thanks all for the help to review, 

Given we have limited usage of this configuration item, I would like to take @lippirk 's advice to keep it simple, and remove the internal type.
However,  I still want to use module to group the related codes. (give variable meaningful names and avoid define them in function which will run again and again)

lippirk (approved):

lindig:
I think a stronger approach would:

* Use a sum type to represent the state; this would limit the state that can be represented to the legal states
* Use `from_string` to convert a string and validate it
* Use `to_string` to create a string representation

I understand that this is fairly inconsequential for this particular case. I'm not going to approve this but also not going to block this.

liulinC:
> I think a stronger approach would:
> 
> * Use a sum type to represent the state; this would limit the state that can be represented to the legal states
> * Use `from_string` to convert a string and validate it
> * Use `to_string` to create a string representation
> 
> I understand that this is fairly inconsequential for this particular case. I'm not going to approve this but also not going to block this.

Is this the original commit? :cry:

ocaml/xapi/xapi_globs.ml
psafont, line 920: In practice this is a default value. I don't like the default string being in two different modules. Please use a variant type (like before), or find a way to define it only once and refer to that binding from modules that use it.

liulinC:
@lindig @psafont @lippirk 
The configuration file parse only support several simple types like string, float, bool, https://github.com/xapi-project/xcp-idl/blob/master/lib/xcp_service.ml#L55, so I put the variable type parse in the logic module just as initial commit, 
Please let me know if you got better idea.

psafont:
Variants can be read from strings using Arg.String, see https://github.com/psafont/xen-api/commit/40e81edec73d2c965a739709f640fc200f204ea4

ocaml/xapi/xapi_globs.ml
```diff
@@ -1255,6 +1257,12 @@ let other_options =
     , (fun () -> !extauth_ad_backend)
     , "Which AD backend used to talk to DC"
     )
+  ; ( "winbind_kerberos_encryption_type"
+    , Arg.Set_string winbind_kerberos_encryption_type
```
lindig, line 1261: I think this should use `Arg.String`, which calls a function that validates the string and stores a value of the proper type directly such that we never deal with the string.  Most of the problems above come from using `Set_string`.

lindig (approved):

ocaml/xapi/xapi_globs.ml
```diff
@@ -1255,6 +1258,19 @@ let other_options =
     , (fun () -> !extauth_ad_backend)
     , "Which AD backend used to talk to DC"
     )
+  ; ( "winbind_kerberos_encryption_type"
+    , Arg.String
+        (fun s ->
+          winbind_kerberos_encryption_type :=
```
psafont, line 1264: This will set the encryption to Strong if the user get the value wrong. For example if it tries to set it to `lagacy`. While this is safe to do security-wise, it might be surprising to the user.

making `of_string` return an option and using option.iter here would be a noop instead of setting it to default when the value is incorect.

psafont (approved):

ocaml/xapi/xapi_globs.ml
liulinC: We have to make a decision even user input typo, 
will not return an option confuse user?

psafont: the option is not exposed to the user, see https://github.com/psafont/xen-api/commit/40e81edec73d2c965a739709f640fc200f204ea4#diff-587411f4c101b535ce411cd08905e6001eda94b44c4e76c889b19dbf5f1e8dd6R933

psafont (approved):

PR #4500, opened 2021-08-17
Backport CP-35026 to havana

lippirk:
Original PR: https://github.com/xapi-project/xen-api/pull/4234

I tested with the python client:

```
>>> s = XenAPI.Session("https://$IP", ignore_ssl=True)
>>> s.xenapi.login_with_password("root", "$passwd")
```
in xensource log this gives you `[ info|..|484 HTTPS 10.71.8.14->:::80|session.login_with_password D:c4c62071c0e0|xapi]`

PR #4501, opened 2021-08-17
Backport CP-35523 to Havana

lippirk:
Please review this first: https://github.com/xapi-project/xen-api/pull/4500

This is supposed to be the havana equivalent of https://github.com/xapi-project/xen-api/pull/4422


psafont (approved):
The second change looks very different because the change to an if else flow was done beforehand in the previous PR.

Looks like a faithful commit

lindig (approved):

PR #4503, opened 2021-08-20
CA-354374: Update pool_cpuinfo and pool_features after the ejected ho…

minglumlu:
…st having been destroyed

The pool_cpuinfo and pool_feautures should be re-calculated after
the ejeced host having been destroyed in DB. Otherwise, if the
ejected host, e.g. doesn't have a common CPU feature, the pool
will not add the common CPU feature back after ejecting.

This commit fixes the issue on both pool_cpuinfo and pool_features.
Prior to this fix, the re-calculations are done by the host which
is being ejected. In this fix, they are moved to the working scope
of the master host.

Signed-off-by: minglumlu <ming.lu@citrix.com>

### DIFF
--- a/ocaml/xapi/xapi_pool.ml
+++ b/ocaml/xapi/xapi_pool.ml
@@ -1857,12 +1857,7 @@ let eject_self ~__context ~host =
     write_first_boot_management_interface_configuration_file () ;
     Net.reset_state () ;
     Xapi_inventory.update Xapi_inventory._current_interfaces "" ;
-    debug "Pool.eject: deleting Host record (the point of no return)" ;
-    Create_misc.create_pool_cpuinfo ~__context ;
-    (* Update pool features, in case this host had a different license to the
-       		 * rest of the pool. *)
-    Pool_features.update_pool_features ~__context ;
-    (* and destroy my control domains, since you can't do this from the API [operation not allowed] *)
+    (* Destroy my control domains, since you can't do this from the API [operation not allowed] *)
     ( try
         List.iter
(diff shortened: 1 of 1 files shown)


lindig (approved):

ocaml/xapi/xapi_pool.ml
```diff
@@ -1857,12 +1857,7 @@ let eject_self ~__context ~host =
     write_first_boot_management_interface_configuration_file () ;
     Net.reset_state () ;
     Xapi_inventory.update Xapi_inventory._current_interfaces "" ;
-    debug "Pool.eject: deleting Host record (the point of no return)" ;
-    Create_misc.create_pool_cpuinfo ~__context ;
-    (* Update pool features, in case this host had a different license to the
-       		 * rest of the pool. *)
-    Pool_features.update_pool_features ~__context ;
-    (* and destroy my control domains, since you can't do this from the API [operation not allowed] *)
```
lippirk, line 1865: we presumably want both the ejected host and the remaining host to recalculate their pool features? I assume that the ejectee will do this when it reboots

lippirk:
I'm surprised the pool does not call `Xapi_host.destroy` on Pool.eject (that would have fix the pool features as a side eject). However this minimal change seems good to me

lippirk (approved):

ocaml/xapi/xapi_pool.ml
minglumlu: Indeed, the ejectee will do this when it reboots. But this is for its own (new and separate) pool  as it has been separated from the others.

PR #4506, opened 2021-08-25
CA-357025 enable TLS cert checking for pool and WLB together

lindig:
When enabling TLS certificate verification for a pool, also enable it
for a Work Load Balancing (WLB). Note that currently TLS certificate
checking can be disabled per host in an emergency but not for WLB.

### DIFF
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -1013,6 +1013,8 @@ functor
                      (Ref.string_of host)
                ) ;
             Db.Pool.set_tls_verification_enabled ~__context ~self ~value:true ;
+            debug "Enabling TLS verification for Work Load Balancing (WLB)" ;
+            Db.Pool.set_wlb_verify_cert ~__context ~self ~value:true ;
             debug "Pool.enable_tls_verification completed (2/2)"
         )
 


lindig:
There is currently no way to disable WLB cert verification because there is no way to disable TLS verification on the pool level - only on the host level. This could pose a problem. We could change this by tying WLB cert verification to both:

```
  let verify_cert =
    match Db.Pool.get_wlb_verify_cert ~__context ~self:pool with
```

We could extend this to

```
  let verify_cert =
    match Db.Pool.get_wlb_verify_cert ~__context ~self:pool && Stunnel_client.get_verify_by_default () with
```

Now WLB verification would be disables if it is disabled on this host (i.e., the primary host)

psafont:
> There is currently no way to disable WLB cert verification because there is no way to disable TLS verification on the pool level - only on the host level. 

This is the intended behaviour, afaik.

psafont (approved):

rosslagerwall (approved):

kc284 (commented):
The code change looks ok, but is the referenced ticket number the right one? (CA-355657 is about bugtool optimisation).

lindig:
Good catch - that should be CA-357025

kc284 (approved):

PR #4508, opened 2021-08-26
CP-37590: Replace negative language with bond/pool member

danilo-delbusso:
Other changes to negative terms were excluded as not relevant to the ticket.

Please also review [xen-api-sdk#189](https://github.com/xapi-project/xen-api-sdk/pull/189)

### DIFF
--- a/ocaml/idl/datamodel_errors.ml
+++ b/ocaml/idl/datamodel_errors.ml
@@ -228,7 +228,7 @@ let _ =
   error Api_errors.incompatible_pif_properties []
     ~doc:"These PIFs cannot be bonded, because their properties are different." ();
   error Api_errors.slave_requires_management_iface []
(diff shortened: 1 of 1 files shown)


lindig:
I can't see how the change is related to the build failure.

```
52 |     | Rpc.Base64 t -> sprintf "Rpc.Base64 %s" t in
           ^^^^^^^^^^
Error: Unbound constructor Rpc.Base64
```

This is the Stockholm branch and I like to believe that we would have noticed any failure here earlier.

lindig:
We merged previously: https://github.com/xapi-project/xen-api/pull/4504 and it looks to me like tests were not run and this could have missed the problem.

lindig:
Indeed, the test on https://github.com/xapi-project/xen-api/pull/4504 failed but it was not obvious enough.

lindig:
I will merge this because the failure is unrelated.

PR #4518, opened 2021-09-09
CA-356959: Decide user account locked out by lockoutTime

liulinC:
Previous solution decide user account locked out by UserAccountControl
However, it is replaced by lockoutTime, see
https://docs.microsoft.com/en-us/windows/win32/adschema/a-lockouttime
This commit decide a user account is locked out if lockoutTime <> 0

### DIFF
--- a/ocaml/xapi/extauth_plugin_ADwinbind.ml
+++ b/ocaml/xapi/extauth_plugin_ADwinbind.ml
@@ -211,62 +211,69 @@ module Ldap = struct
     end in
     let ldap fmt = fmt |> Printf.ksprintf @@ Printf.sprintf "ldap %s" in
     let* kvps = P.parse_kvp_map stdout <!> ldap "parsing failed '%s'" in
-    let get_string k =
-      match Map.find_opt k kvps with
+    let get_string key =
+      match Map.find_opt key kvps with
       | None ->
-          Error (ldap "missing key '%s'" k)
+          Error (ldap "missing key '%s'" key)
       | Some x ->
           Ok x
     in
-    let get_string_with_default ~k ~default =
-      match get_string k with Ok x -> Ok x | Error _ -> Ok default
+    let get_string_with_default ~key ~default =
+      match get_string key with Ok x -> Ok x | Error _ -> Ok default
     in
-    let get_int of_string k =
-      let* str = get_string k in
+    let get of_string key =
+      let* str = get_string key in
       try Ok (of_string str)
-      with _ -> Error (ldap "invalid value for key '%s'" k)
+      with _ -> Error (ldap "invalid value for key '%s'" key)
     in
-    let get_int_with_default of_string ~k ~default =
-      match get_int of_string k with Ok x -> Ok x | Error _ -> Ok default
+    let get_with_default of_string ~key ~default =
+      match get of_string key with Ok x -> Ok x | Error _ -> Ok default
+    in
+    let windows_nt_time_to_unix_time x =
+      Int64.sub (Int64.div x 10000000L) 11644473600L
     in
     let default = "" in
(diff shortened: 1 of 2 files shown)


lindig (approved):

liulinC:
Thanks @lindig @robhoes  for the help to review,
I have just add a new commit for another minor bug, could you hep to review again.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -226,41 +226,54 @@ module Ldap = struct
       try Ok (of_string str)
       with _ -> Error (ldap "invalid value for key '%s'" k)
     in
+    let get_int_with_default of_string ~k ~default =
```
lindig, line 229: I think `k` is not a good name; I assume it stands for `key` - so I would use that instead.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -226,41 +226,54 @@ module Ldap = struct
...
+      match get_int of_string k with Ok x -> Ok x | Error _ -> Ok default
```
lindig, line 230: The name `get_int` could be misleading because the return type is determined by the first argument. `get_int` is indeed polymorphic and could be called `get` and `get_with_default` instead.

lindig (approved):

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -211,62 +211,69 @@ module Ldap = struct
...
       ; upn
-      ; account_expired
+        (* see https://docs.microsoft.com/en-us/windows/win32/adschema/a-accountexpires *)
+      ; account_expired= is_expired false account_expires
       ; account_disabled= logand user_account_control disabled_bit <> 0l
-      ; account_locked= lockout_time <> 0L
-      ; password_expired= logand user_account_control passw_expire_bit <> 0l
+      ; account_locked=
+          lockout_time <> 0L
+          (* see https://ldapwiki.com/wiki/MsDS-UserPasswordExpiryTimeComputed *)
```
edwintorok, line 275: This link doesn't work, is there a microsoft document that replaces it?

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -211,62 +211,69 @@ module Ldap = struct
...
+    let is_expired zero_expired = function
+      | 0L ->
+          zero_expired
+      | i when i = Int64.max_int ->
           false
-      | i ->
+      | _ as t ->
           let expire_unix_time =
-            windows_nt_time_to_unix_time i |> Int64.to_float
+            windows_nt_time_to_unix_time t |> Int64.to_float
```
edwintorok, line 256: The comparison below relies on the host's clock and the DC's clock to be in sync. We use NTP on our host (though not necessarily with the DC as clock source), so that just leaves a possible DC misconfiguration to cause clocks to be out of sync (which we've seen happen, but in that case the DC's clock should be fixed, if the clock on the host is correct). Might be worth checking for host <-> DC timedrift though in some places and warn in the logs if we detect it. Is there an API to query the DC for the time through LDAP/Kerberos?

edwintorok (approved):

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: I found a way to query DC time before, But I cannot remember where, 

Here is actually a non-critical path, where we display domain user status, not for any critical usage, like permit access to dom0.  
The different timezone may cause the display about user information not right for a while, but will not bring critical issues.

PR #4519, opened 2021-09-10
Fix handling of web-dir parameter

### LINKED ISSUE #4512: The value defined in xapi.conf's `web-dir` is ignored
(tests done on Stockholm release)

I have tried to set the `web-dir` parameter, that defaults to `/opt/xensource/www`, in `xapi.conf`. I thought that a different value would make XAPI serve files from a different directory (e.g. `/var/xapi_http_root`).

However, whatever the value I set, HTTP requests on port 443 still provide files from `/opt/xensource/www`.

Is it the expected behaviour or is there a bug?

stormi:
The web-dir parameter is not taken into account in the definition of
common_http_handlers' get_root handler, because !Xapi_globs.web_dir is
evaluated once and for all when the module loads, before the
configuration file is read.

Turn common_http_handlers into a function so that it is evaluated when
called.

Fixes #4512

Co-authored-by: BenjiReis <benjamin.reis@vates.fr>

### DIFF
--- a/ocaml/xapi/xapi.ml
+++ b/ocaml/xapi/xapi.ml
@@ -778,7 +778,7 @@ let master_only_http_handlers =
   ; ("get_updates", Http_svr.FdIO Xapi_pool.get_updates_handler)
   ]
 
-let common_http_handlers =
+let common_http_handlers () =
   [
     ("get_services_xenops", Http_svr.FdIO Xapi_services.get_handler)
   ; ("put_services_xenops", Http_svr.FdIO Xapi_services.put_handler)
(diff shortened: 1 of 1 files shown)


stormi:
While doing the PR, I wondered whether it would be better to not turn the entire `common_http_handlers` into a function, but rather create a function for get_root and put it instead of `(Fileserver.send_file "/" !Xapi_globs.web_dir))`.

robhoes (approved):

robhoes:
> While doing the PR, I wondered whether it would be better to not turn the entire common_http_handlers into a function, but rather create a function for get_root and put it instead of (Fileserver.send_file "/" !Xapi_globs.web_dir)).

It would still get evaluated immediately when the module loads if `common_http_handlers` is not a function. And I don't think that there is a downside to the current solution.

PR #4520, opened 2021-09-14
CA-358816: Updated subject name in DC does not get updated in pam

liulinC:
The permitted user subject name is configured in pam configuration
file /etc/pam.d/hcp_users
however, it is only updated as event callback like add subject
This commit fix this by update the pam configuration if subject
updated in domain controller


lindig (commented):
I'm not an expert for this code base but it looks good to me.

lindig (approved):

edwintorok (approved):

PR #4524, opened 2021-09-15
Use port 443 for client certificate auth (again)

robhoes:
Partial revert of 6d4537904a49984827cbe88b290ba1e52f4b9c00.

This relies on the "redirect" option of stunnel. We found earlier that
this option was incompatible with the "protocol" option that xapi needs.
However, this was fixed in stunnel v5.60.

### DIFF
--- a/ocaml/xapi-consts/constants.ml
+++ b/ocaml/xapi-consts/constants.ml
@@ -291,8 +291,6 @@ let http_port = default_cleartext_port
 
 let https_port = ref default_ssl_port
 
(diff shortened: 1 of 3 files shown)


lindig (approved):

edwintorok:
Do we need a startup version check here to determine that we indeed have stunnel 5.60? We're currently on an older version.

robhoes:
@edwintorok We don't normally do that at this level, but in the spec file.

edwintorok (approved):

PR #4526, opened 2021-09-16
Backport: Fix handling of web-dir parameter

stormi:
The web-dir parameter is not taken into account in the definition of
common_http_handlers' get_root handler, because !Xapi_globs.web_dir is
evaluated once and for all when the module loads, before the
configuration file is read.

Turn common_http_handlers into a function so that it is evaluated when
called.

Backported from 8909814f812312014b3a386f97d20ff06d342d44

Related to issue #4512

Co-authored-by: BenjiReis <benjamin.reis@vates.fr>

### DIFF
--- a/ocaml/xapi/xapi.ml
+++ b/ocaml/xapi/xapi.ml
@@ -708,7 +708,7 @@ let master_only_http_handlers =
     , Http_svr.BufIO remote_database_access_handler_v2 )
   ]
 
-let common_http_handlers =
+let common_http_handlers () =
   [
     ("get_services_xenops", Http_svr.FdIO Xapi_services.get_handler)
   ; ("put_services_xenops", Http_svr.FdIO Xapi_services.put_handler)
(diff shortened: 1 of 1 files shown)


robhoes (approved):

stormi:
Should I ignore the check-format failure on this lcm branch?

lindig (approved):

lindig:
If you have the branch building, I'd suggest to reformat based on `make format`

stormi:
I suspect the diff would then be far from minimal which is probably not what we want in a backport?

stormi:
But I can also just apply the formatting change that the check format action suggests, if that's preferred.

lindig:
I think we should reformat the code to make it pass the check. Otherwise this check becomes useless.

psafont:
> But I can also just apply the formatting change that the check format action suggests, if that's preferred.

Yes, please

stormi:
All checks now pass.

PR #4527, opened 2021-09-17
CA-358904 REQ-403  cross pool migration must not use cert checking

lindig:
When communicating between two pool, certificate checking can't be used
because this requires a prior key exchange. Add a patch to detect this
case.

Still testing this - so that's why it's a draft.


robhoes:
Looks like you need to update the quality gate:

    ERROR expected 8 verify_cert:None usages, got 9

robhoes (approved):

lindig:
Still testing but tests that failed earlier are currently passing

psafont (approved):

PR #4528, opened 2021-09-17
CA-358898: handle IPv6 state when management disabled

robhoes:
Regression in recent commit (d1fea38).


psafont (approved):
There were three branches in the code, the last one didn't set the variable in the original patch

edwintorok (approved):

PR #4536, opened 2021-09-24
Enable formatting for ocaml code and format the code

psafont:
This was missed during the update to dune 2.0


psafont:
allow_aproximate_merlin is dropped because .merlin files are not generated anymore with newish dune versions. This means the warning doesn't appear anymore when building the project

edwintorok (approved):

robhoes (approved):

PR #4537, opened 2021-09-27
CP-38046: Add token in pool.sync_updates to support repository client…

minglumlu:
… authentication

In handling of pool.sync_updates, some HTTP requests could be sent to
the remote repositories, which could be actually hosted in a CDN, to
fetch updates.

The CDN could authenticate a request by a token. In this way, the
un-authozied users can be prevented from accessing the updates.

The rough process is:
1. a user gets the access permission as a short-lived token;
2. the user calls 'pool.sync_updates' with 2 new added (in this commit)
parameters 'token' and 'token_id';
3. the token is injected to the HTTP requests sent by YUM client as a
HTTP header 'X-Access-Token'. And the 'token_id' is injected as a HTTP
header 'Referer' to indicate the owner who is granted with the token.


robhoes:
I think that a change to `exec_command` in xapi_cli.ml is needed to ensure that the token is not logged to xensource.log as part of the command line logging.

ocaml/idl/datamodel_pool.ml
```diff
@@ -746,9 +746,11 @@ open Datamodel_types
...
       ~in_product_since:rel_next
       ~doc:"Sync with the enabled repository"
-      ~params:[
-        Ref _pool, "self", "The pool";
-        Bool, "force", "If true local mirroring repo will be removed before syncing"
+      ~versioned_params:[
+        {param_type=Ref _pool; param_name="self"; param_doc="The pool"; param_release=next_release; param_default=None};
+        {param_type=Bool; param_name="force"; param_doc="If true local mirroring repo will be removed before syncing"; param_release=next_release; param_default=None};
+        {param_type=String; param_name="token"; param_doc="The token for repository client authentication"; param_release=next_release; param_default=Some (VString "")};
+        {param_type=String; param_name="token_id"; param_doc="The ID of the token"; param_release=next_release; param_default=Some (VString "")}
```
lindig, line 753: Small comment: I think parameter names traditionally use a hyphen (`token-id`) instead of an underscore (`token_id`). On the OCaml level, we still would use an underscore because an identifier can't include a hyphen or minus.

ocaml/xapi/message_forwarding.ml
```diff
@@ -1036,11 +1036,15 @@ functor
...
+      let sync_updates ~__context ~self ~force ~token ~token_id =
+        info
+          "Pool.sync_updates: pool = '%s'; force = %s; token = %s; token_id = \
+           %s"
           (pool_uuid ~__context self)
-          (string_of_bool force) ;
-        Local.Pool.sync_updates ~__context ~self ~force
+          (string_of_bool force)
+          (if token = "" then "" else "(token filtered)")
+          (if token_id = "" then "" else token_id) ;
```
lindig, line 1046: What does this expression do? It seems to me this is simply `token_id`

ocaml/xapi/repository_helpers.ml
```diff
@@ -1431,3 +1431,26 @@ let get_single_enabled_update_repository ~__context =
...
+          close_out tmpch ;
+          f (Some tmpfile)
+          )
+        (fun () -> Unixext.unlink_safe tmpfile)
+  | t, tid when t = "" && tid = "" ->
+      f None
+  | _ ->
+      raise
+        Api_errors.(
+          Server_error (internal_error, ["The token or token_id is empty."])
```
lindig, line 1455: Might want to include `__LOC__` to make debugging easier.

ocaml/idl/datamodel_pool.ml
robhoes: Actually, for the API, we use underscores, but for the CLI we use hyphens.

minglumlu: Thanks. I changed "token_id" to "token-id" in CLI part.

lindig (approved):

PR #4538, opened 2021-09-29
CA-359226 backdate new certificates

lindig:
When creating new certs, we made them valid from the current moment. To
help with testing, we backdate them. During testing we change the date
and that introduces the problem to have certificates that are only start
to become valid in the future. Backdating avoids the problem.


psafont:
I don't think a quirk in the tests should be resolved this way, having such a long validity period is bound to make security scans raise alarms in production systems

PR #4539, opened 2021-09-29
Remove old-style xva import code (finally)

robhoes:
Cleaning up some very old, very dead code ("old-style" means "from before 2007").


robhoes:
Still running some tests, just to be sure.

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -5363,291 +5363,52 @@ let vm_import fd printer rpc session_id params =
...
+              Constants.import_metadata_uri
           else
-            uuids
-        in
-        marshal fd (Command (Print (String.concat "," uuids)))
-    | _ ->
-        failwith "Thin CLI protocol error"
+            Constants.import_uri
+          )
+          (Ref.string_of session_id) (Ref.string_of task_id) full_restore
```
lindig, line 5377: This code assumes that no URL escaping is ever required. Is this true? I'm no too fond of constructing a URI this way and would have preferred constructing a map or assoc list which then some function turns into a URI, taking care of escaping.

ocaml/xapi/import.ml
```diff
@@ -2169,121 +2169,100 @@ let metadata_handler (req : Request.t) s _ =
...
+              header.version config.force vdis
+          in
+          (* CA-48768: Stream_vdi.recv_all only checks for task cancellation
+             						   every ten seconds, so we need to check again now. After this
+             						   point, we disable cancellation for this task. *)
+          TaskHelper.exn_if_cancelling ~__context ;
+          TaskHelper.set_not_cancellable ~__context ;
+          (* Pre-miami GA exports have a checksum table at the end of the export. Check the calculated checksums *)
+          (* against the table here. Nb. Rio GA-Miami B2 exports get their checksums checked twice! *)
+          ( if header.version.export_vsn < 2 then
```
lindig, line 2230: This is some deeply nested code. Could this be made flatter?

ocaml/xapi-cli-server/cli_operations.ml
robhoes: This is not new code. This PR mostly removes and reformats code.

robhoes: This view makes it easier to see: https://github.com/xapi-project/xen-api/pull/4539/files?diff=split&w=1

ocaml/xapi/import.ml
robhoes: I guess the next step might be to zap the code that handles pre-miami exports 🤣

robhoes:
@djs55 @jonludlam Can you believe that we are doing this?!

johnelse:
I always wondered if anyone used this stuff :D

jonludlam:
mumble mumble transfer vm...?

jonludlam:
or perhaps the venerable debian vm?

robhoes:
Jon is right! It was the Transfer VM!

djs55 (approved):
LGTM ;)

djs55:
Did this include the logic to include empty blocks in the output stream in case a long sequence of zeros causes the TCP stream to go idle long enough to be silently terminated by middleboxes? I enjoyed that one.

ocaml/xapi/import.ml
avsm: I think I might have some Rio exports lying around... perhaps I should quietly destroy them ;-)

psafont (approved):
The removal of code looks reasonable, especially how the CLI protocol is handled.

I'm also happy an unverified TLS connection got zapped

robhoes:
@djs55 I think that you mean this one: https://github.com/xapi-project/xen-api/commit/bd54ec89438ed1358ac8ba02bf60110c0f71a313. This code is still there! :)

edwintorok (approved):

edwintorok:
I assume you've tested and regular xva import still works with the new `xe`.

PR #4540, opened 2021-09-30
CA-357785: Stop metrics binaries from logging to stdout

psafont:
Before systemd these binaries were daemonized when running on a system,
and not daemonized when ran for debugging purposes.
Now systemd launches them without parameters to be daemonized making
this code obsolete and make it log on daemon.log as well.

Remove the logic that allows daemonization and stop logging to stdout

This avoids the logspam in daemon.log which is produced because the ocaml libraries doesn't set a stricted loglevel. the log facility local0 already avoid logging debug-level messages to xcp-rrdd-plugins.log using syslog configuration

### DIFF
--- a/ocaml/xcp-rrdd/lib/plugin/rrdd_plugin.ml
+++ b/ocaml/xcp-rrdd/lib/plugin/rrdd_plugin.ml
@@ -138,7 +138,6 @@ functor
       ) else (
         D.info "Not daemonizing .." ;
         Sys.catch_break true ;
-        Debug.log_to_stdout ()
       ) ;
       if !pidfile <> "" then (
         D.debug "Storing process id into specified file .." ;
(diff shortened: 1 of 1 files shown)


ocaml/xcp-rrdd/lib/plugin/rrdd_plugin.ml
```diff
@@ -119,32 +119,7 @@ functor
...
-           ]
-        )
-        (fun _ -> failwith "Invalid argument")
-        (Printf.sprintf "Usage: %s [-daemon] [-pidfile filename]" N.name) ;
-      if !daemonize then (
-        D.info "Daemonizing .." ; Unixext.daemonize ()
-      ) else (
-        D.info "Not daemonizing .." ;
-        Sys.catch_break true ;
-        Debug.log_to_stdout ()
```
robhoes, line 141: This was the culprit.

robhoes (approved):

robhoes:
We don't need `-daemon` and `-pidfile` anymore, so this is cleaning up dead code.

edwintorok (approved):

psafont:
Split the commit into two to separate the cleanup from the fix

PR #4541, opened 2021-09-30
CA-359226 add fist point to backdate new certs during testing

lindig:
When creating new certs, we made them valid from the current moment. To
help with testing, we backdate them based if fist point
/tmp/fist_gencert_backdate is present. During testing we change the date
and that introduces the problem to have certificates that are only start
to become valid in the future. Backdating avoids the problem.

### DIFF
--- a/ocaml/gencert/selfcert.ml
+++ b/ocaml/gencert/selfcert.ml
@@ -20,6 +20,27 @@ module D = Debug.Make (struct let name = "gencert_selfcert" end)
 
 let ( let* ) = Result.bind
 
+let fist_path = "/tmp/fist_gencert_backdate"
+
+let rio_release =
+  match Ptime.of_date (2008, 8, 1) with
+  | None ->
+      assert false
+  | Some date ->
+      date
+
+(** A new cert is valid from: now by default, or the given date. A fist
+file forces the date to [rio_release] for testing *)
+let valid_from' date =
+  match (date, Sys.file_exists fist_path) with
+  | _, true ->
+      D.debug "backdating cert to 2007-08-01 (%s)" fist_path ;
+      rio_release
+  | Some date, false ->
+      date
+  | None, false ->
+      Ptime_clock.now ()
+
 (** initialize the random number generator at program startup when this
 module is loaded. *)
 let () = Mirage_crypto_rng_unix.initialize ()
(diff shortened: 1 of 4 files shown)


ocaml/gencert/selfcert.ml
```diff
@@ -159,8 +165,18 @@ let xapi_pool ~uuid pemfile =
   in
   R.failwith_error_msg res
 
-let xapi_cluster ~cn =
-  let expiration = expire_never () in
+let xapi_cluster ?valid_from ~cn () =
+  let fist = "/tmp/fist_gencert_backdate" in
```
robhoes, line 169: Let's define this just once, on top of this file, so that gencert.ml can get it too?

ocaml/gencert/selfcert.ml
```diff
@@ -20,6 +20,8 @@ module D = Debug.Make (struct let name = "gencert_selfcert" end)
 
 let ( let* ) = Result.bind
 
+let default x = function None -> x | Some x -> x
```
psafont, line 23: This is `Option.value ~default:x`

lindig:
This is now a bit more uniform. All the fist point handling is now internal to selfcert.ml

ocaml/gencert/selfcert.mli
```diff
@@ -1,3 +1,7 @@
+val fist_path : string
```
robhoes, line 1: General comment: we a lacking copyright headers on this file.

robhoes (approved):

psafont (approved):
It looks like it does what it's intended to do.

Do we know if this approach is usable by the testing system? I guess it needs some kind of change to force generating the certs with the ancient dates, as the first certificates are generated before the testing harness can write the file into the fs in a reliable way.

lindig:
I am going to test this manually first. A test would create the fist file and then rotate certs. However, this might not cover all cases because clustering doesn't support cert rotation. We will have to discuss this.

lindig:
Manual testing shows that certs are backdated when the fist point exists.

```
[root@xrtuk-09-14 ~]# xe certificate-list
uuid ( RO)           : 94fc7ae0-6ace-149f-7088-ffae23cc0f16
           type ( RO): host_internal
           name ( RO): 
           host ( RO): 4d622da7-1ef3-4e56-a71c-ba9071f2e182
     not-before ( RO): 20080801T00:00:00Z
      not-after ( RO): 20310929T12:33:13Z
    fingerprint ( RO): F5:61:EE:BD:88:6E:CC:C1:3B:6C:D1:46:DE:26:F3:A8:F7:D6:BC:31:B3:0E:71:90:42:66:07:F6:2F:1A:F2:9D


uuid ( RO)           : f3f28934-a014-cb95-b3e4-00e9d23bb80f
           type ( RO): host
           name ( RO): 
           host ( RO): 4d622da7-1ef3-4e56-a71c-ba9071f2e182
     not-before ( RO): 20080801T00:00:00Z
      not-after ( RO): 20310929T12:33:12Z
    fingerprint ( RO): 2A:5A:59:8F:C7:D6:A3:CA:BB:1B:25:E8:F0:F9:3D:00:1F:5E:D5:78:5F:70:A8:98:6F:EF:FB:80:81:29:92:C1
```

PR #4543, opened 2021-10-08
Enhance XAPI `Task` API

benjamreis:
For now a XAPI client can create a task with `Task.create` and set its status.
But it's not possible to set the results or error_info once a task as finished with success or failure.
Which complicates the reports of the end of the task for the XAPI client.

Also a `Task` created via `Task.create` will always have the pool master as its `resident_on`. It should be settable at task creation and default to `!Xapi_globs.localhost_ref` when not provided.

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -245,6 +245,26 @@ module Task = struct
       ~allowed_roles:_R_READ_ONLY (* POOL_OP can set status for any tasks, others can set status only for owned tasks *)
       ()
 
+  let set_result = call ~flags:[`Session]
+      ~in_oss_since:None
+      ~in_product_since:rel_next
+      ~name:"set_result"
+      ~doc:"Set the task result"
+      ~params:[Ref _task, "self", "Reference to the task object";
+              String, "value", "Task result to be set"]
+      ~allowed_roles:_R_READ_ONLY (* POOL_OP can set result for any tasks, others can set result only for owned tasks *)
+      ()
+
+  let set_error_info = call ~flags:[`Session]
+      ~in_oss_since:None
+      ~in_product_since:rel_next
+      ~name:"set_error_info"
+      ~doc:"Set the task error info"
+      ~params:[Ref _task, "self", "Reference to the task object";
+              Set String, "value", "Task error info to be set"]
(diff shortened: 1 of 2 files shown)


stormi:
So, I'll try to give you more context, to explain why we are trying to manipulate a task object from a XAPI plugin, including its `result` and/or `error_info` fields.

Our host update system differs from that of Citrix Hypervisor - we simply rely on a public `updates` RPM repo with no concept of hotfixes - so we can't use XAPI's implementation of such updates. Back when we started XCP-ng, it was hard for us to contribute to XAPI (technically, and also simply difficult to feel legitimate about it), so we implemented a simple **updater** XAPI plugin in python, which offers a [check_update](https://github.com/xcp-ng/xcp-ng-xapi-plugins/blob/1.6.1-8.2/SOURCES/etc/xapi.d/plugins/updater.py#L101) endpoint and an [update]() endpoint. The methods are currently synchronous, so the caller waits for the result, but means any other client (another XO instance for example) won't be aware that an update is going on, its progress, its status, its result... And the lack of feedback via XAPI leads some users to reboot before the updates finished applying.

Now we would like to improve that by creating a XAPI task for the update process. But as it's implemented in a XAPI plugin, we don't have access to TaskHelper and thus can't fill the `results` (if filling it makes sense... The docs say it's only meant to contain object references?) nor `error_info` fields on the task. This is why we thought it may be good to offer to plugins a way to set those fields. It may not be the better solution. We could implement our update process in XAPI itself, but will you accept to have two competing update systems in it (I'm sure some parts can be shared, but there will be differences too)? Or do you see another solution?

lindig:
Is the issue that the Task APi is incomplete in the sense that some API functionality is available internally but not by external clients? And this PR would level the playing field? I looked at the API and indeed, only a task implemented in OCaml would be able to set results. I am now more sympathetic to this change.

psafont:
Citrix has added a mechanism to maintain yum repositories and update hosts through xapi. I'm not sure about the requirements that you have, I'm sure it's worth taking a look at it instead of hacking on the task API.
http://xapi-project.github.io/xen-api/classes/repository.html

stormi:
Is there a chance we could go from "won't object" to "approved"? :)

As Rob said, there's already `task.set_progress` and `task.set_status` not it's already half open to modification of tasks through the API.

robhoes (approved):
Alright then :)

lindig (approved):

PR #4545, opened 2021-10-08
CA-359214: Only restart stunnel if the config file has changed

robhoes:
This avoids unnecessarily breaking connections.

This addresses a regression introduced in d1fea38, which caused stunnel
to be restarted no matter whether a new management server was started.

The patch renames Xapi_stunnel_server.restart to .sync to reflect that
the function no longer always restarts stunnel. Also the logging is
improved.


lindig:
Stunnel can also reload a configuration. Could this be used instead?

robhoes:
> Stunnel can also reload a configuration. Could this be used instead?

I had consider that, but this function is also called when xapi starts up, and then stunnel may not yet be running, or it may be better to really restart it and close any remaining open connections. I'm not sure whether the latter is really needed, but it would need more testing if we want to change that. The current patch restores the behaviour from before last month.

PR #4546, opened 2021-10-11
CA-341715: control-domain-params-init: skip on upgrade

lindig:
This makes it behave the same as network-init and other *-init services,
as well as the old firstboot scripts.

Backport of
  * a014ba952
  * be9ee983a


lindig:
My bad - indeed code was lost in an incomplete rebase.

PR #4550, opened 2021-10-13
stunnel/gencert services: use Wants rather than Requires

robhoes:
This avoids stunnel to go down if forkexecd is restarted.


edwintorok:
CA-359844 if you want a ticket number

PR #4552, opened 2021-10-19
Merge xenopsd

robhoes:
The big one. Xenopsd is coming home.

PR #4553, opened 2021-10-19
specsavers: merge xapi-networkd in

psafont:
Opening the draft PR to be able to test it. Waiting on https://github.com/xapi-project/xen-api/pull/4552 before it's ready to merge

### DIFF
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -2392,6 +2392,7 @@ functor
       let migrate_send ~__context ~vm ~dest ~live ~vdi_map ~vif_map ~options
           ~vgpu_map =
         info "VM.migrate_send: VM = '%s'" (vm_uuid ~__context vm) ;
+        let source_host = Db.VM.get_resident_on ~__context ~self:vm in
         let local_fn =
           Local.VM.migrate_send ~vm ~dest ~live ~vdi_map ~vif_map ~vgpu_map
             ~options
@@ -2464,7 +2465,6 @@ functor
         in
         ( match migration_type with
         | `Live_intrapool host ->
-            let source_host = Db.VM.get_resident_on ~__context ~self:vm in
             let message_body =
               Printf.sprintf
                 "VM '%s' (uuid: %s) migrated from host '%s' (uuid: %s) to host \
--- a/scripts/backup-metadata-cron
+++ b/scripts/backup-metadata-cron
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-if [ -e @ETCDIR@/pool.conf ]; then
-    if [ "$(cat @ETCDIR@/pool.conf)" = "master" ]; then
+if [ -e @ETCXENDIR@/pool.conf ]; then
+    if [ "$(cat @ETCXENDIR@/pool.conf)" = "master" ]; then
         "@OPTDIR@/bin/xe-backup-metadata" -c >/dev/null 2>&1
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -139,6 +142,11 @@ install: build doc sdk doc-json
 	install -m 755 _build/install/default/bin/get_vhd_vsize    $(DESTDIR)/usr/libexec/xapi/get_vhd_vsize
 	install -m 755 ocaml/vhd-tool/scripts/get_nbd_extents.py   $(DESTDIR)$(LIBEXECDIR)/get_nbd_extents.py
 	install -m 644 ocaml/vhd-tool/scripts/python_nbd_client.py $(DESTDIR)$(LIBEXECDIR)/python_nbd_client.py
+# xcp-networkd
+	install -m 755 _build/install/default/bin/xapi-networkd              $(DESTDIR)/usr/sbin/xcp-networkd
+	install -m 755 _build/install/default/bin/networkd_db                $(DESTDIR)/usr/bin/networkd_db
+	install -m 644 _build/default/ocaml/networkd/networkd/xcp-networkd.1 $(DESTDIR)/usr/share/man/man1/xcp-networkd.1
+	install -m 644 scripts/xcp-networkd.conf                             $(DESTDIR)/etc/xcp-networkd.conf
```
robhoes: I think it's better to install scripts/xcp-networkd.conf in scripts/Makefile, otherwise it will get confusing.

robhoes (approved):

ocaml/networkd/networkd_db/dune
```diff
@@ -2,6 +2,7 @@
  (name networkd_db)
  (public_name networkd_db)
  (package xapi-networkd)
+ (modes byte exe)
```
edwintorok: We probably don't want the `byte` modes, we only deploy the native executables, but it doesn't really hurt to have them here (as long as we don't accidentally end up with a bytecode linked with ocamlrun instead of an ocamlopt compiled one, but we can check for that by running `strings -a` on the binary and looking for `caml_interp` I think)

Makefile
```diff
@@ -143,9 +143,9 @@ install: build doc sdk doc-json
 	install -m 755 ocaml/vhd-tool/scripts/get_nbd_extents.py   $(DESTDIR)$(LIBEXECDIR)/get_nbd_extents.py
 	install -m 644 ocaml/vhd-tool/scripts/python_nbd_client.py $(DESTDIR)$(LIBEXECDIR)/python_nbd_client.py
 # xcp-networkd
-	install -m 755 _build/install/default/bin/xapi-networkd              $(DESTDIR)/usr/sbin/xcp-networkd
-	install -m 755 _build/install/default/bin/networkd_db                $(DESTDIR)/usr/bin/networkd_db
-	install -m 644 _build/default/ocaml/networkd/networkd/xcp-networkd.1 $(DESTDIR)/usr/share/man/man1/xcp-networkd.1
+	install -m 755 _build/install/default/bin/xapi-networkd         $(DESTDIR)/usr/sbin/xcp-networkd
+	install -m 755 _build/install/default/bin/networkd_db           $(DESTDIR)/usr/bin/networkd_db
+	install -m 644 _build/default/ocaml/networkd/bin/xcp-networkd.1 $(DESTDIR)/usr/share/man/man1/xcp-networkd.1
```
edwintorok, line 172: we could install all of these with dune, but lets do that cleanup once everything is merged into xapi.

edwintorok (approved):

PR #4560, opened 2021-10-28
Updating README with step-by-step build instructions v2

Cameronsplaze:
Now with signed commits!

A continuation of #4556.

> I would prefer for lines to not exceed 80 characters and therefore makes this work well in tools like `less`. I understand that markdown rendered into HTML does take care of this but I believe these files are also viewed inside shells.

@lindig I don't have a way to count them, but I'm guessing the `PACKAGES=...` line is the main offender? I'd recommend keeping it copy/paste-able at least. This is the first solution that comes to mind, but maybe there's multi-line in bash that'd still work with `opam install $PACKAGES`?

```bash
PACKAGES="First set here"
PACKAGES="$PACKAGES second set here"
...
```

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -18,10 +18,67 @@ Xapi is the main component produced by the Linux Foundation's
 Build and Install
 -----------------
 
-The build install instructions are currently being written. The Xapi
-Project contains a large list of dependencies and sub-projects, which
-are actually quite difficult to build independently. To build xen-api
-from source, we recommend using [opam](https://opam.ocaml.org/doc/Manual.html) with the [xs-opam](https://github.com/xapi-project/xs-opam) remote (detailed explanation in [readme](https://github.com/xapi-project/xs-opam/blob/master/README.md)).
+To build xen-api from source, we recommend using [opam](https://opam.ocaml.org/doc/Manual.html) with the [xs-opam](https://github.com/xapi-project/xs-opam) repository (farther explanation in its [readme](https://github.com/xapi-project/xs-opam/blob/master/README.md)).
+
+1) Install `opam` and `git` with your package manager.
+
+2) Clone this repo, and work from it's root.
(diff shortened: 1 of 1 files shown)


README.markdown
```diff
@@ -18,10 +18,67 @@ Xapi is the main component produced by the Linux Foundation's
...
+
+    ```bash
+    PACKAGES="xapi-cli-protocol xapi-client xapi-consts xapi-database xapi-datamodel xapi-types xapi xe xen-api-sdk xen-api-client xen-api-client-lwt xen-api-client-async xapi-rrdd xapi-rrdd-plugin xapi-rrd-transport xapi-rrd-transport-utils rrd-transport rrdd-plugin rrdd-plugins rrddump gzip http-svr pciutil safe-resources sexpr stunnel uuid xapi-compression xml-light2 zstd vhd-tool xs-toolstack"
+
+    # NOT needed with opam>=2.1.0) Install all the dependances (Including OS):
+    opam --yes depext --yes -u $PACKAGES # The first '--yes' is to install depext itself
+    # Install the Packages finally:
+    opam install $PACKAGES --yes --deps-only --with-test -v
+    # Update the current switch. (You're already on the correct one, just refresh it).
+    eval $(opam env)
```
psafont, line 70:
```suggestion
    # NOT needed with opam>=2.1.0) Install all the dependances (Including OS):
    opam -depext --yes -u xs-toolstack
    # Build and install all the xen-api daemons and their dependencies:
    opam install --with-test xs-toolstack
```
This is works as well and won't need to maintain the list of packages.

It works because xs-toolstack is an empty package that depends on all the code generated here

psafont (approved):
There is a change I'd like to do, it can be done later, this is already improving on what we have

lindig (approved):

PR #4566, opened 2021-11-05
CP-38309 make TLS more explicit in clusterd interface

lindig:
We want to move to a more explicit interface with xapi clusterd. In
particular, when certificate checking is enabled and what certificates
it is using should be more obvious.

This leaves certificate checking still *off* by default. Expect the CI to tail because this depends on changes in https://github.com/xapi-project/message-switch/pull/74


edwintorok (approved):
To be merged together with message-switch PR

psafont (approved):

PR #4568, opened 2021-11-09
CA-359869: Make Sysfs.list robust against disappearing devices

robhoes:


lindig (approved):

ocaml/networkd/lib/network_utils.ml
```diff
@@ -198,7 +198,9 @@ module Sysfs = struct
         false
 
   let list () =
-    let is_dir name = Sys.is_directory ("/sys/class/net/" ^ name) in
+    let is_dir name =
+      try Sys.is_directory ("/sys/class/net/" ^ name) with _ -> false
+    in
     Sys.readdir "/sys/class/net"
     |> Array.to_list
     |> List.filter (fun name -> is_dir name && is_ether_device name)
```
lindig, line 206: An alternative would be to check using `Sys.file_exists` in the predicate in addition. `file_exists` is true for directories, too.

robhoes: I think here we want _only_ directories.

lindig (approved):

ocaml/networkd/lib/network_utils.ml
lindig: My suggestion was to check for: exists && is directory && is_ether_devive.

robhoes: Oh you said _in addition_ (to the directory check)...

edwintorok (approved):
I see that almost all callers of read_one_line need to wrap it in a try/with. Would it be better if read_one_line returned a Result instead to force the caller into handling the error condition?
In fact isn't `is_dir` redundant here? If it is not a dir then trying to open a file inside it will raise ENOENT and cause is_ether_device to return false anyway.

lindig:
>  In fact isn't `is_dir` redundant here? If it is not a dir then trying to open a file inside it will raise ENOENT and cause is_ether_device to return false anyway.

Good observation. Historically `is_ether_device` was introduced after `is_dir`.

robhoes:
Yes, strictly speaking it appears to be redundant (now). However, it may be good to have the explicit check there just in case someone "optimises" or "refactors" it later for some reason.

PR #4569, opened 2021-11-10
CA-360577: Add RBAC checking for client cert HTTPs requests

minglumlu:
When a HTTPs request arrived through client cert path, a 'client_cert' role
has been assigned to this request. There is no need to do authentication again
based on credentials in HTTP headers.

This commit adds RBAC checking for this kind of HTTPs requests. In details,
the `Xapi_session.login_with_password` is re-used to create a session so that
`Rbac.check` can be re-used as well to do the RBAC checking with the session.

### DIFF
--- a/ocaml/xapi/xapi_http.ml
+++ b/ocaml/xapi/xapi_http.ml
@@ -85,6 +85,12 @@ let rbac_audit_params_of (req : Request.t) =
     (fun (n, v) (acc_n, acc_v) -> (n :: acc_n, Rpc.String v :: acc_v))
     all ([], [])
 
+let create_session_for_client_cert req s =
+  let __context = Context.make ~origin:(Http (req, s)) "client_cert" in
+  Xapi_session.login_with_password ~__context ~uname:"" ~pwd:""
+    ~version:Datamodel_common.api_version_string
+    ~originator:Constants.xapi_user_agent
+
 let assert_credentials_ok realm ?(http_action = realm) ?(fn = Rbac.nofn)
     (req : Request.t) ic =
   let http_permission = Datamodel.rbac_http_permission_prefix ^ http_action in
@@ -155,7 +161,15 @@ let assert_credentials_ok realm ?(http_action = realm) ?(fn = Rbac.nofn)
         raise (Failure (Printf.sprintf "Unknown authorization header: %s" x))
     | None, None, None ->
         debug "No header credentials during http connection to %s" realm ;
-        raise (Http.Unauthorised realm)
+        let session_id =
+          try create_session_for_client_cert req ic
+          with _ -> raise (Http.Unauthorised realm)
+        in
+        Xapi_stdext_pervasives.Pervasiveext.finally
+          (fun () -> rbac_check session_id)
+          (fun () ->
+            try Client.Session.logout inet_rpc session_id with _ -> ()
+            )
 
 let with_context ?(dummy = false) label (req : Request.t) (s : Unix.file_descr)
     f =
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_http.ml
```diff
@@ -199,7 +213,11 @@ let with_context ?(dummy = false) label (req : Request.t) (s : Unix.file_descr)
         | None, None, Some (Http.UnknownAuth x) ->
             raise (Failure (Printf.sprintf "Unknown authorization header: %s" x))
         | None, None, None ->
-            raise (Http.Unauthorised label)
+            let session_id =
```
lindig, line 216: While this addition is probably fine by itself, I feel this entire function is increasingly getting out of shape. We have deeply nested pattern matching and exception handling and I believe this is in need of a rethink.

ocaml/xapi/xapi_http.ml
```diff
@@ -155,7 +161,15 @@ let assert_credentials_ok realm ?(http_action = realm) ?(fn = Rbac.nofn)
         raise (Failure (Printf.sprintf "Unknown authorization header: %s" x))
     | None, None, None ->
         debug "No header credentials during http connection to %s" realm ;
-        raise (Http.Unauthorised realm)
+        let session_id =
+          try create_session_for_client_cert req ic
```
robhoes, line 165: Let's add a comment and some debug logging here to avoid confusion.
```suggestion
          (* The connection may have been authenticated using a client cert.
               If so, then following call confirms this. *)
          try
            create_session_for_client_cert req ic;
            debug "This HTTP connection was pre-authenticated using a client cert"
```

ocaml/xapi/xapi_http.ml
```diff
@@ -199,7 +213,11 @@ let with_context ?(dummy = false) label (req : Request.t) (s : Unix.file_descr)
         | None, None, Some (Http.UnknownAuth x) ->
             raise (Failure (Printf.sprintf "Unknown authorization header: %s" x))
         | None, None, None ->
-            raise (Http.Unauthorised label)
+            let session_id =
+              try create_session_for_client_cert req s
```
robhoes, line 217: As above, please add comment/debug.

ocaml/xapi/xapi_http.ml
```diff
@@ -199,7 +213,11 @@ let with_context ?(dummy = false) label (req : Request.t) (s : Unix.file_descr)
...
+              with _ -> raise (Http.Unauthorised label)
+            in
+            (session_id, false)
```
robhoes, line 220: Here, `must_logout` should be `true`, as no one else will logout (the client does not get this session id).

robhoes (changes requested):
Generally fine, but a small update is needed.

ocaml/xapi/xapi_http.ml
minglumlu: Yes. Thanks. Fixed it.

ocaml/xapi/xapi_http.ml
minglumlu: I appended a refine commit for this. Hope this would be better.

ocaml/xapi/xapi_http.ml
```diff
@@ -161,10 +161,14 @@ let assert_credentials_ok realm ?(http_action = realm) ?(fn = Rbac.nofn)
...
     | None, None, None ->
         debug "No header credentials during http connection to %s" realm ;
+        (* The connection may have been authenticated using a client cert.
+         * If so, then following call confirms this. *)
         let session_id =
           try create_session_for_client_cert req ic
           with _ -> raise (Http.Unauthorised realm)
         in
+        debug "Created a session %s for a HTTP(s) request"
+          (Context.trackid_of_session (Some session_id)) ;
```
robhoes: The trackid may already get logged from elsewhere. I wanted to explicitly call out the client cert aspect here.

robhoes (approved):

psafont (approved):

PR #4571, opened 2021-11-11
CA-360485: Fix SR-IOV capability detection

robhoes:
Always require that the device has declared SR-IOV support in sysfs.
This is independent of the way that SR-IOV would eventually be enabled
for the NIC, which may be through either of the "sysfs" or "modprobe"
methods, as before.

### DIFF
--- a/ocaml/networkd/bin/network_server.ml
+++ b/ocaml/networkd/bin/network_server.ml
@@ -106,16 +106,19 @@ module Sriov = struct
       Sysfs.get_driver_name_err dev >>= fun driver ->
       Modprobe.get_config_from_comments driver |> Modprobe.get_maxvfs driver
     and maxvfs_sysfs = Sysfs.get_sriov_maxvfs dev in
-    let is_support =
+    let supported =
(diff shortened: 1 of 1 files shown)


robhoes:
@idruzhinin I think that this is the safest way to fix the problem. What do you think?

idruzhinin:
IMO get_config_from_comments  was always wrong and no longer needed. Max number of VFs should be taken from sysfs file.

robhoes:
Unfortunately, it is a documented manual way to set maxvfs (setting `VFs-maxvfs-by-user`), and it is used by the "enable" function as well. The fix the current bug, I would preserve this. We can have a separate discussion about the future modprobe thing.

idruzhinin:
Maybe we have to keep it - I'm looking at the driver now and it does not allow setting max_vfs higher than 7 while the hardware says it's capable of 8.

robhoes:
Igor helped to verify the fix, so I've now declared it "ready".

psafont (approved):

edwintorok (approved):

PR #4574, opened 2021-11-15
Use stunnel proxy to access internal YUM repo

minglumlu:
The pool members will access the internal YUM repo which is hosted on the pool
master host.

This commit aims to reuse intra-pool server certificate checking to verify
that the YUM repo is a valid one. Consequently, a stunnel client proxy will be
started on-demand to handle the TLS connection from a pool member to the pool
master, especially the server certificate checking.


ocaml/libs/stunnel/stunnel.ml
```diff
@@ -287,6 +296,50 @@ let disconnect ?(wait = true) ?(force = false) x =
...
+  in
+  Xapi_stdext_pervasives.Pervasiveext.finally
+    (fun () -> output_string tmpch config ; close_out tmpch ; f tmpfile)
+    (fun () -> Unixext.unlink_safe tmpfile)
+
+let with_client_proxy verify_cert r_host r_port l_host l_port f =
+  let config =
+    config_file ~accept:(Some (l_host, l_port)) verify_cert r_host r_port
+  in
+  with_client_config_file ~config @@ fun config_file_path ->
```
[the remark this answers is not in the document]
minglumlu: It's possible. But my first impression was it might increase the number of branches in the `with_attempt_one_connect`. Lots of bits in the function are handling the socketpair which is not necessary in this task.
But I would like to have a try to see what it would be in reusing way.

ocaml/libs/stunnel/stunnel.ml
```diff
@@ -287,6 +296,50 @@ let disconnect ?(wait = true) ?(force = false) x =
    exception instead *)
 exception Stunnel_initialisation_failed
 
+let with_client_config_file ~config f =
+  let tmpfile, tmpch =
+    Filename.open_temp_file ~mode:[Open_text] "stunnel-client-proxy-" ".conf"
+  in
+  Xapi_stdext_pervasives.Pervasiveext.finally
```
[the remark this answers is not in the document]
lindig: I am not sure that this protects against leaking `tmpch`. If an exception occurs before `close_out`, it will leak.

ocaml/libs/stunnel/stunnel.ml
```diff
@@ -287,6 +296,50 @@ let disconnect ?(wait = true) ?(force = false) x =
...
+    (fun () -> output_string tmpch config ; close_out tmpch ; f tmpfile)
+    (fun () -> Unixext.unlink_safe tmpfile)
+
+let with_client_proxy verify_cert r_host r_port l_host l_port f =
```
lindig, line 307: I assume the `l` and `r` prefix should denote `local` and `remote` - I'd prefer longer names for parameters and maybe even named parameters to avoid mixing them up at the call site.

ocaml/libs/stunnel/stunnel.mli
```diff
@@ -87,3 +87,12 @@ val move_out_exn : t -> t
 val with_moved_exn : t -> (t -> 'd) -> 'd
 
 val safe_release : t -> unit
+
+val with_client_proxy :
+     verification_config option
+  -> string
```
lindig, line 93: This needs named parameters to avoid mixing them up.

ocaml/libs/stunnel/stunnel.ml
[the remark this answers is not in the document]
psafont: With `Fun.protect` exceptions cannot be raised in the `~finally` function, stdext's finally protects against this.

ocaml/libs/stunnel/stunnel.ml
minglumlu: > I am not sure that this protects against leaking tmpch. If an exception occurs before close_out, it will leak.
The `finally` can handle the exception case.

minglumlu:
Hi reviewers,
I'm sorry I force pushed a commit. This is because the new change is quite different with the previous one. Per @robhoes 's comments and detailed suggestion in a separate message, the new change is to re-use the `with_attempt_one_connect` (renamed to `attempt_one_connect`).
Your reviews will be appreciated!

robhoes (approved):
I think this looks right!

ocaml/libs/stunnel/stunnel.ml
```diff
@@ -345,39 +344,51 @@ let with_attempt_one_connect ?unique_id ?(use_fork_exec_helper = true)
                    )
                  (path :: args)
               )
-          ) ;
+        in
         Unixfd.safe_close config_out ;
-        Unixfd.safe_close data_in ;
-        let config = config_file verify_cert host port in
+        (* The sock_of_stunnel has been passed to stunnel process. Close it in XAPI *)
+        ignore (Option.map (fun s -> Unixfd.safe_close s) sock_of_stunnel) ;
```
edwintorok, line 350: This could use Option.iter instead of map+ignore

ocaml/libs/stunnel/stunnel.ml
```diff
@@ -345,39 +344,51 @@ let with_attempt_one_connect ?unique_id ?(use_fork_exec_helper = true)
...
-        write_to_log "stunnel start" ;
-        t.logfile <- log
-      ) ;
-      f t
+  | Forkhelpers.Success (log, pid) ->
+      if extended_diagnosis then write_to_log "stunnel start" ;
+      (pid, log)
   | Forkhelpers.Failure (log, exn) ->
       write_to_log ("stunnel abort: Log from stunnel: [" ^ log ^ "]") ;
-      disconnect t ;
```
edwintorok, line 380: I think having the `disconnect t` here was important to avoid file descriptor leaks, otherwise the error paths in `start` need to handle disconnection which is more error prone because as we change the code and add more error paths we might forget to free it.
Was there a reason the `t` here had to be moved?

edwintorok (changes requested):

ocaml/libs/stunnel/stunnel.ml
minglumlu: It is moved into the `let start sock_of_stunnel config` function as `disconnect_with_pid`.
And the unix socket will be closed by `Unixfd.with_socketpair`.
The reason is just that the `t` is not applicable to the client proxy case.

ocaml/libs/stunnel/stunnel.ml
```diff
@@ -288,54 +299,42 @@ let disconnect ?(wait = true) ?(force = false) x =
...
     Forkhelpers.with_logfile_fd "stunnel" ~delete:(not extended_diagnosis)
       (fun logfd ->
         let path = stunnel_path () in
-        let fdops =
-          [
-            Unsafe.Dup2 (Unixfd.(!data_in), Unix.stdin)
-          ; Unsafe.Dup2 (Unixfd.(!data_in), Unix.stdout)
-          ; Unsafe.Dup2 (logfd, Unix.stderr)
-          ]
+        let fds_needed, fdops, sock =
```
edwintorok, line 318: `sock` is a raw fd here, I think it might be better if we keep it undereferenced and dereference it with `Unixfd.(!)` only at the callsite. It is only used below in Forkhelpers.safe_close_and_exec currently, but in the future we might accidentally attempt to use it elsewhere (and it will get closed by the with_socketpair function, so we might get EBADF in that case).
I'm not sure what we could do at the compiler/typesystem level to ensure the fd is used correctly (OCaml doesn't have linear types built-in), I think for now we relied on this convention of dereferencing the fd wrapper as late as possible and in the smallest scope possible.

ocaml/libs/stunnel/stunnel.ml
```diff
@@ -288,54 +299,42 @@ let disconnect ?(wait = true) ?(force = false) x =
...
-              FEFork
-                (Forkhelpers.safe_close_and_exec
-                   (Some Unixfd.(!data_in))
-                   (Some Unixfd.(!data_in))
-                   (Some logfd) configs path args
-                )
+        let pid =
+          if use_fork_exec_helper || Option.is_none sock_of_stunnel then
+            FEFork
+              (Forkhelpers.safe_close_and_exec sock sock (Some logfd) configs
```
edwintorok, line 335: This might need an `Option.map Unix.fd(!)` if we remove `sock` from above.

edwintorok (approved):
Now that forkexec is part of xapi we could make it take a Unixfd and simplify this code. That can be done in a separate PR.

Issue #4567, opened 2021-11-07
DUPLICATE_VM error, impossible to find the cause


olivierlambert:
Hello!

We got an interesting case of someone having issues to migrate a VM to another pool, failing with `xe` with:
```
Cannot restore this VM because it would create a duplicate
```
Obviously, same result with Xen Orchestra, it's not a XAPI client issue, but clearly something on XAPI side itself. However, despite trying to find duplicate data between VMs, we couldn't find anything (except `mac_seed` but I don't think that's the culprit).

More details are provided here: https://xcp-ng.org/forum/topic/5151/migrating-vm-fails-with-duplicate_vm-error-part2

olivierlambert:
So after some tests, we confirm that this error is display if you have the same `mac_seed` between 2 VMs.

Is that expected? To me, a MAC seed is just… a seed.

undu:
The VIF mac's are generated in a deterministic manner from the mac seed and the device number, this is why the seeds are forced to be unique

psafont:
I think so, reopen if you think it needs more fixing (I don't think this has been marked for backporting)

PR #4576, opened 2021-11-17
xapi/import: report duplicate mac seeds on import as such

psafont:
Reporting duplicate vm can confuse users, see #4567

### DIFF
--- a/ocaml/xapi/import.ml
+++ b/ocaml/xapi/import.ml
@@ -247,10 +247,7 @@ let assert_can_restore_backup ~__context rpc session_id (x : header) =
       List.iter
         (fun (mac', vm') ->
           if mac = mac' && not (is_compatible vm vm') then
-            raise
-              (Api_errors.Server_error
-                 (Api_errors.duplicate_vm, [vm'.API.vM_uuid])
-              )
+            raise Api_errors.(Server_error (duplicate_mac_seed, [mac']))
           )
         existing_vms
       )


ocaml/xapi/import.ml
```diff
@@ -247,10 +247,7 @@ let assert_can_restore_backup ~__context rpc session_id (x : header) =
       List.iter
         (fun (mac', vm') ->
           if mac = mac' && not (is_compatible vm vm') then
-            raise
-              (Api_errors.Server_error
-                 (Api_errors.duplicate_vm, [vm'.API.vM_uuid])
-              )
+            raise Api_errors.(Server_error (duplicate_mac_seed, [mac']))
```
robhoes, line 250: It also tests for other incompatibilities with `is_compatible`, so maybe split the cases and raise separate errors.

psafont: The check is intertwined, (it's an `&&`, not `||`), there's something that's missing here from the report.

undu: `is_compatible` checks whether the 2 vms are the same, they are snapshots of the same vm or one is a snapshot of the other.
This means that the only when there are two different VMs relates we care that the mac is the same, this is why the duplicate mac is the correct error to raise.

psafont: `is_compatible` checks whether the 2 vms are the same, they are snapshots of the same vm or one is a snapshot of the other.
This means that the only when there are two different VMs relates we care that the mac is the same, this is why the duplicate mac is the correct error to raise.

robhoes (approved):

benjamreis (approved):

lindig (approved):

PR #4578, opened 2021-11-18
Specsavers: import message-switch

robhoes:
...including xcp-idl, forkexecd, xapi-storage, xapi-storage-script.

As usual, this was a large subtree merge, so just look at the final couple of commits.


psafont (approved):
This merge is too big to review, I've seen that the tests from the imported packages are being run and that all checks pass, so it looks good.

Further tests with the whole xenserver are being run, the initial tests were successful.

ocaml/xenopsd/c_stubs/dune
```diff
@@ -16,17 +16,7 @@
  (foreign_stubs
   (language c)
   (names tuntap_stubs xenctrlext_stubs)
-  (flags (:include xentoollog_flags.sexp))
+  (flags (-L/lib64 -lxentoollog))
```
edwintorok: dune-configurator may be able to find out the flags here if needed

edwintorok (commented):
There are probably other improvements we can do to dune configurator, but lets do them after the merge (e.g. around XAPI_VERSION).

edwintorok (approved):

PR #4579, opened 2021-11-22
CA-360754: exclude client-cert sessions from revalidation

robhoes:


lindig (approved):
Looks good. I find "client_certificate" not ideal for conveying the meaning that the session has the property of being authenticated by such a certificate but also struggle to suggest a better name. I was trying in my head with something like `certified` or `client_certified`.

psafont:
I think a name with `authenticated` could work, maybe `authenticated_with_certificate`

minglumlu (approved):

robhoes:
"Authenticated" is a bit redundant to me, because the session is the auth token.

PR #4580, opened 2021-11-23
CA-360951: Failed to lookup workgroup from domain as DNS cache

liulinC:
DNS cache all the domain controller information, if one domain
controller broken down, it still be returned as valid controller

This commit fix this issue by re-validate the controller list and
find out one valid controller for further actions

### DIFF
--- a/ocaml/xapi/extauth_plugin_ADwinbind.ml
+++ b/ocaml/xapi/extauth_plugin_ADwinbind.ml
@@ -678,24 +678,33 @@ let kdcs_of_domain domain =
        )
   with _ -> raise (generic_ex "Failed to lookup kdcs of domain %s" domain)
 
+let workgroup_from_server kdc =
+  let err_msg =
+    Printf.sprintf "Failed to lookup workgroup from server %s" kdc
+  in
+  let key = "Pre-Win2k Domain" in
+  try
+    Helpers.call_script ~log_output:On_failure net_cmd
+      ["ads"; "lookup"; "-S"; kdc; "-d"; debug_level ()]
+    |> Xapi_cmd_result.of_output ~sep:':' ~key
+    |> Result.ok
+  with _ ->
+    debug "Unable to query info from kdc %s, probably is broken down" kdc ;
+    Error (Auth_service_error (E_LOOKUP, err_msg))
+
 let kdc_of_domain domain =
-  let msg = Printf.sprintf "Failed to lookup kdc of domain %s" domain in
-  kdcs_of_domain domain |> hd msg
+  try
+    kdcs_of_domain domain
+    (* Does not trust DNS as it may cache some invalid kdcs, CA-360951 *)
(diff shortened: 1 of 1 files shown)


ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -678,24 +678,35 @@ let kdcs_of_domain domain =
...
+    Printf.sprintf "Failed to lookup workgroup from server %s" kdc
+  in
+  let key = "Pre-Win2k Domain" in
+  try
+    Helpers.call_script ~log_output:On_failure net_cmd
+      ["ads"; "lookup"; "-S"; kdc; "-d"; debug_level ()]
+    |> fun lines -> Xapi_cmd_result.of_output ~sep:':' ~key lines
+  with _ -> raise (Auth_service_error (E_LOOKUP, err_msg))
+
+let rec valid_kdc = function
```
lindig, line 692: This is conceptually a `List.find_opt` that uses `workgroup_from_server` as predicate. I would prefer that to find the fist valid entry.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -678,24 +678,35 @@ let kdcs_of_domain domain =
...
-  kdcs_of_domain domain |> hd msg
+let workgroup_from_server kdc =
+  let err_msg =
+    Printf.sprintf "Failed to lookup workgroup from server %s" kdc
+  in
+  let key = "Pre-Win2k Domain" in
+  try
+    Helpers.call_script ~log_output:On_failure net_cmd
+      ["ads"; "lookup"; "-S"; kdc; "-d"; debug_level ()]
+    |> fun lines -> Xapi_cmd_result.of_output ~sep:':' ~key lines
```
edwintorok, line 689: Do you need to use a label here `~lines`?

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -678,24 +678,35 @@ let kdcs_of_domain domain =
...
+  with _ -> raise (Auth_service_error (E_LOOKUP, err_msg))
```
edwintorok, line 690: Should we log what the actual error was here, at least at debug level?

edwintorok: And instead of raising an exception here it might be better to return `Error (Auth_service_error (E_LOOKUP, err_msg))`, and an `Ok ()` on the `try` side if it succeeds? Then you can use this as a predicate in a find_opt if you transform the result into a bool

liulinC:
@edwintorok @lindig @psafont Thanks for the review, 
Just updated, could you please help to take a look.

lindig (approved):

edwintorok (approved):

PR #4581, opened 2021-11-23
CA-360997: Don't reject imports if the host's major version is larger

robhoes:
The current code always rejects imports of the major version of the
running xapi and the one in the import metadata are not equal. This
change makes it less strict by only rejecting imports from strictly
newer versions of xapi.

### DIFF
--- a/ocaml/xapi/importexport.ml
+++ b/ocaml/xapi/importexport.ml
@@ -107,15 +107,15 @@ let this_version __context =
     This will get complicated over time... *)
 let assert_compatible ~__context other_version =
   let this_version = this_version __context in
-  (* error if major versions differ; also error if this host has a
-     lower minor vsn than the import *)
+  (* error if this host has a lower vsn than the import *)
   if
-    this_version.xapi_vsn_major <> other_version.xapi_vsn_major
-    || this_version.xapi_vsn_minor < other_version.xapi_vsn_minor
+    this_version.xapi_vsn_major < other_version.xapi_vsn_major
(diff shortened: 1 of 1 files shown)


robhoes:
Still running some tests.

robhoes:
Tests are green.

psafont (approved):

ocaml/xapi/importexport.ml
```diff
@@ -107,15 +107,15 @@ let this_version __context =
...
 let assert_compatible ~__context other_version =
   let this_version = this_version __context in
-  (* error if major versions differ; also error if this host has a
-     lower minor vsn than the import *)
+  (* error if this host has a lower vsn than the import *)
   if
-    this_version.xapi_vsn_major <> other_version.xapi_vsn_major
-    || this_version.xapi_vsn_minor < other_version.xapi_vsn_minor
+    this_version.xapi_vsn_major < other_version.xapi_vsn_major
+    || this_version.xapi_vsn_major = other_version.xapi_vsn_major
```
edwintorok, line 113: This relies on && having higher precedence than ||

edwintorok (approved):

ocaml/xapi/importexport.ml
edwintorok: ocamlformat doesn't add parantheses here (they'd be redundant) but it does it show the precedence via indentation, so this should be fine.

robhoes: Yes, I actually had () around it, but `make format` removed them :)

PR #4582, opened 2021-11-24
Yangtze backport of CA-360951: Failed to lookup workgroup from domain as DNS cache

psafont:
See #4580, waiting on tests to confirm the issue is fixed before undrafting.


lindig (approved):
If you cherry-picked it, you might want to add your sign-off.

psafont:
Now it's signed off

psafont:
I've created a build with the change and I've verified the fix works by blocking network packets from all but the last of the domain controllers available to a host, then ran a loop that joins and leaved a domain for 10 times. There weren't any failures. On an unfixed host the first iterations of joining fails.

robhoes (approved):

PR #4584, opened 2021-11-29
CA-361151: Provide default user details on ldap query fail

liulinC:
Ldap fails for cross domain if the target domain does not trust
the joined domain
Ldap is used to query user details like whether an account is
disabled, locked out etc, provide default value on failure

### DIFF
--- a/ocaml/xapi/extauth_plugin_ADwinbind.ml
+++ b/ocaml/xapi/extauth_plugin_ADwinbind.ml
@@ -977,22 +977,13 @@ module ClosestKdc = struct
 
   let lookup domain =
     try
-      let* krbtgt_sid =
-        Wbinfo.sid_of_name (Printf.sprintf "%s@%s" krbtgt domain)
-      in
-      let* domain_netbios_name =
-        Wbinfo.domain_name_of ~target_name_type:NetbiosName ~from_name:domain
-      in
-
       kdcs_of_domain domain
       |> List.map (fun kdc ->
              debug "Got domain '%s' kdc '%s'" domain kdc ;
              kdc
          )
       |> List.map (fun kdc ->
-             mtime_this ~name:kdc ~f:(fun () ->
-                 Ldap.query_user krbtgt_sid domain_netbios_name kdc
-             )
+             mtime_this ~name:kdc ~f:(fun () -> workgroup_from_server kdc)
          )
       |> List.filter_map Result.to_option
       |> List.sort (fun (_, s1) (_, s2) -> Mtime.Span.compare s1 s2)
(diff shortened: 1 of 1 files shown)


liulinC:
more discussions, see the internal CA ticket

psafont:
Can the query fail for other reasons? It doesn't look safe to be that the user query never fails anymore.

It seems to me that the code should request the joined domain for the user information, rather than the "target" one. Is that possible to do?

edwintorok (changes requested):
It would be better to store the account_locked/account_expired as a tri-state instead of a boolean. E.g. a `bool option` would be good here:
* None - means we have no info whether the acocunt is locked or expired
* Some true - means we know for sure the account is locked/expired
* Some false - means we know for sure it isn't (at the time of the query)

Then each caller can determine how to handle the missing case and what the fallback would mean.

Alternatively move all the `account_disabled`, `account_locked`, `account_expired`, `password_expired` into an `extra_info` field, and make that an `extra_info option` type, again forcing each user to deal with the `None` case.

Then we can better review the impact of this change.

psafont:
Changing to a tristate is not that simple, it's stored as a map of string to strings which is exposed in the Auth module datamodel: https://github.com/xapi-project/xen-api/blob/master/ocaml/idl/datamodel.ml#L3111

liulinC:
> 

I tried to do that by set an N/A, but the session re-validate needs a bool to decide whether destroy the session, 
so let's just keep it for now, we will totally drop it in the future.

psafont (changes requested):
This follows the design document:
1. on Lookup we stop using authenticated LDAP queries to time responses
2. on querying subject information, ldap is only used on the single-domain case, on multi-domain case, whether it's cross forest or not, the default value that doesn't invalidate sessions is used.

The second case might not be what we want, we need to make sure this new behaviour is not a regression from PBIS, this needs more discussion

liulinC:
> This follows the design document:
> 
> 1. on Lookup we stop using authenticated LDAP queries to time responses
> 2. on querying subject information, ldap is only used on the single-domain case, on multi-domain case, whether it's cross forest or not, the default value that doesn't invalidate sessions is used.
> 
> The second case might not be what we want, we need to make sure this new behaviour is not a regression from PBIS, this needs more discussion

This is not regression, 
PBIS cannot handle it even in simple domain/forest, see the discussion in the ticket.

psafont (approved):

edwintorok:
Not sure why the bool gets converted into a string when transferred through to xapi_session (and in `get_suspenion_value` it is then compared against `false`)

edwintorok (approved):

liulinC:
We probably does not need this any more as we are not sure whether winbind is the right solution to go to, 
but still merge this PR as
1. not affect other parts.
2. a little piece of refine, just in case we keep winbind in the future.

PR #4585, opened 2021-11-29
Enable to set a `Task`'s `resident_on` field.

benjamreis:
This is a continuation of https://github.com/xapi-project/xen-api/pull/4543.
When creating a Task that is not linked to another XAPI call, a user needs to set on which host this task is running.


benjamreis:
For example, our updater plugin will create a Task for a Host update & XO will read the `resident_on` field to get which host is updating and prevent/warn the user from doing a update breaking operation (ie reboot meanwhile update is still ongoing etc)

robhoes (approved):

psafont (approved):

PR #4586, opened 2021-11-29
utf8_recode: use Uutf.{Buffer.add_utf_8,String.fold_utf_8} instead of Uutf.{encoder,decoder}

edwintorok:
Uutf.encoder allocates a 64k buffer internally (in addition to our 1k buffer).
Use Uutf.Buffer to write utf8 chars directly into our buffer instead.
Also size the buffer based on initial string size (it can grow if needed).

This should reduce allocation rate.

Thanks to a hint from https://discuss.ocaml.org/t/decoding-many-unicode-strings-with-uutf/8910/2

### DIFF
--- a/ocaml/xenopsd/lib/xenops_utils.ml
+++ b/ocaml/xenopsd/lib/xenops_utils.ml
@@ -115,27 +115,17 @@ let ignore_bool (_ : bool) = ()
 let ignore_int (_ : int) = ()
 
 (* Recode an incoming string as valid UTF-8 *)
+let utf8_add_if_valid buf _ next =
+  let uchar = match next with
+  | `Malformed _ -> Uutf.u_rep
+  | `Uchar u -> u in
+  Uutf.Buffer.add_utf_8 buf uchar;
+  buf
+
 let utf8_recode str =
-  let out_encoding = `UTF_8 in
-  let b = Buffer.create 1024 in
-  let dst = `Buffer b in
-  let src = `String str in
-  let rec loop d e =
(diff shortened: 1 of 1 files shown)


ocaml/xenopsd/lib/xenops_utils.ml
```diff
@@ -115,27 +115,17 @@ let ignore_bool (_ : bool) = ()
 let ignore_int (_ : int) = ()
 
 (* Recode an incoming string as valid UTF-8 *)
+let utf8_fold buf _ next =
```
lindig, line 118: This is not a good name - this simply adds to the buffer - so it should not contain `fold`.

lindig (approved):

ocaml/xenopsd/lib/xenops_utils.ml
```diff
@@ -115,27 +115,25 @@ let ignore_bool (_ : bool) = ()
...
 
 (* Recode an incoming string as valid UTF-8 *)
+let utf8_add_if_valid buf _ next =
+  let uchar = match next with
+  | `Malformed _ -> Uutf.u_rep
+  | `Uchar u -> u in
+  Uutf.Buffer.add_utf_8 buf uchar;
+  buf
+
+let is_valid prev _ = function
```
lindig, line 125: It is somewhat unusual to have the predicate take a second argument but here we might want to do it for efficiency. `is_valid_and` might convey this slightly better.

robhoes:
Closing PR, as it was replaced by #4587.

PR #4588, opened 2021-12-01
CA-361151: Use PBIS as the default AD backend

psafont:
I couldn't find any more mentions of winbind that actually allow the code to choose it in xapi's code.


lindig (approved):

edwintorok (approved):

liulinC (approved):
This PR looks good, 
but we also have some logic in the host-installer to `preserve winbind data` or `preserver P BIS data`,  we updated it to reserve `winbind data` and ignored PBIS data, 
we need to restore it also.

PR #4589, opened 2021-12-08
CP-38759: Add pool.deconfigure_repository_proxy

minglumlu:


ocaml/idl/datamodel_pool.ml
```diff
@@ -803,6 +803,14 @@ open Datamodel_types
       ~allowed_roles:_R_POOL_OP
       ()
 
+  let deconfigure_repository_proxy = call
```
lindig, line 806: I think "deconfigure" is a quite unusual verb. I would have preferred any of "decommission", "disable", "remove". Another option is "clear", which has a different meaning.

ocaml/idl/datamodel_pool.ml
```diff
@@ -803,6 +803,14 @@ open Datamodel_types
...
+      ~name:"deconfigure_repository_proxy"
+      ~in_product_since:rel_next
+      ~doc:"Clear the configured proxy for RPM package repositories."
```
lindig, line 809: "Clear" is not the same as "deconfigure" - it reads like a contradiction. Clearing a cache means to reset it to an empty state whereas deconfigure means to remove it (in my understanding).

ocaml/xapi-cli-server/cli_frontend.ml
```diff
@@ -633,6 +633,15 @@ let rec cmdtable_data : (string * cmd_spec) list =
       ; flags= []
       }
     )
+  ; ( "pool-deconfigure-repository-proxy"
+    , {
+        reqd= []
+      ; optn= []
+      ; help= "Clear configured proxy for RPM package repositories"
```
lindig, line 640: See above - what this does is not clear enough.

ocaml/idl/datamodel_pool.ml
minglumlu: Thanks. I prefer to use "remove" in this case.

minglumlu: But the SDK has constraint on the usage of "remove". So I change to use "reset".

lindig: The updated code uses "reset", though. So maybe we should get an idea what this code is supposed to do first. What happens when this API call is executed and how does it affect future API calls?

minglumlu: The call is just to remove the configured proxy, or in other words, change it to the initial default value which stands for "no proxy" configured.

lindig: I think "remove" would be most appropriate, followed by "disable". "Reset" and "clear" would suggest the cache still exists, is initially empty, but would fill up again. If the cache does not fill up again, it is disabled.

robhoes: It's not actually a cache, but a network proxy. "Reset" to me means that it's not necessarily going away. Perhaps "disable" is the best if "remove" isn't possible.

lindig (approved):

robhoes (approved):

PR #4590, opened 2021-12-13
CP-38701: Restrict client-cert role

robhoes:

### DIFF
--- a/ocaml/idl/datamodel_roles.ml
+++ b/ocaml/idl/datamodel_roles.ml
@@ -11,11 +11,11 @@ let role_client_cert = "client-cert"
 let roles_all =
   [ (* in decreasing total linear order of privileges *)
     role_pool_admin;
-    role_client_cert;
     role_pool_operator;
     role_vm_power_admin;
     role_vm_admin;
     role_vm_operator;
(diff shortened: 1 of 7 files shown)


ocaml/idl/datamodel_roles.ml
```diff
@@ -41,3 +41,13 @@ let _R_VM_ADMIN = Some(roles_gte role_vm_admin)
 let _R_VM_OP = Some(roles_gte role_vm_operator)
 let _R_READ_ONLY = Some(roles_gte role_read_only) (* = all *)
 let _R_ALL = _R_READ_ONLY
+
+(* Only client-cert; added to one of the above where required in the datamodel *)
+let _R_CLIENT_CERT = Some([role_client_cert])
+
+let union a b =
```
lindig, line 48: I'm not sure what this is meant to do, but this does not look like a union to me. Why is this case not what we want? `Some x, None -> Some x`

robhoes: Oh yes, I forgot to update that.

lindig (approved):

psafont (approved):

PR #4591, opened 2021-12-15
CA-361988 execute cluster host_resync always locally

lindig:
Host_resync() must be executed on each host locally. For this, iterate
on the master (in message_forwarding) and call via RPC pool_resync,
which then calls host_resync().

This commit splits the code from pool_resync() between
message_forwarding) and pool_resync().


psafont (approved):
The way the distribution of the calls is done makes sense, it's unfortunate that the ~host parameter in pool_resync has to be kept due to backwards compatibility.

robhoes (approved):

PR #4592, opened 2021-12-21
CA-361209: When using WoL find the remote physical PIF

psafont:
This is needed when the host's is not a VLAN to get the correct MAC

I'll undraft once it's tested


scripts/examples/python/wake-on-lan.py
```diff
@@ -46,6 +46,15 @@ def find_host_mgmt_pif(session, host_uuid):
         raise "Failed to find a management interface (PIF) for host uuid %s" % host_uuid
     return mgmt
 
+def get_physical_pif(session, pif_ref):
+    """Return the PIF object underlying an interface"""
+    # When the management interface is on a VLAN, the actual PIF needs to be found
+    vlan = session.xenapi.PIF.VLAN_master_of(pif_ref)
+    if not vlan = "OpaqueRef:NULL":
```
robhoes, line 53: Do you not need `==` here? Or just `!=`?

robhoes (approved):
This looks good now (assuming tests will be successful).

psafont:
Regression tests for Wake on LAN look good (Job Run 3409457)

I've set up a pool with the management interface on VLAN and will test manually WoL there

lindig (approved):

psafont:
This needs another look now that I've deleted a python script, it needs to be compared with the new python file being used for both poweron and standalone script.

Support is trying out the new script

lindig (approved):

robhoes (approved):

psafont:
Confirmed by client that it works, passes current WoL tests

PR #4595, opened 2022-01-09
xenopsd: explicitly clean VM state if VM_restore failed during VM_rec…

cheese:
…eive_memory

If a failure ocurred during this periad, e.g. between Synchronisation point 1
and 2 of VM_receive_memory, and VM has on_shutdown set to Start, VM.reboot will
be triggered by the pending VM_check_state action. This patch explicitly clean
the VM state to prevent VM.reboot to actually take effect.

Similar explicit clean is already implemented for period between Synchronisation
point 2 and 4.

Signed-off-by: Robin Lee <cheeselee@fedoraproject.org>

### DIFF
--- a/ocaml/xenopsd/lib/xenops_server.ml
+++ b/ocaml/xenopsd/lib/xenops_server.ml
@@ -2582,37 +2582,46 @@ and perform_exn ?subtask ?result (op : operation) (t : Xenops_task.task_handle)
             ) ;
             debug "VM.receive_memory: Synchronisation point 1" ;
             debug "VM.receive_memory restoring VM" ;
-            (* Check if there is a separate vGPU data channel *)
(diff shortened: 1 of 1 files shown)


lindig:
Could the scope of the `try .. with` block be reduced to the part where the `VM_restore` operation is executed?

cheese:
> Could the scope of the `try .. with` block be reduced to the part where the `VM_restore` operation is executed?

Not necessary I think. The VM_destroy and VM_remove actions are to revert the result of VM_create in the previous step. So it is reasonable to guard the whole scope between Synchronisation point 1 and 2, just like the guard between Synchronisation point 2 and 4.

lindig (approved):

lindig:
Ping @robhoes

psafont (approved):
Removing whitespace differences from the diff makes the change much clearer to see.

The changes in behaviour is sensible to me

PR #4598, opened 2022-01-10
CP-38850 add xapi.conf option for cert-expiration-days

lindig:
When xapi generates/refreshes a host or pool certificate, the number of
days it will remain valid will be taken from xapi.conf.

Note that initially certficates are generated using a binary 'gencert'
which does not read this value from xapi.conf and will use the current
default of 10 years.

Also note that, for now, certificates used by clusterd have an unlimited
lifetime but this could be changed to also use the value specified in
xapi.conf.

To generate new certificates with a lifetime as in xapi.conf:

* edit xapi.conf
* restart xapi to make it read xapi.conf: service xapi restart
* xe host-reset-server-certificate
* xe host-refresh-server-certificate host=<hostname>

To verify SSL certficate lifetimes:

* openssl s_client -showcerts -connect localhost:443  2> /dev/null < /dev/null
    | openssl x509 -noout -dates

* openssl s_client -showcerts -connect localhost:443
    -servername pool 2> /dev/null < /dev/null | openssl x509 -noout -dates

### DIFF
--- a/ocaml/xapi/xapi_globs.ml
+++ b/ocaml/xapi/xapi_globs.ml
@@ -559,6 +559,8 @@ let default_wlb_timeout = 30.0
 
 let default_wlb_reports_timeout = 600.0
 
+let cert_expiration_days = ref (365 * 10)
+
 (** {2 Settings relating to dynamic memory control} *)
 
 (** A pool-wide configuration key that specifies for HVM guests a lower bound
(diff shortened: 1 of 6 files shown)


psafont (approved):
Just to make it explicit, the configuration option applies to both certificates that are shown to normal API clients and to other hosts in the pool.

This approach looks good to me

ocaml/xapi/xapi_globs.ml
```diff
@@ -1311,6 +1313,11 @@ let other_options =
     , "Frequency at which we alert any failed logins (in seconds; \
        default=3600s)"
     )
+  ; ( "cert-expiration-days"
+    , Arg.Set_int cert_expiration_days
+    , (fun () -> string_of_int !cert_expiration_days)
+    , "Number of days a refreshed certficiate will be valid. Default 10 years."
```
edwintorok, line 1319: typo in `certficiate`

edwintorok (commented):
Code looks good, small typo in doc for the flag.

robhoes:
Indeed, looks good apart from the typo.

lindig:
Fixed the typo and force-pushed to keep it to one commit.

robhoes (approved):

PR #4602, opened 2022-01-11
Use TMPDIR when testing if possible in tests to get round the opam sandbox

psafont:
for xapi_db_upgrade this means having to select one of three directories

This changes are needed for opam's 2.1.0 sanbox, which doesn't allow writing to /tmp

### DIFF
--- a/ocaml/tests/common/test_common.ml
+++ b/ocaml/tests/common/test_common.ml
@@ -15,10 +15,7 @@
 open API
 
 (* A directory to use for temporary files. *)
-let working_area = "/tmp/xapi-test"
-
-(** Utility functions *)
-let id (x : 'a) : 'a = x
+let working_area = Filename.(concat (get_temp_dir_name ()) "xapi-test")
 
 let make_uuid () = Uuid.string_of_uuid (Uuid.make_uuid ())
 
--- a/ocaml/xapi/xapi_db_upgrade.ml
+++ b/ocaml/xapi/xapi_db_upgrade.ml
@@ -303,7 +303,9 @@ let upgrade_bios_strings =
         let oem_manufacturer =
           let test_path =
             Option.map
-              (fun _ -> "/tmp/previousInventory")
+              (fun _ ->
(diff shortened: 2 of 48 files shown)


ocaml/tests/common/test_common.ml
```diff
@@ -15,10 +15,10 @@
 open API
 
 (* A directory to use for temporary files. *)
-let working_area = "/tmp/xapi-test"
-
-(** Utility functions *)
-let id (x : 'a) : 'a = x
+let working_area =
+  Filename.concat
+    (Option.value ~default:"/tmp" (Sys.getenv_opt "TMPDIR"))
```
edwintorok, line 20: Doesn't Filename module in OCaml have a way to read/use TMPDIR automatically? See `Filename.get_temp_dir_name`

edwintorok: We could also use Filename.temp_file which also takes care of creating a unique temp name with a chosen suffix (although cleaning it up is up to the caller).

psafont:
Now all tests pass on an xs-opam switch for all dependencies of xs-toolstack:

```
$ opam list -s --required-by xs-toolstack | xargs opam install -ty
```

ocaml/xapi/xapi_db_upgrade.ml
```diff
@@ -303,7 +303,9 @@ let upgrade_bios_strings =
         let oem_manufacturer =
           let test_path =
             Option.map
-              (fun _ -> "/tmp/previousInventory")
+              (fun _ ->
```
robhoes, line 306: Alternatively, we can just ditch this upgrade rule, which is for upgrades from XS 5.5, released on 2009.

robhoes (approved):

ocaml/xapi/xapi_db_upgrade.ml
psafont: I still would prefer to have this as we might have to backport this to yangtze

lindig (approved):

PR #4604, opened 2022-01-13
Use make check to show all warnings while hiding them from make

psafont:
I've silenced some warning that are difficult to silence, because they are produced by rpclib or are produced by generated code but are harmless. This allows for `make check` to use the debug profile and actually show a lot of warnings, some of them which look like they should be addressed, like when the result type is ignored.

Once build passes, I'll push another commit where `make check` is changed and shows all the issues in CI.

### DIFF
--- a/ocaml/xapi-storage-cli/main.ml
+++ b/ocaml/xapi-storage-cli/main.ml
@@ -312,7 +312,7 @@ let on_vdi' f common_opts sr vdi =
     (fun sr vdi ->
       let () = f sr vdi in
       `Ok ()
-      )
+    )
     common_opts sr vdi
 
 let mirror_start common_opts sr vdi dp url dest =
--- a/ocaml/xapi-storage-cli/main.ml
+++ b/ocaml/xapi-storage-cli/main.ml
@@ -30,9 +30,9 @@ let string_of_mirror id {Mirror.source_vdi; dest_vdi; state; failed} =
           (function
             | Storage_interface.Mirror.Receiving ->
                 "Receiving"
-            | Sending ->
+            | Storage_interface.Mirror.Sending ->
                 "Sending"
-            | Copying ->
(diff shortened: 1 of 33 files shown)


ocaml/xapi-storage-cli/main.ml
```diff
@@ -300,10 +303,13 @@ let on_vdi f common_opts sr vdi =
       `Error (true, "must supply VDI")
   | Some sr, Some vdi ->
       let sr, vdi = Storage_interface.(Sr.of_string sr, Vdi.of_string vdi) in
-      wrap common_opts (fun () -> f sr vdi)
+      wrap_exn common_opts (fun () -> f sr vdi)
+
+let on_vdi' f common_opts sr vdi =
```
lindig, line 308: I think having `on_vdi` and `on_vdi'` in scope is dangerous if they have the same type. I might be better to shadow `on_vdi` in that case.

lindig (approved):

psafont:
This needs more than 1000 changes in xapi-cli-server to eliminate all warnings / errors

lindig:
Is that a problem for merging it? I assume this is intended to make these problems visible but not to eliminate them

psafont:
I think so, merging the PR with the check activated would make CI fail, I can disable it and still merge the rest

psafont:
Warning and errors in the step are shown, while they don't block the CI. I've added a check that maintains previous behaviour so we can detect whenever problems are added when using release mode.

robhoes (approved):
I assume that you can prove that this is a "no functional change" type of PR... ;)

psafont:
Actually [`8a7e488` (#4604)](https://github.com/xapi-project/xen-api/pull/4604/commits/8a7e48804d14fe0785ec5ba0694b548177e3379c) contains a functional fix as a function was not completely applied when it should have :)

PR #4606, opened 2022-01-17
Make reboot equal to shutdown+start for CPUID changes due to a changed domain type

robhoes:
The problem occurs in particular when the domain-type of a running PV VM is changed to PV-in-PVH. The domain-type change is correctly applied when the VM reboots, but the CPUID featureset needs to change along with it, but does not. The result is a crash as soon as the VM reboots.

To fix this, we need to:

- Reset the featureset not just on start, but also on reboot.
- Updated platformdata, which contains the featureset, must be sent to xenopsd immediately and used upon reboot (but not for migrating or resuming!).

Currently in xenopsd, platformdata is stored in:
- `Xenops_interface.VM.t` (persistent)
- `Domain.create_info` (non-persistent)

plaformdata is NOT stored in:
- `VmExtra.persistent_t`

On start:
- xapi pushes `Xenops_interface.VM.t` to xenopsd.
- `Domain.create_info` is recreated in `VM_create` based on `Xenops_interface.VM.t`, and used immediately by low-level code, but not stored.

In xapi:
- `Cpuid_helpers.reset_cpu_flags` is called from VM.start (only) and sets `VM.last_boot_CPU_flags` based on the "pool cpuid level".
- When pushing/updating VM metadata to xenopsd, `Xapi_xenops.MD.of_vm` takes `VM.last_boot_CPU_flags` and puts it into `platformdata:featureset` in the `Xenops_interface.VM.t` record.
  - Other platform keys are taken from `VM.platform`, sanity checked, and included in `platformdata` as well.
  - The events-from-xapi loop regenerates the metadata and sends it to xenopsd, triggered by xapi DB changes, while a VM is running.
  - This includes regular platform keys. The `featureset` won't change, because `VM.last_boot_CPU_flags` never changes.
- `VM.last_boot_CPU_flags` is relied upon for resume to provide the correct featureset.
- For migration, xenopsd pushes `Xenops_interface.VM.t` to the remote xenopsd, which includes `platformdata:featureset`.
  - `VmExtra.persistent_t` is _also_ sent to the remote.

To address the problem, we make the following changes:
- Put the current featureset (in fact, the entire platform map) in `persistent_t` and take it from there for resume and migrate.
- Let `VM.t` be the next-boot config.
- Add featureset to xenopsd's `VM.state` and get it from `persistent_t` if requested by xapi.
- Set `VM.last_boot_CPU_flags` based on the state from xenopsd (only).
- Recompute the feature set "on the fly" when pushing new VM metadata to xenopsd (which is triggered by DB changes), rather than taking it from `VM.last_boot_CPU_flags`.

### DIFF
--- a/ocaml/xapi/cpuid_helpers.ml
+++ b/ocaml/xapi/cpuid_helpers.ml
@@ -144,24 +144,6 @@ let upgrade_features ~__context ~vm ~host host_features vm_features =
       (string_of_features upgraded_features) ;
   upgraded_features
 
-let set_flags ~__context self vendor features =
-  let features = features |> snd features_t in
-  let value =
-    [(cpu_info_vendor_key, vendor); (cpu_info_features_key, features)]
-  in
-  debug "VM's CPU features set to: %s" features ;
-  Db.VM.set_last_boot_CPU_flags ~__context ~self ~value
-
-(* Reset last_boot_CPU_flags with the vendor and feature set.
- * On VM.start, the feature set is inherited from the pool level (PV or HVM) *)
-let reset_cpu_flags ~__context ~vm =
-  let pool_vendor, _, pool_features =
-    let pool = Helpers.get_pool ~__context in
-    let pool_cpu_info = Db.Pool.get_cpu_info ~__context ~self:pool in
-    get_flags_for_vm ~__context vm pool_cpu_info
-  in
-  set_flags ~__context vm pool_vendor pool_features
-
 (* Return the featureset to be used for the next boot of the given VM. *)
 let next_boot_cpu_features ~__context ~vm =
   (* On VM.start, the feature set is inherited from the pool level (PV or HVM) *)
@@ -186,28 +168,6 @@ let next_boot_cpu_features ~__context ~vm =
     Map_check.getf ~default features_field_boot pool_cpu_info
   in
   snd features_t pool_features
-
-(* Update last_boot_CPU_flags with the vendor and feature set.
- * On VM.resume or migrate, the field is kept intact, and upgraded if needed. *)
-let update_cpu_flags ~__context ~vm ~host =
-  let current_features =
-    let flags = Db.VM.get_last_boot_CPU_flags ~__context ~self:vm in
-    Map_check.getf ~default:[||] features flags
-  in
-  debug "VM last boot CPU features: %s" (string_of_features current_features) ;
-  try
-    let host_vendor, host_features, _ =
-      let host_cpu_info = Db.Host.get_cpu_info ~__context ~self:host in
-      get_flags_for_vm ~__context vm host_cpu_info
-    in
-    let new_features =
-      upgrade_features ~__context ~vm host_features current_features
-    in
-    if new_features <> current_features then
-      set_flags ~__context vm host_vendor new_features
-  with Not_found ->
-    (* pre-Dundee? *)
-    failwith "Host does not have new leveling feature keys"
 
 let get_host_cpu_info ~__context ~vm ~host ?remote () =
   match remote with
--- a/ocaml/xapi/cpuid_helpers.ml
+++ b/ocaml/xapi/cpuid_helpers.ml
@@ -121,33 +121,22 @@ let get_flags_for_vm ~__context vm cpu_info =
   let onboot = Map_check.getf ~default:migration features_field_boot cpu_info in
   (vendor, migration, onboot)
 
-(** Upgrade a VM's feature set based on the host's one, if needed.
- *  The output will be a feature set that is the same length as the host's
- *  set, with a prefix equal to the VM's set, and extended where needed.
- *  If the current VM set has 4 words, then we assume it was last running on
- *  a host that did not support "feature levelling v2". In that case, we cannot
- *  be certain about which host features it was using, so we'll extend the set
- *  with all current host features. Otherwise we'll zero-extend. *)
-let upgrade_features ~__context ~vm host_features vm_features =
-  let len = Array.length vm_features in
+(** Upgrade a VM's feature set based on the host's one by zero-extending, if needed. *)
+let upgrade_features ~__context ~vm ~host host_features vm_features =
+  ( if Array.length vm_features <= 4 then
+      let msg =
+        "VM featureset too old to be upgraded; VM needs to be shut down before \
+         it can run here"
+      in
+      raise
+        (Api_errors.Server_error
+           ( Api_errors.vm_incompatible_with_this_host
+           , [Ref.string_of vm; Ref.string_of host; msg]
+           )
+        )
+  ) ;
   let upgraded_features =
-    if len <= 4 then
-      let open Xapi_xenops_queue in
-      let dbg = Context.string_of_task __context in
-      let module Client = (val make_client (default_xenopsd ()) : XENOPS) in
-      let uses_hvm_features =
-        match Helpers.domain_type ~__context ~self:vm with
-        | `hvm | `pv_in_pvh ->
-            true
-        | `pv ->
-            false
-      in
-      let vm_features' =
-        Client.HOST.upgrade_cpu_features dbg vm_features uses_hvm_features
-      in
-      extend vm_features' host_features
-    else
-      zero_extend vm_features (Array.length host_features)
+    zero_extend vm_features (Array.length host_features)
   in
   if vm_features <> upgraded_features then
     debug "VM featureset upgraded from %s to %s"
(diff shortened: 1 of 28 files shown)


edwintorok:
There are more fields that are wrong when using reboot instead of shutdown+start, e.g. the platform: viridian flags. Would it be possible to also ensure all platform: data is correct, not just the CPUID features?

andyhhp:
> The result is a cache as soon as the VM reboots.

crash?

ocaml/xapi/cpuid_helpers.ml
```diff
@@ -121,33 +121,10 @@ let get_flags_for_vm ~__context vm cpu_info =
...
- *  set, with a prefix equal to the VM's set, and extended where needed.
- *  If the current VM set has 4 words, then we assume it was last running on
- *  a host that did not support "feature levelling v2". In that case, we cannot
- *  be certain about which host features it was using, so we'll extend the set
- *  with all current host features. Otherwise we'll zero-extend. *)
+(** Upgrade a VM's feature set based on the host's one by zero-extending, if needed. *)
 let upgrade_features ~__context ~vm host_features vm_features =
-  let len = Array.length vm_features in
   let upgraded_features =
-    if len <= 4 then
```
edwintorok, line 134: Should we refuse to zero extend if the feature is too old? i.e. raise an exception if `len <= 4` with an error saying that the VM is unsupported because it has been created pre 7.x.
That might help find such VMs easier (rather than a more subtle failure due to featuresets being wrong.

ocaml/xenopsd/xc/xenops_server_xen.ml
```diff
@@ -1041,6 +1057,15 @@ module HOST = struct
         }
     )
 
+  let stat () =
+    match !stat_cache with
+    | None ->
+        let s = stat' () in
```
edwintorok, line 1063: This may work, but might be problematic in some cases, e.g. what if a Xen or microcode live-patch changes the available features on the CPU and requires reevaluating the next-boot featureset?
In particular when features like TSX are removed they're still emulated (in a correct, but slow way by aborting it), and on next boot we'd want to remove the TSX cpuid bit completely so the guest doesn't even try to use TSX. This happens because the host has changed its featureset "live".

We'd probably need a way to invalidate this cache, e.g. toolstack restart might work, but might be useful to have something more targeted.

edwintorok: XAPI should be aware when patch RPMs are applied through supported mechanisms, so it could send a command to xenopsd to invalidate the stat cache?

andyhhp: The actions we support to some form or another are this:

- Livepatch Xen knowing about new functionality
- Late load ucode with new functionality
    - Raw features gets updated at this point, admin checks via `xen-cpuid`
- If everything is ok, admin elects to update the default/max policies via `xen-spec-ctrl` (name in need of change)
    - toolstack restart at this point to pick everything new up
- New VMs started pick up new features
    - Obviously, we'd like to include reboots here too so VMs have expected behaviour

The 3rd bullet point is very important.  It offers the admin a way out if something didn't happen as expected with the ucode load.

For the TSX case specifically, we do not hide the RTM/HLE CPUID bits during late loading.  The runtime complexity is just too great, as demonstrated by the number of bugs we've had around maintaining migration compatibility on a clean boot of Xen.

robhoes:
> There are more fields that are wrong when using reboot instead of shutdown+start, e.g. the platform: viridian flags (if you've changed them e.g. due to updating a template and wanting to propagate that to your other VMs via a script). Would it be possible to also ensure all platform: data is correct, not just the CPUID features? (IIRC changing these can cause issues with migration where the new VM gets constructed with one set of flags and the migration stream containing another). I assume [98782a8](https://github.com/xapi-project/xen-api/commit/98782a8a72c4e1ed5a79bfeeebaeaffe163347ae) addresses my concerns though.

Yes, this PR makes it better for most platform keys. The `featureset` is a special case, because it is put into the platform map by xapi, so that it is easily passed to xenguest, which picks it up (in hindsight, it perhaps would've been better for the featureset to have it's own field in `VM.t`). I think the problem for those other platform keys was a little different though (except for `nomigrate` and `nested-virt`, which are already handled specially, and correctly): changing a key already led to an immediate metadata update in xenopsd, so a reboot _would_ pick up the new platform key. However, things might go wrong on resume/migrate, as those operations would also see the new key when creating the domain (depending on what happens with those keys). This PR ensures that updated platform keys are still immediately send to xenopsd, but they will only be used at the next boot.

ocaml/xapi/cpuid_helpers.ml
robhoes: OK, good idea.

robhoes:
Note that there are still be some remaining cases for which reboot != start+shutdown. The first one is adding VIFs or VBDs without hot-plugging: they will get plugged on shutdown+start, but not on reboot. This requires a different solution.

Also, regarding CPUID featuresets, even with this PR, a change in pool level does not yet lead to a new featureset sent to xenopsd for running VMs, so a _VM internal_ reboot would not pick it up (a reboot through the API/CLI would). For that to work, the events_from_xapi thread could be changed to pick it pool-level changes and push it to running VMs. That will lead to the metadata for _all_ running VMs in the pool to be updated though, so we'd have to assess the overhead of it on big pools.

ocaml/xenopsd/xc/xenops_server_xen.ml
robhoes: You're right, but currently xapi only fetches a host's cpu info when it starts up, so this patch doesn't change the current behaviour (see commit message). Indeed a toolstack restart is needed to pick up a change in host features. This can be addressed in a separate PR.

ocaml/xapi/cpuid_helpers.ml
psafont: I don't think this has been addressed yet

psafont: I think this should be addressed

robhoes:
@andyhhp: when this is in, we no longer need ocaml-cpuid-helpers.patch in xen.

ocaml/xapi/cpuid_helpers.ml
robhoes: Done now.

andyhhp:
> @andyhhp: when this is in, we no longer need ocaml-cpuid-helpers.patch in xen.

:smile:    Way ahead of you
```
commit 7b0f284227803e08d41b04ea9a4ca89d3fbe2c8d
Author: Andrew Cooper <andrew.cooper3@citrix.com>
Date:   Tue Jun 22 12:50:39 2021

    CP-37343: Drop Ocaml/CPUID technical debt
    
```

ocaml/xapi/xapi_xenops.ml
```diff
@@ -1213,18 +1213,28 @@ module MD = struct
...
     let platformdata =
       if not (List.mem_assoc Vm_platform.featureset platformdata) then
         let featureset =
-          if
-            List.mem_assoc Xapi_globs.cpu_info_features_key
-              vm.API.vM_last_boot_CPU_flags
-          then
-            List.assoc Xapi_globs.cpu_info_features_key
-              vm.API.vM_last_boot_CPU_flags
+          if vm.API.vM_power_state = `Suspended then
```
lindig, line 1220: I would turn this around to reduce the nesting and combine `match` and `when`:
```
match List.assoc_opt Xapi_globs.cpu_info_features_key vm.API.vM_last_boot_CPU_flags with
| ..  when vm.API.vM_power_state = `Suspended ->
```

lindig (approved):
This has a lot of detail and subtleties - I did not spot anything obvious.

ocaml/xapi/xapi_xenops.ml
robhoes: It'll be like this then, otherwise we need ``= `Suspended`` on two cases:
```
          match
            List.assoc_opt Xapi_globs.cpu_info_features_key
              vm.API.vM_last_boot_CPU_flags
          with
          | _ when vm.API.vM_power_state <> `Suspended ->
            (* The VM's current featureset is now part of xenopsd's
               persistent metadata, and taken from there on resume
               and migrate-receive. However, VMs suspended before this
               change don't have the featureset there yet, and xenopsd
               falls back to the platformdata. We can't detect this case
               here, so we'll therefore fall back to the original source
               (VM.last_boot_CPU_flags) regardless. *)
            Cpuid_helpers.next_boot_cpu_features ~__context ~vm:vmref
          | Some fs ->
              fs
          | None ->
              failwith "VM's CPU featureset not initialised"
```

lindig: I like it but would not insist on it

psafont:
This needs a reformat with ocamlformat 0.19 :D

psafont (approved):

ocaml/xapi/xapi_xenops.ml
robhoes: Updated.

PR #4608, opened 2022-01-28
CP-38978: Add systemd socket for on-demand stunnel in CC mode

minglumlu:


scripts/xe-toolstack-restart
```diff
@@ -34,6 +34,8 @@ SERVICES="perfmon xapi v6d xenopsd xenopsd-xc xenopsd-xenlight
 
 tmp_file=$(mktemp --suffix="xe-toolstack-restart")
 systemctl stop stunnel@xapi > $tmp_file 2>&1
+systemctl stop stunnel-on-demand.socket >> $tmp_file 2>&1
```
minglumlu, line 37: Ross Lagerwall2 days ago
Maybe this should be done in either case to handle the possibility of switching from CC_PREPARATIONS=true to false.

Ming Lu （陆明）2 days ago
We don't support this switching.
A pool would be installed in CC or non-CC mode. It is not allowed to change an existing pool between two modes.

Rob Hoes2 days ago
I'd still just stop them unconditionally (as the operation is idempotent. Also I think you need to stop stunnel-on-demand@xapi?

Ming Lu （陆明）2 hours ago
Sure. It's for additional confidence 🙂

ocaml/xapi/helpers.ml
```diff
@@ -1768,39 +1768,52 @@ end = struct
...
-    | "enabled" ->
-        true
-    | "disabled" ->
-        false
-    | unknown ->
-        D.error
-          "Stunnel.is_enabled: expected 'enabled' or 'disabled', but got: %s"
-          unknown ;
-        false
+  let systemctl_socket cmd = call_script !Xapi_globs.systemctl [cmd; "stunnel-on-demand.socket"]
```
minglumlu, line 1771: Rob HoesYesterday
Why no stunnel-on-demand@xapi here?

Ming Lu （陆明）2 hours ago
Indeed. I added a systemctl stop stunnel-on-demand@*.service to clean it up in restarting.
Otherwise, systemctl disable stunnel-on-demand.socket will clean up all running service instances.

ocaml/xapi/helpers.ml
```diff
@@ -1768,39 +1768,52 @@ end = struct
...
+  let _enable () =
+    if is_fips_enabled () then (
+      systemctl_ "stop" ;
+      systemctl_ "disable" ;
+      systemctl_socket_ "enable"
+    )
+    else (
+      systemctl_socket_ "stop" ;
+      systemctl_socket_ "disable" ; (* This will stop all stunnel-on-demand@*.service instances *)
+      systemctl_ "enable"
```
minglumlu, line 1787: Ross Lagerwall2 days ago
Should this stop and disable the socket in case it's switching from fips to non-fips?

Rob HoesYesterday
Might as well to avoid confusion, and for additional confidence (even if we don't support the switching).

ocaml/xapi/helpers.ml
```diff
@@ -1694,6 +1694,19 @@ module Stunnel : sig
...
+  let is_fips_enabled () =
+    Xapi_inventory.lookup ~default:"false" "CC_PREPARATIONS" |> function
+    | "true" ->
+        true
+    | "false" ->
+        false
+    | unknown ->
+        info
+          "CC_PREPARATIONS has unexpected value '%s'. assuming fips mode \
+           should be disabled"
```
lindig, line 1706: I would replace "should be" with "is". Otherwise this can create a bit of ambiguity between what we assume and what we do.

ocaml/xapi/helpers.ml
```diff
@@ -1704,25 +1717,13 @@ end = struct
...
-            "no"
-        | unknown ->
-            info
-              "CC_PREPARATIONS has unexpected value '%s'. assuming fips mode \
-               should be disabled"
-              unknown ;
-            "no"
-      in
       let conf_contents =
         String.concat "\n"
```
lindig, line 1721:
```
let fips = is_fips_enabled () in
...
if fips then .. else
```

ocaml/xapi/helpers.ml
```diff
@@ -1768,39 +1773,51 @@ end = struct
...
-        true
-    | "disabled" ->
-        false
-    | unknown ->
-        D.error
-          "Stunnel.is_enabled: expected 'enabled' or 'disabled', but got: %s"
-          unknown ;
-        false
+  let systemctl_socket cmd =
+    call_script !Xapi_globs.systemctl [cmd; "stunnel-on-demand.socket"]
```
lindig, line 1777: Just as below, use `|> ignore` directly.

ocaml/xapi/helpers.ml
```diff
@@ -1768,39 +1773,51 @@ end = struct
...
+
+  let systemctl_socket_ cmd = systemctl_socket cmd |> ignore
+
+  let systemctl_socket_stop_on_demand_ () =
+    call_script !Xapi_globs.systemctl ["stop"; "stunnel-on-demand@*.service"]
+    |> ignore
+
+  let _enable () =
```
lindig, line 1785: Why are these names prefixed by underscores? This has a special meaning in OCaml in that they are not reported if unused.

ocaml/xapi/helpers.ml
```diff
@@ -1768,39 +1773,51 @@ end = struct
...
+        "restarting stunnel-on-demand.socket" (fun () ->
+          try
+            (* To make all existing instances be terminated before restarting,
+             * the systemd socket needs to be stopped and then started.
+             *)
+            systemctl_socket_ "stop" ;
+            systemctl_socket_stop_on_demand_ () ;
+            systemctl_socket_ "start" ;
+            true
+          with e -> false)
```
lindig, line 1810: If something goes wrong, create a log entry.

ocaml/xapi/helpers.ml
robhoes: The function is actually used for both: to determine whether we need to enable FIPS mode, and to determine whether we would have enabled it.

Perhaps we call it cc_mode rather than fips in the name of this function: `is_cc_mode_enabled`, or just `cc_mode`.

ocaml/xapi/helpers.ml
minglumlu: Thanks. I renamed them.

ocaml/xapi/helpers.ml
minglumlu: The functions will call `call_script` eventually in which it will write log.

robhoes (approved):

psafont (approved):

minglumlu:
Squashed fixups

PR #4609, opened 2022-01-28
Further winbind for ntlm and cross forest

liulinC:


ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -707,44 +709,66 @@ let query_domain_workgroup ~domain =
...
+          ]
   in
+
   let len = String.length conf_contents in
   Unixext.atomic_write_to_file smb_config 0o0644 (fun fd ->
       let (_ : int) = Unix.single_write_substring fd conf_contents 0 len in
       Unix.fsync fd
   )
 
+let clear_winbind_config () =
```
lindig, line 765: Is this a good function? Does it make sense to either have or not to have a winbind configuration based on the debug level? It seems to me that these are independent concepts. The name is likewise confusing for that reason: it says clear btu only creates an empty configuration in some conditions.

liulinC: The configuration is very import for debug, so I kept it with the max debug level.
but your concern make sense, I will add a new configuration item for this.

ocaml/xapi/xapi_globs.ml
```diff
@@ -1286,6 +1288,17 @@ let other_options =
     , "Encryption types to use when operating as Kerberos client \
        [strong|legacy|all]"
     )
+  ; ( "winbind_allow_kerberos_auth_fallback"
+    , Arg.Set winbind_allow_kerberos_auth_fallback
+    , (fun () -> string_of_bool !winbind_allow_kerberos_auth_fallback)
+    , "Whether allow fallback to other auth on kerberos failure"
```
lindig, line 1294: Language: "whether *to* allow". Likewise below.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -707,44 +711,67 @@ let query_domain_workgroup ~domain =
...
+          ]
   in
+
   let len = String.length conf_contents in
   Unixext.atomic_write_to_file smb_config 0o0644 (fun fd ->
       let (_ : int) = Unix.single_write_substring fd conf_contents 0 len in
       Unix.fsync fd
   )
 
+let clean_winbind_config () =
```
lindig, line 767: Is "clean" here to be read as an adjective or a verb? Using "clear" could be less ambiguous if you want this to be a verb.

liulinC: My dictionary tell me `clean` can also be used as `vert`
but you are expert, I will update.

lindig (approved):

liulinC:
format check failed, is the format updated again?

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -707,44 +711,67 @@ let query_domain_workgroup ~domain =
...
+              (Kerberos_encryption_types.Winbind.to_string
+                 !Xapi_globs.winbind_kerberos_encryption_type
+              )
+          ; Printf.sprintf "workgroup = %s" workgroup
+          ; Printf.sprintf "netbios name = %s" netbios_name
+          ; "idmap config * : range = 3000000-3999999"
+          ; Printf.sprintf "idmap config %s: backend = rid" dom
+          ; Printf.sprintf "idmap config %s: range = 2000000-2999999" dom
+          ; Printf.sprintf "log level = %s" (debug_level ())
+          ; Printf.sprintf "allow kerberos auth fallback = %s" allow_fallback
```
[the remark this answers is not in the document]
liulinC: this is from our internal patch, 
I sent you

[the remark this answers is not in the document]
edwintorok: Would be useful to add a comment about that here, so we'll remember to update this when we update to new version of Samba (that may or may not need this option or a different one).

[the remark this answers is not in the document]
liulinC: updated above, (line 736~~740)

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -729,10 +729,15 @@ let query_domain_workgroup ~domain =
...
-let config_winbind_damon ?(workgroup = "") ?(netbios_name = "") domain =
+let config_winbind_daemon ?(workgroup = "") ?(netbios_name = "") domain =
   let open Xapi_stdext_unix in
   let smb_config = "/etc/samba/smb.conf" in
   let allow_fallback =
+    (*`allow kerberos auth fallback` depends on our internal samba patch,
+     * this patch disalbe fallback to ntlm by default and can be enabled
+     * Looks like upstream is doing something similar on master with
+     * configration `weak_crypto`, check and replace the internal patch when
+     * upgrade to samba packages with this capacity *)
```
edwintorok, line 740: There are quite a few typos in the comment here, but the code looks OK now. 'disalbe', 'configration', please fix those prior to merging.

edwintorok (approved):

ocaml/xapi/extauth_plugin_ADwinbind.ml
psafont: There are also a few typos in the commit messages

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -708,40 +708,49 @@ let query_domain_workgroup ~domain =
     workgroup_from_server kdc |> Result.get_ok
   with _ -> raise (Auth_service_error (E_LOOKUP, err_msg))
 
-let config_winbind_damon ~domain ~workgroup ~netbios_name =
+let config_winbind_damon ?(workgroup = "") ?(netbios_name = "") domain =
```
psafont: These 2 parameters cannot be erased, make them non-optional or add a unit at the end as a parameter

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -708,40 +708,49 @@ let query_domain_workgroup ~domain =
...
+          ; Printf.sprintf "winbind cache time = %d"
+              !Xapi_globs.winbind_cache_time
+          ; Printf.sprintf "machine password timeout = %d"
+              !Xapi_globs.winbind_machine_pwd_timeout
+          ; Printf.sprintf "kerberos encryption types = %s"
+              (Kerberos_encryption_types.Winbind.to_string
+                 !Xapi_globs.winbind_kerberos_encryption_type
+              )
+          ; Printf.sprintf "workgroup = %s" workgroup
+          ; Printf.sprintf "netbios name = %s" netbios_name
```
psafont, line 768: Is it OK to write `workgroup = ` in the file? (same for netbios)

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -821,6 +830,7 @@ let disable_machine_account ~service_name = function
       in
       try
         Helpers.call_script ~env net_cmd args |> ignore ;
+        config_winbind_damon None ;
```
psafont: Using `config_winbind_damon ~workgroup:None ~netbios_name:None ~domain:None ;`

Would make the intent of the call clearer.

ocaml/xapi/xapi_globs.ml
```diff
@@ -931,8 +931,7 @@ let winbind_machine_pwd_timeout = ref (7 * 24 * 3600)
 
 let winbind_update_closest_kdc_interval = ref (3600. *. 24.) (* every day *)
 
-let winbind_kerberos_encryption_type =
-  ref Kerberos_encryption_types.Winbind.Strong
+let winbind_kerberos_encryption_type = ref Kerberos_encryption_types.Winbind.All
```
psafont, line 935: > * Some old DC only support legacy encryption type
> * One way trust needs extra configration in DC to support strong encrytion type

Can we add this to the documentation instead of downgrading the crypto being used?

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -1063,8 +1063,13 @@ let build_dns_hostname_option ~config_params =
...
-  | Some kdc ->
-      kdc
+  | Some kdc -> (
+    match workgroup_from_server kdc with
+    | Error _ ->
+        kdc_of_domain domain
+    | Ok _ ->
+        kdc
+  )
   | None ->
```
psafont, line 1113:
```suggestion
  | Some kdc when Result.is_ok (workgroup_from_server kdc) ->
      kdc
  | _ ->
```

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: I think that has already been updated.

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: why do you think it is a problem?

ocaml/xapi/xapi_globs.ml
liulinC: We will document about this, 
but I am afraid we cannot avoid downgrading the crypto being used here, otherwise it will break our own usage (and thus xenrt) as there is one citrite domain (maybe ROOTCTX, not very sure) only support legacy encrytion type.
also, one way trust is used in citrite (eng.citrite.net ----one way trust--> citrite.net)

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -1170,8 +1185,13 @@ module AuthADWinbind : Auth_signature.AUTH_MODULE = struct
...
     let* {user_name; gecos; gid} = Wbinfo.uid_info_of_uid uid in
     let* domain_netbios, domain = Wbinfo.domain_of_uname user_name in
-    let closest_kdc = closest_kdc_of_domain domain in
-    let {service_name; _} = get_domain_info_from_db () in
+    let upn = "" in
+    let display_name = "" in
+    let account_disabled = false in
+    let account_locked = false in
+    let account_expired = false in
+    let password_expired = false in
```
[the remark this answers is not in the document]
liulinC: yes, that was I tried to do at first place, 
but compiler is not happy with it with following errors , any suggestion?
```File "ocaml/xapi/extauth_plugin_ADwinbind.ml", line 1214, characters 8-12:                                                                         
1214 |         name= user_name
               ^^^^
Error: Unbound record field name
make: *** [build] Error 1```

ocaml/xapi/extauth_plugin_ADwinbind.ml
psafont: The main reason is that having empty values may be handled in a way different from not having it defined at all.

I see in the manpage that some default are defined with an empty value, so it seems this should be OK.
> Default: `add group script = `

ocaml/xapi/xapi_globs.ml
liulinC: I am afraid we have to enable `All` ciphers here, as
- One way trust needs this (otherwise, extra AD configuration is required)
- Customer may upgrade to next, and have their environment broken.  (PBIS do support legacy).
BTW, it is agreed during design review meeting with security team we enable all, and document how to improve security by set to `Strong`.  
per my understand,  this is only about cyphers,  kerberos protocol improve the security also.

psafont (approved):

ocaml/xapi/xapi_globs.ml
Zhengchai: @liulinC The requirement says 5 cihpers but not All - is the Winbind.All exactly equaling below?
Also I think the requirement indicate that we should use the AES first (over the other 3 legacy ciphers), how is this matching the requirement?

- aes256-cts-hmac-sha1-96
- aes128-cts-hmac-sha1-96
- RC4-HMAC
- DES-CBC-CRC
- DES-CBC-MD5

liulinC: I think we have talked about this in the desgin. 
- all means the 5 ciphers, 
- strong means only aes256-cts-hmac-sha1-96 and aes128-cts-hmac-sha1-96
- legacy means only RC4-HMAC etc,
Winbind negotiate ciphers with DC, and will pick aes as possible, but for cases like one-way trust and only old ciphers, it will fallback to legacy.

PR #4612, opened 2022-01-31
Enable certificate checking by default

lindig:
* Enable certificate checking
* Disable certificate rotation but make it available for tests

### DIFF
--- a/scripts/network-init
+++ b/scripts/network-init
@@ -108,13 +108,6 @@ rename_network_label() {
     done
 }
 
-maybe-enable-tls-verification() {
-  verification_enabled="$(${XE} pool-list params=tls-verification-enabled --minimal)"
-  if [ "$FIRSTBOOT_ENABLE_TLS_VERIFICATION" == "true" ] && [ "$verification_enabled" != "true" ] ; then
-      ${XE} pool-enable-tls-verification
-  fi
-}
-
 if [ "$UPGRADE" = "true" ]; then
     rename_network_label
 else
@@ -122,7 +115,7 @@ else
     rename_network_label
     # Ensure changes are synced to disk
     xe pool-sync-database
-    maybe-enable-tls-verification
+    ${XE} pool-enable-tls-verification
 fi
 
 touch /var/lib/misc/ran-network-init
--- a/scripts/certificate-refresh
+++ b/scripts/certificate-refresh
@@ -10,6 +10,8 @@ prio="local5.error"
 tag="$0"
 xe=/opt/xensource/bin/xe
 
-/usr/bin/sleep $((RANDOM % 600))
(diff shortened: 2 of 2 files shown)


scripts/network-init
```diff
@@ -108,21 +108,14 @@ rename_network_label() {
...
-
 if [ "$UPGRADE" = "true" ]; then
     rename_network_label
 else
     prepare_networking
     rename_network_label
     # Ensure changes are synced to disk
     xe pool-sync-database
-    maybe-enable-tls-verification
+    xe pool-enable-tls-verification
```
psafont, line 118:
```suggestion
    ${XE} pool-enable-tls-verification
```

psafont (approved):

scripts/network-init
```diff
@@ -108,21 +108,14 @@ rename_network_label() {
...
-
 if [ "$UPGRADE" = "true" ]; then
     rename_network_label
 else
     prepare_networking
     rename_network_label
     # Ensure changes are synced to disk
     xe pool-sync-database
-    maybe-enable-tls-verification
+    ${XE} pool-enable-tls-verification
```
robhoes, line 118: We don't need to check anymore if it is already enabled, as in the old function?

scripts/network-init
```diff
@@ -108,21 +108,14 @@ rename_network_label() {
...
-  fi
-}
-
 if [ "$UPGRADE" = "true" ]; then
     rename_network_label
 else
     prepare_networking
     rename_network_label
     # Ensure changes are synced to disk
     xe pool-sync-database
```
psafont, line 117: I wonder why plain xe is used here if `$XE `is defined, It would make sense to use only one of them.

scripts/network-init
edwintorok: What if it is an 'update' (i.e. installed as a hotfix in XenCenter) as opposed to an upgrade (reboot into host-installer). Which path will this script take here?

scripts/network-init
psafont: The call exchanges certificates every time. 

I think this behaviour is desired to be able to recover pools, as it should be idempotent, although it may do additional non-essential work.

We may want to reintroduce the check here

robhoes: @lindig ?

lindig: What is the question? About the behaviour when certs already exist and are distributed? We have a design where verification can be enabled once and disabled per host in an emergency. Looking at the code I don't immediately see that it would refresh certs.

psafont: Mh, it might not be needed, first boot is always a single-host pool

psafont (approved):

robhoes (approved):

PR #4616, opened 2022-02-03
CA-363207: SSH access failing when using AD groups with spaces in name

liulinC:
pam module cannot handle subject name with space, instead, it expect
space be replaced with '+'

### DIFF
--- a/scripts/plugins/extauth-hook-AD.py
+++ b/scripts/plugins/extauth-hook-AD.py
@@ -165,6 +165,9 @@ class DynamicPam(ADConfig):
         try:
             sid = subject_rec['subject_identifier']
             name = subject_rec["other_config"]["subject-name"]
+            # CA-363207: Pam cannot handle space, put it inside []
+            if " " in name:
+                name = "[{}]".format(name)
             is_group = subject_rec["other_config"]["subject-is-group"] == "true"
             logger.debug("Permit %s with sid %s is_group as %s", name, sid, is_group)
             condition = "ingroup" if is_group else "="


scripts/plugins/extauth-hook-AD.py
```diff
@@ -165,6 +165,9 @@ def _add_subject(self, subject_rec):
         try:
             sid = subject_rec['subject_identifier']
             name = subject_rec["other_config"]["subject-name"]
+            # CA-363207: Need to replace space to + for pam configuration
+            # "AA BB" -> "AA+BB", "AA   BB" -> "AA+++BB"
+            name = name.replace(' ', "+")
```
robhoes, line 170: Do we need to do the reverse and turn '+' into ' ' somewhere else?

edwintorok:
[Might be useful to test all the special characters supported by AD and check whether they work in a PAM config. (except `#` which we already know is broken because it is a PAM comment char)]
(https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-2000-server/bb726984(v=technet.10)?redirectedfrom=MSDN) says "Logon names can't contain certain characters. Invalid characters are `" / \ [ ] : ; | = , + * ? < >` "

scripts/plugins/extauth-hook-AD.py
edwintorok: Apparently + is not a valid char for a logon name according to some Windows 2000 docs (can't easily find docs for anything newer, whether it has become valid meantime)

liulinC: windows server 2016 keep consistent with the doc.
put name inside [] works both for pbis and winbind, while replace ' ' -> '+' only work for pbis, 
updated the PR.

liulinC: @robhoes  I do not think we need to reverse

robhoes (approved):

psafont (approved):

PR #4617, opened 2022-02-04
CA-363207: SSH access failing when using AD groups with spaces in name (Backport)

robhoes:
pam module cannot handle subject name with space, the subject is split
by space, put it into [] to handle it

(cherry picked from commit 1e5ab3728e34588de1bde833f73bc8b382975c65)


psafont (approved):
This is a faithful backport

lindig (approved):

PR #4618, opened 2022-02-04
CP-39031 keep more xapi version details for Host.software_versions

lindig:
Keep more details about the version of xapi as it is recorded in the
"sofware-versions" field of a host. This is most useful on a host where
a new version has been installed but not yet started. We want to know,
what version is currently running.

To make this useful, we will have to pass a more detailed XAPI_VERSION 
during builds into the environment from where it is picked up. That 
requires changes in the SPEC file - so let's do this first.

### DIFF
--- a/ocaml/xapi/create_misc.ml
+++ b/ocaml/xapi/create_misc.ml
@@ -569,6 +569,7 @@ let make_software_version ~__context host_info =
   @ v6_version
   @ [
       ("xapi", get_xapi_verstring ())
+    ; ("xapi_build", Build_info.version)
     ; ("xen", Option.value ~default:"(unknown)" host_info.xen_verstring)
     ; ("linux", host_info.linux_verstring)
     ; ("xencenter_min", Xapi_globs.xencenter_min_verstring)


ocaml/util/xapi_version.ml
```diff
@@ -20,9 +20,11 @@ let hostname = "localhost"
 
 let date = Build_info.date
 
-let xapi_version_major, xapi_version_minor =
+let xapi_version_major, xapi_version_minor, xapi_version_rel =
```
robhoes, line 23: I think `micro` and `%d`.

lindig: With `%d` we would miss extra information also present: `22.5.0.18.gb60d592f7.2.2.g014ff5d.xs8` - this is currently captured.

ocaml/xapi/create_misc.ml
```diff
@@ -470,7 +470,8 @@ let create_root_user ~__context =
     Db.User.create ~__context ~ref ~fullname ~short_name ~uuid ~other_config:[]
 
 let get_xapi_verstring () =
-  Printf.sprintf "%d.%d" Constants.version_major Constants.version_minor
+  Printf.sprintf "%d.%d.%s" Constants.version_major Constants.version_minor
```
edwintorok, line 473: Not sure whether we should put the extra version info into this field or into a separate field, if there is any software out there that parsed this version number and expected a 2 numbers it'd break.
It is fine to change this on trunk, but if we want to backport it we should probably add a separate full-xapi-version field there?

ocaml/util/xapi_version.ml
robhoes: OK, on a release RPM build, this would only be the micro version, but on a dev build it will have further stuff on it (not the release number though).

lindig:
I've introduced a new field and reformatted using new OCamlformat.

ocaml/xapi/create_misc.ml
```diff
@@ -569,6 +569,10 @@ let make_software_version ~__context host_info =
   @ v6_version
   @ [
       ("xapi", get_xapi_verstring ())
+    ; ( "xapi_release"
```
robhoes, line 572: This isn't the release number in the RPM sense, but a version number. Perhaps name it `toolstack`?

Also, perhaps just use `Build_info.version` here directly rather than unpacking it and putting it back together?

ocaml/xapi-consts/constants.ml
```diff
@@ -383,6 +383,8 @@ let version_major = Xapi_version.xapi_version_major
 
 let version_minor = Xapi_version.xapi_version_minor
 
+let version_release = Xapi_version.xapi_version_rel
```
robhoes, line 386: As below, it's not a release number.

lindig:
Simplified this.

robhoes (approved):

PR #4619, opened 2022-02-07
CA-363154: Use repoquery to get available updates

minglumlu:
Per Edwin Török's comments, 'repoquery' is better than 'yum' to get the
available updates as it can format the output which could be parsed by
XAPI easily.


ocaml/xapi/repository_helpers.ml
```diff
@@ -1143,71 +1151,39 @@ let eval_guidances ~updates_info ~updates ~kind =
...
     [
-      "-q"
+      "-a"
     ; "--disablerepo=*"
     ; Printf.sprintf "--enablerepo=%s" (String.concat "," repositories)
-    ; "list"
-    ; "updates"
+    ; "--pkgnarrow=updates"
+    ; "--qf"
+    ; "%{name}:%{epoch}:%{version}:%{release}:%{arch}:%{repoid}"
```
lindig, line 1178: The parsing of this line relies on the precondition that `:` cannot appear in any of these fields. Is this actually the case? If there is no certainty, characters that I would consider to be less likely to be used in a any of these fields are space, `|`, `<` - or any character that has meaning in a shell because packages would avoid them even when they are not illegal.

robhoes:
I like the `+61 −248`.

robhoes:
We can ignore the "Check format" failure this week: we have a new ocamlformat version (with breaking changes) in xs-opam on github, but not yet in the build system. When we get that updated, we'll have to reformat everything.

robhoes:
>Some errors in parsing updates should be ignored so that they would not
block the consequent applying updates.

Then how to we provide feedback to the user about failed updates? Likewise, would we ourselves still be able to catch problem such as the one that triggered this bugfix, if errors are ignored?

robhoes:
The code looks good, but I'd like answers to Christian's question and my question about error reporting above.

ocaml/xapi/repository_helpers.ml
edwintorok: it would be good to make this and the rpm -qa format consistent: whatever concerns we have for parsing 'repoquery' output the same applies to parsing 'rpm -qa' output, and in fact the query format here and in rpm -qa should use the same field names. So would be good to define this query format as a constant, and then next to it move the function that parses that format: that way it'll be obvious that one needs to be updated when the other one is, and we can share code between rpm -qa and repoquery more easily.

minglumlu: The `:` will not appear in any of these fields as it is a special one for epoch. For example: `libpath-utils-2:0.2.1-29.el7.x86_64` means the epoch is `2`. So I think `:` should be safe.
I can't find a document to clarify this definitively :(

edwintorok: How about using `:\t` as separator? `/usr/lib/rpm/rpmpopt-4.11.3` has a bunch of queries as examples used by RPM itself internally and separators used there are `:`, `\t`,`|` and ` `.
Even if a package would have a whitespace or : somewhere it is very unlikely it'd be able to have both next to each other, and then you could use `cuts` to split on `":\t"` (of course the danger here is that something would convert tabs to spaces, but it'll be very obvious if that happens)

robhoes:
> I confess the 3rd commit is lack of confidence. The origin of this commit is to get the updates applied firstly. Otherwise, the whole stream would be blocked until a fix arrived.

OK, true, if a bug in this code means that no further updates can be applied, including to fix said bug, then we're in trouble.

ocaml/xapi/repository_helpers.ml
edwintorok: https://github.com/xapi-project/xen-api/blob/a9e9763ad2d2282df5426e0959a14df37ab7f658/ocaml/xapi/repository_helpers.ml#L1049 this is the 2nd occurance of 'rpm -qa' that I'm refering to, and it would be nice to use the same format string there and in repoquery, and same parsing function.

minglumlu: Thanks @lindig @edwintorok 
I've appended a commit for this change.

minglumlu:
Rebased and appended a new commit as there were conflicts.

ocaml/xapi/repository_helpers.ml
```diff
@@ -1151,8 +1144,15 @@ let eval_guidances ~updates_info ~updates ~kind =
   |> GuidanceSet.resort_guidances ~kind
   |> GuidanceSet.elements
 
+let repoquery_sep = ":|"
+
+let get_repoquery_fmt () =
```
robhoes, line 1149: This doesn't actually need to be a function, as it is entirely static (so `()` can be omitted). But it doesn't matter much.

robhoes (approved):

ocaml/xapi/repository_helpers.ml
```diff
@@ -731,15 +728,26 @@ module UpdateInfo = struct
...
+                          (* The error should not block update. Ingore it. *)
+                          warn "%s" (ExnHelper.string_of_exn e) ;
+                          acc
+                      )
+                      | Xml.Element ("absolute_guidance", _, [Xml.PCData v])
+                        -> (
+                        try {acc with abs_guidance= Some (Guidance.of_string v)}
+                        with e ->
+                          (* The error should not block update. Ingore it. *)
+                          warn "%s" (ExnHelper.string_of_exn e) ;
```
edwintorok, line 743: Might be useful to be able to report back any parsing errors in a separate part of the reply, probably something as a future enhancement. Then the cloud could keep track of these parsing failures so we eventually resolve them (or we could run a more strict version of this in xenrt to detect errors early).

edwintorok (approved):
Thanks, the shared repoquery sep and fmt should make the parsing more robust, and allow us to detect failures more easily, not just during updates.

PR #4622, opened 2022-02-07
CA-363391: fix wake-on-lan script

robhoes:
Bug in 930f879b (missing function param).


edwintorok (approved):
Would be nice if we could use `mypy`. Once we upgrade to Python3 that is.

lindig (approved):

PR #4623, opened 2022-02-07
Maintenance: reformat with new ocamlformat version

robhoes:
I hate doing seemingly arbitrary changes like this, but we don't have much of a choice if we stick with ocamlformat.


lindig (approved):

liulinC:
will 8.2 LTSR use the same format?

edwintorok (approved):

robhoes:
>will 8.2 LTSR use the same format?

Planning to reformat the lcm branch as well, yes.

PR #4624, opened 2022-02-08
Use Filename to concat varstore dir and file

benjamreis:
The simple string concat can lead to bad filename.
For instance with default `varstore_dir` the filename is: `/usr/share/varstoredKEK.auth`

### DIFF
--- a/ocaml/xapi/xapi_vm.ml
+++ b/ocaml/xapi/xapi_vm.ml
@@ -315,7 +315,7 @@ let with_temp_file_contents ~contents f =
     (fun () -> Sys.remove filename)
 
 let save_uefi_certificates_to_dir ~__context ~pool ~vm =
-  let uefi_key = !Xapi_globs.varstore_dir ^ "KEK.auth" in
+  let uefi_key = Filename.concat !Xapi_globs.varstore_dir "KEK.auth" in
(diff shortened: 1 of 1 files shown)


benjamreis:
This lead to the SB certificates to be overwritten on host disks by XAPI ones at every UEFI VM start up.
Because we always enter in [this test](https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_vm.ml#L319).

lindig (approved):
The existing code relies on `varstore_dir` ending with `/` - that's not robust and this is better.

benjamreis:
Indeed, is there a reason not to overwrite SB certificates on a host disk by the ones in XAPI DB?

edwintorok:
We can change the default dir, but I think it needs updates to varstored itself to, so that it knows where to load it from. If XAPI wrote out the cert to the wrong path though, I wonder whether varstored was ever able to pick it up though...

edwintorok (approved):

stormi:
There's no bug when the file is written. The bug is that it's written when the code intended that it's not (due to checking the presence of a file at the wrong path). If we were to change the default directory for varstored, varstored itself would need to be rebuilt with the path changed too, of course.

PR #4625, opened 2022-02-11
CA-363633: Always take the generation-id directly from xapi

robhoes:
The generation-id of a VM is made up by xapi - specifically for Windows
VMs - and xapi changes it or leaves it the same depending on the use
case. One such use case is revert-from-checkpoint, where the reverted VM
is expected to get a new generation-id.

The generation-id is put into the plaformdata by xapi and xenopsd
just writes it to xenstore when creating a domain. Because of a recent
change, the platformdata is now persisted by xenopsd across
suspend/resume and migration, and will therefore not take any new values
that set by xapi in the VM metadata.

Since a checkpoint is essentially a suspended VM, and a
revert-from-checkpoint results in a VM that can subsequently be resumed,
the VM now does not get a refreshed generation-id after this operation,
which is a bug.

To address this, the generation_id is given a dedicated field in the VM
metadata for xenopsd, so that it behaves as before. Xenopsd itself then
puts it into the platformdata when creating a domain.

### DIFF
--- a/ocaml/xenopsd/xc/xenops_server_xen.ml
+++ b/ocaml/xenopsd/xc/xenops_server_xen.ml
@@ -1332,11 +1332,20 @@ module VM = struct
       @ affinity
       @ weight
     in
+    let set_generation_id platformdata =
+      let key = "generation-id" in
+      match vm.generation_id with
+      | Some genid ->
+          (key, genid) :: List.remove_assoc key platformdata
+      | None ->
+          platformdata
+    in
     let default k v p = if List.mem_assoc k p then p else (k, v) :: p in
     let platformdata =
-      persistent.VmExtra.platformdata
+      persistent.VmExtra.platformdata @ vcpus
       |> default "acpi_s3" "0"
       |> default "acpi_s4" "0"
+      |> set_generation_id
     in
     let is_uefi =
       match ty with HVM {firmware= Uefi _; _} -> true | _ -> false
@@ -1356,7 +1365,7 @@ module VM = struct
     ; hap= hvm
     ; name= vm.name
     ; xsdata= vm.xsdata
-    ; platformdata= platformdata @ vcpus
+    ; platformdata
     ; bios_strings= vm.bios_strings
     ; has_vendor_device= vm.has_vendor_device
     ; is_uefi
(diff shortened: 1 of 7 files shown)


robhoes:
Still awaiting test results.

ocaml/xenopsd/xc/xenops_server_xen.ml
```diff
@@ -1332,9 +1332,16 @@ module VM = struct
       @ affinity
       @ weight
     in
+    let generation_id =
+      match vm.generation_id with
+      | Some genid ->
+          [("generation-id", genid)]
+      | None ->
```
edwintorok, line 1339: Will this handle migration correctly

robhoes: Migration also includes the VM metadata (`vm`). And if this metadata does not include a `generation_id`, in case it came from an older version, then the key will still be in the platformdata. But I'd better do something to ensure that there can't be two versions of the key in platformdata, just in case.

robhoes:
Tests passed.

ocaml/xenopsd/xc/xenops_server_xen.ml
```diff
@@ -1332,11 +1332,20 @@ module VM = struct
...
     in
+    let set_generation_id platformdata =
+      let key = "generation-id" in
+      match vm.generation_id with
+      | Some genid ->
+          (key, genid) :: (List.remove_assoc key platformdata)
+      | None ->
+          platformdata
+    in
     let default k v p = if List.mem_assoc k p then p else (k, v) :: p in
```
lindig, line 1343: This is existing code but `p` does not strike me as a good name; `assoc`, `kvs` would create a stronger sense of meaning for me. If `p` stands for `pairs` or `parameters`, `ps` would be better as well.

lindig (approved):

edwintorok (approved):

ocaml/xenopsd/xc/xenops_server_xen.ml
robhoes: Probably `platformdata`, since that is what is passed to it.

PR #4626, opened 2022-02-14
maintenance: remove warnings uncovered by ocaml 4.13.1

psafont:
There are many more, but this is significant chunk

### DIFF
--- a/ocaml/xapi-client/tasks.ml
+++ b/ocaml/xapi-client/tasks.ml
@@ -23,7 +23,7 @@ module TaskSet = Set.Make (struct
 end)
 
 (* Return once none of the tasks have a `pending status. *)
-let wait_for_all_inner ~rpc ~session_id ?all_timeout ~tasks =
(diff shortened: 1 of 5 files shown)


ocaml/xapi-client/tasks.ml
```diff
@@ -23,7 +23,7 @@ module TaskSet = Set.Make (struct
 end)
 
 (* Return once none of the tasks have a `pending status. *)
-let wait_for_all_inner ~rpc ~session_id ?all_timeout ~tasks =
+let wait_for_all_inner ~rpc ~session_id ~all_timeout ~tasks =
```
edwintorok, line 26: adding a `()` at the end might've been another way to fix it

psafont: Indeed, I thought making the `option` explicit rather than implicit would be a good thing for readability

lindig (approved):

edwintorok (approved):

PR #4628, opened 2022-02-16
certificate checking for clusterd

lindig:
Just to collect some comments - changes in xapi to enable certificate checking in clusterd. Clusterd uses the certificates used also by xapi. 

The protocol/IDL between xapi/clusterd and clusterd/clusterd changes to include a TLS configuration which holds the paths to the files containing the certificates. This simplifies the previous design which involved using single certificate for the cluster.


ocaml/xapi/message_forwarding.ml
```diff
@@ -1008,8 +1008,6 @@ functor
               debug "Pool.enable_tls_verification sleeping on fistpoint" ;
               Thread.delay 5.0
             done ;
-            Db.Cluster.get_all ~__context
-            |> List.iter (Xapi_cluster_helpers.Pem.maybe_write_new ~__context) ;
             all_hosts
             |> List.iter (fun host ->
                    do_op_on ~local_fn ~__context ~host (fun session_id rpc ->
```
edwintorok, line 1013: Since you loop over all hosts then this should be Client.Host.enable_tls_verification, not Client.Pool.enable_tls_verification (unrelated to clusterd changes, perhaps fix it in a separate commit).
If it was really a pool level operation then it should've taken a pool as an argument and do the looping internally (but you'd still need a host-level API to call).

edwintorok (approved):

ocaml/xapi-idl/cluster/cluster_interface.ml
```diff
@@ -79,6 +69,18 @@ let decode_cluster_config x =
 type cluster_config_and_all_members = cluster_config * all_members
 [@@deriving rpcty]
 
+type tls_config = {
+    server_pem_path: string  (** Path containing private and public keys *)
+  ; cn: string
+        (** CN used for verification, typically host UUID, ignored otherwise *)
```
psafont, line 75: If this is shared among the cluster shouldn't this real pool UUID instead of host UUID?

lindig: This is misleading. We are using something static because this is a pool-level configuration. Currently something like "api-cluster" or so. I have to look it up. Should update the comment.

psafont (approved):
I like this easier to reason about than the previous attempt, as well as the solution to certificate generation

PR #4629, opened 2022-02-17
Disable DMC

lindig:
Removing DMC by removing the feature flag and preventing any VM using DMC; such a VM cannot be started, imported, or migrated into a host.


ocaml/xapi/xapi_vm.ml
```diff
@@ -279,10 +279,28 @@ let assert_memory_constraints ~__context ~vm platformdata =
...
-   scheduled_to_be_resident_on atomically. This prevents concurrent API calls on the master from accounting for the
-   same VM twice during memory calculations to determine whether a given VM can start on a particular host..
-*)
+let enforce_memory_constraints ~__context ~vm _platformdata =
+  let module C = Xapi_vm_memory_constraints.Vm_memory_constraints in
+  let constraints = C.get ~__context ~vm_ref:vm in
+  try
+    C.assert_valid_and_pinned_at_static_max ~constraints
+      ~reason:"DMC unavailable"
+  with Api_errors.Server_error _ ->
```
edwintorok, line 288: Could we use `are_valid_and_pinned_at_static_max` directly so we don't need to raise/catch exceptions?

edwintorok (approved):
This seems safe for now: as long as the DMC feature flag is there the enforcement code won't run.

ocaml/xapi/xapi_vm.ml
```diff
@@ -376,6 +392,8 @@ let start ~__context ~vm ~start_paused ~force =
     Pool_features.assert_enabled ~__context ~f:Features.Network_sriov ;
   if not force then
     assert_memory_constraints ~__context ~vm vmr.API.vM_platform ;
+  if not (Pool_features.is_enabled ~__context Features.DMC) then
```
robhoes, line 395: This is too late, because the memory checks (to see if the VM fits on the host) have already been done at this point. This happens in `Xapi_vm_helpers.assert_can_boot_here` in the `assert_enough_memory_available` call, and this check is based on dynamic_min. It is called through `forward_to_suitable_host` (for `VM.start`) or `reserve_memory_for_vm` (for `VM.start_on`) in message_forwarding. Those functions are also used for operations like resume and migrate, so we can't call `enforce_memory_constraints` inside of them. The safest place for it is probably in `VM.start` and `VM.start_on` in message_forwarding.

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -1577,6 +1577,7 @@ let migration_type ~__context ~remote =
 let assert_can_migrate ~__context ~vm ~dest ~live ~vdi_map ~vif_map ~options
     ~vgpu_map =
   assert_licensed_storage_motion ~__context ;
+  Xapi_vm_helpers.assert_dmc_compatible ~__context ~vm ;
```
robhoes, line 1580: You also need this in `VM.pool_migrate` in message_forwarding, as this function isn't called for pool_migrate.

ocaml/xapi/import.ml
```diff
@@ -275,14 +275,26 @@ let assert_can_live_import __context rpc session_id vm_record =
...
       assert_memory_available ()
-  | `Paused ->
+  | `Paused when dmc_compatible ->
       assert_memory_available ()
-  | `Suspended ->
-      () (* disable DMC *)
-  | _other ->
+  | `Suspended when dmc_compatible ->
+      ()
+  | _ when not dmc_compatible ->
```
robhoes, line 288: May be easier to just do `if Xapi_vm_helpers.is_dmc_compatible_vmr ~__context ~vmr:vm_record then ...log and raise error...` above the match to avoid all the `when` clauses.

robhoes: Not also that this function is only called for cross-pool live migration. We need to also handle the normal upgrade case.

lindig: I think a flat `match` structure is easier to read and reason about  compared to nested `if` and it avoids the problem of multiple statements in `if` that need grouping. But it's a matter of taste.

lindig: I think I will try to unify the handling of live and other imports - take back some of the changes I made.

ocaml/xapi/import.ml
```diff
@@ -421,6 +421,21 @@ module VM : HandlerTools = struct
 
   let precheck __context config rpc session_id state x =
     let vm_record = get_vm_record x.snapshot in
+
+    (* we can't import a VM if it is not in a resting state and requires
+       DMC *)
+    let is_dmc_compatible =
```
robhoes, line 427: I think this block could also be written as
```
let is_dmc_compatible =
  vm_record.API.vM_power_state = `Halted
  || Xapi_vm_helpers.is_dmc_compatible_vmr ~__context ~vmr:vm_record
in
```

ocaml/xapi/import.ml
```diff
@@ -553,23 +574,28 @@ module VM : HandlerTools = struct
...
-            API.vM_memory_static_min= safe_constraints.static_min
-          ; vM_memory_dynamic_min= safe_constraints.dynamic_min
-          ; vM_memory_target= safe_constraints.target
-          ; vM_memory_dynamic_max= safe_constraints.dynamic_max
-          ; vM_memory_static_max= safe_constraints.static_max
-          }
-        ) else
-          vm_record
+        match vm_record.API.vM_power_state with
+        | `Paused | `Suspended | `Halted ->
```
robhoes, line 578: Only Halted here, because all the others mean "not at rest".

ocaml/xapi/import.ml
```diff
@@ -275,11 +275,15 @@ let assert_can_live_import __context rpc session_id vm_record =
            )
         )
   in
-  if
-    vm_record.API.vM_power_state = `Running
-    || vm_record.API.vM_power_state = `Paused
-  then
-    assert_memory_available ()
+  match vm_record.API.vM_power_state with
```
robhoes, line 278: I think we don't need this change anymore now, as the changes below apply to but live and non-live?

ocaml/xapi/xapi_db_upgrade.ml
```diff
@@ -230,64 +235,53 @@ let upgrade_wlb_configuration =
...
-            Db.VM.set_memory_target ~__context ~self:vm_ref
-              ~value:safe_constraints.VMC.target ;
-            Db.VM.set_memory_dynamic_max ~__context ~self:vm_ref
-              ~value:safe_constraints.VMC.dynamic_max ;
-            Db.VM.set_memory_static_max ~__context ~self:vm_ref
-              ~value:safe_constraints.VMC.static_max
+          match
+            (vm_rec.API.vM_is_control_domain, vm_rec.API.vM_power_state)
+          with
+          | true, _ ->
```
robhoes, line 251: I think this case is not needed anymore, and we don't need to special case control domains. We could do
```
if vm_rec.API.vM_is_control_domain || vm_rec.API.vM_power_state = `Halted then
```
with the code in the second case below. The control domains already shouldn't use DMC, but we can update them just in case. What is special about them is that they are always "Running" and changes will only take effect after a host reboot (see VM.set_memory_limits).

ocaml/xapi/xapi_db_upgrade.ml
```diff
@@ -230,64 +235,53 @@ let upgrade_wlb_configuration =
...
+              let target = vm_rec.API.vM_memory_target in
+              debug "VM %s (%s) dynamic_{min,max} <- %Ld" vm_rec.API.vM_uuid
+                vm_rec.API.vM_name_label target ;
+              Db.VM.set_memory_dynamic_min ~__context ~self:vm_ref ~value:target ;
+              Db.VM.set_memory_dynamic_max ~__context ~self:vm_ref ~value:target
+          | false, `Halted ->
+              (* Note this will also transform templates *)
```
robhoes, line 258: Can we use VMC.set below rather than setting individual fields? Or even just replace this whole thing with a call to `Xapi_vm_helpers.enforce_memory_constraints` to reduce duplication?

ocaml/xapi/import.ml
lindig: We need this match because based on power state we need to allocate memory or not. It was previously expressed with `if` and is now expressed with `match`.

robhoes: Yes, we need the old code. Changing it seems unnecessary, and the comment about DMC below misleading.

ocaml/xapi/xapi_db_upgrade.ml
```diff
@@ -238,7 +238,7 @@ let upgrade_wlb_configuration =
 let upgrade_vm_memory_for_dmc =
   {
     description= "Upgrading VM memory fields to disable DMC"
-  ; version= (fun x -> x <= yangtze)
+  ; version= (fun x -> true)
```
robhoes, line 241: This is ok, but `x < 751` would be sufficient.

robhoes (approved):

robhoes:
Ready to merge once testing is complete.

PR #4632, opened 2022-02-18
CA-363903: Winbind does not rotate keytab file

liulinC:
with `machine password timeout` and `kerberos method = secrets only`
winbind can rotate the machine password in winbind database,
Refer: https://www.samba.org/samba/docs/current/man-html/smb.conf.5.html
However, upgrade to future new arch with pure kerberos requires
machine password stored in keytab file.
This commit start a xapi periodic task to run `net changetrustpw`
which will rotate machine password both in winbind database and
system keytab file


lindig (approved):

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -994,6 +999,38 @@ module Winbind = struct
...
+    let now = Unix.time () in
+    let now_str = string_of_float now in
+    try
+      let machine_pwd_last_change_time =
+        (get_domain_info_from_db ()).machine_pwd_last_change_time
+      in
+      match machine_pwd_last_change_time with
+      | Some time when now < time +. !Xapi_globs.winbind_machine_pwd_timeout ->
+          ()
+      | _ ->
```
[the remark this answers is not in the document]
liulinC: yes, this is by design, 
if we does not find the machine password change time in xapi database (usually upgrade from PBIS), we will try to do a machine password rotate. (btw, this is not one of the critical logic)

ocaml/xapi/xapi_globs.ml
```diff
@@ -1013,7 +1013,7 @@ let xapi_globs_spec =
   ; ("max_active_sr_scans", Int max_active_sr_scans)
   ; ("winbind_debug_level", Int winbind_debug_level)
   ; ("winbind_cache_time", Int winbind_cache_time)
-  ; ("winbind_machine_pwd_timeout", Int winbind_machine_pwd_timeout)
+  ; ("winbind_machine_pwd_timeout", Float winbind_machine_pwd_timeout)
```
edwintorok, line 1016: This will likely need a XenRT test that sets this to a few minutes to check that it works, we're unlikely to have a testcase that runs for >7 days to find a bug in machine password rotation.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -994,6 +999,40 @@ module Winbind = struct
...
+          debug "%s started at %s" rotate_machine_password_task_name now_str ;
+          Helpers.call_script net_cmd
+            ["changetrustpw"; "--kerberos"]
```
edwintorok, line 1018: I don't think this is guaranteed to use the same KDC as XAPI, and if it doesn't then there will be a time where the machine password will be out of date: the KDC which 'net changetrustpw' used will have the new machine password and the other KDCs will have the old one. So depending which KDC you will use, you will either succeed further operations or fail because the machine password is wrong and out of date.
And if the KDCs are geographically distributed and slow to synchronize we know it can take 15m-30m until a password change update is propagates to all KDCs.

Is it possible to tell winbind here which KDC to prefer, so that at least we have a higher chance of this to succeed? Otherwise you'll get a 15m-30m window where authentications randomly fail or succeed.

ocaml/xapi/xapi_globs.ml
```diff
@@ -927,7 +927,7 @@ let winbind_debug_level = ref 2
 
 let winbind_cache_time = ref 60
 
-let winbind_machine_pwd_timeout = ref (7 * 24 * 3600)
+let winbind_machine_pwd_timeout = ref (7. *. 24. *. 3600.)
```
edwintorok, line 930: It might be best to handle this on a systemd timer, so you can also set the time. E.g. you may want to run this rotation during a maintenance window (see above: password rotations risk introducing a window where password authentications would fail, and you don't want that in the middle of your working day).

liulinC: windows DC keep latest 2 version of password, 
so password rotating does not break the authentication

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: I think both DC and samba/keytab keep the latest 2 version of password, so it is okay as long as we does not changeturst often (default to 2 weeks), then DC will sync the password .

ocaml/xapi/xapi_globs.ml
liulinC: yes, I agree.

liulinC: BTW, I found each `net changetrustpw` will override the `/var/lib/samba/lock/smb_krb5/krb5.conf.{DOMAIN}`, which have domains KDC, and I think Winbind will use this KDC. 
Thus, the KDC updated the machine password  (with latest Machine password) will always be used.

ocaml/xapi/extauth_plugin_ADwinbind.ml
edwintorok: OK, maybe it works differently for machine passwords, changing my user password used to be problematic.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -1118,6 +1084,105 @@ module ClosestKdc = struct
...
+  let generate_krb5_tmp_config ~domain ~kdc_fqdn =
+    (* Configure which server to change the password
+     * https://web.mit.edu/kerberos/krb5-devel/doc/admin/conf_files/krb5_conf.html *)
+    let open Xapi_stdext_unix in
+    let realm = String.uppercase_ascii domain in
+    let conf_contents =
+      String.concat "\n"
+        [
+          "# auto-generated by xapi"
+        ; "[libdefaults]"
```
edwintorok, line 1109: If we generate a krb5.conf then we also need to specify the ciphersuites to accept here. Normally we'd rely on winbind to generate krb5.conf for us.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -1083,6 +1084,105 @@ module ClosestKdc = struct
       Xapi_periodic_scheduler.remove_from_queue periodic_update_task_name
 end
 
+module RotateMachinePassword = struct
+  let task_name = "Rotating machine password"
+
+  let tmp_krb5_conf =
+    Printf.sprintf "%s/pwrotate.krb5" (Filename.get_temp_dir_name ())
```
lindig, line 1091: Is this risking a collision? This is global name - should this name not be unique for this invocation?

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -1083,6 +1084,105 @@ module ClosestKdc = struct
...
+    Unixext.atomic_write_to_file tmp_krb5_conf 0o0644 (fun fd ->
+        let (_ : int) = Unix.single_write_substring fd conf_contents 0 len in
+        Unix.fsync fd
+    )
+
+  let clear_tmp_krb5_conf () =
+    if !Xapi_globs.winbind_keep_configuration then
+      ()
+    else
+      Sys.remove tmp_krb5_conf
```
lindig, line 1129: This raises an exception if the file does not exist.

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -1083,6 +1084,105 @@ module ClosestKdc = struct
...
+              )
+              (fun () -> clear_tmp_krb5_conf ()) ;
+
+            Server_helpers.exec_with_new_task
+              "update machine password last change time"
+            @@ fun __context ->
+            update_extauth_configuration ~__context
+              ~k:"machine_pwd_last_change_time" ~v:now_str
+        | _ ->
+            debug "Xapi database does not has closest KDC, skip this rotate"
```
lindig, line 1173: language: skip this "rotation"

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: This file does not need to be unique as
1. this is not parallel task
2. it can be simply overrided.

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: Just updated with `include <winbind generated conf>`

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -1118,6 +1085,123 @@ module ClosestKdc = struct
...
+
+  let rotate () =
+    let now = Unix.time () in
+    let now_str = string_of_float now in
+    try
+      let {service_name; machine_pwd_last_change_time} =
+        get_domain_info_from_db ()
+      in
+      match machine_pwd_last_change_time with
+      | Some time when now < time +. !Xapi_globs.winbind_machine_pwd_timeout ->
```
edwintorok, line 1158: This will need some testing, but there might be some rounding issue here where we'll always miss the first call to rotate (it is just a bit too soon) and the interval between password rotations will be sometimes machine_password_timeout, sometimes twice that. Probably doesn't matter in practice though. (the scheduler and this code also use different clocks, monotonic vs wall clock)

edwintorok (approved):

ocaml/xapi/extauth_plugin_ADwinbind.ml
liulinC: this can be fix by like update the periodic interval to winbind_machine_pwd_timeout/2, 
but as you said, this does not matter in practice as rotating machine password is not in the critical path.

PR #4633, opened 2022-02-22
CP-38617: Add module with stubs for xenforeignmemory

psafont:
Adding these here, to xenctrlext instead of upstream allows us to iterate faster to adapt them to new features that we need.

Depends on https://github.com/xapi-project/xs-opam/pull/600 and https://github.com/xapi-project/xenctrl/pull/4

### DIFF
--- a/ocaml/xenforeign/dune
+++ b/ocaml/xenforeign/dune
@@ -1,0 +1,4 @@
+(executable
+  (name main)
+  (libraries xenopsd_xc hex)
+  )
--- a/ocaml/xenforeign/main.ml
+++ b/ocaml/xenforeign/main.ml
@@ -1,0 +1,27 @@
+module Xfm = Xenctrlext.Xenforeignmemory
+
+let usage_msg = "foreign-mapper <domid>"
+
+let domid = ref None
+
+let anon_fun param =
+  match !domid with None -> domid := Some (int_of_string param) | _ -> ()
+
+let defer f = Fun.protect ~finally:f
+
+let () =
+  Arg.parse [] anon_fun usage_msg ;
+  let the_domid = Option.get !domid in
+
+  let handle = Xfm.acquire None in
+  defer (fun () -> Xfm.release handle) @@ fun () ->
+  let prot = {Xfm.read= true; write= true; exec= false} in
+  let tpm_addr = 0x110000L in
+  Xfm.with_mapping handle the_domid prot [tpm_addr]
+    ~on_unmap_failure:(Fun.const ())
+  @@ fun mapping ->
+  let readable_region = Bigarray.Array1.(create Char C_layout 1024) in
+  for i = 0 to 1023 do
+    readable_region.{i} <- mapping.{i}
+  done ;
+  Hex.(hexdump (of_bigstring readable_region))
--- a/ocaml/xenopsd/c_stubs/xenctrlext_stubs.c
+++ b/ocaml/xenopsd/c_stubs/xenctrlext_stubs.c
@@ -19,6 +19,10 @@
 #include <string.h>
 #include <unistd.h>
 #include <xenctrl.h>
+#include <xenforeignmemory.h>
+#include <xentoollog.h>
+
+#include <sys/mman.h>
 
 #include <caml/mlvalues.h>
 #include <caml/memory.h>
@@ -28,12 +32,23 @@
 #include <caml/signals.h>
 #include <caml/callback.h>
 #include <caml/unixsupport.h>
+#include <caml/bigarray.h>
 
 #define _H(__h) ((xc_interface *)(__h))
 #define _D(__d) ((uint32_t)Int_val(__d))
 
 /* From xenctrl_stubs */
 #define ERROR_STRLEN 1024
+
+#define Xtl_val(x)(*((struct xentoollog_logger **) Data_custom_val(x)))
+#define Xfm_val(x)(*((struct xenforeignmemory_handle **) Data_abstract_val(x)))
+#define Addr_val(x)(*((void **) Data_abstract_val(x)))
+
+// Defined in OCaml 4.12: https://github.com/ocaml/ocaml/pull/9734
+#if OCAML_VERSION < 41200
+#define Some_val(v) Field(v, 0)
+#define Is_some(v) Is_block(v)
+#endif
 
 static void raise_unix_errno_msg(int err_code, const char *err_msg)
 {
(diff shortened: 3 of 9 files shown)


ocaml/xenforeign/dune
```diff
@@ -0,0 +1,4 @@
+(executable
```
edwintorok, line 1: Could we make this part of the quicktest? Then it will be run on a host where Xen is available and we can check it doesn't break.  (I think the quicktests can use a CoreOS VM, see quicktest_vm_lifecycle.ml)

ocaml/xenopsd/c_stubs/xenctrlext_stubs.c
```diff
@@ -394,6 +409,127 @@ CAMLprim value stub_xenctrlext_cputopoinfo(value xch)
...
+                log_handle = Xtl_val(Some_val(logger));
+        }
+
+        fmem = xenforeignmemory_open(log_handle, Int_val(open_flags));
+
+        if(fmem == NULL) {
+                caml_failwith("Error when opening foreign memory handle");
+        }
+
+        result = caml_alloc(1, Abstract_tag);
```
edwintorok, line 429: Not sure which way around the order would be better here, but if this raises an exception (e.g. OutOfMemory) then we leak a xenforeignmemory handle.
If we allocate first then at worst we have allocated an abstract tag that we do not use, but the GC should collect it, so perhaps doing the caml_alloc first would be better.

ocaml/xenopsd/c_stubs/xenctrlext_stubs.c
```diff
@@ -394,6 +409,127 @@ CAMLprim value stub_xenctrlext_cputopoinfo(value xch)
...
+CAMLprim value stub_xenforeignmemory_close(value fmem)
+{
+        CAMLparam1(fmem);
+        int retval;
+
+        retval = xenforeignmemory_close(Xfm_val(fmem));
+
+        if(retval < 0) {
+                caml_failwith("Error when closing foreign memory handle");
+        }
```
edwintorok, line 444: To protect again double close you could also then set Xfm_val(fmem) to NULL here after the call succeeds. Looking at the implementation of `xenforeignmemory_close` it'll just exist when its arg is null.
We could also explicitly test against NULL before calling it and raise Invalid_argument if it is NULL.

ocaml/xenopsd/xc/xenctrlext.ml
```diff
@@ -82,3 +82,65 @@ type cputopo = {core: int; socket: int; node: int}
 external numainfo : handle -> numainfo = "stub_xenctrlext_numainfo"
 
 external cputopoinfo : handle -> cputopo array = "stub_xenctrlext_cputopoinfo"
+
+module Xenforeignmemory : sig
+  type handle
+
+  type mapping =
```
edwintorok, line 89: maybe it'd be best to hide the type and provide some functions to read/write the data from a string.
Then we could check that it hasn't been freed yet, and the unsafe interface would only be internal to this module: it'd be impossible to use it incorrectly from the outside: we'd raise an exception.

ocaml/xenforeign/dune
psafont: To run successfully it needs a special xenserver build as well as a windows VM, I don't think we can add this to quicktest as it is. 

The code gets checked in the github CI using `make check`. I think this is OK for the time being, towards the end of the feature when it has stabilized in the main branch we can do this along the upstreaming of the bindings

ocaml/xenopsd/c_stubs/xenctrlext_stubs.c
lindig: This is C code, so how can there be an exception? The OCaml code has a `defer` that should ensure the handle is closed.

ocaml/xenopsd/c_stubs/xenctrlext_stubs.c
```diff
@@ -394,6 +409,127 @@ CAMLprim value stub_xenctrlext_cputopoinfo(value xch)
...
+        the_errno = errno;
+
+        free(arr);
+
+        if(retval == NULL) {
+                raise_unix_errno_msg(the_errno,
+                                "Error when trying to map foreign memory");
+        }
+
+        result = caml_ba_alloc_dims(
```
lindig, line 505: I don't know how this works in detail - but this call must no longer depend on `arr` because it was freed.

ocaml/xenopsd/xc/xenctrlext.mli
```diff
@@ -78,3 +78,33 @@ external vcpu_setaffinity_soft : handle -> domid -> int -> bool array -> unit
 external numainfo : handle -> numainfo = "stub_xenctrlext_numainfo"
 
 external cputopoinfo : handle -> cputopo array = "stub_xenctrlext_cputopoinfo"
+
+module Xenforeignmemory : sig
+  type handle
+
+  type mapping =
+    (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.Array1.t
```
lindig, line 86: Could this be made abstract?

lindig (approved):

ocaml/xenopsd/xc/xenctrlext.ml
robhoes: I agree, I think this type should be abstract.

ocaml/xenforeign/main.ml
```diff
@@ -0,0 +1,25 @@
...
+let anon_fun param =
+  match !domid with None -> domid := Some (int_of_string param) | _ -> ()
+
+let defer f = Fun.protect ~finally:f
+
+let () =
+  Arg.parse [] anon_fun usage_msg ;
+  let the_domid = Option.get !domid in
+
+  let handle = Xfm.open_handle None in
```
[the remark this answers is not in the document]
psafont: open_memory doesn't feel like it conveys what's being opened, it's a handle that's able to map memory between domains.

I could use another pair of words: create / destroy, what do you think?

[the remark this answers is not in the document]
lindig: My point was, any abstract value could be considered a handle. I would be fine with open/close but unfortunately open is a keyword. Anything like allocate/free, create/destroy, acquire/release that conveys some duality is fine.

ocaml/xenopsd/c_stubs/xenctrlext_stubs.c
```diff
@@ -394,6 +409,135 @@ CAMLprim value stub_xenctrlext_cputopoinfo(value xch)
...
+        value prot_flags, value pages)
+{
+        CAMLparam4(fmem, dom, prot_flags, pages);
+        CAMLlocal2(cell, result);
+        size_t i, pages_length;
+        xen_pfn_t *arr;
+        int prot, the_errno;
+        void *retval;
+
+        if (Field(prot_flags, 0) == Val_false &&
```
lindig, line 467: I am wondering if the first `if` is needed. Could we start with `prot = PRO_NONE` and then go through the remaining if's? What is the relation between `PROT_NONE` and `prot = 0`?

ocaml/xenopsd/xc/xenctrlext.ml
psafont: The lifetime of the mapping is already limited by the scope of the function `with_mapping`. 

While this interface is unsafe, it's not clear to me how it's going to be used. I would rather change the interface later on once we write code in xenops that handles these arrays to better know the needs for a safe interface

ocaml/xenopsd/c_stubs/xenctrlext_stubs.c
psafont: Indeed, `arr` is only used to feed the pages to Xen's function `xenforeignmemory_map`. It's not used here to allocate the bigarray that is being returned

ocaml/xenopsd/xc/xenctrlext.mli
psafont: I think the with_mapping already limits the lifetime of the bindings for this type and I don't know yet exactly how are the bytes going to be read. I'd rather keep it like it is now before restricting its usage

ocaml/xenopsd/c_stubs/xenctrlext_stubs.c
psafont: > What is the relation between `PROT_NONE` and `prot = 0`?

The short answer is that it depends on the OS. This is why I'd rather not make an assumption, even if on Linux `PROT_NONE` is `0`

ocaml/xenopsd/xc/xenctrlext.ml
```diff
@@ -82,3 +82,37 @@ type cputopo = {core: int; socket: int; node: int}
 external numainfo : handle -> numainfo = "stub_xenctrlext_numainfo"
 
 external cputopoinfo : handle -> cputopo array = "stub_xenctrlext_cputopoinfo"
+
+module Xenforeignmemory = struct
+  type handle
+
+  type map_addr
```
robhoes, line 89: Unused type?

robhoes (approved):

ocaml/xenforeign/dune
edwintorok: Lets think about the quicktest later. In theory it should be possible to add this device to a Linux VM too. If this depends on any changes in Xen then we may want to avoid running that quicktest by default until the branch is merged.
Shouldn't we do this development on a feature branch though? There are likely to be more revisions to this code, and it is unlikely to interfere with code on master. There is a `feature/vtpm` branch that I've created for this purpose.

ocaml/xenopsd/xc/xenctrlext.ml
```diff
@@ -82,3 +82,35 @@ type cputopo = {core: int; socket: int; node: int}
...
+  external unmap : handle -> mapping -> unit = "stub_xenforeignmemory_unmap"
+
+  let with_mapping (handle : handle) (domid : domid) (prot : prot)
+      (pages : Int64.t list) (f : mapping -> unit) =
+    let mapping = map handle domid prot pages in
+    Fun.protect
+      ~finally:(fun () ->
+        try unmap handle mapping
+        with Unix_error (errno, _) ->
+          Printf.printf "Error while unmapping memory: %s\n"
```
edwintorok, line 112: We shouldn't print the error message here, stdout is likely to go nowhere (/dev/null).
If we had access to a logging framework (e.g. Logs) we could call that here, but if the intention is to upstream these bindings we can't do that. Although if the intention is to upstream the bindings we can't print to stdout here either, and we can't raise in the `finally` either, and we don't want to ignore the exception either...

Perhaps add an optional parameter to `with_mapping`, e.g. `?on_unmap_failure` that can be used to log the exception.

psafont: Yeah, I've added the parameter to log it the user wants to, I've made the parameter optional.

ocaml/xenopsd/c_stubs/xenctrlext_stubs.c
psafont: The surprising thing is that for all the warnings there are in the [manua](https://ocaml.org/manual/intfc.html) this is not among them. Maybe this is exceedingly rare in the real world, but it still feels suspect to have this behaviour.

edwintorok (approved):

PR #4635, opened 2022-02-23
Hide credentials in repository configuration from log

minglumlu:
CA-362704: Hide proxy_username and proxy_password for repo proxy

This commit aims to hide the proxy_username and proxy_password of
repository from log file.

CA-362704: Remove credential related info from remote repository conf file

This commit removes credential related info from the remote repository
conf file as there might be proxy credentials (proxy_username and
proxy_password) and the temporary token file path in it.

### DIFF
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -43,6 +43,27 @@ let log_exn_continue msg f x =
 
 type log_output = Always | Never | On_failure
 
+let filter_patterns =
+  [
+    ( Re.Str.regexp "^\\(.*proxy_\\(username\\|password\\)=\\)\\(.*\\)$"
+    , "\\1(filtered)"
+    )
+  ]
+
+let filter_args args =
+  List.map
+    (fun arg ->
+      List.fold_left
+        (fun acc (r, t) ->
+          if Re.Str.string_match r acc 0 then
+            Re.Str.replace_matched t acc
+          else
+            acc
+        )
+        arg filter_patterns
+    )
+    args
+
 let call_script ?(log_output = Always) ?env ?stdin script args =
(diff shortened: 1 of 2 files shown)


ocaml/xapi/helpers.ml
```diff
@@ -43,6 +43,24 @@ let log_exn_continue msg f x =
 
 type log_output = Always | Never | On_failure
 
+let filter_args args =
+  let patterns =
+    ["^\\(.*proxy_username=\\)\\(.*\\)$"; "^\\(.*proxy_password=\\)\\(.*\\)$"]
```
lindig, line 48: These strings are static and thus `Re.Str.regexp` should be called only once for them and not at every invocation of `filter_args`. I also think this should be one pattern and not multiple such that the string being filtered is only scanned once.

lindig (approved):

ocaml/xapi/helpers.ml
lindig: This can be slightly simplified by using `{| .. |}` for the string because`\` does not need to be escaped.

robhoes (approved):

robhoes:
@minglumlu please squash the fixup.

ocaml/xapi/helpers.ml
```diff
@@ -43,6 +43,27 @@ let log_exn_continue msg f x =
...
+    , "\\1(filtered)"
+    )
+  ]
+
+let filter_args args =
+  List.map
+    (fun arg ->
+      List.fold_left
+        (fun acc (r, t) ->
+          if Re.Str.string_match r acc 0 then
```
edwintorok, line 58: Re.Str is not thread-safe, I thought we fixed that, but perhaps not. We should use Re.Posix or Re.Pcre

PR #4636, opened 2022-02-24
maintenance: Emit the same warnings across the toolstack and remove a few dozen of them

psafont:
This was hiding a large amount of compiler warnings, especially in xenopsd.

Now the flags are controlled by the dune file at the root of the project.

There are some exceptions to this rule for heavy ppx and autogenerated code where the warning are not easily avoidable and aren't useful.


ocaml/tests/alerts/test_alert_certificate_check.ml
```diff
@@ -26,24 +26,22 @@ let good_samples =
 
 let expiring_samples =
   [
-    ("20200302T02:00:00Z", 30, Api_messages.host_server_certificate_expiring_30)
-  ; ("20200301T02:00:00Z", 29, Api_messages.host_server_certificate_expiring_30)
-  ; ("20200215T02:00:00Z", 14, Api_messages.host_server_certificate_expiring_14)
-  ; ("20200208T02:00:00Z", 7, Api_messages.host_server_certificate_expiring_07)
-  ; ("20200201T02:00:00Z", 0, Api_messages.host_server_certificate_expiring_07)
+    ("20200302T02:00:00Z", Api_messages.host_server_certificate_expiring_30)
```
edwintorok, line 29: are we sure we want to remove these numbers here?

edwintorok: Ah I see, it wasn't used in the formatted message (it probably changed).

psafont: The numbers are already encoded in the name of the message and not used to generate the message, so pretty sure ;)

lindig (approved):
Looking over it, it looks good to me. Mostly prefixing unused parameters with `_` and some dead code removed.

robhoes (approved):

PR #4637, opened 2022-03-01
Bugfix and improvement for repository GPG key

minglumlu:
CA-364450: Fix YUM repo config for repo metadata checking

The '--setopt' of yum-config-manager requires specifying the repo name.
The bug fixed in this commit is '--setopt=repo_gpgcheck=1' should be
'--setopt=<repo-name>.repo_gpgcheck=1' in parameter of
'yum-config-manager'.

In this commit, additionally the `repo_gpgcheck` is moved to the initial
repo configuration file content which contains static configurations
usually.

CP-39209: Add new field 'gpgkey_name' in repository object

Prior to this series, the GPG key file name is coded as a global
configuration of XAPI. This will get the GPG key rotation difficult in
future.

This commit changes the configuration into XAPI repository object so
that it could be changed through XAPI API.

CP-39209: Use GPG key file path from repository object

Given the gpgkey name has been stored in repository object, this commit
retrieves the gpgkey name from the repository object rather than a
global XAPI configuration.

The global XAPI configuration now is a default one for backward
compatibility

### DIFF
--- a/ocaml/idl/datamodel_lifecycle.ml
+++ b/ocaml/idl/datamodel_lifecycle.ml
@@ -3,9 +3,13 @@ let prototyped_of_class = function
       None
 
 let prototyped_of_field = function
+  | "Repository", "gpgkey_path" ->
+      Some "22.10.0-next"
   | _ ->
       None
 
 let prototyped_of_message = function
+  | "Repository", "set_gpgkey_path" ->
+      Some "22.10.0-next"
   | _ ->
       None
--- a/ocaml/xapi/repository_helpers.ml
+++ b/ocaml/xapi/repository_helpers.ml
@@ -875,38 +875,45 @@ let remove_repo_conf_file repo_name =
   in
   Unixext.unlink_safe path
 
-let write_yum_config ?(source_url = None) binary_url repo_name =
+let write_yum_config ~source_url ~binary_url ~repo_gpgcheck ~gpgkey_path
+    ~repo_name =
   let file_path =
     Filename.concat !Xapi_globs.yum_repos_config_dir (repo_name ^ ".repo")
   in
-  let gpgkey_path =
-    Filename.concat !Xapi_globs.rpm_gpgkey_dir
-      !Xapi_globs.repository_gpgkey_name
+  let opt_repo_gpgcheck, opt_gpgcheck, opt_gpgkey =
+    let opt_gpgkey () =
+      let gpgkey_abs_path =
+        Filename.concat !Xapi_globs.rpm_gpgkey_dir gpgkey_path
+      in
+      if not (Sys.file_exists gpgkey_abs_path) then
+        raise
+          Api_errors.(
+            Server_error (internal_error, ["gpg key file does not exist"])
+          ) ;
+      if not ((Unix.lstat gpgkey_abs_path).Unix.st_kind = Unix.S_REG) then
+        raise
+          Api_errors.(
+            Server_error (internal_error, ["gpg key file is not a regular file"])
+          ) ;
+      Printf.sprintf "gpgkey=file://%s" gpgkey_abs_path
+    in
+    match (!Xapi_globs.repository_gpgcheck, repo_gpgcheck) with
+    | true, true ->
+        ("repo_gpgcheck=1", "gpgcheck=1", opt_gpgkey ())
+    | true, false ->
+        ("repo_gpgcheck=0", "gpgcheck=1", opt_gpgkey ())
+    | false, _ ->
+        ("repo_gpgcheck=0", "gpgcheck=0", "")
   in
-  if !Xapi_globs.repository_gpgcheck then (
-    if not (Sys.file_exists gpgkey_path) then
-      raise
-        Api_errors.(
-          Server_error (internal_error, ["gpg key file does not exist"])
-        ) ;
-    if not ((Unix.lstat gpgkey_path).Unix.st_kind = Unix.S_REG) then
-      raise
-        Api_errors.(
-          Server_error (internal_error, ["gpg key file is not a regular file"])
-        )
-  ) ;
   let content_of_binary =
     [
       Printf.sprintf "[%s]" repo_name
     ; Printf.sprintf "name=%s" repo_name
     ; Printf.sprintf "baseurl=%s" binary_url
     ; "enabled=0"
-    ; (if !Xapi_globs.repository_gpgcheck then "gpgcheck=1" else "gpgcheck=0")
-    ; ( if !Xapi_globs.repository_gpgcheck then
-          Printf.sprintf "gpgkey=file://%s" gpgkey_path
-      else
-        ""
-      )
+    ; opt_repo_gpgcheck
+    ; opt_gpgcheck
+    ; opt_gpgkey
     ]
   in
   let content_of_source =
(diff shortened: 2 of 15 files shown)


ocaml/xapi/repository_helpers.ml
```diff
@@ -875,38 +891,44 @@ let remove_repo_conf_file repo_name =
...
+        raise
+          Api_errors.(
+            Server_error (internal_error, ["gpg key file is not a regular file"])
+          ) ;
+      let opt_gpgkey' = Printf.sprintf "gpgkey=file://%s" gpgkey_path in
+      (* Will not check the metadata in local repository on cooridnator *)
+      if repo_gpgcheck then
+        ("repo_gpgcheck=1", "gpgcheck=1", opt_gpgkey')
+      else
+        ("repo_gpgcheck=0", "gpgcheck=1", opt_gpgkey')
```
edwintorok, line 919: Why would we ever want to not check the GPG key? I think the repository should always be signed with a key, even if just a test key and for testing purposes we could point the host to the right one.
There is a danger here that we might accidentally turn off GPG verification off in production for the sake of testing, I'd rather we always have gpgcheck=1 in the config.
If we really have to then we can add a single boolean in xapi.conf to override this and read it at the point we generate the file instead of plumbing it through lots of function calls with default values

edwintorok (changes requested):

ocaml/xapi/repository_helpers.ml
minglumlu: For the 'gpgcheck', this was originated from very early stage we were using dummy packages for testing only purpose. And I believe it is being used still.
The current logic is just to be determined by `!Xapi_globs.repository_gpgcheck` which is a config can be configured in xapi.conf.

For the 'repo_gpgcheck', this is used to check meta data of a repo. This works when Toolstack syncs repo from remote repository. After syncing, Toolstack creates a local repository. During this process, the signing is not done for simplicity (otherwise a key has to be managed). This results that when the members of the pool sync with the local repository on the coordinator, the metadata will not be checked, but 'gpgcheck=1' means the RPM packages are to be checked.

ocaml/xapi/repository_helpers.ml
```diff
@@ -875,38 +875,44 @@ let remove_repo_conf_file repo_name =
   in
   Unixext.unlink_safe path
 
-let write_yum_config ?(source_url = None) binary_url repo_name =
+let write_yum_config ?(source_url = None) ?(repo_gpgcheck = true) binary_url
+    repo_name gpgkey_name =
```
psafont, line 895: named arguments is better since the arguments share type and can be easily switched

ocaml/xapi/repository_helpers.ml
```diff
@@ -875,38 +875,44 @@ let remove_repo_conf_file repo_name =
...
   let file_path =
     Filename.concat !Xapi_globs.yum_repos_config_dir (repo_name ^ ".repo")
   in
-  let gpgkey_path =
-    Filename.concat !Xapi_globs.rpm_gpgkey_dir
-      !Xapi_globs.repository_gpgkey_name
+  let opt_repo_gpgcheck, opt_gpgcheck, opt_gpgkey =
```
psafont, line 899: This block seems quite convoluted, it would seem like matching values to construct the tuple would be more organized:

```ocaml
  let opt_gpgkey () =
    if not (Sys.file_exists gpgkey_path) then
        raise
          Api_errors.(
            Server_error (internal_error, ["gpg key file does not exist"])
          ) ;
    else if not ((Unix.lstat gpgkey_path).Unix.st_kind = Unix.S_REG) then
        raise
          Api_errors.(
            Server_error (internal_error, ["gpg key file is not a regular file"])
          ) ;
  else
      Printf.sprintf "gpgkey=file://%s" gpgkey_path
  in
  match !Xapi_globs.repository_gpgcheck, repo_gpgcheck with
  | false, _ ->
      ("repo_gpgcheck=0", "gpgcheck=0", "")
  | true, false ->
      ("repo_gpgcheck=0", "gpgcheck=1", opt_gpgkey ())
  | true, true ->
      ("repo_gpgcheck=1", "gpgcheck=1", opt_gpgkey ())
```

minglumlu:
Rebased due to conflict on the DB schema.
Appended 3 commits to resolve issues in comments.

lindig (changes requested):
Using `_name` suffix for a file name parameter is confusing and `_path` would be better.

minglumlu:
Renamed the gpgkey_name to gpgkey_path for better naming.
The commits were sorted out as appending new fixups would be messed up.

ocaml/xapi/repository_helpers.ml
```diff
@@ -817,6 +817,22 @@ let assert_url_is_valid ~url =
...
+let gpgkey_path_pattern = Re.Str.regexp "^[A-Za-z0-9_-]+$"
+
+let assert_gpgkey_path_is_valid path =
+  (* When Xapi_globs.repository_gpgcheck is:
+   * true, an empty gpgkey path will
+   *   1) reuslt to use default one which is configured in repository-gpgkey-path, or,
+   *   2) raise an error to user if no default one configured;
+   * false, an empty gpgkey path will be ignored.
+   * The existence and validaty of the GPG public key file will be verified before using *)
+  if path = "" || Re.Str.string_match gpgkey_path_pattern path 0 then
```
lindig, line 829: Avoid the `Re.Str` module because its thread safety is not clear https://discuss.ocaml.org/t/re-str-is-not-thread-safe/9425/17 and use `Re.Posix`

lindig (approved):

ocaml/xapi/repository_helpers.ml
psafont: Re might be not thread-safe after all. I'd propose instead:

```ocaml
let gpgkey_path_is_valid = function
| 'A' .. 'Z' | 'a'..'z' | '0'..'9' | '_' | '-' -> true
| _ -> false

let assert_gpgkey_path_is_valid path =
  if path = "" || String.forall gpgkey_path_is_valid path then
```
Technicallthe empty string check could be removed and still be valid, but it's better to have it explicitly.

ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -3,9 +3,13 @@ let prototyped_of_class = function
       None
 
 let prototyped_of_field = function
+  | "Repository", "gpgkey_path" ->
+      Some "22.10.0.4.gbf8e4e52c-next"
```
robhoes, line 7: Hmm... this should've been `22.10.0-next`. So I made a mistake in my PR about versions.

ocaml/xapi/repository_helpers.ml
```diff
@@ -817,16 +817,20 @@ let assert_url_is_valid ~url =
     error "Invalid url %s: %s" url (ExnHelper.string_of_exn e) ;
     raise Api_errors.(Server_error (invalid_base_url, [url]))
 
-let gpgkey_path_pattern = Re.Str.regexp "^[A-Za-z0-9_-]+$"
+let is_gpgkey_path_valid = function
```
lindig, line 820: The name is misleading - this functions does not check the validity of a path but is a predicate over characters. The predicate does not include `/` and not `.` - both are typically found in filenames and paths - is this correct?

minglumlu: It's a file name under a specific path `/etc/pki/rpm-gpg/`.

robhoes: Right, and I see that we asked you to change "name" into "path" :(

robhoes (approved):

ocaml/xapi/repository_helpers.ml
robhoes: And also the predicate was a suggestion in the comments above.

PR #4638, opened 2022-03-01
Add pool.set_wlb_enabled permission for client auth

minglumlu:
Authentication with client certificate requires permission to run
'pool.set_wlb_enabled'. This commit is just for this.


robhoes (approved):

psafont (approved):

psafont:
Does it need a change in the lifetime to show in the SDK that the roles have changed?

robhoes:
>Does it need a change in the lifetime to show in the SDK that the roles have changed?

No I don't think so in this case.

PR #4639, opened 2022-03-01
maintenance: remove most usages of Re.Str

psafont:
Re.Str does not seem thread-safe as it caches state. Replace with
String, Astring or Re.Posix functions which are safe.

The remaining usages are in single-threaded cli applicacions where the
usage is not an issue.

I've manually verified the substitutions regarding the behaviour of empty strings and whether the position is the start of the matched group or the end


ocaml/xapi/repository_helpers.ml
```diff
@@ -104,13 +104,13 @@ module Pkg = struct
      *   "qemu-dp-2:2.12.0-2.0.11.x86_64", or
      *   "time-(none):1.7-45.el7.x86_64".
      * These may come from "yum updateinfo list updates" and "rpm -qa" *)
-    let r = Re.Str.regexp "^\\(.+\\)\\.\\(noarch\\|x86_64\\)$" in
-    match Re.Str.string_match r s 0 with
-    | true -> (
+    let r = Re.Posix.compile_pat {|^(.+).(noarch|x86_64)$|} in
```
lindig, line 107: In the original, the dot before `noarch` is escaped - it is not here - so it is not a literal dot.

lindig (commented):
I am always a bit unclear what signifies grouping, alternatives, or non-empty repetition in the various flavors of regular expressions. Some require using `\+` while others use `+`, for example.

ocaml/xapi/repository_helpers.ml
psafont: This is unit-tested, I'm suprised it doesn't provoke any failures

ocaml/vhd-tool/src/cohttp_unbuffered_io.ml
```diff
@@ -100,7 +100,7 @@ let rec read_line ic =
       read_line ic
   | Some buf, i when i < String.length buf -> (
     try
-      let eol = Re.Str.search_forward crlf buf i in
+      let eol = Re.Group.start (Re.exec ~pos:i crlf buf) 0 in
```
edwintorok, line 103: I don't think we necessarily need a regex here, this just looks for the `\r\n` substring, no alternatives are usd. Would something like Astring's String.find_sub work here?

ocaml/xapi-idl/example/example.ml
```diff
@@ -38,9 +38,7 @@ let resources =
 
 let socket_path = ref "/var/xapi/socket"
 
-let comma = Re.Str.regexp_string ","
-
-let csv = Re.Str.split_delim comma
+let csv = Astring.String.cuts ~sep:"," ~empty:false
```
edwintorok, line 41: This could use `String.split_on_char` since we're splitting only on a single character.

ocaml/xapi/repository_helpers.ml
```diff
@@ -104,13 +104,13 @@ module Pkg = struct
      *   "qemu-dp-2:2.12.0-2.0.11.x86_64", or
      *   "time-(none):1.7-45.el7.x86_64".
      * These may come from "yum updateinfo list updates" and "rpm -qa" *)
-    let r = Re.Str.regexp "^\\(.+\\)\\.\\(noarch\\|x86_64\\)$" in
-    match Re.Str.string_match r s 0 with
-    | true -> (
+    let r = Re.Posix.compile_pat {|^(.+)\.(noarch|x86_64)$|} in
```
edwintorok, line 107: This could simply split the string from reverse on `.` at most once and then match `("noarch" | "x86_64" as arch)`. No need for a regex here. E.g. `Astring.String.cut ~rev:true` could be used.
No regex needed.

ocaml/xapi/repository_helpers.ml
```diff
@@ -191,20 +191,20 @@ module Pkg = struct
      *)
     let normalize v =
       let split_letters_and_numbers s =
```
edwintorok, line 193: This parsing looks very fragile and should probably be reimplemented according to official docs on version strings instead of guessing based on number/letter splits. Also it is likely to reach a different conclusion than RPM/Yum would, so it would be best to ask Yum/RPM about this (there is an rpmvercmp executable if we have to use it) rather than reimplementing this ourselves.

For now lets just fix the regex parts though.

edwintorok (commented):
Thanks for addressing this. As discussed on the OCaml forums Re itself may not be entirely thread-safe either, so more changes might be needed here in the future  (e.g. wrapping each Re call with a mutex belonging to that regex)

ocaml/xapi/repository_helpers.ml
psafont: Added a test that detects the failure

ocaml/xapi-idl/example/example.ml
psafont: Unfortunately `split_on_char` allows for empty strings in the list which is not the case for these two.

robhoes (approved):

ocaml/xapi/repository_helpers.ml
```diff
@@ -104,14 +104,9 @@ module Pkg = struct
      *   "qemu-dp-2:2.12.0-2.0.11.x86_64", or
      *   "time-(none):1.7-45.el7.x86_64".
      * These may come from "yum updateinfo list updates" and "rpm -qa" *)
-    let r = Re.Str.regexp "^\\(.+\\)\\.\\(noarch\\|x86_64\\)$" in
-    match Re.Str.string_match r s 0 with
-    | true -> (
+    match Astring.String.cut ~rev:true ~sep:"." s with
```
lindig, line 107: Does this work as expected? A package can have more dots and I would expect the result to have a variable list of segments that need to be matched here. The original regexp was anchored at the end and this is why it worked.

ocaml/xapi/repository_helpers.ml
lindig: I agree that this looks fragile but also with focusing on replacing `Re.Str`.

lindig (approved):

ocaml/xapi/repository_helpers.ml
psafont: Yep: `~rev:true` makes the cut happen at the the rightmost dot of the string

edwintorok (approved):

PR #4640, opened 2022-03-02
Add [post|put]_services_xenops to client auth permission list

minglumlu:
These two methods are used when performing VM.pool_migrate to
migrate VMs from member hosts to coordinator host.

The clients authenticated with certificates require permission to use
these methods. This commit extends the permission list of client auth
with these two methods.


lindig (approved):
The only change is the extended list of permissions? This is not reflected in the title of this commit.

PR #4641, opened 2022-03-03
Fix random insert of fist point

lindig:
The cert distribution code provides a fist point for testing. The fist
point takes a list of operations and inserts randomly a failure. We have
seen logs where that failure was inserted at the end - which does not
make sense for testing. Replace the code that inserts the operations to
make sure it never inserts the failure at the end unless the list is
empty.

### DIFF
--- a/ocaml/xapi/cert_distrib.ml
+++ b/ocaml/xapi/cert_distrib.ml
@@ -364,15 +364,32 @@ let collect_pool_certs ~__context ~rpc ~session_id ~map ~from_hosts =
          map cert
      )
 
-let take_and_append n x xs =
-  (* take_and_append 3 10 [1;2;3;4] = [1;2;3;10] *)
-  let rec loop i acc = function
-    | x :: xs when i < n ->
-        loop (i + 1) (x :: acc) xs
-    | _ ->
-        x :: acc |> List.rev
+(* insert an element randomly *between* existing elements of a list.
+   We are enumerating the gaps between elements in a list:
+
+   List:     a b c d
+   Gaps:    0 1 2 3 4
+
+   A list of length n has gaps 0, 1, ...,n. We are selecting a gap
+   between 1 and n-1 to insert an element - so never at the beginning or
+   end if n >= 2. Otherwise we insert at the front.
+
+   This function is not tail recursive, so only apply to short lists *)
+
+let randomly_insert x ys =
+  let gap = Random.int (List.length ys) + 1 in
+  let rec loop n xs =
+    match (n, xs) with
+    | _, [] ->
+        [x]
+    | _, [y] ->
+        [x; y] (* never insert at the end *)
+    | 0, ys ->
+        x :: ys
+    | n, y :: ys ->
+        y :: loop (n - 1) ys
   in
-  loop 0 [] xs
+  loop gap ys
 
 let exchange_certificates_in_pool ~__context =
   (* here we coordinate the certificate distribution. from a high level
@@ -403,7 +420,6 @@ let exchange_certificates_in_pool ~__context =
     | Some seed ->
         fun ops ->
           Random.init seed ;
-          let rand_i = Random.int (List.length ops) in
           let throw_op =
             ( "FIST"
             , fun () ->
@@ -416,7 +432,7 @@ let exchange_certificates_in_pool ~__context =
                   )
             )
           in
-          let ops' = take_and_append rand_i throw_op ops in
+          let ops' = randomly_insert throw_op ops in
           D.debug "exchange_certificates_in_pool: we are about to..." ;
           List.iteri (fun i (desc, _) -> D.debug "%d. %s" i desc) ops' ;
           ops'


ocaml/xapi/cert_distrib.ml
```diff
@@ -403,7 +420,6 @@ let exchange_certificates_in_pool ~__context =
     | Some seed ->
         fun ops ->
           Random.init seed ;
-          let rand_i = Random.int (List.length ops) in
```
robhoes, line 406: Isn't it enough to use `List.length ops - 1` here?

lindig: I don't like the existing function to take a position where it inserts - this should be the responsibility of the function. I think the existing semantics are not very clear and that is why I rewrote it.

ocaml/xapi/cert_distrib.ml
```diff
@@ -364,15 +364,32 @@ let collect_pool_certs ~__context ~rpc ~session_id ~map ~from_hosts =
...
+   end if n >= 2. Otherwise we insert at the front.
+
+   This function is not tail recursive, so only apply to short lists *)
+
+let randomly_insert x ys =
+  let gap = Random.int (List.length ys) + 1 in
+  let rec loop n xs =
+    match (n, xs) with
+    | _, [] ->
+        [x]
```
edwintorok, line 384: If you want to never insert at the end, shouldn't this be the empty list `[]`?

ocaml/xapi/cert_distrib.ml
```diff
@@ -364,15 +364,32 @@ let collect_pool_certs ~__context ~rpc ~session_id ~map ~from_hosts =
...
+   Gaps:    0 1 2 3 4
+
+   A list of length n has gaps 0, 1, ...,n. We are selecting a gap
+   between 1 and n-1 to insert an element - so never at the beginning or
+   end if n >= 2. Otherwise we insert at the front.
+
+   This function is not tail recursive, so only apply to short lists *)
+
+let randomly_insert x ys =
+  let gap = Random.int (List.length ys) + 1 in
```
psafont, line 380: I'm weary of puttin the random pick far away from the seed initialization as it was before

This would allow to call an insert function with clearly defined semantic which should be clearer to understand:
```ocaml
let insert ~at x xs =
  let front, back = Xapi_stdext_std.Listext.List.chop at xs in
  List.concat [front; x :: back]
``` 

this can be used with
```ocaml
let at = match List.length with 
  | 0 -> 0
  | len ->
(* disallow fistpoints at the end of the operation list*)
Random.int (len - 1)
in
let ops' = insert ~at throw_op ops
```

ocaml/xapi/cert_distrib.ml
```diff
@@ -364,15 +364,32 @@ let collect_pool_certs ~__context ~rpc ~session_id ~map ~from_hosts =
...
-    | _ ->
-        x :: acc |> List.rev
+(* insert an element randomly *between* existing elements of a list.
+   We are enumerating the gaps between elements in a list:
+
+   List:     a b c d
+   Gaps:    0 1 2 3 4
+
+   A list of length n has gaps 0, 1, ...,n. We are selecting a gap
+   between 1 and n-1 to insert an element - so never at the beginning or
```
robhoes, line 374: Why not at the beginning for n >= 2, while it is allowed for n = 1?

ocaml/xapi/cert_distrib.ml
robhoes: I actually thought the old function was easier to understand than the new one.

ocaml/xapi/cert_distrib.ml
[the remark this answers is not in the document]
psafont: To prevent the function from throwing an error when the length of the list is 0

ocaml/xapi/cert_distrib.ml
lindig: This is a matter of definition. If we want to insert something, we have no choice here. I don't think the edge case matters for how we use it but I rather not drop any element.

ocaml/xapi/cert_distrib.ml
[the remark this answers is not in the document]
lindig: Because we never want to insert at 0 - the front, unless we can't avoid it. See the comment - we pick a gap from 1 .. n-1.

ocaml/xapi/cert_distrib.ml
lindig: Unless you drop the element to be inserted, you don't have a choice when you have `[]` or `[a]`. I believe this edge case should never come up but I want to handle it reasonably.

ocaml/xapi/cert_distrib.ml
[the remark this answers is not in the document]
robhoes: Yes that's why I expected a `- 1`

lindig: I'm fine with this but again believe it is difficult to see what happens around edge cases and this is where the previous problem crept in.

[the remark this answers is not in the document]
psafont: The call `Random.int 0` throws an `Invalid_argument`, this should be guarded against.

ocaml/xapi/cert_distrib.ml
robhoes: The old function allows adding the element to the beginning for any length. I think it's valid in the context where we use it, making the call fail right away.

ocaml/xapi/cert_distrib.ml
psafont: The previous code didn't try to handle edge-cases at all. I don't think mixing selection and insertion code is the solution rather, but rather tackling the edge-cases at the selection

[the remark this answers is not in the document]
robhoes: Main point is that `gap` isn't actually the gap where we are inserting, which confused me. Because `gap` cannot be 0, but sometimes we do insert at 0; and `gap` can be = n, but we only insert at gaps < n.

ocaml/xapi/cert_distrib.ml
robhoes: Actually... The bound of `Random.int` is excluded, so we should already not take the last element. So why is this wrong or is there another problem?

lindig:
{noformat}
Mar  1 13:03:56 xrtuk-01-02 xapi: [debug||2128 HTTP 127.0.0.1->:::80|pool.enable_tls_verification D:df00f9ca0f7a|cert_distrib] exchange_certificates_in_pool: we are about to...
Mar  1 13:03:56 xrtuk-01-02 xapi: [debug||2128 HTTP 127.0.0.1->:::80|pool.enable_tls_verification D:df00f9ca0f7a|cert_distrib] 0. send certs to b3c38966
Mar  1 13:03:56 xrtuk-01-02 xapi: [debug||2128 HTTP 127.0.0.1->:::80|pool.enable_tls_verification D:df00f9ca0f7a|cert_distrib] 1. send certs to c15558a3
Mar  1 13:03:56 xrtuk-01-02 xapi: [debug||2128 HTTP 127.0.0.1->:::80|pool.enable_tls_verification D:df00f9ca0f7a|cert_distrib] 2. FIST
{noformat}

In a pool with 2 members we have these logs which suggest that in the list of actions a fist point was inserted after the distribution was done already. So either there were more elements in the list, which got dropped, or the fist point was inserted at the end.

robhoes:
> So either there were more elements in the list, which got dropped, or the fist point was inserted at the end.

This is important, because the old function does indeed drop anything after the inserted error. So perhaps the test problem was not that an error was added at the end.

lindig:
This does not increase my confidence in the existing code and test strategy. We insert a failure but don't document where it was inserted because we remove everything that came after it. But this PR might not be the solution either.

robhoes:
>We insert a failure but don't document where it was inserted because we remove everything that came after it

Yes, it would be useful to log the full list for clarity.

lindig:
Let's close this - there are too many problems in the existing test and my proposed solution.

PR #4642, opened 2022-03-07
Fist point of cert exchange: keep all operations

lindig:
For testing we add at a random point in a list of operations a failure.
The existing code removed all operations after the inserted failure.
This commit changes this to keep them for debugging. They won't be
executed, though, because the failure is reached first.


lindig:
The failure is coming from the warnings gate but I don't see what warning I would have missed.

robhoes:
Looks good.

A test failed, which may be solved by a rebase.

robhoes:
No, the branch already on the latest commit :/

lindig:
But there are no warnings in the file that I touched. Could this has been introduced before?

lindig:
I have cherry-picked Pau's commit

robhoes (approved):

PR #4644, opened 2022-03-08
CP-38626:Allow manipulation of VTPM contents

psafont:
This is restricted to local access only, through the UNIX socket.

This allows the software TPM daemon to retrieve and push the contents of
the TPM from and to the xapi database.

Any remote accesses to the contents are forbidden

### DIFF
--- a/ocaml/idl/datamodel_vtpm.ml
+++ b/ocaml/idl/datamodel_vtpm.ml
@@ -16,17 +16,47 @@ open Datamodel_types
 open Datamodel_common
 open Datamodel_roles
 
+let get_contents =
+  call ~name:"get_contents" ~in_product_since:"rel_next"
+    ~doc:"Obtain the contents of the TPM" ~secret:true
+    ~params:[(Ref _vtpm, "self", "The VTPM reference")]
+    ~result:(String, "The contents") ~hide_from_docs:true
+    ~allowed_roles:_R_LOCAL_ROOT_ONLY ()
+
+let set_contents =
+  call ~name:"set_contents" ~in_product_since:"rel_next"
+    ~doc:"Introduce new contents for the TPM" ~secret:true
+    ~params:
+      [
+        (Ref _vtpm, "self", "The VTPM reference")
+      ; (String, "contents", "The new contents")
+      ]
+    ~hide_from_docs:true ~allowed_roles:_R_LOCAL_ROOT_ONLY ()
+
 let t =
   create_obj ~in_db:true ~in_oss_since:oss_since_303 ~persist:PersistEverything
-    ~lifecycle:[(Published, rel_rio, "Added VTPM stub")]
+    ~lifecycle:
+      [
+        (Published, rel_rio, "Added VTPM stub")
+      ; (Extended, rel_next, "Added ability to manipulate contents")
+      ; (Extended, rel_next, "Added VTPM profiles")
+      ; (Changed, rel_next, "Removed backend field")
+      ]
     ~gen_constructor_destructor:true ~name:_vtpm ~descr:"A virtual TPM device"
     ~gen_events:false ~doccomments:[]
-    ~messages_default_allowed_roles:_R_VM_ADMIN ~messages:[]
+    ~messages_default_allowed_roles:_R_POOL_ADMIN
     ~contents:
       [
         uid _vtpm
-      ; field ~qualifier:StaticRO ~ty:(Ref _vm) "VM" "the virtual machine"
-      ; field ~qualifier:StaticRO ~ty:(Ref _vm) "backend"
-          "the domain where the backend is located"
+      ; field ~qualifier:StaticRO ~ty:(Ref _vm) "VM"
+          "The virtual machine the TPM is attached to"
+      ; field ~qualifier:DynamicRO
+          ~ty:(Map (String, String))
+          ~lifecycle:[(Published, rel_next, "Added VTPM profiles")]
+          "profile" "The security properties that define how the TPM is handled"
+      ; field ~qualifier:DynamicRO ~ty:(Ref _secret) ~internal_only:true
+          ~lifecycle:[(Published, rel_next, "Added VTPM contents")]
+          "contents" "The contents of the TPM"
       ]
+    ~messages:[get_contents; set_contents]
     ()
--- a/ocaml/idl/datamodel_vtpm.ml
+++ b/ocaml/idl/datamodel_vtpm.ml
@@ -1,0 +1,32 @@
+(*
+   Copyright (C) Citrix Systems Inc.
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU Lesser General Public License as published
+   by the Free Software Foundation; version 2.1 only. with the special
+   exception on linking described in file LICENSE.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU Lesser General Public License for more details.
+ *)
+
+open Datamodel_types
+open Datamodel_common
+open Datamodel_roles
+
+let t =
+  create_obj ~in_db:true ~in_oss_since:oss_since_303 ~persist:PersistEverything
+    ~lifecycle:[(Published, rel_rio, "Added VTPM stub")]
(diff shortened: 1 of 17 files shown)


ocaml/idl/datamodel_vtpm.ml
```diff
@@ -0,0 +1,62 @@
...
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU Lesser General Public License for more details.
+ *)
+
+open Datamodel_types
+open Datamodel_common
+open Datamodel_roles
+
+let get_contents =
+  call ~name:"get_contents" ~in_product_since:"rel_next"
```
lindig, line 20: The result is a string but the name is plural, suggesting some collection. I would have use `content`; an alternative would be `secret` - because this is what this data is.

ocaml/idl/datamodel_vtpm.ml
```diff
@@ -0,0 +1,62 @@
...
+    ~messages_default_allowed_roles:_R_POOL_ADMIN
+    ~contents:
+      [
+        uid _vtpm
+      ; field ~qualifier:StaticRO ~ty:(Ref _vm) "VM"
+          "The virtual machine the TPM is attached to"
+      ; field ~qualifier:DynamicRO
+          ~ty:(Map (String, String))
+          ~lifecycle:[(Published, rel_next, "Added VTPM profiles")]
+          "profile" "The security properties that define how the TPM is handled"
```
lindig, line 56: This is a list - what is the semantics of duplicates?

lindig (approved):

ocaml/idl/datamodel_vtpm.ml
psafont: It should not contain any duplicates. It's awkward that it's handled by tuple lists internally, I'll try to see whether there's a more fitting type.

ocaml/xapi-cli-server/records.ml
```diff
@@ -4976,3 +4976,37 @@ let repository_record rpc session_id repository =
...
       ]
   }
+
+let vtpm_record rpc session_id vtpm =
+  let _ref = ref vtpm in
+  let empty_record =
+    ToGet (fun () -> Client.VTPM.get_record rpc session_id !_ref)
+  in
+  let record = ref empty_record in
+  let x () = lzy_get record in
```
lindig, line 4986: I wasn't aware of this lazy business and this code is repeated over and over in this module. I wonder why the OCaml Lazy module could not have been used for this. Now it's too late because we have this record but it looks convoluted.

lindig (approved):

ocaml/xapi-cli-server/records.ml
psafont: This module also has hundreds of calls to functions without labelling their arguments, it's in the best status regarding maintenance.

Last time I did a cleanup here around inconsistent display of lists xenrt broke so I'm a bit wary of being too aggressive here

psafont:
The functions restricted to local work when this python script is run locally:

```python
import XenAPI

x = XenAPI.xapi_local()
x.login_with_password('','')
vm = x.xenapi.VM.get_all()[0]

vtpm = x.xenapi.VTPM.create({"VM": vm})
x.xenapi.VTPM.set_contents(vtpm, "hello, world")

print "vtpm " + vtpm + " created with contents:"
print x.xenapi.VTPM.get_contents(vtpm)

# x.xenapi.VTPM.destroy(vtpm)
```

```
# python test-vtpm-api-access.py 
vtpm OpaqueRef:97c336ec-5e4d-4c4f-8ce0-6d96a9fbadc6 created with contents:
hello, world
# xe vtpm-list
uuid ( RO)       : 6d34bbd3-84ae-f0e1-3635-dfb899e54c6c
         vm ( RO): 2cf37285-57bc-4633-a24f-0c6c825dda66
    profile ( RO): 

```

ocaml/xapi-cli-server/records.ml
```diff
@@ -2431,6 +2431,9 @@ let vm_record rpc session_id vm =
               (x ()).API.vM_pending_guidances
           )
           ()
+      ; make_field ~name:"vtpm"
```
psafont, line 2434: The name is singular, but we're forced to print a list because I couldn't find a way to use an `option` in the datamodel.

ocaml/idl/datamodel_vtpm.ml
edwintorok: XAPI uses lists for too many things, would be nice if the idl generator could convert a `Map(String, String)` into a `string Map.Make(String).t`, but doing that on the vtpm branch would be too much churn. We can try to see how it'd look on master though.

ocaml/idl/datamodel_vtpm.ml
```diff
@@ -16,17 +16,47 @@ open Datamodel_types
 open Datamodel_common
 open Datamodel_roles
 
+let get_contents =
+  call ~name:"get_contents" ~in_product_since:"rel_next"
+    ~doc:"Obtain the contents of the TPM" ~secret:true
+    ~params:[(Ref _vtpm, "self", "The VTPM reference")]
+    ~result:(String, "The contents") ~hide_from_docs:true
```
edwintorok, line 23: We've got a Base64 type in the RPC, can we try using that, would it help? This is a binary string but we want to avoid storing binary strings (for varstored we store it as base64).

psafont: Yes, I think it can help, does the database do the encoding and decoding or should the functions themselves do it?

ocaml/xapi/xapi_vtpm.ml
```diff
@@ -0,0 +1,38 @@
...
+  let secret = Db.VTPM.get_contents ~__context ~self in
+  Db.Secret.destroy ~__context ~self:secret ;
+  Db.VTPM.destroy ~__context ~self
+
+let get_contents ~__context ~self =
+  let secret = Db.VTPM.get_contents ~__context ~self in
+  Db.Secret.get_value ~__context ~self:secret
+
+let set_contents ~__context ~self ~contents =
+  let secret = Xapi_secret.create ~__context ~value:contents ~other_config:[] in
```
edwintorok, line 37: Are we going to leak secrets here? I think we need to destroy the previous secret after calling set_contents below, otherwise we'll keep a growing list of secrets in the secrets table.

ocaml/xapi-cli-server/records.ml
```diff
@@ -4976,3 +4979,37 @@ let repository_record rpc session_id repository =
...
+      )
+  ; record= x
+  ; getref= (fun () -> !_ref)
+  ; fields=
+      [
+        make_field ~name:"uuid" ~get:(fun () -> (x ()).API.vTPM_uuid) ()
+      ; make_field ~name:"vm"
+          ~get:(fun () -> get_uuid_from_ref (x ()).API.vTPM_VM)
+          ()
+      ; make_field ~name:"profile"
```
edwintorok, line 5009: I think we need a way to change the profile as well. It is a bit like the `platform` field in the VM. Although I think we can only change it by resetting the vtpm, once the vtpm is in use I don't think that changing the profile is permitted. So maybe we need a way to specify the profile at `xe vtpm-create` time, is that possible?

ocaml/idl/datamodel_vtpm.ml
psafont: I don't think the variant support base64: https://github.com/xapi-project/xen-api/blob/master/ocaml/idl/datamodel_types.ml#L390
The base64 encoding is available in the RPC interface.

I found that the license uses base64encoding as well, but the conversion is forced upon the user. This could be an option if there are usecase that need to keep the vtpm contents base64-encoded

psafont: The contents are now encoded and decoded: both functions use exceptions to report errors, decoding shouldn't be a problem since we always do the encoding.

Regarding encoding I don't see how the code would throw an exception as I can't see the internal function returning an error.

This of course means the vtpm are always encoded/decoded. If this operation are common enough this might become a performance bottleneck. We can change this afterwards and do the encoding/deciding in the clients if we need more flexibility.

edwintorok: Ah we might need to do the base64 encoding in the clients anyway: you can't have binary data inside JSON/XML.

edwintorok: For now lets keep the base64 encoding here and see what happens with the client, if it cannot send binary data through the wire protocl we'll need to move the base64 encoding there

edwintorok (approved):

PR #4645, opened 2022-03-09
CA-364138 log when about to stop varstored and varstore-guard

lindig:
For debugging lost FDs in varstore-guard, log when we call it to stop serving
a domain.


edwintorok:
Can you also add some debug messages here? https://github.com/xapi-project/xen-api/blob/dfaf3839c9d05a9bff889a7d8de55fef8786d649/ocaml/xenopsd/lib/xenops_sandbox.ml#L148 maybe the problem is that they do get called, but the file doesn't exist anymore (because other parts of xenopsd have already cleaned up the chroot)

robhoes (approved):

PR #4646, opened 2022-03-10
CA-365112: Permit pool admin username with space to ssh login

liulinC:
CA-363207 complains pool admin group name with space cannot login,
that issue is fixed by surround group name with [], which works both
for pbis and winbind.
However, the fix does not work PBIS username with space.
This commit take pbis and winbind seperately
* for pbis, relace space with +, which is the original pbis solution
* for winbind, surround name with []

### DIFF
--- a/scripts/plugins/extauth-hook-AD.py
+++ b/scripts/plugins/extauth-hook-AD.py
@@ -172,20 +172,21 @@ class DynamicPam(ADConfig):
                 self._add_subject(subject_rec)
 
     def _format_item(self, item):
-        # CA-363207: Pam cannot handle space, put it inside []
         space_replacement = "+"
         if space_replacement in item:
             raise ValueError(
                 "{} is not permitted in subject name".format(space_replacement))
 
-        if " " in item:
-            if self._backend == ADBackend.bd_pbis:
-                # PBIS relace space with "+", eg "ab  cd" -> "ab++cd"
-                # PBIS pam module will reverse it back
-                return item.replace(" ", space_replacement)
-            # winbind put name with space within "[]"
-            return "[{}]".format(item)
-        return item
+        if " " not in item:
+            return item
+
+        if self._backend == ADBackend.bd_pbis:
+            # PBIS relace space with "+", eg "ab  cd" -> "ab++cd"
+            # PBIS pam module will reverse it back
+            return item.replace(" ", space_replacement)
+
+        # winbind put name with space within "[]"
+        return "[{}]".format(item)
 
     def _add_subject(self, subject_rec):
         try:
--- a/scripts/plugins/extauth-hook-AD.py
+++ b/scripts/plugins/extauth-hook-AD.py
@@ -27,16 +27,20 @@ from enum import Enum
 # - /etc/pam.d/hcp_users
 # - /etc/ssh/ssh_config
 
-#pylint: disable=super-with-arguments
+# pylint: disable=super-with-arguments
+
 
 def setup_logger():
     logger = logging.getLogger()
-    logging.basicConfig(format='%(asctime)s %(levelname)s %(name)s %(funcName)s %(message)s', level=logging.DEBUG)
+    logging.basicConfig(
+        format='%(asctime)s %(levelname)s %(name)s %(funcName)s %(message)s', level=logging.DEBUG)
     # Send to syslog local5, which will be redirected to xapi log /var/log/xensource.log
-    handler = logging.handlers.SysLogHandler(facility='local5', address='/dev/log')
+    handler = logging.handlers.SysLogHandler(
+        facility='local5', address='/dev/log')
     std_err = logging.StreamHandler(sys.stderr)
     # Send to authpriv, which will be redirected to /var/log/secure
-    auth_log = logging.handlers.SysLogHandler(facility="authpriv", address='/dev/log')
+    auth_log = logging.handlers.SysLogHandler(
+        facility="authpriv", address='/dev/log')
     logger.addHandler(handler)
     logger.addHandler(std_err)
     logger.addHandler(auth_log)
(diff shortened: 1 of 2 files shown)


scripts/plugins/extauth-hook-AD.py
```diff
@@ -164,6 +167,10 @@ def _apply_to_cache(self):
     def _format_item(self, item):
         # CA-363207: Pam cannot handle space, put it inside []
         if " " in item:
+            if self._backend == ADBackend.bd_pbis:
+                # PBIS relace space to +, eg "ab  cd" -> "ab++cd"
```
psafont, line 171:
```suggestion
# PBIS replaces spaces with +, eg "ab  cd" -> "ab++cd"
```

scripts/plugins/extauth-hook-AD.py
```diff
@@ -27,6 +27,7 @@
 # - /etc/pam.d/hcp_users
 # - /etc/ssh/ssh_config
 
+#pylint: disable=super-with-arguments
```
psafont, line 30: I'll be happy to review a PR that introduces pylint checks to the CI ;)

We should replace nosetests with pytest as well

liulinC: yes, there are several pylint warnings I should fix, 
but I like to leave it to anther commit to get your attention on the fix here.

scripts/plugins/extauth-hook-AD.py
```diff
@@ -174,15 +181,17 @@ def _add_subject(self, subject_rec):
...
             if not is_group:
                 subject_upn = self._format_item(subject_rec["other_config"]["subject-upn"])
-                self._lines.append("account sufficient pam_succeed_if.so user {} {}".format(condition, subject_upn))
-        except Exception:
-            logger.warning("Failed to check subject %s for dynamic pam", subject_rec)
+                self._lines.append("account sufficient pam_succeed_if.so user {} {}".format(
+                    condition, subject_upn))
+        #pylint: disable=broad-except
+        except Exception as exp:
+            logger.warning("Failed to ad subject %s for dynamic pam: %s", subject_rec, str(exp))
```
edwintorok, line 194: Failed to `ad` -> Failed to `add`

scripts/plugins/extauth-hook-AD.py
psafont: Something happened, I still see this typo

scripts/plugins/extauth-hook-AD.py
```diff
@@ -164,6 +167,10 @@ def _apply_to_cache(self):
     def _format_item(self, item):
         # CA-363207: Pam cannot handle space, put it inside []
         if " " in item:
```
lindig, line 169: I assume the substitution `item.replace(" " , "+")` can be run every time without checking for space first. It will simply do nothing if there is no space in `item`. Is `+` a legal character or are we creating a problem here by mapping space to plus and making this transformation not reversible?

scripts/plugins/test_extauth_hook_AD.py
```diff
@@ -103,7 +103,27 @@ def test_permit_admin_user(self, mock_rename, mock_chmod):
         user = build_user("CONNAPP", "radmin", True)
         mock_session.xenapi.subject.get_record.return_value = user
         permit_user = r"account sufficient pam_succeed_if.so user = CONNAPP\radmin"
-        dynamic = DynamicPam(mock_session, args_bd_wibind)
+        dynamic = DynamicPam(mock_session, args_bd_winbind)
```
lindig, line 106: You need a test case with the name containing a plus, if that is legal.

lindig (approved):

scripts/plugins/test_extauth_hook_AD.py
psafont: I suspect that names containing a plus sign are not valid, should xapi reject them?

scripts/plugins/extauth-hook-AD.py
liulinC: Maybe a problem of GitHub? The latest code has
`# PBIS relace space with "+", eg "ab cd" -> "ab++cd"`

scripts/plugins/extauth-hook-AD.py
liulinC: we need to check space for winbind anyway.
PBIS will reverse it back. (I can add a comment for it)

scripts/plugins/test_extauth_hook_AD.py
liulinC: yes, plus is not permitted in windows AD, this means it will never happens in username, (if user force it, then AD will replace plus with "_").
xapi can detect and reject it. But that may not be necessary as
1. the username are queried from AD, not input from user, in this case, plus never exists
2. In case of plus do exists, we just put it into the pam configure file, and it does not affect other users. This is similar to detect and skip this `bad` user.

psafont (approved):

scripts/plugins/extauth-hook-AD.py
```diff
@@ -168,35 +175,40 @@ def _format_item(self, item):
         # CA-363207: Pam cannot handle space, put it inside []
         space_replacement = "+"
         if space_replacement in item:
-            raise ValueError("{} is not permitted in subject name".format(space_replacement))
+            raise ValueError(
+                "{} is not permitted in subject name".format(space_replacement))
 
         if " " in item:
             if self._backend == ADBackend.bd_pbis:
```
lindig, line 182: This is a key part of the logic. I would like it if the backends were handled with more symmetry and the code stay flat:

```
if not (" " in item):
  return item

if self._backend == ADBackend.bd_pbis:
   return item.replace(" " , space_replacement)
else
   return "[{}]".format(item)
```

liulinC: just updated, please help to have a check

PR #4648, opened 2022-03-10
Client-cert auth: use CAfile

robhoes:
The CApath stunnel config option is currently used, but it turns out
that the necessary pem file symlinks are not always created. This should
be sorted out, but for now the quickest solution is to switch to CAfile
and use the CA bundle.


psafont (approved):
This is good, this makes all the code to "refresh" the trust root when using `c_rehash` moot. Although not ideal, we can do the cleanup at a later point

lindig (approved):

PR #4650, opened 2022-03-13
backport https://github.com/xapi-project/xen-api/pull/4646 to yangtze

liulinC:
This is backport of  https://github.com/xapi-project/xen-api/pull/4646, target to Yangtze

### DIFF
--- a/scripts/plugins/extauth-hook-AD.py
+++ b/scripts/plugins/extauth-hook-AD.py
@@ -172,20 +172,21 @@ class DynamicPam(ADConfig):
                 self._add_subject(subject_rec)
 
     def _format_item(self, item):
-        # CA-363207: Pam cannot handle space, put it inside []
         space_replacement = "+"
         if space_replacement in item:
             raise ValueError(
                 "{} is not permitted in subject name".format(space_replacement))
 
-        if " " in item:
-            if self._backend == ADBackend.bd_pbis:
-                # PBIS relace space with "+", eg "ab  cd" -> "ab++cd"
-                # PBIS pam module will reverse it back
-                return item.replace(" ", space_replacement)
-            # winbind put name with space within "[]"
-            return "[{}]".format(item)
-        return item
+        if " " not in item:
+            return item
+
+        if self._backend == ADBackend.bd_pbis:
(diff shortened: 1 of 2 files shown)


scripts/plugins/extauth-hook-AD.py
```diff
@@ -205,18 +244,19 @@ def is_special_line(self, line):
 
     def _load_values(self):
```
lindig, line 245: I understand this is a packport but I notice a deeply nested logic that I don't trust to be correct. There are several special cases (empty lines, kv != 2) that are mostly unmotivated. Given that this code handles part of authentication I would like to see more clarity.

lindig (approved):

psafont (changes requested):
Please add the hash of the original commits in the commit message of every commit, this makes it easier to see that the backport is accurate.

psafont:
The first commit is not a cherry-pick, it has additonal changes not present in the original commit. Is there a reason for this? If there are more commit backported please add each one as a cherry-pick.

The strategy with backport is to only import commits that are already on `master`, plus additional commits when some adaptations are needed to make it work on the branch

liulinC:
> Please add the hash of the original commits in the commit message of every commit, this makes it easier to see that the backport is accurate.

Just  updated, p

> The first commit is not a cherry-pick, it has additonal changes not present in the original commit. Is there a reason for this? If there are more commit backported please add each one as a cherry-pick.
> 
> The strategy with backport is to only import commits that are already on `master`, plus additional commits when some adaptations are needed to make it work on the branch

My autopep vim plugin updated it for me, I have just fixup, please help to take a look, thanks.

psafont (approved):
On the first commit a few more changes were added to match the final output of the patch.

Looks like a faithful backport to me.

liulinC:
> On the first commit a few more changes were added to match the final output of the patch.
> 
> Looks like a faithful backport to me.

The original commit do include a little pep8 fixup, see https://github.com/xapi-project/xen-api/pull/4646/commits/93dc7707fd4fa4de55a285ca9a179c3c1124ede0

That is caused by refine a little pep8 during the issue, then I put the remaining to a single commit.
I should squash the commits in master, but since it got merged, let's keep consistency here.

PR #4651, opened 2022-03-14
Remove RPM gpgcheck in reposync

minglumlu:
Prior to this commit, the assumption is the RPM GPG key will be imported
before running 'reposync'. This results in un-necessary dependency and
un-necessary RPM packages signature checking since this will be done by
'yum upgrade' later as well.

In this commit, the '--gpgcheck' paramter is removed. Thereafter,
'reposync' will only check repository metadata. It will not require the
RPM GPG key being imported anymore. And later in 'yum upgrade', the GPG
key will be imported automatically and used for RPM packages signature
checking.


lindig (approved):
This has security implications; I am not familiar enough with the architecture to be sure that the signatures will still be checked. The code change by itself is fine.

robhoes (approved):

PR #4653, opened 2022-03-21
CA-365130: Log the exception causing backups to fail

psafont:
Previously the type of exception was not reported.

Not print the exception with warning level and make an effort to report
whether the process was signaled to stop.


edwintorok:
Might be useful to format Subprocess_killed using https://erratique.ch/software/fmt/doc/Fmt/Dump/index.html#val-signal formatter (the Sys signal numbers and the POSIX ones do not match: the Sys one is just an index of the OCaml variant I think, although they're negative so they can be distinguished that way. E.g. Sys.sigkill is -7, not 9 as it would be on Unix).

robhoes (approved):

PR #4654, opened 2022-03-21
Remove hard-coded paths of repository metadata files

minglumlu:
This series remove the hard-coded paths of repository metadata files. 
Instead, in this series, the paths are retrieved from the repomd.xml.

### DIFF
--- a/ocaml/tests/test_repository_helpers.ml
+++ b/ocaml/tests/test_repository_helpers.ml
@@ -724,15 +724,15 @@ module UpdateInfoMetaDataOfXml = Generic.MakeStateless (struct
   module Io = struct
     type input_t = string
 
-    type output_t = (UpdateInfoMetaData.t, exn) result
+    type output_t = (RepoMetaData.t, exn) result
 
     let string_of_input_t x = x
 
     let fields =
       Fmt.Dump.
         [
-          field "checksum" (fun (r : UpdateInfoMetaData.t) -> r.checksum) string
-        ; field "location" (fun (r : UpdateInfoMetaData.t) -> r.location) string
+          field "checksum" (fun (r : RepoMetaData.t) -> r.checksum) string
+        ; field "location" (fun (r : RepoMetaData.t) -> r.location) string
         ]
       
 
@@ -741,7 +741,7 @@ module UpdateInfoMetaDataOfXml = Generic.MakeStateless (struct
   end
 
   let transform input =
-    try Ok (UpdateInfoMetaData.of_xml (Xml.parse_string input))
+    try Ok RepoMetaData.(of_xml (Xml.parse_string input) UpdateInfo)
     with e -> Error e
 
   let tests =
@@ -808,7 +808,7 @@ module UpdateInfoMetaDataOfXml = Generic.MakeStateless (struct
             </repomd>
            |}
         , Ok
-            UpdateInfoMetaData.
+            RepoMetaData.
               {checksum= "123abc"; location= "repodata/123abc.xml.gz"}
             
         )
--- a/ocaml/xapi/repository_helpers.ml
+++ b/ocaml/xapi/repository_helpers.ml
@@ -566,9 +566,13 @@ end
 module RepoMetaData = struct
   type t = {checksum: string; location: string}
 
-  type datatype = UpdateInfo
+  type datatype = UpdateInfo | Group
 
-  let string_of_datatype = function UpdateInfo -> "updateinfo"
+  let string_of_datatype = function
+    | UpdateInfo ->
+        "updateinfo"
+    | Group ->
+        "group"
 
   let assert_valid = function
     | {checksum= ""; _} | {location= ""; _} ->
--- a/ocaml/xapi/repository_helpers.ml
+++ b/ocaml/xapi/repository_helpers.ml
@@ -563,22 +563,28 @@ module Applicability = struct
         raise Invalid_inequality
 end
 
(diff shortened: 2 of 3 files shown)


ocaml/tests/test_repository_helpers.ml
```diff
@@ -741,7 +741,11 @@ module UpdateInfoMetaDataOfXml = Generic.MakeStateless (struct
   end
 
   let transform input =
-    try Ok (UpdateInfoMetaData.of_xml (Xml.parse_string input))
+    try
+      Ok
+        (UpdateInfoMetaData.of_xml (Xml.parse_string input) "updateinfo"
+        |> UpdateInfoMetaData.assert_valid
+        )
```
psafont, line 748: I think the validation should happens in the conversion function, as well as picking the field, simplifying this code into: 

```suggestion
      Ok (UpdateMetaData.of_xml (Xml.parse_string input))
```
This means `UpdateInfoMetadata.of_xml` could be defined as
```ocaml
  let of_xml xml =
    let update_info = Repometadata.of_xml xml "updateinfo" in
    UpdateInfoMetaData.assert_valid update_info ;
    update_info
```
This would make this function much easier to use and less error-prone

psafont, line 748: I think the validation should happens in the conversion function, as well as picking the field, simplifying this code into: 

```suggestion
      Ok (UpdateMetaData.of_xml (Xml.parse_string input))
```
This means `UpdateInfoMetadata.of_xml` could be defined as
```ocaml
  let of_xml xml =
    let update_info = Repometadata.of_xml xml "updateinfo" in
    UpdateInfoMetaData.assert_valid update_info ;
    update_info
```
This would make this function much easier to use and less error-prone

ocaml/xapi/repository_helpers.ml
```diff
@@ -628,6 +622,22 @@ module UpdateInfoMetaData = struct
     )
 end
 
+module UpdateInfoMetaData = struct
+  include RepoMetaData
```
psafont, line 626: I don't think exposing all the functions in RepoMetadata is to our convenience here, I'd rather be explicit about the functions define in the module.

ocaml/xapi/repository_helpers.ml
```diff
@@ -638,6 +638,19 @@ module UpdateInfoMetaData = struct
     "updateinfo" |> of_xml_file xml_path |> assert_valid
 end
 
+module GroupMetaData = struct
+  include RepoMetaData
```
psafont, line 642: Include adds the fields and methods of `RepoMetaData` into the signature of the module. I don't think this is what should happen, an `open` here is more that enough, or the fully qualified name for fields and functions can be used as well

ocaml/tests/test_repository_helpers.ml
minglumlu: The `of_xml` will be called in `of_xml_file`. So are you suggesting the `of_xml_file` accepts a function as a parameter? It would look like:
```
module RepoMetaData = struct
  type t = {checksum: string; location: string}

  let default = {checksum= ""; location= ""}

  let of_xml xml data_type =
    ...

  let of_xml_file xml_path of_xml =
    match Sys.file_exists xml_path with
    | false ->
        error "No repomd.xml found: %s" xml_path ;
        raise Api_errors.(Server_error (invalid_repomd_xml, []))
    | true -> (
      match Xml.parse_file xml_path with
      | xml ->
          of_xml xml
      | exception e ->
          error "Failed to parse repomd.xml: %s" (ExnHelper.string_of_exn e) ;
          raise Api_errors.(Server_error (invalid_repomd_xml, []))
    )
end

module UpdateInfoMetaData = struct
  let assert_valid = function
    | RepoMetaData.{checksum= ""; _} | RepoMetaData.{location= ""; _} ->
        error
          "Can't find valid 'checksum' or 'location' in updateinfo in \
           repomod.xml" ;
        raise Api_errors.(Server_error (invalid_repomd_xml, []))
    | md ->
        md

  let of_xml xml =
    RepoMetadata.of_xml "updateinfo" xml |> assert_valid

  let of_xml_file xml_path =
    RepoMetaData.of_xml_file xml_path of_xml
end
```

psafont: I don't think that change is necessary, similar to how I suggested to implement `UpdateInfoMetaData.of_xml`, `of_xml_path` can be implemented as

```ocaml
  let of_xml_file xml_path =
    let update_info = Repometadata.of_xml_file xml_path "updateinfo" in
    UpdateInfoMetaData.assert_valid update_info ;
    update_info
```

minglumlu:
Hi Reviewers,
Thanks for @psafont 's suggestions. I reworked the commits based on them. Since changes are not suitable as fixups, I have to push forcibly. Thanks for your reviews again.

psafont (approved):
The code looks reasonable to me.

I think there's an opportunity to read the repomd.xml only once and extract the information as many times as it's needed. Currently it's read twice

robhoes (approved):

minglumlu:
> I think there's an opportunity to read the repomd.xml only once and extract the information as many times as it's needed. Currently it's read twice

Indeed. I thought it would be easier to handle output of single record. But it may be not too difficult as well if there is a `RepoMetaData.of_xml_file` returns an association list: [(UpdateInfo, {...}); (Group, {...})]. And the logic outside parses the assoc list.

PR #4655, opened 2022-03-21
CA-365121: pool join: require common xapi versions

robhoes:
In the pre-join checks, require that the xapi version of the joining
host is equal to that of the pool coordinator. This is stricter than,
and replaces, the check for common platform versions. In particular,
this protects against joining a host with a newer xapi version to a
pool, which may break (the coordinator must at any time have the highest
version).


edwintorok:
How will you ensure that the joining host has the same XAPI version?
Lets say you've got a pool with N hosts and you've applied some XAPI updates so they're at version X+10.
Now you install a new host from the latest snapshot ISO but that gets an older XAPI, lets say version X.
To update the XAPI version on the new host you've got these options:
* join it to an existing pool, and then use the connector from that pool to install updates on the new host
* set up a new connector, or configure the existing connector to also have the credentials for the newly joining host, update its XAPI version (to latest there is no way to ensure to update exactly to the version the pool master has, so this will also require you to immediately update your entire pool to latest of everything in order to join a new host).

This can be quite problematic if all you wanted to do is to provide some quick, extra capacity to your pool.

edwintorok:
I think there should at least be a xapi.conf entry to override the strict check and allow a >= check instead (coordinator >= joining), otherwise you may end up stuck in a situation without a way out of (short of reinstalling your entire pool to get them onto consistent versions)

edwintorok (approved):

psafont (approved):

PR #4656, opened 2022-03-22
CA-355588: Users in groups contains # can not ssh into dom0

liulinC:
Linux PAM take # as comments and there is no way to escape it.
This commit permit access to dom0 by pam_access instead of pam_succeed_if
/etc/security/hcp_access.conf (pam_access config) is used to permit users
and groups instead of /etc/pam.d/hcp_users (linux pam config)

### DIFF
--- a/scripts/plugins/extauth-hook-AD.py
+++ b/scripts/plugins/extauth-hook-AD.py
@@ -104,7 +104,27 @@ class ADConfig(object):
             os.chmod(self._file_path, self._file_mode)
 
 
-class StaticPam(ADConfig):
+# pylint:disable=too-few-public-methods
+class PamConfig(ADConfig):
+    """
+    Base class for Pam configuration
+    """
+    ACCESS_CONFIG = "/etc/security/hcp_access.conf"
+    @abc.abstractmethod
+    def _apply_to_cache(self):
+        pass
+
+    def __init__(self, path, session, args, ad_enabled=True):
+        super(PamConfig, self).__init__(path, session, args, ad_enabled,
+                                        load_existing=False)
+        self.list_sep = args.get("list_sep", ";")
+        self.field_sep = args.get("field_sep", ",")
+
+
+class StaticPam(PamConfig):
+    """
+    Class to manage ssh pam configuration
+    """
     ad_pam_format = """#%PAM-1.0
 auth        required      pam_env.so
 auth        sufficient    pam_unix.so try_first_pass nullok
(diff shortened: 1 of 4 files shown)


scripts/plugins/extauth-hook-AD.py
```diff
@@ -135,77 +138,88 @@ class StaticPam(ADConfig):
...
+                    "{} is not permitted in subject name".format(space_replacement))
+
             # PBIS relace space with "+", eg "ab  cd" -> "ab++cd"
             # PBIS pam module will reverse it back
-            return item.replace(" ", space_replacement)
-
-        # winbind put name with space within "[]"
-        return "[{}]".format(item)
+            item = item.replace(" ", space_replacement)
+        if is_group:  # put group into () to avlid conflict with user
```
edwintorok, line 208: typo: avoid

scripts/plugins/test_extauth_hook_AD.py
```diff
@@ -77,11 +80,10 @@ def build_group(domain, name, is_admin):
 @patch("os.rename")
 class TestStaicPamConfig(TestCase):
     def test_ad_not_enabled(self, mock_rename, mock_chmod):
-        # No hcp_users file should be included
         static = StaticPam(mock_session, args_bd_winbind, ad_enabled=False)
         static.apply()
-        enabled_keyward = "account     include       hcp_users"
-        self.assertFalse(line_exists_in_config(static._lines, enabled_keyward))
+        enabled_keyward = "hcp_access.conf"
+        self.assertFalse(enabled_keyward in "".join(static._lines))
```
edwintorok, line 86: keyward/keyword

scripts/plugins/extauth-hook-AD.py
```diff
@@ -135,77 +138,88 @@ class StaticPam(ADConfig):
...
+            logger.exception("Failed permit subjects")
+        finally:
+            # deny all others by default
+            conf = self.CONF_TEMP.format(
+                perm="-", sep=self.field_sep, name="ALL", origins="ALL")
+            self._lines.append(conf)
+            self._lines.append("")  # Require ending line sep
+
+    def _format_item(self, item, is_group):
         space_replacement = "+"
```
edwintorok, line 199: It would be useful to check that the user/group name doesn't contain any of the field separators. Even if currently AD would forbid those characters:
* the separator is configurable in xapi.conf, so someone could accidentally choose an unsafe one
* future versions of AD may change what characters they allow, and if someone manages to create an AD user with one of the escape chars in their user/group name they'd be able to elevate privileges.

If the user/group contains the field separator then see whether it is possible to escape it, if not then we should reject adding that user/group to XAPI and for an already existing user/group we should log+skip the entry here.

liulinC: @edwintorok  we forbid add separators to avoid injection attack, 
However, we will not forbid add subject with separators, given
- if Microsoft allow it, then we allow it.
- we forbid add to pam configuration, which disable ssh, However, it should not affect other functions like  add users, user login with SDK, CHC.   
We log warning to remind user update separators if they insist to.

liulinC:
close this PR, will raise another PR with pam_filelist

PR #4657, opened 2022-03-22
Avoid compiler warnings

psafont:
The PR is better reviewed commit by commit. The warnings removed are mostly unlabeled parameters in Client calls and unused bindings.

For autogenerated code I've opted for ignoring the warnings emitted rather than trying to change the code generated.

The quality gate doesn't change much because I've held off some changes that happen before the ones in the compiled units fixed here. I believe the other changes are more contentious and I want to test more rigorously.

The changes in here should not be contentious as
1. They add parameter labels, which makes it difficult to introduce errors; and
2. Changes in tests surface errors really fast in obvious ways

### DIFF
--- a/ocaml/message-switch/switch/logging.ml
+++ b/ocaml/message-switch/switch/logging.ml
@@ -71,21 +71,15 @@ let ignore_fmt_lwt fmt = Printf.ksprintf (fun _ -> Lwt.return ()) fmt
 (* General system logging *)
 let logger = create 512
 
-type level = Debug | Info | Warn | Error | Null
-
-let log_level = ref Warn
+type level = Info | Warn | Error
 
 let string_of_level = function
-  | Debug ->
-      "debug"
   | Info ->
       "info"
   | Warn ->
       "warn"
   | Error ->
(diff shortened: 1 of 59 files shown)


ocaml/message-switch/switch/logging.ml
```diff
@@ -71,21 +71,15 @@ let ignore_fmt_lwt fmt = Printf.ksprintf (fun _ -> Lwt.return ()) fmt
 (* General system logging *)
 let logger = create 512
 
-type level = Debug | Info | Warn | Error | Null
-
-let log_level = ref Warn
+type level = Info | Warn | Error
```
lindig, line 74: It looks like the only purpose is to turn these into strings later. I would assume that function like `info` could pass the correct string directly without using a `level` value which is turned into a string. The one thing this has going for it, it prevents arbitrary strings to be used.

lindig (approved):
The pure number of (mostly syntactical) changes make this challenging to review; but they look correct to me.

robhoes (approved):

PR #4658, opened 2022-03-24
users in pool admin group which contains # can not ssh into dom0

liulinC:
CA-355588: users in pool admin group which contains # can not ssh into dom0
    
    Linux PAM take # as comments and there is no way to escape it
    This commit fix the issue by permit users and groups through
    pam_listfile pam module by
    - users, in /etc/security/hcp_ad_users.conf
    - groups, in /etc/security/hcp_ad_groups.conf

### DIFF
--- a/scripts/plugins/extauth-hook-AD.py
+++ b/scripts/plugins/extauth-hook-AD.py
@@ -10,8 +10,6 @@
 # Alternatively, the extauth-hook module can be called, which will
 # dispatch to the correct extauth-hook-<type>.py module automatically.
 import abc
-import XenAPIPlugin
-import XenAPI
 import sys
 import subprocess
 import os
@@ -20,6 +18,9 @@ import logging
 import logging.handlers
 from collections import OrderedDict
 from enum import Enum
+import XenAPIPlugin
+import XenAPI
+
 
 # this plugin manage following configuration files for external auth
 # - /etc/nsswitch.conf
@@ -31,7 +32,8 @@ from enum import Enum
 
 
 def setup_logger():
-    logger = logging.getLogger()
+    """Helper function setup logger"""
+    log = logging.getLogger()
     logging.basicConfig(
         format='%(asctime)s %(levelname)s %(name)s %(funcName)s %(message)s', level=logging.DEBUG)
     # Send to syslog local5, which will be redirected to xapi log /var/log/xensource.log
@@ -41,9 +43,9 @@ def setup_logger():
     # Send to authpriv, which will be redirected to /var/log/secure
     auth_log = logging.handlers.SysLogHandler(
         facility="authpriv", address='/dev/log')
-    logger.addHandler(handler)
-    logger.addHandler(std_err)
-    logger.addHandler(auth_log)
+    log.addHandler(handler)
+    log.addHandler(std_err)
+    log.addHandler(auth_log)
 
 
 setup_logger()
@@ -51,60 +53,73 @@ logger = logging.getLogger(__name__)
 
 
 def run_cmd(cmd, log_cmd=True):
+    """Helper function to run command"""
     try:
         result = subprocess.check_output(cmd)
         if log_cmd:
-            logger.debug("{} -> {}".format(cmd, result))
+            msg = "{} -> {}".format(cmd, result)
+            logger.debug(msg)
         return result.strip()
-    except Exception:
+    except Exception:  # pylint: disable=broad-except
         logger.exception("Failed to run command %s", cmd)
         return None
 
 
 class ADBackend(Enum):
-    bd_pbis = 0
-    bd_winbind = 1
+    """Enum for AD backend"""
+    BD_PBIS = 0
+    BD_WINBIND = 1
 
 
+# pylint: disable=useless-object-inheritance, too-few-public-methods
 class ADConfig(object):
+    """Base class for AD configuration"""
+    #pylint: disable=too-many-arguments
+
     def __init__(self, path, session, args, ad_enabled=True, load_existing=True, file_mode=0o644):
         self._file_path = path
         self._session = session
         self._args = args
         self._lines = []
-        self._backend = self._get_ad_backend(args)
+        self._backend = self._get_ad_backend()
         self._ad_enabled = ad_enabled
         self._file_mode = file_mode
         if load_existing and os.path.exists(self._file_path):
-            with open(self._file_path, 'r') as f:
-                lines = f.readlines()
+            with open(self._file_path, 'r') as file:
+                lines = file.readlines()
                 self._lines = [l.strip() for l in lines]
 
-    def _get_ad_backend(self, args):
-        if args.get("ad_backend", "winbind") == "pbis":
+    def _get_ad_backend(self):
+        """Get atctive AD backend"""
+        if self._args.get("ad_backend", "winbind") == "pbis":
             logger.debug("pbis is used as AD backend")
-            return ADBackend.bd_pbis
+            return ADBackend.BD_PBIS
 
         logger.debug("winbind is used as AD backend")
-        return ADBackend.bd_winbind
+        return ADBackend.BD_WINBIND
 
     @abc.abstractmethod
     def _apply_to_cache(self):
         pass
 
     def apply(self):
+        """Apply configuration"""
         self._apply_to_cache()
         self._install()
 
     def _install(self):
-        with tempfile.NamedTemporaryFile(prefix="extauth-", delete=False) as f:
-            f.write("\n".join(self._lines).encode())
-            f.flush()
-            os.rename(f.name, self._file_path)
+        """Install configuration"""
+        with tempfile.NamedTemporaryFile(prefix="extauth-", delete=False) as file:
+            file.write("\n".join(self._lines).encode())
+            file.flush()
+            os.rename(file.name, self._file_path)
             os.chmod(self._file_path, self._file_mode)
 
 
-class StaticPam(ADConfig):
+class StaticSSHPam(ADConfig):
+    """
+    Class to manage ssh pam configuration
+    """
     ad_pam_format = """#%PAM-1.0
 auth        required      pam_env.so
 auth        sufficient    pam_unix.so try_first_pass nullok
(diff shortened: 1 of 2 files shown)


liulinC:
will raise another PR to xenserver-status-report to collect /etc/security/users.conf (groups.conf) after this merge.

scripts/plugins/extauth-hook-AD.py
```diff
@@ -41,70 +47,83 @@ def setup_logger():
...
-            with open(self._file_path, 'r') as f:
-                lines = f.readlines()
+            with open(self._file_path, 'r') as file:
+                lines = file.readlines()
                 self._lines = [l.strip() for l in lines]
 
-    def _get_ad_backend(self, args):
-        if args.get("ad_backend", "winbind") == "pbis":
+    def _get_ad_backend(self):
+        """Get atctive AD backend"""
```
lindig, line 97: typo: `active`

scripts/plugins/extauth-hook-AD.py
```diff
@@ -41,70 +47,83 @@ def setup_logger():
...
+            msg = "{} -> {}".format(cmd, result)
+            logger.debug(msg)
         return result.strip()
-    except Exception:
+    except Exception:  # pylint: disable=broad-except
         logger.exception("Failed to run command %s", cmd)
         return None
 
 
 class ADBackend(Enum):
```
lindig, line 72: This does not look like an OO design to me. Should not each backend be modeled as a class that knows the constants it needs to know?

lindig (commented):
I'm not a Python developer but my main question would be: if we are using classes, why are the different backends not modeled as classes, inheriting from a super class that captures the common behavior? Instead we have branching inside methods based on the backend.

scripts/plugins/extauth-hook-AD.py
liulinC: > I'm not a Python developer but my main question would be: if we are using classes, why are the different backends not modeled as classes, inheriting from a super class that captures the common behavior? Instead we have branching inside methods based on the backend.

Well, it can be, but I do not find it necessary as it will lead to double number of classes.

scripts/plugins/extauth-hook-AD.py
```diff
@@ -233,6 +223,55 @@ def _install(self):
...
+    def _add_subject(self, subject_rec):
+        try:
+            sid = subject_rec['subject_identifier']
+            name = subject_rec["other_config"]["subject-name"]
+            formatted_name = self._format_item(name)
+            logger.debug("Permit user %s with sid %s", formatted_name, sid)
+            self._lines.append(formatted_name)
+            # If ssh key is permittd in authorized_keys,
+            # The original name is compared, add UPN and original name
+            self._lines.append(subject_rec["other_config"]["subject-upn"])
```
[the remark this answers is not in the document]
liulinC: Does the `half-written`  means `formatted_name` is written, but original name and UPN does not write?
That actually does not matter.   And ```upn=subject_rec["other_config"]["subject-upn"] self._lines.append(upn)``` does not help on this.  But it is a good point that in case of subject does not have `subject-upn`, we should put the original name first.

scripts/plugins/extauth-hook-AD.py
```diff
@@ -233,6 +223,55 @@ def _install(self):
...
+            sid = subject_rec['subject_identifier']
+            name = subject_rec["other_config"]["subject-name"]
+            formatted_name = self._format_item(name)
+            logger.debug("Permit user %s with sid %s", formatted_name, sid)
+            self._lines.append(formatted_name)
+            # If ssh key is permittd in authorized_keys,
+            # The original name is compared, add UPN and original name
+            if self._backend == ADBackend.BD_PBIS and name != formatted_name:
+                self._lines.append(name)
+            self._lines.append(subject_rec["other_config"]["subject-upn"])
```
psafont, line 247: When this fails logs will show "Failed to add user", I think that's bound to cause confusion when triaging issues around this code, I'd rather have a different logline to be shown, if at all.

This could be done with:
```python
            if "subject-upn" in subject_rec["other_config"]:
                self._lines.append(subject_rec["other_config"]["subject-upn"])
            else:
                logger.info("UPN not found for user %s", subject_rec)
```

liulinC: updated, could you please help to check?

psafont (approved):

scripts/plugins/extauth-hook-AD.py
MarkSymsCtx: I'd be inclined to agree with @lindig, this should be a set of three classes

- A common base with everything that is shared
- A legacy PBIS implementation
- A Winbind implementation
otherwise there's a ton of technical debt relating to PBIS and excising it in the future will be much harder than just deleting the class definition

liulinC: Updated, could you please help to take a look. thanks.

scripts/plugins/extauth-hook-AD.py
```diff
@@ -163,74 +192,165 @@ def _apply_to_cache(self):
...
-        if space_replacement in item:
-            raise ValueError(
-                "{} is not permitted in subject name".format(space_replacement))
+    def _is_pool_admin(self, subject_rec):
+        return self.admin_role in subject_rec['roles']
 
-        if " " not in item:
-            return item
+    def _authorized_keys_support(self, subject_rec, formatted_name, name):
+        """Some subject can do extra configure to support authorized keys"""
```
psafont, line 207: The body of this function is empty!

psafont:
Would it be possible to have 3 classes only? An abstract one: `AD_backend`, and two implementations: `AD_WinBind` and `AD_PBIS`. I'm not convinved of the necessity of having different classes for dealing with SSH, PAM, NSS group lists and user lists.

As it is the latest change with many classes introduces too much churn to my taste to be able to confidently review it

liulinC:
@psafont @MarkSymsCtx @lindig I have just revert the latest commit which explosion the classes.

scripts/plugins/extauth-hook-AD.py
```diff
@@ -233,6 +223,69 @@ def _install(self):
...
 
+class UsersList(DynamicPam):
+    #pylint: disable=too-few-public-methods
+    """Class manage users which permit pool admin ssh"""
+
+    def __init__(self, session, arg, ad_enabled=True):
+        super(UsersList, self).__init__(HCP_USERS, session, arg, ad_enabled)
+
+    def _match_subject(self, subject_rec):
+        return subject_rec["other_config"]["subject-is-group"] != "true"
```
edwintorok, line 234: What if `subject-is-group` is missing from `other-config`?

scripts/plugins/extauth-hook-AD.py
```diff
@@ -233,6 +223,69 @@ def _install(self):
...
+
+class GroupsList(DynamicPam):
+    #pylint: disable=too-few-public-methods
+    """Class manage groups which permit pool admin ssh"""
+
+    def __init__(self, session, arg, ad_enabled=True):
+        super(GroupsList, self).__init__(HCP_GROUPS, session, arg, ad_enabled)
+
+    def _match_subject(self, subject_rec):
+        return subject_rec["other_config"]["subject-is-group"] == "true"
```
edwintorok, line 275: as above, what if this key is missing from other-config? it'll raise exception and crash, would be better to check that 'subject-is-group' is in other-config, and if not then return false here. Also is `other_config` spelled correctly here, shouldn't it be `other-config`?

scripts/plugins/extauth-hook-AD.py
edwintorok: Also why `other_config` instead of `other-config`, are you sure it is with `_`? Where is this set? Might be useful to use a string constant for other_config to ensure consistent spelling.

scripts/plugins/extauth-hook-AD.py
```diff
@@ -233,6 +223,69 @@ def _install(self):
...
+            self._lines.append("{}{}{}".format(user, sep, domain))
+        except KeyError:
+            logger.info("subject does not have upn %s", subject_rec)
+        except ValueError:
+            logger.info("UPN format is not right %s", upn)
+
+    def _add_subject(self, subject_rec):
+        try:
+            sid = subject_rec['subject_identifier']
+            name = subject_rec["other_config"]["subject-name"]
```
edwintorok, line 253: What if the subject name contains an @ character, would that confuse things and allow someone else to log in?
E.g. if john.smith@good.example.com has its subject name set to 'John@bad.example.com' then that will allow the user with `John@bad.example.com` UPN to log in (in addition to john.smith@good.example.com who can login with its UPN form)

scripts/plugins/extauth-hook-AD.py
```diff
@@ -233,6 +223,69 @@ def _install(self):
...
+
+    def _match_subject(self, subject_rec):
+        return subject_rec["other_config"]["subject-is-group"] == "true"
+
+    def _add_subject(self, subject_rec):
+        try:
+            sid = subject_rec['subject_identifier']
+            name = self._format_item(
+                subject_rec["other_config"]["subject-name"])
+            logger.debug("Permit group %s with sid %s", name, sid)
```
edwintorok, line 282: the SID is only used for logging but not for login (i.e. if the SID changes the group will still be allowed to log in). If that is intended then maybe make it explicit in logging, e.g. 'Permit group %s, current sid is %s'.

scripts/plugins/extauth-hook-AD.py
```diff
@@ -233,6 +223,69 @@ def _install(self):
...
+            logger.info("subject does not have upn %s", subject_rec)
+        except ValueError:
+            logger.info("UPN format is not right %s", upn)
+
+    def _add_subject(self, subject_rec):
+        try:
+            sid = subject_rec['subject_identifier']
+            name = subject_rec["other_config"]["subject-name"]
+            formatted_name = self._format_item(name)
+            logger.debug("Permit user %s with sid %s", formatted_name, sid)
```
edwintorok, line 255: The sid is not enforced for login, might be useful to make that explicit in the log message (see below).

scripts/plugins/extauth-hook-AD.py
liulinC: The subject name is in the format of `Domain_netbis\username` and UPN is information of `username@domain_name`, so there is no such problem.

scripts/plugins/extauth-hook-AD.py
liulinC: yes, Sid is only for logging.

scripts/plugins/extauth-hook-AD.py
liulinC: `subject-is-group` comes from xapi database, and it will always have a value, 
As it always better to do security check, I will add a try catch block for this.

edwintorok (approved):

PR #4660, opened 2022-03-24
CA-365516: CLI: protect cmdtable population with mutex

robhoes:


robhoes:
Hmm... where is the rest of CI tests?

psafont (approved):

ocaml/xapi-cli-server/cli_frontend.ml
```diff
@@ -3576,15 +3576,19 @@ let cmdtable : (string, cmd_spec) Hashtbl.t = Hashtbl.create 50
...
-  if !populated then
-    ()
-  else (
-    populated := true ;
-    List.iter
-      (fun (n, c) -> Hashtbl.add cmdtable n c)
-      (cmdtable_data @ Cli_operations.gen_cmds rpc session_id)
-    (* Autogenerated commands too *)
+  Xapi_stdext_threads.Threadext.Mutex.execute populated_m (fun () ->
+      if !populated then
```
lindig, line 3583: Could be simplified(?) to
```
if not !populated then ( 

)
```
without `else` unless you want to log something in the else branch. Also: would it be safer to use `Hashtnl.replace` rather than `add` because `add` shadows existing bindings that would re-appear when a key is removed.

lindig (approved):

ocaml/xapi-cli-server/cli_frontend.ml
robhoes: True for the if-statement, yes. For the other point, this table is filled in once and then left alone, so nothing is ever removed.

PR #4664, opened 2022-03-31
CA-365900: Clean up remanent stunnel client proxy

minglumlu:
When a member host is being updated, Toolstack creates a stunnel client
proxy for the connections from local YUM client to the YUM repositories
which locate on the coordinator host. The client stunnel process will be
terminated once the updating on the member host completes.

The client stunnel on a member host, however, will be left there when
the XAPI process has been terminated unexpectedly. And then when the
XAPI process starts again, and tries to create a stunnel client for
update, it will fail as there is already running one.

This commit aims to fix this issue by attempting to clean up running
client stunnel which is listening on the specific TCP port before
creating a new stunnel client proxy.


robhoes (approved):

psafont (approved):
I wish there was a more reasonable approach to handle these dangling processes, but I can't think of any without heavy rearchitecturing.

lindig (approved):

PR #4667, opened 2022-04-04
CP-38554: Start up and stop SWTPM and guard for every domain

psafont:
Xenopsd manages several services that sustain the lifecycle of each
domain. To launch a service, xenopsd usually starts up a daemon and it
waits for this daemon to write its pid on xenstore to signal the service
is ready to attend requests.

While this method is convenient in order to be able to cancel the wait
using xenstore watches, it's unfortunate as it makes the daemons depend
on xenstore even if it isn't needed to supply the service.

This change implements an alternative way to set up a cancellable wait
on the service readiness using inotify and epoll by setting up
filesystem watches instead of xenstore ones.

The polling loop wakes up every second to have a direct method to deal
with timeouts as well as unrelated events that happened in the
filesystem.

For every domain now xenopsd creates a chroot (which starts up a guard
daemon instance), launches an SWTPM instance, then waits on its pidfile
to be created.
On domain destruction is shuts down both instances: first SWTPM, then
the guard.

TODO:
- [x] Communicate swtpm socket path from inside the chroot to the qemu invocation

### DIFF
--- a/ocaml/xenopsd/lib/xenops_sandbox.ml
+++ b/ocaml/xenopsd/lib/xenops_sandbox.ml
@@ -118,7 +118,7 @@ end = struct
 end
 
 module type SANDBOX = sig
-  val prepare : domid:int -> vm_uuid:string -> Chroot.Path.t -> string
+  val create : domid:int -> vm_uuid:string -> Chroot.Path.t -> string
 
   val start :
        string
--- a/ocaml/xenopsd/lib/xenops_sandbox.ml
+++ b/ocaml/xenopsd/lib/xenops_sandbox.ml
@@ -21,31 +21,38 @@ module Chroot : sig
   module Path : sig
     type t
 
+    val root : t
+
     val of_string : relative:string -> t
+
+    val concat : t -> string -> t
   end
 
   val absolute_path_outside : t -> Path.t -> string
-  (** [absolute_path_outside chroot path] returns the absolute path outside the
-      chroot *)
 
   val chroot_path_inside : Path.t -> string
-  (** [chroot_path_inside path] returns the path when inside the chroot *)
 
-  val of_domid : daemon:string -> domid:int -> vm_uuid:string -> t
-  (** [of_domid daemon domid] describes a chroot for specified daemon and domain *)
+  val create_dir : within:t -> int -> Path.t -> unit
 
-  val create : daemon:string -> domid:int -> vm_uuid:string -> Path.t list -> t
-  (** [create daemon domid paths] Creates the specified chroot with appropriate
-      permissions, and ensures that all [paths] are owned by the chrooted daemon
-      and rw- *)
+  val of_domid :
+    base:string -> daemon:string -> domid:int -> vm_uuid:string -> t
+
+  val create :
+       base:string
+    -> daemon:string
+    -> domid:int
+    -> vm_uuid:string
+    -> Path.t list
+    -> t
 
   val destroy : t -> unit
-  (** [destroy chroot] Deletes the chroot *)
 end = struct
   type t = {root: string; uid: int; gid: int}
 
   module Path = struct
     type t = string
+
+    let root = "/"
 
     let of_string ~relative =
       if not (Filename.is_implicit relative) then
@@ -54,30 +61,41 @@ end = struct
              relative __LOC__
           ) ;
       relative
+
+    let concat = Filename.concat
   end
 
   let absolute_path_outside chroot path = Filename.concat chroot.root path
 
-  let chroot_path_inside path = Filename.concat "/" path
+  let chroot_path_inside path = Path.(concat root path)
+
+  let create_dir ~within perm path =
+    let fullpath = absolute_path_outside within path in
+    Xenops_utils.Unixext.with_file fullpath [Unix.O_CREAT; Unix.O_EXCL] perm
+      (fun fd -> Unix.fchown fd within.uid within.gid
+    )
 
   let qemu_base_uid () = (Unix.getpwnam "qemu_base").Unix.pw_uid
 
   let qemu_base_gid () = (Unix.getpwnam "qemu_base").Unix.pw_gid
 
-  let of_domid ~daemon ~domid ~vm_uuid =
+  let of_domid ~base ~daemon ~domid ~vm_uuid =
     let root =
-      if domid = 0 then
-        Printf.sprintf "/var/run/xen/%s-root-%d-%s" daemon domid vm_uuid
-      else
-        Printf.sprintf "/var/run/xen/%s-root-%d" daemon domid
+      let dir =
+        if domid = 0 then
+          Printf.sprintf "%s-root-%d-%s" daemon domid vm_uuid
+        else
+          Printf.sprintf "%s-root-%d" daemon domid
+      in
+      Filename.concat base dir
     in
     (* per VM uid/gid as for QEMU *)
     let uid = qemu_base_uid () + domid in
     let gid = qemu_base_gid () + domid in
     {root; uid; gid}
 
-  let create ~daemon ~domid ~vm_uuid paths =
-    let chroot = of_domid ~daemon ~domid ~vm_uuid in
+  let create ~base ~daemon ~domid ~vm_uuid paths =
+    let chroot = of_domid ~base ~daemon ~domid ~vm_uuid in
     try
       Xenops_utils.Unixext.mkdir_rec chroot.root 0o755 ;
       (* we want parent dir to be 0o755 and this dir 0o750 *)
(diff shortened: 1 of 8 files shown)


lindig:
Could this (daemons per domain) be also controlled by systemd?

ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -4,12 +4,12 @@ let prototyped_of_class = function
 
 let prototyped_of_field = function
   | "Repository", "gpgkey_path" ->
-      Some "22.10.0-next"
+      Some "22.12.0"
```
lindig, line 7: Maye this should not be in a PR and we deal with this once it is merged?

ocaml/xenopsd/lib/xenops_sandbox.ml
```diff
@@ -104,59 +104,104 @@ end = struct
     )
 end
 
-module Varstore_guard = struct
-  let daemon = "varstored"
+module type Chrooter = sig
+  val prepare : domid:int -> vm_uuid:string -> Chroot.Path.t -> string
```
lindig, line 108: I would like `create` better based on the documentation.

ocaml/xenopsd/lib/xenops_sandbox.ml
```diff
@@ -104,59 +104,104 @@ end = struct
...
+  val read : domid:int -> Chroot.Path.t -> vm_uuid:string -> string
+
+  val stop : string -> domid:int -> vm_uuid:string -> unit
+  (** [strop dbg ~domid ~vm_uuid] *)
+end
+
+module type Guard = sig
+  val daemon_name : string
+
+  val create : string -> vm_uuid:Uuidm.t -> domid:int -> path:string -> unit
```
lindig, line 132: Based on the type, `vm` would be good enough as a label. But this might be less convenient on the call side where `vm` could be used for something else.

ocaml/xenopsd/lib/xenops_sandbox.ml
```diff
@@ -104,59 +104,104 @@ end = struct
...
+      the guard to create a socket restricted to [vm_uuid]. Also creates
+      empty files specified in [path] owned by [domid] user. *)
+
+  val read : domid:int -> Chroot.Path.t -> vm_uuid:string -> string
+
+  val stop : string -> domid:int -> vm_uuid:string -> unit
+  (** [strop dbg ~domid ~vm_uuid] *)
+end
+
+module type Guard = sig
```
lindig, line 129: You could use `GUARD` which is quite conventional for module types. This leaves `Guard` for the functor below.

ocaml/xenopsd/lib/xenops_sandbox.ml
```diff
@@ -104,59 +104,104 @@ end = struct
...
       let gid = chroot.Chroot.gid in
       let absolute_socket_path =
         Chroot.absolute_path_outside chroot socket_path
       in
       Xenops_utils.best_effort "Stop listening on deprivileged socket"
-        (fun () ->
-          Varstore_privileged_client.Client.destroy dbg gid absolute_socket_path
+        (fun () -> G.destroy dbg ~domid:gid ~path:absolute_socket_path
       ) ;
       Chroot.destroy chroot
```
lindig, line 180: I believe only this should be guarded by `file_exists` and the rest is best effort - so let that code deal with the fact that the chroot does not exist but it still might want to do other clean-up.

ocaml/xenopsd/xc/device.ml
```diff
@@ -92,6 +90,170 @@ module Profile = struct
...
+          status.active_state
+    ) ;
+    is_active
+
+  type watch_trigger = Finished | Cancelled | Empty
+
+  exception ECancelled of Xenops_task.task_handle
+
+  let raise_e = function
+    | ECancelled t ->
```
lindig, line 132: Typo here? What is `ECancelled` vs `Cancelled`?

ocaml/xenopsd/xc/device.ml
```diff
@@ -92,6 +90,170 @@ module Profile = struct
...
+        [Inotify.S_Create]
+    in
+    Monitor.add pollfd notifd ;
+
+    let close_fds () =
+      Monitor.remove pollfd notifd ;
+      Polly.close pollfd ;
+      Inotify.rm_watch notifd ready_watch ;
+      Inotify.rm_watch notifd cancel_watch ;
+      Unix.close notifd
```
lindig, line 158: Should this be guarded by a try and any exception be logged because it would mean that resources leak?

ocaml/xenopsd/xc/device.ml
```diff
@@ -92,6 +90,170 @@ module Profile = struct
...
+    in
+
+    let wait ~task ~pid_path ~cancel_path ~for_s ~service_name =
+      let start_time = Mtime_clock.elapsed () in
+      let poll_period_ms = 1000 in
+      let cancel () =
+        (* create and empty file to trigger the watch and delete it
```
lindig, line 165: Typo "an"

psafont:
> Could this (daemons per domain) be also controlled by systemd?

The daemons per domain are spawned by systemd already.  The wait for PID could be done with systemd, but then we don't have an immediate feedback when the service is up because it forces xenopsd to do simple polling for state instead of being woken with a system call. (and for cancellation we would still have to use a different method)

ocaml/xenopsd/lib/xenops_sandbox.ml
psafont: I can cherry-pick the commit in the master branch, or we can merge master branch into this one.

ocaml/xenopsd/xc/device.ml
psafont: `ECancelled` is correct, we're matching exceptions. I would have like to not have to define `ECancelled` not `raise_e`, but `xcp-idl`'s `Xenops_task` is obtuse and makes it difficult to actually expose the type of exception, so `raise_cancelled` is used instead

ocaml/xenopsd/lib/xenops_sandbox.ml
psafont: I've changed the names of the module types and functors to something more palatable and consistent

ocaml/idl/datamodel_lifecycle.ml
psafont: I can remove this before merging, I committed because it otherwise disrupts the work-flow as the change reappears every time the project is compiled

ocaml/xenopsd/lib/xenops_sandbox.mli
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  val absolute_path_outside : t -> Path.t -> string
+  (** [absolute_path_outside chroot path] returns the absolute path outside the
+      chroot *)
+
+  val chroot_path_inside : Path.t -> string
+  (** [chroot_path_inside path] returns the path when inside the chroot *)
+
+  val of_domid : daemon:string -> domid:int -> vm_uuid:string -> t
```
lindig, line 17: The is indexed on both domid and uuid; I believe services are per domain, so domid would be enough. What creates the need to have the uuid as well? (Obviously we could also look up the domid for a uuid but not involving the database is a good idea)

lindig (approved):

ocaml/xenopsd/lib/xenops_sandbox.mli
psafont: the root directory needs both, unfortunately:
```ocaml
    let root =
      if domid = 0 then
        Printf.sprintf "/var/run/xen/%s-root-%d-%s" daemon domid vm_uuid
      else
        Printf.sprintf "/var/run/xen/%s-root-%d" daemon domid
```

ocaml/xenopsd/xc/device.ml
```diff
@@ -92,6 +90,170 @@ module Profile = struct
...
+                "Exception while waiting for service %s to be ready: %s"
+                service_name (Printexc.to_string e)
+            in
+            Error (Failure err_msg)
+        in
+
+        Xenops_task.with_cancel task cancel poll_loop
+      in
+      cancellable_watch ()
+    in
```
edwintorok, line 240: This is a lot simpler to do in the Lwt variant of inotify (see https://github.com/xapi-project/xen-api/pull/4670/commits/9d12dcd9ebbe8515606532d5c10b20c6591414a6), but probably because Lwt already takes care of wrapping the event loop and waiting for an event, whereas here you have to do it "manually".

Also cancelation adds another layer of complexity on top of this. I'll see whether I can come up with a simpler implementation.

edwintorok: I think we can avoid using epoll though: xenopsd already has 1 thread/request, and there is nothing else to do if the file doesn't exist (it is not lke Lwt where it can go and execute something else), so doing a blocking Inotify.read should actually work and be simpler.

ocaml/xenopsd/lib/xenops_sandbox.mli
edwintorok: I think that is just to guard against domid reuse (e.g. if your domid has overflown because you've exceed the max domid - which is only 16-bits), to avoid accidentally using another VM's chroot (that may still have services that didn't get cleaned up). Although if we're careful about cleaning up chroot as domains die we don't need it, I thought it'd be worth having a bit of an extra layer of defense to avoid an XSA in the future.

ocaml/xenopsd/xc/device.ml
```diff
@@ -92,6 +90,175 @@ module Profile = struct
...
+    ; domid: domid
+    ; exec_path: string
+    ; pid_path: string
+    ; cancel_path: string
+    ; timeout_seconds: float
+    ; args: string list
+    ; execute: path:string -> args:string list -> domid:domid -> unit -> string
+  }
+
+  let alive service =
```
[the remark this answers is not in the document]
psafont: I've hidden these behind the interface in `ocaml/xenopsd/xc/service.mli`

edwintorok (approved):

PR #4670, opened 2022-04-06
CP-39134: add filtering for vTPM API calls

edwintorok:
swtpm runs in a sandbox and has access to a socket on which it can talk to XAPI.
xapi-guard is used to whitelist what API calls it can make, such that it can only access and change its own VM, and not other VMs.

This is already done for varstored and NVRAM API calls, implement filtering for VTPM API calls similarly.

Also add unit tests, and refactor/clean up the code a bit.


ocaml/xapi-guard/src/varstored_interface.ml
```diff
@@ -122,16 +126,8 @@ let logout session_id =
           Lwt.fail e
       )
 
-let cache = SessionCache.create ~login ~logout
-
 let shutdown = Lwt_switch.create ()
 
-let () =
```
psafont, line 129: Why isn't this needed?

ocaml/xapi-guard/lib/varstored_interface.ml
```diff
@@ -151,22 +152,61 @@ let with_xapi ~cache f =
...
       (* do not close both channels, or we get an EBADF *)
       Lwt_io.close ic
+  | Error e ->
+      D.debug "Waiting for socket to be connectable at %s: %s" path
+        (Printexc.to_string e) ;
+      (* just in case Lwt_unix.listen doesn't get called soon enough,
+         avoid using up 100% CPU waiting for it,
+         better than Lwt.pause ()
+      *)
+      let* () = Lwt_unix.sleep 0.001 in
```
psafont, line 173: The sleep value was changed, was this intentional?

ocaml/xapi-guard/src/varstored_interface.ml
edwintorok: It got moved into main.ml. There is no longer a global `cache` here, so there can't be a global shutdown hook either. The global cache is now declared in main.ml.
This is done so that the unit test can create and tear down the Lwt switch within each test (and test that this teardown actually works) instead of relying on a global (which could only test it once/process)

psafont: Oh, you're right, I missed that piece of code in main.ml

psafont (approved):

psafont:
Are there salvageable commits here?

(I was thinking about the unit-testing)

edwintorok:
I think I've included what is useful in my other PR already that got merged

edwintorok:
Opened ticket to salvage unit tests

PR #4672, opened 2022-04-07
CP-35846: Restrict access to internal yum repo server (members only)

robhoes:
This restricts the /repository endpoint to local-root only. This
endpoint is exposed only on the coordinator, and used by other pool
members to access the pool's yum repo mirror.

Yum calls to this endpoint now require a pool secret to be used in a
cookie, which is implemented through a yum plugin.

### DIFF
--- a/scripts/yum-plugins/ptoken.py
+++ b/scripts/yum-plugins/ptoken.py
@@ -1,0 +1,31 @@
+#!/usr/bin/python
+
+# Drop this file into /usr/lib/yum-plugins/
+# Enable it by creating conf file /etc/yum/pluginconf.d/ptoken.conf:
+#   [main]
+#   enabled=1
+#
+# Configure it by:
+# yum-config-manager --setopt=<repo-name>.ptoken=file://<token-file-path> --save
+
+from yum import config
+from yum.plugins import TYPE_CORE
+
+requires_api_version = '2.5'
+plugin_type = (TYPE_CORE,)
+
+def config_hook(conduit):
+    config.RepoConf.ptoken = config.BoolOption(False)
+
+def init_hook(conduit):
+    with open('/etc/xensource/ptoken') as f:
+        ptoken = f.read().strip()
+
+    repos = conduit.getRepos()
+    for name in repos.repos:
+        repo = repos.repos[name]
+        # Only include the ptoken for repos with a localhost URL, for added safety.
+        # These may be proxied to the coordinator through stunnel, set up by xapi.
+        if len(repo.baseurl) > 0 and repo.baseurl[0].startswith("http://127.0.0.1") \
+            and repo.getConfigOption('ptoken'):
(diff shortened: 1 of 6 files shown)


scripts/yum-plugins/ptoken.py
```diff
@@ -0,0 +1,28 @@
...
+
+def init_hook(conduit):
+    with open('/etc/xensource/ptoken') as f:
+        ptoken = f.read().strip()
+
+    repos = conduit.getRepos()
+    for name in repos.repos:
+        repo = repos.repos[name]
+        if repo.getConfigOption('ptoken'):
+            repo.http_headers['cookie'] = "pool_secret=" + ptoken
```
edwintorok, line 28: If we're sending the pool token along with every request we should probably forbid http:// URIs in our yum config where ptoken is used. i.e. you either use an internal mirror through http, or the one through ptoken but that must be https.

robhoes: Yes, that is already the case. See:
* https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/repository.ml#L466
* https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/fileserver.ml#L83

minglumlu (approved):

scripts/yum-plugins/ptoken.py
```diff
@@ -0,0 +1,28 @@
...
+def config_hook(conduit):
+    config.RepoConf.ptoken = config.BoolOption(False)
+
+def init_hook(conduit):
+    with open('/etc/xensource/ptoken') as f:
+        ptoken = f.read().strip()
+
+    repos = conduit.getRepos()
+    for name in repos.repos:
+        repo = repos.repos[name]
```
minglumlu, line 26: Is it worth to restrict the repo URL to "http://127.0.0.1:8000"?

robhoes: How exactly? Something like `if repo.getConfigOption('baseurl') != "http://127.0.0.1:8000": continue`?

lindig: Do we have to worry about the token containing anything that would need escaping when passed in an URL? Or would whoever constructs the URL take care of this?

lindig (approved):

scripts/yum-plugins/ptoken.py
robhoes: It goes into a cookie header, not the URL. And it's a fixed format, defined by xapi, that doesn't have weird characters.

minglumlu: > How exactly? Something like `if repo.getConfigOption('baseurl') != "http://127.0.0.1:8000": continue`?

Yes. Maybe we have to have some hard-coding in this plugin.

robhoes: I've now added a restriction for the baseurl to localhost.

PR #4676, opened 2022-04-08
CA-366014: avoid qemu device state issues during migration in UEFI mode

edwintorok:
See [this commit message ](https://github.com/xapi-project/xen-api/commit/5f33eb2fd251e7183098f4ef96233f28b6198974) for details.

### DIFF
--- a/ocaml/xenopsd/xc/domain.ml
+++ b/ocaml/xenopsd/xc/domain.ml
@@ -1177,11 +1177,13 @@ let build (task : Xenops_task.task_handle) ~xc ~xs ~store_domid ~console_domid
 
 type suspend_flag = Live | Debug
 
-let dm_flags = function
-  | Device.Profile.Qemu_upstream | Device.Profile.Qemu_upstream_compat ->
-    ["-dm"; "qemu"]
-  | _ ->
-    []
+let dm_flags =
+  let open Device.Profile in
+  function
+  | Qemu_upstream | Qemu_upstream_compat | Qemu_upstream_uefi ->
+      ["-dm"; "qemu"]
+  | Qemu_trad | Qemu_none ->
+      []
 
 let with_emu_manager_restore (task : Xenops_task.task_handle) ~domain_type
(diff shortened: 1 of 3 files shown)


edwintorok:
Discovered while testing vtpm, but affects master as well, so opened PR towards master, will mark it as ready for review once I've done some more testing on it.

ocaml/xenopsd/xc/domain.ml
```diff
@@ -1208,12 +1216,7 @@ let with_emu_manager_restore (task : Xenops_task.task_handle) ~domain_type
     ; "-console_port"
     ; string_of_int console_port
     ]
-    @ ( match dm with
-      | Device.Profile.Qemu_upstream | Device.Profile.Qemu_upstream_compat ->
-          ["-dm"; "qemu"]
-      | _ ->
-          []
-      )
+    @ dm_flags dm
```
lindig, line 1219: Maybe constructing `args` should use `List.concat` to avoid the O(n^2) or `@`?

lindig (approved):

robhoes (approved):

robhoes:
How's testing for this going?

edwintorok:
I've tested this by manually migrating a UEFI windows VM on localhost. Although I was not able to reproduce the crash by moving the mouse on the VNC console I confirmed that the `-dm qemu` parameter gets added properly now:
```
xenopsd-xc: [debug||30 |VM.pool_migrate R:da6ccebcd44b|emu-manager] connect: args = [ -mode hvm_restore -domid 10 -fd bd22e3dc-8a0c-4bce-b8a8-2fb7b01d0301 -store_port 17 -console_port 18 -dm qemu ]
```

vs before:
```
xenopsd-xc: [debug||34 |VM.pool_migrate R:98e7e6896e5f|emu-manager] connect: args = [ -fd 35ce8856-5124-469a-81ec-95236f3ca000 -mode hvm_save -domid 7 -live true ]```

Whether that actually passes the right Qmp command to qemu is a bit harder to tell because we don't log or trace Qmp commands by default, but that'll be up to emu-manager to do the right thing here.

PR #4678, opened 2022-04-11
Support livepatch - part 1

minglumlu:
This series contains the first part of changes to support live patch:
1. some existing codes are split into new files;
2. new functions to parse metadata of live patches from updateinfo;
3. get the status of running live patch (Xen and Kernel) on host;
4. get all applicable live patches from updateinfo;
5. calculate and consolidate guidance.


ocaml/xapi/livepatch.ml
```diff
@@ -0,0 +1,343 @@
...
+module D = Debug.Make (struct let name = "livepatch" end)
+
+open D
+
+exception Invalid_component
+
+type component = Xen | Kernel
+
+let component_of_string = function
+  | "Xen" | "xen" ->
```
lindig, line 24: It would be better to use `String.lowercase_ascii` before the match is made.

ocaml/xapi/livepatch.ml
```diff
@@ -0,0 +1,343 @@
...
+    {
+      component= member "component" js |> to_string |> component_of_string
+    ; base_build_id= member "base_build_id" js |> to_string
+    ; base_version= member "base_version" js |> to_string
+    ; base_release= member "base_release" js |> to_string
+    ; to_version= member "to_version" js |> to_string_option
+    ; to_release= member "to_release" js |> to_string_option
+    }
+  with e ->
+    let msg = "Can't construct a livepatch from json" in
```
lindig, line 74: This should either include a path to the JSON or the JSON itself to facilitate debugging.

ocaml/xapi/livepatch.ml
```diff
@@ -0,0 +1,343 @@
...
+    let r =
+      (* This expects the kernel livepatch module name should be built as:
+       * livepatch_<base_version>__<base_release>__<to_version>__<to_releaes>
+       * The "." in <base_version>, <base_release>, <to_version>, and <to_releaes>
+       * should be replaced with "_".
+       * For example:
+       * livepatch_4_19_19__8_0_20__4_19_19__8_0_22
+       *)
+      Re.Posix.compile_pat
+        {|^[ ]*livepatch_([^ \[]+)__([^ \[]+)__([^ \[]+)__([^ \[]+).*$|}
```
lindig, line 98: Can we make this regex stronger? `([0-9_]+)` seems to capture the expected version strings.

ocaml/xapi/livepatch.ml
```diff
@@ -0,0 +1,343 @@
...
+    in
+    Helpers.call_script !Xapi_globs.kpatch_cmd ["list"]
+    |> Astring.String.cuts ~sep:"\n"
+    |> List.fold_left
+         (fun (acc, acc_flag) line ->
+           match acc_flag with
+           | true ->
+               (acc, true)
+           | false ->
+               if
```
lindig, line 108: This could be expressed as `when` clause on the match case.

ocaml/xapi/livepatch.ml
```diff
@@ -0,0 +1,343 @@
...
+                 , Re.Group.get groups 2 |> replace
+                 , Re.Group.get groups 3 |> replace
+                 , Re.Group.get groups 4 |> replace
+                 )
+           | None ->
+               None
+       )
+    |> get_latest_livepatch
+
+  let get_base_build_id () =
```
lindig, line 134: What is this code doing? It seems quite repetitive and using low-level operations. Parsing binary data could be easier using `Angstrom` which has operations to read bytes and Big Endian, Small Endian numbers.

ocaml/xapi/livepatch.mli
```diff
@@ -0,0 +1,34 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+type component = Xen | Kernel
+
+type t = {
+    component: component
```
lindig, line 18: This needs documentation. What are these fields describing?

ocaml/xapi/rpm.mli
```diff
@@ -0,0 +1,60 @@
...
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+module Epoch : sig
```
lindig, line 15: This needs doucumentation. What are these modules describing?

ocaml/xapi/livepatch.ml
minglumlu: It is going to read ELF notes from file `/sys/kernel/notes` to find out the running kernel's build id.
The file content is organized as some consecutive sections. For each section, at beginning of it, it's a header structure described by `headers_struct`. And the length of a section is determined by its headers. This function needs to find a specific section and read build id in it.

If `Angstrom` works for this in better way, I will look into it. Thanks :)

lindig: * https://github.com/lindig/fit/blob/main/lib/fit.ml#L213-L222

For example, this is code I wrote to read the (binary) header of a FIT file. There is more complex code in that file but it gives you a flavor how to read bytes, multi-byte numbers, and strings.

psafont: There's another example of using angstrom for decoding the output of a command, it might prove useful as well: https://github.com/xapi-project/xen-api/blob/47fae74032aa6ade0fc12e867c530eaf2a96bf75/ocaml/xapi/bios_strings.ml

Looking at the documentation of angstrom was useful when I developed the module: https://docs.ocaml.pro/docs/LIBRARY.angstrom@angstrom.0.15.0/Angstrom/index.html

minglumlu: Hi @lindig @psafont 
Thanks for your pointers. I've reworked this with `Angstrom` which looks very great.

minglumlu:
Hi Reviewers,
I reworked based on existing comments. May I please have your review more?
I squashed the fixup as I think the changes in rework are too dramatic. Sorry for this.

ocaml/xapi/livepatch.ml
```diff
@@ -0,0 +1,343 @@
...
+                   Some (b_v, b_r, t_v, t_r)
+               | _ ->
+                   None
+             )
+           | None ->
+               None
+       )
+    |> get_latest_livepatch
+
+  let get_base_build_id () =
```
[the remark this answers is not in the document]
andyhhp: Yeah - `/sys/hypervisor/properties/buildid` is looking very broken, so use `xl info build_id` for now.

ocaml/xapi/livepatch.ml
minglumlu: The version and release strings may contain other characters. Actually we don't have any expectations on the strings.  Here we only need to separate them.

ocaml/xapi/livepatch.ml
```diff
@@ -0,0 +1,279 @@
...
+      "xen-hypervisor"
+  | Kernel ->
+      "kernel"
+
+let string_of_component = function Xen -> "xen" | Kernel -> "kernel"
+
+type t = {
+    component: component
+  ; base_build_id: string
+  ; to_version: string option
```
lindig, line 43: I am bit surprised about `to_` - it is somewhat unusual. Does this indicate the version and release reached after applying the patch? Maybe a comment could clarify this.

ocaml/xapi/livepatch.ml
```diff
@@ -0,0 +1,279 @@
...
+  |> function
+  | Some (v, r) ->
+      List.find_opt (fun (_, _, t_v, t_r) -> t_v = v && t_r = r) lps
+  | None ->
+      None
+
+module BuildId = struct
+  let one_byte =
+    let open Angstrom in
+    any_char >>= fun c ->
```
lindig, line 88: I'd use `any_uint8` or `any_int8` - which will result in an `int`.

ocaml/xapi/livepatch.ml
```diff
@@ -0,0 +1,279 @@
...
+        take (align desc_size) >>= fun desc ->
+        let name' = Astring.(String.trim ~drop:Char.Ascii.is_control name) in
+        match (note_type, desc_size, name') with
+        | 3l, 20l, "GNU" ->
+            return desc
+        | _ ->
+            section' ()
+      )
+
+  let section = section' ()
```
lindig, line 167: Could we rename `section'` to `section` and get rid of this?

ocaml/xapi/livepatch.ml
```diff
@@ -0,0 +1,279 @@
...
+                 )
+           | None ->
+               None
+       )
+    |> get_latest_livepatch
+
+  let align x =
+    (* The actual size is aligned with 4. *)
+    let open Int32 in
+    (if rem x 4l = 0l then x else add x (sub 4l (rem x 4l))) |> to_int
```
lindig, line 146: The code is using `41` but the comment says something is 4-byte aligned. What is the relevance of `41`? What is the result of this function for a given x?

ocaml/xapi/livepatch.ml
```diff
@@ -0,0 +1,279 @@
...
+end
+
+let get_livepatch_dir ~component ~base_build_id =
+  let comp_str = string_of_component component in
+  Printf.sprintf "/usr/lib/%s-livepatch/%s" comp_str base_build_id
+
+let get_installed_livepatch_file ~component ~base_build_id =
+  let lp_dir = get_livepatch_dir ~component ~base_build_id in
+  let installed_symlink = Filename.concat lp_dir "livepatch.livepatch" in
+  if not (Sys.file_exists installed_symlink) then
```
lindig, line 237: This could be done with a single match: 
```
match Unix.lstat installed_symlink with
| Unix.{st_kind = S_LNK;_} -> ... Option .some
| exception _ -> None
```

ocaml/xapi/rpm.ml
```diff
@@ -0,0 +1,290 @@
...
+
+  type order = LT | EQ | GT
+
+  type segment_of_version = Int of int | Str of string
+
+  let string_of_order = function LT -> "<" | EQ -> "=" | GT -> ">"
+
+  let error_msg = Printf.sprintf "Failed to parse '%s'"
+
+  let parse_epoch_version_release epoch_ver_rel =
```
lindig, line 61: This module looks too complicated to me. I would expect that a version is represented as a list of numbers and that we are comparing two strings to establish their order. I think it needs to be spelled out how strings and numbers are treated: is every string that can be read as a number considered a number, and or still a string? This makes a difference because "01" and "1" are different strings but the same number. Likewise, how is any string that cannot be read as a number treated? I am not sure but I would have started with:
```
type comp = Number of int | String of string
type version = comp list (* assuming a version is a list of components, each either a string or a number *)
val parse: string -> string * version (* split into package and version *)
val compare: version -> version -> int (* -1, 0, 1 *)
```

lindig (commented):
The code to extract the build information is much better now.

ocaml/xapi/livepatch.ml
lindig: Sorry, I misread this. The `1` is actually an `l` - a size indicator. A comment what `align 6` would do would be still helpful. Are we rounding up or down to a multiple of 4? I assume `align x` rounds x up to the next multiple of 4. I believe it could be simplified (here for integers):
```

Comment (minglumlu) [reply ocaml/xapi/livepatch.ml:146]: Yeah, this is great. Thanks.

File: ocaml/xapi/livepatch.ml (hunk 9 above)
Comment (minglumlu) [reply ocaml/xapi/livepatch.ml:43]: Yes. It indicates that the version and release reached after applying the live patch.

File: ocaml/xapi/livepatch.ml (hunk 11 above)
Comment (minglumlu) [reply ocaml/xapi/livepatch.ml:167]: I was not quite understand the error in that case:
`Error: This kind of expression is not allowed as right-hand side of 'let rec'`
So I added a `section`.

File: ocaml/xapi/rpm.ml (hunk 14 above)
Comment (minglumlu) [reply ocaml/xapi/rpm.ml:61]: These bits are some existing lines which are moved into a new file only in this PR.

`type comp = Number of int | String of string`
This is defined as similarly as:
https://github.com/xapi-project/xen-api/pull/4678/files#diff-3c831fd11744c8735ff48fa1f9e436f357cb4316e4394855770ac300ed41dc69R55

`val parse: string -> string * version (* split into package and version *)`
is implemented like:
https://github.com/xapi-project/xen-api/pull/4678/files#diff-3c831fd11744c8735ff48fa1f9e436f357cb4316e4394855770ac300ed41dc69R183

and `val compare: version -> version -> int (* -1, 0, 1 *)` is implemented as:
https://github.com/xapi-project/xen-api/pull/4678/files#diff-3c831fd11744c8735ff48fa1f9e436f357cb4316e4394855770ac300ed41dc69R160

There are some examples to explain the rules of comparison:
https://github.com/xapi-project/xen-api/pull/4678/files#diff-3c831fd11744c8735ff48fa1f9e436f357cb4316e4394855770ac300ed41dc69R166

File: ocaml/xapi/repository.ml [hunk 15]
```diff
@@ -304,6 +334,17 @@ let set_available_updates ~__context =
         Helpers.run_in_parallel funs capacity_in_parallel
     )
   in
+  let applied_livepatches_of_hosts =
```
lindig, line 337: How many lines is this function long? I would encourage to split out more functions that are defined in this function.  If this is done just to keep the name space clean, I would use a local module rather than making this function larger and larger.

lindig (approved):

ocaml/xapi/repository.ml
minglumlu: Thanks. I split some functions out. Indeed, the `set_available_updates` function was too long.

ocaml/xapi/livepatch.ml
robhoes: I see that there are comments that explain this in the mli file.

ocaml/xapi/livepatch.ml
robhoes: This is now logged below.

ocaml/xapi/livepatch.ml
robhoes: Another option might be to chop off the `livepatch_` prefix and then split the string on `__`. I suppose it doesn't really matter.

lindig: If we can describe the operation with a regular expression (because the number of fields  is known in advance) I would prefer that over a procedural approach of chopping an splitting - let the regex engine do the work. A regex doesn't work if this becomes context sensitive - earlier information informs later decisions. In that case I would again consider using Angstrom because it is still quite declarative.

robhoes (approved):
It's not easy to get a good overview of the logic and control flow because there is a lot going on in the >1000 changed lines. I think it comes down to testing well. It mostly looks fine.

robhoes:
Please do not merge yet: we are sorting out some build issues first.

minglumlu:
> It's not easy to get a good overview of the logic and control flow because there is a lot going on in the >1000 changed lines. I think it comes down to testing well. It mostly looks fine.

Sure. I will re-do tests before merging this one. And the following PRs will be unit tests and applying live patches.

minglumlu:
Hi @lindig @robhoes 
I appended a new commit for a minor fixup. https://github.com/xapi-project/xen-api/pull/4678/commits/170e6c91ae64c732a9a015dbd26476ef0a937bc5
Would you please have a review again on it?

robhoes:
Looks like there are now a few conflicts due to the warning-removal PR...

minglumlu:
fe0634c0d is result of rebasing and squashing.
8017dcf8e is a fixup for an issue found in test.
de7c89759 is a fixup for "unused" warnings

PR #4679, opened 2022-04-11
CP-39551: avoid warnings in xapi-cli-server and xenopsd/xc

psafont:
Please review both commits separately, in particular:
- the xenopsd drops code that was surprisingly unused around PCI devices, should it be used somewhere?
- Most of the (huge!) churn in xapi-cli-server is around adding named parameter to api calls and removing unused parameters / bindings

I ran a suite run 167355 with BST and BVT with the changed and it turned out green, do we want to run some more tests to be safe around the cli server?

### DIFF
--- a/ocaml/xenopsd/xc/device.ml
+++ b/ocaml/xenopsd/xc/device.ml
@@ -86,6 +86,8 @@ module Profile = struct
         |> fun s -> Xenopsd_error (Internal_error s) |> raise
     | x when x = Name.qemu_upstream_uefi ->
         Qemu_upstream_uefi
+    | x when x = Name.qemu_none ->
+        Qemu_none
     | x ->
         debug "unknown device-model profile %s: defaulting to %s" x
           Name.qemu_upstream_compat ;
@@ -222,11 +224,6 @@ module Generic = struct
       true
     with Xs_protocol.Enoent _ -> false
 
-  let assert_exists_t ~xs t (x : device) =
-    let backend_stub = backend_path_of_device ~xs x in
-    try ignore_string (t.Xst.read backend_stub)
-    with Xs_protocol.Enoent _ -> raise Device_not_found
-
   (** When hot-unplugging a device we ask nicely *)
   let clean_shutdown_async ~xs (x : device) =
     let backend_path = backend_path_of_device ~xs x in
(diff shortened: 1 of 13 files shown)


ocaml/xenopsd/xc/device.ml
```diff
@@ -2371,9 +2289,6 @@ module Dm_Common = struct
      -net nic,vlan=1,macaddr=00:16:3E:76:CE:44,model=rtl8139 -net
      tap,vlan=1,bridge=xenbr0 -vnc 39 -k en-us -vnclisten 127.0.0.1] *)
 
-  (** Should be <= the hardcoded maximum number of emulated NICs *)
-  let max_emulated_nics = 8
```
edwintorok, line 2375: There is a NIC.max_emulated further down, looks like this was a duplicate.

ocaml/xenopsd/xc/device.ml
```diff
@@ -52,8 +52,6 @@ module Profile = struct
   let all = [Qemu_none; Qemu_upstream_compat; Qemu_upstream]
 
   module Name = struct
-    let qemu_none = "qemu-none"
```
edwintorok, line 55: It looks like there is a missing entry  in `Profile.of_string`. I think it'll be possible to run without Qemu once we have support for PVH (for which I had an old branch of mine). Instead of dropping qemu_none perhaps we can add it to of_string instead? There *is* a `Qemu_none` backend module there already.

ocaml/xenopsd/xc/device.ml
```diff
@@ -1953,89 +1934,26 @@ module PCI = struct
...
-          ; (Printf.sprintf "device-model/%d/parameter" domid, parameter)
-          ]
-    )
-
   (* Return a list of PCI devices *)
   let list = read_pcidir
-
-  (* We explicitly add a device frontend in the hotplug case so the device watch
-     code can find the backend and monitor it. *)
-  let ensure_device_frontend_exists ~xs backend_domid frontend_domid =
```
edwintorok, line 2003: This appears to be a remnant from  this old commit (which didn't remove all dead code):
```
commit a185ec6e1bee1c0957465cb5e60287c3cb487506
Author: Koushik Chakravarty <koushik.chakravarty@citrix.com>
Date:   Thu Aug 6 11:11:22 2015 +0000

    CA-168029: Using xl pci for plug/unplug in xenopsd-xc

    This xl pci-attach and pci-detach are working now and this
    patch makes xenopsd-xc to use them.
    Also removing some dead code paths
```

edwintorok (approved):

PR #4681, opened 2022-04-12
CA-366098: Clarify logs when a xenopsd task times out

psafont:
Previous message was not insightful at all, now it should be more
evident to triage an error when this message happens along a 20 minute
timeout in the logs

### DIFF
--- a/ocaml/xenopsd/lib/xenops_server.ml
+++ b/ocaml/xenopsd/lib/xenops_server.ml
@@ -1649,12 +1649,11 @@ let rec perform_atomic ~progress_callback ?subtask:_ ?result (op : atomic)
       (* make sure that we destroy all the parallel tasks that finished *)
       let errors =
         List.map
-          (fun task_handle ->
-            let id = Xenops_task.id_of_handle task_handle in
-            match Xenops_task.get_state task_handle with
-            | Task.Completed _ ->
+          (fun (id, task_handle, task_state) ->
+            match task_state with
+            | Some (Task.Completed _) ->
                 TASK.destroy' id ; None
-            | Task.Failed e ->
+            | Some (Task.Failed e) ->
                 TASK.destroy' id ;
                 let e =
                   match Rpcmarshal.unmarshal Errors.error.Rpc.Types.ty e with
@@ -1667,10 +1666,18 @@ let rec perform_atomic ~progress_callback ?subtask:_ ?result (op : atomic)
                         )
                 in
                 Some e
-            | Task.Pending _ ->
-                error "Parallel: queue_atomics_and_wait returned a pending task" ;
+            | None | Some (Task.Pending _) ->
+                (* Because pending tasks are filtered out in
+                   queue_atomics_and_wait with task_ended the second case will
+                   never be encountered. The previous boolean used in
+                   event_wait was enough to express the possible cases *)
+                let err_msg =
+                  Printf.sprintf "Timed out while waiting on task %s (%s)" id
+                    (Xenops_task.get_dbg task_handle)
+                in
+                error "%s" err_msg ;
                 Xenops_task.cancel task_handle ;
-                Some (Xenopsd_error (Cancelled id))
+                Some (Xenopsd_error (Internal_error err_msg))
           )
           task_list
       in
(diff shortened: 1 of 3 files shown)


edwintorok (approved):

ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -1668,9 +1668,12 @@ let rec perform_atomic ~progress_callback ?subtask:_ ?result (op : atomic)
                 in
                 Some e
             | Task.Pending _ ->
-                error "Parallel: queue_atomics_and_wait returned a pending task" ;
+                let err_msg =
+                  Printf.sprintf "Timed out while waiting on task %s" id
```
robhoes, line 1672: Yesterday we found that `event_wait` inside `queue_atomics_and_wait` returns a Boolean that indicates whether the wait timed out or returned normally, and that this is ignored everywhere. I think that we need to inspect this to know whether this pending task was cancelled or some operation had timed out.

psafont: Now the code makes a distinction between both when reporting the error.

ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -1667,10 +1666,17 @@ let rec perform_atomic ~progress_callback ?subtask:_ ?result (op : atomic)
...
                 Some e
-            | Task.Pending _ ->
-                error "Parallel: queue_atomics_and_wait returned a pending task" ;
+            | None | Some (Task.Pending _) ->
+                (* Because pending tasks are filtered out in
+                   queue_atomics_and_wait with task_ended the second case will
+                   never be encountered. The previous boolean used in
+                   event_wait was enough to express the possible cases *)
+                let err_msg =
+                  Printf.sprintf "Timed out while waiting on task %s" id
```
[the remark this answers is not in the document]
psafont: added `Xenops_task.get_dbg task_handle` which should contain more information

ocaml/xenopsd/lib/xenops_task.ml
```diff
@@ -39,28 +39,27 @@ let event_wait local_updates task ?from ?timeout_start timeout p =
...
       in
-      let success = List.fold_left (fun acc d -> acc || p d) false deltas in
-      let finished = success in
-      if not finished then
-        let elapsed = Unix.gettimeofday () -. start in
-        inner (timeout -. elapsed) (Some next_id)
-      else
-        success
+      match List.find_map filter deltas with
+      | Some success when p success ->
```
robhoes, line 43: I think that `success` here does actually mean that the task was successful, just that we got an event on the task, and `success` is the state of that task after the event. So perhaps call it `state`.

ocaml/xenopsd/lib/xenops_task.ml
```diff
@@ -39,28 +39,27 @@ let event_wait local_updates task ?from ?timeout_start timeout p =
...
+          Some success
```
robhoes, line 44: We get here if we have detected a change in state on the task, and the task has ended (either succeeded or failed) within the specified timeout.

ocaml/xenopsd/lib/xenops_task.ml
```diff
@@ -39,28 +39,27 @@ let event_wait local_updates task ?from ?timeout_start timeout p =
...
+      | _ ->
+          let elapsed = Unix.gettimeofday () -. start in
+          inner (timeout -. elapsed) (Some next_id)
     else
-      false
+      None
```
robhoes, line 49: We return `None` if the task didn't end before the timeout.

It would be handy to add comments for this and the cases above.

ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -1667,10 +1666,18 @@ let rec perform_atomic ~progress_callback ?subtask:_ ?result (op : atomic)
                         )
                 in
                 Some e
-            | Task.Pending _ ->
-                error "Parallel: queue_atomics_and_wait returned a pending task" ;
+            | None | Some (Task.Pending _) ->
+                (* Because pending tasks are filtered out in
+                   queue_atomics_and_wait with task_ended the second case will
+                   never be encountered. The previous boolean used in
```
robhoes, line 1672: OK I see, so when you get `None`, then the task state would have been Pending.

Also, `event_wait` calls `Xenops_task.check_cancelling task` at the end, which causes an exception to be raised if the task has `cancelling` set to `true`. If fact, this can happen at any time during the 20-minute wait, because of the `~with_cancel:(Xenops_task.with_cancel task)` on `Updates.get`.

So the case here really is only a timeout.

robhoes (approved):

ocaml/xenopsd/lib/xenops_task.ml
psafont: Added some comments in `event_wait`

ocaml/xenopsd/lib/xenops_task.ml
psafont: renamed to found as `event_wait` "looks" for an event.

psafont:
Ring3 BST + BVT run passing: 16812

PR #4683, opened 2022-04-13
CP-39551: avoid warnings in xenopsd/xc and libs

psafont:
The xenopsd warnings were introduced by late merges, the libs warnings are intresting because they highlight that httpsvr never gets used to serve from file descriptors

The libs change passed ring3 BVT + BST, so it's more a matter of remembering this commit whenever we need additional behaviour in http svr or stunnel


edwintorok (approved):
Deleting dead code is useful so we don't get confused when debugging something

PR #4684, opened 2022-04-13
CP-39551: Remove all compiler warnings (finally!)

psafont:
The last part of the repo with warnings is xapi, these are mostly about using named args and unused bindings, with some dead code removal, unused opens and some interesting changes:
- several `call_api_functions` were not needed as the internal functions did not need rpc access at all (for example in xapi.ml or when verifying pool homogeneicity)
- the module xapi_host_cpu was unused, I've removed it
- some minor refactoring in sexpr

the second commit addresses the warnings introduced by the recent uefi certs PR. Could you take a look, @benjamreis ?

The changes pass the ring3 BST + BVT suites, it got run along the rest of the warning removals.


benjamreis (approved):
LGTM.

lindig (approved):
Almost all changes are purely syntactic. The amount of changes make this a challenge to review but I believe this kind of clean-up is necessary as it gives us better compiler support in the long run.

robhoes:
300 lines removed... nice :)

ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -525,22 +525,6 @@ module AuthADlw : Auth_signature.AUTH_MODULE = struct
     *)
     extract_sid_from_group_list subject_attrs
 
-  let pbis_get_sid_bygid gid =
```
psafont, line 528: @liulinC should this function be used somewhere?

robhoes (approved):
Fantastic :)

Now that so many functions have `_` args, I suppose there is scope for removing those from the callers (on another day).

psafont:
> Now that so many functions have _ args, I suppose there is scope for removing those from the callers (on another day).

In the cases where the functions don't need to conform to an interface, this is a possibility yes.

PR #4687, opened 2022-04-18
Add temporary feature 'Internal_repo_access' to allow update in mix mode

minglumlu:
In commit d951478f3, a new function was added to restrict the repository
access to pool members only. With this funciton, the pool members need
to use pool_secret in access requests to the mirrorred repositories on
the pool coordinator for authentication.

But in updating, this would cause the pool members can't access the
mirrored repositories as at that time the pool coordinator has been
updated to check the pool_secret while the pool members have not been
udpated to use pool_secret.

Per suggestion from Rob Hoes, this commit adds a temporary feature flag
to make the pool coordinator be able to distinguish the case in the
mixed mode in updating.

### DIFF
--- a/ocaml/xapi/repository.ml
+++ b/ocaml/xapi/repository.ml
@@ -465,24 +465,44 @@ let get_repository_handler (req : Http.Request.t) s _ =
   req.Request.close <- true ;
   if Fileserver.access_forbidden req s then
     Http_svr.response_forbidden ~req s
-  else if is_local_pool_repo_enabled () then (
-    try
-      let len = String.length Constants.get_repository_uri in
-      match String.sub_to_end req.Request.uri len with
-      | uri_path ->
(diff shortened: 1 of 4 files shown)


ocaml/xapi/repository.ml
```diff
@@ -465,24 +465,44 @@ let get_repository_handler (req : Http.Request.t) s _ =
...
+      Pool_features.is_enabled ~__context Features.Internal_repo_access
+    in
+    match (can_be_authorized, internal_repo_access_only) with
+    | false, true ->
+        error
+          "Invalid secret for authorization when Internal_repo_access is \
+           enabled" ;
+        Http_svr.response_forbidden ~req s
+    | _ -> (
+      try
```
lindig, line 488: This code looks fragile but existed before. We should user higher-level functions to decompose request paths.

lindig (approved):

PR #4688, opened 2022-04-19
ci: fix testing of xapi-xenstored in newer opam's sandboxes

psafont:
xs-opam's CI is the only environment where /tmp is not writable and
XDG_RUNTIME_DIR is not set. Using TMPDIR with /tmp as a backup will mean
the socket is writeable in all environments

xs-opam's CI successfuly passes with this branch: https://github.com/psafont/xs-opam/runs/6075835428?check_suite_focus=true

### DIFF
--- a/ocaml/forkexecd/test/fe_test.sh
+++ b/ocaml/forkexecd/test/fe_test.sh
@@ -1,7 +1,8 @@
 #!/bin/sh
 
 # Use user-writable directories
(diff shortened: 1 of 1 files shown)


edwintorok (approved):

edwintorok:
I thought this worked in the past though, has XDG_RUNTIME_DIR got dropped by accident somewhere in opam?

lindig (approved):

PR #4689, opened 2022-04-20
CA-366430: do not wipe PK.auth/dbx.auth

edwintorok:
These files are owned by other RPM packages.
Also upgrade would need to be taken into consideration where
uefi-certificates would already be set on a pool, but just with KEK+db
without PK+dbx.
In that case if we order uploading PK/dbx and restart xapi the wrong way
around we may get these files wiped completely from the system.
Wiping just KEK+db should be enough to ensure we don't have stale files
around.

Fixes a regression introduced by 53d656bec66d903425fc7bfe580d1425da5f2745

### DIFF
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -2593,7 +2593,7 @@ let write_uefi_certificates_to_disk ~__context ~host =
             let path = Filename.concat !Xapi_globs.varstore_dir name in
             Unixext.unlink_safe path
           )
-          ["PK.auth"; "KEK.auth"; "db.auth"; "dbx.auth"] ;
+          ["KEK.auth"; "db.auth"] ;
         (* No uefi certificates, nothing to do. *)
         if contents <> "" then (
           with_temp_file_contents ~contents


robhoes (approved):

stormi:
Please do not merge right now if possible as I'm writing a comment with a slightly different proposal.

stormi:
I agree that we should not have removed PK.auth, as it is necessary for UEFI VMs even without secure boot.

However, in XCP-ng we do need `dbx.auth` to be properly cleaned up when appropriate. We offer choice to user: they can download the latest DBX from microsoft, or an earlier version, or use none at all if they want. We documented it [here](https://xcp-ng.org/docs/guides.html#configure-the-pool) and users liked the control we gave them.

I think it's the right occasion to make things a bit cleaner, as it's not a good thing that anything writes in `/usr/share/varstored` as is currently the case in Citrix Hypervisor.  

In XCP-ng, we put the certificates in `/var/lib/uefistored`. There's also a `PK.auth` in `/usr/share/uefistored` which comes from the RPM. This `PK.auth` is copied to `/var/lib/uefistored` when the RPM is installed (`%post` scriptlet) and may be changed by users if needed with our `secureboot-certs install` script.

Would this be possible in Citrix Hypervisor:
* Let varstored read certificates from `/var/lib/varstored`, not from `/usr/share/varstored`. Change the directory in XAPI too, and in the systemd service that writes the certs it read from the host.
* Keep providing `PK.auth` and `dbx.auth` with the `varstored` RPM, in `/usr/share/varstored`
* Do not wipe `PK.auth` indeed
* At XAPI startup, if there's no PK in `/var/lib/varstored`, copy it from `/usr/share/varstored/PK.auth`
* Optionnally, do the same for `dbx.auth`

stormi:
By the way, 8.2 CU1 doesn't have `dbx.auth` in the RPM. Was it added recently?

stormi:
Note: there's another use for removing the files before extracting the tarball, by the way. We found out with our automated tests that the extract code is buggy and sometimes corrupts the target files if you don't remove them first. This is probably unlikely to happen to users (and won't happen if the files are already identical so their version in the tarball), but I wouldn't rule it out completely. If some files are not to be removed before the extract operation, then maybe the files should first be extracted to a temporary directory and then copied over the existing files to workaround the issue in the lib that is used for that.

stormi:
(In case you already read it: I updated my proposal above with what I think is a better solution)

lindig (approved):

edwintorok:
The dbx.auth is part of secureboot-certificates RPM which has a different license field on it (due to the license of the dbx.auth file itself).
I agree that we need a better way to manage these certificates, and move files out of the way, but we need to take upgrade scenarios into account as well, and is best addressed in another PR, i.e. lets undo the regression first to unblock testing, and then think about how to best implement and test changes to uefi cert management in a way that doesn't break on upgrades.
The current solution we have in XAPI only ever meant to be temporary

edwintorok:
A possibility might be to make the location that varstored reads the certificates from configurable, so xapi can pass at runtime the directory it used to store the files in.
And then we don't have a problem that we read a file from a directory that we're about to wipe (and risk losing the only copy of the file if scripts run in wrong order), and the complexity of how to deal with the upgrade at the RPM level: the RPMs that ship these .auth files can keep putting them where they currently are, just varstored will read them from somewhere else.
XAPI could then use /usr/share/varstored as a fallback to "fill in" anything that is missing from the tarball (i.e. tarball takes precedence).

i.e. essentially your idea but instead of hardcoding the other path (/var/lib/varstored), we make it configurable at runtime (so we can keep using /usr/share/varstored if asbent, or switch to another location without having to patch/rebuild varstored for that).

PR #4690, opened 2022-04-22
maintenance: make xapi-xenops tests more granular

psafont:
In practice means single values don't block others values from being
tested. Using alcotest failures make test failures easier to read and
reason about.


psafont:
Just saw that there are unit test for VM_platform, so I don't know why are most of these test in testing "xapi_xenops", seems like moving them along the other vm_platform tests is better.

lindig (approved):

robhoes (approved):

PR #4691, opened 2022-04-22
Don't use --force in gzip decompress

lindig:
Normally gzip stops reading input when it has reached its own end-of
input marker even when that is before the end of input. This is
overridden by --force for decompression and it is creating a problem for
us because suspend images on disk are represented as devices that are
larger than the actual content - so --force kept reading beyond the
actual input and this in turn was not accepted by the code taking the
uncompressed output.

### DIFF
--- a/ocaml/libs/xapi-compression/xapi_compression.ml
+++ b/ocaml/libs/xapi-compression/xapi_compression.ml
@@ -44,7 +44,11 @@ module Make (Algorithm : ALGORITHM) = struct
     let open Safe_resources in
     Unixfd.with_pipe ~loc:__LOC__ () @@ fun zcat_out zcat_in ->
     let args =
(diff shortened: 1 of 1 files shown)


robhoes:
Did you test that this works for export/import?

lindig:
I ran BVT/BST - assuming that this would test it

robhoes:
I don't think so. It's better to test it directly to be sure.

lindig:
I could also introduce a force flag to retain the existing behavior. Will do some manual testing, too.

lindig:
I've run a `xe vm-export compress=true` with a subsequent import. This uses gzip compression and it worked.

lindig:
Don't merge yet. There is one failed sub test on the BST which I will re-run

robhoes (approved):

lindig:
The previously failed test has passed in the re-run. So this should be save to merge.

psafont (approved):

PR #4692, opened 2022-04-25
Support livepatch - part 2 - unit tests

minglumlu:
This series contains unit tests against:

- the module Livepatch for handling running live patch;
- the logic to parse livepatch related metadata in updateinfo.xml;
- the logic to calculate guidance with considering livepatchGuidance.


ocaml/tests/test_repository_helpers.ml
```diff
@@ -511,551 +511,1001 @@ module EvalGuidanceForOneUpdate = Generic.MakeStateless (struct
...
       Fmt.(str "%a" Dump.(string)) (UpdateInfo.guidance_to_string g)
   end
 
-  let transform (updates_info, update) =
-    eval_guidance_for_one_update ~updates_info ~update ~kind:Guidance.Absolute
-      ~upd_ids_of_livepatches:UpdateIdSet.empty
+  let transform ((updates_info, update), upd_ids_of_livepatches) =
+    eval_guidance_for_one_update ~updates_info ~update
+      ~kind:Guidance.Recommended
+      ~upd_ids_of_livepatches:(UpdateIdSet.of_list upd_ids_of_livepatches)
```
psafont, line 529: Could the changes here be explained in the commit message? These seem important and introduce quite a lot of churn in the test data, it would be good to know why

minglumlu: Sure. I will add a commit message. The message would be:
Prior to support live patch, the guidance of an update is determined by "recommendedGuidance". With supporting to live patch, a new metadata "livepatchGuidance" is introduced. In an update, if a live patch in it is determined as applicable, its "livepatchGuidance" will overwrite its "recommendedGuidance".

ocaml/xapi/livepatch.ml
```diff
@@ -88,7 +88,7 @@ module BuildId = struct
 end
 
 module KernelLivePatch = struct
-  let get_running_livepatch () =
+  let get_running_livepatch' s =
```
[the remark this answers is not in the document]
robhoes: The original function (now below) called a script and then processed the output, so did more "getting". The "running" bit seems important though, because it looks for already loaded/applied/running livepatches.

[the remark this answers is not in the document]
lindig: This function does nothing about running - it is extracting a version from a string. The string might come from a running system, though. So it would be at the call site where "running" should appear.

ocaml/tests/test_rpm.ml
```diff
@@ -0,0 +1,190 @@
...
+    let string_of_output_t =
+      Fmt.(
+        str "%a"
+          Dump.(result ~ok:(option @@ record @@ fields_of_pkg) ~error:exn)
+      )
+  end
+
+  exception Can_not_parse of string
+
+  let transform input =
```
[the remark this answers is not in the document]
robhoes: This is part of the functor interface. It is meant to transform something of type `input_t` to `output_t`, where `tests` below lists expected input/output combinations.

ocaml/xapi/livepatch.ml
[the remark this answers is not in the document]
robhoes: It does if you look at the tests: it only takes things from the string (which is the output from a command) that indicates applied patches and ignores the not-yet-applied ones.

robhoes:
It looks that 19a60eb22f36cc98313c2ae888110e61ba6ac873 (the second patch) is only moving test code from one file into two new ones.

ocaml/tests/test_repository_helpers.ml
robhoes: Looks like this hasn't yet been added to the commit message.

robhoes:
Hiding whitespace to review the last patch is also helpful to see the actual changes.

ocaml/tests/test_repository_helpers.ml
```diff
@@ -511,551 +511,1001 @@ module EvalGuidanceForOneUpdate = Generic.MakeStateless (struct
...
+                    ; update_type= "security"
+                    ; livepatch_guidance= Some Guidance.RestartDeviceModel
+                    ; livepatches=
+                        [
+                          LivePatch.
+                            {
+                              component= Livepatch.Xen
+                            ; base_build_id=
+                                "9346194f2e98a228f5a595b13ecabd43a99fada0"
+                            ; base_version= "4.19.19"
```
robhoes, line 1218: Is the point here that these versions don't match the ones in the Update below?

robhoes: Or that the update ids don't match? I don't understand "livepatch does not come from myself" in the comment above.

robhoes (approved):
A few minor comments above, mainly to explain things a bit better. Otherwise it looks fine.

ocaml/xapi/livepatch.ml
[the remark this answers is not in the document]
lindig: Fair; my argument was, that this is a string function and not a function that obtains something from the running system. But you can argue that "running" is still a property of the version that this function is extracting. We can keep this.

lindig (approved):

ocaml/tests/test_repository_helpers.ml
minglumlu: I updated the comment to make it clearer. 
Here the `myself` refers to the one `update` in `eval_guidance_for_one_update`. This function evaluate guidance against this RPM update. But this evaluation needs to consider livepatch as well:
1. If  this RPM update has a livepatch (its update info is contained in the ones introduced by applicable livepatches), this is for the original comment "livepatch comes from myself"
2. Otherwise, "livepatch doesn't come from myself".
But anyway, I refined the comments and test data to make it clearer.

minglumlu:
After a fixup commit 0185698, this series have been squashed and re-based (against current master).

PR #4693, opened 2022-04-25
CP-39414: start swtpm depending on the VM's platform metadata

psafont:
Now swtpm is only started for domains that have vtpm=true added to their
platform's metada. This can be done using xe e.g.
xe vm-param-add uuid=VM-XXX param-name=platform vtpm=true

Shutting down domains without the swtpm and the xapi-guard services
running works as normal and without any errorsNow swtpm is only started for domains that have vtpm=true added to their
platform's metada. This can be done using xe e.g.
`xe vm-param-add uuid=VM-XXX param-name=platform vtpm=true`

Shutting down domains without the swtpm and the xapi-guard services
running works as normal and without any errors.


edwintorok:
I think we should create a vTPM object when vtpm=true, and then start/stop the vtpm daemon based on the presence of the vtpm object. Might need a bit more plumbing, see how the bios vs uefi is done in `xenops_types` `firmware_type`.
Do we also want to define a 'module VTPM' in device.ml?

psafont:
Now an optional variant is passed through the idl so all the logic to decide what information gets passed around can be done in xapi. Xenopsd can decide how to set up swtpm based on that information.

This should make creating the vtpm object and changing how to set up swtpm much simpler

ocaml/xapi-idl/xen/xenops_types.ml
```diff
@@ -77,6 +77,8 @@ module Vm = struct
 
   let default_firmware = Bios [@@deriving rpcty]
 
+  type tpm_type = Vtpm [@@deriving rpcty, sexp]
```
lindig, line 80: A type with type it its name? Is `tpm` enough (or we could use something longer because it is less frequently used: `trusted_platform`). I see tpm and vtpm being used. Is there anything other than vtpm?

ocaml/xapi/xapi_xenops.ml
```diff
@@ -259,6 +259,13 @@ let firmware_of_vm vm =
   | exception Not_found ->
       default_firmware
 
+let tpm_of_vm vm =
+  match List.assoc_opt "vtpm" vm.API.vM_platform with
```
lindig, line 263: Should this use `String.lowercase_ascii` and also match `1`?

ocaml/xapi-idl/xen/xenops_types.ml
psafont: I'm using `tpm_type` here because `tpm` is already used for the field for the record.

>  Is there anything other than vtpm?
In theory a hardware TPM could be passed through, in this variant is where more information will be attached so the vTPM can be initialized with the correct information when the contents are stored in the database.

We might have to be careful here to avoid getting stale vtpm data into xenopsd, although usually when building this metadata the domain will not be running.

ocaml/xapi/xapi_xenops.ml
psafont: I should try harded indeed, I'm surprised surrounding code doesn't use the platformdata code to process these fields, will change this here, then other code

lindig (approved):

psafont:
I've tested that xapi now creates and destroys the vtpm objects on a VM's first boot (not on successive boots) and when they are destroyed.

edwintorok (approved):

PR #4696, opened 2022-05-04
Add a fallback system for auth files belonging to RPMs

benjamreis:
This is an implementation in what was proposed in [this comment](https://github.com/xapi-project/xen-api/pull/4689#issuecomment-1103919868).

- Erase all auth files at XAPI startup, `PK` and `dbx` included

- If after the extract of the tarball mandatory auth files are missing:
fetch them from a fallback dir and copy them in the varstore directory

- New XAPI conf entry: `fallback_auth_dir`: Fallback path for auth files installed by RPMs

### DIFF
--- a/ocaml/xapi/xapi_globs.ml
+++ b/ocaml/xapi/xapi_globs.ml
@@ -846,6 +846,8 @@ let varstore_rm = ref "/usr/bin/varstore-rm"
 
 let varstore_dir = ref "/usr/share/varstored"
 
+let fallback_auth_dir = ref "/var/lib/varstored"
+
 let disable_logging_for = ref []
 
 let nvidia_whitelist = ref "/usr/share/nvidia/vgpu/vgpuConfig.xml"
@@ -1501,6 +1503,7 @@ module Resources = struct
       , "Executed to clear certain UEFI variables during clone"
       )
     ; ("varstore_dir", varstore_dir, "Path to local varstored directory")
+    ; ("fallback_auth_dir", fallback_auth_dir, "Fallback path for auth files")
     ; ( "nvidia-sriov-manage"
       , nvidia_sriov_manage_script
       , "Path to NVIDIA sriov-manage script"
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -2591,13 +2591,34 @@ let write_uefi_certificates_to_disk ~__context ~host:_ =
             let path = Filename.concat !Xapi_globs.varstore_dir name in
             Unixext.unlink_safe path
           )
-          ["KEK.auth"; "db.auth"] ;
+          ["PK.auth"; "KEK.auth"; "db.auth"; "dbx.auth"] ;
         (* No uefi certificates, nothing to do. *)
         if contents <> "" then (
           with_temp_file_contents ~contents
             (Tar_unix.Archive.extract extract_certificate_file) ;
           debug "UEFI tar file extracted to temporary directory"
-        )
+        ) ;
+        (* If some mandatory auth files are missing, fetch them from a fallback dir *)
+        List.iter
+          (fun name ->
+            let path = Filename.concat !Xapi_globs.varstore_dir name in
+            let fallback_path =
+              Filename.concat !Xapi_globs.fallback_auth_dir name
+            in
+            if (not (Sys.file_exists path)) && Sys.file_exists fallback_path
+            then (
+              debug "Copy fallback auth %s to varstore dir" name ;
+              let ifd = Unix.openfile fallback_path [Unix.O_RDONLY] 0o0 in
+              let ofd =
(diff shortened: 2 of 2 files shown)


benjamreis:
Note that it'd mean that you install your RPM owned PK and dbx files in `/var/lib/varstored` (the default value can of course change as you wish!) and copy them in `/usr/share/varstored`.

As described by @stormi when talking to @edwintorok: we already install our PK in a different directory (`/usr/share/uefistored`) and copy it in our `varstore_dir` (`/var/lib/uefistored`).

ocaml/xapi/xapi_globs.ml
```diff
@@ -846,6 +846,8 @@ let varstore_rm = ref "/usr/bin/varstore-rm"
 
 let varstore_dir = ref "/usr/share/varstored"
 
+let fallback_auth_dir = ref "/var/lib/varstored"
```
stormi, line 849: How about exchanging the default locations of `varstore_dir` and `fallback_auth_dir` (maybe rename it `varstore_fallback_dir` in order to share a common prefix with `varstore_dir`?).

I know this would temporarily break varstored on Citrix Hypervisor, but it would be easily fixed in `xapi.conf` or `xapi.conf.d` and allow to be consistent right from the start in the code: `varstore_dir`, volatile, in `/var`, and fallback_dir in `/usr`.

ocaml/xapi/xapi_globs.ml
```diff
@@ -1501,6 +1503,10 @@ module Resources = struct
       , "Executed to clear certain UEFI variables during clone"
       )
     ; ("varstore_dir", varstore_dir, "Path to local varstored directory")
+    ; ( "fallback_auth_dir"
+      , fallback_auth_dir
+      , "Fallback path for auth files installed by RPMs"
```
stormi, line 1508: It's not mandatory that the files come from RPMs so I wouldn't mention them here.

ocaml/xapi/xapi_globs.ml
benjamreis: I think it's for @edwintorok to decide.

benjamreis:
The PR has been tested (https://github.com/xcp-ng/xcp-ng-tests/pull/106/files).

After a `Pool.set_uefi_certificates("")` there still is a PK.auth fetched from our fallback directory and so a UEFI VM can be booted.

ocaml/xapi/xapi_host.ml
```diff
@@ -2591,13 +2591,34 @@ let write_uefi_certificates_to_disk ~__context ~host:_ =
...
-        )
+        ) ;
+        (* If some mandatory auth files are missing, fetch them from a fallback dir *)
+        List.iter
+          (fun name ->
+            let path = Filename.concat !Xapi_globs.varstore_dir name in
+            let fallback_path =
+              Filename.concat !Xapi_globs.fallback_auth_dir name
+            in
+            if (not (Sys.file_exists path)) && Sys.file_exists fallback_path
```
lindig, line 2608: Is this sufficient? What about the file at `path` is more recent than the file at `fallback_path`? Would you want to create a new fallback?

benjamreis: This is not what the code do. (Or I misunderstood your comment)
In the case the extract of the XAPI db tarball has no file (file at `path` doesn't exist) the code fetch a fallback file and copy it to `path`.
To update the file in the tarball, an update of varstored (for exemple) install a new file at fallback path & call `Pool.set_uefi_certificate` with a tarball containing the new file.

lindig: Sorry, I misread this. The code is fetching a file if it is missing, not creating the fallback as I had read it.

benjamreis: Exactly, the XAPI does not change/delete the fallback file, only reads it.
To change it it'd come from another process.
IIUC your usecase `PK` would be modified by a new `varstored` RPM and `dbx` by a new `secureboot-certs`, the %post of the RPM would install the new auth files to the fallback directory and the XAPI db would need to be updated (to propagate to the whole pool) with the `Pool.set_uefi_certificates` method with a tarball with the updated auth files. :)

lindig (approved):

benjamreis:
Hi! Any update on this PR? It's my understanding that it respects both our scenarii so it's good to merge AFAICT. :)

lindig:
This is quite a small PR; so have a look @robhoes @edwintorok @psafont

psafont (approved):

psafont (approved):
Looking at the previous PR, this one looks like it avoid the previous pitfall

ocaml/xapi/xapi_globs.ml
stormi: @psafont @edwintorok What do you think about my comment here? (for a possible future PR)

PR #4697, opened 2022-05-04
xsh: fix XAPI blob sync in pools

edwintorok:
xsh was closing the file descriptor both in the 'proxy' function and in
the stunnel wrapper code.
Push down the Unixfd.(!) dereference into the proxy code to make it
obvious that we have a Unix.close there, and replace that with
Unixfd.safe_close which avoids double close.

This was causing blob sync to fail in XAPI with an rsync protocol error
(due to the extra EBADF printed by xsh).

Fixes: c6511f0a48e1c058dd43bb02fc9503be0c1aca42 "CA-337546: update to new Stunnel_cache API with Safe_resources support"

TODO: needs a CA ticket and some testing, so far I just confirmed that the EBADF error is gone, need to check whether blob sync works too.

### DIFF
--- a/ocaml/xsh/xsh.ml
+++ b/ocaml/xsh/xsh.ml
@@ -22,12 +22,14 @@ type endpoint = {
 let make_endpoint fdin fdout =
   {fdin; fdout; buffer= Bytes.make 4096 '\000'; buffer_len= 0}
 
-let proxy (ain : Unix.file_descr) (aout : Unix.file_descr)
-    (bin : Unix.file_descr) (bout : Unix.file_descr) =
-  let a' = make_endpoint ain aout and b' = make_endpoint bin bout in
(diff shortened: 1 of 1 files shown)


psafont (approved):
I can see how this got missed, would it be worth setting up a quality gate that counts the Unix.close in the project?

edwintorok:
Now that almost the entire toolstack is in one repo we might be able to convert some more code over to using Unixfd (although to make that easier might be useful to move Unixfd into XAPI itself: it is probably better to wrap raw FD operations into Unixfd module and expose only Unixfd.t operations).
Not sure the quality gate would help: there are 140 Unix.close calls.

lindig (approved):

robhoes:
Looks like this needs a reformat.

PR #4698, opened 2022-05-05
maintenance: run uuid tests as part of that package

psafont:
This causes a failure when compiling wsproxy in an opam environment as the test binary is built for all the opam packages in the repo and wsproxy lacks the alcotest dependency when testing.

### DIFF
--- a/ocaml/libs/uuid/dune
+++ b/ocaml/libs/uuid/dune
@@ -9,6 +9,7 @@
 
 (test
   (name uuid_test)
+  (package uuid)
   (modules uuid_test)
   (libraries uuid alcotest)
   )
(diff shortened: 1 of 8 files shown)


lindig (approved):

ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -5,6 +5,8 @@ let prototyped_of_class = function
 let prototyped_of_field = function
   | "Repository", "gpgkey_path" ->
       Some "22.12.0"
+  | "pool", "uefi_certificates" ->
+      Some "22.18.0-next"
```
robhoes, line 9: This field hasn't been defined quite right in the datamodel... It is not a new field, but it has changed recently. Please remove this diff and I'll fix it separately.

robhoes (approved):

PR #4699, opened 2022-05-06
Update to mtime 1.4 interface

psafont:
Mtime 1.4 deprecates the constant to convert from seconds to nanoseconds and viceverse.

I've used to opportunity to use the new interface for defining spans and remove quite a few periods defined in seconds, now they should be easier to recognized in the timeouts that happen.

Most of the code uses Mtime and counters to set the point of reference. For AD this is not possible as the state is persisted, so Ptime is used.

I wanted to gather some comments while I test: for example I don't know if changing the xapi globals to ints from floats breaks compatibility or whether I missed some shared frame of reference which would cause failures.

The part I like the least is the use of compare which is not instant to recognize what it means.

The CI will fail until Mtime is updated in xs-opam (and once it's updated it will make the CI fail xen-api until this PR or similar is merged)

### DIFF
--- a/ocaml/xapi-client/tasks.ml
+++ b/ocaml/xapi-client/tasks.ml
@@ -30,7 +30,7 @@ let wait_for_all_inner ~rpc ~session_id ~all_timeout ~tasks =
   let timeout_span =
     match all_timeout with
     | Some t ->
-        Some (t *. Mtime.s_to_ns |> Int64.of_float |> Mtime.Span.of_uint64_ns)
+        Some Mtime.Span.(Float.to_int (t *. 1e9) * ns)
     | None ->
         None
   in
--- a/ocaml/xapi/xapi_event.ml
+++ b/ocaml/xapi/xapi_event.ml
@@ -21,6 +21,9 @@ module D = Debug.Make (struct let name = "xapi_event" end)
 
 open D
 
+let timeout_not_reached elapsed timeout =
+  Mtime.Span.(compare (Mtime_clock.count elapsed) timeout) < 0
+
 module Message = struct
   type t =
     | Create of (API.ref_message * API.message_t)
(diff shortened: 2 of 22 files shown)


ocaml/xapi-client/tasks.ml
```diff
@@ -30,7 +30,7 @@ let wait_for_all_inner ~rpc ~session_id ~all_timeout ~tasks =
   let timeout_span =
     match all_timeout with
     | Some t ->
-        Some (t *. Mtime.s_to_ns |> Int64.of_float |> Mtime.Span.of_uint64_ns)
+        Some Mtime.Span.(Float.to_int (t *. 1e9) * ns)
```
lindig, line 33: Is this more complicated than necessary? `t` is in seconds, so this could be `Span.(t * s)`

ocaml/xapi/xapi_event.ml
```diff
@@ -714,14 +717,17 @@ let from ~__context ~classes ~token ~timeout =
         )
   in
   let subs = List.map Subscription.of_string classes in
-  let deadline = Unix.gettimeofday () +. timeout in
+  let elapsed = Mtime_clock.counter () in
+  let timeout = Mtime.Span.(Float.to_int (timeout *. 1e9) * ns) in
```
lindig, line 721: Again, why not `Span.(timeout * s)`?

ocaml/xapi/xapi_globs.ml
```diff
@@ -681,13 +681,13 @@ let ha_xapi_restart_timeout = ref 300
    logrotate when it exceeds the threshold *)
 let logrotate_check_interval = ref 300.
 
-let rrd_backup_interval = ref 86400.
+let rrd_backup_interval = ref 86400
 
 (* CP-703: Periodic revalidation of externally-authenticated sessions *)
-let session_revalidation_interval = ref 300. (* every 5 minutes *)
+let session_revalidation_interval = ref 300 (* every 5 minutes *)
```
lindig, line 687: The reason that we can't express these in spans is that these refs are updated by command-line options?

ocaml/xapi/xapi_host.ml
```diff
@@ -820,21 +820,23 @@ let is_slave ~__context ~host:_ = not (Pool_role.is_master ())
...
-    let timeout = 10. in
+    let timeout = Mtime.Span.(10 * s) in
     let task_name = Context.get_task_id __context |> Ref.string_of in
     let ip, uuid =
       ( Db.Host.get_address ~__context ~self:host
       , Db.Host.get_uuid ~__context ~self:host
       )
     in
     let rec log_host_slow_to_respond timeout () =
+      let minimum a b = if Mtime.Span.compare a b < 0 then b else a in
```
lindig, line 831: I believe this is wrong. if `compare a b` < 0 then a < b, hence a is the minimum.

lindig (changes requested):

ocaml/xapi-client/tasks.ml
psafont: t is a float, not an integer, so the easy conversion is not available. If, for some reason, the timeout is more finegrained than seconds we need to do the complicated conversion

ocaml/xapi/xapi_globs.ml
psafont: Indeed, I'm not sure these bindings can be changed to integers without breaking anything.

In theory we could deserialize Spans in the form of an integer followed by the unit of time.

ocaml/xapi/xapi_host.ml
psafont: Yeah, we want to return a unless b is a larger span of time.

psafont:
I'll do a small patch to fix the deprecation warnings, while this is a nice change to have it introduces more churn than we're comfortable with

PR #4704, opened 2022-05-16
Missing net new RPMs in picking up metadata from updateinfo

minglumlu:
Following fixes and changes are included in this series:
1. Missing net new RPMs in picking up metadata from updateinfo
2. Add un-installed packages into accumulative update list
3. Add debug logs for outputs of YUM/RPM command lines

### DIFF
--- a/ocaml/xapi/repository_helpers.ml
+++ b/ocaml/xapi/repository_helpers.ml
@@ -459,21 +459,61 @@ let parse_updateinfo_list acc line =
       debug "Ignore unrecognized line '%s' in parsing updateinfo list" line ;
       acc
 
-let get_updates_from_updateinfo ~__context repositories =
-  let params_of_updateinfo_list =
+let is_obsoleted pkg_name repositories =
+  let params =
+    [
+      "-a"
+    ; "--plugins"
+    ; "--disablerepo=*"
+    ; Printf.sprintf "--enablerepo=%s" (String.concat "," repositories)
+    ; "--whatobsoletes"
+    ; pkg_name
+    ; "--qf"
+    ; "%{name}"
+    ]
+  in
+  match
+    Helpers.call_script !Xapi_globs.repoquery_cmd params
+    |> Astring.String.cuts ~sep:"\n" ~empty:false
+  with
+  | [] ->
+      false
+  | _ ->
+      debug "Available package %s has been obsoleted." pkg_name ;
+      true
+  | exception _ ->
+      warn
+        "Can't determine if available package %s is obsoleted or not. Assuming \
+         it is not."
+        pkg_name ;
+      false
+
+let get_pkgs_from_yum_updateinfo_list sub_command repositories =
+  let params =
     [
       "-q"
     ; "--disablerepo=*"
     ; Printf.sprintf "--enablerepo=%s" (String.concat "," repositories)
     ; "updateinfo"
     ; "list"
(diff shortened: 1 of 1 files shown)


ocaml/xapi/repository_helpers.ml
```diff
@@ -459,22 +459,70 @@ let parse_updateinfo_list acc line =
...
+    ; "--whatobsoletes"
+    ; pkg_name
+    ; "--qf"
+    ; "%{name}"
+    ]
+  in
+  match
+    Helpers.call_script !Xapi_globs.repoquery_cmd params
+    |> Astring.String.cuts ~sep:"\n"
+    |> List.filter (fun x -> not (Astring.String.is_empty x))
```
psafont, line 478:
```suggestion
    |> Astring.String.cuts ~sep:"\n" ~empty:false
```

ocaml/xapi/repository_helpers.ml
```diff
@@ -459,22 +459,70 @@ let parse_updateinfo_list acc line =
...
+      debug "Available package %s has been obsoleted." pkg_name ;
+      true
+  | exception _ ->
+      warn
+        "Can't determine if available package %s is obsoleted or not. Assuming \
+         it is not."
+        pkg_name ;
+      false
+
+let run_yum_updateinfo_list sub_command enabled_repos =
```
lindig, line 492: This name suggests the result is a side effect but the actual result is computed with a `fold_left`; I would suggest to rename this to something that better reflects the result.

ocaml/xapi/repository_helpers.ml
```diff
@@ -459,22 +459,70 @@ let parse_updateinfo_list acc line =
...
+  Helpers.call_script !Xapi_globs.yum_cmd params
   |> assert_yum_error
   |> Astring.String.cuts ~sep:"\n"
+  |> List.map (fun x ->
+         debug "yum updateinfo list %s: %s" sub_command x ;
+         x
+     )
   |> List.fold_left parse_updateinfo_list []
 
+let get_updates_from_updateinfo ~__context repositories =
```
lindig, line 512: Would this be better expressed with strings sets? `module StringSet = Set.Make(String)`

ocaml/xapi/repository_helpers.ml
```diff
@@ -600,27 +648,49 @@ let get_installed_pkgs () =
...
   |> Astring.String.cuts ~sep:"\n"
+  |> List.map (fun x ->
+         debug "repoquery available: %s" x ;
+         x
+     )
   |> List.fold_left parse_line_of_repoquery []
 
+let get_updates_from_repoquery repositories =
+  List.iter (fun r -> clean_yum_cache r) repositories ;
+  let enabled_repos = String.concat "," repositories in
```
lindig, line 681: I think the naming is a bit redundant. Why do we need `enabled_repos` when there are not other repos in this code? Again, it might be better to use sets to express this logic.

ocaml/xapi/repository_helpers.ml
minglumlu: The `repositories` comes from query on XAPI DB. So I think it should keep  the uniqueness of elements well. And usually there are only two repositories in this list.

lindig (approved):

robhoes (approved):
Looks good. Please squash the fixup before merging.

PR #4705, opened 2022-05-16
Filter input dns when reconfiguring a pif IP(v6)

benjamreis:
Instead of throwing an exception when a DNS doesn't have
the correct ip familly: just ignore it

It means when reconfiguring:
- an ipv4: only the ipv4 entries on the input dns will kept and
only the ipv6 dns already set to the pif will be kept
- an ipv6: only the ipv6 entries on the input dns will kept and
only the ipv4 dns already set to the pif will be kept

This is usefull in the case both family are configured on a pif

Right now the installer and xsconsole can fail when reconfiguring an
IP or an IPv6 and both family DNS are present, which can often happen
in case of dual stack

### DIFF
--- a/ocaml/xapi/xapi_pif.ml
+++ b/ocaml/xapi/xapi_pif.ml
@@ -728,6 +728,12 @@ let destroy ~__context ~self =
       Client.Client.VLAN.destroy ~rpc ~session_id ~self:vlan
   )
 
+let restrict_to ~domain dns =
+  Astring.String.cuts ~sep:"," ~empty:false dns
+  |> List.filter (fun addr ->
+         Xapi_stdext_unix.Unixext.domain_of_addr addr = Some domain
+     )
+
 let reconfigure_ipv6 ~__context ~self ~mode ~iPv6 ~gateway ~dNS =
   Xapi_pif_helpers.assert_pif_is_managed ~__context ~self ;
   assert_no_protection_enabled ~__context ~self ;
@@ -737,10 +743,8 @@ let reconfigure_ipv6 ~__context ~self ~mode ~iPv6 ~gateway ~dNS =
   (* If we have an IPv6 address, check that it is valid and a prefix length is specified *)
   if iPv6 <> "" then
     Helpers.assert_is_valid_cidr `ipv6 "IPv6" iPv6 ;
-  if dNS <> "" then
-    List.iter
-      (fun address -> Helpers.assert_is_valid_ip `ipv6 "DNS" address)
-      (String.split ',' dNS) ;
+  (* Only set IPv6 dNS *)
+  let ipv6_dNS = restrict_to ~domain:Unix.PF_INET6 dNS in
   (* Management iface must have an address for the primary address type *)
   let management = Db.PIF.get_management ~__context ~self in
   let primary_address_type = Db.PIF.get_primary_address_type ~__context ~self in
@@ -753,15 +757,12 @@ let reconfigure_ipv6 ~__context ~self ~mode ~iPv6 ~gateway ~dNS =
   Db.PIF.set_ipv6_configuration_mode ~__context ~self ~value:mode ;
   Db.PIF.set_ipv6_gateway ~__context ~self ~value:gateway ;
   Db.PIF.set_IPv6 ~__context ~self ~value:[iPv6] ;
-  (* Keep IPv4 DNS entries *)
-  let pif_dns =
-    List.filter
-      (fun dns_addr ->
-        Xapi_stdext_unix.Unixext.domain_of_addr dns_addr = Some Unix.PF_INET
-      )
-      (String.split_on_char ',' (Db.PIF.get_DNS ~__context ~self))
+  (* Only keep existing IPv4 entries *)
+  let pif_ipv4_dNS =
+    restrict_to ~domain:Unix.PF_INET (Db.PIF.get_DNS ~__context ~self)
   in
-  Db.PIF.set_DNS ~__context ~self ~value:(String.concat "," (dNS :: pif_dns)) ;
+  Db.PIF.set_DNS ~__context ~self
+    ~value:(String.concat "," (List.concat [ipv6_dNS; pif_ipv4_dNS])) ;
   if Db.PIF.get_currently_attached ~__context ~self then (
     debug
       "PIF %s is currently_attached and the configuration has changed; calling \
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_pif.ml
```diff
@@ -737,10 +737,14 @@ let reconfigure_ipv6 ~__context ~self ~mode ~iPv6 ~gateway ~dNS =
...
-    List.iter
-      (fun address -> Helpers.assert_is_valid_ip `ipv6 "DNS" address)
-      (String.split ',' dNS) ;
+  (* Only set IPv6 dNS *)
+  let ipv6_dns_to_set =
+    List.filter
+      (fun dns_addr ->
+        Xapi_stdext_unix.Unixext.domain_of_addr dns_addr = Some Unix.PF_INET6
+      )
+      (String.split_on_char ',' dNS)
```
psafont, line 746: Using `Astring.String.cuts ~sep:',' ~empty:false` would allow to match against empty lists, without worrying about lists with empty elements further down.

It would be better dNS would only be processed once instead of twice. An even further enhancement would be to use `let ipv4_dns, ipv6_set = List.partition (fun dns_addr -> ...)`, but this only works if there are 2 types of addresses here

benjamreis: `dNS` is only processed once the 2nd time is the `PIF.DNS` field.

ocaml/xapi/xapi_pif.ml
```diff
@@ -753,15 +757,28 @@ let reconfigure_ipv6 ~__context ~self ~mode ~iPv6 ~gateway ~dNS =
...
+  let new_dns =
+    match (ipv6_dns_to_set, pif_ipv4_dns) with
+    | [], [] ->
+        ""
+    | _, [] ->
+        String.concat "," ipv6_dns_to_set
+    | [], _ ->
+        String.concat "," pif_ipv4_dns
+    | _, _ ->
+        String.concat "," (String.concat "," ipv6_dns_to_set :: pif_ipv4_dns)
```
psafont, line 779: Both `ipv6_dns_to_set` and `pif_ipv4_dns` are lists,  `::` adds a single element to a list. Instead use `List.concat [ipv6; ipv4]` to concatenate the lists.

benjamreis:
Tested and validated:
`xe pif-reconfigure-ip mode=static uuid=... IP=... gateway=... DNS=ipv4,ipv6` only sets the ipv4 and same but ipv6 for `pif-reconfigure-ipv6`

And the method only override the DNS of the same family address.

ocaml/xapi/xapi_pif.ml
```diff
@@ -753,15 +757,28 @@ let reconfigure_ipv6 ~__context ~self ~mode ~iPv6 ~gateway ~dNS =
...
+  let new_dns =
+    match (ipv6_dns_to_set, pif_ipv4_dns) with
+    | [], [] ->
+        ""
+    | _, [] ->
+        String.concat "," ipv6_dns_to_set
+    | [], _ ->
+        String.concat "," pif_ipv4_dns
+    | _, _ ->
+        String.concat "," (List.concat [ipv6_dns_to_set; pif_ipv4_dns])
```
psafont, line 779: I believe this is equivalent to:
```suggestion
    let new_dns =
      String.concat "," (List.concat [ipv6_dns_to_set; pif_ipv4_dns])
    in
```

ocaml/xapi/xapi_pif.ml
```diff
@@ -753,15 +757,28 @@ let reconfigure_ipv6 ~__context ~self ~mode ~iPv6 ~gateway ~dNS =
   Db.PIF.set_ipv6_configuration_mode ~__context ~self ~value:mode ;
   Db.PIF.set_ipv6_gateway ~__context ~self ~value:gateway ;
   Db.PIF.set_IPv6 ~__context ~self ~value:[iPv6] ;
-  (* Keep IPv4 DNS entries *)
-  let pif_dns =
+  (* Only keep existing IPv4 entries *)
+  let pif_ipv4_dns =
```
lindig, line 761: A different name for a similar purpose here.

ocaml/xapi/xapi_pif.ml
```diff
@@ -809,15 +830,28 @@ let reconfigure_ip ~__context ~self ~mode ~iP ~netmask ~gateway ~dNS =
...
   Db.PIF.set_gateway ~__context ~self ~value:gateway ;
-  (* Keep IPv6 DNS entries *)
-  let pif_dns =
+  (* Only keep existing IPv4 entries *)
+  let pif_ipv6_dns =
     List.filter
       (fun dns_addr ->
         Xapi_stdext_unix.Unixext.domain_of_addr dns_addr = Some Unix.PF_INET6
       )
-      (String.split_on_char ',' (Db.PIF.get_DNS ~__context ~self))
```
lindig, line 818: Quite a bit of code repetition.

ocaml/xapi/xapi_pif.ml
benjamreis: Indeed! Thx!

ocaml/xapi/xapi_pif.ml
```diff
@@ -737,10 +737,14 @@ let reconfigure_ipv6 ~__context ~self ~mode ~iPv6 ~gateway ~dNS =
   (* If we have an IPv6 address, check that it is valid and a prefix length is specified *)
   if iPv6 <> "" then
     Helpers.assert_is_valid_cidr `ipv6 "IPv6" iPv6 ;
-  if dNS <> "" then
-    List.iter
-      (fun address -> Helpers.assert_is_valid_ip `ipv6 "DNS" address)
-      (String.split ',' dNS) ;
+  (* Only set IPv6 dNS *)
+  let ipv6_dns_to_set =
```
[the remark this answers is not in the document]
benjamreis: the filter to mutualize the code is great. Thanks :)

ocaml/xapi/xapi_pif.ml
benjamreis: what do you think about:
`<family>_dNS` for the DNS inputed to the reconfigure function and `pif_<family>_dNS` for the DNS already present in the PIF DNS field.
Where family is either `ipv6` or `ipv4`

ocaml/xapi/xapi_pif.ml
benjamreis: I mutualized with your helper :)

lindig (approved):

ocaml/xapi/xapi_pif.ml
```diff
@@ -809,15 +808,12 @@ let reconfigure_ip ~__context ~self ~mode ~iP ~netmask ~gateway ~dNS =
...
   Db.PIF.set_netmask ~__context ~self ~value:netmask ;
   Db.PIF.set_gateway ~__context ~self ~value:gateway ;
-  (* Keep IPv6 DNS entries *)
-  let pif_dns =
-    List.filter
-      (fun dns_addr ->
-        Xapi_stdext_unix.Unixext.domain_of_addr dns_addr = Some Unix.PF_INET6
-      )
-      (String.split_on_char ',' (Db.PIF.get_DNS ~__context ~self))
+  (* Only keep existing IPv4 entries *)
```
robhoes, line 811: IPv6?

benjamreis: Indeed ^^'

benjamreis: Fixed

robhoes (approved):

PR #4706, opened 2022-05-16
CP-38688 introduce Message.destroy_many() API/CLI call

lindig:
The RPC overhead when destroying many messages is high. To avoid it,
introcuce a message to destroy several messages in one call. This
introduces a new API call Message.destroy_many. The corresponding CLI
command message-destroy is extended to accept a parameter "uuids" to
accept multiple UUIDs in additio to the existing "uuid" parameter.

I notice that destroyig a message by itself is not cheap because it
parses the XML content of the message being destroyed. In an unlucky
case it also involves scanning a directory on the file system to remove
an alias.

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -6281,6 +6281,11 @@ module Message = struct
         [(Ref _message, "self", "The reference of the message to destroy")]
       ~flags:[`Session] ~allowed_roles:_R_POOL_OP ()
 
+  let destroy_many =
+    call ~name:"destroy_many" ~in_product_since:rel_next
+      ~params:[(Set (Ref _message), "messages", "Messages to destroy")]
+      ~flags:[`Session] ~allowed_roles:_R_POOL_OP ()
+
   let get_all =
     call ~name:"get_all" ~in_product_since:rel_orlando ~params:[]
       ~flags:[`Session]
@@ -6343,6 +6348,7 @@ module Message = struct
         [
           create
         ; destroy
+        ; destroy_many
         ; get
         ; get_all
         ; get_since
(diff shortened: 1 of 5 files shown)


ocaml/idl/datamodel.ml
```diff
@@ -6281,6 +6281,11 @@ module Message = struct
         [(Ref _message, "self", "The reference of the message to destroy")]
       ~flags:[`Session] ~allowed_roles:_R_POOL_OP ()
 
+  let destroy_many =
+    call ~name:"destroy_many" ~in_product_since:rel_next
```
robhoes, line 6285: Please omit `~in_product_since:rel_next`.

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -1315,9 +1315,25 @@ let message_create (_ : printer) rpc session_id params =
...
-  Client.Message.destroy ~rpc ~session_id ~self:message
+  let fail msg = raise (Cli_util.Cli_failure msg) in
+  let uuid = List.assoc_opt "uuid" params in
+  let uuids = List.assoc_opt "uuids" params in
+  let uuids =
+    match (uuid, uuids) with
+    | Some uuid, None ->
+        [uuid]
+    | None, Some uuids ->
+        String.split_on_char ',' uuids
```
psafont, line 1326: Is comma used for inputting lists elsewhere in the `CLI`?

lindig: Yes

ocaml/idl/datamodel.ml
lindig: This causes:

robhoes: Instead, add `~lifecycle:[]`.

psafont (approved):
Other than the comment about the separator, this looks good. I wish there was a way to enforce either of the parameters in a data-centric way but it looks like the way it's done it's the next-best thing.

Does this need some change in the bash autocompletion script? It could be difficult but it might be worth to autocomplete to `uuids` when tabbing

PR #4707, opened 2022-05-16
CP-39136: Reintroduce backend to vtpms 

psafont:
This type it's for signifying what persistence backend it's being used.
This means its type from a vm ref to an enum, for now with a single
possible value. There's no way to change yet backend, I would be wary of
adding them, as this is intended to permit several persistence backends
at the same time only for the benefit of developers.

### DIFF
--- a/ocaml/idl/datamodel_vtpm.ml
+++ b/ocaml/idl/datamodel_vtpm.ml
@@ -16,6 +16,9 @@ open Datamodel_types
 open Datamodel_common
 open Datamodel_roles
 
+let persistence_backend =
+  Enum ("persistence_backend", [("xapi", "This VTPM is persisted in XAPI's DB")])
+
 let get_contents =
   call ~name:"get_contents" ~in_product_since:"rel_next"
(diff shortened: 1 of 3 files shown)


edwintorok (commented):
Have you checked that upgrades from the old DB work? Last I tried I got some errors from XAPI about missing backend field from the DB. Maybe it'd be better to leave the old `backend` field name alone (delete it if we have to) and add a `persistence_backend` field (to match the persistence_backend type), that way perhaps we'd have less issues during a DB upgrade. I think the schema minor vsn has to be bumped too, that is probably the upgrade wasn't working: without bumping the minor vsn XAPI won't know that it has to upgrade the DB (although perhaps we can improve that: we have a schema hash now, so XAPI could also check the schema hash of the DB it loads vs the one it got compiled with, and if it doens't match then run the DB upgrade anyway?)

lindig (approved):

edwintorok (changes requested):
I think this requires a bump to the schema minor vsn to avoid problems with DB upgrades.

psafont:
Will integrate this change into another PR

PR #4709, opened 2022-05-16
XSI-791/CA-343760: Backport to yangtze

robhoes:


psafont (approved):
Along with the xenopsd and xcp-idl commits, it looks like a faithful commit, although these are a lot of functional changes and will need thorough testing

lindig (approved):

PR #4713, opened 2022-05-17
xapi: avoid spawning processes to touch files

psafont:
The check against an empty string needs to be kept as some of the
variables use an empty string as an alternative to None, meaning not to
create the file at all.

### DIFF
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -1219,16 +1219,13 @@ let local_storage_exists () =
     true
   with _ -> false
 
-let touch_file fname =
-  try
-    if fname <> "" then
-      match Unixext.spawnvp "touch" [|"touch"; fname|] with
-      | Unix.WEXITED 0 ->
-          ()
-      | _ ->
-          warn "Unable to touch ready file '%s': touch exited abnormally" fname
-  with e ->
-    warn "Unable to touch ready file '%s': %s" fname (Printexc.to_string e)
+(** Stdlib's Arg module doesn't support string option ref, instead the empty
+    string is used for expressing the null value *)
+let touch_file cmd_arg =
+  if cmd_arg <> "" then
+    try Unixext.touch_file cmd_arg
+    with e ->
+      warn "Unable to touch file '%s': %s" cmd_arg (Printexc.to_string e)
 
 let vm_to_string __context vm =
   let str = Ref.string_of vm in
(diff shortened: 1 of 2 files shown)


ocaml/xapi/xapi_host.ml
```diff
@@ -1103,8 +1103,12 @@ let syslog_reconfigure ~__context ~host:_ =
   let flag =
     match destination with "" -> "--noremote" | _ -> "--remote=" ^ destination
   in
-  let args = [|!Xapi_globs.xe_syslog_reconfigure; flag|] in
-  ignore (Unixext.spawnvp args.(0) args)
+  let _ =
```
edwintorok, line 1106: Usually it is good to specify the type:
```
let (_: string*string) = ...
```

That way if forkhelpers.execute_command_get_output gains more parameters this won't silently stop working, although in its current form it is correct.

ocaml/xapi/helpers.ml
```diff
@@ -1220,15 +1220,9 @@ let local_storage_exists () =
...
-  try
-    if fname <> "" then
-      match Unixext.spawnvp "touch" [|"touch"; fname|] with
-      | Unix.WEXITED 0 ->
-          ()
-      | _ ->
-          warn "Unable to touch ready file '%s': touch exited abnormally" fname
-  with e ->
-    warn "Unable to touch ready file '%s': %s" fname (Printexc.to_string e)
+  if fname <> "" then
```
lindig, line 1223: What is the point of this test? I would rather see the log entry that this can't be touched than nothing.

lindig (approved):

ocaml/xapi/helpers.ml
psafont: This function gets used on values set by the configuration. These values use the convention of using empty to mean None, as in "please don't try to create the file". I think adding an error to the logs would be confusing in that case (which is the default for some values)

lindig: So we keep it here but this sounds to me like something that should be handled on the calling side and not here or at the very least needs to be made explicit with an option.

psafont: Stdlib's Arg module doesn't support using `string option ref`: it forces to process string parameters with the empty string being the null value. I've made it obvious this function processes args, but I don't think we can do much more short of changing the framework for parsing args

lindig: Fair enough. `Arg` is convenient but I would favour `Cmdliner` for new code because it is more compositional and has richer functionality.

PR #4718, opened 2022-05-19
CP-38583 add Host.last_software_update field with data/time

lindig:
Add a new field that reflects the last time Host.apply_software_updates
was run (or the time of the first installation).

### DIFF
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -2071,6 +2071,10 @@ let t =
         ; field ~qualifier:DynamicRO ~in_product_since:"1.313.0" ~ty:Bool
             "tls_verification_enabled" ~default_value:(Some (VBool false))
             "True if this host has TLS verifcation enabled"
+        ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:DateTime
+            "last_software_update"
+            ~default_value:(Some (VDateTime (Date.of_float 0.0)))
+            "Date and time when the last software update was applied"
         ]
       )
     ()
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -950,7 +950,7 @@ let create ~__context ~uuid ~name_label ~name_description:_ ~hostname ~address
       )
     ~control_domain:Ref.null ~updates_requiring_reboot:[] ~iscsi_iqn:""
     ~multipathing:false ~uefi_certificates:"" ~editions:[] ~pending_guidances:[]
-    ~tls_verification_enabled ;
+    ~tls_verification_enabled ~last_software_update:(Date.localtime ()) ;
   (* If the host we're creating is us, make sure its set to live *)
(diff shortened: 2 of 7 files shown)


ocaml/xapi/xapi_host.ml
```diff
@@ -2810,4 +2814,5 @@ let apply_updates ~__context ~self ~hash =
       ~doc:"Host.apply_updates" ~op:`apply_updates
     @@ fun () -> Repository.apply_updates ~__context ~host:self ~hash
   in
+  Db.Host.set_last_software_update ~__context ~self ~value:(now_rfc3339 ()) ;
```
psafont, line 2817: Won't this update the field for the coordinator only? this needs to happen for supporters as well

lindig: My understanding is this updates the entry for any host (not just the coordinator) on the coordinator.

psafont: y confusion is due to the name of the host parameter which is called `~self`, this can be different from the host where the function is executed

lindig:
This PR adds a new field. Should this PR include an update to `ocaml/idl/datamodel_lifecycle.ml` or not? I'm a bit confused about this. Including this would make the PR dependent on the current xapi tag.
```
--- a/ocaml/idl/datamodel_lifecycle.ml
+++ b/ocaml/idl/datamodel_lifecycle.ml
@@ -5,8 +5,6 @@ let prototyped_of_class = function
       Some "22.12.0"
-  | "host", "last_software_update" ->
-      Some "22.18.0-next"
   | _ ->
       None
```

psafont:
I don't think the lifecycle change is quite right since this change is done on top of 22.19.0. I think the change of the field should reflect it has been introduced in 22.20.0 as 2.19 does _not_ include it. So I think this PR should _not_ contain this change

psafont (approved):

robhoes:
Regarding the lifecycle change, you got `Some "22.18.0-next"` because the tag v22.19.0 tag was added after you did your build. Now it should be `Some "22.19.0-next"`, which means that it will be set to 22.20.0 _after_ that release has been tagged.

ocaml/idl/datamodel_host.ml
```diff
@@ -2071,6 +2071,9 @@ let t =
         ; field ~qualifier:DynamicRO ~in_product_since:"1.313.0" ~ty:Bool
             "tls_verification_enabled" ~default_value:(Some (VBool false))
             "True if this host has TLS verifcation enabled"
+        ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:String
```
robhoes, line 2074: Now that this is a new field, you can use the `DateTime` type rather than `String`.

lindig:
Several changes in response to comments:
* changed data type of the host field to `DateTime`
* updated hash that identifies IDL version
* updated and simplified code to set this field
* include lifetime annotation (based on latest tag)

robhoes (approved):

robhoes (approved):

PR #4719, opened 2022-05-23
CP-39884 generalise interface to gzip/zstd-like tools

lindig:
The existing interface to compressions tools assumes that all tools
receive the exact same command-line arguments. This patch removes this
limitiation while keeping the existing comamnd-line arguments as
defaults. No client code needs to be changed.

The xapi_compression.mli interace is extended by accepting additional
(and optional) flags (or other arguments) that are passed to the
compression binary. For example, the compression level could be passed
in addition.

### DIFF
--- a/ocaml/libs/xapi-compression/xapi_compression.ml
+++ b/ocaml/libs/xapi-compression/xapi_compression.ml
@@ -1,5 +1,25 @@
 module type ALGORITHM = sig
   val executable : string
+
+  val compress_options : string list
+
+  val decompress_options : string list
+end
+
+module type COMPRESSOR = sig
+  val available : unit -> bool
+  (** Returns whether this compression algorithm is available *)
+
+  val compress : Unix.file_descr -> (Unix.file_descr -> 'a) -> 'a
+  (** Runs a compression process which is fed from a pipe whose entrance
+  is passed to 'f' and whose output is 'ofd' *)
+
+  val decompress : Unix.file_descr -> (Unix.file_descr -> 'a) -> 'a
+  (** Runs a decompression process which is fed from a pipe whose
+  entrance is passed to 'f' and whose output is 'ofd' *)
+
+  (* Experimental decompressor which is fed from an fd and writes to a pipe *)
+  val decompress_passive : Unix.file_descr -> (Unix.file_descr -> 'a) -> 'a
 end
 
 module D = Debug.Make (struct let name = "xapi_compression" end)
@@ -46,9 +66,9 @@ module Make (Algorithm : ALGORITHM) = struct
     let args =
       match mode with
       | Compress ->
-          []
+          Algorithm.compress_options
       | Decompress ->
-          ["--decompress"; "--stdout"]
+          Algorithm.decompress_options
     in
     let stdin, stdout, close_now, close_later =
       match input with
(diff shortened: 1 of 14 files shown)


ocaml/libs/xapi-compression/xapi_compression.ml
```diff
@@ -45,10 +51,10 @@ module Make (Algorithm : ALGORITHM) = struct
     Unixfd.with_pipe ~loc:__LOC__ () @@ fun zcat_out zcat_in ->
     let args =
       match mode with
-      | Compress ->
-          []
-      | Decompress ->
-          ["--decompress"; "--stdout"]
+      | Compress args ->
+          List.concat [Algorithm.compress_options; args]
```
edwintorok, line 55: The args here would probably have to be mapped to the underlying compression algo somehow, each tool has its own args, perhaps we could have a variant here and each implementation can map those variants to string args?
There are at least 2 uses of compression so far:
* compress once, use multiple times (.xva), spending  bit more time on compression could be worth it if you save on bandwidth and make downloads faster
* compress/decompress once, during live migrate: balance CPU usage with network bandwidth

So perhaps to start 2 variants:
`type compression_option = BalanceRatio | BalanceTransferSpeed`

For decompression we have no options yet so that could be just the `unit` type for now.
Then gzip and zstd could map these.
E.g. zstd could map it like this:
```
function
| BalanceTransferSpeed -> ["--adapt=min=1,max=3"]
| BalanceRatio -> [] (* we could specify some higher level here, but lets leave as is for now to avoid making it too slow *)
```

And gzip:
```
function
| BalanceTransferSpeed ->  ["--fast"]
| BalanceRatio -> [] (* could specify --best but gzip is too slow already... *)
```

edwintorok (commented):
Looks good in general, the `string list` from the common code could further be eliminated in a separate commit, see my comment.

ocaml/libs/xapi-compression/xapi_compression.ml
lindig: The `Algorithm.compress_options` and `Algorithm.decompress_options` cover the default per algorithm. It is a remaining limitation that any additional options are just added to these and you can't replace the default. I am reluctant to add more abstractions at this point before it is clear what actual use cases will look like.

psafont:
I'd like to get the logic on arguments to be as contained as possible, which makes me hesitant to add optional parameters on each action. Having this allows the logic to set parameters to sprawl.

Could we remove the optional args and instead expose more than a module in gzip.mli? Each module would have different options when comrpessing and decompressing, for example:

```ocaml

(** Path to the gzip binary *)
let exec = "/bin/gzip"

module Normal = Xapi_compression.Make (struct
  let executable = exec

  let compress_options = []

  let decompress_options = ["--decompress"; "--stdout"]
end)

module High_CPU = Xapi_compression.Make (struct
  let executable = exec

  let compress_options = ["--compress-level";  "high"]

  let decompress_options = ["--decompress"; "--stdout"]
end)

module Low_CPU = Xapi_compression.Make (struct
  let executable = exec

  let compress_options = ["--compress-level";  "low"]

  let decompress_options = ["--decompress"; "--stdout"]
end)

include Normal
```

lindig:
I like the idea to support use cases by defining sub-modules and by doing so keeping all relevant parameters closely together.

lindig:
I revised the design in the direction suggested by @psafont

edwintorok (approved):
Looks good, we can add the adaptive zstd as a module later on when testing migration.

lindig:
Yes, I am planning to add more specific compressors when working on new use cases.

PR #4722, opened 2022-05-24
Some SDK fixes

kc284:
The following commits have already been reviewed within XenCenter, so they can be ignored:

- CP-37091: Do not use a loop for only one iteration. https://github.com/xenserver/xenadmin/pull/2985
- CA-365905 (XSI-1215): Create a temporary file in the target download folder instead of the user's temp folder if the latter is in a different drive from the target folder. https://github.com/xenserver/xenadmin/pull/2993

These are new changes:

- Renamed parameter of abstract method to match the parameter name in the method's implementations.
- CA-355432: Fixed generation of method overloads.

The last commit, apart from the bug referenced, it also contains the SDK counterpart of CP-38610.

### DIFF
--- a/ocaml/sdk-gen/common/CommonFunctions.ml
+++ b/ocaml/sdk-gen/common/CommonFunctions.ml
@@ -33,12 +33,22 @@ open Datamodel_types
 
 type wireProtocol = XmlRpc | JsonRpc
 
+type rpm_version = {major: int; minor: int; micro: int}
+
 let finally f ~(always : unit -> unit) =
   match f () with
   | result ->
       always () ; result
   | exception e ->
       always () ; raise e
+
+let parse_to_rpm_version_option inputStr =
+  match String.split_on_char '.' inputStr with
+  | [x; y; z] ->
+      Some
+        {major= int_of_string x; minor= int_of_string y; micro= int_of_string z}
+  | _ ->
+      None
 
 let string_of_file filename =
   let in_channel = open_in filename in
@@ -65,27 +75,7 @@ let escape_xml s =
   |> Astring.String.cuts ~sep:">" ~empty:true
   |> String.concat "&gt;"
 
-let rec list_distinct list =
-  match list with
-  | [] ->
-      []
-  | [x] ->
-      [x]
-  | hd1 :: (hd2 :: _ as tl) ->
-      if hd1 = hd2 then
-        list_distinct tl
-      else
-        hd1 :: list_distinct tl
-
-let rec list_last = function
-  | [x] ->
-      x
-  | _ :: tl ->
-      list_last tl
-  | [] ->
-      failwith "Cannot return the last element of an empty list."
-
-and list_index_of x list =
+let list_index_of x list =
   let rec index_rec i = function
     | [] ->
         raise Not_found
@@ -94,15 +84,7 @@ and list_index_of x list =
   in
   try index_rec 0 list with Not_found -> -1
 
-let rec gen_param_groups_for_releases releaseOrder params =
-  match releaseOrder with
-  | [] ->
-      [("", [])]
-  | hd :: tl ->
-      (hd, List.filter (fun x -> List.mem hd x.param_release.internal) params)
-      :: gen_param_groups_for_releases tl params
-
-and is_method_static message =
+let is_method_static message =
   match message.msg_params with
   | [] ->
       true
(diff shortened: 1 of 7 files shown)


ocaml/sdk-gen/csharp/autogen/src/Session.cs
```diff
@@ -404,11 +405,12 @@ public void login_with_password(string username, string password, API_Version ve
         private void SetAPIVersion()
         {
             Dictionary<XenRef<Pool>, Pool> pools = Pool.get_all_records(this);
-            foreach (Pool pool in pools.Values)
+
+            if (pools.Values.Count > 0)
             {
+                var pool = pools.Values.First();
                 Host host = Host.get_record(this, pool.master);
                 APIVersion = Helper.GetAPIVersion(host.API_version_major, host.API_version_minor);
```
psafont, line 413: Instead of getting the first, wouldn't it be better to get the maximum or the minimum version of all the pools?

kc284: Theoretically speaking it would be more correct. Practically it does not make a difference since there is only one pool record in the pool.

ocaml/sdk-gen/common/CommonFunctions.ml
```diff
@@ -111,137 +93,188 @@ and is_method_static message =
...
+  String.length message.msg_name >= 3 && String.sub message.msg_name 0 3 = "set"
 
-and code_name_of_release_ext x =
-  try code_name_of_release x with UnspecifiedRelease -> ""
+and is_getter message =
+  String.length message.msg_name >= 3 && String.sub message.msg_name 0 3 = "get"
 
-and code_name_order = release_order |> List.map code_name_of_release
+and is_adder message =
+  String.length message.msg_name >= 3 && String.sub message.msg_name 0 3 = "add"
```
psafont, line 103:
```suggestion
and is_adder message = String.starts_with ~prefix:"add" message.msg_name
```

ocaml/sdk-gen/common/CommonFunctions.ml
```diff
@@ -111,137 +93,188 @@ and is_method_static message =
...
-    message.msg_params
-  else
-    List.tl message.msg_params
+and is_setter message =
+  String.length message.msg_name >= 3 && String.sub message.msg_name 0 3 = "set"
 
-and code_name_of_release_ext x =
-  try code_name_of_release x with UnspecifiedRelease -> ""
+and is_getter message =
+  String.length message.msg_name >= 3 && String.sub message.msg_name 0 3 = "get"
```
psafont, line 100:
```suggestion
and is_getter message = String.starts_with ~prefix:"get" message.msg_name
```

ocaml/sdk-gen/common/CommonFunctions.ml
```diff
@@ -111,137 +93,188 @@ and is_method_static message =
   | {param_type= ty; _} :: _ ->
       not (ty = Ref message.msg_obj_name)
 
-and get_method_params_list message =
-  if is_method_static message then
-    message.msg_params
-  else
-    List.tl message.msg_params
+and is_setter message =
+  String.length message.msg_name >= 3 && String.sub message.msg_name 0 3 = "set"
```
psafont, line 97:
```suggestion
and is_setter message = String.starts_with ~prefix:"set" message.msg_name
```

ocaml/sdk-gen/common/CommonFunctions.ml
```diff
@@ -111,137 +93,188 @@ and is_method_static message =
...
-            []
-          else
-            snd hd
-        in
-        valid :: getValid tl
-  in
-  let filteredGroups =
+and is_remover message =
+  String.length message.msg_name >= 6
+  && String.sub message.msg_name 0 6 = "remove"
```
psafont, line 107:
```suggestion
and is_remover message = String.starts_with ~prefix:"remove" message.msg_name
```

ocaml/sdk-gen/common/CommonFunctions.ml
```diff
@@ -111,137 +93,188 @@ and is_method_static message =
...
+      -1
+  | _, None ->
+      1
+  | Some v1, Some v2 ->
+      if v1.major == v2.major && v1.minor == v2.minor && v1.micro == v2.micro
+      then
+        0
+      else if v1.major == v2.major && v1.minor == v2.minor then
+        v1.micro - v2.micro
+      else if v1.major == v2.major then
```
psafont, line 137: Here we want to use value equality, not physical (in memory) equality
```suggestion
      if v1.major = v2.major && v1.minor = v2.minor && v1.micro = v2.micro
      then
        0
      else if v1.major = v2.major && v1.minor = v2.minor then
        v1.micro - v2.micro
      else if v1.major = v2.major then
```

ocaml/sdk-gen/common/CommonFunctions.ml
```diff
@@ -111,137 +93,188 @@ and is_method_static message =
...
+        v1.minor - v2.minor
+      else
+        v1.major - v2.major
+
+let rec lifecycle_matcher milestone lifecycle =
+  match lifecycle with
+  | [] ->
+      ""
+  | (x, y, _) :: tl ->
+      if x == milestone then
```
psafont, line 147:
```suggestion
      if x = milestone then
```

ocaml/sdk-gen/common/CommonFunctions.ml
```diff
@@ -111,137 +93,188 @@ and is_method_static message =
...
+  in
+  collator [] [] l
 
-and get_prototyped_release lifecycle =
-  match lifecycle with [(Prototyped, release, _)] -> release | _ -> ""
+let gen_param_groups message params =
+  let expRelease = get_prototyped_release message.msg_lifecycle in
+  let paramGroups = group_params_per_release params in
+  let overloadGroups =
+    List.rev (List.map List.concat (List.map List.rev (collate paramGroups)))
```
psafont, line 201: This is quite a mouthful, and it's tough to see what is `collate` doing, I'd like to try to come up with something easier to understand. Could you try to explain the relationship between parameter groups and overload groups?

ocaml/sdk-gen/common/CommonFunctions.ml
```diff
@@ -111,137 +93,188 @@ and is_method_static message =
...
-  else if
-    (not (clsReleaseIndex = -1))
-    && (not (msgReleaseIndex = -1))
-    && clsReleaseIndex < msgReleaseIndex
-  then
-    get_first_release_string msgRelease
+  let classRelease = get_published_release cls.obj_lifecycle in
+  let msgRelease = get_published_release message.msg_lifecycle in
+  let expRel = get_prototyped_release message.msg_lifecycle in
+  if msgRelease == "" && expRel <> "" then
```
psafont, line 237:
```suggestion
  if msgRelease = "" && expRel <> "" then
```

ocaml/sdk-gen/common/CommonFunctions.ml
```diff
@@ -111,137 +93,188 @@ and is_method_static message =
...
-  else if
-    (not (clsReleaseIndex = -1))
-    && (not (fieldReleaseIndex = -1))
-    && clsReleaseIndex < fieldReleaseIndex
-  then
-    get_first_release_string fieldRelease
+  let clsRelease = get_published_release cls.obj_lifecycle in
+  let fieldRelease = get_published_release field.lifecycle in
+  let expRel = get_prototyped_release field.lifecycle in
+  if fieldRelease == "" && expRel <> "" then
```
psafont, line 274:
```suggestion
  if fieldRelease = "" && expRel <> "" then
```

ocaml/sdk-gen/common/CommonFunctions.ml
```diff
@@ -285,17 +308,39 @@ let json_releases =
...
   in
   let unique_version_bumps = get_unique release_order_full in
+  let version_index_of x list =
+    let rec index_rec i = function
+      | [] ->
+          raise Not_found
+      | hd :: tl ->
+          if
+            hd.version_major == x.version_major
+            && hd.version_minor == x.version_minor
```
psafont, line 318:
```suggestion
            hd.version_major = x.version_major
            && hd.version_minor = x.version_minor
```

psafont (changes requested):

ocaml/sdk-gen/common/CommonFunctions.ml
lindig, line 97: I'm not a big fan of using `String.sub` for sub-string matching. How about `Astring.String.is_prefix`

ocaml/sdk-gen/common/CommonFunctions.ml
kc284: Although modern versions of C# support optional parameters, I wanted to avoid making too many changes to the existing C# generator, and also wanted to be able to reuse at least some of the code with the generator for java, which does not support optional parameters in a straight forward way, so I kept the generation of traditional method overloads.

This means that if the Ocaml call has a number of parameters added at 4 different releases, e.g.
`call(param_rio, param_rio, param_lima, param_1_302_0, param_1_302_0, param_22_1_18)`
group_params_per_release splits the above parameters in 4 groups and then the "mouthful" line combines them so this results to the following c#/java overloads:
`call(param_rio, param_rio)`
`call(param_rio, param_rio, param_lima)`
`call(param_rio, param_rio, param_lima, param_1_302_0, param_1_302_0)`
`call(param_rio, param_rio, param_lima, param_1_302_0, param_1_302_0, param_22_1_18)`

psafont (approved):
Are there unit-tests for the SDK? Or example binaries that can be run to test these

kc284:
No, there are no unit tests. The way I normally work with these is run the build, get the source rpm and then diff the generated code with the old one. The commit with the corrections produces the same code as the rest of the PR. The whole PR results to several new exposed API members which were previously missing (the ones where the release info is an rpm version).

ocaml/sdk-gen/common/CommonFunctions.ml
```diff
@@ -111,137 +92,180 @@ and is_method_static message =
...
+  message.msg_tag = FromObject Make || message.msg_name = "create"
+
+and is_real_constructor message = message.msg_tag = FromObject Make
+
+and is_destructor message =
+  message.msg_tag = FromObject Delete || message.msg_name = "destroy"
+
+let code_name_order = release_order |> List.map code_name_of_release
+
+let compare_versions x y =
```
lindig, line 113: This could be made simpler by relying on the natural order of tuples.
```
let compare x y =
  let (x1, x2, x3 as xx) = x.major, x.minor, x.micro in
  let (y1, y2, y3 as yy) = y.major, y.minor, y.micro in
    compare xx yy
```

PR #4725, opened 2022-06-01
xenopsd/xc: do not log error when querying for migrability

psafont:
QMP usually returns an error because nvmes are not migratable, but this
is expected, it doesn't make sense to log  this as an error.

### DIFF
--- a/ocaml/xenopsd/xc/device_common.ml
+++ b/ocaml/xenopsd/xc/device_common.ml
@@ -523,12 +523,16 @@ let with_qmp_connection domid f =
     (fun () -> f connection)
     (fun () -> exec (fun () -> Qmp_protocol.close connection))
 
-(** [qmp_send_cmd domid cmd] sends [cmd] to [domid] and checks that the result
-    it returns is Success. Otherwise it will raise [QMP_Error].
+(** [qmp_send_cmd ~send_fd ~may_fail domid cmd] sends [cmd] to [domid] and
+    checks that the result it returns is Success. Otherwise it will raise
+    [QMP_Error].
 
     [qmp_send_cmd] can send optionally a file descriptor [send_fd] over the
-    connection before it sends the command. This is required for some commands. *)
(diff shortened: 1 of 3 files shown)


ocaml/xenopsd/xc/device_common.ml
```diff
@@ -528,7 +528,7 @@ let with_qmp_connection domid f =
 
     [qmp_send_cmd] can send optionally a file descriptor [send_fd] over the
     connection before it sends the command. This is required for some commands. *)
-let qmp_send_cmd ?send_fd domid cmd =
+let qmp_send_cmd ?send_fd ?may_fail domid cmd =
```
edwintorok, line 531: you can do `?(may_fail=false)` to avoid having to match an optional boolean below.

ocaml/xenopsd/xc/device_common.ml
```diff
@@ -553,6 +553,7 @@ let qmp_send_cmd ?send_fd domid cmd =
           result
       | message ->
           let msg' = Qmp.string_of_message message in
-          error "QMP result for domid %d: %s (%s)" domid msg' __LOC__ ;
+          if may_fail = Some true then
```
edwintorok, line 556: wouldn't you want to skip the error message when you expect a failure, does this need to be `if not may_fail then error...`

edwintorok (commented):
the error is expected as long as the VM only has emulated devices (and should be removed later), but yeah it is quite common to see this during VM boot.

ocaml/xenopsd/xc/device_common.ml
psafont: It's when the caller expects the call to fail that it doesn't make sense to log the failure there, this is because on the caller side the failure is accounted for and there is no need to have logs to debug the situation. It's only in cases where the failure is not expected that it's beneficial to log it.

lindig (approved):

edwintorok (approved):

PR #4726, opened 2022-06-01
CA-391660: xe-xentrace: fix and allow dumping on high CPU

edwintorok:
TODO: would be better to rewrite this in python or ocaml

### DIFF
--- a/scripts/xe-xentrace
+++ b/scripts/xe-xentrace
@@ -118,7 +118,7 @@ fi
 echo "done"
 
 echo "Xentrace: starting"
-/usr/bin/xentrace -D -e all -r 1 ${mnt}/trace -T ${TIME}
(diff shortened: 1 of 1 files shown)


edwintorok:
Lets leave rewriting to another PR

psafont (approved):
The base branch is growing, should we prepare for merging? the changes are quite various

edwintorok:
Yes, I've basically been putting everything into feature/perf since XS8 got released, because I'm constantly moved to work on new things rather than finishing any of the old ones first (and testing each small PR individually is too high overhead given that we have no internal CI yet).

I need to create a build, run some BVT/BST, perhaps review the code once more myself (I've forgotten what is actually in the branch...), and then if it works we can merge it.

robhoes (approved):

PR #4729, opened 2022-06-08
CP-39996: Generate and push docs to xapi-storage

psafont:
Xapi-storage still hosts the documentation for SMAPIv3, so to keep it up-to-date this repository needs to push the changes to xapi-project/xapi-storage.

Tested manually that this correctly updates the correct branch. The first commit will create a `.nojekyll` file once this is merged to master. This is expected as the documentation files have not changed since the last commit and the file is specific to github's pages to avoid its internal jekyll deployment.  Currently the deploy action avoids pushing empty commits, so only changes will create documentation pushes.

I believe there might be a step missing to actually publish to github pages but that can be dealt with a github action in xapi-storage.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,55 @@
+name: Generate and upload docs
+
+on:
+  push:
+    branches: master
+
+jobs:
+  ocaml:
+    name: Docs
+    runs-on: ubuntu-20.04
+    env:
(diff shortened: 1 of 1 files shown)


edwintorok (approved):
Yes, something will need to publish from `slate` to `gh_pages` branch

psafont:
Since this just needs to push some files to a remote repository, I don't thin the choice of action to push it is ideal, even if it works

robhoes (approved):

PR #4730, opened 2022-06-08
vTPM state storage v0

edwintorok:
See the description in the commit messages.

### DIFF
--- a/ocaml/xapi/xapi_xenops.ml
+++ b/ocaml/xapi/xapi_xenops.ml
@@ -386,20 +386,28 @@ let builder_of_vm ~__context (vmref, vm) timeoffset pci_passthrough vgpu =
       warn "QEMU stub domains are no longer implemented" ;
 
     let tpm_of_vm () =
-      if bool vm.API.vM_platform false "vtpm" then (
-        if vm.API.vM_VTPMs = [] then (
-          let ref () = Ref.make () in
-          let uuid () = Uuid.(to_string (make ())) in
-          let profile = [] in
-          let other_config = [] in
-          let contents = ref () in
-          Db.Secret.create ~__context ~ref:contents ~uuid:(uuid ()) ~value:""
-            ~other_config ;
-          Db.VTPM.create ~__context ~ref:(ref ()) ~uuid:(uuid ()) ~vM:vmref
-            ~profile ~contents
-        ) ;
-        Some Xenops_interface.Vm.Vtpm
-      ) else
+      if bool vm.API.vM_platform false "vtpm" then
+        let uuid =
+          match vm.API.vM_VTPMs with
+          | [] ->
+              let ref () = Ref.make () in
+              let uuid () = Uuid.(to_string (make ())) in
+              let profile = [] in
+              let other_config = [] in
+              let contents = ref () in
+              Db.Secret.create ~__context ~ref:contents ~uuid:(uuid ())
+                ~value:"" ~other_config ;
+              let vtpm_uuid = uuid () in
+              Db.VTPM.create ~__context ~ref:(ref ()) ~uuid:vtpm_uuid ~vM:vmref
+                ~profile ~contents ;
+              vtpm_uuid
+          | [self] ->
+              Db.VTPM.get_uuid ~__context ~self
+          | _ :: _ :: _ ->
+              failwith "Multiple vTPMs are not supported"
+        in
+        Some (Xenops_interface.Vm.Vtpm (Uuidm.of_string uuid |> Option.get))
+      else
         None
     in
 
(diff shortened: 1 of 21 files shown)


ocaml/xenopsd/xc/device.ml
```diff
@@ -1264,13 +1264,80 @@ module Varstored = SystemdDaemonMgmt (struct
...
+      Unixext.write_string_to_file path state
+    end else
+      debug "vTPM state for domid %d is empty: not restoring" domid
+
+  let start_daemon dbg ~xs ~path ~args ~domid ~vm_uuid ~vtpm_uuid ~index () =
+    let state =
+      Varstore_privileged_client.Client.vtpm_get_contents dbg vtpm_uuid
+      |> Base64.decode_exn
+    in
+    let chroot = Xenops_sandbox.Swtpm_guard.chroot ~domid ~vm_uuid in
```
psafont, line 1310: The root of the sandbox depends on the domid and vm_uuid. Expecting the sandbox to exist to recover state seems suspect to me instead of simply deleting the sandbox when we need to.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -386,20 +386,27 @@ let builder_of_vm ~__context (_, vm) timeoffset pci_passthrough vgpu =
...
+              let contents = ref () in
+              Db.Secret.create ~__context ~ref:contents ~uuid:(uuid ())
+                ~value:"" ~other_config ;
+              let vtpm_uuid = uuid () in
+              Db.VTPM.create ~__context ~ref:(ref ()) ~uuid:vtpm_uuid ~vM:vmref
+                ~profile ~contents;
+              vtpm_uuid
+          | self :: _ ->
+              Db.VTPM.get_uuid ~__context ~self
+          (* TODO: multiple vtpm supports *)
```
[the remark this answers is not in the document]
edwintorok: I replaced this with a failwith

ocaml/xenopsd/xc/device.ml
edwintorok: Yes, we probably want a way to check whether a file exists in the sandbox without necessarily creating the sandbox (which would create the file if it doesn't exist). I'll see whether I can plumb the Chroot.t parameter through.

edwintorok:
Note that this assumes that the backend is always xapi-db. Once we have a 'backend' field again then this should be made more generic, to allow any backend (and to move the backend specific code into a module of its own with a clear signature, so it'll be obvious how to implement new backends).

psafont:
If I read the code correctly, writing the state filepath into a xenstore key containing the vtpm_uuid is done to support mustiple vtpms per VM, but is currently unused.

Is there a reason for using this mechanism instead of, let's say symlinks in the chroot to stablish this relationship off-band?

edwintorok:
Looks like it is used after all: in `vtpms_of_domid` we read the vtpm uuids from xenstore, this makes the stop code generic and works whether or not the VM has a vTPM attached or not (i.e. for the 0 vs 1 usecase rather than 1 vs 1+ vTPM usecase). xenopsd doesn't have access to XAPI's DB so it needs to retrieve that information from somewhere, however I think it should be set/retrieved in the VM's "xenopsd" configuration, and not xenstore directly. (there is a "DB" module in xenops_server that knows about other devices similarly, and doesn't know about TPMs yet).

Using xenstore here was just a shortcut: this needs a proper VTPM_DB module in lib/xenops_server instead. Thanks for spotting this.

edwintorok:
Looks like it is used after all: in `vtpms_of_domid` we read the vtpm uuids from xenstore, this makes the stop code generic and works whether or not the VM has a vTPM attached or not (i.e. for the 0 vs 1 usecase rather than 1 vs 1+ vTPM usecase). xenopsd doesn't have access to XAPI's DB so it needs to retrieve that information from somewhere, however I think it should be set/retrieved in the VM's "xenopsd" configuration, and not xenstore directly. (there is a "DB" module in xenops_server that knows about other devices similarly, and doesn't know about TPMs yet).

Using xenstore here was just a shortcut: this needs a proper VTPM_DB module in lib/xenops_server instead. Thanks for spotting this.

psafont (approved):
I think the chroot function should not be exposed. I'll fix it once this is merged

ocaml/xenopsd/xc/domain.ml
```diff
@@ -1745,12 +1755,30 @@ let write_varstored_record task ~xs domid main_fd =
   Io.write main_fd varstored_record ;
   return ()
 
+let forall f l =
+  let open Suspend_image.M in
```
lindig, line 1759: Is this a monadic `iter`? In that case I would find `forall` a bit misleading.

ocaml/xenopsd/xc/xc_resources.ml
```diff
@@ -95,3 +95,4 @@ let nonessentials =
   @ Resources.hvm_guests
   @ Resources.pv_guests
   @ Resources.pvinpvh_guests
+  @ Resources.vtpm_guests
```
lindig, line 98: Moving to `List.concat` as this gets longer would be good.

lindig (approved):

PR #4731, opened 2022-06-09
CP-39806: remove code without a stable formatting

psafont:
newer ocamlformat are unable to reach a stable format for code that
has long comment inside long boolean expressions, change the code so an
ocamlformat version compatible with ocaml 4.14 can be used to format the
code.

See https://github.com/ocaml-ppx/ocamlformat/issues/2057

### DIFF
--- a/ocaml/xapi/xapi_session.ml
+++ b/ocaml/xapi/xapi_session.ml
@@ -1235,22 +1235,32 @@ let logout_subject_identifier ~__context ~subject_identifier =
   let all_sessions = Db.Session.get_all ~__context in
   let current_session = Context.get_session_id __context in
   (* we filter the sessions to be destroyed *)
+  (* 1. we never allow local_superuser sessions to be forcibly logged out *)
+  let is_not_local_superuser s =
+    not (Db.Session.get_is_local_superuser ~__context ~self:s)
+  in
+  (* 2. we remove the session associated with this function call from the list
+        of all sessions to be destroyed *)
+  let is_not_current_session s =
+    Db.Session.get_uuid ~__context ~self:s
+    <> Db.Session.get_uuid ~__context ~self:current_session
+  in
+  (* 3. we only consider those sessions associated with the specific subject_id
+        received as parameter *)
+  let is_associated_with_user_logging_out s =
+    (* TODO: better to look up the membership closure *)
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_session.ml
```diff
@@ -1235,22 +1235,32 @@ let logout_subject_identifier ~__context ~subject_identifier =
...
-        ((* 3.1. the sid of the authenticated user *)
-         Db.Session.get_auth_user_sid ~__context ~self:s = subject_identifier
-        || (* 3.2. any sids of the group that authenticated the user *)
-           (* TODO: better to look up the membership closure *)
-        get_group_subject_identifier_from_session ~__context ~session:s
-        = subject_identifier
-        )
+        is_not_local_superuser s
+        && is_not_current_session s
+        && is_associated_with_user_logging_out s
```
robhoes, line 1263: Is `a && b && c || d` equal to `a && b && (c || d)` (as the transformation does this)? It looks like this was the intention of the original code, anyway.

robhoes: https://v2.ocaml.org/manual/expr.html#ss%3Aprecedence-and-associativity says that associativity for Boolean operators is "left", so this should be fine.

robhoes (approved):

ocaml/xapi/xapi_session.ml
robhoes: Hmm... but `&&` has higher precedence :/

lindig: `&&` has higher precedence than `||`, so `b && c || d` is `(b && c) || d`

psafont: From the comments it looks like the intent was the new code. Now the question is whether the precious behaviour is considered a bug or a feature

robhoes:
```
utop # false && true && true || true;;
- : bool = true
─( 11:01:14 )─< command 8 >──────────────────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # false && true && (true || true);;
- : bool = false
```

psafont: I think so as well, I'd like to understand the implications

PR #4732, opened 2022-06-10
Compression of migration streams

lindig:
Using Zstd to compress the memory stream during migration. This is beneficial when network capacity is limited, especially when migrating multiple VMs.

### DIFF
--- a/ocaml/xapi-cli-server/cli_frontend.ml
+++ b/ocaml/xapi-cli-server/cli_frontend.ml
@@ -1539,6 +1539,7 @@ let rec cmdtable_data : (string * cmd_spec) list =
           ; "remote-network"
           ; "force"
           ; "copy"
+          ; "compress"
           ; "vif:"
           ; "vdi:"
           ]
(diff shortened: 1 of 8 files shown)


lindig:
Benchmarks have demonstrated that compression offers a clear advantage on 1GB/s networks when migrating multiple VMs.
    The current implementation has passed a BVT and BST test suite, suggesting it does not regress
    However, a test continuously migrating multiple VMs showed failures: xapi would lose track of where a VM is hosted. This is caused by a race condition.
    The race condition is related to a changed handshake protocol when using compression: the last handshake is skipped.
    The reason to skip the handshake is that handshaking is incompatible with compression - handshakes happen on the same channel as the data transfer. The compressor introduces a latency as it accumulates data before sending it on. This leads to a deadlock.
    Skipping the last handshake leads to a race condition: the sender, whithout the last handshake, calls pool_migrate_complete on the receiver, which (among other things) sets the resident-on field.
    This call might happen too early, before the final rename of the received VM has happened, causing the VM not to be found by xenopsd. And this is the moment where xapi then loses track of the VM.

PR #4734, opened 2022-06-15
CP-39805: Adapt code to cmdliner 1.1.0

psafont:
This introduces quite a bit of churn, but it fixes several inconsistencies on how the xcp-idl  binaries handle error codes and simplifies how Xcp_service.configure2 works: now it exists when an error happens, and returns a unit, so no actions from the caller are needed.

This PR depends on rpclib 9.0.0, which forces a bit of complication to process the result of the cmdline generator, but aside from the boilerplate it should be fine.

Marked as draft as it needs changes in xs-opam.

TODO: 
- [x] change `Xcp_service.eval_cmdline` to print errors on failure
- [x] support changes introduced by rpclib 8.1.2

### DIFF
--- a/ocaml/xapi-idl/cluster/cluster_cli.ml
+++ b/ocaml/xapi-idl/cluster/cluster_cli.ml
@@ -4,33 +4,20 @@ open Cluster_interface
 
 module Cmds = LocalAPI (Cmdlinergen.Gen ())
 
-let version_str description =
-  let maj, min, mic = description.Idl.Interface.version in
-  Printf.sprintf "%d.%d.%d" maj min mic
+let doc =
+  String.concat ""
+    [
+      "A CLI for the cluster API. This tool is not intended to be used as "
+    ; "an end user tool"
+    ]
 
-open! Cmdliner
-
-let cmds =
+let cmdline_gen () =
   List.map
-    (fun gen ->
-      let t, i = gen (Cluster_client.rpc_internal Cluster_client.json_url) in
-      Cmd.v i t
-    )
+    (fun t -> t (Cluster_client.rpc_internal Cluster_client.json_url))
     (Cmds.implementation ())
 
-let cli () =
-  let default = Term.(ret (const (fun _ -> `Help (`Pager, None)) $ const ())) in
-  let info =
-    let doc =
-      String.concat ""
-        [
-          "A CLI for the cluster API. This tool is not intended to be used as "
-        ; "an end user tool"
-        ]
-    in
-    Cmd.info "cluster_cli" ~version:(version_str Cmds.description) ~doc
-  in
-  let cmd = Cmd.group ~default info cmds in
-  Cmd.eval_value cmd
+let cli =
+  Xcp_service.cli ~name:"cluster_cli" ~doc ~version:Cmds.description.version
+    ~cmdline_gen
 
-let () = match cli () with Ok (`Ok f) -> f () | _ -> ()
+let () = Xcp_service.eval_cmdline cli
--- a/ocaml/xapi-idl/cluster/cluster_cli.ml
+++ b/ocaml/xapi-idl/cluster/cluster_cli.ml
@@ -8,23 +8,29 @@ let version_str description =
   let maj, min, mic = description.Idl.Interface.version in
   Printf.sprintf "%d.%d.%d" maj min mic
 
-let default_cmd =
-  let doc =
-    String.concat ""
-      [
-        "A CLI for the cluster API. This tool is not intended to be used as an "
-      ; "end user tool"
-      ]
-  in
-  ( Cmdliner.Term.(ret (const (fun _ -> `Help (`Pager, None)) $ const ()))
-  , Cmdliner.Term.info "cluster_cli"
-      ~version:(version_str Cmds.description)
-      ~doc
-  )
+open! Cmdliner
+
+let cmds =
+  List.map
+    (fun gen ->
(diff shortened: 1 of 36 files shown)


ocaml/xapi-storage-cli/main.ml
```diff
@@ -952,10 +945,15 @@ let cmds =
...
-  | `Error _ ->
-      exit 1
-  | _ ->
-      exit 0
+  let default =
+    Term.(ret (const (fun _ -> `Help (`Pager, None)) $ common_options_t))
+  in
+  let doc = "interact with an XCP storage management service" in
+  let info =
+    Cmd.info "sm-cli" ~version:"1.0.0" ~sdocs:_common_options ~doc ~man:help
```
edwintorok, line 956: At some point we can fixup these versions to point to the true version, but for now lets keep it as is

ocaml/xapi-idl/cluster/cluster_cli.ml
```diff
@@ -8,23 +8,29 @@ let version_str description =
...
+        ; "an end user tool"
+        ]
+    in
+    Cmd.info "cluster_cli" ~version:(version_str Cmds.description) ~doc
+  in
+  let cmd = Cmd.group ~default info cmds in
+  Cmd.eval_value cmd
 
-let _ = cli ()
+let () = match cli () with Ok (`Ok f) -> f () | _ -> ()
```
edwintorok, line 36: Can't we use Cmd.eval_result or exit (...) in some way here to also ensure that errors get printed properly?

lindig (approved):
As can be seen, moving to the new Cmdliner requires a lot of changes. Should we keep two versions of Cmdliner in xs-opam such that we can transition packages one by one? And use opam version constraints to pick up the right one? Otherwise it might take a long time to transition all packages

ocaml/xapi-idl/lib/xcp_service.ml
```diff
@@ -476,7 +476,18 @@ let configure ?(options = []) ?(resources = []) () =
...
+  | Ok (`Ok _) ->
+      ()
+  | Ok `Help | Ok `Version ->
+      exit 0
+  | Error `Parse ->
+      exit 123
+  | Error `Term ->
+      exit 124
+  | Error `Exn ->
+      exit 125
```
edwintorok, line 490: Instead of hardcoding the numbers you can use the constants from here: https://erratique.ch/software/cmdliner/doc/Cmdliner/Cmd/Exit/index.html

ocaml/xapi-idl/cluster/cluster_cli.ml
psafont: We could yes, At this point it might be worth thinking about a function in xcp_service to reduce repetition here

edwintorok: Yes something in xapi-idl/, all these CLIs are almost the same except for a few things like names, and those could easily be parameters. Perhaps do that as a commit on top of the other ones? (the conversion from old to new cmdliner is still good to see separately and then the refactoring to use common code).

psafont: indeed

ocaml/xapi-idl/gpumon/gpumon_cli.ml
```diff
@@ -6,23 +6,30 @@ let version_str description =
...
+      let t, i = t Gpumon_client.rpc in
+      Cmd.v i t
+    )
+    (Cmds.implementation ())
 
 let cli () =
-  let rpc = Gpumon_client.rpc in
-  Cmdliner.Term.eval_choice default_cmd
-    (List.map (fun t -> t rpc) (Cmds.implementation ()))
+  let default = Term.(ret (const (fun _ -> `Help (`Pager, None)) $ const ())) in
```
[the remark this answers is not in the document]
psafont: This has been consolidated in a shared function

ocaml/xapi-storage-cli/main.ml
lindig: I am using this to get access to strings that we could use
```
! /bin/bash

cat <<EOF
let git_revision = "$(git describe --always --long)"
let build_time = "$(date)"
let build_timestamp = $(date +%s)
EOF

cat <<'EOF'
let json =
 `O [ "revision", `String git_revision
    ; "date", `String build_time
    ; "timestamp", `Float (float_of_int build_timestamp)
    ]
EOF
```
```
(rule
 (target build.ml)
 (deps (universe) build.sh)
 (action
  (with-stdout-to
   %{target}
   (bash "./build.sh"))))
```

psafont: We have infrastructure to get version number in place (Build_info). It needs to be cleaned up so it doesn't conflict with dune's Build_info and the adopt it elsewhere.

edwintorok: We should see whether we can use Dune's build-info support without requiring to define the `XAPI_VERSION` global, perhaps by having a separate module that parses the build info generated by dune. Dune has some support to avoid needless recompilations whenever the build info changes (it only embeds the true version number into the binary when it is installed, not when it is initially built/executed locally), but only if its own build info module is used.


Although that is a bit complicated because Dune's build info is designed NOT to be used at build time (it returns None there), but we do use version info during build time in gen_lifecycle.ml. With a few tweaks to the Makefile I think we can force these to use a (locally installed) version though, I'll try something and submit a PR to master if I succeed.

edwintorok (approved):

PR #4740, opened 2022-06-22
CP-40032: Expose VTPM creation and destruction both on the CLI and the API

psafont:
For now, only allow a single VTPM to be associated to a VM, and only
allow VTPM creation and destruction on halted VMs to avoid desync of the
DB and the actual running state of the devices.

### DIFF
--- a/ocaml/xapi/xapi_vtpm.ml
+++ b/ocaml/xapi/xapi_vtpm.ml
@@ -12,12 +12,23 @@
    GNU Lesser General Public License for more details.
  *)
 
+let assert_no_vtpm_associated ~__context vm =
+  match Db.VM.get_VTPMs ~__context ~self:vm with
+  | [] ->
+      ()
+  | vtpms ->
+      let amount = List.length vtpms |> Int.to_string in
+      raise Api_errors.(Server_error (vtpm_max_amount_reached, [amount]))
+
 let introduce ~__context ~uuid ~vM ~profile ~contents =
   let ref = Ref.make () in
   Db.VTPM.create ~__context ~ref ~uuid ~vM ~profile ~contents ;
   ref
 
 let create ~__context ~vM =
+  assert_no_vtpm_associated ~__context vM ;
+  Xapi_vm_lifecycle.assert_initial_power_state_is ~__context ~self:vM
+    ~expected:`Halted ;
   let uuid = Uuid.(to_string (make ())) in
   let profile = Db.VM.get_default_vtpm_profile ~__context ~self:vM in
   let contents = Xapi_secret.create ~__context ~value:"" ~other_config:[] in
@@ -25,6 +36,9 @@ let create ~__context ~vM =
   ref
 
 let destroy ~__context ~self =
+  let vm = Db.VTPM.get_VM ~__context ~self in
+  Xapi_vm_lifecycle.assert_initial_power_state_is ~__context ~self:vm
+    ~expected:`Halted ;
   let secret = Db.VTPM.get_contents ~__context ~self in
   Db.Secret.destroy ~__context ~self:secret ;
   Db.VTPM.destroy ~__context ~self
--- a/ocaml/idl/datamodel_errors.ml
+++ b/ocaml/idl/datamodel_errors.ml
@@ -1940,9 +1940,11 @@ let _ =
   error Api_errors.invalid_repository_proxy_url ["url"]
     ~doc:"The repository proxy URL is invalid." () ;
   error Api_errors.invalid_repository_proxy_credential []
-    ~doc:"The repository proxy username/password is invalid." ()
+    ~doc:"The repository proxy username/password is invalid." () ;
 
-let _ =
+  error Api_errors.vtpm_max_amount_reached ["amount"]
+    ~doc:"The VM cannot be associated with more VTPMs." () ;
+
   message
     (fst Api_messages.ha_pool_overcommitted)
     ~doc:
--- a/ocaml/xapi-cli-server/cli_frontend.ml
+++ b/ocaml/xapi-cli-server/cli_frontend.ml
@@ -3569,6 +3569,24 @@ let rec cmdtable_data : (string * cmd_spec) list =
       ; flags= []
       }
     )
+  ; ( "vtpm-create"
+    , {
+        reqd= ["vm-uuid"]
+      ; optn= []
+      ; help= "Create a VTPM associated with a VM."
+      ; implementation= No_fd Cli_operations.VTPM.create
+      ; flags= []
+      }
+    )
+  ; ( "vtpm-destroy"
+    , {
+        reqd= ["uuid"]
+      ; optn= []
+      ; help= "Destroy a VTPM"
+      ; implementation= No_fd Cli_operations.VTPM.destroy
+      ; flags= []
+      }
+    )
   ]
 
 let cmdtable : (string, cmd_spec) Hashtbl.t = Hashtbl.create 50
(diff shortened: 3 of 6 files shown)


ocaml/idl/datamodel_vtpm.ml
```diff
@@ -42,7 +55,7 @@ let t =
       ; (Extended, rel_next, "Added VTPM profiles")
       ; (Changed, rel_next, "Removed backend field")
       ]
-    ~gen_constructor_destructor:true ~name:_vtpm ~descr:"A virtual TPM device"
+    ~gen_constructor_destructor:false ~name:_vtpm ~descr:"A virtual TPM device"
```
edwintorok, line 58: I assume you need to disable autogeneration because you want to override them to add the assertions?

ocaml/xapi/xapi_vtpm.ml
```diff
@@ -12,19 +12,43 @@
...
+  | vtpms ->
+      let uuids =
+        List.map (fun v -> Db.VTPM.get_uuid ~__context ~self:v) vtpms
+      in
+      raise Api_errors.(Server_error (vtpm_max_amount_reached, uuids))
+
+let assert_vm_is_halted ~__context vm =
+  let attachable = [`Halted] in
+  match Db.VM.get_power_state ~__context ~self:vm with
+  | state when List.exists (( = ) state) attachable ->
```
edwintorok, line 28: using ``Halted` directly in the match would be shorter?

ocaml/xapi/xapi_vtpm.ml
```diff
@@ -12,19 +12,43 @@
...
   let profile = Db.VM.get_default_vtpm_profile ~__context ~self:vM in
   let contents = Xapi_secret.create ~__context ~value:"" ~other_config:[] in
   let ref = introduce ~__context ~uuid ~vM ~profile ~contents in
   ref
 
 let destroy ~__context ~self =
+  assert_vm_is_halted ~__context (Db.VTPM.get_VM ~__context ~self) ;
   let secret = Db.VTPM.get_contents ~__context ~self in
   Db.Secret.destroy ~__context ~self:secret ;
   Db.VTPM.destroy ~__context ~self
```
edwintorok, line 54: I assume xapi_db_gc will then go and delete the link from the VM to this vTPM?

ocaml/xapi/xapi_vtpm.ml
edwintorok: Also would be good to have an allowed operation on the VM object that tells you when you can or cannot create vtpms, and an allowed operation on the vtpm object to tell you when you can destroy it.
Then this can be updated (and hopefully this assertion code shared). I think the usual way is to put it into some lifecycle helper module and call the assertion in a strict mode when performing the API call.

ocaml/idl/datamodel_vtpm.ml
psafont: indeed, otherwise it's warnings galore in the server

ocaml/xapi/xapi_vtpm.ml
psafont: That's correct

ocaml/xapi/xapi_vtpm.ml
psafont: > using ``Halted` directly in the match would be shorter?
You're right, I was thinking of adding more that one state there and avoid duplication when generating the error message

I'm wary of adding vtpm creation / deletion to the allowed operations of the VM, as the operation is on the device, not the domain.

Having allowed operations on the vtpm would more sense if there were several operations there (I was thinking about attaching and detaching). Since it's a single operation, I'm not sure this would be of much benefit right now to have this.

ocaml/idl/datamodel_vtpm.ml
robhoes: Hmmm... what's this got to do with anything in the implementation of the module? What sort of warnings in server.ml?

psafont: it creates identical branches when matching for the api call used, so ocaml warns about one of them never being used

robhoes: If you still add the separate function to the `messages` below, yes. But why do you need that?

psafont:
> It would be nice if a client were able to prevent creation depending on whether the max number of vTPMs has been reached. 

Current error shows the vtpms associated with the vm, which can allow the client to know the maximum amount of vtpms allowed for that vm. I understand the UI wants to disallow the user from creating the vtpm if it cannot be created, I'll see how to better request this, and if the current mechanisms for other entities (probably allowed_operations)

ocaml/idl/datamodel_vtpm.ml
psafont: All this is to expose them to the API, couldn't find a way to explicitely expose them without defining them

psafont: The local index.html generated doesn't show implicit messages, with was confusing because there were available in the API regardless. I've removed the commit since it wasn't useful.

edwintorok (approved):

ocaml/xapi/xapi_vtpm.ml
```diff
@@ -12,19 +12,42 @@
...
+      raise Api_errors.(Server_error (vm_bad_power_state, error_params))
+
 let introduce ~__context ~uuid ~vM ~profile ~contents =
   let ref = Ref.make () in
   Db.VTPM.create ~__context ~ref ~uuid ~vM ~profile ~contents ;
   ref
 
 let create ~__context ~vM =
+  assert_no_vtpm_associated ~__context vM ;
+  assert_vm_is_halted ~__context vM ;
```
robhoes, line 42: You could use `Xapi_vm_lifecycle.assert_initial_power_state_is`.

robhoes: An alternative to this is to allow VTPM creation while the VM is running, with the understanding that it will only take effect after a VM reboot (and yes, we should make sure that reboot and shutdown+start behave the same...).

ocaml/xapi/xapi_vtpm.ml
```diff
@@ -12,19 +12,35 @@
...
 
+let assert_no_vtpm_associated ~__context vm =
+  match Db.VM.get_VTPMs ~__context ~self:vm with
+  | [] ->
+      ()
+  | vtpms ->
+      let uuids =
+        List.map (fun v -> Db.VTPM.get_uuid ~__context ~self:v) vtpms
+      in
+      raise Api_errors.(Server_error (vtpm_max_amount_reached, uuids))
```
robhoes, line 23: This should return refs, not uuids (as usual in the API). But doesn't it make more sense to return the number?

ocaml/xapi/xapi_vtpm.ml
robhoes: That's not what the GC does. If the link is defined a one-to-many relation, then if the "one" side is deleted, the other side is automatically updated by the database. A GC rule is needed for the case where the VM object is deleted, and its VTPM objects need to be automatically deleted. Do we have a rule for that?

ocaml/xapi/xapi_vtpm.ml
psafont: I just thought having all the references would be more useful that just having the number, but thinking about it it the information doesn't seem to be very actuable

robhoes (approved):

ocaml/xapi/xapi_vtpm.ml
psafont: > Do we have a rule for that?

No, such rule would delete the tpm state with the current model

psafont: I would think in currently that's true, I've created an issue to track that

PR #4741, opened 2022-06-22
CP-40027 VM migration introduce /services/xenops/migrate-mem

lindig:
* Migrating a VM works currently like this: The source xenopsd opens a
  HTTP connection to the destination xenopsd services/xenops/memory. This
  connection is used as a socket, i.e. in both direction. The handler on
  the destination puts the socket connection into VM_receive_memory. The
  source is using this connection for two purposes: sending the memory
  stream but also a handshake protocol that syncs up source and
  destination. There are more details about the GPU migration stream that
  I'm not touching upon.

* This patch introduces an additional route and handler ..../migrate-mem
  that will be used only for the VM memory stream and leave almost all
  signaling to the two purposes: sending VM memory and keep signaling to
  the original first connection. The motivation is to enable compression
  of the memory stream, which is incompatible with signaling because it
  would introduce latencies.

* All file descriptors (FDs) needed for the migration of a VM end up in
  the VM_migrate and VM_receive_memory operations. On the destination
  the new handler receive_mem stores the FD in a hash table from where
  it is retrieved by the VM_receive_memory implementation and passed on.

* The destination must stay compatible with the old protocol because it
  will receive VMs during a rolling pool upgrade from hosts running
  older software. We signal the use of the new protocol by using a
  cookie mem_migration and the reveiver use its presence to branch on
  the protocol it uses.

* Several changes are administrative: factor out in handlers for the
  error case. Moving parameters of the VM_receive_memory operation into
  a record.

### DIFF
--- a/ocaml/xenopsd/lib/xenops_server.ml
+++ b/ocaml/xenopsd/lib/xenops_server.ml
@@ -51,6 +51,12 @@ let cookie_vgpu_migration = "vgpu_migration"
 
 let cookie_mem_migration = "mem_migration"
 
+let cookie_mem_migration_value = Build_info.version
+
+let cookie_mem_compression = "compress"
+
+let cookie_mem_compression_value = "zstd"
+
 let backend = ref None
 
 let get_backend () =
@@ -163,6 +169,7 @@ type vm_migrate_op = {
   ; vmm_url: string
   ; vmm_tmp_src_id: Vm.id
   ; vmm_tmp_dest_id: Vm.id
+  ; vmm_compress: bool
 }
 [@@deriving rpcty]
 
@@ -172,6 +179,7 @@ type vm_receive_op = {
   ; vmr_memory_limit: int64
   ; vmr_socket: Unix.file_descr
   ; vmr_handshake: string option  (** handshake protocol *)
+  ; vmr_compressed: bool
 }
 [@@deriving rpcty]
 
@@ -2381,6 +2389,16 @@ and perform_exn ?subtask ?result (op : operation) (t : Xenops_task.task_handle)
         let vm = VM_DB.read_exn id in
         let dbg = (Xenops_task.to_interface_task t).Task.dbg in
         let url = Uri.of_string vmm.vmm_url in
+        let compress_memory = vmm.vmm_compress in
+        let compress =
+          match compress_memory with
+          | true ->
+              Zstd.Default.compress
+          | false ->
+              fun fd fn -> fn fd
+          (* do nothing *)
+        in
+        debug "%s compress memory: %b" __FUNCTION__ compress_memory ;
         (* We need to perform version exchange here *)
         let module B = (val get_backend () : S) in
         B.VM.assert_can_save vm ;
@@ -2416,7 +2434,6 @@ and perform_exn ?subtask ?result (op : operation) (t : Xenops_task.task_handle)
             ~path:(Uri.path url ^ snippet ^ id_str)
             ~query:(Uri.query url) ()
         in
-        let memory_url = make_url "/memory/" new_dest_id in
         (* CA-78365: set the memory dynamic range to a single value to stop
            ballooning. *)
         let atomic =
@@ -2437,20 +2454,29 @@ and perform_exn ?subtask ?result (op : operation) (t : Xenops_task.task_handle)
            memory on the receiver *)
         let state = B.VM.get_state vm in
         info "VM %s has memory_limit = %Ld" id state.Vm.memory_limit ;
-        Open_uri.with_open_uri memory_url (fun mem_fd ->
+        let url = make_url "/migrate/vm/" new_dest_id in
+        Open_uri.with_open_uri url (fun vm_fd ->
             let module Handshake = Xenops_migrate.Handshake in
             let do_request fd extra_cookies url =
               Sockopt.set_sock_keepalives fd ;
               let module Request =
                 Cohttp.Request.Make (Cohttp_posix_io.Unbuffered_IO) in
               let cookies =
-                [
-                  ("instance_id", instance_id)
-                ; ("final_id", id)
-                ; ("dbg", dbg)
-                ; (cookie_mem_migration, "")
-                ]
-                @ extra_cookies
+                List.concat
+                  [
+                    [
+                      ("instance_id", instance_id)
+                    ; ("final_id", id)
+                    ; ("dbg", dbg)
+                    ; (cookie_mem_migration, cookie_mem_migration_value)
+                    ]
+                  ; ( if compress_memory then
+                        [(cookie_mem_compression, cookie_mem_compression_value)]
+                    else
+                      []
+                    )
+                  ; extra_cookies
+                  ]
               in
               let headers =
                 Cohttp.Header.of_list
@@ -2465,11 +2491,11 @@ and perform_exn ?subtask ?result (op : operation) (t : Xenops_task.task_handle)
               in
               Request.write (fun _ -> ()) request fd
             in
-            do_request mem_fd
+            do_request vm_fd
               [("memory_limit", Int64.to_string state.Vm.memory_limit)]
-              memory_url ;
+              url ;
             let first_handshake () =
-              ( match Handshake.recv mem_fd with
+              ( match Handshake.recv vm_fd with
               | Handshake.Success ->
                   ()
               | Handshake.Error msg -> (
@@ -2489,9 +2515,9 @@ and perform_exn ?subtask ?result (op : operation) (t : Xenops_task.task_handle)
               debug "VM.migrate: Synchronisation point 1"
             in
             let final_handshake () =
-              Handshake.send ~verbose:true mem_fd Handshake.Success ;
+              Handshake.send vm_fd Handshake.Success ;
               debug "VM.migrate: Synchronisation point 3" ;
-              match Handshake.recv mem_fd with
+              match Handshake.recv vm_fd with
               | Success ->
                   debug "VM.migrate: Synchronisation point 4"
               | Error msg ->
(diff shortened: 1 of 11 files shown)


lindig:
Passed BVT and BST

psafont (approved):
Aside from the small nit, I thinkthe code looks good. I do believe that all documentation explaining the process of migration should be changed to reflect the added protocol, that is analogous  to the vgpu memory migration

ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -81,6 +83,17 @@ let filter_prefix prefix xs =
...
+let internal_error fmt =
+  Printf.kprintf (fun msg -> raise (Xenopsd_error (Internal_error msg))) fmt
+
+(* return last/element/in/a/path *)
+let basename path =
+  match Astring.String.cut ~sep:"/" ~rev:true path with
+  | Some (_, base) ->
+      base
+  | None ->
+      internal_error "invalid path: %s" path
```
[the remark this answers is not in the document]
lindig: The only reason I was not using it was that this is a path in an URL. There are small differences: this expects to find at least one `/`. `Filename.basename` will return `.` for the empty string. So it has slightly more specific semantics which would not hurt here, though.

ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -1312,6 +1334,12 @@ let vgpu_receiver_sync : (Vm.id, vgpu_fd_info) Hashtbl.t = Hashtbl.create 10
 
 let vgpu_receiver_sync_m = Mutex.create ()
 
+type mem_fd_info = {mem_fd: Unix.file_descr; mem_channel: unit Event.channel}
+
+let mem_receiver_sync : (Vm.id, mem_fd_info) Hashtbl.t = Hashtbl.create 10
+
+let mem_receiver_sync_m = Mutex.create ()
```
edwintorok, line 1341: It looks like modification is done using `with_lock` on this lock and the hashtable, but read is done without using the lock. I think that might still be unsafe: the hashtable might get modified as we read it, so the lock should be used in both cases, ideally the raw `mem_receiver_sync` should not be exposed and all accesses should be done through some wrapper functions that always take the lock.

Otherwise this may be a bit fragile and rely on an implementation detail: OCaml can currently only switch threads when a C stub call is made that relinquishes the runtime lock, OR when the called function allocates (and then the signal queued by the periodic ~50ms timer action gets processed that switches threads). I'm not sure we can rely on Hashtbl.find_opt never allocating (apart from the final `Some data`), and OCaml 5.0 would change the "only one OCaml thread may run at one time" if we start using the new Domains feature causing the hashtable to get mutated as we're searching it which might cause us not to find an entry that is otherwise there.

We probably already have this problem with various other global hashtables used in xenopsd, we should probably review its thread safety, regardless of multicore multiple threads are already being used in xenopsd, and it probably happens to work because *currently* Hashtbl.find_opt doesn't allocate during a search.

ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -3367,12 +3448,41 @@ module VM = struct
...
+            with_lock mem_receiver_sync_m (fun () ->
+                Hashtbl.add mem_receiver_sync vm info
+            ) ;
+            (* Inform the sender that everything is in place to start
+               save/restore *)
+            Xenops_migrate.(Handshake.send ~verbose:true fd Handshake.Success) ;
+            debug "VM.receive_mem: Synchronisation point 1-mem" ;
+            (* Keep the thread/connection open until the restore is done on the
+               other thread *)
+            Event.receive info.mem_channel |> Event.sync ;
```
edwintorok, line 3479: What if the migration fails, are we going to wait here forever, potentially "leaking" the file descriptor? I wonder whether this is causing the xenopsd file descriptor leak bug that we have elsewhere (not your new code, but the existing migration code in xenopsd)

ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -81,6 +83,17 @@ let filter_prefix prefix xs =
     )
     xs
 
+let internal_error fmt =
+  Printf.kprintf (fun msg -> raise (Xenopsd_error (Internal_error msg))) fmt
+
+(* return last/element/in/a/path *)
+let basename path =
+  match Astring.String.cut ~sep:"/" ~rev:true path with
```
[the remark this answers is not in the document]
psafont: See https://github.com/xapi-project/xen-api/pull/4741#discussion_r904821034

ocaml/xenopsd/lib/xenops_server.ml
lindig: I have rebased the code and added locking around reading the hash table.

lindig:
@robhoes can you take a look?

ocaml/xenopsd/lib/xenops_server.ml
robhoes: The signal from the main thread is sent in a finally handler that protects most of the code before it.

ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -2612,13 +2676,24 @@ and perform_exn ?subtask ?result (op : operation) (t : Xenops_task.task_handle)
...
+                    receive id s (* new handshake protocol *)
+                | None ->
+                    s
+                (* receiving memory on this conncection *)
+              in
               perform_atomics
                 (List.concat
                    [
-                     vgpu_start_operations
+                     vgpu_start_operations ()
```
robhoes, line 2690: Why does this need to become a function now?

ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -2592,9 +2642,23 @@ and perform_exn ?subtask ?result (op : operation) (t : Xenops_task.task_handle)
...
+                with_lock vgpu_receiver_sync_m (fun () ->
+                    Hashtbl.find_opt vgpu_receiver_sync id
+                )
+              in
               let pcis = PCI_DB.pcis id |> pci_plug_order in
-              let vgpu_start_operations =
+              let receive id s =
+                debug "using new handshake protocol for VM %s" id ;
+                Handshake.recv_success ~verbose:true s ;
+                match Hashtbl.find_opt mem_receiver_sync id with
```
robhoes, line 2654: `with_lock mem_receiver_sync_m`?

ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -2592,9 +2642,23 @@ and perform_exn ?subtask ?result (op : operation) (t : Xenops_task.task_handle)
...
+              let vgpu_info =
+                with_lock vgpu_receiver_sync_m (fun () ->
+                    Hashtbl.find_opt vgpu_receiver_sync id
+                )
+              in
               let pcis = PCI_DB.pcis id |> pci_plug_order in
-              let vgpu_start_operations =
+              let receive id s =
+                debug "using new handshake protocol for VM %s" id ;
+                Handshake.recv_success ~verbose:true s ;
```
robhoes, line 2653: Log `debug "VM.receive_memory: Synchronisation point 1-mem` after this?

ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -2592,9 +2642,23 @@ and perform_exn ?subtask ?result (op : operation) (t : Xenops_task.task_handle)
...
               (* Check if there is a separate vGPU data channel *)
-              let vgpu_info = Hashtbl.find_opt vgpu_receiver_sync id in
+              let vgpu_info =
+                with_lock vgpu_receiver_sync_m (fun () ->
+                    Hashtbl.find_opt vgpu_receiver_sync id
+                )
+              in
               let pcis = PCI_DB.pcis id |> pci_plug_order in
-              let vgpu_start_operations =
+              let receive id s =
```
robhoes, line 2651: The function name could be clearer. What is being received?

ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -2612,13 +2676,24 @@ and perform_exn ?subtask ?result (op : operation) (t : Xenops_task.task_handle)
                       ; [VGPU_start (vgpus, true)]
                       ]
               in
+              let mem_fd =
```
robhoes, line 2679: Here, `mem_fd` _is_ accurately referring to the fd that the memory is transferred over.

ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -2612,13 +2676,24 @@ and perform_exn ?subtask ?result (op : operation) (t : Xenops_task.task_handle)
...
+                match handshake with
+                | Some _ ->
+                    receive id s (* new handshake protocol *)
+                | None ->
+                    s
+                (* receiving memory on this conncection *)
```
robhoes, line 2685: "conncection" -> "connection"

ocaml/xenopsd/lib/xenops_server.ml
robhoes: I think you forgot one: see below.

robhoes (commented):
Overall this looks good, and you have show with stress tests that it is indeed working. I've made some comments about little things that can be improved. More generally, while this seems to be the smallest amount of change to the existing code to achieve the goal, there may be ways to improve the structure of the main `VM_migrate` and `VM_migrate_receive` blocks, to make it easier to follow (though I can't immediately see how).

ocaml/xenopsd/lib/xenops_server.ml
lindig: There is no particular reason except I became a bit nervous that this is evaluated before it is used and given the complexity of the value would like to make sure it is constructed in the right context.

ocaml/xenopsd/lib/xenopsd.ml
```diff
@@ -347,6 +347,7 @@ let handle_received_fd this_connection =
       let common_prefix = "/services/xenops/" in
       let memory_prefix = common_prefix ^ "memory/" in
```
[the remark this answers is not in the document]
robhoes: I think `migrate-mem/` and `migrate-vgpu/` are fine. I just meant replacing `memory/` with `migrate/`.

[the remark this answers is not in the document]
robhoes: You could then even decide to drop the cookie if you receive `migrate/` rather than `memory/`.

lindig:
I have updated the code in response to the reviews but also included the code for compression because the two are a bit too intertwined to keep them separate after I made the requested changes. So the PR is now quite a bit bigger. By default compression is not used but can be either requested explicitly (`compress=true` option in API and CLI) or the default can be changed in `xapi.conf`.

ocaml/xenopsd/lib/xenops_migrate.ml
```diff
@@ -42,7 +42,7 @@ module Handshake = struct
     if n < len then really_read fd buf (ofs + n) (len - n)
 
   (** Receive a 'result' from the remote *)
-  let recv ?(verbose = false) (s : Unix.file_descr) : result =
```
robhoes, line 45: Did you mean to leave verbose-mode on? You may have done this for testing, and may get overly spammy.

lindig: I'll dial it back. It is more important during development.

edwintorok (commented):
To future proof this could we add the compression algorithm to the cookie (`zstd`) and the version of XAPI on the source side? We've had bugs in the past that were unfixable for migration from an old version either because we would've had to fix the bug on the old source, or because we were not able to disambiguate between 2 versions of xapi, one with a bug and one without and had to rely on various other hacks to try and identify the version.
I think we have some kind of version number in the metadata, but we don't bump it often enough.

lindig:
Latest changes in response to review:
* dialed back handshake logging
* Using more useful values in the cookies: handshake uses xapi build, compression uses "zstd". The handshake one could prove useful in the future because it will tell us the exact version that we are migrating from.

PR #4742, opened 2022-06-28
Speed up builds: avoid rebuilding on git commit hash/date changes

edwintorok:
Use dune's promotion mechanism, more info here: https://dune.readthedocs.io/en/stable/concepts.html#diffing-and-promotion

It now takes ~3s to do a `make build` the 2nd time around when there are no changes, and it used to take ~a minute previously (since changing the datamodel lifecycle triggered a rebuild of everything, even if its contents hasn't actually changed).

Draft PR because this also needs testing with .spec builds.


edwintorok:
Looks like this needs some more work for it to work correctly when building outside a git repository and when building from a commit hash without tags.

psafont:
The failure to build is because somehow dune-build-info does not read the version field from `dune-project` and creates a version from the hash version (which does not contain a correct version number because github does a shallow clone)

psafont:
xs-opam is happy with my branch as well: https://github.com/psafont/xs-opam/runs/7990341646

It only needs fixing for xen-api's github CI

robhoes:
This has been superseded by #4779

PR #4743, opened 2022-06-28
CP-39640/CP-39157 Add stream compression for VM migration

lindig:
Optionally compress the memory byte stream that is coming out of
emu-manager as part of a VM migration and extend API and CLI for that
option. The default (when no option is used) is controlled via xapi.conf
and the current default is to use compression for non-local migration.

Compression is implemented by feeding the byte stream through a
compression binary that we spawn with Forkexed. Currently we are using
zstd to achieve high troughput. Compression isbenefical if otherwise the
network connection becomes saturated which often will require multiple
migrations in parallel.

At the API/CLI level compression is enabled by using the option
"compress=true", disabled by "compression=false" and using the default
(controlled via xapi.conf) otherwise.

The sending side signals to the receiving side the use of compression by
using a HTTP cookie key-value pair with "compress=true".

### DIFF
--- a/ocaml/idl/datamodel_vm.ml
+++ b/ocaml/idl/datamodel_vm.ml
@@ -988,7 +988,12 @@ let pool_migrate =
       [
         (Ref _vm, "vm", "The VM to migrate")
       ; (Ref _host, "host", "The target host")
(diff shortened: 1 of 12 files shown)


lindig:
Will move this to a different PR

PR #4744, opened 2022-07-01
CP-39894: move xenopsd's daemon modules from device to service

psafont:
This mirrors the changes in https://github.com/xapi-project/xen-api/pull/4720 except the code that is used exclusively by the swtpm daemon, mainly the code introduced by this commit and successive modifications: [`500bbab` (#4720)](https://github.com/xapi-project/xen-api/pull/4720/commits/500bbab1e5e0bac85cf9ac73972cabb1ef8e2bf1), which I cherry picked, the modified to remove the Swtpm module and functions it depends on (because they would have been unused)

It's pretty to notice the differences between this and the swtpm branch are the removal of functions needed by the swtpm daemon; and the removal of File  in the variant `Pid.path`, which would be unused, and consequent change in behaviour in the uses of this variant.

I used `git diff feature/vtpm..private/paus/pidpaths -- ocaml/xenopsd/xc/service.ml`

For device.ml there are more differences, all the ones I found were cause because of the addition of the vtpm parameter and swtpm management as well.

Ring3 BST + BVT SR: 170941 ✅

### DIFF
--- a/ocaml/xenopsd/lib/xenops_utils.ml
+++ b/ocaml/xenopsd/lib/xenops_utils.ml
@@ -44,6 +44,23 @@ module Unix = struct
       }
 end
 
+(** Represent an IPC endpoint *)
+module Socket = struct
+  type t = Unix of string | Port of int
+
+  module Unix = struct
+    let path x = "unix:" ^ x
+
+    let rm x =
+      let dbg = debug "error cleaning unix socket %s: %s" x in
+      try Unix.unlink x with
+      | Unix.Unix_error (Unix.ENOENT, _, _) ->
+          ()
+      | Unix.Unix_error (e, _, _) ->
+          dbg (Unix.error_message e)
+  end
+end
+
 let all = List.fold_left ( && ) true
 
 let any = List.fold_left ( || ) false
@@ -632,6 +649,12 @@ let chunks size lst =
     [] lst
   |> List.map (fun xs -> List.rev xs)
   |> List.rev
+
+let really_kill pid =
+  try Unixext.kill_and_wait pid
+  with Unixext.Process_still_alive ->
+    debug "%d: failed to respond to SIGTERM, sending SIGKILL" pid ;
+    Unixext.kill_and_wait ~signal:Sys.sigkill pid
 
 let best_effort txt f =
   try f ()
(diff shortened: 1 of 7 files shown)


ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -6,13 +6,13 @@ let prototyped_of_field = function
   | "Repository", "gpgkey_path" ->
       Some "22.12.0"
   | "host", "last_software_update" ->
-      Some "22.19.0-next"
```
lindig, line 9: Is this related or accidental?

ocaml/xenopsd/lib/xenops_utils.ml
```diff
@@ -44,6 +44,23 @@ module Unix = struct
...
 
+(** Represent an IPC endpoint *)
+module Socket = struct
+  type t = Unix of string | Port of int
+
+  module Unix = struct
+    let path x = "unix:" ^ x
+
+    let rm x =
+      let dbg = debug "error cleaning unix socket %s: %s" x in
```
lindig, line 55: I think not having an explicit parameter is slightly confusing.

ocaml/xenopsd/xc/service.mli
```diff
@@ -0,0 +1,113 @@
...
+    -> restore:bool
+    -> Xenops_task.Xenops_task.task_handle
+    -> int
+    -> unit
+
+  val cancel_key : Xenctrl.domid -> Cancel_utils.key
+
+  val state_path : Xenctrl.domid -> string
+
+  val pid : xs:Xenstore.Xs.xsh -> Xenctrl.domid -> int option
```
lindig, line 66: Does a `None` result imply that `is_running` is false? This would suggest that `is_running` is a simple test for `pid <> None`.

lindig (approved):

ocaml/idl/datamodel_lifecycle.ml
psafont: This is a change that is forced by the versioning mechanism we have, everytime the code is compile this source file is changes, which might make it to unrelated commit if we're not careful, hence why I added it a separate commit. This is the same change as in Rob's PR

ocaml/xenopsd/xc/service.mli
psafont: Not exactly, is_running is:
```
  let is_running ~xs domid =
    match pid ~xs domid with
    | None ->
        false
    | Some p -> (
      try Unix.kill p 0 ; (* This checks the existence of pid p *)
                          true
      with _ -> false
    )
```

ocaml/idl/datamodel_lifecycle.ml
robhoes: Yeah, it was actually in 22.20.0 but got missed. I've got the right version in my PR, which was merged, so this change should now be removed.

edwintorok (approved):

PR #4745, opened 2022-07-01
CA-365946: Block VIF and VBD hotplug into dom0

robhoes:


ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -6,13 +6,13 @@ let prototyped_of_field = function
   | "Repository", "gpgkey_path" ->
       Some "22.12.0"
   | "host", "last_software_update" ->
-      Some "22.19.0-next"
```
lindig, line 9: Is this expected to be part of this PR? It seems unrelated.

ocaml/xapi/xapi_vbd_helpers.ml
```diff
@@ -95,6 +95,11 @@ let valid_operations ~expensive_sharing_checks ~__context record _ref' : table =
     set_errors Api_errors.other_operation_in_progress
       ["VBD"; _ref; vbd_operation_to_string (List.hd current_ops)]
       [`unpause] ;
+  (* No hotplug on dom0 *)
+  if Helpers.is_domain_zero ~__context vm then
+    set_errors Api_errors.operation_not_allowed
+      ["Control domain does not support hotplug"]
```
lindig, line 101: Should this be `does not support hot plugging`? Also below.

ocaml/xapi/xapi_xenops.ml
```diff
@@ -4221,21 +4225,25 @@ let vif_unplug ~__context ~self force =
...
+        |> sync_with_task __context queue_name ;
+        (* We need to make sure VIF.stat still works so: wait before calling VIF.remove *)
+        Events_from_xenopsd.wait queue_name dbg (fst vif.Vif.id) () ;
+        if Db.VIF.get_currently_attached ~__context ~self then
+          raise
+            Api_errors.(
+              Server_error
+                ( internal_error
+                , [
+                    Printf.sprintf "vif_unplug: Unable to unplug VIF %s"
```
lindig, line 4239: This can also use `__FUNCTION__` to help finding the origin of the message.

lindig (approved):

ocaml/tests/common/test_common.ml
```diff
@@ -144,20 +144,34 @@ let make_vm ~__context ?(name_label = "name_label")
...
+  match power_state with
+  | `Running | `Paused ->
+      Db.VM.set_power_state ~__context ~self:vm ~value:power_state ;
+      Option.iter
+        (fun host -> Db.VM.set_resident_on ~__context ~self:vm ~value:host)
+        resident_on ;
+      let vmm = Db.VM.get_metrics ~__context ~self:vm in
+      Db.VM_metrics.set_current_domain_type ~__context ~self:vmm
+        ~value:domain_type ;
+      vm
```
psafont, line 172: This seems unrelated to the rest of the changes, why is updating the power state of the VM, resident_on and metrics needed here, only for vm that have been told to run or pause?

ocaml/idl/datamodel_lifecycle.ml
robhoes: We just do these updates along the way as the changes are normally automatic.

psafont (approved):

ocaml/xapi/xapi_vbd_helpers.ml
robhoes: We normally write "hotplug" rather than "hot plug". Regarding "-ing", I'm not convinced...

ocaml/xapi/xapi_xenops.ml
robhoes: That may not be appropriate in an API error; I think it's more useful for logging. I could add it to the new log line below, but if we thing this is useful, shouldn't we add it to the log library itself rather than having to manually add it to every log line?

PR #4746, opened 2022-07-04
Fast resume a VM when its migration has failed

benjamreis:

### DIFF
--- a/ocaml/xapi/xapi_vm_migrate.ml
+++ b/ocaml/xapi/xapi_vm_migrate.ml
@@ -387,8 +387,8 @@ let pool_migrate ~__context ~vm ~host ~options =
         ( try
             let _, state = XenopsAPI.VM.stat dbg vm_uuid in
             if Xenops_interface.(state.Vm.power_state = Suspended) then (
-              debug "xenops: %s: shutting down suspended VM" vm_uuid ;
-              Xapi_xenops.shutdown ~__context ~self:vm None
+              debug "xenops: %s: Resuming suspended VM on source host" vm_uuid ;
+              Xapi_xenops.fast_resume ~__context ~self:vm
             )
           with _ -> ()
         ) ;
@@ -1549,8 +1549,8 @@ let migrate_send' ~__context ~vm ~dest ~live:_ ~vdi_map ~vif_map ~vgpu_map
         try
           let _, state = XenopsAPI.VM.stat dbg vm_uuid in
           if Xenops_interface.(state.Vm.power_state = Suspended) then (
-            debug "xenops: %s: shutting down suspended VM" vm_uuid ;
-            Xapi_xenops.shutdown ~__context ~self:vm None
+            debug "xenops: %s: Resuming suspended VM on source host" vm_uuid ;
+            Xapi_xenops.fast_resume ~__context ~self:vm
           )
         with _ -> ()
     ) ;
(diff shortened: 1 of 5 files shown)


ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -1549,8 +1549,8 @@ let migrate_send' ~__context ~vm ~dest ~live:_ ~vdi_map ~vif_map ~vgpu_map
         try
           let _, state = XenopsAPI.VM.stat dbg vm_uuid in
           if Xenops_interface.(state.Vm.power_state = Suspended) then (
-            debug "xenops: %s: shutting down suspended VM" vm_uuid ;
-            Xapi_xenops.shutdown ~__context ~self:vm None
+            debug "xenops: %s: Resuming suspended VM on source host" vm_uuid ;
+            Xapi_xenops.fast_resume ~__context ~self:vm
           )
         with _ -> ()
```
edwintorok, line 1555: It would be useful to log any exceptions here, perhaps at debug level instead of silently eating it up if we want to do anything other than killing the VM here.

ocaml/xenopsd/xc/domain.ml
```diff
@@ -2000,3 +2000,10 @@ let move_xstree ~xs domid olduuid newuuid =
     )
   in
   List.iter mv_tree search_paths
+
+(** Request a fast_resume, return without waiting for acknowledgement *)
+let fast_resume ~xc ~xs domid =
+  let uuid = get_uuid ~xc domid in
+  debug "VM = %s; domid = %d; Requesting domain fast resume"
+    (Uuid.to_string uuid) domid ;
+  Xenctrl.domain_resume_fast xc domid
```
edwintorok, line 2009: We might want to check whether the guest supports fast resume or not before attempting this, see the comments in Xen's `xg_resume.c` (Unfortunately it doesn't document how it can be determined whether a guest supports this or not, but presumably it can advertise that in xenstore in some way? although the `xl` doesn't seem to check for it either). Perhaps just add a TODO comment till we figure it out?

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -1549,8 +1549,8 @@ let migrate_send' ~__context ~vm ~dest ~live:_ ~vdi_map ~vif_map ~vgpu_map
         try
           let _, state = XenopsAPI.VM.stat dbg vm_uuid in
           if Xenops_interface.(state.Vm.power_state = Suspended) then (
-            debug "xenops: %s: shutting down suspended VM" vm_uuid ;
-            Xapi_xenops.shutdown ~__context ~self:vm None
+            debug "xenops: %s: Resuming suspended VM on source host" vm_uuid ;
+            Xapi_xenops.fast_resume ~__context ~self:vm
```
edwintorok, line 1553: This will need to distinguish whether we've already force reset the lifecycle state above (and that is what failed, in which case we can't resume or somthing earlier).
So probably need to move this `with` right before the 'force_state_reset` `if` condition above. And have another `try/with` that does the force shutdown as before.

We might also need to introduce a new synchronization point where the other end can tell us whether it failed to resume the VM there, so that it can be resumed on the source, and we need to wait for that before attempting to destroy the VM on the source.

lindig:
That's a very desirable feature - thanks for working on it

ocaml/xapi/xapi_vm_migrate.ml
edwintorok: Yes, that would be good to avoid leaving a VM around in a broken state.

lindig:
What is the status? It's fine to open a PR as a draft for early feedback but I would like to see it progress or otherwise be closed.

benjamreis:
Hi!

I was on hollydays the past weeks so that's why this PR has been inactive.
I've taken the comments into account so I removed my changes locally in order to mutualize the modification within `xenopsd`.

robhoes:
This PR currently doesn't have any content, so I'll close it. Thanks for looking into it :)

PR #4750, opened 2022-07-06
Factor out Throttle module

lindig:
We use the Throttle module to execute a function multiple times in
parallel up to a configured number. This generalises the existing module
and makes it accessible in other contexts.

### DIFF
--- a/ocaml/xapi/xapi_sr.ml
+++ b/ocaml/xapi/xapi_sr.ml
@@ -61,7 +61,7 @@ let scan_finished sr =
       Condition.broadcast scans_in_progress_c
   )
 
-module Size = struct let n = !Xapi_globs.max_active_sr_scans end
+module Size = struct let n () = !Xapi_globs.max_active_sr_scans end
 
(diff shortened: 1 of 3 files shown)


ocaml/xapi/xapi_sr.ml
```diff
@@ -61,30 +61,10 @@ let scan_finished sr =
...
-            result
-        | Some s ->
-            s
-    )
-
-  let execute f = Semaphore.execute (get_semaphore ()) f
-end
-
-module AutoScanThrottle = Throttle ()
+module Size = struct let n = !Xapi_globs.max_active_sr_scans end
```
edwintorok, line 64: This will evaluate Xapi_globs.max_active_sr_scans too early, before the config file or cmdline is parsed (it'll run as part of module startup code).
I think we need to delay evaluation of this by making it `unit -> int`.

lindig:
This is optional now because I no longer need the Throttle module in another context. I still think it makes sense to expose it, though.

robhoes (approved):

PR #4752, opened 2022-07-07
VTPM: rework datamodel, fix nits needed by clients / detected by QA

psafont:
Replace the profile with single fields, which as of now cannot be changed. Unique can be made configured later on.

This means deleting the default_vtpm_profile from the VM objects. The way the design is going it looks like "empty" or otherwise vtpms associated will replace this parameter, which contain this information and more.

vtpm.create's role has been changed and it's now similat to other classes like vgpu so it uses individual fields as parameters instead of whole records. This is because VTPM records cannot be easily created by clients. (for comparison VM _does_ have a create method, but it has big warnings against using it and the default way to create VMs is by cloning templates, both in Xencenter and Xen Orchestra)

VTPM.backend has been reintroduced for internal purposes only

The contents of the VTPM are now only encoded once for serialization purposes. There's still a second encoding, but it's used for verification purposes and the result is discarded.

New builds are in progress

### DIFF
--- a/ocaml/idl/datamodel_vm.ml
+++ b/ocaml/idl/datamodel_vm.ml
@@ -2118,11 +2118,6 @@ let t =
             ~ty:(Set update_guidances) "pending_guidances"
             ~default_value:(Some (VSet []))
             "The set of pending guidances after applying updates"
-        ; field ~qualifier:StaticRO ~in_product_since:rel_next
-            ~ty:(Map (String, String))
-            ~default_value:(Some (VMap [])) "default_vtpm_profile"
-            "The security properties that will be used by default when \
-             creating a TPMs attached to the VM / template"
         ]
       )
     ()
--- a/ocaml/idl/datamodel_vtpm.ml
+++ b/ocaml/idl/datamodel_vtpm.ml
@@ -26,11 +26,15 @@ let create =
         (Published, rel_rio, "")
       ; ( Changed
         , rel_next
-        , "Require only a VM reference to create a VTPM instance"
+        , "Require only a VM reference and uniqueness to create a VTPM instance"
         )
       ]
     ~doc:"Create a new VTPM instance, and return its handle."
-    ~params:[(Ref _vm, "vM", "The VM reference the VTPM will be attached to")]
+    ~params:
+      [
+        (Ref _vm, "vM", "The VM reference the VTPM will be attached to")
+      ; (Bool, "is_unique", "Whether the VTPM must be unique")
+      ]
     ~result:(Ref _vtpm, "The reference of the newly created VTPM")
     ~allowed_roles:_R_VM_ADMIN ()
 
@@ -63,7 +67,7 @@ let t =
       [
         (Published, rel_rio, "Added VTPM stub")
       ; (Extended, rel_next, "Added ability to manipulate contents")
-      ; (Extended, rel_next, "Added VTPM profiles")
+      ; (Extended, rel_next, "Added VTPM unique and protected properties")
       ; (Extended, rel_next, "Added Persistence backed")
       ]
     ~gen_constructor_destructor:false ~name:_vtpm ~descr:"A virtual TPM device"
@@ -80,10 +84,13 @@ let t =
       ; field ~qualifier:DynamicRO ~ty:persistence_backend
           ~default_value:(Some (VEnum "xapi")) ~lifecycle:[]
           "persistence_backend" "The backend where the vTPM is persisted"
-      ; field ~qualifier:DynamicRO
-          ~ty:(Map (String, String))
-          ~lifecycle:[(Published, rel_next, "Added VTPM profiles")]
-          "profile" "The security properties that define how the TPM is handled"
+      ; field ~qualifier:StaticRO ~ty:Bool ~default_value:(Some (VBool false))
+          ~lifecycle:[] "is_unique"
+          "Whether the contents are never copied, satisfying the TPM spec"
+      ; field ~qualifier:DynamicRO ~ty:Bool ~default_value:(Some (VBool false))
+          ~lifecycle:[] "is_protected"
+          "Whether the contents of the VTPM are secured according to the TPM \
+           spec"
       ; field ~qualifier:DynamicRO ~ty:(Ref _secret) ~internal_only:true
           ~lifecycle:[(Published, rel_next, "Added VTPM contents")]
           "contents" "The contents of the TPM"
--- a/ocaml/idl/datamodel_vtpm.ml
+++ b/ocaml/idl/datamodel_vtpm.ml
@@ -92,8 +92,7 @@ let t =
           "Whether the contents of the VTPM are secured according to the TPM \
            spec"
       ; field ~qualifier:DynamicRO ~ty:(Ref _secret) ~internal_only:true
-          ~lifecycle:[(Published, rel_next, "Added VTPM contents")]
-          "contents" "The contents of the TPM"
+          ~lifecycle:[] "contents" "The contents of the TPM"
       ]
     ~messages:[create; destroy; get_contents; set_contents]
     ()
--- a/ocaml/idl/datamodel_vtpm.ml
+++ b/ocaml/idl/datamodel_vtpm.ml
@@ -16,6 +16,27 @@ open Datamodel_types
 open Datamodel_common
 open Datamodel_roles
 
+let create =
+  call ~name:"create"
+    ~lifecycle:
+      [
+        (Published, rel_rio, "")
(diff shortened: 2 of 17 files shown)


ocaml/xapi/xapi_vtpm.ml
```diff
@@ -45,11 +45,16 @@ let destroy ~__context ~self =
 
 let get_contents ~__context ~self =
   let secret = Db.VTPM.get_contents ~__context ~self in
-  Base64.decode_exn (Db.Secret.get_value ~__context ~self:secret)
+  Db.Secret.get_value ~__context ~self:secret
```
edwintorok, line 45: This is a breaking change for anyone who has had VM's saved with the previous encoding, but that is probably just me at the moment, and I can probably manually update the DB following this change (otherwise we'd need  DB upgrade rule, but probably not worth writing one now)

ocaml/idl/datamodel_vtpm.ml
```diff
@@ -61,7 +64,10 @@ let t =
         (Published, rel_rio, "Added VTPM stub")
       ; (Extended, rel_next, "Added ability to manipulate contents")
       ; (Extended, rel_next, "Added VTPM profiles")
-      ; (Changed, rel_next, "Removed backend field")
```
edwintorok, line 43: I think we should keep the 'backend' field removed (it used to mean "domain" and had a different type), might avoid issues with DB schema upgrade.

And you've already called the backend 'persistence_backend' above in the type, so we should make the field name match as well.

ocaml/idl/datamodel_vm.ml
```diff
@@ -2118,11 +2118,6 @@ let t =
             ~ty:(Set update_guidances) "pending_guidances"
             ~default_value:(Some (VSet []))
             "The set of pending guidances after applying updates"
-        ; field ~qualifier:StaticRO ~in_product_since:rel_next
```
edwintorok, line 2121: This never reached rt-next, so this is fine to remove.

ocaml/xapi/xapi_vtpm.ml
```diff
@@ -20,20 +20,16 @@ let assert_no_vtpm_associated ~__context vm =
...
   assert_no_vtpm_associated ~__context vM ;
   Xapi_vm_lifecycle.assert_initial_power_state_is ~__context ~self:vM
     ~expected:`Halted ;
+  let ref = Ref.make () in
   let uuid = Uuid.(to_string (make ())) in
-  let profile = Db.VM.get_default_vtpm_profile ~__context ~self:vM in
   let backend = `xapi in
   let contents = Xapi_secret.create ~__context ~value:"" ~other_config:[] in
-  let ref = introduce ~__context ~uuid ~vM ~backend ~profile ~contents in
+  Db.VTPM.create ~__context ~ref ~uuid ~vM ~backend ~unique:false
```
edwintorok, line 31: We should add unique and protected as parameters to create, defaulting to false if absent.

psafont: While I agree on the unique field, I do not think it makes sense to give control to users of protected. It's a property of how the system handles the vtpm contents and should be read-only.

For the unique field I expected to defer this until it's properly implemented. I can change the constructor and ignore the value of the field for the time being.

edwintorok: They are readonly after the object is created, but how else would the user chose whether they want a protected vTPM or not if not at create time?

edwintorok (approved):

ocaml/xapi/xapi_vtpm.ml
psafont: I would say that the whole point of TPMs is to have some secrets protected, so I don't see the point in making users choose. I see it as a way to report the system property.

(I'm working on exposing the unique field in the create)

psafont: I've added the unique parameter, there's no easy way to have it as an option: having option types in `create` functions is disallowed by the api generator, and because a unit parameter needs to be added to remove ambiguity cause by partial application.

I believe this is not such a bad compromise:
- for backwards compatibility we have the platform metadata in templates, which will always create non-unique vtpms, this is fine because this is wanted only to be able to install / boot windows 11 guests.
- New clients will need an updated SDK to create VTPMs anyway, this will make the option obvious to them and can make choices for their users.

ocaml/idl/datamodel_vtpm.ml
```diff
@@ -39,24 +67,30 @@ let t =
...
     ~contents:
       [
         uid _vtpm
       ; field ~qualifier:StaticRO ~ty:(Ref _vm) "VM"
           "The virtual machine the TPM is attached to"
-      ; field ~qualifier:DynamicRO
-          ~ty:(Map (String, String))
-          ~lifecycle:[(Published, rel_next, "Added VTPM profiles")]
-          "profile" "The security properties that define how the TPM is handled"
+      ; field ~qualifier:StaticRO ~ty:(Ref _vm) "backend"
```
robhoes, line 81: Even though you are explicit defining the `create` function, please keep the qualifier's in sync with its arguments, just to avoid confusion. So `VM` and `unique` should be StaticRO, and all others DynamicRO.

ocaml/xapi/xapi_vtpm.ml
edwintorok: Depending on how we implement protection (e.g. per-host or per-pool encryption key) then a VM exported from a pool will only be reimportable on the same pool (unless you provide the encryption keys in some way, but those should be non-exportable...). 
This is fine if you've actually stored secrets in the vTPM, however if you only want to share a Windows11 VM image between multiple pools, which has a vTPM (not for security purposes, but purely to satisfy Win11 requirements), then having it marked as protected will make that quite difficult (in particular none of the existing tooling/applications will work with it because they wouldn't know a key has to be supplied or how to supply it).
In that case you might want to create the VM with unprotected vTPM as a functionality/security tradeoff.

*or* we should provide a way to export a "wiped"/empty vTPM/vTPM with just public certs, in case a protected vTPM is being exported, acknowledging that this may lead to data loss if the VM has actually stored some secrets or data inside the vTPM.

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -7845,7 +7845,14 @@ module VTPM = struct
   let create printer rpc session_id params =
     let vm_uuid = List.assoc "vm-uuid" params in
     let vM = Client.VM.get_by_uuid ~rpc ~session_id ~uuid:vm_uuid in
-    let ref = Client.VTPM.create ~rpc ~session_id ~vM in
+    let unique =
+      match List.assoc_opt "unique" params with
+      | Some value ->
+          Stdlib.bool_of_string value
```
lindig, line 7851: This can fail. Is this string user supplied? Likewise instances below. It might be safer to use a function that catches exceptions.

lindig (approved):

ocaml/xapi-cli-server/cli_operations.ml
psafont: I've replaced it with the module's own 'bool_of_string' to have a better error

ocaml/idl/datamodel_vtpm.ml
```diff
@@ -39,24 +67,29 @@ let t =
...
-          ~lifecycle:[(Published, rel_next, "Added VTPM profiles")]
-          "profile" "The security properties that define how the TPM is handled"
+      ; field ~qualifier:DynamicRO ~ty:(Ref _vm) "backend"
+          "The domain where the backend is located (unused)"
+      ; field ~qualifier:DynamicRO ~ty:persistence_backend ~lifecycle:[]
+          "persistence_backend" "The backend where the vTPM is persisted"
+      ; field ~qualifier:StaticRO ~ty:Bool ~default_value:(Some (VBool false))
+          ~lifecycle:[] "unique"
+          "Whether the contents are never copied, satisfying the TPM spec"
+      ; field ~qualifier:DynamicRO ~ty:Bool ~lifecycle:[] "protected"
```
edwintorok, line 88: There is code in XAPI that would check (see dm_api.ml) that all fields that are not in `rel_rio` have default values, which are required for DB upgrades to work.
However because the object is `rel_rio` I think all fields inherit that. But none of these new fields existed in `rel_rio`, so I think we need to add an explicit `in_product_since` definition to each field that we've added since Yangtze.

edwintorok: This can be fixed in datamodel itself by setting in_product_since to None if lifecycle is None, so we don't need to do anything about Rio here for now.

PR #4753, opened 2022-07-08
SDK fixes (within CP-37225 and CP-37091)

kc284:
Best reviewed with whitespace diff off.


CitrixChris (approved):

danilo-delbusso (approved):
compiles, works, and all that jazz. Thanks for waiting :D

ocaml/sdk-gen/java/autogen/xen-api-samples/src/main/resources/README.txt
```diff
@@ -49,3 +49,22 @@ To run the tests
 Once you compile the library run:
 
 RunTests <host> <username> <password> [nfs server] [nfs path]
+
+Before running, you may need to perform these steps:
+
+1. Copy to your client machine the certificate /etc/xensource/xapi-ssl.pem of
```
edwintorok, line 55: That includes both the public and private key of the server, all you need to connect to the server is the public key, and the private key shouldn't ever be copied off the server: it allows you to decrypt all (future) traffic (past traffic should be safe due to use of ephemeral key exchange).

edwintorok: It would be safer to run this on the server:
```
openssl x509 -in /etc/xensource/xapi-ssl.pem -pubkey -out serverpub.pem
```

edwintorok (commented):
One minor comment on copying certificates in the docs, but looks ok otherwise.

edwintorok (approved):

PR #4755, opened 2022-07-12
CP-40155 Parallelize Host.evacuate

lindig:
Host.evacuate migrates all VMs from a host to another in preparation of
a host update. This is currently done sequentially, one VM after
another. We want to migrate VMs in parallel to reduce total elapsed
time.

Because a host can host hundreds of VMs, we want to limit
parallelization such that at most n migrations are active at the same
time. We implement this using batching: a fixed number of VMs is
migrated off in parallel before the next batch is started. This is
controlled by a new parameter evacuation-batch-size in xapi.conf.

I suspect that this is leaking tasks but I don't understand why this is happening
and would like to ask reviewers to have an eye towards that. Otherwise the
code works well in the tests that I have run where I continuously evacuate hosts in a pool.

### DIFF
--- a/ocaml/xapi/xapi_globs.ml
+++ b/ocaml/xapi/xapi_globs.ml
@@ -926,6 +926,8 @@ let repository_gpgcheck = ref true
 
 let migration_compression = ref false
 
+let evacuation_batch_size = ref 10
+
 type xapi_globs_spec_ty = Float of float ref | Int of int ref
 
 let extauth_ad_backend = ref "winbind"
@@ -1354,6 +1356,11 @@ let other_options =
     , Arg.Set migration_compression
     , (fun () -> string_of_bool !migration_compression)
     , "Use compression during VM migration when no API option provided."
+    )
+  ; ( "evacuation-batch-size"
+    , Arg.Set_int evacuation_batch_size
+    , (fun () -> string_of_int !evacuation_batch_size)
+    , "The number of VMs evacauted from a host in parallel."
     )
   ]
 
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -27,6 +27,21 @@ module D = Debug.Make (struct let name = "xapi_host" end)
 
 open D
 
+(* [take n xs] returns the first [n] elements of [xs] and the remaining
+   tail. Similar to Listext.chop but never raises an exception even if
+   n <= 0 or n > |xs|. *)
+let take n xs =
+  let rec loop n head tail =
+    match tail with
+    | [] ->
+        (List.rev head, tail)
+    | tail when n <= 0 ->
+        (List.rev head, tail)
+    | t :: ts ->
+        loop (n - 1) (t :: head) ts
+  in
+  loop n [] xs
+
 let get_servertime ~__context ~host:_ = Date.of_float (Unix.gettimeofday ())
 
 let get_server_localtime ~__context ~host:_ = Date.localtime ()
@@ -584,8 +599,8 @@ let compute_evacuation_plan ~__context ~host =
         compute_evacuation_plan_no_wlb ~__context ~host ()
 
 let evacuate ~__context ~host ~network =
-  let task = Context.get_task_id __context in
   let plans = compute_evacuation_plan ~__context ~host in
+  let plans_length = float (Hashtbl.length plans) in
   (* Check there are no errors in this list *)
   Hashtbl.iter
     (fun _ plan ->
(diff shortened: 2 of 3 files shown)


ocaml/xapi/xapi_host.ml
```diff
@@ -598,46 +612,73 @@ let evacuate ~__context ~host ~network =
...
+              []
+          in
+          let options = ("live", "true") :: with_network_option in
+          Client.Client.Async.VM.pool_migrate ~rpc ~session_id ~vm ~host
+            ~options
+        with
+        | Api_errors.Server_error (code, _) as e
           when code = Api_errors.vm_bad_power_state ->
-            ()
+            raise e
```
robhoes, line 643: Why?

ocaml/xapi/xapi_host.ml
```diff
@@ -27,6 +27,20 @@ module D = Debug.Make (struct let name = "xapi_host" end)
 
 open D
 
+(* [take n xs] returns the first [n] elements of [xs] and the remaining
+   tail. *)
+let take n xs =
```
psafont, line 32: `Xapi_stdext_std.Listext.List.take` is available as well.

It's mind-boggling how this is not on the standard library yet, on previous attempts, inclusion was struck down stating it was "too niche"

ocaml/xapi/xapi_host.ml
```diff
@@ -598,46 +612,93 @@ let evacuate ~__context ~host ~network =
...
+                      ~__context ~self:network ~host
+             ) ;
+          Client.Client.Async.VM.pool_migrate ~rpc ~session_id ~vm ~host
+            ~options
+      | Error (code, params) ->
+          (* should never happen *)
+          raise (Api_errors.Server_error (code, params))
+    in
+    let progress = Db.Task.get_progress ~__context ~self:task in
+    TaskHelper.set_progress ~__context (progress +. individual_progress) ;
```
robhoes, line 638: This now assigns progress immediately after the async call returns rather than at the end of the migration.

lindig: That's true. Maybe progress it after each batch has completed?

lindig:
I've addressed the helpful comments:

* Tasks are destroyed and no longer leaking
* The results of tasks are checked but 'power state' errors ignored. The reason is that a VM in the list of VMs to migrate can have shut down when it comes to migrating it
* Progress is reported on the task by looking at the number of VMs remaining relative to the total number. This number is updated after each batch.
* I've slightly rewritten `migrate` to reduce code nesting and to better use pattern matching.

ocaml/xapi/xapi_host.ml
```diff
@@ -596,48 +610,91 @@ let evacuate ~__context ~host ~network =
...
-        | Api_errors.Server_error (code, _)
-          when code = Api_errors.vm_bad_power_state ->
-            ()
-        | e ->
-            raise e
-        ) ;
-        let progress = Db.Task.get_progress ~__context ~self:task in
-        TaskHelper.set_progress ~__context (progress +. individual_progress)
+        let options = ("network", Ref.string_of network) :: options in
+        Db.Host.get_all ~__context
```
robhoes, line 621: Hmm... since this is checking all hosts and nothing specific to the VM being migrated, it seems unnecessary to do this for each migrate call. We should be able to do it once, in the beginning of the evacuate function?

robhoes: The options are the same for each VM as well.

lindig: We would have to collect all the hosts that we see in the plans and check these once at the beginning.

robhoes: Could to that as well, but the current code actually checks _all_ hosts and doesn't look at the plan.

robhoes (approved):

lindig:
I have squashed the commits - this would be ready to merge. Will keep testing it.

ocaml/xapi/xapi_globs.ml
```diff
@@ -1355,6 +1357,11 @@ let other_options =
     , (fun () -> string_of_bool !migration_compression)
     , "Use compression during VM migration when no API option provided."
     )
+  ; ( "evacuation-batch-size"
+    , Arg.Set_int evacuation_batch_size
+    , (fun () -> string_of_int !evacuation_batch_size)
+    , "The number of VMs evacauted from a host in parallel."
```
edwintorok, line 1363: typo `evacauted`

ocaml/xapi/xapi_host.ml
```diff
@@ -596,48 +611,104 @@ let evacuate ~__context ~host ~network =
...
+        )
+    in
+
+    let assert_success task =
+      match Db.Task.get_status ~__context ~self:task with
+      | `success ->
+          ()
+      | `failure -> (
+        match Db.Task.get_error_info ~__context ~self:task with
+        | [] | [_] ->
```
edwintorok, line 667: Are we sure that `[_]` belongs here? couldn't it be an error message with 0 params (if such errors exist?), i.e. `code :: []`

ocaml/xapi/xapi_host.ml
```diff
@@ -596,48 +611,104 @@ let evacuate ~__context ~host ~network =
...
+        Client.Client.Async.VM.pool_migrate ~rpc ~session_id ~vm ~host ~options
     | Error (code, params) ->
         (* should never happen *)
         raise (Api_errors.Server_error (code, params))
   in
-  let () = Hashtbl.iter migrate_vm plans in
+
+  (* execute [n] asynchronous API calls [api_fn] for [xs] and wait for them to
+     finish before executing the next batch. *)
+  let batch ~__context n api_fn xs =
```
edwintorok, line 650: In the future we can probably extract this function into a module of its own to run tasks in parallel with batching, but that can wait until we're happy with the way it works in this particular case.

scripts/xapi.conf
```diff
@@ -375,6 +375,9 @@ sm-plugins=ext nfs iscsi lvmoiscsi dummy file hba rawhba udev iso lvm lvmohba lv
 # migration-compression = true
 # if not requested otherwise, use stream compression during migration of a VM
 
+# evacuation-batch-size = 10
```
edwintorok, line 378: xenopsd has 16 workers (to match the 16 vCPUs of Dom0), so we could make this default to 16. For now 10 is probably fine though, we should have some tests that measure what the optimal number would be here (a too high number might increase VM downtime if the memory is not compressible well, or if you have vGPUs)

edwintorok (approved):

PR #4756, opened 2022-07-15
Spawn a thread to run xe-toolstack-restart

minglumlu:
It has been observed that a "host.restart_agent" XAPI call reported
failure with error "CANNOT_CONTACT_HOST". This was caused by the race
condition between: "host.restart_agent" call on the member host
terminated the stunnel connection bearing the call.

This commit uses a new thread to run the "xe-toolstack-restart" command,
and meanwhile adds a 1-second delay in the thread to make the XAPI call
complete before the stunnel connection being terminated by
"xe-toolstack-restart" command in more probability.

### DIFF
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -674,14 +674,24 @@ let retrieve_wlb_evacuate_recommendations ~__context ~self =
     plans []
 
 let restart_agent ~__context ~host:_ =
-  let syslog_stdout = Forkhelpers.Syslog_WithKey "Host.restart_agent" in
-  let pid =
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_host.ml
```diff
@@ -683,6 +683,12 @@ let restart_agent ~__context ~host:_ =
   debug "Created process with pid: %d to perform xe-toolstack-restart"
     (Forkhelpers.getpid pid)
 
+let restart_agent ~__context ~host:_ =
```
lindig, line 686: I'm fine with the overall solution but would suggest to no create a new function but to keep it within the existing `restart_agent`

lindig (approved):

PR #4757, opened 2022-07-18
CA-368912 CP-38583 set Host.last_software_update on upgrade

lindig:
Host.last_software_update is a new field in the host record that
reflects the last time a host was updated. This field has a default of
Date.never, which is misleading after an upgarde. Use the DB upgrade
mechanism to set it to the current date.

### DIFF
--- a/ocaml/xapi/xapi_db_upgrade.ml
+++ b/ocaml/xapi/xapi_db_upgrade.ml
@@ -868,6 +868,18 @@ let remove_legacy_ssl_support =
       )
   }
 
+let upgrade_last_updated_field =
+  {
+    description= "update Host.last_software_update field"
+  ; version= (fun _ -> true)
+  ; fn=
+      (fun ~__context ->
+        let host = Helpers.get_localhost ~__context in
+        Db.Host.set_last_software_update ~__context ~self:host
+          ~value:(Xapi_host.get_servertime ~__context ~host)
+      )
+  }
+
 let rules =
   [
     upgrade_domain_type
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_db_upgrade.ml
```diff
@@ -868,6 +868,18 @@ let remove_legacy_ssl_support =
       )
   }
 
+let upgrade_last_updated_field =
+  {
+    description= "update Host.last_software_update field"
+  ; version= (fun _ -> true)
```
robhoes, line 874: Let do
```suggestion
  ; version= (fun x -> x <= yangtze)
```
so that it doesn't kick in on future upgrades, where the field is set in a different place.

ocaml/xapi/xapi_db_upgrade.ml
```diff
@@ -868,6 +868,18 @@ let remove_legacy_ssl_support =
...
+  ; fn=
+      (fun ~__context ->
+        let host = Helpers.get_localhost ~__context in
+        Db.Host.set_last_software_update ~__context ~self:host
+          ~value:(Xapi_host.get_servertime ~__context ~host)
```
edwintorok, line 879: This might be misleading: we claim we've updated the host "now", when in fact we probably just installed an ISO, which could've been months/years out of date.

What is the semantics of this field:
* does it mean that the host has the latest cloud updates installed as of <timestamp>
* the date the last software update got installed was <timestamp>, but we don't know how out-of-date that is with the cloud

It might be useful to have some kind of field that tells you the latest HCP update number that got installed as well to disambiguate.

Either way I'd suggest to try to set this in some way to the "ISO" build date, e.g. can we get that based on the DBV, or some other ISO metadata?

lindig:
After some discussion, let's keep the existing behavior.

PR #4758, opened 2022-07-18
CA-368910: Allow destruction of PVS_cache_storage if SR is already gone

robhoes:
Check if the SR still exists before attempting to scan it in
PVS_cache_storage.destoy. Without that, the call may fail, making it
impossible to clean up the cache storage object.

### DIFF
--- a/ocaml/xapi/pvs_cache_vdi.ml
+++ b/ocaml/xapi/pvs_cache_vdi.ml
@@ -44,18 +44,24 @@ let get_vdi ~__context ~self =
   if vbds <> [] then
     Some vdi
   else
-    (* Scan the SR. This will cause the removal of VDI records that are not backed
-       by an actual volume on the SR. *)
     let sr = Db.PVS_cache_storage.get_SR ~__context ~self in
-    Helpers.call_api_functions ~__context (fun rpc session_id ->
-        Client.Client.SR.scan ~rpc ~session_id ~sr
-    ) ;
-    (* If our VDI reference is still valid, then we're good. *)
-    if Db.is_valid_ref __context vdi then
-      Some vdi
-    else (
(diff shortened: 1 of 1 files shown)


ocaml/xapi/pvs_cache_vdi.ml
```diff
@@ -44,18 +44,24 @@ let get_vdi ~__context ~self =
...
     let sr = Db.PVS_cache_storage.get_SR ~__context ~self in
-    Helpers.call_api_functions ~__context (fun rpc session_id ->
-        Client.Client.SR.scan ~rpc ~session_id ~sr
-    ) ;
-    (* If our VDI reference is still valid, then we're good. *)
-    if Db.is_valid_ref __context vdi then
-      Some vdi
-    else (
-      info "PVS-cache VDI %s is no longer present on the SR" (Ref.string_of vdi) ;
+    if not (Db.is_valid_ref __context sr) then (
```
lindig, line 48: I would try to implement this with less nesting (just a sketch):
```
if Db.is_valid_ref __context sr then
          Client.Client.SR.scan ~rpc ~session_id ~sr
      ) ;
if Db.is_valid_ref __context sr then
  Some vdi
else 
  None
```

lindig (approved):

ocaml/xapi/pvs_cache_vdi.ml
robhoes: I do want to return `None` right away though, without doing more DB calls.

robhoes: Given the restricted order of operations and side-effects involved here, I'm not sure if this is going to get any better...

edwintorok (approved):

PR #4759, opened 2022-07-19
CA-368806: Workaround pbis get wedged

liulinC:
In this ticket, all PBIS lwsmd service threads keep in waiting
status and not able to service new requests. Just restart the
service can resovle the issue and bring the service back.

This symptom looks like a race condition. Given PBIS is going to
be replaced with winbind, this commit workaground the issue
instead of investigate and patch PBIS

* Restart lwsmd on LW_ERROR_ERRNO_ECONNREFUSED
* Restart lwsmd on xapi start/restart

### DIFF
--- a/ocaml/xapi/extauth_plugin_ADpbis.ml
+++ b/ocaml/xapi/extauth_plugin_ADpbis.ml
@@ -27,6 +27,12 @@ let with_lock = Xapi_stdext_threads.Threadext.Mutex.execute
 let lwsmd_service = "lwsmd"
 
 module Lwsmd = struct
+  (* This can be refined by Mtime.Span.hour when mtime is updated to 1.4.0 *)
+  let restart_interval = Int64.mul 3600L 1000000000L |> Mtime.Span.of_uint64_ns
+
+  let next_check_point =
+    Mtime.add_span (Mtime_clock.now ()) restart_interval |> ref
+
   let is_ad_enabled ~__context =
     ( Helpers.get_localhost ~__context |> fun self ->
       Db.Host.get_external_auth_type ~__context ~self
@@ -51,6 +57,22 @@ module Lwsmd = struct
   let start ~timeout ~wait_until_success =
     Xapi_systemctl.start ~timeout ~wait_until_success lwsmd_service
 
+  let restart ~timeout ~wait_until_success =
+    Xapi_systemctl.restart ~timeout ~wait_until_success lwsmd_service
+
+  let restart_on_error () =
+    (* Only restart once within restart_interval *)
+    let now = Mtime_clock.now () in
+    match !next_check_point with
+    | Some check_point ->
+        if Mtime.is_later now ~than:check_point then (
+          debug "Restart %s due to local server error" lwsmd_service ;
+          next_check_point := Mtime.add_span now restart_interval ;
+          restart ~timeout:0. ~wait_until_success:false
+        )
+    | None ->
+        debug "next_check_point overflow"
+
   let init_service ~__context =
     (* This function is called during xapi start *)
     (* it will start lwsmd service if the host is authed with AD *)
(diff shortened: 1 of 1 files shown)


liulinC:
will backport after merge

ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -51,6 +55,18 @@ module Lwsmd = struct
...
+  let restart ~timeout ~wait_until_success =
+    Xapi_systemctl.restart ~timeout ~wait_until_success lwsmd_service
+
+  let restart_on_error () =
+    (* Only restart once within restart_interval *)
+    let now = Unix.gettimeofday () in
+    if now -. !last_restart >= restart_interval then (
+      debug "Restart %s due to local server error" lwsmd_service ;
+      restart ~timeout:0. ~wait_until_success:false ;
+      last_restart := now
```
edwintorok, line 67: I think we should set last_restart before calling `restart`, otherwise if there are N concurrent AD requests and they all fail with econnrefused we'll attempt to restart lwsmd N times, which in turn can cause more failures, and so on.
It would also be better to use monotonic clocks for measuring duration than Unix.gettimeofday, because Unix.gettimeofday can go backwards (e.g. due to NTP) which can completely disable this codepath then (if last_restart becomes a timestamp very far in the future). See usages of `Mtime` elsewhere in XAPI.

ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -27,6 +27,10 @@ let with_lock = Xapi_stdext_threads.Threadext.Mutex.execute
 let lwsmd_service = "lwsmd"
 
 module Lwsmd = struct
+  let last_restart = Mtime_clock.now () |> ref
+
+  let restart_interval = 3600.
```
lindig, line 32: This should be an `Mtime.Span.t` value:

```
let restart_interval = Mtime.(1 * Span.hour)
```

or similar - depending on what span you want.

liulinC: well, we could do this, 
However, I do not find Span.hour in https://v3.ocaml.org/p/mtime/1.2.0/doc/Mtime/Span/index.html, 
maybe `let restart_interval = Mtime.(3600* 1000000000 * Span.one)` ?

liulinC: transformer used `1.2.0`, 
`mtime                   1.2.0       Monotonic wall-clock time for OCaml`, 
so I will not use `Mtime.Span.(1 * hour)` as this requires backport.

liulinC: I upgraded my Mtime and it worked 
```
utop # let restart_internval = Mtime.Span.hour;;
val restart_internval : Mtime.span = <abstr>
```

ocaml/xapi/extauth_plugin_ADpbis.ml
```diff
@@ -27,6 +27,12 @@ let with_lock = Xapi_stdext_threads.Threadext.Mutex.execute
 let lwsmd_service = "lwsmd"
 
 module Lwsmd = struct
+  (* This can be refined by Mtime.Span.hour when mtime is updated to 1.4.0 *)
```
lindig, line 30: Mtime in xs-opam is at 1.4.0, which as a package is at 6.59.0. This package is in rt-next. So I see no reason to not use 1.4.0.

liulinC: we need to back port this commit to yangtze  (as you see this update is for pbis, and pbis is not in rt-next),
I fix this in master and then back port to yangtze to keep consistency. 
I see yangtze still use 1.2.0.

lindig (approved):

edwintorok (approved):

PR #4760, opened 2022-07-19
CA-368437 remove duplicate keys from SM.features

lindig:
The association list SM.features frequently contains duplicate keys.
This is problematic because they could bind to different values (which
are versions in this case).

When parsing the features from a list of strings, eliminate duplicate
keys. In addition, simplify the implementation and avoid duplicate work
by using List.filter_map.

### DIFF
--- a/ocaml/tests/test_sm_features.ml
+++ b/ocaml/tests/test_sm_features.ml
@@ -46,55 +46,53 @@ let test_sequences =
     {
       raw=
         [
-          "SR_PROBE"
-        ; "SR_UPDATE"
+          "ATOMIC_PAUSE"
         ; "SR_CACHING"
-        ; (* xapi ignores this. *)
-          "VDI_CREATE"
-        ; "VDI_DELETE"
+        ; "SR_PROBE"
+        ; "SR_UPDATE"
         ; "VDI_ATTACH"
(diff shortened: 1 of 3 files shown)


edwintorok (commented):
It could be that an SM interface supports both v1 and v2 of something (e.g. during a transition when v2 is introduced and to allow rolling out the `sm` and `xapi` change independently), in which case I think that should be retained: something that only knows how to talk v1 to it might still want to talk v1. So perhaps sort_uniq should be done on both the key and the value? i.e. just avoid duplicate foo/1, foo/1, but keep foo/1, foo/2

lindig:
Good idea. This would remove duplicate identical entries but keep entries of identical keys with different values. This it not tested and I would assume that this is not panned but we can keep it.

robhoes:
Ah, the value of these keys is the version number?

lindig:
I'm gone back to having no key duplicates. Otherwise we would have several versions per feature, which I doubt it the underlying model.

robhoes (approved):

edwintorok:
https://datatracker.ietf.org/doc/html/rfc8259#section-4 json says that keys should be unique for best interop, although the original json.org defines objects as unordered sets of name/value pairs, and not as a map.

PR #4761, opened 2022-07-21
CP-40175: Strip metadata of non-applicable livepatches

minglumlu:
This commit aims to remove metadata of non-applicable livepatches from
the updateinfo list in return of request to HTTP /updates endpoint.

The metadata of livepatches in an updateinfo may contain non-applicable
livepatches to individual pools. Presenting the non-applicable
livepatches causes confusion.


ocaml/xapi/repository_helpers.ml
```diff
@@ -1079,3 +1079,24 @@ let with_access_token ~token ~token_id f =
   | _ ->
       let msg = Printf.sprintf "%s: The token or token_id is empty" __LOC__ in
       raise Api_errors.(Server_error (internal_error, [msg]))
+
+let sort_in_assoc x =
+  Yojson.Basic.Util.to_assoc x
```
lindig, line 1084: What does the sorting accomplish? It looks like the core of the task is to filter out livepatches. Why does this require sorting?

lindig: Hm. This is to make it work with the built-in equality. I think it would be better to have an OCaml record type that defines a live patch and use equality over that to make it more explicit. The whole quality of JSON assocs is a bit too indirect for my taste.

ocaml/xapi/repository_helpers.ml
```diff
@@ -1079,3 +1079,24 @@ let with_access_token ~token ~token_id f =
...
+let prune_updateinfo_for_livepatches livepatches updateinfo =
+  let updateinfo' = Yojson.Basic.Util.to_assoc updateinfo in
+  match List.assoc_opt "livepatches" updateinfo' with
+  | None ->
+      updateinfo
+  | Some l ->
+      let livepatches' = List.map sort_in_assoc livepatches in
+      let l' =
+        Yojson.Basic.Util.to_list l
+        |> List.filter (fun x -> List.mem (sort_in_assoc x) livepatches')
```
edwintorok, line 1097: Could this be done using Set intersection and Set difference operations instead of operations on lists? Might be more natural to express it that way, and Sets would guarantee uniqueness of entries, and a good sorted order when converting back to lists.

lindig (approved):

robhoes (approved):

ocaml/xapi/repository.ml
```diff
@@ -608,27 +608,31 @@ let get_pool_updates_in_json ~__context ~hosts =
...
-              (json_of_host :: acc1, UpdateIdSet.union upd_ids acc2)
+              (updates_of_host :: acc1, UpdateIdSet.union upd_ids acc2)
             else
               (acc1, acc2)
           )
           updates_in_cache ([], UpdateIdSet.empty)
       in
+      let lps =
+        updates_of_hosts
+        |> List.map (fun x -> x.HostUpdates.livepatches)
```
edwintorok, line 623: For the future: this could've been written as:
```
|> List.to_seq
|> Seq.flat_map (fun x -> x.HostUpdates.livepatches)
|> LivePatchSet.of_seq
```

`map + concat = flat_map`, and using a sequence avoids building an intermediate list, though in this case the lists are probably fairly small so doesn't matter.

ocaml/xapi/repository.ml
```diff
@@ -608,27 +608,31 @@ let get_pool_updates_in_json ~__context ~hosts =
...
+        |> List.concat
+        |> LivePatchSet.of_list
+      in
+      let updateinfo_list =
+        UpdateIdSet.elements ids_of_updates
+        |> List.map (fun upd_id -> List.assoc upd_id updates_info)
```
edwintorok, line 629: Perhaps we can change updates_info to be a map in the future? (e.g. separate PR)

ocaml/xapi/updateinfo.ml
```diff
@@ -573,3 +575,33 @@ module UpdateInfo = struct
         error "Failed to parse updateinfo.xml: %s" (ExnHelper.string_of_exn e) ;
         raise Api_errors.(Server_error (invalid_updateinfo_xml, []))
 end
+
+module HostUpdates = struct
+  type t = {
```
edwintorok, line 580: I wonder why aren't we using xapi-idl here and doing it by hand. Is it because it doesn't support nested types well?

edwintorok (approved):

ocaml/xapi/updateinfo.ml
minglumlu: This is for an external interface which defines the JSON content in the response of the HTTP endpoint "/updates". I don't think it is suitable to put it in xapi-idl which I understand is holding the definitions of Toolstack's internal interfaces.

ocaml/xapi/repository.ml
minglumlu: Yeah. This makes sense. I created an improvement for this. Thanks.

PR #4762, opened 2022-07-22
CA-347473: Minor memory leak from unloaded Xen livepatches

minglumlu:
The latest livepatch will always roll up the previous ones. Once the
latest one has been applied, the previous ones should be un-loaded to
avoid minor memory leak.

### DIFF
--- a/ocaml/xapi/livepatch.ml
+++ b/ocaml/xapi/livepatch.ml
@@ -187,14 +187,16 @@ module KernelLivePatch = struct
 end
 
 module XenLivePatch = struct
-  let get_running_livepatch' s =
-    let r =
-      Re.Posix.compile_pat
-        {|^[ ]*lp_([^- ]+)-([^- ]+)-([^- ]+)-([^- ]+).+APPLIED.*$|}
-    in
+  let get_pattern status =
+    Re.Posix.compile_pat
+      (Printf.sprintf {|^[ ]*lp_([^- ]+)-([^- ]+)-([^- ]+)-([^- ]+).+%s.*$|}
+         status
+      )
+
+  let get_livepatches pattern s =
     Astring.String.cuts ~sep:"\n" s
     |> List.filter_map (fun line ->
-           match Re.exec_opt r line with
+           match Re.exec_opt pattern line with
            | Some groups ->
                let base_version = Re.Group.get groups 1 in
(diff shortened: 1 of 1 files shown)


ocaml/xapi/livepatch.ml
```diff
@@ -243,7 +252,20 @@ module XenLivePatch = struct
...
+     * The status of these livepatches is changed to 'CHECKED' from 'APPLIED' when
+     * the latest livepatch is being applied.
+     *)
+    get_checked_livepatches ()
+    |> List.iter (fun (base_version, base_release, to_version, to_release) ->
+           let name =
+             Printf.sprintf "lp_%s-%s-%s-%s" base_version base_release
+               to_version to_release
+           in
+           Helpers.call_script !Xapi_globs.xen_livepatch_cmd ["unload"; name]
```
edwintorok, line 266: Does the order in which these patches are unloaded matter, i.e. do you need to unload them in the reverse order in which they got loaded, or it doesn't matter.
@andyhhp @rosslagerwall ?

ocaml/xapi/livepatch.ml
```diff
@@ -187,14 +187,16 @@ module KernelLivePatch = struct
 end
 
 module XenLivePatch = struct
-  let get_running_livepatch' s =
-    let r =
-      Re.Posix.compile_pat
-        {|^[ ]*lp_([^- ]+)-([^- ]+)-([^- ]+)-([^- ]+).+APPLIED.*$|}
-    in
+  let get_pattern status =
```
lindig, line 190: A better name would be `regexp`, `get_regexp`, `regexp_for`.

ocaml/xapi/livepatch.ml
minglumlu: It worked well from my test in which 3 simple livepatches were used. At least no errors were reported by the `xen-livepatch` tool.
But, yeah, it would be much better to confirm this from rationale.

robhoes (approved):

robhoes:
Looks good to me, but let's confirm the ordering question before merging (and squash the fixup).

edwintorok:
Looks like we have confirmation about the lack of an ordering requirement (see the conversation above that I've marked as solved)

edwintorok:
merged with 'squash&merge'

PR #4765, opened 2022-07-27
CP-37225: Added unmarshalling code for Ocaml's Set(Set string) for C

kc284:
This is the last installment for this ticket.

### DIFF
--- a/ocaml/sdk-gen/c/autogen/include/xen/api/xen_string_set_set.h
+++ b/ocaml/sdk-gen/c/autogen/include/xen/api/xen_string_set_set.h
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) Citrix Systems, Inc.
+ * All rights reserved.
+ * 
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 
+ *   1) Redistributions of source code must retain the above copyright
+ *      notice, this list of conditions and the following disclaimer.
+ * 
+ *   2) Redistributions in binary form must reproduce the above
+ *      copyright notice, this list of conditions and the following
+ *      disclaimer in the documentation and/or other materials
+ *      provided with the distribution.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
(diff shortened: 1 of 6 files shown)


ocaml/sdk-gen/c/autogen/src/xen_string_set_set.c
```diff
@@ -0,0 +1,56 @@
...
+#include "xen_internal.h"
+#include <xen/api/xen_string_set.h>
+#include <xen/api/xen_string_set_set.h>
+
+
+xen_string_set_set *
+xen_string_set_set_alloc(size_t size)
+{
+    xen_string_set_set *result = calloc(1, sizeof(xen_string_set_set) +
+                                    size * sizeof(char *));
```
robhoes, line 40: Not `sizeof(xen_string_set *)` (it may work out the same size)?

ocaml/sdk-gen/c/autogen/include/xen/api/xen_string_set_set.h
```diff
@@ -0,0 +1,60 @@
...
+
+
+#include "xen_common.h"
+#include <xen/api/xen_string_set.h>
+
+
+typedef struct xen_string_set_set
+{
+    size_t size;
+    xen_string_set *contents[];
```
edwintorok, line 42: Flexible array members without a size need at least C99.
I remember that Visual Studio didn't support C99 (sometimes one had to change the language from C to C++ to get things to compile, but that is not entirely correct either since C++ isn't necessarily a strict superset of C),
other compilers on Windows (and Linux) shouldn't have a problem though.

If you have Visual Studio can you check whether it can compile this as C code? If not a workaround would be to declare the array of having at least 1 member and adjusting the calloc below to allocate n-1 elements (but that is a hack and then I'm not sure it'd be entirely C99 compliant...)

edwintorok: It looks like there are now /std:c11 flags since VIsual Studio 2019 version 16.8, so you can try using that (I don't know which version the default implements, their documentation is unclear about that).

kc284: We always compile it with `-std=c99` (see ocaml/sdk-gen/c/templates/Makefile.mustache).

lindig:
This PR deals in part with formatting of C code. You could consider using a Makefile rule to autoformat it using `indent`. I am using `indent -linux foo.c` but GNU indent supports many other styles.

kc284:
I only changed the indentation in a few lines that were sticking out, I wasn't really planning to format the whole code. It's a good suggestion though and I can implement it in a different PR. From what I see it will need some tweaking because the `-linux` default makes the file less readable (to my eyes at least) - and I don't see an option to disable the generation of back-up files.

lindig:
Since the backup files are not under Git's control they can ignored in `.gitignore` and could be removed with a find(1) command. indent has many options to control the formatting, including some shortcuts to specify common styles like GNU, K&R, or Linux.

edwintorok (approved):

lindig (approved):

kc284:
Are we happy to merge this PR now?

PR #4766, opened 2022-07-27
CA-367236 replace Ezjsonm with Yojson

lindig:
We would like to use just one library for creating and consuming JSON;
so replace Ezjsonm with Yojson.

The serialisation of timestamps in the V1 transport appears to change:
they are now serialised as floats; to avoid errors in the parse, we need
to change the reader for the V1 transport.

### DIFF
--- a/ocaml/xcp-rrdd/lib/transport/base/rrd_protocol_v1.ml
+++ b/ocaml/xcp-rrdd/lib/transport/base/rrd_protocol_v1.ml
@@ -94,7 +94,9 @@ let parse_payload ~(json : string) : payload =
   try
     let rpc = Jsonrpc.of_string json in
     let kvs = Rrd_rpc.dict_of_rpc ~rpc in
-    let timestamp = Rpc.int64_of_rpc (List.assoc "timestamp" kvs) in
+    let timestamp =
+      Rpc.float_of_rpc (List.assoc "timestamp" kvs) |> Int64.of_float
+    in
     let datasource_rpcs =
       Rrd_rpc.dict_of_rpc ~rpc:(List.assoc "datasources" kvs)
     in
--- a/ocaml/xcp-rrdd/lib/transport/base/dune
+++ b/ocaml/xcp-rrdd/lib/transport/base/dune
@@ -10,7 +10,7 @@
     xapi-rrd
     threads
     xapi-idl.rrd
-    ezjsonm
+    yojson
   )
 )
 
--- a/ocaml/xcp-rrdd/lib/transport/base/rrd_json.ml
+++ b/ocaml/xcp-rrdd/lib/transport/base/rrd_json.ml
@@ -48,7 +48,7 @@ let bool b = string "%b" b (* Should use `Bool b *)
 
 let float x = string "%.2f" x
 
-let record xs = `O xs
+let record xs = `Assoc xs
 
 let description = function
   | "" ->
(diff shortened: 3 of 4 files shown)


ocaml/xcp-rrdd/lib/transport/base/rrd_protocol_v1.ml
```diff
@@ -94,7 +94,9 @@ let parse_payload ~(json : string) : payload =
   try
     let rpc = Jsonrpc.of_string json in
     let kvs = Rrd_rpc.dict_of_rpc ~rpc in
-    let timestamp = Rpc.int64_of_rpc (List.assoc "timestamp" kvs) in
+    let timestamp =
+      Rpc.float_of_rpc (List.assoc "timestamp" kvs) |> Int64.of_float
```
robhoes, line 98: Is `` `Float`` encoded differently by yojson and jsonm?

lindig: In the file transport (a 4096-byte file) where the JSON is embedded I noticed that the timestamp was encoded like an integer so far, but with a `.0` suffix with this patch. In pure JSON terms, this is fine: both are numbers but the reader failed because it was expecting an OCaml integer but got an OCaml float.

robhoes: Ah yes, JSON does not distinguish between floats and ints, and just has "numbers" (which may or may not include a decimal point).

lindig:
This passed unit test, BVT, BST and I see no problems in XenCenter when connecting to a machine with this build.

ocaml/xcp-rrdd/lib/transport/base/rrd_protocol_v1.ml
lindig:
```
test-new.v1
00000000  44 41 54 41 53 4f 55 52  43 45 53 0a 30 30 30 30  |DATASOURCES.0000|
00000010  30 32 33 38 0a 33 32 61  62 32 39 34 65 34 65 34  |0238.32ab294e4e4|
00000020  33 33 61 61 38 65 36 39  32 32 36 38 39 33 37 61  |33aa8e692268937a|
00000030  38 30 63 38 62 0a 7b 22  74 69 6d 65 73 74 61 6d  |80c8b.{"timestam|
00000040  70 22 3a 31 33 38 37 38  36 37 32 32 33 2e 30 2c  |p":1387867223.0,|
00000050  22 64 61 74 61 73 6f 75  72 63 65 73 22 3a 7b 22  |"datasources":{"|
test-old.v1
00000000  44 41 54 41 53 4f 55 52  43 45 53 0a 30 30 30 30  |DATASOURCES.0000|
00000010  30 32 33 36 0a 31 33 65  65 33 64 39 32 33 32 62  |0236.13ee3d9232b|
00000020  30 37 62 62 66 64 62 32  62 31 33 38 65 64 32 33  |07bbfdb2b138ed23|
00000030  30 38 38 37 65 0a 7b 22  74 69 6d 65 73 74 61 6d  |0887e.{"timestam|
00000040  70 22 3a 31 33 38 37 38  36 37 32 32 33 2c 22 64  |p":1387867223,"d|
00000050  61 74 61 73 6f 75 72 63  65 73 22 3a 7b 22 74 65  |atasources":{"te|
```
Look at the encoding of the timestamp. It is a number but but could be parsed as int in once case but not the other.

robhoes: Good, I suppose that `Rpc.float_of_rpc` may actually accept both (with and without `.0`). At least `float_of_string` does.

robhoes (approved):

psafont (approved):

ocaml/xcp-rrdd/lib/transport/base/rrd_protocol_v1.ml
edwintorok: https://ocaml.org/p/yojson/2.0.1/doc/Yojson/index.html#type-of-the-json-tree you can use `Int instead of Float to encode integers without a .0 I think.
Jsonm just had a single float type (and tried to avoid the .0 if the number was exactly convertible to int64)

PR #4768, opened 2022-07-29
Simplify and fix UDSTransport implementation

### LINKED ISSUE #4767: XenAPI.py + python 3: TypeError: send_request() takes 4 positional arguments but 5 were given
Opening this issue to discuss this bug in XenAPI.py when executed with python 3. Initially discussed at https://github.com/xenserver/guest-templates-json/pull/1#issuecomment-1179132574

Backtrace:
```
# python3 create-guest-templates 
  File "create-guest-templates", line 10, in <module>
    session.xenapi.login_with_password('', '', '', 'create-template')
  File "/usr/local/lib/python3.6/site-packages/XenAPI/XenAPI.py", line 269, in __call__
    return self.__send(self.__name, args)
  File "/usr/local/lib/python3.6/site-packages/XenAPI/XenAPI.py", line 154, in xenapi_request
    self._login(methodname, params)
  File "/usr/local/lib/python3.6/site-packages/XenAPI/XenAPI.py", line 179, in _login
    getattr(self, 'session.%s' % method)(*params))
  File "/usr/lib64/python3.6/xmlrpc/client.py", line 1112, in __call__
    return self.__send(self.__name, args)
  File "/usr/lib64/python3.6/xmlrpc/client.py", line 1452, in __request
    verbose=self.__verbose
  File "/usr/lib64/python3.6/xmlrpc/client.py", line 1154, in request
    return self.single_request(host, handler, request_body, verbose)
  File "/usr/lib64/python3.6/xmlrpc/client.py", line 1166, in single_request
    http_conn = self.send_request(host, handler, request_body, verbose)
TypeError: send_request() takes 4 positional arguments but 5 were given
```

The bug is likely located at https://github.com/xapi-project/xen-api/blob/master/scripts/examples/python/XenAPI/XenAPI.py#L111 where a class that inherits `xmlrpc.client.Transport` redefines `send_request`.

`xmlrpc.client.Transport` doesn't have the same signature for `send_request` in python2 and python3, so a parameter is missing when called in python3. The backtrace is somewhat misleading as no mention of this function is made at all.

A similar issue is discussed at https://www.reddit.com/r/learnpython/comments/1l38mf/i_cant_get_xmlrpc_on_python_3_to_use_an_http_proxy/

@edwintorok wrote:
> XenAPI.py shouldn't override send_request, it should override make_connection instead, see https://stackoverflow.com/questions/11729159/use-python-xmlrpclib-with-unix-domain-sockets for some ideas.

stormi:
The current implementation of UDSTransport in XenAPI.py and rrdd.py has
several issues:
- It redefines xmlrpclib.Transport.send_request, whose signature has changed
  in python 3, so this breaks XenAPI.py's python 3 port.
- The overridden implementation of send_request adds extra headers that are
  already handled by the parent class. As a result, extra headers are sent
  twice.
- Method __init__ is overridden needlessly, which could cause issues in
  the future if the parent class evolves.

This commit fixes the issues by simply deleting the send_request and
__init__ methods from the implementation of UDSTransport.

At the same time, code that was written to retain compatibility with python
< 2.7 is removed.

The add_extra_header method is kept in XenAPI.py, as `sm` apparently uses it
to add a `Subtask-of:` header. Symmetrically, I also kept it in rrdd.py,
but I'm not sure it's used anywhere.

Fixes #4767

### DIFF
--- a/ocaml/xcp-rrdd/scripts/rrdd/rrdd.py
+++ b/ocaml/xcp-rrdd/scripts/rrdd/rrdd.py
@@ -75,28 +75,13 @@ class UDSHTTPConnection(httplib.HTTPConnection):
     self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
     self.sock.connect(path)
 
-class UDSHTTP(httplib.HTTP):
-  _connection_class = UDSHTTPConnection
-
 class UDSTransport(xmlrpclib.Transport):
-  def __init__(self, use_datetime = 0):
-    self._use_datetime = use_datetime
-    self._extra_headers = []
-
+  # FIXME: is this function actually used anywhere?
   def add_extra_header(self, key, value):
     self._extra_headers += [(key, value)]
 
   def make_connection(self, host):
-    # Python 2.4 compatibility
-    if sys.version_info[0] <= 2 and sys.version_info[1] < 6:
-      return UDSHTTP(host)
-    else:
-      return UDSHTTPConnection(host)
-
-  def send_request(self, connection, handler, request_body):
-    connection.putrequest("POST", handler)
-    for key, value in self._extra_headers:
-      connection.putheader(key, value)
+    return UDSHTTPConnection(host)
 
 class Proxy(xmlrpclib.ServerProxy):
   def __init__(self, uri, transport = None, encoding = None, verbose = 0,
--- a/scripts/examples/python/XenAPI/XenAPI.py
+++ b/scripts/examples/python/XenAPI/XenAPI.py
@@ -92,26 +92,11 @@ class UDSHTTPConnection(httplib.HTTPConnection):
         self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
         self.sock.connect(path)
 
-class UDSHTTP(httplib.HTTPConnection):
-    _connection_class = UDSHTTPConnection
-
 class UDSTransport(xmlrpclib.Transport):
-    def __init__(self, use_datetime=0):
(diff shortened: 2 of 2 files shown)


stormi:
I have tested XenAPI.py with this script and checked the contents of requests with strace:
```
from __future__ import print_function
from guesttemplates import loader
import XenAPI
import errno

if __name__ == '__main__':
    print(XenAPI.__file__)
    session = XenAPI.xapi_local()
    session.transport.add_extra_header('Subtask-of', "POUET")
    session.xenapi.login_with_password('', '', '', 'create-template')
    session.xenapi.session.logout()
```

I also tested the change with `create-guest-templates` that is being ported to python 3, both with python 2 and python 3.

ocaml/xcp-rrdd/scripts/rrdd/rrdd.py
```diff
@@ -75,28 +75,13 @@ def connect(self):
...
 
-class UDSHTTP(httplib.HTTP):
-  _connection_class = UDSHTTPConnection
-
 class UDSTransport(xmlrpclib.Transport):
-  def __init__(self, use_datetime = 0):
-    self._use_datetime = use_datetime
-    self._extra_headers = []
-
+  # FIXME: is this function actually used anywhere?
```
edwintorok, line 79: I think `rrdd.py` is just an example library that one can use to query RRDs from XAPI., I don't think it is actually used by anything in the product, but I could be wrong, so lets keep the method here for now, we can always remove it later.

edwintorok (approved):

psafont (approved):

PR #4770, opened 2022-08-02
Fix quicktest's -default-sr parameter

stormi:
The value of `use_default_sr` is evaluated before it can be set to true
if the `-default-sr` parameter is provided to the command line.

Move the if test inside a function so that `use_default_sr` is only
evaluated when appropriate.

### DIFF
--- a/ocaml/quicktest/qt_filter.ml
+++ b/ocaml/quicktest/qt_filter.ml
@@ -172,15 +172,16 @@ module SR = struct
 
   let all_srs = lazy (list_srs_connected_to_localhost !A.rpc !session_id)
 
-  let all =
+  let all () =
     if !A.use_default_sr then
       let pool = Qt.get_pool !A.rpc !session_id in
       only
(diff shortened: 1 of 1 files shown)


edwintorok (approved):
I don't particularly like that this code uses mutable globals all over the place, that makes it hard to reason about the exact value of various values (not the fault of this PR though).
In this case `default_sr` was wrong because it probbaly got read at module initialization time *before* the cmdline got parsed, and similarly the `rpc` would've had the wrong value if this code ran before we acquired an `rpc`.

Delaying the initialization here sounds good, although this might cause the value to be recomputed multiple times, although the lazy value for `all_srs` should avoid that, and the recomputation would only happen when -default-sr is specified.
This is just test code, not performance critical so probably fine as is for now.

psafont (approved):

stormi:
I'm waiting for the build to complete then I'll push with the formatting fix.

PR #4771, opened 2022-08-10
CP-39850: Copy VTPM on clone

psafont:
On the non-unique mode (default) it copies the VTPM, this allows bitlocker to continue functioning in the clone without re-keying.  On unique mode it creates a new VTPM with the same profile.

This PR is not enough to make snapshots work, this is because it's not easy to notify xapi about the new vtpm contents while the VM is running because xenopsd is not involved. Instead it blocks this operation showing that the functionality is not implemented to avoid loss of data.

I've manually tested the clone with both values of unique and the live snapshot with vtpms associated

### DIFF
--- a/ocaml/xapi/xapi_vm_clone.ml
+++ b/ocaml/xapi/xapi_vm_clone.ml
@@ -421,6 +421,18 @@ let clone ?snapshot_info_record ?(ignore_vdis = []) disk_op ~__context ~vm
       let vgpus = Db.VM.get_VGPUs ~__context ~self:vm in
       let vtpms = Db.VM.get_VTPMs ~__context ~self:vm in
       let power_state = Db.VM.get_power_state ~__context ~self:vm in
+      ( match (power_state, vtpms) with
+      | `Running, _ :: _ ->
+          error "%s: Running VMs with VTPMs cannot be snapshotted yet"
+            __FUNCTION__ ;
+          raise
+            Api_errors.(
+              Server_error
+                (not_implemented, ["VM.clone of running VM with VTPM"])
+            )
+      | _ ->
+          ()
+      ) ;
       (* If a VM is snapshotted, then the new VM must remain halted.
          Otherwise, we keep the same power-state as the initial VM *)
       let new_power_state =
--- a/ocaml/xapi/xapi_vm_clone.ml
+++ b/ocaml/xapi/xapi_vm_clone.ml
@@ -419,6 +419,7 @@ let clone ?snapshot_info_record ?(ignore_vdis = []) disk_op ~__context ~vm
       let vbds = Db.VM.get_VBDs ~__context ~self:vm in
       let vifs = Db.VM.get_VIFs ~__context ~self:vm in
       let vgpus = Db.VM.get_VGPUs ~__context ~self:vm in
+      let vtpms = Db.VM.get_VTPMs ~__context ~self:vm in
       let power_state = Db.VM.get_power_state ~__context ~self:vm in
       (* If a VM is snapshotted, then the new VM must remain halted.
          Otherwise, we keep the same power-state as the initial VM *)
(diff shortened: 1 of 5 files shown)


ocaml/xapi/xapi_vm_clone.ml
```diff
@@ -498,6 +503,10 @@ let clone ?snapshot_info_record ?(ignore_vdis = []) disk_op ~__context ~vm
           let (_ : [`VGPU] Ref.t list) =
             List.map (fun vgpu -> Xapi_vgpu.copy ~__context ~vm:ref vgpu) vgpus
           in
+          (* copy vTPMs *)
+          let (_ : [`VTPM] Ref.t list) =
+            List.map (fun vtpm -> Xapi_vtpm.copy ~__context ~vM:ref vtpm) vtpms
```
lindig, line 508: This looks strange. We are using `List.map` but are not using the result? Why not use `List.iter`? Or maybe use `ignore` and be more explicit?

ocaml/xapi/xapi_vtpm.ml
```diff
@@ -20,19 +20,41 @@ let assert_no_vtpm_associated ~__context vm =
...
-  let persistence_backend = `xapi in
-  let contents = Xapi_secret.create ~__context ~value:"" ~other_config:[] in
   Db.VTPM.create ~__context ~ref ~uuid ~vM ~backend ~persistence_backend
     ~is_unique ~is_protected:false ~contents ;
   ref
 
+(** Contents from unique vtpms cannot be copied! *)
+let get_contents ~__context ?from () =
+  let block_copy ref =
+    if Db.VTPM.get_is_unique ~__context ~self:ref then None else Some ref
```
lindig, line 34: I am confused. We are either returning `None` or the argument as `Some`. This suggests the function is a predicate on `ref` but this not reflected in the name.

ocaml/xapi/xapi_vm_clone.ml
psafont: I also thought of using List.iter, but too keep it in 3 lines it needs the use of ignore, I'd rather use map with the explicit types.

Using iter and ignore would look like this, which is not too far off, but if copy changes we might not detect that change because we don't enforce types:
```
          List.iter
            (fun vtpm -> ignore (Xapi_vtpm.copy ~__context ~vM:ref vtpm))
            vtpms ;
```

lindig: I believe `ignore` would detect the case where the function application becomes partial but I agree that types are not as explicit.

ocaml/xapi/xapi_vtpm.ml
psafont: I was afraid it might cause confusion.
The purpose of this function is to disallowing VTPMs with the unique profile from ever being copied while being able to compose with the optional parameter `?frpm` which contains a `ref option`.

I'll try thinking of a less confusing way to express this

psafont: Is this version easier to follow?

ocaml/xapi/xapi_vm_clone.ml
edwintorok: Seems like all other object copies are implemented weirdly like this, I would've expected that you receive the original vtpm object, you make a copy and you put the ref that got returned into the new object.
Otherwise where does it go?

edwintorok: Can we change the type of `copy` to return `unit`, and in the mli mention that it performs the copy and puts the new object into the specified target VM?
I think it is weird like this because it needs to maintain the relations in the DB: when you copy the VTPM object you also need to update its VM field which is a backreference to its owning VM (which you also changed when copying).

psafont: It is useful to have all the `copy` methods return the reference of the newly created method in case the caller needs to do some operation with them (like the vtpm code does when it copies secrets). In this particular case that is not needed and the new references are simply ignored. This patterns is also followed across xapi for `create`.

> I think it is weird like this because it needs to maintain the relations in the DB: when you copy the VTPM object you also need to update its VM field which is a backreference to its owning VM (which you also changed when copying)
The VM is created by `copy_vm_record` which avoids giving any parameters of relationships to the actual `VM.create`.

I'd rather follow the current pattern rather than make the copy function return `unit` to make this callsite more pleasant to interpret.

lindig (approved):

edwintorok (approved):

PR #4772, opened 2022-08-12
CA-369599: ignore invalid references on eject

psafont:
While ejecting multiple hosts in parallel some references may become
invalid, ignore them as they are from objects pertaining to ejected
hosts.

### DIFF
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -772,7 +772,7 @@ let get_my_pbds __context =
 (* Return the PBD for specified SR on a specific host *)
 (* Just say an SR is shared if it has more than one PBD *)
 let is_sr_shared ~__context ~self =
-  List.length (Db.SR.get_PBDs ~__context ~self) > 1
+  match Db.SR.get_PBDs ~__context ~self with [] | [_] -> false | _ -> true
 
 let get_main_ip_address ~__context =
   try Pool_role.get_master_address () with _ -> "127.0.0.1"
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -122,6 +122,9 @@ let choose_network_name_for_pif device =
 (* !! FIXME(2) - this code could be shared with the CLI? *)
 let checknull f = try f () with _ -> "<not in database>"
 
+let ignore_invalid_ref f (x : 'a Ref.t) =
+  try Ref.to_option (f x) with Db_exn.DBCache_NotFound _ -> None
+
 let get_pool ~__context = List.hd (Db.Pool.get_all ~__context)
 
 let get_master ~__context =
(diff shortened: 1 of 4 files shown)


ocaml/xapi/xapi_pool.ml
```diff
@@ -1745,29 +1745,40 @@ let eject_self ~__context ~host =
...
+       halted), they should have disks on my local storage *)
     let vm_is_resident_on_host vm_rec host =
       Db.is_valid_ref __context vm_rec.API.vM_resident_on
       && vm_rec.API.vM_resident_on = host
     in
-    let vm_has_disks_on_local_sr_of_host vm_ref host =
-      let is_sr_local x = not (Helpers.is_sr_shared ~__context ~self:x) in
+    let vm_has_disks_on_local_sr_of_host vm_rec host =
+      let is_sr_local x =
+        try not (Helpers.is_sr_shared ~__context ~self:x) with _ -> false
```
edwintorok, line 1757: Should we catch just the specific error for invalid reference here instead of a catch all?
E.g. if we have some other genuine error we should probably surface that.

ocaml/xapi/xapi_pool.ml
```diff
@@ -1745,29 +1745,40 @@ let eject_self ~__context ~host =
...
-      (* filter out null ref VDIs (can happen e.g. for CDs) *)
-      |> List.map (fun x -> Db.VDI.get_SR ~__context ~self:x)
+      let ignore_invalid_ref f x =
+        try
+          let ref = f x in
+          if ref <> Ref.null then
+            Some ref
+          else
+            None
+        with _ -> None
```
edwintorok, line 1769: Same here, perhaps we should have a Helpers.ignore_invalid_ref_exn and a Ref.to_option, the former would ignore exceptions (see above, just the specific one we want to ignore), and the other one would just do the Ref.null -> None conversion.
Then we could perhaps reuse this in other places in the code.

ocaml/xapi/helpers.ml
```diff
@@ -772,7 +772,7 @@ let get_my_pbds __context =
 (* Return the PBD for specified SR on a specific host *)
 (* Just say an SR is shared if it has more than one PBD *)
 let is_sr_shared ~__context ~self =
-  List.length (Db.SR.get_PBDs ~__context ~self) > 1
+  match Db.SR.get_PBDs ~__context ~self with [] | [_] -> false | _ -> true
```
robhoes, line 778: Isn't this functionally the same as before?

psafont: indeed, this removes the usage of  list.length

lindig:
Is this in essence working around a race condition? Hosts get ejected but references obtained before they were objected now have become invalid. I struggle to find a clean solution that would avoid this and so detecting the invalid references like this code does is probably the pragmatic solution.

lindig (approved):

psafont:
> Is this in essence working around a race condition?

Yup, we can't fetch all the information in the database atomically, and I suspect the effects of the eject might not atomically alter the database either.

> I struggle to find a clean solution that would avoid this

One solution would be to carefully serialize the eject operations within a pool, this workaround seems easier to do as we don't really care about control domains of ejected hosts which are backed up by shared storage in the current host.

robhoes (approved):

PR #4773, opened 2022-08-15
CP-40392 compress vGPU migration stream

lindig:
During VM migration memory and vGPU memory are streamed from source
to destination host. Currently, the memory stream can be compressed
optionally while the vGPU stream remains uncompressed. This patch adds
compression for the vGPU stream. It is controlled by the same mechanism
and as such is optional, too, and the default it controlled by a flag in
/etc/xapi.conf. Likewise, it us using the same compression framework
(but it would be easy to add one using different paramters).

### DIFF
--- a/ocaml/xenopsd/lib/xenops_server.ml
+++ b/ocaml/xenopsd/lib/xenops_server.ml
@@ -2398,6 +2398,13 @@ and perform_exn ?subtask ?result (op : operation) (t : Xenops_task.task_handle)
               fun fd fn -> fn fd
           (* do nothing *)
         in
+        let compress_vgpu vgpu_fd f =
+          match vgpu_fd with
+          | Some (FD fd) when compress_memory ->
+              compress fd (fun fd -> f (Some (FD fd)))
+          | vgpu_fd ->
+              f vgpu_fd
+        in
         debug "%s compress memory: %b" __FUNCTION__ compress_memory ;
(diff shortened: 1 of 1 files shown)


lindig:
This patch depends on a change in the `vgpu` daemon such that streaming data over a pipe is supported.

ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -2398,6 +2398,13 @@ and perform_exn ?subtask ?result (op : operation) (t : Xenops_task.task_handle)
               fun fd fn -> fn fd
           (* do nothing *)
         in
+        let compress_vgpu vgpu_fd f =
+          match vgpu_fd with
+          | Some (FD fd) when compress_memory ->
+              compress fd (fun fd -> f (Some (FD fd)))
```
robhoes, line 2404: `compress` already becomes the identity function if `compress_memory = false`, so can we just use `compress` directly?

lindig: We can't use `compress` here because the argument type is not a `fd` but more complicated and needs to be handled.

lindig:
We now have the changes in `vgpu` 7.4.9 in place to make this merge safe.

psafont (approved):

robhoes (approved):

PR #4774, opened 2022-08-16
CP-40284, CP-39874: Gate VTPM creation behind an experimental feature and Prevent VTPMs and HA from coexisting

psafont:
This also blocks creation of VMs from templates with vtpm in the platform data.

Tested manually both cases for feature flag, and 4 cases for fencing + vtpm (try to create vtpm with ha, clustering on + try to enable ha, cluster when a vtpm is present)

### DIFF
--- a/ocaml/xapi/xapi_vtpm.ml
+++ b/ocaml/xapi/xapi_vtpm.ml
@@ -23,6 +23,19 @@ let assert_not_restricted ~__context =
   | _ ->
       raise Api_errors.(Server_error (feature_restricted, [feature]))
 
+(** The state in the xapi backend is only up-to-date when the VMs are halted *)
+let assert_no_fencing ~__context ~persistence_backend =
+  let pool = Helpers.get_pool ~__context in
+  let ha_enabled = Db.Pool.get_ha_enabled ~__context ~self:pool in
+  let clustering_enabled = Db.Cluster.get_all ~__context <> [] in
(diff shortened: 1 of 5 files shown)


lindig (approved):

ocaml/xapi/xapi_vtpm.ml
```diff
@@ -12,6 +12,30 @@
    GNU Lesser General Public License for more details.
  *)
 
+(** Don't allow unless VTPM is enabled as an experimental feature *)
+let assert_not_restricted ~__context =
+  let pool = Helpers.get_pool ~__context in
+  let restrictions = Db.Pool.get_restrictions ~__context ~self:pool in
+  let feature = "restrict_vtpm" in
```
edwintorok, line 19: This is probably fine while it is temporary, but once it is not experimental anymore we should remove the restriction from here so this also works on non-Citrix builds (or add `vtpm` to the list of default features in Features.mli.
vtpm is expected to be a feature that is always available (once we finish implementing it...)

So perhaps add a comment here to remind ourselves to remove this.

edwintorok (approved):
Looks good, but we should remember to remove the restriction code once the feature is no longer experimental (or add it to Features. default set at that time).

ocaml/xapi/xapi_vtpm.ml
psafont: Creating an issue so we don't forget about this

PR #4776, opened 2022-08-23
CA-370084: Add tests for accepting PEMs with DOS endlines, make parser stricter

psafont:
Yangtze failed to parse PEMs with these line ending, as they are valid PEMs according to RFC 1421 (and users have tried to install these).

Current parser accepts this, but it's quite lax when parsing them, allowing other whitespace characters.
This adds an explicit test for this case, makes the parser stricter and changes the tests to surface more information.


lindig:
I think it looks fine and has been tested. The big change is around whitespace handling. Could you summarise the change here? The change is not just what is considered white space but where is is expected.

lindig (approved):

psafont:
> I think it looks fine and has been tested. The big change is around whitespace handling. Could you summarise the change here? The change is not just what is considered white space but where is is expected.

The whitespace handling was done all in a single function when it was needed for two different purposes:
1. Handling endlines around keys and certificates, now treated at the high level with `end_of_line`
2. Allowing single spaces in key headers: this was unnecessary and has been hardcoded in the inner case when recognising the kind of header.

I need to change the pem file that was introduced, it was checked out with UNIX endlines, git really doesn't like to use the DOS ones.

edwintorok (approved):

psafont:
While this makes the code to accept DOS endline, these need to be treated so the file is human-readable on the CLI

PR #4777, opened 2022-08-23
CA-370084: Print PEMs containing DOS endlines in yangtze

psafont:
The check to find the header and the footer were too strict, trim the lines before comparing.


lindig (approved):

lindig:
This backport could collide with the Certificate Checking backport because this module has changed and we might have to take the Trunk version when backporting Certificat Checking.

PR #4779, opened 2022-08-24
Speed up builds (v2): avoid rebuilding on git commit hash/date changes

psafont:
Supersedes https://github.com/xapi-project/xen-api/pull/4742

This PR changes the steps where the version is defined by two variables:
- ./configure decides the version depending on the --xapi_version parameter. If the parameter is not defined then dune uses git describe to create the version. On shallow clones this version fails to conform to the required version schema, this is why a workaround is needed on github CI (by commiting a version)

- If it's not defined the version is "0.0.dev". This happens on dev builds and and on non-installed binaries. This is done to avoid changing the file dune-project while developing.

Todo:
- [ ] test all the changed version methods work as expected


ocaml/util/xapi_version.ml
```diff
@@ -18,14 +18,30 @@ let git_id = ""
...
+      let str = Build_info.V1.Version.to_string v in
+      let version =
+        if String.starts_with ~prefix:"v" str then
+          String.sub str 1 (String.length str - 1)
+        else
+          str
+      in
+      try
+        let maj, min =
+          Scanf.sscanf version "%d.%d.%s" (fun maj min _rest -> (maj, min))
```
lindig, line 37: This could be handled with a regular expression that accounts for an optional `v` prefix by simply matching the desired version. `sscanf` is too weak to do this. The disadvantage of using `Re.Posix` would be that conversion from string to int would have to be done explicitly - so this is probably ok.

lindig (approved):
Good to see the amount ot version machinery reduced.

ocaml/util/xapi_version.ml
lindig:
```

**Comment by psafont** [comment]
Unfortunately the mechanism doesn't work for xapi-storage-script:

```
# /usr/sbin/xcp-rrdd --version
22.25.0.11.g08408edf3
# /usr/sbin/xcp-networkd --version
22.25.0.11.g08408edf3
# /usr/sbin/xenopsd-xc --version
22.25.0.11.g08408edf3
# /usr/sbin/xapi-storage-script --version
0.0.dev
```

Trying to figure out how is it failing

**Comment by psafont** [comment]
The versions from the latest release are worse:
```
# /usr/sbin/xcp-rrdd --version
n/a
# /usr/sbin/xcp-networkd --version
22.25.0
# /usr/sbin/xenopsd-xc --version
22.25.0
# /usr/sbin/xapi-storage-script --version
0.0.0
```

**Comment by psafont** [comment]
Not sure why xapi-storage-script is reporting 0.0.dev as it seems to be built like the rest of the packages. in any case it's as broken as in master branch, so I don't think it should block the merge

File: ocaml/idl/gen_lifecycle.ml [hunk 2]
```diff
@@ -15,7 +15,7 @@
 open Datamodel_types
 
 let current_version =
-  Scanf.sscanf Xapi_version.version "%s@-%s" (fun v _rest -> v)
+  Xapi_version.(Printf.sprintf "%d.%d" xapi_version_major xapi_version_minor)
```
robhoes, line 18: We'll lose the micro version here: the old version of the function kept everything before the `-`.

psafont: That commit has been dropped

robhoes (approved):

PR #4780, opened 2022-08-26
Introduce VTPM for guests

psafont:
The feature is experimental. Currently the contents of the TPMs are stored in xapi's database, this means they are not always up-to-date which makes the feature unable to work with HA and with VM snapshots. Cloning and checkpoints work, this means that migrations work as well.

vtpms are always associated with a VM and it's not possible to change the VM once set. The VTPM can only be associated to a VM if the VM is halted.
vtpms have the fields `unique` and `secure`. Unique can be set at creation time, is false by default, and will avoid creating copies of the contents of the vTPMs. secure is not configurable and defaults to false

To enable the experimental feature `xe-enable-experimental-feature vtpm` has to be run on the pool coordinator. In case the coordinator changes it's better if this is run on all hosts.

The last commit introduces `swtpm-wrapper` into the repo, this has been moved from the .spec repo, and the former will need a change to remove its own copy when including these changes.


lindig (approved):

lindig:
A lot of code that has been reviewed previously. I mostly glanced over it and looked more closely at some error handling.

ocaml/database/db_upgrade.ml
```diff
@@ -12,9 +12,7 @@
  * GNU Lesser General Public License for more details.
  *)
 
-module D = Debug.Make (struct
-  let name = "xapi" (* this is set to 'xapi' deliberately! :) *)
```
robhoes, line 16: And now it has been changed away from 'xapi' deliberately? :)

psafont: I spelunked and found no justification for it in the commit that introduced the module. It's more helpful to have the actual module that prints the loglines, same reasoning I used to change all the unjustified ones 3 years back, it just helps whith triaging

ocaml/idl/datamodel_vtpm.ml
```diff
@@ -0,0 +1,98 @@
...
+open Datamodel_roles
+
+let persistence_backend =
+  Enum ("persistence_backend", [("xapi", "This VTPM is persisted in XAPI's DB")])
+
+let create =
+  call ~name:"create"
+    ~lifecycle:
+      [
+        (Published, rel_rio, "")
```
robhoes, line 26: In reality, the VTPM class and anything in it wasn't actually "published" in that it has been completely unusable until now. The class was there in the datamodel, but the create and destroy functions immediately raise NOT_IMPLEMENTED. No one will have a xapi DB with VTPM rows in it and state.db files will always have ` <table name="VTPM"/>`.

I would therefore consider this class to be introduced _now_, and as a prototype. This signals that any part of the class can still change.

I'd therefore like to clear all `lifecycles` in this module completely. Recording changes from Rio here doesn't really add much value.

psafont: Cool, I'll change that.

PR #4781, opened 2022-08-30
CP-40669: Restore database should not require ref as the first

liulinC:
attribute of row

When restore xapi database from xml file, xapi require ref be
the first attribute of row, This is not flexible, especially when
need to parse and save the xml files.

This commit fix this issue by search ref in the attribute list

### DIFF
--- a/ocaml/database/db_xml.ml
+++ b/ocaml/database/db_xml.ml
@@ -146,14 +146,23 @@ module From = struct
               f acc
           | (_, "table"), [((_, "name"), tblname)] ->
               f (tableset, Table.empty, tblname, manifest)
-          | (_, "row"), ((_, "ref"), rf) :: rest ->
-              (* Remove any other duplicate "ref"s which might have sneaked in there *)
-              let rest = List.filter (fun ((_, k), _) -> k <> "ref") rest in
+          | (_, "row"), rest ->
+              let ref_l, rest =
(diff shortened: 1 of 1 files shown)


ocaml/database/db_xml.ml
```diff
@@ -146,15 +146,24 @@ module From = struct
               f acc
           | (_, "table"), [((_, "name"), tblname)] ->
               f (tableset, Table.empty, tblname, manifest)
-          | (_, "row"), ((_, "ref"), rf) :: rest ->
-              (* Remove any other duplicate "ref"s which might have sneaked in there *)
-              let rest = List.filter (fun ((_, k), _) -> k <> "ref") rest in
+          | (_, "row"), rest ->
```
lindig, line 149: Altogether, this code is traversing lists multiple times and is inefficient if these lists are long. But that problem is already pre-existing.

The problem at hand is to find the first "ref" entry and to ignore the others. The solution is to split the list based on the key and we expect to find at least one entry. I think the code achieving this could be kept closer together - a sketch is below:

```
let ref, rest =
  match List.partition (fun ((_, k) -> k = "ref")) rest with
  | r::_, rest -> r,rest
  | [] -> raise ...
```

lindig (approved):

psafont (approved):

robhoes (approved):

PR #4782, opened 2022-09-05
CP-40402: Move C# and Powershell SDK Generation to .NET Standard 2.0 / .NET 6.0

danilo-delbusso:
Best reviewed one commit at a time.

If you want to run this and don't know how, feel free to ping me.

### C# SDK(move to  .NET Standard 2.0)

1. Remove all mentions of using `CookComputing.XmlRpc`
2. Remove `IXmlRpcProxy` and other `XmlRpc` mentions from `Proxy.cs`
3. Remove XmlRpc references in `Event.cs`, `Response.cs`, and `Proxy.cs`
   - `Event.cs`: `XmlRpcMissingMapping`, `XmlRpcMember`, calls to `session.XmlRpcProxy`
   - `Response.cs`: `XmlRpcMissingMapping`
6. Covert `csproj` to SDK style
7. Target `netstandard2.0`
8. Remove reference mention to `CookComputing.XmlRpcV2` in `csproj`
9.  Remove XmlRpc references in `gen_csharp_binding.ml`:
   - Remove XmlRpc returns
   - Remove XmlRpc references in `Maps.cs` code generation
   - Remove XmlRpc references in `Proxy.cs` code generation
13. Update `Session.cs` to only use JsonRpc
14. Update `README`
15. Update target framework of `XenSdkSample`
____________________________________

### Powershell SDK (Move to .NET 6.0):
The Powershell SDK must be moved to .NET Core to satisfy requirements. 

.NET Core 6.0 is the latest available LTS.

**n.b.: Visual Studio 2022 is required to build to .NET 6.0.**

1. Remove use of `WinForms` in `Wait-XenTask.cs`
2. Make sure `ShouldContinue` does not run in a separate thread when called in `Connect-Xenserver.cs` (in .NET Core, the `ValidateServerCertificate` call runs in a separate thread. This causes an `InvalidOperationException` when calling `ShouldContinue`, which must be called within the same thread as `ProcessRecord`)
3. Convert `csproj` to SDK style
4. Target `net6.0`
5. Remove `CookComputing.XmlRpcV2` from references in project file
6. Update `README`
7. Update `.psd1`


CitrixChris (approved):

ocaml/sdk-gen/csharp/templates/XenServer.csproj.mustache
```diff
@@ -1,96 +1,32 @@
...
-  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
-  <!-- To modify your build process, add your task inside one of the targets below and uncomment it.
-       Other similar extension points exist, see Microsoft.Common.targets.
-  <Target Name="BeforeBuild">
-  </Target>
-  <Target Name="AfterBuild">
-  </Target>
-  -->
-</Project>
+</Project>
```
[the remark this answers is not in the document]
danilo-delbusso: Looks like the generated file has it, too. Must be a copy/paste error. I'll add it to avoid the same confusion appearing in case we need to amend the file in the future 💯

ocaml/sdk-gen/csharp/autogen/src/JsonRpc.cs
```diff
@@ -206,13 +206,13 @@ private T Rpc<T>(string callName, JToken parameters, JsonSerializer serializer)
             webRequest.Proxy = WebProxy;
             webRequest.KeepAlive = KeepAlive;
             webRequest.UserAgent = UserAgent;
-            webRequest.ConnectionGroupName = ConnectionGroupName;
-            webRequest.ProtocolVersion = ProtocolVersion;
+            webRequest.ConnectionGroupName = ConnectionGroupName ?? webRequest.ConnectionGroupName;
```
kc284, line 209: What is the reason behind the change in these 3 lines? This undoes partially the commit https://github.com/xapi-project/xen-api-sdk/commit/f094386763b41acfb1426eefbcdee551bb151374 (nasty bug...). It's a tricky area that I'd leave untouched unless we have evidence of malfunction.

ocaml/sdk-gen/csharp/autogen/src/Session.cs
```diff
@@ -254,36 +226,31 @@ public override string SaveChanges(Session session, string _serverOpaqueRef, Ses
             throw new Exception("The method or operation is not implemented.");
         }
 
-        public Proxy XmlRpcProxy { get; private set; }
-
         public JsonRpcClient JsonRpcClient { get; private set; }
 
+        [Obsolete("Use opaque_ref instead.")]
```
kc284, line 231: These Obsolete properties were removed before the SDK generation code came back to this repo, how have they been re-introduced in this code? I think these changes may have been done off the wrong base? I'm reviewing commit by commit and right now I'm a the third commit, but please have a look and rebase as necessary before I proceed further.

kc284 (changes requested):

ocaml/sdk-gen/csharp/autogen/src/Session.cs
danilo-delbusso: Indeed this was an issue.

I went through every file change in this PR, and I can confirm only this file seems to have the issue.

I have double checked it and reverted the changes that were not meant to go in. (commit coming soon)

ocaml/sdk-gen/csharp/autogen/src/JsonRpc.cs
danilo-delbusso: I see.

The issue here is that the values I null coalesced are null.

They were initially instantiated based on the values of XmlRpcProxy, but since that doesn't exist anymore, I just assumed the sensible thing was to use whatever deafult value we had.

now, out of the three values:

danilo-delbusso: right so I had a look at the previous behaviour, to check which values we might want to have.

### `ConnectionGroupName`

 This one is actually set to `null` by default, too.
image.
So no need to coalesce it.

### `ProtocolVersion`

The protocol version defaults to 1.1 when using the XmlRpc version of this code, which is the same one that the new requests are set to using the new logic. In this case, if we don't specify it, we'll get an error thrown at us.

I think we should leave it like this:
```csharp
webRequest.ProtocolVersion = ProtocolVersion ?? webRequest.ProtocolVersion;
```

this is the variable's value when logging in with the old SDK:

image


### `CookieContainer`

 this one is the weirdest one. When using XMLRPC, it has a decent default:
image

but when using the value from webrequest, we're not getting any instantiated cookie container. I checked the XMLRPC library, and it's being created with a normal `new CookieContainer()`

While the value can be null without errors, if we want to replicate the old behaviour, we could do this:
```csharp
 webRequest.CookieContainer = Cookies ?? webRequest.CookieContainer ?? new CookieContainer(); 
 ````
 
 the code above results in the following initial value of the cookie container, which is the same as we previously had. 
 
image


_______________________________

If this all works for you, I'll push a new commit with the tweaks.

File: ocaml/sdk-gen/powershell/autogen/XenServerPSModule.psd1 [hunk 4]
```diff
@@ -41,29 +41,27 @@ CompanyName = 'Citrix Systems, Inc'
 Copyright = 'Copyright (c) Citrix Systems, Inc. All rights reserved.'
 
 # Requirements
-PowerShellVersion = '4.0'
+PowerShellVersion = '7.2'
 PowerShellHostName = ''
 PowerShellHostVersion = ''
-DotNetFrameworkVersion = '4.5'
-CLRVersion = '4.0.30319'
+DotNetFrameworkVersion = '6.0'
```
kc284, line 47: Is this still needed/valid?

ocaml/sdk-gen/powershell/autogen/XenServerPSModule.psd1
```diff
@@ -41,29 +41,27 @@ CompanyName = 'Citrix Systems, Inc'
...
 ProcessorArchitecture = 'None'
 
 #Contents
 ModuleToProcess = 'XenServerPowerShell.dll'
 RequiredModules = @()
 NestedModules = @()
 ModuleList = @('XenServerPowerShell.dll')
-RequiredAssemblies = @('CookComputing.XmlRpcV2.dll',
-                       'Newtonsoft.Json.CH.dll',
+RequiredAssemblies = @('System.Management.Automation.dll',
```
kc284, line 55: I presume this is needed for the Linux case? (Previously it was not necessary for Windows).

ocaml/sdk-gen/powershell/templates/XenServerPowerShell.csproj.mustache
```diff
@@ -1,75 +1,27 @@
...
-    <DefineConstants>TRACE</DefineConstants>
-    <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>4</WarningLevel>
-    <Prefer32Bit>false</Prefer32Bit>
-  </PropertyGroup>
   <ItemGroup>
-    <Reference Include="System" />
-    <Reference Include="System.Data" />
-    <Reference Include="System.Management" />
+    <Reference Include="Newtonsoft.Json.CH">
```
kc284, line 14: This is a dependency in XenServer rather than XenServerPowershell project, in fact only recently was it removed from the latter. It doubt it should be needed now.

ocaml/sdk-gen/powershell/autogen/XenServerPSModule.psd1
```diff
@@ -41,29 +41,23 @@ CompanyName = 'Citrix Systems, Inc'
...
 ProcessorArchitecture = 'None'
 
 #Contents
 ModuleToProcess = 'XenServerPowerShell.dll'
 RequiredModules = @()
 NestedModules = @()
-ModuleList = @('XenServerPowerShell.dll')
-RequiredAssemblies = @('CookComputing.XmlRpcV2.dll',
-                       'Newtonsoft.Json.CH.dll',
+RequiredAssemblies = @('Newtonsoft.Json.CH.dll',
```
kc284, line 53: The latest commit has also removed the ModuleList - was this on purpose or accidental? Is it not needed?

danilo-delbusso: It was on purpose. Not only does it seem like it's not needed, but it was flagged by the `Test-ModuleManifest` cmdlet

ocaml/sdk-gen/csharp/autogen/src/JsonRpc.cs
```diff
@@ -207,12 +207,12 @@ private T Rpc<T>(string callName, JToken parameters, JsonSerializer serializer)
             webRequest.KeepAlive = KeepAlive;
             webRequest.UserAgent = UserAgent;
             webRequest.ConnectionGroupName = ConnectionGroupName;
-            webRequest.ProtocolVersion = ProtocolVersion;
+            webRequest.ProtocolVersion = ProtocolVersion ?? webRequest.ProtocolVersion;
             webRequest.ServicePoint.Expect100Continue = Expect100Continue;
             webRequest.AllowAutoRedirect = AllowAutoRedirect;
```
kc284, line 212: While debugging I noticed that in Xmlrpc the default for AllowAutoRedirect was true. I'm not certain whether it is needed for the server API, but let's keep it as it was to be on the safe side. The place to set it in when we create the Jsonrpc client in the Session constructor.

CitrixChris (approved):

robhoes (approved):
I haven't personally reviewed this, but several people who actually know about this code have approved.

PR #4784, opened 2022-09-07
Update CI badges in README

lindig:
We moved to the GitHub CI - so use them to reflect the build status. Also removed the LoC badges because it appeared unreliable to me.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -1,7 +1,7 @@
 Xapi Project's XenAPI Management Toolstack
 ==========================================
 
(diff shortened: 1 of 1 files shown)


psafont:
coverage hasn't been reported since may 2019, we could remove this, or maybe james could take on the task of enable coverage reporting using the new mechanism together with ppx_bisect

psafont (approved):

edwintorok (approved):

PR #4789, opened 2022-09-15
CA-370575: [XSI-1310] Driver disks / supp packs applied at host

liulinC:
install time cannot be updated

When applied at host install time, the yum is configured with `group_command=compat`, thus yum does not have group membership information if `group_command=objects` is set during update.apply

This commit set `group_command=compat` to support Driver disks/ supp packs applied
* at host install time
* manually (install some packages)

### DIFF
--- a/ocaml/xapi/xapi_pool_update.ml
+++ b/ocaml/xapi/xapi_pool_update.ml
@@ -220,13 +220,14 @@ let with_api_errors f x =
       error "%s" msg ;
       raise (Api_errors.Server_error (Api_errors.invalid_update, [errinfo]))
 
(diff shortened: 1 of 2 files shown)


rosslagerwall (approved):

psafont (approved):
There don't seem to be coding errors being introduced here, I can't speak for the correctness of the logic here as I'm not a subject expert

liulinC:
Thanks @psafont and @rosslagerwall  for the help to review, 
@lindig @edwintorok @robhoes Can I have another review/approve from one of you as the `writer approve`

robhoes (approved):

PR #4790, opened 2022-09-16
xenopsd tests: don't run cancel_utils_test as unit test

psafont:
In practice the watches code was not being exercised by the test in any situation, so transforming it into an executable that doesn't get compiled by default. It is still checked by the CI

@stormi this should fix your failures


psafont:
Sorry, I meant it's not installed by default.

lindig (approved):

PR #4791, opened 2022-09-16
CA-370731: allow only one database entry per ca certificate

psafont:
Previously when overwriting a ca certificate the old object referring to the certificate was not deleted. This meant that when uninstalling that ca certificate, after deleting the certificate file a failure was encountered instead of removing all the database objects referring to the file.

This PR now removes the old certificate objects when installing a new one and removes all of them when the certificate is uninstalled.

Note: the pool join is blocked if there are conflicting certificates so there's no worry of overwriting information in that case

### DIFF
--- a/ocaml/xapi/certificates.ml
+++ b/ocaml/xapi/certificates.ml
@@ -233,7 +233,7 @@ end = struct
     let certs =
       match get_ca_certs ~__context name with
       | [x] ->
-          x
+          [x]
       | [] ->
           D.error "unable to find certificate with name='%s'" name ;
           raise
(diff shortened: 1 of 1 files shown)


ocaml/xapi/certificates.ml
```diff
@@ -239,19 +247,9 @@ end = struct
           D.error
```
edwintorok, line 247: Should this be a warning now that we don't raise anymore?

edwintorok (approved):
would be nice if the DB was able to enforce the desired uniqueness invariant here.

ocaml/xapi/certificates.ml
psafont: rethinking about it... yes, there are too many scary errors around the logs, let's mark it as warning

lindig (approved):

robhoes (approved):

PR #4794, opened 2022-09-27
CP-33973 disable DMC: adjust unit test to disable DMC feature 

lindig:
The handling of DMC is guarded by a corresponding license feature flag. But this is not true for the database upgrade rule that is triggered during a version upgrade - it unconditionally disables DMC on all halted VMs. Change this to respect the feature flag as well. This makes backporting this code to the LTSR easier where DMC is still supported.

### DIFF
--- a/ocaml/tests/test_xapi_db_upgrade.ml
+++ b/ocaml/tests/test_xapi_db_upgrade.ml
@@ -17,7 +17,9 @@ module X = Xapi_db_upgrade
 module Date = Xapi_stdext_date.Date
 
 let upgrade_vm_memory_for_dmc () =
-  let __context = T.make_test_database () in
+  (* disable DMC for unit tests *)
+  let features = Features.(List.filter (( <> ) DMC) all_features) in
+  let __context = T.make_test_database ~features () in
   let self = List.hd (Db.VM.get_all ~__context) in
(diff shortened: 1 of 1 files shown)


lindig:
A unit test is failing - will look into it

ocaml/xapi/xapi_db_upgrade.ml
```diff
@@ -253,7 +253,10 @@ let upgrade_vm_memory_for_dmc =
           else
             ()
         in
-        List.iter update_vm (Db.VM.get_all_records ~__context)
+        if Pool_features.is_enabled ~__context Features.DMC then
```
robhoes, line 256: I don't think this works as expected, as on startup, the DB upgrade runs before the licensing refresh that turns the feature off.

lindig: Ok - so maybe we need to use the version constraint that is part of the rules mechanism.

lindig: I'm reducing this to just changing the unit test to use a disabled DMC, as that is the current scenario. The upgrade rule might have to be removed if we backport this code.

psafont (approved):

robhoes (approved):

PR #4795, opened 2022-09-27
CA-370858: disallow VM exports with VTPMs attached

psafont:
We're unable to serialize the data because the field for the contents is not exposed in the API and it's based on a secret, which can be dangerous once it's been implemented.

Exports are exposed using an HTTP endpoint, this means it's an indirect operation and that other operations that use the  feature will fail in extraneous ways in a non-instantaneous way.

To avoid this the two methods that use it in xapi are changed as well (vm-export and VM cross-pool migrations). This makes the failure immediate and clear.


psafont:
Superceded by https://github.com/xapi-project/xen-api/pull/4796

PR #4796, opened 2022-09-29
CP-39744, CA-370858: Block actions depending on unimplemented VTPM functionality

psafont:
CP-39744: Block BIOS VMs with vTPM attached from booting
The Physical Presence Interface part of the vTPM specification is only
implemented for UEFI guests, other functionality may be missing.

Also makes the check that raises exceptions centralised and makes it togglable using /etc/sysconfic/xapi

CA-370858: disallow VM exports with VTPMs attached
We're unable to serialize the data because the field for the contents is not exposed in the API and it's based on a secret, which can be dangerous once it's been implemented.

Exports are exposed using an HTTP endpoint, this means it's an indirect operation and that other operations that use the feature will fail in extraneous ways in a non-instantaneous way.

To avoid this the two methods that use it in xapi are changed as well (vm-export and VM cross-pool migrations). This makes the failure immediate and clear.

### DIFF
--- a/ocaml/xapi-cli-server/cli_operations.ml
+++ b/ocaml/xapi-cli-server/cli_operations.ml
@@ -5794,6 +5794,13 @@ let export_common fd _printer rpc session_id params filename num ?task_uuid
   in
   let vm_metadata_only = get_bool_param params "metadata" in
   let vm_record = vm.record () in
+  (* disallow exports and cross-pool migrations of VMs with VTPMs *)
+  ( if vm_record.API.vM_VTPMs <> [] then
+      let message = "Exporting VM metadata with VTPMs attached" in
+      (* Helpers.maybe_raise_vtpm_unimplemented cannot be used due to the
+         xapi_globs dependence *)
+      raise Api_errors.(Server_error (not_implemented, [message]))
+  ) ;
   let exporttask, task_destroy_fn =
     match task_uuid with
     | None ->
--- a/ocaml/xapi/xapi_globs.ml
+++ b/ocaml/xapi/xapi_globs.ml
@@ -926,6 +926,8 @@ let repository_gpgcheck = ref true
 
 let migration_compression = ref false
 
+let ignore_vtpm_unimplemented = ref false
+
 let evacuation_batch_size = ref 10
 
 type xapi_globs_spec_ty = Float of float ref | Int of int ref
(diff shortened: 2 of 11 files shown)


ocaml/xapi/xapi_globs.ml
```diff
@@ -1364,8 +1366,16 @@ let other_options =
...
     , "The number of VMs evacauted from a host in parallel."
     )
+  ; ( "ignore-vtpm-unimplemented"
+    , Arg.Set ignore_vtpm_unimplemented
+    , (fun () -> string_of_bool !ignore_vtpm_unimplemented)
+    , "Do not raise errors on use-cases where VTPM codepaths are not finished."
+    )
   ]
 
+(* The options can be set with the variable xapiflags in /etc/sysconfig/xapi.
```
edwintorok, line 1376: I think they can also be set in `/etc/xapi.conf`, that is what I usually change.

edwintorok (approved):

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -5794,6 +5794,13 @@ let export_common fd _printer rpc session_id params filename num ?task_uuid
   in
   let vm_metadata_only = get_bool_param params "metadata" in
   let vm_record = vm.record () in
+  (* disallow exports and cross-pool migrations of VMs with VTPMs *)
+  ( if vm_record.API.vM_VTPMs <> [] then
+      let message = "Exporting VM metadata with VTPMs attached" in
```
edwintorok, line 5799: We intend to implement this, so as a temporary measure this is probably fine (we limit the libraries that the CLI server can access on purpose to avoid it accidentally bypassing RBAC checks)

edwintorok (approved):

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -1167,6 +1167,10 @@ let migrate_send' ~__context ~vm ~dest ~live:_ ~vdi_map ~vif_map ~vgpu_map
       true
     with _ -> false
   in
+  ( if (not is_intra_pool) && Db.VM.get_VTPMs ~__context ~self:vm <> [] then
```
robhoes, line 1170: It's better to do this in `assert_can_migrate` (below).

ocaml/xapi/xapi_xenops.ml
```diff
@@ -1261,6 +1261,14 @@ module MD = struct
       else
         platformdata
     in
+    (* BIOS guests don't seem to detect the attached VTPM, block them *)
+    ( match (firmware, vm.API.vM_VTPMs) with
+    | Xenops_types.Vm.Bios, _ :: _ ->
+        let message = "Booting BIOS VM with VTPMs attached" in
+        Helpers.maybe_raise_vtpm_uninmplemented __FUNCTION__ message
```
lindig, line 1268: Should this include a reference to the VM?

lindig (approved):

ocaml/xapi-cli-server/cli_operations.ml
robhoes: But why do you do this here in the CLI, while the export handle below already has the check?

ocaml/xapi/xapi_xenops.ml
psafont: I don't think it's needed, the whole action to start a VM fails, this already signals the VM that fails to the user, both in CLI and GUI

PR #4797, opened 2022-09-29
CP-40767 CP-40429 Migration Compression - define Zstd.Fast

lindig:
Testing of vCPU/vGPU compression reveals that it does not offer a benefit on 10+ GBit/s networks because we can't saturate the network. So compression causes a slowdown compared to uncompressed migration.

So far:

* We are spawning a zstd process for compression with default parameters
* The default compression level is 3, on a scale 1 to 19, so 1 would be fast/low-compression and 19 slow/high-compression.

Define and use a new compression profile Zstd.Fast that uses less CPU; it causes less slowdown in the case of fast networks. The benefit on slow networks needs to be re-evaluated but this appears to be a better trade off.

TRUNK

evacuating xrtmia-04-21 without compression
real    3m44.421s
real    3m39.811s

evacuating xrtmia-04-24 with compression Zstd.Default
real    4m42.191s
real    4m58.360s

ZSTD.ADAPTIVE

evacuating xrtmia-04-21 (in batches of 20) without compression
real    3m48.964s
real    3m47.949s

evacuating xrtmia-04-24 (in batches of 20) with compression Zstd.Adaptive
real    5m59.555s
real    6m15.327s

This is clearly worse than what we have on Trunk.

ZSTD.FAST

evacuating xrtmia-04-21 (in batches of 20) without compression
real    3m39.868s
real    3m44.444s

evacuating xrtmia-04-24 (in batches of 20) with compression Zstd.Fast
real    4m8.825s
real    4m5.427s

This is substantially better than what we have on Trunk. There is still a slowdwown against uncompressed migration but it is much smaller.

Trunk:          4:42 / 3:44 = 1.26 (26% slowdown)
Zstd.Adaptive:  5:59 / 3:48 = 1.57 (57% slowdown)
Zstd.Fast:      4:08 / 3:39 = 1.13 (13% slowdown)

### DIFF
--- a/ocaml/libs/zstd/zstd.ml
+++ b/ocaml/libs/zstd/zstd.ml
@@ -20,3 +20,33 @@ module Default = Xapi_compression.Make (struct
 
   let decompress_options = ["--decompress"; "--stdout"]
 end)
+
+module Adaptive = Xapi_compression.Make (struct
+  (** Path to the zstd binary *)
+  let executable = "/usr/bin/zstd"
+
+  (* default compression level is 3 on a scale 1..19 with lower=faster
+     and higher=more compression. By default 1 thread is used *)
+  let compress_options = ["--adapt=min=1,max=4"; "--threads=1"]
+
+  let decompress_options = ["--decompress"; "--stdout"]
+end)
+
+module Fast = Xapi_compression.Make (struct
+  (** Path to the zstd binary *)
+  let executable = "/usr/bin/zstd"
+
+  (** --threads=2 doesn't improve speed *)
+  let compress_options = ["--fast"]
+
+  let decompress_options = ["--decompress"; "--stdout"]
+end)
+
+module Null = Xapi_compression.Make (struct
+  (** Path to the zstd binary *)
+  let executable = "/usr/bin/cat"
+
+  let compress_options = []
+
+  let decompress_options = []
+end)
--- a/ocaml/libs/zstd/zstd.mli
+++ b/ocaml/libs/zstd/zstd.mli
@@ -13,3 +13,9 @@
  *)
 
 module Default : Xapi_compression.COMPRESSOR
+
+module Adaptive : Xapi_compression.COMPRESSOR
+
+module Fast : Xapi_compression.COMPRESSOR
+
+module Null : Xapi_compression.COMPRESSOR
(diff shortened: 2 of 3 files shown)


ocaml/libs/zstd/zstd.ml
```diff
@@ -20,3 +20,23 @@ module Default = Xapi_compression.Make (struct
...
   let decompress_options = ["--decompress"; "--stdout"]
 end)
+
+module Adaptive = Xapi_compression.Make (struct
+  (** Path to the zstd binary *)
+  let executable = "/usr/bin/zstd"
+
+  (* default compression level is 3 on a scale 1..19 with lower=faster
+     and higher=more compression. By default 1 thread is used *)
+  let compress_options = ["--adapt=min=1,max=4"; "--threads=2"]
```
edwintorok, line 30: Apparently contrary to the manpage `min=0` is accepted too, might be interesting to see if it has any effect.
If you have a machine and a dump of the stream you'd be compressing you can check the relative speeds this way:
```
zstd -b0 -e4 yourfile
```

And it should print the speed and resulting size for each one.

There is also `lz4` which although doesn't support adaptive compression supports the same `-b` and `-e` flags

edwintorok: Although from a quick test level 0 results in weird speed (slower than 1!), so maybe best avoided and use just the values from the manpage as you've done here:
```
zstd -b0 -e2 foo.tar; lz4 -b0 -e2 foo.tar;
 0#foo.tar           :  30853120 ->   6520977 (4.731), 165.5 MB/s , 656.7 MB/s
 1#foo.tar           :  30853120 ->   7315797 (4.217), 243.4 MB/s , 817.8 MB/s
 2#foo.tar           :  30853120 ->   6971746 (4.425), 213.0 MB/s , 721.8 MB/s
Benchmarking levels from 0 to 2
 0#foo.tar           :  30853120 ->  10587545 (2.914), 394.3 MB/s ,2302.8 MB/s
 1#foo.tar           :  30853120 ->  10587545 (2.914), 396.0 MB/s ,2308.5 MB/s
 2#foo.tar           :  30853120 ->  10587545 (2.914), 398.3 MB/s ,2230.9 MB/s```

edwintorok: Try moving `--threads=2` to the `--fast` section.

edwintorok (approved):

psafont (approved):

PR #4799, opened 2022-09-30
Private/jameshen/host https only

jameshensmancitrix:
Added https_only field to host as well as the setter

### DIFF
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -1926,6 +1926,7 @@ let t =
             ~doc_tags:[Networking] "address"
             "The address by which this host can be contacted from any other \
              host in the pool"
+        ; field ~qualifier:DynamicRO ~ty:Bool ~default_value:(Some (VBool false)) "https_only" "Reflets wheather port 80 is open (false) or not (true)"
         ; field ~qualifier:DynamicRO ~ty:(Ref _host_metrics) "metrics"
             "metrics associated with this host"
         ; field ~in_oss_since:None ~qualifier:DynamicRO
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -1727,6 +1727,23 @@ let copy_primary_host_certs =
       ]
     ~allowed_roles:_R_POOL_ADMIN ~hide_from_docs:true ~pool_internal:true ()
 
+let set_https_only =
+  call ~name:"set_https_only"
+    ~doc:
+      "updates the host firewall to open or close port 80 depending on the \
+       value"
+    ~lifecycle:[]
+    ~params:
+      [
+        (Ref _host, "self", "The Host")
+      ; ( Bool
+        , "value"
+        , "true - http port 80 will be blocked, false - http port 80 will be \
+           open"
+        )
+      ]
+    ~allowed_roles:_R_POOL_OP ()
+
 (** Hosts *)
 let t =
   create_obj ~in_db:true ~in_product_since:rel_rio ~in_oss_since:oss_since_303
@@ -1860,6 +1877,7 @@ let t =
       ; cert_distrib_atom
       ; apply_updates
       ; copy_primary_host_certs
+      ; set_https_only
       ]
     ~contents:
       ([
@@ -1926,7 +1944,6 @@ let t =
             ~doc_tags:[Networking] "address"
             "The address by which this host can be contacted from any other \
              host in the pool"
-        ; field ~qualifier:DynamicRO ~ty:Bool ~default_value:(Some (VBool false)) "https_only" "Reflets wheather port 80 is open (false) or not (true)"
         ; field ~qualifier:DynamicRO ~ty:(Ref _host_metrics) "metrics"
             "metrics associated with this host"
         ; field ~in_oss_since:None ~qualifier:DynamicRO
@@ -2078,6 +2095,9 @@ let t =
             "last_software_update"
             ~default_value:(Some (VDateTime (Date.of_float 0.0)))
             "Date and time when the last software update was applied"
+        ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:Bool
+            ~default_value:(Some (VBool false)) "https_only"
+            "Reflects whether port 80 is open (false) or not (true)"
         ]
       )
     ()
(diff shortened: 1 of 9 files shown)


ocaml/idl/datamodel_host.ml
```diff
@@ -1727,6 +1727,16 @@ let copy_primary_host_certs =
       ]
     ~allowed_roles:_R_POOL_ADMIN ~hide_from_docs:true ~pool_internal:true ()
 
+let set_https_only =
+  call ~name:"set_https_only" ~doc:"updates the host firewall to open or close port 80 depending on the value" 
+  ~lifecycle:[]
+  ~params:
+  [
+      (Ref _host, "host", "the host that the https_only setting is to be set")
+    ; (Bool, "value", "host.https_only will be updated according to this value")
```
lindig, line 1736: I think the doc-strings could be less self-referential: "the host that the https_only setting is to be set" -> "host whose firewall port settings to update"; "host.https_only will be updated according to this value" -> "Boolean for https_only; true - http port 80 will be blocked, false - http port 80 will be open".

ocaml/xapi-cli-server/records.ml
```diff
@@ -2968,6 +2968,7 @@ let host_record rpc session_id host =
           ()
       ; make_field ~name:"hostname" ~get:(fun () -> (x ()).API.host_hostname) ()
       ; make_field ~name:"address" ~get:(fun () -> (x ()).API.host_address) ()
+      ; make_field ~name:"https_only" ~get:(fun () -> string_of_bool (x ()).API.host_https_only) ~set:(fun s -> Client.Host.set_https_only ~rpc ~session_id ~host ~value:(Stdlib.bool_of_string s) ) ()
```
lindig, line 2971: Are these lines not folded?

lindig (approved):

ocaml/idl/datamodel_host.ml
```diff
@@ -1926,6 +1937,7 @@ let t =
             ~doc_tags:[Networking] "address"
             "The address by which this host can be contacted from any other \
              host in the pool"
+        ; field ~qualifier:DynamicRO ~ty:Bool ~default_value:(Some (VBool false)) "https_only" "Reflets wheather port 80 is open (false) or not (true)"
```
robhoes, line 1940: Please add new fields to the end of the list. Also there are spelling mistakes in the doc string.

robhoes: I think you also need to add `~lifecycle:[]` here to get datamodel_lifecycle.ml updated properly.

ocaml/idl/datamodel_host.ml
```diff
@@ -1727,6 +1727,16 @@ let copy_primary_host_certs =
       ]
     ~allowed_roles:_R_POOL_ADMIN ~hide_from_docs:true ~pool_internal:true ()
 
+let set_https_only =
+  call ~name:"set_https_only" ~doc:"updates the host firewall to open or close port 80 depending on the value" 
+  ~lifecycle:[]
+  ~params:
+  [
+      (Ref _host, "host", "the host that the https_only setting is to be set")
```
robhoes, line 1735: The param name should be `self`, not `host`. For the doc string, probably just have "The host".

ocaml/xapi-cli-server/records.ml
robhoes: Indeed, `make format` should be used.

ocaml/xapi/message_forwarding.ml
```diff
@@ -3961,6 +3961,11 @@ functor
         let uuid = host_uuid ~__context self in
         info "Host.apply_updates: host = '%s'; hash = '%s'" uuid hash ;
         Local.Host.apply_updates ~__context ~self ~hash
+
+      let set_https_only ~__context ~host ~value = 
+        info "setting https_only to %B" value;
```
robhoes, line 3966:
```suggestion
        info "Host.set_https_only: self = '%s';  value = %b" self value ;
```
(as per convention)

ocaml/xapi-cli-server/records.ml
```diff
@@ -2968,6 +2968,13 @@ let host_record rpc session_id host =
           ()
       ; make_field ~name:"hostname" ~get:(fun () -> (x ()).API.host_hostname) ()
       ; make_field ~name:"address" ~get:(fun () -> (x ()).API.host_address) ()
+      ; make_field ~name:"https_only"
```
robhoes, line 2971:
```suggestion
      ; make_field ~name:"https-only"
```

ocaml/xapi/message_forwarding.ml
psafont: Even more handily:
```suggestion
        info "%s: self = '%s';  value = %b" __FUNCTION__ self value ;
```

ocaml/idl/datamodel_host.ml
```diff
@@ -2077,6 +2095,9 @@ let t =
             "last_software_update"
             ~default_value:(Some (VDateTime (Date.of_float 0.0)))
             "Date and time when the last software update was applied"
+        ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:Bool
+            ~default_value:(Some (VBool false)) "https_only"
+            "Reflets wheather port 80 is open (false) or not (true)"
```
psafont, line 2100:
```suggestion
            "Reflects whether port 80 is open (false) or not (true)"
```

ocaml/idl/datamodel_host.ml
```diff
@@ -1727,6 +1727,23 @@ let copy_primary_host_certs =
...
+      "updates the host firewall to open or close port 80 depending on the \
+       value"
+    ~lifecycle:[]
+    ~params:
+      [
+        (Ref _host, "self", "The Host")
+      ; ( Bool
+        , "value"
+        , "Boolean for https_only; true - http port 80 will be blocked, false \
+           - http port 80 will be open"
```
psafont, line 1742: I would remove the first sentence altogether, that information is already displayed elsewhere (it's a boolean by the type and the value controls by its name)

```suggestion
        , "true - blocks http port 80; false - opens http port 80"
```

datamodel_lifecycle.ml
```diff
@@ -0,0 +1,43 @@
+let prototyped_of_class = function
```
robhoes, line 1: This one came back...

robhoes: and gone :)

robhoes (approved):

PR #4800, opened 2022-09-30
Add HTTPS support to Open_uri

robhoes:
This required a bunch of libraries to be split off and moved around (hopefully for the better).


ocaml/xapi-idl/lib/dune
```diff
@@ -12,6 +12,7 @@
    message-switch-unix 
    mtime
    mtime.clock.os
+   open-uri
```
edwintorok, line 15: for consistency should it be `xapi-open-uri`?

ocaml/libs/open-uri/open_uri.ml
```diff
@@ -53,6 +53,20 @@ let with_open_uri uri f =
              (Uri.to_string uri)
           )
   )
+  | Some "https" -> (
+    match (Uri.host uri, Uri.port uri) with
+    | Some host, Some port ->
+        Stunnel.with_connect ~verify_cert:(Stunnel_client.pool ()) host port (fun s ->
+          f Safe_resources.Unixfd.(!(s.Stunnel.fd)))
```
edwintorok, line 60: that is not really a safe way to pass that fd, the wrapper might decide to close it at some point, can we instead pass it the `s.Stunnel.fd` directly and let it do the unwrapping internally?

edwintorok:
Looks like a `make format` is needed after the `Uuidx` changes, it probably changed the length of some lines just enough to trigger splitting.

ocaml/xapi-idl/lib/dune
robhoes: Hmm yes, that's probably better.

ocaml/libs/open-uri/open_uri.ml
robhoes: Do you mean that `Safe_resources.Unixfd.(!)` will close it?

I'll take a look, but it comes down to the semantics of the existing function, and the way `f` is used by callers now. If we'd push down the Safe_resources thing, then we'll also need to do that for the other cases.

ocaml/libs/open-uri/open_uri.ml
```diff
@@ -53,6 +53,20 @@ let with_open_uri uri f =
              (Uri.to_string uri)
           )
   )
+  | Some "https" -> (
+    match (Uri.host uri, Uri.port uri) with
+    | Some host, Some port ->
+        Stunnel.with_connect ~verify_cert:(Stunnel_client.pool ()) host port (fun s ->
```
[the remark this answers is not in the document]
robhoes: I've now added `?verify_cert`. It's a bit awkward, as the intention of this function is to be guided by the URL alone, and now takes an argument that is only relevant for one of the cases after parsing the URL. But is seems the best we can do now.

ocaml/libs/open-uri/open_uri.ml
robhoes: The best way forward seems to be to adopt Safe_resources for all fds in this function, but that would be a much bigger change. At least `Stunnel.with_connect` does not close the fd as far as I can see.

ocaml/xapi-idl/lib/dune
robhoes: Done.

psafont:
This is a reminder for myself to push the new .opam files to xs-opam when this is merged

ocaml/libs/open-uri/open_uri.ml
```diff
@@ -53,6 +53,25 @@ let with_open_uri uri f =
              (Uri.to_string uri)
           )
   )
+  | Some "https" -> (
+      let verify_cert =
+        Option.value ~default:(Stunnel_client.pool ()) verify_cert
```
psafont, line 58: I like that users are forced to be explicit about the verification they need if they require something else other than pool verification

psafont (approved):
Changes look good. I think this change needs a build to test and flush out potential issues with open_uris using non-pool hosts

robhoes:
@psafont Thanks. Note that the changes to Open_uri here actually don't yet take effect anywhere, as it is not yet called with `https` URLs. That will be next :)

lindig (approved):
99% of all changes are Uuid/Uuidx. This is best reviewed by commits.

PR #4802, opened 2022-10-03
CP-40751 Added and implemented Pool.set_https_only

jameshensmancitrix:

### DIFF
--- a/ocaml/xapi/xapi_pool.ml
+++ b/ocaml/xapi/xapi_pool.ml
@@ -3596,3 +3596,12 @@ let set_uefi_certificates ~__context ~self ~value =
         )
         (Db.Host.get_all ~__context)
   )
+
+let set_https_only ~__context ~self:_ ~value =
+  Helpers.call_api_functions ~__context (fun rpc session_id ->
+      List.iter
+        (fun host ->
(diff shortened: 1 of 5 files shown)


robhoes:
This generally looks good, but needs a reformat.

ocaml/xapi/xapi_pool.ml
```diff
@@ -3596,3 +3596,12 @@ let set_uefi_certificates ~__context ~self ~value =
...
         (Db.Host.get_all ~__context)
   )
+
+let set_https_only ~__context ~self ~value =
+  Helpers.call_api_functions ~__context (fun rpc session_id ->
+      List.iter
+        (fun host ->
+          Client.Host.set_https_only ~rpc ~session_id ~self:host ~value
+        )
+        (Db.Host.get_all ~__context)
```
psafont, line 3606: Note: it doesn't look like order matters in this case

robhoes: The calls from this host (the coordinator) to all other hosts are always done over HTTPS, so there is no risk of the side-effects of this function caution harm to the completion of the call itself.

robhoes (approved):

psafont (approved):

PR #4806, opened 2022-10-04
CP-40752 Added CLI functionality for pool level setter and getter (https_only)

jameshensmancitrix:


robhoes:
I think you need to rebase, because this includes a commit for "CP-40751 Added and implemented Pool.set_https_only", which is already in.

ocaml/xapi-cli-server/records.ml
```diff
@@ -1430,6 +1430,20 @@ let pool_record rpc session_id pool =
       ; make_field ~name:"repository-proxy-username"
           ~get:(fun () -> (x ()).API.pool_repository_proxy_username)
           ()
+      ; make_field ~name:"https_only"
```
robhoes, line 1433: In the CLI, we use `-` rather than `_`.

ocaml/xapi-cli-server/records.ml
```diff
@@ -1430,6 +1430,20 @@ let pool_record rpc session_id pool =
       ; make_field ~name:"repository-proxy-username"
           ~get:(fun () -> (x ()).API.pool_repository_proxy_username)
           ()
+      ; make_field ~name:"https-only"
+          ~get:(fun () ->
+            Client.Host.get_all ~rpc ~session_id
```
psafont, line 1435: Why isn't a `Client.Pool.get_https_only` enough here? I would rather expect that a Host.set_http_only = false would make the pool field to become false as well.

psafont (approved):

ocaml/xapi-cli-server/records.ml
```diff
@@ -1430,6 +1430,20 @@ let pool_record rpc session_id pool =
...
+            |> List.map (fun h ->
+                   Client.Host.get_https_only ~rpc ~session_id ~self:h
+               )
+            |> List.fold_left ( && ) true
+            |> string_of_bool
```
psafont, line 1440: |> List.for_all (fun h ->
                   Client.Host.get_https_only ~rpc ~session_id ~self:h
               )
            |> string_of_bool

PR #4808, opened 2022-10-05
CP40753 Firewall updates when Host.set_https_only is called

jameshensmancitrix:


ocaml/xapi/xapi_host.ml
```diff
@@ -2927,4 +2927,7 @@ let apply_updates ~__context ~self ~hash =
   warnings
 
 let set_https_only ~__context ~self ~value =
-  Db.Host.set_https_only ~__context ~self ~value
+  Db.Host.set_https_only ~__context ~self ~value ;
```
robhoes, line 2930: Please update the field _after_ changing the firewall. That way you don't have to put back the old value in case the script fails.

jameshensmancitrix: Makes sense, have swapped them around

robhoes (approved):

scripts/plugins/firewall-port
```diff
@@ -19,6 +19,11 @@ PROTOCOL="${3:-tcp}"
 CHAIN="xapi-INPUT"
 RULE="-p $PROTOCOL -m conntrack --ctstate NEW -m $PROTOCOL --dport $PORT -j ACCEPT"
 
+if [ "$PORT" -eq "80" ]
```
lindig, line 22: I think it would be clearer to `CHAIN="xapi-INPUT"` in and `else` branch. Another idea is to use `case` over `$PORT` as this would make adding cases easier.

jameshensmancitrix: Would this new chain be used only for port 80 or on others?

lindig: The current logic could be encoded like this:

```
case "$PORT" in
  80)
    CHAIN="RH-Firewall-1-INPUT"
    ;;
  *)
    CHAIN="xapi-INPUT"
    ;;
esac
```

And my argument is, that this is anticipating future cases. A simpler solution is to use `if then else fi`, which I would prefer over having a default which is over written.

lindig (approved):

PR #4810, opened 2022-10-05
maintenance: explicitely declare direct dependencies

psafont:
Previously a lot of libraries were used but not declared because a dependent library was declared pulled them.

To enable the explicit mode `(implicit_transitive_deps false)` needs to be added to `dune-project`. I believe enabling it requires a high bar of knowledge around libraries, so it's best to leave it disabled. For example, the error when not importing declaring `dune-build-info` when it's needed is indecipherable:
```
File "ocaml/gencert/.gencert.eobjs/build_info_data.ml-gen", line 1:
Error: Could not find the .cmi file for interface
       ocaml/gencert/.gencert.eobjs/build_info_data.ml-gen.
```

While dune does not warn of unused libraries, I've removed some unused libraries, like xapi-idl (because only xapi-log was being used), or threads, which is not needed anymore as dune assumes `-mt` and forces linking against `threads.posix`anyway. Read https://dune.readthedocs.io/en/stable/dune-files.html?highlight=threads#implicit-transitive-deps-1 and https://dune.readthedocs.io/en/stable/advanced-topics.html?highlight=threads#findlib-integration

For users of the gzip and zstd I've decided to reexport the symbols from the library xapi-compression. I saw little value in forcing users to declare both.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,7 +25,7 @@ jobs:
     name: Ocaml tests
     runs-on: ubuntu-20.04
     env:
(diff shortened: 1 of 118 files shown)


lindig (approved):

lindig:
So now that dependencies are correct, does it not make sense to use `(implicit_transitive_deps false)` and be alerted when libraries are missing? Otherwise we will be back where we were quickly.

psafont:
I recommend against that setting, this is because sadly the dune library that provides an ocaml module is often not obvious. While I can guess it by grepping ocamlfind output, this is exceedingly difficult for non-experts.

I understand that it will get out-of-date but I rather have a PR a year updating it until dune includes a helpful error message telling users `library xxx provides module yyy, please add it to the list of dependencies in file zzz`

robhoes (approved):

PR #4811, opened 2022-10-05
CP-40754 Update host.https_only field in dbsync based on firewall state

jameshensmancitrix:
The firewall-port script returns true if port 80 is blocked … and false if it is closed, this is captured in set_https_only to update the DB based on the state of the network, not the requested setting should there be a failure

### DIFF
--- a/ocaml/xapi/dbsync_slave.ml
+++ b/ocaml/xapi/dbsync_slave.ml
@@ -122,8 +122,20 @@ let refresh_localhost_info ~__context info =
       ~key:Xapi_globs.host_no_local_storage ~value:"true"
   ) else
     Db.Host.remove_from_other_config ~__context ~self:host
-      ~key:Xapi_globs.host_no_local_storage
-
+      ~key:Xapi_globs.host_no_local_storage ;
+  let script_output =
+    Helpers.call_script !Xapi_globs.firewall_port_config_script ["check"; "80"]
+  in
+  try
+    let network_state = Scanf.sscanf script_output "Port 80 open: %B" Fun.id in
+    Db.Host.set_https_only ~__context ~self:host ~value:network_state
+  with _ ->
+    let message =
+      Printf.sprintf
+        "unexpected output from /etc/xapi.d/plugins/firewall-port: %s"
+        script_output
+    in
+    raise Api_errors.(Server_error (internal_error, [message]))
 (*************** update database tools ******************)
 
 (** Record host memory properties in database *)
--- a/scripts/plugins/firewall-port
+++ b/scripts/plugins/firewall-port
@@ -47,10 +47,19 @@ case "${OP}" in
             service iptables save
         fi
         ;;
+    check) 
+        if [[ -z `iptables -S $CHAIN | grep " $PORT "` ]]
+        then
+            echo "Port $PORT open: true"
+        else
+            echo "Port $PORT open: false"
+        fi
+        ;;
     *)
-        echo $"Usage: $0 {open|close} {port} {protocol}" 1>&2
+        echo $"Usage: $0 {open|close|check} {port} {protocol}" 1>&2
         exit 1
         ;;
 esac
+
 exit 0
 


ocaml/xapi/xapi_host.ml
```diff
@@ -2928,6 +2928,7 @@ let apply_updates ~__context ~self ~hash =
 
 let set_https_only ~__context ~self ~value =
   let state = match value with true -> "close" | false -> "open" in
-  ignore
-  @@ Helpers.call_script !Xapi_globs.firewall_port_config_script [state; "80"] ;
-  Db.Host.set_https_only ~__context ~self ~value
+  let network_state =
+    Helpers.call_script !Xapi_globs.firewall_port_config_script [state; "80"]
+  in
+  Db.Host.set_https_only ~__context ~self ~value:(bool_of_string network_state)
```
robhoes, line 2934: This change should not be done here, but in `refresh_localhost_info` in dbsync_slave.ml, which is a function that runs when xapi start up. At that point we want to query the state of the firewall and sync the DB.

scripts/plugins/firewall-port
```diff
@@ -52,5 +52,13 @@ case "${OP}" in
         exit 1
         ;;
 esac
+
+if [[ -z `iptables -S $CHAIN | grep " $PORT "` ]]
```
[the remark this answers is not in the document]
jameshensmancitrix: Should I put the query under each of the case statements?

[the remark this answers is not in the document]
robhoes: No, only under one new case. So the script will have three distinct ops: open, close, check.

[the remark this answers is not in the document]
jameshensmancitrix: That makes more sense now, have since adjusted it

scripts/plugins/firewall-port
```diff
@@ -47,10 +47,19 @@ case "${OP}" in
             service iptables save
         fi
         ;;
+    check) 
+        if [[ -z `iptables -S $CHAIN | grep " $PORT "` ]]
+            then
+                echo true
+                else
```
lindig, line 54: Indentation needs a fix. Maybe this is part of the design, but I think a more descriptive output would be better unless the output is directly used somewhere else.

jameshensmancitrix: the stdout feeds into OCaml where it is fed into bool_of_string. This tells Xapi what the state of https_only should be.

lindig: I would still use a more descriptive output and use `sscanf` to parse it on the OCaml side. That is safer than using `bool_of_string` anyway. Something like: `port 80: open` which can be easily parsed.

ocaml/xapi/dbsync_slave.ml
```diff
@@ -122,7 +122,16 @@ let refresh_localhost_info ~__context info =
...
+      ~key:Xapi_globs.host_no_local_storage ;
+  let network_state =
+    Helpers.call_script !Xapi_globs.firewall_port_config_script ["check"; "80"]
+  in
+  Db.Host.set_https_only ~__context ~self:host
+    ~value:
+      (Scanf.bscanf
+         (Scanf.Scanning.from_string network_state)
+         "Port 80 open: %B" Fun.id
+      )
```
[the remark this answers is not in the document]
psafont: ksscanf takes an additional parameter that is a handler in case the string cannot be parsed:

ocaml/xapi/dbsync_slave.ml
```diff
@@ -122,8 +124,25 @@ let refresh_localhost_info ~__context info =
...
+      Api_errors.(
+        Server_error
+          ( internal_error
+          , [
+              Printf.sprintf
+                "unexpected output from /etc/xapi.d/plugins/firewall-port: %s"
+                script_output
+            ]
+          )
+      )
```
psafont, line 145: One way to avoid the indentation and extra lines caused by the formatter is:
```suggestion
    let message =
      Printf.sprintf
        "unexpected output from /etc/xapi.d/plugins/firewall-port: %s"
        script_output
    in
    raise Api_errors.(Server_error (internal_error, [message]))
```

jameshensmancitrix: That looks a lot more efficient

lindig (approved):
Nice improvements!

robhoes (approved):

robhoes:
This looks good now. But please merge only after testing.

jameshensmancitrix:
Am currently testing, I will compare with both network states switched when Xapi is not running

jameshensmancitrix:
Testing complete, xapi correctly sets https_only on the host on the basis of the port after systemctl restart xapi

PR #4813, opened 2022-10-07
xapi-guard: Add unit tests

psafont:
These were developed in conjunction with a deprivileged interface to gather vtpm data, but it was dropped: https://github.com/xapi-project/xen-api/pull/4670

 I've removed that part of the tests and kept the rest, including an file descriptor leak test that found actually found one. The fix was included in the codebase months ago.


psafont:
The FD leak test is noisy, there's probably some interference with dune and opam

psafont:
I'm unable to reproduce the failure locally, neither with dune nor without it:
```
$ dune runtest ocaml/xapi-guard/test/ -f --profile=release --no-buffer -j 12
```
```
$ _build/default/ocaml/xapi-guard/test/xapi_guard_test.exe -v
```

ocaml/xapi-guard/lib/varstored_interface.ml
```diff
@@ -147,29 +143,69 @@ let () =
...
+      let* () = Lwt_io.close oc in
+      Lwt_io.close ic
+  | Error e ->
+      D.debug "Waiting for socket to be connectable at %s: %s" path
+        (Printexc.to_string e) ;
+      (* just in case Lwt_unix.listen doesn't get called soon enough,
+         avoid using up 100% CPU waiting for it,
+         better than Lwt.pause ()
+      *)
+      let* () = Lwt_unix.sleep 0.001 in
```
lindig, line 173: The wait is a lot shorter now. Is this just for Lwt scheduling?

lindig (approved):
Nice to see code more concise and using the `let*`syntax.

ocaml/xapi-guard/lib/varstored_interface.ml
psafont: This is for getting the unit tests to pass quickly while still yielding the CPU to other tasks, see https://github.com/xapi-project/xen-api/pull/4670#discussion_r844046467

jameshensmancitrix (approved):

PR #4815, opened 2022-10-11
idl/json_backend: generate data for new-style xapi-releases

psafont:
Staarting from xapi release 1.250.0, xapi has relied more and more in its own version, rather than on xenserver releases and "API version".
This has meant that in the public docs there is no information of the API changes that have happened in the last 2 years.

This adds the information about these changes to the metadata that gets published. There are some tweaks that could be made, like adding the date of the release for each of these, or tweak the release datatype to better accomodate this new style, or even move all that information to the metadata generation side instead of being a fundamental part of xapi.

I've taken the liberty to port the json generation here to yojson. I've manually checked using jq that the json produced is equivalent, although the new one has a lot lets whitespace.

Example output using this branch:
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,6 @@ python:
 	$(MAKE) -C scripts/examples/python build
 
 doc-json:
-	dune build --profile=$(PROFILE) ocaml/idl/json_backend/gen_json.exe
 	dune exec --profile=$(PROFILE) -- ocaml/idl/json_backend/gen_json.exe -destdir $(XAPIDOC)/jekyll
 
 format:
--- a/ocaml/idl/json_backend/dune
+++ b/ocaml/idl/json_backend/dune
@@ -6,6 +6,7 @@
     xapi-consts
     xapi-stdext-unix
     xapi-stdext-std
+    yojson
   )
 )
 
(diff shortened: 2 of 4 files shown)


ocaml/idl/datamodel_types.ml
```diff
@@ -107,6 +107,16 @@ type api_release = {
   ; release_date: string option
 }
 
+let pp_api_release ppf
+    {code_name; version_major; version_minor; branding; release_date} =
+  Format.fprintf ppf
+    "{code_name=%a; version_major=%d; version_minor=%d; branding=%s; \
```
lindig, line 113: I lack the context but this does not look like JSON. If it was meant to be, it would be better to create an association and let a JSON library create the string representation. Fine otherwise.

lindig (approved):

Makefile
```diff
@@ -78,7 +78,6 @@ python:
 	$(MAKE) -C scripts/examples/python build
 
 doc-json:
-	dune build --profile=$(PROFILE) ocaml/idl/json_backend/gen_json.exe
```
robhoes, line 81: Will `dune exec` below always rebuild this?

ocaml/idl/datamodel_types.ml
robhoes: Yes, please remove if it is not used in the code.

Makefile
psafont: It will rebuild as needed, as it was previously

robhoes (approved):

PR #4820, opened 2022-10-14
CA-364194: Allow creation of statefiles to time out 

psafont:
Currently enabling HA can stall because the GFS2 calls when adding the
HA statefile might stall indefinitely when there are connection issues

Now xapi waits a minute before failing, signalling the user that
something has gone awry

Unfortunately using Python 2.7 blocks us from having a timeout in static-vdis, added a comment so it has more chances of getting added when porting to python 3.5+

Xapi helpers now will also print when a script is called without a callback so it's easier to spot similar situations

I've also converted this infrastructure to [mtime](https://github.com/xapi-project/xen-api/compare/master...psafont:xen-api:private/paus/its-about-mtime?expand=1), but it got big and it's better that this goes in independently

### DIFF
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -62,7 +62,7 @@ let filter_args args =
     )
     args
 
-let call_script ?(log_output = Always) ?env ?stdin script args =
+let call_script ?(log_output = Always) ?env ?stdin ?timeout script args =
   let should_log_output_on_success, should_log_output_on_failure =
     match log_output with
     | Always ->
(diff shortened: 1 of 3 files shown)


ocaml/xapi/helpers.ml
```diff
@@ -72,7 +72,10 @@ let call_script ?(log_output = Always) ?env ?stdin script args =
     | On_failure ->
         (false, true)
   in
-  debug "about to call script: %s %s" script
+  let timeout_msg =
+    if Option.is_none timeout then " without a timeout" else ""
+  in
+  debug "about to call script%s: %s %s" timeout_msg script
```
lindig, line 78: Include the timeout duration in the log message?

lindig (approved):

ocaml/xapi/helpers.ml
psafont: Yeah, makes sense, this way we can try looking for the failure of the script when triaging with the logs. I've added this

jameshensmancitrix (approved):

PR #4823, opened 2022-10-17
idl/json_backend: Process unreleased versioned releases

psafont:
These releases happen when an API change has been introduced and gen_lifecycle is executed before tagging. They are in the form of 22.11.0-next.
For ordering the ints are used like before, and the string is only used to compare if the integers are equal.


ocaml/idl/json_backend/gen_json.ml
```diff
@@ -541,31 +541,71 @@ let write_to_dir dir maybe_file =
...
+     - 22.11.0-next
+     - 22.12.0
+     From the way the lifecycle binary generates versions there only will be a
+     single -next version, with the highest version number
+  *)
+  let compare (xs, x_next) (ys, y_next) =
+    let rec compare_list x y =
+      match (x, y) with
+      | x :: xs, y :: ys when x = y ->
+          compare_list xs ys
```
edwintorok, line 559: 4.12 has List.compare: https://v2.ocaml.org/api/List.html so this could be written more simply

ocaml/idl/json_backend/gen_json.ml
```diff
@@ -541,31 +541,71 @@ let write_to_dir dir maybe_file =
...
+      | [], y :: _ when y = 0 ->
+          0
+      | [], _ :: _ ->
+          -1
+      | [], [] ->
+          0
+    in
+    let cmp = compare_list xs ys in
+    if cmp = 0 then
+      match (x_next, y_next) with
```
edwintorok, line 575: Option.compare exists since 4.08 it could shorten this code

ocaml/idl/json_backend/gen_json.ml
```diff
@@ -541,31 +541,46 @@ let write_to_dir dir maybe_file =
...
+  *)
+  let compare (xs, x_next) (ys, y_next) =
+    let cmp = List.compare Int.compare xs ys in
+    if cmp = 0 then
+      Option.compare String.compare x_next y_next
+    else
+      cmp
+
+  let of_name name =
+    let to_version v = String.split_on_char '.' v |> List.map int_of_string in
```
lindig, line 563: This could be also done with `Astring.String.fields`

```
utop # let sep = function '.' | '-' -> true | _ -> false;;
val sep : char -> bool = <fun>
─( 15:51:55 )─< command 2 >──────────────────────────────────────────{ counter: 0 }─
utop # Astring.String.fields ~is_sep:sep "12.13.14-foo";;
- : string list = ["12"; "13"; "14"; "foo"]
```

lindig (approved):

ocaml/idl/json_backend/gen_json.ml
lindig:
```
utop # Scanf.sscanf "12.23.45" "%d.%d.%d%s" (fun x y z l -> (x,y,z,l));;
- : int * int * int * string = (12, 23, 45, "")

utop # Scanf.sscanf "12.23.45-foo" "%d.%d.%d%s" (fun x y z l -> (x,y,z,l));;
- : int * int * int * string = (12, 23, 45, "-foo")
```

psafont: the leading dash is unimportant right, now, I've used the scanf concoction

minglumlu (approved):

PR #4825, opened 2022-10-18
CP-39594: Add datamodel for Rpm_gpg_key

minglumlu:
This series adds a few functions to manage RPM GPG public keys in dom0.

The 'pool.install_rpmgpgkey' is used to add a new RPM GPG pubkey in a
pool and sync the key to every member host in the pool.
The 'pool.uninstall_rpmgpgkey' is to uninstall a managed RPM GPG pubkey.
The 'pool.sync_rpmgpgkeys' is to sync all managed RPM GPG pubkey
throughout all the hosts in the pool.

With these functions, XAPI users can add a new RPM GPG public key and
then configure it to a repository. In updating of a pool, the RPM GPG
key will be used to verify the metadata of the repository and the RPM
packages in the repository. What's more, the users can uninstall an
existing RPM GPG pubkey and install a new one if the existing on is to
be replaced.


ocaml/xapi/message_forwarding.ml
```diff
@@ -629,6 +629,14 @@ functor
           Ref.string_of repository
       with _ -> "invalid"
 
+    let rpm_gpg_key_uuid ~__context rpm_gpg_key =
+      try
+        if Pool_role.is_master () then
+          Db.Rpm_gpg_key.get_uuid ~__context ~self:rpm_gpg_key
+        else
+          Ref.string_of rpm_gpg_key
+      with _ -> "invalid"
```
lindig, line 638: Is this a good way to handle this? Should this not raise an exception rather than returning some malformed string?

ocaml/xapi/rpm_gpg_key.ml
```diff
@@ -0,0 +1,246 @@
...
+
+let parse_pubkey_metadata md_str_in_colons =
+  let open Angstrom in
+  let s = char ':' in
+  let p = take_till (fun x -> x = ':') in
+  let colons = sep_by s p >>= fun ss -> return ss in
+  let consume = Consume.Prefix in
+  let r = (parse_string ~consume colons) md_str_in_colons in
+  match r with
+  | Ok l ->
```
lindig, line 53: Would it not be easier to match to use Angstrom combinators like `count` to consume the expected number of `:` and let the parser fails rather than doing this manually here? I would also try to let the parser return a list and match the structure of this list here and avoid using something like `List.nth`

lindig:
Architecturally, how many of these GPG keys do we have and how large are they? Would it have been possible to store them in the xapi database and get distribution for free?

ocaml/xapi/rpm_gpg_key.ml
minglumlu: Thanks. Yeah, using `count` looks better:
```
  let single_field =
    take_while (fun x -> x <> ':') >>= fun v ->
    char ':' >>= fun _ ->
    return v
  in
  let pickup =
    single_field >>= fun _ ->
    count 4 single_field >>= fun _ ->
    single_field >>= fun created ->
    count 13 single_field >>= fun _ ->
    single_field >>= fun fingerprint  ->
    return (created, fingerprint)
  in
```

ocaml/idl/datamodel_host.ml
```diff
@@ -1744,6 +1744,34 @@ let set_https_only =
       ]
     ~allowed_roles:_R_POOL_OP ()
 
+let install_rpmgpgkey =
+  call ~name:"install_rpmgpgkey"
+    ~doc:"Install a RPM GPG public key to the host."
```
lindig, line 1749: "Install .. onto .." or "Add .. to .."

ocaml/idl/datamodel_host.ml
```diff
@@ -1744,6 +1744,34 @@ let set_https_only =
...
+      ; (String, "name", "The name of the RPM GPG public key")
+      ; (String, "pubkey", "The RPM GPG public key")
+      ; (String, "fingerprint", "The fingerprint of the RPM GPG public key")
+      ]
+    ~allowed_roles:(_R_POOL_OP ++ _R_CLIENT_CERT)
+    ~lifecycle:[] ()
+
+let uninstall_rpmgpgkey =
+  call ~name:"uninstall_rpmgpgkey"
+    ~doc:"Uninstall a RPM GPG public key on the pool."
```
lindig, line 1762: I think it should be "Uninstall an RPM" - but not a native speaker. While "r" is a consonant, "rpm" is pronounced like starting with a vowel and hence "an" would be correct.

ocaml/idl/datamodel_pool.ml
```diff
@@ -44,6 +44,18 @@ let operations =
         , "Indicates the primary host is sending its certificates to another \
            host"
         )
+      ; ( "install_rpmgpgkey"
+        , "Indicates this pool is in the process of installing a RPM GPG \
```
lindig, line 48: "an RPM" - but see above; also below

ocaml/xapi/rpm_gpg_key.ml
```diff
@@ -0,0 +1,246 @@
...
+  match (parse_string ~consume pickup) md_str_in_colons with
+  | Ok (date_str, fingerprint) ->
+      let created =
+        try Float.of_string date_str
+        with _ ->
+          error "Failed to parse RPM GPG key: invalid creation date: %s"
+            date_str ;
+          raise Api_errors.(Server_error (rpm_gpg_key_is_invalid, []))
+      in
+      let safe_char = function 'A' .. 'Z' | '0' .. '9' -> true | _ -> false in
```
lindig, line 68: This logic could be part of the `pickup` function by using a combinator that accepts only safe characters. But fine to do it here, too.

lindig (approved):

psafont (changes requested):
I don't like that the datatype is so specialised when most of it doesn't need to be, it limits future applicability.

- Please use a gpg key datatype, add a variant to indicate what is the gpg going to be used for (in this case for rpms), similar to what we do with x509 certificates: https://github.com/xapi-project/xen-api/blob/master/ocaml/idl/datamodel_certificate.ml#L21
- Split the functionality of rpm_gpg_key.ml into xapi_gpg_key.ml and gpg.ml, with the API datatype and database going into the first and the gpg processing into the second

minglumlu:
Hello @psafont, @lindig and @robhoes,
Sorry for long delay on this PR. I've refined the code based on the comments. Could you please take a look again? Thanks in advance!

ocaml/xapi/gpg.ml
```diff
@@ -152,3 +152,91 @@ let with_detached_signature filename signature size f =
...
+
+let assert_name_is_valid ~name =
+  let safe_char = function
+    | 'A' .. 'Z' | 'a' .. 'z' | '0' .. '9' | '.' | '_' | '-' ->
+        true
+    | _ ->
+        false
+  in
+  let len = String.length name in
+  let starts_with_dot = Astring.String.is_prefix ~affix:"." name in
```
psafont, line 164:
```suggestion
  let starts_with_dot = String.starts_with ~prefix:"." name in
```

ocaml/xapi/gpg.ml
```diff
@@ -152,3 +152,91 @@ let with_detached_signature filename signature size f =
...
+      raise Api_errors.(Server_error (gpg_key_name_is_invalid, [name]))
+  | true, true, _ ->
+      raise Api_errors.(Server_error (gpg_key_name_is_empty, []))
+  | true, false, true ->
+      raise Api_errors.(Server_error (gpg_key_name_is_too_long, [name]))
+  | true, false, false ->
+      ()
+
+module PubKeyMetaData = struct
+  type t = {created: float; fingerprint: string}
```
psafont, line 177: Could you change the created field to be Ptime.t? This makes it obvious it's a unit timestamp. This needs an awkward conversion from ptime to float to date to store it in the database, I'm fixing this is https://github.com/xapi-project/stdext/pull/68/

ocaml/idl/datamodel_gpg_key.ml
```diff
@@ -0,0 +1,47 @@
...
+          ~default_value:(Some (VString "")) "name"
+          "The name of the GPG key file"
+      ; field ~qualifier:StaticRO ~lifecycle ~ty:DateTime
+          ~default_value:(Some (VDateTime Date.never)) "created"
+          "Date on which the key was created"
+      ; field ~qualifier:StaticRO ~lifecycle ~ty:String
+          ~default_value:(Some (VString "")) "fingerprint"
+          "The fingerprint of the GPG public key"
+      ; field ~qualifier:DynamicRO ~lifecycle ~ty:Bool
+          ~default_value:(Some (VBool false)) "uninstalled"
```
psafont, line 41: What's the lifecycle of the value? one a gpg key has been installed it can never leave the database?

ocaml/xapi/rpm_gpg_key.ml
```diff
@@ -0,0 +1,167 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+module Unixext = Xapi_stdext_unix.Unixext
+
+module D = Debug.Make (struct let name = "rpm_gpg_key" end)
```
psafont, line 17:
```suggestion
module D = Debug.Make (struct let name = __MODULE__ end)
```

ocaml/idl/datamodel_host.ml
```diff
@@ -1744,6 +1744,34 @@ let set_https_only =
...
+    ~doc:"Install an RPM GPG public key onto the host."
+    ~params:
+      [
+        (Ref _host, "self", "The host")
+      ; (String, "name", "The name of the RPM GPG public key")
+      ; (String, "pubkey", "The RPM GPG public key")
+      ; (String, "fingerprint", "The fingerprint of the RPM GPG public key")
+      ]
+    ~allowed_roles:(_R_POOL_OP ++ _R_CLIENT_CERT)
+    ~lifecycle:[] ()
```
psafont, line 1758: shouldn't this one be pool_internal, like uninstall?

ocaml/xapi/xapi_pool.ml
```diff
@@ -1563,6 +1563,42 @@ let join_common ~__context ~master_address ~master_username ~master_password
             "Unable to set the write the new pool certificates to the disk : %s"
             (ExnHelper.string_of_exn e)
       ) ;
+      (* Sync RPM GPG keys to every host of the current pool *)
+      ((* Retrieve keys from the new pool *)
+       let pubkeys =
```
psafont, line 1568: This won't be enough to ensure all hosts have all the keys, we used to do this for pool certificates and it didn't work for all cases.

Consider the case where multiple host are joining a pool at the same time. The first certificate to join will miss the gpg keys of successive joiners before it's able to restart and connect to the pool coordinator as part of the pool.

On pull start supporters needs to import the missing keys from the pool coordinator to be able to solve this race condition

See https://github.com/xapi-project/xen-api/commit/c6ae148d024a6158d3922e0c6fafcddcdc137073 to see how it was solved for pool certificates

ocaml/xapi/xapi_pool.ml
```diff
@@ -3605,3 +3656,223 @@ let set_https_only ~__context ~self:_ ~value =
...
+                   )
+               )
+         | _ ->
+             None
+     )
+  |> fun gpg_key_refs ->
+    (* Install the key on the coordinator host firstly. *)
+    Rpm_gpg_key.install_rpmgpgkey ~name ~pubkey
+      ~fingerprint:md.PubKeyMetaData.fingerprint ;
+    match gpg_key_refs with
```
psafont, line 3699: I believe the key should be installed in the filesystem only if the database record is missing, otherwise we're writing to the filesystem needlessly

ocaml/xapi/xapi_pool.ml
```diff
@@ -3605,3 +3656,223 @@ let set_https_only ~__context ~self:_ ~value =
...
+        ref
+    | _ ->
+        raise Api_errors.(Server_error (multiple_rpm_gpg_keys_exist, [name]))
+  )
+  |> fun gpg_key_ref ->
+  (* Install the key on other member hosts. *)
+  let other_members =
+    let coordinator = Helpers.get_localhost ~__context in
+    Db.Host.get_all ~__context |> List.filter (fun href -> href <> coordinator)
+  in
```
psafont, line 3713:
```suggestion
  let other_members = Xapi_pool_helpers.get_slaves_list ~__context in
```

ocaml/idl/datamodel_pool.ml
```diff
@@ -1023,6 +1035,52 @@ let set_uefi_certificates =
...
+    ~result:(Ref _gpg_key, "The reference of the created Gpg_key object")
+    ()
+
+let uninstall_rpmgpgkey =
+  call ~name:"uninstall_rpmgpgkey"
+    ~doc:"Uninstall an RPM GPG public key from the pool."
+    ~params:
+      [
+        (Ref _pool, "self", "The pool")
+      ; (String, "name", "The name of the RPM GPG public key to be uninstalled")
```
psafont, line 1058: This argument makes the implementation more difficult than it should be, use `(Ref _gpg_key, "gpg_key", "...")` instead

ocaml/idl/datamodel_pool.ml
```diff
@@ -1023,6 +1035,52 @@ let set_uefi_certificates =
...
+    ~allowed_roles:(_R_POOL_OP ++ _R_CLIENT_CERT)
+    ~lifecycle:[] ()
+
+let sync_rpmgpgkeys =
+  call ~name:"sync_rpmgpgkeys"
+    ~doc:"Sync all RPM GPG public keys throughout the pool."
+    ~params:
+      [
+        (Ref _pool, "self", "The pool")
+      ; (Set (Ref _host), "hosts", "The hosts to be synced")
```
psafont, line 1069: What's the usecase for only updating only some of the hosts in the pool? I'd remove this parameter to make the call easier to use

ocaml/idl/datamodel_pool.ml
```diff
@@ -1023,6 +1035,52 @@ let set_uefi_certificates =
...
+      ]
+    ~allowed_roles:(_R_POOL_OP ++ _R_CLIENT_CERT)
+    ~lifecycle:[] ()
+
+let get_rpm_pubkey_string =
```
psafont, line 1074: "string" doesn't have semantic value. Rather use "contents", and the doc state which format are the contents in

psafont (changes requested):

ocaml/xapi/xapi_pool.ml
```diff
@@ -3605,3 +3656,223 @@ let set_https_only ~__context ~self:_ ~value =
...
+                  )
+                  hosts
+              in
+              Helpers.run_in_parallel ~funs ~capacity:16 |> ignore ;
+              Db.Gpg_key.destroy ~__context ~self
+            )
+            refs
+    )
+
+let sync_rpmgpgkeys ~__context ~self:_ ~hosts =
```
psafont, line 3785: A lot of work in this functions seems like it should be part of uninstall, see previous comment

psafont, line 3785: I'm unsure about having such a complex function with such a specific usecase, at this point I'm missing a design.

We have complex machinery around certificate ditribution, it's probably much more complex that what's needed here but it's more principled that this function, and I'd like to see what guarantees are needed here exactly to be able to judge the implementation

ocaml/idl/datamodel_gpg_key.ml
minglumlu: The DB record of the key will be removed only after it has been removed on all member hosts. So this value will be set to `true` if this key is going to be removed. But some member hosts may be offline at that time. There is another chance to clean up all since the key in DB has been marked as `uninstalled`.

ocaml/idl/datamodel_host.ml
minglumlu: This will be used by client to determine if a key (may with same name) has been changed.

ocaml/xapi/xapi_pool.ml
minglumlu: I removed this function completely.

ocaml/idl/datamodel_pool.ml
minglumlu: I removed the `pool.sync_rpmgpgkeys` completely.

ocaml/xapi/xapi_pool.ml
minglumlu: I think this would correct the case that the key file on coordinator is missing.

ocaml/xapi/xapi_pool.ml
```diff
@@ -3605,3 +3656,223 @@ let set_https_only ~__context ~self:_ ~value =
...
+      | [] ->
+          raise
+            Api_errors.(
+              Server_error (rpm_gpg_key_record_does_not_exist, [name])
+            )
+      | refs ->
+          let hosts = Db.Host.get_all ~__context in
+          List.iter
+            (fun self ->
+              Db.Gpg_key.set_uninstalled ~__context ~self ~value:true ;
```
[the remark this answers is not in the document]
minglumlu: Normally 'pool.uninstall_rpmgpgkey' will remove both DB record and files on hosts in the pool. But if some of member hosts failed to be contacted in the call, the DB record would be left there but with `uninstalled=true`, and meanwhile there will be an error like 'host offline'.
User can re-call 'pool.uninstall_rpmgpgkey' until all hosts having reported success later.

ocaml/xapi/gpg.ml
minglumlu: Sure. I will change it once the PR got merged.

ocaml/xapi/xapi_pool.ml
minglumlu: Unlike certificates in which case there are certs exchanging between hosts, the GPG keys (at least RPM GPG public keys) would be synced from coordinator to other members in uni-direction.
But indeed, there may be cases that a key was not synced to a member host. So I appended a commit for the member host to pull the key from coordinator on-demand.

ocaml/idl/datamodel_host.ml
psafont: I'd rather not override an install call to have 2 functionality, why can't a client check if a key has changed by querying the API for all the gpg_keys and checking their names and fingerprints?

ocaml/xapi/xapi_pool.ml
psafont: > Unlike certificates in which case there are certs exchanging between hosts, the GPG keys (at least RPM GPG public keys) would be synced from coordinator to other members in uni-direction.

On pool join the joiner downloads the certificates from the coordinator, and it pushes its certificates to the coordinator, which will push it to the pool supporters. The race condition is the same as here

quality-gate.sh
```diff
@@ -25,7 +25,7 @@ verify-cert () {
 }
 
 mli-files () {
-  N=513
+  N=514
```
psafont, line 28: add a empty `ocaml/tests/test_gpg_key.mli` instead ;)

minglumlu: This should be caused by the `datamodel_gpg_key.ml` since the script doesn't count the files under "tests".

psafont: ahh, yes, nevermind, then

ocaml/xapi/xapi_pool.ml
minglumlu: Discussed offline. The joiner should remove the managed GPG keys in old pool before joining.

ocaml/idl/datamodel_host.ml
minglumlu: Ah. I see. I had a mistake in my mind.

ocaml/xapi/xapi_pool.ml
minglumlu: Appended [113c8ea](https://github.com/xapi-project/xen-api/pull/4825/commits/113c8ea648a6d895d7be373cf4d17316e33a0571) for this.

edwintorok (changes requested):

ocaml/idl/datamodel_pool.ml
```diff
@@ -1023,6 +1031,45 @@ let set_uefi_certificates =
...
+let install_rpmgpgkey =
+  call ~name:"install_rpmgpgkey"
+    ~doc:"Install an RPM GPG public key onto the pool."
+    ~params:
+      [
+        (Ref _pool, "self", "The pool")
+      ; (String, "name", "The name of the RPM GPG public key")
+      ; (String, "pubkey", "The RPM GPG public key")
+      ]
+    ~allowed_roles:(_R_POOL_OP ++ _R_CLIENT_CERT)
```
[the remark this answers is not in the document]
minglumlu: Close this PR due this issue.

PR #4828, opened 2022-10-21
CA-369446 

jameshensmancitrix:
Change the use of doexec in add_vswitch_port to run using the ON_ERROR_FAIL handler. This will prevent the VM from starting in the case of ovs-vsctl failing.

### DIFF
--- a/ocaml/xenopsd/scripts/setup-vif-rules
+++ b/ocaml/xenopsd/scripts/setup-vif-rules
@@ -23,7 +23,7 @@ import sys
 import syslog
 import time
 import common
-from common import send_to_syslog, doexec
+from common import ON_ERROR_FAIL, send_to_syslog, doexec, run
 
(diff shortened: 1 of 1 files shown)


lindig (approved):

lindig:
I think the commit subject could be shorter (to avoid overflow) and be more descriptive: "exit with error if add_vswitch_port fails" or similar. The body could explain how this is implemented (but it is also quite obvious) once you see the diff

robhoes:
Did you manage to test this?

jameshensmancitrix:
@robhoes What would need to be done to have ovs-vsctl fail on vm start?

robhoes:
Firstly, you need to check that the normal case still works by starting a VM. To test the failure case, you could modify the code to call a fake version of ovs-vsctl that exits with an error.

robhoes (approved):
Testing has now happened. All good.

PR #4832, opened 2022-10-24
CP-40823 Corrected blank vm in DP.destroy and VDI.destroy functions

jameshensmancitrix:


ocaml/tests/test_storage_migrate_state.ml
```diff
@@ -32,7 +32,7 @@ let sample_send_state =
           ; vdi= Storage_interface.Vdi.of_string "mirror_vdi"
           ; url= "remote_url"
           }
-    ; local_dp= "local_dp"
+    ; local_dp= "remote_dp"
```
jameshensmancitrix, line 35: By the looks of it there are a few pieces of clean up i should do in this instance

ocaml/xapi-idl/storage/storage_interface.ml
```diff
@@ -482,7 +482,7 @@ module StorageAPI (R : RPC) = struct
 
   let vm_p = Param.mk ~name:"vm" Vm.t
 
-  let dp_p = Param.mk ~name:"dp" dp
+  let dp_p = Param.mk ~name:"dp" Types.string
```
jameshensmancitrix, line 485: change back to dp as this is now just `Types.string`

ocaml/xapi-idl/storage/storage_interface.ml
```diff
@@ -522,7 +522,7 @@ module StorageAPI (R : RPC) = struct
       let id_p = Param.mk ~name:"id" Types.string in
       declare "DP.create"
         ["[DP.create dbg id]: creates and returns a dp"]
-        (dbg_p @-> id_p @-> returning dp_p err)
+        (dbg_p @-> id_p @-> returning id_p err)
```
jameshensmancitrix, line 525: revert this

ocaml/xapi-idl/storage/storage_interface.ml
```diff
@@ -1100,9 +1100,10 @@ module type Server_impl = sig
   end
 
   module DP : sig
-    val create : context -> dbg:debug_info -> id:string -> dp
+    val create : context -> dbg:debug_info -> id:string -> string
```
jameshensmancitrix, line 1103: revert

ocaml/xapi/storage_impl.ml
```diff
@@ -183,9 +185,19 @@ module Vdi = struct
...
+    if dp_on_vdi_by_id dp t then
+      List.assoc dp (List.map (fun ((dp, _vm), st) -> (dp, st)) t.dps)
+    (*perhaps clean up *)
+    else
+      Vdi_automaton.Detached
+
+  let get_dp_state (dp : Dp.t) t =
     if dp_on_vdi dp t then
-      List.assoc dp t.dps
+      List.assoc dp t.dps (*perhaps clean up *)
```
jameshensmancitrix, line 200: remove comment, function did include a convoluted mapping but is not unneeded

ocaml/xapi/storage_migrate.ml
```diff
@@ -494,9 +494,10 @@ let copy' ~task ~dbg ~sr ~vdi ~url ~dest ~dest_vdi =
       None
   in
   try
-    let remote_dp = Uuidx.(to_string (make ())) in
-    let base_dp = Uuidx.(to_string (make ())) in
-    let leaf_dp = Uuidx.(to_string (make ())) in
+    (* FIXME: SMAPIv3 storage migration not implemented, what value to use for domid here? *)
+    let remote_dp = Local.DP.create dbg Uuidx.(to_string (make ())) in
+    let base_dp = Local.DP.create dbg Uuidx.(to_string (make ())) in
+    let leaf_dp = Local.DP.create dbg Uuidx.(to_string (make ())) in
```
jameshensmancitrix, line 500: could be reverted, DP.create is not really needed here

ocaml/xapi/storage_migrate.ml
```diff
@@ -551,7 +552,8 @@ let copy' ~task ~dbg ~sr ~vdi ~url ~dest ~dest_vdi =
         )
       )
       (fun () ->
-        Remote.DP.destroy dbg remote_dp false ;
+        Remote.DP.destroy dbg remote_dp false
+          (Storage_interface.Vm.of_string "0") ;
```
jameshensmancitrix, line 556: Review this

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -24,9 +24,13 @@ let with_lock = Xapi_stdext_threads.Threadext.Mutex.execute
 
 let finally = Xapi_stdext_pervasives.Pervasiveext.finally
 
-module DD = Debug.Make (struct let name = "xapi_vm_migrate" end)
+(* module DD = Debug.Make (struct let name = "xapi_vm_migrate" end)
```
jameshensmancitrix, line 27: remove

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -24,9 +24,13 @@ let with_lock = Xapi_stdext_threads.Threadext.Mutex.execute
...
 
-open DD
+   open DD *)
```
jameshensmancitrix, line 29: remove commented section

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -888,14 +892,16 @@ let vdi_copy_fun __context dbg vdi_map remote is_intra_pool remote_vdis so_far
...
   let with_new_dp cont =
-    let dp =
-      Printf.sprintf
-        (if vconf.do_mirror then "mirror_%s" else "copy_%s")
-        vconf.dp
+    let dp, vm =
+      ( Printf.sprintf
+          (if vconf.do_mirror then "mirror_%s" else "copy_%s")
+          vconf.dp
+      , Vm.of_string "-1" (*Correct bad vm*)
```
jameshensmancitrix, line 899: correct this vm

ocaml/xenopsd/xc/xenops_server_xen.ml
```diff
@@ -3974,6 +3987,7 @@ module VBD = struct
         safe_rm xs (vdi_path_of_device ~xs device) ;
         safe_rm xs (Device_common.backend_path_of_device ~xs device ^ "/sm-data") ;
         Storage.dp_destroy task
+          (Storage_interface.Vm.of_string vm)
```
jameshensmancitrix, line 3990: double check if this good

ocaml/xapi/storage_impl.ml
```diff
@@ -183,7 +185,16 @@ module Vdi = struct
 
   let dp_on_vdi dp t = List.mem_assoc dp t.dps
 
-  let get_dp_state dp t =
+  let dp_on_vdi_by_id dp t =
+    List.mem_assoc dp (List.map (fun ((dp, _vm), st) -> (dp, st)) t.dps)
```
lindig, line 189: `List.mem_assoc` uses physical quality rather than structural quality - is this the right choice? `assoc` can also fail if the element is not present; is this guaranteed not to happen? Otherwise `List.assoc_opt` is a better choice.

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -24,9 +24,9 @@ let with_lock = Xapi_stdext_threads.Threadext.Mutex.execute
 
 let finally = Xapi_stdext_pervasives.Pervasiveext.finally
 
-module DD = Debug.Make (struct let name = "xapi_vm_migrate" end)
+module Impl = Debug.Make (struct let name = "storage_impl" end)
```
lindig, line 27: Is this the right name? By opening the module below, its name becomes irrelevant. Is this not logging errors for `xapi_vm_migrate`? In that case `struct let name = __MODULE__ end` would work. The name `Impl` doesn't make sense to me. This is about debug logging; why is `Impl` better than `DD` from before?

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -887,15 +887,26 @@ let vdi_copy_fun __context dbg vdi_map remote is_intra_pool remote_vdis so_far
     else
       (not is_intra_pool) || dest_sr <> vconf.sr
   in
+  let vbds = Db.VDI.get_VBDs ~__context ~self:vconf.vdi in
+  let _vm =
+    match vbds with
+    | [vbd] ->
+        Db.VBD.get_VM ~__context ~self:vbd
+    | _ ->
+        Ref.of_string ""
```
lindig, line 896: This looks like an error case - should this be logged? The result value looks like it is going to confuse the caller. It might be better to use error handling.

ocaml/xapi/xapi_vm_migrate.ml
jameshensmancitrix: This seems to have been auto generated, and the build did not like having the `module DD ...` included. I am not sure on this.

ocaml/xapi/xapi_vm_migrate.ml
jameshensmancitrix: That makes sense, I am not sure if this will ever happen or the importance of the vm paramter in this contxt.

ocaml/xapi/storage_impl.ml
jameshensmancitrix: In this case, I was modifying an existing function so that there was a function uisng the new dp structure and the old string

lindig: `mem_assoc` looks wrong to me - we should discuss this - also if it exists in existing code.

ocaml/xapi/xapi_vm_migrate.ml
robhoes: This is not right... please undo the change.

ocaml/xapi/storage_impl.ml
robhoes: `List.mem_assoc` does use structural equality; it is `List.mem_assq` that uses physical.

robhoes:
>CP-40823 Addedparamter ~fail_on_leaned in Storage_impl.ml remove_data…
>…paths_andthen_nolockk to fail if there are attached dps in VDI.destory

🤨 

Apart from the obvious mistakes, git works better if the first line of the commit message is a short title, then an empty line followed by more text where useful.

ocaml/xapi/storage_impl.ml
```diff
@@ -570,12 +582,12 @@ functor
         in
         let failures =
           List.fold_left
-            (fun acc dp ->
-              info "Attempting to destroy datapath dp:%s sr:%s vdi:%s" dp
+            (fun acc ((dpid, vm) as _dp) ->
```
robhoes, line 585: You don't need `as _dp` here. It is giving a value a new name that is not used.

ocaml/xapi/storage_migrate.ml
```diff
@@ -26,6 +26,8 @@ open Storage_interface
 open Storage_task
 open Storage_utils
 
+let unkown_domid = "0"
```
robhoes, line 29: `unknown_domid`

ocaml/xapi/xapi_globs.ml
```diff
@@ -976,6 +976,8 @@ let conn_limit_unix = ref 1024
 
 let conn_limit_clientcert = ref 800
 
+let fail_on_leaked = false
```
robhoes, line 979: To be able to set this this from the config file, you need to add it below.

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -887,15 +887,26 @@ let vdi_copy_fun __context dbg vdi_map remote is_intra_pool remote_vdis so_far
     else
       (not is_intra_pool) || dest_sr <> vconf.sr
   in
+  let vbds = Db.VDI.get_VBDs ~__context ~self:vconf.vdi in
+  let _vm =
```
robhoes, line 891: This code is inside `vdi_copy_fun`, which is only called from `migrate_send'`, which knows the `vm`. It would be easier to just add a `vm` param to `vdi_copy_fun`.

robhoes (changes requested):

ocaml/xapi/storage_access.ml
```diff
@@ -1718,7 +1718,6 @@ let on_vdi ~__context ~vbd ~domid f =
   let module C = Storage_interface.StorageAPI (Idl.Exn.GenClient (struct
     let rpc = rpc
   end)) in
-  let dp = C.DP.create dbg dp in
```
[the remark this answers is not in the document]
jameshensmancitrix: We already have the DP at this point

ocaml/xenopsd/dbgring/dune
```diff
@@ -6,7 +6,6 @@
    dune-build-info
    xapi-xenopsd
    xenctrl
-   xenctrl.xenmmap
```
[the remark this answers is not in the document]
jameshensmancitrix: I did not make that change, I think it may have gotten mixed up when I accidentally messed up my own repo

ocaml/xapi/storage_impl.ml
```diff
@@ -573,8 +574,11 @@ functor
               info "Attempting to destroy datapath dp:%s sr:%s vdi:%s" dp
                 (s_of_sr sr) (s_of_vdi vdi) ;
               try
-                destroy_datapath_nolock context ~dbg ~dp ~sr ~vdi ~vm
-                  ~allow_leak:false ;
+                if (not fail_on_leaked) || dps <> [] then
+                  destroy_datapath_nolock context ~dbg ~dp ~sr ~vdi ~vm
+                    ~allow_leak:false
+                else
+                  failwith "Leaked Datapaths not allowed" ;
```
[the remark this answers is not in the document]
jameshensmancitrix: I used `failwith` here so that the failure mechanism would pick it up, I have moved the check outside of the statement

ocaml/xapi/storage_impl.ml
```diff
@@ -568,26 +573,29 @@ functor
...
-                (s_of_sr sr) (s_of_vdi vdi) ;
-              try
-                destroy_datapath_nolock context ~dbg ~dp ~sr ~vdi ~vm
-                  ~allow_leak:false ;
-                acc
-              with e -> e :: acc
-            )
-            [] dps
+          if fail_on_leaked && dps <> [] then
+            [failwith "Leaked Datapaths not allowed"]
```
robhoes, line 577: This raises an exception immediately rather than putting an exception _value_ in a list. I'd really just pull it outside `let failures =`, e.g.:

```
if fail_on_leaked && dps <> [] then
  failwith "Leaked datapaths not allowed" ;
let failures =
  ...
```

ocaml/xenopsd/xc/xenops_server_xen.ml
```diff
@@ -1732,9 +1729,18 @@ module VM = struct
             Domain.destroy task ~xc ~xs ~qemu_domid ~vtpm:(vtpm_of ~vm)
               ~dm:(dm_of ~vm) domid ;
             (* Detach any remaining disks *)
+            let dps =
+              List.map
+                (fun device ->
+                  ( Device.Generic.get_private_key ~xs device _dp_id
+                  , Storage.vm_of_domid (Some domid)
```
robhoes, line 1736: This is still a constant value for all dps. There is no need to put it inside the list. Just leave dps as it was and pass `Storage.vm_of_domid (Some domid)` to `Storage.dp_destroy` below.

ocaml/xapi/storage_access.ml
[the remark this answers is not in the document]
robhoes: Apparently `DP.create` just returns the same `dp` without any side-effects. Perhaps there was a type conversion in the past, but it seems unnecessary indeed.

ocaml/xapi/storage_impl.ml
```diff
@@ -568,26 +573,29 @@ functor
...
+          else
+            List.fold_left
+              (fun acc (dpid, vm) ->
+                info "Attempting to destroy datapath dp:%s sr:%s vdi:%s" dpid
+                  (s_of_sr sr) (s_of_vdi vdi) ;
+                try
+                  destroy_datapath_nolock context ~dbg ~dp:dpid ~sr ~vdi ~vm
+                    ~allow_leak:false ;
+                  acc
+                with e -> Printexc.to_string e :: acc
```
robhoes, line 587: Now you are rewriting exceptions and always raise `Failure` rather than the original exception, which could've been a different one...

ocaml/xapi/storage_impl.ml
```diff
@@ -568,26 +573,29 @@ functor
...
+        match failures with [] -> next () | f :: _ -> failwith f
 
       let epoch_begin context ~dbg ~sr ~vdi ~vm ~persistent =
         info "VDI.epoch_begin dbg:%s sr:%s vdi:%s vm:%s persistent:%b" dbg
           (s_of_sr sr) (s_of_vdi vdi) (s_of_vm vm) persistent ;
         with_vdi sr vdi (fun () ->
-            remove_datapaths_andthen_nolock context ~dbg ~sr ~vdi ~vm Vdi.leaked
-              (fun () ->
+            remove_datapaths_andthen_nolock context ~dbg ~sr ~vdi
+              ~fail_on_leaked:false Vdi.leaked (fun () ->
```
robhoes, line 598: Probably want to refer to the Xapi_globs value rather than `false`.

ocaml/xapi/xapi_host.ml
```diff
@@ -1302,7 +1302,7 @@ let get_thread_diagnostics ~__context ~host:_ =
   Locking_helpers.Thread_state.to_graphviz ()
 
 let sm_dp_destroy ~__context ~host:_ ~dp ~allow_leak =
-  Storage_access.dp_destroy ~__context dp allow_leak
+  Storage_access.dp_destroy ~__context dp allow_leak ~domid:0
```
robhoes, line 1305: I didn't spot this earlier... but it can't be right I'm afraid.

ocaml/xapi/storage_impl.ml
```diff
@@ -113,9 +113,11 @@ let info (fmt : ('a, unit, string, unit) format4) =
 let host_state_path = ref "/var/run/nonpersistent/xapi/storage.db"
 
 module Dp = struct
-  type t = string [@@deriving rpcty]
+  type t = string * Vm.t [@@deriving rpcty]
 
-  let make username = username
+  let make ((_datapath, _vm) as dp) = dp
+
+  let to_string (dp, _) = dp
```
psafont, line 120: Types already let us know that a string is returned, stating the semantic meaning  seems more valuable:
```suggestion
  let to_path (dp, _) = dp
```

ocaml/xapi/storage_impl.ml
```diff
@@ -172,10 +172,11 @@ module Vdi = struct
     ; dps: (Dp.t * Vdi_automaton.state) list
           (** state of the VDI from each dp's PoV *)
     ; leaked: Dp.t list  (** "leaked" dps *)
+    ; dpv: (Dp.t * Vm.t) list
```
psafont, line 175: Some information in a doccoment would be welcome here
```suggestion
    ; dpv: (Dp.t * Vm.t) list  (** VM Datapath *)
```

robhoes: This needs a default added for the upgrade case: `[@default []]` at the end of the line.

ocaml/xapi/storage_impl.ml
```diff
@@ -189,11 +190,42 @@ module Vdi = struct
...
-  let set_dp_state dp state t =
-    let rest = List.filter (fun (u, _) -> u <> dp) t.dps in
+  (** [get_dp_vm] returns [vm] of type Vm.t. If the mapping between a datapath and vm exists, it will return that vm. otherwise it will attempt to claculate it. If that is not pisslbe it will return a blank vm *)
+
+  let get_dp_vm dp t =
+    if dp_on_vdi dp t then
+      let _vm = List.assoc dp t.dpv in
+      match s_of_vm _vm with
+      | "" ->
+          (* Blank vms will occur if host is being upgraded. these should not normally occur*)
```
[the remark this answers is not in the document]
robhoes: It's better to use `List.assoc_opt dp t.dpv`.

ocaml/xapi/storage_impl.ml
```diff
@@ -189,11 +190,42 @@ module Vdi = struct
...
+      | vm ->
+          vm_of_s vm
+    (* This else block should only be called if there is no dp in the dpv list. This should only be an issue during the upgrade*)
+    else
+      let vm =
+        debug
+          "[Vdi.get_dp_vm] No vm was found on the datapath: trying to obtain \
+           vm from parsing dp %s"
+          dp ;
+        try Scanf.sscanf dp "vbd/%8i/%_s" string_of_int
```
robhoes, line 215: I wouldn't bother with this. Many of the dp strings are not formatted this way, so it will only work some of the time, which doesn't buy us much.

ocaml/xapi/storage_impl.ml
robhoes: This is still here.

ocaml/xapi/storage_impl.ml
```diff
@@ -189,11 +190,42 @@ module Vdi = struct
...
+        with Scanf.Scan_failure _ -> ""
+      in
+      vm_of_s vm
+
+  let add_or_update_dp dp vm state t =
+    debug "[Vdi.add_or_update] dp:%s vm:%s" dp (s_of_vm vm) ;
+    let rests = List.filter (fun (u, _) -> u <> dp) t.dps in
```
robhoes, line 222: You can use `List.remove_assoc` here (added to the standard lib after the original code here was written).

ocaml/xapi-idl/lib_test/test_data/storage/requests/DP.destroy.request.0
```diff
@@ -1 +1 @@
-<?xml version="1.0"?><methodCall><methodName>DP.destroy</methodName><params><param><value><struct><member><name>allow_leak</name><value><boolean>1</boolean></value></member><member><name>dp</name><value>dp</value></member><member><name>dbg</name><value>dbg</value></member></struct></value></param></params></methodCall>
\ No newline at end of file
+<?xml version="1.0"?><methodCall><methodName>DP.destroy</methodName><params><param><value><struct><member><name>allow_leak</name><value><boolean>0</boolean></value></member><member><name>dp</name><value>dp</value></member><member><name>dbg</name><value>dbg</value></member></struct></value></param></params></methodCall>
```
robhoes, line 1: Is this useful?

robhoes (commented):
Almost there! This is getting smaller and better at the same time :)

There are still several `~fail_on_leaked:false`, which need to use the value from `Xapi_globs`, and a code comment need to be updated. Then we'll probably want to squash the commits into one or two.

robhoes:
There are still lots of `~fail_on_leaked:false` (see above). There two commits now, but the first one still introduces all the interface changes that the second patch removes. This is hard to disentangle now, so lets just squash them into a single patch.

jameshensmancitrix:
> There are still lots of `~fail_on_leaked:false` (see above). There two commits now, but the first one still introduces all the interface changes that the second patch removes. This is hard to disentangle now, so lets just squash them into a single patch.

The fail_on_leaked business is somewhat a separate issue, it may be best to do an entirely separate PR on this issue, I have removed this entirely and just kept the changes related to the dp, vm mapping.

ocaml/tests/test_storage_impl.ml
```diff
@@ -0,0 +1,128 @@
+open Storage_impl
```
robhoes, line 1: You need copyright headers on top of this new file.

ocaml/tests/test_storage_impl.ml
```diff
@@ -0,0 +1,120 @@
...
+
+let pp_state fmt state =
+  Format.fprintf fmt "%s" (Vdi_automaton.string_of_state state)
+
+let pp_vdi fmt {dps; leaked; dpv; _} =
+  let pp_vm f vm = Format.fprintf f "%s" (Storage_interface.Vm.string_of vm) in
+  let pp_state f state =
+    Format.fprintf f "%s" (Vdi_automaton.string_of_state state)
+  in
+  Format.fprintf fmt "{attach_info=_;dps=%a;leaked=%a;dpv=%a}"
```
psafont, line 31:
```suggestion
```suggestion
These are already defined

File: ocaml/tests/test_storage_impl.ml [hunk 35]
```diff
@@ -0,0 +1,120 @@
...
+let pp_vm fmt vm = Format.fprintf fmt "%s" (Storage_interface.Vm.string_of vm)
+
+let pp_state fmt state =
+  Format.fprintf fmt "%s" (Vdi_automaton.string_of_state state)
+
+let pp_vdi fmt {dps; leaked; dpv; _} =
+  let pp_vm f vm = Format.fprintf f "%s" (Storage_interface.Vm.string_of vm) in
+  let pp_state f state =
+    Format.fprintf f "%s" (Vdi_automaton.string_of_state state)
+  in
```
psafont, line 30:
```suggestion
```
These are already defined and not needed

ocaml/xapi/storage_impl.ml
```diff
@@ -209,13 +223,17 @@ module Vdi = struct
...
     {t' with leaked= dp :: t'.leaked}
 
-  let dps t = List.map fst t.dps
+  let dp t = List.map fst t.dps
 
-  (** [perform dp op t] updates VDI [t] given the request to perform [op] by [dp] *)
-  let perform dp op t =
+  let dpv t = t.dpv
+
+  (** [perform dp vm op t] updates VDI [t] given the request to perform [op] by [dp] and then updates the dp on the vdi *)
```
psafont, line 230: this comment should be on top of the `let perform` instead of here

robhoes (approved):

ocaml/xapi/storage_impl.ml
```diff
@@ -189,16 +186,33 @@ module Vdi = struct
     else
       Vdi_automaton.Detached
 
-  let set_dp_state dp state t =
-    let rest = List.filter (fun (u, _) -> u <> dp) t.dps in
+  (** [get_dp_vm] returns [vm] of type Vm.t. If the mapping between a datapath and vm exists, it will return the vm. Owtherwise a blank vm will be returned*)
+
```
psafont, line 190:
```suggestion
```

ocaml/xapi/storage_impl.ml
```diff
@@ -189,16 +186,33 @@ module Vdi = struct
...
+  let get_dp_vm dp t =
+    let _vm = List.assoc_opt dp t.dpv in
+    match _vm with
+    | Some vm ->
+        vm
+    | None ->
+        debug
+          "[Vdi.get_dp_vm] No vm was found on the datapath please restart host \
+           to fix" ;
```
psafont, line 199: This looks like it should be a `warn` instead of a debug.

Restarting the host might be a bit too much to ask to fix it, maybe restarting xapi is enough?

ocaml/xapi/storage_impl.ml
```diff
@@ -189,16 +186,33 @@ module Vdi = struct
...
+        vm_of_s ""
+
+  let add_or_update_dp dp vm state t =
+    debug "[Vdi.add_or_update] dp:%s vm:%s" dp (s_of_vm vm) ;
```
psafont, line 203: Can we avoid this debug logline? I'd like to reduce noise in the logs if possible

psafont (approved):

PR #4834, opened 2022-10-25
CA-371759: check certificates in xsh

robhoes:
This program is used to sync blobs across hosts in a pool. It should verify certificates, as for other TLS connections within a pool.


edwintorok (approved):

ocaml/xsh/xsh.ml
```diff
@@ -74,12 +78,22 @@ let proxy (ain : Unix.file_descr) (aout : Unix.file_descr) (bin : Unixfd.t)
     try Unix.close bout with _ -> ()
   )
 
+let init_tls_verification () =
+  let file = Constants.verify_certificates_path in
+  match Sys.file_exists file with
+  | false ->
+      warn "TLS verification is disabled on this host: %s" file ;
```
lindig, line 85: This could be misleading: it points to a file that does not exist. So maybe:

```
"TLS verification is disabled on this host: %s is absent"
```

lindig (approved):

ocaml/xsh/xsh.ml
robhoes: True... I copied this from xapi.ml, so will change it there too.

PR #4835, opened 2022-10-27
CA-352073, CA-371780,  CA-372128: reduce overhead in xapi database and rbac checks; and in xcp-rrdd

psafont:
Continuation of https://github.com/xapi-project/xen-api/pull/4829

RBAC checks have complicated list traversals, quite a few of them can be avoidable changing the datastructure to a list of tuples. There were also some database lookups that were unnecessary and have been folded together, there is also some refactoring in rbac_audit that tries to reduce the cognitive complexity of the code, hopefully it's easier to analyse.

Db_cacche_types spent an unreasonable amount of time in map.update. This is because it tried to avoid running update on unchanged values. This incurred a terrible amount of cost. Currently we can use the standard library implementation of map.update which avoids this while being efficient, cutting the cpu runtime of xapi under certain conditions by half.

Additionally I've also changes an n^2 operation in xcp-rrdd that was taking almost 20% of its cpu time, the change seems very effective, I cannot see it anymore in the flamegraphs and the rest of the functions take wider chunk :)

### DIFF
--- a/ocaml/xapi/rbac.ml
+++ b/ocaml/xapi/rbac.ml
@@ -69,9 +69,6 @@ let session_permissions_tbl = Hashtbl.create 256 (* initial 256 sessions *)
 
 module Permission_set = Set.Make (String)
 
-(* This flag enables efficient look-up of the permission set *)
-let use_efficient_permission_set = true
-
 let permission_set permission_list =
   List.fold_left
     (fun set r -> Permission_set.add r set)
@@ -79,7 +76,7 @@ let permission_set permission_list =
 
 let create_session_permissions_tbl ~session_id ~rbac_permissions =
   if
-    use_efficient_permission_set && Pool_role.is_master ()
+    Pool_role.is_master ()
     (* Create this structure on the master only, *)
     (* so as to avoid heap-leaking on the slaves *)
   then (
@@ -92,8 +89,7 @@ let create_session_permissions_tbl ~session_id ~rbac_permissions =
     None
 
 let destroy_session_permissions_tbl ~session_id =
-  if use_efficient_permission_set then
-    Hashtbl.remove session_permissions_tbl session_id
+  Hashtbl.remove session_permissions_tbl session_id
 
 (* create a key permission name that can be in the session *)
 let get_key_permission_name permission key_name =
@@ -152,21 +148,17 @@ let permission_of_action ?args ~keys _action =
 let is_permission_in_session ~session_id ~permission ~session =
   let find_linear elem set = List.exists (fun e -> e = elem) set in
   let find_log elem set = Permission_set.mem elem set in
-  if use_efficient_permission_set then
-    (* use efficient log look-up of permissions *)
-    let permission_tree =
-      try Some (Hashtbl.find session_permissions_tbl session_id)
-      with Not_found ->
-        create_session_permissions_tbl ~session_id
-          ~rbac_permissions:session.API.session_rbac_permissions
-    in
-    match permission_tree with
-    | Some permission_tree ->
-        find_log permission permission_tree
-    | None ->
-        find_linear permission session.API.session_rbac_permissions
-  else (* use linear look-up of permissions *)
-    find_linear permission session.API.session_rbac_permissions
+  let permission_tree =
+    try Some (Hashtbl.find session_permissions_tbl session_id)
+    with Not_found ->
+      create_session_permissions_tbl ~session_id
+        ~rbac_permissions:session.API.session_rbac_permissions
+  in
+  match permission_tree with
+  | Some permission_tree ->
+      find_log permission permission_tree
+  | None ->
+      find_linear permission session.API.session_rbac_permissions
 
 open Db_actions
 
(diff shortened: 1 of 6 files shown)


edwintorok (approved):
Thanks, it looks like OCaml has a specialized `caml_string_compare` it can call, even if String.compare is implemented by just calling Stdlib.compare with known types.
Except in this case the compiler doesn't know the type yet for `let compare` so using string should provide a benefit.

This is how the generated assembly looks like when using plain Stdlib.compare vs String.compare:
[output from godbolt
](https://godbolt.org/z/nfW4KKeo6)

psafont:
Did a run with all the fixes:
- now Rbach_check appears in .22 % of the samples / 6.2 % are xapi samples -> 1.36 % of xapi's CPU time, which means it's not the bottleneck in this situation anymore

Running a BST + BVT suite: 176527

ocaml/xapi/rbac_audit.ml
```diff
@@ -350,85 +313,57 @@ let rec sexpr_args_of __context name rpc_value action =
...
-              match sexpr_args_of __context str_name rpc_value action with
-              | None ->
-                  params
-              | Some p ->
-                  p :: params
+              let myparam = sexpr_of_parameters __context action (Some d) in
+              myparam @ params
+          | _ -> (
+            (* the expected list of xml arguments *)
+            match sexpr_args_of __context str_name rpc_value action with
```
lindig, line 359:
```
sexpr_args_of __context str_name rpc_value action |> Option.fold ~none:params ~some:(fun p -> p :: params)
```

ocaml/xapi/rbac_audit.ml
```diff
@@ -350,85 +313,57 @@ let rec sexpr_args_of __context name rpc_value action =
...
-              match _name_of_ref_value with
-              | None ->
-                  get_db_namevalue __context name action ref_of_ref_value
-              | Some "" ->
-                  get_db_namevalue __context name action ref_of_ref_value
-              | Some a ->
-                  a
-            in
+      match Ref_index.lookup value with
+      | None ->
```
lindig, line 317: could use `when` and and an extra case to make it flatter.

ocaml/xapi/rbac_audit.ml
```diff
@@ -169,22 +138,20 @@ let populate_audit_record_with_obj_names_of_refs line =
...
-              line
+        let (args : SExpr.t) = List.hd (List.rev els) in
+        match List.partition (fun (e : SExpr.t) -> e <> args) els with
+        | prefix, [SExpr.Node arg_list] ->
+            (* paste together the prefix of original audit record *)
+            let the_sexpr =
+              SExpr.Node (prefix @ [SExpr.Node (get_obj_names_of_refs arg_list)])
+            in
+            let to_string () = SExpr.string_of in
+            Printf.sprintf "%s %a" before_sexpr_str to_string the_sexpr
```
lindig, line 152: If we do this a lot, `String.concat` would avoid the overhead of `sprintf`.

ocaml/xapi/rbac.ml
```diff
@@ -114,86 +110,55 @@ let permission_of_action ?args ~keys _action =
...
-                        (Rpc.to_string v) action ;
-                      get_keyERR_permission_name action "DENY_NOVALUE"
+          debug "DENYING access: no 'key' argument in the action %s" action ;
+          get_keyERR_permission_name action "DENY_NOKEY"
+      | Some (Rpc.String key_name_in_args) -> (
+        try
+          let key_name =
+            List.find
+              (fun key_name ->
+                if Astring.String.is_suffix ~affix:"*" key_name then
```
lindig, line 124: Would it be faster to do this with a regular expression because it could avoid repeated traversal?

lindig (approved):

ocaml/xapi/rbac_audit.ml
psafont: yeah, the code around here is smelly, checking whether value is a ref should be much cheaper than looking it up

ocaml/xapi/rbac.ml
psafont: There's a traversal on key_name (to look for '*'), and another in key_name_in_args. Not sure about the "repeated traversal"

ocaml/xapi/rbac_audit.ml
```diff
@@ -127,33 +108,21 @@ let get_sexpr_arg name name_of_ref uuid_of_ref ref_value : SExpr.t =
...
-            ref_value
-      | _ ->
-          obj_ref
-      (* do nothing if not a triplet *)
-    )
+                ("", "")
+            | Some {name_label= None; uuid; _} ->
+                ("", uuid)
+            | Some {name_label= Some name; uuid; _} ->
+                (name, uuid)
```
robhoes, line 120: You can't have a obj_name without a uuid (as the old code suggests)?

robhoes: Seems not indeed, looking at Ref_index. The old code unnecessarily turns the uuid into an option.

psafont: The issue is having a uuid without an object name

robhoes (approved):

PR #4836, opened 2022-10-27
CP-41033: install XenAPI for Python3

MarkSymsCtx:


MarkSymsCtx:
This will need a spec change like
```
commit 8492c3f5391a7f9f135d843bf3a304020e1dc35b
Author: Mark Syms <mark.syms@citrix.com>
Date:   Wed Oct 26 15:37:07 2022 +0100

    CP-41033: install XenAPI for Python 3

diff --git a/xapi.spec b/xapi.spec
index 08f8b8d..130ee28 100644
--- a/xapi.spec
+++ b/xapi.spec
@@ -46,6 +46,7 @@ BuildRequires: gmp-devel
 BuildRequires: libuuid-devel
 BuildRequires: make
 BuildRequires: python2-devel
+BuildRequires: python3-devel
 BuildRequires: xs-opam-repo >= 6.54.0-2
 BuildRequires: libnl3-devel
 BuildRequires: systemd-devel
@@ -85,6 +86,7 @@ Requires: m2crypto
 Requires: net-tools
 Requires: vmss
 Requires: python-six
+Requires: python3-six
 Requires: python-pyudev
 Requires: gmp
 Requires: xapi-storage-plugins
@@ -414,11 +416,13 @@ rm -rf %{buildroot}
 DESTDIR=$RPM_BUILD_ROOT %{__make} install
 
 SITEDIR=$(python -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")
+SITE3DIR=$(python3 -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")
 for f in XenAPI XenAPIPlugin inventory; do
     for e in py pyc pyo; do
         echo $SITEDIR/$f.$e
     done
 done > core-files
+echo "$SITE3DIR/*" >> core-files
 
 ln -s /var/lib/xcp $RPM_BUILD_ROOT/var/xapi
 mkdir $RPM_BUILD_ROOT/etc/xapi.conf.d
```

lindig (approved):

MarkSymsCtx:
Internal BST test executed - 176192 and testing with Python 3 against in progress migrations.

robhoes (approved):
LGTM

scripts/Makefile
```diff
@@ -167,6 +167,7 @@ install:
 	$(IDATA) examples/python/XenAPI/XenAPI.py $(DESTDIR)$(SITE_DIR)/
 	$(IDATA) examples/python/inventory.py $(DESTDIR)$(SITE_DIR)/
 	$(IDATA) examples/python/XenAPIPlugin.py $(DESTDIR)$(SITE3_DIR)/
+	sed -i 's/#!\/usr\/bin\/python/#!\/usr\/bin\/python3/' $(DESTDIR)$(SITE3_DIR)/XenAPIPlugin.py
```
edwintorok, line 170: It is a plugin, you're not meant to be executing it directly, why does it even have a `#!` line?

edwintorok (commented):
LGTM, minor comment on why we need the `#!` line at all, can we delete it and then use the exact same file for python2+3?

MarkSymsCtx:
> LGTM, minor comment on why we need the `#!` line at all, can we delete it and then use the exact same file for python2+3?

Possibly ¯\_(ツ)_/¯, down to the owners of the code, I was just making it install properly. If it never needs to be executed, and I think you're probably correct as it doesn't have a check for `__main__` then for sure we can and simplify things

PR #4837, opened 2022-10-28
CP-41067: Drop eliloader bootloader: obsolete and not needed on any s…

edwintorok:
…upported guest

Eliloader used to supply a patched kernel/initrd needed to boot EL4.x-6.x and Ubuntu 11.04 PV kernels.
All those guests are EOL, and guests nowadays include the required drivers to boot under Xen.

eliloader is being removed from the product:
last update was in 2019 when adding EL6.10 support, and CentOS 6.x is EOL now. Drop it from the supported bootloaders list.

Pygrub is still there, even if booting PV guests isn't strictly supported right now, since it might be useful for PVH guests again, although PVH can also use OVMF to boot.
(Pygrub would extract the kernel to boot from the guest disk prior to booting it)

Also drop function arguments that are now unused (the vm argument).

### DIFF
--- a/ocaml/xenopsd/lib/bootloader.ml
+++ b/ocaml/xenopsd/lib/bootloader.ml
@@ -24,29 +24,15 @@ module D = Debug.Make (struct let name = "bootloader" end)
 
 open D
 
-type supported_bootloader = Pygrub | Eliloader
+type supported_bootloader = Pygrub
 
-let string_of_bootloader = function
-  | Pygrub ->
-      "pygrub"
-  | Eliloader ->
-      "eliloader"
+let string_of_bootloader = function Pygrub -> "pygrub"
 
-let bootloader_of_string = function
-  | "pygrub" ->
-      Some Pygrub
-  | "eliloader" ->
-      Some Eliloader
-  | _ ->
-      None
+let bootloader_of_string = function "pygrub" -> Some Pygrub | _ -> None
 
-let path_of_bootloader = function
-  | Pygrub ->
-      !Resources.pygrub
-  | Eliloader ->
-      !Resources.eliloader
+let path_of_bootloader = function Pygrub -> !Resources.pygrub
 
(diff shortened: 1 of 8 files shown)


edwintorok:
Still needs some testing internally

ocaml/xenopsd/lib/bootloader.ml
```diff
@@ -205,9 +187,7 @@ let sanity_check_path p =
       let canonical_path = Xapi_stdext_unix.Unixext.resolve_dot_and_dotdot p in
       match Filename.dirname canonical_path with
       | "/var/run/xen/pygrub"
-      (* From pygrub, including when called by eliloader *)
-      | "/var/run/xend/boot" (* From eliloader *) ->
-          canonical_path
```
psafont, line 210: Careful, now pygrub will raise an error without:
```suggestion
      (* From pygrub *)
      ->
          canonical_path
```

andyhhp:
`Pygrub is still there, even if booting PV guests isn't strictly supported right now`  Pygrub is also used for pv-in-pvh and that's still supported in some cases

ocaml/xenopsd/lib/bootloader.ml
edwintorok: We've dropped `xend` from the product long ago, I think that was a toolstack before `xl`, there are 2 lines there with pygrub, I only dropped one.

edwintorok: Ah I see what you mean

edwintorok: Fixed in latest push.

edwintorok:
I'll need to make one fixup here: if we detect eliloader then give a better error message than 'malicious guest, unknown bootloader', i.e. have a known_bootloader in addition to supported_bootloaders.

edwintorok:
Lets try this again in 2024. For now this code seems to be needed to boot RHEL6 under PVshim, and that is the only guest we can test with PVshim currently (everything else seems to be EOL).

PR #4841, opened 2022-11-03
cp-40190 Steps towards preventing SWTPM from filling dom0 root partition

xennifer:
Steps are made so the main swtpm software dose not need to create any files on the filesytem.
This incluse:
* pre-creating UDS  and passing the FD to swtpm.
* Switching to linear file for state storage.
* pre-creating the PID file.

Note that the switching to linear file is to be controlled by xenopsd, using a parameter.  Prevously this was incorrectly set as
'file' type when it should be 'dir' type.  This has been changed.  When xenopd is ready to use linear state file, it should pass the 'file' type.  Until this is done, swtpm is allowed to create files, and is hence, still a small risk.

### DIFF
--- a/ocaml/xenopsd/scripts/swtpm-wrapper
+++ b/ocaml/xenopsd/scripts/swtpm-wrapper
@@ -15,20 +15,22 @@
 from __future__ import print_function
 import os
 import stat
+import socket
 import sys
 import pwd
 import grp
 import subprocess
 import ctypes
 import ctypes.util
+import struct
 from resource import getrlimit, RLIMIT_CORE, RLIMIT_FSIZE, setrlimit
-
-
-STATE_FILE = 'tpm2-00.permall'
+from urlparse import urlparse
 
 CLONE_NEWNS  = 0x00020000 # mount namespace
 CLONE_NEWNET = 0x40000000 # network namespace
 CLONE_NEWIPC = 0x08000000 # IPC namespace
+
+SWTPM_NVSTORE_LINEAR_MAGIC = 0x737774706d6c696e
 
 # Set cgroup_slice to the name of the cgroup slice the swtpm process
 # should live in.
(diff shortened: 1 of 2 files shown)


ocaml/xenopsd/scripts/swtpm-wrapper
```diff
@@ -85,38 +83,84 @@ def prepare_exec():
...
     while n < len(argv):
-        if argv[n] == "-priv":
+        if argv[n] == "--priv":
             depriv = False
-            continue
+        else:
+            sys.exit("Unknown option %s\n" % argv[n])
         n += 1
 
+    uri_sep = tpm_state.find("://")
```
psafont, line 134: [urlparse](https://docs.python.org/2/library/urlparse.html) is in the standard library and should be used here:
```python
urlparse('dir://tpm2-00.permall')
ParseResult(scheme='dir', netloc='tpm2-00.permall', path='', params='', query='', fragment='')
```

ocaml/xenopsd/scripts/swtpm-wrapper
```diff
@@ -85,38 +83,84 @@ def prepare_exec():
...
+    tpm_state_prefix = tpm_state[:uri_sep]
+    tpm_state_file = tpm_state[uri_sep+3:].lstrip('/')
+
+    if (tpm_state_prefix == "dir"):
+        tpm_uri = tpm_dir
+        tpm_file = os.path.join(tpm_dir, tpm_state_file)
```
psafont, line 140: tpm_file has the same definiton in both branches, should be defined before to avoid repetition

ocaml/xenopsd/scripts/swtpm-wrapper
```diff
@@ -85,38 +83,84 @@ def prepare_exec():
...
+    if not os.path.exists(fname):
+        return True
+
+    mode = os.stat(fname).st_mode
+
+    if not stat.S_ISBLK(mode):
+        return False
+
+    # Check if block device has non-zero header
+    file = open(fname, "r")
```
lindig, line 105: I like `with open as file` as you will never miss to close a file.

ocaml/xenopsd/scripts/swtpm-wrapper
```diff
@@ -85,38 +83,84 @@ def prepare_exec():
...
+        sys.exit("Not enough arguments.")
 
     domid = int(argv[1])
     tpm_dir = argv[2]
+    tpm_state = argv[3]
     tpm_path = tpm_dir
     depriv = True
 
-    n= 3
+    n = 4
```
lindig, line 126: Put this higher and use `if len(argv) < n` above.

lindig (approved):

ocaml/xenopsd/scripts/swtpm-wrapper
```diff
@@ -132,21 +176,32 @@ def main(argv):
...
+            os.chown(tpm_file, uid, uid)
 
         except OSError as error:
             print(error)
             return
 
         tpm_path = '/'
+        tpm_uri = tpm_state
+
+        if (tpm_state_prefix == "file"):
```
rosslagerwall, line 188: Should this be in the `if depriv` section?

ocaml/xenopsd/scripts/swtpm-wrapper
```diff
@@ -132,21 +176,32 @@ def main(argv):
             os.chown(tpm_dir, uid, uid)
```
rosslagerwall, line 176: I don't think this is right. If the tpm_dir is owned by the same uid as the swtpm process, then it can change the permissions back to 07xx. If the tpm_dir is owned by root, that can't happen since only the owner can change permissions.

ocaml/xenopsd/scripts/swtpm-wrapper
xennifer: It all about restricting priveladge, so yes?

ocaml/xenopsd/scripts/swtpm-wrapper
```diff
@@ -85,38 +84,80 @@ def prepare_exec():
...
+    mode = os.stat(fname).st_mode
+
+    if not stat.S_ISBLK(mode):
+        return False
+
+    # Check if block device has non-zero header
+    with open(fname, "r") as file:
+        hdr = file.read(8)
+
+    if len(hdr) == 8 and hdr != "\0\0\0\0\0\0\0\0":
```
rosslagerwall, line 109: It would be better to see if there is a magic number you can use here instead.

The source code defines `SWTPM_NVSTORE_LINEAR_MAGIC` - does that work?

xennifer: I based this code on what it does to 'delete' the tpm.  It always zeros it.
I only wanted the equivelant of 'if the file exists' which we had before.
I suppose the magic number would work too, I didnt try.

ocaml/xenopsd/scripts/swtpm-wrapper
```diff
@@ -129,24 +170,36 @@ def main(argv):
...
-            os.chown(tpm_dir, uid, uid)
             if os.path.exists(os.path.join(tpm_dir, ".lock")):
                 os.chown(os.path.join(tpm_dir, ".lock"), uid, uid)
-            os.chown(os.path.join(tpm_dir, STATE_FILE), uid, uid)
+            os.chown(tpm_file, uid, uid)
+
+            if (tpm_state_scheme != "file"):
+                os.chown(tpm_dir, uid, uid)
+            else:
+                os.chown(tpm_dir, 0 , uid)
```
rosslagerwall, line 180: Just noting here that this works because xenopsd creates tpm_dir as 0750. After swtpm switches uid/gid, the process gid matches the dir gid and allows swtpm to read files in tpm_dir but not create new ones. It can't change the directory permissions either because that is reserved for the directory owner.

We need to ensure that the xenospd side doesn't ever change to create tpm_dir with more lax permissions at some point in the future.

rosslagerwall: Maybe it should chmod `0750` here as well to ensure that?

psafont (approved):

PR #4842, opened 2022-11-03
Add certificate checking to sparse-dd

robhoes:
This is required to be able to fully support storage migration over HTTPS. This is still off by default.


psafont (approved):
Neat, nothing of concern that I saw

PR #4843, opened 2022-11-03
CA-371780: Reduce overheads in update_rrdds

psafont:
This function had quadratic costs with all the lists it was using. On
top of that all three types of owned datasources were processed in a
different way, especially the VMs, which where map from a list of active
VMs instead of the owned data sources

Further work is needed move collect the datasources into the appropriate
datastructure instead of in this function, and extend the datastructures
into the used functions isntead of converting them to lists

Additionally I've ported the unit tests to alcotests, which allowed to uncover a minor difference in behaviour: previously the host record was always created, not it only is created when there are host sources. This is irrelevant in the grand scheme of things as host sources are fed every time to the function.

I've manually tested updating a host with this change, the overhead from this function has gone down substantially looking at the flamegraphs, and CHC keeps showing stats correctly, both new one and previous ones.

I don't think this change should be backported

Marked as draft as a couple of todos need to be handled

image

### DIFF
--- a/ocaml/xcp-rrdd/bin/rrdd/rrdd_monitor.ml
+++ b/ocaml/xcp-rrdd/bin/rrdd/rrdd_monitor.ml
@@ -64,6 +64,29 @@ let merge_new_dss rrd dss =
     )
     rrd new_dss
 
+module OwnerMap = Map.Make (struct
+  type t = ds_owner
+
+  let compare a b =
+    match (a, b) with
+    | Host, Host ->
+        0
+    | Host, _ | VM _, SR _ ->
+        -1
+    | _, Host | SR _, VM _ ->
+        1
+    | VM a, VM b | SR a, SR b ->
+        String.compare a b
+end)
+
+let owner_to_string () = function
+  | Host ->
+      "host"
+  | VM uuid ->
+      "VM " ^ uuid
+  | SR uuid ->
+      "SR " ^ uuid
+
 (** Updates all of the hosts rrds. We are passed a list of uuids that is used as
     the primary source for which VMs are resident on us. When a new uuid turns
     up that we haven't got an RRD for in our hashtbl, we create a new one. When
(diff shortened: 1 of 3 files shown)


ocaml/xcp-rrdd/bin/rrdd/rrdd_monitor.ml
```diff
@@ -62,14 +62,49 @@ let merge_new_dss rrd dss =
     )
     rrd new_dss
 
+module OwnerMap = Map.Make (struct
+  type t = ds_owner
+
+  let compare a b =
```
lindig, line 68: Is there anything special about this order? Would the built-in `compare` not yield a similar result? In that case `let compare = compare` with a comment would be sufficient.

ocaml/xcp-rrdd/bin/rrdd/rrdd_monitor.ml
```diff
@@ -167,32 +178,32 @@ let update_rrds timestamp dss (uuid_domids : (string * int) list) paused_vms =
...
-          dss
+      let do_host dss =
+        let named_updates =
+          StringMap.to_seq dss |> Seq.map to_named_updates |> List.of_seq
+        in
+        let dss = StringMap.to_seq dss |> Seq.map snd |> List.of_seq in
+
+        match !host_rrd with
+        | None ->
+            debug "Creating fresh RRD for localhost" ;
```
lindig, line 189: Add some `__FUNCTION__` so we know where this is coming from?

ocaml/xcp-rrdd/bin/rrdd/rrdd_monitor.ml
```diff
@@ -167,32 +178,32 @@ let update_rrds timestamp dss (uuid_domids : (string * int) list) paused_vms =
...
-            (List.map
-               (fun ds ->
-                 (ds.ds_name, (ds.ds_value, ds.ds_pdp_transform_function))
-               )
-               host_dss
-            )
+      let process_dss ds_owner dss =
+        match ds_owner with
+        | Host ->
+            do_host dss
```
lindig, line 202: Prefix `process_` instead of `do_` would fit the theme.

lindig (approved):

ocaml/xcp-rrdd/bin/rrdd/rrdd_monitor.ml
psafont: I'm wary about using the polymorphic compare here due to its overhead compared to a hand-made function

ocaml/xcp-rrdd/bin/rrdd/rrdd_monitor.ml
```diff
@@ -62,14 +62,58 @@ let merge_new_dss rrd dss =
...
+      | None ->
+          Some (add_ds_to StringMap.empty)
+      | Some dss ->
+          warn {|%s: found duplicate datasource with key "%s" in owner "%a"|}
+            __FUNCTION__ ds.ds_name owner_to_string owner ;
+          Some (add_ds_to dss)
+    in
+    OwnerMap.update owner merge all
+  in
+  let dss = List.to_seq dss |> Seq.fold_left consolidate OwnerMap.empty in
```
edwintorok, line 110: This could've been just a regular fold_left, avoids constructing a seq just to immediately fold on it (although if you intend to push the list -> seq to the caller might be worth keeping)

edwintorok (approved):
Nice improvement, and shows why most of the time the choice of list as a data structure in XAPI is wrong

PR #4845, opened 2022-11-14
CP-41028: Enforce certificate checking for intra-pool storage migration connections

robhoes:


ocaml/xapi/storage_migrate.ml
```diff
@@ -339,7 +318,10 @@ let vdi_info x =
       failwith "Runtime type error: expecting Vdi_info"
 
 module Local = StorageAPI (Idl.Exn.GenClient (struct
-  let rpc call = rpc ~srcstr:"smapiv2" ~dststr:"smapiv2" (local_url ()) call
+  let rpc call =
+    Storage_utils.rpc ~srcstr:"smapiv2" ~dststr:"smapiv2"
```
lindig, line 322: Should we have a module-local binding for the `smapiv2` string and use that everywhere? That would eliminate any possibility to spell it wrong.

ocaml/xapi/storage_migrate.ml
```diff
@@ -434,18 +416,19 @@ let progress_callback start len t y =
...
   signal (Storage_task.id_of_handle t)
 
-let copy' ~task ~dbg ~sr ~vdi ~url ~dest ~dest_vdi =
-  let remote_url = Http.Url.of_string url in
+let copy' ~task ~dbg ~sr ~vdi ~url ~dest ~dest_vdi ~verify_dest =
+  let remote_url = Storage_utils.connection_args_of_uri ~verify_dest url in
   let module Remote = StorageAPI (Idl.Exn.GenClient (struct
     let rpc =
-      rpc ~srcstr:"smapiv2" ~dststr:"dst_smapiv2" (storage_url remote_url)
+      Storage_utils.rpc ~srcstr:"smapiv2" ~dststr:"dst_smapiv2" remote_url
```
lindig, line 423: Is the `dst_` prefix correct?

lindig (approved):

ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -2408,7 +2408,7 @@ and perform_exn ?subtask ?result (op : operation) (t : Xenops_task.task_handle)
         in
         debug "%s compress memory: %b" __FUNCTION__ compress_memory ;
         let verify_cert =
-          if vmm.vmm_verify_cert then Stunnel_client.pool () else None
+          if vmm.vmm_verify_cert then Some Stunnel.pool else None
```
psafont: This is confusing, but it's exactly what the doctor ordered

robhoes: I agree...

psafont: maybe a comment explaining the situation would help?

ocaml/xapi/storage_migrate.ml
robhoes: This is only for logging, and not used very consistently. I think we should sort this out properly when we get down to do proper distributed tracing. This stuff may then become obsolete.

PR #4847, opened 2022-11-15
CA-369444: Ensure xenopsd still starts if VM state upgrade fails

robhoes:


ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -3777,8 +3777,14 @@ let upgrade_internal_state_of_running_vms () =
...
-      let vm_t = VM_DB.read_exn vm in
-      B.VM.get_internal_state [] [] vm_t |> B.VM.set_internal_state vm_t
+      try
+        let vm_t = VM_DB.read_exn vm in
+        B.VM.get_internal_state [] [] vm_t |> B.VM.set_internal_state vm_t
+      with
+      | Xenops_interface.Xenopsd_error (Does_not_exist _) ->
+          warn "Missing VM extra metadata for VM %s" vm
+      | _ ->
+          warn "VM metadata upgrade failed for VM %s" vm
```
psafont, line 3787: might be worth printing the error here, are there exception serializers in Xenops_interface?

robhoes: Yes, and registered with `Printexc.register_printer`, so I think that gets picked up by `Printexc.to_string`.

ocaml/xenopsd/lib/xenopsd.ml
```diff
@@ -418,8 +418,25 @@ let configure ?(specific_options = []) ?(specific_essential_paths = [])
     ~name:(Filename.basename Sys.argv.(0))
     ~version:Xapi_version.version ~doc ~options ~resources ()
 
+let log_raw_backtrace bt =
+  Option.iter
+    (fun slots ->
```
lindig, line 423: Maybe too cryptic, but I believe you can omit `(fun slots ->` and `) slots` as the inner `Array.iteri` will have the correct type.

lindig (approved):

psafont (approved):

PR #4848, opened 2022-11-15
CA-371419: Always log exceptions when responding with 500 Internal Error

robhoes:
By default, logs from the http server are supressed, because it would log far too much. However, it would be valuable to always log the exception that has led to a 500 response, because this error condition is often due to a bug.

Logging for these cases is done under a separate name, which is enabled by default.


psafont (commented):
How do the other logs get supressed?

ocaml/libs/http-svr/http_svr.ml
```diff
@@ -177,7 +179,9 @@ let response_request_header_fields_too_large s =
   in
   response_error_html s "431" "Request Header Fields Too Large" [] body
 
-let response_internal_error ?req ?extra s =
+let response_internal_error ?req ?extra exc s =
+  E.error "Responding with 500 Internal Error due to %s" (Printexc.to_string exc) ;
+  E.log_backtrace () ;
```
lindig, line 184: Can we be sure that we still have the correct backtrace at this point? We had discussion in the past about lost back traces - obviously we will have some backtrace here but is it the correct one? That's why I would like to capture the backtrace first in an exception handler.

lindig (approved):

ocaml/libs/http-svr/http_svr.ml
```diff
@@ -39,6 +39,8 @@ module D = Debug.Make (struct let name = "http" end)
 
 open D
 
+module E = Debug.Make (struct let name = "http_internal_errors" end)
```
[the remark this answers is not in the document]
robhoes: Here we explicitly want a different name.

ocaml/libs/http-svr/http_svr.ml
robhoes: I'm not sure, but this is a catch-all situation. I guess it's better than nothing at this point.

PR #4850, opened 2022-11-17
CA-369690: Prioritize loglines when backing up RRDs, remove loglines about stunnel

psafont:
By default xcp-rrdd does not log debug lines, elevate two messages so
it's easier to know when this operation starts happening and when it
fails.

Two loglines per call removed: one is not really useful since the first
logline already contains the most important information about the
stunnel configuration and the one about the cache is not important as
the code is known to work as intended.

The main logline about stunnel has been changed to make it obvious there
has been a change in the are about logging when looking at logs

### DIFF
--- a/ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
+++ b/ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
@@ -112,11 +112,11 @@ let backup_rrds (remote_address : string option) () : unit =
   let destination =
     match remote_address with
     | None ->
-        "to local disk"
+        "local disk"
     | Some address ->
-        Printf.sprintf "to host %s" address
+        Printf.sprintf "host %s" address
   in
-  debug "backing up rrds %s" destination ;
+  info "%s: trying to back up RRDs to %s" __FUNCTION__ destination ;
   let total_cycles = 5 in
   let cycles_tried = ref 0 in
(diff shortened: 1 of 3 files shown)


ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
```diff
@@ -112,11 +112,11 @@ let backup_rrds (remote_address : string option) () : unit =
...
     match remote_address with
     | None ->
-        "to local disk"
+        "local disk"
     | Some address ->
-        Printf.sprintf "to host %s" address
+        Printf.sprintf "host %s" address
   in
-  debug "backing up rrds %s" destination ;
+  info "%s: trying to back up RRDs to %s" __FUNCTION__ destination ;
```
robhoes, line 119: How often does this happen?

psafont: every hour during normal operation, once every 6 minutes when testing, plus any manual requests to back the RRDs, if any

robhoes: This includes saving to local disk? RRDs are just in memory for most of the hour?

robhoes (approved):

jameshensmancitrix (approved):

PR #4851, opened 2022-11-17
Eliminate unnecessarily scary log lines

robhoes:


robhoes:
Draft: still running tests.

lindig (approved):

psafont (approved):
The logic to watch event looks quite complex, but I cannot spot any obvious issue with the change.

(previously exceptions were used to interrupt a while loop, not a guard is used)

robhoes:
Test failure looks like something else, so this change is clear.

psafont:
shoutout to make @rwdrich feel vindicated :)

PR #4852, opened 2022-11-18
Enable HTTPS for storage migration on the source host

robhoes:
This is currently set on the destination, in host.migrate_receive. However, all recent changes to make HTTPS migration work were on the source host. Migration is allowed from older to newer software versions, so it is possible that the destination is HTTPS capable, while the source is not. This may result in the source receiving HTTPS URLs, which is cannot handle, breaking this upgrade case.

Instead, let the source decide whether to switch to HTTPS or not, depending on the config key (to be made the default later).

### DIFF
--- a/ocaml/xapi/xapi_vm_migrate.ml
+++ b/ocaml/xapi/xapi_vm_migrate.ml
@@ -94,8 +94,14 @@ let use_compression options src dst =
       !Xapi_globs.migration_compression
 
 let remote_of_dest ~__context dest =
-  let master_url = List.assoc _master dest in
-  let xenops_url = List.assoc _xenops dest in
+  let maybe_set_https url =
+    if !Xapi_globs.migration_https_only then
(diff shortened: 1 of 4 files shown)


ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -111,12 +117,9 @@ let remote_of_dest ~__context dest =
   in
   let sm_url =
     let url = List.assoc _sm dest in
-    if Helpers.this_is_my_address ~__context remote_ip then
-      match Http.Url.of_string url with
-      | Http h, d ->
-          Http.Url.to_string (Http {h with Http.Url.ssl= false}, d)
-      | _ ->
-          url
+    (* Never use HTTPS for local SM calls *)
```
lindig, line 120: I think it would be more readable by avoiding the `not` and switching the branches. It would also agree with the comment better. Minor point, though.

lindig (approved):

psafont (approved):

PR #4853, opened 2022-11-21
CP-40357: Parse all lifecycled in the datamodel, avoid loading removed fields into the database

psafont:
Now all the lifecycles in the datamodel are check by an automaton found in datamodel_types
This allows to enforce lifecycles to all fields and messages and xapi to
know it, especially when loading the database from a file, which will now avoid loading removed fields onto the in-memory database

Marking as draft because I want feedback on the lifecycle automation and the datamodel changes but I haven't tested the database changes


ocaml/idl/datamodel_types.ml
```diff
@@ -444,15 +444,123 @@ type release = {
...
+  (* lifecycle automaton, rules:
+     - Unreleased is the initial state
+     - Extended and Changed do not change the state
+     - Objects cannot be Changed nor Extended before being prototyped or after
+       being deprecated
+     - The rest of the changes are not idempotent, they cannot be applied twice
+       in a row
+     - Objects can only be removed when are prototyped or deprecated
+  *)
+  let automaton =
```
lindig, line 497: Could this be simplified by using wildcards? There are only a few legal transitions in each state - express those explicitly and match all remaining ones with a wildcard and make them illegal?

psafont:
Now the database ignores removed fields when loading, the second time there are no messages about ignored fields.

Now testing the json generation code

psafont:
With the docs fixed, this is ready to be reviewed. I recommend doing it commit by commit, and special attention should be given to `CP-40357: Patch invalid lifecycles` to make sure the changes there make sense, and the last commit where the datamodel changes force changes into the rest of xapi

ocaml/idl/datamodel.ml
```diff
@@ -8414,3 +8414,44 @@ let extra_permissions =
...
+
+type obj_states = {
+    obj_state: Lifecycle.state
+  ; msg_states: Lifecycle.state StringMap.t
+  ; fld_states: Lifecycle.state StringMap.t
+}
+
+let all_lifecycles =
+  let parse_with ~name lifecycle =
+    try Lifecycle.from lifecycle
```
edwintorok: Should this use a `result` type instead, or do you want to preserve the location that raised the invalid message in stacktraces (if it is useful then it should be preserved and using exceptions here is better)

ocaml/idl/json_backend/gen_json.ml
```diff
@@ -381,6 +381,32 @@ end = struct
     else
       cmp
 
+  let rec list_dedup cmp =
```
edwintorok, line 384: I think we have an implementation of this elsewhere already (somewhere in stdext)?
Although there is `List.sort_uniq` would that help, or you need more control over which one to keep?

ocaml/idl/datamodel.ml
psafont: This code is replaced when the lifecycle state is integrated into the datamodel, I don't think it's important enough to change it for the few commits it's used

robhoes (commented):
This mostly looks really good, and is a much needed update to control the evolution of the API. Before merging, we'll need to ensure that we test the upgrade case very well. We'll also need to assess the impact on the API bindings (SDK), as I suspect that the last commit goes a step too far in removing fields, where clients may need them to connect to older server versions.

robhoes (approved):

ocaml/idl/json_backend/gen_json.ml
```diff
@@ -381,6 +381,32 @@ end = struct
...
+    let rec loop acc = function
+      | [] ->
+          List.rev acc
+      | [item] ->
+          loop (item :: acc) []
+      | a :: b :: rest when cmp a b = 0 ->
+          loop (a :: acc) rest
+      | a :: b :: rest ->
```
lindig, line 392: This could be written as `a :: rest -> loop (a :: acc) rest`

ocaml/idl/json_backend/gen_json.ml
```diff
@@ -381,6 +381,32 @@ end = struct
...
+          loop (a :: acc) (b :: rest)
+    in
+    loop []
+
+  let remove_outdated changes =
+    (* When several lifecycles transitions for the same entity, keep the most
+       significative one and drop the rest *)
```
lindig, line 399: significant?

ocaml/idl/json_backend/gen_json.ml
```diff
@@ -381,6 +381,32 @@ end = struct
...
+    changes
+    |> List.sort (fun ((_, nam_a, _, _) as a) ((_, nam_b, _, _) as b) ->
+           let cmp = String.compare nam_a nam_b in
```
lindig, line 402: I am not sure I understand this logic: when the names are identical, call `compare_changes`, which again uses the complete structure. Is the purpose here to compare the two structures by element and controlling the order? I've used the code below in such a situation where I define an order by comparing elements in a particular order.

```
  let (<?>) a b =
    if a = 0 then b else a

  let compare a b =
        compare a.hour    b.hour
    <?> compare a.minutes b.minutes
    <?> compare a.seconds b.seconds
    <?> 0
```

lindig (approved):
Obviously a complex change whose logic I can't review comprehensively. Only a few remarks,

ocaml/idl/json_backend/gen_json.ml
psafont: It orders the list of entity changes in a way that exploits the ordering of the type of changes (removed last, prototyped first).

Once that is done it drops items with a name that's the same as the preceding one. This way only the last change per entity is kept. For example if in a  release a field got deprecated and removed, only the removed change is kept.

ocaml/idl/json_backend/gen_json.ml
psafont: I've used latest, which is better at expressing intention

PR #4854, opened 2022-11-22
CA-391658: toolstack restart when HA is enabled is not safe

edwintorok:
Try to reduce toolstack downtime on toolstack restart by stopping and starting all the services at once: then any timeouts will be handled in parallel instead of sequently (i.e. ~1m30s, after which systemd will SIGKILL, instead of ~32m).

Reject toolstack restart through the API (e.g. from XenCenter) when HA is enabled, requiring you to disable HA first with `xe host-emergency-ha-disable --force` as documented.

We should also tune the default HA xapi check timeouts though, the default 60s for XHA Is wrong (larger than 1m30s in toolstack restart), and even the 120s for SMAPIv3 is dangerously close.

Alternatively toolstack restart could do the emergency HA disable itself, but actions like that are probably best if the user is aware of them, at least when done through the UI. Perhaps the toolstack restart script should either check and reject if HA is enabled or emergency disable it itself to prevent the undesired host fencing.

There are probably more improvements we can do here, hence a draft, because HA doesn't actually detect a stuck XAPI (xapi responds to healthchecks, but you can't start/stop VMs due to a stuck SR for example)

### DIFF
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -766,6 +766,9 @@ let restart_agent ~__context ~host:_ =
 
 let shutdown_agent ~__context =
   debug "Host.restart_agent: Host agent will shutdown in 1s!!!!" ;
+  if Pool_role.is_master () &&
+    try Localdb.get Constants.ha_armed = "true" with _ -> false then
+    raise (Api_errors.Server_error (Api_errors.ha_is_enabled, [])) ;
   Xapi_fuse.light_fuse_and_dont_restart ~fuse_length:1. ()
 
 let disable ~__context ~host =
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -766,8 +766,10 @@ let restart_agent ~__context ~host:_ =
 
 let shutdown_agent ~__context =
   debug "Host.restart_agent: Host agent will shutdown in 1s!!!!" ;
-  if Pool_role.is_master () &&
-    try Localdb.get Constants.ha_armed = "true" with _ -> false then
+  if
+    Pool_role.is_master ()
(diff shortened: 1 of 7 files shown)


ocaml/xapi/xapi_host.ml
```diff
@@ -766,6 +766,9 @@ let restart_agent ~__context ~host:_ =
 
 let shutdown_agent ~__context =
   debug "Host.restart_agent: Host agent will shutdown in 1s!!!!" ;
+  if Pool_role.is_master () &&
+    try Localdb.get Constants.ha_armed = "true" with _ -> false then
+    raise (Api_errors.Server_error (Api_errors.ha_is_enabled, [])) ;
   Xapi_fuse.light_fuse_and_dont_restart ~fuse_length:1. ()
```
lindig, line 772: This should move above the debug line.

scripts/xe-toolstack-restart
```diff
@@ -49,15 +49,13 @@ for svc in $SERVICES ; do
 
 	if [ $? -eq 0 ] ; then
 		TO_RESTART="$svc $TO_RESTART"
-		systemctl stop $svc
 	fi
 done
+systemctl stop ${TO_RESTART}
 
 set -e
```
lindig, line 56: Not sure this makes a lot of sense.

lindig (approved):

ocaml/xapi/xapi_host.ml
psafont: yup, this should be addressed

lindig (approved):

edwintorok:
I don't remember anymore what needs to be done here, probably best to revisit HA again when I'm working on something HA related, probably a deep rabbithole, for now I'm closing this.

PR #4855, opened 2022-11-23
CA-372785 make with-vdi more robust

lindig:
The vbd-plug call in with-vdi may fail after a VBD was already created. The errexit option would cause the script to fail before the VBD would be destroyed. So make sure we destroy the VBD if vbd-plug fails.


MarkSymsCtx (approved):
LGTM

robhoes (approved):

PR #4857, opened 2022-11-24
CA-371790: Restrict the permissions on pool tokens

snwoods:
Depends on the changes in https://github.com/xapi-project/stdext/pull/69

### DIFF
--- a/ocaml/xapi-types/secretString.ml
+++ b/ocaml/xapi-types/secretString.ml
@@ -24,7 +24,7 @@ let with_cookie t request =
   let cookie = [(_pool_secret, t)] in
   {request with Http.Request.cookie}
 
-let write_to_file = Xapi_stdext_unix.Unixext.write_string_to_file
+let write_to_file = Xapi_stdext_unix.Unixext.write_string_to_file ~perms:0o600
 
 let write_to_channel c s = output_string c s
 


lindig:
```
File "ocaml/xapi/helpers.ml", line 1954, characters 4-30:
1954 |     SecretString.write_to_file ~perms:0o600 !Xapi_globs.pool_secret_path ps ;
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: This function has type string -> SecretString.t -> unit
make: *** [Makefile:16: build] Error 1
Error: Process completed with exit code 2.
```

robhoes:
We first need to get the stdext change reviewed/committed and added to xs-opam.

robhoes:
You also need a `make format`.

psafont (approved):
LGTM, please wait to merge until xs-opam includes the necessary code to make the code compile

psafont:
I don't like that users of the function can dictate the permissions, please undo those changes and instead change the implementation of the function to hardcode it:

```ocaml
let write_to_file = Xapi_stdext_unix.Unixext.write_string_to_file ~perms:0o600
```

psafont (changes requested):
I don't like that users of the function can dictate the permissions, please undo those changes and instead change the implementation of the function to hardcode it:

```ocaml
let write_to_file = Xapi_stdext_unix.Unixext.write_string_to_file ~perms:0o600
```

lindig:
> I don't like that users of the function can dictate the permissions, please undo those changes and instead change the implementation of the function to hardcode it:
> 
> ```ocaml
> let write_to_file = Xapi_stdext_unix.Unixext.write_string_to_file ~perms:0o600
> ```

Is this based on the argument that we already know that we want to store a secret and hence need most restrictive permissions? Which `perms` don't you like? I think passing permission to `Xapi_stdext_unix.Unixext.write_string_to_file` is good, but to `SecretString.write_to_file` maybe not.

psafont:
> Is this based on the argument that we already know that we want to store a secret and hence need most restrictive permissions? Which perms don't you like?

Yes, since it's a secrets the permissions need to be more stringent. Currently it's on SecretString's user to use the safe permissions, this is risky since it's done a optional parameter which might be elided. Even if it wasn't I don't think it should be a decision at all as it should always use the safe choice, hence why it only makes sense to encode it in `SecretString.write_to_file` and not expose it to its users

lindig (approved):

psafont (approved):

PR #4858, opened 2022-11-25
Modify Xenctrlext to use its own libxc handle

andyhhp:
Xenopsd's _H() in xenctrlext_stubs blindly reaches sideways into the internals of a different libray (specifically, Xenctrl from upstream Xen).  This violates many API/ABI rules.  It also breaks now that upstream Xen has switched Xenctrl to using a Custom block.

Have Xenctrlext open its own libxc handle, to disentangle it from Xenctrl.

### DIFF
--- a/ocaml/xenopsd/xc/xenops_server_xen.ml
+++ b/ocaml/xenopsd/xc/xenops_server_xen.ml
@@ -1280,12 +1280,13 @@ module VM = struct
 
   let mkints n = List.init n Fun.id
 
-  let generate_create_info ~xc ~xs:_ vm persistent =
+  let generate_create_info ~xs:_ vm persistent =
     let ty = match persistent.VmExtra.ty with Some ty -> ty | None -> vm.ty in
     let hvm = match ty with HVM _ | PVinPVH _ -> true | PV _ -> false in
     (* XXX add per-vcpu information to the platform data *)
     (* VCPU configuration *)
-    let pcpus = Xenctrlext.get_max_nr_cpus xc in
+    let xcext = Xenctrlext.get_handle () in
+    let pcpus = Xenctrlext.get_max_nr_cpus xcext in
     let all_pcpus = mkints pcpus in
     let all_vcpus = mkints vm.vcpu_max in
     let masks =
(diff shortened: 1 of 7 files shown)


andyhhp:
This is the approach suggested by @edwintorok on the upstream thread: https://lore.kernel.org/xen-devel/212D2A4B-9921-4072-8A1A-FCB111251E49@citrix.com/

andyhhp:
I don't like perpetuating the `with_xcext` model of interface.  It would be far better for xenopsd to have real globals for the xc, xs and xcext handles, because they are global singletons

edwintorok:
For xc/ we might be able to create a single global xc and global xce used by everything. Generally it is good to lazy initialize this, so that code in the unit tests which doesn't require xenctrl (but links with it) still works, and that things like --help and --version work even when not running under Xen.
But I think xc/ already has problems in running when not running under the correct version of Xen (which in the CI it won't), so it might not be too bad to do it there. We just can't do it in oxenstored, or it'd break any attempt at testing there.
However xenopsd is a bit more decoupled as it has a simulator that uses just the core from lib/ which doesn't depend on xenctrl.

edwintorok (approved):
Addressing the global xc is probably done in a separate PR, for now this is useful on its own as it is. Thanks.

andyhhp:
> Addressing the global xc is probably done in a separate PR, for now this is useful on its own as it is. Thanks.

Yeah - one problem at once :)

andyhhp:
Untangling the unused variable checks turned into a mess.   I've altered the `with_xcext` interface to be `get_handle` so it can more easily be pulled into a local variable.

psafont (approved):

ocaml/xenopsd/xc/xenops_server_xen.ml
```diff
@@ -1280,12 +1280,13 @@ module VM = struct
...
 
-  let generate_create_info ~xc ~xs:_ vm persistent =
+  let generate_create_info ~xs:_ vm persistent =
     let ty = match persistent.VmExtra.ty with Some ty -> ty | None -> vm.ty in
     let hvm = match ty with HVM _ | PVinPVH _ -> true | PV _ -> false in
     (* XXX add per-vcpu information to the platform data *)
     (* VCPU configuration *)
-    let pcpus = Xenctrlext.get_max_nr_cpus xc in
+    let xcext = Xenctrlext.get_handle () in
+    let pcpus = Xenctrlext.get_max_nr_cpus xcext in
```
robhoes, line 1289: Now (almost) every call to a `Xenctrlext` comes with an additional `Xenctrlext.get_handle ()` call just above. With the handle being cached indefinitely upon the first call, then you might as well call `get_handle ()` inside the library itself as part of every library call.

robhoes (approved):
I'm happy with this as it is. However, I think we can simplify it a little (see comment above).

Issue #4856, opened 2022-11-24
redirect http to https in XAPI when you call the `http(s)://<ip_server>`


benjamreis:
Hi!

As said in the title, we'd like to redirect people accessing the homepage of host through http to https.
Are there, from the go, any problem with that that we're not aware of?

Thanks

robhoes:
Hi, I'm fine with that. However, we have now added an option to close port 80 completely, as some people prefer that for security reason. Port 80 is still open by default, but when closed, then the redirection won't work anymore, as nothing will be listening on port 80.

PR #4859, opened 2022-11-29
Redirect fileserver towards https

benjamreis:
- Add `location` to `Http.Request.Hdr`
- Add `response_redirect` method to `http-svr`
- When `GET` is received by the fileserver and `website-https-only` is true: 
    - if `host` is filled in the request, redirect towards the same URI but in HTTPS 
    - if `host` is not present, reply with a forbidden as before

Solves: https://github.com/xapi-project/xen-api/issues/4856

### DIFF
--- a/ocaml/libs/http-svr/http_svr.ml
+++ b/ocaml/libs/http-svr/http_svr.ml
@@ -210,6 +210,15 @@ let response_method_not_implemented ?req s =
   in
   response_error_html ?version s "501" "Method not implemented" [] body
 
+let response_redirect ?req s dest =
+  let version = Option.fold ~none:"1.1" ~some:get_return_version req in
+  let location = (Http.Hdr.location, dest) in
+  let res =
+    Http.Response.make ~version ~headers:[location] ~body:"" "301"
+      "Moved Permanently"
+  in
+  Unixext.really_write_string s (Http.Response.to_wire_string res)
+
 let response_file ?mime_content_type s file =
   let size = (Unix.LargeFile.stat file).Unix.LargeFile.st_size in
   let mime_header =
--- a/ocaml/xapi/fileserver.ml
+++ b/ocaml/xapi/fileserver.ml
@@ -90,7 +90,14 @@ let send_file (uri_base : string) (dir : string) (req : Request.t)
   let s = Buf_io.fd_of bio in
   Buf_io.assert_buffer_empty bio ;
   if access_forbidden req s then
-    Http_svr.response_forbidden ~req s
+    match req.Request.host with
+    | Some host ->
+        (* Redirect towards HTTPS *)
+        let path = String.concat "" [uri_base; req.Request.uri] in
+        let dest = Uri.make ~scheme:"https" ~host ~path () |> Uri.to_string in
+        Http_svr.response_redirect ~req s dest
+    | None ->
+        Http_svr.response_forbidden ~req s
   else
     let uri = req.Request.uri in
     try
--- a/ocaml/libs/http-svr/http.ml
+++ b/ocaml/libs/http-svr/http.ml
@@ -122,6 +122,8 @@ module Hdr = struct
   let content_disposition = "content-disposition"
 
   let accept = "accept"
+
+  let location = "location"
 end
 
 let output_http fd headers =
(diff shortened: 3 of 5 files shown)


psafont (approved):
This looks good, waiting on Benjamin to post the results of manual testing

ocaml/xapi/fileserver.ml
```diff
@@ -90,7 +90,13 @@ let send_file (uri_base : string) (dir : string) (req : Request.t)
   let s = Buf_io.fd_of bio in
   Buf_io.assert_buffer_empty bio ;
   if access_forbidden req s then
-    Http_svr.response_forbidden ~req s
+    match req.Request.host with
+    | Some host ->
+        (* Redirect towards HTTPS *)
+        let dest = "https://" ^ host ^ uri_base ^ req.Request.uri in
```
lindig, line 96: Is this the best way to construct a URL? At least I would use `String.concat`, which is more efficient. Is there no way to systematically create a URL from components?

lindig (approved):

ocaml/libs/http-svr/http_svr.ml
```diff
@@ -210,6 +210,19 @@ let response_method_not_implemented ?req s =
   in
   response_error_html ?version s "501" "Method not implemented" [] body
 
+let response_redirect ?req s dest =
+  let version =
+    Option.fold ~none:"1.1"
+      ~some:(fun x -> x)
```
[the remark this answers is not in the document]
benjamreis: I've addapt to the name patterns of the other methods: `let response_method_not_implemented`, `response_request_timeout`, `response_file` etc.

ocaml/xapi/fileserver.ml
psafont: Uri can be used here:
```ocaml
Uri.(make ~scheme:"https" ~host ~path:??? |> to_string)
```
https://docs.mirage.io/uri/Uri/index.html#val-make

psafont:
> ```
> File "ocaml/xapi/fileserver.ml", line 97, characters 46-50:
> 97 |         let dest = Uri.(make ~scheme:"https" ~host ~path |> to_string) in
>                                                    ^^^^
> Error: This expression has type Uri.t -> string option
>        but an expression was expected of type string
> ```
> 
> It seems to not take the `host` from the `match` above.

Because Uri is open, host is actually Uri.host, and path is actually Uri.path. The correct expression is:
```ocaml
let dest = Uri.make ~scheme:"https" ~host ~path () |> Uri.to_string
```

benjamreis:
```
File "ocaml/xapi/fileserver.ml", line 97, characters 19-55:
97 |         let dest = Uri.make ~scheme:"https" ~host ~path |> Uri.to_string in
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Warning 5 [ignored-partial-application]: this function application is partial,
maybe some arguments are missing.
File "ocaml/xapi/fileserver.ml", line 97, characters [19](https://github.com/xapi-project/xen-api/actions/runs/3581868804/jobs/6025428280#step:11:20)-55:
97 |         let dest = Uri.make ~scheme:"https" ~host ~path |> Uri.to_string in
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Should I add a `ignore` around it?
Since other args are optionnal this shoulddn't be a problem right?

lindig:
You need a unit `()` argument:
```
 utop # Uri.make;;
- : ?scheme:string ->
    ?userinfo:string ->
    ?host:string ->
    ?port:int ->
    ?path:string ->
    ?query:(string * string list) list -> ?fragment:string -> unit -> Uri.t
```

benjamreis:
Thanks, it seems I need a bit more coffee this morning ^^'

lindig:
"A function taking some optional arguments must also take at least one non-optional argument."

benjamreis:
Okay so testing was successful :) This is good to go for us!

ocaml/libs/http-svr/http_svr.ml
```diff
@@ -210,6 +210,17 @@ let response_method_not_implemented ?req s =
   in
   response_error_html ?version s "501" "Method not implemented" [] body
 
+let response_redirect ?req s dest =
+  let version =
+    Option.fold ~none:"1.1" ~some:Fun.id (Option.map get_return_version req)
```
psafont, line 215: This can be made simpler:
```suggestion
    Option.value ~default:"1.1" get_return_version req
```

psafont (approved):

PR #4861, opened 2022-12-05
Xenctrl: drop interface_close

edwintorok:
This got removed from Xen 4.18. We only use a global, so switch to using Xenctrl.with_intf.

(We can't use a single global 'xc' value, because that might cause some unit tests to fail since initializing xenctrl outside of a Xen system is not possible, so delaying initialization is still needed).


andyhhp:
LGTM

psafont (approved):

robhoes (approved):
This looks backwards compatible, so can go in now.

PR #4864, opened 2022-12-06
CP-41279: add migration_compression pool option

snwoods:

### DIFF
--- a/ocaml/idl/datamodel_lifecycle.ml
+++ b/ocaml/idl/datamodel_lifecycle.ml
@@ -19,6 +19,8 @@ let prototyped_of_field = function
       Some "22.27.0"
   | "host", "last_software_update" ->
       Some "22.20.0"
+  | "pool", "migration_compression" ->
+      Some "22.32.0-next"
   | _ ->
       None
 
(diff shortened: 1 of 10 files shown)


ocaml/idl/datamodel_pool.ml
```diff
@@ -1322,6 +1322,10 @@ let t =
             ~default_value:(Some (VRef null_ref)) "repository_proxy_password"
             "Password for the authentication of the proxy used in syncing with \
              the enabled repositories"
+        ; field ~lifecycle:[] ~ty:Bool ~default_value:(Some (VBool false))
```
robhoes, line 1325: You should add `~qualifier:DynamicRO`. I'm not sure why the compiler didn't complain about this.

ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -19,6 +19,8 @@ let prototyped_of_field = function
       Some "22.27.0"
   | "host", "last_software_update" ->
       Some "22.20.0"
+  | "pool", "migration_compression" ->
+      Some "22.32.0"
```
robhoes, line 23: This should have `22.32.0-next` if the autogenerator works properly...

snwoods: It gave me 22.32.0-next the first time but after some changes only gave me the above

robhoes:
Looks like you need to update the unit tests.

lindig (approved):

ocaml/idl/datamodel_pool.ml
snwoods: Thanks, although it should be RW right?

robhoes (approved):

PR #4866, opened 2022-12-08
CP-41156 Patched XSI 1135 on Havana for NS-SDX

jameshensmancitrix:

### DIFF
--- a/ocaml/xapi/nm.ml
+++ b/ocaml/xapi/nm.ml
@@ -89,7 +89,27 @@ let determine_ethtool_settings properties oc =
       []
   in
   let autoneg = proc "autoneg" in
-  let settings = speed @ duplex @ autoneg in
+  let assoc_opt key value =
+    try Some (List.assoc key value) with Not_found -> None
+  in
+  let advertise =
+    let is_hex x =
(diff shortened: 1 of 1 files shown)


ocaml/xapi/nm.ml
```diff
@@ -89,6 +89,24 @@ let determine_ethtool_settings properties oc =
       []
   in
   let autoneg = proc "autoneg" in
+  let advertise =
+    let is_hex x =
+      try Scanf.sscanf x "0x%x%!" (fun _ -> true)
+      with Stdlib.Scanf.Scan_failure _ -> false
+    in
+    match List.assoc_opt "ethtool-advertise" oc with
```
robhoes, line 97: I think this function is not available due to the compiler version used on this branch. Do it like this: https://github.com/xapi-project/xen-api/pull/4703/files.

lindig: You can use `List.assoc` and cover the exception in the match: `| exception Not_found -> []`

jameshensmancitrix: I also realised I had 2 versions of settings, now amended

robhoes (approved):

psafont (approved):

PR #4868, opened 2022-12-09
CP-40950: Make bias against scheduling vms on pool master configurable

YanRachel:
By design, the pool master has a bias against scheduling VMs on it, preferring other member hosts.
An option to toggle this bias has been added, which can be used to improve performance, for example when the non-master hosts get overloaded.

### DIFF
--- a/ocaml/idl/datamodel_lifecycle.ml
+++ b/ocaml/idl/datamodel_lifecycle.ml
@@ -21,6 +21,8 @@ let prototyped_of_field = function
       Some "22.20.0"
   | "pool", "migration_compression" ->
       Some "22.33.0"
+  | "pool", "coordinator_bias" ->
+      Some "22.34.0-next"
   | _ ->
       None
 
--- a/ocaml/idl/datamodel_pool.ml
+++ b/ocaml/idl/datamodel_pool.ml
@@ -1327,6 +1327,10 @@ let t =
             ~default_value:(Some (VBool false)) "migration_compression"
             "Default behaviour during migration, True if stream compression \
              should be used"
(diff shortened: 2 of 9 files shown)


edwintorok (approved):
Looks good, in the future we may want to switch the default to false once testing has shown that it is not needed anymore.
I think that the protection we have in Dom0 (with control-slice) against qemu taking away CPU from XAPI should be sufficient these days, but some pool bootstorm tests should give us some data on this.

ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -20,7 +20,7 @@ let prototyped_of_field = function
   | "host", "last_software_update" ->
       Some "22.20.0"
   | "pool", "migration_compression" ->
-      Some "22.32.0-next"
```
psafont, line 23: This commit should be dropped, it's now included in the master branch

ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -19,6 +19,8 @@ let prototyped_of_field = function
       Some "22.27.0"
   | "host", "last_software_update" ->
       Some "22.20.0"
+  | "pool", "bias_enabled" ->
```
psafont, line 22: This change is out of date, 22.34.0 has been released. Please change it by removing changes in this file and rebuilding xapi

psafont (changes requested):
The approach seems fine, but a couple of red tape changes got outdated

psafont (approved):

YanRachel:
Tested: https://info.citrite.net/display/~rachely/CP-40950+Notes

PR #4869, opened 2022-12-09
CA-373551: Ensure xapi-events never gets GC'ed

robhoes:


psafont (approved):

robhoes:
I'm going to take a different approach for this event loop.

PR #4870, opened 2022-12-12
CA-373551: register for host events rather than task in events_from_xapi

robhoes:


psafont (approved):
With his change now the loop is woken every time a field from that host object is changed. Even with the increased rate, this seems benign.

PR #4875, opened 2022-12-15
xapi-storage{,-script}: explicitely use python2 instead of python

psafont:
Other distributions other than Xenserver are used for testing the
installation of packages, and they have removed python as an alias for
python2. Being explicit doesn't break compatibility on Xenserver, and
allows us to keep newer versions of Ubuntu for testing in github.

It also has the benefit of reducing confusion when updating to python3 as the change cannot be done implicitly by switching the python binary.

This is a try to fix the latest error in the nightly CI for the docs: https://github.com/xapi-project/xapi-project.github.io/actions/runs/3700548098/jobs/6275403367

- [x] Confirm that xapi-storage-script tests aren't using `python`. (when it's an alias for python3 they fail)
- [x] Pass ring3 BVT and BST, and storage BST  with the changes

### DIFF
--- a/ocaml/xapi-storage-script/examples/datapath/block/Datapath.activate
+++ b/ocaml/xapi-storage-script/examples/datapath/block/Datapath.activate
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 import sys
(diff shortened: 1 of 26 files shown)


lindig (approved):
Mostly just making the binary explicit

edwintorok (approved):

PR #4877, opened 2022-12-16
CP-32622: avoid using select and instead use epoll

snwoods:
To avoid xapi from running out of file descriptors, replace select with epoll in most cases

### DIFF
--- a/ocaml/libs/http-svr/buf_io.ml
+++ b/ocaml/libs/http-svr/buf_io.ml
@@ -79,14 +79,17 @@ let is_full ic = ic.cur = 0 && ic.max = Bytes.length ic.buf
 let fill_buf ~buffered ic timeout =
   let buf_size = Bytes.length ic.buf in
   let fill_no_exc timeout len =
-    let l, _, _ = Unix.select [ic.fd] [] [] timeout in
-    if List.length l <> 0 then (
-      let n = Unix.read ic.fd ic.buf ic.max len in
-      ic.max <- n + ic.max ;
-      if n = 0 && len <> 0 then raise Eof ;
-      n
-    ) else
-      -1
+    Unix.setsockopt_float ic.fd Unix.SO_RCVTIMEO timeout ;
+    let result =
+      try
+        let n = Unix.read ic.fd ic.buf ic.max len in
+        ic.max <- n + ic.max ;
+        if n = 0 && len <> 0 then raise Eof ;
+        n
+      with Unix.Unix_error (Unix.EAGAIN, _, _) -> -1
+    in
+    Unix.setsockopt_float ic.fd Unix.SO_RCVTIMEO 0. ;
+    result
   in
   (* If there's no space to read, shift *)
   if ic.max = buf_size then shift ic ;
--- a/ocaml/libs/stunnel/stunnel.ml
+++ b/ocaml/libs/stunnel/stunnel.ml
@@ -394,7 +394,7 @@ let rec retry f = function
     try f ()
     with Stunnel_initialisation_failed ->
       (* Leave a few seconds between each attempt *)
-      ignore (Unix.select [] [] [] 3.) ;
+      Thread.delay 3. ;
       retry f (n - 1)
   )
 
--- a/ocaml/message-switch/unix/protocol_unix_scheduler.ml
+++ b/ocaml/message-switch/unix/protocol_unix_scheduler.ml
@@ -38,15 +38,15 @@ module Delay = struct
   (* Concrete type is the ends of a pipe *)
   type t = {
       (* A pipe is used to wake up a thread blocked in wait: *)
-      mutable pipe_out: Unix.file_descr option
-    ; mutable pipe_in: Unix.file_descr option
+      mutable sock_out: Unix.file_descr option
+    ; mutable sock_in: Unix.file_descr option
     ; (* Indicates that a signal arrived before a wait: *)
       mutable signalled: bool
     ; m: Mutex.t
   }
 
   let make () =
-    {pipe_out= None; pipe_in= None; signalled= false; m= Mutex.create ()}
+    {sock_out= None; sock_in= None; signalled= false; m= Mutex.create ()}
 
   exception Pre_signalled
 
@@ -59,39 +59,44 @@ module Delay = struct
     finally'
       (fun () ->
         try
-          let pipe_out =
+          let sock_out =
             Mutex.execute x.m (fun () ->
                 if x.signalled then (
                   x.signalled <- false ;
                   raise Pre_signalled
                 ) ;
-                let pipe_out, pipe_in = Unix.pipe () in
+                let sock_out, sock_in =
+                  Unix.socketpair Unix.PF_UNIX Unix.SOCK_STREAM 0
+                in
                 (* these will be unconditionally closed on exit *)
-                to_close := [pipe_out; pipe_in] ;
-                x.pipe_out <- Some pipe_out ;
-                x.pipe_in <- Some pipe_in ;
+                to_close := [sock_out; sock_in] ;
+                x.sock_out <- Some sock_out ;
+                x.sock_in <- Some sock_in ;
                 x.signalled <- false ;
-                pipe_out
+                sock_out
             )
           in
-          let r, _, _ = Unix.select [pipe_out] [] [] seconds in
-          (* flush the single byte from the pipe *)
-          if r <> [] then ignore (Unix.read pipe_out (Bytes.create 1) 0 1) ;
+          (* flush the single byte from the socket *)
+          Unix.setsockopt_float sock_out Unix.SO_RCVTIMEO seconds ;
           (* return true if we waited the full length of time, false if we were woken *)
-          r = []
+          try
+            ignore (Unix.read sock_out (Bytes.create 1) 0 1) ;
+            Unix.setsockopt_float sock_out Unix.SO_RCVTIMEO 0. ;
+            false
+          with Unix.Unix_error (Unix.EAGAIN, _, _) -> true
         with Pre_signalled -> false
       )
       (fun () ->
         Mutex.execute x.m (fun () ->
-            x.pipe_out <- None ;
-            x.pipe_in <- None ;
+            x.sock_out <- None ;
+            x.sock_in <- None ;
             List.iter close' !to_close
         )
       )
 
   let signal (x : t) =
     Mutex.execute x.m (fun () ->
-        match x.pipe_in with
+        match x.sock_in with
         | Some fd ->
             ignore (Unix.write fd (Bytes.of_string "X") 0 1)
         | None ->
--- a/ocaml/networkd/lib/jsonrpc_client.ml
+++ b/ocaml/networkd/lib/jsonrpc_client.ml
@@ -32,8 +32,6 @@ let json_rpc_write_timeout = ref 60000000000L
 
 (* timeout value in ns when writing RPC request *)
 
-let to_s s = Int64.to_float s *. 1e-9
-
 (* Read the entire contents of the fd, of unknown length *)
 let timeout_read fd timeout =
   let buf = Buffer.create !json_rpc_max_len in
@@ -42,13 +40,6 @@ let timeout_read fd timeout =
     Mtime.Span.to_uint64_ns (Mtime_clock.count read_start)
   in
   let rec inner max_time max_bytes =
-    let ready_to_read, _, _ =
-      try Unix.select [fd] [] [] (to_s max_time)
-      with
-      (* in case the unix.select call fails in situation like interrupt *)
-      | Unix.Unix_error (Unix.EINTR, _, _) ->
-        ([], [], [])
-    in
     (* This is not accurate the calculate time just for the select part.
        However, we think the read time will be minor comparing to the scale of
        tens of seconds. the current style will be much concise in code. *)
@@ -60,27 +51,25 @@ let timeout_read fd timeout =
       debug "Timeout after read %d" (Buffer.length buf) ;
       raise Timeout
     ) ;
-    if List.mem fd ready_to_read then
-      let bytes = Bytes.make 4096 '\000' in
-      match Unix.read fd bytes 0 4096 with
-      | 0 ->
-          Buffer.contents buf (* EOF *)
-      | n ->
-          if n > max_bytes then (
-            debug "exceeding maximum read limit %d, clear buffer"
-              !json_rpc_max_len ;
-            Buffer.clear buf ;
-            raise Read_error
-          ) else (
-            Buffer.add_subbytes buf bytes 0 n ;
-            inner remain_time (max_bytes - n)
-          )
-      | exception
-          Unix.Unix_error ((Unix.EAGAIN | Unix.EWOULDBLOCK | Unix.EINTR), _, _)
-        ->
-          inner remain_time max_bytes
-    else
-      inner remain_time max_bytes
+    ( try Unix.setsockopt_float fd SO_RCVTIMEO (Int64.to_float max_time)
+      with Unix.Unix_error (Unix.ENOTSOCK, _, _) ->
+        (* In the unit tests, the fd comes from a pipe... ignore *)
+        ()
+    ) ;
+    let bytes = Bytes.make 4096 '\000' in
+    match Unix.read fd bytes 0 4096 with
+    | 0 ->
+        Buffer.contents buf (* EOF *)
+    | n when n > max_bytes ->
+        debug "exceeding maximum read limit %d, clear buffer" !json_rpc_max_len ;
+        Buffer.clear buf ;
+        raise Read_error
+    | n ->
+        Buffer.add_subbytes buf bytes 0 n ;
+        inner remain_time (max_bytes - n)
+    | exception
+        Unix.Unix_error ((Unix.EAGAIN | Unix.EWOULDBLOCK | Unix.EINTR), _, _) ->
+        inner remain_time max_bytes
   in
   inner timeout !json_rpc_max_len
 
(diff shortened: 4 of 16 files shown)


ocaml/xapi-idl/lib/posix_channel.ml
```diff
@@ -80,12 +80,17 @@ let proxy (a : Unix.file_descr) (b : Unix.file_descr) =
...
       (* If we can't make any progress (because fds have been closed), then stop *)
       if r = [] && w = [] then raise End_of_file ;
-      let r, w, _ = Unix.select r w [] (-1.0) in
-      (* Do the writing before the reading *)
-      List.iter
-        (fun fd -> if a = fd then CBuf.write b' a else CBuf.write a' b)
-        w ;
-      List.iter (fun fd -> if a = fd then CBuf.read a' a else CBuf.read b' b) r ;
+      let epoll = Polly.create () in
+      List.iter (fun fd -> Polly.add epoll fd Polly.Events.inp) (r @ w) ;
```
robhoes, line 84: I think you need `Polly.Events.out` for `w` to wait for it to be ready for writing.

ocaml/xapi-idl/lib/posix_channel.ml
```diff
@@ -80,12 +80,17 @@ let proxy (a : Unix.file_descr) (b : Unix.file_descr) =
...
+      ignore
+      @@ Polly.wait epoll 4 (-1) (fun _ fd _ ->
```
robhoes, line 86: I think that the old code can be summarised as: "Wait for any `r` to be ready to reading or any of the `w` to be ready for writing. Then read/write to the fd found ready. Start again.". In every round of the loop (every time `select` returns) one or more fds would be ready, and it may often be just one. So we can't assume that we can write + read in the same round.

The third argument of the function passed in here, currently ignore with `_`, is set to the kind of event that happened and indicates whether we can read from or write to `fd`.

ocaml/xapi-idl/lib/posix_channel.ml
snwoods: Good point, although we need to read and write: I guess there needs to be two Polly.adds?

lindig: I believe you need to create a set of all events you care about; using `++` in `Polly.Events`.

ocaml/networkd/lib/jsonrpc_client.ml
```diff
@@ -60,27 +53,24 @@ let timeout_read fd timeout =
...
-            Buffer.add_subbytes buf bytes 0 n ;
-            inner remain_time (max_bytes - n)
-          )
-      | exception
-          Unix.Unix_error ((Unix.EAGAIN | Unix.EWOULDBLOCK | Unix.EINTR), _, _)
-        ->
-          inner remain_time max_bytes
-    else
-      inner remain_time max_bytes
+    Unix.setsockopt_float fd Unix.SO_RCVTIMEO (Int64.to_float max_time) ;
```
snwoods, line 56: This line is giving a parse error on `make test`

ocaml/xenopsd/cli/xn.ml
```diff
@@ -1175,27 +1175,38 @@ let raw_console_proxy sockaddr =
...
+                     while
+                       !i < !buf_remote_end + b
+                       && Char.code (Bytes.get buf_remote !i) <> 0x1d
+                     do
+                       incr i
+                     done ;
+                     if !i < !buf_remote_end + b then final := true ;
+                     buf_remote_end := !i
+                   ) ;
+                   if fd = file_desc then
```
lindig, line 1202: An alternative style could use `match`:

```
match file_desc with
| file_desc when file_desc = Unix.stdin -> ..
| file_desc when file_desc = fd -> ..
| _otherwise ->
```

ocaml/forkexecd/src/child.ml
```diff
@@ -59,28 +59,42 @@ let handle_comms_sock comms_sock state =
...
-    debug "comms sock" ;
-    handle_comms_sock comms_sock state
+  debug "Using epoll in handle_comms_no_fd_sock2" ;
+  let epoll = Polly.create () in
+  List.iter
+    (fun fd -> Polly.add epoll fd Polly.Events.inp)
+    [comms_sock; fd_sock] ;
+  (* Although there are two fds, we set max_fds to 1 here as we only want this
+     function to trigger once so that we get one return value *)
+  Polly.wait_fold epoll 1 (-1) state (fun _ fd _ _ ->
```
edwintorok, line 69: Is 1 correct here though, would that prevent it from even watching the other FD?

lindig: My understanding is that one FD would be processed per round - and any other FD that is ready would be reported again in the next round. But does this guarantee fairness?

snwoods: This was the only way I could think to make it work, do you have any other alternatives?

lindig:
Remember that the timeout in `Polly.wait` is specified in milliseconds. Maybe I should change the Polly interface to use a labeled argument `timeout_ms`.

ocaml/xsh/xsh.ml
```diff
@@ -60,12 +60,24 @@ let proxy (ain : Unix.file_descr) (aout : Unix.file_descr) (bin : Unixfd.t)
...
-        w ;
-      List.iter (fun fd -> if ain = fd then read_into a' else read_into b') r
+      let epoll = Polly.create () in
+      List.iter (fun fd -> Polly.add epoll fd Polly.Events.inp) r ;
+      List.iter (fun fd -> Polly.add epoll fd Polly.Events.out) w ;
+      Fun.protect
+        ~finally:(fun () -> Polly.close epoll)
+        (fun () ->
+          ignore
+          @@ Polly.wait epoll 4 (-1) (fun _ fd _ ->
```
lindig, line 70: Note that the code below matches exactly one condition and not all fds that are ready; could this be written as
```
if fd = x then do something;
if fd = y then do something;
..
```

Maybe leave a comment why the selected behavior is the right one.

ocaml/xenopsd/cli/xn.ml
lindig: Do we want to handle one or all FDs that are ready? Up to `2` can be ready. This code handles two if they are ready but we have other places where we handle only one. It's not obvious why we have different behavior.

ocaml/xapi-idl/lib/posix_channel.ml
```diff
@@ -80,19 +80,33 @@ let proxy (a : Unix.file_descr) (b : Unix.file_descr) =
...
-          if CBuf.end_of_writes buf then Unix.shutdown fd Unix.SHUTDOWN_RECEIVE
+      let epoll = Polly.create () in
+      List.iter (fun fd -> Polly.add epoll fd Polly.Events.inp) r ;
+      List.iter (fun fd -> Polly.add epoll fd Polly.Events.out) w ;
+      Fun.protect
+        ~finally:(fun () -> Polly.close epoll)
+        (fun () ->
+          ignore
+          @@ Polly.wait epoll 4 (-1) (fun _ fd event ->
+                 if event = Polly.Events.inp then
```
lindig, line 91: Again, this handles at most one FD. I think the unconditional `else` case is a bit dangerous and I would explicitly check than any FD is ready before acting on it. This applies to other places as well.

ocaml/networkd/lib/jsonrpc_client.ml
```diff
@@ -60,27 +51,28 @@ let timeout_read fd timeout =
...
+    ( try Unix.setsockopt_float fd SO_RCVTIMEO (Int64.to_float max_time)
+      with Unix.Unix_error (Unix.ENOTSOCK, _, _) ->
+        (* In the unit tests, the fd comes from a pipe... ignore *)
+        ()
+    ) ;
+    let bytes = Bytes.make 4096 '\000' in
+    match Unix.read fd bytes 0 4096 with
+    | 0 ->
+        Buffer.contents buf (* EOF *)
+    | n ->
```
lindig, line 63: Could be expressed as 
```
| n when n >= max_bytes -> ..
| n (* otherwise *) ->
| exception .. ->
```

lindig (approved):

ocaml/xapi-idl/lib/posix_channel.ml
```diff
@@ -153,27 +167,40 @@ let send proxy_socket =
...
+                                   (String.length token) [] proxy_socket
+                               in
+                               ()
+                           ) else if s_ip = fd then (
+                             let fd, _peer = Unix.accept s_ip in
+                             List.iter close !to_close ;
+                             to_close := fd :: !to_close ;
+                             proxy fd proxy_socket
+                           ) else
+                             assert false (* can never happen *)
```
edwintorok, line 201: Wouldn't be so sure, we could get spurious notifications from 'epoll', and we don't want to crash the application (or thread). Perhaps just log it as error for now?
Especially that you've added 2 file descriptors s_ip and s_unix, this will NOT be just s_unix, it is entirely possible that s_ip will have something available, otherwise why add it to epoll in the first place?

ocaml/database/block_device_io.ml
```diff
@@ -326,13 +326,13 @@ let accept_conn s latest_response_time =
...
-    (* We've received a connection. Accept it and return the socket. *)
-    fst (Unix.accept s)
-  else (* We must have timed out *)
-    raise Unixext.Timeout
+  Fun.protect
+    ~finally:(fun () -> Unix.setsockopt_float s Unix.SO_RCVTIMEO 0.)
+    (fun () ->
+      Unix.setsockopt_float s Unix.SO_RCVTIMEO timeout ;
+      try fst (Unix.accept s)
+      with Unix.Unix_error (Unix.EAGAIN, _, _) -> raise Unixext.Timeout
```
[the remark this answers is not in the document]
snwoods: It looks like you're right, the only examples online of using a timeout with accept are using select

snwoods:
I believe all of the comments have now been addressed.

ocaml/database/block_device_io.ml
```diff
@@ -326,13 +326,22 @@ let accept_conn s latest_response_time =
...
+        Polly.wait_fold epoll 1
+          (int_of_float (timeout *. 1000.))
+          []
+          (fun _ fd _ fds -> fst (Unix.accept fd) :: fds)
+      )
+  in
+  if fds = [] then (* We must have timed out *)
     raise Unixext.Timeout
+  else (* There will only ever be a maximum of one fd *)
+    List.hd fds
```
psafont, line 344: While List.hd is guarded here, it's preferrable to destructure the list to obtain values from it:
```suggestion
  match fds with
  | [] -> (* We must have timed out *)
      raise Unixext.Timeout
  | fd :: _ -> (* There will only ever be a maximum of one fd *)
      fd
```

snwoods:
All comments addressed and suiterun 179915 passing.

ocaml/libs/stunnel/stunnel.ml
```diff
@@ -394,7 +394,7 @@ let rec retry f = function
     try f ()
     with Stunnel_initialisation_failed ->
       (* Leave a few seconds between each attempt *)
-      ignore (Unix.select [] [] [] 3.) ;
+      Unix.sleep 3 ;
```
robhoes, line 397: Use `Thread.delay`

ocaml/libs/http-svr/buf_io.ml
```diff
@@ -79,14 +79,17 @@ let is_full ic = ic.cur = 0 && ic.max = Bytes.length ic.buf
...
+    Unix.setsockopt_float ic.fd Unix.SO_RCVTIMEO timeout ;
+    let result =
+      try
+        let n = Unix.read ic.fd ic.buf ic.max len in
+        ic.max <- n + ic.max ;
+        if n = 0 && len <> 0 then raise Eof ;
+        n
+      with Unix.Unix_error (Unix.EAGAIN, _, _) -> -1
+    in
+    Unix.setsockopt_float ic.fd Unix.SO_RCVTIMEO 0. ;
```
robhoes, line 91: I think we'd better use `finally` to ensure that this timeout is reset in case of any exception (only one kind is caught above.

ocaml/message-switch/unix/protocol_unix_scheduler.ml
```diff
@@ -59,39 +59,44 @@ module Delay = struct
...
-          let r, _, _ = Unix.select [pipe_out] [] [] seconds in
-          (* flush the single byte from the pipe *)
-          if r <> [] then ignore (Unix.read pipe_out (Bytes.create 1) 0 1) ;
+          (* flush the single byte from the socket *)
+          Unix.setsockopt_float sock_out Unix.SO_RCVTIMEO seconds ;
           (* return true if we waited the full length of time, false if we were woken *)
-          r = []
+          try
+            ignore (Unix.read sock_out (Bytes.create 1) 0 1) ;
+            Unix.setsockopt_float sock_out Unix.SO_RCVTIMEO 0. ;
```
robhoes, line 84: Here we'd miss resetting the socket option in case of an exception.

ocaml/networkd/lib/jsonrpc_client.ml
```diff
@@ -60,27 +51,28 @@ let timeout_read fd timeout =
...
-            Buffer.add_subbytes buf bytes 0 n ;
-            inner remain_time (max_bytes - n)
-          )
-      | exception
-          Unix.Unix_error ((Unix.EAGAIN | Unix.EWOULDBLOCK | Unix.EINTR), _, _)
-        ->
-          inner remain_time max_bytes
-    else
-      inner remain_time max_bytes
+    ( try Unix.setsockopt_float fd SO_RCVTIMEO (Int64.to_float max_time)
```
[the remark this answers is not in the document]
robhoes: Yes, this would avoid mistakes, such as around resetting the socket option.

ocaml/libs/http-svr/buf_io.ml
edwintorok: Every `read` on `ic.fd` seems to go through here, and each read is preceded by setting a timeout, so I don't think we even need to reset the timeout, just drop this line.

edwintorok: There are more reads in http.ml but they also set a socket timeout. Will need to check the other callers too.

edwintorok:
Testing whether 'select' is completely gone or not is tricky due to a lot of dead code in the linked binary.
I hacked up a script to find at least the actual OCaml functions that call 'Unix.select' directly:
```
#!/bin/sh
set -eu

EXE="$1"
ADDR="${EXE}.unix_select.address"
DISASM="${EXE}.disasm"

# Find address of 'unix_select' symbol
addr2line -pae "${EXE}" unix_select \
  | sed -e "s/^0x0*\\([0-9a-fA-F]*\\):.*/\\1/" \
  | tr -d '\n' >"${ADDR}"

# Find all calls to 'unix_select', which is done by
#  mov $0x...,%rax
#  call <caml_c_call>

# disassemble, but without sources
# (source lookup doesn't work for all dependencies, and is very slow on a large binary)
# To make debugging easier the disassembly is saved to a file instead of piping
objdump -d --no-show-raw-insn --no-addresses "${EXE}" >"${DISASM}"

SYMADDR=$(cat "${ADDR}")
# Prints a list of functions that call unix_select
# But we won't know whether it is dead code or not
gawk -f unix_select.gawk "${SYMADDR}" <"${DISASM}"
```

unix_select.gawk:
```
BEGIN { SYMADDR=ARGV[1]; ARGV[1]=""; }

/^<.*/ { SYMBOL=$1 }
match($0, /mov.*0x([0-9a-fA-F]*),%rax/, addr) { RAX=addr[1]; }
/call.*<caml_c_call>/ { if (RAX == SYMADDR) { print SYMBOL }  }
```

It shows these for XAPI and xenopsd:
```
./no-unix-select.sh _build/default/ocaml/xapi/xapi_main.exe
<camlXapi_stdext_threads__Threadext__fun_812>:
<camlXapi_stdext_unix__Unixext__kill_and_wait_inner_2120>:
<camlXapi_stdext_unix__Unixext__proxy_1371>:
<camlXapi_stdext_unix__Unixext__time_limited_write_internal_1424>:
<camlXapi_stdext_unix__Unixext__time_limited_read_1450>:
<camlThread__wait_timed_read_777>:
<camlThread__wait_timed_write_781>:
<camlUnix__fun_2441>:
```

And for xenopsd:
```
./no-unix-select.sh _build/default/ocaml/xenopsd/xc/xenops_xc_main.exe
<camlXapi_stdext_unix__Unixext__kill_and_wait_inner_2120>:
<camlXapi_stdext_unix__Unixext__proxy_1371>:
<camlXapi_stdext_unix__Unixext__time_limited_write_internal_1424>:
<camlXapi_stdext_unix__Unixext__time_limited_read_1450>:
<camlXapi_stdext_threads__Threadext__fun_812>:
<camlWatch__fun_884>:
<camlThread__wait_timed_read_777>:
<camlThread__wait_timed_write_781>:
<camlUnix__fun_2441>:
```

edwintorok:
The 'select' in watch.ml can be replaced similarly as elsewhere by changing pipe->socket and using socket timeouts.

edwintorok:
Removed some Unix.select and Thread.wait* usages in https://github.com/xapi-project/stdext/pull/80.

And a few more from XAPI and xenopsd as well: https://github.com/edwintorok/xen-api/commits/private/edvint/CP-32622

We'll probably need a CA ticket and backport the watch.ml fix since it affects xenopsd which already claims to support >1024 fds.

I think these now remove all the obvious Unix.select usages, but we should implement what has been suggested here and add some CI checks in xs-opam with various methods to check that we really don't call it anymore (and then add some unit tests to all these functions that we modified to check that they still work correctly):
https://discuss.ocaml.org/t/ensuring-that-an-ocaml-binary-never-calls-unix-select/13465/7?u=edwin

edwintorok:
I've updated the `no-unix-select.sh` to be recursive up to a certain depth, and it didn't find more calls.
Going to try the `.cmt` iterator suggested in the forums just to have another way to validate this.

snwoods:
I'm closing this and have opened a PR against Edwin's fork so that he can still keep track of it https://github.com/edwintorok/xen-api/pull/4

PR #4878, opened 2022-12-19
CA-373776 

jameshensmancitrix:
Added unhandled exception handler to nbd to log errors instead of the messages being printed to the host console


lindig:
A better subject line would be "CA-373776 catch unhandled exceptions and log them." "Added unhandled exception handler" is ambiguous.

ocaml/nbd/src/main.ml
```diff
@@ -255,4 +255,9 @@ let () =
      in the failed state, and a backtrace will show up in the logs. *)
   Cleanup.Runtime.register_signal_handler () ;
   setup_logging () ;
+  let handler exn _ =
+    Lwt_log.ign_error (Printexc.to_string exn) ;
+    Lwt_log.ign_error (Printexc.get_backtrace ())
```
lindig, line 260: Could the logging call destroy the backtrace? I think it would be best to let-bind the backtrace before doing anything else in the handler.

lindig (approved):

ocaml/nbd/src/main.ml
jameshensmancitrix: Not sure on that one, I can do a let bind to ensure that

jameshensmancitrix: The handler does have a backtrace on it I am no using to print the backtrace.

stormi:
> A better subject line would be "CA-373776 catch unhandled exceptions and log them." "Added unhandled exception handler" is ambiguous.

It would also be nice to update the PR title with this, as the CA number's signification is not obvious at first read (nor at second read if you don't have access to the internal issue tracker it refers to).

ocaml/nbd/src/main.ml
```diff
@@ -177,14 +177,10 @@ let main port certfile _ =
...
   (* Log unexpected exceptions *)
   let () =
-    Lwt_main.run
-      (Lwt.catch t (fun e ->
-           Lwt_log.fatal_f "Caught unexpected exception: %s"
-             (Printexc.to_string e)
-           >>= fun () -> Lwt.fail e
-       )
-      )
+    try Lwt_main.run (ignore_exn_delayed t ())
```
edwintorok, line 180: This is correct I think: Lwt_main.run according to its documentation will raise an exception if the promise failed, and the try/with will also catch any OCaml exceptions that might escape.

You might also want to set the async exception hook https://ocsigen.org/lwt/latest/api/Lwt#VALasync_exception_hook, which by default prints the exception and terminates the program.

ocaml/nbd/src/main.ml
```diff
@@ -177,14 +177,10 @@ let main port certfile _ =
...
+    with e -> Lwt_log.ign_fatal (Printexc.to_string e)
```
edwintorok, line 181: I think this changes the exitcode of the process which isn't necessarily right. If we had an exception we want a non-zero exitcode, like before.

edwintorok: Try adding a call to 'exit 3' here or something similar after logging the exception. Although the Lwt main loop is not running anymore at this point so I doubt we'll succeed in logging it.
Maybe it'd be better to use Lwt_main.run (Lwt_log.fatal...)

ocaml/nbd/src/main.ml
```diff
@@ -176,15 +176,7 @@ let main port certfile _ =
       (ignore_exn_delayed (fun () -> Lwt_unix.close sock))
   in
   (* Log unexpected exceptions *)
-  let () =
-    Lwt_main.run
-      (Lwt.catch t (fun e ->
-           Lwt_log.fatal_f "Caught unexpected exception: %s"
-             (Printexc.to_string e)
-           >>= fun () -> Lwt.fail e
```
psafont, line 184: The exception is no longer reraised, instead it's ignored. I think in case of unhandled exception nbd should fail with an error code.

I recommend changing this back to how it was and then change `fatal_f` for `error_f`

jameshensmancitrix:
Will do a final test before merging

PR #4879, opened 2022-12-21
Switch from Xenctrl.hvm_check_pvdriver to Xenctrl.hvm_param_get

andyhhp:
The former is technical debt in the hypervisor patchqueue, while the latter is an upstream binding.

No functional change.


edwintorok:
Might also need to update the [mock xenctrl](https://github.com/xapi-project/xenctrl) to make the CI happy

PR #4884, opened 2023-01-10
Reorganise SMAPI code in xapi for compatibility with next-gen SMAPIv3 backends

robhoes:
For SMAPIv1 storage backends to work correctly, xapi has functionality in a wrapper around SMAPIv2 calls that implements:

1. a state machine, which ensures VM storage does not get detached during a localhost migration;
2. locking to correctly handle concurrent calls;
3. handling of "leaked" datapaths for the case errors occur when a VDI is detached.

The current SMAPIv3 backend still relies on the first of these. A new implementation (which is still in the works), however, requires the state machine to be removed. SMAPIv3 already does not require the other two part of the wrapper.

This PR pushes the wrapper down the stack such that it only covers SMAPIv1 calls. This needs quite a bit of change in the Storage_mux module, which is now directly the entry point for SMAPIv2 calls in xapi. In particular, some state needs to be maintained here to keep track of "datapaths", to ensure that the `DP.destroy` call still works. This may get deprecated/removed at some point later.


lindig (approved):
This is obviously complex. I looked mostly for OCaml foot guns but didn't spot any.

ocaml/xapi/xapi.ml
```diff
@@ -1030,11 +1030,14 @@ let server_init () =
           ; ("Initialise TLS verification", [], init_tls_verification)
           ; ("Running startup check", [], startup_check)
           ; ("Registering SMAPIv1 plugins", [Startup.OnlyMaster], Sm.register)
+          ; ( "Initialising SMAPIv1 state"
+            , []
+            , Storage_smapiv1_wrapper.initialise
+            )
```
jameshensmancitrix, line 1036: I am not sure if it would cause any issues, why are these two initialisation tasks swapped in order?

ocaml/xapi/storage_mux.ml
```diff
@@ -151,6 +153,49 @@ module Mux = struct
       )
   end
 
+  module DP_info = struct
+    type t = {sr: Sr.t; vdi: Vdi.t; vm: Vm.t} [@@deriving rpcty]
+
+    let storage_dp_path = "/var/run/nonpersistent/xapi/storage-dps"
```
edwintorok, line 161: What happens if you do a toolstack-restart on a running system that had the old datapaths, will the new XAPI be able to find them, or will we require a host reboot?

ocaml/xapi/storage_mux.ml
```diff
@@ -718,19 +722,8 @@ module Mux = struct
...
-
-    let destroy () ~dbg:_ ~task:_ = assert false
-
-    let list () ~dbg:_ = assert false
-  end
-
-  module UPDATES = struct
-    let get () ~dbg:_ ~from:_ ~timeout:_ = assert false
-  end
+  module TASK = Storage_smapiv1_wrapper.Impl.TASK
```
edwintorok, line 734: there is a smapiv3 task api, but I don't think any SRs actually implement it currently (it is meant to run things like the SMGC in a way that is not totally opaque to XAPI, or long running tasks such as mirroring/etc.).
I don't know how useful it'd be, maybe @MarkSymsCtx knows whether we intend to use the SMAPIv3 tasks API.

ocaml/xapi/storage_mux.ml
MarkSymsCtx: We definitely require a host evacuate, and possibly a reboot. Nothing we can do to avoid unfortunately, the implementation of the original SMAPIv3 adapter violated its design and architectural constraints meaning that it is not possible to make the Storage Management scripts work as the design and architecture says they should as they do not currently get called correctly.

ocaml/xapi/storage_mux.ml
MarkSymsCtx: I never even knew there was such a thing.

edwintorok: https://xapi-project.github.io/xapi-storage/#task-errors I don't know whether it works at all, probably hooked up in all the right places in XAPI, but if it is useful you can using it.

psafont (approved):
I'm not a domain expert, the changes look reasonable here.

The only change that I see that makes sense is adding the redirection of the tasks to smapiv3 in the muxer so they can start being used when the backend uses them. That will allow to spot issues in the area

robhoes:
I have rebased this. We are ready to progress this now.

PR #4887, opened 2023-01-16
CA-374274: Provide more information when datasource is not found

psafont:
Previously when a requested datasource was not recorded, an unhelpful message "Xcp_rrdd_internal_error Not_found" was generated. Instead create a new error to signal when a source is not being recorded. This also contains what source is missing: the object related to the datasource (a VM or SR), or just the datasource.

### DIFF
--- a/ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
+++ b/ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
@@ -12,9 +12,6 @@
  * GNU Lesser General Public License for more details.
  *)
 
-(* The framework requires type 'context' to be defined. *)
-type context = unit
-
 let with_lock = Xapi_stdext_threads.Threadext.Mutex.execute
 
 open Rrdd_shared
--- a/ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
+++ b/ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
@@ -339,33 +339,46 @@ let send_host_rrd_to_master master_address =
   | None ->
       ()
 
+let fail_missing name = raise (Rrdd_error (Datasource_missing name))
+
 (** {add_ds rrdi ds_name} creates a new time series (rrd) in {rrdi} with the
     name {ds_name}. The operation fails if rrdi does not contain any live
     datasource with the name {ds_name} *)
 let add_ds ~rrdi ~ds_name =
-  let open Ds in
-  let ds = List.find (fun ds -> ds.ds_name = ds_name) rrdi.dss in
+  match List.find_opt (fun ds -> ds.Ds.ds_name = ds_name) rrdi.dss with
+  | None ->
+      fail_missing ds_name
+  | Some ds ->
+      let now = Unix.gettimeofday () in
+      Rrd.rrd_add_ds rrdi.rrd now
+        (Rrd.ds_create ds.ds_name ds.ds_type ~mrhb:300.0 Rrd.VT_Unknown)
+
+let add rrds uuid domid ds_name rrdi =
+  let rrd = add_ds ~rrdi ~ds_name in
+  Hashtbl.replace rrds uuid {rrd; dss= rrdi.dss; domid}
+
+let forget rrds ~uuid ~ds_name rrdi =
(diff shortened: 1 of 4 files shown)


ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
```diff
@@ -347,10 +347,12 @@ let send_host_rrd_to_master master_address =
     datasource with the name {ds_name} *)
 let add_ds ~rrdi ~ds_name =
   let open Ds in
-  let ds = List.find (fun ds -> ds.ds_name = ds_name) rrdi.dss in
-  let now = Unix.gettimeofday () in
-  Rrd.rrd_add_ds rrdi.rrd now
-    (Rrd.ds_create ds.ds_name ds.ds_type ~mrhb:300.0 Rrd.VT_Unknown)
+  try
+    let ds = List.find (fun ds -> ds.ds_name = ds_name) rrdi.dss in
```
lindig, line 351: Use `List.find_opt` and make it more explicit which error we are handling?

ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
```diff
@@ -454,30 +456,42 @@ let query_host_ds (ds_name : string) : float =
     {ds_name}. *)
 let add_vm_ds (vm_uuid : string) (domid : int) (ds_name : string) : unit =
   with_lock mutex (fun () ->
-      let rrdi = Hashtbl.find vm_rrds vm_uuid in
-      let rrd = add_ds ~rrdi ~ds_name in
-      Hashtbl.replace vm_rrds vm_uuid {rrd; dss= rrdi.dss; domid}
+      try
+        let rrdi = Hashtbl.find vm_rrds vm_uuid in
```
lindig, line 460: Same here, use `find_opt`?

lindig (approved):
I suggest to use `find_opt` to make it explicit which error we are handling. IN the backport we could a module-local implementation of `find_opt` (or use this code).

ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
```diff
@@ -347,30 +347,42 @@ let send_host_rrd_to_master master_address =
...
     Rrd.rrd_add_ds rrdi.rrd now
       (Rrd.ds_create ds.ds_name ds.ds_type ~mrhb:300.0 Rrd.VT_Unknown)
-  with Not_found -> raise (Rrdd_error (Datasource_missing ds_name))
+  in
+  let fail = raise (Rrdd_error (Datasource_missing ds_name)) in
+
+  List.find_opt (fun ds -> ds.ds_name = ds_name) rrdi.dss
+  |> Option.fold ~some:add ~none:fail
+
+let add_rrdi rrds uuid domid ds_name rrdi =
```
lindig, line 360: Naming is a bit inconsistent - this could be just `add` to match the others or they could have a `_rrdi` suffix.

lindig: I think a global `fail` function that raises `Rrdd_error (Datasource_missing fmt)` and takes a printf-style format string could be useful.

lindig (approved):

ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
psafont: I've changed the name to add and made sure it's not exposed to other modules by adding an mli.

I'll look into the fail one

psafont: I've added the fail function, it does shorten the functions quite a bit

psafont:
I've added some find_opts where it made sense. I think it's a bad fit for raising exception because using Option.fold and exception is awkward: the exception raising needs to be delayed. This can be done by adding a unit parameter, but this affect the functions that return a value.

I've added this option as a separate commit, that I think should be dropped. I can try using results here which are probably better for this, but does it make sense?

lindig:
I agree that `Option.find` and exceptions don't mix. I would have used an explicit `match` where I would have raised the exception or called the `fail` function that does it. You approached this from the other end by using `Option.find` - now raising exceptions is a bit awkward.

PR #4889, opened 2023-01-16
CA-343683 Atomising Networkd pif plugging on Xapi startup

jameshensmancitrix:
network config will not write to disk when network state is being refreshed when starting up

- 	added `config_writable` flag. config will write on if true but will remain in memory otherwise.

- 	`config_writable` is set to false when the state is cleared

- 	added `sync_state` function to enable disk writing then write to disk: synchronising memory and disk.

This will prevent the chance of incomplete config in `networkd.db` which can cause network issues on reboot.

### DIFF
--- a/ocaml/networkd/bin/network_server.ml
+++ b/ocaml/networkd/bin/network_server.ml
@@ -27,11 +27,14 @@ let network_conf = ref "/etc/xcp/network.conf"
 
 let config : config_t ref = ref Network_config.empty_config
 
+let write_lock = ref false
+
 let backend_kind = ref Openvswitch
 
 let write_config () =
-  try Network_config.write_config !config
-  with Network_config.Write_error -> ()
+  if not !write_lock then
+    try Network_config.write_config !config
+    with Network_config.Write_error -> ()
 
 let read_config () =
   try
@@ -58,7 +61,13 @@ let on_shutdown signal =
 
 let on_timer () = write_config ()
 
-let clear_state () = config := Network_config.empty_config
+let clear_state () =
+  write_lock := true ;
+  config := Network_config.empty_config
+
+let sync_state () =
+  write_lock := false ;
+  write_config ()
 
 let reset_state () = config := Network_config.read_management_conf ()
 
--- a/ocaml/xapi-idl/network/network_interface.ml
+++ b/ocaml/xapi-idl/network/network_interface.ml
@@ -336,12 +336,17 @@ module Interface_API (R : RPC) = struct
 
   let clear_state =
     declare "clear_state"
-      ["Clear configuration state"]
+      ["Clear configuration state then lock the writing of the state to disk"]
       (unit_p @-> returning unit_p err)
 
   let reset_state =
     declare "reset_state"
       ["Reset configuration state"]
+      (unit_p @-> returning unit_p err)
+
+  let sync_state =
+    declare "sync_state"
+      ["Allow for the config state to be written to disk then perform a write"]
       (unit_p @-> returning unit_p err)
 
   let set_gateway_interface =
--- a/ocaml/networkd/bin/networkd.ml
+++ b/ocaml/networkd/bin/networkd.ml
@@ -174,6 +174,7 @@ let bind () =
   let open Network_server in
   S.clear_state clear_state ;
   S.reset_state reset_state ;
+  S.sync_state sync_state ;
   S.set_gateway_interface set_gateway_interface ;
   S.set_dns_interface set_dns_interface ;
   S.Interface.get_all Interface.get_all ;
--- a/ocaml/xapi/xapi_pif.ml
+++ b/ocaml/xapi/xapi_pif.ml
@@ -1111,5 +1111,6 @@ let start_of_day_best_effort_bring_up () =
             )
             pif
         )
-        (calculate_pifs_required_at_start_of_day ~__context)
+        (calculate_pifs_required_at_start_of_day ~__context) ;
+      Net.sync_state ()
   )


ocaml/networkd/bin/network_server.ml
```diff
@@ -58,7 +61,13 @@ let on_shutdown signal =
 
 let on_timer () = write_config ()
 
-let clear_state () = config := Network_config.empty_config
+let clear_state () =
+  config := Network_config.empty_config ;
+  config_writable := false
+
+let sync_state () =
```
psafont, line 68: Can you change the name to `persist_state`? Sync can be quite a nebulous action, while persist is much more specific.

ocaml/xapi-idl/network/network_interface.ml
```diff
@@ -336,7 +336,12 @@ module Interface_API (R : RPC) = struct
 
   let clear_state =
     declare "clear_state"
-      ["Clear configuration state"]
+      ["Clear configuration state and pause writing to disk"]
+      (unit_p @-> returning unit_p err)
+
+  let sync_state =
```
psafont, line 342: Can you change the name to `persist_state`? Sync can be quite a nebulous action, while persist is much more specific.

ocaml/networkd/bin/network_server.ml
```diff
@@ -27,11 +27,14 @@ let network_conf = ref "/etc/xcp/network.conf"
 
 let config : config_t ref = ref Network_config.empty_config
 
+let config_writable = ref true
```
psafont, line 30: Adding this reference makes implementation more difficult (on sync_state it should be turned on, write, then turn off the switch)

I think calling write_config only when xapi calls `persist_state` makes sense

That means it should stopped being called on shutdown (`on_shutdown`) and periodically (`on_timer`) because it cannot be guaranteed at those points that the configuration is total

psafont (changes requested):

ocaml/networkd/bin/network_server.ml
jameshensmancitrix: I would have to see if at any time the configuration is changed other than this one point in xapi_pif, if so it may be worth keeping this.

jameshensmancitrix: Taking a deeper look it does seem as if `Nm.bring_pif_up` is called in numerous places which is also the mechanism used to change the config. This mechanism using on_timer and on_shutdown seems to account for this change

psafont: The problem is that on_timer and on_shutdown persist the configuration without ever ensuring that the configuration is not incomplete. This is why I think only the API for explicitely writing is the only way to write the configuration to disk. This allows xapi to ensure that the configuration is total and only then write it.

jameshensmancitrix: The only place I can see this being an issue is in the on-startup iterator which clears the in-memory config and rebuilds it. I wouldn't expect to have an incomplete config in any other place. It is in this place, we need to disable writing so that even on those two calls the state is never persisted and remains to be the existing state. What other places in the code would there be an incomplete state?

psafont: > What other places in the code would there be an incomplete state?

That's the thing, networkd has no way of knowing whether the configuration it holds is complete or not, this means it's open to write incomplete configurations when it writes by itself.

Example: xapi wants to send configuration for 8 interfaces
- xapi sends configuration for 4 interfaces
- networkd periodic timer ends, `on_timer` gets executed and networkd writes the configurations
- xapi sends configuration for 2 more interfaces
- xapi gets interrupted
- now you're left with an incomplete configuration on disk, precisely what we want to avoid in this PR

psafont: > This is why config_writable is used to disable writing

The `sync_state` function is missing the disablement after having written the state, as I said on the first message. 

This error in the implementation is why I'm advising to go for a far simpler solution that is much harder to misuse.

ocaml/xapi/nm.ml
```diff
@@ -566,6 +566,7 @@ let bring_pif_up ~__context ?(management_interface = false) (pif : API.ref_PIF)
               Option.iter
                 (fun (_, name) -> Net.set_dns_interface dbg name)
                 dns_if ;
+              Net.write_config () ;
```
psafont, line 569: This can't be done until all the `Net.*.make_config` calls have completed, at around line 693

ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -20,7 +20,7 @@ let prototyped_of_field = function
   | "host", "last_software_update" ->
       Some "22.20.0"
   | "pool", "coordinator_bias" ->
-      Some "22.34.0-next"
```
psafont, line 23: Please split this change into a separate commit. Having separate changes in different commits makes it easier to reverse any of them separately if needed

psafont:
With the original plan partial configurations could be written to disk when processed were interrupted, which I thought we wanted to avoid. By exposing the syncing to xapi this can be avoided and only be written when xapi has ensured the configuration is total.

To avoid missing writes a `with_` function could be added, or reused to ensure the write happens after the function containing configuration changes have ended without an error.

jameshensmancitrix:
Outside of `Xapi_pif`there are other instances of non-atomic changes to the config which could be written if on_timer or on_shutdown were called in the middle of one of these batches. The idea here is to remove the chance of disk writes during one of these batched changes and only write after. So far this does pass the BST. It is as easy to do this change as adding write prevention to the other changes.

jameshensmancitrix:
I see the `write_config ()` in `nm.ml` which will call at a bad time, I will remove this.

robhoes:
Now the startup case may be fixed, but the config is not persisted after any other network changes. There are many callers of `Nm.bring_pif_up` and `Nm.bring_pif_down` outside of the calls starting with `start_of_day_best_effort_bring_up`.

However, you are right that there is scope for partial updates to be written out between calls to xcp-networkd inside of `Nm.bring_pif_up`. We absolutely have to make sure that any state change coming from whatever direction is eventually persisted, unless there is an interruption. The original plan was to have persistence by default, but add a transaction-like mechanism to _temporarily_ stop this. We could extend this to cover more cases, but have to take care of nested calls from `Xapi_pif` to `Nm` etc.

I am fine with removing the relative unpredictability of the timer in xcp-networkd, if we then ensure that it is impossible to "forget" to persist the state after an update.

robhoes:
I've looked at the latest changes, but I am sorry to say that I am still not happy with this approach, for the same reasons as in my last comment. In particular about making it impossible to forget to get the state persisted. Right now, we need to add `Network.Net.write_config` everywhere.

The fact that it is easy to forget is proven by the fact that `Xapi_pif.plug` does not have this in the current code, which affects the `PIF.plug` API call (and e.g. `xe pif-plug`). That function is called from the start-of-day function as well, which is probably why you didn't call `write_config` in there, because it would give us the original problem back again.

Now we can distinguish cases in `Xapi_pif.plug` etc. to work around this to make it work _here_. But I still don't like the general approach... :(

At this point, I would go back to the original, very simple fix. This addresses the main bug and doesn't make anything else worse. There may be other windows for things to go wrong, but I think those are both much less likely, and less problematic.

psafont (approved):

robhoes (approved):

PR #4863, opened 2022-12-06
Fix extra `/` in https redirection

benjamreis:
Do not use `uri_base` when redirecting towards https

Signed-off-by: BenjiReis <benjamin.reis@vates.fr>

PR #4890, opened 2023-01-19
CA-373785: Deny HTTP requests on website_https_only

psafont:
With the previous patch the requests were redirected to HTTPS instead.

Current behaviour is to deny all HTTP requests when website_https_only
is turned on. When it's not, try to redirect the request from HTTP to
HTTPS.

supersedes #4863

### DIFF
--- a/ocaml/idl/datamodel_lifecycle.ml
+++ b/ocaml/idl/datamodel_lifecycle.ml
@@ -19,10 +19,10 @@ let prototyped_of_field = function
       Some "22.27.0"
   | "host", "last_software_update" ->
       Some "22.20.0"
-  | "pool", "migration_compression" ->
-      Some "22.33.0"
   | "pool", "coordinator_bias" ->
       Some "22.34.0-next"
+  | "pool", "migration_compression" ->
+      Some "22.33.0"
   | _ ->
       None
 
--- a/ocaml/xapi/fileserver.ml
+++ b/ocaml/xapi/fileserver.ml
@@ -78,26 +78,27 @@ let response_file s file_path =
   in
   Http_svr.response_file ~mime_content_type s file_path
 
+let is_external_http req s =
+  (not (Context.is_unix_socket s)) && Context._client_of_rq req = None
+
 let access_forbidden req s =
   (* Reject external non-TLS requests (depending on config) *)
-  !Xapi_globs.website_https_only
-  && (not (Context.is_unix_socket s))
-  && Context._client_of_rq req = None
+  !Xapi_globs.website_https_only && is_external_http req s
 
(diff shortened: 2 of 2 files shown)


benjamreis (approved):

lindig (approved):

psafont:
Tested both cases when http is requested and website-https-only is:
- on: 403 is returned
- off: redirect happens

HTTPS works fine in both cases

ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -19,10 +19,10 @@ let prototyped_of_field = function
       Some "22.27.0"
   | "host", "last_software_update" ->
       Some "22.20.0"
-  | "pool", "migration_compression" ->
-      Some "22.33.0"
   | "pool", "coordinator_bias" ->
       Some "22.34.0-next"
+  | "pool", "migration_compression" ->
+      Some "22.33.0"
```
jameshensmancitrix, line 25: I have similar in other PR, should this be an issue?

jameshensmancitrix (approved):

ocaml/idl/datamodel_lifecycle.ml
psafont: rebasing should be easy, as it's the same change, from what I've seen. worst case you can ignore the change and rerun make to regenerate the file

ocaml/xapi/fileserver.ml
```diff
@@ -78,26 +78,27 @@ let response_file s file_path =
...
-    match req.Request.host with
-    | Some host ->
-        (* Redirect towards HTTPS *)
-        let path = req.Request.uri in
-        let dest = Uri.make ~scheme:"https" ~host ~path () |> Uri.to_string in
-        Http_svr.response_redirect ~req s dest
-    | None ->
-        Http_svr.response_forbidden ~req s
+  let is_external_http = is_external_http req s in
+  if is_external_http && !Xapi_globs.website_https_only then
```
edwintorok, line 94: I think we're also going to close the port completely in the firewall here in the near future, so the behaviour of xapi when https only is enabled won't matter anymore.

edwintorok (approved):

ocaml/xapi/fileserver.ml
```diff
@@ -78,26 +78,27 @@ let response_file s file_path =
   in
   Http_svr.response_file ~mime_content_type s file_path
 
+let is_external_http req s =
+  (not (Context.is_unix_socket s)) && Context._client_of_rq req = None
+
 let access_forbidden req s =
```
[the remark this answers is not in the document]
psafont: no, it's used in `ocaml/xapi/repository.ml`

PR #4893, opened 2023-01-25
CA-362358: Filter out new files when refreshing directory of certific…

snwoods:
…ates

### DIFF
--- a/ocaml/xapi/cert_distrib.ml
+++ b/ocaml/xapi/cert_distrib.ml
@@ -457,7 +457,10 @@ let ( (get_local_ca_certs : unit -> WireProtocol.certificate_file list)
      * executed on [path] *)
     Sys.readdir path
     |> Array.to_list
-    |> List.filter (Fun.flip Filename.check_suffix "pem")
+    |> List.filter (fun x ->
+           Filename.check_suffix x "pem"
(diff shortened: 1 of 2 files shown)


psafont:
There's at least one other place that needs to do a better job of filtering this directory:

https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/cert_distrib.ml#L460

lindig:
I think this looks good. I speculate the problem of suffix clashes could have also been resolved by creating new certificates under `*.pem.new` rather than `*.new.pem`.

lindig (approved):

ocaml/xapi/cert_distrib.ml
```diff
@@ -457,7 +457,10 @@ let ( (get_local_ca_certs : unit -> WireProtocol.certificate_file list)
      * executed on [path] *)
     Sys.readdir path
     |> Array.to_list
-    |> List.filter (Fun.flip Filename.check_suffix "pem")
+    |> List.filter (fun x ->
```
lindig, line 460: Minor detail: a less generic name than `x` would be better - something like `file`, `fname`, `name`, or similar. Below `filename` is used.

psafont (approved):
Once we see there are no more bugs blocking distribution of builds we can merge this

PR #4894, opened 2023-01-25
CA-374238: prevent copying of removed fields when reverting snapshots

psafont:
is_snapshot_from_vmpp cannot be loaded anymore, so do not try to copy it over.

Additionally when encountering an error when writing the field print a warning instead of failing

This fixes both vm-snapshot and cross-pool migration.

I've tested import-export just in case and it works correctly

### DIFF
--- a/ocaml/xapi/xapi_vm_snapshot.ml
+++ b/ocaml/xapi/xapi_vm_snapshot.ml
@@ -169,27 +169,23 @@ let checkpoint ~__context ~vm ~new_name =
 
 (* The following code have to run on the master as it manipulates the DB cache directly. *)
 let copy_vm_fields ~__context ~metadata ~dst ~do_not_copy ~overrides =
-  if not (Pool_role.is_master ()) then
-    raise
-      Api_errors.(
-        Server_error
-          ( internal_error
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_vm_snapshot.ml
```diff
@@ -169,27 +169,23 @@ let checkpoint ~__context ~vm ~new_name =
...
-        if List.mem_assoc key overrides then
-          List.assoc key overrides
-        else
-          value
-      in
+      let value = Option.value ~default:value (List.assoc_opt key overrides) in
       if not (List.mem key do_not_copy) then
-        DB.write_field db Db_names.vm (Ref.string_of dst) key value
+        try DB.write_field db Db_names.vm (Ref.string_of dst) key value
+        with Db_exn.DBCache_NotFound ("missing column", _, name) ->
```
edwintorok, line 184: matching exception based on string might be brittle. Perhaps one compromise would be to define a constant string in schema.ml for "missing column" use that in the raise, and then use it again here.
Or turn that string into a variant.

edwintorok (approved):
We can do the improvement I suggested later, for now this fixes the blocker.

lindig (approved):

PR #4895, opened 2023-01-26
CP-41818: Branding and copyright updates for the SDK

kc284:
Summary of changes
Licence:
- Removed BSD2 licence from Ocaml generation files as I think they should match the LGPLv2 licence of the rest of the Ocaml code in this repo (as also stated in the Readme of the sdk-gen repo).

Copyright:
- Added `<yearRange> <vendor>` to SDK sources and Ocaml generation code
- Added `<yearRange> <vendor> All rights reserved.` to Readmes and metadata files shipped with the binary deliverables.
I've used the same yearRange for both of the above cases. Practically speaking I don't think it's incorrect to do so, plus that from past experience I doubt we'll be remembering to update it every time for each individual file, while using the same makes it easier to change in one go.

### DIFF
--- a/ocaml/sdk-gen/csharp/templates/XenServer.csproj.mustache
+++ b/ocaml/sdk-gen/csharp/templates/XenServer.csproj.mustache
@@ -7,11 +7,11 @@
     <GenerateAssemblyInfo>True</GenerateAssemblyInfo>
     <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
     <Title>XenServer.NET</Title>
-    <Authors>Citrix Systems, Inc.</Authors>
+    <Authors>Cloud Software Group, Inc.</Authors>
     <PackageId>$(AssemblyName).NET</PackageId>
     <Product>$(AssemblyName).NET</Product>
     <Description>.NET wrapper for the XenServer API</Description>
-    <Copyright>Copyright (c) Citrix Systems, Inc. All rights reserved.</Copyright>
+    <Copyright>Copyright (c) 2007-2023 Cloud Software Group, Inc. All Rights Reserved.</Copyright>
     <PackageTags>citrix hypervisor virtualization sdk jsonrpc .net c# xen xenserver</PackageTags>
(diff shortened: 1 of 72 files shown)


edwintorok:
Years aren't strictly needed in copyright notices in files  [see the curl blogpost on this](https://daniel.haxx.se/blog/2023/01/08/copyright-without-years/).
I think copyright exists for 70 years after the death of the author, so the only effect updating copyright has is to push that out a few years for newly added work.

edwintorok:
I don't think the 'All Rights Reserved' is strictly needed either, according to https://en.wikipedia.org/wiki/All_rights_reserved it is obsolete since 2000 and has no legal effect.
The example BSD2 license also doesn't include it https://opensource.org/licenses/BSD-2-Clause

ocaml/sdk-gen/csharp/templates/XenServer.csproj.mustache
```diff
@@ -7,11 +7,11 @@
...
     <Title>XenServer.NET</Title>
-    <Authors>Citrix Systems, Inc.</Authors>
+    <Authors>Cloud Software Group, Inc.</Authors>
     <PackageId>$(AssemblyName).NET</PackageId>
     <Product>$(AssemblyName).NET</Product>
     <Description>.NET wrapper for the XenServer API</Description>
-    <Copyright>Copyright (c) Citrix Systems, Inc. All rights reserved.</Copyright>
+    <Copyright>Copyright (c) 2007-2023 Cloud Software Group, Inc. All Rights Reserved.</Copyright>
     <PackageTags>citrix hypervisor virtualization sdk jsonrpc .net c# xen xenserver</PackageTags>
     <PackageLicenseExpression>BSD-2-Clause</PackageLicenseExpression>
```
danilo-delbusso, line 16: Are we not updating the license of the generated SDKs to LGPLv2, too?

kc284: No need to.

danilo-delbusso (approved):

psafont (approved):

PR #4897, opened 2023-01-30
CP-33338: write physical-device-path to xenstore for nbd devices

MarkSymsCtx:
When the VDI datapath is attached with an NBD export the backend expects the details to be present in physical-device-path.

### DIFF
--- a/ocaml/xenopsd/scripts/block
+++ b/ocaml/xenopsd/scripts/block
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 TYPE=`echo ${XENBUS_PATH} | cut -f 2 -d '/'`
 DOMID=`echo ${XENBUS_PATH} | cut -f 3 -d '/'`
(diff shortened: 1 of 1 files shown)


MarkSymsCtx:
This is a nicely separable preparation change to enable being able to set up the storage backend correctly when the attach information is for an NBD unit socket.

MarkSymsCtx:
BSTs for Storage and Ring 3 are 179608 & 179609. Still in progress, will edit and update once completed.

lindig (commented):
This is using bash features but this is `#!/bin/sh`

MarkSymsCtx:
> This is using bash features but this is `#!/bin/sh`

Is there any reason why we can't make it a bash shell? Matching on the regex makes it a straightforward single operation to determine if we take the path and extracts the necessary data. Using legacy `sh` syntax will make it somewhat more fragile and less easy to read IMO with the extra checks and splits and indexing.

lindig (approved):

robhoes (approved):

PR #4899, opened 2023-01-31
dbgring: switch from Xenmmap to Gnt + Io_page

edwintorok:
Xenmmap.read is being removed from Xen, use upstream mirage libs to access this instead.

This uses OCaml's builtin Unix.map_file+Bigarray instead of the custom Xenmmap. The custom xenmmap is useful if we want precise control over when to unmap (the builtin one will unmap when OCaml value is finalized), but that is not needed here: dbgring is an application that runs once, it is not a long running daemon.


edwintorok:
Cc @andyhhp

edwintorok:
This will allow us to drop some buggy code from Xen's xenmmap (read and write are not safe to use there since they do not perform any bounds checks)

andyhhp (approved):

ocaml/xenopsd/dbgring/dbgring.ml
```diff
@@ -11,27 +11,29 @@
...
  * GNU Lesser General Public License for more details.
  *)
-open Xenops_utils
 
 let xenstored_proc_port = "/proc/xen/xsd_port"
 
 let xenstored_proc_kva = "/proc/xen/xsd_kva"
 
+let gnttab = Gnt.Gnttab.interface_open ()
+
```
lindig, line 20: Should this use `Fun.protect`?

lindig (approved):

ocaml/xenopsd/dbgring/dbgring.ml
edwintorok: There is a 'with' function I could use inside Gnttab, but I'm not sure how safe it is to close the gnt when the mapping is still in use, I'll have to test it.

edwintorok:
Fairly sure I tested this before, but maybe I built the wrong thing because it is not working now.
Unix.map_file fails because it does an fstat and gets back a 0 size and then doesn't let mmap to continue (it is a device, so kind of special that the file size is 0, but you can map it beyond its size...).
And Gnttab somehow manages to end up doing an mmap with length 0, which won't work.

edwintorok:
loks like we need xenmmap still.

PR #4900, opened 2023-01-31
CP-41812: Add total_sessions_count RRD

snwoods:

### DIFF
--- a/ocaml/xapi/xapi_session.ml
+++ b/ocaml/xapi/xapi_session.ml
@@ -23,6 +23,10 @@ open D
 
 let with_lock = Xapi_stdext_threads.Threadext.Mutex.execute
 
+let total_sessions = Atomic.make 0
+
+let get_total_sessions () = Atomic.get total_sessions |> Int64.of_int
+
 module Date = Xapi_stdext_date.Date
 module Listext = Xapi_stdext_std.Listext
 open Client
@@ -633,6 +637,8 @@ let login_no_password_common ~__context ~uname ~originator ~host ~pool
       ~other_config:[] ~subject ~is_local_superuser ~auth_user_sid
       ~validation_time:(Date.of_float (Unix.time ()))
       ~auth_user_name ~rbac_permissions ~parent ~originator ~client_certificate ;
+    if not pool then
+      Atomic.incr total_sessions ;
     Ref.string_of session_id
   in
   let session_id =
--- a/ocaml/xapi/xapi_session.mli
+++ b/ocaml/xapi/xapi_session.mli
@@ -103,3 +103,6 @@ val record_login_failure :
   -> 'a
 
 val get_failed_login_stats : unit -> string option
+
+val get_total_sessions : unit -> Int64.t
+(** Retrieves the amount of sessions opened since the last time xapi was started *)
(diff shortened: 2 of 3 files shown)


snwoods:
I've not touched any database code so I'm not sure if this is the full solution. However, the pool_total_session_count correctly appears in xencenter and the CLI with an ever-increasing value.

ocaml/xapi/xapi_session.ml
```diff
@@ -633,6 +635,7 @@ let login_no_password_common ~__context ~uname ~originator ~host ~pool
       ~other_config:[] ~subject ~is_local_superuser ~auth_user_sid
       ~validation_time:(Date.of_float (Unix.time ()))
       ~auth_user_name ~rbac_permissions ~parent ~originator ~client_certificate ;
+    Atomic.incr total_sessions ;
```
robhoes, line 638: We probably want to exclude sessions with `pool = true` here, which are used by pool members only.

ocaml/xapi/xapi_session.mli
```diff
@@ -103,3 +103,5 @@ val record_login_failure :
   -> 'a
 
 val get_failed_login_stats : unit -> string option
+
+val total_sessions : int Atomic.t
```
psafont, line 107: For isolation purposes it'd be better to expose a function that cannot modify the value

```ocaml
  val get_total_sessions : unit -> Int64.t
  (** Retrieves the amount of sessions opened since the last time xapi was started *)

  let get_total_sessions () = Atomic.get total_sessions |> Int64.of_int
```

ocaml/xapi/xapi_stats.ml
```diff
@@ -35,7 +35,18 @@ let generate_master_stats ~__context =
...
   in
-  [session_count_ds; task_count_ds]
+  let total_session_count =
+    Atomic.get Xapi_session.total_sessions |> Int64.of_int
+  in
+  let total_session_count_ds =
+    ( Rrd.Host
+    , Ds.ds_make ~name:"pool_total_session_count"
+        ~description:"Total number of sessions"
+        ~value:(Rrd.VT_Int64 total_session_count) ~ty:Rrd.Gauge ~default:true
```
psafont, line 45:
```suggestion
        ~value:(Rrd.VT_Int64 total_session_count) ~ty:Rrd.Derive ~default:true
```

We need the RRds to calculate the rate of change, gauge is time-independent so it doesn't do what we want here. Note that the number of tasks and sessions are the ones that are open at a time instead of the total amount, they are different so a rate cannot be calculated from them.

```
(** DS - a data source
    This defines how we deal with incoming data. Type is one of:
    - Absolute: meaning that the incoming data is an absolute rate
    - Derive:   meaning that the rate must come from the difference between the
                incoming data and the previous value
    - Gauge:    meaning that the value isn't a rate at all (e.g. temperature, load avg)
```
https://github.com/xapi-project/xcp-rrd/blob/master/lib/rrd.ml#L98-107

psafont (changes requested):

ocaml/xapi/xapi_stats.ml
```diff
@@ -36,13 +36,13 @@ let generate_master_stats ~__context =
     )
   in
   let total_session_count =
-    Atomic.get Xapi_session.total_sessions |> Int64.of_int
+    Xapi_session.get_total_sessions
```
psafont, line 39:
```suggestion
    Xapi_session.get_total_sessions ()
```

robhoes (approved):

Issue #4822, opened 2022-10-17
RFC: Design - new UEFI certificate management


stormi:
## Context

The requirements around UEFI certificate management differ in XenServer and XCP-ng, but we can do something that works for both.

In XenServer, certificates will now be provided through RPMs (`varstored` (PK, KEK, db) and `secureboot-certs` (dbx), which simplifies a lot the requirements. XAPI doesn't need to play a role anymore, except maybe by providing a way to read the certificates through the API. As far as I understand the requirements, users must not be able to change the certificates through XAPI (`pool.set_uefi_certificates` not supported anymore).

In XCP-ng, we will also provide default certificates through RPMs, but we also want to offer users the ability to override with their own certificates. There are use cases for this. For example, hardware vendors who have their own set of SecureBoot certificates, not Microsoft's ones. So we need to keep a way to set the certificates at the pool level and then propagate them to host disks for varstored (we will likely replace our own implementation, uefistored, with varstored) to consume.

## Proposed Design Change

A `xapi.conf` setting named `override_uefi_certs` governs XAPI's behaviour (suggested by Edwin Török)
* if false, "makes all set attempts fail. Reading the field will always tell you what is on the FS." (quoting Edwin)
* if true, XAPI keeps handling certificates, but now in a way that accounts for both default certs (those in `/usr/share/varstored`) and custom certs (those present in XAPI's db if any, then written to disk to a different place so that default certs are never overwritten).

The way we would like XAPI to behave when `override_uefi_certs` is true would likely be useful to XenServer too if later XenServer wants to offer some users the ability to override the certificates with their custom certs, without overwriting the certificates provides in the RPMs.

## How XAPI would handle certificates if `override_uefi_certs` is true

At startup and when `pool.set_uefi_certificates` is called, XAPI writes certificates to disk, to `varstored_dir` (after emptying it)... Whose new default value is now `/var/lib/varstored` (XAPI must *not* write to `/usr/share` as it did before).
* If `pool.uefi_certificates` is empty, certificates are copied from `default_auth_dir` (`/usr/share/varstored`) to `/var/lib/varstored`
* Else, the certificates stored in XAPI are extracted to `/var/lib/varstored`, and default certificates are not used.

In any case, XAPI *never* writes anything to `/usr/share/varstored`, in order not to interfere with the contents of the RPMs.

## Impact on varstored

The main impact is that, depending on the value of `override_uefi_certs`, varstored will need to read from a different directory:
[...]

PR #4901, opened 2023-02-03
private/marcusg/cp-41675

mg12:
Xapi changes for the new flag override_uefi_certs associated with the design in https://github.com/xapi-project/xen-api/issues/4822
Related to the varstored change in https://github.com/mg12/varstored/commit/3e7b2cd8417283b68f1abdea1e0d0c8cbc09dfea


psafont:
To make the format check happen, `make format` needs to be run. For easiness, you probably want to rebase the patches and run it on every commit.

The tests are failing with:
```
New schema hash ('28dd1b29878ec637f3223fe8c80f6fce') doesn't match the last known one. Please bump the
```
This changes in this schema test to contain the new hash: https://github.com/xapi-project/xen-api/blob/703479fa448a8d7141954bb6e8964d8e25c4ac2e/ocaml/idl/schematest.ml#L4

and to bump the minor version of the schema:
https://github.com/xapi-project/xen-api/blob/703479fa448a8d7141954bb6e8964d8e25c4ac2e/ocaml/idl/datamodel_common.ml#L13

psafont:
This should be reviewed by @benjamreis since he's the toolstack expert for xcp-ng. He's on PTO this week, so I expect this to be merged next week at the earliest

ocaml/xapi/xapi_host.ml
```diff
@@ -2671,44 +2671,165 @@ let with_temp_file_contents ~contents f =
...
-         * to avoid a extract override issue. *)
-        List.iter
-          (fun name ->
-            let path = Filename.concat !Xapi_globs.varstore_dir name in
-            Unixext.unlink_safe path
+let really_write_uefi_certificates_to_disk ~__context ~host:_ ~value =
+  let ( // ) = Filename.concat in
+  match value with
+  | "" ->
+      let files_of_dir path =
```
lindig, line 2678: You could use `Sys.readdir` and `Sys.is_directory`. Using more low level functions is important if you want to control the behavior around links.

ocaml/xapi/xapi_host.ml
```diff
@@ -2671,44 +2671,165 @@ let with_temp_file_contents ~contents f =
...
+      try
+        Helpers.rmtree from_path ;
+        Unix.symlink to_path from_path
+      with e ->
+        let msg = Printf.sprintf "Failed to symlink %s->%s" from_path to_path in
+        error "%s: %s" msg (Printexc.to_string e) ;
+        raise Api_errors.(Server_error (internal_error, [msg]))
+    ) ;
+    fn from_path
+  in
```
lindig, line 2758: There is a lot of repeated code around file and directory handling.

lindig (commented):
Overall I think there is too much code for the functionality it provides.

ocaml/xapi/xapi_host.ml
```diff
@@ -2671,44 +2671,165 @@ let with_temp_file_contents ~contents f =
...
+  let check_valid_uefi_certs_in path =
+    (* check expected uefi certificates are present *)
+    let rec with_uefi_certs_in_disk path =
+      if Sys.file_exists path then
+        Unix.(
+          match (lstat path).st_kind with
+          | S_DIR ->
+              Sys.readdir path |> Array.to_list
+          | S_LNK ->
+              Unix.readlink path |> with_uefi_certs_in_disk
```
psafont, line 2784: Can this lead to and infinite loop chasing links?

ocaml/idl/datamodel_pool.ml
```diff
@@ -1015,7 +1015,15 @@ let disable_repository_proxy =
 
 let set_uefi_certificates =
   call ~name:"set_uefi_certificates"
-    ~lifecycle:[(Published, "22.16.0", "")]
+    ~lifecycle:
+      [
+        (Published, "22.16.0", "")
+      ; ( Deprecated
+        , "23.3.0"
```
psafont, line 1022: This will need to be changed, 23.3.0 will be a bugfix-only release, same below

ocaml/xapi/xapi_host.ml
```diff
@@ -2671,44 +2671,165 @@ let with_temp_file_contents ~contents f =
...
+              []
+        )
+      else
+        []
+    in
+    let uefi_certs_in_disk = with_uefi_certs_in_disk path in
+    ["KEK.auth"; "db.auth"]
+    |> List.iter (fun cert ->
+           let log_of found =
+             (if found then debug else error)
```
psafont, line 2795: I think info is more appropriate than debug level, this is something that users will be interested in reading

ocaml/xapi/xapi_host.ml
```diff
@@ -2671,44 +2671,165 @@ let with_temp_file_contents ~contents f =
...
+               "check_valid_uefi_certs: %s %s in %s"
+               (if found then "found" else "missing")
+               cert path
+           in
+           log_of (List.mem cert uefi_certs_in_disk)
+       )
+  in
+  match !Xapi_globs.override_uefi_certs with
+  | false ->
+      with_valid_symlink !Xapi_globs.varstore_dir !Xapi_globs.default_auth_dir
```
psafont, line 2805: as a curiosity, you might like this way of expressing the with_ functions:

```ocaml
    let ( let@ ) f x = f x
    let@ path = with_valid_symlink !Xapi_globs.varstore_dir !Xapi_globs.default_auth_dir in
    check_valid_uefi_certs_in path
```

similarly in the other case:
```ocaml
      let@ path = with_empty_dir !Xapi_globs.varstore_dir in
      let pool_uefi_certs =
        Db.Pool.get_uefi_certificates ~__context
          ~self:(Helpers.get_pool ~__context)
      in
      really_write_uefi_certificates_to_disk ~__context ~host
        ~value:pool_uefi_certs ;
      check_valid_uefi_certs_in path
```

ocaml/xapi/xapi_host.ml
```diff
@@ -2671,44 +2671,165 @@ let with_temp_file_contents ~contents f =
...
-        ~self:(Helpers.get_pool ~__context)
-        ~value
-  )
+  match !Xapi_globs.override_uefi_certs with
+  | false ->
+      raise
+        Api_errors.(
+          Server_error
+            ( Api_errors.operation_not_allowed
+            , ["override_uefi_certs not enabled"]
```
psafont, line 2827: I would rather leave the message empty, this action might not be supported, depending on the product used. How to enable this can be instead be explained in the docs of xcp-ng since it's a special case for customers already using custom certificates, which Windows guests might not consider kosher

ocaml/xapi/xapi_host.ml
mg12: I've refactored the code to use Helpers.FileSys.* more extensively, the number of line changes reduced considerably since the last patchset.

ocaml/xapi/xapi_host.ml
mg12: thanks, I like this

ocaml/xapi/xapi_host.ml
mg12: fixed in the last patchset

ocaml/xapi/xapi_host.ml
mg12: fixed in the last patchset

ocaml/xapi/xapi_host.ml
mg12: fixed in the last patchset

ocaml/xapi/xapi_host.ml
mg12: I believe I fixed it in the last patchset in the way your expected, I wonder if you could double-check

psafont: It looks good now, thanks

ocaml/xapi/xapi_globs.ml
```diff
@@ -1388,6 +1392,11 @@ let other_options =
     , (fun () -> string_of_bool !ignore_vtpm_unimplemented)
     , "Do not raise errors on use-cases where VTPM codepaths are not finished."
     )
+  ; ( "override-uefi-certs"
+    , Arg.Set override_uefi_certs
+    , (fun () -> string_of_bool !override_uefi_certs)
+    , "Enable/Disable overriding location for varstored UEFI certificates"
+    )
```
benjamreis, line 1399: Is there a reason `varsore_dir` & `default_auth_dir` are not configurable?

psafont: Why would they be? one is defined by the rpm, and the other is an interface with varstored an xapi, neither of which make sense to allow for users to configure

benjamreis: No it's fine by me, i was just wondering ^^.

benjamreis (approved):
LGTM

mg12:
In the latest series above: (cc @kc284) 
- I removed the patch that deprecated `Pool.set_uefi_certificates`, as this api call is still useful for users when `xapi.conf:override_uefi_certificates=true`
- I kept the patch that does not allow use of `Pool.set_uefi_certificates` when `xapi.conf:override_uefi_certificates=false`
- To keep backwards compatibility of the DB field `pool.uefi_certificates` with existing vm-start behaviour such as `platform:secureboot=auto` means `platform:secureboot=true`, and XenCenter and other clients relying on this field being populated to detect if UEFI Secure Boot is available, then when `xapi.conf:override_uefi_certificates=false`, this DB field will be populated with the UEFI certificates present in disk at the moment in which xapi starts.

benjamreis (approved):

ocaml/xapi/xapi_db_upgrade.ml
```diff
@@ -896,6 +909,7 @@ let rules =
   ; upgrade_cluster_timeouts
   ; upgrade_secrets
   ; remove_legacy_ssl_support
+  ; empty_pool_uefi_certificates
```
psafont, line 912: I believe this gets executed before calling `Xapi_host.write_uefi_certificates_to_disk` at startup

psafont (approved):

ocaml/xapi/xapi_host.ml
```diff
@@ -2671,26 +2671,31 @@ let with_temp_file_contents ~contents f =
...
-    Sys.file_exists !Xapi_globs.varstore_dir
-    && Sys.is_directory !Xapi_globs.varstore_dir
-  then
-    match
-      Base64.decode
-        (Db.Pool.get_uefi_certificates ~__context
-           ~self:(Helpers.get_pool ~__context)
-        )
-    with
+let ( let@ ) f x = f x
```
edwintorok, line 2674: This is a nice trick to avoid having to write 'with_file foo @@ fun fd -> '

ocaml/xapi/helpers.ml
```diff
@@ -1955,10 +1955,25 @@ end = struct
     Xapi_psr_util.load_psr_pool_secrets ()
 end
 
+let ( let@ ) f x = f x
+
+let with_temp_fd fd f = finally (fun () -> f fd) (fun () -> Unix.close fd)
+
+let with_temp_file ?mode prefix suffix f =
+  let temp_file, out_ch = Filename.open_temp_file ?mode prefix suffix in
```
lindig, line 1963: I would pick more symmetrical names: `tmp_path`, `tmp_channel` or similar. Because everything is in the local context of a temporary file, `temp` or `tmp` could be also dropped. Likewise below.

ocaml/xapi/xapi_host.ml
```diff
@@ -2701,14 +2706,65 @@ let write_uefi_certificates_to_disk ~__context ~host:_ =
     | Error _ ->
         debug
           "UEFI tar file was not extracted: it was not base64-encoded correctly"
+  )
+
+let write_uefi_certificates_to_disk ~__context ~host =
+  let with_valid_symlink from_path to_path fn =
```
edwintorok, line 2725: Since the order of arguments here is the opposite of the order in Unix.symlink perhaps use labeled arguments here to make it obvious at the callsite which argument means what.

ocaml/xapi/xapi_pool.ml
```diff
@@ -3588,14 +3588,18 @@ let disable_repository_proxy ~__context ~self =
...
-  Helpers.call_api_functions ~__context (fun rpc session_id ->
-      List.iter
-        (fun host ->
-          Client.Host.write_uefi_certificates_to_disk ~rpc ~session_id ~host
-        )
-        (Db.Host.get_all ~__context)
-  )
+  match !Xapi_globs.override_uefi_certs with
+  | false ->
+      raise Api_errors.(Server_error (Api_errors.operation_not_allowed, [""]))
```
edwintorok, line 3593: according to datamodel_errors.ml the argument here should be a 'reason', so using something like `"Setting UEFI certificates is not possible when override_uefi_certs is false"`

ocaml/xapi/xapi_globs.ml
```diff
@@ -1388,6 +1392,11 @@ let other_options =
     , (fun () -> string_of_bool !ignore_vtpm_unimplemented)
     , "Do not raise errors on use-cases where VTPM codepaths are not finished."
     )
+  ; ( "override-uefi-certs"
+    , Arg.Set override_uefi_certs
+    , (fun () -> string_of_bool !override_uefi_certs)
+    , "Enable/Disable overriding location for varstored UEFI certificates"
```
lindig, line 1398: Should this include a hint for the expected value like `Enable (true) or disable (false) overriding ..`?

ocaml/xapi/xapi_host.ml
```diff
@@ -2671,26 +2671,44 @@ let with_temp_file_contents ~contents f =
...
-        (Db.Pool.get_uefi_certificates ~__context
-           ~self:(Helpers.get_pool ~__context)
-        )
-    with
+let ( let@ ) f x = f x
+
+let ( // ) = Filename.concat
+
+let really_read_uefi_certificates_from_disk ~__context ~host:_ from_path =
+  let src_files = Sys.readdir from_path |> Array.map (( // ) from_path) in
```
lindig, line 2679: I find `src` slightly unexpected. Are these not `certs`?

ocaml/xapi/helpers.ml
```diff
@@ -1955,6 +1955,19 @@ end = struct
...
+  finally (fun () -> f (temp_file, out_ch)) (fun () -> Unix.unlink temp_file)
+
+let with_temp_fd_of_temp_file ?mode prefix suffix f =
+  let@ temp_file, out_ch = with_temp_file ?mode prefix suffix in
+  let@ temp_out_fd = out_ch |> Unix.descr_of_out_channel |> with_temp_fd in
```
edwintorok, line 1968: This will cause a warning to be printed to stderr when Sys.enable_runtime_warnings is used:
```
utop # Sys.enable_runtime_warnings true;;
- : unit = ()
─( 10:13:35 )─< command 3 >───────────────────────────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # Filename.open_temp_file "foo" "foo" |> snd |> Unix.descr_of_out_channel |> Unix.close;;
- : unit = ()
[ocaml] (use Sys.enable_runtime_warnings to control these warnings)
[ocaml] channel opened on file '/tmp/foo7d4427foo' dies without being closed
```

We should really try to close the channel instead of the FD it contains inside.

lindig (approved):

ocaml/xapi/xapi_pool.ml
```diff
@@ -3590,7 +3590,16 @@ let disable_repository_proxy ~__context ~self =
 let set_uefi_certificates ~__context ~self ~value =
   match !Xapi_globs.override_uefi_certs with
   | false ->
-      raise Api_errors.(Server_error (Api_errors.operation_not_allowed, [""]))
+      raise
```
psafont, line 3593:
```suggestion
      let msg =
        "Setting UEFI certificates is not possible when override_uefi_certs is \
         false"
      in
      raise Api_errors.(Server_error (operation_not_allowed, [msg]))
```

ocaml/xapi/xapi_pool.ml
```diff
@@ -3590,7 +3590,16 @@ let disable_repository_proxy ~__context ~self =
...
+        Api_errors.(
+          Server_error
+            ( Api_errors.operation_not_allowed
+            , [
+                "Setting UEFI certificates is not possible when \
+                 override_uefi_certs is false"
+              ]
+            )
+        )
```
psafont, line 3602:
```suggestion
      let msg =
        "Setting UEFI certificates is not possible when override_uefi_certs is \
         false"
      in
      raise Api_errors.(Server_error (operation_not_allowed, [msg]))
```

psafont: This will need to be explained in the xenserver docs, I'm sure some customers will try to set their own certs just because they can, and support needs to be ready to tell them it's not supported

edwintorok (approved):

ocaml/xapi/xapi_host.ml
mg12: yes, done in the latest fixup

ocaml/xapi/xapi_globs.ml
mg12: done in the latest fixup

ocaml/xapi/helpers.ml
mg12: done in the latest fixup

PR #4902, opened 2023-02-06
Private/jameshen/cli sessions

jameshensmancitrix:


ocaml/idl/datamodel.ml
```diff
@@ -191,6 +222,10 @@ module Session = struct
         ; local_logout
         ; get_all_subject_identifiers
         ; logout_subject_identifier
+        ; get_all_records
+        ; count_sessions
+        ; group_sessions_by_originator
```
lindig, line 227: This is in the Sessions module, so `group_by_originator` is probably good enogh; similar for the other field.

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -7261,6 +7278,24 @@ let session_subject_identifier_logout_all _printer rpc session_id _params =
     )
     subject_identifiers
 
+let session_count printer rpc session_id _params =
+  let count = Client.Session.count_sessions ~rpc ~session_id in
+  printer (Cli_printer.PMsg (Int64.to_string count))
+
+let session_groupby_originator printer rpc session_id _params =
+  let mapping = Client.Session.group_sessions_by_originator ~rpc ~session_id in
```
lindig, line 7286: Is it possible to sort the output? Is there a default sort (I assume by key) but in this case we might be interested in the highest numbers first. This is just an idea - if it's easy.

ocaml/xapi/xapi_session.ml
```diff
@@ -1337,3 +1349,66 @@ let create_from_db_file ~__context ~filename =
...
+
+let count_sessions ~__context =
+  Db.Session.get_all ~__context |> List.length |> Int64.of_int
+
+let group_sessions_by_originator ~__context =
+  let one = Int64.of_int 1 in
+  Db.Session.get_all ~__context
+  |> List.map (fun self -> Db.Session.get_originator ~__context ~self)
+  |> List.fold_left
+       (fun acc orig ->
```
lindig, line 1391: This looks inefficient - the repeated use of `List.mem`; it might pay off to use a different data structure. On the other hand, we don't expect this is be used frequently. It would be more efficient to sort the entries by originator - now they can be easily grouped in O(n).

ocaml/xapi-cli-server/cli_operations.ml
jameshensmancitrix: No order, I can add this in, per the highest count?

ocaml/xapi/xapi_session.ml
jameshensmancitrix:
```
let group_by_originator ~__context =
  let one = Int64.of_int 1 in
  Db.Session.get_all ~__context
  |> List.map (fun self -> Db.Session.get_originator ~__context ~self)
  |> List.sort (fun orig1 orig2 -> if orig1 > orig2 then 1 else 0)
  |> List.fold_left
       (fun (acc : (string * Int64.t) list) orig -> 
         match acc with
         | []  -> (orig, one) :: acc
         | (o,_)::_ when o <> orig -> (orig, one) :: acc
         | (_,value)::_ -> (
           let acc' = List.tl acc in
           (orig, (Int64.add value one)) :: acc')
       )
       []
  |> List.sort (fun (_,v1) (_,v2) -> Int64.sub v1 v2 |> Int64.to_int)
```

lindig: For clarity, I would write (or find) a function that partitions a sorted (key, value) list into (key, count) similar to what on the command line `uniq -c` does and use it.

ocaml/xapi/xapi_session.ml
```diff
@@ -1337,3 +1349,73 @@ let create_from_db_file ~__context ~filename =
...
+  |> List.map (fun session -> (session, get_api_record ~self:session))
+
+let count_sessions ~__context =
+  Db.Session.get_all ~__context |> List.length |> Int64.of_int
+
+let group_by_originator ~__context =
+  let one = Int64.of_int 1 in
+  Db.Session.get_all ~__context
+  |> List.map (fun self -> Db.Session.get_originator ~__context ~self)
+  |> List.sort (fun orig1 orig2 -> if orig1 > orig2 then 1 else 0)
```
lindig, line 1390: This compare function never returns -1, as it probably should. You can probably use `List.sort String.compare`

ocaml/xapi/xapi_session.ml
```diff
@@ -1337,3 +1349,73 @@ let create_from_db_file ~__context ~filename =
...
+         | [] ->
+             (orig, one) :: acc
+         | (o, _) :: _ when o <> orig ->
+             (orig, one) :: acc
+         | (_, value) :: _ ->
+             let acc' = List.tl acc in
+             (orig, Int64.add value one) :: acc'
+       )
+       []
+  |> List.sort (fun (_, v1) (_, v2) -> Int64.sub v1 v2 |> Int64.to_int)
```
lindig, line 1403: `Int64.compare`

ocaml/xapi/xapi_session.ml
jameshensmancitrix: That would work well

jameshensmancitrix:
due to rebasing issues and the branch getting a little mangled between features I have created one commit with the message describing everything going on.

ocaml/idl/datamodel.ml
```diff
@@ -174,6 +174,35 @@ module Session = struct
       ~in_product_since:rel_george ~in_oss_since:None ~allowed_roles:_R_POOL_OP
       ()
 
+  let get_all_records =
+    call ~name:"get_all_records" ~lifecycle:[] ~params:[] ~flags:[`Session]
+      ~result:
+        ( Map (Ref _session, Record _session)
```
robhoes, line 180: This is still a problem: you can't have a `get_all_records` function for sessions for security reasons.

ocaml/xapi/xapi_session.ml
```diff
@@ -1337,3 +1349,73 @@ let create_from_db_file ~__context ~filename =
...
+    ; API.session_this_user= Db.Session.get_this_user ~__context ~self
+    ; API.session_uuid= Db.Session.get_uuid ~__context ~self
+    ; API.session_validation_time=
+        Db.Session.get_validation_time ~__context ~self
+    ; API.session_creation_ip= Db.Session.get_creation_ip ~__context ~self
+    ; API.session_last_login_ip= Db.Session.get_last_login_ip ~__context ~self
+    }
+  in
+  Db.Session.get_all ~__context
+  |> List.map (fun session -> (session, get_api_record ~self:session))
```
robhoes, line 1381: You are returning the session reference here for all sessions. This is security sensitive data.

ocaml/xapi-cli-server/records.ml
```diff
@@ -1627,6 +1627,98 @@ let subject_record rpc session_id subject =
       ]
   }
 
+let session_record rpc session_id session =
```
robhoes, line 1630: I think we don't need this at all. It doesn't look like it is even used? We shouldn't have `xe session-list` or similar.

jameshensmancitrix: This is needed for autogeneration of the CLI commands *-list *-param-get

ocaml/xapi/xapi_session.ml
jameshensmancitrix: I am not returning the actual session reference, I am creating a new reference and returning that This will leak a random ref.

robhoes: `Db.Session.get_all ~__context` returns a list of session references. You map this a tuple of a session reference with a session record. This is being returned directly to the API client, right here.

jameshensmancitrix: i thought I had changed that, my mistake

lindig:
Could you summarize the discussion? I am not sure if we should continue to review this or if we are waiting for updates.

PR #4905, opened 2023-02-09
CP-42173: xenctrlext: stop using xentoolog bindings

psafont:
These are known to be problematic and cause segfaults on shutdown, the call can be done by passing around NULL to let xen create a handle that can be ignored.

This needs testing with a special build of xen without the bindings from @andyhhp


andyhhp:
I don't see any need to change Xen for this.  In fact, I kinda need this merged before I pull in the latest staging from Xen which deletes the Xentoollog binding

andyhhp (approved):

lindig (approved):

edwintorok (approved):

PR #4906, opened 2023-02-10
CP-40650 Remove vtpm feature restriction

lindig:
So far creation of vtpms is gated by the feature flag check. This needs to be removed before it goes to general access so customers can have access to the feature without friction.

Feature flags work like this:

* v6 provides a script to create a file: scripts/toggle-experimental-feature.sh
* this script creates an entry in /etc/xenserver/features.d
* this directory is read by v6
* the features/restrictions are read by xapi (from v6) and turned into an association list pool.restrictions
* this association list is checked by the code

The change keeps the core of the code but simply reports the state of the feature flag. We might want to remove the code that sets it. It is currently not set from within xapi.

### DIFF
--- a/ocaml/xapi/xapi_vtpm.ml
+++ b/ocaml/xapi/xapi_vtpm.ml
@@ -12,16 +12,24 @@
    GNU Lesser General Public License for more details.
  *)
 
-(** Don't allow unless VTPM is enabled as an experimental feature *)
-let assert_not_restricted ~__context =
+(** Report the state of the feature flag but ignore any restrictions.
+    CP-40650: we remove any restriction to the feature *)
+
+module D = Debug.Make (struct let name = __MODULE__ end)
+
+open D
+
+let log_feature_flag ~__context =
   let pool = Helpers.get_pool ~__context in
   let restrictions = Db.Pool.get_restrictions ~__context ~self:pool in
   let feature = "restrict_vtpm" in
   match List.assoc_opt feature restrictions with
   | Some "false" ->
-      ()
-  | _ ->
-      raise Api_errors.(Server_error (feature_restricted, [feature]))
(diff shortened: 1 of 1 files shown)


psafont:
The experimental feature is set as disabledin the packaging metadata, in xapi.spec:

```ocaml
commit 2b75d6daac99331777ae589942132b9fce98562e
Author: Pau Ruiz Safont <pau.safont@citrix.com>
Date:   2022-08-26 15:17:12 +0100

    CP-40444: Prepare to enable VTPM
    
    Adds dependency on swtpm, installs swtpm-wrapper, disables vtpm feature
    by default.
[...]
@@ -423,6 +425,8 @@ mkdir $RPM_BUILD_ROOT/etc/xapi.conf.d
 mkdir $RPM_BUILD_ROOT/etc/xcp
 
 mkdir -p %{buildroot}/etc/xenserver/features.d
+echo 0 > %{buildroot}/etc/xenserver/features.d/vtpm
+
 mkdir -p %{buildroot}%{_sbindir}
 mkdir -p %{buildroot}%{_tmpfilesdir}
 %{__install} -D -m 0644 %{SOURCE1} %{buildroot}%{_unitdir}/xcp-rrdd.service
```
Maybe it could be set as enabled by default there instead

psafont (approved):

lindig:
Indeed, that would be even simpler. Not sure what what to prefer. Long term I expect the feature flag to disappear; just change xapi.spec for now?

rosslagerwall:
I think just changing the 0 to 1 in xapi.spec will be simplest for now.

psafont:
I think both changes complement each other, they can both go in and we can remove them at the same time when the feature is being cleaned up

lindig:
I think we can shelve this diff for later and change xapi.spec. Let me do this.

PR #4908, opened 2023-02-13
CP-42182 Add CLI tool on xe and rrd-cli to enable saving of rrds to local host

jameshensmancitrix:


ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
```diff
@@ -162,6 +162,10 @@ let backup_rrds (remote_address : string option) () : unit =
     )
   done
 
+let save_rrds =
+  info "%s: trying to save RRDs to localhost" __FUNCTION__ ;
+  backup_rrds None
```
robhoes, line 167: Do we actually need a new function. It looks like `backup_rrds` is there and already sufficient?

ocaml/xapi-idl/rrd/rrd_interface.ml
```diff
@@ -181,12 +181,12 @@ module RPC_API (R : RPC) = struct
 
   let has_vm_rrd =
     let bool_p = Param.mk Types.bool in
-    declare "has_vm_rrd"
+    declare "has-vm-rrd"
```
robhoes, line 184: Hmm... does this actually work? I think these should match the functions in the OCaml interface.

ocaml/idl/datamodel_host.ml
```diff
@@ -1745,6 +1745,13 @@ let set_https_only =
       ]
     ~allowed_roles:_R_POOL_OP ()
 
+let save_rrds =
```
robhoes, line 1748: Why add this to the high-level API and xe CLI if we are just going to call the rrd-cli from the bugtool?

ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
jameshensmancitrix: backup_rrds normally takes in a remote (the pool master), In this case we are passing in None to trigger a local save

robhoes: Yes, `backup_rrds` in the idl can also take `None`. The CLI can call it with `None`.

robhoes:
It would be good to update this (so that we don't lose it) to have just the `rrd-cli` and bits needs for it, plus the bugtool plugin.

jameshensmancitrix:
I had some issues when minimising the PR, then I largely put focus to the new PI 18 work, I could get back on this on Friday and try to get it working. This would also involve modifications to the spec and bugtool itself. So it would be 3 PRs to 3 repos.

lindig (approved):

robhoes (approved):

PR #4912, opened 2023-02-16
OCaml C stub bugfixes

edwintorok:
Fixes a race condition introduced in 40b755b6d, but also some other long standing bugs (wrong number of arguments to C stubs, incorrect O_DIRECT setting).
It also prepares the C stubs for OCaml 5.0 compatibility by removing some global state (although I don't claim this to be fully OCaml 5.0 ready yet).
There are some '[maintenance]' commits as well to make debugging better by preserving errno in a Unix_error instead of losing it with 'failwith', and better editor integration for C stubs (write out a compile_flags.txt with paths to headers/etc. so that `clang` can act as a language server, just like ocaml-lsp would for OCaml file).

There is also a static analyzer to check that these particular race conditions got fixed everywhere, but that depends on a newer version of GCC, and will come in a separate PR.

### DIFF
--- a/ocaml/xenopsd/c_stubs/tuntap_stubs.c
+++ b/ocaml/xenopsd/c_stubs/tuntap_stubs.c
@@ -1,5 +1,5 @@
 /*
- * (c) Citrix 
+ * (c) Citrix
  *
  * This could be replaced by https://github.com/mirage/ocaml-tuntap
  * if more features are required.
@@ -11,12 +11,14 @@
 #include <sys/ioctl.h>
 #include <linux/if_tun.h>
 #include <fcntl.h>
+#include <errno.h>
 
-
+#include <caml/alloc.h>
 #include <caml/mlvalues.h>
 #include <caml/fail.h>
 #include <caml/callback.h>
 #include <caml/memory.h>
+#include <caml/unixsupport.h>
 
 #define PATH_NET_TUN "/dev/net/tun"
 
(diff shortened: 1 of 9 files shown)


psafont (approved):

ocaml/xenopsd/c_stubs/tuntap_stubs.c
```diff
@@ -41,20 +44,23 @@ CAMLprim value stub_tap_open(value ocaml_ifname)
     }
     strncpy(ifr.ifr_name, ifname, IFNAMSIZ);
 
+    path_net_tun = caml_copy_string(PATH_NET_TUN);
     int fd = open(PATH_NET_TUN, O_RDWR);
     if (fd < 0) {
-        caml_failwith("open(" PATH_NET_TUN ") failed in " __FILE__);
+        uerror("open", path_net_tun);
```
lindig, line 50: If you want to preserve the file, you could use string concatenation of static strings in GCC: `"open failed in " __FILE__` Is a single string.

lindig (approved):

ocaml/xenopsd/c_stubs/tuntap_stubs.c
edwintorok: According to https://v2.ocaml.org/api/Unix.html the string there should be the function name,
I could do something like `"open at " __FILE__`, which would print something like this when converting the exception to string:
`"open at ..." failed on "...file...": error description`

Although in this case '/dev/net/tun' is a unique string in the code, and I hope the stacktrace will also provide enough context where the error is raised at. Previously the filename was not printed at all so __FILE__ was essential with locating where this comes from, for now lets try to follow the usual conventions of Unix.error.

PR #4914, opened 2023-02-16
add experimental PVH support

edwintorok:
This enables testing and developing PVH support in OVMF, XTF and mirage unikernels.

Tested an earlier version of this with a Mirage Solo5 unikernel (it might be useful for writing xenstore fuzzers/tests too).

/cc @andyhhp

### DIFF
--- a/ocaml/xapi-idl/xen/xenops_types.ml
+++ b/ocaml/xapi-idl/xen/xenops_types.ml
@@ -129,7 +129,11 @@ module Vm = struct
   }
   [@@deriving rpcty, sexp]
 
-  type builder_info = HVM of hvm_info | PV of pv_info | PVinPVH of pv_info
+  type builder_info =
+    | HVM of hvm_info
+    | PV of pv_info
+    | PVinPVH of pv_info
+    | PVH of pv_info
   [@@deriving rpcty, sexp]
 
   type id = string [@@deriving rpcty, sexp]
@@ -179,6 +183,7 @@ module Vm = struct
     | Domain_HVM
     | Domain_PV
     | Domain_PVinPVH
+    | Domain_PVH
     | Domain_undefined
   [@@deriving rpcty, sexp]
 
--- a/ocaml/xenopsd/lib/xenopsd.ml
+++ b/ocaml/xenopsd/lib/xenopsd.ml
@@ -59,6 +59,8 @@ let feature_flags_path = ref "/etc/xenserver/features.d"
 
 let pvinpvh_xen_cmdline = ref "pv-shim console=xen"
 
+let pvh_ovmf_cmdline = ref ""
+
 let numa_placement = ref false
 
 (* This is for debugging only *)
@@ -241,6 +243,11 @@ let options =
     , Arg.Set_string pvinpvh_xen_cmdline
     , (fun () -> !pvinpvh_xen_cmdline)
     , "Command line for the inner-xen for PV-in-PVH guests"
+    )
+  ; ( "pvh-ovmf-cmdline"
+    , Arg.Set_string pvh_ovmf_cmdline
+    , (fun () -> !pvh_ovmf_cmdline)
+    , "Command line for OVMF for PVH guests"
     )
   ; ( "numa-placement"
     , Arg.Bool (fun x -> numa_placement := x)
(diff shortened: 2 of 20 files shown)


ocaml/xapi-idl/xen/xenops_types.ml
```diff
@@ -129,7 +129,11 @@ module Vm = struct
   }
   [@@deriving rpcty, sexp]
 
-  type builder_info = HVM of hvm_info | PV of pv_info | PVinPVH of pv_info
+  type builder_info =
+    | HVM of hvm_info
+    | PV of pv_info
+    | PVinPVH of pv_info
+    | PVH of pv_info
   [@@deriving rpcty, sexp]
```
andyhhp, line 137: At some point we're going to have to have a conversation about how Xapi's data model ought to change...

ocaml/xenopsd/lib/xenopsd.ml
```diff
@@ -242,6 +244,11 @@ let options =
     , (fun () -> !pvinpvh_xen_cmdline)
     , "Command line for the inner-xen for PV-in-PVH guests"
     )
+  ; ( "pvh-ovmf-cmdline"
+    , Arg.Set_string pvh_ovmf_cmdline
+    , (fun () -> !pvh_ovmf_cmdline)
+    , "Command line for OVMF for PVH guests"
+    )
```
andyhhp, line 251: This is is probably ok for a prototype, but we're going to need to improve the data model in due course

edwintorok: This is currently completely unused, so I might as well drop it. I think for pvinpvh this is needed because there is also a nested Xen and you may want to control the args to that independently of the args for the guest kernel, but for PVH the args to the guest kernel are controllable directly and there is no nested Xen.

ocaml/xapi-idl/xen/xenops_types.ml
edwintorok: I suspect pv_info is not entirely accurate here now and we'd want a mix of hvm_info and pv_info?

andyhhp: I don't think a suitable data model would be expressed in those terms to begin with

edwintorok: maybe it'd be better to express this in terms of devices or components and their configurations. There are some implicit assumptions leaking through here, e.g. 'tpm' is only present in hvm_info because it is currently implemented by qemu, and with no qemu there is no tpm for pvh. If we'd split the bits that are specific to configuring Xen from the bits specific to configuring QEMU, and the bits from configuring devices that are usable by either pv/pvh/hvm that might be better.

One possibility here is to use phantom types, e.g. 'tpm' could be an `'Swtpm_qemu` device because that is how it is currently implemented, and HVM could accept [Swtpm_qemu | Qemu | HVM | Common] configs). Similarly 'acpi' could be an 'HVM' config, so it won't get accepted by PV, but get accepted by PVH (probably?).
And for pvinpvh we need to differentiate arguments given to nested Xen too.

lindig (approved):

lindig:
This needs a squash of the fixup before merge @edwintorok

lindig:
@robhoes @psafont can we review and merge this?

psafont (approved):
I suspect this needs a new schema hash before being merged

ocaml/xenopsd/xc/xenops_server_xen.ml
```diff
@@ -1162,7 +1162,7 @@ module VM = struct
       match persistent.ty with
       | Some (PV _) ->
           Memory.Linux.overhead_mib
-      | Some (PVinPVH _) ->
+      | Some (PVinPVH _ | PVH _ (* TODO: pvh *)) ->
```
robhoes, line 1165: This is not consistent with the change in memory_check.ml, where the HVM overhead is used.

robhoes:
I think this is fine apart from one little inconsistency, and the need for a squash.

psafont:
The DCO bot is not well:
> Commit sha: [1f1bf9d](https://github.com/xapi-project/xen-api/pull/4914/commits/1f1bf9d0100128b552ab60405807b630c9174082), Author: Edwin Török, Committer: Pau Ruiz Safont; Expected "Edwin Török [edvin.torok@citrix.com](mailto:edvin.torok@citrix.com)", but got "Edwin Török [edwin.torok@cloud.com](mailto:edwin.torok@cloud.com)".

This not happening on other PRs where the new mail is being used by Edwin, I'm treating it as glitch and overriding it.

PR #4915, opened 2023-02-16
CP-48840: Static analyzer for OCaml C stub race conditions

edwintorok:
This can find the bug we introduced in December, but it cannot yet find the bug we would have with Abstract_tag and storing a pointer to an OCaml value to a local var (it only tracks variables as OCaml values when their types start out as 'value').

This will depend on a new xs-opam.

TODO: also needs some more unit tests for the bugs it can find, there are already some 'cram' tests but more would be useful.


psafont:
xs-opam now has goblint, this could be rebased and see how it works

dune
```diff
@@ -13,3 +13,94 @@
...
+    )
+   )
+  (action
+    (with-stdout-to primitives.model.c
+      (run ocaml/staticanalyzer/genmain/lintcstubs_genmain.exe %{cmtfiles})
+     )
+   )
+)
+
+; once we upgrade to Dune 3.x we can use glob_files_rec here
```
psafont, line 66: dune 3.x is now available

edwintorok:
There is a newer analyzer at https://github.com/edwintorok/lintcstubs-arity and https://github.com/edwintorok/lintcstubs that need a few tweaks to be more easily integrated into a CI (e.g. auto-generate some of the dunes/makefiles that are used to analyze the stubs).
Closing for now.

PR #4916, opened 2023-02-16
CA-391656: OCaml C stubs: release runtime lock around all xenctrl calls

edwintorok:
There was a commit in the past that added enter/leave around all stubs, however since then more stubs got added, and not all of them released the runtime.

This is opened as draft, since it is a riskier change that needs a bit more stress testing too (not just a BST), and potentially also an improved static analyzer.

This PR builds on top of [the C stub bugfix PR ](https://github.com/xapi-project/xen-api/pull/4912)

Review only these 3 commits here:
```
854574107858b4443b5144d8e13de8bb8e2b334b (HEAD -> private/edvint/toolstack/improve-enter-leave, edwintorok/private/edvint/toolstack/improve-enter-leave) C stubs: add a .clang-format
6b386b8c6dc87d799bdbc23377042f1f9f649e33 CA-375277: unixpwd_stubs.c: factor out common code and use enter/leave blocking section where possible
a7cf4ff74db9a49a1aed670fb776e5c89ec00e38 CA-375277: xenctrlext_stubs.c: add missing enter/leave blocking section
```

### DIFF
--- a/unixpwd/c/unixpwd_stubs.c
+++ b/unixpwd/c/unixpwd_stubs.c
@@ -20,102 +20,88 @@
 #include <caml/fail.h>
 #include <caml/callback.h>
 #include <caml/memory.h>
+#include <caml/threads.h>
+#include <caml/unixsupport.h>
 
 #include "unixpwd.h"
 
-
-CAMLprim        value
-caml_unixpwd_getpwd(value caml_user)
+static CAMLprim value caml_unixpwd_get_(value caml_user, const char *fname, char*(*f)(const char*))
 {
     CAMLparam1(caml_user);
-    const char     *user;
-    char           *passwd;
+    char     *user;
+    char     *passwd;
+    int       saved_errno;
     CAMLlocal1(pw);
 
-    user = String_val(caml_user);
-    passwd = unixpwd_getpwd(user);
-    if (passwd == NULL && errno != 0)
-        caml_failwith(strerror(errno));
-    if (passwd == NULL)
-        caml_failwith("unspecified error in caml_unixpwd_getpwd()");
+    user = caml_stat_strdup(String_val(caml_user));
+    caml_enter_blocking_section();
+    errno = 0;
+    passwd = f(user);
+    saved_errno = errno;
+    caml_stat_free(user); user = NULL;
+    caml_leave_blocking_section();
+    errno = saved_errno;
+
+    if (passwd == NULL) /* errno of 0 will be mapped to `EUNKNOWNERR of 0` */
+        uerror(fname, caml_user);
 
     pw = caml_copy_string(passwd);
     free(passwd);
     CAMLreturn(pw);
+}
+
+CAMLprim        value
+caml_unixpwd_getpwd(value caml_user)
+{
+    return caml_unixpwd_get_(caml_user, "unixpwd_getpwd", unixpwd_getpwd);
 }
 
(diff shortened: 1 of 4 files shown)


edwintorok:
This PR has been open for more than a year with 0 comments.

edwintorok:
I should fix the conflicts here, and run the static analyzer. I'll keep the PR open meanwhile though.

unixpwd/c/unixpwd_stubs.c
```diff
@@ -20,104 +20,90 @@
...
-    password = caml_stat_strdup(String_val(caml_password));
-    rc = unixpwd_setspw(user, password);
-    caml_stat_free(password);
-    if (rc != 0)
-        caml_failwith(strerror(rc));
-    CAMLreturn(Val_unit);
+CAMLprim        value
+caml_unixpwd_setspw(value caml_user, value caml_password)
+{
+    return caml_unixpwd_set_(caml_user, caml_password, "unix_setpwd",
```
robhoes, line 105: String should be `"unixpwd_setspw"`

unixpwd/c/unixpwd_stubs.c
```diff
@@ -20,104 +20,90 @@
...
 
-    user = String_val(caml_user);
+    user = caml_stat_strdup(String_val(caml_user));
     password = caml_stat_strdup(String_val(caml_password));
-    rc = unixpwd_setpwd(user, password);
+    caml_enter_blocking_section();
+    errno = 0;
+    rc = f(user, password);
+    saved_errno = errno;
+    caml_stat_free(user);
```
lindig, line 85: Again, it appears to me that memory ownership is subtle.

lindig (approved):

robhoes (changes requested):
This needs to be updated, as above.

unixpwd/c/unixpwd_stubs.c
edwintorok: Lets keep the `caml_stat*` functions in the section where the runtime lock is held for consistency with the allocation function (where the runtime lock is held).

edwintorok:
Changed the base to `master` because `feature/perf` doesn't build now due to the change in `xs-opam`.
Will change the target back to `feature/perf`, once the feature branch has been updated in https://github.com/xapi-project/xen-api/pull/5913

lindig:
According to the OCaml manual the `*blocking_section` names are superseded by aquire/release, which I find more meaningful. So I would suggest to use those. 

```
  caml_acquire_runtime_system();
  /* Resolve OCaml function vfun to be invoked */
  /* Build OCaml argument varg to the callback */
  vres = callback(vfun, varg);
  /* Copy relevant parts of result vres to C data structures */
  caml_release_runtime_system();

Note: the acquire and release functions described above were introduced in OCaml 3.12. Older code uses the following historical names, declared in <caml/signals.h>:

    caml_enter_blocking_section as an alias for caml_release_runtime_system
    caml_leave_blocking_section as an alias for caml_acquire_runtime_system 
```

lindig:
I stand by my suggestion to use the modern release/acquire calls. I don't find the a "blocking section" to be a meaningful concept when calling C functions.

psafont:
This has been stalled for 10 months already. I think it's best if we close this for the time being, and make a note. Once somebody has time, they can pick this up and open a PR

PR #4917, opened 2023-02-17
python/XenAPI: Replace import six.moves with stdlib imports

bernhardkaindl:
Replace the use of six.moves with directly importing the standard xmlrpc and http libs of Python 2.7 and Python 3.x

Tested by using xenserver-status-report/xen-bugtool with it using on XS 8.3.0 with Python 2.7 and Python 3.6 installed.

Consequently, the Requires: python-six and python3-six can be removed from the xen-api.spec file as well.

### DIFF
--- a/ocaml/tests/test_data/repository_pkg_of_fullname_all
+++ b/ocaml/tests/test_data/repository_pkg_of_fullname_all
@@ -266,7 +266,6 @@ libjpeg-turbo-1.2.90-5.el7.x86_64
 ethtool-4.19-1.xs8.x86_64
 xxhash-libs-0.6.5-1.el7.x86_64
 libcollection-0.7.0-29.el7.x86_64
-python-six-1.9.0-2.el7.noarch
 libarchive-3.1.2-10.el7_2.x86_64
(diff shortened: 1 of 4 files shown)


psafont (approved):

psafont:
Looks good, because xapi is a linux foundation projects, all the commits need to be signed-off.

In this case this can be done with `git amend --sign-off`, then force-pushing

robhoes:
The CI doesn't have it:

    ImportError: No module named xmlrpc.client

bernhardkaindl:
The OpenStack variant of XenAPI https://opendev.org/x/os-xenapi/src/branch/master/os_xenapi/client/XenAPI.py#L20 uses this:
```py
if sys.version_info[0] == 2:
    import httplib as httpclient
    import xmlrpclib as xmlrpcclient
else:
    import http.client as httpclient
    import xmlrpc.client as xmlrpcclient
```
Force-pushed, lets see if this is ok.

robhoes (approved):

psafont (approved):

psafont (approved):

PR #4918, opened 2023-02-17
CA-375256: Fix storage initialisation on xapi startup

robhoes:
The top-level SMAPIv2 server in xapi uses a mux to forward calls to
lower level SMAPI servers depending on the SR. This relies on having
each SR registered with the mux as soon as the related PBD is plugged on
a host. These registrations are held in memory and therefore need to be
refreshed every time xapi starts.

Previously, the mux had a wrapper on top of it, which handled some of
the SMAPIv2 calls itself, without forwarding invoking the mux at all. An
example is the SR.list call, which returns the list of currently
attached SRs on a host. This used state from a file in dom0 and
therefore worked before registering SRs with the mux.

Unfortunately, `SR.list` is used a few times in xapi's startup sequence
before mux registrations are complete. Now that the wrapper has been
removed, the function always returns an empty list until the mux is
fully set up. This breaks several bits of initialisation at dbsync time,
in particular in the `resynchronise_pbds` function in storage_access.ml.

This fix ensures that all SRs are registered with the mux before calling
functions on the top-level SMAPIv2 server.

### DIFF
--- a/ocaml/xapi/storage_smapiv1_wrapper.ml
+++ b/ocaml/xapi/storage_smapiv1_wrapper.ml
@@ -82,7 +82,7 @@ let s_of_vm = Storage_interface.Vm.string_of
 
 let vm_of_s = Storage_interface.Vm.of_string
 
-module D = Debug.Make (struct let name = "storage_impl" end)
+module D = Debug.Make (struct let name = __MODULE__ end)
 
 open D
 
--- a/ocaml/xapi/storage_access.ml
+++ b/ocaml/xapi/storage_access.ml
@@ -709,7 +709,7 @@ let refresh_local_vdi_activations ~__context =
     let i_locked_it = List.mem localhost hosts in
     let all = List.fold_left ( && ) true in
     let someone_leaked_it =
-      all (List.map (fun h -> not (List.mem h hosts)) all_hosts)
+      hosts <> [] && all (List.map (fun h -> not (List.mem h hosts)) all_hosts)
     in
     if i_locked_it || someone_leaked_it then (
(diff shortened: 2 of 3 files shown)


ocaml/xapi/storage_smapiv1_wrapper.ml
```diff
@@ -82,7 +82,7 @@ let s_of_vm = Storage_interface.Vm.string_of
 
 let vm_of_s = Storage_interface.Vm.of_string
 
-module D = Debug.Make (struct let name = "storage_impl" end)
+module D = Debug.Make (struct let name = "storage_smapiv1_wrapper" end)
```
psafont, line 85:
```suggestion
module D = Debug.Make (struct let name = __MODULE__ end)
```

psafont (approved):
This solved the issue with resynchronize_pbds, have you checked if there are other functions where list is used before initialization of the muxes?

robhoes:
Yes, `resynchronize_pbds` is always the first thing that runs, as part of dbsync, immediately after the DB is up.

ocaml/xapi/storage_smapiv1_wrapper.ml
robhoes: Ah, small difference here: `__MODULE__` capitalises the name, so it's Storage_smapiv1_wrapper in the logs. Not a big deal if you know what to look for...

edwintorok (approved):
Looks good. The driver domain support for SRs was quite brittle even previously (just because the VM has started it doesn't mean it has also started its SMAPIv2/v3 server yet, and you will get some errors meanwhile), but we can improve that later I don't think this makes it worse.

PR #4920, opened 2023-02-22
CP-42014: Add last_update_sync to pool datamodel

gangj:
Add a new field "last_update_sync" of type DateTime to record when the last
update synchronization occurred.

### DIFF
--- a/ocaml/idl/datamodel_lifecycle.ml
+++ b/ocaml/idl/datamodel_lifecycle.ml
@@ -21,6 +21,8 @@ let prototyped_of_field = function
       Some "22.20.0"
   | "VM", "actions__after_softreboot" ->
       Some "23.1.0"
+  | "pool", "last_update_sync" ->
+      Some "23.3.0-next"
   | "pool", "coordinator_bias" ->
       Some "22.37.0"
   | "pool", "migration_compression" ->
--- a/ocaml/idl/datamodel_pool.ml
+++ b/ocaml/idl/datamodel_pool.ml
@@ -1331,6 +1331,9 @@ let t =
             "coordinator_bias"
             "true if bias against pool master when scheduling vms is enabled, \
              false otherwise"
+        ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:DateTime
+            ~default_value:(Some (VDateTime Date.epoch)) "last_update_sync"
+            "time of the last update sychronization"
         ]
       )
     ()
--- a/ocaml/xapi/xapi_pool.ml
+++ b/ocaml/xapi/xapi_pool.ml
@@ -3397,6 +3397,7 @@ let sync_updates ~__context ~self ~force ~token ~token_id =
    * But the 'configure_repositories' and 'sync_updates' may do the change.
    *)
   let enabled = Repository_helpers.get_enabled_repositories ~__context in
+  let pool = Helpers.get_pool ~__context in
   match force with
   | true ->
       Xapi_pool_helpers.with_pool_operation ~__context ~self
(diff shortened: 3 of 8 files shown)


ocaml/idl/datamodel_pool.ml
```diff
@@ -1331,6 +1331,9 @@ let t =
             "coordinator_bias"
             "true if bias against pool master when scheduling vms is enabled, \
              false otherwise"
+        ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:DateTime
```
lindig, line 1334: Is this best practice? We have an optional parameter and still use special value as default rather than `None`

ocaml/idl/datamodel_pool.ml
```diff
@@ -1331,6 +1331,9 @@ let t =
...
+            ~default_value:(Some (VDateTime (Date.of_float 0.)))
```
lindig, line 1335: This should use `Date.never`

psafont: `never` is deprecated, `Date.epoch` should be used instead

ocaml/idl/datamodel_pool.ml
gangj: Hi @lindig , sorry, do you suggest to use `None` as default value for `last_update_sync`?
Now I change to use `Date.epoch` for its default value, does this resolve this issue?

ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -21,6 +21,8 @@ let prototyped_of_field = function
       Some "22.20.0"
   | "VM", "actions__after_softreboot" ->
       Some "23.1.0"
+  | "pool", "last_update_sync" ->
+      Some "23.3.0-next"
```
psafont, line 25:
```
      Some "23.5.0-next"
```

ocaml/xapi/xapi_pool.ml
```diff
@@ -3397,6 +3397,7 @@ let sync_updates ~__context ~self ~force ~token ~token_id =
    * But the 'configure_repositories' and 'sync_updates' may do the change.
    *)
   let enabled = Repository_helpers.get_enabled_repositories ~__context in
+  let pool = Helpers.get_pool ~__context in
```
minglumlu, line 3400: The `self` is just the pool.

ocaml/idl/datamodel_lifecycle.ml
minglumlu: These new fields are to be added into feature branch firstly. Maybe the lifecycle of new added fields could be added just before merging to master?

ocaml/xapi/xapi_pool.ml
```diff
@@ -3409,10 +3410,12 @@ let sync_updates ~__context ~self ~force ~token ~token_id =
              sync ~__context ~self:x ~token ~token_id ;
              create_pool_repository ~__context ~self:x
          ) ;
+      Db.Pool.set_last_update_sync ~__context ~self:pool ~value:(Date.now ()) ;
```
minglumlu, line 3413: The `set_available_updates` may fail. I think it's better to set the `last_update_sync` after `set_available_updates`.

ocaml/idl/datamodel_lifecycle.ml
psafont: that sounds good, you'll have to remember to to commit them before that point

gangj: I think before the code is finally merged from the feature branch to master, a local make will re-update datamodel_lifecycle.ml to the latest version, will add that change then.

lindig (approved):

PR #4922, opened 2023-02-23
CP-40388: Call VDI.activate_readonly in SMAPIv3 when supported and appropriate

robhoes:


psafont (approved):

psafont (approved):

ocaml/xapi-storage-script/main.ml
```diff
@@ -1334,6 +1334,31 @@ let bind ~volume_script_dir =
     |> wrap
   in
   S.VDI.activate3 vdi_activate3_impl ;
+  let vdi_activate_readonly_impl dbg _dp sr vdi' vm' =
+    (let vdi = Storage_interface.Vdi.string_of vdi' in
+     let domain = Storage_interface.Vm.string_of vm' in
+     Attached_SRs.find sr >>>= fun sr ->
+     (* Discover the URIs using Volume.stat *)
+     stat ~dbg ~sr ~vdi >>>= fun response ->
+     (* If we have a clone-on-boot volume then use that instead *)
```
edwintorok, line 1343: Can't you reuse 'vdi_attach_common' for this part? The only thing that changes i the Datapath_client.<fun> to be called, so vdi_attach_common could be changed to take an ~f to invoke

edwintorok (commented):
Looks good in general, there might be an opportunity to avoid duplicating code and instead reuse 'vdi_attach_common'.

ocaml/xapi-storage-script/main.ml
robhoes: Unfortunately, the return types of `Datapath_client.attach` and `Datapath_client.activate` are different so that doesn't quite work out. I could probably still share code between the two activate calls.

robhoes:
@edwintorok I've added a commit to reduce duplication.

edwintorok (approved):

PR #4923, opened 2023-02-24
python: Use xapi's versioning scheme for XenAPI package (IH-203)

psafont:
This allows us to do releases much more frequently as pipy's policy
is to have immutable packages.

It also simplifies the build process quite a bit, which is neat.

The current method of building allows to build the package under git and
when git information is not available, using the same mechanism as xapi.
(using XAPI_VERSION if it's set, with a fallback on git, or failing
miserably)

This is not enough to be able to buuld it in a .spec environmtent
because the version schemas generated for the spec file conflict with
the allowed versions of python packages.

e.g.
koji generates
  23.5.0.19.g3657e16cb
python rejects it, it would accept
  23.5.0.19+g3657e16cb
the autogenerated equivalent generated by python would be
  23.5.0.dev19+g3657e16cb

This changeset is not meant to resolve this conflict, as the current
solution is good enough for its main use-case: generating a release
using github actions

I ran a couple of test workflows on my fork and it worked as expected (with a few tweaks to make it build on a commit that's not a tag):
workflow run: https://github.com/psafont/xen-api/actions/runs/4262752035/jobs/7418629627
release generated (contains the results of 2 workflow runs): https://github.com/psafont/xen-api/releases/tag/pyver


edwintorok (approved):

psafont:
Previously the pipeline had to compile xapi, now only the ./configure is needed :)

PR #4924, opened 2023-03-02
CP-40847: synchronize read-only uefi-certificates field for both host & pool

mg12:
in order to maintain backwards compatibility.


edwintorok (approved):

edwintorok:
looks good: Host.get_uefi_certificates will now reflect that host's state, as of the last XAPI startup.

robhoes (approved):

PR #4925, opened 2023-03-07
CA-364049: Tell external auth plugins to use python3

liulinC:
External auth pluins needs to import Enum, which is removed as
    secureboot-certificate does not Requires it any more.
    
    To resovle this issue there are two options:
    * Requires python2-enum34 in xapi.spec in re-introduce the package
    * Update the plugins to use python3
    
    Given python2 are going to be updated to puthon3, this commit fix
    the issue by telling plugins to use python3 explictly

### DIFF
--- a/scripts/plugins/extauth-hook
+++ b/scripts/plugins/extauth-hook
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Plugin called by xapi whenever an external_auth_type is enabled or disabled as an 
 # Also when a subject is added or removed.
--- a/scripts/plugins/extauth-hook-AD.py
+++ b/scripts/plugins/extauth-hook-AD.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 #
 # extauth-hook-AD.py
 #


liulinC:
Please hold this PR , there are some other issues needs to be resovled.

liulinC:
The issue  is caused by unrelated outdated private tag,  refreshed and looks good, 
Please continue review, thanks

scripts/plugins/extauth-hook
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
```
psafont, line 1: why not use `#!/usr/bin/env python3`?

edwintorok: When packages RPM will change .py files from '/usr/bin/env python3' to '/usr/bin/python3', but using '/usr/bin/env' might otherwise be preferable, e.g. if used in virtualenvs, where python3 might not be at /usr/bin/python3. https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_shebang_macros

liulinC: I was following other scripts under /etc/xapi.d/plugis, as follows, 
But let's update to use `env`
```grep "/usr/bin/python3" . -rn 
./on-slave:1:#!/usr/bin/python3
./extauth-hook-AD.py:1:#!/usr/bin/python3
./intellicache-clean:1:#!/usr/bin/python3
./tapdisk-pause:1:#!/usr/bin/python3
./testing-hooks:1:#!/usr/bin/python3
./trim:1:#!/usr/bin/python3
./nfs-on-slave:1:#!/usr/bin/python3
./lvhd-thin:1:#!/usr/bin/python3
./extauth-hook:1:#!/usr/bin/python3
./coalesce-leaf:1:#!/usr/bin/python3
```

psafont (approved):

robhoes (approved):

PR #4926, opened 2023-03-07
Fix total order in Ref.compare

edwintorok:
Explanation in commit messages. Draft, because I only unit tested it.


psafont (approved):

edwintorok:
I think Ref.compare is completely unused at the moment (my code on vtpm branch would be its first user?)

psafont:
I believe that's true, I added with the intention to use it for the database GC, but we ended up not needing it

edwintorok:
forgot to rerun formatting, CI should pass now, lets see.

lindig (approved):
I think the comments in the code should be updated.

lindig:
I believe this should be back ported

psafont:
This code is not present in 8.2.1, there's no need to backport

PR #4927, opened 2023-03-07
CA-375427: Make DP.destroy idempotent again

robhoes:
Before the recent storage API reorganisation, the DP.destroy call was implemented in storage_impl.ml. This file has been renamed storage_smapiv1_wrapper.ml and was pushed down from being the storage API entry point in xapi to only wrap SMAPIv1 calls, and now sits below the storage mux in the stack. The call in storage_impl.ml simply returns if there is no associated VDI with the given datapath in the local state, and there is nothing to "destroy".

The current entry point for the call in storage_mux.ml does raise an error if the datapath is not known, which leads to unexpected errors. This commits makes the call idempotent again.

### DIFF
--- a/ocaml/xapi/storage_mux.ml
+++ b/ocaml/xapi/storage_mux.ml
@@ -229,15 +229,14 @@ module Mux = struct
 
     let destroy _context ~dbg ~dp ~allow_leak =
       info "DP.destroy dbg:%s dp:%s allow_leak:%b" dbg dp allow_leak ;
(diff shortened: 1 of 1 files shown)


psafont (approved):
Such a weird behaviour to have :)

robhoes:
The checker tells me to make the pattern match more complete. Just a sec...

edwintorok (approved):

lindig (approved):

PR #4931, opened 2023-03-10
CP-39935 VTPM import/export

lindig:
Add support for VTPM import/export based on the current xapi database backend.

### DIFF
--- a/ocaml/xapi/export.ml
+++ b/ocaml/xapi/export.ml
@@ -476,6 +476,7 @@ let make_vtpm table __context self =
   let vtpm' =
     {
       vTPM'_VM= lookup' vtpm.vTPM_VM
+    ; vTPM'_uuid= vtpm.vTPM_uuid
     ; vTPM'_is_unique= vtpm.vTPM_is_unique
     ; vTPM'_is_protected= vtpm.vTPM_is_protected
     ; vTPM'_content= secret
--- a/ocaml/xapi/export.ml
+++ b/ocaml/xapi/export.ml
@@ -117,6 +117,10 @@ let rec update_table ~__context ~include_snapshots ~preserve_power_state
              add proxy.API.pVS_proxy_site
            )
        ) ;
+    (* add VTPMs that belong to this VM *)
+    vm.API.vM_VTPMs
+    |> List.iter (fun ref -> if Db.is_valid_ref __context ref then add ref) ;
+
     (* If we need to include snapshots, update the table for VMs in the 'snapshots' field *)
     if include_snapshots then
       List.iter
@@ -211,11 +215,6 @@ let make_vm ?(with_snapshot_metadata = false) ~preserve_power_state table
     __context self =
   let vm = Db.VM.get_record ~__context ~self in
   let vM_VTPMs = filter table (List.map Ref.string_of vm.API.vM_VTPMs) in
-  (* disallow exports and cross-pool migrations of VMs with VTPMs *)
-  ( if vM_VTPMs <> [] then
-      let message = "Exporting VM metadata with VTPMs attached" in
-      Helpers.maybe_raise_vtpm_unimplemented __FUNCTION__ message
-  ) ;
   let vm =
     {
       vm with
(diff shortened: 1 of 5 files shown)


ocaml/xapi/import.ml
```diff
@@ -1838,6 +1838,49 @@ module PVS_Site : HandlerTools = struct
...
+        debug "%s: importing VTPM" __FUNCTION__ ;
+        let vm =
+          log_reraise
+            ("Failed to find VTPM's VM: " ^ Ref.string_of vtpm'.vTPM'_VM)
+            (lookup vtpm'.vTPM'_VM) state.table
+        in
+        let self =
+          Xapi_vtpm.create ~__context ~vM:vm ~is_unique:vtpm'.vTPM'_is_unique
+        in
+        Xapi_vtpm.set_contents ~__context ~self ~contents:vtpm'.vTPM'_content
```
robhoes, line 1881: There are a few more fields that are part of the incoming record. Shouldn't we set those?

robhoes: What about `backend`, `persistence_backend` and `uuid` (other handlers preserve the uuid if `config.full_restore`; see VIF and VBD)? The `protected` field may be hardcoded now, but I assume later won't be, which may lead to surprises.

Many handlers use `Client.X.create_from_record` to set all fields at once. You could do that as well, after reconstructing the actual datamodel record.

edwintorok: can we set those fields to their defaults? doesn't matter what the backend was previously, we want to use the latest one. we do want to preserve the vtpm profile though.

ocaml/xapi/export.ml
```diff
@@ -467,6 +466,33 @@ let make_pvs_sites table __context self =
   ; snapshot= API.rpc_of_pVS_site_t site
   }
 
+let make_vtpm table __context self =
+  debug "exporting vtpm %s" (Ref.string_of self) ;
+  let lookup' ref = lookup table (Ref.string_of ref) in
+  let vtpm = Db.VTPM.get_record ~__context ~self in
+  let secret =
```
edwintorok, line 473: Can we use the API call implementation here? Then when we implement etcd it'll all be transparent to this code.
There is some special handling there for unique vtpms, which we'll need to think about for export. (I think the REQ has some details how to handle those)

ocaml/xapi/export.ml
```diff
@@ -467,6 +466,33 @@ let make_pvs_sites table __context self =
...
+    Db.Secret.get_value ~__context ~self
+  in
+  (* we are using a hand-crafted record that we serialise. The default
+     API type does not include the value of the VTPM *)
+  let vtpm' =
+    {
+      vTPM'_uuid= vtpm.vTPM_uuid
+    ; vTPM'_VM= lookup' vtpm.vTPM_VM
+    ; vTPM'_backend= vtpm.vTPM_backend
+    ; vTPM'_persistence_backend= vtpm.vTPM_persistence_backend
```
edwintorok, line 484: I don't think the persistence backend has to be exported. We'll store the same data just in a different place, so it should be possible to export using XAPI backend and import using ETCD backend

ocaml/xapi/import.ml
lindig: Is it safe to update the UUID? Currently we we get a new UUID but I would assume it could be likewise kept. The problem here is that this is: 
* if the UUID already exists, I probably should not import the VTPM and use the existing one - but sharing between VTPMs seems not like a good idea
* if I create a new VTPM but update the UUID, I could end up with a UUID that is not unique

lindig:
I have added restoring the UUID when a full restore is requested. This includes a check that this UUID does not already exist.

lindig:
I will clean up commits after review because doing it now would mess up the kojienv workflow.

ocaml/xapi/import.ml
```diff
@@ -1838,6 +1838,79 @@ module PVS_Site : HandlerTools = struct
...
+     already but always create it *)
+
+  type precheck_t = Import of vtpm'
+
+  let fail fmt =
+    Printf.kprintf
+      (fun msg -> raise Api_errors.(Server_error (import_error_generic, [msg])))
+      fmt
+
+  let collison ~__context ~uuid =
```
psafont, line 1853:
```suggestion
  let collision ~__context ~uuid =
```

ocaml/xapi/import.ml
```diff
@@ -1838,6 +1838,79 @@ module PVS_Site : HandlerTools = struct
...
+
+  let precheck __context config _rpc _session _state obj =
+    let vtpm' = vtpm'_of_rpc obj.snapshot in
+    debug "%s: importing VTPM" __FUNCTION__ ;
+    match config.full_restore with
+    | false ->
+        Import vtpm'
+    | true ->
+        let uuid = vtpm'.vTPM'_uuid in
+        if collison ~__context ~uuid then
```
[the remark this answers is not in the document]
lindig: I renamed this to `colliding`, which I think it better anyway.

psafont (approved):

PR #4932, opened 2023-03-13
fe_systemctl: fix quoting and add a quicktest

edwintorok:
This is in preparation of running etcd as a (templated/transient) systemd service.

The test here launches 'env' through fe_systemctl and sets an environment variable containing all printable characters as values to check that quoting is done properly.
In particular systemd's quoting doesn't match either `Filename.quote` or `String.escaped` and needs a hand-written escaper. The syntax is [documented here](https://www.freedesktop.org/software/systemd/man/systemd.syntax.html#Quoting), notice how Filename.quote would produce the wrong results when escaping the single quote character (`'\''`) and String.escaped would not quote it at all.
Quote single and double quotes using hex escape sequences (this makes it independent to the kind of quote used to wrap items).

The test is not suitable for a unit test because it depends on external tools (systemd), which may not be available in all build environments (in particular won't be available in the RPM build environment, not even in --user mode).
Quicktests are tests that run in Dom0, so it is a suitable place to run this test automatically without having to add a lot of extra machinery to run the test.

However it is still useful to run this test in a development environment (provided you have 'systemctl --user' working there, which most Linux distros or containers would), so add some flags to quicktest to be able to run without XAPI, which can be used like this:
`dune exec ./quicktest.exe -- -skip-xapi`.
When the argument is passed the XAPI session is not initialized and any queries to XAPI to list VMs/SRs will return the empty list (thus these tests won't run at all). Reading the XAPI inventory for localhost UUID is also skipped when UID is not  
 root (0), otherwise you'd get a permission denied error trying to create it.

If you want to pass additional arguments to alcotest (e.g. to 'list' or run just  a particular test with 'test')  then you can now do:
`dune exec ./quicktest.exe -- -skip-xapi -- list` (the first '--' is interpreted by 'dune' itself to pass through the arguments to the underlying command without trying to interpret them as dune arguments, and the second '--' is parsed by the new code in quicktest).

Also adds a start_templated convenience wrapper that starts a templated 'foo@instance' parametrized systemd template and a 'set_properties' that can be used to override certain settings via systemd drop-in (e.g. Environment variables).
We do not use 'systemctl set-property' because that only supports setting runtime changeable properties (mostly cgroup) properties, and not settings that require a service restart like Environment.
Although we could use entirely transient systemd services to always set the environment, it might be useful to be able to use a templated systemd service with the environment customized per instance, so that the customization survives a reboot and the service is started up even if XAPI would have difficulty starting on next boot (this will be useful to keep the quorum alive with services like 'etcd').

PR to feature branch, since this is likely to evolve as part of its use in etcd and the quicktest changes still need testing in XenRT.

### DIFF
--- a/ocaml/forkexecd/lib/fe_systemctl.ml
+++ b/ocaml/forkexecd/lib/fe_systemctl.ml
@@ -18,6 +18,10 @@ type status = {
   ; active_state: string
 }
 
+module StringMap = Map.Make (String)
+
+type 'a string_map = 'a StringMap.t
+
 let systemctl = "/usr/bin/systemctl"
 
 let action ~service action =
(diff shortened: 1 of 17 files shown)


edwintorok:
('make switch' and 'make start' worked after setting CHECK_IF_PREINSTALLED=false to get ocamlbuild to install)

ocaml/forkexecd/lib/fe_systemctl.ml
```diff
@@ -18,54 +18,126 @@ type status = {
...
+  |> List.map env_pair
+  |> List.map (fun v -> ("Environment", [v]))
+(* we could build just a single environment line, but might be too long and
+   difficult to debug *)
+
+let build_properties env base properties =
+  "[Service]"
+  :: (List.concat [environment env; base; properties]
+     |> List.map (fun (k, v) ->
+            k ^ "=" ^ (List.map systemd_quote v |> String.concat " ")
```
lindig, line 85: I believe `^`is quite inefficient; why is `k` not escaped?

ocaml/forkexecd/lib/fe_systemctl.ml
```diff
@@ -18,54 +18,126 @@ type status = {
...
        ; "DefaultDependencies=no" (* lifecycle tied to domain, not systemd *)
-       ; "[Service]"
        ]
   |> String.concat "\n"
   |> Xapi_stdext_unix.Unixext.write_string_to_file destination ;
   action ~service "start"
 
+let unit_path () =
+  if !test_mode then
+    let home = Sys.getenv_opt "HOME" |> Option.value ~default:"root" in
```
lindig, line 121: I suspect the default is wrong; should this not be `/root`?

ocaml/quicktest/quicktest_fe_systemctl.ml
```diff
@@ -0,0 +1,88 @@
...
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+let home = Sys.getenv_opt "HOME" |> Option.value ~default:"/root"
```
lindig, line 15: Correct here

lindig (approved):

ocaml/forkexecd/lib/fe_systemctl.ml
edwintorok: I replaced `k ^ "=" ^ ..." with `String.concat "" [k; "="; ...]`. I haven't measured but the implementation of 'String.concat' looks quite efficient (it first calculates how big of a buffer it needs, allocates it, then blits all the individual strings with the proper offsets)

ocaml/forkexecd/lib/fe_systemctl.mli
```diff
@@ -22,6 +22,34 @@ type status = {
...
+    [None] removes any override for the property, setting back to default
+
+  @param service the systemd service to change. Changes take effect on next
+  restart.
+
+*)
+
+val start_templated : template:string -> instance:string -> unit
+(** [start_templated ~template ~instance] starts [template@instance.service]
+  instance of the [template@.service] systemd template unit.
```
psafont, line 47: It's problably worth mentioning here that the service files is created at this point, and it is deleted when stopping the service, which may not always be expected as these are not transient

psafont (approved):

ocaml/forkexecd/lib/fe_systemctl.mli
edwintorok: For start the service file is expected to exist already, I'll add this to the docstring.
On stop it indeed deletes the service file, but only if it was transient (a non-transient service won't be found in the transient path, so the unlink_safe will be a no-op). I'll document that too.

PR #4934, opened 2023-03-15
CP-42013: Do not apply recommended guidances automatically

gangj:
1. add a new field "recommended_guidances" in both host and vm datamodel
2. return "recommended_guidances" together with "warnings" as update applying results from host.apply_updates
3. add API host.apply_recommended_guidances
4. refactor "immediate_guidances" to "recommended_guidances"

### DIFF
--- a/ocaml/idl/datamodel_lifecycle.ml
+++ b/ocaml/idl/datamodel_lifecycle.ml
@@ -15,14 +15,10 @@ let prototyped_of_field = function
       Some "22.26.0"
   | "VTPM", "persistence_backend" ->
       Some "22.26.0"
-  | "host", "recommended_guidances" ->
-      Some "23.3.0-next"
   | "host", "https_only" ->
       Some "22.27.0"
   | "host", "last_software_update" ->
       Some "22.20.0"
-  | "VM", "recommended_guidances" ->
-      Some "23.3.0-next"
   | "VM", "actions__after_softreboot" ->
       Some "23.1.0"
   | "pool", "last_update_sync" ->
@@ -49,8 +45,6 @@ let prototyped_of_message = function
       Some "22.26.0"
   | "VTPM", "create" ->
       Some "22.26.0"
-  | "host", "apply_recommended_guidances" ->
-      Some "23.3.0-next"
   | "host", "set_https_only" ->
       Some "22.27.0"
   | "pool", "set_https_only" ->
--- a/ocaml/idl/datamodel_lifecycle.ml
+++ b/ocaml/idl/datamodel_lifecycle.ml
@@ -15,10 +15,14 @@ let prototyped_of_field = function
       Some "22.26.0"
   | "VTPM", "persistence_backend" ->
       Some "22.26.0"
+  | "host", "recommended_guidances" ->
+      Some "23.3.0-next"
   | "host", "https_only" ->
       Some "22.27.0"
   | "host", "last_software_update" ->
       Some "22.20.0"
+  | "VM", "recommended_guidances" ->
+      Some "23.3.0-next"
   | "VM", "actions__after_softreboot" ->
       Some "23.1.0"
   | "pool", "last_update_sync" ->
@@ -45,6 +49,8 @@ let prototyped_of_message = function
       Some "22.26.0"
   | "VTPM", "create" ->
       Some "22.26.0"
+  | "host", "apply_recommended_guidances" ->
+      Some "23.3.0-next"
   | "host", "set_https_only" ->
       Some "22.27.0"
   | "pool", "set_https_only" ->
--- a/ocaml/xapi-cli-server/cli_frontend.ml
+++ b/ocaml/xapi-cli-server/cli_frontend.ml
@@ -1027,6 +1027,15 @@ let rec cmdtable_data : (string * cmd_spec) list =
       ; flags= [Host_selectors]
       }
     )
+  ; ( "host-apply-recommended-guidances"
+    , {
+        reqd= []
+      ; optn= []
+      ; help= "Apply recommended guidances on a host."
+      ; implementation= No_fd Cli_operations.host_apply_recommended_guidances
+      ; flags= [Host_selectors]
+      }
+    )
   ; ( "patch-upload"
     , {
         reqd= ["file-name"]
--- a/ocaml/xapi/repository.ml
+++ b/ocaml/xapi/repository.ml
@@ -688,30 +688,23 @@ let apply_livepatch ~__context ~host:_ ~component ~base_build_id ~base_version
       error "%s" msg ;
       raise Api_errors.(Server_error (internal_error, [msg]))
 
-let set_pending_restart_device_models ~__context ~host =
+let set_restart_device_models ~__context ~host ~kind =
   (* Set pending restart device models of all running HVM VMs on the host *)
   do_with_device_models ~__context ~host @@ fun (ref, record) ->
   match
     (record.API.vM_power_state, Helpers.has_qemu_currently ~__context ~self:ref)
   with
-  | `Running, true | `Paused, true ->
-      Db.VM.add_pending_guidances ~__context ~self:ref
-        ~value:`restart_device_model ;
-      None
-  | _ ->
-      (* No device models are running for this VM *)
-      None
-
-let set_recommended_restart_device_models ~__context ~host =
-  (* Set recommended restart device models of all running HVM VMs on the host *)
-  do_with_device_models ~__context ~host @@ fun (ref, record) ->
-  match
-    (record.API.vM_power_state, Helpers.has_qemu_currently ~__context ~self:ref)
-  with
-  | `Running, true | `Paused, true ->
-      Db.VM.add_recommended_guidances ~__context ~self:ref
-        ~value:`restart_device_model ;
-      None
+  | `Running, true | `Paused, true -> (
+    match kind with
+    | Guidance.Absolute ->
+        Db.VM.add_pending_guidances ~__context ~self:ref
+          ~value:`restart_device_model ;
+        None
+    | Guidance.Recommended ->
+        Db.VM.add_recommended_guidances ~__context ~self:ref
+          ~value:`restart_device_model ;
+        None
+  )
   | _ ->
       (* No device models are running for this VM *)
       None
@@ -730,7 +723,7 @@ let set_pending_guidances ~__context ~host ~guidances =
            Db.Host.add_pending_guidances ~__context ~self:host
              ~value:`restart_toolstack
        | RestartDeviceModel ->
-           set_pending_restart_device_models ~__context ~host
+           set_restart_device_models ~__context ~host ~kind:Absolute
        | g ->
            warn "Unsupported pending guidance %s, ignoring it."
              (Guidance.to_string g)
@@ -747,7 +740,7 @@ let set_recommended_guidances ~__context ~host ~guidances =
            Db.Host.add_recommended_guidances ~__context ~self:host
              ~value:`restart_toolstack
        | RestartDeviceModel ->
-           set_recommended_restart_device_models ~__context ~host
+           set_restart_device_models ~__context ~host ~kind:Recommended
        | g ->
            warn "Unsupported recommended guidance %s, ignoring it."
              (Guidance.to_string g)
@@ -787,10 +780,6 @@ let apply_livepatches' ~__context ~host ~livepatches =
 let apply_updates' ~__context ~host ~updates_info ~livepatches ~acc_rpm_updates
     =
   (* This function runs on coordinator host *)
-  let expected_recommended_guidances =
-    eval_guidances ~updates_info ~updates:acc_rpm_updates ~kind:Recommended
-      ~livepatches ~failed_livepatches:[]
-  in
   (* Install RPM updates firstly *)
   Helpers.call_api_functions ~__context (fun rpc session_id ->
       Client.Client.Repository.apply ~rpc ~session_id ~host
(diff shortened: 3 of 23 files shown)


ocaml/xapi-cli-server/cli_frontend.ml
```diff
@@ -1027,6 +1027,15 @@ let rec cmdtable_data : (string * cmd_spec) list =
       ; flags= [Host_selectors]
       }
     )
+  ; ( "host-apply-recommended-guidances"
```
lindig, line 1030: Small thing; while "guidances" is the plural of guidance, it is quite a rate word. I would have used the singular "guidance" (here and everywhere else) since a guidance may conceptually comprise multiple steps.

ocaml/xapi/repository.ml
```diff
@@ -800,6 +736,23 @@ let set_pending_guidances ~__context ~host ~guidances =
...
+       | RebootHost ->
+           Db.Host.add_recommended_guidances ~__context ~self:host
+             ~value:`reboot_host
+       | RestartToolstack ->
+           Db.Host.add_recommended_guidances ~__context ~self:host
+             ~value:`restart_toolstack
+       | RestartDeviceModel ->
+           set_recommended_restart_device_models ~__context ~host
+       | g ->
+           warn "Unsupported recommended guidance %s, ignore it."
```
lindig, line 752: "ignoring it" would be better

ocaml/xapi/repository_helpers.ml
```diff
@@ -1076,3 +1076,24 @@ let prune_updateinfo_for_livepatches livepatches updateinfo =
     List.filter (fun x -> LivePatchSet.mem x livepatches) updateinfo.livepatches
   in
   {updateinfo with livepatches= lps}
+
+let do_with_device_models ~__context ?(vm_with_recommended_guidances = false)
```
lindig, line 1080: The meaning of `vm_with_recommended_guidances` given the type `bool` is not obvious. What does this flag represent?

ocaml/xapi-cli-server/cli_frontend.ml
gangj: There might be multiple recommended "guidances" after a update, i.e.: "RestartDeviceModel and RestartToolstack", and as the XAPI code from Centaurus was named this way, I would prefer to follow it in this PR. If we decide to change the naming, I think we can do it in a separate PR with low priority.

ocaml/xapi/repository_helpers.ml
gangj: If "vm_with_recommended_guidances " is true, "do_with_device_models" will execute "f" with only vm with recommended guidances,
If "vm_with_recommended_guidances " is false, "do_with_device_models" will execute "f" with all vms no matter if a vm has recommended guidances or not.

lindig: So this acts as a filter. I would suggest to use `only_vm_with_guidances` or `limit_to_vm_with_guidances`.

ocaml/xapi/repository_helpers.ml
```diff
@@ -1076,3 +1076,24 @@ let prune_updateinfo_for_livepatches livepatches updateinfo =
...
+     )
+  |> List.map (fun self -> (self, Db.VM.get_record ~__context ~self))
+  |> List.filter (fun (_, record) -> not record.API.vM_is_control_domain)
+  |> List.filter_map f
+  |> function
+  | [] ->
+      ()
+  | _ :: _ ->
+      let host' = Ref.string_of host in
+      raise Api_errors.(Server_error (cannot_restart_device_model, [host']))
```
lindig, line 1099: Only VMs have a device model - so why is the host part of the message?

gangj: It is expected that any failure when executing "f" should be logged by "f" itself, and as the error here happens when trying to restart device model for the "host", the "cannot_restart_device_model" error of the host is logged here.

ocaml/xapi/repository_helpers.ml
gangj: As now "vm" obj has both "recommended_guidances" and "pending_guidances", renamed it to "only_vm_with_recommended_guidances"

ocaml/xapi/xapi_host.ml
```diff
@@ -2904,14 +2905,97 @@ let apply_updates ~__context ~self ~hash =
   in
   Db.Host.set_last_software_update ~__context ~self
     ~value:(get_servertime ~__context ~host:self) ;
-  Repository.apply_immediate_guidances ~__context ~host:self ~guidances ;
   (* The warnings may not be returned to async caller if this happens
```
minglumlu, line 2908: This comment will not be applicable anymore if the `restartToolstack` will not be performed here.

ocaml/xapi/xapi_host.ml
```diff
@@ -2904,14 +2905,97 @@ let apply_updates ~__context ~self ~hash =
...
+        )
+        host_update_guidances
+    in
+    match host_guidances with
+    | [] ->
+        (* try to restart device models for vm recommended_guidances*)
+        restart_device_models_for_recommended_guidances ~__context ~host:self
+    | [RebootHost] ->
+        reboot ~__context ~host:self
+    | [EvacuateHost] ->
```
minglumlu, line 2977: Unlike the logic you are removing, in which the guidances are parsed from updateinfo and may contain `EvacuateHost`, I think it's better to remove it from `recommended_guidances` set in the first place.

ocaml/xapi/xapi_host.ml
```diff
@@ -2904,14 +2905,97 @@ let apply_updates ~__context ~self ~hash =
...
+        (* EvacuatHost should have been done before applying updates by XAPI users.
+         * Here only the guidances to be applied after applying updates are handled,
+         * nothing need to be done for EvacuatHost.
+         *)
+        ()
+    | [RestartToolstack] ->
+        (* try to restart device models for vm recommended_guidances*)
+        restart_device_models_for_recommended_guidances ~__context ~host:self ;
+        restart_agent ~__context ~host:self
+    | l when GuidanceSet.eq_set1 l ->
```
minglumlu, line 2987: If the `EvacuateHost` will not be expected here, the combination (set2) of `RestartDeviceModel` and `RestartToolstack` should be handled here.

ocaml/xapi/xapi_host.ml
```diff
@@ -2904,14 +2905,97 @@ let apply_updates ~__context ~self ~hash =
...
+        (* EvacuateHost and RestartToolstack *)
+        restart_agent ~__context ~host:self
+    | l ->
+        let host' = Ref.string_of self in
+        error
+          "Found wrong guidance(s) after applying updates on host ref='%s': %s"
```
minglumlu, line 2993: The context `after applying updates` would not be applicable since it is in a separated API.

ocaml/xapi/repository.ml
```diff
@@ -716,70 +702,20 @@ let set_pending_restart_device_models ~__context ~host =
       (* No device models are running for this VM *)
       None
 
-let restart_device_models ~__context ~host =
-  (* Restart device models of all running HVM VMs on the host by doing
-   * local migrations. *)
-  do_with_device_models ~__context ~host ~action_label:"restart"
-  @@ fun (ref, record) ->
+let set_recommended_restart_device_models ~__context ~host =
```
minglumlu, line 705: This should be able to be combined with `set_pending_restart_device_models` as most bits of them are same.

ocaml/xapi/repository_helpers.ml
```diff
@@ -1076,3 +1076,24 @@ let prune_updateinfo_for_livepatches livepatches updateinfo =
     List.filter (fun x -> LivePatchSet.mem x livepatches) updateinfo.livepatches
   in
   {updateinfo with livepatches= lps}
+
+let do_with_device_models ~__context
+    ?(only_vm_with_recommended_guidances = false) ~host f =
```
minglumlu, line 1081: You can also put the filter in the `match` of `restart_device_models_for_recommended_guidances`.
In this case, the specific logic will be in the specific function.

ocaml/xapi/xapi_host.ml
```diff
@@ -2904,14 +2905,97 @@ let apply_updates ~__context ~self ~hash =
...
+      Db.Host.get_recommended_guidances ~__context ~self
+    in
+    let host_guidances =
+      List.map
+        (fun ug ->
+          match ug with
+          | `reboot_host ->
+              Guidance.RebootHost
+          | `reboot_host_on_livepatch_failure ->
+              Guidance.RebootHostOnLivePatchFailure
```
minglumlu, line 2963: This will not be in recommended guidances.

minglumlu: Would this be clearer?
```
Db.Host.get_recommended_guidances ~__context ~self
|> List.map (...)
|> function
| [RebootHost] ->
| [RestartToolstack] ->
......
```

ocaml/xapi/xapi_host.ml
```diff
@@ -2904,14 +2905,97 @@ let apply_updates ~__context ~self ~hash =
...
+      None
+  | `Paused, true ->
+      error "VM 'ref=%s' is paused, can't restart device models for it"
+        (Ref.string_of ref) ;
+      Some ref
+  | _ ->
+      (* No device models are running for this VM *)
+      None
+
+let apply_recommended_guidances ~__context ~self =
```
minglumlu, line 2948: In this function, the host may be rebooted, the toolstack of this host may be restarted, or the HVM VMs with `restart_device_models` may be local migrated.
Putting them together in one API results in the code like 
```
    match host_guidances with
    | [] ->
        (* try to restart device models for vm recommended_guidances*)
        restart_device_models_for_recommended_guidances ~__context ~host:self
    | [RebootHost] ->
```
The whole logic flow may be a little confusing to perform the VMs' guidance based on host's guidances.
I think the VMs' guidance can be handled explicitly and separately in this function rather than being mixed in host guidance handling.

minglumlu, line 2948: Another concern is that this function will run on a member host directly. I don't have strong argument for not doing this given now XAPI doesn't have special logic in handling, i.e. `host.reboot, in `message_forwarding` in coordinator. I think an alternative is just to handle this in coordinator, and send primitive XAPI calls, like `host.reboot` to member hosts.

minglumlu: Basically, I think the existing `host.reboot`, `host.restart_agent`, and `vm.pool_migrate` can be used by clients directly. This will get things simple.

ocaml/xapi/xapi_host.ml
gangj: Updated the code to follow that only "[]", "[`reboot_host]" or "[`restart_toolstack]" is a valid possible value of "Db.Host.get_recommended_guidances"

ocaml/xapi/xapi_host.ml
gangj: error log updated

ocaml/xapi/repository.ml
gangj: combined

ocaml/xapi/xapi_host.ml
gangj: If we expose `host.reboot`, `host.restart_agent`, and `vm.pool_migrate` to clients(like XenCenter), then clients needs to follow the logic here inside `apply_recommended_guidances` to apply recommended guidances, so I think exposing a single API to apply the recommended guidances is more convenient for clients.

ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -15,10 +15,14 @@ let prototyped_of_field = function
       Some "22.26.0"
   | "VTPM", "persistence_backend" ->
       Some "22.26.0"
+  | "host", "recommended_guidances" ->
```
psafont, line 18: Please remove the changes in this file out until it's time to merge to master, the changes are already outdated

ocaml/xapi/xapi_host.ml
```diff
@@ -2904,14 +2905,90 @@ let apply_updates ~__context ~self ~hash =
...
+                   (fun ug ->
+                     match ug with
+                     | `reboot_host ->
+                         Guidance.RebootHost
+                     | `reboot_host_on_livepatch_failure ->
+                         Guidance.RebootHostOnLivePatchFailure
+                     | `restart_toolstack ->
+                         Guidance.RestartToolstack
+                     | `restart_device_model ->
+                         Guidance.RestartDeviceModel
```
psafont, line 2983: Please make this an independent function and place it somewhere where it can be reused (maybe in repository.ml?)
The you can use that function here, instead of having so many nesting levels here

ocaml/idl/datamodel_lifecycle.ml
gangj: removed

ocaml/xapi/xapi_host.ml
gangj: Added a function "of_update_guidance" in module Guidance, which transforms a "update_guidance" to Guidance.t.

psafont (approved):

ocaml/xapi/xapi_host.ml
```diff
@@ -2904,14 +2905,77 @@ let apply_updates ~__context ~self ~hash =
...
   @@ Helpers.call_script !Xapi_globs.firewall_port_config_script [state; "80"] ;
   Db.Host.set_https_only ~__context ~self ~value
+
+let try_restart_device_models_for_recommended_guidances ~__context ~host =
+  (* Restart device models of all running HVM VMs on the host by doing
+   * local migrations if it is required by recommended guidances. *)
+  Repository_helpers.do_with_device_models ~__context ~host
+  @@ fun (ref, record) ->
+  if
+    List.mem `restart_device_model
```
minglumlu, line 2930: This can be in `match` also.

minglumlu (approved):

PR #4936, opened 2023-03-15
xapi_vtpm: do not reuse name for get_contents

psafont:
The internal function that's more complex can have another name, rename so it's easier to see which one is being used and the intention behind it.

### DIFF
--- a/ocaml/xapi/xapi_vtpm.ml
+++ b/ocaml/xapi/xapi_vtpm.ml
@@ -53,7 +53,7 @@ let introduce ~__context ~vM ~persistence_backend ~contents ~is_unique =
   ref
 
 (** Contents from unique vtpms cannot be copied! *)
-let get_contents ~__context ?from () =
+let copy_or_create_contents ~__context ?from () =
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_vtpm.ml
```diff
@@ -53,7 +53,7 @@ let introduce ~__context ~vM ~persistence_backend ~contents ~is_unique =
   ref
 
 (** Contents from unique vtpms cannot be copied! *)
-let get_contents ~__context ?from () =
+let copy_or_create_contents ~__context ?from () =
```
lindig, line 56: Looking from the outside, the semantics are a bit surprising. We either get a copy of an existing secret, or an empty secret. The name captures this (maybe `copy_or_create_secret`) but it's hard to imagine a client that finds this useful.

psafont: This function is not exposed in the interface, it's meant to be used inside the module only

robhoes (approved):

PR #4937, opened 2023-03-17
CP-39863 add "allowed operations" support for VTPMs

lindig:
An object (like a VM, VTPM) can have operations (an enumeration) and can announce which operations are currently allowed. This helps API clients to avoid operations which then fail because the object is not in the correct state. While an API failure is not a problem, it is only detected after the fact. Allowed operations help in particular interactive clients to block operations that would fail.

Currently neither VMs nor VTPMs have declared operations related to VTPMs. Hence, the purpose of this ticket is to introduce them and to implement the corresponding checks.


ocaml/idl/datamodel_vtpm.ml
```diff
@@ -79,5 +101,6 @@ let t =
       ; field ~qualifier:DynamicRO ~ty:(Ref _secret) ~internal_only:true
           ~lifecycle:[] "contents" "The contents of the TPM"
       ]
-    ~messages:[create; destroy; get_contents; set_contents]
+    ~messages:
+      [create; destroy; get_contents; set_contents; assert_operation_valid]
```
edwintorok, line 105: I think there is a shortcut to declare allowed operations: `allowed_and_current_operations operations` that should generate some of the boilerplate.

ocaml/xapi/xapi_vtpm.ml
```diff
@@ -12,6 +12,30 @@
...
+  (* currently every VTPM has an associated VM *)
+  assert (Db.is_valid_ref __context vm) ;
+  let state = Db.VM.get_power_state ~__context ~self:vm in
+  let ok = () in
+  match op with
+  | `destroy when state = `Halted ->
+      ok
+  | `set_contents when state = `Halted ->
+      ok
+  | `get_contents when state = `Halted ->
```
edwintorok, line 34: I think API users are not meant to use set/get contents at all, shouldn't it be purely a XAPI operation?
(and currently swtpm will only use it when the VM is running).
In general API users should use export/import VM metadata instead.

Maybe we should mark set/get contents as internal only so you cannot call it with *any* API role unless you are local root and XAPI itself (or swtpm).

ocaml/idl/datamodel_vtpm.ml
robhoes: Yes, that's a helper function to add the two fields.

ocaml/idl/datamodel_vm.ml
```diff
@@ -1616,6 +1616,8 @@ let operations =
         ; ("metadata_export", "exporting VM metadata to a network stream")
         ; ("reverting", "Reverting the VM to a previous snapshotted state")
         ; ("destroy", "refers to the act of uninstalling the VM")
+        ; ("create_vtpm", "Creating and adding a VTPM to this VM")
+        ; ("destroy_vtpm", "Removing a VTPM from this VM")
```
robhoes, line 1620: The destroy seems to overlap with the operation on the VTPM class itself. I think it belongs on that class?

ocaml/idl/datamodel_vtpm.ml
```diff
@@ -53,6 +53,28 @@ let set_contents =
...
+  Enum
+    ( "vtpm_operations"
+    , [
+        ("destroy", "Destroy a VTPM")
+      ; ("get_contents", "Read data from a VTPM")
+      ; ("set_contents", "Write to the VTPM, overwriting existing content")
+      ]
+    )
+
+let assert_operation_valid =
```
robhoes, line 66: Do we need this function in the API? The GUI normally uses the `allowed_operations` fields in preference to function calls, and we don't have similar functions in the API for other devices (though VM does have it, for some reason).

robhoes: Updates to those fields as well as API-level locking is managed through message_forwarding. See e.g. how `with_vif_marked` is used in message_forwarding.ml.

robhoes: Similarly, you should use `with_vm_operation` in `message_forwarding.ml` on the `VTPM.create` function to ensure that the VM operations fields are set correctly, and the check is enforced on the create function.

ocaml/xapi/xapi_vtpm.ml
lindig: The allowed operations have no direct connection with API calls - they are connected by convention and form a set of names. So we can put anything here we like but I mirrored the API calls. We can still make the get/set contents API calls private - which would leave only create and destroy as operations.

ocaml/idl/datamodel_vm.ml
lindig: I will remove it here (it exists in both places).

lindig:
I have addressed some of the comments:
* removed `destroy_vtpm` from VM
* use `with_vm_operation`

lindig:
I have reworked this - now using a field holding allowed operations. This field is updated when then VTPM's VM operations are updated because all that matters (currently) is the power state of the VM.

ocaml/idl/datamodel_vtpm.ml
robhoes: This is now unused (not mentioned in `~messages` below). I think we can remove it, because XC is unlikely to use it.

ocaml/xapi/xapi_vm_lifecycle.ml
```diff
@@ -728,6 +741,14 @@ let assert_operation_valid ~__context ~self ~op ~strict =
   | Some (a, b) ->
       raise (Api_errors.Server_error (a, b))
 
+(* can't put this into xapi_vtpm because it creates a cycle *)
```
robhoes, line 744: OK now for such a simple function. If this gets more complex, then perhaps put it in a separate file like xapi_vtpm_helpers.ml.

robhoes (approved):
There is a conflict in `schematest.ml` and an unused function `assert_operation_valid` in the datamodel. Otherwise this looks good.

lindig:
I have updated this but excluded lifecycle updates for telemetry code that was automatically generated. We might want to do this separately

ocaml/idl/datamodel_vtpm.ml
```diff
@@ -53,31 +53,46 @@ let set_contents =
       ]
     ~hide_from_docs:true ~allowed_roles:_R_LOCAL_ROOT_ONLY ()
 
+let operations =
+  Enum
+    ( "vtpm_operations"
+    , [
+        ("destroy", "Destroy a VTPM")
+      ; ("get_contents", "Read data from a VTPM")
+      ; ("set_contents", "Write to the VTPM, overwriting existing content")
```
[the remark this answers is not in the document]
lindig: The get/set content methods are forwarded but they are currently not checking the VM state either in the forwarding code or the actual implementation. What should happen here? I think as a minimum we should only forward when the VM is halted.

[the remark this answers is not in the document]
psafont: For snapshots reading the vtpm contents while is paused is needed. These calls are hidden from documentation and seem to exist only to facilitate the implementation.

If you think we'll benefit from having allowed operations, feel free to reinforce them

psafont (approved):

PR #4939, opened 2023-03-20
CP-48838: Various build system and testing improvements

edwintorok:
Draft, still need to make and test the specfiles changes and a chainbuild.
Explanations in commit messages.

Not vTPM specific, but the changes were made to make vtpm/etcd development easier, so test it out here first (and once it works we can push this part to master fairly quickly)


edwintorok:
This depends on having Dune 3.7 in xs-opam, which I thought we did, but looks like that is waiting on the 4.14 branch to be merged:[on this PR here ](https://github.com/xapi-project/xs-opam/pull/619/commits)

xapi-datamodel.opam
```diff
@@ -17,7 +17,6 @@ depends: [
   "rpclib"
   "base-threads"
   "xapi-consts"
-  "xapi-schema"
```
psafont, line 20: This probably was needed before this reversal: https://github.com/xapi-project/xen-api/commit/9608b74224f59c8e6c2b59a039a40dc609618a1c

dune
```diff
@@ -3,7 +3,7 @@
     (ocamlopt_flags (:standard -g -p -w -39))
     (flags (:standard -w -39))
   )
-  (dev (flags (:standard -g -w -39)))
+  (dev (flags (:standard -g -w -39 -warn-error -32-38-67-69)))
```
psafont, line 6: My prediction is we'll forget about these and will be left disabled

psafont:
This PR depends on https://github.com/xapi-project/xs-opam/pull/634

dune
edwintorok: You do see the warnings, they are just not errors (which is why I didn't completely turn off the warnings). When I have some time I'll try to clean them up, but there were too many for a friday.

psafont:
Not sure why the ocaml tests fail, when I run the tests locally, they all pass

psafont:
```
Fatal error: exception Unix.Unix_error(Unix.ECONNRESET, "read", "")
File "ocaml/forkexecd/test/dune", line 6, characters 0-130:
 6 | (rule
 7 |  (alias runtest)
 8 |  (package xapi-forkexecd)
 9 |  (deps fe_test.sh fe_test.exe ../src/fe_main.exe)
10 |  (action
11 |   (run ./fe_test.sh)))
(cd _build/default/ocaml/forkexecd/test && ./fe_test.sh)
Command exited with code 2.
```

edwintorok:
pushed some changes to the script to output a backtrace on failure, and also tweaked how dune runtest works in the CI so we get the output from the failed tests separated from the successful ones.

edwintorok:
There is an ECONNRESET error here from forkexecd, I'll take a look

edwintorok:
There is a race condition in 'fe_test.sh' (that is more apparent now that I reduced the sleeps, but the problem was there before): the test script only checks for the presence of the socket and then starts a client, but the server obviously has to create the socket first and then start listening on it, which leaves a short window where the socket exists, nothing listens on it and clients would get ECONNREFUSED.

This is usually fixed by systemd daemon notify (when run within systemd), or by the parent process existing only once it has 'listen()'-ed on the socket, but 'forkexecd' does neither.

edwintorok:
Hmm, I got a successful build of v6d and I indeed dropped the commits that made those libraries private, looks like I force pushed an outdated branch here. Let me check again against the branch that got tested.

psafont (approved):

edwintorok:
Hmm now we have ECONNRESET bugs in the forkexecd tests instead of ECONNREFUSED. I could move the forkexecd changes out to a separate pull request, but I'll take a look first on whether I can reproduce and track down what goes wrong here (would be good to make these tests more reliable).

edwintorok:
This needs to be rebased, and potentially the forkexec changes separated out if they expose pre-existing bugs.
For now closing the PR, but I'll try to get back to this fairly soon as it reduces build times and fixes a few other issues with the build/tests that improves developer efficiency.

PR #4940, opened 2023-03-20
Start implementing some etcd management code in XAPI

edwintorok:
This is just the very basic design around library layout and some helper data structures.
More to come soon as soon as I split up the config module and test code.


psafont:
`dune-project` needs to be changed to use `(lang dune 2.8)`

edwintorok:
I have a 'private/edvint/etcd4' branch which has both [the maintenance PR ](https://github.com/xapi-project/xen-api/pull/4939)  and this one, just wanted to separate review/comments of vtpm specific code from general improvements.

edwintorok:
https://github.com/edwintorok/xen-api/compare/private/edvint/maintenance...edwintorok:xen-api:private/edvint/etcd4

ocaml/tests/test_metastore_properties.ml
```diff
@@ -0,0 +1,166 @@
...
+let corner_cases gen lst =
+  Crowbar.(choose [gen; choose @@ List.rev_map const lst])
+
+let int = corner_cases Crowbar.int [min_int; max_int]
+
+let char = corner_cases Crowbar.char ['\x00'; '\xff']
+
+let all_str = String.init 255 Char.chr
+
+let string = corner_cases Crowbar.bytes [all_str]
```
psafont, line 19: shouldn't an empty string be a corner case?

ocaml/tests/test_metastore_properties.ml
```diff
@@ -0,0 +1,166 @@
...
+
+let map =
+  Crowbar.map [Crowbar.list (Crowbar.pair string int)] @@ fun lst ->
+  lst |> List.to_seq |> StringMap.of_seq
+
+let test_stringmap map =
+  test_serialize_deserialize (StringMap.equal Int.equal) typ_of_map map
+
+let () =
+  Crowbar.add_test ~name:"serialize/deserialize" [t] test_serialize ;
```
psafont, line 164: wondering if this shouldn't be part of the rpclib tests as well

ocaml/database/metastore/config_field.ml
```diff
@@ -0,0 +1,74 @@
+type nonrec ('a, 'b) result = ('a, ([> `Msg of string] as 'b)) result
```
psafont, line 1:
```suggestion
type ('a, 'b) result = ('a, ([> `Msg of string] as 'b)) Result.t
```

ocaml/database/metastore/config_field.ml
```diff
@@ -0,0 +1,74 @@
...
+
+let typ_of_path =
+  Serialization.using ~aname:"path" Fpath.to_string Fpath.v
+    Rpc.Types.string.Rpc.Types.ty
+
+type initial_cluster_state = New | Existing [@@deriving rpcty]
+
+type log_output = Stdout | Stderr | Default [@@deriving rpcty]
+
+type level = CRITICAL | ERROR | WARNING | INFO | DEBUG [@@deriving rpcty]
```
psafont, line 43: Can't this be taken from the Logs or xapi-log libraries? It's already being duplicated

ocaml/tests/test_metastore_properties.ml
edwintorok: Rpclib tests do not use crowbar, and crowbar and alcotest do not mix well together (crowbar runs all of its tests 'at_exit', whereas with alcotest you can choose what to run and runs it as part of the main program)

ocaml/database/metastore/config_field.ml
edwintorok: This is the list of values accepted by 'etcd', I'll add a comment

ocaml/database/metastore/config_field.mli
```diff
@@ -0,0 +1,93 @@
...
+
+      @param limit when [None] means unlimited (encoded as 0), otherwise [Some p], where [p > 0].
+
+      [0] is not accepted, because it means unlimited, and explicitly raises an
+  exception to avoid accidentally interpreting a "no values permitted" as
+  "unlimited values are permitted".
+   *)
+
+(** milliseconds, always > 0 *)
+type milliseconds = private int64
```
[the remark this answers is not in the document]
edwintorok: They are intervals. The internals representation needs to match what etcd uses (so milliseconds as integers). We *can* chose to use a different type at the high level though, in the 'constructor' that creates 'milliseconds' types and it might make sense for that to take an 'Mtime.Span.t' (might also avoid confusion between seconds and milliseconds).

ocaml/database/metastore/config_field.ml
```diff
@@ -16,7 +16,8 @@ let positive_or_unlimited = function None -> Ok 0L | Some i64 -> positive i64
 
 type milliseconds = int64 [@@deriving rpcty]
 
-let milliseconds f = f *. 1000. |> Float.round |> Int64.of_float |> positive
+let milliseconds t =
+  t |> Mtime.Span.to_ms |> Float.round |> Int64.of_float |> positive
```
psafont, line 20: No need to call `positive` here, Span.t already ensures the property

edwintorok:
Currently there is a conflict between the dune here and in the maintenance PR, I'll change the one here to avoid the conflict (also the tests will only run once the maintenance PR is merged)

ocaml/database/metastore/serialization.ml
```diff
@@ -0,0 +1,147 @@
...
+      true
+  | _ ->
+      false
+
+let max_filename_component = 255
+(* on Linux *)
+
+let max_filename_length = 4096
+
+let max_path_component =
```
lindig, line 117: I think we need some comments here what this is good for.

ocaml/quicktest/quicktest_etcd.ml
```diff
@@ -0,0 +1,110 @@
...
+module StringMap = Map.Make (String)
+
+let kill_after pid =
+  Thread.delay 0.5 ;
+  try Unix.kill pid Sys.sigterm with Unix.Unix_error (Unix.ESRCH, _, _) -> ()
+
+let read_line ch =
+  match input_line ch with
+  | line ->
+      Some (line, ch)
```
lindig, line 63: Why does this return the channel, which is already provided as argument?

ocaml/quicktest/quicktest_etcd.ml
```diff
@@ -0,0 +1,110 @@
...
+      |> StringMap.bindings
+      |> List.rev_map @@ fun (k, v) -> Printf.sprintf "%s=%s" k v
+      )
+  in
+  Logs.debug (fun m -> m "environment: %a" Fmt.Dump.(array string) env) ;
+  config |> Config.serialize |> Serialization.string_to_file_exn conf ;
+  (* TODO: use forkexec helpers or bos here? *)
+  let cmd, args = (!etcd, []) in
+  let stdout, stdin, stderr =
+    Unix.open_process_full (Filename.quote_command cmd args) env
```
lindig, line 88: I think we should avoid strings here and use a string list or array containing all parameters that don't undergo any shell expansion. Use `Unix.open_process_args_full`?

PR #4941, opened 2023-03-21
CA-374989: Avoid using get_record on cross-pool migration

psafont:
Using them makes it impossible to remove unrelated fields that were supposed to have been removed in 2012.

Using the get_record is what caused to revert the datamodel changes in https://github.com/xapi-project/xen-api/pull/4903

If this is backported to 8.2.1 it will allow to integrate that code again once 8.2.1 is the earliest release supported, otherwise we'll have to wait until the version that includes this fix is the earliest one


robhoes (approved):
Was that the only one?

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -347,21 +347,25 @@ let infer_vgpu_map ~__context ?remote vm =
...
+          XenAPI.VGPU.get_scheduled_to_be_resident_on ~rpc ~session_id
+            ~self:vgpu
+          |> fun self ->
           XenAPI.PGPU.get_PCI ~rpc ~session_id ~self |> fun self ->
           XenAPI.PCI.get_pci_id ~rpc ~session_id ~self
           |> Xenops_interface.Pci.address_of_string
         in
         let vf () =
-          vgpu.API.vGPU_PCI |> fun self ->
+          pci |> fun self ->
```
lindig, line 362: Could just use `~self:pci` below

lindig (approved):

PR #4944, opened 2023-03-23
CP-42011: Config scheduled job to synchronize updates

gangj:
1. add new fields: "update_sync_frequency", "update_sync_day", "update_sync_hour", "update_sync_enabled" in pool datamodel
2. add API: pool.configure_update_sync
3. add xe CLI: pool-configure-update-sync

### DIFF
--- a/ocaml/idl/datamodel_pool.ml
+++ b/ocaml/idl/datamodel_pool.ml
@@ -1046,16 +1046,16 @@ let configure_update_sync =
         (Ref _pool, "self", "The pool")
       ; ( update_sync_frequency
         , "update_sync_frequency"
-        , "of type Enum ('daily', 'weekly', 'monthly') to record the frequency \
-           of the scheduled update synchronization"
+        , "The frequency at which updates are synced from remote CDN: daily, \
+           weekly, or monthly."
         )
       ; ( Int
         , "update_sync_day"
-        , "which day of one period the update sychronization is scheduled"
+        , "Which day of one period the update sychronization is scheduled"
         )
       ; ( Int
         , "update_sync_hour"
-        , "which hour of day the update sychronization is scheduled"
+        , "Which hour of day the update sychronization is scheduled"
         )
       ]
     ~allowed_roles:_R_POOL_OP ()
--- a/ocaml/idl/datamodel_pool.ml
+++ b/ocaml/idl/datamodel_pool.ml
@@ -1024,6 +1024,42 @@ let set_uefi_certificates =
       ]
     ~allowed_roles:_R_POOL_ADMIN ()
 
+let update_sync_frequency =
+  Enum
+    ( "update_sync_frequency"
+    , [
+        ("daily", "Indicates the pool update synchronization is scheduled daily")
+      ; ( "weekly"
+        , "Indicates the pool update synchronization is scheduled weekly"
+        )
+      ; ( "monthly"
+        , "Indicates the pool update synchronization is scheduled monthly"
+        )
+      ]
+    )
+
+let configure_update_sync =
+  call ~name:"configure_update_sync"
+    ~doc:"config scheduled job to sync update from remote CDN" ~lifecycle:[]
+    ~params:
+      [
+        (Ref _pool, "self", "The pool")
+      ; ( update_sync_frequency
+        , "update_sync_frequency"
+        , "of type Enum ('daily', 'weekly', 'monthly') to record the frequency \
+           of the scheduled update synchronization"
+        )
+      ; ( Int
+        , "update_sync_day"
+        , "which day of one period the update sychronization is scheduled"
+        )
+      ; ( Int
+        , "update_sync_hour"
+        , "which hour of day the update sychronization is scheduled"
+        )
+      ]
+    ~allowed_roles:_R_POOL_OP ()
+
 (** A pool class *)
 let t =
   create_obj ~in_db:true ~in_product_since:rel_rio ~in_oss_since:None
@@ -1106,6 +1142,7 @@ let t =
       ; disable_repository_proxy
       ; set_uefi_certificates
       ; set_https_only
+      ; configure_update_sync
       ]
     ~contents:
       ([uid ~in_oss_since:None _pool]
@@ -1334,6 +1371,18 @@ let t =
         ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:DateTime
             ~default_value:(Some (VDateTime Date.epoch)) "last_update_sync"
             "time of the last update sychronization"
+        ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:update_sync_frequency
+            ~default_value:(Some (VEnum "daily")) "update_sync_frequency"
+            "frequency of the scheduled update synchronization"
+        ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:Int "update_sync_day"
+            ~default_value:(Some (VInt 1L))
+            "which day of one period the update sychronization is scheduled"
+        ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:Int "update_sync_hour"
+            ~default_value:(Some (VInt 0L))
+            "which hour of day the update sychronization is scheduled"
+        ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:Bool
+            ~default_value:(Some (VBool false)) "update_sync_enabled"
+            "if scheduled update sychronization is enabled or not"
         ]
       )
     ()
--- a/ocaml/idl/datamodel_pool.ml
+++ b/ocaml/idl/datamodel_pool.ml
@@ -1040,7 +1040,7 @@ let update_sync_frequency =
 
 let configure_update_sync =
   call ~name:"configure_update_sync"
-    ~doc:"config scheduled job to sync update from remote CDN" ~lifecycle:[]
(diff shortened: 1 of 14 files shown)


ocaml/idl/datamodel_pool.ml
```diff
@@ -1024,6 +1024,42 @@ let set_uefi_certificates =
...
+        , "Indicates the pool update synchronization is scheduled weekly"
+        )
+      ; ( "monthly"
+        , "Indicates the pool update synchronization is scheduled monthly"
+        )
+      ]
+    )
+
+let configure_update_sync =
+  call ~name:"configure_update_sync"
```
lindig, line 1042: I would try to simplify this: "The frequency at which information about new updates are downloaded:  daily, weekly, or monthly."

ocaml/idl/datamodel_pool.ml
```diff
@@ -1024,6 +1024,42 @@ let set_uefi_certificates =
...
+    ~doc:"config scheduled job to sync update from remote CDN" ~lifecycle:[]
+    ~params:
+      [
+        (Ref _pool, "self", "The pool")
+      ; ( update_sync_frequency
+        , "update_sync_frequency"
+        , "of type Enum ('daily', 'weekly', 'monthly') to record the frequency \
+           of the scheduled update synchronization"
+        )
+      ; ( Int
```
lindig, line 1052: Making day and hour separate parameters which in addition are restricted in range forces a lot of checking. There is also the problem that the legal day range depends on the frequency: on a weekly basis the day is either 0 to 6 or 1 to 7, but larger on a monthly basis. Would it have been simpler to collapse this all into one string?

* `daily:13:00` (every day at 13:00)
* `weekly:3-13:30` (every week on, Wednesday on 13:30)
* `monthly:15-14:45` (every 15th, on 14:45)
Parsing a string like this and finding illegal strings would be one compact function and a single error.

gangj: @lindig Thanks for the suggestion.
I think this API design was chosen in the design: [Design of Toolstack for CP-41451](https://info.citrite.net/display/xenserver/Design+of+Toolstack+for+CP-41451) is because we can do parameter checking easily.
And with your suggested string style parameter, is there a simple way to tell both the API user and xe CLI user the expected format of this string?

lindig: I thought a bit more about it. I think it's fine as is. Using a single string is an alternative design (a domain-specific format) ; the important part is that all parameters are set in one call and not individually because the legal value ranges depend on each other.

ocaml/idl/datamodel_pool.ml
gangj: I changed the description to: "The frequency at which updates are synced from remote CDN: daily, weekly, or monthly."

ocaml/xapi/xapi_pool.ml
```diff
@@ -3609,3 +3609,55 @@ let set_https_only ~__context ~self:_ ~value =
         )
         (Db.Host.get_all ~__context)
   )
+
+let configure_update_sync ~__context ~self ~update_sync_frequency
+    ~update_sync_day ~update_sync_hour =
+  ( match update_sync_frequency with
+  | `daily when update_sync_day <> 1L ->
```
minglumlu, line 3616: I'm not sure if this is correct that no variables in `when` come from matched pattern.
Following let me feel easy. Would appreciate if anyone could have an elaboration here. 
```
match (update_sync_frequency, update_sync_day) with
| `daily, d when d <> 1L
......
```

ocaml/xapi/xapi_pool.ml
```diff
@@ -3609,3 +3609,55 @@ let set_https_only ~__context ~self:_ ~value =
...
+         an integer out of range: -28 ~ -1, 1 ~ 28" ;
+      raise
+        Api_errors.(
+          Server_error
+            (invalid_update_sync_day, [Int64.to_string update_sync_day])
+        )
+  | _ ->
+      ()
+  ) ;
+  if update_sync_hour < 0L || update_sync_hour > 23L then (
```
minglumlu, line 3650: This can be matched together above.

ocaml/xapi/xapi_pool.ml
minglumlu: And would following look better?
```
let day_of_month = function
  | i when i >=1 and i <= 31 ->
      Some i
  | _ ->
      None
in
let hour_of_day = function
  | i when i >=0 and i <= 23 ->
      Some i
  | _ ->
      None
in
let day_of_week = function
  | i when i >= 1 and i <= 7 ->
      Some i
  | _ ->
      None
in
match (frequency,
       day_of_month(update_sync_day),
       day_of_week(update_sync_day),
       hour_of_day(update_sync_hour)) with
| `daily, _, _, Some _
| `weekly, _, Some _, Some _
| `monthly, Some _, _, Some _ ->
    ()
| _ ->
    raise ...
```

lindig: I think having an explicit check for day_of_month, hour. day_of_week is good. But note that existing code allows for negative numbers, too. A `day_of_month = 31` is problematic. What happens in a month with 30 days? Is this synchronization triggered or not in that situation?  This needs to be clearly documented.

lindig:
Where is the code that decides to run a synchronisation based on the parameters set trough the API?

gangj:
Hi @lindig , it is not included in this PR, it is done in the next ticket: CP-42012, and the PR for CP-42012 will be raised after this PR is merged first.

ocaml/idl/datamodel_pool.ml
```diff
@@ -1024,6 +1024,46 @@ let set_uefi_certificates =
...
+        , "The frequency at which updates are synced from remote CDN: daily, \
+           weekly, or monthly."
+        )
+      ; ( Int
+        , "update_sync_day"
+        , "Which day of one period the update sychronization is scheduled. For \
+           'daily' schedule, it should be 1. For 'weekly' schedule, -7..-1 and \
+           1..7, where 1 and -7 is Sunday. For 'monthly' schedule, -28..-1 and \
+           1..28, this ensure that the day exist in every month. For negative \
+           number, it means the day count down from the end of the period."
```
lindig, line 1058: For a month with 30 days, what does `-1` mean? I could read this as `30-1=29` so synchronisation happens on the 29th day. Is this how this works or is the documentation inaccurate? How would you make it run on the 30th day?

ocaml/idl/datamodel_pool.ml
```diff
@@ -1024,6 +1024,46 @@ let set_uefi_certificates =
...
+        )
+      ; ( Int
+        , "update_sync_hour"
+        , "Which hour of day the update sychronization is scheduled, 0..23"
```
lindig, line 1062: So: days in week and in month are 1-based but hours are 0-based.

lindig (approved):

ocaml/idl/datamodel_pool.ml
gangj: Yes, I set "update_sync_hour" to 0-based, it is the same as the clock hour, and "update_sync_day" for `monthly` to 1-based, it is the same as calendar day, and "update_sync_day" for `weekly` to 1-based to align with `monthly`.

But the week day in Unix is 0-based (https://v2.ocaml.org/api/Unix.html):
tm_wday : int;   Day of week (Sunday is 0)
I am a little hesitate if 0-based is better for day for `weekly`.

What is your suggestion here?

ocaml/idl/datamodel_pool.ml
gangj: The meaning of negative number is following python way which is suggested by Rob: -1 means the last day of that period, so for a month with 30 days, `-1` means the last day of that month: day 30.

lindig: Python is 0-based: for an array of size 30, you get 30-1=29 which is the last entry in the array. Using this with a 1-based system is not as natural. Your explanation in the documentation "For negative number, it means the day count down from the end of the period." does not match the intended semantics.

gangj: I think you are right! Then how about changing the `update_sync_day` to 0-based. @robhoes @minglumlu are you happy with this?

And for the comment: "For negative number, it means the day count down from the end of the period.", I am not quiet sure how to express it correctly and briefly, will this be ok:

`For negative number, it means counting down from the end of the period, which follows Python style, For example, -1 with 'monthly' schedule means the last day of the month.`

minglumlu: I think Python uses `1` for the first day of a month (at least from this https://docs.python.org/3/library/datetime.html#datetime.date.day).
I don't think the `0th` day of a month looks good. Every time a user who wants to set to a day has to do a decreasing.

The start day of a week should be clarified in API/CLI document as Monday. But in this case, the number starting from either 0 or 1 can work.

I don't think we need to keep same base for all of month, week, hour. Naturally, the base day of month is 1, and the base hour of day is 0. The base day of week can be either 0 or 1.

minglumlu (approved):

ocaml/idl/datamodel_pool.ml
gangj: Thanks @minglumlu , one thing to note is that: the start day of a week is Sunday.
And as the PR is already merged, will clarify it in API/CLI document in the next PR.

PR #4946, opened 2023-03-23
Toolstack Tracing v1

mg12:
Creating this PR as an early form of Request for Comments, in order to obtain feedback as soon as possible.

PLEASE DO NOT MERGE UNTIL AT LEAST THE FOLLOWING FIXUPS ARE ADDED:
1) tracing starts disabled by default
2) tracing does not add too many lines to debug output

----
This PR contains an initial implementation for xen-api of the distributed tracing concepts in https://opentelemetry.io/docs/reference/specification/trace/api/

It's split in two main parts:

a) ocaml/libs/tracing/tracing.ml: minimal library implementation of the API defined in the link above.
- TracerProvider: configuration for status (enabled/disabled) and endpoints (disk/network)
- Span: abstracts an action with start and finish events
- Export: Content format manipulation and Destination output helpers for files and http endpoints. These helpers pull the finished spans and move them over to the destination endpoints indicated in the TracerProvider. The initial content format present is zipkinv2, which is compatible with a number of different distributed tracing endpoints. It should be possible to produce content also in the OTEL format using a content producer library such as ocaml-opentelemetry. 

b) all the other files: initial xen-api instrumentation using tracing.ml to produce distributed tracing content to the supported destination endpoints.

--
Example usage:

1) install Jaeger using something like:
```
docker run -d --name jaeger   -e COLLECTOR_ZIPKIN_HTTP_PORT=9411   -p 5775:5775/udp   -p 6831:6831/udp   -p 6832:6832/udp   -p 5778:5778   -p 16686:16686   -p 14268:14268   -p 9411:9411   jaegertracing/all-in-one:1.6
```
2) set the destination endpoint to point to (1):
```
   xe tracing-param-set uuid=<tracing-provider-uuid> endpoints=http://<my-jaeger-ip>:9411/api/v2/spans
```
3) enable the tracer provider:
```
   xe tracing-param-set uuid=<tracing-provider-uuid> status=enabled
```
and you should be able to see the spans collected by XAPI in the Jaeger UI at `http://<my-jaeger-ip>:16686/`

### DIFF
--- a/ocaml/libs/tracing/tracing.ml
+++ b/ocaml/libs/tracing/tracing.ml
@@ -1,0 +1,142 @@
+(*
+ * Copyright (C) 2023 Cloud Software Group
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+type config = Undetermined | Disabled | Url of string
+
+let tracing_config = ref Undetermined
+
+let url_file = "/etc/xapi-tracing-url"
+
+module Http = struct
+  module Request = Cohttp.Request.Make (Cohttp_posix_io.Buffered_IO)
+  module Response = Cohttp.Response.Make (Cohttp_posix_io.Buffered_IO)
+
+  let get_user_agent () = Sys.argv.(0)
+
+  let post ~url ~route ~body : (string, exn) result =
+    try
+      let uri = Printf.sprintf "%s%s" url route in
+      let uri' = Uri.of_string uri in
+      Open_uri.with_open_uri uri' (fun fd ->
+          let headers =
+            Cohttp.Header.of_list
+              [
+                ("User-agent", get_user_agent ())
+              ; ("content-type", "application/body")
+              ; ("content-length", string_of_int (String.length body))
+              ]
+          in
+          let request =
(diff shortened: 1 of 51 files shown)


edwintorok:
I've converted the PR to draft given the 'do not merge' request.

edwintorok:
Tip: you can use 'git commit --fixup <commit-id>' to add fixup commits with the right syntax such that git rebase will automatically squash the fixup commit into the correct commit.

ocaml/libs/tracing/tracing.ml
```diff
@@ -491,13 +491,19 @@ module Export = struct
...
+          in
+          let unix_time = timestamp |> Unix.localtime in
           let date =
-            Printf.sprintf "%d%d%d-%d%d%d-%s" unix_time.tm_year unix_time.tm_mon
-              unix_time.tm_mday unix_time.tm_hour unix_time.tm_min
-              unix_time.tm_sec microsec
+            Printf.sprintf "%d%02d%d-%d%d%d-%s"
+              (Int.rem unix_time.tm_year 100)
+              (unix_time.tm_mon + 1) unix_time.tm_mday unix_time.tm_hour
+              unix_time.tm_min unix_time.tm_sec microsec
```
psafont: This is really complex, why can't Ptime be used here? something like:
```suggestion
          let date = Ptime_clock.now () |> Ptime.to_rfc3339 ~frac_int:6
```
We already have the [Date](http://xapi-project.github.io/stdext/xapi-stdext-date/Xapi_stdext_date/Date/index.html) module which deals with dates, I'd rather have it there, or encapsulate it better so it's associated with serialization of some specific format

psafont:
This is very hard to review there is a lot of churn, meaning that a lot of code that gets introduced in a commit gets completely replaced in following commits. there is also very little explanation in the commit messages. A PR that add the end result constructively without all the churn would be much easier to review and take a lot less collective time

jameshensmancitrix (commented):
We could move the changes to Span into the previous commit. For traceparent I changed trace_id to be 32 chars, this would be good to have the change here to.

psafont:
Closing because it got superceded with #4958

PR #4949, opened 2023-03-24
Update API lifecycle data

lindig:


psafont (approved):

robhoes (approved):

psafont:
I was thinking on how to automate this commit to avoid opening these.

Maybe on the release workflow, add a step that detects that the tagged commit is the HEAD of master, and another one that only runs when that is true, creates a new one commit with this change.

PR #4951, opened 2023-03-24
Toolstack Tracing V1

jameshensmancitrix:
This PR contains an initial implementation for xen-api of the distributed tracing concepts in https://opentelemetry.io/docs/reference/specification/trace/api/

It's split in two main parts:

a) ocaml/libs/tracing/tracing.ml: minimal library implementation of the API defined in the link above.

TracerProvider: configuration for status (enabled/disabled) and endpoints (disk/network)
Span: abstracts an action with start and finish events
Export: Content format manipulation and Destination output helpers for files and http endpoints. These helpers pull the finished spans and move them over to the destination endpoints indicated in the TracerProvider. The initial content format present is zipkinv2, which is compatible with a number of different distributed tracing endpoints. It should be possible to produce content also in the OTEL format using a content producer library such as ocaml-opentelemetry.
b) all the other files: initial xen-api instrumentation using tracing.ml to produce distributed tracing content to the supported destination endpoints.

--
Example usage:

install Jaeger using something like:
`docker run -d --name jaeger   -e COLLECTOR_ZIPKIN_HTTP_PORT=9411   -p 5775:5775/udp   -p 6831:6831/udp   -p 6832:6832/udp   -p 5778:5778   -p 16686:16686   -p 14268:14268   -p 9411:9411   jaegertracing/all-in-one:1.6`
set the destination endpoint to point to (1):
`   xe tracing-param-set uuid=<tracing-provider-uuid> endpoints=http://<my-jaeger-ip>:9411/api/v2/spans`
enable the tracer provider:
`   xe tracing-param-set uuid=<tracing-provider-uuid> status=enabled`
and you should be able to see the spans collected by XAPI in the Jaeger UI at http://<my-jaeger-ip>:16686/

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -7732,6 +7732,124 @@ module VUSB = struct
       ~messages:[create; unplug; destroy] ()
 end
 
+module Tracing = struct
+  let lifecycle = []
+
+  let set_status =
+    call ~name:"set_status" ~in_oss_since:None ~lifecycle
+      ~doc:"Set the fields of a TracerProvider in the Tracing Library"
+      ~params:
+        [
+          (Ref _tracing, "self", "The TracerProvider")
+        ; (Bool, "status", "The status the TracerProvider is to be set to")
+        ]
+      ~allowed_roles:_R_POOL_ADMIN ()
+
+  let set_tags =
+    call ~name:"set_tags" ~in_oss_since:None ~lifecycle
+      ~doc:"Set the fields of a TracerProvider in the Tracing Library"
+      ~params:
+        [
+          (Ref _tracing, "self", "The TracerProvider")
+        ; ( Map (String, String)
+          , "tags"
+          , "The tags the TracerProvider is to be set to"
+          )
+        ]
+      ~allowed_roles:_R_POOL_ADMIN ()
+
+  let set_endpoints =
+    call ~name:"set_endpoints" ~in_oss_since:None ~lifecycle
+      ~doc:"Set the fields of a TracerProvider in the Tracing Library"
+      ~params:
+        [
+          (Ref _tracing, "self", "The TracerProvider")
+        ; ( Set String
+          , "endpoints"
+          , "The endpoints the TracerProvider is to be set to"
+          )
+        ]
+      ~allowed_roles:_R_POOL_ADMIN ()
+
+  let set_components =
+    call ~name:"set_components" ~in_oss_since:None ~lifecycle
+      ~doc:"Set the fields of a TracerProvider in the Tracing Library"
+      ~params:
+        [
+          (Ref _tracing, "self", "The TracerProvider")
+        ; ( Set String
+          , "components"
+          , "The components the TracerProvider is to be set to"
+          )
+        ]
+      ~allowed_roles:_R_POOL_ADMIN ()
+
+  let set_filters =
+    call ~name:"set_filters" ~in_oss_since:None ~lifecycle
+      ~doc:"Set the fields of a TracerProvider in the Tracing Library"
+      ~params:
+        [
+          (Ref _tracing, "self", "The TracerProvider")
+        ; ( Set String
+          , "filters"
+          , "The filter the TracerProvider is to be set to"
+          )
+        ]
+      ~allowed_roles:_R_POOL_ADMIN ()
+
+  let set_processors =
+    call ~name:"set_processors" ~in_oss_since:None ~lifecycle
+      ~doc:"Set the fields of a TracerProvider in the Tracing Library"
+      ~params:
+        [
+          (Ref _tracing, "self", "The TracerProvider")
+        ; ( Set String
+          , "processors"
+          , "The processors the TracerProvider is to be set to"
+          )
+        ]
+      ~allowed_roles:_R_POOL_ADMIN ()
+
+  let t =
+    create_obj ~name:_tracing ~descr:"Describes a TracerProvider"
+      ~doccomments:[] ~gen_constructor_destructor:false ~gen_events:true
+      ~in_db:true ~lifecycle ~persist:PersistEverything ~in_oss_since:None
+      ~messages_default_allowed_roles:_R_POOL_ADMIN
+      ~contents:
+        ([uid _tracing ~lifecycle]
+        @ [
+            field ~qualifier:DynamicRO ~ty:(Set (Ref _host)) ~lifecycle "hosts"
+              "A list of hosts the TracerProvider is to be registered on"
+              ~ignore_foreign_key:true
+          ; field ~qualifier:DynamicRO ~ty:String ~lifecycle "name_label"
+              "A user-assinged label for a TracerProvider"
+          ; field ~qualifier:DynamicRO
+              ~ty:(Map (String, String))
+              ~lifecycle "tags"
+              "Tags to be set on spans to provide extra information"
+          ; field ~qualifier:DynamicRO ~ty:(Set String) ~lifecycle "endpoints"
+              "Endpoints where spans are exported to"
+          ; field ~qualifier:DynamicRO ~ty:(Set String) ~lifecycle "components"
+              "Componenets the TraverProvider is to be registered on"
+          ; field ~qualifier:DynamicRO ~ty:(Set String) ~lifecycle "filters"
+              "Filters for operations that spans will be produced for"
+          ; field ~qualifier:DynamicRO ~ty:(Set String) ~lifecycle "processors"
+              "Processors that will filter/map spans before being exported"
+          ; field ~qualifier:DynamicRO ~ty:Bool ~lifecycle "status"
+              "The status of the TracerProvider"
+          ]
+        )
+      ~messages:
+        [
+          set_status
+        ; set_tags
+        ; set_endpoints
+        ; set_components
+        ; set_filters
+        ; set_processors
+        ]
+      ()
+end
 (******************************************************************************************)
 
 (** All the objects in the system in order they will appear in documentation: *)
@@ -7803,6 +7921,7 @@ let all_system =
   ; PUSB.t
   ; USB_group.t
   ; VUSB.t
+  ; Tracing.t
   ; Datamodel_cluster.t
   ; Datamodel_cluster_host.t
   ; Datamodel_certificate.t
@@ -8040,6 +8159,7 @@ let expose_get_all_messages_for =
   ; _certificate
   ; _repository
   ; _vtpm
+  ; _tracing
   ]
 
 let no_task_id_for = [_task; (* _alert; *) _event]
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -7732,124 +7732,6 @@ module VUSB = struct
       ~messages:[create; unplug; destroy] ()
 end
 
-module Tracing = struct
-  let lifecycle = []
-
-  let set_status =
-    call ~name:"set_status" ~in_oss_since:None ~lifecycle
-      ~doc:"Set the fields of a TracerProvider in the Tracing Library"
-      ~params:
-        [
-          (Ref _tracing, "self", "The TracerProvider")
-        ; (Bool, "status", "The status the TracerProvider is to be set to")
-        ]
-      ~allowed_roles:_R_POOL_ADMIN ()
-
-  let set_tags =
-    call ~name:"set_tags" ~in_oss_since:None ~lifecycle
-      ~doc:"Set the fields of a TracerProvider in the Tracing Library"
-      ~params:
-        [
-          (Ref _tracing, "self", "The TracerProvider")
-        ; ( Map (String, String)
-          , "tags"
-          , "The tags the TracerProvider is to be set to"
-          )
-        ]
-      ~allowed_roles:_R_POOL_ADMIN ()
-
-  let set_endpoints =
-    call ~name:"set_endpoints" ~in_oss_since:None ~lifecycle
-      ~doc:"Set the fields of a TracerProvider in the Tracing Library"
-      ~params:
-        [
-          (Ref _tracing, "self", "The TracerProvider")
-        ; ( Set String
-          , "endpoints"
-          , "The endpoints the TracerProvider is to be set to"
-          )
-        ]
-      ~allowed_roles:_R_POOL_ADMIN ()
-
-  let set_components =
-    call ~name:"set_components" ~in_oss_since:None ~lifecycle
-      ~doc:"Set the fields of a TracerProvider in the Tracing Library"
-      ~params:
-        [
-          (Ref _tracing, "self", "The TracerProvider")
-        ; ( Set String
-          , "components"
-          , "The components the TracerProvider is to be set to"
-          )
-        ]
-      ~allowed_roles:_R_POOL_ADMIN ()
-
-  let set_filters =
-    call ~name:"set_filters" ~in_oss_since:None ~lifecycle
-      ~doc:"Set the fields of a TracerProvider in the Tracing Library"
-      ~params:
-        [
-          (Ref _tracing, "self", "The TracerProvider")
-        ; ( Set String
-          , "filters"
-          , "The filter the TracerProvider is to be set to"
-          )
-        ]
-      ~allowed_roles:_R_POOL_ADMIN ()
-
-  let set_processors =
-    call ~name:"set_processors" ~in_oss_since:None ~lifecycle
-      ~doc:"Set the fields of a TracerProvider in the Tracing Library"
-      ~params:
-        [
-          (Ref _tracing, "self", "The TracerProvider")
-        ; ( Set String
-          , "processors"
-          , "The processors the TracerProvider is to be set to"
-          )
-        ]
-      ~allowed_roles:_R_POOL_ADMIN ()
-
-  let t =
-    create_obj ~name:_tracing ~descr:"Describes a TracerProvider"
-      ~doccomments:[] ~gen_constructor_destructor:false ~gen_events:true
-      ~in_db:true ~lifecycle ~persist:PersistEverything ~in_oss_since:None
-      ~messages_default_allowed_roles:_R_POOL_ADMIN
-      ~contents:
-        ([uid _tracing ~lifecycle]
-        @ [
-            field ~qualifier:DynamicRO ~ty:(Set (Ref _host)) ~lifecycle "hosts"
-              "A list of hosts the TracerProvider is to be registered on"
-              ~ignore_foreign_key:true
-          ; field ~qualifier:DynamicRO ~ty:String ~lifecycle "name_label"
-              "A user-assinged label for a TracerProvider"
-          ; field ~qualifier:DynamicRO
-              ~ty:(Map (String, String))
-              ~lifecycle "tags"
-              "Tags to be set on spans to provide extra information"
-          ; field ~qualifier:DynamicRO ~ty:(Set String) ~lifecycle "endpoints"
-              "Endpoints where spans are exported to"
-          ; field ~qualifier:DynamicRO ~ty:(Set String) ~lifecycle "components"
-              "Componenets the TraverProvider is to be registered on"
-          ; field ~qualifier:DynamicRO ~ty:(Set String) ~lifecycle "filters"
-              "Filters for operations that spans will be produced for"
-          ; field ~qualifier:DynamicRO ~ty:(Set String) ~lifecycle "processors"
-              "Processors that will filter/map spans before being exported"
-          ; field ~qualifier:DynamicRO ~ty:Bool ~lifecycle "status"
-              "The status of the TracerProvider"
-          ]
-        )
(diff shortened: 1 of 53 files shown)


ocaml/libs/tracing/tracing.mli
```diff
@@ -0,0 +1,112 @@
...
+  val get_default_tracer : name:string -> Tracer.t
+end
+
+type blob = Span.t
+
+type t = blob option
+
+val empty : t
+
+val null : t -> bool
```
lindig, line 96: `is_empty` would make the connection with `empty` clear beyond the type.

ocaml/libs/tracing/tracing.mli
```diff
@@ -0,0 +1,112 @@
...
+
+val t_of_string : string -> t
+
+val string_of_t : t -> string
```
lindig, line 100: Calling this `to_string` would be more conventional.

ocaml/libs/tracing/tracing.mli
```diff
@@ -0,0 +1,112 @@
...
+
+val json_of_t : t -> string option
```
lindig, line 102: It is not obvious why this ever can be None. Is JSON's Null not the right value?

ocaml/xapi-idl/lib/task_server.ml
```diff
@@ -126,12 +127,22 @@ functor
 
     (* [add dbg f] creates a fresh [t], registers and returns it *)
     let add tasks dbg (f : task_handle -> Interface.Task.async_result option) =
+      let dbg, tracing =
+        match String.index_opt dbg '\n' with
```
lindig, line 131: If this is splitting some string, using Astring functions would be more obvious. Another option is to use sscanf.

ocaml/xapi/storage_smapiv1_wrapper.ml
```diff
@@ -359,6 +359,27 @@ module Everything = struct
     Errors.errors := h.errors
 end
 
+let with_tracing ~name ~dbg f =
+  let dbg, parent =
+    match String.index_opt dbg '\n' with
```
lindig, line 364: Same as above: use string-splitting functions

ocaml/xs-trace/test/test-xs-trace.sh
```diff
@@ -0,0 +1,17 @@
+set -eux
+
+export PORT=9411
+export HOST="127.0.0.1"
```
lindig, line 4: This should have a `#!` line to make clear what shell is expected.

ocaml/xs-trace/xs_trace.ml
```diff
@@ -0,0 +1,35 @@
...
+          ()
+      | Error err ->
+          Printf.eprintf "Error: %s" (Printexc.to_string err) ;
+          exit 1
+  in
+  export_file origin ;
+  let rec remove_all orig =
+    if Sys.is_directory orig then (
+      let files = Sys.readdir orig in
+      let file_paths = Array.map (fun file -> orig ^ "/" ^ file) files in
```
lindig, line 28: Use `Filename.concat`

ocaml/xapi-idl/lib/task_server.ml
jameshensmancitrix: split_on_char could also be used here and pattern match.

robhoes: Hmm... I put the tracing data in the `dbg` string for the hackathon project as a quick hack, which we would consider replacing with proper types when "productising". Either by adding a new function parameter for the tracing data to every function, or changing the type of `dbg` to a tuple or record. This may or may not be easy :) but worth thinking about.

ocaml/xs-trace/xs_trace.ml
psafont: I like to use it this way:
```ocaml
let ( // ) = Filename.concat in
filepath = "/root" // "a" // "path" // "file.ext"
```

ocaml/libs/tracing/tracing.mli
jameshensmancitrix: Because t can be None, would `{}` be a better default? I think we could condense to_string and to_json.

jameshensmancitrix:
> 7a37d01

I noticed this just now, I was previously having issues with runtest (the shebang fixed) so I just clocked that a minute before the comment. Created the fixup

ocaml/libs/tracing/tracing.mli
lindig: I would use JSON's `null`

lindig (approved):

ocaml/xapi/xapi_globs.ml
```diff
@@ -1195,6 +1201,23 @@ let other_options =
       (fun s -> s)
       (fun s -> s)
       disable_dbsync_for
+  ; gen_list_option "tracing-default-endpoints"
+      "space-separated list of endpoints strings for the default tracer \
```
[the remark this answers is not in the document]
jameshensmancitrix: The problem of ctrl+d for editing

ocaml/xapi/context.ml
```diff
@@ -148,6 +148,13 @@ let __destroy_task : (__context:t -> API.ref_task -> unit) ref =
 
 let string_of_task __context = __context.dbg
 
+let string_of_task_and_tracing __context =
+  match Tracing.t_to_string_opt __context.tracing with
+  | None ->
+      __context.dbg
+  | Some tracing ->
+      __context.dbg ^ "\x00" ^ tracing
```
psafont, line 156: Is there a plan to replace `string_of_task` with this function in all situations? I'd try to use the final name instead so the namechange happens only once. I think `to_string` would a good candidate

ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -1864,10 +1864,17 @@ let rec perform_atomic ~progress_callback ?subtask:_ ?result (op : atomic)
           (Xenops_task.id_of_handle t)
           (List.length atoms) description
       in
+      let parallel_id_with_tracing =
+        match Xenops_task.tracing t with
+        | Some t ->
+            parallel_id ^ "\x00" ^ t
```
psafont, line 1870: Can we put this in a function in xenops_task? this is leaking an implementation detail.
Using `Xenops_task.( parallel_id // id)`, similar to how FIlenmae.concat is used would be quite understandable

ocaml/xapi/context.ml
jameshensmancitrix: This came up last Tuesday, There is no clear iea yet on the best way to approach getting tracing into xenopsd, changing the type or adding parameter would require a lot of changes. It terms of naming, I see your point this name is convoluted.

ocaml/xapi/storage_smapiv1_wrapper.ml
```diff
@@ -359,6 +359,25 @@ module Everything = struct
     Errors.errors := h.errors
 end
 
+let with_tracing ~name ~dbg f =
+  let dbg, parent =
+    match String.index_opt dbg '\x00' with
```
psafont, line 364: Similar to the point in xenopsd, this operation here should not depend on an internal representation detail.
Please add a function in Context so this function can use something like: `Context.tracing_parent_of_dbg dbg`

ocaml/xapi/storage_smapiv1_wrapper.ml
```diff
@@ -453,11 +472,12 @@ functor
         in
         List.fold_left perform_one vdi_t ops
 
-      let perform_nolock context ~dbg ~dp ~sr ~vdi ~vm this_op =
+      let perform_nolock context ~dbg ~name ~dp ~sr ~vdi ~vm this_op =
```
edwintorok, line 477: I think I see what your question was @jameshensmancitrix about this function missing in Storage_mux, and I think missing this function is fine, since it is called by the higher level epoch_begin/etc. functions below, the perform function is just a shared implementation of the API, but not exposed on the surface level of the API.

edwintorok: yeah add tracing just for the public APIs that exist in Storage_mux. Is there anything else apart from `perform_nolock` that is missing from Storage_mux?

ocaml/xenopsd/lib/xenops_server.ml
jameshensmancitrix: I can do that?

ocaml/xs-trace/xs_trace.ml
```diff
@@ -0,0 +1,35 @@
+let _ =
+  if Array.length Sys.argv <> 4 then (
+    Printf.eprintf "Usage: %s cp/mv <origin> <destination>\n" Sys.argv.(0) ;
+    exit 1
+  ) ;
+  let action = Sys.argv.(1) in
+  let origin = Sys.argv.(2) in
+  let url = Sys.argv.(3) in
```
psafont, line 8: Parameter parsing is missing, currently it's not resistant to typos.
On top of that, allowing to _delete_ any directory in the system is really dangerous.
There are at least two options:
- Hardcode the root of the path that can be written to (and deleted)
- Use a list of allowed roots that can be used, the directories themselves cannot be erased
- 
```suggestion
  let action = match Sys.argv.(1) with | "mv" -> `mv | "cp" -> `cp | _ -> error_with_message usage in
  let origin =
    let param = Sys.argv.(2) in
    if ".." in param then invalid_argument "not allowed"
    Filename.concat root param
  in
```

The url can be parsed, but if it fails it worries me less as it will be very evident.

ocaml/xenopsd/lib/xenops_server.ml
psafont: I don't see why not, just implement the function in xenops_task as:
```ocaml
let ( // ) parent id = parent ^ "\x00" ^ id
```

ocaml/idl/datamodel.ml
```diff
@@ -7732,6 +7732,124 @@ module VUSB = struct
       ~messages:[create; unplug; destroy] ()
 end
 
+module Tracing = struct
```
psafont, line 7735: Please move this module to a separate file (`ocaml/idl/datamodel/datamodel_tracing.ml`)

ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -1,4 +1,6 @@
 let prototyped_of_class = function
+  | "Tracing" ->
+      Some "23.9.0"
```
psafont, line 3: This version has been released, this file needs to be checked out and regenerated

ocaml/xapi-cli-server/records.ml
```diff
@@ -5242,3 +5242,119 @@ let vtpm_record rpc session_id vtpm =
...
+      ; make_field ~name:"name_label"
+          ~get:(fun () -> (x ()).API.tracing_name_label)
+          ()
+      ; make_field ~name:"tags"
+          ~get:(fun () ->
+            List.map
+              (fun (k, v) -> Printf.sprintf "%s : %s" k v)
+              (x ()).API.tracing_tags
+            |> String.concat ","
+            |> fun s -> if s = "" then "none" else s
```
psafont, line 5285: This doesn't follow the convention elsewhere. Simply use the empty list when there are no values

ocaml/xapi-cli-server/records.ml
```diff
@@ -5242,3 +5242,119 @@ let vtpm_record rpc session_id vtpm =
...
+            |> fun s -> if s = "" then "all" else ""
+          )
+          ()
+      ; make_field ~name:"name_label"
+          ~get:(fun () -> (x ()).API.tracing_name_label)
+          ()
+      ; make_field ~name:"tags"
+          ~get:(fun () ->
+            List.map
+              (fun (k, v) -> Printf.sprintf "%s : %s" k v)
```
psafont, line 5282: Use Record_util.s2sm_to_string instead

ocaml/xapi-cli-server/records.ml
```diff
@@ -5242,3 +5242,119 @@ let vtpm_record rpc session_id vtpm =
...
+          ~get:(fun () ->
+            List.map get_uuid_from_ref (x ()).API.tracing_hosts
+            |> String.concat ","
+            |> fun s -> if s = "" then "all" else ""
+          )
+          ()
+      ; make_field ~name:"name_label"
+          ~get:(fun () -> (x ()).API.tracing_name_label)
+          ()
+      ; make_field ~name:"tags"
```
psafont, line 5279: This is a map, I recommend to follow how other maps are implemented in here. See the field type to see the interesting fields to define. These are get_map, add_to_map, remove_from_map, set_in_map, set_map and clear_map instead of defining a custom set, but I doubt all of them are needed

ocaml/xapi-cli-server/records.ml
```diff
@@ -5242,3 +5242,119 @@ let vtpm_record rpc session_id vtpm =
...
+              ~endpoints
+          )
+          ()
+      ; make_field ~name:"components"
+          ~get:(fun () ->
+            (x ()).API.tracing_components |> String.concat ", " |> fun s ->
+            if s = "" then "all" else s
+          )
+          ~set:(fun s ->
+            String.split_on_char ',' s |> List.filter (( <> ) "all")
```
psafont, line 5320: Define the `_set` fields instead, similar to above with map

ocaml/xapi-cli-server/records.ml
```diff
@@ -5242,3 +5242,119 @@ let vtpm_record rpc session_id vtpm =
...
+            |> fun components ->
+            Client.Tracing.set_components ~rpc ~session_id ~self:tracing
+              ~components
+          )
+          ()
+      ; make_field ~name:"filters"
```
psafont, line 5326: Define the _set fields instead, similar to above with map

ocaml/xapi-cli-server/records.ml
```diff
@@ -5242,3 +5242,119 @@ let vtpm_record rpc session_id vtpm =
...
+          )
+          ~set:(fun s ->
+            String.split_on_char ',' s |> List.filter (( <> ) "all")
+            |> fun filters ->
+            Client.Tracing.set_filters ~rpc ~session_id ~self:tracing ~filters
+          )
+          ()
+      ; make_field ~name:"processors"
+          ~get:(fun () ->
+            (x ()).API.tracing_processors |> String.concat ", " |> fun s ->
```
psafont, line 5339: Define the _set fields instead, similar to above with map

ocaml/xapi-cli-server/records.ml
```diff
@@ -5242,3 +5242,119 @@ let vtpm_record rpc session_id vtpm =
...
+            String.split_on_char ',' s |> List.filter (( <> ) "none")
+            |> fun processors ->
+            Client.Tracing.set_processors ~rpc ~session_id ~self:tracing
+              ~processors
+          )
+          ()
+      ; make_field ~name:"status"
+          ~get:(fun () ->
+            (x ()).API.tracing_status |> fun b ->
+            if b then "enabled" else "disabled"
```
psafont, line 5352: Is this how other booleans are implemented? Please try to follow conventions used in the records.ml module

ocaml/xapi-cli-server/records.ml
```diff
@@ -5242,3 +5242,119 @@ let vtpm_record rpc session_id vtpm =
...
+                   | [k; v] ->
+                       Some (k, v)
+                   | _ ->
+                       None
+               )
+            |> fun tags ->
+            Client.Tracing.set_tags ~rpc ~session_id ~self:tracing ~tags
+          )
+          ()
+      ; make_field ~name:"endpoints"
```
psafont, line 5302: Define the _set fields instead, similar to above with map

ocaml/xapi/message_forwarding.ml
```diff
@@ -6484,4 +6492,105 @@ functor
...
+            hosts ;
+          List.iter (fun _ -> local_fn ~__context) local
+        ) else
+          local_fn ~__context ;
+        ()
+
+      let set_status ~__context ~self ~status =
+        info "Tracing.set_status: self=%s status=%B"
+          (tracing_uuid ~__context self)
+          status ;
```
psafont, line 6523: Using __FUNCTION__ makes these loglines more resilient against errors in copy-pastes, renamings and typos in general
```suggestion
        info "%s: self=%s status=%B" __FUNCTION__
          (tracing_uuid ~__context self)
          status ;
```

ocaml/xapi/message_forwarding.ml
```diff
@@ -6484,4 +6492,105 @@ functor
...
+          (List.map (fun (k, v) -> k ^ ":" ^ v) tags |> String.concat ",") ;
+        let local_fn = Local.Tracing.set_tags ~self ~tags in
+        let client_fn session_id rpc =
+          Client.Tracing.set_tags ~rpc ~session_id ~self ~tags
+        in
+        let db_fn () = Db.Tracing.set_tags ~__context ~self ~value:tags in
+        do_trace_op_on ~__context ~self ~local_fn ~client_fn ~db_fn
+
+      let set_endpoints ~__context ~self ~endpoints =
+        info "Tracing.set_endpoints: self=%s endpoints=%s"
```
psafont, line 6547: URLS can contain secrets, I don't think it's a good idea to log them

ocaml/idl/datamodel_lifecycle.ml
jameshensmancitrix: How would I do this?

ocaml/xapi-cli-server/records.ml
jameshensmancitrix: I can do that, it makes sense if it is converntion

ocaml/idl/datamodel_lifecycle.ml
psafont: This file is regenerated whenever `make` is run

ocaml/xapi-cli-server/records.ml
psafont: This is for tags, which is modelled as a map

ocaml/idl/datamodel.ml
robhoes: Does it make more sense to call the class `Tracer` actually? The classes are usually names of "things", like "Host", "VM" etc.

ocaml/xapi-cli-server/records.ml
robhoes: I think a better name for the field in the datamodel is `enabled`. A "status" of true or false doesn't mean much to me. And then in the CLI it is just `enabled=true|false`.

ocaml/idl/datamodel.ml
```diff
@@ -7732,6 +7732,124 @@ module VUSB = struct
...
+              "Filters for operations that spans will be produced for"
+          ; field ~qualifier:DynamicRO ~ty:(Set String) ~lifecycle "processors"
+              "Processors that will filter/map spans before being exported"
+          ; field ~qualifier:DynamicRO ~ty:Bool ~lifecycle "status"
+              "The status of the TracerProvider"
+          ]
+        )
+      ~messages:
+        [
+          set_status
```
robhoes, line 7844: There is no constructor at all?

ocaml/xapi/storage_smapiv1_wrapper.ml
jameshensmancitrix: No

ocaml/idl/datamodel.ml
robhoes: You somewhere to start though. Are you pre-creating a Tracing object somewhere? I thought we'd have no tracing by default, and then if you want tracing, you create a Tracing object with the relevant parameters such as the endpoint etc.

ocaml/xapi/context.ml
robhoes: > I think to_string would a good candidate

It's not a string of the whole context though. Just of the task and tracing info in the context.

ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -1862,10 +1862,11 @@ let rec perform_atomic ~progress_callback ?subtask:_ ?result (op : atomic)
           (Xenops_task.id_of_handle t)
           (List.length atoms) description
       in
+      let with_tracing = parallel_id_with_tracing parallel_id t in
```
robhoes, line 1867: This name is confusing, as `with_x` normally is a function, while this is a string.

ocaml/xs-trace/test/test-xs-trace.sh
```diff
@@ -0,0 +1,14 @@
...
+
+set -eux
+
+export PORT=9411
+export HOST="127.0.0.1"
+
+./test_xs_trace.exe && ../xs_trace.exe cp test-source.json http://$HOST:$PORT/api/v2/spans
+
+
+if [ $(diff test-source.json test-http-server.out | grep "<") ]
```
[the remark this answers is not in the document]
mg12: @snwoods ^

psafont:
Closing because it got superceded with #4966

PR #4952, opened 2023-03-27
xapi-idl: define PVH memory model and use it

psafont:
PVH needs some additional memory compared to HVM. Since it doesn't use a
shim, do not use PV-in-PVH memory model.


ocaml/xapi-idl/memory/memory.ml
```diff
@@ -169,6 +169,16 @@ module Linux_memory_model_data : MEMORY_MODEL_DATA = struct
 end
 
 module PVinPVH_memory_model_data : MEMORY_MODEL_DATA = struct
+  let extra_internal_mib = 1L +++ 1L
+
+  let extra_external_mib = 1L
+
+  let shim_mib _ = 0L
```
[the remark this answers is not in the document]
psafont: Yep, I copy-pasted the code and modified the wrong one

psafont:
Mirage unikernels still need to have in the platform field both `nx=true` and `pae=true` to be able to boot, but that's unrelated to this issue

ocaml/xenopsd/xc/domain.mli
```diff
@@ -105,7 +105,8 @@ val build_pv_info : build_pv_info Rpc.Types.def
 type build_pvh_info = {
     cmdline: string
   ; (* cmdline for the kernel (image) *)
-    modules: (string * string option) list
+    pv_shim: bool  (** whether the domain will be run with a PV shim *)
```
edwintorok: do you need a default true here for backward compat with already running PVinPVH guests?

edwintorok (approved):
LGTM

ocaml/xenopsd/xc/domain.mli
psafont: I believe so, yes

lindig (approved):

PR #4953, opened 2023-03-31
CP-39863 show allowed VTPM ops in the CLI

lindig:
Add code to show VTPM operations on VM and VTPM objects in the CLI.

### DIFF
--- a/ocaml/xapi-cli-server/records.ml
+++ b/ocaml/xapi-cli-server/records.ml
@@ -5240,5 +5240,15 @@ let vtpm_record rpc session_id vtpm =
       ; make_field ~name:"is_protected"
           ~get:(fun () -> string_of_bool (x ()).API.vTPM_is_protected)
           ()
+      ; make_field ~name:"allowed-operations"
+          ~get:(fun () ->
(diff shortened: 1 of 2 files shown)


edwintorok (approved):

ocaml/xapi-cli-server/records.ml
```diff
@@ -5240,5 +5240,15 @@ let vtpm_record rpc session_id vtpm =
...
           ~get:(fun () -> string_of_bool (x ()).API.vTPM_is_protected)
           ()
+      ; make_field ~name:"allowed-operations"
+          ~get:(fun () ->
+            map_and_concat Record_util.vtpm_operation_to_string
+              (x ()).API.vTPM_allowed_operations
+          )
+          ~get_set:(fun () ->
+            List.map Record_util.vtpm_operation_to_string
+              (x ()).API.vTPM_allowed_operations
```
psafont, line 5250: won't adding this method allow the user to set the allowed operations?

lindig: I don't know; I notice that this is in other records. I was expecting to see only `get`

PR #4956, opened 2023-04-03
CP-42455 Revert "disable DMC"

lindig:
Under CP-3397 we had disabled DMC since the Yangtze release. Revert the corresponding commits. This depends in addition on a change in V6 where the DMC feature is defined.


psafont:
The commits can be signed-off in a single action with `git rebase master --signoff`

ocaml/xapi/xapi_db_upgrade.ml
```diff
@@ -106,11 +106,6 @@ let jura =
   , Datamodel_common.jura_release_schema_minor_vsn
   )
 
-let yangtze =
```
psafont, line 109: Should this be removed?

psafont:
I don't see anything egregious, ballooning down on migration should be double-checked to _not_ happen, regardless of DMC being present or not as it make migrations less reliable

ocaml/xapi/xapi_db_upgrade.ml
lindig: We probably should keep this. Will fix it.

ocaml/idl/datamodel_errors.ml
```diff
@@ -176,8 +172,8 @@ let _ =
     () ;
   error Api_errors.memory_constraint_violation_maxpin ["reason"]
     ~doc:
-      "The dynamic memory range violates constraint static_min <= dynamic_min \
```
robhoes, line 179: I think this change was actually a fix to the error (independent of disabling DMC), as static-min is always allowed to be less than the other three params.

ocaml/xapi/xapi_db_upgrade.ml
```diff
@@ -235,23 +235,64 @@ let upgrade_wlb_configuration =
...
+    		{- t.dynamic_max := s.target}
+    		{- t.dynamic_min := s.target}}}
+    }
+*)
 let upgrade_vm_memory_for_dmc =
   {
-    description= "Upgrading VM memory fields to disable DMC"
-  ; version= (fun _ -> true)
+    description= "Upgrading VM memory fields for DMC"
+  ; version= (fun x -> x <= george)
```
robhoes, line 258: We can actually zap this entire rule now, since it is so old (George = XS 5.6 = 2009).

robhoes (approved):

ocaml/xapi/xapi_db_upgrade.ml
lindig: I've removed this upgrade rule, and the unit test for it. Would you like to take a second look? Otherwise I will merge this.

psafont (approved):

lindig:
Thanks - I fixed the sign-offs and updated the quality gate (after removing the upgrade case and test). Will merge this.

PR #4957, opened 2023-04-04
CA-375358: Parse output of yum upgrade to get RPMs to be updated/inst…

minglumlu:
…alled

In getting which RPM packages are going to be updated and installed before applying updates on a host, following cases are considered:
1. package-A-v1-r1 is installed and package-A-v2-r1 is an update. A new package-B-v1-r1 is required (hence introduced) by package-A-v2-r1.
2. package-X-v1-r1 is required by package-Y-v1-r1, and both are installed. Now package-Y-v2-r1 is an update, and doesn't require package-X anymore.

In case#1, package-B-v1-r1 can't be returned by "yum list updates", but it can be returned by "yum list available". This was fixed in commit c020034a21e. The fix results in another issue that in case#2, the package-X-v1-r1 is returned as an update from "yum list available". Ideally the package-X-v1-r1 should be obsoleted so that the fix in commit c020034a21e can handle it correctly. But it would be better to make the XAPI get the packages to be updated and installed exactly as what will be done in applying updates by "yum upgrade".

This approach was considered in the first place when this feature was implemented. But the difficulty is that the output of "yum upgrade" can't be formatted as other YUM commands.
To avoid more potential issues in future, this commit adds function to parse the RPM packages to be updated and installed from output of dry-run of "yum upgrade".


ocaml/xapi/repository_helpers.ml
```diff
@@ -717,6 +717,146 @@ let get_updates_from_repoquery repositories =
...
 
+(* Parser module for the output of command yum upgrade *)
+module YumUpgradeOutput = struct
+  open Angstrom
+
+  let consume = Consume.Prefix
+
+  let is_eol = function '\n' -> true | _ -> false
+
+  let is_eol_tab_or_space = function '\n' | '\t' | ' ' -> true | _ -> false
```
lindig, line 728: This is traditionally called white space; in programming languages tokens are separated by white space and it is often not significant.

lindig (approved):
The overall approach looks fragile to me. It relies on parsing output that is not meant to be consumed by a machine and relies on recognizing strings that could change and break the tool.

ocaml/xapi/repository_helpers.ml
psafont: This is (Astring.Char.Ascii.is_white](https://ocaml.org/p/astring/latest/doc/Astring/Char/Ascii/index.html#val-is_white)

ocaml/xapi/repository_helpers.ml
```diff
@@ -717,6 +717,196 @@ let get_updates_from_repoquery repositories =
...
+              line (Some section') [] acc
+        )
+        | "Transaction Summary" -> (
+          match section with
+          | Some s ->
+              line_after_txn_summary ((s, section_acc) :: acc)
+          | None ->
+              line_after_txn_summary acc
+        )
+        | l -> (
```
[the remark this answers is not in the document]
psafont: It seems it's collected in a different way (`Some s` instead of `Some l`, `l :: section_acc` instead of `[]`). This seems to have quite complex logic and some labels in the `line` function might help here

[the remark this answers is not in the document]
minglumlu: Added the labels of parameters and some comments.
The structure of the result would be:
`[(section_name:string, lines: string list); (section_name:string, lines: string list); ...]`

ocaml/xapi/rpm.mli
```diff
@@ -56,6 +56,8 @@ module Pkg : sig
   val lte : Epoch.t -> string -> string -> Epoch.t -> string -> string -> bool
 
   val gte : Epoch.t -> string -> string -> Epoch.t -> string -> string -> bool
+
```
lindig, line 59: This interface lacks documentation. What are all the strings for the comparison functions? Using labels could help to document them in code; or some comment is needed. I would have expected that the comparison functions work on `t` values.

minglumlu: I added a few comments as mitigation in this PR. The best may be to refine the typing to abstract a `RpmVersion.t` for  epoch:version-release. Could do it in another PR.

minglumlu:
Hi @lindig @psafont 
I had more tests for more cases. Based on the results, I appended a few changes and more unit tests for this. Could you please take a look again? Thank you very much!

ocaml/xapi/repository_helpers.ml
```diff
@@ -717,6 +717,224 @@ let get_updates_from_repoquery repositories =
...
+    with
+    | Ok ((_ :: _ as sections), txn_line_opt) -> (
+        let txn_file = parse_yum_txn_file txn_line_opt in
+        let get_pkgs lines =
+          let s = String.concat "\n" (List.rev_append lines []) in
+          parse_string ~consume packages s
+        in
+        sections
+        |> List.filter (fun (section, _) ->
+               [
```
lindig, line 868: I would write the filter function as 
```
fun (section, _) -> match section with 
|  "Installing" -> true
| "Updating" -> true
| ..
| _ -> false
```
and avoid building a list and using `List.mem`

lindig (approved):

minglumlu:
Hi @lindig, @psafont,
Would appreciate more reviews here.
I appended two commits, one is a fixup for a review comment, another is to add redundancy for getting the latest updates.

ocaml/xapi/xapi_fist.ml
```diff
@@ -110,6 +110,9 @@ let delay_xenopsd_event_threads () = fistpoint "delay_xenopsd_event_threads"
 
 let pause_after_cert_exchange () = fistpoint "pause_after_cert_exchange"
 
+let raise_error_in_yum_upgrade_dry_run () =
```
lindig, line 113: As mentioned above, I think `fail_on_error` would be better.

lindig (approved):

psafont (approved):

minglumlu:
Squashed fixups.

PR #4958, opened 2023-04-04
Add Tracing Library

jameshensmancitrix:
Created a Tracing library to produce and export spans on functions and send them to a specific http or file endpoint

### DIFF
--- a/ocaml/libs/tracing/tracing.ml
+++ b/ocaml/libs/tracing/tracing.ml
@@ -33,6 +33,12 @@ module SpanKind = struct
         "INTERNAL"
 end
 
+let endpoint_of_string = function
+  | "bugtool" ->
+      Bugtool
+  | url ->
+      Url (Uri.of_string url)
+
 let ok_none = Ok None
 
 module Status = struct
@@ -373,6 +379,52 @@ let lock = Mutex.create ()
 
 let tracer_providers = Hashtbl.create 100
 
+let set ?enabled ?tags ?endpoints ?filters ?processors ~uuid () =
+  Xapi_stdext_threads.Threadext.Mutex.execute lock (fun () ->
+      let provider =
+        match Hashtbl.find_opt tracer_providers uuid with
+        | Some (provider : TracerProvider.t) ->
+            let enabled = Option.value ~default:provider.enabled enabled in
+            let tags = Option.value ~default:provider.tags tags in
+            let endpoints =
+              Option.fold ~none:provider.endpoints
+                ~some:(List.map endpoint_of_string)
+                endpoints
+            in
+            let filters = Option.value ~default:provider.filters filters in
+            let processors =
+              Option.value ~default:provider.processors processors
+            in
+            {provider with enabled; tags; endpoints; filters; processors}
+        | None ->
+            failwith
+              (Printf.sprintf "The TracerProvider : %s does not exist" uuid)
+      in
+      Hashtbl.replace tracer_providers uuid provider
+  )
+
+let create ~enabled ~tags ~endpoints ~filters ~processors ~service_name
+    ~name_label ~uuid =
+  let endpoints = List.map endpoint_of_string endpoints in
+  let provider : TracerProvider.t =
+    {name_label; tags; endpoints; filters; processors; service_name; enabled}
+  in
+  Xapi_stdext_threads.Threadext.Mutex.execute lock (fun () ->
+      match Hashtbl.find_opt tracer_providers uuid with
+      | None ->
+          Hashtbl.add tracer_providers uuid provider
+      | Some _ ->
+          failwith "Tracing : TracerProvider already exists"
+  )
+
+let destroy ~uuid =
+  Xapi_stdext_threads.Threadext.Mutex.execute lock (fun () ->
+      Hashtbl.remove tracer_providers uuid
+  )
+
+let get_tracer_providers () =
+  Hashtbl.fold (fun _ provider acc -> provider :: acc) tracer_providers []
+
 let get_tracer ~name =
   let providers =
     Xapi_stdext_threads.Threadext.Mutex.execute lock (fun () ->
@@ -386,9 +438,6 @@ let get_tracer ~name =
       Tracer.create ~name ~provider
   | None ->
       warn "No provider found" ; Tracer.no_op
-
-let get_tracer_providers () =
-  Hashtbl.fold (fun _ provider acc -> provider :: acc) tracer_providers []
 
 let enable_span_garbage_collector ?(timeout = 86400.) () =
   Spans.GC.initialise_thread ~timeout
--- a/ocaml/libs/tracing/tracing.ml
+++ b/ocaml/libs/tracing/tracing.ml
@@ -70,22 +70,137 @@ module Spans = struct
 
   let spans = Hashtbl.create 100
 
+  let max_spans = ref 1000
+
+  let set_max_spans x = max_spans := x
+
+  let max_traces = ref 1000
+
+  let set_max_traces x = max_traces := x
+
+  let finished_spans = Hashtbl.create 100
+
   let add_to_spans ~(span : Span.t) =
+    let key = span.context.trace_id in
     Xapi_stdext_threads.Threadext.Mutex.execute lock (fun () ->
-        let key = span.context.trace_id in
         match Hashtbl.find_opt spans key with
         | None ->
-            Hashtbl.add spans key [span]
+            if Hashtbl.length spans < !max_traces then
+              Hashtbl.add spans key [span]
         | Some span_list ->
-            if List.length span_list < 1000 then
+            if List.length span_list < !max_spans then
               Hashtbl.replace spans key (span :: span_list)
     )
 
+  let remove_from_spans span =
+    let key = span.Span.context.trace_id in
+    Xapi_stdext_threads.Threadext.Mutex.execute lock (fun () ->
+        match Hashtbl.find_opt spans key with
+        | None ->
+            debug "Span does not exist or already finished" ;
+            None
+        | Some span_list ->
+            ( match
+                List.filter (fun x -> x.Span.context <> span.context) span_list
+              with
+            | [] ->
+                Hashtbl.remove spans key
+            | filtered_list ->
+                Hashtbl.replace spans key filtered_list
+            ) ;
+            Some span
+    )
+
+  let add_to_finished span =
+    let key = span.Span.context.trace_id in
+    Xapi_stdext_threads.Threadext.Mutex.execute lock (fun () ->
+        match Hashtbl.find_opt finished_spans key with
+        | None ->
+            if Hashtbl.length finished_spans < !max_traces then
+              Hashtbl.add finished_spans key [span]
+        | Some span_list ->
+            if List.length span_list < !max_spans then
+              Hashtbl.replace finished_spans key (span :: span_list)
+    )
+
+  let mark_finished span = Option.iter add_to_finished (remove_from_spans span)
+
+  let span_is_finished x =
+    match x with
+    | None ->
+        false
+    | Some (span : Span.t) -> (
+      match Hashtbl.find_opt finished_spans span.context.trace_id with
+      | None ->
+          false
+      | Some span_list ->
+          List.exists (fun x -> x = span) span_list
+    )
+
+  (** since copies the existing finished spans and then clears the existing spans as to only export them once  *)
   let since () =
     Xapi_stdext_threads.Threadext.Mutex.execute lock (fun () ->
-        let copy = Hashtbl.copy spans in
-        Hashtbl.clear spans ; copy
+        let copy = Hashtbl.copy finished_spans in
+        Hashtbl.clear finished_spans ;
+        copy
     )
+
+  module GC = struct
+    let lock = Mutex.create ()
+
+    let span_timeout = ref 86400.
+
+    let span_timeout_thread = ref None
+
+    let gc_inactive_spans () =
+      Xapi_stdext_threads.Threadext.Mutex.execute lock (fun () ->
+          Hashtbl.filter_map_inplace
+            (fun _ spanlist ->
+              let filtered =
+                List.filter_map
+                  (fun span ->
+                    let elapsed =
+                      Unix.gettimeofday () -. span.Span.begin_time
+                    in
+                    if elapsed > !span_timeout *. 1000000. then (
+                      debug "Tracing: Span %s timed out, forcibly finishing now"
+                        span.Span.context.span_id ;
+                      let span =
+                        Span.finish ~span
+                          ~tags:
+                            [
+                              ( "gc_inactive_span_timeout"
+                              , string_of_float elapsed
+                              )
+                            ]
+                          ()
+                      in
+                      add_to_finished span ; None
+                    ) else
+                      Some span
+                  )
+                  spanlist
+              in
+              match filtered with [] -> None | spans -> Some spans
+            )
+            spans
+      )
+
+    let initialise_thread ~timeout =
+      span_timeout := timeout ;
+      span_timeout_thread :=
+        Some
+          (Thread.create
+             (fun () ->
+               while true do
+                 debug "Tracing: Span garbage collector" ;
+                 Thread.delay !span_timeout ;
+                 gc_inactive_spans ()
+               done
+             )
+             ()
+          )
+  end
 end
 
 module TracerProvider = struct
@@ -128,7 +243,17 @@ module Tracer = struct
       let span = Span.start ~tags ~name ~parent () in
       Spans.add_to_spans ~span ; Ok (Some span)
 
-  let finish span = Ok (Option.map (fun span -> Span.finish ~span ()) span)
+  let finish span =
+    Ok
+      (Option.map
+         (fun span ->
+           let span = Span.finish ~span () in
+           Spans.mark_finished span ; span
+         )
+         span
+      )
+
+  let assert_finished x = Spans.assert_finished x
 end
 
 let lock = Mutex.create ()
@@ -148,6 +273,9 @@ let get_tracer ~name =
       Tracer.create ~name ~provider
   | None ->
       warn "No provider found" ; Tracer.no_op
+
+let enable_span_garbage_collector ?(timeout = 86400.) () =
+  Spans.GC.initialise_thread ~timeout
 
 module Export = struct
   module Content = struct
@@ -219,7 +347,6 @@ module Export = struct
           in
           Xapi_stdext_unix.Unixext.mkdir_rec (Filename.dirname file) 0o700 ;
           Xapi_stdext_unix.Unixext.write_string_to_file file span_json ;
-          (* TODO: Ensure file is only written when trace is complete *)
           Ok ""
         with e -> Error e
     end
--- a/ocaml/libs/tracing/tracing.ml
+++ b/ocaml/libs/tracing/tracing.ml
@@ -11,14 +11,11 @@
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU Lesser General Public License for more details.
  *)
-
 module D = Debug.Make (struct let name = "tracing" end)
 
 open D
 
 type endpoint = Bugtool | Url of Uri.t
-
-let url_file = "/etc/xapi-tracing-url"
 
 let ok_none = Ok None
 
@@ -213,6 +210,8 @@ module TracerProvider = struct
     ; enabled: bool
     ; service_name: string
   }
+
+  let endpoints_of t = t.endpoints
 end
 
 module Tracer = struct
@@ -274,10 +273,17 @@ let get_tracer ~name =
   | None ->
       warn "No provider found" ; Tracer.no_op
 
+let get_tracer_providers () =
+  Hashtbl.fold (fun _ provider acc -> provider :: acc) tracer_providers []
+
 let enable_span_garbage_collector ?(timeout = 86400.) () =
   Spans.GC.initialise_thread ~timeout
 
 module Export = struct
+  let export_interval = ref 30.
+
+  let set_export_interval t = export_interval := t
+
   module Content = struct
     module Json = struct
       module Zipkinv2 = struct
@@ -358,7 +364,6 @@ module Export = struct
       let export ~span_json ~url : (string, exn) result =
         try
           let body = span_json in
-          let uri = Uri.of_string url in
           let headers =
             Cohttp.Header.of_list
               [
@@ -367,9 +372,9 @@ module Export = struct
               ; ("content-length", string_of_int (String.length body))
               ]
           in
-          Open_uri.with_open_uri uri (fun fd ->
+          Open_uri.with_open_uri url (fun fd ->
               let request =
-                Cohttp.Request.make ~meth:`POST ~version:`HTTP_1_1 ~headers uri
+                Cohttp.Request.make ~meth:`POST ~version:`HTTP_1_1 ~headers url
               in
               let ic = Unix.in_channel_of_descr fd in
               let oc = Unix.out_channel_of_descr fd in
@@ -400,57 +405,43 @@ module Export = struct
         with e -> Error e
     end
 
-    let export_to_http_server () =
-      debug "Tracing: About to export to http server" ;
-      let url =
-        Xapi_stdext_unix.Unixext.string_of_file url_file |> String.trim
-      in
-      let span_list = Spans.since () in
+    let export_to_endpoint span_list endpoint =
       try
-        Hashtbl.iter
-          (fun _ span_list ->
-            let zipkin_spans = Content.Json.Zipkinv2.content_of span_list in
-            match Http.export ~span_json:zipkin_spans ~url with
-            | Ok _ ->
-                ()
-            | Error e ->
-                raise e
-          )
-          span_list ;
-        Ok ()
-      with e -> Error e
-
-    let export_to_logs () =
-      debug "Tracing: About to export to dom0 logs" ;
-      let span_list = Spans.since () in
-      try
+        debug "Tracing: About to export" ;
         Hashtbl.iter
           (fun trace_id span_list ->
             let zipkin_spans = Content.Json.Zipkinv2.content_of span_list in
             match
-              File.export ~trace_id ~span_json:zipkin_spans
-                ~path:!File.trace_log_dir
+              match endpoint with
+              | Url url ->
+                  Http.export ~span_json:zipkin_spans ~url
+              | Bugtool ->
+                  File.export ~trace_id ~span_json:zipkin_spans
+                    ~path:!File.trace_log_dir
             with
             | Ok _ ->
                 ()
             | Error e ->
                 raise e
           )
-          span_list ;
-        Ok ()
-      with e -> Error e
+          span_list
+      with e -> debug "Tracing: ERROR %s" (Printexc.to_string e)
 
     let main () =
+      enable_span_garbage_collector () ;
       Thread.create
         (fun () ->
           while true do
-            debug "Tracing: Waiting 30s before exporting spans" ;
-            Thread.delay 30. ;
-            match export_to_http_server () with
-            | Ok () ->
-                ()
-            | Error e ->
-                debug "Tracing: Export ERROR %s" (Printexc.to_string e)
+            debug "Tracing: Waiting %d seconds before exporting spans"
+              (int_of_float !export_interval) ;
+            Thread.delay !export_interval ;
+            let span_list = Spans.since () in
+            get_tracer_providers ()
+            |> List.iter (fun x ->
+                   if x.TracerProvider.enabled then
+                     TracerProvider.endpoints_of x
+                     |> List.iter (export_to_endpoint span_list)
+               )
           done
         )
         ()
--- a/ocaml/libs/tracing/tracing.ml
+++ b/ocaml/libs/tracing/tracing.ml
@@ -37,6 +37,16 @@ let ok_none = Ok None
 
 module SpanContext = struct
   type t = {trace_id: string; span_id: string} [@@deriving rpcty]
+
+  let to_traceparent t = Printf.sprintf "00-%s-%s-00" t.trace_id t.span_id
+
+  let of_traceparent traceparent =
+    let elements = String.split_on_char '-' traceparent in
+    match elements with
+    | ["00"; trace_id; span_id; "00"] ->
+        Some {trace_id; span_id}
+    | _ ->
+        None
 end
 
 module Span = struct
@@ -50,6 +60,8 @@ module Span = struct
     ; tags: (string * string) list
   }
   [@@deriving rpcty]
+
+  let get_context t = t.context
 
   let generate_id n = String.init n (fun _ -> "0123456789abcdef".[Random.int 16])
 
@@ -251,6 +263,17 @@ module Tracer = struct
       }
     in
     {name= ""; provider}
+
+  let span_of_span_context context name : Span.t =
+    {
+      context
+    ; name
+    ; parent= None
+    ; span_kind= SpanKind.Client (* This will be the span of the client call*)
+    ; begin_time= Unix.gettimeofday ()
+    ; end_time= None
+    ; tags= []
+    }
 
   let start ~tracer:t ?(span_kind = SpanKind.Internal) ~name ~parent () :
       (Span.t option, exn) result =
--- a/ocaml/libs/tracing/tracing.ml
+++ b/ocaml/libs/tracing/tracing.ml
@@ -200,6 +200,30 @@ module Export = struct
   end
 
   module Destination = struct
+    module File = struct
+      let trace_log_dir = ref "/var/log/dt/zipkinv2/json"
+
+      let host_id = ref "localhost"
+
+      let set_trace_log_dir dir = trace_log_dir := dir
+
+      let set_host_id id = host_id := id
+
+      let export ~trace_id ~span_json ~path : (string, exn) result =
+        try
+          let date = Ptime_clock.now () |> Ptime.to_rfc3339 ~frac_s:6 in
+          let file =
+            path
+            ^ String.concat "-" [trace_id; "xapi"; !host_id; date]
+            ^ ".json"
+          in
+          Xapi_stdext_unix.Unixext.mkdir_rec (Filename.dirname file) 0o700 ;
+          Xapi_stdext_unix.Unixext.write_string_to_file file span_json ;
+          (* TODO: Ensure file is only written when trace is complete *)
+          Ok ""
+        with e -> Error e
+    end
+
     module Http = struct
       module Request = Cohttp.Request.Make (Cohttp_posix_io.Buffered_IO)
       module Response = Cohttp.Response.Make (Cohttp_posix_io.Buffered_IO)
@@ -260,6 +284,26 @@ module Export = struct
           (fun _ span_list ->
             let zipkin_spans = Content.Json.Zipkinv2.content_of span_list in
             match Http.export ~span_json:zipkin_spans ~url with
+            | Ok _ ->
+                ()
+            | Error e ->
+                raise e
+          )
+          span_list ;
+        Ok ()
+      with e -> Error e
+
+    let export_to_logs () =
+      debug "Tracing: About to export to dom0 logs" ;
+      let span_list = Spans.since () in
+      try
+        Hashtbl.iter
+          (fun trace_id span_list ->
+            let zipkin_spans = Content.Json.Zipkinv2.content_of span_list in
+            match
+              File.export ~trace_id ~span_json:zipkin_spans
+                ~path:!File.trace_log_dir
+            with
             | Ok _ ->
                 ()
             | Error e ->
--- a/ocaml/libs/tracing/tracing.ml
+++ b/ocaml/libs/tracing/tracing.ml
@@ -63,6 +63,7 @@ module Span = struct
     ; status: Status.t
     ; parent: t option
     ; name: string
+    ; service_name: string
     ; begin_time: float
     ; end_time: float option
     ; tags: (string * string) list
@@ -73,7 +74,7 @@ module Span = struct
 
   let generate_id n = String.init n (fun _ -> "0123456789abcdef".[Random.int 16])
 
-  let start ?(tags = []) ~name ~parent ~span_kind () =
+  let start ?(tags = []) ~name ~parent ~span_kind ~service_name () =
     let trace_id =
       match parent with
       | None ->
@@ -87,7 +88,17 @@ module Span = struct
     let begin_time = Unix.gettimeofday () in
     let end_time = None in
     let status : Status.t = {status_code= Status.Unset; description= None} in
-    {context; span_kind; status; parent; name; begin_time; end_time; tags}
+    {
+      context
+    ; span_kind
+    ; status
+    ; parent
+    ; name
+    ; service_name
+    ; begin_time
+    ; end_time
+    ; tags
+    }
 
   let finish ?(tags = []) ~span () =
     {span with end_time= Some (Unix.gettimeofday ()); tags= span.tags @ tags}
@@ -312,11 +323,12 @@ module Tracer = struct
     in
     {name= ""; provider}
 
-  let span_of_span_context context name : Span.t =
+  let span_of_span_context t context name : Span.t =
     {
       context
     ; status= {status_code= Status.Unset; description= None}
     ; name
+    ; service_name= t.provider.service_name
     ; parent= None
     ; span_kind= SpanKind.Client (* This will be the span of the client call*)
     ; begin_time= Unix.gettimeofday ()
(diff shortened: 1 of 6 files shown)


ocaml/libs/tracing/tracing.ml
```diff
@@ -0,0 +1,97 @@
...
+
+  let finish ?(tags = []) ~span () =
+    span.span_end_time <- Some (Unix.gettimeofday ()) ;
+    match (span.tags, tags) with
+    | _, [] ->
+        ()
+    | [], new_tags ->
+        span.tags <- new_tags
+    | orig_tags, new_tags ->
+        span.tags <- orig_tags @ new_tags
```
robhoes, line 55: Instead of using mutable fields, you can also construct a new value. Something like this:
```
{ span with
    span_end_time = Some (Unix.gettimeofday ()) ;
    tags = span.tags @ tags
}
```
But from this first patch it is not yet clear what the hash table is about. I guess your trying to update values in the hashtable, in place?

ocaml/xapi/context.ml
```diff
@@ -237,13 +250,25 @@ let from_forwarded_task ?(http_other_config = []) ?session_id
   let dbg = make_dbg http_other_config task_name task_id in
   info "task %s forwarded%s" dbg
     (trackid_of_session ~with_brackets:true ~prefix:" " session_id) ;
+  let tracing =
+    let open Tracing in
+    (* Set parent based on trace context from forwarded call instead! *)
```
robhoes, line 255: This is done in a future PR.

ocaml/xapi/context.ml
```diff
@@ -282,13 +307,27 @@ let make ?(http_other_config = []) ?(quiet = false) ?subtask_of ?session_id
             " by task " ^ make_dbg [] "" subtask_of
         )
   ) ;
+  let tracing =
+    let open Tracing in
+    (* Set parent based on incoming trace context instead! *)
+    let parent = empty in
+    match start ~name:task_name ~parent with
+    | Ok x ->
+        R.debug "Started trace: %s"
```
robhoes, line 316: I think we should drop these. It will get far too spammy.

ocaml/libs/tracing/tracing.ml
```diff
@@ -0,0 +1,97 @@
...
+let spans = Hashtbl.create 100
+
+let add_to_spans ~(span : Span.t) =
+  Xapi_stdext_threads.Threadext.Mutex.execute Span.lock (fun () ->
+      let key = span.span_context.trace_id in
+      match Hashtbl.find_opt spans key with
+      | None ->
+          Hashtbl.add spans key [span]
+      | Some span_list ->
+          if List.length span_list < 1000 then
```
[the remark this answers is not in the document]
jameshensmancitrix: This is handled later, I probably would be best to include that here.

ocaml/libs/tracing/tracing.ml
```diff
@@ -0,0 +1,97 @@
...
+  |> Rpcmarshal.unmarshal blob.Rpc.Types.ty
+  |> Result.to_option
+
+let empty : t = None
+
+let is_empty = function None -> true | Some _ -> false
+
+let start ~name ~parent : (t, exn) result =
+  let span = Span.start ~name ~parent () in
+  add_to_spans ~span ; Ok (Some span)
```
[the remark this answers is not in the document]
jameshensmancitrix: In OpenTelemetry, Spans are created by Tracers and later on this start is moved into the Tracer module. `Span.start` is intended as a constructor.

ocaml/libs/tracing/tracing.ml
```diff
@@ -0,0 +1,81 @@
...
+      context: SpanContext.t
+    ; parent: t option
+    ; name: string
+    ; begin_time: float
+    ; end_time: float option
+    ; tags: (string * string) list
+  }
+  [@@deriving rpcty]
+
+  let generate_id n = String.init n (fun _ -> "0123456789abcdef".[Random.int 16])
```
edwintorok, line 101: Might be more efficient to generate a single int64 number and then encode it in hex, but I haven't measured. For now lets leave it as is and we can add some microbenchmarks later, see what brings up and then tweak the code if needed.

edwintorok: When we start using the Opentelemetry library we can use its method of generating span/trace ids: https://ocaml.org/p/opentelemetry/latest/doc/Opentelemetry/Rand_bytes/index.html

ocaml/libs/tracing/tracing.ml
```diff
@@ -0,0 +1,81 @@
...
+  let lock = Mutex.create ()
+
+  let spans = Hashtbl.create 100
+
+  let add_to_spans ~(span : Span.t) =
+    Xapi_stdext_threads.Threadext.Mutex.execute lock (fun () ->
+        let key = span.context.trace_id in
+        match Hashtbl.find_opt spans key with
+        | None ->
+            Hashtbl.add spans key [span]
```
edwintorok, line 178: This can potentially result in unbounded memory usage since the hashtable is global, but I assume there is a commit later on that implements a GC?

ocaml/libs/tracing/tracing.ml
```diff
@@ -0,0 +1,81 @@
...
+  let of_string s =
+    Jsonrpc.of_string s
+    |> Rpcmarshal.unmarshal t.Rpc.Types.ty
+    |> Result.to_option
+end
+
+module Spans = struct
+  let lock = Mutex.create ()
+
+  let spans = Hashtbl.create 100
```
edwintorok, line 169: Note: once we start using the opentelemetry library we can use its Thread local data structure (I haven't checked how efficiently it is implemented): https://ocaml.org/p/opentelemetry/latest/doc/Opentelemetry/Thread_local/index.html

ocaml/libs/tracing/tracing.ml
```diff
@@ -11,6 +11,25 @@
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU Lesser General Public License for more details.
  *)
+
+module D = Debug.Make (struct let name = "tracing" end)
+
+open D
+
+type endpoint = Bugtool | Url of string [@@deriving rpcty]
```
edwintorok, line 18: Why is the Url a `string` instead of a`Uri.t`? If this is because Uri.t would lack a deriving I have some code that could help with that.

ocaml/libs/tracing/tracing.ml
```diff
@@ -71,11 +90,92 @@ module Spans = struct
...
+        ; enabled= false
+        }
+    in
+    {name= ""; provider}
+
+  let start ~tracer:t ~name ~parent : (Span.t option, exn) result =
+    let provider = !(t.provider) in
+    (* Do not start span if the TracerProvider is diabled*)
+    if not provider.enabled then
+      Ok None
```
edwintorok, line 302: This creates a new 'Ok None' value everytime. Since we probably want to consider tracing disabled as the fastpath it might be useful to have a preallocated `let ok_none = Ok None`. This is what Lwt does to optimize common pattern like `Lwt.return ()` and `Lwt.return None` (it has a `Lwt.return_unit` and `Lwt.return_none`).

ocaml/libs/tracing/tracing.ml
```diff
@@ -71,11 +90,92 @@ module Spans = struct
...
     let _span = Option.map (fun span -> Span.finish ~span ()) span in
     Ok ()
 end
+
+module TracerProvider = struct
+  type t = {tracers: Tracer.t list; config: provider_config_t}
+
+  let get_tracer ~provider:t ~name =
+    match
+      List.filter (fun (tracer : Tracer.t) -> tracer.name = name) t.tracers
```
edwintorok, line 337: Are we going to look up tracers like this often? How many tracers do you normally have?
If it is a small number, like 1 or 2 then the List may be better, if you expect to have lots (hundreds) then a String Map might be better (O(log n) vs O(n) lookup time).

ocaml/libs/tracing/tracing.ml
```diff
@@ -214,13 +224,103 @@ module Export = struct
...
+          span_list ;
+        Ok ()
+      with e -> Error e
+
+    let _ =
+      Thread.create
+        (fun () ->
+          while true do
+            debug "Tracing: Waiting 30s before exporting spans" ;
+            Thread.delay 30. ;
```
edwintorok, line 632: Would be good to make this configurable.

ocaml/libs/tracing/tracing.ml
```diff
@@ -239,6 +241,36 @@ module Export = struct
   end
 
   module Destination = struct
+    module File = struct
+      let export ~trace_id ~span_json : (string, exn) result =
+        try
+          (* TODO: *)
```
edwintorok, line 529: TODO... is there something missing here?

ocaml/libs/tracing/tracing.ml
```diff
@@ -92,11 +94,85 @@ module Spans = struct
...
+        copy
     )
+
+  module GC = struct
+    let span_timeout = ref 86400.
+
+    let span_timeout_thread = ref None
+
+    let gc_inactive_spans () =
+      Hashtbl.iter
```
edwintorok, line 227: This needs a mutex: it modifies a global data structure and that is not safe to do concurrently.
See also https://v2.ocaml.org/api/Hashtbl.html 
"Alert unsynchronized_access. Unsynchronized accesses to hash tables are a programming error." the hashtable doesn't have any internal locking.

ocaml/libs/tracing/tracing.ml
```diff
@@ -87,13 +110,39 @@ module Span = struct
...
     let end_time = None in
-    {context; span_kind; parent; name; begin_time; end_time; tags}
+    let status : Status.t = {status_code= Status.Unset; description= None} in
+    {context; span_kind; status; parent; name; begin_time; end_time; tags}
 
   let finish ?(tags = []) ~span () =
+    let status = span.status in
+    let tags =
+      if status.status_code = Error then
+        (Status.status_to_string status, Printexc.get_backtrace ()) :: tags
```
edwintorok, line 132: This is too late, you've already called 'Status.status_to_string' which might've destroyed the backtrace.
You need to capture the backtrace as close the error as possible, i.e. ideally whatever set the status_code to Error.

ocaml/libs/tracing/tracing.ml
```diff
@@ -87,13 +110,39 @@ module Span = struct
...
+      else
+        tags
+    in
     {span with end_time= Some (Unix.gettimeofday ()); tags= span.tags @ tags}
 
   let set_span_kind span kind = span.span_kind <- kind
 
+  let set_error span exn =
+    let description = Some (Printexc.to_string exn) in
```
edwintorok, line 141: You need to capture the stacktrace here, before you call anything else (in theory even calling Printexc.to_string might've already destroyed the backtrace).

edwintorok:
I would suggest adding some unit tests that check that:
* we do not introduce any space leaks when tracing is enabled (I think the current code does, see my comments) by checking that the Hashtable is empty at the end of a test that uses tracing (and non-empty during the test). Could also try to check OCaml memory usage before/after the test and assert they are equal, but from having tried to do that in the past it is not easy, and might be off by a few bytes due to memory used by the test framework itself.

* exception backtraces are properly retained. E.g. raise a particular exception during a unit test and then check the exception stacktrace, and assert that it contains the function name that raised the exception. It is quite tricky to get exception backtraces correctly retained (XAPI currently doesn't do it correctly in all cases...), so having tests for this would be useful

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: That would be ideal to use that library.

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: Yes we have a GC later on

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: It makes it easy to derive yes, and also to quite easy to get from a string. I am not aware of `Uri.t` so I will look into that.

mg12:
> I would suggest adding some unit tests that check that:
> 
> * we do not introduce any space leaks when tracing is enabled (I think the current code does, see my comments) by checking that the Hashtable is empty at the end of a test that uses tracing (and non-empty during the test). Could also try to check OCaml memory usage before/after the test and assert they are equal, but from having tried to do that in the past it is not easy, and might be off by a few bytes due to memory used by the test framework itself.
> * exception backtraces are properly retained. E.g. raise a particular exception during a unit test and then check the exception stacktrace, and assert that it contains the function name that raised the exception. It is quite tricky to get exception backtraces correctly retained (XAPI currently doesn't do it correctly in all cases...), so having tests for this would be useful

@YanRachel

ocaml/libs/tracing/tracing.ml
mg12: @snwoods

ocaml/libs/tracing/tracing.ml
mg12: @snwoods

ocaml/libs/tracing/tracing.ml
mg12: @jameshensmancitrix

ocaml/libs/tracing/tracing.ml
```diff
@@ -214,13 +224,103 @@ module Export = struct
...
+          in
+          Open_uri.with_open_uri uri (fun fd ->
+              let request =
+                Cohttp.Request.make ~meth:`POST ~version:`HTTP_1_1 ~headers uri
+              in
+              let ic = Unix.in_channel_of_descr fd in
+              let oc = Unix.out_channel_of_descr fd in
+              Request.write
+                (fun writer -> Request.write_body writer body)
+                request oc ;
```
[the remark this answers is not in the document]
mg12: @snwoods

ocaml/libs/tracing/tracing.ml
mg12: @snwoods

ocaml/libs/tracing/tracing.ml
mg12: @YanRachel

ocaml/libs/tracing/tracing.ml
```diff
@@ -239,6 +241,36 @@ module Export = struct
   end
 
   module Destination = struct
+    module File = struct
+      let export ~trace_id ~span_json : (string, exn) result =
+        try
+          (* TODO: *)
+          let host_id = "" in
+          let timestamp = Unix.gettimeofday () in
+          let timestamp_ms =
```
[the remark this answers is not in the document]
mg12: @snwoods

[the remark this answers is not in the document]
YanRachel: https://github.com/xapi-project/xen-api/pull/4946#discussion_r1146487769

ocaml/libs/tracing/tracing.ml
```diff
@@ -0,0 +1,663 @@
...
+    Xapi_stdext_threads.Threadext.Mutex.execute lock (fun () ->
+        let provider = Hashtbl.find tracer_providers "default" in
+        Ok provider
+    )
+  with e -> Error e
+
+let get_tracer ~name =
+  let providers =
+    Xapi_stdext_threads.Threadext.Mutex.execute lock (fun () ->
+        Hashtbl.fold (fun _k v acc -> v :: acc) tracer_providers []
```
jameshensmancitrix, line 472: Should filter if there is a Tracer enabled or not, will add a fixup for this

ocaml/libs/tracing/tracing.ml
snwoods: Resolved by 6f41e5ca3deac76ae8fb2b982df6fb4777f9f39e

ocaml/libs/tracing/tracing.ml
```diff
@@ -0,0 +1,81 @@
...
+    let trace_id =
+      match parent with
+      | None ->
+          generate_id 32
+      | Some span_parent ->
+          span_parent.context.trace_id
+    in
+    let span_id = generate_id 16 in
+    let context : SpanContext.t = {trace_id; span_id} in
+    let begin_time = Unix.gettimeofday () in
```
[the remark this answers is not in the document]
snwoods: Resolved by db13dab112d47ff0e7542dacd9d20a01842b9fbb

ocaml/libs/tracing/tracing.ml
snwoods: Resolved by aab0e0a8f3e772eb2a12232be92100c0cfac94dc

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
snwoods: Resolved by cc11eccbfef01ccbcb8f93c68acd7eb29965b8c9

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: CP-42794

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: CP-42794

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
jameshensmancitrix: resolved by[83a362e](https://github.com/xapi-project/xen-api/pull/4958/commits/83a362ee76427e2664b5315c997dbd0e4904d7b9)

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
jameshensmancitrix: This is added from the begining. Tracers create spans and manage them in the wider context whereas functions of Span are related to mutating / creating / extracting information about Spans.

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: The structure was remade to largely use less side effects, there is one instance where this cannot be done.

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: CP-42794

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: Will delete as not used

ocaml/libs/tracing/tracing.ml
mg12: @jameshensmancitrix

ocaml/libs/tracing/tracing.ml
```diff
@@ -0,0 +1,666 @@
...
+
+  let assert_span_hashtbl_empty () =
+    debug "hashtbl len: %d" (Hashtbl.length spans) ;
+    Hashtbl.length spans = 0
+
+  let add_to_spans ~(span : Span.t) =
+    Xapi_stdext_threads.Threadext.Mutex.execute lock (fun () ->
+        let key = span.context.trace_id in
+        match Hashtbl.find_opt spans key with
+        | None ->
```
mg12, line 200: `if Hashtbl.length spans < !max_trace_ids then`

ocaml/libs/tracing/tracing.ml
```diff
@@ -92,11 +94,85 @@ module Spans = struct
...
+              let span =
+                Span.finish ~span
+                  ~tags:[("gc_inactive_span_timeout", string_of_float elapsed)]
+                  ()
+              in
+              mark_finished ~span
+            )
+            spanlist
+        )
+        spans
```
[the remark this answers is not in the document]
mg12: @jameshensmancitrix do you have the commit id that fixes this comment?

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
snwoods: As per the slack conversation, I don't think we need to flush manually as we are not concerned about the timing of this

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
jameshensmancitrix: This fix was done as a squash so there is no fixup commit

ocaml/xapi/context.ml
jameshensmancitrix: Done in other PR

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: Resolved by [50a18b5](https://github.com/xapi-project/xen-api/pull/4958/commits/50a18b557327e07d332e11f34b0086a85d5155d1)

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: [6be7576](https://github.com/xapi-project/xen-api/pull/4958/commits/6be7576d40f3f6984434efbbddb96fe6a95a9d8c)

ocaml/libs/tracing/tracing.ml
```diff
@@ -0,0 +1,665 @@
...
+  let config : provider_config_t =
+    {name_label; tags; endpoints; filters; processors; service_name; status}
+  in
+  match Hashtbl.find_opt tracer_providers uuid with
+  | None ->
+      Hashtbl.add tracer_providers uuid {tracers; config}
+  | Some _ ->
+      failwith "Tracing : TracerProvider already exists"
+
+let destory ~name_label ~uuid =
```
mg12, line 453: destroy?

jameshensmancitrix: Resolved by 
43d7b59

ocaml/tests/test_tracing.ml
```diff
@@ -0,0 +1,58 @@
...
+          Unix.sleep 30 ;
+          (* Wait for export to clear hashtbl *)
+          not (Tracer.assert_span_hashtbl_empty ())
+    | Error e ->
+        debug "%s" (Printexc.to_string e) ;
+        false
+  in
+  Alcotest.(check bool) "No space leaks introduced in span hashtable" false leak
+
+let test_tracing_exn_backtraces () =
```
YanRachel, line 49: oh this. doesn't look done yet

ocaml/libs/tracing/tracing.ml
```diff
@@ -220,7 +225,8 @@ module Spans = struct
               Hashtbl.replace spans key filtered_list ;
               match Hashtbl.find_opt finished_spans key with
               | None ->
-                  Hashtbl.add finished_spans key [span]
+                  if Hashtbl.length spans < !max_traces then
```
edwintorok, line 228: copy+past error here, did you mean `finished_spans` instead of `spans`?

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
edwintorok: The problem is not whether you write it often or not, it is whether you lose data or not: if you do not flush the output channel and then close the underlying file descriptor then (depending on the size of data written) you may not send the entire string that you intended to send. However looking at the implementation in 'Cohttp_posix_io.Buffered_io.write' it already calls 'flush' internally, so this should be fine.

ocaml/tests/test_tracing.ml
```diff
@@ -0,0 +1,58 @@
...
+let test_hashtbl_leaks () =
+  let leak =
+    let span = start_test_span () in
+    match span with
+    | Ok x ->
+        if Tracer.assert_span_hashtbl_empty () then
+          true
+        else
+          let _ = Tracer.finish x in
+          Unix.sleep 30 ;
```
edwintorok, line 40: Having large sleeps in unit test code is usually not a good idea, why is this exactly 30 seconds, can it be made smaller?
If there is a timeout somewhere else then the unit test should set that timeout to a small value for testing purposes (to something small like 0.2s, and wait a bit more like 0.3 or 0.4s)

ocaml/tests/test_tracing.ml
```diff
@@ -0,0 +1,58 @@
...
+  let span = Tracer.start ~tracer ~name:"test_task" ~parent:None () in
+  set ~name_label:"default" ~status:false () ;
+  span
+
+let test_hashtbl_leaks () =
+  let leak =
+    let span = start_test_span () in
+    match span with
+    | Ok x ->
+        if Tracer.assert_span_hashtbl_empty () then
```
edwintorok, line 36: This works but might be clearer if you assert twice with different error messages, the first one should say that it asserts that we have some spans, and then the last one should assert that there are no leaks.
That would avoid having to negate both sides

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
edwintorok: I think there is still a bug here, looking at the latest code in 'gc_inactive_spans' it does 'Hashtbl.iter', and call 'mark_finished', which may end up calling 'Hashtbl.remove' which mutates the hashtable you are iterating on, which according to Hashtbl's documentation is not safe to do.

edwintorok:
Added some comments: there is one [copy+paste error ](https://github.com/xapi-project/xen-api/pull/4958#discussion_r1169697202) and an issue with using [Hashtbl.remove inside Hashtbl.iter](https://github.com/xapi-project/xen-api/pull/4958#discussion_r1169725905). And not necessarily a bug, but we [shouldn't have hardcoded sleeps of 30s in unit test code ](https://github.com/xapi-project/xen-api/pull/4958#discussion_r1169713695)

ocaml/tests/test_tracing.ml
YanRachel: It's to wait for the export thread in tracing.ml to clear the span hashtable -we could add another function that force-exports without doing the wait and calling that instead?

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: [Resolved ](https://github.com/xapi-project/xen-api/pull/4958/commits/34e9795b2b9c0e4964af7369f7f4b515105c3ea4)

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
jameshensmancitrix: Would iterating on a copy or a list be safer?

ocaml/libs/tracing/tracing.ml
```diff
@@ -0,0 +1,674 @@
...
+                  loop () ;
+                  Ok (Buffer.contents body)
+          )
+        with e -> Error e
+    end
+
+    let export_to_endpoint endpoint =
+      try
+        debug "Tracing: About to export" ;
+        let span_list = Spans.since () in
```
robhoes, line 637: This gets all newly finished spans, since the last export, right? It doesn't seem to specify that the spans are recorded under a certain tracer or tracer provider, where the endpoint is defined. So if you have multiple tracer providers with different endpoints, then all the same traces are sent to the first endpoint in the `List.iter` below, and the following ones get nothing?

ocaml/libs/tracing/tracing.ml
```diff
@@ -0,0 +1,674 @@
...
+      with e -> debug "Tracing: ERROR %s" (Printexc.to_string e)
+
+    let _ =
+      enable_span_garbage_collector () ;
+      Thread.create
+        (fun () ->
+          while true do
+            debug "Tracing: Waiting %d seconds before exporting spans"
+              (int_of_float !export_timeout) ;
+            Thread.delay !export_timeout ;
```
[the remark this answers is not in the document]
jameshensmancitrix: Control how often exports are done

ocaml/libs/tracing/tracing.ml
```diff
@@ -0,0 +1,665 @@
...
+    let description =
+      Some
+        (Printf.sprintf "Error: %s Backtrace: %s" (Printexc.to_string exn)
+           backtrace
+        )
+    in
+    let status_code = Status.Error in
+    match span.status.status_code with
+    | Unset ->
+        {span with status= {status_code; description}}
```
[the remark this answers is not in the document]
jameshensmancitrix: Because we do not know if it will be a client or not (The task associated sends a http request) from the start. We can set it to a client if it is used to a set the traceparent header. If there is a better way to do this then that would be an option

ocaml/libs/tracing/tracing.ml
```diff
@@ -0,0 +1,665 @@
...
+                 gc_inactive_spans ()
+               done
+             )
+             ()
+          )
+  end
+end
+
+module Tracer = struct
+  type t = {name: string; provider: provider_config_t ref}
```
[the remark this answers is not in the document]
jameshensmancitrix: [ OpenTelemetry specification
](https://opentelemetry.io/docs/reference/specification/trace/api/#tracerprovider-operations)

> Note: This could, for example, be implemented by storing any mutable configuration in the TracerProvider and having Tracer implementation objects have a reference to the TracerProvider from which they were obtained. If configuration must be stored per-tracer (such as disabling a certain tracer), the tracer could, for example, do a look-up with its identity in a map in the TracerProvider, or the TracerProvider could maintain a registry of all returned Tracers and actively update their configuration if it changes.

 If it were about functionality I would refactor Tracer and TracerProvider into one module.

ocaml/libs/tracing/tracing.ml
```diff
@@ -0,0 +1,674 @@
...
+  else
+    Hashtbl.remove tracer_providers uuid
+
+let get_tracer_providers () =
+  Hashtbl.fold (fun _ provider acc -> provider :: acc) tracer_providers []
+
+let get_default () =
+  try
+    Xapi_stdext_threads.Threadext.Mutex.execute lock (fun () ->
+        let provider = Hashtbl.find tracer_providers "default" in
```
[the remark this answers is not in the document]
jameshensmancitrix: `get_default` isn't used to I could understand getting rid of it. I could also add default via create, however it also comes back to [ this
](https://opentelemetry.io/docs/reference/specification/trace/api/#tracerprovider)
Normally, the TracerProvider is expected to be accessed from a central place. Thus, the API SHOULD provide a way to set/register and access a global default TracerProvider.

ocaml/libs/tracing/tracing.ml
```diff
@@ -0,0 +1,674 @@
...
+            with
+            | Ok _ ->
+                ()
+            | Error e ->
+                raise e
+          )
+          span_list
+      with e -> debug "Tracing: ERROR %s" (Printexc.to_string e)
+
+    let _ =
```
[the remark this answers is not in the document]
jameshensmancitrix: We need the export thread for multiple daemons as well, this was it is done without needing to start this up in all of them.

ocaml/tests/test_tracing.ml
YanRachel: resolved in ebd6aead72445ef2c9af5090ca9432e065ca5fd6

ocaml/tests/test_tracing.ml
YanRachel: resolved in ebd6aead72445ef2c9af5090ca9432e065ca5fd6

ocaml/tests/test_tracing.ml
YanRachel: resolved in ebd6aead72445ef2c9af5090ca9432e065ca5fd6

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
jameshensmancitrix: Resolved by561538ae93abe485bc236f4ab50da0bd3686d11a

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
jameshensmancitrix: Partly resolved bybbcece5ba2b97f9fcb187f193706b5fb2a364e53

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
robhoes: The alternative to a mutable field in a more functional style is to create a new record with the field updated, as is done here for the status. So I'm questioning why two different methods are used for fields in the same record.

Mutable fields could be convenient if records are stuck in a hash table as you don't have to remove and re-add. Functional updates have other benefits in general.

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
jameshensmancitrix: Resolved by bbcece5ba2b97f9fcb187f193706b5fb2a364e53, 9e0e83cfa9b9d472a3947e271eb1b961143e4c17

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
jameshensmancitrix: The issue was the updating for this field was deeply embedded in the code and not in the instrumentation, however if we called set before we attach a span in a http header then put it back into context then it should propegate.

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
robhoes: For libraries especially it is better to have an init function rather than a `let _ =` "main" function, to give the users of the library more control. Having multiple "main" functions in a program just adds unpredictability and risks race conditions.

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
robhoes: I think we want the library to support creating any number of TracerProviders, including zero. You could have a convention that the first one that is created is the "default" one, if you need a default. Or have some way of designating a default after creating a couple. I think having a separate create function for default provider is useful. And definitely not with a name that is hardcoded in the library.

For our initial implementation in xapi, the common cases will probably be either zero (tracing disabled) or one (tracing enabled) provider. When tracing is enabled, just create one TracerProvider that will then be the global one for the program.

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: Resolved by[ ](https://github.com/xapi-project/xen-api/pull/4958/commits/be140243d2fad8b9f8f4863742a0de5e4700dbfb)

jameshensmancitrix: https://github.com/xapi-project/xen-api/pull/4958/commits/be140243d2fad8b9f8f4863742a0de5e4700dbfb

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
jameshensmancitrix: call it `interval`?

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
jameshensmancitrix: Providers can be disabled hence tracing diabled

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
jameshensmancitrix: I had accidently forgotten some changes I have fixed that now.

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
jameshensmancitrix: resolved by 0e0f99fb2e7c095b232832f0d76e19b928d26f85)

ocaml/libs/tracing/tracing.mli
```diff
@@ -0,0 +1,123 @@
...
+  val set_export_interval : float -> unit
+
+  module Destination : sig
+    module File : sig
+      val set_trace_log_dir : string -> unit
+
+      val set_host_id : string -> unit
+    end
+
+    val main : unit -> Thread.t
```
jameshensmancitrix, line 121: It would be neater to put the `main ()` function on the outside of Export.Destination

jameshensmancitrix: Resolved in fixups

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
jameshensmancitrix: Have removed the two functions in fixups

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
jameshensmancitrix: Convert hash table to a sequence and iterate over that sequence instead.

ocaml/libs/tracing/tracing.ml
```diff
@@ -0,0 +1,658 @@
...
+  | Some _ ->
+      failwith "Tracing : TracerProvider already exists"
+
+let destroy ~uuid =
+  let provider = Hashtbl.find_opt tracer_providers uuid in
+  match provider with
+  | None ->
+      ()
+  | Some provider ->
+      if provider.name_label = "default" then
```
robhoes, line 454: Obsolete now?

robhoes (approved):

xapi-tracing.opam
```diff
@@ -0,0 +1,25 @@
+opam-version: "2.0"
```
psafont, line 1: Note package will need to be added to xs-opam under `packages/xs-extra`

ocaml/libs/tracing/tracing.ml
```diff
@@ -78,6 +80,12 @@ module Spans = struct
             if List.length span_list < 1000 then
               Hashtbl.replace spans key (span :: span_list)
     )
+
+  let since () =
```
psafont, line 236: This should have a doc-comment explaining why it clears the hashtable

ocaml/libs/tracing/tracing.ml
```diff
@@ -71,10 +80,87 @@ module Spans = struct
     )
 end
 
-module Tracer = struct
-  let start ~name ~parent : (Span.t option, exn) result =
-    let span = Span.start ~name ~parent () in
-    Spans.add_to_spans ~span ; Ok (Some span)
+module rec Tracer : sig
```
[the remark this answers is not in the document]
psafont: (unless it's needed for later PRs)

xapi-tracing.opam
jameshensmancitrix: https://github.com/xapi-project/xs-opam/pull/637

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: Added

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
jameshensmancitrix: It has been removed, it could be added back if it is decided it is needed in the future, it is best to keep it simple for now.

ocaml/tests/suite_alcotest.ml
```diff
@@ -46,6 +46,7 @@ let () =
      ; ("Test_storage_migrate_state", Test_storage_migrate_state.test)
      ; ("Test_bios_strings", Test_bios_strings.test)
      ; ("Test_certificates", Test_certificates.test)
+     ; ("Test_tracing", Test_tracing.test)
```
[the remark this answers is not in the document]
jameshensmancitrix: I have made this change

ocaml/libs/tracing/tracing.ml
```diff
@@ -0,0 +1,619 @@
...
+  module GC = struct
+    let lock = Mutex.create ()
+
+    let span_timeout = ref 86400.
+
+    let span_timeout_thread = ref None
+
+    let gc_inactive_spans () =
+      Xapi_stdext_threads.Threadext.Mutex.execute lock (fun () ->
+          Seq.iter
```
edwintorok, line 255: This still has the same problem that the hashtable is being modified as you're iterating through it. Seq.iter is sort of a lazy iterator fetching elements one at a time.

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
edwintorok: iterating over a copy of the keys might be safer yes, see [here ](https://github.com/xapi-project/xen-api/pull/4958/files#r1177850813)

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: What would be a good way to get around this issue

edwintorok: Try `Hashbtl.filter_map_inplace`: it safely iterates over all elements and you get to decide whether to keep it or not by returning 'Some v' or 'None'.
Also check the logic, because it looks like only the debug message is under the 'if' scope and mark_as_finished is always called on all elements currently, was that intended?

jameshensmancitrix: I have added that in. I divided `marked_finished` into two functions, one of which removes the span from the hash-table and one that adds it to the second hash-table. 4d0cbfa2b108943df5691958342af6f7b7d42511 for the new implementation.

edwintorok (approved):
I can't spot more bugs in the hashtable code now.

PR #4961, opened 2023-04-06
rrdd_proxy: compare the localhost uuid with a uuid instead of a ref

psafont:
This incorrect comparison made the collection fail for vms running in the coordinator host as it forced to request for the coordinator's IP


ocaml/xapi/rrdd_proxy.ml
```diff
@@ -98,7 +98,7 @@ let get_vm_rrd_forwarder (req : Http.Request.t) (s : Unix.file_descr) _ =
           let localhost_uuid = Helpers.get_localhost_uuid () in
           let vm_ref = Db.VM.get_by_uuid ~__context ~uuid:vm_uuid in
           let owner = Db.VM.get_resident_on ~__context ~self:vm_ref in
-          let owner_uuid = Ref.string_of owner in
+          let owner_uuid = Db.Host.get_uuid ~__context ~self:owner in
```
robhoes: It's a little more efficient to use `Helpers.get_localhost` above to get a ref, which avoids the additional DB call. `Helpers.get_localhost` caches the ref, so it is quick.

robhoes (approved):

ocaml/xapi/rrdd_proxy.ml
psafont: Yeah, that simplifies the code quite a bit

lindig (approved):

PR #4962, opened 2023-04-06
CA-376319: Ensure that nbd_client_manager cannot block forever.

MarkSymsCtx:
This has been seen to occur in the prescence of a kernel bug which resulted in the NBD device state not being correctly updated. The fix for which is obviously to update the kernel to not have the bug, but also make the calling code robust to the condition and not block for an infinite amount of time.

Remove the lock surrounding the disconnect operation as it is are not required and provides no benefit. Locks are only required to prevent contention between multiple clients trying to acquire the same NBD device having determined that it currently unused.

### DIFF
--- a/scripts/nbd_client_manager.py
+++ b/scripts/nbd_client_manager.py
@@ -14,12 +14,16 @@ import time
 import fcntl
 import json
 import re
+from datetime import datetime, timedelta
 
 
 LOGGER = logging.getLogger("nbd_client_manager")
 LOGGER.setLevel(logging.DEBUG)
 
 LOCK_FILE = '/var/run/nonpersistent/nbd_client_manager'
+
+# Don't wait more than 10 minutes for the NBD device
+MAX_DEVICE_WAIT_MINUTES = 10
 
 
 class NbdDeviceNotFound(Exception):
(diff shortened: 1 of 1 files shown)


scripts/nbd_client_manager.py
```diff
@@ -123,7 +127,14 @@ def _find_unused_nbd_device():
 
 
 def _wait_for_nbd_device(nbd_device, connected):
+    deadline = datetime.now() + timedelta(minutes=MAX_DEVICE_WAIT_MINUTES)
```
edwintorok, line 130: would be good to use a monotonic clock here, otherwise when the host's clock is changed this could still hang for a long time: https://docs.python.org/3/library/time.html#time.monotonic.
Although that might have to wait until this script is converted to Python3 (if it isn't already), or until we no longer need this script (e.g. because we make vhd-util talk NBD directly).

edwintorok (approved):

robhoes (approved):

scripts/nbd_client_manager.py
MarkSymsCtx: We do have monotonic (as SM uses it) but the likelihood of it hitting the issue is so low (in the absence) of the Kernel bug that it's not worthing stressing over.

PR #4963, opened 2023-04-07
CA-376294: Extract hostname from FQDN

liulinC:
The local hostname may be set to FQDN, which is likely to exceed the netbios limit (15 chars), Then the netbios is compressed and randomized to meet the limit.
However, user may pre-create the machine account to join domain with the hostname part of FQDN, thus we need to extract the hostname from the FQDN as netbios to join domain.


lindig:
The hostname part of the FQDN could still exceed 15 characters (but it's not likely). Should this case not be logged with a warning?

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -991,10 +991,17 @@ module Winbind = struct
     in
     String.init len (fun _ -> random_char ())
 
-  let build_netbios_name () =
+  let build_netbios_name localhost_name =
     (* Winbind follow https://docs.microsoft.com/en-US/troubleshoot/windows-server/identity/naming-conventions-for-computer-domain-site-ou#domain-names to limit netbios length to 15
      * Compress the hostname if exceed the length *)
-    let hostname = get_localhost_name () in
+    let netbios_sep = '.' in
```
lindig, line 997: Is this name misleading? The dot is the separator in the FQDN and this code extracts the hostname from the FQDN.

It would be better to rely on a library to split up a FQDN. We already have 

* https://ocaml.org/u/36c4fc6563935290671863522bba451e/domain-name/0.4.0/doc/Domain_name/index.html

in our dependencies.

liulinC:
> The hostname part of the FQDN could still exceed 15 characters (but it's not likely). Should this case not be logged with a warning?

This information can be found directly and accurately in the `net ads join` command, 
but let's log it in case of using compress hostname.

lindig (approved):

ocaml/xapi/extauth_plugin_ADwinbind.ml
```diff
@@ -991,17 +991,32 @@ module Winbind = struct
...
+    in
+    if String.length hostname > max_netbios_name_length then (
       (* format hostname to prefix-random each with 7 chars *)
       let len = 7 in
       let prefix = String.sub hostname 0 len in
       let suffix = random_string len in
-      Printf.sprintf "%s-%s" prefix suffix
-    else
+      let netbios_name = Printf.sprintf "%s-%s" prefix suffix in
+      warn "hostname length exceed limit, use '%s' instead" netbios_name ;
```
psafont, line 1017:
```suggestion
      warn "hostname exceeds allowed length, using '%s' instead" netbios_name ;
```

psafont:
Aside from the comment on the new addition this looks good

psafont (approved):

PR #4970, opened 2023-04-20
CP-42642: Support share server certificate file to group users

DeliZhangX:
- Currently /etc/xensource/xapi-ssl.pem is user (default root) readonly, to share to other users of one group add the setting "server-cert-group-id".
- gencert.service can not read xapi config file, so feed the goup id in service unit file directly. In future can consider to make "xapi_globs" as common library so that gencert.service can read "server-cert-group-id" directly.
- The group id -1 means the default group won't be changed.

### DIFF
--- a/ocaml/gencert/gencert.ml
+++ b/ocaml/gencert/gencert.ml
@@ -41,7 +41,7 @@ let generate_cert_or_fail ~generator ~path =
     exit 1
   )
 
-let main ~dbg ~path ~sni () =
+let main ~dbg ~path ~cert_gid ~sni () =
   let valid_for_days = 365 * 10 in
   let init_inventory () = Inventory.inventory_filename := inventory in
   init_inventory () ;
@@ -60,12 +60,13 @@ let main ~dbg ~path ~sni () =
         let ips = [ip] in
         let (_ : X509.Certificate.t) =
           Gencertlib.Selfcert.host ~name ~dns_names ~ips ~valid_for_days path
+            cert_gid
         in
         ()
     | SNI.Xapi_pool ->
         let uuid = Inventory.lookup Inventory._installation_uuid in
         let (_ : X509.Certificate.t) =
-          Gencertlib.Selfcert.xapi_pool ~valid_for_days ~uuid path
+          Gencertlib.Selfcert.xapi_pool ~valid_for_days ~uuid path cert_gid
         in
         ()
   in
@@ -77,13 +78,13 @@ let () =
   (* if necessary use Unix.localtime to debug *)
   D.debug "%s" dbg ;
   match Sys.argv with
-  | [|_; path; _|] when Sys.file_exists path ->
+  | [|_; path; _; _|] when Sys.file_exists path ->
       D.info "file already exists at path (%s) - doing nothing" path ;
       exit 0
-  | [|_; path; sni|] -> (
+  | [|_; path; cert_gid; sni|] -> (
     match SNI.of_string sni with
     | Some sni ->
-        main ~dbg ~path ~sni ()
+        main ~dbg ~path ~cert_gid:(int_of_string cert_gid) ~sni ()
     | None ->
         D.error "SNI must be default or xapi:pool, but got '%s'" sni ;
         exit 1
--- a/scripts/gencert.service
+++ b/scripts/gencert.service
@@ -6,8 +6,8 @@ After=forkexecd.service
 [Service]
 User=root
 Type=oneshot
-ExecStart=/opt/xensource/libexec/gencert /etc/xensource/xapi-pool-tls.pem xapi:pool
-ExecStart=/opt/xensource/libexec/gencert /etc/xensource/xapi-ssl.pem default
+ExecStart=/opt/xensource/libexec/gencert /etc/xensource/xapi-pool-tls.pem -1 xapi:pool
+ExecStart=/opt/xensource/libexec/gencert /etc/xensource/xapi-ssl.pem -1 default
 RemainAfterExit=yes
 
 [Install]
--- a/ocaml/gencert/lib.ml
+++ b/ocaml/gencert/lib.ml
@@ -125,7 +125,7 @@ let validate_certificate kind pem now private_key =
     )
 
 let install_server_certificate ~pem_chain ~pem_leaf ~pkcs8_private_key
-    ~server_cert_path =
+    ~server_cert_path ~cert_gid =
   let now = Ptime_clock.now () in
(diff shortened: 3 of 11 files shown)


lindig (approved):
Do we really need this much machinery? Why not create a Unix group for all users that can read the file and create the file with this group and readable for the group?

ocaml/gencert/gencert.ml
```diff
@@ -77,13 +78,13 @@ let () =
...
-  | [|_; path; _|] when Sys.file_exists path ->
+  | [|_; path; _; _|] when Sys.file_exists path ->
       D.info "file already exists at path (%s) - doing nothing" path ;
       exit 0
-  | [|_; path; sni|] -> (
+  | [|_; path; cert_gid; sni|] -> (
     match SNI.of_string sni with
     | Some sni ->
-        main ~dbg ~path ~sni ()
+        main ~dbg ~path ~cert_gid:(int_of_string cert_gid) ~sni ()
```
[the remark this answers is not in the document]
DeliZhangX: The group "certusers" in XS is already created https://code.citrite.net/projects/XSC/repos/xenserver-setup/commits/8cbe1c1a5cf76c97ddf752a0561b46de418940e1, how to manage users and group depends on specific hypervisor release, here upstream xapi just provide the mechanism to allow set the file group.

psafont:
I have the same question as christian, why is this so complicated?

> In future can consider to make "xapi_globs" as common library so that gencert.service can read "server-cert-group-id" directly.

I would prefer this was done in this PR: read the xapi_globs on the function that writes the host certificates instead of threading the groupid through so many functions.

DeliZhangX:
> 

The reason is realistic consideration, that are the project time limit, I don't know much about xapi and there are more than 800 references to `xapi_globs`, updating them is also risky.

psafont:
I've reviewed the position of xapi globs with respect of the gencert binary.
The library cannot be used directly as xapi_globs is part of xapi_internal, as it can only be used from the xapi binary.
This unfortunately forces the group id to be threaded to the function that calls the chown.

Having said this:
- gencert uses gid  -1 by default, this forces the permission to be read-only for root, not the group. Why can't the permissions to allow the gid to read it be set unconditionally?

lindig:
Do we know all files that would be affected? Why not use the correct permissions from now on for new files and also iterate over existing (known) files and change their permissions at startup?

psafont:
There's only one file affected: `/etc/xensource/xapi-ssl.pem`. The problem is that there are 2 binaries that write to this file, on at first boot, when the file is missing (gencert), and another when xapi is requested to install a new one. So ensuring the permissions on startup is not enough

DeliZhangX:
You can see `UX.atomic_write_to_file` implement, it creates temp file and rename to original file, it keeps permission but ownership. So that group back to default (root) when xapi-ssl.pem is reset/installed.
`Xapi_stdext_unix.Unixext` is (or will be) deprecated, so I did not make change there.

The default gid -1 is to ensure compatibility for existing hypervisor env. E.g. one customer has xapi-ssl.pem with old ownership of `root:root`, then xapi updated, xapi-ssl.pem will keep the ownership even if customer installed new certificate.
In XS, I will build new xapi rpm where set the gid to 204 (certusers group), so that NRPE service can read xapi-ssl.pem. (Also when NRPE service is enabled, customer is asked to run one script of NRPE package which chmod 440 and chgrp certusers on the existing xapi-ssl.pem)

psafont (approved):

scripts/gencert.service
```diff
@@ -6,8 +6,8 @@ After=forkexecd.service
 [Service]
 User=root
 Type=oneshot
-ExecStart=/opt/xensource/libexec/gencert /etc/xensource/xapi-pool-tls.pem xapi:pool
-ExecStart=/opt/xensource/libexec/gencert /etc/xensource/xapi-ssl.pem default
+ExecStart=/opt/xensource/libexec/gencert /etc/xensource/xapi-pool-tls.pem -1 xapi:pool
+ExecStart=/opt/xensource/libexec/gencert /etc/xensource/xapi-ssl.pem -1 default
```
bernhardkaindl, line 10: Won't this change of the argument list of /opt/xensource/libexec/gencert be wirde consequences?

By inserting the group id as 2nd argument before the (default|xapi:pool), this changes the API and affects all existing users, scripts, XenRT and other automation tools which update certificates? (e.g. internal and maybe by special customers which use automated ssh on Dom0 like the IBM Cloud)?

So far, I found these tools which have the string /opt/xensource/libexec/gencert in them, so it looks like they would be affected by this change:

```py
grep -r /opt/xensource/libexec/gencert /opt/xensource|cut -d' ' -f 3
/opt/xensource/debug/perftest
/opt/xensource/debug/event_listen
/opt/xensource/debug/vncproxy
/opt/xensource/debug/quicktestbin
/opt/xensource/bin/xe
/opt/xensource/bin/xapi-db-process
/opt/xensource/bin/xsh
/opt/xensource/bin/rrd2csv
/opt/xensource/bin/xapi
/opt/xensource/bin/mpathalert
/opt/xensource/libexec/v6d-citrix
/opt/xensource/libexec/alert-certificate-check
/opt/xensource/libexec/daily-license-check
/opt/xensource/libexec/gencert
/opt/xensource/libexec/xcp-rrdd-plugins/xcp-rrdd-squeezed
/opt/xensource/libexec/xcp-rrdd-plugins/xcp-rrdd-xenpm
/opt/xensource/libexec/xcp-rrdd-plugins/xcp-rrdd-gpumon
/opt/xensource/libexec/xcp-rrdd-plugins/xcp-rrdd-iostat
```

psafont: > all existing users, scripts, XenRT and other automation tools which update certificates? 

If there are automation tools calling the script they are using the product in an unintended way. The only supported way to change the certificate is though xapi, the only intended user of gencert. If there is a usecase where the APIs are not enough we need to have a conversation with the clients so xapi can cater to it

> Also, couln't have this been done by a simple Exec of chmod 204 as a separate Exec call after the creation of the certificate? Then this change to Xapi could be avoided completely and only gencert.service would be updated.

Unfortunately, I don't think so. xapi calls the certificate library as well to bve able to install custom certificates (not just generate a new one), so it needs to be modified to know the gid as well. It is also used to control the pool-internal certificates

> So far, I found these tools which have the string /opt/xensource/libexec/gencert in them, so it looks like they would be affected by this change:

I believe most of them are false positives, they contain a library with this string, it doesn't mean they use it in a meaningful way

PR #4975, opened 2023-04-23
Move writing init complete to the end of startup sequence

liulinC:
The init complete file should be touched at the end of startup sequence to state the complete of init

Test:
Toolstack-input: Karma test is happy with it. 182471 (Dev Run)

### DIFF
--- a/ocaml/xapi/xapi.ml
+++ b/ocaml/xapi/xapi.ml
@@ -1291,10 +1291,6 @@ let server_init () =
             , []
             , fun () -> Create_misc.create_pool_cpuinfo ~__context
             )
-          ; ( "writing init complete"
-            , []
-            , fun () -> Helpers.touch_file !Xapi_globs.init_complete
-            )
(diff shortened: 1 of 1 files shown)


liulinC:
Is there any format update recently?  
Lots of unrelated files are complained for format error.

robhoes:
This change looks like it makes a lot of sense. However, I'm always nervous when changing the startup sequence, as there are often hidden dependencies somewhere in the system.

Looking through the history, already in 2007 the waiting for management interface was added after writing init complete. Without explanation, unfortunately. I suspect that many of the other items were simply added to the end of the list without considering the init complete signal.

I am tempted to accept this change, run a variety of  additional tests, and fix any failures that might be found.

ocaml/xapi/xapi.ml
```diff
@@ -1422,6 +1418,10 @@ let server_init () =
                 Xapi_host.write_uefi_certificates_to_disk ~__context
                   ~host:(Helpers.get_localhost ~__context)
             )
+          ; ( "writing init complete"
```
lindig, line 1421: Add a comment that this has to stay the last entry and new entries have to be added before.

PR #4979, opened 2023-04-25
CA-376271: Block vTPM creation on BIOS booted VMs

GeraldEV:
Block the creation of vTPMs on VMs with BIOS firmware in HVM-boot-params (or BIOS by default)
Moved firmware_of_vm and nvram_uefi_of_vm into new module to resolve dependency loop between xapi_vtpm.ml and xapi_xenops.ml - is this the correct way to resolve this?

Manually tested by uploading the built xapi to a host and attempting to create a vTPM:
```
[root@cl01-03 ~]# xe vtpm-create vm-uuid=f571526b-f1d4-ab2f-7052-2e22ed0093d3
VTPMs cannot be created on BIOS booted VMs
```
Confirmed that vTPMs can still be created on UEFI booted hosts
Is there any more testing required?


stormi:
I haven't read the code, but what happens if someone tries to change the boot firmware from UEFI to BIOS *after* having added a vTPM to the VM?

GeraldEV:
The check already exists in XenCenter and is not needed here - closing PR and ticket as not a bug
Thanks for your feedback!

PR #4980, opened 2023-04-27
CA-373074 Fix Fingerprint not writing on boot 

jameshensmancitrix:
An internal test case flags that the host fingerprint is not being found by the testcase. 

The update is done by the script `update_getty`, however if the certificate is not generated then this process will not run properly. This change runs this script at the end of gencert starting.

### DIFF
--- a/scripts/gencert.service
+++ b/scripts/gencert.service
@@ -8,6 +8,8 @@ User=root
 Type=oneshot
 ExecStart=/opt/xensource/libexec/gencert /etc/xensource/xapi-pool-tls.pem -1 xapi:pool
 ExecStart=/opt/xensource/libexec/gencert /etc/xensource/xapi-ssl.pem -1 default
+ExecStartPost=/sbin/update-issue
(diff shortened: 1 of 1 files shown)


lindig:
Is the context that cert generation sometimes fails (typically because of a lack of IP address) and then the fingerprint is not generated? Would like to understand the conditions why the problem of no fingerprint exists. Can we be sure that gencert properly defines the host to be ready? If so this is an elegant fix

jameshensmancitrix:
The problem that we are getting on rt-next is that at the time Xapi attempts to write the finger print to `/etc/issue` gencert is not running so the fingerprint is not written and this test case fails. The fix here places the script to update the file as a script to run after the services has started to ensure this doesn't happen.

edwintorok (approved):
I'm not familiar with how this is supposed to work, but if `update-issue` depends on input that changes when `gencert` is run then rerunning `update-issue` whenever its input changes seems like the right thing to do

psafont (approved):

PR #4981, opened 2023-04-27
CP-40528 VTPM snapshot, revert, clone

lindig:
Implement VM snapshot/revert and VM clone for VMs that have VTPMs.

A VM snapshot is a VM record with an additional field snapshot_metadata
that contains all metadata of the VM we are taking a snapshot of. This
metadata includes a reference to the VTPM the VM is/was using. In
addition, a new VTPM as a copy of the existing one is created and referred
to by the VM record that represents the snapshot. So a snapshot references
two VTPM: one directly (A) and one via snapshot_metadata (B).

Since the VM that took the snapshot is in continued use, it will continue
to update B but A is stable.

When reverting the snapshot we create a new VM from the VM record that
represents the snapshot. This can happen again and again.  Unlike a clone,
reverting a snapshot does not consume it.

As a consequence, the VM we are creating can't use VTPM A (and modify
it) but we create a copy of A and use that. This currently means that
the UUID of the VTPM changes - this is fixed in a later commit.

To help development, this patch extends the fields listed in "xe
vtpm-list" by showing the size and MD5 hash of the content. Otherwise
tracking this is difficult because the content is not stored in the VTPM
object but it refers to another object of class secret that holds the
actual content.

When we revert a VM snapshot, we:

* delete the VMs current VTPM * make a copy of the snapshot VTPM *
assign the copy to the VM

The copy (of the VTPM) has a UUID that is different from the UUID before
the revert. But we can maintain the same UUID because the VTPM before
the revert operation is removed. So we remember that UUID and assign it
to the copy we just created.

### DIFF
--- a/ocaml/xapi/xapi_vm_snapshot.ml
+++ b/ocaml/xapi/xapi_vm_snapshot.ml
@@ -397,6 +397,7 @@ let do_not_copy =
   ; "VIFs"
   ; "VGPUs"
   ; "VUSBs"
+  ; "VTPMs"
   ; (* Stateful fields that will be reset anyway *)
     "power_state"
   ; (* Attached PCIs should not revert from snapshot *)
@@ -478,6 +479,35 @@ let revert_vm_fields ~__context ~snapshot ~vm =
     ~overrides ;
   TaskHelper.set_progress ~__context 0.1
 
+(* See CP-40528. A snapshot is a VM record that contains the record of
+   the VM at snapshot time in snapshot_metadata, including its VTPM.
+   Most data is restored from snapshot_metadata. At the time the
+   snapshot is taken, a copy of the VTPM is taken and this is part of
+   the VM record. We want to use a copy of this VTPM. We need to make a
+   copy because a snapshot can be restored multiple times and we have to
+   ensure the VTPM of the snapshot remains unaltered. *)
+
+let update_vtpm ~__context ~snapshot ~vm =
+  debug "%s for VM %s (1/2)" __FUNCTION__ (Ref.string_of vm) ;
+  let vtpms = Db.VM.get_VTPMs ~__context ~self:vm in
+  (* destroy current VTPM, but don't fail if it doesn't exist *)
+  List.iter
+    (fun vtpm ->
+      if Db.is_valid_ref __context vtpm then
+        Xapi_vtpm.destroy ~__context ~self:vtpm
+    )
+    vtpms ;
+  (* assign copy of snapshot VTPM to VM *)
+  let _vtpms =
+    Db.VM.get_VTPMs ~__context ~self:snapshot
+    |> List.map (fun vtpm -> Xapi_vtpm.copy ~__context ~vM:vm vtpm)
+  in
+  (* A restored VTPM has a different UUID than the one that took the
+     snapshot from. But there is no easy way to maintain it. In
+     particular, the current VM could have a different number of VTPMs
+     than the snapshot. So code can't rely on this anyway. *)
+  debug "%s for VM %s (2/2)" __FUNCTION__ (Ref.string_of vm)
+
 let revert ~__context ~snapshot ~vm =
   debug "Reverting %s to %s" (Ref.string_of vm) (Ref.string_of snapshot) ;
   (* This is destructive and relatively fast. There's no point advertising cancel since it
@@ -489,6 +519,7 @@ let revert ~__context ~snapshot ~vm =
     update_guest_metrics ~__context ~snapshot ~vm ;
     update_metrics ~__context ~snapshot ~vm ;
     update_parent ~__context ~snapshot ~vm ;
+    update_vtpm ~__context ~snapshot ~vm ;
     TaskHelper.set_progress ~__context 1. ;
     Xapi_vm_lifecycle.force_state_reset ~__context ~self:vm ~value:power_state ;
     debug "VM.revert done"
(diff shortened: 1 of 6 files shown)


ocaml/xapi/xapi_globs.ml
```diff
@@ -930,7 +930,7 @@ let repository_gpgkey_name = ref ""
 
 let repository_gpgcheck = ref true
 
-let ignore_vtpm_unimplemented = ref false
+let ignore_vtpm_unimplemented = ref true
```
psafont, line 933: can this be turned to false and removed in the snapshot case?

lindig: So keep it and just remove the checks? Will look for it

ocaml/xapi/xapi_vm_snapshot.ml
```diff
@@ -487,15 +487,35 @@ let revert_vm_fields ~__context ~snapshot ~vm =
    copy because a snapshot can be restored multiple times and we have to
    ensure the VTPM of the snapshot remains unaltered. *)
 
+(** zip pairs two lists into a list of pairs *)
+let rec zip xs ys =
```
psafont, line 491: This is [List.combine](https://v2.ocaml.org/api/List.html#VALcombine)

ocaml/xapi/xapi_vm_snapshot.ml
```diff
@@ -487,15 +487,35 @@ let revert_vm_fields ~__context ~snapshot ~vm =
...
   (* destroy current VTPM *)
-  Db.VM.get_VTPMs ~__context ~self:vm
-  |> List.iter (fun vtpm -> Xapi_vtpm.destroy ~__context ~self:vtpm) ;
+  List.iter (fun vtpm -> Xapi_vtpm.destroy ~__context ~self:vtpm) vtpms ;
   (* use copy of snapshot VTPM *)
   Db.VM.get_VTPMs ~__context ~self:snapshot
   |> List.iter (fun vtpm -> Xapi_vtpm.copy ~__context ~vM:vm vtpm |> ignore) ;
-  debug "%s for VM %s (2/2)" __FUNCTION__ (Ref.string_of vm)
+  (* reset UUIDs *)
+  let vtpms = Db.VM.get_VTPMs ~__context ~self:vm in
```
psafont: What do you think about using?
```suggestion
  let vtpms =
    Db.VM.get_VTPMs ~__context ~self:snapshot
    |> List.map (fun vtpm -> Xapi_vtpm.copy ~__context ~vM:vm vtpm)
  in
  (* reset UUIDs *)
```

psafont (approved):
This needs a squash, and removing the empty commit, otherwise it looks good to go

lindig:
* disabled cluster test because it fails on Koji CA-376993
* moved all code into one commit
* auto format asked for some small changed of existing code

ocaml/xapi/xapi_vm_snapshot.ml
```diff
@@ -478,6 +479,36 @@ let revert_vm_fields ~__context ~snapshot ~vm =
...
+  (* destroy current VTPM *)
+  List.iter (fun vtpm -> Xapi_vtpm.destroy ~__context ~self:vtpm) vtpms ;
+  (* use copy of snapshot VTPM *)
+  let vtpms =
+    Db.VM.get_VTPMs ~__context ~self:snapshot
+    |> List.map (fun vtpm -> Xapi_vtpm.copy ~__context ~vM:vm vtpm)
+  in
+  (* reset UUIDs *)
+  List.combine vtpms uuids
+  |> List.iter (fun (vtpm, uuid) ->
```
robhoes, line 505: I think this makes two assumptions:
* That the current VM and its snapshot have the same number of VTPM objects.
* That the order of the VTPMs coming back from `Db.VM.get_VTPMs` is the same for the VM and for its snapshot.

I think that the first one in particular is not true in general, as one can easily add or remove a VTPM after creating a snapshot.

lindig: Currently you can have at most one VTPM; but you are right that the current VM might have no VTPM - so this needs more careful coding; it might be necessary to remember the UUID of the VTPM that we snapshot as we otherwise might not be able to re-create it

ocaml/xapi/xapi_vm_snapshot.ml
```diff
@@ -478,6 +479,36 @@ let revert_vm_fields ~__context ~snapshot ~vm =
...
+   ensure the VTPM of the snapshot remains unaltered. *)
+
+let update_vtpm ~__context ~snapshot ~vm =
+  debug "%s for VM %s (1/2)" __FUNCTION__ (Ref.string_of vm) ;
+  let vtpms = Db.VM.get_VTPMs ~__context ~self:vm in
+  let uuids =
+    vtpms |> List.map (fun self -> Db.VTPM.get_uuid ~__context ~self)
+  in
+  (* destroy current VTPM *)
+  List.iter (fun vtpm -> Xapi_vtpm.destroy ~__context ~self:vtpm) vtpms ;
```
robhoes, line 497: The function `update_vifs_vbds_vgpus_and_vusbs` above uses `safe_destroy_X` functions that don't raise an error if the ref no longer exists for some reason. I suppose this is to make reverting to a snapshot robust if the current VM is corrupted for some reason.

ocaml/xapi/xapi_vm_snapshot.ml
robhoes: The code in `update_vifs_vbds_vgpus_and_vusbs` seems to make no attempt to preserve the UUID for the other devices upon reverting a VM.

robhoes (approved):

PR #4982, opened 2023-04-27
`rrdd-plugin`: do not write payload if page count is 0

benjamreis:
`page_count` is used by `xcp-rrdd` to define the size to allocate to a `cstruct` to write the payload into.

If `page_count=0` the `cstruct` has a size of 0 and the attempt of writing the payload results in a
`not enough memory` error


benjamreis:
This happened for instance on a 8.2 host using a mocked gpumon

psafont (approved):
I wonder if it would be better to simply not execute the loop further down and simply terminate if the page_count is 0, it cannot be changed after calling start_local

lindig (approved):
Nice find

PR #4991, opened 2023-05-03
CA-376144: handle host.apply_recommended_guidances first on pool coordinator

gangj:
Instead of proxying the request directly to pool member, first handler the
    request on pool coordinator, only proxy "VM.pool_migrate", "Host.reboot",
    "Host.restart_agent" to pool member.

    This resolves issue introduced in 022588de52d567c0d6a6b85d328362ae8de6af75

### DIFF
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -797,6 +797,13 @@ let is_pool_master ~__context ~host =
   let master_id = Db.Host.get_uuid ~__context ~self:master in
   host_id = master_id
 
+let assert_we_are_master ~__context =
+  if not (is_pool_master ~__context ~host:(get_localhost ~__context)) then
+    raise
+      Api_errors.(
+        Server_error (host_is_slave, [Pool_role.get_master_address ()])
+      )
+
 (* Host version compare helpers *)
 let compare_int_lists : int list -> int list -> int =
  fun a b ->
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -798,10 +798,11 @@ let is_pool_master ~__context ~host =
   host_id = master_id
 
 let assert_we_are_master ~__context =
-  if not (is_pool_master ~__context ~host:(get_localhost ~__context)) then
+  let host = get_localhost ~__context in
+  if not (is_pool_master ~__context ~host) then
     raise
       Api_errors.(
-        Server_error (host_is_slave, [Pool_role.get_master_address ()])
+        Server_error (host_is_slave, [Db.Host.get_address ~__context ~self:host])
       )
 
 (* Host version compare helpers *)
--- a/ocaml/xapi/helpers.ml
+++ b/ocaml/xapi/helpers.ml
@@ -798,11 +798,10 @@ let is_pool_master ~__context ~host =
   host_id = master_id
 
 let assert_we_are_master ~__context =
-  let host = get_localhost ~__context in
-  if not (is_pool_master ~__context ~host) then
+  if not (is_pool_master ~__context ~host:(get_localhost ~__context)) then
     raise
       Api_errors.(
-        Server_error (host_is_slave, [Db.Host.get_address ~__context ~self:host])
+        Server_error (host_is_slave, [Pool_role.get_master_address ()])
       )
 
 (* Host version compare helpers *)
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -3982,10 +3982,10 @@ functor
             Client.Host.set_https_only ~rpc ~session_id ~self ~value
         )
 
-      let apply_recommended_guidances ~__context ~host =
-        let uuid = host_uuid ~__context host in
+      let apply_recommended_guidances ~__context ~self =
+        let uuid = host_uuid ~__context self in
         info "Host.apply_recommended_guidances: host = %s" uuid ;
-        Local.Host.apply_recommended_guidances ~__context ~host
+        Local.Host.apply_recommended_guidances ~__context ~self
     end
 
     module Host_crashdump = struct
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -3984,11 +3984,8 @@ functor
 
       let apply_recommended_guidances ~__context ~self =
         let uuid = host_uuid ~__context self in
-        info "Host.apply_recommended_guidance: host = %s" uuid ;
-        let local_fn = Local.Host.apply_recommended_guidances ~self in
-        do_op_on ~local_fn ~__context ~host:self (fun session_id rpc ->
-            Client.Host.apply_recommended_guidances ~rpc ~session_id ~self
-        )
+        info "Host.apply_recommended_guidances: host = %s" uuid ;
+        Local.Host.apply_recommended_guidances ~__context ~self
     end
 
     module Host_crashdump = struct
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -3982,10 +3982,10 @@ functor
             Client.Host.set_https_only ~rpc ~session_id ~self ~value
         )
 
-      let apply_recommended_guidances ~__context ~self =
-        let uuid = host_uuid ~__context self in
+      let apply_recommended_guidances ~__context ~host =
+        let uuid = host_uuid ~__context host in
         info "Host.apply_recommended_guidances: host = %s" uuid ;
-        Local.Host.apply_recommended_guidances ~__context ~self
+        Local.Host.apply_recommended_guidances ~__context ~host
     end
 
     module Host_crashdump = struct
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -3068,13 +3068,21 @@ let apply_recommended_guidances ~__context ~host =
     | [] ->
         try_restart_device_models_for_recommended_guidances ~__context ~host
     | [`reboot_host] ->
-        Helpers.call_api_functions ~__context (fun rpc session_id ->
-            Client.Client.Host.reboot ~rpc ~session_id ~host
+        let local_fn = reboot ~host in
+        Xapi_host_helpers.with_host_operation ~__context ~self:host
+          ~doc:"Host.reboot" ~op:`reboot (fun () ->
+            Message_forwarding.do_op_on ~local_fn ~__context ~host (fun session_id rpc ->
+                Client.Client.Host.reboot ~rpc ~session_id ~host
+            )
         )
     | [`restart_toolstack] ->
         try_restart_device_models_for_recommended_guidances ~__context ~host ;
-        Helpers.call_api_functions ~__context (fun rpc session_id ->
-            Client.Client.Host.restart_agent ~rpc ~session_id ~host
+        let local_fn = restart_agent ~host in
+        Xapi_host_helpers.with_host_operation ~__context ~self:host
+          ~doc:"Host.reboot" ~op:`reboot (fun () ->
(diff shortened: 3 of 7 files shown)


ocaml/xapi/xapi_host.ml
```diff
@@ -3055,18 +3058,23 @@ let try_restart_device_models_for_recommended_guidances ~__context ~host =
       None
 
 let apply_recommended_guidances ~__context ~self =
+  (* This function runs on master host *)
```
psafont, line 3061: While the comment is helpful, it's better to enforce this property instead with code, for example an assertion that can raise an internal server error if the function is not run in the coordinator

minglumlu (approved):

ocaml/xapi/xapi_host.ml
```diff
@@ -3055,18 +3060,24 @@ let try_restart_device_models_for_recommended_guidances ~__context ~host =
       None
 
 let apply_recommended_guidances ~__context ~self =
```
psafont, line 3062: Is it possible to change ~self to ~host? I'm not sure if the API framework allows for it, it would make the code in this functions easier to read

psafont, line 3062: It's a bit confusing to use the name 'self' for an external host.
It's possible to rename self to host in the function like this:
```suggestion
let apply_recommended_guidances ~__context ~self:host =
```

psafont:
Changes look good, only a small naming issue that makes reading apply_recommended_guidances more difficult than it should be

ocaml/xapi/xapi_host.ml
```diff
@@ -3055,18 +3060,24 @@ let try_restart_device_models_for_recommended_guidances ~__context ~host =
...
-        reboot ~__context ~host:self
+        Helpers.call_api_functions ~__context (fun rpc session_id ->
+            Client.Client.Host.reboot ~rpc ~session_id ~host:self
+        )
     | [`restart_toolstack] ->
         try_restart_device_models_for_recommended_guidances ~__context
           ~host:self ;
-        restart_agent ~__context ~host:self
+        Helpers.call_api_functions ~__context (fun rpc session_id ->
+            Client.Client.Host.restart_agent ~rpc ~session_id ~host:self
```
psafont, line 3079: I expect this to close the connection, because xapi will be restarted, is this handled in a graceful way here, does it appear as an error instead?

ocaml/xapi/xapi_host.ml
```diff
@@ -3055,18 +3060,24 @@ let try_restart_device_models_for_recommended_guidances ~__context ~host =
...
+  Helpers.assert_we_are_master ~__context ;
   try
     let open Updateinfo in
     Db.Host.get_recommended_guidances ~__context ~self |> function
     | [] ->
         try_restart_device_models_for_recommended_guidances ~__context
           ~host:self
     | [`reboot_host] ->
-        reboot ~__context ~host:self
+        Helpers.call_api_functions ~__context (fun rpc session_id ->
```
psafont, line 3072: It might be worth emulating the do_host_op_on: if the coordinator needs to be restarted it would be nice if it used Local.Host.reboot instead to avoid creating a new connection

psafont:
It's not clear to me how does this change enforce that the coordinator is updated before the other pool members.
In the case where update_guidances is being introduced by the update it makes sense that these cannot be run unless the coordinator is updated first, and the rest of hosts don't need to be aware of it: this is a positive change and is welcome. In the case where other updates are happening there is no order being enforced since it's a per-host operation, not a pool-wide. Should it use a pool-wide operation instead to orchestrate each update or do requirements need users to be in full control when rebooting each host?

minglumlu:
> It's not clear to me how does this change enforce that the coordinator is updated before the other pool members. In the case where update_guidances is being introduced by the update it makes sense that these cannot be run unless the coordinator is updated first, and the rest of hosts don't need to be aware of it: this is a positive change and is welcome. In the case where other updates are happening there is no order being enforced since it's a per-host operation, not a pool-wide. Should it use a pool-wide operation instead to orchestrate each update or do requirements need users to be in full control when rebooting each host?

I don't think this change does anything with the order of hosts in update. The issue to be fixed is that the new API call was going to be forwarded to supporter hosts directly, which would have been running without the new added API call at that time.
The order of hosts in update is ensured by clients, like XenCenter. I think the coordinator will be required to be the first one always.

psafont:
That makes sense, thanks

ocaml/xapi/xapi_host.ml
gangj: Hi @psafont , sorry, I don't know the differences between "renaming with ~self:host" and "change 'self' directly to 'host'", as this is a new added api, so in the fixup: [1979db5](https://github.com/xapi-project/xen-api/pull/4991/commits/1979db5a88bf5eb12bac8ac9e88c39a152acc439), I renamed "self" directly to "host".
Is this fix ok? Or would you please share more about differences, thank you!

ocaml/xapi/xapi_host.ml
gangj: Hi @psafont , I tried a fix with: [c71c4bc](https://github.com/xapi-project/xen-api/pull/4991/commits/c71c4bc51a7fb8cf6109efbcbbe6026a32f223b1), which "copies" the way "Message_forwarding" handles requests. 
But I still have a question, I greped xen-api code, seems almost all requests to other host is handled with "call_api_functions", while only few special requests, like "hello", "ping" is using the "do_op_on", is there some reason for this? Please help to have a review, thank you!

ocaml/xapi/xapi_host.ml
gangj: As `restart_agent` is called with a new thread, I think there will be no error calling `restart_agent`. While in the new thread requesting a restart of xapi or the host will make the connection closed and thus error will arise, I think this is inevitable, or would you please share what is the "graceful way" to handle it? Thank you!

ocaml/xapi/helpers.ml
```diff
@@ -797,6 +797,13 @@ let is_pool_master ~__context ~host =
   let master_id = Db.Host.get_uuid ~__context ~self:master in
   host_id = master_id
 
+let assert_we_are_master ~__context =
+  if not (is_pool_master ~__context ~host:(get_localhost ~__context)) then
+    raise
+      Api_errors.(
+        Server_error (host_is_slave, [Pool_role.get_master_address ()])
```
lindig, line 804: I think the error message is confusing but consistent with the code base: it reports a host is a slave but includes the address of the master.

ocaml/xapi/message_forwarding.ml
```diff
@@ -3982,13 +3982,10 @@ functor
...
-        info "Host.apply_recommended_guidance: host = %s" uuid ;
-        let local_fn = Local.Host.apply_recommended_guidances ~self in
-        do_op_on ~local_fn ~__context ~host:self (fun session_id rpc ->
-            Client.Host.apply_recommended_guidances ~rpc ~session_id ~self
-        )
+      let apply_recommended_guidances ~__context ~host =
+        let uuid = host_uuid ~__context host in
+        info "Host.apply_recommended_guidances: host = %s" uuid ;
+        Local.Host.apply_recommended_guidances ~__context ~host
     end
```
lindig, line 3989: Why is this correct if host not equal to master? This needs an comment why we don't need to forward this to another host.

gangj: This is the purpose of this PR: it is because we "can not" forward this to pool member, as the pool member is running with an old version of XAPI which does not have the new added API call at that time. I am afraid it is had to make it clear with an comment, below is the detailed scenario(as [mentioned in the design doc](https://info.citrite.net/pages/viewpage.action?spaceKey=xenserver&title=Design+of+Toolstack+for+CP-41451#comment-1631270801)):

let's discuss pool upgrading with a pool with 2 hosts: host A as master, B as slave, both with old xapi version (without the new host.apply_recommended_guidances  API):


step 1: XenCenter call host.apply_updates  on master A, as master XAPI is old, after it installed the new version of xapi-core rpm, it restart xapi immediately as recommended guidance of the update. As toolstack is restarted, XenCenter will not get the result of the host.apply_updates . So it will not got the Api_errors.updates_require_recommended_guidance  warning, thus user will not get the recommended guidance prompt.


step 2: After XenCenter reconnect with the pool, it continues to call host.apply_updates  on host B, here is the place which I missed: the implementation of apply_updates  handler is that, the apply_updates  request is not proxied from master A to host B directly, now most of the handling of this request is executed in master, since now master A has been upgraded to new version, it will not apply recommended guidance within host.apply_updates , and the new version XAPI on master will also respond warning: Api_errors.updates_require_recommended_guidance  for the host.apply_updates  API call, then client would try to call the new API: host.apply_recommended_guidances . The currently implementation of host.apply_recommended_guidances  handler is that: master will simply proxy this request to slave, while XAPI running in slave is still old as toolstack on slave have not been restarted yet, then the host.apply_recommended_guidances  request would fail.

ocaml/xapi/helpers.ml
gangj: Yes, this function was moved from `xapi_psr.ml`, had a fix to report the address of current slave in the error.

ocaml/xapi/xapi_host.ml
```diff
@@ -3054,21 +3059,35 @@ let try_restart_device_models_for_recommended_guidances ~__context ~host =
...
+            )
+        )
     | [`restart_toolstack] ->
-        try_restart_device_models_for_recommended_guidances ~__context
-          ~host:self ;
-        restart_agent ~__context ~host:self
+        try_restart_device_models_for_recommended_guidances ~__context ~host ;
+        let local_fn = restart_agent ~host in
+        Xapi_host_helpers.with_host_operation ~__context ~self:host
+          ~doc:"Host.reboot" ~op:`reboot (fun () ->
```
liulinC, line 3083: Why the doc host be `Host.reboot`?

gangj: Removed "Xapi_host_helpers.with_host_operation" here, as it is not used in "Message_forwarding.restart_agent"

liulinC (approved):

liulinC (approved):

ocaml/xapi/xapi_host.ml
gangj: changed back to use call_api_functions instead of do_op_on with [fa8f1b9](https://github.com/xapi-project/xen-api/pull/4991/commits/fa8f1b96fb9ec1df818263b41fe49b1ece8277f5), as the discussion here: https://citrix.slack.com/archives/C7EPZ2JR3/p1683623556253019?thread_ts=1683535636.421609&cid=C7EPZ2JR3

lindig:
I would like to see a review by @robhoes. You obviously thought about this but I am still confused why certain functions are now only executed on pool masters; for example `try_restart_device_models_for_recommended_guidances` looks like it is never executed on a pool slave when previously it was.

ocaml/xapi/helpers.ml
```diff
@@ -797,6 +797,14 @@ let is_pool_master ~__context ~host =
   let master_id = Db.Host.get_uuid ~__context ~self:master in
   host_id = master_id
 
+let assert_we_are_master ~__context =
+  let host = get_localhost ~__context in
+  if not (is_pool_master ~__context ~host) then
+    raise
+      Api_errors.(
+        Server_error (host_is_slave, [Db.Host.get_address ~__context ~self:host])
```
robhoes, line 805: The parameter to HOST_IS_SLAVE must be the coordinator's address, not the local address (as done in the original function in xapi_psr.ml).

robhoes (changes requested):
This is mostly fine. Just two comments.

ocaml/xapi/xapi_host.ml
robhoes: Pau's way left the name of the argument for callers of the function as `self`, while inside the function, the parameter is renamed to `host`. It is better to do that, because the argument in the API should remain `self` for consistency with the rest of the API.

robhoes (approved):

PR #4996, opened 2023-05-09
Drop xen-gnt-unix dependency

edwintorok:
Tested that a 'dune build', 'dune build @check' and 'dune runtest' pass locally.
Haven't yet tested on Koji (might need tweaking the specfile to drop some files).


psafont (approved):

andyhhp:
:shipit:

PR #4998, opened 2023-05-12
CP-43131: Make gvt-g support configurable

liulinC:
_This PR will not be merged until xs9 is ready, 
Please just help to review this PR and approve if you are happy, we will merge it later._

Edit: This PR is updated to make gvt-g support configurable, so it can be merged to master as early as possible, 
The legacy code can be removed later at proper time.

### DIFF
--- a/ocaml/xapi/xapi_vgpu_type.ml
+++ b/ocaml/xapi/xapi_vgpu_type.ml
@@ -14,7 +14,7 @@
 
 let finally = Xapi_stdext_pervasives.Pervasiveext.finally
 
-module D = Debug.Make (struct let name = "xapi" end)
+module D = Debug.Make (struct let name = "xapi_vgpu_type" end)
 
 open D
 
@@ -1069,7 +1069,7 @@ let find_or_create_supported_types ~__context ~pci =
     match vendor_id with
     | x when x = Nvidia.vendor_id ->
         Nvidia.find_or_create_supported_types
-    | x when x = Intel.vendor_id ->
+    | x when x = Intel.vendor_id && !Xapi_globs.gvt_g_supported ->
         Intel.find_or_create_supported_types
     | x when x = AMD.vendor_id ->
         AMD.find_or_create_supported_types
@@ -1090,3 +1090,14 @@ let requires_passthrough ~__context ~self =
       None
 
 (* Does not require any passthrough *)
+
+let assert_not_legacy ~__context ~self =
+  let legacy =
+    (not !Xapi_globs.gvt_g_supported)
+    && Db.VGPU_type.get_implementation ~__context ~self = `gvt_g
+  in
+  if legacy then
+    let vendor = Db.VGPU_type.get_vendor_name ~__context ~self in
+    let model = Db.VGPU_type.get_model_name ~__context ~self in
+    let msg = [Printf.sprintf "%s: %s" vendor model] in
+    raise Api_errors.(Server_error (vgpu_type_no_longer_supported, msg))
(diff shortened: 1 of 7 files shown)


johnelse:
So long, old friend 👋

ocaml/xapi-idl/xen/xenops_interface.ml
```diff
@@ -219,11 +219,7 @@ end
 module Vgpu = struct
   include Xenops_types.Vgpu
 
-  type implementation =
-    | GVT_g of gvt_g
```
psafont, line 223: Can't this be a problem when upgrading?

liulinC: > So long, old friend 👋

So long, old friend 😄

liulinC:
> So long, old friend 👋

So long, old friend 😄

ocaml/xapi-idl/xen/xenops_interface.ml
liulinC: It shouldn't be, 
The VM will be shutdown during RPU, 
After upgrade, xapi will block VM start if the VM has this vGPU assigned, then it will not go this far.

robhoes:
@johnelse is still around and watching PRs here!!

robhoes:
Oh dear... we'd better be more careful about what we are saying here... 😬

liulinC:
I rebase to one single commit as the `design` now has been updated according to @edwintorok , 
We introduce configurable item, and drop/support this vGPU on demand.

liulinC:
> I think the checks in start/resume/resume_on are redundant and would be covered by the 'allocate_vm_to_host' check.
> 
> What happens to the PGPUs though, will XAPI do the right thing on startup and remove the gvt-g ones if not supported due to the change in the 'find_or_create_supported'?

1. Indeed, all the three function will reach `allocate_vm_to_host`, but that is at very quick deep/later stage, with lots of xapi effort wasted, besides that, that happens when xapi to allocate a host to the VM, thus, the error message looks not straight forward, 
```
xe vm-start uuid=5e4ff268-c692-b202-6c23-8186266d59bd 
There are no suitable hosts to start this VM on.
The following table provides per-host reasons for why the VM could not be started:

#    : Cannot start here [Server_error(VGPU_TYPE_NO_LONGER_SUPPORTED, [ Intel Corporation: Intel GVT-g ])]
```
With current early check and block, 
```
# xe vm-start uuid=5e4ff268-c692-b202-6c23-8186266d59bd 
VGPU type is no longer supported
type: Intel Corporation: Intel GVT-g
```

I would like to argue just keep it (yes, we can handle the exception and re-raise at proper place, but that sounds more complicated.)

2.  `gvt-g` is one type of vGPU, passthrough (intel gvt-d) can still wrok well.

edwintorok (approved):

ocaml/xapi/xapi_vgpu_type.ml
```diff
@@ -1090,3 +1090,22 @@ let requires_passthrough ~__context ~self =
...
+
+let assert_not_legacy ~__context ~self =
+  let legacy =
+    Db.VGPU_type.get_implementation ~__context ~self |> fun x ->
+    match (x, !Xapi_globs.gvt_g_supported) with
+    | `gvt_g, false ->
+        true
+    | _ ->
+        false
+  in
```
psafont, line 1102: We want to avoid database queries if possible, since this feature is going to be disabled most of the time I recommend to use short-circuiting to avoid the db: 
```suggestion
  let legacy =
    not !Xapi_globs.gvt_g_supported
    && Db.VGPU_type.get_implementation ~__context ~self = `gvt_g
  in
```

psafont (approved):

liulinC:
`push -f` just squash the history and rebase master before merge.

PR #4999, opened 2023-05-12
ci: update main workflow to setup-ocaml v2

psafont:
This means we don't have to handle caches manually, or manage the
repositories directly


psafont:
This seems to speed up the ocamlformat check (4 minutes from 5 minutes), while it slightly slows down the warm tests (10:44 minutes from 9:30 minutes)

PR #5000, opened 2023-05-12
Minimal swtpm REST API

edwintorok:
TODO:
* plumb creation of new socket through to 'make_server_vtpm_rest'
* add new enum for new storage backend in addition to xapi (maybe call it xapi2?), the serialization is different and incompatible (unless swtpm would implement a linear op instead of the current per-entry ops, but per-entry seems more robust)
* bring back the tests, but this time just test basic operations/scalability, not correctness in a distributed system (we won't have a distributed system anymore, we'll rely on shared storage to solve that problem)

### DIFF
--- a/ocaml/database/redo_log.ml
+++ b/ocaml/database/redo_log.ml
@@ -37,11 +37,12 @@ let get_static_device reason =
   R.debug "Found %d VDIs matching [%s]" (List.length vdis) reason ;
   match vdis with [] -> None | hd :: _ -> hd.Static_vdis_list.path
 
-(* Make sure we have plenty of room for the database *)
-let minimum_vdi_size =
+let mib megabytes =
   let ( ** ) = Int64.mul in
-  let mib = 1024L ** 1024L in
-  256L ** mib
+  Int64.of_int megabytes ** 1024L ** 1024L
+
+(* Make sure we have plenty of room for the database *)
+let minimum_vdi_size, recommended_vdi_size = (mib 256, mib 4096)
 
 let redo_log_sm_config = [("type", "raw")]
 
--- a/ocaml/xapi-guard/lib/varstored_interface.ml
+++ b/ocaml/xapi-guard/lib/varstored_interface.ml
@@ -27,8 +27,6 @@ type nvram = (string * string) list [@@deriving rpcty]
 
 let originator = "varstored-guard"
 
-let version = "0.1"
-
 type session = [`session] Ref.t
 
 type rpc = call -> response Lwt.t
--- a/ocaml/xapi-guard/lib/varstored_interface.ml
+++ b/ocaml/xapi-guard/lib/varstored_interface.ml
@@ -99,6 +99,95 @@ let serve_forever_lwt_callback rpc_fn path _ req body =
       in
       Cohttp_lwt_unix.Server.respond_string ~status:`Method_not_allowed ~body ()
 
+(* The TPM has 3 kinds of states *)
+type state = {
+    permall: string  (** permanent storage *)
+  ; savestate: string  (** for ACPI S3 *)
+  ; volatilestate: string  (** for snapshot/migration/etc. *)
+}
+
+let split_char = ' '
+
+let join_string = String.make 1 split_char
+
+let deserialize t =
+  match String.split_on_char split_char t with
+  | [permall] ->
+      (* backwards compat with reading tpm2-00.permall *)
+      {permall; savestate= ""; volatilestate= ""}
+  | [permall; savestate; volatilestate] ->
+      {permall; savestate; volatilestate}
+  | splits ->
+      Fmt.failwith "Invalid state: too many splits %d" (List.length splits)
+
+let serialize t =
+  (* it is assumed that swtpm has already base64 encoded this *)
+  String.concat join_string [t.permall; t.savestate; t.volatilestate]
+
+let lookup_key key t =
+  match key with
+  | "/tpm2-00.permall" ->
+      t.permall
+  | "/tpm2-00.savestate" ->
+      t.savestate
+  | "/tpm2-00.volatilestate" ->
+      t.volatilestate
+  | s ->
+      Fmt.invalid_arg "Unknown TPM state key: %s" s
+
+let update_key key state t =
+  if String.contains state split_char then
+    Fmt.invalid_arg
+      "State to be stored (%d bytes) contained forbidden separator: %c"
+      (String.length state) split_char ;
+  match key with
+  | "/tpm2-00.permall" ->
+      {t with permall= state}
+  | "/tpm2-00.savestate" ->
+      {t with savestate= state}
+  | "/tpm2-00.volatilestate" ->
+      {t with volatilestate= state}
+  | s ->
+      Fmt.invalid_arg "Unknown TPM state key: %s" s
+
+let empty = ""
+
+let serve_forever_lwt_callback_vtpm ~cache mutex vtpm _path _ req body =
+  let uri = Cohttp.Request.uri req in
+  (* in case the connection is interrupted/etc. we may still have pending operations,
+     so use a per vTPM mutex to ensure we really only have 1 pending operation at a time for a vTPM
+  *)
+  Lwt_mutex.with_lock mutex @@ fun () ->
+  (* TODO: some logging *)
+  match (Cohttp.Request.meth req, Uri.path uri) with
+  | `GET, key when key <> "/" ->
+      let* contents = with_xapi ~cache @@ VTPM.get_contents ~self:vtpm in
+      let body = contents |> deserialize |> lookup_key key in
+      let headers =
+        Cohttp.Header.of_list [("Content-Type", "application/octet-stream")]
+      in
+      Cohttp_lwt_unix.Server.respond_string ~headers ~status:`OK ~body ()
+  | `PUT, key when key <> "/" ->
+      let* body = Cohttp_lwt.Body.to_string body in
+      let* contents = with_xapi ~cache @@ VTPM.get_contents ~self:vtpm in
+      let contents =
+        contents |> deserialize |> update_key key body |> serialize
+      in
+      let* () = with_xapi ~cache @@ VTPM.set_contents ~self:vtpm ~contents in
+      Cohttp_lwt_unix.Server.respond ~status:`No_content
+        ~body:Cohttp_lwt.Body.empty ()
+  | `DELETE, key when key <> "/" ->
+      let* contents = with_xapi ~cache @@ VTPM.get_contents ~self:vtpm in
+      let contents =
+        contents |> deserialize |> update_key key empty |> serialize
+      in
+      let* () = with_xapi ~cache @@ VTPM.set_contents ~self:vtpm ~contents in
+      Cohttp_lwt_unix.Server.respond ~status:`No_content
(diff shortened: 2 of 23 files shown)


ocaml/xapi-guard/lib/varstored_interface.ml
```diff
@@ -100,6 +75,30 @@ let serve_forever_lwt rpc_fn path =
   Lwt_switch.add_hook (Some shutdown) cleanup ;
   Lwt.return cleanup
 
+let serve_forever_lwt_callback rpc_fn path _ req body =
```
psafont, line 78: The path is only passed for debugging purposes, is it useful enough?

ocaml/xapi-guard/lib/varstored_interface.ml
```diff
@@ -99,53 +99,81 @@ let serve_forever_lwt_callback rpc_fn path _ req body =
...
-      in
-      let headers =
-        Cohttp.Header.of_list [("Content-Type", "application/json")]
-      in
-      Cohttp_lwt_unix.Server.respond_string ~headers ~status:`OK ~body ()
   | `GET, key -> (
-      let* value_opt = VB.get backend (VB.Key.of_string_exn key) in
-      match value_opt with
+      let* contents = with_xapi ~cache @@ VTPM.get_contents ~self:vtpm in
+      let map = deserialize contents in
```
psafont, line 148: I didn't expect the contents to be in json format, is there any explanation on the format of the values stored in the database?

edwintorok: SWTPM can store multiple key-value pairs for a single vtpm. I think in practice that key is only a single thing with a binary value, but I'll have to check.
We should indeed use something more efficient than json here.

ocaml/xapi-guard/lib/varstored_interface.ml
```diff
@@ -102,6 +75,107 @@ let serve_forever_lwt rpc_fn path =
...
+          (alist
+          |> List.to_seq
+          |> Seq.map @@ fun (k, v) ->
+             ( k |> Base64.decode_exn
+             , v
+               |> Yojson.Safe.Util.to_string
+               |> Base64.decode_exn
+             )
+          )
+    | _ ->
```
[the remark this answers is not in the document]
edwintorok: I'm dropping the base64 encoding/decoding from here.

edwintorok:
There are still some changes here that I'm not entirely happy with (disabling seccomp, using uid instead of gid for swtpm), but this is closer to being mergable.
Will do some tests on koji.

ocaml/xenopsd/xc/service.ml
```diff
@@ -803,23 +781,15 @@ module Swtpm = struct
...
-    let contents = suspend dbg ~xs ~domid ~vm_uuid ~vtpm_uuid in
-    let length = String.length contents in
-    if length > 0 then (
-      debug "Storing vTPM state of %d bytes" length ;
-      Varstore_privileged_client.Client.vtpm_set_contents dbg vtpm_uuid
-        (Base64.encode_string contents)
-    ) else
-      debug "vTPM state is empty: not storing" ;
-    (* needed to save contents before wiping the chroot *)
+    D.stop ~xs domid ;
```
psafont, line 792: This wasn't stopped before (!?)

edwintorok: Previously stop would call 'suspend' which would call D.stop.

ocaml/xapi-guard/lib/varstored_interface.ml
edwintorok: The debug output says which VM we are handling the request for, which could be quite useful for debugging, I'd keep it for now.
Maybe once we add distributed tracing here then we won't need this anymore.

psafont (approved):

ocaml/database/redo_log.ml
```diff
@@ -37,11 +37,12 @@ let get_static_device reason =
   R.debug "Found %d VDIs matching [%s]" (List.length vdis) reason ;
   match vdis with [] -> None | hd :: _ -> hd.Static_vdis_list.path
 
-(* Make sure we have plenty of room for the database *)
-let minimum_vdi_size =
+let mib megabytes =
```
lindig, line 40: The argument could be made an `int` to simplify the call site and converted to `int64` here.

ocaml/xapi-guard/lib/varstored_interface.ml
```diff
@@ -102,8 +75,111 @@ let serve_forever_lwt rpc_fn path =
...
+
+let deserialize t =
+  match String.split_on_char ' ' t with
+  | [permall] ->
+      (* backwards compat with reading tpm2-00.permall *)
+      {permall; savestate= ""; volatilestate= ""}
+  | [permall; savestate; volatilestate] ->
+      {permall; savestate; volatilestate}
+  | splits ->
+      Fmt.failwith "Invalid state: too many splits %d" (List.length splits)
```
lindig, line 117: You might want to include `t` in the message.

ocaml/xapi-guard/lib/varstored_interface.ml
```diff
@@ -102,8 +75,111 @@ let serve_forever_lwt rpc_fn path =
...
+
+(* The TPM has 3 kinds of states *)
+type state = {
+    permall: string  (** permanent storage *)
+  ; savestate: string  (** for ACPI S3 *)
+  ; volatilestate: string  (** for snapshot/migration/etc. *)
+}
+
+let deserialize t =
+  match String.split_on_char ' ' t with
```
lindig, line 110: The split character is shared between two functions. I would define a value for it.

ocaml/xapi-guard/lib/varstored_interface.ml
```diff
@@ -102,8 +75,111 @@ let serve_forever_lwt rpc_fn path =
...
+      let contents =
+        contents |> deserialize |> update_key key body |> serialize
+      in
+      let* () = with_xapi ~cache @@ VTPM.set_contents ~self:vtpm ~contents in
+      Cohttp_lwt_unix.Server.respond ~status:`No_content
+        ~body:Cohttp_lwt.Body.empty ()
+  | `DELETE, key when key <> "/" ->
+      let* contents = with_xapi ~cache @@ VTPM.get_contents ~self:vtpm in
+      let contents =
+        contents |> deserialize |> update_key key "" |> serialize
```
lindig, line 172: The empty string here is important. So we set the value to the empty string. Maybe bind it `let empty = "" in`?

ocaml/xapi-guard/src/main.ml
```diff
@@ -66,7 +71,9 @@ let recover_path = "/run/nonpersistent/varstored-guard-active.json"
 
 let store_args sockets =
   Hashtbl.fold
-    (fun path (_, (vm_uuid, gid)) acc -> {Persistent.vm_uuid; path; gid} :: acc)
+    (fun path (_, (vm_uuid, gid, typ)) acc ->
```
lindig, line 74: Could also use `to_seq_values |> Seq.to_list`

ocaml/xenopsd/xc/service.ml
```diff
@@ -668,56 +668,90 @@ module Swtpm = struct
...
     in
-    let args = Fe_argv.Add.many [string_of_int domid; tpm_root; state_uri] in
-    let args = Fe_argv.run args |> snd |> Fe_argv.argv in
+    (* fetch it here instead of swtpm-wrapper: better error reporting,
+       swtpm-wrapper runs as a service and getting the exact error back is
+       difficult. *)
+    let needs_init = check_state_needs_init task vtpm_uuid in
     let timeout_seconds = !Xenopsd.swtpm_ready_timeout in
+    if needs_init then (
+      debug "vTPM is empty, needs to be created" ;
```
lindig, line 739: Should this include a path?

scripts/vtpm_smoke_test.sh
```diff
@@ -0,0 +1,147 @@
+#!/bin/sh
```
lindig, line 1: This should be `bash`

lindig (approved):

edwintorok:
To consider: what happens to existing VMs that use the old storage method. The storage is meant to be backwards compatible (same base64 encoding, etc.), but I removed all the 'read state from filesystem and store back into xapi' code from shutdown/migrate, which means that a running VM with a vTPM will lose some data if it has changed it since VM boot.

We could:
* add back the code that uploads the vTPM state in this case but make it conditional (if file is missing do nothing). This is problematic because if the new code "forgets" a tpm state file then that will overwrite a potentially newer state in the database
* do the vTPM state upload as a one-off during XAPI restart/upgrade (probably preferred).
* introduce a new enum and keep both old and new code, however this multiplies codepaths for testing and inevitably something will go wrong with one of the code paths

Implementing 2) seems like the best way forward.

edwintorok:
Back to draft until upgrade issues are resolved.

edwintorok:
New enum would work better, because we don't have easy in-place migration of VM emulators without migrating the entire VM (for which we may have not enough memory)

lindig:
Where is the state stored that is not in the xapi database that we would have to upload?

psafont:
> Where is the state stored that is not in the xapi database that we would have to upload?

In the sandbox prepared for the domain's SWTPM, exposed as a file

ocaml/xapi-guard/lib/varstored_interface.ml
edwintorok: 't' might be secret though (the entire vTPM state)

ocaml/xapi-guard/src/main.ml
```diff
@@ -141,6 +155,47 @@ let depriv_destroy dbg gid path =
...
+        let+ () = safe_unlink path in
+        Hashtbl.remove sockets path
+      in
+      let* () = Lwt.finalize stop_server finally in
+      D.debug "[%s] stopped swtpm server for gid %d and removed socket" dbg gid ;
+      Lwt.return_unit
+  | Some _ ->
+      D.warn
+        "[%s] asked to stop swtpm server for path %s, but it's not an\n\
+        \      swtpm server" dbg path ;
```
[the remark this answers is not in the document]
edwintorok: It also shouldn't include a \n. I'll fix.

ocaml/xenopsd/xc/service.ml
edwintorok: no path anymore, but I will include the vtpm UUID.

ocaml/xapi-guard/src/main.ml
edwintorok: Needs the key too, but yes I've rewritten this using seq instead of fold, it is clearer.

ocaml/xapi-guard/lib/varstored_interface.ml
edwintorok: I've added an additional check during serialization that it doesn't contain the forbidden separator char (it is base64 encoded, so it shouldn't). This will then reject malformed state early (at serialization time before the state gets corrupted), instead of deserialization time. 
This is part of this fixup: 0dc82ffbbdb1bfe9a27fa72b2588543c5b571f3d

lindig: Good point about not logging secrets; I was looking for something to convey the shape of the string for debugging. It's good enough as it is.

PR #5006, opened 2023-05-18
Observability Extend unit tests

jameshensmancitrix:
Adding in new unit tests to validate operations on the API


ocaml/tests/test_observer.ml
```diff
@@ -187,10 +483,19 @@ let test_attribute_validation () =
...
   [
-    ("test_hashtbl_leaks", `Quick, test_hashtbl_leaks)
-  ; ("test_tracing_exn_backtraces", `Quick, test_tracing_exn_backtraces)
+    ( "test_observer_create_and_destroy"
+    , `Quick
+    , test_observer_create_and_destroy
+    )
+  ; ("test_observer_valid_params", `Quick, test_observer_valid_params)
+  ; ("test_observer_endpoint", `Quick, test_observer_endpoint)
+  ; ("test_file_export", `Quick, test_file_export_writes)
```
jameshensmancitrix, line 492: This will fail first time as the directory would not have been created as the export loop would not have taken place. This test also ends up leaving a next of empty directories in my repo. I don't know too much about xenRT testing this test feels better here as it is quite stateful.

ocaml/libs/tracing/tracing.ml
```diff
@@ -620,6 +625,15 @@ module Export = struct
           span_list
       with e -> debug "Tracing: ERROR %s" (Printexc.to_string e)
 
+    let flush_spans () =
+      let span_list = Spans.since () in
+      get_tracer_providers ()
+      |> List.iter (fun x ->
+             if x.TracerProvider.enabled then
```
lindig, line 632: This could be expressed with `List.filter`. I these lists are long, using sequences would be better (module `Seq`).

ocaml/libs/tracing/tracing.mli
```diff
@@ -76,10 +78,20 @@ module Tracer : sig
...
   val span_hashtbl_is_empty : unit -> bool
+
+  val finished_span_hashtbl_is_empty : unit -> bool
 end
 
 module TracerProvider : sig
-  type t
+  type t = {
+      name_label: string
+    ; tags: (string * string) list
```
lindig, line 88: If the list is long and needs to be accessed, use a more efficient data structure.

ocaml/libs/tracing/tracing.mli
```diff
@@ -76,10 +78,20 @@ module Tracer : sig
...
+    ; endpoints: endpoint list
+    ; filters: string list
+    ; processors: string list
+    ; enabled: bool
+    ; service_name: string
```
lindig, line 93: This needs documentation. What is the difference to `name_label`?

ocaml/tests/test_observer.ml
```diff
@@ -13,27 +13,114 @@
...
-let _ =
-  create ~enabled:false ~name_label:"default" ~uuid:"123" ~tags:[]
-    ~endpoints:["bugtool"] ~processors:[] ~filters:[] ~service_name:"xapi" ;
-  main ()
+let test_trace_log_dir = "/tmp/var/log/dt/zipkinv2/json/"
+
+let () = Export.Destination.File.set_trace_log_dir test_trace_log_dir
+
+module Xapi_DB = struct
+  let assert_x_observers ~__context x =
```
lindig, line 27: What is an `x_observer`?

ocaml/tests/test_observer.ml
```diff
@@ -13,27 +13,114 @@
...
+      x (List.length observers)
+
+  let assert_observer_disabled ~__context ~self =
+    let enabled = Db.Observer.get_enabled ~__context ~self in
+    Alcotest.(check bool) "Observer disabled" false enabled
+
+  let check_endpoints ~__context ~self ~endpoints =
+    let db_endpoints = Db.Observer.get_endpoints ~__context ~self in
+    let _ =
+      List.map2
```
lindig, line 40: This looks wrong? Why use `map` and not `iter` when the result is not used?

ocaml/tests/test_observer.ml
```diff
@@ -13,27 +13,114 @@
...
+  let assert_x_observers ~__context x =
+    let providers = Tracing.get_tracer_providers () in
+    Alcotest.(check int)
+      (Printf.sprintf "%d provider(s) exists in lib " x)
+      x (List.length providers)
+
+  let find_provider_exn ~name =
+    let providers = Tracing.get_tracer_providers () in
+    match
+      List.find_opt (fun x -> x.TracerProvider.name_label = name) providers
```
lindig, line 57: Do we need this often? That would suggest a more efficient data structure.

ocaml/tests/test_observer.ml
```diff
@@ -13,27 +13,114 @@
...
+      |> List.map (fun endpoint ->
+             match endpoint with
+             | Bugtool ->
+                 "bugtool"
+             | Url x ->
+                 Uri.to_string x
+         )
+    in
+    let _ =
+      List.map2
```
lindig, line 94: Same as above - why not using `iter`?

ocaml/tests/test_observer.ml
```diff
@@ -55,35 +142,211 @@ let start_test_trace () =
...
+  Xapi_observer.set_endpoints ~__context ~self ~value:endpoints ;
+  Xapi_DB.check_endpoints ~__context ~self ~endpoints ;
+  TracerProvider.check_endpoints ~name:"test-observer" ~endpoints ;
+  test_destroy ~__context ~self ()
+
+let is_export_dir_empty () =
+  try
+    let log_dir =
+      Sys.readdir test_trace_log_dir
+      |> Array.map (fun x -> test_trace_log_dir ^ x)
```
lindig, line 247: Use `Filename.concat`? Is this constructing file names?

ocaml/libs/tracing/tracing.mli
jameshensmancitrix: service_name refers  to a component name i.e. xapi and is responsible for the difference in bar colours seen in the UI. name_label is a user defined label that is used to identify a provider

ocaml/tests/test_observer.ml
jameshensmancitrix: Resolved

ocaml/tests/test_observer.ml
YanRachel: changed to assert_num_observers if that's more straighforward?

ocaml/libs/tracing/tracing.mli
YanRachel: we're not expecting the tags list to be long

ocaml/tests/test_observer.ml
```diff
@@ -262,32 +265,28 @@ let clear_export_dir () =
...
+      ; ("localEndpoint", `Assoc [("serviceName", `String "xapi")])
+      ; ("duration", `Int _)
+      ; ("timestamp", `Int _)
+      ; ("name", `String "test_task")
+      ; ("traceId", `String _)
+      ; ("id", `String _)
+      ] ->
+      ()
+  | _ ->
+      Alcotest.failf "Trace log json format incorrect"
```
lindig, line 289: If this ever fails you will want to see the actual JSON. Would turn the JSON into a string and include it here.

lindig (approved):

mg12 (approved):

ocaml/libs/tracing/tracing.ml
```diff
@@ -620,6 +631,15 @@ module Export = struct
           span_list
       with e -> debug "Tracing: ERROR %s" (Printexc.to_string e)
 
+    let flush_spans () =
+      let span_list = Spans.since () in
+      get_tracer_providers ()
+      |> List.filter (fun x -> x.TracerProvider.enabled)
+      |> List.iter (fun x ->
```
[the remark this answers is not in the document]
psafont: To avoid the nested iter, concat_map can be used:
```ocaml
get_tracer_providers ()
|> List.filter (fun x -> x.TracerProvider.enabled)
|> List.concat_map (fun x -> Traceprovider.endpooints.of x)
|> List.iter (export_to_endpoint span_list)
```

ocaml/tests/test_observer.ml
```diff
@@ -13,27 +13,111 @@
...
 
 open D
 
 let () = Printexc.record_backtrace true
 
-let _ =
-  create ~enabled:false ~name_label:"default" ~uuid:"123" ~tags:[]
-    ~endpoints:["bugtool"] ~processors:[] ~filters:[] ~service_name:"xapi" ;
-  main ()
+let test_trace_log_dir = "/tmp/var/log/dt/zipkinv2/json/"
```
psafont, line 22: Please use `Filename.get_temp_dir_name` instead of harcoding all the path

ocaml/tests/test_observer.ml
```diff
@@ -55,35 +139,209 @@ let start_test_trace () =
...
+      |> Array.map (Filename.concat test_trace_log_dir)
+    in
+    Array.length log_dir = 0
+  with e ->
+    debug "%s" (Printexc.to_string e) ;
+    true
+
+let clear_export_dir () =
+  try
+    let rec rmdir_rec path =
```
psafont, line 253: Generating a temporary folder for each test is better than reusing a single one for all of them, it may allow to run them in parallel and avoid these cleanup functions

lindig (approved):

ocaml/tests/test_observer.ml
```diff
@@ -55,35 +144,218 @@ let start_test_trace () =
...
+  Alcotest.check_raises
+    "Xapi_observer.set_components should fail on invalid component"
+    Api_errors.(Server_error (invalid_value, ["component"; invalid_component]))
+    (fun () ->
+      Xapi_observer.set_components ~__context ~self ~value:[invalid_component]
+      |> ignore
+    )
+
+let test_endpoints ~__context ~self =
+  let valid_endpoints = ["bugtool"; "http://example.com:9411/api/v2/spans"] in
```
psafont, line 212: It would be interesting to add at least an invalid endpint, it could be used to accumulate regression tests for invalid endpoints

ocaml/tests/test_observer.ml
```diff
@@ -55,35 +144,218 @@ let start_test_trace () =
...
+    let log_dir =
+      Sys.readdir test_trace_log_dir
+      |> Array.map (Filename.concat test_trace_log_dir)
+    in
+    Array.length log_dir = 0
+  with e ->
+    debug "%s" (Printexc.to_string e) ;
+    true
+
+let rec rmdir_rec path =
```
psafont, line 256: I'm surprised the function hasn't been added yet to Unixext

psafont (approved):
Aside from the small comment about the endpoints, this looks good

ocaml/tests/test_observer.ml
YanRachel: because currently the only endpoint validity check we do is Uri.to_url, I believe it accepts _any_ string so the test would always fail -Can always add an invalid_endpoint check when/if we change this

psafont: The validation happens in ocaml/xapi/xapi_observer.ml -> validate_endpoint. Uri.of_string is used to validate the string, on top of that the scheme is enforced and the host of the uri is also validated to be a valid hostname or an IP. I think there can be lots of invalid strings, and even invalid URIs as well :)

YanRachel: oh whoops! that's new, I thought we were adding proper validation later -my bad

PR #5007, opened 2023-05-18
Instrumentation for Xenopsd and Xapi Storage

jameshensmancitrix:
This includes the instrumentation and management of Observability in Xenopsd. It introduces the system to manage and broadcast to different components (this being xapi and xenopsd)

(Minor issues remain so i will continue to iron these out)


robhoes:
Have the minor issues been ironed out now?

jameshensmancitrix:
Yes, Regarding the last commit, there are other changes we have discussed that would contain minor improvements, would it be etter to move this commit to a new branch or keep it here?

ocaml/xapi/storage_mux.ml
```diff
@@ -381,6 +401,8 @@ module Mux = struct
         |> String.concat "; "
         |> Printf.sprintf "[%s]"
         ) ;
+      with_tracing ~name:"Storage_migrate.update_snapshot_info_src" ~dbg
```
robhoes, line 404: The name should be `SR.update_snapshot_info_src`.

ocaml/xapi/storage_mux.ml
```diff
@@ -18,6 +18,19 @@ module D = Debug.Make (struct let name = "mux" end)
 
 open D
 
+let with_tracing ~name ~dbg f =
+  let dbg, parent = Context.tracing_of_dbg dbg in
+  let tracer = Tracing.get_tracer ~name in
```
robhoes, line 23: `name` here is the name of the operation; don't we need the name of the tracer here?

jameshensmancitrix: the `Tracer` name is not the `TracerProvider` name, its somewhat redundent currently but a tracer having a name is in the spec.

ocaml/xapi/storage_mux.ml
```diff
@@ -18,6 +18,19 @@ module D = Debug.Make (struct let name = "mux" end)
...
+  let span = Tracing.Tracer.start ~tracer ~name ~parent () in
```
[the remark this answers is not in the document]
robhoes: Yes, but I want to be able to distinguish easily between high-level XenAPI calls and calls into xapi's the storage API (SMAPIv2). Actually, `SMAPIv2.` would be a good prefix.

jameshensmancitrix:
I have added duel signoff to those commit that were mostly your work originially.

ocaml/xapi/xapi_observer.ml
```diff
@@ -11,11 +11,117 @@
...
+end
+
+let get_forwarder c =
+  let module Forwarder = ( val match c with
+                               | "xapi" ->
+                                   (module Observer)
+                               | "xenopsd" ->
+                                   (module Xapi_xenops.Observer)
+                               | _ ->
+                                   failwith "Not a valid component"
```
lindig, line 109: This should include the matched string

ocaml/xapi/xapi_observer.ml
```diff
@@ -96,24 +210,67 @@ let create ~__context ~name_label ~name_description ~hosts ~attributes
...
+  Db.Observer.get_components ~__context ~self
+  |> observed_components_of
+  |> unregister_components ~__context ~self
 
 let destroy ~__context ~self = Db.Observer.destroy ~__context ~self
 
-let set_trace_log_dir dir =
-  Tracing.Export.Destination.File.set_trace_log_dir dir
+let set_trace_log_dir ~__context dir =
+  List.iter
```
lindig, line 228: This pattern is repeated several times. Could we create the list of modules once? I am also a bit dubious about using first-class modules in so many places.

jameshensmancitrix: I don't see a way to parameterise this more effectively.

ocaml/xapi/xapi_observer.ml
jameshensmancitrix: resolved

lindig (approved):

PR #5008, opened 2023-05-19
Make NBD disconnect robust to the device being gone

MarkSymsCtx:
Noticed this in a test which had failed for other reasons and then wouldn't unplug the datapath. If the device is not found there is nothing to disconnect.

### DIFF
--- a/scripts/nbd_client_manager.py
+++ b/scripts/nbd_client_manager.py
@@ -208,11 +208,16 @@ def disconnect_nbd_device(nbd_device):
     This function is idempotent: calling it on an already disconnected device
     does nothing.
     """
-    if _is_nbd_device_connected(nbd_device=nbd_device):
-        _remove_persistent_connect_info(nbd_device)
-        cmd = ['nbd-client', '-disconnect', nbd_device]
-        _call(cmd)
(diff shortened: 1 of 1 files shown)


scripts/nbd_client_manager.py
```diff
@@ -208,11 +208,16 @@ def disconnect_nbd_device(nbd_device):
     This function is idempotent: calling it on an already disconnected device
     does nothing.
     """
-    if _is_nbd_device_connected(nbd_device=nbd_device):
-        _remove_persistent_connect_info(nbd_device)
-        cmd = ['nbd-client', '-disconnect', nbd_device]
-        _call(cmd)
-        _wait_for_nbd_device(nbd_device=nbd_device, connected=False)
+    try:
+        if _is_nbd_device_connected(nbd_device=nbd_device):
```
robhoes, line 212: So this check doesn't always give the right answer?

robhoes (approved):

scripts/nbd_client_manager.py
MarkSymsCtx: If the device isn't present at all it fails with an exception `NbdDeviceNotFound` so there are three states

- Connected
- Disconnected
- Not present

psafont (approved):

PR #5011, opened 2023-05-19
Manage and Collect spans in logs

jameshensmancitrix:
Collect spans in Bugtool and delete old spans or purge spans when above 100Mb

### DIFF
--- a/scripts/xapi-tracing-log-trim.cron
+++ b/scripts/xapi-tracing-log-trim.cron
@@ -1,0 +1,3 @@
+# Run at 20 minutes past the hour,
+# every hour, day-of-month, month, day-of-week
+20   *  *  *  * root /opt/xensource/libexec/xapi-tracing-log-trim.sh
--- a/scripts/xapi-tracing-log-trim.sh
+++ b/scripts/xapi-tracing-log-trim.sh
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+# find the oldest trace files beyond MAX_MB size and delete
+find /var/log/dt -mindepth 1 -type f -printf '%T@ %k %p\n' | sort -nr | awk -v max_mb=${MAX_MB:=100} 'BEGIN{size=0}{size+=$2; if (size>1000*max_mb) {$1=""; $2=""; print $0}}' | xargs -r rm
+
+# delete any remaining files that are older than MAX_DAYS days
+find /var/log/dt -mindepth 1 -mtime ${MAX_DAYS:=+30} -delete
+
+# clean up any empty directories
+find /var/log/dt -mindepth 1 -type d -empty -delete
--- a/scripts/Makefile
+++ b/scripts/Makefile
@@ -44,6 +44,7 @@ install:
 	$(IDATA) audit-logrotate $(DESTDIR)/etc/logrotate.d/audit
 	$(IDATA) xapi-logrotate.conf $(DESTDIR)$(ETCXENDIR)
 	$(IPROG) xapi-logrotate.sh $(DESTDIR)$(LIBEXECDIR)
+	$(IPROG) xapi-tracing-log-trim.sh $(DESTDIR)$(LIBEXECDIR)
 	$(IPROG) xapi-wait-init-complete $(DESTDIR)$(OPTDIR)/bin
 	$(IPROG) xapi-autostart-vms $(DESTDIR)$(OPTDIR)/bin
 	$(IPROG) udhcpd.skel $(DESTDIR)$(ETCXENDIR)/udhcpd.skel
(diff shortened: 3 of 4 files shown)


scripts/xapi-tracing-log-trim.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+export MAX_MB=100
+
+# find the oldest trace files beyond MAX_MB size and delete
+find /var/log/dt -type f -printf "%T@ %s %p\n" | sort -nr | awk 'BEGIN{size=0}{size+=$2; if (size>1024*1024*MAX_MB) {$1=""; $2=""; print $0}}' | xargs rm
+
+# delete any remaining files that are older than 30 days
+find /var/log/dt -mtime +30 -delete
```
psafont, line 9: Would it make sense to delete more traces if the total amount of traces is bigger than a set amount?

scripts/xapi-tracing-log-trim.cron
```diff
@@ -0,0 +1,3 @@
+# Run at 20 minutes past the hour,
```
edwintorok, line 1: Could you use logrotate here? http://www.skrenta.com/rt/man/logrotate.8.html

That is already set up to run on a schedule and can rotate/compress based on file size.
Although that may only work well if a single well known filename is used, you can use '*' in logrotate, but I'm not sure whether that would actually delete old files completely or you'd have to do it by hand.

jameshensmancitrix: I could look into doing that

edwintorok: Another option is https://www.freedesktop.org/software/systemd/man/tmpfiles.d.html which can clean up files in a directory based on age.

scripts/xapi-tracing-log-trim.sh
jameshensmancitrix: traces can also vary on size, the entire reason to delete them is to avoid too much space being used up. Using size concretely caps how much can be stored whereas with a set amount of traces that wouldn't be as clear.

psafont: Sorry if it wasn't clear, I meant a set amount in terms of disk space, not number of files

jameshensmancitrix: The set amount is 100mb

scripts/xapi-tracing-log-trim.cron
edwintorok: I think tmpfiles.d would. For logrotate you have to use wildcards, but you'll have to do some tests to see how it works.

lindig: Rotating logfiles has the problem that you might want to rotate a file that is currently open - assuming the logging mechanism keeps the file open between writes. How is this handled? It could be an option to log through the syslog service to avoid this problem rather than writing to a regular file.

scripts/xapi-tracing-log-trim.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+export MAX_MB=100
+
+# find the oldest trace files beyond MAX_MB size and delete
+find /var/log/dt -type f -printf "%T@ %s %p\n" | sort -nr | awk 'BEGIN{size=0}{size+=$2; if (size>1024*1024*MAX_MB) {$1=""; $2=""; print $0}}' | xargs rm
+
+# delete any remaining files that are older than 30 days
```
edwintorok, line 8: would be good to parametrize the script (with a default of 30 days), so we can test in XenRT without having to wait for 30days...

scripts/xapi-tracing-log-trim.cron
edwintorok: for now maybe keep this script if neither tmpfiles.d or syslog is appropriate. Syslog would have all kinds of limit on line length and rate limiting, might be more difficult to use.

edwintorok: (also if we want to switch to a binary format later syslog wouldn't work). To change to something else signals could be used if needed (or the application can be mindful of the size of the file it has already written and reopen a new file on its own)

mg12: After the latest changes, I can test the script successfully in dom0 using eg.
```
MAX_MB=10 /opt/xensource/libexec/xapi-tracing-log-trim.sh  #remove the oldest files beyond 10MB total in dir
MAX_DAYS=0 /opt/xensource/libexec/xapi-tracing-log-trim.sh #remove all the files immediately
```

scripts/xapi-tracing-log-trim.sh
mg12: After the latest changes, I can test the script successfully in dom0 using eg.
```
MAX_MB=10 /opt/xensource/libexec/xapi-tracing-log-trim.sh  #remove the oldest files beyond 10MB total in dir
MAX_DAYS=0 /opt/xensource/libexec/xapi-tracing-log-trim.sh #remove all the files immediately
```

edwintorok (approved):

mg12 (approved):

psafont (approved):

PR #5015, opened 2023-05-22
Set PIF's IPv6 Gateway when in DHCP/Autconf

benjamreis:
Read the default route for IPv6 and set it in the PIF

### DIFF
--- a/ocaml/networkd/bin/network_server.ml
+++ b/ocaml/networkd/bin/network_server.ml
@@ -445,6 +445,24 @@ module Interface = struct
       )
       ()
 
+  let get_ipv6_gateway dbg name =
+    Debug.with_thread_associated dbg
+      (fun () ->
+        let output = Ip.route_show ~version:Ip.V6 name in
+        try
+          let line =
+            List.find
+              (fun s -> Astring.String.is_prefix ~affix:"default via" s)
+              (Astring.String.cuts ~empty:false ~sep:"\n" output)
+          in
+          let addr =
+            List.nth (Astring.String.cuts ~empty:false ~sep:" " line) 2
+          in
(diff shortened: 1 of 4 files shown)


ocaml/xapi/helpers.ml
```diff
@@ -331,13 +331,20 @@ let update_getty () =
...
+              ~value:(Unix.string_of_inet_addr addr)
+        | None ->
+            ()
+    ) ;
+    match Net.Interface.get_ipv6_gateway dbg bridge with
+    | Some addr ->
+        Db.PIF.set_ipv6_gateway ~__context ~self:pif
+          ~value:(Unix.string_of_inet_addr addr)
+    | None ->
+        ()
```
psafont, line 347: Shouldn't this be inside the `if ...` block as well?

benjamreis: Yes it should

psafont (approved):

benjamreis:
Validated with IPv6 DHCP & Autoconf (Static was already working and still is :+1: )

lindig (approved):

ocaml/networkd/bin/network_server.ml
```diff
@@ -445,6 +445,24 @@ module Interface = struct
       )
       ()
 
+  let get_ipv6_gateway dbg name =
+    Debug.with_thread_associated dbg
+      (fun () ->
+        let output = Ip.route_show ~version:Ip.V6 name in
+        try
+          let line =
```
lindig, line 453: I think this kind of problem lends itself to this style of coding:
```
output 
|> Astring.String.cuts ~empty:false ~sep:"\n"
|> List.find (fun s -> Astring.String.is_prefix ~affix:"default via" s)
```

PR #5017, opened 2023-05-22
CA-377456 unblock cross-pool migration with VTPM when halted

lindig:
The VTPM of a halted VM is up to date and its content is transferred in the migration stream. Unblock this case.

### DIFF
--- a/ocaml/xapi/xapi_vm_migrate.ml
+++ b/ocaml/xapi/xapi_vm_migrate.ml
@@ -1756,15 +1756,16 @@ let assert_can_migrate ~__context ~vm ~dest ~live:_ ~vdi_map ~vif_map ~options
              , [Ref.string_of vm; Ref.string_of remote.dest_host]
              )
           ) ;
+      let power_state = Db.VM.get_power_state ~__context ~self:vm in
       (* VTPMs can't be exported currently, which will make the migration fail *)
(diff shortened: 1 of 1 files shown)


psafont (approved):

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -1756,15 +1756,16 @@ let assert_can_migrate ~__context ~vm ~dest ~live:_ ~vdi_map ~vif_map ~options
              , [Ref.string_of vm; Ref.string_of remote.dest_host]
              )
           ) ;
+      let power_state = Db.VM.get_power_state ~__context ~self:vm in
       (* VTPMs can't be exported currently, which will make the migration fail *)
-      ( if Db.VM.get_VTPMs ~__context ~self:vm <> [] then
+      ( if power_state <> `Halted && Db.VM.get_VTPMs ~__context ~self:vm <> []
+      then
           let message = "Cross-pool VM migration with VTPMs attached" in
```
edwintorok, line 1763: I think we can drop this check completely now that VM metadata export with vTPM is implemented.
And intra-pool live migration has worked for a long time.

We should test and see whether it works or not.

lindig: To help progress this, maybe merge and we investigate further. This came up in a test.

edwintorok (approved):

PR #5020, opened 2023-05-25
CP-42739: Bump Java SDK to JDK 11 (LTS)

danilo-delbusso:
Also update references to xapi project in the `pom.xml`


kc284 (approved):

psafont (approved):

danilo-delbusso:
> Is there some OIL we can do here to generate the SDK and include it in the gihub releases?
@psafont not sure if there's an OIL but we've had CP-41012 for a while. Just haven't gotten round to it

PR #5022, opened 2023-05-25
General improvements for Tracing

jameshensmancitrix:
General improvements to the library and instrumentation to improve performance, functionality and closeness to OTEL specification

1. Add error attribute to identify to UIs like jaeger that a span has an error contained
2. Ensure all uri endpoints passed to the API have a valid scheme and host
3. The `make_remote_rpc` in helpers.ml is given `__context` to it can send tracing information
4. Remove unneeded fields from tracer providers
5. Move `service_name` from `TracerProvider` into a library wide constant for a component
6.  Add SpanLink
7.  Add SpanEvent
8. Use w3c Format for sending tracing to Xenopsd

Commit messages need to be cleaned

### DIFF
--- a/ocaml/libs/tracing/tracing.ml
+++ b/ocaml/libs/tracing/tracing.ml
@@ -78,7 +78,7 @@ module Span = struct
     ; service_name: string
     ; begin_time: float
     ; end_time: float option
-    ; tags: (string * string) list
+    ; attributes: (string * string) list
   }
   [@@deriving rpcty]
 
@@ -93,7 +93,7 @@ module Span = struct
 
   let generate_id n = String.init n (fun _ -> "0123456789abcdef".[Random.int 16])
 
-  let start ?(tags = []) ~name ~parent ~span_kind ~service_name () =
+  let start ?(attributes = []) ~name ~parent ~span_kind ~service_name () =
     let trace_id =
       match parent with
       | None ->
@@ -116,13 +116,17 @@ module Span = struct
     ; service_name
     ; begin_time
     ; end_time
-    ; tags
+    ; attributes
     }
 
-  let get_tag t tag = snd (List.find (fun s -> fst s = tag) t.tags)
+  let get_tag t tag = snd (List.find (fun s -> fst s = tag) t.attributes)
 
-  let finish ?(tags = []) ~span () =
-    {span with end_time= Some (Unix.gettimeofday ()); tags= span.tags @ tags}
+  let finish ?(attributes = []) ~span () =
+    {
+      span with
+      end_time= Some (Unix.gettimeofday ())
+    ; attributes= span.attributes @ attributes
+    }
 
   let set_span_kind span span_kind = {span with span_kind}
 
@@ -138,7 +142,7 @@ module Span = struct
             )
         in
         let status_code = Status.Error in
-        let exn_tags =
+        let exn_attributes =
           [
             ("exception.message", msg)
           ; ("exception.stacktrace", stacktrace)
@@ -151,7 +155,7 @@ module Span = struct
             {
               span with
               status= {status_code; description}
-            ; tags= span.tags @ exn_tags
+            ; attributes= span.attributes @ exn_attributes
             }
         | _ ->
             span
@@ -280,7 +284,7 @@ module Spans = struct
                         span.Span.context.span_id ;
                       let span =
                         Span.finish ~span
-                          ~tags:
+                          ~attributes:
                             [
                               ( "gc_inactive_span_timeout"
                               , string_of_float elapsed
@@ -319,10 +323,8 @@ end
 module TracerProvider = struct
   type t = {
       name_label: string
-    ; tags: (string * string) list
+    ; attributes: (string * string) list
     ; endpoints: endpoint list
-    ; filters: string list
-    ; processors: string list
     ; enabled: bool
     ; service_name: string
   }
@@ -339,10 +341,8 @@ module Tracer = struct
     let provider : TracerProvider.t =
       {
         name_label= ""
-      ; tags= []
+      ; attributes= []
       ; endpoints= []
-      ; filters= []
-      ; processors= []
       ; enabled= false
       ; service_name= ""
       }
@@ -359,7 +359,7 @@ module Tracer = struct
     ; span_kind= SpanKind.Client (* This will be the span of the client call*)
     ; begin_time= Unix.gettimeofday ()
     ; end_time= None
-    ; tags= []
+    ; attributes= []
     }
 
   let start ~tracer:t ?(span_kind = SpanKind.Internal) ~name ~parent () :
@@ -368,9 +368,9 @@ module Tracer = struct
     if not t.provider.enabled then
       ok_none
     else
-      let tags = t.provider.tags in
+      let attributes = t.provider.attributes in
       let span =
-        Span.start ~tags ~name ~parent ~span_kind
+        Span.start ~attributes ~name ~parent ~span_kind
           ~service_name:t.provider.service_name ()
       in
       Spans.add_to_spans ~span ; Ok (Some span)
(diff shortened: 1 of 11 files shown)


ocaml/libs/tracing/tracing.ml
```diff
@@ -466,6 +444,14 @@ module Export = struct
 
   let set_export_interval t = export_interval := t
 
+  let host_id = ref "localhost"
+
+  let set_host_id id = host_id := id
+
+  let service_name = ref "unknown"
+
+  let set_service_name name = service_name := name
```
psafont, line 480: I would think having unkowns is a rather unwanted effect, would it be worth using a function to log when an unknown is generated?

```suggestion
  let service_name = ref None
  let set_service_name name = service_name := Some name
  let get_service_name () =
    match !sercive_name with
    | None ->
        warn "service name not yet set!" ;
        "unknown"
    | Some name ->
        name
```

ocaml/xapi-idl/xen/xenops_interface.ml
```diff
@@ -1160,6 +1160,10 @@ module XenopsAPI (R : RPC) = struct
       declare "Observer.set_host_id" []
         (debug_info_p @-> string_p @-> returning unit_p err)
 
+    let set_service_name =
+      declare "Observer.set_service_name" []
+        (debug_info_p @-> returning unit_p err)
```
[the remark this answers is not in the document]
jameshensmancitrix: In this unlike most others, it doesn't depend on xapi, also `init` could be made like this. Where is xenops startup sequence?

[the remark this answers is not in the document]
psafont: it's in `ocaml/xenopsd/xc/xenops_xc_main.ml`

robhoes (approved):
Generally looks good to me.

ocaml/xenopsd/lib/xenopsd.ml
```diff
@@ -473,6 +473,7 @@ let main backend =
     )
     () ;
   Xenops_server.WorkerPool.start !worker_pool_size ;
+  Tracing.Export.set_service_name "xenopsd" ;
```
psafont, line 476: This seems like it should be set up earlier, before xenops starts listening, maybe at around line 433?

ocaml/xapi/xapi_observer.ml
```diff
@@ -152,11 +152,18 @@ let assert_valid_endpoints endpoints =
     | "bugtool" ->
         true
     | url -> (
-      try
-        let _ = Uri.of_string url in
-        true
-      with _ -> false
-    )
+        let uri = Uri.of_string url in
```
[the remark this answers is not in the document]
jameshensmancitrix: i can add that back in

ocaml/libs/tracing/tracing.ml
```diff
@@ -138,19 +163,20 @@ module Span = struct
...
+          ; ("error", "true")
           ]
         in
         match span.status.status_code with
         | Unset ->
             {
               span with
               status= {status_code; description}
-            ; tags= span.tags @ exn_tags
+            ; attributes= span.attributes @ exn_attributes
```
lindig, line 179: Could we end up with duplicate keys? I assume these lists are short.

lindig (approved):

ocaml/libs/tracing/tracing.ml
```diff
@@ -54,6 +54,11 @@ module Status = struct
   [@@deriving rpcty]
 end
 
+module SpanEvent = struct
+  type t = {name: string; time: float; attributes: (string * string) list}
```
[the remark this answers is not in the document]
jameshensmancitrix: Following how it is done in `Span`

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: attributes also use namespaces. Default attributes begin with `xs.`, exn with `exception.` and user with `user.`. There is one which does nto have a namespace `error` and this is to work with Jaeger's internal system to detect if a span had an error.

ocaml/libs/tracing/tracing.ml
```diff
@@ -132,14 +134,15 @@ module Span = struct
...
-    {
-      span with
-      end_time= Some (Unix.gettimeofday ())
-    ; attributes= span.attributes @ attributes
-    }
+    let attributes =
+      List.fold_left
+        (fun attrs (k, v) -> Attributes.add k v attrs)
+        span.attributes attributes
+    in
```
psafont, line 144: I'd rather use union here to be explicit on which map do we prefer here when merging the two maps:
```suggestion
    (* prefer attributes from the parameter *)
    let attributes = Attributes.union
      (fun _k a _b -> Some a)
      attributes span.attributes
    in
```
https://v2.ocaml.org/api/Map.S.html#VALunion

ocaml/libs/tracing/tracing.ml
```diff
@@ -138,20 +164,22 @@ module Span = struct
...
         | Unset ->
-            {
-              span with
-              status= {status_code; description}
-            ; tags= span.tags @ exn_tags
-            }
+            let attributes =
+              List.fold_left
+                (fun attrs (k, v) -> Attributes.add k v attrs)
+                Attributes.empty exn_attributes
```
psafont, line 180: The previous code united span.tags and exn_tags. it now unites exn_attributes with en empty map, this doesn't look like it's intended

ocaml/libs/tracing/tracing.ml
```diff
@@ -279,12 +300,10 @@ module Spans = struct
...
                         Span.finish ~span
-                          ~tags:
-                            [
-                              ( "gc_inactive_span_timeout"
-                              , string_of_float elapsed
-                              )
-                            ]
+                          ~attributes:
+                            (Attributes.add "gc_inactive_span_timeout"
+                               (string_of_float elapsed) Attributes.empty
```
psafont, line 305:
```suggestion
                            (Attributes.singleton "gc_inactive_span_timeout"
                               (string_of_float elapsed)
```

ocaml/libs/tracing/tracing.ml
```diff
@@ -400,35 +411,43 @@ let lock = Mutex.create ()
...
         | Some (provider : TracerProvider.t) ->
             let enabled = Option.value ~default:provider.enabled enabled in
-            let tags = Option.value ~default:provider.tags tags in
+            let attributes : string Attributes.t =
+              Option.fold ~none:provider.attributes
+                ~some:
+                  (List.fold_left
+                     (fun attrs (k, v) -> Attributes.add k v attrs)
+                     Attributes.empty
+                  )
```
psafont, line 426: This pattern might be worth encapsulating into a function:

```ocaml
let attrs_of_list ls = List.to_seq |> Attributes.of_seq
```

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: That was a copy paste error, now adjusted

ocaml/xenopsd/lib/xenops_task.ml
```diff
@@ -77,33 +77,13 @@ let parallel_id_with_tracing parallel_id t =
...
-  let* span = Tracing.Span.of_string tracing in
-  Tracing.Span.get_context span
-  |> Tracing.SpanContext.to_traceparent
-  |> Option.some
-
 let dbg_with_traceparent_of_task t =
   Option.fold ~none:(Xenops_task.get_dbg t)
     ~some:(fun traceparent -> Xenops_task.get_dbg t ^ "\x00" ^ traceparent)
-    (traceparent_of_task t)
+    (Xenops_task.tracing t)
```
psafont, line 83: With this change it looks like we're missing a link from the parent, please confirm this is intended

psafont (changes requested):

ocaml/libs/tracing/tracing.ml
```diff
@@ -173,11 +176,12 @@ module Span = struct
...
-            {
-              span with
-              status= {status_code; description}
-            ; attributes= span.attributes @ exn_attributes
-            }
+            let attributes =
+              List.fold_left
+                (fun attrs (k, v) -> Attributes.add k v attrs)
+                span.attributes exn_attributes
+            in
```
psafont, line 183:
```suggestion
            let attributes =
              Attributes.union (fun _k a _b) (attrs_of_list exn_attr) span.attributes
            in
```

ocaml/xenopsd/lib/xenops_task.ml
jameshensmancitrix: This is because xenopsd is now storing the traceparent header instead of the potentially bloated string representation of the Span.

jameshensmancitrix: with this reduction the traceparent_of_task is redundent

ocaml/libs/tracing/tracing.ml
```diff
@@ -138,20 +166,20 @@ module Span = struct
...
         in
         match span.status.status_code with
         | Unset ->
-            {
-              span with
-              status= {status_code; description}
-            ; tags= span.tags @ exn_tags
-            }
+            let attributes =
+              Attributes.union (fun _k a _b) span.attributes (attrs_of_list exn_attributes)
```
psafont, line 180: Sorry, I put an incomplete function, it should have been (fun _k a _b -> Some a), like above

psafont (approved):

jameshensmancitrix:
Going to make sure everything is working then this can be merged.

ocaml/libs/tracing/tracing.ml
```diff
@@ -51,7 +51,14 @@ module Status = struct
   type status_code = Unset | Ok | Error [@@deriving rpcty]
 
   type t = {status_code: status_code; description: string option}
-  [@@deriving rpcty]
+end
+
+module Attributes = Map.Make (String)
```
lindig, line 56: Is this possible? We should keep functions related to attributes in a module.

```
module Attributes = struct
  open Map.Make (String)
  let of_list list = List.to_seq list |> of_seq
end
```

jameshensmancitrix: I wasn't aware this was possible, this looks a lot neater

PR #5029, opened 2023-05-31
CA-375147: UPDATES_REQUIRE_SYNC when toolstack restarted on coordinator

minglumlu:
For repository-based updates, the information of available updates on every host is stored in a cache (in memory) in XAPI process on coordinator. This cache aims to make the HTTP GET /updates requests responded quickly as a fast path. One drawback of this design is that the cache would be flushed when toolstack on coordinator restarted. As part of original design, if this happens, an error 'UPDATES_REQUIRE_SYNC' will be raised to clients so that HTTP-based users would get quick responses in all cases.

This commit aims to fill the cache again (but taking longer time) automatically instead of raising error 'UPDATES_REQUIRE_SYNC' when the cache has been flushed. This will simplify the XAPI call flow from client's perspective.

### DIFF
--- a/ocaml/xapi/repository.ml
+++ b/ocaml/xapi/repository.ml
@@ -615,44 +615,44 @@ let get_repository_handler (req : Http.Request.t) s _ =
 let get_pool_updates_in_json ~__context ~hosts =
   try
     let repository = get_single_enabled_update_repository ~__context in
-    if Hashtbl.length updates_in_cache > 0 then
-      let repository_name = get_repository_name ~__context ~self:repository in
-      let updates_info =
-        parse_updateinfo ~__context ~self:repository ~check:true
-      in
-      let updates_of_hosts, ids_of_updates =
-        Hashtbl.fold
-          (fun host updates_of_host (acc1, acc2) ->
-            if List.mem host hosts then
-              let updates_of_host, upd_ids =
-                consolidate_updates_of_host ~repository_name ~updates_info
-                  (Ref.string_of host) updates_of_host
-              in
-              (updates_of_host :: acc1, UpdateIdSet.union upd_ids acc2)
-            else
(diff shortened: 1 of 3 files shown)


lindig:
> This cache aims to make the HTTP GET /updates requests responded quickly as a fast path.

Why do we need a fast path for this operation? You would think that this happens at most once a day.

minglumlu:
> > This cache aims to make the HTTP GET /updates requests responded quickly as a fast path.
> 
> Why do we need a fast path for this operation? You would think that this happens at most once a day.

This was designed for clients from public cloud. The requests from these clients may go through a long networking path to 
 reach the pool. The long HTTP responding time may cause failure easily in this long networking path.

ocaml/xapi/repository.ml
```diff
@@ -615,44 +615,44 @@ let get_repository_handler (req : Http.Request.t) s _ =
...
-      in
-      `Assoc
-        [
-          ("hosts", `List (List.map HostUpdates.to_json updates_of_hosts))
-        ; ("updates", `List (List.map UpdateInfo.to_json updateinfo_list))
-        ; ("hash", `String (Db.Repository.get_hash ~__context ~self:repository))
-        ]
-    else
-      raise Api_errors.(Server_error (updates_require_sync, []))
+    if Hashtbl.length updates_in_cache = 0 then
```
liulinC, line 618: `apply_updates` will update this hashtable, is the hash table thread safe? or does it need any kind of sync/lock with this API?

liulinC (approved):
Presume the design document has been updated.

ocaml/xapi/repository.ml
minglumlu: It is protected by `with_pool_operation`.

PR #5031, opened 2023-05-31
squeezed: fix link to architectural drawing

ydirson:
This fixes the link from the github repo to the actual drawing.

What's unclear to me is that the contents of http://xapi-project.github.io/squeezed/architecture/architecture.html do look like they were generated from the docs in the same directory, except that
- it is not the case for this image
- the link to the image is the correct one there

The fact that the formatting of that doc does not parse/format the itemized list like github does is also annoying, and I mention it here in the case there would be a single thing to do to fix all those issues.

### DIFF
--- a/ocaml/squeezed/doc/architecture/README.md
+++ b/ocaml/squeezed/doc/architecture/README.md
@@ -6,7 +6,7 @@ Squeezed is responsible for managing the memory on a single host. Squeezed
 
 The following diagram shows the internals of Squeezed:
 
-![Internals of squeezed](http://xapi-project.github.io/squeezed/doc/architecture/squeezed.png)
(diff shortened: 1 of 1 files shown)


psafont (approved):

psafont:
At some point we need to a new framework for the docs (I know that Rob has tried to use hugo for it). The docs are in need of much love, they haven't had attention in years and we need to get better in order to attract technical users

lindig (approved):

ydirson:
> At some point we need to a new framework for the docs

How are the doc pages generated today?

psafont:
They are generated nightly (if there are changes in xapi's datamodel) using jekyll:
https://github.com/xapi-project/xapi-project.github.io/blob/master/.github/workflows/pull-xapi-data.yml
https://github.com/xapi-project/xapi-project.github.io/actions

PR #5032, opened 2023-05-31
Merge vtpm feature branch.

edwintorok:
We are doing most of the development on the master branch nowadays, but there were a few commits still on the vTPM feature branch. Merge them to master.

The 2 fixups are new, they fix a bug in the newly added quicktest (the feature branch doesn't allow force pushes though, so I couldn't squash them).


edwintorok:
Converting to draft again, although the quicktest works now when manually testing there is still something wrong when running it via automated testing (or my build was out of date). Will add some more debugging and test again.

psafont:
Let's close this PR in favour of #5033

PR #5038, opened 2023-06-02
Write bugtool exports in batches

jameshensmancitrix:
Instead of writing a new file for every trace we export, write into files split into 1mb batches. This will save disk space as the number of files will be drastically reduced.

### DIFF
--- a/ocaml/libs/tracing/tracing.ml
+++ b/ocaml/libs/tracing/tracing.ml
@@ -574,29 +574,75 @@ module Export = struct
     module File = struct
       let trace_log_dir = ref "/var/log/dt/zipkinv2/json"
 
+      let max_file_size = ref (1 lsl 20)
+
       let set_trace_log_dir dir = trace_log_dir := dir
 
-      let export ~trace_id ~span_json ~path : (string, exn) result =
-        try
-          let date = Ptime_clock.now () |> Ptime.to_rfc3339 ~frac_s:6 in
-          let file =
-            path
-            ^ String.concat "-" [trace_id; "xapi"; !host_id; date]
-            ^ ".json"
-          in
-          Xapi_stdext_unix.Unixext.mkdir_rec (Filename.dirname file) 0o700 ;
-          Xapi_stdext_unix.Unixext.write_string_to_file file span_json ;
-          Ok ""
-        with e -> Error e
+      let set_max_file_size size = max_file_size := size
+
+      let file_name = ref None
+
+      let file_stream = ref None
+
+      let lock = Mutex.create ()
+
+      let new_file_name () =
+        let date = Ptime_clock.now () |> Ptime.to_rfc3339 ~frac_s:6 in
+        let ( // ) = Filename.concat in
+        let name =
+          !trace_log_dir
+          // String.concat "-" [get_service_name (); !host_id; date]
+          ^ ".json"
+        in
+        file_name := Some name ;
+        name
+
+      let open_stream file_name =
+        Xapi_stdext_unix.Unixext.mkdir_rec (Filename.dirname file_name) 0o700 ;
+        file_stream :=
+          Some (Unix.openfile file_name [O_WRONLY; O_CREAT; O_APPEND] 0o700)
+
+      let close_stream () =
+        Option.iter (fun stream -> Unix.close stream) !file_stream ;
+        file_stream := None
+
+      let check () =
+        match (!file_stream, !file_name) with
+        | Some file_stream, _
+          when (Unix.fstat file_stream).st_size >= !max_file_size ->
+            close_stream () ;
+            new_file_name () |> open_stream
+        | None, None ->
+            new_file_name () |> open_stream
+        | None, Some file_name ->
+            open_stream file_name
+        | _ ->
+            (* We have an open fd which points to a file under the limit*)
+            ()
+
+      let write str =
+        Option.iter
+          (fun stream ->
+            let content = Bytes.of_string (str ^ "\n") in
+            ignore @@ Unix.write stream content 0 (Bytes.length content)
+          )
+          !file_stream
+
+      let export json = try check () ; write json ; Ok () with e -> Error e
+
+      let with_stream f =
+        Xapi_stdext_threads.Threadext.Mutex.execute lock (fun () ->
+            f export ; close_stream ()
+        )
     end
 
     module Http = struct
       module Request = Cohttp.Request.Make (Cohttp_posix_io.Buffered_IO)
       module Response = Cohttp.Response.Make (Cohttp_posix_io.Buffered_IO)
 
-      let export ~span_json ~url : (string, exn) result =
+      let export ~url json =
         try
-          let body = span_json in
+          let body = json in
           let headers =
             Cohttp.Header.of_list
               [
@@ -617,7 +663,7 @@ module Export = struct
               Unix.shutdown fd Unix.SHUTDOWN_SEND ;
               match try Response.read ic with _ -> `Eof with
               | `Eof ->
-                  Ok ""
+                  Ok ()
               | `Invalid x ->
                   Error (Failure ("invalid read: " ^ x))
               | `Ok response ->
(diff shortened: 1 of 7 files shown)


ocaml/libs/tracing/tracing.ml
```diff
@@ -576,17 +576,70 @@ module Export = struct
...
+
+        let new_file_name path =
           let date = Ptime_clock.now () |> Ptime.to_rfc3339 ~frac_s:6 in
-          let file =
-            path
-            ^ String.concat "-" [trace_id; "xapi"; !host_id; date]
+          let ( // ) = Filename.concat in
+          let name =
+            (path // String.concat "-" [get_service_name (); !host_id; date])
             ^ ".json"
```
lindig, line 595: This can be part of the `String.concat`

ocaml/libs/tracing/tracing.ml
```diff
@@ -642,22 +695,22 @@ module Export = struct
...
+                  File.export ~span_json:zipkin_spans ~path:!File.trace_log_dir
             with
             | Ok _ ->
                 ()
             | Error e ->
                 raise e
           )
-          span_list
+          span_list ;
+        File.FileStream.close_stream ()
```
psafont, line 713: Separating concerns leads to simpler-looking code: the export function can be defined first, and then be used for processing the spans:
```ocaml
let export = match endpoint with
  | Url url ->  Http.export ~url
  | Bugtool -> File.export ~path:!File.trace_log_dir
in
Hashtbl.iter (fun _ span ->
  let span_json = Content.Json.Zipkinv2.content_of span in
  match export ~span_json with
  | Ok _ -> ()
  | Error e -> raise e
  )
  span_list ;
```

ocaml/libs/tracing/tracing.ml
```diff
@@ -576,17 +576,70 @@ module Export = struct
...
+          Xapi_stdext_unix.Unixext.mkdir_rec (Filename.dirname file_name) 0o700 ;
+          file_stream :=
+            Some (Unix.openfile file_name [O_WRONLY; O_CREAT; O_APPEND] 0o700)
+
+        let close_stream' () =
+          Option.iter (fun stream -> Unix.close stream) !file_stream ;
+          file_stream := None
+
+        let close_stream () =
+          Xapi_stdext_threads.Threadext.Mutex.execute lock (fun () ->
```
lindig, line 610: Keep these in one function. The easy way to do this is:

```
Xapi_stdext_threads.Threadext.Mutex.execute lock @@ fun () ->
Option.iter (fun stream -> Unix.close stream) !file_stream ;
file_stream := None
```

ocaml/libs/tracing/tracing.ml
```diff
@@ -576,17 +576,70 @@ module Export = struct
...
+              close_stream' ()
+          )
+
+        let check path =
+          Xapi_stdext_threads.Threadext.Mutex.execute lock (fun () ->
```
lindig, line 615: See above how to keep this simple.

ocaml/xapi/xapi_globs.ml
```diff
@@ -988,6 +988,9 @@ let max_spans = ref 1000
 
 let max_traces = ref 10000
 
+(** 1MB *)
+let max_observer_file_size = ref 1048576
```
lindig, line 992: Would be better to express these numbers as a product of powers of two - otherwise it's not clear why we have seemingly special numbers. Or write a functions like `mib` and `kib` to denote common sizes.

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: The reason this is being done is to avoid deadlock in `check` which is used to close a stream

ocaml/libs/tracing/tracing.ml
lindig: Sorry, I was wrong; because `-` is used to concatenate. Otherwise
```
path // String.concat "" [get_service_name (); "-"; !host_id; "-"; date; ".json"]
```

jameshensmancitrix: In that case I may as well just used `^`

ocaml/libs/tracing/tracing.ml
```diff
@@ -576,17 +576,70 @@ module Export = struct
 
       let set_trace_log_dir dir = trace_log_dir := dir
 
-      let export ~trace_id ~span_json ~path : (string, exn) result =
-        try
+      module FileStream = struct
```
psafont, line 579: Because we don't want to leave the file descriptrs only we really want to expose an interface to the Filestream that will close the resources implicitly when the user function finishes writing everything.

This can be done with a Filestream.with function that has a function as a parameter that is given an export function which can use. and once this function finishes, the finaliser can close the file descriptors.

```ocaml
module Filestream = struct
  val with : unit -> (span_json : ... -> (string, exn) result) -> unit) -> unit
end
...
    let export_to_endpoint span_list endpoint =
      try
        debug "Tracing: About to export" ;
        let ( let@ ) f x = f x in
        let@ file_export = Filestream.(with ~dir:!trace_log_dir ()) in
        let export = match endpoint with
        | Url url ->  Http.export ~url
        | Bugtool -> file_export
        Hashtbl.iter (...) span_list
      with e -> debug "Tracing: ERROR %s" (Printexc.to_string e)
```

Once this interface is stablished I think the module File doesn't contain anything as the export function is not provided by Filestream, and can be dropped. This design also allows set_trace_log_dir to be stable over a whole run of span exports, previously it could be changed while the export was in process.

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: [c783311](https://github.com/xapi-project/xen-api/pull/5038/commits/c783311f36d5b9e362947efad4376a27b8fd74b1)

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: [c783311](https://github.com/xapi-project/xen-api/pull/5038/commits/c783311f36d5b9e362947efad4376a27b8fd74b1)

ocaml/libs/tracing/tracing.ml
psafont: "^" has quadratic cost, please don't use it to join so many strings

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: Part of the idea currently is to keep resources open throughout the iter, would this be affect here?

jameshensmancitrix:
Digging some more into the export functions I can see that could reduce the used of `Result` as these are only going to lead to Errors being raised and caught in `export_to_endpoint`. I could also refactor this area to need less result types

psafont (approved):

ocaml/libs/tracing/tracing.ml
```diff
@@ -574,38 +574,90 @@ module Export = struct
...
-            Cohttp.Header.of_list
-              [
-                ("accepts", "application/json")
-              ; ("content-type", "application/json")
-              ; ("content-length", string_of_int (String.length body))
-              ]
-          in
-          Open_uri.with_open_uri url (fun fd ->
+      let export ~url ~span_json =
+        let body = span_json in
```
mg12, line 650: we probably want to keep the try statement at the beginning of the function, so that all the call to libraries below are protected by it.

lindig (approved):

ocaml/libs/tracing/tracing.ml
```diff
@@ -574,38 +574,90 @@ module Export = struct
     module File = struct
       let trace_log_dir = ref "/var/log/dt/zipkinv2/json"
 
+      let max_file_size = ref 1048576
```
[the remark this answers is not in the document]
jameshensmancitrix: I see what you mean, I have added that there and in `xapi_globs`

PR #5040, opened 2023-06-05
Add new fields to SpanContext

jameshensmancitrix:
Based off the OpenTelemetry specification **[here](https://opentelemetry.io/docs/specs/otel/trace/api/#spancontext)**, add more methods and fields to the SpanContext. This also includes adding `tracestate` as a header to `http_svr` and using sampling based off the new flags field.


jameshensmancitrix:
To be moved to master

PR #5041, opened 2023-06-05
Remove most vTPM restrictions

lindig:
We are moving to a new backend for vTPM state storage that is updated as soon as the swTPM updates it. This permits to remove some restrictions that are in place to prohibit vTPM state being lost.

* cloning of a running VM
* cross-pool migration of a running VM
* checkpointing of a running VM
* creating a vTPM on a VM that is protected by high availability (HA)

One restriction remains in place: we still can't create a vTPM for a VM with a BIOS (in contrast to UEFI).

### DIFF
--- a/ocaml/xapi/xapi_vtpm.ml
+++ b/ocaml/xapi/xapi_vtpm.ml
@@ -27,19 +27,6 @@ let assert_not_restricted ~__context =
   | _ ->
       raise Api_errors.(Server_error (feature_restricted, [feature]))
 
-(** The state in the xapi backend is only up-to-date when the VMs are halted *)
-let assert_no_fencing ~__context ~persistence_backend =
-  let pool = Helpers.get_pool ~__context in
-  let ha_enabled = Db.Pool.get_ha_enabled ~__context ~self:pool in
-  let clustering_enabled = Db.Cluster.get_all ~__context <> [] in
-  let may_fence = ha_enabled || clustering_enabled in
-  match (persistence_backend, may_fence) with
-  | `xapi, true ->
-      let message = "VTPM.create with HA or clustering enabled" in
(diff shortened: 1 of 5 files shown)


psafont (approved):

psafont:
This should be merged right after https://github.com/xapi-project/xen-api/pull/5000

ocaml/xapi/xapi_vtpm.ml
```diff
@@ -27,19 +27,6 @@ let assert_not_restricted ~__context =
   | _ ->
       raise Api_errors.(Server_error (feature_restricted, [feature]))
 
-(** The state in the xapi backend is only up-to-date when the VMs are halted *)
-let assert_no_fencing ~__context ~persistence_backend =
-  let pool = Helpers.get_pool ~__context in
-  let ha_enabled = Db.Pool.get_ha_enabled ~__context ~self:pool in
-  let clustering_enabled = Db.Cluster.get_all ~__context <> [] in
-  let may_fence = ha_enabled || clustering_enabled in
```
edwintorok, line 35: This is the one we discussed earlier today.
Perhaps a solution would be to have the equivalent of a mini-redolog active whenever clustering is enabled, using Dom0's filesystem for the redo log. That would not require any action from the user to chose a shared storage for the redo log (although of course if they do chose one then we can use that), and can be done as part of enabling clustering internally in XAPI to also enable this internal redo log mode.

But I wouldn't want to tie together making that change with vTPM, using clustering without enabling any form of redo log or HA is already known to be a bad idea.

edwintorok (approved):

edwintorok:
There is still one check that remained active:
```
assert_ha_vtpms_compatible```

Which prevents enabling HA when vTPM is present ('assert_no_fencing' wasn't the only thing blocking HA).
Can you remove that too?

PR #5042, opened 2023-06-06
CA-354436: pool.is_slave took a long time to respond

minglumlu:
This commit is to fix an issue which may be hit when the pool coordinator is rebooting or its XAPI is restarting.

In either of the cases, other pool members keep trying to connect to the coordinator for DB connections. This connecting holds the DB lock until a timer expired. And the timer uses exponential backoff to avoid overload the coordinator. The maximum of the timer is 256 seconds.

Meanwhile in starting up, the XAPI in coordinator will issue XAPI call pool.is_slave to every pool member. This call on every pool member will try to acquire the DB lock as well and hence be blocked by the re-connecting mentioned above. In the worst case, the start-up of XAPI would be delayed up to more than 4 minutes (256 seconds).

The idea of this commit is that since the pool.is_slave is issued from the coordinator, it is reasonable to abort the waiting for the timer expiry and try connecting again.

### DIFF
--- a/ocaml/database/master_connection.ml
+++ b/ocaml/database/master_connection.ml
@@ -28,6 +28,8 @@ open D
 
 let my_connection : Stunnel.t option ref = ref None
 
+let delay = Scheduler.PipeDelay.make ()
+
 exception Uninitialised
 
 let is_slave : (unit -> bool) ref =
@@ -315,7 +317,9 @@ let do_db_xml_rpc_persistent_with_reopen ~host:_ ~path (req : string) :
           ) ;
         debug "Sleeping %f seconds before retrying master connection..."
           !backoff_delay ;
-        Thread.delay !backoff_delay ;
+        let timed_out = Scheduler.PipeDelay.wait delay !backoff_delay in
+        if not timed_out then
+          debug "%s: Wake up and retry connecting" __FUNCTION__ ;
         update_backoff_delay () ;
         try open_secure_connection () with _ -> ()
         (* oh well, maybe nextime... *)
--- a/ocaml/database/master_connection.ml
+++ b/ocaml/database/master_connection.ml
@@ -319,7 +319,8 @@ let do_db_xml_rpc_persistent_with_reopen ~host:_ ~path (req : string) :
           !backoff_delay ;
         let timed_out = Scheduler.PipeDelay.wait delay !backoff_delay in
         if not timed_out then
-          debug "%s: Wake up and retry connecting" __FUNCTION__ ;
+          debug "%s: Sleep interrupted, retrying master connection now"
+            __FUNCTION__ ;
         update_backoff_delay () ;
         try open_secure_connection () with _ -> ()
         (* oh well, maybe nextime... *)
--- a/ocaml/xapi/xapi_pool.ml
+++ b/ocaml/xapi/xapi_pool.ml
@@ -2120,6 +2120,7 @@ let is_slave ~__context ~host:_ =
   debug
     "About to kick the database connection to make sure it's still working..." ;
   let (_ : bool) =
+    Scheduler.PipeDelay.signal Master_connection.delay ;
     Db.is_valid_ref __context
       (Ref.of_string
          "Pool.is_slave checking to see if the database connection is up"
--- a/ocaml/xapi-idl/lib/scheduler.mli
+++ b/ocaml/xapi-idl/lib/scheduler.mli
@@ -47,3 +47,18 @@ val one_shot : t -> time -> string -> (unit -> unit) -> handle
 
 val cancel : t -> handle -> unit
 (** Cancel an item *)
+
+(** The PipeDelay module here implements simple cancellable delays. *)
+module PipeDelay : sig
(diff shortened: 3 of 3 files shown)


ocaml/database/master_connection.ml
```diff
@@ -215,6 +221,23 @@ let do_db_xml_rpc_persistent_with_reopen ~host:_ ~path (req : string) :
...
+  let rec delay = function
+    | d when d > 16.0 -> (
+        (* Sleep for a small time in case that the coordinator can connect but can't accept *)
+        delay (d /. 2.0) ;
+        match Atomic.get in_call_from_coordinator with
+        | true ->
+            debug
+              "%s: The coordinator is calling me. Wake up and retry connecting."
+              __FUNCTION__ ;
+            Atomic.set in_call_from_coordinator false
```
psafont, line 233: This one doesn't seem needed, it's set to false in line 240 whenever the function delay finishes

ocaml/database/master_connection.ml
```diff
@@ -215,6 +221,23 @@ let do_db_xml_rpc_persistent_with_reopen ~host:_ ~path (req : string) :
...
+        | false ->
+            delay (d /. 2.0)
```
psafont, line 235: It's better if the read and write operations happen atomically:
```suggestion
        let changed = Atomic.compare_and_set in_call_from_coordinator true false in
        if changed then
          debug
            "%s: The coordinator is calling me. Wake up and retry connecting."
            __FUNCTION__ 
        else
          delay (d /. 2.0)
```
https://v2.ocaml.org/api/Atomic.html#VALcompare_and_set

ocaml/database/master_connection.ml
```diff
@@ -215,6 +221,23 @@ let do_db_xml_rpc_persistent_with_reopen ~host:_ ~path (req : string) :
...
+      )
+    | _ as d ->
+        Thread.delay d
+  in
+  Atomic.set in_call_from_coordinator false ;
```
psafont, line 240: Why is it set to false unconditionally? it's not obvious

ocaml/xapi/xapi_pool.ml
```diff
@@ -2120,6 +2120,7 @@ let is_slave ~__context ~host:_ =
   debug
     "About to kick the database connection to make sure it's still working..." ;
   let (_ : bool) =
+    Atomic.set Master_connection.in_call_from_coordinator true ;
```
psafont, line 2123: Shouldn't it be set to false after the database check is done?

minglumlu: The indicator is a boolean value. Setting to false may conflict with other callers (although there are no such callers at the moment).

ocaml/database/master_connection.ml
minglumlu: This is related with the setting in callers (in host.is_slave).
Setting here would get the status as expected before delay regardless how the callers would do.

edwintorok:
We have a Delay module that can wait for a timeout, but is interruptible with a Delay.signal. (it uses pipes internally instead of sleep).
That might allow to interrupt the sleep even sooner. Use the Delay module to perform the sleep, and wherever you would do the Atomic set to abort the sleep use the Delay.signal to interrupt the sleep instead

ocaml/database/master_connection.ml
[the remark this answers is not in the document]
minglumlu: > Why wait when we are in call from the host that we try to connect to?

This is not in call from the coordinator, but it is holding the DB lock.
Another call from the coordinator (XAPI is restarting) would try to acquire the DB lock as well and hence be blocked.

> As written, we always delay d/2 first but skip the second delay d/2 when called from the coordinator. Is this the intention?

This first approach was to keep waking up from small timers (16 secs). Without the first delay, it might be connecting busily if the coordinator couldn't accept any connections yet.

minglumlu:
Hi @psafont, @lindig 
I followed @edwintorok 's suggestion to use `PipeDelay` which could be interrupted by signal directly instead of keeping waking up and checking. Could you please review the new changes? Thanks!

ocaml/database/master_connection.ml
```diff
@@ -315,7 +317,9 @@ let do_db_xml_rpc_persistent_with_reopen ~host:_ ~path (req : string) :
           ) ;
         debug "Sleeping %f seconds before retrying master connection..."
           !backoff_delay ;
-        Thread.delay !backoff_delay ;
+        let timed_out = Scheduler.PipeDelay.wait delay !backoff_delay in
+        if not timed_out then
+          debug "%s: Wake up and retry connecting" __FUNCTION__ ;
```
psafont, line 322:
```suggestion
          debug "%s: Sleep interrupted, retrying master connection now" __FUNCTION__ ;
```

minglumlu:
> This should work, has it been tested? @minglumlu

Yes. I tested it manually and automation of regression tests.

psafont (approved):

robhoes (approved):

PR #5044, opened 2023-06-06
Trace Export functions to monitor performance 

jameshensmancitrix:
Given concerns that exporting over http was being slow, this pull request will create spans that measure how performant the Export functions are.

### DIFF
--- a/ocaml/libs/tracing/tracing.ml
+++ b/ocaml/libs/tracing/tracing.ml
@@ -737,10 +737,6 @@ module Export = struct
     let export_to_endpoint parent traces endpoint =
       debug "Tracing: About to export" ;
       try
-        let attributes = [("export.endpoint", endpoint_to_string endpoint)] in
-        let@ parent =
-          with_tracing ~parent ~attributes ~name:"Tracing.export_to_endpoint"
-        in
         File.with_stream (fun file_export ->
             let export, name =
               match endpoint with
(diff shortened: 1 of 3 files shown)


ocaml/libs/tracing/tracing.ml
```diff
@@ -696,36 +721,44 @@ module Export = struct
...
-            Ok ()
-            |> Hashtbl.fold
-                 (fun _ spans result ->
-                   Content.Json.Zipkinv2.content_of spans
-                   |> export
-                   |> Result.fold ~ok:(fun () -> result) ~error:Result.error
-                 )
-                 traces
-            |> Result.iter_error raise
+        with_tracing ~parent ~name:"Tracing.export_to_endpoint" (fun parent ->
```
psafont, line 727: the new let@ trick can help with minimizing the changeset by keeping indentation the same :)

`let@ parent = with_tracing ~parent ~name:... in`

psafont (approved):

ocaml/libs/tracing/tracing.ml
jameshensmancitrix: If this new, will I be able to do this now or wait until 4.14?

psafont: this is not included in ocaml 4.14, it needs to be defined: `let ( let@ ) f x = f x`

robhoes (approved):

PR #5045, opened 2023-06-06
CA-376864: prefer use of NBD path for static VDIs on SMAPIv1

MarkSymsCtx:


scripts/static-vdis
```diff
@@ -220,12 +220,28 @@ def doexec(args, inputtext=None):
...
     return (rc,stdout,stderr)
 
+
+def connect_smapiv1_nbd(params_nbd):
+    # Format "nbd:unix:<socket path>:exportname=<export name>"
+    parts = params_nbd.split(':')
+    if len(parts) != 4:
+        raise Exception("Malformed NBD params %s" % params_nbd)
+
+    export = parts[3].split('=')[1]
```
lindig, line 230: Is there any chance that this fails because the `=` is missing?

lindig (approved):

scripts/static-vdis
```diff
@@ -220,12 +220,28 @@ def doexec(args, inputtext=None):
     rc = proc.returncode
     return (rc,stdout,stderr)
 
+
+def connect_smapiv1_nbd(params_nbd):
+    # Format "nbd:unix:<socket path>:exportname=<export name>"
```
robhoes, line 225: Does `parse_nbd_uri` not do the right thing?

MarkSymsCtx: You're correct it does exactly what's required, I just didn't spot it in the code.

MarkSymsCtx:
Updated and retested using the `parse_nbd_uri` method

psafont (approved):

PR #5046, opened 2023-06-07
CP-40204, CA-366396: Add "host.up_to_date", remove "repository.up_to_date"

gangj:
In repository-based continuous update, XAPI users/clients would like to know if the versions of software installed on individual hosts are consistent in a CH pool.

The consistency was defined as "a host is running same versions of everything as the coordinator" in hotfix-based discrete update.

Now the consistency is defined as "a host is running same versions of everything as the latest ones known in last pool.sync_updates".

Therefore, a new boolean filed "host.up_to_date" should be added as the indicator.

This field should be set in:

    pool.sync_updates, and
    host.apply_updates

### DIFF
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -1761,6 +1761,24 @@ let apply_recommended_guidances =
       ]
     ~allowed_roles:_R_POOL_OP ()
 
+let up_to_date_state =
+  Enum
+    ( "up_to_date_state"
+    , [
+        ( "yes"
+        , "The host is up to date with the latest updates synced from remote \
+           CDN"
+        )
+      ; ( "no"
+        , "The host is outdated with the latest updates synced from remote CDN"
+        )
+      ; ( "unknown"
+        , "If the host is up to date with the latest updates synced from \
+           remote CDN is unknown"
+        )
+      ]
+    )
+
 (** Hosts *)
 let t =
   create_obj ~in_db:true ~in_product_since:rel_rio ~in_oss_since:oss_since_303
@@ -2120,6 +2138,10 @@ let t =
         ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:(Set update_guidances)
             "recommended_guidances" ~default_value:(Some (VSet []))
             "The set of recommended guidances after applying updates"
+        ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:up_to_date_state
+            "up_to_date" ~default_value:(Some (VEnum "unknown"))
+            "Default as 'unknown', 'yes' if the host is up to date with \
+             updates synced from remote CDN, otherwise 'no'"
         ]
       )
     ()
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -1761,9 +1761,9 @@ let apply_recommended_guidances =
       ]
     ~allowed_roles:_R_POOL_OP ()
 
-let up_to_date_state =
+let latest_synced_updates_applied_state =
   Enum
-    ( "up_to_date_state"
+    ( "latest_synced_updates_applied_state"
     , [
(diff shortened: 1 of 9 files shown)


ocaml/xapi/repository.ml
```diff
@@ -333,92 +292,69 @@ let get_hosts_updates ~__context =
...
+                   enabled
+                )
+            in
+            not
+              (is_livepatchable ~__context update_repo
+                 (get_applied_livepatches_of_host updates_of_host)
+              )
+      in
+      let up_to_date_state =
+        match up_to_date with true -> `yes | false -> `no
```
minglumlu, line 347: No need to match 3 times. Only one bool data is flowing through the whole logic. Would match be enough?

ocaml/idl/datamodel_host.ml
```diff
@@ -2120,6 +2138,10 @@ let t =
         ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:(Set update_guidances)
             "recommended_guidances" ~default_value:(Some (VSet []))
             "The set of recommended guidances after applying updates"
+        ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:up_to_date_state
```
minglumlu, line 2141: Could this be typed as `Option Bool`?

minglumlu:
Another case is when a host has been updated successfully, the field should be set.

ocaml/idl/datamodel_host.ml
```diff
@@ -1761,6 +1761,24 @@ let apply_recommended_guidances =
       ]
     ~allowed_roles:_R_POOL_OP ()
 
+let up_to_date_state =
+  Enum
+    ( "up_to_date_state"
+    , [
+        ( "yes"
```
psafont, line 1768: I would a name that is more informative "up to date" can mean that it has the latest package metadata, or that it has installed it.
```
    ( "latest_known_packages_installed"
    , [
        ( "true"
```

stormi:
So, for months I've been following more or less closely the changes in XAPI regarding host update management, in the hope that I could use it in XCP-ng too, but it's really hard for me to understand where this is going, to the point that for now I have little hope that I can really use XAPI to manage a host's RPM updates in the context of XCP-ng.
 
 Is there a design document that might be shared publicly and commented, so that we may know where XAPI is going regarding update management?

gangj:
> Another case is when a host has been updated successfully, the field should be set.

It is set to "'yes" in xapi_host.ml inside function "apply_updates":
"Db.Host.set_up_to_date ~__context ~self ~value:`yes ;"

ocaml/idl/datamodel_host.ml
gangj: Yes, I think it can, but when getting it, you will have to match it with None to do a translation like:

      ; make_field ~name:"latest-synced-updates-applied"
          ~get:(fun () ->
             match  (x ()).API.host_latest_synced_updates_applied with
             | Some a -> Bool.to_string a
             | None -> "unknown"
          )
          ()
      ]

While with a new type like:

let latest_synced_updates_applied_state =
  Enum
    ( "latest_synced_updates_applied_state"
    , [
        ( "yes"
        , "The host is up to date with the latest updates synced from remote \
           CDN" 
        )
      ; ( "no"
        , "The host is outdated with the latest updates synced from remote CDN" 
        )
      ; ( "unknown"
        , "If the host is up to date with the latest updates synced from \
           remote CDN is unknown"
        )
      ]
    )

The definition of all the state is in one place, I think this is easier to understand and would like to keep it.

ocaml/idl/datamodel_host.ml
gangj: As suggested by Ming, renamed it to "latest_synced_updates_applied", as besides rpm packages, there are also livepatches to be applied.

minglumlu (approved):

liulinC (approved):

PR #5047, opened 2023-06-08
CP-42999: Return new "preview" in return of v6 "get_version"

minglumlu:
This commit adds a new field "preview" in return of v6 interface "get_version".

To achieve this, the change has to extend the return type of "get_version" interface so that more version related information could be supported.

### DIFF
--- a/ocaml/xapi-idl/v6/v6_interface.ml
+++ b/ocaml/xapi-idl/v6/v6_interface.ml
@@ -51,16 +51,6 @@ type edition = {
 }
 [@@deriving rpcty]
 
-(** Record representing version *)
-type version = {
-    dbv: string  (** Date-Based Version *)
-  ; preview: bool  (** In preview phase or not *)
-}
-[@@deriving rpcty]
-
-let assoc_list_of_version version =
-  [("dbv", version.dbv); ("preview", string_of_bool version.preview)]
-
 (** List of edition records *)
 type edition_list = edition list [@@deriving rpcty]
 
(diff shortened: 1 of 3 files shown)


ocaml/xapi-idl/v6/v6_interface.ml
```diff
@@ -51,6 +51,16 @@ type edition = {
 }
 [@@deriving rpcty]
 
+(** Record representing version *)
+type version = {
+    dbv: string  (** Date-Based Version *)
+  ; preview: bool  (** In preview phase or not *)
```
psafont, line 57: I'm sorry, but I have to do the same nitpick as on the previous PR, "preview" is not enough to understand what the field does at first glance

 ```suggestion
  ; is_prerelease: bool  (** In preview / prerelease phase or not *)
```
Adding is_ is not enough to make it clarifying enough, stating it's a prerelease version makes everything clear to me, I'd like to hear other opinions

minglumlu: Thought it about for a while. I would prefer the `is_preview_release` which was suggested by @lindig. I changed it back to `preview` as I thought it would be clear under the context of version-related info.

`prerelease` seems not a release to me. But actually it is a release indeed.

minglumlu: BTW, with @robhoes 's comment above, the specific name will not be hardcoded in XAPI.

psafont: That sounds good as it releases xcp-ng from knowing about specific fields. For their purposes they can change their daemon to use the new interface and pass around any information they might want (or none at all)
@stormi

psafont (approved):

robhoes (approved):

PR #5048, opened 2023-06-08
xapiguard_cli: run its tests as part of varstored package

psafont:
Otherwise dune not find the executable when only xapi-idl is tested

This fixes xs-opam-repo's CI


psafont:
The change is included in #5001

PR #5052, opened 2023-06-08
CA-378222: assert_sr_can_host_statefile has to take available space i…

edwintorok:
…nto consideration

If a statefile of the required size is not already present on the SR then we need to check that there is sufficient space available to create the statefile.

Otherwise when multiple SRs are available we may pick the wrong one to create the statefile on, even though a suitable one would've been found if we tried the other ones.


edwintorok:
Not yet tested, hence a draft

edwintorok:
The error message here is not ideal (it says 'source SR' ran out of space, and doesn't say how much it would need).
Might be useful to introduce a new error message specifically for this (although that'll also require a XenCenter update for localization of the error message).

ocaml/xapi/xha_statefile.ml
```diff
@@ -89,15 +89,54 @@ let assert_sr_can_host_statefile ~__context ~sr ~cluster_stack =
...
         raise
           (Api_errors.Server_error
              (Api_errors.sr_operation_not_supported, [Ref.string_of sr])
+          ) ;
+      let size = minimum_size in
+      match
+        List.filter
+          (fun self ->
+            true
+            && Db.VDI.get_type ~__context ~self = `ha_statefile
```
psafont, line 106:
```suggestion
            Db.VDI.get_type ~__context ~self = `ha_statefile
```

ocaml/xapi/xha_statefile.ml
```diff
@@ -89,15 +89,54 @@ let assert_sr_can_host_statefile ~__context ~sr ~cluster_stack =
...
+            && Db.VDI.get_virtual_size ~__context ~self >= size
           )
+          (Db.SR.get_VDIs ~__context ~self:sr)
+      with
```
psafont, line 110:
```suggestion
      let ha_fits self = 
        Db.VDI.get_type ~__context ~self = `ha_statefile
        && Db.VDI.get_virtual_size ~__context ~self >= size
      in
      match List.filter ha_fits (Db.SR.get_VDIs ~__context ~self:sr) in
```

ocaml/xapi/xha_statefile.ml
```diff
@@ -89,15 +89,54 @@ let assert_sr_can_host_statefile ~__context ~sr ~cluster_stack =
...
+              (Api_errors.Server_error
+                 (Api_errors.sr_source_space_insufficient, [Ref.string_of sr])
+              )
+          ) else
+            None
+    )
+
+let assert_sr_can_host_statefile ~__context ~sr ~cluster_stack =
+  let (_ : 'a option) =
+    check_sr_can_host_statefile ~__context ~sr ~cluster_stack
```
psafont, line 137: Might be worth using Result.t instead of exceptions

ocaml/database/redo_log.ml
```diff
@@ -888,6 +888,26 @@ let flush_db_to_redo_log db log =
     write_db_to_fd log ;
   !(log.currently_accessible)
 
+let flush_db_or_raise db log =
```
lindig, line 891: We have used `exn` as suffix to indicate a function may fail. I like `or_raise` but it's a bit longer.

ocaml/xapi/xha_statefile.ml
```diff
@@ -89,15 +89,54 @@ let assert_sr_can_host_statefile ~__context ~sr ~cluster_stack =
...
       then
         raise
           (Api_errors.Server_error
              (Api_errors.sr_operation_not_supported, [Ref.string_of sr])
+          ) ;
+      let size = minimum_size in
+      match
+        List.filter
+          (fun self ->
+            true
```
lindig, line 105: What is the `true` doing? I notice it is coming from existing code.

lindig (approved):

edwintorok:
Test has passed: without this change we lose database objects/fields (sometimes even the entire VM object is lost), and with this change we no longer lose them (and we don't lose vTPM state either).
I'll address the review comments and then mark this as ready for review.

ocaml/xapi/xha_statefile.ml
psafont: I removed it from my suggestion, it's useless

edwintorok: it aligns the next 2 clauses by having them both have '&&', I'll drop it.

ocaml/xapi/xha_statefile.ml
edwintorok: exceptions are needed because this function is eventually exposed in the API, which has to return an error. We can clean this up, but that'd need a lot more changes inside the 'check_sr_can_host_statefile' function to convert all the exceptions in to results, only for at the end to convert the results back into exceptions.

ocaml/database/redo_log.ml
```diff
@@ -888,6 +888,26 @@ let flush_db_to_redo_log db log =
     write_db_to_fd log ;
   !(log.currently_accessible)
 
+let flush_db_exn db log =
```
psafont, line 891: Might be worth having a small module here:

```ocaml
module Flush = struct
  type t = private Flush of (Db_cache_types.Database.t * redo_log) | No of redo_log

  val no = No

  val flush db log = if db.read_only then No else Flush db log
end
```
then this can be come
```ocaml
let flush_db_exn = function
| No _ ->
  ()
| Flush (db, log) ->
  R.debug "Flushing..."
```
and can be used like
```ocaml
Redo_log.enable Xapi_ha.(Flush.no ha_redo_log) Xapi_globs.gen_metadata_vdi_reason ;
...
let flush = Flush.flush (Context.database_of __context |> Db_ref.get_database) ha_redo_log in
Redo_log.enable flush Xapi_globs.ha_metadata_vdi_reason
```

edwintorok: I've pushed some commits implementing this idea, and split the enable functions into 2: `enable_existing` which doesn't take a database as a parameter and accepts both RO and RW redologs. And `enable_and_flush` which takes only RW redologs and flushes. (enable_existing needs to accept RW because the HA redo log starts out RO, to read back old state, and then becomes RW)

psafont (approved):

PR #5055, opened 2023-06-09
CP-43556 Enhance features in SpanContext

jameshensmancitrix:
Currently the SpanContext only holds 2 values and the [OpenTelemetry Specification](https://opentelemetry.io/docs/specs/otel/trace/api/#spancontext) has 5 and a number of methods. The concept of Sampling, while always set to true is introduced here.

### DIFF
--- a/ocaml/libs/tracing/tracing.ml
+++ b/ocaml/libs/tracing/tracing.ml
@@ -132,6 +132,18 @@ module SpanContext = struct
     | tracestate ->
         Some tracestate
 
+  let tracestate_get key t = List.assoc_opt key t.tracestate
+
+  let tracestate_delete key t =
+    let tracestate = List.filter (fun (k, _) -> k <> key) t.tracestate in
+    {t with tracestate}
+
+  let tracestate_replace key value t =
+    let tracestate =
+      (key, value) :: List.filter (fun (k, _) -> k <> key) t.tracestate
+    in
+    {t with tracestate}
+
   let of_traceparent ?tracestate traceparent =
     let tracestate =
       Option.fold ~none:[]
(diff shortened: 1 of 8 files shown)


ocaml/libs/tracing/tracing.ml
```diff
@@ -78,15 +78,91 @@ module SpanEvent = struct
...
-  let to_traceparent t = Printf.sprintf "00-%s-%s-00" t.trace_id t.span_id
+  let char_of_octet f =
+    try "0x" ^ f |> int_of_string |> char_of_int with _ -> '\x00'
 
-  let of_traceparent traceparent =
+  let is_valid t =
+    String.exists (( <> ) '0') t.span_id
+    && String.exists (( <> ) '0') t.trace_id
+
+  let generate_id n = String.init n (fun _ -> "0123456789abcdef".[Random.int 16])
```
psafont, line 99: This seems a pretty inefficient way of generating the ids

This function is used to generate identifiers of 64 bits (16 hexadecimal digits) and 128 bits (32 hexadecimal digits)
I believe using one or two calls to [Random.bits64](https://v2.ocaml.org/api/Random.html#VALbits64) and serializing with hex might be faster. Tee serialization can be done right away or delay it until it's needed, depending on how often they are serialized it might be worth it, memory and time-wise

lindig: I once proposed this. You get one call to `Random.int` per character (4bit) - so indeed this could be more efficient it we need this a lot.

psafont:
CI should pass after merging https://github.com/xapi-project/xs-opam/pull/644 because of the use of functions introduced in ocaml 4.14

lindig (approved):

lindig:
Given that CI has passed, is there a reason not to merge this?

robhoes:
@mg12, could you review this?

robhoes:
There are conflicts to fix now. I'll have to close this. Someone can fix it and create a new PR if this is still useful.

PR #5057, opened 2023-06-09
Fix build: install xapi-tracing

edwintorok:
This is needed by other packages that depend on xapi.


robhoes (approved):

psafont (approved):

edwintorok:
Lets wait for the CI to see whether xapi-tracing installs successfully on its own or its deps need to be updated.

PR #5060, opened 2023-06-13
CP-43565: xapi-expiry-alerts: a new library to generate expiry alerts

gangj:
The interface of this library is "alert", upon calling this API, any
existing outdated messages will be removed first, and new message will
be created only if it does not exist in Xapi.Message records yet.


ocaml/alerts/lib.mli
```diff
@@ -0,0 +1,44 @@
+val seconds_per_day : float
```
psafont: should be removed from the itnerface

psafont (changes requested):
There are too many changes to be able to trust them. In particular do not change the interface to test the certificate expiration and do _not_ remove the test module for certificate alerts. This is independent from testing the library

ocaml/alerts/expiry_alert.mli
```diff
@@ -0,0 +1,57 @@
...
+
+val message_body : string -> Xapi_stdext_date.Date.t -> string
+
+val expired_message : string -> string
+
+val expiring_message : string -> string
+
+val generate_alert :
+     float
+  -> Xapi_stdext_date.Date.t
```
psafont, line 45:
```suggestion
     now:Xapi_stdext_date.Date.t
  -> expiration:Xapi_stdext_date.Date.t
```

ocaml/tests/alerts/test_expiry_alert.ml
```diff
@@ -0,0 +1,407 @@
...
+open Expiry_alert
+
+let date_of = Xapi_stdext_date.Date.of_string
+
+let host_server_cert_expiring_conditions =
+  [
+    (7, Api_messages.host_server_certificate_expiring_07)
+  ; (14, Api_messages.host_server_certificate_expiring_14)
+  ; (30, Api_messages.host_server_certificate_expiring_30)
+  ]
```
psafont, line 24: These belong to the tests for the certificate checks, not to the library tests

ocaml/alerts/expiry_alert.mli
```diff
@@ -0,0 +1,57 @@
...
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+val update :
```
psafont, line 15: These functions need a details explanation of what they do and why they tale all these parameters.

Avoiding complex functions is why I preferred using the functor, the client module does not need to know about complex functions, instead it provides information topt he functor so it can do its job.

ocaml/tests/alerts/test_expiry_alert.ml
gangj: Create a set of test message ids to use for testing, now no message_id from Api_messages is used in the UT.

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,126 @@
...
+  if create_new_msg then
+    let body, (name, priority) = alert in
+    let (_ : [> `message] API.Ref.t) =
+      XenAPI.Message.create ~rpc ~session_id ~name ~priority ~cls:msg_cls
+        ~obj_uuid:msg_obj_uuid ~body
+    in
+    ()
+
+let update ~rpc ~session_id ~alert_obj_description ~expired_message_id
+    ~expiring_conditions ~expiry ~msg_cls ~msg_obj_uuid =
```
minglumlu, line 114: The function doesn't have constraints on the consistency between these parameters. For example, for a function callers' perspective, it is possible to use `expiring_conditions` which is not applicable to `msg_cls.

ocaml/alerts/expiry_alert.mli
```diff
@@ -0,0 +1,57 @@
...
+     rpc:(Rpc.call -> Rpc.response)
+  -> session_id:[< `session] Ref.t
+  -> alert_obj_description:string
+  -> expired_message_id:string * int64
+  -> expiring_conditions:(int * (string * int64)) list
```
minglumlu, line 20: Would suggest to add a comment to explain the meaning of the components in this type.
Does the `string * int64` refers same type of `expired_message_id`?

ocaml/alerts/expiry_alert.ml
liulinC, line 114: I am very confused by the name `msg_obj_uuid`, I thought it was the uuid to identify a message object, which need to be update.

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,126 @@
...
+    alert_obj_description =
+  let remaining_days =
+    days_until_expiry (Date.to_float now) (Date.to_float expiry)
+  in
+  let sorted_alert_conditions =
+    List.sort
+      (fun (remaining_days_a, _) (remaining_days_b, _) ->
+        remaining_days_a - remaining_days_b
+      )
+      ((0, expired_message_id) :: expiring_conditions)
```
minglumlu, line 53: I think this special case could be avoided from beginning. Generally I think if it could be merged with other cases eventually, it should be possible to treat them in unified way in the first place.

ocaml/alerts/expiry_alert.ml
liulinC: These arguments looks over complicated to me, not straight forward for the callers, at least need some document/comment.

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,126 @@
...
+
+let message_body msg expiry =
+  Printf.sprintf "<body><message>%s</message><date>%s</date></body>" msg
+    (Date.to_string expiry)
+
+let expired_message obj = Printf.sprintf "The %s has expired." obj
+
+let expiring_message obj = Printf.sprintf "The %s is expiring soon." obj
+
+let generate_alert now expiry expired_message_id expiring_conditions
```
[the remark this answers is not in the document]
gangj: Adding now as a parameter is to make it easy for UT

ocaml/alerts/expiry_alert.ml
gangj: Yes, it is a little confusing from the name, this parameter is required by the Xapi_message module: https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_message.ml#LL411C45-L411C53

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,126 @@
...
+    alert_obj_description =
+  let remaining_days =
+    days_until_expiry (Date.to_float now) (Date.to_float expiry)
+  in
+  let sorted_alert_conditions =
```
minglumlu, line 48: Something like this would be able to eliminate the `List.sort` and `List.find_opt`:
```
let rec get_most_critical = function
  | [], curr ->
      curr
  | ((days, _) as cond) :: t, None ->
      if days < remaining_days then
        get_most_critical (t, None)
      else
        get_most_critical (t, Some cond)
  | ((days, _) as cond) :: t, (Some (curr_days, _) as curr) ->
      if curr_days > days then
        get_most_critical (t, Some cond)
      else
        get_most_critical (t, curr)
in
```

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,126 @@
...
+  let outdated, current = List.partition is_outdated related_msgs in
+  (outdated, current = [])
+
+let update_message_internal expired_message_id expiring_conditions msg_obj_uuid
+    alert all_msgs =
+  let expiring_msg_name_list =
+    List.map (fun (_, (msg_name, _)) -> msg_name) expiring_conditions
+  in
+  let expired_expiring_msg_name_list =
+    fst expired_message_id :: expiring_msg_name_list
```
minglumlu, line 86: Similarly, this special logic should be avoided.

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,126 @@
...
+  in
+  let outdated_msgs, create_new_msg =
+    messages_require_update related_msgs alert
+  in
+  (outdated_msgs, create_new_msg)
+
+let update_message rpc session_id expired_message_id expiring_conditions msg_cls
+    msg_obj_uuid alert all_msgs =
+  let outdated_msgs, create_new_msg =
+    update_message_internal expired_message_id expiring_conditions msg_obj_uuid
```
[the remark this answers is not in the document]
minglumlu: The journey of `all_msgs` is: `update_message` -> `update_message_internal` -> `related_messages`. No other usages except in `related_messages`. So it could be get just in the last function using it.

ocaml/alerts/expiry_alert.ml
```diff
@@ -110,7 +112,7 @@ let update_message rpc session_id expired_message_id expiring_conditions msg_cls
 
 let update ~rpc ~session_id ~alert_obj_description ~expired_message_id
     ~expiring_conditions ~expiry ~msg_cls ~msg_obj_uuid =
-  let now = Unix.time () in
+  let now = Date.of_float (Unix.time ()) in
```
psafont, line 115:
```suggest
  let now = Date.now () in
```

psafont:
The previous tests have to be recovered and be in place on top of the new tests

ocaml/alerts/expiry_alert.ml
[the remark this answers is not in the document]
liulinC: then it needs to pass `rpc, session_id` as extra argument?

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,126 @@
...
+let update ~rpc ~session_id ~alert_obj_description ~expired_message_id
+    ~expiring_conditions ~expiry ~msg_cls ~msg_obj_uuid =
+  let now = Date.of_float (Unix.time ()) in
+  let alert =
+    generate_alert now expiry expired_message_id expiring_conditions
+      alert_obj_description
+  in
+  match alert with
+  | Some alert ->
+      all_messages rpc session_id
```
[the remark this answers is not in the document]
gangj: Yes, on every attempt to update an alert message, all existing messages need to be retrieved to do the checking.
Do you mean adding another API like "update_messages" for updating more than one messages at one time? Would you please share your thoughts on how to avoid doing this for every API call?

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,126 @@
...
+      Some (message_body expiring expiry, expiring_message_id)
+
+let messages_require_update related_msgs alert =
+  let msg_body, (msg_name, msg_prio) = alert in
+  let is_outdated (_ref, record) =
+    record.API.message_body <> msg_body
+    || record.API.message_name <> msg_name
+    || record.API.message_priority <> msg_prio
+  in
+  let outdated, current = List.partition is_outdated related_msgs in
```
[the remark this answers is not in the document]
gangj: This "outdated" filtering will only work on "related_msgs": which have the matching name(one of expired_name or expiring_name) and matching "message_obj_uuid"

ocaml/alerts/expiry_alert.ml
gangj: `expiring_conditions` describes when an alert message will be sent, it has no relation with `msg_cls`, while `msg_cls` is one of the property needed to create a Xapi_message, there is no constraints between them.

For example:

let test_expiring_07 = ("TEST_EXPIRING_07", 2L)
let test_expiring_14 = ("TEST_EXPIRING_14", 3L)
let test_expiring_30 = ("TEST_EXPIRING_30", 4L)
let test_expiring_conditions =
  [(7, test_expiring_07); (14, test_expiring_14); (30, test_expiring_30)]

this tells the expiry_alert lib to send:
 `"TEST_EXPIRING_07", 2L` when the remaining days less than 7;
`"TEST_EXPIRING_14", 3L` when the remaining days is between 7 and 14;
`"TEST_EXPIRING_30", 4L` when the remaining days is between 14 and 30.

ocaml/alerts/expiry_alert.ml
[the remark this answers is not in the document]
gangj: `all_msgs` is needed as a parameter for `update_message_internal` for UT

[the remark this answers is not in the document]
minglumlu: No. I understand this would be a monad pattern. The `all_msgs` as the start-up input data.

ocaml/alerts/expiry_alert.ml
gangj: With `expired_message_id` as an individual parameter avoid having to check if an expired message id is set in the list, as I think an expired message is required, while expiring messages are optional.

ocaml/alerts/expiry_alert.ml
gangj: > So far this seems a reasonable abstraction. But I'm not sure if this is good for other alerts.

I think now only certificate alert and CSS alert fit to this requirement: message sent to XAPI need to be updated: sent only if not exist, and remove any outdated ones. And there indeed might be other "resource alert", for example, alert for rest amount of disk space, requires the auto-updated requirement.
For your comment, I did tried to add support for the "resource related alert" together with the "expiry alert", but seems it can not be added with little effort: the main reason is that: for the "expiry alert", an "expiry" datetime is needed to generate the "date" field of message_body, while for "resource alert", there will not be the "date" field in the message_body. So I would suggest to only implement for "expiry alert" here, and extend it to support "resource alert" when required later.

gangj:
> The previous tests have to be recovered and be in place on top of the new tests

recovered tests in ocaml/tests/alerts/test_alert_certificate_check.ml.

ocaml/alerts/expiry_alert.mli
gangj: Add type definitions and comments to make it easy to understand in the header file: https://github.com/xapi-project/xen-api/pull/5060/files#diff-318524569501cd50cb93b08d9e5e86f2c11b3d61946f85b2ac38ea65ef68116e

ocaml/alerts/expiry_alert.ml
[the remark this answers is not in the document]
gangj: Changed the API: `alert` to support alerting for a list of messages now.

ocaml/alerts/expiry_alert.mli
gangj: Added comments in the header file: https://github.com/xapi-project/xen-api/pull/5060/files#diff-318524569501cd50cb93b08d9e5e86f2c11b3d61946f85b2ac38ea65ef68116e

> Does the string * int64 refers same type of expired_message_id?
Yes, and `expired_message_id` and `expiring_conditions` are merged into one list now.

ocaml/alerts/expiry_alert.ml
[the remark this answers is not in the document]
gangj: This logic is copied from original code, I agree this might be a problem: there might be more than 1 alert remains in Message storage, do we need to filter to retain only one, or this is by design? @psafont please help to have a look, thanks!

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,168 @@
...
+  (outdated, current = [])
+
+let update_message_internal message_name_list msg_obj_uuid alert all_msgs =
```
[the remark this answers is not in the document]
gangj: Hi @robhoes , I added the func `update_message_internal` for UT: the calling for `XenAPI.Message.destroy` and `XenAPI.Message.create` is isolated. There are some test cases for `update_message_internal` to verify if the `outdated_msgs` and `create_new_msg` are calculated correctly.

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,168 @@
...
+let expired_message obj = Printf.sprintf "The %s has expired." obj
+
+let expiring_message obj = Printf.sprintf "The %s is expiring soon." obj
+
+let generate_alert now obj_description message_sent_on_remaining_days_list
+    expiry =
+  let remaining_days =
+    days_until_expiry (Date.to_float now) (Date.to_float expiry)
+  in
+  let rec get_most_critical = function
```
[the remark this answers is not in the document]
gangj: > If you assume that the list is already sorted, which the certificate one is, and which you can mandate in the interface
Yes, if the list is already sorted, it is easier here. But then I think we will need to add some code in the interface to verify if it is sorted?

ocaml/tests/alerts/test_alert_certificate_check.ml
```diff
@@ -16,30 +16,33 @@ open Certificate_check
...
   ; "20200302T02:00:01Z" (* +30 days, +1 second *)
+  ; "20200302T02:00:00Z" (* +30 days *)
   ]
 
 let expiring_samples =
   [
-    ("20200302T02:00:00Z", Api_messages.host_server_certificate_expiring_30)
+    ("20200302T01:59:59Z", Api_messages.host_server_certificate_expiring_30)
   ; ("20200301T02:00:00Z", Api_messages.host_server_certificate_expiring_30)
-  ; ("20200215T02:00:00Z", Api_messages.host_server_certificate_expiring_14)
```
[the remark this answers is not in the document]
gangj: It is because I see the last test case of `expiring_samples`:
`("20200201T02:00:00Z", Api_messages.host_server_certificate_expiring_07)`
as the check time is `20200201T02:00:00Z`, I think this means the "datetime" get from a certificate:
  let expiry =
    match cert with Host (_, exp) | Internal (_, exp) | CA (_, exp) -> exp
  in
is the last second when the certificate is valid, the certificate will become invalid from the next second of "expiry" (see the test case: `20200201T01:59:59Z` in `expired_samples`).

And I think the length of the "expiring 7 days" period should be exactly 7 days, while for case
`("20200208T02:00:00Z", Api_messages.host_server_certificate_expiring_07)`
the check_time(`20200201T02:00:00Z`) is still in "expiring 7 days" period after 7 days and 1 second, so I changed this test case to:
`("20200208T01:59:59Z", Api_messages.host_server_certificate_expiring_07)`
and added a case:
`("20200208T02:00:00Z", Api_messages.host_server_certificate_expiring_14)`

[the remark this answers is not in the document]
gangj: It is because I see the last test case of `expiring_samples`:
`("20200201T02:00:00Z", Api_messages.host_server_certificate_expiring_07)`
as the check time is `20200201T02:00:00Z`, I think this means the "datetime" get from a certificate:
  let expiry =
    match cert with Host (_, exp) | Internal (_, exp) | CA (_, exp) -> exp
  in
is the last second when the certificate is valid, the certificate will become invalid from the next second of "expiry" (see the test case: `20200201T01:59:59Z` in `expired_samples`).

And I think the length of the "expiring 7 days" period should be exactly 7 days, while for case
`("20200208T02:00:00Z", Api_messages.host_server_certificate_expiring_07)`
the check_time(`20200201T02:00:00Z`) is still in "expiring 7 days" period after 7 days and 1 second, so I changed this test case to:
`("20200208T01:59:59Z", Api_messages.host_server_certificate_expiring_07)`
and added a case:
`("20200208T02:00:00Z", Api_messages.host_server_certificate_expiring_14)`

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,168 @@
...
+  List.filter
+    (fun (_ref, record) ->
+      let related name = List.mem name related_message_name_list in
+      record.API.message_obj_uuid = msg_obj_uuid
+      && related record.API.message_name
+    )
+    all_msgs
+
+let message_body msg expiry =
+  let valid_till = Date.of_float (Date.to_float expiry -. 1.) in
```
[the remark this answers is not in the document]
gangj: It is because the `expiry` time in the `Expiry_alert` module is taken as the first second when something will expire, the last second when it has not expired is 1 second before `expiry`.
While I think what we need to print in the message body in the `date` field for alerting is the last second when it is still valid. So I add back 1 second for it.

ocaml/alerts/certificate/certificate_check.mli
```diff
@@ -0,0 +1,21 @@
+val alert : (Rpc.call -> Rpc.response) -> [< `session] Ref.t -> unit
```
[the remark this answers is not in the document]
gangj: OK, added.

ocaml/alerts/expiry_alert.ml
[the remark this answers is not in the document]
robhoes: It's an internal library used only by our own code, and doesn't take user input, so I think it's fine. Otherwise, just sort the list with e.g. `List.sort (fun (a, _) (b, _) -> compare a b)`.

ocaml/alerts/certificate/certificate_check.ml
```diff
@@ -38,142 +23,73 @@ let get_certificate_attributes rpc session_id =
...
-      in
-      let messages_in_host =
-        List.filter
-          (fun (_, record) -> record.API.message_obj_uuid = obj_uuid)
-          existing_messages
+let get_expiry = function
+  | Host (_, valid_till) | Internal (_, valid_till) | CA (_, valid_till) ->
+      Date.of_float (Date.to_float valid_till +. 1.)
+
+let certificates_to_expiry_message_info_list rpc session_id certificates =
```
minglumlu, line 68: It seems to me that too many duplicated data are placed in the returned list.

gangj: Yes, some fields of the same type of certificate will have the same value: `message_cls`, `obj_description`, `message_sent_on_remaining_days_list`, but the others(`message_obj_uuid`, `expiry`) might be different. Is this acceptable? Otherwise, to reduce this duplication, I think the input parameter of the `Expire_alert.alert` will be more complicated.

ocaml/alerts/certificate/certificate_check.ml
```diff
@@ -38,142 +23,73 @@ let get_certificate_attributes rpc session_id =
...
-  | None ->
-      ()
+      let expiry = get_expiry cert in
+      Expiry_alert.
+        {
+          message_cls
+        ; message_obj_uuid
+        ; obj_description
+        ; message_sent_on_remaining_days_list
+        ; expiry
```
minglumlu, line 85: I would suggest to replace `obj_description`, `message_sent_on_remaining_days_list` and `expiry` with `msg_name_list` in the interface. The purpose is to minimize/simplify the interface.
Leave `generate_alert` to individual callers (The implementation can be provided as a generic helper function). And the `msg_name_list` would be just a `string list`.

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,167 @@
...
+                (fun (_, (msg_name, _)) -> msg_name)
+                message_sent_on_remaining_days_list
+            in
+            Some (alert, msg_name_list, message_cls, message_obj_uuid)
+        | None ->
+            None
+      )
+      expiry_messaging_info_list
+  in
+  if alert_message_info_list <> [] then
```
minglumlu, line 160: I think this is single data flow, the processing can be appended simply. For example:
```
expiry_messaging_info_list
|> List.filter_map (...)
|> List.iter (...)
```

gangj: `all_msgs` needs to be retrieved before List.iter only if List.filter_map returns an non-empty list.

ocaml/alerts/expiry_alert.ml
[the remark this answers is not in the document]
minglumlu: I think sorting the list is not necessary. What we want is something like finding the maximum value.

ocaml/alerts/certificate/certificate_check.ml
gangj: > Leave generate_alert to individual callers

That is another option: do not generate `alert` within the lib, while the caller need to generate the `alert` by itself following some required format. And provide the generated alert to the lib in some way(as an parameter or implemented a `generate_alert` interface of the lib functor)

> The implementation can be provided as a generic helper function

To provide a generic helper function to generate alert, I think this requires these parameter again.

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,167 @@
...
+ *)
+
+module XenAPI = Client.Client
+module Date = Xapi_stdext_date.Date
+
+type message_name_t = string
+
+type message_priority_t = int64
+
+type message_id_t = message_name_t * message_priority_t
```
minglumlu, line 22: Will the `message_priority_t`be used? I can't find its usage in this module.

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,167 @@
...
+      | `PVS_proxy
+      | `Pool
+      | `SR
+      | `VDI
+      | `VM
+      | `VMPP
+      | `VMSS ]
+  ; message_obj_uuid: string
+  ; obj_description: string
+  ; message_sent_on_remaining_days_list: (remaining_days_t * message_id_t) list
```
minglumlu, line 39: How about `priorities`, or `urgency_levels`?

ocaml/alerts/certificate/certificate_check.ml
minglumlu: > To provide a generic helper function to generate alert, I think this requires these parameter again.
No. The helper function will accept much fewer parameters.

minglumlu: And the whole thing would be something like:
```
type alert_msg_t = {
    cls: ...
  ; obj_uuid: string
  ; filter_names: string list
  ; body: string
  ; name: string
  ; prio: int64
}

all_messages rpc session_id 
|> List.iter (fun x ->
    all_msgs
    |> List.filter (fun (_ref, record) ->
        record.API.message_obj_uuid = x.obj_uuid
        && List.mem record.API.message_name x.filter_names
    )
    |> List.partition (fun (_ref, record) ->
        record.API.message_body <> x.body
        || record.API.message_name <> x.name
        || record.API.message_priority <> x.prio
    )
    |> (fun (outdated, current) ->
        List.iter
          (fun (self, _) -> XenAPI.Message.destroy ~rpc ~session_id ~self)
          outdated ;
        match current with
        | [] -> (
            XenAPI.Message.create ~rpc ~session_id ~name:x.name ~priority:x.prio ~cls:x.cls
              ~obj_uuid:x.obj_uuid ~body:x.body
            |> ignore
        )
        | _ ->
            ()
    )
)
```

ocaml/alerts/expiry_alert.ml
gangj: No, I added this type definition just to make the type definition of `message_id_t` clear. I think this is clearer than comment.

ocaml/alerts/expiry_alert.ml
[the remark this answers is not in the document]
gangj: I guess what @robhoes cares most is correctness, performance is the 2nd one.

ocaml/alerts/certificate/certificate_check.ml
gangj: I see the above logic is the logic of `update_message_internal`, and I think the `body`, `name`, `prio` are from the caller self generated alert, and this only processed 1 alert.
Then 
1. How to provide a generic helper function for callers to generate alert?
2. To process more than 1 alert at a time, I think a list of `alert_msg_t ` will be needed, I agree the parameters are a little simpler than the current implementation, while callers need to generate the alert themselves. From the callers' point of view, I think the current implementation is a little easier.

ocaml/alerts/expiry_alert.ml
gangj: Then `all_msgs` will always be retrieved 1 time no matter if there is any alert required to be updated?

ocaml/alerts/expiry_alert.ml
gangj: Do you mean change `remaining_days_t` to `priorities`? I had considered that way. But to let the lib generate the alert for certificates and CSS, a `expiry` datetime is still needed to generate the `message_body`, so I think it will not make it easier and clearer.

ocaml/alerts/expiry_alert.ml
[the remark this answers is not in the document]
gangj: Changed back to use `List.sort` and `List.find_opt` instead of self-written func: `get_most_critical`.

ocaml/alerts/expiry_alert.ml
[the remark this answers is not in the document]
gangj: Changed to avoid the "-1 and +1" in [5da53be](https://github.com/xapi-project/xen-api/pull/5060/commits/5da53be8c5a09d72ea915f758fe14efdad404edf)

ocaml/alerts/expiry_alert.mli
```diff
@@ -0,0 +1,85 @@
...
+      | `SR
+      | `VDI
+      | `VM
+      | `VMPP
+      | `VMSS ]
+        (* parameter "cls" of XenAPI.Message.create *)
+  ; obj_uuid: string (* parameter "obj_uuid" of XenAPI.Message.create *)
+  ; obj_description: string
+        (* description of the obj which would expire, which will be used in the body of generated message, take host server certificate for example: "TLS server certificate" *)
+  ; msg_sent_on_remaining_days_list: (remaining_days_t * message_id_t) list
```
minglumlu, line 39: How about something like `alert_conditions`, or `alert_priorities`?

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,129 @@
...
+  in
+  match critical_condition with
+  | None ->
+      None
+  | Some (days, expired_message_id) when days <= 0 ->
+      let expired = expired_message obj_description in
+      Some (message_body expired expiry, expired_message_id)
+  | Some (_, expiring_message_id) ->
+      let expiring = expiring_message obj_description in
+      Some (message_body expiring expiry, expiring_message_id)
```
minglumlu, line 79:
```suggestion
msg_sent_on_remaining_days_list
|> List.sort (fun (a, _) (b, _) -> compare a b)
|> List.find_opt (fun (days, _) -> remaining_days < float_of_int days)
|> function
| None ->
    None
| Some (days, expired_message_id) when days <= 0 ->
    let expired = expired_message obj_description in
    Some (message_body expired expiry, expired_message_id)
| Some (_, expiring_message_id) ->
    let expiring = expiring_message obj_description in
    Some (message_body expiring expiry, expiring_message_id)
```

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,129 @@
...
+
+let update_message_internal msg_name_list msg_obj_uuid alert all_msgs =
```
minglumlu, line 81: How about naming it to something like `find_alert_msgs`?

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,129 @@
...
+         ; msg_sent_on_remaining_days_list
+         ; expiry
+         } ->
+      let alert =
+        generate_alert now obj_description msg_sent_on_remaining_days_list
+          expiry
+      in
+      match alert with
+      | Some alert ->
+          let msg_name_list =
```
minglumlu, line 111: I would suggest to move this block into `update_message_internal`.

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,129 @@
...
+          List.iter
+            (fun (self, _) -> XenAPI.Message.destroy ~rpc ~session_id ~self)
+            outdated ;
+          if current = [] then
+            let body, (name, priority) = alert in
+            XenAPI.Message.create ~rpc ~session_id ~name ~priority ~cls
+              ~obj_uuid ~body
+            |> ignore
+      | None ->
+          ()
```
minglumlu, line 127:
```suggestion
         ; alert_priorities
         ; expiry
         } ->
      generate_alert now obj_description alert_priorities expiry
      |> Option.map (fun alert ->
          all_msgs
          |> filter_alert_msgs alert_priorities obj_uuid alert
          |> (fun (outdated, current) ->
              List.iter
                (fun (self, _) ->
                  XenAPI.Message.destroy ~rpc ~session_id ~self
                )
                outdated ;
              if current = [] then
                let body, (name, priority) = alert in
                XenAPI.Message.create ~rpc ~session_id ~name ~priority ~cls
                  ~obj_uuid ~body
          )
      )
      |> ignore
```

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,129 @@
...
+         record.API.message_obj_uuid = msg_obj_uuid
+         && List.mem record.API.message_name msg_name_list
+     )
+  |> List.partition (fun (_ref, record) ->
+         record.API.message_body <> msg_body
+         || record.API.message_name <> msg_name
+         || record.API.message_priority <> msg_prio
+     )
+
+let alert ~rpc ~session_id expiry_messaging_info_list =
```
minglumlu, line 94: I understand the `expiry_messaging_info_list` is something like `raw_alerts`.

ocaml/alerts/certificate/certificate_check.ml
minglumlu: I think the function `alert` now plays two parts: `generate alert` and `update existing alert messages`.
Combining them in one function would result in verbose input parameters as it has to be a union of two inputs.

minglumlu: But I don't have better idea except splitting them which may get the usage of the lib complicated, for example, calling two lib functions instead of only one.

ocaml/alerts/expiry_alert.ml
minglumlu: I meant to rename the `message_sent_on_remaining_days_list`. It's too long.

ocaml/alerts/expiry_alert.ml
gangj: renamed `update_message_internal` to `filter_messages`, as I think it only filters all existing messages, it is not "alert" related yet.

ocaml/alerts/expiry_alert.mli
```diff
@@ -0,0 +1,85 @@
...
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+type message_name_t = string
```
lindig, line 15: Why is every type suffixed with a `_t`? This is not idiomatic for OCaml where types and values live in separate name spaces and can't conflict.

ocaml/alerts/expiry_alert.mli
```diff
@@ -0,0 +1,85 @@
...
+
+type message_priority_t = int64
+
+(* message_id_t is the type of mesage defined in Api_messages *)
+type message_id_t = message_name_t * message_priority_t
+
+type remaining_days_t = int
+
+type raw_alert_t = {
+    cls:
```
lindig, line 25: This is a polymorphic variant and hence extensible. Is this flexibility required and could this be a regular variant instead?

ocaml/alerts/expiry_alert.mli
```diff
@@ -0,0 +1,85 @@
...
+val expiring_message : string -> string
+
+val generate_alert :
+     Xapi_stdext_date.Date.t
+  -> string
+  -> (remaining_days_t * message_id_t) list
+  -> Xapi_stdext_date.Date.t
+  -> (message_name_t * message_id_t) option
+
+val filter_messages :
```
lindig, line 80: This needs documentation how this works.

ocaml/alerts/expiry_alert.mli
```diff
@@ -0,0 +1,85 @@
...
+
+(* Below exposed only for ease of testing *)
+
+val message_body : string -> Xapi_stdext_date.Date.t -> string
+
+val expired_message : string -> string
+
+val expiring_message : string -> string
+
+val generate_alert :
```
lindig, line 73: Why is the result an option? This needs to be documented because the name suggests an alert is generated every time.

ocaml/alerts/expiry_alert.ml
```diff
@@ -0,0 +1,112 @@
...
+  | None ->
+      None
+  | Some (days, expired_message_id) when days <= 0 ->
+      let expired = expired_message obj_description in
+      Some (message_body expired expiry, expired_message_id)
+  | Some (_, expiring_message_id) ->
+      let expiring = expiring_message obj_description in
+      Some (message_body expiring expiry, expiring_message_id)
+
+let filter_messages msg_name_list msg_obj_uuid alert all_msgs =
```
lindig, line 76: This is undocumented both here and in the MLI.

lindig (approved):

ocaml/alerts/expiry_alert.mli
minglumlu: How about `maybe_generate_alert`?

ocaml/alerts/expiry_alert.mli
minglumlu: I think this is aligned with https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_message.ml#L413, which is defined in https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi-cli-server/record_util.ml#L28

ocaml/alerts/expiry_alert.mli
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+type message_name_t = string
+
+type message_priority_t = int64
```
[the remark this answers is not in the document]
gangj: The `message_name` and `message_priority` are those defined in https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi-consts/api_messages.ml, and the `priority` there is 64 bits: https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi-consts/api_messages.ml#L34

ocaml/alerts/expiry_alert.ml
gangj: Added comment for this function.

ocaml/alerts/expiry_alert.mli
gangj: Added

minglumlu (approved):

robhoes (approved):

robhoes:
I think this looks good now, a lot easier to follow than the first iteration. I also believe that Pau's concerns about testing have been addressed. I'll squash and merge.

PR #5062, opened 2023-06-14
CA-378304: check max_file_size limit after writing to tracing file

mg12:
- fix issue in which max_file_size limit wasn't being respected after batching was added
- simplify the code to remove the need for a fd ref and make sure all fds have closed


robhoes (approved):
Smaller and better.

psafont (approved):

PR #5063, opened 2023-06-15
CP-43551: Dump host_rrd latest data to /dev/shm/metrics/host-dss

DeliZhangX:
- Dump all latest data of host dss to file in json format so that any client can read even if it's non-privileged user, such as NRPE.

Test passed, see CP-43551 to get details.

### DIFF
--- a/ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
+++ b/ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
@@ -456,6 +456,41 @@ let query_host_ds (ds_name : string) : float =
           query ds_name rrdi
   )
 
+(** Dump all latest data of host dss to file in json format so that any client
+    can read even if it's non-privileged user, such as NRPE. 
+    Especially, nan, infinity and neg_infinity will be converted to strings 
+    "NaN", "infinity" and "-infinity", the client needs to handle by itself.
+    *)
+let dump_host_dss_to_file (file : string) : unit =
+  let convert_value x =
+    match classify_float x with
+    | FP_nan ->
+        `String "NaN"
+    | FP_infinite ->
+        `String (if x > 0.0 then "infinity" else "-infinity")
+    | _ ->
+        `Float x
+  in
+  let json =
+    with_lock mutex (fun () ->
+        match !host_rrd with
+        | None ->
+            `Assoc []
+        | Some rrdi ->
+            `Assoc
+              (Rrd.ds_names rrdi.rrd
+              |> List.map (fun ds_name ->
+                     (ds_name, convert_value (query ds_name rrdi))
+                 )
+              )
+    )
+  in
+  Xapi_stdext_unix.Unixext.atomic_write_to_file file 0o644 (fun fd ->
+      let oc = Unix.out_channel_of_descr fd in
+      Yojson.Basic.to_channel ~std:true oc json ;
+      flush oc
+  )
+
 (** {add_vm_ds vm_uuid domid ds_name} enables collection of the data produced by
     the data sourced with name {ds_name} for the VM {vm_uuid} into a time series
     (rrd). Throws an exception if there are no time series related to {vm_uuid}
--- a/ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
+++ b/ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
@@ -762,7 +762,11 @@ let do_monitor_write xc writers =
       let stats = List.rev_append plugins_stats dom0_stats in
       Rrdd_stats.print_snapshot () ;
       let uuid_domids = List.map (fun (_, u, i) -> (u, i)) domains in
-      Rrdd_monitor.update_rrds timestamp stats uuid_domids my_paused_vms
+      Rrdd_monitor.update_rrds timestamp stats uuid_domids my_paused_vms ;
+
+      (* Dump to /dev/shm/metrics/host-dss *)
+      Rrdd_server.Plugin.get_path "host-dss"
+      |> Rrdd_server.dump_host_dss_to_file
   )
 
 let monitor_write_loop writers =
--- a/ocaml/xcp-rrdd/bin/rrdd/rrdd_server.mli
+++ b/ocaml/xcp-rrdd/bin/rrdd/rrdd_server.mli
@@ -28,6 +28,8 @@ val query_possible_host_dss : unit -> Data_source.t list
 
 val query_host_ds : string -> float
 
+val dump_host_dss_to_file : string -> unit
+
 val add_vm_ds : string -> int -> string -> unit
 
 val forget_vm_ds : string -> string -> unit


ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
```diff
@@ -456,6 +457,30 @@ let query_host_ds (ds_name : string) : float =
...
+        | Some rrdi ->
+            `Assoc
+              (Rrd.ds_names rrdi.rrd
+              |> List.map (fun ds_name -> (ds_name, `Float (query ds_name rrdi)))
+              )
+    )
+    |> Yojson.Basic.to_string
+    |> Bytes.of_string
+  in
+  let fd =
```
lindig, line 477: `Unixext.write_string_to_file` could be used. But would be better to use the Yojson functions that write to a channel rather than creating a potentially large string first.

lindig (changes requested):

ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
```diff
@@ -775,6 +775,11 @@ let monitor_write_loop writers =
               with_lock Rrdd_shared.last_loop_end_time_m (fun _ ->
                   Rrdd_shared.last_loop_end_time := Unix.gettimeofday ()
               ) ;
+
+              (* Dump to /dev/shm/metrics/host-dss *)
+              let path = Rrdd_server.Plugin.get_path "host-dss" in
+              Rrdd_server.dump_host_dss_to_file path ;
```
psafont, line 781: I would prefer if this was done inside the do_monitor_write timed section, after the update_rrds is executed

DeliZhangX: OK will move there.

ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
```diff
@@ -456,6 +457,30 @@ let query_host_ds (ds_name : string) : float =
...
+    can read even if it's non-privileged user, such as NRPE. *)
+let dump_host_dss_to_file (file : string) : unit =
+  let json =
+    with_lock mutex (fun () ->
+        match !host_rrd with
+        | None ->
+            `Assoc []
+        | Some rrdi ->
+            `Assoc
+              (Rrd.ds_names rrdi.rrd
```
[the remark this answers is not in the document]
DeliZhangX: I wrote test code to dump `stats` in `do_monitor_write`, but found some data is only the raw data than calculated as Data Source Type.
e.g. below is CPU time
```
[root@xrtmia-11-07 ~]# json_reformat< /dev/shm/metrics/host-dss1 | grep cpu_avg
    "cpu_avg": 34199.011874511874,
```
while the queried data from rrd is correct CPU use percentage
```
[root@xrtmia-11-07 ~]# json_reformat< /dev/shm/metrics/host-dss | grep cpu_avg
    "cpu_avg": 0.5052633881568909,
```
So looks still need to query from rrd even if there are about 100~200 metrics query every 5 seconds.

ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
DeliZhangX: Thanks! fixed now.

ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
DeliZhangX: Fixed now.

ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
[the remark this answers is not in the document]
robhoes: Or this thing that collects the data (e.g. Nagios or its plugin) processes the raw data directly?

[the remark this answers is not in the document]
psafont: Ah, that makes sense, the derivation step on the data needs to happen in some sort of way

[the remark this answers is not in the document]
DeliZhangX: Nagios plugin can processes the raw data indeed, while that's duplicated with rrd, and it has to query 2 times with 5 seconds  interval that will slow down Nagios check obviously. 
If you concern the query count very much, I think we can add `filter` to let only specified `ds_name`(s) data are queried. So in future any people want to dump some data they should update the filter firstly.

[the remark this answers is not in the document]
robhoes: I don't know about twice in a 5 sec interval. It just needs to take regular samples at some interval and calculate whatever it needs? Doesn't Nagios actually _want_ the raw data, before averages/min/max etc are taken, so that it can do that based on it's own time scales?

ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
```diff
@@ -456,6 +457,26 @@ let query_host_ds (ds_name : string) : float =
...
+              |> List.map (fun ds_name -> (ds_name, `Float (query ds_name rrdi)))
+              )
+    )
+  in
+  let oc = open_out_gen [Open_wronly; Open_creat; Open_trunc] 0o644 file in
+  finally
+    (fun () -> Yojson.Basic.pretty_to_channel oc json)
+    (fun () -> close_out oc)
```
psafont, line 478: It's fine to not call `close_out_noerr` because our finally deals with exception raised here

ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
```diff
@@ -456,6 +457,26 @@ let query_host_ds (ds_name : string) : float =
...
+        | Some rrdi ->
+            `Assoc
+              (Rrd.ds_names rrdi.rrd
+              |> List.map (fun ds_name -> (ds_name, `Float (query ds_name rrdi)))
+              )
+    )
+  in
+  let oc = open_out_gen [Open_wronly; Open_creat; Open_trunc] 0o644 file in
+  finally
+    (fun () -> Yojson.Basic.pretty_to_channel oc json)
```
psafont, line 477: We probably want a compact representation after the testing is done:
```suggestion
    (fun () -> Yojson.Basic.to_channel oc json)
```

ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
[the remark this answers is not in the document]
DeliZhangX: Nagios NRPE won't collect and save data (like rrd), it just handle a check calling from client and execute corresponding check command, e.g.  `command[check_cpu]=/usr/lib64/nagios/plugins/check_cpu -w 80% -c 90%` , `check_cpu` internally run dom0 cmd `mpstat` to get current cpu usage, and finally return to client.
This process is similar to user executes OS command through SSH.
If NRPE reads the raw data from rrd dump file, then `check_cpu` has to read the file, sleep 5s, read again, calculate the result by self. This duplicates with rrd, and more bad thing is the client will have to wait 5s to get check result.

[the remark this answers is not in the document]
psafont: I agree it's not worth duplicating the calculation inside dom0, so I think it's fine as it is.

As a note, for prometheus / openmetrics the values are exported as raw values, the appliance gathering metrics deal with these

ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
DeliZhangX: I saw `query ds_name rrdi` already ensure return `float` type value, so I still need to handle NaN and infinities cases for Yojson?

psafont: Unfortunately, yes. Json doesn't allow NaN not infinities to be represented in floats, but NaN in this case must be considered a common value. I don't think infinities appear hear, but they are a possibility.

https://datatracker.ietf.org/doc/html/rfc7159.html#section-6
> Numeric values that cannot be represented in the grammar below (such
>   as Infinity and NaN) are not permitted.

>      number = [ minus ] int [ frac ] [ exp ]
>      decimal-point = %x2E       ; .
>      digit1-9 = %x31-39         ; 1-9
>      e = %x65 / %x45            ; e E
>      exp = e [ minus / plus ] 1*DIGIT
>      frac = decimal-point 1*DIGIT
>      int = zero / ( digit1-9 *DIGIT )
>      minus = %x2D               ; -
>      plus = %x2B                ; +
>      zero = %x30                ; 0

>   This specification allows implementations to set limits on the range
>   and precision of numbers accepted.  Since software that implements
>   IEEE 754-2008 binary64 (double precision) numbers [[IEEE754](https://datatracker.ietf.org/doc/html/rfc7159.html#ref-IEEE754)] is
>   generally available and widely used, good interoperability can be
>   achieved by implementations that expect no more precision or range
>   than these provide, in the sense that implementations will
>   approximate JSON numbers within the expected precision.  A JSON
>   number such as 1E400 or 3.141592653589793238462643383279 may indicate
>   potential interoperability problems, since it suggests that the
>   software that created it expects receiving software to have greater
>   capabilities for numeric magnitude and precision than is widely
>   available.

>   Note that when such software is used, numbers that are integers and
>   are in the range [-(2**53)+1, (2**53)-1] are interoperable in the
>   sense that implementations will agree exactly on their numeric
>   values.

I believe some sort of agreement between writer and reader must be done in order to parse these 3 values (NaN, Infinity, -Infinity)

ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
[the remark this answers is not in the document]
robhoes: OK that's fine then.

ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
psafont: When using std=true whenever NaN or the infinities are present the library throws an exception. They need some treatment in the representation before they can be sent. This is what I meant by
> some sort of agreement between writer and reader must be done in order to parse these 3 values 
The representation must be agreed, documented in the design, then implemented in both sides

ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
```diff
@@ -762,7 +762,11 @@ let do_monitor_write xc writers =
       let stats = List.rev_append plugins_stats dom0_stats in
       Rrdd_stats.print_snapshot () ;
       let uuid_domids = List.map (fun (_, u, i) -> (u, i)) domains in
-      Rrdd_monitor.update_rrds timestamp stats uuid_domids my_paused_vms
+      Rrdd_monitor.update_rrds timestamp stats uuid_domids my_paused_vms ;
+
+      (* Dump to /dev/shm/metrics/host-dss *)
+      Rrdd_server.Plugin.get_path "host-dss"
+      |> Rrdd_server.dump_host_dss_to_file
```
minglumlu, line 769: Is it possible for other readers (like NRPE) conflicting with rrdd on accessing the file, like reading corrupted content?

DeliZhangX: The reader takes responsibility to retry. I have updated design https://info.citrite.net/display/xenserver/Design+of+NRPE+integration

ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
DeliZhangX: Fixed now, could you continue review?
I also updated design https://info.citrite.net/display/xenserver/Design+of+NRPE+integration and commit description.

ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
minglumlu: Then how could a reader know that corrupted data is being read?

ocaml/xcp-rrdd/bin/rrdd/rrdd_server.ml
minglumlu: Is it worth to add some unit tests for these agreements?

DeliZhangX: This is simple and not critical function, just let me know if UT is must have.

ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
DeliZhangX: I will change to that.

DeliZhangX: Fixed now by referring `Unixext.atomic_write_to_file`

minglumlu: I seems you can use `Unixext.atomic_write_to_file` directly.
The `fd` passed from it can be changed to channel by `Unix.out_channel_of_descr`.
https://v2.ocaml.org/api/Unix.html

DeliZhangX: It's wonderful!, fixed now!

DeliZhangX:
Python test failed due to
```
  Warning: The support for python 2.7 will be removed on June 19. Related issue: https://github.com/actions/setup-python/issues/672
  Version 2.7 was not found in the local cache
  Error: The version '2.7' with architecture 'x64' was not found for Ubuntu 20.04.
```

minglumlu (approved):

robhoes (approved):

PR #5064, opened 2023-06-15
CA-378035: set nbd client timeout to 60 seconds

MarkSymsCtx:


lindig (approved):

robhoes (approved):
Look good, and as discussed. However, a bit of explanation in the commit message would be helpful.

PR #5065, opened 2023-06-15
CA-378455: Ensure TPM contents are base64-encoded on migration recieve

psafont:
When migrating from earlier xapi releases, the migration block was not base64-encoded while newer versions assume the migration block is. Introduce a new migration block so from this version can tell when to base64-encode the contents for migration contents and when to leave the contents alone.

Drafting until the fix is manually verified

### DIFF
--- a/ocaml/xenopsd/lib/suspend_image.ml
+++ b/ocaml/xenopsd/lib/suspend_image.ml
@@ -42,7 +42,7 @@ module Xenops_record = struct
   [@@deriving sexp]
 
   let make ?vm_str ?xs_subtree () =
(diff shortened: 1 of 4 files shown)


edwintorok (approved):
In theory this looks good. Does it work?

ocaml/xenopsd/lib/suspend_image.ml
```diff
@@ -62,7 +62,8 @@ type header_type =
   | Qemu_xen
   | Demu
   | Varstored
-  | Swtpm
+  | Swtpm_old
```
lindig, line 65: I would use something like `Swtpm_pre_<version>` or `Swtpm0` - just in case you need another increment.

lindig (approved):

PR #5067, opened 2023-06-16
CA-378317: fix EBADF in waitpid_nohang (v1)

edwintorok:
'waitpid' will close the socket when succesful,
so move the 'clear_nonblock' to the EAGAIN|EWOULDBLOCK case where the socket should still be open.

Fixes: 291c5002f ("CA-341921: avoid EINVAL errors for >1024 FDs")

### DIFF
--- a/ocaml/forkexecd/lib/forkhelpers.ml
+++ b/ocaml/forkexecd/lib/forkhelpers.ml
@@ -62,12 +62,9 @@ let waitpid (sock, pid) =
 
 let waitpid_nohang ((sock, _) as x) =
   Unix.set_nonblock sock ;
-  let r =
-    try waitpid x
-    with Unix.(Unix_error ((EAGAIN | EWOULDBLOCK), _, _)) ->
-      (0, Unix.WEXITED 0)
-  in
-  Unix.clear_nonblock sock ; r
+  try waitpid x
+  with Unix.(Unix_error ((EAGAIN | EWOULDBLOCK), _, _)) ->
+    Unix.clear_nonblock sock ; (0, Unix.WEXITED 0)
 
 let dontwaitpid (sock, _pid) =
   ( try


psafont (approved):
This seems to work, but it might be better for clarity to have an internal waitpid function that doesn't close the socket, and have both public functions to close it explicitly so the socket management is in these external functions

edwintorok:
If the process is still running then 'waitpid_nohang' would have to return immediately and leave the socket open. 
If the process is already dead then it would close it, so the same function has different behaviour.
If we separate it then we might introduce leaks.

I'd like to completely remove the open socket from here, and e.g. establish a fresh one to query forkexecd about the status of the process instead of keeping one open, but that has other drawbacks (it prevents FD leaks in the client, but may leave an FD/memory leak in forkexecd if we never 'wait' for a process that exited), I'll need to think about this more to see whether I find an alternative.

lindig (commented):
I think this code is quite subtle in its interaction with `waitpid`. It might be better to implement this my inlining the `waitpid` call.

The current patch calls `Unix.clear_nonblock sock` when `waitpid` has failed before reaching the `Unix.close sock` call. I think we need a comment in `waitpid` that ` Fecomms.read_raw_rpc`
might fail.

Where is the socket closed? So we rely on `waitpid` being called again?

ocaml/forkexecd/lib/forkhelpers.ml
```diff
@@ -62,12 +62,9 @@ let waitpid (sock, pid) =
...
   Unix.set_nonblock sock ;
-  let r =
-    try waitpid x
-    with Unix.(Unix_error ((EAGAIN | EWOULDBLOCK), _, _)) ->
-      (0, Unix.WEXITED 0)
-  in
-  Unix.clear_nonblock sock ; r
+  try waitpid x
+  with Unix.(Unix_error ((EAGAIN | EWOULDBLOCK), _, _)) ->
+    Unix.clear_nonblock sock ; (0, Unix.WEXITED 0)
```
lindig, line 67: This signals that the the child terminated. But don't we have to close the socket here? This is the behavior of `waitpid`. Or where and when would the socket be closed?

lindig:
I think it would help to implement `waitpid_nohang` explicitly. I generally don't like the current protocol how the socket is closed.

```

ocaml/forkexecd/lib/forkhelpers.ml
edwintorok: I think we may have more leaks in the callers, because the callers are not consistent how they use it:
stunnel calls this, and if the pid is still alive then sends a SIGKILL and then leaks the socket.
OTOH the xapi host backup code calls waitpid_nohang in a loop with a delay inbetween, so in that case if the process is still alive we must *not* close the socket.

It might be better to replace this function with 2:
* one which always explicitly closes the socket no matter what, and doesn't get blocked if the PID is still running, e.g. 'waitpid_nohang_close'
* one which leaves the socket open so it can be called in a loop, e.g. 'waitpid_nohang_leaveopen'

Could also try to wrap it with my safe resources code, but that will only detect leaks at runtime, not compile time.

PR #5068, opened 2023-06-22
Filter out link local address when getting host IPv6

benjamreis:
The link local addresses can't be used

### DIFF
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -2527,7 +2527,12 @@ let migrate_receive ~__context ~host ~network ~options:_ =
         let configuration_mode =
           Db.PIF.get_ipv6_configuration_mode ~__context ~self:pif
         in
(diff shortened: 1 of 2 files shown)


lindig (changes requested):
I think this filter should be factored out into one of the helper modules and not repeated.

benjamreis:
So if I mutualize and use this lib: https://opam.ocaml.org/packages/ipaddr/

Would that be okay?
We can then extend bit by bit the use of the lib in other cases.

lindig:
Yes, using such a library to encapsulate the address would be a good step. I don't know how feasible that is, though.

psafont:
I believe this won't be merged as-is. It's better to open a new one once the ipaddr library is used. Note that the ipaddr version used has bugs with certain ranges of IPv6 addresses. I'm working to include a fixed version.

benjamreis:
New version using the `Ipaddr` lib as asked here: https://github.com/xapi-project/xen-api/pull/5463

PR #5069, opened 2023-06-22
Add HTTP Strict Transport Security header

gthvn1:
This header informs the browser that it should use HTTPS. The max-age is set to 2 years.
By default header is not added. To enable it you need to set option add_hsts_response_header to true.


lindig:
Thanks for your contribution. Please rebase to get a working CI. I generally like this. How common is the name `hsts`? For me it's quite cryptic and I wonder if a more descriptive name would be better.

ocaml/libs/http-lib/http_svr.ml
```diff
@@ -221,17 +221,23 @@ let response_redirect ?req s dest =
...
   let size = (Unix.LargeFile.stat file).Unix.LargeFile.st_size in
   let mime_header =
     Option.fold ~none:[]
       ~some:(fun ty -> [(Hdr.content_type, ty)])
       mime_content_type
   in
   let keep_alive = (Http.Hdr.connection, "keep-alive") in
+  let hsts =
+    if add_hsts then
+      [(Http.Hdr.hsts, "max-age=63072000")] (* max-age sets to 2yr *)
```
lindig, line 234: Would be better to compute this as in `Printf.sprintf "max-age=%d" (60*60*24*365*2)` or similar to avoid magic constants.

ocaml/libs/http-lib/http_svr.ml
```diff
@@ -221,17 +221,23 @@ let response_redirect ?req s dest =
   in
   Unixext.really_write_string s (Http.Response.to_wire_string res)
 
-let response_file ?mime_content_type s file =
+let response_file ?mime_content_type ~add_hsts s file =
```
lindig, line 224: Would it not be better to use an option that includes the value actually being used? So either `None` or `Some time`? The time to use could still come from a configuration.

lindig (approved):

gthvn1:
> Thanks for your contribution. Please rebase to get a working CI. I generally like this. How common is the name `hsts`? For me it's quite cryptic and I wonder if a more descriptive name would be better.

Thanks for the feedback. Yes HSTS (HTTP Strict Transport Security) is quiet common and is coming from  RFC 6797. I think it is ok but sure I can use a more descriptive name. Maybe **http_sts** ?

ocaml/libs/http-lib/http_svr.ml
gthvn1: Right. Common values are 1yr and 2yr but it is a good idea to have something that is configurable. If I understand correctly I could make it optional and change the configuration to be **hsts_max_age** and set it to 0 by default. And it also remove the magic constant ;-)... I will do that. Thanks for the comment.

gthvn1: Note that set it to 0 by default is to be able to not activate this feature. So if it is 0 I will not pass the value to response_file otherwise I will pass it. We want to disable it by default because HSTS prevent you to use self signed certificate and it is common to have self signed certificate. So only people that know what they are doing (and so have a valid configuration) should enable it.

lindig: This is my idea, roughly. I did not implement the other comment about magic numbers,

```
diff --git a/ocaml/libs/http-lib/http_svr.ml b/ocaml/libs/http-lib/http_svr.ml
index ea35e02ee..b4ef6600c 100644
--- a/ocaml/libs/http-lib/http_svr.ml
+++ b/ocaml/libs/http-lib/http_svr.ml
@@ -221,17 +221,22 @@ let response_redirect ?req s dest =
   in
   Unixext.really_write_string s (Http.Response.to_wire_string res)
 
-let response_file ?mime_content_type s file =
+let response_file ?mime_content_type ?hsts s file =
   let size = (Unix.LargeFile.stat file).Unix.LargeFile.st_size in
   let mime_header =
     Option.fold ~none:[]
       ~some:(fun ty -> [(Hdr.content_type, ty)])
       mime_content_type
   in
-  let keep_alive = (Http.Hdr.connection, "keep-alive") in
+  let keep_alive = [(Http.Hdr.connection, "keep-alive")] in
+  let hsts =
+    Option.fold ~none:[]
+      ~some:(fun time -> [(Http.Hdr.hsts, "max-age=63072000")])
+      hsts
+  in
   let res =
-    Http.Response.make ~version:"1.1"
-      ~headers:(keep_alive :: mime_header)
+    Http.Response.make ~version:"1.1" ~headers:List.concat
+      [keep_alive; hsts; mime_header]
       ~length:size "200" "OK"
   in
   Unixext.with_file file [Unix.O_RDONLY] 0 (fun f ->
diff --git a/ocaml/libs/http-lib/http_svr.mli b/ocaml/libs/http-lib/http_svr.mli
index d20ddfc38..25a12c4f6 100644
--- a/ocaml/libs/http-lib/http_svr.mli
+++ b/ocaml/libs/http-lib/http_svr.mli
@@ -120,7 +120,7 @@ val response_method_not_implemented :
 val response_redirect : ?req:Http.Request.t -> Unix.file_descr -> string -> unit
 
 val response_file :
-  ?mime_content_type:string -> Unix.file_descr -> string -> unit
+  ?mime_content_type:string -> ?hsts:int -> Unix.file_descr -> string -> unit
 
 val respond_to_options : Http.Request.t -> Unix.file_descr -> unit
 
```

lindig (commented):
I think we now have a mix of conventions: positive/negative coming from the configuration and Some/None used in the function using it. That forces conversion and branching in several places. While I don't like the positive/negative convention too much it is simple. It's probably easiest to make the `hsts_directive` no longer optional and use the convention that a negative value will mean the header is not set.

ocaml/xapi/xapi_vncsnapshot.ml
```diff
@@ -46,8 +46,12 @@ let vncsnapshot_handler (req : Request.t) s _ =
                 ; tmp
                 ]
             in
+            let max_age = !Xapi_globs.hsts_max_age in
             waitpid_fail_if_bad_exit pid ;
-            Http_svr.response_file s tmp
+            if max_age < 0 then
+              Http_svr.response_file s tmp
+            else
+              Http_svr.response_file ~hsts_directive:max_age s tmp
```
gthvn1, line 54: Not sure that it makes sense to add hsts directive here.... any ideas?

gthvn1:
> I think we now have a mix of conventions: positive/negative coming from the configuration and Some/None used in the function using it. That forces conversion and branching in several places. While I don't like the positive/negative convention too much it is simple. It's probably easiest to make the `hsts_directive` no longer optional and use the convention that a negative value will mean the header is not set.

I agree with with `hsts_directive` no longer optional and use negative value to set or not the header. For information the value 0 has a meaning with HSTS. It says to the client (mainly a browser) that it should stop considering the host as a Known HSTS Host. The only question I have is: does HSTS have a meaning for the `vnc_snapshot_handler`. If not this is a sopt where it makes sense to have `hsts_directive` as an option. Otherwise option is not needed. And even for `vnc_snapshot_handler` it will just be ignored so we can always pass it.

 I will fix that and declare `hsts_directive` as a labeled argument instead of an option.

robhoes (approved):

PR #5070, opened 2023-06-22
CA-378837 log results from Host.get_vms_which_prevent_evacuation

lindig:
Host.get_vms_which_prevent_evacuation is used by XenCenter as part of a rolling pool upgrade. But when it fails we are missing information in the logs about VMs that block evacuation of a host. Add logging.

### DIFF
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -448,7 +448,16 @@ let get_vms_which_prevent_evacuation_internal ~__context ~self ~ignore_ha =
 
 (* New Orlando style function which returns a Map *)
 let get_vms_which_prevent_evacuation ~__context ~self =
-  get_vms_which_prevent_evacuation_internal ~__context ~self ~ignore_ha:false
+  let vms =
+    get_vms_which_prevent_evacuation_internal ~__context ~self ~ignore_ha:false
+  in
+
+  let log (vm, reasons) =
+    debug "%s: VM %s preventing evacuation of host %s: %s" __FUNCTION__
+      (Ref.string_of vm) (Ref.string_of self)
+      (String.concat "; " reasons)
+  in
+  List.iter log vms ; vms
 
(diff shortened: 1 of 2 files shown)


ocaml/xapi/xapi_host.ml
```diff
@@ -448,7 +448,16 @@ let get_vms_which_prevent_evacuation_internal ~__context ~self ~ignore_ha =
...
 (* New Orlando style function which returns a Map *)
 let get_vms_which_prevent_evacuation ~__context ~self =
-  get_vms_which_prevent_evacuation_internal ~__context ~self ~ignore_ha:false
+  let vms =
+    get_vms_which_prevent_evacuation_internal ~__context ~self ~ignore_ha:false
+  in
+
+  let log (vm, reasons) =
+    debug "%s: VM %s preventing evacuation of host %s: %s" __FUNCTION__
+      (Ref.string_of vm) (Ref.string_of self)
```
edwintorok, line 457: These will all be database references, making it hard to debug, but better than nothing.

edwintorok (approved):

edwintorok:
There is already a separate API call: `Host.get_vms_which_prevent_evacuation`

ocaml/xapi/xapi_host.ml
robhoes: Yes, it would be better to log UUIDs. We currently have a mixed bag in the logs in general, but whenever I see a ref in the logs when triaging something I get disappointed. Would be good to add a `Db.VM.get_uuid` here.

lindig: I hear ya. I thought Refs are internal and UUIDs are for the API but can fish them out for logging.

lindig:
@robhoes :arrow_up:

robhoes (approved):

robhoes:
Please don't forget to squash the fixup...

PR #5071, opened 2023-06-23
CA-378837 log results from Host.get_vms_which_prevent_evacuation

lindig:
Backport of 6fb867c5d57298a2f5d5d334e838e3f704fa31b5

Host.get_vms_which_prevent_evacuation is used by XenCenter as part of a rolling pool upgrade. But when it fails we are missing information in the logs about VMs that block evacuation of a host. Add logging.


robhoes (approved):

lindig:
@edwintorok :arrow_up:

edwintorok (approved):

PR #5072, opened 2023-06-23
CA-378317 fix EBADF in waitpid_nohang (v2)

lindig:
waitpid_nohang closes the socket by calling waitpid and then calls clear_nonblock on that socket, which fails.

waitpid_nohang's policy around closing the socket is not obvious enough. Make it more explicit by not calling waitpid but inlining this code and lookign at the different cases. Also provide an option to never close the socket such that this resided with the caller - which is a cleaner protocol. The default behaviour is to still close the socket when the child process has terminated and not otherwise.

### DIFF
--- a/ocaml/forkexecd/lib/forkhelpers.ml
+++ b/ocaml/forkexecd/lib/forkhelpers.ml
@@ -72,14 +72,32 @@ let waitpid (sock, pid) =
       in
       failwith msg
 
-let waitpid_nohang ((sock, _) as x) =
-  Unix.set_nonblock sock ;
-  let r =
-    try waitpid x
-    with Unix.(Unix_error ((EAGAIN | EWOULDBLOCK), _, _)) ->
-      (0, Unix.WEXITED 0)
-  in
-  Unix.clear_nonblock sock ; r
+(* [waitpid_nohang] reports the status of a socket to a process. The
+   intention is to make this non-blocking. If the process is finished,
+   the socket is closed by default and not otherwise. The [close] option
+   can be used to never close the socket such that this is left to the
+   caller, which is a cleaner protocol. *)
+let waitpid_nohang ?(close = true) (sock, pid) =
+  let fail fmt = Printf.kprintf failwith fmt in
+  match Fecomms.read_raw_rpc sock with
+  | Ok Fe.(Finished (WEXITED n)) ->
+      if close then Unix.close sock ;
+      (pid, Unix.WEXITED n)
+  | Ok Fe.(Finished (WSIGNALED n)) ->
+      if close then Unix.close sock ;
+      (pid, Unix.WSIGNALED n)
+  | Ok Fe.(Finished (WSTOPPED n)) ->
+      if close then Unix.close sock ;
+      (pid, Unix.WSTOPPED n)
+  | Ok status ->
+      fail "%s: unexpected status received (%s)" __FUNCTION__
+        (Fe.ferpc_to_string status)
+  | Error msg ->
+      fail "%s: error happened when trying to read the status. %s" __FUNCTION__
+        msg
+  | exception exn ->
+      fail "%s: error happened when trying to read the status. %s" __FUNCTION__
+        (Printexc.to_string exn)
 
 let dontwaitpid (sock, _pid) =
   ( try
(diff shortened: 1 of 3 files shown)


ocaml/forkexecd/lib/forkhelpers.ml
```diff
@@ -60,14 +60,32 @@ let waitpid (sock, pid) =
...
+      if close then Unix.close sock ;
+      (pid, Unix.WEXITED n)
+  | Fe.Finished (Fe.WSIGNALED n) ->
+      if close then Unix.close sock ;
+      (pid, Unix.WSIGNALED n)
+  | Fe.Finished (Fe.WSTOPPED n) ->
+      if close then Unix.close sock ;
+      (pid, Unix.WSTOPPED n)
+  | exception Unix.(Unix_error ((EAGAIN | EWOULDBLOCK), _, _)) ->
+      Unix.clear_nonblock sock ;
```
edwintorok, line 82: This is a superset of https://github.com/xapi-project/xen-api/pull/5067, so closing that one, but see that PR for more discussions on what is wrong with forkexec usage.

edwintorok: To summarize that PR, I think I was worried there that the approach with closing the socket or not is too difficult to get right (so we may introduce an EBADF elsewhere by fixing this one), and I think my latest thinking was that we should change the protocol to remove the need for this weird protocol where sometimes the socket needs to stay open or not, but I haven't had time to work on that idea yet:

> I'd like to completely remove the open socket from here, and e.g. establish a fresh one to query forkexecd about the status of the process instead of keeping one open, but that has other drawbacks (it prevents FD leaks in the client, but may leave an FD/memory leak in forkexecd if we never 'wait' for a process that exited), I'll need to think about this more to see whether I find an alternative.

ocaml/forkexecd/lib/forkhelpers.ml
```diff
@@ -60,14 +60,32 @@ let waitpid (sock, pid) =
...
+  match Fecomms.read_raw_rpc sock with
+  | Fe.Finished (Fe.WEXITED n) ->
+      if close then Unix.close sock ;
+      (pid, Unix.WEXITED n)
+  | Fe.Finished (Fe.WSIGNALED n) ->
+      if close then Unix.close sock ;
+      (pid, Unix.WSIGNALED n)
+  | Fe.Finished (Fe.WSTOPPED n) ->
+      if close then Unix.close sock ;
+      (pid, Unix.WSTOPPED n)
```
edwintorok, line 80: These 3 match cases could be collapsed into one:
```
| Fe.Finished ((Fe.WEXITED _ | Fe.WSIGNALED _ | Fe.WSTOPPED  _) as status) ->
   if close then Unix.close sock;
   (pid, status)
```

ocaml/libs/stunnel/stunnel.ml
```diff
@@ -257,7 +257,7 @@ let disconnect_with_pid ?(wait = true) ?(force = false) pid =
           ( if wait then
               Forkhelpers.waitpid
           else
-            Forkhelpers.waitpid_nohang
+            Forkhelpers.waitpid_nohang ~close:true
```
edwintorok, line 260: This is I think the risky part (it should fix the other bug about an FD leak here, or it might introduce an EBADF bug here).

Perhaps split this into 2 commits? One that just does everything except this line (and we can test that and merge it as low risk), and another commit that adds this line which will need a closer review and more wider testing?

lindig:
I've rebased my PR and split out the call side into its own commit.

edwintorok (approved):
I can't spot more things wrong with it at the moment, but we'll need to do some testing.
Perhaps test the 2 commits separately (i.e. one koji build with 1 commit, and another with both) so that if there are bugs in one of them we can identify which one?

lindig:
I'll build this and run some tests

lindig:
Manual testing confirms it fixes the reported problem

ocaml/forkexecd/lib/forkhelpers.ml
psafont: Unfortunately this doesn't work because the types of Fe.WEXITED and Unix.WEXITED are not the same

lindig: Indeed. I tried it and came back to what I had.

psafont (approved):

lindig:
I'm not convinced this is correct - closing for now.

PR #5073, opened 2023-06-25
CP-43545: expose `issued` and `severity` from updateinfo

gangj:
Expose the `issued` and `severity` fields from updateinfo on host http endpoint: /updates.

### DIFF
--- a/ocaml/xapi/updateinfo.ml
+++ b/ocaml/xapi/updateinfo.ml
@@ -603,8 +603,13 @@ module UpdateInfo = struct
                                 Xapi_stdext_date.Date.epoch
                           in
                           {acc with issued}
-                      | Xml.Element ("severity", _, [Xml.PCData v]) ->
-                          {acc with severity= Severity.of_string v}
+                      | Xml.Element ("severity", _, [Xml.PCData v]) -> (
+                        try {acc with severity= Severity.of_string v}
+                        with e ->
+                          (* The error should not block update. Ingore it. *)
+                          warn "%s" (ExnHelper.string_of_exn e) ;
+                          acc
+                      )
                       | _ ->
                           acc
                     )
--- a/ocaml/xapi/updateinfo.ml
+++ b/ocaml/xapi/updateinfo.ml
@@ -578,7 +578,11 @@ module UpdateInfo = struct
                                          h m s
                                    )
                                   )
-                              with _ -> Xapi_stdext_date.Date.epoch
+                              with e ->
+                                (* The error should not block update. Ingore it
+                                   and set "issued" as epoch. *)
+                                warn "%s" (ExnHelper.string_of_exn e) ;
+                                Xapi_stdext_date.Date.epoch
                             )
                             | None ->
                                 Xapi_stdext_date.Date.epoch
--- a/ocaml/xapi/updateinfo.ml
+++ b/ocaml/xapi/updateinfo.ml
@@ -405,6 +405,21 @@ module LivePatch = struct
          )
 end
 
+module Severity = struct
+  type t = None | High
+
+  let to_string = function None -> "None" | High -> "High"
+
+  let of_string = function
+    | "None" ->
+        None
+    | "High" ->
+        High
+    | _ ->
+        error "Unknown severity in updateinfo" ;
+        raise Api_errors.(Server_error (invalid_updateinfo_xml, []))
+end
+
 module UpdateInfo = struct
   type t = {
       id: string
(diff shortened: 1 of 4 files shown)


ocaml/xapi/updateinfo.mli
```diff
@@ -111,6 +111,8 @@ module UpdateInfo : sig
     ; update_type: string
     ; livepatch_guidance: Guidance.t option
     ; livepatches: LivePatch.t list
+    ; issued: Xapi_stdext_date.Date.t
+    ; severity: string
```
minglumlu, line 115: I believe the possible values of `severity` are able to be enumerated. Then a new data type for this would be better.

lindig: It would be better to encode severity as an integer or enum. The domain of a string is too unrestricted.

ocaml/xapi/updateinfo.ml
```diff
@@ -558,6 +564,28 @@ module UpdateInfo = struct
...
+                                Xapi_stdext_date.Date.of_string
+                                  (Scanf.sscanf date
+                                     "%04d-%02d-%02d %02d:%02d:%02d"
+                                     (fun y mon d h m s ->
+                                       Printf.sprintf
+                                         "%04i%02i%02iT%02i:%02i:%02iZ" y mon d
+                                         h m s
+                                   )
+                                  )
+                              with _ -> Xapi_stdext_date.Date.epoch
```
minglumlu, line 581: It's worth to add a warning log for this.

ocaml/xapi/updateinfo.mli
gangj: I think the idea to add an Enum type in XAPI for `severity` is to check the correctness of the value for `severity` from  updateinfo, although it is rarely possible that the value will be invalid, I agree it is an enhancement to add it.
While for the proposal to "match the severity from String to an integer", currently there are only 2 levels of severity are defined: "High" and "None", it is easy to understand the order of the severity, and I think this match might require an extra step to understand the `severity` returned from XAPI, users might need to be aware of what severity in the updateinfo an integer presents.
So how about just add the `severity` enum type which contains "High" and "None"?

lindig: I'm fine with adding an enum; I was just offering a way to do this with fewer changes.

gangj: OK, thank you

ocaml/xapi/updateinfo.ml
```diff
@@ -558,6 +579,32 @@ module UpdateInfo = struct
...
+                                   and set "issued" as epoch. *)
+                                warn "%s" (ExnHelper.string_of_exn e) ;
+                                Xapi_stdext_date.Date.epoch
+                            )
+                            | None ->
+                                Xapi_stdext_date.Date.epoch
+                          in
+                          {acc with issued}
+                      | Xml.Element ("severity", _, [Xml.PCData v]) ->
+                          {acc with severity= Severity.of_string v}
```
minglumlu, line 607: This may raise exception as well. But this should not block the update. So I would suggest to add a warning log and ignore the error with a default value.

gangj: Sorry, I missed that! Added the exception handler.

minglumlu (approved):

gangj:
@lindig Would you please help to review this PR again and approve it if there is no issues, thank you~

ocaml/xapi/updateinfo.ml
```diff
@@ -405,6 +405,21 @@ module LivePatch = struct
          )
 end
 
+module Severity = struct
+  type t = None | High
+
+  let to_string = function None -> "None" | High -> "High"
+
+  let of_string = function
```
lindig, line 413: I would use `match  String.( x |> lowercase_ascii |> trim) with "none" -> .. | "high" -> ..` to be a bit more flexible how these are encoded. But it's fine to be strict, too.

ocaml/xapi/updateinfo.mli
```diff
@@ -97,6 +97,14 @@ module LivePatch : sig
   val of_xml : Xml.xml list -> t list
 end
 
+module Severity : sig
+  type t = None | High
+
+  val to_string : t -> string
+
+  val of_string : string -> t
```
lindig, line 105: Add a comment `(* may fail *)`

lindig (approved):

lindig:
This needs to be rebased onto `master` or the commit from master that disables Python 2.7 in the CI needs to be cherry picked.

ocaml/xapi/updateinfo.ml
gangj: Thank you

ocaml/xapi/updateinfo.mli
gangj: Added

gangj:
> This needs to be rebased onto `master` or the commit from master that disables Python 2.7 in the CI needs to be cherry picked.

Thanks for reminder, will rebase `master` in an individual PR for the feature branch.

PR #5074, opened 2023-06-26
CP-43574: Add host load data source

DeliZhangX:
- Host load refers to the number of vGPU in running or runnable status, it's similar definition to "load average" in "uptime" and "top" commands for process.
- The "hostload" metric is the host load per physical cpu.

### DIFF
--- a/ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
+++ b/ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
@@ -347,6 +347,39 @@ let dss_loadavg () =
     )
   ]
 
+let dss_hostload xc domains =
+  let physinfo = Xenctrl.physinfo xc in
+  let pcpus = physinfo.Xenctrl.nr_cpus in
+  let rec sum acc n f =
+    match n with n when n >= 0 -> sum (acc + f n) (n - 1) f | _ -> acc
+  in
+  let load =
+    List.fold_left
+      (fun acc (dom, _, domid) ->
+        sum 0 dom.Xenctrl.max_vcpu_id (fun id ->
+            let vcpuinfo = Xenctrl.domain_get_vcpuinfo xc domid id in
+            if vcpuinfo.Xenctrl.online && not vcpuinfo.Xenctrl.blocked then
+              1
+            else
+              0
+        )
+        + acc
+      )
+      0 domains
+  in
+  let load_per_cpu = float_of_int load /. float_of_int pcpus in
+  [
+    ( Rrd.Host
+    , Ds.ds_make ~name:"hostload" ~units:"(fraction)"
+        ~description:
+          ("Host load per physical cpu, where load refers to "
+          ^ "the number of vCPU(s) in running or runnable status."
+          )
+        ~value:(Rrd.VT_Float load_per_cpu) ~min:0.0 ~ty:Rrd.Gauge ~default:true
+        ()
+    )
+  ]
+
 (*****************************************************)
 (* network related code                              *)
 (*****************************************************)
@@ -729,6 +762,7 @@ let dom0_stat_generators =
   ; ("pcpus", fun xc _ _ -> dss_pcpus xc)
   ; ("vcpus", fun xc _ domains -> dss_vcpus xc domains)
   ; ("loadavg", fun _ _ _ -> dss_loadavg ())
+  ; ("hostload", fun xc _ domains -> dss_hostload xc domains)
   ; ("netdev", fun _ _ domains -> dss_netdev domains)
   ; ("cache", fun _ timestamp _ -> dss_cache timestamp)
   ]


ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
```diff
@@ -347,6 +347,41 @@ let dss_loadavg () =
     )
   ]
 
+let dss_hostload xc domains =
+  let physinfo = Xenctrl.physinfo xc in
+  let pcpus = physinfo.Xenctrl.nr_cpus in
+  let load =
+    List.fold_left
+      (fun acc (dom, _, domid) ->
+        let vcpu_ids = List.init (dom.Xenctrl.max_vcpu_id + 1) (fun i -> i) in
```
lindig, line 356: This is two nested `List.fold_left` and hence quadratic effort. Is this the best we can do? Also, it appears that `vcpu_ids` is simply to iterate over a list of numbers - we don't have to turn this into a list but could use a loop that increments a counter. Code like this should create less memory garbage.

DeliZhangX: No exiting interface to get all vGPU info directly, so have to get from domains indirectly. 
I have optimized to in loop way, could you help to review further?

ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
```diff
@@ -347,6 +347,37 @@ let dss_loadavg () =
     )
   ]
 
+let dss_hostload xc domains =
+  let physinfo = Xenctrl.physinfo xc in
+  let pcpus = physinfo.Xenctrl.nr_cpus in
+  let rec loop n f = if n < 0 then 0 else f n + loop (n - 1) f in
```
lindig, line 353: I would call this `sum` or `total`- the result is the sum of `f n` + `f (n-1)` + `f (n-2)` + ..; specifically, this is used to compute the number of active vcpus for every domain.

lindig (approved):

ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
```diff
@@ -347,6 +347,37 @@ let dss_loadavg () =
...
+      )
+      0 domains
+  in
+  let load_per_cpu = float_of_int load /. float_of_int pcpus in
+  [
+    ( Rrd.Host
+    , Ds.ds_make ~name:"hostload" ~units:"(fraction)"
+        ~description:
+          ("Host load per physical cpu, where load refers to "
+          ^ "the number of vGPU(s) in running or runnable status."
```
minglumlu, line 374: I can't see how does this relate with `vGPU`?

ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
```diff
@@ -347,6 +347,37 @@ let dss_loadavg () =
...
+            let vcpuinfo = Xenctrl.domain_get_vcpuinfo xc domid id in
+            if vcpuinfo.Xenctrl.online && not vcpuinfo.Xenctrl.blocked then
+              1
+            else
+              0
+        )
+        + acc
+      )
+      0 domains
+  in
```
minglumlu, line 367:
```suggestion
  let load =
    List.fold_left
      (fun acc (dom, _, domid) ->
        let rec sum cnt id =
          match id with
          | id when id >= 0 ->
              let vcpuinfo = Xenctrl.domain_get_vcpuinfo xc domid id in
              if vcpuinfo.Xenctrl.online && not vcpuinfo.Xenctrl.blocked then
                sum (1 + cnt) (id - 1)
              else
                sum cnt (id - 1)
          | _ ->
              cnt
        in
        let vcpus = sum 0 dom.Xenctrl.max_vcpu_id
        vcpus + acc
      )
      0 domains
  in
```

ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
```diff
@@ -347,6 +347,37 @@ let dss_loadavg () =
...
+  let load_per_cpu = float_of_int load /. float_of_int pcpus in
+  [
+    ( Rrd.Host
+    , Ds.ds_make ~name:"hostload" ~units:"(fraction)"
```
minglumlu, line 371: I understand this is vCPU-to-core ratio. How about `vcpu_to_pcpu_ratio`?

ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
minglumlu, line 374: How about `the ratio between the number of running/runnable vCPUs and the number of physical CPU cores`

ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
```diff
@@ -347,6 +347,37 @@ let dss_loadavg () =
     )
   ]
 
+let dss_hostload xc domains =
+  let physinfo = Xenctrl.physinfo xc in
+  let pcpus = physinfo.Xenctrl.nr_cpus in
+  let rec loop n f = if n < 0 then 0 else f n + loop (n - 1) f in
+  let load =
+    List.fold_left
+      (fun acc (dom, _, domid) ->
```
minglumlu, line 356: Not sure if it is possible to read stale data and hence get error or exception. 
I would suggest to ignore the potential errors and exceptions, instead of populate them.

ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
lindig: This is not tail recursive - so `n` should not be larger than 100 or so

ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
DeliZhangX: Fixed!

ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
DeliZhangX: Fixed, changed to tail recursive.

ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
DeliZhangX: There exists another ds `loadavg` which `load` is defined in "uptime" and "top", we just extend the concept to host.

ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
DeliZhangX: I checked other ds using `Xenctrl`, they do not catch exception either.

minglumlu: I think they are catching. https://github.com/xapi-project/xen-api/blob/fa18bc0ace9932011058a1190b34d04c080e56bf/ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml#L292

And this could be confirmed in https://code.citrite.net/projects/XSU/repos/xen/browse/tools/ocaml/libs/xc/xenctrl_stubs.c#522 

So I think it would be better to catch the exceptions.

ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
```diff
@@ -347,6 +347,39 @@ let dss_loadavg () =
...
+  in
+  let load =
+    List.fold_left
+      (fun acc (dom, _, domid) ->
+        sum 0 dom.Xenctrl.max_vcpu_id (fun id ->
+            let vcpuinfo = Xenctrl.domain_get_vcpuinfo xc domid id in
+            if vcpuinfo.Xenctrl.online && not vcpuinfo.Xenctrl.blocked then
+              1
+            else
+              0
```
minglumlu, line 364: A minor observation: adding with 0 seems not necessary, could just return the `acc` here.

minglumlu (approved):

ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
DeliZhangX: `domain_get_vcpuinfo` may raise exception indeed.
I checked following code to call the ds generator which already handled all exception and return `[]` ds
```
let generate_all_dom0_stats xc timestamp domains =
  let handle_generator (name, generator) =
    (name, handle_exn name (fun _ -> generator xc timestamp domains) [])
  in
  List.map handle_generator dom0_stat_generators
```

ocaml/xcp-rrdd/bin/rrdd/xcp_rrdd.ml
DeliZhangX: Confirmed with Ming, won't change.

PR #5075, opened 2023-06-27
CP-40214: **/*.py: raise (AnyException()): Remove optional parentheses

bernhardkaindl:
- Remove optional parentheses for "raise (...AnyException..)))":
  ```ml
  sed -i '/raise (/s/)$//  ;  s/raise (/raise /' $(find ocaml -name *.py
  ```
  Also updates these in with the same change:
  - ocaml/message-switch/python/message_switch.py
  - ocaml/xapi-idl/designs/interprocesscomms.md

### DIFF
--- a/ocaml/message-switch/python/message_switch.py
+++ b/ocaml/message-switch/python/message_switch.py
@@ -210,7 +210,7 @@ def link_send(sock, m):
 def link_recv(reader):
     status = reader.readline()
     if not(status.startswith("HTTP/1.1 200 OK")):
-        raise (Bad_status(status))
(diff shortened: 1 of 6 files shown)


edwintorok (approved):
This should be a no-op on python2.

edwintorok:
(FWIW the parenthesis work on Python3 too, at least from what I tested, but tools like `2to3` get very confused by it and do an incorrect transformation dropping the exception arguments instead. Since parenthesis aren't really required after the `raise` keyword it is better to drop it)

lindig (approved):

PR #5076, opened 2023-06-27
Push CPU feature set logic from xapi to xenopsd

robhoes:
The goal is to make CPU feature set logic abstract to xapi, and eventually delegate all logic to an external library, such as libxc.


ocaml/libs/cpuid/dune
```diff
@@ -0,0 +1,8 @@
+(library
+ (name cpuid)
+ (public_name cpuid)
```
edwintorok, line 3: If you only use it inside xenopsd you don't have to make it a public library (simplifies the spec file as you won't need to install it). Then you also won't need an opam file for it.

ocaml/xenopsd/xc/xenops_server_xen.ml
```diff
@@ -1121,6 +1121,18 @@ module HOST = struct
...
+    let open Cpuid in
+    intersect (features_of_string policy1) (features_of_string policy2)
+    |> string_of_features
+
+  let is_compatible vm_policy host_policy =
+    let open Cpuid in
+    let vm = features_of_string vm_policy in
+    let host = features_of_string host_policy in
+    let vm' = zero_extend vm (Array.length host) in
+    is_subset vm' host
```
edwintorok, line 1135: See the XAPI code, it would be useful to print a diff here, that helps determining what went wrong in many cases.

ocaml/libs/cpuid/dune
robhoes: I needed it public for the unit tests at one point, but perhaps not anymore... Let me check...

ocaml/xapi-idl/xen/xenops_interface.ml
```diff
@@ -618,6 +611,37 @@ module XenopsAPI (R : RPC) = struct
...
+      let policy2_p =
+        Param.mk ~description:["CPU policy 2"] ~name:"policy2" Types.string
+      in
+      let policy3_p =
+        Param.mk ~description:["Combined CPU policy"] ~name:"policy3"
+          Types.string
+      in
+      declare "HOST.combine_cpu_policies"
+        ["Combine CPU policy to get a common subset"]
+        (debug_info_p @-> policy1_p @-> policy2_p @-> returning policy3_p err)
```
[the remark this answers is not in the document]
robhoes: Yes, once we change the format from a bitmap to something else, then we'd need to think about compatibility. However, not yet at this point.

ocaml/xapi-idl/xen/xenops_interface.ml
```diff
@@ -465,11 +465,11 @@ module Host = struct
     ; model: string
     ; stepping: string
     ; flags: string
-    ; features: int64 array
-    ; features_pv: int64 array
-    ; features_hvm: int64 array
-    ; features_pv_host: int64 array
-    ; features_hvm_host: int64 array
+    ; features: string
```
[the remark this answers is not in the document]
robhoes: This thing is that this needs to be a string in order for xapi to store in the database, and it needs to be serialisable for ocaml-rpc as well. I don't think it can be abstract in the interface.

ocaml/libs/cpuid/dune
robhoes: Yep, I can make it non-public.

ocaml/xapi/cpuid_helpers.ml
```diff
@@ -211,19 +112,12 @@ let assert_vm_is_compatible ~__context ~vm ~host ?remote () =
...
            features %s"
-          (string_of_features vm_cpu_features)
-          (string_of_features host_cpu_features') ;
-        let vm_cpu_features' =
-          vm_cpu_features
-          |> upgrade_features ~__context ~vm ~host host_cpu_features'
-        in
-        if not (is_subset vm_cpu_features' host_cpu_features') then (
+          vm_cpu_features host_cpu_features ;
+        if not (Xenopsd.HOST.is_compatible dbg vm_cpu_features host_cpu_features)
```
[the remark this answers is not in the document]
robhoes: No, it doesn't :) Unfortunately ocaml-rpc doesn't provide named arguments to calls in the client module.

ocaml/libs/cpuid/cpuid.ml
```diff
@@ -0,0 +1,81 @@
...
+
+let string_of_features features =
+  Array.map (Printf.sprintf "%08Lx") features
+  |> Array.to_list
+  |> String.concat "-"
+
+exception InvalidFeatureString of string
+
+let features_of_string str =
+  let scanf fmt s = Scanf.sscanf s fmt (fun x -> x) in
```
lindig, line 23: Can use `Fun.id`

ocaml/libs/cpuid/cpuid.ml
```diff
@@ -0,0 +1,81 @@
...
+  try
+    String.split_on_char '-' str
+    |> (fun lst -> if lst = [""] then [] else lst)
```
lindig, line 26:
```
(function [""] -> [] | lst -> lst) 
```

ocaml/libs/cpuid/cpuid.ml
```diff
@@ -0,0 +1,81 @@
...
+    |> Array.of_list
+    |> Array.map (scanf "%08Lx%!")
+  with _ -> raise (InvalidFeatureString str)
+
+(** If arr0 is shorter than arr1, extend arr0 with elements from arr1 up to the
+ *  length of arr1. Otherwise, truncate arr0 to the length of arr1. *)
+let extend arr0 arr1 =
```
lindig, line 33: Is the name misleading? The code returns a new array that is some kind of alignment of `arr0` with `arr1`; this is not necessarily an extension but could be a truncation as well. Could be called `align arr0 ~with`; `pad` could be another name but it implies adding elements.

The semantics of this operation seems strange in the case where data from arr1 is taken to pad arr0. Maybe it would be better to call this step normalisation. All feature sets are normalised before doing arithmetic on them,

robhoes: The code in cpuid.ml is just moved from cpuid_helpers.ml to a library. I don't really want to change it if not necessary, because it is known to work, and will eventually go away.

ocaml/xenopsd/xc/xenops_server_xen.ml
robhoes: Yes, we compute and can print the diff here now. Once we switch to the C version of the function to do the comparison, then we would need that function to return a diff in some form.

robhoes:
@edwintorok see the new commit for the type changes.

edwintorok:
Try in dune: '(flags (:standard -w -27))' to work around the unused var from ppx_deriving_rpc.

ocaml/xenopsd/xc/xenops_server_xen.ml
robhoes: I've done that now.

ocaml/xapi-idl/xen/xenops_interface.ml
```diff
@@ -618,6 +611,37 @@ module XenopsAPI (R : RPC) = struct
...
+      let host_policy_p =
+        Param.mk ~description:["Host CPU policy"] ~name:"host_policy"
+          Types.string
+      in
+      declare "HOST.is_compatible"
+        ["Check whether a VM can live-migrate to or resume on a host"]
+        (debug_info_p
+        @-> vm_policy_p
+        @-> host_policy_p
+        @-> returning (Param.mk Types.bool) err
```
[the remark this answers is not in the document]
robhoes: Logged in xenopsd now.

ocaml/xapi/cpuid_helpers.ml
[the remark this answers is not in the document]
robhoes: I've tried to solve this using an phantom type in the interface (similar to the way `Ref.t` works), to distinguish between host and VM feature sets.

ocaml/xapi/create_misc.ml
```diff
@@ -687,105 +673,81 @@ let create_host_cpu ~__context host_info =
...
       (fun (r, s) -> (r, s.API.host_cpu_info))
       (Db.Host.get_all_records ~__context)
   in
-  let getfdefault ~defaultf key host =
-    let default = getf defaultf host in
-    getf ~default key host
-  in
   let merge pool (hostref, host) =
+    let combine a b =
+      match (CPU_policy.to_string a, CPU_policy.to_string b) with
```
edwintorok, line 687: This could have an 'is_empty' to avoid converting to string only to check for emptyness (well it is a string currently, so it is probably fine as it is for now)?

ocaml/xapi-idl/xen/xenops_interface.ml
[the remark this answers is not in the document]
edwintorok: this got changed to 'host_cpu_policy'.

edwintorok (commented):
Thanks for the phantom type changes, it should make the code more robust, I can't spot any errors, but I'll take a deeper look a bit later.

lindig (approved):

edwintorok (approved):

ocaml/xapi-idl/xen/xenops_interface.ml
```diff
@@ -201,6 +201,53 @@ module Network = struct
...
+
+  val vm : [`vm] t Rpc.Types.def
+
+  val host : [`host] t Rpc.Types.def
+
+  val typ_of : 'a -> 'a t Rpc.Types.typ
+end = struct
+  type 'a t = string [@@deriving rpc]
+
+  let of_string _ s = s
```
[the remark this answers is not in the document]
robhoes: Yes, this data will eventually come directly from libxc, not from user input, and is just passed from xenopsd to xapi. We don't want to restrict it at this point.

ocaml/xenopsd/test/test_cpuid.ml
```diff
@@ -0,0 +1,329 @@
...
+
+    let string_of_output_t = Test_printers.(array int64)
+  end
+
+  let transform (arr0, arr1) = extend arr0 arr1
+
+  let tests =
+    `QuickAndAutoDocumented
+      [
+        (([||], [||]), [||])
```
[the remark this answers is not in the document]
robhoes: Some of the more complicated ones below do that.

ocaml/xapi-idl/xen/xenops_interface.ml
```diff
@@ -618,6 +658,37 @@ module XenopsAPI (R : RPC) = struct
...
+      let policy2_p =
+        Param.mk ~description:["CPU policy 2"] ~name:"policy2" CPU_policy.host
+      in
+      let policy3_p =
+        Param.mk ~description:["Combined CPU policy"] ~name:"policy3"
+          CPU_policy.host
+      in
+      declare "HOST.combine_cpu_policies"
+        ["Combine CPU policy to get a common subset"]
+        (debug_info_p @-> policy1_p @-> policy2_p @-> returning policy3_p err)
```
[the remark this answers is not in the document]
robhoes: Yes, it is possible to reuse a generic one. It's just that you can have different docs strings this way.

psafont (approved):

PR #5077, opened 2023-06-27
CA-379173 handle race condition in stunnel_cache

lindig:
We haev seen a hash table lookup fail unexpectedly. Handle this gracefully.

This does not fix the underlying race condition. The current code strikes me as quite challenging to improve in this regard.


edwintorok (commented):
The bug might originate from the 2 different loops below that handle '!to_gc'. If one of them fails then the other one is not run at all, leading to this inconsistent state.

I'm not sure whether ignoring the entry would result in a leak or not. Would it be safer to add this to the 'to_gc' list anyway if we don't find the times?

lindig:
This could be equally risky: now you are trying to GC something that is in an inconsistent state

robhoes (approved):

edwintorok (approved):

PR #5078, opened 2023-06-27
ocaml/xapi-storage/python/xapi/**.py: modernize -f except,print

bernhardkaindl:
Reviewed by me(@bernhardkaindl):

Migrate `xapi-storage` Python code to Python3: __Initial part of step 1__:

In order for a clear and easy review, this commit adds only two fixes:
```ml
modernize -wf except,print $(find ocaml/xapi-storage/xapi -name *.py)
```
This `modernize` makes these changes:
- `except Exception, e` to: `except Exception as e`
- Updates print() with `from __future__ import print_function`, for forcing Python2 to `raise SyntaxError` on Python2 `print` (the import of print_function could be omitted without problem)

### DIFF
--- a/ocaml/xapi-storage/python/xapi/storage/api/volume.py
+++ b/ocaml/xapi-storage/python/xapi/storage/api/volume.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 from xapi import success, Rpc_light_failure, InternalError, UnmarshalException, TypeError, is_long, UnknownMethod
 import xapi
 import sys
@@ -656,8 +657,8 @@ class Volume_commandline():
             request = self._parse_create()
             use_json = 'json' in request and request['json']
             results = self.dispatcher.create(request)
-            print json.dumps(results)
-        except Exception, e:
+            print(json.dumps(results))
+        except Exception as e:
             if use_json:
                 xapi.handle_exception(e)
             else:
@@ -669,8 +670,8 @@ class Volume_commandline():
             request = self._parse_snapshot()
             use_json = 'json' in request and request['json']
             results = self.dispatcher.snapshot(request)
-            print json.dumps(results)
-        except Exception, e:
+            print(json.dumps(results))
+        except Exception as e:
             if use_json:
                 xapi.handle_exception(e)
             else:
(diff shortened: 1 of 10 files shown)


ocaml/xapi-storage/python/xapi/storage/api/volume.py
```diff
@@ -1106,7 +1107,7 @@ def __init__(self):
     def probe(self, dbg, uri):
         """Operations which act on Storage Repositories"""
         result = {}
-        result["result"] = { "srs": [ { "sr": "string", "name": "string", "description": "string", "free_space": 0L, "total_space": 0L, "datasources": [ "string", "string" ], "clustered": True, "health": None }, { "sr": "string", "name": "string", "description": "string", "free_space": 0L, "total_space": 0L, "datasources": [ "string", "string" ], "clustered": True, "health": None } ], "uris": [ "string", "string" ] }
+        result["result"] = { "srs": [ { "sr": "string", "name": "string", "description": "string", "free_space": 0, "total_space": 0, "datasources": [ "string", "string" ], "clustered": True, "health": None }, { "sr": "string", "name": "string", "description": "string", "free_space": 0, "total_space": 0, "datasources": [ "string", "string" ], "clustered": True, "health": None } ], "uris": [ "string", "string" ] }
```
edwintorok, line 1110: What are the consequences of changing '0' to '0L' here? One will be 'int', the other will be 'long' in Python2. Can that potentially confuse the RPC serialization/deserialization glue code, will it handle both?

lindig: In Python3, all integers are long.

bernhardkaindl: @edwintorok I pushed an update of this commit for fixing it to settle any doubts:

https://github.com/xapi-project/xen-api/pull/5078/commits/ac683ca9815f65ddc48bffdb4e242909387c3daa

Using `futurize --step1` would have replaced `0L` with `0`. But for Python2, that would have changed values from `long` to `int`. Fix this like `ocaml-rpc/src/lib/pythongen.ml` already does it for the generated code:

1. Add:
   ```py
   if sys.version_info[0] > 2:
       long = int
       unicode = str
       str = bytes
   ```

   This is already used in the code generated by by `pythongen.ml`:
   https://github.com/mirage/ocaml-rpc/blob/master/src/lib/pythongen.ml#L102

2. Apply:
   ```ml
   sed -i 's/ 0L, / long(0), /g' ocaml/xapi-storage/python/xapi/storage/api/volume.py
   ```
   This is already used in the code generated by by `pythongen.ml`:
   https://github.com/mirage/ocaml-rpc/blob/master/src/lib/pythongen.ml#L284

edwintorok (approved):

psafont (approved):

PR #5079, opened 2023-06-28
CA-378757: remove "EvacuateHost" from recommended guidance

gangj:
In the update applying process, "EvacuateHost" will be applied before applying updates.

### DIFF
--- a/ocaml/xapi/repository.ml
+++ b/ocaml/xapi/repository.ml
@@ -750,8 +750,12 @@ let apply_updates' ~__context ~host ~updates_info ~livepatches ~acc_rpm_updates
   (* Evaluate recommended/pending guidances *)
   let recommended_guidances, pending_guidances =
     let recommended_guidances' =
-      eval_guidances ~updates_info ~updates:acc_rpm_updates ~kind:Recommended
-        ~livepatches:successful_livepatches ~failed_livepatches
+      (* EvacuateHost will be applied before applying updates *)
+      List.filter
+        (fun g -> not (g = Guidance.EvacuateHost))
+        (eval_guidances ~updates_info ~updates:acc_rpm_updates ~kind:Recommended
+           ~livepatches:successful_livepatches ~failed_livepatches
+        )
     in
     let pending_guidances' =
       List.filter
--- a/ocaml/xapi/repository.ml
+++ b/ocaml/xapi/repository.ml
@@ -751,18 +751,15 @@ let apply_updates' ~__context ~host ~updates_info ~livepatches ~acc_rpm_updates
   let recommended_guidances, pending_guidances =
     let recommended_guidances' =
       (* EvacuateHost will be applied before applying updates *)
-      List.filter
-        (fun g -> g <> Guidance.EvacuateHost)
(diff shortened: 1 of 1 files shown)


ocaml/xapi/repository.ml
```diff
@@ -750,8 +750,12 @@ let apply_updates' ~__context ~host ~updates_info ~livepatches ~acc_rpm_updates
   (* Evaluate recommended/pending guidances *)
   let recommended_guidances, pending_guidances =
     let recommended_guidances' =
-      eval_guidances ~updates_info ~updates:acc_rpm_updates ~kind:Recommended
-        ~livepatches:successful_livepatches ~failed_livepatches
+      (* EvacuateHost will be applied before applying updates *)
+      List.filter
+        (fun g -> not (g = Guidance.EvacuateHost))
```
lindig, line 755: Use `fun guidance -> guidance <> Guidance.EvacuateHost`; can be also written as `((<>) Guidance.EvacuateHost)` but it's a bit cryptic.

Does this make `EvacuateHost` obsolete? Can this guidance still reach a customer? The code also suggests to me it would be better to use sets than lists to compute these (below).

lindig (approved):

ocaml/xapi/repository.ml
gangj: > Does this make EvacuateHost obsolete? Can this guidance still reach a customer?
`EvacuateHost` guidance is available to customer before applying update with the HTTP endpoint "/updates", and the client (XenCenter) should handle the guidance`EvacuateHost` before applying update.

> The code also suggests to me it would be better to use sets than lists to compute these (below).
Yes, but I would like not to change this in this PR. :)

ocaml/xapi/repository.ml
```diff
@@ -750,8 +750,12 @@ let apply_updates' ~__context ~host ~updates_info ~livepatches ~acc_rpm_updates
...
   let recommended_guidances, pending_guidances =
     let recommended_guidances' =
-      eval_guidances ~updates_info ~updates:acc_rpm_updates ~kind:Recommended
-        ~livepatches:successful_livepatches ~failed_livepatches
+      (* EvacuateHost will be applied before applying updates *)
+      List.filter
+        (fun g -> g <> Guidance.EvacuateHost)
+        (eval_guidances ~updates_info ~updates:acc_rpm_updates ~kind:Recommended
+           ~livepatches:successful_livepatches ~failed_livepatches
+        )
```
minglumlu, line 758: Would this way look clearer?
```suggestion
      (* EvacuateHost will be applied before applying updates *)
      eval_guidances ~updates_info ~updates:acc_rpm_updates ~kind:Recommended
           ~livepatches:successful_livepatches ~failed_livepatches
      |> List.filter (fun g -> g <> Guidance.EvacuateHost)
```

minglumlu (approved):

ocaml/xapi/repository.ml
gangj: Changed this way also for the following lines of code to keep the same style.

PR #5083, opened 2023-06-29
Instrument named mutexes with distributed tracing

edwintorok:
When distributed tracing is enabled this will generate spans:
* 'waiting_for' and 'acquired' are begin/end spans for the 'waiting_for(lockname)' span
* 'acquired' and 'released' are begin/end spans for the 'acquired(lockname)' span

An optional '?parent' argument to 'waiting_for' enables tracing, which is in turn retrieved from `Context.t` and `None` by default, unless an Observer has been created and enabled.
I initially wanted to enforce a strict waiting_for..acquired..released order, but due to how callbacks work in Xmlrpc/stunnel there will always be optional values. Also there is no 'waiting_for' event generated for stunnel, because we only know its PID at the very end once it is already connected (instrumenting that would require digging into stunnel.ml which is out of scope for now). I do enforce strict pairing between acquired/released though.

This adds some minimal parent tracing to 3 SMAPI calls that take named mutexes, but I haven't instrumented the SMAPI further (I'll wait for @robhoes 's changes first).

Also I do not create a new context and do not act as a parent span for functions called while the lock is held, that can be fixed in the future by plumbing a __context through to appropriate places and changing Named_mutex.execute to invoke its 'f' argument with a context or parent span.

TODO: instead of constructing a new string every time we wait for / acquire/ release a lock pre-construct the string and store it in 'type resource', that should reduce the overhead of this instrumentation. Hence the draft in the PR.

### DIFF
--- a/ocaml/xapi/locking_helpers.ml
+++ b/ocaml/xapi/locking_helpers.ml
@@ -188,20 +188,16 @@ module Thread_state = struct
     List.map fst
       (IntMap.fold (fun _ ts acc -> ts.acquired_resources @ acc) snapshot [])
 
-  let update f =
-    let old = Thread_local_storage.get thread_states in
-    f old
+  let get_states () = Thread_local_storage.get thread_states
 
   let with_named_thread name task f =
-    update (fun ts ->
-        ts.name <- name ;
-        ts.task <- task
-    ) ;
+    let ts = get_states () in
+    ts.name <- name ;
+    ts.task <- task ;
     finally f (fun () ->
-        update (fun ts ->
-            ts.name <- "" ;
-            ts.task <- Ref.null
-        )
+        let ts = get_states () in
+        ts.name <- "" ;
+        ts.task <- Ref.null
     )
 
   let now () = Unix.gettimeofday ()
@@ -222,7 +218,8 @@ module Thread_state = struct
               None
         )
     in
-    update (fun ts -> ts.waiting_for <- resource) ;
+    let ts = get_states () in
+    ts.waiting_for <- resource ;
     span
 
   let acquired resource parent =
@@ -242,18 +239,16 @@ module Thread_state = struct
               None
         )
     in
-    update (fun ts ->
-        ts.waiting_for <- none ;
-        ts.acquired_resources <- (resource, now ()) :: ts.acquired_resources
-    ) ;
+    let ts = get_states () in
+    ts.waiting_for <- none ;
+    ts.acquired_resources <- (resource, now ()) :: ts.acquired_resources ;
     span
 
   let released resource span =
     let (_ : (_, _) result) = Tracing.Tracer.finish span in
-    update (fun ts ->
-        ts.acquired_resources <-
-          List.filter (fun (r, _) -> r <> resource) ts.acquired_resources
-    )
+    let ts = get_states () in
+    ts.acquired_resources <-
+      List.filter (fun (r, _) -> r <> resource) ts.acquired_resources
 
   let to_graphviz () =
     let t' = now () in
--- a/ocaml/xapi/locking_helpers.ml
+++ b/ocaml/xapi/locking_helpers.ml
@@ -274,12 +274,38 @@ module Thread_state = struct
     String.concat "\n" all
 
   let known_threads () = with_lock m (fun () -> IntMap.cardinal !thread_states)
+
+  let with_resource resource acquire f release arg =
+    let acquired = acquire resource arg in
+    match f () with
+    | r ->
+        release resource acquired ; r
+    | exception e ->
+        let bt = Printexc.get_raw_backtrace () in
+        D.log_and_ignore_exn (fun () -> release resource acquired) ;
+        Printexc.raise_with_backtrace e bt
 end
 
 module Named_mutex = struct
-  type t = {name: string; m: Mutex.t; r: resource}
+  type t = {
+      name: string
+    ; m: Mutex.t
+    ; r: resource
+    ; acquire: t -> Tracing.Span.t option -> Thread_state.acquired
+    ; release: t -> Thread_state.acquired -> unit
+  }
 
-  let create name = {name; m= Mutex.create (); r= lock name}
+  let create name =
+    let acquire t parent =
+      let waiting = Thread_state.waiting_for ?parent t.r in
+      Mutex.lock t.m ;
+      Thread_state.acquired t.r waiting
+    in
+    let release t waiting =
+      Mutex.unlock t.m ;
+      Thread_state.released t.r waiting
+    in
+    {name; m= Mutex.create (); r= lock name; acquire; release}
 
   let execute ?__context ?parent (x : t) f =
     let parent =
@@ -289,9 +315,5 @@ module Named_mutex = struct
       | Some _ as p ->
           p
     in
-    let waiting = Thread_state.waiting_for ?parent x.r in
-    with_lock x.m (fun () ->
-        let acquired = Thread_state.acquired x.r waiting in
-        finally f (fun () -> Thread_state.released x.r acquired)
-    )
+    Thread_state.with_resource x x.acquire f x.release parent
 end
--- a/ocaml/xapi/locking_helpers.ml
+++ b/ocaml/xapi/locking_helpers.ml
@@ -37,6 +37,10 @@ let kill_resource = function
       Unix.kill pid Sys.sigkill
 
 module Thread_state = struct
+  type waiting = unit
+
+  type acquired = unit
+
   type time = float
 
   type t = {
@@ -98,7 +102,7 @@ module Thread_state = struct
   let waiting_for resource =
     update (fun ts -> {ts with waiting_for= Some (resource, now ())})
 
-  let acquired resource =
+  let acquired resource (_ : unit) =
     update (fun ts ->
         {
           ts with
@@ -107,7 +111,7 @@ module Thread_state = struct
         }
     )
 
-  let released resource =
+  let released resource () =
     update (fun ts ->
         {
           ts with
@@ -217,9 +221,9 @@ module Named_mutex = struct
 
   let execute (x : t) f =
     let r = Lock x.name in
-    Thread_state.waiting_for r ;
+    let waiting = Thread_state.waiting_for r in
     with_lock x.m (fun () ->
-        Thread_state.acquired r ;
-        finally f (fun () -> Thread_state.released r)
+        let acquired = Thread_state.acquired r waiting in
+        finally f (fun () -> Thread_state.released r acquired)
     )
 end
--- a/ocaml/xapi/locking_helpers.ml
+++ b/ocaml/xapi/locking_helpers.ml
@@ -85,7 +85,7 @@ module Thread_local_storage = struct
     let self = Thread.self () in
     LiveThreads.replace t.tbl (Thread_key.of_thread self) v
 
-  let set t v = with_lock t set_unlocked v
+  let _set t v = with_lock t set_unlocked v
 
   let snapshot_unlocked t () =
     LiveThreads.fold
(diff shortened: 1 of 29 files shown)


edwintorok:
Screenshot 2023-06-29 at 17 59 20
here is an example of how the generated spans look like

lindig (approved):
I am approving this to help moving this forward.

edwintorok:
Added a small benchmark, mainly interested in the overhead of the Mutex -> Named_mutex switch:
```
Running benchmarks
╭──────────────────────────────────────┬───────────────────────────┬───────────────────────────┬───────────────────────────╮
│name                                  │  major-allocated          │  minor-allocated          │  monotonic-clock          │
├──────────────────────────────────────┼───────────────────────────┼───────────────────────────┼───────────────────────────┤
│  Mutex/Mutex.execute                 │             0.0000 mjw/run│             7.3333 mnw/run│            689.2567 ns/run│
│  Mutex/NamedMutex.execute            │             0.0000 mjw/run│           138.3333 mnw/run│           5408.7165 ns/run│
│  Mutex/NamedMutex.execute (tracing)  │             0.0000 mjw/run│           705.9894 mnw/run│          25930.8053 ns/run│
╰──────────────────────────────────────┴───────────────────────────┴───────────────────────────┴───────────────────────────╯
```

According to this Named mutexes are ~7x slower than regular ones, and enabling tracing further slows it down ~4x. I'll profile this a bit more and see what I can optimise, the goal with instrumenting locks with the named mutex was to improve performance, but if the instrumenation (even when off) is that slow then it might be better to make the named mutex switch optional too (or optimize it).

edwintorok:
Reduced allocations:
```
Running benchmarks
╭──────────────────────────────────────┬───────────────────────────┬───────────────────────────┬───────────────────────────╮
│name                                  │  major-allocated          │  minor-allocated          │  monotonic-clock          │
├──────────────────────────────────────┼───────────────────────────┼───────────────────────────┼───────────────────────────┤
│  Mutex/Mutex.execute                 │             0.0000 mjw/run│             7.1915 mnw/run│            572.9840 ns/run│
│  Mutex/Mutex.lock/unlock             │             0.0000 mjw/run│             3.0612 mnw/run│            529.4312 ns/run│
│  Mutex/NamedMutex.execute            │             0.0030 mjw/run│            25.4971 mnw/run│           3702.8499 ns/run│
│  Mutex/NamedMutex.execute (tracing)  │             0.0000 mjw/run│           490.9894 mnw/run│          21692.9272 ns/run│
╰──────────────────────────────────────┴───────────────────────────┴───────────────────────────┴───────────────────────────╯```

This now 30% faster than before, and memory allocations got reduced ~5.4x, there are still some allocations to eliminate.
The code is becoming less and less functional though (avoiding allocations/optimizing performance drives toward using hashtables instead of maps). The slow paths are still functional though.

I'll try to eliminate those last few allocations, and then also add some unit tests for Locking_helpers/Thread_state and Named_mutex. And also a leak test for the new Thread_local_state which relies on ephemerons now (+its own lock). 

In the end I think there'll still be an overhead for switching to named mutex from the regular one (it does track a lot more data on who is waiting on who and who is holding which mutex). That may be useful information to have (the performance issue happens once the mutex is acquired, not in the mutex itself), but I'll create a separate
InstrumentedMutex nevertheless that could be used just to add Tracing or not to a regular mutex, and when tracing is off is should have minimal to no overhead, so we can more freely switch between implementations as needed.

File: ocaml/libs/log/debug.ml [hunk 1]
```diff
@@ -310,6 +310,10 @@ module type DEBUG = sig
   val log_backtrace : unit -> unit
 
   val log_and_ignore_exn : (unit -> unit) -> unit
+
+  val log_and_ignore_exn1 : ('a -> unit) -> 'a -> unit
```
psafont, line 314: Are these needed? The first log_and_ignore_exn can be used with functions of any number or arguments:
```ocaml
D.log_and_ignore_exn2 (fun () -> release resource acquired) ;
```

edwintorok:
Some more optimizations, I've eliminated most allocations from the named mutex now:
```
╭────────────────────────────┬───────────────────────────┬───────────────────────────┬───────────────────────────╮
│name                        │  major-allocated          │  minor-allocated          │  monotonic-clock          │
├────────────────────────────┼───────────────────────────┼───────────────────────────┼───────────────────────────┤
│  Mutex/Mutex.execute       │             0.0000 mjw/run│             7.4884 mnw/run│            995.6693 ns/run│
│  Mutex/Mutex.lock/unlock   │             0.0000 mjw/run│             3.3333 mnw/run│            705.3188 ns/run│
│  Mutex/NamedMutex.execute  │             0.0000 mjw/run│             5.4884 mnw/run│           3448.5192 ns/run│
╰────────────────────────────┴───────────────────────────┴───────────────────────────┴───────────────────────────╯
```

ocaml/libs/log/debug.ml
edwintorok: That would require allocating a closure, the advantage of having these specialised 1 and 2 argument functions is that closure allocation can be avoided. Since these functions would be used in fast-paths having the optimized version is worhtwhile.

edwintorok: I'm going to build a switch with 'flambda' to test and compare, if there is a significant benefit in general we might consider testing and turning on flambda by default (it'll avoid having to do lots of these small optimizations by hand).
(Compilation times might increase, but only on Koji)

ocaml/xapi/locking_helpers.ml
```diff
@@ -23,28 +23,14 @@ module Thread_local_storage = struct
...
   end
 
-  module WeakSet = Weak.Make (Thread_key)
-  module WeakTable = Ephemeron.K1.Make (Thread_key)
+  module LiveThreads = Hashtbl.Make (Thread_key)
 
   (* While a thread is alive we keep some per-thread data,
      after the thread dies the data will be GC-ed.
      This is exactly what weak Ephemeron hashtables provide,
      they are not thread-safe themselves and require a global mutex
```
psafont, line 31: Needs to be changed now that ephemerons aren't used

ocaml/libs/log/debug.ml
edwintorok: I reverted `f590dc` and added '-O3' and '-unbox-closures' to both the benchmark's and xapi-internal's ocamlopt_flags and looked at the number of allocations.
Without modifying xapi_internal's flags I got 10 'minor-allocated', and when modifying it I got back ~5, which is what I get when *not* reverting the commit.

So it looks like 'flambda' can perform this optimization already on its own and it is worth-while to further investigate applying flambda to XAPI in general:

flambda -O3 -unbox-closures just on the benchmark:
```
╭────────────────────────────┬───────────────────────────┬───────────────────────────┬───────────────────────────╮
│name                        │  major-allocated          │  minor-allocated          │  monotonic-clock          │
├────────────────────────────┼───────────────────────────┼───────────────────────────┼───────────────────────────┤
│  Mutex/Mutex.execute       │             0.0000 mjw/run│             2.9412 mnw/run│            488.3654 ns/run│
│  Mutex/Mutex.lock/unlock   │             0.0000 mjw/run│             2.8302 mnw/run│            489.5659 ns/run│
│  Mutex/NamedMutex.execute  │             0.0000 mjw/run│            10.9412 mnw/run│           3268.9580 ns/run│
╰────────────────────────────┴───────────────────────────┴───────────────────────────┴───────────────────────────╯
```

flambda -O3 -unbox-closures also on xapi-internal:
```
╭────────────────────────────┬───────────────────────────┬───────────────────────────┬───────────────────────────╮
│name                        │  major-allocated          │  minor-allocated          │  monotonic-clock          │
├────────────────────────────┼───────────────────────────┼───────────────────────────┼───────────────────────────┤
│  Mutex/Mutex.execute       │             0.0000 mjw/run│             7.0612 mnw/run│            689.6918 ns/run│
│  Mutex/Mutex.lock/unlock   │             0.0000 mjw/run│             2.8302 mnw/run│            519.8271 ns/run│
│  Mutex/NamedMutex.execute  │             0.0000 mjw/run│             4.6316 mnw/run│           2682.6565 ns/run│
╰────────────────────────────┴───────────────────────────┴───────────────────────────┴───────────────────────────╯```

**Comment by edwintorok** [comment]
This appears to be fast enough now (even though about 5x as slow as a regular mutex, now the overhead is at the order of `214ns` which will be a lot smaller than the time spent with the lock held:
```
╭──────────────────────────────────────┬───────────────────────────┬───────────────────────────┬───────────────────────────╮
│name                                  │  major-allocated          │  minor-allocated          │  monotonic-clock          │
├──────────────────────────────────────┼───────────────────────────┼───────────────────────────┼───────────────────────────┤
│  Mutex/Mutex.execute                 │             0.3558 mjw/run│             4.0005 mnw/run│             64.7886 ns/run│
│  Mutex/Mutex.lock/unlock             │             0.0000 mjw/run│             0.0004 mnw/run│             40.9007 ns/run│
│  Mutex/NamedMutex.execute            │             1.2776 mjw/run│             2.0020 mnw/run│            214.0252 ns/run│
│  Mutex/NamedMutex.execute (tracing)  │            12.2949 mjw/run│           427.0162 mnw/run│           7181.2289 ns/run│
╰──────────────────────────────────────┴───────────────────────────┴───────────────────────────┴───────────────────────────╯
```

Some changes:
* disable 'stabilise_gc' feature of bechamel: it affects measurement results by a factor of about 10 (even though the time spent stabilising the GC should not be included in the measured time, it still is)
* Use `tsc` clocksource for Xen when measuring (in a VM, Dom0 should already use `tsc`, but VMs default to Xen due to migration)
* Optimise lookups between `waiting_for`, `acquired`, and `released`: it is sufficient to look up the thread local storage entry just once
* The Ephemeron I was using does actually allocate on each lookup, which is unavoidable, so have to use a regular hashtable here

There are still some tests I'd like to add here (and a lot more XenRT testing to be done), but I'm happy with the optimizations so far, so I'll clean up the commit history to reflect the final state

**Comment by edwintorok** [comment]
In relative terms it is costly (as you say 30x), but in absolute term it is still fairly small (7 microseconds), assuming that operations inside the lock take a lot longer (which should be the case for things like logging a user in, running SM operations, etc.).

We should look at optimizing tracing in the future, I have a 'zero_events' library in my load generator code that does 0 allocations while collecting data for tracing, but that relies on OCaml 5.1 feature in the standard library: runtime_events (although we could perhaps add a simplified implementation for that for OCaml 4.x based on a (big)array that acts as a ring buffer if it is useful).

**Comment by edwintorok** [comment]
I've rebased and cleaned up some commits (moving the benchmarks ahead in commit history, and folding in some of the benchmark bugfixes). There are still some things missing until this is ready to be merged:
* writing and adding some unit tests for locking helpers
* running a BVT/BST and then some stress tests

**Comment by edwintorok** [comment]
More testing: there are some deadlock issues with the finalizer, e.g.:
```
Responding with 500 Internal Error due to Sys_error("Mutex.lock: Resource deadlock avoided")
[debug||8360 /var/lib/xcp/xapi||http_internal_errors] Raised by primitive operation at Mutex in file \\"mutex.ml\\" (inlined), line 18, characters 0-44\
Called from Locking_helpers.Thread_local_storage.on_thread_gc in file \\"ocaml/xapi/locking_helpers.ml\\", line 56, characters 4-21\
Called from Locking_helpers.Thread_local_storage.with_lock in file \\"ocaml/xapi/locking_helpers.ml\\", line 47, characters 10-17\
Re-raised at Locking_helpers.Thread_local_storage.with_lock in file \\"ocaml/xapi/locking_helpers.ml\\", line 53, characters 8-42\
Called from Locking_helpers.Thread_local_storage.get in file \\"ocaml/xapi/locking_helpers.ml\\" (inlined), line 74, characters 4-44\
Called from Locking_helpers.Thread_state.get_states in file \\"ocaml/xapi/locking_helpers.ml\\" (inlined), line 211, characters 22-60\
Called from Locking_helpers.Thread_state.with_named_thread in file \\"ocaml/xapi/locking_helpers.ml\\" (inlined), line 214, characters 13-26\
Called from Server_helpers.exec_with_context in file \\"ocaml/xapi/server_helpers.ml\\", line 95, characters 2-780\
Called from Api_server.callback1 in file \\"ocaml/xapi/api_server.ml\\" (inlined), line 213, characters 19-51\
Called from Api_server.callback in file \\"ocaml/xapi/api_server.ml\\", line 251, characters 19-47\
Re-raised at Api_server.callback in file \\"ocaml/xapi/api_server.ml\\", line 283, characters 33-40\
Called from Xapi_http.add_handler.(fun) in file \\"ocaml/xapi/xapi_http.ml\\", line 350, characters 16-39\
Called from Http_svr.handle_one in file \\"ocaml/libs/http-lib/http_svr.ml\\", line 496, characters 8-32\
```

This can happen when we hold the lock and try to perform some operation on the global hashtable, but meanwhile a finalizer kicks in and it tries to take a lock (but it can't because the caller has it).
It looks like taking locks in finalizer in general is not very safe: you never know at which point you may run, and if you happen to run when the lock is already held you're doomed to fail.

(We could use try_lock there, but that is not reliable -- could be that another thread holds the lock and then what, we must complete the finalizer....)

I'll need to think about this more on how to best solve it and add some unit tests to trigger it.

PR #5086, opened 2023-07-03
scripts/plugins/extauth-hook-AD.py: Skip init logging on import

bernhardkaindl:
- Re-enable Python 2.7 CI using tiny docker containers(very fast)

- [scripts/plugins/extauth-hook-AD.py](https://github.com/xapi-project/xen-api/blob/master/scripts/plugins/extauth-hook-AD.py) has a main function, but it's [setup_logger()](https://github.com/xapi-project/xen-api/blob/master/scripts/plugins/extauth-hook-AD.py#L38) runs during import, which breaks importing it in tests in containers not providing /dev/log. Call setup_logger() only if the module is started as a script.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,18 +14,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.11"]
+        docker-image:  # See https://hub.docker.com/_/python/tags for images
+          - python:2.7.18-alpine3.11
+          - python:3.11.4-alpine3.18
+    container: ${{ matrix.docker-image }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
 
-      - name: Use python ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
-        with:
-          python-version: ${{ matrix.python-version }}
-
       - name: Install python 2 dependencies
-        if: ${{ matrix.python-version == 2.7 }}
+        if: ${{ startsWith(matrix.docker-image, 'python:2.7.18') }}
         run: pip install enum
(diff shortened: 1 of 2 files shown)


psafont (approved):

psafont:
Thanks!

lindig (approved):

bernhardkaindl:
I've got this comment from LinLiu, so I might try to update it in this way

If the setup_logger cause any problem, then it should be properly mocked in the unittest.

You can do something like this:
```py
diff --git a/scripts/plugins/test_extauth_hook_AD.py b/scripts/plugins/test_extauth_hook_AD.py
 index 71b5b7c95..6f26fa0b4 100644
 --- a/scripts/plugins/test_extauth_hook_AD.py
 +++ b/scripts/plugins/test_extauth_hook_AD.py
 @@ -4,14 +4,16 @@ Test module for extauth_hook_ad
  #pylint: disable=invalid-name
  import sys
  import os
 +import unittest
  from unittest import TestCase
 -from mock import MagicMock, patch
 +from unittest.mock import MagicMock, patch
  # mock modules to avoid dependencies
  sys.modules["XenAPIPlugin"] = MagicMock()
  sys.modules["XenAPI"] = MagicMock()
  # pylint: disable=wrong-import-position
  # Import must after mock modules
 -from extauth_hook_ad import StaticSSHPam, NssConfig, SshdConfig, UsersList, GroupsList
 +with unittest.mock.patch('extauth_hook_ad.setup_logger'):
 +    from extauth_hook_ad import StaticSSHPam, NssConfig, SshdConfig, UsersList, GroupsList
```

PR #5089, opened 2023-07-04
CA-375992: clean up previous sandbox when creating one

psafont:
On some cases where the sandbox may not have properly cleaned up the remains should not block from creating another sandbox for the same domain id

### DIFF
--- a/ocaml/xenopsd/lib/xenops_sandbox.ml
+++ b/ocaml/xenopsd/lib/xenops_sandbox.ml
@@ -12,7 +12,7 @@
  * GNU Lesser General Public License for more details.
  *)
 
-module D = Debug.Make (struct let name = "xenops_sandbox" end)
+module D = Debug.Make (struct let name = __MODULE__ end)
 
 module Chroot : sig
   (* can access fields, but can only be created through of_domid and create *)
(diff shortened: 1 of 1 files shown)


ocaml/xenopsd/lib/xenops_sandbox.ml
```diff
@@ -173,6 +176,10 @@ module Guard (G : GUARD) : SANDBOX = struct
 
   let create ~domid ~vm_uuid path =
     let chroot = chroot ~domid ~vm_uuid in
+    if Sys.file_exists chroot.root then (
```
lindig, line 179: Should this code be moved into the other `Chroot.create` function that creates the new one? Now we are checking the the presence of the directory in two places and react to it.

lindig (approved):

ocaml/xenopsd/lib/xenops_sandbox.ml
psafont: I'll add a parameter to chroot.create so the caller has to be specific whether the existing chroot will be deleted or not, so it can make a specific decision if need be

ocaml/xenopsd/lib/xenops_sandbox.ml
```diff
@@ -96,7 +96,19 @@ end = struct
...
+  let create ~clean chroot paths =
+    if Sys.file_exists chroot.root then
+      if clean then (
+        D.info "%s cleaning up existing chroot %s" __FUNCTION__ chroot.root ;
+        destroy chroot
+      ) else
+        D.info "%s: root %s already exists, continuing anyway" __FUNCTION__
+          chroot.root ;
     try
       Xenops_utils.Unixext.mkdir_rec chroot.root 0o755 ;
```
lindig, line 113: Would this fail if the directory exists? Such that this should be only called when it does not exist?

robhoes (approved):

PR #5090, opened 2023-07-04
CP-27910: allow exporting vm rrds and unarchives in json

psafont:
Until now only the general /rrd_updates and /host_rrds allowed this, which made
it awkward for clients to use the parameter in all endpoints

Tested manually, both json and xml works fro /vm_rrds and /host_rrds endpoints


ocaml/xcp-rrdd/bin/rrdd/rrdd_http_handler.ml
```diff
@@ -5,6 +5,21 @@ open Rrdd_shared
...
+let rrd_handler (req : Http.Request.t) (s : Unix.file_descr) = function
+  | None ->
+      Http_svr.headers s (Http.http_404_missing ())
+  | Some rrd ->
+      let json = List.mem_assoc "json" req.Http.Request.query in
+      let headers =
+        List.concat
+          [
+            Http.http_200_ok ~version:"1.0" ~keep_alive:false ()
+          ; (if json then [] else [Http.Hdr.content_type ^ ": text/xml"])
```
lindig, line 17: Should the content type not always be set?

lindig (approved):

ocaml/xcp-rrdd/bin/rrdd/rrdd_http_handler.ml
psafont: I'll add it, in a different commit since it's not being done in any endpoint, unlike text/xml

psafont:
I've tested the accept headers behaviour using firefox, it all looks good. To test without Accept headers I've used curl:

curl -v -H 'Accept:' "http://hostname.example/host_rrdd?session_id=OpaqueRef:99dfe4d0-ea6d-189a-71d6-42060a066068

ocaml/libs/http-lib/http.ml
```diff
@@ -449,73 +434,120 @@ let read_http_response_header buf fd =
...
+        false
+    | None, None ->
+        true
+    | Some (a, None), Some (b, None) ->
+        String.equal a b
+    | Some (a, Some a_s), Some (b, Some b_s) ->
+        if not (String.equal a b) then false else String.equal a_s b_s
+
+  let equal {ty= a_ty; q= a_q} {ty= b_ty; q= b_q} =
+    if not (Int.equal a_q b_q) then
```
lindig, line 459: This could be expressed with boolean logic alone

ocaml/libs/http-lib/http.ml
```diff
@@ -449,73 +434,120 @@ let read_http_response_header buf fd =
...
+          1
+      | None, None
+      | Some (_, None), Some (_, None)
+      | Some (_, Some _), Some (_, Some _) ->
+          0
+
+  (** [pick_preferred ~allowed accepted] filters [accepted] removing typed that
+      do not match [accepted] and orders them by client weighting *)
+  let pick_preferred ~allowed types =
+    (* Because compare_user_preference is insesible to the actual value of the
```
lindig, line 507: Would `List.stable_sort` help?

lindig (approved):
My overall impression is that there is a lot of branching logic in this patch - I would have hoped that finding the preferred response could have been more table or match driven. Good to see the unit tests.

ocaml/libs/http-lib/http.ml
psafont: The main issue is that the list of matches for each mimetype need to be merged in a way that deduplicates while maintaining the original sorting order.

Using stable sort doesn't help you in some cases when the merging happens because the stable sort happens on the merged list, not the original one, in other words, the algorithm is still sensible to the order of the list in `~allowed`

For example:
``` 
let allowed = ["text/xml"; "application/json"]
let types = "*/*;q=0.8,application/json,text/xml,text/html"
```
Because `text/xml` in the first item of `allowed`, its filtered list for compatible types filters out `application/json`, leaving the first item to be `text/xml`. This means that the merged list begins with `text/xml`, which stable sort will maintain: 
```
   Expected: `[application/json;q=1.000; text/xml;q=1.000; */*;q=0.800]'
   Received: `[text/xml;q=1.000; application/json;q=1.000; */*;q=0.800]'
```

psafont: Another alternative would be to accumulate items that do not match the allowed mimetypes, on the on merge keep the ones that are contained in all the lists, and finally remove those from the end list. But this seems more complex than keeping track of the indices of items that must be kept

lindig: What are the semantics that we need? Set operations on ordered lists? I'm not even sure what that means but I suspect that we would like some algebra over lists where order is meaningful. Would it make sense to write a module for that?

psafont: I've changed the algorithm by sorting the accept list by user preference then consuming elements from one priority at a time to select the content types that match them, returning the first non-empty result, or an empty one if all the list has been consumed.

ocaml/libs/http-lib/http.mli
```diff
@@ -54,15 +54,15 @@ module Accept : sig
...
-  val string_of_t : t -> string
+  val to_string : t -> string
 
-  val matches : string * string -> t -> bool
+  val matches : string -> t -> bool
 
-  val preferred_match : string * string -> t list -> t option
+  val preferred : from:string list -> t list -> string list
+  (** [preferred ~allowed accepted] returns the content types in [~allowed]
+      that are accepted by the most prioritary elements of [accepted] *)
```
robhoes, line 65:
```suggestion
  (** [preferred ~from accepted] returns the content types in [~from]
      that are accepted by elements of [accepted] in priority order *)
```

robhoes (approved):

PR #5092, opened 2023-07-06
CP-43728: Store in DB: if Toolstack restart is required after update

gangj:
Add field "toolstack_restart_required_after_update" to Host objects to
record if Toolstack restart is required after update.

This commit is a fixup for d2290d419f0ecdca28d2c35ca2169a4ca1197a61 
CP-42787: Ensure correct pool update sequence:

 let is_toolstack_restart_required ~__context self =
 \-  Db.Host.get_recommended_guidances ~__context ~self
 \-  |> List.mem `restart_toolstack
 \+  Db.Host.get_toolstack_restart_required_after_update ~__context ~self

As `restart_toolstack will be removed from recommended_guidances
if `reboot_host also exists in recommended_guidances.

### DIFF
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -2144,6 +2144,12 @@ let t =
             ~default_value:(Some (VEnum "unknown"))
             "Default as 'unknown', 'yes' if the host is up to date with \
              updates synced from remote CDN, otherwise 'no'"
+        ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:Bool
+            "toolstack_restart_required_after_update"
+            ~default_value:(Some (VBool false))
+            "True if this host has applied an update with RestartToolstack as \
+             one of its recommended guidance, set to False after Toolstack is \
+             restarted"
         ]
       )
     ()
--- a/ocaml/idl/schematest.ml
+++ b/ocaml/idl/schematest.ml
@@ -2,7 +2,7 @@ let hash x = Digest.string x |> Digest.to_hex
 
 (* BEWARE: if this changes, check that schema has been bumped accordingly in
    ocaml/idl/datamodel_common.ml, usually schema_minor_vsn *)
-let last_known_schema_hash = "3efd34e77e3d098653f4f0e1c89bae1d"
+let last_known_schema_hash = "c2f56456af7c1e9d3c45f5db9e16d06a"
 
 let current_schema_hash : string =
   let open Datamodel_types in
--- a/ocaml/tests/common/test_common.ml
+++ b/ocaml/tests/common/test_common.ml
@@ -208,7 +208,8 @@ let make_host2 ~__context ?(ref = Ref.make ()) ?(uuid = make_uuid ())
     ~multipathing:false ~uefi_certificates:"" ~editions:[] ~pending_guidances:[]
     ~tls_verification_enabled
     ~last_software_update:(Xapi_host.get_servertime ~__context ~host:ref)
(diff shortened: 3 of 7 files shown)


lindig:
Is this sufficiently general? Could this be more general with something like "pending_actions" or "outstanding_actions" per host and this could include other actions that have to be triggered? In general I believe accumulating outstanding actions leads to a situation where we don't understand the system behavior properly. But it is a consequence of using recommendations rather than enforcing these actions programmatically.

gangj:
> Is this sufficiently general? Could this be more general with something like "pending_actions" or "outstanding_actions" per host and this could include other actions that have to be triggered? In general I believe accumulating outstanding actions leads to a situation where we don't understand the system behavior properly. But it is a consequence of using recommendations rather than enforcing these actions programmatically.

@lindig Yes, I think I see your point, `toolstack_restart_required_after_update` seems one kind of "pending_actions". While we already have 2 pending action lists for Host obj: `recommended_guidances` and `pending_guidances`, they are the real pending actions for the Host, while `toolstack_restart_required_after_update` is not a real required action, it can also be cleared by a host reboot. And `toolstack_restart_required_after_update` is needed only to block the applying update on pool members during the update process, I think this might be a specific requirement.

robhoes:
I'd like to avoid new fields in the API for (almost) redundant data, if `recommended_guidances` already contains this information.

> while toolstack_restart_required_after_update is not a real required action, it can also be cleared by a host reboot

Why would `recommended_guidances` not be cleared on reboot?

robhoes:
I'm also not convinced that we need to block updates of other hosts until the coordinator has its xapi restarted...

gangj:
> I'd like to avoid new fields in the API for (almost) redundant data, if `recommended_guidances` already contains this information.
> 
> > while toolstack_restart_required_after_update is not a real required action, it can also be cleared by a host reboot
> 
> Why would `recommended_guidances` not be cleared on reboot?

`recommended_guidances` would also be cleared on reboot, I mean when there are 2 recommended guidance: "RestartToolstack" and "RebootHost", there will only be one item in the `recommended_guidances`: "RebootHost", that means there remains  only 1 required action: reboot host.

minglumlu:
> I'm also not convinced that we need to block updates of other hosts until the coordinator has its xapi restarted...

I'm impressed that it is a danger situation when a member host is newer than the coordinator host? (Features may be designed with an assumption that the coordinator has to be updated (and taken effect) firstly to solve compatibility issues etc.)

But I'm not sure how severely this could be, or whether the pool could be recovered easily from this situation.
If a pool is allowed to install updates on all hosts but not to get the new updates take effect, the longer the pool is kept in this case, it would be more probable to run into this situation, I.E. a member host rebooted unexpectedly.

gangj:
@robhoes , thank you for your suggestion!

Will try to have a fix following Rob's proposal: 
> For this particular case, rather than changing the update process, how about adding some protection to xapi's startup sequence to check the coordinator's xapi version in the "master connection", and block if the coordinator has a lower version? There is already a retry loop in that area.

Close this PR first and will try to have a fix following the new solution later.
And revert the related PR: https://github.com/xapi-project/xen-api/pull/5095

minglumlu:
> For this particular case, rather than changing the update process, how about adding some protection to xapi's startup sequence to check the coordinator's xapi version in the "master connection", and block if the coordinator has a lower version? There is already a retry loop in that area.

This sounds a wonderful solution. Thank you! Generally this unblocks the update process and protects XAPI pool from running into unpredictable situation.

PR #5094, opened 2023-07-10
CP-42019: Update wording for expiry message

gangj:
As the suggested wording from Katherine are:
"Your Customer Success Services (CSS) agreement is expiring soon."
and "Your Customer Success Services (CSS) agreement has expired.",
remove the leading "The".

### DIFF
--- a/ocaml/alerts/certificate/certificate_check.ml
+++ b/ocaml/alerts/certificate/certificate_check.ml
@@ -25,11 +25,11 @@ let get_certificates rpc session_id =
 
 let certificate_description = function
   | Host _ ->
-      "TLS server certificate"
+      "The TLS server certificate"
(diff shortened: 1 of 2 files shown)


liulinC (approved):

robhoes:
The unit test failed, because the change also changes the TLS cert alerts. I think the right thing to do is to change the code that raises the certificate alerts such that the resulting message text is the same as before.

gangj:
> The unit test failed, because the change also changes the TLS cert alerts. I think the right thing to do is to change the code that raises the certificate alerts such that the resulting message text is the same as before.

Sorry for the test failure! Changed.

psafont (changes requested):
Please squash the commtis

PR #5095, opened 2023-07-11
Revert "CP-42787: Ensure correct pool update sequence"

gangj:
This reverts commit d2290d419f0ecdca28d2c35ca2169a4ca1197a61.

Instead of blocking pool member update process, there is a better solution: during xapi's startup on pool member, check the coordinator's xapi version in the "master connection", and block if the coordinator has a lower version.

This revert commit removes the blocking during pool member update, will try to have a fix following the new solution later.


minglumlu (approved):

psafont (approved):
Please open the master connection check pull request against master branch, it's good to have it regardless

robhoes (approved):

PR #5002, opened 2023-05-16
CP-42019: Add messages for updates feature expiry.

gangj:
4 messages are added: UPDATES_FEATURE_EXPIRING_WARNING, UPDATES_FEATURE_EXPIRING_MAJOR, UPDATES_FEATURE_EXPIRING_CRITICAL, UPDATES_FEATURE_EXPIRED

PR #5096, opened 2023-07-11
Revert "CP-42019: Add messages for updates feature expiry. (#5002)"

gangj:
This reverts commit 3492322842a7c3eb33f05b8c29d86caf1f350312.

Will define CSS expiry related alerts in v6d, not in xapi.


psafont (approved):

kc284:
I've just noticed these were reverted because the SDK geneated after the merge to master does not have them. Unfortunately, XC relies on the messages being exposed in the SDK in order to handle them. We have the same problem with messages from SM too. But as a workaround I'll define them in the SDK.

PR #5097, opened 2023-07-11
CA-378778: Calculate host guidance correctly

gangj:
When there remains 'recommended_guidances' or 'pending_guidances' from previous update not applied for a host, the 'recommended_guidances' and 'pending_guidances' are not calculated correctly: the previous guidance should be taken into account to calculate the new 'recommended_guidances' and 'pending_guidances'.

### DIFF
--- a/ocaml/xapi/repository.ml
+++ b/ocaml/xapi/repository.ml
@@ -766,22 +766,27 @@ let apply_updates' ~__context ~host ~updates_info ~livepatches ~acc_rpm_updates
       ) ;
   (* Evaluate recommended/pending guidances *)
   let recommended_guidances, pending_guidances =
-    let recommended_guidances' =
+    let new_recommended_gs =
       (* EvacuateHost will be applied before applying updates *)
       eval_guidances ~updates_info ~updates:acc_rpm_updates ~kind:Recommended
         ~livepatches:successful_livepatches ~failed_livepatches
-        ~unapplied_guidances:
-          (Db.Host.get_recommended_guidances ~__context ~self:host)
       |> List.filter (fun g -> g <> Guidance.EvacuateHost)
     in
+    let recommended_guidances' =
+      merge_with_unapplied_guidances ~__context ~host ~kind:Recommended
+        ~guidances:new_recommended_gs
+    in
 
-    let pending_guidances' =
+    let new_pending_gs =
       eval_guidances ~updates_info ~updates:acc_rpm_updates ~kind:Absolute
         ~livepatches:[] ~failed_livepatches:[]
-        ~unapplied_guidances:
-          (Db.Host.get_pending_guidances ~__context ~self:host)
       |> List.filter (fun g -> not (List.mem g recommended_guidances'))
     in
+    let pending_guidances' =
+      merge_with_unapplied_guidances ~__context ~host ~kind:Absolute
+        ~guidances:new_pending_gs
+    in
+
     match failed_livepatches with
     | [] ->
         (* No livepatch should be applicable now *)
--- a/ocaml/xapi/repository.ml
+++ b/ocaml/xapi/repository.ml
@@ -765,7 +765,7 @@ let apply_updates' ~__context ~host ~updates_info ~livepatches ~acc_rpm_updates
       (* EvacuateHost will be applied before applying updates *)
       eval_guidances ~updates_info ~updates:acc_rpm_updates ~kind:Recommended
         ~livepatches:successful_livepatches ~failed_livepatches
-        ~previous_guidances:
+        ~unapplied_guidances:
           (Db.Host.get_recommended_guidances ~__context ~self:host)
       |> List.filter (fun g -> g <> Guidance.EvacuateHost)
     in
@@ -773,7 +773,8 @@ let apply_updates' ~__context ~host ~updates_info ~livepatches ~acc_rpm_updates
     let pending_guidances' =
       eval_guidances ~updates_info ~updates:acc_rpm_updates ~kind:Absolute
         ~livepatches:[] ~failed_livepatches:[]
-        ~previous_guidances:(Db.Host.get_pending_guidances ~__context ~self:host)
+        ~unapplied_guidances:
+          (Db.Host.get_pending_guidances ~__context ~self:host)
       |> List.filter (fun g -> not (List.mem g recommended_guidances'))
     in
     match failed_livepatches with
--- a/ocaml/xapi/repository.ml
+++ b/ocaml/xapi/repository.ml
@@ -641,12 +641,12 @@ let set_restart_device_models ~__context ~host ~kind =
   | `Running, true | `Paused, true -> (
     match kind with
     | Guidance.Absolute ->
-        Db.VM.add_pending_guidances ~__context ~self:ref
-          ~value:`restart_device_model ;
+        Db.VM.set_pending_guidances ~__context ~self:ref
+          ~value:[`restart_device_model] ;
         None
     | Guidance.Recommended ->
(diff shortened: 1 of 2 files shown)


ocaml/xapi/repository_helpers.ml
```diff
@@ -622,7 +622,7 @@ let append_livepatch_guidances ~updates_info ~upd_ids_of_livepatches guidances =
     upd_ids_of_livepatches guidances
 
 let eval_guidances ~updates_info ~updates ~kind ~livepatches ~failed_livepatches
-    =
+    ~previous_guidances =
```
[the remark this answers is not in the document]
gangj: > What does this parameter mean? that these guidances are pending, or that they were applied all, or it's a mix?

`~previous_guidances` means 'recommended_guidances' or 'pending_guidances' which are from previous update and have not been applied yet, so they are still in the host obj's fields: 'recommended_guidances' or 'pending_guidances'.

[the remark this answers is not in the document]
psafont: `pending_guidances` sounds like a better name for it, it's more specific than `previous_guidances`, please change it

ocaml/xapi/repository.ml
```diff
@@ -654,6 +654,10 @@ let set_restart_device_models ~__context ~host ~kind =
       None
 
 let set_pending_guidances ~__context ~host ~guidances =
+  let pending_guidances = Db.Host.get_pending_guidances ~__context ~self:host in
```
edwintorok, line 657: Would be better to add first and then remove, otherwise what if XAPI crashes in the middle here? We will have no guidances left, not old and not new (although in general XAPI's DB is not atomic to begin with...)

ocaml/xapi/repository_helpers.ml
```diff
@@ -634,6 +634,11 @@ let eval_guidances ~updates_info ~updates ~kind ~livepatches ~failed_livepatches
   in
   let upd_ids_of_livepatches = extract_upd_ids livepatches in
   let upd_ids_of_failed_livepatches = extract_upd_ids failed_livepatches in
+  let initial_gs =
+    List.fold_left
```
lindig, line 638: How about creating a set from a list first and joining the sets? Using `GuidanceSet.of_list` to create the set.

lindig (approved):

ocaml/xapi/repository_helpers.ml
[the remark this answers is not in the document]
gangj: > `pending_guidances` sounds like a better name for it, it's more specific than `previous_guidances`, please change it

Ah, yes, the name `previous_guidances` is not clear enough, how about change it to: "unapplied_guidances", as `pending_guidances` is used to represent the "absolute_guidance" from update now.

ocaml/xapi/repository_helpers.ml
gangj: @lindig , thanks for the suggestion~ But to create a set with `GuidanceSet.of_list`, we need to transfer from "update_guidance" to "Guidance" (with `Guidance.of_update_guidance`) first.

ocaml/xapi/repository.ml
minglumlu: The `diff` operation could get `to_be_removed` and `to_be_added`. So only the minimum changes would be applied on the DB.

ocaml/xapi/repository_helpers.ml
```diff
@@ -622,7 +622,7 @@ let append_livepatch_guidances ~updates_info ~upd_ids_of_livepatches guidances =
     upd_ids_of_livepatches guidances
 
 let eval_guidances ~updates_info ~updates ~kind ~livepatches ~failed_livepatches
-    =
+    ~unapplied_guidances =
```
minglumlu, line 625: I would suggest not to add this parameter in `eval_guidances`. The problem to be solved is to merge two guidance sets which could be considered as:
1. carry on union of two sets
2. perform `GuidanceSet.resort_guidances` on the union.

ocaml/xapi/repository.ml
minglumlu: Not sure if `Db.Host.set_pending_guidances` directly would work?

ocaml/xapi/repository_helpers.ml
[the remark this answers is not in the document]
psafont: that's fine, yes

ocaml/xapi/repository.ml
gangj: Yes, I think it works, fixed with [d197e5d](https://github.com/xapi-project/xen-api/pull/5097/commits/d197e5d17764235997fc70e4aa619076039edc3b), there is a similar usage here: 
https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/dbsync_slave.ml#L95-L96
`virtual_hardware_platform_versions` is also of type Set: https://github.com/xapi-project/xen-api/blob/master/ocaml/idl/datamodel_host.ml#L2085

ocaml/xapi/repository_helpers.ml
gangj: I had a try with your suggestion: https://github.com/xapi-project/xen-api/commit/50f3c30e476f25af76e1efca9c088221683ba788
But I would prefer to keep the current version: fix by adding the `unapplied_guidances` parameter to `eval_guidances`, as otherwise there seems duplicated codes:
https://github.com/gangj/xen-api/blob/private/gangj/CA-378778.2/ocaml/xapi/repository_helpers.ml#L650-L651
https://github.com/gangj/xen-api/blob/private/gangj/CA-378778.2/ocaml/xapi/repository.ml#L775-L784
https://github.com/gangj/xen-api/blob/private/gangj/CA-378778.2/ocaml/xapi/repository.ml#L792-L801

Or could it be fixed in a better way?

minglumlu: I think the duplication could be eliminated by a helper function simply:
```
let merge_guidances ~host ~kind ~guidances =
  let open GuidanceSet in
  (match kind with
  | Abosulte ->
      Db.Host.get_pending_guidances ~__context ~self:host
  | Recommended ->
      Db.Host.get_recommended_guidances ~__context ~self:host
  )
  |> List.map (fun g -> Guidance.of_update_guidance g)
  |> of_list
  |> union (of_list guidances)
  |> resort_guidances ~kind
  |> elements
```
The point is the `eval_guidances` is complicated already. And the change in this PR seems not necessary to be coded in the function.

gangj: OK, changed by adding helper function `merge_with_unapplied_guidances`

ocaml/xapi/repository.ml
```diff
@@ -657,38 +657,55 @@ let set_pending_guidances ~__context ~host ~guidances =
...
+        | RebootHost ->
+            Some `reboot_host
+        | RestartToolstack ->
+            Some `restart_toolstack
+        | _ ->
+            None
+        )
+      guidances
+  in
+  Db.Host.set_recommended_guidances ~__context ~self:host ~value:gs
```
minglumlu, line 708: How about this for the block from line#686 to line#708?
```
guidances
|> List.fold_left (fun acc g
     | RebootHost ->
         `reboot_host :: acc
     | RebootHostOnLivePatchFailure ->
         `reboot_host_on_livepatch_failure :: acc
     | RestartToolstack ->
         `restart_toolstack :: acc
     | RestartDeviceModel ->
         set_restart_device_models ~__context ~host ~kind:Absolute ;
         acc
) []
|> (fun gs -> Db.Host.set_pending_guidances ~__context ~self:host ~value:gs)
```

minglumlu (approved):

PR #5099, opened 2023-07-11
CA-379472 add debugging to redo_log

lindig:
We could add some more debugging for the redo log - which in general is disabled from `xapi.conf` because it creates too much logging.


robhoes:
Have you tested that this doesn't cause excessive logging?

lindig:
I have checked a xenrt test case and it seems fine to me - 150 extra lines

edwintorok (approved):

PR #5103, opened 2023-07-13
CP-43916: Revert xapi-expiry-alerts from feature branch

gangj:
Revert xapi-expiry-alerts related change, as they have already been merged into master, and the commits which have already been merged into master are slightly different from those in the feature branch, as there are some review fixup later when merging into master.

I think the steps to refine this feature branch can be:
1. Revert any commits which have already been merged into master (this will make the below step "merge master into this feature branch" easier)
2. Merge the "current master" into this feature branch. (so that the xapi-expiry-alerts related changes in master are available), this will be the last change for the "current feature branch".
3. Prepare a new feature branch, based on the same "current master", by "cherry-pick"ing all the CP and CA commits from the "current feature branch" (filter out the revert commits, and the merge master commits).
4. Diff the "current feature branch" and "new feature branch", make sure there is not any code change difference.
5. Then I think finally the "new feature branch" can be forcefully pushed into the feature branch: feature/stream-updates, there will have no code change diff for this PR, only code commit history change.

And this PR is the first step.

Can we refine the feature branch this way? Would you please give your suggestions, thank you!


gangj:
OK, will try that.
As this changes the history of the feature branch, I think the final PR will be a merge request from the "new feature branch" to "master" branch, right?
But as the code of "xapi-expiry-alerts" in the feature branch is slightly different from the code in master branch, then there will be a diff between the "new feature branch" and "current feature branch", it will be that difference.

gangj:
Oh, I think I am still a little confused, I think without step1 and step 2, then the diff between the final "new feature branch" which is to be merged into master in the final PR and the "current feature branch" will contains 2 part:
1. the slight difference of "xapi-expiry-alerts"
2. the later merged commits in master, which has not been merged into "current feature branch" yet.
Then how to verify the diff easily?

gangj:
Not needed now.

PR #5106, opened 2023-07-14
CA-380032: Remove the key if it exists before adding it

gangj:
This is a fix for 91104b3a15d07008ff96451a9b469ffd020f6eee:
CP-43799: Add a flag to show if an update sync is in progress

### DIFF
--- a/ocaml/xapi/xapi_pool.ml
+++ b/ocaml/xapi/xapi_pool.ml
@@ -3385,6 +3385,18 @@ let remove_repository ~__context ~self ~value =
     (Db.Pool.get_repositories ~__context ~self) ;
   Db.Pool.remove_repositories ~__context ~self ~value
 
+let sync_with_yum_repos_in_progress = "sync_with_yum_repos_in_progress"
+
+let update_sync_status ~__context ~pool ~value =
+  if
+    List.mem_assoc sync_with_yum_repos_in_progress
+      (Db.Pool.get_other_config ~__context ~self:pool)
+  then
+    Db.Pool.remove_from_other_config ~__context ~self:pool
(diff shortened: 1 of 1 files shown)


robhoes:
See my comment on the other PR (unfortunately already merged): other_config is not right for this use case.

ocaml/xapi/xapi_pool.ml
```diff
@@ -3385,6 +3385,18 @@ let remove_repository ~__context ~self ~value =
...
   Db.Pool.remove_repositories ~__context ~self ~value
 
+let sync_with_yum_repos_in_progress = "sync_with_yum_repos_in_progress"
+
+let update_sync_status ~__context ~pool ~value =
+  if
+    List.mem_assoc sync_with_yum_repos_in_progress
+      (Db.Pool.get_other_config ~__context ~self:pool)
+  then
+    Db.Pool.remove_from_other_config ~__context ~self:pool
```
minglumlu, line 3395: What will be returned from `Db.Pool.remove_from_other_config` if the key doesn't exist? An exception?
Checking it firstly like this won't guarantee the key is there when doing the removal.
Remove it directly and catch the exception, or ignore anything it returned.

minglumlu: If the kv pair with same key can be added multiple times, will the removing get only one removed when there are multiple pairs already in it?

gangj:
> See my comment on the other PR (unfortunately already merged): other_config is not right for this use case.

Thank you, will close this PR and revert the code change for "sync_with_yum_repos_in_progress": https://github.com/xapi-project/xen-api/pull/5107

PR #5108, opened 2023-07-14
fix CI, opam: sync with latest metadata; use ocaml setup v2

psafont:
Changes some download links to stop using the git protocol, updates some dependencies on new libraries and changes the conf-python dependency to depend on python 3

Also changes the ocaml setup action to v2, as this is the one that's maintained and should give less problems than the v1 that's in use currently


robhoes (approved):

robhoes (approved):
Nice to see that package list go away from the workflow.

edwintorok (approved):

PR #5110, opened 2023-07-14
Test dune cache behaviour

psafont:


psafont:
In the rerun the compilation was really slow, not sure why was that. The dependency compilation was faster, as well as the setup phase, and the errors are gone

PR #5111, opened 2023-07-14
Cross-pool live migration: move CPU check to the target host

robhoes:
This required adding VM_metrics to the metadata export/import.


psafont (approved):

lindig (approved):
Just from a review it's difficult to say if this works but no obvious problems for me.

PR #5114, opened 2023-07-17
CA-380043: VM recommended guidances is not set correctly

gangj:
This fixes an issue introduced in 47bfe89e:
CA-378778: Calculate host guidance correctly
where the "recommended_guidances" was set into "pending_guidances" by mistake.

### DIFF
--- a/ocaml/xapi/repository.ml
+++ b/ocaml/xapi/repository.ml
@@ -653,49 +653,40 @@ let set_restart_device_models ~__context ~host ~kind =
       (* No device models are running for this VM *)
       None
 
-let set_pending_guidances ~__context ~host ~guidances =
+let set_guidances ~__context ~host ~guidances ~db_set ~kind =
   let open Guidance in
   guidances
   |> List.fold_left
        (fun acc g ->
-         match g with
-         | RebootHost ->
+         match (g, kind) with
+         | RebootHost, _ ->
              `reboot_host :: acc
-         | RebootHostOnLivePatchFailure ->
-             `reboot_host_on_livepatch_failure :: acc
-         | RestartToolstack ->
+         | RestartToolstack, _ ->
              `restart_toolstack :: acc
-         | RestartDeviceModel ->
-             set_restart_device_models ~__context ~host ~kind:Absolute ;
+         | RestartDeviceModel, _ ->
+             set_restart_device_models ~__context ~host ~kind ;
              acc
-         | g ->
+         | RebootHostOnLivePatchFailure, Absolute ->
+             `reboot_host_on_livepatch_failure :: acc
+         | _, Absolute ->
              warn "Unsupported pending guidance %s, ignoring it."
                (Guidance.to_string g) ;
              acc
-       )
-       []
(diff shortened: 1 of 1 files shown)


minglumlu (approved):

edwintorok (commented):
There is another difference here: set_pending_guidances handles 'RebootHostOnLivePatchFailure' and the other one does not.

It would be better if these 2 functions shared more code and instead have one common function that receives the `kind` as a labeled argument and the DB function to use, to be used like this:
```ocaml
let set_recommended_guidances ~__context ~host ~guidances =
  set_common_guidances ~__context _host ~guidances ~kind:Recommended ~db_set:Db.Host.set_recommended_guidances

let set_pending_guidances ~__context ~host ~guidances =
  set_common_guidances ~__context _host ~guidances ~kind:Absolute ~db_set:Db.Host.set_pending_guidances
```

Assuming that 'kind = Absolute' is what you meant to use for 'pending', I don't know if that is correct (if not replace it with the correct kind)

edwintorok:
What happens if someone tries to apply the recommended guidance, has RebootHostOnLivePatchFailure set as guidance, and the livepatch fails. I think unsupported guidances completely is wrong: unsupported guidances should be treated the same as the max guidance (RebootHost), that is the only way we can introduce new guidances in the future.

Otherwise if you're updating from a sufficiently old XAPI it will ignore the guidance and not do anything which is wrong, because perhaps it should've done something, it just wasn't new enough to realise what. However RebootHost should always get everythingu updated, so that is a safe fallback.

And then in fact you won't have to special case the RebootHostOnLivePatchFailure if you don't want to: it'll do the right thing and fall back to RebootHost too.

lindig (approved):

robhoes (approved):

robhoes:
I agree that this fixes the bug, so this PR is an improvement. But I also agree with Edwin, that an unknown guidance should be translated to "host reboot".

gangj:
Thanks for the review and suggestion!
"Translate" unknown guidance to "host reboot" should be a safe design when we have new guidance later. To do that, I think we might need to change the code a little bit to add some type named "UnknownGuidance" to "Guidance" and do the translation when parse the update in somewhere maybe in "eval_guidance_for_one_update". @minglumlu , what's your thought about this, please help to have a look, thank you!

minglumlu:
The update process is to notify a user that if a host will be rebooted or not before applying update. If the user found the host was rebooting automatically and unexpectedly after applying updates, it would be confusing.
Indeed the case is valid when a pool has not been updated for a long time (a sufficiently old XAPI). But we don't know what is a `valid` unknown/new guidance. Falling back to rebooting host always may be abused due to bugs in updateinfo metadata from repos.
I agree to open a ticket for this potential improvement. Let's merge the bug fix firstly.

edwintorok:
If you've waited long enough there'll inevitably be a kernel or Xen update that requires a host reboot. Avoiding a host reboot is an optimization *if* we know that is the safe thing to do, if we don't know it is safe then we shouldn't make that optimization. But yes lets have another PR for that.

PR #5117, opened 2023-07-18
CA-379928: enable more logging for redo_log_usage

edwintorok:
Recovery from redolog is critical for the correct operation of XAPI (with HA or with the redolog enabled). Do not hide these messages by default: redo_log_usage should have its own log prefix (redo_log is disabled by default).

Also do not eat exceptions: instead use log and ignore. This preserves the previous behaviour where we allow XAPI to start up even if redo log recovery fails (with possibly an older an incorrect database).

### DIFF
--- a/ocaml/xapi/redo_log_usage.ml
+++ b/ocaml/xapi/redo_log_usage.ml
@@ -21,96 +21,94 @@ exception DatabaseWrongSize of int * int
 
 let read_from_redo_log log staging_path db_ref =
   R.log_and_ignore_exn @@ fun () ->
-    (* 1. Start the process with which we communicate to access the redo log *)
-    R.debug "Starting redo log" ;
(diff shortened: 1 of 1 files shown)


ocaml/xapi/redo_log_usage.ml
```diff
@@ -11,7 +11,7 @@
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU Lesser General Public License for more details.
  *)
-module R = Debug.Make (struct let name = "redo_log" end)
+module R = Debug.Make (struct let name = "redo_log_usage" end)
```
lindig, line 14: Use `__MODULE__`?

lindig (approved):

psafont (approved):

PR #5119, opened 2023-07-18
Allow a user to select on which SR to run quicktest

benjamreis:
- New `-sr` option which takes a SR uuid to run the tests on the option is mutually exclusive with `-default-sr`

- If the SR provided is not available on the host, fails

### DIFF
--- a/ocaml/quicktest/qt_filter.ml
+++ b/ocaml/quicktest/qt_filter.ml
@@ -180,7 +180,20 @@ module SR = struct
            ~self:pool
         )
         ()
-    else
+    else if !A.sr <> "" then (
+      let sr =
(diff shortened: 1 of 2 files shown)


psafont (approved):
This seems OK to have, shouldn't interfere with the usage in the internal xenserver tests

benjamreis:
We'd like to run `quicktest` on several storage drivers without having to set the default SR everytime nor running it on ALL SRs ^^

I'll test the patch and report if it behaves as expected.

edwintorok (approved):

benjamreis:
This has been tested and validated and will soon be used in our ci.

PR #5120, opened 2023-07-18
CA-377945: toolstack restart: ensure xapi is stopped first, started last

robhoes:
Stopping xapi appears to take a long time for something to time out if other services have stopped before it. The assumption in xe-toolstack-restart seems to be that `systemctl stop` with multiple service names on the same command line respects the `After:` constraints of the services, but I am not convinced it does. Explicitly stopping xapi first and starting it after everything else has started, makes the whole script execute about a minute faster.

### DIFF
--- a/scripts/xe-toolstack-restart
+++ b/scripts/xe-toolstack-restart
@@ -27,7 +27,7 @@ echo "Executing $FILENAME"
 
 POOLCONF=`cat @ETCXENDIR@/pool.conf`
 if [ $POOLCONF == "master" ]; then MPATHALERT="mpathalert"; else MPATHALERT=""; fi
-SERVICES="perfmon xapi v6d xenopsd xenopsd-xc xenopsd-xenlight
+SERVICES="perfmon v6d xenopsd xenopsd-xc xenopsd-xenlight
   xenopsd-simulator xenopsd-libvirt xcp-rrdd-iostat xcp-rrdd-squeezed
   xcp-rrdd-xenpm xcp-rrdd-gpumon xcp-rrdd xcp-networkd squeezed forkexecd
   $MPATHALERT xapi-storage-script xapi-clusterd varstored-guard message-switch"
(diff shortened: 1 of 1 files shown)


psafont (approved):

rosslagerwall:
From what I can see in the systemd code, `systemctl <op> <unit1> <unit2> ...` will execute each operation in a sequence in the order given on the command-line, not as part of a single transaction. So I _think_ this change has only effectively moved XAPI from after to before perfmon.

I'm not sure how this helps generally though... in the normal case xe-toolstack-restart only takes a few seconds and if the system is in a bad state and some services that XAPI depends on have already been stopped when xe-toolstack-restart is run and XAPI is stopped, then surely it will still hang?

It's a larger change so might not need to be done now, but I think the preferred way to do this rather than encoding a list of services and ordering in this script would be to define a `toolstack.target`. Each toolstack service file can then declare to be part of the toolstack target. The xe-toolstack-restart script could then just stop and start the toolstack target and those actions would be part of a multi-unit transaction so the orderings declared in the service file should be respected.

robhoes:
Thanks Ross, that confirms my suspicion.

I think the problem with this script, and why the change I made actually makes a difference, is that the list of services is reversed before stopping, due to `TO_RESTART="$svc $TO_RESTART"`. We want it reversed for starting, but not for stopping.

I agree with you that a systemd target is a better solution overall. This script feels too much like a big hack at the moment.

robhoes:
>I'm not sure how this helps generally though... in the normal case xe-toolstack-restart only takes a few seconds and if the system is in a bad state and some services that XAPI depends on have already been stopped when xe-toolstack-restart is run and XAPI is stopped, then surely it will still hang?

Turns out it was usually around a minute now. I think it is to do with something timing out eventually, so it is not permanently blocked.

edwintorok (approved):

edwintorok:
I've opened an internal ticket about the systemd target cleanup

PR #5121, opened 2023-07-19
CP-40651 Validate the login username and provide friendly error message

LunfanZhang:
Add a check for an empty username before performing local or external authentication to prevent passing an empty username to the authentication server.

Signed-off-by: Lunfan Zhang luzhan@tibco.com

### DIFF
--- a/ocaml/idl/datamodel_errors.ml
+++ b/ocaml/idl/datamodel_errors.ml
@@ -177,7 +177,7 @@ let _ =
     () ;
 
   (* Session errors *)
-  error Api_errors.session_authentication_failed []
+  error Api_errors.session_authentication_failed ["username"; "msg"]
     ~doc:
       "The credentials given by the user are incorrect, so access has been \
        denied, and you have not been issued a session handle."
--- a/ocaml/xapi/xapi_session.ml
+++ b/ocaml/xapi/xapi_session.ml
@@ -257,6 +257,15 @@ let xapi_internal_originator = "xapi"
 
 let serialize_auth = Mutex.create ()
 
+let assert_username uname = 
+  match uname with
+  | "" ->
+      raise
+        Api_errors.(
+          Server_error
+            (session_authentication_failed, [""; "Username is required to login"])
+        )
+
 let wipe_string_contents str =
   for i = 0 to Bytes.length str - 1 do
     Bytes.set str i '\000'
@@ -766,6 +775,7 @@ let login_with_password ~__context ~uname ~pwd ~version:_ ~originator =
             ~auth_user_sid:"" ~auth_user_name:uname ~rbac_permissions
             ~db_ref:None ~client_certificate:true
       | None -> (
+          let () = assert_username uname in
           let () =
             if Pool_role.is_slave () then
               raise
--- a/ocaml/xapi/xapi_session.ml
+++ b/ocaml/xapi/xapi_session.ml
@@ -257,14 +257,18 @@ let xapi_internal_originator = "xapi"
 
 let serialize_auth = Mutex.create ()
 
-let assert_username uname = 
+let assert_username uname =
   match uname with
   | "" ->
       raise
         Api_errors.(
           Server_error
-            (session_authentication_failed, [""; "Username is required to login"])
+            ( session_authentication_failed
+            , [""; "Username is required to login"]
+            )
         )
+  | _ ->
+      ()
 
 let wipe_string_contents str =
   for i = 0 to Bytes.length str - 1 do
--- a/ocaml/tests/test_session.ml
+++ b/ocaml/tests/test_session.ml
@@ -6,10 +6,17 @@ let future = Date.of_string "2020-09-22T15:03:13Z"
 
 let fail_login ~__context ~uname ~originator ~now () =
   try
+    let user_name = Option.get uname in
     Xapi_session._record_login_failure ~__context ~now ~uname ~originator
       ~record:`log_and_alert (fun () ->
         if Random.bool () then
-          raise Api_errors.(Server_error (session_authentication_failed, [uname, "Authentication failed"]))
+          raise
+            Api_errors.(
+              Server_error
+                ( session_authentication_failed
+                , [user_name; "Authentication failed"]
+                )
+            )
         else
           raise (Auth_signature.Auth_failure "Auth failure")
     )
--- a/ocaml/tests/test_session.ml
+++ b/ocaml/tests/test_session.ml
@@ -9,7 +9,7 @@ let fail_login ~__context ~uname ~originator ~now () =
     Xapi_session._record_login_failure ~__context ~now ~uname ~originator
       ~record:`log_and_alert (fun () ->
         if Random.bool () then
-          raise Api_errors.(Server_error (session_authentication_failed, []))
+          raise Api_errors.(Server_error (session_authentication_failed, [uname, "Authentication failed"]))
         else
           raise (Auth_signature.Auth_failure "Auth failure")
     )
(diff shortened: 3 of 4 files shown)


liulinC:
Describe the problem in the commit message, other than refer to CP-40651.

liulinC:
The `preauth` return some value only when the socket fd with path `/var/lib/xcp/xapi` or `/var/lib/xcp/xapi-clientcert`, 
thus means this two does not requires username while all others do, does this meet the expectation? 
(This should meet the expectation at my first sight, @robhoes  would you like to confirm?)
```119 let preauth ~__context =
120   match __context.origin with
121   | Internal ->
122       None
123   | Http (_, s) -> (
124     match Unix.getsockname s with
125     | Unix.ADDR_UNIX path when path = Xapi_globs.unix_domain_socket ->
126         Some `root
127     | Unix.ADDR_UNIX path when path = Xapi_globs.unix_domain_socket_clientcert
128       ->
129         Some `client_cert
130     | Unix.ADDR_UNIX _ | Unix.ADDR_INET _ ->
131         None
132   )
```

ocaml/xapi/xapi_session.ml
```diff
@@ -1195,6 +1196,13 @@ let local_logout ~__context =
   let session_id = Context.get_session_id __context in
   Xapi_local_session.destroy ~__context ~self:session_id
 
+let check_uname uname =
```
liulinC, line 1199: Better to rename as `assert_uname`

ocaml/xapi/xapi_session.ml
```diff
@@ -1195,6 +1196,13 @@ let local_logout ~__context =
   let session_id = Context.get_session_id __context in
   Xapi_local_session.destroy ~__context ~self:session_id
 
+let check_uname uname = 
+  match uname with
+  | "" ->
+    raise(Api_errors.Server_error
+              ( Api_errors.internal_error, "user name is empty!"))
```
liulinC, line 1203: "user name is empty!" -> "Username is required to login" ?

ocaml/xapi/xapi_session.ml
```diff
@@ -1195,6 +1196,13 @@ let local_logout ~__context =
   let session_id = Context.get_session_id __context in
   Xapi_local_session.destroy ~__context ~self:session_id
 
+let check_uname uname = 
+  match uname with
+  | "" ->
+    raise(Api_errors.Server_error
```
liulinC, line 1202: is `failwith` better in this case as it is not due to server error, but less of argument ?

ocaml/xapi/xapi_session.ml
LunfanZhang: All the comments have been fixed in the latest commit, Thanks.

ocaml/xapi/xapi_session.ml
```diff
@@ -1195,6 +1196,12 @@ let local_logout ~__context =
   let session_id = Context.get_session_id __context in
   Xapi_local_session.destroy ~__context ~self:session_id
 
+let assert_username uname = 
+  match uname with
+  | "" ->
+    failwith ("Username is required to login")
```
robhoes, line 1202: It would be better to raise an API error here. The error `session_authentication_failed` looks suitable. So:

    raise (Api_errors.(Server_error (session_authentication_failed, []))

ocaml/xapi/xapi_session.ml
robhoes: @liulinC The exception `Api_errors.Server_error` is used for all API errors. Even `failwith`, which ends up raising a standard OCaml `Failure` exception, is eventually re-raised as a `Server_error` with `Api_errors.internal_error`.

The issue here is that this should not be an "internal error", because the server is not at fault here, but the client has given a bad value. In such cases, we want to raise a specific API error; see my other comment.

ocaml/xapi/xapi_session.ml
LunfanZhang: How about  `raise (Api_errors.(Server_error (session_authentication_failed, ["Username is required to login"])))`   which will deliver an explicit error message to the client. CC @liulinC

ocaml/xapi/xapi_session.ml
```diff
@@ -257,6 +257,11 @@ let xapi_internal_originator = "xapi"
 
 let serialize_auth = Mutex.create ()
 
+let assert_username uname = 
+  match uname with
+  | "" -> raise (Api_errors.(Server_error (session_authentication_failed, ["Username is required to login"])))
```
edwintorok, line 262:
```suggestion
  | "" -> raise (Api_errors.(Server_error (session_authorization_failed, ["", "Username is required to login"]))) 
```

'session_authentication_failed' is not declared to take any parameters in 'datamodel_errors.ml', but 'session_authorization_failed' does (it takes a username and an error message as parameter).

robhoes: But this is an authentication error, not an authorization error...

robhoes: I think we need to update the error definition in datamodel_errors.ml to take two arguments: the username and a message, as actually most of current cases where this error is raise (several in the same file), these arguments are actually included.

ocaml/xapi/xapi_session.ml
robhoes: See below: we need to consider the error definition in datamodel_errors.ml.

ocaml/xapi/xapi_session.ml
LunfanZhang: In my opinion, an empty string should not be a valid username, and it will be beneficial to prevent empty string from being passed to authentication server and response with a friendly error message. 
I agree with @robhoes , updating error definition in datamode_errors.ml is good solution.

liulinC: @lindig Yes, you are right, empty username does not always fail, like certificate authentication.

The background of this PR is to short circuit the process when try to authenticate to external auth (which is AD domain), which flood AD.  And empty username absolutely have no meaning for Domain Controller.

minglumlu: Hi @LunfanZhang Can you please resolve the build error firstly?

ocaml/xapi/xapi_session.ml
```diff
@@ -766,6 +775,7 @@ let login_with_password ~__context ~uname ~pwd ~version:_ ~originator =
             ~auth_user_sid:"" ~auth_user_name:uname ~rbac_permissions
             ~db_ref:None ~client_certificate:true
       | None -> (
+          let () = assert_username uname in
```
minglumlu, line 778: Would `assert_username_valid uname ;` be better?

psafont: It's a matetr of style, both are fine

LunfanZhang: @minglumlu Thanks for the reminder.  I have updated and fixed format issue, please review again.

ocaml/idl/datamodel_errors.ml
```diff
@@ -177,7 +177,7 @@ let _ =
     () ;
 
   (* Session errors *)
-  error Api_errors.session_authentication_failed []
+  error Api_errors.session_authentication_failed ["username"; "msg"]
```
psafont, line 180: Please update all the users of the error to use the new parameters, I found at least two out of date ones:
- https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_http.ml#L264
- https://github.com/xapi-project/xen-api/blob/master/ocaml/tests/test_session.ml#L12

minglumlu:
Hi @LunfanZhang  Some unit tests look broken. Can you fix them please? You can run the unit tests locally.

lindig:
This has been going of for quite a while. I suggest this is either progressed or move to Draft.

minglumlu:
Hi @LunfanZhang 
I set this PR as draft. Please feel free to resume the review once you have progress on it.

lindig:
This should be progressed or closed. A PR with a long and stale history is difficult to review.

lindig:
No action - closing. Please re-open once this has been updated.

PR #5122, opened 2023-07-19
Introduce LRU Cache and re-implement Stunnel Cache

lindig:
The Stunnel cache implementation is so convoluted that I have little hope finding a race condition that was discovered in a stress test. I am proposing to introduce a general LRU cache and to use that to re-implement the Stunnel cache. This is currently being tested and I am opening a draft PR for transparency.

### DIFF
--- a/lru.opam
+++ b/lru.opam
@@ -1,0 +1,22 @@
+opam-version: "2.0"
+maintainer: "xen-api@lists.xen.org"
+authors: "xen-api@lists.xen.org"
+homepage: "https://xapi-project.github.io/"
+bug-reports: "https://github.com/xapi-project/xen-api.git"
+dev-repo: "git+https://github.com/xapi-project/xen-api.git"
+build: [[ "dune" "build" "-p" name "-j" jobs ]]
+
+available: [ os = "linux" ]
+depends: [
+  "ocaml"
+  "dune"
+  "threads"
+]
+synopsis: "Library required by xapi"
+description: """
+These libraries are provided for backwards compatibility only.
+No new code should use these libraries."""
+url {
+  src:
+    "https://github.com/xapi-project/xen-api/archive/master.tar.gz"
+}
--- a/lru.opam
+++ b/lru.opam
@@ -10,7 +10,7 @@ available: [ os = "linux" ]
 depends: [
   "ocaml"
   "dune"
-  "base-threads"
+  "threads.posix"
 ]
 synopsis: "Library required by xapi"
 description: """
--- a/lru.opam
+++ b/lru.opam
@@ -10,7 +10,7 @@ available: [ os = "linux" ]
 depends: [
   "ocaml"
   "dune"
-  "threads"
+  "base-threads"
 ]
 synopsis: "Library required by xapi"
 description: """
--- a/ocaml/libs/lru-cache/lru.ml
+++ b/ocaml/libs/lru-cache/lru.ml
@@ -1,0 +1,291 @@
+(*
+ * This module implements a cache with support for a least-recently-used
+ * (LRU) replacement policy. Main features:
+ *
+ * Implemented with standard OCaml data types, no dependency on outside
+ * libararies for the main functionality.
+ *
+ * Implemented with mutable state. This keeps the implementation compact
+ * and efficient but requires thinking about state more.
+ *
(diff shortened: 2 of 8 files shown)


lru.opam
```diff
@@ -10,7 +10,7 @@ available: [ os = "linux" ]
 depends: [
   "ocaml"
   "dune"
-  "threads"
+  "base-threads"
```
psafont, line 13:
```
  "threads.posix"
```

psafont:
If the library does not need to be exposed, it can be kept private, and the test binaries be part of the main xapi package

ocaml/libs/lru-cache/lru.ml
```diff
@@ -0,0 +1,294 @@
...
+    ( match np with
+    | None ->
+        t.first <- nn
+    | Some x ->
+        x.next <- nn ;
+        n.prev <- None
+    ) ;
+    match nn with
+    | None ->
+        t.last <- np
```
edwintorok, line 113: assert with phys equality that t.last was equal to 'n'.

ocaml/libs/lru-cache/lru.ml
```diff
@@ -0,0 +1,294 @@
...
+ *)
+
+let invalid_arg fmt = Printf.kprintf invalid_arg fmt
+
+let _fail fmt = Printf.kprintf failwith fmt
+
+module LL : sig
+  (** Doubly linked list ['a t] holding elements of type ['a]. *)
+
+  (** doubly linked list; this is a cyclic data structure; don't use [=]
```
edwintorok, line 25: It *can* be implemented as a cyclic data structure, but the current one doesn't *seem* to be: previous of first is None and next of last is None. Probably best to avoid using '=' anyway though.

ocaml/libs/lru-cache/lru.ml
```diff
@@ -0,0 +1,294 @@
...
+      is returned and [false] otherwise. It signals the caller to [trim]
+      the table.*)
+  let add' t key value =
+    match lookup' t key with
+    | None ->
+        let node = LL.node (key, value) in
+        Hashtbl.add t.table key node ;
+        t.entries <- t.entries + 1 ;
+        LL.append t.queue node ;
+        t.entries > t.cap
```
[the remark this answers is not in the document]
lindig: If we store anything in the LRU that requires handling before it can be removed, the client must be given an opportunity to do that. Hence I avoid removing elements when capacity is exceeded. I don't care so much about temporarily exceeding the capacity.

ocaml/libs/lru-cache/lru.mli
```diff
@@ -0,0 +1,127 @@
...
+val trim : ('a, 'b) t -> unit
+(** [trim] the cache by removing least-used elements until the size does
+   not exceed the capacity. No finalisation of elements that are
+   removed. Use [drop_while] for custom finalisation. *)
+
+val to_list : ('a, 'b) t -> ('a * 'b) list
+(** retrieve all elements as a list. The head of the list is in LRU
+    order: the least-used entry is at the head of the list. *)
+
+module LL : sig
```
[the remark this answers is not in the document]
lindig: Yes, exposed for unit testing.

ocaml/libs/stunnel/stunnel_cache.ml
```diff
@@ -1,250 +1,110 @@
...
+
+(** drop entries until we are no longer exceeding capacity *)
+let trim cache =
+  let evict (k, v) = function
+    | true ->
+        debug "%s: expiring %s:%d from cache" __FUNCTION__ k.host k.port ;
+        Stunnel.disconnect v.stunnel ;
+        true
+    | false ->
+        debug "%s: cache holds %d entries" __FUNCTION__ (LRU.size cache) ;
```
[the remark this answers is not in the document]
lindig: So far have not implemented it. But in general more criteria could be used; the guiding principle is that we visit elements in LRU order.

ocaml/libs/lru-cache/lru.ml
lindig: This is cyclic: you have a cycle between neighbouring elements but not between first/last. The cycle is the reason that `=` must not be used.

lindig:
I am closing this until I have figured out the interaction with the `Safe_resources` module.

PR #5123, opened 2023-07-19
Offload VM CPU Policy checks to Xen

andyhhp:
In order to support MSR_ARCH_CAPS, combining policlies becomes more complicated than a simple intersection.  More generally, it is more conviencient to have all levelling/compatibility logic in a single codebase to avoid needing linked changes in Xapi.

Where possible, offload the decisions into Xen.

### DIFF
--- a/ocaml/xenopsd/c_stubs/xenctrlext_stubs.c
+++ b/ocaml/xenopsd/c_stubs/xenctrlext_stubs.c
@@ -452,6 +452,96 @@ CAMLprim value stub_xenctrlext_cputopoinfo(value xch_val)
 	CAMLreturn(result);
 }
 
+/*
+ * Convert an Ocaml int64 array to a C uint32_t array, zero extending as
+ * necessary.
+ */
+static void ocaml_int64_array_to_c_array(value o, uint32_t *c, mlsize_t c_len)
+{
+	mlsize_t i, o_len = caml_array_length(o);
+
+	for (i = 0; i < o_len; ++i)
+		c[i] = Int64_val(Field(o, i));
+	for (; i < c_len; ++i)
+		c[i] = 0;
+}
+
+#ifndef MAX
+#define MAX(a, b) ((a) > (b) ? (a) : (b))
+#endif
+
+__attribute__((weak))
+void xc_combine_cpu_featuresets(
+	const uint32_t *p1, const uint32_t *p2, uint32_t *out, size_t len);
+
+/* int64 array (p1) -> int64 array (p2) -> int64 array (new) */
+CAMLprim value stub_xenctrlext_combine_cpu_featuresets(value p1, value p2)
+{
+	CAMLparam2(p1, p2);
+	CAMLlocal1(result);
+
+	mlsize_t p1_len = caml_array_length(p1);
+	mlsize_t p2_len = caml_array_length(p2);
+	mlsize_t len = MAX(p1_len, p2_len);
+	mlsize_t i;
+
+	uint32_t c_p1[len], c_p2[len], c_out[len];
+
+	if (!xc_combine_cpu_featuresets)
+		raise_unix_errno_msg(ENOSYS, "xc_combine_cpu_featuresets");
+
+	if (len == 0)
+		CAMLreturn(Atom(0));
+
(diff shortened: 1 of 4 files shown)


MarkSymsCtx:
s/conviencient /convienient / in the commit

robhoes (approved):

andyhhp:
Fixed both spelling mistakes.  (policlies too)

ocaml/xenopsd/c_stubs/xenctrlext_stubs.c
```diff
@@ -452,6 +452,96 @@ CAMLprim value stub_xenctrlext_cputopoinfo(value xch_val)
...
+ * necessary.
+ */
+static void ocaml_int64_array_to_c_array(value o, uint32_t *c, mlsize_t c_len)
+{
+	mlsize_t i, o_len = caml_array_length(o);
+
+	for (i = 0; i < o_len; ++i)
+		c[i] = Int64_val(Field(o, i));
+	for (; i < c_len; ++i)
+		c[i] = 0;
```
edwintorok, line 466: This won't allow you to differentiate between: this entire feature leaf was not present before and got zero-extended, vs it was present before, but with a value of `0`. Especially with some features having opposite polarity "not present before" could actually mean "it was acting as 1 before", and zero-extending here would be wrong. Or do you intend to switch the polarity internally for any new bits that behave that way?

For backwards compatibility calculations the difference may matter at some point (and OCaml could make this distinction by looking at whether the array is empty or not).

Also do you have to special case the entirely zero-length featureset in case some OCaml code does a 'fold' with a starting value of an empty array? (otherwise that would result in the entire pool getting all zeros everywhere, and I think this was special cased before in the OCaml code), or is that already special cased in the caller?

ocaml/xenopsd/c_stubs/xenctrlext_stubs.c
```diff
@@ -452,6 +452,96 @@ CAMLprim value stub_xenctrlext_cputopoinfo(value xch_val)
...
+	ocaml_int64_array_to_c_array(vm, c_vm, len);
+	ocaml_int64_array_to_c_array(host, c_host, len);
+
+	err = xc_cpu_featuresets_are_compatible(c_vm, c_host, len, msg);
+
+	if (!err)
+		result = Val_none;
+	else {
+		result = caml_alloc_small(1, Tag_some);
+		Store_field(result, 0, caml_copy_string(err));
```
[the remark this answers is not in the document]
andyhhp: yes

ocaml/xenopsd/xc/xenops_server_xen.ml
```diff
@@ -1127,24 +1127,49 @@ module HOST = struct
 
   let combine_cpu_policies policy1 policy2 =
     let open Cpuid in
-    intersect
+    let combine p1 p2 =
```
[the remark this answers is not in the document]
andyhhp: I'll leave the Ocaml to others.  (I didn't write this bit of the patch)

ocaml/xenopsd/c_stubs/xenctrlext_stubs.c
edwintorok: Discussed that the special fold case doesn't exist anymore and there is already a CPUID bit that can be used to solve the potential tri-state problem with the additional features (0,1, #GP).

edwintorok (approved):

ocaml/xenopsd/c_stubs/xenctrlext_stubs.c
andyhhp: The zero extending is intentional, to avoid needing memory allocations in the library function, and therefore to avoid ENOMEM failures.

The new logic is not a case of having opposite polarity (I was hoping that was to be the case, but alas).

I could in principle pass `p1_orig_len` and `p2_orig_len` in which would allow distinguishing these cases, but I don't have any use them right now and I'm not sure it would be useful in the future either.

Current VMs don't see the ARCH_CAPS CPUID bit (and MSR_ARCH_CAPS will #GP on access).  There's no room for a VM to see an intermediate state with ARCH_CAPS visible but RSBA/RRSBA incorrectly calculated.

PR #5124, opened 2023-07-22
CA-380368: Replaced &lt; ad &gt; with < and >. Added messages raised by v6 and SM.

kc284:


lindig (approved):

lindig:
We have all kind of problems in the CI that are unrelated to the code change. Is this a glitch or or do we have changes in xs-opam causing this?

psafont:
It's an issue with the base images used by github, the system repositories are not up-to-date and this makes downloading some outdated packages to fail. The setup-ocaml action somehow ignores the error and continues, making the action fail while trying to install some of the dependencies

psafont (approved):

PR #5125, opened 2023-07-24
ci: update base image

psafont:
Previous image was starting to need workaround because it wasn't updated anymore

Set to draft to test the change

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   python-test:
     name: Python tests
-    runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -59,6 +59,9 @@ jobs:
         id: dotenv
         uses: falti/dotenv-action@v1.0.4
 
+      - name: Update Ubuntu repositories
+        run: sudo apt-get update
```
lindig, line 63: Do we need an `apt-get upgrade`, too?

lindig (approved):

edwintorok (approved):

PR #5127, opened 2023-07-25
Add `secureboot-enforce` setting in `xapi.conf`

benjamreis:
This setting will be used by varstored to know
wheter to allow the start of a VM that has no certificates when secureboot is enabled.

Default: false to keep the previous behavior.

See: https://github.com/xapi-project/varstored/pull/19

### DIFF
--- a/ocaml/xapi/xapi_globs.ml
+++ b/ocaml/xapi/xapi_globs.ml
@@ -998,6 +998,8 @@ let prefer_nbd_attach = ref false
 (** 1 MiB *)
 let max_observer_file_size = ref (1 lsl 20)
 
+let secureboot_enforce = ref false
+
 let xapi_globs_spec =
   [
     ( "master_connection_reset_timeout"
@@ -1469,6 +1471,11 @@ let other_options =
     , Arg.Set_int max_observer_file_size
     , (fun () -> string_of_int !max_observer_file_size)
     , "The maximum size of log files for saving spans"
+    )
+  ; ( "secureboot-enforce"
+    , Arg.Set secureboot_enforce
+    , (fun () -> string_of_bool !secureboot_enforce)
+    , "Do not start a VM with no SB certificates if secureboot is set to on"
     )
   ]
 
--- a/ocaml/xapi/xapi_vm.ml
+++ b/ocaml/xapi/xapi_vm.ml
@@ -286,16 +286,23 @@ let assert_memory_constraints ~__context ~vm platformdata =
 *)
 
 let update_platform_secureboot ~__context ~self platform =
-  match List.assoc "secureboot" platform with
-  | exception Not_found ->
-      platform
-  | "auto" ->
-      ( "secureboot"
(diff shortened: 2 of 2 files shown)


benjamreis:
TBD:
- `secureboot-enforce` is it the correct naming?
- for now it only writes when creating a VM, what to do with already existing ones?

ocaml/xapi/xapi_globs.ml
```diff
@@ -1470,6 +1473,11 @@ let other_options =
     , (fun () -> string_of_int !max_observer_file_size)
     , "The maximum size of log files for saving spans"
     )
+  ; ( "secureboot-enforce"
+    , Arg.Set secureboot_enforce
+    , (fun () -> string_of_bool !secureboot_enforce)
+    , "Do not start a VM with no SB certificates if secureboot is set to on"
```
psafont, line 1479: I think using "block" creates the expectations that the system will return an error in this situation
```suggestion
    , "Block secureboot VMs from starting when the secureboot certificates are not valid"
```

psafont (commented):
I think this setting makes more sense to be applied to all the VMs at the same time instead of per-VM. Can't xenopsd enforce this setting when setting up varstored?

ocaml/xapi/xapi_globs.ml
lindig: `Block a secureboot VM from starting when its secureboot certificates are not valid` - the singular is usually easier to understand; we are obviously talking about the general case and it would apply to all such VMs implicitly.

benjamreis:
It's my understanding that going through the xenstore via a VM platform is the standard way of passing information to xenstore (like a VM secureboot certificates) it was discussed in an issue linked in the varstored PR.

ocaml/xapi/xapi_vm.ml
```diff
@@ -286,16 +286,23 @@ let assert_memory_constraints ~__context ~vm platformdata =
...
-      platform
-  | "auto" ->
-      ( "secureboot"
-      , string_of_bool (Db.Pool.get_uefi_certificates ~__context ~self <> "")
-      )
-      :: List.remove_assoc "secureboot" platform
-  | _ ->
-      platform
+  let platform =
+    match List.assoc "secureboot" platform with
```
lindig, line 290: `List.assoc_opt` would work as well

The code below could benefit from a function to update an association list:

```
let update (k,v) assoc = (k,v) :: List.remove_assoc k assoc
```

lindig (approved):

lindig:
Can we progress or close this?

stormi:
As the decision also involves varstored, we asked there a few days ago. To move forward we now need someone at XenServer to choose one of the two proposed solutions.

https://github.com/xapi-project/varstored/issues/16#issuecomment-1689523789

stormi:
I made a mistake, explained here: https://github.com/xapi-project/varstored/pull/19

This PR is not necessary anymore.

PR #5129, opened 2023-07-25
CA-380551: bump minimum HA SR size to 4GiB

edwintorok:
Commit 00cb7f2fa changed the default (recommended) size of the redo log to 4GiB, while keeping the minimum at 256MiB for backwards compatibility with existing VDIs.

However:
* HA must be disabled/re-enabled across updates anyway so backwards compatibility is not needed
* the redo-log on its own is known not to work (bug), keeping backwards compatibility here is not useful
* it makes it more difficult for XenCenter to tell the user when they chose an SR that is too small

If the SR is greater than minimum but below recommended then from the CLI you could still enable HA and create a redo log of the appropriate size, but not from XenCenter (where the size cannot be chosen).

We could expose a new RO pool field with the minimum size and update clients to use that, but that is unnecessary complication for a very rare use case.

Revert the 'recommended_size' introduction, and instead bump the minimum size. The existing 'SR.assert_can_host_statefile' can then report an error when the size is wrong and XenCenter can appropriately grey out the choice for the user with the correct message. (And this method works also for other reasons why we might reject an SR).

Fixes: 00cb7f2fa ("redo-log: bump default size to 4GiB")

### DIFF
--- a/ocaml/database/redo_log.ml
+++ b/ocaml/database/redo_log.ml
@@ -49,7 +49,7 @@ let mib megabytes =
   Int64.of_int megabytes ** 1024L ** 1024L
 
 (* Make sure we have plenty of room for the database *)
-let minimum_vdi_size, recommended_vdi_size = (mib 256, mib 4096)
+let minimum_vdi_size = mib 4096
 
 let redo_log_sm_config = [("type", "raw")]
 
--- a/ocaml/database/redo_log.ml
+++ b/ocaml/database/redo_log.ml
@@ -48,8 +48,17 @@ let mib megabytes =
   let ( ** ) = Int64.mul in
   Int64.of_int megabytes ** 1024L ** 1024L
 
-(* Make sure we have plenty of room for the database *)
-let minimum_vdi_size = mib 4096
+(* Make sure we have plenty of room for the database
+   There is also a 4MiB statefile, so make the sum be 4GB, which is easier to document.
+*)
+let minimum_vdi_size =
+  let ( // ) = Int64.div and ( ** ) = Int64.mul and ( -- ) = Int64.sub in
+  let align = mib 4 in
+  ((4_000_000_000L // align)
+  -- 2L
+     (* -2 because we also need room for a statefile, and an 'empty' SR seems to have a utilization of 4MiB *)
+  )
+  ** align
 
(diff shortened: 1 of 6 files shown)


lindig (approved):

psafont (approved):

edwintorok:
I finally got around to testing this, and I needed to add 1 more commit (+previous refactoring commit), apparently assert_can_host_statefile really just checked for enough space for the statefile, and didn't check for enough space for redolog.
There *is* code in xapi to check for enough space for the metadata VDI (but that is different from the redo log VDI...).

So I've added some more code to check that when we want to chose an SR to host a statefile we have enough space for both the statefile and the redo log (or there is already a preexisting statefile and/or redolog that we can reuse).

Given SRs of sizes 1GiB, 4GiB, and 5GB XenCenter now successfully greys out the small SRs, leaving just the 5GB one available, without any changes required in XenCenter:
Screenshot 2023-08-14 at 15 59 02

I think this PR is now ready.

psafont (approved):
Seems sensible to me

lindig:
We discussed today that the required space is the total of redo-log and state file. It sounded like this is not accounted for so far.

lindig:
I think the issue is we have two calls:
* ` ha_fits_sr ~__context ~what:"statefile" ~sr ~minimum_size ~typ:`ha_statefile`
* ` ha_fits_sr ~__context ~what:"redo-log" ~sr ~minimum_size:Redo_log.minimum_vdi_size ~typ:`redo_log`

But the SR needs to fit both.

lindig:
The easiest change is probably passing the sum of of both sizes to both calls.

edwintorok:
After some manual testing this seems to work now. Although there are still various bugs about whether GB means power of 2 or power of 10, both in XenRT and XenCenter, so I've made the SR minimum size be 4GB instead of 4GiB.
That way it works with both (SR size of 4GB and 4GiB).

I'll need to run some more tests.

lindig:
This needs a reformat; maybe do this and undraft? @edwintorok

edwintorok:
Reformatted, and the tests I've run worked (or failed for unrelated reasons, and a rerun worked). So I think this is ready (pending the CI here becoming green).

PR #5130, opened 2023-07-26
ci: work around outdated ubuntu repositories on all workflows

psafont:
The documentation push failed because the ubuntu repos are not up-to-date in the ubuntu image used in the CI, work around this on all the remaining workflows.

Also update base image and actions, use the latest available. There's no need to pin any packages when checking the format, so skip that step

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   ocaml:
     name: Docs
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     env:
       XAPI_VERSION: "v0.0.0-${{ github.sha }}"
       STORAGE_DOCDIR: .gh-pages-xapi-storage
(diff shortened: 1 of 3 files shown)


edwintorok (approved):

.github/workflows/docs.yml
```diff
@@ -7,7 +7,7 @@ on:
 jobs:
   ocaml:
     name: Docs
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
lindig, line 10: Why not `ubuntu-latest` here?

lindig (approved):

.github/workflows/docs.yml
psafont: testing and building the toolstack is more fragile with regards to distribution updates, this is because the official opam repository protects us from unstability in the packages they maintain (like ocamlformat), but not our own.

For example to update to ubuntu 22.04 we had to change the python testing, because python 2 is simply not available.

PR #5131, opened 2023-07-27
CA-379350: Use up-to-date vTPM UUIDs when creating device models

psafont:
When creating a domain from an snapshot, the UUID of the vTPM as used by xapi
is changed as a new vTPM object is used.
Diagram of introduced behavior and in which cases these happen with analysis
below, it's assumed that xapi updates xenops metadata before doing any
lifecycle VM operation so the second column is always up-to-date with the
current xapi model.

<table>
<tr>
 <td>
 <td>extra vtpm
 <td>VM vtpm
 <td>result
<tr>
 <td>1.
 <td>None
 <td>None
 <td>None
<tr>
 <td>2.
 <td>None
 <td>Some a
 <td>Some a
<tr>
 <td>3.
 <td>Some a
 <td>None
 <td>Some a
<tr>
 <td>4.
 <td>Some a
 <td>Some b
 <td>Some b
<tr>
 <td>5.
 <td>Some a
 <td>Some a
 <td>Some a
</table>

1. Normal case when vtpm is unplugged, OK
2. This case happens in case of a xapi error, where it allowed hotplugging a
   vtpm device into a snapshot. Same with other resume cases, OKj
3. This case happens in case of a xapi error, where it allowed hot unplugging a
   vtpm in a snapshot. Same with other resume cases, OK
4. This is the expected result for snapshots. In other resume cases,
   hotplugging is disallowed by xapi, so this case shouldn't happen elsewhere.
5. Normal case when vtpm is plugged, OK

This fix is in the first commit only, the second commit where changes I made to understand and read the code better while looking for the issue

### DIFF
--- a/ocaml/xenopsd/lib/xenops_server.ml
+++ b/ocaml/xenopsd/lib/xenops_server.ml
@@ -760,11 +760,7 @@ module VUSB_DB = struct
       remove id
 end
 
-module StringMap = Map.Make (struct
-  type t = string
-
-  let compare = compare
-end)
+module StringMap = Map.Make (String)
 
 let push_with_coalesce should_keep item queue =
   (* [filter_with_memory p xs] returns elements [x \in xs] where [p (x_i,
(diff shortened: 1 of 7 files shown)


lindig:
Maybe it's worth encoding the table using pattern matching to make the connection with the comment more explicit, rather than using `Option.*` combinators.

lindig (approved):

ocaml/xenopsd/xc/xenops_server_xen.ml
```diff
@@ -2391,7 +2391,9 @@ module VM = struct
           in
           let () =
             on_domain
```
edwintorok: Didn't you change 'on_domain' to take 'task vm' first?

ocaml/xenopsd/xc/xenops_server_xen.ml
```diff
@@ -2036,15 +2036,15 @@ module VM = struct
                   Device.Dm.Disabled
             in
             let parallel = List.assoc_opt "parallel" vm.Vm.platformdata in
+            let tpm = Option.fold ~none:hvm_info.tpm ~some:Option.some vtpm in
```
edwintorok, line 2007: Is this the same as `Option.value ~default:hvm_info.tpm vtpm`? I'd find that easier to read than the fold.

ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -2223,7 +2223,13 @@ let rec perform_atomic ~progress_callback ?subtask:_ ?result (op : atomic)
       let vifs : Vif.t list = VIF_DB.vifs id in
       let vgpus : Vgpu.t list = VGPU_DB.vgpus id in
       let vusbs : Vusb.t list = VUSB_DB.vusbs id in
-      B.VM.create_device_model t (VM_DB.read_exn id) vbds vifs vgpus vusbs
+      let vtpm : Vm.tpm option =
+        (* The uuid changes when snapshots are restored *)
+        let id = try Some (VM_DB.read_exn id) with _ -> None in
+        Option.bind id @@ fun id ->
+        match id.Vm.ty with HVM {tpm; _} -> tpm | _ -> None
+      in
```
edwintorok, line 2227: Would be useful to add a line here that logs what the outcome of this logic was (e.g. the uuid of the 'vtpm' if not None) so we can more easily see in the logs where it comes from.

ocaml/xenopsd/xc/xenops_server_xen.ml
edwintorok: Also would be useful to add a log line here that shows the final outcome of the 'tpm' object (e.g. its uuid it it exists)

edwintorok (commented):
Looks good, some minor comments

ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -2223,7 +2223,13 @@ let rec perform_atomic ~progress_callback ?subtask:_ ?result (op : atomic)
...
+      B.VM.create_device_model t (VM_DB.read_exn id) vbds vifs vgpus vusbs vtpm
```
robhoes, line 2228: Why can't the backend (xenops_server_xen.ml) itself take the vtpm out of the passed in `vm` parameter? It has a helper function for that already (`vtpm_of ~vm`), which is used in many similar functions. It can be used where currently `hvm_info.tpm` is used in `create_device_model_config`.

That way, we don't need an interface change that essentially adds an parameter with redundant data.

ocaml/xenopsd/xc/xenops_server_xen.ml
psafont: > Is this the same as Option.value ~default:hvm_info.tpm vtpm?
Not exactly, we don't want to unwrap the value, which is what `Option.value` does.
This case can be thought of as an if is none, give a default option value, otherwise leave it unchanged

ocaml/xenopsd/lib/xenops_server.ml
psafont: `vtpm_of ~vm` is limited to what the backend can access. The vm value it has access to ultimately comes from the internal DB module: https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/xc/xenops_server_xen.ml#L2376

This internal module has only access to the xenopsd/classic/extra directory (https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/xc/xenops_server_xen.ml#L188), and not the the xenopsd/classic/VM one.

I find it less invasive to add another device to the call that has already several devices passed as parameters, rather than enabling a new module to give the backend access to the information as the latter seems a breach of the self-imposed responsibilities while the former doesn't

robhoes: I'm not sure what exactly you mean, because the second argument passed to this function here is the vm record from `VM_DB.read_exn id`, which is the same thing you unpack to construct `vtpm` above.

psafont: I see what you mean, I missed that parameter in all the parameter soup

ocaml/xenopsd/xc/xenops_server_xen.ml
```diff
@@ -1798,28 +1772,27 @@ module VM = struct
           )
     )
 
-  let pause =
-    on_domain (fun xc _ _ _ di ->
+  let pause t di =
+    on_domain t di (fun xc _ _ _ di ->
```
robhoes, line 1776: This is confusing: `on_domain` takes a `vm` arg, not a `di`.

ocaml/xenopsd/xc/xenops_server_xen.ml
```diff
@@ -1798,28 +1772,27 @@ module VM = struct
...
         if di.Xenctrl.total_memory_pages = 0n then
           raise (Xenopsd_error Domain_not_built) ;
         Domain.pause ~xc di.Xenctrl.domid
     )
 
-  let unpause =
-    on_domain (fun xc _ _ _ di ->
+  let unpause t di =
+    on_domain t di (fun xc _ _ _ di ->
```
robhoes, line 1783: Here also.

ocaml/xenopsd/xc/device.ml
```diff
@@ -3696,7 +3698,8 @@ module Dm = struct
       | Restore ->
           qemu_args ~xs ~dm info true domid
     in
-    debug "Device.Dm.start domid=%d args: [%s]" domid
+    debug "Device.Dm.start domid=%d restore=%s qemu args: [%s]" domid
```
robhoes, line 3701: This would print `restore=Start qemu`. I think we don't want `restore=`.

psafont: I'll change it to print `Device.Dm.start domid=XX action=Start qemu args:`. It's not ideal because of the repetition, but should be used for debugging only, so it makes sense to look for action there

robhoes (approved):

PR #5132, opened 2023-07-27
CA-380580: cross-pool migration: no CPU checks for halted VMs

robhoes:
CPU checks are needed only for running VMs that are being migrated, to check for compatibility with the remote-host's CPUs.

### DIFF
--- a/ocaml/xapi/xapi_vm_migrate.ml
+++ b/ocaml/xapi/xapi_vm_migrate.ml
@@ -1182,6 +1182,7 @@ let migrate_send' ~__context ~vm ~dest ~live:_ ~vdi_map ~vif_map ~vgpu_map
      We look at the VDIs of the VM, the VDIs of all of the snapshots, and any
      suspend-image VDIs. *)
   let vm_uuid = Db.VM.get_uuid ~__context ~self:vm in
+  let power_state = Db.VM.get_power_state ~__context ~self:vm in
   let vbds = Db.VM.get_VBDs ~__context ~self:vm in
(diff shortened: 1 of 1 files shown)


psafont (approved):

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -1456,7 +1457,7 @@ let migrate_send' ~__context ~vm ~dest ~live:_ ~vdi_map ~vif_map ~vgpu_map
             in
             inter_pool_metadata_transfer ~__context ~remote ~vm ~vdi_map
               ~vif_map ~vgpu_map ~dry_run:false ~live:true ~copy
-              ~check_cpu:(not force)
+              ~check_cpu:((not force) && power_state <> `Halted)
```
edwintorok, line 1460: how about 'Suspended' and 'Paused'? In theory you can move that around freely until you resume (at which point you do the CPU check).
Should the condition here we `power_state = 'Running` instead?

robhoes: This is an odd use case... The checks take into account that one may want to start/resume/unpause/shutdown the VM while it is being migrated to another pool. This is why the checks are always done with `~live:true`, which checks for available memory on the target host, even if the VM is halted. The CPU checks would be similar, except that you can't actually do it for a halted VM.

lindig (approved):

lindig:
Is this risking moving a halted VM to a host where it can't be started later?

PR #5133, opened 2023-07-27
CA-375992 remove stale swtpm chroots after boot

lindig:
When a host crashes, chroots in /var/lib/xcp/run remain. Their names use the domain counter, which resets on reboot. This risks name collisions between a new chroot and an existing old one. After reboot we can be sure that no domain is running and that it is safe to remove stale chroots.


psafont (commented):
This doesn't deal with the domid's wrapping around, which could be an issue for long-running servers.

Not how rare this situation will become, is it a problem for varstored sandboxes as well? If so it seems pretty rare

lindig:
The domid wrapping around is not addressed here, indeed. This only addresses that we don't clean up chroots after a host crash. We could address the wrap-around by using a different naming scheme but that is orthogonal because those would not be cleaned up either.

psafont (approved):

PR #5136, opened 2023-07-31
CA-380581: Remove lock on downloading updates from remote repos

minglumlu:
The 'reposync_mutex' was designed for cases in which a pool is downloading updates from remote repos, but another XAPI call is being carried on to apply updates simultaneously. The assumption of this design is that the downloading may take long time. During this period, it would be nice to allow other operations, e.g get_updates, or apply_updates, etc.

The issue to be fixed in this commit is that the locked status of 'reposync_mutex' can't be populated to clients since the status of the lock is not recorded in XAPI DB.

Additionally, with 80c8153860, only the minimum set of update packages will be downloaded. The assumption above now doesn't hold anymore in most cases.

The 'pool.allowed_operations' now can take the chance to replace 'reposync_mutex' to protect consistency among repo updates related operations. As a result, this will populate the locked status to clients since it is recorded in XAPI DB.

Note: a data change in XAPI DB can be sent/retrieved to/by clients through API calls.

### DIFF
--- a/ocaml/idl/datamodel_errors.ml
+++ b/ocaml/idl/datamodel_errors.ml
@@ -1890,8 +1890,6 @@ let _ =
   error Api_errors.repository_already_exists ["ref"]
     ~doc:"The repository already exists." () ;
   error Api_errors.repository_is_in_use [] ~doc:"The repository is in use." () ;
-  error Api_errors.reposync_in_progress []
-    ~doc:"The pool is syncing with the enabled remote YUM repository." () ;
   error Api_errors.repository_cleanup_failed []
     ~doc:"Failed to clean up local repository on coordinator." () ;
   error Api_errors.no_repository_enabled []
--- a/ocaml/xapi-consts/api_errors.ml
+++ b/ocaml/xapi-consts/api_errors.ml
@@ -1237,8 +1237,6 @@ let repository_already_exists = "REPOSITORY_ALREADY_EXISTS"
 
 let repository_is_in_use = "REPOSITORY_IS_IN_USE"
 
-let reposync_in_progress = "REPOSYNC_IN_PROGRESS"
-
 let repository_cleanup_failed = "REPOSITORY_CLEANUP_FAILED"
 
 let no_repository_enabled = "NO_REPOSITORY_ENABLED"
--- a/ocaml/xapi/repository.ml
+++ b/ocaml/xapi/repository.ml
@@ -23,8 +23,6 @@ module UpdateIdSet = Set.Make (String)
 
 let capacity_in_parallel = 16
 
-let reposync_mutex = Mutex.create ()
-
 (* The cache below is protected by pool's current_operations locking mechanism *)
 let updates_in_cache : (API.ref_host, Yojson.Basic.t) Hashtbl.t =
   Hashtbl.create 64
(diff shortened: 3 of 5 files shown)


psafont:
I would prefer a consistent way of using the pool operation. It's not clear what would be the state when a client removes or adds a yum repository while a download is in progress.

While clients can probably avoid that situation I think it would be clearer if a single pool operation would block the operations from running concurrently rather than forcing all clients to do it. Would it cause issues? usually the only long-running operation is the download.

minglumlu:
> I would prefer a consistent way of using the pool operation. It's not clear what would be the state when a client removes or adds a yum repository while a download is in progress.

If a repo is enabled, it can't be removed. Similarly, adding a repo will not get it enabled automatically. So both cases will not impact the update status since the status is only related with the enabled repos.

minglumlu:
The "pool.sync_updates" can be called by XAPI scheduler periodically and clients at same time. From the XenCenter (client)'s perspective, it may be annoying to make customer get an error if a sync triggered by XAPI scheduler is in progress already. In other words, XenCenter would like to disable the "sync"/"apply" button when it knows a auto-sync is in progress. 
A client can't get the locking status of the mutex proposed to be removed in this PR, but it can get the pool operation through XAPI API e.g. event interface.
So the idea here is to populate the status to clients completely.

lindig:
Could the allowed operations mechanism be used to block update operations while a repository is being updated? The schedule for updates is set by the user so they can change a different one. In any case, I would avoid applying updates while updates themselves are updated.

lindig:
How much do we need this? I am not convinced that this is the right thing to do and this is why I have not approved this so far.

psafont:
I believe with the current approach there are races. I would prefer that the op `configure_repositories is blocked while `sync_updates is ongoing. This can be done in the function valid_operations https://github.com/xapi-project/xen-api/blob/0f717b741939fbe30a8aa150d0285ac6a1d75abd/ocaml/xapi/xapi_pool_helpers.ml#L62C5-L62C21

minglumlu:
Hi @lindig @psafont 
All the repo updates related operations are already protected by `with_pool_operation`.  In this PR, the change is to remove another lock. The lock works mainly in sync_updates as:

**Phase#1**
1 lock the mutex to be removed in this PR
2 downloading updates
3 unlock

**Phase#2**
4 lock with `with_pool_operation`
5 create new repo metadata based on the downloaded updates
6 unlock with end of `with_pool_operation`

This was designed to allow other repo updates operations when downloading the updates. 
With the change in this PR, the sync_updates would be like:
1 lock with `with_pool_operation`
2 downloading updates
3 create new repo metadata based on the downloaded updates
4 unlock with end of `with_pool_operation`

The origin of this change is that a client could only know the pool is in phase#1 (downloading updates) when it issued a "pool.sync_updates" and received an error "reposync_in_progress". It would be annoying for a client to keep issuing the call and receiving the error to know the status. Instead, a client can get the status of `with_pool_operation` by `pool.allowed_operations` and `pool.current_operations`.

So in this PR, the mutex is going to be removed and replaced with `with_pool_operation`.

psafont (approved):

lindig (approved):

lindig:
Thank you for the clarification, @minglumlu

PR #5137, opened 2023-07-31
CP-43977: Fallback un-recognized guidance as RebootHost

gangj:
In this way, future new added guidance can be covered by the max
guidance: RebootHost.

### DIFF
--- a/ocaml/xapi/updateinfo.ml
+++ b/ocaml/xapi/updateinfo.ml
@@ -50,9 +50,13 @@ module Guidance = struct
         EvacuateHost
     | "RestartDeviceModel" ->
         RestartDeviceModel
-    | _ ->
-        error "Unknown node in <absolute|recommended_guidance>" ;
-        raise Api_errors.(Server_error (invalid_updateinfo_xml, []))
+    | g ->
+        warn
+          "Un-recognized guidance in \
+           <absolute|recommended|livepatch_guidance>: %s, fallback to \
+           RebootHost"
+          g ;
+        RebootHost
 
   let of_update_guidance = function
     | `reboot_host ->
@@ -544,33 +548,15 @@ module UpdateInfo = struct
                       | Xml.Element ("description", _, [Xml.PCData v]) ->
                           {acc with description= v}
                       | Xml.Element ("recommended_guidance", _, [Xml.PCData v])
-                        -> (
-                        try {acc with rec_guidance= Some (Guidance.of_string v)}
-                        with e ->
-                          (* The error should not block update. Ingore it. *)
-                          warn "%s" (ExnHelper.string_of_exn e) ;
-                          acc
-                      )
-                      | Xml.Element ("absolute_guidance", _, [Xml.PCData v])
-                        -> (
-                        try {acc with abs_guidance= Some (Guidance.of_string v)}
-                        with e ->
-                          (* The error should not block update. Ingore it. *)
-                          warn "%s" (ExnHelper.string_of_exn e) ;
-                          acc
-                      )
-                      | Xml.Element ("livepatch_guidance", _, [Xml.PCData v])
-                        -> (
-                        try
+                        ->
+                          {acc with rec_guidance= Some (Guidance.of_string v)}
+                      | Xml.Element ("absolute_guidance", _, [Xml.PCData v]) ->
+                          {acc with abs_guidance= Some (Guidance.of_string v)}
+                      | Xml.Element ("livepatch_guidance", _, [Xml.PCData v]) ->
                           {
                             acc with
                             livepatch_guidance= Some (Guidance.of_string v)
                           }
-                        with e ->
-                          (* The error should not block update. Ingore it. *)
-                          warn "%s" (ExnHelper.string_of_exn e) ;
-                          acc
-                      )
                       | Xml.Element ("guidance_applicabilities", _, apps) ->
                           {
                             acc with
--- a/ocaml/tests/test_updateinfo.ml
+++ b/ocaml/tests/test_updateinfo.ml
@@ -786,6 +786,84 @@ module UpdateInfoOfXml = Generic.MakeStateless (struct
               )
             ]
         )
+      ; (* Single update with new guidances *)
+        ( {|
+            <updates>
+              <update type="security">
+                <id>UPDATE-0000</id>
+                <title>title</title>
+                <summary>summary</summary>
+                <description>description</description>
+                <special_info>special information</special_info>
+                <url>https://update.details.info</url>
+                <recommended_guidance>NewGuidance</recommended_guidance>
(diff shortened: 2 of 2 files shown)


ocaml/xapi/updateinfo.ml
```diff
@@ -50,9 +50,13 @@ module Guidance = struct
...
-    | _ ->
-        error "Unknown node in <absolute|recommended_guidance>" ;
-        raise Api_errors.(Server_error (invalid_updateinfo_xml, []))
+    | g ->
+        warn
+          "Un-recognized guidance in \
+           <absolute|recommended|livepatch_guidance>: %s, fallback to \
+           RebootHost to support future new added guidance"
+          g ;
+        RebootHost
```
minglumlu, line 59: I think it may be too strong to say `support` new guidance. This is only to fallback to the safest action on coordinator host only in normal case.
The normal case would be like this:
A pool has not been updated for a long time and hence has missed the update in XAPI to support a new guidance. While with this change, the pool coordinator would be updated firstly and rebooted up. Then the new coordinator can understand the new guidance, and meanwhile, the new guidance has been applied on itself as well which is benefited from this change. 
In other words, normally members hosts anyway will be handled correctly. So this change will benefit the coordinator only. The guidance will be evaluated again when individual member hosts are going to be updated.

Of course if user chose to not apply guidance immediately/automatically, then all the hosts will fall back to `RebootHost`.

minglumlu: I think a `UnknownGuidance` proposed by @gangj may come in again here for distinguish if the `RebootHost` is introduced by the unknown guidance, or there is a `RebootHost` originally. The `UnknownGuidance` will not be exposed to user.

alexbrett: We always want to minimise reboots for customers - as such if we add more subtle guidance in future it will be likely towards this goal. At that point, I think falling back to a reboot host seems to be achieving the exact opposite.

I wonder if we would be better to fallback to e.g. a toolstack restart, and we can then take that in to account if/when we introduce any future guidance (e.g. if we do need something that is closer to a reboot, then we should ensure the toolstack change which introduces the new guidance handling could require the reboot to avoid the problem)

minglumlu: `RestartToolstack` will be there surely since new guidance support in Toolstack must require it. This is only for the case that an old enough pool which has the new guidance support in Toolstack and an update which uses the new guidance as available updates at same time.

gangj: I think to release an update with a new type of guidance, there will be 2 steps:
1. release an update for xapi which add support for that new type of guidance in xapi (with RestartToolstack as recommended guidance)
2. release an update using that new type of guidance(in recommended, absolute or livepatch guidance)
And I think we might probably do above step 1 only when we found there is a new type of guidance in need. (Or do we have a plan to add some other kind of guidance?)

And if the pool have not applied the update from step 1 already, and then there comes the update from step 2, the pool will encounter the "new un-supported" guidance scenario we are discussing here. So I mean the pool do not have to be very old to encounter this scenario.

> We always want to minimise reboots for customers
@alexbrett I see your point. To achieve this goal, I think there is one option in xapi: when the pool runs into this condition, there will be `RestartToolstack` recommended guidance from step 1, if customer choose to apply recommended guidance automatically, first Toolstack will be restarted immediately after applying update on coordinator. After coordinator xapi is restarted, xapi can handle the new type of guidance, then we can let xapi apply the not-applied yet new type of guidance at the end of xapi startup. (add code at the end of [server_init](https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi.ml#L1016))
In this way, we can avoid the "fallback-ed reboot". But we need add code in xapi to support this.

However @minglumlu shared more thoughts about this:
From our update history, it might be a big chance that an future update will have "RebootHost" as recommended guidance, in that case, "new added guidance" will not be omitted. The point is that what we worry about: "new added guidance" might be missed, it might be a small chance issue. If this is true, we might not benefit to let xapi apply new added guidance during startup, this make the code complicated, just fallback to reboot(only coordinator) for rare case seems a better option.

> Of course if user chose to not apply guidance immediately/automatically, then all the hosts will fall back to RebootHost.
I think if customer choose to apply guidance himself/herself, after the update is installed, he/she can perform any exact guidance, reboot is not needed. If this is true, how about just expose the exact new guidance to user?

@alexbrett @robhoes @lindig @edwintorok @psafont @minglumlu Would you please share more thoughts, thank you!

minglumlu: I don't think there will be too many new guidances to be added. Even a new guidance is being added, the probability of hitting this falling-back is low as well.
I think a falling-back to `RebootHost` is the simplest way.

psafont: I agree about having reboot as fallback. I believe it's also important that we communicate to customers that frequent updates minimizes reboots (by avoiding these scenarios). There is tension with the general reticence to update I've seen in customer issues to apply fixes, let's hope that this way of applying updates helps

gangj: For the above options, I think @minglumlu 's proposal wins: `Fallback un-recognized guidance to reboot for pool coordinator only`, will change this way and push later.

gangj: > I think a `UnknownGuidance` proposed by @gangj may come in again here for distinguish if the `RebootHost` is introduced by the unknown guidance, or there is a `RebootHost` originally. The `UnknownGuidance` will not be exposed to user.

Since we will not expose the unknown guidance to user, I think there is no need for another new type guidance, just fallback any unknown guidance to `RebootHost` as current change in the PR will work.

> In other words, normally members hosts anyway will be handled correctly. So this change will benefit the coordinator only. The guidance will be evaluated again when individual member hosts are going to be updated.

Yes, in normal process(Coordinator Toolstack restart before applying updates in pool members), the `RebootHost` guidance which fallback from any unknown guidance will only be available for pool coordinator. But since XAPI API do not have any restriction for the update and guidance applying process: pool member might still be possible to apply updates before pool coordinator apply updates and restart Toolstack. In that case the fallback `RebootHost` guidance will also be available for the pool member.

So here I think we can not hide the Fallback `RebootHost` guidance for pool members from the /updates http api, this means if there is any unknown guidance, the guidance got from the /updates http api will contain `RebootHost` guidance. While if customer apply updates on pool coordinator and restart coordinator Toolstack first, and then after applying update on pool member, the guidance might change: the `RebootHost` guidance will gone if it is only from any unknown guidance, and it might change to some new added guidance. This requires the XAPI client(XC) to update the guidance again after applying updates.

gangj:
Force push to update the commit message.

minglumlu (approved):

robhoes (approved):

PR #5140, opened 2023-08-02
Add `vdi_update` filter to some tests

benjamreis:
These tests call at some point `VDI.update` so
the VDIs not supporting it should be filtered out.


benjamreis:
This has been found by running quicktest on different types of SR and these 4 tests failed with `UNSUPPORTED_OPERATIONS` instead of being ignored.

lindig (approved):

psafont (approved):
I checked that all of them indeed try to use Qt.VDI.Test_update

PR #5142, opened 2023-08-02
forkexecd: better recover from invalid rpc messsages

psafont:
In some situations the string-encoded length is malformed and cannot be converted into an integer. Use Result.t to relay the cause of the error and allow callers to log the string that caused the couldn't be decoded


ocaml/forkexecd/lib/fecomms.ml
```diff
@@ -23,12 +22,16 @@ let open_unix_domain_sock_client path =
...
-  let len = int_of_string (Bytes.unsafe_to_string buffer) in
-  let body = Unixext.really_read_string sock len in
-  ferpc_of_rpc (Jsonrpc.of_string body)
+  let header = Bytes.unsafe_to_string buffer in
+  match int_of_string_opt header with
+  | Some len ->
+      let body = Unixext.really_read_string sock len in
+      Ok (Fe.ferpc_of_rpc (Jsonrpc.of_string body))
+  | None ->
+      Error ("Header is not an integer: " ^ header)
```
edwintorok, line 31: If the RPC message is not valid, how can we trust that anything further we read from the socket will be valid? (the sender has broken the protocol)

We cannot close the socket here (it might lead to EBADF or sending data to the wrong socket), but we could shut it down such that any further attempts to read or write on the socket will fail immediately, see https://man7.org/linux/man-pages/man2/shutdown.2.html, in OCaml that would be `Unix.shutdown sock Unix.SHUTDOWN_ALL`

psafont: Updated the reader to shut down the socket, the writer side was not dealing well with closed pipes either, added code for recovery or ignore the error where appropriate

ocaml/forkexecd/lib/forkhelpers.ml
```diff
@@ -203,10 +213,23 @@ let safe_close_and_exec ?env stdin stdout stderr
...
+                (Fe.ferpc_to_string status)
+            in
+            failwith msg
+        | Error err ->
+            let msg =
+              Printf.sprintf
+                "%s: Received invalid reply from forking executioner (%s)"
+                __FUNCTION__ err
+            in
+            failwith msg
```
edwintorok, line 232: This is fine, you can also use `Fmt.failwith` instead which takes a printf format string and does the internal printf + failwith internally.

edwintorok (approved):

PR #5143, opened 2023-08-04
maintenance: prepare code for newer versions of libraries

psafont:
In particular:
- add dependency to ctypes.stubs to use its new dune-compiled version
- replace ounit with ounit2
- remove usages of Mtime.Span.to_s: avoid conversion to float if possible by comparing timespans, or use manual conversion to seconds in floats, taking care to remove as little precision as possible (floats become imprecise for periods longer than 54 years). The latter is needed when there are interfaces using the float seconds, which are difficult to change because they are outside our control, or they introduce backwards-compatibility headaches.

I've been working on a bigger code change for adopting monotonic timers across the board, but I'd like to get some functions upstreamed to Mtime before that to make the comparisons easier to understand.

### DIFF
--- a/ocaml/xapi/xapi_vdi.ml
+++ b/ocaml/xapi/xapi_vdi.ml
@@ -862,6 +862,7 @@ let snapshot ~__context ~vdi ~driver_params =
     It is sufficient to wait for all the VBDs to disappear, because a VBD
     cannot be destroyed if it is plugged. *)
 let wait_for_vbds_to_be_unplugged_and_destroyed ~__context ~self ~timeout =
+  let timeout = Mtime.Span.(timeout * s) in
   let vdi_uuid = Db.VDI.get_uuid ~__context ~self in
   (* We watch the "VBDs" field of this VDI. *)
   let classes = [Printf.sprintf "VDI/%s" (Ref.string_of self)] in
@@ -891,6 +892,7 @@ let wait_for_vbds_to_be_unplugged_and_destroyed ~__context ~self ~timeout =
         most_recent_snapshot
     in
     let from =
+      let timeout = Scheduler.span_to_s timeout in
       Helpers.call_api_functions ~__context (fun rpc session_id ->
           Client.Event.from ~rpc ~session_id ~classes ~token ~timeout
           |> Event_types.event_from_of_rpc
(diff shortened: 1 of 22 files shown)


lindig (approved):
Good change but it's easy to make a mistake that can't be caught by the compiler.

ocaml/xapi/xapi_vdi.ml
```diff
@@ -905,31 +910,24 @@ let wait_for_vbds_to_be_unplugged_and_destroyed ~__context ~self ~timeout =
...
+  let from_start = Mtime_clock.counter () in
   let token, initial_vbds = next_token_and_vbds ~token:"" ~timeout in
   (* When we use an empty token, we always get back the whole VDI record *)
   let initial_vbds = Option.get initial_vbds in
+  let pp_time () = Fmt.str "%a" Mtime.Span.pp in
   let rec loop ~token ~remaining_vbds =
-    let now = Mtime_clock.now () in
-    if remaining_vbds <> [] && Mtime.is_earlier now ~than:finish then (
+    let elapsed = Mtime_clock.count from_start in
+    if remaining_vbds <> [] && Mtime.Span.compare elapsed timeout < 0 then (
```
edwintorok, line 920: I quite liked the 'is_earlier', and it still seems to be present in Mtime 2.0, why do we have to use compare instead?

psafont: These are timespans, not timestamps.
I agree the comparisons should use be more ergonomic. I think longer and shorter are a very good match (and longest, shortest). I want to work more on the conversion before proposing the change upstream to make the comparisons easier to understand

ocaml/xapi/xapi_vdi.ml
```diff
@@ -858,10 +858,14 @@ let snapshot ~__context ~vdi ~driver_params =
...
   newvdi
 
+let span_to_s span =
+  Mtime.Span.to_uint64_ns span |> Int64.to_float |> fun ns -> ns /. 1e9
+
 (** Wait for all the VBDs of this VDI to be unplugged and destroyed.
     It is sufficient to wait for all the VBDs to disappear, because a VBD
     cannot be destroyed if it is plugged. *)
 let wait_for_vbds_to_be_unplugged_and_destroyed ~__context ~self ~timeout =
+  let timeout = Mtime.Span.(int_of_float (timeout *. 1000.) * ms) in
```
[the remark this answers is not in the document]
psafont: I'll change it to integer seconds, the precisions is not needed in this case as there's only a single, hardcoded value

ocaml/xapi/xapi_vdi.ml
psafont: Opened the PR upstream https://github.com/dbuenzli/mtime/pull/49

lindig: It was accepted. Quick turnaround

psafont: Indeed, the code here can be changed once we take in the (still unreleased) version that adds these functions

lindig:
` Mtime.Span.compare now timeout > 0` appears repeatedly and I find it not very readable:
* "now" suggests a timestamp
* timeout could be either a timestamp or a time span. For a timestamp "deadline" would be better
* "Span" suggest we are dealing with times spans but "now" does not make sense 
* Because it appears repeatedly, we might want to define a predicate that either checks if a timespan has elapsed or if a timestamp is passed a deadline.

robhoes (approved):

PR #5144, opened 2023-08-09
CA-379112 make PBD.plug wait for scan results 

lindig:
We have a race condition: PDB.plug updates the xapi database in a separate thread. This means the information there is accurate only after PBD.plug returns. In this case, the information about available space could be misleading before it is silently corrected. To avoid this, make this update synchronous.


edwintorok (approved):

psafont (approved):
The comment above needs to be changes to remove the mention for the asynchronicity

PR #5145, opened 2023-08-09
CA-378966: Detect state network interface changes

psafont:
Currently only address changes are detected, but state changes are needed
because MTU changes make the state of the link to change.

Not detecting this situation makes xapi hang at boot in certain scenarios.

### DIFF
--- a/ocaml/networkd/bin/network_monitor_thread.ml
+++ b/ocaml/networkd/bin/network_monitor_thread.ml
@@ -31,7 +31,7 @@ let monitor_whitelist =
     ; "vif" (* This includes "tap" owing to the use of standardise_name below *)
     ]
 
-let xapi_rpc xml =
+let rpc xml =
   let open Xmlrpc_client in
   XMLRPC_protocol.rpc ~srcstr:"xcp-networkd" ~dststr:"xapi"
     ~transport:(Unix "/var/xapi/xapi")
@@ -42,8 +42,7 @@ let send_bond_change_alert _dev interfaces message =
   let ifaces = String.concat "+" (List.sort String.compare interfaces) in
   let module XenAPI = Client.Client in
   let session_id =
-    XenAPI.Session.login_with_password ~rpc:xapi_rpc ~uname:"" ~pwd:""
-      ~version:"1.4"
+    XenAPI.Session.login_with_password ~rpc ~uname:"" ~pwd:"" ~version:""
       ~originator:("xcp-networkd v" ^ Xapi_version.version)
   in
   Pervasiveext.finally
@@ -53,13 +52,13 @@ let send_bond_change_alert _dev interfaces message =
       try
         let name, priority = Api_messages.bond_status_changed in
         let (_ : API.ref_message) =
-          XenAPI.Message.create ~rpc:xapi_rpc ~session_id ~name ~priority
-            ~cls:`Host ~obj_uuid ~body
+          XenAPI.Message.create ~rpc ~session_id ~name ~priority ~cls:`Host
+            ~obj_uuid ~body
         in
         ()
       with _ -> warn "Exception sending a bond-status-change alert."
     )
-    (fun _ -> XenAPI.Session.logout ~rpc:xapi_rpc ~session_id)
+    (fun _ -> XenAPI.Session.logout ~rpc ~session_id)
 
 let check_for_changes ~(dev : string) ~(stat : Network_monitor.iface_stats) =
   let open Network_monitor in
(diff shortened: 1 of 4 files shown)


ocaml/networkd/bin/network_monitor_thread.ml
```diff
@@ -371,9 +371,22 @@ let clear_input fd =
   in
   Unix.set_nonblock fd ; loop () ; Unix.clear_nonblock fd
 
+(* Watch for changes in the network interfaces: whether the IP changes, and
+   whether the links go online or offline *)
+let filters =
```
lindig, line 376: I like this design - a list of predicates on a string.

ocaml/xapi-aux/networking_info.ml
```diff
@@ -64,19 +62,14 @@ let get_management_ip_addr ~dbg =
...
+            L.error "%s: %s" __FUNCTION__ msg ;
+            raise (Unexpected_address_type msg)
       in
-      Option.join (List.nth_opt addrs 0)
+      addrs
+      |> List.map (fun (addr, _) -> Ipaddr_unix.of_inet_addr addr)
+      (* Filter out link-local addresses *)
+      |> List.filter (fun addr -> Ipaddr.scope addr <> Ipaddr.Link)
+      |> List.map (fun ip -> (Ipaddr.to_string ip, ipaddr_to_cstruct ip))
+      |> Fun.flip List.nth_opt 0
```
lindig, line 74: Should we log the complete list before we pick the head as a management address? You could bind `Fun.flip List.nth_opt 0` to `hd_opt` - it returns the head of the list or None.

lindig (approved):

ocaml/xapi-aux/networking_info.ml
psafont: > Should we log the complete list before we pick the head as a management address?

We will probably need to change this selection in the dual-stack case, for the time being it's fine as it is, we can add more debugging when exploring host ipv6.

For the other comment I'm not sure what you mean, there is no `List.hd_opt` in stdlib. Do you mean defining a new binding?

lindig: The idea was to do `let hd_opt xs = .. in` locally - just for clarity because I had to think for a moment what happens here.

ocaml/networkd/bin/network_monitor_thread.ml
rosslagerwall, line 376: Does this filter out events related to interfaces the toolstack doesn't need to monitor? Otherwise it's going to signal a networking change every time something happens to the vif and tap interfaces.

psafont: will test that, sounds like xapi is signaled way too many times

robhoes: I don't think that vif/tap devices would turn up in the output from {{ip monitor address}}, because they don't have IP addresses in dom0.

ocaml/networkd/bin/network_monitor_thread.ml
```diff
@@ -42,8 +42,7 @@ let send_bond_change_alert _dev interfaces message =
   let ifaces = String.concat "+" (List.sort String.compare interfaces) in
   let module XenAPI = Client.Client in
   let session_id =
-    XenAPI.Session.login_with_password ~rpc:xapi_rpc ~uname:"" ~pwd:""
-      ~version:"1.4"
+    XenAPI.Session.login_with_password ~rpc ~uname:"" ~pwd:"" ~version:""
```
psafont, line 45: the version parameter is ignored, has been since the first commit, so I doubt it's ever going to be useful
https://github.com/xapi-project/xen-api/blob/6b0ddfe5430f5f588dc7b226ce966434d099a4e9/ocaml/xapi/xapi_session.ml#L728

ocaml/networkd/bin/network_monitor_thread.ml
psafont: Code now ignore vifs and tap devices.

I think the output processing of ip monitor can be fragile, monitoring the sysfs will can be much more robust. I wouldn't try to use angstrom to try to parse the output, seems overkill when monitoring seems easier

robhoes: Or we could switch to NETLINK directly rather than `ip`.

psafont: Ah, didn't know we had these already: https://github.com/xapi-project/ocaml-netlink

robhoes: Yes, it was my first attempt at using ctypes :) It's not complete bindings though. The bit of Netlink that we need isn't there now. But I always wanted to extend it and make more use of it in xcp-network, to replace `ip` calls.

PR #5147, opened 2023-08-11
Revert "Cross-pool live migration: move CPU check to the target host"

lindig:
We see some errors in the context of snapshots that we need to investigate.


edwintorok (approved):
If we don't find another solution that fixes this reliably we can merge this revert (or perhaps we can merge this revert now, and once we found a fix and tested all the remaining corner cases then bring it back?)

psafont:
See the fix in https://github.com/xapi-project/xen-api/pull/5148

robhoes (approved):

psafont:
Needs a `git rebase master --signoff`

edwintorok (approved):

PR #5148, opened 2023-08-14
CA-380789: Not get power_state from snapshots with suspend VDIs

minglumlu:
This commit is to fix an issue which was introduced by 3d039f3. In VM migration, the suspend_vdis will be get from snapshots the power_state of which is `Suspended. The individual snapshots are represented as VM records in XAPI as well. They are different with the base VM record.

This commit reverts this part to the logic prior to 3d039f3.

### DIFF
--- a/ocaml/xapi/xapi_vm_migrate.ml
+++ b/ocaml/xapi/xapi_vm_migrate.ml
@@ -1241,7 +1241,7 @@ let migrate_send' ~__context ~vm ~dest ~live:_ ~vdi_map ~vif_map ~vgpu_map
   let suspends_vdis =
     List.fold_left
       (fun acc vm ->
-        if power_state = `Suspended then
+        if Db.VM.get_power_state ~__context ~self:vm = `Suspended then
           let vdi = Db.VM.get_suspend_VDI ~__context ~self:vm in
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -1241,7 +1241,7 @@ let migrate_send' ~__context ~vm ~dest ~live:_ ~vdi_map ~vif_map ~vgpu_map
   let suspends_vdis =
     List.fold_left
       (fun acc vm ->
-        if power_state = `Suspended then
+        if Db.VM.get_power_state ~__context ~self:vm = `Suspended then
```
psafont, line 1244: It's clear how reverting fixes the issue, maybe rename the binding to make sure it can't be mistaken again?
```suggestion
      (fun acc vm_or_snapshot ->
        let power_state = Db.VM.get_power_state ~__context ~self:vm_or_snapshot then
        if power_state = `Suspended then
```

psafont (approved):

lindig (approved):

ocaml/xapi/xapi_vm_migrate.ml
minglumlu: Let's minimize the changes at this stage :)

PR #5149, opened 2023-08-14
CA-379459 protect redo_log.shutdown with a lock

lindig:
"shutdown log" is called from multiple threads and this causes attempts to do the work twice, which fails. Protect the code by a lock such that an already shut down connection is recognised by later calls.

### DIFF
--- a/ocaml/database/redo_log.ml
+++ b/ocaml/database/redo_log.ml
@@ -269,8 +269,10 @@ let connect sockpath latest_response_time =
         D.debug "Remaining time for retries: %f" remaining ;
         Thread.delay attempt_delay ;
         attempt ()
-      ) else
+      ) else (
+        D.debug "%s timeout reached" __FUNCTION__ ;
         raise Unixext.Timeout
+      )
   in
   attempt ()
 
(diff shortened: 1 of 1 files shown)


ocaml/database/redo_log.ml
```diff
@@ -595,6 +600,12 @@ let shutdown log =
     (* ignore any errors *)
   )
 
+let locked f =
+  let finally () = Mutex.unlock shutdown_m in
```
edwintorok, line 604: I think we have a Mutex.execute somewhere in Threadext

edwintorok (approved):

ocaml/database/redo_log.ml
```diff
@@ -542,7 +544,10 @@ let maybe_retry f log =
 (* Functions relating to the lifecycle of the block device I/O process. *)
 
 (* Close any existing socket and kill the corresponding process. *)
-let shutdown log =
+
+let shutdown_m = Mutex.create ()
```
psafont, line 548: It seems to me that the resource in this case is `!(log.pid)`, having a mutex to be able to access it would make more sense that one for just the shutdown

ocaml/database/redo_log.ml
lindig:
```
let shutdown log =
  Xapi_stdext_threads.Threadext.Mutex.execute shutdown_m (fun () ->
      shutdown' log
  )
```

psafont: It's define in line 18 and ready to be used:

```
let with_lock = Xapi_stdext_threads.Threadext.Mutex.execute
```

ocaml/database/redo_log.ml
psafont, line 548: The log should have the name of the resource being used exclusively, in this case `!(log.pid)`.

We might want to drop mutex inside log `dying_processes_mutex`, it's only used here

lindig: Why do we need a thread to adjust a counter in the first place?

lindig: I don't think this lock can be removed. The update is happening in a separate thread.

PR #5150, opened 2023-08-15
CA-380715: Hit unspecified domain_type of snapshots in VM migration

minglumlu:
In live migration of a VM with snapshots, the "current_domain_type" of a snapshot may be "unspecified" if the snapshot was taken when the VM was in shutdown state. This will fail the CPU checking in migration as the checking has to know the domain type.

On the other hand, it is not necessary to check CPU for a snapshot if its domain type is unknown ('unspecified'). The potential problem might be the snapshot couldn't be reverted and started up in migration destination. But this can't be avoided anyway. For example, a new created VM which never started would hit same issue after migration (not live).

This commit skips the CPU checking on such snapshots.

### DIFF
--- a/ocaml/xapi/import.ml
+++ b/ocaml/xapi/import.ml
@@ -516,12 +516,18 @@ module VM : HandlerTools = struct
                 |> API.vM_metrics_t_of_rpc
               in
               let host = Helpers.get_localhost ~__context in
-              Cpuid_helpers.assert_vm_is_compatible ~__context
-                ~vm:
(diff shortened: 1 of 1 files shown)


minglumlu:
Close this as the related commits have been reverted.

robhoes:
Indeed, we had reverted the original change. However, you are correct that we should not do the CPU checks for snapshots; we didn't do so before my changes either. I'll run some more tests with my changes + your fixes later.

PR #5153, opened 2023-08-17
CA-379459 make shutdown mutex per redo_log

lindig:
The mutex for the shutdown operation should be local to the redo_log rather than the module. We can have many redo logs and they are independent.

This is a mall fixup for commit b687523fc; we ultimately might change how we use locks for redo locks but this is an improvement over what we have.


psafont (approved):

edwintorok (commented):
We could call this operation mutex and use it for shutdown to begin with and other operations later on (or keep it as shutdown mutex for now and rename it in a later commit).

edwintorok (approved):

lindig:
I like @edwintorok's remark and will rename the lock to `mutex` such that we can re-use it.

PR #5154, opened 2023-08-17
CA-381133: Set pending_guidances based on recommended guidance

robhoes:


ocaml/xapi/repository.ml
```diff
@@ -737,40 +737,29 @@ let apply_updates' ~__context ~host ~updates_info ~livepatches ~acc_rpm_updates
...
       (* EvacuateHost will be applied before applying updates *)
       eval_guidances ~updates_info ~updates:acc_rpm_updates ~kind:Recommended
         ~livepatches:successful_livepatches ~failed_livepatches
       |> List.filter (fun g -> g <> Guidance.EvacuateHost)
     in
-    let recommended_guidances' =
+    let guidances' =
       merge_with_unapplied_guidances ~__context ~host ~kind:Recommended
-        ~guidances:new_recommended_gs
+        ~guidances
```
gangj, line 748: Sorry, I think there is one debt from my previous code change:
`merge_with_unapplied_guidances` should be done after the `guidances` is updated by failed livepatch. Otherwise there might be duplicated `Guidance.RebootHostOnLivePatchFailure` guidance.

ocaml/xapi/repository.ml
```diff
@@ -737,40 +737,29 @@ let apply_updates' ~__context ~host ~updates_info ~livepatches ~acc_rpm_updates
...
-    let pending_guidances' =
-      merge_with_unapplied_guidances ~__context ~host ~kind:Absolute
-        ~guidances:new_pending_gs
-    in
-
     match failed_livepatches with
     | [] ->
         (* No livepatch should be applicable now *)
         Db.Host.remove_pending_guidances ~__context ~self:host
           ~value:`reboot_host_on_livepatch_failure ;
```
gangj, line 754: Not needed anymore.

ocaml/xapi/repository.ml
```diff
@@ -737,40 +737,29 @@ let apply_updates' ~__context ~host ~updates_info ~livepatches ~acc_rpm_updates
...
-        (recommended_guidances', pending_guidances')
+        (guidances', guidances')
     | _ :: _ ->
         (* There is(are) livepatch failure(s):
          * the host should not be rebooted, and
          * an extra pending guidance 'RebootHostOnLivePatchFailure' should be set.
          *)
-        ( List.filter (fun g -> g <> Guidance.RebootHost) recommended_guidances'
-        , Guidance.RebootHostOnLivePatchFailure :: pending_guidances'
+        ( List.filter (fun g -> g <> Guidance.RebootHost) guidances'
```
gangj, line 761: Do we need to expose `RebootHost` guidance to user when there is(are) livepatch failure(s)? If user need to tell if a pending “reboot host” is introduced by both "failed livepatch" and "some update", then we do not need to filter it out from the guidance here. 
@kc284 @alexbrett what is your thought?

ocaml/xapi/repository.ml
```diff
@@ -737,40 +737,29 @@ let apply_updates' ~__context ~host ~updates_info ~livepatches ~acc_rpm_updates
...
+        , Guidance.RebootHostOnLivePatchFailure :: guidances'
```
gangj, line 762: I think the 2 guidances returned here should also be equal.

edwintorok (commented):
Bit surprised this didn't require updating tests, don't we have any unit tests for the guidances logic in XAPI?

ocaml/xapi/xapi_host.ml
```diff
@@ -3074,23 +3074,25 @@ let apply_recommended_guidances ~__context ~self:host =
...
     | [] ->
-        try_restart_device_models_for_recommended_guidances ~__context ~host
+        try_restart_device_models ~__context ~host
     | [`reboot_host] ->
         Helpers.call_api_functions ~__context (fun rpc session_id ->
             Client.Client.Host.reboot ~rpc ~session_id ~host
         )
     | [`restart_toolstack] ->
-        try_restart_device_models_for_recommended_guidances ~__context ~host ;
+        try_restart_device_models ~__context ~host ;
```
edwintorok, line 3088: Why do you restart the device model when restarting the toolstack? Aren't they 2 separate guidances that are orthogonal?
(I assume restart device model = restart qemu by migrating the VM?)

ocaml/xapi/repository.ml
```diff
@@ -735,43 +720,33 @@ let apply_updates' ~__context ~host ~updates_info ~livepatches ~acc_rpm_updates
...
       eval_guidances ~updates_info ~updates:acc_rpm_updates ~kind:Recommended
         ~livepatches:successful_livepatches ~failed_livepatches
       |> List.filter (fun g -> g <> Guidance.EvacuateHost)
+      |> fun l -> merge_with_unapplied_guidances ~__context ~host ~guidances:l
     in
-    let guidances' =
-      merge_with_unapplied_guidances ~__context ~host ~kind:Recommended
-        ~guidances
-    in
+    GuidanceSet.assert_valid_guidances guidances' ;
```
psafont, line 732: Is it safe to move the assert to a point were not all guidances have been added yet? In particular the ones affected by the failed livepatches

ocaml/xapi/repository_helpers.ml
```diff
@@ -650,18 +650,14 @@ let eval_guidances ~updates_info ~updates ~kind ~livepatches ~failed_livepatches
...
-  | Guidance.Recommended ->
-      Db.Host.get_recommended_guidances ~__context ~self:host
-  )
+  Db.Host.get_pending_guidances ~__context ~self:host
   |> List.map (fun g -> Guidance.of_update_guidance g)
+  |> List.filter (fun g -> g <> Guidance.RebootHostOnLivePatchFailure)
   |> of_list
   |> union (of_list guidances)
-  |> resort_guidances ~kind
+  |> resort_guidances ~kind:Guidance.Recommended
```
psafont, line 660: Please change resort_guidances to have a different parameter: `remove_evacuations:false`, the current indirection is confusing

robhoes: Adding a new commit.

ocaml/xapi/repository.ml
robhoes: Originally, the assert was only done on `recommended_guidances`, which didn't have that livepatch guidance. I don't know if it interacts badly for some reason. I suggest leaving this now, just in case.

psafont (approved):

ocaml/xapi/xapi_host.ml
robhoes: We'll address this separately (it is not caused by this PR).

ocaml/tests/test_repository_helpers.ml
```diff
@@ -2755,7 +2755,8 @@ module GuidanceSetResortGuidancesTest = Generic.MakeStateless (struct
   let transform (kind, guidances) =
     guidances
     |> GuidanceSet.of_list
-    |> GuidanceSet.resort_guidances ~kind
+    |> GuidanceSet.resort_guidances
+         ~remove_evacuations:(kind = Guidance.Absolute)
```
edwintorok, line 2759: why does this deal with absolute guidances when the concept is being dropped from XAPI?

ocaml/xapi/repository_helpers.ml
```diff
@@ -143,7 +143,10 @@ module GuidanceSet = struct
         )
         gs precedences
     in
-    match kind with Recommended -> gs' | Absolute -> remove EvacuateHost gs'
+    if remove_evacuations then
+      remove EvacuateHost gs'
```
edwintorok, line 147: Why would you remove evacuatehost from absolute guidances? reboothost overrides evacuatehost but it should do that for both recommended and absolute. And if an update really doesn't need evacuation as an absolute guidance then it shouldn't contain it, and if it does XAPI shouldn't remove it.

ocaml/tests/test_repository_helpers.ml
robhoes: It's still part of the update metadata.

ocaml/xapi/repository_helpers.ml
robhoes: This isn't something that is being changed in this PR.

minglumlu (approved):

PR #5156, opened 2023-08-22
CA-381856: preserve host.last_software_update on pool join

robhoes:


ocaml/idl/datamodel_host.ml
```diff
@@ -921,6 +921,13 @@ let create_params =
     ; param_release= dundee_release
     ; param_default= Some (VBool true)
     }
+  ; {
+      param_type= DateTime
+    ; param_name= "last_software_update"
+    ; param_doc= "Date and time when the last software update was applied."
+    ; param_release= dundee_release
+    ; param_default= Some (VDateTime (Date.of_float 0.0))
```
psafont, line 929:
```suggestion
    ; param_default= Some (VDateTime Date.epoch)
```

ocaml/tests/test_host.ml
```diff
@@ -23,6 +23,7 @@ let add_host __context name =
        ~external_auth_service_name:"" ~external_auth_configuration:[]
        ~license_params:[] ~edition:"" ~license_server:[]
        ~local_cache_sr:Ref.null ~chipset_info:[] ~ssl_legacy:false
+       ~last_software_update:Xapi_stdext_date.Date.never
```
psafont, line 26:
```suggestion
       ~last_software_update:Xapi_stdext_date.Date.epoch
```

ocaml/xapi/dbsync_slave.ml
```diff
@@ -59,6 +59,7 @@ let create_localhost ~__context info =
         ~external_auth_configuration:[] ~license_params:[] ~edition:""
         ~license_server:[("address", "localhost"); ("port", "27000")]
         ~local_cache_sr:Ref.null ~chipset_info:[] ~ssl_legacy:false
+        ~last_software_update:Date.never
```
psafont, line 62:
```suggestion
        ~last_software_update:Date.epoch
```

ocaml/tests/test_host.ml
robhoes: Aha, there's 36 more of those in the existing code...

psafont: I'll change them all when I specsave the libraries into xapi

there are a few deprecated bindings in the libirary: https://ocaml.org/p/xapi-stdext-date/latest/doc/Xapi_stdext_date/Date/index.html#val-never

lindig (approved):

ocaml/idl/datamodel_host.ml
robhoes: I've changed them all now.

minglumlu (approved):

minglumlu:
Would the `last_software_update ` be preserved also on a host which has been ejected from a pool? Or should it be preserved in that case?

robhoes:
@minglumlu, not for ejected hosts. The eject case is like a fresh install, where the host becomes a new pool-of-one, with an empty initial DB.

PR #5157, opened 2023-08-23
CP-44988: remove API: host.apply_recommended_guidances

gangj:
This API was added in 23.18.0 for applying all the recommended guidances after updates are applied. While it is found that the order in which some recommended guidances are applied is not correct: a RestartDeviceModel is applied before RestartToolstack, this is a bug:

1. if the new toolstack is already unpacked it may have backward incompatible scripts vs the running xapi/xenopsd, so you really must restart the toolstack asap, and not do some migrations first which will very likely fail completely.

2. if there is a (security) fix in xenopsd in how we launch the device model (qemu) then obviously that code needs to be active first but we don't have a way to express that.

And since XC do not use this API to apply recommended guidance(XC calls corresponding APIs to apply each guidance one by one), this problematic API is simply removed to resolve the problem.


minglumlu:
I think it needs to be removed through `lifecycle`.

gangj:
> I think it needs to be removed through `lifecycle`.

Thanks, now I see `Removed` in some API's lifecycle. I think these might be needed, will add back them.
I removed the code referring this commit: 2f73fa2a38baeb1478c95efc233afb2ca504909f, where all the code is removed, no `Deprecated` and `Removed` are added.

lindig:
I am fine with removing this. I think the guidance is missing the definition of a normal form: a sequence of guidance can be rewritten into a more efficient normal form that avoids doing work twice. The normal form would include which operations cancel each other and how they commute. This normal form needs to be known by whoever controls the application of guidance.

lindig (approved):

edwintorok (approved):

robhoes:
NOTE: don't merge yet, as we need to update some internal tests first.

PR #5158, opened 2023-08-25
CP-42559: Add RBAC info to C#, Java, and C SDK method comments

danilo-delbusso:
The PowerShell SDK needs more work as we'd need to generate a valid `dll-Help.xml` file to properly comment cmdlets: https://learn.microsoft.com/en-us/powershell/scripting/developer/help/how-to-create-the-cmdlet-help-file?view=powershell-7.3

Comment based support for `Get-Help` is only available for modules written directly in PowerShell.

Bit out of scope for this, worth opening a separate ticket.

Example of the new docs:

C#

```csharp
/// <summary>
/// Create a new VTPM instance, and return its handle.
 /// Experimental. First published in 22.26.0.
/// </summary>
/// <param name="session">The session</param>
/// <param name="_vm">The VM reference the VTPM will be attached to</param>
/// <param name="_is_unique">Whether the VTPM must be unique</param>
/// <remarks>
/// Minimum allowed role: vm-admin
/// </remarks>
public static XenRef<VTPM> create(Session session, string _vm, bool _is_unique)
{
    return session.JsonRpcClient.vtpm_create(session.opaque_ref, _vm, _is_unique);
}
```

C
```c
/**
 * Create a new VTPM instance, and return its handle.
 * Minimum allowed role: vm-admin.
 */
extern bool
xen_vtpm_create(xen_session *session, xen_vtpm *result, xen_vm vm, bool is_unique);
```

Java

```java
/**
   * Create a new VTPM instance, and return its handle.
   * Minimum allowed role: vm-admin
   * Experimental. First published in 22.26.0.
   *
   * @param c The connection the call is made on
   * @param vM The VM reference the VTPM will be attached to
   * @param isUnique Whether the VTPM must be unique
   * @return The reference of the newly created VTPM
   * @throws BadServerResponse Thrown if the response from the server contains an invalid status.
   * @throws XenAPIException Thrown if the call failed.
   * @throws XmlRpcException Thrown if the result of an asynchronous call could not be parsed.
   */
public static VTPM create(Connection c, VM vM, Boolean isUnique) throws BadServerResponse, XenAPIException, XmlRpcException 
{
    String method_call = "VTPM.create";
    String session = c.getSessionReference();
    Object[] method_params = {Marshalling.toXMLRPC(session), Marshalling.toXMLRPC(vM), Marshalling.toXMLRPC(isUnique)};
    Map response = c.dispatch(method_call, method_params);
    Object result = response.get("Value");
    return Types.toVTPM(result);
}
```

### DIFF
--- a/ocaml/idl/datamodel_roles.ml
+++ b/ocaml/idl/datamodel_roles.ml
@@ -64,7 +64,23 @@ let roles_all = List.map role_name ordered
 
 let role_description = List.map role_name_with_description ordered
 
-let role_is_internal = List.map role_name_with_internal ordered
+let role_is_internal_map = List.map role_name_with_internal ordered
+
+(** Returns true if role_name_label matches that of an internal role.
+@param role_name_label the string representation of the role we're checking.
+@raise Failure if there is no matching field or the field doesn't have a is_internal value.
+*)
+let role_is_internal role_name_label =
+  match List.assoc_opt role_name_label role_is_internal_map with
+  | Some is_internal ->
+      is_internal
+  | None ->
+      failwith
+        (Printf.sprintf
+           "Check Datamodel_roles.role_is_internal: there's no is_internal \
+            field set for role %s"
+           role_name_label
+        )
 
 (* obtain all roles with at least the specified role privileges *)
 let roles_gte role =
--- a/ocaml/sdk-gen/common/CommonFunctions.ml
+++ b/ocaml/sdk-gen/common/CommonFunctions.ml
@@ -197,6 +197,16 @@ let gen_param_groups message params =
        (if expRelease = "" then overloadGroups else [params])
     )
 
+and get_minimum_allowed_role msg =
+  let get_last_role roles =
+    match List.rev roles with [] -> "Not Applicable" | last :: _ -> last
+  in
+  match msg.msg_allowed_roles with
+  | None ->
+      "Not Applicable"
+  | Some roles ->
+      get_last_role roles
+
 (*** XML documentation ***)
 
 and get_published_info_message message cls =
--- a/ocaml/sdk-gen/common/CommonFunctions.ml
+++ b/ocaml/sdk-gen/common/CommonFunctions.ml
@@ -198,8 +198,19 @@ let gen_param_groups message params =
     )
 
 and get_minimum_allowed_role msg =
-  let get_last_role roles =
-    match List.rev roles with [] -> "Not Applicable" | last :: _ -> last
+  let get_last_role recroles =
+    let reversed_role_list = List.rev recroles in
+    let rec get_role reversed_list =
+      match reversed_list with
+      | [] ->
+          "Not Applicable"
+      | last :: rest ->
+          if not (Datamodel_roles.role_is_internal last) then
+            last
+          else
+            get_role rest
+    in
+    get_role reversed_role_list
   in
   match msg.msg_allowed_roles with
   | None ->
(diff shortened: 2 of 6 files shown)


ocaml/sdk-gen/java/main.ml
```diff
@@ -288,6 +288,8 @@ let gen_method file cls message params async_version =
 
   fprintf file "    /**\n" ;
   fprintf file "     * %s\n" (escape_xml message.msg_doc) ;
+  fprintf file "     * Minimum allowed role: %s\n" (get_minimum_allowed_role
```
danilo-delbusso, line 291: Could not find a better option for [javadoc](https://docs.oracle.com/javase/1.5.0/docs/tooldocs/windows/javadoc.html) style docs. So it's just in the generic method summary

ocaml/sdk-gen/common/CommonFunctions.ml
```diff
@@ -197,6 +197,16 @@ let gen_param_groups message params =
        (if expRelease = "" then overloadGroups else [params])
     )
 
+and get_minimum_allowed_role msg =
```
kc284, line 200: This is probably a question for the Toosltack folk: Should we exclude internal roles?

I suppose this also goes for the API docs, which currently do not hide internal roles. For example, the mimimun role for `repository.set_name_label` is documented as `client-cert` - not sure how much sense this makes to the API users. (Note that on the UI we do hide internal roles).

danilo-delbusso: cc @psafont @lindig

psafont: I think it makes sense as a policy to hide internal roles as a policy. Might as well start now here

kc284 (changes requested):
Ok, based on Pau's reply, let's exclude internal roles.

ocaml/idl/datamodel_roles.ml
```diff
@@ -64,7 +64,21 @@ let roles_all = List.map role_name ordered
...
+*)
+let role_is_internal role_name_label =
+  try List.assoc role_name_label role_is_internal_map
+  with Not_found ->
+    failwith
+      (Printf.sprintf
+         "Check Datamodel_roles.role_is_internal: there's no is_internal field \
+          set for role %s"
+         role_name_label
+      )
```
psafont, line 81:
```suggestion
  match List.assoc_opt role_name_label role_is_internal_map with
  | Some is_internal ->
      is_internal
  | None ->
      failwith
        (Printf.sprintf
           "Check Datamodel_roles.role_is_internal: there's no is_internal field \
            set for role %s"
           role_name_label
        )
```
It's usually more efficient to not raise exceptions to immediately capture it and use options instead

psafont (dismissed):

ocaml/sdk-gen/common/CommonFunctions.ml
```diff
@@ -197,6 +197,27 @@ let gen_param_groups message params =
...
+  let get_last_role recroles =
+    let reversed_role_list = List.rev recroles in
+    let rec get_role reversed_list =
+      match reversed_list with
+      | [] ->
+          "Not Applicable"
```
kc284, line 206: I was diffing the generated code:
- I see `Not Applicable` being attached to all `get_all_records` calls, for which I would expect the role to be `read-only`.
- Toolstack can correct me if I'm wrong, but for the calls that previously had an internal role, shouldn't the docs show the next public available one, i.e. `read-only`?

kc284 (changes requested):
As above.

kc284 (approved):

kc284 (approved):

kc284:
@psafont Danilo dismissed your review due to adding one more commit after you had approved, could you please have one more look - thanks.

PR #5160, opened 2023-08-25
Install suspend_image_viewer

edwintorok:
This is useful for debugging. It cannot be easily run out of Dom0 due to its dependency on verify-stream-v2 provided by Xen.
Also if it is not installed in Dom0 then it is bitrotting and not being tested, so when we need it then we realise it is broken.

Note that `verify-stream-v2` is currently bitrotted and its shebang needs to be edited to python2 to make it work.

Can be used like this:

```
/opt/xensource/debug/with-vdi de9bd384-b866-42b6-a8a0-b4f03a23c0fe
./suspend_image_viewer.exe --path /dev/$DEVICE --verify-stream-v2=/usr/libexec/xen/bin/verify-stream-v2
```

(This will require a specfile update when merged)

### DIFF
--- a/ocaml/xenopsd/suspend_image_viewer/dune
+++ b/ocaml/xenopsd/suspend_image_viewer/dune
@@ -1,5 +1,6 @@
 (executable
- (public_name suspend_image_viewer)
+ (public_name suspend-image-viewer)
+ (name suspend_image_viewer)
  (package xapi-xenopsd-xc)
  (libraries
   cmdliner
--- a/ocaml/xenopsd/suspend_image_viewer/dune
+++ b/ocaml/xenopsd/suspend_image_viewer/dune
@@ -1,5 +1,6 @@
 (executable
- (name suspend_image_viewer)
+ (public_name suspend_image_viewer)
+ (package xapi-xenopsd-xc)
  (libraries
   cmdliner
   forkexec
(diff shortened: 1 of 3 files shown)


ocaml/xenopsd/suspend_image_viewer/dune
```diff
@@ -1,5 +1,6 @@
 (executable
- (name suspend_image_viewer)
+ (public_name suspend_image_viewer)
```
robhoes, line 2: Perhaps call it `suspend-image-viewer` for consistency with other CLI tools.

lindig (approved):
I agree with @robhoes about the name using hyphens.

lindig:
I think we should merge this @robhoes or at least approve such that it has the required ticks.

ocaml/xenopsd/suspend_image_viewer/dune
```diff
@@ -1,5 +1,7 @@
 (executable
+ (public_name suspend-image-viewer)
```
edwintorok, line 2: Note that the module name needs to stay with underscores because OCaml doesn't accept '-' in filenames, but the public name can indeed be anything, including '-'.

ocaml/xenopsd/suspend_image_viewer/view.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-# Usage: copy this script and suspend_image_viewer executable to a box
+# Usage: copy this script and suspend-image-viewer executable to a box
```
edwintorok, line 2: This script probably doesn't work anymore, but I renamed it nevertheless.

robhoes (approved):

PR #5161, opened 2023-08-29
Fix bugs in merge of guidances and livepatch failures

minglumlu:
This commit aims to fix bugs in merging guidances. Following changes are included in this commit:
1. merge_with_unapplied_guidances: merge host|VM.pending_guidances with the guidances calculated from the latest update.
2. merge_livepatch_failures: merge livepatch failures in host.pending_guidances with the failures in the latest update.
3. Remove the merged guidances from the return of `host.apply_updates` as they have been set in host|VM.pending_guidances.
4. Add RebootHostOnXenLivepatchFailure and RebootHostOnKernelLivepatchFailure to deprecate RebootHostOnlLivepatchFailure
5. Add exhaustive unit tests for this part.

### DIFF
--- a/ocaml/tests/test_repository_helpers.ml
+++ b/ocaml/tests/test_repository_helpers.ml
@@ -46,6 +46,62 @@ let fields_of_update =
     ; field "repository" (fun (r : Update.t) -> r.repository) string
     ]
 
+let rec combination' acc = function
+  | [] ->
+      acc @ [[]]
+  | hd :: tl ->
+      let acc' = [[hd]] @ List.map (fun l -> hd :: l) acc @ acc in
+      combination' acc' tl
+
+let combination l = combination' [] l
+
+let join l1 l2 =
+  List.fold_left
+    (fun acc e1 -> List.fold_left (fun acc' e2 -> (e1, e2) :: acc') acc l2)
+    [] l1
+
+let join3 l2 l1 =
+  List.fold_left
+    (fun acc (x1, x2) ->
+      List.fold_left (fun acc' y -> (x1, x2, y) :: acc') acc l2
+    )
+    [] l1
+
+let join4 l2 l1 =
+  List.fold_left
+    (fun acc (x1, x2, x3) ->
+      List.fold_left (fun acc' y -> (x1, x2, x3, y) :: acc') acc l2
+    )
+    [] l1
+
+let guidances_to_string gs =
+  "[" ^ (List.map Guidance.to_string gs |> String.concat "; ") ^ "]"
+
+let livepatch_components_to_string components =
+  "["
+  ^ (List.map Livepatch.string_of_component components |> String.concat "; ")
+  ^ "]"
+
+let livepatching_result_to_string succeeded failed =
+  let to_string = livepatch_components_to_string in
+  "(" ^ to_string succeeded ^ ", " ^ to_string failed ^ ")"
+
+let exhaustive_livepatching_results =
+  let open Livepatch in
+  [
+    ([], [])
+  ; ([], [Xen])
+  ; ([], [Kernel])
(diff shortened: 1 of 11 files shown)


ocaml/tests/test_repository_helpers.ml
```diff
@@ -3608,6 +3608,448 @@ module GetLatestUpdatesFromRedundancy = Generic.MakeStateless (struct
...
+        )
+      ; (([], ([Xen], [])), []) (* 5 ~ 9 *)
+      ; (([], ([Xen], [Kernel])), [RebootHostOnKernelLivePatchFailure])
+      ; (([], ([Kernel], [])), [])
+      ; (([], ([Kernel], [Xen])), [RebootHostOnXenLivePatchFailure])
+      ; (([], ([Xen; Kernel], [])), [])
+      ; ( ([RebootHostOnXenLivePatchFailure], ([], []))
+        , [RebootHostOnXenLivePatchFailure]
+        )
+        (* 10 ~ 14 *)
```
edwintorok, line 3753: These kind of tests would probably be easier to write and maintain using property based testing.
Although the manually written ones are useful too (e.g. a property test could check whether the manual ones match expectations or not).

minglumlu: How about adding a new test case and in the case, the number of exhausted inputs is to be verified against the number of sub cases in this test?
It has to be done manually to map the auto-generated input to an expected output.

ocaml/xapi/repository_helpers.ml
```diff
@@ -653,15 +653,92 @@ let eval_guidances ~updates_info ~updates ~kind ~livepatches ~failed_livepatches
...
+  let unapplied_only = diff unapplied_set coming_set in
+  let to_be_added', to_be_removed', unchanged' =
+    match (is_empty unapplied_only, subset unapplied_only resorted) with
+    | true, _ ->
+        (resorted, empty, empty)
+    | false, true ->
+        (* The unapplied_only is contained in resorted *)
+        (diff resorted unapplied_only, empty, unapplied_only)
+    | false, false ->
+        (diff resorted unapplied_only, unapplied_only, empty)
```
gangj, line 675: Can we get the result directly like (tried this change with the added UT, passed):

  let to_be_added' = diff resorted unapplied_only in
  let to_be_removed' = diff unapplied_set resorted in
  let kept_only_from_unapplied' = diff resorted coming_set in
  let to_be_added, to_be_removed, kept_untouched =
    (elements to_be_added', elements to_be_removed', elements kept_only_from_unapplied')
  in

  # might worth adding comment for the "kept_untouched"(or "unchanged") to make it more clear: those guidance which come only from "unapplied", and not from "coming", and after "resort", those guidance are still needed.

ocaml/tests/test_repository_helpers.ml
```diff
@@ -3608,6 +3608,448 @@ module GetLatestUpdatesFromRedundancy = Generic.MakeStateless (struct
...
+    let open Guidance in
+    let open Livepatch in
+    `QuickAndAutoDocumented
+      [
+        (* 0 ~ 4 *)
+        (([], ([], [])), [])
+      ; (([], ([], [Xen])), [RebootHostOnXenLivePatchFailure])
+      ; (([], ([], [Kernel])), [RebootHostOnKernelLivePatchFailure])
+      ; ( ([], ([], [Xen; Kernel]))
+        , [RebootHostOnXenLivePatchFailure; RebootHostOnKernelLivePatchFailure]
```
gangj, line 3743: Find a limitation of the test: the test will fail if the order of output elements is changed, for example,
this case will fail if this line changes to:
["RebootHostOnXenLivePatchFailure"; "RebootHostOnKernelLivePatchFailure"]

It will be better to check ignoring the order.

PR #5164, opened 2023-08-30
CP-42949: Ensure storage RRDs are created without tapdev in kernel

robhoes:
All the necessary data is available also from tapdisk/blktap3 directly, through the RRD plugin transport protocol. However, this was missed because of an assumption that tapdev kernel devices are always present. This is no longer the case.

The code is reorganised to take this into account. In the near future, we can probably just delete all of the tapdev-dependent code.

### DIFF
--- a/ocaml/xcp-rrdd/bin/rrdp-iostat/rrdp_iostat.ml
+++ b/ocaml/xcp-rrdd/bin/rrdp-iostat/rrdp_iostat.ml
@@ -353,7 +353,10 @@ let refresh_phypath_to_sr_vdi () =
          )
      )
 
-let exec_tap_ctl () =
+(* Get a list of currently-active tapdisk processes as an assoc list from SR/VDI tuple
+   to minor number. As a side-effect, this updates the VDI-to-VM in the process state,
+   if needed. *)
+let exec_tap_ctl_list () : ((string * string) * int) list =
   let tap_ctl = "/usr/sbin/tap-ctl list" in
   let extract_vdis pid minor _state kind phypath =
     if not (kind = "vhd" || kind = "aio") then raise (Failure "Unknown type") ;
@@ -379,7 +382,11 @@ let exec_tap_ctl () =
   let pid_and_minor_to_sr_and_vdi =
     Utils.exec_cmd (module Process.D) ~cmdstring:tap_ctl ~f:process_line
   in
-  let minor_to_sr_and_vdi = List.map snd pid_and_minor_to_sr_and_vdi in
+  let sr_and_vdi_to_minor =
+    List.map
+      (fun (_, (minor, sr_and_vdi)) -> (sr_and_vdi, minor))
+      pid_and_minor_to_sr_and_vdi
+  in
   (let reason_for_updating_vdi_to_vm_map =
      let pid_vdis_to_string pids =
        pids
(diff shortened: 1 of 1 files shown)


ocaml/xcp-rrdd/bin/rrdp-iostat/rrdp_iostat.ml
```diff
@@ -951,9 +952,27 @@ let gen_metrics () =
     Blktap3_stats_wrapper.get_domid_devid_to_stats_blktap3 ()
   in
 
+  let sr_and_vdi_to_minor = exec_tap_ctl_list () in
+  let tapdevs = get_tapdevs () in
+
+  let map_sr_and_vdi_to_stats minor_to_tapdev_stat =
+    List.map
+      (fun (sr_and_vdi, minor) ->
+        let data = List.assoc_opt minor minor_to_tapdev_stat in
```
edwintorok, line 961: Not related to your change, but in the future we should replace these with some Maps at least. Assoc lists are not good for scalability, that N^2 becomes problematic fast...

ocaml/xcp-rrdd/bin/rrdp-iostat/rrdp_iostat.ml
```diff
@@ -998,7 +1017,7 @@ let gen_metrics () =
               None
           | Some s ->
               if Hashtbl.mem s sr_vdi then
-                Some (Hashtbl.find s sr_vdi)
+                Hashtbl.find s sr_vdi
```
lindig, line 1020: The combination of mem/find is prone to race conditions and not efficient. It's better to use `Hashtbl.find_opt`; I am surprised that the current code is type correct.

ocaml/xcp-rrdd/bin/rrdp-iostat/rrdp_iostat.ml
```diff
@@ -353,7 +353,10 @@ let refresh_phypath_to_sr_vdi () =
          )
      )
 
-let exec_tap_ctl () =
+(* Get a list of currently-active tapdisk processes as an assoc list from SR/VDI tuple
+   to minor number. As a side-effect, this updates the VDI-to-VM in the process state,
+   if needed. *)
+let exec_tap_ctl_list () : ((string * string) * int) list =
```
lindig, line 359: Maybe not worth changing when we believe this code is not going to stay around for long: a record type instead of a tuple would help readability.

ocaml/xcp-rrdd/bin/rrdp-iostat/rrdp_iostat.ml
```diff
@@ -445,46 +452,36 @@ let exec_tap_ctl () =
...
   previous_map := pid_and_minor_to_sr_and_vdi ;
-  minor_to_sr_and_vdi
+  sr_and_vdi_to_minor
 
-let minor_of_tdX_unsafe tdX =
+(* Get the minor number of the kernel tapdev device *)
+let minor_of_tapdev_unsafe tapdev =
   int_of_string
     (Unixext.file_lines_fold
        (fun acc l -> acc ^ List.nth (Xstringext.String.split ':' l) 1)
```
lindig, line 461: String concatenations with `^` are expensive. `List.nth` is could be solved with pattern matching; its' ok here because `1` is a small number.

lindig (approved):

ocaml/xcp-rrdd/bin/rrdp-iostat/rrdp_iostat.ml
robhoes: True, I'll change it to `find_opt`.

edwintorok (approved):

PR #5165, opened 2023-08-31
Improve redo_log

lindig:
We have seen "Redo_log.RedoLogFailure("Cannot connect to redo log")" and it mostly stems from threads interfering. The redo_log module was not designed with thread-safety in mind. This path greatly reduces the problem. We might want to study the problem more closely but I would suggest to take this patch to make the problem smaller.

The key ingredient is taking the database lock before opening the redo log flushing the database to the redo_log. This blocks other operations.

### DIFF
--- a/ocaml/database/redo_log.ml
+++ b/ocaml/database/redo_log.ml
@@ -546,7 +546,8 @@ let maybe_retry f log =
 
 (* Close any existing socket and kill the corresponding process. *)
 
-let shutdown' log =
+let shutdown log =
+  with_lock log.mutex @@ fun () ->
   if is_enabled log then (
     D.debug "Shutting down connection to I/O process for '%s'" log.name ;
     try
@@ -600,8 +601,6 @@ let shutdown' log =
     (* ignore any errors *)
   )
 
-let shutdown log = with_lock log.mutex (fun () -> shutdown' log)
-
 let broken log =
   set_time_of_last_failure log ;
   shutdown log ;
@@ -614,6 +613,7 @@ let healthy log =
 exception TooManyProcesses
 
 let startup log =
+  with_lock log.mutex @@ fun () ->
   if is_enabled log then (
     try
       ( match !(log.pid) with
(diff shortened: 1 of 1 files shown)


lindig:
@edwintorok @psafont could we review this? I believe it improves the situation short of re-writing the redo log implementation. See CP-45213

ocaml/database/redo_log.ml
```diff
@@ -886,11 +886,12 @@ let flush_db_exn db log =
...
 
 let enable_and_flush db log reason =
-  enable_existing log reason ; flush_db_exn db log
+  enable_existing log reason ;
+  with_lock redo_log_creation_mutex @@ fun () -> flush_db_exn db log
 
 let enable_block_and_flush db log path =
   enable_block_existing log path ;
-  flush_db_exn db log
+  with_lock redo_log_creation_mutex @@ fun () -> flush_db_exn db log
```
psafont, line 894: Why is this mutex held? I see it's only used for manipulating the Set of redo_logs

lindig: I can test this. This one was suggested by @edwintorok and I tested it.

lindig: Using `log.mutex` instead causes a deadlock.

psafont (approved):

ocaml/database/redo_log.ml
edwintorok: `with_active_redologs` holds this mutex and is what used to synchronize DB delta writes. Although there are some calls such as shutdown which are not always wrapped with that.

edwintorok (approved):
I think there are still race conditions between `delete` (or `shutdown`) and actively writing to the redo log, because not all callers of 'shutdown' or 'delete' call 'with_active_redologs'.
Doesn't matter if you only have 1 redo log (you'll be unsafe once they are shutdown anyway), but if you have multiple redo logs and only shut down 1 then you may get some exceptions raised.

It might be useful to acquire the per-log mutex too whenever performing an operation on it, but we need to be careful not to deadlock in that case, because callers aren't consistent about which locks they take.

OTOH forcibly shutting down a redolog may be useful (e.g. if its device is not able to complete IO operations successfully anymore).

Anyway I think this PR is an improvement and we should merge it, but there is more we could do.

PR #5168, opened 2023-09-01
CP-45006: Define volume.compose API for SMAPIv3.

KevinLCtx:
Description was adapted from SMAPIv2.

### DIFF
--- a/ocaml/xapi-storage/generator/lib/control.ml
+++ b/ocaml/xapi-storage/generator/lib/control.ml
@@ -403,6 +403,15 @@ module Volume (R : RPC) = struct
       @-> returning changed_blocks errors
       )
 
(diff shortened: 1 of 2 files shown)


lindig:
This looks ok but it would be good to provide more context in the commit message.

lindig (approved):

KevinLCtx:
Updated commit message.

lindig:
We have all kinds of CI issues here that need to be addressed

robhoes:
Looks fine, but you may need to define the new function in xapi-storage-script, as it may otherwise complain about a missing implementation when starting up. I'll try it out and will let you know.

robhoes (approved):

PR #5172, opened 2023-09-11
Fix installation of xapi-storage for Python 3

robhoes:

PR #5173, opened 2023-09-12
Update for python3 build

liulinC:
CP-44271: Remove python build/install from source code
    
    This part should be removed to spec file, so it can be handled
    properly basing on python2/python3

CP-44271: Conditionally run python test
    
    The python test for xapi-storage should only be run under python2
    environment, as it is in python2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,9 @@ lint:
 
 test:
 	dune runtest --profile=$(PROFILE) --error-reporting=twice -j $(JOBS)
+ifneq ($(PY_TEST), NO)
 	dune build @runtest-python --profile=$(PROFILE)
+endif
 
 stresstest:
 	dune build @stresstest --profile=$(PROFILE) --no-buffer -j $(JOBS)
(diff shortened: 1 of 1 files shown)


liulinC:
xenrt job: 188376 (Dev Run)

robhoes:
I have a patch in #5172 to add a missing `make install` for python 3 to xapi-storage. Will you handle this in the spec file now?

Makefile
```diff
@@ -30,7 +30,9 @@ lint:
 
 test:
 	dune runtest --profile=$(PROFILE) --error-reporting=twice -j $(JOBS)
+ifeq ($(PY_TEST), YES)
 	dune build @runtest-python --profile=$(PROFILE)
```
psafont, line 34: I run the python tests when developing, this change will make them more difficult to run. Please change the conditional to make the tests be run by default and when a variable is set they are not run

liulinC: is it not default to run?
PY_TEST is not set by default.

liulinC: when you comes to xs9, you may not run it by default (as xs9 do not have python2 for python test).
I will update to run it by default.

psafont (approved):

PR #5175, opened 2023-09-14
CA-371529 XSI-1329 remove license check for has-vendor-device

lindig:
The API call VM.set-has-vendor-device used to be a lincensed feature but it no longer is. As a first step to simplify when Windows VMs automatically or not update device drivers, remove the license checks in the code. The feature flag still remains in place.

### DIFF
--- a/ocaml/idl/datamodel_vm.ml
+++ b/ocaml/idl/datamodel_vm.ml
@@ -2093,34 +2093,8 @@ let t =
             "The host virtual hardware platform version the VM can run on"
         ; field ~qualifier:StaticRO
             ~lifecycle:[(Published, rel_dundee, "")]
-            ~doc_tags:[Windows]
-            ~default_value:
-              (Some
-                 (VCustom
-                    ( String.concat "\n"
-                        [
-                          "(try Rpc.Bool ("
-                        ; "let pool = List.hd \
-                           (Db_actions.DB_Action.Pool.get_all ~__context) in"
-                        ; "let restrictions = \
-                           Db_actions.DB_Action.Pool.get_restrictions \
-                           ~__context ~self:pool in "
-                        ; "let vendor_device_allowed = try List.assoc \
-                           \"restrict_pci_device_for_auto_update\" \
-                           restrictions = \"false\" with _ -> false in"
-                        ; "let policy_says_its_ok = not \
-                           (Db_actions.DB_Action.Pool.get_policy_no_vendor_device \
-                           ~__context ~self:pool) in"
-                        ; "vendor_device_allowed && policy_says_its_ok) with e \
-                           -> D.error \"Failure when defaulting \
-                           has_vendor_device field: %s\" (Printexc.to_string \
-                           e); Rpc.Bool false)"
-                        ]
-                    , VBool false
-                    )
-                 )
-              )
-            ~ty:Bool "has_vendor_device"
+            ~doc_tags:[Windows] ~default_value:(Some (VBool true)) ~ty:Bool
+            "has_vendor_device"
             "When an HVM guest starts, this controls the presence of the \
              emulated C000 PCI device which triggers Windows Update to fetch \
              or update PV drivers."
--- a/ocaml/idl/datamodel_vm.ml
+++ b/ocaml/idl/datamodel_vm.ml
@@ -2092,7 +2092,14 @@ let t =
             "hardware_platform_version"
             "The host virtual hardware platform version the VM can run on"
         ; field ~qualifier:StaticRO
-            ~lifecycle:[(Published, rel_dundee, "")]
+            ~lifecycle:
+              [
+                (Published, rel_dundee, "")
+              ; ( Changed
+                , "24.14.0"
+                , "New default and not consulting Pool.policy_no_vendor_device"
+                )
+              ]
             ~doc_tags:[Windows] ~default_value:(Some (VBool true)) ~ty:Bool
             "has_vendor_device"
             "When an HVM guest starts, this controls the presence of the \
--- a/ocaml/xapi/import.ml
+++ b/ocaml/xapi/import.ml
@@ -629,17 +629,15 @@ module VM : HandlerTools = struct
               ~domain_type:vm_record.API.vM_domain_type
               ~is_a_template:vm_record.API.vM_is_a_template
               vm_record.API.vM_platform
+        ; API.vM_suspend_VDI= Ref.null
+        ; API.vM_power_state= `Halted
         }
       in
       let vm =
         log_reraise
           ("failed to create VM with name-label " ^ vm_record.API.vM_name_label)
           (fun value ->
-            let vm =
-              Xapi_vm_helpers
-              .create_from_record_without_checking_licence_feature_for_vendor_device
-                ~__context rpc session_id value
-            in
+            let vm = Client.VM.create_from_record ~rpc ~session_id ~value in
             if config.full_restore then
               Db.VM.set_uuid ~__context ~self:vm ~value:value.API.vM_uuid ;
             vm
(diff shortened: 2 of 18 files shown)


ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -154,18 +151,13 @@ let update_vm_virtual_hardware_platform_version ~__context ~vm =
 
 let create_from_record_without_checking_licence_feature_for_vendor_device
```
robhoes, line 152: This whole function no longer serves a purpose now. We should remove it, and have its callers call `Client.VM.create_from_record` directly.

ocaml/xapi/xapi_vm_clone.ml
```diff
@@ -246,10 +246,9 @@ let snapshot_metadata ~__context ~vm ~is_a_snapshot =
   else
     ""
 
-(* return a new VM record, in appropriate power state and having the good metrics. *)
-(* N.B. always check VM.has_vendor_device and Features.PCI_device_for_auto_update before calling this,
- * as is done before the single existing call to this function.
- * If ever we need to expose this function in the .mli file then we should do the check in the function. *)
+(* return a new VM record, in appropriate power state and having the
+   good metrics.  If ever we need to expose this function in the .mli
```
robhoes, line 250: The last sentence ("If ever we...") can also be removed now.

ocaml/xapi/xapi_vm_helpers.ml
lindig: What about this part?
```
update_vm_virtual_hardware_platform_version ~__context ~vm ;
```

robhoes: That looks redundant, as `VM.create` does the same. The `update_vm_virtual_hardware_platform_version ` function depends on the current `has_vendor_device` value of the VM, which is why the function-with-the-long-name had call it again after setting the field.

minglumlu:
A typo in commit message: lincensed -> licensed

minglumlu:
The feature related logic appears in https://github.com/xapi-project/xen-api/blob/master/ocaml/idl/datamodel_vm.ml#L2093.
It's in datamodel, and it seems the value will be determined by the feature here also as these are OCaml code lines.

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -154,18 +151,13 @@ let update_vm_virtual_hardware_platform_version ~__context ~vm =
...
-    has_vendor_device
-    && not
-         (Pool_features.is_enabled ~__context
-            Features.PCI_device_for_auto_update
-         )
-  then (
+  if has_vendor_device then (
     (* Avoid the licence feature check which is enforced in VM.create (and create_from_record). *)
     let vm = mk_vm {vm_record with API.vM_has_vendor_device= false} in
     Db.VM.set_has_vendor_device ~__context ~self:vm ~value:true ;
```
minglumlu, line 163: When the license checking is removed, the settings on `has_vendor_device` (to `false` and then to `true`) are not necessary. So they can be removed.
But I can't see the relationship between the feature and `hardware_platform_version`.

lindig:
> The feature related logic appears in https://github.com/xapi-project/xen-api/blob/master/ocaml/idl/datamodel_vm.ml#L2093. It's in datamodel, and it seems the value will be determined by the feature here also as these are OCaml code lines.

I am puzzled about this code; the default value is some computed value. 
```
        ; field ~qualifier:StaticRO
            ~lifecycle:[(Published, rel_dundee, "")]
            ~doc_tags:[Windows]
            ~default_value:
              (Some
                 (VCustom
                    ( String.concat "\n"
                        [
                          "(try Rpc.Bool ("
                        ; "let pool = List.hd \
                           (Db_actions.DB_Action.Pool.get_all ~__context) in"
                        ; "let restrictions = \
                           Db_actions.DB_Action.Pool.get_restrictions \
                           ~__context ~self:pool in "
                        ; "let vendor_device_allowed = try List.assoc \
                           \"restrict_pci_device_for_auto_update\" \
                           restrictions = \"false\" with _ -> false in"
                        ; "let policy_says_its_ok = not \
                           (Db_actions.DB_Action.Pool.get_policy_no_vendor_device \
                           ~__context ~self:pool) in"
                        ; "vendor_device_allowed && policy_says_its_ok) with e \
                           -> D.error \"Failure when defaulting \
                           has_vendor_device field: %s\" (Printexc.to_string \
                           e); Rpc.Bool false)"
                        ]
                    , VBool false
                    )
                 )
              )
            ~ty:Bool "has_vendor_device"
            "When an HVM guest starts, this controls the presence of the \
             emulated C000 PCI device which triggers Windows Update to fetch \
             or update PV drivers."
```

lindig:
Sure. Let's go over it in person briefly so we have a common understanding and I'll complete it.

lindig:
Passed BVT, BST except some DMC problem.

lindig:
@robhoes please take another look

ocaml/idl/datamodel_vm.ml
```diff
@@ -2093,34 +2093,8 @@ let t =
...
-                        ; "let pool = List.hd \
-                           (Db_actions.DB_Action.Pool.get_all ~__context) in"
-                        ; "let restrictions = \
-                           Db_actions.DB_Action.Pool.get_restrictions \
-                           ~__context ~self:pool in "
-                        ; "let vendor_device_allowed = try List.assoc \
-                           \"restrict_pci_device_for_auto_update\" \
-                           restrictions = \"false\" with _ -> false in"
-                        ; "let policy_says_its_ok = not \
-                           (Db_actions.DB_Action.Pool.get_policy_no_vendor_device \
```
robhoes, line 2112: I missed the use of `Pool.policy_no_vendor_device` when looking at this earlier. It is defined in datamodel_pool.ml as:
```
        ; field ~qualifier:RW ~in_product_since:rel_dundee
            ~default_value:(Some (VBool false)) ~ty:Bool
            "policy_no_vendor_device"
            "The pool-wide policy for clients on whether to use the vendor \
             device or not on newly created VMs. This field will also be \
             consulted if the 'has_vendor_device' field is not specified in \
             the VM.create call."
```
We need to check if this is still needed for some reason.

lindig: It looks to me `policy_no_vendor_device` is not used outside this code except for tests. I do think we should try to remove that field as well because it is confusing. The problem is, that customers probably use it.

lindig:
Should we add some note to `Pool.policy_no_vendor_device` that this field is now ignored?

robhoes:
Yes, let's mark it as deprecated with a note about `VM.create`.

lindig:
Moved to Dart again because I can't work on it today

lindig:
@robhoes I've updated the API documentation.

ocaml/xapi/import.ml
```diff
@@ -635,11 +635,7 @@ module VM : HandlerTools = struct
         log_reraise
           ("failed to create VM with name-label " ^ vm_record.API.vM_name_label)
           (fun value ->
-            let vm =
-              Xapi_vm_helpers
-              .create_from_record_without_checking_licence_feature_for_vendor_device
-                ~__context rpc session_id value
-            in
+            let vm = Client.VM.create_from_record ~rpc ~session_id ~value in
```
psafont, line 638: Previously, some values were overridden:
```ocaml
~value:{r with API.vM_suspend_VDI= Ref.null; API.vM_power_state= `Halted}
```
Shouldn't this still be done? Both here and in xapi_vm_snapshot

robhoes: That override came from 07b0fd4d509825277d44258da29de37d1d758aff.

lindig: Yes, will update

psafont (approved):

robhoes (approved):
Looks good. Let's squash the fixup and merge.

PR #5180, opened 2023-09-26
Clean up after unit test: don't leave gigabytes of files in /tmp

edwintorok:
`du -csh /tmp/*.vhd` has shown about 1.3GB of files, and this also caused the readonly unit test to fail:
```
[pid 1698168] openat(AT_FDCWD, "/tmp/parse_test.1.vhd", O_RDWR|O_CREAT|O_TRUNC, 0644) = -1 EACCES (Permission denied)

(* Check we respect RO-ness *)
let check_readonly () =
  let filename = make_new_filename () in
  Vhd_IO.create_dynamic ~filename ~size:0L () >>= fun vhd ->
  Vhd_IO.close vhd >>= fun () ->
  Unix.chmod filename 0o444 ;
  Vhd_IO.openchain filename false >>= fun vhd -> Vhd_IO.close vhd
```

### DIFF
--- a/ocaml/libs/vhd/vhd_format_lwt_test/parse_test.ml
+++ b/ocaml/libs/vhd/vhd_format_lwt_test/parse_test.ml
@@ -298,12 +298,6 @@ let all_program_tests =
     programs
 
 let _ =
-  let verbose = ref false in
-  Arg.parse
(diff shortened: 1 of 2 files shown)


edwintorok:
Hmm there are still 1.3G of files in /tmp, some more cleanup is needed

psafont (approved):

robhoes (approved):

PR #5181, opened 2023-09-28
CA-381047: Add observer capability to bugtool

snwoods:
Move the tracing log output to its own capability as it spans more than just xapi. This also allows the max size to be increased without increasing it for all of xapi.

### DIFF
--- a/scripts/Makefile
+++ b/scripts/Makefile
@@ -33,10 +33,13 @@ install:
 	$(IPROG) nbd_client_manager.py $(DESTDIR)$(LIBEXECDIR)
 	mkdir -p $(DESTDIR)$(ETCXENDIR)/bugtool/xapi
 	mkdir -p $(DESTDIR)$(ETCXENDIR)/bugtool/xenopsd
+	mkdir -p $(DESTDIR)$(ETCXENDIR)/bugtool/observer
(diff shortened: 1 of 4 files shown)


psafont (approved):

mg12 (approved):

scripts/bugtool-plugin/xapi.xml
```diff
@@ -1 +1 @@
-<capability pii="yes" max_size="16384" max_time="60" mime="text/plain" checked="true"/>
+<capability pii="yes" max_size="1638400" max_time="60" mime="text/plain" checked="true"/>
```
mg12, line 1: It looks like this limit applies to all files in the directory together, not individual files: https://github.com/xenserver/status-report/blob/master/xen-bugtool#L517
So this limit is not enough for the up to 100MB of files in /var/log/dt. We either need to increase this value further, or create a new capability of max_size=100MB only for /var/log/dt

mg12 (changes requested):

lindig:
Nit picking comment: the subject line could be made to fit: Increase bugtool size limit to <number>. RIght now one would wonder about the limit for that and its value.

mg12 (approved):

PR #5182, opened 2023-09-29
Futurize rrdd.py for Python3

MarkSymsCtx:


psafont (approved):

Makefile
```diff
@@ -142,6 +142,8 @@ install: build doc sdk doc-json
 # xcp-rrdd
 	install -D _build/install/default/bin/xcp-rrdd $(DESTDIR)/usr/sbin/xcp-rrdd
 	install -D _build/install/default/bin/rrddump $(DESTDIR)/usr/bin/rrddump
+	install -D -m 644 ocaml/xcp-rrdd/scripts/rrdd/rrdd.py $(DESTDIR)/usr/lib/python2.7/site-packages
+	install -D -m 644 ocaml/xcp-rrdd/scripts/rrdd/rrdd.py $(DESTDIR)/usr/lib/python3.6/site-packages
```
MarkSymsCtx, line 146: This should only have been adding the 3.6 one as 2.7 was added in d3a89eb9cd233186b06830c89fe29f803516a94f, unless that subsequently moved after the last release?

psafont: This is done at the spec level now, according to the change that removed it https://github.com/xapi-project/xen-api/commit/40d75db7d910e96e420eb027176c64fb81fe9565

MarkSymsCtx: OK, I'll fix that up although I couldn't see anything in the spec that was doing that so we might have a gap between source and spec versions.

ocaml/xcp-rrdd/scripts/rrdd/rrdd.py
```diff
@@ -44,267 +44,288 @@
...
+                             timestamp))
+        for val in data_values:
+            # This is already big endian encoded
+            self.dest.write(val)
+
+        self.dest.write(pack(">L", len(metadata_json)))
+        self.dest.write(metadata_json.encode())
+        self.dest.flush()
+        self.datasources = []
+        time.sleep(
```
edwintorok, line 326: why the linebreak here?

edwintorok (approved):

ocaml/xcp-rrdd/scripts/rrdd/rrdd.py
MarkSymsCtx: I guess the reformat (which was automated) split the line.

lindig:
Can we merge this?

PR #5183, opened 2023-09-29
Publish new Hugo-based docs

robhoes:
This Github Action publishes the new Hugo-based docs from /doc to a temporary staging area at https://xapi-project.github.io/new-docs/. Eventually, this will replace the old docs on https://xapi-project.github.io. However, we are still transferring docs to the new world, which we do incrementally so that it all can be updated and reviewed again.

### DIFF
--- a/.github/workflows/hugo.yml
+++ b/.github/workflows/hugo.yml
@@ -1,0 +1,37 @@
+name: Generate and upload Hugo docs
+
+on:
+  push:
+    branches: master
+
+jobs:
+  ocaml:
+    name: Docs
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Setup Hugo
+        uses: peaceiris/actions-hugo@v2
+        with:
+          hugo-version: '0.119.0'
+
+      - name: Build
+        run: |
+          cd doc
+          hugo --minify
+
+      - name: Deploy
+        uses: peaceiris/actions-gh-pages@v3
+        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/hugo.yml
```diff
@@ -0,0 +1,37 @@
...
+    branches: master
+
+jobs:
+  ocaml:
+    name: Docs
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
```
edwintorok, line 14: There is an @v4 now

edwintorok (approved):

.github/workflows/hugo.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Build
+        run: |
+          cd doc
+          hugo --minify
+
+      - name: Deploy
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          deploy_key: ${{ secrets.ACTIONS_DOCS_DEPLOY_KEY }}
```
psafont, line 29: I think this should be part of the docs.yml workflow so it's easy to see both are deploying to the same repo.

robhoes: I wasn't sure what the best practice is... But they aren't actually deploying to the same repo. The other one goes to xapi-storage. And the API docs are pulled by xapi-project.github.io.

robhoes: Perhaps we'll merge all that at some point.

lindig (approved):

.github/workflows/hugo.yml
robhoes: Talked to @psafont... we'll merge all docs-related Actions later.

PR #5188, opened 2023-10-05
Allow SDK Consumers to create a custom implementation for JsonRpcClient

MichaelSabin-Citrix:
Some consumers of the Xen SDK may have use cases where they need custom handling of the HTTP requests to be made to the endpoint server. This change will allow the JsonRpc class to be overridden and provide a custom implementation of the PerformPostRequest method.


kc284:
Ah, I forgot: could you just reword the commit message to include the ticket number for traceability: CP-44367. Thanks.

MichaelSabin-Citrix:
Updated. Thanks.

danilo-delbusso (approved):

PR #5190, opened 2023-10-06
CA-383867: xapi-guard: retry on XAPI communication failure

edwintorok:
Try to prevent the TPM from going into failure mode when XAPI is down (e.g. due to a toolstack restart or host reboot).

Before this change running a command like 'Initialize-Tpm' inside a Win11 guest caused the vTPM to go into failure mode due to the communication error with xapi, and it would only recover after a VM reboot. Using TPM_RETRY instead of TPM_FAIL as the failure code doesn't seem to make a difference, the TPM still goes into a fatal failure mode.

After this change a 'systemctl stop xapi', 'Initialize-Tpm' inside a Win11 guest gets stuck, 'systemctl start xapi' - it starts working again.

An 'xe-toolstack-restart' still causes the TPM to go into a failure mode due to the temporary interruption of comms between xapi-guard and swtpm but that can be worked around by retyring on the swtpm side.

### DIFF
--- a/ocaml/xapi-guard/lib/server_interface.ml
+++ b/ocaml/xapi-guard/lib/server_interface.ml
@@ -50,7 +50,22 @@ let () =
    * this is only needed for syscalls that would otherwise block *)
   Lwt_unix.set_pool_size 16
 
-let with_xapi ~cache f =
+let rec with_retry ?(retry_max = 5.) ~retry f =
+  match retry with
+  | Some retry ->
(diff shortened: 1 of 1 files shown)


edwintorok:
(May want to have a different upper bound than 5s, I'm running a test to see how long a VM can survive without a reply from the vTPM)

edwintorok:
This won't work: the guest eventually times out (fairly soon, couple of minutes).

PR #5192, opened 2023-10-06
Add some more documents from xapi-project.github.io

robhoes:


edwintorok (approved):
We should probably merge this as is (it is just moving around existing docs) and then do improvements as separate PRs on top instead of trying to update or correct any out of date information in this one.

robhoes:
I did do a quick review myself and fixed some things that were clearly wrong. I did not review things like the HA docs in detail, but then not much has probably changed there in recent years.

psafont (approved):

PR #5193, opened 2023-10-09
CP-45214: Fix tracing HTTP request not working with Jaegers FQDN by f…

snwoods:
…lushing instead of closing the sending stream

### DIFF
--- a/ocaml/libs/tracing/tracing.ml
+++ b/ocaml/libs/tracing/tracing.ml
@@ -684,9 +684,9 @@ module Export = struct
           let headers =
             Cohttp.Header.of_list
               [
-                ("accepts", "application/json")
-              ; ("content-type", "application/json")
-              ; ("content-length", string_of_int (String.length body))
+                ("Content-Type", "application/json")
+              ; ("Content-Length", string_of_int (String.length body))
+              ; ("Host", "jaeger.xenrt.citrite.net:9411")
               ]
           in
           Open_uri.with_open_uri url (fun fd ->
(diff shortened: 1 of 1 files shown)


ocaml/libs/tracing/tracing.ml
```diff
@@ -684,9 +684,9 @@ module Export = struct
           let headers =
             Cohttp.Header.of_list
               [
-                ("accepts", "application/json")
-              ; ("content-type", "application/json")
+                ("content-type", "application/json")
               ; ("content-length", string_of_int (String.length body))
+              ; ("host", "jaeger.xenrt.citrite.net:9411")
```
edwintorok, line 689: should we use the capitalised version here? Although it should be case insensitive I think you found an article where some proxies are buggy

edwintorok (approved):

ocaml/libs/tracing/tracing.ml
```diff
@@ -684,9 +684,9 @@ module Export = struct
           let headers =
             Cohttp.Header.of_list
               [
-                ("accepts", "application/json")
-              ; ("content-type", "application/json")
-              ; ("content-length", string_of_int (String.length body))
+                ("Content-Type", "application/json")
+              ; ("Content-Length", string_of_int (String.length body))
+              ; ("Host", "jaeger.xenrt.citrite.net:9411")
```
psafont, line 689: Why is the host hardcoded? Is this overwritten in case other hostnames are used?

PR #5194, opened 2023-10-09
Do not fetch random SR in empty list

benjamreis:
When looking for a random SR in an empty list,
quicktest can call `Random.int 0` which will throw. Instead return an empty list.
Instead of throwing, the test won't be passed as there is no compatible SR.

This can happen when calling quickest with the `-sr` on ISO SR.

### DIFF
--- a/ocaml/quicktest/qt_filter.ml
+++ b/ocaml/quicktest/qt_filter.ml
@@ -198,8 +198,11 @@ module SR = struct
 
   let random srs () =
     let srs = srs () in
-    let index = Random.int @@ List.length srs in
-    [List.nth srs index]
+    if srs = [] then
+      []
+    else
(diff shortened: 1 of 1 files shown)


ocaml/quicktest/qt_filter.ml
```diff
@@ -198,8 +198,12 @@ module SR = struct
 
   let random srs () =
     let srs = srs () in
-    let index = Random.int @@ List.length srs in
-    [List.nth srs index]
+    let length = List.length srs in
+    if length = 0 then
```
robhoes, line 202:
```suggestion
    if srs = [] then
```

robhoes (approved):
Looks fine to me, but please squash the commits.

psafont (approved):

PR #5195, opened 2023-10-09
Benchmark code for scalability of XAPI

edwintorok:
Using bechamel to benchmark (it requires this draft version to fix a double-free bug https://github.com/mirage/bechamel/pull/47).

First add some convenience modules that avoids the boilerplate in setting up, running and analyzing benchmarks.
Then add some building blocks:
* overhead of 'ignore'
* fixed work C stub that releases runtime lock
* pingpong tests for Event and Semaphore
* barriers (not provided natively by OCaml). Several ways to implement them, currently the fastest one seems to be to use an array of binary semaphores (which is about twice as fast than using a single counting semaphore, probably due to the OCaml master lock in OCaml 4.x)
* a helper to run benchmarks in worker threads concurrently
* run the fixed work using the concurrent helper

This prepares the code for benchmarking PAM start/stop and its scalability by knowing ahead of time the scalability of basic OCaml building blocks. E.g. one of the machines I used for benchmarking shows very bad scalability with 4 threads having 30% overhead and 8 threads having 70% overhead, whereas another machine showed only an 5% overhead.
Further investigation is needed on why that particular machine is so bad, but for PAM I'll use the other machine which shows better scalability for the 'fixed work' C stub: that'll likely expose the maximum parallelism available in hardware,
making any scalability issues in our PAM bindings more noticable.

They are also useful baselines: the PAM code can only scale as well as this basic 'fixed work' benchmark, any scalability limits in this basic code will also be present in the PAM code through no fault of that code.

Draft PR because we'll need an updated version of bechamel (and a few more dependencies: bechamel-notty, notty, bechamel-js).


edwintorok:
one of the benchmarks also has barriers implemented in a bad way just to see whether it is indeed slower and it is (it uses yield, but yield doesn't target  a specific thread so it is possible to yield to a thread that has just yielded previously and will immediately yield back... kind of like the Xen bug that got recently fixed).
FWIW that approach works well with OCaml Domains in OCaml 5 where instead of yield you can busy loop with cpu_relax and achieve much lower wakeup latencies. But on OCaml 4 there is no equivalent approach.

ocaml/auth/xa_auth_stubs.c
```diff
@@ -70,6 +71,95 @@ CAMLprim value stub_XA_mh_chpasswd(value username, value new_password){
...
+
+  /* guard against use after free and double free */
+  *((pam_handle_t** ) Data_abstract_val(handle)) = NULL;
+
+  caml_enter_blocking_section();
+  rc = XA_mh_authorize_stop(pamh, &error);
+  caml_leave_blocking_section();
+
+  if (XA_SUCCESS != rc)
+    caml_failwith(error ? error : "Unknown error");
```
lindig, line 117: Same here.

ocaml/auth/xa_auth_stubs.c
```diff
@@ -70,6 +71,95 @@ CAMLprim value stub_XA_mh_chpasswd(value username, value new_password){
...
+
+  CAMLreturn(Val_unit);
+}
+
+CAMLprim value stub_XA_mh_authorize_run(value ml_handle, value username, value password){
+    CAMLparam2(username, password);
```
lindig, line 123: There are 3 arguments but only two are declared here.

ocaml/auth/xa_auth_stubs.c
```diff
@@ -70,6 +71,95 @@ CAMLprim value stub_XA_mh_chpasswd(value username, value new_password){
...
+    CAMLlocal1(ret);
+    pam_handle_t *handle = pam_handle_of_val(ml_handle);
+    ret = Val_unit;
+
+    char *c_username = strdup(String_val(username));
```
lindig, line 128: How about `strdupa(3)`

ocaml/auth/xa_auth_stubs.c
```diff
@@ -70,6 +71,95 @@ CAMLprim value stub_XA_mh_chpasswd(value username, value new_password){
...
+    rc = XA_mh_authorize_run(&handle, c_username, c_password, &error);
+    free(c_username);
+    free(c_password);
+    caml_leave_blocking_section();
+
+    /* must only be done with the runtime lock held again */
+    *((pam_handle_t** ) Data_abstract_val(ml_handle)) = handle;
+
+    if (rc != XA_SUCCESS)
+        caml_failwith(error ? error : "Unknown error");
```
lindig, line 143: Use `__LINE__` or `__FUNCTION__` (not sure this is available).

lindig (changes requested):
This is a lot of code; I did not look at all of it so far.

edwintorok:
This needs some small updates (fix conflicts, and the C stub argument errors). Closing for now as discussed.

PR #5198, opened 2023-10-11
CP-43755: parallelise PAM logins

edwintorok:
A quick test shows this is about twice as fast than with 1 thread, although [the benchmark PR ](https://github.com/xapi-project/xen-api/pull/5195) shows that it should be able to scale much better.
I'm getting ~9 logins/s, and should be getting ~3000 with the new code.


edwintorok:
(draft PR, running some tests)

edwintorok:
Tests look good, no more sleeps either. The remaining bottleneck appears to be in jemalloc (sha512_crypt_r spends >90% time allocating and deallocating memory), but that'll be fixed in another PR (I'll try running my benchmarks with jemalloc).

ocaml/auth/xa_auth_stubs.c
```diff
@@ -70,6 +71,30 @@ CAMLprim value stub_XA_mh_chpasswd(value username, value new_password){
     CAMLreturn(ret);
 }
 
+#include <crypt.h>
+/* 'constructor' attribute will ensure this function gets call early during program startup. */
+void __attribute__((constructor)) stub_XA_workaround(void)
```
robhoes, line 76: I don't see an ml file referring to this stub. What actually calls it?

ocaml/xapi/locking_helpers.ml
```diff
@@ -223,3 +223,55 @@ module Named_mutex = struct
         finally f (fun () -> Thread_state.released r)
     )
 end
+
+module Semaphore = struct
+  type t = {
+      name: string
+    ; sem: Semaphore.Counting.t
```
robhoes, line 230: We are using `Xapi_stdext_threads.Semaphore` elsewhere. Can we replace all uses of that module with the standard library's module?

robhoes (approved):

robhoes:
Please commit the datamodel_lifecycle changes.

ocaml/auth/xa_auth_stubs.c
edwintorok: I was thinking about adding a stub and a call in the module initialization `let () =...` but this is simpler.
The constructor gets called automatically when the shared library or process is starting up, and at that point you haven't created any threads yet so there are no race conditions.

ocaml/xapi/locking_helpers.ml
edwintorok: The standard library's semaphore implementation is similarly with mutex+condvar, so yes. It doesn't have an 'execute' function though, and can only acquire 1 resource at a time (but I only see calls to acquire with the value of '1', so should be fine).
The one in stdext also has a spurius Condition.signal inside acquire that'll wake up threads for no good reason (it has taken away some resources so if they were sleeping before they still won't be able to acquire them).

Perhaps we could change Xapi_stdext.Semaphore to use the semaphore module as its implementation and act as a lightweight wrapper that provides the other convenience functions (execute), etc.

ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -55,6 +55,10 @@ let prototyped_of_field = function
       Some "23.9.0"
   | "pool", "telemetry_uuid" ->
       Some "23.9.0"
+  | "pool", "ext_auth_max_threads" ->
+      Some "23.25.0-next"
```
robhoes, line 59: We have tag v23.26.0 already, so I think you need to fetch the tags and run this again.

edwintorok:
```
git describe
v23.26.0-7-gd90369deb
```
something goes wrong on why it picks up 25 instead

edwintorok: This is due to the dune cache, it needs to be told that 'gen_lifecycle' depends on something external (i.e. 'the universe'), otherwise it'll use a cached version that still uses the old version number.
I pushed a commit that makes the dune change and now it prints the right version.

ocaml/xapi/locking_helpers.ml
psafont: We should change the implementation of the module with the stand library one

ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -91,8 +95,14 @@ let prototyped_of_message = function
       Some "22.26.0"
   | "VTPM", "create" ->
       Some "22.26.0"
+  | "host", "apply_recommended_guidances" ->
+      Some "23.18.0"
```
psafont, line 99: This addition doesn't seem right

robhoes: This surprised me as well, but I think it's ok... The datamodel has an explicit:

        ~lifecycle:[(Prototyped, "23.18.0", ""); (Removed, "23.25.0", "")]

I was perhaps manually removed from datamodel_lifecycle.ml and then added to datamodel_host.ml, but the automation put it back.

robhoes (approved):

psafont (approved):

PR #5199, opened 2023-10-12
Do not attempt to start snapshots or templates

benjamreis:
`start` operations are not allowed for these VMs

Fixes: https://github.com/xcp-ng/xcp/issues/625


benjamreis:
You can see [here](https://xcp-ng.org/forum/post/66428) one of our user encountered the issue.
Filtering out snapshots and templates for autostartup seems user friendlier to me than raising `VM_IS_{SNAPSHOT/TEMPLATE}` and stop starting other VM after a host reboot.

psafont (approved):

robhoes (approved):

PR #5201, opened 2023-10-12
CA-371002: Do a usual import when a default template cannot be found

Vincent-lau:


edwintorok:
You'll need to update 'quality-gate.sh' (you've improved the code quality by deleting one List.hd):
```
ERROR expected 321 List.hd usages, got 320
```

ocaml/xapi/import.ml
```diff
@@ -427,82 +427,84 @@ module VM : HandlerTools = struct
...
-                       Ref.string_of vm
-                     ; Int64.to_string existing_version
-                     ; Int64.to_string version_to_import
-                     ]
-                   )
-                )
+
+    let maybe_template =
+      List.nth_opt
+        (Db.VM.get_by_name_label ~__context ~label:vm_record.API.vM_name_label)
```
edwintorok, line 433: 'List.nth_opt' works, although since you have a match below another way to write this would've been to assign `let maybe_templates` and then pattern match that it has at least one element, i.e. `true, template :: _`.
But you can also leave the code as it is.

ocaml/xapi/import.ml
```diff
@@ -427,82 +427,94 @@ module VM : HandlerTools = struct
...
+                       ; Int64.to_string existing_version
+                       ; Int64.to_string version_to_import
+                       ]
+                     )
+                  )
+              else
+                Replace (vm, vm_record)
+          else
+            let vm_record =
+              if not config.force then
```
minglumlu, line 497: Probably the `re-imported` default template would be of same name or UUID. The logic flow having reached here means a VM (template) with a non-existing name and a non-existing UUID is being asked to be imported.
I think the `is_default_template` has clearly indicated this should be imported as a default template. In other words, may I please know if the `config.force` is deliberately involved here for some consideration?

Vincent-lau: This is so that when the force option is not enabled, we make the default template non-default. We do this because importing a default template (with no UUID/names) will make it non-deletable afterwards, which is probably not ideal for customers.

Vincent-lau: The reason (as far as I have seen) why the imported default template does not have a UUID/name is because it has been exported (with changes) and reimported by a user. To me it feels that this reimported template should no longer be considered default since the user has made changes after exporting the template.

robhoes (approved):

Issue #5202, opened 2023-10-13
Dynamical `Host.evacuate` batch size


benjamreis:
Hello!

Here's a feature request a user asked us about: being able to choose how many VMs are migrating at the same time while evacuating a host.

I'm a aware of the `evacuation-batch-size` option in `xapi.conf` but we think it would be user friendlier to be able to choose the batch size at each `evacuate` call.

So I was thinking about adding an optionnal `batch-size` parameter to the `evacuate` call that would override the option when fed to the method.

What do you think about this?

Regards.

PR #5203, opened 2023-10-13
Choose size of batch VM evacuation

benjamreis:
New optional argument to `Host.evacuate`: `evacuate_batch_size` When provided uses it instead of `xapi.conf`'s `evacuation_batch_size` When not provided uses the `xapi.conf` option

Fixes: https://github.com/xapi-project/xen-api/issues/5202

### DIFF
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -356,6 +356,7 @@ let evacuate =
       [
         (Published, rel_miami, "")
       ; (Extended, "1.297.0", "Enable migration network selection.")
+      ; (Extended, "23.27.0", "Choose batch size of VM evacuation.")
       ]
     ~versioned_params:
       [
@@ -372,6 +373,15 @@ let evacuate =
         ; param_doc= "Optional preferred network for migration"
         ; param_release= next_release
         ; param_default= Some (VRef null_ref)
+        }
+      ; {
+          param_type= Int
+        ; param_name= "evacuate_batch_size"
+        ; param_doc=
+            "The maximum number of VMs to be migrated per batch 0 will use the \
+             value `evacuation-batch-size` defined in xapi.conf"
+        ; param_release= next_release
+        ; param_default= Some (VInt 0L)
         }
       ]
     ~allowed_roles:(_R_POOL_OP ++ _R_CLIENT_CERT)
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -3387,14 +3387,14 @@ functor
           (host_uuid ~__context self) ;
         Local.Host.get_vms_which_prevent_evacuation ~__context ~self
 
-      let evacuate ~__context ~host ~network =
+      let evacuate ~__context ~host ~network ~evacuate_batch_size =
         info "Host.evacuate: host = '%s'" (host_uuid ~__context host) ;
         (* Block call if this would break our VM restart plan (because the body of this sets enabled to false) *)
         Xapi_ha_vm_failover.assert_host_disable_preserves_ha_plan ~__context
           host ;
         Xapi_host_helpers.with_host_operation ~__context ~self:host
           ~doc:"Host.evacuate" ~op:`evacuate (fun () ->
-            Local.Host.evacuate ~__context ~host ~network
+            Local.Host.evacuate ~__context ~host ~network ~evacuate_batch_size
         )
 
       let retrieve_wlb_evacuate_recommendations ~__context ~self =
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -607,7 +607,7 @@ let compute_evacuation_plan ~__context ~host =
            Using original algorithm" ;
         compute_evacuation_plan_no_wlb ~__context ~host ()
 
-let evacuate ~__context ~host ~network =
+let evacuate ~__context ~host ~network ~evacuate_batch_size =
   let plans = compute_evacuation_plan ~__context ~host in
   let plans_length = float (Hashtbl.length plans) in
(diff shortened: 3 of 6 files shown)


ocaml/xapi/xapi_host.ml
```diff
@@ -709,7 +709,19 @@ let evacuate ~__context ~host ~network =
...
+        (abs
+           ( match evacuate_batch_size with
+           | Some size ->
+               size
+           | None ->
+               !Xapi_globs.evacuation_batch_size
+           )
+        )
+    )
+  in
```
psafont, line 724: I tend to prefer defining values before using them in expression as they can become unwieldy and more difficult to understand otherwise:
```suggestion
  let batch_size =
    Option.value
      ~default:!Xapi_globs.evacuation_batch_size
      evacuation_batch_size
  in
  (* avoid edge cases from meaningless batch sizes *)
  let batch_size = Int.(max 1 (abs batch_size)) in
```

ocaml/xapi/message_forwarding.ml
```diff
@@ -3387,14 +3387,14 @@ functor
           (host_uuid ~__context self) ;
         Local.Host.get_vms_which_prevent_evacuation ~__context ~self
 
-      let evacuate ~__context ~host ~network =
+      let evacuate ~__context ~host ~network ?evacuate_batch_size =
```
psafont, line 3390: It's not a good idea to add optional arguments to a function without non-labelled arguments, it can lead to partially-applied functions. Unfortunately, a unit parameter can't just be added here.

I recommend making it a labelled argument, the end result is the same: users have to define whether they want to set a custom value of the parameter with an `Option` and this has to get treated in the function body, but the possible errors are avoided.

Note that on the API side it's still defined as an optional argument, this is all about how it's treated / represented on the ocaml side

psafont (changes requested):

ocaml/xapi/message_forwarding.ml
benjamreis: When using an `Option` I get `Fatal error: exception Failure("option types never stored in the database")`

So i think I'll give int64 and ignore when it 0L

psafont: My point was to _not_ change the IDL, it was fine as it was

ocaml/idl/datamodel_host.ml
```diff
@@ -373,6 +374,13 @@ let evacuate =
         ; param_release= next_release
         ; param_default= Some (VRef null_ref)
         }
+      ; {
+          param_type= Int
+        ; param_name= "evacuate_batch_size"
+        ; param_doc= "Number of VM to evacuate at the same time"
```
minglumlu, line 380: How about `The maximum number of VMs to be migrated per batch`?

ocaml/xapi/message_forwarding.ml
minglumlu: In `datamodel_host.ml`, the new parameter is defined as `param_type= Int`.

minglumlu: It seems this kind of `versioned_params` is always  of `Optional` to support backward compatibility?

ocaml/idl/datamodel_host.ml
benjamreis: Right, it's more accurate. thx :)

ocaml/xapi/message_forwarding.ml
benjamreis: I'm not sure to understand your last comment.
It means in the new arg is optionnal only in the API but is not in the implementation?

minglumlu: Yes. That's what I read from the code. A quick experiment could make it confirmed.

ocaml/idl/datamodel_host.ml
```diff
@@ -373,6 +374,13 @@ let evacuate =
         ; param_release= next_release
         ; param_default= Some (VRef null_ref)
         }
+      ; {
+          param_type= Int
+        ; param_name= "evacuate_batch_size"
+        ; param_doc= "The maximum number of VMs to be migrated per batch"
+        ; param_release= next_release
+        ; param_default= None
```
minglumlu, line 382: From https://github.com/xapi-project/xen-api/blob/master/ocaml/idl/ocaml_backend/gen_server.ml#L67, I understand is that `None` means it is a mandatory parameter.
While this should not be as the old clients may send calls without the new parameter. Then `Some <N>` would be a default value in this case.

ocaml/xapi/xapi_host.ml
```diff
@@ -708,8 +708,15 @@ let evacuate ~__context ~host ~network =
     TaskHelper.set_progress ~__context 1.0
   in
 
+  let batch_size =
+    match evacuate_batch_size with
+    | 0L ->
```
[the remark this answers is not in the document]
benjamreis: indeed

benjamreis:
Okay so last forced push should take into account all comments, i'll let you close the conversations when satisfied.

Meanwhile I tested giving a number and letting the default value take over, both works as expected. :)

ocaml/idl/datamodel_host.ml
```diff
@@ -373,6 +374,13 @@ let evacuate =
         ; param_release= next_release
         ; param_default= Some (VRef null_ref)
         }
+      ; {
+          param_type= Int
+        ; param_name= "evacuate_batch_size"
+        ; param_doc= "The maximum number of VMs to be migrated per batch"
```
lindig, line 380: Should this include something like "Zero (0) will use a default from xapi.conf"? Otherwise the default is surprising.

benjamreis: Done :)

psafont (approved):

PR #5205, opened 2023-10-17
CA-381044: Raise error when `pool.set_update_sync_enabled` is called with `true` and empty repos

danilo-delbusso:
Please be nit-picky :)

### DIFF
--- a/ocaml/xapi-consts/api_errors.ml
+++ b/ocaml/xapi-consts/api_errors.ml
@@ -1283,6 +1283,8 @@ let update_guidance_changed = "UPDATE_GUIDANCE_CHANGED"
 
 let invalid_update_sync_day = "INVALID_UPDATE_SYNC_DAY"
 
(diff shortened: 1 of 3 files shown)


ocaml/xapi-consts/api_errors.ml
```diff
@@ -1283,6 +1283,8 @@ let update_guidance_changed = "UPDATE_GUIDANCE_CHANGED"
 
 let invalid_update_sync_day = "INVALID_UPDATE_SYNC_DAY"
 
+let cannot_enable_update_sync = "CANNOT_ENABLE_UPDATE_SYNC"
```
psafont, line 1286: I like errors to have the reason of the failure, rather than what failed, as that is usually obvious to an API client. I'd prefer to have "NO_REPOSITORIES_CONFIGURED" or something to the same effect

psafont (approved):

lindig (approved):

PR #5206, opened 2023-10-17
CP-41844: Add xs-trace, an executable to submit trace files to an endpoint

snwoods:


ocaml/xs-trace/xs_trace.ml
```diff
@@ -0,0 +1,52 @@
...
+        done
+      with End_of_file -> ()
+    else if Filename.check_suffix orig ".ndjson" then
+      Xapi_stdext_unix.Unixext.readfile_line (fun line -> submit_json line) orig
+    else
+      let json = Xapi_stdext_unix.Unixext.string_of_file orig in
+      submit_json json
+  in
+  export_file origin ;
+  let rec remove_all orig =
```
lindig, line 42: Existing code:

```
let rec rmdir_rec path =
  match Sys.is_directory path with
  | true ->
      Sys.readdir path
      |> Array.iter (fun name -> rmdir_rec (Filename.concat path name)) ;
      Unix.rmdir path
  | false ->
      Sys.remove path
```

ocaml/xs-trace/xs_trace.ml
```diff
@@ -0,0 +1,52 @@
+let _ =
+  if Array.length Sys.argv <> 4 then (
+    Printf.eprintf "Usage: %s cp/mv <origin> <destination>\n" Sys.argv.(0) ;
+    exit 1
+  ) ;
+  let action = Sys.argv.(1) in
```
lindig, line 6: This can fail. Better to match over this:
```
match Sys.argv with
| [|_; action; origin|} ->
| _ -> something is wrong
```

ocaml/xs-trace/test/test_xs_trace.ml
```diff
@@ -0,0 +1,29 @@
+let _ =
+  let inet_addr = Unix.ADDR_INET (Unix.inet_addr_of_string "127.0.0.1", 9411) in
+  let read_body ic file len =
+    let line = really_input_string ic !len in
+    Printf.fprintf file "%s\n" line
+  in
+  let len = ref 0 in
+  let rec read_header ic file =
```
psafont, line 8: I wanted to suggest using the library xapi has to implement  http server, but it's a mess :(

ocaml/xs-trace/xs_trace.ml
snwoods: Good find. I see that's in test_observer. How should I reuse it, move the function to somewhere both files could access it? Where would you recommend?

ocaml/xs-trace/test/test_xs_trace.ml
```diff
@@ -0,0 +1,29 @@
...
+    match input_line ic with
+    | line when line = "\r" ->
+        read_body ic file len
+    | line ->
+        if String.starts_with ~prefix:"Content-Length: " line then
+          (*Subtract one more than the length of Content-Length: so we don't read the trailing \r*)
+          len := int_of_string (String.sub line 16 (String.length line - 17)) ;
```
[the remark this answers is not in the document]
snwoods: As noted in the comment, 16 is the length of "Content-Length: " that we wish to remove, and 17 is one more than that to account for the trailing "/r"

ocaml/xs-trace/test/test_xs_trace.ml
```diff
@@ -0,0 +1,29 @@
+let _ =
+  let inet_addr = Unix.ADDR_INET (Unix.inet_addr_of_string "127.0.0.1", 9411) in
+  let read_body ic file len =
+    let line = really_input_string ic !len in
+    Printf.fprintf file "%s\n" line
+  in
+  let len = ref 0 in
+  let rec read_header ic file =
+    match input_line ic with
+    | line when line = "\r" ->
```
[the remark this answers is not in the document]
snwoods: input_line stops reading at the \n and removes it, so it'll just be \r

ocaml/xs-trace/test/test_xs_trace.ml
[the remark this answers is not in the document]
lindig: I would suggest to bind the string and to compute this:
```
let length_header = "Content-Length: " in (** final space is ok *)
let length_header_len = String.length length_header in
...
```

ocaml/xs-trace/xs_trace.ml
lindig: Yes; move into one of the helper modules; if the helper module is crowded you could also create a local sub module like `Files` or so and put the code inside.

ocaml/xs-trace/xs_trace.ml
```diff
@@ -0,0 +1,58 @@
...
+          Xapi_stdext_unix.Unixext.readfile_line
+            (fun line -> submit_json line)
+            orig
+        else
+          let json = Xapi_stdext_unix.Unixext.string_of_file orig in
+          submit_json json
+      in
+      export_file origin ;
+      if action = "mv" then
+        Xapi_stdext_unix.Unixext.rm_rec ~rm_top:true origin ;
```
snwoods, line 55: This relies on a change to stdext https://github.com/xapi-project/stdext/pull/76

mg12 (approved):

lindig (approved):

PR #5207, opened 2023-10-19
CP-43901: Block pool member startup if it has a higher xapi version

gangj:
During pool upgrade, it is expected that pool coordinator needs to be updated first, and a new xapi should start on the coordinator before it starts on any other host in the pool. Otherwise it is dangerous and may cause unpredictable problem.

This commit is to block pool member startup if it has a higher xapi version than coordinator until coordinator has a equal or higher version of xapi running.

### DIFF
--- a/ocaml/util/xapi_version.ml
+++ b/ocaml/util/xapi_version.ml
@@ -43,3 +43,27 @@ let version, xapi_version_major, xapi_version_minor, git_id =
              version
           )
     )
+
+let compare_version version_a version_b =
+  let maj_a, min_a, _ =
+    try Scanf.sscanf version_a "%d.%d.%s" (fun maj min rest -> (maj, min, rest))
+    with _ ->
+      failwith
+        (Printf.sprintf "Couldn't determine xapi version for version_a: '%s'"
+           version_a
+        )
+  in
+  let maj_b, min_b, _ =
+    try Scanf.sscanf version_b "%d.%d.%s" (fun maj min rest -> (maj, min, rest))
+    with _ ->
+      failwith
+        (Printf.sprintf "Couldn't determine xapi version for version_b: '%s'"
+           version_b
+        )
+  in
+  if maj_a > maj_b || (maj_a = maj_b && min_a > min_b) then
+    1
+  else if maj_a = maj_b && min_a = min_b then
+    0
+  else
+    -1
--- a/ocaml/util/xapi_version.ml
+++ b/ocaml/util/xapi_version.ml
@@ -44,9 +44,5 @@ let version, xapi_version_major, xapi_version_minor, git_id =
 let compare_version version_a version_b =
   let maj_a, min_a, _ = parse_xapi_version version_a in
   let maj_b, min_b, _ = parse_xapi_version version_b in
-  if maj_a > maj_b || (maj_a = maj_b && min_a > min_b) then
-    1
-  else if maj_a = maj_b && min_a = min_b then
-    0
-  else
-    -1
+  let ( <?> ) a b = if a = 0 then b else a in
+  Int.compare maj_a maj_b <?> Int.compare min_a min_b <?> 0
--- a/ocaml/util/xapi_version.ml
+++ b/ocaml/util/xapi_version.ml
@@ -18,11 +18,19 @@ let hostname = "localhost"
 
 let date = Xapi_build_info.date
 
+let parse_xapi_version version =
+  try Scanf.sscanf version "%d.%d.%s" (fun maj min rest -> (maj, min, rest))
+  with _ ->
+    failwith
+      (Printf.sprintf "Couldn't determine xapi version from string: '%s'"
+         version
+      )
+
 let version, xapi_version_major, xapi_version_minor, git_id =
   match Build_info.V1.version () with
   | None ->
       ("0.0.dev", 0, 0, "dev")
-  | Some v -> (
+  | Some v ->
       let str = Build_info.V1.Version.to_string v in
       let version =
         if String.starts_with ~prefix:"v" str then
@@ -30,37 +38,12 @@ let version, xapi_version_major, xapi_version_minor, git_id =
         else
           str
       in
-      try
-        let maj, min, git_id =
-          Scanf.sscanf version "%d.%d.%s" (fun maj min rest -> (maj, min, rest))
-        in
-        (version, maj, min, git_id)
-      with _ ->
-        failwith
-          (Printf.sprintf
-             "Couldn't determine xapi version - got unexpected version from \
-              dune: '%s'"
-             version
-          )
-    )
+      let maj, min, git_id = parse_xapi_version version in
+      (version, maj, min, git_id)
 
 let compare_version version_a version_b =
-  let maj_a, min_a, _ =
-    try Scanf.sscanf version_a "%d.%d.%s" (fun maj min rest -> (maj, min, rest))
-    with _ ->
-      failwith
-        (Printf.sprintf "Couldn't determine xapi version for version_a: '%s'"
-           version_a
-        )
-  in
-  let maj_b, min_b, _ =
-    try Scanf.sscanf version_b "%d.%d.%s" (fun maj min rest -> (maj, min, rest))
-    with _ ->
-      failwith
-        (Printf.sprintf "Couldn't determine xapi version for version_b: '%s'"
-           version_b
-        )
-  in
+  let maj_a, min_a, _ = parse_xapi_version version_a in
+  let maj_b, min_b, _ = parse_xapi_version version_b in
   if maj_a > maj_b || (maj_a = maj_b && min_a > min_b) then
     1
   else if maj_a = maj_b && min_a = min_b then
--- a/ocaml/util/xapi_version.mli
+++ b/ocaml/util/xapi_version.mli
@@ -23,3 +23,5 @@ val git_id : string
 val xapi_version_major : int
 
 val xapi_version_minor : int
+
+val compare_version : string -> string -> int
--- a/ocaml/xapi-consts/api_messages.ml
+++ b/ocaml/xapi-consts/api_messages.ml
@@ -355,6 +355,3 @@ let tls_verification_emergency_disabled =
   addMessage "TLS_VERIFICATION_EMERGENCY_DISABLED" 3L
 
 let periodic_update_sync_failed = addMessage "PERIODIC_UPDATE_SYNC_FAILED" 3L
-
-let xapi_startup_blocked_as_version_higher_than_coordinator =
-  addMessage "XAPI_STARTUP_BLOCKED_AS_VERSION_HIGHER_THAN_COORDINATOR" 1L
--- a/ocaml/xapi-consts/api_messages.ml
+++ b/ocaml/xapi-consts/api_messages.ml
@@ -355,3 +355,6 @@ let tls_verification_emergency_disabled =
   addMessage "TLS_VERIFICATION_EMERGENCY_DISABLED" 3L
 
 let periodic_update_sync_failed = addMessage "PERIODIC_UPDATE_SYNC_FAILED" 3L
+
+let xapi_startup_blocked_as_version_higher_than_coordinator =
+  addMessage "XAPI_STARTUP_BLOCKED_AS_VERSION_HIGHER_THAN_COORDINATOR" 1L
--- a/ocaml/xapi-consts/api_messages.ml
+++ b/ocaml/xapi-consts/api_messages.ml
@@ -355,3 +355,6 @@ let tls_verification_emergency_disabled =
   addMessage "TLS_VERIFICATION_EMERGENCY_DISABLED" 3L
 
 let periodic_update_sync_failed = addMessage "PERIODIC_UPDATE_SYNC_FAILED" 3L
+
+let xapi_startup_blocked_as_version_higher_than_coordinator =
+  addMessage "XAPI_STARTUP_BLOCKED_AS_VERSION_HIGHER_THAN_COORDINATOR" 2L
--- a/ocaml/xapi/xapi.ml
+++ b/ocaml/xapi/xapi.ml
@@ -465,11 +465,9 @@ let attempt_host_status_check_with_coordinator ~__context my_ip =
                   ~self:(Helpers.get_localhost ~__context)
               in
               let err_msg =
-                "Xapi startup in pool member "
-                ^ name_label
-                ^ " is blocked as its xapi version("
-                ^ Xapi_version.version
-                ^ ") is higher than xapi version in pool coordinator."
+                Printf.sprintf "Xapi startup in pool member %s is blocked as its xapi version \
+                  (%s) is higher than xapi version in pool coordinator."
+                  name_label Xapi_version.version
               in
               if not !xapi_ver_high_alerted then (
                 let name, priority =
--- a/ocaml/xapi/xapi.ml
+++ b/ocaml/xapi/xapi.ml
@@ -423,6 +423,94 @@ type host_status_check_error =
 
 (* some glitch or other *)
 
+let xapi_ver_high_alerted = ref false
+
+(** Attempt checking host status with pool coordinator:
+ *  1. Pool.hello
+ *  2. if Pool.hello ok, check xapi version
+ *  Return None if ok or Some host_status_check_error otherwise *)
+let attempt_host_status_check_with_coordinator ~__context my_ip =
+  let localhost_uuid = Helpers.get_localhost_uuid () in
+  try
+    Helpers.call_emergency_mode_functions (Pool_role.get_master_address ())
+      (fun rpc session_id ->
+        match
+          Client.Client.Pool.hello ~rpc ~session_id ~host_uuid:localhost_uuid
+            ~host_address:my_ip
+        with
+        | `cannot_talk_back ->
+            error "Master claims he cannot talk back to us on IP: %s" my_ip ;
+            Xapi_host.set_emergency_mode_error
+              Api_errors.host_master_cannot_talk_back [my_ip] ;
+            Some Temporary
+        | `unknown_host ->
+            debug "Master claims he has no record of us being a slave" ;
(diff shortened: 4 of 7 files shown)


ocaml/util/xapi_version.ml
```diff
@@ -43,3 +43,27 @@ let version, xapi_version_major, xapi_version_minor, git_id =
              version
           )
     )
+
+let compare_version version_a version_b =
+  let maj_a, min_a, _ =
+    try Scanf.sscanf version_a "%d.%d.%s" (fun maj min rest -> (maj, min, rest))
```
minglumlu, line 49: It seems this is a code snippet repeating in multiple places.

ocaml/util/xapi_version.ml
```diff
@@ -43,3 +43,27 @@ let version, xapi_version_major, xapi_version_minor, git_id =
...
+  in
+  let maj_b, min_b, _ =
+    try Scanf.sscanf version_b "%d.%d.%s" (fun maj min rest -> (maj, min, rest))
+    with _ ->
+      failwith
+        (Printf.sprintf "Couldn't determine xapi version for version_b: '%s'"
+           version_b
+        )
+  in
+  if maj_a > maj_b || (maj_a = maj_b && min_a > min_b) then
```
minglumlu, line 64: If we don't care the exact result:
```
let is_higher a b =
  match (a.maj > b.maj, a.min > b.min) with
  | true, _
  | false, true -> true
  | _ -> false
```

ocaml/xapi/xapi.ml
```diff
@@ -1138,7 +1150,16 @@ let server_init () =
               (* Try to say hello to the pool *)
               match attempt_pool_hello ip with
               | None ->
-                  finished := true
+                  if compare_xapi_version_with_coordinator ~__context > 0 then (
+                    error
```
minglumlu, line 1154: I think an alert would be much nicer for users.

ocaml/util/xapi_version.ml
psafont: an alternative:
```ocaml
let compare a b =
  let to_maj_min v = try Scanf.sscanf v "%d.%d.%s" (fun maj min rest -> (maj, min, rest))
    with _ ->
      failwith
        (Printf.sprintf "Couldn't determine xapi version: '%s'"
           origin v
        )
  in
  let maj_a, min_a = to_maj_min a in
  let maj_b, min_b = to_maj_min b in
  let (<?>) a b = if a = 0 then b else a in
  compare maj_a maj_b <?> min_a min_b <?> 0
```
See Christian's style guide: https://github.com/lindig/ocaml-style#compare-functions

ocaml/util/xapi_version.mli
```diff
@@ -23,3 +23,5 @@ val git_id : string
 val xapi_version_major : int
 
 val xapi_version_minor : int
+
+val compare_version : string -> string -> int
```
psafont, line 27: It might be interesting to add a small datastructure for xapi versions and use it here, forcing users to convert the versions first:

```ocaml
type version = {major: int; minor: int; build: string}

val of_string : string -> version

val compare : version -> version -> int
```

ocaml/xapi/xapi.ml
gangj: To add an alert for this kind of pool member xapi startup failure, I think we also need to remove the alert later after the pool coordinator xapi got updated and restarted which un-blocks the pool member xapi startup.
Will add and remove alert in this way.

ocaml/util/xapi_version.ml
```diff
@@ -18,28 +18,31 @@ let hostname = "localhost"
 
 let date = Xapi_build_info.date
 
+let parse_xapi_version version =
+  try Scanf.sscanf version "%d.%d.%s" (fun maj min rest -> (maj, min, rest))
```
edwintorok, line 22: We have an `Rpm` module in XAPI which has a `Pkg.parse_epoch_version_release` and various comparison functions for it. In particular a version may contain more than just major and minor and the actual RPM version rules can be quite tricky.
E.g. in a backport we might have to increment (or introduce) a patch version in addition to major and minor in case the major.minor+1 is already taken up by an existing release.

ocaml/util/xapi_version.mli
gangj: Hi @psafont , I think adding a type for the "xapi version" is a good idea. But I feel it is not easy to change it clearly as:
1. now the name of the file is `xapi_version.ml`,
2. we already have "version" field in `xapi_version.ml`.
3. besides `version`, `xapi_version_major`, `xapi_version_minor`, `git_id`, there are also other field in the `xapi_version.ml`

To change it clearly, I would like to rename `xapi_version.ml` to `version.ml`, and add a type `xapi_version` as:
type xapi_version = {major: int; minor: int; build: string}
But this would need to change all the xapi codes using `Xapi_version`.

What's your though on this, or if current code is acceptable, can we keep it unchanged? Thank you!

ocaml/util/xapi_version.ml
gangj: The xapi version string is in format "major.minor.git_id", i.e.: "23.25.0", "0.0.dev" or "22.18.0.11.g93d35f038", I checked related `Rpm` code: I think it is different from `Rpm.epoch_ver_rel` as there is no "epoch" and "release" info inside "xapi_version".

ocaml/util/xapi_version.mli
psafont: Yes, that's too much churn for dubious benefit, thanks for looking into it!

ocaml/util/xapi_version.ml
psafont: I don't think Rpm can be used here. The versions compared here are runtime ones, not rpm ones

ocaml/xapi/xapi.ml
```diff
@@ -1138,7 +1153,57 @@ let server_init () =
...
               | None ->
-                  finished := true
+                  let obj_uuid = Helpers.get_localhost_uuid () in
+                  let message_name, _ =
+                    Api_messages
+                    .xapi_startup_blocked_as_version_higher_than_coordinator
+                  in
+                  let existing_alerts =
+                    Helpers.call_api_functions ~__context (fun rpc session_id ->
+                        Client.Client.Message.get_all_records ~rpc ~session_id
```
minglumlu, line 1163: This call may fail. The failure has to be handled properly.
Basically, this failure should not block the checking.

ocaml/xapi/xapi.ml
```diff
@@ -467,6 +467,18 @@ let attempt_pool_hello my_ip =
         [ExnHelper.string_of_exn exn] ;
       Some Temporary
 
+let compare_xapi_version_with_coordinator ~__context =
+  let coordinator_ref = Helpers.get_master ~__context in
+  let coordinator_xapi_ver =
+    Helpers.call_emergency_mode_functions (Pool_role.get_master_address ())
+      (fun rpc session_id ->
+        Client.Client.Host.get_software_version ~rpc ~session_id
```
minglumlu, line 475: What would happen when this call failed?
I think it should retry.

ocaml/xapi/xapi.ml
```diff
@@ -467,6 +467,18 @@ let attempt_pool_hello my_ip =
         [ExnHelper.string_of_exn exn] ;
       Some Temporary
 
+let compare_xapi_version_with_coordinator ~__context =
+  let coordinator_ref = Helpers.get_master ~__context in
+  let coordinator_xapi_ver =
+    Helpers.call_emergency_mode_functions (Pool_role.get_master_address ())
```
minglumlu, line 473: May I know the difference between `Helpers.call_emergency_mode_functions` and `Helpers.call_api_functions`?
This is called after one usage of `Helpers.call_api_functions` (to get the existing alerts), should it use `Helpers.call_api_functions` also?
Anyway, it's better to understand the difference between the two functions.

gangj: Checked the code, I think the difference between them mainly in if master's DB is used:
`call_api_functions`: uses the master's database
`call_emergency_mode_functions`: uses local storage
As here the host might have not synced with master DB, I think we should only use `call_emergency_mode_functions`.

ocaml/xapi/xapi.ml
```diff
@@ -444,7 +447,60 @@ let attempt_pool_hello my_ip =
               [localhost_uuid] ;
             Some Permanent
         | `ok ->
-            None
+            let obj_uuid = Helpers.get_localhost_uuid () in
```
robhoes, line 450: We've already got `localhost_uuid` above, which can be reused.

ocaml/xapi/xapi.ml
```diff
@@ -444,7 +447,60 @@ let attempt_pool_hello my_ip =
...
+            let message_name, _ =
+              Api_messages
+              .xapi_startup_blocked_as_version_higher_than_coordinator
+            in
+            let existing_alerts =
+              Client.Client.Message.get_all_records ~rpc ~session_id
```
robhoes, line 456: This is potentially a lot of data that gets transferred from the coordinator, every time xapi starts up.

ocaml/xapi/xapi.ml
```diff
@@ -444,7 +447,60 @@ let attempt_pool_hello my_ip =
...
+              |> List.filter (fun (_ref, record) ->
+                     record.API.message_obj_uuid = obj_uuid
+                     && record.API.message_name = message_name
+                 )
+            in
+            let compare_xapi_version_with_coordinator ~__context =
+              let coordinator_ref = Helpers.get_master ~__context in
+              let coordinator_xapi_ver =
+                Client.Client.Host.get_software_version ~rpc ~session_id
```
robhoes, line 465: At this point, we are able to use the `Db` module to more efficiently read from the DB. This is more efficient than doing a `Client` call, which makes a new connection.

robhoes: In fact, `Helpers.get_master` also uses the `Db` module internally, and you're doing it below as well.

ocaml/xapi/xapi.ml
```diff
@@ -444,7 +447,60 @@ let attempt_pool_hello my_ip =
...
+                  ~self:coordinator_ref
+                |> List.assoc "xapi_build"
+              in
+              Xapi_version.compare_version Xapi_version.version
+                coordinator_xapi_ver
+            in
+            if compare_xapi_version_with_coordinator ~__context > 0 then (
+              ( if existing_alerts = [] then
+                  let hostname =
+                    Db.Host.get_hostname ~__context
```
robhoes, line 475: I think you want `get_name_label`.

ocaml/xapi/xapi.ml
robhoes: To be honest, I would just send the alert if needed, and not worry about previous alerts. This is going to be an unusual situation anyway.

ocaml/xapi-consts/api_messages.ml
```diff
@@ -355,3 +355,6 @@ let tls_verification_emergency_disabled =
   addMessage "TLS_VERIFICATION_EMERGENCY_DISABLED" 3L
 
 let periodic_update_sync_failed = addMessage "PERIODIC_UPDATE_SYNC_FAILED" 3L
+
+let xapi_startup_blocked_as_version_higher_than_coordinator =
+  addMessage "XAPI_STARTUP_BLOCKED_AS_VERSION_HIGHER_THAN_COORDINATOR" 1L
```
robhoes, line 360: Priority 2 looks more appropriate, as there is not chance of data loss here.

robhoes (changes requested):
This looks good in general. Just a few small comments.

ocaml/xapi/xapi.ml
gangj: Ah, yes, sorry!

ocaml/xapi/xapi.ml
minglumlu: I looked at the context here again, how about doing `Xapi_host.set_emergency_mode_error` when `Xapi_globs.slave_emergency_mode` is `true`?
It seems in this circumstance, this is an existing way to expose the error.

gangj: The alert would be sent every 1 minutes, I think we had to avoid sending so many duplicated alerts when the xapi in the pool member is updated and restarted earlier than coordinator. While there seems not an easy way to achieve this.
Had a discussion with @minglumlu , I think Ming's suggestion is great, set emergency mode the same as other "hello error" seems a good choice.
But currently, XC has no way to check the host's emergency mode, as the only api: "local_assert_healthy" is not exported and _R_LOCAL_ROOT_ONLY.

ocaml/xapi-consts/api_messages.ml
gangj: I see, changed.

ocaml/xapi/xapi.ml
gangj: OK, changed.
Would you please share what's difference between the two fields of a host: `hostname` and `name_label`? It is strange that the comment for `name_label` field is:
`param_doc= "The name of the new storage repository"`
And from current code, I see both `hostname` and `name_label` are used in logging.

ocaml/xapi/xapi.ml
gangj: Changed to alert only once after xapi restart, and will not remove the alert(s) after it startup successfully.

ocaml/xapi/xapi.ml
```diff
@@ -1129,23 +1085,132 @@ let server_init () =
...
+                            let coordinator_xapi_ver =
+                              Db.Host.get_software_version ~__context
+                                ~self:coordinator_ref
+                              |> List.assoc "xapi_build"
+                            in
+                            Xapi_version.compare_version Xapi_version.version
+                              coordinator_xapi_ver
+                          in
+                          if
+                            compare_xapi_version_with_coordinator ~__context > 0
```
minglumlu, line 1137:
```suggestion
                          let is_higher =
                            Db.Host.get_software_version ~__context
                              ~self:(Helpers.get_master ~__context)
                            |> List.assoc "xapi_build"
                            |> Xapi_version.compare_version Xapi_version.version
                            |> fun r -> r > 0
                          in
                          if is_higher
                              

```

ocaml/xapi/xapi.ml
```diff
@@ -1129,23 +1085,132 @@ let server_init () =
...
             while not !finished do
               (* Grab the management IP address (wait forever for it if necessary) *)
               let ip = wait_for_management_ip_address ~__context in
               debug "Start master_connection watchdog" ;
               ignore (Master_connection.start_master_connection_watchdog ()) ;
               debug "Attempting to communicate with master" ;
-              (* Try to say hello to the pool *)
-              match attempt_pool_hello ip with
+
+              (* Attempt checking host status with pool coordinator:
```
minglumlu, line 1097: The code lines enclosed by the `while` loop look too many.
Would suggest to keep the original structure, with a function being separated.

ocaml/xapi/xapi.ml
```diff
@@ -417,14 +417,19 @@ let wait_for_management_ip_address ~__context =
...
 
-(** Attempt a Pool.hello, return None if ok or Some hello_error otherwise *)
-let attempt_pool_hello my_ip =
+let xapi_ver_high_alerted = ref false
+
+(** Attempt checking host status with pool coordinator:
+ *  1. Pool.hello
+ *  2. if Pool.hello ok, check xapi version
+ *  Return None if ok or Some host_status_check_error otherwise *)
+let attempt_host_status_check_with_coordinator ~__context my_ip =
```
minglumlu, line 432: I think the `pool_hello` is still a proper name.

ocaml/xapi/xapi.ml
```diff
@@ -444,7 +449,46 @@ let attempt_pool_hello my_ip =
...
+                       ("Xapi startup in pool member "
+                       ^ name_label
+                       ^ " is blocked as its xapi version("
+                       ^ Xapi_version.version
+                       ^ ") is higher than xapi version in pool coordinator."
+                       )
+                  ) ;
+                xapi_ver_high_alerted := true
+              ) ;
+              error
```
minglumlu, line 482: This error message could reuse the body message.

ocaml/xapi/xapi.ml
robhoes: `Host.name_label` is a read/write field for users to set to any name they like for their hosts. The `Host.hostname` is the network hostname.

ocaml/xapi/xapi.ml
gangj: OK, moved out.

ocaml/xapi/xapi.ml
gangj: I think this function now does more than the original pool.hello API call, it also compares xapi version.

minglumlu (approved):

ocaml/xapi/xapi.ml
```diff
@@ -456,12 +500,12 @@ let attempt_pool_hello my_ip =
         [localhost_uuid] ;
       Some Permanent
   | Api_errors.Server_error (code, params) as exn ->
-      debug "Caught exception: %s during Pool.hello"
+      debug "Caught exception: %s in attempt_host_status_check_with_coordinator"
         (ExnHelper.string_of_exn exn) ;
```
psafont, line 504:
```suggestion
      debug "Caught exception: %s in %s"
        (ExnHelper.string_of_exn exn) __FUNCTION__ ;
```

ocaml/xapi/xapi.ml
```diff
@@ -456,12 +500,12 @@ let attempt_pool_hello my_ip =
...
       Xapi_host.set_emergency_mode_error code params ;
       Some Temporary
   | exn ->
-      debug "Caught exception: %s during Pool.hello"
+      debug "Caught exception: %s in attempt_host_status_check_with_coordinator"
```
psafont, line 508:
```suggestion
      debug "Caught exception: %s in %s"
        (ExnHelper.string_of_exn exn) __FUNCTION__ ;
```

robhoes (approved):

robhoes:
Squashed and merged.

PR #5213, opened 2023-10-24
Replace duplicate functions rmrf and rmtree with rm_rec from Xapi_std…

snwoods:
…ext_unix.Unixext

### DIFF
--- a/ocaml/tests/test_observer.ml
+++ b/ocaml/tests/test_observer.ml
@@ -269,18 +269,9 @@ let is_dir_empty ~test_trace_log_dir =
     debug "%s" (Printexc.to_string e) ;
     true
 
-let rec rmdir_rec path =
-  match Sys.is_directory path with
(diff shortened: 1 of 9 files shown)


snwoods:
> This module is unknown - so needs a change in the `dune` file. This might be because in the past xenopsd and xapi were separate and did not share code. Not sure this can be resolved easily.

(For some reason github won't let me reply directly)

Sorry I meant to say this relies on this PR: https://github.com/xapi-project/stdext/pull/76
I think it's taking issue with rm_rec rather than the module itself, as Xapi_stdext_unix is used elsewhere in xenopsd without issue and it also builds with no problem locally where I have the new version of stdext.

lindig (approved):

lindig:
Can we move on with this @mg12 ?

psafont (approved):

PR #5215, opened 2023-10-25
XSI-1457: Limit number of sectors to coalesce

Vincent-lau:
To make the progress bar display more interactive, we limit the amount of consecutive sectors to coalesce when generating a list to copy.

### DIFF
--- a/ocaml/libs/vhd/vhd_format/f.ml
+++ b/ocaml/libs/vhd/vhd_format/f.ml
@@ -18,6 +18,14 @@ let sector_size = 512
 
 let sector_shift = 9
 
+let mib n =
+  let ( ** ) = Int64.mul in
+  Int64.(1024L ** 1024L ** of_int n)
+
+(** This is defined to be the same as sync_limit in channels.ml due to circular
+    dependencies.  *)
+let sync_limit = mib 4
+
 exception Cstruct_differ
 
 let cstruct_equal a b =
@@ -2633,7 +2641,11 @@ functor
       | false, None ->
           false
 
-    let rec coalesce_request acc s =
+    (* The coalesced_sectors variable accumulates the number of bytes that have
+       been coalesced so far. It is made optional because we only use it when we
+       continuously match with one pattern, i.e. the pattern where we coalesce
+       consecutive sectors, and default it to 1 when we are not coalescing. *)
+    let rec coalesce_request ?(coalesced_sectors = 1L) acc s =
       let open Int64 in
       s >>= fun next ->
       match (next, acc) with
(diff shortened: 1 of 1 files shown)


ocaml/libs/vhd/vhd_format/f.ml
```diff
@@ -2633,7 +2637,7 @@ functor
       | false, None ->
           false
 
-    let rec coalesce_request acc s =
+    let rec coalesce_request ?(coalesced_sectors = 1L) acc s =
```
lindig, line 2640: The new parameter could be explained in a comment. It is not obvious to me, especially because it it optional.

ocaml/libs/vhd/vhd_format/f.ml
```diff
@@ -2653,11 +2657,16 @@ functor
           return (Cons (x, fun () -> coalesce_request None s))
       | Cons (`Copy (h, ofs, len), next), None ->
           coalesce_request (Some (`Copy (h, ofs, len))) (next ())
-      | Cons (`Copy (h, ofs, len), next), Some (`Copy (h', ofs', len')) ->
+      | Cons (`Copy (h, ofs, len), next), Some (`Copy (h', ofs', len'))
+        when coalesced_sectors ** Int64.of_int sector_size <= sync_limit ->
```
lindig, line 2661: I think it's fine - but I stumbled when we have `when` followed by `if`: the if condition could be folded into the `when` and creating a new match for the `else`.

lindig (approved):

ocaml/libs/vhd/vhd_format/f.ml
```diff
@@ -18,6 +18,10 @@ let sector_size = 512
 
 let sector_shift = 9
 
+(** This is defined to be the same as sync_limit in channels.ml due to circular 
+    dependencies*)
+let sync_limit = Int64.(mul 4L (mul 1024L 1024L))
```
psafont, line 23: It's worth documenting this is in bytes

lindig: Or let code do the talking:
```
let mib n = 
  let ( ** ) = Int64.mul in
  Int64.(1024L ** 1024L ** of_int n)

let sync_limit = 4 mib
```

Vincent-lau: you mean `let sync_limit = mib 4` right?

lindig: Indeed

ocaml/libs/vhd/vhd_format/f.ml
Vincent-lau: done

psafont (approved):

robhoes (approved):

PR #5216, opened 2023-10-26
CA-365059: Clear messages in source pool after migrating VM

danilo-delbusso:
please be nit picky! 😁

tested using migration of w10 VM between 2 standalone hosts.
______

So while making this change I noticed we only move (and now clear) messages connected to the VM.

We are **not** doing the same for messages connected to other objects (such as VDIs). Is this something we should add?

### DIFF
--- a/ocaml/xapi/xapi_message.ml
+++ b/ocaml/xapi/xapi_message.ml
@@ -817,13 +817,15 @@ let handler (req : Http.Request.t) fd _ =
           (ExnHelper.string_of_exn e)
   )
 
-(* Export messages and send to another host/pool over http. *)
-let send_messages ~__context ~cls ~obj_uuid ~session_id ~remote_address =
-  let msgs = get ~__context ~cls ~obj_uuid ~since:(Date.of_float 0.0) in
-  let body = export_xml msgs in
+(* Export and send messages stored on the current host to another host/pool over http. *)
+let send_messages ~__context ~cls ~obj_uuid ~session_id ~remote_address
+    ~messages =
+  let body = export_xml messages in
   let query =
     [
-      ("session_id", Ref.string_of session_id); ("cls", "VM"); ("uuid", obj_uuid)
+      ("session_id", Ref.string_of session_id)
+    ; ("cls", Record_util.class_to_string cls)
+    ; ("uuid", obj_uuid)
     ]
   in
   let subtask_of = Context.string_of_task __context in
--- a/ocaml/xapi/xapi_vm_migrate.ml
+++ b/ocaml/xapi/xapi_vm_migrate.ml
@@ -1603,8 +1603,15 @@ let migrate_send' ~__context ~vm ~dest ~live:_ ~vdi_map ~vif_map ~vgpu_map
         XenAPI.VM.set_ha_always_run ~rpc:remote.rpc ~session_id:remote.session
           ~self:new_vm ~value:true ;
       (* Send non-database metadata *)
-      Xapi_message.send_messages ~__context ~cls:`VM ~obj_uuid:vm_uuid
+      let messages =
+        Xapi_message.get ~__context ~cls:`VM ~obj_uuid:vm_uuid
+          ~since:(Date.of_float 0.0)
+      in
+      Xapi_message.send_messages ~__context ~cls:`VM ~obj_uuid:vm_uuid ~messages
         ~session_id:remote.session ~remote_address:remote.remote_master_ip ;
+      (* CA-365059: Remove messages since they've already been copied to the destination pool *)
+      let message_refs = List.rev_map fst messages in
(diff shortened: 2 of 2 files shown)


ocaml/xapi/xapi_message.ml
```diff
@@ -817,10 +817,10 @@ let handler (req : Http.Request.t) fd _ =
...
-let send_messages ~__context ~cls ~obj_uuid ~session_id ~remote_address =
-  let msgs = get ~__context ~cls ~obj_uuid ~since:(Date.of_float 0.0) in
-  let body = export_xml msgs in
+(* Export and send host messages to another host/pool over http. *)
+let send_messages ~__context ~cls ~obj_uuid ~session_id ~remote_address
+    ~messages =
+  let body = export_xml messages in
   let query =
     [
       ("session_id", Ref.string_of session_id); ("cls", "VM"); ("uuid", obj_uuid)
```
psafont, line 826: There was a latent bug here, the class is ahrdcoded, it should be possible to convert it to a string
```suggestion
      ("session_id", Ref.string_of session_id); ("cls", to_string cls); ("uuid", obj_uuid)
```

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -1603,8 +1603,15 @@ let migrate_send' ~__context ~vm ~dest ~live:_ ~vdi_map ~vif_map ~vgpu_map
...
       (* Send non-database metadata *)
-      Xapi_message.send_messages ~__context ~cls:`VM ~obj_uuid:vm_uuid
+      let messages =
+        Xapi_message.get ~__context ~cls:`VM ~obj_uuid:vm_uuid
+          ~since:(Date.of_float 0.0)
+      in
+      Xapi_message.send_messages ~__context ~cls:`VM ~obj_uuid:vm_uuid ~messages
         ~session_id:remote.session ~remote_address:remote.remote_master_ip ;
+      (* CA-365059: Remove messages since they've already been copied to the destination pool *)
+      let message_refs = List.map fst messages in
```
edwintorok, line 1613: List.map may crash with a stack overflow if you have really lots of messages here (I don't think we have any auto-cleanup for these so e.g. if you've rebooted your VM tens of thousands of times over many years it might happen).

Better to use `List.rev_map` for now if you don't actually care about the order.

OCaml 5.1.0 has finally changed List.map to use the tail-recursion-modulo-const annotation on the function that makes it run in fixed amount of stack space (the annotation is supported since 4.14 but not widely applied within the stdlib yet). We also have various other explicitly tail recursive map implementations in stdext I think.

ocaml/xapi/xapi_message.ml
```diff
@@ -817,10 +817,10 @@ let handler (req : Http.Request.t) fd _ =
           (ExnHelper.string_of_exn e)
   )
 
-(* Export messages and send to another host/pool over http. *)
-let send_messages ~__context ~cls ~obj_uuid ~session_id ~remote_address =
-  let msgs = get ~__context ~cls ~obj_uuid ~since:(Date.of_float 0.0) in
-  let body = export_xml msgs in
+(* Export and send host messages to another host/pool over http. *)
```
edwintorok, line 820: minor nitpick:

are they host messages, or messages for an object (VM) that were stored on that host? Given that this is only invoked by VM migration and the 'cls' parameter is unused here I'd say 'export and send messages stored on the current host' (they are not necessarily messages about the host).

Also if 'cls' really is unused now then you can drop that parameter: the message class is implied by the list of messages you pass.

danilo-delbusso: good point, actually. I'll update the comment.

The `cls` is actually used, that was a mistake on my part

ocaml/xapi/xapi_vm_migrate.ml
robhoes: An alternative is to not create the new list at all, but use `List.iter` and `Xapi_message.destroy`. It's fine the way you've done it now though.

robhoes (approved):
An alternative would have been to modify the `send_messages` function to `send_and_destroy_messages`. That would declutter the already-long `migrate_send'` function, which is the only caller.

psafont (approved):

danilo-delbusso:
> An alternative would have been to modify the `send_messages` function to `send_and_destroy_messages`. That would declutter the already-long `migrate_send'` function, which is the only caller.

I had thought about it but I preferred sticking to one function = one action. Would something like `send_and_destroy_messages` be common in xapi?

robhoes:
>I had thought about it but I preferred sticking to one function = one action. Would something like send_and_destroy_messages be common in xapi?

Take a look at that `migrate_send'` function for a (terrible) counterexample... 😂  I'm in favour of keeping functions short and simple in general. In this case, the combined action would still be a simple function (and you might call it a single "move" action).

PR #5217, opened 2023-10-27
CP-45938: Fixup xs9 failure due to python2 stuff

liulinC:
XS9 does not have python2, the building and installation of python2 stuff should be excluded from XS9

### DIFF
--- a/scripts/Makefile
+++ b/scripts/Makefile
@@ -168,9 +168,11 @@ install:
 	$(IPROG) host-backup-restore/host-backup $(DESTDIR)$(LIBEXECDIR)
 	$(IPROG) host-backup-restore/host-restore $(DESTDIR)$(LIBEXECDIR)
 # example/python
+ifneq ($(BUILD_PY2), NO)
 	$(IDATA) examples/python/XenAPIPlugin.py $(DESTDIR)$(SITE_DIR)/
 	$(IDATA) examples/python/XenAPI/XenAPI.py $(DESTDIR)$(SITE_DIR)/
 	$(IDATA) examples/python/inventory.py $(DESTDIR)$(SITE_DIR)/
+endif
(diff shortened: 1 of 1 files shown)


psafont (approved):

edwintorok (approved):

scripts/Makefile
```diff
@@ -168,9 +168,11 @@ install:
 	$(IPROG) host-backup-restore/host-backup $(DESTDIR)$(LIBEXECDIR)
 	$(IPROG) host-backup-restore/host-restore $(DESTDIR)$(LIBEXECDIR)
 # example/python
+ifneq ($(BUILD_PY2), NO)
 	$(IDATA) examples/python/XenAPIPlugin.py $(DESTDIR)$(SITE_DIR)/
 	$(IDATA) examples/python/XenAPI/XenAPI.py $(DESTDIR)$(SITE_DIR)/
```
MarkSymsCtx, line 173: This is the XenAPI file installed into Dom0, as far as I know it has been made p2/3 compatible and is used by Dom0 Pythn3 systems such as the Storage Manager, removing it from the install is likely to break a lot of things.

psafont: Indeed, this is why BUILD_PY2=NO will only be set for future XS9 builds

edwintorok: There is a 2nd line below that installs it to SITE3_DIR, which is what the XS9 build uses. The SITE_DIR install rules were causing a build failure on XS9, because SITE_DIR ended up empty, and there were unpackaged /XenAPI.py files.

edwintorok: We probably should've called SITE_DIR, SITE2_DIR to make it clear it is python2-only.

liulinC: @edwintorok  is right, `DESTDIR` is for PY2 only.
What this commit do is disable PY2 on purpose (for xs9, for example). PY3 is not affacted.

PR #5219, opened 2023-10-30
CP-44561: Preserve the default attributes when setting attributes on an

snwoods:
Observer. Remove the user. prefix from db attributes

### DIFF
--- a/ocaml/xapi/xapi_observer.ml
+++ b/ocaml/xapi/xapi_observer.ml
@@ -194,20 +194,23 @@ let assert_valid_attributes attributes =
     )
     attributes
 
-let register_components ~__context ~self ~host =
+let default_attributes ~__context ~host ~name_label =
   let pool = Helpers.get_pool ~__context in
   let host_label = Db.Host.get_name_label ~__context ~self:host in
   let host_uuid = Db.Host.get_uuid ~__context ~self:host in
   let pool_uuid = Db.Pool.get_uuid ~__context ~self:pool in
+  [
+    ("xs.pool.uuid", pool_uuid)
+  ; ("xs.host.name", host_label)
+  ; ("xs.host.uuid", host_uuid)
+  ; ("xs.observer.name", name_label)
+  ]
+
+let register_components ~__context ~self ~host =
   let name_label = Db.Observer.get_name_label ~__context ~self in
   let attributes =
-    ("xs.pool.uuid", pool_uuid)
-    :: ("xs.host.name", host_label)
-    :: ("xs.host.uuid", host_uuid)
-    :: ("xs.observer.name", name_label)
-    :: List.map
-         (fun (k, v) -> ("user." ^ k, v))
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_observer.ml
```diff
@@ -194,20 +194,21 @@ let assert_valid_attributes attributes =
     )
     attributes
 
-let register_components ~__context ~self ~host =
+let default_attributes ~__context ~self ~host ~name_label =
   let pool = Helpers.get_pool ~__context in
   let host_label = Db.Host.get_name_label ~__context ~self:host in
   let host_uuid = Db.Host.get_uuid ~__context ~self:host in
   let pool_uuid = Db.Pool.get_uuid ~__context ~self:pool in
+  ("xs.pool.uuid", pool_uuid)
```
lindig, line 202: This could be written more conventionally as a list of key-value pairs.

lindig (approved):

ocaml/xapi/xapi_observer.ml
psafont: Not sure what you mean, this is already a list of key-value pairs (or association list), do you mean using `[(k1, v1); ...]`?

lindig: This is written as `a :: b :: [c]` but why not as `[a; b; c]`?

psafont: It's preserving the original code, I guess the other style of writing lists was more in vogue back then

psafont (approved):

psafont:
needs a small tweak to `default_attributes` to remove the `~self` parameter, otherwise seems fine

ocaml/xapi/xapi_observer.ml
snwoods: Yeah I just blindly copied the old code, but I agree [a; b; c] is nicer :)

robhoes: >I guess the other style of writing lists was more in vogue back then

You mean, six months ago? ;)

lindig: You have to write it using `::` if you prepend to an existing list; but here we know the entire list so I see no good reason to use `a :: b :: [c]`. Anyway, this is a minor point since it will be identical after parsing.

ocaml/xapi/xapi_observer.ml
```diff
@@ -368,11 +371,17 @@ let set_enabled ~__context ~self ~value =
 let set_attributes ~__context ~self ~value =
   assert_valid_attributes value ;
   let uuid = Db.Observer.get_uuid ~__context ~self in
+  let host = Helpers.get_localhost ~__context in
+  let name_label = Db.Observer.get_name_label ~__context ~self in
+  let default_attributes =
+    default_attributes ~__context ~self ~host ~name_label
```
robhoes, line 377: The function doesn't have `~self`.

lindig (approved):

PR #5221, opened 2023-10-31
CP-44477 Add Typescript bindings for XenAPI

acefei:
In this PR, I implemented the Typescript bindings for XenAPI whose usage is almost identical to the [Python XenAPI](https://xapi-project.github.io/xen-api/usage.html), except it's asynchronous and requires async/await.

Meanwhile, publish Typescript distribution for XenAPI automatically to npmjs by github action. Repo Admin need to create a `NPM_TOKEN` variable in repository secrets.

### DIFF
--- a/scripts/examples/typescript/README.md
+++ b/scripts/examples/typescript/README.md
@@ -1,0 +1,33 @@
+## Typescript for XenAPI Usage
+
+### Examples
+Install lib from npm repository
+```
+$ npm install xen-api-ts
+```
+
+The usage of Typescript XenAPI is almost identical to the [Python XenAPI](https://xapi-project.github.io/xen-api/usage.html), except it's asynchronous and requires async/await.
+```
+import { xapi_client } from "xen-api-ts"
+async function main() {
+    const session = xapi_client(process.env.HOST_URL)
+    try:
+        await session.login_with_password(process.env.USERNAME, process.env.PASSWORD)
+        const hosts = await session.xenapi.host.get_all()
+    finally:
+        await session.xenapi.session.logout()
+}
+main()
+```
+
+For more example usage, we can run as follows.
+```
+$ cd <examples/typescript dir>
+$ npm install
+$ echo "HOST_URL=xxx" >> .env
+$ echo "USERNAME=xxx" >> .env
+$ echo "PASSWORD=xxx" >> .env
+$ npm test tests/getXapiVersion.ts
+```
+
+## Packaging
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ JOBS = $(shell getconf _NPROCESSORS_ONLN)
 PROFILE=release
 OPTMANDIR ?= $(OPTDIR)/man/man1/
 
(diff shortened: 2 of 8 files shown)


psafont (changes requested):
While I understand the location was copied from the python SDK, I don't like this location. I would prefer to use `typescript/sdk`

I've checked the official github docs and using the auth token is the supported way of doing it, unlike PyPI which doesn't need it.

I'd like to know more about the publising of the SDK is this something that should be handled by linux foundation / xapi-project?

scripts/examples/typescript/README.md
```diff
@@ -0,0 +1,33 @@
+## Typescript for XenAPI Usage
+
+### Examples
+Install lib from npm repository
+```
+$ npm install xen-api-ts
```
psafont, line 6: Is it standard to add -ts to the library name? The end result should be a javascript library along the typing metadata so it doesn't feel right

acefei: No, there is no standard mentioned this library naming convertion. 
According to the current tsconfig.json and package.json, we can build and publish the TypeScript based NPM package that interops seamlessly with TypeScript and JavaScript, so the -ts is not mandatory.

BTW, there is a official guide to publish npm packages under the [@types](https://www.npmjs.com/~types) organization https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#publish-to-types

acefei:
> While I understand the location was copied from the python SDK, I don't like this location. I would prefer to use `typescript/sdk`
> 
This Typescript XenAPI implatement is inspired by Python XenAPI, which intercept and redefine getter/setter/call function for that object to generate API bindings automatically, it was not like sdk-gen code, so I am not quite sure if calling it as Typescript SDK is appropriate.
I can rename the location as you prefer if there are no objections.

> I've checked the official github docs and using the auth token is the supported way of doing it, unlike PyPI which doesn't need it.
> 
Yeah, and if using registry-url as "https://npm.pkg.github.com" instead of "https://registry.npmjs.org", we can use NODE_AUTH_TOKEN: ${{ github.token }} to publish package, in this way, we won't extra effort to create auth token from npmjs.

> I'd like to know more about the publising of the SDK is this something that should be handled by linux foundation / xapi-project?
In fact, I am not sure. I'm just following the current practice of Python XenAPI, which publishes the package on PyPI.

scripts/examples/typescript/README.md
acefei: the other thing is the name `xen-api` was occupied on npmjs, so I have to add -ts suffix for difference.

Zhengchai:
@psafont @minglumlu @acefei This PR has been a while, can you please suggest the next action item to move this on please? Thanks.

kc284:
@Zhengchai There was another PR on this https://github.com/xapi-project/xen-api/pull/5623
It had been reviewed by me and Danilo, but he requested changes has not been completed. Which of the two PRs is the right one to proceed with?

kc284:
I think #5623 is the one to go with since it had significant reworking.

PR #5222, opened 2023-11-01
CP-45571: Add VM.restart_device_model function (API/CLI)

gangj:
Implemented as a localhost migration, can be made more efficient in future.
As it is simply a "pool_migrate", just reuse message forwarding of "pool_migrate" to handle "allowed operation" and "message routing".

### DIFF
--- a/ocaml/idl/datamodel_vm.ml
+++ b/ocaml/idl/datamodel_vm.ml
@@ -1033,6 +1033,7 @@ let pool_migrate =
       ; Api_errors.other_operation_in_progress
       ; Api_errors.vm_is_template
       ; Api_errors.operation_not_allowed
+      ; Api_errors.vm_bad_power_state
       ]
     ~allowed_roles:(_R_VM_POWER_ADMIN ++ _R_CLIENT_CERT)
     ()
@@ -1694,6 +1695,7 @@ let restart_device_models =
       ; Api_errors.other_operation_in_progress
       ; Api_errors.vm_is_template
       ; Api_errors.operation_not_allowed
+      ; Api_errors.vm_bad_power_state
       ]
     ~allowed_roles:(_R_VM_POWER_ADMIN ++ _R_CLIENT_CERT)
     ()
--- a/ocaml/idl/datamodel_vm.ml
+++ b/ocaml/idl/datamodel_vm.ml
@@ -1033,7 +1033,6 @@ let pool_migrate =
       ; Api_errors.other_operation_in_progress
       ; Api_errors.vm_is_template
       ; Api_errors.operation_not_allowed
-      ; Api_errors.vm_migrate_failed
       ]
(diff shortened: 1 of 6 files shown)


ocaml/xapi/xapi_vm.ml
```diff
@@ -1611,3 +1611,13 @@ let set_NVRAM_EFI_variables ~__context ~self ~value =
...
       Db.VM.set_NVRAM ~__context ~self ~value
   )
+
+let restart_device_models ~__context ~self =
+  let host = Db.VM.get_resident_on ~__context ~self in
+  (* As it is implemented as a localhost migration, just reuse message
+   * forwarding of "pool_migrate" to handle "allowed operation" and "message
+   * routing" *)
+  Helpers.call_api_functions ~__context (fun rpc session_id ->
+      Client.VM.pool_migrate ~rpc ~session_id ~vm:self ~host
```
robhoes, line 1621: Which error will be returned if the VM is not running? It should be a bad power state.

ocaml/idl/datamodel_vm.ml
```diff
@@ -1033,7 +1033,6 @@ let pool_migrate =
       ; Api_errors.other_operation_in_progress
       ; Api_errors.vm_is_template
       ; Api_errors.operation_not_allowed
-      ; Api_errors.vm_migrate_failed
```
robhoes, line 1036: Why is this removed? Are you sure this error is not raised by this call?

ocaml/xapi/xapi_vm.ml
gangj: The handler of "pool_migrate" had handled kinds of errors including the "bad power state", here is the tested result:

```
[root@lcy2-dt44 ~]# xe vm-restart-device-models vm=rhel8_202
You attempted an operation on a VM that was not in an appropriate power state at the time; for example, you attempted to start a VM that was already running. The parameters returned are the VM's handle, and the expected and actual VM state at the time of the call.
vm: 698f91f3-84f8-34e3-714e-f4cd7d8a1207 (rhel8_202)
expected: running
actual: halted
```

ocaml/idl/datamodel_vm.ml
gangj: I copied the errors from `vm.pool_migrate` for `vm.restart_device_models`, yes, I checked the code, it is not raised from the `vm.pool_migrate` API call.

ocaml/xapi/xapi_vm.ml
robhoes: That's good. I was worried that it would first complain about the `host` param being `OpaqueRef:NULL` (from `resident_on`).

robhoes (approved):

psafont (approved):

robhoes:
Squashed+merged.

PR #5223, opened 2023-11-01
CP-45741 VCS support, adjust args for qemu, demu

lindig:
Part of CP-44317, CP-44320 suporting NVidia VCS. Adjust args when we are running a VCS (compute) vGPU when starting qemu and demu:

* don't pass -vgpu to qemu
* pass --noconsole to demu

Remove a redundant match in cmdline_of_disp.

### DIFF
--- a/ocaml/xenopsd/xc/device.ml
+++ b/ocaml/xenopsd/xc/device.ml
@@ -2091,10 +2091,11 @@ module Dm_Common = struct
   let cmdline_of_disp ?domid info =
     let vga_type_opts x =
       let open Xenops_interface.Vgpu in
+      (* We can match on the implementation details to detect the VCS
+         case. Don't pass -vgpu for a compute vGPU. *)
       match x with
-      | Vgpu [{implementation= Nvidia _; _}] ->
-          ["-vgpu"]
-      | Vgpu ({implementation= Nvidia _; _} :: _) ->
+      | Vgpu ({implementation= Nvidia {vclass; _}; _} :: _)
+        when vclass <> Some "Compute" ->
           ["-vgpu"]
       | Vgpu [{implementation= GVT_g gvt_g; _}] ->
           let base_opts =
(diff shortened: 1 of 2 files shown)


ocaml/xenopsd/xc/device.ml
```diff
@@ -2091,10 +2091,11 @@ module Dm_Common = struct
...
     let vga_type_opts x =
       let open Xenops_interface.Vgpu in
+      (* We can match on the implementation details to detect the VCS
+         case. Don't pass -vgpu for a compute vGPU. *)
       match x with
-      | Vgpu [{implementation= Nvidia _; _}] ->
-          ["-vgpu"]
-      | Vgpu ({implementation= Nvidia _; _} :: _) ->
+      | Vgpu ({implementation= Nvidia {vclass; _}; _} :: _)
+        when vclass <> Some "Compute" ->
```
psafont, line 2098: This change will make compute vgpus to raise an exception. It needs another case added for them to return an empty list, probably along the MxGPU case

psafont (changes requested):

ocaml/xenopsd/xc/device.ml
lindig: What exception? No match?

lindig: Will rewrite this

psafont (approved):

ocaml/xenopsd/xc/service.ml
```diff
@@ -504,7 +514,15 @@ module Vgpu = struct
       @ device_args
     in
     let fd_arg = if restore then ["--resume"] else [] in
-    List.concat [base_args; fd_arg]
+    (* support for NVidia VCS (compute) vGPUs *)
+    let no_console =
+      match List.exists is_compute_vgpu vgpus with
```
robhoes, line 519: Hmm... I'd expect to not want a console if _all_ vGPUs are compute ones, not just one of them. Is that right?

lindig:
This should be safe to merge because we don't have compute profiles yet. I would squash the fixup before merging.

ocaml/xenopsd/xc/service.ml
lindig: I have to ask @xennifer. Could switch to `List.for_all`.

lindig: I've changed it to `for_all` after talking to @xennifer

robhoes (approved):

PR #5224, opened 2023-11-01
CP-45304: Remove UUID from span name of dispatch:system.isAlive:<UUID…

snwoods:
…> to reduce spam entries

### DIFF
--- a/ocaml/xapi/context.ml
+++ b/ocaml/xapi/context.ml
@@ -218,7 +218,7 @@ let make_dbg http_other_config task_name task_id =
       (if task_name = "" then "" else " ")
       (Ref.really_pretty_and_small task_id)
 
-let tracing_of_origin (origin : origin) task_name =
+let tracing_of_origin (origin : origin) span_name =
   let open Tracing in
   let ( let* ) = Option.bind in
   let parent =
@@ -226,7 +226,7 @@ let tracing_of_origin (origin : origin) task_name =
     | Http (req, _) ->
         let* traceparent = req.Http.Request.traceparent in
         let* span_context = SpanContext.of_traceparent traceparent in
-        let span = Tracer.span_of_span_context span_context task_name in
+        let span = Tracer.span_of_span_context span_context span_name in
         Some span
     | _ ->
         None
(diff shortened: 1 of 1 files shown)


ocaml/xapi/context.ml
```diff
@@ -309,11 +309,23 @@ let make ?(http_other_config = []) ?(quiet = false) ?subtask_of ?session_id
             " by task " ^ make_dbg [] "" subtask_of
         )
   ) ;
+  let span_details_from_task_name task_name =
+    match String.split_on_char ':' task_name with
+    | [x; y; uuid] when x = "dispatch" && y = "system.isAlive" ->
```
lindig, line 314: The string literals can be matched directly, no need to use `when` for them

lindig:
I suggest to trim the subject line of the commit such that it no longer overflows.

lindig (approved):

ocaml/xapi/context.ml
snwoods: As in ["dispatch"; "system.isAlive"; uuid] ? Then we'd need to repeat the string literals, wouldn't it be better to use variables for reuse?

lindig: Yes, I was think about `| ["dispatch"; "system.isAlive"; uuid] ->`. When using `when` you can match against a let-bound string (and use an arbitrary expression) but your code doesn't do that currently and hence my suggestion. If you want to avoid repeated string literals you have to use `when`.

ocaml/xapi/context.ml
```diff
@@ -309,11 +309,23 @@ let make ?(http_other_config = []) ?(quiet = false) ?subtask_of ?session_id
...
+        let span_name = x ^ ":" ^ y in
```
psafont, line 315: This can avoid the repetition, it's not very readable, however:
```suggestion
    | ["dispatch"; "system.isAlive"; uuid] ->
        let span_name = String.sub task_name 0 23 in
```

mg12 (approved):

ocaml/xapi/context.ml
snwoods: Yeah it is a little unreadable, I think I'll stick with the `when`

PR #5225, opened 2023-11-02
Update datamodel lifecycle

robhoes:


psafont:
I'm thinking of adding a step in the release workflow to push the change after the tag on master

lindig (approved):

PR #5226, opened 2023-11-02
NUMA-aware VM placement: introduce host level control and turn best effort placement on by default

edwintorok:
Draft PR, will need to rerun the perf tests from last time. Also I've only compile tested this.

And I think I have an explanation for the regression that was seen last time with this turned on: it is only a regression if we have a single VM on the host, as soon as you have more and they're all busy it is expected that this change would be the same or slightly better. Yet to confirm that theory, if it turns out to be false I'll remove the last commit that changes the default.

I can also introduce a pool level setting, but if this works well it may not be needed (i.e. just as an escape hatch if something goes wrong).

### DIFF
--- a/doc/content/toolstack/features/NUMA/index.md
+++ b/doc/content/toolstack/features/NUMA/index.md
@@ -122,6 +122,8 @@ See page 13 in [^AMD_numa] for a diagram of an AMD Opteron 6272 system.
    If this is an issue in practice the NUMA matrix could be pre-filtered to contain only reachable nodes.
    NUMA nodes with 0 CPUs *are* accepted (it can result from hard affinity pinnings)
 * NUMA balancing is not considered during HA planning
+* Dom0 is a single VM that needs to communicate with all other VMs, so NUMA balancing is not applied to it (we'd need to expose NUMA topology to the Dom0 kernel so it can better allocate processes)
+* IO NUMA is out of scope for now
 
 ## XAPI datamodel design
 
@@ -153,15 +155,16 @@ Debugging the allocator is done by running `xl vcpu-list` and investigating the 
 
 See the documentation in [softaffinity.mli] and [topology.mli].
 
-[Softaffinity.plan] returns a [CPUSet] given a host's NUMA allocation state and a VM's NUMA allocation request.
-[Topology.CPUSet] provides helpers for operating on a set of CPU indexes.
-[Topology.NUMAResource] is a [CPUSet] and the free memory available on a NUMA node.
-[Topology.NUMARequest] is a request for a given number of vCPUs and memory in bytes.
-[Topology.NUMA] represents a host's NUMA allocation state.
-[Topology.NUMA.candidates] are groups of nodes orderd by minimum average distance. The sequence is limited to [N+65520], where [N] is the number of NUMA nodes.
+* [Softaffinity.plan] returns a [CPUSet] given a host's NUMA allocation state and a VM's NUMA allocation request.
+* [Topology.CPUSet] provides helpers for operating on a set of CPU indexes.
+* [Topology.NUMAResource] is a [CPUSet] and the free memory available on a NUMA node.
+* [Topology.NUMARequest] is a request for a given number of vCPUs and memory in bytes.
+* [Topology.NUMA] represents a host's NUMA allocation state.
+* [Topology.NUMA.candidates] are groups of nodes orderd by minimum average distance.
+The sequence is limited to [N+65520], where [N] is the number of NUMA nodes.
 This avoids exponential state space explosion on very large systems (>16 NUMA nodes).
-[Topology.NUMA.choose] will choose one NUMA node deterministically, while trying to keep overall NUMA node usage balanced.
-[Domain.numa_placement] builds a [NUMARequest] and uses the above [Topology] and [Softaffinity] functions to compute and apply a plan.
+* [Topology.NUMA.choose] will choose one NUMA node deterministically, while trying to keep overall NUMA node usage balanced.
+* [Domain.numa_placement] builds a [NUMARequest] and uses the above [Topology] and [Softaffinity] functions to compute and apply a plan.
 
 We used to have a `xenopsd.conf` configuration option to enable numa placement, for backwards compatibility this is still supported, but only if the admin hasn't set an explicit policy on the Host.
 It is best to remove the experimental `xenopsd.conf` entry though, a future version may completely drop it.
--- a/doc/content/toolstack/features/NUMA/index.md
+++ b/doc/content/toolstack/features/NUMA/index.md
@@ -1,0 +1,186 @@
++++
+title = "NUMA"
++++
+
+## NUMA in a nutshell
+
+Systems that contain more than one CPU socket are typically built on a Non-Uniform Memory Architecture (NUMA) [^xen_numa][^kernel_numa].
+In a NUMA system each node has fast, lower latency access to local memory.
+
+![hwloc](hwloc.svg)
+
+In the diagram [^lstopo] above we have 4 NUMA nodes:
+ * 2 of those are due to 2 separate physical packages (sockets)
+ * a further 2 is due to Sub-NUMA-Clustering (aka Nodes Per Socket for AMD) where the L3 cache is split
+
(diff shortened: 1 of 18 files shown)


edwintorok:
TODO:
* We have some docs in the source code about NUMA design&implementation. Now that we have Hugo add something about NUMA there and link it in some way to the source code docs
* I'll probably split this PR into the datamodel changes, and one additional PR that just flips the default policy so we can test/merge these independently as they are ready
* May want to have a per-VM NUMA affinity policy in the future that allows more strict control over placement (although to see any benefits the host must already use best-effort as default for other VMs)
  * Those stricter controls will require some new `xenctrl` APIs to query where the VM's memory actually ended up, and some integration with XAPI's VM planner (currently based on host free memory)
  * out of scope for this PR, but will attempt to document the changes needed in XAPI&Xen to achieve that

edwintorok:
PR split, 2nd one is here: https://github.com/edwintorok/xen-api/pull/3.
Also moved and updated the design doc about NUMA, I think this is ready for review now.

ocaml/idl/datamodel_common.ml
```diff
@@ -10,7 +10,7 @@ open Datamodel_roles
               to leave a gap for potential hotfixes needing to increment the schema version.*)
 let schema_major_vsn = 5
 
-let schema_minor_vsn = 767
+let schema_minor_vsn = 768
```
psafont, line 13: I believe this needs to be bumped to 769 because it was bumped in another PR

psafont (approved):

ocaml/xapi-cli-server/record_util.ml
```diff
@@ -706,6 +706,24 @@ let host_sched_gran_to_string = function
...
+
+let host_numa_affinity_policy_of_string = function
+  | "any" ->
+      `any
+  | "besteffort" ->
+      `besteffort
+  | "default" ->
+      `default
+  | _ ->
+      failwith "Bad type"
```
mg12, line 725: should we use the other examples around and raise a Record_failure instead with the expected valid values?

ocaml/xapi-idl/xen/xenops_interface.ml
```diff
@@ -538,6 +538,12 @@ module Host = struct
   }
   [@@deriving rpcty]
 
+  type numa_affinity_policy =
```
mg12, line 541: shouldn't there be a Default option here as well?

ocaml/xapi-cli-server/record_util.ml
edwintorok: I did, but followed the wrong example that used failwith. Updated (btw this file should really be  mostly autogenerated).

ocaml/idl/datamodel_host.ml
```diff
@@ -1668,6 +1668,35 @@ let cleanup_pool_secret =
...
+    ~doc:"Set VM placement NUMA affinity policy"
+    ~params:
+      [
+        (Ref _host, "self", "The host")
+      ; ( host_numa_affinity_policy
+        , "value"
+        , "The NUMA affinity policy to apply to a host"
+        )
+      ]
+    ~allowed_roles:_R_LOCAL_ROOT_ONLY ()
```
robhoes, line 1698: Really local root? Not pool admin?

ocaml/xapi/dbsync_slave.ml
```diff
@@ -89,6 +89,9 @@ let refresh_localhost_info ~__context info =
   debug "Updating host software_version and updates_requiring_reboot" ;
   Create_misc.create_updates_requiring_reboot_info ~__context ~host ;
   Create_misc.create_software_version ~__context ~info:(Some info) () ;
+  (* set startup policy in xenopsd *)
+  Xapi_host.set_numa_affinity_policy ~__context ~self:host
```
robhoes, line 93: The function `refresh_localhost_info` is meant to "Update the information in the Host structure". This new call is rather the opposite and is not really a good fit here.

robhoes: Also, `Xapi_host.set_numa_affinity_policy` does another DB-set inside of it, which is a bit of a waste. It would be nicer to split that function in a DB-set part and a call-xenopsd part.

ocaml/idl/datamodel_host.ml
```diff
@@ -1668,6 +1668,35 @@ let cleanup_pool_secret =
       ]
     ~allowed_roles:_R_LOCAL_ROOT_ONLY ~hide_from_docs:true ()
 
+let host_numa_affinity_policy =
+  Enum
+    ( "host_numa_affinity_policy"
+    , [
+        ("any", "VMs are spread across all available NUMA nodes")
+      ; ( "best-effort"
```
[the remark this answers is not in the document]
edwintorok: We don't have enough information: we attempt to set things up such that Xen choses NUMA placement the way we hope it would, but we cannot query what it actually did. Also if you try to start multiple VMs then due to race conditions one NUMA node may run out of memory and you'll get memory allocated from the other one anyway.

Also best effort because if we should never refuse to start a VM if it doesn't fit into its original NUMA nodes (we expand the set of NUMA nodes until it includes the entire host if needed)

ocaml/idl/datamodel_host.ml
edwintorok: Ah initialy I wanted this to be a pool level setting, with the host level setting internal. But since there is no pool level setting I can change this back.

robhoes (changes requested):
As commented above, this needs a small update.

ocaml/xapi/dbsync_slave.ml
edwintorok: I moved this to Xapi_xenopsd and on_xapi_startup now.

ocaml/idl/datamodel_host.ml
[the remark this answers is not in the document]
edwintorok: Updated the comment to say this is not guaranteed and may fall back to 'any' (we don't have any policies with guarantees atm)

edwintorok:
I've rebased on latest master to fix the conflict with the new schema hash (required bumping the minor vsn again too).
Also I had to temporarily undo the changes to datamodel_lifecycle, and then run `make build` again, otherwise it wanted to commit the `23.30.0-next` as `23.32.0` instead of `23.32.0-next`.

So I've dropped all the datamodel lifecycle changes from the PR, they are best left uncommitted until we tag them otherwise a long running PR may end up getting committed with the wrong API version. Should be ready to review again.

lindig (approved):

doc/content/toolstack/features/NUMA/index.md
```diff
@@ -0,0 +1,220 @@
++++
+title = "NUMA"
++++
+
+## NUMA in a nutshell
```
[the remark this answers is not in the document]
edwintorok: I'd probably prefer the one without line length limits: adding or removing a word would result in smaller diffs (the linebreaks in the markdown won't carry over to the final document anyway unless it is a paragraph break).

Some of this got copied as-is from OCaml code which does have line length limits and would explain the weird formatting. I'll make it more consistent.

ocaml/xapi-cli-server/record_util.ml
```diff
@@ -706,6 +706,25 @@ let host_sched_gran_to_string = function
...
+let host_numa_affinity_policy_of_string = function
+  | "any" ->
+      `any
+  | "best-effort" ->
+      `besteffort
+  | "default" ->
+      `default
+  | s ->
+      raise
+        (Record_failure ("Expected 'any', 'best-effort' or 'default', got " ^ s))
```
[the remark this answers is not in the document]
edwintorok: That is a problem everywhere in this file, and would more easily be fixed if the file was auto-generated, I'll take a look whether that is easy to do.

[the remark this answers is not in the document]
edwintorok: We're not very consistent here, sometimes we use `_` as separator for the CLI, sometimes `-`, and the actual polymorphic variant sometimes includes neither (e.g. `balanceslb is `balance-slb` on the CLI). We could autogenerate code to accept both _ and - as separator, but we need to be backwards compatible with the way we printed things before (and changing the enum name to add _ would break some SDKs that got generated without it).

Still we can do better than what we do now by having a list of polymorphic variant and string names and calling a helper function to perform the matching and print the error message.

robhoes (approved):
Another reformat and squash needed, but then it looks good to me.

Issue #4072, opened 2020-04-07
XAPI's runtime dependency to gpumon, which requires a nVIDIA toolkit to build


stormi:
Hi! I've reported it on IRC but I think it is good to have an open detailed issue about this here.

## Context
During the development phase of XCP-ng 8.1, I had (wrongly) decided to drop the `gpumon` package because I saw no way to build it using only open-source components. I also thought that it was only required for the nVIDIA VGPU features, who depend on the closed-source `vgpu` RPM, absent from XCP-ng. Although its source code is available and its license is FOSS-compatible (ISC License), building the binary requires a development toolkit from nVIDIA, whose license doesn't seem to be compatible with the legal requirements of the build system of an open project like XCP-ng.

## Issue
However, we found out the hard way (it went unnoticed during the beta and RC phases, unfortunately) that XAPI's startup phase would stall, without a visible error message, if an nVIDIA GPU is present and `gpumon` is missing.

In short, unless I'm mistaking about the licensing issues, XAPI currently has a runtime strong dependency to a non-free component.

There are other features from the XAPI that depend on closed-source components, but usually it will handle the absence of such a closed-source component gracefully. In the case of `gpumon`, though, the startup stalls.

## Expectation
I would expect that the Xen API be able to start and operate normally in the absence of `gpumon`, unless it turns out that my understanding of the build environment required to build it is wrong. It's not documented, so there may be too much guessing from my part.

Can we work together towards such a change? What would it involve?

robhoes:
We should definitely make it so that `gpumon` is optional and its absence is gracefully handled.

edwintorok:
gpumon has some code to handle the absence of the nvidia library at runtime, perhaps it needs to handle the absence of it at build time too.
Dune now has support for Configurator so it should be possible to check for the presence of the header and substitute either one or another .c/.ml file (where one .ml file would be a stub that always says unavailable)

robhoes:
We could change the `gpumon_client.ml` in xcp-idl to check for gpumon, ideally by confirming that its message-switch queue exists. We'd need some need functionality in message-switch for that (we talked about this before, didn't we @edwintorok ?). Then we can return an error from the rpc function right away, rather than waiting forever, and xapi can handle this error.

stormi:
For XCP-ng 8.2, we're trying a workaround that consists in building a "mock" gpumon, using `make mock`.

With that fake gpumon, XAPI is happy enough to start even if there's a nVIDIA GPU.

[...]

PR #5227, opened 2023-11-06
Add timeout to gpumon client

gthvn1:
It can happens that the message switch queue for gpumon is not available and in this case the RPC is never answered. To avoid waiting forever we add a timeout.

The condition where the RPC is not answered arise when the RRDD gpumon plugin is not started. We observed in this condition and on a host with an Nvidia GPU that the xapi wait forever because it tries to get information about the GPU but we don't see a queue `org.xen.xapi.gpumon`. Adding this timeout allows the RPC to return and xapi continues its initialization. 
We also observed that starting the RRDD gpumon plugin has the effect of creating the `org.xen.xapi.gpumon` queue and so if the plugin is started the timeout is not needed.

### DIFF
--- a/ocaml/xapi-idl/gpumon/gpumon_client.ml
+++ b/ocaml/xapi-idl/gpumon/gpumon_client.ml
@@ -16,7 +16,7 @@ let xml_url () = "file:" ^ Gpumon_interface.xml_path
 
 let rpc call =
   if !Xcp_client.use_switch then
-    Xcp_client.json_switch_rpc Gpumon_interface.queue_name call
+    Xcp_client.json_switch_rpc ~timeout:10 Gpumon_interface.queue_name call
   else
     Xcp_client.xml_http_rpc
       ~srcstr:(Xcp_client.get_user_agent ())


lindig:
This sounds sensible to me and the implementation is simple. Has this been run through system tests?

gthvn1:
On my side I just tested it on a host with Nvidia GPU and xcp-ng. So it is more manual testing. I'm not 100% sure about the side effects.

stormi:
That's the difficulty: the servers we run our test suites on don't have a nVIDIA GPU so they don't trigger the initial issue. But we can at least run more tests on this separate host Guillaume used.

psafont (approved):

psafont:
Is 30 seconds too long to wait? Does this wait happen on every xapi restast?

lindig:
I am wondering why we have not seen any hangs without this change if indeed RPC calls are not returning.

stormi:
You won't have any hangs as long as gpumon is present and running. This is related to #4072. In the past, we released XCP-ng 8.1 without gpumon and then the first users with NVIDIA GPUs reported the hang.

gthvn1:
It hangs on `Oct 25 10:04:15 xcp-ng-zen-4 xapi: [debug||0 |dbsync (update_env) R:35d4b4e35e46|dbsync] Sync: sync_gpus`
When I test it I put 10s so yes maybe 30 seconds it a little bit too long. I don't know which value can be the good one. When it works it takes less than 1s so I think that we can safely reduce the timeout to 5s.

lindig:
We can merge this and adjust to 10s. How about that?

gthvn1:
sounds good. Should I submit a new PR or can you modify it directly?

psafont:
`git commit --amend` this one :)

gthvn1:
> `git commit --amend` this one :)

Yes sure it was what I thought. I said "new" PR but think update this one ;). I will do that :+1:

psafont:
Don't know what the CI is complaining about, annoying

PR #5232, opened 2023-11-07
CA-384936 enable redo log logging

lindig:
We have a case where we don't know why the redo log is not found on boot. We need this log entry to understand if any VDIs are found.


psafont (approved):
Is the R module used anymore?

lindig:
Yes - in the more busy sections. But it makes sense to log anything to do with initialization.

lindig:
This is contained in https://github.com/xapi-project/xen-api/pull/5235 and could be disregarded if we merge the other PR.

psafont:
Yeah, let's close this

PR #5233, opened 2023-11-08
CA-365486: repository-domain-name-allowlist could accept a full hostname

duobei:
For the check of function `assert_url_is_valid`, repository-domain-name-allowlist doesn't work if you use the full FQDN for a server.

The `assert_url_is_valid` only allows if the host ends with .<entry in `repository-domain-name-allowlist`>. Now it is extended to allow if the hosts matches the entry in the list too.

### DIFF
--- a/ocaml/xapi/repository_helpers.ml
+++ b/ocaml/xapi/repository_helpers.ml
@@ -157,37 +157,73 @@ let create_repository_record ~__context ~name_label ~name_description
     ~binary_url ~source_url ~update ~hash:"" ~up_to_date:false ~gpgkey_path ;
   ref
 
+module DomainNameIncludeIP = struct
+  include Domain_name
+
+  let host_exn t =
+    try host_exn t
+    with Invalid_argument _ -> (
+      match Ipaddr.of_string (Domain_name.to_string t) with
+      | Ok ip ->
+          Ipaddr.to_domain_name ip
+      | Error _ ->
+          invalid_arg "invalid host name"
+    )
+end
+
 let assert_url_is_valid ~url =
   try
     let uri = Uri.of_string url in
-    match Uri.scheme uri with
-    | Some "http" | Some "https" -> (
-      match Uri.host uri with
-      | Some h -> (
-        match !Xapi_globs.repository_domain_name_allowlist with
-        | [] ->
+    match Uri.(scheme uri, host uri) with
+    | Some ("http" | "https"), Some h -> (
+        let subdomain =
+          DomainNameIncludeIP.(host_exn (of_string h |> Result.get_ok))
+        in
+        let invalids, valids =
+          List.partition_map
+            (fun d ->
+              match Domain_name.of_string d with
+              | Ok dom ->
+                  Right dom
+              | Error _ ->
+                  Left d
+            )
+            !Xapi_globs.repository_domain_name_allowlist
+        in
+        let hostname_allowed =
+          List.exists (fun d ->
+              DomainNameIncludeIP.(is_subdomain ~subdomain ~domain:(host_exn d))
+          )
+        in
+        match (valids, invalids) with
+        | [], [] ->
             ()
-        | l -> (
-          match
-            List.exists (fun d -> Astring.String.is_suffix ~affix:("." ^ d) h) l
-          with
-          | true ->
-              ()
-          | false ->
-              let msg = "host is not in allowlist" in
-              raise Api_errors.(Server_error (internal_error, [msg]))
-        )
+        | valids, [] when not (hostname_allowed valids) ->
+            let msg = "host is not in allowlist" in
+            raise Api_errors.(Server_error (internal_error, [msg]))
+        | _, [] ->
+            ()
+        | _ ->
+            error "Invalid domains: %s in repository-domain-name-allowlist."
+              (String.concat "," invalids) ;
+            raise
+              Api_errors.(
+                Server_error (invalid_repository_domain_allowlist, invalids)
+              )
       )
-      | None ->
-          raise
-            Api_errors.(Server_error (internal_error, ["invalid host in url"]))
-    )
+    | _, None ->
+        raise
+          Api_errors.(Server_error (internal_error, ["invalid host in url"]))
     | _ ->
         raise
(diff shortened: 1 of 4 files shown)


minglumlu (approved):

ocaml/xapi/repository_helpers.ml
```diff
@@ -169,7 +169,11 @@ let assert_url_is_valid ~url =
...
           match
-            List.exists (fun d -> Astring.String.is_suffix ~affix:("." ^ d) h) l
+            List.exists
+              (fun d ->
+                Astring.String.is_suffix ~affix:("." ^ d) h || String.equal d h
+              )
+              l
           with
           | true ->
               ()
```
psafont, line 179: I don't like all the nesting here I'd rather have a clause with a `when is_allowed domains`, in any case the way to do subdomain checking is to use [Domain_name.is_subdomain](https://hannesm.github.io/domain-name/doc/domain-name/Domain_name/index.html#val-is_subdomain)

```ocaml
let subdomain = Domain_name.host_exn h in 
let allow_hostname =
  List.exists (fun d -> Domain_name.(is_subdomain ~subdomain ~domain:(host_exn d)))
in
...
        | domains when not (allow_hostname domains) ->
            let msg = "host is not in allowlist" in
        | _ ->
            ()
```

lindig: Unrelated, but `repository_domain_name_allowlist` is a poor name. This list is empty by default, which means nothing is allowed?

robhoes: I think that is indeed the case.

duobei: `Domain_name.host_exn` is defined below:

lindig: This is the kind of code we want to avoid. Addresses in IPv6 are looking different and this would not work. At the very least, code like this should be isolated in a module.

psafont: The name of the allowd list states that those are domain names, IPs are not domain names. Whether this is the intention or not, I do not know, I'll refer to Ming to see what is the intent here

minglumlu: In product, we should expect a `static` configuration is done always as something like "xenserver.com" in `repository_domain_name_allowlist`. This is done by another XS specific RPM package.
IPs are used only for private testing purpose. The empty `repository_domain_name_allowlist` means no such allowing test would be done at all.
So I think for the function of checking against allowlist, the IPs should not be supported. For using IPs in testing, the allowlist could be set as empty.

lindig: @psafont The code tests says "TLD should not be all-numeric". TLD stands for Top Level Domain?  In any case, the code is testing for `exists is_letter (Array.get t 0)`; it this to detect IP addresses? In any case, an IPv6 address could contain letters. Altogether, this appears to me fragile.

minglumlu: With the `Domain_name`, it is good to determine the sub domains. And the checking in `Domain_name` seems better to have as the allowlist is a built-in/static configuration (no API to set/get it). In other words, the built-in allowlist should not be written with ambiguities. And since it's not exposed as API, could XAPI just accept what `Domain_name` does currently?
And for the address in URL, the built-in allowlist can enforce the restriction on it.

psafont: the RFC states that domain names are a bunch of labels separated by dots.
> Each node has a label, which is zero to 63 octets in length.  Brother
nodes may not have the same label
> When a user needs to type a domain name, the length of each label is
omitted and the labels are separated by dots (".").
> To simplify implementations, the total number of octets that represent a
domain name (i.e., the sum of all label octets and label lengths) is
limited to 255.

These properties should be all tested, otherwise we should open PRs with known-good code

duobei: In order to support for checking ip host, the `Domain_name` module extended with `Ipaddr` lib.

psafont (changes requested):

ocaml/xapi/repository_helpers.ml
```diff
@@ -174,56 +174,48 @@ end
...
-                raise Api_errors.(Server_error (internal_error, [msg]))
-          | _ ->
-              ()
-        )
-      | None ->
-          raise
-            Api_errors.(Server_error (internal_error, ["invalid host in url"]))
-    )
+      )
     | _ ->
```
psafont, line 214: The distinction between invalid host and invalid scheme should be kept. This can be done by changing the case to two different ones using
```ocaml 
    | _, None ->
          host is incorrect
    | (Some _ | None), _ ->
          scheme is incorrect
```

ocaml/xapi/repository_helpers.ml
```diff
@@ -157,37 +157,74 @@ let create_repository_record ~__context ~name_label ~name_description
...
-    error "Invalid url %s: %s" url (ExnHelper.string_of_exn e) ;
-    raise Api_errors.(Server_error (invalid_base_url, [url]))
+  with e -> (
+    match ExnHelper.error_of_exn e with
+    | error, l when error = Api_errors.invalid_repository_domain_allowlist ->
+        raise Api_errors.(Server_error (error, l))
+    | _ ->
+        error "Invalid url %s: %s" url (ExnHelper.string_of_exn e) ;
+        raise Api_errors.(Server_error (invalid_base_url, [url]))
+  )
```
psafont, line 227: I'm not sure why error_of_exn is used, there's no need for it or a nested match:
```suggestion
  with
  | Api_errors.Server_error (err, _) as e
    when err = Api_errors.invalid_repository_domain_allowlist ->
      raise e
  | e ->
      error "Invalid url %s: %s" url (ExnHelper.string_of_exn e) ;
      raise Api_errors.(Server_error (invalid_base_url, [url]))
```

duobei: Yeah, I didn't know it could directly match before.

psafont (approved):

minglumlu:
Hi @duobei,  Can you please squash the commits into one and fix the typo in commit message `the hosts matches ...`?

PR #5235, opened 2023-11-08
CA-384936 attach static VDIs for redo-log

lindig:
The redo-log without HA is missing to attach the static VDIs that
 actually contain the redo log. Call the same code that we are using in
 the HA case.

### DIFF
--- a/ocaml/xapi/xapi.ml
+++ b/ocaml/xapi/xapi.ml
@@ -531,6 +531,7 @@ let start_redo_log () =
       && not (bool_of_string (Localdb.get Constants.ha_armed))
     then (
       debug "Redo log was enabled when shutting down, so restarting it" ;
+      Static_vdis.reattempt_on_boot_attach () ;
       (* enable the use of the redo log *)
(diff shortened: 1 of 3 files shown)


ocaml/xapi/xapi.ml
```diff
@@ -531,6 +531,7 @@ let start_redo_log () =
       && not (bool_of_string (Localdb.get Constants.ha_armed))
     then (
       debug "Redo log was enabled when shutting down, so restarting it" ;
+      Static_vdis.reattempt_on_boot_attach () ;
```
psafont, line 534: Is there a way to share init code in both cases instead of adding more actions here and comparing with the HA case?

psafont (approved):

ocaml/xapi/xapi.ml
lindig: This is pretty minimal; otherwise we would need to refactor all this redo log code. I suspect this fix is missing in the DR code as well.

lindig:
The test that found this issue passes with this change.

edwintorok (approved):

PR #5236, opened 2023-11-08
uefi-certs: avoid breaking new VMs when override is set

psafont:
Before this change, the use-case to enable custom certificates was:
1. Set override-uefi-certificates=true
2. Run xe-toosltack-restart
   - xapi deletes the symlink where varstored reads from and creates an empty directory
3. User calls pool.set_uefi_certificates

Between 2 and 3, any VMs created will have no UEFI certificates, breaking secureboot for them

Furthermore, when a user wants to reset the remove custom certificates there is no way to do so without restarting the toolstack and having set previously override-uefi-certificates=false.

This path changes the behaviour (case 2b from Marcus' design) so the symlink is maintained when the filesystem contents are synchronised from the database and pool.uefi_certificates is empty. This ensure that between step 2 and 3, VMs get a correct set of certificates; and allow users to clear the value of the field and get the default certificates, without restarting the toolstack.

Tested-by: Samuel Verschelde <stormi-xcp@ylix.fr>

### DIFF
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -2773,8 +2773,12 @@ let write_uefi_certificates_to_disk ~__context ~host =
            uefi_certs_in_disk |> Array.mem cert |> log_of
        )
   in
-  match !Xapi_globs.override_uefi_certs with
-  | false ->
+  (* get from pool for consistent results across hosts *)
+  let pool_uefi_certs =
+    Db.Pool.get_uefi_certificates ~__context ~self:(Helpers.get_pool ~__context)
+  in
+  match (!Xapi_globs.override_uefi_certs, pool_uefi_certs) with
+  | false, _ ->
       let@ path =
         with_valid_symlink ~from_path:!Xapi_globs.varstore_dir
           ~to_path:!Xapi_globs.default_auth_dir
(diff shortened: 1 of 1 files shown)


ocaml/xapi/xapi_host.ml
```diff
@@ -2791,13 +2795,16 @@ let write_uefi_certificates_to_disk ~__context ~host =
...
           ~self:(Helpers.get_pool ~__context)
           ~value:disk_uefi_certs_tar
-  | true ->
-      let@ path = with_empty_dir !Xapi_globs.varstore_dir in
-      (* get from pool for consistent results across hosts *)
-      let pool_uefi_certs =
-        Db.Pool.get_uefi_certificates ~__context
-          ~self:(Helpers.get_pool ~__context)
+  | true, "" ->
+      (* When overriding certificates and user hasn't yet specified a value,
```
stormi, line 2799: This comment somehow implies that the user is going to specify a value, but to me, `override_uefi_certs = true`  doesn't mean "the user will override the certs", but rather "we leave the ability to override the certs if needed". Hence, rather than a transitional state, I see this state where we keep the symlink as XCP-ng's default state, where the certs remain in sync with the defaults unless the user installs new certs to override them.

stormi:
So, my initial tests were incomplete, and I found out later that if you want to transition from `override-uefi-certs = false` to `override-uefi-certs = true` (by modifying the setting in xapi.conf), you start with contents in the database's `pool.uefi-certificates` field. That's why a `/var/lib/varstored` directory gets created as soon as you restart XAPI. However, the contents of this field, which was likely based on the default certificates, is rejected when the time to write the files on disk comes:
```
Nov  8 18:29:12 xcpng-ci-83-b1 xapi: [ warn||0 |server_init D:e86a23dd1a22|startup] task [Sync UEFI certificates on host with XAPI db] exception: Failure("Path in certificate tarball %s contains '/'/var/lib/varstored/db.auth")
```

psafont:
Talking with Samuel, this is probably not the change xcp-ng needs. We need to go back to their requirements to understand what can be done. This is because it's not clear xcp-ng can (legally) distribute all the certificates to provide a good base by default.

PR #5238, opened 2023-11-09
Add space between header name and value

danilo-delbusso:
While I can't find a mention of the space this in the HTTP/1.1 spec, some clients cannot parse this correctly (tested with [Insomnia](https://insomnia.rest/)).

Also changing because we already do have this space for other headers, such as `content-length`.

### DIFF
--- a/ocaml/libs/http-lib/http.ml
+++ b/ocaml/libs/http-lib/http.ml
@@ -757,7 +757,7 @@ module Request = struct
     @ user_agent
     @ traceparent
     @ close
-    @ List.map (fun (k, v) -> k ^ ":" ^ v) x.additional_headers
+    @ List.map (fun (k, v) -> k ^ ": " ^ v) x.additional_headers
 
   let to_headers_and_body (x : t) =
     (* If the body is given then compute a content length *)
@@ -856,7 +856,7 @@ module Response = struct
     let task =
       Option.fold ~none:[] ~some:(fun x -> [Hdr.task_id ^ ": " ^ x]) x.task
     in
-    let headers = List.map (fun (k, v) -> k ^ ":" ^ v) x.additional_headers in
+    let headers = List.map (fun (k, v) -> k ^ ": " ^ v) x.additional_headers in
     status :: (content_length @ task @ headers)
 
   let to_headers_and_body (x : t) =


lindig (approved):

psafont:
From https://datatracker.ietf.org/doc/html/rfc2616#section-4.2
>   Each header field consists
   of a name followed by a colon (":") and the field value. Field names
   are case-insensitive. The field value MAY be preceded by any amount
   of LWS, though a single SP is preferred. Header fields can be
   extended over multiple lines by preceding each extra line with at
   least one SP or HT. Applications ought to follow "common form", where
   one is known or indicated, when generating HTTP constructs, since
   there might exist some implementations that fail to accept anything

>       message-header = field-name ":" [ field-value ]
>       field-name     = token
>       field-value    = *( field-content | LWS )
>       field-content  = <the OCTETs making up the field-value
>                        and consisting of either *TEXT or combinations
>                        of token, separators, and quoted-string>

>   The field-content does not include any leading or trailing LWS:
   linear white space occurring before the first non-whitespace
   character of the field-value or after the last non-whitespace
   character of the field-value. Such leading or trailing LWS MAY be
   removed without changing the semantics of the field value. Any LWS
   that occurs between field-content MAY be replaced with a single SP
   before interpreting the field value or forwarding the message
   downstream.

> LWS: Linear white space

The standard says to ignore leading and trailing whitespace, but I don't see why it would be needed

danilo-delbusso:
> From https://datatracker.ietf.org/doc/html/rfc2616#section-4.2
> >   Each header field consists
>    of a name followed by a colon (":") and the field value. Field names
>    are case-insensitive. The field value MAY be preceded by any amount
>    of LWS, though a single SP is preferred. Header fields can be
>    extended over multiple lines by preceding each extra line with at
>    least one SP or HT. Applications ought to follow "common form", where
>    one is known or indicated, when generating HTTP constructs, since
>    there might exist some implementations that fail to accept anything
> 
> >       message-header = field-name ":" [ field-value ]
> >       field-name     = token
> >       field-value    = *( field-content | LWS )
> >       field-content  = <the OCTETs making up the field-value
> >                        and consisting of either *TEXT or combinations
> >                        of token, separators, and quoted-string>
> 
> >   The field-content does not include any leading or trailing LWS:
>    linear white space occurring before the first non-whitespace
>    character of the field-value or after the last non-whitespace
>    character of the field-value. Such leading or trailing LWS MAY be
>    removed without changing the semantics of the field value. Any LWS
>    that occurs between field-content MAY be replaced with a single SP
>    before interpreting the field value or forwarding the message
>    downstream.
> 
> > LWS: Linear white space
> 
> The standard says to ignore leading and trailing whitespace, but I don't see why it would be needed

It's definitely a client side issue, and they should fix this on their end. 

However this PR is more to keep consistency across our headers. It's very very unimportant but an easy fix nonetheless. One could as easily argue that we should remove the whitespace, but I prefer keeping it for readability purposes (+ as mentioned this one client, which has its fair share of users)

lindig:
I would like to see this merged. The patch increases the consistency how headers are formatted and increase compatibility with clients even when these clients are not adhering to the standard; the patch is not decreasing standard compliance on the xapi side. @psafont @robhoes

psafont (approved):

PR #5240, opened 2023-11-10
CP-45847: Allow any value of trace flag for traceparent

snwoods:


psafont:
The commit message is missing context as to what are the benefits of this change, could you add them?

mg12 (approved):

psafont (approved):

PR #5241, opened 2023-11-13
CP-46004: Finish eventgen span to remove spans table clutter

snwoods:
Fix garbage collector timeout calculation as elapsed is in seconds. Add span table lengths to export span to help catch future issues.


ocaml/libs/tracing/tracing.ml
```diff
@@ -314,7 +314,7 @@ module Spans = struct
                     let elapsed =
                       Unix.gettimeofday () -. span.Span.begin_time
                     in
-                    if elapsed > !span_timeout *. 1000000. then (
+                    if elapsed > !span_timeout then (
```
psafont, line 317: This check is susceptible to clock changes, if the time spans are all created within xapi, it would be better to use [`Mtime_clock.elapsed`](https://ocaml.org/p/mtime/latest/doc/Mtime_clock/index.html#val-elapsed)

lindig (approved):

ocaml/libs/tracing/tracing.ml
Vincent-lau: I don't think we pass spans around, but only span context with its ids, and these are then used to create spans locally within each host. So local monotonic clock should be fine.

Vincent-lau:
Might be a good idea to add a comment on `span_timeout` to say it is in seconds to avoid confusion in the future.

Vincent-lau:
> The key is:
> 
> * A deadline is a point in time in the future - relative to the Epoch
> * A timeout is a timespan that has elapsed since an arbitrary start.
>   All of them can be in seconds. What is it?

Something like "the duration we allow in seconds before timing out"?

lindig:
To know that the time has expired, in one case you need to know the start time (when it's a span) and not in the other (when it's a deadline because it relative to a convention). Both designs work but it's easy to confuse the two.

ocaml/libs/tracing/tracing.ml
robhoes: But this is for _distributed_ tracing, where spans come from different processes and/or hosts, and are combined into a single trace. So we need wall-clock time on them.

ocaml/libs/tracing/tracing.ml
```diff
@@ -761,6 +761,12 @@ module Export = struct
               [
                 ("export.span.count", List.length all_spans |> string_of_int)
               ; ("export.endpoint", endpoint_to_string endpoint)
+              ; ( "xs.tracing.spans_table.count"
```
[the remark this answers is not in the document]
mg12: this field could be used to figure out if the spans table is increasing in size over time, or got full, to detect span leakage of spans that are never finished.

ocaml/libs/tracing/tracing.ml
snwoods: Yes I see a comment about the span begin_time "Using gettimeofday over Mtime as it is better for sharing timestamps between the systems". Would Ptime be better than gettimeofday though as a wall clock?

Would the clock changes matter much in the scale of a 24 hour timeout?

ocaml/libs/tracing/tracing.ml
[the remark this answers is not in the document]
snwoods: It's been added specifically to keep an eye on the span table size in the future (I actually added it after I was finished debugging, on Marcus' request)

ocaml/libs/tracing/tracing.ml
edwintorok: Ptime and gettimeofday are the same, they are both wall clocks. Ptime just offers a nicer way to format the time as a string, for opentelemetry I think you need it as a float so there isn't much difference, 'gettimeofday' should be fine for that purpose.

snwoods: Yeah so I think the best solution is to add a separate monotonic timer for removing inactive spans as Rob suggests. I'll move this to a separate PR

robhoes (approved):

PR #5242, opened 2023-11-14
CA-385080: Finish trace locally for forwarded tasks

Vincent-lau:
Some part of the span are started by `of_http_req` in `do_dispatch` but not finished afterwards. This fixes the issue where the task is forwarded and not completed, i.e. span not finished.


ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -56,9 +56,9 @@ let prototyped_of_field = function
   | "pool", "telemetry_uuid" ->
       Some "23.9.0"
   | "pool", "ext_auth_max_threads" ->
-      Some "23.27.0"
```
psafont, line 59: this change doesn't seem right

robhoes (approved):

lindig (approved):

PR #5243, opened 2023-11-15
CP-45921: Make yum commands nonessential

liulinC:
During xapi startup, it will check availability of the essential executables, and fail if not found.

yum commands should not be essential, the repo managers will be
- yum: for xs8
- dnf: for xs9

### DIFF
--- a/ocaml/xapi/xapi_globs.ml
+++ b/ocaml/xapi/xapi_globs.ml
@@ -1588,15 +1588,9 @@ module Resources = struct
       , set_iSCSI_initiator_script
       , "Path to set-iscsi-initiator script"
       )
-    ; ("yum-cmd", yum_cmd, "Path to yum command")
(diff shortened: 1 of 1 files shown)


liulinC:
toolstack-next, karma test (3866135) so far so good.

edwintorok (commented):
Is Yum and DNF's command-line compatible? You could have a generic 'repomanager_cmd' that is set to 'yum' on XS8 and 'dnf' on XS9 from the .spec file.

liulinC:
> Is Yum and DNF's command-line compatible? You could have a generic 'repomanager_cmd' that is set to 'yum' on XS8 and 'dnf' on XS9 from the .spec file.

Absolutely. not compatible, there will be followup commits for the stuff you talking about.

robhoes (approved):

lindig (approved):

PR #5244, opened 2023-11-15
CA-378591: Clear span tables when all observers are disabled

snwoods:


ocaml/libs/tracing/tracing.ml
```diff
@@ -465,6 +468,14 @@ let set ?enabled ?attributes ?endpoints ~uuid () =
               (Printf.sprintf "The TracerProvider : %s does not exist" uuid)
       in
       Hashtbl.replace tracer_providers uuid provider
+  ) ;
+  if
+    List.for_all
+      (fun provider -> not provider.TracerProvider.enabled)
+      (get_tracer_providers ())
+  then (
+    Hashtbl.clear Spans.spans ;
```
robhoes, line 477: Don't we need to grab a lock here for these two hash tables?

robhoes (approved):

PR #5246, opened 2023-11-16
CP-45888: Replace XML-RPC client in Java SDK with JSON-RPC

danilo-delbusso:
I suggest reviewing commits one at a time.

I also suggest you check out and run the output, I have built it against my SDK tag, feel free to ask for info.

As discussed this makes use of Apache's client since the Java stdlib one was having issues with the connection pool.

I am however not exposing request configuration (except timeouts). This would be trivial-ish but I didn't want to add bloat.

I have only implemented sync calls for the HTTP client. We could add the option to pass callbacks to calls and make it fully async but it's out of scope and I didn't want to make reviews even longer. 

Other notable changes:
- Now using `HashSet` instead of the generic `Set` for relevant types
- Marshalling in `Types.java` and other places has mostly been replaced by type handling from Jackson's `ObjectMapper`. This makes the source much smaller and arguably easier to understand. I have tried to start remote calls without a lot of issues. The only change needed was setting the date format to ISO 8601 Zulu instead of simple ISO 8601. Haven't encountered issues though there's the chance that xapi will send back dates with a different format. In that case I'd argue we should fix xapi, instead.
- As part of this PR I have added the options for SDK consumers to pass their own `ConsumableHttpClient` with their preferred configuration, which should suffice for most advanced usage. We had a similar PR raised against the C# SDK so I thought it'd be a nice addition.
- I have set the max no. of concurrent connections per host connection to `10`. In XenCenter we do 3(?) but I didn't want to limit the SDK. I could also leave them unlimited so feel free to weigh in.
- As part of the review requests, I have added `mli` files to ensure we don't miss unused functions.  I have only added definitions for functions that are actually consumed outside the files, so `mli` definitions for the files that end up as generation binaries are excluded.

### DIFF
--- a/ocaml/sdk-gen/common/CommonFunctions.mli
+++ b/ocaml/sdk-gen/common/CommonFunctions.mli
@@ -14,11 +14,6 @@ val is_setter : Datamodel_types.message -> bool
     @param message Message to check.
     @return [true] if the message is a setter, [false] otherwise. *)
 
-val finally : (unit -> 'a) -> always:(unit -> unit) -> 'a
-(** [finally f ~always] Executes [f] and then [always] regardless of success or failure.
-    @param f Function to execute.
-    @param always Function to execute always. *)
-
 val is_getter : Datamodel_types.message -> bool
 (** [is_getter message] Checks if a message is a getter based on its name.
     @param message Message to check.
--- a/ocaml/sdk-gen/common/CommonFunctions.mli
+++ b/ocaml/sdk-gen/common/CommonFunctions.mli
@@ -58,6 +58,13 @@ val is_remover : Datamodel_types.message -> bool
     @param message Message to check.
     @return [true] if the message is a remover, [false] otherwise. *)
 
+val get_minimum_allowed_role : Datamodel_types.message -> string
+(** [msg message] Get the minimum RBAC role required to run the procedure of the input message.
+    This function ignores internal roles.
+    If no matching role is found, the string "Not Applicable" is returned
+    @param message Input message.
+    @return string the name of the RBAC role if a matching one is found, "Not Applicable" otherwise. *)
+
 val gen_param_groups :
      Datamodel_types.message
   -> Datamodel_types.param list
--- a/ocaml/sdk-gen/common/CommonFunctions.mli
+++ b/ocaml/sdk-gen/common/CommonFunctions.mli
@@ -4,6 +4,12 @@ exception Unknown_wire_protocol
 (** Type representing supported protocols. *)
 type wireProtocol = XmlRpc | JsonRpc
 
+val get_deprecated_release :
+  (Datamodel_types.Lifecycle.change * string * 'a) list -> string
+(** [get_deprecated_release codename] Gets the non-branded release name for a lifecycle if it's deprecated
+    @param lifecycle The lifecycle transitions to check.
+    @return The non-branded release name for a lifecycle if it's deprecated, empty if not deprecated *)
+
 val get_release_branding : string -> string
 (** [get_release_branding codename] Gets the branding for a release codename.
     @param codename Release codename to lookup.
--- a/ocaml/sdk-gen/common/CommonFunctions.mli
+++ b/ocaml/sdk-gen/common/CommonFunctions.mli
@@ -119,12 +119,5 @@ val render_file : string * string -> Mustache.Json.t -> string -> string -> unit
     @param templates_dir Directory containing templates.
     @param dest_dir Directory for the rendered output. *)
 
-val json_releases :
-  [> `O of
-     ( string
-     * [> `A of
-          [> `O of (string * [> `Float of float | `String of string]) list] list
-       | `Float of float ]
-     )
-     list ]
+val json_releases : Mustache.Json.t
 (** JSON structure representing release information. *)
--- a/ocaml/sdk-gen/common/CommonFunctions.mli
+++ b/ocaml/sdk-gen/common/CommonFunctions.mli
@@ -1,0 +1,130 @@
+(** Exception for unknown wire protocol. *)
+exception Unknown_wire_protocol
+
+(** Type representing supported protocols. *)
+type wireProtocol = XmlRpc | JsonRpc
+
+val get_release_branding : string -> string
+(** [get_release_branding codename] Gets the branding for a release codename.
+    @param codename Release codename to lookup.
+    @return Branding for the release codename, or the original codename if not found. *)
+
+val is_setter : Datamodel_types.message -> bool
+(** [is_setter message] Checks if a message is a setter based on its name.
+    @param message Message to check.
+    @return [true] if the message is a setter, [false] otherwise. *)
+
+val finally : (unit -> 'a) -> always:(unit -> unit) -> 'a
+(** [finally f ~always] Executes [f] and then [always] regardless of success or failure.
+    @param f Function to execute.
+    @param always Function to execute always. *)
+
+val is_getter : Datamodel_types.message -> bool
+(** [is_getter message] Checks if a message is a getter based on its name.
+    @param message Message to check.
+    @return [true] if the message is a getter, [false] otherwise. *)
+
+val get_deprecated_info_message : Datamodel_types.message -> string
+(** [get_deprecated_info_message message] Returns a deprecated information message
+    based on the internal_deprecated_since version in the input message.
+    @param message Message containing version information.
+    @return Deprecated information message or an empty string if not deprecated. *)
+
+val is_adder : Datamodel_types.message -> bool
+(** [is_adder message] Checks if a message is an adder based on its name.
+    @param message Message to check.
+    @return [true] if the message is an adder, [false] otherwise. *)
+
+val get_published_info_class : Datamodel_types.obj -> string
+(** [get_published_info_class cls] Returns information about the first publication
+    of a class based on its lifecycle transitions.
+    @param cls Class to retrieve publication information for.
+    @return Information string with the first published version. *)
+
+val is_method_static : Datamodel_types.message -> bool
+(** [is_method_static message] Checks if a method is static based on its parameters.
+    @param message Message to check.
+    @return [true] if the method is static, [false] otherwise. *)
+
+val escape_xml : string -> string
+(** [escape_xml s] Escapes XML special characters in a string.
+    @param s String to escape.
+    @return String with XML special characters escaped. *)
+
+val get_published_info_message :
+  Datamodel_types.message -> Datamodel_types.obj -> string
+(** Gets information about the publication status of a message.
+    @param message - Message to check.
+    @param cls - Class containing the message.
+    @return Information about the publication status. *)
+
+val is_remover : Datamodel_types.message -> bool
+(** [is_remover message] Checks if a message is a remover based on its name.
+    @param message Message to check.
+    @return [true] if the message is a remover, [false] otherwise. *)
+
+val gen_param_groups :
+     Datamodel_types.message
+  -> Datamodel_types.param list
+  -> Datamodel_types.param list list
+(** Generates parameter groups based on a message and its parameters.
+    @param message - Message containing the parameters.
+    @param params - List of parameters.
+    @return List of parameter groups. *)
+
+val get_published_info_param :
+  Datamodel_types.message -> Datamodel_types.param -> string
+(** Gets information about the publication status of a parameter.
+    @param message - Message containing the parameter.
+    @param param - Parameter to check.
+    @return Information about the publication status. *)
+
+val get_published_info_field :
+  Datamodel_types.field -> Datamodel_types.obj -> string
+(** Gets information about the publication status of a field within a class.
+    @param field - Field to check.
+    @param cls - Class containing the field.
+    @return Information about the publication status. *)
+
+val string_of_file : string -> string
+(** [string_of_file filename] Reads the content of a file into a string.
+    @param filename Name of the file to read.
+    @return String containing the file content. *)
+
+val is_constructor : Datamodel_types.message -> bool
+(** [is_constructor message] Checks if a message is a constructor.
+    @param message Message to check.
+    @return [true] if the message is a constructor, [false] otherwise. *)
+
+val with_output : string -> (out_channel -> 'a) -> 'a
+(** [with_output filename f] Opens a file for writing and executes a function with the output channel.
+    @param filename Name of the file to open.
+    @param f Function to execute with the output channel. *)
+
+val is_destructor : Datamodel_types.message -> bool
+(** [is_destructor message] Checks if a message is a destructor.
+    @param message Message to check.
+    @return [true] if the message is a destructor, [false] otherwise. *)
+
+val is_real_constructor : Datamodel_types.message -> bool
+(** [is_real_constructor message] Checks if a message is a real constructor.
+    @param message Message to check.
+    @return [true] if the message is a real constructor, [false] otherwise. *)
+
+val render_file : string * string -> Mustache.Json.t -> string -> string -> unit
+(** [render_file (infile, outfile) json templates_dir dest_dir] Renders a file using a JSON data model and templates.
+    @param infile Input file name.
+    @param outfile Output file name.
+    @param json JSON data model.
+    @param templates_dir Directory containing templates.
+    @param dest_dir Directory for the rendered output. *)
+
+val json_releases :
+  [> `O of
+     ( string
+     * [> `A of
+          [> `O of (string * [> `Float of float | `String of string]) list] list
+       | `Float of float ]
+     )
+     list ]
+(** JSON structure representing release information. *)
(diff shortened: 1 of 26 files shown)


ocaml/sdk-gen/java/main.ml
```diff
@@ -159,36 +159,6 @@ let switch_enum =
 
 let _ = get_java_type switch_enum
 
-(*Helper function for get_marshall_function*)
```
psafont, line 162: I recommend creating an empty file at `ocaml/sdk-gen/java/main.mli` so the ocaml compiler can tell us if there are any unused functions.

danilo-delbusso: all done, I've actually added `mli`s where I saw fit. some comments are a bit obvious but I really wanted to have them for the params

ocaml/sdk-gen/java/main.ml
```diff
@@ -569,16 +519,15 @@ let gen_class cls folder =
   print_license file ;
   fprintf file
     "package com.xensource.xenapi;\n\n\
+     import com.fasterxml.jackson.annotation.JsonProperty;\n\
```
lindig, line 522: It could be easier to use the `{| .. |}` string that does not require escapes for newlines

ocaml/sdk-gen/common/CommonFunctions.mli
```diff
@@ -0,0 +1,130 @@
...
+(** [get_release_branding codename] Gets the branding for a release codename.
+    @param codename Release codename to lookup.
+    @return Branding for the release codename, or the original codename if not found. *)
+
+val is_setter : Datamodel_types.message -> bool
+(** [is_setter message] Checks if a message is a setter based on its name.
+    @param message Message to check.
+    @return [true] if the message is a setter, [false] otherwise. *)
+
+val finally : (unit -> 'a) -> always:(unit -> unit) -> 'a
```
lindig, line 17: This should be coming from the `Fun.protect` and not be required (or just an alias)

ocaml/sdk-gen/common/CommonFunctions.mli
```diff
@@ -0,0 +1,130 @@
...
+
+val render_file : string * string -> Mustache.Json.t -> string -> string -> unit
+(** [render_file (infile, outfile) json templates_dir dest_dir] Renders a file using a JSON data model and templates.
+    @param infile Input file name.
+    @param outfile Output file name.
+    @param json JSON data model.
+    @param templates_dir Directory containing templates.
+    @param dest_dir Directory for the rendered output. *)
+
+val json_releases :
```
lindig, line 122: Just a remark: there are competing ways to denote a JSON value in OCaml - based on the library used. This one uses the notation from Mustache. So it would make sense to give it type `Mustache.Json.t` (or `value`).

ocaml/sdk-gen/java/main.ml
psafont: I'd go for a templating engine here (mustache is available)

ocaml/sdk-gen/common/CommonFunctions.mli
danilo-delbusso: I've replaced all usages of `finally` with `Fun.protect`, instead

ocaml/sdk-gen/java/main.ml
psafont: yes, it's for another PR, my point was that we can leave it as it is, changing the code to use the new style of quoted strings doesn't seem to be worth the effort as I'd rather do something else in the end

lindig:
Can we have more eyes on this from the toolstack? @robhoes @edwintorok @minglumlu

kc284 (changes requested):
This is a first small set of comments, will add some more once I finish testing the generated code.

danilo-delbusso:
rebased on `master` to fix the Ocaml tests check

ocaml/sdk-gen/java/main.ml
```diff
@@ -821,13 +610,20 @@ let gen_enum file name ls =
...
       global_replace (regexp_string "\n") "\n         * " escaped_description
     in
-    "        /**\n"
-    ^ "         * "
-    ^ final_description
-    ^ "\n"
-    ^ "         */\n"
-    ^ "        "
-    ^ enum_of_wire name
+    let comment =
```
lindig, line 613: I suggest `String.concat "\n" [ .... ]`

ocaml/sdk-gen/powershell/common_functions.mli
```diff
@@ -0,0 +1,98 @@
...
+    @param field - OCaml field.
+    @return C# property name. *)
+
+val exposed_type : Datamodel_types.ty -> string
+(** Converts an exposed type from OCaml to its corresponding C# type.
+    @param ty - OCaml type.
+    @return Corresponding C# type. *)
+
+val pascal_case : string -> string
+(** Converts a string to PascalCase.
```
lindig, line 78: There is also `pascal_case_` further up. Maybe keep these together to highlight the difference? One is recursive and the other is not - this could be reflected in the name.

lindig (approved):

ocaml/sdk-gen/java/autogen/xen-api/src/main/java/com/xensource/xenapi/Connection.java
```diff
@@ -29,228 +29,211 @@
...
-        config.setTimeZone(TimeZone.getTimeZone("UTC"));
-        config.setServerURL(url);
-        config.setReplyTimeout(replyWait * 1000);
-        config.setConnectionTimeout(connWait * 1000);
-        XmlRpcClient client = new XmlRpcClient();
-        client.setConfig(config);
-        return client;
+    private void setAPIVersion(Session session) throws IOException {
+        try {
+            long major = session.getThisHost(this).getAPIVersionMajor(this);
```
kc284, line 179: Not caused by this PR, but we have 4 requests here to get the API version which increases the "connection" time. If we do it like the C# SDK, i.e. get the pool records, then the coordinator record, we achieve the same with 2 requests.

ocaml/sdk-gen/java/autogen/xen-api/src/main/java/com/xensource/xenapi/Connection.java
```diff
@@ -29,228 +29,211 @@
...
+     * in to any other API calls.
+     * <p>
+     * Note this constructor does NOT call Session.loginWithPassword; the programmer is responsible for calling it,
+     * passing the Connection as a parameter. No attempt to connect to the server is made until login is called.
+     * <p>
+     * When this constructor is used, a call to dispose() will do nothing. The programmer is responsible for manually
+     * logging out the Session.
+     *
+     * @param httpClient     The HttpClient used to make calls, this will be used by JsonRpcClient for handling requests
+     * @param url            The URL of the server to connect to. Should be of the form http(s)://host-url./jsonrpc or http(s)://host-url.
```
kc284, line 79: Stray fullstop in the first URL.

ocaml/sdk-gen/java/autogen/xen-api/src/main/java/com/xensource/xenapi/Connection.java
```diff
@@ -29,228 +29,211 @@
...
+     * @param requestTimeout The reply timeout for JSON-RPC calls in seconds
      */
-    private final XmlRpcClient client;
+    public Connection(CloseableHttpClient httpClient, URL url, int requestTimeout) {
```
kc284, line 82: We may want to follow the example of the C# SDK and deprecate the overloads that pass the requestTimeout or connectionTimeout, and instead provide public setters. The reason is that if the user wants to override other parameters like the keepAlive etc., adding public setters is probably easier than adding constructor overloads with a train of parameters.

ocaml/sdk-gen/java/autogen/xen-api/src/main/java/com/xensource/xenapi/Connection.java
```diff
@@ -29,228 +29,211 @@
...
 
     /**
-     * The (auto-generated parts of) the bindings dispatch XMLRPC calls on this Connection's client through this method.
+     * Send a method call to xapi's backend. You need to provide the type of the data returned by a successful response.
+     *
+     * @param methodCall            the JSON-RPC xapi method call. e.g.: session.login_with_password
+     * @param methodParameters      the methodParameters of the method call
+     * @param responseTypeReference the type of the response, wrapped with a TypeReference
+     * @param <T>                   The type of the response's payload. For instance, a map of opaque references to VM objects is expected when calling VM.get_all_records
+     * @return The result of the call with the type specified under T.
```
kc284, line 203: If you're making other changes you may want to fix the indentation of the parameters in both dispatch methods.
(Apropos, the indentation is off in some of the autogeenrated files, but no need to bother with those in this PR as they can be fixed when we switch to templates).

lindig (approved):

kc284:
The methods `Types.toX(Task,Connection)` were used to extract the result from the Task returned by the async methods. One could use Task.getResult() instead, but they were quite convenient wrappers which I expect to be used a lot - this means many client code breakages.

I am less fussed about the `Types.toX(Object)` methods because I suspect they may have been exposed to the public interface erroneously; what they do is a bit weird and it would justify internal use only. Normally one would deprecate them for a number of releases and then remove them so people have the time to change their code, but I'd be willing to overlook this since I expect the uses to be very few, if any. (Note that there is a use in the samples which is wrong and needs correction - it passes in a name label which gets eventually assigned to the opaque_ref). We will need to release note it as a breaking change.

The change of Set to HashSet breaks people's implementation too and unfortunately I expect these uses to be many. Is there a strong reason to want to avoid Set?

ocaml/sdk-gen/java/autogen/xen-api/src/main/java/com/xensource/xenapi/JsonRpcClient.java
```diff
@@ -0,0 +1,230 @@
...
+     * Format input URL to the form protocol://host/jsonrpc
+     *
+     * @param url the input URL to format
+     * @return a string version of a valid xen-api backend URL
+     */
+    private String formatBackendUrl(URL url) {
+        // We only replace it when it's empty.
+        // If the user purposely set the path
+        // we use the given value even if incorrect
+        if (!url.getPath().isEmpty()) {
```
kc284, line 225: You probably mean `if (url.getPath().isEmpty())`

danilo-delbusso:
> The change of Set to HashSet breaks people's implementation too and unfortunately I expect these uses to be many. Is there a strong reason to want to avoid Set?


quickly about this, it's just best practice

see https://docs.oracle.com/javase/tutorial/java/generics/rawTypes.html#:~:text=The%20warning%20shows%20that%20raw,Java%20compiler%20uses%20raw%20types.


however it's worse to break people's implementations without even deprecating.

we can leave it for another time, we can't rewrite everything I guess 😅

kc284:
> however it's worse to break people's implementations without even deprecating.

Unfortunately in this case deprecation does not work, so we either decide it's ok to do all major breaking changes in one go or never change it.

danilo-delbusso:
> > however it's worse to break people's implementations without even deprecating.
> 
> Unfortunately in this case deprecation does not work, so we either decide it's ok to do all major breaking changes in one go or never change it.

then I'm happy with making the changes now 🤷 should we get the team's opinion?

danilo-delbusso:
rebased on master and fixed conflict in `quality-gate.sh`

danilo-delbusso:
3rd party packages approved - fully opened PR

danilo-delbusso:
had to fix the original rebase which contained a merge commit in between - turns out I had only rebased half of the commits on top of `master` by mistake🤦‍♂️

psafont:
> > The change of Set to HashSet breaks people's implementation too and unfortunately I expect these uses to be many. Is there a strong reason to want to avoid Set?
> 
> quickly about this, it's just best practice

How is it best practice? Interfaces (which Set is) should be used instead of implementing classes (which HashSet is) in library code: it allows users to keep whatever types they decided to use in their code without any type of conversion. I was coding Java for years and using interface types was best practice, only when creating variables should the class be declared.
like so:

```java
Set<Integer> a_set = new HashSet<>()
```
Now I can use any library that processes sets, without caring how `a_set` is implemented

> see [docs.oracle.com/javase/tutorial/java/generics/rawTypes.html#:~:text=The%20warning%20shows%20that%20raw,Java%20compiler%20uses%20raw%20types](https://docs.oracle.com/javase/tutorial/java/generics/rawTypes.html#:~:text=The%20warning%20shows%20that%20raw,Java%20compiler%20uses%20raw%20types).

While I hate raw types, this is not related to Set / HashSet !

danilo-delbusso:
> had to fix the original rebase which contained a merge commit in between - turns out I had only rebased half of the commits on top of `master` by mistake🤦‍♂️



> > > The change of Set to HashSet breaks people's implementation too and unfortunately I expect these uses to be many. Is there a strong reason to want to avoid Set?
> > 
> > 
> > quickly about this, it's just best practice
> 
> How is it best practice? Interfaces (which Set is) should be used instead of implementing classes (which HashSet is) in library code: it allows users to keep whatever types they decided to use in their code without any type of conversion. I was coding Java for years and using interface types was best practice, only when creating variables should the class be declared. like so:
> 
> ```java
> Set<Integer> a_set = new HashSet<>()
> ```
> 
> Now I can use any library that processes sets, without caring how `a_set` is implemented
> 
> > see [docs.oracle.com/javase/tutorial/java/generics/rawTypes.html#:~:text=The%20warning%20shows%20that%20raw,Java%20compiler%20uses%20raw%20types](https://docs.oracle.com/javase/tutorial/java/generics/rawTypes.html#:~:text=The%20warning%20shows%20that%20raw,Java%20compiler%20uses%20raw%20types).
> 
> While I hate raw types, this is not related to Set / HashSet !

you are right and I had gotten confused - one of the reasons I reverted it :)

danilo-delbusso:
> Sorry, I missed the commit where it was reverted

np and a very valid point that gives more context :)

kc284 (changes requested):

ocaml/sdk-gen/java/main.ml
```diff
@@ -821,11 +851,20 @@ let gen_enum file name ls =
...
-    "        /**\n"
-    ^ "         * "
-    ^ final_description
+    let comment =
+      String.concat "\n"
+        ["        /**"; "         * " ^ final_description; "         */"]
+    in
+    let json_property =
+      if name != "UNRECOGNIZED" then
+        {|@JsonProperty("|} ^ name ^ {|")|}
```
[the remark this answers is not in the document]
danilo-delbusso: may be on an older commit?

it's being added as

```ocaml
    ^ "        \n"
    ^ json_property
```

Also checked the output and it seems fine 👇
image

ocaml/sdk-gen/java/main.ml
```diff
@@ -535,8 +536,9 @@ let gen_record file cls =
   fprintf file "        /**\n" ;
   fprintf file "         * Convert a %s.Record to a Map\n" cls.name ;
   fprintf file "         */\n" ;
+  fprintf file "        @SuppressWarnings(\"unchecked\")\n" ;
```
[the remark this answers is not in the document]
danilo-delbusso: that was an oversight on my part. not needed here. I also made it smarter so we're not adding the suppression to methods that don't need it further below.

See https://github.com/xapi-project/xen-api/compare/4160b11c9f462a4030c95541833cef65df2802e7..905b2922fd47471aa7c97114ded3b8de7be79e4d#diff-74ba59135d80ba9206f9f8478296df534fcb1f12a26991931f6104851ab0cf17L831

danilo-delbusso:
For latest changes see https://github.com/xapi-project/xen-api/compare/4160b11c9f462a4030c95541833cef65df2802e7..905b2922fd47471aa7c97114ded3b8de7be79e4d. I had to force push again because I had forgotten to amend commit messages

ocaml/sdk-gen/java/main.ml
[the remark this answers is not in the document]
kc284: I may have added the comment on the wrong line, it was on the enum members in Types.java, could you please check.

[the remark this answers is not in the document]
danilo-delbusso: damn actually I was the one that looked at the wrong file. the following  _is_ wrong. will update :)

```ocaml
^ "        \n"
^ json_property
```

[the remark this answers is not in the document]
danilo-delbusso: done, see https://github.com/xapi-project/xen-api/compare/e2201756d05d0c0c13f781662ce8001c8d90fc6a..647b15523f2251c7a0d7e9b95dec0dc8643bdd04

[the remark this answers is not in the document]
kc284: There are also a couple of build failures - format and tests.

danilo-delbusso:
force push was simple rebase on `master`

kc284 (approved):

Issue #5245, opened 2023-11-15
Custom UEFI certificates - API change proposal


stormi:
## Requirements

In XCP-ng, we would like to offer the ability to install custom UEFI certificates directly through XAPI, as was the case in 8.2, without requiring to modify xapi.conf on all hosts. This is for user convenience on one hand, but also because we are not sure yet whether we can distribute the UEFI certificates in our RPMs without adding restrictions of use and/or of distribution to the XCP-ng as a whole, which would make XCP-ng non-free by FLOSS standards.

When no custom certificates are set, then the default certificates should be used, exactly as what would happen if support for custom certificates were not enabled. Users would be able to set custom certificates, but also to remove them, which would go back to using the defaults.

The current implementation almost fulfills these requirements, but has a few drawbacks for us:
* When a user switches from `override-uefi-certs=false` to `override-uefi-certs=true` in xapi.conf, the `Pool.uefi_certificates` field begins with a non-empty value (written there by XAPI based on the master host's disk contents). Thus, XAPI believes that there already are custom certificates in the DB, and won't sync again with the defaults, were them to be updated in the future. However, as we intend to enable `override-uefi-certs` at installation time, we will likely avoid this issue, unless a user changes the value to `false` and then back to `true`. And we will have to tell existing testers of XCP-ng 8.3 that they must clear the custom certificates, and it's almost guaranteed that some will not see the message.
* When no custom certificates are set but overriding is supported, `/var/lib/varstored` contains a copy of the default certificates, not a direct symlink to them. Thus, if the default certificates would get updated, the contents of `/var/lib/varstored` would get out of sync.
* Once `override-uefi-certs` is true (and this was the default value we intended for XCP-ng 8.3), the contents of `Pool.uefi_certificates` doesn't not always describe the actual certificates in use. When the default certificates are used, that is no custom certificates are set, the getter will return an empty value, which means "no custom certificates", but doesn't allow to check the value of the default certificates that are used instead.

These issues all stem from the fact that there is only one DB field (Pool.uefi_certificates) for several different needs.

Here's a proposed update to the API which builds on the existing, but separates the values for custom certificates and default certificates in the database. I believe it would not change the behaviour in XenServer's context.

## xapi.conf

- rename the `override-uefi-certs` parameter to `allow-custom-uefi-certs`, which better describes what we need it to do when it's true.
- similarly, rename `override_uefi_certs` to `allow_custom_uefi_certs` in the source code.

## Proposed API changes

[...]

PR #5247, opened 2023-11-17
Custom UEFI certificates - API change

benjamreis:
Solves: https://github.com/xapi-project/xen-api/issues/5245

### DIFF
--- a/ocaml/idl/datamodel_pool.ml
+++ b/ocaml/idl/datamodel_pool.ml
@@ -1026,8 +1026,27 @@ let disable_repository_proxy =
 
 let set_uefi_certificates =
   call ~name:"set_uefi_certificates"
-    ~lifecycle:[(Published, "22.16.0", "")]
-    ~doc:"Sets the UEFI certificates for a pool and all its hosts"
+    ~lifecycle:
+      [
+        (Published, "22.16.0", "")
+      ; (Deprecated, "23.32.0", "use set_custom_uefi_certificates instead")
+      ]
+    ~doc:
+      "Set the UEFI certificates for a pool and all its hosts. Deprecated: use \
+       set_custom_uefi_certificates instead"
+    ~params:
+      [
+        (Ref _pool, "self", "The pool")
+      ; (String, "value", "The certificates to apply to the pool and its hosts")
+      ]
+    ~allowed_roles:_R_POOL_ADMIN ()
+
+let set_custom_uefi_certificates =
+  call ~name:"set_custom_uefi_certificates" ~lifecycle:[]
+    ~doc:
+      "Set custom UEFI certificates for a pool and all its hosts. Need \
+       `allow-custom-uefi-certs` set to true in conf. If empty: default back \
+       to Pool.uefi_certificates"
     ~params:
       [
         (Ref _pool, "self", "The pool")
@@ -1194,6 +1213,7 @@ let t =
       ; configure_repository_proxy
       ; disable_repository_proxy
       ; set_uefi_certificates
+      ; set_custom_uefi_certificates
       ; set_https_only
       ; set_telemetry_next_collection
       ; reset_telemetry_uuid
@@ -1382,6 +1402,9 @@ let t =
               ]
             ~default_value:(Some (VString "")) "uefi_certificates"
             "The UEFI certificates allowing Secure Boot"
+        ; field ~qualifier:StaticRO ~ty:String ~lifecycle:[]
+            ~default_value:(Some (VString "")) "custom_uefi_certificates"
+            "Custom UEFI certificates allowing Secure Boot"
         ; field ~in_product_since:rel_stockholm_psr ~qualifier:RW ~ty:Bool
             ~default_value:(Some (VBool false)) "is_psr_pending"
             "True if either a PSR is running or we are waiting for a PSR to be \
--- a/ocaml/xapi/message_forwarding.ml
+++ b/ocaml/xapi/message_forwarding.ml
@@ -1099,6 +1099,12 @@ functor
           value ;
         Local.Pool.set_uefi_certificates ~__context ~self ~value
 
+      let set_custom_uefi_certificates ~__context ~self ~value =
+        info "Pool.set_custom_uefi_certificates: pool='%s' value='%s'"
+          (pool_uuid ~__context self)
+          value ;
+        Local.Pool.set_custom_uefi_certificates ~__context ~self ~value
+
       let set_https_only ~__context ~self ~value =
         info "Pool.set_https_only: pool='%s' value='%B'"
           (pool_uuid ~__context self)
--- a/ocaml/xapi/xapi_globs.ml
+++ b/ocaml/xapi/xapi_globs.ml
@@ -858,7 +858,7 @@ let varstore_dir = ref "/var/lib/varstored"
 
 let default_auth_dir = ref "/usr/share/varstored"
 
-let override_uefi_certs = ref false
+let allow_custom_uefi_certs = ref false
 
 let disable_logging_for = ref []
 
@@ -1443,11 +1443,11 @@ let other_options =
     , (fun () -> string_of_bool !ignore_vtpm_unimplemented)
     , "Do not raise errors on use-cases where VTPM codepaths are not finished."
     )
-  ; ( "override-uefi-certs"
-    , Arg.Set override_uefi_certs
-    , (fun () -> string_of_bool !override_uefi_certs)
-    , "Enable (true) or Disable (false) overriding location for varstored UEFI \
-       certificates"
+  ; ( "allow-custom-uefi-certs"
+    , Arg.Set allow_custom_uefi_certs
+    , (fun () -> string_of_bool !allow_custom_uefi_certs)
+    , "Enable (true) or Disable (false) setting a custom location for \
+       varstored UEFI certificates"
     )
   ; ( "server-cert-group-id"
     , Arg.Set_int server_cert_group_id
--- a/ocaml/xapi/xapi_host.ml
+++ b/ocaml/xapi/xapi_host.ml
@@ -2725,7 +2725,7 @@ let really_write_uefi_certificates_to_disk ~__context ~host:_ ~value =
                  [Unix.O_WRONLY; Unix.O_CREAT; Unix.O_TRUNC]
                  0o644
              in
-             debug "override_uefi_certs: copy_file %s->%s" src dst ;
+             debug "%s: copy_file %s->%s" __FUNCTION__ src dst ;
              ignore (Unixext.copy_file src_fd dst_fd)
          )
   | base64_value -> (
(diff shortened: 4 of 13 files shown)


ocaml/idl/datamodel_pool.ml
```diff
@@ -1024,10 +1024,31 @@ let disable_repository_proxy =
     ~allowed_roles:(_R_POOL_OP ++ _R_CLIENT_CERT)
     ()
 
+let get_uefi_certificates =
+  call ~name:"get_uefi_certificates"
+    ~lifecycle:[(Published, "0.0.0", "")]
+    ~doc:"Gets the UEFI certificates used by a pool and all its hosts."
+    ~params:[(Ref _pool, "self", "The pool")]
+    ~allowed_roles:_R_POOL_ADMIN ()
+    ~result:(String, "The pool's used UEFI certificates")
```
lindig, line 1033: would drop "used"

ocaml/xapi/xapi_globs.ml
```diff
@@ -1443,11 +1443,11 @@ let other_options =
...
     )
-  ; ( "override-uefi-certs"
-    , Arg.Set override_uefi_certs
-    , (fun () -> string_of_bool !override_uefi_certs)
-    , "Enable (true) or Disable (false) overriding location for varstored UEFI \
-       certificates"
+  ; ( "allow-custom-uefi-certs"
+    , Arg.Set allow_custom_uefi_certs
+    , (fun () -> string_of_bool !allow_custom_uefi_certs)
+    , "Enable (true) or Disable (false) setting custom location for varstored \
```
lindig, line 1449: .. setting *a* custom location ..

ocaml/xapi/xapi_pool.ml
```diff
@@ -3567,16 +3567,31 @@ let disable_repository_proxy ~__context ~self =
         Db.Secret.destroy ~__context ~self:old_secret_ref
     )
 
+let get_uefi_certificates ~__context ~self =
+  let custom_certs = Db.Pool.get_custom_uefi_certificates ~__context ~self in
+  match (!Xapi_globs.allow_custom_uefi_certs, custom_certs) with
+  | false, _ | true, "" ->
```
lindig, line 3573: This suggests the empty string has a special meaning. Is this obvious for the user?

benjamreis: In the design clearing the custom UEFI certificates would default back to using the default ones.

psafont: Some docs can be added clarifying this clearing. An alternative could be to add an api function called `Pool.clear_custom_uefi_certs`

ocaml/idl/datamodel_pool.ml
```diff
@@ -1024,10 +1024,31 @@ let disable_repository_proxy =
     ~allowed_roles:(_R_POOL_OP ++ _R_CLIENT_CERT)
     ()
 
+let get_uefi_certificates =
+  call ~name:"get_uefi_certificates"
+    ~lifecycle:[(Published, "0.0.0", "")]
```
psafont, line 1029: Leave this empty so it can get autogenerated

ocaml/idl/datamodel_pool.ml
```diff
@@ -1382,6 +1405,16 @@ let t =
               ]
             ~default_value:(Some (VString "")) "uefi_certificates"
             "The UEFI certificates allowing Secure Boot"
+        ; field ~qualifier:StaticRO ~ty:String
+            ~lifecycle:
```
psafont, line 1409: Same here

ocaml/xapi/xapi_pool.ml
stormi: Note that this special meaning of an empty value is already present currently in `Db.Pool.{set,}uefi_certificates`. We just moved this to the new field, to keep the amount of changes to a minimum.

benjamreis: I've added in the doc of the setter the defaulting back when empty behavior. :)

benjamreis:
All comments should have addressed and we have tested this PR against our CI so this should be go for review and hopefully merge.
I'll resolve the schematest just before merge as it might change again.

robhoes:
If you have a field called `pool.uefi_certificates` then a getter called `pool.get_uefi_certificates` will always be generated, unless the field is marked as `internal_only` (which may have other implications).

benjamreis:
It means it's not possible to override `pool.get_uefi_certificates`?

robhoes:
> It means it's not possible to override pool.get_uefi_certificates?

If there is a public field and you change its getter, then that could lead to inconsistent behaviour. For example, the value of the field will also be returned by calls like `pool.get_record` and `pool.get_all_records`, and in events returned by `event.from`, but that would be the value as in the database.

stormi:
> > It means it's not possible to override pool.get_uefi_certificates?
> 
> If there is a public field and you change its getter, then that could lead to inconsistent behaviour. For example, the value of the field will also be returned by calls like `pool.get_record` and `pool.get_all_records`, and in events returned by `event.from`, but that would be the value as in the database.

So the issue is not technical and it's mostly consistency of the API? As long as the behaviour of the getter is documented, the override looks good to me, because it offers a single API call to get the actually used certs, without causing any changes on XenServer side. It didn't raise any objections at the design stage in #5245. However, if this is seen as a bad thing, then we'll need to simply forget about this "one API call gives the used certs" thing and just query both custom then default certificates from the client side.

> I don't fully understand the design here, but why not change the value of the actual `pool.uefi_certificates` field in the DB with the custom value, depending on the configuration?

This is what the previous implementation did, but this caused several issues detailed on top of #5245. Two fields is cleaner.

benjamreis:
We removed the override of `Pool.get_uefi_certificates` and we'll let the xapi client fetch the custom ones or the default ones whe necessary.

ocaml/xapi/message_forwarding.ml
```diff
@@ -1099,6 +1099,12 @@ functor
           value ;
         Local.Pool.set_uefi_certificates ~__context ~self ~value
 
+      let set_custom_uefi_certificates ~__context ~self ~value =
+        info "Pool.set_custom_uefi_certificates: pool='%s' value='%s'"
```
psafont, line 1103: Just in case, the certificates printed here don't usually contain private keys, do they?

ocaml/xapi/xapi_host.ml
```diff
@@ -2725,7 +2725,7 @@ let really_write_uefi_certificates_to_disk ~__context ~host:_ ~value =
                  [Unix.O_WRONLY; Unix.O_CREAT; Unix.O_TRUNC]
                  0o644
              in
-             debug "override_uefi_certs: copy_file %s->%s" src dst ;
```
psafont, line 2728: these names are outdated, using `debug "%s: ..." __FUNCTION__ ...` would be better

psafont (commented):
I can't spot any big issues, just a doubt around printing certificates, and the reminder to update the versions in the datamodel when this is ready

psafont:
Just remembered, ocaml/xapi-cli-server/records.ml needs to be modified so xe can push the certificates

stormi:
> Just remembered, ocaml/xapi-cli-server/records.ml needs to be modified so xe can push the certificates

Regarding this, maybe we could wait for the need to arise from users. We already have a script which downloads certificates from the net then sets them directly through a XAPI call, and also allows setting custom certs.

I'm not against the principle, but time flies and we need this to move forward with user testing of XCP-ng 8.3 beta :)

ocaml/idl/datamodel_pool.ml
```diff
@@ -1026,8 +1026,28 @@ let disable_repository_proxy =
 
 let set_uefi_certificates =
   call ~name:"set_uefi_certificates"
-    ~lifecycle:[(Published, "22.16.0", "")]
-    ~doc:"Sets the UEFI certificates for a pool and all its hosts"
+    ~lifecycle:
+      [
+        (Published, "22.16.0", "")
+      ; (Deprecated, "23.31.0", "use set_custom_uefi_certificates instead")
```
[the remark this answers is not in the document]
benjamreis: I'll update it right before the merge since it can chang again.

psafont (approved):
This simplifies the call graph when setting the custom uefi certificates (no more Host.set_uefi_certificates) I think it's worth adding the extra field and function calls to Pool here, at the cost of special handling in xapi_vm.

I searched for more users of uefi_certificates and couldn't find it, so the solution seems complete. If there were more users maybe a function in xapi_helpers could be added to handle this centrally. In the meantime this looks good.

Please do not merge yet, it needs changing versions in the datamodel

benjamreis:
Which version should I put? `v23.32.0`?

ocaml/idl/datamodel_pool.ml
```diff
@@ -1026,8 +1026,28 @@ let disable_repository_proxy =
...
+    ~params:
+      [
+        (Ref _pool, "self", "The pool")
+      ; (String, "value", "The certificates to apply to the pool and its hosts")
+      ]
+    ~allowed_roles:_R_POOL_ADMIN ()
+
+let set_custom_uefi_certificates =
+  call ~name:"set_custom_uefi_certificates"
+    ~lifecycle:[(Published, "0.0.0", "")]
```
psafont, line 1046: Please leave this lifecycle empty, I believe it should work

benjamreis: Done!

psafont (approved):

psafont (approved):

benjamreis:
last force push was to fix typos in the commit messages, everything is good and validated on our side so good for much according to us! :)

stormi (approved):

benjamreis:
The format error does not concern my changes, not sure what to do :thinking:

benjamreis:
rebased on master for format change.

stormi:
How about a merge of this as a Christmas gift? :wink:

ocaml/idl/datamodel_pool.ml
```diff
@@ -1026,8 +1026,27 @@ let disable_repository_proxy =
...
+        (Ref _pool, "self", "The pool")
+      ; (String, "value", "The certificates to apply to the pool and its hosts")
+      ]
+    ~allowed_roles:_R_POOL_ADMIN ()
+
+let set_custom_uefi_certificates =
+  call ~name:"set_custom_uefi_certificates" ~lifecycle:[]
+    ~doc:
+      "Set custom UEFI certificates for a pool and all its hosts. Need \
+       `allow-custom-uefi-verts` set to true in conf. If empty: default back \
```
lindig, line 1048: typo? "verts"?

lindig (approved):

ocaml/idl/datamodel_pool.ml
benjamreis: indeed! Fixed :innocent:

ocaml/idl/datamodel_pool.ml
```diff
@@ -1026,8 +1026,27 @@ let disable_repository_proxy =
 
 let set_uefi_certificates =
   call ~name:"set_uefi_certificates"
-    ~lifecycle:[(Published, "22.16.0", "")]
-    ~doc:"Sets the UEFI certificates for a pool and all its hosts"
+    ~lifecycle:
+      [
+        (Published, "22.16.0", "")
+      ; (Deprecated, "23.32.0", "use set_custom_uefi_certificates instead")
```
psafont, line 1032: This is the latest version

psafont (approved):

PR #5248, opened 2023-11-17
Simplify scanning /sys/block/<dev>/ stats for iostat

lindig:
Use scanf to to read the stats provided by the /sys file system. This rest of the code is still quite convoluted. In case of an error we are returning 0 values - not sure this is a good choice.

### DIFF
--- a/ocaml/xcp-rrdd/bin/rrdp-iostat/rrdp_iostat.ml
+++ b/ocaml/xcp-rrdd/bin/rrdp-iostat/rrdp_iostat.ml
@@ -232,41 +232,49 @@ module Stat = struct
   type t = int64 list
   (* /sys/block/tdX/stats @ /sys/block/tdX/inflight @ [read bytes; write bytes] *)
 
+  let finally = Xapi_stdext_pervasives.Pervasiveext.finally
+
+  (** open [path] to let [f] scan the contents using scanf *)
+  let scanning path f =
+    let io = Scanf.Scanning.open_in path in
+    finally (fun () -> f io) (fun () -> Scanf.Scanning.close_in io)
+
+  let sector_size dev =
+    let path = String.concat "/" ["/sys/block"; dev; "queue/hw_sector_size"] in
+    try scanning path @@ fun io -> Scanf.bscanf io " %Ld" @@ Fun.id
+    with e ->
+      D.error "%s: %s" __FUNCTION__ (Printexc.to_string e) ;
+      0L
+
+  (** read first 11 values from device stat *)
+  let stat dev =
+    let path = String.concat "/" ["/sys/block"; dev; "stat"] in
+    try
+      scanning path @@ fun io ->
+      Scanf.bscanf io " %Ld %Ld %Ld %Ld %Ld %Ld %Ld %Ld %Ld %Ld %Ld"
+      @@ fun d1 d2 d3 d4 d5 d6 d7 d8 d9 d10 d11 ->
+      [d1; d2; d3; d4; d5; d6; d7; d8; d9; d10; d11]
+    with e ->
+      D.error "%s: %s" __FUNCTION__ (Printexc.to_string e) ;
+      [0L; 0L; 0L; 0L; 0L; 0L; 0L; 0L; 0L; 0L; 0L]
+
+  let inflight dev =
+    let path = String.concat "/" ["/sys/block"; dev; "inflight"] in
+    try
+      scanning path @@ fun io ->
+      Scanf.bscanf io " %Ld %Ld" @@ fun d1 d2 -> [d1; d2]
+    with e ->
+      D.error "%s: %s" __FUNCTION__ (Printexc.to_string e) ;
+      [0L; 0L]
+
   let get_unsafe_dev (dev : string) : t =
-    let hw_sector_size =
-      Int64.of_string
-        (Unixext.file_lines_fold
-           (fun acc line -> acc ^ line)
-           ""
-           ("/sys/block/" ^ dev ^ "/queue/hw_sector_size")
-        )
-    in
-    let stats =
-      List.map Int64.of_string
-        (List.hd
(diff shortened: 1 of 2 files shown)


ocaml/xcp-rrdd/bin/rrdp-iostat/rrdp_iostat.ml
```diff
@@ -232,41 +232,51 @@ module Stat = struct
...
+      [d1; d2; d3; d4; d5; d6; d7; d8; d9; d10; d11]
+    with e ->
+      D.error "%s: %s" __FUNCTION__ (Printexc.to_string e) ;
+      [0L; 0L; 0L; 0L; 0L; 0L; 0L; 0L; 0L; 0L; 0L]
+
+  let inflight dev =
+    let path = String.concat "/" ["/sys/block"; dev; "inflight"] in
+    try
+      scanning path @@ fun io ->
+      Scanf.bscanf io " %Ld %Ld" @@ fun d1 d2 -> [d1; d2]
```
psafont, line 267: What's the format of the file? Previous code simply took all the values in the file

psafont (approved):

ocaml/xcp-rrdd/bin/rrdp-iostat/rrdp_iostat.ml
lindig: It's two numbers. I think it's better to encode expectations and to fail if they change.

ocaml/xcp-rrdd/bin/rrdp-iostat/rrdp_iostat.ml
```diff
@@ -232,41 +232,51 @@ module Stat = struct
...
+
+  (** read first 11 values from device stat *)
+  let stat dev =
+    let path = String.concat "/" ["/sys/block"; dev; "stat"] in
+    try
+      scanning path @@ fun io ->
+      Scanf.bscanf io " %Ld %Ld %Ld %Ld %Ld %Ld %Ld %Ld %Ld %Ld %Ld"
+      @@ fun d1 d2 d3 d4 d5 d6 d7 d8 d9 d10 d11 ->
+      [d1; d2; d3; d4; d5; d6; d7; d8; d9; d10; d11]
+    with e ->
```
Vincent-lau, line 259: when would this error occur, is there an example?

ocaml/xcp-rrdd/bin/rrdp-iostat/rrdp_iostat.ml
```diff
@@ -232,41 +232,51 @@ module Stat = struct
   type t = int64 list
   (* /sys/block/tdX/stats @ /sys/block/tdX/inflight @ [read bytes; write bytes] *)
 
+  let finally = Xapi_stdext_pervasives.Pervasiveext.finally
```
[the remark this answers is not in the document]
lindig: Back in the day there was no `Fun.protect` in the standard  library and you had to write something yourself. The name is taken from Java. Implementation can disagree how they treat exceptions in the finally block. I would be fine using `Fun.protect` or have an alias for it.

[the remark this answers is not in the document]
psafont: There's one key difference still, `finally` deals gracefully with exceptions raised in the cleanup function, `Fun.protect` considers this a programming error and will not bother cleaning up

ocaml/xcp-rrdd/bin/rrdp-iostat/rrdp_iostat.ml
lindig: The exception would occur if `path` can't be opened or if the the content does not match the expected format.

ocaml/xcp-rrdd/bin/rrdp-iostat/rrdp_iostat.ml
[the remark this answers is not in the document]
Vincent-lau: maybe depends on the context, but I would think it is a good practice to not have anything that would fail in the cleanup part of the code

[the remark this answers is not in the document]
edwintorok: Sometimes you don't have a choice, like closing file descriptors (you can use the no_err variant, but then you lose exceptions).

ocaml/xcp-rrdd/bin/rrdp-iostat/rrdp_iostat.ml
```diff
@@ -232,41 +232,51 @@ module Stat = struct
...
+
+  (** open [path] to let [f] scan the contents using scanf *)
+  let scanning path f =
+    let io = Scanf.Scanning.open_in path in
```
snwoods, line 239: Would ic be a better name as it's an input_channel, not input/output?

lindig: Maybe. But this is a 3-line function so I don't see room for confusion and `io` is a stronger mnemonic than `ic`. I've used `io` in other parts of the code - writing to it would be detected by the type system.

ocaml/xcp-rrdd/bin/rrdp-iostat/rrdp_iostat.ml
```diff
@@ -232,41 +232,51 @@ module Stat = struct
...
+    finally (fun () -> f io) (fun () -> Scanf.Scanning.close_in io)
+
+  let sector_size dev =
+    let path = String.concat "/" ["/sys/block"; dev; "queue/hw_sector_size"] in
+    try
+      scanning path @@ fun io ->
+      Scanf.bscanf io " %Ld" @@ fun d1 -> d1
```
Vincent-lau, line 246: `Fun.id`?

Vincent-lau (approved):
seems like here we are converting

PR #5250, opened 2023-11-20
CP-45469: Distributed tracing for xapi-clusterd

Vincent-lau:
add tracing support for clusterd.


ocaml/xapi-idl/cluster/cluster_interface.ml
```diff
@@ -271,4 +272,129 @@ module LocalAPI (R : RPC) = struct
...
+      in
+      {name; description; ty= Tuple3 (p1.ty, p2.ty, p3.ty)}
+  end
+
+  module Observer = struct
+    open TypeCombinators
+
+    let endpoints_p = Param.mk ~name:"endpoints" (list Types.string)
+
+    let bool_p = Param.mk ~name:"bool" Types.bool
```
robhoes, line 329: Parameters that are named after their type are not very descriptive. It is better to give them more meaningful names. For these one, however, I realise that you have copied the code from xenops_interface.ml.

ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -100,7 +100,7 @@ let prototyped_of_message = function
   | "host", "set_https_only" ->
       Some "22.27.0"
   | "VM", "restart_device_models" ->
-      Some "23.29.0-next"
```
psafont, line 103: This change is unrelated to the rest, please put it into a different commit

ocaml/xapi-idl/cluster/cluster_interface.ml
```diff
@@ -271,4 +272,129 @@ module LocalAPI (R : RPC) = struct
       @-> enabled_p
       @-> returning unit_p err
       )
+
+  type dict = (string * string) list [@@deriving rpcty]
+
+  module TypeCombinators = struct
```
robhoes, line 278: Looks like we have an identical module in storage_interface.ml, xenops_interface.ml and now here. It's worth putting it somewhere more central to avoid duplication. E.g. a file `ocaml/xapi-idl/lib/type_combinators.ml`.

ocaml/idl/datamodel_lifecycle.ml
Vincent-lau: done

ocaml/xapi-idl/cluster/cluster_interface.ml
Vincent-lau: indeed, and in xenops_interface.ml as well, I will remove the duplicate code

robhoes:
You don't have to do it now, but for future PR, it would be good to split things up into smaller self-contained commits if possible. For example, we could have separate commits for:
* The addition of `flush_spans` to the tracing lib
* The addition of the Observer functions to the cluster interface
* Changes to xapi_observer.ml
* Instrumentation in xapi_cluster*

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -253,7 +254,7 @@ let ip_of_str str = Cluster_interface.IPv4 str
 
 let forget ~__context ~self =
   with_clustering_lock __LOC__ (fun () ->
-      let dbg = Context.string_of_task __context in
+      let dbg = Context.string_of_task_and_tracing __context in
```
robhoes, line 257: L290 has another one.

ocaml/xapi/xapi_observer.ml
```diff
@@ -110,14 +110,128 @@ module Observer : ObserverInterface = struct
...
+let supported_components = ["xapi"; "xenopsd"; "xapi-clusterd"]
+
+(* We startup the observer for clusterd separately in cluster_host so that
+   there is no need to restart xapi in order for clusterd to be observed.
+   This does mean that observer will always be enabled for clusterd. *)
+let startup_components =
+  List.filter (( <> ) "xapi-clusterd") supported_components
 
 let get_forwarder c =
   let module Forwarder = ( val match c with
```
robhoes, line 228: I don't like that this is based on strings... Can we switch to a variant type?

ocaml/xapi/xapi_cluster_host.ml
Vincent-lau: done

ocaml/xapi-idl/cluster/cluster_interface.ml
Vincent-lau: done

quality-gate.sh
```diff
@@ -25,7 +25,7 @@ verify-cert () {
 }
 
 mli-files () {
-  N=519
+  N=520
```
robhoes, line 28: Why not add the mli instead?

ocaml/xapi/xapi_observer.ml
Vincent-lau: there is some complication where we need to store the components to a database, and converting them back to strings (e.g. the `register` function in `xapi_observer.ml`) . I am not sure if it's still worth adding a variant type

Vincent-lau: We can add the to and from string function to this though, when we read from/write to the database

ocaml/xapi/xapi_observer.ml
```diff
@@ -252,92 +368,79 @@ let unregister ~__context ~self ~host:_ =
...
       let host = Helpers.get_localhost ~__context in
       let hosts = Db.Observer.get_hosts ~__context ~self in
       if hosts = [] || List.mem host hosts then
-        register ~__context ~self ~host
+        register_component ~__context ~self ~host ~component
     )
     all
 
+let initialise_component ~__context =
+  let comp_left = ref (List.map (fun c -> (c, 1)) supported_components) in
```
robhoes, line 415: The use of `comp_left` is a bit cryptic... It is essentially doing an intersection between the set of "supported components" and "observed components"? If so, then I think it'd be clearer to literally do that set operation in `initialise` below.

Vincent-lau: The `Db.Observer.get_all` function returns a list of observers, each of which might has its own associated component. These components might overlap, e.g. [xapi], [xapi, xenopsd]. In this case, we want to make sure we call the initialisation functions liek `set_trace_log_dir` once for xapi and xenopsd.

Vincent-lau: However, it is not the case for the `load` function which needs to be called once for *every* component. E.g. with the example above we need to call `load` for `xapi` twice, and once for xenopsd.

Vincent-lau: Although I don't think it is crucially important here as most of the functions like `set_trace_log_dir` is (sort of) idempotent. Although the init function is not, because it will create an exporting thread each time it is called, so we need to be careful with this.

ocaml/xapi-idl/lib/typeCombinators.ml
```diff
@@ -0,0 +1,51 @@
...
+  let open Rpc.Types in
+  let name =
+    match name with Some n -> n | None -> Printf.sprintf "%s option" d.name
+  in
+  {name; description; ty= Option d.ty}
+
+let list ?name ?(description = []) d =
+  let open Rpc.Types in
+  let name =
+    match name with Some n -> n | None -> Printf.sprintf "list of %ss" d.name
```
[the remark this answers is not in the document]
psafont: Your proposed alternative execute the printf even if it's not used, this doesn't happen in Vincent's code and that why I prefer it in this particular case.

One way to get around this issue would be to return a `get_name : unit -> 'a` function, then execute that function to collect the value, this is not very ergonomic

ocaml/xapi/xapi_observer.ml
Vincent-lau: done as separate PR #5269

robhoes:
I think this is fine now. After #5269 has been merged, this needs to be rebased and updated, and commits squashed.

Vincent-lau:
reabsed + commits fixup. Maybe need a final review

robhoes (approved):

robhoes:
That's a clear set of commits now 👍

PR #5252, opened 2023-11-22
CP-45978 update /etc/xapi.d/plugins/power-on-host to python3 for xs9

acefei:
- using xcp.cmd to replace subprocess.popen
- using bytes.fromhex to replace struct.pack
- raise Exception instead of raw string.
- sort out some code style 

Test passed on XenRT, job id 3870994

### DIFF
--- a/scripts/poweron/DRAC.py
+++ b/scripts/poweron/DRAC.py
@@ -1,45 +1,61 @@
-import subprocess, sys, os.path
+#!/usr/bin/env python3
+
+import os.path
+import subprocess
+import sys
+
 
 class DRAC_NO_SUPP_PACK(Exception):
     """Base Exception class for all transfer plugin errors."""
+
     def __init__(self, *args):
         Exception.__init__(self, *args)
-        
+
+
 class DRAC_POWERON_FAILED(Exception):
     """Base Exception class for all transfer plugin errors."""
+
     def __init__(self, *args):
         Exception.__init__(self, *args)
 
+
 def run2(command):
-   run = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    run = subprocess.Popen(
+        command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+    )
 
-   # Wait for the process to return
-   out, err = [ e.splitlines() for e in run.communicate() ]
+    # Wait for the process to return
+    out, err = (e.splitlines() for e in run.communicate())
 
-   return run.returncode, out, err
+    return run.returncode, out, err
 
-drac_path='/opt/dell/srvadmin/sbin/racadm'
-def DRAC( power_on_ip, user, password):
-    if( not os.path.exists(drac_path)):
+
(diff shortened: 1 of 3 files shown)


scripts/poweron/wlan.py
```diff
@@ -1,34 +1,55 @@
...
-    (stdout,stderr) = proc.communicate(inputtext)
+    proc = subprocess.Popen(
+        args,
+        stdin=subprocess.PIPE,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        universal_newlines=True,
+        close_fds=True,
+    )
+    (stdout, stderr) = proc.communicate(inputtext)
```
psafont: These two variables contain bytes because the Popen didn't open them in text mode:

> If encoding or errors are specified, or text is true, the file objects stdin, stdout and stderr are opened in text mode with the specified encoding and errors, as described above in [Frequently Used Arguments](https://docs.python.org/3/library/subprocess.html#frequently-used-arguments). The universal_newlines argument is equivalent to text and is provided for backwards compatibility. By default, file objects are opened in binary mode.

https://docs.python.org/3/library/subprocess.html#subprocess.Popen

Please specify the encoding in Popen to fix this

scripts/poweron/DRAC.py
```diff
@@ -1,46 +1,62 @@
...
     """Base Exception class for all transfer plugin errors."""
+
     def __init__(self, *args):
         Exception.__init__(self, *args)
 
+
 def run2(command):
-   run = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    run = subprocess.Popen(
+        command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE
```
psafont, line 24: Doesn't universal_newlines or encoding need to be set to open the pipes in text mode?

scripts/poweron/wlan.py
```diff
@@ -4,33 +4,19 @@
...
-    )
-    (stdout, stderr) = proc.communicate(inputtext)
-    rc = proc.returncode
-    return (rc, stdout, stderr)
-
-
 def find_interface_broadcast_ip(interface):
     """Return the broadcast IP address of the supplied local interface"""
-    (rc, stdout, stderr) = doexec(["ip", "address", "show", "dev", interface])
+    (rc, stdout, stderr) = cmd.runCmd(
```
psafont, line 17: Why is this dependency preferred?

acefei: > Why is this dependency preferred?

Introducing `xcp.cmd` is because others plugins (eg: disk-space) are also using it, and its upgrade has already been completed (refer to https://github.com/xenserver/python-libs), making it convenient for us to upgrade python3.

BTW, @liulinC had a PR for requiring xcp-python-libs in xapi.spec.

scripts/poweron/DRAC.py
acefei: Now I am using xcp.cmd instead.

psafont (approved):

lindig (approved):

PR #5253, opened 2023-11-22
CP-45978 update /etc/xapi.d/plugins/disk-space to python3 for xs9

acefei:
- Comparisons to singletons like None should always be done with 'is' or 'is not', never the equality operators.
- Reduce Cyclomatic complexity.

Test passed in XenRT, job id: 3871041

### DIFF
--- a/scripts/plugins/disk-space
+++ b/scripts/plugins/disk-space
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 #
 # Copyright (C) 2017 Citrix, Inc.
 #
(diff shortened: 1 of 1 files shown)


psafont:
Please move the formatting to its own commit, it makes reviewing the changes harder than it needs to be

psafont (approved):

acefei:
> Please move the formatting to its own commit, it makes reviewing the changes harder than it needs to be

Most of the changes in this single file are formatting with black, so I'll keep the current commit, and I'll follow your guideline in future pull requests.

liulinC (approved):
Agree with @psafont  to separate the format commit for the reviewers.

PR #5254, opened 2023-11-22
CP-45977: Update scrips/extensions from python2 to python3

stephenchengCloud:


scripts/extensions/pool_update.apply
```diff
@@ -53,6 +53,7 @@ def execute_apply(session, update_package, yum_conf_file):
     cmd = ['yum', 'clean', 'all', '--noplugins', '-c', yum_conf_file]
     p = subprocess.Popen(cmd, shell=False, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, close_fds=True, env=yum_env)
     output, _ = p.communicate()
+    output = output.decode('utf-8')
```
psafont, line 56: would it be better to call `Popen` with `encoding='utf-8'` so stdout is opened in text mode?

stephenchengCloud: Yes, you're right. Corrected!

liulinC: `universal_newlines` is often used to get the `string` output.

stephenchengCloud: Corrected and tested!

psafont (approved):

scripts/extensions/pool_update.apply
Vincent-lau: probably better to use `text=True` as opposed to `universal_newlines` from [here](https://docs.python.org/3/library/subprocess.html#popen-constructor)

> New in version 3.7: text was added as a more readable alias for universal_newlines.

stormi:
Minor comment: `scrips` should be `scripts` in the commit message and in the PR title.

stephenchengCloud:
> Minor comment: `scrips` should be `scripts` in the commit message and in the PR title.

Done!

liulinC (approved):
Approved with minor comments, Please check.

scripts/extensions/pool_update.precheck
```diff
@@ -206,6 +206,7 @@ def execute_precheck(session, control_package, yum_conf_file, update_precheck_fi
     if os.path.isfile(update_precheck_file):
         pp = subprocess.Popen(update_precheck_file, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, close_fds=True)
         precheck_output, _ = pp.communicate()
+        precheck_output = precheck_output.decode("utf-8")
```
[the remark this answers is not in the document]
stephenchengCloud: Corrected!

PR #5258, opened 2023-11-22
Add interface for flush spans and exit the export thread

Vincent-lau:
This is so that daemons like clusterd can call it and flush the spans before they exit

### DIFF
--- a/ocaml/libs/tracing/tracing.ml
+++ b/ocaml/libs/tracing/tracing.ml
@@ -13,6 +13,7 @@
  *)
 module D = Debug.Make (struct let name = "tracing" end)
 
+module Delay = Xapi_stdext_threads.Threadext.Delay
 open D
 
 type endpoint = Bugtool | Url of Uri.t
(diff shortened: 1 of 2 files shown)


ocaml/libs/tracing/tracing.ml
```diff
@@ -803,19 +804,30 @@ module Export = struct
...
       Thread.create
         (fun () ->
           while true do
             debug "Tracing: Waiting %d seconds before exporting spans"
               (int_of_float !export_interval) ;
-            Thread.delay !export_interval ;
+            if not (Delay.wait delay !export_interval) then (
+              debug "Tracing: we are signaled, export spans now and exit" ;
+              flush_spans () ;
+              raise Thread.Exit
```
robhoes, line 822: I'd prefer to change the while loop to exit naturally rather than raising an exception.

psafont: I would prefer if a separate function was provided for lwt-based executables, which returns a promise (` -> unit Lwt.t`).
But that probably needs a separate library to avoid including Lwt into xapi

Vincent-lau: > exit naturally rather than raising an exception

you mean having a mutable flag and check for that?

Vincent-lau: > But that probably needs a separate library to avoid including Lwt into xapi

This probably needs a larger effort of introudcing Lwt into xapi

robhoes: > you mean having a mutable flag and check for that?

Yes, your last update is what I meant.

robhoes (approved):

mg12 (approved):

robhoes:
Looks like this needs a reformat now.

PR #5261, opened 2023-11-24
CP-45978 update /etc/xapi.d/plugins/openvswitch-config-update to python3 for xs9 

acefei:
- format with black
- using the shared lib xcp.cmd instead of subprocess.popen
- optimize code logic

Test passed with XenRT, job id: 3871613

### DIFF
--- a/scripts/plugins/openvswitch-config-update
+++ b/scripts/plugins/openvswitch-config-update
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 #
 # xapi plugin script to update the cache of configuration items in the
 # ovs-vswitchd configuration that are managed in the xapi database when
@@ -23,7 +23,7 @@
 
 import os
 import re
-import subprocess
+from xcp.cmd import runCmd
 
 import XenAPIPlugin
 
@@ -36,11 +36,8 @@ iptablectl = "/etc/xapi.d/plugins/firewall-port"
 
 # Delete the CA certificate, so that we go back to boot-strapping mode
 def delete_cacert():
-    try:
+    if os.path.exists(bootstrap_cacert):
         os.remove(bootstrap_cacert)
-    except OSError:
-        # Ignore error if file doesn't exist
-        pass
 
 
 def update(session, args):
@@ -75,8 +72,7 @@ def update(session, args):
 
     if not controller and currentControllers:
         delete_cacert()
-        cmd = [iptablectl, "close", "6640"]
-        output = subprocess.Popen(cmd, stdout=subprocess.PIPE).communicate()
+        runCmd([iptablectl, "close", "6640"])
         try:
             emergency_reset(session, None)
         except:
@@ -94,10 +90,7 @@ def update(session, args):
         new_controller = True
         ret_str += "Successfully set controller to %s.  " % controller
 
-    try:
-        pool_fail_mode = pool["other_config"]["vswitch-controller-fail-mode"]
-    except KeyError as e:
-        pool_fail_mode = None
+    pool_fail_mode = pool["other_config"].get("vswitch-controller-fail-mode")
 
     bton = {}
 
@@ -188,7 +181,7 @@ def update(session, args):
         try:
             other_config = bton[bridge]["other_config"]
             fail_mode = other_config["vswitch-controller-fail-mode"]
-        except KeyError as e:
+        except KeyError:
             fail_mode = None
 
         if fail_mode not in ["secure", "standalone"]:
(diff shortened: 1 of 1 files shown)


scripts/plugins/openvswitch-config-update
```diff
@@ -75,8 +72,7 @@ def update(session, args):
 
     if not controller and currentControllers:
         delete_cacert()
-        cmd = [iptablectl, "close", "6640"]
-        output = subprocess.Popen(cmd, stdout=subprocess.PIPE).communicate()
```
psafont, line 76: I believe the communicate is there to wait until the execution of iptablectl finishes, have you checked that runCmd waits as well?

scripts/plugins/openvswitch-config-update
```diff
@@ -94,10 +90,7 @@ def update(session, args):
         new_controller = True
         ret_str += "Successfully set controller to %s.  " % controller
 
-    try:
-        pool_fail_mode = pool["other_config"]["vswitch-controller-fail-mode"]
-    except KeyError as e:
-        pool_fail_mode = None
+    pool_fail_mode = pool["other_config"].get("vswitch-controller-fail-mode")
```
psafont, line 93: I don't think this is equivalent code, fetching other-config can also raise a KeyError

scripts/plugins/openvswitch-config-update
```diff
@@ -114,187 +110,232 @@ def update(session, args):
...
+            ]
+        ).strip('"')
 
         # Do nothing if setting is invalid, and warn the user.
-        if xapi_dib not in ['true', 'false', '']:
-            ret_str += '"' + xapi_dib + '"' + \
-                ' is an invalid value for vswitch-disable-in-band on ' + \
-                bridge + '  '
+        if xapi_dib not in ["true", "false", ""]:
+            ret_str += (
```
lindig, line 140: This looks unidiomatic; why not using string formatting to create a single string that is added to `ret_str`? Again, it is existing code.

scripts/plugins/openvswitch-config-update
```diff
@@ -114,187 +110,232 @@ def update(session, args):
...
+        ovs_dib = vswitchCfgQuery(
+            [
+                "--",
+                "--if-exists",
+                "get",
+                "Bridge",
+                bridge,
+                "other_config:disable-in-band",
+            ]
+        ).strip('"')
```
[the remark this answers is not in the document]
robhoes: I think the ovs commands may return values in quotes.

Vincent-lau:
Just curious: why changing all the single quotes to double?

psafont:
Because the authors of the formatter (black) decided to use double-quotes. It's a long-standing debate in the community, since python documentation favoured single-quoted comments.

https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#labels-strings

scripts/plugins/openvswitch-config-update
acefei: The `communicate` is also used in runCmd, refer to https://github.com/xenserver/python-libs/blob/master/xcp/cmd.py#L70.

scripts/plugins/openvswitch-config-update
acefei: Indeed, I tried to use f-string, but it led to build failures whilst py_byte_compile running (https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_source_files_and_bytecode_cache), it need to be compatible with both py2 and py3 in the current xapi.spec.

If you agreed, I can use classic string format with the % operator, even thought its performance is not as good as f-string in py3.

scripts/plugins/openvswitch-config-update
```diff
@@ -199,16 +190,12 @@ def update(session, args):
...
                 mgmt_bridge = bridge
                 host_mgmt_mac = m.group(1)
                 host_mgmt_device = pool_mgmt_macs[host_mgmt_mac]
 
-    if (
-        host_mgmt_mac is not None
-        and mgmt_bridge is not None
-        and host_mgmt_device is not None
-    ):
+    if all([host_mgmt_mac, mgmt_bridge, host_mgmt_device]):
```
psafont, line 198: While the checks are not equivalent, it's better to check for falseness in this case since the empty values would also be incorrect

psafont (approved):

scripts/plugins/openvswitch-config-update
acefei: Re-run vswitch testcase,  job (3876591) passed

PR #5266, opened 2023-11-28
CP-46149: Create new Xapi_Observer forwarder for Dom0 SMAPI

snwoods:


ocaml/xapi/xapi_observer.ml
```diff
@@ -110,14 +110,139 @@ module Observer : ObserverInterface = struct
...
+    ; otel_exporter_zipkin_endpoints: string list
+    ; xs_exporter_bugtool_endpoint: string option
+  }
+
+  let filtered_endpoints regex endpoints =
+    List.filter (fun endpoint -> Str.string_match regex endpoint 0) endpoints
+
+  let config_of_observer ~__context ~component ~observer =
+    let endpoints = Db.Observer.get_endpoints ~__context ~self:observer in
+    let zipkin_reg = Str.regexp "http://.*:9411" in
```
psafont, line 157: Why is the port hardcoded?

ocaml/xapi/xapi_observer.ml
```diff
@@ -110,14 +110,139 @@ module Observer : ObserverInterface = struct
...
+
+module Dom0ObserverConfig (Component : COMPONENT) : ObserverInterface = struct
+  let config_root = "/etc/xensource/observer/"
+
+  let env_vars_of_config (config : ObserverConfig.t) =
+    [
+      "OTEL_SERVICE_NAME=" ^ config.otel_service_name
+    ; "OTEL_RESOURCE_ATTRIBUTES="
+      ^ String.concat ","
+          (List.map (fun (k, v) -> k ^ "=" ^ v) config.otel_resource_attributes)
```
psafont: using `=` here seems fragile since it's also used for separating the env vars with their values, is this the format that opentelemetry standard dictates?

ocaml/xapi/xapi_observer.ml
```diff
@@ -110,14 +110,139 @@ module Observer : ObserverInterface = struct
...
+  let set_host_id ~__context ~host_id = ()
+
+  let set_compress_tracing_files ~__context ~enabled = ()
+end
+
+module SMObserverConfig = Dom0ObserverConfig (struct
+  let component_name = "smapi"
+end)
+
+let supported_components = ["xapi"; "xenopsd"; "smapi"]
```
psafont, line 272: we have now 3 supported components, these are strings. Do we need a variant instead here to avoid typos?

robhoes: Four: see Vincent's PR, where I suggested the same :)

mg12: Also, I think the same idea applies to this string "bugtool" in these two places:
* https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_observer.ml#L163
* https://github.com/xapi-project/xen-api/pull/5266/commits/aa2d80506f23c04e7e51252dab74e455fa99e628#diff-a31ed9ca9a7a8e83afdb6bee0ad80e86361374cd45dd4c29babf21ffcef09978R127

they should be using `Tracing.endpoint_to_string Bugtool`: https://github.com/xapi-project/xen-api/blob/master/ocaml/libs/tracing/tracing.ml#L54

Vincent-lau: might be easier if we do this in one place. I can create a new PR just for this variant and then we can buidl on that

Vincent-lau: see #5269

ocaml/xapi/xapi_observer.ml
```diff
@@ -110,14 +110,192 @@ module Observer : ObserverInterface = struct
...
+    lst
+    |> List.map (fun (key, value) ->
+           String.concat "=" [key; Filename.quote value]
+       )
+    |> String.concat "\n"
+end
+
+module ObserverConfig = struct
+  type t = {
+      otel_service_name: string
```
Vincent-lau, line 146: what does otel mean?

ocaml/xapi/xapi_observer.ml
```diff
@@ -110,14 +110,192 @@ module Observer : ObserverInterface = struct
...
+
+  let init ~__context =
+    let observer_all = Db.Observer.get_all ~__context in
+    update_all_configs ~__context ~observer_all
+
+  let set_trace_log_dir ~__context ~dir:_ =
+    let observer_all = Db.Observer.get_all ~__context in
+    update_all_configs ~__context ~observer_all
+
+  let set_export_interval ~__context:_ ~interval:_ = ()
```
Vincent-lau, line 272: so where is the export_interval set then?

psafont:
The commit adding the smapi component needs to be changed to use a variant instead of a plain string, otherwise I only have a nitpick about style

ocaml/xapi/xapi_observer.ml
```diff
@@ -160,7 +338,7 @@ let assert_valid_components components =
 
 let assert_valid_endpoints endpoints =
   let validate_endpoint = function
-    | "bugtool" ->
+    | str when Tracing.(endpoint_of_string str = Bugtool) ->
```
Vincent-lau, line 341: minor style point: we have several places where we want to use the string "bugtool", maybe instead of doing the convertion each time, we could define an Endpoint module and then hardcode a "bugtool" variable there (and then export it)?

ocaml/xapi/message_forwarding.ml
```diff
@@ -6586,7 +6586,7 @@ functor
         let fn ~rpc:_ ~session_id:_ ~host =
           do_op_on ~__context ~host ~local_fn client_fn
         in
-        Xapi_pool_helpers.call_fn_on_slaves_then_master ~__context fn
+        Xapi_pool_helpers.call_fn_on_master_then_slaves ~__context fn
```
Vincent-lau, line 6589: why are we reversing the order?

snwoods: I explained in the commit message: there was no specific reason for doing slaves then master and reversing it means that master (and therefore the DB) gets updated before all the forwarders. This means the Dom0ObserverConfig can rely on the data in the DB to update the config files. If we don't reverse it, the DB is updated last so the config files are based on old info.

ocaml/xapi/xapi_observer.ml
snwoods: It's done by the Observer (the default Xapi one)

ocaml/libs/tracing/tracing.ml
```diff
@@ -15,6 +15,63 @@ module D = Debug.Make (struct let name = "tracing" end)
...
+            false
+      in
+      let chars = explode str in
+      List.for_all (fun c -> is_tchar c) chars
+  end
+
+  module Value = struct
+    type t = string
+
+    let serialise str =
```
mg12, line 53: make instead of serialize? https://github.com/lindig/ocaml-style#introduce-and-document-interfaces

ocaml/xapi/xapi_observer.ml
```diff
@@ -124,6 +124,8 @@ module Component = struct
         "xenopsd"
     | Xapi_clusterd ->
         "xapi-clusterd"
+    | SMApi ->
+        "smapi"
```
mg12, line 128: I think we should really have something like
```
module Component
  module Name
    xapi = "xapi"
    xenopsd = "xenopsd"
    xapi_clusterd = "xapi-clusterd"
    smapi = "smapi"
  end
...
```
and then reuse these constants Component.Name.xapi, Component.Name.xenopsd etc instead of strings throughout the code.

The same thing with Tracing.Endpoint.Name.bugtool = "bugtool", as I think @Vincent-lau  suggested in another comment.

ocaml/xapi/xapi_observer.ml
```diff
@@ -249,6 +253,146 @@ end
...
+    ; otel_resource_attributes=
+        attributes_to_W3CBaggage
+          (Db.Observer.get_attributes ~__context ~self:observer)
+    ; otel_exporter_zipkin_endpoints= zipkin_endpoints endpoints
+    ; xs_exporter_bugtool_endpoint= bugtool_endpoint endpoints
+    }
+end
+
+module type OBSERVER_COMPONENT = sig
+  val component_name : string
```
mg12, line 334: instead of using `component_name : string`, it's better to use `component: Component.t` here

ocaml/libs/tracing/tracing.ml
```diff
@@ -15,13 +15,65 @@ module D = Debug.Make (struct let name = "tracing" end)
...
+        | '^'
+        | '_'
+        | '`'
+        | '|'
+        | '~' ->
+            true
+        | _ ->
+            false
+      in
+      String.to_seq str |> Seq.for_all (fun c -> is_tchar c)
```
lindig, line 44: Why not `String.for_all`? This should be most efficient because it does not allocate.

ocaml/libs/tracing/tracing.ml
```diff
@@ -15,13 +15,65 @@ module D = Debug.Make (struct let name = "tracing" end)
...
+    let serialise str =
+      let encode_char x =
+        match x with
+        (* Encode anything that isn't in basic US-ASCII or is a Control, whitespace, DQUOTE , ; or \ *)
+        | '\000' .. '\032' | '"' | ',' | ';' | '\\' | '\128' .. '\255' ->
+            Printf.sprintf "%%%02X" (Char.code x)
+        | _ ->
+            String.make 1 x
+      in
+      String.to_seq str
```
lindig, line 59: If this is used a lot, we should use something more efficient: 
* first count the number of chars that need to be escaped by iterating over it
* if zero - do nothing
* if non-zero: create a string of the right size and copy from old to new while escaping those that need it. An escaped char has a defined length - so it is possible to know the required space.
* if this is used rarely or the chance of hitting an escape is low, the escaping mechanism could still use something similar to what you are doing now but it creates a lot of garbage.

ocaml/libs/tracing/tracing.ml
```diff
@@ -15,13 +15,65 @@ module D = Debug.Make (struct let name = "tracing" end)
...
 type endpoint = Bugtool | Url of Uri.t
 
 let attribute_key_regex =
   Re.Posix.compile_pat "^[a-z0-9][a-z0-9._]{0,253}[a-z0-9]$"
 
 let validate_attribute (key, value) =
-  Re.execp attribute_key_regex key && String.length value <= 4095
+  Re.execp attribute_key_regex key
+  && W3CBaggage.Key.is_valid_key key
+  && String.length value <= 4095
```
lindig, line 76: This should be tested first - because it is cheap; you want early failure.

ocaml/xapi/xapi_observer.ml
Vincent-lau: depends on how heavily we need to use the òf_string` function. If it's only once or twice while interfacing with the external world, then this should be fine. Otherwise we might need to consider something else

ocaml/libs/tracing/tracing.ml
Vincent-lau: same question, what's the benefit of using sequence here?

ocaml/libs/tracing/tracing.ml
snwoods: I think both the chance of hitting an escape is low (because these are user input characters and users wouldn't typically use " , ; \ in an attribute value) and it is rarely used because it is only executed when an observer parameter is changed, which wouldn't be very often. So I think it is fine as is for now :)

mg12 (approved):

ocaml/libs/tracing/tracing.ml
lindig: But this code is doing a lot of work even when there is nothing to do: for a string of length _n_ it creates _n_ (new) strings of length 1 and later concatenates them (into a new string) when it could have returned the original string. And this being part of tracing, I would not assume that this function is called infrequently.

snwoods: Yes that is rather inefficient, I'll take a look at it when I can. Though whilst it's part of tracing, it isn't used as frequently as traces, it's only used when someone changes an Observer's attributes on the commandline

robhoes:
This needs a rebase.

robhoes (approved):

PR #5267, opened 2023-11-28
CP-46677: Create correct span names associated with tasks.

GabrielBuica:
Some tasks have the name "Export of VM" concatenated with the corresponding uuid. These creates thousands of different span names for the same task. This change prevents that and slightly refactors the code to remove duplicated code when creating the span for different types of tasks.

### DIFF
--- a/ocaml/xapi/context.ml
+++ b/ocaml/xapi/context.ml
@@ -218,23 +218,45 @@ let make_dbg http_other_config task_name task_id =
       (if task_name = "" then "" else " ")
       (Ref.really_pretty_and_small task_id)
 
-let tracing_of_origin (origin : origin) span_name =
+let span_kind_of_parent parent =
+  let open Tracing in
+  Option.fold ~none:SpanKind.Internal ~some:(fun _ -> SpanKind.Server) parent
+
+let parent_of_origin (origin : origin) span_name =
   let open Tracing in
   let ( let* ) = Option.bind in
-  let parent =
-    match origin with
-    | Http (req, _) ->
-        let* traceparent = req.Http.Request.traceparent in
-        let* span_context = SpanContext.of_traceparent traceparent in
-        let span = Tracer.span_of_span_context span_context span_name in
-        Some span
+  match origin with
+  | Http (req, _) ->
+      let* traceparent = req.Http.Request.traceparent in
+      let* span_context = SpanContext.of_traceparent traceparent in
+      let span = Tracer.span_of_span_context span_context span_name in
+      Some span
+  | _ ->
+      None
+
+let start_tracing_helper parent_fn task_name =
+  let open Tracing in
+  let span_details_from_task_name task_name =
+    match String.split_on_char ':' task_name with
+    | [x; y; uuid] when x = "dispatch" && y = "system.isAlive" ->
+        let span_name = x ^ ":" ^ y in
+        (span_name, [("xs.xapi.rpc.msg.uuid", uuid)])
     | _ ->
-        None
+        (task_name, [])
   in
-  let span_kind =
-    Option.fold ~none:SpanKind.Internal ~some:(fun _ -> SpanKind.Server) parent
-  in
-  (parent, span_kind)
+  let span_name, span_attributes = span_details_from_task_name task_name in
+  let parent = parent_fn span_name in
+  let span_kind = span_kind_of_parent parent in
+  let tracer = get_tracer ~name:span_name in
+  match
+    Tracer.start ~span_kind ~tracer ~attributes:span_attributes ~name:span_name
+      ~parent ()
+  with
+  | Ok x ->
+      x
+  | Error e ->
+      R.warn "Failed to start tracing: %s" (Printexc.to_string e) ;
+      None
 
 (** constructors *)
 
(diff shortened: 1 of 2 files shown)


ocaml/xapi/context.ml
```diff
@@ -347,13 +346,7 @@ let make_subcontext ~__context ?task_in_database task_name =
...
     Option.bind __context.tracing (fun parent ->
         let parent = Some parent in
-        let tracer = get_tracer ~name:task_name in
-        match Tracer.start ~tracer ~name:task_name ~parent () with
-        | Ok x ->
-            x
-        | Error e ->
-            R.warn "Failed to start tracing: %s" (Printexc.to_string e) ;
-            None
+        start_tracing_helper (fun span_name -> parent) task_name
```
psafont, line 349: The function, while correct, discards the parameter so the compiler will complain about it. Other ways of expressing this, without upsetting the compiler are: `(Fun.id parent)` or `(fun _ -> parent`

psafont:
Please check the errors, an open should be removed and some names are unused, which should be replaced as per my previous comment: https://github.com/xapi-project/xen-api/actions/runs/7018171159/job/19094357816?pr=5267

ocaml/xapi/context.ml
```diff
@@ -236,6 +236,32 @@ let tracing_of_origin (origin : origin) span_name =
...
+  let span_details_from_task_name task_name =
+    match String.split_on_char ':' task_name with
+    | [x; uuid] when x = "Export of VM" ->
+        let span_name = x in
+        (span_name, [("xs.span.arg.vm.uuid", String.trim uuid)])
+    | [x; y; uuid] when x = "dispatch" && y = "system.isAlive" ->
+        let span_name = x ^ ":" ^ y in
+        (span_name, [("xs.xapi.rpc.msg.uuid", uuid)])
+    | _ ->
+        (task_name, [])
```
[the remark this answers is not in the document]
GabrielBuica: I added a new commit for this change https://github.com/xapi-project/xen-api/pull/5267/commits/dcab4c90d96cee994abc49c309f486a6c86653b5

ocaml/xapi/context.ml
```diff
@@ -425,13 +424,12 @@ let get_user_agent context =
 let with_tracing context name f =
   let open Tracing in
   let parent = context.tracing in
-  let tracer = get_tracer ~name in
-  match Tracer.start ~tracer ~name ~parent () with
-  | Ok span ->
+  let tracing = start_tracing_helper (fun _ -> parent) name in
+  match tracing with
```
duobei, line 428: It's not necessary to bind the result of `start_tracing_helper (fun _ -> parent) name` to the `tracing` variable. `match start_tracing_helper (fun _ -> parent) name with` is ok.

ocaml/xapi/context.ml
```diff
@@ -237,15 +237,28 @@ let parent_of_origin (origin : origin) span_name =
...
-    | [x; y; uuid] when x = "dispatch" && y = "system.isAlive" ->
-        let span_name = x ^ ":" ^ y in
-        (span_name, [("xs.xapi.rpc.msg.uuid", uuid)])
-    | _ ->
-        (task_name, [])
+    let uuid_length = 36 in
+    let export_of_vm = "Export of VM" in
+    let dispatch_system_is_alive = "dispatch:system.isAlive" in
+    let open String in
+    if starts_with ~prefix:export_of_vm task_name then
```
psafont, line 244: The prefix should be "Export of VM:", like it's done in the second case

ocaml/xapi/context.ml
```diff
@@ -237,15 +237,28 @@ let parent_of_origin (origin : origin) span_name =
...
+      ( export_of_vm
+      , [
+          ( "xs.span.arg.vm.uuid"
+          , sub task_name (length export_of_vm + 2) uuid_length
+          )
+        ]
+      )
```
psafont, line 251:
```suggestion
      let uuid = 
        sub task_name (length export_of_vm + 2) uuid_length
      in
      (export_of_vm, [("xs.span.arg.vm.uuid", uuid)])
```
Is the `+2` there because there's a space between the `:` and the `uuid`?

ocaml/xapi/context.ml
```diff
@@ -237,15 +237,28 @@ let parent_of_origin (origin : origin) span_name =
...
+    else if starts_with ~prefix:"dispatch:system.isAlive:" task_name then
+      ( dispatch_system_is_alive
+      , [
+          ( "xs.span.arg.vm.uuid"
+          , sub task_name (length dispatch_system_is_alive + 1) uuid_length
+          )
+        ]
+      )
```
psafont, line 259:
```suggestion
      let uuid = 
        sub task_name (length dispatch_system_is_alive + 1) uuid_length
      in
      (dispatch_system_is_alive, [("xs.span.arg.vm.uuid", uuid)])
```

psafont:
Almost there, just a couple of cosmetic changes to make the code more compact and a comment on the first match to let the reader understand the format of the string more easily

robhoes:
The task name in question comes from the `xe vm-export` code: https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi-cli-server/cli_operations.ml#L5877-L5879. The CLI server here creates a task for itself to easily track the export call. Any API client can call `task.create` to do this, and come up with their own task name.

I worry that we are now going to add lots of rules to a fundamental xapi server module like `Context` based on what API client might use for their task names. In this case, we now rely on the task name to be formatted as `"Export of VM: %s"` elsewhere in our own code, and this string is defined in two entirely separate places.

I like some of the refactoring in this PR to reduce code duplication in context.ml, but I don't think that this is the right approach for the span-naming problem.

robhoes:
For this case, I think we can just remove the UUID from the task name in cli_operations.ml, as it serves no particular purpose. We could put it in the task description instead if we want to.

ocaml/xapi/context.ml
GabrielBuica: This one solves the issue: https://github.com/xapi-project/xen-api/pull/5267/commits/3893e6240640e893ebe4be74af1fa32466336062 .
I've also addressed @robhoes 's comments and removed the uuid from the cli instead: https://github.com/xapi-project/xen-api/pull/5267/commits/29a58d03b810c759fa0cb41f05991e0df42226c5 .

ocaml/xapi/context.ml
GabrielBuica: This change should fix it: https://github.com/xapi-project/xen-api/pull/5267/commits/a1c796623ad4bcf6ed1b5fc3696d4a350f56c79b .

PR #5269, opened 2023-11-28
Change type of observer components to use variant

Vincent-lau:
Note that we keep the interface defined in `xapi_observer.mli` the same, i.e. the `create` and `set_components` functions continue to accept string as parameters, but internally we use ocaml's variant type.

### DIFF
--- a/ocaml/xapi/xapi_observer.ml
+++ b/ocaml/xapi/xapi_observer.ml
@@ -110,33 +110,42 @@ module Observer : ObserverInterface = struct
     Tracing.Export.Destination.File.set_compress_tracing_files enabled
 end
 
-let supported_components = ["xapi"; "xenopsd"]
+module Component = struct
+  type t = Xapi | Xenopsd [@@deriving ord]
+
+  exception Unsupported_Component of string
+
+  let all = [Xapi; Xenopsd]
+
+  let to_string = function Xapi -> "xapi" | Xenopsd -> "xenopsd"
+
+  let of_string = function
+    | "xapi" ->
+        Xapi
+    | "xenopsd" ->
+        Xenopsd
+    | c ->
+        raise (Unsupported_Component c)
+end
 
 let get_forwarder c =
   let module Forwarder = ( val match c with
-                               | "xapi" ->
+                               | Component.Xapi ->
                                    (module Observer)
-                               | "xenopsd" ->
+                               | Component.Xenopsd ->
                                    (module Xapi_xenops.Observer)
-                               | _ ->
-                                   failwith
-                                     (Printf.sprintf
-                                        "Not a valid component: %s. Valid \
-                                         components are: %s "
-                                        c
-                                        (String.concat ", " supported_components)
-                                     ) : ObserverInterface
+                             : ObserverInterface
                          )
   in
   (module Forwarder : ObserverInterface)
 
-module StringSet = Set.Make (String)
+module ComponentSet = Set.Make (Component)
 
 let observed_hosts_of ~__context hosts =
   match hosts with [] -> Db.Host.get_all ~__context | hosts -> hosts
 
 let observed_components_of components =
-  match components with [] -> supported_components | components -> components
+  match components with [] -> Component.all | components -> components
 
 let assert_valid_hosts ~__context hosts =
   List.iter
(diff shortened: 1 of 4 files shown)


ocaml/xapi/xapi_observer.ml
```diff
@@ -110,27 +110,34 @@ module Observer : ObserverInterface = struct
     Tracing.Export.Destination.File.set_compress_tracing_files enabled
 end
 
-let supported_components = ["xapi"; "xenopsd"]
+module Observer_component = struct
```
psafont, line 113: I believe `Component` can be used instead

ocaml/xapi/xapi_observer.ml
```diff
@@ -110,27 +110,34 @@ module Observer : ObserverInterface = struct
...
+  let to_string = function Xapi -> "xapi" | Xenopsd -> "xenopsd"
+
+  let of_string = function
+    | "xapi" ->
+        Xapi
+    | "xenopsd" ->
+        Xenopsd
+    | c ->
+        failwith (Printf.sprintf "Not a valid component: %s" c)
+end
```
psafont, line 125: I recommend adding `let all = [Xapi; Xenopsd]`, and use `Components.all` instead of `supported_components`.

It's both shorter and lets the programmer know that it should contain all the values of the variant, maybe define it right after the variant so it's even more obvious it needs to be updated.

Alternatively, check whether [ppx_enumerate](https://github.com/janestreet/ppx_enumerate) is available

robhoes:
The CI says:
```
File "ocaml/xapi/dune", line 156, characters 50-66:
156 |   (preprocess (pps ppx_deriving_rpc ppx_sexp_conv ppx_deriving.ord))
                                                        ^^^^^^^^^^^^^^^^
Error: Library "ppx_deriving.ord" not found.
```

What do you need it for, by the way? For the set?

Vincent-lau:
> The CI says:
> 
> ```
> File "ocaml/xapi/dune", line 156, characters 50-66:
> 156 |   (preprocess (pps ppx_deriving_rpc ppx_sexp_conv ppx_deriving.ord))
>                                                         ^^^^^^^^^^^^^^^^
> Error: Library "ppx_deriving.ord" not found.
> ```
> 
> What do you need it for, by the way? For the set?

Yes, for the set

robhoes (approved):

ocaml/xapi/xapi_observer.ml
```diff
@@ -152,7 +163,8 @@ let assert_valid_hosts ~__context hosts =
 let assert_valid_components components =
   List.iter
     (fun component ->
-      if not (List.mem component supported_components) then
+      if not (List.mem component (List.map Component.to_string Component.all))
+      then
         raise
           Api_errors.(Server_error (invalid_value, ["component"; component]))
     )
```
psafont, line 170: We can exploit exceptions here:
```
      try List.find (Component.of_string component) Component.all
      with _ ->
          raise
            Api_errors.(Server_error (invalid_value, ["component"; component]))
```

Vincent-lau: or maybe just a `List.iter` to let the `of_string` raise the exception

psafont: I don't understand, there's only one string that needs to be verified, so why would List.iter be useful here?

Vincent-lau: ah I mean remove the `List.find` and let `of_string` raise the exception. Look at my new commits and see what you think

psafont (approved):

ocaml/xapi/xapi_observer.ml
psafont: ah, much better

ocaml/xapi/xapi_observer.ml
```diff
@@ -381,7 +392,10 @@ let set_attributes ~__context ~self ~value =
         Forwarder.set_attributes ~__context ~uuid
           ~attributes:(default_attributes @ value)
       )
-      (observed_components_of (Db.Observer.get_components ~__context ~self))
+      (Db.Observer.get_components ~__context ~self
+      |> List.map Component.of_string
+      |> observed_components_of
+      )
```
duobei, line 398: Maybe we could let 
```OCaml
   Db.Observer.get_components ~__context ~self
   |> List.map Component.of_string
   |> observed_components_of
```` 
to be a function here, it appears three times in this place.

PR #5272, opened 2023-11-29
CA-365059: Clear source pool messages after migrating VM

danilo-delbusso:
Please be nit-picky ;) 

Looks like we weren't moving messages if the VM was on a supporter at all in the past (`xapi_messages:get` only looks on the file system of the host it's called on, so the messages weren't being returned).

This PR fixes both the issue in the ticket and moves VM messages if the VM was originally on a coordinator. Using RPC calls since internal xapi calls fail, see #5216 and its revert PR #5239.


ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -1603,8 +1603,22 @@ let migrate_send' ~__context ~vm ~dest ~live:_ ~vdi_map ~vif_map ~vgpu_map
         XenAPI.VM.set_ha_always_run ~rpc:remote.rpc ~session_id:remote.session
           ~self:new_vm ~value:true ;
       (* Send non-database metadata *)
-      Xapi_message.send_messages ~__context ~cls:`VM ~obj_uuid:vm_uuid
+      let rpc = Helpers.make_rpc ~__context in
+      let session_id = Context.get_session_id __context in
+      (* We fetch and destroy messages via RPC calls because they are stored on the master host *)
+      let messages =
+        XenAPI.Message.get ~rpc ~session_id ~cls:`VM ~obj_uuid:vm_uuid
+          ~since:(Date.of_float 0.0)
```
psafont, line 1611:
```suggestion
          ~since:Date.epoch
```

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -1603,8 +1603,22 @@ let migrate_send' ~__context ~vm ~dest ~live:_ ~vdi_map ~vif_map ~vgpu_map
...
+      in
+      Xapi_message.send_messages ~__context ~cls:`VM ~obj_uuid:vm_uuid ~messages
```
psafont, line 1613: Currently, there's no way to know whether this fails in the code, only a logline is printed. This means that the messages will be lost as the messages will be destroyed.

Since the code is not allowed to fail in any case, this probably doesn't matter much as there's no nice way to treat this

psafont (approved):

Vincent-lau:
I assume when we are migrating a vm from a host to itself nothing would happen?

ocaml/xapi/xapi_message.ml
```diff
@@ -817,13 +817,15 @@ let handler (req : Http.Request.t) fd _ =
           (ExnHelper.string_of_exn e)
   )
 
-(* Export messages and send to another host/pool over http. *)
-let send_messages ~__context ~cls ~obj_uuid ~session_id ~remote_address =
-  let msgs = get ~__context ~cls ~obj_uuid ~since:(Date.of_float 0.0) in
-  let body = export_xml msgs in
+(* Export and send messages stored on the current host to another host/pool over http. *)
```
robhoes, line 820: Not "messages stored on the current host", but messages explicitly passed to this function.

danilo-delbusso:
> I assume when we are migrating a vm from a host to itself nothing would happen?

Yup see line `1600`:

```ocaml
if (not is_intra_pool) && not copy then (
      [...]
let rpc = Helpers.make_rpc ~__context in
      let session_id = Context.get_session_id __context in
      (* We fetch and destroy messages via RPC calls because they are stored on the master host *)
      let messages =
        XenAPI.Message.get ~rpc ~session_id ~cls:`VM ~obj_uuid:vm_uuid
          ~since:Date.epoch
      in
      Xapi_message.send_messages ~__context ~cls:`VM ~obj_uuid:vm_uuid ~messages
        ~session_id:remote.session ~remote_address:remote.remote_master_ip ;
      info
        "Destroying %s messages belonging to VM ref=%s uuid=%s from the source \
         host"
        (List.length messages |> Int.to_string)
        (Ref.string_of vm) vm_uuid ;
      let message_refs = List.rev_map fst messages in
      XenAPI.Message.destroy_many ~rpc ~session_id ~messages:message_refs ;
```

ocaml/xapi/xapi_vm_migrate.ml
danilo-delbusso: what about preventing the destruction and creating a new message?

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -1603,8 +1603,22 @@ let migrate_send' ~__context ~vm ~dest ~live:_ ~vdi_map ~vif_map ~vgpu_map
...
         ~session_id:remote.session ~remote_address:remote.remote_master_ip ;
+      info
+        "Destroying %s messages belonging to VM ref=%s uuid=%s from the source \
+         host"
```
robhoes, line 1617:
```suggestion
         pool, after sending them to the remote pool"
```

robhoes:
>Looks like we weren't moving messages if the VM was on a supporter at all in the past (xapi_messages:get only looks on the file system of the host it's called on, so the messages weren't being returned).

Right, looks like this has never worked at all!

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -1603,8 +1603,22 @@ let migrate_send' ~__context ~vm ~dest ~live:_ ~vdi_map ~vif_map ~vgpu_map
         XenAPI.VM.set_ha_always_run ~rpc:remote.rpc ~session_id:remote.session
           ~self:new_vm ~value:true ;
       (* Send non-database metadata *)
-      Xapi_message.send_messages ~__context ~cls:`VM ~obj_uuid:vm_uuid
+      let rpc = Helpers.make_rpc ~__context in
+      let session_id = Context.get_session_id __context in
```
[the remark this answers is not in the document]
danilo-delbusso: would it be better design to have `Xapi_message.send_messages` in `Helpers.call_api_functions`'s `f` here or two make two `Helpers.call_api_functions` calls to wrap the two RPC calls we're making?

ocaml/xapi/xapi_vm_migrate.ml
danilo-delbusso: what's usually done in these cases? I can't see a customer being able to resolve it so maybe a new message isn't the correct way :/  I do not like however that we're destroying data here - even if it's not that important

ocaml/xapi/xapi_vm_migrate.ml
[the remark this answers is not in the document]
danilo-delbusso: I'm thinking that the former is a bit cleaner-looking and having `Xapi_message.send_messages` in `f` doesn't matter, but I'm not sure of what's considered best in this code base :)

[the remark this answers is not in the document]
robhoes: Something like this is fine:
```
   let messages = XenAPI.Message.get ... in
   Xapi_message.send_messages ... messages ... ;
   XenAPI.Message.destroy_many .... messages ...
)
```
Xapi_message.send_messages doesn't use the "local" rpc and session_id, but it is still allowed in there.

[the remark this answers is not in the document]
danilo-delbusso: thought so! thanks for confirming :)

robhoes (approved):

robhoes:
I did a "Squash and Merge".

PR #5280, opened 2023-12-06
CP-45981: Update xenopsd from python2 to python3

stephenchengCloud:
1. Update xenopsd to python3
2. Fix some tab and space inconsistency
3. Remove unused import modules

Test:
3879814 auto cert kit (network and VM related test)
3879823 vCPUPinning (qemu-related test)
191817 Ring 0 BST 
191818 ToolStack Input

### DIFF
--- a/ocaml/xenopsd/scripts/common.py
+++ b/ocaml/xenopsd/scripts/common.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 #
 # Copyright (c) 2011-2013 Citrix Systems, Inc.
 #
@@ -38,7 +38,8 @@ def send_to_syslog(msg):
 def doexec(args):
     """Execute a subprocess, then return its return code, stdout and stderr"""
     send_to_syslog(args)
-    proc = subprocess.Popen([ "/usr/bin/env", "PATH=%s" % path ] + args,stdin=None,stdout=subprocess.PIPE,stderr=subprocess.PIPE,close_fds=True)
+    proc = subprocess.Popen([ "/usr/bin/env", "PATH=%s" % path ] + args, stdin=None, stdout=subprocess.PIPE,
+                            stderr=subprocess.PIPE, close_fds=True, universal_newlines=True)
     rc = proc.wait()
     stdout = proc.stdout
     stderr = proc.stderr
@@ -101,7 +102,7 @@ def add_to_bridge(mode, dev, bridge, address, external_ids):
         cmd = ["ovs-vsctl", "--timeout=30", "--", "--if-exists", "del-port", dev, "--", "add-port", bridge, dev]
         for (key, value) in external_ids:
              cmd = cmd + ["--", "set", "interface", dev, 'external-ids:"%s"="%s"' % (key, value) ]
-	run(ON_ERROR_LOG, cmd)
+        run(ON_ERROR_LOG, cmd)
 
 def remove_from_bridge(mode, dev, bridge):
     if mode == MODE_BRIDGE:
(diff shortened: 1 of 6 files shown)


ocaml/xenopsd/scripts/setup-vif-rules
```diff
@@ -196,7 +193,7 @@ def clear_xs_ofport_path(vif):
 
 def make_vswitch_external_ids(vif):
     args = []
-    for (k, v) in vif.get_external_ids().items():
+    for (k, v) in list(vif.get_external_ids().items()):
```
liulinC, line 196: I know the return type of `{}.items()` are different, 
but are't both of them are iterable?  is this `list` necessary?  (but has not bad anyway, just curious)

stephenchengCloud: Yes, you're right. It's not necessary. :)

stephenchengCloud: I think you mean line 95:
def do_chain_action(executable, action, chain_name, args=[]):

ocaml/xenopsd/scripts/common.py
```diff
@@ -206,11 +207,11 @@ def get_words(value, separator):
         private = self.json["extra_private_keys"]
```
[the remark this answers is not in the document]
stephenchengCloud: For set_ethtool(), the caller does pass a `mode`. I don't know if the original code wants to keep the possibility for further extension. So keep it as it is.

ocaml/xenopsd/scripts/qemu-vif-script
```diff
@@ -1,6 +1,6 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
+
 
-from __future__ import print_function
```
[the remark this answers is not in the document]
stephenchengCloud: yes. The original code imports common which has sys.
Anyway, importing sys explicitly is the best practice.

stephenchengCloud:
Updated according to all the comments above.
Testing...

ocaml/xenopsd/scripts/qemu-wrapper
```diff
@@ -327,7 +326,7 @@ def main(argv):
 
     qemu = subprocess.Popen(qemu_args, executable=qemu_dm, env=qemu_env,
                             stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
-                            preexec_fn=prepare_exec)
+                            preexec_fn=prepare_exec, pass_fds=open_fds)
```
rosslagerwall, line 329: I can't see how the above `subprocess.call([clipboardd, ...)` works after the Py3 conversion. It expects an open fd but it has neither `close_fds=False` or `pass_fds=...`.

In any case, I think it would be preferable to use the `pass_fds` option in the call to xs-clipboardd and get rid of the preexec_fn option.

stephenchengCloud: Thanks for the review.
Code updated!

robhoes (approved):
This looks fine as far as I can see, but I'd like @rosslagerwall to confirm if his comment has been addressed appropriately.

ocaml/xenopsd/scripts/setup-vif-rules
```diff
@@ -92,7 +89,9 @@ def remove_bridge_port(vif_name):
 def get_chain_name(vif_name):
     return ("FORWARD_%s" % vif_name)
 
-def do_chain_action(executable, action, chain_name, args=[]):
+def do_chain_action(executable, action, chain_name, args = None):
```
rosslagerwall, line 92: PEP 8 says no spaces around keyword arguments.

ocaml/xenopsd/scripts/qemu-wrapper
```diff
@@ -310,7 +309,7 @@ def main(argv):
 
     clipboardd = '/opt/xensource/libexec/xs-clipboardd'
     subprocess.call([clipboardd, "-d", str(domid), "-s", str(s2.fileno())],
-                    preexec_fn=close_fds)
+                    pass_fds=[s2.fileno()])
```
rosslagerwall, line 312: The code could probably do with some refactoring after this change. `close_fds` is short and only used once so it could probably be inlined. It also weirdly relies on global state rather than taking a parameter.

This refactoring can be done later.

rosslagerwall (approved):

ocaml/xenopsd/scripts/setup-vif-rules
stephenchengCloud: Thanks! Modified.

stephenchengCloud:
Make the commits one and the keep the commit message clean.

liulinC (approved):

PR #5276, opened 2023-12-02
CP-45565: Add new guidance fields to API and CLI

gangj:
This is the initial commit for the update guidance improvement:
    
    1. Update comments for re-purposing host.pending_guidances.
    2. Add new pending-guidance fields: host.pending_guidances_recommended,
       host.pending_guidances_full, vm.pending_guidances_recommended and
       vm.pending_guidances_full.
    3. Add definition for new guidance: RestartVM,
       reboot_host_on_kernel_livepatch_failure and
       reboot_host_on_xen_livepatch_failure
    4. Add DB upgrade rule: replace reboot_host_on_livepatch_failure in
       "host.pending_guidances" with both
       reboot_host_on_kernel_livepatch_failure and
       reboot_host_on_xen_livepatch_failure in
       "host.pending_guidances_recommended”
    5. Clear reboot_host_on_kernel_livepatch_failure and
       reboot_host_on_xen_livepatch_failure on reboot
    6. CLI: expose those new added pending guidance fields

PR #5281, opened 2023-12-07
CP-45566: Support guidance data in new format in updateinfo.xml

minglumlu:
This PR includes changes to support guidance data in new format in updateinfo.xml. The XAPI will support guidance data in updateinfo.xml in either old or new format for sake of backwards compatibility. Then XAPI will convert the guidance data and expose to XAPI clients through interfaces like HTTP /updates and pending guidance lists. But for simplicity, XAPI will not support backwards compatibility on these XAPI interfaces. This means a new XenCenter is required to co-work with the XAPI changes in this PR.

### DIFF
--- a/ocaml/tests/test_repository_helpers.ml
+++ b/ocaml/tests/test_repository_helpers.ml
@@ -187,118 +187,6 @@ module UpdateOfJsonTest = Generic.MakeStateless (struct
       ]
 end)
 
-module GuidanceSetAssertValidGuidanceTest = Generic.MakeStateless (struct
-  module Io = struct
-    type input_t = Guidance.t list
-
-    type output_t = (unit, exn) result
-
-    let string_of_input_t l =
-      Fmt.(str "%a" Dump.(list string)) (List.map Guidance.to_string l)
-
-    let string_of_output_t =
-      Fmt.(str "%a" Dump.(result ~ok:(any "()") ~error:exn))
-  end
-
-  let transform input =
-    try Ok (GuidanceSet.assert_valid_guidances input) with e -> Error e
-
-  let tests =
-    let open Guidance in
-    `QuickAndAutoDocumented
-      [
-        ([], Ok ())
-      ; ([RebootHost], Ok ())
-      ; ([RestartToolstack], Ok ())
-      ; ([RestartDeviceModel], Ok ())
-      ; ([EvacuateHost], Ok ())
-      ; ([EvacuateHost; RestartToolstack], Ok ())
-      ; ([RestartDeviceModel; RestartToolstack], Ok ())
-      ; ( [RestartDeviceModel; EvacuateHost]
-        , Error
-            Api_errors.(
-              Server_error
-                ( internal_error
-                , [GuidanceSet.error_msg [RestartDeviceModel; EvacuateHost]]
-                )
-            )
-        )
-      ; ( [EvacuateHost; RestartToolstack; RestartDeviceModel]
-        , Error
-            Api_errors.(
-              Server_error
-                ( internal_error
-                , [
-                    GuidanceSet.error_msg
-                      [EvacuateHost; RestartToolstack; RestartDeviceModel]
-                  ]
-                )
-            )
-        )
-      ; ( [RebootHost; RestartToolstack]
-        , Error
-            Api_errors.(
-              Server_error
-                ( internal_error
-                , [GuidanceSet.error_msg [RebootHost; RestartToolstack]]
-                )
-            )
-        )
-      ; ( [RebootHost; RestartDeviceModel]
-        , Error
-            Api_errors.(
-              Server_error
-                ( internal_error
-                , [GuidanceSet.error_msg [RebootHost; RestartDeviceModel]]
-                )
-            )
-        )
-      ; ( [RebootHost; EvacuateHost]
-        , Error
-            Api_errors.(
-              Server_error
-                ( internal_error
-                , [GuidanceSet.error_msg [RebootHost; EvacuateHost]]
-                )
-            )
-        )
-      ]
-end)
-
-let fields_of_updateinfo =
-  Fmt.Dump.
-    [
-      field "id" (fun (r : UpdateInfo.t) -> r.id) string
-    ; field "summary" (fun (r : UpdateInfo.t) -> r.summary) string
-    ; field "description" (fun (r : UpdateInfo.t) -> r.description) string
-    ; field "rec_guidance"
-        (fun (r : UpdateInfo.t) -> UpdateInfo.guidance_to_string r.rec_guidance)
-        string
-    ; field "abs_guidance"
-        (fun (r : UpdateInfo.t) -> UpdateInfo.guidance_to_string r.abs_guidance)
-        string
-    ; field "guidance_applicabilities"
-        (fun (r : UpdateInfo.t) ->
-          List.map Applicability.to_string r.guidance_applicabilities
-        )
-        (list string)
-    ; field "spec_info" (fun (r : UpdateInfo.t) -> r.spec_info) string
-    ; field "url" (fun (r : UpdateInfo.t) -> r.url) string
-    ; field "update_type" (fun (r : UpdateInfo.t) -> r.update_type) string
-    ; field "livepatch_guidance"
-        (fun (r : UpdateInfo.t) ->
-          UpdateInfo.guidance_to_string r.livepatch_guidance
-        )
-        string
-    ; field "livepatches"
-        (fun (r : UpdateInfo.t) ->
-          List.map
-            (fun x -> x |> LivePatch.to_json |> Yojson.Basic.pretty_to_string)
-            r.livepatches
-        )
-        (list string)
-    ]
-
 module AssertUrlIsValid = Generic.MakeStateless (struct
   module Io = struct
     type input_t = string * string list
@@ -520,10 +408,10 @@ module EvalGuidanceForOneUpdate = Generic.MakeStateless (struct
         updates_info: (string * UpdateInfo.t) list
       ; update: Update.t
       ; upd_ids_of_livepatches: string list
-      ; upd_ids_of_failed_livepatches: string list
+      ; kind: Guidance.kind
     }
 
-    type output_t = Guidance.t option
+    type output_t = Guidance.t list
 
     let fields_of_input =
       Fmt.Dump.
@@ -548,37 +436,24 @@ module EvalGuidanceForOneUpdate = Generic.MakeStateless (struct
               |> Printf.sprintf "[%s]"
             )
             string
-        ; field "upd_ids_of_failed_livepatches"
-            (fun (r : input_t) ->
-              r.upd_ids_of_failed_livepatches
-              |> String.concat ";"
-              |> Printf.sprintf "[%s]"
-            )
+        ; field "kind"
+            (fun (r : input_t) -> Guidance.kind_to_string r.kind)
             string
         ]
 
     let string_of_input_t = Fmt.(str "%a" Dump.(record @@ fields_of_input))
 
-    let string_of_output_t g =
-      Fmt.(str "%a" Dump.(string)) (UpdateInfo.guidance_to_string g)
+    let string_of_output_t l =
+      Fmt.(str "%a" Dump.(list string)) (List.map Guidance.to_string l)
   end
 
-  let transform
-      Io.
-        {
-          updates_info
-        ; update
-        ; upd_ids_of_livepatches
-        ; upd_ids_of_failed_livepatches
-        } =
-    eval_guidance_for_one_update ~updates_info ~update
-      ~kind:Guidance.Recommended
+  let transform Io.{updates_info; update; upd_ids_of_livepatches; kind} =
+    eval_guidance_for_one_update ~updates_info ~update ~kind
       ~upd_ids_of_livepatches:(UpdateIdSet.of_list upd_ids_of_livepatches)
-      ~upd_ids_of_failed_livepatches:
-        (UpdateIdSet.of_list upd_ids_of_failed_livepatches)
 
   let tests =
     let open Io in
+    let open Guidance in
     `QuickAndAutoDocumented
       [
         (* Update ID in update can't be found in updateinfo list *)
@@ -600,9 +475,9 @@ module EvalGuidanceForOneUpdate = Generic.MakeStateless (struct
                 ; repository= "regular"
                 }
           ; upd_ids_of_livepatches= []
-          ; upd_ids_of_failed_livepatches= []
+          ; kind= Mandatory
           }
-        , None
+        , []
         )
       ; (* Update ID in update can't be found in updateinfo list *)
         ( {
@@ -614,13 +489,17 @@ module EvalGuidanceForOneUpdate = Generic.MakeStateless (struct
                       id= "UPDATE-0000"
                     ; summary= "summary"
                     ; description= "description"
-                    ; rec_guidance= Some Guidance.EvacuateHost
-                    ; abs_guidance= Some Guidance.RebootHost
+                    ; guidance=
+                        [
+                          (Mandatory, [EvacuateHost])
+                        ; (Recommended, [])
+                        ; (Full, [RebootHost])
+                        ; (Livepatch, [])
+                        ]
                     ; guidance_applicabilities= []
                     ; spec_info= "special info"
                     ; url= "https://update.details.info"
                     ; update_type= "security"
-                    ; livepatch_guidance= None
                     ; livepatches= []
                     ; issued= Xapi_stdext_date.Date.epoch
                     ; severity= Severity.None
@@ -632,13 +511,17 @@ module EvalGuidanceForOneUpdate = Generic.MakeStateless (struct
                       id= "UPDATE-0001"
                     ; summary= "summary"
                     ; description= "description"
-                    ; rec_guidance= Some Guidance.EvacuateHost
-                    ; abs_guidance= Some Guidance.RebootHost
+                    ; guidance=
+                        [
+                          (Mandatory, [EvacuateHost])
+                        ; (Recommended, [])
+                        ; (Full, [RebootHost])
+                        ; (Livepatch, [])
+                        ]
                     ; guidance_applicabilities= []
                     ; spec_info= "special info"
                     ; url= "https://update.details.info"
                     ; update_type= "security"
-                    ; livepatch_guidance= None
                     ; livepatches= []
                     ; issued= Xapi_stdext_date.Date.epoch
                     ; severity= Severity.None
@@ -661,9 +544,9 @@ module EvalGuidanceForOneUpdate = Generic.MakeStateless (struct
                 ; repository= "regular"
                 }
           ; upd_ids_of_livepatches= []
-          ; upd_ids_of_failed_livepatches= []
+          ; kind= Mandatory
           }
-        , None
+        , []
         )
       ; (* No update ID in update *)
         ( {
@@ -675,13 +558,17 @@ module EvalGuidanceForOneUpdate = Generic.MakeStateless (struct
                       id= "UPDATE-0000"
                     ; summary= "summary"
                     ; description= "description"
-                    ; rec_guidance= Some Guidance.EvacuateHost
-                    ; abs_guidance= Some Guidance.RebootHost
+                    ; guidance=
+                        [
+                          (Mandatory, [EvacuateHost])
+                        ; (Recommended, [])
+                        ; (Full, [RebootHost])
+                        ; (Livepatch, [])
+                        ]
                     ; guidance_applicabilities= []
                     ; spec_info= "special info"
                     ; url= "https://update.details.info"
                     ; update_type= "security"
-                    ; livepatch_guidance= None
                     ; livepatches= []
                     ; issued= Xapi_stdext_date.Date.epoch
                     ; severity= Severity.None
@@ -703,9 +590,9 @@ module EvalGuidanceForOneUpdate = Generic.MakeStateless (struct
                 ; repository= "regular"
                 }
           ; upd_ids_of_livepatches= []
-          ; upd_ids_of_failed_livepatches= []
+          ; kind= Recommended
           }
-        , None
+        , []
         )
       ; (* Empty applicabilities *)
         ( {
@@ -717,13 +604,17 @@ module EvalGuidanceForOneUpdate = Generic.MakeStateless (struct
                       id= "UPDATE-0000"
                     ; summary= "summary"
                     ; description= "description"
-                    ; rec_guidance= None
-                    ; abs_guidance= None
+                    ; guidance=
+                        [
+                          (Mandatory, [])
+                        ; (Recommended, [])
+                        ; (Full, [])
+                        ; (Livepatch, [])
+                        ]
                     ; guidance_applicabilities= []
                     ; spec_info= "special info"
                     ; url= "https://update.details.info"
                     ; update_type= "security"
-                    ; livepatch_guidance= None
                     ; livepatches= []
                     ; issued= Xapi_stdext_date.Date.epoch
                     ; severity= Severity.None
@@ -735,13 +626,17 @@ module EvalGuidanceForOneUpdate = Generic.MakeStateless (struct
                       id= "UPDATE-0001"
                     ; summary= "summary"
                     ; description= "description"
-                    ; rec_guidance= Some Guidance.RebootHost
-                    ; abs_guidance= None
+                    ; guidance=
+                        [
+                          (Mandatory, [])
+                        ; (Recommended, [RebootHost])
+                        ; (Full, [])
+                        ; (Livepatch, [])
+                        ]
                     ; guidance_applicabilities= [] (* No applicabilities *)
                     ; spec_info= "special info"
                     ; url= "https://update.details.info"
                     ; update_type= "security"
-                    ; livepatch_guidance= None
                     ; livepatches= []
                     ; issued= Xapi_stdext_date.Date.epoch
                     ; severity= Severity.None
@@ -763,9 +658,9 @@ module EvalGuidanceForOneUpdate = Generic.MakeStateless (struct
                 ; repository= "regular"
                 }
           ; upd_ids_of_livepatches= []
-          ; upd_ids_of_failed_livepatches= []
+          ; kind= Recommended
           }
-        , Some Guidance.RebootHost
+        , [RebootHost]
         )
       ; (* Matched applicability *)
         ( {
@@ -777,13 +672,17 @@ module EvalGuidanceForOneUpdate = Generic.MakeStateless (struct
                       id= "UPDATE-0000"
                     ; summary= "summary"
                     ; description= "description"
-                    ; rec_guidance= None
-                    ; abs_guidance= None
+                    ; guidance=
+                        [
+                          (Mandatory, [])
+                        ; (Recommended, [])
+                        ; (Full, [])
+                        ; (Livepatch, [])
+                        ]
                     ; guidance_applicabilities= []
                     ; spec_info= "special info"
                     ; url= "https://update.details.info"
                     ; update_type= "security"
-                    ; livepatch_guidance= None
                     ; livepatches= []
                     ; issued= Xapi_stdext_date.Date.epoch
                     ; severity= Severity.None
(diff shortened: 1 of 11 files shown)


ocaml/xapi/updateinfo.ml
```diff
@@ -441,13 +449,76 @@ module Severity = struct
...
 end
 
+module GuidanceOfUpdateInfo = struct
+  type t = (Guidance.kind * Guidance.t list) list
+
+  let value_of_xml = function
+    | Xml.Element ("value", _, [Xml.PCData v]) ->
+        Some (Guidance.of_string v)
+    | Xml.Element (unexpected, _, _) ->
+        warn "Unexpected guidance value XML tag %s" unexpected ;
```
liulinC, line 477: `Unexpected guidance value XML tag` -> 'Ignore unexpected guidance value XML tag' ?

ocaml/xapi/updateinfo.ml
```diff
@@ -441,13 +449,76 @@ module Severity = struct
...
+
+  let of_xml xml_blocks =
+    List.filter_map
+      (function
+        | Xml.Element (kind_xml, _, guidances) -> (
+          match kind_of_xml kind_xml with
+          | kind ->
+              let values = List.filter_map value_of_xml guidances in
+              Some (kind, values)
+          | exception Unsupported_kind_xml ->
```
liulinC, line 509: Ocaml can match type as well as exception in such way, good to know

ocaml/xapi/updateinfo.ml
```diff
@@ -441,13 +449,76 @@ module Severity = struct
...
+              warn "Unsupported guidance kind XML tag %s" kind_xml ;
+              None
+        )
+        | _ ->
```
liulinC, line 513: The piece of xml is ignored, do you want to log it?

gangj:
Maybe to update the commit message:

In the commit message of the first commit: 3b6ea2cc8f06d9527fac1b37c94de6d2177da64b:

`XAPI parses the guidance from the data in old format only if there is no
data in new format presented in the XML file.`

From the code, I think all the data including the old guidances are parsed first, but will convert the old ones only if there is no new one.

ocaml/xapi/repository_helpers.ml
```diff
@@ -672,22 +673,16 @@ let eval_guidances ~updates_info ~updates ~kind ~livepatches ~failed_livepatches
...
-          GuidanceSet.add g acc
-      | None ->
-          acc
+      eval_guidance_for_one_update ~updates_info ~update:u ~kind
+        ~upd_ids_of_livepatches
+      |> GuidanceSet.of_list
+      |> GuidanceSet.union acc
     )
     GuidanceSet.empty updates
   |> append_livepatch_guidances ~updates_info ~upd_ids_of_livepatches
```
gangj, line 657: Sorry, I cannot understand the logic here, I see that the livepatch guidances have been taking into account in `eval_guidance_for_one_update` for `recommended` guidance, but why we need to `append_livepatch_guidances ` for every `kind` of guidance again?
Would you please help to explain it a bit, Thank you!

minglumlu: This is for the stand-alone livepatches. RPM packages have been updated but livepatches are still available.
```
 622 (* In case that the RPM in an update has been installed (including livepatch file),
 623  * but the livepatch has not been applied.
 624  * In other words, this RPM update will not appear in parameter [updates] of
 625  * function [eval_guidances], but the livepatch in it is still applicable.
 626  *)
```

gangj:
For the commit message of the 2nd one: 740bbd9bc2fbe1e07bdf352137a72f2856c97f99

`Additionally, the evaluation is simplified by not considering the failed
livapatches any more. This may cause un-necessary carrying out of
livepatch guidances. But the overall impact of carrying out these
guidances is considered as ignorable.
`

might be good to add:
1. how livepatch guidance is used now: livepatch guidance will be used as recommended guidance no matter if the livepatch is applied successfully or failed
2. why. Something like; as update guidance will be carried out by xapi client(XenCenter), this will make the client's implementation easy, it doesn't need to check the result of the livepatch to determine what guidances are pending to be applied.

ocaml/xapi/xapi_host.ml
```diff
@@ -3010,15 +3010,7 @@ let apply_updates ~__context ~self ~hash =
...
-  List.map
-    (fun g ->
-      [
-        Api_errors.updates_require_recommended_guidance
-      ; Updateinfo.Guidance.to_string g
-      ]
-    )
-    guidances
-  @ warnings
+  warnings
```
gangj, line 3013: Do we need to remove error `updates_require_recommended_guidance` as it is not used now?

ocaml/xapi/repository.ml
```diff
@@ -720,40 +668,41 @@ let apply_updates' ~__context ~host ~updates_info ~livepatches ~acc_rpm_updates
...
-         *)
-        guidances'
-        |> List.filter (fun g -> g <> Guidance.RebootHost)
-        |> List.cons Guidance.RebootHostOnLivePatchFailure
+      )
+
+let apply_updates' ~__context ~host ~updates_info ~livepatches ~acc_rpm_updates
+    =
+  let open Guidance in
+  (* This function runs on coordinator host *)
```
gangj, line 676: Would be better to move this line of comment up as the first line of this function as original or just change it to a function comment above the function.

ocaml/xapi/repository.ml
```diff
@@ -720,40 +668,41 @@ let apply_updates' ~__context ~host ~updates_info ~livepatches ~acc_rpm_updates
...
+  let get_guidances kind =
+    eval_guidances ~updates_info ~updates:acc_rpm_updates ~kind ~livepatches
+    |> GuidanceSet.remove EvacuateHost
+    |> GuidanceSet.elements
   in
+  let mandatory = get_guidances Mandatory in
+  (* Install RPM updates firstly *)
+  Helpers.call_api_functions ~__context (fun rpc session_id ->
+      Client.Client.Repository.apply ~rpc ~session_id ~host
+  ) ;
```
gangj, line 686: As it is the first thing done in this function, would be clear to move these lines to the start of this function as original.

ocaml/xapi/xapi_host.ml
```diff
@@ -3010,15 +3010,7 @@ let apply_updates ~__context ~self ~hash =
   Db.Host.set_last_software_update ~__context ~self
     ~value:(get_servertime ~__context ~host:self) ;
   Db.Host.set_latest_synced_updates_applied ~__context ~self ~value:`yes ;
-  List.map
-    (fun g ->
-      [
-        Api_errors.updates_require_recommended_guidance
```
gangj, line 3016: Do we need to remove error `updates_require_recommended_guidance` as it is not used now?

ocaml/xapi/repository.ml
```diff
@@ -695,18 +654,7 @@ let apply_livepatches' ~__context ~host ~livepatches =
...
-    =
-  (* This function runs on coordinator host *)
-  (* Install RPM updates firstly *)
-  Helpers.call_api_functions ~__context (fun rpc session_id ->
-      Client.Client.Repository.apply ~rpc ~session_id ~host
-  ) ;
-  (* Apply live patches with best effort *)
-  let successful_livepatches, failed_livepatches =
-    apply_livepatches' ~__context ~host ~livepatches
-  in
```
gangj, line 708: Sorry, may I know if there is any change on how livepatches are applied, seems the function `apply_livepatches` is not changed, but the comment changed from
(* Apply live patches with best effort *)
to
(* Apply livepatches *)

ocaml/xapi/repository_helpers.ml
```diff
@@ -1361,15 +1324,139 @@ let prune_updateinfo_for_livepatches livepatches updateinfo =
...
+         | `reboot_host_on_xen_livepatch_failure
+         | `reboot_host_on_kernel_livepatch_failure ]
+      -> unit
+  ; vms_get: unit -> (string * Guidance.t list) list
+  ; vm_add: string -> [`restart_device_model | `restart_vm] -> unit
+  ; vm_remove: string -> [`restart_device_model | `restart_vm] -> unit
+}
+
+let get_ops_of_pending ~__context ~host ~kind =
+  let get_pending_guidances_of_host ~get =
```
gangj, line 1391: `get` -> `db_get`

ocaml/xapi/repository_helpers.ml
```diff
@@ -1361,15 +1324,139 @@ let prune_updateinfo_for_livepatches livepatches updateinfo =
...
+          ~value
+      in
+      {host_get; host_add; host_remove; vms_get; vm_add; vm_remove}
+  | _ ->
+      raise Api_errors.(Server_error (internal_error, ["Not implemented kind"]))
+
+let set_pending_guidances ~ops ~coming =
+  let pending_of_host =
+    ops.host_get () |> List.filter (fun x -> not (is_livepatch_failure x))
+  in
```
gangj, line 1446: I think any livepatch guidances in `host.pending_guidances` have been "updated" in the first [commit](https://github.com/xapi-project/xen-api/pull/5276/files#diff-0b5dc72783f904599eeed5fdcd93fbdcf368b71edc18f657d123a8982bc5402c) of the feature branch?

ocaml/xapi/repository_helpers.ml
```diff
@@ -1361,15 +1324,139 @@ let prune_updateinfo_for_livepatches livepatches updateinfo =
...
+  ; vms_get: unit -> (string * Guidance.t list) list
+  ; vm_add: string -> [`restart_device_model | `restart_vm] -> unit
+  ; vm_remove: string -> [`restart_device_model | `restart_vm] -> unit
+}
+
+let get_ops_of_pending ~__context ~host ~kind =
+  let get_pending_guidances_of_host ~get =
+    get ~__context ~self:host
+    |> List.map (fun g -> Guidance.of_pending_guidance g)
+  in
```
gangj, line 1394: Seems we can use `Db.Host.get_pending_guidances` directly inside the function without passing it as parameter `get`?

ocaml/xapi/repository.ml
```diff
@@ -720,40 +668,41 @@ let apply_updates' ~__context ~host ~updates_info ~livepatches ~acc_rpm_updates
...
-  , List.map
-      (fun (lp, _) ->
-        [Api_errors.apply_livepatch_failed; LivePatch.to_string lp]
-      )
-      failed_livepatches
-  )
+    (fun g -> debug "mandatory pending_guidance: %s" (to_string g))
+    mandatory ;
+  let ops = get_ops_of_pending ~__context ~host ~kind:Mandatory in
+  set_pending_guidances ~ops ~coming:mandatory ;
```
gangj, line 702: Only mandatory guidances are set? When will recommended and full guidances be set?

ocaml/xapi/repository_helpers.ml
```diff
@@ -1361,15 +1324,139 @@ let prune_updateinfo_for_livepatches livepatches updateinfo =
...
+  | _ ->
+      raise Api_errors.(Server_error (internal_error, ["Not implemented kind"]))
+
+let set_pending_guidances ~ops ~coming =
+  let pending_of_host =
+    ops.host_get () |> List.filter (fun x -> not (is_livepatch_failure x))
+  in
+  let to_be_added, to_be_removed =
+    merge_with_pending_guidances ~pending:pending_of_host ~coming
+  in
```
gangj, line 1449: As only new XenCenter can use this new version of xapi, then XenCenter should finish applying any pending mandatory guidances if there is any, when XC start to apply update, I think there should have no pending mandatory guidances on any host?
Then when will we need to merge coming mandatory ones with pending? Is this needed only when those pending guidance are from the original `host.pending_guidances` for recommended guidances of old xapi? Then should XC also ensure those legacy guidances got applied before start applying a new update?

ocaml/xapi/repository_helpers.ml
```diff
@@ -1361,15 +1324,139 @@ let prune_updateinfo_for_livepatches livepatches updateinfo =
...
+          op vm `restart_vm
+      | _ ->
+          ()
+    )
+    guidances
+
+let merge_with_pending_guidances ~pending ~coming =
+  let open GuidanceSet in
+  let pending = of_list pending in
+  union pending (of_list coming) |> remove EvacuateHost |> fun unioned ->
```
gangj, line 1358: For mandatory guidances we will not filter guidance will partial order, we will only remove any duplicated ones. Then when will we need to remove some guidance from pending ones?
Is the `to_be_removed` guidances only would be `EvacuateHost`? But I think `EvacuateHost` will not exist in the legacy `host.pending_guidances` neither?

Would you please give an example when there would be `to_be_removed` returned here? Thank you!

ocaml/tests/test_updateinfo.ml
```diff
@@ -1117,11 +1132,295 @@ module UpdateInfoOfXml = Generic.MakeStateless (struct
...
+                <recommended_guidance>EvacuateHost</recommended_guidance>
+                <absolute_guidance>RebootHost</absolute_guidance>
+                <guidance_applicabilities/>
+                <guidance>
+                  <mandatory>
+                    <value>RestartDeviceModel</value>
+                    <value>EvacuateHost</value>
+                    <value>RestartToolstack</value>
+                  </mandatory>
+                </guidance>
```
gangj, line 1266: I think this is also an invalid update? Same as above.

ocaml/tests/test_updateinfo.ml
```diff
@@ -1117,11 +1132,295 @@ module UpdateInfoOfXml = Generic.MakeStateless (struct
...
+                  ; update_type= "security"
+                  ; livepatch_guidance= Some RestartDeviceModel
+                  ; livepatches= []
+                  ; issued= Xapi_stdext_date.Date.epoch
+                  ; severity= Severity.None
+                  }
+              )
+            ]
+        )
+      ; (* guidance in new format: empty mandatory and full *)
```
gangj, line 1198: How about a test case with only new `guidance`, but without old `recommended_guidance`, `absolute_guidance`?

ocaml/tests/test_repository_helpers.ml
```diff
@@ -952,8 +874,9 @@ module EvalGuidanceForOneUpdate = Generic.MakeStateless (struct
                       id= "UPDATE-0001"
                     ; summary= "summary"
                     ; description= "description"
-                    ; rec_guidance= Some Guidance.RestartDeviceModel
+                    ; rec_guidance= Some RestartDeviceModel
                     ; abs_guidance= None
+                    ; guidance= [(Mandatory, [RestartDeviceModel])]
                     ; guidance_applicabilities=
                         [
                           Applicability.
```
gangj, line 882: Some comments seems need update, like
line 888: "old version 0.2.1 is greater than 0.2.0"

ocaml/tests/test_repository_helpers.ml
```diff
@@ -1163,13 +1090,18 @@ module EvalGuidanceForOneUpdate = Generic.MakeStateless (struct
                       id= "UPDATE-0000"
                     ; summary= "summary"
                     ; description= "description"
-                    ; rec_guidance= Some Guidance.RebootHost
+                    ; rec_guidance= Some RebootHost
                     ; abs_guidance= None
+                    ; guidance=
+                        [
+                          (Recommended, [RebootHost])
```
gangj, line 1097: I think for a valid update, the `RebootHost` guidance here should be a `Mandatory` one?

ocaml/tests/test_repository_helpers.ml
```diff
@@ -1163,13 +1090,18 @@ module EvalGuidanceForOneUpdate = Generic.MakeStateless (struct
                       id= "UPDATE-0000"
                     ; summary= "summary"
                     ; description= "description"
-                    ; rec_guidance= Some Guidance.RebootHost
+                    ; rec_guidance= Some RebootHost
```
gangj, line 1093: Do we need the old fields `rec_guidance` and `abs_guidance` for the new update? As we need new client(new version of XC) to talk to xapi.

ocaml/tests/test_repository_helpers.ml
```diff
@@ -2750,52 +2757,720 @@ module ParseUpdateInfoList = Generic.MakeStateless (struct
...
+  end
+
+  let transform (l1, l2) =
+    let open GuidanceSet in
+    reduce (of_list l1) (of_list l2) |> elements
+
+  let tests =
+    let open Guidance in
+    `QuickAndAutoDocumented
+      [
```
gangj, line 2868: I think the test cases exhausted all the combination possibilities, we'd better have a mechanism to ensure no one is missed later.

ocaml/xapi/repository_helpers.ml
```diff
@@ -99,54 +99,35 @@ module GuidanceSet = struct
...
-          else
-            acc
-        )
-        gs precedences
-    in
-    if remove_evacuations then
-      remove EvacuateHost gs'
-    else
-      gs'
+  let resort gs =
```
lindig, line 109: What is this function doing? I understand lists are short but it has quadratic behavior.

ocaml/xapi/repository_helpers.ml
```diff
@@ -99,54 +99,35 @@ module GuidanceSet = struct
...
+          diff acc lowers
+        else
+          acc
+      )
+      gs precedences
+
+  (** remove the guidances which are in s1 from s2. *)
+  let reduce s1 s2 = diff (resort (union s1 s2)) s1
+
+  (** for each set in the list, remove the guidances which are in previous sets from it. *)
```
lindig, line 122: Are we dealing with lists (order is important) or sets (order is not important)? If we are operating on sets I would prefer sets being used.

ocaml/xapi/repository_helpers.ml
```diff
@@ -1361,15 +1324,139 @@ let prune_updateinfo_for_livepatches livepatches updateinfo =
...
+      | _ ->
+          ()
+    )
+    guidances
+
+let merge_with_pending_guidances ~pending ~coming =
+  let open GuidanceSet in
+  let pending = of_list pending in
+  union pending (of_list coming) |> remove EvacuateHost |> fun unioned ->
+  (diff unioned pending |> elements, diff pending unioned |> elements)
```
duobei, line 1359:
```OCaml
let merge_with_pending_guidances ~pending ~coming =
  let open GuidanceSet in
  let pending = of_list pending in
  let unioned = remove EvacuateHost @@ union pending (of_list coming) in
  (diff unioned pending |> elements, diff pending unioned |> elements)
```
The code above is a little more readable.

ocaml/tests/test_repository_helpers.ml
```diff
@@ -2750,52 +2757,720 @@ module ParseUpdateInfoList = Generic.MakeStateless (struct
...
+module GuidanceSetReduceTest = Generic.MakeStateless (struct
+  module Io = struct
+    type input_t = Guidance.t list * Guidance.t list
+
+    type output_t = Guidance.t list
+
+    let string_of_input_t (l1, l2) =
+      Fmt.(str "%a" Dump.(list string)) (List.map Guidance.to_string l1)
+      ^ " + "
+      ^ Fmt.(str "%a" Dump.(list string)) (List.map Guidance.to_string l2)
```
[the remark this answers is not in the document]
lindig: The `^` operation is generally expensive; prefer using `String.concat`; could this not be part of a single format? `"%a + %a`?

ocaml/xapi/updateinfo.mli
```diff
@@ -113,14 +117,28 @@ module Severity : sig
...
+
+(** The metadata of one update in updateinfo. *)
 module UpdateInfo : sig
   type t = {
       id: string
     ; summary: string
     ; description: string
     ; rec_guidance: Guidance.t option
     ; abs_guidance: Guidance.t option
+    ; guidance: GuidanceOfUpdateInfo.t
```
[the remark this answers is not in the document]
minglumlu: The `guidance` (non-countable noun) is a general concept in updateinfo metadata for all tasks to be done in update

ocaml/xapi/repository_helpers.ml
minglumlu: I added a comment for this:
`(** reduce the set [gs] by sorting partially with the defined [precedences] *)`

ocaml/xapi/repository_helpers.ml
minglumlu: In this module, only `set`s are used. The `list` here is for a list of sets.

ocaml/xapi/repository_helpers.ml
minglumlu: Yeah. This is not correct. Fixed it up.

ocaml/xapi/repository.ml
minglumlu: Updated the comment. This is not the first thing :)

ocaml/xapi/repository.ml
minglumlu: As described in commit message, others will be set in next PR.

ocaml/xapi/repository_helpers.ml
minglumlu: Indeed, `to_be_removed` will be empty always in the current implementation.  It is to keep an ability of removing something from the list.

ocaml/xapi/repository_helpers.ml
minglumlu: Yes. But this is a general logic for all kinds of guidance.

ocaml/xapi/repository_helpers.ml
minglumlu: This is a general logic for all kinds of guidance, including `Recommended` and `Full` which will be added in following PR.

ocaml/xapi/xapi_host.ml
minglumlu: Yes.

ocaml/tests/test_updateinfo.ml
```diff
@@ -1117,11 +1132,295 @@ module UpdateInfoOfXml = Generic.MakeStateless (struct
...
+                <special_info>special information</special_info>
+                <url>https://update.details.info</url>
+                <recommended_guidance>EvacuateHost</recommended_guidance>
+                <absolute_guidance>RebootHost</absolute_guidance>
+                <guidance_applicabilities/>
+                <guidance>
+                  <mandatory/>
+                  <full>
+                  </full>
+                </guidance>
```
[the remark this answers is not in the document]
minglumlu: This is to test the guidance in new format will be used in favor.

ocaml/tests/test_updateinfo.ml
minglumlu: This is to test the guidance in new format will be used in favor.

ocaml/tests/test_repository_helpers.ml
minglumlu: The "old version" means the current installed version in an update. Here it means the applicability is not applicable to the update.

ocaml/tests/test_repository_helpers.ml
minglumlu: Yes. It's fine to get rid of them from internal data structure.

ocaml/tests/test_repository_helpers.ml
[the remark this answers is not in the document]
duobei: > String.concat is actually pretty efficient as it computes the length of the result from its argument (a list) and then blits the list items one by one. So there isn’t any allocation of intermediate results.

Cool!

minglumlu:
Squashed the fixups to make reviewers easier.

ocaml/xapi/updateinfo.ml
```diff
@@ -441,18 +449,79 @@ module Severity = struct
...
+          match kind_of_xml kind_xml with
+          | kind ->
+              let values = List.filter_map value_of_xml values_in_xml in
+              (kind, values) :: List.remove_assoc kind acc
+          | exception Unsupported_kind_xml ->
+              warn "Unsupported guidance kind XML tag %s" kind_xml ;
+              acc
+        )
+        | _ ->
+            warn "Ignore unexpected XML node in guidance." ;
```
gangj, line 514: Can we log something about the unexpected node, I think then the log will be meaningful.

ocaml/xapi/updateinfo.mli
```diff
@@ -113,19 +117,30 @@ module Severity : sig
...
+
+(** The metadata of one update in updateinfo. *)
 module UpdateInfo : sig
   type t = {
       id: string
     ; summary: string
     ; description: string
-    ; rec_guidance: Guidance.t option
-    ; abs_guidance: Guidance.t option
+    ; guidance: GuidanceInUpdateInfo.t
```
gangj, line 150: As we discussed, keep the name `guidance` align with the key in updateinfo.xml is ok, it is better to add a comment here that it is not one guidance for quick understanding.

ocaml/xapi/updateinfo.mli
```diff
@@ -113,19 +117,30 @@ module Severity : sig
   val of_string : string -> t
 end
 
-(** The metadata of one update in updateinfo *)
+(** The type of [guidance] in updateinfo metadata. *)
+module GuidanceInUpdateInfo : sig
```
gangj, line 132: add an "s" -> `GuidancesInUpdateInfo` ?

gangj:
> Maybe to update the commit message:
> 
> In the commit message of the first commit: [3b6ea2c](https://github.com/xapi-project/xen-api/commit/3b6ea2cc8f06d9527fac1b37c94de6d2177da64b):
> 
> `XAPI parses the guidance from the data in old format only if there is no data in new format presented in the XML file.`
> 
> From the code, I think all the data including the old guidances are parsed first, but will convert the old ones only if there is no new one.

I think an accurate commit message would be: ("parses" -> "uses")
"XAPI uses the guidance from the data in old format only if there is no
data in new format presented in the XML file."

ocaml/xapi/repository_helpers.ml
```diff
@@ -596,72 +604,77 @@ let eval_guidance_for_one_update ~updates_info ~update ~kind
...
+          GuidanceSet.union s acc
+      | None ->
           acc
     )
-    upd_ids_of_livepatches guidances
+    upd_ids_of_livepatches guidance_tasks
 
-let eval_guidances ~updates_info ~updates ~kind ~livepatches ~failed_livepatches
-    =
+let eval_guidances ~updates_info ~updates ~kind ~livepatches =
```
gangj, line 650: UT to check it works for all `kind`s of guidance would be better.

gangj:
> For the commit message of the 2nd one: [740bbd9](https://github.com/xapi-project/xen-api/commit/740bbd9bc2fbe1e07bdf352137a72f2856c97f99)
> 
> `Additionally, the evaluation is simplified by not considering the failed livapatches any more. This may cause un-necessary carrying out of livepatch guidances. But the overall impact of carrying out these guidances is considered as ignorable. `
> 
> might be good to add:
> 
> 1. how livepatch guidance is used now: livepatch guidance will be used as recommended guidance no matter if the livepatch is applied successfully or failed
> 2. why. Something like; as update guidance will be carried out by xapi client(XenCenter), this will make the client's implementation easy, it doesn't need to check the result of the livepatch to determine what guidances are pending to be applied.

I think it would be better to clarify the why for "the evaluation is simplified by not considering the failed
livapatches any more"

ocaml/xapi/repository_helpers.ml
gangj: Understand, we can keep this logic, but would be better to add a comment to avoid any confusion later.

ocaml/tests/test_updateinfo.ml
gangj: Sorry, I can't see a new test case added. We'd better add it now as there would have no xapi code change when the old updates fields are removed by factory.

ocaml/xapi/updateinfo.mli
minglumlu: I think the type should be good to explain the structure.

ocaml/xapi/updateinfo.mli
minglumlu: I think using `guidance` is consistent with the name of the field.

ocaml/xapi/updateinfo.ml
minglumlu: It may not be worth to log a XML node the structure of which is unknown.

robhoes (approved):

gangj (approved):

PR #5287, opened 2023-12-08
CP-45979 update host-display to python3

acefei:
- format with black
- use xcp.cmd to replace subprocess.popen
Test passed on Xenrt, job id: 3880959

### DIFF
--- a/scripts/host-display
+++ b/scripts/host-display
@@ -8,26 +8,23 @@ n.b. the value returned by the "status" command will be correct as of the next
 boot (assuming that the boot config is not modified further), and is not
 necessarily correct at the time this script is called.
 """
-from __future__ import print_function
 
 import os
-import os.path
 import re
-import subprocess
 import sys
 import syslog
 from xcp.bootloader import Bootloader
 
-COMMAND_NAME = sys.argv[0]
-SHORT_COMMAND_NAME = os.path.basename(COMMAND_NAME)
+SHORT_COMMAND_NAME = Path(__file__).name
 
-XE_SERIAL = 'xe-serial'
-XEN_CMDLINE = '/opt/xensource/libexec/xen-cmdline'
+XE_SERIAL = "xe-serial"
+XEN_CMDLINE = "/opt/xensource/libexec/xen-cmdline"
+
 
 def send_to_syslog(msg):
-    """ Send a message to syslog. """
+    """Send a message to syslog."""
     pid = os.getpid()
-    syslog.syslog("%s[%d] - %s" %(SHORT_COMMAND_NAME, pid, msg))
+    syslog.syslog("%s[%d] - %s" % (SHORT_COMMAND_NAME, pid, msg))
 
 def doexec(args):
     """Execute a subprocess, then return its return code, stdout and stderr"""
@@ -46,8 +43,8 @@ def doexec(args):
 
 def disable(bootloader):
     """
-        Configure the host not to output its console to the physical display on
-        next boot.
+    Configure the host not to output its console to the physical display on
+    next boot.
     """
     if bootloader.default == XE_SERIAL:
         doexec([XEN_CMDLINE, "--set-xen", "console=%s" % (_get_com_port())])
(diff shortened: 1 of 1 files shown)


scripts/host-display
```diff
@@ -91,25 +73,24 @@ def _get_com_port():
     """
     Determine from the current bootloader config which COM port is in use
     """
-    (_, stdout, _) = doexec([XEN_CMDLINE, "--get-xen", "console"])
-    m = re.match(r"console=.*(com\d).*", stdout.readline().strip())
+    (_, stdout) = runCmd([XEN_CMDLINE, "--get-xen", "console"], with_stdout=True)
+    m = re.match(r"console=.*(com\d).*", stdout)
```
psafont, line 77: Why isn't the strip needed here anymore?

acefei: I believe the strip is unnecessary as the regex won't capture the trailing whitespaces.

psafont: Strip also removes leading whitespace, which the regex doesn't take into account. (I don't know if it's relevant)

scripts/host-display
```diff
@@ -8,108 +8,96 @@ n.b. the value returned by the "status" command will be correct as of the next
...
+
 def usage():
-    """ Display a usage string. """
-    command_names = list(COMMANDS.keys())
-    command_names.sort()
+    """Display a usage string."""
+    command_names = sorted(COMMANDS.keys())
     print("Usage:")
-    print("    %s {%s}" % (sys.argv[0], '|'.join(command_names)))
+    print("    {} {{{}}}".format(sys.argv[0], "|".join(command_names)))
```
bernhardkaindl, line 88: The existing version is shorter and more readable to me:
```suggestion
    print("    %s {%s}" % (sys.argv[0], '|'.join(command_names)))
```
Also, please don't mix actually needed cosmetic changes (like removing spaces aboce) with changes needed Python3, this makes it needlessly hard to review the changes.

bernhardkaindl (changes requested):

scripts/host-display
acefei: I tested with this change which works well, but to align with the previous behavior, I will add `strip()`

scripts/host-display
acefei: str.format has more performance, but bad readablity in this case, I will revert this change here.

scripts/host-display
acefei: Hi @psafont , I've updated as your comment, could you review again and approve this PR?

acefei:
updated as comments, job 3888721 is passed.

psafont (approved):
Please squash the last comment into the previous one

scripts/host-display
```diff
@@ -8,108 +8,96 @@ n.b. the value returned by the "status" command will be correct as of the next
...
-        args,
-        stdin=None,
-        stdout=subprocess.PIPE,
-        stderr=subprocess.PIPE,
-        close_fds=True,
-        universal_newlines=True)
-    rc = proc.wait()
-    stdout = proc.stdout
-    stderr = proc.stderr
-    return (rc, stdout, stderr)
```
bernhardkaindl, line 45: By replacing calls to `doexec()` (which calls send_to_syslog) with `xcp.runCmd()` (which does not log the commands to syslog), you remove the logging of the executed commands to the syslog.

Please revert the changes to call `xcp.runCmd()` directly and incorporate `syslog.syslog("%s[%d] - %s" % (SHORT_COMMAND_NAME, os.getpid(), msg))` from the now unused `send_to_syslog()` into the updated `doexec()`.

bernhardkaindl (changes requested):
See my code review comment: Please fix that the new exec function logs like the old did, using the code from the now unused `send_to_syslog()` function,

scripts/host-display
acefei: updated and tested with job id 3889671

psafont (approved):

psafont:
The commits need to be signed off, this can be done by running `git rebase master --signoff`

acefei:
split commits and tested job id 3895260

liulinC:
job 3895260 passed, PR looks good.
Merge this pR.

PR #5288, opened 2023-12-08
CP-45979 update link-vms-by-sr.py to python3

acefei:
- refactor code to put the entire loop into the try-except block for performance
- change print syntax
- use argparse instead of getopt
- remove unused arguments
- rename `dir` variable as it is python buildin variable
- improve code readablity

Test manunally passed.
image

### DIFF
--- a/scripts/link-vms-by-sr.py
+++ b/scripts/link-vms-by-sr.py
@@ -1,41 +1,32 @@
-#!/usr/bin/python
+#!/usr/bin/env python3
 # Populate a directory of symlinks partitioning VMs by SR
 # (c) Anil Madhavapeddy, Citrix Systems Inc, 2008
 
+from __future__ import print_function
+
+import argparse
 import atexit
+import contextlib
+import os
+import sys
+from pathlib import Path
+
 import XenAPI
-import os, sys
-import getopt
 
-def logout():
-    try:
+
+def logout(session):
+    with contextlib.suppress(Exception):
         session.xenapi.session.logout()
-    except:
-        pass
-atexit.register(logout)
 
-def usage():
-    print >> sys.stderr, "%s [-d <directory>]" % sys.argv[0]
-    sys.exit(1)
-   
-def main(argv):
-    session = XenAPI.xapi_local()
-    session.xenapi.login_with_password("", "", "1.0", "xen-api-scripts-linkvmsbysr.py")
 
-    try:
-        opts, args = getopt.getopt(sys.argv[1:], "hd:", [])
-    except getopt.GetoptError, err:
-        print str(err)
-        usage()
+def get_input_dir():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-d", dest="input_dir", required=True, help="Specify the input directory")
+    args = parser.parse_args()
+    return args.input_dir
 
-    dir = None
-    for o,a in opts:
-        if o == "-d":
-            dir = a
 
-    if dir == None:
-        usage()
- 
+def get_vms_in_sr(session):
     vms = session.xenapi.VM.get_all_records()
     vbds = session.xenapi.VBD.get_all_records()
     vdis = session.xenapi.VDI.get_all_records()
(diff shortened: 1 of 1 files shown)


scripts/link-vms-by-sr.py
```diff
@@ -1,41 +1,37 @@
...
 
-def logout():
-    try:
+
+def logout(session):
+    with contextlib.suppress(Exception):
         session.xenapi.session.logout()
-    except:
-        pass
-atexit.register(logout)
```
psafont, line 15: Where is the registration of the logout function happening now?

scripts/link-vms-by-sr.py
```diff
@@ -46,61 +42,64 @@ def main(argv):
...
+            Path(linkdir).mkdir(parents=True)
         except:
-            print >> sys.stderr, "Failed to create directory: %s" % linkdir
-        for vmuuid in vms_in_sr[sruuid].keys():
-            try:
-                src = "../../all/%s.vmmeta" % vmuuid
-                targ = "%s/%s.vmmeta" % (linkdir, vmuuid)
+            print("Failed to create directory: %s" % linkdir, file=sys.stderr)
+
+        try:
```
psafont, line 93: You say this is for performance, but this changes behaviour, now if a symlink for a vm fails to be created can block the creation for the rest of the symlinks, this didn't happen before

acefei: > You say this is for performance, but this changes behaviour, now if a symlink for a vm fails to be created can block the creation for the rest of the symlinks, this didn't happen before

Good catch!
I will revert this change.

scripts/link-vms-by-sr.py
acefei: The origin code is invalid, because the session is never passed in the `logout()` function and the trick is that the exception was catched.
Now I put the registration after session created https://github.com/xapi-project/xen-api/pull/5288/files#diff-12459e8d686a2efa25406c72b7fd866bb61c08c614c5030588fdff5304538499R77

scripts/link-vms-by-sr.py
```diff
@@ -1,41 +1,37 @@
...
-    try:
+
+def logout(session):
+    with contextlib.suppress(Exception):
         session.xenapi.session.logout()
-    except:
-        pass
-atexit.register(logout)
-
-def usage():
```
stephenchengCloud, line 17: I don't see usage in your change. Did you remove it？

scripts/link-vms-by-sr.py
```diff
@@ -1,41 +1,37 @@
...
-    except getopt.GetoptError, err:
-        print str(err)
-        usage()
 
-    dir = None
-    for o,a in opts:
-        if o == "-d":
-            dir = a
+def get_input_dir():
+    parser = argparse.ArgumentParser()
```
liulinC, line 23: ArgumentParser is usually used in some dedicated function for arg parsers. 
So it can be extended if we need some more args. 

But it is good enough for this case.

scripts/link-vms-by-sr.py
liulinC: ArgumentParser take care of it?

scripts/link-vms-by-sr.py
```diff
@@ -1,41 +1,37 @@
...
+    parser.add_argument("-d", dest="input_dir", help="Specify the input directory")
+
+    args = parser.parse_args()
 
-    if dir == None:
-        usage()
- 
+    if not args.input_dir:
```
liulinC, line 28: https://docs.python.org/3/library/argparse.html#required can be used to force the argument, 
why do not you use it? To override the error message?

acefei: image
updated

liulinC:
@acefei  Please squash your commits before merge.

liulinC (approved):

psafont:
This is used exclusively by the script `scripts/xe-backup-metadata` with `link-vms-by-sr.py" -d ${metadir}`.

Have xe-backup-metadata been tested after the change to verify it still works as expected?

psafont:
The makefile should install:
```
        $(IPROG) xe-backup-metadata $(DESTDIR)$(OPTDIR)/bin 
```

And indeed I can find the script in a host:
```
# ls -l /opt/xensource/bin/xe-backup-metadata 
-rwxr-xr-x 1 root root 8025 Dec  3  2021 /opt/xensource/bin/xe-backup-metadata
```

psafont:
Ah, that's fair enough

PR #5292, opened 2023-12-11
CP-45974-port examples folder under /xen-api/scripts/examples to python3.

ashwin9390:
PR contains below changes  to support python3,

1. Shebang has been changed to  #!/usr/bin/env python3
2. Certain modules required update to make it compatible with python3 .
3. Used Generic  exception to fix "Python 3 does not support string exceptions"
4. Fixed extra line issues when making changes.
5. smapiv2 throws "ValueError: must have exactly one of create/read/write/append mode" made necessary changes to the code to make it compatible with python3.

Signed-off-by: Ashwinh <ashwin.h@cloud.com>

### DIFF
--- a/scripts/examples/python/exportimport.py
+++ b/scripts/examples/python/exportimport.py
@@ -20,7 +20,7 @@
 #  - connect an export to an import to copy a raw disk image
 
 from __future__ import print_function
-import sys, os, socket, urllib2, urlparse, XenAPI, traceback, ssl, time
+import sys, os, socket, urllib.request, urllib.error, urllib.parse, XenAPI, traceback, ssl, time
 
 def exportimport(url, xapi, session, src_vdi, dst_vdi):
   # If an HTTP operation fails then it will record the error on the task
@@ -35,12 +35,12 @@ def exportimport(url, xapi, session, src_vdi, dst_vdi):
     put_url = "/import_raw_vdi?session_id=%s&vdi=%s&task_id=%s" % (session, dst_vdi, import_task)
 
     # 'data' is the stream of raw data:
-    data = urllib2.urlopen(url + get_url)
+    data = urllib.request.urlopen(url + get_url)
 
     # python's builtin library doesn't support HTTP PUT very well
     # so we do it manually. Note xapi doesn't support Transfer-encoding:
     # chunked so we must send the data raw.
-    url = urlparse.urlparse(url)
+    url = urllib.parse.urlparse(url)
     host = url.netloc.split(":")[0] # assume port 443
     if url.scheme != "https":
       print("Sorry, this example only supports HTTPS (not HTTP)", file=sys.stderr)
--- a/scripts/examples/python/exportimport.py
+++ b/scripts/examples/python/exportimport.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Copyright (c) 2014 Citrix, Inc.
 #
--- a/scripts/examples/python/mini-xenrt.py
+++ b/scripts/examples/python/mini-xenrt.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Receive multiple VMs
 # Issue parallel loops of: reboot, suspend/resume, migrate
--- a/scripts/examples/python/mini-xenrt.py
+++ b/scripts/examples/python/mini-xenrt.py
@@ -15,9 +15,9 @@ stop = False
 
 class Operation:
     def __init__(self):
-        raise "this is supposed to be abstract, dummy"
+        raise Exception("this is supposed to be abstract, dummy")
     def execute(self, server, session_id):
-        raise "this is supposed to be abstract, dummy"
+        raise Exception("this is supposed to be abstract, dummy")
 
 class Reboot(Operation):
     def __init__(self, vm):
@@ -120,7 +120,7 @@ if __name__ == "__main__":
             allowed_ops = vms[vm]["allowed_operations"]
             for op in [ "clean_reboot", "suspend", "pool_migrate" ]:
                 if op not in allowed_ops:
-                    raise "VM %s is not in a state where it can %s" % (vms[vm]["name_label"], op)
+                    raise Exception("VM %s is not in a state where it can %s" % (vms[vm]["name_label"], op))
             workers.append(Worker(x, session, make_operation_list(vm)))
     for w in workers:
         w.start()
--- a/scripts/examples/python/mini-xenrt.py
+++ b/scripts/examples/python/mini-xenrt.py
@@ -4,7 +4,7 @@
 # Issue parallel loops of: reboot, suspend/resume, migrate
 
 from __future__ import print_function
-import xmlrpclib
+import xmlrpc.client
 from threading import Thread
 import time, sys
 
@@ -109,7 +109,7 @@ if __name__ == "__main__":
         print("  -- performs parallel operations on VMs with the specified other-config key")
         sys.exit(1)
     
-    x = xmlrpclib.Server(sys.argv[1])
+    x = xmlrpc.client.server(sys.argv[1])
     key = sys.argv[2]
     session = x.session.login_with_password("root", "xenroot", "1.0", "xen-api-scripts-minixenrt.py")["Value"]
     vms = x.VM.get_all_records(session)["Value"]
--- a/scripts/examples/python/mini-xenrt.py
+++ b/scripts/examples/python/mini-xenrt.py
@@ -138,5 +138,4 @@ if __name__ == "__main__":
         sys.exit(0)
     else:
         print("FAIL")
-        sys.exit(1)
-        
+        sys.exit(1)  
--- a/scripts/examples/python/provision.py
+++ b/scripts/examples/python/provision.py
@@ -65,7 +65,7 @@ def parseProvisionSpec(txt):
     doc = xml.dom.minidom.parseString(txt)
     all = doc.getElementsByTagName("provision")
     if len(all) != 1:
-        raise "Expected to find exactly one <provision> element"
+        raise Exception("Expected to find exactly one <provision> element")
     ps = ProvisionSpec()
     disks = all[0].getElementsByTagName("disk")
     for disk in disks:
@@ -107,5 +107,5 @@ if __name__ == "__main__":
     txt2 = printProvisionSpec(ps)
     print(txt2)
     if txt != txt2:
-        raise "Sanity-check failed: print(parse(print(x))) <> print(x)"
+        raise Exception("Sanity-check failed: print(parse(print(x))) <> print(x)")
     print("* OK: print(parse(print(x))) == print(x)")
--- a/scripts/examples/python/provision.py
+++ b/scripts/examples/python/provision.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # Copyright (c) 2007 XenSource, Inc.
 #
 # Permission to use, copy, modify, and distribute this software for any
--- a/scripts/examples/python/xva.py
+++ b/scripts/examples/python/xva.py
@@ -3,7 +3,7 @@
 # Rewrite the VDI.sm_config:SCSIid fields in XVA metadata
 
 from __future__ import print_function
-import tarfile, xmlrpclib, optparse, StringIO, sys
+import tarfile, xmlrpc.client, optparse, io, sys
 
 class Object(object):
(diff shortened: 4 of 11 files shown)


scripts/examples/python/mini-xenrt.py
```diff
@@ -15,9 +15,9 @@
 
 class Operation:
     def __init__(self):
-        raise "this is supposed to be abstract, dummy"
+        raise Exception("this is supposed to be abstract, dummy")
```
psafont, line 18:
```suggestion
        raise NotImplementedError
````

File: scripts/examples/python/mini-xenrt.py [hunk 2]
```diff
@@ -15,9 +15,9 @@
...
     def execute(self, server, session_id):
-        raise "this is supposed to be abstract, dummy"
+        raise Exception("this is supposed to be abstract, dummy")
```
psafont, line 20:
```suggestion
        raise NotImplementedError
````

File: scripts/examples/python/provision.py [hunk 3]
```diff
@@ -65,7 +65,7 @@ def parseProvisionSpec(txt):
     doc = xml.dom.minidom.parseString(txt)
     all = doc.getElementsByTagName("provision")
     if len(all) != 1:
-        raise "Expected to find exactly one <provision> element"
+        raise Exception("Expected to find exactly one <provision> element")
```
bernhardkaindl, line 68:
```suggestion
        raise ValueError("Expected to find exactly one <provision> element")
```
raise a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError):
* Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError).

Reference 1: https://docs.python.org/3/library/exceptions.html
Reference 2: https://pylint.readthedocs.io/en/stable/user_guide/messages/warning/broad-exception-raised.html
> Raising exceptions that are too generic force you to catch exceptions generically too. It will force you to use a naked ``except Exception:`` clause. You might then end up catching exceptions other than the ones you expect to catch. This can hide bugs or make it harder to debug programs when unrelated errors are hidden.

Examples:

Problematic code:
  ```py
def small_apple(apple, length):
    if len(apple) < length:
        raise Exception("Apple is too small!")  # [broad-exception-raised]
    print(f"{apple} is proper size.")
  ```
Correct code:
```py
def small_apple(apple, length):
    if len(apple) < length:
        raise ValueError("Apple is too small!")
    print(f"{apple} is proper size.")
```

scripts/examples/python/mini-xenrt.py
```diff
@@ -120,7 +120,7 @@ def make_operation_list(vm):
             allowed_ops = vms[vm]["allowed_operations"]
             for op in [ "clean_reboot", "suspend", "pool_migrate" ]:
                 if op not in allowed_ops:
-                    raise "VM %s is not in a state where it can %s" % (vms[vm]["name_label"], op)
+                    raise Exception("VM %s is not in a state where it can %s" % (vms[vm]["name_label"], op))
```
bernhardkaindl, line 123:
```suggestion
                    raise RuntimeError("VM %s is not in a state where it can %s" % (vms[vm]["name_label"], op))
```
raise a [RuntimeError](https://docs.python.org/3/library/exceptions.html#RuntimeError)
* Raised when an error is detected that doesn’t fall in any of the other categories. The associated value is a string indicating what precisely went wrong.

Having the information which class of exception 
Reference 1: https://docs.python.org/3/library/exceptions.html

Reference 2: https://pylint.readthedocs.io/en/stable/user_guide/messages/warning/broad-exception-raised.html

> Raising exceptions that are too generic force you to catch exceptions generically too. It will force you to use a naked ``except Exception:`` clause. You might then end up catching exceptions other than the ones you expect to catch. This can hide bugs or make it harder to debug programs when unrelated errors are hidden.

scripts/examples/python/xva.py
```diff
@@ -61,7 +61,7 @@ def save(self, fileobj):
         output = tarfile.TarFile(mode='w', fileobj=fileobj)
         tarinfo = tarfile.TarInfo("ova.xml")
         tarinfo.size = len(ova_txt)
-        output.addfile(tarinfo, StringIO.StringIO(ova_txt))
+        output.addfile(tarinfo, io.StringIO(ova_txt))
```
bernhardkaindl, line 64: Requesting a change and a test here, if the script is used (not obsolete):

From my experience with the Python3 changes for https://github.com/xenserver/status-report I am quite certain that using io.StringIO as the IO for Python3's `tarfile` implementation will not work.

In Python2, `str` and `bytes` are the same type, so it works, but AFAIK, Python3 `tarfile` won't be able to work with Pyton3's *Unicode* str for I/O, it still needs the I/O based on bytes. Therefore, `io.BytesIO()` would likely have to be used to arrive at actually working, correctly running code that can be covered by executing a test case in CI.

scripts/examples/smapiv2.py
```diff
@@ -1,19 +1,22 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 from __future__ import print_function
-import os, sys, time, socket, traceback
+import os, sys, io, time, socket, traceback
 
-log_f = os.fdopen(os.dup(sys.stdout.fileno()), "aw")
+log_f = io.open(os.dup(sys.stdout.fileno()), "w")
```
bernhardkaindl, line 6: Opening without the append mode may cause the output file (descriptor in this case) to be truncated. The 1:1 conversion would be to use `"a"` everywhere where `"aw"` had do be used with Python2.

The explanation is here:
https://stackoverflow.com/questions/42824474/python-open-built-in-function-difference-between-modes-aw-and-a-or-aw-and
```suggestion
log_f = io.open(os.dup(sys.stdout.fileno()), "a")
```

scripts/examples/smapiv2.py
```diff
@@ -1,19 +1,22 @@
...
 def reopenlog(log_file):
     global log_f
     if log_f:
         log_f.close()
     if log_file:
-        log_f = open(log_file, "aw")
+        try:
+            log_f = io.open(log_file, "a")
+        except FilenotFoundError:
+            log_f = io.open(log_file, "w")
```
bernhardkaindl, line 17: Opening without the append mode may cause the output file (descriptor in this case) to be truncated. The 1:1 conversion would be to use `"a"` everywhere where `"aw"` had do be used with Python2.

The explanation is here:
https://stackoverflow.com/questions/42824474/python-open-built-in-function-difference-between-modes-aw-and-a-or-aw-and
File creation also works in append mode:
```py
$ rm foo;python3 -c 'import io; f = io.open("foo", "a"); f.write("Hi!\n")';cat foo
Hi!
```
```suggestion
        log_f = io.open(log_file, "a")
```

bernhardkaindl (changes requested):
See my 5 comments, they have the respective descriptions and the changes could be applied with the button labelled "Commit suggestion", but of course pull the changes and compare the result using `diff -ru --exclude=.git directory1 directory2` to review the cumulated changes.

For the change from StringIO to BytesIO, further changes are likely needed because BytesIO does not work with Unicode Python3 `str`. So `.encode()` and/or `.decode()` and/or `b"bytes-content"` might be needed on some places.

Independent of what change are made, if this script is not obsolete (cannot be removed), It would be best to add a testcase which executes the changed code to see that it works.

scripts/examples/python/mini-xenrt.py
ashwin9390: I will change this to reflect  NotimplementedError exception.

scripts/examples/python/mini-xenrt.py
ashwin9390: This will be addressed too.

acefei:
There are some general python code style need to update in these files.

- remove the unnecessary `.keys()`, eg: `for ref in pifs`
- `if pif is not None` instead of `pif != None`
- compare types by using `isinstance()`
- be aware of `map()` which return map object instead of list in python3, suggest to use list/tuple comprehension. eg:`(vdi_info(vi) for vi in vis)` instead of  `map(lambda vi: vdi_info(vi), vis)`

scripts/examples/python/provision.py
ashwin9390: Will change the code to include specific exception instead of generic exception.

scripts/examples/python/mini-xenrt.py
ashwin9390: This code will be addressed too.

scripts/examples/python/xva.py
ashwin9390: I do not have information about xva.py being obsolete .i will  make above changes and test the code.

scripts/examples/smapiv2.py
ashwin9390: @bernhardkaindl with append mode ,we will get illegal seek error.

ashwin9390: @acefei Let me go through the link and do necessary changes.

scripts/examples/python/exportimport.py
```diff
@@ -35,12 +35,12 @@ def exportimport(url, xapi, session, src_vdi, dst_vdi):
...
 
     # 'data' is the stream of raw data:
-    data = urllib2.urlopen(url + get_url)
+    data = urllib.request.urlopen(url + get_url)
 
     # python's builtin library doesn't support HTTP PUT very well
     # so we do it manually. Note xapi doesn't support Transfer-encoding:
     # chunked so we must send the data raw.
-    url = urlparse.urlparse(url)
+    url = urllib.parse.urlparse(url)
```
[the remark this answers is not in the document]
ashwin9390: Acknowledged.

scripts/examples/smapiv2.py
bernhardkaindl: Oops, I overlooked that this is just accessing stdout, not an actual file.
In this case, accessing stdout with "w" just works, as there is no "truncating" of stdout.

On @acefei's comment regarding `with`:

Normally, when opening files, the recommendation to use `with` to open files is sound.

But in these cases, an already opened stdout file descriptor is just reused, it's not a real `open()`.

In the worst case, `with` may close stdout, and if you end up really closing stdout, you terminate your ability to send any output to stdout easily.

So @acefei recommendation/comment only applies to regular open(), not when just accessing a file descriptor that is already opened like here with `.fileno()`.

scripts/examples/python/mini-xenrt.py
ashwin9390: Exception has been corrected ,commit id "862b8140d91f56e9a8e9fa4cceb83a319a5ca4d2"

scripts/examples/python/mini-xenrt.py
ashwin9390: Exception has been corrected ,commit id "862b8140d91f56e9a8e9fa4cceb83a319a5ca4d2"

scripts/examples/python/provision.py
ashwin9390: Done ,included in commit bea2f7cd95615c90e9894efc23974c48301b4b85

scripts/examples/python/xva.py
ashwin9390: After  Edwin Torok and Rob Hoes approval  deleted xva.py under /xen-api/scripts/examples/python/   commit 2a6aaaea534544cc7ebe5dcde4b78cc4bc94a106

scripts/examples/python/mini-xenrt.py
```diff
@@ -15,9 +15,9 @@
 
 class Operation:
     def __init__(self):
-        raise Exception("this is supposed to be abstract, dummy")
+        raise NotImplementedError("this is supposed to be abstract, dummy")
```
psafont, line 18: This exception does not allow arguments, instead apply my precious suggested change

ashwin9390: Few sites had argument ,stuck with it . Removed the argument as suggested commit 1186ee869e001c565ea95bc174f3d8b60d250c01

scripts/examples/smapiv2.py
ashwin9390: @acefei changed from io.open to open ,unnecessary keys were removed , changed result = map(lambda vi: vdi_info(vi), vis) to list result = [vdi_info(vi) for vi in vis] commit db38533188e21b4507994590f24cf2e8440332ce

acefei (approved):

psafont (approved):

GeraldEV (approved):

lindig (approved):

lindig (approved):

PR #5293, opened 2023-12-11
CP-46379 Propagate 'traceparent' from 'storage_smapiv1.ml' to 'sm.ml'

GabrielBuica:
Functions inside 'storage_smapiv1.ml' do not propagate the 'traceparent' and/or do not start spans when called. This change makes sure the propagation is done, slightly refactors the code to remove code duplication, and ensure proper closing of spans when exceptions happen.

### DIFF
--- a/ocaml/xapi-idl/lib/debuginfo.ml
+++ b/ocaml/xapi-idl/lib/debuginfo.ml
@@ -47,3 +47,21 @@ let to_string t =
 
 (* Used for xapi-idl servers that do not yet accept tracing data in the debuginfo *)
 let to_log_string t = t.log
+
+(* Sets the logging context based on `dbg`.
+   Also adds a new tracing span, linked to the parent span from `dbg`, if available. *)
+let with_dbg ?(with_thread = false) ~module_name ~name ~dbg f =
+  let di = of_string dbg in
+  let f_with_trace () =
+    let name =
+      match module_name with "" -> name | _ -> module_name ^ "." ^ name
+    in
+    Tracing.with_tracing ~parent:di.tracing ~name (fun span ->
+        match span with Some _ -> f {di with tracing= span} | None -> f di
+    )
+  in
+  match with_thread with
+  | true ->
+      Debug.with_thread_associated di.log f_with_trace ()
+  | false ->
+      f_with_trace ()
--- a/ocaml/xapi/sm.ml
+++ b/ocaml/xapi/sm.ml
@@ -59,6 +59,8 @@ let driver_filename driver =
 
 (*****************************************************************************)
 
+let with_dbg ~name ~dbg f = Debuginfo.with_dbg ~module_name:"SM" ~name ~dbg f
+
 let debug operation driver msg = debug "SM %s %s %s" driver operation msg
 
 let srmaster_only (_, dconf) =
@@ -132,9 +134,10 @@ let sr_update dconf driver sr =
   let call = Sm_exec.make_call ~sr_ref:sr dconf "sr_update" [] in
   Sm_exec.parse_unit (Sm_exec.exec_xmlrpc (driver_filename driver) call)
 
-let vdi_create dconf driver sr sm_config vdi_type size name_label
+let vdi_create ~dbg dconf driver sr sm_config vdi_type size name_label
     name_description metadata_of_pool is_a_snapshot snapshot_time snapshot_of
     read_only =
+  with_dbg ~dbg ~name:"vdi_create" @@ fun _ ->
   debug "vdi_create" driver
     (sprintf "sr=%s sm_config=[%s] type=[%s] size=%Ld" (Ref.string_of sr)
        (String.concat "; " (List.map (fun (k, v) -> k ^ "=" ^ v) sm_config))
@@ -157,13 +160,15 @@ let vdi_create dconf driver sr sm_config vdi_type size name_label
   in
   Sm_exec.parse_vdi_info (Sm_exec.exec_xmlrpc (driver_filename driver) call)
 
-let vdi_update dconf driver sr vdi =
+let vdi_update ~dbg dconf driver sr vdi =
+  with_dbg ~dbg ~name:"vdi_update" @@ fun _ ->
   debug "vdi_update" driver
     (sprintf "sr=%s vdi=%s" (Ref.string_of sr) (Ref.string_of vdi)) ;
   let call = Sm_exec.make_call ~sr_ref:sr ~vdi_ref:vdi dconf "vdi_update" [] in
   Sm_exec.parse_unit (Sm_exec.exec_xmlrpc (driver_filename driver) call)
 
-let vdi_introduce dconf driver sr new_uuid sm_config location =
+let vdi_introduce ~dbg dconf driver sr new_uuid sm_config location =
+  with_dbg ~dbg ~name:"vdi_introduce" @@ fun _ ->
   debug "vdi_introduce" driver
     (sprintf "sr=%s new_uuid=%s sm_config=[%s] location=%s" (Ref.string_of sr)
        new_uuid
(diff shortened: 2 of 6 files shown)


ocaml/xapi/storage_smapiv1_wrapper.ml
```diff
@@ -92,26 +92,8 @@ let indent x = "    " ^ x
...
-      | Ok span_context ->
-          let result = f {di with tracing= span_context} in
-          let _ = Tracing.Tracer.finish span_context in
-          result
-      | Error e ->
-          D.warn "Failed to start tracing: %s" (Printexc.to_string e) ;
-          f di
-    )
-    ()
+  Debuginfo.with_dbg ~module_name:"SMAPIv1-Wrapper" ~name ~dbg f
```
psafont, line 96: This changes SMAPIv1 to SMAPIv1-Wrapper, is this an intended change? It's not mentioned in the commit message

ocaml/xapi/storage_smapiv1_wrapper.ml
```diff
@@ -603,8 +603,8 @@ functor
         match failures with [] -> next () | f :: _ -> raise f
 
       let epoch_begin context ~dbg ~sr ~vdi ~vm ~persistent =
-        with_dbg ~name:"VDI.epoch_begin" ~dbg @@ fun {log= dbg; _} ->
-        info "VDI.epoch_begin dbg:%s sr:%s vdi:%s vm:%s persistent:%b" dbg
+        with_dbg ~name:"VDI.epoch_begin" ~dbg @@ fun {log; _} ->
```
psafont, line 588: Maybe I'm reading this wrong, but this change seems to contradict the commit message. This looks like a simple rename to me, but on the commit message it state that the type being passed around changes, same with the other functions.

What am I missing?

ocaml/xapi/sm.ml
```diff
@@ -276,23 +302,31 @@ let vdi_compose dconf driver sr vdi1 vdi2 =
...
   debug "vdi_epoch_begin" driver
     (sprintf "sr=%s vdi=%s" (Ref.string_of sr) (Ref.string_of vdi)) ;
   let call =
     Sm_exec.make_call ~sr_ref:sr ~vdi_ref:vdi dconf "vdi_epoch_begin" []
   in
   Sm_exec.parse_unit (Sm_exec.exec_xmlrpc (driver_filename driver) call)
 
-let vdi_epoch_end dconf driver sr vdi =
+let vdi_epoch_end ?traceparent dconf driver sr vdi =
+  Tracing.with_tracing ?parent:(Option.some traceparent) ~name:"SM.epoch_end"
```
psafont, line 317:
```suggestion
  Tracing.with_tracing ?parent:(Option.some traceparent) ~name:"SM.vdi_epoch_end"
```

ocaml/xapi/sm.ml
```diff
@@ -132,9 +132,11 @@ let sr_update dconf driver sr =
   let call = Sm_exec.make_call ~sr_ref:sr dconf "sr_update" [] in
   Sm_exec.parse_unit (Sm_exec.exec_xmlrpc (driver_filename driver) call)
 
-let vdi_create dconf driver sr sm_config vdi_type size name_label
+let vdi_create ?traceparent dconf driver sr sm_config vdi_type size name_label
     name_description metadata_of_pool is_a_snapshot snapshot_time snapshot_of
     read_only =
+  Tracing.with_tracing ?parent:(Option.some traceparent) ~name:"SM.vdi_create"
```
psafont, line 138: Please try to see if this works:
```suggestion
  Tracing.with_tracing ~parent:traceparent ~name:"SM.vdi_create"
```

ocaml/xapi/storage_smapiv1_wrapper.ml
GabrielBuica: This was @robhoes's suggestion to be more clear about the module names. As in the next commit https://github.com/xapi-project/xen-api/pull/5293/commits/b69cc0f38e09407e318b931c546d2d413fe2d071, I instrumented 'storage_smapiv1.ml' and put the 'module_name' as 'SMAPIv1'.

ocaml/xapi/storage_smapiv1_wrapper.ml
GabrielBuica: My understanding was that 'with_dbg' transforms 'dbg' into a 'Debuginfo.t' record `{log:dbg_log; tracing: dbg_tracing}`. Therefore, 'dbg' is shadowed in the scope of 'f' with 'dbg_log' and the 'traceparent' is lost for 'Impl' functions.

ocaml/xapi/sm.ml
GabrielBuica: Thanks, I missed that!

ocaml/xapi/sm.ml
GabrielBuica: I will try it.

ocaml/xapi/sm.ml
GabrielBuica: https://github.com/xapi-project/xen-api/pull/5293/commits/81310de0ec3c3af9a8e00ed239b782c65556ea1b This should fix it.

ocaml/xapi/sm.ml
GabrielBuica: This commit https://github.com/xapi-project/xen-api/pull/5293/commits/81310de0ec3c3af9a8e00ed239b782c65556ea1b should work now. Also, after talking with @mg12, I've made a second commit which use `with_dbg` instead of `with_tracing`.

ocaml/xapi/sm.ml
```diff
@@ -276,23 +290,26 @@ let vdi_compose dconf driver sr vdi1 vdi2 =
...
   debug "vdi_epoch_begin" driver
     (sprintf "sr=%s vdi=%s" (Ref.string_of sr) (Ref.string_of vdi)) ;
   let call =
     Sm_exec.make_call ~sr_ref:sr ~vdi_ref:vdi dconf "vdi_epoch_begin" []
   in
   Sm_exec.parse_unit (Sm_exec.exec_xmlrpc (driver_filename driver) call)
 
-let vdi_epoch_end dconf driver sr vdi =
+let vdi_epoch_end ~dbg dconf driver sr vdi =
+  with_dbg ~dbg ~name:"epoch_end" @@ fun _ ->
```
mg12, line 303: vdi_epoch_end ?

ocaml/xapi/static_vdis.ml
```diff
@@ -77,7 +77,7 @@ let permanent_vdi_deactivate_by_uuid ~__context ~uuid =
   try
     let vdi = Db.VDI.get_by_uuid ~__context ~uuid in
     Sm.call_sm_vdi_functions ~__context ~vdi (fun srconf srtype sr ->
-        Sm.vdi_deactivate srconf srtype sr vdi
+        Sm.vdi_deactivate ~dbg:"" srconf srtype sr vdi
```
mg12, line 80: instead of exposing and passing an empty string, perhaps we could use stronger typing by:
- making ~dbg an optional ?dbg in Debuginfo.with_dbg.
- inside Debuginfo.with_dbg, if ?dbg matches None, then use whatever is equivalent to "" for dbg, else use Some dbg.
- make ~dbg optional ?dbg in Sm.vdi_deactivate only, the other functions in Sm.* still use ~dbg.
- so this line in static_vdis.ml would work without passing dbg to Sm.vdi_deactivate.

GabrielBuica: This makes `dbg` optional inside `Debuginfo.with_dgb` https://github.com/xapi-project/xen-api/pull/5293/commits/47170c82a588e218038284ce036dfdad2b1a597a. And this updates `Sm.vdi_deactivate` https://github.com/xapi-project/xen-api/pull/5293/commits/f2448fa4c67403a74d4f69ea259b8db6ddb5096a

mg12 (approved):

ocaml/xapi-idl/lib/debuginfo.ml
```diff
@@ -47,3 +47,21 @@ let to_string t =
 
 (* Used for xapi-idl servers that do not yet accept tracing data in the debuginfo *)
 let to_log_string t = t.log
+
+(* Sets the logging context based on `dbg`.
+   Also adds a new tracing span, linked to the parent span from `dbg`, if available. *)
+let with_dbg ?(with_thread = false) ?(dbg = "") ~module_name ~name f =
```
robhoes, line 53: Why is `dbg` and optional argument? It seems odd to have a function called `with_dbg` that doesn't require a `dbg`. And the empty string doesn't help.

ocaml/xapi/sm.ml
```diff
@@ -210,15 +219,17 @@ let vdi_activate dconf driver sr vdi writable =
   in
   Sm_exec.parse_unit (Sm_exec.exec_xmlrpc (driver_filename driver) call)
 
-let vdi_deactivate dconf driver sr vdi =
+let vdi_deactivate ?dbg dconf driver sr vdi =
```
robhoes, line 222: Is there a reason for the optional argument here? Is it because of the call in static_vdis.ml? I think we should change _that_ one to add a `~dbg` to the call with

    let dbg = Context.string_of_task_and_tracing __context in

ocaml/xapi-idl/lib/debuginfo.ml
robhoes: Correction... the function is called `with_dbg` because _provides_ a `dbg` to the function. Still, it would be better to make `~dbg` mandatory, since this is the thing that links the spans.

ocaml/xapi/sm.ml
robhoes: I see that Marcus had suggested this in a previous comment. However, I think it is easy to construct a `dbg` as above in static_vdis.ml. This ends up making tracing more complete.

robhoes (approved):
Look good, thanks. Please squash the last fixup and then we can merge it.

PR #5294, opened 2023-12-11
CA-386582: Always create exporting thread for observers

Vincent-lau:
create observer exporting thread regardless of whether we have observers in database.

### DIFF
--- a/ocaml/libs/tracing/tracing.ml
+++ b/ocaml/libs/tracing/tracing.ml
@@ -820,7 +820,7 @@ module Export = struct
       |> List.concat_map (fun x -> TracerProvider.get_endpoints x)
       |> List.iter (export_to_endpoint parent span_list)
 
-    let main () =
+    let create_exporter () =
       enable_span_garbage_collector () ;
       Thread.create
         (fun () ->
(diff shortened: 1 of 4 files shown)


psafont:
I might look obvious, but the reason for the behaviour won't be for people reading the commits in the future. Please explain it in the commit

ocaml/libs/tracing/tracing.ml
```diff
@@ -832,6 +832,21 @@ module Export = struct
           done
         )
         ()
+
+    let exporter_created = ref None
```
lindig, line 836: I think a better name would be `exporter`. The fact that it exists is expressed in the value.

ocaml/libs/tracing/tracing.ml
```diff
@@ -832,6 +832,21 @@ module Export = struct
...
+
+    let lock = Mutex.create ()
+
+    let main () =
+      Xapi_stdext_threads.Threadext.Mutex.execute lock (fun () ->
+          match !exporter_created with
+          | None ->
+              let tid = create_exporter () in
```
lindig, line 844: We are creating an exporter - but assign to a `tid` - which is a thread ID? Naming could be more congruent such that the function creating a value is related to the variable holding that value. We have here `exporter_created`, `create_exporter` and `tid` which refer to related values but you can't tell this from their names.

Vincent-lau: tid is the thread id of the exporter, it is a temporary variable holding the value of that id. It's just easier to have this temporary variable rather than storing the value in exporter directly, which can be a bit awkward since it is a option value

lindig (approved):

psafont (approved):

robhoes (approved):

PR #5295, opened 2023-12-12
Update to python3: hfx_filename, perfmon, static-vdis, xe-scsi-dev-map

stephenchengCloud:
Overview:
- Update to Python3
- Fix some tab-space issues
- rename `list` variable as it is python build-in module
- rename `xmlrpc` varialbe as it is python3 module

Tests:
1. hfx_filename:
    3880043 miscsinglehost.seq passed (one case failed, but it also faied with rt-next, see 3879984)

2. perfmon:
    - manually tested
    [root@lcy2-dt49 bin]# ps -ef | grep perfmon 
    root     32049 17718  0 05:52 pts/1    00:00:00 python3 perfmon
    root     32221 21004  0 05:53 pts/2    00:00:00 grep --color=auto perfmon
    - 3880044 alerts.seq passed

3. static-vdis:
    - 3882705 corosync4hostsHAP1 passed
    - 3882706 Debian passed

4. xe-scsi-dev-map:
    - manually tested
    [root@lcy2-dt49 bin]# python3 xe-scsi-dev-map 

serial number 'W470E22S' maps the following devices 
            /dev/sda6
            /dev/sda5
            /dev/sda4
            /dev/sda3
            /dev/sda2
            /dev/sda1
            /dev/sda
            /dev/disk/by-id/dm-uuid-LVM-8dmUOsdXCl6tof8r3q1FxX4YOFqHDecih3iU9TH7zw6xnbLDAb7Evwq8Tvj9vhZa
            /dev/disk/by-id/dm-name-VG_XenStorage--e73aee11--7462--33fe--176d--3e2aa73ed049-MGT
            /dev/disk/by-id/wwn-0x5000c500a978a868-part2
            /dev/disk/by-id/ata-ST1000NX0423_W470E22S-part2
            /dev/disk/by-id/wwn-0x5000c500a978a868-part4
            /dev/disk/by-id/ata-ST1000NX0423_W470E22S-part4
            /dev/disk/by-id/wwn-0x5000c500a978a868-part3
            /dev/disk/by-id/ata-ST1000NX0423_W470E22S-part3
            /dev/disk/by-id/wwn-0x5000c500a978a868-part6
            /dev/disk/by-id/ata-ST1000NX0423_W470E22S-part6
            /dev/disk/by-id/wwn-0x5000c500a978a868-part5
            /dev/disk/by-id/ata-ST1000NX0423_W470E22S-part5
            /dev/disk/by-id/wwn-0x5000c500a978a868-part1
            /dev/disk/by-id/ata-ST1000NX0423_W470E22S-part1
            /dev/disk/by-id/wwn-0x5000c500a978a868
            /dev/disk/by-id/ata-ST1000NX0423_W470E22S
            /dev/mapper/VG_XenStorage--e73aee11--7462--33fe--176d--3e2aa73ed049-MGT

5. Test with all files:
    191960 Storage: GFS2 Functional
    191961 Storage-Input: karma
    191962 ToolStack-Input
    191963 Ring0 BST
    191964 Storage VDI Encryption Tests

### DIFF
--- a/scripts/perfmon
+++ b/scripts/perfmon
@@ -340,7 +340,7 @@ class RRDUpdates:
     def get_uuid_list_by_objtype(self, objtype):
         "Return a list of uuids corresonding to the objects of this type for which we have ObjectReports"
         return [ objrep.uuid 
-                   for objrep in list(self.report.obj_reports.values()) 
+                   for objrep in self.report.obj_reports.values() 
                      if objrep.objtype == objtype ]
 
 
@@ -1103,18 +1103,18 @@ def main():
             print_debug("sr_uuid_list = %s" % sr_uuid_list)
 
             # Remove monitors for SRs no longer listed in the rrd_updates page
-            for uuid in list(sr_mon_lookup.keys()):
+            for uuid in sr_mon_lookup:
                 if uuid not in sr_uuid_list:
                     sr_mon_lookup.pop(uuid)
             # Create monitors for SRs that have just appeared in rrd_updates page
             for uuid in sr_uuid_list:
-                if uuid not in list(sr_mon_lookup.keys()):
+                if uuid not in sr_mon_lookup:
                     sr_mon_lookup[uuid] = SRMonitor(uuid)
                 else:
                     sr_mon_lookup[uuid].refresh_config()
 
             # Go through each vm_mon and update it using the rrd_udpates - this may generate alarms
-            for vm_mon in list(vm_mon_lookup.values()):
+            for vm_mon in vm_mon_lookup.values():
                 vm_mon.process_rrd_updates(rrd_updates, session)
 
             # Ditto for the host_mon
@@ -1122,7 +1122,7 @@ def main():
                 host_mon.process_rrd_updates(rrd_updates, session)
 
             # And for the sr_mons
-            for sr_mon in list(sr_mon_lookup.values()):
+            for sr_mon in sr_mon_lookup.values():
                 sr_mon.process_rrd_updates(rrd_updates, session)
 
         except socket.error as e:
--- a/scripts/perfmon
+++ b/scripts/perfmon
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 #
 # perfmon - a daemon for monitoring performance of the host on which it is run 
 # and of all the local VMs, and for generating events based on configurable 
@@ -29,13 +29,13 @@
 #
 # The "cf" CGI param specfies the row.  (All rows are returned if it's missing.)
 
-from __future__ import print_function
+
 import sys
 import os
 import getopt
 import traceback
 import XenAPI
-import urllib
+import urllib.request
 from xml import sax            # used to parse rrd_updates because this may be large and sax is more efficient
 from xml.dom import minidom    # used to parse other-config:perfmon.  Efficiency is less important than reliability here
 from xml.parsers.expat import ExpatError
@@ -46,7 +46,7 @@ import syslog
 import socket
 import gc
 import signal
-import commands
+import subprocess
 
 def print_debug(string):
     if debug: 
@@ -118,7 +118,7 @@ class ObjectReport:
     def get_uuid(self):
         return self.uuid
     def get_var_names(self):
-        return self.vars.keys()
+        return list(self.vars.keys())
     def get_value(self, var_name, row):
         try:
             return (self.vars[var_name])[row]
@@ -312,9 +312,7 @@ class RRDUpdates:
         paramstr = "&".join(["%s=%s"  % (k,params[k]) for k in params])
         print_debug("Calling http://localhost/rrd_updates?%s" % paramstr)
 
-        # this is better than urllib.urlopen() as it raises an Exception on http 401 'Unauthorised' error 
-        # rather than drop into interactive mode
-        sock = urllib.URLopener().open("http://localhost/rrd_updates?%s" % paramstr)
+        sock = urllib.request.urlopen("http://localhost/rrd_updates?%s" % paramstr)
         xmlsource = sock.read()
         sock.close()
 
@@ -342,7 +340,7 @@ class RRDUpdates:
     def get_uuid_list_by_objtype(self, objtype):
         "Return a list of uuids corresonding to the objects of this type for which we have ObjectReports"
         return [ objrep.uuid 
-                   for objrep in self.report.obj_reports.values() 
+                   for objrep in list(self.report.obj_reports.values()) 
                      if objrep.objtype == objtype ]
 
 
@@ -354,8 +352,8 @@ def average(mylist):
 
 def get_percent_log_fs_usage(ignored):
     "Get the percent usage of the host filesystem for logs partition. Input list is ignored and should be empty"
-    fs_output = commands.getoutput('df /etc/passwd')
-    log_fs_output = commands.getoutput('df /var/log')
+    fs_output = subprocess.getoutput('df /etc/passwd')
+    log_fs_output = subprocess.getoutput('df /var/log')
     fs_output = ' '.join(fs_output.splitlines()[1:])
     log_fs_output = ' '.join(log_fs_output.splitlines()[1:])
     # Get the percent usage only when there is a separate logs partition
@@ -369,7 +367,7 @@ def get_percent_log_fs_usage(ignored):
 def get_percent_fs_usage(ignored):
     "Get the percent usage of the host filesystem.  Input list is ignored and should be empty"
     # this file is on the filesystem of interest in both OEM and Retail
-    output = commands.getoutput('df /etc/passwd')
+    output = subprocess.getoutput('df /etc/passwd')
     output = ' '.join(output.splitlines()[1:]) # remove header line and rewrap on single line
     percentage = output.split()[4]
     # remove % character and convert to float
@@ -639,10 +637,10 @@ class ObjectMonitor:
         
         for var in self.get_active_variables():
             # find the subset of the params returned for this object that we need to consolidate into var
-            params_to_consolidate = filter(var.rrd_regex.match, params_in_obj_report)
+            params_to_consolidate = list(filter(var.rrd_regex.match, params_in_obj_report))
             for row in range(num_rows):
                 # Get the values to consolidate
-                values_to_consolidate = map(lambda param: obj_report.get_value(param, row), params_to_consolidate)
+                values_to_consolidate = [obj_report.get_value(param, row) for param in params_to_consolidate]
                 # Consolidate them
                 value = var.consolidation_fn(values_to_consolidate)
                 # Pass result on to the variable object - this may result in an alarm being generated
@@ -941,7 +939,7 @@ def update_all_xmlconfigs(session):
     for recs in (all_host_recs, all_vm_recs, all_sr_recs):
         all_otherconfigs.update([
                 (recs[ref]['uuid'], recs[ref]['other_config'])
-                for ref in recs.keys()
+                for ref in recs
                 ])
 
     # rebuild dictionary mapping uuids to xmlconfigs
@@ -1069,13 +1067,13 @@ def main():
             vm_uuid_list = rrd_updates.get_uuid_list_by_objtype('vm')
 
             # Remove any monitors for VMs no longer listed in rrd_updates page
-            for uuid in vm_mon_lookup.keys():
+            for uuid in vm_mon_lookup:
                 if uuid not in vm_uuid_list:
                     vm_mon_lookup.pop(uuid)
                 
             # Create monitors for VMs that have just appeared in rrd_updates page
             for uuid in vm_uuid_list:
-                if uuid not in vm_mon_lookup.keys():
+                if uuid not in vm_mon_lookup:
                     vm_mon_lookup[uuid] = VMMonitor(uuid)
                 else:
                     # check if the config has changed, e.g. by XenCenter
@@ -1105,18 +1103,18 @@ def main():
             print_debug("sr_uuid_list = %s" % sr_uuid_list)
 
             # Remove monitors for SRs no longer listed in the rrd_updates page
-            for uuid in sr_mon_lookup.keys():
+            for uuid in list(sr_mon_lookup.keys()):
                 if uuid not in sr_uuid_list:
                     sr_mon_lookup.pop(uuid)
             # Create monitors for SRs that have just appeared in rrd_updates page
             for uuid in sr_uuid_list:
-                if uuid not in sr_mon_lookup.keys():
+                if uuid not in list(sr_mon_lookup.keys()):
                     sr_mon_lookup[uuid] = SRMonitor(uuid)
                 else:
                     sr_mon_lookup[uuid].refresh_config()
 
             # Go through each vm_mon and update it using the rrd_udpates - this may generate alarms
-            for vm_mon in vm_mon_lookup.values():
+            for vm_mon in list(vm_mon_lookup.values()):
                 vm_mon.process_rrd_updates(rrd_updates, session)
 
             # Ditto for the host_mon
@@ -1124,7 +1122,7 @@ def main():
                 host_mon.process_rrd_updates(rrd_updates, session)
 
             # And for the sr_mons
-            for sr_mon in sr_mon_lookup.values():
+            for sr_mon in list(sr_mon_lookup.values()):
                 sr_mon.process_rrd_updates(rrd_updates, session)
 
         except socket.error as e:
--- a/scripts/static-vdis
+++ b/scripts/static-vdis
@@ -153,7 +153,7 @@ def add(session, vdi_uuid, reason):
 
     sm = None
     all_sm = session.xenapi.SM.get_all_records()
-    for sm_ref in list(all_sm.keys()):
+    for sm_ref in all_sm:
         if all_sm[sm_ref]['type'] == ty:
             sm = all_sm[sm_ref]
             break
@@ -189,7 +189,7 @@ def add(session, vdi_uuid, reason):
     fresh = fresh_name()
     path = main_dir + "/" + fresh
     os.mkdir(path)
-    for key in list(data.keys()):
+    for key in data:
         write_whole_file(path + "/" + key, data[key])
 
 def delete(vdi_uuid):
--- a/scripts/static-vdis
+++ b/scripts/static-vdis
@@ -1,12 +1,12 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Common functions for managing statically-attached (ie onboot, without xapi) VDIs
 
-from __future__ import print_function
-import sys, os, subprocess, json, urlparse, itertools
+
+import sys, os, subprocess, json, urllib.parse
 import os.path
 import time
-import XenAPI, inventory, xmlrpclib
+import XenAPI, inventory, xmlrpc.client
 
 main_dir = "/etc/xensource/static-vdis"
 
@@ -22,27 +22,21 @@ def call_volume_plugin(name, command, args):
         cmd_args.extend(args)
     # on Python >= 3.3 a timout can be set, not on 2.7
     # when porting please add a timeout
-    output = subprocess.check_output(cmd_args)
+    output = subprocess.check_output(cmd_args, universal_newlines=True)
     return json.loads(output)
 
 def call_datapath_plugin(name, command, args):
     args = [ xapi_storage_script + "/datapath/" + name + "/" + command, "static-vdis" ] + args
-    output = subprocess.check_output(args)
+    output = subprocess.check_output(args, universal_newlines=True)
     return json.loads(output)
-
+        
 def read_whole_file(filename):
-    f = open(filename)
-    try:
-        return reduce(lambda x, y: x + y.decode('utf8'), f.readlines(), "").strip()
-    finally:
-        f.close()
-
+    with open(filename, 'r', encoding='utf-8') as f:
+        return ''.join(f.readlines()).strip()
+        
 def write_whole_file(filename, contents):
-    f = open(filename, "w")
-    try:
-        f.write(contents.encode('utf8'))
-    finally:
-        f.close()
+    with open(filename, "w", encoding='utf-8') as f:
+        f.write(contents)
 
 def load(name):
     """Return a dictionary describing a single static VDI"""
@@ -70,7 +64,7 @@ def wait_for_corosync_quorum():
     cmd_args = ['xcli', 'diagnostics', '--json', 'static-vdis']
     quorate = False
     while not quorate:
-        output = subprocess.check_output(cmd_args)
+        output = subprocess.check_output(cmd_args, universal_newlines=True)
         output_map = json.loads(output)
         quorate = output_map.get('is_quorate')
         if not quorate:
@@ -85,17 +79,17 @@ def check_clusterstack(ty):
 def sr_attach(ty, device_config):
     check_clusterstack(ty)
 
-    args = [arg for (k,v) in device_config.iteritems()
+    args = [arg for (k,v) in device_config.items()
             for arg in ["--configuration", k, v]]
     return call_volume_plugin(ty, "SR.attach", args)
 
-def list():
+def list_vdis():
     all = []
     try:
         all = os.listdir(main_dir)
     except:
         pass
-    return map(load, all)
+    return list(map(load, all))
 
 def fresh_name():
     all = []
@@ -131,7 +125,7 @@ def to_string_list(d):
     return l
 
 def add(session, vdi_uuid, reason):
-    for existing in list():
+    for existing in list_vdis():
         if existing['vdi-uuid'] == vdi_uuid:
             if existing['delete-next-boot'] == "True":
                 # Undo the 'delete-next-boot' flag to reinstitute
@@ -159,7 +153,7 @@ def add(session, vdi_uuid, reason):
 
     sm = None
     all_sm = session.xenapi.SM.get_all_records()
-    for sm_ref in all_sm.keys():
+    for sm_ref in list(all_sm.keys()):
         if all_sm[sm_ref]['type'] == ty:
             sm = all_sm[sm_ref]
             break
@@ -195,12 +189,12 @@ def add(session, vdi_uuid, reason):
     fresh = fresh_name()
     path = main_dir + "/" + fresh
     os.mkdir(path)
-    for key in data.keys():
+    for key in list(data.keys()):
         write_whole_file(path + "/" + key, data[key])
 
 def delete(vdi_uuid):
     found = False
-    for existing in list():
+    for existing in list_vdis():
         if existing['vdi-uuid'] == vdi_uuid:
             found = True
             path = main_dir + "/" + existing['id']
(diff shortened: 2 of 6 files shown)


scripts/perfmon
```diff
@@ -354,8 +352,8 @@ def average(mylist):
 
 def get_percent_log_fs_usage(ignored):
     "Get the percent usage of the host filesystem for logs partition. Input list is ignored and should be empty"
-    fs_output = commands.getoutput('df /etc/passwd')
-    log_fs_output = commands.getoutput('df /var/log')
+    fs_output = subprocess.getoutput('df /etc/passwd')
```
liulinC, line 355: do you expect string or byte result?

stephenchengCloud: Both the original code in python2 and the modified code in python3 returns a string
https://www.programcreek.com/python/example/599/commands.getoutput
https://docs.python.org/3/library/subprocess.html

scripts/static-vdis
```diff
@@ -22,27 +22,21 @@ def call_volume_plugin(name, command, args):
...
+        
 def read_whole_file(filename):
-    f = open(filename)
-    try:
-        return reduce(lambda x, y: x + y.decode('utf8'), f.readlines(), "").strip()
-    finally:
-        f.close()
-
+    with open(filename, 'r', encoding='utf-8') as f:
+        return ''.join(map(str.strip, f.readlines()))
```
bernhardkaindl, line 35: There are 2 bugs - and here is the exhibit which shows it:
```py
cat >test-pr5295.py <<END
import sys

if sys.version_info > (3,0):
    def read_whole_file(filename):
        with open(filename, 'r', encoding='utf-8') as f:
            return ''.join(map(str.strip, f.readlines()))
            #return ''.join(f.readlines()).strip()
else:
    def read_whole_file(filename):
        f = open(filename)
        try:
            return reduce(lambda x, y: x + y.decode('utf8'), f.readlines(), "").strip()
        finally:
            f.close()

with open(__file__ + "%s.out" % sys.version_info.major, "w") as o:
    o.write(read_whole_file(__file__))
END
python2 test-pr5295.py
python3 test-pr5295.py
diff -u test-pr5295.py[23].out
md5sum test-pr5295.py[23].out
```
For best results, paste it into and editor and run it, see below:

Output:
```py
bash test-pr5295.sh
--- test-pr5295.py2.out 2023-12-12 20:32:42.273484598 +0100
+++ test-pr5295.py3.out 2023-12-12 20:32:42.283484596 +0100
@@ -1,17 +1 @@
-import sys
-
-if sys.version_info > (3,0):
-    def read_whole_file(filename):
-        with open(filename, 'r', encoding='utf-8') as f:
-            return ''.join(map(str.strip, f.readlines()))
-            #return ''.join(f.readlines()).strip()
-else:
-    def read_whole_file(filename):
-        f = open(filename)
-        try:
-            return reduce(lambda x, y: x + y.decode('utf8'), f.readlines(), "").strip()
-        finally:
-            f.close()
-
-with open(__file__ + "%s.out" % sys.version_info.major, "w") as o:
-    o.write(read_whole_file(__file__))
\ No newline at end of file
+import sysif sys.version_info > (3,0):def read_whole_file(filename):with open(filename, 'r', encoding='utf-8') as f:return ''.join(map(str.strip, f.readlines()))#return ''.join(f.readlines()).strip()else:def read_whole_file(filename):f = open(filename)try:return reduce(lambda x, y: x + y.decode('utf8'), f.readlines(), "").strip()finally:f.close()with open(__file__ + "%s.out" % sys.version_info.major, "w") as o:o.write(read_whole_file(__file__))
\ No newline at end of file
f54a639ff4bcd8c662bc35be4def7493  test-pr5295.py2.out
aafbc3b6bce179b66d14cbc75d1b8c12  test-pr5295.py3.out
```
Expected: No output and identical md5sum for both:
```py
bash test-pr5295.sh
c389102aa49e776235c234768286658c  test-pr5295.py2.out
c389102aa49e776235c234768286658c  test-pr5295.py3.out
```
I get this when moving the hash '#' to the other line in the new function of the sample test.

```suggestion
    with open(filename, 'r', encoding='utf-8') as f:
        return ''.join(f.readlines()).strip()
```
Your change removes the newlines and all leading and trailing space from each line, but the original code does not do that as you see: strip() only used once at the end, once, not for each line!

To make sure this is not overlooked, I'm requesting this fix as a requested change, and I'll be on PTO until January, so I might not be able to dismiss the status of requested change until January 2.

This would have nearly escaped my review, and it would be much easier to deal with and find such issues if you would not put everything into one large pull request. In the future, make small PRs, one for each file, for example.

bernhardkaindl (changes requested):
As said in my code review comment: The change in behaviour of `read_whole_file` would have nearly escaped my review.

To make sure this is not overlooked, I'm requesting this fix as a requested change, and I'll be on PTO until January, so I might not be able to dismiss the status of requested change until January 2.

This would have nearly escaped my review, and it would be much easier to deal with and find such issues if you would not put everything into one large pull request. In the future, make small PRs, one for each file, for example.

JFYI: @psafont @lindig @liulinC

liulinC:
> As said in my code review comment: The change in behaviour of `read_whole_file` would have nearly escaped my review.
> 
> To make sure this is not overlooked, I'm requesting this fix as a requested change, and I'll be on PTO until January, so I might not be able to dismiss the status of requested change until January 2.
> 
> This would have nearly escaped my review, and it would be much easier to deal with and find such issues if you would not put everything into one large pull request. In the future, make small PRs, one for each file, for example.
> 
> JFYI: @psafont @lindig @liulinC

Small PRs are always preferred for reviewing and merging. 
I have no objections for this point, even separated PRs means more testing.

scripts/static-vdis
stephenchengCloud: Thank you Bernhard for the review and test.
Code updated.
Totally agree on small PRs for better review.

scripts/static-vdis
bernhardkaindl, line 1: Just a correction on the commit message of my fix resulting from my review of `read_whole_file`:

> Update scripts/static-vdis to keep the same logic as the original code

The commit message suggests that just the change in the logic escaped the testing which was done on this PR.

```py
def read_whole_file(filename):
    with open(filename, 'r', encoding='utf-8') as f:
-        return ''.join(map(str.strip, f.readlines()))
+        return ''.join(f.readlines()).strip()
```
        
In fact, it was the returned data that was changed. Talking about just the logic suggests that just some internal way of how the implementation works was changed (which was not the only effect of the initial change).

Instead, the actually returned data was changed, which has an effect beyond the internals of the function.

A change of logic could still produce the same output.

So speaking of "keep the same logic" is a misrepresentation of the change.
A correct description is:

> "Fix the converted code to not strip all leading and trailing whitespace and all newlines".

In comparison, the current comment to keep the same logic can be read like an optional change, while in fact, it is a fix to preserve the existing functionality.

The distinction between these different meanings is important to me because the changed output was not detected during testing.

Therefore, I'd like to make the point that a unit test would be due in such cases where we have to rewrite a function.

scripts/perfmon
```diff
@@ -938,20 +936,20 @@ def update_all_xmlconfigs(session):
     for recs in (all_host_recs, all_vm_recs, all_sr_recs):
         all_otherconfigs.update([
                 (recs[ref]['uuid'], recs[ref]['other_config'])
-                for ref in recs.keys()
+                for ref in list(recs.keys())
```
bernhardkaindl, line 939: It appears to be sufficient to write it this way:
```suggestion
                for ref in recs
```

Example code:
```py
recs = {"key1": {"attr": "item1"}, "key2": {"attr":"item2"}}
print("Output:", [(ref, recs[ref]["attr"]) for ref in recs])
Output: [('key1', 'item1'), ('key2', 'item2')]
```

You find a good explanation and many conformations here:
https://stackoverflow.com/questions/27476079/why-does-2to3-change-mydict-keys-to-listmydict-keys
Quote:
> In Python 3, the .keys() method returns a view object rather than a list, for efficiency's sake.
>
> In the iteration case, this doesn't actually matter, but where it would matter is if you were doing something like foo.keys()[0] - you can't index a view. Thus, 2to3 always adds an explicit list conversion to make sure that any potential indexing doesn't break.
>
> You can manually remove the list() call anywhere that a view would work fine; 2to3 just isn't smart enough to tell which case is which.

The same change has been requested by `rosslagerwall` in other reviews and other I sat another pull request which directly changed the code this way.

This applies to all uses of `for key in dict.keys()` in this PR.

In my testing, I also saw that the same applies to all uses of `for key in dict.items()` and `for key in dict.values()` where these forms appear to be sufficient.

It's just that 2to3 does not check if `dict.keys()` / `.items()` / `.values()` is used as `....()[0]`, and only in the latter case, the view needs to be converted into a list before the `[0]` or other direct indexing, not for the `for` loop.

scripts/perfmon
```diff
@@ -938,20 +936,20 @@ def update_all_xmlconfigs(session):
...
                 ])
 
     # rebuild dictionary mapping uuids to xmlconfigs
     all_xmlconfigs.clear()
     all_xmlconfigs.update([
             (uuid, other_config['perfmon'])
-            for (uuid, other_config) in all_otherconfigs.items()
+            for (uuid, other_config) in list(all_otherconfigs.items())
```
bernhardkaindl, line 946:
```suggestion
            for (uuid, other_config) in all_otherconfigs.items()
```
Looks similar here, except of course you can't remove `.items()` itself:
```py
recs = {"key1": {"attr": "item1"}, "key2": {"attr":"item2"}}
print("Output:", [(key, value["attr"]) for key, value in recs.items()])
Output: [('key1', 'item1'), ('key2', 'item2')]
```

scripts/perfmon
```diff
@@ -938,20 +936,20 @@ def update_all_xmlconfigs(session):
...
             if 'perfmon' in other_config
             ])
 
     # Rebuild another map
     sruuids_by_hostuuid.clear()
-    for (sr, rec) in all_sr_recs.items():
+    for (sr, rec) in list(all_sr_recs.items()):
```
bernhardkaindl, line 952:
```suggestion
    for (sr, rec) in all_sr_recs.items():
```
ditto, AFAICS.

scripts/perfmon
```diff
@@ -1066,13 +1064,13 @@ def main():
             vm_uuid_list = rrd_updates.get_uuid_list_by_objtype('vm')
 
             # Remove any monitors for VMs no longer listed in rrd_updates page
-            for uuid in vm_mon_lookup.keys():
+            for uuid in list(vm_mon_lookup.keys()):
```
bernhardkaindl, line 1067:
```suggestion
            for uuid in vm_mon_lookup:
```
ditto, AFIACS

scripts/perfmon
```diff
@@ -1066,13 +1064,13 @@ def main():
...
                 if uuid not in vm_uuid_list:
                     vm_mon_lookup.pop(uuid)
                 
             # Create monitors for VMs that have just appeared in rrd_updates page
             for uuid in vm_uuid_list:
-                if uuid not in vm_mon_lookup.keys():
+                if uuid not in list(vm_mon_lookup.keys()):
```
bernhardkaindl, line 1073:
```suggestion
                if uuid not in vm_mon_lookup:
```
And so on for all `for uuid in list(some_dict.keys())` below.

bernhardkaindl (changes requested):
AFAICS, foreach:
```py
-   for uuid in list(somedict.keys()):
+   for uuid in somedict:
```

bernhardkaindl:
> Small PRs are always preferred for reviewing and merging. I have no objections for this point, even separated PRs means more testing.
> 
> Edit: The request to split make sense for reviewers. However, considering the big test effort, (Refer to the `Tests` in @stephenchengCloud), Can I suggest to split the files into commits (as current status)? Then reviewers can review commit by commit, Thus keep each review small and do not explode the Testing. @bernhardkaindl

Of course, repeated manual testing would be bad, but the fact that the change in output from the rewrite of `def read_whole_file()` escaped testing means we should learn from it and improve the process.

Instead of repeated manual testing, adding unit tests when functions are rewritten would be preferred. I already provided the core of a simple test for this function. It would have been easy and not time-consuming at all to add a unit test for it. The strategy would have been:

1. PR: Add unit tests to GitHub wherever a larger change is needed. Ensure that GitHub CI fails when the unit test fails.
   **Important: Get the PR reviewed and merged to verify that the unit test is correct and tests the exitsing code, not the future change**
2. PR: Update the code. The unit test result will immediately (already on the git push) but also in the PR itself whether the change was correct.

liulinC:
> > Small PRs are always preferred for reviewing and merging. I have no objections for this point, even separated PRs means more testing.
> > Edit: The request to split make sense for reviewers. However, considering the big test effort, (Refer to the `Tests` in @stephenchengCloud), Can I suggest to split the files into commits (as current status)? Then reviewers can review commit by commit, Thus keep each review small and do not explode the Testing. @bernhardkaindl
> 
> Of course, repeated manual testing would be bad, but the fact that the change in output from the rewrite of `def read_whole_file()` escaped testing means we should learn from it and improve the process.
> 
> Instead of repeated manual testing, adding unit tests when functions are rewritten would be preferred. I already provided the core of a simple test for this function. It would have been easy and not time-consuming at all to add a unit test for it. The strategy would have been:
> 
> 1. PR: Add unit tests to GitHub wherever a larger change is needed. Ensure that GitHub CI fails when the unit test fails.
>    **Important: Get the PR reviewed and merged to verify that the unit test is correct and tests the exitsing code, not the future change**
> 2. PR: Update the code. The unit test result will immediately (already on the git push) but also in the PR itself whether the change was correct.
> 3. PR: When the code is ready to be parsed by Python3 at some point, enable the unit test to be run
>    under Python3.
> 
> To import python scripts for unit testing individual functions in it, you can use this pytest fixture: https://github.com/xenserver/status-report/blob/master/tests/unit/conftest.py
> 
> 1. Simply change the references to `bugtool` in it to the other script(s) that you want to test
> 2. place it into the tests subdirectory where your unit tests are.
> 3. request pytest to pass the fixture as an argument to the `test_` function like here:
>    https://github.com/xenserver/status-report/blob/master/tests/unit/test_xapidb_filter.py
> 4. Then, you can access all functions and variables in the script using the name of the passed fixture like `fixture_name.function_name()`

UT is good and preferred, 
However, is should be a separated task,  and should not block XS9.

@stephenchengCloud  Please estimate the effort for adding UT, 
If you can finish it quickly, then I am okay with it.

scripts/static-vdis
stephenchengCloud: Fully understood. Thank you for the detailed explanation. 
I'll try to add unit tests if necessary.

scripts/perfmon
stephenchengCloud: Got it. I'll check each change by 2to3.
Thank you so much!

scripts/static-vdis
stephenchengCloud: Will change the commit message too.

stephenchengCloud:
> > > Small PRs are always preferred for reviewing and merging. I have no objections for this point, even separated PRs means more testing.
> > > Edit: The request to split make sense for reviewers. However, considering the big test effort, (Refer to the `Tests` in @stephenchengCloud), Can I suggest to split the files into commits (as current status)? Then reviewers can review commit by commit, Thus keep each review small and do not explode the Testing. @bernhardkaindl
> > 
> > 
> > Of course, repeated manual testing would be bad, but the fact that the change in output from the rewrite of `def read_whole_file()` escaped testing means we should learn from it and improve the process.
> > Instead of repeated manual testing, adding unit tests when functions are rewritten would be preferred. I already provided the core of a simple test for this function. It would have been easy and not time-consuming at all to add a unit test for it. The strategy would have been:
> > 
> > 1. PR: Add unit tests to GitHub wherever a larger change is needed. Ensure that GitHub CI fails when the unit test fails.
> >    **Important: Get the PR reviewed and merged to verify that the unit test is correct and tests the exitsing code, not the future change**
> > 2. PR: Update the code. The unit test result will immediately (already on the git push) but also in the PR itself whether the change was correct.
> > 3. PR: When the code is ready to be parsed by Python3 at some point, enable the unit test to be run
> >    under Python3.
> > 
> > To import python scripts for unit testing individual functions in it, you can use this pytest fixture: https://github.com/xenserver/status-report/blob/master/tests/unit/conftest.py
> > 
> > 1. Simply change the references to `bugtool` in it to the other script(s) that you want to test
> > 2. place it into the tests subdirectory where your unit tests are.
> > 3. request pytest to pass the fixture as an argument to the `test_` function like here:
> >    https://github.com/xenserver/status-report/blob/master/tests/unit/test_xapidb_filter.py
> > 4. Then, you can access all functions and variables in the script using the name of the passed fixture like `fixture_name.function_name()`
> 
> UT is good and preferred, However, is should be a separated task, and should not block XS9.
> 
> @stephenchengCloud Please estimate the effort for adding UT, If you can finish it quickly, then I am okay with it.

After checking the changes I made, I'm afraid right now I have no effort to add UT:
- It's good to have unit tests, but I think they should be done by the author of the function.
- Most of the changes I made in `perfmon` file are in the main() function which has around 300 lines. It would take a lot of time to figure out the logic of the main() function, the inputs and outputs.
- In the file `xe-scsi-dev-map`, similarly, most of the code is in the main() function, and I have no idea what the expected output is.
- Instead, like what @bernhardkaindl has done, I can write small testing to locally test the compatibility between python2 and python3 In response to my changes
- If I have time and effort to spare, I'll consider adding unit tests, but currently, I have other tasks.

psafont:
I think there are some misconceptions here about developing and maintaining software that need to be straightened out:
1. Maintainers have the responsibility to not let regressions be merged. Until maintainers have confidence that the changes won't introduce regressions, these will not get approved nor merged.
2. Because much of the code was developed years ago by people that are no longer active in the project, maintainers can't just ask them to write unit-tests for functions they wrote. Instead they rely on the people changing the functions to add them to prove they are not changing behaviour, or even when they are fixing bugs to test is doesn't regress. These people include [themselves](https://github.com/xapi-project/xcp-rrd/commit/2d303ab82077d42a0d37ea2ffd22c2fef051cd26)
3. Unfortunately because of python being interpreted, there's generally low confidence that changes are generally correct. To add to this, handing unicode and process output are specially easy to get wrong.
4. To raise confidence, testing is needed; and the less costly way (IMO) is to add tests is through unit-test, you are however free to make other kinds of testing to convince maintainers.
5. Churning code fast without doing due process to prove the code is not reliable is something that happens when engineers give in to managers deadlines and demands. The bugs introduced cost orders of magnitude to fix that the ones that are never introduced. Do not fall into this trap by saying unit-tests can be postponed. We both know that if we don't have time to add tests now we'll never have the time because managers are always asking for new features.

Other maintainers will have their own opinion, but as things stand I can't approve this PR in good conscience before unit tests appear.

edwintorok:
I agree with what @psafont said. In addition because Python3 is a language that doesn't have full static type checking  it is difficult to gain confidence in large changes regarding Python. This is a lot easier when a compiled language (like OCaml is used) where the compiler ensures a lot of properties and proves some errors cannot happen at runtime.

With Python you need to either:
* have almost full coverage via unit tests to ensure each line is even syntactically correct (e.g. there could be error paths that are completely broken and would fail with a syntax or type error at runtime). Note that this means that the code you are changing is already unreliable and broken in certain corner cases.
* use mypy or other static type checking (although they do not fully guarantee that the types are correct, it falls back to 'any' often)
* use a statically type checked language instead (like OCaml), although that would be a larger task

edwintorok:
If writing unit tests would be too tedious I'd suggest:
* add some minimal tests in the makefile to ensure that python can at least import the module (see the makefile rules that got added in xsconsole, see e.g. the precommit checks here https://github.com/xapi-project/xsconsole/blob/master/.pre-commit-config.yaml)
* add a mypy rule in CI and start adding some minimal mypy type annotations (this can be done on the Python2 code first by using the comment syntax for types). You don't have to fully annotate everything, just enough to get mypy working.
* then do the python3 update and see whether there are any new mypy failures

stephenchengCloud:
Remove unnecessary "list()" added by 2to3.

- Manually tested  perfmon:
image

- Test in XenRT:

3888759 alerts for perfmon
3888761 corosync4hostsHAP1 for static_vdis

stephenchengCloud:
Going to add unit tests.

stephenchengCloud:
@psafont @bernhardkaindl @edwintorok @liulinC 
Based on the suggestion, I've added some unit tests.
Besides these unit tests, I have already done the XenRT test and manual test:

> Overview:

Update to Python3
Fix some tab-space issues
rename list variable as it is python build-in module
rename xmlrpc varialbe as it is python3 module
Tests:

hfx_filename:
3880043 miscsinglehost.seq passed (one case failed, but it also faied with rt-next, see 3879984)

perfmon:

manually tested
[root@lcy2-dt49 bin]# ps -ef | grep perfmon
root 32049 17718 0 05:52 pts/1 00:00:00 python3 perfmon
root 32221 21004 0 05:53 pts/2 00:00:00 grep --color=auto perfmon
3880044 alerts.seq passed
static-vdis:

3882705 corosync4hostsHAP1 passed
3882706 Debian passed
xe-scsi-dev-map:

manually tested
[root@lcy2-dt49 bin]# python3 xe-scsi-dev-map
serial number 'W470E22S' maps the following devices
/dev/sda6
/dev/sda5
/dev/sda4
/dev/sda3
/dev/sda2
/dev/sda1
/dev/sda
/dev/disk/by-id/dm-uuid-LVM-8dmUOsdXCl6tof8r3q1FxX4YOFqHDecih3iU9TH7zw6xnbLDAb7Evwq8Tvj9vhZa
/dev/disk/by-id/dm-name-VG_XenStorage--e73aee11--7462--33fe--176d--3e2aa73ed049-MGT
/dev/disk/by-id/wwn-0x5000c500a978a868-part2
/dev/disk/by-id/ata-ST1000NX0423_W470E22S-part2
/dev/disk/by-id/wwn-0x5000c500a978a868-part4
/dev/disk/by-id/ata-ST1000NX0423_W470E22S-part4
/dev/disk/by-id/wwn-0x5000c500a978a868-part3
/dev/disk/by-id/ata-ST1000NX0423_W470E22S-part3
/dev/disk/by-id/wwn-0x5000c500a978a868-part6
/dev/disk/by-id/ata-ST1000NX0423_W470E22S-part6
/dev/disk/by-id/wwn-0x5000c500a978a868-part5
/dev/disk/by-id/ata-ST1000NX0423_W470E22S-part5
/dev/disk/by-id/wwn-0x5000c500a978a868-part1
/dev/disk/by-id/ata-ST1000NX0423_W470E22S-part1
/dev/disk/by-id/wwn-0x5000c500a978a868
/dev/disk/by-id/ata-ST1000NX0423_W470E22S
/dev/mapper/VG_XenStorage--e73aee11--7462--33fe--176d--3e2aa73ed049-MGT

Test with all files:
191960 Storage: GFS2 Functional
191961 Storage-Input: karma
191962 ToolStack-Input
191963 Ring0 BST
191964 Storage VDI Encryption Tests

scripts/test_perfmon.py
```diff
@@ -0,0 +1,112 @@
...
+@unittest.skipIf(sys.version_info < (3, 0), reason="requires python3")
+@patch("perfmon.Variable.update")
+@patch("perfmon.get_percent_fs_usage")
+@patch("perfmon.get_percent_log_fs_usage")
+@patch("perfmon.get_percent_mem_usage")
+class TestVMMonitor(unittest.TestCase):   
+    def test_process_rrd_updates(self, mock_get_percent_fs_usage, mock_get_percent_log_fs_usage, 
+                                 mock_get_percent_mem_usage, mock_update):
+        uuid = 'e1ae3f5d-4c8b-4575-bbb8-2af7e8a2c31e'
+        perfmon.all_xmlconfigs = {'e1ae3f5d-4c8b-4575-bbb8-2af7e8a2c31e': '<config><variable><name value="fs_usage"/><alarm_trigger_level value="0.9"/><alarm_trigger_period value="60"/><alarm_auto_inhibit_period value="3600"/></variable><variable><name value="mem_usage"/><alarm_trigger_level value="0.95"/><alarm_trigger_period value="60"/><alarm_auto_inhibit_period value="3600"/></variable><variable><name value="log_fs_usage"/><alarm_trigger_level value="0.9"/><alarm_trigger_period value="60"/><alarm_auto_inhibit_period value="3600"/></variable></config>'}
```
lindig, line 95: Is it possible to make this more diff-friendlly by avoiding overlong lines? I understand there are no spaces in this string. But maybe a similar case below could be improved.

lindig (approved):
This looks good to me but should be reviewed by someone with more Python experience before merging. This PR has accumulated a lot of comments which makes it difficult to review. This suggests it was raised prematurely.

scripts/test_perfmon.py
stephenchengCloud: Thanks!
Code updated.

scripts/static-vdis
```diff
@@ -227,41 +221,42 @@ def connect_smapiv1_nbd(params_nbd):
 
 
 def call_backend_attach(driver, config):
-    args = map(lambda arg: arg.encode('utf8'), [driver, config])
+    args = [arg.encode('utf8') for arg in [driver, config]]
```
liulinC, line 224: Previous updated `utf8` to `utf-8`, seems this one is missed.

stephenchengCloud: There's no difference between utf8 and utf-8.
https://stackoverflow.com/questions/35383506/what-is-the-difference-between-encoding-utf-8-and-utf8-in-python-3-5

liulinC: yes, It is just alias. 
You updated to `utf-8` in other place in this PR,  I thought you want to update all to keep consistency.

stephenchengCloud: Updated!

liulinC (approved):

stephenchengCloud:
- rebase commits
- squash related commits

liulinC:
Merge the code considering
- Unittest added as required.
- XenRT is happy with it.
- The PR LGTM.

scripts/static-vdis
```diff
@@ -1,12 +1,12 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Common functions for managing statically-attached (ie onboot, without xapi) VDIs
 
-from __future__ import print_function
-import sys, os, subprocess, json, urlparse, itertools
+
+import sys, os, subprocess, json, urllib.parse
```
psafont, line 6: There are multiple import that do not work in python2, this was never properly tested

stephenchengCloud: If I understand correctly, after updating to python3, it can only run under Python 3.
I noticed the issue in [CA-388124](https://issues.citrite.net/browse/CA-388124), I'll sort it out ASAP.

scripts/perfmon
stephenchengCloud: Re-checked the code.
The `list` is necessary for the code, for the code in the loop would modify the dictionary and would cause ` RuntimeError: dictionary changed size during iteration` in python3.
Will raised another PR to fix this.

scripts/perfmon
stephenchengCloud: Re-checked the code.
The `list` is necessary for the code, for the code in the loop would modify the dictionary and would cause ` RuntimeError: dictionary changed size during iteration` in python3.
Will raise another PR to fix this.

scripts/static-vdis
psafont: Sorry, my comment was misleading. the script is always run using python 3, as enforced by the shebang

stephenchengCloud: It's OK. If I were you, I would thought CA-388124 was caused by the python3 migration.

PR #5296, opened 2023-12-12
update mail-alarm and usb_scan to python3

acefei:
As @bernhardkaindl mentioned, those file already support python3, so this PR only changed:
- update shebang
- sort out imports

### DIFF
--- a/scripts/mail-alarm
+++ b/scripts/mail-alarm
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 #
 # mail-alarm: uses ssmtp to send a mail message, to pool:other_config:mail-destination
 #
@@ -11,18 +11,18 @@
 # the only thing that needs be set is pool:other-config:ssmtp-mailhub
 
 from __future__ import print_function
-import XenAPI
-import sys
+
+import json
 import os
+import re
+import sys
+import syslog
 import tempfile
 import traceback
-import syslog
-import json
-import re
-from xml.dom import minidom
-from xml.sax.saxutils import unescape
-from xml.parsers.expat import ExpatError
 from socket import getfqdn
+from xml.dom import minidom
+
+import XenAPI
 from xcp import branding
 
 # Go read man ssmtp.conf
--- a/scripts/usb_scan.py
+++ b/scripts/usb_scan.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 #
 # Copyright (C) Citrix Systems Inc.
 #
@@ -21,15 +21,16 @@
 # 2. check if device can be passed through based on policy file
 # 3. return the device info to XAPI in json format
 
-from __future__ import print_function
+
 import abc
 import argparse
 import json
-import xcp.logger as log
 import logging
-import pyudev
 import re
 import sys
+
+import pyudev
+import xcp.logger as log
 
 
 def log_list(l):


scripts/usb_scan.py
```diff
@@ -159,7 +158,7 @@ class UsbDevice(UsbObject):
     _PRODUCT_DETAILS = [_VERSION, _ID_VENDOR, _ID_PRODUCT, _BCD_DEVICE, _SERIAL,
                         _CLASS, _CONF_VALUE, _NUM_INTERFACES, _USB_SPEED]
     _PROPS = _PRODUCT_DESC + _PRODUCT_DETAILS
-    _PROPS_NONABLE = _PRODUCT_DESC + [_SERIAL]
+    _PROPS_NONABLE = [*_PRODUCT_DESC, _SERIAL]
```
psafont, line 161: It's not obvious why this transformation is made, nor why it's correct

bernhardkaindl: It's also not obvious why any of the transformations in this PR are made.

Both, mail-alarm and usb_scan already support Python3, since this commit in 2019:
https://github.com/xapi-project/xen-api/commit/3179d669739bda1c67e9859a83f4535b18d60476

This commit exactly describes how the changes were made, which is not described here.

The changes are described as:
```py
    remove redundant else after return
    remove redundant parameters in super(), ref https://docs.python.org/3/library/functions.html#super
    use with open() statement
    remove redundant/unused variables
```
Clearly, many changes were made by automated tools, others seem to be manual changes.

Don't mix automated changes with manual changes in a single PR and even a single commit, as this makes them basically needlessly hard to review in a sensible manner.

This raise the chances to overlook such unexplained changes.
Kudos to Pau that he found this change buried in all the automatic reformatting.

While the use "with open()" would have theoretically some merit, if this these scripts were long-running daemons, but they aren't, so any unclosed open files are closed by the OS when the program is terminated.

That is not to say that `with open` should not be used, but just that it is not needed.

Without further explanations, this PR is a code cleanup PR, not a Python3 PR.

From my point of view, this PR is a hard NACK (will not approve), because it:
1. Mixes automated with manual changes, making them hard to review.
1. It is not obviously needed, these tools already work with Python3.

If you like to contribute the `with open` change, open a new PR with only it.

Likewise for other 'else after return', but it is harmless, except that your IDE may warn you about it.

The other changes are even far less of a fixed "problem" that would belong to a Python3 PR.

bernhardkaindl (changes requested):
See my comment:
https://github.com/xapi-project/xen-api/pull/5296#discussion_r1424543694

I'll be on away until 2nd January, and as I wrote in the comment following Pau's finding, for me, this PR is not review and approvable for me due to the mixed amount of changes (see my longer comment for details).

The `with open` is a nice modernisation and file descriptor leak fix, but as these are scripts which run to completion, not long-running daemons, it's also not a real problem.

If you have actual fixes where the code does not work with Python3 in practice, open a new PR for each problem, with a description of the issue and how to test it.

scripts/usb_scan.py
acefei: > It's not obvious why this transformation is made, nor why it's correct
The * iterable unpacking operator was introduced in Python 3.5, which is useful as a more readable way of summing iterables into a list, ref https://peps.python.org/pep-0448/

acefei:
> See my comment: [#5296 (comment)](https://github.com/xapi-project/xen-api/pull/5296#discussion_r1424543694)
> 
> I'll be on away until 2nd January, and as I wrote in the comment following Pau's finding, for me, this PR is not review and approvable for me due to the mixed amount of changes (see my longer comment for details).
> 
> The `with open` is a nice modernisation and file descriptor leak fix, but as these are scripts which run to completion, not long-running daemons, it's also not a real problem.
> 
> If you have actual fixes where the code does not work with Python3 in practice, open a new PR for each problem, with a description of the issue and how to test it.

1. About Pau's comments, I've replied in https://github.com/xapi-project/xen-api/pull/5296#discussion_r1424803536
2. `The other changes are even far less of a fixed "problem" that would belong to a Python3 PR.`  What I did is writting a modenize python code instead of python2 and python3 compitable code, so most of changes is introduced after python3.
3.  It's not good practice that harmless code that is bad code smell should be left unmodified, because cognitive inertia may lead to future mistakes.
4. There is an unittest script in github action to check if the changes in script run as expect.

scripts/usb_scan.py
robhoes: The CI tests are failing with a syntax error on this line.

acefei: > The CI tests are failing with a syntax error on this line.
I will revert this change first, and will do other ticket to remove the python2 CI, since all of existed python scripts covered by unittests are upgraded python3.

bernhardkaindl:
> 1. About Pau's comments, I've replied in [update mail-alarm and usb_scan to python3 #5296 (comment)](https://github.com/xapi-project/xen-api/pull/5296#discussion_r1424803536)

ACK
 
> 2. `The other changes are even far less of a fixed "problem" that would belong to a Python3 PR.`  What I did is writting a modernize python3 code instead of python2 and python3 compatible code, in other word, most of changes are introduced after python3.

Thanks for admitting that your commits are not updating to Python3, but instead modernize Python3 code instead and remove Python2 compatibility.

As you admitted that, correct your commit messages and the PR title from "update ... to Python3" to:

"Modernize ... and remove Python2 compatibility"

Please apply this correction to:
- The commit message of commit 1
- The commit message of commit 2
- The caption title of this PR

> 3. It's not good practice that harmless code that is bad code smell should be left unmodified, I think cognitive inertia may lead to future mistakes.

I don't see the latter claim ("I think cognitive inertia may lead to future mistakes") as having been proven. If you make such claims, please provide some evidence for it.

I claim instead:

Sticking to what is proven to work well may even be a good thing, if there is no cost to it.

But there is one thing that's worse to any of that:

Lying and misrepresenting changes or work as doing X while doing Y.

You admitted yourself that this PR is modernizing the code and removing Python2 compatibility, which is not the same as updating - it already supports Python3. So correct your commits and the PR title.

Lying about what you do, possibly fooling others about it and may be yourself at some point later as well when you described something as something different is also a bad habit that may lead to future mistakes in general, I think.
 
> 4. There is a github action to run unittest to check if the changes in script run as expect.

Some evidence that is easy to verify that the unit tests are sufficiently verifying all your modernisations would be good.

For example, is there a code coverage report web page, e.g. at cover from CI or diff-coverage check against the master branch that shows that your modernisations are covered by the unit tests?

Example Codecov report: https://app.codecov.io/gh/xenserver/python-libs/pull/40?src=pr&el=h1

scripts/usb_scan.py
```diff
@@ -435,20 +433,19 @@ def parse_line(self, line):
         # 4. parse key=value pairs
         # pattern = r"\s*(class|subclass|prot|vid|pid|rel)\s*=\s*([0-9a-f]+)"
         last_end = 0
-        for matchNum, match in enumerate(re.finditer(self._PATTERN, target,
```
bernhardkaindl, line 438: I like this removal of the not needed `enumerate()`, but what is the coding rule for renaming `match` to `_match` here?

I did not see other examples for single pre underscored variables in this code base here, so it looks inconsistent. Furthermore, I also found no explanation that would match its use in a local loop variable on a very quick search.

scripts/usb_scan.py
```diff
@@ -435,20 +433,19 @@ def parse_line(self, line):
...
-                                                     re.IGNORECASE)):
-            if last_end != match.start():
-                self.parse_error(last_end, match.start(), target, line)
+        for _match in re.finditer(self._PATTERN, target, re.IGNORECASE):
```
bernhardkaindl, line 436: I like this removal of the not needed enumerate(), but what is the coding rule for renaming match to _match here?

I did not see other examples for single pre underscored variables in this code base here, so it looks inconsistent. Furthermore, I also found no explanation that would match its use in a local loop variable on a quick search.

scripts/mail-alarm
```diff
@@ -11,18 +11,18 @@
...
-import syslog
-import json
-import re
-from xml.dom import minidom
-from xml.sax.saxutils import unescape
-from xml.parsers.expat import ExpatError
 from socket import getfqdn
+
+import XenAPI
+from xml.dom.minidom import parseString
```
bernhardkaindl, line 25: The Import rule from PEP 8 at https://peps.python.org/pep-0008/#imports says:

> Imports should be grouped in the following order:
>
> 1. Standard library imports.
> 2. Related third party imports.
> 3. Local application/library specific imports.

Accordingly, `isort` sorts `from xml.dom.minidom import parseString` directly after `from socket import getfqdn` because it is a 3rd party import, not a local application/library specific import like `import XenAPI` and `from xcp import branding`:
```suggestion
from xml.dom.minidom import parseString

import XenAPI
```

bernhardkaindl (changes requested):
Please fix the import order of `from xml.dom.minidom import parseString` to be the last entry of the 3rd party libraries, not a local/application specific import, conforming to the PEP8 import rules, like `isort` applies them.

You can apply this change by clicking the button "Commit suggestion" in my review comment.

scripts/usb_scan.py
acefei: The change here is the avoidance ofusing the same name as a builtin variable,  `match` is a function in builtin module `re.match`.

acefei:
> As you admitted that, correct your commit messages and the PR title from "update ... to Python3" to:
> "Modernize ... and remove Python2 compatibility"

Fair enough, I will update this.

> I don't see the claim ("I think cognitive inertia may lead to future mistakes") as having been proven. If you make such claims, please provide some evidence for it.

When I said that, I intended to respond to this segment as follows.

> While the use "with open()" would have theoretically some merit, if this these scripts were long-running daemons, but they aren't, so any unclosed open files are closed by the OS when the program is terminated.

> That is not to say that with open should not be used, but just that it is not needed.

I insist on making `with open()` changes because I feel that it dont make sense to use inconsistent code style between daemon and non-daemon programs.

> You could have submitted the clean-up of bad code smells completely independent of the above as separate PRs. This would make each of these PRs more clear, with the correct commit messages and PR titles, of course.

You're right. I'll elaborate on my commit messages.

psafont (changes requested):
I'm blocking the merge because the PR breaks xs8 builds of xapi.spec.

Current xs8 builds use python2. As long as this happens this PR cannot be merged. Is there a PR missing to xapi.spec that removes the python2 dependency?

```
Bytecompiling .py files below /builddir/build/BUILDROOT/xapi-23.32.0.40.g5683e9378-1.2.g838bda5.xs8.x86_64/usr/lib/python2.7 using /usr/bin/python2.7
Bytecompiling .py files below /builddir/build/BUILDROOT/xapi-23.32.0.40.g5683e9378-1.2.g838bda5.xs8.x86_64/usr/lib/python3.6 using /usr/bin/python3.6
Compiling /builddir/build/BUILDROOT/xapi-23.32.0.40.g5683e9378-1.2.g838bda5.xs8.x86_64/opt/xensource/libexec/usb_scan.py ...
  File "/opt/xensource/libexec/usb_scan.py", line 161
    _PROPS_NONABLE = [*_PRODUCT_DESC, _SERIAL]
                      ^
SyntaxError: invalid syntax



RPM build errors:
error: Bad exit status from /var/tmp/rpm-tmp.RjFDNt (%install)
    Bad exit status from /var/tmp/rpm-tmp.RjFDNt (%install)
```

I recommend reevaluating the approach, I think what's valuable right now is making sure all used code is python-3 compatible, not removing python2 compatibility. Once the former is done the package can be build with using only python 3, and only then these cleanups / improvements can be integrated

acefei:
I will re-work for this PR for python2 and python3 compatible.

acefei:
Karma Testing 192605 (Dev Run)

psafont (approved):

liulinC (approved):

liulinC:
Merge the commit considering:
- The PR LGTM,
- XenRT is happy.

robhoes:
>Karma Testing 192605 (Dev Run) passed.

And unfortunately, that test suite did not actually include tests that exercise the change code.

PR #5297, opened 2023-12-12
CA-386620: Convert `kbps` to `Mbps` when passing QoS limit

danilo-delbusso:
I have found further issues while making this change (see ticket). This one is just to fix a long hanging fruit reported under CA-386620.

### DIFF
--- a/ocaml/networkd/lib/network_utils.ml
+++ b/ocaml/networkd/lib/network_utils.ml
@@ -695,13 +695,38 @@ module Ip = struct
     with _ ->
       Result.Error (Fail_to_set_vf_vlan, "Failed to set VF VLAN for: " ^ dev)
 
-  (* We know some NICs do not support config VF Rate, so will explicitly tell
-     XAPI this error*)
+  (** [set_vf_rate dev index rate] Updates the VF rate for a specified network interface card (NIC).
+    Note that this function converts the input rate from kilobits per second (kbps) to
+    megabits per second (Mbps) using division followed by ceil.
+
+    This function accounts for the fact that some NICs may not support VF rate configuration.
+    Any encountered errors during the process are explicitly reported to XAPI.
+
+    @param dev    The network device identifier.
+    @param index  The index of the VF.
+    @param rate   The desired rate in kbps, it will be converted into Mbps using division and ceil.
+
+    @return [Result.Ok] with the updated configuration upon success, or [Result.Error] wrapping [Fail_to_set_vf_rate] with the specific error message upon failure.
+*)
   let set_vf_rate dev index rate =
+    let divide_and_ceil (numerator : int) (denominator : int) : int =
+      let quotient = numerator / denominator in
+      let remainder = numerator mod denominator in
+      if remainder > 0 then
+        quotient + 1
+      else
+        quotient
+    in
     try
-      debug "Setting VF rate for dev: %s, index: %d, rate: %d" dev index rate ;
+      let megabits_per_second = divide_and_ceil rate 1000 in
+      debug "Setting VF rate for dev: %s, index: %d, rate: %d Mbps" dev index
+        megabits_per_second ;
       Result.Ok
-        (link_set dev ["vf"; string_of_int index; "rate"; string_of_int rate])
+        (link_set dev
+           [
+             "vf"; string_of_int index; "rate"; string_of_int megabits_per_second
+           ]
+        )
     with _ ->
       Result.Error (Fail_to_set_vf_rate, "Failed to set VF rate for: " ^ dev)
 end
--- a/ocaml/xenopsd/xc/device.ml
+++ b/ocaml/xenopsd/xc/device.ml
@@ -807,7 +807,7 @@ module Vif = struct
       match rate with
       | None ->
           []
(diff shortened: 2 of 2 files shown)


ocaml/xenopsd/xc/device.ml
```diff
@@ -816,14 +816,14 @@ module Vif = struct
               50000L
             (* 50ms by default *)
           in
-          let bytes_per_interval =
-            ((kbytes_per_s ^* 1024L) ^* timeslice_us) ^/ 1000000L
+          let bits_per_interval =
+            ((kbps ^* 1000L) ^* timeslice_us) ^/ 1000000L
```
psafont, line 820: Please explain in the commit message the change of unit base (from 2 to 10) as well

danilo-delbusso: mh I did assume that this was being passed to `ip link` tbh. @edwintorok can we easily confirm that's the case? 👀 

It does look like we're just writing it to `xenstore` here so not sure where it's picked up :/

robhoes: This is not related to the `set_vf_rate` function in networkd... That one is for SR-IOV  virtual functions (VFs). This code here is setting values in xenstore for netback.

robhoes: See L950 and below in this file for the SR-IOV/VF case.

robhoes: Also, we could check what netback really does, but this code explicitly talks about _kbytes_ per second, not kbits, and I have no reason to believe this is wrong. Other than the fact that the XenAPI parameter is unfortunately called `kbps`... :/ The XS docs at https://docs.xenserver.com/en-us/xenserver/8/networking/manage also talk about bytes.

ocaml/networkd/lib/network_utils.ml
```diff
@@ -695,13 +695,38 @@ module Ip = struct
...
+    This function accounts for the fact that some NICs may not support VF rate configuration.
+    Any encountered errors during the process are explicitly reported to XAPI.
+
+    @param dev    The network device identifier.
+    @param index  The index of the VF.
+    @param rate   The desired rate in kbps, it will be converted into Mbps using division and ceil.
+
+    @return [Result.Ok] with the updated configuration upon success, or [Result.Error] wrapping [Fail_to_set_vf_rate] with the specific error message upon failure.
+*)
   let set_vf_rate dev index rate =
```
lindig, line 711: Would suggest to rename `rate` to `kbit_rate` or similar. It would be even better to make it a label such that this is obvious in client code. `kb` could mean kbit or kbyte and I think we should avoid it,

ocaml/xenopsd/xc/device.ml
```diff
@@ -816,14 +816,14 @@ module Vif = struct
...
           in
-          if bytes_per_interval > 0L && bytes_per_interval < 0xffffffffL then
-            [("rate", sprintf "%Lu,%Lu" bytes_per_interval timeslice_us)]
+          if bits_per_interval > 0L && bits_per_interval < 0xffffffffL then
```
lindig, line 822: Would suggest to write the constant as `0xffff_ffffL` to make it easier to parse; so this is expected to be a 32-bit value? I am somewhat surprised that the existing code is not using `>=` and `<=`.

ocaml/networkd/lib/network_utils.ml
```diff
@@ -695,13 +695,38 @@ module Ip = struct
...
+      let quotient = numerator / denominator in
+      let remainder = numerator mod denominator in
+      if remainder > 0 then
+        quotient + 1
+      else
+        quotient
+    in
     try
-      debug "Setting VF rate for dev: %s, index: %d, rate: %d" dev index rate ;
+      let megabits_per_second = divide_and_ceil rate 1000 in
```
edwintorok, line 721: I think @robhoes found that this does indeed happen in another place already (that I missed). Except it divides by 1024 and not 1000, so would need to fix up the divisor.

edwintorok: See this line in device.ml: 'Network_client.Client.Sriov.{mac= Some mac; vlan; rate= rate_Mbps}'

ocaml/xenopsd/xc/device.ml
edwintorok: Using `kbps` to mean kilobytes/s is very confusing, `kbps` isn't ideal but almost everywhere it means `Kbit/s` (with K power of 10).

I think we should deprecate that parameter (accept it as kilobytes for backward compat) and add a properly named parameter. Maybe `bitrate`? And then the CLI and XC can do whatever unit conversions they want but the API will have the full value (although obviously we cannot actually consistently set very small values, so the granularity might be confusing).

ocaml/networkd/lib/network_utils.ml
edwintorok: Yes we also have 2 fields, both called `rate` in different IDLs, with different units, which is very confusing. (that is why I thought this 'ip link set' code would be used for regular VIFs, but it is only used for SRIOV ones)

danilo-delbusso:
Closing as it needs further investigation

PR #5298, opened 2023-12-12
CA-386552 XSI-1534 Failed to disable pool HA after missing HA statefile

lindig:
The issue arises because:

* xapi_ha.ml, function disable_internal
* When a static VDI is removed from an SR (manually by deleting it from an NFS server, for example)
* As the SR is re-scanned, it is also removed from the xapi database
* But the re-scan does not remove it from the list of static VDIs in /etc/xensource/static-vidis
* when the function is called, it obtains the list of static VDIs (the UUIds)
* and looks them up in the database - which fails
* the lookup happens quite a while before the value is used
* we will try to move the lookup close to the usage - hoping that it won't be needed
* another option would be to make sure that SR scans remove references to static VDIs

With this patch I was able to use `xe host-emergency-ha-disable force=true` but I can't be sure that I recreated exactly the situation the customer was facing. In any case, I don't expect this patch to cause any harm,


edwintorok (approved):
This also filters and removes static VDIs that have been removed from the DB, which is fine if we want to disable anyway.

psafont (approved):

robhoes:
> With this patch I was able to use xe host-emergency-ha-disable force=true 

It's `xe pool-ha-disable` that we are updating here though, isn't it? This is the "clean" HA disable. The emergency command already worked, but only on a single host, which then lead to fencing of all other hosts because of the violation of "survival rule 2".

lindig:
I have re-worked this. Now I can `xe pool-disable-ha` even when the statefile VDI is missing. So rebustness is increased. I still think we should not spend too much time on this because this is a corner case where the customer has messed with our internal representation - you can't protect yourself against all possible cases.

robhoes (approved):

PR #5301, opened 2023-12-15
update print-custom-templates to python3

acefei:
- format with `2to3`
- In Python 2, it was possible to compare strings without using quotes, but python3 not
- fix invalid `logout` function 

Tested manually on host

### DIFF
--- a/scripts/print-custom-templates
+++ b/scripts/print-custom-templates
@@ -1,42 +1,40 @@
-#!/usr/bin/python
+#!/usr/bin/env python3
 # Print out custom template UUIDs in CLI comma-separated minimal format
 # (c) Anil Madhavapeddy, Citrix Systems Inc, 2008
 
 import atexit
-import XenAPI
-import os, sys, time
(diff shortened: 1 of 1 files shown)


psafont (approved):

scripts/print-custom-templates
```diff
@@ -1,42 +1,40 @@
...
         tmplrec = templates[tmplref]
         try:
-            if not tmplrec['other_config'].has_key('default_template'):
-                output.append(tmplrec['uuid'])
-            elif tmplrec['other_config']['default_template'] != true:
-                output.append(tmplrec['uuid'])
-        except:
+            if "default_template" not in tmplrec["other_config"] or tmplrec["other_config"]["default_template"] != 'true':
+                output.append(tmplrec["uuid"])
+        except KeyError:
```
liulinC, line 33: Can this be replaced by
` with contextlib.suppress(KeyError)`

liulinC (approved):

PR #5302, opened 2023-12-16
Allow passing extra headers into the HTTP calls (CP-33676) +  markdown fixes

kc284:


kc284:
The format failure in this PR is misleading as there are no OCaml changes here.

kc284:
Rebased on master to fix the prechecks.

ocaml/sdk-gen/csharp/autogen/src/HTTP.cs
```diff
@@ -687,43 +687,57 @@ private static Stream DoHttp(Uri uri, IWebProxy proxy, bool nodelay, int timeout
...
+            while (ReadHttpHeaders(ref stream, proxy, noDelay, timeoutMs));
 
             return stream;
         }
 
         /// <summary>
         /// Adds HTTP CONNECT headers returning the stream ready for use
         /// </summary>
-        public static Stream HttpConnectStream(Uri uri, IWebProxy proxy, String session, int timeoutMs)
+        public static Stream HttpConnectStream(Uri uri, IWebProxy proxy, string session, int timeoutMs, params string[] additionalHeaders)
```
danilo-delbusso, line 698: shouldn't we use a dictionary of strings? `CONNECT <uri> HTTP/1.0` is not really a normal header. As a consumer I'd expect to pass headers as name value pairs, not just the string.

Not a biggie so I'm approving anyway. If you agree feel free to change

danilo-delbusso (approved):

ocaml/sdk-gen/csharp/autogen/src/HTTP.cs
kc284: Hmm, now that you mentioned it, how about using an optional System.Net.WebHeaderCollection object? This might be more flexible as it allows adding headers either as a string or as a key-value pair.

danilo-delbusso: sounds perfect, let's do that

ocaml/sdk-gen/csharp/autogen/src/HTTP.cs
```diff
@@ -668,9 +668,9 @@ private static void AuthenticateProxy(ref Stream stream, Uri uri, IWebProxy prox
         }
 
 
-        private static Stream DoHttp(Uri uri, IWebProxy proxy, bool nodelay, int timeout_ms, params string[] headers)
+        private static Stream DoHttp(Uri uri, IWebProxy proxy, bool noDelay, int timeoutMs, params string[] headers)
```
danilo-delbusso, line 671: As discussed, let's use `WebHeaderCollection `

danilo-delbusso (changes requested):

ocaml/sdk-gen/csharp/autogen/src/HTTP.cs
kc284: It turns out that the WebHeaderCollection does not accept a single string as a header unless it contains a colon - for example, a string like `CONNECT <uri> HTTP/1.0` makes it crash - which means the flexibility is only syntactic and it does enforce entering headers as key-value pairs.

ocaml/sdk-gen/csharp/autogen/src/HTTP.cs
kc284: I went for the dictionary in the end for the reason mentioned [here ](https://github.com/xapi-project/xen-api/pull/5302#discussion_r1446778061)

danilo-delbusso (approved):

PR #5303, opened 2023-12-17
[OIL] Moved some more generation code to templates.

kc284:
The first commit is cleaning some leftovers from PR #4782. The others are part of the on(slow)going effort to make the generation code dev friendlier.


psafont:
The failure seems genuine to me:
```
File "ocaml/sdk-gen/csharp/gen_csharp_binding.ml", lines 868-887, characters 15-16:
868 | ...............(fun (Map (l, r)) ->
869 |                  `O
870 |                    [
871 |                      ("map_key", `String (exposed_type l))
872 |                    ; ("map_value", `String (exposed_type r))
...
884 |                          (convert_from_proxy_hashtable_value "table[key]" r)
885 |                      )
886 |                    ]
887 |                )
Error (warning 8 [partial-match]): this pattern-matching is not exhaustive.
Here is an example of a case that is not matched:
(SecretString|String|Int|Float|Bool|DateTime|Enum (_, _)|Set _|Ref _|
Record _|Option _)
```

kc284:
Hmm, `convert_from_proxy_hashtable_value` does use `_` as the last match, so I'm not sure I understand where the warning comes from 😕

robhoes:
@kc284 is this this function that is not complete:
```
(fun (Map (l, r)) ->
```
as it only matches `Map`.

kc284:
Ahh 🤦‍♀️

ocaml/sdk-gen/csharp/gen_csharp_binding.ml
```diff
@@ -867,68 +858,45 @@ and has_unknown_entry contents =
...
-      assert false
+  let mapList = List.rev (TypeSet.fold (fun x acc -> x :: acc) !maps []) in
+  let json =
+    `O
+      [
+        ( "all_maps"
+        , `A
+            (List.map
+               (fun x ->
+                 match x with
```
psafont, line 869: You can skip some indentation with
```suggestion
               (function
```

kc284: It saves one line, so happy to remove it, but the formatter doesn't seem to change the identation.

psafont (approved):

kc284:
I rebased on master to resolve conflicts due to reformatting.

psafont:
I guess this needs a review from interfaces, or is another toolstack one enough?

kc284:
I think @danilo-delbusso wanted to review it.

kc284:
Rebased on master to resolve conflicts.

PR #5304, opened 2023-12-18
IH-397: remove all dependencies on async and core 

psafont:
This is a resubmission of https://github.com/xapi-project/xen-api/pull/5275

Now the code has passed internal tests that it initially failed (SR192267) with PBD plugs happily working.
There were two bugs in the initial PR:
- Double-closing of the launched process' stdin: lwt_process forced the double close, now a wrapper is added to work around the library issue
- Error on creating a dir that already existed: now this is taken into account, and even recursively creates the directory, if necessary

The more involved change is the replacement of async with lwt in xapi-storage-script

The error handling used in async is translated into Lwt_result type, this is
because the base Lwt only allows exception, which isn't enough to code the
errors produced.

Logging has been changed to use logs with an asynchronous reporter. Now
loglines print the loglevel.

The setup of the loops / promises has been changed, the outer loop has been
removed and its retry-on-error logic has been moved to the two inner watch
loops.

The inotify loop needed quite a few changes as async_inotify was doing complex
handling of events, even consolidating them. This has been simplified by
removing the watches on files and maintaining the directory one, and creating a
list of commands, which the watch loops can act upon, like registering and
deregistering plugins, for example.

This PR depends on https://github.com/xapi-project/xs-opam/pull/663 being merged for a fixed version of inotify

There's a failing test in the internal test suite, but it seems to happen at a lower level, and it fails differently from previous attempts (it's not a timeout before):
```
Dec 18 17:09:27 xs-host-41 SM: [10081] Got exception: Command [Dec 18 17:09:27 | 36001405324955ef764c40cfa4e4912c4: map in use
Dec 18 17:09:27 xs-host-41 SM: [10081] Dec 18 17:09:27 | failed to remove multipath map 36001405324955ef764c40cfa4e4912c4]. Retry number: 1
Dec 18 17:09:31 xs-host-41 SM: [10081] Got exception: Command [Dec 18 17:09:31 | 36001405324955ef764c40cfa4e4912c4: map in use
Dec 18 17:09:31 xs-host-41 SM: [10081] Dec 18 17:09:31 | failed to remove multipath map 36001405324955ef764c40cfa4e4912c4]. Retry number: 2
Dec 18 17:09:31 xs-host-41 SMAPIv3: [10081] - ERROR - Error detaching SR file:///run/sr-mount/847d50c2-dc5f-15ac-50f5-fd56a8dac832
```

A rerun passed the test, doesn't seem likely this change affected the results


ocaml/xapi-storage-script/main.ml
```diff
@@ -372,168 +678,152 @@ let fork_exec_rpc :
...
-                       in a RPC Dict"
-                    ]
-                 )
-              )
+        let w = p.proc#stdin in
+        Lwt_io.write w (Jsonrpc.to_string args) >>= fun () -> Lwt_io.close w
+      )
+      (fun _ -> Process.close p >>= fun _ -> Lwt.return_unit)
+    >>= fun () ->
+    Process.collect_output_and_wait p >>= fun output ->
```
edwintorok, line 714: Can we use `Lwt_process.pmap` to simplify some of this?

psafont: I'll try, but that method doesn't handle stderr

psafont: I don't believe it's usable, there's no way to extract the exit status when using pmap

lindig (approved):

ocaml/xapi-storage-script/main.ml
```diff
@@ -120,7 +118,19 @@ let pvs_version = "3.0"
...
 let supported_api_versions = [pvs_version; "5.0"]
 
-let api_max = List.fold_left ~f:String.max supported_api_versions ~init:""
+let string_min a b =
+  if String.compare a b <= 0 then
+    a
+  else
+    b
+
+let string_max a b =
```
edwintorok, line 434: I see you are keeping `Base`, and there is a `Base.String.{min,max}` that can be used.

ocaml/xapi-storage-script/main.ml
```diff
@@ -217,8 +227,8 @@ end = struct
       old PVS scripts *)
   let compat_uri device_config =
     match !V.version with
-    | Some version when String.(version = pvs_version) -> (
-      match List.Assoc.find ~equal:String.equal device_config "uri" with
```
edwintorok, line 221: I think Base has this too, so we could reduce the changes needed here by replacing `open Core` with `open Base` at the top. (Unless at some point we'd want to remove the dependency on Base, but I see no reason to do that, reusing code is better than reinventing the wheel). But this is fine too, since you've already done it.

ocaml/xapi-storage-script/main.ml
```diff
@@ -11,18 +11,293 @@
...
+let return = Lwt_result.return
+
+let fail = Lwt_result.fail
+
+let ( // ) = Filename.concat
+
+module Deferred = struct
+  let all_unit (lst : unit Lwt.t list) : unit Lwt.t =
+    let (_ : unit list Lwt.t) = Lwt.all lst in
+    Lwt.return ()
```
edwintorok, line 36: this isn't exactly equivalent: it won't wait until all promises in 'lst' are resolved, it will instead return immediately.

edwintorok: What you want here is Lwt.join

ocaml/xapi-storage-script/main.ml
```diff
@@ -11,18 +11,293 @@
...
+        access path [Unix.X_OK] >>= function
+        | Error exn ->
+            fail exn
+        | Ok () ->
+            return ()
+      )
+
+  let read_file_contents path =
+    let open Lwt_io in
+    with_file ~mode:input ~flags:[O_RDONLY] ~perm:0o000 path @@ fun chan ->
```
edwintorok, line 99: I think you can pass just `read`, the `@@ fun chan -> read chan` is redundant.

ocaml/xapi-storage-script/main.ml
```diff
@@ -11,18 +11,293 @@
...
+      | Regular | Other | Unknown ->
+          let msg =
+            Printf.sprintf
+              {|Could not create directory "%s": already exists and it's not a directory|}
+              path
+          in
+          Lwt.fail (Failure msg)
+      | Missing ->
+          let parent = Filename.dirname path in
+          loop ((fun () -> Lwt_unix.mkdir path perm) :: acc) parent
```
edwintorok, line 126: This could still return EEXIST if something else (whether in this process or elsewhere) races and succeeds creating the directory between checking for existence and creating it.
I think it should be possible to call mkdir directly (without checking first), handle EEXIST by ignoring it, and calling mkdir_p with Filename.parent when it returns ENOENT, and failing on ENOTDIR? (see also how this is implemented in unixext).

Anyway minimal fix here would be to wrap the mkdir with a try/catch and ignore EEXIST.

ocaml/xapi-storage-script/main.ml
```diff
@@ -11,18 +11,293 @@
...
+    close p >>= fun status ->
+    let exit_status = Output.exit_or_signal_of_unix status in
+    Lwt.return {Output.exit_status; stdout; stderr}
+
+  let run ~prog ~args ~input =
+    let p = create ~prog ~args in
+    Lwt.catch
+      (fun () ->
+        let w = p.proc#stdin in
+        Lwt_io.write w input >>= fun () -> Lwt_io.close w
```
edwintorok, line 217: In theory this could deadlock: the process may write something to its output every time it consumes chunks from the input, and because these are pipes they have a limited capacity. Once full they block and the other side needs to consume them.
I think you need to write to stdin and read from stdout/stderr concurrently to prevent deadlocks.

edwintorok: See how pmap is implemented: it starts both the getter and sender, and then it tries to wait for them (doesn't matter in what order because they'll keep running concurrently):[ ](https://github.com/ocsigen/lwt/blob/master/src/unix/lwt_process.cppo.ml#L501).

This is why I usually recommend to use `pmap` where possible, although if you need separate stdout and stderr then we don't have much choice but to reimplement it.

ocaml/xapi-storage-script/main.ml
psafont: Unixext uses a more pessimistic strategy: it expect to create all parent directories by default, so it does n mkdirs, where n is the number of segments in the path. This strategy is more optimistic in the sense that it expects the path to be there, but doing up to 2n mkdirs. Most of the time the directory will be there in this case, so it makes sense to be optimistic.

ocaml/xapi-storage-script/main.ml
psafont: I've reworked this, which makes for a simpler implementation, still needs to be tested.

psafont:
Rerun internal tests with the latest code, all green: SR193050

psafont:
Running a GFS2 test 10 times to check whether there might be some latent issues, all the ring3 tests passed, and so did the functional GFS2 ones

psafont:
All tests runs passed, we can think of merging this for the next version

psafont:
Due to the lack of appetite to review this, I'm splitting off the newly added functions into a private library and will develop a test suite for it. Hopefully it will raise confidence in the changes done

edwintorok:
I think the reason I didn't approve this before was because we were doing the XS8 release, but this should be mergable now once the conflicts are fixed.

psafont:
Closing for the time being while the tests are not done, a rebase is also needed.

PR #5309, opened 2023-12-20
CP-47033: Allow `protocol_lwt` server to handle multiple requests at the same time

Vincent-lau:
Previously the `protocol_lwt` server handles one request at a time which makes clusterd, for example, blocked if one request is taking a long time. Modifying the `protocol_lwt` (and `protocol_async` until it's removed) to make it possible to handle multiple requests.

### DIFF
--- a/ocaml/message-switch/core/make.ml
+++ b/ocaml/message-switch/core/make.ml
@@ -352,16 +352,23 @@ functor
 
     let listen ~process ~switch:port ~queue:name () =
       let token = Printf.sprintf "%d" (Unix.getpid ()) in
-      M.connect port >>= fun c ->
-      Connection.rpc c (In.Login token) >>|= fun (_ : string) ->
-      Connection.rpc c (In.CreatePersistent name) >>|= fun _ ->
+      let reconnect () =
+        M.connect port >>= fun request_conn ->
+        Connection.rpc request_conn (In.Login token) >>|= fun (_ : string) ->
+        M.connect port >>= fun reply_conn ->
+        Connection.rpc reply_conn (In.Login token) >>|= fun (_ : string) ->
+        return (Ok (request_conn, reply_conn))
+      in
+      reconnect () >>|= fun ((request_conn, reply_conn) as c) ->
       let request_shutdown = M.Ivar.create () in
       let on_shutdown = M.Ivar.create () in
+      let mutex = M.Mutex.create () in
+      Connection.rpc request_conn (In.CreatePersistent name) >>|= fun _ ->
       let t = {request_shutdown; on_shutdown} in
       let rec loop c from =
         let transfer = {In.from; timeout; queues= [name]} in
         let frame = In.Transfer transfer in
-        let message = Connection.rpc c frame in
+        let message = Connection.rpc request_conn frame in
         any [map (fun _ -> ()) message; M.Ivar.read request_shutdown]
         >>= fun () ->
         if is_determined (M.Ivar.read request_shutdown) then (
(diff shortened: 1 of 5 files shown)


ocaml/message-switch/core/make.ml
```diff
@@ -352,26 +352,36 @@ functor
...
     let listen ~process ~switch:port ~queue:name () =
       let token = Printf.sprintf "%d" (Unix.getpid ()) in
-      M.connect port >>= fun c ->
-      Connection.rpc c (In.Login token) >>|= fun (_ : string) ->
-      Connection.rpc c (In.CreatePersistent name) >>|= fun _ ->
+      let reconnect () =
+        M.connect port >>= fun request_conn ->
+        Connection.rpc request_conn (In.Login token) >>|= fun (_ : string) ->
+        Connection.rpc request_conn (In.CreatePersistent name) >>|= fun _ ->
+        M.connect port >>= fun reply_conn ->
```
edwintorok, line 359: copy/paste error? why are you logging in twice?

Vincent-lau: for two different connections? I just followed what `protocol_unix` did

```
 do_rpc request_conn (In.Login token) >>|= fun (_ : string) ->
      protect_connect switch @@ fun reply_conn ->
      do_rpc reply_conn (In.Login token) >>|= fun (_ : string) ->
```

ocaml/message-switch/core/make.ml
```diff
@@ -352,26 +352,36 @@ functor
...
         >>= fun () ->
         if is_determined (M.Ivar.read request_shutdown) then (
           M.Ivar.fill on_shutdown () ; return (Ok ())
         ) else
           message >>= function
           | Error _e ->
-              M.connect port >>= fun c ->
-              Connection.rpc c (In.Login token) >>|= fun (_ : string) ->
-              loop c from
+              M.Mutex.with_lock mutex (fun () ->
```
edwintorok, line 380: I think there is some duplication now with the mutex in Protocol_unix.ml
Would be useful to delete the code from Protocol_unix now that the mutex has been added to the shared code.

ocaml/message-switch/core/make.ml
edwintorok: Ah nevermind the copy/paste error is in Protocol_unix, this one is fine.
Anyway if we delete most of protocol_unix and rely on the code here that'll remove the duplication

ocaml/message-switch/core/make.ml
Vincent-lau: I feel it is non-trivial to funtorise the `protocol_unix` module, maybe we could do that as a separate PR?

ocaml/message-switch/switch/switch_main.ml
```diff
@@ -75,6 +75,13 @@ module Lwt_result = struct
   let ( >>= ) m f = m >>= fun x -> f (Stdlib.Result.get_ok x)
 end
 
+let exn_hook e =
+  let bt = Printexc.get_raw_backtrace () in
+  error "Caught exception in Lwt.async: %s\n%s" (Printexc.to_string e)
```
lindig, line 80: I think using `\n` in log messages is not a good idea. I believe it gets escaped anyway and so it would be better to use something else.

lindig (approved):

ocaml/message-switch/core/make.ml
```diff
@@ -352,33 +352,43 @@ functor
...
-      M.connect port >>= fun c ->
-      Connection.rpc c (In.Login token) >>|= fun (_ : string) ->
-      Connection.rpc c (In.CreatePersistent name) >>|= fun _ ->
+      let reconnect () =
+        M.connect port >>= fun request_conn ->
+        Connection.rpc request_conn (In.Login token) >>|= fun (_ : string) ->
+        Connection.rpc request_conn (In.CreatePersistent name) >>|= fun _ ->
+        M.connect port >>= fun reply_conn ->
+        Connection.rpc reply_conn (In.Login token) >>|= fun (_ : string) ->
+        Connection.rpc reply_conn (In.CreatePersistent name) >>|= fun _ ->
```
edwintorok, line 361: The old code didn't seem to create a persistent one on reply conn, and the Protocol_unix doesn't either (only on reuqest, not on reply).

ocaml/message-switch/switch/switch_main.ml
edwintorok: maybe we have something in the debug module already to print a backtrace?

ocaml/message-switch/core/make.ml
Vincent-lau: makes sense

robhoes (approved):
Approving on behalf of @edwintorok

PR #5310, opened 2023-12-20
Fix 'make install' and add a CI rule

edwintorok:
Opening PR to test CI, this will also need a .spec change


psafont (approved):

edwintorok:
That is probably a preexinsting bug in the Makefile, I'll try without '-j'. I wonder why we're not seeing this in koji.

robhoes (approved):

PR #5312, opened 2023-12-20
CP-46378: Propagate `dbg` to `Sm_exec.exec_xmlrpc`

GabrielBuica:
The `traceparent` was not propagated from the `sm.ml` functions. This commit solve this by propagating `dbg` into `Sm_exec.exec_xmlrpc` as an optional argument.
When `dbg` is passed, `Sm_exec.exec_xmlrps` now creates a span around the call.

### DIFF
--- a/ocaml/xapi/sm_exec.ml
+++ b/ocaml/xapi/sm_exec.ml
@@ -320,7 +320,7 @@ let with_session sr f =
       finally (fun () -> f session_id) (fun () -> destroy_session session_id)
   )
 
-let exec_xmlrpc ?context:_ ?(needs_session = true) (driver : string)
+let exec_xmlrpc ?dbg ?context:_ ?(needs_session = true) (driver : string)
     (call : call) =
   let do_call call =
     let xml = xmlrpc_of_call call in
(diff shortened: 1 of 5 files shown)


lindig (approved):
I am not too familiar with this code but this looks like mostly a mechanical change to thread an additional argument.

ocaml/xapi/sm_exec.ml
```diff
@@ -320,7 +320,7 @@ let with_session sr f =
       finally (fun () -> f session_id) (fun () -> destroy_session session_id)
   )
 
-let exec_xmlrpc ?context:_ ?(needs_session = true) (driver : string)
+let exec_xmlrpc ?dbg ?context:_ ?(needs_session = true) (driver : string)
```
robhoes, line 323: I think there are no callers of this function without `~dbg` anymore after both commits? Then we should make this a non-optional argument.

robhoes (approved):

ocaml/xapi/sm_exec.ml
GabrielBuica: Done in here https://github.com/xapi-project/xen-api/pull/5312/commits/4fced77c8c38335d315d72242048172f3d7578d9 and here https://github.com/xapi-project/xen-api/pull/5312/commits/366c55dff31b8be28aef9dc632ba86743944d277.

PR #5314, opened 2023-12-21
CP-46917 wait for DEMU "running" before unpausing

lindig:
Wait for the state of DEMU to be runnung before unpausing a guest.

* split out the code for waiting into wait_for_vgpu_state
* adjust timeouts

### DIFF
--- a/ocaml/xenopsd/xc/device.ml
+++ b/ocaml/xenopsd/xc/device.ml
@@ -3629,6 +3629,37 @@ module Dm = struct
     let module Q = (val Backend.of_profile dm) in
     Q.Dm.pci_assign_guest ~xs ~index ~host
 
+  let wait_for_vgpu_state states ~timeout ~xs ~domid ~task vgpus =
+    let open Xenops_interface.Vgpu in
+    match vgpus with
+    | {implementation= Nvidia {vclass= _; _}; _} :: _ -> (
+        let ioemu_failed fmt =
+          Printf.kprintf (fun msg -> raise (Ioemu_failed ("vgpu", msg))) fmt
+        in
+        let error_path =
+          Printf.sprintf "/local/domain/%d/vgpu/error-code" domid
+        in
+        let state_path = Service.Vgpu.state_path domid in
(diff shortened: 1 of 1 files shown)


ocaml/xenopsd/xc/device.ml
```diff
@@ -3833,6 +3840,14 @@ module Dm = struct
                       crash_reason
             )
         )
+    ) ;
+    (* CP-46917 Wait for vgpu/demu to be running. Timeout needs to be
+       long enough for migrations to complete *)
```
robhoes, line 3845: This function is called as part of the `VM_create_device_model` op that is part of the `VM_restore_devices` op. In the receiving host of a live migration, the `VM_receive_memory` op is executed, which runs `VM_restore_devices` after `VM_restore` and (just) before `VM_unpause` (after Synchronisation Point 3). See step 14 in [this diagram](https://xapi-project.github.io/new-docs/xenopsd/walkthroughs/live-migration/index.html). At this point, DEMU has already finished restoring the data from the sending host, so it won't take very long to get to the "running" state anymore.

Resume should be similar.

lindig: To be safe, I restored the timeout to the value we had before but happy to adjust it down. The first timeout I've set to 60s because we should see "initialising" almost immediately.

robhoes (approved):
I think this is correct, but please test all scenarios carefully.

psafont (approved):
Only question I had was about timeouts, but it has been answered. The change looks reasonable having read the explanation on why two waits are actually needed: one for the emulator to start initialising, and another to reach the running state to be abe to unpause the VM

robhoes:
>The change looks reasonable having read the explanation on why two waits are actually needed: one for the emulator to start initialising, and another to reach the running state to be abe to unpause the VM

And the other important bit is that QEMU is started in between those DEMU states. This is due to interaction between DEMU and QEMU. So what this PR does is wait for DEMU to be "running" after QEMU has been fully started up, and before unpausing the domain. It would be good to document these constraints actually.

PR #5316, opened 2023-12-21
CP-47046 optimise rule emitter

lindig:
Emit port-oriented rules for IPv4 and IPv6 only when needed.

The complication is that a single PVS server can have multiple addresses each of which can be IPv4 or IPv6. So we need both kind of rules if addresses as mixed but can emit only one kind if addresses are of one kind.

This was tested manually and by @rosslagerwall for IPv6.

### DIFF
--- a/ocaml/xenopsd/pvs/pvs_proxy_setup.ml
+++ b/ocaml/xenopsd/pvs/pvs_proxy_setup.ml
@@ -85,6 +85,20 @@ let is_ipv6 str =
   | Unix.PF_UNIX ->
       error "unexpected address format: %s" str
 
+let ( let+ ) x f = Result.map f x
+
+let is_ipv4 str =
+  let+ res = is_ipv6 str in
+  not res
+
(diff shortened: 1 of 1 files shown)


ocaml/xenopsd/pvs/pvs_proxy_setup.ml
```diff
@@ -85,6 +85,16 @@ let is_ipv6 str =
   | Unix.PF_UNIX ->
       error "unexpected address format: %s" str
 
+let is_ipv4 str = is_ipv6 str >>= fun res -> ok (not res)
```
psafont, line 88: Might be nicer to read by introducing the let binding for map:

```suggestion
let ( let+ ) x f = Result.map f x

let is_ipv4 str =
  let+ res = is_ipv6 str in
  not res
```

psafont (approved):

robhoes (approved):

PR #5318, opened 2023-12-22
CP-45570: Clear pending guidance when it got applied

gangj:
This PR fixes issue about clearing RestartToolstack guidance and RestartDeviceModel guidance, and clear RestartVM guidance.

Please refer commit message of each commit for details.

### DIFF
--- a/ocaml/xapi/xapi_host_helpers.ml
+++ b/ocaml/xapi/xapi_host_helpers.ml
@@ -344,6 +344,42 @@ let assert_xen_compatible () =
   if not compatible then
     raise Api_errors.(Server_error (xen_incompatible, []))
 
+let remove_pending_guidance ~__context ~self ~value =
+  let h = Db.Host.get_name_label ~__context ~self in
+  if
+    List.exists
+      (fun g -> g = value)
+      (Db.Host.get_pending_guidances ~__context ~self)
+  then (
+    debug "Remove guidance [%s] from host [%s]'s pending_guidances list"
+      Updateinfo.Guidance.(of_pending_guidance value |> to_string)
+      h ;
+    Db.Host.remove_pending_guidances ~__context ~self ~value
+  ) ;
+
+  if
+    List.exists
+      (fun g -> g = value)
+      (Db.Host.get_pending_guidances_recommended ~__context ~self)
+  then (
+    debug
+      "Remove guidance [%s] from host [%s]'s pending_guidances_recommended list"
+      Updateinfo.Guidance.(of_pending_guidance value |> to_string)
+      h ;
+    Db.Host.remove_pending_guidances_recommended ~__context ~self ~value
+  ) ;
+
+  if
+    List.exists
+      (fun g -> g = value)
+      (Db.Host.get_pending_guidances_full ~__context ~self)
+  then (
+    debug "Remove guidance [%s] from host [%s]'s pending_guidances_full list"
+      Updateinfo.Guidance.(of_pending_guidance value |> to_string)
+      h ;
+    Db.Host.remove_pending_guidances_full ~__context ~self ~value
+  )
+
 let consider_enabling_host_nolock ~__context =
   debug "Xapi_host_helpers.consider_enabling_host_nolock called" ;
   (* If HA is enabled only consider marking the host as enabled if all the storage plugs in successfully.
@@ -372,8 +408,6 @@ let consider_enabling_host_nolock ~__context =
        		   letting a machine with no fencing touch any VMs. Once the host reboots we can safely clear
        		   the flag 'host_disabled_until_reboot' *)
     let pool = Helpers.get_pool ~__context in
-    Db.Host.remove_pending_guidances ~__context ~self:localhost
-      ~value:`restart_toolstack ;
     let if_no_pending_guidances f =
       let host_pending_mandatory_guidances =
         Db.Host.get_pending_guidances ~__context ~self:localhost
@@ -395,18 +429,7 @@ let consider_enabling_host_nolock ~__context =
         f ()
     in
     if !Xapi_globs.on_system_boot then (
-      debug
-        "Host.enabled: system has just restarted: remove livepatch failure \
-         guidances" ;
-      Db.Host.remove_pending_guidances ~__context ~self:localhost
-        ~value:`reboot_host ;
-      Db.Host.remove_pending_guidances ~__context ~self:localhost
-        ~value:`reboot_host_on_livepatch_failure ;
-      Db.Host.remove_pending_guidances_recommended ~__context ~self:localhost
-        ~value:`reboot_host_on_kernel_livepatch_failure ;
-      Db.Host.remove_pending_guidances_recommended ~__context ~self:localhost
-        ~value:`reboot_host_on_xen_livepatch_failure ;
-
+      debug "Host.enabled: system has just restarted" ;
       if_no_pending_guidances (fun () ->
           debug
             "Host.enabled: system has just restarted and no pending mandatory \
--- a/ocaml/xapi/xapi_vm_lifecycle.ml
+++ b/ocaml/xapi/xapi_vm_lifecycle.ml
@@ -816,6 +816,42 @@ let checkpoint_in_progress ~__context ~vm =
     (List.map snd (Db.VM.get_current_operations ~__context ~self:vm))
   |> List.mem `checkpoint
 
+let remove_pending_guidance ~__context ~self ~value =
+  let v = Db.VM.get_name_label ~__context ~self in
+  if
+    List.exists
+      (fun g -> g = value)
+      (Db.VM.get_pending_guidances ~__context ~self)
+  then (
+    debug "Remove guidance [%s] from vm [%s]'s pending_guidances list"
+      Updateinfo.Guidance.(of_pending_guidance value |> to_string)
+      v ;
+    Db.VM.remove_pending_guidances ~__context ~self ~value
+  ) ;
+
+  if
+    List.exists
+      (fun g -> g = value)
+      (Db.VM.get_pending_guidances_recommended ~__context ~self)
+  then (
+    debug
+      "Remove guidance [%s] from vm [%s]'s pending_guidances_recommended list"
+      Updateinfo.Guidance.(of_pending_guidance value |> to_string)
+      v ;
+    Db.VM.remove_pending_guidances_recommended ~__context ~self ~value
+  ) ;
+
+  if
+    List.exists
+      (fun g -> g = value)
+      (Db.VM.get_pending_guidances_full ~__context ~self)
+  then (
+    debug "Remove guidance [%s] from vm [%s]'s pending_guidances_full list"
+      Updateinfo.Guidance.(of_pending_guidance value |> to_string)
+      v ;
+    Db.VM.remove_pending_guidances_full ~__context ~self ~value
+  )
+
 (** 1. Called on new VMs (clones, imports) and on server start to manually refresh
     the power state, allowed_operations field etc.  Current-operations won't be
     cleaned
@@ -823,7 +859,11 @@ let checkpoint_in_progress ~__context ~vm =
 let force_state_reset_keep_current_operations ~__context ~self ~value:state =
   (* First update the power_state. Some operations below indirectly rely on this. *)
   Db.VM.set_power_state ~__context ~self ~value:state ;
+  if state = `Suspended then
+    remove_pending_guidance ~__context ~self ~value:`restart_device_model ;
   if state = `Halted then (
+    remove_pending_guidance ~__context ~self ~value:`restart_device_model ;
+    remove_pending_guidance ~__context ~self ~value:`restart_vm ;
     (* mark all devices as disconnected *)
     List.iter
       (fun vbd ->
--- a/ocaml/xapi/xapi_vm_migrate.ml
+++ b/ocaml/xapi/xapi_vm_migrate.ml
@@ -479,6 +479,9 @@ let remove_stale_pcis ~__context ~vm =
 let pool_migrate_complete ~__context ~vm ~host:_ =
   let id = Db.VM.get_uuid ~__context ~self:vm in
   debug "VM.pool_migrate_complete %s" id ;
+  (* clear RestartDeviceModel guidance on VM migrate *)
+  Xapi_vm_lifecycle.remove_pending_guidance ~__context ~self:vm
+    ~value:`restart_device_model ;
   let dbg = Context.string_of_task __context in
   let queue_name = Xapi_xenops_queue.queue_of_vm ~__context ~self:vm in
   if Xapi_xenops.vm_exists_in_xenopsd queue_name dbg id then (
--- a/ocaml/xapi/xapi_xenops.ml
+++ b/ocaml/xapi/xapi_xenops.ml
@@ -1999,6 +1999,9 @@ let update_vm ~__context id =
             xenapi_of_xenops_power_state
               (Option.map (fun x -> (snd x).Vm.power_state) info)
           in
+          let power_state_before_update =
+            Db.VM.get_power_state ~__context ~self
+          in
           (* We preserve the current_domain_type of suspended VMs like we preserve
              the currently_attached fields for VBDs/VIFs etc - it's important to know
              whether suspended VMs are going to resume into PV or PVinPVH for example.
@@ -2275,13 +2278,28 @@ let update_vm ~__context id =
               try
                 Option.iter
                   (fun (_, state) ->
-                    debug "xenopsd event: Updating VM %s last_start_time <- %s"
-                      id
-                      (Date.to_string (Date.of_float state.Vm.last_start_time)) ;
                     let metrics = Db.VM.get_metrics ~__context ~self in
                     let start_time = Date.of_float state.Vm.last_start_time in
-                    Db.VM_metrics.set_start_time ~__context ~self:metrics
-                      ~value:start_time ;
+                    if
+                      start_time
+                      <> Db.VM_metrics.get_start_time ~__context ~self:metrics
+                    then (
+                      debug
+                        "xenopsd event: Updating VM %s last_start_time <- %s" id
+                        (Date.to_string (Date.of_float state.Vm.last_start_time)) ;
+                      Db.VM_metrics.set_start_time ~__context ~self:metrics
+                        ~value:start_time ;
+                      if
+                        (* VM start and VM reboot *)
+                        power_state = `Running
+                        && power_state_before_update <> `Suspended
+                      then (
+                        Xapi_vm_lifecycle.remove_pending_guidance ~__context
+                          ~self ~value:`restart_device_model ;
+                        Xapi_vm_lifecycle.remove_pending_guidance ~__context
+                          ~self ~value:`restart_vm
+                      )
+                    ) ;
                     create_guest_metrics_if_needed () ;
                     let gm = Db.VM.get_guest_metrics ~__context ~self in
                     let update_time =
@@ -3484,10 +3502,7 @@ let set_resident_on ~__context ~self =
   refresh_vm ~__context ~self ;
   !trigger_xenapi_reregister () ;
   (* Any future XenAPI updates will trigger events, but we might have missed one so: *)
-  Xenopsd_metadata.update ~__context ~self ;
-  Db.VM.remove_pending_guidances ~__context ~self ~value:`restart_device_model ;
-  Db.VM.remove_recommended_guidances ~__context ~self
-    ~value:`restart_device_model
+  Xenopsd_metadata.update ~__context ~self
 
 let update_debug_info __context t =
   let task = Context.get_task_id __context in
--- a/ocaml/xapi/dbsync_slave.ml
+++ b/ocaml/xapi/dbsync_slave.ml
@@ -232,6 +232,21 @@ let resynchronise_pif_params ~__context =
   (* Ensure that all DHCP PIFs have their IP address updated in the DB *)
   Helpers.update_pif_addresses ~__context
 
+let remove_pending_guidances ~__context =
+  let localhost = Helpers.get_localhost ~__context in
+  Xapi_host_helpers.remove_pending_guidance ~__context ~self:localhost
+    ~value:`restart_toolstack ;
+  if !Xapi_globs.on_system_boot then (
+    Xapi_host_helpers.remove_pending_guidance ~__context ~self:localhost
+      ~value:`reboot_host ;
+    Xapi_host_helpers.remove_pending_guidance ~__context ~self:localhost
+      ~value:`reboot_host_on_livepatch_failure ;
+    Xapi_host_helpers.remove_pending_guidance ~__context ~self:localhost
+      ~value:`reboot_host_on_kernel_livepatch_failure ;
+    Xapi_host_helpers.remove_pending_guidance ~__context ~self:localhost
+      ~value:`reboot_host_on_xen_livepatch_failure
+  )
+
 (** Update the database to reflect current state. Called for both start of day and after
     an agent restart. *)
 let update_env __context sync_keys =
@@ -351,4 +366,6 @@ let update_env __context sync_keys =
   switched_sync Xapi_globs.sync_chipset_info (fun () ->
       Create_misc.create_chipset_info ~__context info
   ) ;
-  switched_sync Xapi_globs.sync_gpus (fun () -> Xapi_pgpu.update_gpus ~__context)
+  switched_sync Xapi_globs.sync_gpus (fun () -> Xapi_pgpu.update_gpus ~__context) ;
+
+  remove_pending_guidances ~__context
--- a/ocaml/xapi/xapi_host_helpers.mli
+++ b/ocaml/xapi/xapi_host_helpers.mli
@@ -100,6 +100,20 @@ val assert_xen_compatible : unit -> unit
     currently installed xenctrl library and the currently running Xen hypervisor.)
     Raises XEN_INCOMPATIBLE if not, and caches the outcome of the check. *)
 
+val remove_pending_guidance :
+     __context:Context.t
+  -> self:API.ref_host
+  -> value:
+       [ `reboot_host
+       | `reboot_host_on_kernel_livepatch_failure
+       | `reboot_host_on_livepatch_failure
+       | `reboot_host_on_xen_livepatch_failure
+       | `restart_device_model
+       | `restart_toolstack
+       | `restart_vm ]
+  -> unit
+(** Removes update guidance from a host's all pending guidances lists. *)
+
 module Host_requires_reboot : sig
   val set : unit -> unit
   (** [set ()] is used to signal the host needs a reboot. This could be, for


ocaml/xapi/xapi_xenops.ml
```diff
@@ -2082,8 +2082,14 @@ let update_vm ~__context id =
                   (fun (_, state) ->
                     match state.Vm.domids with
                     | value :: _ ->
-                        Db.VM.set_domid ~__context ~self
-                          ~value:(Int64.of_int value)
+                        if
+                          Db.VM.get_domid ~__context ~self <> Int64.of_int value
```
minglumlu, line 2089: Would the line `( if different (fun x -> x.domids) then` already ensure the difference?

ocaml/xapi/xapi_vm_lifecycle.ml
```diff
@@ -875,6 +883,8 @@ let force_state_reset_keep_current_operations ~__context ~self ~value:state =
     Db.VM.set_resident_on ~__context ~self ~value:Ref.null ;
     (* make sure we aren't reserving any memory for this VM *)
     Db.VM.set_scheduled_to_be_resident_on ~__context ~self ~value:Ref.null ;
+    if not (Db.VM.get_is_a_template ~__context ~self) then
```
minglumlu, line 886: Could this be cleared together with `RestartVM`?

ocaml/xapi/xapi_xenops.ml
```diff
@@ -2281,13 +2284,25 @@ let update_vm ~__context id =
...
-                    debug "xenopsd event: Updating VM %s last_start_time <- %s"
-                      id
-                      (Date.to_string (Date.of_float state.Vm.last_start_time)) ;
                     let metrics = Db.VM.get_metrics ~__context ~self in
                     let start_time = Date.of_float state.Vm.last_start_time in
-                    Db.VM_metrics.set_start_time ~__context ~self:metrics
-                      ~value:start_time ;
+                    if
+                      start_time
+                      <> Db.VM_metrics.get_start_time ~__context ~self:metrics
```
minglumlu, line 2291: Again, would the line `( if different (fun x -> x.last_start_time) then` ensure the difference?

ocaml/xapi/xapi_xenops.ml
```diff
@@ -2281,13 +2284,25 @@ let update_vm ~__context id =
...
+                    then (
+                      debug
+                        "xenopsd event: Updating VM %s last_start_time <- %s" id
+                        (Date.to_string (Date.of_float state.Vm.last_start_time)) ;
+                      Db.VM_metrics.set_start_time ~__context ~self:metrics
+                        ~value:start_time ;
+                      if
+                        power_state <> `Suspended (* not VM suspend *)
```
minglumlu, line 2299: Would suspending a VM reach here? And what's more, resuming a VM may not reach here also.
This is based on what I read from:
https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/xc/xenops_server_xen.ml#L1433-L1473

ocaml/xapi/xapi_xenops.ml
```diff
@@ -2281,13 +2284,25 @@ let update_vm ~__context id =
...
+                        && power_state_before_update
+                           <> `Suspended (* not VM resume *)
+                      then
+                        Xapi_vm_lifecycle.remove_pending_guidance ~__context
```
minglumlu, line 2303: Would it always be safe to clear the `restart_device_model` here?
(Otherwise, `RestartVM` might not supersede the `RestartDeviceModel`)

ocaml/xapi/xapi_xenops.ml
gangj: No

ocaml/xapi/xapi_vm_lifecycle.ml
gangj: Yes, it can be cleared with no problem, but it will generate useless logs...

ocaml/xapi/xapi_xenops.ml
gangj: Yes, based on tests, and this change passed my local test.
From test, when a VM is suspended, the "Suspended" power state changed comes together with "last_start_time" update:

Below is one piece of log:

Dec 22 06:02:16 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: processing event for VM b3587353-cbcf-6008-dc9b-e00038467dbf
Dec 22 06:02:16 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf power_state <- Suspended
Dec 22 06:02:16 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf domid
Dec 22 06:02:16 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf memory_target <- 0
Dec 22 06:02:16 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf hvm <- false
Dec 22 06:02:16 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf last_start_time <- 19700101T00:00:00Z
Dec 22 06:02:16 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf vcpu_target <- 0
Dec 22 06:02:16 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: ignoring event for VBD (VM b3587353-cbcf-6008-dc9b-e00038467dbf not resident)
Dec 22 06:02:16 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: ignoring event for VIF (VM b3587353-cbcf-6008-dc9b-e00038467dbf not resident)

ocaml/xapi/xapi_xenops.ml
gangj: Yes, I think it should be safe to also clear `restart_device_model` here to make sure `restart_device_model` would always be when VM is restarted.
But if we indeed have a scenario which misses changing `domid` on VM restart, there should be a bug need to be fixed.

ocaml/xapi/xapi_vm_lifecycle.ml
minglumlu: I understand here is to clear `RestartDeviceModel` at the point of VM stop and suspend. Would the domid changing point capture them already?

ocaml/xapi/xapi_xenops.ml
minglumlu: The `different` with cache recognizes the differences already. So here the change is trying to filter out a fake positive? E.g. in cache it is A, but from xenopsd, it is B, and actually it is B ? 
But I understand the case should be considered is E.g. in cache it is A, but from xenopsd, it is B and then A, so that the `different` can't spot the change.

ocaml/xapi/xapi_xenops.ml
minglumlu: OK. For suspend, `last_start_time <- 19700101T00:00:00Z`.
What about resume? The last_start_time will be changed as well?

ocaml/xapi/xapi_xenops.ml
minglumlu: `But if we indeed have a scenario which misses changing domid on VM restart`
I don't think this will be a valid case. And I think your tests also confirmed this. It's agreed that RestartVM supersedes RestartDeviceModel.

ocaml/xapi/xapi_vm_lifecycle.ml
gangj: `domid` is changed in below line:
`Db.VM.set_domid ~__context ~self ~value:(-1L) ;`

ocaml/xapi/xapi_xenops.ml
gangj: No

ocaml/xapi/xapi_xenops.ml
gangj: Set to the time when it is resumed:

Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: processing event for VM b3587353-cbcf-6008-dc9b-e00038467dbf
Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf current_domain_type <- hvm
Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf power_state <- Running
Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf domid
Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf consoles
Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf memory_target <- 4294967296
Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf platform:timeoffset <- 0
Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf hvm <- true
Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf nomigrate <- false
Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf nested_virt <- false
Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf last_start_time <- 20231222T05:59:07Z
Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: VM b3587353-cbcf-6008-dc9b-e00038467dbf domid 59 uncooperative_balloon_driver = false
Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf domid 59 guest_agent
Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf PV drivers detected false
Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf domid 59 xsdata
Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf domid 59 memory target
Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf vcpu_target <- 2
Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf shadow_multiplier <- 1.00
Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VM b3587353-cbcf-6008-dc9b-e00038467dbf last_boot_CPU_flags <- 1fcbfbff-f7ba2223-2c100800-00000001-00000001-00000281-00000000-00000000-00001000-9c000400-00000000-00000000-00000000-00000000-00000000-00000000-00000000-00000000-00000000-00000000-00000000-00000000
Dec 22 06:14:12 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VBD b3587353-cbcf-6008-dc9b-e00038467dbf.xvda device <- xvda; currently_attached <- true
Dec 22 06:14:13 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenopsd event: Updating VIF b3587353-cbcf-6008-dc9b-e00038467dbf.0 currently_attached <- true
Dec 22 06:14:13 xrtmia-12-07 xapi: [debug||88 |org.xen.xapi.xenops.classic events D:642e31003c24|xenops] xenops event on VM b3587353-cbcf-6008-dc9b-e00038467dbf

ocaml/xapi/xapi_xenops.ml
gangj: I mean there should not have such case, as once a VM is restarted, its `domid` will always be changed.

minglumlu (approved):

ocaml/xapi/xapi_host_helpers.ml
```diff
@@ -411,8 +411,7 @@ let consider_enabling_host_nolock ~__context =
       debug
         "Host.enabled: system has just restarted: remove livepatch failure \
          guidances" ;
-      Db.Host.remove_pending_guidances ~__context ~self:localhost
-        ~value:`reboot_host ;
+      remove_pending_guidance ~__context ~self:localhost ~value:`reboot_host ;
```
robhoes, line 414: Should we use the same helper function for the three livepatch-failure cases below?

ocaml/xapi/xapi_vm_migrate.ml
```diff
@@ -438,7 +438,9 @@ let pool_migrate ~__context ~vm ~host ~options =
         detach_local_network_for_vm ~__context ~vm ~destination:host ;
         Helpers.call_api_functions ~__context (fun rpc session_id ->
             XenAPI.VM.pool_migrate_complete ~rpc ~session_id ~vm ~host
-        )
+        ) ;
+        Xapi_vm_lifecycle.remove_pending_guidance ~__context ~self:vm
```
robhoes, line 442: If you do this in `pool_migrate_complete`, then it will work for both `VM.pool_migrate` as well as `VM.migrate_send`.

ocaml/xapi/xapi_xenops.ml
robhoes: And the cache may be empty, e.g. after xapi restarts.

ocaml/xapi/xapi_xenops.ml
robhoes: >Set to the time when it is resumed:

I don't think so. The `last_start_time` value is persisted in the xenopsd metadata record and saved in the `VM.last_booted_record` field in xapi. On resume, this xenopsd metadata is restored in xenopsd. So `last_start_time` should still be the actual last start time.

Your example has:

    Dec 22 06:14:12 (...) last_start_time <- 20231222T05:59:07Z

which is a `last_start_time` 15 min before this is being logged. I assume this is when you had started the VM.

robhoes: It doesn't matter for this patch though.

ocaml/xapi/xapi_xenops.ml
robhoes: In theory, a VM can get the same domid when resumed as it had before it was suspended. For example, if it resumes on another host, or if the host has rebooted. Perhaps we should not rely on the domid change, and just remove the guidance in the `resume` function. You're doing that for migration as well.

ocaml/xapi/xapi_xenops.ml
gangj: Oh, I didn't notice that, thank you!

ocaml/xapi/xapi_host_helpers.ml
gangj: I think as in our design, livepatch-failure guidance will be added only to the recommended guidance list.

ocaml/xapi/xapi_xenops.ml
gangj: Yes, a VM might get the same domid when resumed:
1. resumes on another host, I think it is a VM migrate, and will be covered in `pool_migrate` in `xapi_vm_migrate.ml`
2. if the host has rebooted, I think then the VM will be closed first and it's domid will be set to `-1` first, so it will be covered in `force_state_reset_keep_current_operations` in `xapi_vm_lifecycle.ml`

Do you mean to remove the `restart_device_model` guidance in `resume` in `xapi_xenops.ml`? But I think then it will not be triggered when the VM is rebooted by XC or inside itself?

ocaml/xapi/xapi_vm_migrate.ml
gangj: Sorry, I missed the "storage live migration" case, thank you!

ocaml/xapi/xapi_xenops.ml
robhoes: Although suspend+resume and live migration use similar mechanisms internally, they are different higher up in xapi. You suspend a VM using the `VM.suspend` call, which saves its memory image to a VDI and writes the xenopsd state to a field in the xapi database. You then resume it using `VM.resume`, on any host in the pool and at any time. No code in xapi_vm_migrate.ml is used here.

But yes, it looks like we are always setting the domid to -1 when entering the suspend state, in `force_state_reset_keep_current_operations`, which would cover both cases. And you are already clearing the guidance fields there. Then the question is: do we really need to catch the changing domid case here again?

robhoes: >Do you mean to remove the restart_device_model guidance in resume in xapi_xenops.ml? But I think then it will not be triggered when the VM is rebooted by XC or inside itself?

We'd still need to watch the `last_start_time` to catch the internal reboot case.

gangj: [Clearing `restart_device_model` guidance when setting `domid` to -1](https://github.com/gangj/xen-api/blob/private/gangj/CP-45570/ocaml/xapi/xapi_vm_lifecycle.ml#L887-L888) in `force_state_reset_keep_current_operations` covers 2 scenarios: "VM shutdown", "VM suspend". And It does not cover "VM migrate" case. "VM migrate" is covered in `pool_migrate_complete` [here](https://github.com/gangj/xen-api/blob/private/gangj/CP-45570/ocaml/xapi/xapi_vm_migrate.ml#L482-L483).

For "VM shutdown" and "VM suspend", `restart_device_model` will not be cleared again when `domid` change is detected [here](https://github.com/gangj/xen-api/blob/private/gangj/CP-45570/ocaml/xapi/xapi_xenops.ml#L2091-L2092), because `force_state_reset_keep_current_operations` is called [when `power_state` change is detected](https://github.com/gangj/xen-api/blob/private/gangj/CP-45570/ocaml/xapi/xapi_xenops.ml#L2053), which is before `domid` change can be detected, and as `domid` is set to `-1` in `force_state_reset_keep_current_operations` already, so `domid` changing will not be detected for "VM shutdown" and "VM suspend".

If we clear `restart_device_model` guidance when `last_start_time` change is detected(instead of `domid` change is detected) the same as we clear the `restart_vm` guidance now, then for "VM shutdown", the `restart_device_model` guidance will be cleared twice: the first time [when `domid` is set to `-1`](https://github.com/gangj/xen-api/blob/private/gangj/CP-45570/ocaml/xapi/xapi_vm_lifecycle.ml#L887-L888) in `force_state_reset_keep_current_operations`, and then again when `last_start_time` is detected.

robhoes: The overall code change is quite small, but it is hard to get a good overview of the logic. If fields are updated in too many places, then it becomes really hard to understand what is going on. And if we rely too much on "if X happens, then we no longer need to do Y" logic, then I quickly get lost.

It's complicated enough now, while we are working on the feature; imagine looking at this again in two years when fixing some bug. I prefer fewer places where a given field is set, as long as it is overall correct.

I think:

* `force_state_reset` always handles the shutdown and suspend cases
* `pool_migrate_complete` handles live migration
* watching `last_start_time` handles reboots

and

* on shutdown and reboot, all guidance is cleared
* on suspend and live migrate, only restart_device_model is removed

Is tracking domid changes then _necessary_?

gangj: OK, changed to clear `RestartDeviceModel` guidance on `last_start_time` change, and with this change, clearing `RestartDeviceModel` inside `force_state_reset_keep_current_operations` is not needed: https://github.com/xapi-project/xen-api/pull/5318/commits/09be7631117cad1dd4a35b046b9a33edc21cb0c3

ocaml/xapi/xapi_xenops.ml
```diff
@@ -2275,13 +2278,25 @@ let update_vm ~__context id =
...
+                    then (
+                      debug
+                        "xenopsd event: Updating VM %s last_start_time <- %s" id
+                        (Date.to_string (Date.of_float state.Vm.last_start_time)) ;
+                      Db.VM_metrics.set_start_time ~__context ~self:metrics
+                        ~value:start_time ;
+                      if
+                        power_state <> `Suspended (* not VM suspend *)
+                        && power_state_before_update
+                           <> `Suspended (* not VM resume *)
```
minglumlu, line 2299: What about 'force shutdown' on a suspended VM?

ocaml/xapi/xapi_host_helpers.ml
minglumlu: Clearing 4 `reboot_host* tasks from 3 lists could be wrapped into one helpful. It would be safer in case there are any bugs.

ocaml/xapi/xapi_xenops.ml
gangj: Thank you for finding this missed scenario, this needs to be added.

robhoes: The function `force_state_reset` is the safest way to clear state (as you did earlier), as it it always called when a VM is shut down or suspended.

ocaml/xapi/xapi_host_helpers.ml
```diff
@@ -411,8 +411,7 @@ let consider_enabling_host_nolock ~__context =
       debug
```
minglumlu, line 411: As revealed from the latest test result, this variable will be `true` in following case:
The current host is a pool supporter. The  toolstack on the pool coordinator restarted. The call flow (**on_master_restart** -> Xapi_host_helpers.**consider_enabling_host_request**) on the pool supporter would reach here. So it is considered as a system boot although the supporter had neither a host reboot nor a toolstack restart.

minglumlu: Could it be fixed by adding an extra parameter in `Xapi_host_helpers.consider_enabling_host_request` to indicate that this caller is `on_master_restart `?

ocaml/xapi/xapi_host_helpers.ml
```diff
@@ -295,6 +295,9 @@ let startup_complete_m = Mutex.create ()
 let signal_startup_complete () =
   with_lock startup_complete_m (fun () -> startup_complete := true)
 
+let is_startup_complete () =
+  with_lock startup_complete_m (fun () -> !startup_complete)
```
lindig, line 299: Why does this need a lock? It only reads a value

ocaml/xapi/xapi_vm_helpers.ml
```diff
@@ -1579,3 +1579,11 @@ let ensure_device_model_profile_present ~__context ~domain_type ~is_a_template
   else
     (* only add device-model to an HVM VM platform if it is not already there *)
     default :: platform
+
+let remove_pending_guidance ~__context ~self ~value =
```
lindig, line 1583: Is this more or less the same code as above? Why have the same function twice?

lindig (approved):

ocaml/xapi/xapi_host_helpers.ml
minglumlu: Are operations on a boolean variable in OCaml always atomic?

ocaml/xapi/xapi_vm_helpers.ml
minglumlu: I think one is for VM, and another is for host.

ocaml/xapi/xapi_host_helpers.ml
gangj: I see it is read with lock this way: https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_host_helpers.ml#L300

ocaml/xapi/xapi_host_helpers.ml
robhoes: I'm not sure why `consider_enabling_host` is called in `on_master_restart`. The former clearly assumes that it is called as part of the startup sequence. Need to do some archeology...

minglumlu (approved):

ocaml/xapi/xapi_host_helpers.ml
```diff
@@ -394,17 +407,16 @@ let consider_enabling_host_nolock ~__context =
       else
         f ()
     in
-    if !Xapi_globs.on_system_boot then (
+    if !Xapi_globs.on_system_boot && not (is_startup_complete ()) then (
```
robhoes, line 410: `startup_complete` is set at the end of this function, outside of the if-statement that we are inside of here. This means that it will always be set the first time `consider_enabling_host` is called. Now this code is part of an if-statement that checks for other things, like `all_pbds_ok`, which may be `false` the first time `consider_enabling_host` is called. Then you may miss out on clearing the guidance fields.

Now all this interaction with `consider_enabling_host` is getting hard to follow. Also, the function `consider_enabling_host` is meant to enable hosts, and that is not related to clearing guidance DB fields. So overall, I think it is better and simpler to not clear those fields here, but move this to Dbsync_slave, which is meant to sync DB fields when xapi starts, and is guaranteed to be called only when xapi starts.

ocaml/xapi/xapi_vm_lifecycle.ml
```diff
@@ -816,14 +816,29 @@ let checkpoint_in_progress ~__context ~vm =
...
+
 (** 1. Called on new VMs (clones, imports) and on server start to manually refresh
     the power state, allowed_operations field etc.  Current-operations won't be
     cleaned
     2. Called on update VM when the power state changes *)
 let force_state_reset_keep_current_operations ~__context ~self ~value:state =
   (* First update the power_state. Some operations below indirectly rely on this. *)
   Db.VM.set_power_state ~__context ~self ~value:state ;
+  if state = `Suspended then
+    if not (Db.VM.get_is_a_template ~__context ~self) then
```
robhoes, line 835: Why do we need to check for templates? I don't see why pending_guidance needs to be preserved for those?

robhoes (changes requested):
See above... almost there ;)

ocaml/xapi/xapi_host_helpers.ml
gangj: Ah, @robhoes , I see the problem, thank you!

ocaml/xapi/xapi_vm_lifecycle.ml
gangj: Yes, `pending_guidance` don't needs to be preserved for templates. The if judgement here is to avoid clearing guidance for a template as it is not needed for templates. (And otherwise I think we will also need to update the log in `remove_pending_guidance` to avoid meaningless clearing guidance logs for templates.)

ocaml/xapi/xapi_host_helpers.ml
gangj: @robhoes , thanks for the guidance, I checked the code, now `Dbsync_slave.update_env` is called only when xapi startup. But I think the name `update_env` might be a little confusing, will it be called somewhere else later?
Is it ok to clear the host guidance in `dbsync.setup`? Just after `update_env`. I think maybe the name `setup` is more secure, it should not be called anywhere else except during startup. I had a try: https://github.com/xapi-project/xen-api/pull/5318/commits/79a049e58477269d6ac3f94e5877677fe1745411, please help to review again, thank you!

ocaml/xapi/xapi_vm_lifecycle.ml
robhoes: Templates are normally running, so this function `force_state_reset` is rarely (never?) called on templates. I'd like the keep the code simple and avoid two many cases, to not make people reading the code wonder what it is for and to avoid unexpected side-effects.

ocaml/xapi/xapi_host_helpers.ml
robhoes: `Dbsync_slave.update_env` is the right place to do this sort of thing.

gangj: OK, changed: https://github.com/xapi-project/xen-api/pull/5318/commits/a3028e1d8677207c8e52f07625887a3f815bc685

ocaml/xapi/xapi_xenops.ml
```diff
@@ -2275,13 +2278,29 @@ let update_vm ~__context id =
...
+                    then (
+                      debug
+                        "xenopsd event: Updating VM %s last_start_time <- %s" id
+                        (Date.to_string (Date.of_float state.Vm.last_start_time)) ;
+                      Db.VM_metrics.set_start_time ~__context ~self:metrics
+                        ~value:start_time ;
+                      if
+                        (* VM reboot *)
+                        power_state = `Running
+                        && power_state_before_update = `Running
```
robhoes, line 2295: Why do we need this condition? I believe we found that the start_time is only updated on start or reboot.

With this condition, we potentially miss the case where, during a reboot the power_state is temporarily set to Halted. This depends on when the state is checked.

ocaml/xapi/xapi_vm_lifecycle.ml
gangj: > Templates are normally running
Sorry, do you mean template's `power-state` is be running, but I checked their power-state, it is halted, for example:
`uuid ( RO)                : 03d30863-86a2-428e-94c8-88ddfcb8cd29
          name-label ( RW): Red Hat Enterprise Linux 7
    name-description ( RW): To use this template from the CLI, install your VM using vm-install, then set other-config-install-repository to the path to your network repository, e.g. http://<server>/<path> or nfs:server:/<path>


uuid ( RO)                : 7774689b-4ca1-4dea-8545-dddd6b64c17f
          name-label ( RW): Windows 10 (64-bit)
    name-description ( RW): Clones of this template will automatically provision their storage when first booted and then reconfigure themselves with the optimal settings for Windows 10 (64-bit).`

`[root@xrtmia-12-02 ~]# xe template-param-list uuid=7774689b-4ca1-4dea-8545-dddd6b64c17f | grep power-state
                           power-state ( RO): halted
[root@xrtmia-12-02 ~]# xe template-param-list uuid=03d30863-86a2-428e-94c8-88ddfcb8cd29 | grep power-state
                           power-state ( RO): halted
`

And with the `if` check removed, during xapi startup, `remove_pending_guidance` is called for each template, here is the log during xapi startup:
`Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [SUSE Linux Enterprise Server 12 SP4 (64-bit) (deprecated)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [SUSE Li nux Enterprise Server 12 SP4 (64-bit) (deprecated)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Windows Server 2016 (64-bit)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Windows  Server 2016 (64-bit)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Red Hat Enterprise Linux 9 (preview)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Red Hat  Enterprise Linux 9 (preview)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Debian Buster 10]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Debian  Buster 10]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Windows Server 2019 (64-bit)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Windows  Server 2019 (64-bit)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Red Hat Enterprise Linux 8]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Red Hat  Enterprise Linux 8]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Ubuntu Jammy Jellyfish 22.04]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Ubuntu  Jammy Jellyfish 22.04]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Gooroom Platform 2.0]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Gooroom  Platform 2.0]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Ubuntu Focal Fossa 20.04]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Ubuntu  Focal Fossa 20.04]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Oracle Linux 8]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Oracle  Linux 8]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [CentOS 7]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [CentOS  7]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Oracle Linux 7]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Oracle  Linux 7]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Windows Server 2022 (64-bit)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Windows  Server 2022 (64-bit)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Debian Bullseye 11]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Debian  Bullseye 11]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Debian Bookworm 12 (preview)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Debian  Bookworm 12 (preview)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [test 2: no memory, taken with guidance]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [test 2:  no memory, taken with guidance]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Ubuntu Bionic Beaver 18.04 (deprecated)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Ubuntu  Bionic Beaver 18.04 (deprecated)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Scientific Linux 7]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Scienti fic Linux 7]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Other install media]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Other i nstall media]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Rocky Linux 9 (preview)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Rocky L inux 9 (preview)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Windows 11 (preview)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Windows  11 (preview)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [NeoKylin Linux Server 7]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [NeoKyli n Linux Server 7]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [CentOS Stream 9 (preview)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [CentOS  Stream 9 (preview)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [SUSE Linux Enterprise 15 (64-bit)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [SUSE Li nux Enterprise 15 (64-bit)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [SUSE Linux Enterprise Server 12 SP5 (64-bit)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [SUSE Li nux Enterprise Server 12 SP5 (64-bit)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Rocky Linux 8]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Rocky L inux 8]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Red Hat Enterprise Linux 7]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Red Hat  Enterprise Linux 7]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartDeviceModel] from vm  [Windows 10 (64-bit)]'s all pending guidances lists
Jan 18 14:40:28 xrtmia-12-02 xapi: [debug||0 |dbsync (update_env) R:82eafac50a6b|xapi_vm_lifecycle] Remove guidance [RestartVM] from vm [Windows  10 (64-bit)]'s all pending guidances lists
`

ocaml/xapi/xapi_xenops.ml
gangj: From test, I see the `start_time` is updated on VM start, VM restart, VM resume. And updated to "19700101T00:00:00Z" on VM suspend and VM shutdown.

gangj: From my test, this always matches a reboot and it also passed xenrt tests. OK, I see, this may potentially miss some VM reboot.
Then how about changing to the last implementation, update on "VM start", "VM shutdown", "VM restart" by filtering "VM suspend" and "VM resume":
`
                      if
                        power_state <> `Suspended (* not VM suspend *)
                        && power_state_before_update
                           <> `Suspended (* not VM resume *)
                      then
                        (* clear RestartVM guidance on VM start, VM shutdown,
                           VM restart *)
                        Xapi_vm_helpers.remove_pending_guidance ~__context ~self
                          ~value:`restart_vm
`

ocaml/xapi/xapi_vm_lifecycle.ml
robhoes: Sorry I missed the word "not", for "not running" 😅

ocaml/xapi/xapi_xenops.ml
robhoes: The transition to 1970 for suspended VMs is unfortunate... Then I'd do
```
power_state = `Running
&& (power_state_before_update <> `Suspended)
```

ocaml/xapi/xapi_vm_lifecycle.ml
gangj: Sure :)
Since this causes confusion, I will remove the `if` check here and update the log in `remove_pending_guidance` to avoid those meaningless log.

ocaml/xapi/xapi_xenops.ml
gangj: OK, I see, can you please help to confirm we will clear the VM guidance both when VM is "down"(VM shutdown, VM suspend) and when VM is "up" (VM start, VM reboot):

```
power_state = `Running
&& (power_state_before_update <> `Suspended)
```
would match "VM start", "VM restart", and we can clear ``restart_device_model`  and ``restart_vm` for the 2 "VM up" scenarios.

And with the code in `force_state_reset_keep_current_operations`:
```
  if state = `Suspended then
    remove_pending_guidance ~__context ~self ~value:`restart_device_model ;
  if state = `Halted then (	  if state = `Halted then (
    remove_pending_guidance ~__context ~self ~value:`restart_device_model ;
    remove_pending_guidance ~__context ~self ~value:`restart_vm
    ) ;
```
which is for "VM down": clears ``restart_device_model` on "VM suspend" and "VM shutdown", and clears ``restart_vm` on "VM shutdown".

Thank you!

robhoes: I don't mind that... call it "defence in depth" :)

robhoes (approved):

robhoes:
Please squash and merge :)

gangj:
> Please squash and merge :)

Thank you, it finally finished~ :)

PR #5321, opened 2023-12-27
CP-46909: Update error returned from HTTP endpoint /updates

gangj:
The error was designed as HTTP error code only. This is not friendly to clients like XenCenter, i.e. XenCenter can't get more information from the error. The improvement is to return a JSON format data like:
{
    "error": {
        "code": 1,
        "data": ["<data1>", "<data2>", ...],
        "message": "<XAPI error code>"
    }
}.
The data list are the parameters of the error defined in datamodel_error, and the HTTP response code will be 200 always.

### DIFF
--- a/ocaml/xapi/xapi_pool.ml
+++ b/ocaml/xapi/xapi_pool.ml
@@ -3500,26 +3500,13 @@ let get_updates_handler (req : Http.Request.t) s _ =
       in
       match hs with
       | _ :: _ as hosts ->
-          let failures_of_404 =
-            [
-              Api_errors.no_repository_enabled
-            ; Api_errors.invalid_repomd_xml
-            ; Api_errors.invalid_updateinfo_xml
-            ]
-          in
           Xapi_pool_helpers.with_pool_operation ~__context
             ~self:(Helpers.get_pool ~__context)
             ~doc:"pool.get_updates" ~op:`get_updates (fun () ->
               try
                 http_res (Repository.get_pool_updates_in_json ~__context ~hosts)
               with
-              | Api_errors.(Server_error (failure, _)) as e
-                when List.mem failure failures_of_404 ->
-                  error "404: can't get updates for pool: %s"
-                    (ExnHelper.string_of_exn e) ;
-                  http_err failure []
-              | Api_errors.(Server_error (failure, _)) as e
-                when not (List.mem failure failures_of_404) ->
+              | Api_errors.(Server_error (failure, _)) as e ->
                   error "getting updates for pool failed: %s"
                     (ExnHelper.string_of_exn e) ;
                   http_err failure []
--- a/ocaml/xapi/xapi_pool.ml
+++ b/ocaml/xapi/xapi_pool.ml
@@ -3454,6 +3454,30 @@ let get_updates_handler (req : Http.Request.t) s _ =
   req.Http.Request.close <- true ;
   let query = req.Http.Request.query in
   Xapi_http.with_context "Getting available updates" req s (fun __context ->
+      let http_res json =
+        let json_str = Yojson.Basic.to_string json in
+        let size = Int64.of_int (String.length json_str) in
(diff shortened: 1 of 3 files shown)


ocaml/xapi/xapi_pool.ml
```diff
@@ -3487,36 +3511,27 @@ let get_updates_handler (req : Http.Request.t) s _ =
...
                   error "404: can't get updates for pool: %s"
                     (ExnHelper.string_of_exn e) ;
-                  Http_svr.headers s (Http.http_404_missing ())
+                  http_err failure []
               | Api_errors.(Server_error (failure, _)) as e
                 when not (List.mem failure failures_of_404) ->
                   error "getting updates for pool failed: %s"
                     (ExnHelper.string_of_exn e) ;
-                  raise e
+                  http_err failure []
```
Vincent-lau, line 3525: This is changing the behaviour of the code, isn't? Previously we had an exception whereas now we are returning normal responses. I wonder if this is ok to clients of xapi other than xencentre. Would it be better to preserve this raising of Api_error?

Vincent-lau, line 3525: This is changing the behaviour of the code, isn't it? Would this have wider impact since we no longer raise an exception? I wonder if it would be better to preserve the exception.

Vincent-lau:
Why not embed the json data in a http 4xx response?

gangj:
> Question regarding the design: why not embed the json data in a http 4xx response?

I think this is because the limitation in client now, I found this in design: "This is because currently the SDK will not bother with the body of the message when the HTTP response code is 404. So it's up to the clients to deduce from the parsing whether it is an error or not."
@kc284  , with the SDK, does XC had no chance to handle body for some http error like 404?

ocaml/xapi/xapi_pool.ml
gangj: I think the `raise` in the http handler will return client some 5xx error, which we also need to change to 200 to let client get this detailed error.

kc284:
It would need investigation because changing the handling would affect all http actions exposed via the SDK. However, I'm not convinced this is the right thing to do. Error 404 means an endpoint is not found, which is not the case here. Here the endpoint exists and returns specific error messages related to the underlying API functionality, I don't think we should mix them with low level http errors.

ocaml/xapi/xapi_pool.ml
minglumlu: I think the idea in design is, if it's one of `Api_errors`, it can be encoded into the new error structure. So all of these can be handled in this logic. `failures_of_404` can be removed actually.

lindig: I don't have the full context but there is a big difference between a HTTP 200 and 500. You can't use 200 to return information about an error - the request failed and that needs to be reflected in the error code.

kc284: There is a statement in the API reference:

```
When a low-level transport error occurs, or a request is malformed at the HTTP or RPC level,
the server may send an HTTP 500 error response, or the client may simulate the same.
...
All other failures are reported with a more structured error response, to allow better automatic
response to failures, proper internationalisation of any error message, and easier debugging.
```

We don't return 500 if RPCs fail due to xapi throwing an error, we return 200 with the error object in the response. The call `/updates` is implemented as an http backend, but the SDK expects a logic similar to the other RPCs, so currently it does not handle non-200 codes and cannot deduce the specific API error.

Note that this is the case with all the http actions, but I found it particularly problematic in the case of the updates, hence the current ticket. We might want to consider reviewing the way the SDK handles http codes at some point in future, but it would require assessing the impact on the implementing clients first.

minglumlu (approved):

kc284 (changes requested):
Please do not merge yet as I have stumbled upon an issue that will require further discussion.

PR #5324, opened 2023-12-28
Merge latest master branch into feature branch: private-network

duobei:


gangj (approved):
The last commit of the source branch is the last commit in master branch now, it can be merged directly is because there is no "private-network" related code commit in this feature branch yet:

```
git show -s feature/private-network.merge-master

commit 01dfc98b0bda437001bf2a27cd00759b66d0e916
Author: Christian Lindig <christian.lindig@cloud.com>
Date:   Wed Dec 20 13:29:39 2023 +0000

    CP-47046 optimise rule emitter for xenopsd PVS Proxy setup

    Emit port-oriented rules for IPv4 and IPv6 only when needed.

    The complication is that a single PVS server can have multiple addresses
    each of which can be IPv4 or IPv6. So we need both kind of rules if
    addresses as mixed but can emit only one kind if addresses are of one
    kind.

    Signed-off-by: Christian Lindig <christian.lindig@cloud.com>
```

```
git show -s master

commit 01dfc98b0bda437001bf2a27cd00759b66d0e916
Author: Christian Lindig <christian.lindig@cloud.com>
Date:   Wed Dec 20 13:29:39 2023 +0000

    CP-47046 optimise rule emitter for xenopsd PVS Proxy setup

    Emit port-oriented rules for IPv4 and IPv6 only when needed.

    The complication is that a single PVS server can have multiple addresses
    each of which can be IPv4 or IPv6. So we need both kind of rules if
    addresses as mixed but can emit only one kind if addresses are of one
    kind.

    Signed-off-by: Christian Lindig <christian.lindig@cloud.com>
```

minglumlu:
Yes. @duobei let's close this PR and reset the feature branch to master branch as no private changes on the feature branch yet.

PR #5325, opened 2024-01-01
CA-382035: xenopsd killed the wrong process

gangj:
When trying to stop Qemu service, if process qemu-dm-<domid> is not
running, besides trying to remove its pidfile, also trys to clear
related info from Xenstore.

When xenopsd-xc getting process id of a service, checks that its
commandline args include expected args.


ocaml/xenopsd/xc/service.ml
```diff
@@ -376,31 +376,32 @@ module Qemu = struct
...
-        best_effort "unmasking signals, qemu-pid is already gone from xenstore"
-          (fun () -> SignalMask.unset signal_mask domid
-        ) ;
-        best_effort "removing device model path from xenstore" (fun () ->
-            xs.Xs.rm (Device_common.device_model_path ~qemu_domid domid)
-        ) ;
-        best_effort (Printf.sprintf "removing %s" file_path) (fun () ->
-            Unix.unlink file_path
-        )
+    let pid = pid ~xs domid |> Option.value ~default:(-1) in
```
minglumlu, line 408: I can't see why this `-1` is being introduced.
Wouldn't the intrinsic `Option` work well here?

ocaml/xenopsd/xc/service.ml
```diff
@@ -256,31 +258,56 @@ module DaemonMgmt (D : DAEMONPIDPATH) = struct
...
+  (* Look up the commandline args for the process pid; *)
+  (* Check that they include expected args. *)
+  let is_cmdline_valid pid expected_args =
     try
-      match D.pid_location with
-      | (File file | Path_of {file; _}) when Sys.file_exists (file domid) ->
-          let path = file domid in
-          let ( let* ) = Option.bind in
-          let* pid = Unixext.pidfile_read path in
-          Unixext.with_file path [Unix.O_RDONLY] 0 (fun fd ->
```
minglumlu, line 266: Could you please explain why this check doesn't work?
The expectation here is that if the target process has terminated, this pidfile should have been unlocked, so `None` is supposed to be returned. But the issue is `Some pid` was returned.

ocaml/xenopsd/xc/service.ml
duobei: It's used in below code:
`if pid <> -1 `

ocaml/xenopsd/xc/service.ml
```diff
@@ -376,31 +376,32 @@ module Qemu = struct
...
+      ) ;
+      best_effort "killing qemu-dm" (fun () -> really_kill pid)
+    ) ;
+    best_effort "removing qemu-pid from xenstore" (fun () ->
+        xs.Xs.rm xenstore_path
+    ) ;
+    if pid <> -1 then
+      best_effort "unmasking signals, qemu-pid is already gone from xenstore"
+        (fun () -> SignalMask.unset signal_mask domid
+      ) ;
```
duobei, line 427: Could we merge those two `if pid <> -1` into a one?

ocaml/xenopsd/xc/service.ml
```diff
@@ -256,31 +258,56 @@ module DaemonMgmt (D : DAEMONPIDPATH) = struct
...
+    | Some pid ->
+        if is_cmdline_valid pid (D.expected_cmdline_args domid) then
+          Some pid
+        else (
+          warn "%s: pid not valid (domid = %d pid = %d)" D.name domid pid ;
           None
-    with _ -> None
+        )
+    | None ->
+        None
```
duobei, line 310: We could simply the code a little:
```ocaml
let pid ~xs domid =
    let ( let* ) = Option.bind in
    let* pid =
      try
        match D.pid_location with
        | (File file | Path_of {file; _}) when Sys.file_exists (file domid) ->
            let path = file domid in
            let* pid = Unixext.pidfile_read path in
            Unixext.with_file path [Unix.O_RDONLY] 0 (fun fd ->
                try
                  Unix.lockf fd Unix.F_TRLOCK 0 ;
                  (* we succeeded taking the lock: original process is dead.
                   * some other process might've reused its pid *)
                  None
                with Unix.Unix_error (Unix.EAGAIN, _, _) ->
                  (* cannot obtain lock: process is alive *)
                  Some pid
            )
        | Xenstore key | Path_of {key; _} ->
            (* case 1: backward compatibility during update installation: only has
                 xenstore pid available.
               case 2: pidfile(file domid) got removed unexpectly *)
            let pid = xs.Xs.read (key domid) in
            Some (int_of_string pid)
        | _ ->
            None
      with _ -> None
    in
    if is_cmdline_valid pid (D.expected_cmdline_args domid) then
      Some pid
    else (
      warn "%s: pid not valid (domid = %d pid = %d)" D.name domid pid ;
      None
    )
```

ocaml/xenopsd/xc/service.ml
gangj: Sure, I commented in the ticket, OK, let me also explain here. 

We can reproduce the problem this way:

1. setup a host with a rhel guest running.
2. kill the qemu for that guest in host console.
3. "Force Shut Down" the rhel guest in XC
4. xapi still try to kill the process with the log: "qemu-dm: stopping qemu-dm with SIGTERM"

Below is the details about how it is reproduced:

when the VM is "Force Shut Down" in XC, xenopsd-xc received 2 stop requests
1. for the first one, it did not return the pid as it found the pid invalid because we succeeded taking the lock: https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/xc/service.ml#L269-L271, but it removed the pidfile: https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/xc/service.ml#L380-L381
2. for the second stop request, when checking for pid, it did not match the `File` case as the pidfile doesn't exist: https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/xc/service.ml#L262, it matches the `Xenstore` case and returned the pid stored in Xenstore: https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/xc/service.ml#L279-L280, and then it will continue to kill the process with pid as process id: https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/xc/service.ml#L391

ocaml/xenopsd/xc/service.ml
gangj: Thanks, fixed.

ocaml/xenopsd/xc/service.ml
gangj: Yes, I think the unmasking should not related to `xs.Xs.rm xenstore_path`, thank you~

ocaml/xenopsd/xc/service.ml
lindig: I would suggest to add logging to see exactly which case is responsible for returning the PID.

ocaml/xenopsd/xc/service.ml
```diff
@@ -256,31 +258,52 @@ module DaemonMgmt (D : DAEMONPIDPATH) = struct
...
-          Some (int_of_string pid)
-      | _ ->
-          None
-    with _ -> None
+      let cmdline =
+        Printf.sprintf "/proc/%d/cmdline" pid
+        |> Unixext.string_of_file
+        |> Astring.String.cuts ~sep:"\000"
+      in
+      List.for_all (fun arg -> List.mem arg cmdline) expected_args
```
lindig, line 270: Is this the right test? Don't we expect the two lists to be identical using string quality?

ocaml/xenopsd/xc/service.ml
```diff
@@ -256,31 +258,56 @@ module DaemonMgmt (D : DAEMONPIDPATH) = struct
 
   let name = D.name
 
-  let pid ~xs domid =
+  (* Look up the commandline args for the process pid; *)
+  (* Check that they include expected args. *)
+  let is_cmdline_valid pid expected_args =
     try
```
edwintorok, line 264: Would be useful to also check the `exe` symlink. Although it is a bit unlikely that there will be a different process with same cmdline args.

ocaml/xenopsd/xc/service.ml
```diff
@@ -256,31 +258,56 @@ module DaemonMgmt (D : DAEMONPIDPATH) = struct
...
-             xenstore pid available *)
-          let pid = xs.Xs.read (key domid) in
-          Some (int_of_string pid)
-      | _ ->
+      let cmdline =
+        Printf.sprintf "/proc/%d/cmdline" pid
+        |> Unixext.string_of_file
+        |> Astring.String.cuts ~sep:"\000"
+      in
+      List.for_all (fun arg -> List.mem arg cmdline) expected_args
```
edwintorok, line 270: might be useful to add a comment here that this relies on the cmdline containing the domid (which it currently does with `-domid ...`, but just in case we change that in the future).
Otherwise we'd still risk killing another qemu instead of the correct one.

ocaml/xenopsd/xc/service.ml
```diff
@@ -826,36 +870,15 @@ module PV_Vnc = struct
 
   let pid ~xs domid = D.pid ~xs domid
 
-  (* Look up the commandline args for the vncterm pid; *)
-  (* Check that they include the vncterm binary path and the xenstore console
-     path for the supplied domid. *)
-  let is_cmdline_valid domid pid =
```
edwintorok, line 831: Thanks for spotting that we can reuse/repurpose this existing check.

edwintorok (commented):
I think this is going in the right direction, some small comments.

ocaml/xenopsd/xc/service.ml
gangj: The service has lot of cmdline arguments, for example for a qemu instance, please refer this:
`65568     1938  1881  0 Jan03 ?        00:10:25 qemu-dm-33 -machine pc-0.10,accel=xen,max-ram-below-4g=4026531840,allow-unassigned=true,trad_compat=True -vnc unix:/var/run/xen/vnc-33,lock-key-sync=off -monitor null -pidfile /var/run/xen/qemu-dm-33.pid -xen-domid 33 -m size=4088 -boot order=dc -usb -device usb-tablet,port=2 -smp 2,maxcpus=2 -chardev file,id=serial0,append=on,path=/tmp/local/scratch/xenrt/guest-console-logs/console.33.log -serial chardev:serial0 -display none -nodefaults -trace enable=xen_platform_log -sandbox on,obsolete=deny,elevateprivileges=allow,spawn=deny,resourcecontrol=deny -S -global PIIX4_PM.revision_id=0x1 -global ide-hd.ver=0.10.2 -global piix3-ide-xen.subvendor_id=0x5853 -global piix3-ide-xen.subsystem_id=0x0001 -global piix3-usb-uhci.subvendor_id=0x5853 -global piix3-usb-uhci.subsystem_id=0x0001 -global rtl8139.subvendor_id=0x5853 -global rtl8139.subsystem_id=0x0001 -parallel null -qmp unix:/var/run/xen/qmp-libxl-33,server,nowait -qmp unix:/var/run/xen/qmp-event-33,server,nowait -device xen-platform,addr=3,device-id=0x0001,revision=0x2,class-id=0x0100,subvendor_id=0x5853,subsystem_id=0x0001 -drive file=/dev/sm/backend/f4099437-827f-9c2f-4917-09ba8134822c/82d9cfa5-0edd-4654-b263-88f802f12167,if=none,id=ide0-hd0,auto-read-only=off,format=raw -device ide-hd,drive=ide0-hd0,bus=ide.0,unit=0,bios-chs-trans=forcelba -device rtl8139,netdev=tapnet0,mac=46:41:4a:2c:fd:f2,addr=4 -netdev tap,id=tapnet0,fd=7 -device VGA,vgamem_mb=8,addr=2,romfile=,rombar=1,subvendor_id=0x5853,subsystem_id=0x0001,qemu-extended-regs=false -vnc-clipboard-socket-fd 4 -chardev stdio,id=ovmf -device isa-debugcon,chardev=ovmf,iobase=0x402 -xen-domid-restrict -chroot /var/xen/qemu/root-33 -runas 65568:1023
qemu-dm-33-machinepc-0.10,accel=xen,max-ram-below-4g=4026531840,allow-unassigned=true,trad_compat=True-vncunix:/var/run/xen/vnc-33,lock-key-sync=off-monitornull-pidfile/var/run/xen/qemu-dm-33.pid-xen-domid33-msize=4088-bootorder=dc-usb-deviceusb-tablet,port=2-smp2,maxcpus=2-chardevfile,id=serial0,append=on,path=/tmp/local/scratch/xenrt/guest-console-logs/console.33.log-serialchardev:serial0-displaynone-nodefaults-traceenable=xen_platform_log-sandboxon,obsolete=deny,elevateprivileges=allow,spawn=deny,resourcecontrol=deny-S-globalPIIX4_PM.revision_id=0x1-globalide-hd.ver=0.10.2-globalpiix3-ide-xen.subvendor_id=0x5853-globalpiix3-ide-xen.subsystem_id=0x0001-globalpiix3-usb-uhci.subvendor_id=0x5853-globalpiix3-usb-uhci.subsystem_id=0x0001-globalrtl8139.subvendor_id=0x5853-globalrtl8139.subsystem_id=0x0001-parallelnull-qmpunix:/var/run/xen/qmp-libxl-33,server,nowait-qmpunix:/var/run/xen/qmp-event-33,server,nowait-devicexen-platform,addr=3,device-id=0x0001,revision=0x2,class-id=0x0100,subvendor_id=0x5853,subsystem_id=0x0001-drivefile=/dev/sm/backend/f4099437-827f-9c2f-4917-09ba8134822c/82d9cfa5-0edd-4654-b263-88f802f12167,if=none,id=ide0-hd0,auto-read-only=off,format=raw-deviceide-hd,drive=ide0-hd0,bus=ide.0,unit=0,bios-chs-trans=forcelba-devicertl8139,netdev=tapnet0,mac=46:41:4a:2c:fd:f2,addr=4-netdevtap,id=tapnet0,fd=7-deviceVGA,vgamem_mb=8,addr=2,romfile=,rombar=1,subvendor_id=0x5853,subsystem_id=0x0001,qemu-extended-regs=false-vnc-clipboard-socket-fd4-chardevstdio,id=ovmf-deviceisa-debugcon,chardev=ovmf,iobase=0x402-xen-domid-restrict-chroot/var/xen/qemu/root-33-runas65568:1023`

ocaml/xenopsd/xc/service.ml
```diff
@@ -376,40 +401,44 @@ module Qemu = struct
...
-        best_effort (Printf.sprintf "removing %s" file_path) (fun () ->
-            Unix.unlink file_path
-        )
+    let pid = pid ~xs domid in
+    let xenstore_path = pidxenstore_path domid in
+    let best_effort = Xenops_utils.best_effort in
+    let really_kill = Xenops_utils.really_kill in
+    if pid <> None then (
+      debug "qemu-dm: stopping qemu-dm with SIGTERM (domid = %d pid = %d)" domid
+        (Option.get pid) ;
```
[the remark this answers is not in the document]
gangj: Thanks, fixed

ocaml/xenopsd/xc/service.ml
```diff
@@ -405,23 +405,23 @@ module Qemu = struct
 
   let stop ~xs ~qemu_domid domid =
     let file_path = pidfile_path domid in
-    let pid = pid ~xs domid |> Option.value ~default:(-1) in
+    let pid = pid ~xs domid in
     let xenstore_path = pidxenstore_path domid in
     let best_effort = Xenops_utils.best_effort in
     let really_kill = Xenops_utils.really_kill in
-    if pid <> -1 then (
+    if pid <> None then (
```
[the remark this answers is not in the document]
gangj: Thanks, fixed

ocaml/xenopsd/xc/service.ml
gangj: Comments updated: https://github.com/xapi-project/xen-api/pull/5325/commits/8b8d84df18368975f52167c42e8f810107542c00

ocaml/xenopsd/xc/service.ml
gangj: I found it when I was checking all the code which get process id from the function `pid` in base module `DaemonMgmt`, and luckily it can be reused. :)

ocaml/xenopsd/xc/service.ml
```diff
@@ -235,6 +235,14 @@ module type DAEMONPIDPATH = sig
...
 
   val pid_location : Pid.path
+
+  (* To check if a pid is valid, look up the commandline args for the process
+     pid, check that expected args are included. The expected args are set by
+     (expected_cmdline_args domid) for each service.
+     Note that the parameter "domid" here is required as otherwise we can not
+     distinguish between the process instances of the same service for
+     different domains. *)
+  val expected_cmdline_args : int -> string list
```
lindig, line 245: I would suggest to make this a named argument `domid` - there is potential to confuse this with a pid.

ocaml/xenopsd/xc/service.ml
lindig: Again, would use `~pid` here to avoid confusion.

ocaml/xenopsd/xc/service.ml
lindig: Are the two not identical because we have some wrapper script that makes adjustment? In that case the current approach to look up a few critical command line arguments would be ok.

gangj:
> Are the two not identical because we have some wrapper script that makes adjustment? In that case the current approach to look up a few critical command line arguments would be ok.

Oh, I get your point. Yes, for example, some of the args for qemu-dm-<domid> may changes: https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/xc/device.ml#L3763

robhoes:
@gangj is there still anything missing from this?

gangj:
> @gangj is there still anything missing from this?

Yes, I think the change in this PR will break the [`pid` usage](https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/xc/device.ml#L3820) in qemu crash handler: as the `pid` will never return Some pid when the qemu process is gone.

Will check related code further to update this fix.

ocaml/xenopsd/xc/device.ml
```diff
@@ -3828,20 +3828,21 @@ module Dm = struct
...
+                    (Service.Qemu.pidxenstore_path_signal domid)
+                    crash_reason
+                with _ ->
+                  (* read Service.Qemu.pidxenstore_path from xs failed, should
+                     be after expected qemu stop or domain xs tree destroyed:
+                     this event arrived too late, nothing to do *)
+                  debug
+                    "domid=%d qemu-pid=%d: already removed from xenstore \
+                     during domain destroy"
+                    domid (getpid qemu_pid)
```
gangj, line 3845: @psafont Would you please help to review this change, thank you!

robhoes: Yes, this code in device.ml should not need any details about whether the pid in read from a file, xenstore, or both. The current code in the exception handler writes to a xenstore path related to the pid, which is already dubious (and I've heard it doesn't quite work anyway).

gangj: Hi @psafont , thanks for the review!
The change here is made is not because the main behavior of `pid` is changed: the pid is still get from pidfile and from xenstore as a fallback.
This change is made as some [validation](https://github.com/gangj/xen-api/blob/private/gangj/CA-382035/ocaml/xenopsd/xc/service.ml#L307-L311) for the `pid` is done before returning from the `pid` function in this PR, as this [comment](https://github.com/xapi-project/xen-api/pull/5325#issuecomment-1886514599).

ocaml/xenopsd/xc/service.ml
```diff
@@ -235,6 +235,14 @@ module type DAEMONPIDPATH = sig
   val name : string
 
   val pid_location : Pid.path
+
+  (* To check if a pid is valid, look up the commandline args for the process
+     pid, check that expected args are included. The expected args are set by
+     (expected_cmdline_args domid) for each service.
+     Note that the parameter "domid" here is required as otherwise we can not
```
robhoes, line 242: ...and the process name...

In some cases, the process name actually includes the domid, so then we don't need to rely on another `domid` parameter. But we _always_ want to match the process name.

gangj: Yes, the `domid` is a argument to generate the `expected args`, for example, for qemu, its expected arg is just its process name: "qemu-dm-<domid>" as [configured ](https://github.com/gangj/xen-api/blob/private/gangj/CA-382035/ocaml/xenopsd/xc/service.ml#L395)here.

ocaml/xenopsd/xc/device.ml
robhoes: I'm keen to merge the main fix in service.ml soon, and separate think about how to improve this on-crash behaviour.

gangj: I think the purpose of the [pid call](https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/xc/device.ml#L3820) here is to check if the qemu-pid is still available in xs as:
1) if qemu-dm-<domid> is [killed by xapi](https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/xc/service.ml#L391), the crash handler should not be triggered, it is implemented by [clearing the `qemu-pid` immediately after killing and surrounded by a signal_mask](https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/xc/service.ml#L387-L397). I think the surrounding `signal_mask` is to avoid an extreme condition: just after the [killing](https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/xc/service.ml#L391) and before the [xs rm](https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/xc/service.ml#L393), the crash handler is triggered, then it can not distinguish if the killing is done by xapi or not.
2) I think as `qemu-pid` will only be cleared by xapi: [clearing after killing](https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/xc/service.ml#L393) or cleared when the [while dom_path is cleared](https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/xc/domain.ml#L738-L742). it is used to check if the qemu process is stopped by xapi, if yes, no crash handling is needed, otherwise, we need the crash handler to do some cleanup.

ocaml/xenopsd/xc/service.ml
gangj: Ah, I see your point, comment updated: https://github.com/xapi-project/xen-api/pull/5325/commits/61354b1cf32471d90a1e55f62b181676ed380ac8

ocaml/xenopsd/xc/device.ml
gangj: > I'm keen to merge the main fix in service.ml soon, and separately think about how to improve this on-crash behaviour.

OK, let me add some more logs finally, should not take much time.

gangj: Log added.

ocaml/xenopsd/xc/service.ml
gangj: Log added: https://github.com/xapi-project/xen-api/commit/40016d998d3cc402cf390fa4c7220833d2edfa0b

ocaml/xenopsd/xc/service.ml
```diff
@@ -268,18 +268,34 @@ module DaemonMgmt (D : DAEMONPIDPATH) = struct
...
+          if not valid then
+            error "%s: pid read from %s not valid (pid = %d)" D.name pid_source
+              pid ;
+          valid
     with _ -> false
 
   let pid ~xs domid =
     let ( let* ) = Option.bind in
+    let pid_source = ref "unknown" in
     let* pid =
```
psafont, line 299:
```suggestion
    let* pid, pidsource =
```

ocaml/xenopsd/xc/service.ml
```diff
@@ -294,26 +310,27 @@ module DaemonMgmt (D : DAEMONPIDPATH) = struct
                   None
                 with Unix.Unix_error (Unix.EAGAIN, _, _) ->
                   (* cannot obtain lock: process is alive *)
-                  debug "%s: pid get from pidfile: %d" D.name pid ;
+                  pid_source := "pidfile" ;
                   Some pid
```
psafont, line 314:
```suggestion
                  Some (pid, "pidfile")
```

ocaml/xenopsd/xc/service.ml
```diff
@@ -294,26 +310,27 @@ module DaemonMgmt (D : DAEMONPIDPATH) = struct
...
             )
         | Xenstore key | Path_of {key; _} ->
             (* case 1: backward compatibility during update installation: only has
                  xenstore pid available.
                case 2: pidfile(file domid) got removed unexpectly *)
             let pid = xs.Xs.read (key domid) in
-            debug "%s: pid get from xs: %s" D.name pid ;
+            pid_source := "xenstore" ;
             Some (int_of_string pid)
```
psafont, line 322:
```suggestion
            Some ((int_of_string pid), "xenstore")
```

ocaml/xenopsd/xc/service.ml
```diff
@@ -294,26 +310,27 @@ module DaemonMgmt (D : DAEMONPIDPATH) = struct
...
         | _ ->
             None
       with _ -> None
     in
-    if is_cmdline_valid ~pid (D.expected_cmdline_items ~domid) then
+    if
+      is_cmdline_valid ~pid ~pid_source:!pid_source
```
psafont, line 328:
```suggestion
      is_cmdline_valid ~pid ~pid_source
```

psafont (commented):
There's no need to add a ref for the source, reuse the option type to contain the source as well

ocaml/xenopsd/xc/service.ml
robhoes: I don't think so, because of the `!`.

robhoes: Oh I see this is part of removing the `ref`.

ocaml/xenopsd/xc/service.ml
```diff
@@ -256,31 +265,66 @@ module DaemonMgmt (D : DAEMONPIDPATH) = struct
...
+            let pid = xs.Xs.read (key domid) in
+            Some (int_of_string pid, "xenstore")
+        | _ ->
+            None
+      with _ -> None
+    in
+    if is_cmdline_valid ~pid ~pid_source (D.expected_cmdline_items ~domid) then
+      Some pid
+    else
+      None
```
lindig, line 327: Should this not be logged? We failed to find the PID and but there is no explanations in the log. I think we should have reasons why we could not find a pid. That requires logging earlier - not just here.

robhoes (approved):

ocaml/xenopsd/xc/service.ml
gangj: Oh, sorry, see it just now! I think you are right, to make future debug easy, we need log inside `is_cmdline_valid` whenever `false` is returned.

PR #5326, opened 2024-01-02
CP-44637: Update data model of Tunnel for cross-server-private-network

duobei:
1. a new field tunnel_id of type int to record the id for all the tunnels in this "cross-server-private-network"
2. a new boolean field called cross_server to indicate whether tunnels to the remote server are set up automatically

### DIFF
--- a/ocaml/idl/datamodel.ml
+++ b/ocaml/idl/datamodel.ml
@@ -2214,6 +2214,16 @@ module Tunnel = struct
           ; param_release= next_release
           ; param_default= Some (VEnum "gre")
           }
+        ; {
+            param_type= Bool
+          ; param_name= "cross_server"
+          ; param_doc=
+              "Cross_server used to indicate what scenario is the tunnel used \
+               in: cross_server is true for a cross-server private network, \
+               false for XCP-ng"
+          ; param_release= next_release
+          ; param_default= Some (VBool false)
+          }
         ]
       ~result:(Ref _tunnel, "The reference of the created tunnel object")
       ~lifecycle:[(Published, rel_cowley, "Create a tunnel")]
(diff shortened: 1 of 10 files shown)


ocaml/idl/datamodel.ml
```diff
@@ -2214,6 +2214,15 @@ module Tunnel = struct
           ; param_release= next_release
           ; param_default= Some (VEnum "gre")
           }
+        ; {
+            param_type= Bool
+          ; param_name= "cross_server"
+          ; param_doc=
+              "Different scenario used for the tunnel \
```
lindig, line 2221: The documentation does not explain the purpose and consequence of this value. What does it mean when this value is true?

ocaml/idl/datamodel.ml
```diff
@@ -2274,6 +2283,18 @@ module Tunnel = struct
         ; field ~ty:tunnel_protocol ~default_value:(Some (VEnum "gre"))
             ~lifecycle:[(Published, "1.250.0", "Add protocol field to tunnel")]
             "protocol" "The protocol used for tunneling (either GRE or VxLAN)"
+        ; field ~qualifier:StaticRO ~ty:Int ~default_value:(Some (VInt 0L))
+            ~lifecycle:[(Published, rel_next, "Set vxlan id to tunnel")]
```
lindig, line 2287: The  documentation is confusing. This is a read-only value but the documentation uses active language: "set ... to". The documentation needs to describe what the observed int value means: "if true, <this and that happens or is also true>".

ocaml/idl/datamodel.ml
```diff
@@ -2274,6 +2283,18 @@ module Tunnel = struct
...
+            "tunnel_id" "VXLAN network identifier of tunnel"
+        ; field ~qualifier:StaticRO ~ty:Bool ~default_value:(Some (VBool false))
+            ~lifecycle:
+              [
+                ( Published
+                , rel_next
+                , "Set true if it is cross-server-private-network"
```
lindig, line 2294: Same problem: "set" is the wrong word: "true for a cross-server private network (observe where to put hyphens"

lindig (changes requested):

gangj:
@duobei To resolve the build failure here, quite a few of code in different places needs to be updated, I have a try here: https://github.com/xapi-project/xen-api/commit/100cf526d94ffd42637559c77a0c9a7911a12bc2 just for your reference, it will become easier when you got familiar with related xapi code. :)

ocaml/idl/datamodel.ml
duobei: I'd like to fix it.

duobei:
> @duobei To resolve the build failure here, quite a few of code in different places needs to be updated, I have a try here: [100cf52](https://github.com/xapi-project/xen-api/commit/100cf526d94ffd42637559c77a0c9a7911a12bc2) just for your reference, it will be a piece of cake when you got familiar with related xapi code. :) BTW, you can't resolve build failure without a local build env.

Thank you so much.

ocaml/xapi/xapi_tunnel.mli
```diff
@@ -20,6 +20,7 @@ val create :
   -> transport_PIF:[`PIF] Ref.t
   -> network:[`network] Ref.t
   -> protocol:API.tunnel_protocol
+  -> cross_server:bool
```
gangj, line 23: I think we should make the `cross_server` parameter optional and with default value as `false` for backward compatibility, then XCP-ng's usage of the internal tunnel will not break.

lindig (approved):

PR #5328, opened 2024-01-03
CA-384537 support NBD for CD device

lindig:
Passing an NBD to qemu using qmp: open the socket and pass it to qmp.
Compared to the existing case that the CD is opened as a raw device the
handling of the socket is slightly different: it does not require a `Qmp.Remove_fd`.

### DIFF
--- a/ocaml/xenopsd/xc/device.ml
+++ b/ocaml/xenopsd/xc/device.ml
@@ -1411,14 +1411,8 @@ module PCI = struct
     | Some (Supported I915) ->
         ()
     | Some drv ->
-        raise
-          (Xenopsd_error
-             (Internal_error
-                (Printf.sprintf "Fail to bind to i915, device is bound to %s"
-                   (string_of_driver drv)
-                )
-             )
-          )
+        internal_error "Fail to bind to i915, device is bound to %s"
+          (string_of_driver drv)
 
   let unbind devstr driver =
     let driverstr = string_of_driver driver in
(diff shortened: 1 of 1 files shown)


lindig:
This was manually tested with a build that uses NBD to provide the CD for guests.

ocaml/xenopsd/xc/device.ml
```diff
@@ -3024,6 +3016,33 @@ module Backend = struct
           match params with
           | "" ->
               qmp_send_cmd domid Qmp.(Eject (cd, Some true)) |> ignore
+          | params when is_nbd params ->
+              let path, exportname = nbd params in
+              info "%s: domain=%d NBD socket=%s" __FUNCTION__ domid path ;
+              with_socket path @@ fun fd ->
+              let cmd = Qmp.(Add_fd None) in
```
edwintorok, line 3023: This code could be shared with the non-nbd case by defining a function and giving it the path and whether to close the FD or not as parameters.

Unless we want to completely drop the non-nbd path eventually?

edwintorok (approved):
The changes themselves look ok

ocaml/xenopsd/xc/device.ml
lindig: I expect that we will drop the non-nbd code eventually. If you look closely there are some difference between the two cases that are not so easy to factor out and so I am not convinced that a more complicated logic buys much.

lindig:
Reviewers might want to review by commit because I simplified some error reporting code; the actual change is in one commit and the rest is improvements.

Vincent-lau (approved):
why is Remove_fd not needed in this special case?

lindig:
@rosslagerwall might be able to explain. We initially had it and it didn't work. I suspect in the raw device case the FD is dup'ed but not in the socket case. It does feel inconsistent, though.

rosslagerwall:
> @rosslagerwall might be able to explain. We initially had it and it didn't work. I suspect in the raw device case the FD is dup'ed but not in the socket case. It does feel inconsistent, though.

Indeed, this does appear to be what it is doing but I can't explain why it is inconsistent.

PR #5331, opened 2024-01-05
Add "xe host-updates-show-available" CLI and update "xe host-apply-updates"

minglumlu:
To facilitate the users who use XE CLI to perform update, a new XE CLI "xe host-show-updates" is added in this series to show the available updates on a host. Prior to this change, the information can be get from XAPI HTTP /updates endpoint. In addition, the available updates will be printed out in existing "xe host-apply-updates".

### DIFF
--- a/ocaml/xapi-cli-server/cli_frontend.ml
+++ b/ocaml/xapi-cli-server/cli_frontend.ml
@@ -1032,7 +1032,7 @@ let rec cmdtable_data : (string * cmd_spec) list =
         reqd= ["hash"]
       ; optn= []
       ; help= "Apply updates from enabled repository on specified host."
-      ; implementation= No_fd Cli_operations.host_apply_updates
+      ; implementation= With_fd Cli_operations.host_apply_updates
       ; flags= [Host_selectors]
       }
     )
--- a/ocaml/xapi-cli-server/cli_frontend.ml
+++ b/ocaml/xapi-cli-server/cli_frontend.ml
@@ -1047,6 +1047,15 @@ let rec cmdtable_data : (string * cmd_spec) list =
       ; flags= [Neverforward]
       }
     )
+  ; ( "host-updates-show-available"
+    , {
+        reqd= []
+      ; optn= []
+      ; help= "Show available updates for a specified host."
+      ; implementation= With_fd Cli_operations.host_updates_show_available
+      ; flags= [Host_selectors]
+      }
+    )
   ; ( "patch-upload"
     , {
         reqd= ["file-name"]
--- a/ocaml/xapi-cli-server/cli_operations.ml
+++ b/ocaml/xapi-cli-server/cli_operations.ml
@@ -5462,11 +5462,14 @@ let wait_for_task_complete rpc session_id task_id =
     Thread.delay 1.0
   done
 
-let check_task_status ~rpc ~session_id ~task ~fd ~label ~ok =
+let check_task_status ?(quiet_on_success = false) ~rpc ~session_id ~task ~fd
+    ~label ~ok () =
   (* if the client thinks it's ok, check that the server does too *)
   match Client.Task.get_status ~rpc ~session_id ~self:task with
-  | `success when ok ->
+  | `success when ok && not quiet_on_success ->
       marshal fd (Command (Print (Printf.sprintf "%s succeeded" label)))
+  | `success when ok && quiet_on_success ->
+      ()
   | `success ->
       marshal fd
         (Command
@@ -5514,7 +5517,7 @@ let download_file rpc session_id task fd filename uri label =
   wait_for_task_complete rpc session_id task ;
   (* Check the server status -- even if the client thinks it's ok, we need
      	   to check that the server does too. *)
-  check_task_status ~rpc ~session_id ~task ~fd ~label ~ok
+  check_task_status ~rpc ~session_id ~task ~fd ~label ~ok ()
 
 let download_file_with_task fd rpc session_id filename uri query label task_name
     =
@@ -5715,7 +5718,7 @@ let blob_get fd _printer rpc session_id params =
       in
       wait_for_task_complete rpc session_id blobtask ;
       check_task_status ~rpc ~session_id ~task:blobtask ~fd ~label:"Blob get"
-        ~ok
+        ~ok ()
     )
     (fun () -> Client.Task.destroy ~rpc ~session_id ~self:blobtask)
 
@@ -5757,7 +5760,7 @@ let blob_put fd _printer rpc session_id params =
       wait_for_task_complete rpc session_id blobtask ;
       (* if the client thinks it's ok, check that the server does too *)
       check_task_status ~rpc ~session_id ~task:blobtask ~fd ~label:"Blob put"
-        ~ok
+        ~ok ()
     )
     (fun () -> Client.Task.destroy ~rpc ~session_id ~self:blobtask)
 
@@ -7642,6 +7645,57 @@ let update_resync_host _printer rpc session_id params =
   let host = Client.Host.get_by_uuid ~rpc ~session_id ~uuid in
   Client.Pool_update.resync_host ~rpc ~session_id ~host
 
+let get_avail_updates_uri ~session_id ~task ~host =
+  let query =
+    [
+      ("session_id", [Ref.string_of session_id])
+    ; ("task_id", [Ref.string_of task])
+    ; ("host_refs", [Ref.string_of host])
+    ]
+  in
+  Uri.make ~path:Constants.get_updates_uri ~query () |> Uri.to_string
+
+let command_in_task ~rpc ~session_id ~fd ~host ~label f =
+  let task =
+    Client.Task.create ~rpc ~session_id
+      ~label:(Printf.sprintf "%s for host (ref=%s)" label (Ref.string_of host))
+      ~description:""
+  in
+  Client.Task.set_progress ~rpc ~session_id ~self:task ~value:(-1.0) ;
+  let command = f session_id task host in
+  finally
+    (fun () ->
+      marshal fd (Command command) ;
+      let response = ref (Response Wait) in
+      while !response = Response Wait do
+        response := unmarshal fd
+      done ;
+      let ok =
+        match !response with
+        | Response OK ->
+            true
+        | Response Failed ->
+            (* Need to check whether the thin cli managed to contact the server
+             * or not. If not, we need to mark the task as failed.
+             *)
+            if Client.Task.get_progress ~rpc ~session_id ~self:task < 0.0 then
+              Client.Task.set_status ~rpc ~session_id ~self:task ~value:`failure ;
+            false
+        | _ ->
+            false
+      in
+      wait_for_task_complete rpc session_id task ;
+      check_task_status ~rpc ~session_id ~task ~fd ~label ~ok
+        ~quiet_on_success:true ()
+    )
+    (fun () -> Client.Task.destroy ~rpc ~session_id ~self:task)
+
+let print_avail_updates ~rpc ~session_id ~fd ~host =
+  command_in_task ~rpc ~session_id ~fd ~host ~label:"Print available updates"
+    (fun session_id task host ->
+      PrintHttpGetJson (get_avail_updates_uri ~session_id ~task ~host)
+  )
+
 let host_apply_updates _printer rpc session_id params =
   let hash = List.assoc "hash" params in
   ignore
@@ -7655,6 +7709,15 @@ let host_apply_updates _printer rpc session_id params =
        )
        params ["hash"]
     )
+
+let host_updates_show_available fd _printer rpc session_id params =
+  do_host_op rpc session_id ~multiple:false
+    (fun _ host ->
+      let host = host.getref () in
+      print_avail_updates ~rpc ~session_id ~fd ~host
+    )
+    params []
+  |> ignore
 
 module SDN_controller = struct
   let introduce printer rpc session_id params =
--- a/ocaml/xapi-cli-server/cli_operations.ml
+++ b/ocaml/xapi-cli-server/cli_operations.ml
@@ -7696,19 +7696,31 @@ let print_avail_updates ~rpc ~session_id ~fd ~host =
       PrintHttpGetJson (get_avail_updates_uri ~session_id ~task ~host)
   )
 
-let host_apply_updates _printer rpc session_id params =
+let print_update_guidance ~rpc ~session_id ~fd ~host =
+  command_in_task ~rpc ~session_id ~fd ~host ~label:"Print update guidance"
+    (fun session_id task host ->
+      PrintUpdateGuidance (get_avail_updates_uri ~session_id ~task ~host)
+  )
+
+let host_apply_updates fd printer rpc session_id params =
   let hash = List.assoc "hash" params in
-  ignore
-    (do_host_op rpc session_id ~multiple:false
-       (fun _ host ->
-         let host = host.getref () in
-         Client.Host.apply_updates ~rpc ~session_id ~self:host ~hash
-         |> List.iter (fun l ->
-                _printer (Cli_printer.PMsg (String.concat "; " l))
-            )
-       )
-       params ["hash"]
+  do_host_op rpc session_id ~multiple:false
+    (fun _ host ->
+      let host = host.getref () in
+      printer (Cli_printer.PMsg "Guidance of updates:") ;
+      print_update_guidance ~rpc ~session_id ~fd ~host ;
+      printer (Cli_printer.PMsg "Applying updates ...") ;
+      match Client.Host.apply_updates ~rpc ~session_id ~self:host ~hash with
+      | [] ->
+          printer (Cli_printer.PMsg "Updated.")
+      | warnings ->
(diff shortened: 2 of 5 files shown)


ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -5462,34 +5462,13 @@ let wait_for_task_complete rpc session_id task_id =
...
-     	   to check that the server does too. *)
+let check_task_status ~rpc ~session_id ~task ~fd ~label ~ok =
+  (* if the client thinks it's ok, check that the server does too *)
   match Client.Task.get_status ~rpc ~session_id ~self:task with
   | `success ->
-      if ok then (
-        if filename <> "" then
-          marshal fd (Command (Print (Printf.sprintf "%s succeeded" label)))
-      ) else (
+      if ok then
```
lindig, line 5469: Just an idea but I like:
```
| `success when ok -> ..
| `success -> ..
| _ -> (* this is not handled in the existing code? *)
```

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -5513,6 +5492,32 @@ let download_file rpc session_id task fd filename uri label =
...
+  let ok =
+    match (!response, filename <> "") with
+    | Response OK, true ->
+        true
+    | Response OK, false ->
+        false
+    | Response Failed, _ ->
+        (* Need to check whether the thin cli managed to contact the server
+           				   or not. If not, we need to mark the task as failed *)
+        if Client.Task.get_progress ~rpc ~session_id ~self:task < 0.0 then
```
lindig, line 5510: Is `<` correct? Can this be negative? If not, it should be `<=`

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -7681,19 +7641,64 @@ let update_resync_host _printer rpc session_id params =
...
+let print_available_updates ~rpc ~session_id ~fd ~printer ~host =
+  let task =
+    Client.Task.create ~rpc ~session_id
+      ~label:
+        (Printf.sprintf "Get updates for host (ref=%s)" (Ref.string_of host))
+      ~description:""
+  in
+  Client.Task.set_progress ~rpc ~session_id ~self:task ~value:(-1.0) ;
+  let uri =
+    Printf.sprintf "%s?session_id=%s&task_id=%s&host_refs=%s"
```
lindig, line 7653: This is probably ok - but would fail if any of the strings would require escaping. In that case it would be better to use the URI module to construct this.

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -7681,19 +7641,64 @@ let update_resync_host _printer rpc session_id params =
...
+      Constants.get_updates_uri (Ref.string_of session_id) (Ref.string_of task)
+      (Ref.string_of host)
+  in
+  finally
+    (fun () ->
+      printer (Cli_printer.PMsg "Get available updates:") ;
+      marshal fd (Command (PrintHttpGetJson uri)) ;
+      let response = ref (Response Wait) in
+      while !response = Response Wait do
```
lindig, line 7662: Is this a spinning loop or is `unmarshal` blocking and this is not spinning?

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -7681,19 +7641,64 @@ let update_resync_host _printer rpc session_id params =
...
+        response := unmarshal fd
+      done ;
+      let ok =
+        match !response with
+        | Response OK ->
+            true
+        | Response Failed ->
+            if Client.Task.get_progress ~rpc ~session_id ~self:task < 0.0 then
```
lindig, line 7670: Again: `<=` is probably safer

lindig (commented):
I would have expected that more of the required functionality already exists. This needs a review from someone who know this area.

ocaml/xapi-cli-server/cli_operations.ml
Vincent-lau: It probably can be negative
```ocaml
(* Initially mark the task progress as -1.0. The first thing the HTTP handler does it to mark it as zero *)
  (* This is used as a flag to show that the 'ownership' of the task has been passed to the handler, and it's *)
  (* not our responsibility any more to mark the task as completed/failed/etc. *)
  Client.Task.set_progress ~rpc ~session_id ~self:task ~value:(-1.0) ;
 ```

Comment (minglumlu) [reply ocaml/xapi-cli-server/cli_operations.ml:5510]: So here the error category of "< 0.0" is that CLI has not contacted the server.

File: ocaml/xapi-cli-server/cli_operations.ml (hunk 4 above)
Comment (minglumlu) [reply ocaml/xapi-cli-server/cli_operations.ml:7662]: I think the `unmarshal` will block on reading. 
https://github.com/xapi-project/stdext/blob/7c5c22b94bab31eb2c498ef47e284391420b9c8e/lib/xapi-stdext-unix/unixext.ml#L491C22-L491C30
If it was in a non-blocking way, the `read_bytes` might be zero and actually not finished.

File: ocaml/xapi-cli-server/cli_operations.ml (hunk 5 above)
Comment (minglumlu) [reply ocaml/xapi-cli-server/cli_operations.ml:7670]: Similarly, this is for the error category that the CLI client has not contacted with the server. Added a comment for this.

**Comment by minglumlu** [comment]
Rebased with feature branch.
The examples of outputs:
```
# xe host-show-updates host=xrtmia-07-47
{
  "hosts": [
    {
      "ref": "OpaqueRef:9707cebd-0a2b-994f-48bf-1d5fa77788f0",
      "guidance": {
        "mandatory": [ "RestartToolstack", "EvacuateHost", "RestartVM" ],
        "recommended": [],
        "full": []
      },
      "RPMS": [ "test-update-0.6.0-1.xs8.x86_64.rpm" ],
      "updates": [ "CITRIX-HYPERVISOR-2021-0000" ],
      "livepatches": []
    }
  ],
  "updates": [
    {
      "id": "CITRIX-HYPERVISOR-2021-0000",
      "summary": "This is an update for testing only",
      "description": "This is an update for testing only",
      "special-info": "This is special update information to be acknowledged by user before applying this update",
      "URL": "https://ctx.citrix.com",
      "type": "Improvement",
      "issued": "20231213T01:59:16Z",
      "severity": "None",
      "livepatches": [],
      "guidance": {
        "mandatory": [ "RestartVM", "RestartToolstack", "EvacuateHost" ],
        "recommended": [],
        "full": [],
        "livepatch": []
      },
      "title": "Title"
    }
  ],
  "hash": "c891dd78f5aa9228811715c8abade849d5a280a951552bb431ee48fe2f8de062"
}

# xe host-apply-updates host=xrtmia-07-47 hash=c891dd78f5aa9228811715c8abade849d5a280a951552bb431ee48fe2f8de062
Guidance of updates:
{
  "mandatory": [ "RestartToolstack", "EvacuateHost", "RestartVM" ],
  "recommended": [],
  "full": []
}
Applying updates ...
Updated.
```

File: ocaml/xapi-cli-server/cli_frontend.ml [hunk 6]
```diff
@@ -1047,6 +1047,15 @@ let rec cmdtable_data : (string * cmd_spec) list =
       ; flags= [Neverforward]
       }
     )
+  ; ( "host-show-updates"
```
gangj, line 1050: I see `xe host-available-updates` in the design: https://code.citrite.net/projects/XSC/repos/xs-documents/browse/designs/on-prem-cdn-update/implementation.md?useDefaultHandler=true#379

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -7684,19 +7644,91 @@ let update_resync_host _printer rpc session_id params =
...
+              Client.Task.set_status ~rpc ~session_id ~self:task ~value:`failure ;
+            false
+        | _ ->
+            false
+      in
+      wait_for_task_complete rpc session_id task ;
+      check_task_status ~rpc ~session_id ~task ~fd ~label ~ok
+        ~quiet_on_success:true ()
+    )
+    (fun () -> Client.Task.destroy ~rpc ~session_id ~self:task)
```
gangj, line 7690: I see several pieces of similar task related code(search with `wait_for_task_complete`) in this file, we might be able to resue this `command_in_task` for other code, I had a rough try: https://github.com/xapi-project/xen-api/commit/154b7339660cbb83164d29945870c917e2cc717a, it passes "make check".

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -7684,19 +7644,91 @@ let update_resync_host _printer rpc session_id params =
...
-         |> List.iter (fun l ->
-                _printer (Cli_printer.PMsg (String.concat "; " l))
-            )
-       )
-       params ["hash"]
+  do_host_op rpc session_id ~multiple:false
+    (fun _ host ->
+      let host = host.getref () in
+      printer (Cli_printer.PMsg "Guidance of updates:") ;
+      print_update_guidance ~rpc ~session_id ~fd ~host ;
```
gangj, line 7710: Or print the guidance after applying updates?

ocaml/xapi-cli-server/cli_frontend.ml
minglumlu: I think `host-show-updates` complies with the convention of <object>-<verb>-<noun>.

ocaml/xe-cli/newcli.ml
```diff
@@ -435,6 +435,38 @@ let assert_filename_permitted ?(permit_cwd = false) permitted_filenames filename
...
+          (* see above about Unixfd.with_connection *)
+          close_in_noerr ic ; close_out_noerr oc ; doit newloc
+      | _ ->
+          failwith "Unhandled response code"
+    in
+    doit url
+  with
+  | ClientSideError msg ->
+      marshal ofd (Response Failed) ;
+      Printf.fprintf stderr "Operation failed. Error: %s\n" msg ;
```
gangj, line 464: Does the `msg` contains detailed info like `url`? If not then it might be helpful to log more details like:
`Printf.fprintf stderr "HTTP GET %s failed. Error: %s\n" url msg ;`

ocaml/xe-cli/newcli.ml
```diff
@@ -435,6 +435,38 @@ let assert_filename_permitted ?(permit_cwd = false) permitted_filenames filename
...
+      exit_code := Some 1
+  | e ->
+      debug "HTTP GET failure: %s\n%!" (Printexc.to_string e) ;
```
gangj, line 467: Same as above, log `url` might be helpful.

ocaml/xapi-cli-server/cli_frontend.ml
gangj: Agree, then how about `host-show-available-updates`? I think it will be easier to understand.

lindig: This is difficult to discover using auto complete. So I would suggest `host-updates-show-avaiable` The reason is that a user wants to know something about updates, so will start from `host-updates-…` to see available commands. The noun in this case is `host-updates`.

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -5462,17 +5462,49 @@ let wait_for_task_complete rpc session_id task_id =
...
   done ;
   let ok =
-    match !response with
-    | Response OK ->
+    match (!response, filename <> "") with
+    | Response OK, true ->
         true
-    | Response Failed ->
+    | Response OK, false ->
+        false
```
gangj, line 5506: Seems when filename is "", it will print to `stdout`: https://github.com/xapi-project/xen-api/blob/8ab1fe3340f79299204939ed7b2a1f6cde8e3d7a/ocaml/xe-cli/newcli.ml#L726-L727, so it is also supported then it is not a failure?

minglumlu: The code `newcli.ml` runs as `xe` client, and the code in `cli_operations.ml` runs in xapi as server.
In this case, the xapi receives the request from `xe` client, send `HttpGet` to the `xe` client, and then
1. the `xe` client opens `stdout` indeed,
2. but the xapi writes error to client and raise exception eventually.
 No behaviour being changed here. The xapi (server) determines if it is an error or not.

ocaml/xe-cli/newcli.ml
minglumlu: The `url` has been logged in line#442 and even earlier in a more general timing before matching `cmd` in the main loop.

ocaml/xe-cli/newcli.ml
minglumlu: The url has been logged in line#442 and even earlier in a more general timing before matching cmd in the main loop.

ocaml/xapi-cli-server/cli_operations.ml
minglumlu: The working pattern of`xe` CLI and xapi is:
1.`xe` CLI connect to xapi CLI handler
2. xapi CLI handler sends a `command` back to `xe` CLI 
3. in the `command` handler, `xe` CLI issues a XAPI call according to the `command`.

The `command` handler in `xe` CLI can't remember the state of the previoius commands. And after applying updates, the guidance from the available updates will be cleaned up in xapi.

ocaml/xapi-cli-server/cli_operations.ml
```diff
@@ -5462,17 +5462,49 @@ let wait_for_task_complete rpc session_id task_id =
     Thread.delay 1.0
   done
 
+let check_task_status ?(quiet_on_success = false) ~rpc ~session_id ~task ~fd
+    ~label ~ok () =
+  (* if the client thinks it's ok, check that the server does too *)
+  match Client.Task.get_status ~rpc ~session_id ~self:task with
+  | `success when ok ->
```
lindig, line 5469: Could be simplified to `when ok && not quiet_on_success ->`

ocaml/xe-cli/newcli.ml
```diff
@@ -435,6 +435,38 @@ let assert_filename_permitted ?(permit_cwd = false) permitted_filenames filename
   | _ ->
       ()
 
+let do_http_get ofd url exit_code f =
+  try
+    let rec doit url =
+      let server, path = parse_url url in
+      debug "Opening connection to server '%s' path '%s'\n%!" server path ;
```
lindig, line 442: Debug logs should avoid using `\n` - it will get escaped anyway.

ocaml/xe-cli/newcli.ml
```diff
@@ -435,6 +435,38 @@ let assert_filename_permitted ?(permit_cwd = false) permitted_filenames filename
...
+      with_open_tcp server @@ fun (ic, oc) ->
+      Printf.fprintf oc "GET %s HTTP/1.0\r\n\r\n" path ;
+      flush oc ;
+      (* Get the result header immediately *)
+      let resultline = input_line ic in
+      debug "Got %s\n%!" resultline ;
```
lindig, line 448: Same here

lindig (approved):

ocaml/xapi-cli-server/cli_frontend.ml
gangj: How about the `host-apply-updates`? Then rename it to `host-updates-apply`?

ocaml/xapi-cli-server/cli_operations.ml
gangj: Thanks for the explanation!
`No behaviour being changed here.`, I think there is some problem with the original code: https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi-cli-server/cli_operations.ml#L5490-L5491, seems there was no handling for the case: `ok && filename = ""`?

ocaml/xapi-cli-server/cli_frontend.ml
lindig: I would find that better discoverable; however, if the command already exists the change would break existing code. In that case having a second name could be a choice to improve discoverability.

gangj (approved):

gangj (approved):
I think the code change without this commit: https://github.com/xapi-project/xen-api/pull/5331/commits/a0b6821990db90411ce16e7f4977047a0ccd9db5 makes sense: `download_file` will raise error when the parameter `filename` is empty.
While with this commit, I think now the behavior is exactly the same as original: when `download_file` with empty filename, it will not fail with an error raised, it only keeps silence.

minglumlu:
I would like to keep the behaviour as before. Specifically, the output will reach `stdout` in `xe` CLI side. It's possible to redirect to other data sinks.

gangj:
> I would like to keep the behaviour as before. Specifically, in the case, the output will reach `stdout` in `xe` CLI side. It's possible to redirect to other data sinks.

Yes, you are right! We must keep the behavior as before.
Checked where `download_file` is used. Sometimes the `filename` is optional, for example: `pool-retrieve-wlb-report`, when `filename` is not provided, the report will be printed to stdout, and no error is expected.
I think the only small issue here is the confusing func name: `download_file`, maybe just `download` would be more appropriate.

minglumlu:
Merge this PR as `xe vm-export` and `xe snapshot-export-to-template` have been tested manually.

PR #5333, opened 2024-01-05
CA-386457: fix environment variable loading of `$PERFMON_FLAGS` in perfmon.service

duobei:
For example, we set the `$PERFMON_FLAGS` in scripts/sysconfig-perfmon:
```bash
PERFMON_FLAGS=""
PERFMON_FLAGS="$PERFMON_FLAGS --debug"
PERFMON_FLAGS="$PERFMON_FLAGS --interval=60"
PERFMON_FLAGS="$PERFMON_FLAGS --rrdstep=5"
```
What we ultimately want the value of `PERFMON_FLAGS` is `--debug --interval=60 --rrdstep=5`, then pass `PERFMON_FLAGS` to @OPTDIR@/bin/perfmon.

But there are some limits in EnvironmentFile of service in the [document of systemd](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#Environment):
> Variable expansion is not performed inside the strings and the "$" character has no special meaning.
>
> Settings from these files override settings made with Environment=. If the same variable is set twice from these files, the files will be read in the order they are specified and the later setting will override the earlier setting.

So if we configure the EnvironmentFile in perfmon.service, what we will only get is `$PERFMON_FLAGS --rrdstep=5` in fact.

### DIFF
--- a/scripts/sysconfig-perfmon
+++ b/scripts/sysconfig-perfmon
@@ -3,26 +3,22 @@
 # /etc/init.d/perfmon and used to construct the command line
 # arguments to pass to @OPTDIR@/bin/perfmon.
 
-# uncomment lines to enable options.
-
-PERFMON_FLAGS=""
-
 # send lots of debug to /var/log/messages
-#PERFMON_FLAGS="$PERFMON_FLAGS --debug"
+#PERFMON_FLAGS=" --debug"
 
 # change loop time from default of 5 minutes to 1 minute
-#PERFMON_FLAGS="$PERFMON_FLAGS --interval=60"
+#PERFMON_FLAGS=" --interval=60"
 
 # Tell perfmon to request "5 second averages" from XAPI.  By default
 # it requests "60 second averages" from XAPI.  The only valid options
 # are 5 or 60.
-#PERFMON_FLAGS="$PERFMON_FLAGS --rrdstep=5"
+#PERFMON_FLAGS=" --rrdstep=5"
 
 #####################################################################
 # Advanced
 
 # run only 10 loops then exit
-#PERFMON_FLAGS="$PERFMON_FLAGS --numloops=10"
+#PERFMON_FLAGS=" --numloops=10"
 
 # Re-read the perfmon configuration for the host and all VMs
 # every 10 minutes, instead of the default of every 1/2 hour.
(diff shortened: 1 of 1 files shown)


lindig:
Why can't we keep using systemd's EnvironmentFile but use supported syntax in that file? It seems to me that this is the best supported scenario (like when files are missing).

duobei:
> Why can't we keep using systemd's EnvironmentFile but use supported syntax in that file? It seems to me that this is the best supported scenario (like when files are missing).

It couldn't get we want.

For example we execute those in bash:
```shell
PERFMON_FLAGS=""
PERFMON_FLAGS="$PERFMON_FLAGS --debug"
PERFMON_FLAGS="$PERFMON_FLAGS --interval=60"
PERFMON_FLAGS="$PERFMON_FLAGS --rrdstep=5"

echo $PERFMON_FLAGS
``` 
It will get the value:
```shell
--debug --interval=60 --rrdstep=5
``` 
We configured `EnvironmentFile=-/etc/sysconfig/perfmon` in the service file and put below in `/etc/sysconfig/perfmon`
:
```shell
PERFMON_FLAGS=""
PERFMON_FLAGS="$PERFMON_FLAGS --debug"
PERFMON_FLAGS="$PERFMON_FLAGS --interval=60"
PERFMON_FLAGS="$PERFMON_FLAGS --rrdstep=5"
``` 
in order to get the same result `--debug --interval=60 --rrdstep=5` in above.
But it only gets `$PERFMON_FLAGS --rrdstep=5`  finally. 
The reasons are: 
> Variable expansion is not performed inside the strings and the "$" character has no special meaning.
> 
> Settings from these files override settings made with Environment=. If the same variable is set twice from these files, the files will be read in the order they are specified and the later setting will override the earlier setting.

The related document at https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#Environment.

lindig:
I understand that. Why can't we have a file that sets the variable in one line and does not rely on expansion?

duobei:
> I understand that. Why can't we have a file that sets the variable in one line and does not rely on expansion?

It's a good idea. I would try in this way.

lindig (approved):

minglumlu (approved):

PR #5334, opened 2024-01-05
OIL: introduce internal_error(fmt) for error reporting

lindig:
Simplify the logging and reporting of internal errors in xenopsd. Rather than constructing a complex exception use a function that accepts a printf-style string for convenience. It logs the error and raises the exception.

### DIFF
--- a/ocaml/xenopsd/lib/xenops_server.ml
+++ b/ocaml/xenopsd/lib/xenops_server.ml
@@ -23,6 +23,14 @@ module D = Debug.Make (struct let name = "xenops_server" end)
 
 open D
 
+let internal_error fmt =
+  Printf.kprintf
+    (fun str ->
+      error "%s" str ;
+      raise (Xenopsd_error (Internal_error str))
+    )
+    fmt
+
 let rpc_of ty x = Rpcmarshal.marshal ty.Rpc.Types.ty x
 
 let finally = Xapi_stdext_pervasives.Pervasiveext.finally
@@ -88,9 +96,6 @@ let filter_prefix prefix xs =
         None
     )
     xs
-
-let internal_error fmt =
-  Printf.kprintf (fun msg -> raise (Xenopsd_error (Internal_error msg))) fmt
 
 (* return last/element/in/a/path *)
 let basename path =
@@ -647,10 +652,7 @@ module VGPU_DB = struct
     | [a; b] ->
         (a, b)
     | _ ->
-        raise
-          (Xenopsd_error
-             (Internal_error ("String cannot be interpreted as vgpu id: " ^ str))
-          )
+        internal_error "String cannot be interpreted as vgpu id: %s" str
 
   let ids vm : Vgpu.id list =
     list [vm] |> filter_prefix "vgpu." |> List.map (fun id -> (vm, id))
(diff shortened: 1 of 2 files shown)


ocaml/xenopsd/lib/xenops_server.ml
```diff
@@ -2731,10 +2727,9 @@ and perform_exn ?subtask ?result (op : operation) (t : Xenops_task.task_handle)
                    the destination host. even though the destination host
                    failed to respond successfully to our handshake, the VM
                    should still be running correctly *)
-                error
+                internal_error
```
robhoes, line 2730: This changes the string of the exception compared to the old code. But it just adds more info to the message, which should be fine.

ocaml/xenopsd/xc/xenops_server_xen.ml
```diff
@@ -247,12 +255,7 @@ let di_of_uuid ~xc uuid =
...
       let uuid' = Uuidx.to_string uuid in
-      let err_msg =
-        Printf.sprintf "More than one domain with uuid %s: (%s)" uuid'
-          domid_list
-      in
-      error "%s: %s" __FUNCTION__ err_msg ;
-      raise (Xenopsd_error (Internal_error err_msg))
+      internal_error "More than one domain with uuid %s: (%s)" uuid' domid_list
   | exception Failure r ->
       raise (Xenopsd_error (Internal_error r))
```
robhoes, line 260: Do this one as well?

robhoes (approved):

ocaml/xenopsd/xc/xenops_server_xen.ml
```diff
@@ -30,6 +30,14 @@ let finally = Xapi_stdext_pervasives.Pervasiveext.finally
 
 let with_lock = Xapi_stdext_threads.Threadext.Mutex.execute
 
+let internal_error fmt =
```
edwintorok, line 33: You have 2 copies of this, I assume because the 'error' function is different (it includes the module name).

edwintorok (approved):

Vincent-lau (approved):
The original code has lots of duplications and nestings 🙅 

One question: if we raise, then that gets logged anyway, do we need to err to log and then raise as well?

ocaml/xenopsd/xc/xenops_server_xen.ml
lindig: Indeed; otherwise you need to make sure it's logged under the correct name.

lindig:
In my new code I log the error just to be very close to the origin of that error and because I don't want to rely on top-level code that might or might not log the error. But you are right that we could decide not to log at this point.

lindig:
I was too eager in one case and changed the exception - fixed that back to the original code:

```
-            let m =
-              Printf.sprintf
-                "VM = %s; domid = %d; Bootloader.Error_from_bootloader %s"
-                vm.Vm.id domid x
-            in
-            debug "%s" m ;
-            raise (Xenopsd_error (Bootloader_error (vm.Vm.id, x)))
+            internal_error
+              "VM = %s; domid = %d; Bootloader.Error_from_bootloader %s"
+              vm.Vm.id domid x
```

PR #5335, opened 2024-01-08
CP-43875: Record the repository hash on the host object when updating

gangj:
Add field "last_update_hash" on host object to record the SHA256 checksum of updateinfo of the most recently applied update on the host.

### DIFF
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -2176,6 +2176,10 @@ let t =
              user should follow to make some updates, e.g. specific hardware \
              drivers or CPU features, fully effective, but the 'average user' \
              doesn't need to"
+        ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:String
+            ~default_value:(Some (VString "")) "last_update_hash"
+            "The SHA256 checksum of updateinfo of the most recently applied \
+             update on the host"
         ]
       )
     ()
--- a/ocaml/idl/schematest.ml
+++ b/ocaml/idl/schematest.ml
@@ -2,7 +2,7 @@ let hash x = Digest.string x |> Digest.to_hex
 
 (* BEWARE: if this changes, check that schema has been bumped accordingly in
    ocaml/idl/datamodel_common.ml, usually schema_minor_vsn *)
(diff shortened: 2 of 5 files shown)


ocaml/idl/datamodel_host.ml
```diff
@@ -2176,6 +2176,10 @@ let t =
              user should follow to make some updates, e.g. specific hardware \
              drivers or CPU features, fully effective, but the 'average user' \
              doesn't need to"
+        ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:String
+            ~default_value:(Some (VString "")) "last_update_hash"
```
lindig, line 2180: I am not a native speaker but wonder if "latest_update_hash" would be better.

ocaml/idl/datamodel_host.ml
```diff
@@ -2176,6 +2176,10 @@ let t =
...
+            "The SHA256 checksum of updateinfo of the most recently applied \
```
lindig, line 2181: An update that has been applied ma still not be active because it might require a restart of some kind. I assume we updating this when applying the update even when this may mean that it is not yet active.

lindig (approved):

robhoes (approved):

ocaml/idl/datamodel_host.ml
minglumlu: Yes. The "not yet active" is tracked by pending guidance. Overall, the "host.last_update_hash" + host pending guidance lists could reflect the status of the update.

ocaml/idl/schematest.ml
```diff
@@ -2,7 +2,7 @@ let hash x = Digest.string x |> Digest.to_hex
 
 (* BEWARE: if this changes, check that schema has been bumped accordingly in
    ocaml/idl/datamodel_common.ml, usually schema_minor_vsn *)
-let last_known_schema_hash = "9c650ad57273c375b9f82f26f82aa75f"
+let last_known_schema_hash = "a2378b3ff9c452cb61f0cab7d9f84a46"
```
psafont, line 5: This was merged without a bump to schema_minor_vsn, this will need to be done when merging to master

PR #5337, opened 2024-01-08
Call resume fast if an exception occurs during migration

gthvn1:
Hello,

I have a look to use domain resume fast. So if an exception is raised when performing the migration of a VM we call resume fast to resume the VM on the original host if the VM is in suspended state. Based on some previous work done in #4746 I started to write this piece of code. I have some issue for testing it. I tried to introduce an error in the handshake during the migration but unfortunately the VM resume on the original host without the need to trigger the resume fast. To be more accurate I just replace the success arm of the match of the [point 4 of the synchronisation](https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/lib/xenops_server.ml#L2723) to also raise an exception. The migration failed as expected but the VM was just still running on the original host without calling fast resume. I guess that it is not the right way to test it and I just de-synchronized both hosts...

Any comments about the patch is welcome but also a way for testing it will be welcome :)

### DIFF
--- a/ocaml/idl/datamodel_vm.ml
+++ b/ocaml/idl/datamodel_vm.ml
@@ -827,6 +827,22 @@ let unpause =
       ]
     ~allowed_roles:_R_VM_OP ()
 
+(* VM.ResumeFast *)
+
+let resumeFast =
+  call ~in_product_since:rel_rio ~name:"resume_fast"
+    ~doc:
+      "Resume the specified VM cooperatively. This can only be called when the \
+       specified VM is in the Suspended state."
+    ~params:[(Ref _vm, "vm", "The VM to resume fast")]
+    ~errs:
+      [
+        Api_errors.vm_bad_power_state
+      ; Api_errors.operation_not_allowed
+      ; Api_errors.vm_is_template
+      ]
+    ~allowed_roles:_R_VM_OP ()
+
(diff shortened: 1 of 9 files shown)


lindig:
Fast resume on migration failure is a desirable feature - than`ks for tackling it. I assume @robhoes might have ideas for how to test this. For conditional test code you might want to look at `ocaml/xapi/xapi_fist.ml` and how it is used in existing code.

robhoes:
Yes I tried something like this myself last year, but didn't get round to finishing it. The error handling for live migration is unfortunately quite complicated as errors can occur at many points in the process, and a different kind of clean-up needs to happen in each case. Right now, it simply shuts down everything related to the running VM for every error. I meant to get back to this at some point...

robhoes:
This needs more investigation on error handling for live migrations in xenopsd, which I am planning to do. It's unfortunately more complex than I had hoped. I'd close this PR for now, but will get back to it once I know more.

psafont:
> Yes I tried something like this myself last year, but didn't get round to finishing it.

@robhoes Is there a public branch that somebody can take and continue the work on?

robhoes:
I don't think I had pushed it to github. I'll have a look through my local branches at some point.

lindig:
🚧  Expect excavation work in this area ⚠️

PR #5338, opened 2024-01-08
[OIL] Split the API reference markdown into smaller files and use templates to generate it.

kc284:
The API reference file has become too big (~1.5MB) and keeps increasing. The result is that the corresponding webpage takes several seconds to download, so we decided to split it to smaller bits and have a file per class.


ocaml/doc/basics.md
```diff
@@ -1,3 +1,7 @@
+---
```
kc284, line 1: These headers are a particularity of XS docs. If there is an issue with having them here, I can remove them and have the pipeline insert them later (since it will have to replace the `@root@` placeholders in some files anyway).

ocaml/idl/autogen/api-ref-autogen.md
```diff
@@ -0,0 +1,12 @@
+---
+ layout: doc
+---
+
+# API Reference
+
+Version **@xapi-version@**
```
kc284, line 7: Is there an easy way to have this replaced during build time? If not, I can have the pipeline replace it later).

kc284:
Note that this PR does not necessarily need Interfaces reviewers.

ocaml/idl/autogen/api-ref-autogen.md
psafont: Currently, the version is embedded into binaries using the `Xapi_version` module and using the `dune-build-info` library. One way to do it would be to it would be to create an ocaml binary that processes this file.

One way you can try is to use `dune subst`, and use `%%VERSION%%`. I believe it's intended to be used project-wide, but you're welcome ot try

kc284: Thanks for the tip; I won't consider this as a blocker for this PR though (it's probably not a big change, but I only have time to work on this on and off, so I'd rather get the PR in the soonest possible and improve the versioning later).

psafont: Had a look, and the binary that generated the files already has access to this library, it should as easy as generating the markdown file from a mustache template to fill in the version... but for some reason I keep getting 0.0.0.dev :(

https://github.com/xapi-project/xen-api/commit/55a970baf71e44571b1eff72018bbb100ab667e9

kc284: As discussed, building in the CI yields `0.0.dev`. In the interest of avoiding taking up too much of people's time, I can always leave it as is and replace it later in the pipeline alongside `@root@`.

ocaml/idl/autogen/dune
```diff
@@ -0,0 +1,6 @@
+(alias
```
edwintorok, line 1: this is an alias with a name and dependency, but where is the action?

Makefile
```diff
@@ -42,15 +42,20 @@ schema:
...
+#html
 	dune build --profile=$(PROFILE) -f @ocaml/doc/jsapigen
 	mkdir -p $(XAPIDOC)/html
 	cp -r _build/default/ocaml/doc/api $(XAPIDOC)/html
 	cp _build/default/ocaml/doc/branding.js $(XAPIDOC)/html
 	cp ocaml/doc/*.js ocaml/doc/*.html ocaml/doc/*.css $(XAPIDOC)/html
-	dune exec --profile=$(PROFILE) -- ocaml/idl/datamodel_main.exe -closed -markdown $(XAPIDOC)/markdown
-	cp ocaml/doc/*.dot ocaml/doc/doc-convert.sh $(XAPIDOC)
+#markdown
+	dune build --profile=$(PROFILE) -f @ocaml/idl/markdowngen
```
edwintorok, line 52: we could define some install sections in dune and let dune install things into the right places. The method of installing files in the Makefile was from the time when we didn't have dune.
Then also the version substitutions inside the markdown files might work

edwintorok: See https://dune.readthedocs.io/en/stable/dune-files.html#install

ocaml/idl/autogen/dune
kc284: No action, just using `source_tree` to include all the files that will be generated in here. We have a few more dune files with rules like this.

Makefile
kc284: Unless I'm missing something, it doesn't seem to me that I can do this efficiently with the version of dune we're currently using. All the useful stanzas (source_trees, glob_files, include) have been added in various 3.x versions.The install sections I see in other dune files in the repo handle single files, but here there is a good number of autogenerated files that have to be installed.

ocaml/idl/markdown_backend.ml
```diff
@@ -152,231 +147,393 @@ let string_of_qualifier = function
...
+let string_of_file filename =
+  let io = open_in filename in
+  Fun.protect
+    (fun () ->
+      let rec read_lines acc =
+        try read_lines (input_line io :: acc) with End_of_file -> acc
+      in
+      read_lines [] |> List.rev |> String.concat "\n"
+    )
+    ~finally:(fun () -> close_in io)
```
psafont, line 159:
```suggestion
let string_of_file = Xapi_stdext_Unix.Unixext.string_of_file
```

psafont:
Is there a way to check the output of this PR?

For now I can see that classes.png is not being generated from classes.dot

psafont:
How nice, that hasn't happened since before the introduction of GPU datatypes :(

Makefile
```diff
@@ -42,15 +42,20 @@ schema:
...
+	mkdir -p $(XAPIDOC)/markdown
+	cp -r _build/default/ocaml/idl/autogen/*.md $(XAPIDOC)/markdown
+	cp -r _build/default/ocaml/idl/autogen/*.yml $(XAPIDOC)/markdown
 	find ocaml/doc -name "*.md" -not -name "README.md" -exec cp {} $(XAPIDOC)/markdown/ \;
+#other
+	cp ocaml/doc/*.dot ocaml/doc/doc-convert.sh $(XAPIDOC)
```
psafont, line 58:
```suggestion
	cp ocaml/doc/*.dot ocaml/doc/doc-convert.sh $(XAPIDOC)
	cd _build/install/default/xapi/doc; ./doc-convert.sh
```

kc284: The output of this PR is in the xapi-doc rpm. Classes.png was never generated automatically. In the past this repo contained the pngs, but they were generated by running the script manually on demand, it was not part of the (make) build. Therefore I removed the pngs and left just the dot files; now the build just packages the dot files in the rpm, and the pngs are created by running the script in the pipeline that packages the API reference for delivery to the main product documentation.  If changes are needed in this workflow I'd rather make them in a different PR - I'm keen on making only the necessary changes needed to merge this one because there is further work depending on it.

psafont (approved):

ocaml/idl/markdown_backend.ml
kc284: Fixed [here](https://github.com/xapi-project/xen-api/compare/8db76b33d881f818bd65da65d47d676eb74f8e87..36219ef3c573ee3c5e8cb7d9137e06996e84f36e)

ocaml/doc/basics.md
danilo-delbusso, line 1: existing issue but since you're here, on line `13` it should read "recognized" instead of "recognised"

danilo-delbusso (changes requested):
One last comment. This may need input from other xapi devs:

[`Certificate`'s description is just the word "Description"](https://github.com/xapi-project/xen-api/blob/master/ocaml/idl/datamodel_certificate.ml#L37) and [`Bond` doesn't even have one](https://github.com/xapi-project/xen-api/blob/master/ocaml/idl/datamodel.ml#L2060).

We could add them in this PR, maybe?

I was thinking @psafont or @edwintorok could weigh in on a proper description

ocaml/idl/templates/relationships.mustache
```diff
@@ -0,0 +1,19 @@
...
+|:---------------------------------------|:---------------------------------------|:--------------|
+{{#relationships}}
+|{{{a_field}}}|{{{b_field}}}|{{relationship}}|
+{{/relationships}}
+
+The following figure represents bound fields (as specified above)
+diagramatically, using crow's foot notation to specify one-to-one,
+one-to-many, or many-to-many relationships:
+
+![Class relationships](classes.png 'Class relationships')
```
danilo-delbusso, line 19: I'm a bit confused, where is this image in the source? I can see it on https://xapi-project.github.io/xen-api/ but not here. :/

danilo-delbusso (changes requested):

psafont:
> One last comment. This may need input from other xapi devs:
> 
> [`Certificate`'s description is just the word "Description"](https://github.com/xapi-project/xen-api/blob/master/ocaml/idl/datamodel_certificate.ml?rgh-link-date=2024-03-08T10%3A14%3A19Z#L37) and [`Bond` doesn't even have one](https://github.com/xapi-project/xen-api/blob/master/ocaml/idl/datamodel.ml?rgh-link-date=2024-03-08T10%3A14%3A19Z#L2060).
> 
> We could add them in this PR, maybe?

For Certificate, the description can be "An X509 certificate used for TLS connections". For Bond, maybe "A Network bond that combines physical network interfaces, also known as link aggregation"?

ocaml/idl/templates/relationships.mustache
kc284: It's autogenerated from the .dot file. The internal pipeline does it, I haven't looked into github actions because that was done a long time ago (and I'll consider it out of scope for this PR).

ocaml/idl/templates/api_errors.mustache
```diff
@@ -0,0 +1,136 @@
...
+content-length: 297
+content-type:text/html
+connection:close
+cache-control:no-cache, no-store
+
+<html><body><h1>HTTP 500 internal server error</h1>An unexpected error occurred;
+ please wait a while and try again. If the problem persists, please contact your
+ support representative.<h1> Additional information </h1>Xmlrpc.Parse_error(&quo
+t;close_tag&quot;, &quot;open_tag&quot;, _)</body></html>
+```
```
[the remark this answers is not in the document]
kc284: First of all, the above has a syntax error: fenced code blocks should be surrounded by blank lines. Beyond that, we can split request and response, but one cannot have two fenced code blocks appearing as one and we shouldn't split the response in two even if it's not highlighted properly.

ocaml/idl/templates/class.mustache
```diff
@@ -0,0 +1,89 @@
...
+
+_Overview:_
+{{#rpc_has_descr}}
+
+{{{rpc_descr}}}
+{{/rpc_has_descr}}
+
+_Signature:_
+
+```
```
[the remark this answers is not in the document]
kc284: I think this is just pseudocode, there isn't a proper syntax highlight for it.

ocaml/idl/templates/class.mustache
```diff
@@ -0,0 +1,89 @@
...
+{{{return_type}}} {{{rpc_name}}} ({{#session}}session ref session_ref{{#has_rpc_params}}, {{/has_rpc_params}}{{/session}}{{{rpc_param_csv}}})
+```
+
+_Arguments:_
+
+|type                          |name                          |description                             |
+|:-----------------------------|:-----------------------------|:---------------------------------------|
+{{#session}}
+|session ref                   |session_ref                   |Reference to a valid session            |
```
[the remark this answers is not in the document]
kc284: "Session ref" is on purpose not wrapped in `` because it is not the name of a type, but rather a description of the type.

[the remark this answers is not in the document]
kc284: Ah, the autogenerated "X ref" instances are wrapped in `` indeed. Ok, I will correct it.

ocaml/idl/markdown_backend.ml
```diff
@@ -152,231 +148,383 @@ let string_of_qualifier = function
...
+    ~finally:(fun () -> close_out io)
 
-let is_deprecation_marker x =
-  match x with Lifecycle.Deprecated, _, _ -> true | _ -> false
+let generate_class cls =
+  let class_json =
+    `O
+      [
+        ("class_name", `String (escape cls.name))
+      ; ("class_descr", `String (escape cls.description))
```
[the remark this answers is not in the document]
kc284: Some times they contain special characters which might create issues in the generated markdown, for example, they may contain underscores which markdown interpretes as italics notation.

ocaml/idl/templates/types.mustache
```diff
@@ -0,0 +1,40 @@
...
+
+The following enumeration types are used:
+{{#enums}}
+
+|enum {{{enum}}}|                                        |
+|:---------------------------------------|:---------------------------------------|
+{{#enum_options}}
+|{{{option_name}}}|{{{option_descr}}}|
+{{/enum_options}}
+{{/enums}}
```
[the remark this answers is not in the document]
kc284: Happy to do this. A related issue: It would be nice if the enums appeared also on the right hand side navigation bar, but I think it doesn't show level ### headers. I thought of splitting the enums in a separate file with header "Types - Enums", but this leaves the primitive and high order types on their own and I can't think of a good header describing both - any ideas?

kc284:
I also rebased on master. The changes requested are in a separate commit, see here https://github.com/xapi-project/xen-api/pull/5338/commits/37821d8828a8366c79676027cc73ffac440e01c8

ocaml/idl/templates/class.mustache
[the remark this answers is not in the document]
danilo-delbusso: +1 but I like some highlighting if possible - though it's just a preference of mine so up to you

ocaml/idl/templates/types.mustache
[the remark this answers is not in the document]
danilo-delbusso: best I could think of is to add a table of contents before listing all of them but I think this is a question best directed to docs, maybe?

danilo-delbusso (approved):

PR #5339, opened 2024-01-08
CP-47153 add task list to bugtool

lindig:
The xapi database holds task but this table is not persisted. In conjuction with XSI-1559 it would be useful to have the currently active tasks. Add the output of "xe task-list" to the bugtool.

This was manually tested. In the typical case the task list is empty. So this adds no weight to the bugtool.


robhoes (approved):

Vincent-lau (approved):
stuff.xml 😆

PR #5340, opened 2024-01-08
CP-46323: Expose quorum and cluster membership through the API

Vincent-lau:
Results look like this:
```bash
$ xe cluster-param-list uuid=690d6398-9f00-9ca9-497d-9ad5d700cdfd    
uuid ( RO)                         : 690d6398-9f00-9ca9-497d-9ad5d700cdfd
                cluster-hosts (SRO): 190aff88-c954-4286-ead7-7c81b3a00bd3; 81efff95-1baa-2801-3397-2b5a4b8c7ec4; 160139e0
-4dc2-be55-3cc8-02812536cd9a
                cluster-token ( RO): b03a7918-2ca7-4d9f-a1a1-5d5391eeaba1
                cluster-stack ( RO): corosync
                token-timeout ( RO): 20.000
    token-timeout-coefficient ( RO): 1.000
           allowed-operations (SRO): destroy; disable; enable; remove; add
           current-operations (SRO):
               pool-auto-join ( RO): true
               cluster-config (MRO):
                 other-config (MRW):
+                   is_quorate ( RO): true
+               quorum_members (SRO): 10.62.49.47; 10.62.49.39; 10.62.49.35
+                       quorum ( RO): 2
+                  total_votes ( RO): 3
```

### DIFF
--- a/ocaml/idl/datamodel_cluster.ml
+++ b/ocaml/idl/datamodel_cluster.ml
@@ -194,6 +194,15 @@ let t =
              (Some (VString Constants.default_smapiv3_cluster_stack))
            "Simply the string 'corosync'. No other cluster stacks are \
             currently supported"
+       ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:Bool "is_quorate"
+           ~default_value:(Some (VBool false))
+           "Whether the cluster stack thinks the cluster is quorate"
+       ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:Int "quorum"
+           ~default_value:(Some (VInt 0L))
+           "Number of live hosts in order to be quorate"
+       ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:Int "live_hosts"
+           ~default_value:(Some (VInt 0L))
+           "Current number of live hosts, according to the cluster stack"
        ]
       @ allowed_and_current_operations cluster_operation
       @ [
--- a/ocaml/idl/datamodel_cluster_host.ml
+++ b/ocaml/idl/datamodel_cluster_host.ml
@@ -103,17 +103,24 @@ let t =
            "Reference to the Cluster object"
        ; field ~qualifier:StaticRO ~lifecycle ~ty:(Ref _host) "host"
            ~default_value:(Some (VRef null_ref)) "Reference to the Host object"
-       ; field ~qualifier:StaticRO ~lifecycle ~ty:Bool "enabled"
+       ; field ~qualifier:DynamicRO ~lifecycle ~ty:Bool "enabled"
            ~default_value:(Some (VBool false))
            "Whether the cluster host believes that clustering should be \
-            enabled on this host"
+            enabled on this host. This field can be altered by calling the \
+            enable/disable message on a cluster host. Only enabled members run \
+            the underlying cluster stack. Disabled members are still \
+            considered a member of the cluster (see joined), and can be \
+            re-enabled by the user."
        ; field ~qualifier:StaticRO ~lifecycle ~ty:(Ref _pif) "PIF"
            ~default_value:(Some (VRef null_ref)) "Reference to the PIF object"
-       ; field ~qualifier:StaticRO ~lifecycle ~ty:Bool "joined"
+       ; field ~qualifier:DynamicRO ~lifecycle ~ty:Bool "joined"
            ~default_value:(Some (VBool true))
-           "Whether the cluster host has joined the cluster"
-         (* TODO: add `live` member to represent whether corosync believes that this
-                  cluster host actually is enabled *)
+           "Whether the cluster host has joined the cluster. Contrary to \
+            enabled, a host that is not joined is not considered a member of \
+            the cluster, and hence no operations (e.g. enable/disable) can be \
+            performed on this host. This field can be altered by calling leave \
+            or destroy on a cluster host. It can also be set automatically if \
+            cluster stack believes that this node is not part of the cluster. "
        ]
       @ allowed_and_current_operations cluster_host_operation
       @ [
--- a/ocaml/idl/datamodel_cluster_host.ml
+++ b/ocaml/idl/datamodel_cluster_host.ml
@@ -121,6 +121,15 @@ let t =
             performed on this host. This field can be altered by calling leave \
             or destroy on a cluster host. It can also be set automatically if \
             cluster stack believes that this node is not part of the cluster. "
+       ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:Bool "live"
+           ~default_value:(Some (VBool false))
+           "Whether the underlying cluster stack thinks we are live. This \
+            field is set automatically based on updates from the cluster stack \
+            and cannot be altered by the user."
+       ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:DateTime
+           "last_update_live" ~default_value:(Some (VDateTime Date.epoch))
+           "Time when the live field was last updated based on information \
+            from the cluster stack"
        ]
       @ allowed_and_current_operations cluster_host_operation
       @ [
--- a/ocaml/idl/datamodel_lifecycle.ml
+++ b/ocaml/idl/datamodel_lifecycle.ml
@@ -21,6 +21,16 @@ let prototyped_of_field = function
       Some "23.14.0"
   | "Repository", "gpgkey_path" ->
       Some "22.12.0"
+  | "Cluster_host", "last_update_live" ->
+      Some "24.2.1-next"
+  | "Cluster_host", "live" ->
+      Some "24.2.1-next"
+  | "Cluster", "live_hosts" ->
+      Some "24.2.1-next"
+  | "Cluster", "quorum" ->
+      Some "24.2.1-next"
+  | "Cluster", "is_quorate" ->
+      Some "24.2.1-next"
   | "VTPM", "contents" ->
       Some "22.26.0"
   | "VTPM", "is_protected" ->
--- a/ocaml/idl/datamodel_lifecycle.ml
+++ b/ocaml/idl/datamodel_lifecycle.ml
@@ -30,7 +30,7 @@ let prototyped_of_field = function
   | "VTPM", "persistence_backend" ->
       Some "22.26.0"
   | "host", "numa_affinity_policy" ->
-      Some "23.32.0-next"
+      Some "24.2.1"
   | "host", "latest_synced_updates_applied" ->
       Some "23.18.0"
   | "host", "recommended_guidances" ->
@@ -66,7 +66,7 @@ let prototyped_of_field = function
   | "pool", "migration_compression" ->
       Some "22.33.0"
   | "pool", "custom_uefi_certificates" ->
-      Some "23.32.0-next"
+      Some "24.2.1"
   | _ ->
       None
 
@@ -104,7 +104,7 @@ let prototyped_of_message = function
   | "host", "set_https_only" ->
       Some "22.27.0"
   | "host", "set_numa_affinity_policy" ->
-      Some "23.32.0-next"
+      Some "24.2.1"
   | "VM", "restart_device_models" ->
       Some "23.30.0"
   | "pool", "set_ext_auth_max_threads" ->
@@ -122,6 +122,6 @@ let prototyped_of_message = function
   | "pool", "set_https_only" ->
       Some "22.27.0"
   | "pool", "set_custom_uefi_certificates" ->
-      Some "23.32.0-next"
+      Some "24.2.1"
   | _ ->
       None
--- a/ocaml/idl/schematest.ml
+++ b/ocaml/idl/schematest.ml
@@ -3,7 +3,7 @@ let hash x = Digest.string x |> Digest.to_hex
 (* BEWARE: if this changes, check that schema has been bumped accordingly in
    ocaml/idl/datamodel_common.ml, usually schema_minor_vsn *)
 
-let last_known_schema_hash = "ab570d662d657ee2055194110d536302"
+let last_known_schema_hash = "f10afca39b85b1c9fd7554b09f749fa5"
 
 let current_schema_hash : string =
   let open Datamodel_types in
--- a/ocaml/idl/schematest.ml
+++ b/ocaml/idl/schematest.ml
@@ -3,7 +3,7 @@ let hash x = Digest.string x |> Digest.to_hex
 (* BEWARE: if this changes, check that schema has been bumped accordingly in
    ocaml/idl/datamodel_common.ml, usually schema_minor_vsn *)
 
-let last_known_schema_hash = "f10afca39b85b1c9fd7554b09f749fa5"
+let last_known_schema_hash = "a3ddab734ad88468120269cf3d0e3bd9"
 
 let current_schema_hash : string =
   let open Datamodel_types in
--- a/ocaml/tests/test_cluster.ml
+++ b/ocaml/tests/test_cluster.ml
@@ -24,6 +24,8 @@ let test_clusterd_rpc ~__context call =
         {success= true; contents= Rpc.String test_token; is_notification= false}
   | ("enable" | "disable" | "destroy" | "leave" | "set-tls-verification"), _ ->
       Rpc.{success= true; contents= Rpc.Null; is_notification= false}
+  | "UPDATES.get", _ ->
+      Rpc.{success= true; contents= Rpc.Null; is_notification= false}
   | ( ( "Observer.create"
       | "Observer.destroy"
       | "Observer.set_enabled"
--- a/ocaml/xapi-idl/cluster/cluster_interface.ml
+++ b/ocaml/xapi-idl/cluster/cluster_interface.ml
@@ -284,6 +284,21 @@ module LocalAPI (R : RPC) = struct
       @-> returning unit_p err
       )
 
+  module UPDATES = struct
+    open TypeCombinators
+
+    let get =
+      let timeout_p = Param.mk ~name:"timeout" Types.float in
+      let result_p = Param.mk ~name:"updates" (list Types.string) in
+      declare "UPDATES.get"
+        [
+          "Get updates from corosync-notifyd, this blocking call will return \
+           when there is an update from corosync-notifyd or it is timed out \
+           after timeout_p seconds"
+        ]
+        (debug_info_p @-> timeout_p @-> returning result_p err)
+  end
+
   module Observer = struct
     open TypeCombinators
 
--- a/ocaml/xapi/xapi_cluster_host.ml
+++ b/ocaml/xapi/xapi_cluster_host.ml
@@ -194,6 +194,8 @@ let resync_host ~__context ~host =
             (* Note that join_internal and enable both use the clustering lock *)
             Client.Client.Cluster_host.enable ~rpc ~session_id ~self
           ) ;
+          (* create the watcher here so that the watcher exists after toolstack restart *)
+          create_cluster_watcher_on_master ~__context ~host ;
           Xapi_observer.initialise_observer ~__context
             Xapi_observer.Component.Xapi_clusterd ;
           let verify = Stunnel_client.get_verify_by_default () in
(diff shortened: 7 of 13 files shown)


ocaml/xapi-idl/cluster/cluster_interface.ml
```diff
@@ -284,6 +284,23 @@ module LocalAPI (R : RPC) = struct
...
+  module UPDATES = struct
+    let get =
+      let timeout_p = Param.mk ~name:"timeout" Types.float in
+      let result_p = Param.mk ~name:"updates" Types.string in
+      declare "UPDATES.get"
+        [
+          "Get updates from corosync-notifyd, this blocking call will return \
+           when there is an update from corosync-notifyd or it is timed out"
+        ]
+        (debug_info_p @-> timeout_p @-> returning result_p err)
```
lindig, line 296: I suggests to include the unit of the timeout in the documentation. I also assume the timeout is a time span - not a point in time.

ocaml/xapi/xapi_clustering.ml
```diff
@@ -401,3 +403,68 @@ let compute_corosync_max_host_failures ~__context =
     ((nhosts - disabled_hosts - 1) / 2) + disabled_hosts
   in
   corosync_ha_max_hosts
+
+let on_corosync_update ~__context ~cluster update =
```
lindig, line 407: The error messages going into the logs could use `"%s: .." __FUNCTION__` to make it more obvious where they are coming from.

ocaml/xapi-idl/cluster/cluster_interface.ml
Vincent-lau: This is using Lwt_unix.with_timeout, which says the following:
timeout d is a thread that remains suspended for d seconds and then fails with [Lwt_unix.Timeout](https://ocsigen.org/lwt/3.2.1/api/Lwt_unix#EXCEPTIONTimeout).

ocaml/idl/datamodel_cluster.ml
```diff
@@ -194,6 +194,16 @@ let t =
              (Some (VString Constants.default_smapiv3_cluster_stack))
            "Simply the string 'corosync'. No other cluster stacks are \
             currently supported"
+       ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:Bool "is_quorate"
+           ~default_value:(Some (VBool false)) "Whether we are quorate"
```
psafont, line 198:
```suggestion
           ~default_value:(Some (VBool false)) "Whether the cluster is quorate"
```

ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -21,6 +21,14 @@ let prototyped_of_field = function
       Some "23.14.0"
   | "Repository", "gpgkey_path" ->
       Some "22.12.0"
+  | "Cluster", "total_votes" ->
+      Some "23.32.0-next"
+  | "Cluster", "quorum" ->
+      Some "23.32.0"
```
psafont, line 27: Was this added previously? otherwise it should be 
```suggestion
      Some "23.32.0-next"
```

ocaml/idl/schematest.ml
```diff
@@ -3,7 +3,7 @@ let hash x = Digest.string x |> Digest.to_hex
 (* BEWARE: if this changes, check that schema has been bumped accordingly in
    ocaml/idl/datamodel_common.ml, usually schema_minor_vsn *)
 
-let last_known_schema_hash = "96da2c136aa13d4f8e71dd40fe2b84af"
```
psafont, line 6: Please change `schema_minor_vsn` in `ocaml/idl/datamodel_common.ml`, as the comment above instructs

ocaml/xapi-idl/cluster/cluster_interface.ml
Vincent-lau: hmm, I don't see the point of using monotonic time here. The timeout parameter represents how long we should wait for the command before we give up, and there is no comparison with other clock readings at all.

ocaml/idl/datamodel_lifecycle.ml
Vincent-lau: yes, you are right, should be next

ocaml/xapi-idl/cluster/cluster_interface.ml
psafont: The wait should use monotonic time to know how much it has waited :)
The other reason is that the constructors for `Mtime.Span.t` enforces unit of time, which removes ambiguity regarding these

Vincent-lau: But `Lwt_unix.with_timeout` expects a float as its timeout parameter. You mean have this `UPDATES.get` function take in a `Mtime.Span.t` and then convert it to a float when we pass it to `Lwt_unix.with_timeout`?

Vincent-lau: Ok, if I understand correctly, this is what you mean
```
      let timeout_p = Param.mk ~name:"timeout" Types.mtime
 ```
 where we need to add a new `mtime` into the rpclib? 
 
 Or maybe I misunderstand this:
 > Can we use an Int64 here so an Mtime.Span.t can be used and deserialized?
 Where does an Int64 come into play here?

Comment (psafont) [reply ocaml/xapi-idl/cluster/cluster_interface.ml:299]: Mtime.span provides function to convert losslessly from int64 (representing microseconds) https://ocaml.org/p/mtime/latest/doc/Mtime/Span/index.html#val-to_uint64_ns.

These could be serialized as Int64

File: ocaml/idl/datamodel_cluster.ml [hunk 6]
```diff
@@ -194,6 +194,16 @@ let t =
...
             currently supported"
+       ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:Bool "is_quorate"
+           ~default_value:(Some (VBool false)) "Whether the cluster is quorate"
+       ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:(Set String)
+           "quorum_members" ~default_value:(Some (VSet []))
+           "Current members that are in the quorum"
+       ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:Int "quorum"
+           ~default_value:(Some (VInt 0L))
+           "Number of nodes in order to be quorate"
+       ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:Int "total_votes"
```
robhoes, line 205: Is this simply the number of elements in `quorum_members`?

ocaml/idl/datamodel_cluster.ml
```diff
@@ -194,6 +194,16 @@ let t =
              (Some (VString Constants.default_smapiv3_cluster_stack))
            "Simply the string 'corosync'. No other cluster stacks are \
             currently supported"
+       ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:Bool "is_quorate"
+           ~default_value:(Some (VBool false)) "Whether the cluster is quorate"
+       ; field ~qualifier:DynamicRO ~lifecycle:[] ~ty:(Set String)
```
robhoes, line 199: This is a `string set`, which ends up getting populated with IP addresses if I see it right. In the context of the XenAPI, it may be better to have `host` references or `cluster_host` references. Still better may be to record this liveness state in a field in the relevant `cluster_host`.

ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -21,6 +21,14 @@ let prototyped_of_field = function
       Some "23.14.0"
   | "Repository", "gpgkey_path" ->
       Some "22.12.0"
+  | "Cluster", "total_votes" ->
```
robhoes, line 24: The file will have to be regenerated to update the versions.

ocaml/idl/datamodel_cluster.ml
Vincent-lau: I think the semantics of the vote is up to the corosync implementation. There are potential complicatiosn where quorum-device/pseudo node can have votes and nodes with more than one vote (very rare). But it does seem to me that in most cases this would be the same as the length of `quorum_members`

ocaml/idl/datamodel_cluster.ml
Vincent-lau: There seems to be a `joined` field in the datamodel already, maybe we can use this one?
```ocaml
    ; field ~qualifier:StaticRO ~lifecycle ~ty:Bool "joined"
           ~default_value:(Some (VBool true))
           "Whether the cluster host has joined the cluster"
```

robhoes: Yeah I saw that and am not clear on what it means exactly. Immediately below there is this comment:
```
         (* TODO: add `live` member to represent whether corosync believes that this
                  cluster host actually is enabled *)
```

Vincent-lau: Seems like we are doing the TODO here, but I feel we can reuse the `joined` field though.

robhoes: Perhaps yes, but we need to check how exactly it is used at the moment.

Vincent-lau: I have decided to introduce a new field called `live` as well as a corresponding `last_update_live` field to make sure that the user is not confused by stale info. I feel using `joined` field complicates the current situation with other methods that can modify join field such as `destroy` and `forget`. The `live` field itself cannot be modified by any API functions, which makes it a bit simpler.

Vincent-lau:
API rewritten accroding to @robhoes and @edwintorok suggestions. Introducing new fields to indicate whether the cluster stack thinks a host is alive. 

While there, also updated the API doc a little bit to clarify the difference between `enabled` and `joined`.

ocaml/idl/datamodel_cluster_host.ml
```diff
@@ -111,9 +111,21 @@ let t =
            ~default_value:(Some (VRef null_ref)) "Reference to the PIF object"
        ; field ~qualifier:StaticRO ~lifecycle ~ty:Bool "joined"
            ~default_value:(Some (VBool true))
-           "Whether the cluster host has joined the cluster"
-         (* TODO: add `live` member to represent whether corosync believes that this
-                  cluster host actually is enabled *)
+           {|Whether the cluster host has joined the cluster. Contrary to \
+            `enabled`, a host that is not joined is not considered a member of \
```
minglumlu, line 119: Could the coordinator be not joined the cluster? This is a configuration, rather than a state?

ocaml/xapi/xapi_clustering.ml
```diff
@@ -401,3 +424,99 @@ let compute_corosync_max_host_failures ~__context =
...
+            "call cluster watcher" 3.
+        in
+        match Idl.IdM.run @@ Cluster_client.IDL.T.get m with
+        | Ok updates -> (
+          match find_cluster_host ~__context ~host with
+          | Some ch ->
+              let cluster = Db.Cluster_host.get_cluster ~__context ~self:ch in
+              on_corosync_update ~__context ~cluster updates
+          | None ->
+              ()
```
minglumlu, line 510: Does this mean the current coordinator is not in cluster?

ocaml/xapi/xapi_cluster_host.ml
```diff
@@ -303,6 +304,7 @@ let enable ~__context ~self =
           "Cluster_host.enable: xapi-clusterd not running - attempting to start" ;
         Xapi_clustering.Daemon.enable ~__context
       ) ;
+      create_cluster_watcher_on_master ~__context ~host ;
```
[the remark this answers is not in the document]
Vincent-lau: probably when xapi exits? In which case the wather thread will exit as well after the main thread exits.

ocaml/idl/datamodel_cluster_host.ml
Vincent-lau: I think it is possible for the pool coordinator to not be in the cluster. In that case these fields will reflect the view from the pool coordinator, i.e. whehther the pool coordinator thinks another cluster host is alive, even if the coor is not in the majority quorum.

Of course, this information is probably inaccurate, but there isn't much we can do if the coordinator is not in the cluster. I guess we could disable the watcher when this is the case?

ocaml/xapi/xapi_clustering.ml
Vincent-lau: possibly, in which case doing nothing could be a sensible choice, I suppose...

ocaml/idl/datamodel_cluster_host.ml
minglumlu: If the pool coordinator is not in the cluster, would it be able to get the update from corosync?

ocaml/idl/datamodel_cluster_host.ml
```diff
@@ -105,15 +105,31 @@ let t =
            ~default_value:(Some (VRef null_ref)) "Reference to the Host object"
        ; field ~qualifier:StaticRO ~lifecycle ~ty:Bool "enabled"
            ~default_value:(Some (VBool false))
-           "Whether the cluster host believes that clustering should be \
-            enabled on this host"
+           {|Whether the cluster host believes that clustering should be \
```
psafont, line 108: This seems to break the SDK, somehow

ocaml/idl/datamodel_cluster_host.ml
Vincent-lau: I think so. Even if the coordinator is not in the majority quorum, it should still be able to contact corosync, but get back info on what it thinks the quorum state is.

Vincent-lau: Don't know if it is actually a good idea to do that, we can either say that these fields always reflect the view of the coordinator (which implies that when the coordinator is not in the majority quorum, its view will be inaccurate), or we can disable this update when the coordinator is not in the majority quorum.

ocaml/idl/datamodel_cluster_host.ml
Vincent-lau: maybe it does not like quoted strings?

ocaml/idl/datamodel_cluster_host.ml
Vincent-lau: Re the config vs state: I think we are using the state/config a bit interchangbly. I guess you probably mean that a config is a corosync config file, in which case I think we do change the config file each time when we update the joined field.

ocaml/idl/datamodel_cluster_host.ml
Vincent-lau: removing the quoted string makes it happy again

ocaml/idl/datamodel_cluster_host.ml
minglumlu: The `config` means a host is required (by users) to join into a cluster. While the `state` means it is required but is failing due to e.g. networking issue.
(Sorry I'm not familiar with all the concepts in cluster)
If the coordinator is not required to be in a cluster, I don't understand why its corosync can get the information from the cluster.

ocaml/xapi/xapi_cluster_host.ml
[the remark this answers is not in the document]
minglumlu: Yeah. I just see that it is called in `enable`. And meanwhile there is a `disable`.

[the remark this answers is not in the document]
Vincent-lau: good point, I feel it is a bit hard to pass the pid across different methods, so I have added an additional guard on when the watcher should be doing work.

ocaml/idl/datamodel_cluster_host.ml
```diff
@@ -106,14 +106,30 @@ let t =
...
+            corosync believes that this node is not part of the cluster. "
+       ; field ~qualifier:StaticRO ~lifecycle:[] ~ty:Bool "live"
+           ~default_value:(Some (VBool false))
+           "Whether the underlying cluster stack thinks we are live. This \
+            field is set automatically based on updates from the cluster stack \
+            and cannot be altered by the user."
+       ; field ~qualifier:StaticRO ~lifecycle:[] ~ty:DateTime "last_update_live"
+           ~default_value:(Some (VDateTime Date.epoch))
+           "Time when we were last able to retrieve the joined information \
+            from corosync"
```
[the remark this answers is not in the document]
Vincent-lau: and so is the new fields `live` and `last_update_live`?

Vincent-lau:
updated again based on the comments. Note I also changed the place where `create_cluster_watcher_on_master` is called. It should be called on the master when a cluster is created, and no need to invoke it in `resync_host` since that is only called for pool member hosts.

ocaml/idl/schematest.ml
```diff
@@ -3,7 +3,7 @@ let hash x = Digest.string x |> Digest.to_hex
 (* BEWARE: if this changes, check that schema has been bumped accordingly in
    ocaml/idl/datamodel_common.ml, usually schema_minor_vsn *)
 
-let last_known_schema_hash = "ab570d662d657ee2055194110d536302"
+let last_known_schema_hash = "a3ddab734ad88468120269cf3d0e3bd9"
```
robhoes, line 6: You still need to update the schema version in ocaml/idl/datamodel_common.ml.

ocaml/xapi/xapi_clustering.ml
```diff
@@ -401,3 +424,99 @@ let compute_corosync_max_host_failures ~__context =
...
+
+let create_cluster_watcher_on_master ~__context ~host =
+  if Helpers.is_pool_master ~__context ~host then (
+    let watch () =
+      while !Daemon.enabled do
+        let m =
+          Cluster_client.LocalClient.UPDATES.get (rpc ~__context)
+            "call cluster watcher" 3.
+        in
+        match Idl.IdM.run @@ Cluster_client.IDL.T.get m with
```
robhoes, line 503: I think we should add an exception handler around here (or an `| exception` case) with a delay to retry the call, just in case. Otherwise you may find the watcher thread exit unexpectedly and never get restarted.

ocaml/tests/test_cluster.ml
```diff
@@ -24,6 +24,8 @@ let test_clusterd_rpc ~__context call =
         {success= true; contents= Rpc.String test_token; is_notification= false}
   | ("enable" | "disable" | "destroy" | "leave" | "set-tls-verification"), _ ->
       Rpc.{success= true; contents= Rpc.Null; is_notification= false}
+  | ("UPDATES.get" | "UPDATES.create_stream"), _ ->
```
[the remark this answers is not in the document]
Vincent-lau: removed

ocaml/xapi/xapi_clustering.ml
Vincent-lau: The `Cluster_client.IDL.T.get` will not throw an exception I suppose? You mean we can avoid calling `handle_error` in the last case which will throw an exception and terminate the thread?

ocaml/idl/schematest.ml
Vincent-lau: Sorry I thought I had done this already. It might have been lost during one of my rebase. Updated now.

ocaml/xapi/xapi_clustering.ml
Vincent-lau: and perhaps L489 as well

ocaml/xapi/xapi_clustering.ml
```diff
@@ -164,6 +164,27 @@ let with_clustering_lock_if_cluster_exists ~__context where f =
   | _ ->
       with_clustering_lock where f
 
+let find_cluster_host_by_network ~__context ~addr =
+  match
+    Db.Cluster_host.get_all ~__context
```
robhoes, line 169: This is called for each host in `nodel` below, while you have `all_cluster_hosts` already available. So you can avoid all those DB calls by passing `all_cluster_hosts` to this function as an argument.

ocaml/xapi/xapi_clustering.ml
```diff
@@ -164,6 +164,27 @@ let with_clustering_lock_if_cluster_exists ~__context where f =
...
+    |> List.filter (fun cluster_host ->
+           let pIF = Db.Cluster_host.get_PIF ~__context ~self:cluster_host in
+           let (Cluster_interface.IPv4 ip) =
+             ip_of_pif (pIF, Db.PIF.get_record ~__context ~self:pIF)
```
robhoes, line 173: The DB calls here are also done for each host in `nodel` again and again. It is worth creating a list of addresses with cluster_host refs once, and then have the find function use that.

robhoes (approved):

ocaml/xapi/xapi_clustering.ml
Vincent-lau: should be fine as according to the ocaml-rpc doc: 

> `GenClientExn` raises an exception in case the response indicates a failure, instead of returning a result

which presumably means that the LocalClient here, generated by `GenClient` does not raise exceptions. But I have added an exception case just in case.

PR #5342, opened 2024-01-09
CP-46122: Support PAX/POSIX tar on import

edwintorok:
POSIX/PAX headers in TAR may be larger than 1 TAR block (otherwise we see the Pax global headers as a "file" in backwards compatible mode).

Calling Tar_unix.get_next_header would already read the appropriate amount from a FD to find the first true file header.
However an imported file can be either compressed or not, and we need to retry with decompression if parsing as a Tar failed. But we might be importing from a socket or pipe, so we cannot just seek back to the beginning, and we cannot use Tar_unix.get_next_header either because we don't have access to its internal buffer.

Implement a custom Tar header reader using the functor, that first feeds the block that we already read to the reader, and then the file itself directly (at which point we turn off the retry because we no longer buffer all that we read).

### DIFF
--- a/ocaml/xapi/import.ml
+++ b/ocaml/xapi/import.ml
@@ -2018,6 +2018,56 @@ let assert_filename_is hdr =
     raise (IFailure (Unexpected_file (expected, actual)))
   )
 
+(* We want to read tar headers from a file descriptor (possibly a socket/pipe),
+   and retry with decompression if that fails.
+   To be able to do that we store the first block that we read, and try to parse it with Tar first,
+   and if that fails try zstd/gzip.
+
+   We need a Tar HeaderReader that reads either from a buffer, and then continues reading from an FD
+*)
+module BufferOrFile = struct
+  type in_channel = {
+      mutable buffer: Cstruct.t option
+    ; fd: Unix.file_descr
+    ; retry_with_compression: bool ref
+  }
+
+  type 'a t = 'a
+
+  let make retry_with_compression buffer fd =
+    {buffer= Some buffer; retry_with_compression; fd}
+
+  let really_read t dest =
+    match t.buffer with
+    | Some first ->
+        t.buffer <- None ;
+        assert (Cstruct.length dest = Cstruct.length first) ;
+        Cstruct.blit first 0 dest 0 (Cstruct.length first)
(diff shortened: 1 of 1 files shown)


liulinC:
I did manual test on both xs8 and xs9, with command `/usr/bin/create-guest-templates`, works well on both.

edwintorok:
Looks like I'll need to reformat this.

lindig:
Should the first 100 bytes or so not be enough to know if a stream is compressed or not based on magic numbers?

edwintorok:
> Should the first 100 bytes or so not be enough to know if a stream is compressed or not based on magic numbers?

Yes, but by that point you've already read those 100 bytes and can't put them back into the file descriptor (and if it is a socket/pipe you can't seek back either). So we need the ability to parse a tar by parsing something from a buffer followed by something from a FD

lindig (approved):

edwintorok:
There is some code below that looks at the zstd magic, we could potentially refactor this by making it always look at the file magic and decide based on that (although it'll still need the tar reader code here), although that could potentially break import of existing archives if we don't get the magic code right so I'd rather do it in a separate PR.

ocaml/xapi/import.ml
```diff
@@ -2028,7 +2078,10 @@ let with_open_archive fd ?length f =
...
     Tar_unix.really_read fd buffer ;
     (* we assume the first block is not all zeroes *)
-    let hdr = Option.get (Tar.Header.unmarshal buffer) in
+    let hdr =
+      TarHeaderReader.read (BufferOrFile.make retry_with_compression buffer fd)
+      |> Result.get_ok
+    in
     assert_filename_is hdr ;
     (* successfully opened uncompressed stream *)
     retry_with_compression := false ;
```
Vincent-lau, line 2087: Is this the reason why a ref variable is used in the `in_channel` record rather than a mutable record?

Vincent-lau (approved):

ocaml/xapi/import.ml
edwintorok: it is set from 2 different places, a mutable record field could've been used but there was no record here previously, just local values.

PR #5343, opened 2024-01-09
Fixup: filter out EvacuateHost before safety check

minglumlu:
A safety check in host.apply_updates will raise error if there is a mandatory EvacuateHost guidance but some VMs are running on the host. Meanwhile, EvacuateHost should not be added into pending list as it should be be carried out before host.apply_updates. Hence it needs to be filtered out.

This commit fixes a bug that filering out the EvacuateHost before the safety check. It should be filtered out after the safety check.

### DIFF
--- a/ocaml/xapi/repository.ml
+++ b/ocaml/xapi/repository.ml
@@ -727,13 +727,11 @@ let apply_updates' ~__context ~host ~updates_info ~livepatches ~acc_rpm_updates
     =
   (* This function runs on coordinator host *)
   let open Guidance in
-  let guidance =
+  let guidance' =
     reduce_guidance ~updates_info ~updates:acc_rpm_updates ~livepatches
-    (* EvacuateHost should be carried out before host.apply_updates *)
-    |> List.map (fun (k, l) -> (k, List.filter (fun x -> x <> EvacuateHost) l))
   in
   let mandatory =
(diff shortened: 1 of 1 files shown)


robhoes (approved):

lindig (approved):

Vincent-lau (approved):

ocaml/xapi/repository.ml
```diff
@@ -727,20 +727,23 @@ let apply_updates' ~__context ~host ~updates_info ~livepatches ~acc_rpm_updates
...
-    (* EvacuateHost should be carried out before host.apply_updates *)
-    |> List.map (fun (k, l) -> (k, List.filter (fun x -> x <> EvacuateHost) l))
   in
   let mandatory =
-    match List.assoc_opt Mandatory guidance with
+    match List.assoc_opt Mandatory guidance' with
     | Some tasks ->
         tasks
     | None ->
         warn "No mandatory guidance found. Ignore it." ;
```
gangj, line 738: Does this requires a warning? Or debug?

gangj (approved):

ocaml/xapi/repository.ml
minglumlu: Mandatory guidance is expected to be there. But the absence of it should not block the update.

PR #5345, opened 2024-01-10
CP-46377: Add env vars `TRACEPARENT` & `OBSERVER_CONFIG_DIR` to `exec_xmlrpc` calls

GabrielBuica:
Sets up the necessary environment variables, `TRACEPARENT` and `OBSERVER_CONFIG_DIR`  for  SMAPI `observer.py` configuration.

### DIFF
--- a/ocaml/forkexecd/lib/forkhelpers.ml
+++ b/ocaml/forkexecd/lib/forkhelpers.ml
@@ -23,6 +23,8 @@
 
 let default_path = ["/sbin"; "/usr/sbin"; "/bin"; "/usr/bin"]
 
+let default_path_env_pair = [|"PATH=" ^ String.concat ":" default_path|]
+
 (* /var/ may not be writable when testing, use XDG_RUNTIME_DIR instead in that
    case. Avoid changing the directory unless the code is being tested. *)
 let test_path =
@@ -182,13 +184,7 @@ let safe_close_and_exec ?env stdin stdout stderr
         List.fold_left maybe_add_id_to_fd_map dest_named_fds predefined_fds
       in
 
-      let env =
-        match env with
-        | Some e ->
-            e
-        | None ->
-            [|"PATH=" ^ String.concat ":" default_path|]
-      in
+      let env = match env with Some e -> e | None -> default_path_env_pair in
       let syslog_stdout =
         match syslog_stdout with
         | NoSyslogging ->
--- a/ocaml/xapi-aux/env_record.mli
+++ b/ocaml/xapi-aux/env_record.mli
@@ -20,7 +20,7 @@
 (** A value of type [t] represents a value. 
   * It is constructed recursively from OCaml values using functions like [str], 
   * [option], or [list]. An environment value of type [t] can be observed using 
-  * [to_shell_string].
+  * [to_shell_string] or [to_string_array].
   *)
 type t
 
@@ -45,3 +45,7 @@ val to_shell_string : (string * t) list -> string
 (** [to_shell_string env_vars] is [env_vars] quoted as needed for use in a shell 
   * script.
   *)
+
+val to_string_array : t list -> string array
+(** Transforms a list of environment variables to a string array of environment 
+  * variables.*)
--- a/ocaml/xapi-aux/env_record.mli
+++ b/ocaml/xapi-aux/env_record.mli
@@ -1,0 +1,47 @@
+(*
+ * Copyright (C) 2023 Cloud Software Group
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+(** A Unix environment variable is a pair of a name and a value, both of which 
+  * are strings. This module helps to construct a complex environment value 
+  * that follows specific conventions.
+  *)
+
+(** A value of type [t] represents a value. 
+  * It is constructed recursively from OCaml values using functions like [str], 
+  * [option], or [list]. An environment value of type [t] can be observed using 
+  * [to_shell_string].
+  *)
+type t
+
+val str : string -> t
+(** Returns an environment variable value from a string.
+  *)
+
+val option : string option -> t
+(** Returns an environment variable value from a string option.
+  * If [None], returns an empty string value.
+  *)
+
+val list : ('a -> t) -> 'a list -> t
+(** [list element lst] is the list [lst] with elements converted by [element] 
+  * and separated by [,]. 
+  *)
+
+val pair : string * string -> t
+(** [pair (key, value)] is [key=value] pair. *)
+
+val to_shell_string : (string * t) list -> string
+(** [to_shell_string env_vars] is [env_vars] quoted as needed for use in a shell 
+  * script.
+  *)
--- a/ocaml/xapi/sm_exec.ml
+++ b/ocaml/xapi/sm_exec.ml
@@ -327,10 +327,29 @@ let is_smapi_enabled () =
   Xapi_observer_components.is_component_enabled
     ~component:Xapi_observer_components.SMApi
 
+let env_vars_of_observer dbg =
+  if is_smapi_enabled () then
+    let make_env k v = k ^ "=" ^ v in
+    let span_context_of_di (di : Debuginfo.t) =
+      Option.map (fun span -> Tracing.Span.get_context span) di.tracing
+    in
+    Debuginfo.of_string dbg
+    |> span_context_of_di
+    |> Option.map Tracing.SpanContext.trace_id_of_span_context
+    |> Option.map (fun v ->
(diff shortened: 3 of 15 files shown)


ocaml/xapi/sm_exec.ml
```diff
@@ -323,6 +323,30 @@ let with_session sr f =
...
+    Server_helpers.exec_with_new_task
+      (Printf.sprintf "check if component %s is enbled " component)
+      (fun __context ->
+        let observers = Db.Observer.get_all ~__context in
+        List.exists
+          (fun observer ->
+            let components =
+              Db.Observer.get_components ~__context ~self:observer
+            in
+            List.length components = 0
```
psafont, line 341:
```
            components = []
```

Vincent-lau: why do we use `components = []` here? When the component list is empty, we know that a component is enabled?

GabrielBuica: Solved with  this reorganizing https://github.com/xapi-project/xen-api/pull/5345/commits/443adb83c760b04f6b9b3d95285d000c48c7cd38 and this fixup https://github.com/xapi-project/xen-api/pull/5345/commits/5ba8c33688c59ee52da06b7221afa40ce825cddc.

ocaml/xapi/xapi_observer.ml
```diff
@@ -736,3 +656,5 @@ let set_components ~__context ~self ~value =
       to_add
   in
   do_set_op ~__context ~self ~observation_fn ~db_fn
+
+let dummy_enabler () = true
```
psafont, line 660: This name is not exposed in the commit

ocaml/xapi/sm_exec.ml
psafont: The original suggestion still stands :)

ocaml/xapi/xapi_observer.ml
GabrielBuica: Thanks, I missed removing this. Removed here https://github.com/xapi-project/xen-api/pull/5345/commits/cc2487298d1be6433d2eee553deaaffe9dac808b.

ocaml/xapi/xapi_observer_components.ml
```diff
@@ -0,0 +1,68 @@
...
+         (Component.to_string component)
+      )
+      (fun __context ->
+        let observers = Db.Observer.get_all ~__context in
+        List.exists
+          (fun observer ->
+            Db.Observer.get_components ~__context ~self:observer
+            |> List.map Component.of_string
+            |> observed_components_of
+            |> List.exists (( = ) component)
```
lindig, line 64: Is this not simply `List.mem`?

ocaml/xapi/xapi_observer_components.mli
```diff
@@ -0,0 +1,17 @@
+module Component : sig
+  type t = Xapi | Xenopsd | Xapi_clusterd | SMApi [@@deriving ord]
+
+  val all : t list
+
+  val to_string : t -> string
+
+  val of_string : string -> t
```
lindig, line 8: Might want to add that this may fail with an exception.

ocaml/xapi/xapi_observer_components.ml
GabrielBuica: I changed it to `List.mem`.

ocaml/xapi/xapi_observer_components.mli
GabrielBuica: Added.

ocaml/xapi/sm_exec.ml
GabrielBuica: I used @lindig's suggestion in the end.

ocaml/xapi/xapi_observer_components.ml
```diff
@@ -0,0 +1,106 @@
...
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+module Component = struct
```
Vincent-lau, line 15: Not a big issue but do we need this module any more as we are moving to a new ml file? I think it is ok to make all of these in the `Component` module toplevel. Or does it require lots of refactoring?

GabrielBuica: Seems like most of the changes required were in `xapi_observer.ml`. I created a fixup for this https://github.com/xapi-project/xen-api/pull/5345/commits/b9e3d635d1dfb9a2da7454245e974573fb870267.

mg12 (approved):
lgtm & passed dt bvt

ocaml/forkexecd/lib/forkhelpers.ml
```diff
@@ -182,13 +184,7 @@ let safe_close_and_exec ?env stdin stdout stderr
...
       in
 
-      let env =
-        match env with
-        | Some e ->
-            e
-        | None ->
-            [|"PATH=" ^ String.concat ":" default_path|]
-      in
+      let env = match env with Some e -> e | None -> default_path_env_pair in
```
lindig, line 187: Could be written as:
```
Option.fold ~some:Fun.id ~none:default_path_env_pair env
```

ocaml/xapi-aux/env_record.mli
```diff
@@ -0,0 +1,27 @@
...
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+type t
+
```
lindig, line 16: This needs documentation what is being constructed here.

ocaml/xapi/xapi_observer_components.mli
```diff
@@ -0,0 +1,38 @@
...
+ * by the Free Software Foundation; version 2.1 only. with the special
+ * exception on linking described in file LICENSE.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *)
+
+type t = Xapi | Xenopsd | Xapi_clusterd | SMApi [@@deriving ord]
```
lindig, line 15: What does this module provide? Some minimal high-level documentation to guide clients is required.

ocaml/xapi-aux/env_record.mli
lindig: The way I understand it, the documentation could be something like this:

A Unix environment variable is a pair of a name and a value, both of which are strings. This module helps to construct a complex environment value that follows conventions specific conventions.

A value of type [t] represents a value. It is constructed recursively from OCaml values using functions like [str], [option], or [list]. An environment value of type [t] can be observed using  [to_shell_string].

I am not sure the above is correct. I would have modeled this in a way that captures the name of an environment variable together with its value but I don't see that, which adds to my confusion.

snwoods: @edwintorok Created the first draft of env_record so he might be able to explain it a bit better but I think the reason that the environment variable isn't captured with its value is because it will always be a simple string. It doesn't need any processing on it like quoting or turning it into a comma-separated list like the environment values do.

snwoods (approved):

ocaml/xapi-aux/env_record.mli
GabrielBuica: https://github.com/xapi-project/xen-api/pull/5345/commits/6cae17d7935b39dc98dad5273fb85fa5922eeb20. I tried to summaries what I understood from @snwoods and @edwintorok's first draft.

ocaml/xapi/xapi_observer_components.ml
```diff
@@ -0,0 +1,99 @@
...
+    (function Xapi_clusterd -> !Xapi_clustering.Daemon.enabled | _ -> true)
+    all
+
+let assert_valid_components components =
+  try List.iter (fun c -> ignore @@ of_string c) components
+  with Unsupported_Component component ->
+    raise Api_errors.(Server_error (invalid_value, ["component"; component]))
+
+let observed_components_of components =
+  match components with [] -> startup_components () | components -> components
```
lindig, line 58: This looks dubious to me. How can an empty argument observe anything? I think the handling of the empty list is unexpected for clients.

GabrielBuica: As far as I understand it, it is because of how things are stored in the database. When you create an observer and you don't specify the components, the expectation is that all components are enabled. Querying the database for this observer returns empty.  I don't know the details of why was implemented like this, but I agree with it being annoying for clients.

lindig: The problem with the current design is that you can't have no components being observed because the empty list maps to all components. I think @mg12 should review this. It would be cleaner to use an option:

```
let observed_components_of = function
| None -> startup_components ()
| Some comps -> comps
```

Now one could calls `observed_components of (Some [])` and get back the empty list. Alternatively the behaviour of the empty list needs to be documented.

GabrielBuica:
Passed bvt after addressing comments.

PR #5346, opened 2024-01-10
fix(NUMA): 'default' is a keyword in some SDK languages

edwintorok:
'auto' is a keyword too.
Lets use 'default-policy' instead.

Fixes: d557b697a ("CP-38020: introduce Host.numa_affinity_policy")

### DIFF
--- a/ocaml/idl/datamodel_host.ml
+++ b/ocaml/idl/datamodel_host.ml
@@ -1673,12 +1673,12 @@ let host_numa_affinity_policy =
     ( "host_numa_affinity_policy"
     , [
         ("any", "VMs are spread across all available NUMA nodes")
-      ; ( "best-effort"
+      ; ( "best_effort"
         , "VMs are placed on the smallest number of NUMA nodes that they fit \
            using soft-pinning, but the policy doesn't guarantee a balanced \
            placement, falling back to the 'any' policy."
         )
-      ; ( "default"
+      ; ( "default_policy"
         , "Use the NUMA affinity policy that is the default for the current \
            version"
         )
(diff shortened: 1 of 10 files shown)


edwintorok:
I would've preferred 'default_policy' for the enum but lets be consistent with how 'best-effort ' and 'besteffort' looks like.

ocaml/idl/datamodel_host.ml
```diff
@@ -1678,7 +1678,7 @@ let host_numa_affinity_policy =
            using soft-pinning, but the policy doesn't guarantee a balanced \
            placement, falling back to the 'any' policy."
         )
-      ; ( "default"
+      ; ( "default-policy"
```
robhoes, line 1681: Better use _ instead of -. That's the convention in the API. I see I've missed `best-effort`...

robhoes: So let's also change `best-effort` to `best_effort`...

edwintorok: We have 'balance-slb' that has a '-'. I'm fine either way

robhoes: Then that one is bad also :) All enums in `host` and `VM` at least have _

robhoes: A bit too much of a replace-all now I think...

ocaml/xapi/xapi_xenops.ml
```diff
@@ -593,11 +593,11 @@ module MD = struct
             RealTime (qos_class params)
         | "idle" ->
             Idle
-        | "best-effort" ->
+        | "best_effort" ->
```
robhoes, line 596: Too many things being replaced now...

edwintorok: ah we already have 'best-effort' elsewhere

robhoes: I'd just update the datamodel and let the compiler tell me what else to update.

robhoes (approved):

psafont (approved):

psafont:
The commit [7de8c8f](https://github.com/xapi-project/xen-api/pull/5347/commits/7de8c8fdf82116225c4c2ec1babd10c4be56c02d) should be added as well

PR #5347, opened 2024-01-10
build: add a C# SDK sanity test

edwintorok:
I've recently broken the C# SDK build by introducing an enum that happened to be a keyword.

Would be good to test all SDK languages in the CI, but lets start with just one: add a command to the makefile to build-test the SDK, and then run the command in the CI.

This also works on Linux using the 'dotnet' build tool (which IIRC should be preinstalled on github workers, but we'll find out).

A hack was needed to get this to compile: generating FriendlyNames.cs from FriendlyNames.resx is not yet supported on Linux (needs some changes to the C# project definition/code/build system), but we don't want to run the SDK here, just test that it builds, so setting to null achieves that.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -91,6 +91,10 @@ jobs:
           opam exec -- make install DESTDIR=$(mktemp -d)
           opam exec -- make install DESTDIR=$(mktemp -d) BUILD_PY2=NO
 
+      - name: Sanity test SDK
+        run: |
+          opam exec -- make sdksanity
+
       - name: Uninstall unversioned packages and remove pins
         # This should purge them from the cache, unversioned package have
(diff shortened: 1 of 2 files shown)


edwintorok:
This has failed as expected: "/home/runner/work/xen-api/xen-api/_build/install/default/xapi/sdk/csharp/src/host_numa_affinity_policy.cs(45,21): error CS1041: Identifier expected; 'default' is a keyword [/home/runner/work/xen-api/xen-api/_build/install/default/xapi/sdk/csharp/src/XenServer.csproj::TargetFramework=netstandard2.0]"

edwintorok:
This has now been included as part of https://github.com/xapi-project/xen-api/pull/5346, which has a successful CI run. Closing.

PR #5348, opened 2024-01-11
CA-387201: Pool.last_sync_date not reset if the user changes update channel

minglumlu:
When a user adds more update repositories as enabled ones, the pool.last_update_sync will be invalid as at the moment not all the enabled repositories have been synced at the time which is recorded in pool.last_update_sync.

To avoid confusion, this commit resets the field to epoch so that the user can be notified to do a pool.sync_updates soon.

Disabling an update repository will not make the pool.last_update_sync be invalid unless no enabled repositories after the disabling.

### DIFF
--- a/ocaml/xapi/xapi_pool.ml
+++ b/ocaml/xapi/xapi_pool.ml
@@ -3352,11 +3352,14 @@ let set_repositories ~__context ~self ~value =
     (fun x ->
       if not (List.mem x existings) then (
         Db.Repository.set_hash ~__context ~self:x ~value:"" ;
-        Repository.reset_updates_in_cache ()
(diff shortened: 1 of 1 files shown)


lindig (approved):
This looks correct to me. Maybe any change to the list of repos should invalidate this date such that we don't need special cases? I understand the argument that removing a repo does not invalidate the date unless it was the last one - which is what the code does.

robhoes (approved):

minglumlu:
>  Maybe any change to the list of repos should invalidate this date such that we don't need special cases?
Yes. The special case is annoying. But I think here the date needs to be reset as soon as possible so that the "never synced" state could be populated to users in time.

PR #5349, opened 2024-01-11
Add record_util backwards compatibility test

edwintorok:
This is a followup on this review comment on the NUMA PR:
https://github.com/xapi-project/xen-api/pull/5226#discussion_r1441994990

It doesn't yet auto-generate record_util, but adds a test that checks that any changes to record_util are fully backwards compatible by exhaustively testing all enum value conversions to and from string.
Exception checks are less strict: if they raise an exception then it must be the same type, but we allow its string argument (and obviously its module name) to be different: this will allow us to improve the error messages in the future without breaking the test.

Last time we tried to auto-generate this (I think by @psafont ) it was difficult to test that it was backwards compatible so lets do test first this time.


psafont (approved):

lindig (approved):

edwintorok:
Rebased on latest master, will merge to feature branch once CI passes.

PR #5350, opened 2024-01-11
CP-47043: Port usb_reset.py to python3. 

ashwin9390:
PR#5350 contains the following changes:

- Shebang changed to python3,
- map() needs to be converted to list()
- Change `setup_cgroup()` to binary I/O:

Python3 does not support unbuffered Raw I/O in Text mode:

Fix "can't have unbuffered text I/O" by changing the code to use binary mode instead of text mode:
when executing `/opt/xensource/libexec/usb_reset.py attach 1-13.2 -d 16 -p 8470`. 
```py
Error code:
  File "/opt/xensource/libexec/usb_reset.py", line 343, in <module>
    attach(arg.device, arg.domid, arg.pid, arg.reset_only)
  File "/opt/xensource/libexec/usb_reset.py", line 285, in attach
    setup_cgroup(domid, pid)
  File "/opt/xensource/libexec/usb_reset.py", line 215, in setup_cgroup
    with open(cg_dir + "/tasks", "w", 0) as tasks, \
ValueError: can't have unbuffered text I/O  
```
Testing information:
Koji rpm package created with converted code: [taskID=567641], XenRT Job 3902923 completed succesfully

Signed-off-by: Ashwinh <ashwin.h@cloud.com>
Co-authored-by: Bernhard Kaindl <bernhard.kaindl@cloud.com> 

@psafont @Vincent-lau @lindig

### DIFF
--- a/scripts/usb_reset.py
+++ b/scripts/usb_reset.py
@@ -236,14 +236,10 @@ def setup_cgroup(domid, pid):  # type:(str, str) -> None
         # Using the unbuffered Raw IO mode, we know the write was done
         # in exactly this way by the write function call itself, not later.
         #
-        # With such short writes, splitting the writes because of a
-        # too small buffer will not usually happen, but be paranoid
-        # and keep the same code for Python3 too. We've to use binary
-        # mode for Python3 for this as Python3 does now support unbuffered
-        # Raw I/O in Text mode (with implicit unicode str to byte encode()).
-        #
-        # In Python2 str is bytes anyway so in Python2 it was already binary.
-        # Using bytes for "wb" means also no change for use with Python3:
+        # With small writes like this , splitting them because of overflowing the
+        # buffer is not expected to happen. To stay safe and keep using unbuffered I/O
+        # We have to migrate to binary mode in python3,as python3 supports unbuffered 
+        # raw I/O in binary mode.
         #
         with open(cg_dir + "/tasks", "wb", 0) as tasks, \
                 open(cg_dir + "/devices.deny", "wb", 0) as deny, \
--- a/scripts/usb_reset.py
+++ b/scripts/usb_reset.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 #
 # Copyright (C) Citrix Systems Inc.
 #
@@ -122,7 +122,7 @@ def load_device_ids(device):
     ids_path = get_ids_path(device)
     try:
         with open(ids_path) as f:
-            uid, gid = map(int, f.readline().split())
+            uid, gid = list(map(int, f.readline().split()))
     except (IOError, ValueError) as e:
         log.error("Failed to load device ids: {}".format(str(e)))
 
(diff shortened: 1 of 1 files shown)


scripts/usb_reset.py
```diff
@@ -212,17 +212,17 @@ def setup_cgroup(domid, pid):
 
     try:
         # unbuffered write to ensure each one is flushed immediately
-        with open(cg_dir + "/tasks", "w", 0) as tasks, \
-                open(cg_dir + "/devices.deny", "w", 0) as deny, \
-                open(cg_dir + "/devices.allow", "w", 0) as allow:
+        with open(cg_dir + "/tasks", "wb") as tasks, \
+                open(cg_dir + "/devices.deny", "wb") as deny, \
+                open(cg_dir + "/devices.allow", "wb") as allow:
```
bernhardkaindl, line 217: Due to cgroupfs and the comment above it, keep using unbuffered I/O and explain why switching to binary I/O (the written bytes are the same):
```suggestion
        # Writing to cgroupfs files is done unbuffered to ensure the order of writes.
        # Unbuffered I/O needs binary I/O in Python3:
        with open(cg_dir + "/tasks", "wb", 0) as tasks, \
                open(cg_dir + "/devices.deny", "wb", 0) as deny, \
                open(cg_dir + "/devices.allow", "wb", 0) as allow:
```

scripts/usb_reset.py
```diff
@@ -212,17 +212,17 @@ def setup_cgroup(domid, pid):
...
 
             # deny all
-            deny.write("a")
+            deny.write(b"a")
 
             # grant rw access to /dev/null by default
-            allow.write(get_ctl("/dev/null", "rw"))
+            allow.write(get_ctl("/dev/null",b"rw"))
```
bernhardkaindl, line 223: And keep the space for readability / standard formatting:
```suggestion
            allow.write(get_ctl("/dev/null", b"rw"))
```

bernhardkaindl (changes requested):

psafont (approved):
This looks reasonable, it's reassuring that the pertinent regression tests pass internally (job run 3902923)

scripts/usb_reset.py
```diff
@@ -212,17 +228,36 @@ def setup_cgroup(domid, pid):
...
+        # may have to be a single write() system call for the entire string.
+        #
+        # Using the unbuffered Raw IO mode, we know the write was done
+        # in exactly this way by the write function call itself, not later.
+        #
+        # With such short writes, splitting the writes because of a
+        # too small buffer will not usually happen, but be paranoid
+        # and keep the same code for Python3 too. We've to use binary
+        # mode for Python3 for this as Python3 does now support unbuffered
+        # Raw I/O in Text mode (with implicit unicode str to byte encode()).
```
bernhardkaindl, line 243: I wrote this comment in a bit of a hurry.
When reading it now, I found that its wording was not great.
I propose to rephrase it like in this proposed commit/change:
```suggestion
        # With small writes like thise, splitting them because of overflowing the
        # buffer is not expected to happen. But stay on safe and keep using unbuffered IO.
        # We've to migrate to binary mode for Python3 for this because
        # Python3 only supports unbuffered raw I/O in binary mode:
```

scripts/usb_reset.py
```diff
@@ -212,17 +228,36 @@ def setup_cgroup(domid, pid):
...
+        #
+        # In Python2 str is bytes anyway so in Python2 it was already binary.
+        # Using bytes for "wb" means also no change for use with Python3:
```
bernhardkaindl, line 246: Ditto, I wrote this comment in a hurry, the comment could be better.
As it was only meant to describe the change for reviewers, it can be dropped for the merge.
Or I could rephrase it, as this last paragraph was only meant for the review.
```suggestion
```

bernhardkaindl (approved):
While I've submitted two proposed commits for updating the docs, they are just improvements for the wording of a comment that I wrote.

scripts/usb_reset.py
ashwin9390: Done

PR #5352, opened 2024-01-11
CP-46374: Add a minimal observer.py

snwoods:
This can be used to test the forthcoming switch which will run sm scripts through observer.py when there is an observer.


snwoods:
Requires: https://code.citrite.net/projects/XSS/repos/xapi.spec/pull-requests/204/overview

scripts/observer.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/python3
+
+if __name__ == '__main__':
+  # run a program passed as parameter, with its original arguments
+  import runpy
+  import sys
+
+  # shift all the argvs one left
+  sys.argv = sys.argv[1:]
+  argv0=sys.argv[0]
```
duobei, line 10: Does shift all the argvs one left only to get the head after shifted?
Could we get the value directly:
```python
file=sys.argv[1]

def run(file):
    runpy.run_path(file, run_name='__main__')

run(file)
```

scripts/Makefile
```diff
@@ -177,6 +177,7 @@ endif
 	sed -i 's/#!\/usr\/bin\/python/#!\/usr\/bin\/python3/' $(DESTDIR)$(SITE3_DIR)/XenAPIPlugin.py
 	$(IDATA) examples/python/XenAPI/XenAPI.py $(DESTDIR)$(SITE3_DIR)/
 	$(IDATA) examples/python/inventory.py $(DESTDIR)$(SITE3_DIR)/
+	$(IDATA) observer.py $(DESTDIR)$(SITE3_DIR)/
```
lindig, line 180: Does this require a change in xapi.spec?

scripts/observer.py
snwoods: It's not just to get the head, runpy.run_path passes sys.argv on to the new file. So we shift sys.argv so that if you did e.g. python3 observer.py test.py 5, in test.py sys.argv[0] = test.py and sys.argv[1] = 5, the same as if test.py had been run directly (python3 test.py 5)

Vincent-lau: I think this doc explains it
```
A number of alterations are also made to the [sys](https://docs.python.org/3/library/sys.html#module-sys) module. Firstly, [sys.path](https://docs.python.org/3/library/sys.html#sys.path) may be altered as described above. sys.argv[0] is updated with the value of path_name
```

robhoes (approved):

Vincent-lau (approved):

PR #5355, opened 2024-01-12
build: set a timeout for the tests

edwintorok:
We have some non-deterministic timeouts on github actions. Introduce some shorter timeouts that is under our control,
 and print more information about what is stuck.

GH actions take <2m to run the tests, so add 5m as a timeout for the children, and twice that for the parent.


edwintorok:
This may need some tweaks to cancel the 'sleep && ps' which would otherwise spam the local console a while later, but lets first see whether we can repro the timeout in GH actions by rerunning the test until it gets stuck.

edwintorok:
Hmm forkexecd is failing consistently like this on github (and sometimes locally too):
```
 63  11 ########Fatal error: exception Unix.Unix_error(Unix.ECONNRESET, "read", "")
make: *** [Makefile:59: test] Error 1
```

We should fix that, but that is not the 30m+ hang bug I'm looking to track down, so lets try without these tests.

edwintorok:
Closing in favour of the real PR where tests are green.

PR #5357, opened 2024-01-12
build: set a timeout for the tests and fix forkexecd unit test

edwintorok:
We have some non-deterministic timeouts on github actions. Introduce some shorter timeouts that is under our control,
 and print more information about what is stuck.

GH actions take <2m to run the tests, so add 5m as a timeout for the children, and twice that for the parent.

This is like https://github.com/xapi-project/xen-api/pull/5355 but without the hacks to disable unreliable tests.

If the CI passes on this one then we should probably merge this to make diagnosing test failures in other PRs easier. I added the missing cleanup code so the ps output should only be printed if there is an actual timeout (if the tests fail or succeed sooner then the sleeping process is killed and ps tree is never dumped).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,8 +28,41 @@ lint:
 	pylint --disable=line-too-long,too-few-public-methods,unused-argument,no-self-use,invalid-name,broad-except,protected-access,redefined-builtin,too-many-lines,wildcard-import,too-many-branches,too-many-arguments,unused-wildcard-import,raising-format-tuple,too-many-statements,duplicate-code _build/default/xapi-storage/python/xapi/storage/api/v5/*.py
 	pycodestyle --ignore=E501 _build/default/xapi-storage/python/xapi/storage/api/v5/*.py
 
+
+# ulimit -S -t <N>
+#  Set a soft CPU time quota which will kill processes with SIGXCPU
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -28,8 +28,41 @@ lint:
...
+
+# ulimit -n <NFILES>
+#  By default the ulimit on some systems is very large (e.g. Fedora39 distrobox 1048576)
+#  which causes some tests to take very long to run (e.g. forkexec tests which loop through and close all fds up to limit)
+
+TEST_TIMEOUT=600
+TEST_TIMEOUT2=1200
 test:
-	dune runtest --profile=$(PROFILE) --error-reporting=twice -j $(JOBS)
+	ulimit -S -t $(TEST_TIMEOUT); \
```
lindig, line 59: This is a tricky bit of code but it looks plausible. Indentation is a bit arbitrary. For anyone wondering: each line in Makefile recipe is executed in its own shell. Here it is important to keep this in a single shell, so we need continuation escapes.

edwintorok:
I'm fixing the forkexecd bug in this PR, because it always reproduces here (this PR explicitly sets an `ulimit -n 1024` instead of leaving it up to the environment to chose a default).

lindig (approved):

psafont (approved):

PR #5358, opened 2024-01-15
Fix Protocol_async.with_lock bug

edwintorok:
@Vincent-lau noticed that our Mutex implementation in message-switch async is completely broken: it doesn't guarantee mutual exclusion at all: the Async scheduler can switch away when the bind is executed (and unlike the Lwt scheduler I think it even guarantees it *will* switch away) and execute some other code which will think it can grab the mutex because it is free.

Considering the hangs in the async unit tests, this looks like a plausible reason. Replace it with code from Async itself (there is no Mutex implementation there anymore, there used to be one in Async_extended, but there is a Sequencer which does pretty much the same thing).

Another way to fix this would be to move the 't.m <- true' code inside the way, just before 'return ()', because at that point we'd be guaranteed not to have switched away to another Async thread (of course we could've switched away to another pure OCaml thread, there are no guarantees against that).

Draft, because we need to test this change a lot more, also to see whether it fixes the underlying hang (and if so then decide which fix to backport to Yangtze).


psafont:
This PR deletes all the related code: https://github.com/xapi-project/xen-api/pull/5304 ;)

edwintorok:
Looks like we'll need to fix that forkexecd unit test problem next.

edwintorok:
Merged the alternative fix instead.

PR #5359, opened 2024-01-15
CA-387699: Fix Protocol_async.with_lock bug spotted by Vincent

edwintorok:
Monadic concurrency libraries can switch away to another 'promise' whenever the bind operator is called.
In fact Async will always switch away, but Lwt would only switch away if the promise is blocked (this is probably the origin of the bug).

Move the 't.m <- true' next to where we checked that it is false to ensure that we are the only ones holding it.
(This is still vulnerable to race conditions with pure OCaml threads, but not with Async promises).

Another alternative would be to use Async.Throttle.Sequencer, but this change is a minimal one that could be backported to Yangtze even.

### DIFF
--- a/ocaml/message-switch/async/protocol_async.ml
+++ b/ocaml/message-switch/async/protocol_async.ml
@@ -81,14 +81,14 @@ module M = struct
       {m; c}
 
     let with_lock t f =
-      let rec wait state =
-        if Bool.(t.m = state) then
+      let rec wait () =
+        if Bool.(t.m = false) then (
+          t.m <- true ;
           return ()
-        else
-          Condition.wait t.c >>= fun () -> wait state
+        ) else
+          Condition.wait t.c >>= wait
       in
-      wait false >>= fun () ->
-      t.m <- true ;
+      wait () >>= fun () ->
       Monitor.protect f ~finally:(fun () ->
           t.m <- false ;
           Condition.broadcast t.c () ;


edwintorok:
This is a minimal alternative to https://github.com/xapi-project/xen-api/pull/5358 that should be backportable to Yangtze too.

ocaml/message-switch/async/protocol_async.ml
```diff
@@ -82,13 +82,13 @@ module M = struct
 
     let with_lock t f =
       let rec wait state =
-        if Bool.(t.m = state) then
+        if Bool.(t.m = state) then (
+          t.m <- true ;
```
robhoes, line 86: As `state` is a variable in this function, while called with `false` below, I think this line should be 

    t.m <- not state ;

edwintorok: This is only ever called with `state = false`, I changed it to take a `()`, and that actually simplified this line too:
``` diff
-          Condition.wait t.c >>= fun () -> wait state
+          Condition.wait t.c >>= wait
```

lindig (approved):

edwintorok:
Our internal test infrastructure is broken at the moment, but the github CI is happier, so we should probably merge this to unblock other PRs?

psafont (approved):

edwintorok:
Internal test passed, so I'm merging this.

PR #5360, opened 2024-01-16
CI: scripts/usb_reset_test.py: Test usb_reset.py function setup_cgroup()

bernhardkaindl:
@ashwin9390's PR 5350 needs a unit test before changing the code of `scripts/usb_reset.py` to make sure that it does not negatively affect the product.

IT is using unbuffered I/O for writing kernel cgroupfs files, and to support it with Python3, its function setup_cgroup() needs to be converted to use RawIO (unbuffered binary IO) 

Maybe unbuffered I/O is not really needed, but to be on the safe side, continue using it in this special case (cgroupfs is a special sysfs-based kernel file system, and writes should write complete strings).

For reference of Python Raw I/O, see: https://docs.python.org/3.11/library/io.html#raw-i-o

The first commit only adds the testcase for the existing Python2 code to ensure that the test case of correct.

(work-in-progress, but if all tests pass, it can be merged already)


liulinC:
why not name the unittest as `test_xxx.py`, is't  encouraged by unittest?

.github/workflows/main.yml
```diff
@@ -7,30 +7,46 @@ on:
     # run daily, this refreshes the cache
     - cron: '13 2 * * *'
 
+# Cancel a currently running workflow from the same PR, branch or tag when
+# a new workflow is triggered:
+# https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre
+concurrency:
```
psafont, line 13: Please add this change into its own separate commit, it's unrelated to the python changes

robhoes:
Where did that Codecov comment come from? Did anyone turn this on?

bernhardkaindl:
The comment seems to have been queued and added when you approved allowing Codecov to make comments.

#5350 merged the fix, but not the unit test for the fix. I'll submit the unit test as a separate PR.

PR #5362, opened 2024-01-16
Add test for lock implementation in message_switch

Vincent-lau:
The lock implementation for `protocol_async` was not correct, hence adding the test suggested by Edwin to trigger race.

Test works by trying to acquire a standard Ocaml mutex in a critical section provided by the `with_lock` implementation. This means that any failed attempt to acquire the lock implies the `with_lock` implementation does not provide mutual exclusion.

### DIFF
--- a/ocaml/message-switch/core/mtest.ml
+++ b/ocaml/message-switch/core/mtest.ml
@@ -1,0 +1,42 @@
+module type MTEST = sig
+  type +'a io
+
+  val mutex_provides_mutal_exclusion : unit -> unit io
+end
+
+module Make =
+functor
+  (M : S.BACKEND)
+  ->
+  struct
+    open M.IO
+
+    type 'a io = 'a M.IO.t
+
+    let ocaml_lock = Mutex.create ()
+
+    let mu = M.Mutex.create ()
+
+    let cond = M.Condition.create ()
+
+    let broadcast () = M.Condition.broadcast cond ()
+
+    let mutex_provides_mutal_exclusion () : unit io =
+      let promises =
+        List.init 100 (fun _ ->
+            M.Condition.wait cond >>= fun () ->
+            M.Mutex.with_lock mu (fun () ->
+                M.IO.return_unit >>= fun () ->
+                (* the with_lock implementation should ensure that only one
+                   monad can try to acquire this lock *)
+                assert (Mutex.try_lock ocaml_lock) ;
+                M.IO.return_unit >>= fun () ->
+                Mutex.unlock ocaml_lock ; M.IO.return_unit
+            )
+        )
+      in
+      broadcast () ;
+      ignore @@ all promises ;
+      Printf.printf "%s test.\n" (M.whoami ()) ;
+      M.IO.return_unit
+  end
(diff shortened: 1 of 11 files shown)


Vincent-lau:
Related fix for the race condition in async #5359

Vincent-lau:
Some quality gate issues regarding ml files. Do we care about ml files in `core_test`? We can either add path to exclude or add mli files for these tests.

ocaml/message-switch/core/mtest.ml
```diff
@@ -0,0 +1,40 @@
...
+    let mu = M.Mutex.create ()
+
+    let cond = M.Condition.create ()
+
+    let broadcast () = M.Condition.broadcast cond ()
+
+    let mutex_provides_mutal_exclusion () : unit io =
+      ignore @@ M.Clock.run_after 1 broadcast ;
+      ignore
+      @@ Array.init 100 (fun _ ->
```
edwintorok, line 27: For Lwt I think we'll need to wait until all of these are complete. Instead of ignoring the array try storing it in a value (or as a list).

ocaml/message-switch/core/mtest.ml
```diff
@@ -0,0 +1,40 @@
...
+    let ocaml_lock = Mutex.create ()
+
+    let mu = M.Mutex.create ()
+
+    let cond = M.Condition.create ()
+
+    let broadcast () = M.Condition.broadcast cond ()
+
+    let mutex_provides_mutal_exclusion () : unit io =
+      ignore @@ M.Clock.run_after 1 broadcast ;
```
edwintorok, line 25: Move this after the array creation: then you don't need to sleep for 1s, you can immediately broadcast after the array has been created.

ocaml/message-switch/core/mtest.ml
```diff
@@ -0,0 +1,40 @@
...
+                 M.IO.return_unit >>= fun () ->
+                 (* the with_lock implementation should ensure that only one monad can
+                    try to acquire this lock *)
+                 assert (Mutex.try_lock ocaml_lock) ;
+                 M.IO.return_unit >>= fun () ->
+                 Mutex.unlock ocaml_lock ; M.IO.return_unit
+             )
+         ) ;
+      Printf.printf "%s test.\n" (M.whoami ()) ;
+      M.IO.return_unit
```
edwintorok, line 39: would be good to wait until all promises in the array are resolved to ensure we've let the code run fully. Then we can remove the arbitrary 2s sleep, and make this test complete more quickly.

edwintorok (commented):
Thanks, this test is a good start. See my comments for some suggestions on how to eliminate the sleeps and make the test more robust for Lwt.

lindig:
Are we not trying to remove all Async code and is this adding more?

psafont:
It's adding a test to ensure the async code works as expected. This will be removed in the coming months, but I believe it's good we have proof of the fix in the repo

psafont:
Please rebase, tests should be more stable now

psafont (approved):

lindig:
@edwintorok you had some initial comments; so would suggest to revisit :arrow_double_up:

lindig (approved):
Looks good to me but I think Edvin should take a look because he had comments.

Vincent-lau:
We can probably merge this one if @edwintorok is busy, it's a small change in unit tests anyway.

psafont:
I agree, this is low risk

PR #5363, opened 2024-01-16
CA-387456 serialise Pool.eject

lindig:
When a host is ejected from a pool, its certificate for pool-internal communication is removed from all remaining hosts. This is unlikely to work when we have multiple eject operations overlapping. Hence, introduce an eject operation and make sure ejects are serialised.

### DIFF
--- a/ocaml/idl/schematest.ml
+++ b/ocaml/idl/schematest.ml
@@ -3,7 +3,7 @@ let hash x = Digest.string x |> Digest.to_hex
 (* BEWARE: if this changes, check that schema has been bumped accordingly in
    ocaml/idl/datamodel_common.ml, usually schema_minor_vsn *)
 
-let last_known_schema_hash = "29a820cd4c81617f07b43c9b4f934317"
+let last_known_schema_hash = "ab570d662d657ee2055194110d536302"
 
 let current_schema_hash : string =
   let open Datamodel_types in
--- a/ocaml/idl/datamodel_common.ml
+++ b/ocaml/idl/datamodel_common.ml
@@ -10,7 +10,7 @@ open Datamodel_roles
               to leave a gap for potential hotfixes needing to increment the schema version.*)
 let schema_major_vsn = 5
 
-let schema_minor_vsn = 771
+let schema_minor_vsn = 772
 
 (* Historical schema versions just in case this is useful later *)
 let rio_schema_major_vsn = 5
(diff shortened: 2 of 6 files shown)


ocaml/idl/datamodel_lifecycle.ml
```diff
@@ -30,7 +30,7 @@ let prototyped_of_field = function
   | "VTPM", "persistence_backend" ->
       Some "22.26.0"
   | "host", "numa_affinity_policy" ->
-      Some "23.32.0-next"
+      Some "24.2.0"
```
robhoes, line 33: Ah, could you change these here and below to 24.0.0 please?

lindig: How about removing these changes entirely (I usually don't include them)?

robhoes (approved):
This looks right. Have you verified that parallel ejects behave correctly without raising errors? You should see OTHER_OPERATION_IN_PROGRESS in the coordinator logs while an eject call is being "queued".

ocaml/idl/datamodel_lifecycle.ml
robhoes: It needs to be done eventually...

lindig:
Yes, I see them:
```
api: [debug||5737 /var/lib/xcp/xapi|pool.eject R:54b37a4a4896|helpers] Pool.eject locking failed: caught transient failure OTHER_OPERATION_IN_PROGRESS: [ pool; OpaqueRef:52251de1-4366-6c23-2ca3-c16ae442f0c2 ]
api: [debug||5737 /var/lib/xcp/xapi|pool.eject R:54b37a4a4896|helpers] Waiting for up to 5.998897 seconds before retrying...
```

ocaml/idl/schematest.ml
```diff
@@ -3,7 +3,7 @@ let hash x = Digest.string x |> Digest.to_hex
 (* BEWARE: if this changes, check that schema has been bumped accordingly in
    ocaml/idl/datamodel_common.ml, usually schema_minor_vsn *)
```
psafont, line 4: please change schema_minor_vsn to 772

psafont (approved):

lindig:
@minglumlu I have not looked; this one was found in testing. I assume joining is less critical with respect to certificates because we are adding capabilities rather taken them away.

psafont:
I did take into account parallel join when designing the feature, and we took care in ensuring it works in any case.

In particular, xapi checks for missing certificates on startup and downloads them from the coordinator

PR #5364, opened 2024-01-17
CP-46264 deprecate host.bios_strings[hp-rombios] entry

lindig:
The host object maintains a map bios_strings that has an entry for the "hp-rombios key. The value is based on a value read from dom0 memory using /dev/mem. This is incompatible with secure boot and this patch makes the value returned the empty string. The value is patched into a guest's memory so we can't return anything else.

We might want to remove the key `hp-rembios` in the future.

### DIFF
--- a/ocaml/xapi/bios_strings.ml
+++ b/ocaml/xapi/bios_strings.ml
@@ -186,21 +186,13 @@ let get_oem_strings decode =
   in
   standard @ List.mapi convert values
 
-(* Get the HP-specific ROMBIOS OEM string:
- * 6 bytes from the memory starting at 0xfffea *)
-let get_hp_rombios () =
-  let hp_rombios = Bytes.make 6 ' ' in
-  ( try
-      let mem = Unix.openfile "/dev/mem" [Unix.O_RDONLY] 0 in
-      Xapi_stdext_pervasives.Pervasiveext.finally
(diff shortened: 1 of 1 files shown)


ocaml/xapi/bios_strings.ml
```diff
@@ -186,21 +186,13 @@ let get_oem_strings decode =
...
-    with _ -> ()
-  ) ;
-  match Bytes.unsafe_to_string hp_rombios with "COMPAQ" -> "COMPAQ" | _ -> ""
+(* CP-46264 - this used to read a string from /dev/mem but we can no
+   longer support this. It would either find "COMPAQ" and return it, or
+   return the empty string. We are now always returning the
+   "". The string is patched into a guest's memory - so it would not be
+   safe to return an arbitrary string. *)
+
+let get_hp_rombios () = ""
```
psafont, line 195: A couple more lines from this ugly matter can be deleted by directly defining the list here:
```suggestion
let hp_rombios = [("hp-rombios", "")] in
```

psafont:
Line 204 and 205 in that file should be deleted :)

psafont (approved):
The CI error is unrelated, I'll push for a new version of stdext to fix it

robhoes (approved):

PR #5366, opened 2024-01-17
CP-46939: Add config option observer_endpoint_http_enabled

snwoods:
This option defaults to false as it is unsafe for customers to use http endpoints. This can be enabled for dev testing and XenRT.


robhoes:
This also blocks HTTPS, which is not intended?

lindig (approved):

mg12 (approved):

lindig:
I've added a label

snwoods:
The XenRT fix has now been merged

PR #5369, opened 2024-01-18
CA-387560 add support for more systemd execution types

lindig:
We use systemd to start transient services. So far, we use the Simple type but would like to add oneshot and potentially more.

### DIFF
--- a/ocaml/forkexecd/lib/fe_systemctl.ml
+++ b/ocaml/forkexecd/lib/fe_systemctl.ml
@@ -28,7 +28,7 @@ let action ~service action =
 
 (** Systemd execution type *)
 module Type = struct
-  type t = Simple | OneShot | Forking
+  type t = Simple | OneShot | Forking | Notify | Idle
 
   let to_string = function
     | Simple ->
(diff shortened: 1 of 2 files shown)


ocaml/forkexecd/lib/fe_systemctl.ml
```diff
@@ -26,12 +26,25 @@ let action ~service action =
...
+module Type = struct
+  type t = Simple | OneShot | Forking
+
+  let to_string = function
+    | Simple ->
+        "simple"
+    | OneShot ->
+        "oneshot"
+    | Forking ->
+        "forking"
```
edwintorok, line 39: Can you add `notify` here? Full list of types: https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#Type=, and in our version of systemd the manpage says:
`simple, forking, oneshot, dbus, notify, or idle`.

`dbus` we can probably leave out (we don't have dbus, we have message-switch).

edwintorok (approved):

psafont (approved):

PR #5370, opened 2024-01-18
CA-387698: datamodel: eliminate next_release

robhoes:
Replace all occurances of `next_release` and `rel_next` with appropriate version numbers.

### DIFF
--- a/ocaml/idl/datamodel_common.ml
+++ b/ocaml/idl/datamodel_common.ml
@@ -10,7 +10,7 @@ open Datamodel_roles
               to leave a gap for potential hotfixes needing to increment the schema version.*)
 let schema_major_vsn = 5
 
-let schema_minor_vsn = 773
+let schema_minor_vsn = 774
 
 (* Historical schema versions just in case this is useful later *)
 let rio_schema_major_vsn = 5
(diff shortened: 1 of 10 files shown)


ocaml/idl/schematest.ml
```diff
@@ -3,7 +3,7 @@ let hash x = Digest.string x |> Digest.to_hex
 (* BEWARE: if this changes, check that schema has been bumped accordingly in
    ocaml/idl/datamodel_common.ml, usually schema_minor_vsn *)
```
psafont, line 4: image

robhoes: I did..............(though not strictly needed here)? 🤨

edwintorok (commented):
This is good, but how are we going to introduce new parameters now? We'll need to specify a version, but you don't have a version until you tag it.
Do we need something equivalent to the 'prototyped' lifecycle that we have for APIs, but for API named parameters?

ocaml/idl/datamodel_common.ml
```diff
@@ -349,9 +349,9 @@ let get_product_releases in_product_since =
   in
   go_through_release_order release_order
 
-let next_release =
+let numbered_release version =
```
lindig, line 352: I would have called this just "release"; there is no check that this is indeed numbered or using any particular format. Which could be done with a regular expression or Scanf.

lindig (approved):

robhoes:
>This is good, but how are we going to introduce new parameters now? We'll need to specify a version, but you don't have a version until you tag it.

We still need to automated this for parameters, like we do for fields and functions. Until then, I think the best option is to manually put something like `1.2.3-next` like the automation would do. At least then we have more info to fix things up then in the old case, which would make it just `next`.

ocaml/idl/datamodel_common.ml
robhoes: It's just to distinguish this case from the old releases, which have names.

psafont:
> I think the best option is to manually put something like 1.2.3-next

What's the best place to document this?

robhoes:
Perhaps I'll make that `numbered_release` accept only `x.y.z` and `x.y.z-next`, and raise an error from there explaining the rules.

psafont:
Conflicts are because the lifecycle got updated incorrectly, we should prioritise to merge this PR

psafont (approved):

